PR #1788, opened 2020-04-22
OpenTitan: README: Update the OpenTitan SHA

alistair23:
### Pull Request Overview

Update the doc SHA.

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.


bradjc:
Is this the commit you have been using for testing?

alistair23:
Yep.

You probably could use an older commit, but I don't think what's in the README now is correct and I figured it's a good one to update to.

Issue #994, opened 2018-06-13
RFC: Future of Tock Hardware


bradjc:
_Writing down some things from recent discussions._

With new tutorials coming up, the issue of what hardware to use is important, and getting an early start gives us time to make sure things will work. This then prompts the obvious question of what hardware do we want to support generally going forward.

Current Issues
----------------

- Doing tutorials with Hail doesn't seem to be getting us much buy-in, despite that Hail is available for purchase. It might be cost, or that researchers don't see how they can effectively use hail, or something else.
- The SAM4L is a source of numerous bugs and weirdnesses. It is also rather large and expensive. Plus it is difficult to get the hardware correct when making a new board.
- There is no LoRa support on any of our HW, yet many researchers are using LoRa radios.

Thoughts for the Tutorial
-------------------------------

- Want hardware that researchers can seem themselves building off of. This would either be for sensing applications or networking experimentation.
- Would like to do something with the 6lowpan stack.

Ideas
-----

- Use the nRF52840dk board with an arduino sensor shield. This is appealing because the nRF52840 is a very compelling chip that supports BLE and 15.4, the nRF52840dk is readily available, and arduino is popular. However, after looking into this, there seem to be almost no arduino sensor shields. It's a little crazy that that can be true, but it seems the arduino crowd uses jumpers and breadboards for their sensors.

Potential Future HW Platforms
----------------------

- The [ACD52832](https://www.mouser.com/ProductDetail/aconno/ACD52832?qs=sGAEpiMZZMve4%2fbfQkoj%252bAOKB4ochF2XCuH5Oopp1aE%3d) is very interesting. It has some useful peripherals, and they are working on a nRF52840 version.
- The [permamote](https://github.com/lab11/permamote) could be a very useful base for an energy-harevesting / low power Tock board.
- Something with LoRaWAN. Based on my searching, this would have to be a custom board as well.

brghena:
Two options on the arduino shield side:

- [Sparkfun Weather Shield](https://www.sparkfun.com/products/13956) which has temperature, humidity, pressure, and light (uses an SI7021 which we already have a driver for)
- [Grove kit](https://www.seeedstudio.com/Grove-Starter-Kit-for-Arduino-p-1855.html) has several sensors and actuators. Grove allows you to plug sensors into any of several available slots, so we'd really have to get our configuration story down to get this working.

dcellucci:
Hello! I thought I'd weigh in since I've been doing my own trades for my research. 

[...]

PR #1791, opened 2020-04-23
boards: add a 'where to get it' link

ppannuto:
If one of our supported board requirements is availability, let's
help folks out with that.

### Pull Request Overview

Adds where to get it column to boards.

### Testing Strategy

N/A

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.

---

Rendered: https://github.com/tock/tock/blob/where-to-get-it/boards/README.md

### DIFF
--- a/boards/README.md
+++ b/boards/README.md
@@ -4,32 +4,19 @@ Platforms Supported by Tock
 The `/boards` directory contains the physical hardware platforms
 that Tock supports.
 
-| Board                                                         | Architecture    | MCU            | Interface  | App deployment | Where to Get It                          |
+| Board                                                         | Architecture    | MCU            | Interface  | App deployment | More Information |
 |---------------------------------------------------------------|-----------------|----------------|------------|----------------|------------------------------------------|
-| [Hail](hail/README.md)                                        | ARM Cortex-M4   | SAM4LC8BA      | Bootloader | tockloader     | [Lab11llc][tockhw]                       |
-| [Imix](imix/README.md)                                        | ARM Cortex-M4   | SAM4LC8CA      | Bootloader | tockloader     | [Lab11llc][tockhw]                       |
-| [Nordic nRF52-DK](nordic/nrf52dk/README.md)                   | ARM Cortex-M4   | nRF52832       | jLink      | tockloader     | [Nordic distributors][nrf52dk-hw]        |
-| [Nordic nRF52840-DK](nordic/nrf52840dk/README.md)             | ARM Cortex-M4   | nRF52840       | jLink      | tockloader     | [Nordic distributors][nrf52840dk-hw]     |
-| [Nordic nRF52840-Dongle](nordic/nrf52840_dongle/README.md)    | ARM Cortex-M4   | nRF52840       | jLink      | tockloader     | [Nordic distributors][nrf52840dongle-hw] |
-| [ACD52832](acd52832/README.md)                                | ARM Cortex-M4   | nRF52832       | jLink      | tockloader     | [Aconno][aconno]                         |
-| [TI LAUNCHXL-CC26x2](launchxl/README.md)                      | ARM Cortex-M4   | CC2652R        | openocd    | tockloader     | [TI or distributors][launchxl-hw]        |
-| [ST Nucleo F446RE](nucleo_f446re/README.md)                   | ARM Cortex-M4   | STM32F446      | openocd    | custom         | [ST or distributors][f446re-hw]          |
-| [ST Nucleo F429ZI](nucleo_f429zi/README.md)                   | ARM Cortex-M4   | STM32F429      | openocd    | custom         | [ST distributors][f429zi-hw]             |
-| [STM32F3Discovery kit](stm32f3discovery/README.md)            | ARM Cortex-M4   | STM32F303VCT6  | openocd    | custom         | [ST distributors][discovery-hw]          |
-| [SiFive HiFive1](hifive1/README.md)                           | RISC-V          | FE310-G000     | openocd    | tockloader     | [Crowdsupply][hifive1-revB-hw]*          |
-| [Digilent Arty A-7 100T](arty-e21/README.md)                  | RISC-V RV32IMAC | SiFive E21     | openocd    | tockloader     | [Digilent or distributors][arty-hw]      |
-| [Nexys Video OpenTitan](opentitan/README.md)                  | RISC-V RV32IMC  | Ibex           | custom     | custom         | _See board README_                       |
+| [Hail](hail/README.md)                                        | ARM Cortex-M4   | SAM4LC8BA      | Bootloader | tockloader     | https://tockos.org/hardware/#hail
+| [Imix](imix/README.md)                                        | ARM Cortex-M4   | SAM4LC8CA      | Bootloader | tockloader     | https://tockos.org/hardware/#imix
+| [Nordic nRF52-DK](nordic/nrf52dk/README.md)                   | ARM Cortex-M4   | nRF52832       | jLink      | tockloader     | https://tockos.org/hardware/#nrf52-family
+| [Nordic nRF52840-DK](nordic/nrf52840dk/README.md)             | ARM Cortex-M4   | nRF52840       | jLink      | tockloader     | https://tockos.org/hardware/#nrf52-family
+| [Nordic nRF52840-Dongle](nordic/nrf52840_dongle/README.md)    | ARM Cortex-M4   | nRF52840       | jLink      | tockloader     | https://tockos.org/hardware/#nrf52-family
+| [ACD52832](acd52832/README.md)                                | ARM Cortex-M4   | nRF52832       | jLink      | tockloader     | https://tockos.org/hardware/#aconno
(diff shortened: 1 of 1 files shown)


boards/README.md
```diff
@@ -4,17 +4,30 @@ Platforms Supported by Tock
...
+| [Nordic nRF52840-DK](nordic/nrf52840dk/README.md)    | ARM Cortex-M4   | nRF52840       | jLink      | tockloader     | [Nordic distributors][nrf52840dk-hw] |
+| [ACD52832](acd52832/README.md)                       | ARM Cortex-M4   | nRF52832       | jLink      | tockloader     | [Aconno][aconno]                     |
+| [TI LAUNCHXL-CC26x2](launchxl/README.md)             | ARM Cortex-M4   | CC2652R        | openocd    | tockloader     | [TI or distributors][launchxl-hw]    |
+| [ST Nucleo F446RE](nucleo_f446re/README.md)          | ARM Cortex-M4   | STM32F446      | openocd    | custom         | [ST or distributors][f446re-hw]      |
+| [ST Nucleo F429ZI](nucleo_f429zi/README.md)          | ARM Cortex-M4   | STM32F429      | openocd    | custom         | [ST distributors][f429zi-hw]         |
+| [STM32F3Discovery kit](stm32f3discovery/README.md)   | ARM Cortex-M4   | STM32F303VCT6  | openocd    | custom         | [ST distributors][discovery-hw]      |
+| [SiFive HiFive1](hifive1/README.md)                  | RISC-V          | FE310-G000     | openocd    | tockloader     | [Crowdsupply][hifive-revB-hw]*       |
```
bradjc, line 18:
```suggestion
| [SiFive HiFive1](hifive1/README.md)                  | RISC-V          | FE310-G000     | openocd    | tockloader     | *[Crowdsupply][hifive-revB-hw]       |
```
Maybe? The asterisk after doesn't render correctly.

boards/README.md
```diff
@@ -4,17 +4,30 @@ Platforms Supported by Tock
...
-| [Nexys Video OpenTitan](opentitan/README.md)         | RISC-V RV32IMC  | Ibex           | custom     | custom         |
+| Board                                                | Architecture    | MCU            | Interface  | App deployment | Where to Get It                      |
+|------------------------------------------------------|-----------------|----------------|------------|----------------|--------------------------------------|
+| [Hail](hail/README.md)                               | ARM Cortex-M4   | SAM4LC8BA      | Bootloader | tockloader     | [Lab11llc][tockhw]                   |
+| [Imix](imix/README.md)                               | ARM Cortex-M4   | SAM4LC8CA      | Bootloader | tockloader     | [Lab11llc][tockhw]                   |
+| [Nordic nRF52-DK](nordic/nrf52dk/README.md)          | ARM Cortex-M4   | nRF52832       | jLink      | tockloader     | [Nordic distributors][nrf52dk-hw]    |
+| [Nordic nRF52840-DK](nordic/nrf52840dk/README.md)    | ARM Cortex-M4   | nRF52840       | jLink      | tockloader     | [Nordic distributors][nrf52840dk-hw] |
```
gendx, line 12: I just noticed that the nRF52840-Dongle is missing from this list (https://www.nordicsemi.com/Software-and-tools/Development-Kits/nRF52840-Dongle/GetStarted and https://www.nordicsemi.com/About-us/BuyOnline?search_token=nRF52840DONGLE&series_token=nRF52840).

ppannuto: Ah! Thanks, just added.

lschuermann:
I'm really conflicted about this. I don't think that Tock as a project should endorse any vendor or distributor. We are kind of doing this already by supporting boards of different vendors, but that can't be avoided.

While this in its current state arguably does not endorse any vendor or more important any distributor specifically, it gets us a lot closer to doing that. Once this table is in place, its easy to accidentally or deliberately add a link to a specific distributor - something I'd like to avoid at all cost. Also, linking to the _buy now_ page of a vendor could be interpreted as us trusting or being sponsored by this vendor. I'd much rather see a link to the product overview page, or something comparable.

gendx:
> I'm really conflicted about this. I don't think that Tock as a project should endorse any vendor or distributor. We are kind of doing this already by supporting boards of different vendors, but that can't be avoided.
> 
> While this in its current state arguably does not endorse any vendor or more important any distributor specifically, it gets us a lot closer to doing that. Once this table is in place, its easy to accidentally or deliberately add a link to a specific distributor - something I'd like to avoid at all cost. Also, linking to the _buy now_ page of a vendor could be interpreted as us trusting or being sponsored by this vendor. I'd much rather see a link to the product overview page, or something comparable.

I tend to agree. The proposed "buy now" links point to a page that directly dumps a list of prices, without a description of the board.

As an alternative to the "buy now" page, there could be a link to the product description page. For example, in https://github.com/tock/tock/blob/master/boards/nordic/nrf52840dk/README.md we already have a link to https://www.nordicsemi.com/Software-and-Tools/Development-Kits/nRF52840-DK, so we could reuse that instead.

This page from the manufacturer presents the product characteristics, and also includes a "buy now" button, so IMO this is a good trade-off, because users falling on this page can:
- make an informed decision about whether the board fits their needs,
- easily find where to buy it, if they want to.

ppannuto:
I'm not sure I see this as endorsing a board or vendor (indeed, they all point to the generic list of distributors rather than any one distributor (except when there is only one distributor)). I see this more as a pragmatic step, cutting out the intermediate search of 'where to get it'.

More importantly, however, one of the early principles we laid out is that boards in mainline Tock *must* be easy to acquire. There is a problem historically with embedded OSes that they "support" a long list of obsolete boards. I very intentionally pointed these links to "buy" pages because for a board to be included in mainline Tock, a buy page must exist; generic product description pages do not fulfill this requirement. For this reason, I would like the addition of a "you can get this HW here" link as a hard gating mechanism for inclusion.

I could see a compromise point of adding an additional "Description / Specs" column before there "Where to get it" column?

gendx:
> More importantly, however, one of the early principles we laid out is that boards in mainline Tock _must_ be easy to acquire. There is a problem historically with embedded OSes that they "support" a long list of obsolete boards. I very intentionally pointed these links to "buy" pages because for a board to be included in mainline Tock, a buy page must exist; generic product description pages do not fulfill this requirement. For this reason, I would like the addition of a "you can get this HW here" link as a hard gating mechanism for inclusion.

On the other hand, I want to point out that Tock's official boards (https://www.tockos.org/hardware/) don't really fall in the "easy" to acquire category. I'd rather say they are "possible" to acquire:
- Hail is currently "out of stock".
- Imix "will be available in the coming weeks". There's no timestamp, but as far as I remember it was like this in 2019 as well. One can "contact Amit", but I wouldn't consider it "easy" to acquire from the perspective of a newcomer (who's Amit? what's the delivery time? in which countries? how much does it cost? is there any in stock?).
  - Now, if one manually visits https://www.tockos.org/hardware/imix, it's available for 100$, but this page is not linked from the front page (I just remembered that there was such a page and randomly tried to append `/imix` to the URL).

So I think we should fix the website. It's an easily accessible entry-point to Tock, and should be up-to-date. I may be wrong, but I think that newcomers are more likely to visit https://www.tockos.org/hardware/ than https://github.com/tock/tock/tree/master/boards (at least when I was new to Tock I started by visiting the https://www.tockos.org/ website).

Should the webpage at https://www.tockos.org/hardware/ be updated to include all the boards of this README?

ppannuto:
Yes! We should absolutely update that page.

Hail is a slightly complicated question. We never really wanted to have a custom board as the "default" for Tock, but we also really wanted one that was in some way "interesting" out-of-the-box; that is, something that ships with sensors on board that could run more interesting applications than blink. We (really Brad, thanks Brad!) have been tracking replacement candidates here: https://github.com/tock/tock/issues/994

Our last hail run finally ran out of stock ~2 weeks ago [what timing..], and I don't think we've strongly concluded or agreed to this, but I _think_ the general feeling among @bradjc @brghena and myself is that we would like to sunset hail and replace it with something mass-produced. That would likely include deprecating hail as a Tock platform. Running out of hails does motivate that much more now.

---

I think having one canonical source for hardware information would be better. I have to catch a talk, but I'll update this PR in a bit to point to the Tock hardware website for "where to get it" information.

gendx:
> We (really Brad, thanks Brad!) have been tracking replacement candidates here: #994

Thanks for pointing it out!

ppannuto:
Blocked on https://github.com/tock/tock-www/pull/46

ppannuto:
This is no longer blocked.

bradjc:
I thought this was a great idea when it linked to an order page. Now I'm not sure if this is worth keeping up to date. I also think we should encourage in general "more information" to be directly included in the board readme where we control it and the information is centralized. For example, I can imagine that if I was an external user, clicking on the opentitan link would be frustrating.

ppannuto:
Yeah. I too am underwhelmed by these links now :)

gendx:
However, adding the nrf52840-dongle to this list (https://github.com/tock/tock/pull/1791#discussion_r414694585) would still be useful.

PR #1793, opened 2020-04-23
Hifive updates

bradjc:
### Pull Request Overview

Some updates I found while trying to get hifive1a to work.

- Enable external interrupts (which is how the PLIC is connected).
- Better label LEDs.
- Fix register field name in PWM.

### Testing Strategy

This pull request was tested by running the kernel on a hifive1a board.

It doesn't seem to really work, perhaps because there is a bootloader that runs.

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.

### DIFF
--- a/chips/sifive/src/pwm.rs
+++ b/chips/sifive/src/pwm.rs
@@ -61,6 +61,6 @@ impl Pwm {
         let regs = self.registers;
 
         // Turn the interrupt compare off so we don't get any RTC interrupts.
-        regs.cfg.write(cfg::enalways::CLEAR + cfg::enalways::CLEAR);
+        regs.cfg.write(cfg::enalways::CLEAR + cfg::enoneshot::CLEAR);
     }
 }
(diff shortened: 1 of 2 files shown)


chips/sifive/src/pwm.rs
```diff
@@ -61,6 +61,6 @@ impl Pwm {
         let regs = self.registers;
 
         // Turn the interrupt compare off so we don't get any RTC interrupts.
-        regs.cfg.write(cfg::enalways::CLEAR + cfg::enalways::CLEAR);
+        regs.cfg.write(cfg::enalways::CLEAR + cfg::enoneshot::CLEAR);
```
ppannuto, line 64: Should this possibly be both lines rather than a replacement (clear the 'always' (level?) interrupt and the 'oneshot' interrupt)?

bradjc: Before it was setting the same bit twice, and I think my intention was to clear both bits.

ppannuto: Wow..., yes, yes it was, I can read.

Would be interesting to see whether such an error could be automatically caught, it's meaningless to 'add' a bit to itself in a mask like it was.

PR #1422, opened 2019-10-22
Add support for multiple GPIO ports in chips/nrf5x.

gendx:
### Pull Request Overview

This pull request add support for multiple GPIO ports in chips/nrf5x.

Enabling of the pins is done with a `nrf52840` feature, which adds the port 1 pins available on the respective board. This feature is not yet enabled on the board, it will follow in further pull-requests/commits.

### Testing Strategy

This pull request was tested by:

- Running `make ci-travis`.
- Flashing Tock on an nRF52840-dk board and running a simple program (blink).

### TODO or Help Wanted

N/A

### Documentation Updated

- [ ] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

PR #1794, opened 2020-04-23
import cfg_if, use for stm IRQS

hudson-ayers:
### Pull Request Overview

This pull request copies the cfg_if!() crate (https://github.com/alexcrichton/cfg-if) into the kernel, and uses to configure which `IRQS` array is enabled in `stm32f4xx/src/lib.rs`. This change prevents both arrays being enabled simultaneously, which fixes clippy when run from the root cargo workspace (currently it fails because clippy runs in parallel across all crates in the workspace, such that both board features are enabled when the stm324xx crate is checked).

Notably, `cfg_if!()` is widely used throughout the Rust ecosystem, and is used internally in libstd and libcore. 

### Testing Strategy

This pull request was tested by building all stm boards and checking that clippy no longer fails on `IRQS` being defined multiple times.

### TODO or Help Wanted

Does this belong in libraries instead? It is not a crate written by us, so the README there doesn't totally apply.

I included the MIT license from the crate at the top of the file, is that okay?

Should I include the tests copied from the original crate or not?

Should we use this macro elsewhere? Currently this is the only location where clippy requires it,
but I think it could improve readability of `cfgs` elsewhere.

### Documentation Updated

- [x] No updates required.

### Formatting

- [x] Ran `make formatall`.


ppannuto:
>> Does this belong in libraries instead?

> https://github.com/tock/tock/blob/master/doc/Design.md#no-external-dependencies
>
> We also realize, however, that external libraries can be very useful. Tock's compromise has been to pull in specific portions of libraries into the libraries folder. This puts the library's source in the same repository, while keeping the library as a clearly separate crate. We do try to limit how often this happens.

>> It is not a crate written by us, so the README there doesn't totally apply.

We should probably fix the README to match our policy :)

---

We *might* consider addressing the license issue by changing tacks slightly and re-exporting an external dependency* pinned to `cfg-if@0.1.10` [there have been two commits since then, but they're just gh-pages stuff] from the libraries crate.

*With a block comment above in the toml that includes
 - rationale for exception: It's used heavily by Rust internals
 - list of tock-core members who audited that release

The `cfg-if` crate has two dependencies:
 - `compiler_builtins ^0.1.2` -- [part of rust-lang](https://github.com/rust-lang/compiler-builtins) and something we have depended on before (and may still be?)
 - `rustc-std-workspace-core ^1.0.0` -- [not a real dependency](https://github.com/rust-lang/rust/tree/master/src/tools/rustc-std-workspace-core)

My understanding is that published crates on cargo [are immutable](https://doc.rust-lang.org/cargo/reference/publishing.html) (unlike pointing to a release tag on github, which could be changed).

gendx:
> My understanding is that published crates on cargo [are immutable](https://doc.rust-lang.org/cargo/reference/publishing.html)

Correct, but note that one needs to write the dependency in `Cargo.toml` as `=1.2.3` to pin to it. And I think that recursive dependencies must also be declared and pinned in `Cargo.toml`. The alternative is to provide a `Cargo.lock` file with the pinned versions in it.

> (unlike pointing to a release tag on github, which could be changed).

Pointing to a commit instead of a tag should work (as long as a forced push doesn't remove the commit, in which case it may not be available anymore).

hudson-ayers:
Pinning it as an external dependency has some downsides -- you have to pin it in the `Cargo.toml` for each crate that uses it, and they need to each pin the same version (or at least have compatible version requests). For this specific PR, the only change needed would be depending on `cfg_if` in the stm32f4xx crate, but this is not a great place to centralize notes like the rationale for exception and the Tock core members that audited the release -- especially because if `cfg_if` is used in additional crates in the future we would end up having to duplicate that comment in each Cargo.toml file. (Alternatively we could have the `kernel::common` crate export `cfg_if`, but that is sort of too bad because the Kernel doesn't need `cfg_if` at all anyway).

> The alternative is to provide a Cargo.lock file with the pinned versions in it.

My understanding is that Cargo.lock should not be manually edited, but so long as each crate pins the same version in Cargo.toml, Cargo.lock will resolve the same dependency for each crate using the dependency.

Ultimately I think it is cleaner to just include it as a library, assuming that the license issue is acceptable (it seems like having an MIT licensed file in libraries should be okay?).

gendx:
As an alternative to Cargo features (cc @bradjc), a pattern we've used for nrf52x chips is to have a hierarchy of crates that depend on some base implementation, add some chip-specific elements on top of it, and re-export all the modules in the final crate. This works quite well.

Something that also works well is the [`InterruptService` pattern](https://github.com/tock/tock/blob/master/chips/nrf52/src/interrupt_service.rs), which I think could be generalized to all the chips.

Somehow, the nrf52x chips all have the same uninteresting `IRQS` array, I'm not sure what we should do about that.

https://github.com/tock/tock/blob/446df4a4d8dc8c6263d66da1ecf2180240e29c66/chips/nrf52/src/crt1.rs#L91-L97

Is there some Tock-wide policy to have some consistent style regarding IRQs across all chips & boards?

hudson-ayers:
For something like the IRQS array for the stm32f4xx, it seems unfortunate that we would have to create two additional crates (stm32f446re and stm32f429zi) exclusively to have a file to export the IRQS array, when all of the other chip code is identical between the two chips. 

But it does seem true that we could use this pattern to move away from all use of cfg except for enabling tests to run on development platforms.

ppannuto:
Maybe I mis-understand dependency resolution, but I think the suggestion I was making is something more like this: https://github.com/tock/tock/commit/53f3294993de28c54026aaa9c8329c5bc332bb08

I don't think this should require any lockfiles, and any part of Tock could safely use this "crate". The auditing happens in exactly one place and the rationale and audit results always live in the "external" crate.

hudson-ayers:
> Maybe I mis-understand dependency resolution, but I think the suggestion I was making is something more like this: [53f3294](https://github.com/tock/tock/commit/53f3294993de28c54026aaa9c8329c5bc332bb08)

Oh, I actually misunderstood your original suggestion. That looks great. I vote we move to this approach

gendx:
> For something like the IRQS array for the stm32f4xx, it seems unfortunate that we would have to create two additional crates (stm32f446re and stm32f429zi) exclusively to have a file to export the IRQS array, when all of the other chip code is identical between the two chips.

I ended up introducing the new crates in #1422 (a change that would have been quite simple with features - like we have here), in particular after the following review: https://github.com/tock/tock/pull/1422#issuecomment-545900262 (cc @bradjc).

Let's be consistent when it comes to features: why should features be banned from Nordic chips but allowed in stm32f4xx chips? Given how much this has been debated in various pull requests, I think there should be a clear decision regarding the use of features in Tock.

Now, what would be the overhead of such chip crates? Not much actually:
- A `README.md` + `Cargo.toml` + `lib.rs`
- The `lib.rs` is quite thin actually. See for example the `nrf52840`. It re-exports common symbols from the common crate - just one line. https://github.com/tock/tock/blob/119957277bd1ea209ff9c20a122b0216ad96b3af/chips/nrf52840/src/lib.rs#L3-L6  And following that we can have the chip-specific parts (for a start, the IRQS).

Overall, I think that adding the `cfg_if` dependency goes against the no-feature principle of Tock (besides going against the no-dependency principle).

hudson-ayers:
@gendx That is pretty convincing. I think you are right that we should be consistent here, and that importing cfg_if seems generally to run counter to two established Tock policies. I don't want to reignite the discussion from #1422 . 

I suppose I should try to migrate the stm32f4 chips to the multiple chip crates approach instead.
Unless anyone feels strongly otherwise, I will close this PR soon in favor of that change.

(As an aside, I am starting to see why we adding clippy to travis has taken so long..doing so seems to rely on fixing a lot of stuff that has slipped through the cracks)

hudson-ayers:
Closing in favor of #1803

PR #1512, opened 2019-12-17
Add cargo test for capsules on ci-travis.

gendx:
### Pull Request Overview

This pull request adds `cargo test` for the `capsules` crate to the Makefile.

I added a simple test in the virtual alarm capsule, but in general I think that capsules should be of great interest for functional unit tests, due to the non-trivial logic and the abstraction provided by HILs. In particular, one can write mock implementations of HILs to simulate various behaviors and test that the capsules react as expected.

### Testing Strategy

This pull request was tested by running `make ci-travis` and checking that unit tests indeed appear in the log.

### TODO or Help Wanted

- This pull request still needs fixing a bunch of doc tests in capsules.
- I hope that we can agree on the usefulness of such tests now that a lot of discussion happened on https://github.com/tock/tock/pull/1393. 

### Documentation Updated

- [ ] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

PR #1795, opened 2020-04-24
tools: update makefile to partition CI

ppannuto:
This eases running the failing portions of CI locally

### Pull Request Overview

Inspired by moving rules around while working on #1512, this partitions each phase of CI into its own rule. It should effectively be a no-op.

### Testing Strategy

This pull request was tested by making my poor laptop very warm by running lots of `make ci-X`

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,64 +39,140 @@ allcheck:
 alldoc:
 	@for f in `./tools/list_boards.sh`; do echo "$$(tput bold)Documenting $$f"; $(MAKE) -C "boards/$$f" doc || exit 1; done
 
+
+
+###################################################################
+##
+## Continuous Integration Targets
+##
+## To run all CI locally, use the meta-target `make ci`.
+##
+## Each of the phases of CI is broken into its own target to enable
+## quick local iteration without re-running all phases of CI.
+##
+
+
+## Meta-Targets
+
+.PHONY: ci
+ci: ci-travis ci-netlify
+
 .PHONY: ci-travis
-ci-travis:
+ci-travis:\
+	ci-formatting\
+	ci-tools\
+	ci-libraries\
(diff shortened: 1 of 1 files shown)


ppannuto:
Added bonus: Cargo is smart enough to lock build directories when needed, so parallel builds also seem to work. On my laptop, after an already successful run (i.e. everything already built):

```
$ time make ci-travis
...
real	1m10.186s
user	0m42.431s
sys	0m30.381s
$ time make -j4 ci-travis
real	0m41.973s
user	0m48.517s
sys	0m33.782s
```

Makefile
```diff
@@ -39,64 +39,140 @@ allcheck:
...
+ci-compilation:
 	@printf "$$(tput bold)*******************$$(tput sgr0)\n"
 	@printf "$$(tput bold)* CI: Compilation *$$(tput sgr0)\n"
 	@printf "$$(tput bold)*******************$$(tput sgr0)\n"
 	@CI=true $(MAKE) allboards
+
+.PHONY: ci-special-targets
+ci-special-targets:
 	@printf "$$(tput bold)***********************$$(tput sgr0)\n"
 	@printf "$$(tput bold)* CI: Special Targets *$$(tput sgr0)\n"
```
gendx, line 147: Could we rename this one to something more descriptive? I tend to keep asking myself what those special targets are when seeing this part in the console, especially given that there is no console message at all mentioning the `lst` part, which by the way is quite slow.

ppannuto: Good point. Also added a comment explaining a bit more what's built and why.

hudson-ayers:
bors retry

PR #1796, opened 2020-04-24
mark get_pin_mut() as unsafe

hudson-ayers:
### Pull Request Overview

This pull request adds unsafe to `get_pin_mut()` to indicate that calling this function could allow for multiple mutable references to a single pin. Currently this function was only called in `panic_fmt()` for various boards, so none of the calling instances had to be modified. 

I also added a clippy allow associated with this behavior to indicate what we are allowing is (I think) effectively no different than if the calling instances directly accessed the global `PINS` variable.

I am in favor of this fix being included in 1.5.

### Testing Strategy

This pull request was tested by running `run_clippy.sh` on a branch with a number of other clippy fixes included, and by travis-ci.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.

### DIFF
--- a/chips/stm32f3xx/src/gpio.rs
+++ b/chips/stm32f3xx/src/gpio.rs
@@ -472,7 +472,11 @@ impl PinId {
         unsafe { &PIN[usize::from(port_num)][usize::from(pin_num)] }
     }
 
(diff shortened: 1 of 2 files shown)


ppannuto:
bors d+

PR #1797, opened 2020-04-24
Add a git-diff line to tools/post_size_changes_to_github.sh

gendx:
### Pull Request Overview

I noticed that the size diffing tool has been added back in #1753. However, I find it hard to pinpoint which sections of the code have the most impact in the size difference, because the diff only summarizes global flash and RAM differences.
However, the size reports are more detailed than that.

This pull request adds a simple `git diff` command over the raw size reports. Although it doesn't compute the difference as a number or percentage, it's a low-hanging fruit which can help understanding where the difference comes from.

### Testing Strategy

This pull request was tested by Travis.

### TODO or Help Wanted

- [ ] Should this line be part of `tools/diff_memory_usage.py` instead?

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.

### DIFF
--- a/tools/post_size_changes_to_github.sh
+++ b/tools/post_size_changes_to_github.sh
@@ -58,6 +58,10 @@ if [ -n "$TRAVIS_PULL_REQUEST_BRANCH" ]; then
     for elf in $(find . -maxdepth 8 | grep 'release' | egrep '\.elf$' | grep -v 'riscv'); do
         tmp=${elf#*release/}
         b=${tmp%.elf}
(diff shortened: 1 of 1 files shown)


gendx:
Should this be merged or is it superseded by the GitHub workflows?

hudson-ayers:
I think this should still be merged as long as we are still using Travis, once we move away from Travis this file will be deleted. But it is nice to have the info in both places for now.

hudson-ayers:
bors d+

PR #1799, opened 2020-04-25
nRF52 spi: fix get_rate

lschuermann:
# Pull Request Overview
The nRF52 SPI implementation's get_rate() method has been returning the actual register content instead of the converted SPI operating frequency.

This PR fixes this issue specifically. However, because the `Frequency` enum relates to two `u32`s, namely the register content and the actual frequency, using `From<>` traits are confusing which is likely how the bug was introduced originally. Those traits are replaced with methods indicating their purpose.

The back-conversion of the register content to a frequency may fail if the register contains an invalid value. As the reset value is valid and all other values written should be valid too, this should work. Nonetheless, a `.expect()` must be introduced.

### Testing Strategy

Tested on nRF52DK.

```rust
    debug!(
        "Old rate: {:?}",
        hil::spi::SpiMaster::get_rate(&nrf52::spi::SPIM0)
    );
    debug!(
        "Set rate: {:?}",
        hil::spi::SpiMaster::set_rate(&nrf52::spi::SPIM0, 5_000_000)
    );
    debug!(
        "New rate: {:?}",
        hil::spi::SpiMaster::get_rate(&nrf52::spi::SPIM0)
    );
```

Output on latest master:
```
NRF52 HW INFO: Variant: AAE0, Part: N52832, Package: QF, Ram: K64, Flash: K512
Old rate: 67108864
Set rate: 4000000
New rate: 1073741824
```

Output with this PR:
```
NRF52 HW INFO: Variant: AAE0, Part: N52832, Package: QF, Ram: K64, Flash: K512
Old rate: 250000
Set rate: 4000000
New rate: 4000000
```

Verifying the actual SPI rate will need to wait for me to get my logic analyzer.

### TODO or Help Wanted

- [ ] The `nRF52840` also supports 16 and 32 Mbps SPI rates. Those should probably be implemented using compile features on the chip crate. Can we do it in this PR as well, or should I open a new one?

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.

### DIFF
--- a/chips/nrf52/src/spi.rs
+++ b/chips/nrf52/src/spi.rs
@@ -190,9 +190,22 @@ pub enum Frequency {
     M8 = 0x80000000,
 }
 
-impl From<Frequency> for u32 {
-    fn from(freq: Frequency) -> u32 {
-        match freq {
+impl Frequency {
+    pub fn from_register(reg: u32) -> Option<Frequency> {
+        match reg {
+            0x02000000 => Some(Frequency::K125),
+            0x04000000 => Some(Frequency::K250),
+            0x08000000 => Some(Frequency::K500),
+            0x10000000 => Some(Frequency::M1),
+            0x20000000 => Some(Frequency::M2),
+            0x40000000 => Some(Frequency::M4),
+            0x80000000 => Some(Frequency::M8),
+            _ => None,
+        }
+    }
+
+    pub fn into_spi_rate(&self) -> u32 {
+        match *self {
             Frequency::K125 => 125_000,
             Frequency::K250 => 250_000,
             Frequency::K500 => 500_000,
@@ -202,10 +215,8 @@ impl From<Frequency> for u32 {
             Frequency::M8 => 8_000_000,
         }
     }
-}
 
-impl From<u32> for Frequency {
-    fn from(freq: u32) -> Frequency {
+    pub fn from_spi_rate(freq: u32) -> Frequency {
         if freq < 250_000 {
             Frequency::K125
         } else if freq < 500_000 {
@@ -413,14 +424,19 @@ impl hil::spi::SpiMaster for SPIM {
     // Returns the actual rate set
     fn set_rate(&self, rate: u32) -> u32 {
         debug_assert!(self.initialized.get());
-        let f = Frequency::from(rate);
+        let f = Frequency::from_spi_rate(rate);
         self.registers.frequency.set(f as u32);
-        f.into()
+        f.into_spi_rate()
     }
 
     fn get_rate(&self) -> u32 {
         debug_assert!(self.initialized.get());
-        self.registers.frequency.get().into()
+
+        // Reset value is a valid frequency (250kbps), so .expect
+        // should be safe here
+        let f = Frequency::from_register(self.registers.frequency.get())
+            .expect("nrf52 unknown spi rate");
+        f.into_spi_rate()
     }
 
     fn set_clock(&self, polarity: hil::spi::ClockPolarity) {


chips/nrf52/src/spi.rs
```diff
@@ -190,9 +190,22 @@ pub enum Frequency {
     M8 = 0x80000000,
 }
 
-impl From<Frequency> for u32 {
-    fn from(freq: Frequency) -> u32 {
-        match freq {
+impl Frequency {
+    pub fn from_register(reg: u32) -> Option<Frequency> {
```
niklasad1, line 194: why not implement `TryFrom` instead?

chips/nrf52/src/spi.rs
```diff
@@ -202,10 +215,8 @@ impl From<Frequency> for u32 {
             Frequency::M8 => 8_000_000,
         }
     }
-}
 
-impl From<u32> for Frequency {
-    fn from(freq: u32) -> Frequency {
+    pub fn from_spi_rate(freq: u32) -> Frequency {
```
niklasad1, line 219: this still seems error-prone but seems to only be used internally with validated input? Perhaps make it private or document it?

chips/nrf52/src/spi.rs
```diff
@@ -413,14 +424,19 @@ impl hil::spi::SpiMaster for SPIM {
     // Returns the actual rate set
     fn set_rate(&self, rate: u32) -> u32 {
```
niklasad1, line 425: hmm, I don't like this `set_rate` and `get_rate` API. Do you know the reason why we cant have a more typed API such as:

- `fn set_rate(&self, freq: Frequency)`
- `fn get_rate(&self) -> Frequency`

?

chips/nrf52/src/spi.rs
lschuermann: I removed all `From<>` trait implementations in an effort to avoid confusion. Implementing `TryFrom<u32> for Frequency` again doesn't clearly state, whether we're talking about register value or frequency.

chips/nrf52/src/spi.rs
lschuermann: I didn't really touch the existing methods because I don't want to break any existing changes. It always returns a `Frequency` for every input, namely one that is lower or equal to the requested one, except for the lowest (which may be higher than requested). I think this makes sense in this context. I'm not yet sure why the `Frequency` enum is public at all - making it private would limit those methods to be called inside this module anyways.

chips/nrf52/src/spi.rs
lschuermann: It's not great, but I don't think we can do much better easily. For instance, the SiFive SPI implementation allows setting arbitrary SPI clocks as the register value is just a divider of the bus frequency. Ultimately, having an enum of all allowed frequency values would get very large very quickly - and will set a deviating frequency on boards where the requested one isn't available (so no real benefit here).

niklasad1: Ok, I see. We could do is to make `Frequency` an associated type and each impl has to decide how to construct it but it is out-of-scope for this PR.

chips/nrf52/src/spi.rs
niklasad1: Ok, but in this case it is the frequency value, no?

niklasad1: Ok, I get your point

chips/nrf52/src/spi.rs
niklasad1: Right, I got confused by `register_value` vs the `actual rate`

chips/nrf52/src/spi.rs
lschuermann: Yes, I thought about that for a while. I'm not sure whether that is worth the additional effort for clients. We'd need to at least make it a `dyn Trait` which allows calling a `frequency_lower_eq()`, returning a Frequency variant being lower or equals to the requested one. Otherwise HW drivers will be unable to automatically set the required frequency agnostic of the underlying SPI implementation.

chips/nrf52/src/spi.rs
```diff
@@ -190,9 +190,22 @@ pub enum Frequency {
...
+impl Frequency {
+    pub fn from_register(reg: u32) -> Option<Frequency> {
+        match reg {
+            0x02000000 => Some(Frequency::K125),
+            0x04000000 => Some(Frequency::K250),
+            0x08000000 => Some(Frequency::K500),
+            0x10000000 => Some(Frequency::M1),
+            0x20000000 => Some(Frequency::M2),
+            0x40000000 => Some(Frequency::M4),
+            0x80000000 => Some(Frequency::M8),
```
bradjc, line 202: image

Shouldn't this have more fields?

lschuermann: Yes, I've stated this in my todos. However the highest frequencies only apply to the nRF52840. I can fix that in this PR too, or create another one for that.

PR #1773, opened 2020-04-20
Fix clippy lints + Run Clippy in Travis-CI

hudson-ayers:
### Pull Request Overview

This pull request fixes all of the errors triggered when running `run_clippy.sh` on master.
It also removes `check_wildcard_imports.sh` in favor of the recently introduced clippy lint that handles that same task (https://github.com/rust-lang/rust-clippy/pull/5029). It also adds a call to `run_clippy.sh` to Travis-CI to prevent future clippy regressions.

In order to get `run_clippy.sh` to pass, I had to make a few noteworthy changes:
- Fixed a permissions check in `network_capabilities.rs` that was previously incorrect (yay clippy)
- I had to add `#[cfg(not(feature = "cargo-clippy"))]` to the definition of one of the stm32 `IRQS` arrays so that clippy would not detect `IRQS` as being defined twice
- Added "-A clippy::enum_clike_unportable_variant" because otherwise I got several errors associated with `tock_registers`, such as:
```rust
error: Clike enum variant discriminant is not portable to 32-bit targets
   --> chips/nrf52/src/ficr.rs:77:1
    |
77  | / register_bitfields! [u32,
78  | |     /// Code memory page size
79  | |     CodePageSize [
80  | |         /// Code memory page size
...   |
235 | |     ]
236 | | ];
    | |__^
    |
    = note: `#[deny(clippy::enum_clike_unportable_variant)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#enum_clike_unportable_variant
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```
  I am not sure if we should be concerned about that error?

- Removed "-D clippy::default_trait_access" from `run-clippy.sh` because it resulted in like 100 errors and the suggested changes IMO added a ton of undesirable verbosity:
```rust
error: Calling `tock_cells::volatile_cell::VolatileCell<u8>::default()` is more clear than this expression
  --> capsules/src/usb/descriptors.rs:25:19
   |
25 |             buf: [Default::default(); 64],
   |                   ^^^^^^^^^^^^^^^^^^ help: try: `tock_cells::volatile_cell::VolatileCell<u8>::default()`
```

### Testing Strategy

This pull request was tested by Travis-CI, and by testing that adding a wildcard import in capsules causes `run_clippy.sh` to fail.

### TODO or Help Wanted

Are others on board with running clippy in Travis-CI? Should we add in additional lints?

[...]

PR #1648, opened 2020-03-03
kernel: use NonNull, iterate on grant and appslice interfaces

mcy:
### Pull Request Overview

This change is effectively `s/Unique/NonNull/g`, and is the supported upgrade path out of `ptr_internals`; it is effectively a no-op.

### Testing Strategy

`make allcheck` passes; presumably unit tests will catch any other silliness.

### TODO or Help Wanted

N/A

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

PR #1364, opened 2019-08-16
kernel: process/hdr: no clippy::cast_ptr_alignment

bradjc:
### Pull Request Overview

This pull request adds some clippy flags in the kernel to turn off errors for potentially misaligned pointers. This lint is by default an error, so we have to either change the pointers (not sure that is possible) or disable the lint for these two files.

https://rust-lang.github.io/rust-clippy/master/index.html#cast_ptr_alignment

### Testing Strategy

This pull request was tested by travis.

### TODO or Help Wanted

n/a

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

PR #1800, opened 2020-04-26
Remaining clippy allows/fixes

hudson-ayers:
### Pull Request Overview

This pull request includes just the fixes from #1773, except those included in ~~#1794~~ #1803 or #1792 or #1796. It does not add clippy to the CI pipeline so that change can be reviewed seperately from these fixes. Once this PR and #1803 are merged, `tools/run_clippy.sh` will pass.

It includes several clippy allows, with justifications included, as well as fixes requested by clippy. I had to apply the clippy fixes manually as `cargo clippy --fix` does not seem to work with our current nightly version.

### Testing Strategy

This pull request was tested by Travis-CI and `run_clippy.sh`.

### TODO or Help Wanted

N/A

### Documentation Updated

- [ ] No updates are required.

### Formatting

- [x] Ran `make formatall`.


boards/imix/src/test/ipv6_lowpan_test.rs
```diff
@@ -469,6 +469,10 @@ impl<'a, A: time::Alarm<'a>> TxClient for LowpanTest<'a, A> {
     }
 }
 
+#[allow(clippy::cast_ptr_alignment)]
+// This test contains non-idiomatic rust that involves casting
+// raw packet buffers to headers, but it works for this platform
+// and is just a test, so we included a clippy allow.
```
gendx, line 475: If it's just a test, do we care about the performance of making a copy instead of an `unsafe` memory transmute? There are reasons why this lint exists, and maybe so far it worked on this platform because the data was already aligned, but maybe it will break on this platform in the future if the data becomes unaligned.

I'd vote for fixing the problem rather than allowing this lint.

chips/nrf52/src/ficr.rs
```diff
@@ -127,6 +127,8 @@ register_bitfields! [u32,
     ],
     /// Part code
     InfoPart [
+        // clippy allow needed for same reason overflowing_literals needed below
+        #[allow(clippy::enum_clike_unportable_variant)]
```
gendx, line 131: Is there actually a reason for the `Unspecified` variant? IIRC this doesn't exist in the data sheet, so we could remove it, and instead use `Option<Enum>` to indicate an unspecified value (i.e. acceptable values will be `None`, and `Some(N528xx)` for the relevant `xx`).

cc @jmichelp.

kernel/src/process.rs
```diff
@@ -1197,6 +1197,7 @@ impl<C: Chip> ProcessType for Process<'a, C> {
         Some(grant_pointer)
     }
 
+    #[allow(clippy::cast_ptr_alignment)]
```
gendx, line 1200: IIRC we discussed it briefly in #1648.

There are reasons why this lint exists, and we should rather make sure that the data is indeed aligned as expected rather than ignoring the problem. In any case, this warrants at least a comment explaining why this lint is allowed.

I'd vote for fixing the problem rather than allowing this lint.

boards/imix/src/test/ipv6_lowpan_test.rs
hudson-ayers: Fixed, and tested again on Imix.

kernel/src/process.rs
hudson-ayers: `process.rs` currently has 4 other functions annotated with this allow, all seemingly used for casting `mem_end()` to different pointer types (`*mut *mut usize` or `*mut *mut u8` or `*const *mut u8`). 

The other allows were added in #1364, but looking back at that PR I am not sure there was any reason to believe those pointers are in fact guaranteed to be aligned? cc @bradjc

ppannuto: The grant pointer ones are safe, as that's all an internal walking of an array of pointers that's guaranteed to be aligned on word-boundaries due to MPU region restrictions.

The `sp` and tbf header stuff, however, is probably not safe.

boards/imix/src/test/ipv6_lowpan_test.rs
```diff
@@ -482,118 +481,140 @@ fn ipv6_check_receive_packet(
     unsafe {
```
gendx, line 481: Is this `unsafe` still necessary now that the `ptr::read` have been removed?

I also want to make an observation that is not really related to this pull-request. In general I don't think having such a broad `unsafe` block is a good pattern:
- Either the `unsafe` part is located in some specific line of code (`ptr::read`), and should only cover this line, to avoid allowing unforeseen unsafe operations elsewhere in the block of code by mistake.
- Or the whole function is `unsafe`, pushing the callers to prove it's indeed safe to call it.
- In either case, there should be some comment or documentation explaining why this `unsafe` block exists.

But wrapping the whole body of a function in an `unsafe` block without marking the function as `unsafe` seems sneaky and IMO is the recipe for ending up with undefined behavior.

hudson-ayers: The unsafe is still needed because the function relies on a bunch of accesses of global static mut variables (it could be scoped down a little bit now but not much). 

I completely agree with your comments regarding when it is appropriate to use unsafe, and agree that this test does not set a good example there. But I am a little reluctant to refactor this whole test when it is just a test for a single board that is not used elsewhere and will not be included in normal use of Tock. Either way such a refactoring is outside the scope of this PR.

kernel/src/process.rs
hudson-ayers: To be honest I am not really familiar enough to know how to fix the `sp` and tbf header pointer casts to be guaranteed to be aligned (or whether those are in fact UB).

But if pat is confident that the allow added here is acceptable, perhaps we could add a comment indicating the grant pointer casts are safe, and then create an issue indicating the `sp` and tbf header pointer casts may not be?

ppannuto: I added comments for the grants and stack pointer, as those are okay.

The `create` one still needs thinking about / tending to.

kernel/src/process.rs
```diff
@@ -1834,6 +1844,12 @@ impl<C: 'static + Chip> Process<'a, C> {
     }
 
     /// Get the current stack pointer as a pointer.
+    // This is currently safe as the the userspace/kernel boundary
+    // implementations of both Risc-V and ARM would fault on context switch if
+    // the stack pointer were misaligned.
+    //
+    // This is a bit of an undocumented assumption, but I'm not sure I forsee
+    // an architecture in the near future where this is realistically a risk.
```
gendx, line 1852: Nit: without context, it's unclear who is "I" in this comment.

hudson-ayers:
I moved the `create()` allow from the body of the function to annotate the two specific pointer casts that trigger it, and attempted to document why I believe those casts are guaranteed to be aligned. I would appreciate someone checking my work though.

chips/nrf52/src/ficr.rs
jmichelp: The `Unspecified` values are only used when converting between `u32` and the `Enum`. I think it's just old-style code and this could be replaced by implementing the `TryFrom` and `TryInto` traits IMO.

jmichelp: But they actually are mentioned in the product specification BTW :)

hudson-ayers: Given that the datasheet does list the `Unspecified` variants, it seems a little weird to remove them from the `register_bitfields()` call. But I am happy to make that change if people agree that is the preferred approach.

ppannuto: For low-level stuff like this, I generally prefer matching what the hardware actually does (not that consumers are ever likely to have hands-on of a test chip set to this value, but nonetheless I don't think it's a value we are at liberty to specify).

---

Partially related, from the clippy source:
```rust
    /// **What it does:** Checks for C-like enumerations that are
    /// `repr(isize/usize)` and have values that don't fit into an `i32`.
    ///
    /// **Why is this bad?** This will truncate the variant value on 32 bit
    /// architectures, but works fine on 64 bit.
```

I don't understand why a **usize** or **u32** would be truncated.

hudson-ayers: Yeah, I am continuing to dig into this and it seems incorrect that we labeled these with `#[allow(overflowing_literals)]` in the first place. The issue is that Rust is parsing these values as isize`, not `u32`.

Removing the overflowing_literalls allow leads to the following warning:
```
error: literal out of range for isize
   --> chips/nrf52/src/ficr.rs:140:27
    |
140 |             Unspecified = 0xffffffff
    |                           ^^^^^^^^^^
    |
    = note: `#[deny(overflowing_literals)]` on by default
    = note: the literal `0xffffffff` (decimal `4294967295`) does not fit into an `isize` and will become `-1isize`

error: could not compile `nrf52`.
```

Is this an error in the tock-register-interface crate?

hudson-ayers: Okay, I have convinced myself this is not an error. I extracted the relevant portions of the tock-register-interface macros as the following rust code:

```rust
pub trait TryFromValue<V> {
    type EnumType;

    fn try_from(v: V) -> Option<Self::EnumType>;
}

#[repr(i32)] // This is not included in the macro, instead the enum defaults to the default isize representation, which on our devices is i32.
#[derive(Debug)]
pub enum Value {
    First = 0,
    #[allow(overflowing_literals)]
    Second = 0xffffffff,
}

impl TryFromValue<u32> for Value {
    type EnumType = Value;

    fn try_from(v: u32) -> Option<Self::EnumType> {
        match v {
            x if x == Value::First as u32 => Some(Value::First),
            x if x == Value::Second as u32 => Some(Value::Second),
            _ => Option::None,
        }
    }
}

fn main() {
    let tmp = Value::try_from(0xffffffff);
    println!("tmp: {:?}", tmp);
    // works! tmp == Some(Second)
}
```

Without the overflowing literals allow, this wont compile, but with it, the `try_from` call is interpreted correctly. 

The basic problem is that the enum defined here: https://github.com/tock/tock/blob/master/libraries/tock-register-interface/src/macros.rs#L90-L113 uses the default representation (`isize`). This specific problem can be fixed by annotating the enum with `repr($valtype)`, but then the macro fails when passed zero-variant enums. It is probably possible to fix this problem within the macro so that the alternative representation is only used when the enum has 1+ variants, but given that regardless of the enum representation the registers will be parsed correctly, it seems more correct to annotate the enum *in the register_bitmasks macro* with `allow[overflowing_literals]` and `#[allow(clippy_enum_clike_unportable_variant)]`, so that users of the interface don't have to worry about this non-issue.

hudson-ayers: actually, that is also not perfect, because it allows passing a value greater than the maximum usize...

hudson-ayers:
Pushed a commit. It is a change to the `register_bitmasks!()` macro that I think resolves the requirement for `#[allow(overflowing_literals)]` and `#[allow(clippy::enum_clike_unportable_variant)]` when registering bitmasks larger than `isize::MAX`. Basically, the enum should be `[repr(usize)]`, except when it is a zero variant enum, in which case it must be the default representation (hence the need for the extra match arm almost entirely duplicating the previous one, because there is no clean way to do if/else in rust macros except for by adding match arms or seperating out into smaller submacros).

If this change should be a separate PR so it can be more broadly discussed let me know.

ppannuto:
Wow, that was a bit of a rabbit hole -- nice work tracking all that down!

I really feel like this is a Rust shortcoming, indeed there is a somewhat closely related issue posted recently for enums mis-sizing themselves when all of the variants are empty https://github.com/rust-lang/rust/issues/68065 -- I don't think we'll ever hit that case as our variants are all still meaningfully typed, but it's a bit of a land mine lying around nonetheless.

---

~Given that's just a big copy/paste of the existing macro I'm okay with including it here~ Actually, I changed my mind mid-sentence. I think it'd be nice to have that in its own issue so that we can point to this as a discrete oddity if/when the enum representation of registers becomes an issue in the future. As it's just touching those two files, hopefully it's easy to pull out.

libraries/tock-register-interface/src/macros.rs
```diff
@@ -51,13 +51,79 @@ macro_rules! register_bitmasks {
...
+            #[allow(unused_imports)]
+            use $crate::registers::{FieldValue, TryFromValue};
+            use super::$reg_desc;
+
+            $(
+            #[allow(non_upper_case_globals)]
+            #[allow(unused)]
+            $(#[$inner])*
+            pub const $valname: FieldValue<$valtype, $reg_desc> =
+                FieldValue::<$valtype, $reg_desc>::new((1<<($numbits-1))+((1<<($numbits-1))-1),
```
gendx, line 74: This seems like an obscure arithmetic expression which would IMO deserve its own macro to clarify what it does, given that it's used in many places now.

gendx: Also, alternative proposal that may be more readable: `!0 >> (core::mem::size_of::<$valtype>() * 8 - $numbits)`

hudson-ayers:
I reverted the last commit on this PR, and moved it into #1823 so it can be reviewed separately.

chips/nrf52/src/ficr.rs
```diff
@@ -144,6 +146,7 @@ register_bitfields! [u32,
         /// Part Variant, Hardware version and Production configuration, encoded as ASCII
         // Note, some of these are not present in datasheets
         // but are in nrf52.svd or are observed in the wild
+        #[allow(clippy::enum_clike_unportable_variant)]
```
ppannuto, line 149: I think we should simply remove all of these from this PR and resolve this in #1823. I know that means this PR doesn't _quite_ get clippy passing, but it's also been through a lot of churn at this point, so it'd be nice to land all of the uncontroversial stuff. I think without these, it's ready to go.

hudson-ayers: Yeah of course, sorry I meant to include that change originally

hudson-ayers:
bors retry

PR #1625, opened 2020-02-25
CI: build OS X on master and bors merges to it

ppannuto:
### Pull Request Overview

This pull request attempts to implement the idea suggested in #1568. 

### Testing Strategy

This pull request was tested by... well, you're looking at it. I've successfully demonstrated with this branch that an OS X build is correctly not yet triggered, but linux builds are. Then the question will be if/when we use bors to merge this whether an OS X build runs first.

### TODO or Help Wanted

If someone's more a travis guru than me and sees a flaw, holler, but I think this is right.

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

PR #1801, opened 2020-04-26
RISC-V: Add automated QEMU testing

alistair23:
### Pull Request Overview

This starts the process of automating QEMU testing for certain targets.

Currently we just do very basic grep based testing. The idea is to move this to a more capable test-runner infrastructure. We want to make sure we utilise the `make qemu` target for each board to be as close as possible to user testing.

### Testing Strategy

Push to GitHub and have Travis run tests.

### TODO or Help Wanted

Nothing TODO for this PR.

Once OT support is merged to mainline QEMU ([patches are on list](https://patchwork.kernel.org/project/qemu-devel/list/?series=277631)) we can add OpenTitan testing to this as well.

Future TODO is to use a test-runner and to have more complex test cases.

### Documentation Updated

### Formatting

- [X] Ran `make formatall`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,6 @@ doc/rustdoc
 # Cargo.lock files have a lot of noisy churn during development. Instead they
 # are manually included for releases.
 Cargo.lock
+
+# Used to build QEMU for emulation testing
+tools/qemu
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,10 @@ branches:
     - staging.tmp
     - trying.tmp
 
-
 cache:
   cargo: true
+  directories:
+    - tools/qemu
 
 os:
   - linux
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,8 @@ ci-travis:\
 	ci-syntax\
 	ci-compilation\
 	ci-debug-support-targets\
-	ci-documentation
+	ci-documentation \
+	emulation-check
 	@printf "$$(tput bold)********************$$(tput sgr0)\n"
 	@printf "$$(tput bold)* CI-Travis: Done! *$$(tput sgr0)\n"
 	@printf "$$(tput bold)********************$$(tput sgr0)\n"
--- a/Makefile
+++ b/Makefile
@@ -15,15 +15,17 @@ usage:
 	@echo "Check out the README in your board's folder for more information."
 	@echo
 	@echo "This root Makefile has a few useful targets as well:"
-	@echo "  allboards: Compiles Tock for all supported boards"
-	@echo "   allcheck: Checks, but does not compile, Tock for all supported boards"
-	@echo "     alldoc: Builds Tock documentation for all boards"
-	@echo "      audit: Audit Cargo dependencies for all kernel sources"
-	@echo "         ci: Run all continuous integration tests"
-	@echo "      clean: Clean all builds"
-	@echo "     format: Runs the rustfmt tool on all kernel sources"
-	@echo "  formatall: Runs all formatting tools"
-	@echo "       list: Lists available boards"
+	@echo "        allboards: Compiles Tock for all supported boards"
+	@echo "        allcheck: Checks, but does not compile, Tock for all supported boards"
+	@echo "          alldoc: Builds Tock documentation for all boards"
+	@echo "           audit: Audit Cargo dependencies for all kernel sources"
+	@echo "              ci: Run all continuous integration tests"
+	@echo " emulation-setup: Setup QEMU for the emulation tests"
+	@echo " emulation-check: Run the emulation tests for supported boards"
+	@echo "           clean: Clean all builds"
+	@echo "          format: Runs the rustfmt tool on all kernel sources"
+	@echo "       formatall: Runs all formatting tools"
+	@echo "            list: Lists available boards"
 	@echo
 	@echo "$$(tput bold)Happy Hacking!$$(tput sgr0)"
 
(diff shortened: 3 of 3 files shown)


torfmaster:
Nice! Would it make sense to run the `libtock-rs` test suite to check the kernel?

alistair23:
A test suite would be nice, but I'm not sure it's the right answer. How do we build it? What happens if there is a bug in `libtock-rs`? What happens if we change the kernel ABI (Tock 2.0)?

torfmaster:
I would (inside .travis.yml for example) choose a fixed version of `libtock-rs`which could be bumped from time to time. If the ABI changes then there will (hopefully) be a version of `libtock-rs` which is compatible with that which could be chosen.

bradjc:
Is the qemu folder cached?

alistair23:
I don't think so, I can see if I can.

alistair23:
It should be cached now

.travis.yml
```diff
@@ -43,6 +44,7 @@ before_script:
 script:
   - export PATH=$HOME/.cargo/bin:$PATH
   - make ci-travis
+  - make emulation-check
```
ppannuto, line 47: Is there a reason to put this here rather than inside of the Makefile? I don't like that this now means `make ci-travis` != run locally what runs on travis.

Given the aspiration of multi-platform travis, the make rule might need to be something like:

```diff
.PHONY: ci-travis
ci-travis:\
	ci-formatting\
	ci-tools\
	ci-libraries\
	ci-archs\
	ci-kernel\
	ci-chips\
	ci-syntax\
	ci-compilation\
	ci-debug-support-targets\
	ci-documentation
+        @if [[ $TRAVIS_OS_NAME == "linux" ]]; $(MAKE) emulation-check; else; printf "$$(tput bold)Skipping QEMU, unsupported platform(tput sgr0)\n"; fi
	@printf "$$(tput bold)********************$$(tput sgr0)\n"
	@printf "$$(tput bold)* CI-Travis: Done! *$$(tput sgr0)\n"
	@printf "$$(tput bold)********************$$(tput sgr0)\n"
```

alistair23: The reason was that maybe QEMU builds are too slow for some people and if they are just making ARM related changes they don't need to run RISC-V tests locally.

Are we targeting a platform where QEMU won't build?

ppannuto: I think I would rather preserve the idiom that `make ci-travis` is exactly what travis runs.

Putting QEMU last hopefully mitigates any build time concerns -- in the hopefully-not-too-distant github actions future, that won't be an issue anyway.

---

We have a goal of running travis on OS X on master-only https://github.com/tock/tock/pull/1625; though again, maybe with github actions, that looks different as well. 

> OH, QEMU works on Mac, interesting

Umm.. I had it in my head that it didn't work on OS X; let me try getting this running locally once

Makefile
```diff
@@ -183,6 +185,22 @@ ci-rustdoc:
 audit:
 	@for f in `./tools/list_lock.sh`; do echo "$$(tput bold)Auditing $$f"; (cd "$$f" && cargo audit || exit 1); done
 
+.PHONY: emulation-setup
+emulation-setup:
+	@#Use the latest QEMU as it has OpenTitan support
+	@if [[ ! -d qemu || ! -f qemu/VERSION ]]; then \
+		rm -rf qemu; \
+		git clone https://github.com/qemu/qemu.git; \
```
ppannuto, line 193: probably better to clone this into `tools/qemu` rather than adding stuff to top-level directory; we install a bunch of other stuff into `tools` already

.travis.yml
alistair23: Ok, I have moved it to the ci Make target.

QEMU should build and run on OS X. I don't think we would need any changes to the scripts to have it work.

.gitignore
```diff
@@ -52,3 +52,6 @@ doc/rustdoc
 # Cargo.lock files have a lot of noisy churn during development. Instead they
 # are manually included for releases.
 Cargo.lock
+
+# Used to build QEMU for emulation testing
+qemu
```
ppannuto, line 57: Need to move to `tools/.gitignore` now too

Makefile
```diff
@@ -183,6 +186,22 @@ ci-rustdoc:
...
+		rm -rf tools/qemu; \
+		cd tools; git clone https://github.com/qemu/qemu.git; \
+		cd qemu; ./configure --target-list=riscv32-softmmu; \
+	fi
+	@$(MAKE) -C "tools/qemu" > /dev/null
+
+.PHONY: emulation-check
+emulation-check: emulation-setup
+	@$(MAKE) -C "boards/hifive1"
+	@PATH="${PATH}:../../tools/qemu/riscv32-softmmu/" timeout --foreground 5s $(MAKE) qemu -C "boards/hifive1" | tee log > /dev/null
```
ppannuto, line 202: Can we somehow make this more self-contained so that it doesn't just drop a `log` file, maybe follow the technique described here: https://unix.stackexchange.com/a/181938/12313 ?

alistair23:
Ok, back to passing. I think I have addressed all comments.

bradjc:
Do you think I can volunteer you @hudson-ayers and @lschuermann to convert this to github actions? All signs point to that being strictly better than travis...maybe we should just freeze and eventually deprecate our travis setup, and move all new CI development over.

bradjc:
But we can also use this now, and switch at some later point.

alistair23:
I'm happy to help with the swap as well

hudson-ayers:
Sure, here is a test: https://github.com/Tock-CI/tock/pull/10 . 

It seems to work pretty much as-is, with the caveat that both the travis build log and the github actions build log show `/bin/sh: 1: [[: not found
/bin/sh: 1: -f: not found` as the first two lines of the emulation steps

hudson-ayers:
The fix for that is to change emulation-setup to use bash instead of sh (which will use dash on ubuntu):
```bash
.PHONY: emulation-setup
emulation-setup: SHELL:=/usr/bin/env bash
emulation-setup:
	@#Use the latest QEMU as it has OpenTitan support
	@if [[ ! -d tools/qemu || ! -f tools/qemu/VERSION ]]; then \
# ...
```

alistair23:
I didn't notice that warning. I have pushed a fix.

PR #1802, opened 2020-04-27
Updated SPI hold_low and release_low for STM

alexandruradovici:
### Pull Request Overview

This pull request:
- ~~Adds the `hold_low` and `release_low` functions to the SpiMasterDevice trait, so that the VirtualSpiMasterDevice may expose them. They exist in the SpiMaster, but were missing from the SpiMasterDevice. Some devices may require this.~~
- Fixes the implementation of these functions for stm32f3xx and stm32f4xx chips. Previously these functions where just activating and deactivating the chip select instead of keeping the chip select line low.

### Testing Strategy

This pull request was tested using the nucleof429zi board.

### TODO or Help Wanted

Any feedback is welcome.

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.

### DIFF
--- a/capsules/src/virtual_spi.rs
+++ b/capsules/src/virtual_spi.rs
@@ -189,6 +189,14 @@ impl<Spi: hil::spi::SpiMaster> hil::spi::SpiMasterDevice for VirtualSpiMasterDev
     fn get_rate(&self) -> u32 {
         self.mux.spi.get_rate()
     }
+
+    fn hold_low (&self) {
+        self.mux.spi.hold_low();
+    }
+
+    fn release_low (&self) {
+        self.mux.spi.release_low();
+    }
 }
 
 pub struct VirtualSpiSlaveDevice<'a, Spi: hil::spi::SpiSlave> {
--- a/capsules/src/virtual_spi.rs
+++ b/capsules/src/virtual_spi.rs
@@ -189,14 +189,6 @@ impl<Spi: hil::spi::SpiMaster> hil::spi::SpiMasterDevice for VirtualSpiMasterDev
     fn get_rate(&self) -> u32 {
         self.mux.spi.get_rate()
     }
-
-    fn hold_low(&self) {
-        self.mux.spi.hold_low();
-    }
-
-    fn release_low(&self) {
-        self.mux.spi.release_low();
-    }
 }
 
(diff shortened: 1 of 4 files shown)


alexandruradovici:
> I'm not sure this is the right change for virtual_spi. At minimum configuring the underlying spi has to go through the `self.operation` queue somehow, otherwise this could be changing the CS characteristics for some other SPI client.
> 
I see your point , I totally overlooked the fact that spi requests are interleaved by the mux. I removed the `hold_low` and `release_low` from the trait.

> More generally, however, I'm not sure the virtual spi device can handle multiple back-to-back transfers, since the mux would have to know not to handle another client once the first transfer has finished. But if the mux waits, and a second request doesn't happen, then the mux will continue to block on that client.
> 

Maybe a transactional approach would be good, the virtual spi should request the next buffer using a `read_write_continue` function? If the client replies with None, the request is finished and another client is served. I'll look into this.

> I think this is why we didn't include those functions in the device interface, but I'm not entirely sure. A driver that needs that functionality can use a non-virtualized SPI, otherwise I think the virtual_spi will have to get more sophisticated.

alexandruradovici:
The bug in the STM chips still remains, I left only these modifications.

alexandruradovici:
> > I'm not sure this is the right change for virtual_spi. At minimum configuring the underlying spi has to go through the `self.operation` queue somehow, otherwise this could be changing the CS characteristics for some other SPI client.
> 
> I see your point , I totally overlooked the fact that spi requests are interleaved by the mux. I removed the `hold_low` and `release_low` from the trait.
> 
> > More generally, however, I'm not sure the virtual spi device can handle multiple back-to-back transfers, since the mux would have to know not to handle another client once the first transfer has finished. But if the mux waits, and a second request doesn't happen, then the mux will continue to block on that client.
> 
> Maybe a transactional approach would be good, the virtual spi should request the next buffer using a `read_write_continue` function? If the client replies with None, the request is finished and another client is served. I'll look into this, I think this is enough for some devices that have the data ahead, they just need to send it divided in packets. 
> 
> > I think this is why we didn't include those functions in the device interface, but I'm not entirely sure. A driver that needs that functionality can use a non-virtualized SPI, otherwise I think the virtual_spi will have to get more sophisticated.

PR #1803, opened 2020-04-27
refactor stmf32f4xx chip code into multiple crates

hudson-ayers:
### Pull Request Overview

This pull request refactors the `stm32f4xx` code into three crates. In addition to the base crate, it adds two new crates -- `stm32f446re` and `stm32429zi` -- which contain all code specific to those chip versions. So far, that is just a few NVIC IDs and the `IRQS` array.

Importantly, this PR removes any need for cargo features (`#[cfg( feature = x)]` statements) for functionality which differs between these two chips. It also absolves any need for `cfg_if!()` or other similar approach to enable clippy to pass on these crates, so I am closing #1794 in favor of this.

### Testing Strategy

This pull request was tested by travis-ci and clippy.
I do not have any nucleo boards, so I can't test this on hardware.

### TODO or Help Wanted

My approach to separating out the NVIC ids was to leave the shared IDs in `stm32f4xx/src/nvic.rs` and put chip-specific IDs in `stm32f446re/src/stm32f446re_nvic.rs and stm32f429zi/src/stm32f429zi_nvic.rs`. We don't currently use any of the chip-specific IDs, so the usability of this approach is pretty untested -- I think we would have to turn to something similar to the `interrupt_service` trait used by the `nrf52` family of chips.

### Documentation Updated

- [x] I updated all doc tests that used `stm32f4xx`, and didn't see any other documentation that needs updating, but could have missed something.

### Formatting

- [x] Ran `make formatall`.


Cargo.toml
```diff
@@ -32,6 +32,8 @@ members = [
     "chips/sifive",
     "chips/stm32f3xx",
     "chips/stm32f4xx",
+    "chips/stm32f446re",
+    "chips/stm32f429zi",
```
gendx, line 36: Nit: maybe we can keep this list sorted?

chips/stm32f4xx/src/nvic.rs
```diff
@@ -83,36 +83,6 @@ pub const OTG_HS_EP1_IN: u32 = 75;
 pub const OTG_HS_WKUP: u32 = 76;
```
gendx, line 83: I can't comment directly on unchanged lines but lines 66-69 should be removed.

boards/nucleo_f429zi/src/main.rs
```diff
@@ -258,132 +258,132 @@ pub unsafe fn reset_handler() {
...
+                                                        // Enable the to use the ADC pins as GPIO
+                                                        // stm32f429zi::gpio::PIN[0][3].as_ref().unwrap(), //A0
+                                                        // stm32f429zi::gpio::PIN[2][0].as_ref().unwrap(), //A1
+                                                        // stm32f429zi::gpio::PIN[2][3].as_ref().unwrap(), //A2
+                                                        // stm32f429zi::gpio::PIN[5][3].as_ref().unwrap(), //A3
+                                                        // stm32f429zi::gpio::PIN[5][5].as_ref().unwrap(), //A4
+                                                        // stm32f429zi::gpio::PIN[5][10].as_ref().unwrap(), //A5
+                                                        // stm32f429zi::gpio::PIN[1][1].as_ref().unwrap(), //A6
+                                                        // stm32f429zi::gpio::PIN[2][2].as_ref().unwrap(), //A7
+                                                        // stm32f429zi::gpio::PIN[5][4].as_ref().unwrap()  //A8
```
gendx, line 386: I'm wondering what's happening with the formatting here. Maybe a block comment would be better than line-by-line comments?

hudson-ayers: Rustfmt seems to handle this case really poorly, even with block comments. I moved the commented out pins to the top of the argument list instead.

gendx:
Can you remove the features from `chips/stm32f4xx/Cargo.toml` as well?

hudson-ayers:
bors r+

Given that no one was able to do 1.5 release testing for the nucleo boards, I don't think it makes sense to wait for someone with hardware to test on this board.

PR #1805, opened 2020-04-28
rv32i: pmp: Remove power of two allignment constraint

alistair23:
### Pull Request Overview

The TOR PMP implementation (the one we are using) does not require a
power of two size allignment so let's remove it to save space.

### Testing Strategy

Running OpenTitan in QEMU with PMP enabled.

### Documentation Updated

### Formatting

- [X] Ran `make formatall`.

### DIFF
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -5,7 +5,6 @@ use core::fmt;
 
 use crate::csr;
 use kernel;
-use kernel::common::math;
 use kernel::common::registers::register_bitfields;
 use kernel::mpu;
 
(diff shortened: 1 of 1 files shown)


alistair23:
Are you worried we would have a size less then the pointer size, or not aligned to the pointer size?

The TOR PMP still needs to be 4byte aligned as we shift (>> 2) when writing to a register so we should be ok.

ppannuto:
I'm both not actually worried and worried that it won't be aligned to pointer size. Several other parts of the kernel assume that region boundaries are aligned to at least pointer width (see the recent series clippy / cast stuff). On the one hand, I really don't see it as likely for an MPU region to have granularity finer than pointer width; on the other hand, we c/would start having alignment issues if anything ever did that currently.

Issue #1730, opened 2020-04-02
Inefficiencies in the system call path on RISC-V


alphan:
I recently took a close look at Tock OS system call performance on OpenTitan using Verilator and noticed some inefficiencies in the system call path.

To measure the system call performance, I implemented a simple capsule with a `command` system call that returns the value of the `mcycle` register and triggered this system call from a libtock-rs application. Both tock and libtock-rs were built with `opt-level = 3` and `lto = true`.

According to the verilator execution trace, roundtrip latency of this system call is 5172 cycles.

Looking at the execution trace and disassembly of the kernel side by side I was able to find a couple of places with room for improvement. I share them below because they seem to be relatively safe:
- In `Process::swtich_to()` (line 1124) using `self.stored_state.take()` instead of `self.stored_state.get()` saves 1225 cycles.
- Skipping MPU configuration if returning to the same process saves an extra 2751 cycles.
- Commenting out debug_* calls in `Process::switch_to()` and `Kernel::do_process()` saves an extra 102 cycles.

I hope I didn't miss anything obvious and please let me know if I can provide any additional information.

bcantrill:
Good stuff! @alphan, if you haven't seen it, you might be interested in [Tockilator](https://github.com/oxidecomputer/tockilator), a (very) crude tool we built to facilitate this kind of analysis...

alphan:
Thanks @bcantrill! I wish I had seen it before, looks pretty cool! I'll definitely check it out.

gendx:
> * Commenting out debug_* calls in `Process::switch_to()` and `Kernel::do_process()` saves an extra 102 cycles.

I've been thinking about having a more principled way of removing debugging statements, see #1372 if you find it useful.

And I've been tracking code-size improvements in #1663. This issue could be generalized to all kinds of performance improvements, or someone could add another tracking issue for saving CPU cycles.

gkelly:
I can't assign this to @tock/opentitan-wg for some reason, @bradjc is assigning this to you-in-lieu as the chair of that WG a reasonable alternative?

vsukhoml:
i think there is potential to reduce syscall overhead down  to 50-100 cycles by streamlining storing/restoring context and reducing it to RISC-V ABI conventions on syscalls (don't save caller-saved regs).

hudson-ayers:
cc #2101 which identifies inefficiencies (~130 cycles) on `allow()` calls and on any system calls that access grant regions.

PR #1613, opened 2020-02-20
kernel: don't cache the process appid in main loop

bradjc:
As we get closer to finishing #1082, I've been testing restarting various apps and found that caching the appid in the main loop can lead to errors since the appid can change if a process restarts. The issue happens when:

1. an app is executed and crashes during its execution
1. the main loop handles this and restarts it
1. its appid has now changed
1. the scheduler loop re-runs the process
1. the process calls a syscall, but when the kernel loop handles that syscall it uses the process's old cached appid

I think this is fine to merge ahead of the other #1082 changes, but at least it is easier to review separated out in this PR.

### Testing Strategy

Using the crash_dummy app to verify that the app restarts correctly.

### TODO or Help Wanted

n/a

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

PR #1566, opened 2020-01-27
kernel: Add an "identifier" to AppId

bradjc:
### Pull Request Overview

Currently a capsule can store an `AppId` to refer to a process, and there is not really a mechanism for ensuring that that `AppId` is still valid. Currently in Tock we are OK with this since we only create processes once, they don't restart, and we don't move them around in the processes array. However, as process management becomes more sophisticated, the idea that an AppId is always valid will no longer be true. Since `AppId` only stores the index of the relevant process in that array, if a process restarts, or if a process is removed and then a new one is added in the same slot, then the stored `AppId` will appear valid, even though it actually refers to a different process.

To enable verifying that `AppId`s are still valid, this PR adds a new field (`identifier`) to the `AppId` object. This allows `AppId`s to be used as before (i.e. they are still copyable), but now operations using `AppId` can verify that the identifier in the AppId matches the identifier assigned to the process at a particular index.

AppId retains the `index` field which does correspond to the index in the processes array (at least while the AppId is valid). This allows us to retain direct lookups, as well as simplifies the IPC implementation.

#### Changes to AppId API

The existing `.idx()` function has been removed, and `.index()` and `.id()` have been added. `.index()` checks that the AppId is still valid before returning the index in the array of the corresponding app. `.id()` just returns the identifier, which can be passed to userspace or over a UART channel, for example.

These changes resulted in a couple changes to capsules which need `usize` identifiers for apps, and now must call `.id()`.

#### Changes to kernel APIs

The major change in the kernel (beyond AppId) is that functions which used to take an index now just take an `AppId`. Mostly this is a minor change, except for `kernel.process_map_or()`. This function runs a closure on a specific app if it is exists, and uses the index and identifier to verify the app is still valid.

Otherwise, IPC had to be updated quite a bit to handle the changes. I also took the opportunity to update how grant iters are implemented since this PR reduces the reliance on an app's identifier being its index in the process array.

#### How does this fix the capsule problem?

A capsule can still store an AppId, and that AppId can still become invalid. However, the main use of an AppId between a capsule and the kernel is to call `grant.enter(appid)`. Without this PR, if appid refers to an old process, and a new process has started in the same array index, `grant.enter(appid)` would succeed (after initializing a new grant region for the new process). Now, when `grant.enter()` calls `.kernel.process_map_or(appid)`, an error will be returned which will bubble back to the capsule.

#### What does this PR not do?

[...]

PR #1822, opened 2020-05-04
kernel: configure MPU only if switching to a different process

alphan:
### Pull Request Overview

Currently, the scheduler configures the MPU every time when it switches
to a process. This change aims to improve syscall latency by caching the
last running process and configuring the MPU only when switching to a
different process.

See also: #1730.

### Testing Strategy

I ran a simple `libtock-rs` application for measuring syscall latency with
and without this change on opentitan on verilator. This change reduces the
latency of a command syscall by ~~2751~~ 1711 cycles (the exact number of cycles
saved obviously depends on the actual implementation, e.g. would probably
be less after #1821).

Also ran `make allcheck`.

### TODO or Help Wanted

I couldn't find any doc files or tests that should be updated. Since
this is my first real Tock PR, it would be great if you could point me
in the right direction if I missed anything.

### Documentation Updated

### Formatting

- [X] Ran `make formatall`.

### DIFF
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -178,8 +178,14 @@ impl PMPConfig {
         }
         None
     }
+}
 
-    fn clear_mpu_registers(&self) {
+impl kernel::mpu::MPU for PMPConfig {
+    type MpuConfig = PMPConfig;
+
+    fn enable_mpu(&self) {}
+
+    fn disable_mpu(&self) {
         for x in 0..self.total_regions {
             // If PMP is supported by the core then all 16 register sets must exist
             // They don't all have to do anything, but let's zero them all just in case.
@@ -355,17 +361,9 @@ impl PMPConfig {
         csr::CSR.pmpcfg0.modify(csr::pmpconfig::pmpcfg::w0::SET);
         csr::CSR.pmpcfg0.modify(csr::pmpconfig::pmpcfg::x0::SET);
         csr::CSR.pmpcfg0.modify(csr::pmpconfig::pmpcfg::a0::TOR);
-        // Set the dirty flag
-        self.is_dirty.set(true);
+        // MPU is not configured for any process now
+        self.last_configured_for.take();
     }
-}
-
-impl kernel::mpu::MPU for PMPConfig {
-    type MpuConfig = PMPConfig;
-
-    fn enable_mpu(&self) {}
-
-    fn disable_mpu(&self) {}
 
     fn number_total_regions(&self) -> usize {
         self.total_regions
@@ -505,7 +503,6 @@ impl kernel::mpu::MPU for PMPConfig {
         // Skip PMP configuration if it is already configured for this app and the MPU
         // configuration of this app has not changed.
         if !last_configured_for_this_app || config.is_dirty.get() {
-            // tock/tock#1860
             let mut regions_sorted = config.regions.clone();
             regions_sorted.sort_unstable_by(|a, b| {
                 let (a_start, _a_size) = match a.location() {
@@ -518,9 +515,6 @@ impl kernel::mpu::MPU for PMPConfig {
                 };
                 a_start.cmp(&b_start)
             });
-
-            // tock/tock#1860
-            self.clear_mpu_registers();
 
             for x in 0..self.total_regions {
                 let region = regions_sorted[x];
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -105,7 +105,10 @@ impl PMPRegion {
 pub struct PMPConfig {
     regions: [PMPRegion; 8],
     total_regions: usize,
+    /// Indicates if the configuration has changed since the last time it was written to hardware.
     is_dirty: Cell<bool>,
+    /// The application that the MPU was last configured for. Used (along with the `is_dirty` flag)
+    /// to determine if MPU can skip writing the configuration to hardware.
     last_configured_for: MapCell<AppId>,
 }
 
@@ -160,6 +163,7 @@ impl PMPConfig {
             // As we use the PMP TOR setup we only support half the number
             // of regions as hardware supports
             total_regions: pmp_regions / 2,
+
             is_dirty: Cell::new(true),
             last_configured_for: MapCell::empty(),
         }
@@ -496,6 +500,7 @@ impl kernel::mpu::MPU for PMPConfig {
     }
 
     fn configure_mpu(&self, config: &Self::MpuConfig, app_id: &AppId) {
+        // Is the PMP already configured for this app?
         let last_configured_for_this_app = self
             .last_configured_for
             .map_or(false, |last_app_id| last_app_id == app_id);
@@ -503,6 +508,7 @@ impl kernel::mpu::MPU for PMPConfig {
         // Skip PMP configuration if it is already configured for this app and the MPU
         // configuration of this app has not changed.
         if !last_configured_for_this_app || config.is_dirty.get() {
+            // Sort the regions before configuring PMP in TOR mode.
             let mut regions_sorted = config.regions.clone();
             regions_sorted.sort_unstable_by(|a, b| {
                 let (a_start, _a_size) = match a.location() {
@@ -641,7 +647,6 @@ impl kernel::mpu::MPU for PMPConfig {
                     None => {}
                 };
             }
-            // Clear the dirty flag
             config.is_dirty.set(false);
             self.last_configured_for.put(*app_id);
         }
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -498,10 +498,13 @@ impl kernel::mpu::MPU for PMPConfig {
     }
 
     fn configure_mpu(&self, config: &Self::MpuConfig, app_id: &AppId) {
-        self.last_configured_for.map(|last_app_id| {
-            if last_app_id == app_id && config.is_dirty.get() == false {
-                return;
-            }
+        let last_configured_for_this_app = self
+            .last_configured_for
+            .map_or(false, |last_app_id| last_app_id == app_id);
+
+        // Skip PMP configuration if it is already configured for this app and the MPU
+        // configuration of this app has not changed.
+        if !last_configured_for_this_app || config.is_dirty.get() {
             // tock/tock#1860
             let mut regions_sorted = config.regions.clone();
             regions_sorted.sort_unstable_by(|a, b| {
@@ -646,7 +649,7 @@ impl kernel::mpu::MPU for PMPConfig {
             }
             // Clear the dirty flag
             config.is_dirty.set(false);
-            *last_app_id = *app_id;
-        });
+            self.last_configured_for.put(*app_id);
+        }
     }
 }
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -1,12 +1,15 @@
 //! Implementation of the physical memory protection unit (PMP).
 
+use core::cell::Cell;
 use core::cmp;
 use core::fmt;
 
 use crate::csr;
 use kernel;
+use kernel::common::cells::MapCell;
 use kernel::common::registers::register_bitfields;
 use kernel::mpu;
+use kernel::AppId;
 
 // This is the RISC-V PMP support for Tock
 // We use the PMP TOR alignment as there are alignment issues with NAPOT
@@ -99,10 +102,11 @@ impl PMPRegion {
 }
 
 /// Struct storing region configuration for RISCV PMP.
-#[derive(Copy, Clone)]
 pub struct PMPConfig {
     regions: [PMPRegion; 8],
     total_regions: usize,
+    is_dirty: Cell<bool>,
+    last_configured_for: MapCell<AppId>,
 }
 
 const APP_MEMORY_REGION_NUM: usize = 0;
@@ -122,6 +126,8 @@ impl Default for PMPConfig {
                 PMPRegion::empty(7),
             ],
             total_regions: 8,
+            is_dirty: Cell::new(true),
+            last_configured_for: MapCell::empty(),
         }
     }
 }
@@ -154,6 +160,8 @@ impl PMPConfig {
             // As we use the PMP TOR setup we only support half the number
             // of regions as hardware supports
             total_regions: pmp_regions / 2,
+            is_dirty: Cell::new(true),
+            last_configured_for: MapCell::empty(),
         }
     }
 
@@ -170,14 +178,8 @@ impl PMPConfig {
         }
         None
     }
-}
 
-impl kernel::mpu::MPU for PMPConfig {
-    type MpuConfig = PMPConfig;
-
-    fn enable_mpu(&self) {}
-
-    fn disable_mpu(&self) {
+    fn clear_mpu_registers(&self) {
         for x in 0..self.total_regions {
             // If PMP is supported by the core then all 16 register sets must exist
             // They don't all have to do anything, but let's zero them all just in case.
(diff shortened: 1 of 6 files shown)


kernel/src/sched.rs
```diff
@@ -331,6 +336,7 @@ impl Kernel {
         systick.reset();
         systick.set_timer(KERNEL_TICK_DURATION_US);
         systick.enable(false);
+        let identifier: usize = process.appid().id();
```
ppannuto, line 339:
```suggestion
        let identifier = process.appid();
```

Ah, I see where this is a bit non-intuitive, but the plain `appid` should be used in-kernel; the `.id()` method is only for special situations: https://docs.tockos.org/kernel/struct.appid#method.id

bradjc:
We would need to update the kernel/src/platform/mpu.rs documentation, particularly to clarify that the MPU implementation must be setup to only enforce in the userspace privilege level.

kernel/src/sched.rs
bradjc: This is subtle, but we should never make a copy of a process's ID like this, since it can change (see 4003b297c90f693ea297a3c70de35f28a0d49c7f for an example of where this has lead to a bug in the past). Instead just use `process.appid()` when needed.

kernel/src/sched.rs
```diff
@@ -47,6 +47,10 @@ pub struct Kernel {
     /// created and the data structures for grants have already been
     /// established.
     grants_finalized: Cell<bool>,
+
+    /// Identifier of the last running process. This is used to determine if
+    /// we are swtiching to a different process and need to configure the MPU.
+    last_running_process: Cell<Option<usize>>,
```
bradjc, line 53:
```suggestion
    last_running_process: OptionalCell<AppId>,
```

kernel/src/sched.rs
alphan: Thank you very much for reviewing this @ppannuto and @bradjc! Your comments raise two very important points:

**Caching process IDs should, in general, be avoided**
Thanks for pointing me to #1613 @bradjc! It is indeed a subtle bug caused by stale process IDs. However, this PR actually relies on the fact that the process IDs returned by `process.appid().id()` is unique among different processes and different _instances_ of the _same_ process [1]. This ensures that the MPU will be configured whenever we switch to a different process. I guess, another alternative could be to store this information as a flag in the process structure itself, but I find this approach more intuitive.

**`process.appid()` vs `process.appid().id()`**
@ppannuto, the reason for using `process.appid().id()` was performance. Using `process.appid()` introduces an overhead of 150 cycles.

[1]  Apologies for the abuse of terminology. Here, I am referring to processes that are restarted after crashes or updated at runtime.

alphan:
>  Did your testing include running multiple apps as well?

Yes, I also tried this with multiple apps and didn't notice any issues.

> Generally, I think this is a good change. However, I'm concerned there may be some extra considerations where we may need to identify the MPU configuration as "tainted". Specifically:
> 
> * if there are additional grant allocations, then the deny region may need to grow (though, this is likely not an immediate issue as the allowed region is generally minimal)
> * if there was a change in permitted region, i.e. the application called `brk` or an IPC `allow`* then these need to be recalculated
> * Maybe others I'm not thinking of?
> 
> Basically, I think this is a good common-case optimization, but I think we also need a way for something to "poison" the MPU configuration to force it to be recalculated.
> 
> *though that second case may actually be fine since in the current architecture as it's always a different process that is allowing a region, so we'll go through regular process switch

Thank you for catching this! I actually looked at various places that can change the MPU config but missed at least one case where MPU config can be modified and not applied immediately, e.g. `sbrk` and `brk` eventually call `configure_mpu()` but `alloc` doesn't. Since cache invalidation is tricky, does it make sense to modify `MPU` and `MPUConfig` to keep track of the _taintedness_ of `MPUConfig` at a higher level?

ppannuto:
>  Since cache invalidation is tricky, does it make sense to modify MPU and MPUConfig to keep track of the taintedness of MPUConfig at a higher level?

Yeah, I think that's a great idea. I believe the mpu configuration structure is already well-equipped for this as all of its state is private and can only be modified by getters and setters; hooking the latter to mark the MPU configuration as dirty, and then perhaps updating `configure_mpu` to know what process it's configured for, pushing the hot-path logic into the MPU itself seems like a good design to me.

In that case, I actually imagine we could _remove_ the `configure_mpu`s that are sprinkled around, instead architecting things such that whenever a process is returned to the MPU layer is consulted first, updating itself if necessary.

kernel/src/sched.rs
ppannuto: I'm afraid this does the opposite of what you might want. It is perfectly valid for the `.id()` value to be re-used. In fact, right now it is simply the index into the static array of `Process` pointers. If a process restarts, `.appid()` will change and `.appid().id()` will not.

However, if a process is restarted, its memory layout likely _has_ changed (i.e. no `brk` yet), thus the MPU regions should be reconfigured. I believe `.appid()` is still the correct choice, as it is what is unique among instances of a process.

alphan: Oh, I was under the impression that `index` can be reused but `.id()`, i.e. `identifier`, is unique because `create_process_identifier()` is called when a process is created or restarted.

ppannuto: Yes, sorry, you are right. I described the old behavior (id used to be the index; #1566 updated this).

Nonetheless, comparing AppId should do the right thing (it internally compares the identifiers returned by `.id()`):
```rust
impl PartialEq for AppId {
    fn eq(&self, other: &AppId) -> bool {
        self.identifier == other.identifier
    }
}
```

---

I know the _intent_ of the `AppId`/`identifier` split was to use `AppId`s internally and to provide `identifier`s for external use, but I'm not sure I could articulate when/where/why i'ts preferred to use `AppId` over the `identifer`

alphan: No worries, sounds good! I will update this PR to use `AppId` (we can hopefully avoid the overhead by using `MapCell` instead of `Cell`).

alphan:
- Reverted changes in `kernel/src/sched.rs`.
- Added `app_id` parameter to `configure_mpu`.
- Implemented the idea captured in this PR as a risc-v-only optimization in `arch/rv32i/src/pmp.rs`.
- Rebased on master.
- Also made `disable_mpu` a no-op for a more accurate latency measurement (moved its contents to `PMPConfig::clear_mpu_registers`, called from `configure_mpu` to stay close to current behavior). It looks like we don't need this function at all but I am deferring this change to #1860, which I created to keep track of the things that I noticed while working on this.

alphan:
@ppannuto, thank you for reviewing this change and your comments. I have a question regarding the changes in `arch/cortex-m3/src/mpu.rs`. Shouldn't `hardware_is_configured_for` be defined in `MPU` (to track the state of the hw)? Also, don't we need a `is_dirty` field in `CortexMConfig`?

ppannuto:
Hmm.., you're right we need some form of structure that is the global to the MPU hardware. The MPU configuration is stored _per process_ ([here](https://github.com/tock/tock/blob/master/kernel/src/process.rs#L789)), so doing dirty tracking in the `MpuConfig` type won't be sufficient for resetting the MPU when we change processes.

I don't think the `CortexMConfig` needs an explicit `dirty` flag. Instead, the cell indicates what application the hardware "is configured for"; whenever the MPU configuration is changed, that cell is cleared to indicate that the MPU is not configured correctly for any application, which acts similarly to a separate dirty signal.

alphan:
The `MPU` struct that I was referring to is the one defined [here](https://github.com/tock/tock/blob/master/arch/cortex-m3/src/mpu.rs#L124). IIUC, this should track the state of the hardware, i.e. `hardware_is_configured_for`, while `CortexMConfig` tracks if it is dirty or not.

alphan:
@ppannuto, 121b33f844b62930babbc2cd3d013606b14f28b6 LGTM, thanks! I also pushed a commit to fix the way `last_configured_for` is used in `configure_mpu`.

arch/rv32i/src/pmp.rs
```diff
@@ -352,8 +354,18 @@ impl kernel::mpu::MPU for PMPConfig {
...
+        self.is_dirty.set(true);
     }
+}
+
+impl kernel::mpu::MPU for PMPConfig {
+    type MpuConfig = PMPConfig;
+
+    fn enable_mpu(&self) {}
+
+    fn disable_mpu(&self) {}
```
alistair23, line 368: This came up in: https://github.com/tock/tock/pull/1871 and isn't the right approach as people might still want to disable PMP at boot.

alistair23: It also means we now don't match the trait documentation.

arch/rv32i/src/pmp.rs
```diff
@@ -478,148 +492,164 @@ impl kernel::mpu::MPU for PMPConfig {
...
+                };
+                let (b_start, _b_size) = match b.location() {
+                    Some((start, size)) => (start as usize, size),
+                    None => (0xFFFF_FFFF, 0xFFFF_FFFF),
+                };
+                a_start.cmp(&b_start)
+            });
+
+            // tock/tock#1860
+            self.clear_mpu_registers();
```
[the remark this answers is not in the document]
alphan: As I mentioned in #1860, I don't think we need to clear the PMP registers when returning to the kernel since we don't use the lock bit. Calling this function here just to stay as close as possible to the current implementation, was planning to remove it as part of #1860.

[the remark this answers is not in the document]
alistair23: This is confusing then. Can we leave the current `disable_mpu()` as is and work on fixing that? There is already a PR for this: https://github.com/tock/tock/pull/1873

arch/rv32i/src/pmp.rs
alphan: I think there is some room for improvement in the trait. The doc says that this will be called before returning to the kernel, so I think it is reasonable to implement it as NOP for risc-v similar to `enable_mpu`, but the name is too general so it amy be misleading.

alistair23: There is a pending PR to update the trait which will then stop calling disable on every transition.

alphan:
Reverted changes in `disable_mpu` to address @alistair23's comments. We should see the expected latency improvements when `do_process` stops clearing PMP registers every time it returns to kernel (#1873).

arch/rv32i/src/pmp.rs
```diff
@@ -478,148 +490,161 @@ impl kernel::mpu::MPU for PMPConfig {
...
-                            csr::CSR.pmpaddr7.set((start as u32 + size as u32) >> 2);
+    fn configure_mpu(&self, config: &Self::MpuConfig, app_id: &AppId) {
+        let last_configured_for_this_app = self
+            .last_configured_for
+            .map_or(false, |last_app_id| last_app_id == app_id);
+
+        // Skip PMP configuration if it is already configured for this app and the MPU
+        // configuration of this app has not changed.
+        if !last_configured_for_this_app || config.is_dirty.get() {
+            // tock/tock#1860
```
alistair23, line 506: Can you remove this?

alphan: Hmm, I thought the comment and the PR description were pretty clear. This skips PMP configuration if it is already configured for the given app and has not changed since the last time it was committed. Please see also #1730 and #1860.

alistair23: The comment does that?

alphan: Just to be clear, are you referring to just `tock/tock#1860` or the `if` statement? It is hard to tell looking at the way your comment is rendered on github (my reply above is for the latter but probably you meant the former). I added a reference to #1860 because the regions are sorted every time PMP is configured.

alistair23: Yeah, I was referring to the comment. The GitHub UI is really lacking unfortunately, so it can be confusing.

Can we remove the comment then? I don't think it adds anything and we have a tracking issue to track the inefficiently.

alphan: Ah, apologies for the confusion (I know github shows some context above the line that you comment but I was confused because this one is perfectly aligned). Done.

alistair23: Thanks! It's good to have someone else looking at the PMP code :)

arch/rv32i/src/pmp.rs
```diff
@@ -99,10 +102,11 @@ impl PMPRegion {
 }
 
 /// Struct storing region configuration for RISCV PMP.
-#[derive(Copy, Clone)]
 pub struct PMPConfig {
     regions: [PMPRegion; 8],
     total_regions: usize,
+    is_dirty: Cell<bool>,
+    last_configured_for: MapCell<AppId>,
```
silvestrst, line 109: Could be useful to document these fields.

arch/rv32i/src/pmp.rs
```diff
@@ -154,6 +160,8 @@ impl PMPConfig {
             // As we use the PMP TOR setup we only support half the number
             // of regions as hardware supports
             total_regions: pmp_regions / 2,
+            is_dirty: Cell::new(true),
+            last_configured_for: MapCell::empty(),
```
silvestrst, line 164: Could you maybe drop an empty line before `is_dirty`, otherwise it looks like the comment applies to the block of code and not just `is_dirty`.

arch/rv32i/src/pmp.rs
```diff
@@ -478,148 +492,161 @@ impl kernel::mpu::MPU for PMPConfig {
...
-                        _ => break,
                     }
-                }
-                None => {}
-            };
-        }
+                    None => {}
+                };
+            }
+            // Clear the dirty flag
```
silvestrst, line 644: Non informative comment, it is obvious from the code that the dirty flag is cleared.

arch/rv32i/src/pmp.rs
```diff
@@ -478,148 +492,161 @@ impl kernel::mpu::MPU for PMPConfig {
...
-                                    + csr::pmpconfig::pmpcfg::x3::CLEAR
-                                    + csr::pmpconfig::pmpcfg::a3::TOR,
-                            );
-                            csr::CSR.pmpaddr14.set((start as u32) >> 2);
-
-                            // Set access to end address
-                            csr::CSR.pmpcfg3.set(cfg_val << 24 | csr::CSR.pmpcfg0.get());
-                            csr::CSR.pmpaddr15.set((start as u32 + size as u32) >> 2);
+    fn configure_mpu(&self, config: &Self::MpuConfig, app_id: &AppId) {
+        self.last_configured_for.map(|last_app_id| {
```
silvestrst: I think it could be useful to document what this block does.

arch/rv32i/src/pmp.rs
```diff
@@ -478,148 +492,161 @@ impl kernel::mpu::MPU for PMPConfig {
...
+            if last_app_id == app_id && config.is_dirty.get() == false {
+                return;
+            }
+            // tock/tock#1860
+            let mut regions_sorted = config.regions.clone();
```
silvestrst, line 506: I think it could be useful to document what this block does.

arch/rv32i/src/pmp.rs
silvestrst: I think it might be useful to refer to the Github issue in the PR description, and instead add a descriptive comment here.

arch/rv32i/src/pmp.rs
alphan: Done.

arch/rv32i/src/pmp.rs
alphan: Done.

arch/rv32i/src/pmp.rs
alphan: Done.

arch/rv32i/src/pmp.rs
alphan: Latest version should be clearer but added a short comment nevertheless.

arch/rv32i/src/pmp.rs
alphan: This part was already there and will hopefully be removed, added a brief comment.

alphan:
Thank you for reviewing this @bradjc and @silvestrst!

Added comments to address @silvestrst's comments.

bradjc:
big queue i guess?

PR #1823, opened 2020-05-05
Fix Tock register interface to allow large unsigned values as bitmasks

hudson-ayers:
### Pull Request Overview

This pull request changes the representation of the `Value` enum in the `register_bitmasks!()` macro to be `#[repr(u64)]` instead of the default representation for Rust enums, `#[repr(isize)]`.

This change allows for values larger than `0x7FFFFFFF` to be stored in 32-bit bitmasks of type `u32`, without confusingly requiring the `#[allow(overflowing_literals)]` lint that suggests matching may not work, and fixes it being impossible to store 64-bit bitmasks with values larger than `0xffffffff`.

Because Rust does not allow for zero-variant enums to be given a non-default representation, this change required duplicating the entire final match arm of the `register_bitmasks!()` macro so that the representation is left unchanged in the case that no bitmasks are passed to an invocation of the macro (as happens often when `register_bitmasks!()` is called from within `register_bitfields!()`.

This change is part of a collection of recent PRs targeted at getting clippy to pass so that we can incorporate clippy into the CI pipeline.

### Testing Strategy

This pull request was tested by compiling on nrf52840dk after removing `#[allow(overflowing_literals)` and by running `run_clippy.sh` succesfully.

### TODO or Help Wanted

This pull request could probably incorporate other changes to clean up this macro -- the largely duplicated match arms could be separated out into smaller macros, for example -- but I was wary to change much more than this as this is the first time I have looked at the internals of this crate.

Also, is this change deserving of a version bump for the tock-register-interface crate?

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.

### DIFF
--- a/libraries/tock-register-interface/CHANGELOG.md
+++ b/libraries/tock-register-interface/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Changelog
 
-## master
+## v0.6
 
+ - #1823: Allow large unsigned values as bitmasks + add bitmask! helper macro
  - #1554: Allow lifetime parameters for `register_structs! { Foo<'a> { ..`
  - #1661: Add `Aliased` register type for MMIO with differing R/W behavior
 
--- a/libraries/tock-register-interface/CHANGELOG.md
+++ b/libraries/tock-register-interface/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## v0.6
+## master
 
  - #1823: Allow large unsigned values as bitmasks + add bitmask! helper macro
  - #1554: Allow lifetime parameters for `register_structs! { Foo<'a> { ..`
--- a/libraries/tock-register-interface/Cargo.toml
+++ b/libraries/tock-register-interface/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tock-registers"
-version = "0.5.0"
+version = "0.6.0"
 authors = ["Tock Project Developers <tock-dev@googlegroups.com>"]
 description = "Memory-Mapped I/O and register interface developed for Tock."
 homepage = "https://www.tockos.org/"
--- a/libraries/tock-register-interface/Cargo.toml
+++ b/libraries/tock-register-interface/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tock-registers"
-version = "0.6.0"
+version = "0.5.0"
 authors = ["Tock Project Developers <tock-dev@googlegroups.com>"]
 description = "Memory-Mapped I/O and register interface developed for Tock."
 homepage = "https://www.tockos.org/"
--- a/libraries/tock-register-interface/src/macros.rs
+++ b/libraries/tock-register-interface/src/macros.rs
@@ -127,8 +127,8 @@ macro_rules! register_bitmasks {
     {
         $valtype:ident, $reg_desc:ident, $(#[$outer:meta])* $field:ident,
                     $offset:expr, $numbits:expr,
-                    [$( $(#[$inner:meta])* $valname:ident = $value:expr ),*]
-    } => { //same pattern as previous match arm, except allows for 0 elements in array. Required because [repr(usize)] cannot be used for zero-variant enums.
+                    []
+    } => { //same pattern as previous match arm, for 0 elements in array. Removes code associated with array.
         #[allow(non_upper_case_globals)]
         #[allow(unused)]
         pub const $field: Field<$valtype, $reg_desc> =
@@ -141,15 +141,6 @@ macro_rules! register_bitmasks {
             #[allow(unused_imports)]
             use $crate::registers::{FieldValue, TryFromValue};
             use super::$reg_desc;
-
-            $(
-            #[allow(non_upper_case_globals)]
-            #[allow(unused)]
-            $(#[$inner])*
-            pub const $valname: FieldValue<$valtype, $reg_desc> =
-                FieldValue::<$valtype, $reg_desc>::new($crate::bitmask!($numbits),
-                    $offset, $value);
-            )*
 
             #[allow(non_upper_case_globals)]
             #[allow(unused)]
@@ -166,25 +157,13 @@ macro_rules! register_bitmasks {
             #[allow(dead_code)]
             #[allow(non_camel_case_types)]
             $(#[$outer])*
-            pub enum Value {
-                $(
-                    $(#[$inner])*
-                    $valname = $value,
-                )*
-            }
+            pub enum Value {}
 
             impl TryFromValue<$valtype> for Value {
                 type EnumType = Value;
 
-                fn try_from(v: $valtype) -> Option<Self::EnumType> {
-                    match v {
-                        $(
-                            $(#[$inner])*
-                            x if x == Value::$valname as $valtype => Some(Value::$valname),
-                        )*
-
-                        _ => Option::None
-                    }
+                fn try_from(_v: $valtype) -> Option<Self::EnumType> {
+                    Option::None
                 }
             }
         }
--- a/libraries/tock-register-interface/src/macros.rs
+++ b/libraries/tock-register-interface/src/macros.rs
@@ -51,8 +51,74 @@ macro_rules! register_bitmasks {
     {
         $valtype:ty, $reg_desc:ident, $(#[$outer:meta])* $field:ident,
                     $offset:expr, $numbits:expr,
-                    [$( $(#[$inner:meta])* $valname:ident = $value:expr ),*]
+                    [$( $(#[$inner:meta])* $valname:ident = $value:expr ),+]
     } => {
+        #[allow(non_upper_case_globals)]
+        #[allow(unused)]
+        pub const $field: Field<$valtype, $reg_desc> =
+            Field::<$valtype, $reg_desc>::new((1<<($numbits-1))+((1<<($numbits-1))-1), $offset);
+
+        #[allow(non_snake_case)]
+        #[allow(unused)]
+        $(#[$outer])*
+        pub mod $field {
+            #[allow(unused_imports)]
+            use $crate::registers::{FieldValue, TryFromValue};
+            use super::$reg_desc;
+
+            $(
+            #[allow(non_upper_case_globals)]
+            #[allow(unused)]
+            $(#[$inner])*
+            pub const $valname: FieldValue<$valtype, $reg_desc> =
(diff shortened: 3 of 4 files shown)


libraries/tock-register-interface/src/macros.rs
```diff
@@ -51,13 +51,79 @@ macro_rules! register_bitmasks {
...
+                    }
+                }
+            }
+        }
+    };
+    {
+        $valtype:ty, $reg_desc:ident, $(#[$outer:meta])* $field:ident,
+                    $offset:expr, $numbits:expr,
+                    [$( $(#[$inner:meta])* $valname:ident = $value:expr ),*]
+    } => { //same pattern as previous match arm, except allows for 0 elements in array. Required because [repr(usize)] cannot be used for zero-variant enums.
```
ppannuto, line 121: Can you add an equivalent comment to the first match arm that explains to future, linear readers that this arm is/will be duplicated and why?

hudson-ayers:
> Looking at the register interface, specifically:
> 
> ```rust
> register_bitfields! [
>     // First parameter is the register width for the bitfields. Can be u8, u16,
>     // u32, or u64.
>     u8,
> ```
> 
> Should we instead be using `repr( width )` as specified here, rather than always `usize`?

I didn't realize the interface claimed to supported 64 bit register widths, and looking throughout the codebase I don't see any instances in Tock where we specify a 64 bit bitmask, but I am pretty confident that doing so would not actually work in the current interface. I think that using `repr( width )` would be ideal, but when I tried doing `repr($valtype)` rust complains `error: expected unsuffixed literal or identifier, found `u32``. However I think `repr(u64)` should work for all cases because it fits 64 bits and the masks stored in the enum are casted to `$valtype` in `try_from()`.

> > Because Rust does not allow for zero-variant enums to be given a non-default representation
> 
> This feels possibly problematic. I don't have my finger on the exact path, but I'm concerned that flipping between a something like a `u8` and effectively an `i32` could get into trouble if something like sign extension pulls a large positive value negative and then tries to preserve that during truncation.
> 
> Do we know why Rust doesn't allow this?

I am a little confused here. Basically rust does not allow the following:
```rust
#[repr(anything)]
enum Value{}
```
because an enum with no variants is effectively just a never type (https://doc.rust-lang.org/reference/items/enumerations.html#zero-variant-enums). Given that the enum contains no values, it can't be instantiated, so I don't think I see a way that this could become problematic.

bradjc:
https://github.com/tock/tock/blob/903b90c0a4cdda691b549c6f235c87fe62459464/arch/rv32i/src/machine_timer.rs#L10-L24

hudson-ayers:
> https://github.com/tock/tock/blob/903b90c0a4cdda691b549c6f235c87fe62459464/arch/rv32i/src/machine_timer.rs#L10-L24

Yes I saw that, but we don't actually declare any bitmasks there

gendx:
> > Looking at the register interface, specifically:
> > ```rust
> > register_bitfields! [
> >     // First parameter is the register width for the bitfields. Can be u8, u16,
> >     // u32, or u64.
> >     u8,
> > ```
> > 
> > 
> > Should we instead be using `repr( width )` as specified here, rather than always `usize`?
> 
> I didn't realize the interface claimed to supported 64 bit register widths, and looking throughout the codebase I don't see any instances in Tock where we specify a 64 bit bitmask, but I am pretty confident that doing so would not actually work in the current interface. I think that using `repr( width )` would be ideal, but when I tried doing `repr($valtype)` rust complains `error: expected unsuffixed literal or identifier, found `u32``. However I think `repr(u64)` should work for all cases because it fits 64 bits and the masks stored in the enum are casted to `$valtype` in `try_from()`.

Regarding `repr($valtype)` in the macro, have you tried using `$valtype:ident` instead of `$valtype:ty`? The former will only allow an identifier token, which could be reused in the `repr`, whereas the latter in principle allows types of arbitrary complexity. But because Rust syntax forbids `repr(HashMap<Foo, Bar>)` for example, the macro engine rejects it alltogether (even if the actual type you pass is a single identifier). From my experience, doing this kind of change in a macro definition can remove the compilation error.

libraries/tock-register-interface/src/macros.rs
```diff
@@ -48,11 +48,79 @@ macro_rules! register_bitmasks {
...
+        $valtype:ty, $reg_desc:ident, $(#[$outer:meta])* $field:ident,
+                    $offset:expr, $numbits:expr,
+                    [$( $(#[$inner:meta])* $valname:ident = $value:expr ),+]
+    } => { // this match arm is duplicated below with an allowance for 0 elements in the valname -> value array,
+        // to seperately support the case of zero-variant enums not supporting non-default
+        // representations.
+        #[allow(non_upper_case_globals)]
+        #[allow(unused)]
+        pub const $field: Field<$valtype, $reg_desc> =
+            Field::<$valtype, $reg_desc>::new((1<<($numbits-1))+((1<<($numbits-1))-1), $offset);
```
gendx, line 61: I think I commented on this pattern before, but can't retrieve where. This arithmetic expression is quite unreadable, and appears many times in these macros. Could you add some helper macro or `const fn` that computes it? As I understand it, it a bitmask for `$numbits`, so `bitmask($numbits)` would be more readable here.

Also, an alternative expression would be: `!0 >> (8 * core::mem::size_of::<$valtype>() - $numbits)`, which may be more readable (fewer nested parentheses to parse for the reader).

hudson-ayers:
@gendx thanks for the hint on replacing `$ty` with `$ident`! That seems like the most correct change.

I also went ahead and separated out that arithmetic into a helper macro, but that change is logically separate from the rest of this PR so if anyone has reservations about it I will probably revert that change and it can be left for a future PR.

libraries/tock-register-interface/src/macros.rs
```diff
@@ -49,14 +57,82 @@ macro_rules! register_bitmasks {
...
+                        _ => Option::None
+                    }
+                }
+            }
+        }
+    };
+    {
+        $valtype:ident, $reg_desc:ident, $(#[$outer:meta])* $field:ident,
                     $offset:expr, $numbits:expr,
                     [$( $(#[$inner:meta])* $valname:ident = $value:expr ),*]
```
brghena, line 130: If this is really only for the case where there are zero elements in the array, why use `*` which specifies zero or more elements rather than having:

```
{ 
$valtype:ident, $reg_desc:ident, $(#[$outer:meta])* $field:ident,
                    $offset:expr, $numbits:expr,
                    []
}
```

The answer might be that my version straight up doesn't work for some reason.  I don't really understand macros very well. Right now it seems that you're relying on the macro_rules! to match with the prior case for anything with 1 or more elements which need the `repr`, which I suppose is how it works, but feels implicit.

brghena:
I think this combined with what's already in master is a good time for a `0.6.0` release.

libraries/tock-register-interface/src/macros.rs
hudson-ayers: using exactly that does not compile, for it to compile you also have to remove everything that uses `$inner`, `$valname` or `$value` within the portion of the macro executed by that match arm. Which I believe is actually the more correct approach, and doing so does make the macro compile, but I also am not on expert on macros and was a little worried that by changing more I would be more likely to break something.

brghena: I think making the change to remove `$inner`, `$valname`, and `$value` as well looks pretty straightforward. I'm looking at this code and it looks like the kind of situation where either it will all explode and not compile at all, or else all work.

brghena: Feel free to push back though.

hudson-ayers: went ahead and took that code out and everything still compiles -- I can't think of any reason the behavior should be different

libraries/tock-register-interface/src/macros.rs
```diff
@@ -1,12 +1,20 @@
 //! Macros for cleanly defining peripheral registers.
 
+/// Helper macro for computing bitmask of variable number of bits
+#[macro_export]
+macro_rules! bitmask {
+    ($numbits:expr) => {
+        (1 << ($numbits - 1)) + ((1 << ($numbits - 1)) - 1)
```
gendx, line 7: Would this be clearer as `!(!0 << $numbits)`? Or rather `!(!1 << ($numbits - 1))` to support the case where `$numbits` is equal to the size of the integer?

hudson-ayers: `!(!0 << $numbits)` is not equivalent to the existing code when `$numbits = 32` and compiled in release mode (see this rust playground example: https://play.rust-lang.org/?version=nightly&mode=release&edition=2018&gist=bd7681e2e730ce2830fe9b14523cffe8).

`!(!1 << ($numbits - 1))` is equivalent to the existing code for all values of numbits in the range (1..64) when compiled in the rust playground in release mode, but it makes me slightly uneasy that the current code panics for `$numbits = 32` when compiled with debug_assertions=true and this new example does not -- does that mean they are not guaranteed to be equivalent to the existing code in the case where `$numbits = 32`?

It seems like `!(!1 << ($numbits - 1))` might be more correct than the current code, but I don't totally understand the current code and am wary of adding a commit that appears to introduce a possible logic change when we may be relying on the existing behavior somehow

gendx: Yes, the `!(!0 << $numbits)` formula only works for `$numbits` in `0..32`, whereas the other two work for `$numbits` in `1..=32` (assuming a `u32`). This is because shifting by 32 bits is undefined behavior (contrary to what one may expect of it "clearing" all bits, some/most CPUs will mask the shift and shift zero bits instead).

I don't think we consider the case where `$numbits` is zero (which would be quite uninteresting), so the current formula is fine, and so would be `!(!1 << ($numbits - 1))`. I can propose a separate PR for that.

libraries/tock-register-interface/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tock-registers"
-version = "0.5.0"
+version = "0.6.0"
```
ppannuto, line 3: Please don't change this here; I'll make a separate release PR

libraries/tock-register-interface/CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 # Changelog
 
-## master
+## v0.6
```
ppannuto, line 3: (same), but leave the added bullet

hudson-ayers:
Release update removed @ppannuto

hudson-ayers:
bors retry

hudson-ayers:
bors retry

PR #1825, opened 2020-05-05
Run Clippy in CI

hudson-ayers:
### Pull Request Overview

This pull request adds a make target to run clippy, and runs clippy in CI (both Travis and gh-actions). 

It updates the PR template to prompt users to run `make format` and correct errors surfaced by `make clippy`.

This PR also replaces `check_wildcard_imports.sh` with the newly introduced clippy flag that performs that check.

### Testing Strategy

This pull request was tested by running `make clippy` and by our CI suite.

### TODO or Help Wanted

Opinions on the Makefile name changes are welcome.

### Documentation Updated

### Formatting

- [x] Ran `make format`.
- [x] Ran `make clippy` without errors.


Makefile
```diff
@@ -24,7 +24,8 @@ usage:
 	@echo " emulation-check: Run the emulation tests for supported boards"
 	@echo "           clean: Clean all builds"
 	@echo "          format: Runs the rustfmt tool on all kernel sources"
-	@echo "       formatall: Runs all formatting tools"
+	@echo "          clippy: Runs the clippy code linter on all kernel sources"
```
ppannuto, line 27: Ah! I missed this when emulation-* came in, but this list used to be in alpha-order; can you fix emulation and put clippy in the right place too?

tools/run_clippy.sh
```diff
@@ -31,6 +35,7 @@ CLIPPY_ARGS="
 -D clippy::default_trait_access
 -D clippy::option_map_unwrap_or
 -D clippy::option_map_unwrap_or_else
+-D clippy::wildcard_imports
```
gendx, line 38: Awesome!

Makefile
```diff
@@ -20,11 +20,12 @@ usage:
...
 	@echo "              ci: Run all continuous integration tests"
-	@echo " emulation-setup: Setup QEMU for the emulation tests"
-	@echo " emulation-check: Run the emulation tests for supported boards"
 	@echo "           clean: Clean all builds"
+	@echo "          clippy: Runs the clippy code linter on all kernel sources"
+	@echo " emulation-check: Run the emulation tests for supported boards"
+	@echo " emulation-setup: Setup QEMU for the emulation tests"
 	@echo "          format: Runs the rustfmt tool on all kernel sources"
-	@echo "       formatall: Runs all formatting tools"
+	@echo "       formatall: Runs all formatting tools (rustfmt and clippy)"
```
gendx, line 28: I find this confusing. `rustfmt` will modify the code by formatting it. But `clippy` is only a linter, and will not format the code at all.

Also, I think we have too many of these rules, which is too complex to reason about: `fmt`, `format`, `formatall`, `ci-formatting`, `ci-format`.

To me, there are two categories:
- A rule that actually modifies the code, which runs `./tools/run_cargo_fmt.sh`, and should be named `format`. I don't think we need `fmt` nor `formatall` spellings. There is no `ci` equivalent because CI only checks the code.
- Rules that check the code. I suggest:
  - `check-format` (or `format-check`), which runs `CI=true ./tools/run_cargo_fmt.sh diff`
  - `lint` (or `lints`), which runs both `check-format` and `clippy`
  - `ci-lint`, with the CI banner.

I'm not sure what `ci-documentation` has to do with any of these.

hudson-ayers: `ci-format` is the grouping of stuff run by the github actions job with the name `ci-format`. I didn't want to break down every single little ci target we have to run in a separate job in github actions, but wanted to have a top level target people can run for when the job with that name fails.

`ci-documentation` is included in that because I consider validating a table of contents as a formatting check.

replacing `formatall` with `check-format` as you have described makes sense to me.

I don't see any need to get rid of the `fmt` spelling as some people may be used to using it.

Makefile
```diff
@@ -234,10 +234,18 @@ clean:
 	@echo "$$(tput bold)Clean rustdoc" && rm -Rf doc/rustdoc
 	@echo "$$(tput bold)Clean ci-artifacts" && rm -Rf ./ci-artifacts
 
-.PHONY: fmt format formatall
+.PHONY: fmt format
 fmt format formatall:
```
gendx, line 238: There is still `formatall` on this line.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -19,4 +19,5 @@ This pull request still needs...
 
 ### Formatting
 
-- [ ] Ran `make formatall`.
+- [ ] Ran `make format`.
+- [ ] Fixed errors surfaced by `make lints`.
```
brghena, line 23: Why not tell them to run `make clippy` instead since they're already running `format`? Then you could get rid of `lints` altogether and just have `ci-lints` call `check-format` and `clippy` separately.

brghena: Or put a different way, do we really need a single target for running both operations?

hudson-ayers:
Rebased on master, this is unblocked now that #1823 is merged

.github/workflows/ci.yml
```diff
@@ -29,7 +29,9 @@ jobs:
         with:
           components: rustfmt, clippy
       - name: ci-formatting
```
gendx, line 31: This could be named "check-format", given that there is no ci-formatting rule in the Makefile.

Makefile
```diff
@@ -108,13 +111,13 @@ ci-tests:\
...
 ## Actual Rules (Travis)
 
-.PHONY: ci-formatting
-ci-formatting:
-	@printf "$$(tput bold)******************$$(tput sgr0)\n"
-	@printf "$$(tput bold)* CI: Formatting *$$(tput sgr0)\n"
-	@printf "$$(tput bold)******************$$(tput sgr0)\n"
-	@CI=true ./tools/run_cargo_fmt.sh diff
-	@./tools/check_wildcard_imports.sh
+.PHONY: ci-lints
```
gendx, line 114: I don't see this rule used anywhere. Is it needed?

hudson-ayers: Oh, good catch, I meant for it to be used by Travis.

hudson-ayers:
Same issue as previously that Bors cannot merge changes to Github workflows, so merging manually.

Issue #178, opened 2016-11-07
TakeCell takes many cycles to run


brghena:
While debugging why Tock couldn't keep up with UART bytes, I realized that the issue seems to be the amount of time `takecell.map()` is taking when mapping over the App struct in the `nrf51822_serialization` capsule. I believe this is contrary to our expectations that `TakeCell` operations are simple and efficient.

The [takecell-timing branch](https://github.com/helena-project/tock/tree/takecell-timing) has code that demonstrates the issue. To test, program on a firestorm along with the `ble-serialization` app and probe pins P2, P3, and P4. The results are documented inline with comments in the ['nrf51822_serialization` capsule](https://github.com/helena-project/tock/blob/takecell-timing/capsules/src/nrf51822_serialization.rs#L253)

The example that is not commented out uses an instrumented version of `TakeCell` in order to break down where the time is spent. (All instrumentation performed by toggling GPIO pins) The pin toggles can be seen [inline](https://github.com/helena-project/tock/blob/8d407a31523f8ae7fee58e157ae9144c587f8112/capsules/src/instrumented_takecell.rs#L108). The results are: 

| Call | Duration |
|:----|:----------|
| `self.take()` | 10.49 us |
| `maybe_val.map()` | 4.57 us |
| `closure()` | 9.82 us |
| `replace()` | 22.22 us |
| `maybe_val.map()`'s `)};` | 1.33 us |

Attached is the assembly `.lst` file from this example. The code is marked up with `AMIT` on lines where GPIO pins are toggled. **Attachment:** [instrumented_takecell_lst.txt](https://github.com/helena-project/tock/files/574289/instrumented_takecell_lst.txt)

As an example of why this issue is important, when bytes are coming in to the `nrf51822_serialization` capsule at 250000 baud, it has 24 us to process and start a new transaction between bytes. Due to the overhead from `TakeCell`, it takes 48 us to do the processing (TakeCell accounting for 38 us, Closure for 10s). There are workarounds to deal with this (which is how the code is currently implemented), but it would be better to not have to deal with high overhead in the kernel.

**TLDR:** `TakeCell` takes much longer than expected to run and it's complicating drivers by slowing them down.

immesys:
[...]

PR #1826, opened 2020-05-05
kernel: use MapCell for Process::stored_state

alphan:
### Pull Request Overview

Currently, `Process::stored_state` is defined as a `Cell`. This change
aims to improve syscall latency by using `MapCell` instead.

See also: #1730, #178.

### Testing Strategy

I ran the same `libtock-rs` app used to test #1822 with and without this
change on opentitan on verilator. This change saves ~1200 cycles for
a `command` syscall.

### TODO or Help Wanted

The main goal of `MapCell` is to avoid panics at runtime, but I'm not sure
if it is OK to silently continue in cases where `Process::stored_state`can 
_possibly_ be empty, which should never happen. The current
implementation exclusively uses `Cell::get`, but a `take` could always
leave a `default` `StoredState` behind so this change doesn't seem to be
worse in that regard. I tried to preserve the current behavior but please let
me know if I missed anything. 

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -736,7 +736,7 @@ pub struct Process<'a, C: 'static + Chip> {
     /// State saved on behalf of the process each time the app switches to the
     /// kernel.
     stored_state:
-        Cell<<<C as Chip>::UserspaceKernelBoundary as UserspaceKernelBoundary>::StoredState>,
+        MapCell<<<C as Chip>::UserspaceKernelBoundary as UserspaceKernelBoundary>::StoredState>,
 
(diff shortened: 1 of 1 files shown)


alphan:
> This looks good! This also enables us to remove the Copy bound from the `StoredState` associated type, and the concrete implementations of it (`RiscvimacStoredState`, `CortexMStoredState`) as we no longer copy `StoredState` objects anywhere -- that might be a good change to add to this PR?

Good point! Created #1833, happy to take care of it, please feel free to assign to me.

PR #1828, opened 2020-05-05
OpenTitan: Add support for receiving data over the UART

alistair23:
### Pull Request Overview

This PR adds support to the LowRISC UART driver to read data from the UART.

### Testing Strategy

Tested by running the `console_recv_short` libtock-c example on the OpenTitan QEMU model and OpenTitan FPGA.

### Documentation Updated

### Formatting

- [X] Ran `make formatall`.

### DIFF
--- a/chips/lowrisc/src/uart.rs
+++ b/chips/lowrisc/src/uart.rs
@@ -9,6 +9,7 @@ use kernel::common::registers::{
 };
 use kernel::common::StaticRef;
 use kernel::hil;
+use kernel::hil::uart;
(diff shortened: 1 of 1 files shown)


bradjc:
It appears to be OK per the HIL to not wait until all the requested bytes are received, but the callback should return `ESIZE` in that case:

https://github.com/tock/tock/blob/903b90c0a4cdda691b549c6f235c87fe62459464/kernel/src/hil/uart.rs#L266-L278

alistair23:
Updated

PR #1443, opened 2019-11-01
Add strace feature to trace syscalls in the kernel.

gendx:
### Pull Request Overview

This pull request adds an "strace" configuration to the kernel, which causes a trace of all syscalls and callbacks to be displayed on the debug port.

This also adds a new way of statically configure the kernel, via a `kernel/src/config.rs` file containing a static const configuration object. This allows to quickly turn on/off configuration values during development, while making sure that un-configured code is still checked for syntax and types, and allowing the compiler to optimize away dead code (due to the configuration being `const`).

### Testing Strategy

This pull request was tested by:
- Running `make ci-travis`.
- Flashing Tock and an app on an nRF52840dk board and checking the debug output.

### TODO or Help Wanted

N/A

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

PR #1765, opened 2020-04-17
Remove &dyn references from GPIO-related HILs and capsules.

gendx:
### Pull Request Overview

This pull request experiments with removing `&dyn` references from GPIO-related capsules, as a supporting example for #1761.

Until now, the capsules were using dynamic references to use GPIO pins, even though for a given board/chip the GPIOs have a type known at compile time. This pull request proposes to replace such dynamic references with explicit types.

The resulting code is a bit more verbose, but surprisingly not so much. Most of the extra code comes from components, where `static_init_half` is now necessary to instantiate the static types that are now generic.

By applying this refactoring to the most common capsules (GPIO, button, LED), savings of the order of ~1KB can be obtained on the boards (depending on how many GPIOs are declared by the board). More savings could be obtained by extending this refactoring to all capsules using GPIOs. And more generally many HILs in the kernel rely on `&dyn` references, so removing them from HILs and capsules in general could achieve even bigger savings.

Before:

```
   text	   data	    bss	    dec	    hex	filename
 102400	   3196	  62336	 167932	  28ffc	/tock/target/thumbv7em-none-eabi/release/hail
 164864	   3272	  54072	 222208	  36400	/tock/target/thumbv7em-none-eabi/release/imix
  68609	   1952	  79968	 150529	  24c01	/tock/target/thumbv7em-none-eabi/release/nucleo_f429zi
```

After:

```
   text	   data	    bss	    dec	    hex	filename
 101888	   3196	  62336	 167420	  28dfc	/tock/target/thumbv7em-none-eabi/release/hail
   -512       =       =            -200
 163840	   3272	  54072	 221184	  36000	/tock/target/thumbv7em-none-eabi/release/imix
  -1024       =       =            -400
  66561	   1952	  79968	 148481	  24401	/tock/target/thumbv7em-none-eabi/release/nucleo_f429zi
  -2048       =       =            -800
```

### Testing Strategy

This pull request was tested by building a few boards and checking the binary size. Not much more testing at this stage.

### TODO or Help Wanted

This pull request still needs:
- Feedback about such a revamp of the HIL.
- Feedback about whether reducing code size by this method is useful (see also #1663).
- Feedback about the components.
- Updating the documentation.
- Testing on physical boards.

### Documentation Updated

- [ ] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

Issue #1761, opened 2020-04-14
About the overhead and constraints of dynamic dispatch (`&dyn` references)


gendx:
Currently, the Tock kernel makes extensive use of "trait objects" (recognizable by `&dyn` references). One such example is the `Process` type, which implements the `ProcessType` trait. The rest of the kernel references processes via `&dyn ProcessType`.

The use of such a dynamic trait object is motivated by the fact that the actual `Process` type is parametrized by a `Chip`, so using a `&Process<ChipFoo>` everywhere instead of `&ProcessType` would make the code more verbose because the `Chip` parameter would have to be recursively given by all structs/traits referring to a process.

However, I see multiple problems with this approach.
- **Dynamic trait objects** have in principle some **overhead**. This is because in principle each object refers to a vtable to resolve at runtime all of the functions in the trait. The reason being that there can be multiple implementations of the trait, and `&dyn ProcessType` could be any struct that implements the trait.
  However, in practice, for a given board there is a single chip and therefore a single implementation of a `Process`. So all the methods in a process could instead be resolved at compile time, and even optimized (inlined by the compiler, etc.).
- Because of the constraints of storing all methods in a vtable, traits used by trait objects have some programming limitations. For example, the **methods in such traits cannot themselves be parametrized** by other types. This problem came up in https://github.com/tock/tock/pull/1648#discussion_r405022427. Essentially, we cannot have the following `alloc` function:
  ```rust
  pub trait ProcessType {
      fn alloc<T: Sized>(&self) -> Option<NonNull<T>>;
  }
  ```
  and then use `&dyn ProcessType`. Essentially, this is because the compiler cannot give a dynamic vtable pointer for such a generic `alloc` function - as it would need to have a function pointer for every possible `T` one might instantiate `alloc` with. This limitation is shown by the following compiler error.
  ```
  error[E0038]: the trait `process::ProcessType` cannot be made into an object
     --> tock/kernel/src/sched.rs:327:18
      |
  327 |         process: &dyn process::ProcessType,
      |                  ^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `process::ProcessType` cannot be made into an object
      |
     ::: tock/kernel/src/process.rs:183:11
      |
  183 | pub trait ProcessType {
      |           ----------- this trait cannot be made into an object...
  ...
  350 |     fn alloc<T: Sized>(&self) -> Option<NonNull<T>>;
      |        ----- ...because method `alloc` has generic type parameters
      |
[...]

Issue #1663, opened 2020-03-05
[Tracking] Optimizing Tock for code size


gendx:
Given the recent interest in optimizing Tock for code size, I think it would make sense to summarize and track existing and future ways of decreasing the size.

- [x] **Capsule configuration in boards.** This is a well-established way of trimming unused code. Despite dozens of capsules being available, each board can pick those that are relevant. By default, board configurations include many drivers as a baseline, but one can easily remove unused drivers to trim down Tock's size for their particular use case.
- [x] (#1443) **Kernel's const configuration.** By tuning `const` parameters in `kernel/src/config.rs`, once can enable "niche" debugging features, such as syscall tracing or loading processes. When turned off, the Rust compiler can fold the constants and remove dead code.
- [ ] (#1660) **Disabling frame pointers.** This is work-in-progress to decrease code size by 2-10%. Frame pointers might be useful for debugging, but we could add configuration for it.
- [ ] (#1372) **Stripping debug statements.** The idea would be to replace the `debug!` macro's generated code by an empty block, under some configuration option (e.g. the kernel's const configuration). This would remove code in itself, as well as a bunch of strings. The compiler could then further optimize the rest of the code. A more aggressive pruning strategy would also remove the UART-related capsules under such as "no-debug" mode.
- [ ] (#1662) **[Unclear] Static mut doesn't help inlining?** Because the compiler can assume that static mutable data is modified elsewhere, it may not optimize (inlining, constant folding) expressions that are in fact constant.

gkelly:
I think removing the debug statements (or making a pluggable logging driver that could do things like printing offsets into `.rodata` and format arguments, for example) might be a big win, especially if results in us completely removing the `fmt` machinery.

ppannuto:
Re the first checkbox, there is still some work to do:
https://github.com/tock/tock/issues/625

gendx:
> Re the first checkbox, there is still some work to do: #625

Thanks for the pointer! I added it to the list.

alevy:
RE: debug statements, one "trivial" way to strip code down by a lot is to remove modify the `panic` handler for a board with one that doesn't output any formatted strings.

E.g., on the Hail board, the following change:

```diff
diff --git a/boards/hail/src/io.rs b/boards/hail/src/io.rs
index 4aef525e0..a3c6a5aab 100644
--- a/boards/hail/src/io.rs
+++ b/boards/hail/src/io.rs
[...]

PR #1830, opened 2020-05-06
kernel: add a config switch for process debug info

alphan:
### Pull Request Overview

This change adds a config switch to control the collection of debugging information for processes and improve syscall latency.

I also added the `debug_` prefix to two functions and refactored some code to new `debug_*` functions.

See also: #1730, #1443.

### Testing Strategy

I ran the app used to test #1826 with `debug_processes` both `true` and `false`. Setting `debug_processes = false` saves ~60 cycles when `opt-level=3`, `lto=true`, `debug=false` and ~200 cycles when `codegen-units=1` in addition to the previous settings.

### Documentation Updated

### Formatting

- [X] Ran `make formatall`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -299,15 +299,6 @@ pub trait ProcessType {
     /// writeable flash region.
     fn get_writeable_flash_region(&self, region_index: usize) -> (u32, u32);
 
-    /// Debug function to update the kernel on where the stack starts for this
-    /// process. Processes are not required to call this through the memop
-    /// system call, but it aids in debugging the process.
-    fn update_stack_start_pointer(&self, stack_pointer: *const u8);
-
-    /// Debug function to update the kernel on where the process heap starts.
-    /// Also optional.
-    fn update_heap_start_pointer(&self, heap_pointer: *const u8);
-
     // additional memop like functions
 
     /// Creates an `AppSlice` from the given offset and size in process memory.
@@ -415,6 +406,15 @@ pub trait ProcessType {
     unsafe fn print_full_process(&self, writer: &mut dyn Write);
 
     // debug
+
+    /// Debug function to update the kernel on where the stack starts for this
+    /// process. Processes are not required to call this through the memop
+    /// system call, but it aids in debugging the process.
+    fn debug_update_stack_start_pointer(&self, stack_pointer: *const u8);
+
+    /// Debug function to update the kernel on where the process heap starts.
+    /// Also optional.
+    fn debug_update_heap_start_pointer(&self, heap_pointer: *const u8);
 
     /// Returns how many syscalls this app has called.
     fn debug_syscall_count(&self) -> usize;
@@ -793,9 +793,7 @@ impl<C: Chip> ProcessType for Process<'a, C> {
         // Make a note that we lost this callback if the enqueue function
         // fails.
         if ret == false {
-            self.debug.map(|debug| {
-                debug.dropped_callback_count += 1;
-            });
+            self.debug_callback_dropped();
         }
 
         ret
@@ -918,26 +916,6 @@ impl<C: Chip> ProcessType for Process<'a, C> {
 
     fn get_writeable_flash_region(&self, region_index: usize) -> (u32, u32) {
         self.header.get_writeable_flash_region(region_index)
-    }
-
-    fn update_stack_start_pointer(&self, stack_pointer: *const u8) {
-        if stack_pointer >= self.mem_start() && stack_pointer < self.mem_end() {
-            self.debug.map(|debug| {
-                debug.app_stack_start_pointer = Some(stack_pointer);
-
-                // We also reset the minimum stack pointer because whatever value
-                // we had could be entirely wrong by now.
-                debug.min_stack_pointer = stack_pointer;
-            });
-        }
-    }
-
-    fn update_heap_start_pointer(&self, heap_pointer: *const u8) {
-        if heap_pointer >= self.mem_start() && heap_pointer < self.mem_end() {
-            self.debug.map(|debug| {
-                debug.app_heap_start_pointer = Some(heap_pointer);
-            });
-        }
     }
 
     fn setup_mpu(&self) {
@@ -1182,7 +1160,7 @@ impl<C: Chip> ProcessType for Process<'a, C> {
 
                 // Update helpful debugging metadata.
                 self.current_stack_pointer.set(stack_bottom as *mut u8);
-                self.debug_set_max_stack_depth();
+                self.debug_update_max_stack_depth();
             }
 
             Some(Err(bad_stack_bottom)) => {
@@ -1193,12 +1171,7 @@ impl<C: Chip> ProcessType for Process<'a, C> {
                 // We also update the debugging metadata so that if the process
                 // fault message prints then it should be easier to debug that
                 // the process exceeded its stack.
-                self.debug.map(|debug| {
-                    let bad_stack_bottom = bad_stack_bottom as *const u8;
-                    if bad_stack_bottom < debug.min_stack_pointer {
-                        debug.min_stack_pointer = bad_stack_bottom;
-                    }
-                });
+                self.debug_update_max_stack_depth_using_pointer(bad_stack_bottom as *const u8);
                 self.set_fault_state();
             }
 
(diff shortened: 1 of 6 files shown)


kernel/src/process.rs
```diff
@@ -1804,8 +1847,9 @@ impl<C: 'static + Chip> Process<'a, C> {
                 process
                     .current_stack_pointer
                     .set(new_stack_pointer as *mut u8);
-                process.debug_set_max_stack_depth();
-                process.stored_state.set(stored_state);
```
hudson-ayers, line 1808: I think this line needs to stay (though this will be moot when your other PR is merged).

alphan: Yes, maybe I should have included the commit in #1826 in this PR too.

bradjc:
> I also think that when this config option is enabled, the process console should report that the values returned are inaccurate, and can be enabled by toggling this option in the kernel.

This is one example of why these config options get scary...reasoning about what works and doesn't work with every single combination of config options, and trying to correctly handle every case is very difficult to scale.

kernel/src/process.rs
alphan: Done.

alphan:
- `debug_processes` is now on by default.
- Added warning messages in process console (list command output) and `print_full_process()`.
  - Introduced an associated function (could be an associated constant too, not sure which one is preferred) to make `debug_processes` available in the process console.

alphan:
> > I also think that when this config option is enabled, the process console should report that the values returned are inaccurate, and can be enabled by toggling this option in the kernel.
> 
> This is one example of why these config options get scary...reasoning about what works and doesn't work with every single combination of config options, and trying to correctly handle every case is very difficult to scale.

I agree that these config options can have unintended consequences, especially when they break the assumptions of existing clients as in this case, but hopefully the latest changes address @hudson-ayers's comments.

kernel/src/process.rs
```diff
@@ -1528,6 +1570,15 @@ impl<C: Chip> ProcessType for Process<'a, C> {
...
     unsafe fn print_full_process(&self, writer: &mut dyn Write) {
+        if config::CONFIG.debug_processes == false {
+            let _ = writer.write_str(
+                "\
+                \r\nWARNING: Values reported below are inaccurate because\
+                \r\nprocess debugging is currently disabled. This setting\
+                \r\ncan be enabled in `kernel/src/config.rs`.\
+                \r\n",
+            );
+        }
```
gendx, line 1581: If `debug_processes` is disabled, could this function simply write this banner and skip all of the rest? This would also reduce code size because the rest of the function would be dead code.

Also, will this banner be printed once per process, or once for all processes?

kernel/src/process.rs
```diff
@@ -431,6 +441,13 @@ pub trait ProcessType {
     /// Increment the number of times the process called a syscall and record
     /// the last syscall that was called.
     fn debug_syscall_called(&self, last_syscall: Syscall);
+
+    /// Reset some debug information after the process restarts.
+    fn debug_restarted(&self);
+
+    /// Collect information about context switches. Currently only counts
+    /// the number of context switches due to timeslice expirations.
+    fn debug_context_switch(&self, switch_reason: &Option<syscall::ContextSwitchReason>);
```
bradjc, line 450: Many of these do not need to be in the trait. If they aren't called from outside process.rs we can just implement them on the struct.

kernel/src/sched.rs
```diff
@@ -284,6 +284,14 @@ impl Kernel {
         }
     }
 
+    /// Returns whether the kernel is configured to collect debugging information for processes.
+    ///
+    /// Clients that rely on this information, e.g. process console, should call this method
+    /// to check if process debugging is enabled and warn the user if it is not.
+    pub fn process_debugging_enabled() -> bool {
```
bradjc, line 291: Rather than this, which is very easy to forget to call, what about making the debug functions that can return useless values return `None`?

kernel/src/sched.rs
```diff
@@ -364,7 +374,9 @@ impl Kernel {
                             process.set_fault_state();
                         }
                         Some(ContextSwitchReason::SyscallFired { syscall }) => {
-                            process.debug_syscall_called(syscall);
+                            if config::CONFIG.debug_processes {
+                                process.debug_syscall_called(syscall);
+                            }
```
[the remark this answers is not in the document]
brghena: If so, it might be more straightforward to relocate all of the `if config...` checks into the debug functions themselves rather than having them sprinkled throughout the code.

alphan:
- Moved some `debug_*` functions from the trait to the struct, added early returns, and removed if statements from call sites.
- Rebased on master.

kernel/src/process.rs
alphan: > If debug_processes is disabled, could this function simply write this banner and skip all of the rest? This would also reduce code size because the rest of the function would be dead code.

It could, but this function also prints some other useful information that is not affected by `debug_processes`, e.g. stored state of the process and MPU config.


> Also, will this banner be printed once per process, or once for all processes?

This banner will be printed per process. I can move the banner to `panic_process_info` since it is the only caller for now but I thought keeping it here would be safer, just in case this function is called from somewhere else in the future. It is also harder to miss when there is a lot of debug output.

Please let me know what you think.

kernel/src/sched.rs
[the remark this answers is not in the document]
alphan: I actually tried this before creating this PR but tried again after implementing @bradjc's comment below to make sure. Benchmarking something like this is tricky but IIUC it doesn't work for functions that are on the trait but it works when they are implemented on the struct. 
This conclusion is not based on the number of cycles, because it depends on compiler settings and can be misleading, but `objdump` output. I see calls to `debug_*` functions that are on the trait when `debug_processes = false` (even when they have an early return) unless they are wrapped in an if statement. Thus, I was able to move the if statements into the `debug_*` functions implemented on the struct but kept the rest (for calls outside process.rs). I hope this is an acceptable compromise.

kernel/src/process.rs
alphan: Done. Thanks for pointing this out, this lead me to a rabbit hole of compiler settings :) Please see my comment above.

kernel/src/sched.rs
alphan: I added this to be able to print the warning banner in the process console.

`debug_*` functions currently return `0` if `Process::debug` is not there when `debug_processes` is disabled.  It looks like returning `None` instead of `0` would result in a lot of changes because callers expect `usize` from these functions and it could still possibly be misinterpreted by someone not familiar with the output. Please let me know what you think.

kernel/src/process.rs
gendx: I thought that printing the banner once would be enough, and avoid verbosity, but if you think it's better to print it multiple times to avoid missing it I don't have a strong opinion either.

kernel/src/sched.rs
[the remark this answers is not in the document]
gendx: I don't think functions on the `ProcessType` trait can be effectively optimized away by the compiler, because this trait is currently used as trait objects, that have dynamic dispatch (even though in practice there's only one type per compiled board).

If you're interested, I've started documenting this overhead in #1761 (see also #1663).
Removing `&dyn` on some GPIOs types recently led to code size improvements in #1765.

kernel/src/sched.rs
gendx: If the kernel config needs to be exposed outside of the kernel crate, shouldn't we make the config module public, or expose a function returning a (read-only) reference to `config::CONFIG`?

kernel/src/sched.rs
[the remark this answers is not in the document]
alphan: I find it natural that the functions on the trait remain, but I am not sure why the calls are not optimized away. Maybe they are optimized in other cases as a side-effect? Thanks for the issue numbers, I'll take a look!

kernel/src/sched.rs
hudson-ayers: Is it safe to assume that anything in config::CONFIG should be public to capsules? 

Though I agree that a function returning a specific configuration feature probably belongs in `config.rs`, not `sched.rs`. I suppose that an alternative would be to make `Config` a pub struct, but then internally use the `crate` modifier on any fields we don't want exposed to capsules?

alphan: @gendx, I agree but I just wasn't sure how much of it should be exposed so I exposed the new flag only. The function is implemented in `kernel/src/sched.rs` because I thought `Kernel` was a reasonable home for a kernel config but if it is OK to make `Config` a pub struct, I can definitely implement @hudson-ayers's suggestion.

kernel/src/process.rs
```diff
@@ -1986,4 +1966,64 @@ impl<C: 'static + Chip> Process<'a, C> {
...
         current_state != State::StoppedFaulted && current_state != State::Fault
     }
+
+    /// Update the maximum stack depth that the process has ever used.
+    fn debug_update_max_stack_depth(&self) {
+        if config::CONFIG.debug_processes == false {
+            return;
+        }
+        self.debug_update_max_stack_depth_using_pointer(self.current_stack_pointer.get());
+    }
```
brghena, line 1976: Does this compile differently than:

```rust
    fn debug_update_max_stack_depth(&self) {
        if config::CONFIG.debug_processes {
             self.debug_update_max_stack_depth_using_pointer(self.current_stack_pointer.get());
        }
    }
```

Generally, I'd say writing it this way reads clearer.

alphan: Done.

kernel/src/sched.rs
gendx: I think @hudson-ayers's suggestion would be the way to go. By the way, we should use `pub(crate)` now, as was finally decided in https://github.com/tock/tock/pull/1650#issuecomment-629399631.

On the other hand, it seems to me that instead of exposing the config to capsules, we could just add a function to the kernel that contains the following code, and only expose this function for capsules.

https://github.com/tock/tock/blob/44875ac55169c99bcb15222e2c84d0726dce3d2f/capsules/src/process_console.rs#L252-L255

The code and text for this banner is already duplicated anyway, so having a single function would also reduce the size of the binary.

https://github.com/tock/tock/blob/44875ac55169c99bcb15222e2c84d0726dce3d2f/kernel/src/process.rs#L1412-L1420

alphan:
Implemented @hudson-ayers's and @gendx's comments (removed `process_debugging_enabled()` and exposed `debug_processes` to capsules).

@gendx, I think capsules may want to do different things based on the value of this flag instead of just printing a warning. So, I thought it would be better to expose this flag. Re. duplication, how about moving the message to a function that returns a static str? Also, where should such a function be defined?

gendx:
I think the main concern about exposing the configuration outside of the kernel would be that it could be harder to refactor in the future if capsules start relying on it. So I see the argument of keeping it private unless putting it in the public API is really necessary.

kernel/src/sched.rs
bradjc: I do think it would require at least some changes, but if a `debug_` call can now fail when before it couldn't, I think that should be reflected in its return type. Expecting callers to first check a flag in a completely different file before knowing if a return value is valid seems like an API we don't want.

phil-levis:
Debugging information is probably one of the toughest of these (config option questions), because you want to be able to selectively choose outputs depending on what you are debugging. Code space means you can't just always leave it on.

The process console is a capsule, though. We find it useful, but it shouldn't be that everyone has to pay a tax for it. None seems like avoiding the problem: we should be able to detect this kind of mismatch at compile time. Else we are paying a big cost for this super-long string that's explaining our configuration mismatch. Is there some way to redesign this interface so we can detect it at compile-time (e.g., the absence of a trait or module)?

bradjc:
Moved to #1995

PR #1831, opened 2020-05-06
Allow/subscribe null pointer doc clarifications

jrvanwhy:
Correct the documentation of the allow() system call to indicate that passing a null pointer only asks a driver to stop accessing a region; it does not revoke that driver's access.

I also clarified the subscribe documentation to indicate that passing a null callback asks the driver to stop calling it rather than revokes its access to the callback.

### Pull Request Overview

This pull request changes the system call documentation for `allow` and `subscribe`.

### Testing Strategy

This pull request was tested by looking [the rendered documentation](https://www.github.com/jrvanwhy/tock/blob/allow-subscribe-null/doc/Syscalls.md).

### Documentation Updated

### Formatting

- [X] Ran `make formatall`.

### DIFF
--- a/doc/Syscalls.md
+++ b/doc/Syscalls.md
@@ -157,8 +157,9 @@ A callback function is uniquely identified by the pair (`driver`,
 are any pending callbacks for this callback ID, they are removed from the queue
 before the new callback function is bound to the callback ID.
 
-Passing a null pointer callback function disables a previously set callback
-(besides flushing pending callbacks for this callback ID).
+A process can pass a null pointer as the callback argument to request the driver
+disable a previously set callback (besides flushing pending callbacks for this
+callback ID).
 
 ```rust
 subscribe(driver: u32, subscribe_number: u32, callback: u32, userdata: u32) -> ReturnCode as u32
@@ -224,7 +225,8 @@ additional meaning such as the number of devices present, as is the case in the
 ### 3: Allow
 
 Allow marks a region of memory as shared between the kernel and application.
-A null pointer revokes sharing a region.
+Passing a null pointer requests the corresponding driver to stop accessing the
+shared memory region.
 
 ```rust
 allow(driver: u32, allow_number: u32, pointer: usize, size: u32) -> ReturnCode as u32


doc/Syscalls.md
```diff
@@ -224,7 +225,8 @@ additional meaning such as the number of devices present, as is the case in the
 ### 3: Allow
 
 Allow marks a region of memory as shared between the kernel and application.
-A null pointer revokes sharing a region.
+Passing a null pointer requests the corresponding driver to stop accessing the
+shared memory region.
```
ppannuto, line 229: I _think_ the intent was actually a strong revoke -- namely that if a capsule held an `AppSlice` it would fail on the next attempt to access that `AppSlice`. I'm not sure that's implemented, but I think that's still the aspiration.

We could consider updating this to caveat that this is imperfectly implemented, but I don't think we want to change this contract?

doc/Syscalls.md
```diff
@@ -157,8 +157,9 @@ A callback function is uniquely identified by the pair (`driver`,
 are any pending callbacks for this callback ID, they are removed from the queue
 before the new callback function is bound to the callback ID.
 
-Passing a null pointer callback function disables a previously set callback
-(besides flushing pending callbacks for this callback ID).
+A process can pass a null pointer as the callback argument to request the driver
+disable a previously set callback (besides flushing pending callbacks for this
+callback ID).
```
ppannuto, line 162:
```suggestion
A process can pass a null pointer as the callback argument to request the driver
disable a previously set callback. Note that it is possible for a pending callback
to be issued during the attempt to disable a callback. The callback is not disabled
until the call setting it to null returns.
```

doc/Syscalls.md
bradjc: I think it should be as strong as the subscribe case. That is, there is expected behavior that the appslice will not still be used, but if the system doesn't behave that way then the fault lies with the capsule and not with the kernel.

doc/Syscalls.md
jrvanwhy: I'm not sure that is entirely accurate. Does something prevent the following sequence of events? (I'm not sure, as I haven't touched this code much)

1. A process calls `subscribe` to set a callback.
1. The driver triggers the callback. The callback is enqueued in the process' callback queue.
1. The process passes a null pointer to `subscribe` to disable the callback, and the driver correctly stops queueing callbacks for that subscription ID.
1. The process calls `yield` and the previously-queued callback runs.

ppannuto: Mhm, unfortunately I think you are correct. That is currently possible.

Again, I think I projected here what I would _want_ the kernel to do rather than what it does. My concern with the current interface is that it becomes unreasonable to write predictable apps. I think I would like to see the kernel handle this as described, but for now, what you wrote is accurate.

jrvanwhy: +1 on "what I want the kernel to do != what the kernel does". I think that discussion is out of the scope of this PR, however, as changing how subscriptions work is not something that can be done quickly, and we should correct the documentation in the interim.

doc/Syscalls.md
jrvanwhy: I've observed some disagreement between contributors on what the contract is. I want to resolve that by making the documentation reflect the current implementation (note that this is an important part of the Tock threat model).

I prefer to have a strong revoke, but that's a chance that will require a significant amount of engineering effort to implement.

gendx:
Although this matches what the code actually does, to me this is a breaking change in the API. Instead of changing the documentation, shouldn't we fix the code and create a bug fix release (e.g. Tock 1.6.1) that actually implements the behavior that was documented so far?

doc/Syscalls.md
```diff
@@ -157,8 +157,9 @@ A callback function is uniquely identified by the pair (`driver`,
 are any pending callbacks for this callback ID, they are removed from the queue
 before the new callback function is bound to the callback ID.
 
-Passing a null pointer callback function disables a previously set callback
-(besides flushing pending callbacks for this callback ID).
+A process can pass a null pointer as the callback argument to request the driver
```
gendx, line 160: I think this is much less clear than the previous documentation. It doesn't specify at all what happens with this "request" to disable the callback. When does the driver take it into account? Can a driver reject such a request?

Without any such guarantees that the callback is indeed disabled once the syscall returns, how can one program anything useful in userspace?

PR #1837, opened 2020-05-08
Frame buffer HIL and ST7735 screen support

alexandruradovici:
### Pull Request Overview

This pull request adds:
  1. A HIL proposal for frame buffer support. This will allow applications to use graphical LCD and EPD screens. The longer purpose is to enable littlevgl.
  2. The frame buffer capsule to be used by the user applications
  3. A capsule to support the ST7735 SPI screen.

The frame buffer implementation allows zero copy mechanism for writing data from the user application to the screen memory. This is done by using a callback function to fill a screen buffer from an AppSlice.

```rust
pub trait ScreenClient {
    // this function is called by the screen whenever it has to send data
    // it returns the number of bytes placed in the buffer or 
    // 0 if there are no more bytes to write
    fn fill_next_buffer_for_write(&self, buffer: &'a mut [u8]) -> usize;
    fn command_complete(&self, r: ReturnCode);
}
```

In the screen implementation, this function is used like this:
```rust
self.buffer.take().map_or_else(
    || panic!("st7735: send parameters has no buffer"),
    |buffer| {
        let len = client.fill_next_buffer_for_write(buffer);
        if len > 0 {
            self.status.set(Status::SendCommandSlice(len));
            self.dc.set();
            self.spi.read_write_bytes(buffer, None, len);
        } else {
            self.buffer.replace(buffer);
            self.do_next_op();
        }
    },
);
```

### Testing Strategy

This pull request was tested with a nucleo429zi and an Adafruit ST7735 screen.

### TODO or Help Wanted

This pull request still needs some feedback regarding the frame buffer API,. Any suggestion is welcome.

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.


kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn get_color_depth_modes(&self) -> usize;
+    fn get_color_depth_bits(&self, index: usize) -> usize;
+
+    fn write(&self, x: usize, y: usize, width: usize, height: usize) -> ReturnCode;
+
+    fn set_client(&self, client: Option<&'static dyn ScreenClient>);
+
+    fn init(&self) -> ReturnCode;
+    fn on(&self) -> ReturnCode;
+    fn off(&self) -> ReturnCode;
```
bradjc, line 48: `init`, `on`, and `off` are generally undesirable functions in HILs, as it is not typically clear who should call them and when. However, on and off probably make sense for a screen, but some clarification on what on or off means would be useful. Can a screen that is off be updated? What would happen for something like e-ink? Is there a low power implication?

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    fn invert_on(&self) -> ReturnCode;
+    fn invert_off(&self) -> ReturnCode;
```
bradjc, line 51: Is `invert` something that is generally supported by displays?

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,57 @@
...
+            ScreenRotation::Rotated180 => 2,
+            ScreenRotation::Rotated270 => 3,
+        }
+    }
+}
+
+pub trait Screen {
+    fn set_resolution(&self, width: usize, height: usize) -> ReturnCode;
+    fn set_color_depth(&self, depth: usize) -> ReturnCode;
+    fn set_rotation(&self, rotation: ScreenRotation) -> ReturnCode;
```
bradjc, line 30: I wonder if these would make it too difficult to implement this HIL, if the screen does not somehow support these operations in hardware. I would think we don't need to support changing the resolution, etc, but these could be in an "Advanced" HIL.

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn get_color_depth(&self) -> usize;
+    fn get_rotation(&self) -> ScreenRotation;
+
+    fn get_resolution_modes(&self) -> usize;
+    fn get_resolution_size(&self, index: usize) -> (usize, usize);
+
+    fn get_color_depth_modes(&self) -> usize;
+    fn get_color_depth_bits(&self, index: usize) -> usize;
+
+    fn write(&self, x: usize, y: usize, width: usize, height: usize) -> ReturnCode;
```
bradjc, line 42: How does this interface work?

kernel/src/hil/framebuffer.rs
alexandruradovici: On and off should switch off the screen/go to low power mode, while still allowing users to draw into it. For the ST7735 this is possible, I will also test it on some e-ink display. So for, I think this is still possible, as these displays usually have some SRAM attached to them.

About init, I was not sure where to place it. Should the frame buffer automatically init the screen before the app starts?

kernel/src/hil/framebuffer.rs
alexandruradovici: I am still waiting for an e-ink display, but I think so. I'll get back on that.

kernel/src/hil/framebuffer.rs
alexandruradovici: I was not sure how to handle this. First I also thought about two HILs, what is not clear to me is how these would interact with the frame buffer capsule? I think there might be two options:
  - have another advanced frame buffer capsule, actually a layer on top of frame buffer
  - add a default implementation for these functions, reporting SUCCESS if the resolution and color depth fits the one implemented by the display or ENOSUPORT otherwise.

I think ``set_rotation`` is available on most displays, as it actually changes only the way the bytes are written to the memory.

kernel/src/hil/framebuffer.rs
alexandruradovici: ``get_color_depth_modes`` - returns the count of the color depths supported
``get_color_depth_bits`` - can be called with index 0 .. count-1 and will return the value of each color depth mode
``write`` - sends a write command to write data in the selected memory window, it seems that several screens can do that. The screen will then call ``fill_next_buffer_for_write`` from the ``ScreenClient`` for subsequent writes. This function will fill the buffer and return the number of bytes written. If it returns 0, the write stops and the screen issues ``command_complete``. This was there is no need for triple buffering (an app buffer, a frame buffer buffer and a screen buffer), data is transfered from the app directly. This also allows writing a repeated pattern (for now only one color) without the app having to fill a buffer with the same pattern. It also allow the screen to have internal an arbitrary size buffer.

brghena: It would be great to put this documentation right in the HIL file.

alexandruradovici: I added it to the HIL file.

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,143 @@
+//! Interface for FrameBuffer
+use crate::returncode::ReturnCode;
+use enum_primitive::cast::FromPrimitive;
+use enum_primitive::enum_from_primitive;
+
+enum_from_primitive! {
+    #[derive(Copy, Clone, PartialEq)]
+    pub enum ScreenRotation {
```
ppannuto, line 8: Is there a reason to give these enums numeric values here? I feel like any specific numeric value is too likely to be specific to an implementation to be appropriate for a HIL.

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,143 @@
...
+enum_from_primitive! {
+    #[derive(Copy, Clone, PartialEq)]
+    #[allow(non_camel_case_types)]
+    pub enum ScreenColorDepth {
+        None = 0,
+        Mono = 1,
+        RGB_233 = 8,
+        RGB_565 = 16,
+        RGB_888 = 24,
+        ARGB_8888 = 32,
```
ppannuto, line 36: This also feels like it might be too device-specific for a generic framebuffer HIL (I assume there are other 8-bit encodings that are not `RBG_233`).

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+enum_from_primitive! {
+    #[derive(Copy, Clone, PartialEq)]
+    #[allow(non_camel_case_types)]
+    pub enum ScreenColorDepth {
+        None = 0,
+        Mono = 1,
+        RGB_233 = 8,
```
ppannuto, line 33:
```suggestion
        /// Pixels encoded as 2-bit red channel, 3-bit green channel, 3-bit blue channel.
        RGB_233,
```

Something explaining what each of these modes are is probably appropriate for a HIL

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,143 @@
...
+            ScreenColorDepth::RGB_565 => 16,
+            ScreenColorDepth::RGB_888 => 16,
+            ScreenColorDepth::ARGB_8888 => 32,
+        }
+    }
+}
+
+pub trait Screen {
+    /// Sets the screen resolution (in pixels). Returns ENOSUPPORT if the resolution is
+    /// not supported. The function should return SUCCESS for at least one resolution.
```
ppannuto, line 55: Is this function synchronous? The later ones specify explicitly that they are.

If it's async, this call should probably have no return value (as the `ENOSUPPORT` will come from the complete call)

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,143 @@
...
+    fn set_resolution(&self, width: usize, height: usize) -> ReturnCode;
+
+    /// Sets the color depth (in bits per pixel). Returns ENOSUPPORT if the color depth is
+    /// not supported. The function should return SUCCESS for at least one color depth.
```
ppannuto, line 59: sync?

kernel/src/hil/framebuffer.rs
ppannuto: We have the pattern of layered advanced HILs in Uart currently, though I don't think any implementation of this idiom has ever landed in upstream Tock. The idea is that a capsule can implement both the basic Uart HIL as well as an Advanced HIL, eg: https://github.com/lab11/signpost-software/blob/master/signpost/kernel/signpost_drivers/src/gps_console.rs#L39

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,143 @@
...
+    fn set_color_depth(&self, depth: ScreenColorDepth) -> ReturnCode;
+
+    /// Sets the rotation of the display.
+    /// note this can swap the width with height.
+    fn set_rotation(&self, rotation: ScreenRotation) -> ReturnCode;
+
+    /// Returns a tuple (width, height) with the current resolution (in pixels)
+    /// note that width and height may change due to rotation
+    /// This function is synchronous.
+    fn get_resolution(&self) -> (usize, usize);
```
ppannuto, line 69: Should `set_resolution ` be updated to accept a tuple? Bit weird that these two don't match.

kernel/src/hil/framebuffer.rs
alexandruradovici: I gave them numeric values as they need to be passed from the user space. Any suggestion is welcome.

kernel/src/hil/framebuffer.rs
alexandruradovici: Fixed

alexandruradovici:
> I'm lightly concerned about all of these interfaces being synchronous. I could imagine a display connected over I2C or similar bus with a display controller on the other end that could never implement all of these synchronously. Maybe the overhead of async is too much for the local screen case though -- we certainly have the pattern of a `sync` and `async` version already from the GPIO HILs for this reason (i.e. a GPIO extender must be async, but making GPIO async in the common case would be obnoxious)

All the set functions are asynchronous as they need to interact with the display. The get functions are not, as the driver will have to configure the display to a certain resolution and pixel format, and I thought that the driver can memorize these options. Is there any way that the display could have a resolution and pixel format unknown to the driver?

kernel/src/hil/framebuffer.rs
alexandruradovici: This is async, but if the driver has already a command on going, it will not be able to handle it. If this returns SUCCESS, the user has to wait for the callback, otherwise the command fail immediately. 

I added this to the documentation.

kernel/src/hil/framebuffer.rs
alexandruradovici: I detailed the functionality in the documentation.

kernel/src/hil/framebuffer.rs
alexandruradovici: I decided to remove ``init``, I think you are right, this is something that the driver should do.

kernel/src/hil/framebuffer.rs
alexandruradovici: I create a new enum ``ScreenPixelFormat`` that allows the definition of various pixel formats. I added some, others can add other formats. I also created an impl for the enum with one function that returns the number of bits per fixel.

kernel/src/hil/framebuffer.rs
alexandruradovici: I changed the enum to ``ScreenPixelFormat`` so that it allows the definition of various pixel formats decoupled from the number of bits. I added some, others can add other formats. I also created an impl for the enum with one function that returns the number of bits per pixel.

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// Sets the rotation of the display.
+    /// The function should return SUCCESS if the request is registered
+    /// and will be sent to the screen.
+    /// Upon SUCCESS, the caller has to wait for the `command_complete` callback function
+    /// that will return the actual ReturnCode after setting the rotation.
+    ///
+    /// note this can swap the width with height.
+    fn set_rotation(&self, rotation: ScreenRotation) -> ReturnCode;
+
+    /// Returns a tuple (width, height) with the current resolution (in pixels)
```
bradjc, line 97:
```suggestion
    /// Returns a tuple (width, height) with the current resolution (in pixels).
```

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+}
+
+enum_from_primitive! {
+    #[derive(Copy, Clone, PartialEq)]
+    #[allow(non_camel_case_types)]
+    pub enum ScreenPixelFormat {
+        /// No pixel format
+        None = 0,
```
bradjc, line 32: What would this be?

kernel/src/hil/framebuffer.rs
bradjc: The SAM4L implements the advanced UART, and the tock bootloader uses it.

kernel/src/hil/framebuffer.rs
alexandruradovici: I's a marker for an invalid pixel format. It is useful when a call to ``get_supported_pixel_format`` with an index out of bounds. It could change it and return an Option, maybe that would be a better way.

alexandruradovici:
> This seems like a great candidate to be split into basic and advanced HILs. The basic HIL would just support the get_ commands and write, and the current HIL would be come the advanced version. The impacts on the other code in this PR would be minimal: there would just be a little renaming. I think the split would be a little nicer than just having an implementation for a simple display return errors for most functions, and making the upper layers handle that.
> 

I'll change it this way, it makes sense.

> I think some more clarity around how pixel values should be packed is needed. I'm not sure if this should use a type of some sort, or more comments, or just a comment saying it is up to specific implementations.
> 
These seems to be pretty standard ways of encoding the pixels. I added comments to each of them and most of the screen / ui libraries support them (ex: lvgl).

> The other bit is the `.write()` structure, which is an inverse to how all other operations in Tock work. Namely: the buffer is owned by the lower layer, and is passed up in a callback when requested, filled by the upper layer, and then used when the callback returns.
> 
> I'm actually not really sure why this works. How does the lower layer know it will have access to the buffer once `fill_next_buffer_for_write()` returns? What if the upper layer needs to wait for more data from an app, and wants to save the reference to the buffer?
> 
The lifetime for the buffer is related to the function's scope, not to the trait scope. The function should not be able to keep the buffer. I may be wrong as I do not have a lot of experience with reference lifetimes.

```rust
fn fill_next_buffer_for_write(&self, buffer: &'b mut [u8]) -> usize;
```

> In any case, it would be helpful to articulate and document why this doesn't follow the normal pattern of buffers being owned by the upper layers and passed to the lower layers. I'm not sure I see why the buffer couldn't be passed in with `write()`, and returned in the callback.
> 
> Am I correct in assuming that display hardware often keeps track of where the user is writing to, and just expects more data to be fed until all pixels in that area have been set? If so, I think it makes sense to capture that in the HIL. Adding a `continue_write(self, buf)` function would allow for this.

I chose this implementation in order to minimize the number of buffer copies required. Screen buffers can be large, so I wanted a way to copy data directly from the AppSlice to the screen. I haven't found a way to pass the AppSlice down (probably there is no safe way).

The screen driver has a buffer for sending data over to the screen (in this example an SPI buffer). The buffer size varies depending on the screen's implementation and bus type. As the frame buffer has no knowledge about this, I thought that the best way would be for the screen to ask for an amount of data from the frame buffer before performing any write.

If the buffer is passed down, I still need a small buffer for other command writes inside the screen and the buffer passed down from the frame buffer, in other words both capsules (framebuffer and screen) need a static buffer.

I think I could pass the buffer down using ``continue_write`` ... let me try an implementation like that.

kernel/src/hil/framebuffer.rs
alexandruradovici: I added a ScreenSetup trait and an Option for it to the framebuffer capsule.

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    /// Returns the number of the resolutions supported.
+    /// should return at least one (the current resolution)
+    /// This function is synchronous as the driver should know this value without
+    /// requesting it from the screen (most screens do not support such a request,
+    /// resolutions are described in the data sheet).
+    ///
+    /// If the screen supports such a feature, the driver should request this information
+    /// from the screen upfront.
+    fn get_supported_resolutions(&self) -> usize;
```
alistair23, line 80: Shouldn't this be something like: `get_num_supported_resolutions()`?

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Returns the number of the pixel formats supported.
+    /// This function is synchronous as the driver should know this value without
+    /// requesting it from the screen (most screens do not support such a request,
+    /// pixel formats are described in the data sheet).
+    ///
+    /// If the screen supports such a feature, the driver should request this information
+    /// from the screen upfront.
+    fn get_supported_pixel_formats(&self) -> usize;
+
+    /// Can be called with index 0 .. count-1 and will returns
```
alistair23, line 98: will return

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// an internal arbitrary size buffer.
+    fn write(&self, x: usize, y: usize, width: usize, height: usize) -> ReturnCode;
+
+    fn set_client(&self, client: Option<&'static dyn ScreenClient>);
+
+    /// Powers up the display.
+    fn on(&self) -> ReturnCode;
+
+    /// Powers down the display. The screen should be able to accept data even when the display is off.
+    fn off(&self) -> ReturnCode;
```
alistair23, line 141: What about setting a brightness level instead? Something between 0 and 1 maybe, where 0 is off?

alexandruradovici: Most of the embedded screens have the brightness set as a background LED. I could maybe add that function in the ScreenSetup (advanced) HIL?

kernel/src/hil/framebuffer.rs
alexandruradovici: Fixed

kernel/src/hil/framebuffer.rs
alexandruradovici: That makes more sense, I renamed it.

kernel/src/hil/framebuffer.rs
alistair23: If the hardware only supports on/off then the brightness could maybe just be rounded up/down. I think for hardware that supports multiple brightness levels (if it exists) if would be useful to set a brightness.

kernel/src/hil/framebuffer.rs
alexandruradovici: I changed the way the buffer is used, now the buffer is passed down from the frame buffer driver to the screen driver.

kernel/src/hil/framebuffer.rs
alexandruradovici: Done, thanks for the suggestions. I used a value in percent, as sending float values is not trivial.

alexandruradovici: I made the following changes to the HIL so that the buffer is sent from the frame buffer driver towards the lower level screen driver:

```rust
/// Sets the video memory frame.
fn set_write_frame(&self, x: usize, y: usize, width: usize, height: usize) -> ReturnCode;

/// Sends a write command to write data in the selected video memory frame.
fn write(&self, buffer: &'static mut [u8], len: usize) -> ReturnCode;
```

I added another function to the ScreenClient to signal when a write command is complete.

```rust
/// The screen will call this function to notify that a command (except write) has finished.
fn command_complete(&self, r: ReturnCode);

/// The screen will call this function to notify that the write command has finished.
/// This is different from `command_complete` as it has to pass back the write buffer
fn write_complete(&self, buffer: &'static mut [u8], r: ReturnCode);
```
Any feedback is very useful.

alexandruradovici:
I made the following changes to the HIL so that the buffer is sent from the frame buffer driver towards the lower level screen driver:

```rust
/// Sets the video memory frame.
fn set_write_frame(&self, x: usize, y: usize, width: usize, height: usize) -> ReturnCode;

/// Sends a write command to write data in the selected video memory frame.
fn write(&self, buffer: &'static mut [u8], len: usize) -> ReturnCode;
```

I added another function to the ScreenClient to signal when a write command is complete.

```rust
/// The screen will call this function to notify that a command (except write) has finished.
fn command_complete(&self, r: ReturnCode);

/// The screen will call this function to notify that the write command has finished.
/// This is different from `command_complete` as it has to pass back the write buffer
fn write_complete(&self, buffer: &'static mut [u8], r: ReturnCode);
```
Any feedback is very useful.

alexandruradovici:
Any ideas why the check fails? It seems some timeout on opentitan, but I see no error.

hudson-ayers:
I re-ran it and it passed -- it seems the QEMU test is not completely deterministic at this point (cc @alistair23 ).

alexandruradovici:
> I re-ran it and it passed -- it seems the QEMU test is not completely deterministic at this point (cc @alistair23 ).

Thank you

alistair23:
I can't see what failed, but if it was Travis then that is probably related to cache.

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,164 @@
...
+    fn get_pixel_format(&self) -> ScreenPixelFormat;
+
+    /// Returns the current rotation.
+    /// This function is synchronous as the driver should know this value without
+    /// requesting it from the screen.
+    fn get_rotation(&self) -> ScreenRotation;
+
+    /// Sets the video memory frame.
+    /// This function has to be called before the first call to the write function.
+    fn set_write_frame(&self, x: usize, y: usize, width: usize, height: usize) -> ReturnCode;
```
bradjc, line 127: Is this synchronous?

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    /// Sends a write command to write data in the selected video memory frame.
+    fn write(&self, buffer: &'static mut [u8], len: usize) -> ReturnCode;
```
bradjc, line 130: What are the return values?

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,164 @@
+//! Interface for FrameBuffer
```
bradjc, line 1: Rename to screen.rs.

kernel/src/hil/framebuffer.rs
bradjc: I don't think we want to establish a precedent that a HIL also defines a syscall interface. This mapping can be done in the capsule, much like how GPIO works (with pull up/down and rising/falling edge). This also removes the need to add a dependency to the kernel crate.

kernel/src/hil/framebuffer.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    /// Returns the number of the resolutions supported.
+    /// should return at least one (the current resolution)
+    /// This function is synchronous as the driver should know this value without
+    /// requesting it from the screen (most screens do not support such a request,
+    /// resolutions are described in the data sheet).
+    ///
+    /// If the screen supports such a feature, the driver should request this information
+    /// from the screen upfront.
+    fn get_num_supported_resolutions(&self) -> usize;
```
bradjc, line 82: I'm not sure I understand how this would be used. I imagine it would be easy to implement, but what can clients expect to do with the number of supported resolutions?

kernel/src/hil/framebuffer.rs
alexandruradovici: Got it, solved.

kernel/src/hil/framebuffer.rs
alexandruradovici: No

kernel/src/hil/framebuffer.rs
alexandruradovici: They signal whether the command can be executed. They can return either SUCCESS or EBUSY (if another command is in progress).

kernel/src/hil/framebuffer.rs
alexandruradovici: ok

kernel/src/hil/framebuffer.rs
alexandruradovici: Some screens like ST7735 seem to support several resolution, I thought the user might want to select one.

kernel/src/hil/screen.rs
```diff
@@ -0,0 +1,156 @@
...
+    fn get_pixel_format(&self) -> ScreenPixelFormat;
+
+    /// Returns the current rotation.
+    /// This function is synchronous as the driver should know this value without
+    /// requesting it from the screen.
+    fn get_rotation(&self) -> ScreenRotation;
+
+    /// Sets the video memory frame.
+    /// This function has to be called before the first call to the write function.
+    fn set_write_frame(&self, x: usize, y: usize, width: usize, height: usize) -> ReturnCode;
```
bradjc, line 119:
```suggestion
    /// Sets the video memory frame.
    /// This function has to be called before the first call to the write function.
    /// This will generate a `command_complete()` callback when finished.
    ///
    /// Return values:
    /// - `SUCCESS`: The write frame is valid.
    /// - `EINVAL`: The parameters of the write frame are not valid.
    /// - `EBUSY`: Unable to set the write frame on the device.
    fn set_write_frame(&self, x: usize, y: usize, width: usize, height: usize) -> ReturnCode;
```

kernel/src/hil/screen.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    /// Sends a write command to write data in the selected video memory frame.
+    fn write(&self, buffer: &'static mut [u8], len: usize) -> ReturnCode;
```
bradjc, line 122:
```suggestion
    /// Sends a write command to write data in the selected video memory frame.
    /// When finished, the driver will call the `write_complete()` callback.
    ///
    /// Return values:
    /// - `SUCCESS`: Write is valid and will be sent to the screen.
    /// - `EINVAL`: Write is invalid or length is wrong.
    /// - `EBUSY`: Another write is in progress.
    fn write(&self, buffer: &'static mut [u8], len: usize) -> ReturnCode;
```

kernel/Cargo.toml
```diff
@@ -7,3 +7,4 @@ edition = "2018"
 [dependencies]
 tock-registers = { path = "../libraries/tock-register-interface" }
 tock-cells = { path = "../libraries/tock-cells" }
+enum_primitive = { path = "../libraries/enum_primitive" }
```
bradjc, line 10: Remove?

kernel/src/hil/screen.rs
```diff
@@ -0,0 +1,156 @@
+//! Interface for FrameBuffer
```
bradjc, line 1:
```suggestion
//! Interface for screens and displays.
```

kernel/Cargo.toml
alexandruradovici: Solved

capsules/src/screen.rs
```diff
@@ -0,0 +1,590 @@
+//! Provides userspace with access to the screen.
+//!
+//! Usage
+//! -----
+//!
+//! You need a screen that provides the `hil::screen::Screen` trait.
+//!
+//! ```rust
+//!
```
ppannuto, line 9:
```suggestion
```

capsules/src/screen.rs
```diff
@@ -0,0 +1,590 @@
...
+            screen_setup: screen_setup,
+            apps: grant,
+            current_app: OptionalCell::empty(),
+            screen_ready: Cell::new(false),
+            pixel_format: Cell::new(screen.get_pixel_format()),
+            buffer: TakeCell::new(buffer),
+        }
+    }
+
+    // Check so see if we are doing something. If not,
```
ppannuto, line 139:
```suggestion
    // Check to see if we are doing something. If not,
```

capsules/src/screen.rs
```diff
@@ -0,0 +1,590 @@
...
+                })
+                .unwrap_or_else(|err| err.into()),
+            _ => ReturnCode::ENOSUPPORT,
+        }
+    }
+
+    fn command(&self, command_num: usize, data1: usize, data2: usize, appid: AppId) -> ReturnCode {
+        match command_num {
+            0 =>
+            /* This driver exists. */
```
ppannuto, line 512:
```suggestion
            // This driver exists.
```

Match the others

kernel/src/hil/screen.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// that will return the actual ReturnCode after setting the pixel format.
+    fn set_pixel_format(&self, depth: ScreenPixelFormat) -> ReturnCode;
+
+    /// Sets the rotation of the display.
+    /// The function should return SUCCESS if the request is registered
+    /// and will be sent to the screen.
+    /// Upon SUCCESS, the caller has to wait for the `command_complete` callback function
+    /// that will return the actual ReturnCode after setting the rotation.
+    ///
+    /// note this can swap the width with height.
```
ppannuto, line 63:
```suggestion
    /// Note that in the case of `Rotated90` or `Rotated270`, this will swap the width and height.
```

Can/should this say *will* like this? Is it a guarantee? Or a thing that some screen might do / expect?

kernel/src/hil/screen.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// resolutions are described in the data sheet).
+    ///
+    /// If the screen supports such a feature, the driver should request this information
+    /// from the screen upfront.
+    fn get_num_supported_resolutions(&self) -> usize;
+
+    /// Can be called with an index from 0 .. count-1 and will
+    /// a tuple (width, height) with the current resolution (in pixels).
+    /// note that width and height may change due to rotation
+    /// This function is synchronous as the driver should know this value without
```
ppannuto, line 79:
```suggestion
    /// note that width and height may change due to rotation
    ///
    /// This function is synchronous as the driver should know this value without
```

kernel/src/hil/screen.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// pixel formats are described in the data sheet).
+    ///
+    /// If the screen supports such a feature, the driver should request this information
+    /// from the screen upfront.
+    fn get_num_supported_pixel_formats(&self) -> usize;
+
+    /// Can be called with index 0 .. count-1 and will return
+    /// the value of each pixel format mode.
+    /// This function is synchronous as the driver should know this value without
+    /// requesting it from the screen.
```
ppannuto, line 95:
```suggestion
    /// the value of each pixel format mode.
    ///
    /// This function is synchronous as the driver should know this value without
    /// requesting it from the screen.
```

kernel/src/hil/screen.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// Sends a write command to write data in the selected video memory frame.
+    /// When finished, the driver will call the `write_complete()` callback.
+    ///
+    /// Return values:
+    /// - `SUCCESS`: Write is valid and will be sent to the screen.
+    /// - `EINVAL`: Write is invalid or length is wrong.
+    /// - `EBUSY`: Another write is in progress.
+    fn write(&self, buffer: &'static mut [u8], len: usize) -> ReturnCode;
+
+    fn set_client(&self, client: Option<&'static dyn ScreenClient>);
```
ppannuto, line 136:
```suggestion

    /// Set the object to receive the asynchronous command callbacks.
    fn set_client(&self, client: Option<&'static dyn ScreenClient>);
```

capsules/src/st7735.rs
```diff
@@ -0,0 +1,1040 @@
+//! ST7735 SPI Screen
+//!
+//! Usage
+//! -----
+//!
+//! ```rust
+//!
```
ppannuto, line 7:
```suggestion
```

bradjc:
Why does this delete an I2C driver?

alexandruradovici:
> Why does this delete an I2C driver?

The file was not supposed to be there, the I2C is in stm32f303xc, the deleted file was left there by mistake.

PR #1841, opened 2020-05-08
kernel: move restart logic to function

bradjc:
### Pull Request Overview

This pull request moves the process restart logic to its own function. This is a bit cleaner, but really it structures things more effectively for adding an `exit` syscall that can also be used to ask the kernel to restart the process.

This is just a small refactor and doesn't change any current behavior.

I also updated some comments to make them more clear.

### Testing Strategy

This pull request was tested by running the MPU fault test app and verifying that the app restarts several times on hail before the kernel panics.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -861,109 +861,8 @@ impl<C: Chip> ProcessType for Process<'a, C> {
                 // process faulted. Panic and print status
                 panic!("Process {} had a fault", self.process_name);
             }
-            FaultResponse::Restart(restart_policy) => {
-                // Start with the generic terminate operations. This frees state
-                // for this process and removes any pending tasks from the
-                // scheduler's queue.
-                self.terminate();
-
-                // Decide what to do with this process. Should it be restarted?
-                // Or should we leave it in a stopped & faulted state? If the
-                // process is faulting too often we might not want to restart.
-                // If we are not going to restart the process then we can just
-                // leave it in the stopped faulted state by returning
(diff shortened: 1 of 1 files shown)


kernel/src/process.rs
```diff
@@ -1857,6 +1756,144 @@ impl<C: 'static + Chip> Process<'a, C> {
...
+    ///    kernel is using for restarting a specific process. For example, if a
+    ///    process has restarted a number of times in a row the kernel may
+    ///    decide to stop executing it.
+    ///
+    /// 2. Some state can no long be configured for the process. For example,
+    ///    the syscall state for the process fails to initialize.
+    ///
+    /// After `restart()` runs the process will either be queued to run its
+    /// `_start` function, or it will be left in `failure_state`.
+    fn restart(&self, failure_state: State) {
```
ppannuto, line 1777: Since restarting can fail, should the function return some indication of success or failure?

bradjc: I'm not sure. I have no idea what the caller could do with that information, but it might be useful to know what happened and potentially report it back for informative purposes.

But, I think it's better if this PR doesn't change any behavior, just re-organizes code. Actual changes can be part of a different patch.

ppannuto: I think if these was a `pub` I'd push back, but given that it's contained, this is fine for now.

hudson-ayers:
bors retry

Issue #1465, opened 2019-11-15
Tracking: Revamp TBF Parsing


bradjc:
- [x] Break parsing function `parse_and_validate_tbf_header()` into multiple smaller functions that take slices to what they are working on. (#1480)
  ~~- [ ] Make the actually `unsafe` operations clearly marked. However, the base function is still unsafe.~~
- [x] Add a TLV for specifying the memory address the process _must_ be assigned. This is needed for the N x K app linking alternative to PIC.
- [x] Move the TBF code to its own crate in `/libraries`.

bradjc:
I'm working on an extension to the TBF format so that elf2tab can specify fixed addresses for processes, and the kernel can verify the resulting address. This won't functionally change much, but will help with debugging and error reporting, since right now address mismatch issues lead to difficult to catch errors.

Issue #1283, opened 2019-02-23
Tracking: Support multiple static apps in Tock


rajivr:
- [x] Add support in the TBF header for apps to specify where their RAM region must begin.
- [x] Add support in elf2tab for adding the TBF headers. This is merged, but not released quite yet (June 2020).
- [x] Add support for the kernel checking that addresses are met.
- [ ] Updated loading processes so that the kernel uses the correct memory locations. (#1928)

@jrvanwhy recently added statically linked app support to `libtock-rs`.

The limitation right now is that we do not have a good way to support multiple statically linked apps. I was wondering if there was interest upstream for supporting multiple statically linked apps?

My motivation for being in favor of supporting multiple statically linked apps is as follows.

1. Dynamic Apps requires `ROPI-RWPI` which is not fully supported by Rust for ARM M profile targets

2. Even if Rust on ARM M profile targets eventually gets `ROPI-RWPI` support, there is no guarantee that other Rust embedded targets would have a position independent model similar to `ROPI-RWPI` that would be suitable for Tock Apps

One argument against adding support for statically linked app could be,

- In Tock, it is currently possible to update Tock Apps independently, as long as syscall ABI is the same. If upstream considers this a critical feature of Tock, then statically linked Apps will not allow for this possibility 

While the following list is not meant to be exhaustive, I want to capture the potential changes that we might need in order to properly support multiple statically linked Apps.

#### TBF Changes (might require a version bump)

1. Introduce a field to specify if an app is static or a dynamic app

2. For a static app, capture `app_flash_start`, in order to indicate where in flash the TBF file would be placed.

3. For a static app, capture `app_memory_start` in order to indicate where in memory we expect App's SRAM region to start.

#### Kernel MPU changes

1. Introduce `allocate_region_static` with a signature similar to

```
    fn allocate_region_static(
        &self,
        unallocated_memory_start: *const u8,
        unallocated_memory_size: usize,
        app_flash_start: usize,
        app_flash_size: usize,
        permissions: Permissions,
        config: &mut Self::MpuConfig,
    ) -> Option<Region> { ... }
```

where `app_flash_size` would be same as existing `app_flash_size` in `process.rs` which is derived from `tbf_header.get_total_size()`.

2. Introduce `allocate_app_memory_region_static` with a signature similar to

```
    fn allocate_app_memory_region_static(
        &self,
        unallocated_memory_start: *const u8,
        unallocated_memory_size: usize,
        app_ram_start: usize,
[...]

PR #1845, opened 2020-05-12
TBF: Add "Fixed Addresses" TLV

bradjc:
### Pull Request Overview

Part of: #1283 and #1465 

This adds a new TLV header (`FixedAddresses`) for processes to specify the addresses in flash and ram they need (if they require fixed addresses, i.e. are not position independent. The kernel then checks these addresses when creating processes, and throws errors if they are not met. Note: this does not change the kernel to try to give a process its requested memory address.

The main benefit of this is that non-PIC apps can use this header and the kernel can catch errors when loading processes. Right now misalignments cause other errors to pop up once a process is running, and these are difficult to debug.

I have related changes to elf2tab and tockloader in progress, but will wait to PR those until any changes are ironed out here.

### Testing Strategy

Adding the headers to Hail and verifying that apps work when they match and an error is displayed when they do not.

### TODO or Help Wanted

- Is this the right TLV header structure to use?
- Does this break any 1.0 promises?
  - I don't think so, as it doesn't change any existing headers and any existing apps won't change, so old apps will run just fine. In particular, ARM libtock-c apps won't change at all.

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.

### DIFF
--- a/kernel/src/tbfheader.rs
+++ b/kernel/src/tbfheader.rs
@@ -147,9 +147,8 @@ crate struct TbfHeaderV2WriteableFlashRegion {
 #[derive(Clone, Copy, Debug, Default)]
 crate struct TbfHeaderV2FixedAddresses {
     /// The absolute address of the start of RAM that the process expects. For
-    /// example, if the process was linked with a RAM region of length 4 kB
-    /// starting at address `0x00023000`, then this would be set to
-    /// `0x00023000`.
+    /// example, if the process was linked with a RAM region starting at 
+    /// address `0x00023000`, then this would be set to `0x00023000`.
     start_process_ram: u32,
     /// The absolute address of the start of the process binary. This does _not_
     /// include the TBF header. This is the address the process used for the
--- a/kernel/src/tbfheader.rs
+++ b/kernel/src/tbfheader.rs
@@ -96,6 +96,7 @@ crate enum TbfHeaderTypes {
     TbfHeaderMain = 1,
     TbfHeaderWriteableFlashRegions = 2,
     TbfHeaderPackageName = 3,
+    TbfHeaderFixedAddresses = 5,
 
     /// Some field in the header that we do not understand. Since the TLV format
     /// specifies the length of each section, if we get a field we do not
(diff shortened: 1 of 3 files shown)


kernel/src/tbfheader.rs
```diff
@@ -351,6 +401,36 @@ impl TbfHeader {
...
+        match *self {
+            TbfHeader::TbfHeaderV2(hd) => hd.fixed_addresses.map_or(None, |fa| {
+                if fa.start_process_ram == 0xFFFFFFFF {
+                    None
+                } else {
+                    Some(fa.start_process_ram)
+                }
+            }),
+            _ => None,
+        }
```
jrvanwhy, line 417: This implementation feels a bit awkward to me. How about the following?

```
        let hd = match self {
            TbfHeader::TbfHeaderV2(hd) => hd,
            _ => return None,
        };
        match hd.fixed_addresses.as_ref()?.start_process_ram {
            0xFFFFFFFF => None,
            start => Some(start),
        }
```

kernel/src/tbfheader.rs
```diff
@@ -531,6 +612,15 @@ crate fn parse_tbf_header(header: &'static [u8], version: u16) -> Result<TbfHead
                                 .or(Err(TbfParseError::BadProcessName))?;
                         }
 
+                        TbfHeaderTypes::TbfHeaderFixedAddresses => {
+                            let entry_len = mem::size_of::<TbfHeaderV2FixedAddresses>();
```
jrvanwhy, line 616: TbfHeaderV2FixedAddresses needs to be `repr(C)` for this check to make sense. It probably makes more sense to hard-code 8 here.

kernel/src/tbfheader.rs
bradjc: I will take your word for it!

bradjc:
Updated.

kernel/src/tbfheader.rs
```diff
@@ -128,6 +132,31 @@ crate struct TbfHeaderV2WriteableFlashRegion {
...
+///
+/// If this header is included, the kernel will check these values when setting
+/// up the process. If a process wants to set one fixed address but not the other, the unused one
+/// can be set to 0xFFFFFFFF.
+#[derive(Clone, Copy, Debug, Default)]
+crate struct TbfHeaderV2FixedAddresses {
+    /// The absolute address of the start of RAM that the process expects. For
+    /// example, if the process was linked with a RAM region of length 4 kB
+    /// starting at address `0x00023000`, then this would be set to
+    /// `0x00023000`.
```
ppannuto, line 152:
```suggestion
    /// example, if the process was linked with a RAM region starting at 
    /// address `0x00023000`, then this would be set to `0x00023000`.
```
What does length have to do with anything here?

bradjc: Nothing, which was my point, but yeah probably makes it more confusing.

bradjc:
Any other comments @tock/core-wg?

PR #1850, opened 2020-05-13
Add licensing information to CONTRIBUTING

ppannuto:
### Pull Request Overview

This updates Tock policy to require license headers at the top of files.

This is conforming to a norm that we probably should have done all along (however much personal distaste I may have for the need).

### Testing Strategy

N/A.

### TODO or Help Wanted

(1) Agree on the text here.
--- [propose merging this PR at this step; then bulk change next two as a big PR] ---
(2) Update extant files to include header.
(3) Add CI to enforce this going forward.

### Documentation Updated

### Formatting

- [ ] Ran `make formatall`.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -212,7 +212,20 @@ invoke `make format`.
 The target `make formatall` in the root will automatically run all style checks
 and make any required changes. PRs must pass the formatting checks before landing.
 
-### Step 8: Landing
+### Step 8: Licensing
+
+Tock is dual-licensed under Apache2/MIT. Files in Tock require a short header
+indicating this. In 99% of cases, the following text should be at the top of every file:
+
+> This file is part of the Tock Project.
+>
+> Dual licensed under the Apache License, Version 2.0 or the MIT License, at your option.
+> See the License for the specific language governing permissions and limitations under the License.
+> This file may not be copied, modified, or distributed except according to those terms.
+
+This header should be automatically enforced by CI tooling.
+
+### Step 9: Landing
 
 In order to land, a Pull Request needs to be reviewed and
 [approved](#getting-approvals-for-your-pull-request) by at least one person with
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -1,10 +1,18 @@
-Tock Operating System
-Copyright 2016 The Tock OS Developers
+Short version for non-lawyers:
 
-Licensed under the Apache License, Version 2.0
-<LICENSE-APACHE or
-http://www.apache.org/licenses/LICENSE-2.0> or the MIT
-license <LICENSE-MIT or http://opensource.org/licenses/MIT>,
-at your option. All files in this project may not be copied,
-modified, or distributed except according to those terms.
+The Tock Project is dual-licensed under Apache 2.0 and MIT terms.
+
+
+Longer version:
+
+Copyrights in the Tock project are retained by their contributors. No
+copyright assignment is required to contribute to the Tock project.
+
+Some files include explicit copyright notices and/or license notices.
+For full authorship information, see the version control history.
+
+Except as otherwise noted (below and/or in individual files), Tock is
+licensed under the Apache License, Version 2.0 <LICENSE-APACHE> or
+<http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+<LICENSE-MIT> or <http://opensource.org/licenses/MIT>, at your option.
 


.github/CONTRIBUTING.md
```diff
@@ -212,7 +212,21 @@ invoke `make format`.
...
+
+> Copyright The Tock Project Authors.
+>
+> Dual licensed under the Apache License, Version 2.0 or the MIT License, at your option.
+> See the License for the specific language governing permissions and limitations under the License.
+> This file may not be copied, modified, or distributed except according to those terms.
+
+This should be automatically enforced by CI tooling. If you need different licensure
+for some reason, please get in touch with the core team directly.
+
```
rajivr, line 228: IANAL, but I don't think the way Tock contributions are setup right now, it would be possible to re-licence Tock in a different licence unless all contributors agree. Re-licensing would typically be possible when there has been copyright assignment. I don't think neither ASL nor MIT requires copyright assignment.

ppannuto: Indeed. I rephrased that slightly. The goal is basically to try to be friendly still with this text :)

brghena:
What led to this PR? Is not having individual file licenses problematic for one of our collaborators?

jrvanwhy:
> What led to this PR? Is not having individual file licenses problematic for one of our collaborators?

The lack of license headers on individual files is problematic for the OpenTitan project.

brghena:
Okay, then we should definitely add them (despite my shared distaste).

The text looks reasonable enough to me, but we should make sure that someone authoritative from the OpenTitan team agrees that it would meet their needs.

jrvanwhy:
> The text looks reasonable enough to me, but we should make sure that someone authoritative from the OpenTitan team agrees that it would meet their needs.

I can't make any promises here. Once a consensus on wording is reached here, I will ask around and see if I can get that approval.

.github/CONTRIBUTING.md
gendx: IAMAL, but I'm also wondering what `we will see if we can work out a solution` would entail. I see the following possible outcomes:
1. After all, the contributor agrees to Apache License, Version 2.0 + MIT License. Easy case.
1. Another license is proposed. Accepting the contribution would likely require re-licensing the whole of Tock, which would mean either of:
  a. Agreement from all the existing [contributors](https://github.com/tock/tock/graphs/contributors) (currently 94), as well as their (former) employers in case such employers hold copyright, their heirs in case of a tragic event, etc. This is basically intractable.
  b. Removing commits from existing contributors who don't agree on the re-licensing, or don't reply in time. Depending on the scale of the contribution, this can affect core functionality of Tock.
1. The contributor creates a fork of the project, with their contributions in this fork licensed however they like as long as it complies with either of Apache License, Version 2.0 or MIT License, at their option. But forking isn't really "contributing" in the scope of this page.

All in all, (1) is a non-event (apart from interesting discussions about licensing & copyright), (2) is a tremendous amount of work (even if the contribution is good, this is likely not worth it) and (3) is not "contributing" per-se.

Also, I'm not sure there will even be such a case of a contributor coming to Tock but requesting another license. Apache + MIT are a common setup for open-source nowadays, and anyone seeing the project is aware of the license before they decide to contribute or not.

ppannuto: Okay then, I just removed that sentence.

ppannuto:
Question:

>  Copyright The Tock Project Authors. 

Should this have an escape clause allowing other copyrights, i.e. if a corporate contribution requires that their company maintain the copyright, I don't _think_ we care?

jrvanwhy:
> Should this have an escape clause allowing other copyrights, i.e. if a corporate contribution requires that their company maintain the copyright, I don't _think_ we care?

In principle, I agree. However, I am unsure how to handle the situation where we accept such a contribution, then make maintenance changes to the file ourselves.

For reference, here is how Google handles that situation for its own open source projects: https://opensource.google/docs/releasing/authors/. Taking that approach is another possibility.

.github/CONTRIBUTING.md
```diff
@@ -212,7 +212,20 @@ invoke `make format`.
 The target `make formatall` in the root will automatically run all style checks
 and make any required changes. PRs must pass the formatting checks before landing.
 
-### Step 8: Landing
+### Step 8: Licensing
+
+Tock is dual-licensed under Apache2/MIT. Files in Tock require a short header
+indicating this. In 99% of cases, the following text should be at the top of every file:
+
+> Copyright The Tock Project Authors.
```
jrvanwhy, line 220: Do we want to allow an optional year here? i.e.
```
Copyright [optional YYYY] The Tock Project Authors
```

Then we could leave off the year for existing files (which is difficult to determine in bulk), and new files can have accurate years.

brghena: The same maintenance question applies. Is that the year the file was first committed or the year of the last update to it?

My real opinion though remains: if this is necessary for collaborators, then we should do it, otherwise keep the bare minimum.

ppannuto: Okay, so from my Google-supported* armchair lawyering: in the US we would likely be fine without a year [it would default to the time of publication; adding a year would allow for subsequent renewal if there were substantial updates, but we don't really care about that]. Once you start looking internationally, however, the year is sometimes still significant.

Relatedly, putting a year _too early_ can invalidate a copyright, but putting one later than publication is okay (with the understanding that _technically_ if someone could produce evidence of their own published copy of the work prior to the date you declared, they could then invalidate your copyright; not something I think we are concerned with. In our case, courtesy of GitHub we have strong evidence of the first publication date of all of our work, should we ever need to get in such a fight).

Bonus things learned, while simply the word "Copyright" is sufficient, it is strongly recommended to also include ©, as some courts [e.g. Canada apparently](https://www.copyrightlaws.com/copyright-symbol-notice-year/) will give extra weight to the ©.

I will update this PR to try to capture these things.

For the initial bulk update, I believe we should mark everything as © 2020.

*The colloquial me-as-person-who-searched-things support, not any formal support from Google the corporate entity

brghena: Did something suggest that we should be putting a copyright range? I don't know if I've ever seen "Copyright 2020-2021" or similar in a file before. It looks to me as though it would imply that there is not a copyright after that date...

ppannuto:
I think I like mirroring the AUTHORS approach. We already have a "Contributors" list at the bottom of the main README that is effectively this.

I've updated this with proposed changes to support this as well.

.github/CONTRIBUTING.md
```diff
@@ -212,7 +212,29 @@ invoke `make format`.
...
+> See the License for the specific language governing permissions and limitations under the License.
+> This file may not be copied, modified, or distributed except according to those terms.
+
+If you create a new "version or edition" of a given work, then the date range
+should be extended to include the newest work. In practice, this means any
+update to a source file will require updating the range (and some churn
+hopefully largely restricted to January each year).
+
+For simplicity, we ask that copyright should always be held by the Tock Project
+Authors. We are happy to recognize new contributors in the
```
jrvanwhy, line 232: "For simplicity, we ask that copyright should always be held by the Tock Project Authors."

This wording implies copyright assignment, which is not what we want here.

jrvanwhy:
What repositories should we make this change in? As far as I understand, OpenTitan uses or plans to use:

1. `elf2tab`
1. `libtock-c`
1. `libtock-rs`
1. `tock`

brghena:
Those seem necessary for sure. Additional repos may be:
 * [design-explorations](https://github.com/tock/design-explorations)
 * [book](https://github.com/tock/book)
 * [tockloader](https://github.com/tock/tockloader)

.github/CONTRIBUTING.md
gendx: > Did something suggest that we should be putting a copyright range? I don't know if I've ever seen "Copyright 2020-2021" or similar in a file before. It looks to me as though it would imply that there is not a copyright after that date...

FWIW I've seen such year ranges quite often. But IANAL in terms of what putting a year or not and which year to put implies in terms of copyright.

ppannuto: Re range: yes. I thought I explained that in the text I wrote:
```
If you create a new "version or edition" of a given work, then the date range
should be extended to include the newest work. In practice, this means any
update to a source file will require updating the range (and some churn
hopefully largely restricted to January each year).
```
Happy to take language updates to make that clearer.

.github/CONTRIBUTING.md
ppannuto: I just removed that sentence. I guess my question is will/should the CI look at who is listed as the author, or just ignore that field?

AUTHORS.md
```diff
@@ -0,0 +1,22 @@
+The Tock Project Developers include dozens of contributors, some on behalf of
```
bradjc, line 1:
```suggestion
# Authors

The Tock Project Developers include dozens of contributors, some on behalf of
```

AUTHORS.md
```diff
@@ -0,0 +1,23 @@
...
+  * Google
+  * Helium
+  * MIT
+  * MIT Lincoln Laboratory
+  * Princeton University
+  * Stanford University
+  * University of California, Berkeley
+  * University of California, San Diego
+  * University of Michigan
+  * University of Virginia
```
alistair23, line 20: Can we add WD? WD is a top 10 contributor now :)

.github/CONTRIBUTING.md
```diff
@@ -212,7 +212,27 @@ invoke `make format`.
 The target `make formatall` in the root will automatically run all style checks
 and make any required changes. PRs must pass the formatting checks before landing.
 
-### Step 8: Landing
+### Step 8: Licensing
+
+Tock is dual-licensed under Apache2/MIT. Files in Tock require a short header
+indicating this. In 99% of cases, the following text should be at the top of every file:
+
+> © Copyright 2020[-20XX] The Tock Project Authors.
```
alistair23, line 220: As I mentioned in the call and speaking as not a lawyer. I don't see how you can do this per file without a CLA. Other projects without CLAs (Linux, QEMU, etc.) have this Copyrighted to major contributors.

hudson-ayers:
The `COPYRIGHT` file currently begins: 
````
Tock Operating System
Copyright 2016 The Tock OS Developers
```

Should this be updated to `The Tock Project Authors` to match the text in the `AUTHORS` file? Should it be updated to point to the authors listed in that file?

.github/CONTRIBUTING.md
ppannuto: Agreed. Rust slowly purged copyright notices from all of their projects. I think Cargo's PR sums up the rationale the best: https://github.com/rust-lang/cargo/pull/4331

I've updated this to remove copyright language from the header template.

AUTHORS.md
ppannuto: Yes!! Done :)

ppannuto:
Okay, we need to disentangle two issues here:

---

The first is copyright, which I think we really don't want to touch at all if we can help it. We have not been (nor do I think we want to) enforcing copyright attribution, so we shouldn't be putting copyright notices on our files. I have updated our root `COPYRIGHT` to match the copyright text from Rust.

---

The second is licensing, which is slightly more complex. As I understand it, including a single license file at the root is technically sufficient to license the project. However, the Apache License includes these clauses:
```
  (b) You must cause any modified files to carry prominent notices
          stating that You changed the files; and
  (c) You must retain, in the Source form of any Derivative Works
          that You distribute, all copyright, patent, trademark, and
          attribution notices from the Source form of the Work,
          excluding those notices that do not pertain to any part of
          the Derivative Works; and
```
This is much easier to satisfy if there are notices in each file that is modified.


At this stage, I do not believe that it is strictly a requirement that every file include a license header (indeed the Apache usage guidelines say headers **should** be included, not **must**, and Rust excludes them universally).

However, it is also my understanding that including headers _while again technically completely unnecessary_ helps to nudge things in an otherwise legal possibly grey area. In the absence of headers, if someone making a derivative were ever legally challenged, the maker of the derivative work would have to prove when, where, and how they made the derivative, that the derivative file was covered by the blanket LICENSE, and that all contributions to that file over time were covered by that LICENSE. While the status quo (i.e. tracing back to 1eb06da30, assuming transitive licensing form there, and trusting that at that point we did get all contributors up to that point to agree to the license terms [which we did, orally, sorry future lawyers]) is again technically sufficient, the addition of a header to each file helps to make this licensing argument much _easier_ if needed. Of particular (legal) significance, there is a difference in _intent_ when a license notice is attached to each file explicitly rather than implicitly covered by a blanket LICENSE.

All of this is a long-winded way to say that adding (or not adding) headers in no material way changes how things have been or will be licensed. However, their addition de-risks entities looking to make derivative works, and if that's something that's important for our collaborators, then we should do it.

ppannuto:
[Note: there are separate and next questions around "which files need headers" and "where they are placed", but I'd like to come to consensus on adding them at all first]

jrvanwhy:
Here is an example of the friction that results from not having copyright headers in Tock's source code.

OpenTitan pull request [2119](https://github.com/lowRISC/opentitan/pull/2119/files) adds sw/device/tock/chips/earlgrey/src/chip.rs, which is a modified form of Tock's https://github.com/tock/tock/blob/master/chips/ibex/src/chip.rs. Following OpenTitan's policies, we must add a copyright header to OpenTitan's version of the file. However, because the original file does not have a copyright header, it would end up with _only_ OpenTitan's header, which risks misleading people.

Issue #1839, opened 2020-05-08
Tracking: Refactor nrf52 boards organization


hudson-ayers:
The discussion in #1784 seems to have arrived at the conclusion that the organization of the nordic boards no longer makes sense, now that we have components and support multiple non-development-kit nordic boards.

The desired result is to refactor the nordic boards such that each board has its own folder (like how all other boards are handled), and use components for all initialization to minimize the repeated code between board files.

Below are a series of largely independent steps that would get us to that result:

- [ ] move `nrf52dk_base/src/nrf52_components/` into top level of `boards`
- [x] create nrf-startup component that replaces beginning of `nrf52dk_base/src/lib.rs::setup_board()` (activate reset button, make flash writable, enable nfc pin protection, configure voltage regulator, start rtc, etc.) (#1853)
- [ ] consider component for uart channel initialization?, would use in `nrf52dk_base/src/lib.rs`
- [x] create nrf52-specific analog comparator component, or repurpose `imix_components/analog_compartor.rs` to be usable for initialization across different chips and move the result to the general components folder (#1853)
- [x] above, but for `nonvolatile_storage` initialization (#1853 )
- [x] Integrate nrf-specific ieee802154 component with imix-component that does the same thing, move this component to the general components folder (#1878  )
- [x] create `mx25r6435f` component (#1853)
- [x] transition `nrf52dk_base/src/lib.rs` to use SPI component (#1853 )
- [x] refactor si7021 component to pull out temperature component into individual component, which is then used by both Imix/Hail and nrf boards temperature sensor (#1853 )
-----------
Once those steps are done..
- [ ] Create individual directories for `nrf52dk`, `nrf52840dk`, and `nrf52840_dongle` in the top level of `boards/`, each with its own `main.rs` that copies in the completely-componentized `setup_board()` currently shared by those three boards. Delete `boards/nordic/` (#1859)

bradjc:
1. Why delete `boards/nordic`? It seems like grouping all of the nordic produced boards has served us well so far.

2. Why would nrf52dk_base still be necessary?

hudson-ayers:
> 1. Why delete `boards/nordic`? It seems like grouping all of the nordic produced boards has served us well so far.

Well, I always thought it was a little weird that the acd52832 wasn't in that folder (I realize that the *board* isn't made by nordic, just the chip, but still..), and also that we only do it for nordic boards but not, say, stm boards. But we could certainly still keep the folder.

[...]

PR #1853, opened 2020-05-14
Components for all things nrf52

hudson-ayers:
### Pull Request Overview

This pull request creates components for all portions of nrf52dk-specific initialization that does not already rely on components.

Specifically:

- Converts Imix-specific analog comparator component into generic analog comparator component, uses it for Imix and nrf52*

- Converts Imix-specific nonvolatile storage component into generic component, uses it for Imix and nrf52dk (required some HIL changes!)

- Converts si7021 specific temperature sensor component into generic temperature sensor component, uses it for Imix, nrf52*, and Hail.

- Creates a component for standalone initialization of the mx25r6435f flash chip used on the nrf52840dk

- Creates NrfStartupComponent and NrfClockComponent, to store repeated initialization code in functions that can be reused.

- Uses the existing SPI compnents to initialize SPI for the nrf52* boards.

This PR is the first step towards addressing #1839 .

### Testing Strategy

This pull request was tested by running the in-kernel log_test on Imix, running the nonvolatile_storage test on Imix, running the analog_comparator test on Imix, and running the analog_comparator and nonvolatile_storage tests on the nrf52840dk.

### TODO or Help Wanted

I am not sure if adding a `configure()` function to the flash HIL was the correct call given that only the sam4l flash controller requires it.

### Documentation Updated

- [x] No updates are required?

### Formatting

- [x] Ran `make formatall`.


boards/components/src/nonvolatile_storage.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        let kernel_len = self.kernel_end - self.kernel_start;
+
+        let nonvolatile_storage = static_init!(
+            NonvolatileStorage<'static>,
+            NonvolatileStorage::new(
+                nv_to_page,
+                self.board_kernel.create_grant(&grant_cap),
+                0x60000,           // Start address for userspace accessible region
+                0x20000,           // Length of userspace accessible region
```
bradjc, line 110: Can you make these two arguments as well? Users of this component should have to think about these values, otherwise this is even more dangerous than it already is.

boards/nordic/nrf52dk_base/src/lib.rs
```diff
@@ -323,113 +254,64 @@ pub unsafe fn setup_board<I: nrf52::interrupt_service::InterruptService>(
...
-    );
-    nrf52::acomp::ACOMP.set_client(analog_comparator);
+    let analog_comparator = components::analog_comparator::AcComponent::new(
+        &nrf52::acomp::ACOMP,
+        components::acomp_component_helper!(nrf52::acomp::Channel, &nrf52::acomp::CHANNEL_AC0),
+    )
+    .finalize(components::acomp_component_buf!(nrf52::acomp::Comparator));
 
     // Start all of the clocks. Low power operation will require a better
     // approach than this.
```
bradjc, line 299: I think this comment should move too.

kernel/src/hil/flash.rs
```diff
@@ -107,7 +113,11 @@ pub trait HasClient<'a, C> {
 /// A page of writable persistent flash memory.
 pub trait Flash {
     /// Type of a single flash page for the given implementation.
-    type Page: AsMut<[u8]>;
+    type Page: AsMut<[u8]> + Default;
+
+    /// Called by nonvolatile storage component. Responsible for chip-specific
+    /// configuration to enable flash.
+    fn configure(&self);
```
bradjc, line 120: I see where this comes from, but I think we would rather avoid adding `configure` as its unclear who should call it and when, particularly when it takes no arguments.

I think it would be pretty straightforward to check for the `FlashState::Unconfigured` state in the sam4l driver and call `self.configure()` if needed.

PR #1857, opened 2020-05-17
Initial support for the Artemis platform

alistair23:
### Pull Request Overview

SparkFun have create the [Artemis MCU module](https://www.sparkfun.com/artemis). It's a Ambiq Apoll3 MCU broken out so that it's ready to be used. On top of this they have a range of boards that you can use.

This PR is the first go at supporting the Apollo3 MCU and the RedBoard Artemis Nano.

Datasheet for the MCU: https://ambiqmicro.com/static/mcu/files/Apollo3_Blue_MCU_Data_Sheet_v0_11_0.pdf

What works:

- Tock runs
- UART with interrupts
- LED (tested with panic handler)
- `libtock-rs` apps

### Testing Strategy

Running on the SparkFun RedBoard Artemis Nano.

### TODO or Help Wanted

The main missing features right now are:
 - Add a Cortex-M4F crate in arch
 - Add the Apollo3 timer
 - Add the Apollo3 UART receive
 - Implement all of the GPIO functions

Not all of them need to be done before this can be merged.

### Documentation Updated

### Formatting

- [x] Ran `make formatall`.


alistair23:
Ok, so I figured out that with this diff the hard fault is fixed:

```diff
diff --git a/arch/cortex-m4/src/lib.rs b/arch/cortex-m4/src/lib.rs
index 00fe99d84..2827d602a 100644
--- a/arch/cortex-m4/src/lib.rs
+++ b/arch/cortex-m4/src/lib.rs
@@ -89,10 +89,6 @@ pub unsafe extern "C" fn generic_isr() {
     mov r0, #0
     msr CONTROL, r0
 
-    // This is a special address to return Thread mode with Main stack
-    movw LR, #0xFFF9
-    movt LR, #0xFFFF
-
     // Now need to disable the interrupt that fired in the NVIC to ensure it
     // does not trigger again before the scheduler has a chance to handle it. We
     // do this here in assembly for performance.
```

hudson-ayers:
Those two lines of assembly write a particular EXC_RETURN value to the cortex-m4 link register.

Accordingly, This may be a clue (from ARMs notes on what registers are different for the m4f): http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dai0298a/BCGHEEFD.html :

"The EXC_RETURN code value which is used in the exception mechanism has an additional bit field to define floating-point stack status. When the value of EXC_RETURN[4] is 0, it indicates that the exception stack frame for an exception return contains floating-point registers. If this bit is 1, then it means the exception stack frame does not contain floating-point registers."

So perhaps this value should be different because you are using an m4f? In particular, perhaps the first line should be `movw LR, #0xFFF1`?

Disclaimer: I am not an expert on Tock's interrupt handling or on Cortex-m interrupt handling in general so maybe this is totally wrong

alistair23:
That changes the error from (with the two lines included):

```
Precise Data Bus Error:             true
Bus Fault Address:                  0x72616F6A
Fault Status Register (CFSR):       0x00008200
Hard Fault Status Register (HFSR):  0x40000000
```

To (with `movw LR, #0xFFF1`)

```
Invalid PC Load Usage Fault:        true
Fault Status Register (CFSR):       0x00040000
Hard Fault Status Register (HFSR):  0x40000000
```

So an improvement?

hudson-ayers:
Wait, 0-indexing - try 0xFFE9?

alistair23:
That's it! Awesome

.gitignore
```diff
@@ -55,3 +55,6 @@ Cargo.lock
 
 # Used to build QEMU for emulation testing
 tools/qemu
+
+Python scripts
```
ppannuto, line 59:
```suggestion
# Python scripts
```

?

rajivr:
> Wait, 0-indexing - try 0xFFE9?

In Interrupt mode, with an `EXC_RETURN` value of `0xFFFF_FFE9`, we would return back to Thread mode with MSP when FPU is used. When FPU is not used the `EXC_RETURN` value would be `0xFFFF_FFF9`.

One subtlety here is that when FPU is used, the exception stack frames are different. There is a reasonable chance that our user-space programs would break.

hudson-ayers:
> In Interrupt mode, with an `EXC_RETURN` value of `0xFFFF_FFE9`, we would return back to Thread mode with MSP when FPU is used. When FPU is not used the `EXC_RETURN` value would be `0xFFFF_FFF9`.
> 
> One subtlety here is that when FPU is used, the exception stack frames are different. There is a reasonable chance that our user-space programs would break.

Looking through the apollo3 code in this PR I don't see anything that looks like it should be enabling the FPU, and per https://developer.arm.com/docs/dui0553/b/cortex-m4-peripherals/floating-point-unit-fpu/enabling-the-fpu it is disabled by default. So I wonder why `0xFFFF_FFE9` was needed to avoid a hard fault.

rajivr:
> Looking through the apollo3 code in this PR I don't see anything that looks like it should be enabling the FPU, and per https://developer.arm.com/docs/dui0553/b/cortex-m4-peripherals/floating-point-unit-fpu/enabling-the-fpu it is disabled by default. So I wonder why `0xFFFF_FFE9` was needed to avoid a hard fault.

I have a strange suspicion that the MCU vendor might be enabling FPU in their mask ROM, before handing control over to us. 

@alistair23 - I was wondering what is the value of the `CONTROL` register that you see when you set a breakpoint in `reset_handler`?

alistair23:
Which one is the control register?

```
r0             0x1d727f            1929855
r1             0xe000e104          -536813308
r2             0xffffffff          -1
r3             0xffffffff          -1
r4             0x1005ffd9          268828633
r5             0x7294              29332
r6             0x10008008          268468232
r7             0x10000ff8          268439544
r8             0x100               256
r9             0x0                 0
r10            0x0                 0
r11            0x0                 0
r12            0x80000000          -2147483648
sp             0x10000e00          0x10000e00 <STACK_MEMORY+3584>
lr             0x2ec9              0x2ec9
pc             0x11b48             0x11b48 <reset_handler+136>
xpsr           0x21000000          553648128
fpscr          0x2000000           33554432
msp            0x10000e00          0x10000e00 <STACK_MEMORY+3584>
psp            0x0                 0x0
special        0x4000000           67108864
```

alistair23:
I have pushed an update that works with the LED now. I also have tidied up the GPIO driver (although some bits are unimplemented). I have also fixed the UART to run the callback so that multiple prints work.

Let me know what you think.

ppannuto:
The CONTROL register isn't printed by default unfortunately. You can use the move special register (`msr`) instruction in assembly to load it into an architectural register. The existing context switch code has examples of how to do this.

chips/apollo3/src/nvic.rs
```diff
@@ -0,0 +1,35 @@
+//! Named constants for NVIC ids shared across the Apollo3 family of chips
+
+#![allow(non_upper_case_globals)]
```
hudson-ayers, line 3: no need for this it seems?

chips/apollo3/src/clkgen.rs
```diff
@@ -0,0 +1,63 @@
...
+        (0x100 => intrpten: ReadWrite<u32>),
+        (0x104 => intrptstat: ReadWrite<u32>),
+        (0x108 => intrptclr: ReadWrite<u32>),
+        (0x10c => intrptset: ReadWrite<u32>),
+        (0x110 => @END),
+    }
+}
+
+pub enum ClockFrequency {
+    FreqMax,
```
hudson-ayers, line 39: Is this 48MHz or 96 MHz? Not clear to me because of how apollo refers to 96 MHz as the TurboSpot frequency which seems to apply it can't run at that speed full time

alistair23:
So after running

```
msr	CONTROL, r0
```

r0 is:

```
0x006d79e8
```

ppannuto:
Oops, sorry, that was a lazy pointer on my part. I believe that will write r0 into control; you need to MRS ( http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0489c/Cihjcedb.html ):

`mrs r0, CONTROL`

In general, in ARM assembly, the leftmost operand is the destination

alistair23:
Ha, I should have thought about it and known that.

Anyway, `CONTROL` is apparently:

```
0x00000004
```

rajivr:
> Ha, I should have thought about it and known that.
> 
> Anyway, `CONTROL` is apparently:
> 
> ```
> 0x00000004
> ```

As I suspected, `FPCA` bit (Bit 2) is turned on. Could you disable it in reset handler by setting `CP10` and `CP11` in SCB.CPACR to `0`.

chips/apollo3/src/clkgen.rs
alistair23: I'm pretty sure it's 48MHz, I'll update this.

alistair23:
Ok. Thanks for @rajivr I can now disable the FPU and don't need anything special in the interrupt handling. It strangely needs a `svc` call, otherwise the FPU isn't disabled, not sure why.

Otherwise I have addressed all the comments. I'm going to mark this as ready for review.

alistair23:
I have the hello_world and LED blink examples work (although the timer doesn't) from `libtock_rs`

alistair23:
`libtock-rs` PR has been created: https://github.com/tock/libtock-rs/pull/183

arch/cortex-m/src/scb.rs
```diff
@@ -293,3 +293,15 @@ pub unsafe fn reset() {
...
+/// Disable the FPU
+pub unsafe fn disable_fpca() {
+    SCB.cpacr
+        .modify(CoprocessorAccessControl::CP10::CLEAR + CoprocessorAccessControl::CP11::CLEAR);
+
+    if SCB.cpacr.read(CoprocessorAccessControl::CP10) != 0
+        || SCB.cpacr.read(CoprocessorAccessControl::CP11) != 0
+    {
+        panic!("Unable to disable FPU");
+    }
```
hudson-ayers, line 306: Section 4.6.6 suggests that a dsb and isb instruction are needed after setting these bits: https://www.st.com/resource/en/programming_manual/dm00046982-stm32-cortex-m4-mcus-and-mpus-programming-manual-stmicroelectronics.pdf
Perhaps that would remove the need for the svc call?

chips/apollo3/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    cortexm4::nvic::disable_all();
+    cortexm4::nvic::clear_all_pending();
+    cortexm4::nvic::enable_all();
+}
+
+// Mock implementation for tests
+#[cfg(not(any(target_arch = "arm", target_os = "none")))]
+pub unsafe fn init() {
+    scb::disable_fpca();
```
hudson-ayers, line 134: I don't think this should be here -- this function should never be called on a host architecture. Or is this just to prevent an unused code warning?

chips/apollo3/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+    // correctly for Tock. The bootloader _may_ set this for us, but it may not
+    // so that any errors early in the Tock boot process trap back to the bootloader.
+    // To be safe we unconditionally set the vector table.
+    scb::set_vector_table_offset(BASE_VECTORS.as_ptr() as *const ());
+
+    // Disable the FPU (it might be enalbed by a prior stage)
+    scb::disable_fpca();
+
+    // This ensures the FPU is actually disabled
+    asm!("svc 0xff");
```
rajivr, line 124: I think we should probably indicate to the `asm!` that `r0-r3, r12` is going to potentially get clobbered by `svc 0xff`.

chips/apollo3/src/lib.rs
alistair23: It's just to prevent unused warnings

arch/cortex-m/src/scb.rs
alistair23: I tested with these in and I still need the svc call.

chips/apollo3/src/lib.rs
ppannuto:
```suggestion
    // Prevent unused code warning.
    scb::disable_fpca();
```

boards/redboard_artemis_nano/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+```shell
+$ make flash
+
+(or)
+
+$ make flash-debug
+```
+
+This will flash Tock onto the board via the /dev/ttyUSB0 port. If you would like to use a different port you can specify it from the `PORT` varialbe.
```
bradjc, line 31:
```suggestion
This will flash Tock onto the board via the /dev/ttyUSB0 port. If you would like to use a different port you can specify it from the `PORT` variable.
```

chips/apollo3/README.md
```diff
@@ -0,0 +1,7 @@
+For Apollo3 micro controllers
+
```
bradjc, line 2:
```suggestion
Apollo3 Microcontrollers
==================

```

alistair23:
I pushed an update with the doc fixes requested by @bradjc. This should be good to go :)

Issue #1252, opened 2018-12-14
Cortex-M `generic_isr()` needs to ensure pending bit is set


bradjc:
https://groups.google.com/forum/#!topic/tock-dev/fZQYq0dpeSQ

A very short pulse interrupt could cause `generic_isr()` to end without the corresponding pending bit set, and the interrupt would not be handled properly. `generic_isr()` should manually set the pending bit in the ISR so that in all cases the pending bit is set after `generic_isr()` ends so that the interrupt will always be handled.

PR #1858, opened 2020-05-18
Set interrupt pending bit manually

cody82:
### Pull Request Overview
This pull request fixes the bug mentioned here:
https://github.com/tock/tock/issues/1252

generic_isr() is called but the interrupt is not handled in service_pending_interrupts() as the pending bit is not set.
While this bug does not seem to be triggered on tock supported boards, I encountered the problem on an automotive board.

### Testing Strategy
This was tested on a hardware security module. After this fix the interrupt is handled properly in service_pending_interrupts().

### TODO or Help Wanted
This should be tested on tock boards as I do not have one.

### Documentation Updated
No documentation was updated.

### DIFF
--- a/arch/cortex-m0/src/lib.rs
+++ b/arch/cortex-m0/src/lib.rs
@@ -90,6 +90,18 @@ _ggeneric_isr_no_stacking:
 
     /* *r3 = r2 */
     str r2, [r3]
+
+    /* The pending bit in ISPR might be reset by hardware for pulse interrupts
+     * at this point. So set it here again so the interrupt does not get lost
+     * in service_pending_interrupts()
+     *
+     * The NVIC.ISPR base is 0xE000E200, which is 0x20 (aka #32) above the
+     * NVIC.ICER base.  Calculate the ISPR address by offsetting from the ICER
+     * address so as to avoid re-doing the [r0 / 32] index math.
+     */
+    adds r3, #32
+    str r2, [r3]
+
     bx lr /* return here since we have extra words in the assembly */
 
 .align 4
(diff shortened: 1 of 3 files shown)


rajivr:
@cody82 I am curious to see if the peripheral in this case was doing _pulsed_ interrupt request rather than _high level_ interrupt request?

If I recall correctly, this was something we had discussed in the past on the mailing list. The reason why our current `generic_isr` works is kind of subtle - It uses a combination of,

1. Peripheral issues *only* _high level_ interrupt request

2. When we enter `generic_isr`, ISPR is automatically cleared.

3. In `generic_isr`, we disable ICER.

4. Because the peripheral has NVIC line high, ISPR is set by hardware again, but we don't enter `generic_isr` because ICER is disabled.

5. Finally in `next_pending`, we rely on ISPR to service the interrupt.

cody82:
It is a pulse interrupt and the pending bit is reset by hardware when entering generic_isr().
I will add an explanation to the code.

cody82:
The m0 code looks quite different and I am not sure whats the best way to fix it. Can we just copy the m3/m4 code or is it intended to be so different?

ppannuto:
@alistair23 This travis failure looks unrelated to this PR, did something change in QEMU? https://travis-ci.org/github/tock/tock/builds/689229773#L1213

hudson-ayers:
> @alistair23 This travis failure looks unrelated to this PR, did something change in QEMU? https://travis-ci.org/github/tock/tock/builds/689229773#L1213

I think there were issues with the travis qemu cache -- I am going to rerun the build.

alistair23:
I'm assuming that the Travis cache issue is because some of the PRs are based on older versions of Tock and that build is cached. Travis no longer runs QEMU checks and doesn't have caching anymore so it should be fixed once rebased (I hope).

ppannuto:
bors r+

Tested on hail with hail app that happily still hails.

PR #1861, opened 2020-05-19
Separate the Makefile rules that require qemu, and stop silently installing qemu.

gendx:
### Pull Request Overview

This pull request adds a new `ci-no-qemu` rule to the Makefile, which runs all CI rules except the ones that require qemu. This also removes the implicit dependency on `emulation-setup` for the `emulation-check` rule.

I find the `make ci` rule convenient for locally testing my commits before sending a pull request. But it now requires qemu, via `ci -> ci-travis -> emulation-check -> emulation-setup`. However, this automatic `emulation-setup` is problematic because:

1. I find it surprising that it automatically tries to install qemu "behind my back".
1. It anyway doesn't work out of the box without installing some other Debian packages.
1. As evidenced in [this workflow](https://github.com/tock/tock/pull/1861/checks?check_run_id=688603466) the qemu installation takes a bit of time (more than 5 minutes on the GitHub workflow runner).
1. The qemu tests are currently specific to some boards. Although this may change in the future, they may not be of high interest for developers of other boards.

I don't mind some rules silently installing Rust tools (Cargo, Clippy, Rustfmt, some compiler target), as these are anyway required to build Tock. But I don't think qemu should necessarily be in scope for **local** testing. Of course, developers who want to use it locally are free to do so, and the GitHub workflows check these tests, but I think there should be some freedom in what part of the CI a developer wants to run.

This pull request therefore adds a few rules to the Makefile that don't require qemu. It also removes the `emulation-setup` implicit dependency from `emulation-check` - one now has to explicitly choose to install qemu.

~~This allowed me to run the CI locally again, and identify some warning in the documentation, which is also fixed in this pull request.~~ Fixing the documentation is now in #1862.

### Testing Strategy

This pull request was tested by Travis, GitHub workflows, and `make ci-no-qemu` locally.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make format`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ before_script:
 
 script:
   - export PATH=$HOME/.cargo/bin:$PATH
+  - make emulation-setup
   - make ci-travis
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ before_script:
 script:
   - export PATH=$HOME/.cargo/bin:$PATH
   - make emulation-setup
-  - make ci-travis
+  - make ci-travis-with-qemu
 
 after_success:
   - export PATH=$HOME/.local/bin:$PATH
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,6 @@ before_script:
 
 script:
   - export PATH=$HOME/.cargo/bin:$PATH
-  - make emulation-setup
   - make ci-travis-with-qemu
 
 after_success:
--- a/Makefile
+++ b/Makefile
@@ -60,18 +60,12 @@ alldoc:
 .PHONY: ci
 ci: ci-travis ci-netlify
 
+.PHONY: ci-no-qemu
+ci-no-qemu: ci-travis-no-qemu ci-netlify
+
 .PHONY: ci-travis
 ci-travis:\
-	ci-lints\
-	ci-tools\
-	ci-libraries\
-	ci-archs\
-	ci-kernel\
-	ci-chips\
-	ci-syntax\
-	ci-compilation\
-	ci-debug-support-targets\
-	ci-documentation \
+	ci-travis-no-qemu \
 	emulation-check
 	@printf "$$(tput bold)********************$$(tput sgr0)\n"
 	@printf "$$(tput bold)* CI-Travis: Done! *$$(tput sgr0)\n"
@@ -83,6 +77,19 @@ ci-netlify:\
 	@printf "$$(tput bold)*********************$$(tput sgr0)\n"
 	@printf "$$(tput bold)* CI-Netlify: Done! *$$(tput sgr0)\n"
 	@printf "$$(tput bold)*********************$$(tput sgr0)\n"
+
+.PHONY: ci-travis-no-qemu
+ci-travis-no-qemu:\
+	ci-lints\
+	ci-tools\
+	ci-libraries\
+	ci-archs\
+	ci-kernel\
+	ci-chips\
+	ci-syntax\
+	ci-compilation\
+	ci-debug-support-targets\
+	ci-documentation \
 
 .PHONY: ci-cargo-tests
 ci-cargo-tests:\
--- a/Makefile
+++ b/Makefile
@@ -57,14 +57,14 @@ alldoc:
 
 ## Meta-Targets
 
-.PHONY: ci
-ci: ci-travis ci-netlify
+.PHONY: ci-with-qemu
+ci: ci-travis-with-qemu ci-netlify
 
 .PHONY: ci-no-qemu
 ci-no-qemu: ci-travis-no-qemu ci-netlify
 
-.PHONY: ci-travis
-ci-travis:\
+.PHONY: ci-travis-with-qemu
+ci-travis-with-qemu:\
 	ci-travis-no-qemu \
 	emulation-check
 	@printf "$$(tput bold)********************$$(tput sgr0)\n"
@@ -240,7 +240,7 @@ emulation-setup:
 
 
 .PHONY: emulation-check
-emulation-check:
+emulation-check: emulation-setup
 	@$(MAKE) -C "boards/hifive1"
 	@$(MAKE) -C "boards/opentitan"
 	@cd tools/qemu-runner; PATH="$(shell pwd)/tools/qemu/riscv32-softmmu/:${PATH}" cargo run
--- a/Makefile
+++ b/Makefile
@@ -240,7 +240,7 @@ emulation-setup:
 
 
 .PHONY: emulation-check
-emulation-check: emulation-setup
+emulation-check:
 	@$(MAKE) -C "boards/hifive1"
 	@$(MAKE) -C "boards/opentitan"
 	@cd tools/qemu-runner; PATH="$(shell pwd)/tools/qemu/riscv32-softmmu/:${PATH}" cargo run
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,5 +92,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
+      - name: qemu setup
+        run: make emulation-setup
       - name: qemu tests
         run: make emulation-check
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,5 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
-      - name: qemu setup
-        run: make emulation-setup
       - name: qemu tests
         run: make emulation-check


Makefile
```diff
@@ -233,7 +240,7 @@ emulation-setup:
 
 
 .PHONY: emulation-check
-emulation-check: emulation-setup
+emulation-check:
```
ppannuto, line 243: I understand the motivation behind removing the implicit setup step, but in the spirit of "help new people to do the right thing automatically", could you add a check to the target here, probably something like:

`if [ ! -x tools/qemu/riscv32-softmmu/qemu-system-riscv32 ]; echo "Error: Running emulation checks requires qemu. Please run 'make emulation-setup' to prepare. This will take several minutes, but only needs to be run once" && exit 1; fi`

---

Generally, the Tock build system tries pretty hard to avoid the idiom of "the command you tried to run failed, please type this other command and now try again", and instead simply does the work for you.

While I am sympathetic to the long build of qemu, it's also a one-off cost. I might be more inclined to just add a print to the `emulation-setup` rule that says something like, "First run requires building qemu. This will take a few minutes. You can cancel and run `make ci-no-qemu` to skip for now".

.travis.yml
```diff
@@ -42,6 +42,7 @@ before_script:
 
 script:
   - export PATH=$HOME/.cargo/bin:$PATH
+  - make emulation-setup
```
hudson-ayers, line 45: one downside of this is that we lose fail-fast behavior for Travis when the error is something simple like a formatting error that would be caught in the first step. I only think this is okay because we have Github Actions now which should fail fast instead.

Makefile
bradjc: Shouldn't `emulation-check` always depend on setup? How can it do the check without the setup? I guess it doesn't matter to me if it is a check in the makefile or a target dependency, but since we don't want to run it on every make call it kind of has to be a target dependency.

gendx: > Generally, the Tock build system tries pretty hard to avoid the idiom of "the command you tried to run failed, please type this other command and now try again", and instead simply does the work for you.

On the other hand, I would argue that the current setup doesn't try that hard and fails pretty quickly without the necessary (Debian) packages.

```
$ make emulation-check 
Cloning into 'qemu'...
remote: Enumerating objects: 19007, done.
remote: Counting objects: 100% (19007/19007), done.
remote: Compressing objects: 100% (6200/6200), done.
remote: Total 482048 (delta 14287), reused 16209 (delta 12647), pack-reused 463041
Receiving objects: 100% (482048/482048), 195.57 MiB | 27.38 MiB/s, done.
Resolving deltas: 100% (394231/394231), done.
Checking connectivity... done.

ERROR: pkg-config binary 'pkg-config' not found

Makefile:219: recipe for target 'emulation-setup' failed
make: *** [emulation-setup] Error 1
```

And installing the required dependencies will likely require `root` privileges, which will anyway require manual user action beyond Tock's Makefile.

---

Could you add a check to the target here, probably something like:

`if [ ! -x tools/qemu/riscv32-softmmu/qemu-system-riscv32 ]; echo "Error: Running emulation checks requires qemu. Please run 'make emulation-setup' to prepare. This will take several minutes, but only needs to be run once" && exit 1; fi`

I'd prefer someone who has actually used the qemu rules to add that.

But in any case, yes the `emulation-check` rule should check for required dependencies and ask the user to install them before downloading 330MB of qemu history. By the way, can't the rule use `git clone --depth 1` to avoid cloning the whole history of qemu?

gendx: Also, the next error is even less clear:

```bash
$ sudo apt install pkg-config
...
$ make emulation-check 
make[1]: *** [config-host.mak] Error 1
Makefile:219: recipe for target 'emulation-setup' failed
make: *** [emulation-setup] Error 2
```

ppannuto:
Attn @alistair23 for some of the qemu ergonomics issued raised here

alistair23:
A few things here:

>     1. I find it surprising that it automatically tries to install qemu "behind my back".

QEMU is never installed. There is a local clone, but never an install. Ever.

The `emulation-setup` will instead clone and build a local QEMU isntance. Without it the `emulation-check` won't work so it depends on the setup. Also, that setup should only happen once and until you run a `git clean` it won't be rebuilt.

> * It anyway doesn't work out of the box without installing some other Debian packages.

I'm sorry you are missing some development packages, but there isn't really a good fix here. I don't think a Makefile should install packages so I don't want to install anything. We could document that you need to install `build-essentials` in order to build QEMU if that helps?

Makefile
alistair23: We could stop piping the output of make to `/dev/null`. That is why you don't have a clear error.

alistair23:
I have sent a PR that:
 - Should make the build errors clearer
 - Does a shallow clone

https://github.com/tock/tock/pull/1866

bradjc:
I think this is a work in progress, and that is fine. QEMU support is brand new, and there are bound to be some issues. Makefile progress is incremental.

What seems to be more fundamental and long lasting is the naming for the make targets. I propose we have three categories:
1. Individual commands. These are targets that do narrow and specific actions. Things like `make format` or `make checkall`. We will probably have many of these.
2. Developer commands. These are targets that individual developers would use before submitting pull requests. These would do things like run checks or formatting or tests. Ideally they should be easy to run (i.e. would only use cargo or tools contained in the tock repository).
3. Automated CI commands. These would only be expected to run by automated CI tools (although nothing would stop them from running locally as well). Developers would not be expected to run these manually before submitting a PR.

The naming should then reflect these categories. In particular, I think the "*ci*" targets should be reserved for category 3, and we need a new name for category 2 commands.

This then would help with growth with our CI infrastructure, as it would make a place for checks that we don't expect users to run locally, without having to do something like `make ci-no-qemu-no-rpi-no-XYZ`.

gendx:
Replying here on the points still relevant to this pull request.

> What seems to be more fundamental and long lasting is the naming for the make targets. I propose we have three categories:
> 
> 1. Individual commands. These are targets that do narrow and specific actions. Things like `make format` or `make checkall`. We will probably have many of these.
> 2. Developer commands. These are targets that individual developers would use before submitting pull requests. These would do things like run checks or formatting or tests. Ideally they should be easy to run (i.e. would only use cargo or tools contained in the tock repository).
> 3. Automated CI commands. These would only be expected to run by automated CI tools (although nothing would stop them from running locally as well). Developers would not be expected to run these manually before submitting a PR.
> 
> The naming should then reflect these categories. In particular, I think the "_ci_" targets should be reserved for category 3, and we need a new name for category 2 commands.

This is where we disagree. Being able to test ci targets locally is essential for my development UX. This is for several reasons:
- I don't want to submit a pull request every time I start experimenting with something, in order to get feedback from Travis-CI or GitHub-Actions (and Tock reviewers probably don't want that either).
- The latency of Travis-CI is too slow, because it doesn't cache my build locally. If I run `make ci` locally and some error happens, I can fix the code and re-run `make ci` benefiting from Rust's incremental compilation rather than having to wait for the equivalent of `make clean && make ci`.

Until now, the `make ci` rule was encompassing all I need as a developer in category 2. But the new qemu rules are too heavy for me.

Also, how can the `make ci` rule be reserved in category 3 when it's not used in any CI pipeline? Travis-CI uses `make ci-travis`, Netlify uses `make ci-netlify`, the GitHub actions use more specific rules. So to me, `make ci` was really this command allowing to locally run all the CI as a sanity check before uploading a pull request.

However, another developer may find the qemu rules useful in their category 2. So I don't think any rule should be reserved to category 3. Any rule should be easily pre-viewable by anyone locally.

> This then would help with growth with our CI infrastructure, as it would make a place for checks that we don't expect users to run locally, without having to do something like `make ci-no-qemu-no-rpi-no-XYZ`.

Then, something as simple as `make ci` shouldn't include the qemu installation, at least not without any prompt.

If you don't like the negative formulation, we could have `make ci` (which does what it did until recently) and `make ci-with-qemu` (with does everything, where one clearly opts-in to run qemu-based rules). Or `make ci && make qemu-tests`. Or find some other name like `make tests` to replace `make ci`?

gendx:
Any update on this? I really don't want to have this problem lingering around until the following concerns are resolved.
- **Insecure integration**: As described in https://github.com/tock/tock/issues/1865#issuecomment-630976331, in particular not coming from an "official" source such as https://github.com/qemu/qemu, no pinning to a specific commit vetted by Tock, and (as a corollary of the previous points) arbitrary code execution in the `./configure` step.
- **Problematic installation steps**:
  - This pulls hundreds of MB.
  - There is not prompt asking for confirmation of the user.
  - The installation rule is still changing fast (e.g. https://github.com/tock/tock/pull/1866), so we should wait a bit for more user feedback before putting it by default in `make ci`.

While the `emulation-check` is useful for CI, the current integration is still IMO a draft and as such shouldn't be part of `make ci`.

From a latency point of view anyway, it would be better to leave `emulation-check` out of Travis-CI for now and have it only in the relevant workflow.

ppannuto:
I have had a *lot* of frustration over the years trying to replicate things that fail in various CIs locally. I place very high value on being able to run `make ci` locally and having that match the result of CI. I would consider it strictly a bug if the success/fail of CI (in the aggregate) != the result of `make ci` locally.

I am, however, sympathetic to the point that the QEMU stuff is experimental and a bit heavyweight to foist on people while we figure out what it will look like long term. To that end, hopefully #1869 is a reasonable compromise.

---

I'm not sure that I agree with the proposed categorization. I see the rules breaking down more along the axes of:

1) Things that validate things (the `ci-*` family)

2) Things that take actions (formatting, etc)

I am very strongly opposed to creating a world where there are things that are expected to only to be able to run in CI infrastructure.

bradjc:
I don't think I worded my categories well. I'm not suggesting that there would be something special about the CI environment that would prevent those targets from running locally. But not all developers have the same requirements, and many (myself included) are more than happy to let cloud CI do the real checks in exchange for my own convenience. If you want the full suite then the CI targets are there. If you want something more user friendly, there would be a target for you. That seems much more sustainable than a series of custom "-no-X" targets.

From a practical standpoint, there are some easy checks to run locally that catch a majority of travis errors. We should make those easy for users to run. We also should pick a name that won't change, but will still contain all of the "easy" checks (i.e. don't contain QEMU).

>Until now, the make ci rule was encompassing all I need as a developer in category 2. But the new qemu rules are too heavy for me (besides my security concerns described in #1865 (comment)).

I don't understand this in relation to your point about needing to run all CI checks locally. Going forward, QEMU is going to be a huge part of CI testing. We are also adding kernel testing infrastructure which may be part of CI. Further out, we will be running automatic tests on actual embedded hardware as part of CI. This problem is only going to get worse. Running all of the CI checks locally is going to be a big lift, I don't see any way around that.

ppannuto:
> From a practical standpoint, there are some easy checks to run locally that catch a majority of travis errors. We should make those easy for users to run. We also should pick a name that won't change, but will still contain all of the "easy" checks (i.e. don't contain QEMU).

This makes sense. Something like `make ci-fast` or `make prepush` (could even suggest installing a git pre-push hook that runs that automatically).

ppannuto:
Also, FWIW, while the QEMU installation is long / slow, that is strictly a one-time cost. Running the QEMU tests on my machine once installed takes less than a second.

bradjc:
> Also, FWIW, while the QEMU installation is long / slow, that is strictly a one-time cost. Running the QEMU tests on my machine once installed takes less than a second.

Does it re-use the existing built objects? I expect that the required version of QEMU will change frequently, and that `qemu-setup` will have to be run periodically.

alistair23:
It shouldn't change frequently. Once the patches we are waiting on are reviewed and merged into mainline we can use submodules that don't change very often.

gendx:
> I don't understand this in relation to your point about needing to run all CI checks locally. Going forward, QEMU is going to be a huge part of CI testing. We are also adding kernel testing infrastructure which may be part of CI. Further out, we will be running automatic tests on actual embedded hardware as part of CI. This problem is only going to get worse. Running all of the CI checks locally is going to be a big lift, I don't see any way around that.

I think a good compromise could be to categorize the CI checks in terms of what they need. E.g. (exact names TBD) `cargo` (all tests/checks/lints/formatting that only require Cargo, and were the `make ci` until recently), `website` (what's needed for netlify), `emulation` (qemu-based), `hardware`, etc.

Then it should be clearer and easier to pick what to run locally as a developer.

gendx:
> It shouldn't change frequently. Once the patches we are waiting on are reviewed and merged into mainline we can use submodules that don't change very often.

We should still keep an eye on potential security bugs fixed by upstream for example, so I'd still expect some amount of updating from time to time (but not every day).

bradjc:
Is this all addressed in #1880?

gendx:
> Is this all addressed in #1880?

Yes. Let's focus on that pull request.

PR #1867, opened 2020-05-19
platform: mpu: Add function to set kernel permissions

alistair23:
### Pull Request Overview

The RISC-V enhanced PMP spec is making progress and it seems like it is unlikely to have any major changes before ratification.

In anticipation of this let's start to prepare adding support to Tock.

This PR updates the MPU trait to allow marking regions of kernel memory with permissions.

The idea here is that before the MPU is enabled Tock will set read/write/execute permissions for itself. This will limit what Tock itself can access, for example by removing write permissions from code.

We can also add a check to see if it is already enabled. If the previous stage has enabled lock down mode we won't set it ourselves. We will have to trust the previous stage (that would have loaded us) to have done it correctly. If no one else has set it, we can set it for enhanced security.

Once enabled we can't change the execute permission (read/write can change) so let's enforce that all regions are added before enabling.

### Testing Strategy

Untested.

### TODO or Help Wanted

I need feedback on the generic API before I go too far in enabling it and test with QEMU.

### Documentation Updated

### Formatting

- [X] Ran `make format`.

### DIFF
--- a/kernel/src/platform/mpu.rs
+++ b/kernel/src/platform/mpu.rs
@@ -112,6 +112,43 @@ pub trait MPU {
         0
     }
 
+    /// Mark a region of memory that the Tock kernel owns
+    ///
+    /// This function will optionally set the MPU to enforce the specified
+    /// constraints for all accessess (even from the kernel).
+    /// This should be used to mark read/write/execute only areas of the Tock
+    /// kernel to have the hardware enforce those permissions.
+    ///
+    /// If using this function, all kernel executable code must be marked as
+    /// executable prior to enabling the MPU (by calling `enable_mpu()).
+    /// Also all kernel read/write regions must be marked. That is any
+    /// unmarked kernel regions will cause a fault if accesses.
+    ///
+    /// Not all architectures support this, so don't assume this will be
+    /// implemented.
+    ///
+    /// # Arguments
+    ///
+    /// - `memory_start`:             start of memory region
+    /// - `memory_size`:              size of unallocated memory
+    /// - `permissions`:              permissions for the region
+    /// - `config`:                   MPU region configuration
+    ///
+    /// # Return Value
+    ///
+    /// Returns the start and size of the requested memory region. If it is
+    /// infeasible to allocate the MPU region, returns None.
+    #[allow(unused_variables)]
+    fn allocate_kernel_region(
+        &self,
+        memory_start: *const u8,
+        memory_size: usize,
+        permissions: Permissions,
+        config: &mut Self::MpuConfig,
+    ) -> Option<Region> {
+        None
+    }
+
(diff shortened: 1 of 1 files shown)


kernel/src/platform/mpu.rs
```diff
@@ -112,6 +112,43 @@ pub trait MPU {
...
 
+    /// Mark a region of memory that the Tock kernel owns
+    ///
+    /// This function will optionally set the MPU to enforce the specified
+    /// constraints for all accessess (even from the kernel).
+    /// This should be used to mark read/write/execute only areas of the Tock
+    /// kernel to have the hardware enforce those permissions.
+    ///
+    /// If using this function, all kernel executable code must be marked as
+    /// executable prior to enalbing the MPU (by calling `enable_mpu()).
```
bradjc, line 123:
```suggestion
    /// executable prior to enabling the MPU (by calling `enable_mpu()`).
```

kernel/src/platform/mpu.rs
```diff
@@ -112,6 +112,43 @@ pub trait MPU {
...
+    /// - `memory_size`:              size of unallocated memory
+    /// - `permissions`:              permissions for the region
+    /// - `config`:                   MPU region configuration
+    ///
+    /// # Return Value
+    ///
+    /// Returns the start and size of the requested memory region. If it is
+    /// infeasible to allocate the MPU region, returns None.
+    #[allow(unused_variables)]
+    fn mark_kernel_region(
```
bradjc, line 142:
```suggestion
    fn allocate_kernel_region(
```

alistair23:
This is the type of flow, for RISC-V with enhanced PMP and no previous stage has setup PMP.

1. Tock boots with PMP disabled and un-configured
2. Tocks adds all of the regions that it needs as PMP regions. This will give itself read/write/execute to all the segments of memory it needs.
3. Tock enables PMP
4. At this point PMP can never be disabled. That is mostly enforced by the hardware (nothing related to execution can change) but Tock won't change it's own mappings. The PMP disable function will change to a nop
5. Before starting an app, we configure the apps PMP regions (the same as today)
6. After coming back from an app we can disable the app region (although we don't have to)

The enhanced PMP means that we can't disable anything related to execution for the kernel once enabled. This is enforced by HW so disable will become a nop.

The interface will be that everything must be explicitly allowed otherwise it will fault. For apps that isn't a change, for the kernel we just need to make sure we mark regions before enabling it.

kernel/src/platform/mpu.rs
alistair23: I didn't go with allocate originally as we aren't actually allocating anything, it's already allocated, so I worried that was a little confusing.

alistair23:
Replaced with: https://github.com/tock/tock/pull/1873

Issue #1860, opened 2020-05-19
Possible RISC-V PMP improvements


alphan:
Creating this issue to keep track of things that I noticed while working on #1822: 
- [x] `disable_mpu` should be a no-op (similar to `enable_mpu`) since permissions are not enforced on M-mode accesses (unless the lock bits are set, which we don't).
    Note: #1822 moves its implementation to `PMPConfig` and calls it from `configure_mpu` but it looks like we don't really need it.
    Note: Now part of #1873.
- [x] `rv32i::pmp::PMPConfig` is used as both `MPUConfig` (in `arch/rv32i/src/pmp.rs`) and `MPU` (in `chips/ibex/src/chip.rs`), which is confusing.
- [x] Do the regions really have to be sorted each time the MPU is configured? It looks like we can sort the list when it is modified if we can keep track of the app memory region instead of fixing its location. #2021

bradjc:
Hopefully a fair summary of the discussion so far:

- `disable()` will remain as a utility function for anyone who might want it, but it will be renamed to "clear" or something to imply that it _cannot_ guarantee that it can disable the MPU (certain hardware may lock access to _any_ software in a certain configuration).
- A new function "disable for kernel" or some name will replace the current `disable()` call. On current platforms this will be a no-op.
- Protecting the kernel from itself will be a new trait. The current trait will be clearly labeled as for applications.
- Performance optimizations will be left to specific implementations (subject to minor trait changes as needed). We don't expect too many implementations to ever exist, and if caching can be done internally to the interface, then better to not have to change the interface.

alistair23:
Sounds good :)

That is what is being done here: https://github.com/tock/tock/pull/1873

bradjc:
Over time I think these issues have been addressed.

PR #1871, opened 2020-05-20
rv32i: Convert PMP disable to a nop

alistair23:
### Pull Request Overview

We never set the lock bit for PMP (and if we did we can't unset it
anyway) so PMP won't apply to Tock only apps. In that case we never need
to disable it as it will be fully reconfigured before running the next
app.

### Testing Strategy

This pull request was tested by running an app on the HiFive1 revB and the QEMU CI.

### Documentation Updated

### Formatting

- [X] Ran `make format`.


arch/rv32i/src/pmp.rs
```diff
@@ -177,183 +177,7 @@ impl kernel::mpu::MPU for PMPConfig {
...
-        }
-        //set first PMP to have permissions to entire space
-        csr::CSR.pmpaddr0.set(0xFFFF_FFFF);
-        //enable R W X fields
-        csr::CSR.pmpcfg0.modify(csr::pmpconfig::pmpcfg::r0::SET);
-        csr::CSR.pmpcfg0.modify(csr::pmpconfig::pmpcfg::w0::SET);
-        csr::CSR.pmpcfg0.modify(csr::pmpconfig::pmpcfg::x0::SET);
-        csr::CSR.pmpcfg0.modify(csr::pmpconfig::pmpcfg::a0::TOR)
-    }
+    fn disable_mpu(&self) {}
```
hudson-ayers, line 180: This probably needs a comment explaining why we don't do need to do anything here

alistair23:
There is no harm in disabling the PMP, this is just more efficient. We waste a lot of cycles writing all these CSRs. In saying that if someone has set the lock bits, it can not be disabled, so the function as is won't work.

https://github.com/tock/tock/pull/1867 is more focused on ePMP so I would like to keep that a littler separate. https://github.com/tock/tock/pull/1867 will make no difference to regular PMP.

`disable_mpu_for_kernel()` isn't really right either as with ePMP it will never be disabled. Maybe we should just remove the disable function? I know Tock doesn't like init functions, but maybe we can have an init/clean function to call at bootup that resets anything the prior stage has done (if possible).

alphan:
#1860

alistair23:
Replaced with: https://github.com/tock/tock/pull/1873

PR #1873, opened 2020-05-21
PMP and MPU Improvements

alistair23:
### Pull Request Overview

This PR replaces:
 - https://github.com/tock/tock/pull/1867
 - https://github.com/tock/tock/pull/1871

And combines a few different MPU/PMP improvements

 1. Rename the `enable/disable_mpu()` functions to be `enable/disable_app_mpu()` functions. This is because these functions only enable and disable the MPU for apps.
 2. Conver the current `disable_mpu()` function to be `clear_mpu()` and run it at boot. Instead of clearing the MPU/PMP after returning to the kernel from an app let's instead do nothing as we don't need to. We still leave the function around though, encase a future architectures wants to use it. This should improve syscall performance, especially for RISC-V. We also clear the MPU/PMP at boot encase a previous stage has set anything up.
 3. We also add new functions in preperation for RISC-V ePMP (see below)

#### RISC-V ePMP

The RISC-V enhanced PMP spec is making progress and it seems like it is unlikely to have any major changes before ratification.

In anticipation of this let's start to prepare adding support to Tock.

This PR updates the MPU trait to allow marking regions of kernel memory with permissions.

The idea here is that before the MPU is enabled Tock will set read/write/execute permissions for itself. This will limit what Tock itself can access, for example by removing write permissions from code.

We can also add a check to see if it is already enabled. If the previous stage has enabled lock down mode we won't set it ourselves. We will have to trust the previous stage (that would have loaded us) to have done it correctly. If no one else has set it, we can set it for enhanced security.

Once enabled we can't change the execute permission (read/write can change) so let's enforce that all regions are added before enabling.

### Testing Strategy

Running on QEMU, I'll test the Unleashed and Apollo3 as well.

### Documentation Updated

### Formatting

- [X] Ran `make format`.

### DIFF
--- a/kernel/src/platform/mpu.rs
+++ b/kernel/src/platform/mpu.rs
@@ -93,20 +93,29 @@ pub trait MPU {
     /// current state to help with debugging.
     type MpuConfig: Default + Display = MpuConfigDefault;
 
-    /// Enables the MPU.
+    /// Clears the MPU.
+    ///
+    /// This function will clear any access control enforced by the
+    /// MPU where possible.
+    /// On some hardware it is impossible to reset the MPU after it has
+    /// been locked, in this case this function wont change those regions.
+    fn clear_mpu(&self) {}
+
+    /// Enables the MPU for userspace apps.
     ///
     /// This function must enable the permission restrictions on the various
     /// regions protected by the MPU.
-    fn enable_mpu(&self) {}
+    fn enable_app_mpu(&self) {}
 
-    /// Disables the MPU.
+    /// Disables the MPU for userspace apps.
     ///
-    /// This function must completely disable any access control enforced by the
-    /// MPU. This will be called before the kernel starts to execute as on some
+    /// This function must disable any access control that was previously setup
+    /// for an app if it will interfere with the kernel.
+    /// This will be called before the kernel starts to execute as on some
     /// platforms the MPU rules apply to privileged code as well, and therefore
-    /// the MPU must be completely disabled for the kernel to effectively manage
-    /// processes.
-    fn disable_mpu(&self) {}
+    /// some of the MPU configuration must be disabled for the kernel to effectively
+    /// manage processes.
+    fn disable_app_mpu(&self) {}
 
     /// Returns the maximum number of regions supported by the MPU.
     fn number_total_regions(&self) -> usize {
--- a/kernel/src/platform/mpu.rs
+++ b/kernel/src/platform/mpu.rs
@@ -266,3 +266,76 @@ pub trait MPU {
 
 /// Implement default MPU trait for unit.
 impl MPU for () {}
+
+/// The generic trait that particular kernel level memory protection unit
+/// implementations need to implement.
+///
+/// This trait provides generic functionality to extend the MPU trait above
+/// to also allow the kernel to protect itself. It is expected that only a
+/// limited number of SoCs can support this, which is why it is a seperate
+/// implementation.
+pub trait KernelMPU {
+    /// MPU-specific state that defines a particular configuration for the kernel
+    /// MPU.
+    /// That is, this should contain all of the required state such that the
+    /// implementation can be passed an object of this type and it should be
+    /// able to correctly and entirely configure the MPU.
+    ///
+    /// It is `Default` so we can create empty state when the kernel is
+    /// created, and `Display` so that the `panic!()` output can display the
+    /// current state to help with debugging.
+    type KernelMpuConfig: Default + Display = MpuConfigDefault;
+
+    /// Mark a region of memory that the Tock kernel owns.
+    ///
+    /// This function will optionally set the MPU to enforce the specified
+    /// constraints for all accessess (even from the kernel).
+    /// This should be used to mark read/write/execute areas of the Tock
+    /// kernel to have the hardware enforce those permissions.
+    ///
+    /// If the KernelMPU trait is supported a board should use this function
+    /// to set permissions for all areas of memory the kernel will use.
+    /// Once all regions of memory have been allocated, the board must call
+    /// enable_kernel_mpu(). After enable_kernel_mpu() is called no changes
+    /// to kernel level code permissions can be made.
+    ///
+    /// Note that kernel level permissions also apply to apps, although apps
+    /// will have more constraints applied on top of the kernel ones as
+    /// specified by the `MPU` trait.
+    ///
+    /// Not all architectures support this, so don't assume this will be
+    /// implemented.
+    ///
+    /// # Arguments
+    ///
+    /// - `memory_start`:             start of memory region
+    /// - `memory_size`:              size of unallocated memory
+    /// - `permissions`:              permissions for the region
+    /// - `config`:                   MPU region configuration
+    ///
+    /// # Return Value
+    ///
+    /// Returns the start and size of the requested memory region. If it is
+    /// infeasible to allocate the MPU region, returns None.
+    #[allow(unused_variables)]
+    fn allocate_kernel_region(
+        &self,
+        memory_start: *const u8,
+        memory_size: usize,
+        permissions: Permissions,
+        config: &mut Self::KernelMpuConfig,
+    ) -> Option<Region>;
+
+    /// Enables the MPU for the kernel.
+    ///
+    /// This function must enable the permission restrictions on the various
+    /// kernel regions specified by `allocate_kernel_region()` protected by
+    /// the MPU.
+    ///
+    /// It is expected that this function is called in `reset_handler()`.
+    ///
+    /// Once enabled this cannot be disabled. It is expected there won't be any
+    /// changes to the kernel regions after this is enabled.
+    #[allow(unused_variables)]
+    fn enable_kernel_mpu(&self, config: &mut Self::KernelMpuConfig);
+}
--- a/kernel/src/platform/mpu.rs
+++ b/kernel/src/platform/mpu.rs
@@ -195,7 +195,7 @@ pub trait MPU {
     /// This function returns the start address and the size of the memory block
     /// chosen for the process. If it is infeasible to find a memory block or
     /// allocate the MPU region, or if the function has already been called,
-    /// returns None.
+    /// returns None. If None is returned no changes are made.
     #[allow(unused_variables)]
     fn allocate_app_memory_region(
         &self,
@@ -234,7 +234,8 @@ pub trait MPU {
     /// # Return Value
     ///
     /// Returns an error if it is infeasible to update the MPU region, or if it
-    /// was never created.
+    /// was never created. If an error is returned no changes are made to the
+    /// configuration.
     #[allow(unused_variables)]
     fn update_app_memory_region(
         &self,
@@ -316,7 +317,8 @@ pub trait KernelMPU {
     /// # Return Value
     ///
     /// Returns the start and size of the requested memory region. If it is
-    /// infeasible to allocate the MPU region, returns None.
+    /// infeasible to allocate the MPU region, returns None. If None is
+    /// returned no changes are made.
     #[allow(unused_variables)]
     fn allocate_kernel_region(
         &self,
--- a/arch/cortex-m3/src/mpu.rs
+++ b/arch/cortex-m3/src/mpu.rs
@@ -363,7 +363,12 @@ impl CortexMRegion {
 impl kernel::mpu::MPU for MPU {
     type MpuConfig = CortexMConfig;
 
-    fn enable_mpu(&self) {
+    fn clear_mpu(&self) {
+        let regs = &*self.registers;
+        regs.ctrl.write(Control::ENABLE::CLEAR);
+    }
+
+    fn enable_app_mpu(&self) {
         let regs = &*self.registers;
 
         // Enable the MPU, disable it during HardFault/NMI handlers, and allow
@@ -372,9 +377,9 @@ impl kernel::mpu::MPU for MPU {
             .write(Control::ENABLE::SET + Control::HFNMIENA::CLEAR + Control::PRIVDEFENA::SET);
     }
 
-    fn disable_mpu(&self) {
-        let regs = &*self.registers;
-        regs.ctrl.write(Control::ENABLE::CLEAR);
+    fn disable_app_mpu(&self) {
+        // The MPU is not enabled for privileged mode, so we don't have to do
+        // anything
     }
 
     fn number_total_regions(&self) -> usize {
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -244,9 +244,7 @@ impl PMPConfig {
 impl kernel::mpu::MPU for PMP {
     type MpuConfig = PMPConfig;
 
-    fn enable_mpu(&self) {}
-
-    fn disable_mpu(&self) {
+    fn clear_mpu(&self) {
         // We want to disable all of the hardware entries, so we use `$x` here,
         // and not `$x / 2`.
         for x in 0..$x {
@@ -299,8 +297,15 @@ impl kernel::mpu::MPU for PMP {
         csr::CSR.pmpcfg[0].modify(csr::pmpconfig::pmpcfg::w0::SET);
         csr::CSR.pmpcfg[0].modify(csr::pmpconfig::pmpcfg::x0::SET);
         csr::CSR.pmpcfg[0].modify(csr::pmpconfig::pmpcfg::a0::TOR);
-        // MPU is not configured for any process now
+        // PMP is not configured for any process now
         self.last_configured_for.take();
+    }
+
+    fn enable_app_mpu(&self) {}
+
+    fn disable_app_mpu(&self) {
+        // PMP is not enabled for machine mode, so we don't have to do
+        // anything
     }
 
     fn number_total_regions(&self) -> usize {
--- a/boards/nano33ble/src/main.rs
+++ b/boards/nano33ble/src/main.rs
@@ -266,7 +266,7 @@ pub unsafe fn reset_handler() {
     CHIP = Some(chip);
 
     // Need to disable the MPU because the bootloader seems to set it up.
-    chip.mpu().disable_mpu();
+    chip.mpu().clear_mpu();
 
     // Configure the USB stack to enable a serial port over CDC-ACM.
     cdc.enable();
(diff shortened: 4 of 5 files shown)


alistair23:
I pushed an update that splits out the KernelMPU and fixes some comments.

alistair23:
Ping!

kernel/src/platform/mpu.rs
```diff
@@ -257,3 +267,70 @@ pub trait MPU {
...
+/// limited number of SoCs can support this, which is why it is a seperate
+/// implementation.
+pub trait KernelMPU {
+    /// MPU-specific state that defines a particular configuration for the kernel
+    /// MPU.
+    /// That is, this should contain all of the required state such that the
+    /// implementation can be passed an object of this type and it should be
+    /// able to correctly and entirely configure the MPU.
+    ///
+    /// It is `Default` so we can create empty state when the process is
```
ppannuto, line 285:
```suggestion
    /// It is `Default` so we can create empty state when the kernel is
```

?

kernel/src/platform/mpu.rs
```diff
@@ -257,3 +267,70 @@ pub trait MPU {
...
+    /// This should be used to mark read/write/execute only areas of the Tock
+    /// kernel to have the hardware enforce those permissions.
+    ///
+    /// If using this function, all kernel executable code must be marked as
+    /// executable prior to enabling the MPU (by calling `enable_mpu()).
+    /// Also all kernel read/write regions must be marked. That is any
+    /// unmarked kernel regions will cause a fault if accesses.
+    ///
+    /// Not all architectures support this, so don't assume this will be
+    /// implemented.
```
ppannuto, line 303: FWIW, ARM *could* support this. We simply choose to turn the MPU off on context switch to the kernel for simplicity, but the context switch code could reconfigure the MPU for 'kernel-mode' before finishing the switch -- we'd just have to decide if that's worth the overhead on context switches.

kernel/src/platform/mod.rs
```diff
@@ -111,6 +112,13 @@ pub trait Chip {
     /// chip and resumes the scheduler.
     fn sleep(&self);
 
+    /// Called at startup to reset any configuration from previous stages.
```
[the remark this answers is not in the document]
alistair23: So if a board doesn't call it then we are still in the same position as we are today (where it isn't called). The idea of calling it is that then we know we are starting with a clean slate.

I don't think it can go in `tock-rt0` as it needs the full PMP/MPU implementation.

kernel/src/platform/mod.rs
```diff
@@ -111,6 +112,13 @@ pub trait Chip {
...
+    /// On some hardware it is impossible to reset the MPU after it has
+    /// been locked, in this case this function wont change those regions.
+    fn clear_mpu(&self) {
+        self.mpu().clear_mpu()
+    }
```
bradjc, line 120: Boards can do this if they need to, but I don't think we need to change this interface.

phil-levis:
I will take a look by end of week.

kernel/src/platform/mod.rs
phil-levis: I agree with Brad -- since this would be in `main` and is board/use case specific, it doesn't and shouldn't be in the trait.

kernel/src/platform/mod.rs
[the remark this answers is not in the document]
phil-levis: Right -- some boards need to call it because their initialization sequence has MPU state to clean out. But this suggests it shouldn't be in the chip-independent trait. The trait is for what chip-independent (e.g., kernel) code needs to do. The MPU implementations can always have additional functionality.

kernel/src/platform/mpu.rs
```diff
@@ -93,20 +93,30 @@ pub trait MPU {
...
-    fn enable_mpu(&self) {}
+    fn enable_app_mpu(&self) {}
 
-    /// Disables the MPU.
+    /// Disables the MPU for userspace apps.
     ///
-    /// This function must completely disable any access control enforced by the
-    /// MPU. This will be called before the kernel starts to execute as on some
+    /// This function must disable any access control that was previously setup
+    /// for an app if it will interfere with the kernel.
```
phil-levis, line 114: I don't understand what this comment means. It seems to contradict the second part (unmodified) of the comment. What does this do? What is "access control that was previously setup for an app if it will interfere with the kernel?" Does this mean "give access to kernel memory"?

kernel/src/platform/mpu.rs
```diff
@@ -257,3 +267,70 @@ pub trait MPU {
...
+        memory_size: usize,
+        permissions: Permissions,
+        config: &mut Self::KernelMpuConfig,
+    ) -> Option<Region> {
+        None
+    }
+
+    /// Enables the MPU for the kernel.
+    ///
+    /// This function must enable the permission restrictions on the various
```
phil-levis, line 329: I don't understand the semantics of this operation. Aren't these regions disabled while in userspace? The comment "cannot be disabled" confuses me.

alistair23: No, these will always be enabled.

The idea is that the kernel can specify that this region of memory is read/write/execute for the kernel. All other accesses are blocked, which includes user space. Once enabled this can't be disabled by either the kernel or the app.

The kernel should allow itself access to what it needs, then lock out everything else from itself and apps.

kernel/src/platform/mpu.rs
alistair23: The idea is to "undo" any app specific changes ONLY if they interfere with the running of the kernel.

So if we support some new architecture that requires disabling the MPU before we can run the rest of the kernel then we should do that here.

kernel/src/platform/mod.rs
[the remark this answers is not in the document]
alistair23: I have removed the mention of calling it at boot.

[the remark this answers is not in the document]
alistair23: Note, that I have still left this function in the trait as I think it's useful. If people feel strongly I can remove it, but it seems worth having, then `Chips` can call it as required.

kernel/src/platform/mpu.rs
alistair23: Ah that's cool

kernel/src/platform/mod.rs
alistair23: I have removed all the calls to this, but left it in the trait.

If it's not in the trait we will end up implementing it anyway and just making it public so that chpis/boards can access it.

alistair23:
I just pushed an update. I have tried to improve the documentation.

I removed the boards clearing the MPU. I have left the `clear_mpu()` function in the trait as I think it is very useful to have. If we remove it from the trait we will just end up implementing the function publicly in the MPU implementations anyway.

Let me know what you think.

alistair23:
Ping!

alistair23:
Ping^2

@phil-levis  or @bradjc any comments?

alistair23:
Ping^3

kernel/src/platform/mpu.rs
```diff
@@ -257,3 +266,75 @@ pub trait MPU {
...
+    /// kernel regions specified by `allocate_kernel_region()` protected by
+    /// the MPU.
+    /// Once enabled this cannot be disabled. It is expected there won't be any
+    /// changes to the kernel regions after this is enabled.
+    #[allow(unused_variables)]
+    fn enable_kernel_mpu(&self, config: &mut Self::KernelMpuConfig) {}
```
hudson-ayers, line 339: This PR does not add any calls to the `KernelMPU` trait anywhere, so I assume that it is intended that setting up and enabling the `KernelMPU` will be done from `reset_handler()`. Accordingly, maybe there is no need for default implementations of this trait -- if you don't want to use it, then don't implement it, as the Kernel does not depend on it existing. Also, maybe it would be good to add a comment instructing users of the trait to ensure all calls to the implementation occur in `reset_handler()`.

alistair23: Yes, it will be up to a specific board to call enable_kernel_mpu(). The reason is that this only needs to be called once and it will be architecture dependent on when this can be called.

Ok, I'll update that.

hudson-ayers (dismissed):

alistair23:
Can this be merged?

alistair23:
Ping again

hudson-ayers:
cc @phil-levis bors can't merge this until you remove your request for changes

phil-levis (dismissed):
Looks good to me.

alistair23:
I just pushed a patch adding the comments discussed in the OT call.

Can we re-run bors?

hudson-ayers (approved):

ia0:
Hi @alistair23,

Is there an issue I could follow tracking the implementation of `KernelMpuConfig` for Cortex-M? In particular, I believe this PR breaks support for writable flash regions for boards with direct write to the flash like for the nRF52 family, because the kernel cannot write to the flash now that the MPU is left enabled when transitioning from app back to kernel (i.e. using `disable_app_mpu` instead of `clear_mpu`).

Thanks!

alistair23:
Hey @ia0. I am not doing anything related to `KernelMpuConfig` for Cortex-M. My focus is extending Tock to support RISC-V ePMP using `KernelMpuConfig`.

Can you point to documentation that describes the issue? Can you also point to a way to reproduce it?

You are saying that because we no longer do `regs.ctrl.write(Control::ENABLE::CLEAR);` after returning from an app that the kernel can't write to direct write flash? A partial revert of https://github.com/tock/tock/pull/1873/commits/128782dad5ad410d3e48c16aadba60d9ed6d842b should fix this right?

ia0:
> A partial revert of 128782d should fix this right?

Yes, this is essentially what we currently do as a workaround until we can setup a kernel MPU config.

> Can you also point to a way to reproduce it?

I'll try to create a "minimal" reproduction case when I get time.

> I am not doing anything related to KernelMpuConfig for Cortex-M.

I see. However, we might need some work in that direction to avoid the regression. Implementing `KernelMpuConfig` should probably not be complicated for someone familiar with the architecture. And updating Tock to configure the Kernel MPU according to the writable flash regions should also not be complicated (I might be able to do it if I look into it).

ppannuto:
@tock/core-wg this sounds like it could be a regression that we may want to fix before a 1.6 release? @alistair23 or @ia0, would it be possible to get the partial revert fix into a PR relatively quickly?

alistair23:
The `disable_app_mpu()` function "must be disabled for the kernel to effectively manage processes", so if it is not doing that we should fix the function for Cortex-M. I don't think we need any changes or implementation of `KernelMpuConfig` for this.

We could either clear the enable bit when returning to the kernel (as we did previously) or we could change the MPU permissions to only affect userspace. The second option matches what we do for RISC-V, so I think that's better.

Something like this diff:

```diff
diff --git a/arch/cortex-m3/src/mpu.rs b/arch/cortex-m3/src/mpu.rs
index 004d52194..729a65977 100644
--- a/arch/cortex-m3/src/mpu.rs
+++ b/arch/cortex-m3/src/mpu.rs
@@ -275,14 +275,14 @@ impl CortexMRegion {
                 RegionAttributes::XN::Disable,
             ),
             mpu::Permissions::ReadExecuteOnly => {
-                (RegionAttributes::AP::ReadOnly, RegionAttributes::XN::Enable)
+                (RegionAttributes::AP::UnprivilegedReadOnly, RegionAttributes::XN::Enable)
             }
             mpu::Permissions::ReadOnly => (
-                RegionAttributes::AP::ReadOnly,
+                RegionAttributes::AP::UnprivilegedReadOnly,
                 RegionAttributes::XN::Disable,
             ),
             mpu::Permissions::ExecuteOnly => {
-                (RegionAttributes::AP::NoAccess, RegionAttributes::XN::Enable)
+                (RegionAttributes::AP::PrivilegedOnly, RegionAttributes::XN::Enable)
             }
         };
 ```

ia0:
I confirm that the diff in https://github.com/tock/tock/pull/1873#issuecomment-696354110 fixes the issue.

It would take some work to make a minimal or even small reproduction case. But here are the reproduction steps for the actual issue:
- Install OpenSK on a nRF52840:
    ```sh
    git clone git@github.com:jmichelp/OpenSK.git
    cd OpenSK
    rm patches/tock/05-mpu-fix.patch
    ./setup.sh
    # The following line assumes a dev-kit board is plugged and configured
    ./deploy.py --board=nrf52840dk --opensk --panic-console --clear-storage --programmer=jlink
    ```
- Look at RTT debugging:
    ```sh
    JLinkExe -device nrf52840_xxAA -if swd -speed 1000 -AutoConnect 1
    # In another shell
    JLinkRTTClientExe
    ```
- Press the reset button of the dev-kit and look at the kernel hardfault due to data access violation:
    ```
    panicked at 'Kernel HardFault.
    	Kernel version release-1.5-1043-gc5b7a4f2c
    	r0  0xffff8000
    	r1  0xffffffff
    	r2  0xffff8000
    	r3  0x4
    	r12 0x2003fb90
    	lr  0xc99d
    	pc  0xc9b4
    	prs 0x81001800 [ N 1 Z 0 C 0 V 0 Q 0 GE 0000 ; ICI.IT 6 T true ; Exc 0-Thread Mode ]
    	sp  0x20000a88
    	top of stack     0x20001000
    	bottom of stack  0x20000000
    	SHCSR 0x0
    	CFSR  0x82
    	HSFR  0x40000000
    	Instruction Access Violation:       false
    	Data Access Violation:              true
    	Memory Management Unstacking Fault: false
    	Memory Management Stacking Fault:   false
    	Memory Management Lazy FP Fault:    false
    	Instruction Bus Error:              false
    	Precise Data Bus Error:             false
    	Imprecise Data Bus Error:           false
    	Bus Unstacking Fault:               false
    	Bus Stacking Fault:                 false
    	Bus Lazy FP Fault:                  false
    	Undefined Instruction Usage Fault:  false
    	Invalid State Usage Fault:          false
    	Invalid PC Load Usage Fault:        false
    	No Coprocessor Usage Fault:         false
    	Unaligned Access Usage Fault:       false
    	Divide By Zero:                     false
    	Bus Fault on Vector Table Read:     false
    	Faulting Memory Address: (valid: true) 0x000C0000
    	Bus Fault Address:       (valid: false) 0x000C0000
    ', arch/cortex-m/src/lib.rs:300:5
    	Kernel version release-1.5-1043-gc5b7a4f2c

    ---| Fault Status |---
    No faults detected.

    ---| App Status |---
    App: ctap2   -   [Running]
     Events Queued: 0   Syscall Count: 18   Dropped Callback Count: 0
     Restart Count: 0
     Last Syscall: Some(COMMAND { driver_number: 327683, subdriver_number: 2, arg0: 786432, arg1: 4 })

     ╚0x20040000═╪══════════════════════════════════════════╝
                 │ ▼ Grant        1156 |   1156
      0x2003FB7C ┼───────────────────────────────────────────
                 │ Unused
      0x20039FA4 ┼───────────────────────────────────────────
                 │ ▲ Heap        90000 | 113512               S
      0x20024014 ┼─────────────────────────────────────────── R
                 │ Data             20 |     20               A
      0x20024000 ┼─────────────────────────────────────────── M
                 │ ▼ Stack        1008 |  16384
      0x20023C10 ┼───────────────────────────────────────────
                 │ Unused
      0x20020000 ┴───────────────────────────────────────────
                 .....
      0x00080000 ┬─────────────────────────────────────────── F
                 │ App Flash    262080                        L
      0x00040040 ┼─────────────────────────────────────────── A
                 │ Protected        64                        S

      R0 : 0x00050003    R6 : 0x00000001
      R1 : 0x00000002    R7 : 0x20023C48
      R2 : 0x000C0000    R8 : 0x20024014
      R3 : 0x00000004    R10: 0x000C0000
      R4 : 0x00000000    R11: 0x00000000
      R5 : 0x000C0000    R12: 0x00000004
      R9 : 0xFFFFFFF8 (Static Base Register)
      SP : 0x20023C10 (Process Stack Pointer)
      LR : 0x00000003
      PC : 0x000513FA
     YPC : 0x000513FA

     APSR: N 0 Z 0 C 0 V 0 Q 0
           GE 0 0 0 0
     EPSR: ICI.IT 0x00
           ThumbBit true

     Cortex-M MPU
      Region 0: [0x20020000:0x20040000], length: 131072 bytes; ReadWrite (0x3)
        Sub-region 0: [0x20020000:0x20024000], Enabled
        Sub-region 1: [0x20024000:0x20028000], Enabled
        Sub-region 2: [0x20028000:0x2002C000], Enabled
        Sub-region 3: [0x2002C000:0x20030000], Enabled
        Sub-region 4: [0x20030000:0x20034000], Enabled
        Sub-region 5: [0x20034000:0x20038000], Enabled
        Sub-region 6: [0x20038000:0x2003C000], Enabled
        Sub-region 7: [0x2003C000:0x20040000], Disabled
      Region 1: [0x00040000:0x00080000], length: 262144 bytes; ReadOnly (0x6)
        Sub-region 0: [0x00040000:0x00048000], Enabled
        Sub-region 1: [0x00048000:0x00050000], Enabled
        Sub-region 2: [0x00050000:0x00058000], Enabled
        Sub-region 3: [0x00058000:0x00060000], Enabled
        Sub-region 4: [0x00060000:0x00068000], Enabled
        Sub-region 5: [0x00068000:0x00070000], Enabled
        Sub-region 6: [0x00070000:0x00078000], Enabled
        Sub-region 7: [0x00078000:0x00080000], Enabled
      Region 2: [0x000C0000:0x00100000], length: 262144 bytes; ReadOnly (0x6)
        Sub-region 0: [0x000C0000:0x000C8000], Enabled
        Sub-region 1: [0x000C8000:0x000D0000], Enabled
        Sub-region 2: [0x000D0000:0x000D8000], Enabled
        Sub-region 3: [0x000D8000:0x000E0000], Enabled
        Sub-region 4: [0x000E0000:0x000E8000], Enabled
        Sub-region 5: [0x000E8000:0x000F0000], Enabled
        Sub-region 6: [0x000F0000:0x000F8000], Enabled
        Sub-region 7: [0x000F8000:0x00100000], Enabled
      Region 3: Unused
      Region 4: Unused
      Region 5: Unused
      Region 6: Unused
      Region 7: Unused
    ```
Address `0xC0000` is where OpenSK has its persistent storage on nRF52840 boards. After reset, it tries to initialize the storage by writing some metadata.

alistair23:
PR created: https://github.com/tock/tock/pull/2120

PR #1879, opened 2020-05-26
OpenTitan: Initial support for I2C

alistair23:
### Pull Request Overview

This is the initial support for an I2C master, based on the OT documentation here: https://docs.opentitan.org/hw/ip/i2c/doc/index.html

### Testing Strategy

None. The I2C master device doesn't have an address in OT (at least not that [I can find](https://docs.opentitan.org/hw/top_earlgrey/doc/)) so I'm assuming it's not actually connected.

Also the I2C bus appears to not be broken out of the FPGA, but instead just connects to the audio IC ([page 28](https://reference.digilentinc.com/_media/nexys-video:nexys_video_rm.pdf)), so I can't test it on an already supported device or connect an LA.

### TODO or Help Wanted

 - Test when a HW implementation is available
 - Add support for SMBus (or at least test with SMBus devices)

### Documentation Updated

### Formatting

- [X] Ran `make format`.

### DIFF
--- a/chips/ibex/src/i2c.rs
+++ b/chips/ibex/src/i2c.rs
@@ -1,0 +1,9 @@
+use crate::chip;
+use kernel::common::StaticRef;
+use lowrisc::i2c::{I2c, I2cRegisters};
+
+pub static mut I2C: I2c = I2c::new(I2C_BASE, (1 / chip::CHIP_FREQ) * 1000 * 1000);
+
+// This is a placeholder address as the I2C MMIO interface isn't avaliable yet
+const I2C_BASE: StaticRef<I2cRegisters> =
+    unsafe { StaticRef::new(0x4005_0000 as *const I2cRegisters) };
--- a/chips/lowrisc/src/i2c.rs
+++ b/chips/lowrisc/src/i2c.rs
@@ -1,0 +1,496 @@
+//! I2C Master Driver
+
+use core::cell::Cell;
+use kernel::common::cells::OptionalCell;
+use kernel::common::cells::TakeCell;
+use kernel::common::registers::{
+    register_bitfields, register_structs, ReadOnly, ReadWrite, WriteOnly,
+};
+use kernel::common::StaticRef;
+use kernel::hil;
+use kernel::hil::i2c;
+
+register_structs! {
+    pub I2cRegisters {
+        (0x00 => intr_state: ReadWrite<u32, INTR::Register>),
+        (0x04 => intr_enable: ReadWrite<u32, INTR::Register>),
+        (0x08 => intr_test: WriteOnly<u32, INTR::Register>),
+        (0x0C => ctrl: ReadWrite<u32, CTRL::Register>),
+        (0x10 => status: ReadOnly<u32, STATUS::Register>),
+        (0x14 => rdata: ReadOnly<u32, RDATA::Register>),
+        (0x18 => fdata: WriteOnly<u32, FDATA::Register>),
+        (0x1C => fifo_ctrl: ReadWrite<u32, FIFO_CTRL::Register>),
+        (0x20 => fifo_status: ReadOnly<u32, FIFO_STATUS::Register>),
+        (0x24 => ovrd: ReadWrite<u32, OVRD::Register>),
+        (0x28 => val: ReadOnly<u32, VAL::Register>),
+        (0x2C => timing0: ReadWrite<u32, TIMING0::Register>),
+        (0x30 => timing1: ReadWrite<u32, TIMING1::Register>),
+        (0x34 => timing2: ReadWrite<u32, TIMING2::Register>),
+        (0x38 => timing3: ReadWrite<u32, TIMING3::Register>),
+        (0x3C => timing4: ReadWrite<u32, TIMING4::Register>),
+        (0x40 => timeout_ctrl: ReadWrite<u32, TIMEOUT_CTRL::Register>),
+        (0x44 => @END),
+    }
+}
+
+register_bitfields![u32,
+    INTR [
+        FMT_WATERMARK OFFSET(0) NUMBITS(1) [],
+        RX_WATERMARK OFFSET(1) NUMBITS(1) [],
+        FMT_OVERFLOW OFFSET(2) NUMBITS(1) [],
+        RX_OVERFLOW OFFSET(3) NUMBITS(1) [],
+        NAK OFFSET(4) NUMBITS(1) [],
+        SCL_INTERFERENCE OFFSET(5) NUMBITS(1) [],
+        SDA_INTERFERENCE OFFSET(6) NUMBITS(1) [],
+        STRETCH_TIMEOUT OFFSET(7) NUMBITS(1) [],
+        SDA_UNSTABLE OFFSET(8) NUMBITS(1) []
+    ],
+    CTRL [
+        ENABLEHOST OFFSET(0) NUMBITS(1) []
+    ],
+    STATUS [
+        FMTFULL OFFSET(0) NUMBITS(1) [],
+        RXFULL OFFSET(1) NUMBITS(1) [],
+        FMTEMPTY OFFSET(2) NUMBITS(1) [],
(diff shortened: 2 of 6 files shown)


ppannuto:
> Add support for SMBus (or at least test with SMBus devices)

In practice, SMBus is largely just a protocol layer atop the I2C PHY (it does change some rules around ACK/NAK and address response, but nothing that we use in Tock anyway). I'd imagine we would implement this as a generic SMBus capsule that sits atop the I2C capsule, no?

chips/ibex/src/i2c.rs
```diff
@@ -0,0 +1,8 @@
+use kernel::common::StaticRef;
+use lowrisc::i2c::{I2c, I2cRegisters};
+
+pub static mut I2C: I2c = I2c::new(I2C_BASE);
+
+// Random guess, this isn't documents
```
ppannuto, line 6: Random guess that's working on FPGA image, or just a placeholder address?

chips/lowrisc/src/i2c.rs
```diff
@@ -0,0 +1,275 @@
+//! SHA256 HMAC (Hash-based Message Authentication Code).
```
ppannuto, line 1: Update header :)

chips/lowrisc/src/i2c.rs
```diff
@@ -0,0 +1,275 @@
...
+        TSU_STO OFFSET(0) NUMBITS(16) [],
+        T_BUF OFFSET(16) NUMBITS(16) []
+    ],
+    TIMEOUT_CTRL [
+        VAL OFFSET(0) NUMBITS(31) [],
+        EN OFFSET(31) NUMBITS(1) []
+    ]
+];
+
+pub struct I2c<'a> {
```
ppannuto, line 125: It looks like Tock writes this `i2c` everywhere else; is there a reason for different capitalization here?

chips/lowrisc/src/i2c.rs
```diff
@@ -0,0 +1,275 @@
...
+    pub fn handle_interrupt(&self) {
+        unimplemented!();
+    }
+
+    fn timing_parameter_init(&self) {
+        let regs = self.registers;
+
+        // Setup the timing variables
+        // This is taken from the OT documentation example
+        // Assume a desired datarate of 1 Mbaud (the bus maximum) for an SCL
```
ppannuto, line 148: Hmm.. 400 KHz is standard for I2C (well, strictly speaking 100 KHz is standard, 400 KHz is "fast mode", 1 MHz is "fast mode plus", and 3.2 MHz is "ultra fast"), but in practice unlabeled I2C is usually "fast mode".

Does the I2C HIL not expose bus speed control? (It probably should..). But I think 400 KHz would be a better default for common-case compatibility.

chips/lowrisc/src/i2c.rs
```diff
@@ -0,0 +1,275 @@
...
+        let read_addr = addr | 1;
+
+        // Set the start condition and the address
+        regs.fdata
+            .write(FDATA::START::SET + FDATA::FBYTE.val(read_addr as u32));
+
+        for i in 0..len {
+            // Read the data
+            buffer[i as usize] = regs.rdata.read(RDATA::RDATA) as u8;
+        }
```
bradjc, line 273: Will this actually work? I would think it would read much faster than bytes are received.

chips/ibex/src/i2c.rs
alistair23: It's just a place holder. Although there is I2C RTL it doesn't seem to be exposed via MMIO in OT yet.

chips/lowrisc/src/i2c.rs
alistair23: I was just following CamelCase for the struct names. Should I change it to all lower case?

chips/ibex/src/i2c.rs
ppannuto: Ah, so this fully isn't expected to work yet. Okay, for OT that's understandable. Can this comment be updated to reflect that this is a placeholder and there's not official MMIO interface (yet)?

chips/lowrisc/src/i2c.rs
ppannuto: Sigh. I don't _really_ care that much. This is firmly on the list of things I want rustfmt to decide for us, I was just surprised that it was allowed to be different than elsewhere in Tock.

Leaving it camel case to follow Rust conventions is probably fine.

chips/ibex/src/i2c.rs
alistair23: Done

alistair23:
I have updated this to use 400kHz I2C and be non-blocking.

chips/lowrisc/src/i2c.rs
```diff
@@ -0,0 +1,494 @@
...
+    ]
+];
+
+pub struct I2c<'a> {
+    registers: StaticRef<I2cRegisters>,
+    clock_period_nanos: u32,
+
+    master_client: OptionalCell<&'a dyn hil::i2c::I2CHwMasterClient>,
+
+    // Set whe calling the write_read operation
```
ppannuto, line 133:
```suggestion
    // Set when the write_read operation is called
```

chips/lowrisc/src/i2c.rs
```diff
@@ -0,0 +1,494 @@
...
+                + INTR::RX_OVERFLOW::SET
+                + INTR::NAK::SET
+                + INTR::SCL_INTERFERENCE::SET
+                + INTR::SDA_INTERFERENCE::SET
+                + INTR::STRETCH_TIMEOUT::SET
+                + INTR::SDA_UNSTABLE::SET,
+        );
+
+        if irqs & (1 << 0) != 0 {
+            // FMT Watermark
```
ppannuto, line 177: Is there a reason for using bare constants here rather than the register interface?

chips/lowrisc/src/i2c.rs
```diff
@@ -0,0 +1,494 @@
...
+            if self.address.get() != 0 {
+                self.write_read_data();
+            } else {
+                self.write_data();
+            }
+        }
+
+        if irqs & (1 << 1) != 0 {
+            // RX Watermark
```
ppannuto, line 186: Bare constants here too?

chips/lowrisc/src/i2c.rs
```diff
@@ -0,0 +1,494 @@
...
+    }
+
+    fn read_data(&self) {
+        let regs = self.registers;
+        let mut data_popped = self.read_index.get();
+        let len = self.read_len.get();
+
+        self.buffer.map(|buf| {
+            for i in data_popped..len {
+                if regs.status.read(STATUS::RXEMPTY) != 0 {
```
ppannuto, line 227:
```suggestion
                if regs.status.is_set(STATUS::RXEMPTY) {
```

Maybe?

chips/lowrisc/src/i2c.rs
ppannuto: Actually, I don't think I quite understand what `address` this is referring to. Is this the address of this device when it's acting as a slave? Or is this the address on the slave device that's being read from? (I think the latter?) A clearer name might help here.

chips/lowrisc/src/i2c.rs
alistair23: There isn't, I just didn't know how to use the register interface on a stored value.

chips/lowrisc/src/i2c.rs
alistair23: I have changed it to `slave_read_address` and added a better comment.

chips/lowrisc/src/i2c.rs
hudson-ayers: `let irqs = regs.intr_state.extract();`

alistair23:
Addressed all comments

ppannuto:
Oh; well the changes look good. 👍 when that merge conflict is resolved.

alistair23:
Rebased

PR #1872, opened 2020-05-21
[RFC] Codify CI policy

ppannuto:
### Pull Request Overview

This takes a first cut at writing down a structure and policy for CI.

### Testing Strategy

N/A.

### TODO or Help Wanted

Feedback!

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [ ] Ran `make format`.
- [ ] Fixed errors surfaced by `make clippy`.

---

[Rendered](https://github.com/tock/tock/blob/ci-policy/doc/CodeReview.md#3-continuous-integration)

Issue #1865, opened 2020-05-19
Make emulation-check rule is broken


gendx:
As mentioned in https://github.com/tock/tock/pull/1861#discussion_r427336544, the `make emulation-check` is broken when relevant Debian packages are missing.

As a newcomer, it's unclear what to do after the following steps.

```
$ make emulation-check 
Cloning into 'qemu'...
remote: Enumerating objects: 19007, done.
remote: Counting objects: 100% (19007/19007), done.
remote: Compressing objects: 100% (6200/6200), done.
remote: Total 482048 (delta 14287), reused 16209 (delta 12647), pack-reused 463041
Receiving objects: 100% (482048/482048), 195.57 MiB | 27.38 MiB/s, done.
Resolving deltas: 100% (394231/394231), done.
Checking connectivity... done.

ERROR: pkg-config binary 'pkg-config' not found

Makefile:219: recipe for target 'emulation-setup' failed
make: *** [emulation-setup] Error 1

$ sudo apt install pkg-config
...

$ make emulation-check 
make[1]: *** [config-host.mak] Error 1
Makefile:219: recipe for target 'emulation-setup' failed
make: *** [emulation-setup] Error 2
```

alistair23:
The solution seems to be just to remove the pipe to `/dev/null`. Then you will get a nice error that can be fixed.

I'll send a patch

gendx:
Replying here to https://github.com/tock/tock/pull/1861#issuecomment-630895222, given that it's not on-topic for that pull request anymore.

> A few things here:
> 
> > ```
> > 1. I find it surprising that it automatically tries to install qemu "behind my back".
> > ```
> 
> QEMU is never installed. There is a local clone, but never an install. I never want random scripts to install things on my machine.

I don't think we have the same definition of "install". Sure, this isn't installed as `root` in a system folder like `/usr/bin`, but it's still some installation on my current folder with my user's privileges. In particular, the `./configure` script could do anything out of my control.

To me, the following facts are problematic.
- This qemu repository contains hundreds of MB. I consider this amount to be an installation. I don't consider it essential to regular Tock development, so there should be a prompt for me to confirm that I want to install it.
[...]

PR #1771, opened 2020-04-18
boards/Makefile.common: tool-chain customization

vsukhoml:
Added more flexible customization of toolchain options:

1) Defined TARGET_DIRECTORY variable to move build artifacts to
    desired location. Default is same - tock/target. Allows to have build outside repository, or at local 
   directory if repository is accessed via sshfs or NFS.
2) Add OBJCOPY_FLAGS, SIZE_FLAGS and CARGO_FLAGS to allow boards to use board- 
    specific options in generic way rather than having special hacks in common makefile or use
    proprietary makefiles.
3) Add CARGO_OPTIONS from #1741 to CARGO_FLAGS to unify both approaches.
4) OBJDUMP_FLAGS completely re-definable by board, also allows to add
    board specific like --arch-type
5) Enabled --demangle by default for OBJDUMP, though it doesn't work
    as expected in llvm-objdump (but you can always use objdump from gcc)
6) split RUSTFLAGS_FOR_CARGO into RUSTFLAGS_FOR_BOARD
    Separate settings which are likely to be configurable by board from settings specific to 
    Makefile.common, so board makefiles will less likely require change if any changes done by 
    boards/Makefile.common

### Pull Request Overview

This pull request adds means to customize tool-chain flags for boards
without introducing changes in Makefile.common or creating board-
specific build rules.

### Testing Strategy

This pull request was tested by building all boards. Default settings
didn't change. 
make ci-netlify

### TODO or Help Wanted

This pull request still needs boards to start using customization and to remove special handling from Makefile.common.

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

PR #1880, opened 2020-05-27
Overhaul Tock CI

ppannuto:
### Pull Request Overview

This implements the policy proposed in #1872.

This includes a number of ergonomics and reliability enhancements to pieces of the CI infrastructure as well.

### Testing Strategy

Lots of local building, and now this PR is testing the CI itself.

### TODO or Help Wanted

Feedback on the `make prepush` target. I was conservative in what I added to that target, but we could possibly add more.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


.github/workflows/ci.yml
```diff
@@ -47,22 +47,23 @@ jobs:
...
+        run: make ci-job-syntax
+      - name: ci-job-compilation
+        run: make ci-job-compilation
+      - name: ci-job-debug-support-targets
+        run: make ci-job-debug-support-targets
 
-      - name: collect-build-artifacts
-        run: make ci-collect-artifacts
+      - name: ci-job-collect-build-artifacts
+        run: make ci-job-collect-artifacts
```
brghena, line 58: Every other `name` matches the rule to run, except this one

README.md
```diff
@@ -1,6 +1,9 @@
 # ![TockOS](http://www.tockos.org/assets/img/tock.svg "TockOS Logo")
 
-[![Build Status](https://travis-ci.org/tock/tock.svg?branch=master)](https://travis-ci.org/tock/tock)
+[![Formatting](https://github.com/tock/tock/workflows/format/badge.svg)](https://github.com/tock/tock/actions?query=branch%3Amaster)
+[![Build](https://github.com/tock/tock/workflows/build/badge.svg)](https://github.com/tock/tock/actions?query=branch%3Amaster)
+[![Unit Tests](https://github.com/tock/tock/workflows/testing/badge.svg)](https://github.com/tock/tock/actions?query=branch%3Amaster)
+[![Emulation](https://github.com/tock/tock/workflows/qemu/badge.svg)](https://github.com/tock/tock/actions?query=branch%3Amaster)
```
brghena, line 6: Do we need these separately? They all must be green at all times anyways, right?

hudson-ayers:
Three cheers for no more Travis!

As an aside, I would like to alias `make clippy` as a second name for `make ci-job-clippy`.

README.md
brghena: If they are going to be separate, can the links go to the individual workflows?

boards/opentitan/Makefile
```diff
@@ -6,9 +6,11 @@ PLATFORM=opentitan
 include ../Makefile.common
 
 qemu: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
+	$(call check_defined, OPENTITAN_BOOT_ROM)
```
brghena, line 9: Is there an "expected correct" default value we could give to `OPENTITAN_BOOT_ROM` so calling `make qemu` from the `opentitan/` directory just works? I'm not really sure why it needs to be a variable in the first place.

ppannuto:
@hudson-ayers It looks like the github workflow isn't running at all now -- is that because I edited it in this PR, or because I messed up the YAML somehow? What's the best way to test?

hudson-ayers:
"Invalid workflow file
You have an error in your yaml syntax on line 18"

I don't think you are supposed to assign jobs a name

hudson-ayers:
(you can see the error by clicking the checks tab at the top of the PR, and then selecting the workflow that did not run)

.github/workflows/ci.yml
```diff
@@ -71,26 +69,22 @@ jobs:
...
+        run: make ci-job-archs
+      - name: ci-job-kernel
+        run: make ci-job-kernel
+      - name: ci-job-chips
+        run: make ci-job-chips
+      - name: ci-job-tools
+        run: make ci-job-tools
 
-  emulation-check:
+  ci-qemu:
```
hudson-ayers, line 83: if you change the name of a workflow job, you have to also change the name of the job that bors blocks on in bors.toml

Makefile
```diff
@@ -8,268 +28,404 @@ usage:
...
+		cd ../..;\
+		done
+
+.PHONY: ci-job-kernel
+ci-job-kernel:
+	$(call banner,CI-Job: Kernel)
+	@cd kernel && CI=true RUSTFLAGS="-D warnings" TOCK_KERNEL_VERSION=ci_test cargo test
 
-.PHONY: emulation-setup
-emulation-setup: SHELL:=/usr/bin/env bash
```
hudson-ayers, line 216: I think this line is required for the test to run correctly

boards/opentitan/Makefile
alistair23: You need to download the file and we need to know where it is. The CI should do this for you, but for users they probably want to specify where it is.

ppannuto: I think this is a case where if you want to step outside of what the CI does for you, you need to know this for yourself. The OT + qemu stuff is too alpha to try to do more right now.

ppannuto:
> As an aside, I would like to alias make clippy as a second name for make ci-job-clippy.

I thought about that, and did intentionally remove it. My rationale was that targets should do something materially different, i.e. `make format` _does_ the formatting while `make ci-job-format` just checks it. For clippy, is there anything meaningfully different that the `clippy` target should do compared to the `ci-job-clippy` target? The makefile is pretty beastly with a lot of targets, so I'm feeling slightly motivated to eliminate duplicate targets in the interest of being a less overwhelming surface of options.

README.md
ppannuto: Turns out you can do a badge for the whole "tock-ci" workflow 👍

Makefile
ppannuto: I just removed the `mktemp`, same effect

bors.toml
```diff
@@ -2,7 +2,7 @@
 # pushed to master.
 status = [
   "continuous-integration/travis-ci/push",
```
hudson-ayers, line 4:
```suggestion
```

Makefile
```diff
@@ -8,268 +28,405 @@ usage:
...
 	@echo "This root Makefile has a few useful targets as well:"
-	@echo "        allboards: Compiles Tock for all supported boards"
+	@echo "        allaudit: Audit Cargo dependencies for all kernel sources"
+	@echo "       allboards: Compiles Tock for all supported boards"
 	@echo "        allcheck: Checks, but does not compile, Tock for all supported boards"
 	@echo "          alldoc: Builds Tock documentation for all boards"
-	@echo "           audit: Audit Cargo dependencies for all kernel sources"
-	@echo "              ci: Run all continuous integration tests"
 	@echo "           clean: Clean all builds"
-	@echo "          clippy: Runs the clippy code linter with Tock's default arguments"
```
bradjc, line 24: `make clippy` seems like a useful shortcut to keep around. In general, it seems useful to keep all chunks of `prepush` as easy targets, since `allcheck` takes forever.

tools/run_cargo_fmt.sh
```diff
@@ -49,15 +51,22 @@ let FAIL=0
 set -e
 
 # Find folders with Cargo.toml files in them and run `cargo fmt`.
-if [ "$1" == "diff" ]; then
-	# Just print out diffs and count errors, used by Travis
-	CARGO_FMT_ARGS="-- --check"
-fi
 for f in $(find . | grep Cargo.toml); do
+	printf "\rFormatting %-$((39))s" $(dirname $f)
```
bradjc, line 55: I was skeptical about this line, but it works well. However, I don't think we should add it. Using `cargo fmt --all` instead of this loop is over twice as fast on my computer. I think we should revisit my PR and switch. 2x performance seems worth having to custom case the tools directory.

bradjc:
> > As an aside, I would like to alias make clippy as a second name for make ci-job-clippy.
> 
> I thought about that, and did intentionally remove it. My rationale was that targets should do something materially different, i.e. `make format` _does_ the formatting while `make ci-job-format` just checks it. For clippy, is there anything meaningfully different that the `clippy` target should do compared to the `ci-job-clippy` target? The makefile is pretty beastly with a lot of targets, so I'm feeling slightly motivated to eliminate duplicate targets in the interest of being a less overwhelming surface of options.

Yes, it's more memorable and easier to type.

Makefile
```diff
@@ -8,268 +28,405 @@ usage:
...
+.PHONY: ci-all
+ci-all:\
+	ci-runner-github\
+	ci-runner-netlify
+
+# Run the fast jobs.
+# This is designed for developers, to be run often and before submitting code upstream.
+.PHONY: prepush
+prepush:\
+	ci-job-format\
```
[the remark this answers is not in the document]
gendx: I think it's intended that prepush only performs checks but doesn't edit any files, right?

Makefile
```diff
@@ -8,268 +28,405 @@ usage:
...
+.PHONY: ci-setup-qemu
+ci-setup-qemu: tools/qemu/riscv32-softmmu tools/qemu-runner/opentitan-boot-rom.elf
+
+tools/qemu/riscv32-softmmu:
+	@# Use the latest QEMU as it has OpenTitan support
+	@printf "Building QEMU, this could take a few minutes\n\n"
+	# Download Tock qemu fork if needed
 	@if [[ ! -d tools/qemu || ! -f tools/qemu/VERSION ]]; then \
 		rm -rf tools/qemu; \
 		cd tools; git clone https://github.com/alistair23/qemu.git --depth 1 -b riscv-tock.next; \
```
gendx, line 396: I think it's fine to leave it like this in the scope of this pull request, but let's not forget to add some kind of integrity check from Tock's side, i.e. cloning at a specific commit vetted by Tock (or checking that HEAD is at a specific commit).

Makefile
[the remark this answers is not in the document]
ppannuto: That was the thought.., but really why would you as a developer want to read diffs at this stage? Especially because the `format` job does check that all changes are at least staged before writing any files (i.e. it's easy to undo its actions). I'll change this.

Makefile
```diff
@@ -8,268 +28,405 @@ usage:
...
 	@echo "          alldoc: Builds Tock documentation for all boards"
-	@echo "           audit: Audit Cargo dependencies for all kernel sources"
-	@echo "              ci: Run all continuous integration tests"
 	@echo "           clean: Clean all builds"
-	@echo "          clippy: Runs the clippy code linter with Tock's default arguments"
-	@echo " emulation-check: Run the emulation tests for supported boards"
-	@echo " emulation-setup: Setup QEMU for the emulation tests"
 	@echo "          format: Runs the rustfmt tool on all kernel sources"
-	@echo "    format-check: Checks if the rustfmt tool would require changes, but doesn't make them"
-	@echo "            list: Lists available boards"
```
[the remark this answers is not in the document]
ppannuto: `format-check` is the exact same as `ci-job-format` and doesn't seem like a useful alias

`make list` basically prints the boards directory (caveat nordic for now) and a plain `make` also lists all of the boards already anyway. Nothing was using it, and we have the `list_boards` in tools if something elsewhere needs it. Was trying to reduce clutter in the options presented to users and this doesn't seem useful.

ppannuto:
I'm still lightly invested in keeping the surface area of commands down (i.e. not adding back `make clippy`); to that end, I've dramatically improved the performance the `prepush` target.

On my machine, with changes to kernel crate [which rebuilds everything] `prepush` now takes a median 5.5s; with no changes `prepush` takes less than a second.

I would rather keep `prepush` performant than ask developers to know about the internal parts of it.

tools/run_cargo_fmt.sh
ppannuto: Hah, I implemented this before I saw this comment :)

Makefile
ppannuto: I propose the alternate solution: `allcheck` is now fast :)

Makefile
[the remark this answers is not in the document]
bradjc: I was thinking of make list, which is a user-facing API, so it's not too surprising that nothing in the tree is using it. I used it with tinyos all the time, which is why I added it here a few years ago. Also, we reference it in the book.

bradjc:
> I'm still lightly invested in keeping the surface area of commands down (i.e. not adding back `make clippy`); to that end, I've dramatically improved the performance the `prepush` target.
> 
> On my machine, with changes to kernel crate [which rebuilds everything] `prepush` now takes a median 5.5s; with no changes `prepush` takes less than a second.
> 
> I would rather keep `prepush` performant than ask developers to know about the internal parts of it.

I think it's fine to not document it and just make it an alias, but I don't see anything wrong with making it easy for developers to run. Particularly if you want to experiment with new lints.

Makefile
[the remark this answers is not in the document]
ppannuto: Hmm.. thinking on this more, I wonder if this isn't a bit of an anti-pattern. I think part of my takeaway from #1784 was that we should push towards the `boards` folder just being `boards`, and it shouldn't require tooling to figure what boards exist in the ecosystem.

I'm okay with adding this back in the short term if you think it's necessary, but I think once we've finished component-izing the nordic boards and re-org'd that directory to be a flat namespace of boards, this should go away (and the tutorial can say `ls boards/` to see supported boards).

ppannuto:
Fair enough, one 'hidden' alias added.

Makefile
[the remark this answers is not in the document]
bradjc: In my opinion the point of supporting `make` is it just works and is developer and new user friendly. I don't see a reason to shift away from that thinking. If we can make any step easier for users, who maybe don't know to just switch to `/boards`, why not? What do we lose? The Makefile is already 435 lines, why not spend more of those on providing guiding hooks for new, unfamiliar users? I suppose if it is a burden to maintain it might not be worth it, but this has been working for the last four years I think.

Makefile
```diff
@@ -8,268 +28,417 @@ usage:
...
 
 
+# Commands
+.PHONY: fmt format
+fmt format: tools/.format_fresh
+	$(call banner,Formatting complete)
 
-###################################################################
+# Get a list of all rust source files (everything fmt operates on)
+$(eval RUST_FILES_IN_TREE := $(shell git ls-files | grep '\.rs$$'))
```
bradjc, line 109: This breaks `make format` if you download the .zip file. It also causes this less-than-friendly message to print: `fatal: not a git repository (or any of the parent directories): .git`.

boards/Makefile.common
```diff
@@ -176,7 +176,7 @@ all: release
 # binary. This makes checking for Rust errors much faster.
 .PHONY: check
 check:
-	$(Q)RUSTFLAGS="$(RUSTFLAGS_FOR_CARGO)" $(CARGO) check $(VERBOSE) --target=$(TARGET) --package $(PLATFORM) --release
+	$(Q)$(CARGO) check $(VERBOSE)
```
gendx, line 179: What are the consequences of removing these flags?
- `--package`: Shouldn't this be scoped to the current board (when running `make check` within a board), instead of the whole workspace?
- `--target`: Some code (e.g. in arch) depends on the target CPU. Is this checking the intended code?

Makefile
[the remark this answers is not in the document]
ppannuto: It's nothing to do with the complexity of the Makefile, it's trying to make things easier for new developers by presenting _fewer_ options. Right now running just `make` dumps out 40 lines of text and 12 possible commands. My eyes certainly tend to glaze over when I run `random_command -h` and it dumps a page of text. The goal is to reduce the number of commands that new users have to learn.

What value is `make list` if all it does is run `ls boards/`?

tools/run_cargo_fmt.sh
```diff
@@ -29,35 +29,62 @@ fi
...
+# them twice
+csplit Cargo.toml '/exclude = \[/' > /dev/null
+rm xx01
+
 # Find folders with Cargo.toml files in them and run `cargo fmt`.
-if [ "$1" == "diff" ]; then
-	# Just print out diffs and count errors, used by Travis
-	CARGO_FMT_ARGS="-- --check"
-fi
 for f in $(find . | grep Cargo.toml); do
```
gendx, line 59: Is there any reason to keep this loop, as opposed to running `cargo fmt` directly on the workspace?

bradjc: I agree we should switch to `cargo fmt --all`.

Makefile
```diff
@@ -8,268 +28,417 @@ usage:
...
+# Run all possible CI. If this passses locally, all cloud CI *must* pass as well.
+.PHONY: ci-all
+ci-all:\
+	ci-runner-github\
+	ci-runner-netlify
+
+# Run the fast jobs.
+# This is designed for developers, to be run often and before submitting code upstream.
+.PHONY: prepush
+prepush:\
```
gendx, line 136: As a developer, I'd still like to have a rule that builds boards and runs checks/tests locally, without having to upload my work-in-progress code to a pull request to get feedback from the GitHub actions. Maybe in the "hidden convenience targets" section?

ppannuto: Does `make ci-all` cover that need for you? Or is there something else you have in mind?

boards/Makefile.common
ppannuto: re `--package`, [according to the docs](https://doc.rust-lang.org/stable/cargo/reference/workspaces.html#the-workspace-section):
> If neither of those flags are specified, Cargo will use the package in the current working directory.

Empirically, this does this thing (i.e. only checks `hail` & dependencies in the hail folder)

---

re `--target`, I mentioned this in the commit message: yes, it does cause check to skip over any `cfg(arch`-gated code. However, there is very little of this in Tock, and the rules that actually build targets will catch them. The gain is that crates are not recompiled when you change which board you're checking for (as no flags to cargo have changed). This is a significant speedup on my machine

Makefile
ppannuto: Fixed.

tools/run_cargo_fmt.sh
ppannuto: I think this is already doing what we want.

When run in the root of a workspace, `cargo fmt` runs on all modules in the workspace. That is what the `== ./Cargo.toml` case covers. The `--all` flag is only meaningful if you want to run `cargo fmt` from in a module directory but want it to format the whole workspace instead of just the module you're running in (i.e. you typed `cargo fmt` in the `boards/hail` directory).

In no case does `cargo fmt --all` choose to format modules that are listed as excluded from the workspace. The format script first runs `cargo fmt` in the root, which formats the whole workspace, and then the loop finds all `Cargo.toml`s in the tree and skips any that were already formatted as part of the workspace.

I tested this fairly extensively, making sure that nothing is double-formatted nor is anything skipped in formatting. Is there any incorrect behavior you're observing?

Makefile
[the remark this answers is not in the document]
bradjc: It doesn't!

```
Supported Tock Boards:
 - nordic/nrf52dk
 - nordic/nrf52840_dongle
 - nordic/nrf52840dk
 - opentitan
 - stm32f3discovery
 - redboard_artemis_nano
 - launchxl
 - hifive1
 - hail
 - arty_e21
 - nucleo_f429zi
 - acd52832
 - imix
 - nucleo_f446re

To build the kernel for a particular board, change to that directory
and run make:
    cd boards/hail
    make
```

is not

```
ls boards/
Makefile.common		hail			nano33ble		redboard_artemis_nano
README.md		hifive1			nordic			stm32f3discovery
acd52832		imix			nucleo_f429zi
arty_e21		kernel_layout.ld	nucleo_f446re
components		launchxl		opentitan
```

One provides context and help, plus much better formatting, and doesn't include things that are not boards. If it essential to remove things from the printout, why does a new user need to know how CI is implemented? Or how to build the documentation?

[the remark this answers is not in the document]
ppannuto: All right, you've convinced me. I've added it back.

tools/run_cargo_fmt.sh
bradjc: For me, it is:

```
$ time make fmt
Formatting complete.

********************************************************************************
* Formatting complete                                                          *
********************************************************************************


real	0m10.809s
user	0m3.021s
sys	0m3.869s

 $ time cargo fmt --all

real	0m4.538s
user	0m3.467s
sys	0m0.933s
```

ppannuto: It takes longer because it does more. `cargo fmt --all` doesn't format every module in the tock tree:

```
[-bash] Thu 28 May 15:56 [[ci-policy-impl $] ~/code/helena-project/tock]
$ cargo fmt --all -- --check
[-bash] Thu 28 May 15:56 [[ci-policy-impl $] ~/code/helena-project/tock]
$ make ci-job-format

********************************************************************************
* CI-Job: Format Check                                                         *
********************************************************************************

Formatting tools/alert_codes                      <- Contains formatting errors!
Diff in /Volumes/code/helena-project/tock/tools/alert_codes/src/main.rs at line 17:

 /// Returns the alert code specified on the command line, or prints a usage
 /// error if it was omitted or incorrectly specified.
-   fn get_alert_code() -> Result<u32, ()> {
+fn get_alert_code() -> Result<u32, ()> {
     let mut args = std::env::args_os();
     if args.len() != 2 {
         usage_error("Incorrect number of arguments");

Formatting complete.

Formatting errors.
See above for details
make[1]: *** [tools/.format_fresh] Error 1
make: *** [ci-job-format] Error 2
```

There is literally no difference between running `cargo fmt` and `cargo fmt --all` **in the root of the workspace**. The only thing the `--all` flag does is change the behavior when run in a module (aka in a subdirectory).

The script runs `cargo fmt` in the root which is exactly equivalent to running `cargo fmt --all` and then checks the crates that excluded from the cargo workspace as well.

bradjc: I now realize that this version is 2x faster than what is in master now, so that's good. I worry about the maintainability of this script, but `cargo fmt --all` works well enough if we need to fall back to that.

gendx: I think a good solution to this problem would be to have a second workspace for tools. Then we can run `cargo fmt` twice: in the root directory and in the `tools/` directory. But this is not in scope for this pull request.

Makefile
gendx: I've already mentioned it in https://github.com/tock/tock/pull/1861 (in particular https://github.com/tock/tock/pull/1861#issuecomment-630982811), which started all this discussion.

Specifically, I've already mentioned (https://github.com/tock/tock/issues/1865#issuecomment-630976331) that the following chain is highly problematic on a development machine.

```
ci-all
    -> ci-runner-github
    -> ci-runner-github-qemu
    -> ci-job-qemu
    -> ci-setup-qemu
    -> git clone https://github.com/alistair23/qemu.git && ./configure
    -> execution of arbitrary code without any integrity check from Tock
```

The currently proposed `make prepush` is fine and fast, but not that useful as it only checks code formatting and simple lints. I'd expect at the very least a rule that also includes `make allboards` (so that boards indeed compile). And more generally all the rules that only depend on Cargo.

One could do `make prepush && make allboards && make foobar`, but I'd be nice to simply have a rule for that (like there is the "hidden" `make clippy` rule).

Makefile
```diff
@@ -8,268 +28,428 @@ usage:
...
-
-## Actual Rules (Netlify)
+.PHONY: ci-job-collect-artifacts
+ci-job-collect-artifacts: ci-job-compilation
+	# Collect binary images for each board
+	#
+	# This is currently used only for code size detection changes, but in
+	# the future may also be used to support checks for deterministic builds.
+	@mkdir -p ./ci-artifacts
+	@rm -rf "./ci-artifacts/*"
```
gendx, line 332: Wouldn't it be clearer to first remove the folder and then create it, instead of creating it and removing every file in it? The rm rule would be simpler: `rm -rf ci-artifacts`.

I also don't get why some paths start with `./` and some not (throughout this Makefile). Isn't this prefix useful only when running an executable?

Makefile
ppannuto: OH! I messed that up! That setup rule should prompt in non-CI context. I'll update that, sorry!

ppannuto: Okay, this is now fixed. This `ci-all` rule will now prompt and offer to do the install for you. There is also a new `ci-nosetup` rule that runs all the CI that requires no new installation on the machine it's being run on.

Makefile
ppannuto: I've updated this to wipe and re-create the folder. I've also moved the folder into `tools/` to avoid cluttering the root if run locally.

As for `./` I assume that was just style of the people who wrote rules. I've updated such that its' only used for executables now.

ppannuto:
I don't think this needs to wait on #1771, there's only one line of changes in the `boards/Makefile.common` here, it should be a minimal conflict (and easy to resolve, simply choose this side of the diff).

Makefile
```diff
@@ -8,268 +27,508 @@ usage:
...
+		tar -xf opentitan-dist/opentitan-snapshot-20191101-*.tar.xz; \
+		mv opentitan-snapshot-20191101-*/sw/device/boot_rom/boot_rom_fpga_nexysvideo.elf $$pwd/tools/qemu-runner/opentitan-boot-rom.elf
+endef
 
+.PHONY: ci-setup-qemu
+ci-setup-qemu:
+	$(call ci_setup_helper,\
+		cd tools/qemu && [[ $$(git rev-parse --short HEAD) == "1ef6d40" ]] && [ -x riscv32-softmmu ] && echo yes,\
+		Clone QEMU fork (with riscv fixes) and run its build scripts,\
+		ci_setup_qemu_riscv,\
```
alistair23, line 501: This will print a message asking the user to install something. We aren't installing anything in this case though.

Makefile
```diff
@@ -8,268 +27,508 @@ usage:
...
+		CI_JOB_QEMU_RISCV)
+	$(call ci_setup_helper,\
+		[[ $$(cksum tools/qemu-runner/opentitan-boot-rom.elf | cut -d' ' -f1) == "2835238144" ]] && echo yes,\
+		Download opentitan archive and unpack a ROM image,\
+		ci_setup_qemu_opentitan,\
+		CI_JOB_QEMU_OPENTITAN)
```
alistair23, line 507: Same here, nothing is being installed.

Makefile
ppannuto: I'm open to a word other than "install", but the intent is to prompt before running lots of big arbitrary commands automatically.

Makefile
ppannuto: Same answer I think, I read "install" here as "install bits onto disk":

Tock needs to install something to run this step.

> You have run a (likely CI) rule that requires Tock to install something on your
> machine. Tock can do this automatically for you, or you can look at the recipe
> for 'ci_setup_qemu_opentitan' and install yourself.
> 
> Continuing will: Download opentitan archive and unpack a ROM image.
> 
> You can use 'make ci-nosetup' to run all CI with no new install requirements.
> 
> Should Tock run install steps for you? [y/N] ^C

I think the "Continuing will" line conveys the intent well. Do you have suggestions for what you think would be better?

Makefile
bradjc: "Install locally"?

alistair23: Maybe just change it to "run"

> You have run a (likely CI) rule that requires Tock's build system to run something on your
> machine.

ppannuto: changed to:
> You have run a (likely CI) rule that requires Tock to run setup commands on your machine.

Makefile
```diff
@@ -8,268 +27,510 @@ usage:
...
+	$(eval guard_variable := $(strip $(4)))
+	@# First, if the dependency is installed, we can bail early
+	$(if $(shell bash -c '$(1)'),$(eval $(guard_variable) := true),
+	@# If running in CI context always yes
+	$(if $(CI),$(eval do_install := yes_CI),
+	@# If running nosetup always no
+	$(if $(TOCK_NOSETUP),$(eval do_install := ),
+	@# Otherwise, ask
+	$(info )
+	$(info Tock needs to install something to run this step.)
```
alistair23, line 86: run here as well?

alistair23: Or maybe just remove this line?

Makefile
```diff
@@ -8,268 +27,510 @@ usage:
...
+	$(info )
+	$(info You have run a (likely CI) rule that requires Tock to run setup commands on your)
+	$(info machine. Tock can do this automatically for you, or you can look at the recipe)
+	$(info for '$(build_function)' and install yourself.)
+	$(info )
+	$(info Continuing will: $(explanation).)
+	$(info )
+	$(info You can use 'make ci-nosetup' to run all CI with no new install requirements.)
```
alistair23, line 94: Same issue with install

Makefile
ppannuto: Deleted it.

Makefile
ppannuto: Removed that, and one other one as well.

ppannuto:
The updated prompt:

> You have run a (likely CI) rule that requires Tock to run setup commands on your
> machine. Tock can do this automatically for you, or you can look at the recipe
> for 'ci_setup_tools' and do it yourself.
> 
> Continuing will: Install 'libusb' for development using your package manager.
> 
> You can use 'make ci-nosetup' to run all CI with no new setup requirements.
> 
> Should Tock run setup commands for you? [y/N]

Makefile
alistair23: Thanks! Sorry to be a pain, I think it just turns people off running the tests if you tell them we need to install things.

ppannuto:
That's the second build in a row where one of the checks randomly cancelled itself. I've submitted a ticket to github to see if they have any insight.

bradjc:
Can we merge this? We have to rebase #1771 anyway.

I don't think this is the end of CI updates, so if there are lingering issues we can fix them. But I think we need more PRs running with these changes to actually test this fully.

Issue #1552, opened 2020-01-18
"Safe" wait primitive?


ppannuto:
While reviewing #1551, I noticed this:

```rust
        // start comparator
        regs.events_ready.set(0);
        regs.tasks_start.set(1);
        // wait for comparator to be ready
        // delay is on order of 3 microseconds so spin wait is OK
        while regs.events_ready.get() == 0 {}
    }
```

We do stuff like this in a few places, but as we start busy-looping on more complex peripherals, I wonder if we perhaps introduce the possibility for the kernel to hang in a way that's challenging to debug. Should we have a kernel primitive of some form that safely spins?

```rust
fn wait_for_hardware<F: Frequency> (fn: [closure type], max_µs: u32) -> bool {
     // Compute max number of tries given `max_us` and core speed
     let max_tries = ...
     while (max_tries) {
          if (fn()) return true;
     }
     false
}

        // wait for comparator to be ready
        if ! wait_for_hardware(|| {regs.events_ready.get() == 0}, 3) {
              return InitializationError; //?
              panic!("useful error message"); //?
        }
```

Or is this the kind of situation where we'd never really be able to do anything useful in the case of failure?

lgnashold:
With reference to #1551, it could be useful if the analog comparator is broken for some reason but the rest of the chip is okay. It would definitely be easier to debug an error in initializing the comparator rather than just getting stuck spinning.

phil-levis:
I think the standard approach is have a watchdog. Its handler could print the stack frame.

Or we could have lots of kernel debug messages, @bradjc ;)

ppannuto:
Sure, I guess the question is whether this is something better than a standard watchdog that we can do now. Especially in the context of multi-app boards, one might imagine a board with a dead sensor having one application fail (`ENODEVICE` perhaps from powering on the device failing) while the rest carries on.

Perhaps it's a bit fanciful -- or prematurely complex for the current reality of Tock -- and not useful / needed in practice. However, I guess one hope could be that a 'robust / reliable / etc' OS would aspire to never watchdog.

(Aside: This isn't meant to block #1551 in any way; I pulled this into a separate issue deliberately. The question here is more about this underlying philosophical question of how to handle dead hardware as an OS, to the extent that we can anyway)

bradjc:
I think this is an interesting push in a very Tock direction. There are others along this line:
- `map()` calls where the option is unexpectedly `None`
- functions/operations that cause an unexpected `panic!()`

[...]

PR #1882, opened 2020-05-27
chips: ibex: Enable low power state

alistair23:
### Pull Request Overview

This PR configures the OpenTitan Power Management IP to enter low power mode before calling the WFI.

### Testing Strategy

This pull request was tested by running Tock and apps on the FPGA OT instance.

### Documentation Updated

### Formatting

- [X] Ran `make format`.

### DIFF
--- a/chips/lowrisc/src/pwrmgr.rs
+++ b/chips/lowrisc/src/pwrmgr.rs
@@ -1,0 +1,110 @@
+//! Power Mangement for LowRISC
+
+use kernel::common::registers::{register_bitfields, register_structs, ReadOnly, ReadWrite};
+use kernel::common::StaticRef;
+
+register_structs! {
+    pub PwrMgrRegisters {
+        (0x00 => ctrl_cfg_regwen: ReadOnly<u32, CTRL_CFG_REGWEN::Register>),
+        (0x04 => control: ReadWrite<u32, CONTROL::Register>),
+        (0x08 => cfg_cdc_sync: ReadWrite<u32, CFG_CDC_SYNC::Register>),
+        (0x0C => wakeup_en_regwen: ReadWrite<u32, WAKEUP_EN_REGWEN::Register>),
+        (0x10 => wakeup_en: ReadWrite<u32, WAKEUP_EN::Register>),
+        (0x14 => wake_status: ReadOnly<u32, WAKE_STATUS::Register>),
+        (0x18 => reset_en_regwen: ReadWrite<u32, RESET_EN_REGWEN::Register>),
+        (0x1C => reset_en: ReadWrite<u32, RESET_EN::Register>),
+        (0x20 => reset_status: ReadOnly<u32, RESET_STATUS::Register>),
+        (0x24 => wake_info_capture_dis: ReadWrite<u32, WAKE_INFO_CAPTURE_DIS::Register>),
+        (0x28 => wake_info: ReadWrite<u32, WAKE_INFO::Register>),
+        (0x2C => @END),
+    }
+}
+
+register_bitfields![u32,
+    CTRL_CFG_REGWEN [
+        EN OFFSET(0) NUMBITS(1) []
+    ],
+    CONTROL [
+        LOW_POWER_HINT OFFSET(0) NUMBITS(1) [],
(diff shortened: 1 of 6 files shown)


chips/lowrisc/src/pwrmgr.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    pub fn handle_interrupt(&self) {
+        let regs = self.registers;
+
+        // Disable power saving
+        regs.control.write(CONTROL::LOW_POWER_HINT::CLEAR);
+    }
+
+    pub fn enalbe_low_power(&self) {
```
ppannuto, line 80:
```suggestion
    pub fn enable_low_power(&self) {
```

chips/lowrisc/src/pwrmgr.rs
```diff
@@ -0,0 +1,101 @@
...
+            CONTROL::LOW_POWER_HINT::SET
+                + CONTROL::CORE_CLK_EN::CLEAR
+                + CONTROL::IO_CLK_EN::SET
+                + CONTROL::MAIN_PD_N::CLEAR,
+        );
+
+        // Propagate changes to slow clock domain
+        regs.cfg_cdc_sync.write(CFG_CDC_SYNC::SYNC::SET);
+
+        loop {
```
ppannuto, line 95: Stuff like this always makes me nervous (and wishing again for something like #1552...) -- is there an expected number of cycles (order of magnitude accurate) we could comment this with? I could imagine this taking a couple hundred cycles if these are 32 kHz vs 16 MHz clocks and they have to do handshaking...

chips/lowrisc/src/pwrmgr.rs
```diff
@@ -0,0 +1,101 @@
...
+impl PwrMgr {
+    pub const fn new(base: StaticRef<PwrMgrRegisters>) -> PwrMgr {
+        PwrMgr { registers: base }
+    }
+
+    pub fn handle_interrupt(&self) {
+        let regs = self.registers;
+
+        // Disable power saving
+        regs.control.write(CONTROL::LOW_POWER_HINT::CLEAR);
```
ppannuto, line 77: This doesn't need all the sync machinery of enabling power saving?

chips/lowrisc/src/pwrmgr.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    pub fn enalbe_low_power(&self) {
+        let regs = self.registers;
+
+        // Next WFI should trigger low power entry
+        // Leave the IO clock enabled as we need to get interrupts
+        regs.control.write(
```
ppannuto, line 85: Should this check the current power state, such that it could skip if already configured for this (especially given the potentially long sync after this and the fact that this is called on every WFI)?

chips/lowrisc/src/pwrmgr.rs
alistair23: I just assumed it didn't, but I think it also needs the same sync.

chips/lowrisc/src/pwrmgr.rs
alistair23: I'll add a check

chips/lowrisc/src/pwrmgr.rs
alistair23: I can't see an indication in the OT repo about how many cycles this should take.

I have created a PR to move towards a safe loop (at least here): https://github.com/tock/tock/pull/1886

alistair23:
I have updated this addressing comments and rebasing it on https://github.com/tock/tock/pull/1886

ppannuto:
I think this is blocked on #1886 for now.

In the meantime, I'm still slightly worried about understanding how peripherals interact with the power manager. For example, on the SAM4L [there is fairly complex logic](https://github.com/tock/tock/blob/master/chips/sam4l/src/pm.rs#L1020) that checks which sleep state is safe to enter based on the peripherals that are active. Does ibex manage the various sleep states automatically, or is this something that Tock will need to manage?

alistair23:
I have updated this and it's no longer blocking.

As for if the sleep state is safe to enter I don't see anything in the documentation describing that we need to check anything.

PR #1886, opened 2020-05-28
kernel: platform: Add a loop with interrupts function

alistair23:
### Pull Request Overview

This PR adds a function to the `Chip` trait that allows us to loop while waiting on interrupts.

This is along the lines of what https://github.com/tock/tock/issues/1552 was looking at.

#### Why not just add a macro for this?

The `Chip` trait is not available globally and as it already has all the interrupt checking functions set-up it makes sense to re-use it.

#### When is this useful?

This is useful for possible long running loops done by the `Chip`. An example of this is entering low power mode, where we want to wait until the HW is ready, unless there is an interrupt.

### Testing Strategy

Running on OT.

### Documentation Updated

### Formatting

- [X] Ran `make format`.

### DIFF
--- a/kernel/src/platform/mod.rs
+++ b/kernel/src/platform/mod.rs
@@ -94,6 +94,39 @@ pub trait Chip {
     /// Ask the chip to check if there are any pending interrupts.
     fn has_pending_interrupts(&self) -> bool;
 
+    /// Run a function in an interruptable loop.
+    ///
+    /// The function will run until it returns true, an interrupt occurs or if
+    /// `max_tries` is not `None` and that limit is reached.
+    /// If the function returns true this call will also return true. If an
+    /// interrupt occurs or `max_tries` is reached this call will return false.
+    fn check_until_true_or_interrupt<F>(&self, f: F, max_tries: Option<usize>) -> bool
+    where
+        F: Fn() -> bool,
+    {
+        match max_tries {
+            Some(t) => {
+                for _i in 0..t {
+                    if self.has_pending_interrupts() {
+                        return false;
+                    }
+                    if f() {
+                        return true;
+                    }
+                }
+            }
+            None => {
+                while !self.has_pending_interrupts() {
+                    if f() {
+                        return true;
+                    }
+                }
+            }
+        }
+
+        false
+    }
+
     /// Returns a reference to the implementation for the MPU on this chip.
     fn mpu(&self) -> &Self::MPU;
 


alistair23:
See https://github.com/tock/tock/pull/1882 for an example of this in use

kernel/src/platform/mod.rs
```diff
@@ -94,6 +94,24 @@ pub trait Chip {
     /// Ask the chip to check if there are any pending interrupts.
     fn has_pending_interrupts(&self) -> bool;
 
+    /// Run a function in an interruptable loop.
+    ///
+    /// The Function will run untilit retruns true or until an interrupt occurs.
```
ppannuto, line 99:
```suggestion
    /// The function will run until it returns true or until an interrupt occurs.
```

alistair23:
This isn't in the full spirit of #1552 as it couldn't be used in device drivers (like the example in #1552). It's more of a step in that direction. The main limitation is access to the `Chip`.

I'm not sure if `max_tries` makes sense. For the OT use case there is no max tries as we want to hang forever (or until interrupted). I think for lots of HW interfaces this is the case as well.

Generally, what action do we take if we hit `max_tries`, panic?

alistair23:
Also, instead of max_tries it might make more sense to think about a watch dog implementation? Like other OSes do. That way it isn't up to users of loops to pick safe values but instead there is a global watch dog to detect stalls.

hudson-ayers:
> Also, instead of max_tries it might make more sense to think about a watch dog implementation? Like other OSes do. That way it isn't up to users of loops to pick safe values but instead there is a global watch dog to detect stalls.

I think that max_tries feels like a significant advantage over a watchdog. A watchdog can cause problems if there are long computations in the kernel, so often watchdog timers are set to much higher values than something like `max_tries` would be in this case, and it is still possible to accidentally set a watchdog timer to too low a value and disrupt correctly executing code. 

Also, with the `max_tries` approach it is possible to handle a failure by disabling the associated peripheral while allowing the rest of the system to continue, which I think is in line with Tock's dependability promises to apps. (i.e. if app1 is using the ADC and some point in the ADC configuation loops forever, it probably makes more sense to disable the ADC and let other apps continue running rather than cause all apps to hang(currently what happens)/fail(if we paniced)/restart (what I assume a watchdog might do).

It is much harder with a watchdog approach to simply disable the responsible hardware peripheral.

alistair23:
Ok, I have added a `max_tries` argument. It can be set to `None` to run forever.

alistair23:
I have also changed the name.

@bradjc The idea is that this is a generic function that can be used in other places. There are a large number of "loop on HW" loops in Tock and it would be nice to replace them with something that could be a little safer.

As for pushing for low power I still think there will be cases where you need to wait on the HW before calling a WFI and this is very helpful for that.

bradjc:
I think we should either actually use it for one of these loops, or just to move this into the crate that needs it.

alistair23:
The use case is: https://github.com/tock/tock/pull/1882

Are you saying you want a use case in other `Chips` to justify a new function? Doesn't #1552 help justify this (although we don't cover all of those cases).

hudson-ayers:
For many of the use cases described in #1882, I think it might actually be a problem that this returns false when an interrupt arrives. There are plenty of cases where the workflow is something like:
```rust
registers.enable_task.set(true);
while !registers.enabled.is_set() {}
```
Where an interrupt does actually arrive when the peripheral becomes enabled, but it is much easier to spin on the enabled register than to rewrite the code to use some asynchronous state machine, especially when the enabling is never supposed to fail. 

However for any instance where an interrupt arriving could cause this to return false, the interrupt that should indicate the peripheral has been enabled would instead cause loop_with_interrupts to indicate a failure.

Perhaps it should also be a parameter whether to break on interrupts? Or perhaps there should just be a second function that does not require a chip reference which is a loop_max_tries?

My personal experience is that there are a lot more instances in Tock where a loop_max_tries would be useful than instances that want to return false when there is an interrupt, but maybe that is just an artifact of the device drivers I have worked on.

alistair23:
Closing this PR and making the new function OpenTitan specific in https://github.com/tock/tock/pull/1882.

We can always pull it out in the future if required.

PR #1767, opened 2020-04-17
Scheduler trait + transition all boards to round robin scheduler

hudson-ayers:
### Pull Request Overview

This pull request creates a Scheduler trait that allows for different scheduling algorithms to be used by the Kernel, and implements 4 schedulers using this new trait:

1.  A round robin scheduler which resolves a number of the problems with the current scheduling implementation, such that it is no longer possible for any app to starve another of the CPU indefinitely, and which fairly distributes timeslices even when interrupts arrive during execution.
2. A cooperative scheduler, which can be used on platforms without a systick.
3. A multi-level feedback queue scheduler which defers bottom-half interrupt handling until timeslice expiration or yield. This shows more complex scheduling algorithms are possible, that storing per-process state is possible, and that this design allows for schedulers to choose when to process interrupts.

This PR puts each scheduling algorithm into a `sched/` folder in the Kernel crate, and creates components for each scheduler.

I am interested in this from the perspective of wanting to eventually implement a real-time scheduler of some sort, which may require the scheduler to have fine-tuned control of when to handle interrupts. I could also imagine a power-aware scheduler, which would need the ability to control when to call chip.sleep(), among other things. Implementing a priority scheduler may require the ability to run different apps with different time quantums. 

### Testing Strategy

Tested several combinations of apps, including whileone, on each scheduler. 

### TODO or Help Wanted

We need to decide which schedulers are worth keeping upstream, as code rot is inevitable for schedulers not used.

### Documentation Updated

- [ ] Once this interface is finalized, will need to add documentation describing how/why to choose different schedulers.

### Formatting

- [x] Ran `make formatall`.

PR #1887, opened 2020-05-28
sched: Add a watchdog

alistair23:
### Pull Request Overview

As part of https://github.com/tock/tock/pull/1882 we started talking about safe waits (https://github.com/tock/tock/issues/1552).

I have a PR with some initial safe wait support (https://github.com/tock/tock/pull/1886) in that we can be interrupted while looping on a hardware condition. The idea here is that while entering low power mode we can skip the infinite loop if we get an interrupt.

https://github.com/tock/tock/pull/1886 doesn't address a large number of loops in the driver capsules as they don't have access to `Chip` so it's more difficult to wait on interrupts. The harder part is what to do when a wait "fails". What happens when returning to full power fails to occur in a specific time, what happens when a bus fails to respond in a reasonable amount of time? How do we gracefully handle this and continue running in some capacity.

This PR starts to implement a watchdog functionality. This is not instead of a safe wait (in lots of cases a safe wait would be more graceful) but is generally to help avoid infinite loops and help debugging. It is also something that has come up to help make Tock more reliable.

It is up to `Chips` and boards to handle the watchdog interrupt and decided how to act.

Currently we tickle the watch dog when iterating on each process and at the start of each kernel loop.

### Testing Strategy

None

### TODO or Help Wanted

Feedback

### Documentation Updated

### Formatting

- [X] Ran `make format`.


kernel/src/platform/mod.rs
```diff
@@ -101,6 +101,20 @@ pub trait Chip {
...
+    /// Tickles the watchdog timer if the Chip has one.
+    ///
+    /// This is called at the start of every kernel loop and when about to jump
+    /// to an app.
+    /// If a watchdog timer is supported, the board is expected to already have
+    /// enabled it with the correct frequency (using the `enable()` function in
+    /// the watchdog HIL). The interval must be at least as long as the
+    /// app time slice.
+    /// The chip must ensure that the watchdog timer is disabled while sleeping
+    /// and re-enabled afterwards.
```
bradjc, line 113: Shouldn't the kernel manage this, rather than each chip having to remember to do this?

alistair23: The disable and enable for sleep?

The `sleep()` is part of the `Chip` so right now it will be up to the `Chip`. We also don't currently expose a WDT to the kernel.

We could look at exposing the WDT to the kernel (like the MPU or systick) if that is a better option.

bradjc:
My thought would be to add a watchdog trait, inline with systick, mpu, and UKB. Then the kernel could entirely manage this, and if a chip wants to opt-out it could just not provide a watchdog implementation.

alistair23:
Pushed an update where it's not a general trait like the MPU or systick.

hudson-ayers:
I feel that whether to include a watchdog is probably more of a board-specific decision than a chip specific one. For example, people maintaining out of tree boards probably should not be forced to also maintain an out-of-tree version of an otherwise in-tree chip just because the board maintainer has different thoughts on whether to use a watchdog. Maybe `kernel_loop()` could take in a boolean that indicates whether to use the chip watchdog or the unit version?

Also, I am interested to see an example of how watchdog interrupts would be handled.

kernel/src/platform/watchdog.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn tickle(&self) {}
+
+    /// Suspends the watchdog timer. After calling this the timer should not
+    /// fire until after `tickle()` has been called. This function is called
+    /// before sleeping.
+    fn suspend(&self) {}
+}
+
+/// Implement default WatchDog trait for unit.
+impl WatchDog for () {}
```
hudson-ayers, line 27: There is already a watchdog trait in kernel/src/hil/watchdog.rs that is implemented by the sam4l (though not currently used in the kernel). It probably makes sense to modify that one rather than have two different watchdog traits.
Or just remove that one in favor of this one, and change the sam4l to implement the new one.

alistair23: Before I go ahead and convert it over to a new implementation I would like to have the implementation agreed upon. Once that's done I'll update the SAM4L

kernel/src/sched.rs
```diff
@@ -313,6 +317,7 @@ impl Kernel {
                         && !DynamicDeferredCall::global_instance_calls_pending().unwrap_or(false)
                         && self.processes_blocked()
                     {
+                        chip.watchdog().suspend();
                         chip.sleep();
```
ppannuto, line 321:
```suggestion
                        chip.watchdog().suspend();
                        chip.sleep();
                        chip.watchdog.().tickle();
```

I think given the HIL interface as listed, there should be a `tickle` here to make sure the watchdog is active once the chip is active again

kernel/src/platform/watchdog.rs
```diff
@@ -0,0 +1,27 @@
...
+/// This trait is called from the `kernel_loop()` code to setup
+/// and maintain the watchdog timer.
+/// It is up to the specific `Chip` how it will handle watchdog interrupts.
+pub trait WatchDog {
+    /// This function must enable the watchdog timer and configure it to
+    /// trigger regulary. The period of the timer is left to the implementation
+    /// to decide. The implementation must ensure that it doesn't trigger too
+    /// early (when we haven't hung for example) or too late as to not catch
+    /// faults.
+    fn setup(&self) {}
```
ppannuto, line 13: Is the watchdog running once `setup` returns, or does it start running after the first `tickle`?

kernel/src/platform/watchdog.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    /// This function must tickle the watchdog to reset the timer.
+    /// If the watchdog was previously suspended then this should also
+    /// resume the timer.
+    fn tickle(&self) {}
+
+    /// Suspends the watchdog timer. After calling this the timer should not
+    /// fire until after `tickle()` has been called. This function is called
+    /// before sleeping.
+    fn suspend(&self) {}
```
ppannuto, line 23: Should there possibly be a dedicated `resume` method (perhaps with a default implementation that calls `tickle`?).

I'm not exactly well-versed in the array of watchdog interfaces in the world, but I just want to make sure that we aren't conflating two operations into one interface that would make it hard for some chip to implement -- I think it might also simplify `tickle` implementations on platforms where tickling doesn't auto re-enable, as it would not have to check the status first on every tickle.

ppannuto:
> It is up to Chips and boards to handle the watchdog interrupt and decided how to act.

I'm not quite sure I see in this interface how _boards_ get to configure the watchdog. I imagine this is actually fairly important, as watchdogging is much more of a board-specific question that a chip-specific one.

I do think this is a pretty good interface for the chip part of the watchdog system.

kernel/src/platform/watchdog.rs
alistair23: I have changed this to specify that the timer must be running

kernel/src/platform/watchdog.rs
alistair23: Good point. I did think about it and thought `tickle()` should be enough, but you are right that maybe on some hardware that could add complexity.

I'll add a `resume()` function

alistair23:
That is a good point. Currently there is no hookup to boards. That is probably something that the `Chips` crate will have to expose to a board.

alistair23:
I have updated this so it passes the tests and remove the old watchdog HIL and replaced it with this one.

Untested on the SAM4L though

alistair23:
I have updated this so it passes the tests and remove the old watchdog HIL and replaced it with this one.

Untested on the SAM4L though

kernel/src/platform/watchdog.rs
```diff
@@ -0,0 +1,35 @@
...
+
+    /// Resumes the watchdog timer. After calling this the timer should be
+    /// running again. This is called after returning from sleep, after
+    /// `suspecd()` was called.
```
hudson-ayers, line 28:
```suggestion
    /// `suspend()` was called.
```

alistair23:
I have fixed the spelling issue.

I agree about configuring from boards. There are a few places where we would like to do this (see https://github.com/tock/tock/issues/1998) so it's something that should happen eventually. We just need to figure out a good way to do it.

hudson-ayers:
> I have fixed the spelling issue.

I don't think you pushed the commit

alistair23:
I did, GitHub is just really slow.

bradjc:
I think the tension here is that after the scheduler PR is merged a board will be able to select a cooperative scheduler, but not disable the watchdog (if the chip crate has enabled the watchdog). A watchdog timer is probably under the purview of the chip crate (I don't think that boards should have to decide what watchdog to use, and there probably is only one choice anyway). Perhaps this falls to the scheduler PR: just like the scheduler algorithm decides how to use the systick, maybe it should also decide how to use the watchdog.

Overall this PR looks good and moves things forward so we should merge it soon.

If we start using the watchdog in the future, it would be nice to add a function to the trait that allows the kernel to check if the last reset was due to the watchdog. This would help with debugging by allowing the kernel to somehow notify developers that the watchdog is triggering.

bradjc:
@alistair23 @hudson-ayers How do you want to merge this? This before or after #1767?

hudson-ayers:
I think it is fine for this to go first, I can rebase #1767 after.

PR #1888, opened 2020-05-29
nRF52: uart fixes for unknown bootloader configurations

bradjc:
### Pull Request Overview
Trying to setup the nano33ble, I ran into some issues with UART.

1. The bootloader seems to be enabling UART TX interrupt, which was causing an early interrupt, which caused Tock to re-send its TX buffer, which caused the number of TX'd bytes to be wrong, which caused UART to stop working.
2. There is a debug!() message that can never print (no callback is issued and interrupts are disabled), so it was very misleading since that case _was_ occurring, but yet there was no debug message. So I removed it.

### Testing Strategy

This pull request was tested by verifying UART now works on the nano33ble.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [ ] Ran `make format`.
- [ ] Fixed errors surfaced by `make clippy`.

### DIFF
--- a/chips/nrf52/src/uart.rs
+++ b/chips/nrf52/src/uart.rs
@@ -448,6 +448,14 @@ impl<'a> Uarte<'a> {
         self.set_tx_dma_pointer_to_buffer();
 
         let regs = &*self.registers;
+
+        // Make sure we clear the endtx interrupt since that is what we rely on
+        // to know when the DMA TX finishes. Normally, we clear this interrupt
+        // as we handle it, so this is not necessary. However, a bootloader (or
+        // some other startup code) may have setup TX interrupts, and there may
+        // be one pending. We clear it to be safe.
+        regs.event_endtx.write(Event::READY::CLEAR);
+
         regs.txd_maxcnt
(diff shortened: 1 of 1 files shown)


chips/nrf52/src/uart.rs
```diff
@@ -448,6 +440,14 @@ impl<'a> Uarte<'a> {
         self.set_tx_dma_pointer_to_buffer();
 
         let regs = &*self.registers;
+
+        // Make sure we clear the endtx interrupt since that is what we rely on
+        // to know when the DMA TX finishes. Normally, we clear this interrupt
+        // as we handle it, so this is not necessary. However, a bootloader (or
+        // some other startup code) may have setup TX interrupts, and there may
+        // be one pending. We clear it to be safe.
+        regs.event_endtx.write(Event::READY::CLEAR);
```
brghena, line 449: Do we need to do this every time when setting up the buffer, or is once in initialize enough?

bradjc: Turns out we don't. Fixed.

PR #1889, opened 2020-05-29
Apollo3: Add an alarm

alistair23:
### Pull Request Overview

Adds the alarm component to the Apollo3 SoC and Artemis Redboard Nano board.

### Testing Strategy

Running the libtock-rs blinky app on the board.

### Documentation Updated

### Formatting

- [X] Ran `make format`.

### DIFF
--- a/boards/redboard_artemis_nano/src/main.rs
+++ b/boards/redboard_artemis_nano/src/main.rs
@@ -8,6 +8,7 @@
 #![cfg_attr(not(doc), no_main)]
 #![deny(missing_docs)]
 
+use capsules::virtual_alarm::VirtualMuxAlarm;
 use kernel::capabilities;
 use kernel::common::dynamic_deferred_call::DynamicDeferredCall;
 use kernel::common::dynamic_deferred_call::DynamicDeferredCallClientState;
@@ -43,6 +44,10 @@ pub static mut STACK_MEMORY: [u8; 0x1000] = [0; 0x1000];
 /// A structure representing this platform that holds references to all
 /// capsules for this platform.
 struct RedboardArtemisNano {
+    alarm: &'static capsules::alarm::AlarmDriver<
+        'static,
+        VirtualMuxAlarm<'static, apollo3::stimer::STimer<'static>>,
+    >,
     led: &'static capsules::led::LED<'static, apollo3::gpio::GpioPin>,
     gpio: &'static capsules::gpio::GPIO<'static, apollo3::gpio::GpioPin>,
     console: &'static capsules::console::Console<'static>,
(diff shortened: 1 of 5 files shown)


boards/redboard_artemis_nano/src/main.rs
```diff
@@ -143,6 +150,30 @@ pub unsafe fn reset_handler() {
...
+    );
+    let alarm = static_init!(
+        capsules::alarm::AlarmDriver<'static, VirtualMuxAlarm<'static, apollo3::stimer::STimer>>,
+        capsules::alarm::AlarmDriver::new(
+            virtual_alarm_user,
+            board_kernel.create_grant(&memory_allocation_cap)
+        )
+    );
+    hil::time::Alarm::set_client(virtual_alarm_user, alarm);
+
```
hudson-ayers, line 176: Could you use the alarm components here instead?

boards/redboard_artemis_nano/src/main.rs
```diff
@@ -143,6 +149,18 @@ pub unsafe fn reset_handler() {
     )
     .finalize(components::gpio_component_buf!(apollo3::gpio::GpioPin));
 
+    // Create a shared virtualisation mux layer on top of a single hardware
+    // alarm.
+    let mux_alarm = static_init!(
+        MuxAlarm<'static, apollo3::stimer::STimer>,
+        MuxAlarm::new(&apollo3::stimer::STIMER)
+    );
+    hil::time::Alarm::set_client(&apollo3::stimer::STIMER, mux_alarm);
```
hudson-ayers, line 158: There is also `AlarmMuxComponent` for instantiating the `MuxAlarm` (sorry I know this feels kind of silly, but having components for as much initialization is the first step towards making it easy to reconfigure boards, so I am trying to enforce use of components whereever possible on new boards being added to the repo)

chips/apollo3/src/stimer.rs
```diff
@@ -0,0 +1,183 @@
...
+        let regs = self.registers;
+
+        // Set the clock source
+        regs.stcfg.write(STCFG::CLKSEL::XTAL_DIV2);
+
+        // Enable interrupts
+        regs.stminten.modify(STMINT::COMPAREA::SET);
+
+        // Set the delta
+        regs.scmpr[0].set(tics - regs.sttmr.get());
```
hudson-ayers, line 137: This is a little different from every other `set_alarm()` implementation in Tock right now, because the compare register is set to a delta in the future rather than a precise tick count at which an interrupt will fire (Because the precise tick count is set in hardware). I dont think this is an issue, but may be worth considering as part of the time/alarm HIL redesign cc @phil-levis .

PR #1894, opened 2020-05-29
sam4l: usb: handle return code, long buffers

bradjc:
1. The `OkSetAddress` return type was split off from `Ok`, but not added to the return code handler for the SAM4L. This puts it back.
2. The capsule uses 64 byte buffers, while the sam4l driver originally only used 8 byte buffers. Since we only seem to use the pointer address in the driver, and not the actual length of the buffer, I think it should be fine if the buffers are too long.

### Testing Strategy

This pull request was tested by running the usb test on imix and verifying that my laptop recognizes the USB device.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make format`.


bradjc:
Ok yup I agree. The first one I think is an "extra" check, because the received packet is always 8 bytes. We could remove the check, but it would fail if we set it to 64.

PR #1896, opened 2020-05-31
apollo3: uart: Fix the UART callback

alistair23:
### Pull Request Overview

Fixes the UART callback for short strings.

### Testing Strategy

Running on the RedBoard Artemis Nano.

### Documentation Updated

### Formatting

- [X] Ran `make format`.


bradjc:
Why not call the callback instead of calling `tx_progress()`? I think this implementation is fine, but it gets a little close to the issue where a callback is triggered not from an interrupt callback. I'm worried about future changes or copy/paste problems for new boards.

alistair23:
I think this might be wrong. We weren't clearing the interrupt. I have pushed an update that should fix it as well.

alistair23:
I pushed an update that fixes some other issues as well.

PR #1899, opened 2020-06-01
Add support for SMBus

alistair23:
### Pull Request Overview

Add some functions to support SMBus read/write operations. Although SMBus is similar to I2C it has a few differences, mostly in regards to frequency, ACK/NACK, timings and ARP.

Some I2C hardware exposes timing options that can be tweaked to better match the SMBus timing requirements. This is why we are adding two new functions. It tells the driver to tweak the timing to match the SMBus spec.

This PR adds new functions to the I2C HIL and also extends the `virtual_i2c.rs` to Mux SMBus devices with I2C devices.

### Testing Strategy

With some patches on top of this I can communicate with an SMBus IR temperature sensor. Once this is merged I'll send a PR to add the sensor.

### Documentation Updated

### Formatting

- [X] Ran `make format`.

### DIFF
--- a/boards/components/src/i2c.rs
+++ b/boards/components/src/i2c.rs
@@ -17,6 +17,7 @@
 
 use capsules::virtual_i2c::{I2CDevice, MuxI2C};
 use core::mem::MaybeUninit;
+use kernel::common::dynamic_deferred_call::DynamicDeferredCall;
 use kernel::component::Component;
 use kernel::hil::i2c;
 use kernel::static_init_half;
@@ -45,6 +46,7 @@ macro_rules! i2c_component_helper {
 pub struct I2CMuxComponent {
     i2c: &'static dyn i2c::I2CMaster,
     smbus: Option<&'static dyn i2c::SMBusMaster>,
+    deferred_caller: &'static DynamicDeferredCall,
 }
 
 pub struct I2CComponent {
@@ -56,8 +58,13 @@ impl I2CMuxComponent {
     pub fn new(
         i2c: &'static dyn i2c::I2CMaster,
         smbus: Option<&'static dyn i2c::SMBusMaster>,
+        deferred_caller: &'static DynamicDeferredCall,
     ) -> Self {
-        I2CMuxComponent { i2c, smbus }
+        I2CMuxComponent {
+            i2c,
+            smbus,
+            deferred_caller,
+        }
     }
 }
 
@@ -69,7 +76,13 @@ impl Component for I2CMuxComponent {
         let mux_i2c = static_init_half!(
             static_buffer,
             MuxI2C<'static>,
-            MuxI2C::new(self.i2c, self.smbus)
+            MuxI2C::new(self.i2c, self.smbus, self.deferred_caller)
+        );
+
+        mux_i2c.initialize_callback_handle(
+            self.deferred_caller
+                .register(mux_i2c)
+                .expect("no deferred call slot available for I2C mux"),
         );
 
         self.i2c.set_master_client(mux_i2c);
--- a/boards/components/src/i2c.rs
+++ b/boards/components/src/i2c.rs
@@ -44,6 +44,7 @@ macro_rules! i2c_component_helper {
 
 pub struct I2CMuxComponent {
     i2c: &'static dyn i2c::I2CMaster,
+    smbus: Option<&'static dyn i2c::SMBusMaster>,
 }
 
 pub struct I2CComponent {
@@ -52,8 +53,11 @@ pub struct I2CComponent {
 }
 
 impl I2CMuxComponent {
-    pub fn new(i2c: &'static dyn i2c::I2CMaster) -> Self {
-        I2CMuxComponent { i2c: i2c }
+    pub fn new(
+        i2c: &'static dyn i2c::I2CMaster,
+        smbus: Option<&'static dyn i2c::SMBusMaster>,
+    ) -> Self {
+        I2CMuxComponent { i2c, smbus }
     }
 }
 
@@ -62,7 +66,11 @@ impl Component for I2CMuxComponent {
     type Output = &'static MuxI2C<'static>;
 
     unsafe fn finalize(self, static_buffer: Self::StaticInput) -> Self::Output {
-        let mux_i2c = static_init_half!(static_buffer, MuxI2C<'static>, MuxI2C::new(self.i2c));
+        let mux_i2c = static_init_half!(
+            static_buffer,
+            MuxI2C<'static>,
+            MuxI2C::new(self.i2c, self.smbus)
+        );
 
         self.i2c.set_master_client(mux_i2c);
 
--- a/capsules/src/virtual_i2c.rs
+++ b/capsules/src/virtual_i2c.rs
@@ -5,6 +5,9 @@
 
 use core::cell::Cell;
 use kernel::common::cells::{OptionalCell, TakeCell};
+use kernel::common::dynamic_deferred_call::{
+    DeferredCallHandle, DynamicDeferredCall, DynamicDeferredCallClient,
+};
 use kernel::common::{List, ListLink, ListNode};
 use kernel::hil::i2c::{self, Error, I2CClient, I2CHwMasterClient};
 
@@ -16,6 +19,8 @@ pub struct MuxI2C<'a> {
     enabled: Cell<usize>,
     i2c_inflight: OptionalCell<&'a I2CDevice<'a>>,
     smbus_inflight: OptionalCell<&'a SMBusDevice<'a>>,
+    deferred_caller: &'a DynamicDeferredCall,
+    handle: OptionalCell<DeferredCallHandle>,
 }
 
 impl I2CHwMasterClient for MuxI2C<'_> {
@@ -37,6 +42,7 @@ impl<'a> MuxI2C<'a> {
     pub const fn new(
         i2c: &'a dyn i2c::I2CMaster,
         smbus: Option<&'a dyn i2c::SMBusMaster>,
+        deferred_caller: &'a DynamicDeferredCall,
     ) -> MuxI2C<'a> {
         MuxI2C {
             i2c: i2c,
@@ -46,7 +52,13 @@ impl<'a> MuxI2C<'a> {
             enabled: Cell::new(0),
             i2c_inflight: OptionalCell::empty(),
             smbus_inflight: OptionalCell::empty(),
+            deferred_caller: deferred_caller,
+            handle: OptionalCell::empty(),
         }
+    }
+
+    pub fn initialize_callback_handle(&self, handle: DeferredCallHandle) {
+        self.handle.replace(handle);
     }
 
     fn enable(&self) {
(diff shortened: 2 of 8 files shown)


alistair23:
The idea with the default implementation is that we can do something like this:

```rust
match self.smbus_dev.smbus_write_read(buf, 1, 1) {
    Ok(_) => {}
    Err(e) => {
        if e.0 == Error::NotSupported {
            // SMBus operations aren't supported. Try normal I2C
            self.smbus_dev.write_read(e.1, 1, 1)
        }
    }
}
```

alistair23:
Yeah, I agree that is confusing. I originally had the functions as part of the I2C traits to not imply that we support SMBus.

I'm expecting the `virtual_smbus` or something similar to handle the fallback. The reason is that some I2C modules don't expose anything to software to correctly handle SMBus. In that case I think we should just fall-back to the standard I2C commands (some SMBus devices are lenient with the spec and should work).

In saying that most Tock I2C devices use 400kHz and SMBus is 100kHz, so it most likely won't work without the hardware having at least some implementation for these new functions.

alistair23:
I pushed an update that fleshes this out more.

capsules/src/virtual_i2c.rs
```diff
@@ -68,8 +81,41 @@ impl<'a> MuxI2C<'a> {
...
+                let mnode = self
+                    .smbus_devices
+                    .iter()
+                    .find(|node| node.operation.get() != Op::Idle);
+                mnode.map(|node| {
+                    node.buffer.take().map(|buf| match node.operation.get() {
+                        Op::Write(len) => {
+                            match self.i2c.smbus_write(node.addr, buf, len) {
+                                Ok(_) => {}
+                                Err(e) => self.command_complete(e.1, e.0),
```
bradjc, line 98: One issue is if the `smbus_write` call returns an error, we can't unconditionally call a callback here, since `do_next_op()` is called from the functions in the interface. That would cause a callback in the middle of the original call, which is against the execution model. There would have to be a deferred call before calling `do_next_op()`, as in virtual_uart.

alistair23: Fixed. It now uses deferred calls.

kernel/src/hil/i2c.rs
```diff
@@ -53,6 +58,73 @@ pub trait I2CMaster {
...
+    /// the hardware can support.
+    /// This function is expected to make any hardware changes required to
+    /// support SMBus and then revert those changes to support future I2C.
+    ///
+    /// addr: The address of the device to write to
+    /// data: The buffer to write the data from and read back to
+    /// write_len: The length of the write operation
+    /// read_len: The length of the read operation
+    #[allow(unused_variables)]
+    fn smbus_write_read(
```
brghena, line 77: I'm a little conflicted on these and I would very much appreciate thoughts from others.

It doesn't seem to me that all I2C masters should be required to implement SMBus functionality. If that's the case, then we should probably split the SMBus stuff into a separate trait, where an I2C master could implement either both or only I2CMaster.

However, you already provided defaults for the SMBus functions, which means it is allowable to not implement them. Further, by modifying the virtual_i2c to require SMBus support (even if just to return an error), you're essentially requiring that all I2C implementations would have to support both. In practice with could be a single `impl SMBusMaster for XXX {}` (which would use the default implementations) for everything, which seems pointless.

Also, I don't think it was a bad idea to include SMBus devices as part of the virtual_i2c. You could imagine a copy of the code that was `virtual_i2c_smbus` and requires an i2c that implements the SMBus trait, but it would just be code duplication that has upkeep problems.

So I'm left unsure how I feel.

alistair23: I also felt conflicted about splitting them. If split then it makes the `virtual_i2c` support much more difficult. Currently we can mux `I2CDevices` and `SMBusDevices` with the `I2CMaster` driver. If we split the implementations (but the underlying hardware will be a single device) then we can't mux virtual_i2c and virtual_smbus without another layer of virtulisation. So it becomes much more complex.

Splitting the two would required more virtulisation without much gain, as I don't see any chips that support just a SMBus device.

brghena: Well, you could split them and then `virtual_i2c` could require an i2c that implements `I2CMaster + SMBusMaster`. That would allow them to be split without requiring any additional verification. But that's also the case above where every i2c chip driver ends up with the additional line: `impl SMBusMaster for XXX {};` just to implement the default functions for SMBus that return errors so that they work with `virtual_i2c`, which seems pretty pointless.

hudson-ayers:
Personally, I think I am okay with the HIL changes, but may need to think about them a little more.

However I don't really like adding a new SMBusDevice struct that is exactly identical to the I2CDevice struct, and implements the same traits with identical code...could we just modify the I2CDevice to have a SMBusMode boolean? Or a mode field that is an enum for either I2CMode or SMBusMode, and that defaults to I2CMode?

hudson-ayers:
Would it be possible to separate the I2CMaster and SMBusMaster HILs, and then have the virtualI2C capsule look like this:
```
pub struct MuxI2C<'a> {
    i2c: &'a dyn hil::i2c::I2CMaster,
    smbus: Option<&'a dyn hil::smbus::SMBusMaster>,
    devices: List<'a, I2CDevice<'a>>,
    i2c_devices: List<'a, I2CDevice<'a>>,
    smbus_devices: List<'a, SMBusDevice<'a>>,
    enabled: Cell<usize>,
    inflight: OptionalCell<&'a I2CDevice<'a>>,
    i2c_inflight: OptionalCell<&'a I2CDevice<'a>>,
    smbus_inflight: OptionalCell<&'a SMBusDevice<'a>>,
    deferred_caller: &'a DynamicDeferredCall,
    handle: OptionalCell<DeferredCallHandle>,
}
```

Then any board supporting smbus could just pass the same trait object for the i2c and smbus fields, while boards not supporting smbus would pass `None` for the smbus field. Ideally, the i2c component could check at compile-time whether the smbus trait was implemented and pass the appropriate value to the `MuxI2C::new()`.

alistair23:
The problem with that is that we then can't mux the `i2c` and `smbus` devices. How do we ensure that only a single operation is going on?

hudson-ayers:
> The problem with that is that we then can't mux the i2c and smbus devices. How do we ensure that only a single operation is going on?

In this scenario virtual_i2c.rs is still responsible for muxing both i2c and smbus -- it just lets the SMBus HIL be completely seperate from the I2C HIL. But `MuxI2C::do_next_op()`, for example, would remain basically as it is now, but lines like `self.i2c.smbus_read()` would be replaced with `self.smbus.unwrap().read()`.

I can try to prototype this if that doesn't make sense

hudson-ayers:
https://github.com/hudson-ayers/tock/tree/smbus is what I have in mind. It seperates the SMBusMaster HIL from the I2CMaster, and removes the separate SMBusDevice struct in favor of adding a boolean to the I2CDevice struct. 

It requires a lot less changes to the virtual_i2c capsule and *I think* lets you do everything that the current implementation does, but I am not totally familiar with SMBus so I could be wrong. Let me know what you think @alistair23

alistair23:
I have pushed an update that follows what Hudson did and separates the SMBusMaster trait.

I prefer the separate SMBusDevice so I have left that in. Let me know what people think. If we think we should squash the device and use a bool I can do that instead.

kernel/src/hil/i2c.rs
```diff
@@ -113,6 +181,69 @@ pub trait I2CDevice {
...
+        &self,
+        data: &'static mut [u8],
+        write_len: u8,
+        read_len: u8,
+    ) -> Result<(), (Error, &'static mut [u8])> {
+        Err((Error::NotSupported, data))
+    }
```
hudson-ayers, line 206: why provide a default implementation for these? There are already implementations on the SMBusDevice struct in virtual_i2c.rs, and normal I2CDevices don't have to implement this trait anyway.

alistair23: Removed

boards/components/src/i2c.rs
```diff
@@ -52,8 +54,14 @@ pub struct I2CComponent {
 }
 
 impl I2CMuxComponent {
-    pub fn new(i2c: &'static dyn i2c::I2CMaster) -> Self {
-        I2CMuxComponent { i2c: i2c }
+    pub fn new(
+        i2c: &'static dyn i2c::I2CMaster,
+        deferred_caller: &'static DynamicDeferredCall,
```
brghena, line 59: Should this not also take in a `smbus: Option<&'a dyn i2c::SMBusMaster>` and pipe it to the MuxI2C::new call?

capsules/src/virtual_i2c.rs
```diff
@@ -64,14 +94,87 @@ impl<'a> MuxI2C<'a> {
...
+                        }
                         Op::Idle => {} // Can't get here...
                     }
                 });
                 node.operation.set(Op::Idle);
-                self.inflight.set(node);
+                self.i2c_inflight.set(node);
             });
+
+            if self.i2c_inflight.is_none() && self.smbus.is_some() {
```
brghena, line 107: Don't you also need to check that `self.smbus_inflight.is_none()` here?

capsules/src/virtual_i2c.rs
```diff
@@ -5,35 +5,62 @@
...
+        }
         self.do_next_op();
     }
 }
 
 impl<'a> MuxI2C<'a> {
-    pub const fn new(i2c: &'a dyn i2c::I2CMaster) -> MuxI2C<'a> {
+    pub const fn new(
+        i2c: &'a dyn i2c::I2CMaster,
+        smbus: Option<&'a dyn i2c::SMBusMaster>,
```
brghena, line 44: Did anyone test that this actually works if you pass in the same `i2c` as both arguments? This seems like the kind of double-reference thing Rust might have problems with.

boards/components/src/i2c.rs
brghena: It would be fine to leave this as-is if you think the plan will be to have a different component for SMBus use.

kernel/src/hil/i2c.rs
brghena: This implementation solves my concerns and it avoids too much duplication!

capsules/src/virtual_i2c.rs
alistair23: I compile tested it, Rust seems happy

capsules/src/virtual_i2c.rs
alistair23: We are still inside the `if self.i2c_inflight.is_none() && self.smbus_inflight.is_none() {` statement.

boards/components/src/i2c.rs
alistair23: Yes, I have fixed this. The board can specify to use the same as I2C or something else

capsules/src/virtual_i2c.rs
brghena: Oops. Sorry about that! I didn't see the indentation level.

Issue #1893, opened 2020-05-29
Tracking: Virtual serial over USB


brghena:
Tracking items:

- [x] Implement some sort of check so that the CDC layer doesn't send data before the USB stack is ready.
- [x] Allow boards to configure the `STRINGS` for the CDC descriptors.
- [x] Decide on a VENDOR_ID and PRODUCT_ID.

**Motivation:**
Many news boards with the nRF52840 that would otherwise be great for Tock use the native USB peripheral on the nRF52840 to connect to it. To make use of these boards, Tock should implement a virtual serial interface over this USB peripheral. This could then be used both for debugging at runtime and secondarily for creating a USB bootloader.

**Background:**
Virtual serial ports can be created through the CDC (Communications Device Class) through the ACM (Abstract Control Model) subclass. This requires the use of three endpoints, one `Interrupt IN` for status notifications (referred to as the "communication" interface) and a `Bulk IN` and `Bulk OUT` for actually sending the serial data (referred to collectively as the "data" interface). The descriptors for this class include two interface descriptors, and several class-specific descriptors as shown in the image below. Here's a good article explaining CDC serial ports specifically: http://janaxelson.com/usb_virtual_com_port.htm
and here's general documentation on descriptors (and general USB background): https://www.beyondlogic.org/usbnutshell/usb5.shtml

![vcp_figure2](https://user-images.githubusercontent.com/235448/83300134-48707d80-a1ac-11ea-9430-621127043d88.jpg)

**Existing Works:**
There are several implementations of CDC serial consoles for the nRF52840. Particularly, the [TinyUSB](https://github.com/hathach/tinyusb) implementation seems pretty trustworthy and worth looking at to understand descriptor details and a minimal implementation of the command interface.

[OpenSK](https://github.com/google/OpenSK) also has an implementation of a USB HID class that works with Tock! The code is easiest to look at locally after setting up a repo as it is in a [patch file](https://github.com/google/OpenSK/blob/master/patches/tock/02-usb.patch). `usb_ctap_hid.rs` mostly layers on top of `usbc_client_ctrl.rs`, which layers on top of the USB HIL, although both the `usb_ctap_hid` and `usbc_client_ctrl` make direct calls to the USB HIL. The file creates several descriptors for a single interface.

**Plan:**
[...]

Issue #1048, opened 2018-06-28
Tracking: Implement serial over USB


bradjc:
Not requiring an FTDI chip would be awesome, and since modern MCUs have USB support this is at least plausible on our current platforms.

- [ ] Implement the CDC ACM profile for USB. #1902
- [ ] See if our USB code can be ported to the nRF52.
- [ ] Verify that Tockloader can work with the CDC port. #1902

bradjc:
This will be very important to support the https://store.arduino.cc/usa/nano-33-ble board correctly.

PR #1901, opened 2020-06-02
sam4l: usb: don't just queue resume in

bradjc:
### Pull Request Overview

While working on #1048 #1893, getting UART.tx to work (sending from device to host) I noticed that transmissions only worked when a character was received. I tracked it down to this change.

Basically, we want to ask the host to send by asking to resume the IN endpoint. The SAM4L driver handled our request, but then queued it until some later event (hence why it didn't work until some receive event). With this change the SAM4L driver now immediately handles our resume in request.

I'm not sure why it was queued in the first place.

### Testing Strategy

This pull request was tested by using hello_loop and console_recv apps on imix with the CDC driver.

### TODO or Help Wanted

What am I missing? What was the queue for? Maybe this change is good enough since I don't know how much interest there is in debugging the SAM4L USB stack.

### Documentation Updated

### Formatting

- [x] Ran `make format`.

### DIFF
--- a/chips/sam4l/src/usbc/mod.rs
+++ b/chips/sam4l/src/usbc/mod.rs
@@ -1558,8 +1558,7 @@ impl<'a> hil::usb::UsbController<'a> for Usbc<'a> {
         requests.resume_in = true;
         self.requests[endpoint].set(requests);
 
-        // TODO: Not sure why this needs to be manually called. Not sure why
-        // it wasn't before.
+        // Immediately handle the request to resume the endpoint.
         self.handle_requests();
     }
 
@@ -1568,8 +1567,7 @@ impl<'a> hil::usb::UsbController<'a> for Usbc<'a> {
         requests.resume_out = true;
         self.requests[endpoint].set(requests);
 
-        // TODO: Not sure why this needs to be manually called. Not sure why
-        // it wasn't before.
+        // Immediately handle the request to resume the endpoint.
         self.handle_requests();
     }
 }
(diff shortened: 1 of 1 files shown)


chips/sam4l/src/usbc/mod.rs
```diff
@@ -1557,12 +1557,20 @@ impl<'a> hil::usb::UsbController<'a> for Usbc<'a> {
         let mut requests = self.requests[endpoint].get();
         requests.resume_in = true;
         self.requests[endpoint].set(requests);
+
+        // TODO: Not sure why this needs to be manually called. Not sure why
+        // it wasn't before.
```
brghena, line 1562:
```suggestion
        // Immediately handle the request to resume the endpoint.
```

chips/sam4l/src/usbc/mod.rs
```diff
@@ -1557,12 +1557,20 @@ impl<'a> hil::usb::UsbController<'a> for Usbc<'a> {
...
+        self.handle_requests();
     }
 
     fn endpoint_resume_out(&self, endpoint: usize) {
         let mut requests = self.requests[endpoint].get();
         requests.resume_out = true;
         self.requests[endpoint].set(requests);
+
+        // TODO: Not sure why this needs to be manually called. Not sure why
+        // it wasn't before.
```
brghena, line 1572:
```suggestion
        // Immediately handle the request to resume the endpoint.
```

brghena:
My opinion is that there seems like some unnecessary design choices in the SAM4L USB stack. However, since we're more focused on the nRF platforms in the long run, doing little things to get the SAM4L working in mostly in its current state seems good.

PR #1902, opened 2020-06-02
USB: Add CDC support, update capsule stack

bradjc:
### Pull Request Overview

This pull request is part of #1048, #1893.

This adds CDC support which implements `kernel::hil::Uart`. This allows us to run console over USB without having to write our own host-side kernel drivers.

On that path, this also updates the capsules USB stack to allow for the larger descriptors we use with the CDC device. This is also the first time we have needed multiple interface descriptors, so it adds support for that as well.

The bug fixes needed along the way are in separate PRs.

### Testing Strategy

This pull request was tested by running apps that use the console on imix, and setting up the console to use UART over CDC rather than the UART driver.

### TODO or Help Wanted

This does not work on the nRF52 (at least on my computer) because that driver does not support CTRL WRITE messages. FYI.

I had to add a platform-specific variable, since the nRF only works if the device descriptor says the packet size is 64, and the SAM4L driver only works if the device descriptor says the packet size is 8. This also makes the USB stack work on the SAM4L again.

### Documentation Updated

### Formatting

- [x] Ran `make format`.

### DIFF
--- a/capsules/src/usb/cdc.rs
+++ b/capsules/src/usb/cdc.rs
@@ -38,6 +38,9 @@ static STRINGS: &'static [&'static str] = &[
     "aSerial No. 5",   // Serial number
 ];
 
+pub const MAX_CTRL_PACKET_SIZE_SAM4L: u8 = 8;
+pub const MAX_CTRL_PACKET_SIZE_NRF52840: u8 = 64;
+
 const N_ENDPOINTS: usize = 3;
 
 pub struct Cdc<'a, C: 'a> {
@@ -61,7 +64,7 @@ pub struct Cdc<'a, C: 'a> {
 }
 
 impl<'a, C: hil::usb::UsbController<'a>> Cdc<'a, C> {
-    pub fn new(controller: &'a C) -> Self {
+    pub fn new(controller: &'a C, max_ctrl_packet_size: u8) -> Self {
         let interfaces: &mut [InterfaceDescriptor] = &mut [
             InterfaceDescriptor {
                 interface_number: 0,
@@ -141,7 +144,7 @@ impl<'a, C: hil::usb::UsbController<'a>> Cdc<'a, C> {
                     product_string: 2,
                     serial_number_string: 3,
                     class: 0x2, // Class: CDC
-                    max_packet_size_ep0: 64,
+                    max_packet_size_ep0: max_ctrl_packet_size,
                     ..descriptors::DeviceDescriptor::default()
                 },
                 descriptors::ConfigurationDescriptor {
--- a/capsules/src/usb/cdc.rs
+++ b/capsules/src/usb/cdc.rs
@@ -128,7 +128,7 @@ impl<'a, U: hil::usb::UsbController<'a>> CdcAcm<'a, U> {
                 endpoint_address: EndpointAddress::new_const(4, TransferDirection::DeviceToHost),
                 transfer_type: TransferType::Interrupt,
                 max_packet_size: 8,
-                interval: 100,
+                interval: 16,
             }],
             &[
                 EndpointDescriptor {
@@ -138,7 +138,7 @@ impl<'a, U: hil::usb::UsbController<'a>> CdcAcm<'a, U> {
                     ),
                     transfer_type: TransferType::Bulk,
                     max_packet_size: 64,
-                    interval: 100,
+                    interval: 0,
                 },
                 EndpointDescriptor {
                     endpoint_address: EndpointAddress::new_const(
@@ -147,7 +147,7 @@ impl<'a, U: hil::usb::UsbController<'a>> CdcAcm<'a, U> {
                     ),
                     transfer_type: TransferType::Bulk,
                     max_packet_size: 64,
-                    interval: 100,
+                    interval: 0,
                 },
             ],
         ];
--- a/capsules/src/usb/cdc.rs
+++ b/capsules/src/usb/cdc.rs
@@ -141,6 +141,7 @@ impl<'a, C: hil::usb::UsbController<'a>> Cdc<'a, C> {
                     product_string: 2,
                     serial_number_string: 3,
                     class: 0x2, // Class: CDC
+                    max_packet_size_ep0: 64,
                     ..descriptors::DeviceDescriptor::default()
                 },
                 descriptors::ConfigurationDescriptor {
--- a/capsules/src/usb/cdc.rs
+++ b/capsules/src/usb/cdc.rs
@@ -1,18 +1,31 @@
 //! CDC
 
-use kernel::common::cells::OptionalCell;
+use core::cmp;
+use core::cell::Cell;
+
 use super::descriptors::{
-    self, Buffer8, EndpointAddress, CsInterfaceDescriptor, EndpointDescriptor, InterfaceDescriptor, TransferDirection,
+    self, Buffer64, EndpointAddress, CsInterfaceDescriptor, EndpointDescriptor, InterfaceDescriptor, TransferDirection,
 };
 use super::usbc_client_ctrl::ClientCtrl;
-use core::cell::Cell;
+
+use kernel::common::cells::OptionalCell;
+use kernel::common::cells::TakeCell;
 use kernel::common::cells::VolatileCell;
 use kernel::debug;
 use kernel::hil;
 use kernel::hil::usb::TransferType;
+use kernel::hil::uart;
+use kernel::ReturnCode;
 
 const VENDOR_ID: u16 = 0x6668;
 const PRODUCT_ID: u16 = 0xabce;
+
+/// Identifying number for the endpoint when transferring data from us to the
+/// host.
+const ENDPOINT_IN_NUM: usize = 2;
+/// Identifying number for the endpoint when transferring data from the host to
+/// us.
+const ENDPOINT_OUT_NUM: usize = 3;
 
 static LANGUAGES: &'static [u16; 1] = &[
     0x0409, // English (United States)
@@ -26,67 +39,32 @@ static STRINGS: &'static [&'static str] = &[
 
 const N_ENDPOINTS: usize = 3;
 
-// // Communication interface descriptor
-
-// 0x09,            // Descriptor size in bytes
-// 0x04,            // INTERFACE descriptor type
-// 0x00,            // Interface number
-// 0x00,            // Alternate setting number
-// 0x01,            // Number of endpoints
-// 0x02,            // Class: CDC communication
-// 0x02,            // Subclass: abstract control model
-// 0x02,            // Protocol: V.25ter (AT commands)
-// 0x00,            // Interface string index
-
-// // Data interface descriptor
-
-// 0x09,            // Descriptor size in bytes
-// 0x04,            // INTERFACE descriptor type
-// 0x01,            // Interface number
-// 0x00,            // Alternate setting number
-// 0x02,            // Number of endpoints
-// 0x0a,            // Class: CDC data
-// 0x00,            // Subclass: none
-// 0x00,            // Protocol: none
-// 0x00,            // Interface string index
 
 
-pub struct Client<'a, C: 'a> {
+
+pub struct Cdc<'a, C: 'a> {
     client_ctrl: ClientCtrl<'a, 'static, C>,
 
     // An eight-byte buffer for each endpoint
-    buffers: [Buffer8; N_ENDPOINTS],
+    buffers: [Buffer64; N_ENDPOINTS],
 
-    // State for a debugging feature: A buffer for echoing bulk data
-    // from an OUT endpoint back to an IN endpoint
-    echo_buf: [Cell<u8>; 8], // Must be no larger than endpoint packet buffer
-    echo_len: Cell<usize>,
-    last_char: OptionalCell<u8>,
+    /// A holder reference for the TX buffer we are transmitting from.
+    tx_buffer: TakeCell<'static, [u8]>,
+    /// The number of bytes the client has asked us to send. We track this so we
+    /// can pass it back to the client when the transmission has finished.
+    tx_len: Cell<usize>,
+    /// How many more bytes we need to transmit. This is used in our TX state
+    /// machine.
+    tx_remaining: Cell<usize>,
+    /// Where in the `tx_buffer` we need to start sending from when we continue.
+    tx_offset: Cell<usize>,
+    /// The TX client to use when transmissions finish.
+    tx_client: OptionalCell<&'a dyn uart::TransmitClient>,
 }
 
-// device
-// {
-// 0x12,            // Descriptor size in bytes
-// 0x01,            // DEVICE descriptor type
-// 0x0200,          // USB version, BCD (2.0)
-// 0x02             // Class: CDC
-// 0x00,            // Subclass: none
-// 0x00,            // Protocol: none
-// 0x08,            // Max. packet size, Endpoint 0
-// 0x0925,          // USB Vendor ID
-// 0x9060,          // USB Product ID
-// 0x0100,          // Device release, BCD (1.0)
-// 0x00,            // Manufacturer string index
-// 0x00,            // Product string index
-// 0x01,            // Serial number string index
-// 0x01             // Number of configurations
-// };
 
 
-
-
-
-impl<'a, C: hil::usb::UsbController<'a>> Client<'a, C> {
+impl<'a, C: hil::usb::UsbController<'a>> Cdc<'a, C> {
     pub fn new(controller: &'a C) -> Self {
 
         let interfaces: &mut [InterfaceDescriptor] = &mut [
@@ -110,8 +88,8 @@ impl<'a, C: hil::usb::UsbController<'a>> Client<'a, C> {
         let cdc_descriptors: &mut [CsInterfaceDescriptor] = &mut [
             CsInterfaceDescriptor {
                 subtype: descriptors::CsInterfaceDescriptorSubType::Header,
-                field1: 0x10,
-                field2: 0x11,
+                field1: 0x10, // CDC
+                field2: 0x11, // CDC
             },
             CsInterfaceDescriptor {
                 subtype: descriptors::CsInterfaceDescriptorSubType::CallManagement,
@@ -121,13 +99,13 @@ impl<'a, C: hil::usb::UsbController<'a>> Client<'a, C> {
             CsInterfaceDescriptor {
                 subtype: descriptors::CsInterfaceDescriptorSubType::AbstractControlManagement,
                 field1: 0x06,
-                field2: 0x00,
+                field2: 0x00, // unused
             },
             // make the length work for now......
             CsInterfaceDescriptor {
                 subtype: descriptors::CsInterfaceDescriptorSubType::Union,
-                field1: 0x00,
-                field2: 0x01,
+                field1: 0x00, // Interface 0
+                field2: 0x01, // Interface 1
             },
         ];
 
@@ -143,15 +121,13 @@ impl<'a, C: hil::usb::UsbController<'a>> Client<'a, C> {
                 EndpointDescriptor {
                     endpoint_address: EndpointAddress::new_const(2, TransferDirection::DeviceToHost),
                     transfer_type: TransferType::Bulk,
-                    // max_packet_size: 16,
-                    max_packet_size: 8,
+                    max_packet_size: 64,
                     interval: 100,
                 },
                 EndpointDescriptor {
                     endpoint_address: EndpointAddress::new_const(3, TransferDirection::HostToDevice),
                     transfer_type: TransferType::Bulk,
-                    // max_packet_size: 16,
-                    max_packet_size: 8,
+                    max_packet_size: 64,
                     interval: 100,
                 },
             ]
@@ -177,7 +153,7 @@ impl<'a, C: hil::usb::UsbController<'a>> Client<'a, C> {
                 Some(cdc_descriptors),
                 );
 
-        Client {
+        Cdc {
             client_ctrl: ClientCtrl::new(
                 controller,
                 device_descriptor_buffer,
@@ -187,25 +163,14 @@ impl<'a, C: hil::usb::UsbController<'a>> Client<'a, C> {
                 LANGUAGES,
                 STRINGS,
             ),
-            buffers: Default::default(),
-            echo_buf: Default::default(),
-            echo_len: Cell::new(0),
-            last_char: OptionalCell::empty(),
+            buffers: [Buffer64::default(); N_ENDPOINTS],
+            tx_buffer: TakeCell::empty(),
+            tx_len: Cell::new(0),
+            tx_remaining: Cell::new(0),
+            tx_offset: Cell::new(0),
+            tx_client: OptionalCell::empty(),
         }
     }
-
-    // fn alert_full(&'a self) {
-    //     // Alert the controller that we now have data to send on the Bulk IN endpoint 1
-    //     self.controller().endpoint_resume_in(2);
-    // }
-
-    // fn alert_empty(&'a self) {
-    //     // In case we reported Delay before, alert the controller
-    //     // that we can now receive data on the Bulk OUT endpoint 2
-    //     if self.delayed_out.take() {
-    //         self.controller().endpoint_resume_out(2);
-    //     }
-    // }
 
     #[inline]
     fn controller(&'a self) -> &'a C {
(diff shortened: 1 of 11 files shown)


capsules/src/usb/cdc.rs
```diff
@@ -0,0 +1,477 @@
...
+    /// when we tell the controller we want to resume the IN endpoint (meaning
+    /// we know we have data to send) and afterwards until we return
+    /// `hil::usb::InResult::Delay` from this function. That means we can use
+    /// this as a callback to mean that the transmission finished by waiting
+    /// until this function is called when we don't have anything left to send.
+    fn packet_in(&'a self, transfer_type: TransferType, endpoint: usize) -> hil::usb::InResult {
+        match transfer_type {
+            TransferType::Interrupt => {
+                debug!("interrupt_in({}) not implemented", endpoint);
+                hil::usb::InResult::Error
```
alistair23, line 275: Should this be an `unimplemented!()` instead

capsules/src/usb/cdc.rs
```diff
@@ -0,0 +1,477 @@
...
+/// us.
+const ENDPOINT_OUT_NUM: usize = 3;
+
+static LANGUAGES: &'static [u16; 1] = &[
+    0x0409, // English (United States)
+];
+
+static STRINGS: &'static [&'static str] = &[
+    "TockOS",         // Manufacturer
+    "The Zorpinator", // Product
```
alistair23, line 37: Ha

capsules/src/usb/cdc.rs
```diff
@@ -0,0 +1,477 @@
...
+
+    fn attach(&'a self) {
+        self.client_ctrl.attach();
+    }
+
+    fn bus_reset(&'a self) {
+        // Should the client initiate reconfiguration here?
+        // For now, the hardware layer does it.
+
+        debug!("Bus reset");
```
alistair23, line 236: Should we remove this? For new board bring-up debug prints can cause panics and are a pain.

capsules/src/usb/cdc.rs
```diff
@@ -0,0 +1,477 @@
...
+            TransferType::Interrupt => {
+                debug!("interrupt_in({}) not implemented", endpoint);
+                hil::usb::InResult::Error
+            }
+            TransferType::Bulk => {
+                self.tx_buffer
+                    .take()
+                    .map_or(hil::usb::InResult::Delay, |tx_buf| {
+                        // Check if we have any bytes to send.
+                        let remaining = self.tx_remaining.get();
```
alistair23, line 282: Why not replace `tx_remaining` with `len - tx_offset`

capsules/src/usb/cdc.rs
```diff
@@ -0,0 +1,477 @@
...
+    fn packet_out(
+        &'a self,
+        transfer_type: TransferType,
+        endpoint: usize,
+        packet_bytes: u32,
+    ) -> hil::usb::OutResult {
+        match transfer_type {
+            TransferType::Interrupt => {
+                debug!("interrupt_out({}) not implemented", endpoint);
+                hil::usb::OutResult::Error
```
alistair23, line 341: Same comment as above.

capsules/src/usb/usbc_client_ctrl.rs
```diff
@@ -416,7 +343,11 @@ impl<'a, 'b, C: hil::usb::UsbController<'a>> ClientCtrl<'a, 'b, C> {
                         hil::usb::CtrlSetupResult::ErrGeneric
                     }
                 }
-                _ => hil::usb::CtrlSetupResult::ErrGeneric,
+                //FIXME might need to include one or more CDC descriptors here
+                _ => {
+                    debug!("we get here?");
```
alistair23, line 348: `unimplemented!()`?

capsules/src/usb/cdc.rs
```diff
@@ -0,0 +1,496 @@
...
+            }],
+            &[
+                EndpointDescriptor {
+                    endpoint_address: EndpointAddress::new_const(
+                        2,
+                        TransferDirection::DeviceToHost,
+                    ),
+                    transfer_type: TransferType::Bulk,
+                    max_packet_size: 64,
+                    interval: 100,
```
brghena, line 141: The interval for bulk endpoints is ignored, so setting this to zero would be preferable so people don't puzzle about its meaning. https://www.beyondlogic.org/usbnutshell/usb5.shtml#EndpointDescriptors

```suggestion
                    interval: 0,
```

capsules/src/usb/cdc.rs
```diff
@@ -0,0 +1,496 @@
...
+                },
+                EndpointDescriptor {
+                    endpoint_address: EndpointAddress::new_const(
+                        3,
+                        TransferDirection::HostToDevice,
+                    ),
+                    transfer_type: TransferType::Bulk,
+                    max_packet_size: 64,
+                    interval: 100,
```
brghena, line 150:
```suggestion
                    interval: 0,
```

capsules/src/usb/cdc.rs
```diff
@@ -0,0 +1,496 @@
...
+                field2: 0x01, // Interface 1
+            },
+        ];
+
+        let endpoints: &[&[EndpointDescriptor]] = &[
+            &[EndpointDescriptor {
+                endpoint_address: EndpointAddress::new_const(4, TransferDirection::DeviceToHost),
+                transfer_type: TransferType::Interrupt,
+                max_packet_size: 8,
+                interval: 100,
```
brghena, line 131: I don't think this value actually matters very much since we don't do anything with the interrupt endpoint, but TinyUSB sets their interval to 16 and copying them seems reasonable. https://github.com/hathach/tinyusb/blob/e9b41cb97d17e82287a146aba1159897a3bcfed2/src/device/usbd.h#L211

```suggestion
                interval: 16,
```

capsules/src/usb/cdc.rs
```diff
@@ -0,0 +1,496 @@
...
+
+    fn packet_transmitted(&'a self, _endpoint: usize) {
+        // Check if more to send.
+        self.tx_buffer.take().map(|tx_buf| {
+            // Check if we have any bytes to send.
+            let remaining = self.tx_len.get() - self.tx_offset.get();
+            if remaining > 0 {
+                // We do, so ask to send again.
+                self.tx_buffer.replace(tx_buf);
+                self.controller().endpoint_resume_in(ENDPOINT_IN_NUM);
```
brghena, line 400: Shouldn't we be checking if `ENDPOINT_IN_NUM == _endpoint` here?

capsules/src/usb/usbc_client.rs
```diff
@@ -56,22 +48,57 @@ pub struct Client<'a, C: 'a> {
...
+                endpoint_address: EndpointAddress::new_const(1, TransferDirection::DeviceToHost),
+                transfer_type: TransferType::Bulk,
+                max_packet_size: 8,
+                interval: 100,
+            },
+            EndpointDescriptor {
+                endpoint_address: EndpointAddress::new_const(2, TransferDirection::HostToDevice),
+                transfer_type: TransferType::Bulk,
+                max_packet_size: 8,
+                interval: 100,
```
brghena, line 74: These aren't your fault, but we might as well fix them now so they stop perpetuating.

```suggestion
                interval: 0,
            },
            EndpointDescriptor {
                endpoint_address: EndpointAddress::new_const(2, TransferDirection::HostToDevice),
                transfer_type: TransferType::Bulk,
                max_packet_size: 8,
                interval: 0,
```

capsules/src/usb/usbc_client_ctrl.rs
```diff
@@ -22,58 +34,60 @@ use core::cmp::min;
...
 
 const N_ENDPOINTS: usize = 3;
 
-pub struct ClientCtrl<'a, 'b, C: 'a> {
-    // The hardware controller
-    controller: &'a C,
+/// Handler for USB control endpoint requests.
+pub struct ClientCtrl<'a, 'b, U: 'a> {
+    /// The USB hardware controller.
+    controller: &'a U,
```
gendx, line 44: What's the rationale for renaming the controller type from `C` to `U`? I don't see any "u" in "controller".

capsules/src/usb/usbc_client_ctrl.rs
```diff
@@ -22,58 +34,60 @@ use core::cmp::min;
...
+    /// Buffer containing the byte-packed representation of the device
+    /// descriptor. This is expected to be created and passed from the user of
+    /// `ClientCtrl`.
+    device_descriptor_buffer: DeviceBuffer,
 
-    // ...with only one interface
-    interface_descriptor: InterfaceDescriptor,
+    /// Buffer containing the byte-packed representation of the configuration
+    /// descriptor and all other descriptors for this device.
+    other_descriptor_buffer: DescriptorBuffer,
```
gendx, line 63: I don't think "other" is a meaningful name. Which descriptors are exactly in there? Maybe `configuration_descriptors_buffer` would make more sense?

capsules/src/usb/usbc_client_ctrl.rs
```diff
@@ -22,58 +34,60 @@ use core::cmp::min;
...
-    // For now we only support one configuration...
-    configuration_descriptor: ConfigurationDescriptor,
+    /// Buffer containing the byte-packed representation of the device
+    /// descriptor. This is expected to be created and passed from the user of
+    /// `ClientCtrl`.
+    device_descriptor_buffer: DeviceBuffer,
 
-    // ...with only one interface
-    interface_descriptor: InterfaceDescriptor,
+    /// Buffer containing the byte-packed representation of the configuration
```
gendx, line 61: What does "byte-packed" mean? Do you mean "serialized"? I've heard of "bit packing" before, but never of "byte packing".

capsules/src/usb/usbc_client_ctrl.rs
```diff
@@ -22,58 +34,60 @@ use core::cmp::min;
...
+    /// descriptor and all other descriptors for this device.
+    other_descriptor_buffer: DescriptorBuffer,
 
-    // A list of endpoints for the configuration
-    endpoint_descriptors: &'b [EndpointDescriptor],
-
-    // A HID descriptor for the configuration, if any
+    /// An optional HID descriptor for the configuration. This can be requested
+    /// separately. It should also be included in `other_descriptor_buffer`.
```
gendx, line 66: What do you mean by "should"?

capsules/src/usb/usbc_client_ctrl.rs
```diff
@@ -22,58 +34,60 @@ use core::cmp::min;
...
+    /// requested separately. It should also be included in
+    /// `other_descriptor_buffer`.
     report_descriptor: Option<&'b ReportDescriptor<'b>>,
 
-    // Supported language (only one for now)
+    /// Supported language (only one for now).
     language: &'b [u16; 1],
 
-    // Strings
+    /// Other strings.
```
gendx, line 77: Why "other" strings? These are just "strings" in USB nomenclature (see also "string descriptors"). And I don't see strings other than these.

capsules/src/usb/usbc_client_ctrl.rs
```diff
@@ -274,56 +246,25 @@ impl<'a, 'b, C: hil::usb::UsbController<'a>> ClientCtrl<'a, 'b, C> {
...
-                                }
-
-                                let end = min(len, requested_length as usize);
-                                self.state[endpoint].set(State::CtrlIn(0, end));
-                                hil::usb::CtrlSetupResult::Ok
-                            }
-                            _ => hil::usb::CtrlSetupResult::ErrInvalidConfigurationIndex,
-                        }
-                    }
-                    DescriptorType::Interface => match descriptor_index {
```
gendx, line 317: I haven't tested it. But does anything break if a host sends us a query for an Interface descriptor anyway?

capsules/src/usb/descriptors.rs
```diff
@@ -370,6 +375,183 @@ impl Descriptor for DeviceDescriptor {
...
+        self.len
+    }
+}
+
+/// Buffer for holding the configuration, interface(s), and endpoint(s)
+/// descriptors. Also includes class-specific functional descriptors.
+pub struct DescriptorBuffer {
+    pub buf: [Cell<u8>; 128],
+    pub len: usize,
+}
```
gendx, line 400: From a memory usage point of view, I find it unfortunate to store some descriptors (e.g. HID) twice: one time already serialized here, and another time in the `ClientCtrl` object. Can't we serialize everything on-demand in the `handle_standard_device_request` function, as before?

capsules/src/usb/usbc_client_ctrl.rs
bradjc: It can't be just _any_ controller, but has to be a USB controller. This mirrors other capsules where the types match the actual resource (A for Alarm, for example).

capsules/src/usb/usbc_client_ctrl.rs
bradjc: Well it isn't just configuration descriptors, it really is all other descriptors.

capsules/src/usb/usbc_client_ctrl.rs
bradjc: I mean it in this context https://docs.python.org/3/library/struct.html

capsules/src/usb/usbc_client_ctrl.rs
bradjc: If you want it included in the descriptor list, it should be included. If it doesn't need to be, then it isn't required.

capsules/src/usb/descriptors.rs
bradjc: Would you be willing to add a commit? This code works, and gets us closer to a long standing item on our wishlist. The USB stack is by no means final, and I'm sure there are optimizations possible.

capsules/src/usb/usbc_client_ctrl.rs
bradjc, line 77:
```suggestion
    /// USB strings to provide human readable descriptions of certain descriptor attributes.
```

capsules/src/usb/cdc.rs
bradjc: We only have one IN endpoint.

capsules/src/usb/usbc_client_ctrl.rs
gendx: Yes, but there is some structure in these descriptors, I assume they are not stored there in arbitrary order. Maybe "configuration descriptor tree" would match the concept more closely?

capsules/src/usb/usbc_client_ctrl.rs
gendx: Well, there is no Python in this context, so it's not clear here.

capsules/src/usb/usbc_client_ctrl.rs
gendx: I assume it's not a matter of a developer "wanting" to include it or not, but rather what the USB specs require or recommend.

capsules/src/usb/descriptors.rs
gendx: > Would you be willing to add a commit?

As far as I'm aware, I'm not able to push commits in this branch, as I'm not a "reviewer with write access".

capsules/src/usb/cdc.rs
brghena: Seems fine then.

capsules/src/usb/usbc_client_ctrl.rs
brghena: I don't think  byte-packed is confusing, but we could just throw the word serialized in instead.

```suggestion
    /// Buffer containing the byte-serialized representation of the configuration
```

capsules/src/usb/usbc_client_ctrl.rs
brghena:
```suggestion
    /// separately. It must also be included in `other_descriptor_buffer` if it exists.
```

capsules/src/usb/usbc_client_ctrl.rs
```diff
@@ -22,58 +34,60 @@ use core::cmp::min;
...
-
-    // A HID descriptor for the configuration, if any
+    /// An optional HID descriptor for the configuration. This can be requested
+    /// separately. It should also be included in `other_descriptor_buffer`.
     hid_descriptor: Option<&'b HIDDescriptor<'b>>,
 
-    // A report descriptor for the configuration, if any
+    /// An optional report descriptor for the configuration. This can be
+    /// requested separately. It should also be included in
+    /// `other_descriptor_buffer`.
```
brghena, line 71:
```suggestion
    /// An optional report descriptor for the configuration. This can be
    /// requested separately. It must also be included in
    /// `other_descriptor_buffer` if it exists.
```

capsules/src/usb/usbc_client_ctrl.rs
brghena: Interface descriptor requests are not possible. Per multiple resources, GET_DESCRIPTOR cannot be used to access interfaces or endpoints separately from configurations. https://www.beyondlogic.org/usbnutshell/usb6.shtml So it should be fine to hit the default match case.

capsules/src/usb/descriptors.rs
brghena: I think serializing everything immediately is actually the way to go. The ClientCtrl doesn't actually need to know much about the descriptors other than just hang on to several buffers of memory and provide them when desired. So it doesn't need a full struct sitting around. A way to stop the duplication would be to have separate buffers for endpoints, interfaces, and other stuff (like the HID) and then combine buffers when things are requested by the host. That would allow you to have only one copy of the HID. We didn't go that full route because the HID descriptor isn't very large so just having two copies lying around doesn't seem to matter much.

The other reason to go with a serialized approach is that it was getting difficult to make lifetimes work for the descriptors. The ClientCtrl can't have structs that it copies fields into because it doesn't know which descriptors are going to exist or how many. It's way easier for the ClientCtrl to have an array of bytes with a fixed size that it can just copy descriptors into than having pointers to arrays of structs that you have to reason about the lifetimes of. I'm sure there's some way to get things working that way, but I couldn't figure it out.

brghena: If you think this is really important to change before this PR goes in and you have some ideas about how to do so, I think you could pull this branch into your own Tock repo, make changes, and then make a PR against this branch rather than against master.

capsules/src/usb/usbc_client_ctrl.rs
brghena: This is the other location that same language is used and is still relevant.

capsules/src/usb/descriptors.rs
```diff
@@ -370,6 +375,183 @@ impl Descriptor for DeviceDescriptor {
...
+/// corresponds to the matching index in the interface descriptor list. For
+/// example, if the interface descriptor list contains `[ID1, ID2, ID3]`,
+/// and the endpoint descriptors list is `[[ED1, ED2], [ED3, ED4, ED5],
+/// [ED6]]`, then the third interface descriptor (`ID3`) has one
+/// corresponding endpoint descriptor (`ED6`).
+pub fn create_descriptor_buffers(
+    device_descriptor: DeviceDescriptor,
+    mut configuration_descriptor: ConfigurationDescriptor,
+    interface_descriptor: &mut [InterfaceDescriptor],
+    endpoint_descriptors: &[&[EndpointDescriptor]],
```
gendx, line 422: I find it weird to separate the arrays for interface and endpoint descriptors in this way. The assumption that isn't mentioned in this API is that `interface_descriptor.len() == endpoint_descriptors.len()`.

Could it be possible to have some `InterfaceHierarchy` struct:

```rust
struct InterfaceHierarchy {
    interface: InterfaceDescriptor,
    endpoints: &[EndpointDescriptor],
}
```

Then, this function would take an `interface_hierarchies: &mut [InterfaceHierarchy]` parameter which would logically tie the interfaces to their endpoints.

capsules/src/usb/descriptors.rs
```diff
@@ -370,6 +375,183 @@ impl Descriptor for DeviceDescriptor {
...
+    // Setup certain descriptor fields since now we know the tree of
+    // descriptors.
+
+    // Configuration Descriptor. We assume there is only one configuration
+    // descriptor, since this is very common for most USB devices.
+    configuration_descriptor.num_interfaces = interface_descriptor.len() as u8;
+
+    // Calculate the length of all dependent descriptors.
+    // TODO should we be erroring here if len > 128? Otherwise we'll probably
+    // buffer overrun and panic.
```
gendx, line 500: This should definitely be handled explicitly, e.g. `panic!("Configuration descriptor too large")`. This shouldn't be a problem given that all of the configuration is defined at compile time, so even if this panic happens at runtime, it's reliably triggered depending on the compile-time descriptors.

This is also IMO a reason why these descriptors should rather be serialized on-demand following get-descriptor requests (which then doesn't impose a limit on the total buffer size), rather than eagerly serialized in a temporary buffer here.

capsules/src/usb/usbc_client_ctrl.rs
```diff
@@ -189,7 +161,7 @@ impl<'a, 'b, C: hil::usb::UsbController<'a>> ClientCtrl<'a, 'b, C> {
     }
 
     #[inline]
-    pub fn controller(&'a self) -> &'a C {
+    pub fn controller(&self) -> &'a U {
```
gendx, line 164: I don't get how the lifetime parameter was removed in the input but not in the output.

hudson-ayers: I don't see why the lifetime of the controller should be tied to the lifetime of the ClientCtrl struct? The controller reference is passed into the ClientCtrl struct when it is created.

I imagined the case where a controller with a static lifetime is passed to `ClientCtrl::new()` -- I don't think that should mean that a later reference to the ClientCtrl struct should have to be static in order for `controller()` to return a static reference to the controller.

gendx: Makes sense. So the answer is that the previous lifetime was unnecessarily too restrictive, right?

hudson-ayers: Yes, I believe so

capsules/src/usb/descriptors.rs
bradjc: Yes, of course, but USB is a WIP stack, so I don't think we should block this change (with a tracking issue dating back years) on the most preferred design, instead we should continue our policy of forward progress is good and iterate quickly.

capsules/src/usb/descriptors.rs
brghena: What _is_ the best way to handle this? We don't want `panic!` in capsules generally, but if this overruns we do want a fast, obvious error.

Just leaving a note for now, is honestly probably sufficient for the foreseeable future (at least until a USB capsule cleanup and rewrite) since 128 bytes fits CDC, OpenSK stuff, and also would fit USB Mass Storage or additional HID stuff if we want to expand to that in the future.

brghena: Unfortunately, I think even serializing on demand doesn't actually fix this simply. You would just end up with structs that, when serialized sometime at runtime, end up either overrunning the buffer or more likely getting truncated. You could handle this at runtime, but doing so would require you to account for partial progress in serializing some struct or another and picking up serializing part-way through that struct for the next packet. (You can't just stop at the last struct that fits because I believe you have to send the full 64 byte packet in order to fragment.) At least with the descriptors serialized at initialization time, the error occurs immediately.

capsules/src/usb/descriptors.rs
gendx: > If you think this is really important to change before this PR goes in and you have some ideas about how to do so, I think you could pull this branch into your own Tock repo, make changes, and then make a PR against this branch rather than against master.

It's not really practical to make a PR against another PR to be honest...

capsules/src/usb/descriptors.rs
gendx: > What _is_ the best way to handle this? We don't want `panic!` in capsules generally, but if this overruns we do want a fast, obvious error.

I don't see any `unsafe` in this file. So if the buffer overruns, the current behavior is that it *will panic* at one of the array indexing operations (e.g. within a `write_to` where the target buffer is smaller than the source). So the current code would yield a panic (which we want to avoid apparently), a fast error (in this function), but not so obvious (array indexing out of bounds somewhere).

What I suggest is an explicit panic with a clear message here, which improves over the current state by making the error obvious.

To avoid a panic altogether, the best would probably be to put all of this code in a `const fn` so that the contents of the descriptor buffer is pre-computed at compile time, and that any failure would prevent it from compiling. However, I'm not sure what the current status of `const fn` is in terms of supporting things like loops (although I remember there has been some improvements lately).

capsules/src/usb/descriptors.rs
gendx: Fair point. My point was mostly that in my experience working with USB, a lot of things can be brittle and therefore overlooking design could fire back in terms of long hours of subtle debugging. But sure this can be improved in a separate pull request.

bradjc:
Are there any major blocking issues for this?

bradjc:
@tock/core-wg

gendx:
> I think the only blockers I see are @gendx's comments:
> 
> * [ ]  [This one should be addressed by adding an explicit `panic!`](https://github.com/tock/tock/pull/1902#discussion_r436592017) -- since it will panic either way, better to have something explicit than it happening from an array access.
> * [ ]  [This one should be added to a tracking issue so we don't lose it](https://github.com/tock/tock/pull/1902#discussion_r436591970)
> * [ ]  [This one could either be addressed by a comment, or added to a tracking issue](https://github.com/tock/tock/pull/1902#discussion_r435800117)

+1 to having (a) tracking issue(s) for the comments I left, so that they are not forgotten in the long term once USB stabilizes, but without unnecessarily blocking progress of useful features in the short term.

bradjc:
I created #1940.

PR #1906, opened 2020-06-03
OpenTitan: Document and support flashing apps

alistair23:
### Pull Request Overview

OpenTitan needs the apps built into the binary. Add a section to the documentation on how to do this and add a target to the Makefile.

We need to use the GCC objcopy as the LLVM one doesn't support updating sections.

### Testing Strategy

Flashing an app to the OT board.

### Documentation Updated

### Formatting

- [X] Ran `make format`.

### DIFF
--- a/boards/opentitan/README.md
+++ b/boards/opentitan/README.md
@@ -50,6 +50,42 @@ You can also just use the `spiflash` program manually to download the image to t
 
 NOTE: You will need to download the Tock binary after every power cycle.
 
+Programming Apps
+----------------
+
+Tock apps for OpenTitan must be included in the Tock binary file flashed with the steps mentioned above.
+
+Apps are built out of tree. Currently [libtock-rs](https://github.com/tock/libtock-rs) apps work well while [libtock-c](https://github.com/tock/libtock-c) apps require a special branch and complex work arounds. It is recomended that libtock-rs apps are used.
+
+Once an app is built and a tbf file is generated, you can use `riscv32-none-elf-objcopy` with `--update-section` to create an ELF image with the
+apps included.
+
+```shell
+$ riscv32-oe-elf-objcopy \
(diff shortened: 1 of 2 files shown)


boards/opentitan/README.md
```diff
@@ -50,22 +50,58 @@ You can also just use the `spiflash` program manually to download the image to t
...
 
+Programming Apps
+----------------
+
+Tock apps for OpenTitan must be included in the Tock binary file flashed with the steps mentioned above.
+
+Apps are built out of tree. Currently [libtock-rs](https://github.com/tock/libtock-rs) apps work well while [libtock-c](https://github.com/tock/libtock-c) apps require a special branch and complex work arounds. It is recomended that libtock-rs apps are used.
+
+Once an app is built, you can use `riscv32-none-elf-objcopy` with `--update-section` to create an ELF image with the
+apps included.
```
brghena, line 61: To make the tab file, do users have to run `make flash-opentitan` (possibly with additional flags) in the libtock-rs repo? That is probably worth noting in the README as my initial assumption was that `make opentitan` would be sufficient.

alistair23: Yes, they do have to run `make flash-opentitan` I would rather not document the libtock-rs flow here as that could change and it will be out of sync. It's probably better to improve the libtock-rs flow.

brghena: Yeah, that's a fair concern. Maybe a small statement like:

```suggestion
Once an app is built and TAB file is created, you can use `riscv32-none-elf-objcopy` with `--update-section` to create an ELF image with the
apps included.
```

That at least gives users a hint that they need to do something more than just build the app. And the libtock-rs main README already says that `make flash-<platform>` creates a TAB file.

alistair23: Done.

alistair23: I wrote tbf file though, as we actually use the tbf file.

PR #1907, opened 2020-06-03
nrf52: usb: handle control OUT transfer

bradjc:
### Pull Request Overview

This pull request is part of #1048, #1893.

This adds support for control write (OUT) transfers in the nRF52 USB stack. This is needed for CDC since the host client sends control messages (which we ignore, but still need to respond to).

### Testing Strategy

This pull request was tested by running the CDC capsule and verifying that console tx and rx works using it on the nrf52840dk.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make format`.

### DIFF
--- a/chips/nrf52/src/usbd.rs
+++ b/chips/nrf52/src/usbd.rs
@@ -1409,6 +1409,9 @@ impl<'a> Usbd<'a> {
                 self.client.map(|client| {
                     match client.ctrl_out(endpoint, regs.epout[endpoint].amount()) {
                         hil::usb::CtrlOutResult::Ok => {
+                            // TODO: Check if the CTRL WRITE is longer
+                            // than the amount of data we have received,
+                            // and receive more data before completing.
                             self.complete_ctrl_status();
                         }
                         hil::usb::CtrlOutResult::Delay => {}
--- a/chips/nrf52/src/usbd.rs
+++ b/chips/nrf52/src/usbd.rs
@@ -1361,7 +1361,11 @@ impl<'a> Usbd<'a> {
         let state = self.descriptors[endpoint].state.get().ctrl_state();
         match state {
             CtrlState::ReadIn => {
-                self.transmit_in_ep0();
+                if self.dma_pending.get() {
+                    self.descriptors[endpoint].request_transmit_in.set(true);
+                } else {
+                    self.transmit_in_ep0();
+                }
             }
 
             CtrlState::ReadStatus => {
@@ -1371,8 +1375,12 @@ impl<'a> Usbd<'a> {
             CtrlState::WriteOut => {
                 // We just completed the Setup stage for a CTRL WRITE transfer,
                 // and now we need to enable DMA so the USBD peripheral can copy
-                // the received data.
-                self.transmit_out_ep0();
+                // the received data. If the DMA is in use, queue our request.
+                if self.dma_pending.get() {
+                    self.descriptors[endpoint].request_transmit_out.set(true);
+                } else {
+                    self.transmit_out_ep0();
+                }
             }
 
             CtrlState::Init => {
@@ -1651,8 +1659,15 @@ impl<'a> Usbd<'a> {
                                         self.descriptors[endpoint]
                                             .state
                                             .set(EndpointState::Ctrl(CtrlState::ReadIn));
-                                        // Transmit first packet
-                                        self.transmit_in_ep0();
+                                        // Transmit first packet if DMA is
+                                        // available.
+                                        if self.dma_pending.get() {
+                                            self.descriptors[endpoint]
+                                                .request_transmit_in
+                                                .set(true);
+                                        } else {
+                                            self.transmit_in_ep0();
+                                        }
                                     }
                                     None => unreachable!(),
                                 }
@@ -1698,13 +1713,21 @@ impl<'a> Usbd<'a> {
 
         for (endpoint, desc) in self.descriptors.iter().enumerate() {
             if desc.request_transmit_in.take() {
-                self.transmit_in(endpoint);
+                if endpoint == 0 {
+                    self.transmit_in_ep0();
+                } else {
+                    self.transmit_in(endpoint);
+                }
                 if self.dma_pending.get() {
                     break;
                 }
             }
             if desc.request_transmit_out.take() {
-                self.transmit_out(endpoint);
+                if endpoint == 0 {
+                    self.transmit_out_ep0();
+                } else {
+                    self.transmit_out(endpoint);
+                }
                 if self.dma_pending.get() {
                     break;
                 }
--- a/chips/nrf52/src/usbd.rs
+++ b/chips/nrf52/src/usbd.rs
@@ -324,10 +324,6 @@ mod detail {
             self.ptr.set(slice.as_ptr() as *const u8);
             self.maxcnt.write(Count::MAXCNT.val(slice.len() as u32));
         }
-
-        pub fn amount(&self) -> u32 {
-            self.amount.get()
-        }
     }
 }
 
@@ -1407,7 +1403,7 @@ impl<'a> Usbd<'a> {
                 // Now we can handle it and pass it to the client to see
                 // what the client returns.
                 self.client.map(|client| {
-                    match client.ctrl_out(endpoint, regs.epout[endpoint].amount()) {
+                    match client.ctrl_out(endpoint, regs.size_epout[endpoint].get()) {
                         hil::usb::CtrlOutResult::Ok => {
                             // TODO: Check if the CTRL WRITE is longer
                             // than the amount of data we have received,
--- a/chips/nrf52/src/usbd.rs
+++ b/chips/nrf52/src/usbd.rs
@@ -1370,10 +1370,9 @@ impl<'a> Usbd<'a> {
 
             CtrlState::WriteOut => {
                 // We just completed the Setup stage for a CTRL WRITE transfer,
-                // and the DMA has received data. Next step is to signal
-                // `startepout[0]` to let the hardware move the data to RAM.
-                debug_tasks!("- task: startepout[{}]", endpoint);
-                regs.task_startepout[endpoint].write(Task::ENABLE::SET);
+                // and now we need to enable DMA so the USBD peripheral can copy
+                // the received data.
+                self.transmit_out_ep0();
             }
 
             CtrlState::Init => {
@@ -1405,6 +1404,9 @@ impl<'a> Usbd<'a> {
                 self.client.map(|client| {
                     match client.ctrl_out(endpoint, regs.size_epout[endpoint].get()) {
                         hil::usb::CtrlOutResult::Ok => {
+                            // We only handle the simple case where we have
+                            // received all of the data we need to.
+                            //
                             // TODO: Check if the CTRL WRITE is longer
                             // than the amount of data we have received,
                             // and receive more data before completing.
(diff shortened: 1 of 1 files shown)


brghena:
I don't understand this code. We could merge it and move forward since USB support is still nascent in Tock, but I'd be happier if @gendx has the time to take a quick look at it.

chips/nrf52/src/usbd.rs
```diff
@@ -1563,16 +1605,26 @@ impl<'a> Usbd<'a> {
...
                                     Some(RequestType::DIRECTION::Value::HostToDevice) => {
-                                        unimplemented!("CTRL write transaction");
+                                        // CTRL WRITE transfer with data to
+                                        // receive. We first need to setup DMA
+                                        // so that the hardware can write the
+                                        // data to us.
+                                        self.descriptors[endpoint]
+                                            .state
+                                            .set(EndpointState::Ctrl(CtrlState::WriteOut));
+                                        self.transmit_out_ep0();
```
gendx, line 1627: Here, the specification (https://infocenter.nordicsemi.com/pdf/nRF52840_OPS_v0.5.pdf page 618) invokes the `EP0RCVOUT` task **before** setting up the DMA and transmission. See my above comment, I think it would be much simpler if here we only invoke the `EP0RCVOUT` task, and just wait for the `EP0DATADONE` event.

Once `EP0DATADONE` has arrived, we can set up the DMA transfer like any other endpoint and launch it. See my comment in the `handle_ep0datadone` above.

This would be consistent with the other endpoints, and the sequence of events would be equivalent. Indeed, my understanding is that the specification proposes the sequence of events (1) `EP0RCVOUT` task (2) set up the DMA (3) `EP0DATADONE` event (4) `STARTEPOUT[0]` task.
But once the task (1) has been invoked, (2) and (3) are interchangeable because we have no control over when the USB controller with send the `EP0DATADONE` event. Even if we're fast at executing (2), the event (3) could still happen before we're done because this event happens in parallel in the USB controller. So we might as well wait for (3) to start executing (2), as we do for the other endpoints. What matters is that setting up the DMA in (2) happens before we invoke the task (4).

bradjc: That is not how I read "The software has to prepare EasyDMA by pointing to the buffer in Data RAM that shall contain the incoming data. If no other EasyDMA transfer is on-going with USBD, software can then send the EP0RCVOUT task, which will make the USBD to accept (ACK) the first OUT+DATA transaction from the host."

chips/nrf52/src/usbd.rs
```diff
@@ -1380,10 +1400,29 @@ impl<'a> Usbd<'a> {
...
+                            // Respond with STALL to any following transactions
+                            // in this request
+                            debug_tasks!("- task: ep0stall");
+                            regs.task_ep0stall.write(Task::ENABLE::SET);
+                            self.descriptors[endpoint]
+                                .state
+                                .set(EndpointState::Ctrl(CtrlState::Init));
+                        }
+                    };
+                });
```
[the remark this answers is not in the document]
bradjc: It's not. Don't need it right now so I didn't add the extra complexity.

chips/nrf52/src/usbd.rs
gendx: Meh... I'm not convinced as to why the wording is different for the control endpoint vs. the bulk/interrupt endpoints.

But as I understand it:
- `EP0RCVOUT` task only makes the USBD accept (ACK) the first OUT+DATA transaction from the host. It's the equivalent of "writing any value to `SIZE.EPOUT[n]`" for bulk/interrupt endpoints.
- The DMA transfer won't start until the `STARTEPOUT[i]` task is invoked (for control and bulk/interrupt endpoints alike). So it's only necessary to setup EasyDMA before this point. But not necessarily before `EP0RCVOUT`.
- If one enables the `EP0DATADONE -> STARTEPOUT[0]` shortcut, then of course it's necessary to setup EasyDMA before sending `EP0RCVOUT`, because we have no control over when the `STARTEPOUT[0]` task is triggered. I think that was the spirit of the specs. But if we don't enable this shortcut, we can wait for the `EP0DATADONE` event before setting up the EasyDMA.
- If we wait for the `EP0DATADONE` event to setup EasyDMA, rather than directly after the `EP0SETUP` event, we block EasyDMA (`self.set_pending_dma()`) for a shorter amount of time, which is preferable.

To me, it would be much clearer to have a consistent implementation for all endpoints, rather than special casing the control endpoint beyond necessary.

chips/nrf52/src/usbd.rs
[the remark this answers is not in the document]
gendx: Fair enough, but then it should be documented here as a TODO that only a single control write transfer is currently supported.

chips/nrf52/src/usbd.rs
```diff
@@ -1380,10 +1400,29 @@ impl<'a> Usbd<'a> {
...
+                // We got data on the control endpoint during a CTRL WRITE
+                // transfer. Let the client handle the data, and then finish up
+                // the control write by moving to the status stage.
+
+                // Now we can handle it and pass it to the client to see
+                // what the client returns.
+                self.client.map(|client| {
+                    match client.ctrl_out(endpoint, regs.epout[endpoint].amount()) {
+                        hil::usb::CtrlOutResult::Ok => {
+                            self.complete_ctrl_status();
```
bradjc, line 1412:
```suggestion
                            // TODO: Check if the CTRL WRITE is longer
                            // than the amount of data we have received,
                            // and receive more data before completing.
                            self.complete_ctrl_status();
```

chips/nrf52/src/usbd.rs
bradjc: Doesn't it make sense then to leave the control flow that matches the datasheet to support a future developer who wants to add the shortcuts?

gendx: As Tock we can make the choice not to use these shortcuts. IMO they add a bit of complexity (enabling/disabling them + a control flow that doesn't match the other endpoints) without clear benefits (the stack is a bit faster to reply, but it's only for control so is it necessary/noticeable?).

My suggestion would be to match what happens for the endpoints for the sake of clarity in Tock's code and to avoid code duplication. But of course one has to test whether that works.

TBH I haven't looked at the CDC protocol at all. Is there a test host program for this (in [tools/usb](https://github.com/tock/tock/tree/master/tools/usb) maybe)?

bradjc:
Ok, state machine is updated (I think) to match the IN side.

bradjc:
I added another commit. I wasn't checking if the DMA was active when handling a CTRL OUT transfer, and this can conflict with an ongoing data IN transfer. Now all calls to `transmit_in|out_ep0()` can be queued, just as the `transmit_in|out()` versions.

chips/nrf52/src/usbd.rs
```diff
@@ -1671,6 +1767,16 @@ impl<'a> Usbd<'a> {
...
+    /// Setup a reception for a CTRL WRITE transaction.
+    ///
+    /// We have received the EP0DATADONE event signaling that the host has sent
+    /// us data. We now need to configure DMA so that the peripheral can copy us
+    /// the data.
+    fn transmit_out_ep0(&self) {
+        let endpoint = 0;
+        self.start_dma_out(endpoint);
+    }
+
```
gendx, line 1779: LGTM

chips/nrf52/src/usbd.rs
```diff
@@ -1625,13 +1713,21 @@ impl<'a> Usbd<'a> {
...
                     break;
                 }
             }
             if desc.request_transmit_out.take() {
-                self.transmit_out(endpoint);
+                if endpoint == 0 {
+                    self.transmit_out_ep0();
+                } else {
+                    self.transmit_out(endpoint);
+                }
```
gendx, line 1730: LGTM.

chips/nrf52/src/usbd.rs
```diff
@@ -1346,20 +1352,37 @@ impl<'a> Usbd<'a> {
...
+                // We just completed the Setup stage for a CTRL WRITE transfer,
+                // and now we need to enable DMA so the USBD peripheral can copy
+                // the received data. If the DMA is in use, queue our request.
+                if self.dma_pending.get() {
+                    self.descriptors[endpoint].request_transmit_out.set(true);
+                } else {
+                    self.transmit_out_ep0();
+                }
+            }
+
```
gendx, line 1385: LGTM. Good point on queuing the requests.

chips/nrf52/src/usbd.rs
```diff
@@ -1380,10 +1403,35 @@ impl<'a> Usbd<'a> {
...
+                        hil::usb::CtrlOutResult::Ok => {
+                            // We only handle the simple case where we have
+                            // received all of the data we need to.
+                            //
+                            // TODO: Check if the CTRL WRITE is longer
+                            // than the amount of data we have received,
+                            // and receive more data before completing.
+                            self.complete_ctrl_status();
+                        }
+                        hil::usb::CtrlOutResult::Delay => {}
```
gendx, line 1423: Note: I'm not sure what would cause a delay here. However, I don't see it happening with only 1 packet of ctrl write data, so until we get a use case with multiple ctrl write transfers, LGTM.

chips/nrf52/src/usbd.rs
```diff
@@ -1380,10 +1403,35 @@ impl<'a> Usbd<'a> {
...
+                            // Respond with STALL to any following transactions
+                            // in this request
+                            debug_tasks!("- task: ep0stall");
+                            regs.task_ep0stall.write(Task::ENABLE::SET);
+                            self.descriptors[endpoint]
+                                .state
+                                .set(EndpointState::Ctrl(CtrlState::Init));
+                        }
+                    };
+                });
```
gendx, line 1434: LGTM.

chips/nrf52/src/usbd.rs
```diff
@@ -1594,7 +1682,7 @@ impl<'a> Usbd<'a> {
                 });
             }
 
-            CtrlState::ReadIn | CtrlState::ReadStatus => {
+            CtrlState::ReadIn | CtrlState::ReadStatus | CtrlState::WriteOut => {
```
gendx, line 1685: LGTM.

chips/nrf52/src/usbd.rs
```diff
@@ -1563,23 +1614,60 @@ impl<'a> Usbd<'a> {
...
-                                        self.transmit_in_ep0();
+                                        // Transmit first packet if DMA is
+                                        // available.
+                                        if self.dma_pending.get() {
+                                            self.descriptors[endpoint]
+                                                .request_transmit_in
+                                                .set(true);
+                                        } else {
+                                            self.transmit_in_ep0();
+                                        }
```
gendx, line 1670: LGTM.

chips/nrf52/src/usbd.rs
```diff
@@ -1563,23 +1614,60 @@ impl<'a> Usbd<'a> {
...
+                                        // > EP0RCVOUT task.
+                                        //
+                                        // But, since we are not using the
+                                        // EP0DATADONE->STARTEPOUT[0] shortcut,
+                                        // and DMA only needs to be setup to
+                                        // copy the bytes from the USBD
+                                        // peripheral, we can wait until we get
+                                        // the EP0DATADONE event to enable DMA.
+                                        debug_tasks!("- task: ep0rcvout");
+                                        regs.task_ep0rcvout.write(Task::ENABLE::SET);
```
gendx, line 1656: LGTM.

bradjc:
Yes this has been tested on the nano33ble board the nano33_3 branch. I will keep testing as these PRs are merged, and then we will have a board that uses it by default to make it more noticeable if something breaks.

PR #1908, opened 2020-06-03
Update Rust To June 2020 Nightly

bradjc:
### Pull Request Overview

This pull request updates the nightly we are on.

### Testing Strategy

This pull request was tested by travis.

### TODO or Help Wanted

Are the warnings valid?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers:
the warnings about variables not being read actually caught 2 long standing bugs I believe. The trick in both cases is that using a `move ||` closure causes that closure to take ownership of all variables used in the closure. Once the variables are moved into the closure, they cannot be used again after the closure, so normally doing so would cause a compile error...unless the moved variable implements `Copy` in which case Rust's move semantics state that a move is just a copy! 

So Rust copies the mutable variable into the closure, and then when it is modified in the closure only the *copy* is modified, not the original variable. Then when the variable name is read after the closure, it is referencing the original, unmodified variable, not the one that was copied into the closure (and modified therein)!

Two fixes for this exist: the closure [crate](https://lib.rs/crates/closure) allows for passing some variables by reference and others by value, which is one fix. If you don't want to use an external crate, the solution is to shadow the reference within the closure, forcing it to capture the variable by reference instead of by value (https://stackoverflow.com/questions/58459643/is-there-a-way-to-have-a-rust-closure-that-moves-only-some-variables-into-it). That is the fix I went with.


The AES test clippy failure also looks like a real bug, as I doubt it is intentional to pass an empty slice to the aes test, but I have no idea what the correct slice to pass to that test is..

hudson-ayers:
I went ahead and "fixed" the clippy error by replacing `&BEACON_UNSECURED[26..26]` with `&[]`, as the test was written that way in the original commit and has not changed since.

@bbbert if you know whether this was actually intended to be something different please let me know what it should be instead.

capsules/src/sdcard.rs
```diff
@@ -1444,8 +1444,8 @@ impl<'a, A: hil::time::Alarm<'a>> SDCardClient for SDCardDriver<'a, A> {
     fn read_done(&self, data: &'static mut [u8], len: usize) {
         self.kernel_buf.replace(data);
         self.app.map(|app| {
-            #[allow(unused_mut)]
-            let mut read_len: usize = 0;
+            let mut read_len = 0;
+            let read_len_ref = &mut read_len; // needed to mutate read_len in move closure
```
brghena, line 1448: Presumably this has been a bug for years, but the sdcard testing app I wrote ignores the size of the the read done callback: https://github.com/tock/libtock-c/blob/master/libtock/sdcard.c#L55

Was the required `#[allow(unused_mut))]` we we upgraded to rust 2018 a hint that something was wrong with the code?

hudson-ayers: yep, that was a clue

capsules/src/ieee802154/framer.rs
```diff
@@ -843,6 +843,7 @@ impl<'a, M: Mac, A: AES128CCM<'a>> CCMClient for Framer<'a, M, A> {
     fn crypt_done(&self, buf: &'static mut [u8], res: ReturnCode, tag_is_valid: bool) {
         let mut tx_waiting = false;
         let mut rx_waiting = false;
+        let rx_waiting_ref = &mut rx_waiting; //needed to mutate rx_waiting in move closure
```
ppannuto, line 846: I believe the other way we can do this is changing this closure to no longer be `move` and convincing the closure to capture `buf` explicitly:

```diff
@@ -843,7 +843,6 @@ impl<'a, M: Mac, A: AES128CCM<'a>> CCMClient for Framer<'a, M, A> {
     fn crypt_done(&self, buf: &'static mut [u8], res: ReturnCode, tag_is_valid: bool) {
         let mut tx_waiting = false;
         let mut rx_waiting = false;
-        let rx_waiting_ref = &mut rx_waiting; //needed to mutate rx_waiting in move closure

         // The crypto operation was from the transmission pipeline.
         let opt_buf = if let Some(state) = self.tx_state.take() {
@@ -880,7 +879,8 @@ impl<'a, M: Mac, A: AES128CCM<'a>> CCMClient for Framer<'a, M, A> {

         // The crypto operation was from the reception pipeline.
         if let Some(buf) = opt_buf {
-            self.rx_state.take().map(move |state| {
+            self.rx_state.take().map(|state| {
+                let buf = buf;
                 match state {
                     RxState::Decrypting(info) => {
                         let next_state = if tag_is_valid {
@@ -892,7 +892,7 @@ impl<'a, M: Mac, A: AES128CCM<'a>> CCMClient for Framer<'a, M, A> {
                         self.step_receive_state();
                     }
                     other_state => {
-                        *rx_waiting_ref = match other_state {
+                        rx_waiting = match other_state {
                             RxState::ReadyToDecrypt(_, _) => true,
                             _ => false,
                         };
```

IIUC correctly, the fundamental problem is that `||` and `|move |` choose between referencing captured variables and moving them into the closure respectively, but we want to do one of each. In this case, because `buf` has a `'static` lifetime, which means that `||` can't automatically create a reference to it, which is why this closure was using `move` in the first place. For variables that can't be moved, `|move |` will implicitly `Copy` them if possible. Unfortunately, because `rx_waiting` is used after this closure, it can't safely be moved into the closure, so `move` helpfully implicitly copied it too.

The proposed patch in the PR resolved this by switching `rx_waiting` to a type where the implicit copy is to a reference. This alternative solution eliminates implicit copies (so `rx_waiting` is passed by reference) and adds an explicit copy of `buf`.

I _believe_ these would actually both compile down to the same code in practice. However, I think the implicit copy of `|move |` is more error-prone, and switching to the explicit-copy model here might be safer / clearer in the long run.

[I found this blog post helpful in understanding closure nuances.](https://stevedonovan.github.io/rustifications/2018/08/18/rust-closures-are-hard.html)

brghena: I can definitely say that the sdcard capsule did not take a nuanced approach. `move` closures are just something I apply when the compiler tells me to in an error message.

capsules/src/test/aes_ccm.rs
```diff
@@ -34,7 +34,8 @@ impl<'a, A: AES128CCM<'a>> Test<'a, A> {
             tests: [
                 (
                     &BEACON_UNSECURED[0..26],
-                    &BEACON_UNSECURED[26..26],
+                    //&BEACON_UNSECURED[26..26],
+                    &[],
```
ppannuto, line 38:
```suggestion
                    &[],
```

I get why this was just commented, but I think it would be clearer to just be explicit that it's empty instead of implying that the array matters or that there's something reasonable to uncomment.

PR #1909, opened 2020-06-04
Add Nano 33 BLE Board

bradjc:
### Pull Request Overview

This pull request adds a board main.rs for the long-discussed (see #994) Nano 33 BLE board. Since it is based on the nrf52840, this is pretty straightforward.

The bootloader situation is not the best at the moment, but that is a work in progress. Also right now the UART requires attaching an FTDI chip to the UART pins. However, when the CDC USB stack is merged we can switch to using the USB header.

### Testing Strategy

This pull request was tested by running the hello loop and blink apps on the board.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nano33ble/Makefile
+++ b/boards/nano33ble/Makefile
@@ -10,7 +10,7 @@ ifdef PORT
   FLAGS += --port $(PORT)
 endif
 
-# Upload the kernel over using bossac
+# Upload the kernel using bossac
 .PHONY: program
 program: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
 	bossac $(FLAGS) -U -i -e -w $< -R
--- a/boards/nano33ble/Makefile
+++ b/boards/nano33ble/Makefile
@@ -1,0 +1,16 @@
+# Makefile for building the tock kernel for the Arduino Nano 33 BLE board.
+
+TOCK_ARCH=cortex-m4
+TARGET=thumbv7em-none-eabi
+PLATFORM=nano33ble
+
+include ../Makefile.common
+
+ifdef PORT
+  FLAGS += --port $(PORT)
+endif
+
+# Upload the kernel over using bossac
+.PHONY: program
+program: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
+	bossac $(FLAGS) -U -i -e -w $< -R
--- a/boards/nano33ble/README.md
+++ b/boards/nano33ble/README.md
@@ -27,6 +27,9 @@ $ cd BOSSA
 $ make bossac
 ```
 
+Then you will need to add `BOSSA/bin` to your `$PATH` variable so that your
+system can find the `bossac` program.
+
 ## Programming the Kernel
 
 To program the kernel we use the BOSSA tool to communicate with the bootloader
--- a/boards/nano33ble/README.md
+++ b/boards/nano33ble/README.md
@@ -1,0 +1,83 @@
+Arduino Nano 33 BLE
+===================
+
+<img src="https://store-cdn.arduino.cc/usa/catalog/product/cache/1/image/1040x660/604a3538c15e081937dbfbd20aa60aad/a/b/abx00031_featured.jpg" width="35%">
+
+The [Arduino Nano 33 BLE](https://store.arduino.cc/usa/nano-33-ble) and [Arduino
+Nano 33 BLE Sense](https://store.arduino.cc/usa/nano-33-ble-sense) are compact
+boards based on the Nordic nRF52840 SoC. The "Sense" version includes the
+following sensors:
+
+- 9 axis inertial sensor
+- humidity and temperature sensor
+- barometric sensor
+- microphone
+- gesture, proximity, light color and light intensity sensor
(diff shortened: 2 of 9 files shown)


boards/nano33ble/Makefile
```diff
@@ -0,0 +1,16 @@
...
+TARGET=thumbv7em-none-eabi
+PLATFORM=nano33ble
+
+include ../Makefile.common
+
+ifdef PORT
+  FLAGS += --port $(PORT)
+endif
+
+# Upload the kernel over using bossac
```
hudson-ayers, line 13:
```suggestion
# Upload the kernel using bossac
```

boards/nano33ble/src/main.rs
```diff
@@ -0,0 +1,283 @@
...
+pub static mut STACK_MEMORY: [u8; 0x1000] = [0; 0x1000];
+
+/// Supported drivers by the platform
+pub struct Platform {
+    // ble_radio: &'static capsules::ble_advertising_driver::BLE<
+    //     'static,
+    //     nrf52::ble_radio::Radio,
+    //     VirtualMuxAlarm<'static, Rtc<'static>>,
+    // >,
+    // ieee802154_radio: Option<&'static capsules::ieee802154::RadioDriver<'static>>,
```
hudson-ayers, line 69: I realize this is commented out anyway, but no need for it to be an Option when you aren't using the `nrf52dk_base` setup, if you want to exclude it you will just comment it out.

boards/nano33ble/src/main.rs
```diff
@@ -0,0 +1,283 @@
+//! Tock kernel for the Arduino Nano 33 BLE.
+//!
+//! It is based on nRF52840 SoC (Cortex M4 core with a BLE transceiver).
```
hudson-ayers, line 3:
```suggestion
//! It is based on nRF52840 SoC (Cortex M4 core with a BLE + IEEE 802.15.4 transceiver).
```

boards/nano33ble/src/main.rs
```diff
@@ -0,0 +1,284 @@
...
+    kernel::debug::assign_gpios(Some(&nrf52840::gpio::PORT[LED_KERNEL_PIN]), None, None);
+
+    //--------------------------------------------------------------------------
+    // GPIO
+    //--------------------------------------------------------------------------
+
+    let gpio = components::gpio::GpioComponent::new(
+        board_kernel,
+        components::gpio_component_helper!(
+            nrf52840::gpio::GPIOPin,
```
alexandruradovici, line 135: As this is a newly added board, what about defining the pin numbers starting from 2 instead of 0 so that they match the print on the board (D2, D3, ...)?

bradjc: Good point. Updated!

boards/nano33ble/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+First, follow the [Tock Getting Started guide](../../../doc/Getting_Started.md)
+
+You will need the bossac bootloader tool:
+
+```shell
+$ git clone https://github.com/arduino/BOSSA
+$ cd BOSSA
+$ make bossac
+```
```
brghena, line 28: Need to say something along the lines of "You will need to add the `BOSSA/bin/` directory to your PATH."

boards/nano33ble/README.md
```diff
@@ -0,0 +1,103 @@
...
+| LED[2]            | Tri-color LED Blue  |
+
+## UART Debugging
+
+Currently the Nano 33 board file uses the UART peripheral for all UART
+debugging. However, that UART is not connected to the USB header on the board.
+So, all `debug!()` messages or application `printf()` calls will not be
+displayed. If you want to see the debug output, you need to connect to the two
+UART pins on the Nano 33 headers. You should be able to connect a serial <-> USB
+converter, like an FTDI board, to retrieve the UART output.
```
brghena, line 103: I think it's worth not merging this PR until #1902 is merged and it can use the USB virtual console.

brghena: (This is what's slowing me down from further review of this PR right now. I don't have an FTDI setup for my nano33 BLE boards, but will happily test once they have USB.)

boards/nano33ble/src/io.rs
```diff
@@ -0,0 +1,67 @@
...
+                width: uart::Width::Eight,
+            });
+        }
+        for &c in buf {
+            unsafe {
+                uart.send_byte(c);
+            }
+            while !uart.tx_ready() {}
+        }
+    }
```
brghena, line 46: How hard would it be to have the panic dump print out over USB instead (or in addition)?

bradjc: Really hard, maybe impossible? I don't know how to do it without interrupts.

It might be worth thinking about handling panics in apps differently so that the kernel is still running while the panic message is printed.

brghena: I won't block this PR on it, but we're going to have to think of some way to at least be able to debug crashing apps over USB serial if we want this to be a first-class board.

brghena: #1972

PR #1911, opened 2020-06-05
New Platform: Msp432

lebakassemmerl:
### New Platform Support for the MSP-EXP432P401R Eval-Board

This pull-request adds support for MSP432P401R chip from Texas Instruments.
The implementation is currently very, very basic, only the GPIO-module is properly implemented in order to create a panic-handler which blinks an LED and to be able to test a simple user-space application. 

If this PR will be accepted, the next step would be to implement the UART and timer modules for supporting debug-output and delays in userspace.

## Adding a Platform to Tock Repository

- [x] The hardware must be widely available. Generally that means the hardware platform can be purchased online.
- The port of Tock to the platform must include at least:
   - [x] Console support so that debug!() and printf() work.
   - [x] Timer support.
   - [x] GPIO support with interrupt functionality.
- [x] The contributor must be willing to maintain the platform, at least initially, and help test the platform for future releases.


alistair23:
Thanks for the PR!!!

Would you mind rebasing on master? It looks like GitHub is confused and has included other patches in the PR. It makes it hard to review your changes.

Also the Tock porting guide gives some hints for porting: https://github.com/tock/tock/blob/master/doc/Porting.md#adding-a-platform-to-tock-repository
It also specifies that you need a working console (and timer but that might be less important) before the board is merged.

alistair23:
It also looks like the tests are failing. Make sure to run at least `make ci-nosetup` locally. For new boards the CI check can be helpful in finding bugs (it checks all of your register offsets).

lebakassemmerl:
Hi!
Yes, of course. I will rebase it on master. Also I will make sure that the test won't fail.

Concerning the console and timer: I read this guide and somewhere at the beginning it said that after GPIO is working it's a good time to open a pull request. So should I close/delete this PR, fix my tests, get the console working and open a new one?

alistair23:
Great! Thanks.

I think opening a PR makes sense. It lets everyone know what you are working on (so we don't duplicate efforts). It is also a great way to get feedback.

It might make sense to convert it to a draft PR until you have the tests and UART working though.

bradjc:
Very cool!

Can I ask what attracted you to this board/MCU?

lebakassemmerl:
Yeah, sure!

I'm doing my master thesis in porting Tock to a new platform, and my only 'requirement' was to choose a platform which is of course not supported yet and where a cheap evaluation board exists. In our university (FH Joanneum) we work a lot with all forms of TI products, so I had a look at their microcontroller boards. I chose this particular one because openOCD had good support for it, it contains a quite new Cortex-M4-based chip and the board costs only 20$.

lebakassemmerl:
Hi!

So far I managed it to get the UART in TX direction working in order to print panic- and debug-messages. Is it necessary for a first pull into the main-repo to get also the RX-functionality working? 

Also the GPIO implementation doesn't support interrupts at the moment. Is this also necessary to get merged?

alistair23:
You don't need to have RX working, so that's fine. For GPIO I think that's also ok.

.vscode/launch.json
```diff
@@ -15,6 +15,21 @@
...
+            "gdbpath": "arm-none-eabi-gdb",
+            "autorun": [
+                "monitor reset",
+                "monitor speed auto",
+                "b reset_handler"
+            ]
         }
     ]
-}
+}
```
alistair23, line 35: Can you keep the newline here

boards/msp_exp432p401r/Makefile
```diff
@@ -0,0 +1,27 @@
+# Makefile for building the tock kernel for the MSP-EXP432P401 launchpad
+#
+TARGET=thumbv7em-none-eabi
+PLATFORM=msp-exp432p401r
+
+include ../Makefile.common
+
+# APP=../../../libtock-rs/target/thumbv7em-none-eabi/tab/msp432/blink/cortex-m4.tbf
+# KERNEL=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/debug/$(PLATFORM).elf
+# KERNEL_WITH_APP=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/debug/$(PLATFORM)-app.elf
```
alistair23, line 10: Can you remove the commented out code

boards/msp_exp432p401r/Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+.PHONY: flash
+flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
+	$(OPENOCD) $(OPENOCD_OPTIONS) -c "init; reset halt; flash write_image erase $<; verify_image $<; reset; shutdown"
+
+.PHONY: program
+program: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
+	$(error See README.md and update this section accordingly)
+	# arm-none-eabi-objcopy --update-section .apps=$(APP) $(KERNEL) $(KERNEL_WITH_APP)
+	# $(OPENOCD) $(OPENOCD_OPTIONS) -c "init; reset halt; flash write_image erase $(KERNEL_WITH_APP); verify_image $(KERNEL_WITH_APP); reset; shutdown"
```
alistair23, line 27: Same comment here about commented out code

boards/msp_exp432p401r/README.md
```diff
@@ -0,0 +1,62 @@
+MSP432 Evaluation kit MSP432P401R MCU
+========================================================
```
alistair23, line 2: Can you match the length of `===` with the header

boards/msp_exp432p401r/src/io.rs
```diff
@@ -0,0 +1,58 @@
...
+/// Writer is used by kernel::debug to panic message to the serial port.
+pub struct Writer {
+    initialized: bool,
+}
+
+/// Global static for debug writer
+pub static mut WRITER: Writer = Writer { initialized: false };
+
+impl Writer {
+    /// Indicate that USART has already been initialized.
```
alistair23, line 20: UART?

boards/msp_exp432p401r/src/io.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+#[no_mangle]
+#[panic_handler]
+pub unsafe extern "C" fn panic_fmt(info: &PanicInfo) -> ! {
+    const LED1_PIN: PinNr = PinNr::P01_0;
+    let led = &mut led::LedHigh::new(&mut msp432::gpio::PINS[LED1_PIN as usize]);
+    let writer = &mut WRITER;
+
+    //debug::panic_blink_forever(&mut [led]);
```
alistair23, line 48: Remove commented out code.

boards/msp_exp432p401r/src/main.rs
```diff
@@ -0,0 +1,163 @@
+#![no_std]
+#![cfg_attr(not(doc), no_main)]
+#![feature(asm, core_intrinsics)]
+// #![deny(missing_docs)]
```
alistair23, line 4: This should be:

```
#![no_std]
#![deny(missing_docs)]
```

Instead

boards/msp_exp432p401r/src/main.rs
```diff
@@ -0,0 +1,163 @@
...
+
+// RAM to be shared by all application processes.
+#[link_section = ".app_memory"]
+static mut APP_MEMORY: [u8; 32768] = [0; 32768];
+
+// Force the emission of the `.apps` segment in the kernel elf image
+// NOTE: This will cause the kernel to overwrite any existing apps when flashed!
+#[used]
+#[link_section = ".app.hack"]
+static APP_HACK: u8 = 0;
```
alistair23, line 37: Do you mind explaining why you need this? Can apps not be flashed to the board?

boards/msp_exp432p401r/src/main.rs
```diff
@@ -0,0 +1,163 @@
...
+    msp432::pcm::PCM.set_high_power();
+    msp432::flctl::FLCTL.set_waitstates(msp432::flctl::WaitStates::_1);
+    msp432::flctl::FLCTL.set_buffering(true);
+
+    // Setup 48MHz from external oscillator
+    msp432::gpio::PINS[msp432::gpio::PinNr::PJ_2 as usize].enable_primary_function();
+    msp432::gpio::PINS[msp432::gpio::PinNr::PJ_3 as usize].enable_primary_function();
+    msp432::cs::CS.set_clk_48mhz();
+    msp432::cs::CS.set_smclk_12mhz();
+
```
alistair23, line 79: Here would be a good place to setup the debug GPIO with `kernel::debug::assign_gpios`

chips/msp432/src/chip.rs
```diff
@@ -0,0 +1,83 @@
...
+        &self.systick
+    }
+
+    fn userspace_kernel_boundary(&self) -> &cortexm4::syscall::SysCall {
+        &self.userspace_kernel_boundary
+    }
+
+    fn sleep(&self) {
+        unsafe {
+            cortexm4::scb::unset_sleepdeep();
```
alistair23, line 68: Why do you need this?

chips/msp432/src/cs.rs
```diff
@@ -0,0 +1,269 @@
...
+
+pub static mut CS: ClockSystem = ClockSystem::new();
+
+const CS_BASE: StaticRef<CsRegisters> =
+    unsafe { StaticRef::new(0x4001_0400u32 as *const CsRegisters) };
+
+const KEY: u32 = 0x695A;
+
+#[repr(C)]
+struct CsRegisters {
```
alistair23, line 14: There is a `register_structs!` macro you can use for this

chips/msp432/src/cs.rs
```diff
@@ -0,0 +1,269 @@
...
+     // DCO external resistor calibration 1 register
+    CSDCOERCAL1 [
+        // DCO frequency calibration for DCO frequency range (DCORSEL) 5
+        DCO_FCAL_RSEL5 OFFSET(0) NUMBITS(10)
+    ]
+];
+
+#[allow(dead_code)]
+#[repr(u32)]
+enum DcoFrequency {
```
alistair23, line 208: Is this used?

chips/msp432/src/cs.rs
```diff
@@ -0,0 +1,269 @@
...
+pub struct ClockSystem {
+    registers: StaticRef<CsRegisters>,
+}
+
+impl ClockSystem {
+    pub const fn new() -> ClockSystem {
+        ClockSystem { registers: CS_BASE }
+    }
+
+    #[inline]
```
alistair23, line 226: You probably can just let the compiler inline this, I don't think you need it to be explicit.

chips/msp432/src/lib.rs
```diff
@@ -0,0 +1,140 @@
+#![crate_name = "msp432"]
+#![crate_type = "rlib"]
+#![feature(asm, const_fn, in_band_lifetimes)]
```
alistair23, line 3: `in_band_lifetimes` have been removed from Tock, can you avoid re-introducing them?

chips/msp432/src/pcm.rs
```diff
@@ -0,0 +1,113 @@
...
+
+pub static mut PCM: Pcm = Pcm::new();
+
+const PCM_BASE: StaticRef<PcmRegisters> =
+    unsafe { StaticRef::new(0x4001_0000 as *const PcmRegisters) };
+
+const PCMKEY: u32 = 0x695A; // for unlocking PCMCTL0 and PCMCTL1
+
+#[repr(C)]
+struct PcmRegisters {
```
alistair23, line 14: Same comment about `register_structs!`

chips/msp432/src/pcm.rs
```diff
@@ -0,0 +1,113 @@
...
+        // invalid transition setting during a active power mode request
+        AM_INVALID_TR_IFG OFFSET(2) NUMBITS(1),
+        // 'a DC-DC operation cannot be achieved or maintained'
+        DCDC_ERROR_IFG OFFSET(6) NUMBITS(1)
+    ]
+];
+
+#[allow(dead_code)]
+#[repr(u32)]
+enum ActiveMode {
```
alistair23, line 77: Is this used?

chips/msp432/src/wdt.rs
```diff
@@ -0,0 +1,69 @@
...
+
+const WATCHDOG_BASE: StaticRef<WatchdogRegisters> =
+    unsafe { StaticRef::new(0x4000_480Cu32 as *const WatchdogRegisters) };
+
+/// every write access has to set this 'password' in the upper 8 bit of the
+/// register, otherwise the watchdog resets the whole system
+const PASSWORD: u16 = 0x5A;
+
+#[repr(C)]
+struct WatchdogRegisters {
```
alistair23, line 17: Same comment about `register_structs!`

chips/msp432/src/wdt.rs
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+}
+
+impl watchdog::Watchdog for Watchdog {
+    fn start(&self, _period: usize) {
+        // TODO: implement the period for the watchdog
+        self.registers
+            .ctl
+            .modify(WDTCTL::WDTPW.val(PASSWORD) + WDTCTL::WDTHOLD::CLEAR);
```
alistair23, line 55: Maybe add a call to `unimplemented!()` as this won't work

chips/msp432/src/sysctl.rs
```diff
@@ -0,0 +1,201 @@
...
+
+const SYSCTL_BASE: StaticRef<SysCtlRegisters> =
+    unsafe { StaticRef::new(0xE004_3000 as *const SysCtlRegisters) };
+
+#[allow(dead_code)]
+const WKEY: u8 = 0x69; // for writing to REBOOT_CTL register
+#[allow(dead_code)]
+const UNLKEY: u16 = 0x695A; // for unlocking IP protected secure zone
+#[allow(dead_code)]
+const UNLOCKED: u16 = 0xA596; // value if zone is unlocked
```
alistair23, line 16: Why include these if they aren't used?

lebakassemmerl: Of course I can remove them if you want, but these values are currently not used but are necessary for some register-accesses. I thought i define them although they are not used because then it's not necessary to look them up again in the datasheet.

boards/msp_exp432p401r/src/main.rs
lebakassemmerl: I implemented it like the stm32f3discovery board, where the applications are flashed combined with the kernel. For me this seemed to be the easiest method to get the platform working at all. Is there a better/easier method for doing this?

boards/msp_exp432p401r/src/main.rs
lebakassemmerl: Sure, of course! I totally forgot about the GPIOs besides the LED.

chips/msp432/src/chip.rs
lebakassemmerl: Ohh, another copy-paste bug from a different platform^^
I'll remove this line.

chips/msp432/src/lib.rs
lebakassemmerl: Sure, I'll remove it!

chips/msp432/src/wdt.rs
lebakassemmerl: Ok!

.vscode/launch.json
lebakassemmerl: Yeah sure!

boards/msp_exp432p401r/src/main.rs
alistair23: Most other boards allow you to flash the kernel and apps separately.You can just basically run the same flash command but with different addresses.

Something like this: https://github.com/tock/tock/blob/76b00caa8a7bb54778dcbbcdc9a1ce38148ef1fc/boards/nano33ble/README.md#programming-applications

chips/msp432/src/sysctl.rs
alistair23: If you think they should be left in can you add a comment?

Someone will probably have to look at the datahseet to use them anyway so it still seems worth removing them if they are unused.

boards/msp_exp432p401r/src/main.rs
lebakassemmerl: Ah, I see!
Should I change this that I can flash the apps separately with openOCD (I guess this should be possible)? Or should I keep as it is now?

chips/msp432/src/sysctl.rs
lebakassemmerl: Ok, then I'll remove them (:

boards/msp_exp432p401r/src/main.rs
alistair23: Unless there is a reason you can't (the HW won't let you) you should allow flashing them separately. You can add a `make flash-app` target in the `Makefile` to make it easier to do if you want.

lebakassemmerl: Ok, I added a new target in the makefile. Now it looks like the following:
```make
# Makefile for building the tock kernel for the MSP-EXP432P401 launchpad
#
TARGET=thumbv7em-none-eabi
PLATFORM=msp-exp432p401r

include ../Makefile.common

OPENOCD=openocd
OPENOCD_OPTIONS=-f openocd.cfg
APP_START_ADDR=0x20000
APP?=

.PHONY: flash-debug
flash-debug: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
	$(OPENOCD) $(OPENOCD_OPTIONS) -c "init; reset halt; flash write_image erase $<; verify_image $<; reset; shutdown"

.PHONY: flash
flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
	$(OPENOCD) $(OPENOCD_OPTIONS) -c "init; reset halt; flash write_image erase $<; verify_image $<; reset; shutdown"

.PHONY: flash-app
flash-app:
	$(OPENOCD) $(OPENOCD_OPTIONS) -c "init; reset halt; flash write_image erase $(APP) $(APP_START_ADDR) bin; verify_image_checksum $(APP) $(APP_START_ADDR) bin; reset; shutdown"

.PHONY: program
program: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
	$(error See README.md and update this section accordingly)`

Can I keep it like that? Or should I do something in a different way?

**Comment by lebakassemmerl** [comment]
> Looks REALLY good! :)
> 
> I have added some comments, they should all be pretty small. Overall this looks really good. I didn't compare it to the datasheet but everything looks reasonable.
> 
> Do you mind addressesing the comments and then I'll have another look?

I think I have addressed all your comments with the latest commits. 
So, when you find the time, please have another look (:

File: boards/msp_exp432p401r/src/main.rs [hunk 20]
```diff
@@ -0,0 +1,165 @@
...
+/// Static reference to chip for panic dumps.
+static mut CHIP: Option<&'static msp432::chip::Msp432> = None;
+
+/// How should the kernel respond when a process faults.
+const FAULT_RESPONSE: kernel::procs::FaultResponse = kernel::procs::FaultResponse::Panic;
+
+/// RAM to be shared by all application processes.
+#[link_section = ".app_memory"]
+static mut APP_MEMORY: [u8; 32768] = [0; 32768];
```
ppannuto, line 32: Please update the `APP_MEMORY` to the new model in #1962 -- sorry these came somewhat concurrently :/

chips/msp432/src/cs.rs
```diff
@@ -0,0 +1,275 @@
...
+        // turn on the VLO oscillator
+        VLO_EN OFFSET(8) NUMBITS(1),
+        // turn on the REFO oscillator
+        REFO_EN OFFSET(9) NUMBITS(1),
+        // turn on the MODOSC oscillator
+        MODOSC_EN OFFSET(10) NUMBITS(1),
+        // select REFO nominal frequency: 0 = 32.768kHz, 1=128kHz
+        REFOFSEL OFFSET(15) NUMBITS(1)
+    ],
+    // status of the different clock-sources, if they are active or not
```
ppannuto, line 135: Please promote these to doc comments (`//` -> `///`) and add descriptors for the other register types as well so that documentation is generated for them.

chips/msp432/src/cs.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    fn unlock_registers(&self) {
+        self.registers.key.modify(CSKEY::KEY.val(KEY));
+    }
+
+    fn lock_registers(&self) {
+        // Every value except KEY written to the key register will perform the lock
+        self.registers.key.modify(CSKEY::KEY.val(0));
+    }
```
ppannuto, line 240: This is okay, but it would be better to use the [`PeripheralManager`](https://docs.tockos.org/kernel/common/peripherals/index.html) `before/after_peripheral_access` hooks, which allow an RAII-style interface that will automatically lock and unlock correctly.

chips/msp432/src/flctl.rs
```diff
@@ -0,0 +1,849 @@
...
+        /// Length of the time for which LDO Boost Signal is kept active
+        BOOST_ACTIVE OFFSET(0) NUMBITS(8) [],
+        /// Length for which Flash deactivates the LDO Boost signal before processing any ne
+        BOOST_HOLD OFFSET(8) NUMBITS(8) []
+    ]
+];
+
+#[repr(u32)]
+#[derive(Copy, Clone)]
+pub enum WaitStates {
```
ppannuto, line 802: `///` comment here (what are the `WaitStates` used for?)

chips/msp432/src/pcm.rs
```diff
@@ -0,0 +1,188 @@
...
+impl Pcm {
+    const fn new() -> Pcm {
+        Pcm {
+            registers: PCM_BASE,
+        }
+    }
+    // currently not sure about the interface, so just implement a simple
+    // method for activating AM_LDO_VCORE1 to provide enough power for 48MHz
+    pub fn set_high_power(&self) {
+        while self.registers.ctl1.is_set(PCMCTL1::PMR_BUSY) {}
```
ppannuto, line 182: :/, another one for the #1552 wishlist

(nothing for you to do here; just want to flag it on that issue)

chips/msp432/src/lib.rs
```diff
@@ -0,0 +1,140 @@
+#![crate_name = "msp432"]
+#![crate_type = "rlib"]
+#![feature(asm, const_fn)]
+#![no_std]
+#![allow(unused_doc_comments)]
```
ppannuto, line 5: Remove this `allow`, it shouldn't be needed (though it was apparently used elsewhere; #1992...)

chips/msp432/src/uart.rs
```diff
@@ -0,0 +1,278 @@
...
+
+const DEFAULT_CLOCK_FREQ_HZ: u32 = 12_000_000;
+
+struct BaudFraction {
+    frac: f32,
+    reg_val: u8,
+}
+
+#[rustfmt::skip]
+// Table out of the datahseet to correct the baudrate
```
ppannuto, line 20:
```suggestion
// Table out of the datasheet to correct the baudrate
```

chips/msp432/src/uart.rs
```diff
@@ -0,0 +1,278 @@
...
+        self.enable_tx_interrupt();
+        self.transmit();
+    }
+
+    fn transmit(&self) {
+        let idx = self.tx_index.get();
+
+        self.tx_buffer.map(|tx_buf| self.write_byte(tx_buf[idx]));
+        self.tx_index.set(idx + 1);
+    }
```
ppannuto, line 113: It looks like this is (asynchronously, good) transmitting one byte at a time. That's going to create a ton of interrupts. Is there no easy way to pass a length for the buffer to the uart peripheral?

boards/msp_exp432p401r/Makefile
```diff
@@ -0,0 +1,23 @@
...
+#
+TARGET=thumbv7em-none-eabi
+PLATFORM=msp-exp432p401r
+
+include ../Makefile.common
+
+OPENOCD=openocd
+OPENOCD_OPTIONS=-f openocd.cfg
+APP_START_ADDR=0x20000
+APP?=
```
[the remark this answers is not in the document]
lebakassemmerl: It is just for flashing the tbf-file to the correct flash-location via openOCD, and with the `APP` variable you set the path of the tbf-file.

I added already a description and an example in the README. Should I add one in the makefile too?

boards/msp_exp432p401r/src/main.rs
lebakassemmerl: Sure, I will convert it to the new model

chips/msp432/src/cs.rs
lebakassemmerl: Of course, I oversaw these comments.

chips/msp432/src/lib.rs
lebakassemmerl: Ok!

chips/msp432/src/uart.rs
lebakassemmerl: This UART-module unfortunately doesn't have any kind of FIFO, it can only transmit one byte at a time. It could use a DMA-channel for avoiding these interrupts, but I thought for the first implementation I just go the easier way without DMA.

chips/msp432/src/flctl.rs
lebakassemmerl: Ok!

chips/msp432/src/cs.rs
lebakassemmerl: Ok, I will try to convert it to this model!

boards/msp_exp432p401r/Makefile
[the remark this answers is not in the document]
ppannuto: The README is probably fine for now.

chips/msp432/src/uart.rs
ppannuto: I think that's reasonable as a start for basic functionality.

DMA should be on the shortlist of next implementation priorities once this is merged, however.

chips/msp432/src/cs.rs
lebakassemmerl: I have now implemented the `PeripheralManagement` trait for the clock system, I hope I did it correct!

alistair23:
When you address a comment can you mark it as resolved?

lebakassemmerl:
> When you address a comment can you mark it as resolved?

Sure! I didn't know that I was supposed to mark the comments as resolved.

chips/msp432/src/cs.rs
```diff
@@ -0,0 +1,308 @@
...
+
+impl ClockSystem {
+    const fn new() -> ClockSystem {
+        ClockSystem { registers: CS_BASE }
+    }
+
+    // Not sure about the interface, so for testing provide a function to set
+    // the master-clock to 48Mhz
+    pub fn set_mclk_48mhz(&self) {
+        self.before_peripheral_access(self.get_clock(), self.get_registers());
```
ppannuto, line 258: You should not need the explicit calls like this. The idea of the PeripheralManager is to use RAII, that is you create a `RegisterManager` object to get access to the registers. The manager object will automatically call `before` during creation and `after` when it goes out of scope.

chips/msp432/src/cs.rs
```diff
@@ -0,0 +1,308 @@
...
+    ],
+    /// DCO external resistor calibration 1 register
+    CSDCOERCAL1 [
+        /// DCO frequency calibration for DCO frequency range (DCORSEL) 5
+        DCO_FCAL_RSEL5 OFFSET(0) NUMBITS(10)
+    ]
+];
+
+pub struct ClockSystem {
+    registers: StaticRef<CsRegisters>,
```
ppannuto, line 247:
```suggestion
```

With the PeripheralManager interface, this structure should not hold a reference to the MMIO registers. Register access needs to go through the peripheral manager instead.

lebakassemmerl: Ahh ok, now I get it, thank you!
I think I use the `PeripheralMangager` now in the correct way

ppannuto:
I think you can move this from a Draft PR to a full-fledged one at the point :)

chips/msp432/src/uart.rs
```diff
@@ -0,0 +1,278 @@
...
+            }
+            ucbrs = val.reg_val;
+        }
+        regs.mctlw.modify(usci::UCAxMCTLW::UCBRS.val(ucbrs as u16));
+
+        // Enable module
+        regs.ctlw0.modify(usci::UCAxCTLW0::UCSWRST::CLEAR);
+
+        // Enable receive interrupt
+        // self.registers.ie.modify(usci::UCAxIE::UCRXIE::SET);
```
alistair23, line 218: Why isn't this enabled?

lebakassemmerl: Because the receive part of the UART isn't implemented yet. I didn't want to enable an interrupt when the whole handling logic is not implemented. My thought was that this would cause only problems.

alistair23: Ah ok. I think it's best just to remove this then. Otherwise looks good :)

lebakassemmerl: Ok, sure!

lebakassemmerl:
I implemented no a very first version for handling GPIO interrupts. However, there is a problem with this controller, it does NOT support detecting interrupts at the falling edge and at the rising edge simultaneously. Just for testing the implementation with the button-capsule I detect only falling edges where actually both should be detected. 

How should I solve this problem? Just 'ignore' the fact that both should be detected? Maybe `panic!()`? Or maybe any kind of software implementation (However I have no idea how this could be done within Tock)?

hudson-ayers:
> I implemented no a very first version for handling GPIO interrupts. However, there is a problem with this controller, it does NOT support detecting interrupts at the falling edge and at the rising edge simultaneously. Just for testing the implementation with the button-capsule I detect only falling edges where actually both should be detected.
> 
> How should I solve this problem? Just 'ignore' the fact that both should be detected? Maybe `panic!()`? Or maybe any kind of software implementation (However I have no idea how this could be done within Tock)?

Regarding a software implementation, I think it would basically involve re-configuring a pin to check for the other edge type each time an interrupt arrived for that pin. This would just be done in the chip GPIO driver -- basically when a falling edge interrupt arrives, check that the pin is still low, and if so configure that pin to look for a rising edge interrupt, and vice-versa. Unfortunately it would have to take place in the bottom half interrupt handler, which means there is the possibility for large enough delays that interrupts could be missed, but I think this is still a better solution than only handling interrupts in one direction.

lebakassemmerl:
> > I implemented no a very first version for handling GPIO interrupts. However, there is a problem with this controller, it does NOT support detecting interrupts at the falling edge and at the rising edge simultaneously. Just for testing the implementation with the button-capsule I detect only falling edges where actually both should be detected.
> > How should I solve this problem? Just 'ignore' the fact that both should be detected? Maybe `panic!()`? Or maybe any kind of software implementation (However I have no idea how this could be done within Tock)?
> 
> Regarding a software implementation, I think it would basically involve re-configuring a pin to check for the other edge type each time an interrupt arrived for that pin. This would just be done in the chip GPIO driver -- basically when a falling edge interrupt arrives, check that the pin is still low, and if so configure that pin to look for a rising edge interrupt, and vice-versa. Unfortunately it would have to take place in the bottom half interrupt handler, which means there is the possibility for large enough delays that interrupts could be missed, but I think this is still a better solution than only handling interrupts in one direction.

Ok, thanks!
Then I will try to implement a first software based approach for detecting both edges simultaneously.

lebakassemmerl:
I have now implemented a first approach for detecting both interrupts by software. Can somebody please have a look on it if the implementation makes sense?
I tested it within the debugger, the basic functionality seems to work, after a rising edge it detects a falling one and vice versa. However I am not sure if I didn't make any subtle mistakes in this implementation.

hudson-ayers:
> I have now implemented a first approach for detecting both interrupts by software. Can somebody please have a look on it if the implementation makes sense?
> I tested it within the debugger, the basic functionality seems to work, after a rising edge it detects a falling one and vice versa. However I am not sure if I didn't make any subtle mistakes in this implementation.

This seems like a reasonable approach! Have you tried the button example?

lebakassemmerl:
> > I have now implemented a first approach for detecting both interrupts by software. Can somebody please have a look on it if the implementation makes sense?
> > I tested it within the debugger, the basic functionality seems to work, after a rising edge it detects a falling one and vice versa. However I am not sure if I didn't make any subtle mistakes in this implementation.
> 
> This seems like a reasonable approach! Have you tried the button example?

Yes I did and it works. However, it also worked when I configured the controller to detect only the falling or rising edge. I think that's due to the bouncing of the button itself, maybe the button example isn't the best one to test this implementation.

lebakassemmerl:
I have now implemented a first version of an alarm timer. Is it also necessary to implement the `Timer` trait for accepting the PR? Or is the `Alarm` trait sufficient?

bradjc:
As long as the blink app works, and that only needs alarm.

lebakassemmerl:
> As long as the blink app works, and that only needs alarm.

Ok, thanks!
With the latest commits the blink-app works now.

lebakassemmerl:
Thanks for approving my changes!
Am I supposed to do something else in order to get the PR accepted? Or is the current implementation sufficient for now?

alistair23:
I think you have everything you need. You just need to wait for some more reviews

chips/msp432/src/chip.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn mpu(&self) -> &cortexm4::mpu::MPU {
+        &self.mpu
+    }
+
+    fn scheduler_timer(&self) -> &cortexm4::systick::SysTick {
+        &self.scheduler_timer
+    }
+
+    fn watchdog(&self) -> &Self::WatchDog {
+        unsafe { &wdt::WDT }
```
bradjc, line 77: This is concerning, why is it unsafe to use the watchdog here?

lebakassemmerl: Because the rust-compiler won't let me return a reference to a static mutable variable. Because there is only 1 watchdog instances I thought this should not be a big problem although it is marked as unsafe.

hudson-ayers: I think that rather than initializing the watchdog timer as a global, it should just be a field of the MSP432 struct, and be initialized in the constructor for that struct.

Using it this way opens the door for future code in the kernel to take a simultaneous mutable reference to the WDT, violating Rust's aliasing guarantees.

lebakassemmerl: Ahh, I see!
Then I create a field within the Msp432 struct for the watchdog and delete the global instance.

lebakassemmerl:
With the latest commit I added some missing lifetime specifiers in the GPIO module. However, since the struct `PinJ` basically doesn't need any lifetime (because it doesn't support interrupts), the macro I created doesn't work any more. For this reason I added a `PhantomData<'a ()>` member to  `PinJ` which makes it behave as it would have a lifetime. Is this the intended way to go or is there a better one?

chips/msp432/src/chip.rs
hudson-ayers: Is it necessary to stop it immediately at startup? Could that just be done as part of the call to `msp432::new()`? Either way, I think that temporarily allocating a short lived instance on the stack (as you are doing) is also fine.

lebakassemmerl: I think it is necessary to stop it immediately at startup, because the interval for it is only ~10ms and depending on what is done before the Msp432 struct is instantiated it could be that the watchdog resets the system.

PR #1913, opened 2020-06-06
Added i.MX RT 1052 EVKB Board 

valexandru:
### Pull Request Overview

This pull request adds support for the NXP i.MX RT 1052 Crossover MCU and  i.MX RT 1052 EVKB Board.

Supported features so far are:
- UART
- GPIO
- Leds
- I2C
- Timer

This also adds support for the Cortex M7 MCU.

This also adds the fxos8700 component.

### Testing Strategy

This pull request was tested using the i.MX RT 1052 EVKB Board.

### TODO or Help Wanted

This pull request still uses a binary FCB Header as header creation is still a problem. Looking into using https://github.com/imxrt-rs/imxrt-boot-gen . Would this be a problem with the licensing?

The timer frequency is not properly working due to clock setup issues.

Due to the high speed of the MCU, i had to add two kernel parameters:
```rust
/// The time a process is permitted to run before being pre-empted
pub(crate) kernel_tick_duration_us: u32,
/// Skip re-scheduling a process if its quanta is nearly exhausted
pub(crate) min_quanta_threshold_us: u32,
```
```rust
    trace_syscalls: false,
    debug_load_processes: false,
    kernel_tick_duration_us: 1000000,
    min_quanta_threshold_us: 50000,
};
```

This is a temporary solution, any other solution would be great.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.


arch/cortex-m7/src/mpu.rs
```diff
@@ -0,0 +1 @@
+../../cortex-m3/src/mpu.rs
```
lschuermann, line 1: This should probably be copied or otherwise imported using Rust methods, not a symlink. Explains why you applied the M7 changes to the M3 files.

arch/cortex-m3/src/mpu.rs
```diff
@@ -1,5 +1,5 @@
-//! Implementation of the memory protection unit for the Cortex-M3 and
-//! Cortex-M4.
+//! Implementation of the memory protection unit for the
+//! Cortex-M7.
```
lschuermann, line 2: This affects the Cortex M3 MPU.

chips/imxrt1050/src/ccm.rs
```diff
@@ -0,0 +1,711 @@
...
+    // clock gating register 5
+    ccgr5: ReadWrite<u32, CCGR5::Register>,
+    _reserved6: [u8; 8],
+}
+
+register_bitfields![u32,
+    CCR [
+        /// Enable for REG_BYPASS_COUNTER
+        RBC_EN OFFSET(27) NUMBITS(1) [],
+
```
lschuermann, line 40: Nitpick, but probably should be formatted without additional newlines as most register definitions around Tock?

valexandru: I fixed the issues and created a new pull request, since this branch was not up to date with the master. Thank you for your input! I will close this pull request and we can continue on the new one.

PR #1916, opened 2020-06-06
Added STM32F412G Discovery kit

alexandruradovici:
### Pull Request Overview

This pull request adds the STM32F412G Discovery Kit. Features working:
- LEDs
- Joystick
- Virtual COM

This uses the same chip as the nucleo429 so all the low level hardware implemented for it should work with this device.

### Testing Strategy

This pull request was tested using an STM32F412G Discovery.

### Documentation Updated

- [X] Updated the relevant files in `/docs`

### Formatting

- [X] Ran `make prepush`.


boards/discovery_f412g/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+(or)
+
+$ make flash-debug
+```
+
+> **Note:** Unlike other Tock platforms, the default kernel image for this
+> board will clear flashed apps when the kernel is loaded. This is to support
+> the non-tockloader based app flash procedure below. To preserve loaded apps,
+> comment out the `APP_HACK` variable in `src/main.rs`.
```
bradjc, line 23: Why are all of these platforms still doing this? My impression is the naming implied this was short-term, but if it is getting copied over and over, this is going to become less a hack and more of a standard. What would it take to fix this?

bradjc:
My other question, and maybe opening an issue for this would be good, is what is the roadmap/vision for Tock support on STM boards? I imagine there are probably hundreds of them, do we want to support as many as possible? I'm not saying we should or should not, but having some shared understanding of the direction would be helpful.

I'm concerned that having many boards with only OK support and testing is less beneficial than having say two boards with robust support that gets tested for releases. But there is also merit in supporting the hardware that people already have. I would probably be supportive of having a "flagship" STM board that many of us buy and is included in release testing that we concentrate support around. Then, there could be other related boards with support in Tock, with the hope that fixes/improvements for the flagship help all of the related boards. But, I'm not sure how realistic that is, or if there is a clear STM board that would make sense to prioritize.

boards/discovery_f412g/README.md
alexandruradovici: I did not implement the original Nucelo boards, I think tockloader does not support this platform so far.

alexandruradovici:
> Remove ` tools/.format_fresh`
> 
> Also, should probably match the naming for the other discovery board.

I will rename the other board also, to be consistent with the initial naming used by nucleo.

alexandruradovici:
> My other question, and maybe opening an issue for this would be good, is what is the roadmap/vision for Tock support on STM boards? I imagine there are probably hundreds of them, do we want to support as many as possible? I'm not saying we should or should not, but having some shared understanding of the direction would be helpful.
> 
> I'm concerned that having many boards with only OK support and testing is less beneficial than having say two boards with robust support that gets tested for releases. But there is also merit in supporting the hardware that people already have. I would probably be supportive of having a "flagship" STM board that many of us buy and is included in release testing that we concentrate support around. Then, there could be other related boards with support in Tock, with the hope that fixes/improvements for the flagship help all of the related boards. But, I'm not sure how realistic that is, or if there is a clear STM board that would make sense to prioritize.

This discovery board has more or less the same MCU as the supported nucelo (stmf4xx). I intend to fully support this board as I need it for my teaching. It will take some time to reach that point.

hudson-ayers:
> My other question, and maybe opening an issue for this would be good, is what is the roadmap/vision for Tock support on STM boards? I imagine there are probably hundreds of them, do we want to support as many as possible? I'm not saying we should or should not, but having some shared understanding of the direction would be helpful.
> 
> I'm concerned that having many boards with only OK support and testing is less beneficial than having say two boards with robust support that gets tested for releases. But there is also merit in supporting the hardware that people already have. I would probably be supportive of having a "flagship" STM board that many of us buy and is included in release testing that we concentrate support around. Then, there could be other related boards with support in Tock, with the hope that fixes/improvements for the flagship help all of the related boards. But, I'm not sure how realistic that is, or if there is a clear STM board that would make sense to prioritize.

I think the stm32f3discovery board might make sense as a "flagship" STM board if we were to go that route. It has 10 LEDs, buttons, accel/gyro/compass, and costs $15. Most importantly, it is the board required to complete the [embedded rust book](https://rust-embedded.github.io/book/), so lots of people interested in getting started with embedded Rust might already have one. The main downside is it doesn't have wireless.

alexandruradovici:
> > Remove ` tools/.format_fresh`
> > Also, should probably match the naming for the other discovery board.
> 
> I will rename the other discovery board also, to be consistent with the initial naming used by nucleo.

I rename this board as STM uses this naming in their website.

chips/stm32f412g/README.md
```diff
@@ -0,0 +1,5 @@
+# ST Micro stm32f412g MCU
+
+Builds upon the `stm32f4xx` crate and includes hardware setup that is specific to this version of
+the stm32f4xx series. Boards that include an stm32f429zi MCU should use this crate as a dependency
```
ppannuto, line 4: > Boards that include an **stm32f429zi MCU**

Quick sanity check: I'm afraid the answer is going to be that STM just has a confusing naming scheme, but the name / title of this file is `stm32f412g`. Is this all correct as-written?

boards/stm32f412gdiscovery/src/main.rs
```diff
@@ -0,0 +1,463 @@
...
+        /// End of the ROM region containing app images.
+        ///
+        /// This symbol is defined in the linker script.
+        static _eapps: u8;
+    }
+
+    kernel::procs::load_processes(
+        board_kernel,
+        chip,
+        core::slice::from_raw_parts(
```
ppannuto, line 443: Attn #1962

boards/stm32f412gdiscovery/Makefile
```diff
@@ -0,0 +1,21 @@
+# Makefile for building the tock kernel for the NUCLEO-429ZI platform
+#
```
hudson-ayers, line 2:
```suggestion
# Makefile for building the tock kernel for the stm32412gdiscovery platform
#
```

boards/stm32f412gdiscovery/chip_layout.ld
```diff
@@ -0,0 +1,14 @@
+/* Memory layout for the STM32F446RE
```
hudson-ayers, line 1:
```suggestion
/* Memory layout for the STM32F412G
```

boards/stm32f412gdiscovery/src/io.rs
```diff
@@ -0,0 +1,79 @@
...
+pub struct Writer {
+    initialized: bool,
+}
+
+/// Global static for debug writer
+pub static mut WRITER: Writer = Writer { initialized: false };
+
+impl Writer {
+    /// Indicate that USART has already been initialized. Trying to double
+    /// initialize USART2 causes STM32F446RE to go into in in-deterministic state.
```
hudson-ayers, line 28:
```suggestion
    /// initialize USART2 causes STM32F412G to go into in in-deterministic state.
```

chips/stm32f412g/README.md
alexandruradovici: It is based on the same stm32f4 family, but the stm32f429zi line is wrong. Fixing it.

boards/stm32f412gdiscovery/src/main.rs
alexandruradovici: I know about it, but is is not merged yet. I'll update the stm boards as soon as this gets merged. Should I do it differently?

ppannuto: Just flagging the (potential) conflict here; whichever goes in first, the other will have to update

alexandruradovici:
bors seems to be stuck here ...

ppannuto:
bors retry

Hmmm..

PR #1917, opened 2020-06-07
hil: ble_advertising: Pass the transmit buffer on callback

alistair23:
### Pull Request Overview

Pass the BLE transmit buffer back to the capsule.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


hudson-ayers:
I tested this on the nrf52840dk with the libtock-c ble_advertising example and it works. 

One downside of this is that if a transmission fails for some reason the buffer is lost forever, whereas the previous design could just call `transmit_advertisement()` again...but I agree that this is more in line with our other drivers

PR #1919, opened 2020-06-08
Core working group notes from June 05

brghena:
### Pull Request Overview

Adds core working group notes from June 5th.

### Testing Strategy

N/A

### TODO or Help Wanted

None.

### Documentation Updated

### Formatting

- [N/A] Ran `make prepush`.


doc/wg/core/notes/core-notes-2020-06-05.md
```diff
@@ -0,0 +1,96 @@
...
...portant for libtock-rs as its using lambda callback functions. And local buffers for slices. If we can't get back ownership, the lifetime is likely gone. The app might exit the function and the memory may not exist anymore. The kernel could end up corrupting the stack. So it's important that unallow exists. But the Tock kernel doesn't actually enforce that capsules do the right thing and give back ownership. I'll focus on slices but the problem is similar for both. We have to declare clear semantics about who has ownership of a slice. For example, a USB driver that wants to receive packets into userspace, usually you would allow a slice from userspace and subscribe to a callback then call a command to start it. Then the kernel would fill the slice and call the callback. After that userspace can interact with the data. It's important to define who owns the slice at what point in these operations. Because the kernel can't overwrite the slice with the next packet while the userland is interacting with it. So the solution in userland seems to be to allow and disallow slices in a double-buffer sense. But there is nothing enforcing that a capsule properly gives back the memory right now.
```
gendx, line 79: Should be "Credit to Woyten and Johnathan who raised the original issue (https://github.com/tock/libtock-rs/issues/143)."

brghena:
```suggestion
 * Guillaume: Credit to Woyten and Johnathan who raised the issue (https://github.com/tock/libtock-rs/issues/143). The subscribe/allow syscalls that give callbacks or slices and can later unallow or unsubscribe with a null pointer. The assumption in libtock-rs is that once you unsubscribe you have ownership again. I think this is going to be important for libtock-rs as its using lambda callback functions. And local buffers for slices. If we can't get back ownership, the lifetime is likely gone. The app might exit the function and the memory may not exist anymore. The kernel could end up corrupting the stack. So it's important that unallow exists. But the Tock kernel doesn't actually enforce that capsules do the right thing and give back ownership. I'll focus on slices but the problem is similar for both. We have to declare clear semantics about who has ownership of a slice. For example, a USB driver that wants to receive packets into userspace, usually you would allow a slice from userspace and subscribe to a callback then call a command to start it. Then the kernel would fill the slice and call the callback. After that userspace can interact with the data. It's important to define who owns the slice at what point in these operations. Because the kernel can't overwrite the slice with the next packet while the userland is interacting with it. So the solution in userland seems to be to allow and disallow slices in a double-buffer sense. But there is nothing enforcing that a capsule properly gives back the memory right now.
```

brghena: Thank you! I'm glad that didn't slip through.

PR #1766, opened 2020-04-17
Unify arithmetic implementation of field registers and add unit tests.

gendx:
### Pull Request Overview

While reviewing https://github.com/tock/tock/pull/1661#discussion_r388306189, it came up that there is some amount of code duplication in the arithmetic operations on various register types.

This pull request unifies the implementation, by adding new functions to the `Field` and `FieldValue` types, and making the other types use these. Unit tests are also added for these unified functions.

Some formulas are also simplified:
- `(value << shift) & (mask << shift)` => `(value & mask) << shift`
- `(val & (field.mask << field.shift)) >> field.shift != T::zero()` => `val & (field.mask << field.shift) != T::zero()`

### Testing Strategy

This pull request was tested by Travis-CI.

### TODO or Help Wanted

- `matches_any(&self, field: FieldValue<T, R>) -> bool` doesn't makes a lot of sense to me, because it doesn't use any "value" of the field. It could be `matches_any(&self, field: Field<T, R>) -> bool` instead. But then it's the same as the existing `is_set(&self, field: Field<T, R>) -> bool`. Should `matches_any` be removed? Tock doesn't use a lot of `matches_any`.
- `ReadWrite` and `InMemoryRegister` have the same implementation. Is that intended? I wonder why volatile read/write are needed for `InMemoryRegister` - given that such "register" is not MMIO. Tock doesn't use a lot of `InMemoryRegister`.
- This pull request could be extended by defining new traits to further remove code duplication. However, this would mean more refactoring to expose these traits, and a breaking change for the published crate (https://crates.io/crates/tock-registers).
  Here is a draft of that:

```rust
trait ReadAble<T: IntLike, R: RegisterLongName> {
    /// Get the raw register value
    fn get(&self) -> T;

    /// Read the value of the given field
    fn read(&self, field: Field<T, R>) -> T {
        field.read(self.get())
    }

    /// Read value of the given field as an enum member
    fn read_as_enum<E: TryFromValue<T, EnumType = E>>(&self, field: Field<T, R>) -> Option<E> {
        field.read_as_enum(self.get())
    }

    ...
}

trait WriteAble<T: IntLike, R: RegisterLongName> {
    /// Set the raw register value
    fn set(&self, value: T);

    /// Write the value of one or more fields, overwriting the other fields with zero
    fn write(&self, field: FieldValue<T, R>) {
        self.set(field.value);
    }
}

[...]

PR #1926, opened 2020-06-10
Prepare for 64 PMP config registers

alistair23:
### Pull Request Overview

This PR was supposed to add support for 64 PMP config register on RISC-V now that it is supported by the spec.

Instead this patch has turned more info converting the riscv-csr library to use the new inline assembly macro and to allow dynamic array access to the RISC-V registers.

This PR does a few things:
 - Converts the PMP CSRs to an array (to allow simplifying the PMP code)
 - Uses the new inline assembly. Unfortunately Rust won't let use access CSR dynamically like we used to (see https://github.com/rust-lang/rust/issues/73220) so we now have a giant if statement to see which one to access. I couldn't see any other way to make the value const.
 - We also implement the register functions on specific types (u32) instead of generics. This is because with generics we can't dynamically access arrays of CSRs with the inline assembly.
 - Finally we clean up some of the PMP code.

### Testing Strategy

None yet.

### TODO or Help Wanted

Comments please, this is uglier then I would like it to be.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


arch/rv32i/src/pmp.rs
```diff
@@ -531,115 +408,115 @@ impl kernel::mpu::MPU for PMPConfig {
...
                                         + csr::pmpconfig::pmpcfg::x0::CLEAR
                                         + csr::pmpconfig::pmpcfg::a0::TOR,
                                 );
-                                csr::CSR.pmpaddr4.set((start as u32) >> 2);
+                                csr::CSR.pmpaddr[4].set((start as u32) >> 2);
 
                                 // Set access to end address
-                                csr::CSR.pmpcfg1.set(cfg_val << 8 | csr::CSR.pmpcfg0.get());
-                                csr::CSR.pmpaddr5.set((start as u32 + size as u32) >> 2);
+                                csr::CSR.pmpcfg[1].set(cfg_val << 8 | csr::CSR.pmpcfg[0].get());
```
a-pronin, line 448: should be pmpcfg[1]
please see all places where changes are needed in https://github.com/tock/tock/pull/1932

libraries/riscv-csr/src/csr.rs
```diff
@@ -33,73 +69,272 @@ impl<T: IntLike, R: RegisterLongName> ReadWriteRiscvCsr<T, R> {
...
+    pub fn matches_any(&self, field: FieldValue<u32, R>) -> bool {
+        self.get() & field.mask != u32::zero()
     }
 
     #[inline]
-    pub fn matches_all(&self, field: FieldValue<T, R>) -> bool {
+    pub fn matches_all(&self, field: FieldValue<u32, R>) -> bool {
         self.get() & field.mask == field.value
     }
 }
```
gendx, line 340: All of these functions look very similar to what's already available in `libraries/tock-register-interface/src/registers.rs`. Is there any reason not to use that?

arch/rv32i/src/csr/mod.rs
```diff
@@ -23,26 +23,8 @@ pub struct CSR {
...
-    pub pmpaddr11: ReadWriteRiscvCsr<u32, pmpaddr::pmpaddr::Register>,
-    pub pmpaddr12: ReadWriteRiscvCsr<u32, pmpaddr::pmpaddr::Register>,
-    pub pmpaddr13: ReadWriteRiscvCsr<u32, pmpaddr::pmpaddr::Register>,
-    pub pmpaddr14: ReadWriteRiscvCsr<u32, pmpaddr::pmpaddr::Register>,
-    pub pmpaddr15: ReadWriteRiscvCsr<u32, pmpaddr::pmpaddr::Register>,
+    pub pmpcfg: [ReadWriteRiscvCsr<u32, pmpconfig::pmpcfg::Register>; 4],
+    pub pmpaddr: [ReadWriteRiscvCsr<u32, pmpaddr::pmpaddr::Register>; 16],
     pub mie: ReadWriteRiscvCsr<u32, mie::mie::Register>,
     pub mscratch: ReadWriteRiscvCsr<u32, mscratch::mscratch::Register>,
     pub mepc: ReadWriteRiscvCsr<u32, mepc::mepc::Register>,
```
gendx, line 30: It looks like this definition could use the `register_structs` macro defined in `libraries/tock-register-interface/src/macros.rs`. Is there any reason not to use it?

libraries/riscv-csr/src/csr.rs
bradjc: Reading a CSR requires an assembly instruction, not accessing a memory location.

arch/rv32i/src/csr/mod.rs
bradjc: Alistair is just updating what is already there, I think it is fine to not add additional changes.

gendx: Note: my question was beyond the scope of this PR, to see if there's any improvement that could be made in `libraries/tock-register-interface` to improve adoption of this library, not only in `chips/` but in `arch/` as well.

libraries/riscv-csr/src/csr.rs
gendx: A lot of the logic still overlaps with methods such as `FieldValue::matches_all`. The only difference that I see is in the implementation of the `set` and `get` methods - all of the other methods are derived from it in the same manner as in the `libraries/tock-register-interface` library.

It looks like the `ReadAble` and `WriteAble` traits that I sketched in https://github.com/tock/tock/pull/1766#issue-405159759 could be beneficial here as well.

ppannuto:
bors d=alistair23,gendx

PR #1928, opened 2020-06-12
kernel: process: try to use memory fixed address

bradjc:
### Pull Request Overview

This pull request adds logic to process.rs to make it try to find a matching memory address if a specific address was requested by the app. Right now all it tries to do is push the app back in memory (i.e. start the app's memory region at a later address).

This also changes the return values from `create()`. This fixes the trace process loading debugging messages so that the correct memory region is printed if an app does not start at the beginning of the memory region.

### Testing Strategy

Running multiple apps on arty-e21.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -1688,6 +1688,51 @@ impl<C: 'static + Chip> Process<'_, C> {
         // Minimum memory size for the process.
         let min_total_memory_size = min_app_ram_size + initial_kernel_memory_size;
 
+        // Check if this process requires a fixed memory start address. If so,
+        // try to adjust the memory region to work for this process.
+        //
+        // Right now, we only support skipping some RAM and leaving a chunk
+        // unused so that the memory region starts where the process needs it
+        // to.
+        let remaining_memory = if let Some(fixed_memory_start) = tbf_header.get_fixed_address_ram()
+        {
+            // The process does have a fixed address.
+            if fixed_memory_start == remaining_memory.as_ptr() as u32 {
+                // Address already matches.
+                remaining_memory
+            } else if fixed_memory_start > remaining_memory.as_ptr() as u32 {
+                // Process wants a memory address farther in memory. Try to
+                // advance the memory region to make the address match.
+                let diff = (fixed_memory_start - remaining_memory.as_ptr() as u32) as usize;
+                if diff > remaining_memory.len() {
+                    // We ran out of memory.
+                    let actual_address =
+                        remaining_memory.as_ptr() as u32 + remaining_memory.len() as u32 - 1;
+                    let expected_address = fixed_memory_start;
+                    return Err(ProcessLoadError::MemoryAddressMismatch {
+                        actual_address,
+                        expected_address,
+                    });
+                } else {
+                    // Change the memory range to start where the process
+                    // requested it.
+                    remaining_memory
+                        .get_mut(diff..)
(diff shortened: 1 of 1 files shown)


kernel/src/process.rs
```diff
@@ -1619,10 +1627,45 @@ impl<C: 'static + Chip> Process<'_, C> {
...
+            if let Some(fixed_memory_start) = tbf_header.get_fixed_address_ram() {
+                // The process does have a fixed address. See if it comes after
+                // where we are in memory.
+                if fixed_memory_start > remaining_app_memory as u32 {
+                    let diff: usize = (fixed_memory_start - remaining_app_memory as u32) as usize;
+                    if diff > remaining_app_memory_size {
+                        // We ran out of memory. Return the original range since (in
+                        // this step) we were unable to meet the process's
+                        // requirements.
+                        (remaining_app_memory, remaining_app_memory_size)
```
ppannuto, line 1646: This is confusing. Why would this not return an error here?

kernel/src/process.rs
```diff
@@ -1619,10 +1627,45 @@ impl<C: 'static + Chip> Process<'_, C> {
...
+                        // from the length of the remaining memory.
+                        (
+                            fixed_memory_start as *mut u8,
+                            remaining_app_memory_size - diff,
+                        )
+                    }
+                } else {
+                    // Address already matches, or is earlier in memory, nothing we
+                    // have to do (or can do).
+                    (remaining_app_memory, remaining_app_memory_size)
```
ppannuto, line 1659: Same question, why not return an error here?

kernel/src/process.rs
```diff
@@ -1619,10 +1627,45 @@ impl<C: 'static + Chip> Process<'_, C> {
         // Minimum memory size for the process.
         let min_total_memory_size = min_app_ram_size + initial_kernel_memory_size;
 
+        // Check if this process requires a fixed memory start address. If so,
+        // try to adjust the memory region to work for this process.
+        //
+        // Right now, we only support skipping some RAM and leaving a chunk
+        // unused so that the memory region starts where the process needs it
+        // to.
+        let (new_memory_start, new_memory_size) =
```
ppannuto, line 1636:
```suggestion
        let (requested_memory_start, requested_memory_size) =
```

"new" is confusing.., whether fixed or not, I think maybe "requested" better expresses that here it's what the loader is requesting.

Now that there are variations of "memory_start", the variables below should probably also change to be something more specific, maybe:
```diff
-let (memory_start, memory_size) = match chip.mpu().allocate_app_memory_region(
+let (actual_memory_start, actual_memory_size) = match chip.mpu().allocate_app_memory_region(
```

kernel/src/process.rs
bradjc: My thought was that this step is very narrowly defined in the scope of what it can do. However, perhaps the MPU step could figure something better out that would support the application.

bradjc:
I'll mark this blocked for now, since it really requires the updates to libtock-c to be usable, and there are other, more pressing, issues to work on in process.rs at the moment.

bradjc:
I have updated this for after #2022 is merged.

bradjc:
Ok! This is ready to go now.

bradjc:
@ppannuto This is the last piece (I think) for basic fixed address support to be reasonably usable.

PR #1930, opened 2020-06-12
kernel: process: omit debug addrs for fixed apps

bradjc:
If an app is compiled for a fixed address, then we do not need to custom make an LST file based on where it actually executed. The `make lst` command will work just fine. This updates the message that is printed on a panic for apps which are compiled with fixed addresses.

### Testing Strategy

This pull request was tested by verifying the correct message is printed on hail.

### TODO or Help Wanted

n/a

### Documentation Updated

Our example panic printout is still valid since the PIC app version did not change.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -1488,16 +1488,65 @@ impl<C: Chip> ProcessType for Process<'_, C> {
             let _ = writer.write_fmt(format_args!("{}", config));
         });
 
-        let sram_start = self.memory.as_ptr() as usize;
-        let flash_start = self.flash.as_ptr() as usize;
-        let flash_init_fn = flash_start + self.header.get_init_function_offset() as usize;
+        /// Return Ok(Flash Address, RAM address) if this app has
+        /// fixed_addresses, Err(()) otherwise.
+        fn check_for_fixed_address(flash: &'static [u8]) -> Result<(u32, u32), ()> {
+            // Need to re-parse the TBF header to find if this app has fixed
+            // addresses or not.
+            let test_header_slice = flash.get(0..8).ok_or(())?;
+            let (version, header_length, _) =
+                tbfheader::parse_tbf_header_lengths(test_header_slice.try_into().or(Err(()))?)
+                    .or(Err(()))?;
+            let header_flash = flash.get(0..header_length as usize).ok_or(())?;
 
-        let _ = writer.write_fmt(format_args!(
-            "\
-             \r\nTo debug, run `make debug RAM_START={:#x} FLASH_INIT={:#x}`\
-             \r\nin the app's folder and open the .lst file.\r\n\r\n",
-            sram_start, flash_init_fn
-        ));
+            // Parse the full TBF header to check if there is a fixed addresses
+            // header.
+            let tbf_header = tbfheader::parse_tbf_header(header_flash, version).or(Err(()))?;
+
+            // If either address is specified, we say that there is a fixed
+            // address.
+            if tbf_header.get_fixed_address_flash().is_some()
+                || tbf_header.get_fixed_address_ram().is_some()
+            {
+                Ok((
+                    tbf_header.get_fixed_address_flash().unwrap_or(0),
+                    tbf_header.get_fixed_address_ram().unwrap_or(0),
+                ))
+            } else {
+                Err(())
+            }
+        }
+
+        // Print a helpful message on how to re-compile a process to view the
+        // listing file. If a process is PIC, then we also need to print the
+        // actual addresses the process executed at so that the .lst file can be
+        // generated for those addresses. If the process was already compiled
+        // for a fixed address, then just generating a .lst file is fine.
+
+        match check_for_fixed_address(self.flash) {
+            Ok((flash, ram)) => {
+                // Fixed addresses, can just run `make lst`.
+                let _ = writer.write_fmt(format_args!(
+                    "\
+                     \r\nTo debug, run `make lst` in the app's folder\
+                     \r\nand open the arch.{:#x}.{:#x}.lst file.\r\n\r\n",
+                    flash, ram
+                ));
+            }
+            Err(()) => {
+                // PIC, need to specify the addresses.
+                let sram_start = self.memory.as_ptr() as usize;
+                let flash_start = self.flash.as_ptr() as usize;
+                let flash_init_fn = flash_start + self.header.get_init_function_offset() as usize;
+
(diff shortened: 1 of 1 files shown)


kernel/src/process.rs
```diff
@@ -1473,16 +1473,65 @@ impl<C: Chip> ProcessType for Process<'_, C> {
             let _ = writer.write_fmt(format_args!("{}", config));
         });
 
-        let sram_start = self.memory.as_ptr() as usize;
-        let flash_start = self.flash.as_ptr() as usize;
-        let flash_init_fn = flash_start + self.header.get_init_function_offset() as usize;
+        /// Return Ok(Flash Address, RAM address) if this app has
+        /// fixed_addresses, Err(()) otherwise.
+        fn check_for_fixed_address(flash: &'static [u8]) -> Result<(u32, u32), ()> {
```
brghena, line 1478: Why make this a nested function if you're only going to call it once?

kernel/src/process.rs
```diff
@@ -1473,16 +1473,65 @@ impl<C: Chip> ProcessType for Process<'_, C> {
...
+                // PIC, need to specify the addresses.
+                let sram_start = self.memory.as_ptr() as usize;
+                let flash_start = self.flash.as_ptr() as usize;
+                let flash_init_fn = flash_start + self.header.get_init_function_offset() as usize;
+
+                let _ = writer.write_fmt(format_args!(
+                    "\
+                     \r\nTo debug, run `make debug RAM_START={:#x} FLASH_INIT={:#x}`\
+                     \r\nin the app's folder and open the .lst file.\r\n\r\n",
+                    sram_start, flash_init_fn
```
brghena, line 1531: Will `make debug` generate a `arch.{}.{}.lst` file too?

kernel/src/process.rs
bradjc: So I can use the `?` operator. I'm not sure how else to do it.

kernel/src/process.rs
bradjc: Yes. But it will be at a useless address.

brghena: I guess I meant instead of saying "open the .lst file" should we be saying "open the arch.{}.{}.lst file" but I guess in this case there should only be one LST file, right?

bradjc: If we knew the architecture we could print `open the arch.userland_debug.lst` file.

kernel/src/process.rs
```diff
@@ -1473,16 +1473,65 @@ impl<C: Chip> ProcessType for Process<'_, C> {
             let _ = writer.write_fmt(format_args!("{}", config));
         });
 
-        let sram_start = self.memory.as_ptr() as usize;
-        let flash_start = self.flash.as_ptr() as usize;
-        let flash_init_fn = flash_start + self.header.get_init_function_offset() as usize;
+        /// Return Ok(Flash Address, RAM address) if this app has
+        /// fixed_addresses, Err(()) otherwise.
+        fn check_for_fixed_address(flash: &'static [u8]) -> Result<(u32, u32), ()> {
+            // Need to re-parse the TBF header to find if this app has fixed
```
ppannuto, line 1479: This feels like a lot of work and non-trivial duplicated logic. Wouldn't it be easier to just store whether the app is PIC in the `struct ProcessDebug` ?

bradjc: What would happen if #1830 (or something similar) were merged?

ppannuto: Oy.. preaching to the choir, but oh what a nightmare `cfg`s are.

By the same token though, if we're in the space-optimized side of the `cfg`, this whole detail-laden panic handler is gone anyway, right?

bradjc:
Ok changed it to save the fixed addresses rather than re-parse them. I don't really know which is better.

kernel/src/process.rs
```diff
@@ -1488,16 +1498,36 @@ impl<C: Chip> ProcessType for Process<'_, C> {
...
+                // PIC, need to specify the addresses.
+                let sram_start = self.memory.as_ptr() as usize;
+                let flash_start = self.flash.as_ptr() as usize;
+                let flash_init_fn = flash_start + self.header.get_init_function_offset() as usize;
+
+                let _ = writer.write_fmt(format_args!(
+                    "\
+                     \r\nTo debug, run `make debug RAM_START={:#x} FLASH_INIT={:#x}`\
+                     \r\nin the app's folder and open the .lst file.\r\n\r\n",
+                    sram_start, flash_init_fn
```
alistair23, line 1527: Won't this have an effect on code size?

bradjc: I guess so? Is that bad?

alistair23: There are a lot of strings here, I'm worried this will be a big increase.

Do you have a diff of the size?

bradjc:
```
hail: flash use +320 bytes (+0.30%), RAM use +0 bytes(+0.00%)
diff --git a/previous-benchmark-hail b/current-benchmark-hail
index 31658c6..52f9757 100644
--- a/previous-benchmark-hail
+++ b/current-benchmark-hail
@@ -1,6 +1,6 @@
 Tock memory usage report for ./target/thumbv7em-none-eabi/release/hail.elf
-Kernel occupies 106716 bytes of flash
-  103520	code and constant strings
+Kernel occupies 107036 bytes of flash
+  103840	code and constant strings
     3196	variable initializers
 Kernel occupies 16380 bytes of RAM
```

```
nrf52840dk: flash use +240 bytes (+0.19%), RAM use +0 bytes(+0.00%)
diff --git a/previous-benchmark-nrf52840dk b/current-benchmark-nrf52840dk
index c2cc239..ef5eb1a 100644
--- a/previous-benchmark-nrf52840dk
+++ b/current-benchmark-nrf52840dk
@@ -1,6 +1,6 @@
 Tock memory usage report for ./target/thumbv7em-none-eabi/release/nrf52840dk.elf
-Kernel occupies 123668 bytes of flash
-  121888	code and constant strings
+Kernel occupies 123908 bytes of flash
+  122128	code and constant strings
     1780	variable initializers
 Kernel occupies 16384 bytes of RAM
```

From: https://github.com/tock/tock/pull/1930/checks?check_run_id=850230327

alistair23: Ah, I forgot we run size diffs on PRs.

Ok, that seems pretty small so that's fine.

PR #1931, opened 2020-06-13
virtual_uart: fix sync transmitted_buffer case

a-pronin:
Before this change, a call to MuxUart::do_next_op set self.inflight to
the node that had pending operation *after* calling the handler for the
operation (transmitted_buffer or transmitted_word). As a result, if the
operation was performed synchronously, MuxUart::transmitted_buffer was
called while self.inflight was still None. Which led to not calling
transmitted_buffer for the originating device and not returning the
buffer to it. So, after a single debug!, all subsequent debug! calls
didn't have an output_buffer and kept adding to the ring_buffer w/o
printing.

Here we fix it by setting self.inflight for MuxUart before performing
the operatiion. Note that we can only set it if we know tx_buffer exists
(otherwise can end up for a non-empty inflight with no operation
pending), so need to set inside the `node.tx_buffer.take().map` block.

To avoid nested calls to do_next_op if buffers were pending transmission
at this time (do_next_op -> uart.transmit_buffer -> transmitted_buffer
-> do_next_op -> uart.transmit_buffer -> ...etc...), we replace
do_next_op with do_next_op_async in transmitted_buffer. To avoid losing
scheduled async operations, clearing node.operation had also be made
before calling transmit_buffer (which can call transmited_buffer and
post a new operation) rather than after it, thus node.operation.map()
was changed to node.operation.take().map().

### DIFF
--- a/capsules/src/virtual_uart.rs
+++ b/capsules/src/virtual_uart.rs
@@ -70,7 +70,7 @@ impl<'a> uart::TransmitClient for MuxUart<'a> {
             self.inflight.clear();
             device.transmitted_buffer(tx_buffer, tx_len, rcode);
         });
-        self.do_next_op();
+        self.do_next_op_async();
     }
 }
 
@@ -220,20 +220,23 @@ impl<'a> MuxUart<'a> {
             let mnode = self.devices.iter().find(|node| node.operation.is_some());
             mnode.map(|node| {
                 node.tx_buffer.take().map(|buf| {
-                    node.operation.map(move |op| match op {
+                    self.inflight.set(node);
+                    node.operation.take().map(move |op| match op {
                         Operation::Transmit { len } => {
-                            let (rcode, rbuf) = self.uart.transmit_buffer(buf, *len);
+                            let (rcode, rbuf) = self.uart.transmit_buffer(buf, len);
                             if rcode != ReturnCode::SUCCESS {
                                 node.tx_client.map(|client| {
+                                    self.inflight.clear();
                                     node.transmitting.set(false);
                                     client.transmitted_buffer(rbuf.unwrap(), 0, rcode);
                                 });
                             }
                         }
                         Operation::TransmitWord { word } => {
-                            let rcode = self.uart.transmit_word(*word);
+                            let rcode = self.uart.transmit_word(word);
                             if rcode != ReturnCode::SUCCESS {
                                 node.tx_client.map(|client| {
+                                    self.inflight.clear();
                                     node.transmitting.set(false);
                                     client.transmitted_word(rcode);
                                 });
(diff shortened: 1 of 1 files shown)


phil-levis:
There shouldn't ever be a synchronous case: issuing the callback inside the call isn't allowed.

phil-levis:
Here's the discussion from tock-dev:

https://groups.google.com/forum/#!searchin/tock-dev/callback%7Csort:date/tock-dev/FBlMxpVVGg0/jbTSvho0BAAJ

That being said, you are not the first person to be surprised by this. This is a general principle we've seen that's important for implementing APIs that are simple, but it is not clearly written down anywhere. @alevy -- after the discussion with Oxide did we write this down somewhere?

a-pronin:
> There shouldn't ever be a synchronous case

Good to know, thanks! 
Would it make sense to introduce an alt way to signal synchronous completion - e.g. some pre-defined return code? Or we don't want to encourage non-optimal driver implementations and at the same time make the ownership story less clear (with return codes)? And thus want to always rely on having transmitted_buffer() called asynchronously from service_pending_interrupts()?

> not the first person to be surprised by this

I suspect the surprise often comes from inability to use kernel/debug (primarily, DebugWriter through debug!) before the main loop starts - it uses MuxUart and is thus stuck waiting for transmitted_buffer after the 1st buffer. 

Having spent a bit of time with the OS now, I see that you can go with Uart::transmit_sync, but iiuc (or I just don't see it - quite possible :)) there's no universal interface for defining an "early debug console". Given that during bringup for new boards you often need to print stuff early and may have a simplistic synchronous uart driver (e.g. based on Uart::transmit_sync), and you often die before starting the main loop, there's an impression that something is wrong with console/virtual_uart drivers. :) 

Does it make sense to introduce some early_debug interface (e.g. that uses the same synchronous writer that panic uses that every board implements anyway) in kernel/debug? 

> did we write this down somewhere?

yes, that was my next question. I had this feeling coming up with this change, but wasn't sure :) This patch still made sense for us locally as we are in bringup stage and have a temporary UART driver that does synchronous completion.

phil-levis:
I think you're right, there isn't a universal interface for an early debug console. The thought is that this is a narrow and temporary case, so having a universal interface for it isn't high priority. E.g., you are always welcome to have your uart implementation implement a synchronous write (as you note), but that shouldn't be part of a kernel HIL.

One place where you do see something like this is `panic`, which use a synchronous writer, since it expects to never return.

bradjc:
For whatever it is worth, I tend to use `debug_gpio!()` for this initial bootup case.

a-pronin:
> One place where you do see something like this is panic, which use a synchronous writer, since it expects to never return.

Exactly.
And thanks for looking into the issue in https://github.com/tock/tock/pull/1941.
Meanwhile, how do we close this PR?

PR #1932, opened 2020-06-13
rv32i: fix PMP configuration

a-pronin:
This patch fixes how PMP registers are configured beyond the first two
regions (which occupy 4 blocks in pmpcfg0 - each unaligned region taking
two blocks). After that, the offsets & values used for pmpcfg1..3 were
often incorrect.


a-pronin:
Actually, conflicts with https://github.com/tock/tock/pull/1926. Given other changes, probably makes sense to wait for 1926, but need to fix the same things there.

a-pronin:
I don't seem to be able to properly rebase this PR on top of recent changes. Besides, more problems discovered in PMP configuration after recent refactoring.
I give up, https://github.com/tock/tock/pull/1969 is filed instead (and a couple of less critical fixes/improvement to follow).

PR #1933, opened 2020-06-14
Bind AppSlice to app instance (AppId), fixes #1914

### LINKED ISSUE #1914: AppSlices are not bounded to process instance lifetime
## Issue

While thinking about #1905, I noticed that when accessing an `AppSlice`, presumably no checks are performed with regards to whether the application the memory is referencing currently runs, or has been restarted in the meantime.

This essentially equals indefinite access to a bounded region of memory in the user space RAM area. I'm guessing this is not the expected behavior. If apps can be dynamically relocated in RAM upon restart that would be even worse.

https://github.com/tock/tock/blob/e5e781df0aba07a09e83badff43acab7c75cbdc1/kernel/src/mem.rs#L130-L134

While closely linked to #1905 and a solution to that would probably also fix this issue, these appear to be separate issues; with slices being accessible longer than the app's lifetime - as far as I can tell - being a bug with regards to Tock's design.

## Possible solutions

A check could be introduced to make sure that the app `allow`ing this memory is currently running, and that it's the same instance which originally `allow`ed it.
Given that the `AppId` associated identifier changes upon process restarts, this should be as simple as to check whether that identifier is still valid with respect of the process array.

Also, solutions to #1905 would probably apply, such as dynamically checking a kernel-held `allow` table, or enforcing a move of the shared `AppSlice` back into the kernel on both unallow & app restart.

lschuermann:
### Pull Request Overview

This pull request fixes #1914 by checking whether the memory an AppSlice points to still belongs to the same app instance (more precisely, the `instance` field in `AppSlice`).

Thanks to @alevy for implementing the main part of this which turned out to be working perfectly. The commits on top make sure that the returned `AppSlice` length is consistent with the `AsRef` and `AsMut` implementations and document this new behavior.

### Testing Strategy

This pull request was tested by developing a crude capsule which deliberately keeps an `AppSlice` out of a grant and does not accept a second `allow`. This ensures that the AppSlice is shared from the first app instance, but is used in the second app instance.

The test capsule (integrated with the nRF52840) along the userspace `libtock_c` app can be found [here](https://github.com/lschuermann/tock/commit/6403b2e4305177dccf968f4f5f65dfd62c84069f).

The output validates that indeed the `AppSlice` returns a length of 0 and hands out a immutable / mutable slice of length `0`.
```
NRF52 HW INFO: Variant: AAC0, Part: N52840, Package: QI, Ram: K256, Flash: K1024

--> AppSlice safety test app!
Allow buffer of length 64
AppSlice shared by app!
Tell capsule to print AppSlice info
Shared slice reports length 64
Shared slice as_ref length 64
Shared slice as_mut length 64

<fault app by button press>

--> AppSlice safety test app!
Allow buffer of length 64
AppSlice already shared, refusing to replace
Tell capsule to print AppSlice info
Shared slice reports length 0
Shared slice as_ref length 0
Shared slice as_mut length 0
```

### TODO

We should probably run the release test suite along with the `DebugProcessRestart` capsule to make sure this doesn't break any existing capsules. I can verify this once this gets "yes, we indeed want this" feedback. :smile: 

### Documentation Updated

  I believe there not to be a section in `/docs` which needs to elaborate on this behavior. I hope the generated rustdoc is sufficient. If indeed an update in `/docs` is required, please provide pointers to the respective section.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/mem.rs
+++ b/kernel/src/mem.rs
@@ -108,16 +108,22 @@ impl<L, T> AppSlice<L, T> {
 
 impl<L, T> AsRef<[T]> for AppSlice<L, T> {
     fn as_ref(&self) -> &[T] {
-        self.ptr.process.kernel.process_map_or(&[], self.ptr.process, |_| {
-            unsafe { slice::from_raw_parts(self.ptr.ptr.as_ref(), self.len) }
-        })
+        self.ptr
+            .process
+            .kernel
+            .process_map_or(&[], self.ptr.process, |_| unsafe {
+                slice::from_raw_parts(self.ptr.ptr.as_ref(), self.len)
+            })
     }
 }
 
 impl<L, T> AsMut<[T]> for AppSlice<L, T> {
     fn as_mut(&mut self) -> &mut [T] {
-        self.ptr.process.kernel.process_map_or(&mut [], self.ptr.process, |_| {
-            unsafe { slice::from_raw_parts_mut(self.ptr.ptr.as_mut(), self.len) }
-        })
+        self.ptr
+            .process
+            .kernel
+            .process_map_or(&mut [], self.ptr.process, |_| unsafe {
+                slice::from_raw_parts_mut(self.ptr.ptr.as_mut(), self.len)
+            })
     }
 }
--- a/kernel/src/mem.rs
+++ b/kernel/src/mem.rs
@@ -97,24 +97,77 @@ impl<L, T> AppSlice<L, T> {
         }
     }
 
+    /// Returns an iterator over the slice
+    ///
+    /// See
+    /// [`std::slice::iter()`](https://doc.rust-lang.org/std/primitive.slice.html#method.iter).
+    ///
+    /// Internally this uses
+    /// [`AsRef`](struct.AppSlice.html#impl-AsRef<[T]>), hence when
+    /// the app dies, restarts or the
+    /// [`AppId`](crate::callback::AppId) changes for any other
+    /// reason, the iterator will be of zero length.
     pub fn iter(&self) -> slice::Iter<T> {
         self.as_ref().iter()
     }
 
+    /// Returns an iterator that allows modifying each value
+    ///
+    /// See
+    /// [`std::slice::iter_mut()`](https://doc.rust-lang.org/std/primitive.slice.html#method.iter_mut).
+    ///
+    /// Internally this uses
+    /// [`AsMut`](struct.AppSlice.html#impl-AsMut<[T]>), hence when
+    /// the app dies, restarts or the
+    /// [`AppId`](crate::callback::AppId) changes for any other
+    /// reason, the iterator will be of zero length.
     pub fn iter_mut(&mut self) -> slice::IterMut<T> {
         self.as_mut().iter_mut()
     }
 
+    /// Iterate over `chunk_size` elements at a time, starting at the
+    /// beginning of the AppSlice.
+    ///
+    /// See
+    /// [`std::slice::chunks()`](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks).
+    ///
+    /// Internally this uses
+    /// [`AsRef`](struct.AppSlice.html#impl-AsRef<[T]>), hence when
+    /// the app dies, restarts or the
+    /// [`AppId`](crate::callback::AppId) changes for any other
+    /// reason, a [`Chunks`](core::slice::Chunks) iterator of zero length will
+    /// be returned.
     pub fn chunks(&self, size: usize) -> slice::Chunks<T> {
         self.as_ref().chunks(size)
     }
 
+    /// Mutably iterate over `chunk_size` elements at a time, starting at the
+    /// beginning of the AppSlice.
+    ///
+    /// See
+    /// [`std::slice::chunks_mut()`](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks_mut).
+    ///
+    /// Internally this uses
+    /// [`AsMut`](struct.AppSlice.html#impl-AsMut<[T]>), hence when
+    /// the app dies, restarts or the
+    /// [`AppId`](crate::callback::AppId) changes for any other
+    /// reason, a [`ChunksMut`](core::slice::ChunksMut) iterator of zero length will
+    /// be returned.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `chunk_size` is 0.
     pub fn chunks_mut(&mut self, size: usize) -> slice::ChunksMut<T> {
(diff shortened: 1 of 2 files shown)


kernel/src/mem.rs
```diff
@@ -77,8 +62,16 @@ impl<L, T> AppSlice<L, T> {
     }
 
     /// Number of bytes in the `AppSlice`.
+    ///
+    /// If the app died, has restarted, or its AppId identifier
+    /// changed for any other reason, return an accessible length of
+    /// zero, consistent with the [`AsRef`](struct.AppSlice.html#impl-AsRef<[T]>)
+    /// and [`AsMut`](struct.AppSlice.html#impl-AsMut<[T]>) implementations.
     pub fn len(&self) -> usize {
-        self.len
```
alevy, line 81: Oh good catch!

kernel/src/mem.rs
```diff
@@ -104,31 +97,93 @@ impl<L, T> AppSlice<L, T> {
...
+    /// Internally this uses
+    /// [`AsMut`](struct.AppSlice.html#impl-AsMut<[T]>), hence when
+    /// the app dies, restarts or the
+    /// [`AppId`](crate::callback::AppId) changes for any other
+    /// reason, the iterator will be of zero length.
     pub fn iter_mut(&mut self) -> slice::IterMut<T> {
         self.as_mut().iter_mut()
     }
 
+    /// Returns an iterator over the slice
```
lschuermann, line 128: Whoop, this should be the description of `chunks` instead.

lschuermann, line 128: Whoop, this should be the description of chunks instead.

kernel/src/mem.rs
```diff
@@ -104,31 +97,93 @@ impl<L, T> AppSlice<L, T> {
...
+    /// [`AsRef`](struct.AppSlice.html#impl-AsRef<[T]>), hence when
+    /// the app dies, restarts or the
+    /// [`AppId`](crate::callback::AppId) changes for any other
+    /// reason, a [`Chunks`](core::slice::Chunks) iterator of zero length will
+    /// be returned.
     pub fn chunks(&self, size: usize) -> slice::Chunks<T> {
         self.as_ref().chunks(size)
     }
 
+    /// Returns an iterator that allows modifying each value
```
lschuermann, line 143: Here as well.

lschuermann:
Given the potentially high impact of this, I'll try to run the release test suite today. I'm not expecting anything to fail, but this way we'd be on the safe side.

lschuermann:
### [WIP] tests on nRF52840DK
(_Explanations shamelessly stolen from @ppannuto's nRF52840DK testing of Release 1.5_)

I've changed the kernel to restart all apps in case of a fault and integrated the `DebugProcessRestart` bound to button 1 on the board. I'm expecting the apps behave _exactly_ the same after the manually triggered fault, as that would mean that all shared AppSlices are handled as expected. This may not uncover _all_ potential issues but should give a good impression of the impact of this.

* All Boards
  * [x]  `examples/sensors`
  * [ ]  `examples/services/ble-env-sense` and `examples/services/ble-env-sense/test-with-sensors`
    * **Actually causes the entire kernel to restart without any indication of panic etc.** Though probably has nothing to do with this change, as the app isn't even deliberately faulted. Will investigate further.
  * [x]  `examples/c_hello` and `examples/tests/printf_long`
  * [x]  `examples/tests/console_recv_short` and `examples/tests/console_recv_long`
    `[LONG] Error doing UART receive: -2`
    but console is responsive and shows identical behavior on faulted app
  * [x]  `examples/blink`
  * [ ]  `examples/rot_client` and `examples/rot_service`
    * [x] With only client: `No rot13 service`
    * [ ] Flashing the service makes my board unusable. Have to flash all zeros over kernel and app memory to revive it. Hangs in a loop, trying to print the initialization message:
           `Ini\x374Ini\x374Ini\x374Ini\x374` etc.
           Using tockloader 4.0.0 for flashing kernel & managing apps.
  * [ ]  `examples/blink` and `examples/c_hello` and `examples/buttons`
          After fault:
          `No available GPIOTE interrupt channels`
          `Hello World!`
          Though unlikely to be related to the AppSlice issue.
  * [x]  `examples/lua-hello`
  * [x]  `examples/tests/console_timeout`
  * [x]  `examples/tests/malloc_test01`
  * [x]  `examples/tests/stack_size_test01`
  * [x]  `examples/tests/stack_size_test02`
  * [x]  `examples/tests/mpu_stack_growth`
    Continuous loop (restart policy):
    ```
    stack: 0x200043e0 - buffer: 0x200043e0 - at_least: 0x   0
    write to 0x200043e0
    read from 0x200043e0
    value 33
    ```
  * [ ]  `examples/tests/mpu_walk_region`
     
  * [ ]  `examples/tests/multi_alarm_test`
     
  * [ ]  `examples/tests/adc`
     
  * [ ]  `examples/tests/adc_continuous`
     
  * [ ]  `examples/tutorials/05_ipc/led` and `examples/tutorials/05_ipc/rng` and `examples/tutorials/05_ipc/logic`
     
  * [ ]  `examples/tests/gpio` with mode set to 0
    <!-- Touch `P0.03` to Vdd (randomly tested with a few other pins to _not_ fire an interrupt):
    ```
    GPIO Test Application
    Print userspace GPIO 0 pin reading whenever its value changes
    Jump pin high to test
    GPIO Interrupt!
    GPIO Interrupt!
    ``` -->
* nRF specific
  * [ ]  `examples/ble_advertising`
    <!-- ![tock-ble](https://user-images.githubusercontent.com/5341193/80621896-464dbe80-8a48-11ea-81c8-315f3358a8f3.jpg) -->
  * [ ]  `examples/ble_passive_scanning`
    <!-- Address and data redacted for privacy reasons, addresses cross-checked with Nordic nRF Connect app
    ```
    [Tutorial] BLE Passive Scanner
    --------------------------LIST-------------------------
    PDU Type: 0 ADV_IND
    PDU TxAdd: 1
    PDU RxAdd: 0
    PDU Length: 24
    Address: XX XX XX XX XX XX
    Data: XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX
    --------------------------END---------------------------
    ``` -->

### [preliminary] conclusion

Although I haven't conducted all tests yet, I'm already seeing a lot of weird behavior. It could be that my setup is not alright or my nRF52840DK is broken somehow, but with a few things I'm actually fearing that some unsafe code is writing to the location of the returned slice from the AppSlice (which should be either on stack or undefined for an empty slice? In any case, I'll conduct the rest of the tests and also test my hail, whether it shows identical symptoms.

PR #1934, opened 2020-06-14
Initial support for Apollo3 BLE

alistair23:
### Pull Request Overview

This PR adds initial support for the Apollo3 BLE module.

This PR can power up the module and write data to it and receive interrupts.

~~This PR applies on top of: https://github.com/tock/tock/pull/1917~~

### Testing Strategy

Try to run the `libtock-c` examples on the Apollo3.

### TODO or Help Wanted

Lots of things still aren't done, if anyone is willing to help I would be very grateful.

BLE still doesn't work (I can't see the device advertise itself from other devices).

This PR also doesn't apply the binaries (called patches) that are required to use the BLE module. I have a branch where I call the to HAL's C code to do this.

I currently just see the `ble_advertisment_driver` alarm firing and the same data being written to the BLE device over and over again. I'm not sure what else should be happening, but BLE is currently not working.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


chips/apollo3/src/lib.rs
```diff
@@ -2,15 +2,17 @@
 
 #![crate_name = "apollo3"]
 #![crate_type = "rlib"]
-#![feature(llvm_asm, const_fn, naked_functions)]
+#![feature(llvm_asm, asm, const_fn, naked_functions)]
```
bradjc, line 5: Is `asm` used?

alistair23: Nope, removed.

alistair23:
Ping!

PR #1935, opened 2020-06-14
Makefile: Update the OpenTitan boot ROM

alistair23:
### Pull Request Overview

Updates the OT boot ROM to fix the failing CI.

### Testing Strategy

CI

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -483,11 +483,11 @@ endef
 define ci_setup_qemu_opentitan
 	$(call banner,CI-Setup: Get OpenTitan boot ROM image)
 	# Download OpenTitan image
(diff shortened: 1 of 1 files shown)


lschuermann:
Having the CI break randomly because of the CI mirror changing / deleting old artifacts is very unfortunate. Is there some other mirror we could use to achieve this? Does the Azure Dev pipeline make any guarantees regarding the availability of artifacts?

If not, can we mirror this ourselves and verify the integrity by hashing after download so we don't have to trust the mirror? If the (binary artifact's) license allows it, publishing to [Zenodo](https://zenodo.org/) could be an option.

PR #1939, opened 2020-06-15
Make the Field::mask and FieldValue::mask fields private.

gendx:
### Pull Request Overview

The `mask` fields don't need to be exposed publicly, as their use cases overlap with the existing functions to manipulate fields.

- This pull request makes them private.
- As a side effect, the `libraries/riscv-csr/src/csr.rs` is refactored to use the unified arithmetic methods introduced in #1766.

### Testing Strategy

### TODO or Help Wanted

- These changes currently overlap with #1926, although are compatible with that PR, as discussed in https://github.com/tock/tock/pull/1926#discussion_r440143726.
- The `Field::shift` field should also be made private, but is currently used in `chips/lowrisc/src/gpio.rs`. That code should probably be refactored so that the shift is not exposed as-is. https://github.com/tock/tock/blob/555b3a0441697d111a0330c01d392c15b36712f6/chips/lowrisc/src/gpio.rs#L94-L104
- Likewise, the `FieldValue::value` field should be made private, but is currently used in https://github.com/tock/tock/blob/555b3a0441697d111a0330c01d392c15b36712f6/libraries/riscv-csr/src/csr.rs#L76-L79

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/libraries/riscv-csr/src/csr.rs
+++ b/libraries/riscv-csr/src/csr.rs
@@ -58,14 +58,12 @@ impl<T: IntLike, R: RegisterLongName> ReadWriteRiscvCsr<T, R> {
 
     #[inline]
     pub fn read(&self, field: Field<T, R>) -> T {
-        (self.get() & (field.mask << field.shift)) >> field.shift
+        field.read(self.get())
     }
 
     #[inline]
     pub fn read_as_enum<E: TryFromValue<T, EnumType = E>>(&self, field: Field<T, R>) -> Option<E> {
-        let val: T = self.read(field);
-
-        E::try_from(val)
+        field.read_as_enum(self.get())
     }
 
     #[inline]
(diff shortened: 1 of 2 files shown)


andre-richter:
@ppannuto Can currently only check on mobile, therefore a bit limited in my judgement. I am using `mask` for the CPU traits, but it looks like I only need to adapt the same way as `libraries/riscv-csr/src/csr.rs` in this PR has been changed to be good?

bradjc:
Do we merge this first? Or #1926?

ppannuto:
bors d=[alistair23, gendx]

I'm happy to let Alistair and Guillaume figure out whichever is easier

ppannuto:
bors d=alistair23,gendx

I was close bors.

alistair23:
I merged this one and rebased my PR

PR #1941, opened 2020-06-15
debug: add flag for kernel_loop begin

ppannuto:
### Pull Request Overview

As noted in #1931, it's a bit non-obvious to folks working on board bringup that `debug!` can't be used before the kernel loop has begun.

This is a bit brute force-y of an implementation, but it's in line with the `DEBUG_WRITER` and kind of the ugly necessity of the lower-level debug stuff.

### Testing Strategy

Compiling.

### TODO or Help Wanted

N/A; feedback if folks have it.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/kernel/src/debug.rs
+++ b/kernel/src/debug.rs
@@ -355,7 +355,15 @@ pub struct DebugWriter {
 /// needed so the debug!() macros have a reference to the object to use.
 static mut DEBUG_WRITER: Option<&'static mut DebugWriterWrapper> = None;
 
+/// The `debug!` macro only works correctly once the kernel loop is running.
+/// This is a guard to help prevent folks working on board bringup from
+/// accidentally trying to use `debug!` too early in boot.
+pub(crate) static mut DEBUG_KERNEL_LOOP_STARTED: bool = false;
+
 pub unsafe fn get_debug_writer() -> &'static mut DebugWriterWrapper {
+    if !DEBUG_KERNEL_LOOP_STARTED {
+        panic!("Cannot use `debug!` before kernel_loop");
+    }
     DEBUG_WRITER
         .as_deref_mut()
         .expect("Must call `set_debug_writer_wrapper` in board initialization.")
--- a/kernel/src/hil/emergency_serial.rs
+++ b/kernel/src/hil/emergency_serial.rs
@@ -1,0 +1,28 @@
+//! Hardware interface layer (HIL) for Emergency Serial
+//!
+
+use crate::returncode::ReturnCode;
+
+/// This is a serial-like interface designated for use only in exceptional
+/// situations. In contrast to other kernel interfaces, all methods here are
+/// synchronous. Implementations of this HIL should be as conservative as
+/// possible, use as few chip/board features as possible, and rely on as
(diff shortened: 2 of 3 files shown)


bradjc:
Did you run this? Won't this crash every board?

ppannuto:
> Testing Strategy
> 
> Compiling.

Nope.

Because of the "initialization complete" lines? Yeah, that's a good point. Though those are perhaps even further misleading to people, as it would lead folks to believe that `debug!` is usable in board main.rs's, when it's really not.

I need to think about this more; will circle back to it in a bit.

bradjc:
I'm not sure what you mean by work/usable. If you call `debug!()`, a UART transmission _could_ be triggered (based on how the debug stack is configured). If you use `virtual_uart`, then a deferred call will be registered. In both cases, everything is working as expected. Do you mean you want to re-visit virtual_uart? I imagine that virtual uart could only register a deferred call if there is an error, otherwise the write could go through immediately.

a-pronin:
Won't try to comment how acceptable this approach is w/o knowing the history of this design. But a related question: In addition, would it be reasonable to introduce a combo of set_sync_writer(), which requires the same sync writer and sets it in a global var (similar to set_debug_writer_wrapper and DEBUG_WRITER), and debug_sync! macro that uses that writer if available (nop if not) to provide a way to print out from places like process.rs before the main loop starts (or when debugging virtual_uart.rs itself, for example).
All boards implement such sync writers anyway for panic.

bradjc:
> Won't try to comment how acceptable this approach is w/o knowing the history of this design. But a related question: In addition, would it be reasonable to introduce a combo of set_sync_writer(), which requires the same sync writer and sets it in a global var (similar to set_debug_writer_wrapper and DEBUG_WRITER), and debug_sync! macro that uses that writer if available (nop if not) to provide a way to print out from places like process.rs before the main loop starts (or when debugging virtual_uart.rs itself, for example).
> All boards implement such sync writers anyway for panic.

I could see that as another debugging tool, absolutely. However, I think I would prefer that it come with a clear disclaimer that we won't ever merge a PR that uses the `debug_sync!()` macro, as a way to make it clear that it is invalid according to Tock's asynchronous design.

kernel/src/debug.rs
```diff
@@ -355,7 +355,15 @@ pub struct DebugWriter {
...
 
+/// The `debug!` macro only works correctly once the kernel loop is running.
+/// This is a guard to help prevent folks working on board bringup from
+/// accidentally trying to use `debug!` too early in boot.
+pub(crate) static mut DEBUG_KERNEL_LOOP_STARTED: bool = false;
+
 pub unsafe fn get_debug_writer() -> &'static mut DebugWriterWrapper {
+    if !DEBUG_KERNEL_LOOP_STARTED {
+        panic!("Cannot use `debug!` before kernel_loop");
+    }
```
hudson-ayers, line 366: I think this use of static mut (reading from the reference in one place and mutating it elsewhere) is technically UB, even if it would be unlikely to do anything other than what you expect here in practice.

bradjc:
Should we close this?

Issue #1900, opened 2020-06-02
Console can report the wrong number of received bytes


bradjc:
A UART can return more bytes than the console asked for. This doesn't cause a crash because of how the console capsule copies into the appslice:

https://github.com/tock/tock/blob/1aa6e1bc071e7916205c3968356728d6dec9d423/capsules/src/console.rs#L358-L378

If `rx_len` is longer than the size of `app_buffer`, the `app_buffer.iter_mut()` makes sure we don't copy too many bytes. However, in the callback, we use the rx_len.

### Why an issue an not a PR?

I'm not _entirely_ sure what the correct fix is. We can certainly check for how many bytes we copied, and report that instead. Should there be an error? Is it ok if we drop any extra bytes?

ppannuto:
Since an app is calling down with a fixed buffer length, I think the solution would be to not allow the underlying UART hardware to receive too much. That is, [instead of passing the whole console capsule buffer here, pass a slice up to the length asked for by the app](https://github.com/tock/tock/blob/1aa6e1bc071e7916205c3968356728d6dec9d423/capsules/src/console.rs#L171).

In that way, we match the semantics of the underlying UART hardware (which I believe will drop bytes in the case of buffer overrun).

bradjc:
But then we lose access to the rest of that buffer forever, right?

hudson-ayers:
Aha..sounds like a job for [LeasableBuffer!](https://github.com/tock/tock/blob/master/kernel/src/common/leasable_buffer.rs)

bradjc:
The other problem, is what happens if the hardware _does_ receive too much? For example, serial over USB isn't a real UART bus, and we might get longer packets from the host than the console asked for. Or a UART implementation might just want to return all of the bytes it has received into a buffer by the time it reads from the HW buffer.

ppannuto:
At a certain point, there's just nothing you can do about this case if the bus doesn't have flow control.

Technically, the serial standard includes RTS/CTS [ready-to-send / clear-to-send] pins for this purpose. I assume that's therefore part of the serial-over-USB protocol.

In the absence of flow control, I think the best/only thing that the underlying driver can do is report to the upper layers when bytes were dropped due to buffer overrun.

bradjc:
That's not the only thing, the driver can also just pass up more bytes than were asked for (of course until the buffer runs out of space). That is what can happen right now.

ppannuto:
[...]

PR #1942, opened 2020-06-15
Console: Check for receive length just in case

bradjc:
### Pull Request Overview

This pull request adds a check in the console capsule to ensure that the console doesn't report that more bytes were received than were actually copied into the buffer. This shouldn't ever happen, but a buggy driver could lead to this behavior. This prevents that bug from propagating to userspace. However, userspace might want to know that the kernel is broken, so an error is reported since likely bytes were lost due to this bug.

Also fixes a comment in UART HIL.

### Testing Strategy

Hopefully simple enough to not be needed.

### TODO or Help Wanted

I could see closing this PR, since if we write UART HIL implementations according to the spec this case won't happen. But I wanted to open it to wrap up #1900.

### Documentation Updated

I'm not actually sure how the console syscall doc matches the console driver, but the return values weren't specified, so I didn't make any changes.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/console.rs
+++ b/capsules/src/console.rs
@@ -368,7 +368,36 @@ impl uart::ReceiveClient for Console<'_> {
                                         for (a, b) in app_buffer.iter_mut().zip(rx_buffer) {
                                             *a = *b;
                                         }
-                                        cb.schedule(From::from(rcode), rx_len, 0);
+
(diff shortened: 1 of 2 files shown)


bradjc:
Lonely PR.

PR #1943, opened 2020-06-16
kernel: TBF parsing: match the threat model

bradjc:
I was testing Tockloader, and a bug I introduced caused the header_len field to be 14 (when it should have been 16). This caused the kernel to give me an internal kernel error, which is against the threat model. This change fixes that issue, and gets the TBF parsing logic inline with the threat model.

The main change is being much more clear about what errors can be returned with doing the initial parsing of the tbf header in flash. Now there are only two options: we are at the end of apps, or we return the length of the entire app (which is a trusted field). This allows us to more confidently stop parsing for apps, as well as skip over invalid apps.

### Testing Strategy

I tested this with tockloader to corrupt the header length field of an app, and it no longer causes an internal error. Additionally, I flashed two apps, and corrupted the first header length, and the second app was successfully parsed and executed.

### TODO or Help Wanted

I have one question for @jrvanwhy: is it OK that I am assuming that the version is trusted as part of the total length being trusted? If not, then we have to assume that the total length is _always_ 4 bytes into the tbf header, which we might not want to do. Right now, if the version doesn't match I assume the "app" is just erased flash at the end of the app linked list. 

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -163,16 +163,16 @@ pub fn load_processes<C: Chip>(
             // Pass the first eight bytes to tbfheader to parse out the length
             // of the tbf header and app. We then use those values to see if we
             // have enough flash remaining to parse the remainder of the header.
-            let (version, header_length, app_length) = match tbfheader::parse_tbf_header_lengths(
+            let (version, header_length, entry_length) = match tbfheader::parse_tbf_header_lengths(
                 test_header_slice
                     .try_into()
                     .or(Err(ProcessLoadError::InternalError))?,
             ) {
                 Ok((v, hl, al)) => (v, hl, al),
-                Err(tbfheader::InitialTbfParseError::InvalidHeader(app_length)) => {
+                Err(tbfheader::InitialTbfParseError::InvalidHeader(entry_length)) => {
                     // If we could not parse the header, then we want to skip
                     // over this app and look for the next one.
-                    (0, 0, app_length)
+                    (0, 0, entry_length)
                 }
                 Err(tbfheader::InitialTbfParseError::UnableToParse) => {
                     // Since Tock apps use a linked list, it is very possible
@@ -183,21 +183,22 @@ pub fn load_processes<C: Chip>(
                 }
             };
 
-            // Now we can get a slice which only encompasses the app. We will
-            // either parse this as an actual app, or skip over this region.
-            let app_flash = remaining_flash
-                .get(0..app_length as usize)
+            // Now we can get a slice which only encompasses the length of
+            // flash described by this tbf header.  We will either parse this
+            // as an actual app, or skip over this region.
+            let entry_flash = remaining_flash
+                .get(0..entry_length as usize)
                 .ok_or(ProcessLoadError::NotEnoughFlash)?;
 
-            // Check if we should continue parsing this app. If the initial
-            // header parsing failed (i.e. the header_length is 0), we don't
-            // bother.
-            let memory_offset = if header_length > 0 {
+            // Determine how much of the available app memory this entry uses.
+            // Either enough for the newly created process, or none if this
+            // entry is not a process.
+            let process_memory_length = if header_length > 0 {
                 // Try to create a process object from that app slice.
-                let (process, memory_offset) = Process::create(
+                let (process, process_memory_length) = Process::create(
                     kernel,
                     chip,
-                    app_flash,
+                    entry_flash,
                     header_length as usize,
                     version,
                     app_memory_ptr,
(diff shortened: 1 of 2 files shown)


kernel/src/process.rs
```diff
@@ -178,44 +183,51 @@ pub fn load_processes<C: Chip>(
...
-                        i,
-                        app_flash.as_ptr() as usize,
-                        app_flash.as_ptr() as usize + app_flash.len(),
-                        app_memory_ptr as usize,
-                        app_memory_ptr as usize + memory_offset,
-                        process.map(|p| p.get_process_name())
-                    );
+            // Check if we should continue parsing this app. If the initial
+            // header parsing failed (i.e. the header_length is 0), we don't
+            // bother.
```
ppannuto, line 194: Something about either the logic or the phrasing here is confusing.

If "initial header parsing failed", shouldn't `parse_tbf_header_lengths` have returned an error and we would never get here?

bradjc: `parse_tbf_header_lengths` can fail for basically two reasons: either we are at the end of apps or there is a corrupt app header. If we are at the end of apps we have returned already above. This check differentiates a successful initial header parse from a corrupt header.

jrvanwhy:
> I have one question for @jrvanwhy: is it OK that I am assuming that the version is trusted as part of the total length being trusted? If not, then we have to assume that the total length is _always_ 4 bytes into the tbf header, which we might not want to do. Right now, if the version doesn't match I assume the "app" is just erased flash at the end of the app linked list.

That seems fine to me.

bradjc:
Also if any Rustaceans are out there, this PR isn't _quite_ the code that I want (although it is definitely good enough). In `tbfheader::parse_tbf_header_lengths()`, the slice being parsed is a fixed length. That means it should be possible to extract subranges from it with hard-coded start and end points that can never panic and without requiring an `Option` (and the resulting error check). That is:

```rust
pub(crate) fn parse_tbf_header_lengths(
    app: &'static [u8; 8],
) -> Result<(u16, u16, u32), InitialTbfParseError> {
    let version = u16::from_le_bytes(
        app.get(0..2)
            .ok_or(InitialTbfParseError::UnableToParse)?
            .try_into()?,
    );
```
should, in at least my mind, be able to be written as:
```rust
pub(crate) fn parse_tbf_header_lengths(
    app: &'static [u8; 8],
) -> Result<(u16, u16, u32), InitialTbfParseError> {
    let version = u16::from_le_bytes(app.get(0..2));
```
since `app` is exactly 8 bytes long, the `get()` range is within those bounds, and a u16 requires exactly two bytes. I don't think there is any way that code can fail. As written now, it looks like it can return an `UnableToParse` error (for two different reasons) when it really can't. I guess from a correctness point of view this is OK, but the code is a little misleading when reading it. It also forces me to make a choice: do I re-use the `UnableToParse` error, conflating what that error really means to a reader of the code, or add a new error to make it clear this is a separate issue and then make the caller deal with it, or use an API that can potentially panic which would be fine now, but could lead to issues if the code is updated in the future? I chose to re-use the error, but I'm not sure what is the best option.

I don't think there is a `safe` way to implement the second version, however, but maybe I'm missing something? To me, having a slice type that can never panic (both for fixed slices like this and other slices) would be a huge addition for Rust since getting this buffer parsing logic right is really tedious. Another huge boost would be pushing more checks into the compiler so that programmers only have to handle error cases that are actually relevant to their code.

ppannuto:
```diff
-    let version = u16::from_le_bytes(
-        app.get(0..2)
-            .ok_or(InitialTbfParseError::UnableToParse)?
-            .try_into()?,
-    );
+    let version = u16::from_le_bytes([app[0], app[1]]);
```

I think this does the right thing?

bradjc:
`[]` can panic, no?

ppannuto:
Rust is a bit smarter than that. The argument to the function is a fixed array of size 8, so it can safely elide the range checks.

bradjc:
Ah yes, OK that's what I thought as well, and I think that is why there isn't more work in this area. My concern is that panic can be subtly re-introduced without ever touching that line of code again if the function signature is changed to `&[u8]`.

ppannuto:
Indeed :/

Something like this would be interesting: https://github.com/dtolnay/no-panic, though I'm not sure there's anything sanctioned in Rust proper.

bradjc:
I would eventually like to spin tbfheader.rs into its own crate merged with elf2tab's TBF code. Then use it as a case study on if it is possible to write low-level buffer manipulation code without unsafe and without ever being able to `panic!()`.

kernel/src/process.rs
ppannuto: I just pushed some comment and name updates that helps clarify what was confusing me. The key for me was making explicit a tbf entry, that is the region of flash described by a tbf header, which may or may not actually be an app (I was really hung up on why there was any meaningful "app_flash" in the case where it's not a valid app).

kernel/src/process.rs
```diff
@@ -183,21 +183,22 @@ pub fn load_processes<C: Chip>(
...
-            // header parsing failed (i.e. the header_length is 0), we don't
-            // bother.
-            let memory_offset = if header_length > 0 {
+            // Determine how much of the available app memory this entry uses.
+            // Either enough for the newly created process, or none if this
+            // entry is not a process.
+            let process_memory_length = if header_length > 0 {
                 // Try to create a process object from that app slice.
-                let (process, memory_offset) = Process::create(
+                let (process, process_memory_length) = Process::create(
```
bradjc, line 198: `process_memory_length` is technically not correct. However, an older PR of mine that I trampled all over with this PR would clarify it, and we should be using slices anyway (another branch I have that I trampled all over with this change). This return value is really `memory_bytes_consumed`, and the process's start of memory does not have to be at the same address of the range passed to `create()`. Just FYI if you are curious, and to help motivate my other (and forthcoming) PRs!

ppannuto: Yeah, that all makes sense. I look forward to future PRs and even better names 👍

PR #1945, opened 2020-06-16
nrf52: usb: do not print debug!() info on startup

bradjc:
Debug statements at the beginning of the USB driver init process conflict with using the USB stack for UART/debug!(), since those debug outputs cause the USB driver to try to start sending before it is fully initialized. Eventually, an assertion fails and everything crashes.

They are also a little noisy for users not debugging the USB stack.

### Testing Strategy

This pull request was tested by getting CDC working on the nano33ble board.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


brghena:
Disabling debug in the USB stack by default seems good.

bors r+

PR #1949, opened 2020-06-17
CDC: Wait for CDC messages, add delay to nRF USB RESET

bradjc:
### Pull Request Overview

This pull request allows the CDC layer to work on nRF even if it transmits a message immediately. We now wait for CDC messages that correlate with a CDC client being connected before requesting an endpoint IN transfer.

This also includes a delay for the nRF chip after a USB RESET event. Without this, the nRF chip could setup an IN transfer, but not ever get the interrupt that the transfer finished. According to the datasheet, we are supposed to wait after a USB reset event:

image

I experimentally determined a long enough wait such that the IN transfers would work.

### Testing Strategy

This pull request was tested by running a hello loop app on nano33ble.

### TODO or Help Wanted

This allows CDC to work as expect on my mac: all messages sent are displayed in the terminal after running `tockloader listen`. On linux, however, I'm still seeing the very first message be lost.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/usb/cdc.rs
+++ b/capsules/src/usb/cdc.rs
@@ -51,6 +51,28 @@ pub const MAX_CTRL_PACKET_SIZE_IBEX: u8 = 64;
 
 const N_ENDPOINTS: usize = 3;
 
+/// States of the CDC driver.
+#[derive(Debug, Copy, Clone, PartialEq)]
+enum State {
+    /// Default state. User must call `enable()`.
+    Disabled,
+    /// `enable()` has been called. The descriptor format has been passed to the
+    /// hardware.
+    Enabled,
+    /// `attach()` has been called. The hardware should be ready for a host to
+    /// connect.
+    Attached,
+    /// The host has enumerated this USB device. Things should be functional at
+    /// this point.
+    Enumerated,
+    /// We have seen the CDC messages that we expect to signal that a CDC client
+    /// has connected. We stay in the "connecting" state until the USB transfer
+    /// has completed.
+    Connecting,
+    /// A CDC client is connected. We can safely send data.
+    Connected,
+}
+
 /// Implementation of the Abstract Control Model (ACM) for the Communications
 /// Class Device (CDC) over USB.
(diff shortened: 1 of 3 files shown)


chips/nrf52/src/usbd.rs
```diff
@@ -1305,6 +1305,16 @@ impl<'a> Usbd<'a> {
 
         self.dma_pending.set(false);
 
+        // Wait for at least T_RSTRCY for the hardware to be ready after the USB
+        // RESET (§6.35.6). I measured the loop using GPIO pins from `0..800000`
+        // as a 62.5 ms delay, and that was enough to allow the CDC layer to
+        // work. I tried shorter time than that (`0..700000`, measured at 54.7
+        // ms), but then the EPDATA event on the very first IN transfer
+        // immediately after the `client.bus_reset()` call below never occurs.
+        for _ in 0..800000 {
```
ppannuto, line 1314: :(

We have a few of these in the kernel now. We should add a delay primitive at some point. That shouldn't block this PR.

capsules/src/usb/cdc.rs
```diff
@@ -229,24 +252,61 @@ impl<'a, U: hil::usb::UsbController<'a>> hil::usb::Client<'a> for CdcAcm<'a, U>
...
+                            self.state.set(State::Enumerated)
+                        }
+                    }
+                }
+                0x23 => {
+                    // On Mac, we seem to get the SEND_BREAK to signal that a
+                    // client disconnects.
+                    self.state.set(State::Enumerated)
+                }
+                _ => {}
```
ppannuto, line 306: Is there a bunch of unrelated traffic, or maybe future debugging efforts will be grateful for a debug print `Ignoring unknown CDC control message: {}`

bradjc: I don't think there are messages generally worth caring about, and it might vary based on implementations/operating systems. But in any case I don't think anyone would care about what the messages are.

PR #1951, opened 2020-06-17
ci: run miri where we can

### LINKED ISSUE #1713: Using Miri to help checking unsafe code?
I've recently seen a talk about `unsafe` code, in which the speaker recommends using Miri as a tool to help finding bugs in `unsafe` code: https://youtu.be/QAz-maaH0KM?t=1426. Given the amount of `unsafe` code in Tock, this could be a helpful tool to find and fix potential bugs.

I haven't tried it, and I'm not sure in which scope it could be used (given the specifics of embedded systems), but it could be relevant for the Tock-cells library for example (which are architecture agnostic, and don't depend on memory mapped registers).

For now it's just an idea, but it seems that it could be deployed a check in CI for example.

ppannuto:
### Pull Request Overview

This pull request runs miri where we can.

Closes #1713.

I don't think this does terribly much in practice today, but who knows, maybe someday it'll catch a bug.

### Testing Strategy

Compiling.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -285,7 +285,8 @@ ci-runner-github-tests:\
 	ci-job-kernel\
 	ci-job-capsules\
 	ci-job-chips\
-	ci-job-tools
+	ci-job-tools\
+	ci-job-miri
 	$(call banner,CI-Runner: GitHub tests runner DONE)
 
 .PHONY: ci-runner-github-qemu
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -470,6 +471,25 @@ ci-job-tools: ci-setup-tools
...
+	@# This is highly experimental at the moment, so just run miri in the
+	@# places it's known to work and there are any tests.
+	@#
+	@# Hangs forever during `Building` for this one :shrug:
+	@#cd libraries/tock-register-interface && CI=true cargo miri test
+	@cd kernel && CI=true cargo miri test
+	@cd arch/cortex-m && CI=true cargo miri test
+	@cd arch/cortex-m0 && CI=true cargo miri test
+	@cd arch/cortex-m3 && CI=true cargo miri test
+	@cd arch/cortex-m4 && CI=true cargo miri test
```
hudson-ayers, line 489: why not include arch/rv32i? or just use list_archs.sh the same way you used list_chips.sh?

ppannuto: Good point; fixed.

bradjc:
You should add miri to the update rust script so we get a nightly that has it in the future.

ppannuto:
Rustup manages it correctly for us. Rust toolchain ecosystem++ for this one. Rustup is aware of the `rust-toolchain` that's active, and so the `rustup component` bits will do the right thing automatically whenever the `rust-toolchain` file changes.

bradjc:
Ah ok the name for what I'm thinking of is "profiles" https://blog.rust-lang.org/2019/10/15/Rustup-1.20.0.html#profiles, and it appears that rustup won't guarantee that if we update rust we will get a nightly that has a valid copy of miri. https://rust-lang.github.io/rustup-components-history/

ppannuto:
Ah, interesting. I don't see any way of doing that automatically, so I added it to the prompt for the user to validate.

bradjc:
It looks like maybe we can pass `--component miri` when rustup chooses a version. https://github.com/rust-lang/rustup#installing-nightly

ppannuto:
Okay, I removed the component selection and rust upgrade parts from this PR and moved them to #1974, as that seems to really be a more general problem given that we weren't doing this for any of the many components we were using before.

PR #1760, opened 2020-04-14
LoRa radio capsule and nRF52 board i/f.

nikulshr:
### Pull Request Overview

Adds a LoRa interface for mirco-controlling a LoRa modem (SX127X) using SPI. So far, the PHY layer has been implemented. Note: Changes to the nRF52 board are mainly to add the underlying SPI mux and driver component.

### Testing Strategy

Tested on MCCI Catena 4610 (SX1276) using Nordic's nRF52840dk by checking for SPI communication over a logic analyzer. TBD: Check for actual OTA packets.

### TODO or Help Wanted

Needs some review of the existing SPI (enable certain blocks currently commented, cleaner code, bugs), general Tock interfaces and possibly LoRa functions.

### Documentation Updated

- Added README under capsules/src/lora

### Formatting
- Ran `make formatall`

PR #1953, opened 2020-06-18
nrf: fixup Platform struct

ppannuto:
### Pull Request Overview

In the process of pulling #1760 up to master, I noticed that the big Nordic refactor mis-ordered several of the `Platform` elements.

I am extremely confused right now. Are capsule types somehow so weak that any capsule can slot in for any other? Or is Rust just matching up the types for the struct elements auto-magically independent of their order?

### Testing Strategy

Compiling... though I'm losing confidence in that.

### TODO or Help Wanted

Why did the existing code compile?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nordic/nrf52840_dongle/src/main.rs
+++ b/boards/nordic/nrf52840_dongle/src/main.rs
@@ -292,18 +292,18 @@ pub unsafe fn reset_handler() {
     nrf52_components::NrfClockComponent::new().finalize(());
 
     let platform = Platform {
-        button,
         ble_radio,
         ieee802154_radio,
+        button,
         pconsole,
         console,
-        led,
         gpio,
+        led,
         rng,
         temp,
-        alarm,
-        analog_comparator,
         ipc: kernel::ipc::IPC::new(board_kernel, &memory_allocation_capability),
+        analog_comparator,
+        alarm,
     };
 
     platform.pconsole.start();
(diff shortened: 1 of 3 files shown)


ppannuto:
Well, update to myself. Apparently you can write these in any order so long as the types are unique and Rust will just magically match them up. I do not like that feature Sam-I-Am.

Still, I think it's slightly cleaner for the order to match, and am leaving this PR open.

ppannuto:
@hudson-ayers helpfully corrected me that it's nothing to do with types, and apparently this black magic works because the variable name matches the structure field name (i.e. it replaces `button: button,`, with just `button,`).

I find the readability of that criminally bad, but maybe that's just my history and biases.

_Still_, I think ordering them to match the declaration is useful.

ppannuto:
The evidence is mounting that I'm on the wrong side of this somehow. I went to clippy to see if there was a way to turn off this nonsense and require explicit field names... turns out the have a lint for the opposite: https://rust-lang.github.io/rust-clippy/master/#unneeded_field_pattern

hudson-ayers:
yeah, rust-analyzer will warn by default if you include the unneeded field name, so I tend to not include them so my editor doesn't yell at me

ppannuto:
Upon brief reflection, I actually think this will eventually come to join the long list of things that annoyed me at first about Rust and later will annoy me that other languages don't do. Been a while since I hit one of those, got caught a bit off guard :)

bradjc:
I think the issue here is Rust adding it later and making it optional. This is something the language should decide, so there is no ambiguity for users or discussions on doing it one way or another.

ppannuto:
I think I've decided this is just noise, so I'm just going to close it :/

PR #1960, opened 2020-06-19
Increase delay in Segger RTT debugging (fix #1697).

### LINKED ISSUE #1697: [Bug] Corrupted console output with large buffers.
### System specifications

- nRF52840-DK board
- `USB_DEBUGGING` enabled on the board's `src/main.rs`
- [Latest Tock kernel](https://github.com/tock/tock/tree/527afb41687da5b53b8e65c6033dba23d3b24c13), commit 527afb41687da5b53b8e65c6033dba23d3b24c13

### Commands run

On Tock repository.

```shell
cd boards/nordic/nrf52840dk
make flash
```

On libtock-rs at my [console-check branch](https://github.com/gendx/libtock-rs/tree/console-check/).

```shell
./run_example console_triangle
```

### Expected behavior

The `console_triangle` example from my `console-check` branch on `libtock-rs` is using the console driver with buffers of increasing size (hence the "triangle") https://github.com/gendx/libtock-rs/blob/console-check/examples/console_triangle.rs.

The output is supposed to look like this:

```
1
22
333
4444
55555
666666
7777777
88888888
999999999
0000000000
11111111111
222222222222
3333333333333
44444444444444
555555555555555
...
```

### Observed behavior

With a maximum buffer size of 128 bytes, I observe corrupted console output such as the following:

```
...
6666666666666666666666666666666666666666666666666666666666666666666666666666
77777777777777777777777777777777777777777777777777777777777777777777777777777
888888888888888888888888888888888888888888888888888888888888888888888888888888
9999999999999999999999999999999999999999999999999999999999999999999999999999999
00000000000000000000000000000000000000000000000000000000000000000000000000000000
111111111111111111111111111111111111111111111111111111111111111111111111111111111
2222222222222222222222222222222222222222222222222222222222222222222222222222222222
3333333333333333333333333333333333333334444444444
333333333333333333333333333333333
4444444444444445555555555555555555555555555555555555555555555
4444444444444444444444
5566666666666666666666666666666666666666666666666666666666666666666666666
7777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777
88888888888888888888888888888888888888888888888888888888888888888888888888888888888888888888888888
...
```

When the maximum buffer size is 64 bytes however, there's no such corrupted output.

### Analysis

Incidentally, the console driver has buffers of 64 bytes, which leads me to believe that the bug could come from there.

https://github.com/tock/tock/blob/527afb41687da5b53b8e65c6033dba23d3b24c13/capsules/src/console.rs#L59-L60

It could also be another side-effect of alarms not working properly (see #1651).

### Help wanted

I can currently only test on nRF52840-DK via Segger RTT output, so it would be helpful to know if anyone can reproduce on another board and/or on UART output.

gendx:
### Pull Request Overview

This pull request fixes #1697 by increasing the delay in Segger RTT debugging.

I tried various parameters, and 1ms of delay seems reliable enough, even for large buffers (tested up to 1500 bytes).

Somehow, when I originally filed #1697, I don't recall that increasing this delay would work, but it seems to fix it in the current kernel.

### Testing Strategy

This pull request was tested on a nRF52840-DK board with the "console triangle" test mentioned in #1697 and located on https://github.com/gendx/libtock-rs/blob/console-check/examples/console_triangle.rs. I also tested increasing the "width" of the triangle up to 1500 bytes, and the output works properly.

I had a look at https://www.segger.com/downloads/jlink but couldn't find a reliable source of ground truth on the acceptable frequency to refresh the RTT buffers...

### TODO or Help Wanted

This change makes the console slower, especially if one only uses very small buffers. However, reliability is likely preferable over premature optimization. Better start from a working code and optimize from there than the opposite.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/segger_rtt.rs
+++ b/capsules/src/segger_rtt.rs
@@ -238,7 +238,7 @@ impl<'a, A: hil::time::Alarm<'a>> uart::Transmit<'a> for SeggerRtt<'a, A> {
                     //
                     // This heuristic interval was tested with the console capsule on a nRF52840-DK
                     // board, passing buffers up to 1500 bytes from userspace. 100 micro-seconds
-                    // was to short, even for buffers as small as 128 bytes. 1 milli-second seems to
+                    // was too short, even for buffers as small as 128 bytes. 1 milli-second seems to
                     // be reliable.
                     let interval = (1000 as u32) * <A::Frequency>::frequency() / 1000000;
                     let tics = self.alarm.now().wrapping_add(interval);
--- a/capsules/src/segger_rtt.rs
+++ b/capsules/src/segger_rtt.rs
@@ -233,9 +233,14 @@ impl<'a, A: hil::time::Alarm<'a>> uart::Transmit<'a> for SeggerRtt<'a, A> {
                     // Save the client buffer so we can pass it back with the callback.
                     self.client_buffer.replace(tx_data);
 
-                    // Start a short timer so that we get a callback and
-                    // can issue the callback to the client.
-                    let interval = (100 as u32) * <A::Frequency>::frequency() / 1000000;
+                    // Start a short timer so that we get a callback and can issue the callback to
+                    // the client.
+                    //
+                    // This heuristic interval was tested with the console capsule on a nRF52840-DK
+                    // board, passing buffers up to 1500 bytes from userspace. 100 micro-seconds
+                    // was to short, even for buffers as small as 128 bytes. 1 milli-second seems to
+                    // be reliable.
+                    let interval = (1000 as u32) * <A::Frequency>::frequency() / 1000000;
                     let tics = self.alarm.now().wrapping_add(interval);
                     self.alarm.set_alarm(tics);
                 })


capsules/src/segger_rtt.rs
```diff
@@ -233,9 +233,14 @@ impl<'a, A: hil::time::Alarm<'a>> uart::Transmit<'a> for SeggerRtt<'a, A> {
...
 
-                    // Start a short timer so that we get a callback and
-                    // can issue the callback to the client.
-                    let interval = (100 as u32) * <A::Frequency>::frequency() / 1000000;
+                    // Start a short timer so that we get a callback and can issue the callback to
+                    // the client.
+                    //
+                    // This heuristic interval was tested with the console capsule on a nRF52840-DK
+                    // board, passing buffers up to 1500 bytes from userspace. 100 micro-seconds
+                    // was to short, even for buffers as small as 128 bytes. 1 milli-second seems to
```
brghena, line 241:
```suggestion
                    // was too short, even for buffers as small as 128 bytes. 1 milli-second seems to
```

PR #1962, opened 2020-06-19
Use linker file to create `APP_MEMORY` slice.

### LINKED ISSUE #1703: [Tracking] Automatic APP_MEMORY buffer size
Currently, each boards defines RAM space for applications in the form of an `APP_MEMORY` array. This array is in the `.app_memory` linker section, so that the linker can lay it out after the kernel's RAM memory.

However, the size of the `APP_MEMORY` for each board has been set to "whatever worked" at some commit. Because the layout depends on the kernel's RAM usage, which itself varies depending on kernel code, notably which capsules are enabled/disabled, this can lead to two issues:

- If the kernel memory increases, there may not be enough space anymore in RAM to layout everything. This leads to errors like:
  ```
  = note: rust-lld: error: section '.sram' will not fit in region 'ram': overflowed by 8192 bytes
  ```
- If the kernel memory decreases, there may be unused memory at the end of the RAM (such as 8KB). But because of the way the MPU currently works, there is some "cascading" effect, and half of the RAM can become unavailable for apps (128KB on nRF52840-DK).

See also https://github.com/tock/tock/pull/1696#issuecomment-600508212 for more details.

---

Although the size can be manually adjusted, it's often unclear to the developer why some error happened and how to fix it, which makes the current system cumbersome.

- In case of overflow, the linker error doesn't mention the `.app_memory` section nor that `APP_MEMORY`'s size should be decreased.
- In case of underflow, the error can come undetected for a while, until an app tries to use too much memory, causing a memory fault. The mistake won't be detected if apps don't use that much memory, and a developer may think that it's a legitimate out-of-memory error. In the case of libtock-rs, RAM space is [reserved for the heap in the application's entry point](https://github.com/tock/libtock-rs/blob/4d27d242cb64b6b0a264bc5fb174b2e10f4fca20/src/entry_point/mod.rs#L119-L134), so a fault would occur early, but still it's unclear to the developer that the fix is in the kernel's `.app_memory` section.

---

This issue is here to track potential designs to make the `APP_MEMORY` automatically adjust its size to the available RAM space.

bradjc:
### Pull Request Overview

Use the linker file to set what was the `APP_MEMORY` slice. Very similar to how app flash is handled.

Also, fix Hail's RAM size.

Fixes #1703.

### Testing Strategy

Running 10+ apps on Hail to verify that it would exhaust the available memory and display an error (which it does).

I also printed out the start address and length of the new code and the existing `APP_MEMORY` buffer, and they match.

### TODO or Help Wanted

Probably makes sense to move all boards over to this?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alistair23:
Can this be done for all boards?

brghena:
Why didn't we do this ages ago? Was there something holding us back previously?

bradjc:
> Why didn't we do this ages ago? Was there something holding us back previously?

I don't think so. It does look much more obvious now, but remember not too long ago we were using just a pointer for the start of apps in flash, and not a slice.

ppannuto:
Rebased and updated to include stm32f412gdiscovery

PR #1669, opened 2020-03-06
Print a SHA-256 sum of built binaries, to check for reproducibility.

gendx:
### Pull Request Overview

This pull request adds a SHA-256 sum of the produced binary when making boards.

### Testing Strategy

This pull request was tested by running `make` on boards.

### TODO or Help Wanted

I'm not sure what to do in the Makefile when `sha256sum` is not present.

I don't think it's a requirement to build Tock, more a nice-to-have to check whether multiple developers reproduce the same binaries. So if there's no such `sha256sum` binary, computing the sum should simply be skipped.

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

PR #1967, opened 2020-06-20
Makefile.common: use system sha256sum if available

lschuermann:
### Pull Request Overview
Use the `sha256sum` binary from coreutils/busybox/... if installed.

Somehow the compilation of the custom shipped `sha256sum` binary often fails for me, chooses the wrong interpreter or does other weird stuff. Often deleting the respective `target/` directory helps, sometimes "uninstalling" rustup (I'm on Nix, so I just leave the nix-shell) helps. In any case this is annoying and since a lot of systems have this preinstalled (almost Linux distributions I'm aware of), we can avoid building our custom application.

#### Going further

There's no _technical_ reason to not do what we currently do, but building our own replacement for something available on virtually every OS's package manager - which could instead be specified as an (optional) build dependency - feels over-engineered. It makes it _our_ responsibility that the produced `sha256sum` binary is bug-free and secure and we must ensure the build process works.

Even #1669 mentioned that this [isn't mandatory for the build](https://github.com/tock/tock/pull/1669#issuecomment-596427676). For the reasons outlined above, I would propose to use the binary if present, specify it as an optional build-time dependency and only run it when it's present.

Tagging @gendx @bradjc, since they were involved #1669.

### Testing Strategy

Building a board with `sha256sum` in $PATH and excluded from $PATH.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/Makefile.common
+++ b/boards/Makefile.common
@@ -247,9 +247,17 @@ target:
 %.elf: %
 	$(Q)cp $< $@
 
-%.bin: %.elf $(TOCK_ROOT_DIRECTORY)tools/sha256sum/target/debug/sha256sum
+# Check whether the system already has a sha256sum application
+# present, if not use the custom shipped one
+ifeq (, $(shell which sha256sum))
+  %.bin: %.elf $(TOCK_ROOT_DIRECTORY)tools/sha256sum/target/debug/sha256sum
 	$(Q)$(OBJCOPY) --output-target=binary $(OBJCOPY_FLAGS) $< $@
 	$(Q)$(TOCK_ROOT_DIRECTORY)tools/sha256sum/target/debug/sha256sum $@
+else
+  %.bin: %.elf
+	$(Q)$(OBJCOPY) --output-target=binary $(OBJCOPY_FLAGS) $< $@
+	$(Q)sha256sum $@
+endif
 
 %.lst: %.elf
 	$(Q)$(OBJDUMP) $(OBJDUMP_FLAGS) $< > $@
--- a/boards/Makefile.common
+++ b/boards/Makefile.common
@@ -161,7 +161,7 @@ endif
 
 # Check whether the system already has a sha256sum application
 # present, if not use the custom shipped one
-ifeq (, $(shell which sha256sum))
+ifeq (, $(shell sha256sum --version 2>/dev/null))
   # No system sha256sum available
   SHA256SUM := $(TOCK_ROOT_DIRECTORY)tools/sha256sum/target/debug/sha256sum
 else
--- a/boards/Makefile.common
+++ b/boards/Makefile.common
@@ -163,7 +163,7 @@ endif
 # present, if not use the custom shipped one
 ifeq (, $(shell sha256sum --version 2>/dev/null))
   # No system sha256sum available
-  SHA256SUM := $(TOCK_ROOT_DIRECTORY)tools/sha256sum/target/debug/sha256sum
+  SHA256SUM := $(CARGO) run --manifest-path $(TOCK_ROOT_DIRECTORY)tools/sha256sum/Cargo.toml --
 else
   # Use system sha256sum
   SHA256SUM := sha256sum
@@ -258,7 +258,7 @@ target:
 	$(Q)cp $< $@
 
 
-%.bin: %.elf $(TOCK_ROOT_DIRECTORY)tools/sha256sum/target/debug/sha256sum
+%.bin: %.elf
 	$(Q)$(OBJCOPY) --output-target=binary $(OBJCOPY_FLAGS) $< $@
 	$(Q)$(SHA256SUM) $@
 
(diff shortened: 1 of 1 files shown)


boards/Makefile.common
```diff
@@ -247,9 +247,17 @@ target:
...
+# present, if not use the custom shipped one
+ifeq (, $(shell which sha256sum))
+  %.bin: %.elf $(TOCK_ROOT_DIRECTORY)tools/sha256sum/target/debug/sha256sum
 	$(Q)$(OBJCOPY) --output-target=binary $(OBJCOPY_FLAGS) $< $@
 	$(Q)$(TOCK_ROOT_DIRECTORY)tools/sha256sum/target/debug/sha256sum $@
+else
+  %.bin: %.elf
+	$(Q)$(OBJCOPY) --output-target=binary $(OBJCOPY_FLAGS) $< $@
+	$(Q)sha256sum $@
+endif
```
ppannuto, line 260: Instead of having the ifdef here, I think it'd be better to do what we do elsewhere through the Tock make system and have a check at the start that sets a `SHA256SUM` variable, which is then used unconditionally in the rules. It makes it a bit easier to read to separate the 'configuration' from the 'build rules'.

The other small change is a light preference for the built-in `command -v` rather than `which`; I no longer recall when and why this was a problem, but I remember that it was once.

So this would be something more like

```make
ifeq (,$(shell command -v sha256sum)
  # No system sha256 available
  SHA256SUM := ....
else
   # Use system sha256
  SHA256SUM := sha256sum
endif


...

rule:
     $(Q)$(SHA256SUM) $@
```

lschuermann: Thanks for the hints! I did actually try that approach, but must've been doing something wrong as it didn't work and the `make` errors were not particularly helpful. I'll apply the changes requested.

gendx:
The pull request that added that originally used the system sha256, but then it was noted in https://github.com/tock/tock/pull/1669#pullrequestreview-370547167 that this could break the Makefile. So putting back in the Makefile a check for whether a system binary exists could go back to square one of breaking the Makefile.

I'm also quite surprised that the current setup is broken for you, as it should be self-contained and the binary is relatively simple. What error do you get?

boards/Makefile.common
lschuermann: Thanks for the hints! I actually tried that approach at first, but must've been doing something wrong and the `make` errors were not particularly helpful. I'll apply the changes requested.

lschuermann:
> but then it was noted in [#1669 (review)](https://github.com/tock/tock/pull/1669#pullrequestreview-370547167) that this could break the Makefile

Shoot, I misread @bradjc's review as having no binary at all, instead of a different one with other options.

Nonetheless, this proposed approach shouldn't break anything, as it first uses the system `sha256sum`, but also falls back onto the custom one. In any case, I think that we could easily support the coreutils/busybox one along with the macOS one - this is only an optional indicator for reproducibility right? The current setup goes to great lengths just to have that hash printed at the end.

> I would like to understand why the build is failing

Me as well. I'm not currently aware under what circumstances this actually produces a broken binary, but I suspect it being the result of a combination of the `shell.nix` build environment together with `rustup` (which unfortunately is required for some tasks in the current Makefile setup). I'll analyze further once I next hit this error.

It's someting along the lines of `no such file or directory` when executing the binary - even though it's there, and if I remember correctly is (a) for the correct architecture and (b) has the right `ld` path set. Note that in that exact same shell I can build other programs with cargo correctly and Tock itself builds perfectly fine as well.

gendx:
> It's someting along the lines of `no such file or directory` when executing the binary - even though it's there, and if I remember correctly is (a) for the correct architecture and (b) has the right `ld` path set. Note that in that exact same shell I can build other programs with cargo correctly and Tock itself builds perfectly fine as well.

Maybe instead of invoking the binary directly as `$(TOCK_ROOT_DIRECTORY)tools/sha256sum/target/debug/sha256sum <options>`, we should do `cargo run -- <options>`, which should be independent of the platform.

lschuermann:
I changed it to work using a conditional variable.

However having a conditionally set prerequisite for a make rule gave me a hard time, so applying @gendx's recommendation of `cargo run` solved that, since it'll automatically verify that the target binary is indeed there, current, and for the right platform (maybe also ensuring the right linker path which can change on NixOS? we'll see). I'm fairly confident that this will work - still slightly objected to shipping a custom `sha256sum`, but ultimately that's fine with me.

As per @ppannuto's recommendation to use `command` instead of `which`: though `command` has a more consistent behavior, it is actually a bash builtin, and we don't seem to be using bash to run the shell commands in the Makefile. I found another approach to be working which is calling the binary and checking whether it errors. This should be very portable, probably also work in Cygwin.

I can also add checks for the `shasum -a 256` command if desired. I don't have access to a Mac, but could maybe abuse the GitHub actions for that.

gendx:
This is fine by me, although given that we have the `cargo run` based fallback anyway, I'm not convinced that we should attempt to check for a system binary at all (which is biased towards some specific OS anyway).

PR #1968, opened 2020-06-21
Stm32f4 I2C HAL

alexandruradovici:
### Pull Request Overview

This pull request adds the I2C HAL for the stm32f4xx chip.

The implementation is without DMA as the I2C overlaps with SPI and USART. The DMA HAL should be rewritten to allocate the streams on a need basis.

### Testing Strategy

This pull request was tested using an stm32f412g discovery kit.

### TODO or Help Wanted

Testing on other devices would be useful.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/stm32f4xx/src/i2c.rs
+++ b/chips/stm32f4xx/src/i2c.rs
@@ -194,7 +194,6 @@ pub struct I2C<'a> {
     slave_address: Cell<u8>,
 
     status: Cell<I2CStatus>,
-    // transfers: Cell<u8>
 }
 
 #[derive(Copy, Clone, PartialEq)]
--- a/chips/stm32f4xx/src/i2c.rs
+++ b/chips/stm32f4xx/src/i2c.rs
@@ -3,9 +3,9 @@ use core::cell::Cell;
 use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::common::registers::{register_bitfields, ReadWrite};
 use kernel::common::StaticRef;
+use kernel::hil;
 use kernel::hil::i2c::{self, Error, I2CHwMasterClient, I2CMaster};
 use kernel::ClockInterface;
-use kernel::{debug, hil};
 
 use crate::rcc;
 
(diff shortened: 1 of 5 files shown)


chips/stm32f4xx/src/i2c.rs
```diff
@@ -0,0 +1,462 @@
...
+    buffer: TakeCell<'static, [u8]>,
+    tx_position: Cell<u8>,
+    rx_position: Cell<u8>,
+    tx_len: Cell<u8>,
+    rx_len: Cell<u8>,
+
+    slave_address: Cell<u8>,
+
+    status: Cell<I2CStatus>,
+    // transfers: Cell<u8>
```
ppannuto, line 197:
```suggestion
```

?

chips/stm32f4xx/src/i2c.rs
```diff
@@ -0,0 +1,462 @@
...
+            if self.buffer.is_some() && self.tx_position.get() < self.tx_len.get() {
+                self.buffer.map(|buf| {
+                    let byte = buf[self.tx_position.get() as usize];
+                    self.registers.dr.write(DR::DR.val(byte as u32));
+                    self.tx_position.set(self.tx_position.get() + 1);
+                });
+            }
+        }
+
+        while self.registers.sr1.is_set(SR1::RXNE) {
```
ppannuto, line 298: I'm a bit confused by this loop -- does this effectively become a synchronous transfer at this point?

alexandruradovici: As far as I understand, the i2c implemented in stm32f4xx has a byte in the data register and might immediatly have another one in the shift register. The idea was to process both bytes without needing another interrupt cycle.

chips/stm32f4xx/src/i2c.rs
alexandruradovici: Deleted

chips/stm32f4xx/src/i2c.rs
ppannuto: That makes sense -- in theory then, if in some universe the I2C bus were fast enough and the core were slow enough, this would become synchronous -- but without this it would keep throwing interrupts for new data anyway.. okay, that all makes sense to me. Thanks!

PR #1973, opened 2020-06-22
Remove reborrows?

ppannuto:
### Pull Request Overview

We use a re-borrow idiom everywhere we access registers. Once upon a lifetime, when MMIO looked very different, I think this was necessary as we changed types to something that let us access the hardware. After the introduction of `StaticRef`, however, I don't think we actually need this anymore and we can simplify all of the register accesses.

There ~300 more instances of this `&*` register pattern in the kernel, so I wanted to take a temperature check on this before doing more of them.

### Testing Strategy

Compiling.

### TODO or Help Wanted

Is there are a reason to keep all of these `&*` around?

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m3/src/mpu.rs
+++ b/arch/cortex-m3/src/mpu.rs
@@ -364,27 +364,25 @@ impl kernel::mpu::MPU for MPU {
     type MpuConfig = CortexMConfig;
 
     fn clear_mpu(&self) {
-        let regs = &*self.registers;
-        regs.ctrl.write(Control::ENABLE::CLEAR);
+        self.registers.ctrl.write(Control::ENABLE::CLEAR);
     }
 
     fn enable_app_mpu(&self) {
-        let regs = &*self.registers;
-
         // Enable the MPU, disable it during HardFault/NMI handlers, and allow
         // privileged code access to all unprotected memory.
-        regs.ctrl
+        self.registers
+            .ctrl
             .write(Control::ENABLE::SET + Control::HFNMIENA::CLEAR + Control::PRIVDEFENA::SET);
     }
 
     fn disable_app_mpu(&self) {
         // The MPU is not enabled for privileged mode, so we don't have to do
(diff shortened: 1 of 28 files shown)


arch/cortex-m3/src/mpu.rs
```diff
@@ -364,22 +364,19 @@ impl kernel::mpu::MPU for MPU {
     type MpuConfig = CortexMConfig;
 
     fn enable_mpu(&self) {
-        let regs = &*self.registers;
-
```
brghena, line 368: If we don't need this line of code with the obscure `&*` everywhere, then I definitely want it gone.

brghena:
This could also be a good time to standardize whether the reference to registers is called `regs` or `registers`. Your two example files make different choices.

bradjc:
I support this.

>This could also be a good time to standardize whether the reference to registers is called regs or registers. Your two example files make different choices.

I don't, however, think we need to manage variable names at this level.

phil-levis:
> I support this.
> 
> > This could also be a good time to standardize whether the reference to registers is called regs or registers. Your two example files make different choices.
> 
> I don't, however, think we need to manage variable names at this level.

I don't think we need to manage them, but consistency is good. I.e., we're not going to reject a PR because it says `regs` instead of `registers`, but there is no good reason to have variation. However, if you have consistency, then new contributors will see it and follow it.

alevy (dismissed):

ppannuto:
I think I got all of the ones that make sense to do as part of a mass refactor. We sill have the `&*` idiom in a few places, but most of those seem necessary.

alevy (approved):

bradjc (approved):
Too much for me to look through...hopefully it is all good :)

hudson-ayers (approved):

ppannuto:
bors retry

PR #1974, opened 2020-06-23
tools: pick toolchain that has all our components

ppannuto:
### Pull Request Overview

This pull request uses rustup to automatically pick a toolchain that has all the components we need.

### Testing Strategy

Not.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/tools/update_rust_version.sh
+++ b/tools/update_rust_version.sh
@@ -4,7 +4,7 @@
 # Ask rustup to pick the latest version that will work.
 # This requires rustup >= 1.20.0.
 echo "Updating rustc to latest compatible version..."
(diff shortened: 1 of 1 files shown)


bradjc:
Why do we need to specify rustc, rust-std, and cargo, rust-docs, rustfmt, and clippy? Those are included in the default rustup profile.

ppannuto:
I felt like the goal of this script was to automate things. What's included in the default rustup profile could change over time. I believe this is the list of the components we are actually using -- even if default rustup drops them, we still need them. It doesn't hurt to re-specify them I don't think.

PR #1978, opened 2020-06-24
Makefile: Fix the temp directory creation when cloning OT ROM

alistair23:
### Pull Request Overview

Fix the temp directory creation when cloning OT ROM

### Testing Strategy

Running the CI

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -510,7 +510,7 @@ define ci_setup_qemu_opentitan
 	# Download OpenTitan image
 	@printf "Downloading OpenTitan boot rom from: 1beb08b474790d4b6c67ae5b3423e2e8dfc9e368\n"
 	@pwd=$$(pwd) && \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -492,7 +492,7 @@ define ci_setup_qemu_opentitan
 	# Download OpenTitan image
 	@printf "Downloading OpenTitan boot rom from: 1beb08b474790d4b6c67ae5b3423e2e8dfc9e368\n"
 	@pwd=$$(pwd) && \
-		temp=$$(mktemp -d)\
+		temp=$$(mktemp -d); \
```
ppannuto, line 495:
```suggestion
		temp=$$(mktemp -d) && \
```

Not that I expect `mktemp` to fail, but if it does, this should short-circuit

PR #1981, opened 2020-06-25
LowRISC: Initial commit of Pad Controller

alistair23:
### Pull Request Overview

This PR adds support for the Pad Controller to the GPIO device.

### Testing Strategy

Booting Tock on the OT FPGA.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/chips/lowrisc/src/gpio.rs
+++ b/chips/lowrisc/src/gpio.rs
@@ -1,5 +1,6 @@
 //! General Purpose Input/Output driver.
 
+use crate::padctrl;
 use kernel::common::cells::OptionalCell;
 use kernel::common::registers::{
     register_bitfields, register_structs, Field, ReadOnly, ReadWrite, WriteOnly,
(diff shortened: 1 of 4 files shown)


chips/lowrisc/src/gpio.rs
```diff
@@ -104,7 +111,7 @@ impl GpioPin {
     }
 
     pub fn handle_interrupt(&self) {
-        let regs = self.registers;
+        let regs = self.gpio_registers;
```
ppannuto, line 114: Is there a reason for this renaming (and I mean the local `regs` variable from `self.gpio_registers`; the `registers` -> `gpio_registers` makes sense)?

I see how this is sort of copying what exists elsewhere in Tock, but I think that's more an accident of legacy MMIO design than intentional, and is hopefully on the way out (see #1973). I think while this is renaming this should just remove the `regs`.

alistair23: Removed

ppannuto:
bors retry

PR #1985, opened 2020-06-26
Update Systick trait, implement it for e310x and ibex

hudson-ayers:
### Pull Request Overview

This PR updates the SysTick trait to be less specific to the ARM systick peripheral, to instead represent a System Timer abstraction that can be more effectively used by chips without this optional peripheral. It also adds an implementation of this trait for two RISC-V chips, ibex and the e310x, by virtualizing the single available hardware timer available on each such that it can be used by both the scheduler and by kernel capsules/userspace apps.

This includes renaming SysTick to SystemTimer (name subject to bikeshedding) and modifying the functions and documentation of the trait slightly. This PR also removes the reliance on a dedicated interrupt handler for system timer interrupts, instead relying on there simply being some interrupt that fires, and calling systick.expired() whenever userspace processes are interrupted to determine if there has been a timeslice expiration. This change is critical for enabling the ability to implement the system timer on top of a hardware peripheral that is shared with other timer functionality.

This PR also includes a new `VirtualSystemTimer` type which can be constructed by passing anything implementing the `Alarm` trait, allowing for `VirtualMuxAlarm`s to be used to create a systick. Notably, the design of `MuxAlarm` is such that there is still constant-time overhead on all systick operations except for handling the systick interrupt when it fires (which is O(N) where N is the number of virtual alarms). This seems like a reasonable tradeoff to me, given that systick expirations already lead to the 
higher overhead of a context switch, and a well designed system (and scheduler) should minimize the frequency of app preemption.

This PR moves opentitan and the hifive1 to use the new `VirtualSystemTimer`.

This PR also removes the reliance of the ARM systick on modifying a global static mut variable to signal timeslice expirations.

Finally, this PR also implements `SystemTimer` directly on the RISC-V mtimer, but this will only work if a different hardware timer is used to provide alarms to the kernel/userspace, and this implementation is untested given that we do not support doing so on any of our current RISC-V chips.

### Testing Strategy

This pull request (well, a slightly earlier version) was tested on HiFive1 in QEMU and appeared to work (apps still functioned, and putting a busy loop in apps led to timeslice expirations that preempted the app).

I also tested it on OpenTitan in verilator, but have run into issues where after a timeslice expirations I receive spurious interrupts. I still need to debug this to determine whether it is an issue with this PR or the tock implementation of `rv_timer()`.

I still need to test on existing ARM boards.

### TODO or Help Wanted

This pull request still needs feedback on names, and feedback on the changes to the trait. 
I learned today that @rajivr has also been working on modifying the SysTick trait, and he has taken a somewhat different approach, so it may be useful to compare and contrast these approaches.

More testing will be needed.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


chips/ibex/src/chip.rs
```diff
@@ -21,16 +22,18 @@ use crate::usbdev;
...
 
-impl Ibex {
-    pub unsafe fn new() -> Ibex {
-        Ibex {
+impl<A: 'static + Alarm<'static>> Ibex<A> {
+    pub unsafe fn new(alarm: &'static A) -> Self {
+        Self {
             userspace_kernel_boundary: SysCall::new(),
             pmp: rv32i::pmp::PMPConfig::new(4),
+            systick: kernel::VirtualSystick::new(alarm), //TODO: Make Optional
```
alistair23: Does this really need to be optional?

kernel/src/platform/system_timer.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// Sets the timer as close as possible to the given interval in
+    /// microseconds, and starts counting down. Interrupts are not enabled.
+    ///
+    /// Callers can assume at least a 24-bit wide clock. Specific timing is
+    /// dependent on the driving clock. In practice, increments of 10ms are most
+    /// accurate and values up to 400ms are valid.
+    fn start_timer(&self, us: u32);
+
+    /// Returns if there is at least `us` microseconds left
+    fn greater_than(&self, us: u32) -> bool;
```
bradjc, line 25:
```suggestion
    fn at_least_us_remaining(&self, us: u32) -> bool;
```
maybe?

kernel/src/platform/system_timer.rs
```diff
@@ -0,0 +1,139 @@
...
+///
+/// A system tick timer provides a countdown timer to enforce process scheduling
+/// quantums.  Implementations should have consistent timing while the CPU is
+/// active, but need not operate during sleep.
+///
+/// On most chips, this will be implemented by the core (e.g. the ARM core systick), but
+/// some chips lack this optional peripheral, in which case it might be
+/// implemented by another timer or alarm controller, or require virtualization
+/// on top of a single hardware timer.
+pub trait SystemTimer {
```
bradjc, line 15:
```suggestion
pub trait TimesliceAlarm {
```
Maybe? It seems we should at least be consistent with the HIL in regards to alarm vs. timer. This alarm is a bit strange in that it doesn't have a client, so it's largely not useful except for in the core kernel loop that can check for `expired()`, so something specific like "timeslice" or "scheduling" makes sense to me.

kernel/src/platform/system_timer.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// the return value is unspecified (implementations can return whatever they like)
+    fn expired(&self) -> bool;
+
+    /// Resets the timer
+    ///
+    /// Resets the timer to 0 and disables it
+    fn reset(&self);
+
+    /// Enable or disable interrupts
+    ///
```
bradjc, line 38:
```suggestion
    /// Enable or disable the alarm interrupt.
    ///
    /// This must not start or stop the alarm, but only change whether an
    /// interrupt will be generated when the alarm expires or not.
```

kernel/src/platform/system_timer.rs
```diff
@@ -0,0 +1,139 @@
...
+    fn config_interrupts(&self, enabled: bool);
+}
```
bradjc, line 40:
```suggestion

    /// Return the number of microseconds remaining before the alarm expires.
    fn get_remaining_us(&self) -> u32;
}
```

You need that for the scheduler updates, right?

kernel/src/platform/system_timer.rs
ppannuto: I agree with Brad here. I would go further and say that we should try to use the word `alarm` throughout the comments, documentation, and filenames as well for this, as in Tock semantics, this is acting as an `alarm` (never mind what ARM happened to call it).

kernel/src/platform/system_timer.rs
```diff
@@ -0,0 +1,139 @@
...
+/// some chips lack this optional peripheral, in which case it might be
+/// implemented by another timer or alarm controller, or require virtualization
+/// on top of a single hardware timer.
+pub trait SystemTimer {
+    /// Sets the timer as close as possible to the given interval in
+    /// microseconds, and starts counting down. Interrupts are not enabled.
+    ///
+    /// Callers can assume at least a 24-bit wide clock. Specific timing is
+    /// dependent on the driving clock. In practice, increments of 10ms are most
+    /// accurate and values up to 400ms are valid.
```
ppannuto, line 21: Are these comments universally true, or a holdover of the ARM implementation?

chips/ibex/src/chip.rs
hudson-ayers: It does not, the comment has been removed.

kernel/src/platform/system_timer.rs
hudson-ayers: From `kernel/src/hil/time.rs`:
```rust
/// The `Alarm` trait models a wrapping counter capable of notifying when the
/// counter reaches a certain value.

//...

/// The `Timer` trait models a timer that can notify when a particular interval
/// has elapsed.
```

Given that the exposed interface is "fire an interrupt x us in the future" I think that it is in fact a timer according to those definitions?

kernel/src/platform/system_timer.rs
hudson-ayers: These are a holdover from the ARM implementation. It might be reasonable to add a function to the trait to provide the minimum valid interval. I think it makes sense to leave the maximum interval as 400 ms -- RISC-V has no limitations on the maximum systick, but I think keeping schedulers platform independent is preferable and I don't imagine many use cases requiring timeslices that long.

kernel/src/platform/system_timer.rs
hudson-ayers: yes, added.

kernel/src/platform/system_timer.rs
hudson-ayers: done

kernel/src/platform/system_timer.rs
hudson-ayers: done

kernel/src/platform/system_timer.rs
hudson-ayers: I update the comment.

kernel/src/platform/system_timer.rs
bradjc: Haha well I'm not going to argue that Tock isn't stingy with documentation. But that should read "every time a particular interval has elapsed" (I'm pretty sure). It would be nice to see the updated time HIL to confirm.

kernel/src/process.rs
```diff
@@ -1263,7 +1264,7 @@ impl<C: Chip> ProcessType for Process<'_, C> {
             return None;
```
bradjc, line 1264: Hmm...I think we should change the return type of `switch_to`. You're right that the arch crate can't differentiate between a timeslice expiration and any other interrupt, and therefore we shouldn't include the systick expired option in `ContextSwitchReason`. But the kernel still needs to know, so we need this to return something like `Option<ProcessStoppedReason>`.

kernel/src/platform/system_timer.rs
hudson-ayers: The trait currently specifies oneshot and repeating timers. 

```rust
/// The `Timer` trait models a timer that can notify when a particular interval
/// has elapsed.
pub trait Timer<'a, W = u32>: Time<W> {
    /// Set the client for interrupt events.
    fn set_client(&'a self, client: &'a dyn TimerClient);

    /// Sets a one-shot timer to fire in `interval` clock-tics.
    ///
    /// Calling this method will override any existing oneshot or repeating timer.
    fn oneshot(&self, interval: W);

    /// Sets repeating timer to fire every `interval` clock-tics.
    ///
    /// Calling this method will override any existing oneshot or repeating timer.
    fn repeat(&self, interval: W);

    /// Returns the interval for a repeating timer.
    ///
    /// Returns `None` if the timer is disabled or in oneshot mode and `Some(interval)` if it is
    /// repeating.
    fn interval(&self) -> Option<W>;
```

cc @phil-levis thoughts on if this should be called a timer or an alarm?

phil-levis: Timers allow you to precisely control the interval, but not the "when" (because there may be latency before the interval is put into a register). In that way, this looks more like a timer.  I'd suggest making its method names more closely aligned to timer. E.g., we probably don't want to say "interrupts" in the trait.

Can someone explain the division between HIL traits and platform traits? Should this perhaps be in the time HIL?

hudson-ayers: The watchdog trait is currently in HIL which seems inconsistent with this one being in platform, but I am not sure which should move.

Also, I am not sure what to say other than interrupts..that function call is used only to affect whether an interrupt will fire when the timer hits 0. just having `enable()` and `disable()` would be misleading, because it seems like `disable()` would stop the timer, but all we want it to do in this case is make sure no interrupt fires when the timer expires, without otherwise changing the state of the timer.

I am open to name suggestions though. I could replace `config_interrupts(enabled: bool)` with `enable_interrupts()` and `disable_interrupts()`?

kernel/src/process.rs
hudson-ayers: After some thought about my changes in #1767, it probably makes more sense for `do_process()` to check for timeslice expirations, so that if a scheduler called `do_process(no_systick)`, checking for timeslice expirations could be avoided. I updated the code so that `ContextSwitchReason` no longer has `TimesliceExpiration` as a variant, and instead check for expiration in the `Interrupts` match arm.

arch/rv32i/src/machine_timer.rs
```diff
@@ -71,6 +72,8 @@ impl<'a> hil::time::Alarm<'a> for MachineTimer<'a> {
     }
 
     fn set_alarm(&self, tics: u32) {
+        // TODO: this may generate spurious interrupts due to it taking 2 instructions to write 64
+        // bit registers, see RISC-V Priveleged ISA
```
alistair23, line 76: Should we disable timer interrupts before setting this then?

kernel/src/platform/system_timer.rs
hudson-ayers: I spoke with Phil about a name for `config_interrupts()` today, we settled on `arm()` and `disarm()`.

I am leaning towards `SchedulerTimer` instead of `SystemTimer` but will get final feedback on the call tomorrow before changing all the instances

kernel/src/platform/system_timer.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// but may disable the underlying interrupt if one has been set,
+    /// preventing overhead from the timer firing after an executing application
+    /// has returned to the kernel.
+    fn disarm(&self);
+
+    /// Arm the underlying timer. This does not start the timer,
+    /// it just guarantees that an interrupt will be generated
+    /// if an already started timer expires, which is useful for
+    /// preempting userspace applications.
+    fn arm(&self);
```
bradjc, line 50: Why not `enable_interrupt()`?

kernel/src/platform/system_timer.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// Resets the timer
+    ///
+    /// Resets the timer to 0 and disables it
+    fn reset(&self);
+
+    /// Disarm the underlying timer. This does not stop the timer,
+    /// but may disable the underlying interrupt if one has been set,
+    /// preventing overhead from the timer firing after an executing application
+    /// has returned to the kernel.
+    fn disarm(&self);
```
bradjc, line 44:
```suggestion
    fn disable_interrupt(&self);
```

kernel/src/platform/system_timer.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    /// Resets the timer
+    ///
+    /// Resets the timer to 0 and disables it
+    fn reset(&self);
+
+    /// Disarm the underlying timer. This does not stop the timer,
+    /// but may disable the underlying interrupt if one has been set,
+    /// preventing overhead from the timer firing after an executing application
+    /// has returned to the kernel.
```
bradjc, line 43:
```suggestion
    /// Disable the interrupt for the underlying timer. This does nothing if the
    /// interrupt was not enabled.
    ///
    /// This is useful for preventing overhead from the timer triggering an interrupt
    /// after an executing process has returned to the kernel.
```

kernel/src/platform/system_timer.rs
hudson-ayers: @phil-levis didn't like the name `enable_interrupt()` because it might imply that all interrupts for the underlying hardware timer are disabled, which is not the case if the trait is on top of a virtualized timer.

kernel/src/platform/system_timer.rs
hudson-ayers: Phil wanted this interface to leave flexibility to the implementation about whether the underlying interrupts is actually disabled, because in certain scenarios where this trait sits on top of a virtualized alarm, if the interrupt is going to occur too soon disabling the interrupt may be problematic for other clients of the virtual alarm, so he said the interface should not guarantee that the alarm will not fire.

kernel/src/platform/system_timer.rs
bradjc: Ok overall I think the symmetry in arm/disarm probably makes for the most clear interface, since disarm doesn't _have_ to actually disable anything. I think just capturing the importance of this interrupt in the documentation better to explain this subtlety will avoid any questions along this line in the future. I can add a commit.

arch/rv32i/src/machine_timer.rs
hudson-ayers: The correct approach involves setting the upper register of the timer to the max value before setting the lower register. I added a comment to describe this, but think that change belongs in its own PR.

kernel/src/platform/system_timer.rs
hudson-ayers: renamed to `SchedulerTimer`, marking this as resolved.

hudson-ayers:
Merge conflicts resolved, and I think all comments have been addressed.

The coupling between the `VirtualSchedulerTimer` and the implementation of `VirtualMuxAlarm` is a little strong still, but I would prefer to resolve that when the new time HIL arrives.

bradjc:
I added the doc commit I promised. I also really like the name.

My last two questions:

- Why `start_timer()` instead of `start()`? start/reset seems like a more obvious pair. Also, now that it is called `SchedulerTimer`, `start()` should be pretty obvious it starts the timer.
- Why both us_remaining() and get_remaining_us()?

hudson-ayers:
> My last two questions:
> 
> * Why `start_timer()` instead of `start()`? start/reset seems like a more obvious pair. Also, now that it is called `SchedulerTimer`, `start()` should be pretty obvious it starts the timer.

Agreed, changed.

> * Why both us_remaining() and get_remaining_us()?

The reason I kept `at_least_us_remaining()` was because I believed that it was possible for such an implementation to be more efficient -- in theory if we only ever call this function with a single value, the compiler should be able to precompute the us --> ticks conversion for that constant, so at runtime there is only a single register read followed by a comparison, while calling `get_remaining_us()` will always require converting the remaining tics to microseconds, which requires 64 bit multiplication and division. However I just benchmarked both approaches and found that the constant propagation I was expecting for `at_least_us_remaining()` does not happen, and in fact that approach takes more cycles (47, as compared to 22 for using `get_remaining_us() < MIN_QUANTA_THRESHOLD`), so I went ahead and removed `at_least_us_remaining()`

hudson-ayers:
The updated version has now been tested on ARM by checking that apps are preempted as expected and that these preemptions are counted by the process console.

I will try to re-test on RISC-V before this is merged.

kernel/src/platform/scheduler_timer.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// `set_timer()` is called again (to start a new timeslice). If `expired()`
+    /// is called again after returning `true` without an intervening call to
+    /// `set_timer()`, the return the return value is unspecified and
+    /// implementations may return whatever they like.
+    ///
+    /// The requirement that this may not be called again after it returns
+    /// `true` simplifies implementation on hardware platforms where the
+    /// hardware automatically clears the expired flag on a read, as with the
+    /// ARM SysTick peripheral.
+    fn expired(&self) -> bool;
```
ppannuto, line 104: I might bikeshed this to `is_expired`; the return type is clear, but the name sounds closed to the use of a function that would be called on the implementing object when the timer expires

bradjc: I could see `has_expired()`, since it can only be called once. `expired()` is good too though.

hudson-ayers:
rebased on master, changed `expired()` --> `has_expired()`. I tested opentitan on verilator and verified that a long running app was preempted by the scheduler timer, then resumed and finished. The timing felt a little weird to me, as it seemed that it probably should have been preempted multiple times, but I don't fully understand how verilator works so I am gonna chalk that up to weirdness with the emulated timer frequency.

I think this is ready to go.

PR #1986, opened 2020-06-28
Touch Panel HIL and ft6x06 driver

alexandruradovici:
### Pull Request Overview

This pull request adds:

* a HIL for touch panels with optional multi touch and gestures
* a driver for the ft6x06 touch panel

### Testing Strategy

This pull request was tested using the STM32F412g Discovery Kit.

### TODO or Help Wanted

This pull request still needs some feedback. 

Even tough the documentation for [ft6x06](http://www.tvielectronics.com/ocart/download/controller/FT6206.pdf) states that gestures should be recognized, my device does not seem to recognize them. I might be doing something wrong, any ideas would be helpful. The test setup is https://github.com/UPBIoT/tock/tree/touch.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,83 @@
+//! Interface for touch input devices
+
+use crate::ReturnCode;
+
+#[derive(Debug, Copy, Clone)]
+pub enum TouchStatus {
```
ppannuto, line 6: We generally require doc comments (`///`) for every `pub` entity and interface in HILs, even for the seemingly obvious stuff, so that the docs (https://docs.tockos.org/kernel/hil/index.html) render nicely.

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,83 @@
...
+    // touch weight
+    pub weight: Option<usize>,
+}
+
+pub trait Touch {
+    // enable touch
+    fn enable(&self) -> ReturnCode;
+
+    // disable touch
+    fn disable(&self) -> ReturnCode;
```
ppannuto, line 43: Is the purpose of enable/disable to start/stop events?

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,83 @@
...
+    // enable touches
+    fn enable(&self) -> ReturnCode;
+
+    // disable touches
+    fn disable(&self) -> ReturnCode;
+
+    /// Returns the number of concurently supported touches
+    /// This function must be called in the same interrupt
+    /// as the event, otherwise data might not be available.
+    fn get_num_touches(&self) -> usize;
```
ppannuto, line 58: This time-dependent bit makes me a bit nervous. Is it expensive to capture the number of touches, or is this something that could just be always included as part of the event?

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    /// Returns the touch event at index or None
+    /// This function must be called in the same interrupt
+    /// as the event, otherwise data might not be available.
+    fn get_touch(&self, index: usize) -> Option<TouchEvent>;
```
ppannuto, line 63: Same comment here. It's also a bit confusing that the single `Touch` will return the information in the event, but this needs the extra lookup.

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,83 @@
...
+    fn set_client(&self, multi_touch_client: &'static dyn MultiTouchClient);
+}
+
+pub trait TouchClient {
+    fn touch_event(&self, touch_event: TouchEvent);
+}
+
+pub trait MultiTouchClient {
+    /// num touches represents the number of touches detected
+    fn touch_event(&self, num_touches: usize);
```
ppannuto, line 74: Ahhh... is the problem that this interface would be an array of unknown size if it tried to report all touch events?

Maybe the right approach is to make an associated type for the MultiTouch HIL that is an array of touches, where the length is specified by the instatiator of the HIL?

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,83 @@
...
+    Pressed,
+    Released,
+}
+
+#[derive(Debug, Copy, Clone)]
+pub enum GestureEvent {
+    MoveUp,
+    MoveDown,
+    MoveLeft,
+    MoveRight,
```
ppannuto, line 16: `Move` is maybe ambiguous here (what is moving, the finger, the object being clicked on?) -- perhaps `SwipeUp` etc as "Swipe" is more clearly a gesture?

alexandruradovici: Good point.

kernel/src/hil/touch.rs
alexandruradovici: It enables and disables the touch panel, this is useful for low power

kernel/src/hil/touch.rs
alexandruradovici: I changed this

kernel/src/hil/touch.rs
alexandruradovici: Done

kernel/src/hil/touch.rs
alexandruradovici: Done, the length is specified in the actual touch panel driver, the hardware has a touch limit usually.

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// Enable the touche panel
+    fn enable(&self) -> ReturnCode;
+
+    /// Disable the touch panel
+    fn disable(&self) -> ReturnCode;
+
+    /// Returns the number of concurently supported touches
+    /// This function must be called in the same interrupt
+    /// as the event, otherwise data might not be available.
+    fn get_num_touches(&self) -> usize;
```
bradjc, line 64: I'm not sure I follow this API. Will the number of concurrent touches supported change? What is the event? Doesn't the event report how many touches there were?

capsules/src/touch.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+impl<'a> hil::touch::TouchClient for Touch<'a> {
+    fn touch_event(&self, mut event: TouchEvent) {
+        // update rotation if there is a screen attached
+        self.update_rotation(&mut event);
+        debug!(
+            "touch {:?} x {} y {} area {:?} weight {:?}",
+            event.status, event.x, event.y, event.area, event.weight
+        );
```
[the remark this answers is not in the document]
alexandruradovici: No, they are just for debugging, will delete them once everyone agrees with the API.

capsules/src/touch.rs
```diff
@@ -0,0 +1,326 @@
...
+                    ReturnCode::ENODEVICE
+                }
+            }
+            // Touch Disable
+            2 => {
+                if let Some(touch) = self.touch {
+                    touch.disable()
+                } else if let Some(multi_touch) = self.multi_touch {
+                    multi_touch.disable()
+                } else {
```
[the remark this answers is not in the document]
alexandruradovici: You have a point here, this was for allowing apps to really disable it. What about allowing this if there are no subscriptions?

capsules/src/touch.rs
```diff
@@ -0,0 +1,326 @@
...
+/// Syscall driver number.
+use crate::driver;
+pub const DRIVER_NUM: usize = driver::NUM::Touch as usize;
+
+pub struct App {
+    touch_callback: Option<Callback>,
+    gesture_callback: Option<Callback>,
+    multi_touch_callback: Option<Callback>,
+    events_buffer: Option<AppSlice<Shared, u8>>,
+    ack: bool,
```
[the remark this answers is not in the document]
alexandruradovici: This avoids a read write race. Apps will get the scheduled callback, but they might be in the middle of it when the touch screen sends a new set of touches. The data would be overwritten by the capsule right when the app reads it. To avoid this, I will add a command so that apps can ack that they have read the data.

Am I maybe wrong about how callbacks work and the ack is not necessary?

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,97 @@
...
+/// A single touch event's data
+#[derive(Copy, Clone)]
+pub struct TouchEvent {
+    pub status: TouchStatus,
+    /// touch (x, y) position
+    pub x: usize,
+    pub y: usize,
+
+    /// touch id, value defined by the driver
+    pub id: usize,
```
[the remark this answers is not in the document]
alexandruradovici: touches might not be in order, the id allows an app to identify the same touch. When the touch screen detects one touch, it assigns an id and returns it. Whenever another touch is detected, another id is generated. When only one of these touches is moved, only that moving touch will be reported. This id allows apps to distinguish between actual touches.

I will test this a little more, I am wring the user land library.

kernel/src/hil/touch.rs
alexandruradovici: Each touch screen might support a different number of touches, I expect this function to be constant. Maybe add const fn?

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    /// touch area
+    pub area: Option<usize>,
```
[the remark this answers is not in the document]
alexandruradovici: I think it should be the diameter or the radius of the touch. This is not precisely defined for ft6x06 and seems to always return 0.

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    /// touch weight
+    pub weight: Option<usize>,
```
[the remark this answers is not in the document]
alexandruradovici: Same as area, this is not precisely defined for ft6x06 and seems to always return 0.

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,98 @@
...
+pub trait MultiTouch<'a> {
+    /// Enable the touche panel
+    fn enable(&self) -> ReturnCode;
+
+    /// Disable the touch panel
+    fn disable(&self) -> ReturnCode;
+
+    /// Returns the number of maximum concurently supported touches.
+    /// This function must be called in the same interrupt
+    /// as the event, otherwise data might not be available.
```
bradjc, line 63:
```suggestion
```

capsules/src/touch.rs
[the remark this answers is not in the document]
bradjc: I think there should be a way to do that, but it shouldn't return an error if it can't disable anything. That would tell the app that another app is using the touchscreen, and leak that information. I think it would overall be easier to just disable the screen automatically if nothing is using it, and not have this API.

capsules/src/touch.rs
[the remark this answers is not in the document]
bradjc: No you're right, that is something that has to be considered.

kernel/src/hil/touch.rs
alevy: This is not necessarily a suggestion, but the way we handle something similar to this in the time HIL is by using an associated type to denote the clock frequency. The associated type itself has only a static method (i.e., one that doesn't take a `self`) and therefore, presumably, is easier for the compiler to resolve statically and inline the constant.

But... I don't know that we ever evaluated that it actually makes a difference, and I don't know if others agree that it's a better interface than just having an instance method in the trait.

alexandruradovici: I get your point, but it does add some complexity to the code and seems to make it more difficult to read.

kernel/src/hil/touch.rs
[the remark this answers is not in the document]
ppannuto: Given that area (and touch) seem a little underspecified and perhaps unimplemented by the ft6x06, perhaps it makes sense to omit them from the HIL for now?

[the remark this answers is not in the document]
bradjc: I think we should just define these, since these are features on modern touchscreens.

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub status: TouchStatus,
+    /// touch (x, y) position
+    pub x: usize,
+    pub y: usize,
+
+    /// touch id, value defined by the driver
+    pub id: usize,
+
+    /// touch area
+    pub area: Option<usize>,
```
bradjc, line 35:
```suggestion
    /// Optional scaled value for the size of the touch. A larger value
    /// corresponds to a "fatter" touch. The size values range from 0
    /// to 65535.
    ///
    /// If a touchscreen does not provide information about the size of the touch,
    /// this must be set to `None`.
    pub size: Option<usize>,
```

I'm copying from the android API largely.

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    /// touch weight
+    pub weight: Option<usize>,
```
bradjc, line 38:
```suggestion
    /// Optional scaled value for the pressure of the touch. A larger value
    /// corresponds to a "firmer" press. The pressure values range from 0
    /// to 65536.
    ///
    /// If a touchscreen does not provide information about the pressure of a touch,
    /// this must be set to `None`.
    pub pressure: Option<usize>,
```

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,98 @@
...
+
+/// A single touch event's data
+#[derive(Copy, Clone)]
+pub struct TouchEvent {
+    pub status: TouchStatus,
+    /// touch (x, y) position
+    pub x: usize,
+    pub y: usize,
+
+    /// touch id, value defined by the driver
```
bradjc, line 31:
```suggestion
    /// Numeric ID assigned to this touch. This ID allows the client to
    /// to match different `TouchEvent`s to the same physical touch.
    ///
    /// The driver must assign a unique ID to each touch.
```

capsules/src/touch.rs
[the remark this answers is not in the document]
alexandruradovici: I commented them for now

capsules/src/touch.rs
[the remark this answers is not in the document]
alexandruradovici: It is now enabled and disabled automatically

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// The driver must assign a unique ID to each touch.
+    pub id: usize,
+
+    /// Optional scaled value for the size of the touch. A larger value
+    /// corresponds to a "fatter" touch. The size values range from 0
+    /// to 65535.
+    ///
+    /// If a touchscreen does not provide information about the size of the touch,
+    /// this must be set to `None`.
+    pub size: Option<usize>,
```
ppannuto, line 43: If the range is 0..65535, should this be a `u16` rather than a `usize`?

ppannuto: Separately: I'm mildly concerned that we're encoding something specific to one touch driver in a general-purpose HIL here. I think this is okay for today; we have changed HILs before, and this is a reasonable enough range for drivers to adapt to.

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    /// Optional scaled value for the pressure of the touch. A larger value
+    /// corresponds to a "firmer" press. The pressure values range from 0
+    /// to 65536.
+    ///
+    /// If a touchscreen does not provide information about the pressure of a touch,
+    /// this must be set to `None`.
+    pub pressure: Option<usize>,
```
ppannuto, line 51: Also `u16` here?

kernel/src/hil/touch.rs
```diff
@@ -0,0 +1,117 @@
...
+/// Single touch panels should implement this
+pub trait Touch<'a> {
+    /// Enable the touch panel
+    ///
+    /// returns SUCCESS even if device is already enabled
+    fn enable(&self) -> ReturnCode;
+
+    /// Disable the touch panel
+    ///
+    /// /// returns SUCCESS even if device is already disabled
```
ppannuto, line 63:
```suggestion
    /// returns SUCCESS even if device is already disabled
```

kernel/src/hil/touch.rs
alexandruradovici: I agree with u16, I don't think there will ever be a screen that large

kernel/src/hil/touch.rs
alexandruradovici: Agree

PR #1918, opened 2020-06-08
Add Cortex-M7 Architecture and i.MX RT 1052 EVKB Board

valexandru:
### Pull Request Overview

This pull request adds support for the NXP i.MX RT 1052 Crossover MCU and i.MX RT 1052 EVKB Board.

Supported features so far are:

UART
GPIO
Leds
I2C
Timer

This also adds support for the Cortex M7 MCU.

This also adds the fxos8700 component.

### Testing Strategy

This pull request was tested using the i.MX RT 1052 EVKB Board.

### TODO or Help Wanted

This pull request still uses a binary FCB Header as header creation is still a problem. Looking into using https://github.com/imxrt-rs/imxrt-boot-gen . Would this be a problem with the licensing?

Added clock frequency Freq2475MHz for the General Purpose Timer.

### Documentation Updated

- [x] Updated the relevant files in `/docs`.

### Formatting

- [x] Ran `make prepush`.

### New Platform Checklist

- [x] Hardware is widely available.
- [x] I can support the platform, at least initially.
- Basic features are implemented:
  - [x] `Console`, including `debug!()` and userspace `printf()`.
  - [x] Timers.
  - [ ] GPIO with interrupts.

PR #1987, opened 2020-06-28
Added support for external clock source and systick user-defined clock value for ARM

valexandru:
### Pull Request Overview

This pull request adds support for setting external clock source for systick
and extends the support for setting a user-defined clock speed for the 
systick.

This pull request is needed for [#1918](https://github.com/tock/tock/pull/1918) to work because for the i.MX RT 1052 EVKB 
board, even though syst_calib.SKEW = 0 and syst.calib.NOREF = 0, the 
calibration value is not correct and causes thesystick to overflow too fast. 

Also, if I select external clock source in syst_csr, which causes an external 
24 MHz clock source to be prescaled to 100KHz and used as systick, the 
system will work.

### Testing Strategy

This pull request was tested using the i.MX RT 1052 EVKB Board.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/systick.rs
+++ b/arch/cortex-m/src/systick.rs
@@ -1,6 +1,6 @@
 //! ARM Cortex-M SysTick peripheral.
 
-use kernel::common::registers::{register_bitfields, ReadOnly, ReadWrite};
+use kernel::common::registers::{register_bitfields, ReadOnly, ReadWrite, FieldValue};
 use kernel::common::StaticRef;
 
 #[repr(C)]
@@ -53,6 +53,7 @@ register_bitfields![u32,
 /// Documented in the Cortex-MX Devices Generic User Guide, Chapter 4.4
 pub struct SysTick {
     hertz: u32,
+    external_clock: bool
 }
 
 const BASE_ADDR: *const SystickRegisters = 0xE000E010 as *const SystickRegisters;
@@ -65,7 +66,7 @@ impl SysTick {
     /// Use this constructor if the core implementation has a pre-calibration
     /// value in hardware.
     pub unsafe fn new() -> SysTick {
-        SysTick { hertz: 0 }
+        SysTick { hertz: 0, external_clock: false }
     }
 
     /// Initialize the `SysTick` with an explicit clock speed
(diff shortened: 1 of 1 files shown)


arch/cortex-m/src/systick.rs
```diff
@@ -143,16 +163,21 @@ impl kernel::SysTick for SysTick {
     }
 
     fn enable(&self, with_interrupt: bool) {
+        let clock_source: FieldValue<u32, self::ControlAndStatus::Register> = if self.external_clock
+        {
+            ControlAndStatus::CLKSOURCE::CLEAR
+        } else {
+            ControlAndStatus::CLKSOURCE::SET
+        };
```
ppannuto, line 171:
```suggestion
        {
            // CLKSOURCE 0 --> external clock
            ControlAndStatus::CLKSOURCE::CLEAR
        } else {
            // CLKSOURCE 1 --> internal clock
            ControlAndStatus::CLKSOURCE::SET
        };
```

arch/cortex-m/src/systick.rs
```diff
@@ -81,12 +85,28 @@ impl SysTick {
...
+
     // Return the tic frequency in hertz. If the calibration value is set in
     // hardware, use `self.hertz`, which is set in the `new_with_calibration`
-    // constructor.
+    // constructor. However, if there is value configured by the user, choose
+    //`self.hertz` instead.
     fn hertz(&self) -> u32 {
         let tenms = SYSTICK_BASE.syst_calib.read(CalibrationValue::TENMS);
-        if tenms == 0 {
+        if tenms == 0 || self.hertz != 0 {
```
ppannuto, line 109: Something about the way the is commented / implemented is confusing. Given the comments, could this be simplified to just:

```rust
fn hertz(&self) -> u32 {
    if self.hertz != 0 {
         self.hertz
    } else {
         let tenms = SYSTICK_BASE.syst_calib.read(CalibrationValue::TENMS);
         tenms * 100
    }
}
```

?

valexandru: Yes and it is far more clear this way. I also updated the comment to reflect the implementation.

valexandru:
> Though it would lead to some duplicated code, I would lean towards just creating a new `ExternalSystick` type and reimplementing the `SysTick` trait for it.
> 
> Many of these functions are called on every single iteration of the scheduler, so it seems a shame to have to make these decisions dynamically on every call to `enable()` and `hertz()`

The hertz() function was not significantly changed. If a value was configured in the chip, self.hertz would be different than 0, otherwise everything remains the same. Would there be a problem for checking the value for external_clock in `enable()`?

PR #1988, opened 2020-06-29
chips: apollo3: Allow a few tries to set the stimer

alistair23:
### Pull Request Overview

Apparently the stimer isn't always set on the first time (a chip errata) so let's take a few goes.

### Testing Strategy

Running on the board

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/chips/apollo3/src/stimer.rs
+++ b/chips/apollo3/src/stimer.rs
@@ -133,8 +133,15 @@ impl<'a> hil::time::Alarm<'a> for STimer<'a> {
         // Enable interrupts
         regs.stminten.modify(STMINT::COMPAREA::SET);
 
-        // Set the delta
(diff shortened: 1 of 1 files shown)


chips/apollo3/src/stimer.rs
```diff
@@ -133,8 +133,14 @@ impl<'a> hil::time::Alarm<'a> for STimer<'a> {
         // Enable interrupts
         regs.stminten.modify(STMINT::COMPAREA::SET);
 
-        // Set the delta
-        regs.scmpr[0].set(tics - regs.sttmr.get());
+        // Set the delta, this can take a few goes
+        let timer_delta = tics - regs.sttmr.get();
+        let mut tries = 0;
+
```
ppannuto, line 139: Can you add a pointer to the errata that we're working around here?

PR #1990, opened 2020-06-29
[RFC] arch/cortex-{m3,m4}: Disable SysTick in trap handler

rajivr:
As I mentioned [here](https://github.com/tock/tock/pull/1767#discussion_r442563432) there is a possibility that we might not accurately capture the time spent in the user-space.

Here is a bit of assembly to disable SysTick in `systick_handler`, `generic_isr` and `svc_handler`. This allows us to accurately know how much time was used by the user process.

### DIFF
--- a/arch/cortex-m3/src/lib.rs
+++ b/arch/cortex-m3/src/lib.rs
@@ -40,6 +40,14 @@ pub unsafe extern "C" fn systick_handler() {
 pub unsafe extern "C" fn systick_handler() {
     llvm_asm!(
         "
+    // Clear SYSTICK_CTRL.ENABLE[0]. This will stop the SysTick
+    // counter, so we can accurately know how much time was used by the
+    // user process.
+    ldr r0, =0xE000E010
+    ldr r1, [r0]
+    bic r1, #1
+    str r1, [r0]
+
     /* Mark that the systick handler was called meaning that the process */
     /* stopped executing because it has exceeded its timeslice. */
     ldr r0, =SYSTICK_EXPIRED
@@ -70,6 +78,14 @@ pub unsafe extern "C" fn generic_isr() {
 pub unsafe extern "C" fn generic_isr() {
     llvm_asm!(
         "
+    // Clear SYSTICK_CTRL.ENABLE[0]. This will stop the SysTick
+    // counter, so we can accurately know how much time was used by the
+    // user process.
+    ldr r0, =0xE000E010
+    ldr r1, [r0]
+    bic r1, #1
+    str r1, [r0]
+
     /* Skip saving process state if not coming from user-space */
     cmp lr, #0xfffffffd
     bne _ggeneric_isr_no_stacking
@@ -162,6 +178,14 @@ pub unsafe extern "C" fn svc_handler() {
     movt lr, #0xffff
     bx lr
   to_kernel:
+    // Clear SYSTICK_CTRL.ENABLE[0]. This will stop the SysTick
+    // counter, so we can accurately know how much time was used by the
+    // user process.
+    ldr r0, =0xE000E010
+    ldr r1, [r0]
+    bic r1, #1
+    str r1, [r0]
+
     ldr r0, =SYSCALL_FIRED
     mov r1, #1
     str r1, [r0, #0]
--- a/arch/cortex-m4/src/lib.rs
+++ b/arch/cortex-m4/src/lib.rs
@@ -43,6 +43,14 @@ pub unsafe extern "C" fn systick_handler() {
 pub unsafe extern "C" fn systick_handler() {
     llvm_asm!(
         "
+    // Clear SYSTICK_CTRL.ENABLE[0]. This will stop the SysTick
+    // counter, so we can accurately know how much time was used by the
+    // user process.
+    ldr r0, =0xE000E010
+    ldr r1, [r0]
+    bic r1, #1
+    str r1, [r0]
+
     // Mark that the systick handler was called meaning that the process stopped
     // executing because it has exceeded its timeslice. This is a global
     // variable that the `UserspaceKernelBoundary` code uses to determine why
@@ -86,6 +94,14 @@ pub unsafe extern "C" fn generic_isr() {
 pub unsafe extern "C" fn generic_isr() {
     llvm_asm!(
         "
+    // Clear SYSTICK_CTRL.ENABLE[0]. This will stop the SysTick
+    // counter, so we can accurately know how much time was used by the
+    // user process.
+    ldr r0, =0xE000E010
+    ldr r1, [r0]
+    bic r1, #1
+    str r1, [r0]
+
     // Set thread mode to privileged to ensure we are executing as the kernel.
     // This may be redundant if the interrupt happened while the kernel code
     // was executing.
@@ -215,6 +231,14 @@ pub unsafe extern "C" fn svc_handler() {
     bx lr
 
   to_kernel:
+    // Clear SYSTICK_CTRL.ENABLE[0]. This will stop the SysTick
+    // counter, so we can accurately know how much time was used by the
+    // user process.
+    ldr r0, =0xE000E010
+    ldr r1, [r0]
+    bic r1, #1
+    str r1, [r0]
+
     // An application called a syscall. We mark this in the global variable
     // `SYSCALL_FIRED` which is stored in the syscall file.
     // `UserspaceKernelBoundary` will use this variable to decide why the app


ppannuto:
Is this intended to work with #1985, or is it independent?

rajivr:
@hudson-ayers I was wondering if you happen to have a PR in progress that combines #1985 with #1767?

Without stopping the SysTick timer, I am not entirely sure how we can accurately capture the remaining timeslice of a process.

> I have four concerns here:
> 
> 1. RISC-V does not have support for pausing the mtimer (the closest equivalent to Arm SYSTICK), so an implementation for RISC-V would require software support for reading the value of the mtimer and storing it somewhere -- assuming a dedicated hardware peripheral is available -- and restoring the mtimer to that value when returning to userspace.
> 2. Each RISC-V board we currently support in Tock does not have multiple hardware timing peripherals available, so supporting application preemption on those platforms requires virtualization of the mtimer. Therefore, accurately tracking time spent in userspace would require the mtimer handler to be somehow aware of this virtualization to determine whether to consider a given mtimer interrupt as being an interrupt intended for application preemption or an interrupt intended for a userspace process or kernel capsule using the timer. This feels like it could get somewhat heavy and remove some of the gains we have made on reducing system call overhead in Tock.

IMHO RISC-V not having a dedicated timer for scheduler is a major limitation. Scheduler code is one of the most critical parts of any kernel. Sharing scheduler timer with drivers and user-space can lead potentially hard to debug scenarios. 

My hope is that ARM does not have to pay the cost of limitation from RISC-V.

> 3. This design decides that Tock only cares to enforce application preemption based on time spent in userspace by the application, such that time spent in the kernel handling syscalls triggered by an application is not attributed to that process. This approach seems inconsistent with how the Linux scheduler, for example, calculates the time used by a process (source)[] , and is inconsistent with how we attribute time to processes in the Tock kernel now. My personal feeling is that it makes sense to charge userspace applications for the in-kernel time they spend handling syscalls. At the very least I think there should be some discussion before we commit to a design that enforces one way or the other.

I agree there has to more discussion on this.

One question that comes to my mind is how would we separate the time spent in handling process A's syscalls from the time spent in bottom half interrupt processing? Process B might be responsible for most of the bottom half interrupt processing, but that would potentially get attributed to process A's timeslice.

bradjc:
I think the change in systick_handler makes sense, for the reason specified in the source comment.

The others are potentially a rather large policy change, as with this change time spent on handling commands for example are not charged against the process.

Also, that this only disables the systick in assembly, but doesn't start it, is a little confusing. It seems like that would be easy to get out of sync. Plus, it seems like the process time slice is something the core kernel is responsible for, but here it is in the arch crate.

bradjc:
If we are going to make any changes along these lines, I think it makes sense to incorporate this into #1985.

hudson-ayers:
> I think the change in systick_handler makes sense, for the reason specified in the source comment.

If the systick_handler executes, we know that the process used its entire timeslice anyway (stopping the countdown timer just stops it at 0 in this case), so I am not really sure what the benefit is of it being stopped as opposed to just calling systick.expired() in the code that returns to the kernel (which is what #1985 does)

hudson-ayers:
> One question that comes to my mind is how would we separate the time spent in handling process A's syscalls from the time spent in bottom half interrupt processing? Process B might be responsible for most of the bottom half interrupt processing, but that would potentially get attributed to process A's timeslice.

The kernel handles all bottom half interrupts outside of `do_process()`, so bottom half interrupt handling is not attributed to any process.

> @hudson-ayers I was wondering if you happen to have a PR in progress that combines #1985 with #1767?
>
> Without stopping the SysTick timer, I am not entirely sure how we can accurately capture the remaining timeslice of a process.

I am planning to rebase one once the other is merged. I am going to add get_time_remaining_us() to the `SysTick` trait in #1985 , which would be called in `do_process()` to capture the remaining timeslice before returning to the Kernel loop.

rajivr:
> > @hudson-ayers I was wondering if you happen to have a PR in progress that combines #1985 with #1767?
> > Without stopping the SysTick timer, I am not entirely sure how we can accurately capture the remaining timeslice of a process.
> 
> I am planning to rebase one once the other is merged. I am going to add get_time_remaining_us() to the `SysTick` trait in #1985 , which would be called in `do_process()` to capture the remaining timeslice before returning to the Kernel loop.

Sorry for the delay in replying.

If possible, it would be great if we can have a candidate PR that combines #1985 and #1767. At this stage, I am not entirely sure about how [`switch_to_process`](https://github.com/hudson-ayers/tock/blob/6c204b666f567b6f7d619ec5358b76cff288bf1f/arch/cortex-m/src/syscall.rs#L140) would be correctly managed, with SysTick not being disabled in trap handler (This PR).

Another related issue with SysTick trait is that there seems to an assumption on being able to accurately get remaining time-slice in micro-seconds (in `get_remaining_us`). The way SysTick reference clock (`STCLK`) gets configured can vary depending on the tune-able parameters that chip vendor provides. Currently SysTick is initialized in `chip.rs` in the chip crate.

I recently learned the following for STM32F446RE.

_The SysTick calibration value is fixed to 18750, which gives a reference time base of 1 ms with the SysTick clock set to 18.75 MHz (HCLK/8, with HCLK set to 150 MHz)._

The default HCLK for STM32 using internal RC oscillator is 16Mhz. This means, almost certainly the calibration value that we are relying on is wrong for STM32. I am not sure how things are for other chips.

Clocks are usually setup in the board crate. We might have to consider some way allowing SysTick calibration to be done from board crate.

rajivr:
> > I think the change in systick_handler makes sense, for the reason specified in the source comment.
> 
> If the systick_handler executes, we know that the process used its entire timeslice anyway (stopping the countdown timer just stops it at 0 in this case), so I am not really sure what the benefit is of it being stopped as opposed to just calling systick.expired() in the code that returns to the kernel (which is what #1985 does)

I think I should clarify the constraints that I've in my mind for SysTick trait.

1. SysTick trait needs to be marked as `unsafe`, so that the implementer is aware of additional constrains.

2. The moment we return back into the kernel, specifically at [this](https://github.com/tock/tock/blob/ec7aa6678898483d136432222c95d6ef5caa0115/arch/cortex-m/src/syscall.rs#L145) point, we need to be able to have an accurate snapshot as to why we returned back. If this snapshot changes under our feet, then it would become really hard to make correct higher-level decisions (such as scheduler decisions) based on it.

3. While we don't have to technically stop SysTick timer in `systick_handler()`, I am doing it so that we can have an uniform trait constraint that when we return back to kernel, the SysTick timer is no longer running.

4. There is a valid point around whether we should account for system call processing time into a process' time-slice. I don't have an opinion on this. My primary concern is around ensuring that we are able to correctly determine the remaining time-slice correctly (with or without system call time) and avoiding subtle bugs such as [this](https://github.com/tock/tock/pull/1767#discussion_r442563432).

hudson-ayers:
> If possible, it would be great if we can have a candidate PR that combines #1985 and #1767. At this stage, I am not entirely sure about how [`switch_to_process`](https://github.com/hudson-ayers/tock/blob/6c204b666f567b6f7d619ec5358b76cff288bf1f/arch/cortex-m/src/syscall.rs#L140) would be correctly managed, with SysTick not being disabled in trap handler (This PR).

I think the way that systick expirations are handled in the updated #1985  should work fine when combined with #1767 -- I am not sure what about #1767 precludes that. 1985 changes things so that calls to `systick.expired()` are used to detect timeslice expiration after `switch_to_process()` in the latter half of the `switch_to()` code in `process.rs`. Do you have a concern with how systick expirations are detected in #1985?

> Another related issue with SysTick trait is that there seems to an assumption on being able to accurately get remaining time-slice in micro-seconds (in `get_remaining_us`). The way SysTick reference clock (`STCLK`) gets configured can vary depending on the tune-able parameters that chip vendor provides. Currently SysTick is initialized in `chip.rs` in the chip crate.
> 
> I recently learned the following for STM32F446RE.
> 
> _The SysTick calibration value is fixed to 18750, which gives a reference time base of 1 ms with the SysTick clock set to 18.75 MHz (HCLK/8, with HCLK set to 150 MHz)._
> 
> The default HCLK for STM32 using internal RC oscillator is 16Mhz. This means, almost certainly the calibration value that we are relying on is wrong for STM32. I am not sure how things are for other chips.
> 
> Clocks are usually setup in the board crate. We might have to consider some way allowing SysTick calibration to be done from board crate.

This seems like a very legitimate concern to me. If you would be willing to make a tracking issue or PR to correct calibration values / allow calibration to be set by boards, that would be great. But I do think that concern is largely separate from whether this PR is needed.

> I think I should clarify the constraints that I've in my mind for SysTick trait.
>
> 1. SysTick trait needs to be marked as unsafe, so that the implementer is aware of additional constrains.

This seems reasonable. By additional constraints do you just mean calibration constraints?

> 2. The moment we return back into the kernel, specifically at this point, we need to be able to have an accurate snapshot as to why we returned back. If this snapshot changes under our feet, then it would become really hard to make correct higher-level decisions (such as scheduler decisions) based on it.

It is still not clear to me why this is important -- it seems to me that establishing why we have returned to the kernel [here](https://github.com/tock/tock/blob/21e0a6d8c0072cf84afe189cc5d6722e61e34713/kernel/src/process.rs#L1280) should work fine, as it still establishes the correct return reason before re-entering scheduler code.

> 3. While we don't have to technically stop SysTick timer in systick_handler(), I am doing it so that we can have an uniform trait constraint that when we return back to kernel, the SysTick timer is no longer running.
>
> 4. There is a valid point around whether we should account for system call processing time into a process' time-slice. I don't have an opinion on this. My primary concern is around ensuring that we are able to determine the remaining time-slice correctly (with or without system call time) and avoiding subtle bugs such as this.

I think these two concerns are related, but basically my thought is that if we are charging system call processing time to a process' time-slice, it is not unreasonable to also include the small amount of instructions executed in the Kernel as part of the path from the `systick_handler` up to the return from `do_process()`. The one potential bug I see with the design I have proposed in #1985  is the possibility for the systick to wrap around after a call to `systick.expired()` but before checking `get_remaining_time_us()`. However it seems this could be resolved by checking COUNTFLAG in the call to `get_remaining_time_us()`, and returning 0 in the case that it has been raised (or even by using a call to `get_remaining_time_us()` in `process.rs` when checking for expiration and saving the value from there). For RISC-V the timer counts up, so this is less of a concern, and we can just check for wraparound when subtracting.

If I make those adjustments to #1985, would you be satisfied with not disabling the SysTick in systick_handler()? I would prefer to attribute timeslice use to processes equally across architectures, unless doing so leads to a significant inefficiency for ARM, and I don't think the approach I am proposing is a significant cost for ARM chips.

Also, if I am missing some potential bug here please let me know. I really appreciate all your feedback on this :)

ppannuto:
> > SysTick trait needs to be marked as unsafe, so that the implementer is aware of additional constrains.

> This seems reasonable. By additional constraints do you just mean calibration constraints?

There's nothing unsafe about calibration in the Rust sense of unsafe is there? I think we need to use a different mechanism for this.

rajivr:
> The one potential bug I see with the design I have proposed in #1985 is the possibility for the systick to wrap around after a call to `systick.expired()` but before checking `get_remaining_time_us()`. However it seems this could be resolved by checking COUNTFLAG in the call to `get_remaining_time_us()`, and returning 0 in the case that it has been raised (or even by using a call to `get_remaining_time_us()` in `process.rs` when checking for expiration and saving the value from there). For RISC-V the timer counts up, so this is less of a concern, and we can just check for wraparound when subtracting.
> 
> If I make those adjustments to #1985, would you be satisfied with not disabling the SysTick in systick_handler()?

I spent sometime thinking about this. I can't think of any potential bugs in the new design. Thanks again for clarifying the new design.

I am still a bit concerned about sharing scheduler timer on Earlgrey and potential downstream effects of that. By the time we consider building a product based on RISC-V, I think this design choice would have gotten tested out.

hudson-ayers:
> I spent sometime thinking about this. I can't think of any potential bugs in the new design. Thanks again for clarifying the new design.
> 
> I am still a bit concerned about sharing scheduler timer on Earlgrey and potential downstream effects of that. By the time we consider building a product based on RISC-V, I think this design choice would have gotten tested out.

Thanks for all the thought you have put into this! I agree that sharing the scheduler timer is nonideal, but I think it is the best option available to us now, and is preferable to having no scheduler timer for Earlgrey.

Should we close this PR now?

rajivr:
> Should we close this PR now?

Sounds good. Thanks!

PR #1984, opened 2020-06-26
[DO NOT MERGE] kernel: Export types for external process management

jon-flatley:
This is a reference commit to outline what functions and types are
needed to externally implement the `ProcessType` trait.

For some of these it may not make sense to just expose publicly, but
should be guarded behind a capability.

The following are justifications for exposing each type and function as
why they are necessary to support the host-side testing work found
here: https://github.com/jon-flatley/tock/tree/tock_testing

Functions changed from `crate` visibility to `pub`:

`callback::AppId::new()` - Used in process creating, needed to satisfy
    `process::ProcessType::appid()`.

`mem::AppSlice::new()` - Needed for the host-side testing implementation
    of the ALLOW syscall, and is returned from
    `Process::ProcessType::allow()`.

`sched::Kernel::increment_work()` - Needed when creating a new process,
    calling `enqueue_task()`, and calling `set_process_function()`.

`sched::Kernel::decrement_work()` - Needed when calling `dequeue_task()`
    and calling `set_yielded_state()`.

`sched::Kernel::get_grant_count_and_finalize()` - Needed to manage the
    creation of grants and for parameter sanity checking in
    `get_grant_ptr()`.

Types changed from `crate` visibility to `pub`:
`callback::CallbackId` - In the signature of
    `process::ProcessType::remove_pending_callbacks()`

`process::FunctionCallSouce` - Needed to manage tasks.

`process::State` - Needed to satisfy numerous signatures and manage
    process state.

`process::Task` - Needed to manage tasks.

### Pull Request Overview

This pull request adds/changes/fixes...

### Testing Strategy

This pull request was tested by...

### TODO or Help Wanted

This pull request still needs...

### Documentation Updated

- [ ] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [ ] Ran `make prepush`.

PR #1994, opened 2020-07-01
kernel: expose APIs for external `ProcessType`s

bradjc:
### Pull Request Overview

This pull request adds new functions with `pub` visibility modifiers to enable an implementation of `ProcessType` to exist outside of the kernel/ crate. These new functions are protected with capabilities since these are very sensitive operations that shouldn't be generally called.

I named the new functions `_public`. I don't think we have an existing convention for this, so I made one up.

I added a new capability `ExternalProcessCapability` to capture the use case for making these functions and constructors public. Increment/decrement work I re-used the `MainLoopCapability` since I think that accurately captures the permissions needed to increment or decrement work. For the others I wasn't sure what to do, so I added a new capability.

Replaces #1984.

### Testing Strategy

travis

### TODO or Help Wanted

@jon-flatley will this work?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/lib.rs
+++ b/kernel/src/lib.rs
@@ -81,7 +81,7 @@
 //!
 //!    One note is that there are currently very few extensions to the core
 //!    kernel that live outside of the kernel crate. That means we have not
-//!    necessarily crated `_extern` functions for all the interfaces needed for
+//!    necessarily created `_extern` functions for all the interfaces needed for
 //!    this use case. It is likely we will have to create new interfaces as new
 //!    use cases are discovered.
 
--- a/kernel/src/lib.rs
+++ b/kernel/src/lib.rs
@@ -5,6 +5,85 @@
 //! Interface Layer (HIL) definitions.
 //!
 //! Most `unsafe` code is in this kernel crate.
+//!
+//!
+//! ## Core Kernel Visibility
+//!
+//! As the root crate in the Tock operating system, this crate serves multiple
+//! purposes:
+//!
+//! 1. It includes the logic for the core kernel, including process management,
+//!    grants, scheduling, etc.
+//!
+//! 2. It includes important interfaces for hardware and other device
+//!    abstractions. These are generally in the HIL and platform folders.
+//!
+//! 3. It includes utility functions used elsewhere in the kernel, generally by
+//!    multiple different crates such that it makes sense to have shared
+//!    implementations in the core kernel crate.
+//!
+//! Because of these different features of the core kernel, managing visibility
+//! of the various objects and functions is a bit tricky. In general, the kernel
+//! crate only exposes what it absolutely needs to. However, there are three
+//! cases where resources in this crate _must_ be exposed.
+//!
+//! 1. The shared utility functions and structs must be exposed. These are
+//!    marked `pub` and are used by many other kernel crates.
+//!
+//!    Some utility objects and abstractions, however, expose memory unsafe
+//!    behavior. These are marked as `unsafe`, and require an `unsafe` block to
+//!    use them. One example of this is `StaticRef` which is used for accessing
+//!    memory-mapped I/O registers. Since accessing the addresses through just a
+//!    memory address is potentially very unsafe, instantiating a `StaticRef`
+//!    requires an `unsafe` block.
+//!
+//! 2. The core kernel types generally have to be exposed as other layers of the
+//!    OS need to use them. However, generally only a very small interface is
+//!    exposed, and using that interface cannot compromise the overall system or
+//!    the core kernel. These functions are also marked `pub`. For example, the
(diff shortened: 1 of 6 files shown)


kernel/src/sched.rs
```diff
@@ -61,16 +61,37 @@ impl Kernel {
...
+    ///
+    /// This is exposed publicly, but restricted with a capability. The intent
+    /// is that external implementations of `ProcessType` need to be able to
+    /// indicate there is more process work to do.
+    pub fn increment_work_public(&self, _capability: &dyn capabilities::MainLoopCapability) {
+        self.work.increment();
+    }
+
     /// Something finished for a process, so we decrement how much work there is
     /// to do.
```
hudson-ayers, line 80:
```suggestion
    /// Something finished for a process, so we decrement how much work there is
    /// to do.
    ///
    /// This is only exposed in the core kernel crate
```
(to match added annotation for increment_work)

kernel/src/sched.rs
```diff
@@ -61,16 +61,37 @@ impl Kernel {
...
     pub(crate) fn increment_work(&self) {
         self.work.increment();
     }
 
+    /// Something was scheduled for a process, so there is more work to do.
+    ///
+    /// This is exposed publicly, but restricted with a capability. The intent
+    /// is that external implementations of `ProcessType` need to be able to
+    /// indicate there is more process work to do.
+    pub fn increment_work_public(&self, _capability: &dyn capabilities::MainLoopCapability) {
```
hudson-ayers, line 75: Would it make more sense for this to just take in `ExternalProcessCapability`? It doesn't make a lot of sense to me to give some process type the ability to start the main loop, which is what this capability was previously used for.

For reference, [here](https://github.com/jon-flatley/tock/blob/124383bbc7dbc0916800359adc67bbd73da8f48d/host-testing/src/process.rs#L266) is where it is called in the host side code.

Most current uses of this function in Tock come from within `process.rs` anyway, as opposed to `sched.rs`

bradjc: I can get behind this idea.

bradjc:
> The alternative to this approach would be to simply have all of these operations require a `ProcessCapability` that could be given to any process type (external or otherwise), and then give that capability to the internal ProcessType as well.
> We could even require `MainLoopCapability` types implement `ProcessCapability` so that the main loop would be able to directly call these functions as well.
> 
> This would prevent the use of duplicated functions with different visibilities that could get out of sync. Just an idea though.

I think that is a reasonable approach certainly. I didn't want to change the convention that functions that require capabilities are exclusively public and not designed to be used internally. You did remind me that I meant to use the existing implementations and I added a commit to do exactly that.

jon-flatley:
This looks great, thanks @bradjc!

I'm not so sure about the `_public` naming convention as it's inconsistent with existing functions in the kernel API. I think ideally we would have the `pub` functions exposed under their existing `pub(crate)` names, but I'm not sure how this could be done ergonomically, apart from @hudson-ayers' idea to use capabilities internally.

bradjc:
I agree a different name would be better, since this one does conflate the different types of "public". I don't think we currently have any "public" functions for extensions of the kernel crate. We have `pub` for external users of the core kernel, and `pub(crate)` for internal implementations of the core kernel, but nothing public for implementations of the core kernel that just so happen to be outside of the kernel crate.

I like something like naming them with `_external`.

jon-flatley:
Ahh, I see the distinction. Under that mode of thought `_public` or `_external` names make more sense.

hudson-ayers:
+1 for `_external`

bradjc:
I switched to _external and standardized the capability.

I also captured the discussion from this PR in lib.rs documenting why various pieces of the kernel are public.

kernel/src/lib.rs
```diff
@@ -5,6 +5,85 @@
...
+//!    hold a Capability to call them. This helps restrict their use and makes
+//!    it very clear that calling them requires special permissions.
+//!    Additionally, to differentiate these interfaces, which are for external
+//!    extensions of core kernel functionality, from the other public but
+//!    sensitive interfaces (item 2 above), we append the name `_external` to
+//!    the function name.
+//!
+//!    One note is that there are currently very few extensions to the core
+//!    kernel that live outside of the kernel crate. That means we have not
+//!    necessarily crated `_extern` functions for all the interfaces needed for
```
hudson-ayers, line 84:
```suggestion
//!    necessarily created `_extern` functions for all the interfaces needed for
```

bradjc:
@tock/core-wg

bradjc:
@jon-flatley mentioned he rebased his work on top of this and things worked as expected.

ppannuto:
bors d+

PR #2000, opened 2020-07-04
Adding ADC support for stm323f303xc

cosmindanielradu19:
### Pull Request Overview

This pull request adds single sample ADC support for stm323f303xc

### Testing Strategy

This pull request was tested by using a stm323f3discovery kit

### TODO or Help Wanted

Continuous mode implementation troubles. The speed of interrupts is too high and can't send response with Data to the client.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


boards/stm32f3discovery/src/main.rs
```diff
@@ -497,6 +659,63 @@ pub unsafe fn reset_handler() {
...
+            // PE10 ADC3
+            &stm32f303xc::adc::Channel::Channel14,
+            // OA1 ADC1 (INTERNAL), PE11 ADC3
+            &stm32f303xc::adc::Channel::Channel15,
+            // TS ADC1 (INTERNAL), PE12 ADC3
+            &stm32f303xc::adc::Channel::Channel16,
+            // BT/2 ADC1 (INTERNAL), OA2 ADC2 (INTERNAL), OA3 ADC3 (INTERNAL), OA4 ADC4 (INTERNAL)
+            &stm32f303xc::adc::Channel::Channel17,
+            // VRI on ADC1, ADC2, ADC3 and ADC4 (INTERNALS)
+            &stm32f303xc::adc::Channel::Channel18,
```
hudson-ayers, line 702: we should probably just pick a few ADC channels to include by default so that GPIOs are still available.

chips/stm32f303xc/src/adc.rs
```diff
@@ -0,0 +1,740 @@
...
+//     // adc12: AdcSeparateRegisters,
+//     // _reserved0: [u32; 48],
+
+//     // adc34: AdcSeparateRegisters,
+//     // _reserved3: [u32; 109],
+
+//     // common: AdcCommonRegisters,
+
+// }
+
```
hudson-ayers, line 24: is there any reason to keep these comments around?

chips/stm32f303xc/src/adc.rs
```diff
@@ -0,0 +1,740 @@
...
+        DELAY OFFSET(8) NUMBITS(4) [],
+        // Dual ADC mode selection
+        DUAL OFFSET(0) NUMBITS(5) []
+    ],
+    // Common regular data register for dual mode
+    CDR [
+        // Regular data of the slave ADC
+        RDATA_SLV OFFSET(16) NUMBITS(16) [],
+        // Regular data of the master ADC
+        RDATA_MST OFFSET(0) NUMBITS(16) []
```
hudson-ayers, line 414: These descriptive comments are awesome, but it would be great if you could make them doc comments so they show up in the docs (i.e. replace `//` --> `///`)

chips/stm32f303xc/src/adc.rs
```diff
@@ -0,0 +1,740 @@
...
+pub static mut ADC1: Adc = Adc::new();
+
+impl Adc {
+    const fn new() -> Adc {
+        Adc {
+            registers: ADC1_BASE,
+            //rcc copy
+            common_registers: ADC12_COMMON_BASE,
+            clock: AdcClock(rcc::PeripheralClock::AHB(rcc::HCLK::ADC1)),
+            // clock34: AdcClock(rcc::PeripheralClock::AHB(rcc::HCLK::ADC34)),
```
hudson-ayers, line 527: I noticed this clock is enabled in rcc.rs, is there a reason it is enabled there but ignored here?

chips/stm32f303xc/src/adc.rs
```diff
@@ -0,0 +1,740 @@
...
+        // Wait for ADVRGEN to enable
+        // This needs to be synchronous because there is no interrupt signaling
+        // when ADVRGEN becomes enabled
+        for _i in 0..1000000 {
+            unsafe {
+                llvm_asm!(
+                "nop"
+            : : : : "volatile" );
+            }
+        }
```
hudson-ayers, line 554: is there a bit we can check that will indicate when ADVRGEN has been enabled? Looping for a million cycles seems a shame if we can short circuit once it becomes ready, and it would be nice to not add `llvm_asm` to this crate if we don't have to.

chips/stm32f303xc/src/adc.rs
```diff
@@ -0,0 +1,740 @@
...
+
+        // Enable the temperature sensor
+        self.common_registers.ccr.modify(CCR::TSEN::SET);
```
hudson-ayers, line 557: What does this have to do with the ADC? Is it used invisibly by the hardware to correct for temperature affects somehow?

chips/stm32f303xc/src/adc.rs
```diff
@@ -0,0 +1,740 @@
...
+        self.registers.ier.modify(IER::ADRDYIE::SET);
+
+        // Clear registers
+        self.registers.isr.modify(ISR::ADRDY::CLEAR);
+        self.registers.cr.modify(CR::ADEN::CLEAR);
+        self.registers.cr.modify(CR::ADCALDIF::CLEAR);
+        self.registers.cr.modify(CR::ADCAL::SET);
+
+        // Wait for calibration
+        while self.registers.cr.is_set(CR::ADCAL) {}
```
hudson-ayers, line 569: Do you have a sense of how many cycles this wait should take? As a general rule, we favor asynchronous designs unless the wait is expected to be very small.

chips/stm32f303xc/src/adc.rs
```diff
@@ -0,0 +1,740 @@
...
+            // Clear interrupt
+            self.registers.ier.modify(IER::OVRIE::CLEAR);
+            self.registers.isr.modify(ISR::OVR::SET);
+        }
+    }
+
+    pub fn set_client<C: EverythingClient>(&self, client: &'static C) {
+        self.client.set(client);
+    }
+    // 12
```
hudson-ayers, line 621:
```suggestion

```

chips/stm32f303xc/src/adc.rs
```diff
@@ -0,0 +1,740 @@
...
+        _length1: usize,
+        _buffer2: &'static mut [u16],
+        _length2: usize,
+    ) -> (
+        ReturnCode,
+        Option<&'static mut [u16]>,
+        Option<&'static mut [u16]>,
+    ) {
+        (ReturnCode::ENOSUPPORT, None, None)
+    }
```
hudson-ayers, line 714: If high speed is not supported, it seems like it makes more sense to simply not implement the trait. Is there something about how the ADC capsule is written that makes doing so not ergonomic?

chips/stm32f303xc/src/adc.rs
cosmindanielradu19: No, it just needs a wait time for it to be enabled and to continue the ADC powering on

chips/stm32f303xc/src/adc.rs
cosmindanielradu19: At the moment just ADC12 is used and tested, ADC34 is on progress, just need to test a bit more

chips/stm32f303xc/src/adc.rs
cosmindanielradu19: No, there is not a fixed amount of cycles needed that's why I chose busy waiting for this to wait the completion

chips/stm32f303xc/src/adc.rs
cosmindanielradu19: Couldn't compile ADC and integrate it in the kernel without high speed Hil implementation

chips/stm32f303xc/src/adc.rs
cosmindanielradu19: It will be enabled on temperature capsule, it was just for testing. I will do this as soon as I get ADC mux working

chips/stm32f303xc/src/adc.rs
cosmindanielradu19: This is the link to the reference manual for stm32f3xx https://www.st.com/resource/en/reference_manual/dm00043574-stm32f303xb-c-d-e-stm32f303x6-8-stm32f328x8-stm32f358xc-stm32f398xe-advanced-arm-based-mcus-stmicroelectronics.pdf and on page 316 it is described how to enable ADVREGEN. Am I missing anything? I’m open to any ideas.

chips/stm32f303xc/src/adc.rs
cosmindanielradu19: This is the link to the reference manual for stm32f3xx https://www.st.com/resource/en/reference_manual/dm00043574-stm32f303xb-c-d-e-stm32f303x6-8-stm32f328x8-stm32f358xc-stm32f398xe-advanced-arm-based-mcus-stmicroelectronics.pdf and on page 317 it is described how to do the calibration process. It says that the ADCAL bit will be cleared by the hardware as soon as the process is completed.

chips/stm32f303xc/src/adc.rs
cosmindanielradu19: If we can split the ADC capsule in two different parts, more exactly ADC and High Speed ADC, this would work, otherwise the implementation for High Speed is needed for the capsule to be compiled alongside the kernel.

boards/stm32f3discovery/src/main.rs
cosmindanielradu19: Ok, I will remove some of the ADC pins and leave them as GPIOs and I will come back with an update

chips/stm32f303xc/src/adc.rs
hudson-ayers: That seems to indicate the worst case delay is 10 microseconds. Also, my understanding is that the maximum core clock speed is 72 MHz on this chip, so I think the worst case number of cycles should be 720. Can you replace 1000000 with 720, and add a comment explaining why that number is used? 

Ideally we would have a safe wait primitive like that described in #1552 , but for now I think limiting this to 720 cycles should reflect the spirit of that.

chips/stm32f303xc/src/adc.rs
hudson-ayers: Should this not be enabled in the temperature capsule then?

boards/stm32f3discovery/.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,16 @@
+{
+    "configurations": [
+        {
+            "name": "Linux",
+            "includePath": [
+                "${workspaceFolder}/**"
+            ],
+            "defines": [],
+            "compilerPath": "/usr/bin/gcc",
```
alistair23, line 9: This probably shouldn't be included in the Tock repo

boards/stm32f3discovery/src/main.rs
```diff
@@ -358,94 +436,103 @@ pub unsafe fn reset_handler() {
...
+            86 => stm32f303xc::gpio::PinId::PC09.get_pin().as_ref().unwrap()
         ),
     )
     .finalize(components::gpio_component_buf!(
         stm32f303xc::gpio::Pin<'static>
     ));
 
+    //start adc12
+    stm32f303xc::adc::ADC1.enable();
+
```
bradjc, line 535: Can you move this down to the other ADC setup code?

chips/stm32f303xc/src/adc.rs
```diff
@@ -0,0 +1,726 @@
+use crate::rcc;
```
bradjc, line 1:
```suggestion
//! Analog to Digital Converter Peripheral

use crate::rcc;
```

chips/stm32f303xc/src/adc.rs
bradjc: Hmm ok that makes sense. I'm not sure about providing a fake interface to userspace, the ADC capsule still isn't supported on the stm32 at this point. I can imagine being on the other side as a developer and thinking ADC will work and then learning why all of the syscalls are returning errors.

cosmindanielradu19: I still suggest splitting the ADC capsule in two as this will allow an easier implementation of a virtual ADC

boards/stm32f3discovery/src/main.rs
```diff
@@ -497,6 +581,40 @@ pub unsafe fn reset_handler() {
     let ninedof = components::ninedof::NineDofComponent::new(board_kernel)
         .finalize(components::ninedof_component_helper!(l3gd20, lsm303dlhc));
 
+    //start adc12
+    stm32f303xc::adc::ADC1.enable();
```
brghena, line 585: Why enable the ADC here instead of within the `sample()` call? The general idea when the ADC HIL was written was that calls to `sample()` should enable, configure, and sample the ADC so that a developer can't ever forget to do part of that and have the ADC in an indeterminate state.

chips/stm32f303xc/src/adc.rs
brghena: I think that the ADC capsule should have been (and still should be) split in two. That didn't happen at the time because of the complexity of implementing virtualization to share one ADC between two interfaces, which remains true today.

I think in lieu of more immediate work on the ADC capsule side, this ENOSUPPORT implementation is a reasonable choice. I would follow this up with a comment in main.rs by the ADC configuration that the highspeed ADC trait is not actually implemented for now, and when you fill out https://github.com/tock/tock/blob/master/chips/README.md#hil-support, I would not mark highspeed ADC as implemented.

chips/stm32f303xc/src/rcc.rs
```diff
@@ -594,6 +595,36 @@ impl Rcc {
...
     }
+
+    // ADC12 clock
+
+    fn is_enabled_adc12_clock(&self) -> bool {
+        self.registers.ahbenr.is_set(AHBENR::ADC12EN)
+    }
+
+    fn enable_adc12_clock(&self) {
+        debug!("enable adc12 clock");
```
bradjc, line 606:
```suggestion
```

chips/stm32f303xc/src/adc.rs
```diff
@@ -0,0 +1,752 @@
...
+            common_registers: ADC12_COMMON_BASE,
+            clock: AdcClock(rcc::PeripheralClock::AHB(rcc::HCLK::ADC1)),
+            // clock34: AdcClock(rcc::PeripheralClock::AHB(rcc::HCLK::ADC34)),
+            status: Cell::new(ADCStatus::Off),
+            client: OptionalCell::empty(),
+        }
+    }
+
+    pub fn enable(&self) {
+        debug!("ENABLE");
```
bradjc, line 526:
```suggestion
```

chips/stm32f303xc/src/adc.rs
```diff
@@ -0,0 +1,752 @@
+//! Analog to Digital Converter Peripheral
+
+use crate::rcc;
+use core::cell::Cell;
+use kernel::common::cells::OptionalCell;
+use kernel::common::registers::{register_bitfields, ReadOnly, ReadWrite};
+use kernel::common::StaticRef;
+use kernel::debug;
```
bradjc, line 8:
```suggestion
```

bradjc:
@hudson-ayers

PR #2001, opened 2020-07-06
boards: Make .app section default, do not include it in kernel binary

bradjc:
### Pull Request Overview

Having a .apps section makes it straightforward to concatenate application binaries into the kernel elf to create a monolithic 
image that flashes the kernel and applications together. However, this has had the side effect that flashing only the kernel erases applications. And, it is only supported on some boards. This PR makes creating a .apps section default, and also removes it from the kernel binary so that the kernel alone will not erase apps.

#### More detail:

The .apps section in flash is a placeholder section that satisfies two properties:

1. Its physical address is the same address as the start of apps in flash.
2. The linker places it into a segment, so that when the .elf file is used, the .apps section will be loaded at the correct address.

In the kernel build system, we have no application binaries, so the
.apps section only includes a placeholder byte so that the linker keeps
it around and creates a segment for it. Later, the .elf file can be
updated with an actual application binary in the .apps section, and the
application binary will be loaded with the kernel binary.

Due to the requirements above, that single placeholder byte means the
kernel .elf file creates a binary with one byte at the address where
applications start. This, when flashed, overwrites the first application
header, effectively erasing all apps.

However, when we are creating just the kernel binary file in the build
system, we do not need to include the .apps section (it is just a
placeholder after all). So, we explicitly ignore it in the .elf file
when creating the kernel binary.

### Testing Strategy

Verifying the hail kernel does not overwrite apps.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/Makefile.common
+++ b/boards/Makefile.common
@@ -140,9 +140,13 @@ SIZE      ?= $(TOOLCHAIN)-size
 OBJCOPY   ?= $(TOOLCHAIN)-objcopy
 OBJDUMP   ?= $(TOOLCHAIN)-objdump
 
-# Set additional flags to produce binary from .elf
-# --strip-sections prevents enormous binaries when SRAM is below flash
-OBJCOPY_FLAGS ?= --strip-sections -S
+# Set additional flags to produce binary from .elf.
+# * --strip-sections prevents enormous binaries when SRAM is below flash.
+# * --remove-section .apps prevents the .apps section from being included in the
+#   kernel binary file. This section is a placeholder for optionally including
+#   application binaries, and only needs to exist in the .elf. By removing it,
+#   we prevent the kernel binary from overwriting applications.
+OBJCOPY_FLAGS ?= --strip-sections -S --remove-section .apps
 # This make variable allows board-specific Makefiles to pass down options to
 # the Cargo build command. For example, in boards/<custom_board>/Makefile:
 # `CARGO_FLAGS += --features=foo` would pass feature `foo` to the top level
--- a/boards/kernel_layout.ld
+++ b/boards/kernel_layout.ld
@@ -185,9 +185,10 @@ SECTIONS
     PROVIDE_HIDDEN (__exidx_end = .);
 
     /* Region for on-chip kernel non-volatile storage.
-     * Align on 512 bytes since that is the page size.
-     * Volumes within this region are allocated with the
-     * storage_volume! macro in utils.rs. */
+     *
+     * Align on 512 bytes since that is the page size. Volumes within this
+     * region are allocated with the storage_volume! macro in utils.rs.
+     */
     .storage :
     {
       . = ALIGN(512);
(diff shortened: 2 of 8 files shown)


doc/Compilation.md
```diff
@@ -103,6 +104,28 @@ has three main ramifications:
...
+To use the special `.apps` section, `objcopy` can replace the placeholder with an actual app binary. The general command looks like:
+
+```bash
+$ arm-none-eabi-objcopy --update-section .apps=libtock-c/examples/c_hello/build/cortex-m4/cortex-m4.tbf target/thumbv7em-none-eabi/release/stm32f412gdiscovery.elf target/thumbv7em-none-eabi/release/stm32f4discovery-app.elf
+```
+
+This replaces the placeholder section `.apps` with the "c_hello" application TBF
+in the stm32f412gdiscovery.elf kernel ELF, and creates a new .elf called
+`stm32f4discovery-app.elf`.
+
```
brghena, line 128: Should there be a discussion here about what to do with that elf file in order to get it on a board? Maybe a comment that says "board-specific README files explain this process and how to load the ELF onto that particular board".

brghena: What's missing right now is how I would load a monolithic kernel on something like Hail or an nRF dev kit where it isn't the default. Running `make flash` or `make program` after running this objcopy would just load a kernel with the apps section stripped. I would end up having to parse through the board make files for a bit to find some `objcopy -O binary` incantation followed by another JTAG incantation.

boards/kernel_layout.ld
```diff
@@ -185,9 +185,10 @@ SECTIONS
     PROVIDE_HIDDEN (__exidx_end = .);
 
     /* Region for on-chip kernel non-volatile storage.
-     * Align on 512 bytes since that is the page size.
-     * Volumes within this region are allocated with the
-     * storage_volume! macro in utils.rs. */
+     *
+     * Align on 512 bytes since that is the page size. Volumes within this
+     * region are allocated with the storage_volume! macro in utils.rs.
+     */
```
hudson-ayers, line 191: not really related to this PR, but it is not generally the case that the page size is 512 bytes. That is true for the sam4l, but  the stm32f303 has 2kB pages. Perhaps we should make this configurable somehow?

doc/Compilation.md
bradjc: Add a commit? I don't think we can promise that board readmes will explain this. My intent was to explain this enough that it is written down somewhere in some form, but not to imply this is the default or intended behavior.

boards/kernel_layout.ld
bradjc: Yeah, we should.

hudson-ayers: Opened #2006

boards/Makefile.common
```diff
@@ -140,9 +140,13 @@ SIZE      ?= $(TOOLCHAIN)-size
...
-# Set additional flags to produce binary from .elf
-# --strip-sections prevents enormous binaries when SRAM is below flash
-OBJCOPY_FLAGS ?= --strip-sections -S
+# Set additional flags to produce binary from .elf.
+# * --strip-sections prevents enormous binaries when SRAM is below flash.
+# * --remove-section .apps prevents the .apps section from being included in the
+#   kernel binary file. This section is a placeholder for optionally including
+#   application binaries, and only needs to exist in the .elf. By removing it,
+#   we prevent the kernel binary from overwriting applications.
+OBJCOPY_FLAGS ?= --strip-sections -S --remove-section .apps
```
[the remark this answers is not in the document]
bradjc: ?

bradjc:
@tock/core-wg

PR #2002, opened 2020-07-06
cortex-m: Detect stack overflow

### LINKED ISSUE #1176: Hard Fault Handler Pushes to Stack
Currently, overflowing the kernel stack results in the hard fault handler being called, as the MPU ensures only valid memory regions can be accessed. However, the hard fault handler immediately pushes two registers onto the stack, resulting in yet another hard fault. This manifests in a complete lockup of the device, with no panic or debug info printed to the console, and no blinking LEDs to indicate a panic. 

Stepping through such a fault in gdb only shows that the board has halted within `hard_fault_handler()` in lib.rs, along with the message "clearing lockup after double fault". Further, backtrace does not work after this has occurred, as gdb detects that the stack has been corrupted. As a result, the only definitive clue that a stack overflow has occurred is that printing the registers reveals that the stack pointer is below 0x20000000. Stack overflows should lead to clear debug messages.

bradjc:
### Pull Request Overview

This pull request adds logic to the kernel hardfault handler to detect a stack overflow condition, and resets the stack before using it. With IPC on hail I'm able to trigger this case and verify I get the panic handler.

This also moves the userland state saving down below since I'm not sure why that code was run in both the kernel and userspace fault case.

Fixes #1176.

### Testing Strategy

Verifying that IPC on hail causes this panic handler to be displayed.

### TODO or Help Wanted

1. Someone to look over the assembly.

2. I can copy to m3 once we are happy with it. I'm not sure this will work on cortex-m0 due to differences in the SCB.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m4/src/lib.rs
+++ b/arch/cortex-m4/src/lib.rs
@@ -444,52 +444,96 @@ pub unsafe extern "C" fn hard_fault_handler() {
     let faulting_stack: *mut u32;
     let kernel_stack: bool;
 
+    // First need to determine if this a kernel fault or a userspace fault.
     llvm_asm!(
-     "/* Read the SCB registers. */
-     ldr r0, =SCB_REGISTERS
-     ldr r1, =0xE000ED14
-     ldr r2, [r1, #0] /* CCR */
-     str r2, [r0, #0]
-     ldr r2, [r1, #20] /* CFSR */
-     str r2, [r0, #4]
-     ldr r2, [r1, #24] /* HFSR */
-     str r2, [r0, #8]
-     ldr r2, [r1, #32] /* MMFAR */
-     str r2, [r0, #12]
-     ldr r2, [r1, #36] /* BFAR */
-     str r2, [r0, #16]
-
-     mov    r1, 0                       \n\
-     tst    lr, #4                      \n\
-     itte   eq                          \n\
-     mrseq  r0, msp                     \n\
-     addeq  r1, 1                       \n\
-     mrsne  r0, psp                     "
+    "
+    mov    r1, 0     /* r1 = 0 */
+    tst    lr, #4    /* bitwise AND link register to 0b100 */
+    itte   eq        /* if lr==4, run next two instructions, else, run 3rd instruction. */
+    mrseq  r0, msp   /* r0 = kernel stack pointer */
+    addeq  r1, 1     /* r1 = 1, kernel was executing */
+    mrsne  r0, psp   /* r0 = userland stack pointer */"
     : "={r0}"(faulting_stack), "={r1}"(kernel_stack)
     :
     : "r0", "r1"
-    : "volatile"
-    );
+    : "volatile" );
 
     if kernel_stack {
-        kernel_hardfault(faulting_stack);
-    } else {
-        // hard fault occurred in an app, not the kernel. The app should be
-        //  marked as in an error state and handled by the kernel
+        // Need to determine if we had a stack overflow before we push anything
+        // on to the stack. We check this by looking at the BusFault Status
+        // Register's (BFSR) `LSPERR` and `STKERR` bits to see if the hardware
+        // had any trouble stacking important registers to the stack during the
+        // fault. If so, then we cannot use this stack while handling this fault
+        // or we will trigger another fault.
+        let stack_overflow: bool;
         llvm_asm!(
-            "ldr r0, =APP_HARD_FAULT
-              mov r1, #1 /* Fault */
-              str r1, [r0, #0]
+        "
+        ldr   r2, =0xE000ED29  /* SCB BFSR register address */
+        ldrb  r2, [r2]         /* r2 = BFSR */
+        tst   r2, #0x30        /* r2 = BFSR & 0b00110000; LSPERR & STKERR bits */
+        ite   ne               /* check if the result of that bitwise AND was not 0 */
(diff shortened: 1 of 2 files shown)


arch/cortex-m4/src/lib.rs
```diff
@@ -444,52 +444,96 @@ pub unsafe extern "C" fn hard_fault_handler() {
...
+            // To make room for a panic!() handler stack, we just re-use the
+            // kernel's original stack. This should in theory leave the bottom
+            // of the stack where the problem occurred untouched should one want
+            // to further debug.
+            llvm_asm!(
+            "
+            mov sp, r0   /* Set the stack pointer to _estack */"
+            :
+            : "{r0}"((_estack as *const ()) as u32)
+            : );
```
ppannuto, line 497: This needs `volatile` (as there are no "externally visible" effects, aka no code depends on any register or variable values written to by this assembly)

arch/cortex-m4/src/lib.rs
```diff
@@ -444,52 +444,96 @@ pub unsafe extern "C" fn hard_fault_handler() {
...
+
+        /* Set thread mode to privileged */
+        mov r0, #0
+        msr CONTROL, r0
+        /* CONTROL writes must be followed by ISB */
+        /* http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dai0321a/BIHFJCAC.html */
+        isb
+
+        movw LR, #0xFFF9
+        movt LR, #0xFFFF"
```
ppannuto, line 536: (comment for next line)

This doesn't seem to have been a problem before, but this assembly should mark its clobbers...

arch/cortex-m4/src/lib.rs
bradjc, line 497:
```suggestion
            : "volatile" );
```

arch/cortex-m4/src/lib.rs
```diff
@@ -444,52 +444,96 @@ pub unsafe extern "C" fn hard_fault_handler() {
...
+        tst   r2, #0x30        /* r2 = BFSR & 0b00110000; LSPERR & STKERR bits */
+        ite   ne               /* check if the result of that bitwise AND was not 0 */
+        movne r3, #1           /* BFSR & 0b00110000 != 0; r3 = 1 */
+        moveq r3, #0           /* BFSR & 0b00110000 == 0; r3 = 0 */"
+        : "={r3}"(stack_overflow)
+        :
+        : "r3"
+        : "volatile" );
+
+        if stack_overflow {
```
brghena, line 482: Shouldn't this run in both cases? Conceptually, an app error could cause a kernel stack overflow if the stack was already really close to full when the app had its error. I agree that that will likely never occur, but if it's possible and this logic can cleanly be moved above the `if kernel_stack {` then why not?

bradjc: In the app case we could. I guess the idea is who cares about the kernel stack if an app faulted?

In the kernel case we probably can? We do want to read values the hw pushed on the kernel stack but assuming calling `kernel_hardfault` doesn't overwrite those it would be fine. But I think the bigger issue is someone would probably want to be able to debug the kernel stack if their kernel hardfaults.

brghena: I guess what I'm saying is that I think an application error could still cause a double-hardfault. For example, assume an application error occurs, so the `hardfault_handler` is called, but before that, various registers are placed on the kernel stack. If the kernel stack was nearly full at that time, it could now overflow which would be detected by your stack_overflow check, but we came from an app instead of the kernel.

We don't actually use the stack pointer in the from-userland portion of the `hardfault_handler` though, so another hardfault would only occur when returning from this handler into the middle of the user swap code. So maybe that would just be end up triggering a from-kernel hardfault which you would catch? I'm not really sure of the specifics there.

A defense against this would be to just check the kernel stack for overflows no matter where we come from. Because if there was an overflow, I've got problems I want to know about, even if it was an app that surfaced it.

bradjc: The HW pushes values on to the kernel stack if an app faults? Where do we reset the kernel stack when that happens? Do we actually need to? Arg why is none of this code commented. I feel like I missed the class where all of this cortexm hardware magic was explained.

bradjc: Is it possible for the kernel stack to be nearly full when an application is executed? Since we only call switch_to in one place, can a long stack actually occur? Or, how could one test this to see what happens?

brghena: The hardware needs to push all caller-saved registers so that the interrupt handler doesn't overwrite them. (https://interrupt.memfault.com/blog/arm-cortex-m-exceptions-and-nvic#exception-entry--exit) It automatically adjusts the stack pointer too. However, the ARM documentation suggests that the hardware uses the _current_ stack to place the caller-saved registers on. (https://stackoverflow.com/questions/45461739/which-stack-is-used-when-entering-interrupt-handler-from-user-program-for-cortex) So an exception from userland cannot cause a kernel stack overflow (although it could cause an application stack overflow).

I also agree that in proper execution, the kernel stack should not be nearly full when an application is executed, so this shouldn't come up anyways.

bradjc:
@tock/core-wg

arch/cortex-m4/src/lib.rs
ppannuto: Moved to #2037.

PR #2003, opened 2020-07-06
kernel: grant: do not pass T::default()

bradjc:
This changes how grant allocation is structured in grant.rs so that the
allocation function does not take a copy of the data to be written when
creating a new grant. This data can be large if the grant region is
large, and it can cause a stack overflow when the allocate function is
called.

The change is pretty simple, but I'm not sure if there is some subtle reason to not do it this way.

Fixes the IPC stack overflow issue found in #1933. Replaces #1976.

### Testing Strategy

Running the hail app on hail.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -81,16 +81,16 @@ impl<T: ?Sized> DerefMut for Owned<T> {
 }
 
 impl Allocator {
-    pub fn alloc<T>(&mut self, data: T) -> Result<Owned<T>, Error> {
+    pub fn alloc<T: Default>(&mut self) -> Result<Owned<T>, Error> {
         unsafe {
-            let ptr = self.alloc_unowned(data)?;
+            let ptr = self.alloc_unowned()?;
             Ok(Owned::new(ptr, self.appid))
         }
     }
 
     // Like `alloc`, but the caller is responsible for free-ing the allocated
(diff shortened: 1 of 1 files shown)


kernel/src/grant.rs
```diff
@@ -102,7 +102,7 @@ impl Allocator {
 
                         // We use `ptr::write` to avoid `Drop`ping the uninitialized memory in
                         // case `T` implements the `Drop` trait.
-                        write(ptr.as_ptr(), data);
+                        write(ptr.as_ptr(), T::default());
```
brghena, line 105: I'm pretty uncertain why this change would actually solve the "grants temporarily take up a ton of stack space" problem. `T::default()` is still going to return a T, which will be its full size and on the stack. Is passing it into the `alloc_unowned()` function taking double the size of T on the stack or something?

brghena: Per a slack discussion with bradjc, this change greatly increases the possible size of IPC grants before a kernel stack overflow occurs (previously 364 bytes increases to 2364 bytes). I'm guessing some compiler optimizations are making the magic happen here.

alevy: Right, presumably the compiler can better guess to just allocate the space for the default value directly at `ptr.as_ptr()` as opposed to on the stack first.

PR #2004, opened 2020-07-07
Adding ADC support for stm323f4xx

cosmindanielradu19:
### Pull Request Overview

This pull request adds single sample ADC support for stm323f4xx.

### Testing Strategy

This pull request was tested by using a stm323f412g discovery

### TODO or Help Wanted

Continuous mode implementation troubles. The speed of interrupts is too high and can't send response with Data to the client.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/stm32f3discovery/Makefile
+++ b/boards/stm32f3discovery/Makefile
@@ -8,6 +8,10 @@ include ../Makefile.common
 OPENOCD=openocd
 OPENOCD_OPTIONS=-f openocd.cfg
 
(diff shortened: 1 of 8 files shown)


brghena:
Thanks for all the STM32 ADC work! I think whatever issues and decisions come up in #2000 apply equally here as well. And once #2000 is merged, this should be pretty straightforward to pull.

Are there any major differences between the ADC implementation for the F3 and F4 that we should be on the lookout for?

cosmindanielradu19:
There are some differences, there are different registers, the powering on sequence is different and the way that samples are requested differ.

Issue #1654, opened 2020-03-03
Building Tock with stable rustc


mcy:
This issue tracks attempting to build as many Tock libraries as possible with a stable compiler, that is, minimizing the number of `#![feature(...)]` pragmas used in `lib.rs` files.

This work is especially attractive to OpenTitan's use of Tock as a secure embedded operating system. Being able to build code without exercising unstable (and, as such, untested and unproven) code paths in the compiler increases confidence in the language-level and hardware-level guarantees provided by Tock. This is not to mention that unstable features are not guaranteed to be stabilized, or to remain in the compiler indefinitely.

This work is divided up according to pragmas still present after https://github.com/tock/tock/pull/1643, which every pragma that could be removed without causing build failures. Below is a list of all pragmas Tock still uses.

- [x] (#1643) Remove all pragmas currently not required to build Tock.
- [ ] (#????) `asm` and `naked_functions`. Assembly is a hard requirement for any kernel, it is possible to move all of it into `.S` files instead. This will require build system work to invoke LLVM's assembler through `clang` or similar. There have been attempts to stabilize this recently (which I have been involved in); driving stabilization forward is also an option, though the stable syntax is likely to be incompatible with the current one, which is just a Clang/GCC dialect.
- [ ] (#????) `core_intrinsics`. Intrinsics will never be stabalized, and are subject to unannounced changes by rustc's developers. We use three classes of intrinsics:
  - `core::intrinsics::offset`, which is just pointer arithmetic. We should be using `<*mut T>::offset`, which is guaranteed to have the same effect and the same optimization characteristics.
  - `core::intrinsics::atomic_*`, various atomic intrinsics. These will be very difficult to get rid of due to how Rust decides to provide `AtomicT` types, and, in the case of `kernel`'s `AtomicUsize` will require consultation with an atomics expert.
  - `core::intrinsics::sqrtf32`, a square root intrinsic. `f32::sqrt` is not present in `core` even though it just calls the intrinsic. This is because on many platforms, the intrinsic expands into a `libm` call, though not on any platforms Tock is built for. Removing this may require implementing a polyfill in assembly.
- [ ] (#????) `lang_items`. The main purpose of this pragma is declaring the global panic handler in a `no_std` binary. Once `no_std` binaries are stabilized, we will have an upgrade path out of this pragma; until then, we are very, very stuck with it.
[...]

PR #2009, opened 2020-07-10
Remove OptionalCell's dependency on feature(const_fn) using where clauses

jrvanwhy:
### Pull Request Overview

This pull requests makes `OptionalCell` work with non-`Copy` types. This allows it to work without `#![feature(const_fn)]`. I rewrote a few methods that used `Cell::get` (which requires `T` to be `Copy`) that didn't need `T: Copy`.

`tock-cells` still needs `#![feature(const_fn)]` because `TakeCell` contains a mutable reference type, which cannot be initialized in `const` context without `const_fn`. Removing that dependency is possible but would require more `unsafe`, as we would need to replace `TockCell`'s reference with a raw pointer.

I have made the same semantic change using different syntax in #2010; I'd like input on which approach seems better.

This change helps migrate towards stable Rust (#1654).

### Testing Strategy

I commented out `#![feature(const_fn)]` and `pub mod take_cell;` in `libraries/tock-cells/src/lib.rs` then ran `cargo check` in `tock-cells` to verify I removed the dependency on `const_fn`. I reverted that change before committing.

I also ran `make allcheck` in the root of this repository.

### TODO or Help Wanted

I also sent #2010, which makes the same improvement in an equivalent but syntactically-different manner (using an `impl<T: Copy>` block rather than `where` clauses). I'd like feedback on which approach you find more maintainable.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/libraries/tock-cells/src/optional_cell.rs
+++ b/libraries/tock-cells/src/optional_cell.rs
@@ -1,16 +1,15 @@
 //! `OptionalCell` convenience type
 
 use core::cell::Cell;
-use core::marker::Copy;
 
 /// `OptionalCell` is a `Cell` that wraps an `Option`. This is helper type
 /// that makes keeping types that can be `None` a little cleaner.
 #[derive(Default)]
-pub struct OptionalCell<T: Copy> {
+pub struct OptionalCell<T> {
     value: Cell<Option<T>>,
 }
 
-impl<T: Copy> OptionalCell<T> {
+impl<T> OptionalCell<T> {
     /// Create a new OptionalCell.
     pub const fn new(val: T) -> OptionalCell<T> {
         OptionalCell {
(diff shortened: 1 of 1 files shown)


libraries/tock-cells/src/optional_cell.rs
```diff
@@ -55,38 +51,54 @@ impl<T: Copy> OptionalCell<T> {
...
+        self.value.set(value);
+        out
     }
 
     /// Returns the contained value or panics if contents is `None`.
-    pub fn expect(&self, msg: &str) -> T {
+    pub fn expect(&self, msg: &str) -> T
+    where
+        T: Copy,
+    {
```
hudson-ayers, line 83: This still seems to require the contained type is `Copy`?

jrvanwhy: Several methods of `OptionalCell` -- including `expect` -- require `T: Copy`.

Unless it panics, `expect` returns a copy of the value contained in the `OptionalCell` without modifying the `OptionalCell`, which definitely requires `T: Copy`.

hudson-ayers: This makes sense, I was just confused by the first sentence of the PR "This pull requests makes OptionalCell work with non-Copy types"

bradjc:
this: image

vs
#2010: image

I don't really have a preference.

jrvanwhy:
I personally have a slight preference for #2010, as it more closely mirrors `core::cell::Cell`'s implementation.

Issue #1662, opened 2020-03-05
Missed optimization opportunities with `static mut` data?


gendx:
Given that the topic of optimizing binary size came up in #1660, I want to mention that I recently observed what seems to be another missed optimization opportunity.

Looking at the nRF52840-DK's panic handler code, a `const` LED is chosen as a blinking signal, and therefore the corresponding GPIO is obtained from `nrf52840::gpio::PORT`. The LED index is compile-time, so I was expecting the array indexing to be optimized away.

https://github.com/tock/tock/blob/fbc863faf0c9615537ee52dcdccdfcb9204d2467/boards/nordic/nrf52840dk/src/io.rs#L89-L102

However, the disassembled code is the following:

```
00007d4e <core::panicking::panic_fmt>:
    7d4e:	b580      	push	{r7, lr}
    7d50:	466f      	mov	r7, sp
    7d52:	b084      	sub	sp, #16
    7d54:	e9cd 0102 	strd	r0, r1, [sp, #8]
    7d58:	f646 1074 	movw	r0, #26996	; 0x6974
    7d5c:	f2c0 0001 	movt	r0, #1
    7d60:	9001      	str	r0, [sp, #4]
    7d62:	f24c 2060 	movw	r0, #49760	; 0xc260
    7d66:	f2c0 0001 	movt	r0, #1
    7d6a:	9000      	str	r0, [sp, #0]
    7d6c:	4668      	mov	r0, sp
    7d6e:	f006 fe35 	bl	e9dc <rust_begin_unwind>
    7d72:	defe      	udf	#254	; 0xfe

0000e9dc <rust_begin_unwind>:
    e9dc:	b5f8      	push	{r3, r4, r5, r6, r7, lr}
    e9de:	af04      	add	r7, sp, #16
    e9e0:	4604      	mov	r4, r0
    e9e2:	f006 fd3b 	bl	1545c <<nrf5x::gpio::Port as core::ops::index::IndexMut<nrf5x::gpio::Pin>>::index_mut>
    e9e6:	4905      	ldr	r1, [pc, #20]	; (e9fc <rust_begin_unwind+0x20>)
    e9e8:	e9cd 0101 	strd	r0, r1, [sp, #4]
    e9ec:	a801      	add	r0, sp, #4
    e9ee:	9003      	str	r0, [sp, #12]
    e9f0:	a803      	add	r0, sp, #12
    e9f2:	4621      	mov	r1, r4
[...]

PR #2010, opened 2020-07-10
Remove OptionalCell's dependency on feature(const_fn) using `impl<T: Copy>`

jrvanwhy:
### Pull Request Overview

This pull requests makes `OptionalCell` work with non-`Copy` types. This allows it to work without `#![feature(const_fn)]`. I rewrote a few methods that used `Cell::get` (which requires `T` to be `Copy`) that didn't need `T: Copy`.

`tock-cells` still needs `#![feature(const_fn)]` because `TakeCell` contains a mutable reference type, which cannot be initialized in `const` context without `const_fn`. Removing that dependency is possible but would require more `unsafe`, as we would need to replace `TockCell`'s reference with a raw pointer.

I have made the same semantic change using different syntax in #2009 ; I'd like input on which approach seems better.

This change helps migrate towards stable Rust (#1654).

### Testing Strategy

I commented out `#![feature(const_fn)]` and `pub mod take_cell;` in `libraries/tock-cells/src/lib.rs` then ran `cargo check` in `tock-cells` to verify I removed the dependency on `const_fn`. I reverted that change before committing.

I also ran `make allcheck` and `make fmt` in the root of this repository.

### TODO or Help Wanted

I also sent #2009, which makes the same improvement in an equivalent but syntactically-different manner (using an `impl<T: Copy>` block rather than `where` clauses). I'd like feedback on which approach you find more maintainable.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/libraries/tock-cells/src/optional_cell.rs
+++ b/libraries/tock-cells/src/optional_cell.rs
@@ -1,16 +1,15 @@
 //! `OptionalCell` convenience type
 
 use core::cell::Cell;
-use core::marker::Copy;
 
 /// `OptionalCell` is a `Cell` that wraps an `Option`. This is helper type
 /// that makes keeping types that can be `None` a little cleaner.
 #[derive(Default)]
-pub struct OptionalCell<T: Copy> {
+pub struct OptionalCell<T> {
     value: Cell<Option<T>>,
 }
 
-impl<T: Copy> OptionalCell<T> {
+impl<T> OptionalCell<T> {
     /// Create a new OptionalCell.
     pub const fn new(val: T) -> OptionalCell<T> {
         OptionalCell {
@@ -33,10 +32,7 @@ impl<T: Copy> OptionalCell<T> {
     /// Insert the value of the supplied `Option`, or `None` if the supplied
     /// `Option` is `None`.
     pub fn insert(&self, opt: Option<T>) {
-        match opt {
-            Some(v) => self.set(v),
-            None => self.clear(),
-        }
+        self.value.set(opt);
     }
 
     /// Replace the contents with the supplied value.
@@ -55,12 +51,18 @@ impl<T: Copy> OptionalCell<T> {
 
     /// Check if the cell contains something.
     pub fn is_some(&self) -> bool {
-        self.value.get().is_some()
+        let value = self.value.take();
+        let out = value.is_some();
+        self.value.set(value);
+        out
     }
 
     /// Check if the cell is None.
     pub fn is_none(&self) -> bool {
-        self.value.get().is_none()
+        let value = self.value.take();
+        let out = value.is_none();
+        self.value.set(value);
+        out
     }
 
     /// Returns true if the option is a Some value containing the given value.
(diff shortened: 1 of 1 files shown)


bradjc:
It's too bad we can't mark just one file in tock-cells with the feature to make it clear if we ever made a change to OptionalCell that undid this improvement.

hudson-ayers:
This doesn't seem to have really removed the dependency on `#[feature(const_fn)]`, because any crate using `OptionalCell` is still required to use the feature in order for code to compile. For example, if I remove `#[feature(const_fn)]` from `chips/sam4l/src/lib.rs` and try to compile Imix, I get this list of errors:

```
error[E0723]: trait bounds other than `Sized` on const fn parameters are unstable
   --> chips/sam4l/src/acifc.rs:284:21
    |
284 |             client: Cell::new(None),
    |                     ^^^^^^^^^^^^^^^
    |
    = note: see issue #57563 <https://github.com/rust-lang/rust/issues/57563> for more information
    = help: add `#![feature(const_fn)]` to the crate attributes to enable

error[E0723]: trait bounds other than `Sized` on const fn parameters are unstable
   --> chips/sam4l/src/adc.rs:383:21
    |
383 |             client: OptionalCell::empty(),
    |                     ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #57563 <https://github.com/rust-lang/rust/issues/57563> for more information
    = help: add `#![feature(const_fn)]` to the crate attributes to enable

error[E0723]: trait bounds other than `Sized` on const fn parameters are unstable
   --> chips/sam4l/src/aes.rs:167:21
    |
167 |             client: OptionalCell::empty(),
    |                     ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #57563 <https://github.com/rust-lang/rust/issues/57563> for more information
    = help: add `#![feature(const_fn)]` to the crate attributes to enable

error[E0723]: trait bounds other than `Sized` on const fn parameters are unstable
   --> chips/sam4l/src/crccu.rs:244:21
    |
244 |             client: OptionalCell::empty(),
    |                     ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #57563 <https://github.com/rust-lang/rust/issues/57563> for more information
    = help: add `#![feature(const_fn)]` to the crate attributes to enable
...
```

(there are a bunch more).

One interesting bit is that in the ACIFC code where `Cell<Option<>>` is used we still see the error.

hudson-ayers:
#2009 has the same issue

jrvanwhy:
> This doesn't seem to have really removed the dependency on `#[feature(const_fn)]`, because any crate using `OptionalCell` is still required to use the feature in order for code to compile.

That's not strictly true, although I can believe that it is the case for most existing use cases. The following works in a crate with no unstable features enabled:

```
use tock_cells::optional_cell::OptionalCell;
static mut CELL: OptionalCell<bool> = OptionalCell::new(false);

fn main() {
    unsafe {
        println!("false: {}", CELL.expect("Error A"));
        CELL.set(true);
        println!("true: {}", CELL.expect("Error B"));
    }
}
```

> One interesting bit is that in the ACIFC code where `Cell<Option<>>` is used we still see the error.

I don't understand why we get that particular error. It appears to be something related to trait objects. Here's a smaller reproduction case:

```
use core::cell::Cell;

trait Bar {}

struct Foo {
    bar: Cell<Option<&'static dyn Bar>>,
}

impl Foo {
    pub const fn new() -> Foo {
        Foo { bar: Cell::new(None), }
    }
}

static mut TEST: Foo = Foo::new();

fn main() {}
```

That said, that error is in code that doesn't touch `OptionalCell`.

I'm okay with backing out this change if we feel that it makes things worse without any benefits, but as far as I can tell this PR is still a step in the right direction (even if it is a smaller step than we'd like).

hudson-ayers:
I don't think we should back this out, I agree it is still a step in the right direction! 

Thanks for that reproduction, it makes things more clear. You can actually remove the `Cell<>` as well.
It looks like the actual problem is that there is no stable way to initialize an `Option<&'static dyn Bar>` in a const function on stable. We need to do this a lot because of our broad reliance on global `static mut FOO` variables which require const initialization. This is sort of disappointing, because I was really starting to believe that atomics were the last blocker to stable with no immediate resolution in sight :/ 

Restructuring the kernel (and capsules) to not use trait objects is one resolution, the other is to not use global variables for any struct that has to hold a reference to a trait object. We have talked some about moving away from trait objects and globals (https://github.com/tock/tock/issues/1662 , https://github.com/tock/tock/issues/1761), but no particularly appealing approach seems to have arisen for either yet.

jrvanwhy:
> Restructuring the kernel (and capsules) to not use trait objects is one resolution, the other is to not use global variables for any struct that has to hold a reference to a trait object. We have talked some about moving away from trait objects and globals (#1662 , #1761), but no particularly appealing approach seems to have arisen for either yet.

I'm trying to avoid using trait objects (to the extent possible) in `libtock-rs`. If that turns out well then we can copy it in the kernel.

hudson-ayers:
I think there would be a lot of benefits to finding a good way to go without trait objects in the kernel, and this is just another reason that would be nice to have.

PR #2015, opened 2020-07-13
tock-registers: use UnsafeCell in writeable registers

### LINKED ISSUE #2005: tock-register-interface ReadWrite / WriteOnly potentially unsound without UnsafeCell?
Hi,

I was looking through the code in `tock-register-interface`, and found this line:

https://github.com/tock/tock/blob/f2b110ddbb9a962c4583087f8236a6e899a19b48/libraries/tock-register-interface/src/registers.rs#L164-L166

As I understand it, this is mutating arbitrary `Copy` data through an immutable reference. This may be a bit naive of me, but I think this brings up two safety issues.

First, according to [the nomicon](https://doc.rust-lang.org/reference/behavior-considered-undefined.html), mutating through an immutable reference like this is undefined behavior:

> ## Behavior considered undefined
> [...]
> - Mutating immutable data. All data inside a const item is immutable. Moreover, all data reached through a shared reference or data owned by an immutable binding is immutable, unless that data is contained within an UnsafeCell<U>.

The specific problem is that this `*mut T` is obtained from an `&T`.

Second, as far as I can tell, `ReadWrite` is `Sync`. According to the [ptr docs](https://doc.rust-lang.org/std/ptr/index.html#safety), `read_volatile` and `write_volatile` aren't thread safe:

> The precise rules for validity are not determined yet. The guarantees that are provided at this point are very minimal:
> [...]
> - All accesses performed by functions in this module are non-atomic in the sense of atomic operations used to synchronize between threads. This means it is undefined behavior to perform two concurrent accesses to the same location from different threads unless both accesses only read from memory. Notice that this explicitly includes `read_volatile` and `write_volatile`: Volatile accesses cannot be used for inter-thread synchronization.

I'm not sure how much tock kernel code running on microcontrollers would run into accidentally abusing it being `Sync`. But the undefined behavior from using a `*mut ` ptr gotten from an immutable reference seems like something to avoid.

Would it be reasonable to use an `UnsafeCell` here, instead? It would fix both issues, as with it one can avoid creating an `&T`, and it is non-`Sync`. It also has `#[repr(transparent)]`, so I don't think it'd change the behavior otherwise.

With that said, I'm a novice at embedded development, so there's likely something I missed. Thoughts?

daboross:
### Pull Request Overview

This pull request changes all writeable registers in `tock-register-interface` to use `UnsafeCell`-wrapped values, rather than just storing `T`.

In particular, `ReadWrite`, `WriteOnly`, `Aliased` and `InMemoryRegister` now wrap `UnsafeCell<T>`, and use `UnsafeCell::get` to retrieve mutable pointers rather than casting immutable references into mutable pointers. The intent is to avoid the undefined behavior of mutating through an immutable reference.

As a consequence, I had to remove `Copy` and `Clone` implementations for `InMemoryRegister`, and all changed registers are no longer `Sync`.

Fixes #2005.

### Testing Strategy

This pull request was tested by running the existing test suite. I ran `make ci-nosetup`, `make prepush`, and flashed a freshly built tock with these changes onto my `nrf52840dk` board.

### TODO or Help Wanted

This pull request still needs verification that removing `Sync` impls, and the `Copy` / `Clone` impls on `InMemoryRegisters` is OK.

There was some discussion of `Sync` in #2005, but none of the `Copy`/`Clone` changes. I'm not sure if this changes respects the purpose of `InMemoryRegister`, although it seems all code continues to compile without these impls?

In addition, this appears to introduce a number of clippy lint errors about not copying const values with inner mutability, despite that inner mutability being behind a pointer in `StaticRef`. I believe this is a false positive, and `clippy` is ignore the fact that the interior mutability is behind a pointer. But this doesn't resolve the errors.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/libraries/tock-register-interface/src/registers.rs
+++ b/libraries/tock-register-interface/src/registers.rs
@@ -59,6 +59,7 @@
 #![allow(clippy::suspicious_op_assign_impl)]
 #![allow(clippy::suspicious_arithmetic_impl)]
 
+use core::cell::UnsafeCell;
 use core::fmt;
 use core::marker::PhantomData;
 use core::ops::{Add, AddAssign, BitAnd, BitOr, BitOrAssign, Not, Shl, Shr};
@@ -117,7 +118,7 @@ pub trait TryFromValue<V> {
 // struct must be exactly the size of the `T`.
 #[repr(transparent)]
 pub struct ReadWrite<T: IntLike, R: RegisterLongName = ()> {
-    value: T,
+    value: UnsafeCell<T>,
     associated_register: PhantomData<R>,
 }
 
@@ -135,7 +136,7 @@ pub struct ReadOnly<T: IntLike, R: RegisterLongName = ()> {
 // struct must be exactly the size of the `T`.
 #[repr(transparent)]
 pub struct WriteOnly<T: IntLike, R: RegisterLongName = ()> {
-    value: T,
+    value: UnsafeCell<T>,
     associated_register: PhantomData<R>,
 }
 
@@ -148,7 +149,7 @@ pub struct WriteOnly<T: IntLike, R: RegisterLongName = ()> {
 // struct must be exactly the size of the `T`.
 #[repr(transparent)]
 pub struct Aliased<T: IntLike, R: RegisterLongName = (), W: RegisterLongName = ()> {
-    value: T,
+    value: UnsafeCell<T>,
     associated_register: PhantomData<(R, W)>,
 }
 
@@ -156,13 +157,13 @@ impl<T: IntLike, R: RegisterLongName> ReadWrite<T, R> {
     #[inline]
     /// Get the raw register value
     pub fn get(&self) -> T {
-        unsafe { ::core::ptr::read_volatile(&self.value) }
+        unsafe { ::core::ptr::read_volatile(self.value.get()) }
     }
 
     #[inline]
     /// Set the raw register value
     pub fn set(&self, value: T) {
-        unsafe { ::core::ptr::write_volatile(&self.value as *const T as *mut T, value) }
+        unsafe { ::core::ptr::write_volatile(self.value.get(), value) }
     }
 
     #[inline]
@@ -269,7 +270,7 @@ impl<T: IntLike, R: RegisterLongName> WriteOnly<T, R> {
     #[inline]
     /// Set the raw register value
     pub fn set(&self, value: T) {
-        unsafe { ::core::ptr::write_volatile(&self.value as *const T as *mut T, value) }
+        unsafe { ::core::ptr::write_volatile(self.value.get(), value) }
     }
 
     #[inline]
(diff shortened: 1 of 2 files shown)


daboross:
I've filed https://github.com/rust-lang/rust-clippy/issues/5796 to report the false positive clippy lint result. It seems there are a lot of places which trigger it, though, as this is a common pattern. Would it be appropriate to add a blanket allow for `clippy::borrow_interior_mutable_const` in the affected crates?

bradjc:
> I've filed [rust-lang/rust-clippy#5796](https://github.com/rust-lang/rust-clippy/issues/5796) to report the false positive clippy lint result. It seems there are a lot of places which trigger it, though, as this is a common pattern. Would it be appropriate to add a blanket allow for `clippy::borrow_interior_mutable_const` in the affected crates?

I think it is better to allow it in the run_clippy.sh script.

daboross:
> I think it is better to allow it in the run_clippy.sh script.

:+1:

I've done this.

libraries/tock-register-interface/src/registers.rs
```diff
@@ -429,29 +430,28 @@ impl<R: RegisterLongName> From<LocalRegisterCopy<u64, R>> for u64 {
 /// In memory volatile register.
 // To successfully alias this structure onto hardware registers in memory, this
 // struct must be exactly the size of the `T`.
-#[derive(Copy, Clone)]
 #[repr(transparent)]
 pub struct InMemoryRegister<T: IntLike, R: RegisterLongName = ()> {
-    value: T,
+    value: UnsafeCell<T>,
```
gendx, line 435: I still don't fully get the point of having non-MMIO mapped registers that use `UnsafeCell` and `volatile`. The only current Tock use cases are `chips/sam4l/src/usbc/mod.rs` and `chips/nrf52/src/usbd.rs`, yet I'm not 100% sure about the reason for them to use the volatile capabilities.

The [documentation](https://github.com/tock/tock/blob/master/libraries/tock-register-interface/README.md) for `InMemoryRegister` mentions sharing the memory with the hardware via DMA, but this seems a somewhat orthogonal problem to me. I don't remember seeing a wrapper type for objects (namely slices) shared via DMA, but without such a wrapper we surely get things wrong sometimes.

ppannuto: In the case of the SAM4L USB, that peripheral is perhaps best thought of as having MMIO with a dynamically assigned address. Configuration, status, and data for are stored in main memory, and can be written to by the CPU core **or** the peripheral. It must be volatile, as the hardware is allowed to change the underlying memory without notifying the CPU.

gendx: OK, this seems like a quite specific use case. I don't recall such requirements for `chips/nrf52/src/usbd.rs` though, so we should probably replace `InMemoryRegister` by a regular `LocalRegisterCopy` there.

PR #2022, opened 2020-07-14
kernel: process: use slice for app memory

bradjc:
### Pull Request Overview

Convert `load_processes` and `process::create()` to use a slice for process memory rather than a pointer and length. This ultimately ends up being safer as we can rely on more rust functions to manage the slice, and it reduces the portion of the code that has to be marked `unsafe`.

~~A while ago I changed load_processes and process::create to use a slice for flash rather than a raw pointer and length. I have always intended to do the same for memory. In general, using a slice seems to be less error prone than using a raw pointer and length. This would also address a small bug currently in load_processes, where the debug print assumes the app's memory region starts at the beginning of the remaining app memory, which isn't necessarily true.~~

~~This is my attempt. I've tried this a couple times now, and I think I'm close, but I can't get Rust to fully compile it. The issue seems to stem from memory being `mut`, and the process struct keeping a reference to its memory slice. To make rust happy with that I had to mark the memory slice as `'static`. That _almost_ works, but there are some ownership issues, so my thought was to pass the entire memory slice to process:create(), and then have process::create() return what it doesn't use. There is conveniently a `split_as_mut()` function, but it seems to lose the `'static` lifetime, causing an error again. That is where it stands now.~~

Compiling hail:

```
make
   Compiling tock-cells v0.1.0 (/Users/bradjc/git/tock/libraries/tock-cells)
   Compiling tock-registers v0.5.0 (/Users/bradjc/git/tock/libraries/tock-register-interface)
   Compiling hail v0.1.0 (/Users/bradjc/git/tock/boards/hail)
error[E0499]: cannot borrow `*remaining_memory` as mutable more than once at a time
   --> kernel/src/process.rs:207:21
    |
201 |                   if let Some((process, unused_memory)) = Process::create(
    |  _________________________________________________________-
202 | |                     kernel,
203 | |                     chip,
204 | |                     entry_flash,
...   |
207 | |                     remaining_memory,
    | |                     ^^^^^^^^^^^^^^^^ mutable borrow starts here in previous iteration of loop
208 | |                     fault_response,
209 | |                     i,
210 | |                 )? {
    | |_________________- argument requires that `*remaining_memory` is borrowed for `'static`

error[E0502]: cannot borrow `*app_memory` as immutable because it is also borrowed as mutable
    --> kernel/src/process.rs:1772:51
     |
1766 |         process.memory = app_memory;
     |         ---------------------------
     |         |                |
     |         |                mutable borrow occurs here
     |         assignment requires that `*app_memory` is borrowed for `'static`
...
1772 |         process.allow_high_water_mark = Cell::new(app_memory.as_ptr());
     |                                                   ^^^^^^^^^^ immutable borrow occurs here

error[E0502]: cannot borrow `*app_memory` as immutable because it is also borrowed as mutable
    --> kernel/src/process.rs:1773:50
     |
1766 |         process.memory = app_memory;
     |         ---------------------------
     |         |                |
     |         |                mutable borrow occurs here
     |         assignment requires that `*app_memory` is borrowed for `'static`
...
1773 |         process.original_allow_high_water_mark = app_memory.as_ptr();
     |                                                  ^^^^^^^^^^ immutable borrow occurs here

error: aborting due to 3 previous errors

Some errors have detailed explanations: E0499, E0502.
For more information about an error, try `rustc --explain E0499`.
error: could not compile `kernel`.
```

### Testing Strategy

todo

### TODO or Help Wanted

Anyone have any ideas?

Also there are a few old comments from past attempts that need I will clean up if we can find a working solution.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -143,7 +143,7 @@ pub fn load_processes<C: Chip>(
 ) -> Result<(), ProcessLoadError> {
     if config::CONFIG.debug_load_processes {
         debug!(
-            "Loading processes from flash=[{:#010X}:{:#010X}] into sram=[{:#010X}:{:#010X}]",
+            "Loading processes from flash={:#010X}-{:#010X} into sram={:#010X}-{:#010X}",
             app_flash.as_ptr() as usize,
             app_flash.as_ptr() as usize + app_flash.len() - 1,
             app_memory.as_ptr() as usize,
@@ -224,7 +224,7 @@ pub fn load_processes<C: Chip>(
             process_option.map(|process| {
                 if config::CONFIG.debug_load_processes {
                     debug!(
-                        "Loaded process[{}] from flash=[{:#010X}:{:#010X}] into sram=[{:#010X}:{:#010X}] = {:?}",
+                        "Loaded process[{}] from flash={:#010X}-{:#010X} into sram={:#010X}-{:#010X} = {:?}",
                         i,
                         entry_flash.as_ptr() as usize,
                         entry_flash.as_ptr() as usize + entry_flash.len() - 1,
@@ -1607,16 +1607,16 @@ impl<C: 'static + Chip> Process<'_, C> {
             if config::CONFIG.debug_load_processes {
                 if !tbf_header.is_app() {
                     debug!(
-                        "Padding in flash=[{:#010X}:{:#010X}]",
+                        "Padding in flash={:#010X}-{:#010X}",
                         app_flash.as_ptr() as usize,
-                        app_flash.as_ptr() as usize + app_flash.len()
+                        app_flash.as_ptr() as usize + app_flash.len() - 1
                     );
                 }
                 if !tbf_header.enabled() {
                     debug!(
-                        "Process not enabled flash=[{:#010X}:{:#010X}] process={:?}",
+                        "Process not enabled flash={:#010X}-{:#010X} process={:?}",
                         app_flash.as_ptr() as usize,
-                        app_flash.as_ptr() as usize + app_flash.len(),
+                        app_flash.as_ptr() as usize + app_flash.len() - 1,
                         process_name
                     );
                 }
@@ -1648,9 +1648,9 @@ impl<C: 'static + Chip> Process<'_, C> {
         {
             if config::CONFIG.debug_load_processes {
                 debug!(
-                    "[!] flash=[{:#010X}:{:#010X}] process={:?} - couldn't allocate MPU region for flash",
+                    "[!] flash={:#010X}-{:#010X} process={:?} - couldn't allocate MPU region for flash",
                     app_flash.as_ptr() as usize,
-                    app_flash.as_ptr() as usize + app_flash.len(),
+                    app_flash.as_ptr() as usize + app_flash.len() - 1,
                     process_name
                 );
             }
@@ -1704,7 +1704,7 @@ impl<C: 'static + Chip> Process<'_, C> {
                 // Failed to load process. Insufficient memory.
                 if config::CONFIG.debug_load_processes {
                     debug!(
-                        "[!] flash=[{:#010X}:{:#010X}] process={:?} - couldn't allocate memory region of size >= {:#X}",
+                        "[!] flash={:#010X}-{:#010X} process={:?} - couldn't allocate memory region of size >= {:#X}",
                         app_flash.as_ptr() as usize,
                         app_flash.as_ptr() as usize + app_flash.len(),
                         process_name,
@@ -1894,7 +1894,7 @@ impl<C: 'static + Chip> Process<'_, C> {
             _ => {
                 if config::CONFIG.debug_load_processes {
                     debug!(
-                        "[!] flash=[{:#010X}:{:#010X}] process={:?} - couldn't initialize process",
+                        "[!] flash={:#010X}-{:#010X} process={:?} - couldn't initialize process",
                         app_flash.as_ptr() as usize,
                         app_flash.as_ptr() as usize + app_flash.len(),
                         process_name
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -1706,7 +1706,7 @@ impl<C: 'static + Chip> Process<'_, C> {
                     debug!(
                         "[!] flash={:#010X}-{:#010X} process={:?} - couldn't allocate memory region of size >= {:#X}",
                         app_flash.as_ptr() as usize,
-                        app_flash.as_ptr() as usize + app_flash.len(),
+                        app_flash.as_ptr() as usize + app_flash.len() - 1,
                         process_name,
                         min_total_memory_size
                     );
@@ -1896,7 +1896,7 @@ impl<C: 'static + Chip> Process<'_, C> {
                     debug!(
                         "[!] flash={:#010X}-{:#010X} process={:?} - couldn't initialize process",
                         app_flash.as_ptr() as usize,
-                        app_flash.as_ptr() as usize + app_flash.len(),
+                        app_flash.as_ptr() as usize + app_flash.len() - 1,
                         process_name
                     );
                 }
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -1563,7 +1563,6 @@ fn exceeded_check(size: usize, allocated: usize) -> &'static str {
 }
 
 impl<C: 'static + Chip> Process<'_, C> {
-    #[inline(always)]
     pub(crate) unsafe fn create(
         kernel: &'static Kernel,
         chip: &'static C,
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -119,11 +119,19 @@ impl fmt::Debug for ProcessLoadError {
 /// is able to create its own `load_processes()` function and use that instead.
 ///
 /// Processes are found in flash starting from the given address and iterating
-/// through Tock Binary Format headers. Processes are given memory out of the
-/// `app_memory` buffer until either the memory is exhausted or the allocated
-/// number of processes are created, with process structures placed in the
-/// provided array. How process faults are handled by the kernel is also
-/// selected.
+/// through Tock Binary Format (TBF) headers. Processes are given memory out of
+/// the `app_memory` buffer until either the memory is exhausted or the
+/// allocated number of processes are created. A reference to each process is
+/// stored in the provided `procs` array. How process faults are handled by the
+/// kernel must be provided and is assigned to every created process.
+///
+/// This function is made `pub` so that board files can use it, but loading
+/// processes from slices of flash an memory is fundamentally unsafe. Therefore,
+/// we require the `ProcessManagementCapability` to call this function.
+///
+/// Returns `Ok(())` if process discovery went as expected. Returns a
+/// `ProcessLoadError` if something goes wrong during TBF parsing or process
+/// creation.
 pub fn load_processes<C: Chip>(
     kernel: &'static Kernel,
     chip: &'static C,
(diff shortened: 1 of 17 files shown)


hudson-ayers:
I pushed a commit that makes it work. You just have to show the compiler that you are reassigning `remaining_memory` on every iteration of the loop, so it knows you are never reborrowing the same slice (or if you are, it is because the function that borrowed it returned it). This requires returning a slice from `Process::create()` even in the case where you do not create a process. 

A major upside of this PR is that this allows us to narrow the scope of `unsafe` in `load_processes()` to just the call to the unsafe fn `Process::create()`, rather than encompassing almost the entire body of the function.

I didn't change over all the other boards to pass a slice yet, but I did test this with a few apps on Imix and it seems to work fine. The comments could probably still use a little work also.

kernel/src/process.rs
```diff
@@ -1510,17 +1511,17 @@ fn exceeded_check(size: usize, allocated: usize) -> &'static str {
 }
 
 impl<C: 'static + Chip> Process<'_, C> {
+    #[inline(always)]
```
bradjc, line 1514: What does this do?

hudson-ayers: Oh, sorry, that's left over from when I was trying a different way, you shouldn't need it.

bradjc:
Ok updated and rebased. I went through all of the comments and updated them as needed.

kernel/src/process.rs
```diff
@@ -1540,17 +1550,17 @@ fn exceeded_check(size: usize, allocated: usize) -> &'static str {
...
```
hudson-ayers, line 1553:
```suggestion
```

kernel/src/process.rs
```diff
@@ -119,127 +119,137 @@ impl fmt::Debug for ProcessLoadError {
...
     if config::CONFIG.debug_load_processes {
         debug!(
-            "Loading processes from flash={:#010X} into sram=[{:#010X}:{:#010X}]",
+            "Loading processes from flash=[{:#010X}:{:#010X}] into sram=[{:#010X}:{:#010X}]",
             app_flash.as_ptr() as usize,
-            app_memory_ptr as usize,
-            app_memory_ptr as usize + app_memory_size
+            app_flash.as_ptr() as usize + app_flash.len() - 1,
+            app_memory.as_ptr() as usize,
+            app_memory.as_ptr() as usize + app_memory.len() - 1
```
gendx, line 150: I don't see why this is changed to a range including the end, rather than excluding it. Since the start and length are generally both have a large alignment, this means the newly printed address will finish with a lot of `f` characters, which I think is less readable.

Also, this change isn't applied consistently across this file, some ranges are still exclusive-ended.

If it makes thinks clearer, the format should use `..` or `..=` rather than `:` so that it's clear. But again I think the open-ended range `..` will be clearer, as it is also generally more commonly used than `..=` in Rust.

kernel/src/process.rs
```diff
@@ -1579,27 +1588,28 @@ impl<C: 'static + Chip> Process<'_, C> {
...
                         app_flash.as_ptr() as usize,
-                        app_flash.as_ptr() as usize + app_flash.len(),
-                        process_name
+                        app_flash.as_ptr() as usize + app_flash.len()
                     );
                 }
                 if !tbf_header.enabled() {
                     debug!(
-                        "[!] flash=[{:#010X}:{:#010X}] process={:?} - process isn't enabled",
+                        "Note flash=[{:#010X}:{:#010X}] process={:?} not enabled",
```
gendx, line 1604: Nit: `Process {:?} at flash=[..] isn't enabled`? I don't see a reason to be verbose and have "Note" in this line.

kernel/src/process.rs
```diff
@@ -1579,27 +1588,28 @@ impl<C: 'static + Chip> Process<'_, C> {
...
         if !tbf_header.is_app() || !tbf_header.enabled() {
             if config::CONFIG.debug_load_processes {
                 if !tbf_header.is_app() {
                     debug!(
-                        "[!] flash=[{:#010X}:{:#010X}] process={:?} - process isn't an app",
+                        "Padding in flash=[{:#010X}:{:#010X}]",
                         app_flash.as_ptr() as usize,
-                        app_flash.as_ptr() as usize + app_flash.len(),
-                        process_name
+                        app_flash.as_ptr() as usize + app_flash.len()
```
gendx, line 1599: Here for example the flash range doesn't include the end (which I find clearer).

kernel/src/process.rs
bradjc, line 1604:
```suggestion
                        "Process not enabled flash=[{:#010X}:{:#010X}] process={:?}",
```

kernel/src/process.rs
bradjc: I guess I find it more precise to use the ']' bracket in the mathematical sense to mean inclusive.

ppannuto: I think in a programming / CS context, `]` unfortunately means exclusive (e.g. `"abc"[0:2]` is `"ab"` in Rust, Python; `.slice(0,2)` would be "ab" in Javascript..).

I looked up a few usages where people write memory maps. In written text, ARM seems to exclusively use inclusive ranges with `–`, i.e. `0x00000000-0x1FFFFFFF`.

I think we should maybe avoid `[]` given the ambiguity here. I'd support `..` or `–` or even `[)`

kernel/src/process.rs
```diff
@@ -119,127 +119,137 @@ impl fmt::Debug for ProcessLoadError {
...
-                    }
-                    procs[i] = process;
-                    process_memory_length
-                } else {
-                    0
-                }
-            } else {
-                // We are just skipping over this region in flash.
-                0
+                )?
```
ppannuto, line 222: I'm not sure we want to use the `?` operator here, as a non-fatal in one process (e.g. `MemoryAddressMismatch`) will stop later processes from loading.

bradjc: But if we don't, then the error doesn't get propagated to main.rs and the user won't know that the app wasn't loaded, putting us right back where we were.

ppannuto: Mhm.., makes sense. I think in the long run, we'll probably want the process load policy to be configurable by the board, but that's an enhancement beyond this PR.

kernel/src/process.rs
bradjc: So you are +1 on inclusive ranges, but not using `[]`?

gendx: Thanks for mentioning Python's `[0:2]` syntax (I guess Rust moved the vertical dots into horizontal as `[0..2]` instead). I'd support:
- `0x00000000..0x20000000` (exclusive)
- `[0x00000000:0x20000000)` (exclusive)
- `[0x00000000..0x20000000)` (exclusive)
- `[0x00000000:0x20000000[` (exclusive)
- `[0x00000000..0x20000000[` (exclusive)
- `0x00000000-0x1FFFFFFF` (inclusive)
- `0x00000000..=0x1FFFFFFF` (inclusive)

with a slight preference for the exclusive end, which I find more readable for highly-aligned addresses.

bradjc:
Thoughts on these debug prints?

kernel/src/process.rs
```diff
@@ -1691,7 +1691,7 @@ impl<C: 'static + Chip> Process<'_, C> {
                 // Failed to load process. Insufficient memory.
                 if config::CONFIG.debug_load_processes {
                     debug!(
-                        "[!] flash=[{:#010X}:{:#010X}] process={:?} - couldn't allocate memory region of size >= {:#X}",
+                        "[!] flash={:#010X}-{:#010X} process={:?} - couldn't allocate memory region of size >= {:#X}",
                         app_flash.as_ptr() as usize,
                         app_flash.as_ptr() as usize + app_flash.len(),
```
gendx, line 1696: This should be "len - 1".

kernel/src/process.rs
```diff
@@ -1881,7 +1881,7 @@ impl<C: 'static + Chip> Process<'_, C> {
             _ => {
                 if config::CONFIG.debug_load_processes {
                     debug!(
-                        "[!] flash=[{:#010X}:{:#010X}] process={:?} - couldn't initialize process",
+                        "[!] flash={:#010X}-{:#010X} process={:?} - couldn't initialize process",
                         app_flash.as_ptr() as usize,
                         app_flash.as_ptr() as usize + app_flash.len(),
```
gendx, line 1886: This should be "len - 1".

kernel/src/process.rs
```diff
@@ -1856,7 +1881,7 @@ impl<C: 'static + Chip> Process<'_, C> {
...
```
bradjc, line 1886:
```suggestion
                        app_flash.as_ptr() as usize + app_flash.len() - 1,
```

kernel/src/process.rs
```diff
@@ -1679,7 +1691,7 @@ impl<C: 'static + Chip> Process<'_, C> {
                 // Failed to load process. Insufficient memory.
                 if config::CONFIG.debug_load_processes {
                     debug!(
-                        "[!] flash=[{:#010X}:{:#010X}] process={:?} - couldn't allocate memory region of size >= {:#X}",
+                        "[!] flash={:#010X}-{:#010X} process={:?} - couldn't allocate memory region of size >= {:#X}",
                         app_flash.as_ptr() as usize,
                         app_flash.as_ptr() as usize + app_flash.len(),
```
bradjc, line 1696:
```suggestion
                        app_flash.as_ptr() as usize + app_flash.len() - 1,
```

kernel/src/process.rs
```diff
@@ -119,127 +119,137 @@ impl fmt::Debug for ProcessLoadError {
...
+                return Ok(());
+            }
+        };
+
+        // Now we can get a slice which only encompasses the length of flash
+        // described by this tbf header.  We will either parse this as an actual
+        // app, or skip over this region.
+        let entry_flash = remaining_flash
+            .get(0..entry_length as usize)
+            .ok_or(ProcessLoadError::NotEnoughFlash)?;
```
gendx, line 199: Instead of calling a `get` twice (here, and at the end of this function to update the `remaining_flash`), we can use the [slice::split_at](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at) method which is more idiomatic and does the bounds check once (I see that we already use `slice::split_at_mut` in this file).

The only caveat is that it would panic on bounds checking, so we have to do a check ourselves.

```rust
// We do a manual bounds check here to return an error rather than panicking in split_at.
if entry_length > remaining_flash.len() {
    return Err(ProcessLoadError::NotEnoughFlash);
}
let (entry_flash, rest) = remaining_flash.split_at(entry_length);
remaining_flash = rest;
```

bradjc: This only runs at boot, and I'm partial to using non panic APIs.

gendx: I agree that we need to report an error rather than panicking (as suggested in my code snippet), but apart from that `split_at` is more idiomatic, and doing twice the bounds check in separate sections of this function looks like an anti-pattern to me. With `split_at`, there's no risk of accidentally having `entry_flash` and the next `remaining_flash` overlap each other.

bradjc: If there is a split_at function that cannot panic I would agree with that being more idiomatic. But I think we need to push back against panic-able functions being idiomatic. While the compiler can omit the panic when it can verify it cannot occur, we 1) have to rely on the compiler to do that, and hope that it always can, and 2) changes to unrelated code can cause the panic to come back.

gendx: Can we move the following lines just after these ones, so that it's clear that the split is happening at the same offset?
```rust
        // Advance the flash slice for process discovery beyond this last entry.
        // This will be the start of where we look for a new process since Tock processes
        // are allocated back-to-back in flash.
        remaining_flash = remaining_flash
            .get(entry_flash.len()..)
            .ok_or(ProcessLoadError::NotEnoughFlash)?;
```

PR #2027, opened 2020-07-16
Remove duplicated code in cortex-m crates

bradjc:
### Pull Request Overview

Having copies of code in cortex-m3 and cortex-m4 is more and more unmaintainable, and with cortex-m7 coming it will only get worse. This moves duplicated code to a shared crate, and then specific crates can re-export it.

This is kind of a placeholder PR, more should be done here when other PRs are merged.

### Testing Strategy

todo

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.


arch/cortex-m/src/lib.rs
```diff
@@ -13,6 +13,255 @@ pub mod support;
 pub mod syscall;
 pub mod systick;
 
+#[cfg(all(target_arch = "arm", target_os = "none"))]
+#[inline(never)]
+unsafe fn kernel_hardfault_m3plus(faulting_stack: *mut u32) {
```
ppannuto, line 18:
```suggestion
unsafe fn kernel_hardfault_arm_v7m(faulting_stack: *mut u32) {
```

This is a name that better captures what cores can run this. (i.e. the new ARM M-23 cores are v8-m, which should use this just fine). The M0 is a v6-m.

bradjc:
Ok many updates to this PR, and some questions:

- The m3 implementation of generic_isr was different from the m4. Is that intentional? Or evidence that this PR is needed?
- The `switch_to_user` function feels like a bit of hack, and you can see my workaround. Is there a better way to get make the linker happy?
- Are there any more arch-specific functions that should be marked as such? I think implementing a cortex-m4f arch would help with finding these, but if there are any that we know of now we should go ahead and label them as such.

ppannuto:
>     * The `switch_to_user` function feels like a bit of hack, and you can see my workaround. Is there a better way to get make the linker happy?

I'm not quite sure I see what the hack was (though this diff is annoying to view on the web :/).

---

More generally, if we refactor this, I think it makes senses to lean into ARM's name for the material differences of these things: specifically, Cortex-M## is largely just a marketing name. From the perspective of a arch folder in an OS, we probably should be using the architecture names.

To that end, TIL that if you tack `--print cfg` onto rustc you can see the options, for hail:

```
$ rustc [...] --print cfg
target_arch="arm"
target_endian="little"
target_env=""
target_feature="dsp"
target_feature="mclass"
target_feature="v5te"
target_feature="v6"
target_feature="v6k"
target_feature="v6t2"
target_feature="v7"   <--------- Most of our M0 vs M3/4/7 stuff should really be gated on this cfg
target_has_atomic="16"
target_has_atomic="32"
target_has_atomic="8"
target_has_atomic="ptr"
target_has_atomic_load_store="16"
target_has_atomic_load_store="32"
target_has_atomic_load_store="8"
target_has_atomic_load_store="ptr"
target_os="none"
target_pointer_width="32"
target_vendor=""
```

bradjc:
> > ```
> > * The `switch_to_user` function feels like a bit of hack, and you can see my workaround. Is there a better way to get make the linker happy?
> > ```
> 
> I'm not quite sure I see what the hack was (though this diff is annoying to view on the web :/).

`cortex-m/syscall.rs` uses an external C function called "switch_to_user" which is resolved at linking time, rather than it being exposed through the rust module system or passed in as a parameter.

hudson-ayers:
rebased on current master and included the changes from #2043 .

PR #2036, opened 2020-07-18
BLE Capsule based on Rubble

daboross:
### Pull Request Overview

This adds an alternate capsule providing BLE support based on the [rubble](https://github.com/jonas-schievink/rubble) project. My intention is to provide pure-Rust advertising, and (insecure) connections to userspace apps, and to coexist as an alternative to the existing ble_advertising module.

This PR has:
- advertisements:
  - [x] a single app can trigger advertisements
- integration:
  - [x] create initial rubble capsule
  - hardware interaction:
    - [x] cordon off `rubble-nrf5x` dependency into `boards/nordic/nrf52_components`
    - [x] migrate from using `rubble-nrf5x` to using Tock's existing BLE radio driver

This PR does not have:
- connections:
  - [ ] can be established
  - [ ] can be controlled via syscall
  - [ ] have event callbacks
  - [ ] have attribute info callbacks
- receiving BLE data:
  - [ ] scan requests
  - [ ] scanning other advertisements
  - [ ] connection requests
- improvements to rubble which would improve integration
  - [ ] Make rubble hardware interface explicitly totally non-blocking
- [ ] vendored rubble with unnecessary dependencies entirely removed 

While this list is incomplete, I'm hoping the PR can be. The PR fully uses Tock's hardware access primitives, cordons off `rubble` into its own crate, and can run a minimal app which just transmits advertisements.

If it's alright, I'm planning on submitting support for connections, and further improvements based off of improvements in rubble, in future PRs.

I've submitted a PR adding a libtock-c interface to this capsule in https://github.com/tock/libtock-c/pull/103.

Depends on #2074 and #2077.

### Testing Strategy

I've applied the changes needed to enable the `rubble` capsule listed in `boards/nordic/nrf52_rubble/README.md`, then flashed the kernel & the example app from https://github.com/tock/libtock-c/pull/103 on a local nRF52840DK board. I verified everything runs correctly, and a second device picks up the advertisements.

Separately, I've run `make clean; make check` to ensure that without making the above changes, `rubble` is _not_ compiled as an indirect dependency of either the kernel or of `boards/nordic/nrf52840dk`. I also checked `Cargo.lock` to double check there was nothing besides the crate chain `nrf52_rubble` -> `tock_rubble` -> `rubble` which depends on `rubble` directly or indirectly by default.

### TODO or Help Wanted

Reviews!

Some of the latest things I've added to this PR mess around in radio code I don't fully understand yet. It works, but could definitely use more eyes.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


bradjc:
Very cool! Having good BLE support for the Nano 33 BLE board would be huge. The main hurdle of course is how to include the dependency One option might be to have a new crate in say boards/nordic called "tock-rubble" that would be the only crate that depends on rubble. It would have to implement some adapters to go from Tock interfaces to Rubble interfaces. The capsule would need to take in some "rubble engine" type that this adapter crate would provide. I think this is doable, albeit with some boilerplate code. I think having the external dependency only in boards/ would be easier to swallow.

alistair23:
It would be great not to make Rubble board specific (in the sense only 1 board supports it). I would like to use it with the Apollo3 Chip/board as well

daboross:
Moving rubble underneath the hil sounds reasonable. I want to ideally still keep hardware-agnostic and hardware-specific code separate, though. Would it be reasonable to move the current "hardware" interface to this new tock-rubble crate and implement that in `nrf52_components`?

I realize the point is to keep the rubble stuff removed from the rest of the codebase, but I don't want to limit its usefulness by keeping it associated with only nordic hardware, if it could just as easily be more than that.

bradjc:
I didn't mean to imply this would only work with nordic boards, only that containing the dependency to boards/ might be a good (short-term?) compromise between Tock's "no external dependencies" policy and Tock's desire to have a working BLE stack.

bradjc:
The core group talked about including external dependencies, and we generally agreed that containing them to a very specific crate (with /boards/ being a reasonable place to store it) that is documented and clear why that crate exists is a reasonable way to support rubble.

daboross:
Thanks for confirming that design can work!

I've been implementing it, and I've just pushed an update which adds the "rubble like" interface to `kernel::hil::rubble`, implements it in `boards/components/tock_rubble`, and removes references to the rubble crate from capsules.

If it's reasonable, I think the capsule, and the general structure of `tock_rubble` should be ready for review. Or, it could wait until I've also finished the hardware side using tock's existing BleRadio and finished fixing up the other TODOs.

I'm hoping that after removing the dependency on `rubble-nrf5x` for the hardware, fixing the remaining TODOs, and fixing anything that comes up in review, that this could be ready for merge. It doesn't yet have connection support, but I've gotten feedback that multiple, smaller PRs are better, and I could submit that as a second PR when it's ready to be reviewed separately from this base. Does that sound like a reasonable plan?

boards/components/tock_rubble/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+authors = ["Tock Project Developers <tock-dev@googlegroups.com>"]
+edition = "2018"
+
+[dependencies]
+kernel = { path = "../../../kernel" }
+
+[dependencies.rubble]
+version = "0.0.3"
+git = "https://github.com/daboross/rubble.git"
+rev = "601ef83eabcec6431d4cf27f0b8c714e0e1d6b87"
```
alistair23, line 13: This should probably be mainline Rubble before it's merged. It looks like only https://github.com/daboross/rubble/commit/601ef83eabcec6431d4cf27f0b8c714e0e1d6b87 still needs to be merged from your branch.

alevy: Looks like it's been merged!

daboross:
I have a WIP removal of the dependency on rubble's hardware crate pushed now. I'm running into a bug where I never get a `transmit_event` callback which I haven't tracked down yet, but besides that (and other minor bugs I'm sure I'll run into when it's fixed) the general form of the codebase is here.

There's one issue in the way I've implemented this pattern, which is that while `tock_rubble` is the only crate actually depending on rubble, `nrf52_components` still depends on `tock_rubble` directly, and thus brings `rubble` into the main dependency tree for `nrf52*` boards. If we keep `tock_rubble` board-independent, could it make sense to introduce a second crate, `nrf52_rubble`, which is only brought in by `nrf52840dk`? Or brought in by nothing, if we disable the `rubble` capsule by default?

daboross:
I've completed the move away from rubble's hardware interface!

If it's alright, I'm marking this PR as "ready to review".

A short list of things which are currently missing, and I think could fit in future PRs if people agree:
- any support for connections in the capsule interface
- support for receiving scan requests, or in general any kind of scanning - the Radio driver usage is currently transmit-only
- vendored rubble with even more dependencies removed

hudson-ayers:
Would it be possible to have the rubble crate not be in `boards/components` but instead at the top level of `boards/`? My concern with the current implementation is that Rubble is fetched from `crates.io` even when building a board like Imix that does not use Rubble at all. I think this is because `Imix` (and all other boards) depends on `components`, and `components` includes the Rubble dependency. 

However, it might be that any inclusion in the top level workspace leads to this, and we may need to make the Rubble dependency "optional" in Cargo.toml to prevent building it for boards that do not use it. Do you know of a clean way to enable this, or are there downsides I have not thought of?

> There are still things missing, but I think they could fit in future PRs. A short list:
>
> Of course, I'd be happy to work on adding them to this - it would just mean a larger PR.

I think this PR as it stands represents a good base to be reviewed, and putting other features in separate PRs makes sense.

> I've force pushed so this PR is sitting directly on top of #2074 and a more recent master, but I haven't yet cleaned up the commit history. Could you let me know if/when doing that would be most helpful?

I think you can worry about cleaning up the history after #2074 is merged, as that will require another rebase anyway.

daboross:
I can certainly move it out from under `components`, but I believe it is inclusion in the workspace which is causing this.

Unfortunately, I think it would still be downloaded if it's included in the crate graph at all, even if it's marked "optional". Specifically, `cargo` always downloads all optional (and even platform-dependent) dependencies so that it can build an accurate `Cargo.lock` - even for crates disabled in the current configuration.

We could fix this by removing `rubble-nrf5x` from the workspace entirely? The disadvantage would be that `cargo check` on the workspace no longer checks `rubble-nrf5x` and `tock_rubble`, so I think it would want to be added as a separate CI check. If you think this would be a better solution, I can definitely change this.

> I think this PR as it stands represents a good base to be reviewed, and putting other features in separate PRs makes sense.

> I think you can worry about cleaning up the history after #2074 is merged, as that will require another rebase anyway.

:+1: Thanks for responding to these! Sounds good - I'll clean up the commit log if/when the two PRs this depends on are merged.

boards/components/tock_rubble/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+    }
+}
+
+impl<'a, A, R> RubbleBleRadio<'a, A, Implementation<'a, A, R>> for BleRadioWrapper<'a, R>
+where
+    A: Alarm<'a>,
+    R: RubbleDataDriver<'a> + BleAdvertisementDriver<'a> + BleConfig + 'a,
+{
+    fn accept_cmd(&mut self, _cmd: RadioCmd) {
+        // TODO: The current NRF52840 ble_radio driver doesn't support
```
alistair23, line 205: This shouldn't be NRF specific

boards/components/tock_rubble/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+                rubble_hil::NextUpdate::At(rubble_hil::Instant::from_raw_micros(time.raw_micros()))
+            }
+            NextUpdate::Disable => rubble_hil::NextUpdate::Disable,
+            NextUpdate::Keep => rubble_hil::NextUpdate::Keep,
+        }
+    }
+    fn queued_work(&self) -> bool {
+        self.0.queued_work
+    }
+    fn into_radio_cmd(self) -> RadioCmd {
```
alistair23, line 164: New lines between functions.

boards/components/tock_rubble/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+    type Responder = ResponderWrapper<'a, A, R>;
+    type Cmd = CmdWrapper;
+    type PacketQueue = refcell_packet_queue::RefCellQueue;
+    fn get_device_address() -> rubble_hil::DeviceAddress {
+        R::get_device_address()
+    }
+    fn rx_packet_queue() -> &'static Self::PacketQueue {
+        &RX_PACKET_QUEUE
+    }
+    fn tx_packet_queue() -> &'static Self::PacketQueue {
```
alistair23, line 65: New lines between the functions.

boards/components/tock_rubble/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+}
+
+impl<'a, R> Transmitter for BleRadioWrapper<'a, R>
+where
+    R: RubbleDataDriver<'a> + BleAdvertisementDriver<'a> + BleConfig + 'a,
+{
+    fn tx_payload_buf(&mut self) -> &mut [u8] {
+        // TODO: To fully comply with what the rubble stack expects, we would
+        // need to block here until the transmission is done. This should be
+        // fixed on rubble's side.
```
alistair23, line 244: Is there an issue or PR tracking this?

boards/components/tock_rubble/src/refcell_packet_queue.rs
```diff
@@ -0,0 +1,119 @@
+//! Safe, non-interrupt-driven packet queue written using RefCells.
+//!
+//! Copied from https://github.com/jonas-schievink/rubble/pull/137/files#diff-80283f6441097bbff4aad2c54957a72a.
```
alistair23, line 3: This is 0-clause BSD code and not Apache/MIT like the rest of Tock. Legally I think that's ok, but it should be documented here.

boards/nordic/nrf52_rubble/src/lib.rs
```diff
@@ -0,0 +1,83 @@
+//! Component for Rubble radio on nRF52 based platforms.
+//!
+//! Usage
+//! -----
+//! ```rust
+//! let ble_radio =
+//!     nrf52_rubble::RubbleComponent::new(board_kernel, &nrf52840::ble_radio::RADIO, mux_alarm)
+//!        .finalize(());
+//! ```
```
alistair23, line 9: Couldn't this be a generic component instead?

capsules/src/rubble.rs
```diff
@@ -0,0 +1,274 @@
...
+        RubbleLinkLayer, RubblePacketQueue, RubbleResponder,
+    },
+    time::Alarm,
+};
+use kernel::{AppId, AppSlice, ReturnCode, Shared};
+
+use crate::driver;
+
+// Syscall driver number.
+
```
alistair23, line 51: No newline

hudson-ayers:
> I can certainly move it out from under `components`, but I believe it is inclusion in the workspace which is causing this.
> 
> Unfortunately, I think it would still be downloaded if it's included in the crate graph at all, even if it's marked "optional". Specifically, `cargo` always downloads all optional (and even platform-dependent) dependencies so that it can build an accurate `Cargo.lock` - even for crates disabled in the current configuration.
> 
> We could fix this by removing `rubble-nrf5x` from the workspace entirely? The disadvantage would be that `cargo check` on the workspace no longer checks `rubble-nrf5x` and `tock_rubble`, so I think it would want to be added as a separate CI check. If you think this would be a better solution, I can definitely change this.

We talked about this briefly on last weeks core call, and the conclusion was that it is fine for the crate to be downloaded even if it is not being used by the board that is being compiled. However, it should be the case that out-of-tree boards can depend on the Kernel (e.g. chips / capsules / kernel / libraries) without having to download the dependency. My understanding is that this should not be an issue even if the dependency is part of the workspace, as out of tree boards depend on individual crates within the Tock kernel rather than the workspace as a whole. 

Given that `components` are likely useful for out of tree boards, I think it is still preferable to move the dependency out of `components` into a different directory which out of tree boards will not have to depend on.

bradjc:
Replaced by #2233

PR #2038, opened 2020-07-20
arch/rv32i: pmp: Make the PMP struct as long as required

alistair23:
### Pull Request Overview

RISC-V allows up to 64 PMP regions, but most SoCs only support a smaller
number (currenlty 4 or 8). Allocating an array of all 64 regions for a
SoC that will only ever use 4 seems very wasteful. At the same time
restricting this amount will hinder devices that can support all 64.

Luckily Rust has a solution with generics. Unfortunately we can't have
generic arrays (there is an external crate that helps). I did manage to
get something almost working, but it was then impossible to iterate or
index into the array.

This patch converts the PMPConfig to be a generic type with specific
implementations created by macros. This allows us to easily support
a dynamic (compile time) number of regions.

### Testing Strategy

Testing with libtock-rs in QEMU on HiFive1 board.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -1,13 +1,16 @@
 //! Implementation of the physical memory protection unit (PMP).
 
+#[macro_export]
+macro_rules! PMPConfigMacro {
+    ( $x:expr ) => {
 use core::cell::Cell;
 use core::cmp;
 use core::fmt;
 use kernel::common::cells::OptionalCell;
 
-use crate::csr;
-use kernel;
+use rv32i::csr;
 use kernel::common::cells::MapCell;
+use kernel::common::registers;
 use kernel::common::registers::register_bitfields;
 use kernel::mpu;
 use kernel::AppId;
@@ -40,7 +43,7 @@ register_bitfields![u8,
 #[derive(Copy, Clone)]
 pub struct PMPRegion {
     location: (*const u8, usize),
-    cfg: tock_registers::registers::FieldValue<u8, pmpcfg::Register>,
+    cfg: registers::FieldValue<u8, pmpcfg::Register>,
 }
 
 impl fmt::Display for PMPRegion {
@@ -130,8 +133,6 @@ pub struct PMPConfig<N: PMPConfigType> {
     app_region: OptionalCell<usize>,
 }
 
-macro_rules! PMPConfigMacro {
-    ( $x:expr ) => {
 impl PMPConfigType for [Option<PMPRegion>; $x] {}
 
 impl Default for PMPConfig<[Option<PMPRegion>; $x]> {
@@ -503,6 +504,3 @@ impl kernel::mpu::MPU for PMPConfig<[Option<PMPRegion>; $x]> {
         }
     };
 }
-
-PMPConfigMacro!(4);
-PMPConfigMacro!(8);
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -116,9 +116,11 @@ impl PMPRegion {
     }
 }
 
+pub trait PMPConfigType: Default + Copy + Clone + Sized {}
+
 /// Struct storing region configuration for RISCV PMP.
-pub struct PMPConfig {
-    regions: [Option<PMPRegion>; 32],
+pub struct PMPConfig<N: PMPConfigType> {
+    regions: N,
     total_regions: usize,
     /// Indicates if the configuration has changed since the last time it was written to hardware.
     is_dirty: Cell<bool>,
@@ -128,12 +130,18 @@ pub struct PMPConfig {
     app_region: OptionalCell<usize>,
 }
 
-impl Default for PMPConfig {
+macro_rules! PMPConfigMacro {
+    ( $x:expr ) => {
+impl PMPConfigType for [Option<PMPRegion>; $x] {}
+
+impl Default for PMPConfig<[Option<PMPRegion>; $x]> {
     /// number of regions on the arty chip
-    fn default() -> PMPConfig {
+    fn default() -> Self {
         PMPConfig {
-            regions: [None; 32],
-            total_regions: 8,
+            regions: [None; $x],
+            // As we use the PMP TOR setup we only support half the number
+            // of regions as hardware supports
+            total_regions: $x / 2,
             is_dirty: Cell::new(true),
             last_configured_for: MapCell::empty(),
             app_region: OptionalCell::empty(),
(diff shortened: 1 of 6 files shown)


bradjc:
For me, this falls on the side of too clunky to include. Having to wrap everything in a macro makes code too hard to read, in my opinion.

alistair23:
Fair, is it that hard to read though? It also reduces memory and stack usage considerably. The HiFive1 cant' run apps without this change for example.

Unfortunately we can't do generic arrays without including this: https://github.com/fizyk20/generic-array

hudson-ayers:
> Unfortunately we can't do generic arrays without including this: https://github.com/fizyk20/generic-array

We can do generic arrays with const generics, though that is currently an unstable feature. I got generic IPC arrays working with const generics here: https://github.com/hudson-ayers/tock/commit/9b5f7cc57132e4446b884605602e68f8ec04ced0 , but we decided not to merge that because of the instability of const generics. 

Still, I would prefer we not go through all of the effort of pulling in external dependencies for something which will be possible without it; hopefully by the end of 2020 (https://without.boats/blog/shipping-const-generics/).

bradjc:
I think Alistair buried the lede on this one. Hifive1 is broken right now, so I'm much more motivated to incorporate this. How hard is it to use the const generics feature to get this to work?

alistair23:
I did have a quick look at using it, but it doesn't seem that close to done and then it would add another step away from using stable Rust (which I wanted to avoid).

bradjc:
Well this PR serves as an escape plan, and if this use of const generics is the only thing keeping us from stable rust I am documenting now I will merge the macro version immediately. :)

alistair23:
It looks like there is a fair number of bugs with const generics (https://github.com/rust-lang/rust/labels/A-const-generics) and some key parts missing (https://github.com/rust-lang/rust/issues/60471) so I still prefer the mild ugliness of macros then using something that still seems to unstable.

We would also probably need `const_in_array_repeat_expressions` as well, so that's two unstable features.

hudson-ayers:
while there are bugs with const generics, my understanding is they are mostly bugs with some of the more complex aspects of the feature rather than the more basic subset we would need here. Unfortunately the working, well tested subset of const generics has not yet been separated out into its own feature, but basic use of const generics for array lengths has been used in the std library for the last year or so. 

Still I didn't realize that hi five was broken, so I think it is fine to rush the macro approach in if needed - I would just want to somehow make sure we switch over to the idiomatic approach once const generics are stable.

bradjc:
Avoiding adding features is good, but we do want to make sure we are using macros in reasonable cases. Here the macro is a workaround for a feature that rust may provide someday, but the macro also hides some complexity in RISC-V hardware: namely that the PMP is not a fully fixed peripheral as defined by the architecture, instead it can be different sizes.

On today's OT call we discussed this, and where we ended up was that expanding the macro to encompass the entire pmp.rs file, and moving the use of the macro to the chip crates might be a good combination of macro use and clarity.

Since the PMP is _not_ entirely specified by the architecture, I would argue it _does_ make sense for the chip crate to implement it since different chips can have different PMP sizes. However, the pmp usage is shared, and duplicating `pmp.rs` would be a maintenance burden. Providing a macro allows for chips to define the specific pmp code for the size they need, addressing the memory and stack problem. However, since the macro is essentially doing a simple find and replace, it is pretty straightforward, and the usage would give a lot of hints as to what it is doing.

I think a clarifying comment would help, particularly to specify that if rust does stabilize this feature we should use it instead.

The current macro is similar, but it does rely on the compiler to do the optimization to remove the implementation for the size(s) that are not used. It also is a little bit harder to read, since it is not initially obvious why the macro gets called multiple times.

ppannuto:
> Unfortunately the working, well tested subset of const generics has not yet been separated out into its own feature, but basic use of const generics for array lengths has been used in the std library for the last year or so.

It looks like there actually is an active effort to stabailize the stable parts of const generics: https://internals.rust-lang.org/t/stabilizing-a-const-generics-mvp/12727

I think our use case here is comfortably in the window of things that will be stabilized on a short-term time frame. With that in mind, I think I would advocate for jumping directly to the generics (with the transient feature), rather than a short-lived, more complex macro implementation.

bradjc:
I guess the risk there is that it turns out our use case is somehow different and unsupported in what gets put on the path to stabilization.

Given that the hifive1 is currently broken, I think I would prefer to merge the macro version, and then replace it with a nightly feature if rust provides a specific feature that is clearly on the path to stabilization.

ppannuto:
Fair enough; let's move forward with macros for now

alistair23:
I don't think this use case is in the first round.

From https://without.boats/blog/shipping-const-generics/:

> That means the signed and unsigned integer types, booleans, and chars. For now, no compound or user-defined types will be allowed

From https://internals.rust-lang.org/t/stabilizing-a-const-generics-mvp/12727:

> Due to this issue, this may need to be restricted to just integral primitive types (ie the int types, bool, and char). Other types would probably be a nearer term addition beyond the MVP. https://github.com/rust-lang/compiler-team/issues/323

Even on top of that `PMPRegion` does not implement `Eq` or `PartialEq` so I think we are still a way out.

alistair23:
I have updated this based on the discussions.

arch/rv32i/src/pmp.rs
```diff
@@ -128,20 +133,22 @@ pub struct PMPConfig {
...
+
+impl Default for PMPConfig<[Option<PMPRegion>; $x]> {
     /// number of regions on the arty chip
-    fn default() -> PMPConfig {
+    fn default() -> Self {
         PMPConfig {
-            regions: [None; 32],
-            total_regions: 8,
+            regions: [None; $x],
+            total_regions: $x / 2,
```
hudson-ayers, line 143: Could there be a comment indicating why this is `$x / 2`?

hudson-ayers:
> I don't think this use case is in the first round.
> 
> From https://without.boats/blog/shipping-const-generics/:
> 
> > That means the signed and unsigned integer types, booleans, and chars. For now, no compound or user-defined types will be allowed
> 
> From https://internals.rust-lang.org/t/stabilizing-a-const-generics-mvp/12727:
> 
> > Due to this issue, this may need to be restricted to just integral primitive types (ie the int types, bool, and char). Other types would probably be a nearer term addition beyond the MVP. [rust-lang/compiler-team#323](https://github.com/rust-lang/compiler-team/issues/323)
> 
> Even on top of that `PMPRegion` does not implement `Eq` or `PartialEq` so I think we are still a way out.

I think that if you wanted to implement this with const generics, you would just keep everything the same as it is before this PR, but with this change:
```
pub struct PMPConfig<const N: usize> {
    regions: [Option<PMPRegion>; N],
    total_regions: N/2,
...
```
This way, the only generic is an int type, which is supported by the MVP.

(Not that this matters for now, I agree with using the macro approach until the MVP is separated out)

arch/rv32i/src/pmp.rs
bradjc: I am adding comments.

alistair23: I didn't see this, I just pushed an update which doesn't remove the comments.

bradjc:
My updates:
- Added comments
- Make the regions array $x/2 and removed the total_regions variable.
- Moved the pmp instantiation to its file for the arty platform.

PR #2039, opened 2020-07-24
chips: Only report a PLIC interrupt as pending if enabled

alistair23:
### Pull Request Overview

There is currently no change to functionality as all PLIC interrupts are
enabled in Tock.

As discussed today this fixes the problem where we will service disabled interrupts before enabled interrupts. This only fixes the problem for RISC-V PLIC systems.

### Testing Strategy

Run Tock on QEMU and OpenTitan hardware.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/chips/e310x/src/plic.rs
+++ b/chips/e310x/src/plic.rs
@@ -89,7 +89,13 @@ pub unsafe fn complete(index: u32) {
 pub unsafe fn has_pending() -> bool {
     let plic: &PlicRegisters = &*PLIC_BASE;
 
-    plic.pending.iter().fold(0, |i, pending| pending.get() | i) != 0
+    for (i, pending) in plic.pending.iter().enumerate() {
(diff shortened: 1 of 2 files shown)


bradjc:
But we _should_ be disabling them, so that others can fire and the just fired interrupt won't re-fire. I think this is a weakness it at least the e310 implementation.

alistair23:
That seems like a different problem.

I have a patch set to do this. The problem is that we need the `Chip` state in the `handle_interrupt()` function called from the traps

alistair23:
Closing this in favour of: https://github.com/tock/tock/pull/2040

I initially thought this patch would be part of https://github.com/tock/tock/pull/2040, but due to the way the PLIC claiming works we don't need to keep track of what is enabled, just what we have already claimed but not completed.

PR #2040, opened 2020-07-24
chips: plic: Disable only specific interrupts in the top half

alistair23:
### Pull Request Overview

When handling an interrupt only disable the specific interrupt instead
of all external interrupts.

See individual commits for more information.

### Testing Strategy

Run Tock on QEMU and OpenTitan hardware.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/chips/e310x/src/chip.rs
+++ b/chips/e310x/src/chip.rs
@@ -4,6 +4,7 @@ use core::fmt::Write;
 use kernel;
 use kernel::debug;
 use kernel::hil::time::Alarm;
+use kernel::Chip;
 use rv32i;
 use rv32i::csr::{mcause, mie::mie, mip::mip, CSR};
 use rv32i::PMPConfigMacro;
@@ -96,7 +97,9 @@ impl<'a, A: 'static + Alarm<'static>, I: InterruptService<()> + 'a> E310x<'a, A,
             if !self.plic_interrupt_service.service_interrupt(interrupt) {
                 debug!("Pidx {}", interrupt);
             }
-            self.plic.complete(interrupt);
+            self.atomic(|| {
+                self.plic.complete(interrupt);
+            });
         }
     }
 }
@@ -149,7 +152,7 @@ impl<'a, A: 'static + Alarm<'static>, I: InterruptService<()> + 'a> kernel::Chip
     }
 
     fn has_pending_interrupts(&self) -> bool {
-        self.plic.has_pending()
+        self.plic.get_saved_interrupts().is_some()
     }
 
     fn sleep(&self) {
--- a/chips/e310x/src/chip.rs
+++ b/chips/e310x/src/chip.rs
@@ -93,7 +93,7 @@ impl<'a, A: 'static + Alarm<'static>, I: InterruptService<()> + 'a> E310x<'a, A,
     }
 
     unsafe fn handle_plic_interrupts(&self) {
-        while let Some(interrupt) = self.plic.next_pending() {
+        while let Some(interrupt) = self.plic.get_saved_interrupts() {
             if !self.plic_interrupt_service.service_interrupt(interrupt) {
                 debug!("Pidx {}", interrupt);
             }
@@ -135,13 +135,13 @@ impl<'a, A: 'static + Alarm<'static>, I: InterruptService<()> + 'a> kernel::Chip
             if mip.is_set(mip::mtimer) {
                 self.timer.handle_interrupt();
             }
-            if mip.is_set(mip::mext) {
+            if self.plic.get_saved_interrupts().is_some() {
                 unsafe {
                     self.handle_plic_interrupts();
                 }
             }
 
-            if !mip.matches_any(mip::mext::SET + mip::mtimer::SET) {
+            if !mip.matches_any(mip::mtimer::SET) && self.plic.get_saved_interrupts().is_none() {
                 break;
             }
         }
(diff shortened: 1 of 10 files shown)


alistair23:
This failure is actually a QEMU bug, I'll send patches out but we are currently in a release freeze so it could take some time until it's merged.

bradjc:
I'm nervous about the "saved interrupt" part. We originally were doing that with cortex-m and the NVIC, and it lead to some subtle bugs. Reliability improved after we switched to using the hardware to store state. But, I'm not sure if that issue apply here.

phil-levis:
> I'm nervous about the "saved interrupt" part. We originally were doing that with cortex-m and the NVIC, and it lead to some subtle bugs. Reliability improved after we switched to using the hardware to store state. But, I'm not sure if that issue apply here.

The bug in Cortex-M, IIRC, was using a FIFO queue. This meant that you could lose an interrupt if the queue overflowed. The current Cortex-M approach of has the outstanding bug that bottom halves are invoked on disabled interrupts.

chips/e310x/src/chip.rs
```diff
@@ -167,7 +169,18 @@ unsafe fn handle_interrupt(intr: mcause::Interrupt) {
             CSR.mie.modify(mie::mtimer::CLEAR);
         }
         mcause::Interrupt::MachineExternal => {
+            // We recieved an interrupt, disable interrupts while we handle them
             CSR.mie.modify(mie::mext::CLEAR);
+
+            // Claim the interrupt, unwrap() as we know an interrupt exists
```
phil-levis, line 175: Would it make sense to merge the arrays, rather than claim the next?

I.e., is it possible to add multiple interrupts at once to the saved set, based on which are pending? I don't think this is a big deal, just wondering.

alistair23: Do you mean loop over `PLIC.next_pending()` until there are no more pending interrupts?

alistair23:
Ping

chips/e310x/src/chip.rs
phil-levis: I'd been thinking more of merging the pending bits in the interrupt pending bits block with this software state using a logical OR. But when I suggested this, I didn't realize that the block is 32 registers (128 bytes) long. The benefit from reducing the number of interrupts handled when there are multiple pending would in most cases be small compared to the cost of having to merge 32 words on each interrupt.

However, I do think that looping over PLIC.next_pending() a few times would be good; if there happen to be a few pending interrupts, amortizing the handling cost across them would be a big savings.

alistair23: Done! We now loop over `next_pending()` in the interrupt handler.

alistair23:
I have addressed all the comments, let me know what you think.

alistair23:
I have updated this after the discussion today so that interrupts are only completed atomically.

@phil-levis, @bradjc any more comments?

chips/e310x/src/chip.rs
```diff
@@ -163,7 +169,25 @@ unsafe fn handle_interrupt(intr: mcause::Interrupt) {
...
+            // Once claimed this interrupt won't fire until it's completed
+            // NOTE: The interrupt is no longer pending in the PLIC
+            loop {
+                let interrupt = PLIC.next_pending();
+
+                match interrupt {
+                    Some(irq) => {
+                        PLIC.save_interrupt(irq);
```
hudson-ayers, line 183: I think this use of the global PLIC struct is unsound. Even with the guarantee that the main thread will not modify the array in a critical section. We call `save_interrupt()` from one thread (interrupt context) then read the saved state from another thread (non-interrupt context). `Cell` is not sufficient to ensure correct results will be read in the non-interrupt context after an interrupt fires.

Cell is not intended as a concurrency primitive. For the following code snippet

```rust
let cell: Cell<u32> = Cell::new(0);

let b = cell.get();
// various stuff that does not touch `cell`
let c = cell.get();
```
the compiler is allowed to assume that cell has not changed between the two loads, and elide the load that would happen when assigning `c` and instead use the result from the load when `b` was assigned. If an interrupt arrives in the interim and modified the cell, the value assigned to `c` would be incorrect.

In the case of saved interrupts, I think this could lead to extremely nefarious bugs.

This does not even consider the possibility of preemption of interrupt handlers -- which may not be possible in Tock on the HiFive? I am not sure? -- but if possible would lead to additional opportunities for bugs here as `save_interrupt` is not called atomically.

alistair23: We only ever read and write the `saved` `Cell` with interrupts disabled (I just fixed a bug where interrupts could be enabled too early), shouldn't that be enough to ensure it's not unsound?

chips/e310x/src/chip.rs
```diff
@@ -177,7 +201,7 @@ unsafe fn handle_interrupt(intr: mcause::Interrupt) {
 /// For the e310 this gets called when an interrupt occurs while the chip is
 /// in kernel mode. All we need to do is check which interrupt occurred and
 /// disable it.
```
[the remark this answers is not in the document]
alistair23: Updated

chips/e310x/src/chip.rs
hudson-ayers: That protects against the potential for memory errors from the `Cell` being modified partway through a nonatomic read/write. But I don't think it protects against the compiler reordering or eliding loads, because AFAIK enabling/disabling interrupts is not a memory fence. If interrupts are disabled during all accesses to the `saved` `Cell`, I think using `VolatileCell` would be sufficient. Alternatively, using an atomic type would work, but that has other issues with forcing us to stay on nightly for `earlgrey` (which may be a problem we want to postpone to later, idk).

hudson-ayers: After some additional thought: If you never read or write the `saved` `Cell` with interrupts enabled, this may be okay. But I think there are several calls right now that read the cell without disabling interrupts first. 

Also, I generally don't like that future modifications to this code would have to be aware that the PLIC struct is shared beyond the kernel thread into interrupt context - I don't think we currently share Rust structs between the kernel thread and interrupt context anywhere else in Tock, and doing so adds all sorts of implicit constraints on how you can interact with those objects safely (e.g. having to disable interrupts when accessing fields that can change) which might easily be missed in future modifications to this code.

alistair23: Ok, I have pushed an update that uses `VolatileCell` (good call on that one). I have changed some of the PLIC functions to be `unsafe` so that no one calls them with interrupts still enabled.

bradjc (approved):

hudson-ayers (dismissed):
Thanks for making those changes

bradjc:
Oh I added a merge conflict...my bad.

alistair23:
I just pushed a rebase

bradjc (approved):

alistair23:
It seems like this fails on QEMU, as the claim register is always something so we get stuck in a loop. This is a QEMU bug in the Ibex specific PLIC, sending a patch now.

Issue #2030, opened 2020-07-17
Release 1.6


bradjc:
## Release Date: ~~Summer 2020~~ Fall 2020

Milestones to be included in the release:

- [x] Scheduler options #1441 
- [x] Alarm HIL updates #2089
- [x] Fixed address support
  - [x] New TBF header
  - [x] elf2tab and tockloader support
  - [x] process.rs can choose specific memory addresses

## Testing

To test the release we use the following applications and run them on all applicable boards and ensure they execute successfully. This list should be copied for each platform and will be tested once a release candidate is ready.

- All Boards
  - [ ] `examples/sensors`
    - sense_to_console: Application that samples a sensor and outputs values to console
  - [ ] `examples/services/ble-env-sense` and `examples/services/ble-env-sense/test-with-sensors`
    - sense_to_ble: Application that samples a sensor and advertises value with BLE
  - [ ] `examples/c_hello` and `examples/tests/printf_long`
    - uart_tx_small and uart_tx_large: applications that write to console with small and large buffers; run both in parallel to properly test virtualization
  - [ ] `examples/tests/console_recv_short` and `examples/tests/console_recv_long`
    - We expect this fail for 1.5 (console capsule not fully virtualized), and might work for 1.6.
  - [ ] `examples/blink`
    - blink: blinks LEDs
  - [ ] `examples/rot_client` and `examples/rot_service`
    - rot_ipc: tests IPC with a simple service
  - [ ] `examples/blink` and `examples/c_hello` and `examples/buttons`
  - [ ] `examples/lua-hello`
  - [ ] `examples/tests/console_timeout`
  - [ ] `examples/tests/malloc_test01`
  - [ ] `examples/tests/stack_size_test01`
  - [ ] `examples/tests/stack_size_test02`
  - [ ] `examples/tests/mpu_stack_growth`
  - [ ] `examples/tests/mpu_walk_region`
  - [ ] `examples/tests/multi_alarm_test`
  - [ ] `examples/tests/adc`
  - [ ] `examples/tests/adc_continuous`
  - [ ] `examples/tutorials/05_ipc/led` and `examples/tutorials/05_ipc/rng` and `examples/tutorials/05_ipc/logic`
  - [ ] `examples/tests/gpio` with mode set to 0
  - [ ] `examples/tests/whileone`
 

- Hail
  - [ ] `examples/tests/hail`
  - [ ] `examples/witenergy`
  - [ ] `examples/accel_leds`
[...]

PR #2042, opened 2020-07-25
Intel 8080 parallel interface HIL, Bus library with I2C, SPI and Intel 8080 parallel interface implementation

alexandruradovici:
### Pull Request Overview

Several hardware components work with the same protocol over several buses. For instance, sensors like l3gd20, lsm303 work over I2C and SPI, screens like ST7735 and ST7789H2 work over SPI or Intel 8080 parallel interface. Using the Bus library, capsules for these devices will work regardless of which hardware communication bus is used.

This PR
  * adds screen support for stm32f412gdiscovery
  * adds FSMC support for stm32f4xx
  * a Intel 8080 parallel interface (Bus8080) HIL
  * changes st7735 driver to st77xx single driver for ST7735 , ST7789H2 and LS016B8UY (reversed engineered, no documentation available) over the new Bus HIL
  * fixes the screen HIL and capsule (adds a `write_continue` function to the screen trait)

#### Intel 8080 parallel interface HIL

```rust
pub trait Bus8080<'a> {
    /// Set the address to write to
    fn set_addr(&self, addr_width: BusWidth, addr: usize) -> ReturnCode;

    /// Write data items to the previously set address
    fn write(&self, data_width: BusWidth, buffer: &'static mut [u8], len: usize) -> ReturnCode;

    /// Read data items from the previously set address
    fn read(&self, data_width: BusWidth, buffer: &'static mut [u8], len: usize) -> ReturnCode;

    fn set_client(&self, client: &'a dyn Client);
}

pub trait Client {
    /// Called when set_addr, write or read are complete
    ///
    /// set_address does not return a buffer
    /// write and read return a buffer
    /// len should be set to the number of data elements written
    fn command_complete(&self, buffer: Option<&'static mut [u8]>, len: usize);
}
```

The Bus8080 HIL is required as FSMC seems to be specific to STMs and is used to connect devices that support Parallel 8080 interface. Several screens work with this interface.

#### Bus Librarry

```rust
pub trait Bus<'a> {
    /// Set the address to write to
    ///
    /// If the underlaying bus does not support addresses (eg UART)
    /// this function returns ENOSUPPORT
    fn set_addr(&self, addr_width: BusWidth, addr: usize) -> ReturnCode;

    /// Write data items to the previously set address
    ///
    /// data_width specifies the encoding of the data items placed in the buffer
    /// len specifies the number of data items (the number of bytes is len * data_width.width_in_bytes)
    fn write(&self, data_width: BusWidth, buffer: &'static mut [u8], len: usize) -> ReturnCode;

    /// Read data items from the previously set address
    ///
    /// data_width specifies the encoding of the data items placed in the buffer
    /// len specifies the number of data items (the number of bytes is len * data_width.width_in_bytes)
    fn read(&self, data_width: BusWidth, buffer: &'static mut [u8], len: usize) -> ReturnCode;

    fn set_client(&self, client: &'static dyn Client);
}

pub trait Client {
    /// Called when set_addr, write or read are complete
    ///
    /// set_address does not return a buffer
    /// write and read return a buffer
    /// len should be set to the number of data elements written
    fn command_complete(&self, buffer: Option<&'static mut [u8]>, len: usize);
}
```

#### Example

Example for ST7735 screen before bus (spi dependent driver)
```rust
//! ```rust
 let tft = components::st7755::ST7735Component::new(alarm_mux).finalize(
     components::st7735_component_helper!(
         // spi type
         stm32f4xx::spi::Spi,
         // chip select
         stm32f4xx::gpio::PinId::PE03,
         // spi mux
         spi_mux,
         // timer type
         stm32f4xx::tim2::Tim2,
         // dc pin
         stm32f4xx::gpio::PinId::PA00.get_pin().as_ref().unwrap(),
         // reset pin
         stm32f4xx::gpio::PinId::PA00.get_pin().as_ref().unwrap()
     )
 );
```

Example for ST7735 screen after bus (bus dependent library)
```rust
let bus = components::bus::Bus8080BusComponent::new().finalize(
    components::bus8080_bus_component_helper!(
        // bus type
        stm32f412g::fsmc::Fsmc,
        // bus
        &stm32f412g::fsmc::FSMC
    ),
);

let tft = components::st77xx::ST77XXComponent::new(mux_alarm).finalize(
    components::st77xx_component_helper!(
        // screen
        &capsules::st77xx::ST7789H2,
        // bus type
        capsules::bus::Bus8080Bus<'static, stm32f412g::fsmc::Fsmc>,
        // bus
        &bus,
        // timer type
        stm32f412g::tim2::Tim2,
        // pin type
        stm32f412g::gpio::Pin,
        // dc pin (optional)
        None,
        // reset pin
        stm32f412g::gpio::PinId::PD11.get_pin().as_ref().unwrap()
    ),
);
```

### Testing Strategy

This pull request was tested using ST7735 and ST7789H2 screens using nano33ble and stm32f412gdiscovery

### TODO or Help Wanted

This pull request still needs some help for the FSMC. I have no idea how to make the write asynchronous, there is a FSMC interrupt but I'm not sure how to call it. One option would be DMA, but it does not seem to be available for FSMC. Any idea would be very useful, as the screen for stm32f412gdiscovery is connected the FSMC.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alexandruradovici:
I updated the bus and the st77xx drivers to use generics, it seems to reduce the code size by about 600 B. The code does look a little more complex.

bradjc:
I'm a little wary of adding another layer of abstraction (the Bus interface). I see what this would provide, but:
- I'm concerned this won't actually work, due to potential corner cases in different implementations between different chips (both MCUs and sensors). For example, implementing `Bus` for one chip might work for one sensor, but not for a different sensor. To capture these potential subtleties the Bus trait would have to get a lot more complicated.
- This will be very hard to test, since a chip capsule developer is very unlikely to have hardware to test the chip with both interfaces. This will almost ensure that we have capsules that look like they provide more support than has ever been tested.
- This adds another layer of abstraction which adds more complexity when debugging or understanding the code. And, we haven't run into the issue where a board uses an existing capsule but with a different bus.

alexandruradovici:
I think this discussion is going to be longer, so let me split this into several PR. I'll leave the bus here, add the FMSC and screens to other PRs.

alexandruradovici:
> I'm a little wary of adding another layer of abstraction (the Bus interface). I see what this would provide, but:
> 
>     * I'm concerned this won't actually work, due to potential corner cases in different implementations between different chips (both MCUs and sensors). For example, implementing `Bus` for one chip might work for one sensor, but not for a different sensor. To capture these potential subtleties the Bus trait would have to get a lot more complicated.

If the chips use some specific functions (or timing) for a specific protocol, I think they should the protocol capsules directly.

> 
>     * This will be very hard to test, since a chip capsule developer is very unlikely to have hardware to test the chip with both interfaces. This will almost ensure that we have capsules that look like they provide more support than has ever been tested.
> 

I don;t thnik this is a problem, components for chips might be defined for several buses. The developer would simply not write the component that he did not test.

```rust
#[macro_export]
macro_rules! st7735_i2c_component_helper {
  // ...
}

macro_rules! st7735_spi_component_helper {
  // ...
}

macro_rules! st7735_intel8080_component_helper {
  // ...
}
```

>     * This adds another layer of abstraction which adds more complexity when debugging or understanding the code. And, we haven't run into the issue where a board uses an existing capsule but with a different bus.

I have run into it when writing drivers for st77xxx screens. The LSM sensors have the same issue, they support both I2C and SPI.

boards/stm32f412gdiscovery/src/main.rs
```diff
@@ -7,14 +7,16 @@
 // https://github.com/rust-lang/rust/issues/62184.
 #![cfg_attr(not(doc), no_main)]
 #![deny(missing_docs)]
-
+#![feature(llvm_asm)]
```
alistair23, line 10: Why is this needed?

alistair23:
I agree with Brad here, at first I thought this would be really useful, but I think there will just be too many corner cases for this to be reliable.

For devices where this is the case couldn't you just implement both I2C and SPI for the device and then call generic functions? That will result in supporting both protocols with little work.

boards/stm32f412gdiscovery/src/main.rs
alexandruradovici: I left it there by mistake, deleting it.

alexandruradovici:
> I agree with Brad here, at first I thought this would be really useful, but I think there will just be too many corner cases for this to be reliable.
> 
I understand your point.
> For devices where this is the case couldn't you just implement both I2C and SPI for the device and then call generic functions? That will result in supporting both protocols with little work.

What about adding this as a library in libraries, I think I can use it for several capsules.

alistair23:
I don't think it needs to be a library, it won't be very long.

alexandruradovici:
It's not long, I just need to use it in several capsules and I am not sure where to place the code.

alistair23:
What I'm picturing (which could be wrong) isn't too much code duplication that it's probably not an issue. Up to you though

bradjc:
The core team discussed how to move forward with this and we think the best plan is to make `Bus` a library in capsules/. Ideally it would have a more specific name for the protocol on top of the physical bus (if there is a reasonable name). This main difference between this and a HIL is just that the implementations for different physical busses would be in capsules/ rather than chips/.

We also think that separating the transport layer from the operational logic on a chip-by-chip basis would work well, but if there are multiple chips with the same interface protocol then consolidating the interfaces into a library makes a lot of sense.

alexandruradovici:
> The core team discussed how to move forward with this and we think the best plan is to make `Bus` a library in capsules/. Ideally it would have a more specific name for the protocol on top of the physical bus (if there is a reasonable name). This main difference between this and a HIL is just that the implementations for different physical busses would be in capsules/ rather than chips/.
> 

Ok, that makes sense. I'm not sure how to proceed. Should there be a trait declared in capsules?

> We also think that separating the transport layer from the operational logic on a chip-by-chip basis would work well, but if there are multiple chips with the same interface protocol then consolidating the interfaces into a library makes a lot of sense.

alexandruradovici:
> > The core team discussed how to move forward with this and we think the best plan is to make `Bus` a library in capsules/. Ideally it would have a more specific name for the protocol on top of the physical bus (if there is a reasonable name). This main difference between this and a HIL is just that the implementations for different physical busses would be in capsules/ rather than chips/.
> 
> Ok, that makes sense. I'm not sure how to proceed. Should there be a trait declared in capsules?
> 
> > We also think that separating the transport layer from the operational logic on a chip-by-chip basis would work well, but if there are multiple chips with the same interface protocol then consolidating the interfaces into a library makes a lot of sense.

I'd love to close this as soon as possible, but @bradjc I would really need some clarifications on what you have in mind. Thank you.

alexandruradovici:
Hi guys, just a kind reminder for some feedback here about the bus library? Thanks.

hudson-ayers:
@alexandruradovici Here is a link to the discussion in the working group meeting a few weeks back: https://github.com/tock/tock/blob/master/doc/wg/core/notes/core-notes-2020-08-07.md#generic-bus-hil . I think the desired changes include declaring the trait in one file in `capsules/` (rather than in `kernel/hil') and then implementing the trait in other files in `capsules/`. Is there a reason that is problematic for your use case?

alexandruradovici:
I made the change from the HIL to the library Trait and updated the PR description.

alexandruradovici:
Hi, any feedback here? It would really help us to close this. Thanks.

chips/stm32f4xx/src/fsmc.rs
```diff
@@ -0,0 +1,358 @@
...
+    /// Address
+    reg: ReadWrite<u16>,
+    /// Data
+    ram: ReadWrite<u16>,
+}
+
+const FSMC_BANK1: StaticRef<FsmcBank> = unsafe { StaticRef::new(0x60000000 as *const FsmcBank) };
+const FSMC_BANK2_RESERVED: StaticRef<FsmcBank> = unsafe { StaticRef::new(0x0 as *const FsmcBank) };
+const FSMC_BANK3: StaticRef<FsmcBank> = unsafe { StaticRef::new(0x68000000 as *const FsmcBank) };
+const FSMC_BANK4_RESERVED: StaticRef<FsmcBank> = unsafe { StaticRef::new(0x0 as *const FsmcBank) };
```
hudson-ayers, line 149: I don't like this practice of treating an arbitrary memory location as though it is an `FsmcBank`. This could easily lead to memory unsafety if these slots were used. If empty slots are needed in the FSMC array, the array should hold `Option<StaticRef<FsmcBank>>`

kernel/src/hil/screen.rs
```diff
@@ -135,6 +186,15 @@ pub trait Screen {
...
     fn write(&self, buffer: &'static mut [u8], len: usize) -> ReturnCode;
 
+    /// Continues the previous previous write command
+    /// When finished, the driver will call the `write_complete()` callback.
+    ///
+    /// Return values:
+    /// - `SUCCESS`: Write is valid and will be sent to the screen.
+    /// - `EINVAL`: Write is invalid or length is wrong.
+    /// - `EBUSY`: Another write is in progress.
+    fn write_continue(&self, buffer: &'static mut [u8], len: usize) -> ReturnCode;
```
hudson-ayers, line 196: I am not clear on why this function is being added to the trait. What can lead to a write beginning, but not being finished? How should clients know when `write_continue()` needs to be called?

hudson-ayers (changes requested):
A couple comments. I think this is close to being merged, though I would like another set of eyes on the changes in `kernel/src/hil/`.

chips/stm32f4xx/src/fsmc.rs
alexandruradovici: Thanks for the tip, I modified it.

kernel/src/hil/screen.rs
alexandruradovici: The frame buffer data might be larger than the allocated RAM, for instance when performing a fill. 

- `write` sets the memory address at the beginning of the screen frame window
- `write_continue` does the same thing as `write`, but does not reset the frame windows position

chips/stm32f4xx/src/fsmc.rs
```diff
@@ -144,13 +144,13 @@ struct FsmcBank {
 }
 
 const FSMC_BANK1: StaticRef<FsmcBank> = unsafe { StaticRef::new(0x60000000 as *const FsmcBank) };
-const FSMC_BANK2_RESERVED: StaticRef<FsmcBank> = unsafe { StaticRef::new(0x0 as *const FsmcBank) };
+// const FSMC_BANK2_RESERVED: StaticRef<FsmcBank> = unsafe { StaticRef::new(0x0 as *const FsmcBank) };
 const FSMC_BANK3: StaticRef<FsmcBank> = unsafe { StaticRef::new(0x68000000 as *const FsmcBank) };
-const FSMC_BANK4_RESERVED: StaticRef<FsmcBank> = unsafe { StaticRef::new(0x0 as *const FsmcBank) };
+// const FSMC_BANK4_RESERVED: StaticRef<FsmcBank> = unsafe { StaticRef::new(0x0 as *const FsmcBank) };
```
hudson-ayers: Should these be labeled BANK0..BANK3 instead of BANK1..BANK4 to match the indicies in the array in which they are contained? It seems odd that programming the bank labeled BANK1 requires passing a `bank_id` of `0` to `write_data()`

kernel/src/hil/screen.rs
hudson-ayers: That makes sense. Can you add that description to the comments in this file? I initially misread this as `write_continue()` somehow resuming a paused write.

alexandruradovici: Done

chips/stm32f4xx/src/fsmc.rs
alexandruradovici: I added an enum for the Banks instead of using an index.

hudson-ayers (approved):
Looks good to me, would be great to have one more set of eyes on the HIL changes

alevy:
This looks good to me. I suggest we wait until after 1.6 to merge (#2030), and then merge immediately.

alevy (dismissed):

hudson-ayers (approved):
bors r+

Issue #2037, opened 2020-07-20
cortex-m: assembly missing some clobbers information


ppannuto:
Found during review of unrelated PR: https://github.com/tock/tock/pull/2002/files#r450389964

We need a "possibly good first issue" or something; this is nice an self-contained, but does assume you understand inline assembly.

Doesn't seem to be causing problems not to do these currently. But, we should do a pass making sure that all the clobbers are marked correctly to prevent future optimizations from causing unexpected errors. This issue tracks that need.

ppannuto:
Okay, since #2002 has landed, this has ratcheted up the priority list a bit, as it dumps out this warning:

```
[-bash] Mon 20 Jul 10:55 [[master $] ~/code/helena-project/tock/boards/hail]
$ make
[...]
warning: expected a clobber, found an option
   --> arch/cortex-m4/src/lib.rs:491:15
    |
491 |             : "volatile" );
    |               ^^^^^^^^^^

warning: 1 warning emitted
```

---

This feels like a bug twice over: the erroneous clobber, and the fact that CI didn't block this warning from coming in :/

PR #2043, opened 2020-07-27
cortex-m: fix warning about volatile keyword in incorrect position

hudson-ayers:
### Pull Request Overview

#2002 added a volatile keyword in the position intended for clobbers in the assembly responsible for resetting the stack pointer.
This surfaced a warning whenever compiling a cortex-m4 architecture, but this warning did not cause CI to fail because `cd arch/ && RUSTFLAGS="-D warnings" cargo test` ignores code that can't be compiled on the host platform, and we do not deny warnings when compiling all boards.

### Testing Strategy

This pull request was tested by compiling Imix and seeing that the warning is gone.

### TODO or Help Wanted

Ideally we would deny warnings in `ci-job-compilation` so mistakes like this couldn't slip past CI, but passing `RUSTFLAGS="-D warnings"` to the call to `make allboards` within `ci-job-compilation` did not seem to work, and I did not try anything further.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m3/src/lib.rs
+++ b/arch/cortex-m3/src/lib.rs
@@ -170,7 +170,7 @@ pub unsafe extern "C" fn svc_handler() {
     movw LR, #0xFFF9
     movt LR, #0xFFFF
     bx lr"
-    : : : "r0", "r1", "cc" : "volatile" );
+    : : : "r0", "r1", "lr", "cc", "memory" : "volatile" );
 }
 
 // Mock implementation for tests on Travis-CI.
@@ -417,7 +417,8 @@ pub unsafe extern "C" fn hard_fault_handler() {
             mov sp, r0   /* Set the stack pointer to _estack */"
             :
             : "{r0}"((_estack as *const ()) as u32)
-            : : "volatile" );
+            :
+            : "volatile" );
 
             // Panic to show the correct error.
             panic!("kernel stack overflow");
@@ -457,7 +458,7 @@ pub unsafe extern "C" fn hard_fault_handler() {
 
         movw LR, #0xFFF9
         movt LR, #0xFFFF"
-        : : : "r1", "r0", "r2" : "volatile" );
+        : : : "r1", "r0", "r2", "memory" : "volatile" );
     }
 }
 
--- a/arch/cortex-m3/src/lib.rs
+++ b/arch/cortex-m3/src/lib.rs
@@ -417,7 +417,7 @@ pub unsafe extern "C" fn hard_fault_handler() {
             mov sp, r0   /* Set the stack pointer to _estack */"
             :
             : "{r0}"((_estack as *const ()) as u32)
-            : "volatile" );
+            : : "volatile" );
 
             // Panic to show the correct error.
             panic!("kernel stack overflow");
--- a/arch/cortex-m3/src/lib.rs
+++ b/arch/cortex-m3/src/lib.rs
@@ -49,7 +49,7 @@ pub unsafe extern "C" fn systick_handler() {
 
     movw LR, #0xFFF9
     movt LR, #0xFFFF"
-    : : : : "volatile" );
+    : : : "r0" : "volatile" );
 }
 
 // Mock implementation for tests on Travis-CI.
@@ -170,7 +170,7 @@ pub unsafe extern "C" fn svc_handler() {
     movw LR, #0xFFF9
     movt LR, #0xFFFF
     bx lr"
-    : : : : "volatile" );
+    : : : "r0", "r1", "cc" : "volatile" );
 }
 
 // Mock implementation for tests on Travis-CI.
@@ -378,7 +378,7 @@ pub unsafe extern "C" fn hard_fault_handler() {
     mrsne  r0, psp   /* r0 = userland stack pointer */"
     : "={r0}"(faulting_stack), "={r1}"(kernel_stack)
     :
-    : "r0", "r1"
+    : "cc"
     : "volatile" );
 
     if kernel_stack {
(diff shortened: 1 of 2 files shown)


ppannuto:
There are a few other issues here too, namely that not all of the assembly blocks properly specify their clobbers (#2037) -- it would be good to fix all of this while we're here if we can.

hudson-ayers:
> There are a few other issues here too, namely that not all of the assembly blocks properly specify their clobbers (#2037) -- it would be good to fix all of this while we're here if we can.

Okay, I pushed a commit that I think fixes a number of issues with the cortex-m clobbers. I could have made some mistakes though, as I am hardly an inline assembly expert. The fixes can be summarized as follows:

1. anywhere there was a clobber of a register that is also an input or output register, I removed it as a listed clobber (see https://doc.rust-lang.org/unstable-book/library-features/llvm-asm.html#clobbers).

2. anywhere that used instructions which at a glance I knew affected the APSR register (i.e. `cmp` and similar) I added the "cc" clobber indicating that condition codes are modified by the assembly therein.

3. Added a bunch of registers to the clobbers list where I could see instructions in the assembly modifying those registers despite them not being listed as inputs, outputs, or clobbers.

I tried to match the changes to cortex-m4 and cortex-m3. I did not touch cortex-m0. Someone checking my work would be appreciated.

bradjc:
Arg can we merge this into #2027? I don't know how we're going to keep all of this straight.

hudson-ayers:
yeah I can port these changes over to #2027, but I would maybe like if they could be reviewed here first. I think that #2027 will be reviewed as though it is just reorganizing existing code, and its probably good to have at least one more set of eyes on these changes.

bradjc:
This looks fine to me (but I'm no expert), but I would rather see a more comprehensive review happen with a switch to llvm_asm to asm.

arch/cortex-m3/src/lib.rs
```diff
@@ -170,7 +170,7 @@ pub unsafe extern "C" fn svc_handler() {
     movw LR, #0xFFF9
     movt LR, #0xFFFF
     bx lr"
-    : : : : "volatile" );
+    : : : "r0", "r1", "cc" : "volatile" );
```
ppannuto, line 173:
```suggestion
    : : : "r0", "r1", "lr", "cc", "memory" : "volatile" );
```

arch/cortex-m3/src/lib.rs
```diff
@@ -417,7 +417,7 @@ pub unsafe extern "C" fn hard_fault_handler() {
             mov sp, r0   /* Set the stack pointer to _estack */"
             :
             : "{r0}"((_estack as *const ()) as u32)
-            : "volatile" );
+            : : "volatile" );
```
ppannuto, line 420:
```suggestion
            :
            : "volatile" );
```

arch/cortex-m3/src/lib.rs
```diff
@@ -457,7 +457,7 @@ pub unsafe extern "C" fn hard_fault_handler() {
 
         movw LR, #0xFFF9
         movt LR, #0xFFFF"
-        : : : : "volatile" );
+        : : : "r1", "r0", "r2" : "volatile" );
```
ppannuto, line 460:
```suggestion
        : : : "r1", "r0", "r2", "memory" : "volatile" );
```

arch/cortex-m4/src/lib.rs
```diff
@@ -228,7 +228,7 @@ pub unsafe extern "C" fn svc_handler() {
     movw LR, #0xFFF9
     movt LR, #0xFFFF
     bx lr"
-    : : : : "volatile" );
+    : : : "r0", "r1", "cc" : "volatile" );
```
ppannuto, line 231:
```suggestion
    : : : "r0", "r1", "cc", "memory" : "volatile" );
```

arch/cortex-m4/src/lib.rs
```diff
@@ -488,7 +488,7 @@ pub unsafe extern "C" fn hard_fault_handler() {
             mov sp, r0   /* Set the stack pointer to _estack */"
             :
             : "{r0}"((_estack as *const ()) as u32)
-            : "volatile" );
+            : : "volatile" );
```
ppannuto, line 491:
```suggestion
            :
            : "volatile" );
```

arch/cortex-m4/src/lib.rs
```diff
@@ -528,7 +528,7 @@ pub unsafe extern "C" fn hard_fault_handler() {
 
         movw LR, #0xFFF9
         movt LR, #0xFFFF"
-        : : : : "volatile" );
+        : : : "r1", "r0", "r2" : "volatile" );
```
ppannuto, line 531:
```suggestion
        : : : "r1", "r0", "r2", "memory" : "volatile" );
```

hudson-ayers:
> Huh, TIL about the `cc` and `memory` clobbers; neat.
> 
> I think we need `memory` clobbers wherever we're doing `str`'s IIUC the documentation correctly.

Yeah, agreed about the `memory` clobbers. Slightly better documentation is available here: https://llvm.org/docs/LangRef.html#clobber-constraints , and in general the llvm syntax seems based off of the gcc syntax, which has even better documentation here: https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Clobbers-and-Scratch-Registers 

> The "memory" clobber tells the compiler that the assembly code performs memory reads or writes to items other than those listed in the input and output operands (for example, accessing the memory pointed to by one of the input parameters). To ensure memory contains correct values, GCC may need to flush specific register values to memory before executing the asm. Further, the compiler does not assume that any values read from memory before an asm remain unchanged after that asm; it reloads them as needed. Using the "memory" clobber effectively forms a read/write memory barrier for the compiler.

PR #2044, opened 2020-07-27
Screen support for STM32F412G Discovery

alexandruradovici:
### Pull Request Overview

This pull request adds screen support for STM32F412G Discovery. It adds a Parallel 8080 Interface Bus HIL as the screen is connected to the FSMC of the STM chip (that seems to be an interface specific to STM) and uses the 8080 bus protocol.

### Testing Strategy

This pull request was tested with an STM32F412G Discovery

### TODO or Help Wanted

This pull request still needs some help for the FSMC. I have no idea how to make the write asynchronous, there is a FSMC interrupt but I'm not sure how to call it. One option would be DMA, but it does not seem to be available for FSMC. Any idea would be very useful, as the screen for stm32f412gdiscovery is connected the FSMC.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alexandruradovici:
Closing due to #2056

PR #2047, opened 2020-07-28
Virtual ADC support

cosmindanielradu19:
### Pull Request Overview

This pull request adds support for virtual ADC. It allows using a virtual ADC from a capsule and also provides a virtual ADC for the User Space.

The syscall interface mimics the existing ADC syscalls, so no changes are needed in the User Space.

For now, the virtual ADC supports only single sample requests.

This PR adds a support for the STM integrated ADC MCU temperature sensor.

### Testing Strategy

This pull request was tested using STM32F3discovery and STM32F412G-discovery

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


boards/stm32f3discovery/Makefile
```diff
@@ -22,6 +18,4 @@ flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
 
 .PHONY: program
 program: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
-	arm-none-eabi-objcopy --update-section .apps=$(APP) $(KERNEL) $(KERNEL_WITH_APP)
-	$(OPENOCD) $(OPENOCD_OPTIONS) -c "init; reset halt; flash write_image erase $(KERNEL_WITH_APP); verify_image $(KERNEL_WITH_APP); reset; shutdown"
-
```
alistair23, line 27: This should be in a separate PR

chips/sam4l/src/adc.rs
```diff
@@ -845,6 +846,13 @@ impl hil::adc::Adc for Adc {
     fn get_voltage_reference_mv(&self) -> Option<usize> {
         Some(3300)
     }
+
+    /// Sets the client for this driver.
+    ///
+    /// - `client`: reference to capsule which handles responses
+    fn set_client(&self, _client: &'static dyn hil::adc::Client) {
+        panic!("This is not implemented yet!");
```
alistair23, line 854: There is an `unimplemented` macro you can use instead

chips/stm32f4xx/src/adc.rs
```diff
@@ -264,7 +264,7 @@ const ADC_COMMON_BASE: StaticRef<AdcCommonRegisters> =
 
 #[allow(dead_code)]
 #[repr(u32)]
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, PartialEq)]
```
alistair23, line 267: Why do we need the `PartialEq`?

chips/sam4l/src/adc.rs
bradjc: There is already a set_client function which should be updated to this one.

capsules/src/virtual_adc.rs
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    fn call_driver(&self, command: Operation, channel: usize) -> ReturnCode {
+        match command {
+            Operation::OneSample => self.drivers[channel].sample(),
+        }
+    }
+}
+
+impl Driver for AdcSyscall<'_> {
```
bradjc, line 220: This is out-of-scope for a `virtual_X.rs` file. A new Driver should be in its own capsule.

boards/stm32f3discovery/Makefile
cosmindanielradu19: It was for testing only and I deleted it when the PR was done

chips/sam4l/src/adc.rs
cosmindanielradu19: The existing set_client function is not in the HIL and is defined for EverythingClient. Can you be a little bit more specific? I am not sure I understand the request.

chips/stm32f4xx/src/adc.rs
cosmindanielradu19: Several apps can request values from the same channel, there is an if condition to reply with the same sample at the same time to them.

```rust
impl<'a, A: hil::adc::Adc> hil::adc::Client for MuxAdc<'a, A> {
    fn sample_ready(&self, sample: u16) {
        for node in self.devices.iter() {
            self.inflight.map(|inflight| {
                if node.channel == inflight.channel {
                    node.operation.map(|operation| match operation {
                        Operation::OneSample => {
                            node.operation.clear();
                            node.client.map(|client| client.sample_ready(sample));
                        }
                    });
                }
            });
        }
        self.inflight.clear();
        self.do_next_op();
    }
}
```

boards/components/src/adc_syscall.rs
```diff
@@ -0,0 +1,65 @@
+use capsules::virtual_adc::AdcSyscall;
```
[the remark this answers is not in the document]
cosmindanielradu19: Yes

capsules/src/virtual_adc.rs
cosmindanielradu19: I will move it.

cosmindanielradu19:
> Can you describe what the use case is for using virtual_adc? It's not intuitive to me at least when an ADC channel would be shared.
> 
The idea is to be able to use the ADC from a capsule and also from the User Space.

> The virtual_X drivers are very tricky to get right, so it's worth agreeing that we need a new one. For i2c/uart/spi I think it makes sense that the kernel wants to be able to share the actual bus. But we don't have a virtual_gpio for example, because how can you share a particular gpio pin? With an ADC channel I guess some hardware would probably allow you to connect the same channel to different physical pins, so you would need a virtual_adc to share one adc channel with multiple adc pins. But I don't think that is the case here.
> 
At the GPIO, each pin works independeently, this is not the case for the ADC channels.

> If we assume that one adc channel is assigned to one adc pin, then that adc pin is likely connected to one analog device. So, rather than sharing at the adc channel level, we could share at the level of whatever the analog device is. For example, if the pin is connected to an analog pressure sensor, we could share pressure values rather than adc samples.

bradjc:
> The idea is to be able to use the ADC from a capsule and also from the User Space.

Do you have a use case in mind? I can think of contrived examples of where this could be used, but I haven't yet thought of a case where the kernel and userspace need to sample the same adc pin.

alexandruradovici:
> > The idea is to be able to use the ADC from a capsule and also from the User Space.
> 
> Do you have a use case in mind? I can think of contrived examples of where this could be used, but I haven't yet thought of a case where the kernel and userspace need to sample the same adc pin.

If not for the Mux, to use a channel from user space and another channel from a capsule you would have to set on the ADC two clients: one the driver exporting ADC channels to user space and another one the capsule using the another non-exported ADC channel. Maybe we got it wrong?

boards/stm32f3discovery/Makefile
alistair23: It's still in this PR though, it makes it very hard to review when you add code then remove it later in the same PR. It also makes it harder to bisect regressions once this is merged. Can you please try to avoid doing this and only keep the relevant code in the PR.

cosmindanielradu19: I am so sorry @alistair23 . I cleaned up the branch and everything should be fine now.

cosmindanielradu19:
Hei guys, have you got any feedback for this PR?

phil-levis:
I think this is valuable -- everything should be virtualized, at the very least for the reason that it means both userspace and the kernel can use the abstraction. I'll do a detailed read-through tomorrow. Sorry for the delay.

phil-levis (commented):
Does `capsules/src/adc.rs` remain for cases when we want a single application to be able to do continuous sampling? Why keep it and `capsules/src/adc_syscall.rs`?

cosmindanielradu19:
> Does `capsules/src/adc.rs` remain for cases when we want a single application to be able to do continuous sampling? Why keep it and `capsules/src/adc_syscall.rs`?

Yes, you are right. Are you suggesting to merge them into one single file?

phil-levis:
> Yes, you are right. Are you suggesting to merge them into one single file?
> Technically speaking I could implement AdcHighSpeed into the mux with all the functions returning ENOSUPPORT.

It's better to not implement it. That way, if your code needs to be able sample continuously, you detect this lack of functionality at compilation rather than at runtime.

One approach might be to put both implementations in `adc_syscalls`, name them appropriately, and add a comment at the top of the file describing their different uses.

cosmindanielradu19:
> > Yes, you are right. Are you suggesting to merge them into one single file?
> > Technically speaking I could implement AdcHighSpeed into the mux with all the functions returning ENOSUPPORT.
> 
> It's better to not implement it. That way, if your code needs to be able sample continuously, you detect this lack of functionality at compilation rather than at runtime.
> 
> One approach might be to put both implementations in `adc_syscalls`, name them appropriately, and add a comment at the top of the file describing their different uses.

I have merged both files in adc.rs because it was already implemented in most of the boards and now there is no need for major updates for the boards that already supported ADC

phil-levis:
OK, seems reasonable. I think the one change that's necessary is renaming Adc and AdcSyscall to be representative of their differences, and adding much clearer explanatory comments that describe their differences and tradeoffs. Right now, were I to come across this file without much background, it would be extremely confusing.

phil-levis (changes requested):
OK, seems reasonable. I think the one change that's necessary is renaming Adc and AdcSyscall to be representative of their differences, and adding much clearer explanatory comments that describe their differences and tradeoffs. Right now, were I to come across `capsules/src/adc.rs` without much background, it would be extremely confusing.

chips/stm32f303xc/src/adc.rs
```diff
@@ -502,7 +502,9 @@ pub struct Adc {
     common_registers: StaticRef<AdcCommonRegisters>,
     clock: AdcClock,
     status: Cell<ADCStatus>,
-    client: OptionalCell<&'static dyn EverythingClient>,
+    client: OptionalCell<&'static dyn hil::adc::Client>,
+    requested: Cell<ADCStatus>,
+    requeste_channel: Cell<u32>,
```
hudson-ayers, line 507:
```suggestion
    requested_channel: Cell<u32>,
```

chips/stm32f303xc/src/adc.rs
```diff
@@ -564,18 +572,24 @@ impl Adc {
...
             // Clear interrupt
-            self.registers.isr.modify(ISR::ADRDY::CLEAR);
             self.registers.ier.modify(IER::ADRDYIE::CLEAR);
             // Set Status
             if self.status.get() == ADCStatus::PoweringOn {
                 self.status.set(ADCStatus::Idle);
+                match self.requested.get() {
+                    ADCStatus::OneSample => {
+                        // self.registers.isr.modify(ISR::EOS::SET);
+                        self.sample_u32(self.requeste_channel.get());
```
hudson-ayers, line 582:
```suggestion
                        self.sample_u32(self.requested_channel.get());
```

chips/stm32f303xc/src/adc.rs
```diff
@@ -623,6 +633,31 @@ impl Adc {
     pub fn disable_clock(&self) {
         self.clock.disable();
     }
+
+    fn enable_special_channels(&self) {
+        // enabling temperature channel
+        if self.requeste_channel.get() == 16 {
```
hudson-ayers, line 639:
```suggestion
        if self.requested_channel.get() == 16 {
```

chips/stm32f303xc/src/adc.rs
```diff
@@ -623,6 +633,31 @@ impl Adc {
...
+            self.registers.sqr1.modify(SQR1::SQ1.val(channel));
+            self.registers.ier.modify(IER::EOSIE::SET);
+            self.registers.ier.modify(IER::EOCIE::SET);
+            self.registers.ier.modify(IER::EOSMPIE::SET);
+            self.registers.cr.modify(CR::ADSTART::SET);
+            ReturnCode::SUCCESS
+        } else {
+            ReturnCode::EBUSY
+        }
+    }
```
hudson-ayers, line 660: Its not intuitive to me why a function named `sample_u32` enables the temperature channel any time it is called

chips/stm32f303xc/src/adc.rs
```diff
@@ -646,20 +681,12 @@ impl hil::adc::Adc for Adc {
 
     fn sample(&self, channel: &Self::Channel) -> ReturnCode {
         if self.status.get() == ADCStatus::Off {
+            self.requested.set(ADCStatus::OneSample);
+            self.requeste_channel.set(*channel as u32);
```
hudson-ayers, line 685:
```suggestion
            self.requested_channel.set(*channel as u32);
```

kernel/src/hil/adc.rs
```diff
@@ -37,6 +37,8 @@ pub trait Adc {
     ///
     /// The returned reference voltage is in millivolts, or `None` if unknown.
     fn get_voltage_reference_mv(&self) -> Option<usize>;
+
+    fn set_client(&self, client: &'static dyn Client);
```
hudson-ayers, line 41: I think this should need a doc comment for rustdoc to pass

hudson-ayers (commented):
Some small nits.

Zoomed out, it seems this solves a problem (before this PR, either a capsule could use the ADC, or userspace could use the ADC -- there was no way for a capsule to use one channel and a userspace app to use another), however it seems it maybe goes *too* far in allowing simultaneous access to a given ADC channel by both userspace and a capsule, as doing so feels like a dark pattern -- generally this is the type of thing that would occur when a capsule and an app are sharing some resource that is known apriori, like a sensor, and rather than exposing that sensor as a raw ADC pin it should be exposed by the capsule that knows what it is (in the case of this PR, the stm_mcu_temperature capsule could provide its own system call driver to allow apps to get the temperature of the MCU, rather than apps also using the ADC pin and performing the calculations themselves).

I think that having a virtual ADC capsule is valuable, however, as there are many cases where it can be imagine that a capsule needs one ADC channel and an app needs another. And I am not sure that it is necessarily important we prevent the case where capsules and apps use the same channel.

cosmindanielradu19:
> Some small nits.
> 
> Zoomed out, it seems this solves a problem (before this PR, either a capsule could use the ADC, or userspace could use the ADC -- there was no way for a capsule to use one channel and a userspace app to use another), however it seems it maybe goes _too_ far in allowing simultaneous access to a given ADC channel by both userspace and a capsule, as doing so feels like a dark pattern -- generally this is the type of thing that would occur when a capsule and an app are sharing some resource that is known apriori, like a sensor, and rather than exposing that sensor as a raw ADC pin it should be exposed by the capsule that knows what it is (in the case of this PR, the stm_mcu_temperature capsule could provide its own system call driver to allow apps to get the temperature of the MCU, rather than apps also using the ADC pin and performing the calculations themselves).
> 
> I think that having a virtual ADC capsule is valuable, however, as there are many cases where it can be imagine that a capsule needs one ADC channel and an app needs another. And I am not sure that it is necessarily important we prevent the case where capsules and apps use the same channel.

Do you have any better idea of how to change the ADC? I mean there might not be any type of problems because the MUX will handle everything and it can't get blocked. The only reason that ADC might be blocked is the HighSpeed ADC which is not implemented yet.

boards/stm32f3discovery/Makefile
alistair23: This is still in the PR. You have completely remove the ability to program the board.

alistair23 (changes requested):

phil-levis:
I did a bit more tidying on the names and expanded the comments a bit:

https://github.com/phil-levis/tock/commit/029d7f125252a115770aef1f14b04338ceb7c29a

Please apply this and I think I'm ready to approve.

cosmindanielradu19:
> I did a bit more tidying on the names and expanded the comments a bit:
> 
> [phil-levis@029d7f1](https://github.com/phil-levis/tock/commit/029d7f125252a115770aef1f14b04338ceb7c29a)
> 
> Please apply this and I think I'm ready to approve.

Applied and rebased

phil-levis (dismissed):

hudson-ayers (dismissed):

phil-levis:
Looks like there's a small conflict -- resolve it and we're good to go!

phil-levis (approved):
Looks good!

PR #2052, opened 2020-07-29
Fix potential unsoundness in dynamic grant allocation

daboross:
### Pull Request Overview

In the original code, `Allocator::alloc` returned `Owned`, a type which asserts that the memory it is pointing to is accessible.

If a capsule stores the returned `Owned` outside of an app-specific `Grant`, and then tries to access it after the app has crashed/restarted, it will access deallocated memory.

To fix this, I've changed `Allocator::alloc` to return a new type, `DynamicGrant`, which requires an `enter` call, and then exposes `Borrowed`.

This API exposes `Borrowed` rather than `Owned` as I believe it matches the nature of this `enter` function better. In particular, the value will be accessible multiple times, and the closure doesn't actually get ownership of it - only temporary access.

CC @alevy 

### Testing Strategy

I applied changes from #2053 on top of these changes, and then ran the tests from PR. Mainly, creating a temporary test capsule and app which allocate, check the results, and then re-allocate every other tick until running out of grant memory.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -34,6 +34,43 @@ impl<T> AppliedGrant<T> {
     }
 }
 
+/// Grant which was dynamically allocated in a particular app's memory.
+pub struct DynamicGrant<T: ?Sized> {
+    data: NonNull<T>,
+    appid: AppId,
+}
+
+impl<T: ?Sized> DynamicGrant<T> {
+    /// Creates a new `DynamicGrant`.
+    ///
+    /// # Safety
+    ///
+    /// `data` must point to a valid, initialized `T`.
+    unsafe fn new(data: NonNull<T>, appid: AppId) -> Self {
+        DynamicGrant { data, appid }
+    }
+
+    pub fn appid(&self) -> AppId {
+        self.appid
+    }
+
+    /// Gives access to inner data within the given closure.
+    ///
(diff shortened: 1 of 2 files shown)


kernel/src/grant.rs
```diff
@@ -34,6 +34,43 @@ impl<T> AppliedGrant<T> {
...
+    where
+        F: FnOnce(Borrowed<'_, T>) -> R,
+    {
+        self.appid.kernel.process_map_or(None, self.appid, |_| {
+            let data = unsafe { self.data.as_mut() };
+            let borrowed = Borrowed::new(data, self.appid);
+            Some(fun(borrowed))
+        })
+    }
+}
```
hudson-ayers, line 72: Do you think it makes more sense for this function to return something like a `Result<R, ReturnCode>`?

hudson-ayers (commented):
One minor comment, overall I think this looks good.

I do think that eventually this will need to add some mechanism for freeing data (`Owned` implemented `Drop`, which called `process.free()` when it went out of scope). But given that we have yet to implement actually implement `free()` I don't think it makes sense to block on that.

ppannuto:
@daboross, we discussed this on the call today and decided we would like to include this fix in the upcoming 1.6 release. It looks like there is currently a merge conflict, however.

hudson-ayers:
replaced by #2229

PR #2053, opened 2020-07-29
Add utilities for allocating dynamic grants.

daboross:
### Pull Request Overview

This pull request is intended to expand the basic functionality for dynamically allocating values. It:

- changes the existing `alloc` method to accept non-`Default` values
- adds a new method, `alloc_n_with`, for allocating a runtime-determined number of contiguous values
- splits the internal `alloc_unowned` into `alloc_raw` and `alloc_default_unowned` to allow for allocating non-`Default` values

I believe all of these changes will be useful, but the one which motivated this PR was `alloc_n_with`. To support connections in the Rubble BLE capsule, we'll need to support attributes, and the capsule won't know ahead of time how many attributes any given app will want to have in a connection. Attributes are most naturally stored in an array, and since the kernel won't know at compile time how many attributes an app has, we need dynamic allocation. Hence, `alloc_n_with`.

Note that this PR leaves the usage of `Owned`, which I believe is unsound. Once we re-export `Owned` so that applications can store them, those applications could potentially store them outside of app-specific grant memory, and then access them after the app has been restarted (thus accessing deallocated memory). I have a fix for this, but since the changes are fairly separate, I've submitted it as a separate PR - #2052.

If one PR is merged first, the other will need a rebase, but it shouldn't matter which one that is.

CC @alevy 

### Testing Strategy

I created a temporary test capsule and app which calls `alloc` and `alloc_n_with` every other tick until we hit `OutOfMemory`, keeps each allocation one tick, and prints the data out after retrieving it.

### TODO or Help Wanted

While this was manually tested, automated testing would be better. Do we have a place for unit tests for kernel functionality exposed to capsules?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -81,14 +81,28 @@ impl<T: ?Sized> DerefMut for Owned<T> {
 }
 
 impl Allocator {
-    /// Allocates a new owned grant initialized to the given value.
-    pub fn alloc<T>(&mut self, val: T) -> Result<Owned<T>, Error> {
+    /// Allocates a new owned grant initialized using the given closure.
+    ///
+    /// The closure will be called exactly once, and the result will be used to
+    /// initialize the owned value.
+    ///
+    /// This interface was chosen instead of a simple `alloc(val)` as it's
+    /// much more likely to optimize out all stack intermediates. This
+    /// helps to prevent stack overflows when allocating large values.
+    ///
+    /// # Panic Safety
+    ///
+    /// If `init` panics, the freshly allocated memory may leak.
+    pub fn alloc_with<T, F>(&mut self, init: F) -> Result<Owned<T>, Error>
+    where
+        F: FnOnce() -> T,
+    {
         unsafe {
             let ptr = self.alloc_raw()?;
 
             // We use `ptr::write` to avoid `Drop`ping the uninitialized memory in
             // case `T` implements the `Drop` trait.
-            write(ptr.as_ptr(), val);
+            write(ptr.as_ptr(), init());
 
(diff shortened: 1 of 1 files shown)


alevy:
I'm dropping a note that @daboross and I discussed this change one-on-one and I believe it is (a) _necessary_ for safety and (b) correct. Because this is a security critical change, we should review this more carefully than normal and probably document it.

Currently, I'm looking through this style (importantly, can we reduce the number of types which would help reason about security) as well as convince myself this is indeed correct this time.

bradjc:
I in general like these changes, but I'm worried they undo https://github.com/tock/tock/pull/2003 which can cause stack overflows.

alevy:
@bradjc that's a good catch. @dabross was that intentional? I.e. is there a reason to avoid that now, or can we switch back?

alevy:
Also, otherwise this looks correct me now that I've looked through it more carefully

daboross:
@bradjc @alevy 

Thank you for catching that! It was not intentional - I didn't realize this was explicitly chosen to avoid stack overflows.

I've pushed an updated interface which I believe both protects against stack overflows and allows non-`Default` values.

alistair23:
Ping

hudson-ayers (approved):

alistair23:
I added the last call label, hopefully someone can merge this soon.

ppannuto (approved):
I think this looks good -- I'll give @bradjc and @alevy an explicit ping and a day as they've commented before and this is sensitive code

bradjc (approved):
tested with IPC on hail and looks good

PR #2055, opened 2020-07-29
Remove tock-registers' dependence on feature(const_fn)

regexident:
### Pull Request Overview

This pull request aims to remove **tock-registers**' dependence on `feature(const_fn)` by conditionally switching between `fn` and `const fn` based on a new `unstable_const_fn` crate feature, instead of enforcing the latter (and thus `feature(const_fn)`).

Since **tock-registers** is available as a stand-alone library crate this change is desirable, even if the effect on **tock** as a whole is negligible, as it removes the nightly-enforcing use of `feature(const_fn)`, allowing the use of **tocks-registers** (or the related [**rust-embedded/register-rs**](https://crates.io/crates/register) crate) in projects using stable Rust.

To achieve this I added a dependency for the [`const_fn`](https://crates.io/crates/const_fn) crate, which then allows for conveniently decorating `const fn`s with `#[const_fn(feature = "unstable_const_fn")]`, instead of the more involved alternative of providing two copies of the same function, each with corresponding  `#[cfg(feature = "unstable_const_fn")]` &  `#[cfg(not(feature = "unstable_const_fn"))]` conditional compilation guards.

This causes the `const fn` in the **tock-registers** crate to compile to `const fn` if the `unstable_const_fn` feature is enabled, otherwise just `fn`.

### Testing Strategy

Building `tock-registers` with the following configurations has the expected results:

| Command                                               | Compiles? |
| ----------------------------------------------------- | --------- |
| `cargo +stable check --features "unstable_const_fn"`  | NO        |
| `cargo +stable check`                                 | YES       |
| `cargo +nightly check --features "unstable_const_fn"` | YES       |
| `cargo +nightly check --features "unstable_const_fn"` | YES       |

While building `tock` with `cargo check` does not break.

### TODO or Help Wanted

This pull request still needs to resolve the following open questions:

#### Open Questions

1. **Use proc-macro dependency or write and use own macro by example?**

   The pull request currently introduces a dependency for a crate that has so far only had 12 versions published and 1,820 downloads (94 commits; last from 2 months ago).

   If you think this is too much a risk, then alternatively one could use a [macro by example](https://stackoverflow.com/questions/49479854/how-do-i-change-a-functions-qualifiers-via-conditional-compilation) instead, which would change things from this:

   ```rust
   #[const_fn(feature = "unstable_const_fn")]`
   const fn foo(...) {
       ...
   }
   ```

   … to this:

   ```rust
   maybe_const_fn! {
       const fn foo(...) {
           ...
       }
   }
   ```

   … effectively trading convenience for reliability & trust.

   

   It is worth noting however that either way the use of such "hacks" would become obsolete (and thus could be removed without breaking, since `fn` -> `const fn` is a non-breaking change) as soon as the required subset of `const_fn` is stable.

2. **Make `unstable_const_fn` a default feature?**

   It is my understanding that while **tock-registers** is published as a stand-alone library crate it is considered mostly an internal dependency.

   As such (and in order to conform to the idiomatic "opt-in to experimental stuff" pattern of Rust) I decided to _not_ make `unstable_const_fn` a default feature, and instead chose to update the *internal* dependencies of **tock** like so:

   ```toml
   [dependencies.tock-registers]
   ...
   features = ["unstable_const_fn"]
   ```

   This however may lead to breakage for depending crates, in case they are actually making use of the existing `const fn`s in a `const` context.

   However:

   - according to [crates.io](https://crates.io/) the [only crate dependent on tocks-registers](https://crates.io/crates/tock-registers/reverse_dependencies) is [**rust-embedded/register-rs**](https://crates.io/crates/register), which is merely a thin name-erasing shim around **tocks-registers** anyway, and could thus easily be updated accordingly.
   - **tocks-registers** is currently at `0.5.0`. As such we actually could just release it as `0.6.0`, taking the breaking change as `0.x`-releases are considered breaking anyway.

#### TODO:

- [ ] Bump version to `0.6.0` & update internal dependencies

****

### Documentation Updated

- [x] ~~Updated the relevant files in `/docs`~~, or <u>no updates are required</u>.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/Cargo.toml
+++ b/arch/rv32i/Cargo.toml
@@ -7,6 +7,8 @@ edition = "2018"
 [dependencies]
 kernel = { path = "../../kernel" }
 tock-rt0 = { path = "../../libraries/tock-rt0" }
-tock-registers = { path = "../../libraries/tock-register-interface" }
 riscv-csr = { path = "../../libraries/riscv-csr" }
 
+[dependencies.tock-registers]
+path = "../../libraries/tock-register-interface"
+features = ["unstable_const_fn"]
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -5,5 +5,8 @@ authors = ["Tock Project Developers <tock-dev@googlegroups.com>"]
 edition = "2018"
 
 [dependencies]
-tock-registers = { path = "../libraries/tock-register-interface" }
 tock-cells = { path = "../libraries/tock-cells" }
+
+[dependencies.tock-registers]
+path = "../libraries/tock-register-interface"
+features = ["unstable_const_fn"]
--- a/libraries/riscv-csr/Cargo.toml
+++ b/libraries/riscv-csr/Cargo.toml
@@ -14,5 +14,6 @@ edition = "2018"
 [badges]
 travis-ci = { repository = "tock/tock", branch = "master" }
 
-[dependencies]
-tock-registers = { path = "../tock-register-interface" }
+[dependencies.tock-registers]
+path = "../tock-register-interface"
+features = ["unstable_const_fn"]
--- a/libraries/tock-register-interface/Cargo.toml
+++ b/libraries/tock-register-interface/Cargo.toml
@@ -14,5 +14,9 @@ edition = "2018"
 [badges]
 travis-ci = { repository = "tock/tock", branch = "master" }
 
+[dependencies]
+const_fn = "^0.3"
+
 [features]
 no_std_unit_tests = []
+unstable_const_fn = []
--- a/libraries/tock-register-interface/src/lib.rs
+++ b/libraries/tock-register-interface/src/lib.rs
@@ -2,7 +2,7 @@
 //!
 //!
 
-#![feature(const_fn)]
+#![cfg_attr(feature = "unstable_const_fn", feature(const_fn))]
 #![no_std]
 
 pub mod macros;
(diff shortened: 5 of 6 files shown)


regexident:
Hm, looks like `cargo check` alone wasn't enough to test. :/
Looking at the stuff that's failing I seem to have slightly underestimated the use of those `const fn`s.
What a bummer. Feel free to close this PR, I guess?

ppannuto:
Hi @regexident, thanks for trying! It certainly would have been exciting to have solved this, as we are on a path towards stable rust (#1654) for the whole of Tock.

I think the challenge here is that the const fn crate doesn't actually _do_ anything; it looks to be just syntactic sugar around removing `const`. In Tock, however, we lean pretty heavily on const methods to allow for rich configuration idioms that do not impose any runtime overhead. As a result, many things have transitive const requirements. While a technique of 'conditional const-ness' might be feasible, I don't think it would be a great fit for Tock as (1) it would pollute a non-trivial portion of the codebase, and (2) it would potentially bloat runtime sizes -- we already have several platforms which capacity issues, so we unfortunately don't have a lot of headroom here.

It does looks like there's some motion in upstream Rust to begin stabilizing at least some portions of const fn's, so hopefully things aren't too far off.

regexident:
100% agreed.

I was of the wrong impression that the use of `const fn` was mostly of the cosmetic/optimization kind.

To figure out how `feature(const_fn)` is used I searched for `const fn`, which only showed occurrences in `registers.rs` and naïvely ignored how these were used in `macros.rs` via `const _: _ = …` and hence figured "well, this should work!".

Oh well, one more reason why `feature(const_fn)` can't be stabilized early enough.

PR #2064, opened 2020-07-31
pmp: add PMP struct

bradjc:
### Pull Request Overview

The PMP wasn't working on the arty because the configure function wasn't actually iterating the PMP regions. This separates out the PMP from the config, and makes a lot more sense to me. And it works on the arty-e21 board.

### Testing Strategy

This pull request was tested by running on arty e21.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -168,7 +168,6 @@ impl Default for PMPConfig {
         PMPConfig {
             regions: [None; $x / 2],
             is_dirty: Cell::new(true),
-
             app_memory_region: OptionalCell::empty(),
         }
     }
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -56,6 +56,22 @@ register_bitfields![u8,
     ]
 ];
 
+/// Main PMP struct.
+pub struct PMP {
+    /// The application that the MPU was last configured for. Used (along with
+    /// the `is_dirty` flag) to determine if MPU can skip writing the
+    /// configuration to hardware.
+    last_configured_for: MapCell<AppId>,
+}
+
+impl PMP {
+    pub const unsafe fn new() -> PMP {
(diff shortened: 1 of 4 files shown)


arch/rv32i/src/pmp.rs
```diff
@@ -136,37 +152,29 @@ impl PMPRegion {
     }
 }
 
-pub trait PMPConfigType: Default + Copy + Clone + Sized {}
-
 /// Struct storing region configuration for RISCV PMP.
-pub struct PMPConfig<N: PMPConfigType> {
+pub struct PMPConfig {
```
alistair23, line 156: This change is fine, but just noting that it is moving us further towards the macro approach as we get further away from the array generators.

arch/rv32i/src/pmp.rs
```diff
@@ -136,37 +152,29 @@ impl PMPRegion {
...
-impl PMPConfigType for [Option<PMPRegion>; $x / 2] {}
-
-impl Default for PMPConfig<[Option<PMPRegion>; $x / 2]> {
+impl Default for PMPConfig {
     fn default() -> Self {
         PMPConfig {
             regions: [None; $x / 2],
             is_dirty: Cell::new(true),
-            last_configured_for: MapCell::empty(),
+
```
alistair23, line 171: There is an extra new line here

arch/rv32i/src/pmp.rs
hudson-ayers: For general awareness, the minimum subset of const generics that will be stabilized now has its own feature gate (https://github.com/rust-lang/rust/pull/74877). Its not stable yet, but clearly on the fast track now.

arch/rv32i/src/pmp.rs
bradjc, line 171:
```suggestion
```

PR #2066, opened 2020-08-04
Msp432: DMA-support

lebakassemmerl:
### Pull Request Overview

This PR adds DMA-support, which will be used with UART TX and RX and later with SPI, ADC, etc.
This is just a first version where only UART TX is supported.

### Testing Strategy

Testing with UART on eval-board

### Help Wanted
* It would be nice to get some feedback if the interface-definitions make sense how they are now implemented
* Any recommendations/ideas how to better align the `DMA_CONFIG` struct, I just wanted to make a standard array and no struct, but I didn't find anything to align arrays in rust.

### TODO
- [x] implement `transfer_mem_to_peripheral()` for UART TX
- [x] implement `transfer_peripheral_to_mem()` for UART RX
- [x] implement `transfer_mem_to_mem()`

### Documentation Updated

n/a

### Formatting
Ran `make ci-nosetup`

### DIFF
--- a/chips/msp432/src/dma.rs
+++ b/chips/msp432/src/dma.rs
@@ -3,7 +3,7 @@
 use core::cell::Cell;
 use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::common::registers::{
-    register_bitfields, register_structs, ReadOnly, ReadWrite, WriteOnly,
+    register_bitfields, register_structs, InMemoryRegister, ReadOnly, ReadWrite, WriteOnly,
 };
 use kernel::common::StaticRef;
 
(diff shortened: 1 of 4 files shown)


ppannuto:
image

The changes in these files look pretty separate from the DMA stuff. In general, we try to keep PRs focused to make them easier to review. To the extent reasonable, more, smaller PRs are better than big ones. These changes also look like stuff that can/should sail through quickly as they're general platform bugfixes, while the DMA will probably take some time to review. Can you pull these out?

lebakassemmerl:
> image
> 
> The changes in these files look pretty separate from the DMA stuff. In general, we try to keep PRs focused to make them easier to review. To the extent reasonable, more, smaller PRs are better than big ones. These changes also look like stuff that can/should sail through quickly as they're general platform bugfixes, while the DMA will probably take some time to review. Can you pull these out?

Sure!
So then I guess it's the best to just create an own platform-fix PR with these changes, right?

hudson-ayers:
> So then I guess it's the best to just create an own platform-fix PR with these changes, right?

Yep!

lebakassemmerl:
I also implemented already an `abort_transfer()` function, which stops any active DMA transfers. My plan was to use in the `transmit_abort()` and `receive_abort()` function from the UART traits. Should I add these changes here? Or should I create another PR with them?

An do you have any recommendations how to test the UART abort-functions? I actually don't really know how to trigger them.

alistair23:
It should probably be a separate PR.

I have no idea how to test it though

lebakassemmerl:
Am I supposed to change something or add any missing features here? Or is it just necessary to wait for some more feedback? (:

chips/msp432/src/dma.rs
```diff
@@ -0,0 +1,916 @@
...
+
+        for i in 0..AVAILABLE_DMA_CHANNELS {
+            let bit = (1 << i) as u32;
+            if (bit & int) > 0 {
+                unsafe {
+                    // Clear interrupt-bit
+                    DMA_BASE.int0_clrflg.set(bit);
+
+                    DMA_CHANNELS[i].handle_interrupt();
+                }
```
ppannuto, line 908: Can you add a comment here explaining why this unsafe block is needed?

Also, I think you can move the `set()` part out of the unsafe block, no?

lebakassemmerl: Yes you are right, `set()` can be moved out, thanks for the hint!

I didn't add a comment to the `unsafe` block because it is accessing a global mutable variable, so it seemed obvious to me that this must be unsafe. But I will add a comment here (:

lebakassemmerl:
> Thanks for iterating on this a few times with us! Really exciting to have DMA on a new platform +1

Of course, no problem! I'm also very excited to see new features being accepted in the main-tree (:

PR #2068, opened 2020-08-05
schedulers: store references to slots in processes array instead of AppIds

### LINKED ISSUE #2063: Scheduler: AppIds can change
On hail, if an app faults the kernel should allow the app to restart several times before causing a full kernel panic. Right now, however, if an app faults and restarts once the entire kernel seems to stop.

I've traced this back to that when an app restarts its AppId changes (so that any old references to the old process execution are no longer valid). However, the scheduler (round robin) too has an old copy, and the scheduler just keeps returning this non existent app from `next()`, but the kernel immediately checks and realizes the appid is not valid. So the kernel gets stuck in this loop.

hudson-ayers:
### Pull Request Overview

This pull request fixes #2063 , by modifying each scheduler that stored a local copy of an AppId to instead store a pointer to a slot in the processes array. It also modifies initialization of these lists such that each scheduler has a list of all slots in the process array, and simply skips empty slots when scheduling. This makes restartable apps work again, and is also a first step towards supporting dynamically loaded apps (I imagine that will also involve replacing `Option<&'static dyn ProcessType>` with `OptionalCell<&'static dyn ProcessType>`).

This PR also changes the components for the mlfq and cooperative scheduler to include improvements I had made to the round robin component.

One downside of this approach is that if a process moves to a new slot in the app array, the scheduler state associated with that process is lost (i.e. position in the round robin queue is changed, and for mlfq the tracker of total CPU time used by the process during the current scheduling period will be inaccurate until the next resurrection). For the current schedulers I think this is fine, as it seems reasonable that moving processes around in the array might reset some state temporarily. If in the future schedulers are added that need to track state for all-time, we may want to reconsider this.

### Testing Strategy

This pull request was tested by running two apps with each scheduler on Imix, and verifying that process restarts work using the process console.

### TODO or Help Wanted

I ended up choosing this design over alternatives due to its simplicity. Eventually, we may want to move to the Kernel owning a generic `ProcessCollection` type, but doing this without adding additional trait-object indirection/overhead feels difficult.

### Documentation Updated

- [x] No updates are required

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/sched/cooperative.rs
+++ b/boards/components/src/sched/cooperative.rs
@@ -11,17 +11,20 @@
 
 // Author: Hudson Ayers <hayers@stanford.edu>
 
+use core::mem::MaybeUninit;
 use kernel::component::Component;
 use kernel::procs::ProcessType;
-use kernel::static_init;
+use kernel::{static_init, static_init_half};
 use kernel::{CoopProcessNode, CooperativeSched};
 
 #[macro_export]
 macro_rules! coop_component_helper {
     ($N:expr) => {{
-        use kernel::static_init;
+        use core::mem::MaybeUninit;
+        use kernel::static_buf;
         use kernel::CoopProcessNode;
-        static_init!([Option<CoopProcessNode<'static>>; $N], [None; $N])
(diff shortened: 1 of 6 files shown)


bradjc:
Last call as I want to merge this quickly (it really is just a bug fix for the scheduler PR).

Issue #1545, opened 2020-01-16
`static mut` data considered harmful (as opposed to `static_init`)


gendx:
## Problem description

Despite the existing [`static_init` macro](https://github.com/tock/tock/blob/master/kernel/src/common/utils.rs) to initialize static data, there are still many uses of "raw" `static mut` buffers or other data. `git grep "^static mut"` and `git grep "^pub static mut"` yield many results across the code base.

However, `static mut` is fundamentally unsafe, because it's relatively easy to alias static data from anywhere in the program. There are indeed evidence of why it is fundamentally problematic in the Rust language - and it may even be deprecated in the future:
- "Why is static mut bad?" https://users.rust-lang.org/t/why-is-static-mut-bad/32888/6
- "Consider deprecation of UB-happy `static mut`" https://github.com/rust-lang/rust/issues/53639

In particular, given the pervasive definition of `unsafe` functions in Tock, it's quite easy to forget that accessing `static mut` data is unsafe and why it is unsafe. All one needs to do to alias a `static mut BUF` is to write `&mut BUF` in an unsafe block of code (and a large chunk of Tock's code is wrapped in `unsafe` functions). This was one of my concerns in https://github.com/tock/tock/issues/1447.

An example is the debug-writer component, which eagerly creates mutable references to the kernel's `INTERNAL_BUF` and `OUTPUT_BUF` static mut objects.

https://github.com/tock/tock/blob/b25a1f80e6db330988b6c8ce039b80351cc7402f/boards/components/src/debug_writer.rs#L41-L56

On the other hand the `static_init` macro "hides" the underlying static buffer (it's private to the macro's block of code), and returns a single `&'static mut` reference instead. The advantage of `&'static mut` over plain `static mut` is that it requires much more effort to alias it. Although it's still possible to alias in unsafe code, one needs to write something like `&mut *(buf as *mut T)` to alias the reference, and the `as *mut T` part is much more visible as a red flag that unsafe aliasing is happening.

Here is another toy example showing how much effort is required to alias data in both cases (when the code is not globally wrapped in an `unsafe` block).

```rust
static mut FOO: [u8; 128] = [0; 128];
static mut BAR: [u8; 128] = [0; 128];

fn main() {
    foo();
    bar(unsafe { &mut BAR });
}

fn foo() {
    let x = unsafe { &mut FOO };
    let y = unsafe { &mut FOO };
    // Undefined behavior because x and y both alias FOO.
    concurrent(x, y);
}

[...]

Issue #1501, opened 2019-12-10
[RFC] trait for servicing interrupts


gendx:
## Observations

While working on chips from the nRF5x family, one thing that came up is that some chips support more interrupts than others. Currently, the code to service pending interrupts is in the `nrf52` crate.

https://github.com/tock/tock/blob/b22a314a8c55e8d8d7925d5e7eff5bbac0a0c07f/chips/nrf52/src/chip.rs#L52-L62

There is some common code and then a `match` over the interrupt number, that calls the suitable `handle_interrupt` function for the suitable driver.

With this architecture, an interrupt only defined on say `nrf52840` would have to be added to `nrf52`, so the driver would also need to be in `nrf52` instead of `nrf52840`. This isn't separating well various chips of the same family.

Alternatively, `nrf52840` could redefine the whole `chips.rs`, which would be a lot of duplicated code for maybe a few unique drivers.

## Draft Proposal

To make servicing of interrupts more flexible, an `InterruptServiceTrait` could be define as follows:

```rust
trait InterruptServiceTrait {
    unsafe fn service_interrupt(&self, interrupt: u32) -> bool;
}
```

- The function is `unsafe`, similarly to the `match` inside the current `service_pending_interrupts`.
- It returns a boolean, whether this interrupt number was serviced or not.

This allows composable interrupt services, for example `nrf52` defines a set of interrupt handlers, and on top of that `nrf52840` only defines its specific interrupts, returning `false` otherwise to indicate that the parent `nrf52` should handle the interrupt.

The chip's constructor, currently here:

https://github.com/tock/tock/blob/b22a314a8c55e8d8d7925d5e7eff5bbac0a0c07f/chips/nrf52/src/chip.rs#L22-L23

would be updated to take a `interrupt_service: &'static dyn InterruptServiceTrait` as parameter to service the interrupt. Then, each "end-user" chip such as `nrf52840` or `nrf52833` can pass a suitable `InterruptServiceTrait` to the generic `NRF52` chip constructor.

It seems that such a trait wouldn't be specific to Nordic chips but could be useful for all chips, as a trait in the kernel.

Any thoughts?

## Misc

Another trait that could be useful to clarify things would be an `InterruptHandler` trait that defines the `handle_interrupt` method. For now, drivers have this method by convention, but it isn't formalized as a trait. It could also be useful to put the "enable/disable all interrupts" pattern in this trait.

bradjc:
[...]

Issue #625, opened 2017-10-01
Unused chip drivers still take up RAM


brghena:
I was looking through RAM use on Signpost today, and wanted to mark this down while I was thinking about it. It's not a blocking problem in any way, but seems like undesirable behavior.

On Signpost (the Audio Module specifically) the USB client driver for the SAM4L takes up 288 Bytes of RAM even though it isn't in use and isn't included in the board's `main.rs` file. The RAM is spent on the USBC struct (allocated as `pub static mut` [here](https://github.com/helena-project/tock/blob/master/chips/sam4l/src/usbc/mod.rs#L870)).

My guess as to why it's taking space at all is because it's included in the `chip.rs` and `mod.rs` files in order to connect it to the interrupt handlers. This feels generally undesirable.

alevy:
> My guess as to why it's taking space at all is because it's included in the chip.rs and mod.rs files in order to connect it to the interrupt handlers. This feels generally undesirable.

It's indeed exactly that. I have changes (long overdue for master) I'm getting ready to make into a PR that change how we handle interrupts (as discussed on the mailing list previously), which will get us part of the way there (removing the ISRs array), but will not fix the issue in `chip.rs`'s `handle_interrup` dispatch code.

This seems like a job for some sort of macro, but i'm not totally sure how to do it.

enzuru:
@alevy is there any issue with me trying my hand at this ticket?

alevy:
@enzuru of course not!

hudson-ayers:
This is addressed by the design in #2069, but needs to be applied to every board/chip combo before this can be closed.

hudson-ayers:
Closed by #2069 and the other related PRs. 

For an example of how to create a minimal board that does not use all chip drivers, see this PR: https://github.com/tock/tock-bootloader/pull/19

PR #2069, opened 2020-08-07
Board-based instantiation of chip drivers and interrupt --> driver mapping for Apollo3 + SAM4L

hudson-ayers:
### Pull Request Overview

This pull request proposes a new design for managing chip drivers and interrupts.
The new design makes boards responsible for instantiating chip drivers, and for defining the mapping from interrupt numbers --> drivers.
It also adds the `InterruptService` trait to `kernel/src/platform/mod.rs`, as it will no longer only be used by nordic boards.

I would appreciate feedback before I implement this for all chips/boards, and answers to my outstanding questions below.

#### Advantages of the design:
- makes it possible for boards to exclude code for chip drivers that they are not using (e.g. solves #625)
- makes boards responsible for static initialization of all chip drivers, much in the same way that boards are responsible for initializing capsules, which removes reliance on global `static mut` variables in these scenarios (a step in the right direction for #1545 / #1662 ).
- If so desired, out-of-tree board authors can write chip drivers that live in the boards crate, and still use the upstream chip crate. 
- introduces macros that live in `chips/` to define the default interrupt mapping, so that upstream boards will not have to reimplement interrupt -> driver mappings in the common case.
- removes all dependence on `const_fn` throughout Tock, except for in `tock-register-interface`. Most uses of `const_fn` in Tock today are only required because of the widespread practice of initializing drivers as globals. By transitioning away from this design, most constructors can be made non-const, a big step towards #1654 .
- I think these changes would enable OpenTitan to use the upstream `earlgray` crate, and maintain only a board crate out-of-tree, which is probably a win for both projects.

So far, this PR is only implemented for the `apollo3`/`redboard_artemis_nano` combination (I picked it because there aren't very many apollo3 drivers at this point, so the total work was smallest).

I have verified that this PR excludes chip driver code in a way that is not currently possible on master, by removing the ble and i2c drivers from the board. Unfortunately removing const_fn everywhere seems to have introduced some code size cost.

### Testing Strategy

#### Outstanding Questions

1. How to handle deferred calls? I propose that they should be handled identically to interrupts in this design, that is `handle_deferred_call()` should be added to the `InterruptService` trait, and populated by a macro for the common case. -- Done.

2. ~~Is modifying the `Platform` trait ideal, or would it be better to use basically the `InterruptService` trait suggested in #1501, which could be implemented for the Platform, or for a struct containing just the drivers?~~ I decided that using `InterruptService` is clearer and makes more sense than conflating `Platform` with interrupt mapping. This technique allows for chips to still define some set of required interrupt mappings if it wishes, and should make adapting this to the nordic boards easier.

3. ~~Is it in fact desirable to remove all uses of `#[const_fn]`, or should I remove that portion of this PR and save it as a followup PR? It seems to me that const constructors would be preferable if the feature was stable, so maybe I am jumping the gun on this.~~ Decided to revert the const removals, for now.

4. Some drivers do not handle interrupts (`mcuctrl`, `pwrctrl`, `clkctrl` in the `Apollo3` case), but still do not need to be initialized as globals. I propose that when these drivers are only used in `main.rs`, they should simply be stack allocated in `reset_handler()`. If used in `chip.rs`, they should be made fields on the respective chip struct, and initialized with the Chip. This makes it clear which drivers are necessary for chip operation vs. optionally included by boards. In fact, this approach could be taken one step further, by making "required" drivers fields of the chip struct. Things like uart and timer which are prerequisites for including a board in Tock could be placed on the chip struct.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/apollo3/src/chip.rs
+++ b/chips/apollo3/src/chip.rs
@@ -3,31 +3,87 @@
 use core::fmt::Write;
 use cortexm4;
 use kernel::Chip;
+use kernel::InterruptService;
 
-use crate::ble;
-use crate::gpio;
-use crate::iom;
-use crate::nvic;
-use crate::stimer;
-use crate::uart;
-
-pub struct Apollo3 {
+pub struct Apollo3<I: InterruptService<()> + 'static> {
     mpu: cortexm4::mpu::MPU,
     userspace_kernel_boundary: cortexm4::syscall::SysCall,
     scheduler_timer: cortexm4::systick::SysTick,
+    interrupt_service: &'static I,
 }
 
-impl Apollo3 {
-    pub unsafe fn new() -> Apollo3 {
-        Apollo3 {
+impl<I: InterruptService<()> + 'static> Apollo3<I> {
+    pub unsafe fn new(interrupt_service: &'static I) -> Self {
+        Self {
             mpu: cortexm4::mpu::MPU::new(),
             userspace_kernel_boundary: cortexm4::syscall::SysCall::new(),
             scheduler_timer: cortexm4::systick::SysTick::new_with_calibration(48_000_000),
+            interrupt_service,
         }
     }
 }
 
-impl Chip for Apollo3 {
+/// This struct, when initialized, instantiates all peripheral drivers for the apollo3.
+/// If a board wishes to use only a subset of these peripherals, this
+/// should not be used or imported, and a modified version should be
+/// constructed manually in main.rs.
+pub struct Apollo3DefaultPeripherals {
+    pub stimer: crate::stimer::STimer<'static>,
+    pub uart0: crate::uart::Uart<'static>,
+    pub uart1: crate::uart::Uart<'static>,
+    pub gpio_port: crate::gpio::Port<'static>,
+    pub iom0: crate::iom::Iom<'static>,
+    pub iom1: crate::iom::Iom<'static>,
+    pub iom2: crate::iom::Iom<'static>,
+    pub iom3: crate::iom::Iom<'static>,
+    pub iom4: crate::iom::Iom<'static>,
+    pub iom5: crate::iom::Iom<'static>,
+    pub ble: crate::ble::Ble<'static>,
+}
+
+impl Apollo3DefaultPeripherals {
+    pub fn new() -> Self {
+        Self {
+            stimer: crate::stimer::STimer::new(),
+            uart0: crate::uart::Uart::new_uart_0(),
+            uart1: crate::uart::Uart::new_uart_1(),
+            gpio_port: crate::gpio::Port::new(),
+            iom0: crate::iom::Iom::new0(),
+            iom1: crate::iom::Iom::new1(),
+            iom2: crate::iom::Iom::new2(),
+            iom3: crate::iom::Iom::new3(),
+            iom4: crate::iom::Iom::new4(),
+            iom5: crate::iom::Iom::new5(),
+            ble: crate::ble::Ble::new(),
+        }
+    }
+}
+
+impl kernel::InterruptService<()> for Apollo3DefaultPeripherals {
+    unsafe fn service_interrupt(&self, interrupt: u32) -> bool {
+        use crate::nvic;
+        match interrupt {
+            nvic::STIMER..=nvic::STIMER_CMPR7 => self.stimer.handle_interrupt(),
+            nvic::UART0 => self.uart0.handle_interrupt(),
+            nvic::UART1 => self.uart1.handle_interrupt(),
+            nvic::GPIO => self.gpio_port.handle_interrupt(),
+            nvic::IOMSTR0 => self.iom0.handle_interrupt(),
+            nvic::IOMSTR1 => self.iom1.handle_interrupt(),
+            nvic::IOMSTR2 => self.iom2.handle_interrupt(),
+            nvic::IOMSTR3 => self.iom3.handle_interrupt(),
+            nvic::IOMSTR4 => self.iom4.handle_interrupt(),
+            nvic::IOMSTR5 => self.iom5.handle_interrupt(),
+            nvic::BLE => self.ble.handle_interrupt(),
+            _ => return false,
+        }
+        true
+    }
+    unsafe fn service_deferred_call(&self, _: ()) -> bool {
+        false
+    }
+}
+
+impl<I: InterruptService<()> + 'static> Chip for Apollo3<I> {
     type MPU = cortexm4::mpu::MPU;
     type UserspaceKernelBoundary = cortexm4::syscall::SysCall;
     type SchedulerTimer = cortexm4::systick::SysTick;
@@ -37,21 +93,8 @@ impl Chip for Apollo3 {
         unsafe {
             loop {
                 if let Some(interrupt) = cortexm4::nvic::next_pending() {
-                    match interrupt {
-                        nvic::STIMER..=nvic::STIMER_CMPR7 => stimer::STIMER.handle_interrupt(),
-                        nvic::UART0 => uart::UART0.handle_interrupt(),
-                        nvic::UART1 => uart::UART1.handle_interrupt(),
-                        nvic::GPIO => gpio::PORT.handle_interrupt(),
-                        nvic::IOMSTR0 => iom::IOM0.handle_interrupt(),
-                        nvic::IOMSTR1 => iom::IOM1.handle_interrupt(),
-                        nvic::IOMSTR2 => iom::IOM2.handle_interrupt(),
-                        nvic::IOMSTR3 => iom::IOM3.handle_interrupt(),
-                        nvic::IOMSTR4 => iom::IOM4.handle_interrupt(),
-                        nvic::IOMSTR5 => iom::IOM5.handle_interrupt(),
-                        nvic::BLE => ble::BLE.handle_interrupt(),
-                        _ => {
-                            panic!("unhandled interrupt {}", interrupt);
-                        }
+                    if !self.interrupt_service.service_interrupt(interrupt) {
+                        panic!("unhandled interrupt, {}", interrupt);
                     }
 
                     let n = cortexm4::nvic::Nvic::new(interrupt);
--- a/chips/apollo3/src/gpio.rs
+++ b/chips/apollo3/src/gpio.rs
@@ -8,11 +8,72 @@ use kernel::common::registers::{register_bitfields, register_structs, ReadWrite}
 use kernel::common::StaticRef;
 use kernel::hil::gpio;
 
+pub const GPIO_BASE_RAW: usize = 0x4001_0000; //safe to export outside crate
+
 const GPIO_BASE: StaticRef<GpioRegisters> =
-    unsafe { StaticRef::new(0x4001_0000 as *const GpioRegisters) };
+    unsafe { StaticRef::new(GPIO_BASE_RAW as *const GpioRegisters) };
 
 pub struct Port<'a> {
     pins: [GpioPin<'a>; 50],
+}
+
+impl<'a> Port<'a> {
+    pub const fn new() -> Self {
+        Self {
+            pins: [
+                GpioPin::new(GPIO_BASE, Pin::Pin00),
+                GpioPin::new(GPIO_BASE, Pin::Pin01),
+                GpioPin::new(GPIO_BASE, Pin::Pin02),
+                GpioPin::new(GPIO_BASE, Pin::Pin03),
+                GpioPin::new(GPIO_BASE, Pin::Pin04),
+                GpioPin::new(GPIO_BASE, Pin::Pin05),
+                GpioPin::new(GPIO_BASE, Pin::Pin06),
+                GpioPin::new(GPIO_BASE, Pin::Pin07),
+                GpioPin::new(GPIO_BASE, Pin::Pin08),
+                GpioPin::new(GPIO_BASE, Pin::Pin09),
+                GpioPin::new(GPIO_BASE, Pin::Pin10),
+                GpioPin::new(GPIO_BASE, Pin::Pin11),
+                GpioPin::new(GPIO_BASE, Pin::Pin12),
+                GpioPin::new(GPIO_BASE, Pin::Pin13),
+                GpioPin::new(GPIO_BASE, Pin::Pin14),
+                GpioPin::new(GPIO_BASE, Pin::Pin15),
+                GpioPin::new(GPIO_BASE, Pin::Pin16),
+                GpioPin::new(GPIO_BASE, Pin::Pin17),
+                GpioPin::new(GPIO_BASE, Pin::Pin18),
+                GpioPin::new(GPIO_BASE, Pin::Pin19),
+                GpioPin::new(GPIO_BASE, Pin::Pin20),
+                GpioPin::new(GPIO_BASE, Pin::Pin21),
+                GpioPin::new(GPIO_BASE, Pin::Pin22),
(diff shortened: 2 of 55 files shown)


bradjc:
I was skeptical at first, but now I think this is a very nice change that I would like to see merged.

My main comment is just around the naming. I think using "peripheral" instead of "driver" would be less confusing, since `Driver` is the trait for implementing a syscall. I also think the macro name should include "default".

1. Agreed.
2. Makes sense.
3. no comment
4. This would all still happen in the macro wouldn't it? So our current boards wouldn't see a difference? I'm not sure we need to specify anything at this point.

bradjc:
I think it would also make sense to start a "layer responsibilities" doc that describes what various layers are tasked with doing, and why. I think this PR adds a rather nonintuitive responsibility for board crates.

chips/apollo3/src/chip.rs
```diff
@@ -3,31 +3,27 @@
...
     mpu: cortexm4::mpu::MPU,
     userspace_kernel_boundary: cortexm4::syscall::SysCall,
     scheduler_timer: cortexm4::systick::SysTick,
+    platform: &'static P,
 }
 
-impl Apollo3 {
-    pub unsafe fn new() -> Apollo3 {
-        Apollo3 {
+impl<P: Platform + 'static> Apollo3<P> {
```
silvestrst: For `earlgrey`, I think it is worth now initialise the `chip_config::Config, from chip_config.rs` in the corresponding `board` and passing the structure as a parameter into `new`.

For OpenTitan it will also allow to remove the following features from the `.../earlgrey/Cargo.toml`:
```
[features]
config_fpga_nexysvideo = ["config_disable_default"]
config_sim_verilator = ["config_disable_default"]
config_disable_default = []
```
And instead have the config related features entirely in the board Cargo.toml.

P.S.

I know this is `apollo` related change, but seems like the most appropriate place to leave this comment.

chips/apollo3/src/gpio.rs
```diff
@@ -8,13 +8,72 @@ use kernel::common::registers::{register_bitfields, register_structs, ReadWrite}
 use kernel::common::StaticRef;
 use kernel::hil::gpio;
 
-const GPIO_BASE: StaticRef<GpioRegisters> =
+pub const GPIO_BASE: StaticRef<GpioRegisters> =
```
silvestrst: I think that base address, together with the register definitions should probably be split out in a separate `.rs`, as they remain the same regardless what driver we use. Allows to avoid reinventing the wheel for the "board" drivers.

Basically decouple implementation from the register and base definitions.

chips/apollo3/src/chip.rs
```diff
@@ -3,31 +3,89 @@
...
     userspace_kernel_boundary: cortexm4::syscall::SysCall,
     scheduler_timer: cortexm4::systick::SysTick,
+    interrupt_service: &'static I,
 }
 
-impl Apollo3 {
-    pub unsafe fn new() -> Apollo3 {
-        Apollo3 {
+impl<I: InterruptService + 'static> Apollo3<I> {
+    pub unsafe fn new(interrupt_service: &'static I) -> Self {
```
silvestrst, line 16: I think `apollo3_driver_definitions` would need to be passed into the chip, as the chip might want internally use a driver which has been overridden by the board. Otherwise we would have a mismatch of board using one  driver implementation, and the board another.

In this case the macro below, probably could just be a function, something like `apollo3_default_drivers_get` or something similar.

hudson-ayers: I think that if a chip needs to use a driver internally, that driver should be a field on the `Chip` struct itself, revealing that the Chip depends on that specific driver (and thereby showing that driver is not optional). If we want that driver to further be replaceable by multiple implementations, then introducing a trait for that particular driver would be appropriate. But I think we want to make it explicitly not possible for the chip to depend on the particular contents of whatever struct implements `InterruptService`, and instead depend only on the interface exposed by `InterruptService`.

chips/apollo3/src/uart.rs
```diff
@@ -9,14 +9,12 @@ use kernel::common::StaticRef;
...
+pub const UART0_BASE: StaticRef<UartRegisters> =
     unsafe { StaticRef::new(0x4001_C000 as *const UartRegisters) };
 
-pub static mut UART1: Uart = Uart::new(UART1_BASE);
-
-const UART1_BASE: StaticRef<UartRegisters> =
+pub const UART1_BASE: StaticRef<UartRegisters> =
     unsafe { StaticRef::new(0x4001_D000 as *const UartRegisters) };
 
 register_structs! {
```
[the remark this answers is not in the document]
bradjc: I think that can be done on a per-chip basis as needed. Having everything in one place is convenient in the common case.

chips/apollo3/src/chip.rs
bradjc: It seems like having that driver be an argument to `chip::new()` is reasonable to handle this case.

hudson-ayers:
I have added support for deferred calls, and implemented this for the sam4l + Imix. Doing so ended up requiring..a lot of code changes. I will do this for hail next, but after that I think it might be best to get this in as separate PRs for each subsequent chip.

Imix code size report from before this PR:
```
   text    data     bss     dec     hex filename
 174081    3272   21304  198657   30801 /home/hudson/tock/target/thumbv7em-none-eabi/release/imix
```

Imix code size report from after this PR:
```
   text    data     bss     dec     hex filename
 176641      64   32704  209409   33201 /home/hudson/tock/target/thumbv7em-none-eabi/release/imix
```

Noticeably, almost nothing left in `.data`, which is pretty cool to see, though much of that size appears to be moved into `.text`. But now it is possible to build a minimal board without the many sam4l drivers!

For a minimal "Imix_mini" board that only supports gpio + led + timer, with no modifications to the sam4l crate:

with this PR:
```
   text    data     bss     dec     hex filename
  52737       0   16384   69121   10e01 /home/hudson/tock/target/thumbv7em-none-eabi/release/imix_mini

Function groups (flash): 31491 bytes
  capsules::*          2418 bytes
  compiler_builtins::* 76 bytes
  components::*        304 bytes
  core::*              11655 bytes
  cortexm4::*          2200 bytes
  cortexm::*           1720 bytes
  imix_mini::*         288 bytes
  kernel::*            11628 bytes
  sam4l::*             1202 bytes
```

before this PR: 
```
   text    data     bss     dec     hex filename
  61953    3148   13232   78333   131fd /home/hudson/tock/target/thumbv7em-none-eabi/release/imix_mini2

Function groups (flash): 40379 bytes
  capsules::*          2414 bytes
  compiler_builtins::* 76 bytes
  components::*        308 bytes
  core::*              11947 bytes
  cortexm4::*          2200 bytes
  cortexm::*           1720 bytes
  imix_mini2::*        368 bytes
  kernel::*            16152 bytes
  sam4l::*             5100 bytes
  tock_cells::*        74 bytes
  tock_registers:      20 bytes
```

This was a rough/quick comparison, and does not exclude debug machinery or the power manager in either example. But it is at least clear that excluding drivers can provide meaningful benefit for code size, both by removing code size from drivers, and by reducing the portion of the core kernel code that is actually included in the final binary.

brghena:
Is the data just going into `.bss` instead? In the imix_mini example, it grows by the exact amount data shrinks, plus 4.

hudson-ayers:
> Is the data just going into `.bss` instead? In the imix_mini example, it grows by the exact amount data shrinks, plus 4.

Well, we set the size of .bss 

> Is the data just going into `.bss` instead? In the imix_mini example, it grows by the exact amount data shrinks, plus 4.

I am not sure..is this maybe an MPU alignment thing, where the RAM section is always going to be a power of 2, and the leftover RAM is given to uninitialized variables?

Have some numbers:

imix_mini2 (current master):
```
(.venv) hudson: ~/tock (no-more-globals) $ ./tools/print_tock_memory_usage.py --depth=3 -v target/thumbv7em-none-eabi/release/imix_mini2.elf
Tock memory usage report for target/thumbv7em-none-eabi/release/imix_mini2.elf
Kernel occupies 64852 bytes of flash
   61704        code and constant strings
    3148        variable initializers
Kernel occupies 16380 bytes of RAM
    8192        stack
    5040        uninitialized variables
    3148        initialized variables
    8188        variables total
Applications allocated 0 bytes of RAM

Variable groups (RAM): 3942 bytes
  imix_mini2::CHIP:                     4 bytes
  imix_mini2::PROCESSES:                32 bytes
  imix_mini2::io::WRITER:               1 bytes
  imix_mini2::reset_handler::BUF1:      24 bytes
  imix_mini2::reset_handler::BUF2:      20 bytes
  imix_mini2::reset_handler::BUF::*     300 bytes
  kernel::common::deferred_call:        4 bytes
  sam4l::adc::ADC0:                     52 bytes
  sam4l::aes::AES:                      40 bytes
  sam4l::ast::AST:                      12 bytes
  sam4l::crccu::CRCCU:                  548 bytes
  sam4l::dma::DMA_CHANNELS:             384 bytes
  sam4l::dma::NUM_ENABLED:              4 bytes
  sam4l::flashcalw::FLASH_CONTROLLER:   32 bytes
  sam4l::gpio::INTERRUPT_COUNT:         4 bytes
  sam4l::gpio::PA:                      516 bytes
  sam4l::gpio::PB:                      516 bytes
  sam4l::gpio::PC:                      516 bytes
  sam4l::i2c::I2C0:                     72 bytes
  sam4l::i2c::I2C1:                     72 bytes
  sam4l::i2c::I2C2:                     72 bytes
  sam4l::i2c::I2C3:                     72 bytes
  sam4l::pm::PM:                        8 bytes
  sam4l::spi::SPI:                      32 bytes
  sam4l::trng::TRNG:                    12 bytes
  sam4l::usart::IS_PANICING:            1 bytes
  sam4l::usart::USART0:                 56 bytes
  sam4l::usart::USART1:                 56 bytes
  sam4l::usart::USART2:                 56 bytes
  sam4l::usart::USART3:                 56 bytes
  sam4l::usbc::USBC:                    368 bytes
   ! 7 bytes of data or padding after sam4l::usart::IS_PANICING
   ! 3 bytes of data or padding after imix_mini2::io::WRITER
```

imix_mini (this PR):
```
(.venv) hudson: ~/tock (no-more-globals) $ ./tools/print_tock_memory_usage.py --depth=3 -v target/thumbv7em-none-eabi/release/imix_mini.elf
Tock memory usage report for target/thumbv7em-none-eabi/release/imix_mini.elf
Kernel occupies 52560 bytes of flash
   52560        code and constant strings
       0        variable initializers
Kernel occupies 16384 bytes of RAM
    8192        stack
    8192        uninitialized variables
       0        initialized variables
    8192        variables total
Applications allocated 0 bytes of RAM

Variable groups (RAM): 1966 bytes
  imix_mini::CHIP:                  4 bytes
  imix_mini::PROCESSES:             32 bytes
  imix_mini::io::WRITER:            1 bytes
  imix_mini::reset_handler::BUF1:   24 bytes
  imix_mini::reset_handler::BUF2:   20 bytes
  imix_mini::reset_handler::BUF::*  1876 bytes
  kernel::common::deferred_call:    4 bytes
  sam4l::gpio::INTERRUPT_COUNT:     4 bytes
  sam4l::usart::IS_PANICING:        1 bytes
   ! 4 bytes of data or padding after imix_mini::reset_handler::BUF
   ! 2 bytes of data or padding after sam4l::usart::IS_PANICING
```

Clearly some of .data has moved to .bss, such as the gpio::PA struct. But I don't see why, say, there would be space for an unitinitialized USBC struct, unless I have done something wrong.

bradjc:
Hmm...given that the sam4l has 8x more flash than RAM, this seems like a change we _don't_ want.

RAM usage goes up by (32704-21304) - (3272-64) = 8192 bytes.

hudson-ayers:
> Hmm...given that the sam4l has 8x more flash than RAM, this seems like a change we _don't_ want.
> 
> RAM usage goes up by (32704-21304) - (3272-64) = 8192 bytes.

I think this is just an artifact of MPU alignment. If i decrease the stack size on Imix by 100 bytes, then the bss size drops to 24512 (a decrease of 8192 bytes, leaving the total RAM usage the same as without this PR). So the impact of this change on RAM usage is much smaller than that in reality.

hudson-ayers:
I think this is ready for review, for apollo3 + sam4l. I have reverted the `const_fn` removal based on the core call discussion today.

By playing with the stack size on current master and this PR to see what stack size marks the boundary where RAM use increases by 8k, I have determined that this PR actually increases memory usage by about 175 bytes.

One thing to pay attention to: constructors of chip peripherals are now public, so they can be called from boards. This has the potential to be a bit of a footgun, as only one instance of the peripheral can receive interrupts. We could potentially prevent this with a capability, but for now I think we should just see how it works out.

alistair23 (dismissed):
I'm not a fan of the giant macro but otherwise I think this is a good improvement. The current way of handling interrupts has always annoyed me and I think this is a bit better.

I reviewed the Apollo3 parts, no the SAM4L though.

hudson-ayers:
I have removed the large macros, and just use structs defined in chips.rs instead. I also added a little documentation to `Porting.md`.

phil-levis (changes requested):
Can you fix the merge conflicts?

alistair23 (approved):
Looks good for the Apollo3

hudson-ayers:
> Can you fix the merge conflicts?

done

alistair23 (approved):

phil-levis:
Some new conflicts, resolve and I'm ready to approve.

bradjc:
@hudson-ayers Have you verified that the non-macro approach does make a smaller binary if you do not use all peripherals?

hudson-ayers:
> @hudson-ayers Have you verified that the non-macro approach does make a smaller binary if you do not use all peripherals?

Just did, the results seem pretty much the same as using the macro approach -- still significant savings.

hudson-ayers:
Squashed/rebased

alistair23 (dismissed):

libraries/tock-cells/src/lib.rs
```diff
@@ -1,14 +1,13 @@
...
-#![feature(const_fn)]
 // Feature used to opt-in the new `core::Option::contains()` API.
 //
 // This feature can be removed if needed by manually reimplementing the
 // `contains` logic for `Option`.
 //
 // Tock expects this feature to stabilize in the near future.
 // Tracking: https://github.com/rust-lang/rust/issues/62358
-#![feature(option_result_contains)]
+#![feature(option_result_contains, const_fn)]
```
bradjc, line 10: Can you undo this?

hudson-ayers: yeah, sorry

bradjc:
I'm ready to merge this once its rebased. @phil-levis?

alistair23 (approved):

hudson-ayers:
rebased

alistair23 (approved):

bradjc (approved):

hudson-ayers:
looks like a network connection error.

bors retry

hudson-ayers:
https://git.qemu.org/ is down, so the QEMU CI cannot pass. But neither of these chips is one of the QEMU tested chips anyway, so I think this is safe to manually merge.

alistair23:
That's a pain. We could use the GitHub mirror instead?

hudson-ayers:
Given that we already have to rely on Github being up to run CI, that would probably be an improvement. This may be a sufficiently rare occurrence it isn't worth investing the time to switch it though.

PR #2071, opened 2020-08-13
Update tock-registers doc to fix mistake

brghena:
Default is u32, and register fields in the example were greater than 8 bits in size, so we should just use 32-bits as the register size in the example.

### Pull Request Overview

Modifies tock-registers README.

### Testing Strategy

N/A

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [N/A] Ran `make prepush`.


brghena:
@alistair23 any ideas why the QEMU check is failing?

alistair23:
The CI should be fixed now.

PR #2074, opened 2020-08-14
Add ability to get full address bytes & address type to nrf52 Ficr

daboross:
### Pull Request Overview

This adds the ability to get the full address as bytes (as opposed to part of the address, or the full address as a string), and a secondary function to get the address type.

I'm mainly interested in using this in order to use the device's address as the device's Bluetooth LE address.

I added the "full_address" function alongside the existing address function for backwards compatibility, but I think it might be better overall to expose just one function returning a `[u8; 6]`. As far as I can tell from the documentation, deviceaddr0 and deviceaddr1 are always meant to be used as a pair?

### Testing Strategy

Printing the device address & address type on a local nrf52840dk module

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


chips/nrf52/src/ficr.rs
```diff
@@ -376,6 +383,28 @@ impl Ficr {
         regs.deviceaddr0.read(DeviceAddress0::DEVICEADDRESS)
     }
 
+    pub fn full_address(&self) -> [u8; 6] {
```
bradjc, line 386: Can we just call this `address()` and remove the address function that exists? I'm not sure what the old version is for, and `make allcheck` still works after I deleted it.

Otherwise looks good.

daboross: Sounds good. I've renamed it!

And thanks for looking over this!

PR #2049, opened 2020-07-28
Integrate HD44780 into Screen HIL

teonaseverin:
### Pull Request Overview

This pull request integrates the HD44780 LCD screen into the screen HIL. It adds the TEXT pixel format, as well as a new function to the `hil::screen::ScreenSetup` trait: `fn screen_command()`. This function can be used to send specific driver commands to the device, that cannot be sent through the API exported by the screen HIL. Such commands, for example for HD44780, can be clearing everything written on the screen, making the cursor visible on the screen or making the cursor blink. 
By integrating the capsule into the screen HIL, the userspace will comunicate with it only through the screen syscalls and not by using the hd44780's specific syscalls. Because of that, I removed the Driver implementation from the capsule. 

### Testing Strategy

This pull request was tested using a NUCLEO-F429ZI board. 

### Documentation Updated

- [x] Removed the documentation about the capsule's syscalls from /docs/syscalls/

### Formatting

- [x] Ran `make prepush`.

Issue #1074, opened 2018-07-05
Tracking: Generic over static lifetimes in HIL interfaces


ppannuto:
HILs with `'static` not in the context of buffers:

- [x] adc (#3455)
- [x] ble_advertising.rs (#1964)
- [x] crc.rs (#1956)
- [x] gpio.rs (#1965)
- [x] i2c.rs (#3467)
- [x] radio.rs (#3459)
- [x] screen (#3461)
- [x] sensors.rs (#1959)
- [x] spi.rs (#3460)

Early in Tock we used `'static` lifetimes fairly liberally. As interfaces matured, we shifted towards generic lifetimes (`'a`) for the flexibility it (could eventually) affords components. Unfortunately, many of the HILs currently specific static rather than generic lifetimes.

As noted in (2) of #1069, one example of a negative consequence is an artificial restriction of where and how clients can be set.

I believe that the use of static lifetimes is largely a historical artifact, however I wanted to open this as an RFC in case this is something folks feel we need to discuss further before adapting.

## Current Status

From a highly scientific and imprecise study, we have more HILs with static lifetimes than generic, but I think that can be attributed to history. Loosely, the newer HILs seem to prefer generic lifetimes:

```bash
# Static
$ git grep static | grep -v '//!' | cut -d':' -f1 | sort | uniq -c
   9 adc.rs
   5 ble_advertising.rs
   4 flash.rs
  11 i2c.rs
   4 led.rs
   1 ninedof.rs
   5 nonvolatile_storage.rs
  12 radio.rs
   4 sensors.rs
  14 spi.rs
   3 symmetric_encryption.rs
   6 uart.rs
# Generic
$ git grep "'a" | grep -v '//!' | cut -d':' -f1 | sort | uniq -c
   2 flash.rs
  10 symmetric_encryption.rs
```

bradjc:
I thought we had a general notion in Tock that changes have to be used to be added. Is there a use for the generic lifetimes?

ppannuto:
What started me down this path was trying to move the `set_client` call for the UART from the top-level board file into the `initialize` method of the console capsule. We currently can't do that because the UART HIL requires static lifetimes but the console capsule has a generic lifetime.

I suspect there are other uses that we aren't currently doing because we can't. Let me flip the question, is there a reason that the HILs should be enforcing static lifetimes rather than permitting generic ones?

bradjc:
[...]

PR #2079, opened 2020-08-17
Added TextScreen HIL and driver and integrated HD44780 capsule

teonaseverin:
### Pull Request Overview

This pull request adds a new TextScreen HIL, to be used with text screen displays, like HD44780, as @bradjc suggested on #2049 . I have also implemented the `text_screen` capsule and component and modified the `hd44780` capsule accordingly. The userspace will comunicate with the hd44780 capsule only through the text_screen syscalls and not by using the its specific syscalls. I have removed the Driver implementation from the capsule.

The functions exported by the TextScreen Trait are for:
1. controlling the screen: set the position of the cursor, show/hide the cursor or turn on/off the blinking cursor
2. printing a message on the screen. 

This PR closes #2049 .

### Testing Strategy

This pull request was tested using a NUCLEO-F429ZI board.
### Documentation Updated

- [x] Removed the documentation about the capsule's syscalls from /docs/syscalls/

### Formatting

- [x] Ran `make prepush`.


stappersg:
@teonaseverin FYI /  FWIW  CI discovered conflicting files

kernel/src/hil/text_screen.rs
```diff
@@ -0,0 +1,99 @@
+//! Interface for text screen and displays.
+use crate::returncode::ReturnCode;
+
+pub trait TextScreen {
+    fn set_client(&self, client: Option<&'static dyn TextScreenClient>);
+
+    /// Returns a tupel (width, height) with the resolution of the
```
bradjc, line 7:
```suggestion
    /// Returns a tuple (width, height) with the resolution of the
```

kernel/src/hil/text_screen.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    /// Sends to the driver a command to show the cursor. When finised,
+    /// the driver will call the `command_complete()` callback.
+    ///
+    /// Return values:
+    /// - `SUCCESS`: The command is valid and will be sent to the driver.
+    /// - `EBUSY`: Another command is in progress.
+    fn show_cursor(&self) -> ReturnCode;
+
+    /// Sends to the driver a command to turn on the blinking cursor. When finised,
```
bradjc, line 48:
```suggestion
    /// Sends to the driver a command to turn on the blinking cursor. When finished,
```

kernel/src/hil/text_screen.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    /// Sends to the driver a command to hide the cursor. When finised,
+    /// the driver will call the `command_complete()` callback.
+    ///
+    /// Return values:
+    /// - `SUCCESS`: The command is valid and will be sent to the driver.
+    /// - `EBUSY`: Another command is in progress.
+    fn hide_cursor(&self) -> ReturnCode;
+
+    /// Sends to the driver a command to show the cursor. When finised,
```
bradjc, line 40:
```suggestion
    /// Sends to the driver a command to show the cursor. When finished,
```

kernel/src/hil/text_screen.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// Sends to the driver a command to set the cursor at a given position
+    /// (x_position, y_position). When finished, the driver will call the
+    /// `command_complete()` callback.
+    ///
+    /// Return values:
+    /// - `SUCCESS`: The command is valid and will be sent to the driver.
+    /// - `EBUSY`: Another command is in progress.
+    fn set_cursor(&self, x_position: usize, y_position: usize) -> ReturnCode;
+
+    /// Sends to the driver a command to hide the cursor. When finised,
```
bradjc, line 32:
```suggestion
    /// Sends to the driver a command to hide the cursor. When finished,
```

kernel/src/hil/text_screen.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    /// Sends to the driver a command to turn on the blinking cursor. When finised,
+    /// the driver will call the `command_complete()` callback.
+    ///
+    /// Return values:
+    /// - `SUCCESS`: The command is valid and will be sent to the driver.
+    /// - `EBUSY`: Another command is in progress.
+    fn blink_cursor_on(&self) -> ReturnCode;
+
+    /// Sends to the driver a command to turn off the blinking cursor. When finised,
```
bradjc, line 56:
```suggestion
    /// Sends to the driver a command to turn off the blinking cursor. When finished,
```

kernel/src/hil/text_screen.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// the driver will call the `command_complete()` callback.
+    ///
+    /// Return values:
+    /// - `SUCCESS`: The command is valid and will be sent to the driver.
+    /// - `EBUSY`: Another command is in progress.
+    fn blink_cursor_off(&self) -> ReturnCode;
+
+    /// Sends to the driver a command to turn on the display of the screen.
+    /// When finised, the driver will call the `command_complete()` callback.
```
bradjc, line 65:
```suggestion
    /// When finished, the driver will call the `command_complete()` callback.
```

kernel/src/hil/text_screen.rs
```diff
@@ -0,0 +1,99 @@
...
+    ///
+    /// Return values:
+    /// - `SUCCESS`: The command is valid and will be sent to the driver.
+    /// - `EBUSY`: Another command is in progress.
+    fn display_on(&self) -> ReturnCode;
+
+    /// Sends to the driver a command to turn off the display of the screen.
+    /// When finised, the driver will call the `command_complete()` callback.
```
bradjc, line 73:
```suggestion
    /// When finished, the driver will call the `command_complete()` callback.
```

kernel/src/hil/text_screen.rs
```diff
@@ -0,0 +1,99 @@
...
+    ///
+    /// Return values:
+    /// - `SUCCESS`: The command is valid and will be sent to the driver.
+    /// - `EBUSY`: Another command is in progress.
+    fn display_off(&self) -> ReturnCode;
+
+    /// Sends to the driver a command to cleat the display of the screen.
+    /// When finised, the driver will call the `command_complete()` callback.
```
bradjc, line 81: Does the cursor reset on a clear?

kernel/src/hil/text_screen.rs
```diff
@@ -0,0 +1,99 @@
...
+pub trait TextScreenClient {
+    /// The driver calls this function when any command (but a write one)
+    /// finishes executing.
+    fn command_complete(&self, r: ReturnCode);
+
+    /// The driver calls this function when a write command finishes executing.
+    fn write_complete(&self, buffer: &'static mut [u8], r: ReturnCode);
+
+    /// The driver calls this function when the screen is ready to use after power on.
+    fn text_screen_is_ready(&self);
```
bradjc, line 98: This callback is a bit unexpected for two reasons. One, there is no `init()` (or similar) function in the HIL. So there would have to be some external code path the causes this callback to occur, _and_ it would have to occur after `set_client()` is called. Two, it's not clear why a client would care about this event. Presumably the only thing stopping the display from being ready is something actually trying to use it, and if so, then the display can just configure itself the first time the client tries to use it. In the case some external event has to happen (like maybe a person has to push a button or something) then I could see this callback being useful.

I think for simplicity this callback can just be removed.

bradjc (commented):
Looks good, I just have a question about the screen is ready callback.

hudson-ayers:
Something seems to have gone wrong with a rebase or merge, as this now changes 17K lines of code

teonaseverin:
> Something seems to have gone wrong with a rebase or merge, as this now changes 17K lines of code

I just saw that, let me try to fix it

kernel/src/hil/text_screen.rs
teonaseverin: Hi! So, I wrote this callback because of two reasons. First, I saw that the `screen` hil has a similar callback, also with no `init()` function. Also, these text screen usually have a power-on sequence that needs to be done before the user can interact with the screen. 

At the moment, there is a `init()` function in the capsule that implements this hil, function that is called from the `main.rs` file, after the `set_client()` is called and executed. 

When I wrote this, I was thinking it would be a good idea if there were a mechanism to stop the capsule from executing commands from the user until that power-on sequence was done, so the hil won't send any command to the screen until the `text_screen_is_ready()` is called. But I believe I could force the init sequence to be executed with the first command from the user.

Tell me what you think and thank you!

kernel/src/hil/text_screen.rs
teonaseverin: Yes. And everything on the screen will be flushed.

capsules/src/screen.rs
```diff
@@ -563,7 +563,6 @@ impl<'a> Driver for Screen<'a> {
             200 => self.enqueue_command(ScreenCommand::Write, data1, data2, appid),
             // Fill
             300 => self.enqueue_command(ScreenCommand::Fill, data1, data2, appid),
-
             _ => ReturnCode::ENOSUPPORT,
```
bradjc, line 566:
```suggestion

            _ => ReturnCode::ENOSUPPORT,
```

capsules/src/hd44780.rs
```diff
@@ -1,31 +1,31 @@
-//! Provides userspace access to LCD connected on the board, but defined in the
-//! kernel.
+//! Provides userspace access (through Screen capsule and userspace library)
+//! to LCD connected on the board, but defined in the kernel.
```
bradjc, line 2:
```suggestion
//! Driver for the HD44780 LCD screen.
```

kernel/src/hil/text_screen.rs
alexandruradovici: Done, the screen now gets initialized when the first `display_on` command is issued.

capsules/src/text_screen.rs
```diff
@@ -0,0 +1,290 @@
...
+        } else {
+            self.current_app.take().map(|appid| {
+                let _ = self.apps.enter(appid, |app, _| {
+                    app.pending_command = false;
+                    app.callback.map(|mut cb| {
+                        cb.schedule(data1, data2, data3);
+                    });
+                });
+            });
+        }
```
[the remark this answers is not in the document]
alexandruradovici: Done.

hudson-ayers (dismissed):
I have not reviewed this in its entirety, but all of my previous comments have been addressed and I am happy with the HIL. Would be good to get this merged so the 2.0 port for the hd44780 can be based on this version

alexandruradovici:
As soon as it is merged, I will port it to the 2.0 API.

teonaseverin:
I have tested the TextScreen HIL and driver and the HD44780 capsule after the latest modifications. I have solved any problem I could find. It works well now.

kernel/src/hil/text_screen.rs
```diff
@@ -0,0 +1,96 @@
+//! Interface for text screen and displays.
+use crate::returncode::ReturnCode;
+
+pub trait TextScreen {
+    fn set_client(&self, client: Option<&'static dyn TextScreenClient>);
```
bradjc, line 5:
```suggestion
pub trait TextScreen<'a> {
    fn set_client(&self, client: Option<&'a dyn TextScreenClient>);
```

I didn't notice this before, but browsing https://github.com/tock/tock/issues/1074 and I remembered that we want to avoid using `&'static` for clients.

alexandruradovici: Done

hudson-ayers (approved):

bradjc (approved):

PR #2080, opened 2020-08-19
OpenTitan: Add support for detecting the USB device

alistair23:
### Pull Request Overview

This allows me to see the OpenTitan USB device when running Tock.

There still seems to be some timing issues and the device doesn't always appear, but at least this is better then the current nothing.

### Testing Strategy

Plugging in the OpenTitan FPGA.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/chips/lowrisc/src/usbdev.rs
+++ b/chips/lowrisc/src/usbdev.rs
@@ -541,6 +541,107 @@ impl<'a> Usb<'a> {
         });
     }
 
+    fn control_ep_receive(&self, ep: usize, buf_id: usize, size: u32, setup: u32) {
+        let hw_buf = self.registers.buffer[buf_id * 8].extract();
+
+        match self.descriptors[ep].state.get() {
+            EndpointState::Disabled => unimplemented!(),
+            EndpointState::Ctrl(state) => {
+                let request_type = hw_buf.read(BUFFER::REQUEST_TYPE);
+                let length = hw_buf.read(BUFFER::LENGTH);
+
+                let ep_buf = &self.descriptors[ep].slice_out;
+                let ep_buf = ep_buf.expect("No OUT slice set for this descriptor");
+                if ep_buf.len() < 8 {
+                    panic!("EP0 DMA buffer length < 8");
+                }
+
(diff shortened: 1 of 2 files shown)


bradjc (dismissed):

alistair23:
Ping!

chips/lowrisc/src/usbdev.rs
```diff
@@ -422,6 +463,199 @@ impl<'a> Usb<'a> {
...
+        self.bufs.set(bufs);
+    }
+
+    fn stall(&self, endpoint: usize) {
+        self.registers
+            .stall
+            .set(1 << endpoint | self.registers.stall.get());
+    }
+
+    fn copy_slice_out_to_hw(&self, ep: usize, buf_id: usize, size: usize) {
```
phil-levis, line 476: This should use enumerations rather than usize, for ep and buf_id, to allow compile-time rather than runtime checking.

chips/lowrisc/src/usbdev.rs
```diff
@@ -422,6 +463,199 @@ impl<'a> Usb<'a> {
...
+        }
+
+        self.registers.configin[ep].write(
+            CONFIGIN::BUFFER.val(buf_id as u32)
+                + CONFIGIN::SIZE.val(size as u32)
+                + CONFIGIN::RDY::SET,
+        );
+    }
+
+    fn copy_from_hw(&self, ep: usize, buf_id: usize, size: usize) {
```
phil-levis, line 517: Same as copy_slice_out_to_hw.

phil-levis (changes requested):

chips/lowrisc/src/usbdev.rs
alistair23: That would required converting from the uszie from the HIL to a enum, what advantage does that give us? It is then written to the hardware as just a number.

alistair23:
Ping!

phil-levis (approved):
After discussion with Alistair, I agree that the issue is with the HIL passing numbers and that this PR should not block on that.

PR #2081, opened 2020-08-19
MSP432: Change SMCLK from 750kHz to 1.5MHz

lebakassemmerl:
### Pull Request Overview

The SMCLK is currently used by the UART-module. It turned out that with 750kHz the baud-error was in a range where the characters could be transmitted correctly, but at receiving characters the error was too high so that the received data was not correct anymore. By changing the clock from 750kHz to 1.5MHz also receiving is now working correctly

### Testing Strategy

Testing on the eval-board by receiving characters over UART.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


chips/msp432/src/cs.rs
```diff
@@ -287,14 +287,14 @@ impl ClockSystem {
     }
 
     // Setup the low-speed subsystem master clock (SMCLK) to 1/64 of the master-clock -> 750kHz
```
hudson-ayers, line 289:
```suggestion
    // Setup the low-speed subsystem master clock (SMCLK) to 1/32 of the master-clock -> 1.5MHz
```

PR #2082, opened 2020-08-22
capsules: usb/cdc: Fixup the CDC control message logic

alistair23:
### Pull Request Overview

Update the comment to use SET_CONTROL_LINE_STATE to match the phrase in
the spec. As well as that accept the request no matter what the value
is, but let's document what the options are.

### Testing Strategy

Test USB CDC on OpenTitan.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/usb/cdc.rs
+++ b/capsules/src/usb/cdc.rs
@@ -285,12 +285,7 @@ impl<'a, U: hil::usb::UsbController<'a>> hil::usb::Client<'a> for CdcAcm<'a, U> 
                     // Currently we don't care about the value
 
                     if self.state.get() != State::Connected {
-                        // We weren't previously connected so this must mean
-                        // a client connected.
                         self.state.set(State::Connecting);
-                    } else {
-                        // We were connected, so disconnect.
-                        self.state.set(State::Enumerated)
                     }
                 }
                 0x23 => {
--- a/capsules/src/usb/cdc.rs
+++ b/capsules/src/usb/cdc.rs
@@ -269,29 +269,28 @@ impl<'a, U: hil::usb::UsbController<'a>> hil::usb::Client<'a> for CdcAcm<'a, U> 
     fn ctrl_setup(&'a self, endpoint: usize) -> hil::usb::CtrlSetupResult {
         descriptors::SetupData::get(&self.client_ctrl.ctrl_buffer.buf).map(|setup_data| {
             let b_request = setup_data.request_code;
-            let value = setup_data.value;
 
             // Match on the two CDC control messages we care about:
-            // - `0x22`: SET_LINE_CONTROL_STATE
+            // - `0x22`: SET_CONTROL_LINE_STATE
             // - `0x23`: SEND_BREAK
             match b_request {
                 0x22 => {
-                    // This message seems to come with different values. We
-                    // don't care about the actual value's meaning, except for
-                    // value=0 which seems to bookend when the CDC client is
-                    // actually attached.
-                    if value == 0 {
-                        // On Linux, this seems to be a good signal of both when
-                        // a client connects and disconnects. So, based on our
-                        // current state, we can update our state.
-                        if self.state.get() != State::Connected {
-                            // We weren't previously connected so this must mean
-                            // a client connected.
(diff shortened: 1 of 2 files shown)


hudson-ayers:
cc @gendx , just to confirm that always accepting the request won't be a problem for OpenSK.

bradjc:
Why the change in the logic? I don't think my approach was particularly robust, but I do remember getting multiple of the various control messages.

alistair23:
My Linux (5.7) kernel sends a value of 3, so with the current logic the CDC device doesn't work.

bradjc:
On my mac with tockloader, I get two CONTROL_LINE_STATE messages, 0x02 and then 0x00. With miniterm I get one, just 0x03. So with this change tockloader doesn't work for me. Maybe we should use this change, but remove the disconnect case in the if statement?

gendx:
> cc @gendx , just to confirm that always accepting the request won't be a problem for OpenSK.

I didn't test the CDC capsule, but it shouldn't be a problem when that capsule is disabled. I didn't check the latest developments, but I don't think the current Tock implementation allows to have multiple USB capsules running at the same time anyway.

alistair23:
I have removed the disconnect logic.

capsules/src/usb/cdc.rs
```diff
@@ -269,29 +269,23 @@ impl<'a, U: hil::usb::UsbController<'a>> hil::usb::Client<'a> for CdcAcm<'a, U>
...
         descriptors::SetupData::get(&self.client_ctrl.ctrl_buffer.buf).map(|setup_data| {
             let b_request = setup_data.request_code;
-            let value = setup_data.value;
 
             // Match on the two CDC control messages we care about:
-            // - `0x22`: SET_LINE_CONTROL_STATE
+            // - `0x22`: SET_CONTROL_LINE_STATE
             // - `0x23`: SEND_BREAK
             match b_request {
                 0x22 => {
```
phil-levis, line 277: These should be an enum, not embedded constants.

bradjc:
My proposed change doesn't work either :(.

I was hoping to avoid changing tockloader, but it seems like I will have to. There doesn't seem to be any way for the host to signal over CDC-ACM "ok, you may now send serial data to me". The serial library in python has this pesky line in it: https://github.com/pyserial/pyserial/blob/master/serial/serialposix.py#L352 which seems to be the issue for tockloader. Basically, anything received before all configuration is finished is just dropped. And since each serial tool has its own setup process, the device can't really know when it is safe to send.

So, my updated approach is:
- On the device, wait for the host to configure the baud rate to 115200 before trying to send anything. This will be useful for using baud rate configuration to enter bootloader mode in the future.
- On the host, edit tockloader to remove the call to `reset_input_buffer()` inside the serial library.

bradjc:
I couldn't push to this PR so I put my changes here: https://github.com/tock/tock/compare/pr/2082

alistair23:
I have added your patches @bradjc and converted the values to enums.

capsules/src/usb/cdc.rs
```diff
@@ -63,6 +63,34 @@ enum State {
...
+#[derive(Debug, Copy, Clone, PartialEq)]
+enum CtrlState {
+    /// No ongoing ctrl transcation.
+    Idle,
+    /// Host has sent a SET_LINE_CODING configuration request.
+    SetLineCoding,
+}
+
+#[derive(PartialEq)]
+enum CDCCntrlMessage {
```
gendx, line 76: Nit pick: "ctrl" is a much more common abbreviation for "control". Otherwise, is saving two letters worth it over fully spelling "control"?

capsules/src/usb/cdc.rs
```diff
@@ -211,6 +243,7 @@ impl<'a, U: hil::usb::UsbController<'a>> CdcAcm<'a, U> {
                 Buffer64::default(),
             ],
             state: Cell::new(State::Disabled),
+            ctrl_state: Cell::new(CtrlState::Idle),
```
gendx, line 246: And here "ctrl" is used.

bradjc (approved):

PR #2083, opened 2020-08-24
Stm32f3: add flash support

krady21:
### Pull Request Overview

This pull request adds flash memory support for stm32f303xc (embedded flash and option bytes). 

### Testing Strategy

This pull request was tested using a stm32f3discovery kit. All the main operations have been tested and work properly (read, write, erase).

### TODO or Help Wanted

I am still somewhat unsure about how to treat the callbacks from writing and erasing the option bytes.

### Documentation Updated

-  [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make prepush`.


chips/stm32f303xc/src/flash.rs
```diff
@@ -0,0 +1,608 @@
...
+
+/// This is a wrapper around a u8 array that is sized to a single page for the
+/// stm32f303xc. Users of this module must pass an object of this type to use the
+/// `hil::flash::Flash` interface.
+///
+/// An example looks like:
+///
+/// ```rust
+/// # extern crate stm32f303xc;
+/// # use stm32f303xc::flash::NrfPage;
```
hudson-ayers, line 232:
```suggestion
/// # use stm32f303xc::flash::StmF303Page;
```

chips/stm32f303xc/src/flash.rs
```diff
@@ -0,0 +1,608 @@
...
+    }
+
+    /// Allows programming the 8 option bytes:
+    /// 0: RDP, 1: USER, 2: DATA0, 3:DATA1, 4. WRP0, 5: WRP1, 6.WRP2, 7. WRP3
+    pub fn write_option(&self, byte_number: usize, value: u8) -> ReturnCode {
+        if byte_number > 7 {
+            return ReturnCode::EINVAL;
+        }
+
+        while self.registers.sr.is_set(Status::BSY) {}
```
hudson-ayers, line 542: Tock tries to avoid blocking interfaces when possible (see https://github.com/tock/tock/blob/master/doc/Design.md#split-phase-operation), with an exception for cases where a datasheet indicates the cycles spent blocking will be less than the cycles it would take to set up and process a callback. Can you provide a pointer here on how long this blocking is expected to last, or why a split-phase design is infeasible? Same for other instances of spinning on the status bit.

chips/stm32f303xc/src/flash.rs
```diff
@@ -0,0 +1,608 @@
...
+                self.registers.cr.modify(Control::PG::CLEAR);
+            }
+
+            if self.registers.cr.is_set(Control::OPTPG) {
+                self.registers.cr.modify(Control::OPTPG::CLEAR);
+            }
+
+            // Reset by writing a 1.
+            self.registers.sr.modify(Status::WRPRTERR::SET);
+            panic!("WRPRTERR: programming a write-protected address.");
```
hudson-ayers, line 408: Writing a 1 to an already set register causes a reset? What does it reset? Also, why did you choose to panic here instead of returning an error? I assume that if the address was write protected no write actually occurred.

chips/stm32f303xc/src/flash.rs
```diff
@@ -0,0 +1,608 @@
...
+                self.registers.cr.modify(Control::PG::CLEAR);
+            }
+
+            if self.registers.cr.is_set(Control::OPTPG) {
+                self.registers.cr.modify(Control::OPTPG::CLEAR);
+            }
+
+            // Reset by writing a 1.
+            self.registers.sr.modify(Status::PGERR::SET);
+            panic!("PGERR: address contains a value different from 0xFFFF before programming.");
```
hudson-ayers, line 422: also confused here by the motivation for modifying all these registers only to panic

krady21: Forgot to delete the panics. I used them for debugging here.

chips/stm32f303xc/src/flash.rs
krady21: My bad. What i actually meant was that if you want to clear WRPRTERR, PGERR or EOP (all the bits that are set by hardware interrupts) you actually have to write a 1 instead of a 0, hence the CLEAR instead of SET. I will change the comment to "Cleared by writing a 1".

chips/stm32f303xc/src/flash.rs
krady21: Btw, do you think i should replace them with debugs or should I just delete them outright?

hudson-ayers: Do each of these conditions indicate that the (read/write/erase) operation failed? If so, I would say that it is probably best to deliver an error to the client via the appropriate callback.

krady21: Yes. Both PGERR and WRPTERR indicate failure. I guess that I will use the write_complete callback as I have done further above.

chips/stm32f303xc/src/flash.rs
krady21: At first, I implemented all the operations with deferred calls. In that case it was necessary to wait for the BSY bit. It was just that after switching to hardware interrupts, I forgot get to delete them. I haven't found any mention in the datasheet about time spent blocking being less than the cycles need to process the callback. After testing the code without the while loops, I am getting the same results, so I suppose I will just delete them. Thanks for the heads-up.

chips/stm32f303xc/src/flash.rs
hudson-ayers: sounds good!

krady21: While we are here, it would be nice to make the kernel flash Error enum generic so that users could define specific error messages. Here I will have to send the same FlashError via callback for two different issues(page write protection vs. page not being erased before programming)

chips/stm32f303xc/src/flash.rs
```diff
@@ -0,0 +1,638 @@
...
+        self.registers
+            .ar
+            .write(Address::FAR.val((PAGE_START + page_number * PAGE_SIZE) as u32));
+        self.registers.cr.modify(Control::STRT::SET);
+
+        ReturnCode::SUCCESS
+    }
+
+    pub fn erase_all(&self) -> ReturnCode {
+        while self.registers.sr.is_set(Status::BSY) {}
```
hudson-ayers, line 501: does this loop still need to be here?

chips/stm32f303xc/src/flash.rs
hudson-ayers: Got it, thanks!

krady21:
Could you give me some pointers in regards to adding the virtual flash? Do i have to write an additional flash capsule that i would use along the virtual flash one similar to what is done with the [Alarm](https://github.com/tock/tock/blob/517ab5a4388c7c2d10e1a8a1ff3e342258515342/boards/stm32f3discovery/src/main.rs#L60) one?

chips/stm32f303xc/src/flash.rs
hudson-ayers: Yeah, that enum definitely needs some work, but making it generic probably doesn't work well with capsules being able to make decisions independent of the underlying driver. Instead we probably just need to add additional, more specific error variants to the enum.

chips/stm32f303xc/src/flash.rs
krady21: nope. Forgot about that one.

hudson-ayers:
> Could you give me some pointers in regards to adding the virtual flash? Do i have to write an additional flash capsule that I would use alongside the virtual flash capsule similar to what is done with the [Alarm](https://github.com/tock/tock/blob/517ab5a4388c7c2d10e1a8a1ff3e342258515342/boards/stm32f3discovery/src/main.rs#L60) one?

Sorry, I shouldn't have said virtual flash in my comment -- I just meant that there should be some capsule using the driver, instantiated in `boards/stm32f3discovery/src/main.rs`. Implementing the nonvolatile storage driver should be fine. There are examples of how to do so in https://github.com/tock/tock/blob/master/boards/imix/src/main.rs and https://github.com/tock/tock/blob/master/boards/nordic/nrf52840dk/src/main.rs

krady21:
> > Could you give me some pointers in regards to adding the virtual flash? Do i have to write an additional flash capsule that I would use alongside the virtual flash capsule similar to what is done with the [Alarm](https://github.com/tock/tock/blob/517ab5a4388c7c2d10e1a8a1ff3e342258515342/boards/stm32f3discovery/src/main.rs#L60) one?
> 
> Sorry, I shouldn't have said virtual flash in my comment -- I just meant that there should be some capsule using the driver, instantiated in `boards/stm32f3discovery/src/main.rs`. Implementing the nonvolatile storage driver should be fine. There are examples of how to do so in https://github.com/tock/tock/blob/master/boards/imix/src/main.rs and https://github.com/tock/tock/blob/master/boards/nordic/nrf52840dk/src/main.rs

Ok. I'm on it

krady21:
I have one question regarding the userspace addresses and length used by  `NonvolatileStorageComponent::new()`. I used the ones I found in chip_layout.ld. Can you confirm that they are the right ones? I am worried that i might overwrite the code.

boards/stm32f3discovery/src/main.rs
```diff
@@ -607,6 +609,18 @@ pub unsafe fn reset_handler() {
     );
     stm32f303xc::adc::ADC1.set_client(adc);
 
+    let nonvolatile_storage = components::nonvolatile_storage::NonvolatileStorageComponent::new(
+        board_kernel,
+        &stm32f303xc::flash::FLASH,
+        0x08020000,
+        0x20000,
+        0x08000000,
+        0x20000,
```
hudson-ayers, line 618: I think this should probably use the estorage / sstorage symbol approach the same way that Imix / nrf boards do

hudson-ayers:
> I have one question regarding the userspace addresses and length used by `NonvolatileStorageComponent::new()`. In the last commit, I used the ones that I found in chip_layout.ld. Can you confirm that they are the right ones? I am worried that i might overwrite the code segment.

I am actually not sure how the userspace addresses/length are supposed to be chosen. Neither Imix nor the nrf52840dk assign the entire prog region as userspace accessible though, which makes me nervous that this does... @bradjc ?

boards/stm32f3discovery/src/main.rs
krady21: Ok. I'll apply your suggestion.

krady21:
Ended up choosing 0x08038000 and 0x8000 (32KB) for the userspace address and length (prog being between 0x08020000 and 0x08040000). This leaves us with 96KB for the application binaries and 16 pages available for programming from userspace. Imix does a similar thing, using 0x60000, while its prog area is between 0x40000 and 0x80000. Let me know if you have a different split in mind.

hudson-ayers (dismissed):
I am happy with this PR, but would like if someone more familiar than me with how we assign userspace flash regions could double check that.

alistair23 (dismissed):

chips/stm32f303xc/src/flash.rs
```diff
@@ -0,0 +1,636 @@
...
+            return Err((ReturnCode::EINVAL, buffer));
+        }
+
+        let mut byte: *const u8 = (PAGE_START + page_number * PAGE_SIZE) as *const u8;
+        unsafe {
+            for i in 0..buffer.len() {
+                buffer[i] = *byte;
+                byte = byte.offset(1);
+            }
+        }
```
krady21, line 556: I have just realised that I am only doing normal pointer reads here.  The reference manual states the following "The embedded Flash module can be addressed directly, as a common memory space.", which makes me think that it is fine as it is. Do you think I should make them volatile?

alistair23 (approved):

chips/stm32f303xc/src/flash.rs
ppannuto: I think they only need to be volatile if there is a chance that the memory values will change from something _other_ than the executing code. That's necessary when doing DMA with peripherals, where a sensor/radio/etc might independently write to memory. As the flash controller won't change memory, I don't think volatile is required here.

boards/stm32f3discovery/src/main.rs
```diff
@@ -607,6 +609,26 @@ pub unsafe fn reset_handler() {
...
+        /// Beginning on the ROM region containing app images.
+        static _sstorage: u8;
+        static _estorage: u8;
+    }
+
+    let nonvolatile_storage = components::nonvolatile_storage::NonvolatileStorageComponent::new(
+        board_kernel,
+        &stm32f303xc::flash::FLASH,
+        0x08038000,
+        0x8000,
```
ppannuto, line 624: Given that these are somewhat arbitrary selections, it'd be nice to add a comment explaining how they were chosen (and that they're flexible).

ppannuto (commented):
I think this looks good -- one place where some comments would be valuable, but otherwise looks good to me

boards/stm32f3discovery/src/main.rs
krady21: I agree. The problem I see with these arbitrary selections (this is how they are done in other boards too) is that nothing stops tock from loading apps over the userspace start address. Correct me if I am wrong. I think this could be solved in the kernel linker script by adding start and end symbols for this userspace region.

chips/stm32f303xc/src/flash.rs
krady21: Makes sense.

boards/stm32f3discovery/src/main.rs
ppannuto: Agreed. This part of the design got hung up a bit a little while ago in a discussion about application identifiers (i.e. how to map storage space to applications) I think. Agreed that a dedicated section would probably be better in the short term, but I think in the long term we'd have a richer concept of 'application flash storage'.

ppannuto (approved):

bradjc (approved):

PR #2084, opened 2020-08-25
Board-based instantiation of chip drivers and interrupt mappings for nordic boards/chips

hudson-ayers:
### Pull Request Overview

This pull request applies the new peripheral instantiation approach proposed in #2069 to the nrf52840 and nrf52832 chips and associated boards. These chips were already hierarchically structured, so I applied that same approach to the macro used to create the default peripherals. Also, the nordic chips have some peripherals (such as USBD and POWER) which depend on each other, so I had to add an additional init() function to allow for finalizing these circular dependencies after instantiation of the drivers.

~~This PR sits on top of #2069, which is why it is so large.~~

### Testing Strategy

This pull request was tested by compiling, and by running the following apps on the nrf52840dk (several at a time):
- blink
- ieee802154/radio_tx
- buttons
- ble_advertising
- rng
- nonvolatile_storage
- c_hello

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


boards/nordic/nrf52840_dongle/src/io.rs
```diff
@@ -25,7 +25,10 @@ impl Write for Writer {
 
 impl IoWrite for Writer {
     fn write(&mut self, buf: &[u8]) {
-        let uart = unsafe { &mut nrf52840::uart::UARTE0 };
+        // Here, we create a second instance of the Uarte struct.
+        // This is okay because we only call this during a panic, and
+        // we will never actually process the interrupts
+        let uart = nrf52840::uart::Uarte::new();
```
gendx, line 31: Does this not require `unsafe`?

boards/nordic/nrf52840_dongle/src/main.rs
```diff
@@ -56,7 +56,8 @@ static mut PROCESSES: [Option<&'static dyn kernel::procs::ProcessType>; NUM_PROC
     [None; NUM_PROCS];
 
 // Static reference to chip for panic dumps
-static mut CHIP: Option<&'static nrf52840::chip::Chip> = None;
+nrf52840::create_default_nrf52840_peripherals!(Nrf52840Peripherals);
```
gendx, line 59: How can this be invoked outside of a function?

boards/nordic/nrf52840_dongle/src/main.rs
```diff
@@ -119,6 +120,15 @@ impl kernel::Platform for Platform {
 pub unsafe fn reset_handler() {
     // Loads relocations and clears BSS
     nrf52840::init();
+    let ppi = static_init!(nrf52840::ppi::Ppi, nrf52840::ppi::Ppi::new());
+    // Initialize chip peripheral drivers
+    let nrf52840_peripherals = static_init!(Nrf52840Peripherals, Nrf52840Peripherals::new(ppi));
+
+    // set up circular peripheral dependencies
+    {
```
gendx, line 128: I don't see why this pair of braces is needed.

chips/nrf52/src/acomp.rs
```diff
@@ -217,9 +217,9 @@ pub struct Comparator<'a> {
 }
 
 impl<'a> Comparator<'a> {
-    const fn new(registers: StaticRef<CompRegisters>) -> Self {
+    pub const fn new() -> Self {
```
gendx, line 220: How are such `new()` functions not marked as `unsafe`? One can instantiate them twice with safe code, which then aliases the registers, violating Rust's fundamental memory safety rules. On the contrary, the previous code only exposed `static mut` variables as public code, and `static mut` required `unsafe` to use.

I think we should come up with a better API.

chips/nrf52/src/chip.rs
```diff
@@ -1,20 +1,18 @@
...
-use crate::interrupt_service::InterruptService;
-use crate::nvmc;
 use core::fmt::Write;
 use cortexm4::{self, nvic};
 use kernel::common::deferred_call;
-use kernel::debug;
+use kernel::InterruptService;
 
-pub struct NRF52<I: InterruptService> {
+pub struct NRF52<I: InterruptService<DeferredCallTask> + 'static> {
```
gendx, line 7: Why is a reference now needed? And if we're introducing new references, can we use a generic `'a` lifetime parameter, in the spirit of #1074?

chips/nrf52/src/chip.rs
```diff
@@ -26,7 +24,142 @@ impl<I: InterruptService> NRF52<I> {
...
-impl<I: InterruptService> kernel::Chip for NRF52<I> {
+/// This macro defines a struct that, when initialized,
+/// instantiates all peripheral drivers shared across all supported nrf52 chips
+/// in Tock. If a board
+/// wishes to use only a subset of these peripherals, this
+/// macro cannot be used, and this struct should be
+/// constructed manually in main.rs. The input to the macro is the name of the struct
+/// that will hold the peripherals, which can be chosen by the board.
+#[macro_export]
+macro_rules! create_default_nrf52_peripherals {
```
gendx, line 35: I'm not sure that "create" is the clearest name for this macro. Without context, I had assumed that the contents of the macro were some code that initializes the peripherals, to put inside a function (similarly to `static_init`). I didn't assume the macro's contents were a struct definition.

Can we come up with a better naming?

chips/nrf52/src/chip.rs
```diff
@@ -26,7 +24,142 @@ impl<I: InterruptService> NRF52<I> {
...
+    ($N:ident) => {
```
gendx, line 36: Why is this name a parameter? I only see the macro invoked with "Nrf52BasePeripherals".

gendx (changes requested):
I had a look only at the Nordic-specific parts. I have two main concerns with this design.

- The new peripherals APIs seem to violate Rust's memory safety rules. Whereas an `unsafe` block was required before to access them (due to `static mut` variables being `unsafe` to access), there are now public `new()` functions exposed that are not marked as `unsafe`. I think they should be `unsafe`, as one must only be allowed to create a single peripheral in the program to avoid aliasing registers.
- Besides an unclear naming, I don't see what the `create_default_foo_peripherals` macros are trying to achieve. The `$N` parameter is not useful (for Nordic at least), as each macro is always used with the same name. Can't the corresponding structures be defined once in the corresponding chip (without any macro), and then have the boards simply `use` the structures to import them in the current namespace, if that's the goal? I think macros should generally be avoided as they make code less clear (similarly to Cargo features), except for really useful cases like `register_structs` or `static_init`.

---

Besides that, would it make sense to define a `Peripheral` trait, that would declare the `new` function? That is:

```rust
trait Peripheral {
    const unsafe fn new() -> Self;
}
```

Or are traits too restrictive regarding `const` and/or `unsafe`?

chips/nrf52/src/chip.rs
hudson-ayers: It doesn't really have to be a parameter. I made it a parameter so that if someone wanted to switch between the macro version and a version with fewer drivers instantiated in boards, it would be a little easier. But I am fine with a version that does not pass a name.

chips/nrf52/src/chip.rs
hudson-ayers: How about `define_default_nrf52_peripherals`?

chips/nrf52/src/acomp.rs
hudson-ayers: I spoke with @alevy about this, and I do not believe it violates Rust's memory safety rules. `StaticRef` provides access to the underlying registers by wrapping the underlying registers in an `UnsafeCell`, but it is not memory unsafe for multiple references to a `StaticRef` to exist. Notably, this design does expose a footgun in that it is possible to instantiate a peripheral twice, when only one can be hooked up to receive interrupts. But that is actually occasionally useful (for the UART in the panic handler, for example). We could have each new function require a capability to lessen the impact of this footgun, but I am not sure if that is necessary. 

Note that within a given module it was already possible to alias the registers without unsafe by simply using the `const BASE_REGISTERS` multiple times, which did not require unsafe. In fact, there are already instances of code that does so in Tock drivers today.

That being said, I would be happy to hear suggestions for an alternate API.

boards/nordic/nrf52840_dongle/src/main.rs
hudson-ayers: thanks, fixed

boards/nordic/nrf52840_dongle/src/main.rs
hudson-ayers: It only defines a struct and methods on that struct, it doesn't do anything at runtime.

boards/nordic/nrf52840_dongle/src/io.rs
hudson-ayers: currently no. We could consider making it take a 'PeripheralCreationCapability'.

chips/nrf52/src/chip.rs
hudson-ayers: A reference is now needed because the struct that implements the interrupt service now also holds all the peripherals, and there are instances where the peripherals need to be instantiated before the Chip struct itself (maybe not for the nordic boards, but I think its good to be consistent across all chips). 

I changed it to use a generic lifetime.

hudson-ayers:
> * Besides an unclear naming, I don't see what the `create_default_foo_peripherals` macros are trying to achieve. The `$N` parameter is not useful (for Nordic at least), as each macro is always used with the same name. Can't the corresponding structures be defined once in the corresponding chip (without any macro), and then have the boards simply `use` the structures to import them in the current namespace, if that's the goal? I think macros should generally be avoided as they make code less clear (similarly to Cargo features), except for really useful cases like `register_structs` or `static_init`.

Agreed that the naming parameter is not that useful, I can get rid of that. I also agree that it would be possible to take the approach you describe that does not require a macro. However, the main motivation of this change is that it allows for out-of-tree boards to selectively exclude peripherals they do not need. To do so, we need to ensure that all chip peripherals and interrupt mappings can be instantiated from within `main.rs`. By using a macro, we enforce that even in upstream boards, these things are executed in `main.rs`, so it is easy to see that out-of-tree boards could exclude peripherals and do the same. If we moved to an approach where upstream boards just use the structures defined in the chip, it would be easy to end up with peripherals that do not expose constructors outside the chip crate, but would still work for the in tree boards. Also, I think the macro approach makes it clear precisely what code should be copied into `main.rs` to create a more limited out-of-tree board.

> Besides that, would it make sense to define a `Peripheral` trait, that would declare the `new` function? That is:
> 
> ```rust
> trait Peripheral {
>     const unsafe fn new() -> Self;
> }
> ```
> 
> Or are traits too restrictive regarding `const` and/or `unsafe`?

Traits are currently too restrictive regarding `const`. We could make `new()` not const with this new approach that does not rely on globals, however, which would make this possible.

boards/nordic/nrf52840_dongle/src/main.rs
gendx: Ok, see my other comment regarding the naming. Let's discuss that there.

chips/nrf52/src/acomp.rs
gendx: > `StaticRef` provides access to the underlying registers by wrapping the underlying registers in an `UnsafeCell`.

I don't see any `UnsafeCell` https://github.com/tock/tock/blob/master/kernel/src/common/static_ref.rs. But after giving it a second look, the `StaticRef` API indeed looks fine to me.
- There is an `unsafe` function to create a `StaticRef`, the prerequisite being a non-null pointer with `'static` lifetime to a `T` object that is otherwise not aliased in the program. I don't see these prerequisites violated in the peripherals we instanciate.
- Cloning is safe. No issue with that, cloning a pointer is in itself safe.
- Dereferencing to a `&'static T`. This is safe because the pointer passed to `new` has `'static` lifetime, and indeed points to a `T` (in particular, is aligned correctly). It's also fine to dereference clones of a `StaticRef`, as we obtain *immutable* references.

The `UnsafeCell` part is from https://github.com/tock/tock/blob/master/libraries/tock-register-interface/src/registers.rs. This also looks fine to me, w.r.t. the `UnsafeCell` rules (https://doc.rust-lang.org/std/cell/struct.UnsafeCell.html).

- The register API doesn't expose any `&T` or `&mut T` from the `UnsafeCell` in register. Simple `set`/`get` API.
- No multi-threading in Tock.
- I wonder however how that plays with re-entrancy. In normal operation the Tock scheduler doesn't call code re-entrantly (just queues pending interrupts for the main loop to process).
- The panic handler path should be fine, but is not 100% clear to me. What if a panic happens from within the UART code, and that we re-entrantly call UART code again?

> Note that within a given module it was already possible to alias the registers without unsafe by simply using the `const BASE_REGISTERS` multiple times, which did not require unsafe. In fact, there are already instances of code that does so in Tock drivers today.

Good point.

So I think I don't have much concerns left here. In short, it's possible to instantiate a set of registers twice, and have the two instances make unrelated writes to the underlying registers (but not *concurrently* in the multi-threaded sense, as Tock is single-threaded), which would certainly violate the logic of the underlying peripheral, but wouldn't trigger any UB at the Rust language level (so there is no risk that compiler optimizes away or things like that).
I think it's fine, and best to keep `unsafe` outside of that (I'm all for reducing the amount of `unsafe` code to the minimum).

But I still want to note that depending on what the peripheral does of it, logic bugs where a peripheral's registers are put in an inconsistent state (due to multiple instantiations of the same registers) may still trigger very bad stuff (though I don't have a specific example in mind). So it would be nice to have a way to prevent instantiating registers twice (except for those required in panic handlers), to avoid logic bugs, but that's out of this pull request's scope.

gendx:
> To do so, we need to ensure that all chip peripherals and interrupt mappings can be instantiated from within `main.rs`. By using a macro, we enforce that even in upstream boards, these things are executed in `main.rs`, so it is easy to see that out-of-tree boards could exclude peripherals and do the same.

I don't really see the point here. Nothing is *instantiated* within `main.rs`, but simply *defined* (as we discuss in the other comment). I don't see how defining things in the `main` module provides anything better than defining them in a chip module, and `use` them in main.

If a custom board wants to exclude peripherals, they can define their own functions instead of using the chip's defaults. So I don't see what the macro provides that can't be done with simple definitions and `use`.

Do you have an example of an out-of-tree board and how it would leverage the macro?

chips/nrf52/src/acomp.rs
hudson-ayers: Thanks for looking into this some more!  Agreed about the potential for logic bugs, but I don't think its that much worse than before where there was widespread use of the globals that allowed for the same bugs, just requiring an `unsafe`.

hudson-ayers:
> I don't really see the point here. Nothing is _instantiated_ within `main.rs`, but simply _defined_ (as we discuss in the other comment). I don't see how defining things in the `main` module provides anything better than defining them in a chip module, and `use` them in main.
> 
> If a custom board wants to exclude peripherals, they can define their own functions instead of using the chip's defaults. So I don't see what the macro provides that can't be done with simple definitions and `use`.
> 
> Do you have an example of an out-of-tree board and how it would leverage the macro?

An out of tree board would not leverage the macro, it would simply copy past the code that currently resides within the macro into its own `main.rs`, and remove the peripherals it does not want. 

Using a macro makes it so upstream boards define the peripherals struct in main as well, ensuring that the visibility of all structs and functions necessary to define peripherals and interrupt mappings can be accessed from a dependent crate (the board crate). If we moved to a non-macro definition, that was just imported into main, I worry about a scenario like this:

```rust
//! chips/apollo3/src/chip.rs

struct DefaultApollo3Peripherals {
    stimer: crate::stimer::STimer<'static>,
    // ...
}
impl DefaultApollo3Peripherals {
    unsafe fn new() -> Self {
        Self {
            stimer: crate::stimer::STimer::new(),
            // ...
        }
    }
}
impl kernel::InterruptService<()> for DefaultApollo3Peripherals {
    unsafe fn service_interrupt(&self, interrupt: u32) -> bool {
        use crate::nvic;
        match interrupt {
            nvic::STIMER..=nvic::STIMER_CMPR7 => self.stimer.handle_interrupt(),
            // ...
```

```rust
//! chips/apollo3/src/stimer.rs

pub struct STimer<'a> {
    registers: StaticRef<STimerRegisters>,
    client: OptionalCell<&'a dyn hil::time::AlarmClient>,
}

impl<'a> STimer<'_> {
    pub(crate) const fn new() -> STimer<'a> { // <-- Works for boards that use the struct in chips/, but cant call outside chips/
        STimer {
            registers: STIMER_BASE,
            client: OptionalCell::empty(),
        }
    }
}
```

Also, in this scenario it is somewhat less straightforward to paste the struct definition into main and edit it, because the board author has to modify all the paths to be from the board crate instead of from within the chip crate.

Does that make sense? I agree that macros are non-ideal, but I think if we use the non-macro approach here it will inevitably be the case that visibility will not be sufficient for out of tree board authors to only define an out of tree board crate, forcing them to maintain an out of tree chip crate as well.

chips/nrf52/src/chip.rs
```diff
@@ -52,12 +185,14 @@ impl<I: InterruptService> kernel::Chip for NRF52<I> {
         unsafe {
             loop {
                 if let Some(task) = deferred_call::DeferredCall::next_pending() {
-                    match task {
-                        DeferredCallTask::Nvmc => nvmc::NVMC.handle_interrupt(),
+                    match self.interrupt_service.service_deferred_call(task) {
+                        true => {}
+                        false => panic!("unhandled deferred call task"),
                     }
```
gendx, line 191: Matching over a boolean to only handle the `false` case seems overkill. Why not using `if !self.interrupt_service.service_deferred_call(task) { panic!(...); }`?

gendx:
So as I understand it, the macro would allow to circumvent the visibility rules to make it easier to support out-of-tree boards/chips. I think this raises multiple questions.

- Regarding this specific problem: if peripherals have to be available outside of the chip crate, why not simply make them public? It would be much simpler to define appropriate visibility rules than trying to circumvent them with macros. In other words, `pub const fn new() -> STimer<'a>`.
- More broadly, should Tock aim to support out-of-tree boards/chips? I'd argue that if they are out-of-tree, they are unsupported by default, and Tock shouldn't go to great lengths to support them. It doesn't mean that Tock should make it hard to integrate out-of-tree boards/chips, but I don't think out-of-tree boards/chips should steer Tock towards complex designs that are harder to maintain (e.g. features, macros, etc.).
- I don't see a big issue in out-of-tree board maintainers having to also maintain out-of-tree chips, if the chip in question differs from the default in Tock. Also, how much maintenance is that really? If it's a simple `git patch` to apply on the chip crate I don't think it's that much of a burden for unsupported boards.
- If maintaining out-of-tree boards is too difficult, shouldn't Tock have more boards/chips in-tree? In the past year, we've already added quite a few more boards/chips in Tock, so I don't think adding more of them is a problem per se. Actually, having more supported boards allows to make sure that core changes in the kernel are suited for a wide range of use cases (rather than forcing maintainers of out-of-tree boards to figure it out).
- Of course, I don't think we should remove the possibility of defining out-of-tree boards, but these will necessarily have a maintenance burden for whoever manages them.

chips/nrf52/src/acomp.rs
alistair23: There are some good use cases where it is useful to have multiple instances of a set of registers.

For example with a pin controller. Let's say there is a pin controller that can be used to allow either I2C or SPI. You could pass the pin controller registers to both the I2C and the SPI devices and then have the I2C/SPI device configure the pins before it starts. Obviously you would then need to make sure only 1 is operating at a time (maybe with a virtual_bus capsule).

The same sort of thing would apply to power management, where each device can enable and disable itself as it is started/stopped. For the Apollo3 for example it would be great to only enable devices when accessed, instead of just enabling everything at boot. Stopping devices is another problem that is much more complex.

hudson-ayers:
> So as I understand it, the macro would allow to circumvent the visibility rules to make it easier to support out-of-tree boards/chips. I think this raises multiple questions.
>
> * Regarding this specific problem: if peripherals have to be available outside of the chip crate, why not simply make them public? It would be much simpler to define appropriate visibility rules than trying to circumvent them with macros. In other words, `pub const fn new() -> STimer<'a>`.

The goal of using a macro is to ensure that the peripherals *are* public. This PR makes all constructors public, but if they are not instantiated via a macro this will not be enforced in CI, and I suspect that future PRs will add peripherals with non-public constructors (or with constructors that rely on access to other non public fields), because doing so will not cause compilation failures for boards that use the default (which all upstream boards will).

> * More broadly, should Tock aim to support out-of-tree boards/chips? I'd argue that if they are out-of-tree, they are unsupported by default, and Tock shouldn't go to great lengths to support them. It doesn't mean that Tock should make it hard to integrate out-of-tree boards/chips, but I don't think out-of-tree boards/chips should steer Tock towards complex designs that are harder to maintain (e.g. features, macros, etc.).

I think this is a reasonable point, and we could argue that it is up to out-of-tree to submit patches if peripherals are mistakenly made not public. This would allow us to use this approach, without a macro.

> * I don't see a big issue in out-of-tree board maintainers having to also maintain out-of-tree chips, if the chip in question differs from the default in Tock. Also, how much maintenance is that really? If it's a simple `git patch` to apply on the chip crate I don't think it's that much of a burden for unsupported boards.

One point of concern is that rather than using patches, many out of tree boards (such as opentitan and the h1) currently maintain their own version of the chip crate entirely. This makes it harder for out of tree boards to stay up to date with the upstream chip crate, and makes out of tree boards less likely to submit bug fixes upstream, as the bug fixes are not widely available. Even OpenSK currently has outstanding bug fixes for the nrf52 USB driver that have not been submitted to upstream. Personally, I do think their is value to Tock in reducing the friction required to maintain boards out of tree.

> * If maintaining out-of-tree boards is too difficult, shouldn't Tock have more boards/chips in-tree? In the past year, we've already added quite a few more boards/chips in Tock, so I don't think adding more of them is a problem per se. Actually, having more supported boards allows to make sure that core changes in the kernel are suited for a wide range of use cases (rather than forcing maintainers of out-of-tree boards to figure it out).

Unfortunately there is a substantial burden to Tock developers to maintain more boards/chips in-tree, which is part of why we also deprecate little-used board/chip combos regularly. Regardless, I think a change to what boards Tock maintains upstream is a bit out of scope of this PR.

chips/nrf52/src/acomp.rs
hudson-ayers: This can be done by making the `const XXX: StaticRef<Registers>` --> `pub const XXX: StaticRef<Registers>` instead. But I don't think we want to take that approach universally, as it will often only make sense when there is some virtualization ensuring concurrent accesses do not lead to unexpected behavior.

For power management the state-of-the-art in Tock previously was anything that wanted to manage power accessing some global `PowerManager` struct. This change at least requires all devices which want to use the struct to take a reference to the power manager when they are instantiated, so it is clearer which peripherals will interact with it.

alistair23:
+1 for not using macros. It doesn't seem unreasonable to just have out of tree board maintainers submit fixes to mainline Tock to fix any non public peripherals. Hindering everyone else for a few out of tree users doesn't seem like the right approach. The out of tree users should just upstream their work.

hudson-ayers:
Okay, I think I am convinced. I will try to update both PRs to use a non-macro approach, and add some docs updates to indicate that chips should keep peripherals publicly exposed to support out-of-tree boards.

hudson-ayers:
I got rid of the large macros in favor of structs (and did so on the apollo3/sam4l version of this PR as well).

chips/nrf52/src/chip.rs
```diff
@@ -26,7 +24,133 @@ impl<I: InterruptService> NRF52<I> {
...
 
-impl<I: InterruptService> kernel::Chip for NRF52<I> {
+/// This struct, when initialized, instantiates all peripheral drivers for the apollo3.
+/// If a board wishes to use only a subset of these peripherals, this
+/// should not be used or imported, and a modified version should be
+/// constructed manually in main.rs.
+pub struct Nrf52DefaultPeripherals<'a> {
+    pub acomp: crate::acomp::Comparator<'a>,
+    pub ecb: crate::aes::AesECB<'a>,
+    pub gpio_port: &'static crate::gpio::Port<'static>,
```
gendx, line 34: Does this have to be `'static` too?

chips/nrf52/src/chip.rs
gendx: Still unresolved.

chips/nrf52832/src/interrupt_service.rs
```diff
@@ -1,21 +1,30 @@
...
 }
-
-impl Nrf52832InterruptService {
-    pub unsafe fn new() -> Nrf52832InterruptService {
-        Nrf52832InterruptService {
-            nrf52: nrf52::interrupt_service::Nrf52InterruptService::new(&gpio::PORT),
+impl<'a> Nrf52832DefaultPeripherals<'a> {
+    pub fn new(ppi: &'a crate::ppi::Ppi) -> Self {
+        Self {
+            nrf52_base: unsafe { Nrf52DefaultPeripherals::new(&crate::gpio::PORT, ppi) },
```
gendx, line 15: Does hard-coding `crate::gpio::PORT` here fits in the new model introduced in this pull request?

chips/nrf52832/src/interrupt_service.rs
```diff
@@ -1,21 +1,30 @@
...
+use nrf52::chip::Nrf52DefaultPeripherals;
 
-pub struct Nrf52832InterruptService {
-    nrf52: nrf52::interrupt_service::Nrf52InterruptService<'static>,
+/// This struct, when initialized, instantiates all peripheral drivers for the nrf52840.
+/// If a board wishes to use only a subset of these peripherals, this
+/// should not be used or imported, and a modified version should be
+/// constructed manually in main.rs.
+pub struct Nrf52832DefaultPeripherals<'a> {
+    pub nrf52_base: Nrf52DefaultPeripherals<'a>,
```
gendx, line 9: Not sure we should name it `nrf52_base` as opposed to simply `nrf52` (which follows the name of the `nrf52` crate).

chips/nrf52840/src/interrupt_service.rs
```diff
@@ -1,25 +1,39 @@
...
+    pub fn new(ppi: &'a crate::ppi::Ppi) -> Self {
+        Self {
+            nrf52_base: unsafe { Nrf52DefaultPeripherals::new(&crate::gpio::PORT, ppi) },
+            usbd: crate::usbd::Usbd::new(),
         }
     }
+    // Necessary for setting up circular dependencies
+    pub fn init(&'a self) {
+        self.nrf52_base.pwr_clk.set_usb_client(&self.usbd);
+        self.usbd.set_power_ref(&self.nrf52_base.pwr_clk);
```
gendx, line 24: This is fine, but I wonder whether these two calls cannot be bundled in one?

chips/nrf52840/src/interrupt_service.rs
```diff
@@ -1,25 +1,39 @@
...
 
-pub struct Nrf52840InterruptService {
-    nrf52: nrf52::interrupt_service::Nrf52InterruptService<'static>,
+/// This struct, when initialized, instantiates all peripheral drivers for the nrf52840.
+/// If a board wishes to use only a subset of these peripherals, this
+/// should not be used or imported, and a modified version should be
+/// constructed manually in main.rs.
+//create all base nrf52 peripherals
+pub struct Nrf52840DefaultPeripherals<'a> {
+    pub nrf52_base: Nrf52DefaultPeripherals<'a>,
```
gendx, line 10: Same question about `nrf52_base`.

kernel/src/platform/mod.rs
```diff
@@ -139,6 +139,56 @@ pub trait Chip {
...
+/// ```ignore
+///           +---->nrf52840
+///           |        |
+///           |        |
+///           |        v
+///           |      nrf52
+///           |        |
+///           |        |
+///           |        v
+/// kernel-->nrf5     nrf5
```
gendx, line 178: Is it still up-to-date now that it's migrated to the kernel? Same question for the text above.

kernel/src/platform/mod.rs
```diff
@@ -139,6 +139,56 @@ pub trait Chip {
...
+/// ```
+/// where the kernel instructs the `nrf5` crate to handle interrupts, and if
+/// there is an interrupt ready then that interrupt is passed through the crates
+/// until something can service it.
+pub trait InterruptService<T> {
```
gendx, line 183: Shouldn't `T` be bounded by some kind of "task" trait?

gendx (commented):
There are still a few pending comments, but apart from that it generally looks fine (for Nordic boards/chips at least). One question is whether this needs to be reviewed & submitted as a single pull request or split by board/chip groups (I don't mind either way).

kernel/src/platform/mod.rs
hudson-ayers: That would make sense, but we don't have any `DeferredCallTask` trait today.

kernel/src/platform/mod.rs
hudson-ayers: updated.

chips/nrf52840/src/interrupt_service.rs
hudson-ayers: done

chips/nrf52832/src/interrupt_service.rs
hudson-ayers: done

chips/nrf52840/src/interrupt_service.rs
hudson-ayers: It could be done with a single (non-method) function, but would require changing the visibility of fields of one of the structs. I think this is more in line with how we set up circular dependencies elsewhere anyways.

chips/nrf52832/src/interrupt_service.rs
hudson-ayers: It does not really fit the model. Unfortunately, keeping a single `Port` type despite differing number of pins between boards is not very ergonomic without `min_const_generics`, so I was planning to wait on that to stabilize rather than transition to something where we continue using slices but stores the pins separately from the `Port`.

chips/nrf52/src/chip.rs
hudson-ayers: see my other comment on why `gpio_port` did not move to the new model

kernel/src/platform/mod.rs
gendx: Shouldn't we introduce one with this pull-request then? 
- If we always use `T = DeferredCallTask` for now, I don't see the need for it to be generic.
- If we use various task types, I think we should unify them with a trait.

chips/nrf52832/src/interrupt_service.rs
gendx: Makes sense. We could mention that in a comment in the code, so that later we think about it when `min_const_generics` gets more stable.

chips/nrf52/src/chip.rs
gendx: Why was this comment marked as resolved? I still don't see the code that fixes that, nor any alternative explanation.

chips/nrf52832/src/interrupt_service.rs
```diff
@@ -1,21 +1,30 @@
...
 }
-
-impl Nrf52832InterruptService {
-    pub unsafe fn new() -> Nrf52832InterruptService {
-        Nrf52832InterruptService {
-            nrf52: nrf52::interrupt_service::Nrf52InterruptService::new(&gpio::PORT),
+impl<'a> Nrf52832DefaultPeripherals<'a> {
+    pub fn new(ppi: &'a crate::ppi::Ppi) -> Self {
+        Self {
+            nrf52: unsafe { Nrf52DefaultPeripherals::new(&crate::gpio::PORT, ppi) },
```
gendx, line 15: So, why is `Nrf52DefaultPeripherals::new` unsafe but not `Nrf52832DefaultPeripherals::new`? This `unsafe` block deserves an explanation why it's safe.

chips/nrf52840/src/interrupt_service.rs
```diff
@@ -1,25 +1,39 @@
...
 }
 
-impl Nrf52840InterruptService {
-    pub unsafe fn new() -> Nrf52840InterruptService {
-        Nrf52840InterruptService {
-            nrf52: nrf52::interrupt_service::Nrf52InterruptService::new(&gpio::PORT),
+impl<'a> Nrf52840DefaultPeripherals<'a> {
+    pub fn new(ppi: &'a crate::ppi::Ppi) -> Self {
+        Self {
+            nrf52: unsafe { Nrf52DefaultPeripherals::new(&crate::gpio::PORT, ppi) },
```
gendx, line 17: Same here.

kernel/src/platform/mod.rs
hudson-ayers: Each chip defines its own `Task` type (sam4l calls it `Task`, nrf52 calls it `DeferredCallTask`), so it has to be a generic or an associated type for it to be part of the `InterruptService` trait definition. Perhaps an associated type would be better?

Currently, each `Task` type is just an enum that specifies different task types and can be converted to/from usize. Given that there are no common functions implemented across all `Task` types I am not sure how much value is gained from making a trait with no functions. 

Would adding the last line of the following code snippet really help that much with unifying?

```rust
/// A type of task to defer a call for
#[derive(Copy, Clone)]
pub enum DeferredCallTask {
    Nvmc = 0,
}

impl DeferredTaskTrait for DeferredCallTask {}
```

chips/nrf52832/src/interrupt_service.rs
hudson-ayers: Hmm..`Nrf52DefaultPeripherals::new()` is unsafe because it instantiates the MPU, syscalls, and systicks, all functions which are marked unsafe. I should probably just bubble that unsafe all the way to main for now, though eventually it might be better for us to rethink whether those constructors are actually unsafe in the rust sense.

Changed `Nrf52832DefaultPeripherals::new()` and `Nrf52840DefaultPeripherals::new()` to be marked unsafe.

chips/nrf52840/src/interrupt_service.rs
hudson-ayers: responded on other thread.

chips/nrf52/src/chip.rs
hudson-ayers: oh, sorry, I fixed the instance below where I did the same for the `service_interrupt` call and did not realize I had made the error twice. Actually fixed now.

chips/nrf52832/src/interrupt_service.rs
hudson-ayers: I added a comment in both places indicating that this is waiting on `min_const_generics`

hudson-ayers:
Rebased on top of the changes in #2069

kernel/src/platform/mod.rs
gendx: I think such a marker trait would be useful, in restricting what kind of type can be applied there. Besides, when someone implements a new chip, they can look where `DeferredTaskTrait` is already implemented and more quickly understand what to do.

chips/nrf52832/src/interrupt_service.rs
gendx: Whether the underlying function should be `unsafe` is a question we can indeed ask. But I don't see any reason to mark the relatively simple `Nrf52832DefaultPeripherals::new` wrapper function as safe (unless the unsafety comes fom the GPIO port, and that using `&crate::gpio::PORT` suddenly makes it safe).

> I should probably just bubble that unsafe all the way to main for now.

I think that's how `unsafe` should work, unless there is a clear reason why some intermediate function is always safe to call, by guaranteeing that the safety invariants are always respected. As far as I understand it, these invariants are currently only guaranteed in the main function.

bradjc (approved):

hudson-ayers:
Rebased on master. One change I had to make in capsules was to make the `controller()` function on the `CdcAcm` struct be `pub`, so that the nano33 could still easily access it from the panic handler without requiring two globals.

If anyone thinks that is problematic I could go with the 2 globals approach instead, but I think the way I chose is cleaner.

bradjc (approved):

PR #2085, opened 2020-08-26
Allow creating a Spi Peripheral

alistair23:
### Pull Request Overview

In the context of this pull request the SPI Slave is refereed to as SPI Peripheral and the SPI Master is the SPI Controller. See https://hackaday.com/2020/06/29/updating-the-language-of-spi-pin-labels-to-remove-casual-references-to-slavery/ for details on why I picked these names.

Previously the SPI peripheral was tightly coupled to the SPI controller capsule in Tock. I'm assuming the idea was that a SPI device would support both controller and peripheral functionality. In practice most hardware seems to have a separate module for controller and peripheral (see Apollo3 documentation).

As well as this SPI generally only supports a single controller, so it's unlikely that a single hardware device will swap between controller and peripheral modes. It also doesn't appear like Tock supports that. In theory a virtual SPI capsule could allow one app to send data on the bus as a controller while another app receives data as a peripheral, but that doesn't seem like a workable hardware design.

In light on this let's split out the SPI peripheral to be independent. This makes it possible for OpenTitan (which only supports peripheral mode) and Apollo3 (which has separate controller and peripheral devices) to use the SPI peripheral.

This will allow boards to attach a SPI peripheral and allow apps to access the SPI peripheral.

While touching the code let's also rename SPISlave to SPIPeripheral. I haven't renamed everything, just bits that I'm touching.

Otherwise no changes have been made to the current implementation. That is left for future patches.

### Testing Strategy

I have some OpenTitan patches on top of this that create a SPI peripheral device that I have used for testing. I can now get userspace apps to call the SPI peripheral device functions (which aren't implemented yet).

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


hudson-ayers:
I think that this split makes sense. It seems notable that despite the SAM4L having an implementation for `SpiSlave` none of our SAM4L boards actually instantiate a `VirtualSpiSlaveDevice` or a `SpiSlaveClient`, presumably because the chip driver does not support dynamically switching between peripheral / controller modes. 

Also, it seems that tock-on-titan had to make changes compared to our capsule to support spi peripheral functionality (https://github.com/google/tock-on-titan/pull/162/files).

I support changing the names, though I am not sure if that should be part of this PR or would be better off as an independent PR. I also think that we should consider that the OpenTitan project uses the terms `Device`/`Host`.

alistair23:
Until specs catch up we are going to have some naming pain. SPI is a little different as there isn't a spec. CIPO/COPI seems to be the way people are going (or the original MISO/MOSI) which is why I picked it.

osenft:
+1 on splitting SPI device / peripheral and SPI host / controller. I found that having both in the same file makes working with the capsule, HIL and driver code a bit awkward, since you easily end up looking at the "wrong" code.

hudson-ayers:
We talked about naming on the call today, and there was consensus that Controller/Peripheral is a reasonable approach to take for now, and that we can always change again in the future if industry converges on some other terminology.

alistair23:
I pushed an update renaming the `spi.rs` file to `spi_controller.rs` and updated the README.

bradjc (approved):

hudson-ayers (approved):

PR #2087, opened 2020-08-27
opentitan: README: Update the OpenTitan minimum version SHA

alistair23:
### Pull Request Overview

Update the OpenTitan minimum SHA to the latest from master. Apparently the old SHA didn't work so update the minimum version to something newer. I tested today and I can run a small C app with the latest Tock master.

Also improve the documentation to indicate that users should try to use the latest bitstream.

### Testing Strategy

Booting an app on OpenTitan.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/README.md
+++ b/boards/opentitan/README.md
@@ -12,7 +12,7 @@ You can get started with OpenTitan using either the Nexys Video FPGA board or si
 Programming
 -----------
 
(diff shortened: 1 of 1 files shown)


alistair23:
@phil-levis  This works for me

Issue #1691, opened 2020-03-16
delay_ms (1) blocks the process (and maybe the whole alarm system) on stm boards


alexandruradovici:
Using delay_ms with 0 or 1 will block the running process (and I think it might lock any other alarm, as this will be considered the first one) on stm boards (and maybe others). 

I tried to do some debugging, it seems that yield never returns. The alarm driver sets the timeout to a tick that is before the cnt counter, so the interrupt never fires (might fire after overflow).

A quick fix is in capsules/src/alarm.rs, adding a few extra ticks if data is less then now(), but this is not really generic. Maybe the underlying alarm should fire immediately if the tick is below (and very close to) the current counter?

````rust
let time = if data > self.alarm.now() as usize { data } else { (self.alarm.now () + 10) as usize };
````

Application code is
````c
int main(void) {
  putnstr_async(hello, sizeof(hello), nop, NULL);
  int i = 0;
  while (1)
  {
    i = i + 1;
    snprintf (hello, sizeof (hello)-1, "%d\r\n", i);
    putnstr_async(hello, strlen(hello), nop, NULL);
    if ((i / 100)%2 == 0)
    {
      led_off(0);
    }
    else
    {
      led_on(0);
    }
    delay_ms (1);
    // execution does not arrive here
  }
  return 0;
}
````

alevy:
> Maybe the underlying alarm should fire immediately if the tick is below (and very close to) the current counter?

Indeed, this is what the SAM4L does (while also allowing for time the chip requires to sync with asynchronous timer):

https://github.com/tock/tock/blob/master/chips/sam4l/src/ast.rs#L321

This is the right strategy

alevy:
Nice catch!

alexandruradovici:
I am getting some very interesting behavior. I modified the stm32f4xx code with the following:

(ALARM0_SYNC_TICS is set to 8, tried several values with same result)

````rust
let now = self.registers.cnt.get ();
if tics.wrapping_sub(now <= ALARM0_SYNC_TICS {
   tics = now.wrapping_add(ALARM0_SYNC_TICS);
}
````
When tics is smaller than now, wrapping_sub will return 4294967294. I assume this is the expected result (it is the same in the rust playground). Obviously the if does not go through and tics is set to a value lower than now.

I tried the second approach:

````rust
let now = self.registers.cnt.get();
let itics: i64 = i64::from(tics);
let inow: i64 = i64::from(now);
debug!("{}", inow - itics);
[...]

Issue #1651, opened 2020-03-03
[Bug] Incorrect VirtualMuxAlarm implementation


gendx:
## Context

When setting an alarm with `VirtualMuxAlarm`, a reference point (`self.mux.prev`) is set along with the alarm.

https://github.com/tock/tock/blob/f30961f2772a0ef5f41e6f1a04843adf51151a3c/capsules/src/virtual_alarm.rs#L84-L95

This `prev` reference point is then used to decide whether alarms in the mux have expired or not.

https://github.com/tock/tock/blob/f30961f2772a0ef5f41e6f1a04843adf51151a3c/capsules/src/virtual_alarm.rs#L149

## Problem

The problem is that inside `VirtualMuxAlarm::set_alarm` there is no guarantee that `cur_alarm >= now`. We can be in a situation where an alarm 1 is already set, and by the time we set another alarm 2, alarm 1 is already in the past (but it hasn't fired yet).

```
> in VirtualMuxAlarm::set_alarm
alarms:       | alarm 1   |      | alarm 2
before:  prev | cur_alarm | now  | when
after:        |           | prev | cur_alarm
```

In that case, because the new `prev` is just after alarm 1, the alarm 1 won't be considered as expired next time `MuxAlarm::fired` is called - it would instead take time for the ticks to wrap-around for the alarm to be considered expired.

This is in particular observable with the Segger RTT debugging on Nordic, which sets a timer very close in the future (100us). By the time other code has run, this can already be in the past.

https://github.com/tock/tock/blob/f30961f2772a0ef5f41e6f1a04843adf51151a3c/capsules/src/segger_rtt.rs#L243

But **all capsules using a virtual alarm are affected**.

The observable result is that the alarm client (in this case Segger RTT) waits forever (for Segger RTT, kernel debugging & console seem to freeze).

## Reproducing the bug

It's easy to reproduce by adding the following check to `VirtualAlarmMux::set_alarm` on an nRF52840-DK board with `USB_DEBUGGING` and `trace_syscalls` enabled. Note that I've already applied the fix from https://github.com/tock/tock/pull/1636.

```rust
if enabled > 0 {
    let cur_alarm = self.mux.alarm.get_alarm();
    let now = self.now();
    let prev = self.mux.prev.get();

    if now.wrapping_sub(prev) > cur_alarm.wrapping_sub(prev) {
        panic!("prev = {}, cur_alarm = {}, now = {}");
    }
    ...
}
```

[...]

PR #1521, opened 2020-01-06
[RFC] Generalize ticks in the time HIL, and add support for 24- and 32-bit ticks

gendx:
### Pull Request Overview

This pull request modifies the time HIL to support various width of clock ticks, as an alternative proposal to https://github.com/tock/tock/pull/1413. In particular, support for 24-bit and 32-bit ticks are added, but it could easily be extended to other widths in the future as needed.

- Added a `Ticks` associated type for the `Time` trait, in lieu of the currently unused `W` parameter. This is more consistent with the `Frequency` associated type and is more convenient to work it through dependencies.
- The `Ticks` trait represent a type encapsulating a ticks value (i.e. a 24-bit or 32-bit integer), with support for wrapping arithmetic and conversion to/from `u32`/`usize`.
- Removed the `Time::max_tics(&self)` function. This was unused. Instead, `Ticks::max_value` is now available (and doesn't take any parameter, as it is a constant of the type).
- Added the `Ticks::expired` primitive, to unify an implementation that was duplicated in various capsules.
- Added `Time::ticks_from_ms` and similar utility functions, to avoid the need for dealing with frequencies in all the capsules. This centralizes the arithmetic logic in one place, to reduce code duplication.
- Added a new `Alarm::set_alarm_from_now` function, to support the common `self.set_alarm(self.now().wrapping_add(duration))` pattern and reduce code duplication.

### Testing Strategy

This pull request was tested with:
- `ci-travis`
- new unit tests added in `kernel::hil::timer`.

I haven't done much more tests for now, to first gather feedback on the implementation.

### TODO or Help Wanted

This pull request still needs:
- [ ] Checking that the widths are correct for all clocks.
- [ ] Double-checking that the affected capsules are still correct.

### Documentation Updated

- [ ] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

Issue #1513, opened 2019-12-17
Correctness of has_elapsed formula in virtual_alarm capsule?


gendx:
This issue adds to the list of recent issues/bugs related to timers (https://github.com/tock/tock/pull/1413, https://github.com/tock/tock/issues/1496, https://github.com/tock/tock/pull/1499).

**Context**: I started noticing issues when integrating with the [Segger RTT capsule](https://github.com/tock/tock/blob/master/capsules/src/segger_rtt.rs) on a nRF52840-DK board and writing a lot of messages on the console (via the console syscall). The app would freeze at some point, seemingly waiting forever for a "write completed" callback to come back from the console driver. I'm not 100% sure because the debug output itself would freeze, but by doing some "LED debugging" it seems to be the problem.

In the virtual_alarm capsule, various checks aims at comparing timestamps with each other, with respect to some reference point (such as "now").

The first one is checking if a new alarm set at timestamp `when` is before the current alarm (in which case the alarm should be updated to point at `when`).

https://github.com/tock/tock/blob/d52a619e640a5c8cdc9deb74232d145f66fade2d/capsules/src/virtual_alarm.rs#L85-L91

The second one is checking if an alarm has expired at timestamp `now`.

https://github.com/tock/tock/blob/d52a619e640a5c8cdc9deb74232d145f66fade2d/capsules/src/virtual_alarm.rs#L131-L133

I'm not sure what the logic is intended with the wrapping arithmetic, but it seems to assume that `cur_alarm >= now && when >= now` (in the first case), and that `now >= prev && alarm >= prev` (in the second case).
Otherwise, because timestamps are unsigned integers, the wrapping arithmetic will produce a very large number if one of the arguments is slighly before `now` (in the first case) or `prev` (in the second case).

However, these assumptions don't hold in practice. It's easy to generate examples by inserting some panic statements.

In the first case, a panic happens quite quickly with my example app.

```rust
if (cur_alarm as i32).wrapping_sub(now as i32) < 0
    || (when as i32).wrapping_sub(now as i32) < 0
{
    panic!("cur_alarm={}, when={}, now={}", cur_alarm, when, now);
}
```

```
Kernel panic at [...]/tock/capsules/src/virtual_alarm.rs:91:
	"cur_alarm=10351, when=10356, now=10353"
```

The second case panics much less often, but I could still trigger it at some point (you can see that the number of system ticks is much larger).

```rust
if (now as i32).wrapping_sub(prev as i32) < 0
[...]

PR #2089, opened 2020-09-01
Time redesign v3

phil-levis:
### Pull Request Overview

This pull request updates the time HIL to address a series of bug reports with the previous API (#1651, #1691, #1513). It also incorporates proposed changes by @gendx to generalize the width of counters/alarms/timers with an associated type rather than assume 32 bits (#1521). 

This has been implemented on all of the chips. It has been tested for the 24-bit nRF52 series, the 32-bit SAM4L, and 64-bit OpenTitan. 

The overall design and summary of the traits is described in 

https://github.com/tock/tock/blob/time-redesign-v3/doc/reference/trd-time.md

We will update this document and give it a TRD number when ready to merge.

There is also an update to the system call API: a new command for Alarm passes both a reference time and a dt. This new API can be used by using the `timer_v3_updates` branch of libtock-c. 

### Testing Strategy

This pull request was tested by compiling and testing on nrf52, SAM4L (imix), and OpenTitan (FPGA) boards. For imix and OT, it was tested using the multi_alarm_test and multi_timer_test tests in the kernel. On imix, it was tested in userspace by running a pair of multi_alarm_test processes. 

I was not able to test the userspace alarm driver on OpenTitan -- after struggling to get libtock-rs applications to run and librtock-c ones to compile I gave up. This is an important test because the capsule is 32 bits, and tries to automatically handle an underlying 64-bit Alarm.

### TODO or Help Wanted

This pull request needs userspace testing on OT (to test that 64-to-32 conversion works correctly for the userspace API). This PR updates the mtimer implementation to seed it with a value close to a 32-bit overflow. So you do not have to run the test very long. Any userspace application that uses an alarm should be a good test.
 
This pull request needs kernel testing on 
  - [x] arty_e1 (@bradjc )
  - [x] hifive (@alevy  )
  - [x] msp (@hudson-ayers )
  - [x] nano33ble (OK)
  - [x] nucleo ()
  - [x] redboard (@alistair23 )
  - [x] stm32f ()

To test, you need run a `multi_alarm_test`. I've added a `multi_alarm_test` for each board and modified each board's `main.rs` to  invoke it. Double-check you see a call to `multi_alarm_test::run_multi_alarm(mux_alarm)`.

This test starts 3 alarms (A, B, C). The `dt` of these alarms is random, with one in 11 alarms (randomly) having a `dt` of 0. A typical output of the test looks something like this (this is from OpenTitan):

```TestA: Alarm fired.
TestA@Ticks64(17033607736): Expected at Ticks64(17033607729) (diff = Ticks64(7)), setting alarm to Ticks64(17033616266) (delay = Ticks64(8537))
TestB: Alarm fired.
TestB@Ticks64(17033614398): Expected at Ticks64(17033614391) (diff = Ticks64(7)), setting alarm to Ticks64(17033626851) (delay = Ticks64(12462))
TestC: Alarm fired.
TestC@Ticks64(17033614581): Expected at Ticks64(17033614576) (diff = Ticks64(5)), setting alarm to Ticks64(17033618165) (delay = Ticks64(3592))
TestA: Alarm fired.
TestA@Ticks64(17033616273): Expected at Ticks64(17033616266) (diff = Ticks64(7)), setting alarm to Ticks64(17033629481) (delay = Ticks64(13214))
TestC: Alarm fired.
TestC@Ticks64(17033618172): Expected at Ticks64(17033618165) (diff = Ticks64(7)), setting alarm to Ticks64(17033626435) (delay = Ticks64(8270))
```

The `delay` value is the `dt` set for the next invocation of this Alarm. The `diff` value is the number of ticks between the desired firing time and a call to `now` in the firing. Note that this value is large (e.g., 7 ticks above!) mostly because of these print statements: formatting the numbers takes significant cycles at these timescales.

The three things to look for to make sure the test is running properly are:
  - All 3 Alarms are firing (one has not been lost or dropped or otherwise miscalculated)
  - The diff values are always positive
  - The diff of alarm firings after a delay of 0 are not excessively high (they will be higher than non-zero delays)

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


capsules/src/alarm.rs
```diff
@@ -1,22 +1,19 @@
...
-/// that the new tics value is not below the current counter
-/// so the alarm gets fired
-const MIN_TICS_AT_16KHZ: usize = 5;
-
+// This should transition to using Ticks
 #[derive(Copy, Clone, Debug)]
 enum Expiration {
     Disabled,
-    Abs(u32),
+    Enabled(u32, u32), // reference, dt
```
gendx, line 16: Note: I recently discovered that Rust's enumerations allow to use named fields. I think the following would be more readable and less error-prone than having a comment explaining which value is which.

```rust
enum Expiration {
    // ...
    Enabled {
        reference: u32,
        dt: u32,
    }
}
```

capsules/src/ble_advertising_driver.rs
```diff
@@ -133,7 +133,7 @@ enum BLEState {
 #[derive(Copy, Clone)]
 enum Expiration {
     Disabled,
-    Abs(u32),
+    Enabled(u32, u32),
```
gendx, line 136: Same comment here.

kernel/src/platform/scheduler_timer.rs
```diff
@@ -117,8 +109,6 @@ impl SchedulerTimer for () {
 
     fn disarm(&self) {}
 
-    fn arm(&self) {}
-
```
hudson-ayers, line 121: Given that disarm is now not called from `sched.rs`, does it still make sense for it to be part of this trait?

chips/earlgrey/src/timer.rs
```diff
@@ -71,59 +75,121 @@ impl<'a> RvTimer<'a> {
...
+        let val = expire.into_u64();
+        let high = (val >> 32) as u32;
+        let low = (val & 0xffffffff) as u32;
+
+        // Recommended approach for setting the two compare registers
+        // (RISC-V Privileged Architectures 3.1.15) -pal 8/6/20
+        regs.compare_low.set(0xffffffff);
+        regs.compare_high.set(high);
+        regs.compare_low.set(low);
+        self.registers.intr_enable.write(intr::timer0::SET);
```
hudson-ayers, line 167: This should also be done for the generic mtimer implementation in `arch/rv32i`, or else this will work on OpenTitan but fail on our other RISC-V chips as soon as the timer hits 2^32 ticks.

kernel/src/sched.rs
```diff
@@ -611,9 +611,7 @@ impl Kernel {
                     // underlying timer is not affected.
                     process.setup_mpu();
                     chip.mpu().enable_mpu();
-                    scheduler_timer.arm();
                     let context_switch_reason = process.switch_to();
-                    scheduler_timer.disarm();
```
hudson-ayers, line 616: comment to note that I discussed this with Amit out of band, and this change should be fine now that we no longer rely on a global variable for detecting systick expirations.

kernel/src/hil/time.rs
```diff
@@ -1,33 +1,270 @@
...
 
 use crate::ReturnCode;
+use core::cmp::{Eq, Ord, Ordering, PartialOrd};
+use core::fmt;
 
-pub trait Time<W = u32> {
+/// An integer type defining the width of a time value, which allows
+/// clients to know when wraparound will occur.
+
+pub trait Ticks: Clone + Copy + From<u32> + fmt::Debug + Ord + PartialOrd + Eq {
```
hudson-ayers, line 19: Given that we now support 64 bit timers below the HIL (RISC-V), should this also implement `From<u64>`?

kernel/src/hil/time.rs
```diff
@@ -1,33 +1,270 @@
...
+    if val <= T::max_value().into_u32() as u64 {
+        T::from(val as u32)
+    } else {
+        T::max_value()
+    }
+}
+
+/// Represents a static moment in time, that does not change over
+/// repeated calls to `Time::now`.
+pub trait Timestamp: Time {}
```
hudson-ayers, line 107: This trait is not implemented or used anywhere throughout this PR, is it actually needed?

kernel/src/hil/time.rs
```diff
@@ -1,33 +1,270 @@
...
+
+/// Interface for receiving notification when a particular time
+/// (`Counter` value) is reached. Clients use the
+/// [`AlarmClient`](trait.AlarmClient.html) trait to signal when the
+/// counter has reached a pre-specified value set in
+/// [`set_alarm`](#tymethod.set_alarm). Alarms are intended for
+/// low-level time needs that require precision (i.e., firing on a
+/// precise clock tick). Software that needs more functionality
+/// but can tolerate some jitter should use the `Timer` trait
+/// instead.
```
hudson-ayers, line 165: My observation is that currently `Timer` is barely used throughout Tock, despite there being many instances where jitter can be tolerated. What is the benefit of using `Timer` instead of an alarm? Is it just a cleaner interface, or are there performance benefits to using `Timer` when precision is not needed?

tools/print_tock_memory_usage.py
```diff
@@ -0,0 +1,422 @@
...
+                      together. Default: 1
+  -v, --verbose       Print verbose output.
+  -s, --show-waste    Show where RAM is wasted (due to padding)
+'''
+
+import os
+import re
+import sys
+import getopt
+import cxxfilt   # Demanging C++/Rust symbol names
```
alistair23: Why change this file in this PR?

arch/cortex-m4/src/lib.rs
```diff
@@ -58,7 +58,7 @@ pub unsafe extern "C" fn systick_handler() {
     // This will resume in the switch to user function where application state
     // is saved and the scheduler can choose what to do next.
     "
-    : : : : "volatile" );
+    : : : : : "volatile" );
```
alistair23: Why do we need to change this?

capsules/src/virtual_alarm.rs
```diff
@@ -4,7 +4,7 @@
 use core::cell::Cell;
 use kernel::common::cells::OptionalCell;
 use kernel::common::{List, ListLink, ListNode};
-use kernel::debug;
+//use kernel::debug;
```
alistair23: There are lots of debug prints that should be removed.

arch/cortex-m4/src/lib.rs
```diff
@@ -264,125 +39,3 @@ pub unsafe extern "C" fn switch_to_user(
...
-    mov    r1, 0     /* r1 = 0 */
-    tst    lr, #4    /* bitwise AND link register to 0b100 */
-    itte   eq        /* if lr==4, run next two instructions, else, run 3rd instruction. */
-    mrseq  r0, msp   /* r0 = kernel stack pointer */
-    addeq  r1, 1     /* r1 = 1, kernel was executing */
-    mrsne  r0, psp   /* r0 = userland stack pointer */"
-    : "={r0}"(faulting_stack), "={r1}"(kernel_stack)
-    :
-    : "r0", "r1"
-    : "volatile" );
```
alistair23: Why are you removing this?

kernel/src/hil/time.rs
```diff
@@ -1,35 +1,186 @@
...
+
+pub trait Alarm<'a>: Time {
+
+  /// Specify the callback for when the counter reaches the alarm
+  /// value. If there was a previously installed callback this call
+  /// replaces it.
+  fn set_client(&'a self, client: &'a dyn AlarmClient);    
+
+    
+  /// Specify when the callback should be called and enable it. Tthe
```
alistair23: s/Tthe/The/g

alistair23: Although the sentence doesn't make sense with the typo fixed

kernel/src/hil/time.rs
```diff
@@ -1,35 +1,186 @@
...
+  /// it it assured to execute *after* `now + dt`: it can be delayed
+  /// but will never fire early. The method takes `now` and `dt`
+  /// rather than a single value denoting the counter value so it can
+  /// distinguish between alarms which have very recently already
+  /// passed and those in the far far future (see #1651).
+  fn set_alarm(&self, now: Self::Ticks, dt: Self::Ticks);
+
+  /// Return the current alarm value. This is undefined at boot and
+  /// otherwise returns `now + dt` from the last call to `set_alarm`.
+  fn get_alarm(&self) -> Self::Ticks;
```
alistair23: What happens when an alarm is disarmed but this is called?

kernel/src/hil/time.rs
```diff
@@ -1,35 +1,186 @@
...
+
+  /// Disable the alarm and stop it from firing in the future.    
+  fn disarm(&self) -> ReturnCode;
```
alistair23: Shouldn't this have the same note as below about this possibly not clearing a pending callback?

capsules/src/ble_advertising_driver.rs
```diff
@@ -338,22 +338,25 @@ where
...
+                        next_dt = dt;
                         next_dist = t_dist;
                     }
                 }
                 Expiration::Disabled => {}
             });
         }
-        if next_alarm != u32::max_value() {
-            self.alarm.set_alarm(next_alarm);
+        if next_ref != u32::max_value() {
```
alistair23, line 361: Why not use an option here?

capsules/src/buzzer_driver.rs
```diff
@@ -40,7 +40,7 @@ use core::cmp;
 
 use kernel::common::cells::OptionalCell;
 use kernel::hil;
-use kernel::hil::time::Frequency;
+use kernel::hil::time::{Frequency};
```
alistair23: You don't need the curly braces here.

chips/earlgrey/src/plic.rs
```diff
@@ -42,7 +43,11 @@ pub unsafe fn clear_all_pending() {
 pub unsafe fn enable_all() {
     let plic: &PlicRegisters = &*PLIC_BASE;
     for enable in plic.enable.iter() {
-        enable.set(0xFFFF_FFFF);
+        // DANGER: For some reason USBDEV seems to be issuing
+        // unhandled CONNECT interrupts. So I have disabled them
+        // here. This should be changed back to 0xFFFFFFFF once
+        // this is resolved and before merging. -pal 8/26/20
+        enable.set(0x0000_00FF);
```
alistair23, line 50: This shouldn't be part of this PR

chips/earlgrey/src/plic.rs
```diff
@@ -2,6 +2,7 @@
 
 use kernel::common::registers::{register_bitfields, register_structs, ReadOnly, ReadWrite};
 use kernel::common::StaticRef;
+use kernel::debug;
```
alistair23, line 5: Why is this needed?

chips/earlgrey/src/plic.rs
alistair23: Also in general this just covers up bugs, the real cause should be identified.

I thought this was a HW issue fixed in the more recent bit streams?

doc/reference/trd-time.md
```diff
@@ -28,148 +29,342 @@ broad categories: alarms and timers. Alarms continuously increment a clock and
...
+```rust
+pub trait AlarmClient: OverflowClient {
+  fn alarm(&self);
+}
+
+pub trait Alarm: Time {
+  fn set_alarm(&self, reference: Self::Ticks, dt: Self::Ticks);
+  fn get_alarm(&self) -> Self::Ticks;
+  fn disarm(&self) -> ReturnCode;
+  fn set_alarm_client(&'a self, client: &'a dyn AlarmClient);
```
alistair23, line 216: Why not just set_client() like all the others?

kernel/src/hil/time.rs
phil-levis: There are lots of cases where we use Alarm where Timer is a better bet; typical cases are periodic operations (such as sensing, sending a packet). It happens to be historically that we didn't have a solid Timer implementation so everyone used Alarm.

chips/earlgrey/src/plic.rs
phil-levis: Yes, I haven't gotten to the bottom of this yet -- it will be fixed in the eventual PR.

kernel/src/platform/scheduler_timer.rs
phil-levis: If you think so, sure, although perhaps that should be a separate PR on main which we can merge into this?

chips/sam4l/src/ast.rs
```diff
@@ -25,7 +26,7 @@ use kernel::hil::Controller;
 /// less than or equal to four tics ahead of the current counter value, the
 /// alarm interrupt doesn't fire. Thus, we simply round up to at least eight
 /// tics. Seems safe enough and in practice has seemed to work.
-const ALARM0_SYNC_TICS: u32 = 8;
+const ALARM0_SYNC_TICS: u32 = 10;
```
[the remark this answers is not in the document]
phil-levis: Nice catch. Turns out having debug!() statements requires a larger value. :)

arch/rv32i/src/machine_timer.rs
```diff
@@ -54,50 +54,61 @@ impl MachineTimer<'_> {
...
         // Correct approach will require seperating MTIMECMP into separate registers. Then,
         // first, set the upper register to the largest value possible without setting
         // off the alarm; this way, we can set the lower register without setting
         // off the alarm, then set the upper register to the correct value.
+        let mut expire = reference.wrapping_add(dt);
+        let now = Self::Ticks::from(self.registers.mtime.get());
+        if !now.within_range(reference, expire) {
+            // expire has already passed, so fire immediately
+            // TODO(alevy): we probably need some wiggle room, but
+            //              I can't trivially figure out how much
```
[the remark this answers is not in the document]
phil-levis: Ah - completely missed this because earlgrey has its own timer. Will definitely try to implement this, although I cannot test it.

chips/earlgrey/src/timer.rs
phil-levis: Yes. Sorry I missed the machine_timer; we are currently only implemented on the platforms we could test.

kernel/src/hil/time.rs
phil-levis: It actually shouldn't. You can't make a Ticks32 from a 64 bit value. The presence of From<u32> is basically for the syscall driver, so it can convert from 32 bit values into Ticks. The assumption is that it is on top of a Ticks that is >= 32 bits in width. The nRF 24 bit Ticks have to expand to 32 (e.g.. by handing overflow) for the syscall driver to use.

kernel/src/hil/time.rs
phil-levis: It seemed like a good idea.

tools/print_tock_memory_usage.py
phil-levis: No idea -- shouldn't be.

capsules/src/virtual_alarm.rs
phil-levis: Later.

doc/reference/trd-time.md
phil-levis: Good question. It's because of cases in which you want to handle both overflows (from Counter) and compares (from Alarm). E.g, if a structure  If they are named the same thing it's a pain. Basically, for these abstractions it is quite possible that a single structure implements multiple traits, and so easily disambiguating their methods makes things easier.

chips/earlgrey/src/plic.rs
phil-levis: From when I was discovering that USBDEV was issuing connected interrupts. I'll clean this (and all of the other debug code) up before being ready for merging.

arch/cortex-m4/src/lib.rs
phil-levis: It's removed in master... not sure why it's showing up.

capsules/src/alarm.rs
phil-levis: Good idea.

capsules/src/ble_advertising_driver.rs
phil-levis: Because the previous implementation didn't and I was trying to minimize the changes. It would make a lot of sense (although use an additional word of memory).

capsules/src/ble_advertising_driver.rs
```diff
@@ -379,21 +383,22 @@ where
...
+                let exp = A::Ticks::from(reference.wrapping_add(dt));
+                let t0 = A::Ticks::from(app.alarm_data.t0);
+                let expired = t0.within_range(exp, now);
                 if expired {
                     if self.busy.get() {
                         // The radio is currently busy, so we won't be able to start the
                         // operation at the appropriate time. Instead, reschedule the
                         // operation for later. This is _kind_ of simulating actual
-                        // on-air interference
+                        // on-air interference. 3 seems like a small number of ticks.
```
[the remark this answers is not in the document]
phil-levis: No idea -- this was a long time ago.

doc/reference/trd-time.md
```diff
@@ -28,148 +29,342 @@ broad categories: alarms and timers. Alarms continuously increment a clock and
...
+pub trait OverflowClient {
+  fn overflow(&self);
+}
+
+pub trait Counter<'a>: Time {
+  fn start(&self) -> ReturnCode;
+  fn stop(&self) -> ReturnCode;
+  fn reset(&self) -> ReturnCode;
+  fn is_running(&self) -> bool;
+  fn set_overflow_client(&'a self, &'a dyn OverflowClient);
```
jrvanwhy, line 167: Presumably, `OverflowClient::overflow` would be called in an interrupt handler or a deferred callback, as that is the normal Tock practice for callbacks. If so, it is possible that the type implementing `OverflowClient` calls `Time::now` between when the overflow has occurred and when the callback is called.

In other words, the following sequence of events is possible (assumes an 8-bit counter):

1. Client code calls `Time::now`, which returns 250.
2. An overflow happens, marking an interrupt as pending.
3. Client code calls `Time::now`, which returns 12.
4. The main event loop runs, invoking the interrupt handler.
5. The Counter calls `OverflowClient::overflow`, notifying the client of the overflow.

This seems confusing, and it appears it would be difficult to handle correctly. Worse, it's unclear how this sequence interacts with `stop` and`reset` (what if `reset` is called between steps 2 and 5?).

I recognize that this design may be the best solution to a difficult problem, but if so there needs to be documentation of how these methods interact with delayed overflows.

tools/print_tock_memory_usage.py
alistair23: It's included in the PR. It looks like you need to rebase on master as you have ~60 commits here.

capsules/src/virtual_alarm.rs
alistair23: Isn't this a PR for ready for review and merge?

capsules/src/ble_advertising_driver.rs
alistair23: You have almost 3000 lines of additions in a single PR, I don't think converting this to an option would be too much change :P

chips/earlgrey/src/plic.rs
alistair23: I didn't realise this wasn't intended to be ready for merging. Why not mark it as a draft PR then?

doc/reference/trd-time.md
alistair23: Is it not worth naming the other ones this way as well? It seems strange that this set_client has a different name to all the others in Tock.

chips/earlgrey/src/plic.rs
alistair23: Ah, you do mention it in the description

kernel/src/platform/scheduler_timer.rs
hudson-ayers: this PR is the one that removes the use of `arm()` and `disarm()` from `sched.rs`, I can break all those changes out into a separate PR if you like, or just followup with that change once this merges.

chips/earlgrey/src/timer.rs
hudson-ayers: perhaps @bradjc can test on the Arty-e21 once the machine_timer is implemented

kernel/src/hil/time.rs
phil-levis: Note https://github.com/tock/tock/blob/a97b7cd66b6ed7e8d3e5658d51f0155de8c82c1d/kernel/src/hil/time.rs#L433

doc/reference/trd-time.md
```diff
@@ -28,148 +29,342 @@ broad categories: alarms and timers. Alarms continuously increment a clock and
...
+    // rounding down any fractions.
+    fn ticks_from_seconds(s: u32) -> Self::Ticks;
 
-    fn is_armed(&self) -> bool;
+    // Returns the number of ticks in the provided number of milliseconds,
+    // rounding down any fractions.
+    fn ticks_from_ms(ms: u32) -> Self::Ticks;
+
+    // Returns the number of ticks in the provided number of microseconds,
+    // rounding down any fractions.
```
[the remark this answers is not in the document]
phil-levis: Good question. I guess the issue is whether you want the semantics of "at least as long" or "at most as long." The motivation for rounding down (at most as long) was that if you are a little early you can always then do something to be exact to match an exact deadline. But if you are a little late there is nothing you can do to be exact.

doc/reference/trd-time.md
phil-levis: It really has to do with the fact that this set of traits are tightly entwined and it's quite normal for a single struct to implement multiple of them. I don't recall ever seeing this issue before in Tock, although it might have happened and so I missed it. If we wanted to change all of the set_clients I wouldn't be against it, but I don't think it's necessary. The analogy for me in natural language is when you have two things you disambiguate them with an adjective, but when you have one thing you don't ("the smaller binary" vs "the binary").

doc/reference/trd-time.md
phil-levis: Yes, this is a good point to mention in the document. It's a basic fact of live with deferred callbacks.

capsules/src/ble_advertising_driver.rs
phil-levis: Sure, happy to change it, I just wanted to scope this as "updating everything to the new abstractions" not "rewriting code to be more idiomatic." Given that it is 3,000 lines, asking people to also reason about the correctness of rewrites for different programming style seemed to me that it could be the straw that broke the camel's back.

Although utltmately at this point the hope is to get a detailed review of the abstractions and for people to see how those abstractions manifest in code. It's intended to be a final design review more than a specific code review.

doc/reference/trd-time.md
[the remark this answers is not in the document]
jrvanwhy: Even RTOSes don't guarantee an exact wakeup, they just guarantee a wakeup within `[target_wakeup_time, target_wakeup_time + max_permissible_latency]`.

arch/rv32i/src/machine_timer.rs
```diff
@@ -46,20 +39,27 @@ impl MachineTimer<'_> {
...
-        // Disable by setting the mtimecmp register to its max value, which
-        // we will never hit.
-        self.registers
-            .mtimecmp
-            .write(MTimeCmp::MTIMECMP.val(0xFFFF_FFFF_FFFF_FFFF));
+        // Disable by setting the mtimecmp register to its max value.
+        // In theory we won't hit it if we start at 0, but it's ridiculous that
+        // mtimer has no enable/disable mechanism.
+        self.registers.compare_high.set(0xFFFF_FFFF);
+        self.registers.compare_low.set(0xFFFF_FFFF);
```
hudson-ayers, line 46: I notice that in `set_alarm()` we set the interrupt enable bit for the mtimer, could we not just clear that bit here?

phil-levis:
This is blocked on #2100 for OpenTitan support.

kernel/src/hil/time.rs
phil-levis: It returns the value from the last call.

kernel/src/hil/time.rs
phil-levis: It should clear the pending callback. The behavior for `is_armed` is for when the callback is pending but has not executed (e.g., kernel is in a different bottom half that ran first). However, if you call `disarm()` you would make it no longer pending.

chips/earlgrey/src/plic.rs
phil-levis: #2100

doc/reference/trd-time.md
phil-levis: Addressed.

doc/reference/trd-time.md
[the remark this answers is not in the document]
phil-levis: Fair enough -- but it seemed useful to have an API that at least gives the possibility of high precision. Of course software delays may render this moot...

chips/earlgrey/src/timer.rs
phil-levis: @bradjc I updated rv32i `machine_timer`.

phil-levis:
I've incorporated almost all of the comments (except one that suggested I reimplement an existing approach).

I implemented the remaining chips.

I believe there is a bug in the Apollo3 stimer, but need to talk with someone who has one to verify. There's a comment in stimer on this.

phil-levis:
> > Ah, I find representing it as inequalities to be confusing, because who's to say whether `a <= b` when you have wraparound.
> > Would you agree with the statement "returns true if b is within the range [a, c) and false if it lines outside that range"?
> 
> Yes indeed, a two-way comparison `a <= b` does't mean anything in wraparound arithmetic. What I meant is indeed `b in [a, c)` in terms of range, i.e. `a <= b < c` (although such notation doesn't exist in common programming languages), which isn't the same as `a <= b && b < c`.
> 
> > Alternatively, using unsigned arithmetic, `(b - a) < (c - a)`?
> 
> This will be the implementation. But I find `b in [a, c)` clearer in terms of "human-language" description.
> 
> By the way, if one would want to check `b in [a, c]`, this would be equivalent to `c not in [a, b)`, i.e. `!is_chronological(a, c, b)`. One may think to write it as `b in [a, c+1)`, except that the semantics are different if `c+1 == a`.
> 
> We could provide two functions though:
> 
>     * `is_chronological_open(a, b, c)` i.e. `b in [a, c)` i.e. `(b - a) < (c - a)`.
> 
>     * `is_chronological_closed(a, b, c)` i.e. `b in [a, c]` i.e. `(b - a) <= (c - a)`.

This was resolved to be named `within_range` and self, so you would call `b.within_range(a, c)`.

kernel/src/platform/scheduler_timer.rs
```diff
@@ -64,26 +64,6 @@ pub trait SchedulerTimer {
...
-    ///
-    /// This does not stop the timer, but indicates to the SchedulerTimer that
-    /// an interrupt is no longer required (i.e. the process is no longer
-    /// executing). By not requiring an interrupt this may allow certain
-    /// implementations to be more efficient by removing the overhead of
-    /// handling the interrupt. The implementation may disable the underlying
-    /// interrupt if one has been set, depending on the requirements of the
-    /// implementation.
-    fn disarm(&self);
-
```
bradjc, line 86: I'm not sure why this interface has changed. We discussed this at length when adding the scheduler, and decided to decouple time keeping from interrupt generation. At the very least the comments in this file need to be updated.

However, I think changing this interface right after we just changed it should at least be its own PR.

bradjc (changes requested):

kernel/src/platform/scheduler_timer.rs
phil-levis: https://github.com/tock/tock/pull/2089#discussion_r481251980

phil-levis:
One of the ci-build test is failings for reasons I don't understand:

```
---- machine_timer::test_register_structs::MachineTimerRegisters::test_offsets stdout ----
thread 'machine_timer::test_register_structs::MachineTimerRegisters::test_offsets' panicked at 'Invalid start offset for field value_low (expected 0x4000 but was 0x0)', arch/rv32i/src/machine_timer.rs:10:1
error: test failed, to rerun pass '--lib'
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

How do I reproduce this test?

arch/rv32i/src/machine_timer.rs
```diff
@@ -2,31 +2,24 @@
...
-    MTimeCmp [
-        MTIMECMP OFFSET(0) NUMBITS(64) []
-    ],
-    MTime [
-        MTIME OFFSET(0) NUMBITS(64) []
-    ]
-];
+register_structs! {
+    pub MachineTimerRegisters {
+        (0x4000 => value_low: ReadWrite<u32>),
```
alistair23, line 12: You need a reserved entry before 0x4000

alistair23:
You should be able to reproduce the CI failure with: `make prepush`

Here is a link to where the failure is: https://github.com/tock/tock/pull/2089/files#r486688376

kernel/src/platform/scheduler_timer.rs
hudson-ayers: I actually think Brad is right -- I agreed too prematurely to the removal of `disarm()` and `arm()` from `sched.rs`. My mistake.

When we first added them to the interface, the idea was that they are an optimization, but are not required -- if a given `SchedulerTimer` implements both `arm()` and `disarm()`, it must implement bot, but if it implements neither, it can simply enable interrupts in `start()`. For all of the `SchedulerTimer` implementations that are not built on top of alarm, they can probably remain as they were, and for implementations built on top of alarm (aka `VirtualSchedulerTimer`), they can go without the optimization of disabling interrupts while not in userspace. So it probably makes more sense to keep `arm()` and `disarm()` in `sched.rs`, and not implement them for `VirtualSchedulerTimer`.

phil-levis: OK, sounds good.

phil-levis: This is a little more complicated. There isn't an `Alarm::enable` any more, in part due to concerns that you set the alarm to the future, disarm it, you pass the compare value, then re-arm it. Instead, you have to set the alarm again with set_alarm. Should I store those values in the struct and re-invoke `set_alarm`?

phil-levis:
> You should be able to reproduce the CI failure with: `make prepush`
> 
> Here is a link to where the failure is: https://github.com/tock/tock/pull/2089/files#r486688376

I have done `make prepush` and didn't seem to trigger it. Weird.

kernel/src/platform/scheduler_timer.rs
hudson-ayers: This only applies to the `VirtualSchedulerTimer`, right? I am proposing this as the new implementation:

```rust
impl<A: 'static + time::Alarm<'static>> SchedulerTimer for VirtualSchedulerTimer<A> {
    fn reset(&self) {
        self.alarm.disable();
    }

    fn start(&self, us: u32) {
        let tics = {
            // We need to convert from microseconds to native tics, which could overflow in 32-bit
            // arithmetic. So we convert to 64-bit. 64-bit division is an expensive subroutine, but
            // if `us` is a power of 10 the compiler will simplify it with the 1_000_000 divisor
            // instead.
            let us = us as u64;
            let hertz = A::Frequency::frequency() as u64;

            (hertz * us / 1_000_000) as u32
        };
        let fire_at = self.alarm.now().wrapping_add(tics);
        self.alarm.set_alarm(fire_at);
    }

    fn arm(&self) {}

    fn disarm(&self) {}
```

hudson-ayers:
> > You should be able to reproduce the CI failure with: `make prepush`
> > Here is a link to where the failure is: https://github.com/tock/tock/pull/2089/files#r486688376
> 
> I have done `make prepush` and didn't seem to trigger it. Weird.

`make prepush` only runs a quick set of checks, not the full CI suite, such that it is suitable for being wired to a pre-push hook. `make ci-nosetup` runs all CI except stuff that requires installing extra tools like QEMU.

`make  ci-job-archs` is probably the fastest way to run the specific test you see failing.

kernel/src/platform/scheduler_timer.rs
hudson-ayers: We should update the comments on the trait to indicate that `arm()`/`disarm()` are optional, but that if `disarm()` is implemented `arm()` must be implemented, and that if `arm()` is not implemented `start()` must enable interrupts.

hudson-ayers: Happy to make a separate PR with those changes if that's easier.

arch/rv32i/src/machine_timer.rs
phil-levis: Fixed. Thanks!

kernel/src/platform/scheduler_timer.rs
phil-levis: OK, sounds good.

phil-levis: I will update the VirtualSchedulerTimer implementation accordingly.

phil-levis: OK, fixed. `arm`/`disarm` are back in.

phil-levis:
> > > You should be able to reproduce the CI failure with: `make prepush`
> > > Here is a link to where the failure is: https://github.com/tock/tock/pull/2089/files#r486688376
> > 
> > 
> > I have done `make prepush` and didn't seem to trigger it. Weird.
> 
> `make prepush` only runs a quick set of checks, not the full CI suite, such that it is suitable for being wired to a pre-push hook. `make ci-nosetup` runs all CI except stuff that requires installing extra tools like QEMU.
> 
> `make ci-job-archs` is probably the fastest way to run the specific test you see failing.

Awesome, thank you.

bradjc (approved):

hudson-ayers (commented):
This PR now adds individual `multi_alam_test.rs` tests to every board, is there a reason for not using the (also introduced in this PR) `MultiAlarmTestComponent`?

hudson-ayers:
@lebakassemmerl are you able to test this PR on the MSP432? Testing should require just checking out this branch, flashing the kernel to your board, and observing the debug output, as this branch currently calls the `multi_alarm_test` by default for each board.

lebakassemmerl:
> @lebakassemmerl are you able to test this PR on the MSP432? Testing should require just checking out this branch, flashing the kernel to your board, and observing the debug output, as this branch currently calls the `multi_alarm_test` by default for each board.

I tested it right now, and as far as I can tell, it seems to work fine. I flashed the kernel plus the blink-example. 
Unfortunately I don't have the time right now to check if everything really works as expected, because I go on vacation tomorrow for about a week. However, you find the output of the serial console below if you want to check something by yourself (:


test_output

alistair23:
I think something is wrong with the Apollo3, I see this:

```
Apollo3 chip revision: B
Loading processes from flash=0x00040000-0x0800FFFF into sram=0x10002000-0x1005FFFF
Starting multi alarm test.
Starting random alarm test TestA.
TestA@Ticks32(0): Expected at Ticks32(0) (diff = Ticks32(0)), setting alarm to Ticks32(19175) (delay = Ticks32(19181))
Starting random alarm test TestB.
TestB@Ticks32(0): Expected at Ticks32(0) (diff = Ticks32(0)), setting alarm to Ticks32(0) (delay = Ticks32(0))
Starting random alarm test TestC.
TestC@Ticks32(0): Expected at Ticks32(0) (diff = Ticks32(0)), setting alarm to Ticks32(1804) (delay = Ticks32(1808))
```

phil-levis:
> 
> 
> I think something is wrong with the Apollo3, I see this:
> 
> ```
> Apollo3 chip revision: B
> Initialization complete. Entering main loop
> Loading processes from flash=0x00040000-0x0800FFFF into sram=0x10002000-0x1005FFFF
> Starting multi alarm test.
> Starting random alarm test TestA.
> TestA@Ticks32(0): Expected at Ticks32(0) (diff = Ticks32(0)), setting alarm to Ticks32(19175) (delay = Ticks32(19181))
> Starting random alarm test TestB.
> TestB@Ticks32(0): Expected at Ticks32(0) (diff = Ticks32(0)), setting alarm to Ticks32(0) (delay = Ticks32(0))
> Starting random alarm test TestC.
> TestC@Ticks32(0): Expected at Ticks32(0) (diff = Ticks32(0)), setting alarm to Ticks32(1804) (delay = Ticks32(1808))
> ```

This looks like interrupts aren't being triggered. Can you check the `stimer` code to see what is changed/might be the problem? I can stare at it but can't test it...

phil-levis:
> I think something is wrong with the Apollo3, I see this:
> 
> ```
> Apollo3 chip revision: B
> Initialization complete. Entering main loop
> Loading processes from flash=0x00040000-0x0800FFFF into sram=0x10002000-0x1005FFFF
> Starting multi alarm test.
> Starting random alarm test TestA.
> TestA@Ticks32(0): Expected at Ticks32(0) (diff = Ticks32(0)), setting alarm to Ticks32(19175) (delay = Ticks32(19181))
> Starting random alarm test TestB.
> TestB@Ticks32(0): Expected at Ticks32(0) (diff = Ticks32(0)), setting alarm to Ticks32(0) (delay = Ticks32(0))
> Starting random alarm test TestC.
> TestC@Ticks32(0): Expected at Ticks32(0) (diff = Ticks32(0)), setting alarm to Ticks32(1804) (delay = Ticks32(1808))
> ```

I just pushed a change, where the apollo3 driver now uses a `minimum_dt`. It didn't in the past, which was likely a bug (delta of 0). Can you try again?

hudson-ayers (dismissed):

bradjc:
This is not working for me on the arty-e21. The timer has always been, shall we say, finicky on the arty-e21, in that setting "too short" of a timeout doesn't seem work. I'm trying an app I know works on master (and just re-tested) that just blinks an led. On this branch I never seem to get a callback.

Preliminary strace:

```
Loading processes from flash=0x40430000-0x4052FFFF into sram=0x80004000-0x8000FFFF
Loaded process[0] from flash=0x40430000-0x4043FFFF into sram=0x80004000-0x80005EC3 = "led-1"
[0] function_call @0x40430088(0x40430060, 0x80004000, 0x1ec4, 0x80004c00)
[0] memop(0, 0x80005318) = 0x0 = SUCCESS
[0] memop(10, 0x80004800) = 0x0 = SUCCESS
[0] memop(11, 0x80005318) = 0x0 = SUCCESS
[0] cmd(0x2, 2, 0x0, 0x0) = 0x0 = SUCCESS
[0] cmd(0x0, 1, 0x0, 0x0) = 0x8000 = SuccessWithValue { value: 32768 }
[0] cmd(0x0, 2, 0x0, 0x0) = 0xffffffff = SuccessWithValue { value: 4294967295 }
[0] cmd(0x0, 2, 0x0, 0x0) = 0xffffffff = SuccessWithValue { value: 4294967295 }
[0] remove_pending_callbacks[0x0:0] = 0 callback(s) removed
[0] subscribe(0x0, 0, @0x404301da, 0x0) = 0x0 = SUCCESS
[0] cmd(0x0, 4, 0x13e7f, 0x0) = 0x13e7f = SuccessWithValue { value: 81535 }
[0] yield
```

alevy:
Works for me as expected on Hail (unsurprising, but a sanity check), but not for HiFive1. I'm getting qemu warnings suggesting the clic is not actually implemented in Qemu:

```
qemu-system-riscv32 -M sifive_e,revb=true -kernel /home/alevy/hack/tock/target/riscv32imac-unknown-none-elf/release/hifive1.elf  -nographic
qemu-system-riscv32: clint: time_lo write not implemented
qemu-system-riscv32: clint: time_hi write not implemented
qemu-system-riscv32: clint: time_lo write not implemented
qemu-system-riscv32: clint: time_lo write not implemented
qemu-system-riscv32: clint: time_hi write not implemented
qemu-system-riscv32: clint: time_lo write not implemented
qemu-system-riscv32: clint: time_lo write not implemented
qemu-system-riscv32: clint: time_hi write not implemented
qemu-system-riscv32: clint: time_lo write not implemented
```

Maybe @alistair23 has an idea about this? (i'm running qemu 5.1, for reference)

hudson-ayers:
@alexandruradovici @cosmindanielradu19 @rajivr is one of you able to test this PR for either the stm32f3 or stm32f4, or one of the nucleo boards? I am not sure any of the core WG members have hardware, but we don't want to merge this until it has been tested on each chip. Testing should require simply flashing the board and posting the debug output in this thread, as the multi_alarm_test is enabled in main for each board.

cosmindanielradu19:
> @alexandruradovici @cosmindanielradu19 @rajivr is one of you able to test this PR for either the stm32f3 or stm32f4, or one of the nucleo boards? I am not sure any of the core WG members have hardware, but we don't want to merge this until it has been tested on each chip. Testing should require simply flashing the board and posting the debug output in this thread, as the multi_alarm_test is enabled in main for each board.

this is the output for the stm32f3 board

image

cosmindanielradu19:
> @alexandruradovici @cosmindanielradu19 @rajivr is one of you able to test this PR for either the stm32f3 or stm32f4, or one of the nucleo boards? I am not sure any of the core WG members have hardware, but we don't want to merge this until it has been tested on each chip. Testing should require simply flashing the board and posting the debug output in this thread, as the multi_alarm_test is enabled in main for each board.

this is the output for the stm32f3 board

image

cosmindanielradu19:
> @alexandruradovici @cosmindanielradu19 @rajivr is one of you able to test this PR for either the stm32f3 or stm32f4, or one of the nucleo boards? I am not sure any of the core WG members have hardware, but we don't want to merge this until it has been tested on each chip. Testing should require simply flashing the board and posting the debug output in this thread, as the multi_alarm_test is enabled in main for each board.

this is the output for the stm32f4

image

alexandruradovici:
This is the output from Nucelo 429ZI

```
␀Initialization complete. Entering main loop
Starting multi alarm test.
Starting random alarm test TestA.
TestA@Ticks32(16): Expected at Ticks32(0) (diff = Ticks32(16)), setting alarm to Ticks32(19191) (delay = Ticks32(19181))
Starting random alarm test TestB.
TestB@Ticks32(31): Expected at Ticks32(0) (diff = Ticks32(31)), setting alarm to Ticks32(31) (delay = Ticks32(0))
Starting random alarm test TestC.
TestC@Ticks32(46): Expected at Ticks32(0) (diff = Ticks32(46)), setting alarm to Ticks32(1850) (delay = Ticks32(1808))
TestC: Alarm fired.
TestC@Ticks32(61): Expected at Ticks32(1850) (diff = Ticks32(4294965507)), setting alarm to Ticks32(9349) (delay = Ticks32(9291))
TestB: Alarm fired.
TestB@Ticks32(76): Expected at Ticks32(31) (diff = Ticks32(45)), setting alarm to Ticks32(76) (delay = Ticks32(0))
TestC: Alarm fired.
TestC@Ticks32(9352): Expected at Ticks32(9349) (diff = Ticks32(3)), setting alarm to Ticks32(26125) (delay = Ticks32(16775))
TestB: Alarm fired.
TestB@Ticks32(9367): Expected at Ticks32(76) (diff = Ticks32(9291)), setting alarm to Ticks32(9367) (delay = Ticks32(0))
TestB: Alarm fired.
TestB@Ticks32(9382): Expected at Ticks32(9367) (diff = Ticks32(15)), setting alarm to Ticks32(13503) (delay = Ticks32(4129))
TestB: Alarm fired.
TestB@Ticks32(13506): Expected at Ticks32(13503) (diff = Ticks32(3)), setting alarm to Ticks32(25112) (delay = Ticks32(11613))
TestA: Alarm fired.
TestA@Ticks32(19193): Expected at Ticks32(19191) (diff = Ticks32(2)), setting alarm to Ticks32(21253) (delay = Ticks32(2064))
TestA: Alarm fired.
TestA@Ticks32(21256): Expected at Ticks32(21253) (diff = Ticks32(3)), setting alarm to Ticks32(30801) (delay = Ticks32(9548))
TestB: Alarm fired.
TestB@Ticks32(25115): Expected at Ticks32(25112) (diff = Ticks32(3)), setting alarm to Ticks32(44205) (delay = Ticks32(19096))
TestC: Alarm fired.
TestC@Ticks32(26128): Expected at Ticks32(26125) (diff = Ticks32(3)), setting alarm to Ticks32(50385) (delay = Ticks32(24258))
TestA: Alarm fired.
TestA@Ticks32(30804): Expected at Ticks32(30801) (diff = Ticks32(3)), setting alarm to Ticks32(47833) (delay = Ticks32(17031))
TestB: Alarm fired.
TestB@Ticks32(44207): Expected at Ticks32(44205) (diff = Ticks32(2)), setting alarm to Ticks32(46182) (delay = Ticks32(1979))
TestB: Alarm fired.
TestB@Ticks32(46185): Expected at Ticks32(46182) (diff = Ticks32(3)), setting alarm to Ticks32(55644) (delay = Ticks32(9462))
TestA: Alarm fired.
TestA@Ticks32(47836): Expected at Ticks32(47833) (diff = Ticks32(3)), setting alarm to Ticks32(72349) (delay = Ticks32(24514))
TestC: Alarm fired.
TestC@Ticks32(50388): Expected at Ticks32(50385) (diff = Ticks32(3)), setting alarm to Ticks32(57520) (delay = Ticks32(7141))
TestB: Alarm fired.
TestB@Ticks32(55647): Expected at Ticks32(55644) (diff = Ticks32(3)), setting alarm to Ticks32(72591) (delay = Ticks32(16946))
TestC: Alarm fired.
TestC@Ticks32(57523): Expected at Ticks32(57520) (diff = Ticks32(3)), setting alarm to Ticks32(72139) (delay = Ticks32(14624))
TestC: Alarm fired.
TestC@Ticks32(72142): Expected at Ticks32(72139) (diff = Ticks32(3)), setting alarm to Ticks32(94243) (delay = Ticks32(22108))
TestA: Alarm fired.
TestA@Ticks32(72351): Expected at Ticks32(72349) (diff = Ticks32(2)), setting alarm to Ticks32(79739) (delay = Ticks32(7397))
TestB: Alarm fired.
TestB@Ticks32(72594): Expected at Ticks32(72591) (diff = Ticks32(3)), setting alarm to Ticks32(97022) (delay = Ticks32(24429))
TestA: Alarm fired.
TestA@Ticks32(79742): Expected at Ticks32(79739) (diff = Ticks32(3)), setting alarm to Ticks32(94615) (delay = Ticks32(14881))
TestC: Alarm fired.
TestC@Ticks32(94246): Expected at Ticks32(94243) (diff = Ticks32(3)), setting alarm to Ticks32(99232) (delay = Ticks32(4991))
TestA: Alarm fired.
TestA@Ticks32(94618): Expected at Ticks32(94615) (diff = Ticks32(3)), setting alarm to Ticks32(116975) (delay = Ticks32(22364))
TestB: Alarm fired.
TestB@Ticks32(97025): Expected at Ticks32(97022) (diff = Ticks32(3)), setting alarm to Ticks32(104328) (delay = Ticks32(7312))
TestC: Alarm fired.
TestC@Ticks32(99235): Expected at Ticks32(99232) (diff = Ticks32(3)), setting alarm to Ticks32(111705) (delay = Ticks32(12474))
TestB: Alarm fired.
TestB@Ticks32(104331): Expected at Ticks32(104328) (diff = Ticks32(3)), setting alarm to Ticks32(119118) (delay = Ticks32(14795))
TestC: Alarm fired.
TestC@Ticks32(111708): Expected at Ticks32(111705) (diff = Ticks32(3)), setting alarm to Ticks32(131662) (delay = Ticks32(19957))
TestA: Alarm fired.
TestA@Ticks32(116978): Expected at Ticks32(116975) (diff = Ticks32(3)), setting alarm to Ticks32(122220) (delay = Ticks32(5247))
TestB: Alarm fired.
TestB@Ticks32(119121): Expected at Ticks32(119118) (diff = Ticks32(3)), setting alarm to Ticks32(141393) (delay = Ticks32(22279))
TestA: Alarm fired.
TestA@Ticks32(122223): Expected at Ticks32(122220) (diff = Ticks32(3)), setting alarm to Ticks32(134949) (delay = Ticks32(12730))
```

alexandruradovici:
This app still crashes:

```c
int main(void) {
  putnstr_async(hello, sizeof(hello), nop, NULL);
  for (int i=0; ; i++) {
    printf ("%d\n", i);
    delay_ms (0);
  }
  return 0;
}
```

Output:

```
Hello World!
␀0
```

bradjc:
I've gotten things closer with arty-e21 and riscv, but something is weird with the `dt` value I'm getting in machine_timer for `set_alarm()`. Here is a debug!:

```
val high 1 low 269321 expire Ticks64(4295236617)
```

The expire I'm getting is huge when trying to set a 2500 ms alarm. For some reason a `1` is getting put in the upper 32 bits.

bradjc:
And my commit that updates the register map for machine timer and changes the disable back: https://github.com/tock/tock/commit/220e9f952d6425514c550192a1ca07660ee0b686

alevy:
On the HiFive, with @bradjc's changes so far, it seems to work (the buffer becomes full and it stops printing, but I think that's a separate issue), except the numbers are not consistent. Looking into that.

phil-levis:
> 
> 
> I've gotten things closer with arty-e21 and riscv, but something is weird with the `dt` value I'm getting in machine_timer for `set_alarm()`. Here is a debug!:
> 
> ```
> val high 1 low 269321 expire Ticks64(4295236617)
> ```
> 
> The expire I'm getting is huge when trying to set a 2500 ms alarm. For some reason a `1` is getting put in the upper 32 bits.

Just to double-check: this is occurring in a call from user space or from the kernel?

bradjc:
I tried the kernel multi alarm test on arty-e21 with the latest commits and it appears to work. I also tried my `led-1` test app and that now works as well.

However, I also switched back the disable_system_timer and the is_armed check to use the interrupt rather than the 0xfffffffffffffff trick, and it seemed like multi_alarm_test still works, but the led-1 userspace app did not work.

phil-levis:
OK. so it sounds like the current code does work? We can revisit the MIE vs. 0xFFFF_FFFF later once things settle down.

I tried for about two hours to get userspace libtock-c apps working on OpenTitan with no success.

bradjc:
Yes I'll check the box. Just need to remove the test from running by default and I'll approve.

phil-levis:
> Yes I'll check the box. Just need to remove the test from running by default and I'll approve.

Great. I'll wait on apollo3, when it is good I will then go through and remove all of the tests/debug output, do the final cleanup, and ping you.

chips/apollo3/src/stimer.rs
```diff
@@ -115,45 +119,87 @@ impl<'a> STimer<'_> {
...
-    fn set_alarm(&self, tics: u32) {
-        let regs = self.registers;
+impl<'a> Counter<'a> for STimer<'a> {
+    fn set_overflow_client(&'a self, _client: &'a dyn OverflowClient) {
+        //self.overflow_client.set(client);
+    }
 
+    fn start(&self) -> ReturnCode {
         // Set the clock source
-        regs.stcfg.write(STCFG::CLKSEL::XTAL_DIV2);
```
alistair23, line 131: This is no longer correctly called.

`STCFG::CLKSEL::XTAL_DIV2` is set only in `Counter` `start()`.

If not using the `Counter` but only using the `Alarm` this is never set.

Adding `self.registers.stcfg.write(STCFG::CLKSEL::XTAL_DIV2);` to `set_alarm()` fixes this.

alistair23 (changes requested):

bradjc:
My latest output from hifive1b hardware after some fixes:

```
ATE0-->ATE0
OK
AT+BLEINIT=0-->OK
AT+CWMODE=0-->OK

HiFive1 initialization complete.
Entering main loop.
Starting multi alarm test.
Starting random alarm test TestA.
now Ticks64(1938) dt Ticks64(39284) val 41183
TestA@Ticks64(1905): Expected at Ticks64(0) (diff = Ticks64(1905)), setting alarm to Ticks64(41183) (delay = Ticks64(39284))
Starting random alarm test TestB.
now Ticks64(2430) dt Ticks64(0) val 2430
TestB@Ticks64(2415): Expected at Ticks64(0) (diff = Ticks64(2415)), setting alarm to Ticks64(2415) (delay = Ticks64(0))
Starting random alarm test TestC.
now Ticks64(2675) dt Ticks64(3703) val 6359
TestC@Ticks64(2660): Expected at Ticks64(0) (diff = Ticks64(2660)), setting alarm to Ticks64(6359) (delay = Ticks64(3703))
here??
TestC: Alarm fired.
TestC@Ticks64(6445): Expected at Ticks64(6359) (diff = Ticks64(86)), setting alarm to Ticks64(25471) (delay = Ticks64(19029))
TestB: Alarm fired.
TestB@Ticks64(6544): Expected at Ticks64(2415) (diff = Ticks64(4129)), setting alarm to Ticks64(6544) (delay = Ticks64(0))
TestA: Alarm fired.
TestA@Ticks64(6641): Expected at Ticks64(41183) (diff = Ticks64(18446744073709517074)), setting alarm to Ticks64(10865) (delay = Ticks64(4228))
now Ticks64(6795) dt Ticks64(0) val 6795
here??
TestC: Alarm fired.
TestC@Ticks64(6882): Expected at Ticks64(25471) (diff = Ticks64(18446744073709533027)), setting alarm to Ticks64(41235) (delay = Ticks64(34355))
TestB: Alarm fired.
TestB@Ticks64(7001): Expected at Ticks64(6544) (diff = Ticks64(457)), setting alarm to Ticks64(7001) (delay = Ticks64(0))
TestA: Alarm fired.
TestA@Ticks64(7088): Expected at Ticks64(10865) (diff = Ticks64(18446744073709547839)), set
*** DEBUG BUFFER FULL ***
```

The "now" lines are when `set_alarm()` is called in machine_timer.rs, and the "here??" lines are when `handle_interrupt()` is called in machine_timer.rs.

It does still look like the virtualizer is calling set_alarm for C even though B should be earlier. When the interrupt comes through for C, it looks like the virtualizer does detect that both C and B should be fired.

I'm guessing that the machine timer _does_ fire for B right away, but the interrupt is not handled because timer C's alarm is set before the scheduler gets around to handling interrupts.

phil-levis:
> It does still look like the virtualizer is calling set_alarm for C even though B should be earlier. When the interrupt comes through for C, it looks like the virtualizer does detect that both C and B should be fired.
> 
> I'm guessing that the machine timer _does_ fire for B right away, but the interrupt is not handled because timer C's alarm is set before the scheduler gets around to handling interrupts.

This is super helpful, thank you. I will look tomorrow.

phil-levis:
> It does still look like the virtualizer is calling set_alarm for C even though B should be earlier. When the interrupt comes through for C, it looks like the virtualizer does detect that both C and B should be fired.
> 
> I'm guessing that the machine timer _does_ fire for B right away, but the interrupt is not handled because timer C's alarm is set before the scheduler gets around to handling interrupts.

Whatever is going on is more insidious than just this. It also looks like whenever an interrupt fires, all three Alarms fire even if some of them are in the future (hence the huge diff values, which represent small negative numbers). These point at a common problem, incorrectly calculating if an Alarm X has expired at time T.

I recreated these cases on imix (ordering of Alarms) and it works fine. I am wondering if this has something to do with Ticks64 and if its inequality tests are wrong; I am going to test on OpenTitan.

phil-levis:
I think that the bugs related to the speed at which the RISCV clocks were going in comparison to their CPU speed. The fact that the tests start in the boot sequence exacerbates this, as there's a long-running computation. The key problem was this:

```
Alarm B    |--------|
Alarm C                   |------------------|
```

Alarm B is very short, and the time until AlarmC is long enough that its reference (start point) is outside the interval of B. The core logic in the virtualizer assumed this wouldn't happen. If it saw that the current earliest expiration fell outside the interval of an alarm A, it assumed it mean that alarm A had expired. I added logic to handle the case that considers A may be in the future and not just the past. Please try now.

kernel/src/hil/time.rs
```diff
@@ -1,33 +1,276 @@
...
+    fn timer(&self);
+}
+
+/// Interface for controlling callbacks when an interval has passed.
+/// This interface is intended for software that requires repeated
+/// and/or one-shot timers and is willing to experience some jitter or
+/// imprecision in return for a simpler API that doesn't require
+/// actual calculation of counter values. Software that requires more
+/// precisely timed callbacks should use the `Alarm` trait instead.
+pub trait Timer<'a>: Time {
```
lschuermann, line 224: I don't think that this supertrait bound to `Time` is justified. A `Timer` is a device operating on a relative time interval, namely the one from enabling until the specified number of ticks have been counted down. Thereby it does not necessarily have a notion of "absolute" (monotonically increasing) time as mandated by the `Time` trait. An example for hardware that implements solely a `Timer` without an "uptime value" is the LiteX Timer core in its default configuration.

I have [implemented a proposed fix here](https://github.com/lschuermann/tock/commit/3f9892c336efc9a5c650594e546bf17309adbdce), feel free to pull this in or request changes.

This actually solves a second problem with my hardware: **if** (not always the case) configured with a system clock, its width is `u64` whereas the timer will be `u32` by default. Therefore it is important to have `Ticks` (and possibly `Frequency`) independent from the `Time` trait as well.

For `Alarm` both of the constraints make sense. Setting an alarm to a timestamp in the future necessarily requires knowing the current time and the alarm time must be in the same field (word size) as the time.

lschuermann (changes requested):

kernel/src/hil/time.rs
phil-levis: Not changing the APIs at the zeroth hour of final testing. 

It inherits from Time in part because this gives it Ticks and Frequency. 

Why is its width u32 by default?

lschuermann: > Not changing the APIs at the zeroth hour of final testing.

Makes sense. I am sorry that this came up so late and I was not aware of the plans on merging in the (very) near future. I can open another PR once this one it merged. My "fix" does not necessarily need to meet any release deadlines and therefore has a bit of time.

> It inherits from Time in part because this gives it Ticks and Frequency.

I see. If that is the _only_ reason, my suspicion of `Timer` not requiring a measure of "absolute" time, and hence my fix does seem to be correct.

> Why is its width u32 by default?

LiteX Timers (`Timer`-trait) simply have a default width of 32bits. Enabling the uptime feature on LiteX SoCs adds a 64bit uptime (`Time`-trait) counter to this timer. I'd like to allow Tock to run on the defaults.

I don't however think that this is the issue here, just an instance of where the flexibility of the current HIL is insufficient. I don't have other specific examples but could think that other chips might have different word-widths for different time/alarm/timer peripherals attached to different clocks or even an integrated RTC. I think we should allow this flexibility which is as simple as adding a `Ticks` type to the timer independent of `Time`/`Alarm` to allow independent implementation of a `Timer` without a `Time` (whereas `Ticks` and `Frequency` of `Time` and `Alarm` should always coincide since an `Alarm` must always have a notion of time).

lschuermann: I can't seem to take this "change suggestion" back so ignore please. :)

chips/apollo3/src/stimer.rs
phil-levis: Does it make sense to select the clock source every time an alarm is set? This is a configuration action so should in theory only be done whenever the clock needs to be changed. Other alarms have a configure call that does this. Or does the apollo3 require you reselect the clock every time you set the compare register?

phil-levis: I updated the boot sequence to start the counter.

kernel/src/hil/time.rs
phil-levis: A `Timer` can need `Time` beyond `Frequency` and `Ticks`. They aren't necessarily always operating in terms of a relative time. For example, it could be that a network protocol does an exchange to decide on a regular communication interval. It needs to be able to set not only the frequency (the interval) but also its phase. By removing `Time`, you're unnecessarily constraining `Timer` only to settings in which you do not require any notion of a shared or global time base. 

Another example use case: your software wants to sleep/do nothing until a particular point in future time (sync to a phase) then start a periodic timer that fires 5 times. E.g., wants to wait until 10:23 then sample temperature 5 times at 1Hz.

lschuermann: I get your point. I don't however think that the current `Timer` trait provides us sufficient methods to guarantee a constant phase offset with respect to a global/shared time, since we can only set it in terms of relative (countdown) time. I can imagine a Timer with constant phase offset with respect to a global/shared time can be much better built using a repeatedly set Alarm, since I can set the timer to an absolute value of the global/shared time. Currently `Timer` only offers a constant phase offset with respect to the previous callback, namely using the `repeating()` method and given HW support.

More pressing for me is what I should do in my scenario. I do have a `Timer`-compatible HW peripheral, but that is not also a `Time`-compatible peripheral. Which combination of traits could abstract this peripheral?

What would you think of a trait `FixedPhaseTimer: Time + Timer` that could be implemented given sufficient HW support and actually provides a method such as `phase_repeating(start: Self::Ticks, interval: Self::Ticks)`? I can imagine that being very useful in my upcoming work.

phil-levis: I guess I don't understand how your hardware is incompatible. It's OK to have multiple implementations of Time/Counter. E.g., you could have one module that reads it as 64 bits, another that reads it as 32. Can you point me at the sections of the datasheet so I can read and understand better?

bradjc:
Ok looks good on hifive1b:

```
ATE0-->ATE0
OK
AT+BLEINIT=0-->OK
AT+CWMODE=0-->OK

HiFive1 initialization complete.
Entering main loop.
Starting multi alarm test.
Starting random alarm test TestA.
val 41172
TestA@Ticks64(1894): Expected at Ticks64(0) (diff = Ticks64(1894)), setting alarm to Ticks64(41172) (delay = Ticks64(39284))
Starting random alarm test TestB.
val 2356
TestB@Ticks64(2332): Expected at Ticks64(0) (diff = Ticks64(2332)), setting alarm to Ticks64(2332) (delay = Ticks64(0))
Starting random alarm test TestC.
TestC@Ticks64(2595): Expected at Ticks64(0) (diff = Ticks64(2595)), setting alarm to Ticks64(6294) (delay = Ticks64(3703))
int
TestB: Alarm fired.
TestB@Ticks64(2812): Expected at Ticks64(2332) (diff = Ticks64(480)), setting alarm to Ticks64(2812) (delay = Ticks64(0))
val 2903
int
TestB: Alarm fired.
TestB@Ticks64(3037): Expected at Ticks64(2812) (diff = Ticks64(225)), setting alarm to Ticks64(3037) (delay = Ticks64(0))
val 3143
int
TestB: Alarm fired.
TestB@Ticks64(3229): Expected at Ticks64(3
*** DEBUG BUFFER FULL ***
int
TestC: Alarm fired.
TestC@Ticks64(6370): Expected at Ticks64(6294) (diff = Ticks64(76)), setting alarm to Ticks64(25396) (delay = Ticks64(19029))
val 11678
int
TestB: Alarm fired.
TestB@Ticks64(11759): Expected at Ticks64(11678) (diff = Ticks64(81)), setting alarm to Ticks64(35535) (delay = Ticks64(23783))
val 25396
int
TestC: Alarm fired.
TestC@Ticks64(25472): Expected at Ticks64(25396) (diff = Ticks64(76)), setting alarm to Ticks64(59825) (delay = Ticks64(34355))
val 35535
int
TestB: Alarm fired.
TestB@Ticks64(35616): Expected at Ticks64(35535) (diff = Ticks64(81)), setting alarm to Ticks64(74719) (delay = Ticks64(39109))
val 41172
int
TestA: Alarm fired.
TestA@Ticks64(41252): Expected at Ticks64(41172) (diff = Ticks64(80)), setting alarm to Ticks64(45476) (delay = Ticks64(4228))
val 45476
int
TestA: Alarm fired.
TestA@Ticks64(45552): Expected at Ticks64(45476) (diff = Ticks64(76)), setting alarm to Ticks64(65103) (delay = Ticks64(19554))
val 59825
int
TestC: Alarm fired.
TestC@Ticks64(59912): Expected at Ticks64(59825) (diff = Ticks64(87)), setting alarm to Ticks64(109592) (delay = Ticks64(49681))
val 65103
int
TestA: Alarm fired.
TestA@Ticks64(65170): Expected at Ticks64(65103) (diff = Ticks64(67)), setting alarm to Ticks64(100048) (delay = Ticks64(34880))
val 74719
int
TestB: Alarm fired.
TestB@Ticks64(74780): Expected at Ticks64(74719) (diff = Ticks64(61)), setting alarm to Ticks64(78829) (delay = Ticks64(4053))
val 78829
int
TestB: Alarm fired.
TestB@Ticks64(78892): Expected at Ticks64(78829) (diff = Ticks64(63)), setting alarm to Ticks64(98268) (delay = Ticks64(19379))
val 98268
int
TestB: Alarm fired.
TestB@Ticks64(98340): Expected at Ticks64(98268) (diff = Ticks64(72)), setting alarm to Ticks64(133043) (delay = Ticks64(34705))
val 100048
int
TestA: Alarm fired.
TestA@Ticks64(100125): Expected at Ticks64(100048) (diff = Ticks64(77)), setting alarm to Ticks64(150330) (delay = Ticks64(50206))
```

Although not sure I can check the box until #2116 is merged since without that the timers don't work.

bradjc (approved):

kernel/src/hil/time.rs
lschuermann: > Can you point me at the sections of the datasheet so I can read and understand better?

Since LiteX is an SoC-Generator, it unfortunately does not provide a "datasheet" in the classical sense.

My intention was not to hijack this PR. Since any potential change won't be included here, I will open an issue later this week where we can discuss whether the current traits can be used for my hardware. I will attach a table of timer registers and explain how the peripheral works.

hudson-ayers (approved):
cc @alistair23 I think the only thing blocking this now is the apollo3 test

alistair23:
It's currently working on Apollo3

chips/apollo3/src/stimer.rs
alistair23: Works

bradjc (approved):

bradjc:
@jrvanwhy @alevy

alexandruradovici:
This code still hangs up `delay_ms(0)` using stm32f4:

```rust
int main(void) {
  putnstr_async(hello, sizeof(hello), nop, NULL);
  delay_ms (0);
  putnstr_async(hello, sizeof(hello), nop, NULL);
  return 0;
}
```

phil-levis:
> ~This code still hangs up when using `delay_ms(0)` using stm32f4:~
> 
> Never mind, I was using the old libtock-c, it seems to work with command_num 6
> 
> ```rust
> int main(void) {
>   putnstr_async(hello, sizeof(hello), nop, NULL);
>   delay_ms (0);
>   putnstr_async(hello, sizeof(hello), nop, NULL);
>   return 0;
> }
> ```

Yes, the next step is to upgrade libtock-c to use the new command.

kernel/src/hil/time.rs
lschuermann: @phil-levis After close inspection of the new HIL, I think that my issues can be worked around. In particular while the traits in question have not changed much, the fact that a `Time` implementation is not semantically equivalent to a `Timestamp` anymore resolves the primary issue of having to implement `Time` for `Timer`. Please excuse my confusion which mostly originated from the legacy (and thankfully) replaced HIL.

Issue #1977, opened 2020-06-24
Tracking: Full Support for Nano33BLE


bradjc:
With Tock running on this board with virtual_uart sitting on top of USB, the Nano 33 BLE is a reasonably promising board to be a standard, well supported Tock board. Additionally, if the board turns out to not be great, the infrastructure which would make the Nano 33 BLE work well would apply to any other board that has the nRF52840 + USB combination.

- [ ] Make the bootloader and tockloader experience "just work". This means `tockloader install` should work on a compiled app.
  - [ ] Replace the bootloader on the board.
    - [ ] Trojan horse kernel that replaces the bootloader.
    - [ ] tock-bootloader ported on top of the USB-CDC-ACM stack.
      - Need some way for the bootloader to detect it should _stay_ in bootloader mode. I'm not sure how Arduino does this.
  - [ ] Need USB serial name that tockloader understands so that it can auto-detect this is a nRF52840 + USB board.
  - [ ] Need the CDC driver to detect when tockloader is trying to connect, and reset the board into the bootloader mode. This should be doable with the baud rate trick that Arduino does.
    - [ ] CDC capsule parses CDC messages looking for a baud rate setting of 1200.
    - [ ] Need a reset() function passed into the CDC capsule. Perhaps with a capability.
- [ ] Make panic debugging work over USB #1972 
- [ ] Implement sensor drivers.
- [ ] Implement BLE.
- [ ] Implement 15.4.

bradjc:
One thought on this topic: proper Tock support really requires breaking backwards compatibility with other nrf52 SDKs and workflows (particularly anything based on the softdevice). That is, by asking users to flash our bootloader, their board will no longer work with any adafruit or arduino instructions. Now we can write a good readme that explains what is happening, and they certainly can run different code on the board of course, but it might not run out of the box anymore.

Do you think this is a concern? Is it worth it to have a mechanism for switching the bootloader back?

alexandruradovici:
I am speaking now from a teaching / beginner point of view and this is only my personal opinion: any action that might "damage" (please see the quotes) the board might be something that might scare people. If the "student" breaks the board and is not able to get it back to its factory state is not something that I as a teacher want to use.

I have been searching for a while for a device that can be used with some scripting language and with Tock. A good candidate has been nano ble 33, as it run CircuitPython [1] until I got to this [2]. I have to replace the boot loader by soldering pins.

People usually want something that can work out of the box and is not easily brick-able.

[...]

PR #2091, opened 2020-09-01
Nano33 apds9960

arjundeopujari:
### Pull Request Overview

Part of #1977

Adds proximity sensor interface and driver for Apds9960 chip for Nano33ble board.
Hil Design:
    - proximity.rs sensor interface:
         read_proximity()
         read_proximity_on_interrupt(low,high)

    - Files added:
         proximity.rs (sensor - userland interface)
         apds9960.rs (IC driver)
  
### Testing Strategy

Tested with corresponding libtock-c app for proximity sensor.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.


kernel/src/hil/sensors.rs
```diff
@@ -31,6 +31,18 @@ pub trait HumidityClient {
     fn callback(&self, value: usize);
 }
 
+/// A basic interface for a proximity sensor
+pub trait ProximityDriver<'a> {
+    fn set_client(&self, client: &'a dyn ProximityClient);
+    fn read_proximity(&self) -> ReturnCode; // Instantaneous proximity reading
```
hudson-ayers, line 37: It is a little misleading to call this instantaneous, as it is still asynchronous (values are returned in a callback), and the time for this to return probably depends on the underlying driver.

kernel/src/hil/sensors.rs
```diff
@@ -31,6 +31,18 @@ pub trait HumidityClient {
...
+    fn read_proximity_on_interrupt(&self, low: u8, high: u8) -> ReturnCode; // Proximity reading returned only after interrupt is detected
```
hudson-ayers, line 38: What are `low` and `high`? What are the units for each? Also, can you specify what type of interrupt this might be waiting for -- at first I was confused because I thought this just meant waiting for the I2C interrupt after the sensor read completes, but it seems it actually refers to returning a reading after some interrupt from the proximity sensor itself.

kernel/src/hil/sensors.rs
```diff
@@ -31,6 +31,18 @@ pub trait HumidityClient {
...
+}
+
+pub trait ProximityClient {
+    /// Called when proximity reading has completed (`command_type` is used by proximity.rs to match callbacks to the apps issuing the appropriate commands)
+    fn callback(&self, value: usize, command_type: usize);
```
hudson-ayers, line 43: What are the units of `value`? What are the possible values of `command_type`? Ideally, the HIL should not be specific to any particular capsule or underlying driver.

capsules/src/proximity.rs
```diff
@@ -0,0 +1,310 @@
...
+        for cntr in self.apps.iter() {
+            cntr.enter(|app, _| {
+                if app.subscribed {
+                    num_commands += 1;
+                }
+            });
+        }
+        if num_commands == 1 {
+            self.run_next_command();
+        }
```
hudson-ayers, line 147: If app1 issues a `ReadProximityOnInterrupt` command, then app2 issues a `ReadProximity` command, doesn't this force the "instantaneous" reading to not occur until after the reading that is waiting on an interrupt occurs?

capsules/src/proximity.rs
```diff
@@ -0,0 +1,310 @@
...
+                    lowest_upper_proximity = app.upper_proximity;
+                }
+            });
+        }
+
+        // return values
+        Thresholds {
+            lower: highest_lower_proximity,
+            upper: lowest_upper_proximity,
+        }
```
hudson-ayers, line 209: Shouldn't this set the thresholds of the underlying driver to be the lowest lower proximity and the highest upper proximity value? 

Ex:
app1: lower_prox=10, higher_prox = 20
app2: lower_prox=5, higher_prox = 25

It seems the current algorithm would set thresholds of 10/20 in this case, thereby missing readings between 5 and 10 and 20/25 that app2 would actually want to detect.

Perhaps I am misunderstanding?

capsules/src/proximity.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    fn call_driver(&self, command: ProximityCommand, arg1: usize, arg2: usize) -> ReturnCode {
+        match command {
+            ProximityCommand::ReadProximity => self.driver.read_proximity(),
+            ProximityCommand::ReadProximityOnInterrupt => self
+                .driver
+                .read_proximity_on_interrupt(arg1 as u8, arg2 as u8),
+            _ => ReturnCode::ENOSUPPORT,
+        }
+    }
```
hudson-ayers, line 220: Having this as its own function seems unnecessary, and passing arg1/arg2 as usizes even though they are really optional arguments is not particularly "rusty". It probably makes more sense to replace calls to this function with the appropriate calls to `self.driver.read_proximity()` or `self.driver.read_proximity_on_interrupt()`

capsules/src/proximity.rs
```diff
@@ -0,0 +1,310 @@
...
+                }
+            }
+
+            command_type if command_type == ProximityCommand::ReadProximityOnInterrupt as usize => {
+                // Schedule callbacks for appropriate apps
+                for cntr in self.apps.iter() {
+                    cntr.enter(|app, _| {
+                        if app.subscribed
+                            && (command_type
+                                == (ProximityCommand::ReadProximityOnInterrupt as usize))
```
hudson-ayers, line 257: If an app has requested an immediate reading, but the next reading to arrive is triggered by a proximity interrupt, might that app not still want to just receive that reading, rather than waiting for a second reading?

capsules/src/driver.rs
```diff
@@ -46,6 +46,7 @@ pub enum NUM {
     Humidity              = 0x60001,
     AmbientLight          = 0x60002,
     NINEDOF               = 0x60004,
+    Proximity             = 0x60003,
```
hudson-ayers, line 49: This syscall value was previously assigned to the pressure sensor HIL, perhaps it would be better to use `0x60005`?
Also, make sure to update `doc/syscalls/README.md` with this value!

boards/nano33ble/src/main.rs
```diff
@@ -227,6 +243,50 @@ pub unsafe fn reset_handler() {
...
+
+    let proximity = static_init!(
+        capsules::proximity::ProximitySensor<'static>,
+        capsules::proximity::ProximitySensor::new(apds9960, board_kernel.create_grant(&grant_cap))
+    );
+
+    kernel::hil::sensors::ProximityDriver::set_client(apds9960, proximity);
+
+    // apds9960.take_measurement();
+
```
hudson-ayers, line 289: remove before merging

boards/nano33ble/src/main.rs
```diff
@@ -273,6 +334,8 @@ pub unsafe fn reset_handler() {
 
     debug!("Initialization complete. Entering main loop.");
 
+    // apds9960.take_measurement();
+
```
hudson-ayers, line 338: remove before merging

hudson-ayers (changes requested):
Thanks for this very nice contribution! Looking forward to having robust sensor support for the nano33!

I reviewed everything except the apds9960 capsule, and have a few comments throughout.

kernel/src/hil/sensors.rs
arjundeopujari: Hi @hudson-ayers, no problem and sorry for the delay!  I have addressed all of your comments but feel free to check if I have accidentally missed one.  I have provided better comments to detail the interface to the user.

kernel/src/hil/sensors.rs
arjundeopujari: I have addressed now in the comments.  `low` and `high` are the interrupt thresholds for the proximity sensor IC.  When 
((reading >= high_thresholds) || (reading <= low_threshold)), an interrupt is fired which alerts the driver to then query the IC for the exact proximity reading which is then returned to the relevant apps in a callback.  Apps can use this when they want to halt the program flow until a certain event (proximity reading crossing these thresholds) occurs.

kernel/src/hil/sensors.rs
arjundeopujari: @bradjc and I have suggested "value" should be unitless as it is impractical to use a formula of some sort to convert the proximity data to inches or millimeters because this reading depends on a multitude of factors (size of object, whether it emits infrared radiation, distance).  Looking around, most proximity sensing IC's use 8 bit registers to store their proximity readings so the range of a possible reading would be in 8 bit resolution (0 - 255) with 255 being the closest measurable distance.  If other drivers and IC's posses different resolutions, I suppose they can do some shifting to get the data into 0 - 255 range.  Ex: if 10-bit resolution then data = data>>2 so 1023 becomes 255.

capsules/src/proximity.rs
arjundeopujari: This is a great observation.  I have decided to put a "ReturnCode::EBUSY" if the currently running command is an interrupt command and an app is issuing a "ReadProximity" command.  I have not found a good solution general to all chips where the interrupt command can be briefly interrupted by a "instantaneous" read command and then resume afterwards.  If you think this is a bad fix then I would be happy to consider your input here.

capsules/src/proximity.rs
arjundeopujari: the sensor is set up in such a way that an interrupt will fire if (reading < 10) or (reading > 20).  If reading == 8 for example, an interrupt will fire and the callback code will alert app1 as it requires this information.  However, the callback code would not alert app2 because its thresholds indicate that it does not require this information.  Lines 288-289 in proximity.rs outline the "callback code" I am referring to.

capsules/src/proximity.rs
arjundeopujari: A good suggestion.  I have addressed this.

capsules/src/proximity.rs
arjundeopujari: I have addressed this.  Another good suggestion.

capsules/src/driver.rs
arjundeopujari: Done and done.

boards/nano33ble/src/main.rs
arjundeopujari: Done.

boards/nano33ble/src/main.rs
arjundeopujari: Done.

kernel/src/hil/sensors.rs
arjundeopujari: "comments" as in comments in the code

kernel/src/hil/sensors.rs
hudson-ayers: That seems reasonable to me, thanks for explaining.

kernel/src/hil/sensors.rs
```diff
@@ -34,12 +34,23 @@ pub trait HumidityClient {
...
+    /// Threshold: A value of range [0 , 255] which represents at what proximity reading ranges an interrupt will occur.
+    fn read_proximity_on_interrupt(&self, low_threshold: u8, high_threshold: u8) -> ReturnCode;
 }
 
 pub trait ProximityClient {
-    /// Called when proximity reading has completed (`command_type` is used by proximity.rs to match callbacks to the apps issuing the appropriate commands)
+    /// Called when a proximity reading has completed.
+    ///
+    /// - `value`: the most recently read proximity value which ranges [0 , 255] where 255 -> object is closest readable distance, 0 -> object is farthest readable distance.
+    /// - 'command_type': callback is being issued after driver serviced this type of command (1 -> read_proximity, 2 -> read_proximity_on_interrupt)
```
hudson-ayers, line 53: If the only valid values are in the range [0, 255], could this just return a u8?

capsules/src/proximity.rs
hudson-ayers: Oh, I see! My mistake.

capsules/src/apds9960.rs
```diff
@@ -0,0 +1,400 @@
...
+            buffer[0] = Registers::ENABLE as u8;
+            buffer[1] = PON | PEN;
+
+            self.i2c.write(buffer, 2);
+
+            self.state.set(State::TakeMeasurement1);
+        });
+    }
+
+    // Take Simple proximity measurement with persistence=4, low threshold=0, high threshold=175, and Sleep-After-Interrupt Mode enabled
```
hudson-ayers, line 200: this comment seems out of date, as the thresholds are now inputs to the function

capsules/src/proximity.rs
hudson-ayers: Hmm, that is a little unfortunate because it means that any app configuring a `ReadProximityWithInterrupt` command effectively takes an indefinite lock on the sensor, which violates the Tock threat model. What do you think of the following approach:

Currently, you acknowledge that multiple `ReadProximityWithInterrupt` commands can be processed simultaneously, and that multiple `ReadProximity` commands can be processed simultaneously. 

If a `ReadProximity` command is issued while a `ReadProximityWithInterrupt` command is "running", the capsule should  change the running command to `ReadProximity` and tell the driver to execute an immediate read instead. The outstanding command(s) will still be enqueued (the apps marked as subscribed), and thus are not forgotten. Also, the callback logic should be changed:

- when a callback arrives for a `ReadProximity` command, still check it against all apps with outstanding `ReadProximityWithInterrupt` commands, and if the proximity value falls within the interrupt-causing range deliver that callback to that client and clear its outstanding command. This helps ensure that proximity readings which arrive during the period when interrupts are disabled still can be delivered to clients that were waiting for interrupts.

Whenever a `ReadProximity` command arrives, it should only be possible for `ReadProximityWithInterrupt` commands to be outstanding, (as any `ReadProximity` callback will remove unsubscribe all `ReadProximity` clients). Also, the driver always handles all `ReadProximityWithInterrupt` commands simultaneously via the call to `find_thresholds()`, so the interrupt detection should only be briefly disabled while the sensor processes the immediate read and services the interrupt. This approach does allow for the small possibility of missed interrupts if the sensor detects the desired proximity value in the time period after the immediate reading is taken and before the capsule has re-enabled interrupts. However this should be quite unlikely, and I think is probably worth making the application interface non-locking.

hudson-ayers (commented):
A few more comments, this is looking close.

capsules/src/proximity.rs
arjundeopujari: This is a pretty good resolution. Ill try the implementation and some testing.

kernel/src/hil/sensors.rs
arjundeopujari: Good point.

capsules/src/proximity.rs
```diff
@@ -249,17 +246,26 @@ impl hil::sensors::ProximityClient for ProximitySensor<'_> {
...
-                // Schedule callbacks for appropriate apps
+                // Schedule callbacks for appropriate apps (any apps waiting for a proximity command)
                 for cntr in self.apps.iter() {
                     cntr.enter(|app, _| {
                         if app.subscribed
-                            && (command_type
-                                == (ProximityCommand::ReadProximityOnInterrupt as usize))
+                            && ((app.enqueued_command_type
+                                == ProximityCommand::ReadProximityOnInterrupt)
+                                || (app.enqueued_command_type == ProximityCommand::ReadProximity))
```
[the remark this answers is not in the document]
arjundeopujari: Another good point.

capsules/src/proximity.rs
arjundeopujari: The callback logic is much cleaner now.  Anytime a reading comes in (no matter for which command it is in response to), the callback logic will check against all outstanding apps and if the app is waiting on an interrupt then it will check against the thresholds.  Any app waiting for a immediate reading or `ReadProximity` should be able to get the command regardless.  Thus, there was really no need for the `command_type` argument in the callback.

I have also implemented the condition in the `enqueue_command` function which checks if the command running is an interrupt command and then substitutes it with the `ReadProximity` command.

kernel/src/hil/sensors.rs
arjundeopujari: I have changed this interface to a u8.

capsules/src/proximity.rs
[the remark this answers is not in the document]
arjundeopujari: I have done this.

capsules/src/proximity.rs
```diff
@@ -0,0 +1,347 @@
...
+    callback: Option<Callback>,
+    subscribed: bool,
+    enqueued_command_type: ProximityCommand,
+    lower_proximity: u8,
+    upper_proximity: u8,
+}
+
+#[derive(Clone, Copy, PartialEq)]
+pub enum ProximityCommand {
+    Exists = 0,
```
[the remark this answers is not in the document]
arjundeopujari: Good suggestion.  I have changed it to a `NoCommand` and removed the `Exists` variant.

capsules/src/proximity.rs
```diff
@@ -0,0 +1,347 @@
...
+    ReadProximity = 1,
+    ReadProximityOnInterrupt = 2,
+    None = 4,
```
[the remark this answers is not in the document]
arjundeopujari: Done.

capsules/src/apds9960.rs
arjundeopujari: Done.

capsules/src/proximity.rs
hudson-ayers: awesome!

hudson-ayers (approved):

bradjc:
It looks like something in upstream tock changed along the way. Look at https://github.com/tock/tock/blob/master/capsules/src/fxos8700cq.rs#L181 for an example how to address the build errors.

capsules/src/apds9960.rs
```diff
@@ -53,7 +53,6 @@ pub static mut BUFFER: [u8; 16] = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
 //                             ^take_meas() callback stored    ^take_meas_int callback stored       ^low thresh           ^high thresh
 
 #[allow(dead_code)]
```
hudson-ayers, line 55: what dead code are we allowing here?

arjundeopujari: I just realized this.  It was leftover from some masks I had before but since deleted.  Ill remove it.

arjundeopujari:
@hudson-ayers @bradjc Sorry for the commotion there.  Ill remember to merge more often for my later works and run the ci jobs locally ahead of time.  It should be good to go here.

hudson-ayers (approved):

bradjc (approved):

ppannuto (approved):
Just looked at HIL, looks good

Issue #1923, opened 2020-06-09
[Tracking] Convert from llvm_asm!() to new asm!()


bradjc:
With a new `asm!()` [macro in Rust nightly](https://blog.rust-lang.org/inside-rust/2020/06/08/new-inline-asm.html) that has a path to stabilization, we should switch our uses of `llvm_asm` to the new format. This is part of #1654.

- [x] arch/cortex-m
- [x] arch/cortex-m0
- [x] arch/cortex-m3
- [x] arch/cortex-m4
- [x] arch/rv32i
- [x] chips/apollo3
- [x] chips/arty_e21_chip
- [x] chips/earlgrey
- [x] chips/stm32f303xc
- [x] chips/stm32f4xx
- [x] libraries (#1926)

gendx:
This is really great given that inline assembly is likely the main reason to use Rust nightly at the moment. And the new macro seems to be getting quite a bit of traction so we can expect it to stabilize in the medium term.

From a programming point of view this is IMO much cleaner than having to link external `.S` files, and if I understood correctly inline `asm!` also allows the compiler to apply more optimizations than with raw `.S` files being linked blindly at the linker step.

alistair23:
https://github.com/tock/tock/pull/1926 is a step in this direction and I think completes the swap for `libraries/riscv-csr``

hudson-ayers:
With the outstanding PRs merged, there are only 4 uses of `llvm_asm` remaining (ignoring all the uses in `cortex-m0`)

lschuermann:
As indicated in #2308, once that change is in, I'll do the migration of the rest of the `rv32i` crate.

bradjc:
Woo!

```
$ git grep llvm_asm | wc -l
       0
```

PR #2092, opened 2020-09-02
arch/cortex-m: change a few `llvm_asm!`s to `asm!`s

JOE1994:
### Pull Request Overview

This PR changes 3 invocations of `llvm_asm!` to `asm!`.
This PR intends to make a small step towards the direction of #1923 .
I made the changes after reading through the [inline-asm RFC](https://github.com/Amanieu/rfcs/blob/inline-asm/text/0000-inline-asm.md#summary) .

##### Brief overview of changes made
1. `llvm_asm!("nop" :::: "volatile");` => `asm!("nop");`
2. `llvm_asm!("wfi" :::: "volatile");` => `asm!("wfi");`

* The only notable change in the first two cases is that the `volatile` keyword is gone. In the new `asm!` macro, `volatile` is assumed by default and there is an option called `pure` which can be used to get rid of the `volatile` semantics.
  * reference: [Link to Q&A on `volatile` from Rust Zulip chat](https://rust-lang.zulipchat.com/#narrow/stream/216763-project-inline-asm/topic/volatile.20question/near/208320361) (requires sign-in)

3. `llvm_asm!("mrs    r0, ipsr": "={r0}"(interrupt_number) : :"r0" :);`
    => `asm!("mrs r0, ipsr", out("r0") interrupt_number);`

* The current `llvm_asm!` invocation specifies `r0` both as output and clobbered register. [It is specified in the LLVM docs that "clobbering named registers that are also present in output constraints is not legal."](http://llvm.org/docs/LangRef.html#clobber-constraints), and it is not possible to do that under the new `asm!` macro syntax. Note the `asm!` macro invocation doesn't have any clobber constraints.

### TODO or Help Wanted

Although the changes made in this PR are quite simple, I'm not sure how to properly test the changes.
I'd appreciate any advice on testing strategy for this PR :smiley_cat: 

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/lib.rs
+++ b/arch/cortex-m/src/lib.rs
@@ -197,6 +197,7 @@ pub unsafe extern "C" fn unhandled_interrupt() {
     asm!(
         "mrs r0, ipsr",
         out("r0") interrupt_number,
+        options(nomem, nostack, preserves_flags)
     );
 
     interrupt_number = interrupt_number & 0x1ff;
--- a/arch/cortex-m/src/lib.rs
+++ b/arch/cortex-m/src/lib.rs
@@ -3,6 +3,7 @@
 #![crate_name = "cortexm"]
 #![crate_type = "rlib"]
 #![feature(llvm_asm)]
+#![feature(asm)]
 #![feature(naked_functions)]
 #![no_std]
 
@@ -193,12 +194,9 @@ pub unsafe extern "C" fn unhandled_interrupt() {
     let mut interrupt_number: u32;
 
     // IPSR[8:0] holds the currently active interrupt
-    llvm_asm!(
-    "mrs    r0, ipsr                    "
-    : "={r0}"(interrupt_number)
-    :
-    : "r0"
-    :
+    asm!(
+        "mrs r0, ipsr",
+        out("r0") interrupt_number,
     );
 
     interrupt_number = interrupt_number & 0x1ff;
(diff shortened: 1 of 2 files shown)


ppannuto:
Following the linked zulip post leads to these docs: https://github.com/rust-lang/rfcs/blob/master/text/2873-inline-asm.md#options-1

For the `nop`, I think we can add `nomem`, `preserves_flags`, and `nostack`.
For the `wfi`, I believe that we can/should also specify `nomem` [We might be able to add `preserves_flags` here too, but probably not `nostack` since the interrupt can/will push to the stack and that might muck with things depending what games the compiler is playing].
For the `mrs`, I think we can add `nomem`, `preserves_flags`, and `nostack`.

---

Regarding testing, I'm not sure there's so much that can be done beyond validating that basic functionality works (i.e. user process with interrupt, like buttons should catch the `wfi` and `mrs`; then any app that panics will exercise `nop`). The additional flags and options come into play when the compiler starts trying to optimize things, which isn't really easy to deterministically test around.

JOE1994:
Thank you for your feedback! I pushed the suggested updates :)

ppannuto (approved):
(Assuming tested on HW) looks good to me!

JOE1994:
I tested the current fork with 4 examples from `libtock-rs`: [`panic`](https://github.com/tock/libtock-rs/blob/master/examples-features/panic.rs), [`blink`](https://github.com/tock/libtock-rs/blob/master/examples/blink.rs), [`button_leds`](https://github.com/tock/libtock-rs/blob/master/examples/button_leds.rs), [`button_subscribe`](https://github.com/tock/libtock-rs/blob/master/examples/button_subscribe.rs).
At least these examples had no issues while running on the modified kernel :)

ppannuto:
bors r+

Sounds good to me, thanks!!

Issue #1940, opened 2020-06-15
Tacking: USB Stack


bradjc:
The USB stack is a work-in-progress. Some identified TODOs:

 - [ ] [This one should be addressed by adding an explicit `panic!`](https://github.com/tock/tock/pull/1902#discussion_r436592017) -- since it will panic either way, better to have something explicit than it happening from an array access.
 - [ ] [This one should be added to a tracking issue so we don't lose it](https://github.com/tock/tock/pull/1902#discussion_r436591970)
 - [ ] [This one could either be addressed by a comment, or added to a tracking issue](https://github.com/tock/tock/pull/1902#discussion_r435800117)

brghena:
Some thoughts I wrote up a couple weeks ago about on a rewrite of `usbc_client_ctrl.rs` and the USB HIL that might be useful for the future.

Goal: control_handler capsule for USB
  * All USB devices must support endpoint zero, which is a control endpoint
  * USB devices do not have multiple control endpoints: https://stackoverflow.com/questions/22417493/what-is-the-use-of-multiple-control-endpoints-non-ep0
  * control_handler should:
	* take a descriptor setup from the higher layer
	* automatically handle control requests from lower layer
	* automatically handle things without an application needing to exist
  * control_handler may:
    * send requests it doesn't know how to handle up the chain (if the higher layer wants them)
  * control_handler should not:
    * get in the way of handling the other (non-control) end points

Questions:
* Should the HIL have two interfaces, one for endpoints and one for control?
  * A good answer seems to me to be have one HIL interface, then one layer on top of it that handles the control endpoint and multiplexes. That layer's interface would be descriptors and endpoints.
* How would a board combine two desired USB capsules: e.g. CDC and HID or similar?
  * Can they be split into configurations?
    * No. The host picks one configuration that the entire device acts as, so a single configuration is best.
  * Ideally, they would be mutiplexed on top of the control layer that lives on top of the HIL. For now, they can be combined into a single capsule manually.
* Do the normal endpoints need to live on top of control, or in parallel?
  * They will need to provide descriptor configuration information to the control capsule.
  * The control capsule is not necessary to be between the HIL and something using bulk, interrupt, or isochronous endpoints.
* Why is the current interface for endpoints on top of buffers of VolatileCells? Can't it just be `[u8]`?
[...]

PR #1467, opened 2019-11-18
Add a nRF52-specific driver for the NVMC

ia0:
### Pull Request Overview

The purpose of this driver is to provide low-level access to the embedded flash of nRF52 boards to allow applications to implement flash-aware (like wear-leveling) data-structures. The driver only permits applications to operate on their writeable flash regions. The API is blocking since the CPU is halted during write and erase operations.

### Testing Strategy

In addition to `make ci-travis`, this pull request was tested on the nrf52840dk board by implementing in libtock-rs:
- a syscall interface
- a multi-set data-structure
- an example application similar to `blink.rs` but with a persistent counter

And updating the linker script.

### TODO or Help Wanted

Which driver number should be used?

The following boards could be tested:
- nRF52833
- nRF52811
- nRF52810

### Documentation Updated

- [x] No updates are required. This is a chip-specific driver, not a generic syscall driver.

### Formatting

- [x] Ran `make formatall`.

PR #2094, opened 2020-09-03
USB: CTAP: Initial support for the FIDO CTAP HID Spec

alistair23:
### Pull Request Overview

This PR adds initial support for the FIDO Alliance CTAP spec, descried here: https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html

This PR adds:
 * USB HID HIL (not CTAP specific)
 * A CTAP capsule
 * A CTAP driver
 * A USB HID CTAP board component

The CTAP capsule has been designed to support BLE, NFC and USB HID transport layers. Currently only USB HID is supported, but in theory BLE and NFC could be fairly easily added. For this reason the CTAP capsule makes USB optional.

### Testing Strategy

Running this PR on OpenTitan with the corresponding [full feature libtock-rs](https://github.com/tock/libtock-rs/pull/239) support.
The [Yubico Python FIDO2](https://github.com/Yubico/python-fido2) library detects OpenTitan as a CTAP device. I can send CBOR/CTAP data to OpenTitan which correctly passes it to the app. The userspace app can decode the CBOR/CTAP data and pass is back to the kernel. This allows me to communicate with OpenTitan over the CTAP protocol from the Python library. I can run their examples.

I am able to register and login to https://github.com/duo-labs/webauthn and https://demo.yubico.com/webauthn with OpenTitan as a security key. NOTE: This doesn't work in Firefox due to https://bugzilla.mozilla.org/show_bug.cgi?id=1530370

@bradjc has tested that the OS X will see the nrf52840dk as a HID device.

In summary, webauthm authentication works on two separate servers when using OpenTitan on a Linux host. OSX with the nrf52840dk is able to detect Tock as a CTAP device.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/nordic/nrf52840dk/src/main.rs
+++ b/boards/nordic/nrf52840dk/src/main.rs
@@ -72,6 +72,8 @@ use kernel::common::dynamic_deferred_call::{DynamicDeferredCall, DynamicDeferred
 use kernel::component::Component;
 use kernel::hil::time::Counter;
 #[allow(unused_imports)]
+use kernel::hil::usb::Client;
+#[allow(unused_imports)]
 use kernel::{capabilities, create_capability, debug, debug_gpio, debug_verbose, static_init};
 use nrf52840::gpio::Pin;
 use nrf52_components::{self, UartChannel, UartPins};
@@ -418,6 +420,38 @@ pub unsafe fn reset_handler() {
     //     components::test::multi_alarm_test::MultiAlarmTestComponent::new(&mux_alarm).finalize(
     //         components::multi_alarm_test_component_buf!(nrf52840::rtc::Rtc),
     //     );
+
+    //--------------------------------------------------------------------------
+    // USB CTAP EXAMPLE
+    //--------------------------------------------------------------------------
+    // Uncomment to experiment with this.
+
+    // // Create the strings we include in the USB descriptor.
+    // let strings = static_init!(
+    //     [&str; 3],
+    //     [
+    //         "Nordic Semiconductor", // Manufacturer
+    //         "nRF52840dk - TockOS",  // Product
+    //         "serial0001",           // Serial number
+    //     ]
+    // );
+
+    // let ctap_send_buffer = static_init!([u8; 64], [0; 64]);
+    // let ctap_recv_buffer = static_init!([u8; 64], [0; 64]);
+
+    // let (ctap, _ctap_driver) = components::ctap::CtapComponent::new(
+    //     &nrf52840::usbd::USBD,
+    //     0x1915, // Nordic Semiconductor
+    //     0x503a, // lowRISC generic FS USB
+    //     strings,
+    //     board_kernel,
+    //     ctap_send_buffer,
+    //     ctap_recv_buffer,
+    // )
+    // .finalize(components::usb_ctap_component_helper!(nrf52840::usbd::Usbd));
+
+    // ctap.enable();
+    // ctap.attach();
 
     let platform = Platform {
         button,
--- a/capsules/src/ctap.rs
+++ b/capsules/src/ctap.rs
@@ -360,6 +360,67 @@ impl<'a, U: usb_hid::UsbHid<'a, [u8; 64]>> Driver for CtapDriver<'a, U> {
                     }
                 })
                 .unwrap_or_else(|err| err.into()),
+            // Send or receive
+            // This command has two parts.
+            //    Part 1: Receive
+            //            This will allow receives, the same as the Allow
+            //            receive command above. If data is ready to receive
+            //            the `packet_received()` callback will be called.
+            //            When this happens the client callback will be
+            //            scheduled and no send event will occur.
+            //    Part 2: Send
+            //            If no receive occurs we will be left in a start where
+            //            future recieves will be allowed. This is the same
+            //            outcome as calling the Allow receive command.
+            //            As well as that we will then send the data in the
+            //            send buffer.
+            4 => self
+                .app
+                .enter(appid, |app, _| {
+                    if let Some(usb) = self.usb {
+                        if app.can_receive.get() {
+                            // We are already receiving
+                            ReturnCode::EBUSY
+                        } else {
+                            app.can_receive.set(true);
+                            if let Some(buf) = self.recv_buffer.take() {
+                                match usb.receive_buffer(buf) {
+                                    Ok(_) => ReturnCode::SUCCESS,
+                                    Err((err, buffer)) => {
+                                        self.recv_buffer.replace(buffer);
+                                        return err;
+                                    }
+                                }
+                            } else {
+                                return ReturnCode::EBUSY;
+                            };
+
+                            if !app.can_receive.get() {
+                                // The call to receive_buffer() collected a pending packet.
+                            } else {
+                                match app.send_buf.as_ref() {
+                                    Some(d) => {
+                                        self.send_buffer.take().map(|buf| {
+                                            let data = d.as_ref();
+
+                                            // Copy the data into the static buffer
+                                            buf.copy_from_slice(&data[0..]);
+
+                                            let _ = usb.send_buffer(buf);
+                                        });
+                                    }
+                                    None => {
+                                        return ReturnCode::ERESERVE;
+                                    }
+                                };
+                            }
+                            ReturnCode::SUCCESS
+                        }
+                    } else {
+                        ReturnCode::ENOSUPPORT
+                    }
+                })
+                .unwrap_or_else(|err| err.into()),
 
             // default
             _ => ReturnCode::ENOSUPPORT,
--- a/capsules/src/ctap.rs
+++ b/capsules/src/ctap.rs
@@ -1,0 +1,368 @@
+//! Provides userspace with access CTAP devices over any transport
+//! layer (USB HID, BLE, NFC). Currently only USB HID is supported.
+//!
+//! Setup
+//! -----
+//!
+//! You need a device that provides the `hil::usb::UsbController` and
+//! `hil::usb_hid::UsbHid` trait.
+//!
+//! ```rust
+//!     let ctap_send_buffer = static_init!([u8; 64], [0; 64]);
+//!     let ctap_recv_buffer = static_init!([u8; 64], [0; 64]);
+//!
+//!     let (ctap, ctap_driver) = components::ctap::CtapComponent::new(
+//!         &earlgrey::usbdev::USB,
+//!         0x1337, // My important company
+//!         0x0DEC, // My device name
+//!         strings,
+//!         board_kernel,
+//!         ctap_send_buffer,
+//!         ctap_recv_buffer,
+//!     )
+//!     .finalize(components::usb_ctap_component_helper!(lowrisc::usbdev::Usb));
+//!
+//!     ctap.enable();
+//!     ctap.attach();
+//! ```
+//!
+
+use core::cell::Cell;
+use core::marker::PhantomData;
+use kernel::common::cells::{OptionalCell, TakeCell};
+use kernel::hil::usb_hid;
+use kernel::{AppId, AppSlice, Callback, Driver, Grant, ReturnCode, Shared};
+
+/// Syscall driver number.
+use crate::driver;
+pub const DRIVER_NUM: usize = driver::NUM::CtapHid as usize;
+
+pub struct App {
+    callback: OptionalCell<Callback>,
+    recv_buf: Option<AppSlice<Shared, u8>>,
+    send_buf: Option<AppSlice<Shared, u8>>,
+    can_receive: Cell<bool>,
+}
+
+impl Default for App {
+    fn default() -> App {
+        App {
+            callback: OptionalCell::empty(),
+            recv_buf: None,
+            send_buf: None,
+            can_receive: Cell::new(false),
+        }
+    }
+}
+
+pub struct CtapDriver<'a, U: usb_hid::UsbHid<'a, [u8; 64]>> {
+    usb: Option<&'a U>,
+
+    app: Grant<App>,
+    appid: OptionalCell<AppId>,
+    phantom: PhantomData<&'a U>,
+
+    send_buffer: TakeCell<'static, [u8; 64]>,
+    recv_buffer: TakeCell<'static, [u8; 64]>,
+}
+
+impl<'a, U: usb_hid::UsbHid<'a, [u8; 64]>> CtapDriver<'a, U> {
+    pub fn new(
+        usb: Option<&'a U>,
+        send_buffer: &'static mut [u8; 64],
+        recv_buffer: &'static mut [u8; 64],
+        grant: Grant<App>,
+    ) -> CtapDriver<'a, U> {
+        CtapDriver {
+            usb: usb,
+            app: grant,
+            appid: OptionalCell::empty(),
+            phantom: PhantomData,
+            send_buffer: TakeCell::new(send_buffer),
+            recv_buffer: TakeCell::new(recv_buffer),
+        }
+    }
+}
+
+impl<'a, U: usb_hid::UsbHid<'a, [u8; 64]>> usb_hid::UsbHid<'a, [u8; 64]> for CtapDriver<'a, U> {
+    fn send_buffer(
+        &'a self,
+        send: &'static mut [u8; 64],
+    ) -> Result<usize, (ReturnCode, &'static mut [u8; 64])> {
+        if let Some(usb) = self.usb {
+            usb.send_buffer(send)
+        } else {
+            Err((ReturnCode::ENOSUPPORT, send))
+        }
+    }
+
+    fn send_cancel(&'a self) -> Result<&'static mut [u8; 64], ReturnCode> {
+        if let Some(usb) = self.usb {
+            usb.send_cancel()
+        } else {
+            Err(ReturnCode::ENOSUPPORT)
+        }
+    }
+
+    fn receive_buffer(
+        &'a self,
+        recv: &'static mut [u8; 64],
+    ) -> Result<(), (ReturnCode, &'static mut [u8; 64])> {
+        if let Some(usb) = self.usb {
+            usb.receive_buffer(recv)
+        } else {
+            Err((ReturnCode::ENODEVICE, recv))
+        }
+    }
+
+    fn receive_cancel(&'a self) -> Result<&'static mut [u8; 64], ReturnCode> {
+        if let Some(usb) = self.usb {
+            usb.receive_cancel()
+        } else {
+            Err(ReturnCode::ENOSUPPORT)
+        }
+    }
+}
+
+impl<'a, U: usb_hid::UsbHid<'a, [u8; 64]>> usb_hid::Client<'a, [u8; 64]> for CtapDriver<'a, U> {
+    fn packet_received(
+        &'a self,
+        _result: ReturnCode,
+        buffer: &'static mut [u8; 64],
+        _endpoint: usize,
+    ) {
+        self.appid.map(|id| {
+            self.app
+                .enter(*id, |app, _| {
+                    match app.recv_buf.as_mut() {
+                        Some(dest) => {
+                            dest.as_mut().copy_from_slice(buffer.as_ref());
+                        }
+                        None => {}
+                    };
+
+                    app.callback.map(|cb| {
+                        cb.schedule(0, 0, 0);
+                        // Set that we can't receive until the app says we can
+                        app.can_receive.set(false);
+                    });
+                })
+                .map_err(|err| {
+                    if err == kernel::procs::Error::NoSuchApp
+                        || err == kernel::procs::Error::InactiveApp
+                    {}
+                })
+        });
+
+        self.recv_buffer.replace(buffer);
+    }
+
+    fn packet_transmitted(
+        &'a self,
+        _result: ReturnCode,
+        buffer: &'static mut [u8; 64],
+        _endpoint: usize,
+    ) {
+        self.appid.map(|id| {
+            self.app
+                .enter(*id, |app, _| {
+                    app.callback.map(|cb| {
+                        cb.schedule(1, 0, 0);
+                    });
+                })
+                .map_err(|err| {
+                    if err == kernel::procs::Error::NoSuchApp
+                        || err == kernel::procs::Error::InactiveApp
+                    {}
+                })
+        });
+
+        // Save our send buffer so we can use it later
+        self.send_buffer.replace(buffer);
+    }
+
+    fn can_receive(&'a self) -> bool {
+        self.appid
+            .map(|id| {
+                self.app
+                    .enter(*id, |app, _| app.can_receive.get())
+                    .unwrap_or(false)
+            })
+            .unwrap_or(false)
+    }
+}
+
+impl<'a, U: usb_hid::UsbHid<'a, [u8; 64]>> Driver for CtapDriver<'a, U> {
+    fn allow(
+        &self,
+        appid: AppId,
+        allow_num: usize,
+        slice: Option<AppSlice<Shared, u8>>,
+    ) -> ReturnCode {
+        match allow_num {
+            // Pass buffer for the recvieved data to be stored in
+            0 => self
+                .app
+                .enter(appid, |app, _| {
+                    app.recv_buf = slice;
+                    ReturnCode::SUCCESS
+                })
+                .unwrap_or(ReturnCode::FAIL),
+
+            // Pass buffer for the sent data to be stored in
+            1 => self
+                .app
+                .enter(appid, |app, _| {
+                    app.send_buf = slice;
+                    ReturnCode::SUCCESS
+                })
+                .unwrap_or(ReturnCode::FAIL),
+
+            // default
+            _ => ReturnCode::ENOSUPPORT,
+        }
+    }
+
+    /// Subscribe to CtapDriver events.
+    ///
+    /// ### `subscribe_num`
+    ///
+    /// - `0`: Subscribe to interrupts from Ctap events.
+    ///        The callback signature is `fn(direction: u32)`
+    ///        `fn(0)` indicates a packet was recieved
+    ///        `fn(1)` indicates a packet was transmitted
+    fn subscribe(
+        &self,
+        subscribe_num: usize,
+        callback: Option<Callback>,
+        appid: AppId,
+    ) -> ReturnCode {
+        match subscribe_num {
+            0 => {
+                // set callback
+                self.app
+                    .enter(appid, |app, _| {
+                        app.callback.insert(callback);
+                        ReturnCode::SUCCESS
+                    })
+                    .unwrap_or(ReturnCode::FAIL)
+            }
+
+            // default
+            _ => ReturnCode::ENOSUPPORT,
+        }
+    }
+
+    fn command(
+        &self,
+        command_num: usize,
+        _data1: usize,
+        _data2: usize,
+        appid: AppId,
+    ) -> ReturnCode {
+        let can_access = self.appid.map_or(true, |owning_app| {
+            if owning_app == &appid {
+                // We own the Ctap device
+                true
+            } else {
+                false
+            }
+        });
+
+        if !can_access {
+            return ReturnCode::EBUSY;
+        }
+
+        match command_num {
+            // Send data
+            0 => self
+                .app
+                .enter(appid, |app, _| {
+                    self.appid.set(appid);
+                    if let Some(usb) = self.usb {
+                        match app.send_buf.as_ref() {
+                            Some(d) => {
+                                self.send_buffer.take().map(|buf| {
+                                    let data = d.as_ref();
+
+                                    // Copy the data into the static buffer
+                                    buf.copy_from_slice(&data[0..]);
+
+                                    let _ = usb.send_buffer(buf);
+                                });
+                            }
+                            None => {
+                                return ReturnCode::ERESERVE;
+                            }
+                        };
+                        ReturnCode::SUCCESS
+                    } else {
+                        ReturnCode::ENOSUPPORT
+                    }
+                })
+                .unwrap_or_else(|err| err.into()),
+            // Allow receive
+            1 => self
+                .app
+                .enter(appid, |app, _| {
+                    self.appid.set(appid);
+                    if let Some(usb) = self.usb {
+                        app.can_receive.set(true);
+                        if let Some(buf) = self.recv_buffer.take() {
+                            match usb.receive_buffer(buf) {
+                                Ok(_) => ReturnCode::SUCCESS,
+                                Err((err, buffer)) => {
+                                    self.recv_buffer.replace(buffer);
+                                    err
+                                }
+                            }
+                        } else {
+                            ReturnCode::EBUSY
+                        }
+                    } else {
+                        ReturnCode::ENOSUPPORT
+                    }
+                })
+                .unwrap_or_else(|err| err.into()),
+            // Cancel send
+            2 => self
+                .app
+                .enter(appid, |_app, _| {
+                    self.appid.set(appid);
+                    if let Some(usb) = self.usb {
+                        match usb.receive_cancel() {
+                            Ok(buf) => {
+                                self.recv_buffer.replace(buf);
+                                ReturnCode::ENOSUPPORT
+                            }
+                            Err(err) => err,
+                        }
+                    } else {
+                        ReturnCode::ENOSUPPORT
+                    }
+                })
+                .unwrap_or_else(|err| err.into()),
+            // Cancel receive
+            3 => self
+                .app
+                .enter(appid, |_app, _| {
+                    self.appid.set(appid);
+                    if let Some(usb) = self.usb {
+                        match usb.receive_cancel() {
+                            Ok(buf) => {
+                                self.recv_buffer.replace(buf);
+                                ReturnCode::ENOSUPPORT
+                            }
+                            Err(err) => err,
+                        }
+                    } else {
+                        ReturnCode::ENOSUPPORT
+                    }
+                })
+                .unwrap_or_else(|err| err.into()),
+
+            // default
+            _ => ReturnCode::ENOSUPPORT,
+        }
+    }
+}
--- a/capsules/src/driver.rs
+++ b/capsules/src/driver.rs
@@ -36,6 +36,7 @@ pub enum NUM {
     Rng                   = 0x40001,
     Crc                   = 0x40002,
     Hmac                  = 0x40003,
+    CtapHid               = 0x40004,
 
     // Storage
     AppFlash              = 0x50000,
--- a/capsules/src/usb/ctap.rs
+++ b/capsules/src/usb/ctap.rs
@@ -1,0 +1,412 @@
+//! Client to Authenticator Protocol CTAPv2 over USB HID
+//!
+//! Based on the spec avaliable at: https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html
+
+use core::cell::Cell;
+use core::cmp;
+
+use super::descriptors;
+use super::descriptors::Buffer64;
+use super::descriptors::DescriptorType;
+use super::descriptors::EndpointAddress;
+use super::descriptors::EndpointDescriptor;
+use super::descriptors::HIDCountryCode;
+use super::descriptors::HIDDescriptor;
+use super::descriptors::HIDSubordinateDescriptor;
+use super::descriptors::InterfaceDescriptor;
+use super::descriptors::ReportDescriptor;
+use super::descriptors::TransferDirection;
+use super::usbc_client_ctrl::ClientCtrl;
+
+use kernel::common::cells::OptionalCell;
+use kernel::common::cells::TakeCell;
+use kernel::common::cells::VolatileCell;
+use kernel::hil;
+use kernel::hil::usb::TransferType;
+use kernel::ReturnCode;
+
+/// The spec defines 1 Interrupt transfer in endpoint
+const ENDPOINT_IN_NUM: usize = 2;
+/// The spec defines 1 Interrupt transfer out endpoint
+const ENDPOINT_OUT_NUM: usize = 1;
+
+static LANGUAGES: &'static [u16; 1] = &[
+    0x0409, // English (United States)
+];
+/// Max packet size specified by spec
+pub const MAX_CTRL_PACKET_SIZE: u8 = 64;
+
+const N_ENDPOINTS: usize = 2;
+
+/// The HID report descriptor for CTAP
+/// This is a combinfrom of:
+///     - the CTAP spec, example 8
+///     - USB HID spec examples
+/// Plus it matches: https://chromium.googlesource.com/chromiumos/platform2/+/master/u2fd/u2fhid.cc
+static REPORT_DESCRIPTOR: &'static [u8] = &[
+    0x06, 0xD0, 0xF1, // HID_UsagePage ( FIDO_USAGE_PAGE ),
+    0x09, 0x01, // HID_Usage ( FIDO_USAGE_CTAPHID ),
+    0xA1, 0x01, // HID_Collection ( HID_Application ),
+    0x09, 0x20, // HID_Usage ( FIDO_USAGE_DATA_IN ),
+    0x15, 0x00, // HID_LogicalMin ( 0 ),
+    0x26, 0xFF, 0x00, // HID_LogicalMaxS ( 0xff ),
+    0x75, 0x08, // HID_ReportSize ( 8 ),
+    0x95, 0x40, // HID_ReportCount ( HID_INPUT_REPORT_BYTES ),
+    0x81, 0x02, // HID_Input ( HID_Data | HID_Absolute | HID_Variable ),
+    0x09, 0x21, // HID_Usage ( FIDO_USAGE_DATA_OUT ),
+    0x15, 0x00, // HID_LogicalMin ( 0 ),
+    0x26, 0xFF, 0x00, // HID_LogicalMaxS ( 0xff ),
+    0x75, 0x08, // HID_ReportSize ( 8 ),
+    0x95, 0x40, // HID_ReportCount ( HID_OUTPUT_REPORT_BYTES ),
+    0x91, 0x02, // HID_Output ( HID_Data | HID_Absolute | HID_Variable ),
+    0xC0, // HID_EndCollection
+];
+
+static REPORT: ReportDescriptor<'static> = ReportDescriptor {
+    desc: REPORT_DESCRIPTOR,
+};
+
+static SUB_HID_DESCRIPTOR: &'static [HIDSubordinateDescriptor] = &[HIDSubordinateDescriptor {
+    typ: DescriptorType::Report,
+    len: REPORT_DESCRIPTOR.len() as u16,
+}];
+
+static HID_DESCRIPTOR: HIDDescriptor<'static> = HIDDescriptor {
+    hid_class: 0x0110,
+    country_code: HIDCountryCode::NotSupported,
+    sub_descriptors: SUB_HID_DESCRIPTOR,
+};
+
+/// Implementation of the CTAP HID (Human Interface Device)
+pub struct CtapHid<'a, U: 'a> {
+    /// Helper USB client library for handling many USB operations.
+    client_ctrl: ClientCtrl<'a, 'static, U>,
+
+    /// 64 byte buffers for each endpoint.
+    buffers: [Buffer64; N_ENDPOINTS],
+
+    client: OptionalCell<&'a dyn hil::usb_hid::Client<'a, [u8; 64]>>,
+
+    /// A buffer to hold the data we want to send
+    send_buffer: TakeCell<'static, [u8; 64]>,
+
+    /// A holder for the buffer to receive bytes into. We use this as a flag as
+    /// well, if we have a buffer then we are actively doing a receive.
+    recv_buffer: TakeCell<'static, [u8; 64]>,
+    /// How many bytes the client wants us to receive.
+    recv_len: Cell<usize>,
+    /// How many bytes we have received so far.
+    recv_offset: Cell<usize>,
+
+    saved_endpoint: OptionalCell<usize>,
+}
+
+impl<'a, U: hil::usb::UsbController<'a>> CtapHid<'a, U> {
+    pub fn new(
+        controller: &'a U,
+        vendor_id: u16,
+        product_id: u16,
+        strings: &'static [&'static str; 3],
+    ) -> Self {
+        let interfaces: &mut [InterfaceDescriptor] = &mut [InterfaceDescriptor {
+            interface_number: 0,
+            interface_class: 0x03,    // HID
+            interface_subclass: 0x00, // No subcall
+            interface_protocol: 0x00, // No protocol
+            ..InterfaceDescriptor::default()
+        }];
+
+        let endpoints: &[&[EndpointDescriptor]] = &[&[
+            EndpointDescriptor {
+                endpoint_address: EndpointAddress::new_const(0x02, TransferDirection::DeviceToHost),
+                transfer_type: TransferType::Interrupt,
+                max_packet_size: 64,
+                interval: 5,
+            },
+            EndpointDescriptor {
+                endpoint_address: EndpointAddress::new_const(0x01, TransferDirection::HostToDevice),
+                transfer_type: TransferType::Interrupt,
+                max_packet_size: 64,
+                interval: 5,
+            },
+        ]];
+
+        let (device_descriptor_buffer, other_descriptor_buffer) =
+            descriptors::create_descriptor_buffers(
+                descriptors::DeviceDescriptor {
+                    vendor_id: vendor_id,
+                    product_id: product_id,
+                    manufacturer_string: 1,
+                    product_string: 2,
+                    serial_number_string: 3,
+                    class: 0x03, // Class: HID
+                    max_packet_size_ep0: MAX_CTRL_PACKET_SIZE,
+                    ..descriptors::DeviceDescriptor::default()
+                },
+                descriptors::ConfigurationDescriptor {
+                    ..descriptors::ConfigurationDescriptor::default()
+                },
+                interfaces,
+                endpoints,
+                Some(&HID_DESCRIPTOR),
+                None,
+            );
+
+        CtapHid {
+            client_ctrl: ClientCtrl::new(
+                controller,
+                device_descriptor_buffer,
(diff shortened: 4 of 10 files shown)


bradjc (approved):

capsules/src/usb/ctap.rs
```diff
@@ -0,0 +1,391 @@
...
+    /// The RX client to use when RX data is received.
+    rx_client: OptionalCell<&'a dyn uart::ReceiveClient>,
+}
+
+impl<'a, U: hil::usb::UsbController<'a>> CtapHid<'a, U> {
+    pub fn new(
+        controller: &'a U,
+        vendor_id: u16,
+        product_id: u16,
+        strings: &'static [&'static str; 3],
```
phil-levis, line 115: These should be typed (e.g., pass a struct that has three strings as named fields).

capsules/src/usb/ctap.rs
```diff
@@ -0,0 +1,391 @@
...
+                unimplemented!();
+            }
+        }
+    }
+
+    /// Handle a Bulk/Interrupt OUT transaction
+    fn packet_out(
+        &'a self,
+        transfer_type: TransferType,
+        endpoint: usize,
```
phil-levis, line 319: Can endpoints be an enum? Endpoint 0xffffffff seems unlikely.

phil-levis (changes requested):
Generally, endpoints should be a type-checked enum, not a usize (if this is feasible).

capsules/src/usb/ctap.rs
alistair23: That is a large change. This function comes from the HIL so the USB HIL would need to be changed as well as everything that uses the HIL. Then the USB packet decoded to a enum (not that hard).

What advantage would that get us?

capsules/src/usb/ctap.rs
alistair23: Then this would differ from the CDC device and the ClientCtrl::new() arguments.

gendx:
FYI, this overlaps with what we've done in OpenSK to support HID for CTAP2: https://github.com/google/OpenSK/blob/master/patches/tock/02-usb.patch

Given that we've already tested it quite a bit in the context of OpenSK, wouldn't it make sense to start from that patch instead? There may be a couple things to synchronize w.r.t. the latest changes done in Tock's USB HIL, but I'd be happy to propose a pull request for that.

I see some differences here, and am worried that this pull request hasn't been tested as much as our OpenSK work.

capsules/src/usb/ctap.rs
```diff
@@ -0,0 +1,391 @@
...
+use super::descriptors::InterfaceDescriptor;
+use super::descriptors::ReportDescriptor;
+use super::descriptors::TransferDirection;
+use super::usbc_client_ctrl::ClientCtrl;
+
+use kernel::common::cells::OptionalCell;
+use kernel::common::cells::TakeCell;
+use kernel::common::cells::VolatileCell;
+use kernel::hil;
+use kernel::hil::uart;
```
gendx, line 25: Why using the UART HIL?

capsules/src/usb/ctap.rs
```diff
@@ -0,0 +1,391 @@
...
+        // Setup buffers for IN and OUT data transfer.
+        self.controller()
+            .endpoint_set_in_buffer(ENDPOINT_IN_NUM, self.buffer(ENDPOINT_IN_NUM));
+        self.controller()
+            .endpoint_in_enable(TransferType::Bulk, ENDPOINT_IN_NUM);
+
+        self.controller()
+            .endpoint_set_out_buffer(ENDPOINT_OUT_NUM, self.buffer(ENDPOINT_OUT_NUM));
+        self.controller()
+            .endpoint_out_enable(TransferType::Bulk, ENDPOINT_OUT_NUM);
```
gendx, line 208: This section is misleading, because in reality `ENDPOINT_IN_NUM == ENDPOINT_OUT_NUM`. In that case, the correct way to call the HIL is to use `endpoint_in_out_enable`.

Also, at least for the OpenSK case, this should be an Interrupt transfer, not Bulk.

That's the kind of things I'm worried about in terms of the lack of robustness of this pull request.

capsules/src/usb/ctap.rs
```diff
@@ -0,0 +1,391 @@
...
+
+    /// Handle a Bulk/Interrupt IN transaction.
+    ///
+    /// This is called when we can send data to the host. It should get called
+    /// when we tell the controller we want to resume the IN endpoint (meaning
+    /// we know we have data to send) and afterwards until we return
+    /// `hil::usb::InResult::Delay` from this function. That means we can use
+    /// this as a callback to mean that the transmission finished by waiting
+    /// until this function is called when we don't have anything left to send.
+    fn packet_in(&'a self, transfer_type: TransferType, endpoint: usize) -> hil::usb::InResult {
```
gendx, line 258: This function looks much more complicated than what we have in https://github.com/google/OpenSK/blob/master/patches/tock/02-usb.patch. Same for `packet_out`.

gendx (changes requested):
> ### Testing Strategy
> Running this on OpenTitan, the Yubico Python FIDO2 https://github.com/Yubico/python-fido2/ library detects OpenTitan as a CTAP device.

I'm worried that "detects the device" is not enough in-depth testing, as opposed to the extensive testing we've put in https://github.com/google/OpenSK/blob/master/patches/tock/02-usb.patch.

I'd however be happy to propose a pull request from https://github.com/google/OpenSK/blob/master/patches/tock/02-usb.patch, and see whether it works on OpenTitan or if adaptations are needed.

alistair23:
I thought OpenSK had CTAP support, but when I had a quick look I couldn't find a Tock repo. I didn't realise it was stored in a single patch.

This is just the very start of the code I have for CTAP support. I currently have a HIL and driver working with userspace apps getting the CBOR/CTAP data over USB. As USB is generally unstable I wanted to send the patches as I go so that the overall diff is pretty small. I was hoping to get this one merged first before sending the others, but I can just send them instead.

I would prefer to use this version. It's based on the latest CDC support so it matches the other USB code in Tock. it also seems a little strange that the OpenSK work couldn't be upsteamed until someone else comes along with the same features.

capsules/src/usb/ctap.rs
alistair23: Removed. That was just leftover.

capsules/src/usb/ctap.rs
alistair23: Fixed.

capsules/src/usb/ctap.rs
alistair23: Fixed.

alistair23:
I pushed an update that adds a HIL, capsule and component. This works with userspace apps, although I'm unable to test sending as the OT USB stack doesn't support it yet.

gendx:
> This is just the very start of the code I have for CTAP support. I currently have a HIL and driver working with userspace apps getting the CBOR/CTAP data over USB. As USB is generally unstable I wanted to send the patches as I go so that the overall diff is pretty small. I was hoping to get this one merged first before sending the others, but I can just send them instead.
> 
> I would prefer to use this version. It's based on the latest CDC support so it matches the other USB code in Tock.

On the one hand, you mention that it's quite experimental, lacks a lot of testing, etc., yet on the other hand you want to push for this version.

To be honest, given that we now have USB support for multiple boards, and for multiple protocols with now competing proposed implementations, I think it's time to revisit the "USB is generally unstable" statement, and take time to figure out a design that works for all of these use cases, rather than pushing in a random direction that will then require a lot of redesign and refactoring to support the other use cases. Otherwise, it's just a recipe for getting forks of Tock...

> This works with userspace apps, although I'm unable to test sending as the OT USB stack doesn't support it yet.

This is what worries me about this implementation. The OpenSK implementation was tested with both sending and receiving, including a "cancellation" operation.

> It also seems a little strange that the OpenSK work couldn't be upsteamed until someone else comes along with the same features.

There had been some push-back on other pull-requests that were quite OpenSK-specific (e.g. #1467), and at the time it seemed that CTAP was quite a niche so I was afraid it would face similar push back. Preparing a pull request is quite some work, so given the reviewing costs there were no clear benefits to proposing it at the time. There was also not much interest for USB in Tock at the time - OpenTitan and CDC support came much later.

But now the question is asked in #2099.

capsules/src/ctap.rs
```diff
@@ -0,0 +1,249 @@
...
+            send_buf: None,
+        }
+    }
+}
+
+pub static mut WRITE_BUF: [u8; 64] = [0; 64];
+pub static mut READ_BUF: [u8; 64] = [0; 64];
+
+pub struct CtapDriver<'a, U: usb_hid::UsbHid<'a>> {
+    usb: Option<&'a U>,
```
bradjc, line 59: Rather than an option, why not have a `CtapTransport` trait?

bradjc (commented):
Maybe one direction forward is looking at what interfaces we want between layers, particularly since we have multiple implementations.

Looking through this PR, it seems the largest stumbling block is we don't have a "plugable" architecture for setting up USB descriptors and endpoints. I'm assuming that we could have a cleaner separation between USB HID and USB HID-CTAP if we could augment the USB HID descriptors with the CTAP-specific setup somehow, then the capsules/usb/ctap file would be more ctap specific, I think. Is that something we could/should add?

capsules/src/ctap.rs
alistair23: I have updated the PR description to clarify this. The idea is that the CTAP capsule can support USB HID, BLE or NFC. USB is an option here to allow extending this to BLE or NFC in the future.

alistair23:
> Looking through this PR, it seems the largest stumbling block is we don't have a "plugable" architecture for setting up USB descriptors and endpoints. I'm assuming that we could have a cleaner separation between USB HID and USB HID-CTAP if we could augment the USB HID descriptors with the CTAP-specific setup somehow, then the capsules/usb/ctap file would be more ctap specific, I think. Is that something we could/should add?

I'm not fully clear on what you mean. I have tried to keep CTAP and USB/CTAP separate as CTAP supports multiple transport layers.

It would be nice to split up the USB capsules a little better so we could just have a HID capsule and the CTAP one could call that, passing in it's report descriptor. That should probably wait until we support multiple configuration descriptors and virtual USB support though.

Something I want to work on after CTAP and CDC are working on OpenTitan is supporting both at the same time. I think at that point we can refactor some of this to be more modular. Maybe even just support each endpoint type and use that as a building block for HID/CDC support? I think all of that should come after this PR (and after general CTAP support).

capsules/src/usb/ctap.rs
alistair23: I'm going to mark this as resolved. As part of the general USB TODO (https://github.com/tock/tock/issues/1940) we want to convert USB endpoints to be enums. That needs to happen in the HIL though.

alistair23:
Rebased on master so now the OpenTitan tests are passing!

bradjc (dismissed):
I added a commit which adds commented out code to getting this running on the nrf52840dk. I tried it as well, and my mac at least recognized the usb device, which is at least a good sign.

Why is one of the USB components OT only?

Overall I think this is progress in the right direction, particularly since most of the actual CTAP code will be in userspace with this design.

alistair23:
Awesome! Thanks for testing it Brad. As I mentioned today without a fully CTAP app that is all that will work.

From memory the USB component is OT only as whatever I based it on was also platform specific. It was just easier at the time, that should be fixed.

bradjc:
@phil-levis

capsules/src/usb/ctap.rs
```diff
@@ -0,0 +1,397 @@
...
+    }
+
+    fn bus_reset(&'a self) {}
+
+    /// Handle a Control Setup transaction.
+    fn ctrl_setup(&'a self, endpoint: usize) -> hil::usb::CtrlSetupResult {
+        descriptors::SetupData::get(&self.client_ctrl.ctrl_buffer.buf).map(|setup_data| {
+            let _b_request = setup_data.request_code;
+            let _value = setup_data.value;
+        });
```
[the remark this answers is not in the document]
alistair23: Nope. Removed.

kernel/src/hil/usb_hid.rs
```diff
@@ -0,0 +1,39 @@
...
+        len: usize,
+        endpoint: usize,
+    );
+
+    /// Called when checking if we can receive any more data.
+    /// Should return true if we are ready to receive.
+    fn can_receive(&'a self) -> bool;
+}
+
+pub trait UsbHid<'a> {
```
phil-levis, line 24: Unlike in the UART, I think it makes sense to have receive and transmit in the same trait here, because for USB the two are coupled, and for a particular endpoint we won't virtualize; virtualization would occur for higher layer abstractions, not at this level.

That being said, I think that these send/receive APIs should follow the same structure as the UART ones. The call to receive should pass the buffer. With this API, it's possible to lose a buffer, e.g., if there already is one and then `set_recv_buffer` is called again. Also, there are no error conditions returned. You need a way to cancel/halt receptions, rather than enabling them. So it should be `receive` and `cancel`, not `set_recv_buffer` and `allow_receive`. These APIs should clearly describe what the ownership assumptions are.

Why does `packet_transmitted` return the endpoint? It's not passed in `send_buffer`.

phil-levis (changes requested):
Overall, getting this code in is good, for when a working USB is merged with mainline OpenTitan.

My one set of concerns relates to the HIL; as we're seeing with Alarm, buggy HILs can be very difficult to unwind. The HIL should follow the calling conventions and designs we see in the rest of the kernel (e.g., with buffer management) unless there's a reason why they can't, and if they can't, that should be a general point rather than a hardware-specific one. Otherwise the HIL happens to fit the first hardware and then runs into problem with later ones (an issue we saw with bit widths of counters with Alarm). Each call should return a ReturnCode, with the possible errors being clearly enumerated.

kernel/src/hil/usb_hid.rs
alistair23: I have converted this so that receive now takes a buffer (`receive_buffer()`) and there is a cancel call in the HIL (`receive_cancel()`).

I have added more documentation in the HIL as well.

`packet_transmitted()` returns the endpoint so that users know what endpoint the data was transfered over if there is more then one. It's not currently used, but it seems like it might be useful for more complex HID devices in the future.

jmichelp: Looking at the implementation that @gendx did for OpenSK, it seems we even needed to be able to cancel both IN and OUT transactions for CTAP2.

capsules/src/usb/ctap.rs
```diff
@@ -0,0 +1,401 @@
...
+
+        self.send_buffer.replace(send);
+        self.controller().endpoint_resume_in(ENDPOINT_IN_NUM);
+
+        Ok(len)
+    }
+
+    fn receive_buffer(
+        &'a self,
+        recv: &'static mut [u8; 64],
```
jmichelp, line 211: Why would the buffers need to be with a static lifetime? Also applies to `send_buffer` above.

kernel/src/hil/usb_hid.rs
alistair23: How would you cancel a send from Tock? If the transaction has already started you have already sent the length on the USB bus, you don't then want to cancel mid transaction.

The send command only sends a single buffer, I don't see a use case for when you would want to cancel that.

capsules/src/usb/ctap.rs
```diff
@@ -0,0 +1,401 @@
...
+
+use kernel::common::cells::OptionalCell;
+use kernel::common::cells::TakeCell;
+use kernel::common::cells::VolatileCell;
+use kernel::hil;
+use kernel::hil::usb::TransferType;
+use kernel::ReturnCode;
+
+/// The spec defines 1 Interrupt transfer in endpoint
+const ENDPOINT_IN_NUM: usize = 2;
```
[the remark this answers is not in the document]
alistair23: I don't understand, the IN and OUT endpoints have different addresses: https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html#usb-descriptors

Endpoint 1 descriptor
bEndpointAdresss | 0x01 | 1, OUT

Endpoint 2 descriptor
bEndpointAdresss | 0x81 | 1, IN

capsules/src/usb/ctap.rs
```diff
@@ -0,0 +1,401 @@
...
+            interface_number: 0,
+            interface_class: 0x03,    // HID
+            interface_subclass: 0x00, // No subcall
+            interface_protocol: 0x00, // No protocol
+            ..InterfaceDescriptor::default()
+        }];
+
+        let endpoints: &[&[EndpointDescriptor]] = &[&[
+            EndpointDescriptor {
+                endpoint_address: EndpointAddress::new_const(0x02, TransferDirection::DeviceToHost),
```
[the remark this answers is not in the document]
alistair23: This is the endpoint number internally to Tock, not the address.

capsules/src/usb/ctap.rs
```diff
@@ -0,0 +1,401 @@
...
+                    product_id: product_id,
+                    manufacturer_string: 1,
+                    product_string: 2,
+                    serial_number_string: 3,
+                    class: 0x03, // Class: HID
+                    max_packet_size_ep0: MAX_CTRL_PACKET_SIZE,
+                    ..descriptors::DeviceDescriptor::default()
+                },
+                descriptors::ConfigurationDescriptor {
+                    ..descriptors::ConfigurationDescriptor::default()
```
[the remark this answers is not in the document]
alistair23: I don't see any issues with dmesg. I would like to leave this as is unless there is a good reason not to.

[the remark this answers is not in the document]
alistair23: I don't see any issues with dmesg. I would like to leave this as is unless there is a good reason not to.

capsules/src/usb/ctap.rs
```diff
@@ -0,0 +1,401 @@
...
+
+impl<'a, U: hil::usb::UsbController<'a>> hil::usb::Client<'a> for CtapHid<'a, U> {
+    fn enable(&'a self) {
+        // Set up the default control endpoint
+        self.client_ctrl.enable();
+
+        // Setup buffers for IN and OUT data transfer.
+        self.controller()
+            .endpoint_set_in_buffer(ENDPOINT_IN_NUM, self.buffer(ENDPOINT_IN_NUM));
+        self.controller()
```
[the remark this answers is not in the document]
alistair23: That's not the case though. I don't think that is even allowed in the USB spec.

The CTAP spec specifies two endpoints: https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html#usb-descriptors

capsules/src/usb/ctap.rs
alistair23: This is what the other Tock HILs do as well.

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
bradjc: Yes using 0 is an issue as it is not valid and mac enforces that. However we changed the default to 1 in descriptors.rs so this shouldn't be a problem now.

kernel/src/hil/usb_hid.rs
jmichelp: As a USB peripheral you're not really deciding when you'll send the buffer. You can only enqueue it to be sent on the next transaction decided by the host, right? So conditions may occur when your enqueued buffer for transmission is not valid anymore because the host sent you another packet.
At least that would be my guess but Guillaume would know better than me here.

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
jmichelp: The endpoint number is encoded on 7 bits. The MSB is used to encode the direction (IN vs. OUT). So it's effectively endpoint 1 being used in both direction.

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
jmichelp: See comment above, it's perfectly valid because the endpoint number is using 7 bits and the MSB is used to distinguish the direction.

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
alistair23: But the spec still specifies two and we follow that. This is clearer to readers what we are doing. Having a single endpoint seems unclear.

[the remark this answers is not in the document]
gendx: The problem is that the USB HIL has a special case for an endpoint enabled in both directions - `endpoint_in_out_enable(endpoint)` - which is NOT equivalent to `endpoint_in_enable(endpoint)` followed by `endpoint_out_enable(endpoint)`.

kernel/src/hil/usb_hid.rs
alistair23: That's true, but the USB implementation only has a single send buffer so in practice this gets send straight away.

I also don't really see a use case where the data would become invalid that quickly. This is a HID style device so the small payload is sent very quickly.

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
alistair23: The spec says "The device implements two endpoints (except the control endpoint 0), one for IN and one for OUT transfers." which is what we are doing here. I don't see any mention of the same endpoint enabled in both directions.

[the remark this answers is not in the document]
gendx: It's written in the table in the "bEndpointAdresss" field. Both endpoints use the endpoint 1, each in a different direction.

> Endpoint 1 descriptor
> ...
> bEndpointAdresss | 0x01 | 1, OUT
> ...
> Endpoint 2 descriptor
> ...
> bEndpointAdresss | 0x81 | 1, IN

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
gendx: > That's not the case though. I don't think that is even allowed in the USB spec.

That's what we've implemented in OpenSK, and it works. Do you have an actual reference from the USB spec that forbids it?

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
alistair23: They list two different endpoint numbers. I don't understand how that becomes a single endpoint.

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
alistair23: I have implemented two here and it also works.

Apparently a IN and OUT endpoint is allowed, sorry about that. In this case thought the CTAP spec still specifies two.

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
jmichelp: No it's not. The first argument is going to be used to compute the descriptor and the CTAP specification expects `0x81` (i.e. endpoint number 1 + `TransferDirection::DeviceToHost`) and `0x01` (i.e. endpoint number 1 + `TransferDirection::HostToDevice`) to be used.
See https://master--docs-tockosorg.netlify.app/src/capsules/usb/descriptors.rs.html#662-666

kernel/src/hil/usb_hid.rs
gendx: In USB, nothing gets sent until the host decided to initiate an OUT transaction (for which we receive interrupts). One important scenario is to send a message fragmented into multiple packets to the host: for each packet, we have to wait for the host to initiate the transaction (first interrupt) and to complete it (second interrupt), before we move on to the next packet. Otherwise, we may end up writing the second packet while the first one is still in progress, i.e. we may end up overwriting a DMA buffer while it's still being read. And this can happen however small the send buffer and "straight away" data gets sent.

The issue is if the host doesn't initiate any OUT transaction. This can happen realistically in practice:
- The host sends a "cancel" message ahead of time rather than expecting a response. This happens when the user presses "cancel" in the Chrome browser popup when prompted for a touch on the security key for example.
- The host application crashed/terminated.

Without support for cancellation, we're stuck in these realistic scenarios:
- If we wait for interrupts, we can wait forever because no host application sends us an OUT transaction.
- If we don't wait for interrupts, and hope that it gets sent "straight away" we DO overwrite DMA buffers in practice.

I don't know the specifics of OpenTitan chip w.r.t. this scenarios, but given how the nRF52840 interrupts are set up, that's something we cannot ignore.

Another important scenario is to be able to handle both send and receive concurrently. Because as a Tock device we can expect to send a packet (OUT transaction), but actually the host decided to send us a packet instead - typically a "cancel" message in CTAP (IN transaction). Because USB is driven by the host, we cannot completely decouple sending and receiving, nor expect either sending or receiving to complete quickly (nor to ever complete).

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
gendx: In both cases, the USB endpoint address is "1". At the hardware level, this corresponds to a specific set of registers.

Have you actually tested this pull request with these two endpoints ("1" and "2") with a real host application? I don't mean that such a configuration would never work, but it's not what is described in the spec's example.

---

The pull request description mentions:

> Running this on OpenTitan, the Yubico Python FIDO2 https://github.com/Yubico/python-fido2/ library detects OpenTitan as a CTAP device.

I assume this is the control setup.

> I can send CBOR/CTAP data to OpenTitan which correctly passes it to an app. The userspace app can decode the CBOR/CTAP data and pass is back to the kernel.

I assume this is the IN endpoint.

> This allows me to communicate with OpenTitan over the CTAP protocol.

I don't see any evidence of testing the OUT endpoint.

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
gendx: I assumed that the specs requires one to use endpoint 1 in both modes, but indeed maybe it would work to pick (2, IN) and (5, OUT) if one wants to. However, I haven't tested any of that on a host application, so maybe it wouldn't be compatible with all host applications.

kernel/src/hil/usb_hid.rs
alistair23: The buffers should not be overwritten as the data should not be changed until the `packet_transmitted()` callback is called, no matter how long it takes for the transfer.

So you are saying that if a user hits cancel during a CTAP operation from the host, that the data we want to send might never get polled by the host and we will be stuck? I think I'm missing how that can happen, the host still continuously polls the device for interrupt data. So the data should still be sent over USB, although it will then not be processed by the host.

I'll add a `send_cancel()` to the HIL so that in future if there is a reason to cancel a send it can be implemented.

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
alistair23: I have tested this with both IN and OUT transactions. Full two way communication in generating keys and signing data is working.

kernel/src/hil/usb_hid.rs
alistair23: Marking as resolved as there is now a `send_cancel()` in the HIL.

gendx: > I think I'm missing how that can happen, the host still continuously polls the device for interrupt data.

That's not what I've observed happening, at least on a Linux host. The host polls for interrupt data only when the userspace application requests it.

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
alistair23: Marking as resolved as this works with the Yubico tests and follows the CTAP spec (although there are also other interpretations of the spec). Feel free to unresolve it if you feel it's not addressed.

capsules/src/usb/ctap.rs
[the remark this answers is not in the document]
alistair23: Marking as resolved as this works with the Yubico tests and follows the CTAP spec (although there are also other interpretations of the spec). Feel free to unresolve it if you feel it's not addressed.

alistair23:
Ok, all comments should be addressed.

capsules/src/ctap.rs
```diff
@@ -0,0 +1,310 @@
...
+            callback: OptionalCell::empty(),
+            recv_buf: None,
+            send_buf: None,
+            can_receive: Cell::new(false),
+        }
+    }
+}
+
+pub static mut WRITE_BUF: [u8; 64] = [0; 64];
+pub static mut READ_BUF: [u8; 64] = [0; 64];
```
gendx, line 59: Static mut buffers is a pattern to avoid (#1545). We have `static_init` for that.

capsules/src/ctap.rs
```diff
@@ -0,0 +1,310 @@
...
+
+pub struct CtapDriver<'a, U: usb_hid::UsbHid<'a>> {
+    usb: Option<&'a U>,
+
+    app: Grant<App>,
+    recv_appid: OptionalCell<AppId>,
+    send_appid: OptionalCell<AppId>,
```
gendx, line 66: Does it make sense to allow different applications for receiving and sending?

capsules/src/ctap.rs
```diff
@@ -0,0 +1,310 @@
...
+                    ReturnCode::SUCCESS
+                })
+                .unwrap_or(ReturnCode::FAIL),
+
+            // default
+            _ => ReturnCode::ENOSUPPORT,
+        }
+    }
+
+    /// Subscribe to HmacDriver events.
```
gendx, line 221: This is not up-to-date.

capsules/src/driver.rs
```diff
@@ -36,6 +36,7 @@ pub enum NUM {
     Rng                   = 0x40001,
     Crc                   = 0x40002,
     Hmac                  = 0x40003,
+    Ctap                  = 0x40004,
```
gendx, line 39: I don't think "cryptography" is the right category for that. It's more a transport layer than cryptography (the cryptography happens in the userspace). Also, this should be named CtapUsb or CtapHidUsb, as it's possible to use CTAP over other protocols (Bluetooth, NFC), which would be different drivers.

capsules/src/ctap.rs
alistair23: Removed

capsules/src/driver.rs
alistair23: I have changed the name to `CtapHid`.

I see what you mean about it not being cryptography, but I still feel that this is the right section. Although it isn't cryptography itself, it is very related to it.

capsules/src/ctap.rs
alistair23: Nope. Fixed.

boards/opentitan/src/main.rs
```diff
@@ -286,20 +291,22 @@ pub unsafe fn reset_handler() {
         ]
     );
 
-    let cdc = components::cdc::CdcAcmComponent::new(
+    let ctap_send_buffer = static_init!([u8; 64], [0; 64]);
+    let ctap_recv_buffer = static_init!([u8; 64], [0; 64]);
+
+    let (ctap, ctap_driver) = components::ctap::CtapComponent::new(
         &earlgrey::usbdev::USB,
-        capsules::usb::cdc::MAX_CTRL_PACKET_SIZE_NRF52840,
```
gendx, line 291: It seems to me that the max packet size should be provided to the API for CTAP as well. But maybe we should rather fix #2123 first instead?

alistair23: CTAP requires a max packet of 64 bytes, I don't think it makes sense to specify.

capsules/src/driver.rs
gendx: Well, I don't feel that this is the right section. Should driver number assignments be reviewed and agreed upon by the Tock core team (e.g. in a weekly call)?

gendx: On the other hand, I don't think these categories have been of much use, and from Tock 2.0 on we could simply allocate driver numbers linearly instead. This would enable some potential optimizations that have been discussed such as storing the driver number on 16 bits, and the sub-driver number on 16 bits as well, both being packed in a single 32-bit register.

alistair23: Agreed. The driver assignment is pretty unimportant, just as long as it doesn't change. libtock-rs is already using this driver number so I would like to avoid changing it unless we have to.

I also agree with 16-bit numbers as part of Tock 2.0.

bradjc (dismissed):

bradjc:
@phil-levis

boards/opentitan/src/main.rs
gendx: Then it will break for the current sam4l, right?

gendx (changes requested):
I don't think this pull request is ready. There are still references to "HMAC", cancellation is not implemented, etc.

alistair23:
I removed all of the Hmac references, there was 1 left over.

As for cancellation:
 - It is in the HIL, which is the part that is hard to change in the future.
 - There isn't an implementation yet, but currently I have no use or test case for it.
 - I don't see how we could cancel a USB transaction via the current USB HIL, so I think it will require USB HIL changes (that is a growing list).

kernel/src/hil/usb_hid.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn send_buffer(
+        &'a self,
+        send: &'static mut [u8; 64],
+    ) -> Result<usize, (ReturnCode, &'static mut [u8; 64])>;
+
+    /// Cancels a send called by `send_buffer()`. Once called the
+    /// `packet_transmitted()` callback won't occur and no new data
+    /// will be sent. Any data currently being sent will stil be sent.
+    /// Returns the buffer passed in to `send_buffer()`.
+    fn send_cancel(&'a self) -> &'static mut [u8; 64];
```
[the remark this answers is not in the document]
alistair23: Good point, fixing now.

kernel/src/hil/usb_hid.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+pub trait UsbHid<'a> {
+    /// Sets the buffer to be sent and starts a send transaction.
+    /// On success returns the number of bytes sent.
+    /// On failure returns an error code and the buffer passed in.
+    fn send_buffer(
+        &'a self,
+        send: &'static mut [u8; 64],
+    ) -> Result<usize, (ReturnCode, &'static mut [u8; 64])>;
```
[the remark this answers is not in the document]
alistair23: This matches what some HILs such as digest do: https://github.com/tock/tock/blob/master/kernel/src/hil/digest.rs

kernel/src/hil/usb_hid.rs
```diff
@@ -0,0 +1,67 @@
+//! Interface for USB HID (Human Interface Device) class
+
+use crate::returncode::ReturnCode;
+
+/// Implement this trait and use `set_client()` in order to receive callbacks.
+pub trait Client<'a> {
+    /// Called when a packet is received.
+    /// This will return the buffer passed into `receive_buffer()` as well as
+    /// the endpoint where the data was received.
+    fn packet_received(&'a self, buffer: &'static mut [u8; 64], len: usize, endpoint: usize);
```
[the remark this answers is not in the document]
alistair23: I have added a `result: Result<(), ReturnCode>`

kernel/src/hil/usb_hid.rs
```diff
@@ -0,0 +1,67 @@
...
+        &'a self,
+        result: Result<(), ReturnCode>,
+        buffer: &'static mut [u8; 64],
+        len: usize,
+        endpoint: usize,
+    );
+
+    /// Called when checking if we can receive any more data.
+    /// Should return true if we are ready to receive.
+    fn can_receive(&'a self) -> bool;
```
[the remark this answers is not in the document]
alistair23: I have changed this to:

```Rust
    /// Called when checking if we can start a new receive operation.
    /// Should return true if we are ready to receive and not currently
    /// in the process of receiving anything. That is if we are currently
    /// idle.
    /// If there is an outstanding call to receive, a callback already
    /// waiting to be called then this will return false.
```
to be more clear

alistair23:
I added cancellation support, I think that addresses most of @gendx 's comments

kernel/src/hil/usb_hid.rs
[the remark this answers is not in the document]
alistair23: Yep! I like `Result` :) 

I don't agree with the all or nothing approach, I think it makes sense to gradually transition across with new HILs and updated HILs. Result works well with other Rust ideas as well.

Let me know what you think of the updated interface and I'll go from there.

alistair23:
I also updated the testing description to be more explicit about what works.

alistair23:
I have added a `send or receive` command so that the new userspace API can be used by OpenSK. That should address all concerns raised in https://github.com/tock/tock/pull/2125.

boards/opentitan/src/main.rs
alistair23: Yes, the SAM4L hardware can not meet the CTAP spec of a max packet size of 64 bytes, https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html#usb-descriptors

phil-levis (approved):
HIL looks reasonable to me.

phil-levis (commented):
The USB HIL seems reasonable, but I agree with @gendx that we should check it is sufficient/works with the full functionality in the OpenSK implementation. If we can get agreement on the HIL, I'm OK with there being multiple implementations at differing levels of functionality. But in the context of #2125 I think we should break out the HIL into a separate PR, then have PRs for each implementation.

alistair23:
So you want me to split the HIL out and we both rebase our work on the HIL?

Then OpenSK tests with that to make sure that works. Then we can merge two similar implementations, but both using the same HIL?

alistair23:
I split out a HID HIL: https://github.com/tock/tock/pull/2130

phil-levis:
> 
> 
> So you want me to split the HIL out and we both rebase our work on the HIL?

That's the goal, yes.

> 
> Then OpenSK tests with that to make sure that works. 

Yes. The major purpose of this is verification that the HIL covers the cases and isn't too implementation-specific.

> Then we can merge two implementations, but both using the same HIL?

I don't know -- let's see how it goes?  That seems like a good path but let's see whether we run into problems between the implementations.

alistair23:
Now that the HID has been merged this is ready to merge.

phil-levis:
I'll do a detailed read-through of this PR and #2125 today so we can move forward.

phil-levis:
I did a read-though of this PR and #2125. At the CTAP implementation level, they are mostly equivalent in terms of functionality. The major behavioral difference at the USB level is how the endpoint ID is filled in on delay receptions; this PR can handle when there are multiple endpoints for OUT(although there aren't currently), while #2125 just assumes the static endpoint identifier. I think the implementation in #2125 is slightly cleaner: `packet_out` and `packet_in` in this PR are more complex, but not so complex that it's a problem.

The difference at the system call level is whether there is a single buffer (for send or receive) or two buffers (one send, one receive). I personally think the double-buffer approach is preferable: I recall all of the pain of the RF233 radio having a shared receive/transmit buffer, such that if you go to transmit, but then a frame is received, you lose the contents of the transmit buffer. 

 So I'm going to make the (probably unpopular) suggestion that we use the CTAP capsule from #2125 and the system call capsule from #2094 . This will require a little bit of testing to make sure they work correctly together, but if that finds ambiguities and implementation assumptions that's a good thing.

gendx:
One unexpected thing that came up while trying to adapt #2125 to the USB-HID HIL (while testing on nRF52840-DK), is how to initialize the USB stack - in particular how to handle the life-cycle of USB getting plugged in, ready, or plugged out.

So far, the implementation for the nRF52* chip (https://github.com/tock/tock/blob/master/chips/nrf52/src/usbd.rs) was modeled over the `usb_user.rs` example capsule. In this example, the USB controller in only enabled and attached when an application calls a specific command syscall:

https://github.com/tock/tock/blob/ba867c2c269325133f212b757b125b114ab9f2ad/capsules/src/usb/usb_user.rs#L74-L76

In practice this allowed enough time for the USB controller to initialize itself upon boot (provided Tock is started with the USB cable plugged in), and we followed the same model in #2125.

However, this #2094 changes this model, by doing this initialization earlier - when the board is initialized.

https://github.com/tock/tock/blob/ee90aad807bbc785215aad14f6e5354b300eda87/boards/nordic/nrf52840dk/src/main.rs#L453-L454

As it turns out, this breaks assumptions that were made when implementing the USB HIL on the nRF52* chip - i.e. assuming that the USB cable is always plugged in and ready when we interact with it. In practice this doesn't work with #2094 because the enable/attach calls happen too early (the USB is not ready yet).

Besides, the `usb_user.rs` model (i.e. enable/attach following a userspace command) doesn't work with the new USB-HID HIL (#2130), because the capsule now gets a reference to the `UsbHid` trait - which itself doesn't expose enable/attach.

https://github.com/tock/tock/blob/d1f8b0deffc680a9eed8922ea2a2717133987088/capsules/src/usb/usb_ctap.rs#L88

To overcome this problem, we could expose enable/attach to the `UsbHid` trait. But it doesn't seem like a clean approach - I think that we should rather fix the chip implementation to for example fully initialize itself only when the relevant "ready" interrupt arrives. However, but this raises another issue: the USB HIL (and *a fortiori* the USB-HID HIL) currently doesn't represent the lifecycle of the USB - being plugged in or not, and being ready or not.

In other words:
- What should a capsule do if USB is not ready yet?
- What should a capsule do if USB is plugged out and re-plugged in?
- Should we surface events back to userspace? I'd assume yes, because the application layer should somehow reinitialize its communication state when the USB cable is plugged out and then in again. Or the application may want to wait until USB is ready.

I think we should resolve this questions and have a clearer vision on what the USB HILs and APIs should look like before pushing too much implementation in a chip- and use-case-specific direction. I think it's becoming clear that we're past the "USB is an unstable & iterating fast prototype" stage.

alistair23:
@phil-levis as you already know that view is unpopular :)

Merging both together is a large amount of work that I really don't want to do. It also looses the advantages of #2094 in being generic over multiple transport protocols (USB, BLE and NFC).

gendx:
> As well as that #2094 follows the CTAP reference example in using two endpoints while #2125 doesn't.

It's the opposite: #2125 closely follows the [CTAP reference example](https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html#usb-descriptors) by using endpoint 1 in both IN an OUT modes, with the addresses `0x01` and `0x81`.

Could you please tell which CTAP reference example is followed by #2094 where two separate endpoints are described?

alistair23:
As discussed before, the CTAP reference example you linked to explicitly says:

> The device implements two endpoints (except the control endpoint 0), one for IN and one for OUT transfers.

That is what this implementation does.

The spec later on says:

> The actual endpoint order, intervals, endpoint numbers and endpoint packet size may be defined freely by the vendor and the host application is responsible for querying these values and handle these accordingly. For the sake of clarity, the values listed above are used in the following examples.

So combining the two into a single endpoint, as done in #2125 is allowed and matches the spec. It just doesn't match the reference example. That might cause issues with host implementations that assume all devices use the reference example, but in those cases the host won't be spec compliant.

gendx:
> So combining the two into a single endpoint, as done in #2125 is allowed and matches the spec. It just doesn't match the reference example.

Again, the [reference example](https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html#usb-descriptors) explicitly mentions:

```
Endpoint 1 descriptor
Mnemonic         | Value | Description
...
bEndpointAdresss | 0x01  | 1, OUT
...

Endpoint 2 descriptor
Mnemonic         | Value | Description
...
bEndpointAdresss | 0x81  | 1, IN
...
```

Which means that the receive and transmit directions (from the device's point of view) both use endpoint number 1 (as indicated in the "endpoint address" field, "description" column), in OUT and IN mode respectively.

Therefore, #2125 follows this example from the spec, while #2094 doesn't.

> That might cause issues with host implementations that assume all devices use the reference example, but in those cases the host won't be spec compliant.

This I can agree with, although haven't tested that on real hosts so cannot tell at this point whether there are non-compliant hosts.

jmichelp:
> #2094 on the other hand has a OUT endpoint 1 at address `0x01` and a IN endpoint 2 at address `0x81`. This is also valid AND matches the reference implementation described in the CTAP spec.

Per USB specification, an endpoint address is stored on **7 bits**. The MSB indicates whether the endpoint is IN or OUT. So address `0x01` and address `0x81` are actually the same endpoint address: `0x01`. Your endpoint 2 will end up at address `0x82` on the descriptor.

And I think that all the confusion on this discussion comes from that point.

alistair23:
Yep, agreed. Although I didn't realise that initially I completely agree that it is valid. They are the same address just a IN and OUT endpoint.

I'm not saying that #2125 is invalid, it just doesn't match the reference implementation example in having 3 total (including control) endpoints.

As for #2094 having an endpoint at address `0x82` that is not true. It is at `0x81` as identified by hosts and a logic analyser.

jmichelp:
It's getting a bit off-topic so I'm happy taking it offline (email or slack for example) to avoid creating too much noise on this PR. But it might reveal an underlying bug somewhere, either in Tock or in hardware so I think it's worth having this conversation somewhere.

In this PR, in `capsules/src/usb/ctap.rs` we have:
```rust
let endpoints: &[&[EndpointDescriptor]] = &[&[
    EndpointDescriptor {
        endpoint_address: EndpointAddress::new_const(0x02, TransferDirection::DeviceToHost),
        ...
    },
    EndpointDescriptor {
        endpoint_address: EndpointAddress::new_const(0x01, TransferDirection::HostToDevice),
        ...
    },
]];
```

Let's follow the code, `EndpointAddress::new_const()` is implemented in [capsules/src/usb/descriptors.rs](https://github.com/tock/tock/blob/master/capsules/src/usb/descriptors.rs#L673-L675) and here are the relevant snippets:
```rust
#[derive(Debug)]
pub enum TransferDirection {
    HostToDevice = 0,
    DeviceToHost = 1,
}
[...]
pub struct EndpointAddress(u8);

impl EndpointAddress {
   [...]
    pub const fn new_const(endpoint: usize, direction: TransferDirection) -> Self {
        EndpointAddress(endpoint as u8 & 0xf | (direction as u8) << 7)
    }
```

So the resulting `endpoint_address` values are going to be, respectively:
* `0x82` (`0x02 | TransferDirection::DeviceToHost`) and
* `0x01` (`0x01 | TransferDirection::HostToDevice`).

This `endpoints` array is then passed as an argument to `descriptors::create_descriptor_buffers()` (also implemented in [capsules/src/usb/descriptors.rs](https://github.com/tock/tock/blob/master/capsules/src/usb/descriptors.rs#L418)) which will simply serialize the Rust structures into an `u8` buffer. This buffer will be transmitted as-is by the controler when it will receive the configuration/interface enumeration commands on the control endpoint.

So if your resulting endpoint address on the hardware is `0x81`, I have no idea how you got it but that's not what the code should produce from what I'm reading.

alistair23:
Ok, I have converted this to use a single endpoint for the IN and OUT transactions.

That should hopefully address both of your concerns.

alistair23:
Ping! From the discussion today it sounds like this is ready to be merged.

chips/lowrisc/src/usbdev.rs
```diff
@@ -1140,8 +1140,14 @@ impl<'a> hil::usb::UsbController<'a> for Usb<'a> {
...
-    fn endpoint_in_out_enable(&self, _transfer_type: TransferType, _endpoint: usize) {
-        unimplemented!()
+    fn endpoint_in_out_enable(&self, transfer_type: TransferType, endpoint: usize) {
+        match transfer_type {
+            TransferType::Bulk => {
+                panic!("IN/OUT Bulk transfer is unsupported ");
+            }
+            _ => {}
+        }
+        self.endpoint_out_enable(transfer_type, endpoint)
```
gendx, line 1150: Seems strange than the endpoint doesn't need to be enabled in IN mode as well. But then, I'm not familiar with this chip.

capsules/src/usb/ctap.rs
```diff
@@ -0,0 +1,405 @@
...
+            interface_number: 0,
+            interface_class: 0x03,    // HID
+            interface_subclass: 0x00, // No subcall
+            interface_protocol: 0x00, // No protocol
+            ..InterfaceDescriptor::default()
+        }];
+
+        let endpoints: &[&[EndpointDescriptor]] = &[&[
+            EndpointDescriptor {
+                endpoint_address: EndpointAddress::new_const(0x01, TransferDirection::DeviceToHost),
```
gendx, line 119: Could be worth re-using `ENDPOINT_NUM` here rather than `0x01`. Same below.

capsules/src/usb/ctap.rs
```diff
@@ -0,0 +1,405 @@
...
+}
+
+impl<'a, U: hil::usb::UsbController<'a>> hil::usb::Client<'a> for CtapHid<'a, U> {
+    fn enable(&'a self) {
+        // Set up the default control endpoint
+        self.client_ctrl.enable();
+
+        // Setup buffers for IN and OUT data transfer.
+        self.controller()
+            .endpoint_set_in_buffer(ENDPOINT_NUM, self.buffer(ENDPOINT_NUM));
```
gendx, line 259: I'm surprised no OUT buffer is set.

boards/nordic/nrf52840dk/src/main.rs
```diff
@@ -419,6 +421,38 @@ pub unsafe fn reset_handler() {
...
+    //     0x503a, // lowRISC generic FS USB
+    //     strings,
+    //     board_kernel,
+    //     ctap_send_buffer,
+    //     ctap_recv_buffer,
+    // )
+    // .finalize(components::usb_ctap_component_helper!(nrf52840::usbd::Usbd));
+
+    // ctap.enable();
+    // ctap.attach();
```
gendx, line 454: From what I tested, this is the part that currently doesn't work on nRF52840-DK, as the USB is generally not ready yet this early in the board initialization process just after reboot.

alistair23: Do you want to send a patch that changes this to something that does work for the nRF52840-DK?

Doing this works fine for OpenTitan. @bradjc added this after testing it on a board, I don't have one to test on.

jmichelp: I don't think it's a reasonable goal for this PR to fix the usb lifecycle. It's a bigger patch, at least on the nordic chip, that needs to happen separately.
Unless someone is interested taking it, as the main users of this board, that's a problem that the OpenSK team will have to work on.

bradjc (approved):

jmichelp:
Thanks a lot for the hard work @alistair23!

bradjc:
Marking this as last call.

alistair23:
Ping!

PR #2097, opened 2020-09-03
OpenTitan: Initial support for the SPI Perhiperal

alistair23:
### Pull Request Overview

This adds initial support for the OpenTitan SPI.

### Testing Strategy

Run on the board. This PR doesn't break anything that currently works, but I could not test the actual SPI bus. The SPI bus isn't broken out and the connection isn't in the schematic so I can't figure out how to test it.

My plan is to come back to this at a later point. Until then let's upstream the current work to avoid code duplication.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/components/src/spi.rs
+++ b/boards/components/src/spi.rs
@@ -207,6 +207,16 @@ impl<S: 'static + spi::SpiSlave + kernel::hil::spi::SpiSlaveDevice> Component
             SpiPeripheral::new(self.device)
         );
 
+        let spi_read_buf =
+            static_init!([u8; DEFAULT_READ_BUF_LENGTH], [0; DEFAULT_READ_BUF_LENGTH]);
+
+        let spi_write_buf = static_init!(
+            [u8; DEFAULT_WRITE_BUF_LENGTH],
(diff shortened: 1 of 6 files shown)


hudson-ayers:
All of the implemented content in this PR looks right to me!

But it feels pretty weird to merge a peripheral driver that does not work at all, especially when the userspace driver is also being included. Tock convention is that userspace examples will check for functionality by querying the existence of the capsule, and that will no longer work. For OpenTitan this maybe applies less because the only hardware available is expensive FPGA hardware and we can currently probably expect that users will be more familiar with OpenTitan/Tock, but I still think it is probably better to leave this as a draft PR and hope that is enough to avoid duplication of work. Happy to be overridden on this if others feel differently though.

alistair23:
I understand. As a user seeing that something exists, but then it not working is a pain. The risk with not merging it though is that we end up with duplicate work.

It currently looks like I'm the only person working on OpenTitan/Tock at the moment though, so I'm going to close this and come back to it later. Hopefully a quick search for Tock PRs/issues will avoid too much duplication here.

PR #2098, opened 2020-09-06
Kernel work increment/decrement prevents MCU going to sleep

alexandruradovici:
### Pull Request Overview

This pull request (hopefully) fixes an issue that seems to prevent the kernel from putting the MCU to sleep. The issue seems to be from the kernel work counter.

While running the following user space app, it seems that the kernel does not sleep the MCU after the process's last yield system call.

```c
char hello[] = "Hello World!\r\n";
char run[] = "system run\r\n";
char texte[] = "texte\r\n";

static void nop(
  int a __attribute__((unused)),
  int b __attribute__((unused)),
  int c __attribute__((unused)),
  void* d __attribute__((unused))) {}


int main(void) {
  putnstr_async(hello, sizeof(hello), nop, NULL);
  putnstr_async(run, sizeof(run), nop, NULL);
  putnstr_async(texte, sizeof(texte), nop, NULL);
  return 0;
}
```

I did some research and found out that the issue pops up only when there are several subscribes for the same callback (driver_id, subscribe_number). In the kernel:

1. I found that when a callback is overwritten:
  a. the kernel work is incremented
  b. the previous callback is removed, but the kernel work is not decremented accordingly

   This prevents sends the scheduler into an infinite loop.

2. I also noticed that `enqueue_task` increments the kernel work regardless of whether a callback is scheduled or not. We also experienced an interesting issue when trying to implement a real-time external interrupt counter. It seems that if the kernel drops callbacks, it starts having a strange behavior, the apps jam. I think it might be related to the kernel work. We still need to do some more research.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -901,8 +901,11 @@ impl<C: Chip> ProcessType for Process<'_, C> {
                 },
                 _ => true,
             });
+            let count_after = tasks.len();
+            for _ in 0..count_before - count_after {
+                self.kernel.decrement_work();
+            }
             if config::CONFIG.trace_syscalls {
-                let count_after = tasks.len();
                 debug!(
                     "[{:?}] remove_pending_callbacks[{:#x}:{}] = {} callback(s) removed",
                     self.appid(),
@@ -910,9 +913,6 @@ impl<C: Chip> ProcessType for Process<'_, C> {
                     callback_id.subscribe_num,
                     count_before - count_after,
                 );
-                for _ in 0..count_before - count_after {
-                    self.kernel.decrement_work();
-                }
             }
         });
     }
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -897,15 +897,19 @@ impl<C: Chip> ProcessType for Process<'_, C> {
                 // to `callback_id`.
                 Task::FunctionCall(function_call) => match function_call.source {
                     FunctionCallSource::Kernel => true,
-                    FunctionCallSource::Driver(id) => id != callback_id,
+                    FunctionCallSource::Driver(id) => {
+                        if id != callback_id {
+                            true
+                        } else {
+                            self.kernel.decrement_work();
+                            false
+                        }
+                    }
                 },
                 _ => true,
             });
-            let count_after = tasks.len();
-            for _ in 0..count_before - count_after {
-                self.kernel.decrement_work();
-            }
             if config::CONFIG.trace_syscalls {
+                let count_after = tasks.len();
                 debug!(
                     "[{:?}] remove_pending_callbacks[{:#x}:{}] = {} callback(s) removed",
                     self.appid(),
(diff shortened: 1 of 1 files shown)


hudson-ayers:
> 1. I found that when a callback is overwritten:
>    a. the kernel work is incremented
>    b. the previous callback is removed, but the kernel work is not decremented accordingly
>    This prevents sends the scheduler into an infinite loop.

Good catch!

> 2. I also noticed that `enqueue_task` increments the kernel work regardless of whether a callback is scheduled or not. We also experienced an interesting issue when trying to implement a real-time external interrupt counter. It seems that if the kernel drops callbacks, it starts having a strange behavior, the apps jam. I think it might be related to the kernel work. We still need to do some more research.

Another good catch! It also seems incorrect that the kernel work is incremented even in the case that `enqueue_task()` fails, I think we should fix that in this PR as well.

kernel/src/process.rs
```diff
@@ -901,8 +901,11 @@ impl<C: Chip> ProcessType for Process<'_, C> {
                 },
                 _ => true,
             });
+            let count_after = tasks.len();
+            for _ in 0..count_before - count_after {
+                self.kernel.decrement_work();
```
hudson-ayers, line 906: Is it feasible to just call this within the closure passed to `retain()`, whenever that closure returns true? It seems like a shame to include another iteration through the callbacks ring buffer when its not really needed. In fact, it seems like calculating `count_before` probably should have previously been gated by `CONFIG.trace_syscalls`

alexandruradovici:
> > 1. I found that when a callback is overwritten:
> >    a. the kernel work is incremented
> >    b. the previous callback is removed, but the kernel work is not decremented accordingly
> >    This prevents sends the scheduler into an infinite loop.
> 
> Good catch!
> 
> > 1. I also noticed that `enqueue_task` increments the kernel work regardless of whether a callback is scheduled or not. We also experienced an interesting issue when trying to implement a real-time external interrupt counter. It seems that if the kernel drops callbacks, it starts having a strange behavior, the apps jam. I think it might be related to the kernel work. We still need to do some more research.
> 
> Another good catch! It also seems incorrect that the kernel work is incremented even in the case that `enqueue_task()` fails, I think we should fix that in this PR as well.

I think I fixed this already by moving the increment_work function call. Am I missing something else here?

kernel/src/process.rs
alexandruradovici: Modified it, that is a good idea.

kernel/src/process.rs
```diff
@@ -879,6 +877,8 @@ impl<C: Chip> ProcessType for Process<'_, C> {
             self.debug.map(|debug| {
                 debug.dropped_callback_count += 1;
             });
+        } else {
+            self.kernel.increment_work();
```
alexandruradovici, line 881: increment_work is called only when ret is true and the callback has been added

hudson-ayers:
> I think I fixed this already by moving the increment_work function call. Am I missing something else here?

Oh, I must have been looking at a single commit or something. You did fix it.

hudson-ayers (approved):

bradjc (approved):

alistair23 (approved):

PR #2104, opened 2020-09-11
opentitan: Makefile: Use riscv64-elf-objcopy to prepare images

alistair23:
### Pull Request Overview

Call the riscv64-elf-objcopy binary to create binaries.

The LLVM objcopy doesn't correctly create working binaries. So let's
avoid calling it. At the same time the riscv64 will correctly work and
is more likely to be installed so let's use it instead.

### Testing Strategy

Running the Makefile to flash images to the OT FPGA.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/Makefile
+++ b/boards/opentitan/Makefile
@@ -26,6 +26,6 @@ flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
 	$(OPENTITAN_TREE)/build-out/sw/host/spiflash/spiflash --input=$(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
 
 flash-app: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
-	riscv32-none-elf-objcopy --update-section .apps=$(APP) $^ $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.elf
-	$(OBJCOPY) --output-target=binary $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.elf $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.bin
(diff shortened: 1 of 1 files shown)


bradjc (changes requested):
Can we dig into why? Conceptually what is happening is pretty simple. And, we have seen how painful it can be to install riscv64-gcc, which is why using the rust provided tools is so convenient.

alistair23:
I have no idea and I don't have a good way to figure out. It's just a binary blob. One boots on OT and one doesn't.

ppannuto:
Maybe worth pinging the LLVM devs? They've been responsive to issues like this before, seems they aspire to parity in tooling where sensible / possible, and this is probably a bug there.

alistair23:
I filed a bug, let's see what happens: https://bugs.llvm.org/show_bug.cgi?id=47563

hudson-ayers:
@alistair23 does `riscv32-none-elf-objcopy` not work if used in both positions?

Given that this rule already requires a risc-v specific objcopy because llvm objcopy does not support `--update-section`, it seems like it might be worth merging this to fix master, especially if we can keep it so we only require the objcopy that people had to use previously.

alistair23:
@hudson-ayers Do you mean `llvm-objcopy`?

We can't use `llvm-objcopy` for the first objcopy as it doesn't support `--update-sections`.

Most machines don't use `riscv32-none-elf-objcopy` but instead use `riscv64-elf-objcopy` hence the change to it. Both are GCC tools though. `riscv32-none-elf-objcopy` came from my OE toolchain, most users will just use `riscv64-elf-objcopy`.

hudson-ayers:
> @hudson-ayers Do you mean `llvm-objcopy`?
> 
> We can't use `llvm-objcopy` for the first objcopy as it doesn't support `--update-sections`.
> 
> Most machines don't use `riscv32-none-elf-objcopy` but instead use `riscv64-elf-objcopy` hence the change to it. Both are GCC tools though. `riscv32-none-elf-objcopy` came from my OE toolchain, most users will just use `riscv64-elf-objcopy`.

I meant `riscv32-none-elf-objcopy` just because that is what was there previously. But that makes sense, I think it is reasonable to merge this because a non-llvm objcopy is required for this make rule anyway.

hudson-ayers (approved):

bradjc (approved):

PR #2107, opened 2020-09-11
Scheduler timer bugfix + documentation updates

hudson-ayers:
### Pull Request Overview

This pull request fixes a potential (though as of yet undetected) bug in the kernel's use of the SchedulerTimer trait, where the kernel checked for expiration of the timer, then checked the amount of time remaining, not considering the possibility that the timer could wraparound in period between those checks, causing the kernel to read an invalid time remaining value. This was identified by @bradjc .

Rather than just fix the bug, we decided  to change the interface slightly to make this misuse impossible -- rather than separate `has_expired()` and `get_remaining_us()` functions, there is now only a `get_remaining_us()` function, which returns `None` if the timer has expired.

This PR also includes lots of documentation updates by @bradjc that clarify the `SchedulerTimer` interface further.

### Testing Strategy

This pull request was tested by running several apps simultaneously on Imix, incuding whileone, and verifying that timeslice expiration still works as expected.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/systick.rs
+++ b/arch/cortex-m/src/systick.rs
@@ -147,10 +147,6 @@ impl kernel::SchedulerTimer for SysTick {
             .write(ControlAndStatus::ENABLE::SET + clock_source);
     }
 
-    fn has_expired(&self) -> bool {
-        SYSTICK_BASE.syst_csr.is_set(ControlAndStatus::COUNTFLAG)
-    }
-
     fn reset(&self) {
         SYSTICK_BASE.syst_csr.set(0);
         SYSTICK_BASE.syst_rvr.set(0);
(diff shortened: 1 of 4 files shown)


bradjc:
The other issue with the current interface is a user could call `has_expired()`, get a `false` return value, and then reasonably call `get_remaining_us()` to see how much time is remaining. However, the timer could have expired between those two calls, and once it has, the call to `get_remaining_us()` is invalid.

bradjc:
I'm nominating this for release-blocker.

bradjc (approved):

hudson-ayers:
bors retry

PR #2108, opened 2020-09-11
MSP432: ADC support

lebakassemmerl:
### Pull Request Overview

This pull request adds ADC support where for the beginning only single samples are supported. When this PR will be accepted, I will start implementing the rest of the ADC-API.

### Testing Strategy

It was tested on the evalboard by measuring actual voltages and comparing the output.

### TODO or Help Wanted

- [x] Support for single samples
- [x] Support for continuous mode
- [x] Support for highspeed mode

### Documentation Updated

### Formatting

- [x] Ran `make ci-nosetup`.

### DIFF
--- a/boards/msp_exp432p401r/src/main.rs
+++ b/boards/msp_exp432p401r/src/main.rs
@@ -53,6 +53,7 @@ struct MspExp432P401R {
         capsules::virtual_alarm::VirtualMuxAlarm<'static, msp432::timer::TimerA<'static>>,
     >,
     ipc: kernel::ipc::IPC,
+    adc: &'static capsules::adc::Adc<'static, msp432::adc::Adc>,
 }
 
 /// Mapping of integer syscalls to objects that implement syscalls.
@@ -68,6 +69,7 @@ impl Platform for MspExp432P401R {
             capsules::gpio::DRIVER_NUM => f(Some(self.gpio)),
             capsules::alarm::DRIVER_NUM => f(Some(self.alarm)),
             kernel::ipc::DRIVER_NUM => f(Some(&self.ipc)),
+            capsules::adc::DRIVER_NUM => f(Some(self.adc)),
             _ => f(None),
         }
     }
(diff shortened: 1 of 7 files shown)


boards/msp_exp432p401r/src/main.rs
```diff
@@ -183,18 +218,75 @@ pub unsafe fn reset_handler() {
...
+            &msp432::adc::ADC,
+            grant_adc,
+            adc_channels,
+            &mut capsules::adc::ADC_BUFFER1,
+            &mut capsules::adc::ADC_BUFFER2,
+            &mut capsules::adc::ADC_BUFFER3
+        )
+    );
+
+    msp432::adc::ADC.set_client(adc);
```
alistair23, line 277: Shouldn't this use a board component?

lebakassemmerl: Actually I really don't know.
I copied this part from the main.rs of the hail- and stm32-discovery board.
Do you mean that the `adc` variable should be a board component?

bradjc:
This PR looks good to me, but I'm unsure about adding the capsule and syscall interface when most of the interface isn't functional.

lebakassemmerl:
> This PR looks good to me, but I'm unsure about adding the capsule and syscall interface when most of the interface isn't functional.

Ok, this makes sense. But I thought it would be easier to review only single sample functionality instead of the complete traits.

I could remove the capsule support or also add the continuous mode and high speed mode if you want.

boards/msp_exp432p401r/src/main.rs
alistair23: Yeah, like the alarm component for example.

lebakassemmerl: > Yeah, like the alarm component for example.

I think I don't exactly get what you mean. As far as I can tell it's already a board-component. In line 289 in the main.rs this `adc` variable is added to the `MspExp432P401R` struct.

hudson-ayers:
> > This PR looks good to me, but I'm unsure about adding the capsule and syscall interface when most of the interface isn't functional.
> 
> Ok, this makes sense. But I thought it would be easier to review only single sample functionality instead of the complete traits.
> 
> I could remove the capsule support or also add the continuous mode and high speed mode if you want.

I think either route would be okay, but adding continuous/high speed would be great.

lebakassemmerl:
Ok!
In this case I will add continuous and high speed mode support.

I guess it would make sense to convert the PR to a draft in the meantime?

lebakassemmerl:
With the latest commit the complete ADC traits are now implemented and can be used. But I have still one design/implementation specific question:

The internal ADC is able to operate at a sampling frequency of 1Msps with 14bit. The very unfortunate thing is, that in this case the limiting part are the 16bit timers. I have to choose between the high sampling rate up to 1Msps or between accurate low-frequent samples e.g. 1Hz to 100Hz. I now focused the low frequencies and limited the maximum sampling frequency to 150kHz, because the timer is clocked with 1.5MHz and can be divided perfectly down to 150kHz.
Is this approach fine for you? Or are the high frequencies more important than the extremely low one?

hudson-ayers:
> With the latest commit the complete ADC traits are now implemented and can be used. But I have still one design/implementation specific question:
> 
> The internal ADC is able to operate at a sampling frequency of 1Msps with 14bit. The very unfortunate thing is, that in this case the limiting part are the 16bit timers. I have to choose between the high sampling rate up to 1Msps or between accurate low-frequent samples e.g. 1Hz to 100Hz. I now focused the low frequencies and limited the maximum sampling frequency to 150kHz, because the timer is clocked with 1.5MHz and can be divided perfectly down to 150kHz.
> Is this approach fine for you? Or are the high frequencies more important than the extremely low one?

This approach seems right to me. As long as we are relying on static clocks for most peripherals this is a fundamental tradeoff.

lebakassemmerl:
>
> This approach seems right to me. As long as we are relying on static clocks for most peripherals this is a fundamental tradeoff.

Ok, fine!
Then I will leave the implementation like it is now (:

boards/msp_exp432p401r/src/main.rs
hudson-ayers: Alistair is referring to "components", a tool that Tock uses to ensure uniform initialization across boards. There is an ADC component in `boards/components/src/adc.rs` which is used by several of the STM boards.  However, I think those components only work to construct a virtual ADC, which does not appear to be what you are trying to do here, so I think it is fine if this does not use a component for now.

chips/msp432/src/adc.rs
```diff
@@ -0,0 +1,982 @@
...
+
+fn buf_u8_to_buf_u16(buf: &'static mut [u8]) -> &'static mut [u16] {
+    let buf_ptr = unsafe { mem::transmute::<*mut u8, *mut u16>(buf.as_mut_ptr()) };
+    unsafe { slice::from_raw_parts_mut(buf_ptr, buf.len() / 2) }
+}
+
+fn buf_u16_to_buf_u8(buf: &'static mut [u16]) -> &'static mut [u8] {
+    let buf_ptr = unsafe { mem::transmute::<*mut u16, *mut u8>(buf.as_mut_ptr()) };
+    unsafe { slice::from_raw_parts_mut(buf_ptr, buf.len() * 2) }
+}
```
hudson-ayers, line 579: These functions are not safe to call for arbitrary input slices, and so should be marked `unsafe` and their requirements documented. If an odd length slice is passed to `buf_u8_to_buf_u16`, it should probably panic rather than go ahead with a transmute. 

This is chip specific code, so we don't really have to worry about endianness, but we do have to consider the possibility for alignment issues when converting from u8 to u16 -- the individual bytes might not be aligned for access as u16s.

hudson-ayers (changes requested):

boards/msp_exp432p401r/src/main.rs
lebakassemmerl: Ok, then I'll leave it now as it is. Thanks for the explanation!

chips/msp432/src/adc.rs
lebakassemmerl: Ahh yeah, of course! I completely oversaw that fact, that u8 to u16 can be a big problem.
I'll fix that as soon as possible!

lebakassemmerl: With the latest commit I marked the 2 functions as unsafe and documented why they are necessary and what the requirement for `buf_u8_to_buf_u16` is. I hope these functions now correctly implemented.

hudson-ayers (approved):
This looks good to me now, it is exciting to have ADC support for another platform.

I still don't love the need for unsafe outside of declaring registers, but at this point this is in line with what the SAM4L ADC driver does. It seems like somewhat of a design wart that this `u16` -> `u8` buffer conversion is required in each ADC peripheral driver because of the conflicting requirements of the ADC HIL and the typical DMA interface which requires a `[u8]` slice.

bradjc (approved):

PR #2111, opened 2020-09-15
Run Tock as a standard POSIX process

alexandruradovici:
### Pull Request Overview

This pull request adds the ability to run Tock as a standard POSIX process. For now, it may run in x86_64 Linux. The purpose of this port is:
  1. to allow easier debugging (ex: using the rust `std` library, this is how we got to #2098).
  2. be able to run the Tock stack natively without any emulation

#### Roadmap
 - [x] Context Switch
 - [x] Simulated UART Console
 - [ ] Alarm
 - [ ] Simulated GPIO
 - [ ] Memory Protection

#### Kernel
The Tock kernel is built as a standard POSIX process using the platform executable (ELF so far). Interrupts are simulated using POSIX signals (SIGALRM, ...).

#### POSIX Registers
As POSIX systems run using several CPU architectures, we have defined the following register convention for every CPU:
  - there is a PC register
  - there is a SP register
  - there are five registers for performing system calls: R0 - R4

For x86_64 we have:

| POSIX | x86_64 |
|-----|------|
| PC | RIP |
| SP | RSP |
| R0 | RAX |
| R1 | RBX |
| R2 | RCX |
| R3 | RDX |
| R4 | R10 |

#### Context Switch

The context switch and system calls are simulated using SIGSEGV and SIGALRM. The context saved and restored using the `sigaction`'s third parameter.

```rust
extern "C" fn handle_sigsegv(
    _signal: libc::c_int,
    _siginfo: *mut libc::siginfo_t,
    signal_context: *mut libc::c_void,
) {
    let mut context = unsafe { &mut *(signal_context as *mut usize as *mut PosixContext) };
    /// ...
}
```

For a syscall, apps will have to perform the following action:
 1. place a _magic number_ (`0x515CA110`) into R0 so that the Tock kernel can distinguish this from a fault
 2. add the syscall number to R0 using R0 = R0 | _syscall_number_
 3. add the syscall parameters to R1-R4
 4. perform a fault (usually mov R0 to address NULL)
 5. get the return valur from R0

```c
static void * tock_syscall(uint64_t syscall_number, uint64_t arg1, uint64_t arg2, uint64_t arg3, uint64_t arg4) {
  syscall_number = 0x515CA110 | syscall_number;
  register uint64_t rax asm ("%rax");
  asm volatile (
    ".intel_syntax\n"
    "mov %%rax, %1\n"
    "mov %%rbx, %2\n"
    "mov %%rcx, %3\n"
    "mov %%rdx, %4\n"
    "mov %%r10, %5\n"
    "mov [0], %%rax\n"
    ".att_syntax\n"
    : "=r"(rax)
    : "r"(syscall_number), "r"(arg1), "r"(arg2), "r"(arg3), "r"(arg4)
    : "rbx", "rcx", "rdx", "r10", "memory");
    return (void*)rax;
}
```
When a SIGSEGV arrives, Tock reads R0 and decides if it is a syscall or a fault.

To switch back to the user space, Tock will perform another fault (to be able to restore the context). The action is:
 1. place a _magic number_ (`0x515CA118`) into R0 (the 4th bit of the LSB is 1)
 2. parform a fault (mov R0 to address NULL)

```rust
extern "C" fn handle_sigsegv(
    _signal: libc::c_int,
    _siginfo: *mut libc::siginfo_t,
    signal_context: *mut libc::c_void,
) {
    let mut context = unsafe { &mut *(signal_context as *mut usize as *mut PosixContext) };
    let key: usize = context.read_register(Regs::r0());
    context.write_register(Regs::r0(), key & 0x7);
    if (key & 0x515CA110) == 0x515CA110 {
        let action = key & 0x7;
        let direction = (key & 0xF) >> 3;
        if direction == 1 {
            // to user space
            unsafe {
                switch_context_user(&mut CURRENT_PROCESS, &mut context);
                switched_context_user();
            }
        } else {
            // to kernel space
            // println!("action {}", action);
            match action {
                0..=4 => unsafe {
                    SYSCALL_FIRED = 1;
                    context.write_register(
                        Regs::pc(),
                        context.read_register(Regs::pc()) + POSIX_NEXT_USER_PC,
                    );
                    switch_context_kernel(&mut CURRENT_PROCESS, &mut context);
                    // uncomment for debug
                    // switched_context_kernel();
                },
                _ => {
                    // fault syscall, consider it a fault
                    unsafe {
                        APP_HARD_FAULT = 1;
                    }
                }
            }
        }
    } else {
        // a fault
        unsafe {
            APP_HARD_FAULT = 1;
        }
    }
}
```

#### Userland

Apps will run with the same privileges as the kernel. Nothing stops the apps from registering another signal handler. As this is a testing platform, we make the assumption that apps are behaving nicely.

#### Relocation

Relocation is a little difficult for x86, as it has much more complex relocation mechanisms. For simplicity, we have chose do perform .text relocation. For this we use a shadow flash. When started, the flash contents is loaded from the executable `.apps` section (defined as `ORIGINAL_FLASH`) to a same size are in memory (defined as `FLASH`).

When starting an app, the apps `.text`, `.rodata` and `.data` sections are copied again from the `ORIGINAL_FLASH` to `FLASH`. This is a must as apps will relocate themselves upon start by modifying `FLASH`. As apps do not know if they have been restarted, upon the first restart, apps will try to relocate themselves, resulting an error.

### Testing Strategy

This pull request was tested using Ubuntu 20.04 and a Intel x86_64 CPU.

### TODO or Help Wanted

This is a work in progress, feedback is greatly appreciated.

This pull request still needs some help top pass the MacOS build test, it should just ignore those builds for now.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


arch/posix/src/support.rs
```diff
@@ -0,0 +1,55 @@
...
+    // Unmask Signals
+    signal::sigprocmask(SigmaskHow::SIG_SETMASK, Some(&current_signals), None).unwrap();
+    res
+}
+
+#[inline(always)]
+/// NOP instruction
+pub fn nop() {
+    unsafe {
+        llvm_asm!("nop" :::: "volatile");
```
alistair23, line 53: Instead of using assembly couldn't this just do nothing?

arch/posix/src/syscall.rs
```diff
@@ -0,0 +1,539 @@
...
+unsafe fn switched_context_kernel() {
+    llvm_asm! ("nop"::::"volatile");
+    write_volatile(&mut RUNNING_KERNEL, true);
+}
+
+// used for debugging
+#[no_mangle]
+#[inline(never)]
+unsafe fn switched_context_user() {
+    llvm_asm! ("nop"::::"volatile");
```
alistair23, line 203: Can we remove these nops?

arch/posix/src/syscall.rs
```diff
@@ -0,0 +1,539 @@
...
+    unsafe {
+        #[cfg(all(target_arch = "x86_64", target_os = "linux"))]
+        llvm_asm!(
+            "
+            mov rax, 0x515CA11F
+            mov [0], rax
+            "
+            : : : "rax" : "volatile", "intel"
+        )
+    }
```
alistair23, line 255: This seems to be the only part that is actually x86 specific (from a quick skim). Is there anyway we could change this to not call assembly code?

If this wasn't x86 specific we could run Tock on any architecture. So we could test RISC-V Tock and RISC-V applications as a userspace application. That would be really cool.

It would not use the RISC-V or ARM arch or board, but it would allow us to test syscalls from applicaions to Tock while running on Linux.

alistair23 (changes requested):

arch/posix/src/support.rs
ppannuto: While it's a bit of a hack, at certain places where we want a few cycles of delay (entering panic comes to mind) we issue calls to `nop` that the kernel expects to actually take real-world time. It probably doesn't matter in practice for the posix platform, but I do think that the `nop` exposed by the arch should be an actual, architectural `nop` rather than something that could be optimized out.

ppannuto: Re the question below, I think that `nop` is a valid instruction on RISC-V, ARM, x86, and x64. That's probably enough to effectively be cross platform anyway.

alistair23: Good point. This is probably ok then. It would be better if we could use a POSIX nop though. Ideally this shouldn't be architecture specific at all.

alexandruradovici: Is there such a thing available? Do you have anything particular in mind?

arch/posix/src/syscall.rs
alexandruradovici: I'm not sure how, the value has to be placed in a register that is architecture dependent. Mayer adding several cfgs?

arch/posix/src/syscall.rs
alexandruradovici: Good point, done.

arch/posix/src/support.rs
alistair23: Not really unfortunately. I can't think of a POSIX NOP syscall. As Par mentioned a NOP should be fine here.

arch/posix/src/syscall.rs
alistair23: Yeah, I see what you mean.

I think the current context switch is a little clunky. It seems best to try and convert it to not be architecture specific at all, instead just using POSIX signals.

Couldn't the app be a signal handler and the kernel return another signal handler (alarm signal for example)?

alexandruradovici: There are some limitations when running code inside a signal handler like:
- unable to use some POSIX API (like print)
- unable to preempt a process (or simulate interrupts)

Using this, it will still be platform dependent.

My suggestions is to use #[cfg(....)] and have some code differently. The number of code lines should not be significant.

alistair23: Your current implementation also requires changes to userspace though.

What we want is that the same ARM or RISC-V app that runs on a board will run on a POSIX Tock kernel (on the same architecture).

alexandruradovici: Porting the POSIX arch to ARM and RISC-V should not require any change in the user space (except svc, as that would trigger the guest os system calls).

The changes made to the user space right now are only due to relocation issues for x86.

bradjc: It might not be possible to run the same say RISC-V app binary on a riscv microcontroller and a riscv linux machine. Maybe there is a way to get that to work, but it's probably not an important goal at this point.

However, the "posix" arch in tock shouldn't be x86 specific. Can we save the syscall state somewhere else? Would just a global variable work? Or sent with a signal somehow?

alexandruradovici: > It might not be possible to run the same say RISC-V app binary on a riscv microcontroller and a riscv linux machine. Maybe there is a way to get that to work, but it's probably not an important goal at this point.
> 
> However, the "posix" arch in tock shouldn't be x86 specific. Can we save the syscall state somewhere else? Would just a global variable work? Or sent with a signal somehow?

The state comes within the signal function parameters, saving and restoring it is not a problem. The problem is that it needs to be changed, to "jump" over the actual fault (PC must be incremented). Additionally, system call parameters are in the registers (even tough this might be solved by using the stack).

Another issue that is different is function parameter passing, as this is different from x86, x86_64, arm, ...

What about exposing the architecture as a feature? It would still use #[cfg(...)] ...

arch/posix/src/syscall.rs
```diff
@@ -0,0 +1,536 @@
...
+
+/// This is used in the syscall handler. When set to 1 this means the
+/// svc_handler was called. Marked `pub` because it is used in the cortex-m*
+/// specific handler.
+#[no_mangle]
+#[used]
+pub static mut SYSCALL_FIRED: usize = 0;
+
+/// This is called in the hard fault handler. When set to 1 this means the hard
+/// fault handler was called. Marked `pub` because it is used in the cortex-m*
```
bradjc, line 20: update

bradjc (commented):
I'd like to make a proposal on how to move forward with this:

1. Rename `arch/posix` to `arch/posix_x86_64` and remove the internal cfgs. If running on posix hosts takes off we can worry about different underlying architectures later.
2. Remove the changes to `process.rs` (assuming the effect is the formatting looks wrong and not that it doesn't compile). We really need something more robust in how we handle debugging for processes. In the short term, complicating the core kernel to run Tock on top of a different OS seems like the wrong tradeoff.
3. Should the chip crate be called "linux"? Is it linux specific? I think everything would run on mac, wouldn't it?

alexandruradovici:
> I'd like to make a proposal on how to move forward with this:
> 
>     1. Rename `arch/posix` to `arch/posix_x86_64` and remove the internal cfgs. If running on posix hosts takes off we can worry about different underlying architectures later.
> 
Ok, make sense.
>     2. Remove the changes to `process.rs` (assuming the effect is the formatting looks wrong and not that it doesn't compile). We really need something more robust in how we handle debugging for processes. In the short term, complicating the core kernel to run Tock on top of a different OS seems like the wrong tradeoff.
> 
Ok, will do.
>     3. Should the chip crate be called "linux"? Is it linux specific? I think everything would run on mac, wouldn't it?
It all depends on the 3rd parameter of a signal handler. I am not sure that macOS has the same opaque structure. It is used for saving the context. For now, it does not build on mac due to a linker issue, I would need some help with that.

arch/posix/src/syscall.rs
alexandruradovici: Fixed

bradjc:
There's also the issue of the dependency in arch/posix. I'm a staunch anti-dependency advocate, but it seems like in this case the dependency doesn't really matter as no one will ever run Tock on top of posix for "real". I do think we should make a note in the arch/posix readme, however, addressing this case and explaining the exception from the rule.

@tock/core-wg thoughts?

alistair23:
I think the dependency is fine here. As you said the posix use case is more for testing/development.

hudson-ayers:
@jrvanwhy is a dependency in one of the `arch/` crates not used by, say, tock-on-titan, going to cause problems for the sandboxed invocation of `Cargo` that is used in the build system of that repository?

I think it is important that out-of-tree boards be able to simply pin a release and build Tock without having to connect to the network or download dependencies, but I am not sure if the dependency here prevents that.

jrvanwhy:
> @jrvanwhy is a dependency in one of the `arch/` crates not used by, say, tock-on-titan, going to cause problems for the sandboxed invocation of `Cargo` that is used in the build system of that repository?
> 
> I think it is important that out-of-tree boards be able to simply pin a release and build Tock without having to connect to the network or download dependencies, but I am not sure if the dependency here prevents that.

I do not believe this will be a problem. My understanding is that as long as we don't depend on the arch crate, we won't need to vendor its dependencies.

If I'm wrong, I can eat my words later and send a PR to fix it :-)

hudson-ayers:
In that case, I am fine with the dependency :)

A note in the README would still be good.

alevy:
@alexandruradovici we'd like to get this in ASAP, ideally before 2.0. What's the status? How can we help?

alevy:
@alexandruradovici I pushed a rebased and slightly fixed (to match changes in master) version to github.com:alevy/tock on branch posix. Can you update from there.

alexandruradovici:
Hey, sorry for the delay, this somehow got to the back of by backlog. @alevy I'll try to get this fixed in the net few days.

alexandruradovici:
I did the requested renames and it seems to work. What do you think should be ready for Tock 2.0?

I could really use some help with the macOS compilation issue.

boards/linux_process_x86_64/src/main.rs
```diff
@@ -0,0 +1,146 @@
...
+static mut APP_FLASH: [u8; 256 * 1024] = [0; 256 * 1024];
+static mut APP_RAM: [u8; 256 * 1024] = [0; 256 * 1024];
+
+// #[no_mangle]
+// #[link_section = ".stack_buffer"]
+// pub static mut STACK_MEMORY: [u8; 0x1000] = [0; 0x1000];
+
+struct LinuxProcess {
+    console: &'static capsules::console::Console<'static>,
+    ipc: kernel::ipc::IPC,
```
bradjc, line 42:
```suggestion
    ipc: kernel::ipc::IPC<NUM_PROCS>,
```

alexandruradovici: Where is this defined? In ipc.rs is does not seem to have any parameter, am I missing something?

bradjc: I think you just need to rebase to master. This is a recent change.

alexandruradovici: It works now, but why did the tests run against the master branch?

lschuermann: The CI always runs against the merged result, as that's what's interesting to see for Tock: do we still pass all tests if this is merged, along with all changes that made it into master in the meantime? Moreover, Bors will queue up each change and run the CI prior to a merge again, such that the status _just before the merge_ and not the CI status from the last push on the PR counts.

bradjc:
I added two commits here: https://github.com/tock/tock/compare/pr/2111-bradjc?expand=1 that allow things to compile for me on mac. I can run the kernel on mac, but since the `obj*` tools are not the same it's not quite ergonomic. However, I think it is good enough for now.

Need to double check that linux still works, however.

bradjc:
Also, we should open a tracking issue for this, since this is just a start and we need a bit more development to make this truly useful.

bradjc:
Closing due to age, but this is a good start if anyone wants to revisit this and update it to match the latest tock apis.

PR #2112, opened 2020-09-16
QDEC Hil and userspace driver + hardware driver for nrf52

mmurray22:
### Pull Request Overview

This pull request adds the QDEC peripheral functionality to TockOS. This includes an addition of HIL and Capsule files. The associated documentation that helped guide the design and implementation of the qdec can be found [here](https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52832.ps.v1.1%2Fqdec.html).

### Testing Strategy

There is a corresponding userspace application which actively tests the full end-to-end functionality of the QDEC peripheral. In order to test the QDEC, a rotary encoder,  nrf52840dk board, and three wires hooked up between the rotary encoder and the GPIO Pins 02, 29, and GND.

This pull request was tested and verified by Michaela Murray (author) and Hudson Ayers.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nordic/nrf52840dk/src/main.rs
+++ b/boards/nordic/nrf52840dk/src/main.rs
@@ -70,6 +70,7 @@
 use capsules::virtual_alarm::VirtualMuxAlarm;
 use kernel::common::dynamic_deferred_call::{DynamicDeferredCall, DynamicDeferredCallClientState};
 use kernel::component::Component;
+use kernel::hil::gpio::Configure;
 #[allow(unused_imports)]
 use kernel::{capabilities, create_capability, debug, debug_gpio, debug_verbose, static_init};
 use nrf52840::gpio::Pin;
@@ -100,6 +101,9 @@ const SPI_CLK: Pin = Pin::P0_19;
 const SPI_MX25R6435F_CHIP_SELECT: Pin = Pin::P0_17;
 const SPI_MX25R6435F_WRITE_PROTECT_PIN: Pin = Pin::P0_22;
 const SPI_MX25R6435F_HOLD_PIN: Pin = Pin::P0_23;
+
+const QDEC_PIN_A: Pin = Pin::P0_02;
+const QDEC_PIN_B: Pin = Pin::P0_29;
 
 // Constants related to the configuration of the 15.4 network stack
 const SRC_MAC: u16 = 0xf00f;
@@ -158,6 +162,7 @@ pub struct Platform {
         capsules::virtual_alarm::VirtualMuxAlarm<'static, nrf52840::rtc::Rtc<'static>>,
     >,
     nonvolatile_storage: &'static capsules::nonvolatile_storage_driver::NonvolatileStorage<'static>,
+    qdec: &'static capsules::qdec::QdecInterface<'static>,
 }
 
 impl kernel::Platform for Platform {
@@ -177,6 +182,7 @@ impl kernel::Platform for Platform {
             capsules::temperature::DRIVER_NUM => f(Some(self.temp)),
             capsules::analog_comparator::DRIVER_NUM => f(Some(self.analog_comparator)),
             capsules::nonvolatile_storage_driver::DRIVER_NUM => f(Some(self.nonvolatile_storage)),
+            capsules::qdec::DRIVER_NUM => f(Some(self.qdec)),
             kernel::ipc::DRIVER_NUM => f(Some(&self.ipc)),
             _ => f(None),
         }
@@ -411,6 +417,26 @@ pub unsafe fn reset_handler() {
         nrf52840::acomp::Comparator
     ));
 
+    // qdec initialization
+    gpio_port[QDEC_PIN_A].make_input();
+    gpio_port[QDEC_PIN_B].make_input();
+    gpio_port[QDEC_PIN_A].set_floating_state(kernel::hil::gpio::FloatingState::PullUp);
+    gpio_port[QDEC_PIN_B].set_floating_state(kernel::hil::gpio::FloatingState::PullUp);
+
+    let qdec_nrf52 = &mut nrf52840::qdec::QDEC;
+    qdec_nrf52.set_pins(
+        nrf52840::pinmux::Pinmux::new(QDEC_PIN_A as u32),
+        nrf52840::pinmux::Pinmux::new(QDEC_PIN_B as u32),
+    );
+    let qdec = static_init!(
+        capsules::qdec::QdecInterface<'static>,
+        capsules::qdec::QdecInterface::new(
+            qdec_nrf52,
+            board_kernel.create_grant(&memory_allocation_capability)
+        )
+    );
+    kernel::hil::qdec::QdecDriver::set_client(&nrf52840::qdec::QDEC, qdec);
+
     nrf52_components::NrfClockComponent::new().finalize(());
 
     let platform = Platform {
@@ -426,6 +452,7 @@ pub unsafe fn reset_handler() {
         alarm,
         analog_comparator,
         nonvolatile_storage,
+        qdec,
         ipc: kernel::ipc::IPC::new(board_kernel, &memory_allocation_capability),
     };
 
--- a/capsules/src/driver.rs
+++ b/capsules/src/driver.rs
@@ -70,6 +70,6 @@ pub enum NUM {
     Buzzer                = 0x90000,
     Screen                = 0x90001,
     Touch                 = 0x90002,
-    Qdec                  = 0x90003
+    Qdec                  = 0x90003,
 }
 }
--- a/capsules/src/driver.rs
+++ b/capsules/src/driver.rs
@@ -69,6 +69,7 @@ pub enum NUM {
     // Misc
     Buzzer                = 0x90000,
     Screen                = 0x90001,
-    Touch                 = 0x90002
+    Touch                 = 0x90002,
+    Qdec                  = 0x90003
 }
 }
--- a/capsules/src/qdec.rs
+++ b/capsules/src/qdec.rs
@@ -1,0 +1,108 @@
+//! Provides userspace access to the Qdec on a board.
+//!
+//! Usage
+//! -----
+//!
+//! ````
+//! let qdec = static_init!(
+//!     capsules::qdec::Qdec<'static>,
+//!     capsules::qdec::QdecInterface::new(&nrf52::qdec::QDEC,
+//!                                         kernel::Grant::create())
+//! );
+//! kernel::hil::QdecDriver.set_client(qdec);
+//! ````
+//! #Interrupt Spurred Readings versus Regular Readings
+//! An application can either enable interrupts to get the
+//! accumulation value or manually read it whenever it wants
+
+use crate::driver;
+use kernel::hil;
+use kernel::{AppId, Callback, Driver, Grant, ReturnCode};
+
+pub const DRIVER_NUM: usize = driver::NUM::Qdec as usize;
+
+/// This struct contains the resources necessary for the QdecInterface
+pub struct QdecInterface<'a> {
+    driver: &'a dyn hil::qdec::QdecDriver,
+    apps: Grant<App>,
+}
+
+#[derive(Default)]
+/// This struct contains the necessary fields for an app
+pub struct App {
+    callback: Option<Callback>,
+    pos: u32,
+}
+
+impl<'a> QdecInterface<'a> {
+    /// Create a new instance of the QdecInterface
+    pub fn new(driver: &'a dyn hil::qdec::QdecDriver, grant: Grant<App>) -> Self {
+        Self {
+            driver: driver,
+            apps: grant,
+        }
+    }
+
+    fn configure_callback(&self, callback: Option<Callback>, app_id: AppId) -> ReturnCode {
+        self.apps
+            .enter(app_id, |app, _| {
+                app.callback = callback;
+                ReturnCode::SUCCESS
+            })
+            .unwrap_or_else(|err| err.into())
+    }
+}
+
+impl<'a> hil::qdec::QdecClient for QdecInterface<'a> {
+    /// Goes through all the apps and if the app is
+    /// subscribed then it sends back the acc value
+    fn sample_ready(&self) {
+        for cntr in self.apps.iter() {
+            cntr.enter(|app, _| {
+                app.pos = app.pos + self.driver.get_acc();
+                app.callback
+                    .map(|mut cb| cb.schedule(app.pos as usize, 0, 0));
+            });
+        }
+    }
+
+    /// Goes through all the apps and if the app recently
+    /// had an overflow, it records the occurance
+    fn overflow(&self) {
+        /*for now, we do not handle overflows*/
+    }
+}
+
+impl<'a> Driver for QdecInterface<'a> {
+    /// Subscribes a client to (newly enabled) interrupts
+    fn subscribe(
+        &self,
+        subscribe_num: usize,
+        callback: Option<Callback>,
+        app_id: AppId,
+    ) -> ReturnCode {
+        match subscribe_num {
+            0 => self.configure_callback(callback, app_id),
+            _ => ReturnCode::ENOSUPPORT,
+        }
+    }
+
+    /// Command switch statement for various essential processes
+    /// 0 is a sanity check for the switch statement
+    /// 1 enables the qdec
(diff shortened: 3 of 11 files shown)


capsules/src/driver.rs
```diff
@@ -69,6 +69,7 @@ pub enum NUM {
     // Misc
     Buzzer                = 0x90000,
     Screen                = 0x90001,
-    Touch                 = 0x90002
+    Touch                 = 0x90002,
+    Qdec                  = 0x90003
```
ppannuto, line 73:
```suggestion
    Qdec                  = 0x90003,
```

So that future additions don't have to touch unneeded lines.

capsules/src/qdec.rs
```diff
@@ -0,0 +1,108 @@
+//! Provides userspace access to the Qdec on a board.
```
ppannuto, line 1: Can you expand on this a bit? For one, I had to google `QDec` as it's not a shorthand for Quadrature Decoder that I'd seen before.

Is the intent of this capsule exclusively to provide an interface for hardware decoders? This is something that's fairly easy (although possibly compute-intensive) to emulate in software. I think the driver interface here probably can/should be abstract over the type of underlying implementation?

capsules/src/qdec.rs
```diff
@@ -0,0 +1,108 @@
...
+//! let qdec = static_init!(
+//!     capsules::qdec::Qdec<'static>,
+//!     capsules::qdec::QdecInterface::new(&nrf52::qdec::QDEC,
+//!                                         kernel::Grant::create())
+//! );
+//! kernel::hil::QdecDriver.set_client(qdec);
+//! ````
+//! #Interrupt Spurred Readings versus Regular Readings
+//! An application can either enable interrupts to get the
+//! accumulation value or manually read it whenever it wants
```
ppannuto, line 16: This reads as something very specific to the nRF implementation. At this layer, however a capsule is ideally a more generic interface to something providing quadrature decoding. It would be good to find one or two other (non-Nordic) chips and quickly verify that their hardware engines can support this interface (no need to implement, just a datasheet read).

capsules/src/qdec.rs
```diff
@@ -0,0 +1,108 @@
...
+pub struct QdecInterface<'a> {
+    driver: &'a dyn hil::qdec::QdecDriver,
+    apps: Grant<App>,
+}
+
+#[derive(Default)]
+/// This struct contains the necessary fields for an app
+pub struct App {
+    callback: Option<Callback>,
+    pos: u32,
```
ppannuto, line 34:
```suggestion
    position: u32,
```

Generally, Tock style prefers spelled out variable names.

---

A more semantic question, should this be called `position` or `ticks`? Generally position has some tie-in to physical space and implies an absolute position. If a wheel goes forward for a long time, it's meaningful for a number of encoder ticks to wraparound, it's not really meaningful for a position to wrap.

chips/nrf52/src/qdec.rs
```diff
@@ -0,0 +1,278 @@
...
+        LedPol OFFSET(0) NUMBITS(1) [
+            ActiveLow = 0,
+            ActiveHigh = 1
+        ]
+    ],
+    Sample [
+        SAMPLE 1
+    ],
+    SampPer [
+        SAMPLEPER OFFSET(0) NUMBITS(4) [
```
ppannuto, line 114: Several of these are missing doc comments, which are useful for the auto-generated docs.

chips/nrf52/src/qdec.rs
```diff
@@ -0,0 +1,278 @@
...
+        let qdec = Qdec {
+            registers: QDEC_BASE,
+            client: OptionalCell::empty(),
+            state: QdecState::Start,
+        };
+        qdec
+    }
+
+    /// sets pins_a and pins_b to be the output pins for whatever the encoding device is  
+    pub fn set_pins(&self, pin_a: pinmux::Pinmux, pin_b: pinmux::Pinmux) {
```
ppannuto, line 180: Is there some expectation here about which pin is A and B in order to get the direction correct?

chips/nrf52/src/qdec.rs
```diff
@@ -0,0 +1,278 @@
...
+            // For each of 4 possible compare events, if it's happened,
+            // clear it and sort its bit in val to pass in callback
+            for i in 0..regs.events_arr.len() {
+                if regs.events_arr[i].is_set(Event::READY) {
+                    regs.events_arr[i].set(0);
+                    match i {
+                        0 => {
+                            client.sample_ready();
+                        }
+                        1 => { /*No handling for REPORTRDY*/ }
```
ppannuto, line 209: Should this (and 3 below) be `unimplemented!()`, or is this something that will never need to be implemented (if the latter, a bit more exposition in the comment would be helpful)?

kernel/src/hil/qdec.rs
```diff
@@ -0,0 +1,33 @@
...
+//! quadrature decoder. Note this interface is experimental and
+//! may need further updates once implemented on additional chips
+
+use crate::returncode::ReturnCode;
+
+pub trait QdecDriver {
+    /// Sets the client which will receive interrupts
+    fn set_client(&self, client: &'static dyn QdecClient);
+
+    /// Enables the SAMPLERDY interrupt
```
ppannuto, line 13: I think this is too chip-specific for a HIL. What does this interrupt do? When does it fire? Can multiple chips reasonably support this?

kernel/src/hil/qdec.rs
```diff
@@ -0,0 +1,33 @@
...
+    /// Enables the Qdec, returning error if Qdec does not exist
+    fn enable_qdec(&self) -> ReturnCode;
+
+    /// Checks if the qdec has been enabled
+    fn enabled(&self) -> ReturnCode;
+
+    /// Reads the accumulator value and resets it
+    /// Note accumulator means the measure of how many ticks the
+    /// QDEC has moved since the last time the function was called
+    fn get_acc(&self) -> u32;
```
ppannuto, line 25: I'm surprised that this is a `u32`, how is direction encoded?

kernel/src/hil/qdec.rs
```diff
@@ -0,0 +1,33 @@
...
+
+pub trait QdecDriver {
+    /// Sets the client which will receive interrupts
+    fn set_client(&self, client: &'static dyn QdecClient);
+
+    /// Enables the SAMPLERDY interrupt
+    fn enable_interrupts(&self) -> ReturnCode;
+
+    /// Enables the Qdec, returning error if Qdec does not exist
+    fn enable_qdec(&self) -> ReturnCode;
```
ppannuto, line 17: Generally, we try to make hardware presence / failure a compile-time check rather than a runtime check in the kernel. If a kernel can access a QDEC object (as created by the board file), then one must physically exist (otherwise the board should not create it).

kernel/src/hil/qdec.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    /// Checks if the qdec has been enabled
+    fn enabled(&self) -> ReturnCode;
```
ppannuto, line 20: Should there be a corresponding disable method?

boards/nordic/nrf52840dk/src/main.rs
```diff
@@ -101,6 +102,9 @@ const SPI_MX25R6435F_CHIP_SELECT: Pin = Pin::P0_17;
 const SPI_MX25R6435F_WRITE_PROTECT_PIN: Pin = Pin::P0_22;
 const SPI_MX25R6435F_HOLD_PIN: Pin = Pin::P0_23;
 
+const QDEC_PIN_A: Pin = Pin::P0_02;
+const QDEC_PIN_B: Pin = Pin::P0_29;
```
ppannuto, line 106: Is this something that's wired up for all 54820dk's, or something specific to a local testing setup? I don't think we want to wire this up by default unless all boards have something quadrature hanging out at these pins.

ppannuto (changes requested):
TIL there's a quadrature decoder built-in to the nRF -- neat!

First pass, mostly with higher-level comments on the interface. Biggest question is how specific it is to the nRF?

capsules/src/driver.rs
mmurray22: Thanks for the comment! Committed the suggesiton!

kernel/src/hil/qdec.rs
mmurray22: Thanks for the comment! The main motivation for using u32 was the eventual conversion to usize that is done in the capsule because the value passed into the callback schedule function needs to be a usize to my knowledge (When attempting this conversion with an i32, it presented some challenges). Since the value actually provided to userspace is an int, the cast correctly changes the overflow from turned the rotary encoder backwards into negative values. Thus, given the testing done so far, it seems the direction is accurately calculated and communicated to the user. However, if you have further input on how to make this process more bulletproof and reliable then I'd be happy to discuss further!

capsules/src/qdec.rs
mmurray22: Thanks for the query! The value returned is the absolute position relative to the starting location of the rotary encoder. Currently, the value we return does not wrap if the rotary encoder makes a full rotation-it continues ticking up or down (depending on the direction). Do you think this should be updated to cater more to `ticks`?

kernel/src/hil/qdec.rs
mmurray22: Thanks for pointing this out! I will make sure to add a disable function here shortly.

boards/nordic/nrf52840dk/src/main.rs
hudson-ayers: These pins are specific to the local testing setup, however I don't think it is possible to include the QDEC code from main without picking 2 pins. This raises an interesting question, because if the QDEC code is not included in a main somewhere it is likely to rot. Should this pick 2 pins that are not exposed (and therefore unusable)? There is currently no mechanism for selecting pins from userspace AFAIK.

kernel/src/hil/qdec.rs
mmurray22: Upon further consideration, the u32 has now been changed to an i32 to better represent the position value!

chips/nrf52/src/qdec.rs
mmurray22: Thanks for pointing this out! I have now added comments to all sections of the register bitfields.

chips/nrf52/src/qdec.rs
```diff
@@ -0,0 +1,316 @@
...
+                    regs.events_arr[i].set(0);
+                    match i {
+                        0 => {
+                            client.sample_ready();
+                        }
+                        1 => { /*For the time being, there will be no implementation in respone
+                                 to the firing of the REPORTRDY interrupt. It fires too frequently
+                                 in order to utilize it for position measuring. However, by that same 
+                                 token, it cannot merely be marked as unimplemented since it would
+                                 crash the program.*/ }
```
ppannuto, line 247: Since we're just ignoring this interrupt anyway, is it possible to configure the peripheral to not generate `REPORTRDY` interrupts? Right now I think this will be a lot of interrupt load that isn't doing anything in practice.

boards/nordic/nrf52840dk/src/main.rs
ppannuto: I think #1760 has a similar problem; it's code that's only meaningful if you have an external LoRa radio module attached, which no "stock" board will have. As we expand the set of peripherals, we're going to have more and more of these hardware add-ons. It would be nice to have some more structured mechanism to handle this.

kernel/src/hil/qdec.rs
ppannuto: I'm not sure this is documented anywhere except prior HIL discussion threads, but I think we have decided that an `is_enabled` check is a bit an anti-pattern for devices.

Callers should know whether the device is enabled or not (i.e. if you intend to use it, you need to enable it first) and if there are multiple users, then there should be a virtualization layer that multiplexes hardware access and handles {en/dis}able correctly. We have the `EOFF` return code for the case where there is a runtime bug as well.

In summary: I think we should remove the `enabled()` method, and add a `disable` method.

kernel/src/hil/qdec.rs
mmurray22: Thanks for comment! I will update the comments/function names to prevent the HIL from being too closely tied with the nRF. I am currently looking into whether the current QDEC setup is compatible with other chips.

kernel/src/hil/qdec.rs
mmurray22: Good point!`enabled()` has been removed from the HIL

capsules/src/qdec.rs
mmurray22: Thanks for the comment! I will update the comment to include a brief summary of what a QDEC is, as well as a note that Quadrature Decoder is the full name.  

As for the second part of your comment, this capsule does provide an interface for hardware decoders (specifically those that decode quadrature encoded signals), so it should be abstract over the underlying implementations/chips it is applied to. Are there sections of the capsule that look too nRF specific?

capsules/src/qdec.rs
mmurray22: As of right now, I have looked at the SAM 3/4 chip and it seem its QDEC on board does behave differently than the nRF. The datasheet for that chip's QDEC can be found [here](http://ww1.microchip.com/downloads/en/AppNotes/Atmel-42706-Quadrature-Decoder-QDEC-for-SAM3-4-Devices_ApplicationNote_AT11483.pdf). I'm not sure how to proceed from this-I'm currently looking at other chips to see if it is the same situation for them (particularly chips that support tockOS already). Right now, it seems most don't have a QDEC.

chips/nrf52/src/qdec.rs
mmurray22: Pin A should be Pin02 and Pin B should be Pin029. If they are put backwards than the direction will be switched. Should I put that in this file?

chips/nrf52/src/qdec.rs
mmurray22: Right now, the current interpretation of the datasheet makes REPORTRDY a critical interrupt to have on because that's how the samples will be generated.

boards/nordic/nrf52840dk/src/main.rs
hudson-ayers: I suppose we could resolve this temporarily using the mechanism we use for kernel tests -- put the code that would call this in a qdec_test.rs file, which picks pins and sets up the qdec peripheral and creates a qdec capsule. Then include this file with an `#[allow(unused_code)]` annotation so that code is always built, but not actually run unless it is included from main. The downside is that there is still a lot of friction to actually using it in main, because it will have to be added to the syscall mapping and the Platform struct, and the interrupt mapping will have to be made.

kernel/src/hil/qdec.rs
```diff
@@ -0,0 +1,33 @@
...
+//! may need further updates once implemented on additional chips
+
+use crate::returncode::ReturnCode;
+
+pub trait QdecDriver {
+    /// Sets the client which will receive interrupts
+    fn set_client(&self, client: &'static dyn QdecClient);
+
+    /// Enables the interrupt collecting and storing samples
+    fn enable_interrupts(&self) -> ReturnCode;
```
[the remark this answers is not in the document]
bradjc: Actually, should these be `enable_callbacks()` and `disable_callbacks()`? Certain hardware may need to use interrupts to implement the accumulator function, even if the client doesn't want any callbacks. I think what really matters here from an API standpoint is does the client get notified. Then if the hardware doesn't need interrupts to implement accumulate then that implementation can disable interrupts.

kernel/src/hil/qdec.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    /// Reads the accumulator value and resets it
+    /// Note accumulator means the measure of how many ticks the
+    /// QDEC has moved since the last time the function was called
+    fn get_acc(&self) -> i32;
+}
+
+pub trait QdecClient {
+    /// Indicate to the client that the status of the accumulator has changed
+    fn sample_ready(&self);
```
bradjc, line 30: Why doesn't this include an argument indicating the direction of the new sample?

```suggestion
    pub enum Direction {
        Increment,
        Decrement,
    }

    fn changed(&self, direction: Direction);
```

bradjc:
Moved to #2345

PR #2115, opened 2020-09-17
opentitan: Disable USB support by default

alistair23:
### Pull Request Overview

Due to an OpenTitan hardware bug (see lowRISC/opentitan#2598)
USB doesn't work. Let's just disable USB by default to avoid confusing
people.

This applies on top of: https://github.com/tock/tock/pull/2094

### Testing Strategy

N/A

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/components/src/ctap.rs
+++ b/boards/components/src/ctap.rs
@@ -1,0 +1,128 @@
+//! Component for CTAP HID over USB support.
+//!
+//! This provides a component for using the CTAP driver. This allows for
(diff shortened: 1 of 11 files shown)


alistair23:
@phil-levis this PR disables USB support in OT.

alistair23:
Do we not still wan this? I thought everyone was on board with disabling USB support by default as the master branch of OT is broken?

Although it looks like the fix is progressing and will hopefully be merged into OT soon.

Issue #1306, opened 2019-04-25
Buggy behavior in cortexm NVIC


phil-levis:
https://github.com/tock/tock/blob/85d066d678002a5069fb770170bdeb2d8b1c08e4/arch/cortex-m/src/nvic.rs#L59

I ran into this bug on H1B. I'd argue the expected behavior is that if you disable an interrupt (NVIC number), then there won't be subsequent triggers of the interrupt handler. However, that's not what happens. Instead, whether the handler is trigger is independent of whether the interrupt is disabled or enabled, since this code traverses the pending bits (a disabled interrupt is still pending, it just doesn't actually trigger the ISR).

However, if you disable *all* interrupts, then all interrupts are disabled. This is because if *any* interrupt fires, then it will trigger *all* pending interrupts, whether they are enabled or disabled.

One simple fix is to check both whether the interrupt is pending and enabled. This will add cycles, though.

alevy:
I doubt the cycles would be significant. It should be just indexing into like two or three extra instructions (just indexing into the same word of the IER and taking a bitwise intersection).

Seems worth adding. It's kind of not "expected" behavior (which is why we didn't expect it :) ), but totally could happen and probably should be handled.

phil-levis:
I lost a few hours trying to figure out why a disabled interrupt continued to be delivered, but if I disabled all interrupts then none were delivered...

phil-levis:
_I doubt the cycles would be significant._ 

The number of instructions is probably small, but these are interrupt handlers; they need to be fast for energy and performance reasons. This is one of the places we want to be most careful about bloat.

bradjc:
Documenting from a recent discussion:

The approach of checking if an interrupt is pending and enabled won't work in Tock because Tock's interrupt handling approach is to disable interrupts in the interrupt controller immediately after they fire (so they don't trigger the ISR again). Later when the kernel is ready to service the interrupts it checks if the interrupt pending bit is set, and if the kernel also checked the enable bit it would always be false.

phil-levis:
It sounds like we need to keep track of this in software, then.

PR #2116, opened 2020-09-18
chips: e310: improve interrupt handling

bradjc:
### Pull Request Overview

This pull request changes the e310 chip to:

1. Not only look at the plic, but also the MIE register for if interrupts have occurred.
2. Not accidentally disable interrupts when servicing them.

This fixes timer issues on the hifive1b.

### Testing Strategy

Running the alarm multi test on hifive1b hardware.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/e310x/src/chip.rs
+++ b/chips/e310x/src/chip.rs
@@ -2,7 +2,6 @@
 
 use core::fmt::Write;
 use kernel;
-use kernel::common::registers::FieldValue;
 use kernel::debug;
 use kernel::hil::time::Alarm;
 use rv32i;
@@ -76,8 +75,6 @@ impl<A: 'static + Alarm<'static>> kernel::Chip for E310x<A> {
     }
 
     fn service_pending_interrupts(&self) {
-        let mut reenable_intr = FieldValue::<u32, mie::Register>::new(0, 0, 0);
-
         loop {
             let mip = CSR.mip.extract();
 
@@ -85,13 +82,11 @@ impl<A: 'static + Alarm<'static>> kernel::Chip for E310x<A> {
                 unsafe {
                     timer::MACHINETIMER.handle_interrupt();
                 }
-                reenable_intr += mie::mtimer::SET;
             }
             if mip.is_set(mip::mext) {
                 unsafe {
                     Self::handle_plic_interrupts();
                 }
-                reenable_intr += mie::mext::SET;
             }
 
             if !mip.matches_any(mip::mext::SET + mip::mtimer::SET) {
@@ -99,12 +94,13 @@ impl<A: 'static + Alarm<'static>> kernel::Chip for E310x<A> {
             }
         }
 
-        // re-enable any interrupt classes which we handled
-        CSR.mie.modify(reenable_intr);
+        // Re-enable all MIE interrupts that we care about. Since we looped
+        // until we handled them all, we can re-enable all of them.
+        CSR.mie.modify(mie::mext::SET + mie::mtimer::SET);
     }
 
     fn has_pending_interrupts(&self) -> bool {
-        unsafe { plic::has_pending() }
+        CSR.mip.matches_any(mip::mext::SET + mip::mtimer::SET)
     }
 
     fn sleep(&self) {


chips/e310x/src/chip.rs
```diff
@@ -76,35 +75,32 @@ impl<A: 'static + Alarm<'static>> kernel::Chip for E310x<A> {
...
-        // re-enable any interrupt classes which we handled
-        CSR.mie.modify(reenable_intr);
+        // Re-enable all MIE interrupts that we care about. Since we looped
+        // until we handled them all, we can re-enable all of them.
+        CSR.mie.modify(mie::mext::SET + mie::mtimer::SET);
     }
 
     fn has_pending_interrupts(&self) -> bool {
-        unsafe { plic::has_pending() }
+        CSR.mip.matches_any(mip::mext::SET + mip::mtimer::SET)
```
ppannuto, line 103: This change confuses me, will this match list need to grow as more interrupts are supported?

ppannuto (commented):
The interrupt handling change looks good and makes sense, one other question though

chips/e310x/src/chip.rs
ppannuto: Specifically, your comment said
> Not only look at the plic, but also the MIE register for if interrupts have occurred.
But this change seems to look at MIE instead of rather than in addition to?

bradjc: PLIC == `mext`

ppannuto (approved):

chips/e310x/src/chip.rs
```diff
@@ -76,35 +75,32 @@ impl<A: 'static + Alarm<'static>> kernel::Chip for E310x<A> {
...
             if !mip.matches_any(mip::mext::SET + mip::mtimer::SET) {
                 break;
             }
         }
 
-        // re-enable any interrupt classes which we handled
-        CSR.mie.modify(reenable_intr);
+        // Re-enable all MIE interrupts that we care about. Since we looped
+        // until we handled them all, we can re-enable all of them.
+        CSR.mie.modify(mie::mext::SET + mie::mtimer::SET);
```
alistair23, line 99: This doesn't seem right as we disable the mtimer when we want to disable the alarm: https://github.com/tock/tock/blob/master/arch/rv32i/src/machine_timer.rs#L141

chips/e310x/src/chip.rs
```diff
@@ -76,35 +75,32 @@ impl<A: 'static + Alarm<'static>> kernel::Chip for E310x<A> {
     }
 
     fn service_pending_interrupts(&self) {
-        let mut reenable_intr = FieldValue::<u32, mie::Register>::new(0, 0, 0);
```
alistair23, line 79: This change also needs to happen to OpenTitan, which uses the same logic.

chips/e310x/src/chip.rs
alistair23: This looks good. Can OT have the same change as well?

alistair23 (dismissed):

chips/e310x/src/chip.rs
bradjc: But the e310 platform (hifive) isn't using that SchedulerTimer implementation. I agree they aren't compatible, but we can't have the scheduler timer changing interrupts for the only timer source we have.

alistair23: If someone adds a RISC-V board that does have other timers they are going to look at this code (and OT which is the same) as an example. I don't think we should be enabling timer interrupts if they weren't already enabled AND have code that disables them and expects them to stay disabled..

Ideally I think it would be best to record that we disabled them handling an interrupt and then only enable them if that's true.

bradjc: I actually think that is more confusing, since Tock's design is all interrupts are enabled at the NVIC/MIE/PLIC layer, but putting that aside, I don't actually know how to write that code in a way that is both correct and I would be ok with someone else copying.

chips/e310x/src/chip.rs
phil-levis: Let's make that a separate PR?

chips/e310x/src/chip.rs
phil-levis: But this (always re-enabling mext and mtimer) is clearly incorrect -- what if the kernel disabled mtimer (for whatever reason)? When there is only a single interrupt source, the scheduler timer should not be disabling its interrupts; it goes through a virtualizer. The virtualizer *may* disable underlying interrupts if there are no outstanding alarms. Breaking abstraction boundaries (e.g., who owns mtimer) in random ways due to software engineering issues is bad.

phil-levis: This is especially pernicious for RISC-V, as the mtimer interrupt will fire continually on <= rather than only fire once on ==.

bradjc: Right now in Tock the kernel cannot dynamically disable the mtimer interrupt (it could choose not to use it at all, and it can disable it when it is also pending, but the design as I understand it is all interrupts at the NVIC/PLIC/CLIC/CLINT/MIE layer are always enabled unless pending).

alistair23: Even if it is always enabled, blinding force enabling it still seems wrong. What if someone is trying to debug something and wants the timer disabled? This will force enable it after it's disabled. There could also be future code changes that allow it to be disabled in the future.

phil-levis: > but the design as I understand it is all interrupts at the NVIC/PLIC/CLIC/CLINT/MIE layer are always enabled unless pending).

Nope. This would mean you cannot suppress interrupts and therefore wakeup. Super-bad.

Instead, the current design -- which is fundamentally flawed -- is that you can suppress interrupts from occurring/waking up the MCU but cannot suppress their handling. I.e, #1306  . https://github.com/tock/tock/blob/ad9387a577405675b044d5bde85badf0274995c8/arch/cortex-m/src/nvic.rs#L167 and https://github.com/tock/tock/blob/ad9387a577405675b044d5bde85badf0274995c8/arch/cortex-m/src/nvic.rs#L109 let you disable interrupts in the CortexM, NVIC, for example.

The CortexM semantics are a problem for all kinds of reasons. But it's even *more* of a problem is the semantics aren't even consistent across architectures!

chips/e310x/src/chip.rs
alistair23: Fine with me, as long as it doesn't get lost. They are pretty much exactly the same code.

chips/e310x/src/chip.rs
alevy: I agree with @bradjc on this: it is currently how we are treating interrupts. I agree with @phil-levis and @alistair23 that is not necessarily the _right_ way to be dealing with interrupts, but that _is_ the architecture we're using in Tock atm.

And generally, you _cannot_ currently disable interrupts from waking the CPU: `nvic::disable` is an instance method of a struct that requires `unsafe` to construct and is handed to **nobody** and `nvic::disable_all` is `unsafe`.

alistair23: I don't understand why it would be too hard to just record that we disabled the interrupt in interrupt handling and then only re-enable interrupts if that happened. What am I missing here?

hudson-ayers:
This is one of two remaining blockers for 1.6. I propose that we have two options here:

1. Modify this PR to use Alistair's approach of saving and re-enabling interrupts. This is probably a better design, but also  important to think about carefully as it has been a source of bugs in the past, and is not necessarily aligned with the spirit of how Tock treats interrupts today.

2. Leave this PR as-is, but delete the lines of the machine timer `SchedulerTimer` implementation that turn off the machine timer interrupt enable (this is okay, because the scheduler timer trait specifies that `arm()`/`disarm()` are optional optimizations). This approach keeps things in line with how the Tock interrupt architecture currently works, and gets us to a release sooner. 

I am in favor of option 2, combined with @alistair23 or @phil-levis submitting a separate PR for review after the release that introduces the save/restore approach on all risc-v platforms, as well as possibly any changes on Cortex-M to allow the kernel to disable interrupts.

bradjc:
I'm writing code that only re-enables interrupts that were serviced, instead of all (both) interrupts unconditionally. I can't say I'm a fan of the code. Here is my function:

```rust
fn service_pending_interrupts(&self) {
        let mut serviced_interrupt_mtimer: bool = false;
        let mut serviced_interrupt_mext: bool = false;

        loop {
            let mip = CSR.mip.extract();

            if mip.is_set(mip::mtimer) {
                unsafe {
                    timer::MACHINETIMER.handle_interrupt();
                }
                serviced_interrupt_mtimer = true;
            }
            if mip.is_set(mip::mext) {
                unsafe {
                    Self::handle_plic_interrupts();
                }
                serviced_interrupt_mext = true
            }

            if !mip.matches_any(mip::mext::SET + mip::mtimer::SET) {
                break;
            }
        }

        // Enable the interrupts that we serviced. Since the interrupts we
        // serviced were pending, we assume they should be re-enabled and are
        // not intended to be disabled for some reason.
        if serviced_interrupt_mtimer {
            CSR.mie.modify(mie::mtimer::SET);
        }
        if serviced_interrupt_mext {
            CSR.mie.modify(mie::mext::SET);
        }
    }
```

First, this still goes against Tock's current interrupt model where all PLIC/NVIC interrupts and above are always enabled. Second, without a _clear_ description somewhere of what the interrupt model is exactly (if not our current model), this change can lead to some weird behavior if different layers can assume they can disable interrupts. For example, what if the peripheral does the (I would argue reasonable) thing of disabling the interrupt in `handle_interrupt()` after handling the interrupt? This code will then immediately re-enable it. At least with the current version in this PR there is no pretense that other layers can disable the interrupt.

bradjc:
Implementing a new interrupt model for Tock is a bigger change, and one that I don't think this PR should start on.

hudson-ayers:
I submitted #2134 which removes the peripheral code that directly disabled the mtimer interrupt.

hudson-ayers (approved):

alevy (approved):

PR #2118, opened 2020-09-21
stm32f3: Watchdog Timers

krady21:
### Pull Request Overview

This pull request adds both watchdogs supported by the stm32f3 boards. The difference between the two of them is thoroughly described [here](https://electronics.stackexchange.com/questions/123080/independent-watchdog-iwdg-or-window-watchdog-wwdg). TLDR: The independent watchdog  is clocked by its own dedicated low-speed clock, but is not as precise, while the window watchdog is more precise, has a configurable time window that can be used to detect early or late abnormalities and can also generate an interrupt just before resetting.

At this point, someone could configure the board to use one, both or neither of the two watchdogs. I don't know if there's need for both, but i wanted to consult with you first before deleting one or the other. I also tried to have this configuration done in the boards file, as it was discussed in a previous [pr](https://github.com/tock/tock/pull/1887#issuecomment-654236192), but I am not entirely satisfied with how i did it.

### Testing Strategy

This pull request was tested using the stm32f3discovery board.

### TODO or Help Wanted

The main problem with both of the watchdogs is that none of them provides a way to suspend them once they are started, except by a full system reset. Since suspend and resume functions are unimplemented, sleeping in the kernel_loop [function](https://github.com/tock/tock/blob/ad9387a577405675b044d5bde85badf0274995c8/kernel/src/sched.rs#L495-L514) will probably end up causing a watchdog reset.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/stm32f3discovery/src/main.rs
+++ b/boards/stm32f3discovery/src/main.rs
@@ -713,7 +713,7 @@ pub unsafe fn reset_handler() {
         debug!("{:?}", err);
     });
 
-    chip.enable_iwdg();
+    chip.enable_watchdog();
     let scheduler = components::sched::round_robin::RoundRobinComponent::new(&PROCESSES)
         .finalize(components::rr_component_helper!(NUM_PROCS));
 
--- a/boards/stm32f3discovery/src/main.rs
+++ b/boards/stm32f3discovery/src/main.rs
@@ -713,6 +713,7 @@ pub unsafe fn reset_handler() {
         debug!("{:?}", err);
     });
 
+    chip.enable_iwdg();
     let scheduler = components::sched::round_robin::RoundRobinComponent::new(&PROCESSES)
         .finalize(components::rr_component_helper!(NUM_PROCS));
 
--- a/boards/stm32f3discovery/src/main.rs
+++ b/boards/stm32f3discovery/src/main.rs
@@ -713,7 +713,9 @@ pub unsafe fn reset_handler() {
         debug!("{:?}", err);
     });
 
-    chip.enable_watchdog();
+    // Uncomment this to enable the watchdog
+    // chip.enable_watchdog();
+
     let scheduler = components::sched::round_robin::RoundRobinComponent::new(&PROCESSES)
         .finalize(components::rr_component_helper!(NUM_PROCS));
 
--- a/chips/stm32f303xc/src/chip.rs
+++ b/chips/stm32f303xc/src/chip.rs
@@ -14,11 +14,13 @@ use crate::nvic;
 use crate::spi;
 use crate::tim2;
 use crate::usart;
+use crate::wdt;
 
 pub struct Stm32f3xx {
     mpu: cortexm4::mpu::MPU,
     userspace_kernel_boundary: cortexm4::syscall::SysCall,
     scheduler_timer: cortexm4::systick::SysTick,
+    watchdog: wdt::Wdt,
 }
 
 impl Stm32f3xx {
@@ -27,7 +29,16 @@ impl Stm32f3xx {
             mpu: cortexm4::mpu::MPU::new(),
             userspace_kernel_boundary: cortexm4::syscall::SysCall::new(),
             scheduler_timer: cortexm4::systick::SysTick::new(),
+            watchdog: wdt::Wdt::new(),
         }
+    }
+
+    pub fn enable_wwdg(&self) {
+        self.watchdog.enable_windo();
+    }
+
+    pub fn enable_iwdg(&self) {
+        self.watchdog.enable_indep();
     }
 }
 
@@ -35,7 +46,7 @@ impl Chip for Stm32f3xx {
     type MPU = cortexm4::mpu::MPU;
     type UserspaceKernelBoundary = cortexm4::syscall::SysCall;
     type SchedulerTimer = cortexm4::systick::SysTick;
-    type WatchDog = ();
+    type WatchDog = wdt::Wdt;
 
     fn service_pending_interrupts(&self) {
         unsafe {
@@ -46,6 +57,7 @@ impl Chip for Stm32f3xx {
                     }
                 } else if let Some(interrupt) = cortexm4::nvic::next_pending() {
                     match interrupt {
+                        nvic::WWDG => wdt::WATCHDOG.windo_wdg.handle_interrupt(),
                         nvic::USART1 => usart::USART1.handle_interrupt(),
 
                         nvic::TIM2 => tim2::TIM2.handle_interrupt(),
@@ -94,7 +106,7 @@ impl Chip for Stm32f3xx {
     }
 
     fn watchdog(&self) -> &Self::WatchDog {
-        &()
+        &self.watchdog
     }
 
     fn userspace_kernel_boundary(&self) -> &cortexm4::syscall::SysCall {
--- a/chips/stm32f303xc/src/chip.rs
+++ b/chips/stm32f303xc/src/chip.rs
@@ -53,8 +53,6 @@ impl Chip for Stm32f3xx {
                     }
                 } else if let Some(interrupt) = cortexm4::nvic::next_pending() {
                     match interrupt {
-                        nvic::WWDG => wdt::WATCHDOG.handle_interrupt(),
-
                         nvic::USART1 => usart::USART1.handle_interrupt(),
 
                         nvic::TIM2 => tim2::TIM2.handle_interrupt(),
--- a/chips/stm32f303xc/src/chip.rs
+++ b/chips/stm32f303xc/src/chip.rs
@@ -20,7 +20,7 @@ pub struct Stm32f3xx {
     mpu: cortexm4::mpu::MPU,
     userspace_kernel_boundary: cortexm4::syscall::SysCall,
     scheduler_timer: cortexm4::systick::SysTick,
-    watchdog: wdt::Wdt,
+    watchdog: wdt::WindoWdg,
 }
 
 impl Stm32f3xx {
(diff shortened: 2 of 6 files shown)


hudson-ayers:
The watchdog firing when the board sleeps seems like a significant issue. I thought about setting the pre-scalar and reset values to the maximum values in `disable()`, but for the IWDG that gives a maximum sleep time of 26 seconds, which is no good. 

For the WWDG it seems that it is effectively possible to disable by disabling the APB1 source clock, but that is not super helpful because I don't know what negative impact that could have on other peripherals that use APB1 as a source clock.

Perhaps use of a global variable and a specialized watchdog top-half interrupt handler could work...something like:

```rust
static mut WATCHDOG_SLEEP_FLAG: bool = false;

impl Watchdog ... {
    suspend(&self) {
        unsafe { WATCHDOG_SLEEP_FLAG = true }
    }

    resume(&self) {
        unsafe { WATCHDOG_SLEEP_FLAG = true }
    }
}

#[naked]
pub unsafe extern "C" fn watchdog_handler() {
    llvm_asm!(
        "
    ld r0, WATCHDOG_SLEEP_FLAG
    // if 0, call actual handler, else call wfi
    );
}
```

Not sure how that approach would be received for inclusion upstream though.

alistair23:
That's crazy! Why can you not suspend the WDT.

Wouldn't it be best to just set the WDT timer as long as possible when suspending and call an alarm to wake up and reset it?

krady21:
First of all, thank you for the feedback and suggestions. 
> That's crazy! Why can you not suspend the WDT.
It is kind of ludicrous, but this seems to be the case for a lot of the watchdogs in st boards.

> For the WWDG it seems that it is effectively possible to suspend by disabling the APB1 source clock, but that is not super helpful because I don't know what negative impact that could have on other peripherals that use APB1 as a source clock.
Yes, that would probably yield more problems for the rest of the peripherals using this clock. 

> Wouldn't it be best to just set the WDT timer as long as possible when suspending and call an alarm to wake up and reset it?
At the moment, this probably seems to be the best course of action. Does constantly entering and leaving the low power mode (because of the alarm wakeup) affect performance in any serious way?

alistair23:
Obviously it would be better not to have to wake up just to service the WDT, but it sounds like that isn't possible. I wonder what other users of STM do, maybe it's worth checking Zephyr?

If there is a better option I'm happy to work on changing the HIL, I just can't think of anything better.

Waking up every 25 seconds (Hudson mentioned the longest WDT time as 26 seconds) isn't ideal and it's hard to tell what impact on power that will have. You would probably have to benchmark it. Maybe it then becomes a board designer decision, what is more important, low power or WDT? It really would depend on your use case though. If you are waking up every 5 seconds for some other task, then it probably doesn't matter. If you wake up once every 24 hours, then it will matter more.

In reality though, Tock isn't super low power in general. It for example doesn't turn off devices that aren't currently being used, so I don't know how much of an overall impact this will have.

Ideally we only want to set the alarm while sleeping, and then disable it when we wakeup. So it shouldn't impact running performance.

krady21:
After going one more time over @hudson-ayers 's response, I realized that I overlooked one important thing. Getting an early wakeup interrupt from the window watchdog actually takes you out of sleep mode (because you entered through wfi). This allows tickling the watchdog in the reset handler, based on whether the WATCHDOG_SLEEP_FLAG is set similar to the above mentioned answer. This effectively removes the need for an alarm with the observation that I will still need to do some benchmarking, as you have mentioned.

hudson-ayers:
> After going one more time over @hudson-ayers 's response, I realized that I overlooked one important thing. Getting an early wakeup interrupt from the window watchdog actually takes you out of sleep mode (because you entered through wfi). This allows tickling the watchdog in the reset handler, based on whether the WATCHDOG_SLEEP_FLAG is set, similar to the above mentioned answer. This effectively removes the need for an alarm with the observation that I will still need to do some benchmarking, as you mentioned.

By reset handler I assume you mean watchdog handler? Because `reset_handler()` is only called at boot or after the board is reset

krady21:
My bad. Meant to say interrupt handler. I'll edit it.

chips/stm32f303xc/src/chip.rs
```diff
@@ -46,6 +53,8 @@ impl Chip for Stm32f3xx {
                     }
                 } else if let Some(interrupt) = cortexm4::nvic::next_pending() {
                     match interrupt {
+                        nvic::WWDG => wdt::WATCHDOG.handle_interrupt(),
+
```
hudson-ayers, line 57: `service_pending_interrupts` will never be called if the kernel hangs

hudson-ayers (changes requested):

chips/stm32f303xc/src/wdt.rs
```diff
@@ -0,0 +1,172 @@
...
+        // t_WWDG = (1 / 36000) * 4096 * 2^3 * (63 + 1) = 58ms
+        self.registers.cr.modify(Control::WDGA::SET);
+    }
+
+    pub fn tickle(&self) {
+        self.registers.cr.modify(Control::T.val(0x7F));
+    }
+}
+
+impl kernel::watchdog::WatchDog for WindoWdg {
```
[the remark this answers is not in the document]
krady21: I think I now understand your concerns regarding the top and bottom half split. After some testing, it also seems that the bottom half never gets executed, even with the biggest prescaler value(3) which would leave you with 8 clock cycles until reset.

kernel/src/platform/watchdog.rs
```diff
@@ -31,5 +31,9 @@ pub trait WatchDog {
     }
 }
 
+pub trait WatchdogClient {
+    fn reset_happened(&self) {}
+}
+
```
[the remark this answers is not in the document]
krady21: My idea, altough not implemented yet, was to have this trait used by the client to do some logging or initiate some corrective action, but probably it was a bit too bold and as you've said, you could properly only handle it in the top half. I'll delete it for the time being.

krady21:
Well, after some testing it seems that enabling and disabling the watchdog clock with the RCC_APB1ENR would make suspend and resume work exactly as we wished in the current kernel_loop context (although ST probably doesn't want you to use their watchdog this way). I think it doesn't affect the other peripherals using APB1, as each of them has a specific field for enabling or disabling their connection to their respective clock source. In this case, I don't know if the interrupt is needed anymore, unless we do want to do some logging before reset with it (or take some other action).
And sorry for changing the approach once again.

chips/stm32f303xc/src/wdt.rs
```diff
@@ -164,9 +160,13 @@ impl kernel::watchdog::WatchDog for WindoWdg {
...
-            unsafe {
-                WATCHDOG_SLEEP_FLAG = true;
-            }
+            self.clock.disable();
+        }
+    }
+
+    fn resume(&self) {
+        if self.enabled.get() {
+            self.clock.enable();
```
hudson-ayers, line 169: `resume()` should probably also tickle the watchdog

chips/stm32f303xc/src/wdt.rs
```diff
@@ -81,33 +80,10 @@ impl WindoWdg {
...
-        }
-    }
-
     /// This interrupt is only cleared by hardware after a reset.
-    fn enable_interrupt(&self) {
-        self.registers.cfr.modify(Config::EWI::SET);
-    }
+    // fn enable_interrupt(&self) {
+    //     self.registers.cfr.modify(Config::EWI::SET);
+    // }
```
hudson-ayers, line 86: Where is this interrupt enabled now?

hudson-ayers (commented):
I like this approach much more, thanks for investigating further. A couple small comments. 

I also like that this leaves enabling the watchdog up to the board.

chips/stm32f303xc/src/wdt.rs
krady21: I was intending to delete this, as I am not sure what to do when we receive the interrupt (if we activate it at all).

hudson-ayers: So the reset still happens even if the interrupt is disabled?

chips/stm32f303xc/src/wdt.rs
krady21: I don't think it is absolutely necessary to tickle it here. The watchdog value is not decremented during sleep because the clock is disabled right before it, in the suspend function. More so, the next thing that happens exactly after resume is actually the first tickle at the beginning of the loop block. I've got no problem also adding one here too, tho.

hudson-ayers: oh, that makes sense, I think it is fine either way then.

chips/stm32f303xc/src/wdt.rs
krady21: yes

alistair23 (dismissed):

boards/stm32f3discovery/src/main.rs
```diff
@@ -713,6 +713,7 @@ pub unsafe fn reset_handler() {
         debug!("{:?}", err);
     });
 
+    chip.enable_watchdog();
```
hudson-ayers, line 716: I don't think we want to enable the watchdog by default for any of our upstream boards.

krady21: Should I comment or delete it outright?

hudson-ayers: Commenting is good!

hudson-ayers (approved):

bradjc (approved):

PR #2120, opened 2020-09-22
arch/cortex-m3: Allow the kernel to access protected memory

alistair23:
### Pull Request Overview

Commit 128782dad5ad410d3 "mpu: Change the disable_mpu API" converted the
disable_app_mpu() function to not make any changes. This means when we
return from an app to the kernel we don't disable the MPU.

This resulted in some access failures when the kernel tried to access
direct write flash regions (https://github.com/tock/tock/pull/1873#issuecomment-696217269).

We can either disable the MPU when returning to the kernel or change the
app MPU configuration permissions to not impact the kernel.

This patch converts the Cortex-M3 MPU implementation to always allow the
kernel access when configuring the app. This way we can avoid the
overhead of disabling the MPU on context switches. There is no security
gap here as the kernel could just disable the MPU anyway if it was
malicious.

### Testing Strategy

None.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m3/src/mpu.rs
+++ b/arch/cortex-m3/src/mpu.rs
@@ -274,16 +274,18 @@ impl CortexMRegion {
                 RegionAttributes::AP::ReadWrite,
                 RegionAttributes::XN::Disable,
             ),
-            mpu::Permissions::ReadExecuteOnly => {
-                (RegionAttributes::AP::ReadOnly, RegionAttributes::XN::Enable)
-            }
+            mpu::Permissions::ReadExecuteOnly => (
+                RegionAttributes::AP::UnprivilegedReadOnly,
+                RegionAttributes::XN::Enable,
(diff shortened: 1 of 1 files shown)


bradjc (dismissed):
Looks good to me. Although I'm a little confused how the m3/m4 can differ since they are symlinked to the same file.

alistair23:
Ha!

I didn't realise they were symlinked, I checked the M4 after I changed the M3. Ignore that part then.

ppannuto (approved):
Looks good to me too, thanks for the quick turnaround on this

hudson-ayers (dismissed):

alistair23:
Fixed the format failure.

hudson-ayers (approved):

bradjc (approved):

PR #2122, opened 2020-09-23
Board based instantiation of chip drivers and interrupt mappings: Earlgrey/Opentitan

hudson-ayers:
### Pull Request Overview

This pull request applies the new, non-global based peripheral design to the earlgrey chip and OpenTitan board. Rather than apply this on top of #2069 / #2084 , the first commit just duplicates the kernel changes from the original PR, so that it is easy to review only the earlgrey/opentitan changes in this PR.

### Testing Strategy

This pull request was tested by compiling. Testing in verilator or on the FPGA board is needed before this is merged.

### TODO or Help Wanted

N/A

### Documentation Updated

- [ ] Docs updated in the other PRs associated with this change.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/earlgrey/src/chip.rs
+++ b/chips/earlgrey/src/chip.rs
@@ -5,36 +5,95 @@ use core::hint::unreachable_unchecked;
 use kernel;
 use kernel::debug;
 use kernel::hil::time::Alarm;
-use kernel::Chip;
+use kernel::{Chip, InterruptService};
 use rv32i::csr::{mcause, mie::mie, mip::mip, mtvec::mtvec, CSR};
 use rv32i::syscall::SysCall;
 use rv32i::PMPConfigMacro;
 
 use crate::chip_config::CONFIG;
-use crate::flash_ctrl;
-use crate::gpio;
-use crate::hmac;
 use crate::interrupts;
 use crate::plic;
-use crate::pwrmgr;
-use crate::timer;
-use crate::uart;
-use crate::usbdev;
 
 PMPConfigMacro!(4);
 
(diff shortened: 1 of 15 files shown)


chips/earlgrey/src/chip.rs
```diff
@@ -48,28 +100,11 @@ impl<A: 'static + Alarm<'static>> EarlGrey<A> {
...
-                        || pwrmgr::PWRMGR.check_clock_propagation(),
-                        None,
-                    );
-                }
-                _ => debug!("Pidx {}", interrupt),
+            if interrupt == interrupts::PWRMGRWAKEUP {
+                self.pwrmgr.handle_interrupt();
+                self.check_until_true_or_interrupt(|| self.pwrmgr.check_clock_propagation(), None);
+            } else if !self.plic_interrupt_service.service_interrupt(interrupt) {
+                debug!("Pidx {}", interrupt);
```
alistair23, line 107: Why do we have the two levels of interrupts?

hudson-ayers: The way the chip implementation works now, power manager interrupts must be handled for the chip to work at all, so it isn't necessary for the power manager to be part of the `DefaultPeripherals` struct because any reimplementation would *have* to include it in order to use the other chip code. Also, calling `check_until_true_or_interrupt()` requires a reference to the Chip, and I didn't want to create a circular dependency here.

alistair23 (approved):

phil-levis:
@hudson-ayers -- there are now conflicts, can you resolve them?

alistair23 (approved):

bradjc (approved):

PR #1543, opened 2020-01-15
Add support for USB bulk/interrupt on the nrf52 chip.

gendx:
### Pull Request Overview

This pull request updates the USB HIL and implements it on the nrf52 chip. There are still unimplemented parts (isochronous endpoints, handling of "USB events"), but I tested it on the USB [echo app](https://github.com/tock/libtock-c/tree/master/examples/tests/usb) for bulk endpoints, and some interrupt endpoints (example not public yet, though the logic in the chip is pretty much the same for interrupt and bulk).

Notable changes to the HIL in order to support this chip as well as interrupt endpoints:
- The `UsbController` and `Client` traits now take a lifetime parameter. This is ultimately to use lifetime-bound buffers, as previously discussed in https://github.com/tock/tock/pull/1427.
- Unified the `endpoint_*_enable` methods to take the transfer type as parameter. This avoids code duplication down the road, as interrupt and bulk endpoints are handled very similarly by the nrf52 chip. Similarly, `bulk_*` methods in the USB client are replaced by `packet_*` methods. And `Bulk(In|Out)Result` enums are generalized to interrupts as well.
- Added a possibility to enable an endpoint both for IN and OUT directions. Some protocols such as USB-HID have such bi-directional endpoints.
- Split `endpoint_bulk_resume` into `endpoint_resume_in` and `endpoint_resume_out` functions, to specify which direction should be resumed, given that an endpoint can now be both IN and OUT directions.
- Due to the way nrf52 triggers USB events, I had to remove the `delayed_in` check in `capsules/src/usb/usbc_client.rs`, otherwise the driver never initiates IN transactions (from the device to the host). I had a look at the sam4l code and this doesn't seem problematic ("superfluous" resumes are ignored).

I propagated these changes on the sam4l, but not implementing the new combinations there as I don't have a board to test on. Code compiles though.

Isochronous endpoints are left un-implemented.

Another thing to note: some revisions of the nrf52840 chip need many more errata to function properly. As I don't have the hardware to test, and to keep things simple, these extra errata are not implemented. If one enables the USB stack on such chip, it will panic, given that the rest of the code will be unstable on these chips.

### Testing Strategy

This pull request was tested by:
- Running the USB bulk-echo test [app](https://github.com/tock/libtock-c/tree/master/examples/tests/usb) with the [host test program](https://github.com/tock/tock/tree/master/tools/usb/bulk-echo).
- Testing a USB-interrupt-based (HID) protocol.

### TODO or Help Wanted

This pull request still needs:
[...]

PR #1434, opened 2019-10-28
Move USB control transaction handling to a separate module and add more descriptors.

gendx:
### Pull Request Overview

Until now, the [`usbc_client` capsule](https://github.com/tock/tock/blob/master/capsules/src/usbc_client.rs) has hard-coded USB descriptors for a dummy USB echo protocol.

This pull request extracts the USB control protocol into a separate generic module that can be given any set of USB descriptors and manage interactions with USB control packets.

Also, I added HID and report descriptors.

This will allow to more easily define real-world USB protocols.

### Testing Strategy

This pull request was tested by:
- Flashing Tock on an nRF52840-DK board, and testing a USB-HID protocol (with other patches coming in further pull-requests). In particular, USB enumeration was tested on multiple operating systems, including Linux and OSX.
- Running `make ci-travis`.

### TODO or Help Wanted

A USB-HID protocol is pending for another pull-request.

### Documentation Updated

- [ ] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

PR #1427, opened 2019-10-24
Add USB register mapping (for nRF52840-DK) to chips/nrf52.

gendx:
### Pull Request Overview

This pull request is a first of a series to add USB support for the nRF52840-DK board.
This pull request is only about the register mapping.

### Testing Strategy

This pull request was tested by:
- Running `make ci-travis`.
- Implementing a USB protocol and testing it on the nRF52840-DK board. However, this requires more pull-requests to reproduce.

### TODO or Help Wanted

We still need to decide whether there should be a specific `chip/nrf52840` folder, or if this should be enabled by a feature (`#[cfg(feature = "nrf52840")]`).

Implementing the actual USB logic will follow in further pull-requests, but it seemed reasonable to me to split it in smaller changes, given all the logic involved.

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

PR #2125, opened 2020-09-24
USB: CTAP: support for FIDO HID as tested with OpenSK.

### LINKED ISSUE #2123: [sam4l] Uniformization of USB packet lengths across platforms?
In `capsules/src/usb/usbc_client.rs`, we define platform-dependent maximum USB packet lengths.

https://github.com/tock/tock/blob/77598c10a7ff40e1911e66731d10381ae4255244/capsules/src/usb/usbc_client.rs#L28-L33

The distinction for sam4l seems to come from the current configuration, i.e. using `EPSIZE::Bytes8` in e.g. the following.

https://github.com/tock/tock/blob/77598c10a7ff40e1911e66731d10381ae4255244/chips/sam4l/src/usbc/mod.rs#L1525-L1528

However, it seems that sam4l can support many more packet lengths.

https://github.com/tock/tock/blob/77598c10a7ff40e1911e66731d10381ae4255244/chips/sam4l/src/usbc/mod.rs#L164-L173

Other platform would also work with other packet lengths I think, it's mainly a matter of setting the length in the `DeviceDescriptor`.

https://github.com/tock/tock/blob/ea4acbe3423253a3e5b67fc4af6cb661d98f1b41/capsules/src/usb/descriptors.rs#L310-L311

Can we uniformize across platforms by always using 64-byte USB packets? This would greatly simplify things by removing the "platform-specific" constants (which really are not platform-specific but configuration-specific), at least until a compelling use case arises for another packet length.

Does anyone has context on the sam4l to confirm it would work? If nobody has context, I haven't seen much support for sam4l's USB lately, so as a default can we assume that 64 bytes is fine?

---

I also don't understand why these constants are duplicated in the CDC capsule:

https://github.com/tock/tock/blob/ea4acbe3423253a3e5b67fc4af6cb661d98f1b41/capsules/src/usb/cdc.rs#L35-L40

gendx:
### Pull Request Overview

This pull request adds support for the USB-CTAP-HID protocol as specified by FIDO2. This is an alternative proposal to #2094 - contrary to #2094, this proposal was extensively tested on nRF52840 chips in the context of [OpenSK](https://github.com/google/OpenSK).

The main differences with the other proposal are the following.
- The driver supports all necessary operations to build a fully-fledged FIDO2 application (as opposed to a simple demo of exchanging HID messages), namely:
  - **Transfer direction**: #2094 only supports "send" and "receive" operations, which isn't sufficient. Indeed, with USB the host decides whether to initiate an OUT transaction or an IN transaction. So it's possible that while a Tock app wants to send a packet to the host, the host actually decided to send a packet to Tock. This in particular happens with FIDO2 "cancel" messages - the host stops receiving any response from Tock and instead sends a cancel packet to Tock. This proposal supports a more general "send_or_receive" operation for this use case where the app aims to send a packet, but is ready to instead receive a packet if the host sends one (in which case sending the packet is aborted).
  - **Cancellation**: Again, because USB transactions are initiated by the host, it's possible that no transaction happens for some time. An application may want to setup a timeout (timeouts are actually mandated by FIDO2 specs). To implement that, the application can separately use the alarm driver to get notified of the timeout, and in that case send a cancellation command to the CTAP-USB driver. This use case isn't supported in #2094.
  - **Allow**: #2094 supports an allow buffer in each direction. This proposal has a single allow buffer, but keeps track of the current direction (send, receive or send+receive). I don't think it makes sense to concurrently set up separate send & receive buffers (the send+receive operation supports that).
  - **Subscribe**: #2094 only supports a callback for receiving. This is incorrect, as even a send operation should get a completion callback (similarly to the console driver for example). Otherwise, the application has no way to know when it can claim back the allowed buffer! This proposal supports a single callback, but keeping track of the current direction (send, receive or send+receive).
  - **Command**: #2094 supports commands for sending and receiving data. This proposal additionally support the aforementioned send+receive, as well as cancel operations. Besides, an explicit "connect" command is required to identify which app currently "owns" the USB stack - whereas #2094 makes it implicit.
- **Testing**: This proposal was extensively tested on nRF52840. #2094 only mentions basic tests on OpenTitan. This proposal was not yet tested on OpenTitan, although I don't expect much chip-specific traps above the USB HIL.
- **Component**: This proposal has a parameter for the `max_ctrl_packet_size` to forward to the USB controller. #2094 unconditionally uses a `max_packet_size_ep0` of `MAX_CTRL_PACKET_SIZE = 64` in the `DeviceDescriptor`, which is not compatible with Tock's current USB implementation on sam4l (#2123). We should fix #2123 or remove USB support on sam4l (is it really supported?) before hard-coding a control packet of 64-bytes.
- **Endpoint configuration**: #2094 unconditionally calls `endpoint_in_enable` and `endpoint_out_enable` separately, even if the same endpoint number is used for the sending and receiving ends (as in the [CTAP spec example](https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html#usb-descriptors)). This doesn't work with the current USB HIL, which expects `endpoint_in_out_enable` in that case. This proposal works with a single endpoint in IN+OUT mode.
- **HIL**: #2094 adds a new USB-HID HIL. I don't think it makes sense to stabilize such a HIL in the kernel at this point because (1) there is currently only one user of this HIL - the CTAP-USB driver - so we don't know what to unify on and (2) there is currently only one implementation of this HIL - based on USB. It seems therefore premature to me to define a kernel HIL without much more thought given on it. There is already a USB HIL to abstract over multiple chips at a lower level.

### Testing Strategy

This pull request was extensively tested together with the [OpenSK](https://github.com/google/OpenSK) app, on a wide range of FIDO2 scenarios, on various boards based on nRF52840 chips.

### TODO or Help Wanted

This pull request still needs to agree on an implementation that fits the needs of both #2094 and OpenSK.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/boards/components/src/lib.rs
+++ b/boards/components/src/lib.rs
@@ -39,3 +39,4 @@ pub mod temperature;
 pub mod temperature_stm;
 pub mod test;
 pub mod touch;
+pub mod usb_ctap;
--- a/boards/components/src/usb_ctap.rs
+++ b/boards/components/src/usb_ctap.rs
@@ -1,0 +1,88 @@
+//! Component for CTAP over USB.
+
+use capsules::usb::usb_ctap::CtapUsbSyscallDriver;
+use capsules::usb::usbc_ctap_hid::ClientCtapHID;
+use core::mem::MaybeUninit;
+use kernel::capabilities;
+use kernel::component::Component;
+use kernel::create_capability;
+use kernel::hil;
+use kernel::static_init_half;
+
+// Setup static space for the objects.
+#[macro_export]
+macro_rules! usb_ctap_component_buf {
+    ($C:ty) => {{
+        use capsules::usb::usb_ctap::CtapUsbSyscallDriver;
+        use capsules::usb::usbc_ctap_hid::ClientCtapHID;
+        use core::mem::MaybeUninit;
+        static mut BUF1: MaybeUninit<ClientCtapHID<'static, 'static, $C>> = MaybeUninit::uninit();
+        static mut BUF2: MaybeUninit<CtapUsbSyscallDriver<'static, 'static, $C>> =
+            MaybeUninit::uninit();
+        (&mut BUF1, &mut BUF2)
+    };};
+}
+
+pub struct UsbCtapComponent<C: 'static + hil::usb::UsbController<'static>> {
+    board_kernel: &'static kernel::Kernel,
+    controller: &'static C,
+    max_ctrl_packet_size: u8,
+    vendor_id: u16,
+    product_id: u16,
+    strings: &'static [&'static str],
+}
+
+impl<C: 'static + hil::usb::UsbController<'static>> UsbCtapComponent<C> {
+    pub fn new(
+        board_kernel: &'static kernel::Kernel,
+        controller: &'static C,
+        max_ctrl_packet_size: u8,
+        vendor_id: u16,
+        product_id: u16,
+        strings: &'static [&'static str],
+    ) -> Self {
+        Self {
+            board_kernel,
+            controller,
+            max_ctrl_packet_size,
+            vendor_id,
+            product_id,
+            strings,
+        }
+    }
+}
+
+impl<C: 'static + hil::usb::UsbController<'static>> Component for UsbCtapComponent<C> {
+    type StaticInput = (
+        &'static mut MaybeUninit<ClientCtapHID<'static, 'static, C>>,
+        &'static mut MaybeUninit<CtapUsbSyscallDriver<'static, 'static, C>>,
+    );
+    type Output = &'static CtapUsbSyscallDriver<'static, 'static, C>;
+
+    unsafe fn finalize(self, static_buffer: Self::StaticInput) -> Self::Output {
+        let grant_cap = create_capability!(capabilities::MemoryAllocationCapability);
+
+        let usb_ctap = static_init_half!(
+            static_buffer.0,
+            ClientCtapHID<'static, 'static, C>,
+            ClientCtapHID::new(
+                self.controller,
+                self.max_ctrl_packet_size,
+                self.vendor_id,
+                self.product_id,
+                self.strings,
+            )
+        );
+        self.controller.set_client(usb_ctap);
+
+        // Configure the USB userspace driver
+        let usb_driver = static_init_half!(
+            static_buffer.1,
+            CtapUsbSyscallDriver<'static, 'static, C>,
+            CtapUsbSyscallDriver::new(usb_ctap, self.board_kernel.create_grant(&grant_cap))
+        );
+        usb_ctap.set_client(usb_driver);
+
+        usb_driver
+    }
+}
--- a/boards/components/src/usb_ctap.rs
+++ b/boards/components/src/usb_ctap.rs
@@ -7,7 +7,8 @@ use kernel::capabilities;
 use kernel::component::Component;
 use kernel::create_capability;
 use kernel::hil;
-use kernel::static_init_half;
+use kernel::hil::usb::Client;
+use kernel::{static_init, static_init_half};
 
 // Setup static space for the objects.
 #[macro_export]
@@ -17,8 +18,7 @@ macro_rules! usb_ctap_component_buf {
         use capsules::usb::usbc_ctap_hid::ClientCtapHID;
         use core::mem::MaybeUninit;
         static mut BUF1: MaybeUninit<ClientCtapHID<'static, 'static, $C>> = MaybeUninit::uninit();
-        static mut BUF2: MaybeUninit<CtapUsbSyscallDriver<'static, 'static, $C>> =
-            MaybeUninit::uninit();
+        static mut BUF2: MaybeUninit<CtapUsbSyscallDriver<'static>> = MaybeUninit::uninit();
         (&mut BUF1, &mut BUF2)
     };};
 }
@@ -55,9 +55,9 @@ impl<C: 'static + hil::usb::UsbController<'static>> UsbCtapComponent<C> {
 impl<C: 'static + hil::usb::UsbController<'static>> Component for UsbCtapComponent<C> {
     type StaticInput = (
         &'static mut MaybeUninit<ClientCtapHID<'static, 'static, C>>,
-        &'static mut MaybeUninit<CtapUsbSyscallDriver<'static, 'static, C>>,
+        &'static mut MaybeUninit<CtapUsbSyscallDriver<'static>>,
     );
-    type Output = &'static CtapUsbSyscallDriver<'static, 'static, C>;
+    type Output = &'static CtapUsbSyscallDriver<'static>;
 
     unsafe fn finalize(self, static_buffer: Self::StaticInput) -> Self::Output {
         let grant_cap = create_capability!(capabilities::MemoryAllocationCapability);
@@ -75,13 +75,24 @@ impl<C: 'static + hil::usb::UsbController<'static>> Component for UsbCtapCompone
         );
         self.controller.set_client(usb_ctap);
 
+        let send_buffer = static_init!([u8; 64], [0; 64]);
+        let recv_buffer = static_init!([u8; 64], [0; 64]);
+
         // Configure the USB userspace driver
         let usb_driver = static_init_half!(
             static_buffer.1,
-            CtapUsbSyscallDriver<'static, 'static, C>,
-            CtapUsbSyscallDriver::new(usb_ctap, self.board_kernel.create_grant(&grant_cap))
+            CtapUsbSyscallDriver<'static>,
+            CtapUsbSyscallDriver::new(
+                usb_ctap,
+                self.board_kernel.create_grant(&grant_cap),
+                send_buffer,
+                recv_buffer
+            )
         );
         usb_ctap.set_client(usb_driver);
+
+        usb_ctap.enable();
+        usb_ctap.attach();
 
         usb_driver
     }
--- a/capsules/src/driver.rs
+++ b/capsules/src/driver.rs
@@ -26,6 +26,7 @@ pub enum NUM {
     I2cMaster             = 0x20003,
     UsbUser               = 0x20005,
     I2cMasterSlave        = 0x20006,
+    UsbCtap               = 0x20009,
 
     // Radio
     BleAdvertising        = 0x30000,
--- a/capsules/src/usb/mod.rs
+++ b/capsules/src/usb/mod.rs
@@ -1,5 +1,7 @@
 pub mod cdc;
 pub mod descriptors;
+pub mod usb_ctap;
 pub mod usb_user;
 pub mod usbc_client;
 pub mod usbc_client_ctrl;
+pub mod usbc_ctap_hid;
--- a/capsules/src/usb/usb_ctap.rs
+++ b/capsules/src/usb/usb_ctap.rs
@@ -1,0 +1,355 @@
+use super::usbc_ctap_hid::ClientCtapHID;
+use kernel::hil;
+use kernel::hil::usb::Client;
+use kernel::{AppId, AppSlice, Callback, Driver, Grant, ReturnCode, Shared};
+
+/// Syscall number
+use crate::driver;
+pub const DRIVER_NUM: usize = driver::NUM::UsbCtap as usize;
+
+pub const CTAP_CMD_CHECK: usize = 0;
+pub const CTAP_CMD_CONNECT: usize = 1;
+pub const CTAP_CMD_TRANSMIT: usize = 2;
+pub const CTAP_CMD_RECEIVE: usize = 3;
+pub const CTAP_CMD_TRANSMIT_OR_RECEIVE: usize = 4;
+pub const CTAP_CMD_CANCEL: usize = 5;
+
+pub const CTAP_ALLOW_TRANSMIT: usize = 1;
+pub const CTAP_ALLOW_RECEIVE: usize = 2;
+pub const CTAP_ALLOW_TRANSMIT_OR_RECEIVE: usize = 3;
+
+pub const CTAP_SUBSCRIBE_TRANSMIT: usize = 1;
+pub const CTAP_SUBSCRIBE_RECEIVE: usize = 2;
+pub const CTAP_SUBSCRIBE_TRANSMIT_OR_RECEIVE: usize = 3;
+
+pub const CTAP_CALLBACK_TRANSMITED: usize = 1;
+pub const CTAP_CALLBACK_RECEIVED: usize = 2;
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum Side {
+    Transmit,
+    Receive,
+    TransmitOrReceive,
+}
+
+impl Side {
+    fn can_transmit(&self) -> bool {
+        match self {
+            Side::Transmit | Side::TransmitOrReceive => true,
+            Side::Receive => false,
+        }
+    }
+
+    fn can_receive(&self) -> bool {
+        match self {
+            Side::Receive | Side::TransmitOrReceive => true,
+            Side::Transmit => false,
+        }
+    }
+}
+
+#[derive(Default)]
+pub struct App {
+    // Only one app can be connected to this driver, to avoid needing to route packets among apps.
+    // This field tracks this status.
+    connected: bool,
+    // Currently enabled transaction side. Subscribing to a callback or allowing a buffer
+    // automatically sets the corresponding side. Clearing both the callback and the buffer resets
+    // the side to None.
+    side: Option<Side>,
+    callback: Option<Callback>,
+    buffer: Option<AppSlice<Shared, u8>>,
+    // Whether the app is waiting for the kernel signaling a packet transfer.
+    waiting: bool,
+}
+
+impl App {
+    fn check_side(&mut self) {
+        if self.callback.is_none() && self.buffer.is_none() && !self.waiting {
+            self.side = None;
+        }
+    }
+
+    fn set_side(&mut self, side: Side) -> bool {
+        match self.side {
+            None => {
+                self.side = Some(side);
+                true
+            }
+            Some(app_side) => side == app_side,
+        }
+    }
+
+    fn is_ready_for_command(&self, side: Side) -> bool {
+        self.buffer.is_some() && self.callback.is_some() && self.side == Some(side)
+    }
+}
+
+pub trait CtapUsbClient {
+    // Whether this client is ready to receive a packet. This must be checked before calling
+    // packet_received().
+    fn can_receive_packet(&self) -> bool;
+
+    // Signal to the client that a packet has been received.
+    fn packet_received(&self, packet: &[u8; 64]);
+
+    // Signal to the client that a packet has been transmitted.
+    fn packet_transmitted(&self);
+}
+
+pub struct CtapUsbSyscallDriver<'a, 'b, C: 'a> {
+    usb_client: &'a ClientCtapHID<'a, 'b, C>,
+    apps: Grant<App>,
+}
+
+impl<'a, 'b, C: hil::usb::UsbController<'a>> CtapUsbSyscallDriver<'a, 'b, C> {
+    pub fn new(usb_client: &'a ClientCtapHID<'a, 'b, C>, apps: Grant<App>) -> Self {
+        CtapUsbSyscallDriver { usb_client, apps }
+    }
+}
+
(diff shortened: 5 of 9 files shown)


bradjc:
I don't claim to understand the intricacies of this stack anywhere near as well as @gendx or @alistair23 (all I've done is get the nano33 board to show up as a CTAP HID device on my laptop). But, I will try to structure a conversation so we can resolve the differences and hopefully converge. Also, I'm still of the mindset that USB is experimental and we haven't decided to stabilize on the current stack.

I see three potential friction points:
1. What should the userspace interface (i.e. `Driver` implementation) be?
2. Layering and abstractions.
3. Feature completeness/robustness/testedness of the code in capsules/usb.


### 1. What should the userspace interface (i.e. `Driver` implementation) be?

This seems like an area where we have a lot of flexibility. I imagine that OpenSK wants to keep the userspace interface the same. @alistair23 do you have a sense of how different the version in this PR is? Would the version in this PR work with the libraries you have been using in userspace?

If not, could we extend the `Driver` impl in this PR in a backwards-compatible way?

If not, let's go forward with two userspace APIs (for now at least). This has been a design argument from the beginning that we aren't always going to know what the userspace-kernel interface should be, and having a 32 bit address space for Driver naming gives us a lot of room to explore.

### 2. Layering and abstractions.

#### USB HID interface

As I understand it, most of the USB-related work with CTAP is really a HID implementation. Given that, it would be nice to be able to re-use the same code for other HID drivers in Tock. HID (I would argue like CDC-ACM) is one of those APIs that is quite powerful because of how ubiquitous it is. I can totally see someone building a keyboard of sorts on top of Tock. Therefore, I think adding new interfaces as in #2094 is a step in the right direction, as it helps lower the bar for future development.

Now, I too didn't think that USB HID should be in kernel/hil, but for better or worse that is our only central folder with interfaces, and we have several interfaces in there already that are not hardware peripheral interfaces (light sensor, for example). That might be something we want to change in the future, but I don't think is relevant here.

Lastly, we can't quite get a clean separation between CTAP and HID because of how USB descriptors are created in Tock (namely all at once), hence the `CtapHid` type in #2094. However, it seems like there is growing need to allow for a more flexible configuration of descriptors, and I think having the USB HID interface helps push in that direction. Specifically, it makes for a nice way to demonstrate that a more pluggable USB init architecture is actually working (because the HID and CTAP layers can set their descriptors independently).

#### CTAP interface

Additionally, as I understand it CTAP can be layered on top of more than just USB. If so, I think it makes sense to structure things to make it easier to choose a different transport in the future. However, I think currently this is a minor issue (boils down to an `Option<UsbCtapHid>` in capsules/ctap in #2094 rather than a required reference). So I'm not sure I feel too strongly either way.

### 3. Feature completeness/robustness/testedness of the code in capsules/usb.

I think it makes sense that this PR has more mileage on it, and Tock can benefit from the more tested code. In just my opinion I think we should converge on the other issues first.




------

My take, for what it's worth: add @alistair23's layers from #2094 to this PR, and either merge the two Driver impls or merge both.

alistair23:
Thanks for writing that up Brad, my thoughts are below.

1. They don't seem that different. I'm sure OpenSK wants to keep them the same as well. The original upstream CTAP implementation (#2094) is already merged in libtock-rs with an example. There is also another libtock-rs PR pending that is using that interface. I think in this case it makes sense to stick with the current #2094 implementation as it has userspace support and has already been received.

If there are missing calls or functionality, they can easily be added without breaking the current support.

2. I prefer adding a USB HID HIL. To me it feels like it matches what other Tock devices do and moves us in a direction where we can re-use more of the USB code between different devices.

#2094 should also allow CTAP support for non-USB (like BLE or NFC) which I see as an advantage.

3. This PR doesn't connect the CTAP with any board, so it's hard to tell how tested it is. I'm sure it's well tested with the OpenSK fork, but that is somewhat different to the current Tock kernel and userspace. As well as that most of the functionality is in userspace, so I'm not sure how much that matters.

#2094 has been open for almost a month. It has been approved by @bradjc and reviewed/comment on by multiple other people. This PR was only opened because of #2094. I feel that #2094 should be merged and any missing features can be ported from this PR to mainline. For example #2094 does not have cancellation. It would be useful to port that from this PR. In saying that, this PR doesn't seem to actually cancel a transfer once it has been sent to the USB driver.

#2094 has an upstream userspace implementation and I plan to continue to work on this and improve it. While OpenSK has had this code public for 9 months and failed to send it upstream. It was even originally licensed differently to Tock, which means that others couldn't use it for an upstream implementation.
It seems entirely possible that once this is merged it will be abandoned. It isn't connected to any boards, so I'm not sure who will be testing it. I also don't see any corresponding userspace support with this PR, which means that it can't be tested for CTAP functionality. While #2094 has a upstream libtock-rs implementation to test the driver interface and a pending PR to add full CTAP functionality. #2094 even has been somewhat tested on two different platforms and architectures.

gendx:
Thanks @bradjc for the summary. My thoughts on your points, and @alistair23's view.

### 1. What should the userspace interface (i.e. Driver implementation) be?

We can definitely make both drivers available in Tock. We have enough space in the driver numbers. They can both be marked as "experimental" for now, and with more time and experience building applications on top of them we can hopefully unify them in a meaningful manner.

> They don't seem that different.

Please read [my description above](https://github.com/tock/tock/pull/2125#issue-492595854). We cannot get far with USB without send+receive and cancellation operations. And not having any callback on write is incorrect.

> The original upstream CTAP implementation (#2094) is already merged in libtock-rs with an example.

I find it strange that libtock-rs already supports an API that is not yet available in the Tock kernel. I don't think that "it's already there" is a good argument to favor one implementation over the other, because it doesn't show robustness of said implementation - it simply shows that one implementation came first.

> If not, let's go forward with two userspace APIs (for now at least). This has been a design argument from the beginning that we aren't always going to know what the userspace-kernel interface should be, and having a 32 bit address space for Driver naming gives us a lot of room to explore.

+1 to this compromise proposed by @bradjc.

### 2. Layering and abstractions.

> I prefer adding a USB HID HIL. To me it feels like it matches what other Tock devices do and moves us in a direction where we can re-use more of the USB code between different devices.

I'm not against adding a USB HID HIL, but a HIL should IMO require a bit more design and review from the Tock core team than an experimental pull request that works for a single & simple use case. Otherwise we'll soon end up having to spend much more time on revising this HIL. See e.g. what happened with the time HIL, it took months for a separate working group to come up with the time HIL v3 and migrate everything to use it.

My point was simply that it's not clear yet what the USB HID HIL should be. It should at least be compatible with both #2094 and #2125.

### 3. Feature completeness/robustness/testedness of the code in capsules/usb.

> In just my opinion I think we should converge on the other issues first.

I agree with that.

> #2094 has been open for almost a month.

To be accurate, it's been just 3 weeks.

> It has been approved by @bradjc and reviewed/comment on by multiple other people.

Indeed, many shortcomings have only been addressed in the past week, following some recent reviews. So far, only @bradjc has approved it, there are still some pending reviews and comments.

If time was an argument, I don't think it's fair to say that #2094 has been around for a month - the current iteration has only been around for a few days.
Even a month is not that long compared to other pull requests in the queue - most have been around for >1 month.

But in any case, should we prioritize merging pull requests based on a "first come first served" basis, or based on technical robustness?

> For example #2094 does not have cancellation. It would be useful to port that from this PR.

I think it would indeed be useful to port cancellation before rushing to merge #2094.

> In saying that, this PR doesn't seem to actually cancel a transfer once it has been sent to the USB driver.

Please be more specific before making such broad claims. This PR has a `cancel_transaction` operation in `capsules/src/usb/usbc_ctap_hid.rs`. Please be more specific about what is not cancelled.

> While OpenSK has had this code public for 9 months and failed to send it upstream.

9 months ago there was no interest in USB in upstream Tock apart from OpenSK. And CTAP seemed like a niche use case. So it didn't seem clear at that time whether it should be beneficial to upstream it to Tock.

But most of the USB code from OpenSK was actually [sent and merged upstream](https://github.com/tock/tock/pulls?q=is%3Apr+is%3Aclosed+author%3Agendx+usb), notably the nRF52840 chip support (e.g. #1427, #1543), as well as support for more generic USB control transactions (#1434).

> It was even originally licensed differently to Tock, which means that others couldn't use it for an upstream implementation.

OpenSK is licensed as Apache 2.0 which is a common open-source license. If that's preventing others from using it, the minimum would have been to ask OpenSK maintainers - it's not like the OpenSK project is unmaintained at the moment, see the [history of commits](https://github.com/google/OpenSK/commits/master).

And it's not like OpenSK is an unknown project either - we've contributed [quite a few commits](https://github.com/tock/tock/commits?author=gendx) to upstream Tock in the context of OpenSK...

> It seems entirely possible that once this is merged it will be abandoned.

It's possible for any of Tock's capsules or boards. It's likewise entirely possible that #2094 will be abandoned once it is merged. Please be mindful before making such broad statements.

> It isn't connected to any boards,

I didn't include the code to pipe it to the Nordic boards because I assumed that merging a capsule is a different concern than enabling it on a board. Thanks to components, it's now quite easy to configure boards to add/remove capsules.

But if you insist, I can add that, it's on the [patch](https://github.com/google/OpenSK/blob/master/patches/tock/02-usb.patch).

> so I'm not sure who will be testing it.

For a start, anyone testing OpenSK will be testing it.

> I also don't see any corresponding userspace support with this PR, which means that it can't be tested for CTAP functionality.

By definition, there is no corresponding userspace support in this repository. There is however support [in the OpenSK repository](https://github.com/google/OpenSK/blob/master/third_party/libtock-drivers/src/usb_ctap_hid.rs).

As to whether userspace support is better in libtock-rs vs. in the OpenSK repository, I'll point out that there have been substantial changes in libtock-rs with the new `async` runtime just when we originally open-sourced OpenSK, and further changes are still pending, which is why things have not converged yet in userspace. But the intent is not to keep a fork forever on OpenSK's side, it's simply blocked on implementing and stabilizing https://github.com/tock/libtock-rs/issues/217.

> While #2094 has a upstream libtock-rs implementation to test the driver interface and a pending PR to add full CTAP functionality.

If you refer to [examples/ctap.rs](https://github.com/tock/libtock-rs/blob/master/examples/ctap.rs), I only see it receiving a single packet and writing it back to the console. That's a very narrow test scenario.

On the contrary this PR has been tested on a fully-fledged FIDO2 application. There is a difference between a "receive 1 packet" example and the whole logic in https://github.com/google/OpenSK/tree/master/src.

> #2094 even has been somewhat tested on two different platforms and architectures.

This PR has been more than "somewhat" tested on Nordic boards, it's been extensively tested there with OpenSK. I don't see anything specific to a given platform/architecture as capsules are above the USB HIL, so I don't think it's fair to say #2094 has received more testing simply because an example was tried on OpenTitan as well.

Also, according to https://github.com/tock/tock/pull/2115, USB doesn't work on OpenTitan, so I don't see how you can at the same time claim that #2094 was tested on the OpenTitan platform.

alistair23:
I have added a `send or receive` command so that the new userspace API can be used by OpenSK. That should address all concerns raised here.

jmichelp:
Aside of the technical debate, I would like to address one comment:
>  I imagine that OpenSK wants to keep the userspace interface the same.

To be honest, not necessarily. If the interface changes, of course it gives us more work to migrate and test again, but it's still 1 patch less that we have to maintain so the effort of migrating is worth it on the long run. Our ideal world would be to not apply patches and just run a stock kernel.
If the new interface/driver breaks us, it means we're drifting even more from a stock kernel because we have to either rollback a PR or pin a commit before this PR and never upgrade again. We upstreamed as much as we could and we even tried to do so with our flash-friendly persistent storage (with a bit less success though). And if there are code we have on OpenSK that you think would make sense in the kernel or in libtock-rs, we can make it happen (e.g. other boards we support or the "DFU" flavor of the boards that allow flashing over USB without external hardware).

But testing a USB stack for OpenSK is not cheap and that's where our main concerns live. We tested our USB stack, end-to-end, on multiple platforms (Macbook, Android, Linux, Windows, Chromebook) and against different websites and web browsers. So we're pretty confident that it's reliable. The discussion about the endpoint numbers on #2094 might seem anecdotical but if this breaks our interoperability because it works on some platforms but not on others, that's a big issue for us. The interface is less of a concern for me than reliability :)

alistair23:
> If the new interface/driver breaks us, it means we're drifting even more from a stock kernel because we have to either rollback a PR or pin a commit before this PR and never upgrade again.

This is why we are in this mess in the first place. If something in the new interface breaks, why not fix it and send the patch upstream? That way you don't need to pin anything AND don't need any patches on top of mainline. Everybody wins.

> We tested our USB stack, end-to-end, on multiple platforms (Macbook, Android, Linux, Windows, Chromebook) and against different websites and web browsers.

You have been testing the OpenSK fork of Tock and OpenSK fork of userspace though. For the mainline kernel that's worrying. I'm guessing this implementation has never been tested with mainline Tock OR with upstream libtock-rs/libtock-c.

There is a growing list of reasons to re-write some of the USB HIL. When that happens this will all need to be re-tested anyway. With this implementation no one else can test CTAP as there is no board connection and no libtock-rs/libtock-c support. That is pretty worrying to me.

On the other hand #2094 has support for libtock-rs as well as boards. It can easily be tested. Your testing efforts will need to be done again in the future for other changes, so one more doesn't seem like such a problem.

Specifically with the endpoint numbers, if there are implementations that don't follow the spec and use two endpoints then we can address that with a fix in the future. Currently there is no reason to think that is the case, and following the spec with two endpoints seems like a better approach.

The worst case with #2094 is that you can continue using the current patch you have. If testing takes awhile you can put it off until you are ready. Nothing in #2094 will break anything for your current approach.

phil-levis:
I think it's OK to have multiple implementations which are at different levels of functionality. But it's very important that we have different implementations of the same abstractions, not fork into multiple different software decompositions. This is why I've been pushing on the HIL in #2094; if we get the HIL right, then implementations can improve and change independently. But if we have a HIL that's not right, or has bugs, or is tied to one implementation, then changes to it require coordinating across the implementations: get the interfaces right first! This is why people do thinks like debate POSIX for years. Thankfully our job is much smaller and simpler.

I'd like to propose that we split out the HIL from #2094 into a separate PR, and make sure that it has the complete functionality that the OpenSK stack has found is needed, while remaining compatible with both the OT and OpenSK stacks. Once we have that, we can then PR OT and nRF implementations. We should then talk about the Driver/syscall interface. In the best world. we'll reach a consensus on what that interface should be; we could have one or two or N implementations. If we don't reach consensus, then I'd argue it's suboptimal but OK to have two different userspace APIs.

@alistair23 -- with respect to the fact that #2094 PR is 3 weeks old and was here first, I don't think that's particularly relevant. We want Tock to have the best code that is robust, well designed, and easy to maintain. A "first stake in the ground wins" policy is fundamentally at odds with this, because it encourages quick work that can create technical debt.

alistair23:
I have split out the HIL to https://github.com/tock/tock/pull/2130. I'll leave the re-write of this PR to use a HID HIL up to @gendx.

Just a note that if this PR is rebased on a USB HID HIL, I expect the two implementations to be practically identical.

I agree with the "first stake in the ground" isn't a great policy, I don't think I was saying that. In fact this implementation was first to exist. I think #2094 has the advantages that the CTAP capsule is not being USB HID specific, it is connected to a board, has a userspace test case that others can run, has been fully tested on 1 platform and partially tested on a second and has been reviewed by multiple people.

phil-levis:
> I have split out the HIL to #2130. I'll leave the re-write of this PR to use a HID HIL up to @gendx.

Great.
 
> Just a note that if this PR is rebased on a USB HID HIL, I expect the two implementations to be practically identical.

I see this as a plus -- if they differ, then that suggests an inconsistency and potentially a bug.
 
> I agree with the "first stake in the ground" isn't a great policy, I don't think I was saying that. In fact this implementation was first to exist. I think #2094 has the advantages that the CTAP capsule is not being USB HID specific, it is connected to a board, has a userspace test case that others can run, has been fully tested on 1 platform and partially tested on a second and has been reviewed by multiple people.

Makes sense.

gendx:
Seems that the discussion is moving forward! I'll check how I can make this proposal compatible with the HIL proposed in #2130 - if that works fine we can make progress from there :)

---

I also want to clarify a point regarding testing.

> You have been testing the OpenSK fork of Tock and OpenSK fork of userspace though. For the mainline kernel that's worrying. I'm guessing this implementation has never been tested with mainline Tock OR with upstream libtock-rs/libtock-c.

In an ideal world, everyone would run exhaustive tests on every pull-request submitted to Tock or libtock-rs. In practice, that's very hard to do because:
- Most practical end-to-end tests are very hard to automate. We can write unit tests that run in CI, and even some tests with qemu. But (1) qemu still diverges a bit from real hardware and (2) it's quite hard to automate a real scenario with many components like a web browser, a user touching a button & observing LEDs, etc.
- Manual tests are expensive to run, so not practical to do on every single pull-request. That's why we have releases of Tock where we run the manual tests (and multiple release candidates to fix the bugs we discover).
- The bigger the tests, the higher the cost to maintain them. Even if we could automate large end-to-end tests, code evolves so tests have to evolve too. But while it's easy to re-write unit tests when we update the functionality of a capsule, it's much harder to refactor an end-to-end test with plenty of components.
- Tock and libtock-rs are not necessarily in sync. Some (most) combinations of a Tock commit and libtock-rs commit are incompatible.
- Tock - and to some degree libtock-rs - is moving fast with many new commits every week. Making a mini-fork is unavoidable for each change (even regular pull requests of Tock have plenty of merges/rebases).

Therefore, it's more practical to start from a relatively stable checkpoint and temporarily fork from there - then aim to catch up and upstream the changes while syncing with the fast-moving upstream (and resolving the occasional conflicts). There's a balance between reliability and velocity - if we aim for each commit to be 100% reliable we'd make 10x fewer progress on Tock every month.

That being said, (1) the OpenSK fork of the Tock kernel has diverged less and less as we upstreamed most of the code and (2) we regularly sync with some recent Tock commit & perform manual tests there. So even if we don't necessarily have confidence on HEAD on the mainline kernel, there is still some level of confidence that at least the kernel commit pinned by OpenSK is stable enough - at least for the capsules used by OpenSK (though we sometimes need some bugfix patches until things are being fixed upstream).

Another question is who should have the burden of performing manual tests. Should someone proposing a pull-request to add a capsule to Tock also perform exhaustive OpenSK tests (I don't think so)? Should OpenSK developers perform exhaustive tests on each commit whenever a pull-request is proposed to Tock (I don't think so either)? A good balance is IMO to simply run manual tests on checkpoints/releases, i.e. when enough changes have been added for the overhead of running tests to make sense.

---

However, I still think that we could do better in terms of automated unit tests for individual components (e.g. at the scope of a capsule, a scheduler). For the most part, Tock's CI currently only checks the basic things (formatting, lints, "does it compile?"), and I agree that this is worrying. Of course, some bugs can slip through when integrating all the components together, but at least having well-tested individual components would raise the bar on reliability & confidence in the code.

It could also be useful to separate the code in "tiers" (well, it's already the case to some extent) - e.g. the core `kernel/` should have plenty of testing, whereas individual capsules or chip-specific elements have a level of scrutiny depending on their actual usage (maybe some capsules are proof-of-concepts without extensive testing, while others like GPIOs are more essential). We could even explicitly split capsules into `core/` and `experimental/` or something like that so that users know what to expect (and e.g. require unit tests and qemu tests for `core/` capsules).

gendx:
This is now compatible with #2130 - at least it compiles. I still have to test on real hardware and scenarios.

phil-levis:
@gendx -- once you have fixed the conflicts tested, I think this raises the question of which implementation we should include. I'll put the discussion on this in #2094 since it's the earlier PR.

gendx:
> @gendx -- once you have fixed the conflicts tested, I think this raises the question of which implementation we should include. I'll put the discussion on this in #2094 since it's the earlier PR.

I've added some comment on an issue I stumbled upon when trying to migrate to the new USB-HID HIL, regarding managing the lifecycle of USB (plugged in, ready, plugged out) https://github.com/tock/tock/pull/2094#issuecomment-706092725.

To me it seems quite important, but orthogonal to the #2094 vs. #2125 debate - and I think converging between #2094 and #2125 should be the easy part compared to that.

bradjc:
Leaving this open as I would like to see the capsule with the `Driver` impl from this PR merged.

alevy:
I think this is _mostly_ resolved at this point with the exception of the system call interface. I'll open an issue to track unifying the OpenSK USB stack and the upstream USB stack and we can close this PR for now (and of course pull over the still relevant part)

PR #2126, opened 2020-09-25
Earlgrey: dont call debug in interrupt top half

hudson-ayers:
### Pull Request Overview

This pull request replaces calls to `debug!()` in the top half interrupt handler for the `earlgrey` chip with `panic!()`. If these calls executed it could cause UB, as the Tock kernel assumes that kernel code is not reentrant, and `debug!()` calls UART code.

I came across this while symbolically executing interrupt handlers in Tock.

### Testing Strategy

N/A

### TODO or Help Wanted

How bad is the possibility of a reentrant panic that this in theory could introduce?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


phil-levis:
I agree with removing debug!() from a safety standpoint, but do we need to panic!()?

alistair23 (approved):
I agree with panic here, these should never happen.

PR #2128, opened 2020-09-25
opentitan: Disable USB support by default

bradjc:
Due to an OpenTitan hardware bug (see https://github.com/lowRISC/opentitan/issues/2598)
USB doesn't work. Let's just disable USB by default to avoid confusing
people.

Replaces #2115

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers (approved):

alistair23:
Ah, I see. This will conflict with https://github.com/tock/tock/pull/2094 which is why it was separate in the first place. I'll just rebase it after this.

alistair23 (approved):

PR #2130, opened 2020-09-26
kernel: hil: Add a USB HID HIL

alistair23:
### Pull Request Overview

See https://github.com/tock/tock/pull/2094 for full details.

### Testing Strategy

See https://github.com/tock/tock/pull/2094 for full details.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/kernel/src/hil/usb_hid.rs
+++ b/kernel/src/hil/usb_hid.rs
@@ -1,0 +1,93 @@
+//! Interface for USB HID (Human Interface Device) class
+
+use crate::returncode::ReturnCode;
+
+/// The 'types' of USB HID, this should define the size of send/received packets
+pub trait UsbHidType: Copy + Clone + Sized {}
+
+impl UsbHidType for [u8; 64] {}
+impl UsbHidType for [u8; 32] {}
+impl UsbHidType for [u8; 16] {}
+impl UsbHidType for [u8; 8] {}
+
+/// Implement this trait and use `set_client()` in order to receive callbacks.
+pub trait Client<'a, T: UsbHidType> {
+    /// Called when a packet is received.
+    /// This will return the buffer passed into `receive_buffer()` as well as
+    /// the endpoint where the data was received. If the buffer length is smaller
+    /// then the data length the buffer will only contain part of the packet and
+    /// `result` will contain indicate an `ESIZE` error. Result will indicate
+    /// `ECANCEL` if a receive was cancelled by `receive_cancel()` but the
+    /// callback still occured. See `receive_cancel()` for more details.
+    fn packet_received(&'a self, result: ReturnCode, buffer: &'static mut T, endpoint: usize);
+
+    /// Called when a packet has been finished transmitting.
+    /// This will return the buffer passed into `send_buffer()` as well as
+    /// the endpoint where the data was sent. If not all of the data could
+    /// be sent the `result` will contain the `ESIZE` error. Result will
+    /// indicate `ECANCEL` if a send was cancelled by `send_cancel()`
+    /// but the callback still occured. See `send_cancel()` for more
+    /// details.
+    fn packet_transmitted(&'a self, result: ReturnCode, buffer: &'static mut T, endpoint: usize);
+
+    /// Called when checking if we can start a new receive operation.
+    /// Should return true if we are ready to receive and not currently
+    /// in the process of receiving anything. That is if we are currently
+    /// idle.
+    /// If there is an outstanding call to receive, a callback already
+    /// waiting to be called then this will return false.
+    fn can_receive(&'a self) -> bool;
+}
+
+pub trait UsbHid<'a, T: UsbHidType> {
+    /// Sets the buffer to be sent and starts a send transaction.
+    /// On success returns the number of bytes that will be sent.
+    /// On failure returns an error code and the buffer passed in.
+    fn send_buffer(&'a self, send: &'static mut T) -> Result<usize, (ReturnCode, &'static mut T)>;
+
+    /// Cancels a send called by `send_buffer()`.
+    /// If `send_cancel()` successfully cancels a send transaction
+    /// before the transaction has been acted upon this function will
+    /// return the buffer passed via `send_buffer()` and no callback
+    /// will occur.
+    /// If there is currently no send transaction (`send_buffer()`
+    /// hasn't been called) this will return `Err(EINVAL)`.
+    /// If the transaction can't be cancelled cleanly, either because
+    /// the send has already occured, a partial send has occured or the
+    /// send can not be cancelled by the hardware this will return
+    /// `Err(EBUSY)` and the callback will still occur.
+    /// Note that unless the transaction completes the callback will
+    /// indicate a result of `ECANCEL`.
+    fn send_cancel(&'a self) -> Result<&'static mut T, ReturnCode>;
+
+    /// Sets the buffer for received data to be stored and enables receive
+    /// transactions. Once this is called the implementation will enable
+    /// receiving via USB. Once a packet is received the `packet_received()`
+    /// callback will be triggered and no more data will be received until
+    /// this is called again.
+    ///
+    /// Once this is called, the implementation will wait until either
+    /// a packet is received or `receive_cancel()` is called.
+    ///
+    /// Calling `receive_buffer()` while there is an outstanding
+    /// `receive_buffer()` operation will return EBUSY.
+    ///
+    /// On success returns nothing.
+    /// On failure returns an error code and the buffer passed in.
+    fn receive_buffer(&'a self, recv: &'static mut T) -> Result<(), (ReturnCode, &'static mut T)>;
+
+    /// Cancels a receive called by `receive_buffer()`.
+    /// If `receive_cancel()` successfully cancels a receive transaction
+    /// before the transaction has been acted upon this function will
+    /// return the buffer passed via `receive_buffer()` and no callback
+    /// will occur.
+    /// If there is currently no receive transaction (`receive_buffer()`
+    /// hasn't been called) this will return `Err(EINVAL)`.
+    /// If the transaction can't be cancelled cleanly, either because
+    /// the receive has already occured, a partial receive has occured or the
+    /// receive can not be cancelled by the hardware this will return
+    /// `Err(EBUSY)` and the callback will still occur.
+    /// Note that unless the transaction completes the callback will
+    /// indicate a result of `ECANCEL`.
+    fn receive_cancel(&'a self) -> Result<&'static mut T, ReturnCode>;
+}
--- a/kernel/src/hil/mod.rs
+++ b/kernel/src/hil/mod.rs
@@ -26,6 +26,7 @@ pub mod time;
 pub mod touch;
 pub mod uart;
 pub mod usb;
+pub mod usb_hid;
 
 /// Shared interface for configuring components.
 pub trait Controller {


phil-levis:
@gendx -- Can you take a look at this HIL?

phil-levis (commented):
Overall, looks good. One question on the length return value of transmit.

kernel/src/hil/usb_hid.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// in the process of receiving anything. That is if we are currently
+    /// idle.
+    /// If there is an outstanding call to receive, a callback already
+    /// waiting to be called then this will return false.
+    fn can_receive(&'a self) -> bool;
+}
+
+pub trait UsbHid<'a> {
+    /// Sets the buffer to be sent and starts a send transaction.
+    /// On success returns the number of bytes sent.
```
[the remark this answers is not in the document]
alistair23: Fixed

kernel/src/hil/usb_hid.rs
```diff
@@ -0,0 +1,81 @@
...
+    );
+
+    /// Called when a packet has been finished transmitting.
+    /// This will return the buffer passed into `send_buffer()` as well as
+    /// the endpoint where the data was sent.
+    fn packet_transmitted(
+        &'a self,
+        result: ReturnCode,
+        buffer: &'static mut [u8; 64],
+        len: usize,
```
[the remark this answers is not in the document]
alistair23: It's the number of bytes sent and it can be different. I have updated the comment to make this clear.

kernel/src/hil/usb_hid.rs
```diff
@@ -0,0 +1,86 @@
...
+        send: &'static mut [u8; 64],
+    ) -> Result<usize, (ReturnCode, &'static mut [u8; 64])>;
+
+    /// Cancels a send called by `send_buffer()`. Once called the
+    /// `packet_transmitted()` callback won't occur and no new data
+    /// will be sent. Any data currently being sent will stil be sent.
+    /// On success (a send was cancelled) returns the buffer passed to
+    /// `send_buffer()`.
+    /// On failure returns a return code
+    fn send_cancel(&'a self) -> Result<&'static mut [u8; 64], ReturnCode>;
```
ppannuto, line 58: I recall @phil-levis thinking pretty carefully through the async cancellation semantics elsewhere (I believe in the radio stack?). I think it would be beneficial to spell out the cancellation semantics a bit more explicitly so as to make sure that when cancellation occurs, it's clearly deterministic to the caller what will happen.

I believe the radio semantics were that either (a) cancellation failed and a callback will still come, OR (b) cancellation succeeded and no callback will come.

The "any data currently being sent will still be sent" part here reads a bit ambiguously to me -- if I successfully cancel "more" (?) data from being sent, will I get both a successful cancel and a send-complete callback?

kernel/src/hil/usb_hid.rs
```diff
@@ -0,0 +1,86 @@
...
+        &'a self,
+        recv: &'static mut [u8; 64],
+    ) -> Result<(), (ReturnCode, &'static mut [u8; 64])>;
+
+    /// Cancels a receive called by `receive_buffer()`. Once called the
+    /// `packet_received()` callback won't occur and no data will be
+    /// received until `receive_buffer()` is called.
+    /// On success (a receive was cancelled) returns the buffer passed to
+    /// `receive_buffer()`.
+    /// On failure returns a return code
```
ppannuto, line 84: I think the cancellation semantics are cleaner here, only question is the "on failure returns a return code" --> will all failures still results in eventual callbacks later? What return codes are valid here?

ppannuto (changes requested):

kernel/src/hil/usb_hid.rs
```diff
@@ -0,0 +1,86 @@
...
+        endpoint: usize,
+    );
+
+    /// Called when checking if we can start a new receive operation.
+    /// Should return true if we are ready to receive and not currently
+    /// in the process of receiving anything. That is if we are currently
+    /// idle.
+    /// If there is an outstanding call to receive, a callback already
+    /// waiting to be called then this will return false.
+    fn can_receive(&'a self) -> bool;
```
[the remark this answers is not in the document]
phil-levis: Given the synchronous nature of the Tock kernel, I don't think you can have TOCTOU bugs; it would have to be you receive a callback preemptively. 

The presence of a call like this allows the client to push out replicated internal state. Suppose the client wants to know if there's currently a reception (e.g., whether it can power something down). It can replicate this part of the USB state machine internally by tracking calls to receive, or it can rely on calling this method. It might want to know the state without starting a new reception.

kernel/src/hil/usb_hid.rs
phil-levis: @ppannuto You have the semantics right: SUCCESS means there will be no callback, an error code means there will be (and the callback may return a SUCCESS or an ECANCEL).

kernel/src/hil/usb_hid.rs
phil-levis: The return codes should be specified as well as that a callback will be invoked.

Note one issue here: what if there is no call outstanding? You don't want to return a buffer and you don't want to signal a callback. This is why the radio moved towards cancel basically always invoking a callback: SUCCESS meant there was no outstanding call. This also simplified buffer management as there's a single call path to get the buffer back.

kernel/src/hil/usb_hid.rs
phil-levis: Note one issue here: what if there is no call outstanding? You don't want to return a buffer and you don't want to signal a callback. This is why the radio moved towards cancel basically always invoking a callback: SUCCESS meant there was no outstanding call. This also simplified buffer management as there's a single call path to get the buffer back.

kernel/src/hil/usb_hid.rs
alistair23: I have updated this to match the `send_cancel()`

kernel/src/hil/usb_hid.rs
```diff
@@ -0,0 +1,86 @@
...
+        buffer: &'static mut [u8; 64],
+        len: usize,
+        endpoint: usize,
+    );
+
+    /// Called when a packet has been finished transmitting.
+    /// This will return the buffer passed into `send_buffer()` as well as
+    /// the endpoint where the data was sent. The `len` argument contains
+    /// the number of bytes sent. If this is less then the size of the
+    /// buffer `result` will contain the `ESIZE` error.
```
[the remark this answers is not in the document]
alistair23: The [USB HID spec](https://www.usb.org/sites/default/files/hid1_11.pdf) says: 

> A transaction may be made up of multiple packets (token, data, handshake) but is limited in size to 8 bytes for low-speed devices and 64 bytes for high-speed devices.

So for high speed devices it should always be 64-bytes.

I guess here we only support high speed devices (USB 2.0+), but I really don't think there are a lot of low speed only devices.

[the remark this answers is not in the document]
alistair23: Also, we could get transactions of a size less then 64-bytes for high-speed. But using a buffer less then 64-bytes just seems to be at a high risk of loosing data, so we don't allow it.

kernel/src/hil/usb_hid.rs
gendx: I think the following semantics would make sense:
- `Ok(buffer)` => A buffer was already set, and could successfully be cancelled.
- `Err(ECANCEL)` => A buffer was already set, but couldn't be cancelled as the send operation is in progress. A callback returning the buffer will be issued later.
- `Err(EALREADY)` => No buffer is set at the moment (i.e. either no buffer was ever set, or the previously set buffer was already returned by a callback or cancel operation). I think `EALREADY` is the closest error code among what we already have (i.e. "already cancelled").

gendx: > SUCCESS meant there was no outstanding call.

I'm not sure it's better than `EALREADY`, because even though the result is a success (nothing is left pending), no "cancelling" happened. If the caller expects to get back a buffer at this point (likely due to an invalid state tracking by the caller - i.e. a bug in the caller's code), they cannot just ignore a "success".

gendx:
I started porting #2125 to this HIL, and should have feedback tomorrow.

kernel/src/hil/usb_hid.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// idle.
+    /// If there is an outstanding call to receive, a callback already
+    /// waiting to be called then this will return false.
+    fn can_receive(&'a self) -> bool;
+}
+
+pub trait UsbHid<'a> {
+    /// Sets the buffer to be sent and starts a send transaction.
+    /// On success returns the number of bytes that will be sent.
+    /// On failure returns an error code and the buffer passed in.
```
phil-levis, line 46: How is the number of bytes that will be sent calculated?

phil-levis:
> 
> 
> I started porting #2125 to this HIL, and should have feedback tomorrow.

Great. If the traits look good then we can unblock this and move forward with 1.6.

kernel/src/hil/usb_hid.rs
alistair23: It should always be 64 for high speed or 8 for low speed devices

hudson-ayers: 64 is for full speed, right? I thought high speed was 1024? (not that I have ever heard of a high speed HID device)

alistair23: High speed is USB 2.0, the HID spec says:

> A transaction may be made up of multiple packets (token, data, handshake) but is limited in size to 8 bytes for low-speed devices and 64 bytes for high-speed devices.

hudson-ayers: Oh, interesting, thanks

gendx:
I managed to port #2125 to use this HIL and make it compile. I still need to confirm this with manual testing.

In the meantime, I have some comments and questions that we need to resolve.
- What is the purpose of the `len` parameter in the `Client` HIL? We only use `[u8; 64]` so I don't see in which scenario the length is not 64 bytes. I don't see either what the client should do if a packet received/transmitted isn't 64 bytes.
- What is the purpose of the `endpoint` parameter in the `Client` HIL? There is no way to specify which endpoint to use in the `UsbHid` HIL, so as far as the client is concerned, the endpoints being used are an implementation detail below the HIL - and there is no value to expose them.
- What are the semantics of the return codes, and what return codes can be expected in which scenarios? We need to clearly define them in the HIL.
  - In terms of `packet_received|transmitted`, the code can be `SUCCESS`, or `ECANCEL` if it was cancelled. The `ECANCEL` case only happens if the client requests both reception and transmission: the implementation automatically cancels a transmission if a packet is received first, and vice versa cancels a reception if a packet is transmitted first. This allows the userspace app to take a suitable decision. For example, the app wanted to send a packet X, but a cancel message was sent by the host instead, and the app now wants to reply to this cancel message with another packet Y rather than with packet X which is obsolete. Does this make sense?
  - As far as explicit cancellation is concerned, here is what I implemented - does it make sense?
    - Cancelling a "receive" operation can only be a `SUCCESS` (if a buffer was passed down the HIL) or an `EALREADY` (if the buffer was already returned back up the HIL, or if no buffer was ever passed down). There is no "partial" state - if the USB controller started to receive some packet, it will be waiting there until the next request to receive a packet is issued.
    - Cancelling a "send" operation can be a `SUCCESS` (if a buffer was passed down the HIL, but not yet consumed by the controller), a `EBUSY` (if a buffer was passed down the HIL, and the controller started sending it, but no "transmitted" interrupt was issued yet) or an `EALREADY` (if the buffer was already returned back up the HIL, or if no buffer was ever passed down).

alistair23:
@phil-levis I address the comments we talked about today.

`packet_transmitted()` no longer returns the len, although `packet_received()` still does.

`send_buffer()` takes a length argument if the caller wants to send a buffer less then 64-bytes

alistair23:
>     * What is the purpose of the `len` parameter in the `Client` HIL? We only use `[u8; 64]` so I don't see in which scenario the length is not 64 bytes. I don't see either what the client should do if a packet received/transmitted isn't 64 bytes.

I have remove the `len` from the transmitted. In the receive callback you could have received data less then the maximum packet size.

>     * What is the purpose of the `endpoint` parameter in the `Client` HIL? There is no way to specify which endpoint to use in the `UsbHid` HIL, so as far as the client is concerned, the endpoints being used are an implementation detail below the HIL - and there is no value to expose them.

I think the endpoint could be useful information for some implementations, especially for packet received if there are multiple endpoints.


I will update the error codes to be a little more explicit.

phil-levis:
I read up on the HID exchange. I found these guides to be really good references when I was doing the USB stack on H1B:

<http://www.usbmadesimple.co.uk/ums_5.htm>

HID devices can send any number of bytes of data in a packet, from 1-64; the HID report descriptor for a given endpoint defines what this should be. However, it'll always send a full sized frame at the physical layer. The maximum size of the frame is 64 bytes, but this can be configured with the wb_max_packet_size field of a device descriptor to be 8, 16, 32, or 64 bytes.

So I think there are three ways forward:
    1. Have separate traits for different max packet sizes: 8, 16, 32, and 64 buffers.
    2. Have one trait, which takes a slice and can send less than a maximum packet size (it'll just fill into a packet sized buffer), and which you can query for the maximum packet size. It does not do fragmentation/assembly: if you try to send/receive more than the maximum packet size you get ESIZE.
    3. Have one trait, which takes a slice of arbitrary length. You can query for the maximum packet size (since it's useful). If you send less, it sends a partially filled packet. If you send more, it fragments into the number of needed maximum sized packets. It fills a receive buffer with the requested number of bytes from multiple packets.

Option 2 is what the tock-on-titan code does.

gendx:
> So I think there are three ways forward:
> 1. Have separate traits for different max packet sizes: 8, 16, 32, and 64 buffers. A caller passes an array of that size.

Option 4 (the simplest IMO) - have only a single trait supporting 64-byte buffers. Since Tock is the device, we can deliberately choose to only support 64-byte buffers (at least for now), and not have to worry about the packet length or handling `ESIZE` errors.

For now, supporting various lengths comes with extra complexity, but this extra complexity doesn't come with any benefits as we only need 64-byte buffers.

We can revisit that in the future if a needs arise, but I think the HIL should be kept as simple as possible for what we need to express.

gendx:
> I think the endpoint could be useful information for some implementations, especially for packet received if there are multiple endpoints.

I don't think it's useful information as a callback if the client doesn't have a means to select an endpoint the other way around when triggering a send/receive operation. Either endpoints are managed above the HIL or below the HIL, but it doesn't make sense to half expose them.

For now, it seems that endpoints are managed below the HIL, as the endpoint descriptors are hard-coded below the HIL, so endpoints are an implementation detail that doesn't need to be exposed back when a callback comes back up the HIL.

And once again, I think we should keep the HIL as simple as possible for what we need to support, in order to avoid unnecessary complexity in checking endpoint numbers etc. We can revisit that later if a need arises for managing endpoint numbers above the HIL, or for having multiple endpoints, etc.

alistair23:
I went with option 1, I have pushed an update doing that. It seems the simplest as implementations are likely going to always use a single size.

phil-levis:
> 
> 
> I went with option 1, I have pushed an update doing that. It seems the simplest as implementations are likely going to always use a single size.

The issue isn't implementations; it's users. There are two sides to an API. This approach means that any given client will only work for a single max packet size. I.e., it binds a client to particular implementations.

alistair23:
Yes, that is the case. A HID device needs to specify at start-up the maximum size it supports so it isn't dynamically changing. I don't see an issue with having an implementation always use the same maximum size. Currently this is a hard coded buffer passed to the host so we don't support changing the max packet size.

phil-levis:
> 
> 
> I don't see an issue with having an implementation always use the same maximum size. Currently this is a hard coded buffer passed to the host so we don't support changing the max packet size.

I think it's fine for a HID implementation to always use the same maximum size (it'll be in the report descriptor); the issue is the client of the HID. USB is pretty clear about wanting to separate out the application-level data sizes and the physical frame size. By having this be the API, nothing written to use this HIL could use a different maximum packet size. This violates the protocol layering. 

That being said, I think that a reasonable way to handle this is to keep the HIL as you have it now, then later have a different HIL that does fragmentation/assembly, which can sit on top of a 64, 32, 16, or 8. We could then transition stuff like CTAP to use this more general abstraction. So this HIL is about physical framing, rather than application framing. If you rename the current trait to UsbHid64 I'm happy.

gendx:
> If you rename the current trait to UsbHid64 I'm happy.

+1 to that.

> That being said, I think that a reasonable way to handle this is to keep the HIL as you have it now, then later have a different HIL that does fragmentation/assembly, which can sit on top of a 64, 32, 16, or 8. We could then transition stuff like CTAP to use this more general abstraction.

We could, although unless there is value in implementing fragmentation/assembly, this is also an extra layer of complexity, and complexity has its drawbacks, among others (1) more opportunities for bugs (2) more maintenance/review cost (3) larger binary size.

I'd propose to revisit this if/when there is a more concrete driver proposal (beyond the CTAP/HID use case) that needs another size.

alistair23:
I am still confused. The client knowing the packet size seems reasonable. For example the CTAP client knows that the packet will always be 64 bytes.

As for renaming doesn't the types make the size clear?

phil-levis:
> 
> 
> > If you rename the current trait to UsbHid64 I'm happy.
> 
> +1 to that.
> 
> > That being said, I think that a reasonable way to handle this is to keep the HIL as you have it now, then later have a different HIL that does fragmentation/assembly, which can sit on top of a 64, 32, 16, or 8. We could then transition stuff like CTAP to use this more general abstraction.
> 
> We could, although unless there is value in implementing fragmentation/assembly, this is also an extra layer of complexity, and complexity has its drawbacks, among others (1) more opportunities for bugs (2) more maintenance/review cost (3) larger binary size.
> 
> I'd propose to revisit this if/when there is a more concrete driver proposal (beyond the CTAP/HID use case) that needs another size.

Sure -- the question is whether the complexity is in this single, shared implementation or whether it's done by each client.

When it comes to API design, I'm generally against on-demand design; it introduces an activation entry for changing APIs. E.g.,

phil-levis:
> 
> 
> I am still confused. The client knowing the packet size seems reasonable. For example the CTAP client knows that the packet will always be 64 bytes.

Does CTAP specify this? What if it is on a high speed link? 

> 
> As for renaming doesn't the types make the size clear?

What if we need a trait with another size? How will we name it?

alistair23:
The CTAP size is technically vendor defined, but the reference implementation is 64 bytes: https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html#usb-descriptors

If you need a different size just use the `UsbHid` but with a different `UsbHidType`. For example for 32 bytes: `hil::usb_hid::UsbHid<'a, [u8; 32]>`

phil-levis:
> If you need a different size just use the `UsbHid` but with a different `UsbHidType`. For example for 32 bytes: `hil::usb_hid::UsbHid<'a, [u8; 32]>`

Ah, sorry, I missed this. A type is good too!

phil-levis (approved):

phil-levis:
> The CTAP size is technically vendor defined, but the reference implementation is 64 bytes: https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html#usb-descriptors

*shrug* I am pretty confident that CTAP is supposed to be independent of this, so should be able to work on all sizes op wMaxPacketSize. So this means Tock should be able to express all of them but it's fine we currently only have an implementation for the one that every implementation we've seen uses.

phil-levis:
@ppannuto Just waiting for you to clear the change request -- I think @alistair23 addressed all of the issues you raised.

ppannuto (approved):
Looks good to me -- glad to see the error states spelled out more clearly, I think that's a pretty valuable / important contract at the HIL layer

gendx:
Overall this looks fine. A few comments:
- I like the new API with packet lengths known at compile time, and `ESIZE` looks like a good error code if the size is not as expected.
- The error codes for `send_buffer` are not documented. I assume the only error case is `EBUSY` if a buffer was already passed down the HIL - similarly to `receive_buffer`?
- I still don't see the value in having the endpoint numbers only in the `Client` trait. Although it's still unnecessary complexity IMO, we can live with it for now.

PR #2131, opened 2020-09-27
panic handler: don't panic if DEBUG_WRITER is not set

lschuermann:
### Pull Request Overview

During development of a new board one might panic prior to a debug
writer component being registered. Especially if no serial console is
working yet, seeing the blinking LEDs in the recognizable pattern is a
good first indicator.

If no debug writer component is set, the panic handler will itself
panic, resulting in a panic loop. The LED pattern will not be
displayed.

This checks whether the DEBUG_WRITER global variable is set and prints
a message that it should be set with `set_debug_writer_wrapper`
instead.

### Testing Strategy

This pull request was tested by running on an in-development board that does not call to `set_debug_writer_wrapper`.

### TODO or Help Wanted
N/A

### Documentation Updated

- N/A Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/kernel/src/debug.rs
+++ b/kernel/src/debug.rs
@@ -343,10 +343,12 @@ pub struct DebugWriter {
 /// needed so the debug!() macros have a reference to the object to use.
 static mut DEBUG_WRITER: Option<&'static mut DebugWriterWrapper> = None;
 
-pub unsafe fn get_debug_writer() -> &'static mut DebugWriterWrapper {
-    DEBUG_WRITER
-        .as_deref_mut()
-        .expect("Must call `set_debug_writer_wrapper` in board initialization.")
+unsafe fn try_get_debug_writer() -> Option<&'static mut DebugWriterWrapper> {
+    DEBUG_WRITER.as_deref_mut()
+}
+
+unsafe fn get_debug_writer() -> &'static mut DebugWriterWrapper {
+    try_get_debug_writer().expect("Must call `set_debug_writer_wrapper` in board initialization.")
 }
 
 /// Function used by board main.rs to set a reference to the writer.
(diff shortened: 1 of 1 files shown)


bradjc (dismissed):

hudson-ayers (approved):

kernel/src/debug.rs
```diff
@@ -343,10 +343,12 @@ pub struct DebugWriter {
 /// needed so the debug!() macros have a reference to the object to use.
 static mut DEBUG_WRITER: Option<&'static mut DebugWriterWrapper> = None;
 
+pub unsafe fn try_get_debug_writer() -> Option<&'static mut DebugWriterWrapper> {
```
ppannuto, line 346: I don't think this should be a `pub` function. I see that this mirrors the non-try version, but that really should not be `pub` either.. 

The underlying `DEBUG_WRITER` is pretty unsafe (and not a `pub` object). Constraining its access to this file seems valuable.

--

Could you remove `pub` here and while we're at it from the non-try variant?

ppannuto (changes requested):

kernel/src/debug.rs
lschuermann: Can do :+1:.

lschuermann: Thanks for the hint. I've pushed the changes and added the following to the commit message:

_Due to the static mutable debug writer being inherently unsafe, this also reduces its visibility._

Feel free to resolve this if you are comfortable with the result

hudson-ayers (approved):

alistair23 (approved):

ppannuto (approved):

PR #2137, opened 2020-10-02
Simplify grant implementation and fix soundness errors

### LINKED ISSUE #2135: Grant API is unsound?
After some investigation, I think the grant API is unsound.

The following code compiles in a Tock capsule today. Incidentally, it currently prints "1" , but I believe it could print "11" if no-aliasing optimizations were enabled in Rust (which has happened in the past and will happen in the future).

```rust
struct App {
    dummy: u32,
}
self.apps.enter(appid, |app, _| {
    let mutable_ref = &mut app.dummy;
    *mutable_ref = 10;
    self.apps.enter(appid, |app, _| {
        app.dummy = 0;
    }).unwrap();
    *mutable_ref += 1;
    kernel::debug!("dummy: {:?}", mutable_ref);
});
```

The problem here is that it is possible to enter a grant region that has already been entered -- `Grant::enter()` and `Grant::each()` do not check if the grant region being entered has already been entered when providing mutable access to the contents. 

https://github.com/tock/tock/blob/e9c5508d2debd1d63eff1d0cad94248dab85df8f/kernel/src/grant.rs#L310 is the particular problematic line of unsafe in `enter()`. 

https://github.com/tock/tock/blob/e9c5508d2debd1d63eff1d0cad94248dab85df8f/kernel/src/grant.rs#L68 is the problematic line of unsafe that is made accessible by `each()` (Confusingly, `each()` is not just syntactic sugar for `for app in apps.iter() { app.enter() }`).

Unfortunately, code already exists that utilizes this API in a way that could be unsound.

For example, in the ble_advertising_driver: https://github.com/tock/tock/blob/2acd4c56ab56b9d98c9501256b63594165a8e1ca/capsules/src/ble_advertising_driver.rs#L574

A closure passed to `Grant::enter()` contains a function call that calls `enter()` itself.

I found another example in the UDP driver: https://github.com/tock/tock/blob/e9c5508d2debd1d63eff1d0cad94248dab85df8f/capsules/src/net/udp/driver.rs#L531

I suspect there are others.

So far, I can think of three possible solutions:

1. Change the definition of `Grant` and `Grant::enter()` as follows:
```rust
pub struct Grant<const NUM_PROCS: u32, T: Default> {
    pub(crate) kernel: &'static Kernel,
    grant_num: usize,
    ptr: PhantomData<T>,
    accessed: [Cell<bool>; NUM_PROCS],
}
    pub fn enter<F, R>(&self, appid: AppId, fun: F) -> Result<R, Error>
    where
        F: FnOnce(&mut Borrowed<T>, &mut Allocator) -> R,
        R: Copy,
    {
        appid
            .kernel
            .process_map_or(Err(Error::NoSuchApp), appid, |process| {
                if let Some(untyped_grant_ptr) = process.get_grant_ptr(self.grant_num) {
                        // ... allocate code omitted
                    } else {
                        untyped_grant_ptr as *mut T
                    };
                    if !accessed[appid.idx].get() {
                         accessed[appid.idx].set(true);
                         // Dereference the typed GrantPointer to make a GrantRegion
                         // reference.
                        let region = unsafe { &mut *typed_grant_pointer };
                        let mut borrowed_region = Borrowed::new(region, appid);

                        // Call the passed in closure with the borrowed grant region.
                        let res = fun(&mut borrowed_region, &mut allocator);
                        accessed.set[appid.idx](false);
                        Ok(res)
                    } else {
                        Err(Error::Busy)
                    }
                } else {
                    Err(Error::InactiveApp)
                }
            })
    }
```
This requires capsules to detect and handle reentrant calls at runtime.

2. Change `grant.enter()` to provide the passed closure with an immutable reference to the type stored in the grant, and fields of that type can only be mutated if they are behind some type of cell. So you would see code like this:

```rust
struct App {
    num: Cell<u32>,
    buf: TakeCell<'static, [u8]>,
    appslice: RefCell<AppSlice<Shared, u8>>,
}
```
This allows for primitives to be accessed and modified safely despite multiple simultaneous accesses to grants (I think Cell should be sufficient because of the single threaded nature of the Tock Kernel). For things like `AppSlice`, simultaneous access would lead to panics rather than unsoundness, thanks to the semantics of `RefCell`.

3. Make `grant.enter()` and `grant.each()` require `&mut self` instead of `&self`, to make reentrant calls impossible. This would probably require large scale changes to capsule code, but maybe could be made easier by storing the Grant in something like a `TakeCell`?
- This approach has the significant downside that it prevents entering grant regions for different applications simultaneously, despite it being safe to do so.

alevy:
### Pull Request Overview

This pull request addresses #2135 and partially the issue that #2052 addresses:

- Grants do not currently prevent re-entrancy which, in turn, allows mutable aliasing of Rust memory (that happens to be located in grant regions)
- The Allocator interface (which is currently totally unused except for @daboross's in progress BLE work) has similar, but harder to fix issues.

This pull request does three things:

1. It removes the `Allocator` struct and replaces the closure `enter` expects to take a `()` instead. This works because the allocator is totally unused at the moment, so users of `enter` don't care about the type. We _should_ bring this back and #2052 is likely close to the solution, but as an initial step, removing it is good too.

2. It attempts to dramatically simplify the grant module, leaving only four types:
    - `Borrowed` - a wrapper around a mutable reference to reified Grant memory and an appid that's yielded when "entering" a grant
    - `AppliedGrant` which is type that contains a grant that's guaranteed to be allocated and usable
    - `Grant` the main type a user interacts with. `Grant`s differ from `AppliedGrant` in that calling `enter` may initialize the grant for a particular process if it has not yet been allocated.
    - `Iter` - a helper type for iterating over all processes for a particular grant.

    In addition, it moves all use of `unsafe` to the creation of `AppliedGrant`s and implements everything else in terms of creation and access to `AppliedGrant`s.

3. Finally, it fixes #2135 by applying `TakeCell`-like semantics to the Grant's `enter` API. Entering a grant leaves behind a placeholder `0xffffffff` value in place of the pointer that means it's currently being used. Re-entrance to a grant checks for this value and will fail if it is present.
  

### Testing Strategy

So far tested on Imix and Hail with the `alarm` driver (which had reentrant code) and the UDP driver (which has reentrant code). Without fixing these capsules, both now fail, but can be fixed to work.

### TODO or Help Wanted

  * [ ] Document the grant module better
  * [ ] Find and fix other incorrectly written capsules

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -1,35 +1,215 @@
 //! Data structure to store a list of userspace applications.
 
-use crate::callback::AppId;
-use crate::process::{Error, ProcessType};
-use crate::sched::Kernel;
 use core::marker::PhantomData;
 use core::mem::{align_of, size_of};
 use core::ops::{Deref, DerefMut};
 use core::ptr::{slice_from_raw_parts_mut, write, NonNull};
 
-/// Region of process memory reserved for the kernel.
-pub struct Grant<T: Default> {
-    pub(crate) kernel: &'static Kernel,
-    grant_num: usize,
-    ptr: PhantomData<T>,
+use crate::callback::AppId;
+use crate::process::{Error, ProcessType};
+use crate::sched::Kernel;
+
+/// Type that indicates a grant region has been entered and borrowed.
+/// This is passed to capsules when they try to enter a grant region.
+pub struct Borrowed<'a, T: 'a + ?Sized> {
+    data: &'a mut T,
+    appid: AppId,
 }
 
-pub struct AppliedGrant<T> {
-    appid: AppId,
-    grant: NonNull<T>,
+impl<'a, T: 'a + ?Sized> Borrowed<'a, T> {
+    fn new(data: &'a mut T, appid: AppId) -> Borrowed<'a, T> {
+        Borrowed {
+            data: data,
+            appid: appid,
+        }
+    }
+
+    pub fn appid(&self) -> AppId {
+        self.appid
+    }
+}
+
+impl<'a, T: 'a + ?Sized> Deref for Borrowed<'a, T> {
+    type Target = T;
+    fn deref(&self) -> &T {
+        self.data
+    }
+}
+
+impl<'a, T: 'a + ?Sized> DerefMut for Borrowed<'a, T> {
+    fn deref_mut(&mut self) -> &mut T {
+        self.data
+    }
+}
+
+pub struct AppliedGrant<'a, T: 'a> {
+    process: &'a dyn ProcessType,
+    grant_num: usize,
+    grant: &'a mut T,
     _phantom: PhantomData<T>,
 }
 
-impl<T> AppliedGrant<T> {
+impl<'a, T: Default> AppliedGrant<'a, T> {
+    fn get_or_allocate(grant: &Grant<T>, process: &'a dyn ProcessType) -> Result<Self, Error> {
+        // Here is an example of how the grants are laid out in a
+        // process's memory:
+        //
+        // Mem. Addr.
+        // 0x0040000  ┌────────────────────
+        //            │   GrantPointer0 [0x003FFC8]
+        //            │   GrantPointer1 [0x003FFC0]
+        //            │   ...
+        //            │   GrantPointerN [0x0000000 (NULL)]
+        //            ├────────────────────
+        //            │   Process Control Block
+        // 0x003FFE0  ├────────────────────
+        //            │   GrantRegion0
+        // 0x003FFC8  ├────────────────────
+        //            │   GrantRegion1
+        // 0x003FFC0  ├────────────────────
+        //            │
+        //            │   --unallocated--
+        //            │
+        //            └────────────────────
+        //
+        // An array of pointers (one per possible grant region)
+        // point to where the actual grant memory is allocated
+        // inside of the process. The grant memory is not allocated
+        // until the actual grant region is actually used.
+        //
+        // A `GrantPointer` can be set to all 1s (0xFFFFFFFFF) as a flag
+        // that that grant is currently entered (aka being accessed). The actual
+        // pointer address is restored when the access finishes.
+        //
+        // This function provides the app access to the specific
+        // grant memory, and allocates the grant region in the
+        // process memory if needed.
+
+        // Get the GrantPointer to start. Since process.rs does not know
+        // anything about the datatype of the grant, and the grant
+        // memory may not yet be allocated, it can only return a `*mut
+        // u8` here. We will eventually convert this to a `*mut T`.
+        let grant_num = grant.grant_num;
+        let appid = process.appid();
+        if let Some(untyped_grant_ptr) = process.get_grant_ptr(grant_num) {
+            // If the grant pointer is NULL then the memory for the
+            // GrantRegion needs to be allocated. If the grant pointer is all 1s
+            // then the grant is already enetered and we return an error. Otherwise, we can
+            // convert the pointer to a `*mut T` because we know we
+            // previously allocated enough memory for type T.
+            unsafe {
+                let typed_grant_pointer = if untyped_grant_ptr.is_null() {
+                    // Allocate space in the process's memory for
+                    // something of type `T` for the grant.
+                    //
+                    // Note: This allocation is intentionally never
+                    // freed.  A grant region is valid once allocated
+                    // for the lifetime of the process.
+                    let alloc_size = size_of::<T>();
+                    let new_region =
+                        appid
+                            .kernel
+                            .process_map_or(Err(Error::NoSuchApp), appid, |process| {
+                                process.alloc(alloc_size, align_of::<T>()).map_or(
+                                    Err(Error::OutOfMemory),
+                                    |buf| {
+                                        // Convert untyped `*mut u8` allocation to allocated type
+                                        let ptr = NonNull::cast::<T>(buf);
+
+                                        Ok(ptr)
+                                    },
+                                )
+                            })?;
+
+                    // We use `ptr::write` to avoid `Drop`ping the uninitialized memory in
+                    // case `T` implements the `Drop` trait.
+                    write(new_region.as_ptr(), T::default());
+
+                    // Update the grant pointer in the process. Again,
+                    // since the process struct does not know about the
+                    // grant type we must use a `*mut u8` here.
+                    process.set_grant_ptr(grant_num, new_region.as_ptr() as *mut u8);
+
+                    // The allocator returns a `NonNull`, we just want
+                    // the raw pointer.
+                    new_region.as_ptr()
+                } else if untyped_grant_ptr == (!0 as *mut u8) {
+                    // Grant region currently entered, cannot enter again.
+                    return Err(Error::AlreadyInUse);
+                } else {
+                    // Grant region previously allocated, just convert the
+                    // pointer.
+                    untyped_grant_ptr as *mut T
+                };
+
+                Ok(AppliedGrant {
+                    process: process,
+                    grant_num: grant.grant_num,
+                    grant: &mut *(typed_grant_pointer as *mut T),
+                    _phantom: PhantomData,
+                })
+            }
+        } else {
+            Err(Error::InactiveApp)
+        }
+    }
+
+    /// Return an `AppliedGrant` for a grant in a process if that
+    /// grant has already been allocated.
+    ///
+    /// On `Err()`, returns `Err(None)` if the grant has not been allocated
+    /// for this process. Returns `Err(Some(Error))` with an appropriate
+    /// error otherwise.
+    fn get_if_allocated(
+        grant: &Grant<T>,
+        process: &'a dyn ProcessType,
+    ) -> Result<Self, Option<Error>> {
+        process
+            .get_grant_ptr(grant.grant_num)
+            .map_or(Err(None), |grant_ptr| {
+                if grant_ptr.is_null() {
+                    // Grant has not been allocated.
+                    Err(None)
+                } else if grant_ptr == (!0 as *mut u8) {
+                    // A grant pointer set to 0xFFFFFFFF is a flag signaling
+                    // that the grant has already been entered.
+                    Err(Some(Error::AlreadyInUse))
+                } else {
+                    Ok(AppliedGrant {
+                        process: process,
+                        grant_num: grant.grant_num,
+                        grant: unsafe { &mut *(grant_ptr as *mut T) },
+                        _phantom: PhantomData,
+                    })
+                }
+            })
+    }
+
+    /// Run a function with access to the contents of the grant region.
+    /// This is "entering" the grant region, and the _only_ time when
+    /// the contents of a grant region can be accessed.
     pub fn enter<F, R>(self, fun: F) -> R
     where
-        F: FnOnce(&mut Owned<T>, &mut Allocator) -> R,
+        F: FnOnce(&mut Borrowed<T>, &mut Allocator) -> R,
         R: Copy,
     {
-        let mut allocator = Allocator { appid: self.appid };
-        let mut root = Owned::new(self.grant, self.appid);
-        fun(&mut root, &mut allocator)
+        let mut allocator = Allocator {
+            appid: self.process.appid(),
+        };
+        let mut root = Borrowed::new(self.grant, self.process.appid());
+        // Mark the grant region as entered by replacing its grant pointer with
+        // all 1s. This allows us to check whether the grant region is already entered.
+        unsafe {
+            self.process.set_grant_ptr(self.grant_num, !0 as *mut u8);
+        }
+        let res = fun(&mut root, &mut allocator);
+        // Restore the grant pointer indicating that we are no longer accessing
+        // the grant.
+        unsafe {
+            self.process
+                .set_grant_ptr(self.grant_num, root.data as *mut _ as *mut u8);
+        }
+        res
     }
 }
 
@@ -58,6 +238,10 @@ impl<T: ?Sized> DynamicGrant<T> {
     /// If the app has since been restarted or crashed, or the memory is otherwise no longer
     /// present, then this function will not call the given closure, and will
     /// instead directly return `Err(Error::NoSuchApp)`.
+    ///
+    /// Because this function requires `&mut self`, it should be impossible to access
+    /// the inner data of a given `DynamicGrant` reentrantly. Thus the reentrance detection
+    /// we use for non-dynamic grants is not needed here.
     pub fn enter<F, R>(&mut self, fun: F) -> Result<R, Error>
     where
         F: FnOnce(Borrowed<'_, T>) -> R,
@@ -76,39 +260,8 @@ pub struct Allocator {
     appid: AppId,
 }
 
-pub struct Owned<T: ?Sized> {
-    data: NonNull<T>,
-    appid: AppId,
-}
-
-impl<T: ?Sized> Owned<T> {
-    fn new(data: NonNull<T>, appid: AppId) -> Owned<T> {
-        Owned {
-            data: data,
-            appid: appid,
-        }
-    }
-
-    pub fn appid(&self) -> AppId {
-        self.appid
-    }
-}
-
-impl<T: ?Sized> Deref for Owned<T> {
-    type Target = T;
-    fn deref(&self) -> &T {
-        unsafe { self.data.as_ref() }
-    }
-}
-
-impl<T: ?Sized> DerefMut for Owned<T> {
-    fn deref_mut(&mut self) -> &mut T {
-        unsafe { self.data.as_mut() }
-    }
-}
-
 impl Allocator {
-    /// Allocates a new owned grant initialized using the given closure.
+    /// Allocates a new dynamic grant initialized using the given closure.
     ///
     /// The closure will be called exactly once, and the result will be used to
     /// initialize the owned value.
@@ -170,21 +323,6 @@ impl Allocator {
         }
     }
 
-    /// Like `alloc`, but the caller is responsible for free-ing the allocated
-    /// memory, as it is not wrapped in a type that implements `Drop`.
-    ///
-    /// In contrast to `alloc_raw`, this method does initialize the returned
-    /// memory.
-    unsafe fn alloc_default_unowned<T: Default>(&mut self) -> Result<NonNull<T>, Error> {
-        let ptr = self.alloc_raw()?;
-
-        // We use `ptr::write` to avoid `Drop`ping the uninitialized memory in
-        // case `T` implements the `Drop` trait.
-        write(ptr.as_ptr(), T::default());
-
-        Ok(ptr)
-    }
-
     /// Allocates uninitialized memory appropriate to store a `T`, and returns a
     /// pointer to said memory. The caller is responsible for both initializing the
     /// returned memory, and dropping it properly when finished.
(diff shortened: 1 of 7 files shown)


hudson-ayers:
One thing to draw attention to is that `iter()` now silently skips processes that are already entered within a given grant, without providing any errors. This makes it easy to accidentally write helper functions that work differently when called from within a closure passed to a grant.

For example, the 802.15.4 stack provides a helper function `get_next_tx_if_idle()` that iterates through all grant regions to find the next pending transmission. This is called in several places, but one place where it is called is during the normal flow for synchronously sending a packet, where it is called from within a closure passed to a grant region, as a result, the next transmission is now not found. I worry that we are unlikely to catch all of the paths that could lead to this bug in Tock today just by performing our usual release testing.

Unfortunately, I am not sure I can think of a better solution. Panicking seems wrong, as often when `iter()` is called skipping entered processes is desirable. Returning an error seems wrong for the same reason. Adding an `iter_all()` function that panics if it skips a process might help, but it would be hard to enforce callers must use it every time they actually *need* to iterate over all processes.

bradjc:
> One thing to draw attention to is that `iter()` now silently skips processes that are already entered within a given grant, without providing any errors. This makes it easy to accidentally write helper functions that work differently when called from within a closure passed to a grant.

I think this is a big change and important issue. It doesn't seem like it should be valid for iter to skip apps that have setup grant regions set up as that seems like a really difficult API to program against.

hudson-ayers:
I pushed a commit that includes fixes for places where the UDP and 15.4 drivers re-entered the grant.

I also pushed a commit that removes a `Copy` bound from the return type of the closure passed to `Grant::enter()` or `AppliedGrant::enter()`. I can't see a reason this bound was useful, and it made it harder to write code that does not re-enter grants. However I would like some verification that the bound is not in fact useful.

hudson-ayers:
Pushed fixes for the 15.4, BLE, and UDP drivers, which each had grant re-entrance. Tested a modified version that panics if grant reentrance is detected on Imix and nrf52840dk with the following apps:

- udp_send
- udp_rx
- udp_virt_test
- radio_rx
- radio_tx
- blink
- nonvolatile_storage
- imix
- ble_advertising

There are now no failures.

bradjc:
How many instances of calling `iter()` from inside of a grant do you think there are? Is that a common pattern?

hudson-ayers:
> How many instances of calling `iter()` from inside of a grant do you think there are? Is that a common pattern?

It's pretty rare. This PR has fixed every instance that I could find by running test apps on Imix (I used a modified iter() that panics when called from inside a grant)

lschuermann:
> One thing to draw attention to is that iter() now silently skips processes that are already entered within a given grant

Would iterating over `Option`s of grant regions be a viable solution? In this case an application would know whether a specific grant region is entered and be forced to handle such cases at the typesystem level (or, if it's confident enough, simply use `.expect(`). The overhead should be negligible.

hudson-ayers:
> > One thing to draw attention to is that iter() now silently skips processes that are already entered within a given grant
> 
> Would iterating over `Option`s of grant regions be a viable solution? In this case an application would know whether a specific grant region is entered and be forced to handle such cases at the typesystem level (or, if it's confident enough, simply use `.expect(`). The overhead should be negligible.

This is an option, but would require us to change every capsule that includes calls to `iter()` to the new API (which might be worth it!)

hudson-ayers:
To summarize some of the discussion from the call today (mostly myself, @alevy , @bradjc , @lschuermann ):

- having `iter()` return an `Iterator` over `Option<AppliedGrant>` makes sense in some ways (the types now encode that is possible the Iterator will not be able to enter each grant region), but has downsides:
  - Most of the time, people will just `map()` over the `Option`, which is really no different than a silent skip
  - The alternative is people use `unwrap()`, which we currently explicitly discourage in capsule code
  - An `Option` being `None` doesn't *really* communicate that the grant in question was already entered
  - This is still more of a fail-slow approach

- An alternative approach is to have any call of `iter()` when a grant region is entered fail immediately (either panic, or return an Error). This has advantages and disadvantages:
  - Fails fast and cleanly, so if code that does this is tested it should be quickly caught
  - There can still be cases where the failure scenario is not exercised in normal tests, but only in the case of multiple apps using a capsule, or only for particular interrupt orderings etc. In this case, it is bad that we have added the potential for lots of panics.
  - Returning an error requires checking all the grant regions before you actually return the iterator, which adds overhead and makes the code more complex
  - There are cases where iterating over all regions *except* the currently entered region is useful

- I do not think that a static checking approach is feasible. Tock requires multiple references to userspace drivers (1 for delivering system calls, another for delivering interrupts from a virtualizer).

- Having additional functions with more descriptive names might help handle different scenarios, but makes the API less clean and simple than before

hudson-ayers:
I updated this to panic on re-entrance by a call to `iter()` or `each()`. 

I added a function -- `iter_unentered_regions()` that can be used to enter only those grant regions which are not already entered. I did not find a need to use it anywhere, however, as we had already entered fixes to make just using `iter()` usable everywhere where this was an issue.

I also fixed an example of re-entrance in the proximity driver.

alevy:
Note that 2.0 drivers may not be fixed for this yet

kernel/src/grant.rs
```diff
@@ -352,34 +257,30 @@ pub struct Iter<'a, T: 'a + Default> {
         core::slice::Iter<'a, Option<&'static dyn ProcessType>>,
         fn(&Option<&'static dyn ProcessType>) -> Option<&'static dyn ProcessType>,
     >,
+    skips_allowed: bool,
```
bradjc, line 260:
```suggestion
    /// Whether this iterator must visit every grant region, or if
    /// it should skip grant regions which are already entered.
    /// A grant region cannot be entered twice. Therefore, if a capsule
    /// tries to iterate over all grant regions while inside of a grant region,
    /// the kernel will `panic!()` to prevent double entry. However, if this
    /// is set, then the iterator will skip over already entered grant regions,
    /// avoiding a panic but potentially creating subtly unintended code.
    skip_entered_grants: bool,
```

kernel/src/grant.rs
```diff
@@ -352,34 +257,30 @@ pub struct Iter<'a, T: 'a + Default> {
...
-            } else {
-                false
-            }
-        });
-
-        // Check if our find above returned another `AppId`, or if we hit the
-        // end of the iterator. If we found another app, try to access its grant
-        // region.
-        res.map_or(None, |process| self.grant.grant(process.appid()))
+        let skips_allowed = self.skips_allowed;
```
bradjc, line 271:
```suggestion
        let skip_entered_grants = self.skip_entered_grants;
```

kernel/src/grant.rs
```diff
@@ -352,34 +257,30 @@ pub struct Iter<'a, T: 'a + Default> {
...
+        self.subiter.find_map(
+            |process| match AppliedGrant::get_if_allocated(grant, process) {
+                Ok(ag) => Some(ag),
+                Err(None) => None,
+                Err(Some(_err)) => {
+                    if skips_allowed {
```
bradjc, line 277:
```suggestion
                    if skip_entered_grants {
```

kernel/src/grant.rs
```diff
@@ -216,132 +202,51 @@ impl<T: Default> Grant<T> {
...
             grant: self,
             subiter: self.kernel.get_process_iter(),
+            skips_allowed: false,
+        }
+    }
+
+    /// Get an iterator over all processes and their active grant regions for
+    /// this particular grant, except grant regions that are already currently
+    /// entered.
+    pub fn iter_unenterted_regions(&self) -> Iter<T> {
```
bradjc, line 245:
```suggestion
    pub fn iter_unentered_grants(&self) -> Iter<T> {
```

kernel/src/grant.rs
```diff
@@ -352,34 +257,30 @@ pub struct Iter<'a, T: 'a + Default> {
...
-        // region.
-        res.map_or(None, |process| self.grant.grant(process.appid()))
+        let skips_allowed = self.skips_allowed;
+        self.subiter.find_map(
+            |process| match AppliedGrant::get_if_allocated(grant, process) {
+                Ok(ag) => Some(ag),
+                Err(None) => None,
+                Err(Some(_err)) => {
+                    if skips_allowed {
+                        None
```
bradjc, line 278:
```suggestion
                        // This iter was created from `iter_unentered_grants()`,
                        // so we skip entered grants to avoid double entry.
                        None
```

kernel/src/grant.rs
```diff
@@ -352,34 +257,30 @@ pub struct Iter<'a, T: 'a + Default> {
...
+                    } else {
+                        panic!("Attempted to re-enter a grant region.")
```
bradjc, line 280:
```suggestion
                        // Panic if the capsule asked us to enter an already entered
                        // grant region. This preserves type correctness (but does crash
                        // the system).
                        //
                        // This panic represents a tradeoff. While it is undesirable to have the
                        // potential for a runtime crash in this grant region code, it balances
                        // usability with type correctness. The challenge is that calling `self.apps.iter()`
                        // is a common pattern in capsules to access the grant region of every
                        // app that is using the capsule, and sometimes it is intuitive to call that
                        // inside of a `self.apps.enter(app_id, |app| {...})` closure. However, `.enter()`
                        // means that app's grant region is entered, and then a naive `.iter()` would
                        // re-enter the grant region and cause undefined behavior. We considered
                        // different options to resolve this.
                        //
                        // 1. Have `.iter()` only iterate over grant regions which are not entered.
                        //    This avoids the bug, but could lead to unexpected behavior, as `self.apps.iter()` will
                        //    do different things depending on where in a capsule it is called.
                        // 2. Have the compiler detect when `.iter()` is called when a grant region has
                        //    already been entered. We don't know of a viable way to implement this.
                        // 3. Panic if `.iter()` is called when a grant is already entered.
                        //
                        // We decided on option 3 because it balances minimizing surprises (`self.apps.iter()`
                        // will always iterate all grants) while also protecting against the bug. We expect
                        // that any code that attempts to call `self.apps.iter()` after calling `.enter()` will
                        // immediately encounter this `panic!()` and have to be refactored before any tests
                        // will be successful. Therefore, this `panic!()` should only occur at development/testing
                        // time.
                        //
                        // ## How to fix this error
                        //
                        // If you are seeing this panic, you need to refactor your capsule to not
                        // call `.iter()` or `.each()` from inside a `.enter()` closure. That is, you
                        // need to close the grant region you are currently in before trying to iterate
                        // over all grant regions.
                        panic!("Attempted to re-enter a grant region.")
```

kernel/src/grant.rs
```diff
@@ -216,132 +202,51 @@ impl<T: Default> Grant<T> {
...
+    }
+
+    /// Get an iterator over all processes and their active grant regions for
+    /// this particular grant, except grant regions that are already currently
+    /// entered.
+    pub fn iter_unenterted_regions(&self) -> Iter<T> {
+        Iter {
+            grant: self,
+            subiter: self.kernel.get_process_iter(),
+            skips_allowed: true,
```
bradjc, line 249:
```suggestion
            skip_entered_grants: true,
```

kernel/src/grant.rs
```diff
@@ -216,132 +202,51 @@ impl<T: Default> Grant<T> {
...
 
     /// Get an iterator over all processes and their active grant regions for
     /// this particular grant.
+    /// Calling this function when a grant region is currently entered
+    /// will lead to a panic.
     pub fn iter(&self) -> Iter<T> {
         Iter {
             grant: self,
             subiter: self.kernel.get_process_iter(),
+            skips_allowed: false,
```
bradjc, line 238:
```suggestion
            skip_entered_grants: false,
```

kernel/src/grant.rs
```diff
@@ -56,155 +27,170 @@ impl<T: ?Sized> Owned<T> {
...
-            .kernel
-            .process_map_or(Err(Error::NoSuchApp), self.appid, |process| {
-                process
-                    .alloc(alloc_size, align_of::<T>())
-                    .map_or(Err(Error::OutOfMemory), |buf| {
-                        // Convert untyped `*mut u8` allocation to allocated type
-                        let ptr = NonNull::cast::<T>(buf);
-
-                        Ok(ptr)
+    fn get_if_allocated(
```
bradjc, line 149:
```suggestion
    /// Return an `AppliedGrant` for a grant in a process if that
    /// grant has already been allocated.
    ///
    /// On `Err()`, returns `Err(None)` if the grant has not been allocated
    /// for this process. Returns `Err(Some(Error))` with an appropriate
    /// error otherwise.
    fn get_if_allocated(
```

kernel/src/grant.rs
```diff
@@ -56,155 +27,170 @@ impl<T: ?Sized> Owned<T> {
...
+        grant: &Grant<T>,
+        process: &'a dyn ProcessType,
+    ) -> Result<Self, Option<Error>> {
+        process
+            .get_grant_ptr(grant.grant_num)
+            .map_or(Err(None), |grant_ptr| {
+                if grant_ptr.is_null() {
+                    Err(None)
```
bradjc, line 157:
```suggestion
                    // Grant has not been allocated.
                    Err(None)
```

kernel/src/grant.rs
```diff
@@ -56,155 +27,170 @@ impl<T: ?Sized> Owned<T> {
...
+                } else if grant_ptr == (!0 as *mut u8) {
+                    Err(Some(Error::AlreadyInUse))
```
bradjc, line 159:
```suggestion
                    // A grant pointer set to 0xFFFFFFFF is a flag signaling
                    // that the grant has already been entered.
                    Err(Some(Error::AlreadyInUse))
```

kernel/src/grant.rs
```diff
@@ -56,155 +27,170 @@ impl<T: ?Sized> Owned<T> {
...
+        //            │
+        //            │   --unallocated--
+        //            │
+        //            └────────────────────
+        //
+        // An array of pointers (one per possible grant region)
+        // point to where the actual grant memory is allocated
+        // inside of the process. The grant memory is not allocated
+        // until the actual grant region is actually used.
+        //
```
bradjc, line 77:
```suggestion
        //
        // A `GrantPointer` can be set to all 1s (0xFFFFFFFFF) as a flag
        // that that grant is currently entered (aka being accessed). The actual
        // pointer address is restored when the access finishes.
        //
```

kernel/src/grant.rs
```diff
@@ -56,155 +27,170 @@ impl<T: ?Sized> Owned<T> {
...
+
+        // Get the GrantPointer to start. Since process.rs does not know
+        // anything about the datatype of the grant, and the grant
+        // memory may not yet be allocated, it can only return a `*mut
+        // u8` here. We will eventually convert this to a `*mut T`.
+        let grant_num = grant.grant_num;
+        let appid = process.appid();
+        if let Some(untyped_grant_ptr) = process.get_grant_ptr(grant_num) {
+            // If the grant pointer is NULL then the memory for the
+            // GrantRegion needs to be allocated. Otherwise, we can
```
bradjc, line 90:
```suggestion
            // GrantRegion needs to be allocated. If the grant pointer is all 1s
            // then the grant is already enetered and we return an error. Otherwise, we can
```

kernel/src/grant.rs
```diff
@@ -56,155 +27,170 @@ impl<T: ?Sized> Owned<T> {
...
+                    // Update the grant pointer in the process. Again,
+                    // since the process struct does not know about the
+                    // grant type we must use a `*mut u8` here.
+                    process.set_grant_ptr(grant_num, new_region.as_ptr() as *mut u8);
+
+                    // The allocator returns a `NonNull`, we just want
+                    // the raw pointer.
+                    new_region.as_ptr()
+                } else if untyped_grant_ptr == (!0 as *mut u8) {
+                    return Err(Error::AlreadyInUse);
```
bradjc, line 130:
```suggestion
                    // Grant region currently entered, cannot enter again.
                    return Err(Error::AlreadyInUse);
```

kernel/src/grant.rs
```diff
@@ -56,155 +27,170 @@ impl<T: ?Sized> Owned<T> {
...
-    data: &'a mut T,
-    appid: AppId,
-}
-
-impl<'a, T: 'a + ?Sized> Borrowed<'a, T> {
-    pub fn new(data: &'a mut T, appid: AppId) -> Borrowed<'a, T> {
-        Borrowed {
-            data: data,
-            appid: appid,
+    pub fn enter<F, R>(self, fun: F) -> R
```
bradjc, line 171:
```suggestion
    /// Run a function with access to the contents of the grant region.
    /// This is "entering" the grant region, and the _only_ time when
    /// the contents of a grant region can be accessed.
    pub fn enter<F, R>(self, fun: F) -> R
```

kernel/src/grant.rs
```diff
@@ -56,155 +27,170 @@ impl<T: ?Sized> Owned<T> {
...
+    where
+        F: FnOnce(&mut Borrowed<T>, ()) -> R,
+        R: Copy,
+    {
+        let mut root = Borrowed::new(self.grant, self.process.appid());
+        unsafe {
```
bradjc, line 177:
```suggestion
        // Mark the grant region as entered by replacing its grant pointer with
        // all 1s. This allows us to check whether the grant region is already entered.
        unsafe {
```

kernel/src/grant.rs
```diff
@@ -56,155 +27,170 @@ impl<T: ?Sized> Owned<T> {
...
-        self.appid
-    }
-}
-
-impl<'a, T: 'a + ?Sized> Deref for Borrowed<'a, T> {
-    type Target = T;
-    fn deref(&self) -> &T {
-        self.data
+        let res = fun(&mut root, ());
+        unsafe {
```
bradjc, line 181:
```suggestion
        // Restore the grant pointer indicating that we are no longer accessing
        // the grant.
        unsafe {
```

kernel/src/grant.rs
```diff
@@ -3,49 +3,20 @@
...
-    }
-}
-
-pub struct Allocator {
-    appid: AppId,
-}
-
-pub struct Owned<T: ?Sized> {
-    data: NonNull<T>,
+pub struct Borrowed<'a, T: 'a + ?Sized> {
```
bradjc, line 12: /// Type that indicates a grant region has been entered and borrowed.
/// This is passed to capsules when they try to enter a grant region.

bradjc (commented):
Updated the comments to not lose the wisdom behind this PR.

One question though: is it possible for `Borrowed` to set the grant pointer to 0xfffffff when it is created and to restore the grant pointer when it is finished?

hudson-ayers:
> One question though: is it possible for `Borrowed` to set the grant pointer to 0xfffffff when it is created and to restore the grant pointer when it is finished?

This could be done by implementing `Drop` on `Borrowed` I think. We would want to adapt `Borrowed` to store a reference to the `Process`, otherwise we would have to go through the whole `AppId` --> `kernel` --> `process_map_or` flow unnecessarily in order to access the grant pointer.

I'm not sure its worth the added overhead though, and in my experience reasoning about when `Drop` will be called can be challenging / lead to mistakes during refactoring.

hudson-ayers:
I had a hell of a time doing this rebase because of the changes from #2229 and because of how many commits in this PR modified `grant.rs` -- so I squashed all the commits in this PR, and then rebased.

Rebasing required bringing back the use of `Allocator` (which this PR had originally deleted) in order to keep the benefits from #2229. I don't think I introduced any additional soundness errors with `DynamicGrant`, because it requires `&mut self` to use, and thus does not need to be checked in the way that `Grant` / `AppliedGrant` does. But this rebase substantially changes this PR, so another review (by at least @alevy ) would be appreciated.

bradjc (approved):

bradjc:
Why haven't we merged this?

hudson-ayers:
Mostly because I made some big changes 21 days ago and would like @alevy to review them first, if possible

alevy:
@hudson-ayers since I'm technically the author of the PR I can't approve the changes in GitHub's UI, so: I hereby approve. Thanks for taking this over the finish line.

PR #2140, opened 2020-10-08
OpenTitan: Initial support for the Flash Controller

alistair23:
### Pull Request Overview

This adds initial support for the OpenTitan Flash Controller, as documented here: https://docs.opentitan.org/hw/ip/flash_ctrl/doc/index.html

### Testing Strategy

Running the `examples/tests/nonvolatile_storage` libtock-c app, all tests pass.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/chips/lowrisc/src/flash_ctrl.rs
+++ b/chips/lowrisc/src/flash_ctrl.rs
@@ -411,17 +411,11 @@ impl hil::flash::Flash for FlashCtrl<'_> {
         self.read_index.set(0);
 
         // Start the transaction
-        let info_sel = if self.region_num as u32 > 3 {
-            CONTROL::INFO_SEL::SET
-        } else {
-            CONTROL::INFO_SEL::CLEAR
-        };
         self.registers.control.write(
             CONTROL::OP::READ
-                + CONTROL::PARTITION_SEL::INFO
+                + CONTROL::PARTITION_SEL::DATA
                 + CONTROL::NUM.val(((PAGE_SIZE / 4) - 1) as u32)
-                + CONTROL::START::SET
-                + info_sel,
+                + CONTROL::START::SET,
         );
 
         Ok(())
@@ -454,17 +448,11 @@ impl hil::flash::Flash for FlashCtrl<'_> {
         self.write_index.set(0);
 
         // Start the transaction
-        let info_sel = if self.region_num as u32 > 3 {
-            CONTROL::INFO_SEL::SET
-        } else {
-            CONTROL::INFO_SEL::CLEAR
-        };
         self.registers.control.write(
             CONTROL::OP::PROG
-                + CONTROL::PARTITION_SEL::INFO
+                + CONTROL::PARTITION_SEL::DATA
                 + CONTROL::NUM.val(((PAGE_SIZE / 4) - 1) as u32)
-                + CONTROL::START::SET
-                + info_sel,
+                + CONTROL::START::SET,
         );
 
         // Write the data until we are full or have written all the data
@@ -514,17 +502,11 @@ impl hil::flash::Flash for FlashCtrl<'_> {
         self.registers.addr.write(ADDR::START.val(addr as u32));
 
         // Start the transaction
-        let info_sel = if self.region_num as u32 > 3 {
-            CONTROL::INFO_SEL::SET
-        } else {
-            CONTROL::INFO_SEL::CLEAR
-        };
         self.registers.control.write(
             CONTROL::OP::ERASE
                 + CONTROL::ERASE_SEL::PAGE
-                + CONTROL::PARTITION_SEL::INFO
-                + CONTROL::START::SET
-                + info_sel,
+                + CONTROL::PARTITION_SEL::DATA
+                + CONTROL::START::SET,
         );
 
         // Disable erase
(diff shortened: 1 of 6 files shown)


chips/lowrisc/src/flash_ctrl.rs
```diff
@@ -0,0 +1,553 @@
...
+    }
+
+    fn read(&self, _buffer: &'a mut [u8], _address: usize, _length: usize) -> ReturnCode {
+        unimplemented!()
+    }
+
+    fn write(&self, _buffer: &'a mut [u8], _address: usize, _length: usize) -> ReturnCode {
+        unimplemented!()
+    }
+}
```
hudson-ayers, line 553: I don't think it makes sense to provide this basically empty implementation of the `NonvolatileStorage` trait. Lots of chips (every chip?), implement only flash, and then use the nonvolatile storage capsule to provide the `NonvolatileStorage` interface, which looks like what you want to do here as well.

chips/lowrisc/src/flash_ctrl.rs
```diff
@@ -0,0 +1,553 @@
...
+
+        // Start the transaction
+        let info_sel = if self.region_num as u32 > 3 {
+            CONTROL::INFO_SEL::SET
+        } else {
+            CONTROL::INFO_SEL::CLEAR
+        };
+        self.registers.control.write(
+            CONTROL::OP::READ
+                + CONTROL::PARTITION_SEL::INFO
```
hudson-ayers, line 423: If I understand the docs correctly, this means that all flash reads and writes through this interface are exclusively being made to the info partition. However, the docs also state "The data partition holds generic data like a generic memory would. The information partition holds metadata about the data partition as well design specific secret data", which to me suggests that for the generic flash interface exposed by Tock we should be conducting all reads and writes in the data partition, rather than an info partition.

Am I missing something, or should this change here and in `write_page()` / `erase_page()` to select the data partition?

chips/lowrisc/src/flash_ctrl.rs
alistair23: Removed.

chips/lowrisc/src/flash_ctrl.rs
alistair23: The docs says that

> data partition - this is the portion of flash that is accessible both by the host and by the controller.

So I don't think data partition is the right place to store sensitive data as then the host can read it.

hudson-ayers: What is the host in this case? Does that just mean that the CPU can directly perform memory mapped reads from flash?

I thought that is how the flash works on most chips in Tock -- the nonvolatile storage interface is not assumed to be used for secrets, but just exposes a way to read/write to allowed locations in flash memory. Generally, we rely on `#[forbid_unsafe]` to prevent capsules from accessing arbitrary flash memory for reads using memory accesses.

But I am not the most familiar with how the nonvolatile storage interface is intended to work, so I will defer to others on this.

jrvanwhy: In the case of EarlGrey, I believe the host is Ibex.

alistair23: Ah, I was under the impression that host meant the CPU OpenTitan was connected to.

Ok, so it sounds like the data partition would be a better place to use for Tock's flash.

bradjc (approved):

hudson-ayers (changes requested):
Putting a request changes because I don't think this should be merged until it is changed to use the data partition.

alistair23:
I changed it to use the data partition

chips/lowrisc/src/flash_ctrl.rs
```diff
@@ -0,0 +1,537 @@
...
+
+        // Reset the write index
+        self.write_index.set(0);
+
+        // Start the transaction
+        let info_sel = if self.region_num as u32 > 3 {
+            CONTROL::INFO_SEL::SET
+        } else {
+            CONTROL::INFO_SEL::CLEAR
+        };
```
hudson-ayers, line 461: I believe this is no longer needed since we are now writing to the data partition?

chips/lowrisc/src/flash_ctrl.rs
```diff
@@ -0,0 +1,537 @@
...
+
+        // Set the address
+        self.registers.addr.write(ADDR::START.val(addr as u32));
+
+        // Start the transaction
+        let info_sel = if self.region_num as u32 > 3 {
+            CONTROL::INFO_SEL::SET
+        } else {
+            CONTROL::INFO_SEL::CLEAR
+        };
```
hudson-ayers, line 521: same here

chips/lowrisc/src/flash_ctrl.rs
```diff
@@ -0,0 +1,537 @@
...
+        // Save the buffer
+        self.read_buf.replace(buf);
+        self.read_index.set(0);
+
+        // Start the transaction
+        let info_sel = if self.region_num as u32 > 3 {
+            CONTROL::INFO_SEL::SET
+        } else {
+            CONTROL::INFO_SEL::CLEAR
+        };
```
hudson-ayers, line 418: same here

hudson-ayers (approved):

PR #2141, opened 2020-10-08
Makefile rule updated

skyplane23:
### Pull Request Overview

This pull request changes the makefile rule for flashing the kernel on the ```nano33ble``` to be in line with the other makefile semantics boards' makefiles.

### Testing Strategy

This pull request was tested by running the ```make flash``` command.

### TODO or Help Wanted

N/A

### Documentation Updated

-  [X] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/nano33ble/Makefile
+++ b/boards/nano33ble/Makefile
@@ -11,6 +11,6 @@ ifdef PORT
 endif
 
 # Upload the kernel using bossac
-.PHONY: program
-program: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
+.PHONY: flash
+flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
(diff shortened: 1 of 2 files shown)


bradjc:
In other boards `make program` uses a bootloader, and since bossac is also communicating with a bootloader I think `make program` is appropriate. `make flash` is used when using a JTAG-esque programmer.

skyplane23:
In other makefiles ```make program``` is used to flash the apps and here only flash the kernel. Shouldn't be at least ```make program-kernel``` to avoid any confusion?

bradjc:
The general format is:
- `make program` uses a bootloader
- `make flash` uses JTAG (or similar)
And both conventionally only flash the kernel.

We have a few other make targets in various boards, and certain boards handle things differently, but that is the standard format.

skyplane23:
Ok, that makes sense. Thank you!

PR #2145, opened 2020-10-09
capsules: fxo: need to enable both mag and accel

bradjc:
Enabling just the magnetometer causes the accelerometer to stop working.
We never tested the magnetometer at the same time before and never
noticed this.

### Testing Strategy

This pull request was tested by running sensors app on hail.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers (approved):
bors r+

PR #2146, opened 2020-10-09
kernel: process: reconfigure mpu on restart

bradjc:
If a process restarts we need to setup the MPU based on the original app break and not the one the app had when it crashed.

If an app shrank its memory before crashing then not doing this causes an immediate MPU fault if the kernel puts the stack outside of the app's old memory region.

If an app grew its memory before crashing then not doing this causes a security issue since an app can grow its heap, crash, then restart and have access to memory above its app break address.

### Testing Strategy

This pull request was tested by using the stack test 01 app on hail with the bug in libtock-c that causes the app to crash.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -2075,6 +2075,28 @@ impl<C: 'static + Chip> Process<'_, C> {
         self.allow_high_water_mark
             .set(self.original_allow_high_water_mark);
 
+        // Reset the MPU configuration based on the original app_break, which is
+        // likely different than what the app break was when the app crashed.
+        let mpu_config_res = self.mpu_config.map_or(Err(()), |mut config| {
+            self.chip.mpu().update_app_memory_region(
+                self.app_break.get(),
+                self.kernel_memory_break.get(),
+                mpu::Permissions::ReadWriteOnly,
+                &mut config,
+            )?;
+            Ok(())
+        });
+        match mpu_config_res {
+            Ok(_) => {}
+            Err(_) => {
+                // We couldn't configure the MPU for the process. This shouldn't
+                // happen since we were able to start the process before, but at
+                // this point it is better to leave the app faulted and not
+                // schedule it.
+                return;
+            }
+        }
+
         // Handle any architecture-specific requirements for a process when it
         // first starts (as it would when it is new).
         let new_stack_pointer_res = self.stored_state.map_or(Err(()), |stored_state| unsafe {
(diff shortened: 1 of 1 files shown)


kernel/src/process.rs
```diff
@@ -2075,6 +2075,28 @@ impl<C: 'static + Chip> Process<'_, C> {
         self.allow_high_water_mark
             .set(self.original_allow_high_water_mark);
 
+        // Reset the MPU configuration based on the original app_break, which is
+        // likely different than what the app break was when the app crashed.
+        let mpu_config_res = self.mpu_config.map_or(Err(()), |mut config| {
+            self.chip.mpu().update_app_memory_region(
+                self.app_break.get(),
+                self.kernel_memory_break.get(),
```
hudson-ayers, line 2083:
```suggestion
                self.original_app_break,
                self.original_kernel_memory_break,
```
(I realize this is basically equivalent, because `app_break` is set to `original_app_break` above, but this way removes an extra `get()` call and protects this mistake from being recreated if the code in this function is shuffled such that the MPU reconfig happens before changing the non-original values)

kernel/src/process.rs
```diff
@@ -2075,6 +2075,28 @@ impl<C: 'static + Chip> Process<'_, C> {
...
+                mpu::Permissions::ReadWriteOnly,
+                &mut config,
```
hudson-ayers, line 2085: This appears to still use the MPU config of the failed process, which is not guaranteed to be the same as the MPU config of a newly created process, as I understand it (`allocate_region()`, for example, modifies the config).

Should we be replacing the config with the `Default` before this?

hudson-ayers:
I pushed a commit that I believe addresses my concerns. I believe that there are better ways to do this, but the better ways involve changing the MPU trait which goes beyond the purview of a pre-release bug fix.

Now additional allocated MPU regions for a process (such as those allocated by IPC) will not carry over past a restart of that process.

hudson-ayers:
The QEMU "push" check only fails because this PR has not been rebased to use the new QEMU download location, but bors will not block on that check.

hudson-ayers (approved):
Giving my approval to brads commit, but someone should review my added commit before we merge this.

bradjc:
Looks good to me.

After several discussions it does seem that recomputing is the approach to take.

PR #2147, opened 2020-10-10
capsules: si7021: only use state for state var

bradjc:
This PR fixes a Rule 2 (Never use more than one state variable) violation.

The si7021 capsule cannot use the presence of a buffer as a signal that the driver is free since the buffer is free when the sensor is waiting for the measurement to finish.

### Testing Strategy

This pull request was tested by running sensors and hail on hail.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/si7021.rs
+++ b/capsules/src/si7021.rs
@@ -239,16 +239,12 @@ impl<'a, A: time::Alarm<'a>> i2c::I2CClient for SI7021<'a, A> {
 
 impl<'a, A: time::Alarm<'a>> kernel::hil::sensors::TemperatureDriver<'a> for SI7021<'a, A> {
     fn read_temperature(&self) -> kernel::ReturnCode {
-        self.buffer.take().map_or_else(
-            || {
-                if self.on_deck.get() != OnDeck::Nothing {
-                    ReturnCode::EBUSY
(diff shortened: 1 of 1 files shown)


ppannuto (approved):

capsules/src/si7021.rs
```diff
@@ -239,25 +239,29 @@ impl<'a, A: time::Alarm<'a>> i2c::I2CClient for SI7021<'a, A> {
...
-                    ReturnCode::SUCCESS
-                }
-            },
-            |buffer| {
+        // This chip handles both humidity and temperature measurements. We can
+        // only start a new measurement if the chip is idle. If it isn't then we
+        // can put this request "on deck" and it will happen after the
+        // temperature measurement has finished.
+        if self.state.get() == State::Idle {
+            self.buffer.take().map_or(ReturnCode::EBUSY, |buffer| {
```
hudson-ayers, line 247: Not a big deal, but maybe this should be `ReturnCode::FAIL`, as its a logic error for the buffer to be missing rather than something that could happen with a certain sequence of calls?

hudson-ayers (approved):

PR #2154, opened 2020-10-12
tools/board-runner: Add automated OpenTitan and Apollo3 tests

alistair23:
### Pull Request Overview

This PR adds automated tests to run the libtock-c release tests on OpenTitan and the Apollo3.

There is a fair bit of duplicated code here, so more clean-up can be done. For now though this seems like a good starting point to continue improving the Tock testing infrastructure.

### Testing Strategy

Running the release tests on OpenTitan and the Apollo3.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,10 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
+      - name: Install dependencies
+        continue-on-error: true
+        run: |
+          sudo apt install libudev-dev
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
       - name: ci-job-libraries
--- a/tools/board-runner/Cargo.toml
+++ b/tools/board-runner/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "board-runner"
+version = "0.1.0"
+authors = ["Tock Project Developers <tock-dev@googlegroups.com>"]
+edition = "2018"
+
+[dependencies]
+rexpect = "0.4"
+serialport = "3.3.0"
--- a/tools/board-runner/src/opentitan.rs
+++ b/tools/board-runner/src/opentitan.rs
@@ -1,0 +1,423 @@
+use rexpect::errors::Error;
+use rexpect::spawn_stream;
+use serialport::prelude::*;
+use serialport::SerialPortSettings;
+use std::env;
+use std::fs::OpenOptions;
+use std::process::{Command, Stdio};
+use std::time::Duration;
+use std::{thread, time};
+
+fn opentitan_flash(
+    app_name: &str,
+) -> Result<rexpect::session::StreamSession<std::boxed::Box<dyn serialport::SerialPort>>, Error> {
+    let s = SerialPortSettings {
+        baud_rate: 115200,
+        data_bits: DataBits::Eight,
+        flow_control: FlowControl::None,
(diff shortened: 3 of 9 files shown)


tools/board-runner/src/opentitan.rs
```diff
@@ -0,0 +1,423 @@
...
+    //    APP_HEAP_SIZE    = 4096
+    //    KERNEL_HEAP_SIZE = 2048
+    // opentitan_stack_size_test1().unwrap_or_else(|e| panic!("opentitan job failed with {}", e));
+
+    // Disabled by default.
+    // Requires:
+    //    STACK_SIZE       = 2048
+    //    APP_HEAP_SIZE    = 4096
+    //    KERNEL_HEAP_SIZE = 2048
+    // opentitan_stack_size_test2().unwrap_or_else(|e| panic!("opentitan job failed with {}", e));
```
hudson-ayers, line 409: Why do these require different app/kernel heap sizes on OpenTitan, but not on other platforms?

alistair23: I'm not sure. Maybe it's related to app addresses not exactly matching what the kernel expects?

I did try using a few different addresses, but that didn't seem to help.

tools/board-runner/src/opentitan.rs
```diff
@@ -0,0 +1,435 @@
...
+    p.exp_string("panicked at 'Process mpu_walk_region had a fault'")?;
+
+    Ok(())
+}
+
+fn opentitan_multi_alarm_test() -> Result<(), Error> {
+    let app = format!(
+        "{}/{}",
+        env::var("LIBTOCK_C_TREE").unwrap(),
+        "examples/tests/multi_alarm_test/build/rv32imc/rv32imc.0x20030080.0x10005000.tbf"
```
hudson-ayers, line 363: it might be good if this script started with a `make clean` of `libtock-c`, and then just matches any `rv32imc.*.tbf` file in each app directory, just so that this script doesn't break next time we need to update the app addresses for opentitan

hudson-ayers (commented):
I think this is really cool.

tools/board-runner/src/opentitan.rs
alistair23: I can't just swap this to `rv32imc.*.tbf` as it then fails to find the elfs.

I'm also not sure that it's a good idea, it's then possible that we will run old elfs. Having to change this when addresses in libtock-c seems somewhat reasonable at least for now based on how much manual intervention is required to run this currently.

tools/board-runner/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "qemu-runner"
```
bradjc, line 2:
```suggestion
name = "board-runner"
```

bradjc (dismissed):

bradjc (approved):

.github/workflows/ci.yml
```diff
@@ -67,6 +67,10 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
+      - name: Install dependencies
```
hudson-ayers, line 70: Why add this to the CI if the CI does not actually call this new runner?

alistair23: It still ends up being built, so it was failing building the serial crate I was using.

hudson-ayers (approved):

Issue #1692, opened 2020-03-16
Persistent Key-Value Store


gkelly:
One area that was identified by the Tock on OpenTitan WG is the lack of a persistent key value store (or filesystem) in Tock.  The current model of having apps modify their own flash to store values is very constrained and has a few issues:
- If the kernel and apps are signed as part of a monolithic signature then modifying the app's flash region breaks signing.
- Any update system breaks the storage.  A/B updates with a monolith mean it's a whole new app getting run on the other slot.  In place updates probably overwrite storage unless careful consideration is paid to the section.  Side-by-side updated apps come up as whole new apps.
- Flash topology tolerance.  Most (all?) systems that Tock targets are XIP from flash. During flash erases a bank is usually unavailable for reads/XIP.  A kv-store/filesystem that is aware of flash topology could avoid writing/erasing banks that are currently in use, or even just defer flash layout questions to the platform itself.

To that end, I believe it makes sense to consider a persistent key-value store API for Tock.  I think reasonable goals are to allow the platform to provide an implementation of the lower-level details (bank management, flash controller details) but provide common things like access control (isolating apps to only reading things that they wrote, size limits, for example).

To get the conversation started I'd like to propose an API:
- Read(&str) -> &[u8]
- Write(&str, &[u8])
- Delete(&str)

The keys would be string-named, the data stored is [u8].  The semantics are that no function will return until the corresponding action is visible to subsequent calls referring to that key.  To begin with, I think apps should be isolated from interacting with each other via this API.

----------

- [x] Add a Key-Value library #2156 
- [x] Add kernel interfaces for key-value stores
  - [x] lower interface (exposes low level detail)
  - [x] upper interface (easier to use API)
- [x] Userspace Driver interface for persistent key-value stores
- [x] Permissions for process key-value store access

bradjc:
This is only tangentially related, but Tock's bootloader supports key-value "attributes" that we use to automatically identify what kind of board is connected (https://github.com/tock/tock/blob/master/doc/Memory_Layout.md#flash-1).

Also, a while ago we discussed "kernel attributes" that would be persistent key,value settings or state relevant to the kernel (like the address of apps), rather than the physical board (https://github.com/tock/tock/issues/695). We ended up deciding it wasn't worth the effort to implement. But, if boards have a key-value mechanism already, might be worth revisiting.

labbott:
[...]

Issue #1459, opened 2019-11-12
RFC: Flash HIL improvements


ia0:
## Motivation

Flash technology does not permit the usage of arbitrary data-structures due to restrictions like:
- It is not possible to switch a single bit back 0 to 1 without erasing all 0s to 1s in the page containing that bit. But it is possible to switch a single bit from 1 to 0.
- A page can only be erased a given amount of times before exhibiting undefined behavior. In other words, the flash has a short lifespan.

To implement flash-friendly data-structures, it is useful to take benefit of all features the flash provides like writing multiple times the same unit of flash (smaller than a full page) before having to erase it.

## Description

The current flash HIL interface is:
- The type of a page (must implement `AsMut<[u8]>`).
- Read a page asynchronously.
- Write a page asynchronously (page to read from must be mutable).
- Erase a page asynchronously.

This interface has the following issues:

### Issue 1: Writing a page takes a mutable page to read from.

This is not clear why the write function has the following signature:

```rust
fn write_page(&self, page_number: usize, buf: &'static mut Self::Page) -> ReturnCode;
```

It would seem more natural to take a shared slice to read from:

```rust
fn write_page(&self, page_number: usize, buf: &'static Self::Page) -> ReturnCode;
```

### Issue 2: The read and write buffer lifetimes are `'static`.

This probably requires the user of the HIL (e.g. a capsule) to have a static buffer allocated and probably copy data back and forth between this backend interface and its own frontend interface (like `AppSlice` in case of a capsule). This issue is due to the fact that the API is non-blocking and that the HIL implementation probably cannot allocate or free a buffer.

### Issue 3: It is not possible to read a smaller slice than a page.

This is not clear why such restriction is in place.

Note however that exposing a read function is probably not useful in the first place if it is possible to read the flash directly, which is the case for nrf52840dk flash for example. So it is as well not useful to expose such function in a capsule since the application can already read its own flash.

See the proposal for the same issue for write operations if read operations should be provided.

### Issue 4: It is not possible to write a smaller slice than a page.

This is not clear why such restriction is in place. If flashs permit to write at a smaller granularity, this option should be exposed.

#### Proposal 1: Expose the smallest write granularity.

```rust
fn write_size(&self) -> usize;
/// Both `ptr` and `buf.len()` must be `write_size` aligned (otherwise `EINVAL`).
fn write_slice(&self, ptr: usize, buf: &[u8]) -> ReturnCode;
```

[...]

PR #2156, opened 2020-10-14
TicKV: Implement a persistent key/value storage

alistair23:
### Pull Request Overview

TicKV (Tiny Circular Key Value) is a small file system allowing key value pairs to be stored in Flash Memory.

Tock currently has a flash HIL, but there is no implementation on top of that HIL to reliably and persistently store data over a long period of time for multiple apps.

This is the first step in that direction by implementing a key/value filesystem designed to store key/value pairs in flash. This also addresses: https://github.com/tock/tock/issues/1692

The design has been implemented as a library, which hopefully will allow others to use it if they want to.

For details of the design read the PR changes, I'm not going to repeat the design and goals here.

#### Why write a new filesystem?

 * I couldn't find one already implemented in Rust
 * I couldn't find one already implemented that would meet Tock's strict requirements
 * I couldn't find one small and simple enough to just do what we need

#### Next steps

Step 1 is to get this merged into Tock.

Then I will have to connect it up with the OpenTitan flash controller and expose it to Tock.

At the same time I want to work on reliable app IDs that can be used for access control.

Finally I need to write a capsule that allows apps to read/write data to/from flash in a secure way in which we can enforce access controls.

### Testing Strategy

Unit tests.

### TODO or Help Wanted

Feedback mostly.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/libraries/tickfs/README.md
+++ b/libraries/tickfs/README.md
@@ -1,0 +1,110 @@
+# TickFS
+
+TickFS (Tiny Circular Key Value File System) is a small file system allowing
+key value pairs to be stored in Flash Memory.
+
+TickFS was written to allow the Tock OS kernel to persistently store app data
+on flash. It was written to be generic though, so other Rust applications can
+use it if they want.
+
+TickFS is based on similar concepts as
+[Yaffs1](https://yaffs.net/documents/how-yaffs-works]).
+
+## Goals of TickFS
+
+TickFS is designed with these main goals (in order)
+
+ * Fully implemented in no_std Rust
+ * Power loss resilient
+ * Maintain data integrity and detect media errors
+ * Wear leveling
+ * Low memory usage
+ * Low storage overhead
+ * No external crates in use (not including unit tests)
+
+TickFS is also designed with some assumptions
+
+ * Most operations will be retrieving keys
+ * Some operations will be storing keys
+ * Keys will rarely be deleted
+ * Key values will rarely need to be modified
+
+## ACID characteristics
+
+TickFS provides some ACID properties. For the purpose of ACID a transaction is
+a key operation, that is finding, adding, invalidating or fully removing
+(garbage collection) a key.
+
+Atomicity: TickFS does not provide atomicity. If a operation fails mid write
+then there will be an invalid key/value stored to flash.
+
+Consistency: Consistency is maintained similar to atomicity. All operations
+can only take the database from a valid state to another valid state.
+
+Isolation: TickFS only allows a single operation at a time. In this way it
+provides isolation. The layer above TickFS is responsible for handling
+concurrent accesses by deferring operations for example.
+
+Durability: TickFS ensures durability and once a transaction has completed
+and been committed to flash it will remain there.
+
+## Using TickFS
+
+See the generated Rust documentation for details on using this in your project.
+
+## How TickFS works
+
+Unlike a regular File System (FS) TickFS is only designed to store Key/Value (KV)
+pairs in flash. It does not support writing actual files, directories or other
+complex objects. Although a traditional file system layer could be added on top
+to add such features.
+
+TickFS allows writing new key/value pairs (by appending them) and removing
+old key/value pairs.
+
+TickFS has two important types, regions and objects.
+
+A TickFS region is the smallest region of the flash memory that can be erased
+in a single command.
+
+TickFS saves and restores objects from flash. TickFS objects contain the value
+the user wanted to store as well as extra header data. Objects are internal to
+TickFS and users don't need to understand them in detail to use it.
+
+For more details on the technical implementation see the [SPEC.md](./spec.md) file.
+
+### Collisions
+
+TickFS will prevent a new key/value pair with a colliding hash of the key to be
+added. The collision will be reported to the user with
+`ErrorCode::KeyAlreadyExists`.
+
+### Power loss protection
+
+TickFS ensures that in the event of a power loss, all commited data remains
+commited. This is the durability guarantee as part of the ACID semantics.
+The only data that can be lost in the event of a power loss is
+the data which hasn't been write to flash yet.
+
+If a power loss occurs after calling `append_key()` or `invalidate_key()`
+before it has completed then the operation probably did not complete and
+that data is lost.
+
+### Security
+
+TickFS uses checksums to check data integrity. TickFS does not have any measures
+to prevent malicious manipulation or privacy. An attacker with access to the
+flash can change the values without being detected. An attacked with access
+to flash can also read all of the information. Any privacy, security or
+authentication measures need to be layered on top of TickFS.
+
+## Versions
+
+TickFS stores the version when adding objects to the flash storage.
+
+TickFS is currently version 0.
+
+ * Version 0
+   * Version 0 is a draft version. It should NOT be used for important data!
+     Version 0 maintains no backwards compatible support and could change at
+     any time.
--- a/libraries/tickfs/SPEC.md
+++ b/libraries/tickfs/SPEC.md
@@ -1,0 +1,495 @@
+# TickFS Technical Details
+
+## How TickFS works
+
+Unlike a regular File System (FS) TickFS is only designed to store Key/Value (KV)
+pairs in flash. It does not support writing actual files, directories or other
+complex objects. Although a traditional file system layer could be added on top
+to add such features.
+
+TickFS allows writing new key/value pairs (by appending them) and removing
+old key/value pairs.
+
+Similar to (Yaffs1)[https://yaffs.net/documents/how-yaffs-works] TickFS uses a
+log structure and circles over the flash data. This means that the file
+system is inherently wear leveling as we don't regularly write to the same
+flash region.
+
+## Storage Format
+
+TickFS flash storage is formatted as a sorting of regions with each region
+containing TickFS objects.
+
+### TickFS Regions
+
+A TickFS region is the smallest set of contiguous flash storage space that can
+be erased with a single command.
+
+For example if a flash controller allows erasing lengths of 1024 (0x400) bytes
+then the length of a region will be 1024 (0x400) bytes.
+
+The number of regions is determined by the capacity of the flash storage and
+the size of regions.
+
+The start and end address of flash used for TickFS must be region aligned.
+
+### TickFS Objects
+
+A TickFS object is the representation of a key/value pair in flash. An object
+contains the value to be saved as well as useful header data.
+
+TickFS saves and reads objects from flash. TickFS objects contain the value
+the user wanted to store as well as extra header data.
+
+A TickFS object consists of multiple parts:
+
+```
+|||||||||||||||||
+|               |
+| Object Header |
+|               |
+|||||||||||||||||
+|               |
+|     Value     |
+|               |
+|||||||||||||||||
+|               |
+|   Check Sum   |
+|               |
+|||||||||||||||||
+```
+
+#### Object Header
+
+Currently ObjectHeader includes these fields:
+
+```Rust
+struct ObjectHeader {
+    version: u8,
+    flags: u4,
+    len: u12,
+    hashed_key: u64,
+}
+```
+
+The `version` field is a byte containing the version of TickFS used when the
+object was written.
+This allows us to upgrade this library in the future, while still supporting
+old data formats.
+
+The `flags` field is a bitmap of at most 4 flags that can be OR-ed together to
+describe an object state or features. The only flag defined is the `valid` flag
+(bit 3), indicating that an object is valid.
+
+It looks like this in flash:
+
+```
+|valid|Reserved|Reserved|Reserved|
+|     |        |        |        |
+|  1  |    0   |    0   |    0   |
+```
+
+Where `valid` indicates if an object is valid. A `1` indicates it is a valid
+object, a `0` indicates that it has been marked as invalid (see below).
+
+The `len` field is 12-bits long.
+This field indicates the total length of the object, including the
+header and check sum. The maximum length of the entire object is
+4KiB (0xFFF) or the region size, whichever is smaller.
+
+The `hashed_key` field stores the 64-bit (8 byte) output of the key hash.
+
+ObjectHeader is internal to TickFS and users of TickFS do not need to
+understand it.
+
+#### Object Value
+
+The Value component of the TickFS object is the value that the user wants to
+store.
+
+The values can be any length as long as they follow both:
+ * Don't span multiple regions. That limits the maximum value length to
+   `region_size - size_of::<ObjectHeader>()`
+ * Don't have a maximum length greater then 4KiB (0xFFF).
+
+#### Checksum
+
+The checksum is a hash of the entire object (not including the checksum).
+The checksum is calculated using the same hash algorithm used for the keys.
+
+A simple CRC isn't used to avoid depending on external crates and to
+re-use the already implemented Hash. This will potentially allow
+hardware accelerated implementation of `core::hash::Hasher` to be used
+for the checksum.
+
+### Object overhead
+
+Currently the overhead of an TickFS object is 21 bytes. Most of this is the 8
+bytes for the key hash and 8 bytes for a checksum.
+
+### Location of objects
+
+The region where a TickFS object is stored is dependent on the output of the
+key hash and the number of regions.
+
+TickFS determines the region an object will be stored or retrieved from using
+region numbers, starting from 0. The region number of an object is equal to the
+last two bytes of an object hashed-key modulo the number of regions.
+
+This will produce a number that is between zero and the total number of
+TickFS regions. This number is the regions number where the data is stored
+or loaded from.
+
+This allows us to quickly determine which flash region a key is stored in. This
+should improve lookup time by reducing the number of reads required.
+
+If a give region is full, we will start to look in neighboring regions. We first
+perform the same search in the next (increment region number) region. If this
+region doesn't exist (we are at the end of the regions) or if the region is full
+we then search in the previous region (decrement region number).
+
+When storing a object this process continues until we either:
+ * Search all regions
+ * Find a free space
+
+When retrieving an object the process continues until we either:
+ * Search all regions
+ * Find the key we are looking for
+ * Find a region that is empty
+
+### Invalidating keys
+
+Flash has the characteristic that although read/writes can happen at small
+granularities an erase operation requires a entire block (specified
+by the region size).
+
+Due to this removing a key with the `invalidate_key()` function does NOT
+remove anything from flash. Instead calling `invalidate_key()`
+will mark the `valid` boolean for that object as `false` (0).
+
+This is done because changing a `1` to a `0` in flash can be done with a
+write to a single byte (where only 1 bit changes). While changing a `0`
+to a `1` requires an erase of the entire region.
+
+The key point here is that removing a key does NOT remove it or the object
+from the flash storage. If the key contains sensitive information it will
+still be in flash. Once invalidated though `get_key()` will not return the key
+any more.
+
+If all the objects in a region are no longer valid then that region will be
+erased when `garbage_collect()` is called. Note that even if the flash is
+full `garbage_collect()` will not be called automatically.
+
+### Initialisation
+
+When setting up a block of flash for the first time the entire size of flash
+is erased. Then a super key called "tickfs-super-key" is added with no
+attached data.
+
+On future initialisation the implementation will check for the
+"tickfs-super-key" key. If it exists no erase operations will occur. If it
+doesn't exist the entire block of flash will be erased.
+
+## What is looks like in flash
+
+### Adding a key
+
+This is an example of what `TickFS::new(..., 0xC00, 0x400)` will
+look like in flash.
+
+```
+0x000                  0x400                  0x800                 0xC00
+--------------------------------------------------------------------------
+|||||     Region 0     |||||     Region 1     |||||     Region 2     |||||
+|||||                  |||||                  |||||                  |||||
+|||||                  |||||                  |||||                  |||||
+--------------------------------------------------------------------------
+```
+
+All values in flash will be `0xFF` as the flash has just been erased.
+
+When a key is added with a `[u8; 32]` `value` with `append_key()` it will
+look like this:
+
+```
+Key:    "ONE"
+Hash:   0xeda10078886193bb
+Region: 1 (0x93bb % 3)
+```
+
+`Region` is the flash region where the object will be stored. In this case it's
+region 1 of 3.
+
+```
+0x000                  0x400                  0x800                 0xC00
+--------------------------------------------------------------------------
+|||||     Region 0     |||||     Region 1     |||||     Region 2     |||||
+|||||                  |||||                  |||||                  |||||
+|||||                  |||||ONE               |||||                  |||||
+--------------------------------------------------------------------------
+```
+
+Where the TickFS object ONE will look like this
+
+```
+0x400                                                                                              0x42C
+--------------------------------------------------------------------------------------------------------
+||||| version|len/flag|   len  |                              hashed_key                               |
+|||||        |        |        |        |        |        |        |        |        |        |        |
+|||||    0x00|10000000|    0x34|    0xed|    0xa1|    0x00|    0x78|    0x88|    0x61|    0x93|    0xbb|
+-------------------------------------------------------------------------------------------------------|
+```
+
+```
+0x42C                                                                                              0x52C
+--------------------------------------------------------------------------------------------------------
+|||||                                               value                                              |
+|||||                                                                                                  |
+|||||xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx|
+-------------------------------------------------------------------------------------------------------|
+```
+
+```
+0x52C                                                                   0x54C
+-----------------------------------------------------------------------------
+|||||                                checksum                               |
+|||||        |        |        |        |        |        |        |        |
+|||||    0xe5|    0x5d|    0x99|    0x22|    0x0a|    0x3c|    0xe1|    0x17|
+----------------------------------------------------------------------------|
+```
+
+In this case the header information takes up 19 bytes of a total of 51 bytes,
+which is around 37% of the space.
+
+Note that if region 1 is full, we would then try to save the ONE object in
+region 2, then try again in region 0.
+
+### Adding a second key
+
+When a new key TWO with a `[u8; 32]` `value` is added with `append_key()` the
+flash might look like this:
+
+```
+0x000                  0x400                  0x800                 0xC00
+--------------------------------------------------------------------------
+|||||     Region 0     |||||     Region 1     |||||     Region 2     |||||
+|||||                  |||||                  |||||                  |||||
+|||||                  |||||ONE|TWO           |||||                  |||||
+--------------------------------------------------------------------------
+```
+
+Where TWO will have the same structure as ONE, except with a different hash
+value, different checksum and starts at address 0x54C. Note that depending
+on the hash of TWO it could be added to any region.
+
+### Adding a third key
+
+When a new key THRID with a `[u8; 32]` `value` is added with `append_key()` the
+flash might look like this
+
+```
+0x000                  0x400                  0x800                 0xC00
+--------------------------------------------------------------------------
+|||||     Region 0     |||||     Region 1     |||||     Region 2     |||||
+|||||                  |||||                  |||||                  |||||
+|||||THRID             |||||ONE|TWO           |||||                  |||||
+--------------------------------------------------------------------------
+```
+
+Note that although region 1 isn't full, it is placed in region 0 based on the
+region calculated from the lower byte of it's hash.
+
+Objects will be placed in chronological order inside a region. Objects will
+not be placed in order inside flash though. Instead they are placed at a
+region depending on their hash. This is described in more detail above.
+
+### Finding keys
+
+Locating keys in flash follows a similar process to writing them.
+
+First we calculate the hash and determine the region where we expect the
+object to be stored. With the same key input and hash function we will always
+get the same hash and region.
+
+Using the same example as above, we will get this:
+
+```
+Key:    "ONE"
+Hash:   0xeda10078886193bb
+Region: 1 (0x93bb % 3)
+```
+
+Then we load the entire region from flash. In this example that would be
+region 1. So we read the entire region 1 from flash.
+
+We then iterate over the loaded region, starting with the first byte.
+
+We check to make sure the version is supported and that the object isn't
+marked as !`valid`.
+
+We then check to see if the object hash matches the hash we are
+looking for. If it doesn't we move forward in the loaded region by the
+total length of the object we just checked and start the process again.
+
+We continue this loop until we either find the key we are looking for or
+find a version 0xFF, indicating the end of the blocks in that region.
+
+This method allows a quick retrieval of data from a given key. We only
+require a single read and store the entire region in memory. This should
+only have a small memory foot print as regions are generally less then 4KiB.
+
+### Adding keys to full regions
+
+In the previous example we added the ONE object to an empty flash region. This
+example shows what would happen if region 1 was full.
+
+Image if the flash storage looked like this:
+
+```
+0x000                  0x400                  0x800                 0xC00
+--------------------------------------------------------------------------
+|||||     Region 0     |||||     Region 1     |||||     Region 2     |||||
+|||||                  |||||                  |||||                  |||||
+|||||                  |||||KEY|EXAMPLE|TEST| |||||                  |||||
+--------------------------------------------------------------------------
+```
+
+and we wanted to add the same ONE key:
+
+```
+Key:    "ONE"
+Hash:   0xeda10078886193bb
+Region: 1 (0x93bb % 3)
+```
+
+We can't add the key to region 1 as it is full. Instead we fall back to trying
+region 2. If region 2 was full we would then try region 0 and so on.
+
+Adding key ONE would look like this:
+
+
+```
+0x000                  0x400                  0x800                 0xC00
+--------------------------------------------------------------------------
+|||||     Region 0     |||||     Region 1     |||||     Region 2     |||||
+|||||                  |||||                  |||||                  |||||
+|||||                  |||||KEY|EXAMPLE|TEST| |||||ONE               |||||
+--------------------------------------------------------------------------
+```
+
+### Finding keys from full regions
+
+Image the flash layout from above and then we wanted to find the non-existent
+key TWO.
+
+The hash of TWO indicates it should be in region 1. First we check region 1,
+but we don't find TWO. Next we try to find TWO in region 2. We don't kind the
+TWO object there, but as the region isn't empty we can't determine if it didn't
+fit. Next we try region 0. As this region is empty we stop looking.
+
+### Invalidating a key
+
+When the key ONE is invalidated with `invalidate_key()`, the only change in
+flash will be the `valid` flag. The object header for ONE will now look like:
+
+```
+0x400                                                                                              0x52C
+--------------------------------------------------------------------------------------------------------
+||||| version|len/flag|   len  |                              hashed_key                               |
+|||||        |        |        |        |        |        |        |        |        |        |        |
+|||||    0x00|00000000|    0x34|    0xed|    0xa1|    0x00|    0x78|    0x88|    0x61|    0x93|    0xbb|
+--------------------------------------------------------------------------------------------------------
+              ^
+```
+
+No changes will happen in flash until key TWO has also been invalidated.
+At which point `garbage_collect()` can erase the region.
+
+## Limitations of TickFS
+
+### Fragmentation
+
+Although TickFS has a `garbage_collect()` function, it makes no effort to
+handle fragmentation.
+
+That means that if you have the following objects in a region
+
+```
+0x000                  0x400
+----------------------------
+|||||     Region 0     |||||
+|||||                  |||||
+|||||ONE|TWO|THREE|FOUR|||||
+----------------------------
+```
+
+Where ONE, TWO and FOUR have been marked as invalid the entire region will still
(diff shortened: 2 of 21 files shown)


libraries/tickfs/SPEC.md
```diff
@@ -0,0 +1,494 @@
...
+
+### Memory usage
+
+At a minimum TickFS requires a `PageSize` amount of free memory to store the
+buffer obtained when reading flash. This could be reduced by instead performing
+multiple reads of a smaller fixed size.
+
+### Synchronous
+
+TickFS is fully synchronous and every call is blocking.
```
bradjc, line 487: Do you have an idea as to how you would implement this? I'm guessing by using a new flash driver?

libraries/tickfs/SPEC.md
```diff
@@ -0,0 +1,494 @@
...
+
+This is unfortunately inherently different to Tock's asynchronous design. The
+justification for this design is that in general writing and erasing flash
+can be tricky if we are executing from the same bank. A synchronous design
+seems like an overall more useful implementation to allow this. It also means
+we block in the critical moments where a power loss would loose the data we
+want to preserve.
```
bradjc, line 494: What does the complexity for executing from the same bank as the file system look like?

libraries/tickfs/SPEC.md
alistair23: I think no matter what a new flash driver would need to be implemented. The current Tock interfaces are very Tock specific and I was trying to make this generic.

There are two options (or both):
 1. Update TickFS to also support asynchronous operations, we can use the callback style method used in Tock. This would take some work, but isn't impossible. Then we would just need a small translation layer to convert between Tock and TickFS.
 1. Add a new flash interface using synchronous calls.

libraries/tickfs/SPEC.md
alistair23: That will be hardware dependent. Most flash controllers have limitations when writing to/from flash. For example you can't load other data from flash while performing an erase. In past experience usually interrupts are disabled during a flash erase to make it a critical section.

There is some discussion about this here: https://github.com/tock/tock/issues/1459#issuecomment-554742330

By making this a synchronous operation we can hide all of that complexity and just offload it to the `FlashController` implementation.

libraries/tickfs/SPEC.md
bradjc: The flash HIL is read/write/erase page with callbacks. I don't see how that is overly Tock specific. There may have to be a shim layer to mostly do address conversion/restriction, but that would be a standard shim that any board with a flash driver could use.

The other benefit to using an asynchronous interface is that this FS would work on an external flash chip (like the nrf52840dk has) as well.

libraries/tickfs/SPEC.md
bradjc: Making an asynchronous operation look synchronous is challenging, but Tock has support calls for deferred calls which enable a synchronous operation to look asynchronous. If a flash controller driver needs to block during an erase it can, then it can set a deferred call to trigger a callback. However, if the flash controller knows the operation can happen asynchronously, it still has to block when it could instead wait for an interrupt.

alistair23: That's a good idea. The defered calls should let us re-use Tocks' Flash HIL with this current interface.

libraries/tickfs/SPEC.md
hudson-ayers: It only requires unsafe code to construct a static buffer that is initially passed to the capsule using the filesystem, so it is at least contained within `main.rs`. But I see that if not for the requirement of async operations, an anonymous lifetime would work, which is much nicer as it would allow clients of the filesystem to use stack allocated buffers for filesystem reads/writes

hudson-ayers:
> Tock currently has a flash HIL, but there is no implementation on top of that HIL to reliably and persistently store data over a long period of time for multiple apps.

This is true, but it is worth noting that the log interface exists (https://github.com/tock/tock/blob/master/capsules/src/log.rs), which allows for persistent storage of values in an append-only log across restarts. The log interface is missing some of the functionality in this PR - keys are just sequentially increasing numbers based on the number of writes, rather than allowing arbitrary user-set keys, and entries cannot be modified, only new entries can be added. Also, the log is not exposed to applications yet, and is not yet virtualized within the kernel.

libraries/tickfs/src/tickfs.rs
```diff
@@ -0,0 +1,673 @@
...
+                    return None;
+                }
+            }
+        };
+
+        Some(tickfs)
+    }
+
+    /// Generate the hash and region number from a key
+    fn get_hash_and_region(&self, hash_function: &mut H, key: &str) -> (u64, usize) {
```
hudson-ayers, line 108: The discussion in https://github.com/tock/tock/issues/1692 seemed to conclude that having keys be `&[u8]` or `u32` values, instead of `&str` values, would be preferred. Can you elaborate on why you chose `&str` keys?

alistair23: I have changed it to `&[u8]`

alistair23:
Ping!

libraries/tickfs/README.md
```diff
@@ -0,0 +1,95 @@
...
+TickFS saves and restores objects from flash. TickFS objects contain the value
+the user wanted to store as well as extra header data. Objects are internal to
+TickFS and users don't need to understand them in detail to use it.
+
+For more details on the technical implementation see the [SPEC.md](./spec.md) file.
+
+### Collisions
+
+TickFS will prevent a new key/value pair with a colliding hash of the key to be
+added. The collision will be reported to the user with the `KeyAlreadyExists`
```
phil-levis, line 60: Should this be `ErrorCode::KeyAlreadyExists`?

libraries/tickfs/README.md
```diff
@@ -0,0 +1,95 @@
...
+`ErroCode`.
+
+### Power loss protection
+
+TickFS ensures that in the event of a power loss, no stored data is lost or
```
phil-levis, line 65: Can you phrase TickFS's guarantees in terms of ACID semantics? "No stored data is lost or corrupted" could mean a lot of things. E.g., what if the data isn't stored? I think what you're getting at is both Atomicity and Durability but I am not sure.

libraries/tickfs/README.md
```diff
@@ -0,0 +1,95 @@
...
+that data is lost.
+
+To help reduce this time to be as short as possible the `FlashController`
+is synchronous. Although flash writes can take a considerable amount of time
+and this will stall the application, this still seems like a good idea
+to avoid loosing data.
+
+### Security
+
+TickFS uses check sums to check data integrity. TickFS does not have any measures
```
phil-levis, line 80: Checksums, CRCs, or some form of hash? Or undefined/implementation-specific? Why are checksums needed?

libraries/tickfs/SPEC.md
```diff
@@ -0,0 +1,494 @@
...
+
+The values can be any length as long as they follow both:
+ * Don't span multiple regions. That limits the maximum value length to
+   `region_size - size_of::<ObjectHeader>()`
+ * Don't have a maximum length greater then 4KiB (0xFFF).
+
+#### Check Sum
+
+The check sum is a hash of the entire object (not including the check sum).
+The check sum is calculated using the same hash algorithm used for the keys.
```
phil-levis, line 117: Then it's not a checksum.

Please note that hashes (especially cryptographic hashes) and CRCs have very different error detection behaviors. A CRC (depending on the generator polynomial) provides guarantees on certain errors it can detect (e.g., an odd number of bit errors). A hash provides no guarantees. A strong cryptographic hash, for example, means that flipping one bit in the input makes every bit in the output effectively random. This means with N bits there is a 2^-N probability you get a collision. It can be OK to make this tradeoff, but a CRC is actually in some ways better.

libraries/tickfs/SPEC.md
```diff
@@ -0,0 +1,494 @@
...
+### Somewhat high storage overhead
+
+The storage overhead is somewhat high for TickFS. This is mostly due to the
+two 64-bit hashes that are stored with every object.
+
+The 64-bit value is determined by the output of the standard Rust
+`core::hash::Hasher` trait.
+
+For the hash of the key a 64-bit value can be justified by the lack of
+collision avoidance in the implementation. If two keys have the same
```
phil-levis, line 461: Could you add salt, to avoid collisions? E.g., if you insert and it fails due to a collision, you can retry with a different salt?

libraries/tickfs/SPEC.md
phil-levis: This needs to be asynchronous: we talked about this in a call. While on-chip flash can behave synchronously, that's only true if you're writing to the bank you are executing out of. If you have off-chip flash (e.g., over a SPI) you do not want to put a blocking interface on top of it.

libraries/tickfs/src/flash_controller.rs
```diff
@@ -0,0 +1,79 @@
+//! The Flash Controller interface with hardware
```
phil-levis, line 1: What is the relationship between a FlashController and an allocated in-kernel storage region made with the `storage_volume` macro? Does this represent a controller over a logical region (e.g., a storage region) or the physical entire flash? It seems like TickFS should operate over a storage region -- is that how it's intended to operate?

phil-levis (commented):
Overall, I think this looks like a good design. It's nice that it's lean and fits in 673 lines of code. 

It's a little hard to do a detailed code review for bugs because the guarantees of operations are not 100% clear. I think it would be good to phrase the specification using TRD terms, so saying what an implementation MUST do, or SHOULD do, or MAY do. Separating its layout and API from a specific implementation would be good.

libraries/tickfs/README.md
alistair23: Checksums are used for data integrity to ensure bits don't flip on the flash. There are more details in the spec document.

libraries/tickfs/SPEC.md
alistair23: Yep, I thought about this.

I would prefer to use a CRC here, the problem is we then also need a CRC implementation as well as a hash implementation in the kernel. The hope is that the hash implementation can be offloaded to hardware, so re-using it for a checksum will hopefully speed things up.

I think I'll move toward using a CRC though, as I agree it will provide much more useful error checking.

libraries/tickfs/SPEC.md
alistair23: The interface in Tock will be async, see https://github.com/tock/tock/pull/2205 for how this is done.

libraries/tickfs/SPEC.md
alistair23: That's an option, but how do we find the salt when looking up the keys? Do we just have a list of salts to walk through?

libraries/tickfs/README.md
alistair23: Done.

libraries/tickfs/src/flash_controller.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use crate::error_codes::ErrorCode;
+
+/// Implementation required for the flash controller hardware. This
+/// should read, write and erase flash from the hardware using the
+/// flash controller.
+///
+/// This is the public trait for the Flash controller implementation.
+///
+/// The PageSize 'P' must be the smallest size that can be erased in
```
bradjc, line 11: Does `PageSize` refer to something in the code?

libraries/tickfs/src/flash_controller.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    /// This function must write the length of `buf` to the specified address
+    /// in flash.
+    /// If the length of `buf` is smaller then the minimum supported write size
+    /// the implementation can write a larger value. This should be done by first
+    /// reading the value, making the changed from `buf` and then writing it back.
+    ///
+    /// On success it should return nothing, on failure it
+    /// should return ErrorCode::WriteFail.
+    fn write_region(&self, address: usize, buf: &[u8]) -> Result<(), ErrorCode>;
```
bradjc, line 72: This doesn't need a region num?

libraries/tickfs/src/flash_controller.rs
```diff
@@ -0,0 +1,79 @@
...
+/// should read, write and erase flash from the hardware using the
+/// flash controller.
+///
+/// This is the public trait for the Flash controller implementation.
+///
+/// The PageSize 'P' must be the smallest size that can be erased in
+/// a single operation.
+/// The start and end address of the FlashController must be aligned
+/// to the PageSize.
+/// All `region_number`s and `address`es are offset from zero. If you
```
bradjc, line 15: How do an implementer and user agree on how large a region is?

libraries/tickfs/src/flash_controller.rs
alistair23: Good catch, this is an out of date comment. I'll remove it.

libraries/tickfs/src/flash_controller.rs
alistair23: No, it's much more efficient to just write the smallest amount of data possible to a specific address. Generally flash controllers don't require an entire page write.

libraries/tickfs/src/flash_controller.rs
alistair23: I have updated the documentation to make this clearer.

libraries/tickfs/src/flash_controller.rs
bradjc: Maybe this should just be called `write()` then?

libraries/tickfs/src/flash_controller.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// should be read.
+    ///
+    /// On success it should return nothing, on failure it
+    /// should return ErrorCode::ReadFail.
+    fn read_region(
+        &self,
+        region_number: usize,
+        offset: usize,
+        buf: &mut [u8],
+    ) -> Result<(), ErrorCode>;
```
bradjc, line 64: Is `buf` required to end up entirely within a single region?

libraries/tickfs/src/flash_controller.rs
bradjc: What about making `get_region_size()` a member function in the trait (or maybe some other rust-y approach)?

libraries/tickfs/src/flash_controller.rs
alistair23: So right now buf will contain the entire region size. In future it might be worth allowing a smaller buf and read part way through a region.

libraries/tickfs/src/flash_controller.rs
alistair23: Good point, I'll fix this.

libraries/tickfs/src/flash_controller.rs
alistair23: With const generic arrays I think this would be a little clearer. I'll see if I can find a rust-y way to make this nicer.

alistair23:
Ok, I have pushed an update.

@hudson-ayers and @bradjc I have included the async changes in this PR now, hopefully that will help with review. I have also included a little more justification for the API at the end of the SPEC document. They basically say the same thing I was saying this morning, in that I think the sync design is helpful for others and unit testing and that due to the loop nature of lookups even with callbacks we will still end up with a retry style loop anyway after the callback has occurred.

@bradjc I have switched over to const generics for the flash region accesses. This then nicely ensures that the buffers all match the region sizes and that it is clear to users how everything should work together. If we don't want to introduce an unstable feature I can revert the change, I feel it indicates what it will look like eventually anyway.

Let me know what you think.

libraries/tickfs/src/tickfs.rs
```diff
@@ -0,0 +1,930 @@
...
+                    }
+                }
+                State::AppendKey(key_state) => match key_state {
+                    KeyState::ReadRegion(reg) => reg as isize,
+                },
+                State::GarbageCollect(state) => match state {
+                    RubbishState::ReadRegion(reg) => reg as isize,
+                    _ => unreachable!(),
+                },
+                _ => unreachable!(),
```
bradjc, line 445: What prevents a user from calling `read_key()`, then an async operation starting, but then calling `append_key()` before calling `continue_op`? I.e., why can't the current state be `GetRead`?

alistair23: This prevents that. It doesn't support two concurrent operations.

Handling that would be very complex. What if two append_key operations called, how do we then make guarantees about ordering and not loosing data?

Only 1 operation can be in progress at a time, we require the operation to complete before a new one can start.

libraries/tickfs/SPEC.md
```diff
@@ -0,0 +1,509 @@
...
+be classified as full. It will only not be full when all four objects are marked
+for invalid.
+
+It would be possible to have a reserved region and then free up the region with
+the following steps:
+ 1. Move the region to the reserved region
+ 1. Erase the above region
+ 1. Move the valid objects from the reserved region to the original region,
+    after being defraged in memory
+ 1. Erase the reserved region
```
phil-levis, line 435: Does this mean that the reserved region is erased for every erase? I.e. it will fail first.

libraries/tickfs/SPEC.md
```diff
@@ -0,0 +1,509 @@
...
+collision avoidance in the implementation. If two keys have the same
+hash the second key will be dropped. In this case a 64-bit hash should
+hopefully make that occurrence very unlikely. Also by following the standard
+Rust `core::hash::Hasher` trait the user is free to implement any standard
+Hasher of their choosing. Some systems will even be able to offload the
+hash to hardware.
+
+As for the check-sum. It was seriously considered to change this to a
+32-bit or even 16-bit CRC. This would reduce the overhead for each object
+by 4 or 6 bytes. Although a CRC is relatively easy to implement, it can still
```
phil-levis, line 470: How long is it in comparison to a flash erase/write? This seems like a handwavy "well I'd have to implement it". 

What are typical error patterns in MCU flash chips? It would be good to see what other systems do and why.

libraries/tickfs/SPEC.md
```diff
@@ -0,0 +1,509 @@
...
+Another reason for the synchronous support is that generally when there is
+only a single flash bank all flash writes and erases must be done synchronously.
+The synchronous API should allow these type of operations to occur.
+
+On top of that it is much simpler to unit test a synchronous API. This means the
+TickFS implementation can take advantage of a large range of unit tests to
+ensure correctness in the design. This is more difficult and complex to do in
+a async only operation.
+
+The async functions (see the documentation) can be used to implement a fully
```
phil-levis, line 504: If you want TickFS to be completely general for the type of NV storage it's on top of, having async and sync APIs makes sense. For example, FRAM is close enough to RAM speeds (and is usually integrated with RAM e.g. in terms of addressing) that it's not useful to have an async API.

Can this document describe the API and how this works?

Why can't the callback that tells a client that an async operation has completed be part of the API? It could be a client doesn't bother registering these callbacks, or a synchronous implementation doesn't ever issue them. I think it's very important that this callback originate from the same module, at least -- you shouldn't have to register a callback with some board-and-TickFS-implementation-specific callback to be able to use TickFS.

Do you retry an operation because it wasn't complete? Or because the implementation was busy?

libraries/tickfs/README.md
```diff
@@ -0,0 +1,93 @@
...
+key value pairs to be stored in Flash Memory.
+
+TickFS was written to allow the Tock OS kernel to persistently store app data
+on flash. It was written to be generic though, so other Rust applications can
+use it if they want.
+
+TickFS is based on similar concepts as
+[Yaffs1](https://yaffs.net/documents/how-yaffs-works]).
+
+TickFS provides ACID properties.
```
phil-levis, line 13: How does it provide isolation? This implies a particular ordering of different clients. It would be good to explain how it provides each.

libraries/tickfs/SPEC.md
```diff
@@ -0,0 +1,509 @@
...
+last two bytes of an object hashed-key modulo the number of regions.
+
+This will produce a number that is between zero and the total number of
+TickFS regions. This number is the regions number where the data is stored
+or loaded from.
+
+This allows us to quickly determine which flash region a key is stored in. This
+should improve lookup time by reducing the number of reads required.
+
+If a give region is full, we will start to look in neighboring regions. We first
```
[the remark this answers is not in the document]
alistair23: Thanks for pointing those out. Doesn't double hashing have the disadvantage of having to calculate two hashes? That could potentially slow down lookups. Also it seems to take up more space right?

Cuckoo hashing changes the medium right? Which is difficult to do safely (power loss resilient) in flash.

I'm all on board with other methods, it is just everything seems to have compromises and it's hard to get the balance right.

libraries/tickfs/SPEC.md
```diff
@@ -0,0 +1,509 @@
...
+regularly write to the same flash region.
+
+## Storage Format
+
+TickFS flash storage is formatted as a sorting of regions with each region
+containing TickFS objects.
+
+### TickFS Regions
+
+A TickFS region is the smallest set of contiguous flash storage space that can
```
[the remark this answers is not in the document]
alistair23: Good questions.

Right now it can not be larger. The main reason is to avoid having to do multiple reads once we have found a key. Also writing over flash bounds will be a little more tricky. I think this is something that could be added in the future. It's not a fundamental limitation, just an implementation limitation.

There is a KV header entry. Right now it's empty. I will look at saving the region size in there.

libraries/tickfs/SPEC.md
```diff
@@ -0,0 +1,509 @@
...
+marked as !`valid`.
+
+We then check to see if the object hash matches the hash we are
+looking for. If it doesn't we move forward in the loaded region by the
+total length of the object we just checked and start the process again.
+
+We continue this loop until we either find the key we are looking for or
+find a version 0xFF, indicating the end of the blocks in that region.
+
+This method allows a quick retrieval of data from a given key. We only
```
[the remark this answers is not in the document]
alistair23: When all regions are full then yes. A lookup will search all of flash.

I can't really think of a way to avoid doing that though. Having some sort of list of keys would use space, and I think space is going to be too important for that.

libraries/tickfs/SPEC.md
```diff
@@ -0,0 +1,509 @@
...
+
+### Memory usage
+
+At a minimum TickFS requires a `PageSize` amount of free memory to store the
+buffer obtained when reading flash. This could be reduced by instead performing
+multiple reads of a smaller fixed size.
+
+### Synchronous and Asynchronous Usage
+
+TickFS supports both an synchronous and asynchronous usage. For developers
```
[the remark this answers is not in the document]
alistair23: This is the spec document. The idea is that it covers the spec in details for those interested. Most users won't need to read this, instead they can read the Rust documentation (auto generated from the code) to see the API. This also ensures that cargo will test the documentation to ensure it's always up to date.

libraries/tickfs/SPEC.md
```diff
@@ -0,0 +1,509 @@
...
+
+One of the goals of TickFS is to allow other non Tock users to use it. The
+current method should allow the library to be used to crates.io by other
+developers.
+
+Another reason for the synchronous support is that generally when there is
+only a single flash bank all flash writes and erases must be done synchronously.
+The synchronous API should allow these type of operations to occur.
+
+On top of that it is much simpler to unit test a synchronous API. This means the
```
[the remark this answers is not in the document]
alistair23: I would disagree. Part of designing something is to allow it to be easy to test. TickFS would be useless if we can't reliably test it for both successes and failures.

libraries/tickfs/SPEC.md
alistair23: This is just an example. I would expect that a reserved region would end up being used for every erase yes. It would be possible to move the reserved region (it changes every time to the last region you wanted to erase) but that then makes lookups slower as we loose ordering.

libraries/tickfs/SPEC.md
alistair23: Good questions. I honestly don't know.

I was originally torn what to do here, I was starting to think a CRC would be better here. The reason I didn't is that I know some MCUs offer CRC offloading, in which case I'm assuming it is slow enough that a hardware off loader becomes beneficial.

libraries/tickfs/SPEC.md
alistair23: This is the spec documetnation, it describes the underlying spec and not the usage interface. The information for how to use the API is in the rust docs.

Do you mean add a Tock callback function to the API that is then set as a client in the kernel?

We retry because it wasn't complete.

libraries/tickfs/README.md
alistair23: It doesn't support concurrency so everything is isolated :)

I'll add some explanation and probably remove isolation.

libraries/tickfs/SPEC.md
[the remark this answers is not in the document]
phil-levis: Depending on the second hash, you could be orders of faster on speed. The time it takes to parse a single header is probably greater than to compute a simple hash. Note that the second hash can be much simpler, if the first one is strong. E.g., you could divide by the number of regions and then mod by the number of regions (assuming you don't have too many regions).

libraries/tickfs/SPEC.md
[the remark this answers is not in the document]
phil-levis: But the point isn't whether you can test it, but rather how much work writing those tests are. Of course you can test an asynchronous system.

libraries/tickfs/SPEC.md
phil-levis: I think you want to do some back of the envelope calcuations on how long it takes to compute a reasonable CRC and hash compared to the time it takes to erase a page and time it takes to write a page. I think you will find they are quite fast.

libraries/tickfs/README.md
phil-levis: ACID semantics are usually defined in terms of transactions, so you need to state what a transaction is. A single read or write? If you define its transactions as single operations, and there are no concurrent operations, then yes, it has isolation. But this means, for example, that a client can't safely check for a key, and if it's not present then insert it.

libraries/tickfs/SPEC.md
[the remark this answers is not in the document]
phil-levis: You mention multi-page writes: what are the durability and atomicity guarantees of flash writes/erases? What happens if you lose power in the middle of a write/erase? If you do not have atomicity guarantees here then you have to design the system differently to provide ACID semantics.

alistair23:
@bradjc, @hudson-ayers, @lschuermann I have re-worked the async interface. It no longer uses retries and is more in line with Tock style code.

This has the disadvantage of now requiring static lifetime buffers. Which requires unsafe to create and in the Tock kernel we need to copy data from app slices to kernel buffers, which now means we need even more static buffers.

Let me know what you think.

libraries/tickfs/SPEC.md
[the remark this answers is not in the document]
alistair23: Aren't you always going to have that problem though? If we loss power when some data has been written, but not all the data then we are going to loose that key. Is there much we can do about that?

alistair23:
@phil-levis I updated some parts of the spec and readme and also changed the name.

Pending spec issues:
 - What should happen to the checksum? Change to a CRC, remove the checksum?
 - Options to avoid collisions. Use a salt?
 - Possibly swap to double hash or cuckoo hashing

Am I missing anything?

libraries/tickfs/SPEC.md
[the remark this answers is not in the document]
phil-levis: You do not always have that problem: https://en.wikipedia.org/wiki/Write-ahead_logging

This is how all modern file systems and databases are built. Filesystems usually rely on block atomicity, which storage units provide (i.e., have enough capacitance to complete an operation once started). But they use this block atomicity to build higher-level atomicity. Note that in RAM-scarce settings, WAL does require two writes.

It's OK if TicKV doesn't provide full ACID. But you should be very clear about what guarantees it does provide.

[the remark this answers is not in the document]
alistair23: Ah, yep. WAL doesn't really work on flash though, at least not without wear levelling issues.

Ok, I think I have described that. We are dependent on the flash implementation for the operations. Let me know if you think I should change the README.

libraries/tickv/README.md
```diff
@@ -0,0 +1,110 @@
...
+## ACID characteristics
+
+TicKV provides some ACID properties. For the purpose of ACID a transaction is
+a key operation, that is finding, adding, invalidating or fully removing
+(garbage collection) a key.
+
+Atomicity: TicKV does not provide atomicity. If a operation fails mid write
+then there will be an invalid key/value stored to flash.
+
+Consistency: Consistency is maintained similar to atomicity. All operations
```
phil-levis, line 41: What does "similar to atomicity" mean? The atomicity comment says a database can have an invalid key/value. So it is possible for an operation to take the store into an invalid state?

It is useful to state when it *is* atomic or consistent. E.g., you can say that "Under the constraint that underlying flash operations succeed, TicKV provides atomicity. If the underlying flash has an error, the system crashes mid-operation, or power is lost mid-operation, an operation may be partially committed."

libraries/tickv/README.md
```diff
@@ -0,0 +1,110 @@
...
+can only take the database from a valid state to another valid state.
+
+Isolation: TicKV only allows a single operation at a time. In this way it
+provides isolation. The layer above TicKV is responsible for handling
+concurrent accesses by deferring operations for example.
+
+Durability: TicKV ensures durability and once a transaction has completed
```
phil-levis, line 48: Should put a note about wear-leveling here.

libraries/tickv/README.md
```diff
@@ -0,0 +1,110 @@
...
+ * No external crates in use (not including unit tests)
+
+TicKV is also designed with some assumptions
+
+ * Most operations will be retrieving keys
+ * Some operations will be storing keys
+ * Keys will rarely be deleted
+ * Key values will rarely need to be modified
+
+## ACID characteristics
```
phil-levis, line 32: I would just call this "System Guarantees"

libraries/tickv/SPEC.md
```diff
@@ -0,0 +1,510 @@
...
+
+Flash has the characteristic that although read/writes can happen at small
+granularities an erase operation requires a entire block (specified
+by the region size).
+
+Due to this removing a key with the `invalidate_key()` function does NOT
+remove anything from flash. Instead calling `invalidate_key()`
+will mark the `valid` boolean for that object as `false` (0).
+
+This is done because changing a `1` to a `0` in flash can be done with a
```
phil-levis, line 170: Many chips allow a limited number of writes per erase cycle on a page. E.g., the EFMPG1 Reference Manual reads:

"It is possible to write words twice between each erase by keeping at 1 the bits that are not to be changed. Let us take as an example
writing two 16 bit values, 0xAAAA and 0x5555. To safely write them in the same flash word this method can be used:
• Write 0xFFFFAAAA (word in flash becomes 0xFFFFAAAA)
• Write 0x5555FFFF (word in flash becomes 0x5555AAAA)
Note:
• There is a maximum of two writes to the same word between each erase due to a physical limitation of the flash.
• During a write or erase, flash read accesses will be stalled, effectively halting code execution from flash. Code execution continues
upon write/erase completion. Code residing in RAM may be executed during a write/erase operation."

It's OK to say that TicKV only works on flash devices that allow N>K writes to a word. It looks like TicKV requires 2 writes be allowed. That means almost all (but perhaps not all) can work. You want this limitation up at the top, not at the bottom.

What happens if you re-insert a key? Does it fail or do you replace?

phil-levis (commented):
The absence of any real in-memory data structures is a limitation here. It means, for example, once a key write is corrupted there's no way out, because you can't do a page-remapping. It could be, for example, that each page says its logical page number (which might be different than its physical), and at boot software scans the pages and build a logical->physical table mapping.

Given how far the design is from ACID, you might just want to instead explicitly state the guarantees:
  - A successful write or erase is durably committed.
  - The design does not support concurrency, such that it imposes a total order on all read, write, and delete. operations. Successful individual operations are therefore atomic. Applications that require higher-level atomicity (e.g., read/modify/delete/write, multiple writes) need to build this on top of these operations.
  -  TicKV does is not robust to low-level flash failures, power loss, or system crashes. However, a failure only affects a single key: a failure to write or delete key K will corrupt at most only K (*is this correct? I think it is from the append structure*).

libraries/tickv/README.md
phil-levis: What happens if a write fails? Is the rest of the page considered full?

bradjc (commented):
note: I'm not reviewing the library itself.

At the risk of going in circles, I think `async_ops.rs` can be removed from this PR. I believe that file can be entirely contained in a Tock capsule which implements the future KV HIL using TickV. Otherwise I think this library is a good addition to Tock (and perhaps should be spun out someday!).

libraries/tickv/README.md
alistair23: I have updated this in the doc to answer all your questions. Let me know what you think. I have pasted it below as well:

> ## System Guarantees
> 
> A successful write or erase is durably committed.
> 
> The design does not support concurrency, such that it imposes a total order
> on all read, write and delete operations. Successful individual operations
> are therefore atomic. Applications that require higher-level atomicity
> (e.g., read/modify/delete/write, multiple writes) need to build this on top
> of these operations.
> 
> TicKV is not robust to low-level flash failures, power loss, or system
> crashes. However, a failure only affects a single key: a failure to write
> or delete key K will corrupt at most only K. A value to read will have no
> effect.
> 
> Under the constraint that underlying flash operations succeed, TicKV provides
> atomicity. If the underlying flash has an error, the system crashes
> mid-operation, or power is lost mid-operation, an operation may be
> partially committed.
> 
> If a write fails with only a partial write to flash the entire length of
> the key will be used in the storage.
> 
> TicKV ensures durability and once a transaction has completed
> and been committed to flash it will remain there. TicKV also takes measures
> to apply wear levelling to the flash storage.

libraries/tickv/SPEC.md
alistair23: Yeah, I have seen some devices with limitations.

As long as we can write twice between erases. Once to store the data and once to invalidate then it should work. From what I can tell it's not very common that you can't write that many times.

I have added a note to the README.

To re-insert a key you first have to delete the key then append it again. It won't replace the original key, it will be appended. The first key (that was invalidated) is marked for deletion and can't be read.

libraries/tickv/src/async_ops.rs
```diff
@@ -0,0 +1,923 @@
...
+type ContinueReturn = (
+    // Result
+    Result<SuccessCode, ErrorCode>,
+    // Buf Buffer
+    Option<&'static mut [u8]>,
+);
+
+/// The struct storing all of the TicKV information for the async implementation.
+pub struct AsyncTicKV<'a, C: FlashController<S>, H: Hasher, const S: usize> {
+    /// The controller used for flash commands
```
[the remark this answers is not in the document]
alistair23: I agree that it can be removed with the HIL. Whatever order it all ends up being approved I can rebase from there.

There will be more changes required to match the new HIL, but that should be easy to address.

alistair23:
> note: I'm not reviewing the library itself.
> 
> At the risk of going in circles, I think `async_ops.rs` can be removed from this PR. I believe that file can be entirely contained in a Tock capsule which implements the future KV HIL using TickV. Otherwise I think this library is a good addition to Tock (and perhaps should be spun out someday!).

I agree that it can be removed with the HIL. Whatever order it all ends up being approved I can rebase from there.

There will be more changes required to match the new HIL, but that should be easy to address.

alistair23:
> The absence of any real in-memory data structures is a limitation here. It means, for example, once a key write is corrupted there's no way out, because you can't do a page-remapping. It could be, for example, that each page says its logical page number (which might be different than its physical), and at boot software scans the pages and build a logical->physical table mapping.

Yes, it will be saved in flash. The implementation can probably be improved to allow the invalid key to be marked as invalid, so that it won't conflict if a new key is added. That way if a corrupt key is saved a real key can be saved again on reboot.

The logical->physical map is a good idea. I think that should be possible to add in the future, without a major re-work. It will have the disadvantage of larger memory usage, but it might be worth it.

> 
> Given how far the design is from ACID, you should say it's not a ACID system and instead explicitly state the guarantees:
> 
>     * A successful write or erase is durably committed.
> 
>     * The design does not support concurrency, such that it imposes a total order on all read, write, and delete. operations. Successful individual operations are therefore atomic. Applications that require higher-level atomicity (e.g., read/modify/delete/write, multiple writes) need to build this on top of these operations.
> 
>     * TicKV does is not robust to low-level flash failures, power loss, or system crashes. However, a failure only affects a single key: a failure to write or delete key K will corrupt at most only K (_is this correct? I think it is from the append structure_).

I put most of this in the README.

libraries/tickv/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+//!
+//!  * Version 0
+//!    * Version 0 is a draft version. It should NOT be used for important data!
+//!      Version 0 maintains no backwards compatible support and could change at
+//!      any time.
+//!
+
+#![feature(min_const_generics)]
+#![no_std]
+#![deny(missing_docs)]
```
bradjc, line 217:
```suggestion
#![feature(min_const_generics)]
#![no_std]
#![forbid(unsafe_code)]
#![deny(missing_docs)]
```

bradjc (approved):
Alright in my opinion this is great addition to Tock. It's an sync and async k-v implementation entirely in safe rust.

We also now have a low-level key-value store HIL interface which I believe a small shim capsule should be able to provide that trait with this implementation. Then ~~we~~ K-V master Alistair can build the rest of the stack.

@phil-levis any remaining issues?

libraries/tickv/src/lib.rs
hudson-ayers: once rebased, `min_const_generics` is no longer needed

libraries/tickv/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+//!
+//! TicKV is currently version 0.
+//!
+//!  * Version 0
+//!    * Version 0 is a draft version. It should NOT be used for important data!
+//!      Version 0 maintains no backwards compatible support and could change at
+//!      any time.
+//!
+
+#![feature(min_const_generics)]
```
bradjc, line 215:
```suggestion
```

alistair23:
Rebased and addresses comments

bradjc (approved):

Issue #2143, opened 2020-10-09
Alarm driver does not work with 64-bit alarms


lschuermann:
While porting to a new board that features a 64-bit HW Timer and hence a 64-bit SW Alarm peripheral, I've noticed weird behavior where an Alarm would be set, fire, and then be immediately reset to the same alarm time (now in the past).

I've managed to trace this down to the Alarm driver, specifically the Alarm callback function that checks for each app whether a set alarm has expired.

Since the alarm times (`reference` and `dt`) are stored as 32-bit values, adding `dt` to `reference` will potentially wrap at reaching `u32::MAX`, however the comparison to the current time will still indicate that the current time is in between `reference` and `dt`, even though the alarm time would have passed.

https://github.com/tock/tock/blob/ba867c2c269325133f212b757b125b114ab9f2ad/capsules/src/alarm.rs#L260-L270

This will cause `num_armed` to not be decreased, which will in turn call `reset_active_alarm()`, setting it to the original alarm time  and hence firing immediately, resulting in a loop.

https://github.com/tock/tock/blob/ba867c2c269325133f212b757b125b114ab9f2ad/capsules/src/alarm.rs#L284-L291

## Current behavior

Given a 64-bit alarm peripheral, once an app-alarm is set in a way that `reference < u32::MAX && (reference + dt) > u32::MAX`, the alarm driver will call `set_alarm(` on the alarm with constant arguments in a loop. 

## Expected behavior

Handle alarm times correctly and don't hang in a loop.

## Example

For example, given

| Field                                        | Value                    |
|---|---|
| `reference`                              | `4_294_967_263` |
| `dt`                                          | `64`                       |
| `reference.wrapping_add(dt)` | `32`                       |
| `now`                                      | `4_294_967_370`  |

the problem is evident, since technically `reference < reference + dt < now` (and hence the alarm should fire), but given the particular wrapping semantics in the comparison as done currently, `reference < now < reference.wrapping_add(dt)` (`4_294_967_263 < 4_294_967_370 < 32`).

While changing this code to read
```rust
                if !now.within_range(
                    A::Ticks::from(reference),
                    A::Ticks::from(reference).wrapping_add(A::Ticks::from(dt)),
                ) {
```
and hence use 64-bit wrapping both in the comparison and the addition does solve the alarm looping issue, it's technically incorrect since it will compare the 64-bit absolute time to the 32-bit alarm time set in userspace and hence at some point fire *before* the userspace alarm time.

## Proposed approach

[...]

PR #2160, opened 2020-10-15
alarm-driver: migrate to native Alarm::Ticks

lschuermann:
### Pull Request Overview

This pull request changes the alarm driver capsule to store pending alarms using the underlying `<Alarm as kernel::hil::time::Time>::Ticks` type instead of `u32`.

Subjectively this makes the alarm driver more comprehensible, by dividing the responsibilities:
- to set an alarm, an initial conversion to the native `A::Ticks` is required once
- when resetting an alarm, during comparison and while checking whether an alarm already expired (which was the source of #2143), the capsule exclusively works with the native `A::Ticks` type, making this (more complex) code easier to reason about

### Testing Strategy

This pull request was tested by running on a board with a 64-bit 100MHz Alarm and does seems to solve #2143. Further testing across different boards and Alarms is required for me to be confident that this works. With some work I can change my FPGA-Platform to use various timer widths.

### TODO

This pull request still needs extensive testing.

It is an opposing approach to [@phil-levis strategy to pretend the underlying alarm is 32-bit wide](https://github.com/tock/tock/issues/2143#issuecomment-708561532), hence the advantages and disadvantages of each approach should be compared, deciding which strategy to go with (given that they work).

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/alarm.rs
+++ b/capsules/src/alarm.rs
@@ -2,28 +2,27 @@
 //! a point in time has been reached.
 
 use core::cell::Cell;
-use kernel::hil::time::{self, Alarm, Frequency, Ticks};
+use kernel::hil::time::{self, Alarm, Frequency, Ticks, Ticks32};
 use kernel::{AppId, Callback, Driver, Grant, ReturnCode};
 
 /// Syscall driver number.
 use crate::driver;
 pub const DRIVER_NUM: usize = driver::NUM::Alarm as usize;
 
-// This should transition to using Ticks
 #[derive(Copy, Clone, Debug)]
-enum Expiration {
+enum Expiration<T: Ticks> {
     Disabled,
-    Enabled { reference: u32, dt: u32 },
+    Enabled { reference: T, dt: T },
 }
 
 #[derive(Copy, Clone)]
-pub struct AlarmData {
-    expiration: Expiration,
+pub struct AlarmData<T: Ticks> {
+    expiration: Expiration<T>,
     callback: Option<Callback>,
 }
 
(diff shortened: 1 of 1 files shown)


phil-levis:
I kind of think this is 6 of one, half a dozen of the other, in that there are two options, either is fine. You can either store the alarm in the underlying Ticks type and do the conversion from 32 bits when you set the alarm, or you can store the alarm in the syscall 32 bits and do the conversion when you set the actual alarm. I personally chose the latter approach because the conversion itself is quite simple and just happens in one code sequence.  This patch does make the code 13 lines longer.

I'm kinda ambivalent at a basic level, since either approach is fine. But we did test this implementation a lot in a lot of different cases and I'm wary of doing that again.

lschuermann:
I guess I don't have much stake in it either, the current way is fine and tested.

Personally I still think that converting to native Ticks while setting the alarm is easier to reason about, in particular when debugging the code (as the alarm times can then be compared to the alarm peripheral directly). This does however introduce some amount of code duplication. Potentially it might reduce runtime complexity. Both effects are however on a scale where it shouldn't matter much.

In any case, we should remove this comment then (which is why I implemented it this way in the first place):

https://github.com/tock/tock/blob/23b0737ee6654f1193f11d93291e8319c7e2aa8e/capsules/src/alarm.rs#L12-L17

@phil-levis Depending on your decision, I can rebase this or you are free to close the PR. I'm fine either way.

phil-levis:
#2186

PR #2163, opened 2020-10-15
Fix ble_advertising_driver alarm logic

hudson-ayers:
### Pull Request Overview

The update to the new alarm HIL introduced incorrect logic in the `ble_advertising_driver`, such that it never recognized alarm callbacks as corresponding to any apps. The logic bug seems to stem from an assumption that the `t0` variable contained the time at which the alarm fired, rather than the reference time at which the alarm was set. The move to the alarm HIL introduced separate tracking of the reference time, so I just deleted the `t0` variable to prevent future confusion.

FWIW, after debugging this, I think that the `within_range()` function in the new alarm HIL is extremely error prone - without looking at the documentation of the method, I would expect `let expired = expiry_time.within_range(reference, now)` to assign true in a callback for when an alarm set for expiry_time expires. On the nrf52dk it *sometimes* does not, because `expiry_time == now` depending how quickly the callback is handled, and when they are equal the function returns false. I understand that the intended way to check for alarm expiration is `let expired = !now.within_range(reference, expiry_time)`, but it is not at all intuitive from the function signature that only that approach will be correct. IMO the publicly exported function should be like `has_expired(reference, now, expiry_time)` to prevent the very nefarious bugs that can arise from misuse of `within_range()`.

Changes to the time HIL are of course outside the purview of this PR.

### Testing Strategy

This pull request was tested by running the ble apps on an nrf52dk.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/ble_advertising_driver.rs
+++ b/capsules/src/ble_advertising_driver.rs
@@ -138,14 +138,12 @@ enum Expiration {
 
 #[derive(Copy, Clone)]
 struct AlarmData {
-    t0: u32,
     expiration: Expiration,
 }
 
 impl AlarmData {
     fn new() -> AlarmData {
         AlarmData {
-            t0: 0,
             expiration: Expiration::Disabled,
         }
     }
@@ -288,7 +286,6 @@ impl App {
 
     // Set the next alarm for this app using the period and provided start time.
     fn set_next_alarm<F: Frequency>(&mut self, now: u32) {
-        self.alarm_data.t0 = now;
         let nonce = self.random_nonce() % 10;
 
         let period_ms = (self.advertisement_interval_ms + nonce) * F::frequency() / 1000;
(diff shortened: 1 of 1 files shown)


bradjc:
`let expired = !now.within_range(reference, expiry_time)` is kind of interesting, because that won't work for checking expiration if `now < reference`. I find your argument for a `has_expired()` function compelling (would that even requiring formally changing the HIL, or just adding a helper function?). I also think that `if expired <= now {}` would be clear (if it can be made to work with wrapping).

bradjc (approved):

alevy (approved):

phil-levis:
> `let expired = !now.within_range(reference, expiry_time)` is kind of interesting, because that won't work for checking expiration if `now < reference`. I find your argument for a `has_expired()` function compelling (would that even requiring formally changing the HIL, or just adding a helper function?). I also think that `if expired <= now {}` would be clear (if it can be made to work with wrapping).

Hudson and I chatted about this in our weekly meeting today. Our conclusion is that the HIL should not change. `within_range` is important for Alarm/Timer implementations. That being said, the argument that you'd like to be able to reverse the operation (having a particular alarm, and asking "has this expired", rather than having a now and going across alarms and asking "has this one expired?") makes sense. 

However, this "check if alarm has expired" operation should not be on the Ticks type: it's weird to ask if an alarm has expired by calling a method on its reference and passing its dt (`reference.has_expired(dt, now)`). Instead, there should be a struct data type representing an alarm, which has a reference and dt fields, and which has a method `has_expired`. The method takes a value of the same Ticks type as the structure's fields.

PR #2167, opened 2020-10-16
MSP432: Fix bug in interrupt-handling of GPIOs

lebakassemmerl:
### Pull Request Overview
This PR fixes a bug in handling interrupts for GPIOs. I initially thought that every pin except the PIN_J port was capable of detecting interrupts. Obviously this is not the case, only the first 6 ports (equals the first 48 pins) are able to detect interrupts.

Now I renamed all interrupt-capable pins `IntPin` and all the others just `Pin`. And the core change was the interrupt-handler itself, since it mapped the occurred interrupt to the wrong pin.

Also the user gpio-pins were not declared until now.

I detected the bug when I wanted to run the `test/gpio` example in mode 0 which is a necessary test for release 1.6

### Testing Strategy
This PR was tested by running the example `button`, `test/gpio` and `test/print_button`. 

### TODO or Help Wanted
n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/msp_exp432p401r/src/main.rs
+++ b/boards/msp_exp432p401r/src/main.rs
@@ -9,6 +9,7 @@
 #![feature(const_in_array_repeat_expressions)]
 #![deny(missing_docs)]
 
+use components::gpio::GpioComponent;
 use kernel::capabilities;
 use kernel::common::dynamic_deferred_call::DynamicDeferredCall;
 use kernel::common::dynamic_deferred_call::DynamicDeferredCallClientState;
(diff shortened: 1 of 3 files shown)


boards/msp_exp432p401r/src/main.rs
```diff
@@ -131,38 +134,88 @@ pub unsafe fn reset_handler() {
...
+            // 26 => &msp432::gpio::INT_PINS[msp432::gpio::IntPinNr::P03_7 as usize], // debug-gpio
+            // Right outer connector, top to bottom
+            27 => &msp432::gpio::INT_PINS[msp432::gpio::IntPinNr::P02_5 as usize],
+            28 => &msp432::gpio::INT_PINS[msp432::gpio::IntPinNr::P03_0 as usize],
+            29 => &msp432::gpio::INT_PINS[msp432::gpio::IntPinNr::P05_7 as usize],
+            30 => &msp432::gpio::INT_PINS[msp432::gpio::IntPinNr::P01_6 as usize],
+            31 => &msp432::gpio::INT_PINS[msp432::gpio::IntPinNr::P01_7 as usize],
+            // 32 => &msp432::gpio::INT_PINS[msp432::gpio::IntPinNr::P05_0 as usize], // A5
+            // 33 => &msp432::gpio::INT_PINS[msp432::gpio::IntPinNr::P05_2 as usize], // A3
+            34 => &msp432::gpio::INT_PINS[msp432::gpio::IntPinNr::P03_6 as usize]
```
hudson-ayers, line 213: Is there a reason that only the commented out pins have labels indicating which actual pin on the board they correspond to?

hudson-ayers (approved):
As a bug fix I think this looks good.

boards/msp_exp432p401r/src/main.rs
lebakassemmerl: I just wanted to indicate what they are actually used for and why they are not used as standard GPIOs. Should I remove these comments?

hudson-ayers: No, I think the comments are useful. I was just curious that you did not also include comments on the other pins indicating what label they map to on the board.

PR #1300, opened 2019-04-09
Driver Permissions Policy

ssands1:
### Pull Request Overview

This pull request implements a permissions policy which blocks apps from accessing drivers to which they lack explicit access as set in the app's header. If the app lacks the requisite permissions, the board pretends the driver doesn't exist.

The TBF header now includes a new `tipe = 5` which is a series of bytes which form a bitmap with each bit corresponding to one driver, as determined by the order of driver numbers in `capsules/src/driver.rs`.

See companion pull request [elf2tab PR #9](https://github.com/tock/elf2tab/pull/9). Tockloader PR coming soon.

### Testing Strategy

To test this PR alone, I specify which permissions I want for a given app by hard-coding elf2tab to write a specific u64 in all TBF headers. To test different permissions, I currently change the hard-coded value `DEFAULT_PERMS` in [elf2tab/header.rs](https://github.com/tock/elf2tab/pull/9/files), but a future PR to tockloader will allow the user to set specific permissions from the command line (coming soon).

### TODO or Help Wanted

TODO:

- [X] Create [PR for elf2tab](https://github.com/tock/elf2tab/pull/9) to work with this feature
- [ ] Create PR for tockloader to work with this feature
- [ ] Update the relevant files in `/docs`

Help Wanted: 

There are two main downsides that I see so far:

1. tockloader must know the mapping of drivers to bits so when I type `tockloader add-permission my_app LED` (possible in a coming PR), tockloader knows which bit corresponds to the LED. This means another list of drivers to maintain in tockloader.

2. elf2tab must know how many bytes to leave blank as permission bits in each TAB. This means a constant at the top of elf2tab/header.rs that needs to be incremented each time a driver is added.

Any thoughts on a better way to do this would be appreciated.

Also: Thoughts on what drivers an app should have access to by default? All? None? Specific drivers?

### Documentation Updated

- [ ] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [X] Ran `make formatall`.

PR #2172, opened 2020-10-23
TBF Header Permissions

alistair23:
### Pull Request Overview

This PR is a second attempt at my [original `app_id` PR](https://github.com/tock/elf2tab/pull/27).

This is similar to an old PR: https://github.com/tock/tock/pull/1300

This PR adds a permission section to the TBF header. This permission section will contain a list of syscalls that the app is allowed and going to perform.

This has two functions.

1. Only allow apps to interact with approved capsules. There are cases where we only want a single app to use a capsule. Take for example a Root of Trust (RoT) that uses I2C to control other devices. In this case we only want 1 approved app to have that control, while other (possibly malicious) apps are blocked from accessing the hardware.
2. Eventually this can be used to enforce flash access permissions. App A can be granted permission with a unique ID stored in the driver options, only other apps with the same ID can read/write the data stored by app A.

### Security

From a thread model perspective this isn't ideal as we don't expect to always trust the TBF, see: https://github.com/tock/tock/blob/master/doc/threat_model/TBF_Headers.md

Unfortunately without a app manifest I can't think of a way to do this where we don't have to trust the TBF.

Really the only threat we open ourselves up to is somehow a malicious app is loaded that gives itself permissions to access something secure. In the I2C example above, a malicious app could be loaded that has I2C permissions. To mitigate this I have added the exclusive access bit. The exclusive access bit might also be useful for hardware that isn't visualised (such as USB) as that restricts only a single app access to it.

### Testing Strategy

I have tested this by adding support to elf2tab and by auto-generating the list in libtock-rs.

### TODO or Help Wanted

A implementation in Tock after this is approved.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


jrvanwhy:
@gkelly I think this is related to your syscall filtering work.

jrvanwhy:
You are correct to say that this PR is not secure under Tock's threat model.

Unfortunately, I think solving this limitation requires figuring out the app ID story. I don't think we can have a secure mechanism to enforce "only app X can do Y" unless we have a mechanism to verify that an app that claims it is X is actually X.

What we might need is a concept of resource ownership. Who "owns" a particular system call (i.e. determines who can call it)? Is it the kernel? I'm going to try to hash this out more on the core WG call and/or dev mailinglist.

alistair23:
An app ID will have the same problem. Eventually the app ID will be stored in the TBF header as well.

My thinking here is that the loader (which we trust) can check this. For example for Tockloader could take a config file like this:

```
app1: "My test app"
  file: app1.elf
  appid: 0x12345678
  permissions:
    i2c: read, write
    console: write

app2: "My second app"
  file: app2.elf
...
```

Then Tockloader can compare the config file permissions with the individual app permissions to make sure they match. Then it can generate and sign a binary ready to deploy.

I'm not clear how ownership would help. Would an app own a driver? What happens if that app is compromised?

I think it makes sense to have the kernel enforce access. The question is how to tell the kernel to do that. Hard coding a function in Rust in a board specific function is one option, but seems too inflexible. Also how would that function know which app can do what, unless we trust something in the TBF header?

alevy:
Is this PR still reflective of the current state of the discussion?

alistair23:
I think it is

jrvanwhy:
> Is this PR still reflective of the current state of the discussion?

Based on the current state of discussion:

1. It is fine for an application to use this mechanism to sandbox *itself*, e.g. to limit the privileges an adversary would gain of the adversary compromises the app. This is equivalent to Unix daemons that drop privileges after startup.
1. We could make tockloader read this header and tell the user what permissions the app needs before the user decides whether or not to load the app. From the perspective of the threat model, we're making the user running tockloader the arbiter of app permissions, and enforcing those permissions through a combination of tockloader and the mechanism described in this PR.
1. We could add a permissions system to the kernel that contains a kernel-provided permissions list (keyed on application ID). The kernel would verify the list in this PR is consistent with the kernel-provided permissions for the app before loading an app. From the perspective of the threat model, the kernel-provided permissions list would be the arbiter of app permissions and this PR's mechanism would be part of the enforcement implementation.

So yes, it is consistent with the current state of the discussion. It is not a mechanism to isolate a possibly-malicious app *on its own*.

alistair23:
That was my understanding as well.

hudson-ayers:
Is https://github.com/tock/elf2tab/pull/28 up to date with the design here, or are you waiting for approvals on this PR before updating the elf2tab PR?

alistair23:
tock/elf2tab#28 is up to date.

alistair23:
Ping!

hudson-ayers:
I looked through this PR and the corresponding elf2tab PR and think the design looks reasonable. I assume the idea is that this would be implemented by the already existing system call filtering mechanism -- perhaps the docs should mention that somewhere?

doc/TockBinaryFormat.md
```diff
@@ -293,6 +309,35 @@ a non-position-independent binary at an incorrect location.
...
+`TbfHeaderDriverPermission`. The length of the TLV can be used to determine
+the number of array elements. The elements in `TbfHeaderDriverPermission` are
+described below:
+
+`driver_number` is the number of the driver that is allowed. This for example
+could be `0x00000` to indicate that the `Alarm` syscalls are allowed.
+
+`allowed_commands` is a bit mask of the allowed commands. For example a value
+of `0b0001` indicates that only command 0 is allowed. `0b0111` would indicate
+that commands 2, 1 and 0 are all allowed. Subscribe and allow commands are
```
ppannuto, line 338: Should `subscribe` be mass-permitted? One might imagine an information leakage channel coming from events being reported. I believe that in many cases right now, if Application A tells something to start sampling, an Application B can subscribe to the event stream without explicitly starting the sampling itself.

doc/TockBinaryFormat.md
```diff
@@ -293,6 +309,35 @@ a non-position-independent binary at an incorrect location.
...
     `0xFFFFFFFF`.
 
+#### `6` Permissions
+
+The `Permissions` section allows an app to specify driver permissions that it
+is allowed to use. All driver syscalls that an app will use must be listed. The
+list should not include drivers that are not being used by the app.
+
+The data is stored in the optional `TbfHeaderV2Permissions` field. This
+includes an array of all the `perms`.
```
ppannuto, line 319: What is the behavior when things are not specified? The obvious backwards-compatible choice is to default to permitted, but as a security-oriented operating system, is that actually the default we want to put into the world?

---

A perhaps more nuanced version might be that if an application specifies _no_ permissions, then it has all permissions. However, once an application chooses to specify _any_ permissions, we expect it to specify all of its requested permissions?

alistair23: I think by default we would allow all with no permissions.

Eventually I think we could then migrate to a disabled by default once all the usersapce apps/tools have migrated.

doc/TockBinaryFormat.md
alistair23: I thought application B shouldn't get an event callback unless it started the command.

doc/TockBinaryFormat.md
jrvanwhy: Does "disabled by default" mean disabled if the permissions header is not present, or disabled if the permissions header is present but empty?

I would oppose a disable by default if the permissions header is not present for the following reasons:

A. This is not compatible with drivers that use all 32 bits of the command ID (i.e. have command numbers >= 64)
B. Particular projects using Tock that have their own security concerns may need their own permissions header, and that would require them to carry both sets of headers. That could waste a lot of space.

alistair23: Eventually I would say "disabled by default" means both, so access is disabled if there is no permissions header or if it is empty. Like I said though, I think initially we would allow everything and only once everything is working and we have had a release or two for everything to catch up would we make that switch.

In saying that, this PR has no impact on that decision, we could just always default to enable by default and have a kernel config that disables access.

Yep, this PR is limited to less then 64 command numbers. Some drivers will need to have this numbers changed (ninedef) to meet this. As discussed on the call we allow doubling the driver numbers to meet more bit masks.

The idea is this would be the header all apps use. If others want to duplicate this we can't really do anything to help that.

doc/TockBinaryFormat.md
alistair23: Also, I'll clarify this that `allow` and `subscribe` are only permitted if the driver is permitted. If the driver is permitted all `allow` and `subscribe` calls are allowed.

doc/TockBinaryFormat.md
jrvanwhy: > Eventually I would say "disabled by default" means both, so access is disabled if there is no permissions header or if it is empty. Like I said though, I think initially we would allow everything and only once everything is working and we have had a release or two for everything to catch up would we make that switch.
> 
> In saying that, this PR has no impact on that decision, we could just always default to enable by default and have a kernel config that disables access.

If you think that discussion is outside the scope of this PR, then I am fine having it elsewhere.

I do want to point out that OpenTitan has not decided what process permissions model it needs, and we cannot determine now whether OpenTitan will use this permissions model. By default, I'd assume that OpenTitan will have its own permissions system, so we want to support that case.

doc/TockBinaryFormat.md
ppannuto: I think that clarification is good. However, we do have implementations that don't enforce that right now, e.g. for the ADC:

```rust
            // perform callback
            self.appid.map(|id| {
                self.apps
                    .enter(*id, |app, _| {
                        app.callback.map(|callback| {
                            calledback = true;
                            callback.schedule(
                                AdcMode::ContinuousSample as usize,
                                self.channel.get(),
                                sample as usize,
                            );
                        });
                    })
```

On the upcall path, callbacks will be issued to any apps that have subscribed, regardless of which app started the ADC. That may be a bug / unintended consequence in the ADC interface, but it seems sensible that if we're going to filter defensively, we should filter against apps "listening-in" when they didn't declare their intent/ability to do so.

doc/TockBinaryFormat.md
```diff
@@ -293,6 +309,44 @@ a non-position-independent binary at an incorrect location.
...
+could be `0x00000` to indicate that the `Alarm` syscalls are allowed.
+
+`allowed_commands` is a bit mask of the allowed commands. For example a value
+of `0b0001` indicates that only command 0 is allowed. `0b0111` would indicate
+that commands 2, 1 and 0 are all allowed. Subscribe and allow commands are
+always allowed as long as the specific `driver_number` has been specified. If
+a `driver_number` has not been specified for the capsule driver then `allow`
+and `subscribe` will be blocked.
+
+`driver_number` can be repeated. Each time `driver_number` is repeated the
```
phil-levis, line 343: I worry about stateful specifications like this. It places requirements on a parser to maintain that state. 

Instead, and I think this came up in the call, there should be an offset field that specifies the offset of the bits. So you can split this into 2 32-bit values, and offset and a bitmask. The value of a bit in the bitmask is `bit_position` + `offset`. I.e., if bit 7 is set and the offset value is 300, it means command 307 is allowed.

This means that dense command spaces are more efficient to permission, but you can permission arbitrarily sparse ones not too badly (unlike say with this next-64 approach, which would not like the 9dof identifiers).

It would be good to also explicitly specify the layout of the the `TbfHeaderdriverPermission` structure as well.

phil-levis (changes requested):
Following the discussion on the core call, this proposal was generally accepted, with one change, the exact format of a permissions object. In order to support sparse identifier spaces (although perhaps not as efficiently), there should be an offset field. Detailed comments inline.

doc/TockBinaryFormat.md
alistair23: I was hesitant to do it that way as it adds an extra 4 bytes to every driver specified, when I think it will be rare that there will be more then 64 commands (assuming we change the current drivers to all use dense numbering).

I have converted it to using the `offset` field though.

alistair23:
Ping!

alistair23:
Ping!

alistair23:
This must be ready to merge

doc/TockBinaryFormat.md
```diff
@@ -293,6 +310,50 @@ a non-position-independent binary at an incorrect location.
...
++-------------+-------------+---------...--+
+| Type (6)    | Length      | perms        |
++-------------+-------------+---------...--+
+```
+
+The `perms` array is made up of a number of elements of
+`TbfHeaderDriverPermission`. The length of the TLV can be used to determine
+the number of array elements. The elements in `TbfHeaderDriverPermission` are
+described below:
+
```
bradjc, line 333:
```suggestion

```text
Driver Permission Structure:
0             2             4             6             8
+-------------+-------------+---------------------------+
| driver_number             | offset                    |
+-------------+-------------+-------------+-------------+
| allowed_commands                                      |
+-------------------------------------------------------+
```

```

File: doc/TockBinaryFormat.md [hunk 5]
```diff
@@ -293,6 +310,50 @@ a non-position-independent binary at an incorrect location.
...
+`driver_number` is the number of the driver that is allowed. This for example
```
bradjc, line 334:
```suggestion
* `driver_number` is the number of the driver that is allowed. This for example
```

doc/TockBinaryFormat.md
```diff
@@ -293,6 +310,50 @@ a non-position-independent binary at an incorrect location.
...
+could be `0x00000` to indicate that the `Alarm` syscalls are allowed.
+
+`allowed_commands` is a bit mask of the allowed commands. For example a value
```
bradjc, line 337:
```suggestion
* `allowed_commands` is a bit mask of the allowed commands. For example a value
```

doc/TockBinaryFormat.md
```diff
@@ -293,6 +310,50 @@ a non-position-independent binary at an incorrect location.
...
+```
+
+The `perms` array is made up of a number of elements of
+`TbfHeaderDriverPermission`. The length of the TLV can be used to determine
+the number of array elements. The elements in `TbfHeaderDriverPermission` are
+described below:
+
+`driver_number` is the number of the driver that is allowed. This for example
+could be `0x00000` to indicate that the `Alarm` syscalls are allowed.
+
```
bradjc, line 336:
```suggestion

* The `offset` field in `TbfHeaderDriverPermission` indicates the offset of the
`allowed_commands` bitmask. All of the examples described in the paragraph
above assume an `offset` of 0. The `offset` field indicates the start of the
`allowed_commands` bitmask. The `offset` is multiple by 64 (the size of the
`allowed_commands` bitmask). For example an `offset` of 1 and a
`allowed_commands` value of `0b0001` indicates that command 64 is allowed.

```

doc/TockBinaryFormat.md
```diff
@@ -293,6 +310,50 @@ a non-position-independent binary at an incorrect location.
...
+Subscribe and allow commands are always allowed as long as the specific
+`driver_number` has been specified. If a `driver_number` has not been specified
+for the capsule driver then `allow` and `subscribe` will be blocked.
+
+The `offset` field in `TbfHeaderDriverPermission` indicates the offset of the
+`allowed_commands` bitmask. All of the examples described in the paragraph
+above assume an `offset` of 0. The `offset` field indicates the start of the
+`allowed_commands` bitmask. The `offset` is multiple by 64 (the size of the
+`allowed_commands` bitmask). For example an `offset` of 1 and a
+`allowed_commands` value of `0b0001` indicates that command 64 is allowed.
```
bradjc, line 351:
```suggestion
```

doc/TockBinaryFormat.md
```diff
@@ -293,6 +310,50 @@ a non-position-independent binary at an incorrect location.
...
+
+Multiple `offset` and `allowed_commands` can be included, so long as no
```
bradjc, line 353:
```suggestion
Multiple `TbfHeaderDriverPermission` with the same `driver_numer` can be included, so long as no
```

hudson-ayers (approved):
LGTM with Brad's doc updates

hudson-ayers:
cc @phil-levis you need to dismiss your "request changes" for this to be mergeable

alistair23:
Updated and ready to go

hudson-ayers (approved):

ppannuto (approved):

alistair23:
Ready to go!

PR #2173, opened 2020-10-23
riscv: pmp: disallow access above app brk

bradjc:
### Pull Request Overview

This updates the PMP driver to only allow processes access to the process memory below the app break. Before it was allowing access to the entire process memory region.

### Testing Strategy

This pull request was tested by using the mpu test app in libtock-c. I now get a fault on arty-e21 when trying to read from the grant region.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -390,6 +390,13 @@ impl kernel::mpu::MPU for PMP {
             config.unused_region_number()?
         };
 
+        // App memory size is what we actual set the region to. So this region
+        // has to be aligned to 4 bytes.
+        let mut initial_app_memory_size: usize = initial_app_memory_size;
+        if initial_app_memory_size % 4 != 0 {
+            initial_app_memory_size += 4 - (initial_app_memory_size % 4);
+        }
+
         // Make sure there is enough memory for app memory and kernel memory.
         let mut region_size = cmp::max(
             min_memory_size,
@@ -411,7 +418,7 @@ impl kernel::mpu::MPU for PMP {
             return None;
         }
 
-        let region = PMPRegion::new(region_start as *const u8, region_size, permissions);
+        let region = PMPRegion::new(region_start as *const u8, initial_app_memory_size, permissions);
 
         config.regions[region_num] = Some(region);
         config.is_dirty.set(true);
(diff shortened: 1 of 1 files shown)


arch/rv32i/src/pmp.rs
```diff
@@ -411,7 +418,7 @@ impl kernel::mpu::MPU for PMP {
             return None;
         }
 
-        let region = PMPRegion::new(region_start as *const u8, region_size, permissions);
+        let region = PMPRegion::new(region_start as *const u8, initial_app_memory_size, permissions);
```
alistair23, line 421: This doesn't seem right, we should be using `region_size`.

Maybe instead the fix would be:
```diff
diff --git a/arch/rv32i/src/pmp.rs b/arch/rv32i/src/pmp.rs
index bfa682d03..7ab539b99 100644
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -393,7 +393,7 @@ impl kernel::mpu::MPU for PMP {
         // Make sure there is enough memory for app memory and kernel memory.
         let mut region_size = cmp::max(
             min_memory_size,
-            initial_app_memory_size + initial_kernel_memory_size,
+            initial_app_memory_size,
         ) as usize;
 
         // Region size always has to align to 4 bytes
```
Note that the same fix is required in the ARM MPU.

bradjc: The returned `region_size` has to be the entire app's memory, not just the accessible portion. The accessible region is defined by the initial app memory size.

alistair23: That doesn't seem to be what the ARM MPU does though.

bradjc: It uses subregions to restrict access beyond the initial app break, and has to return the entire region size due to power of 2 restrictions.

I tried the same test on hail and it fails as expected.

alistair23: Ok, looks fine then.

alistair23 (approved):

alistair23:
It also seems to pass all of the automated release tests.

PR #2177, opened 2020-10-26
TBF Header Persisent Access Control

alistair23:
### Pull Request Overview

This PR builds on top of: https://github.com/tock/tock/pull/2172 to add a list of write and read permissions for persistent storage.

Apps can have a single write_id to allow them to write data with that id and multiple read IDs to read back the data. This allows the flexibility to read from multiple apps if required.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/doc/TockBinaryFormat.md
+++ b/doc/TockBinaryFormat.md
@@ -15,6 +15,7 @@
     + [`3` Package Name](#3-package-name)
     + [`5` Fixed Addresses](#5-fixed-addresses)
     + [`6` Permissions](#6-permissions)
+    + [`7` Persistent ACL](#7-persistent-acl)
 - [Code](#code)
 
 <!-- tocstop -->
@@ -88,6 +89,7 @@ struct TbfHeader {
     flash_regions: Option<TbfHeaderWriteableFlashRegions>,
     fixed_address: Option<TbfHeaderV2FixedAddresses>,
     permissions: Option<TbfHeaderV2Permissions>,
+    persistent_acl: Option<TbfHeaderV2PersistentAcl>,
 }
 
 // Identifiers for the optional header structs.
@@ -98,6 +100,7 @@ enum TbfHeaderTypes {
     TbfHeaderPicOption1 = 4,
     TbfHeaderFixedAddresses = 5,
     TbfHeaderPermissions = 6,
+    TbfHeaderPersistent = 7,
 }
 
 // Type-length-value header to identify each struct.
@@ -154,6 +157,16 @@ struct TbfHeaderV2Permissions {
     base: TbfHeaderTlv,
     length: u16,
     perms: [TbfHeaderDriverPermission],
+}
+
+// A list of persistent access permissions
+struct TbfHeaderV2PersistentAcl {
+    base: TbfHeaderTlv,
+    write_id: u32,
+    read_length: u16,
+    read_ids: [u32],
+    access_length: u16,
+    access_ids: [u32],
 }
 ```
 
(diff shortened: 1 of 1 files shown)


doc/TockBinaryFormat.md
```diff
@@ -293,6 +319,55 @@ a non-position-independent binary at an incorrect location.
...
+0             2             4             6             8
++-------------+-------------+---------------------------+---------...--+
+| Type (6)    | Length      | write_id                  | read_ids     |
++-------------+-------------+-------------+-------------+---------...--+
+```
+
+`write_id` indicates the id that all persistent data is written with. Only
+apps with the same id listed in the `read_ids` can read the data.
+
+`read_ids` list all of the ids that this app has permission to read
```
hudson-ayers, line 369: It's not immediately clear to me from this documentation how these IDs map to ability to store stuff.

Is the intended design that any app with a `write_id` can write wherever it wants, and once it writes somewhere that region is reserved for writing by this particular app?

Is there some out-of-band mapping of `write_id`s to regions of flash?

alistair23: My original idea is that this would not tie to regions but instead to K/V pairs.
For example app A uses ID `0x12345678` to write it's K/V data. App B has a read ID of `0x12345678` so it can read the same pairs. This way the flash memory isn't partitioned up.

This can avoid fragmentation wasting space, the obvious downside is the intentional of DoS attacks by filling up flash.

The other implementation option is that the flash region could be partitioned with each one having a specific ID.
For example the flash could be partitioned with two regions in `reset_handler()`. One with ID `0x12345678` so all apss with a `write_id` or `0x12345678` can write and `read_id` of `0x12345678` can read and a second region with a different ID.

The hope is that this is flexible enough to allow different implementations in capsules depending on the use case.

hudson-ayers: The K/V approach suggests only a single writer is possible for a given "region" (region == a given K/V combo) while the ID'ed partitions approach suggests multiple writers of a given region might be possible. I think it is important to clarify whether multiple apps can have the same `write_id`, as that is probably something the loader would need to enforce.

Otherwise I understand the desire to leave this general for now. I think it would be possible for a given implementation to still prevent DoS attacks by setting a per-app limit on the amount that can be stored (something like allow each app to write up to `TOTAL_SPACE / NUM_UNIQUE_WRITE_IDS`)

alistair23: The idea was that multiple apps can have the same `write_id`. They could then write over each others data, but that would be on purpose.

hudson-ayers: I think it would be good to state that in the document.

Also, given that, one could imagine a use case for a single app having multiple write_ids.
Example: each app logs errors in its own write_region. An error-reporting app reports these errors over the network, and once the reported errors are acked erases them from the log. You wouldn't want a single shared write_region for this because one app with a lot of errors could fill it up quickly. But you would want your error reporting app to be able to modify the write_region of every other app.

Thoughts? If this was desired, it might make more sense to just have a list of ids, with two bits of the ID reserved to indicate whether the ID is read/write/both

alistair23: The problem with multiple `write_ids` is which one is used to write data? Is it just the first one?

That does seem like an good use case though

hudson-ayers: That is a good point also. I think that is going to be part of the challenge of having a single tbf header that is intended to serve pretty different use cases

alistair23:
Ping!

doc/TockBinaryFormat.md
ppannuto: Following this thinking, we have a pretty generous address space for TBF types (effectively 2^15). As there is an effort to keep this general, maybe the right approach is to have a `read_ids` type and a `write_ids` type, each of which allow any number of IDs?

I think this would have the added benefit of more cleanly allowing for read-only apps. Right now, the `write_id` field is required, but there are likely apps that should never write. We could use some sentinel id like 0, but if there simply were no `write_ids` that would be a bit cleaner perhaps.

alistair23: The problem with multiple write_ids is which one is used to write data? Is it just the first one or do we allow apps to specify the `write_id`?

If we make write_id optional then we need something to indicate if it's included or not, which doesn't really buy us anything. I have updated this to include that a `write_id` of `0` means no write.

alistair23: Ok, I have addedd `access_ids` which should handle the case that @hudson-ayers metnioned

ppannuto: > The problem with multiple write_ids is which one is used to write data? Is it just the first one or do we allow apps to specify the write_id?

I think that might actually make a lot of sense for apps to specify the id they want to use when writing. I could imagine that you want to have some app available to tertiary app A, and some available to tertiary app B. The writer can use multiple `write_id`s to control which apps will be able to read which data.

ppannuto: Ah, I re-read the `access_ids` and I think it actually handles the use cases I brought up here.

ppannuto (approved):
@tock/core-wg Following the conversation today of settling PRs, this one has been sitting for a while and I think should be good to go pre 2.0 freeze.

alistair23:
Whoop! Ready to go

hudson-ayers (approved):

PR #2178, opened 2020-10-27
Synchronous panic output over USB for nano33ble

### LINKED ISSUE #1972: Debugging over USB serial
For platforms with only USB virtual consoles, we need to support debugging over that interface.

On a panic, we traditionally use the synchronous UART to print debugging information. However, this synchronous method seems very hard to impossible to implement for USB CDC. Crashes in applications should still be debuggable since the kernel is trustworthy, but would require some changes in the kernel for how they are handled. Crashes in the kernel are a harder question.

hudson-ayers:
### Pull Request Overview

This pull request fixes #1972 by implementing a "synchronous" panic output via USB CDC for the nano33ble. It does so by spin-waiting for USB interrupts to allow the USB and CDC state machines to operate as intended despite the fact that the core kernel is no longer running and dispatching interrupts. It also requires some creative use of `static mut` global variables to give the panic handler access to the resources it needs.

Of course, this approach will not provide panic output if the USB connection is not yet established. Still it is a substantial improvement over the current state of affairs which requires using a separate serial port adapter to get panic output off the device.

### Testing Strategy

This pull request was tested by manually inserting a panic in the LED capsule, and observing the panic output using `tockloader listen` when running the blink app.

### TODO or Help Wanted

N/A

### Documentation Updated

- [x] Updated the relevant files in `/docs`

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nano33ble/src/main.rs
+++ b/boards/nano33ble/src/main.rs
@@ -71,6 +71,7 @@ const NUM_PROCS: usize = 8;
 static mut PROCESSES: [Option<&'static dyn kernel::procs::ProcessType>; NUM_PROCS] = [None; 8];
 
 static mut CHIP: Option<&'static nrf52840::chip::Chip> = None;
+static mut CDC_REF: Option<&'static capsules::usb::cdc::CdcAcm<'static, nrf52::usbd::Usbd>> = None;
 
 /// Dummy buffer that causes the linker to reserve enough space for the stack.
 #[no_mangle]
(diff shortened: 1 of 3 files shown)


boards/nano33ble/src/main.rs
```diff
@@ -71,6 +71,7 @@ const NUM_PROCS: usize = 8;
 static mut PROCESSES: [Option<&'static dyn kernel::procs::ProcessType>; NUM_PROCS] = [None; 8];
 
 static mut CHIP: Option<&'static nrf52840::chip::Chip> = None;
+static mut CDC_REF: Option<&'static capsules::usb::cdc::CdcAcm<'static, nrf52::usbd::Usbd>> = None;
```
ppannuto, line 74:
```suggestion
static mut CDC_REF_FOR_PANIC: Option<&'static capsules::usb::cdc::CdcAcm<'static, nrf52::usbd::Usbd>> = None;
```

ppannuto (approved):
This is actually pretty clean all things considered, nicely done.

ppannuto (approved):

bradjc (approved):

PR #1031, opened 2018-06-26
Use LLVM Tools

bradjc:
### Pull Request Overview

At long last, this pull request removes the need for arm-none-eabi-gcc and instead uses the llvm tools (size, objcopy, and objdump).

This requires the nightly upgrade and the switch to lld, so it really needs to be viewed with just individual commits.

### Testing Strategy

This pull request was tested by programming the kernel on hail.

### TODO or Help Wanted

n/a

### Documentation Updated

- [ ] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [ ] Ran `make formatall`.

PR #2180, opened 2020-10-28
dont let linker assume 512 byte page size

hudson-ayers:
### Pull Request Overview

Now that we support non-SAM4L chips, I don't think it is safe for the linker to assume a maximum page size of 512 bytes. While https://github.com/tock/tock/pull/2090/ added a variable to the layout file for each board to specify its page size, I am not sure how to access that variable, so this PR just removes the linker option. However if someone better at Makefiles than me does know an easy way to access that, we should just use it instead of 512.

### Testing Strategy

CI

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/Makefile.common
+++ b/boards/Makefile.common
@@ -24,15 +24,11 @@ TARGET_DIRECTORY ?= $(TOCK_ROOT_DIRECTORY)target/
 
 # RUSTC_FLAGS allows boards to define board-specific options.
 # This will hopefully move into Cargo.toml (or Cargo.toml.local) eventually.
-# lld uses the page size to align program sections. It defaults to 4096 and this
-# puts a gap between before the .relocate section. `zmax-page-size=512` tells
-# lld the actual page size so it doesn't have to be conservative.
 RUSTC_FLAGS ?= \
   -C link-arg=-Tlayout.ld \
   -C linker=rust-lld \
   -C linker-flavor=ld.lld \
   -C relocation-model=dynamic-no-pic \
-  -C link-arg=-zmax-page-size=512 \
   -C link-arg=-icf=all \
 
 # RISC-V-specific flags.


ppannuto:
Interesting. From a quick search, I think we can actually go the other way around -- the linker script can access the command line argument: https://stackoverflow.com/questions/33540989/is-it-possible-to-change-the-value-of-constantmaxpagesize-in-a-linker-script

We could have both and have them in sync then if it's a Make variable set in each board's makefile, used as an argument to the linker option, and via the `CONSTANT` mechanism in the linker script.

_However_, not doing all that convoluted stuff seems like a win if we can. Why was this originally passed to the linker, does it need it for anything?

bradjc:
I'm pretty sure I added this after a prolonged battle with the linker and llvm. Are you sure it is safe to remove this?

hudson-ayers:
[Here](https://github.com/tock/tock/commit/e8b50080f7ed994ade5d3f20d589619ba8ff2cc6) is the original commit that added it. From reading some comments on various LLVM posts like [this one](https://reviews.llvm.org/D56205), it seems that by default lld aligns all sections to MAXPAGESIZE. 

The comment above the flag seems to indicate that is why it was used:
```
# lld uses the page size to align program sections. It defaults to 4096 and this
# puts a gap between before the .relocate section. `zmax-page-size=512` tells
# lld the actual page size so it doesn't have to be conservative.
```

I *think* this means that removing it will waste up to 4KB of flash space on platforms like the SAM4L with page sizes much smaller than 4kB. But perhaps there is actually an issue with having a gap between these sections beyond wasted flash?

Actually, I just went back to the PR that originally added this, and it seems there was some discussion between you and Brad: https://github.com/tock/tock/pull/1031#discussion_r198653012 . Unfortunately the link in Brad's comment no longer seems to link to an actual line of code, so I am not sure what it was supposed to match..

hudson-ayers:
> I'm pretty sure I added this after a prolonged battle with the linker and llvm. Are you sure it is safe to remove this?

My most recent comment was a race with this one, but after looking back at the PR that added this I am not sure it is safe to remove it...I don't really understand why it was added though, or if it is potentially problematic that we don't use the PAGE_SIZE for the board being built

bradjc:
Thank you for digging up the old thread. It's entirely possible that over time and with changes to newer llvm versions whatever caused this flag to be needed on llvm but not gcc has gone away. The flag was the best patch I could come up with for the differences between gcc and llvm at the time, not something to enforce a specific page size or anything like that.

I'm pretty sure that if boards still work without this flag then it is safe to remove it.

hudson-ayers:
Here is a relevant discussion from the rust embedded wg: https://github.com/rust-embedded/cortex-m-quickstart/pull/95#issuecomment-691711983 . It seems that we might just want to be telling the linker not to align sections on page boundaries at all, using the -nmagic flag or similar. 

...But also I'm starting to feel like I'm in over my head on this and I should just leave well enough alone, haha

lenary:
I recall that the max page size has some implicit behaviour that is only mentioned in passing in the ld documentation, I'm going to see if I can find it. There's a little about this in Section 4.4 of "Linkers and Loaders", and more in Section 4.7. This should affect the "segments" (program headers), and implicitly the data segment starts in one of two places compared to the end of the text section:
- at the next page boundary (NMAGIC)
- exactly one page later, so you can map that physical page twice into two different virtual pages at different addresses. 

I *think* if you're controlling the specific addresses of these sections in your linker script, this won't be an issue - but if you're only setting the alignment, your linker may round that up to max page size in certain circumstances, especially when allocating sections into segments - this is done implicitly if you don't have program headers in the linker script, which kernel_layout.ld does not seem to have. 

This is probably the "extra space" that the comment refers to around the `.relocate` section. - that section may have different permissions to the sections that came before, so may have been over-aligned for the original device.

hudson-ayers:
I am not confident that the change proposed in this PR will always work, and also not going to look into a more correct fix in the near future, so I am moving this to an issue: #2219.

PR #2181, opened 2020-10-30
Virtual log

kevtan:
### Pull Request Overview

This pull request implements the virtualization of the log interface, which makes it possible for multiple in-kernel capsules to write persistent logs to flash. It also fixes a minor error in the original `Log::seek` method where the log device would not check to see it was in an idle state before it initiated a seek operation.

### Testing Strategy

@hudson-ayers @TPackard @phil-levis 
(I'm tagging you guys because I'm not sure who to tag)

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/capsules/src/log.rs
+++ b/capsules/src/log.rs
@@ -627,21 +627,21 @@ impl<'a, F: Flash + 'static> LogRead<'a> for Log<'a, F> {
         &self,
         buffer: &'static mut [u8],
         length: usize,
-    ) -> Result<(), (ReturnCode, Option<&'static mut [u8]>)> {
+    ) -> Result<(), (ReturnCode, &'static mut [u8])> {
         // Check for failure cases.
         if self.state.get() != State::Idle {
             // Log busy, try reading again later.
-            return Err((ReturnCode::EBUSY, Some(buffer)));
+            return Err((ReturnCode::EBUSY, buffer));
         } else if buffer.len() < length {
             // Client buffer too small for provided length.
-            return Err((ReturnCode::EINVAL, Some(buffer)));
+            return Err((ReturnCode::EINVAL, buffer));
         } else if self.read_entry_id.get() > self.append_entry_id.get() {
             // Read entry ID beyond append entry ID, must be invalid.
             self.read_entry_id.set(self.oldest_entry_id.get());
-            return Err((ReturnCode::ECANCEL, Some(buffer)));
+            return Err((ReturnCode::ECANCEL, buffer));
         } else if self.read_client.is_none() {
             // No client for callback.
-            return Err((ReturnCode::ERESERVE, Some(buffer)));
+            return Err((ReturnCode::ERESERVE, buffer));
         }
 
         // Try reading next entry.
@@ -654,7 +654,7 @@ impl<'a, F: Flash + 'static> LogRead<'a> for Log<'a, F> {
                 self.deferred_client_callback();
                 Ok(())
             }
-            Err(return_code) => Err((return_code, Some(buffer))),
+            Err(return_code) => Err((return_code, buffer)),
         }
     }
 
--- a/capsules/src/log.rs
+++ b/capsules/src/log.rs
@@ -802,12 +802,14 @@ impl<'a, F: Flash + 'static> LogWrite<'a> for Log<'a, F> {
     }
 
     /// Erase the entire log.
+    ///
     /// ReturnCodes used:
-    ///     * SUCCESS: flush started successfully.
-    ///     * EBUSY: log busy, try again later.
+    /// * SUCCESS: flush started successfully.
+    /// * EBUSY: log busy, try again later.
+    ///
     /// ReturnCodes used in erase_done callback:
-    ///     * SUCCESS: erase succeeded.
-    ///     * EBUSY: erase interrupted by busy flash driver. Call erase again to resume.
+    /// * SUCCESS: erase succeeded.
+    /// * EBUSY: erase interrupted by busy flash driver. Call erase again to resume.
     fn erase(&self) -> ReturnCode {
         if self.state.get() != State::Idle {
             // Log busy, try appending again later.
--- a/capsules/src/log.rs
+++ b/capsules/src/log.rs
@@ -10,7 +10,7 @@
 //! after an entry with a smaller ID. IDs can also be used to determine the physical position of
 //! entries within the log's underlying storage volume - taking the ID modulo the size of the
 //! underlying storage volume yields the position of the entry's header relative to the start of
-//! the volume. Entries should not be created manually by clients, only retrieved through the
+//! the volume. Entry IDs should not be created manually by clients, only retrieved through the
 //! `log_start()`, `log_end()`, and `next_read_entry_id()` functions.
 //!
 //! Entry IDs are not explicitly stored in the log. Instead, each page of the log contains a header
--- a/capsules/src/log.rs
+++ b/capsules/src/log.rs
@@ -73,6 +73,7 @@
 //! ```
 
 use core::cell::Cell;
+use core::cmp;
 use core::convert::TryFrom;
 use core::mem::size_of;
 use core::unreachable;
--- a/capsules/src/log.rs
+++ b/capsules/src/log.rs
@@ -40,32 +40,36 @@
 //! Usage
 //! -----
 //!
-//! ```
-//!     storage_volume!(VOLUME, 2);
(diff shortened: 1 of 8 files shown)


lschuermann:
Thanks for the contribution. Unfortunately the emojis in the commit messages are not displayed correctly in my tools and break my autocomplete workflow. Would you mind rewording those commits? I'm not sure whether we're at a stage of adoption where this won't break most setups. :confused:

ppannuto:
@lschuermann I am kinda saddened to hear that there are still tools that break on emoji -- doesn't really bode well for their i18n support.. I quite like the quick visual record the icons provide here. Out of curiosity, what breaks for you?

lschuermann:
> Out of curiosity, what breaks for you?

Essentially, I see only unicode boxes except one emoji, indicating lacking support in my fonts. One terminal emulator cuts up the emoji as they would be too wide for the otherwise monospace font. In the TTY (which I find myself working in if my GPU is busy serving a VM) I see diamond characters. My script to checkout a commit based on the first characters of a commit message doesn't work (as I can't type emoji). Disclaimer: personal opinion. I don't think that this adds any value, but much rather distracts from the commit message - being a precise description of what was done - and _if_ one can decipher the emoji leading to thinking ["it's just a simple fix for a non critical issue"](https://github.com/carloscuesta/gitmoji/issues/605) or the likes.

bradjc:
Virtualizing log read is fairly complicated just because there are multiple related calls, and allowing those to be interleaved in any order is probably not what a user of the log read interface expects. Is virtualizing read necessary? Would it be simpler to just virtualize write and assume there will only be one reader?

alistair23:
I'm with @lschuermann on the emojis, it's distracting, doesn't really add anything and can easily break tools. Why does one of the commits have a poop in front of it? What does that mean?

It's good to see someone splitting up their commits nicely though :)

kevtan:
@lschuermann @alistair23 Sorry to hear that the emojis were distracting and breaking tools! I've reworded the commit messages accordingly. I've been following [this](https://gitmoji.carloscuesta.me/) commit message emoji standard in case either of you was curious, but I'll stick with regular ASCII characters for Tock.

kevtan:
@bradjc Before making a final decision about whether or not read should be virtualized, can we talk about who will actually need to read the log? It's unclear to me what the use case of the log actually is, considering it's not used anywhere in our codebase except for a test harness. It seems reasonable to me that kernel capsules only ever need to write info/error/warning messages and never really need to read them. But who _should_ have read functionality?

phil-levis:
If there's a log, why shouldn't you be able to virtualize it? You don't have to. But multiple capsules in the kernel might want to be able to read the log, and it would be a problem if you had to choose between them. E.g., you might want ProcessConsole to be able to print the log for you, but also have a network service that lets you request it over TCP/IP.

Virtualizing should be pretty simple: you just need to virtualize the seek pointer. This is what all POSIX file systems do with the open file table; you can open a file multiple times, and each entry in the open file table has its own seek pointer.

capsules/src/log.rs
```diff
@@ -666,7 +666,10 @@ impl<'a, F: Flash + 'static> LogRead<'a> for Log<'a, F> {
     ///
     /// ReturnCodes used:
     /// * SUCCESS: seek succeeded.
+<<<<<<< HEAD
     /// * EBUSY: log busy with another operation, try again later.
+=======
+>>>>>>> 7598e876f... 💡 Make "append" documentation render correctly
```
alistair23: You have some conflict markers left over.

capsules/src/virtual_log.rs
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+}
+
+impl<'a, Log: LogRead<'a> + LogWrite<'a>> LogRead<'a> for VirtualLogDevice<'a, Log> {
+    type EntryID = usize;
+
+    // This method is used by a capsule to register itself as a read client of the virtual log device.
+    fn set_read_client(&'a self, read_client: &'a dyn LogReadClient) {
+        // TODO: Should we check if we're already part of the mux's devices list?
```
alistair23, line 62: There probably should be a check.

capsules/src/virtual_log.rs
```diff
@@ -0,0 +1,171 @@
...
+    fn next_read_entry_id(&self) -> Self::EntryID {
+        self.mux.log.next_read_entry_id()
+    }
+
+    // The seek function on the virtual log device doesn't actually cause a seek to occur on the
+    // underlying persistent storage device. All it does is update a state variable representing
+    // the location of its position in the log file.
+    // TODO: check for errors
+    fn seek(&self, entry: Self::EntryID) -> ReturnCode {
+        self.read_entry_id.set(entry);
```
alistair23, line 96: I know there is a TODO, but can we return errors?

capsules/src/virtual_log.rs
```diff
@@ -18,11 +18,11 @@ enum Op {
 }
 
 pub struct VirtualLogDevice<'a, Log: LogRead<'a> + LogWrite<'a>> {
-    // Every virtual log device has a reference to the mux.
+    // A reference to the mux
     mux: &'a MuxLog<'a, Log>,
-    // Every virtual log device has a pointer to the next virtual log device.
+    // A pointer to the next virtual log device
```
alistair23: These changes should probably be squashed

capsules/src/virtual_log.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    fn read(
+        &self,
+        buffer: &'static mut [u8],
+        length: usize,
+    ) -> Result<(), (ReturnCode, Option<&'static mut [u8]>)> {
+        self.operation.set(Op::Read(buffer, length));
+        self.mux.do_next_op();
+        Ok(()))
```
alistair23: There are too many brackets here

capsules/src/virtual_log.rs
```diff
@@ -0,0 +1,171 @@
...
+use kernel::ReturnCode;
+
+// The purpose of this type alias is to make it clear when a usize represents a log entry ID.
+type EntryID = usize;
+
+// This enum represents the current operation that a virtual log device is performing.
+#[derive(Copy, PartialEq)]
+enum Op {
+    Idle,
+    Read(&'static mut [u8], usize),
```
alistair23: Does storing the buffer in the state really work?

alistair23 (changes requested):

bradjc:
Regarding read: the readers that stand out to me are some external tool (like tockloader) that reads the log via reading flash directly (i.e. not involving kernel code at all), or a capsule that wirelessly reports the log. In the first case the kernel does not need to be able to read the log at all, and in the second case there would only be one reader.

I agree virtualizing the read side is entirely possible, but if we merge virtual_log.rs with read virtualization it needs to actually work. One option is to implement it, which would be great. Another option is to only virtualize write and ignore read, which I also think would be fine for many use cases.

capsules/src/virtual_log.rs
```diff
@@ -0,0 +1,281 @@
...
+    fn next_read_entry_id(&self) -> Self::EntryID {
+        self.mux.log.next_read_entry_id()
+    }
+
+    // The seek function on the virtual log device doesn't actually cause a seek to occur on the
+    // underlying persistent storage device. All it does is update a state variable representing
+    // the location of its position in the log file.
+    // TODO: check for errors
+    fn seek(&self, entry: Self::EntryID) -> ReturnCode {
+        self.read_entry_id.set(entry);
```
bradjc, line 92: Need to actually use `self.read_entry_id`.

bradjc (changes requested):

phil-levis:
Could we clear the changes requested if they should be cleared, then we can look into review for approval?

capsules/src/virtual_log.rs
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    fn log_end(&self) -> Self::EntryID {
+        self.mux.log.log_end()
+    }
+
+    // TODO: this needs to be virtualized
+    fn next_read_entry_id(&self) -> Self::EntryID {
+        self.mux.log.next_read_entry_id()
+    }
```
bradjc, line 108: Need to take care of this TODO?

bradjc:
I think this is close. There is a TODO in the code that mentions virtualization, and it needs to be switched from draft to full PR.

alevy:
What's the state of this PR @kevtan?

phil-levis:
Looks like this won't get wrapped up.

PR #2182, opened 2020-10-30
Stm32f412g: Add rng support

krady21:
### Pull Request Overview

This pull request adds true random number generation support for the stm32f412gdiscovery board. 
I had to tweak the Stm32f4xx struct a little bit, making it generic over an interrupt service. This makes it possible to access interrupts that are specific to supersets of the stm32f4xx chip (stm32f412 in this case). 

### Testing Strategy

This pull request was tested using a stm32f412gdisco board by running the rng test from libtock-c.

### TODO or Help Wanted

N/O

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/stm32f412gdiscovery/src/main.rs
+++ b/boards/stm32f412gdiscovery/src/main.rs
@@ -10,6 +10,7 @@
 #![feature(const_in_array_repeat_expressions)]
 use capsules::virtual_alarm::VirtualMuxAlarm;
 use components::gpio::GpioComponent;
+use components::rng::RngComponent;
 use kernel::capabilities;
 use kernel::common::dynamic_deferred_call::{DynamicDeferredCall, DynamicDeferredCallClientState};
 use kernel::component::Component;
@@ -59,6 +60,7 @@ struct STM32F412GDiscovery {
     touch: &'static capsules::touch::Touch<'static>,
     screen: &'static capsules::screen::Screen<'static>,
     temperature: &'static capsules::temperature::TemperatureSensor<'static>,
+    rng: &'static capsules::rng::RngDriver<'static>,
 }
 
 /// Mapping of integer syscalls to objects that implement syscalls.
@@ -79,6 +81,7 @@ impl Platform for STM32F412GDiscovery {
             capsules::touch::DRIVER_NUM => f(Some(self.touch)),
             capsules::screen::DRIVER_NUM => f(Some(self.screen)),
             capsules::temperature::DRIVER_NUM => f(Some(self.temperature)),
+            capsules::rng::DRIVER_NUM => f(Some(self.rng)),
             _ => f(None),
         }
     }
@@ -335,7 +338,11 @@ unsafe fn set_pin_primary_functions(
 }
 
 /// Helper function for miscellaneous peripheral functions
-unsafe fn setup_peripherals(tim2: &stm32f412g::tim2::Tim2, fsmc: &stm32f412g::fsmc::Fsmc) {
+unsafe fn setup_peripherals(
+    tim2: &stm32f412g::tim2::Tim2,
+    fsmc: &stm32f412g::fsmc::Fsmc,
+    trng: &stm32f412g::trng::Trng,
+) {
     // USART2 IRQn is 38
     cortexm4::nvic::Nvic::new(stm32f412g::nvic::USART2).enable();
 
@@ -346,6 +353,8 @@ unsafe fn setup_peripherals(tim2: &stm32f412g::tim2::Tim2, fsmc: &stm32f412g::fs
 
     // FSMC
     fsmc.enable();
+
+    trng.enable_clock();
 }
 
 /// Reset Handler.
@@ -372,7 +381,11 @@ pub unsafe fn reset_handler() {
     );
     peripherals.init();
     let base_peripherals = &peripherals.stm32f4;
-    setup_peripherals(&base_peripherals.tim2, &base_peripherals.fsmc);
+    setup_peripherals(
+        &base_peripherals.tim2,
+        &base_peripherals.fsmc,
+        &peripherals.trng,
+    );
 
     // We use the default HSI 16Mhz clock
     set_pin_primary_functions(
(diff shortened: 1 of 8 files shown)


chips/stm32f4xx/src/trng.rs
```diff
@@ -0,0 +1,154 @@
...
+    pub fn disable_clock(&self) {
+        self.clock.disable();
+    }
+
+    pub fn handle_interrupt(&self) {
+        if self.registers.sr.is_set(Status::SEIS) {
+            self.registers.sr.modify(Status::SEIS::CLEAR);
+            // Restart the rng
+            self.registers.cr.modify(Control::RNGEN::CLEAR);
+            self.registers.cr.modify(Control::RNGEN::SET);
```
hudson-ayers, line 86: It makes sense that if there is a seed error, we restart the RNG. But shouldn't we inform the client of the error, as the client is probably blocked waiting for a callback that now will never arrive?

krady21: Restarting the rng process is the suggested recovery procedure in the reference manual. After restart, the peripheral is supposed to continue generating correct random numbers and the corresponding interrupts, so I don't see the client getting stuck waiting.

krady21: Although, It seems that I have omitted to also read the data register before restarting in order to avoid having the client read invalid data.

hudson-ayers: that makes sense, thanks

chips/stm32f4xx/src/chip.rs
```diff
@@ -2,39 +2,32 @@
...
             mpu: cortexm4::mpu::MPU::new(),
             userspace_kernel_boundary: cortexm4::syscall::SysCall::new(),
             scheduler_timer: cortexm4::systick::SysTick::new(),
+            interrupt_service,
         }
     }
 }
 
-impl Chip for Stm32f4xx {
+impl<I: InterruptService> Chip for Stm32f4xx<I> {
```
alistair23, line 30: Shouldn't this be a separate PR?

chips/stm32f412g/src/chip.rs
```diff
@@ -0,0 +1,8 @@
+use crate::interrupt::Stm32f412gInterruptService;
+use stm32f4xx::chip::Stm32f4xx;
+
+pub type Chip = Stm32f4xx<Stm32f412gInterruptService>;
+
+pub unsafe fn new() -> Chip {
+    Stm32f4xx::new(Stm32f412gInterruptService::new())
+}
```
alistair23, line 8: This seems unrelated to RNG support

boards/stm32f412gdiscovery/src/main.rs
```diff
@@ -139,8 +142,7 @@ unsafe fn set_pin_primary_functions() {
     });
     PinId::PA03.get_pin().as_ref().map(|pin| {
         pin.set_mode(Mode::AlternateFunctionMode);
-        // AF7 is USART2_RX
-        pin.set_alternate_function(AlternateFunction::AF7);
+        // AF7 is USART2_RX pin.set_alternate_function(AlternateFunction::AF7);
```
alistair23, line 145: Why is this commented out?

alistair23 (changes requested):

chips/stm32f4xx/src/chip.rs
krady21: I was not sure about this. Before this pr, from the service_pending_interrupts() function in stm32f4xx/chip.rs one could only process interrupts that were common to to all stm324xx chips. Because the random number generator stm32f412g specific, in order to use it and receive the associated interrupts, one needs to have access to the nvic id defined in the stm32f412g_nvic.rs file. It would not have made sense for the stm32f412g crate to be imported in the stm32f4xx crate, as that would lead to circular dependencies. By making Chip generic over InterruptService, this allows each variant of the f4 boards to define their own Chip struct where they can process interrupts that are not common to [f4xx](https://github.com/krady21/tock/blob/aafad112e09c41e0f424efca37a03a6bbc2817a0/chips/stm32f412g/src/interrupt.rs#L18-L21). This is not new in tock. Nordic boards do the [same thing](https://github.com/krady21/tock/blob/stm32f4-rng/chips/nrf52840/src/interrupt_service.rs). If needed, I can break this up into two PRs. Let me know.

boards/stm32f412gdiscovery/src/main.rs
krady21: It seems that I've messed up this line after solving some merge conflicts. There was no reason for it to be commented. Thanks for spotting it!

chips/stm32f4xx/src/chip.rs
alistair23: It should probably be a separate PR I think.

hudson-ayers: Yeah, its clearly a desired change but I think that it is kind of obscured by being in what appears to just be a PR for adding a single hardware peripheral to a single chip.

Also, now that #2069 is merged, the `InterruptService` trait is now available in `kernel/src/platform` and has changed slightly. It would be good to use the updated, broadly usable trait rather than an stm specific one.

krady21: Didn't know about #2069 , so thanks for pointing it out. @hudson-ayers , I see that you also have an ongoing PR (#2187) that implements the interrupt changes for the f3. Do you have another one in the making for the f4 chips and boards? I'm asking in order to avoid overlapping with your work.

hudson-ayers: I do have another PR in the works for the f4 chips and boards, it should be ready within a day or two

krady21:
I did the rebase, but the rng is not working anymore. I'll have to look more into it.

hudson-ayers:
> I did the rebase, but the rng is not working anymore (or any other libtock example for that matter). I'll have to look more into it.

Oh dear. Perhaps some initialization was messed up in #2188 . Please let me know what you find, I do not have hardware to test on.

krady21:
Tested this with the changes added in #2223 and it works now.

hudson-ayers (approved):

bradjc:
@alistair23

alistair23 (approved):

PR #2183, opened 2020-10-30
Core WG notes for October 30

brghena:
### Pull Request Overview

Notes for Core Working Group meeting October 30th.

### Testing Strategy

Reading it with spellcheck on.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [N/A] Ran `make prepush`.

### DIFF
--- a/doc/wg/core/notes/core-notes-2020-10-30.md
+++ b/doc/wg/core/notes/core-notes-2020-10-30.md
@@ -93,7 +93,7 @@
  * Amit: We can add more path java-style to make them more unique. Basically I was saying that more than 32-bits is really helpful to having UUIDs.
 * Brad: I'm a little worried about "testapp" because those are exactly the users that we want this to "just work" for. People assigning keys are already gonna figure this out.
 * Amit: So those cases won't set anything in the TLV and the board will just automatically create one.
-* Johnathan: Could even have elf2tbf enforce that two things named the same always map to same ID. Could be something complex internally, but with a simple rule.
+* Johnathan: Could have elf2tab default to generating an app ID by hashing the package name. This would have the effect of making binaries with the same package name map to the same app ID. Our documentation could then describe that different binaries should have different names. Would be complex internally, but with a simple rule for users.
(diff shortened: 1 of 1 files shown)


doc/wg/core/notes/core-notes-2020-10-30.md
```diff
@@ -0,0 +1,101 @@
...
+ * Johnathan: I had kind of assumed originally that it would be super hard to do generically. But I think the language has improved which will make it better.
+ * Leon: I have a mostly-worked-through example that will improve this.
+ * Amit: Plus in the common case, the 48-byte IDs would live mostly in flash. Plus right now we use arbitrary strings, which are reasonably long anyways.
+ * Brad: Why is 48-bytes better for unique IDs? Won't many people just call themselves "test" or something that all hash the same?
+ * Amit: We can add more path java-style to make them more unique. Basically I was saying that more than 32-bits is really helpful to having UUIDs.
+* Brad: I'm a little worried about "testapp" because those are exactly the users that we want this to "just work" for. People assigning keys are already gonna figure this out.
+* Amit: So those cases won't set anything in the TLV and the board will just automatically create one.
+* Johnathan: Could even have elf2tbf enforce that two things named the same always map to same ID. Could be something complex internally, but with a simple rule.
```
jrvanwhy, line 96: I said a bit more that was missed in the notes. I believe the following is an accurate paraphrasing of what I said:

Could have elf2tab default to generating an app ID by hashing the package name. This would have the effect of making binaries with the same package name map to the same app ID. Our documentation could then describe that different binaries should have different names. Would be complex internally, but with a simple rule for users.

brghena: Updated. Thanks for checking this Johnathan! There's always a chance I miss something important when a paraphrase.

jrvanwhy: Thanks for changing it! I'm being a bit nitpicky on the topic of app IDs because I think the discussion is highly technically nuanced :-)

jrvanwhy (approved):

hudson-ayers (approved):

lschuermann (approved):

PR #2184, opened 2020-10-31
[WIP] Rename AppId to ProcessId

lschuermann:
### Pull Request Overview

The `AppId` struct changes the returned `.id()` whenever an application restarts. It thereby refers to an _application binary instance_, or _process_ instead of the _application binary_ itself. This terminology is in line with the [definitions of threat model](https://github.com/tock/tock/blob/bb8c021833355bb69bd75e8ed652e9e88dcd07eb/doc/threat_model/README.md#definitions):

> A process is a runtime instantiation of an application binary. When an application binary "restarts", its process is terminated and a new process is started using the same binary. Note that the kernel is not considered a process, although it is a thread of execution.

I personally found this confusing [in the past](https://github.com/tock/tock/issues/1914). Especially [with the introduction of a proper application id](https://groups.google.com/forum/#!topic/tock-dev/aduN7fHWXdI), the current terminology would either collide or be very confusing.

This pull request is, in its current form, for the most part a very mechanical search-and-replace to rename the `AppId` struct to `ProcessId`. In addition to that, some variables and methods in the kernel have been renamed in line with the updated terminology. Some documentation comments have been updated.

In general much more work must be put into such a change to ensure a **consistent** naming throughout the kernel, which is in line with the definition of a threat model. Variables in capsules (`appid`, `app_id`, etc.) have not been renamed at this stage - I don't know whether those would need to be renamed as part of this changeset. If this change is beneficial, I am ready to invest time into continuing to develop it.

I'm open to suggestions on better naming schemes, other ways to approach these issues or to make these changes and whether such a change is good at all.

### Testing Strategy

This pull request ~was tested by CI~ needs testing.

### TODO

- [ ] **Verify that this does not change any functionality.**
- [ ] Check whether the MPU code refers to application binaries or processes in the individual statements and carefully update variable naming.
- [ ] Resolve `TODO`s in `kernel/src/grant.rs`.
- [ ] Thoroughly go through every changed file, verify the change, manually adjusting the variable naming whether it refers to a process or application binary respectively.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m3/src/mpu.rs
+++ b/arch/cortex-m3/src/mpu.rs
@@ -10,7 +10,7 @@ use kernel::common::math;
 use kernel::common::registers::{register_bitfields, FieldValue, ReadOnly, ReadWrite};
 use kernel::common::StaticRef;
 use kernel::mpu;
-use kernel::AppId;
+use kernel::ProcessId;
 
 /// MPU Registers for the Cortex-M3 and Cortex-M4 families
 /// Described in section 4.5 of
@@ -130,10 +130,10 @@ const MPU_BASE_ADDRESS: StaticRef<MpuRegisters> =
 pub struct MPU {
     /// MMIO reference to MPU registers.
     registers: StaticRef<MpuRegisters>,
-    /// Optimization logic. This is used to indicate which application the MPU
+    /// Optimization logic. This is used to indicate which process the MPU
     /// is currently configured for so that the MPU can skip updating when the
-    /// kernel returns to the same app.
-    hardware_is_configured_for: OptionalCell<AppId>,
+    /// kernel returns to the same process.
+    hardware_is_configured_for: OptionalCell<ProcessId>,
 }
 
 impl MPU {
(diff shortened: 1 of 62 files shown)


hudson-ayers:
I think this seems desirable -- in the future, is the idea that a ProcessId will be an ID for a single instance of an app on a particular board, while an AppId will persist across restarts/updates?

If we change this, do we also need to rename `AppPtr` and `AppSlice`?

lschuermann:
> I think this seems desirable -- in the future, is the idea that a
> ProcessId will be an ID for a single instance of an app on a
> particular board, while an AppId will persist across restarts/updates?

Yes, this is the exact idea. It's still very important to be able to
reference an application binary instance or process, hence ProcessId and
(the new) AppId would coexist, each with their respective use cases.

> If we change this, do we also need to rename `AppPtr` and `AppSlice`?

That's a good question. Technically, `AppPtr` and `AppSlice` are
referring to functions and slices in the running process. However, given
the current design of the Tock kernel, as far as I'm aware, multiple
instances of the same application binary can't run simultaneously. Given
that in the proposed Tock 2.0 syscall design those types are going to be
reworked such that they also have a default instance which does not
actually point to any process, it doesn't seem urgent or "more correct"
to change these. More importantly, there is no immediate upcoming
collision of these identifiers, as in the case of `AppId`.

In any case I'd like to only do minimal changes to reach a consistent
`ProcessId` naming in the kernel - at least as part of this
patchset. It'll make it easy to verify that indeed only the naming and
documentation has been changed.

-- 
This email was composed and sent using free software.
emacs, notmuch, msmtp, qmail, linux, nixos, gnupg, ...

Proudly brought to you without php, outlook, and win32.

bradjc:
I think this is a change we have been avoiding, and we should probably just do this at some point.

I propose we do this after tagging 2.0rc1 (i.e. in the switch from 1.x to 2.0). That way we don't require many capsules to update multiple times.

lschuermann:
> I propose we do this after tagging 2.0rc1

I also thought of doing this as part of the switch form 1.x to 2.0. I'm not sure what the benefit is of doing this after tagging 2.0rc1 though, because that would imply we'd have to touch every capsule twice: once to adjust to the new driver trait, once for this change.

How about I change this PR to include only the kernel changes (breaking capsule code), and set the target branch to `tock-2.0-dev`? Then we have a place to review and discuss the change, and merge this as soon as the new Driver trait is finished such that only one big change to every capsule is required.

phil-levis:
> I propose we do this after tagging 2.0rc1 (i.e. in the switch from 1.x to 2.0). That way we don't require many capsules to update multiple times.

I'd suggest pre-rc; since it won't be a release candidate. Perhaps 2.0alpha1?

lschuermann:
Sure, we can do that. My original plan was to do this prior to people porting capsules to the Tock 2.0 interface, as they can then also update internal variable names. I can also just do this in the next couple of days, which might have the negative impact of causing a few conflicts which the porting authors would need to resolve.

lschuermann:
I realize this has been stale for some time. I hope it's okay to leave it open until it eventually makes it into the `tock-2.0-dev` branch (as a reminder to myself).

hudson-ayers:
Replaced by #2507

PR #2185, opened 2020-11-02
Adding NFC support in nRF52840 DK 

MirnaMuhammad98:
### Pull Request Overview

This pull request adds:
* An NFC HIL
* An NFC tag peripheral driver
* An NFC capsule
* An NFC component.

### Testing Strategy

This pull request was tested by compiling and an [example application](https://github.com/google/OpenSK/pull/199) that will be part of [OpenSK](https://github.com/google/OpenSK).

### TODO or Help Wanted

Approval of the proposed NFC HIL design.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/components/src/nfc.rs
+++ b/boards/components/src/nfc.rs
@@ -7,7 +7,6 @@
 //! ```
 
 // Author: Mirna Al-Shetairy <mshetairy@google.com>
-// Last modified: 19/10/2020
 
 use capsules::nfc;
 use kernel::capabilities;
--- a/boards/components/src/nfc.rs
+++ b/boards/components/src/nfc.rs
@@ -1,0 +1,60 @@
+//! Component for NFC Tag.
+//!
+//! Usage
+//! -----
+//! ```rust
+//! let nfct = components::nfct::NfcComponent::new(board_kernel, &nrf52840::nfct::NFCT).finalize(());
+//! ```
+
+// Author: Mirna Al-Shetairy <mshetairy@google.com>
+// Last modified: 19/10/2020
+
+use capsules::nfc;
+use kernel::capabilities;
+use kernel::component::Component;
+use kernel::create_capability;
+use kernel::hil::nfc::NfcTag;
+use kernel::static_init;
+
+pub struct NfcComponent {
+    board_kernel: &'static kernel::Kernel,
+    nfct: &'static dyn NfcTag<'static>,
+}
+
+impl NfcComponent {
+    pub fn new(
+        board_kernel: &'static kernel::Kernel,
+        nfct: &'static dyn NfcTag<'static>,
+    ) -> NfcComponent {
+        NfcComponent {
+            board_kernel: board_kernel,
+            nfct: nfct,
+        }
+    }
+}
+
+impl Component for NfcComponent {
+    type StaticInput = ();
+    type Output = &'static nfc::NfcDriver<'static>;
+
+    unsafe fn finalize(self, _static_buffer: Self::StaticInput) -> Self::Output {
+        let grant_cap = create_capability!(capabilities::MemoryAllocationCapability);
+
+        let tx_buffer = static_init!([u8; nfc::MAX_LENGTH], [0u8; nfc::MAX_LENGTH]);
+        let rx_buffer = static_init!([u8; nfc::MAX_LENGTH], [0u8; nfc::MAX_LENGTH]);
+
+        let nfct = static_init!(
+            // Supply to the capsule: the driver and a grant
+            nfc::NfcDriver<'static>,
+            nfc::NfcDriver::new(
+                self.nfct,
+                tx_buffer,
+                rx_buffer,
+                self.board_kernel.create_grant(&grant_cap)
+            )
+        );
+        self.nfct.set_client(nfct);
+        self.nfct.enable();
+        nfct
+    }
+}
--- a/boards/nordic/nrf52840dk/src/main.rs
+++ b/boards/nordic/nrf52840dk/src/main.rs
@@ -120,7 +120,7 @@ pub mod io;
 // Whether to use UART debugging or Segger RTT (USB) debugging.
 // - Set to false to use UART.
 // - Set to true to use Segger RTT over USB.
-const USB_DEBUGGING: bool = true;
+const USB_DEBUGGING: bool = false;
 
 // State for loading and holding applications.
 // How should the kernel respond when a process faults.
--- a/boards/nordic/nrf52840dk/src/main.rs
+++ b/boards/nordic/nrf52840dk/src/main.rs
@@ -120,7 +120,7 @@ pub mod io;
 // Whether to use UART debugging or Segger RTT (USB) debugging.
 // - Set to false to use UART.
 // - Set to true to use Segger RTT over USB.
-const USB_DEBUGGING: bool = false;
+const USB_DEBUGGING: bool = true;
 
 // State for loading and holding applications.
 // How should the kernel respond when a process faults.
@@ -168,6 +168,7 @@ pub struct Platform {
     >,
     nonvolatile_storage: &'static capsules::nonvolatile_storage_driver::NonvolatileStorage<'static>,
     udp_driver: &'static capsules::net::udp::UDPDriver<'static>,
+    nfct: &'static capsules::nfc::NfcDriver<'static>,
 }
 
 impl kernel::Platform for Platform {
@@ -188,6 +189,7 @@ impl kernel::Platform for Platform {
             capsules::analog_comparator::DRIVER_NUM => f(Some(self.analog_comparator)),
             capsules::nonvolatile_storage_driver::DRIVER_NUM => f(Some(self.nonvolatile_storage)),
             capsules::net::udp::DRIVER_NUM => f(Some(self.udp_driver)),
+            capsules::nfc::DRIVER_NUM => f(Some(self.nfct)),
             kernel::ipc::DRIVER_NUM => f(Some(&self.ipc)),
             _ => f(None),
         }
@@ -367,6 +369,9 @@ pub unsafe fn reset_handler() {
         nrf52_components::BLEComponent::new(board_kernel, &base_peripherals.ble_radio, mux_alarm)
             .finalize(());
 
+    let nfc_driver =
+        components::nfc::NfcComponent::new(board_kernel, &nrf52840::nfct::NFCT).finalize(());
+
     let serial_num = nrf52840::ficr::FICR_INSTANCE.address();
     let serial_num_bottom_16 = serial_num[0] as u16 + ((serial_num[1] as u16) << 8);
     let src_mac_from_serial_num: MacAddress = MacAddress::Short(serial_num_bottom_16);
@@ -533,6 +538,7 @@ pub unsafe fn reset_handler() {
         nonvolatile_storage,
         udp_driver,
         ipc: kernel::ipc::IPC::new(board_kernel, &memory_allocation_capability),
+        nfct: nfc_driver,
     };
 
     platform.pconsole.start();
--- a/boards/nordic/nrf52840dk/src/main.rs
+++ b/boards/nordic/nrf52840dk/src/main.rs
@@ -370,7 +370,7 @@ pub unsafe fn reset_handler() {
             .finalize(());
 
     let nfc_driver =
-        components::nfc::NfcComponent::new(board_kernel, &nrf52840::nfct::NFCT).finalize(());
+        components::nfc::NfcComponent::new(board_kernel, &base_peripherals.nfct).finalize(());
 
     let serial_num = nrf52840::ficr::FICR_INSTANCE.address();
     let serial_num_bottom_16 = serial_num[0] as u16 + ((serial_num[1] as u16) << 8);
--- a/capsules/src/nfc.rs
+++ b/capsules/src/nfc.rs
@@ -27,6 +27,8 @@ pub struct NfcDriver<'a> {
     rx_in_progress: OptionalCell<AppId>,
     rx_buffer: TakeCell<'static, [u8]>,
     driver_selected: Cell<bool>,
+    tag_configured: Cell<bool>,
+    current_field: Cell<nfc::NfcFieldState>,
 }
 
 impl<'a> NfcDriver<'a> {
@@ -44,11 +46,23 @@ impl<'a> NfcDriver<'a> {
             rx_in_progress: OptionalCell::empty(),
             rx_buffer: TakeCell::new(rx_buffer),
             driver_selected: Cell::new(false),
+            tag_configured: Cell::new(false),
+            current_field: Cell::new(nfc::NfcFieldState::None),
         }
+    }
+
+    /// Helper function to reset the state of the capsule
+    fn reset(&self) {
+        self.driver_selected.set(false);
+        self.current_field.set(nfc::NfcFieldState::Off);
+        self.tag_configured.set(false);
     }
 
     /// Internal helper function for setting up frame transmission
     fn transmit_new(&self, app_id: AppId, app: &mut App, len: usize) -> ReturnCode {
+        if self.current_field.get() != nfc::NfcFieldState::On {
+            return ReturnCode::ECANCEL;
+        }
         // Driver not ready yet
         if !self.driver_selected.get() {
             return ReturnCode::EOFF;
@@ -92,14 +106,16 @@ impl<'a> NfcDriver<'a> {
 
     /// Internal helper function for starting a receive operation
     fn receive_new(&self, app_id: AppId, app: &mut App, _len: usize) -> ReturnCode {
-        // Driver not ready yet
-        if !self.driver_selected.get() {
+        if !self.tag_configured.get() {
             return ReturnCode::EOFF;
         }
-        if self.rx_buffer.is_none() {
+        if self.current_field.get() != nfc::NfcFieldState::On {
+            return ReturnCode::ECANCEL;
+        }
+        // Driver not ready yet
+        if !self.driver_selected.get() | self.rx_buffer.is_none() {
             return ReturnCode::EBUSY;
         }
-
         if app.rx_buffer.is_some() {
             self.rx_buffer
                 .take()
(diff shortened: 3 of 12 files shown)


chips/nrf52/src/nfct.rs
```diff
@@ -0,0 +1,668 @@
...
+            .errorstatus
+            .set(self.registers.errorstatus.get());
+        self.registers
+            .framestatus_rx
+            .set(self.registers.framestatus_rx.get());
+    }
+
+    pub fn handle_interrupt(&self) {
+        let saved_inter = self.registers.intenset.extract();
+        self.disable_all_interrupts();
```
hudson-ayers, line 493: In Tock, disabling interrupts while handling the bottom half is generally not necessary, as the top half interrupt handler disables the interrupt that fired (see `arch/cortex-m/src/lib.rs` line 140) and the bottom half interrupt handler does not re-enable the interrupt until the bottom half handler (i.e. this function) has executed (see `nrf52/src/chip.rs` `service_pending_interrupts()` function).

I realize that this pattern exists in several other places in Tock, but I generally think that is a result of people basing drivers off of C drivers written for other operating systems without this property.

hudson-ayers (commented):
Very cool to see NFC support! This is an incomplete review with just a comment on something I saw when glancing through the changes.

chips/nrf52/src/nfct.rs
MirnaMuhammad98: I disable the interrupts as I keep track at a time of which interrupt interests the sequence of operations. For example, once the NFC tag has been selected, I'm not interested in the `select` interrupt. Also when the capsule requests a `receive` or `transmit` operation it unmasks the corresponding interrupt.

That was the sequence I adapted.

hudson-ayers: Oh I see, so the assumption is that there will never be more than one possible outstanding interrupt at a time, and disabling all interrupts is equivalent to clearing the flag for the one interrupt that arrived

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,50 @@
...
+
+pub trait NfcTag<'a> {
+    fn set_client(&self, client: &'a dyn Client<'a>);
+
+    /// Enable NFC sense field mode, and subscribe
+    /// to the relevant interrupt.
+    fn enable(&self);
+
+    /// Enables tag emulation.
+    fn activate(&self);
```
alistair23, line 11: Same question, what if this fails?

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// Disables tag emulation.
+    fn deactivate(&self);
+
+    /// Set the buffer to be transmitted and the amount of data
+    /// trigger the task for transmission.
+    fn transmit_buffer(&self, tx_buffer: &'static mut [u8], tx_amount: usize);
```
alistair23, line 17: This should return as result as well.

Something like:

```rust
/// Returns size to be sent on success. On error returns the buffer and a return code
-> Result<usize, (ReturnCode, &'static mut [u8])>
```

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// Pass a buffer for receiving.
+    fn receive_buffer(&self, rx_buffer: &'static mut [u8]);
+
+    /// Configuration of the Tag according to its Type.
+    fn configure(&self, tag_type: u8);
```
alistair23, line 23: I expect that this could fail if a tag isn't supported right?

chips/nrf52/src/nfct.rs
```diff
@@ -0,0 +1,703 @@
...
+        }
+        if events_to_process.is_set(Interrupt::SELECTED) {
+            self.handle_selected();
+        }
+        if events_to_process.is_set(Interrupt::TXFRAMESTART) {
+            self.handle_txstart();
+            // Enable TXFRAMEEND
+            self.registers.intenset.modify(Interrupt::TXFRAMEEND::SET);
+        }
+        self.clear_errors();
```
alistair23, line 506: Why do you clear the errors?

chips/nrf52/src/nfct.rs
```diff
@@ -0,0 +1,703 @@
...
+    fn take_event(event: &ReadWrite<u32, Event::Register>) -> bool {
+        let result = event.is_set(Event::GENERATED);
+        if result {
+            event.write(Event::GENERATED::CLEAR);
+        }
+        result
+    }
+
+    fn disable_all_interrupts(&self) {
+        self.registers.intenclr.set(0xffffffff);
```
alistair23, line 570: Does this disable interrupts or just clear them?

chips/nrf52/src/nfct.rs
```diff
@@ -0,0 +1,703 @@
...
+
+    fn enable(&self) {
+        self.registers.errata98.set(0x38148);
+        self.registers
+            .framedelay_mode
+            .write(FrameDelayMode::FRAMEDELAYMODE::WindowGrid);
+        self.registers.framedelay_max.set(0x1000);
+        // TODO: Remove TASKS_ACTIVATE and Enable TASKS_SENSE instead.
+        self.registers.task_activate.write(Task::ENABLE::Trigger);
+        // self.registers.task_sense.write(Task::ENABLE::Trigger);
```
alistair23, line 592: Don't have commented out code.

capsules/src/nfc.rs
```diff
@@ -0,0 +1,270 @@
...
+                .unwrap_or_else(|err| err.into()),
+            2 => self
+                .application
+                .enter(appid, |app, _| {
+                    app.rx_callback = callback;
+                    ReturnCode::SUCCESS
+                })
+                .unwrap_or_else(|err| err.into()),
+            _ => ReturnCode::ENOSUPPORT,
+        }
```
alistair23, line 225: Why have seperate receive/send callbacks? This makes the user space implementation in Rust tricky as you can't share data between the two.

capsules/src/nfc.rs
```diff
@@ -0,0 +1,270 @@
...
+                self.application.enter(appid, |app, _| {
+                    self.receive_new(appid, app, len)
+                }).unwrap_or_else(|err| err.into())
+            },
+            3 /* enable tag emulation */=> {
+                match arg1 as u8 {
+                    0 /* false */ => self.driver.deactivate(),
+                    _ /* true */ => self.driver.activate(),
+                }
+                ReturnCode::SUCCESS
```
alistair23, line 260: Shouldn't this also be matched by appid?

capsules/src/nfc.rs
```diff
@@ -0,0 +1,270 @@
...
+            }
+            4 /* tag type configuration */ => {
+                let tag_type = arg1;
+                self.driver.configure(tag_type as u8);
+                ReturnCode::SUCCESS
+            }
```
alistair23, line 266: Same here

boards/components/src/nfc.rs
```diff
@@ -0,0 +1,60 @@
+//! Component for NFC Tag.
+//!
+//! Usage
+//! -----
+//! ```rust
+//! let nfct = components::nfct::NfcComponent::new(board_kernel, &nrf52840::nfct::NFCT).finalize(());
+//! ```
+
+// Author: Mirna Al-Shetairy <mshetairy@google.com>
+// Last modified: 19/10/2020
```
alistair23, line 10: Do we really need a last modified date? Git can tell us that and this is already out of date

boards/nordic/nrf52840dk/src/main.rs
```diff
@@ -119,7 +119,7 @@ pub mod io;
 // Whether to use UART debugging or Segger RTT (USB) debugging.
 // - Set to false to use UART.
 // - Set to true to use Segger RTT over USB.
-const USB_DEBUGGING: bool = false;
+const USB_DEBUGGING: bool = true;
```
alistair23: Don't change this

chips/nrf52/src/nfct.rs
```diff
@@ -459,10 +459,6 @@ impl<'a> NfcTag<'a> {
         self.clear_errors();
     }
 
-    fn handle_txstart(&self) {
-        self.set_tx_registers();
-    }
```
alistair23: Just squash this into the original commit

alistair23 (changes requested):
Awesome! It's great to see NFC support and the patches are nice and split up which really helps with reviewing.

I think the HIL needs some more documentation and the ability to return errors.

Can you share the libtock-rs/libtock-c patches you are using for testing? It's probably worth opening a PR for those repos now, even as a draft PR. It makes it easier to understand how everything works together.

chips/nrf52/src/nfct.rs
jmichelp: It disables them.
Nordic has 3 registers for interrupt handling: `INTEN`, `INTENSET` and `INTENCLR`. The last two are only SR flip-flops modifying the `INTEN` register. It's the same as writing `self.registers.inten.set(0)` if you think this is clearer for the reviewer.

jmichelp:
> Can you share the libtock-rs/libtock-c patches you are using for testing? It's probably worth opening a PR for those repos now, even as a draft PR. It makes it easier to understand how everything works together.

This is developed as part of OpenSK and I'm not sure Mirna modified anything in `libtock-rs` because we don't use `futures` yet. That's why we had to create our own `libtock-drivers` mod as part of our repository to keep the old behaviour of libtock-rs before they moved to `futures`.
You can see the file here for NFC: https://github.com/google/OpenSK/blob/master/third_party/libtock-drivers/src/nfc.rs

chips/nrf52/src/nfct.rs
alistair23: `self.registers.inten.set(0)` seems clearer to me, but if what's already there actually does disable them that's fine.

capsules/src/nfc.rs
MirnaMuhammad98: I'm a bit confused as to what exactly do you mean by share data between the two? Do you mean working with one buffer for both operations?

boards/components/src/nfc.rs
MirnaMuhammad98: Yes, my apologies 😅  I was using a rather old file for reference and as my template. I will update it in my upcoming commit 👍

chips/nrf52/src/nfct.rs
MirnaMuhammad98: Those error registers that get cleared are related to transmitting or receiving. They're handled properly in the relevant `TX/RX` functions. And at the end of the `handle_interrupt()` I clear them once more just to ensure there won't be any spurious errors.

I will try to convey that with comments in code.

boards/nordic/nrf52840dk/src/main.rs
MirnaMuhammad98: Yes, my bad 👍

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,50 @@
+use crate::returncode::ReturnCode;
+
+pub trait NfcTag<'a> {
+    fn set_client(&self, client: &'a dyn Client<'a>);
```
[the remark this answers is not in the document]
MirnaMuhammad98: I updated the HIL documentation 👍

capsules/src/nfc.rs
alistair23: From my experience you usually have some struct in userspace that you use for parsing the data you send/receive. My having a separate send and receive callback you can't share that one struct between both callbacks in libtock-rs.

arch/rv32i/src/lib.rs
```diff
@@ -227,7 +227,7 @@ pub extern "C" fn _start_trap() {
             // Jump to board-specific trap handler code. Likely this was an
             // interrupt and we want to disable a particular interrupt, but each
             // board/chip can customize this as needed.
-            jal ra, _start_trap_rust
+            jal ra, _start_trap_rust_from_kernel
```
alistair23: Why rename this?

MirnaMuhammad98: I didn't, this is from one of the commits I pulled recently to update my PR.

alistair23: Maybe it's best to rebase your patches then? The GitHub UI is showing me a large amount of other changes that it's hard to keep track of.

MirnaMuhammad98: But aren't the changed files easy to review the latest commit changes 🤔 ?

alistair23: I'm not clear on what you mean. It's much more difficult to follow changes with merge commits in the history. GitHub shows me a large amount of unrelated changes. A rebase on master is much cleaner to review and makes the Git history tidier and more bisectable.

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,105 @@
+use crate::returncode::ReturnCode;
```
ppannuto, line 1: Really a comment for line 0: Can you add some top-level documentation (`//!` comments) about what this HIL is for -- especially what parts of the wide array of possible NFC implementations that it does and doesn't cover?

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,105 @@
...
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+/// An enum to keep track of the NFC field status
+pub enum NfcFieldState {
+    /// Initial value that indicates no NFCT field events.
+    None,
+    /// The NFCT FIELDLOST event has been set.
+    On,
+    /// The NFCT FIELDDETECTED event has been set.
+    Off,
```
ppannuto, line 11: I'm a bit confused by the words `On` and `Off` here, as compared to the documentation. If I understand correctly, this driver is exclusively for a NFC edge device (i.e. it cannot turn on and generate a field, but it can respond to one when it finds it).

Would this maybe better expressed as a `NfcFieldPresent` enum?

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,105 @@
...
+
+/// Controls an NFC tag main functionalities
+pub trait NfcTag<'a> {
+    /// Set the client instance that will handle callbacks
+    fn set_client(&self, client: &'a dyn Client<'a>);
+
+    /// Enable NFC sense field mode, and subscribe to the relevant interrupts.
+    /// Also set up the default configurations of how frame delay should be
+    /// dealt with (e.g. the maximum delay before a timeout).
+    /// This function should never fail.
```
ppannuto, line 24: "subscribe to the relevant interrupts" is a bit too generic for a HIL.

ppannuto: Should the timeout be a parameter?

Alternatively, is that just something intrinsic to NFC? (Does it need to be mentioned as part of the enable HIL?)

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn enable(&self);
+
+    /// Enable tag emulation by triggering the necessary task.
+    /// This function should never fail.
+    fn activate(&self);
```
ppannuto, line 29: What are the "necessary task"s here?

"should" is too weak, if the function _cannot_ fail, we can say so. However, it seems possible that some NFC implementations may exist but fail to activate for various reasons. This probably needs a `ReturnCode`

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Notify the client by calling `field_lost()`. Then disable tag emulation
+    /// by triggering the necessary task. And go back to state of sense field.
+    /// This function should never fail.
+    fn deactivate(&self);
```
ppannuto, line 33: Same ideas here, what specifically is deactivated, and what happens if it does fail?

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    /// Pass the buffer to be transmitted and the amount of data and take
+    /// ownership of it. Subscribe to the relevant interrupt and trigger
+    /// the task for transmission.
+    ///     
+    /// On success returns the length of data to be sent.
+    /// On failure returns an error code and the buffer passed in.
```
ppannuto, line 40: For HILs, we try to enumerate the possible errors that can be returned when possible

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn transmit_buffer(
+        &self,
+        tx_buffer: &'static mut [u8],
+        tx_amount: usize,
+    ) -> Result<usize, (ReturnCode, &'static mut [u8])>;
+
+    /// Pass a buffer for receiving data and take ownership of it.
+    ///     
+    /// On success returns nothing.
+    /// On failure returns an error code and the buffer passed in.
```
ppannuto, line 50: What is the behavior when a packet is received by the radio, but there is no receive buffer?

What is the behavior when a packet is received by the radio that does not fit in this buffer?

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn receive_buffer(
+        &self,
+        rx_buffer: &'static mut [u8],
+    ) -> Result<(), (ReturnCode, &'static mut [u8])>;
+
+    /// Configuration of the Tag according to its Type.
+    ///
+    /// Return `SUCCESS` if the tag type is supported and
+    /// `ENOSUPPORT` otherwise.
+    fn configure(&self, tag_type: u8) -> ReturnCode;
```
ppannuto, line 60: What is "type" in this context? Is this part of the NFC spec? Can/should this be an enum defined in the HIL instead of a `u8`?

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    /// Set the maximum frame delay in number of 13.56 MHz clocks.
+    fn set_framedelaymax(&self, max_delay: u32);
```
ppannuto, line 63: Is 13.56 MHz a universal NFC clock rate? If it's not part of the NFC spec, we probably shouldn't have a clock rate in the HIL

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    /// Enable the interrupt for event SELECTED
+    fn unmask_select(&self);
```
ppannuto, line 66: I don't think I quite understand what this is, but it feels too chip-specific for a HIL

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/// Implement this trait and use `set_client()` in order to receive callbacks.
```
ppannuto, line 69: How does `tag_selected`/`tag_activated` interact with `field_detected`/`field_lost`? I think some explanation of the expected state progression (either here, or at the very top) would help make this clear for future implementers.

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn tag_deactivated(&'a self);
+
+    /// Called when a field is detected.
+    /// This will notify the app of the presence of a field to activate the tag.
+    fn field_detected(&'a self);
+
+    /// Called when a field is lost.
+    /// This will notify the app of the absence of a field.
+    /// Returns any buffers passed either in `receive_buffer()` or `transmit_buffer()`
+    /// back to the capsule. Also trigger any ready callback functions.
```
ppannuto, line 86: The buffer explanation here is great ✅ 

I'm not sure I quite understand which callbacks might be triggered here. Which ones are they, and will they be triggered in a way that indicates an error?

kernel/src/hil/nfc.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn field_lost(
+        &'a self,
+        rx_buffer: Option<&'static mut [u8]>,
+        tx_buffer: Option<&'static mut [u8]>,
+    );
+
+    /// Called when a frame is received.
+    /// This will return the buffer passed into `receive_buffer()`.
+    /// If the buffer length is smaller then the data length the buffer will only contain part
+    /// of the frame the `result` will contain an `ENOMEM` error. If the received frame contained
```
ppannuto, line 96:
```suggestion
    /// of the frame the `result` will contain an `ESIZE` error. If the received frame contained
```

I think this would be the better error here, `ENOMEM` is usually for when something tries to allocate memory and fails

ppannuto (changes requested):
This is awesome! Super exciting to see NFC support!

I mostly looked over the HIL. This is an important component to get right, as the HIL has to be implementable by _any_ NFC chip. Most of my questions are around how _other_ chips would be able to implement this in the future. Hopefully they all make sense.

bradjc:
If I can quickly summarize the state of this PR: it adds a HIL, and we've found that a little more careful review when a HIL is added saves quite a bit of effort later so we want to make sure we strike a balance of "it works" and "it isn't too chip specific". We don't expect it to be final by any means.

alevy:
@MirnaMuhammad98 This would be great to have! What's the status on addressing/responding to the comments?

jmichelp:
Mirna's internship ended. We have someone who is continuing the effort on OpenSK to have NFC working but as we are facing a few issues on the app side code, I was thinking about waiting that everything works end-to-end in case we discover something that needs to redesign the kernel code.

But if you prefer to do like USB and have it early in and then we iterate on it, I'm fine with it and we'll address the comments to finish this PR.

bradjc:
Closing due to age, but please re-open if/when an updated version is available.

PR #2187, opened 2020-11-03
Board based instantiation of chip drivers and interrupt mappings: Stm32f3

hudson-ayers:
### Pull Request Overview

This pull request ports the stm32f3xx chip and the stm32f3discovery board to the new peripheral instantiation approach that does not rely on global variables (first proposed in #2069). 

This chip was more challenging than most of the previous ones I have switched over, because it extensively relied on shared use of global statics for GPIO Pins / Ports, the clock manager (RCC), and the interrupt controller (EXTI). As a result I had to do a semi-significant refactoring of the GPIO driver to not rely on the assumption that references to gpio pins could be created from anywhere. The GPIO driver is now closer to the design of the GPIO driver for most other chips, but the "ports" and "pins" are still stored separately so that I could minimize the total amount of code I had to change.

Bright side: This PR removes 60 uses of `unsafe` across the `stm32f3xx` and `stm32f3discovery` crates!

### Testing Strategy

This pull request was tested by compiling. But there are enough changes in this PR that hardware testing would be greatly appreciated.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/stm32f3discovery/src/io.rs
+++ b/boards/stm32f3discovery/src/io.rs
@@ -38,7 +38,8 @@ impl Write for Writer {
 
 impl IoWrite for Writer {
     fn write(&mut self, buf: &[u8]) {
-        let uart = unsafe { &mut stm32f303xc::usart::USART1 };
+        let rcc = stm32f303xc::rcc::Rcc::new();
+        let uart = stm32f303xc::usart::Usart::new_usart1(&rcc);
 
         if !self.initialized {
             self.initialized = true;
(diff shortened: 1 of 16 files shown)


bradjc:
In my mind the PRs needed to complete this transition to board-specified chip instantiation are all automatically last-call, but I think we should leave the PRs open for a couple days to give people a chance to take a look.

bradjc (approved):

PR #2189, opened 2020-11-10
Board based instantiation of chip drivers and interrupt mappings: Msp432

hudson-ayers:
### Pull Request Overview

This pull request ports the MSP432 chip/board to the new peripheral instantiation approach that does not rely on global variables (first proposed in #2069).

This chip was relatively easy to change compared to some of the others, but @lebakassemmerl I would still appreciate a quick look over the changes or a test of a couple apps to ensure I haven't done anything stupid.

This is the last remaining chip before all upstream chips/boards have been ported!

### Testing Strategy

This pull request was tested by compiling, hardware testing would be nice.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/msp432/src/uart.rs
+++ b/chips/msp432/src/uart.rs
@@ -76,13 +76,14 @@ pub struct Uart<'a> {
 
 impl<'a> Uart<'a> {
     pub const fn new(
+        registers: StaticRef<UsciARegisters>,
         tx_dma_chan: usize,
         rx_dma_chan: usize,
         tx_dma_src: u8,
         rx_dma_src: u8,
     ) -> Self {
         Self {
-            registers: usci::USCI_A0_BASE,
+            registers,
             clock_frequency: DEFAULT_CLOCK_FREQ_HZ,
 
             tx_client: OptionalCell::empty(),
--- a/chips/msp432/src/uart.rs
+++ b/chips/msp432/src/uart.rs
@@ -9,8 +9,6 @@ use kernel::common::StaticRef;
 use kernel::hil;
 use kernel::ReturnCode;
 
-pub static mut UART0: Uart<'static> = Uart::new(usci::USCI_A0_BASE, 0, 1, 1, 1);
-
 const DEFAULT_CLOCK_FREQ_HZ: u32 = crate::cs::SMCLK_HZ;
 
 struct BaudFraction {
@@ -77,15 +75,14 @@ pub struct Uart<'a> {
 }
 
 impl<'a> Uart<'a> {
-    pub(crate) const fn new(
-        regs: StaticRef<UsciARegisters>,
+    pub const fn new(
         tx_dma_chan: usize,
         rx_dma_chan: usize,
         tx_dma_src: u8,
         rx_dma_src: u8,
-    ) -> Uart<'static> {
-        Uart {
-            registers: regs,
+    ) -> Self {
+        Self {
+            registers: usci::USCI_A0_BASE,
             clock_frequency: DEFAULT_CLOCK_FREQ_HZ,
 
             tx_client: OptionalCell::empty(),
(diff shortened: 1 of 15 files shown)


lebakassemmerl:
Sure, no problem!
I'm currently not at home but I will run the tests on the hardware in the evening today or tomorrow morning (UTC +1).

chips/msp432/src/uart.rs
```diff
@@ -77,15 +75,14 @@ pub struct Uart<'a> {
...
         rx_dma_chan: usize,
         tx_dma_src: u8,
         rx_dma_src: u8,
-    ) -> Uart<'static> {
-        Uart {
-            registers: regs,
+    ) -> Self {
+        Self {
+            registers: usci::USCI_A0_BASE,
             clock_frequency: DEFAULT_CLOCK_FREQ_HZ,
```
lebakassemmerl, line 87: This constructor seems wrong to me since the reference to the registers isn't passed as argument anymore. If other UARTs are instantiated they all point to the same registers.

hudson-ayers: That's a good point, I didn't realize the MSP432 had multiple UARTs. I'll fix

lebakassemmerl:
I ran now a couple of test programs (blink, button, helloworld, ipc, etc.) , and everything seems to work fine. However, I stumbled across a different problem which might be a bug in the linking-stage or within the kernel-linker-script.

When I run `make flash` everything works fine, and the binary fits into the reserved flash-space.
Running `make flash-debug` creates a binary which is slightly too big for the reserved flash. And then I realized that that the `.bss` section is copied into the actual hex-file. I get the following output:

```
   text    data     bss     dec     hex 
  81921       0   16384   98305   18001
```

Initially I thought maybe `.bss` and `.data` are interchanged, but when I make the following changes to `main.rs`:
```Rust
pub static mut TEST: usize = 334;

unsafe fn startup_intilialisation() {
    msp432::init();

    TEST += 1;
    ...
}
```
I get the following output:
```
   text    data     bss     dec     hex 
 114689       4   16128  130821   1ff05
```
So obviously `.bss` and `.data` are not interchanged. 
And as far as I know the `.bss` contains only uninitialized data which gets zeroed after startup. Should this really be in the binary?
And if this is a bug, should I open an issue?

hudson-ayers:
If by "end up in the binary" you mean "ends up in flash" I don't think that is happening. The GNU-size utility used to print the size information says the following:

> The Berkeley style output counts read only data in the text column, not in the data column, the dec and hex columns both display the sum of the text, data, and bss columns in decimal and hexadecimal respectively.

So I think we are seeing exactly what we should expect.

bradjc (dismissed):

lebakassemmerl:
> If by "end up in the binary" you mean "ends up in flash" I don't think that is happening. The GNU-size utility used to print the size information says the following:
> 
> > The Berkeley style output counts read only data in the text column, not in the data column, the dec and hex columns both display the sum of the text, data, and bss columns in decimal and hexadecimal respectively.
> 
> So I think we are seeing exactly what we should expect.

Ok, thanks for the answer!

I am still a bit confused about this because when flashing the binary with openocd it always prints the amount of flashed bytes. And this number equals exactly the decimal or hexadecimal column. But it's not a big problem, I'll try to find the problem when I have time (:

ppannuto:
bors d+

hudson-ayers:
rebased to resolve conflicts

lebakassemmerl:
Will this PR be merged in the near future? Because I have some fixes pending for this platform but I thought it doesn't make sense to open a PR before this one isn't merged.

ppannuto (approved):
bors r+

hudson-ayers:
bors retry

PR #2190, opened 2020-11-10
DALS: A dynamic app-loading system for the kernel

arjundeopujari:
### Pull Request Overview

@bradjc  and I propose "DALS" (acronym for dynamic app-loading system).  This is a series of capsules which work in unison to take raw received data from standard embedded systems communication protocols such as BLE, Zigbee, or USB and load it into memory and verify it at runtime.  In essence, a system to dynamically load apps onto Tock-compatible HW at runtime. The design is detailed in the README and more code-specific comments and interface specifications are in `interfaces.rs`.  

**NOTE**:  The purpose of this PR is not necessarily to have these changes merged but rather to solicit feedback/constructive
criticism/questions from the community on suggestions, potential problems, etc. with this design.  We believe this design is a worthwhile future direction for Tock especially with respect to IoT market demands.

### Testing Strategy

N/A

### TODO or Help Wanted

N/A

### Documentation Updated

- Added `README.md` in capsules/src/dals/

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/boards/imxrt1050-evkb/Makefile
+++ b/boards/imxrt1050-evkb/Makefile
@@ -1,16 +1,7 @@
 # Makefile for building the tock kernel for the i.MX RT1052 platform
-#
-APP=../../../libtock-c/examples/c_hello/build/cortex-m7/cortex-m7.tbf
-# APP=../../../libtock-rs/target/thumbv7em-none-eabi/tab/imxrt1050/hello_world/cortex-m7.tbf
+
 TARGET=thumbv7em-none-eabi
 PLATFORM=imxrt1050-evkb
-
-D_KERNEL=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/debug/$(PLATFORM).elf
-D_KERNEL_WITH_APP=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/debug/$(PLATFORM)-app.axf
-R_KERNEL=$(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
-R_KERNEL_WITH_APP=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/release/$(PLATFORM)-app.axf
-
-MCUX_IDE_BIN=/Applications/MCUXpressoIDE_11.1.1_3241/ide/plugins/com.nxp.mcuxpresso.tools.bin.macosx_11.1.0.202002241259/binaries/
 
 include ../Makefile.common
 
@@ -18,20 +9,19 @@ include ../Makefile.common
 .PHONY: install
 install: flash
 
-.PHONY: $(TOCK_ROOT_DIRECTORY)program
-program: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
-	arm-none-eabi-objcopy --update-section .apps=$(APP) $(D_KERNEL) $(D_KERNEL_WITH_APP)
-	$(warning Tock OS was compiled. In order to flash it on the board, you will need MCUExpresso. See README.md for details.)
-
-.PHONY: $(TOCK_ROOT_DIRECTORY)flash-debug
-flash-debug: program
-	$(MCUX_IDE_BIN)/crt_emu_cm_redlink --flash-load-exec "$(D_KERNEL_WITH_APP)" -p MIMXRT1052xxxxB \
-		--ConnectScript RT1050_connect.scp --flash-driver= -x . \
-		--flash-dir $(MCUX_IDE_BIN)/Flash --flash-hashing
-
-.PHONY: $(TOCK_ROOT_DIRECTORY)flash
+.PHONY: flash
 flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
-	arm-none-eabi-objcopy --update-section .apps=$(APP) $(R_KERNEL) $(R_KERNEL_WITH_APP)
-	$(MCUX_IDE_BIN)/crt_emu_cm_redlink --flash-load-exec "$(R_KERNEL_WITH_APP)" -p MIMXRT1052xxxxB \
-		--ConnectScript RT1050_connect.scp --flash-driver= -x . \
-		--flash-dir $(MCUX_IDE_BIN)/Flash --flash-hashing
+	crt_emu_cm_redlink \
+		--flash-load-exec "$(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf" \
+		-p MIMXRT1052xxxxB --ConnectScript RT1050_connect.scp \
+		--flash-driver= -x . --flash-dir $(MCUX_IDE_BIN)/Flash --flash-hashing
+
+.PHONY: flash-app
+flash-app: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
+	@: $(if $(value APP),,$(error Please set APP to the path of a TBF file to program applications))
+	arm-none-eabi-objcopy --update-section .apps=$(APP) $< \
+		$(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-apps.elf
+	crt_emu_cm_redlink \
+		--flash-load-exec "$(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-apps.elf" \
+		-p MIMXRT1052xxxxB --ConnectScript RT1050_connect.scp \
+		--flash-driver= -x . --flash-dir $(MCUX_IDE_BIN)/Flash --flash-hashing
--- a/boards/imxrt1050-evkb/README.md
+++ b/boards/imxrt1050-evkb/README.md
@@ -4,11 +4,48 @@ i.MX RT1050 Crossover MCU with Arm® Cortex®-M7 core
 For more details about the board [visit the NXP  board website](https://www.nxp.com/design/development-boards/i-mx-evaluation-and-development-boards/i-mx-rt1050-evaluation-kit:MIMXRT1050-EVK). Details about the chip can be found [here](https://www.nxp.com/products/processors-and-microcontrollers/arm-microcontrollers/i-mx-rt-crossover-mcus/i-mx-rt1050-crossover-mcu-with-arm-cortex-m7-core:i.MX-RT1050).
 
 
-## Building the app
+## Flashing the kernel
+
+Prerequisites:
+- NXP MCUXpresso SDK software
+
+  The `crt_emu_cm_redlink` binary for flashing the board is not added
+  to the PATH in a standard MCUXpresso installation. Due to the
+  version- and OS-dependent paths, this has to be done manually. It
+  will be located in paths such as
+
+  - on macOS: `/Applications/MCUXpressoIDE_11.1.1_3241/ide/plugins/com.nxp.mcuxpresso.tools.bin.macosx_11.1.0.202002241259/binaries/`
+  - on Debian-based Linux distributions: `/usr/local/mcuxpressoide-11.1.1_3241/ide/plugins/com.nxp.mcuxpresso.tools.bin.linux_10.3.1.201811211038/binaries/`
+
+  Please add these directories to your path, like so:
+  ```bash
+  export PATH=$PATH:<the directory containing crt_emu_cm_redlink>`
+  ```
+- for the `flash-app` target: `gcc-arm-none-eabi`
+
+To compile Tock and flash the kernel onto the board, run `make
+flash`. Its output should look something like:
+
+```bash
+$ make flash
+    Finished release [optimized + debuginfo] target(s) in 0.02s
+   text	   data	    bss	    dec	    hex	filename
+  82365	      0	  16384	  98749	  181bd	/home/USER/tock/target/thumbv7em-none-eabi/release/imxrt1050-evkb
+crt_emu_cm_redlink \
+	--flash-load-exec "/home/USER/tock/target/thumbv7em-none-eabi/release/imxrt1050-evkb.elf" \
+	-p MIMXRT1052xxxxB --ConnectScript RT1050_connect.scp \
+	--flash-driver= -x . --flash-dir /Flash --flash-hashing
+Ns: MCUXpresso IDE RedlinkMulti Driver v11.3 (Mar 30 2021 17:55:57 - crt_emu_cm_redlink build 18)
+Wc(03). No cache support.
+Nc: Found chip XML file in ./MIMXRT1052xxxxB.xml
+Nc: Restarted LinkServer process (PID 22344).
+```
+
+## Running an app
 
 Apps are built out-of-tree. Once an app is built, you can use
-`arm-none-eabi-objcopy` with `--update-section` to create an ELF image with the
-apps included.
+`arm-none-eabi-objcopy` with `--update-section` to create an ELF image
+with your app(s) included.
 
 ```bash
 $ arm-none-eabi-objcopy  \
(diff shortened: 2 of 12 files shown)


capsules/src/dals/README.md
```diff
@@ -0,0 +1,100 @@
...
+verify that no bit-errors occured during wireless transmission or that the loaded app is not malware.
+Examples of this include checksums, digital signatures or more complex cryptographic computations.
+
+This capsule must contain an immutable reference to the app binary data in flash space.
+
+## Validator
+
+This capsule implements `pub trait Validator<'a>`.  This capsule contains an algorithm which 
+performs validation on the app data in flash according to a user specification.  It could check the 
+app's length, version, or header info for any salient features.
```
bradjc, line 92:
```suggestion
This capsule implements `pub trait Validator<'a>`.  This capsule contains an algorithm which 
performs validation on the app data in flash according to a user specification. This allows the
kernel to check if it actually wants to run the newly loaded app. The validator can check the 
app's length, version, or header info for any salient features.
```

capsules/src/dals/README.md
```diff
@@ -0,0 +1,100 @@
...
+This capsule implements `pub trait AppLoaderClient<'a>`.  This capsule can be any capsule
+which receives data via standard communication protocols for embedded systems such as BLE, 
+Zigbee, USB.  This capsule's main function is to send this raw received data buffer-by-buffer
+to the **AppLoader** capsule.  In order to prevent bottleneck issues, this capsule must send an ack 
+only after its buffer is returned to it from the **AppLoader**.
+
+This capsule must also include a buffer the size of one data segment (different for different protocols) which
+it sends to the **AppLoader** to send one segment of the received data.
+
+## Decompressor
```
alistair23, line 69: Does the decompressor write the data to flash?

capsules/src/dals/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+**DALS** support data decompression methods which "operate on-the-fly".  That is, the received app data
+is decompressed segment-by-segment. This capsule must implement `pub trait Decompressor<'a>` and contain 
+a decompression algorithm which conforms to the above constraints.
+
+This capsule must include a buffer in which to store the resulting decompressed data which is passed
+back to the **AppLoader**.
+
+
+## Verifier
```
alistair23, line 79: What happens if the data isn't valid? It's already in flash, is it erased is the packet re-sent?

capsules/src/dals/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+This capsule implements `pub trait Verifier<'a>`.  This capsule contains an algorithm which can
+perform data verification on the app data loaded into flash.  Such an algorithm could be used to 
+verify that no bit-errors occured during wireless transmission or that the loaded app is not malware.
+Examples of this include checksums, digital signatures or more complex cryptographic computations.
+
+This capsule must contain an immutable reference to the app binary data in flash space.
+
+## Validator
```
alistair23, line 88: Crypto checks would probably occur here, to make sure the app matches signatures that we allow to run for example.

capsules/src/dals/README.md
```diff
@@ -0,0 +1,100 @@
...
+the raw app data from another capsule, moves it through a pipeline of steps, places the
+resulting app binary in an empty slot in flash space, and, finally, sets up its process-memory region
+so the scheduler can run it at the appropriate time.
+
+Basic Pipeline at the core of **DALS**.
+
+
+Collect raw data from peripheral capsule (BLE, Zigbee, etc.) -----> Decompress Data  ----->  Find empty slot in flash --- 
+
+---> Verify Data ------> Validate Data ----> Set up process-specific structures (stack, heap, etc.)
```
alistair23, line 27: Some of these happen in parallel right? Maybe there is a better way to illustrate that

capsules/src/dals/README.md
```diff
@@ -0,0 +1,100 @@
...
+The peripheral modules are detailed below.  Please see `interfaces.rs` for more information on functions and such.
+
+
+## AppLoaderClient
+
+This capsule implements `pub trait AppLoaderClient<'a>`.  This capsule can be any capsule
+which receives data via standard communication protocols for embedded systems such as BLE, 
+Zigbee, USB.  This capsule's main function is to send this raw received data buffer-by-buffer
+to the **AppLoader** capsule.  In order to prevent bottleneck issues, this capsule must send an ack 
+only after its buffer is returned to it from the **AppLoader**.
```
hudson-ayers, line 64: This may not be possible for arbitrary communication protocols. For example 802.15.4 expects an ACK within 1-2 ms or an automatic retry will be triggered, and Zigbee will give up on a MAC layer retransmissions (after sending 5x) after 37 ms. It seems that this realistically will create a requirement for higher layer flow control (i.e. use of TCP or similar).

Even for BLE, I don't think our current BLE stack supports L2CAP flow control, which is the only way I know of to achieve this.

capsules/src/dals/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+
+## Verifier
+
+This capsule implements `pub trait Verifier<'a>`.  This capsule contains an algorithm which can
+perform data verification on the app data loaded into flash.  Such an algorithm could be used to 
+verify that no bit-errors occured during wireless transmission or that the loaded app is not malware.
+Examples of this include checksums, digital signatures or more complex cryptographic computations.
+
+This capsule must contain an immutable reference to the app binary data in flash space.
```
hudson-ayers, line 86:
```suggestion
This capsule must accept an immutable reference to the app binary data in flash space.
```
Maybe? It seems like this capsule would not always own such a reference but just receive it as part of the interaction with the `AppLoader`

capsules/src/dals/interfaces.rs
```diff
@@ -0,0 +1,107 @@
...
+/// such as Wifi, Bluetooth, Zigbee, USB, etc.
+pub trait AppLoaderClient<'a>{
+    /// Called by `AppLoader` implementation when data buffer is ready to be returned to `AppLoaderClient`
+    ///
+    /// - `data_buffer`: Buffer which was passed to `AppLoader` via the `next_buffer()` call 
+    fn return_buffer(&self, data_buffer: &'static mut [u8]);
+}
+
+/// Implemented by main `AppLoader` module
+pub trait AppLoader<'a>{
```
hudson-ayers, line 33: Does it make sense to express the dependence on the client traits here?
```suggestion
pub trait AppLoader<'a>: DecompressorClient<'a> + VerifierClient<'a> + ValidatorClient<'a> {
```

capsules/src/dals/interfaces.rs
```diff
@@ -0,0 +1,107 @@
...
+
+/// Implemented by any module which provides an algorithm to verify the loaded app binary for security
+/// purposes (ex: SHA, Checksum, MD5 implementations)
+pub trait Verifier<'a>{
+    fn set_client(&self, client: &'a dyn VerifierClient);
+    /// Called by `VerifierClient` implementation
+    ///
+    /// - `app_flash` : Immutable reference to app data stored in flash
+    /// - `len` : Size of app in flash
+    fn verify_data(&self, app_flash: &'static [u8], len: u32);
```
hudson-ayers, line 80: Given that slices already encode length, why do we need to pass the length separately here? It seems that we should just pass a slice that only provides access to the relevant portion of flash and no more.

(This applies to all the other methods in this file that use this pattern)

capsules/src/dals/interfaces.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn set_client(&self, client: &'a dyn ValidatorClient);
+    /// Called by `ValidatorClient`
+    ///
+    /// - `app_flash` : Immutable reference to app data stored in flash
+    /// - `len` : Size of app in flash 
+    fn validate_data(&self, app_flash: &'static [u8], len: u32);
+}
+
+/// Implemented by main `AppLoader` module
+pub trait ValidatorClient<'a>{
```
hudson-ayers, line 102:
```suggestion
pub trait ValidatorClient {
```
The trait doesn't seem to need the lifetime?

capsules/src/dals/interfaces.rs
arjundeopujari: Hmmm... that makes sense.

capsules/src/dals/README.md
arjundeopujari: No. the decompressor will put the decompressed data into a buffer which AppLoader will read from.  Ill make it clearer in the README.

capsules/src/dals/README.md
arjundeopujari: @bradjc and I discussed that the flash memory used will be deallocated if checks do not pass and an error will propagate up the stack towards the AppLoaderClient implementing capsule.  At that point, that capsule can choose how to handle the error (re-send some packets, give up, etc.)

capsules/src/dals/README.md
arjundeopujari: @bradjc and I discussed we think its a good idea to leave crypto checks in Verification stage.  Validation stage is designed to be more based around system designer policies.  For example, the user/designer could institute a policy to not run apps which need x, y, or z hardware resources.  Verification is geared more towards running some algorithm on the whole app binary in flash which checks for bit-errors or digital signature matches.  Let me know if this works.

capsules/src/dals/README.md
alistair23: Ok, can you make that clearer here?

capsules/src/dals/README.md
alistair23: Won't a signature require the entire app to be loaded though?

capsules/src/dals/README.md
```diff
@@ -0,0 +1,100 @@
...
+to be loaded and run by Tock is transmitted to a Tock-compatible board through any 
+standard  communication protocol (BLE, Zigbee, USB, etc.) which provides "acks" or a similar service
+to lower data rates.  This data is then pushed 
+segment-by-segment through **DALS** which places it in an empty space in flash space and
+then sets up the process-specific elements for the newly-loaded app.
+
+## Design
+
+The design of **DALS** consists of a series of capsules with one capsule, the **AppLoader**, considered
+to be trusted.  This main trusted capsule is the heart of **DALS** and contains the code which collects
```
jrvanwhy, line 17: Several months ago, on the core WG call, we discussed the "capsule" terminology and whether or not a capsule can be trusted ([notes](https://github.com/tock/tock/blob/master/doc/wg/core/notes/core-notes-2020-04-03.md)). We came to the conclusion that "capsule" never refers to code that the core kernel trusts. You are correct to state that some component of the DALS [must be trusted by the core kernel](https://github.com/tock/tock/blob/master/doc/threat_model/Application_Loader.md).

Based on those discussions, I think the trusted code should live outside the "capsules" directory. The rest of DALS -- the code that doesn't need to be trusted under Tock's threat model -- can be in capsules.

capsules/src/dals/interfaces.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/// Implemented by any module which aims to validate the loaded app according to user specifications 
+pub trait Validator<'a>{
+    fn set_client(&self, client: &'a dyn ValidatorClient);
+    /// Called by `ValidatorClient`
+    ///
+    /// - `app_flash` : Immutable reference to app data stored in flash
+    /// - `len` : Size of app in flash 
+    fn validate_data(&self, app_flash: &'static [u8], len: u32);
```
jrvanwhy, line 98: By handing the Validator a `&'static [u8]`, the AppLoader loses the ability to modify that data in flash until the system is restarted. Is that intended?

capsules/src/dals/README.md
arjundeopujari: Sure.

capsules/src/dals/README.md
arjundeopujari: Hmmm... @bradjc would you agree with moving the "trusted code" outside of the capsules directory?

bradjc: Yes this is a good point. Something needs to check that the start of the app header is valid when it is written to flash. One way to do this would be to embed that check in the implementation of the flash interface given to DALS. Something (very roughly) like:

```rust
impl NonvolatileStorage for AppFlashWriter {
    fn write(offset, buffer, len) {
        if offset < 8 {
            // make sure header in buffer is valid
        } else {...}
    }
    ...
}
```

That implementation would 1) confine the app to loading only in the flash space allocated for the incoming app, and 2) verify that the start of the TBF header it writes to flash has the correct version and length.

The code outlined in this PR would then be untrusted, and the trust lies in a proper sandboxed nonvolatile storage interface being provided to DALS.

capsules/src/dals/interfaces.rs
bradjc: Implicitly, yes, as in the current design AppLoader would have no need to write the flash at this point. The app has already been entirely loaded.

hudson-ayers: This suggests that once an app has been loaded into a location it cannot be removed and replaced by another app without a system restart?

capsules/src/dals/README.md
arjundeopujari: Sorry for missing this.  The pipeline is loading --> decompressing --> verification --> validation.  After the decompressing phase the `AppLoader` will write the app binary into flash space.  The verification phase can then process on the app binary data in flash to perform any crypto checks/signature-checking.

capsules/src/dals/README.md
arjundeopujari: Sorry for missing this.  @bradjc and I seem converged on the idea of leaving this "traffic control" idea up to the actual protocol instead.  Obviously since there does not seem to be a one size fit all solution.  For protocols such as 802.15.4 for which timing is an issue, we think it is best to leave this problem up to the system designer.

capsules/src/dals/interfaces.rs
arjundeopujari: Yes.

hudson-ayers: hmm, okay. I imagined that DALS would also enable app updates by making it possible to replace an existing app with a new version, but I suppose that will require some other mechanism.

arjundeopujari: I suppose a slight change to the interface could be made where Verifier/Validator was pass a reference to the flash region back and forth.  This would enable the AppLoader to receive this pointer again after the Verification and Validation stages are over and make any necessary changes to the app.

capsules/src/dals/interfaces.rs
arjundeopujari: Yes. It does.  That is a good idea to make sure they are all implemented by the same module.

capsules/src/dals/interfaces.rs
bradjc: The core kernel can create any slices it wants, so perhaps we want the core kernel to provide the slice when it allocates space for DALS to store the app.

In general, the interface to flash is not specified in this PR, but I'm imagining that AppLoader will have to ask the core kernel where it can store an incoming app of a certain length, and the core kernel will provide some mechanism for DALS to write the app in that location.

bradjc:
We got good feedback on the design, and are now working on the implementation.

arjundeopujari:
Yes.  Thank you to everyone for your feedback.

kaczmarczyck:
Hi, this looks cool! You say apps can be loaded by
> protocols such as BLE, Zigbee, or USB

I have a question about a potential use case: One app uses some channel to download a new app. Can it use DALS to install that app next to it?

bradjc:
> Hi, this looks cool! You say apps can be loaded by
> 
> > protocols such as BLE, Zigbee, or USB
> 
> I have a question about a potential use case: One app uses some channel to download a new app. Can it use DALS to install that app next to it?

In the initial version, no. We are starting with only support in the kernel. However, the design will make it possible for a capsule to provide a syscall interface for userspace to allow an app to handle loading apps. That is in phase 2 of this project.

arjundeopujari:
> Hi, this looks cool! You say apps can be loaded by
> 
> > protocols such as BLE, Zigbee, or USB
> 
> I have a question about a potential use case: One app uses some channel to download a new app. Can it use DALS to install that app next to it?

Yes. Currently, we are in the testing phase of the initial implementation.  However, we expect that down the road we will add an interface for applications to replace themselves with updated versions or load other apps.  This PR will see a lot of new content in the coming weeks so we would be glad if you want to check out the new changes or suggest features to add :) @kaczmarczyck

kaczmarczyck:
> an interface for applications to replace themselves with updated versions

This is exactly what I'm interested in! :) Feel free to ping me on your new changes.

arjundeopujari:
> > an interface for applications to replace themselves with updated versions
> 
> This is exactly what I'm interested in! :) Feel free to ping me on your new changes.

Ok.  @kaczmarczyck you can also join the Tock slack channel and query us on project progress there if you would like! Thanks for your interest.

kaczmarczyck:
Thanks! I joined and will check in there next time.

PR #2193, opened 2020-11-12
Clarify the semantics of erase_page

kevtan:
### Pull Request Overview

Users of Tock would like to know that data is not being unintentionally leaked. This was not clear from the flash HIL originally, but is fixed now.

### Testing Strategy

Admittedly, there is an implied assumption that chips will actually zero out the page, but this seems like a reasonable one. I consulted @hudson-ayers on this one.

### Documentation Updated

No updates are required.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/hil/flash.rs
+++ b/kernel/src/hil/flash.rs
@@ -127,7 +127,7 @@ pub trait Flash {
         buf: &'static mut Self::Page,
     ) -> Result<(), (ReturnCode, &'static mut Self::Page)>;
 
-    /// Erase a page of flash.
+    /// Erase a page of flash by setting every byte to 0xFF.
     fn erase_page(&self, page_number: usize) -> ReturnCode;
 }
 


lschuermann:
I'm not quite sure how well this operation maps to different flash devices. Secure erase (/overwrite) is a difficult issue, if your device would feature wear leveling etc. it might not even be possible to securely overwrite data at all. Maybe it might be a good idea to introduce a separate Trait for secure erase operations, which would then actually make certain guarantees regarding the data confidentiality? Devices supporting these features could implement is, whereas others would only implement the base trait?

Are there any flash devices where the native erase command would actually set all bits to one (the inverted behavior as you described)? In that case those would need to do both an erase and a write operation to comply with the specification.

These are just some thoughts, let's hear what the others think first before making any changes. :)

kevtan:
These are all great points that I think need to be addressed. I looked at the 6 implementers of the `kernel::hil::flash::Flash` trait but was unable to confirm what the actual behavior of erasing is. It looks like most write to registers in order to trigger an erase operation and the specifics are left up to the hardware (so maybe we'd have to look at individual data sheets to find out what actually happens)? I'm not 100% certain.

The two-trait idea sounds really promising to me. Devices that support secure erasures can use the secure trait while devices that don't support secure erasures will implement the base one. It makes it clear to uses what guarantees they can expect from the hardware!

kernel/src/hil/flash.rs
```diff
@@ -127,7 +127,7 @@ pub trait Flash {
         buf: &'static mut Self::Page,
     ) -> Result<(), (ReturnCode, &'static mut Self::Page)>;
 
-    /// Erase a page of flash.
+    /// Erase a page of flash by setting every byte to zero.
```
alistair23, line 130: This doesn't seem right. Normally a flash erase sets everything to 1

alistair23: https://en.wikipedia.org/wiki/Flash_memory#Block_erasure

alistair23:
We can reasonably assume that when calling erase page on flash that it will be erased. Flash erasure isn't something that can just be skipped as then the flash page becomes unusable. As I mentioned in the comment though, I would expect it to set to all 1s as that is usually how flash erasure works.

kernel/src/hil/flash.rs
lschuermann: Ah, then my suspicion of the previous comment is confirmed. Wasn't sure anymore. :+1:

kevtan:
@alistair23 Thanks for linking to the Wikipedia article! I didn't know that's how flash memory works. Do you think it'd be still valuable to add a comment there (maybe with the link you provided) or at the module-level for the `Flash` trait? Or do you think it is sufficiently obvious to the average user of flash memory?

alistair23:
I don't think you really need a comment. If you want to clarify that all bits/bytes should be 1/0xFF then that's fine as well. Up to you.

kevtan:
@alistair23 I think it's still valuable to put this in the off chance it'll help someone down the line!

alistair23 (approved):

hudson-ayers (approved):

lschuermann (approved):
I'm still not entirely convinced that it's a good idea to make this guarantee in a hardware-independent abstraction layer. On the other hand this does seem to be the behavior of (at least) the vast majority of flash chips.

*If* we encounter & implement a flash device which would violate this guarantee down the line, the change would have to be reverted and all implementations checked on relying on this guarantee.

bradjc:
Is `erase` ambiguous? I suppose a `hil::Flash` implementation could just skip implementing erase if the actual storage hardware could always be written and didn't need to be erased first. Clarifying that tock expects erase to actually _do_ something is fine, but I think it would be simpler to just say "sets all bits to the same value" and avoid any issues with some hypothetical future hardware.

lschuermann:
> I suppose a `hil::Flash` implementation could just skip implementing erase if the actual storage hardware could always be written and didn't need to be erased first

That's a great point! For such HW we wouldn't want unnecessary overwrites for every page write (would mean two writes for one operation). In this case, maybe it'd be a great idea to have an explicit `overwrite` function, which independently of the hardware would set all the bits to the same value? This would also work with the idea of an "extension trait", implemented by HW which can actually guarantee that such overwrites will happen.

alistair23:
Unless there is some new flash medium invented I think this is fine as is. Flash should always be erased to 1 as flash can only ever be written from a 1 to a 0. This is common to all flash. Flash based file systems use this assumption in their design as well.

kevtan:
@alistair23 @bradjc @lschuermann After doing a little more digging, I found that the behavior Alistair initially pointed out is _inherent_ to the way flash memory works; it's not a convention or a "most common case." See [here](https://www.microchip.com/forums/FindPost/833493) for a pretty decent explanation. That being said, I'm going to keep the comment. Seeing as there's been so much debate over this, I'm convinced it will be useful for people down the line.

How do I actually merge this?

alistair23 (approved):

PR #2194, opened 2020-11-12
capsules/led: migrate from hil::gpio::Pin to hil::led::Led

lschuermann:
### Pull Request Overview

This pull request
- changes the LED driver capsule to use the trait `hil::led::Led` instead of `hil::gpio::Pin`
- renames the LED driver capsule from `LED` to `LedDriver`, to disambiguate it from the `Led` trait
- changes the `LedHigh` and `LedLow` wrappers to use immutable references of `hil::gpio::Pin` structs, as `Pin`s don't require mutable references for mutation and this would otherwise break reference handling in some board files
- updates the led component to take instances of `hil::led::Led`
  - makes the variadic `led_component_helper!` interface accept trailing commas
- updates the boards accommodating for these changes

These changes are done as I'm porting to a board which has a dedicated LED controller. Technically the LEDs are therefore not connected to GPIOs. I would still like to expose them to userspace though. These changes seem reasonable, since in the future one might want to support shift registers, dedicated LED driver chips or chained LEDs through the driver.

Implications:
- In this implementation, the `LedDriver` is made generic over a specific `hil::led::Led` type. As the LED polarity is encoded in the wrapper types, this makes it impossible to operate LEDs that are active-low and active-high through the same driver.

  This can be solved in two ways
  - use `dyn hil::led::Led` in the `LedDriver`: this will make all calls go through a vtable lookup and has a runtime performance impact for all boards
  - create a `DynLed<'a>` wrapper in `kernel/hil/led.rs` that wraps a `&'a mut dyn hil::led::Led`, which can then be used as the generic type for `LedDriver`

    This approach is essentially the same as the first one, but only boards which had the requirement of using different `hil::led::Led` types would go through the vtable lookup. It would add one level of indirection for these boards though, and the amount of initialization code for the boards would increase.
  
- This changeset should not have any (significant) runtime performance or code size impact. The `LedHigh` and `LedLow` wrappers are essentially zero-sized. This does add another layer of references, as each GPIO wrapper has to be statically allocated and calls go through a reference.

### Testing Strategy

- the blink example on an nRF52840DK
- the blink example on an nRF52840DK with `LedHigh`, to confirm that the pattern is inverted

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

Requesting a review of @hudson-ayers because of changes to the component and macros.

### DIFF
--- a/boards/acd52832/src/io.rs
+++ b/boards/acd52832/src/io.rs
@@ -10,6 +10,6 @@ use nrf52832::gpio::Pin;
 #[panic_handler]
 pub unsafe extern "C" fn panic_fmt(_pi: &PanicInfo) -> ! {
     const LED1_PIN: Pin = Pin::P0_22;
-    let led = &mut led::LedLow::new(&mut nrf52832::gpio::PORT[LED1_PIN]);
+    let led = &mut led::LedLow::new(&nrf52832::gpio::PORT[LED1_PIN]);
     debug::panic_blink_forever(&mut [led])
 }
--- a/boards/acd52832/src/main.rs
+++ b/boards/acd52832/src/main.rs
@@ -15,6 +15,7 @@ use kernel::hil;
 use kernel::hil::adc::Adc;
 use kernel::hil::entropy::Entropy32;
 use kernel::hil::gpio::{Configure, InterruptWithValue, Output};
+use kernel::hil::led::LedLow;
 use kernel::hil::rng::Rng;
 use kernel::hil::time::{Alarm, Counter};
 #[allow(unused_imports)]
@@ -64,7 +65,10 @@ pub struct Platform {
     button: &'static capsules::button::Button<'static, nrf52832::gpio::GPIOPin<'static>>,
     console: &'static capsules::console::Console<'static>,
     gpio: &'static capsules::gpio::GPIO<'static, nrf52832::gpio::GPIOPin<'static>>,
-    led: &'static capsules::led::LED<'static, nrf52832::gpio::GPIOPin<'static>>,
+    led: &'static capsules::led::LedDriver<
+        'static,
+        LedLow<'static, nrf52832::gpio::GPIOPin<'static>>,
+    >,
     rng: &'static capsules::rng::RngDriver<'static>,
     temp: &'static capsules::temperature::TemperatureSensor<'static>,
     ipc: kernel::ipc::IPC,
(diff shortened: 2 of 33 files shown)


bradjc (dismissed):

alistair23 (approved):
Nice cleanup

hudson-ayers (dismissed):
Looks good to me!

This change does mean that the LED component doesn't support boards with a mixture of ActiveHigh / ActiveLow LEDs, but I think that is okay, as no Tock boards today are like that and components are optional anyway.

lschuermann:
> This change does mean that the LED component doesn't support boards with a mixture of ActiveHigh / ActiveLow LEDs

Correct. See the _Implications_ section of the PR description for potential solution to these issue. I don't think making the `LedDriver` use dynamic trait reference is worth the performance penalty for all boards though, so a reasonable solution is the introduction of the `DynLed` type as proposed above. I can already do this in this PR or we can wait until the need arises.

hudson-ayers:
> > This change does mean that the LED component doesn't support boards with a mixture of ActiveHigh / ActiveLow LEDs
> 
> Correct. See the _Implications_ section of the PR description for potential solution to these issue. I don't think making the `LedDriver` use dynamic trait references is worth the performance penalty for all boards though, so a reasonable solution is the introduction of the `DynLed` type as proposed above. I can already do this in this PR or we can wait until the need arises.

Oh sorry, I didn't read the whole description haha. Yeah I agree with your proposed solution and that it is fine to wait until the need arises.

LGTM

lschuermann:
Rebased onto current master, integrating @hudson-ayers' changes to the `stm32f2` peripheral instantiation.

alistair23 (dismissed):

hudson-ayers (dismissed):

bradjc (dismissed):

lschuermann:
Resolved conflicts introduced by `465bbae6ea`, rebased, tested, clippy-validated. Let's give try this one more time @hudson-ayers @bradjc @alistair23. :smile:

hudson-ayers (approved):

alistair23 (approved):

ppannuto (approved):
bors  r+

alistair23:
bors retry

hudson-ayers:
bors retry

hudson-ayers:
bors retry

PR #2196, opened 2020-11-13
Virtual AES Driver

Ouyancheng:
### Pull Request Overview

This pull request adds the virtualizer for the AES driver 

### Testing Strategy

- File boards/imix/src/test/virtual_aes_ccm_test.rs contains testing with 3 clients. Each of the client is performing the tests specified at capsules/src/test/aes_ccm.rs 

### TODO or Help Wanted

- Currently we couldn't remove the original aes_ccm.rs file, as it's used by somewhere else. There should be further transitions to only use the virtual_aes_ccm driver. 

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


capsules/src/virtual_aes_ccm.rs
```diff
@@ -0,0 +1,798 @@
...
+        self.aes.disable();
+    }
+
+    fn do_next_op(&self) {
+        if self.inflight.is_none() {
+            let mnode = self.clients.iter().find(|node| node.queued_up.is_some());
+            mnode.map(|node| {
+                self.inflight.set(node);
+                let parameters: CryptFunctionParameters = node.queued_up.take().unwrap();
+                let (res, _) = node.crypt_r(
```
niklasad1, line 165: would be nicer to just pass/borrow `CryptParameters` instead of each of its entries AFAIU.

capsules/src/virtual_aes_ccm.rs
```diff
@@ -0,0 +1,798 @@
...
+enum CCMState {
+    Idle,
+    Auth,
+    Encrypt,
+}
+
+// to cache up the function parameters of the crypt() function
+struct CryptFunctionParameters {
+    buf: &'static mut [u8],
+    a_off: usize,
```
niklasad1, line 100: please explicit variable names, `a_off` is hard to understand I think `authdata_offset` is better or something similar.

capsules/src/virtual_aes_ccm.rs
```diff
@@ -0,0 +1,798 @@
...
+                    parameters.m_off,
+                    parameters.m_len,
+                    parameters.mic_len,
+                    parameters.confidential,
+                    parameters.encrypting,
+                );
+                // notice that we didn't put the parameters back...
+
+                if res != ReturnCode::SUCCESS {
+                    // notify the client that there's a failure
```
niklasad1, line 177: Do we want to indicate error/debug log if `buf` or `client` is None?

capsules/src/virtual_aes_ccm.rs
Ouyancheng: Ah sorry, it was originally called a_off in aes_ccm.rs, I don't know what they exactly mean TBH... sorry...

capsules/src/virtual_aes_ccm.rs
Ouyancheng: added checks for buf / client being None in recent commit.

capsules/src/virtual_aes_ccm.rs
Ouyancheng: Yeah you are right! Thanks for pointing out!

Ouyancheng: resolved in recent commit~

capsules/src/virtual_aes_ccm.rs
Ouyancheng: So do you want me to expand those names? I am not very sure about them and there's a risk of mis-interpretation... Sorry for this... The code is from aes_ccm.rs.

niklasad1: Let's hear what the others think but if no one understands what they mean that's a problem :)

Could be maybe fixed in another PR then.

Ouyancheng: Thanks Niklas!

capsules/src/virtual_aes_ccm.rs
```diff
@@ -0,0 +1,802 @@
...
+                    }
+                    if node.buf.is_none() {
+                        debug!("virtual_aes_ccm: no buffer is binded with VirtualAES128CCM");
+                    }
+                    // notify the client that there's a failure
+                    node.buf.take().map(|buf| {
+                        node.crypt_client.map(move |client| {
+                            client.crypt_done(buf, res, false);
+                        });
+                    });
```
bradjc, line 183: Unfortunately it is problematic to have an error callback here. The issue is that when the user calls `crypt()` and nothing else is using AES, do_next_op runs immediately. If there is an error, this callback is called synchronously. However, tock requires that callbacks _only_ be generated stemming from interrupts.

The general way to solve this is to use a deferred call in `crypt()` so that do_next_op is only ever called from a callback context. The virtual_uart driver illustrates this.

bradjc (commented):
I think this should either be merged with the existing aes_ccm capsule, or replace it. The virtualizer layer shouldn't have to duplicate all of the logic from the aes_ccm capsule.

capsules/src/virtual_aes_ccm.rs
Ouyancheng: Thanks Brad, I've now modified it to use dynamic deferred call to call `do_next_op()` in `crypt()`

Ouyancheng: > I think this should either be merged with the existing aes_ccm capsule, or replace it. The virtualizer layer shouldn't have to duplicate all of the logic from the aes_ccm capsule.

Yeah, there should be further transitions to use the virtualized driver. But currently I couldn't perform such transition as aes_ccm.rs is used somewhere else. Probably we should do the transition later. Anyways, the virtual_aes_ccm is also functional.

hudson-ayers:
I pushed a commit that changes all uses of the old non-virtualized AESCCM to use the new, virtualized one.

Ouyancheng:
> I pushed a commit that changes all uses of the old non-virtualized AESCCM to use the new, virtualized one.

Thanks a lot Hudson!!!!!!

hudson-ayers (changes requested):
Mostly small changes

capsules/src/virtual_aes_ccm.rs
```diff
@@ -40,39 +40,219 @@
...
+    }
+}
+
+pub struct MuxAES128CCM<'a, A: AES128<'a> + AES128Ctr + AES128CBC> {
     aes: &'a A,
+    clients: List<'a, VirtualAES128CCM<'a, A>>,
+    inflight: OptionalCell<&'a VirtualAES128CCM<'a, A>>,
+    deferred_caller: &'a DynamicDeferredCall,
+    handle: OptionalCell<DeferredCallHandle>,
+    // debug_counter: Cell<i32>,
```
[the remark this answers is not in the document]
Ouyancheng: debug stuffs are removed in recent commit.

capsules/src/virtual_aes_ccm.rs
```diff
@@ -40,39 +40,219 @@
...
+};
+use kernel::common::{List, ListLink, ListNode};
+use kernel::debug;
 use kernel::hil::symmetric_encryption;
 use kernel::hil::symmetric_encryption::{
     AES128Ctr, AES128, AES128CBC, AES128_BLOCK_SIZE, AES128_KEY_SIZE, CCM_NONCE_LENGTH,
 };
 use kernel::ReturnCode;
 
+// the CCM state is not public is aes_ccm.rs
```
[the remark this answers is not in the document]
Ouyancheng: Yep, removed in recent commit.

hudson-ayers (approved):
This PR has the same issue as several other virtualizers that certain types of errors are replaced with debug statements, because there is complexity required to ensure those errors are reported via callbacks, and that those callbacks are only called in callback context. I am not sure it is fair to block this PR on that issue when it exists in many of the other virtual capsules.

hudson-ayers:
This does need to be rebased on master before it can be merged

hudson-ayers:
Rebased/squashed onto master (#2069 caused several conflicts)

hudson-ayers:
I would appreciate you running the AES hardware tests again to confirm I didn't break anything

Ouyancheng:
> I would appreciate you running the AES hardware tests again to confirm I didn't break anything

virtual_aes_ccm_test.rs tests passed!

hudson-ayers (approved):

boards/imix/src/test/virtual_aes_ccm_test.rs
```diff
@@ -0,0 +1,105 @@
...
+//! aes_ccm_test passed: (current_test=0, encrypting=true, tag_is_valid=true)
+//! aes_ccm_test passed: (current_test=0, encrypting=false, tag_is_valid=true)
+//! aes_ccm_test passed: (current_test=1, encrypting=true, tag_is_valid=true)
+//! aes_ccm_test passed: (current_test=1, encrypting=false, tag_is_valid=true)
+//! aes_ccm_test passed: (current_test=2, encrypting=true, tag_is_valid=true)
+//! aes_ccm_test passed: (current_test=2, encrypting=false, tag_is_valid=true)
+use capsules::test::aes_ccm::Test;
+use capsules::virtual_aes_ccm;
+use kernel::common::dynamic_deferred_call::DynamicDeferredCall;
+// use kernel::debug;
```
alistair23, line 33: Can we remove this commented out code?

boards/imix/src/test/virtual_aes_ccm_test.rs
```diff
@@ -0,0 +1,105 @@
+//! To run this test, include the code
+//! ```
```
alistair23, line 2: Can this be ` ```rust ` instead?

boards/imix/src/test/virtual_aes_ccm_test.rs
```diff
@@ -0,0 +1,105 @@
...
+    t1.run();
+    t2.run();
+    t3.run();
+    // notice that there will be 18 tests, 6 for each,
+    // and the test output will make the debug buffer full,
+    // please go to boards/components/src/debug_writer.rs and change
+    //     let buf = static_init!([u8; 1024], [0; 1024]);
+    // to
+    //     let buf = static_init!([u8; 4096], [0; 4096]);
+    // Thanks!
```
alistair23, line 104: Can you put this note in the header instead?

capsules/src/virtual_aes_ccm.rs
```diff
@@ -71,8 +102,153 @@ enum CCMState {
...
     aes: &'a A,
+    clients: List<'a, VirtualAES128CCM<'a, A>>,
+    inflight: OptionalCell<&'a VirtualAES128CCM<'a, A>>,
+    deferred_caller: &'a DynamicDeferredCall,
+    handle: OptionalCell<DeferredCallHandle>,
+}
+
+impl<'a, A: AES128<'a> + AES128Ctr + AES128CBC> MuxAES128CCM<'a, A> {
+    pub fn new(aes: &'a A, deferred_caller: &'a DynamicDeferredCall) -> MuxAES128CCM<'a, A> {
+        // aes.enable(); // enable the hardware
```
alistair23, line 148: Why is this commented out?

Ouyancheng: Hi Alistair, 
The device should be globally enabled in the board's related file instead of here. It's same semantic for the original AES CCM driver. 
Regards, 
Yancheng Ou

boards/imix/src/test/virtual_aes_ccm_test.rs
Ouyancheng: resolved in recent commit

boards/imix/src/test/virtual_aes_ccm_test.rs
Ouyancheng: resolved in recent commit

boards/imix/src/test/virtual_aes_ccm_test.rs
Ouyancheng: resolved in recent commit

capsules/src/virtual_aes_ccm.rs
hudson-ayers: I think this line should either be deleted or commented in. I think it makes the most sense to just enable it here, even if the original driver did not do it that way. That helps prevent board authors from forgetting the enable call

bradjc (approved):

capsules/src/virtual_aes_ccm.rs
Ouyancheng: OK~ then let me make another commit and uncomment the line~ 
Sorry for this... 

Regards, 
Yancheng Ou

Ouyancheng: It's uncommented in recent commit. And the tests are passed. 

Regards, 
Yancheng Ou

alistair23 (approved):

bradjc (approved):

hudson-ayers:
I am very confused as to how that single line change broke all those tests in a seemingly unrelated area of code

Ouyancheng:
> I am very confused as to how that single line change broke all those tests in a seemingly unrelated area of code

I guess it's because Nano33BLE uncommented its wireless component, which is not fully compatible with the virtualized code.

hudson-ayers:
Ah, okay. I'll rebase.

Ouyancheng:
> Ah, okay. I'll rebase.

Thanks Hudson!

hudson-ayers (approved):

PR #2197, opened 2020-11-13
tock-registers: accept trailing comma in bitfields and bitmasks

lschuermann:
### Pull Request Overview

This pull request changes the macro rules to accept one trailing comma at the end of the `register_bitfields` and `register_bitmasks`-macro variadic arguments interfaces.

This is not a deal breaker, but an error I've done countless number of times. (For me) It's intuitive to write the trailing comma in lists which span multiple lines and it's generally valid in Rust syntax. Also, having a trailing comma prevents the macro from expanding, which Rusts treats as if it wasn't called, so the compiler output is flooded with subsequent errors. Changing the syntax to accept the comma seems appropriate.

With this change, the syntax is accepted:
```
register_bitfields![u32,
    TESTREG [
        TESTBIT OFFSET(7) NUMBITS(1) [
            TestStateA = 0,
            TestStateB = 1,
        ],
    ],
];
```

This pull request was tested by compiling with trailing commas.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/libraries/tock-register-interface/src/macros.rs
+++ b/libraries/tock-register-interface/src/macros.rs
@@ -15,7 +15,7 @@ macro_rules! register_bitmasks {
         // BITFIELD_NAME OFFSET(x)
         $(#[$outer:meta])*
         $valtype:ident, $reg_desc:ident, [
-            $( $(#[$inner:meta])* $field:ident OFFSET($offset:expr)),+
+            $( $(#[$inner:meta])* $field:ident OFFSET($offset:expr)),+ $(,)?
         ]
     } => {
         $(#[$outer])*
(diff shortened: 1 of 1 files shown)


alistair23 (approved):

hudson-ayers (approved):
In retrospect it probably would have been nice to do this to most of the component macros as well, thanks for this nice change.

bradjc (approved):

lschuermann:
> In retrospect it probably would have been nice to do this to most of the component macros as well, thanks for this nice change.

I agree. I'll happily volunteer to do this!

PR #2175, opened 2020-10-26
rust: update to 2020-10-25

bradjc:
### Pull Request Overview

This pull request updates the rust nightly version to 2020-10-25.

I had to add a feature in two places. I have no idea why or what changed, but the compiler instructed me to add them.

Two issues:

```
error[E0658]: mutable references are not allowed in constant functions
  --> chips/lowrisc/src/hmac.rs:87:21
   |
87 |             digest: Cell::new(None),
   |                     ^^^^^^^^^^^^^^^
   |
   = note: see issue #57349 <https://github.com/rust-lang/rust/issues/57349> for more information
   = help: add `#![feature(const_mut_refs)]` to the crate attributes to enable

error[E0658]: mutable references are not allowed in constant functions
  --> chips/lowrisc/src/hmac.rs:87:31
   |
87 |             digest: Cell::new(None),
   |                               ^^^^
   |
   = note: see issue #57349 <https://github.com/rust-lang/rust/issues/57349> for more information
   = help: add `#![feature(const_mut_refs)]` to the crate attributes to enable

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0658`.
error: could not compile `lowrisc`
```

and 
```
error[E0658]: mutable references are not allowed in constant functions
  --> libraries/tock-cells/src/take_cell.rs:21:18
   |
21 |             val: Cell::new(None),
   |                  ^^^^^^^^^^^^^^^
   |
   = note: see issue #57349 <https://github.com/rust-lang/rust/issues/57349> for more information
   = help: add `#![feature(const_mut_refs)]` to the crate attributes to enable

error[E0658]: mutable references are not allowed in constant functions
  --> libraries/tock-cells/src/take_cell.rs:21:28
   |
21 |             val: Cell::new(None),
   |                            ^^^^
   |
   = note: see issue #57349 <https://github.com/rust-lang/rust/issues/57349> for more information
   = help: add `#![feature(const_mut_refs)]` to the crate attributes to enable

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0658`.
error: could not compile `tock-cells`
```

### Testing Strategy

This pull request was tested by travis.

### TODO or Help Wanted

n/a

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make prepush`.

PR #2198, opened 2020-11-14
build: handle missing tty gracefully

ppannuto:
### Pull Request Overview

This should hopefully address the recent netflify build update.

### Testing Strategy

You're looking at it.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/boards/components/src/hmac.rs
+++ b/boards/components/src/hmac.rs
@@ -2,7 +2,7 @@
 //!
 //! Usage
 //! -----
-//! ```rust
+//! ```ignore
 //!    let hmac_data_buffer = static_init!([u8; 64], [0; 64]);
 //!    let hmac_dest_buffer = static_init!([u8; 32], [0; 32]);
 //!
@@ -20,7 +20,6 @@
 //!        lowrisc::hmac::Hmac,
 //!        [u8; 32]
 //!    ));
-//! ));
 //! ```
 
 use capsules;
--- a/boards/components/src/hmac.rs
+++ b/boards/components/src/hmac.rs
@@ -2,7 +2,7 @@
 //!
 //! Usage
 //! -----
-//! ```ignore
+//! ```rust
 //!    let hmac_data_buffer = static_init!([u8; 64], [0; 64]);
 //!    let hmac_dest_buffer = static_init!([u8; 32], [0; 32]);
 //!
--- a/boards/Makefile.common
+++ b/boards/Makefile.common
@@ -245,7 +245,7 @@ debug-lst:  $(TARGET_PATH)/debug/$(PLATFORM).lst
 doc: | target
 	@# This mess is all to work around rustdoc giving no way to return an
 	@# error if there are warnings. This effectively simulates that.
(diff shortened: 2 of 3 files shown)


hudson-ayers:
still no luck with netlify it seems

hudson-ayers:
I think the tty issue is a red herring on what is actually wrong. The tty errors were printed even back when the build passed (see https://app.netlify.com/sites/docs-tockosorg/deploys/5f9ad29ca1bd4c0007bed3fd for example).

The netlify build stopped passing on master on the commit where #2175 was merged (we missed this because we thought it was just because the qemu site was down and merged manually, but now I am not so sure).

Another PR that seems to have caused some issues is #2040 , which updated the QEMU submodule. I say this because the build log shows failures for that PR even when other PRs were passing.

Also, interestingly, clearing the netlify cache partially fixes a single build (it gets further along before failing), but you have to reclear the cache before every new commit to see the benefit. So there seem to be two significant issues -- one related to the cache, and the other possibly related to the PR that updated Rust somehow causing builds to timeout even when the cache issue is temporarily fixed.

hudson-ayers:
okay, now I'm getting close! Updating Rust means a new version of rustdoc, and the new version of rustdoc is better at generating warnings -- it found 2 warnings which did not previously exist. Unfortunately, those terrifying bash lines meant to detect warnings and error instead seem to detect warnings and hang when run on the netlify hosted machines. Why? No idea.

But with those warnings fixed, netlify now passes, so long as you manually rebuild the cache. In fact, now that it passes, I am not sure if rebuilding the cache is needed for subsequent builds

alistair23:
WooHoo!

boards/components/src/hmac.rs
```diff
@@ -2,7 +2,7 @@
 //!
 //! Usage
 //! -----
-//! ```rust
+//! ```ignore
```
alistair23, line 5: I think it should be `rust,ignore`

hudson-ayers: it should be just rust, that was a leftover from earlier debugging

hudson-ayers:
For posterity:

this problem has probably existed for awhile, but it only manifests when there are warnings when running `rustdoc`, so we got by not encountering it. 

A minimal reproducible case I could find was this:
```bash
cargo doc 2>&1 | tee /dev/stderr | grep --quiet warning
```
which hangs forever when the output of `cargo doc` includes a warning.

Interestingly, both
```bash
cargo doc 2>&1 | grep --quiet warning
```
and
```bash
cargo doc 2>&1 | tee /dev/stderr | grep warning
```
work correctly in the presence of warnings.

Notably, per `man grep`:
```
-q, --quiet, --silent
      Quiet; do not write anything to standard output.  Exit immediately with zero status if any match is found, even if an error was  detected. 
```

The issue is clearly related to the fact that `grep -q` exits early, closing a pipe when preceding commands are still running.

By monitoring the output of `ps -e` before and after the faulty command hangs, I can see that both the `tee` and `grep` processes do terminate. `cargo doc`, however, does not. So this seems to be some sort of issue with `cargo doc` not correctly handling when its `stdout` or `stderr` is closed. But I fail to understand why this issue only manifests when `tee` is being used between `cargo doc` and the failing command.

My solution was as follows:

1. tee to `/dev/stderr` instead of `/dev/tty`, as there is no `/dev/tty` on the netlify CI machines, so we can't actually see the warnings produced by `cargo doc`  EDIT: It seems this still doesn't work, there is a permissions error. AFAICT this is a docker issue.

2. Rather than use `grep -q`, which exits early, use `grep >> /dev/null`, to avoid issues with `cargo doc` incorrectly handling closed outputs.

3. Fix the `cargo doc` warnings

hudson-ayers (approved):
I think this is ready for merge

ppannuto (commented):
You are a hero for tracking this down. Ironically, I can't approve this because I'm the original PR author, but it looks 👍 to me.

alistair23 (approved):

bradjc (approved):

PR #2200, opened 2020-11-15
Add Teensy 4 board

mciantyre:
### Pull Request Overview

The PR proposes Tock support for the Teensy 4. As of this PR, the Tock kernel, and a variety of small examples, can run on [Teensy 4.0][teensy40] and [Teensy 4.1][teensy41] development boards. The PR builds on the i.MX RT chip foundation added in #1918.

[teensy40]: https://www.pjrc.com/store/teensy40.html
[teensy41]: https://www.pjrc.com/store/teensy41.html

The board makes use of

- the LED on pin 13
- UART2 on pins 14 and 15
- GPT1 as an alarm

The Teensy 4 boards use i.MX RT **1062** chips. Given our current chip features, the 1060 chip family is identical to the 1050 chip family. Before integrating the Teensy 4 board, I *renamed `chips/imxrt1050` to `chips/imxrt10xx`*. I updated the `imxrt1050-evkb` board to use the renamed crate.

Additional changes to the i.MX RT chip crate include

- adding a LPUART2 peripheral
- renaming the `gpt1` module to `gpt`, and adding `GPT2`
- supporting periodic clock selection, allowing a user to select the static crystal oscillator as the GPT clock source

Changes to the chip crate should be backwards compatible for `imxrt1050-evkb` users. Let me know if we see an issue.

### Testing Strategy

I tested the PR by running `blink` and `console` libtock-c examples on both a Teensy 4.0 and 4.1 board. I repackaged the examples [here](https://github.com/mciantyre/tock-teensy4-apps).

The PR was **not** tested on an NXP i.MX RT 1050 evaluation board; I don't have that hardware.

### TODO or Help Wanted

This pull request may be tested on a 1050 evaluation board. @valexandru, if you have an opportunity to test this work and review these changes, I'd appreciate it!

This PR does not address any of the TODOs noted in #1918. In particular, the `imxrt10xx` chip still does not use the new peripheral instantiation approach (#2069). If this is still TODO and not urgent, I'm happy to support that cleanup in a separate PR.

This Teensy 4 support was based on a different chip implementation. That chip implementation supported DMA, and a UART driver that used DMA. If we see value in a DMA driver for i.MX RT chips, I'm happy to contribute the driver.

### Documentation Updated

- [x] ~~Updated the relevant files in `/docs`~~, or **no updates are required**

I've added documentation in `boards/teensy4`. The documentation

- lists the tools that you need to program a Teensy 4
- how to build the kernel and apps
- how to convert the program to a HEX file, which is necessary to program a board

### Formatting

- [x] Ran `make prepush`.

### New Platform Checklist

- [x] Hardware is widely available.
- [x] I can support the platform, which includes release testing for the platform, at least initially.
  - [x] `Console`, including `debug!()` and userspace `printf()`.
  - [x] Timers.
  - [x] GPIO with interrupts.

I assume the basic chip features were provided by #1918. Let me know if we need to make all three features available through the board. As of this writing, the Teensy 4 board does not expose an input GPIO that responds to an interrupt.


boards/teensy4/Makefile
```diff
@@ -0,0 +1,32 @@
...
+
+# Unsupported; there's no easily-accessible JTAG interface
+.PHONY: flash
+flash:
+	echo "Use 'make program' to program the Teensy 4"
+	exit 1
+
+# For testing with a blinky LED
+BLINK=../../../libtock-c/examples/blink/build/cortex-m7/cortex-m7.tbf
+.PHONY: program
```
ppannuto, line 25:
```suggestion
```

kernel/src/hil/time.rs
```diff
@@ -282,15 +282,6 @@ impl Frequency for Freq16MHz {
...
 
-/// 24.75MHz `Frequency`
-#[derive(Debug)]
-pub struct Freq2475MHz;
-impl Frequency for Freq2475MHz {
-    fn frequency() -> u32 {
-        24750000
-    }
-}
-
```
ppannuto, line 293: Why is this being deleted?

boards/teensy4/layout.ld
```diff
@@ -0,0 +1,73 @@
...
+    /* End of iMXRT10xx magic
+     *
+     * Need to align vector table on a 1024-byte boundary given the size of the table
+     */
+    . = ALIGN(1024);
+  } > rom
+}
+
+INCLUDE ../kernel_layout.ld
+ENTRY(_ivt);
```
ppannuto, line 73: Interesting -- is this actually used by the tooling?

ppannuto (commented):
I looked over most of this this evening, and this generally looks fantastic. I really appreciate the detailed comments through some of the more complex bits; this was really quite smooth to review. The fancy boot sector is interesting, haven't seen anything like that before.

The README for the board is great. It'd be nice to maybe have a few less manual steps and some supporting make targets that do that work for you, but that's not a blocker by any means.

Only real nit is the possibly spurious deletion of a frequency from the time HIL?

boards/teensy4/README.md
```diff
@@ -0,0 +1,74 @@
+Teensy 4 Development Board
+==========================
+
+The `teensy4` board supports both the Teensy 4.0 and Teensy 4.1 form factors.
+For more information, visit the product pages:
+
+- [Teensy 4.0 Development Board](https://www.pjrc.com/store/teensy40.html)
+- [Teensy 4.1 Development Board](https://www.pjrc.com/store/teensy41.html)
```
bradjc, line 8: It looks like (based on https://www.pjrc.com/store/teensy41.html) these two boards vary quite a bit. Can we choose just one that this has been tested on? Otherwise it's not clear how the linker file should be set, or if it is ok to add an sdcard driver.

bradjc (commented):
Overall looks good, but I'm a little wary of having a general board crate rather than being specific about exactly what hardware was used and tested.

hudson-ayers:
> This PR does not address any of the TODOs noted in #1918. In particular, the imxrt10xx chip still does not use the new peripheral instantiation approach (#2069). If this is still TODO and not urgent, I'm happy to support that cleanup in a separate PR.

That transition is still TODO and is not urgent, this can be merged first. And if you are able to support that cleanup in a separate PR I would really appreciate it :)

kernel/src/hil/time.rs
mciantyre: #1918 introduced `Freq2475MHz`. But after this PR, it's unused. It's replaced by [a frequency value computed at runtime][freq-rt-diff], based on an input clock selection and divider.

I thought this removal, plus the change to compute clock frequency, would do for our current and future goals:

- It let me express the original author's intent -- run the general purpose timers (GPT) on the IPG clock -- while also supporting my goal: run the GPT on the oscillator clock, with root- and peripheral-level clock dividers<sup>*</sup>.
- From the original author's notes, this frequency would eventually change once we configured the ARM clock (which generates the IPG clock):

https://github.com/tock/tock/blob/d695df5223204fe0cf102ccffe097cc208efcc42/chips/imxrt1050/src/gpt1.rs#L247-L253

Although the runtime frequency computation works, I'm looking for other suggestions to solve a perceived _"how to I set a chip's clock frequency, based on the board?"_ problem. It seems I'm the first to propose a `Frequency::frequency()` implementation that doesn't return a constant value, and I'm not sure I want to set that example. Just want to make sure that it's OK to go with this approach.

<sup>* I prefer the 24MHz oscillator clock source since it's not affected by varing processor speeds, it's consistent across all i.MX RT processors, and it's fast enough for me.</sup>

[freq-rt-diff]: https://github.com/tock/tock/pull/2200/files#diff-265db427da6c50259e02f26382d7804a971db41f4b4023ad0c794ce982c80072R318-R333

boards/teensy4/layout.ld
mciantyre: Nope! Just a leftover from some earlier prototyping.

I'll remove it. There's some other comments in this file that need correction.

boards/teensy4/README.md
mciantyre: I updated the README to state that we guarantee support for the Teensy **4.0** board. The `teensy4` crate will never include features that are only for the 4.1 board. I've left more thoughts in f981a847's commit message.

Would you prefer to rename the crate to `teensy40` as a stronger support signal? Or, could there be a future where the `teensy4` board supports the common functions between 4.0 and 4.1, and we use cargo feature flags to enable 4.1 features?

bradjc (approved):

boards/teensy4/README.md
hudson-ayers: Typically, Tock avoids cargo feature flags for enabling features specific to a particular chip within a family of supported chips. Instead, Tock structures modules in a specific way so that more specific chips inherit from a library of shared functionality, re-exporting from the shared library all functionality that is unchanged by the more specific chip. For the best example of this, see `chips/nrf5x', `chips/nrf52`, and `chips/nrf52840`.

Given that you currently don't have any features only supported by one chip and not another, and have only tested on a teensy 4.0, I think it is best to name the chip crate `teensy40`. We can revisit that if explicit, tested support for the teensy 4.1 is added later.

hudson-ayers (approved):

bradjc (approved):

bradjc (approved):

PR #2202, opened 2020-11-17
Virtual RNG

nirvikbaruah:
### Pull Request Overview

This pull request implements the virtualisation of the current rng capsule, which allows multiple instances of the rng to request randomness.

### Testing Strategy

Created a test file which allows 3 clients to request randomness at the same time.

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


boards/nano33ble/src/main.rs
```diff
@@ -18,6 +18,7 @@ use kernel::hil::gpio::Interrupt;
 use kernel::hil::gpio::Output;
 use kernel::hil::i2c::I2CMaster;
 use kernel::hil::time::Counter;
+
```
bradjc, line 21:
```suggestion
```

boards/nano33ble/src/main.rs
```diff
@@ -180,7 +181,6 @@ pub unsafe fn reset_handler() {
         (&nrf52840::gpio::PORT[LED_BLUE_PIN], ActiveLow)
     ))
     .finalize(components::led_component_buf!(nrf52840::gpio::GPIOPin));
```
bradjc, line 183:
```suggestion
    .finalize(components::led_component_buf!(nrf52840::gpio::GPIOPin));

```

capsules/src/virtual_rng.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+impl<'a, R: Rng<'a>> Rng<'a> for VirtualRngMasterDevice<'a, R> {
+    fn get(&self) -> ReturnCode {
+        self.operation.set(Op::Get);
+        self.mux.do_next_op()
+    }
+
+    fn cancel(&self) -> ReturnCode {
+        self.mux.rng.cancel()
```
bradjc, line 111: What prevents one client from cancelling another client's RNG read?

capsules/src/virtual_rng.rs
```diff
@@ -0,0 +1,130 @@
...
+                .find(|node| node.operation.get() != Op::Idle);
+            mnode.map(|node| {
+                let op = node.operation.get();
+                // Need to set idle here in case callback changes state
+                node.operation.set(Op::Idle);
+                match op {
+                    Op::SetClient => {
+                        self.rng.set_client(node);
+                        ReturnCode::SUCCESS
+                    }
```
bradjc, line 46: What ensures that the correct client is set when the `Op::Get` executes?

capsules/src/virtual_rng.rs
nirvikbaruah: Clarified my logic with @hudson-ayers and modified `cancel()` to account for the current inflight device :)

capsules/src/virtual_rng.rs
nirvikbaruah: Just modified the implementation for `Op::Get` to set the correct client prior to every call

capsules/src/virtual_rng.rs
```diff
@@ -0,0 +1,167 @@
...
+
+                match op {
+                    Op::Get => {
+                        let client = node.client.take();
+                        match client {
+                            Some(p) => {
+                                // Set rng client to current node
+                                self.rng.set_client(p);
+                                node.client.insert(client);
+                                self.rng.get()
```
hudson-ayers, line 52: we should only set the current node as `inflight` if `get()` returns `SUCCESS`. Otherwise no callback will be generated and the RNG will become unusable.

hudson-ayers (changes requested):

capsules/src/test/virtual_rng.rs
```diff
@@ -0,0 +1,30 @@
+// Test file for virtual_rng
+// Currently tests out getting rng sequentially
```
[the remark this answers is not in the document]
lschuermann: I agree. We could also add a check that a random sequence shouldn't repeat or given out to multiple clients. If we want something more deterministic, we could have a non-random test RNG, over which we could verify that the sequences are actually uniquely distributed to the clients.

Depends on how far we're willing to go, but a basic test with two clients seems easy to implement and better tests the virtualization layer.

bradjc (dismissed):

capsules/src/test/virtual_rng.rs
```diff
@@ -0,0 +1,30 @@
...
+    pub fn new(rng: &'a VirtualRngMasterDevice<'a, R>) -> TestRng<'a, R> {
+        debug!("Initialized virtual_rng tester");
+        TestRng { rng: rng }
+    }
+
+    pub fn run(&self) {
+        debug!("Starting virtual_rng get tests:");
+        for x in 1..ELEMENTS {
+            match self.rng.get() {
+                ReturnCode::SUCCESS => debug!("virtual_rng test: get {} SUCCESS", x),
```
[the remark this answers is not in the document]
phil-levis: Yes, we want to actually check the right callbacks are called, they get randomnness appropriately, etc.

capsules/src/virtual_rng.rs
```diff
@@ -0,0 +1,175 @@
...
+#[derive(Copy, Clone, PartialEq)]
+enum Op {
+    Idle,
+    Get,
+}
+// TODO: REMOVE SIZED
+
+// Struct to manage multiple rng requests
+pub struct MuxRngMaster<'a, R: Rng<'a> + ?Sized> {
+    rng: &'a dyn Rng<'a>,
```
hudson-ayers, line 17:
```suggestion
pub struct MuxRngMaster<'a, R: Rng<'a>> {
    rng: &'a R,
```

capsules/src/virtual_rng.rs
```diff
@@ -0,0 +1,175 @@
...
+                device.get();
+            }
+
+            cont_code
+        })
+    }
+}
+
+// Struct for a single rng device
+pub struct VirtualRngMasterDevice<'a, R: Rng<'a> + ?Sized> {
```
hudson-ayers, line 94:
```suggestion
pub struct VirtualRngMasterDevice<'a, R: Rng<'a>> {
```

boards/nano33ble/src/test/virtual_rng_test.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    // Create virtual rng mux device
+    let mux = static_init!(
+        capsules::virtual_rng::MuxRngMaster<'static, dyn kernel::hil::rng::Rng<'static>>,
+        capsules::virtual_rng::MuxRngMaster::new(rng_obj)
+    );
+
+    // Create all devices for the virtual rng
+    let device1 = static_init!(
+        capsules::virtual_rng::VirtualRngMasterDevice<'static, dyn kernel::hil::rng::Rng<'static>>,
```
hudson-ayers, line 27:
```suggestion
        capsules::virtual_rng::VirtualRngMasterDevice<'static, nrf52::trng::Trng>,
```

hudson-ayers (commented):
This currently has an unusual mix of trait objects and generics. I think you should either use only trait objects, or only generics. I included a few suggested changes that I think show how you could move to a generics-only approach, and not rely on `?sized`

capsules/src/virtual_rng.rs
```diff
@@ -0,0 +1,169 @@
...
+        self.mux.inflight.map_or_else(
+            || {
+                // If no node inflight, just set node to idle and return
+                self.operation.set(Op::Idle);
+                ReturnCode::SUCCESS
+            },
+            |current_node| {
+                // Find if current device is the one in flight or not
+                if *current_node == self {
+                    self.mux.rng.cancel()
```
hudson-ayers, line 144: Shouldn't this branch also call `self.operations.set(Op::Idle)`?

capsules/src/virtual_rng.rs
```diff
@@ -0,0 +1,169 @@
...
+        self.mux.devices.push_head(&self);
+        self.client.set(client);
+    }
+}
+
+impl<'a> Client for VirtualRngMasterDevice<'a> {
+    fn randomness_available(
+        &self,
+        randomness: &mut dyn Iterator<Item = u32>,
+        _error: ReturnCode,
```
hudson-ayers, line 163: parameter names should only be prefixed with an underscore when they are not used in the function. error is used, so it should just be named `error` rather than `_error`

hudson-ayers (changes requested):
This is looking very close! Probably the most complicated thing remaining is the rebase -- check out https://github.com/tock/tock/blob/master/.github/CONTRIBUTING.md#step-4-rebase for instructions.

capsules/src/virtual_rng.rs
```diff
@@ -0,0 +1,169 @@
...
+    ) -> Continue {
+        // Try find if randomness is available, or return done
+        self.inflight.take().map_or(Continue::Done, |device| {
+            self.do_next_op();
+            let cont_code = device.randomness_available(_randomness, _error);
+
+            // Get more randomness if requested
+            if cont_code == Continue::More {
+                device.get();
+            }
```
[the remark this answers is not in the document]
nirvikbaruah: From my test files, this code is necessary for more randomness to be generated from a client. From my understanding, when the client returns `Continue::More`, the virtualizer should request for more randomness and return it. This doesn't seem to happen automatically, which is why an explicit call to `get()` is necessary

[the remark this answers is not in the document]
hudson-ayers: If so, I think that would indicate a bug in the `nrf5x/trng` driver. But looking at the driver code, it does seem to handle this case:
```rust
                    let result = client.entropy_available(&mut TrngIter(self), ReturnCode::SUCCESS);
                    if Continue::Done != result {
                        // need more randomness i.e generate more randomness
                        self.start_rng();
                    }
```

I don't think we should call `get()` here. Also, I am realizing that there are potentially complex interactions between `Continue::More` and `do_next_op()`. What if `do_next_op()` starts another `get()`? Which client should receive the eventual callback -- the one that returned `Continue::More` or the one that was scheduled in the call to `do_next_op()`?

I see two possible solutions to this:
1. Move the call to `do_next_op()` back to after the client has a chance to request more randomness, and only call `do_next_op()` if no randomness is requested. This is the simplest solution, but makes starvation of other clients trivial.

2. Don't propagate the Continue code back to the underlying hardware RNG driver -- instead, cache it as a new get() request being made by the client.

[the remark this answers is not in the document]
nirvikbaruah: That makes sense - I've implemented your first solution and modified my test files accordingly.

hudson-ayers (approved):
Do the tests pass now? If so, I think this looks good.

nirvikbaruah:
> Do the tests pass now? If so, I think this looks good. 

Yes, tests pass now @hudson-ayers

PR #2203, opened 2020-11-18
Port to LiteX+VexRiscv-based SoCs

lschuermann:
### Pull Request Overview

This PR introduces support for LiteX SoCs built around a VexRiscv CPU. [Timer](https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/timer.py), [UART](https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/uart.py), [LED](https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/led.py) and [LiteEth](https://github.com/enjoy-digital/liteeth) cores are currently supported. The `litex_sim` and `litex_arty` boards target the default [`litex_sim.py`](https://github.com/enjoy-digital/litex/blob/master/litex/tools/litex_sim.py) and [`arty.py`](https://github.com/enjoy-digital/litex/blob/master/litex/boards/targets/arty.py) SoC definitions of a [specific release of LiteX](https://github.com/enjoy-digital/litex/tree/444a605deae6a561dbe2c49bf3062eae6f3cd887) respectively.

- `litex_arty` targets the [Digilent Arty-A7 FPGA development board](https://reference.digilentinc.com/reference/programmable-logic/arty-a7/start)
  A working bitstream (`arty.bit`) is provided in a companion repository, [tock-litex-support](https://git.currently.online/~leons/tock-litex-support/tree/master/generated/arty/gateware).
- `litex_sim` is a verilated simulation of a LiteX SoC, with UART, Timer and Ethernet cores

#### LiteX

> LiteX is a Migen/MiSoC based Core/SoC builder that provides the infrastructure to easily create Cores/SoCs (with or without CPU).

By its nature, LiteX supports generating a plethora of different SoCs. This PR implements drivers for a selection of LiteX cores (Timer, UART, LED, Ethernet). Naturally, these cores - being described in the hardware description language Migen - can be changed to break compatibility with the drivers in Tock. The idea is to pin, per board in the README, a specific version of LiteX and parameters for generating the SoC definition with which Tocks is compatible. Variations in the design _should_ be reflected in Tock by adjustable constants, types and other compile time options wherever it makes sense.

#### Ethernet

Both the Arty A7 board and the verilated simulation feature an Ethernet MAC, which connects to an on-board PHY or a `tap` network interface respectively. In the near future, I will create a PR for VirtIO-based network interfaces. This deliberately does not try to define a HIL yet, as at least two different implementations should be considered for the design. Hence the interface of the Ethernet MAC is subject to change.

#### `litex_registers.rs` / `litex-register-gen`

LiteX has the notion of _Configuration Status Registers_, which are not to be confused with RISC-V's CSRs. These registers are exposed on a Wishbone bus over the SoC. The CSRs can be generated with different data widths (for efficient routing on small FPGAs), aligment settings, bus widths, etc. `litex_registers.rs` abstracts these unconventional and varying layouts using structs named according to the scheme `<AccessTypes><RegisterWidth>C<CSRDataWidth>B<BaseWidth>`. Register abstraction types of a common layout configuration are grouped using the `LiteXSoCRegisterConfiguration` trait. Register structs can then be made generic over this trait. Refer to `chips/litex/src/timer.rs` for a usage example. [This comment](https://github.com/tock/tock/pull/2203#issuecomment-733955775) futher explains how the `LiteXSoCRegisterConfiguration` works.

~To avoid a massive amount of repetitive code, `litex_registers.rs` generates these types using a procedural macro defined in `libraries/litex-register-gen`. If this macro should not be included within Tock, I will moved it to the companion support repository and include the expanded Rust code instead.~ The procedural macro to generate the LiteX register types is removed from this PR and the generated output is statically placed into `litex_registers.rs` instead. It can be reviewed in a followup PR instead.

As generic associated types are an unstable Rust feature, the wrapper types `{Read,Write,ReadWrite}RegWrapper` are required to access bitfields of registers. As soon as this feature lands in stable Rust, the zero-cost wrappers can be removed.

### Testing Strategy

- running different userspace apps on both `litex_arty` and `litex_sim`
- implementing a small IPv6 ND/ping stack in the kernel and pinging both `litex_arty` and `litex_sim`

### TODO or Help Wanted

**VexRiscv is still missing PMP support upstream!** While there is [a pull request](https://github.com/SpinalHDL/VexRiscv/pull/147) for adding support, currently the boards lack memory protection entirely. This is expliclity documented in the respective board READMEs.

There are still some minor TODOs until this is ready
- [x] document the `litex_registers` and `litex-register-gen`
- [x] ~there is some weird behaviour with the scheduler timer, which needs to be further diagnosed~ `VirtualSchedulerTimer` issue identified to be independent of this PR, fix proposed in #2267 
- [x] in general, method documentation of the peripherals should be improved / added
- [x] check that 64-bit timer overflows are handled correctly in `LiteXAlarm` (overflows after >5000 years @100MHz)
- [ ] ~(the Ethernet MAC TX currently only supports a single slot -> non-blocking issue, Ethernet should still work fine)~

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### New Platform Checklist

- [X] Hardware is widely available: yes, and the verilated simulation does not require hardware
- [x] I can support the platform, which includes release testing for the platform, at least initially.
  - [X] `Console`, including `debug!()` and userspace `printf()`.
  - [X] Timers.
  - [X] ~GPIO with interrupts.~ LEDs work through a special LED driver core, LiteX does not include a GPIO core in the default target definitions


boards/litex_arty/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+build = "build.rs"
+edition = "2018"
+
+[dependencies]
+components = { path = "../components" }
+rv32i = { path = "../../arch/rv32i" }
+capsules = { path = "../../capsules" }
+kernel = { path = "../../kernel" }
+litex = { path = "../../chips/litex" }
+litex_vexriscv = { path = "../../chips/litex_vexriscv" }
```
hudson-ayers, line 14: For most boards in Tock that use a chip that depends on some collection of shared libraries (e.g. `nrf52840` depends on the shared libraries in `nrf52`), the more specific crate simply re-exports the portions of the shared crate that might be needed outside the more specific crate. (e.g. `chips/nrf52840` re-exports much of `chips/nrf52`). This allows for boards to depend only on the more specific crate, rather than on both the more specific and less specific crates. I have found that this design simplifies refactoring -- if nrf52840 previously used the generic nrf52 spi library, but is later updated to use an nrf52840-specific version, I do not have to modify every board to reflect the change.

Would it be a good idea to continue that approach here? Or is there some reason that doing so is impossible for the litex chips?

lschuermann: I wasn't aware of this approach, but it makes sense. I'll happily update it.

hudson-ayers: Thanks! Re: awareness, thats probably because the approach isn't documented anywhere yet. We probably should try to codify how we support chip families in Tock soon. When it was just nrf52 it didn't make sense, but now we have similar patterns for nrf52, stm32, lowrisc, imxrt, and soon litex.

lschuermann: I've updated the board to only depend on the `litex_vexriscv`-crate, which now exposes both the chip module (which is depending on the specific CPU architecture used, in our case VexRiscv) and the rest of the LiteX core drivers and support modules. Feel free to resolve if the current design matches the other chips as requested.

boards/litex_arty/Makefile
```diff
@@ -0,0 +1,7 @@
+# Makefile for building the tock kernel for the HiFive1 platform
+
+TARGET=riscv32i-unknown-none-elf
```
alistair23, line 3: There shouldn't be an `i` in the `riscv32` name.

boards/litex_arty/Makefile
```diff
@@ -0,0 +1,7 @@
+# Makefile for building the tock kernel for the HiFive1 platform
```
alistair23, line 1: Comment is out of date.

boards/litex_arty/Makefile
alistair23, line 3: There shouldn't be an i in the riscv32 name.

boards/litex_arty/README.md
```diff
@@ -0,0 +1,137 @@
...
+Since LiteX is a SoC builder, the individual generated bitstreams can
+differ significantly depending on the LiteX release and configuration
+options used. This board definition currently targets and has been
+tested with
+- [the LiteX SoC generator, revision
+  444a605dea](https://github.com/enjoy-digital/litex/tree/444a605deae6a561dbe2c49bf3062eae6f3cd887)
+- using the included [target
+  file](https://github.com/enjoy-digital/litex/blob/444a605deae6a561dbe2c49bf3062eae6f3cd887/litex/boards/targets/arty.py)
+- built around a VexRiscv-CPU
+- along with the folling configuration options:
```
alistair23, line 18: s/folling/following/g

boards/litex_arty/README.md
```diff
@@ -0,0 +1,137 @@
...
+  ```
+
+Many bitstream customizations can be represented in the Tock board by
+simply changing the variables in `src/litex_generated.rs`. To support
+a different set of FPGA cores and perform further modifications, the
+`src/main.rs` file will have to be modified.
+
+
+Warning
+-------
```
alistair23, line 36: I'm not sure if we really need a warning, pretty much every board supported by Tock is under development.

boards/litex_arty/README.md
```diff
@@ -0,0 +1,137 @@
...
+can run multiple userspace applications.
+
+The following on-board components and cores are supported:
+- [X] Timer (with uptime support)
+- [X] UART output via USB-FTDI
+- [X] Green LEDs
+- [X] 100MBit/s Ethernet MAC
+
+The following components and cores require porting:
+- [ ] Compressed instruction support in the VexRiscv CPU
```
alistair23, line 50: This is a hardware limitation right? Do we need to list that here?

boards/litex_sim/Makefile
```diff
@@ -0,0 +1,7 @@
+# Makefile for building the tock kernel for the HiFive1 platform
+
+TARGET=riscv32i-unknown-none-elf
+PLATFORM=litex_sim
```
alistair23, line 4: Same comments here

boards/litex_sim/README.md
```diff
@@ -0,0 +1,86 @@
+LiteX SoC on the Digilent Arty-A7 FPGA Board
```
alistair23, line 1: Wrong title here, all the comments in the previous readme apply here as well.

libraries/tock-cells/src/take_cell.rs
```diff
@@ -23,7 +23,7 @@ impl<'a, T: ?Sized> TakeCell<'a, T> {
     }
 
     /// Creates a new `TakeCell` containing `value`
-    pub fn new(value: &'a mut T) -> TakeCell<'a, T> {
+    pub const fn new(value: &'a mut T) -> TakeCell<'a, T> {
```
alistair23, line 26: This could probably be split out as it's own PR.

libraries/litex-register-gen/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "litex-register-gen"
```
alistair23, line 2: This is a long PR, maybe this could be split out separately?

chips/litex_vexriscv/src/chip.rs
```diff
@@ -0,0 +1,223 @@
...
+use kernel::common::registers::FieldValue;
+use kernel::debug;
+use kernel::hil::time::Alarm;
+use kernel::InterruptService;
+use rv32i::csr::{mcause, mie::mie, mip::mip, CSR};
+use rv32i::syscall::SysCall;
+use rv32i::PMPConfigMacro;
+
+// TODO: Rename this module? Is this really a PLIC?
+use crate::plic;
```
alistair23, line 14: It doesn't look like it.

chips/litex_vexriscv/src/chip.rs
```diff
@@ -0,0 +1,223 @@
...
+            soc_identifier,
+            userspace_kernel_boundary: SysCall::new(),
+            _pmp: PMP::new(),
+            scheduler_timer: kernel::VirtualSchedulerTimer::new(alarm),
+            interrupt_service,
+        }
+    }
+
+    pub unsafe fn enable_plic_interrupts(&self) {
+        plic::enable_interrupts();
```
alistair23, line 43: The interrupt controller should be included in the `Chip` (see the RISC-V boards today). Also it doesn't sound like a PLIC.

chips/litex_vexriscv/src/chip.rs
```diff
@@ -0,0 +1,223 @@
...
+                reenable_intr += mie::mext::SET;
+            }
+
+            if !mip.matches_any(mip::mext::SET + mip::mtimer::SET) {
+                break;
+            }
+        }
+
+        // re-enable any interrupt classes which we handled
+        CSR.mie.modify(reenable_intr);
```
alistair23, line 108: This looks out of date, can you update this?

chips/litex_vexriscv/src/chip.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+/// Trap handler for board/chip specific code.
+///
+/// For the Ibex this gets called when an interrupt occurs while the chip is
+/// in kernel mode. All we need to do is check which interrupt occurred and
+/// disable it.
+///
+/// TODO: Only for ibex?
```
alistair23, line 195: Nope :) This is a RISC-V thing. These are a little out of date as the interrupt handling has changed. Check the current HiFive1 code.

chips/litex_vexriscv/src/plic.rs
```diff
@@ -0,0 +1,93 @@
+// TODO: This module may require renaming, as in VexRiscv the
+// interrupt controller probably isn't really called "PLIC"
```
alistair23, line 2: Yeah it should be renamed. PLIC is actually a RISC-V spec so you either support it or you don't (lowRISC has a non-compliant PLIC unfortunately).

chips/litex_vexriscv/src/plic.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/// Disable all interrupts
+pub unsafe fn disable_interrupts() {
+    vexriscv_irq_raw::irq_setie(false);
+}
+
+/// Enable all interrupts
+pub unsafe fn enable_interrupts() {
+    vexriscv_irq_raw::irq_setie(true);
```
alistair23, line 20: This doesn't seem right. You shouldn't be changing the global interrupt enable (MIE) in the interrupt controller. This should instead enable or disable interrupts in the interrupt controller.

If they are always enabled then this should be a nop.

chips/litex_vexriscv/src/plic.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    #[cfg(not(any(target_arch = "riscv32", target_os = "none")))]
+    pub unsafe fn irq_getmask() -> usize {
+        0
+    }
+
+    #[cfg(all(target_arch = "riscv32", target_os = "none"))]
+    pub unsafe fn irq_getmask() -> usize {
+        let mask: usize;
+        asm!("csrr {mask}, {csr}", mask = out(reg) mask, csr = const CSR_IRQ_MASK);
```
alistair23, line 70: I think keeping these vendor specific CSRs in chips is the right call here.

boards/litex_sim/src/main.rs
```diff
@@ -0,0 +1,415 @@
...
+    uart0: &'static litex_vexriscv::uart::LiteXUart<'static, socc::SoCRegisterFmt>,
+    timer0: &'static litex_vexriscv::timer::LiteXTimer<
+        'static,
+        socc::SoCRegisterFmt,
+        socc::ClockFrequency,
+    >,
+    ethmac0: &'static litex_vexriscv::liteeth::LiteEth<'static, socc::SoCRegisterFmt>,
+}
+
+impl InterruptService<()> for LiteXSimInterruptablePeripherals {
```
alistair23, line 44: I'm assuming this isn't in chips because it's configurable? Maybe just add a comment saying that.

alistair23 (changes requested):

chips/litex_vexriscv/src/plic.rs
lschuermann: Right, I see. The global external interrupt source should not be controlled in the interrupt controller itself, but rather in the chip crate. This then reduces the features of the interrupt controller to masking interrupts and checking which are pending, which is fine. Individual peripherals have their own event manager, which they have full control over.

lschuermann: @alistair23 I've reworked the interrupt handling mechanisms, they should now closely match the `e310x` implementation. LiteX-VexRiscv does not have an `mtimer` by default, so I've thrown that out, but we can always introduce it again when necessary. If these changes look good to you, feel free to resolve this.

boards/litex_arty/README.md
lschuermann: I agree, most boards are in development. I may remove the implementation state under the warning or move it somewhere else. I do however think that lacking memory protection is an issue which should be prominently stated somewhere. I'm hoping that the PMP is integrated into VexRiscv sooner than later, however it could take a few months.

alistair23: It's worth saying that PMP isn't there, just maybe "Warning" is too strong, maybe "Notice" or something instead.

It's not really a issue though, so I'll resolve this.

chips/litex_vexriscv/src/interrupt_controller.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// The ordering is determined by the interrupt number, lower
+    /// having a higher priority.
+    pub fn next_saved(&self) -> Option<usize> {
+        let saved_interrupts = self.saved_interrupts.get();
+        let interrupt_bits = size_of::<usize>() * 8;
+
+        // This is essentially an inefficient version of C's find first
+        // set (ffs()) function, giving the index of the least significant
+        // bit that is set
+        (0..interrupt_bits).find(|test| saved_interrupts & (1 << test) != 0)
```
alistair23, line 63: Wouldn't `leading_zeros()` work here: https://doc.rust-lang.org/std/primitive.usize.html#method.leading_zeros

chips/litex_vexriscv/src/interrupt_controller.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// function returns `None`.
+    pub unsafe fn next_pending() -> Option<usize> {
+        let pending_interrupts = vexriscv_irq_raw::irq_pending();
+        let interrupt_bits = size_of::<usize>() * 8;
+
+        // This is essentially an inefficient version of C's find first
+        // set (ffs()) function, giving the index of the least significant
+        // bit that is set
+        (0..interrupt_bits).find(|test| pending_interrupts & (1 << test) != 0)
+    }
```
alistair23, line 89: If you are saving interrupts in the interrupt handler then this won't do what users expect, it's probably worth removing it.

chips/litex/src/uart.rs
```diff
@@ -0,0 +1,538 @@
...
+
+    pub fn initialize(&self, deferred_call_handle: DeferredCallHandle) {
+        self.uart_regs.ev().disable_all();
+        self.deferred_handle.set(deferred_call_handle);
+    }
+
+    pub fn transmit_sync(&self, bytes: &[u8]) {
+        // We need to make sure that we're not modifying interrupt
+        // pending and enabled bits here!
+        let regs = self.uart_regs;
```
alistair23, line 135: Don't save the registers in a temp value.

alistair23 (commented):
I'm a little confused by the need for the `LiteXSoCRegisterConfiguration`.

Is the problem that the Tock register interface assumes a usize length for register accesses, while LiteX allows any length (8/6/32/64 bits) registers?

boards/litex_arty/README.md
lschuermann: Right, I'll put it on my local todo list instead. VexRiscv _does_ support compressed instruction support (which would allow us to use the `riscv32imc` target, hence have hardware multiplication support), though it's not enabled in the `Full`-build which LiteX ships. I'll create a PR with them.

boards/litex_arty/README.md
lschuermann: Alright, I see. I've changed it to say "Please note", which is not as drastic as "Warning".

libraries/tock-cells/src/take_cell.rs
lschuermann: Right. I've opened #2214 accordingly, whose merge commit I'll cherry-pick onto this branch such that this isn't part of the merge diff anymore.

libraries/litex-register-gen/Cargo.toml
lschuermann: I agree that this should probably be reviewed separately. However swapping that out against the generated code won't exactly make the PR shorter, rather a bit longer. It'll be easier to review though.

chips/litex_vexriscv/src/chip.rs
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        // Re-enable all MIE interrupts that we care about. Since we
+        // looped until we handled them all, we can re-enable all of
+        // them.
+        CSR.mie.modify(mie::mext::SET);
+    }
+
+    fn has_pending_interrupts(&self) -> bool {
+        VexRiscvInterruptController::global_instance()
```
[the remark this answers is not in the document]
lschuermann: Right, the accessor method is probably not required and it's better to add a local reference to the `Chip` struct.

chips/litex/src/uart.rs
lschuermann: Why shouldn't I do this? This reduces verbosity a bit and it only saves the reference to the register struct in a variable (so no real copying of registers is happening here).

chips/litex_vexriscv/src/interrupt_controller.rs
lschuermann: Close! `trailing_zeros` appears to do the job. Unfortunately (and understandably) `usize::trailing_zeros(0) == usize::BITS == core::mem::size_of<usize>() * 8` and because `::BITS` is unstable, we're going to have to use the latter method to check whether there are any interrupts pending. Still worth the potential performance improvements by using LLVM intrinsics.

boards/litex_sim/src/main.rs
```diff
@@ -0,0 +1,420 @@
...
+    );
+
+    PANIC_REFERENCES.uart = Some(uart0);
+
+    // Create a shared UART channel for the console and for kernel debug.
+    //
+    // The baudrate is ingnored, as no UART phy is present in the
+    // verilated simulation.
+    let uart_mux =
+        components::console::UartMuxComponent::new(uart0, 115200, dynamic_deferred_caller)
```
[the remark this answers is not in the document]
lschuermann: Here the scenario is somewhat different. A few lines above I'm initializing the UART without PHY registers (passing a `None`). This means that all calls to `configure()` end up with an `ReturnCode::ENOSUPPORT` -- which makes sense, since this is a simulation after all.

This baud rate is therefore just a dummy value. We could replace it with `1` or some other value? I would've hoped the comment above explains this sufficiently. What do you think?

boards/litex_arty/src/main.rs
```diff
@@ -0,0 +1,469 @@
...
+        dynamic_deferred_caller
+            .register(uart0)
+            .expect("dynamic deferred caller out of slots"),
+    );
+
+    PANIC_REFERENCES.uart = Some(uart0);
+
+    // Create a shared UART channel for the console and for kernel debug.
+    let uart_mux =
+        components::console::UartMuxComponent::new(uart0, 1_000_000, dynamic_deferred_caller)
```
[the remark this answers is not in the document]
lschuermann: Right, I'll move it to the other constants in `litex_generated_constants.rs`.

chips/litex/src/uart.rs
hudson-ayers: I think this is fine. The pattern we got rid of was `let regs = &*self.uart_regs`, which used to be necessary for some reason but no longer is.

lschuermann:
> I'm a little confused by the need for the `LiteXSoCRegisterConfiguration`.

Right, it's a little confusing. LiteX generates SoC with peripherals on a bus (for instance Wishbone), which then expose so called _configuration status registers_ (CSRs). Because routing wide buses on an FPGA is difficult and may consume more resources, LiteX allows you to specify the bus data width, to for instance `8` or `32` bits. Also endianness of the CSRs or the base integer width of the CPU may vary. This leads to predictable changes in the CSR layouts, which the `LiteXSoCRegisterConfiguration` abstracts over with different register types.

For instance, with `--csr-data-width=32`, the ethmac registers look like this:
| Addr | Register |
|---|---|
| 0x00 | `rx_slot` (2 bits wide) |
| 0x04 | `rx_length` (32 bits wide) |

whereas with `--csr-data-width=8` it would look like this:
| Addr | Register |
|---|---|
| 0x00 | `rx_slot` (2 bits wide) |
| 0x04 | `rx_length` MSB byte 0 |
| 0x08 | `rx_length` MSB byte 1 |
| 0x0C | `rx_length` MSB byte 2 |
| 0x10 | `rx_length` MSB byte 3 |

We _could_ just opt for supporting one of these variants. Limiting ourselves to 8 bits data width is very inefficient on large platforms, whereas limiting to 32 bits could exceed the HW limits of potentially interesting small targets such as [FOMU](https://tomu.im/fomu.html). LiteX has just recently switched to 32 as the default, as the impact is not too bad on most FPGAs, whereas on the iCE40 we're _really_ close to the FPGA limits, so 8 might be beneficial.

My suggestion: we keep this complexity for now (since it's written anyways, and IMHO is an interesting example of how hardware-software codesign could look in Tock). I'm trying to get Tock to run on FOMU. LiteX is also an interesting experimentation candidate to try 64bit RISC-V CPUs, for which we'd need this abstraction too. If 8 bit wide CSRs loose significance within the interesting targets in Tock+LiteX and/or we decide that we don't want 64-bit RISC-V support, we can always replace it by standard `tock-registers` types later.

I'm going to throw out the macro for generating these types from this PR and just insert the generated code. The macro could then be reviewed in a follow-up PR.

libraries/litex-register-gen/Cargo.toml
lschuermann: Done!

boards/litex_sim/src/main.rs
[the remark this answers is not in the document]
alistair23: Ah, I see. This is probably fine then.

boards/litex_sim/src/main.rs
```diff
@@ -0,0 +1,420 @@
...
+            >,
+            LiteXSimInterruptablePeripherals,
+        >,
+    >,
+    uart: Option<&'static litex_vexriscv::uart::LiteXUart<'static, socc::SoCRegisterFmt>>,
+}
+static mut PANIC_REFERENCES: LiteXSimPanicReferences = LiteXSimPanicReferences {
+    chip: None,
+    uart: None,
+};
```
hudson-ayers, line 101: This seems like a slight improvement on my approach of re-initializing these items in io.rs. Unifies what we do for CHIP with everything that has to be used in panic context, and makes it clear what is happening. I may try to port this to the other boards at some point

lschuermann: I'm not entirely sure whether it's a large improvement, but at one stage while porting I had 5 different `static`s which would need to be passed to the panic handler, where the benefits of this approach clearly showed. :smile:

bradjc (dismissed):
Looks good.

I'm a little hesitant about having two boards and the potential difficulty of maintaining them, but I suspect that either it won't be a problem or we can re-evaluate in the future.

Also, the register thing is pretty weird, but the new register interface is self contained, so that seems like a good solution.

chips/litex/src/led_controller.rs
```diff
@@ -0,0 +1,152 @@
...
+    fn destroy_led(&self, index: usize) {
+        self.led_references
+            .set(self.led_references.get() & !(1 << index));
+    }
+
+    fn set_led(&self, index: usize, val: bool) {
+        if val ^ self.active_low {
+            self.regs
+                .leds_out
+                .set(self.regs.leds_out.get() | (1 << index));
```
alistair23, line 95: Why not just use `modify()`?

chips/litex_vexriscv/src/chip.rs
```diff
@@ -0,0 +1,215 @@
...
+
+            // Save the interrupts and check whether at least one
+            // interrupt is to be handled
+            //
+            // If no interrupt was saved, reenable interrupts
+            // immediately
+            if !INTERRUPT_CONTROLLER.save_pending() {
+                CSR.mie.modify(mie::mext::SET);
+            }
+        }
```
alistair23, line 178: This should be done in a loop.

alistair23 (commented):
Overall looks good.

I think it would be great if the duplicate board code could be reduced between the two boards. The new component method of creating boards should hopefully be able to help with that.

For the Litex Registers I would like to see that functionality added to Tock registers instead.  Although it seems to be a specific LiteX use case it would still remove a lot of duplicate code by allowing non target length accesses in Tock registers. Maybe there are other platforms where this would be useful as well. In saying that if it ends up a pain to implement maybe this is better. I just see it diverging from Tock registers and as bugs are fixes there, they won't be fixed here.

Also the current timer implementation seems a little sketchy and doesn't handle overflows. That should be improved.

I don't think any of those 3 issues block merging this though, they can be future improvements.

chips/litex/src/led_controller.rs
lschuermann: `modify()` takes a `FieldValue<T, R>` and this logic should be entirely generic over the register length (as adding more LEDs will cause the register width to simply grow). While I could technically construct a FieldValue with a matching mask, shift and value here, this seems like the more natural and readable option in this case, since I'm effectively only setting single bits in a flat integer.

chips/litex_vexriscv/src/chip.rs
lschuermann: I'm not quite sure I follow. On `e310x` for example, we run this in a loop as `save_interrupt(&self, index: u32)` on the PLIC saves only a single interrupt, whereas this `save_pending()` saves the entire interrupt register at once.

The only difference is that on the `e310x`, as soon as all interrupts are saved internally, the `mext` is enabled again, whereas on `litex_vexriscv` they're left disabled until all of them are handled (in line 96). What's the approach to use here?

chips/litex/README.md
```diff
@@ -0,0 +1,14 @@
+LiteX SoC Peripherals
+=====================
+
+LiteX is a Migen based Core / SoC builder. It allows developers to
+combine peripherals and CPUs into a custom SoC easily. One of the
+supported CPUs is the VexRiscv processor described in SpinalHDL.
```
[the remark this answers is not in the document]
lschuermann: Absolutely. I've added links to LiteX, LiteEth and internal links to the respective core implementations.

lschuermann:
Thanks for the in-depth reviews @alistair23 @bradjc @hudson-ayers!

> I'm a little hesitant about having two boards and the potential difficulty of maintaining them, but I suspect that either it won't be a problem or we can re-evaluate in the future.

@bradjc I agree, it's not the best solution, especially given that they are very similar. However, I expect the boards to further deviate in the future and to be honest, there's not too much complicated going on in there. If it ever becomes an issue I'll be sure to think of a solution. Could it already help to put these into a `litex/` subdirectory, similar to the Nordic boards?

> Also, the register thing is pretty weird, but the new register interface is self contained, so that seems like a good solution.

Right, that was the most important design consideration. I don't want to leak any of this weirdness into other parts of the code. As I outlined in my explanation above, for now the abstraction is good to have, but we might throw it out in the future if 8-bit wide CSRs loose relevance for Tock (i.e. if my port to FOMU won't work because of resource constraints).

---

> The new component method of creating boards should hopefully be able to help with that.

@alistair23 Isn't this already using the new component method? It just doesn't define a default set of peripherals for a chip because it doesn't make sense for a dynamically generated SoC.

> For the Litex Registers I would like to see that functionality added to Tock registers instead.

I don't see any point in doing so. I'm always trying to find ways to improve tock-registers, but quite honestly this is a crazy edge-case, which might not even remain in Tock for forever. If no other chip requires similar measures, I don't see any justification for adding this complexity to tock-registers.

> Also the current timer implementation seems a little sketchy and doesn't handle overflows. That should be improved.

Right! I'll be sure to tick the boxes in the PR's TODO-section in the next few days, fixing the Timer, from which point on this should be ready (from my side).

bradjc:
> @bradjc I agree, it's not the best solution, especially given that they are very similar. However, I expect the boards to further deviate in the future and to be honest, there's not too much complicated going on in there. If it ever becomes an issue I'll be sure to think of a solution. Could it already help to put these into a `litex/` subdirectory, similar to the Nordic boards?

Up to you.

alistair23:
Whoops, yes you already are.

True, but the large amount of duplicate code is a pain. It's fine for now though, maybe this is the best bet anyway.

Great!

lindemer:
Just FYI, a new core variant with 16 PMP registers and M/U modes has been upstreamed in both VexRiscv and LiteX. You can find it under the names `VexRiscv_Secure.v` and `VexRiscv_SecureDebug.v` in the LiteX cores repository: https://github.com/litex-hub/pythondata-cpu-vexriscv/pull/9

lschuermann:
@lindemer That's pretty awesome! It makes this a much more viable target for Tock. I'll fix the LiteX timers today though, and since this is a long and large PR already I'd prefer not to switch the CPU or LiteX release in this, but rather in a follow up PR. :+1:

lschuermann:
I rewrote large parts of the timer and alarm infrastructure which should now be more reliable and handle a wrapping uptime register reliably (although it's very unlikely given it's a 64-bit timer at ~100MHz-1GHz). In general, the documentation is improved and many rough edges have been removed. With the scheduler timer bug identified to be independent of this PR and a solution proposed in #2267, I think this could be ready to go! I'll take care of the (now integrated) PMP and Ethernet improvements (along with a trait and TAP-networking driver) in followup PRs.

Thank you for your reviews and patience, @alistair23 @bradjc @hudson-ayers!

Cargo.toml
```diff
@@ -1,3 +1,4 @@
+
```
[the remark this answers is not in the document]
lschuermann: Right, must've missed that from my diff, sorry.

boards/litex/arty/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+
+Programming
+-----------
+
+By default, the LiteX SoC will feature an integrated BIOS in ROM,
+which acts as a bootloader. The Tock kernel binary can be loaded
+either using `litex_term.py` (available as `lxterm`) via serial, or
+using TFTP via Ethernet. The uploaded image will be placed into the
+`main_ram` section and executed.
```
bradjc, line 103: Can you add instructions on how to do this, or add it to the Makefile?

boards/litex/sim/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+This command will build the SoC Verilog files and a ROM-bios.
+Afterwards, it will use Verilator to create a C++ simulation program,
+which is compiled and run. The script may ask for an administrator
+password to create a `tap0` network device on the host if built with
+Ethernet support.
+
+
+Debugging
+---------
```
bradjc, line 72: How do you run the kernel?

bradjc (dismissed):
I didn't look through all the files in detail, but looks like a good new platform addition!

boards/litex/arty/README.md
lschuermann: Sure. If you're referring to the TFTP-based upload, that's described below. I can add a code snippet for the invocation of `litex_term.py` though.

boards/litex/sim/README.md
lschuermann: I can add a section, but it's just `./litex_sim.py` with the configuration options stated above, so:

```
./litex_sim.py \
  --csr-data-width=32 \
  --integrated-rom-size=1048576 \
  --cpu-variant=full \
  --with-ethernet \
  --rom-init $PATH_TO_TOCK_BINARY (e.g. target/riscv32i-unknown-none-elf/release/litex_sim.bin)
```

alistair23 (approved):

lschuermann:
@bradjc Addressed your documentation comments, which unfortunately dismissed the reviews @alistair23 @bradjc.

alistair23 (approved):

bradjc (approved):

PR #2205, opened 2020-11-19
OpenTitan-Earlgrey: Integrate TickFS into Tock

alistair23:
### Pull Request Overview

This is the first stab at integrating [TickFS](https://github.com/tock/tock/pull/2156), which needs review, into Tock.

This adds a internal state machine to TickFS and some continue functions to allow it to work async while also being friendly to sync users.

This is a WIP PR with debug statements, lots that isn't yet implementation and using the deprecated `core::hash::SipHasher`. It's hopefully just a way to get reviews for https://github.com/tock/tock/pull/2156

### Testing Strategy

This pull request was tested by running on OpenTitain.

I can store a key in flash and then reboot Tock and see that it's still in flash and read it back.

### TODO or Help Wanted

Review: https://github.com/tock/tock/pull/2156

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/kv_store.rs
+++ b/capsules/src/kv_store.rs
@@ -10,17 +10,9 @@ use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::hil::flash::{self, Client, Flash};
 use kernel::{AppId, AppSlice, Callback, Driver, Grant, ReturnCode, Shared};
 use tickfs;
-use tickfs::TickFS;
+use tickfs::AsyncTickFS;
 
-#[derive(Clone, Copy)]
-enum State {
-    None,
-    ReadComplete(isize),
-    WriteComplete,
-    EraseComplete(usize),
-}
-
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, PartialEq)]
 enum Operation {
     None,
     Init,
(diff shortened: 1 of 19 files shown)


alistair23:
@bradjc I just pushed an update that now has app callbacks.

I tested from user space and I can retrieve a key and I get a callback when the operation is done.

It still uses the deprecated `core::hash::SipHasher` and is missing functionality and error checking, so not ready to merge, but should be enough justification to merge #2156 and then continue improving both.

libraries/tickfs/src/flash_controller.rs
```diff
@@ -0,0 +1,100 @@
...
+///     fn write_region(&self, address: usize, buf: &[u8]) -> Result<(), ErrorCode> {
+///         unimplemented!()
+///     }
+///
+///     fn erase_region(&self, region_number: usize) -> Result<(), ErrorCode> {
+///         unimplemented!()
+///     }
+/// }
+/// ```
+pub trait FlashController {
```
hudson-ayers, line 49: Why do we need a new flash interface, rather than reusing the interface in `kernel/src/hil/flash.rs`? I understand that removing static lifetimes was one motivation, but this makes additional modifications (operating on regions instead of pages, offering offset support) -- should we just be making those changes to the HIL?

capsules/src/kv_store.rs
```diff
@@ -0,0 +1,434 @@
...
+
+    fn erase_region(&self, region_number: usize) -> Result<(), tickfs::error_codes::ErrorCode> {
+        self.flash.erase_page(self.region_offset + region_number);
+
+        Err(tickfs::error_codes::ErrorCode::EraseNotReady(region_number))
+    }
+}
+
+pub struct KVStoreDriver<'a, F: Flash + 'static> {
+    tickfs: TickFS<'a, TickFSFlastCtrl<'a, F>, SipHasher>,
```
hudson-ayers, line 131: this setup, where the userspace driver owns the filesystem, suggests that the file system will only be for use by applications. I think this is fine as an initial approach, but it would ultimately be nice for capsules to be able to write to the file system. Do you think anything about the design of TickFS precludes virtualization within the kernel?

libraries/tickfs/src/flash_controller.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// On success it should return nothing, on failure it
+    /// should return ErrorCode::WriteFail.
+    ///
+    /// If the write operation is to be complete asynchronously then
+    /// `write_region()` can return `ErrorCode::WriteNotReady(region_number)`.
+    /// By returning `ErrorCode::WriteNotReady(region_number)`
+    /// `read_region()` can indicate that the operation should be retried in
+    /// the future. Note that that region will not be written
+    /// again so the write must occur otherwise the operation fails.
+    fn write_region(&self, address: usize, buf: &[u8]) -> Result<(), ErrorCode>;
```
hudson-ayers, line 89: This interface seems pretty hard to get right. It sounds like any failed write should be retried, but probably not retried in a loop (because the kernel is single threaded, so any bottom half interrupts needed to make TickFS ready to write will not run until the caller somehow yields). So the caller probably needs to schedule a deferred call anytime a write fails? But I don't see any use of `DeferredCall` in this PR. Retrying the call from within a deferred call will require that the buffer being written be stored across invocations of the kernel loop, which will require a buffer with a lifetime at least as long as that of the calling object (i.e. I see no way to reliably write from a stack allocated buffer with this interface, because any failure will require storing the buffer in something that outlives the stack frame of the caller).

libraries/tickfs/src/flash_controller.rs
alistair23: The idea is that TickFS can be split out as a separate crate (like the Tock registers). The hope is then that others can use it as there doesn't seem to be anything else in Rust for storing data on flash. That is why a new interface is added.

It is probably worth improving the current Tock Flash HIL, as allowing address writes is pretty useful. I think that can be done separately though.

capsules/src/kv_store.rs
alistair23: Yeah, my current focus has only been exposing it to userspace applications.

I can't think of a reason we can't virtualise it, but I haven't thought about it much to be honest.

hudson-ayers: Sounds good. Given that TickFS does not implement a HIL-like interface, it seems fine to just punt this problem to the future.

libraries/tickfs/src/flash_controller.rs
alistair23: Yep, the interface is a little more complex then I would like it to be. This is the tradeoff to allow non `'static` buffers for other users (if there are any).

So a write doesn't have to be re-tried. A write is the last operation and nothing happens after a write. So the idea is a write will occur and then we are done. If the write actually fails in the driver, then we should get an error callback which we have to handle (I don't handle errors here yet). Eventually the write will complete in the driver, so we don't need to reschedule it.

hudson-ayers: Oh, I think I understand, `WriteNotReady` means that the write is going to happen eventually, it just didn't happen immediately. That is more usable. I am not sure that should be reported as an error -- perhaps this could return something like `Result<SuccessCode, ErrorCode>`.
with
```rust
pub enum SuccessCode {
    Written
    Queued
}
```?

It seems weird that if TickFS is implemented on off-chip flash every single write would return an error, despite eventually succeeding.

alistair23: Yep, exactly!

I agree, it isn't really an error, but I also didn't want to report a success.

I was trying to avoid having a `SuccessCode` as I think that's a little convoluted. I'll add it in though.

alistair23: Done.

Issue #2201, opened 2020-11-16
Make netlify a required check 


hudson-ayers:
The netlify CI is currently having issues, so I made it a non-required check in the Github UI. We should require it again once netlify is fixed. This issue serves as a reminder.

alistair23:
Any idea how to fix it? It seems like I can't even see the error without a login.

hudson-ayers:
I don't have access either, and haven't really looked into it.

ppannuto:
There are two issues, one is hopefully transient and had to do with the netlify cache doing bad things when a submodule updates (that's all the PR's hanging at `Preparing Git Reference refs/heads/master`) -- a clear cache and rebuild seems to fix those that still have a bad cache. I don't have the support thread open on this computer, but I think Netlify is working on or had a fix.

The other is that something changed in the tty/term config causing hangs, which I started diagnosing here, but ran out of time for. Hopefully will be able to get to it soonish.

hudson-ayers:
Netlify has been back working correctly for awhile, so I made it a required check again, which closes this.

PR #2210, opened 2020-11-22
Add missing IntLike implementation

berkus:
For u128 type.

- [x] Ran `make prepush`.

I'm using this type to define a very wide bitfield type and need the impl.

### DIFF
--- a/libraries/tock-register-interface/src/registers.rs
+++ b/libraries/tock-register-interface/src/registers.rs
@@ -99,6 +99,11 @@ impl IntLike for u64 {
         0
     }
 }
+impl IntLike for u128 {
+    fn zero() -> Self {
+        0
+    }
+}
 
 /// Descriptive name for each register.
 pub trait RegisterLongName {}


lschuermann (approved):
Looks good. I don't see anything speaking against this addition, especially as with `u64` we already support an integer type larger than the word size of the currently supported architectures.

lschuermann:
The CI failure you're experiencing should be fixed in [this PR](https://github.com/tock/tock/pull/2212).

berkus:
@lschuermann shall we merge that PR first? I see there's still one check failing there.

I can wait until that's fixed, currently using my own fork of tock-registers with more optimizations and fixes - hopefully will upstream it at some point.

lschuermann:
It doesn't really matter. People with permissions can merge your PR despite the CI failing. The Netlify docs preview failing is a known issue, see #2201.

ppannuto (approved):
bors r+

I've merged the other PR, this should either go in with that batch or follow it, and bors / CI should clear.

berkus:
🙇 thanks folks!

berkus:
@ppannuto @lschuermann question: I'm sitting on some changes, namely reducing the verbiage for repeated u8, u16, u32 etc impls: see [here](https://github.com/metta-systems/vesper/commit/0f50bd1e19505832dc53ebcda4e5ede3d880b72b) for example.

Is there a chance these could be merged? If so I can prepare a proper PR with changes (full list [here](https://github.com/metta-systems/vesper/compare/fb607cebd41898d867fdce5b8e2f54cd6387043a..9a0a3396ade5c6b2e66228c26c2fce97bdc91f68)).

lschuermann:
@berkus Those are some good cleanups. Adding macros is always a tradeoff, however here it should be fine.

You're currently invoking, per type, several macros which take variadic arguments. It's better for one macro to generate all code for a specific type, and then invoke that multiple times for the respective types. This reduces complexity and interdependence of the macros.

So going from this:
```rust
IntImpl_for!(u8, u16);
FromImpl_for!(u8, u16);
```

to something like:
```
register_type(u8);
register_type(u16);
```

Or is there any reason to call one macro on a type, but not the others? Then this of course doesn't apply.

berkus:
@lschuermann It's a tradeoff, but I wanted to keep it small and clean, so it's actually useful. I even used a decl-macro first, but it seems that it doesn't work (has too many limitations atm, and couldn't generate code properly), so I had to switch to macro_rules with slightly uglier syntax.

The original `macro` implementation did use the approach you listed, but that's because decl-macro does not have this pattern matching abilities.

Since I switched to macro_rules I decided to not mash all the unrelated code into single macro and keep it clean and separated - IntLike impls in one macro, From impls in another, etc. The final task was reducing cognitive complexity of a) repeated code with slight variations, b) too much unrelated code smashed under a single macro. I believe this is a reasonably fine balance I found here. 

Also I don't see any interdependence between these macros, they are fairly independent and can be modified/extended separately if necessary.

lschuermann:
@berkus I get your point, the implementations of `From<`, `IntLike`, etc. are technically usable separately, but in the current case there doesn't seem to be a use case for implementing those separately, or is there?

If you insist on keeping those in separate macros, I'm still vouching against the punctuated matches (e.g. operate on multiple types in a single invocation). As a general rule of thumb (to extend the idea of macro hygiene), a macro invocation should generate code that is inherently coupled. Sometimes variadic parameters make sense or are even required. In this case, the implementations over types such as `u8`, `u16`, etc. really don't have anything to do which each other, so they should be generated in individual calls to the common macro.

berkus:
@lschuermann sure, i can split those out into separate calls, it wont add anything (except line count) but wont take anything away either. Shall I prepare the PR then?

lschuermann:
That'd be great. There will probably be some more feedback from the others and I naturally won't (and can't :smile:) _promise_ it'll get merged, but as I said, it seems to be a good cleanup!

brghena:
I can add that I would absolutely accept it. We do not like code duplication, as it leads to bugs when you forget to update some of the duplicates. Plus the macro for this solution is pretty darn simple, even moreso if you go with lschuermann's suggestion of a single type per macro call (which I agree with).

berkus:
Ok, I've submitted #2220

PR #2212, opened 2020-11-22
Unbreak CI by updating the package repositories

lschuermann:
### Pull Request Overview

This pull request
- updates the package repositories on Ubuntu-runners where packages are installed using apt
- calls to `apt` only when running on a `ubuntu-latest` machine

This unbreaks the `ci-tests` job on `ubuntu-latest`. It also makes the CI more resilient to outdated package repositories in the provided images from GitHub. Furthermore, it doesn't fail the _Install dependencies_ job silently to succeed on a macOS machine, which causes real errors on Ubuntu to be ignored.

### Testing Strategy

You're looking at it.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Update package repositories
+        run: |
+          sudo apt update
       - name: Install dependencies
         continue-on-error: true
         run: |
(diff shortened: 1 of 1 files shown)


ppannuto (approved):
bors r+

I'm surprised an mildly annoyed this is necessary given our use of `ubuntu-latest`, but it does fix the problem.

hudson-ayers (approved):

PR #2213, opened 2020-11-22
kernel,arch,components: accept trailing commas in macro_rules

lschuermann:
### Pull Request Overview

This pull request makes the various `macro_rules!` macros in the `kernel`, `arch` and `components` crates accept trailing commas, wherever the Rust function calling syntax would allow it too.

This is in line with the changes of #2197, inspired by a comment of @hudson-ayers. :)

While these changes may not make much sense in `macro_rules!` with only a single argument, I'd rather change these consistently and to the way how Rust syntax would behave. Furthermore, this might increase the chance that someone copying a component helper macro will get this right on the first try.

### Testing Strategy

### TODO or Help Wanted

I haven't touched `chips` and `capsules` yet, as I do not know whether it is fine to "police" / update these macros. Furthermore, I'm not sure whether we want to further deviate from upstream in the copied and modified `enum_primitive` library by possibly changing those macros too.

The kernel-provided (especially `static_init!`) and component-helper macros seemed the most important to change (as I've often ran into errors here :) ).

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/gpio.rs
+++ b/boards/components/src/gpio.rs
@@ -51,8 +51,8 @@ use kernel::static_init_half;
 #[macro_export]
 macro_rules! gpio_component_helper_max_pin {
     () => { 0usize };
-    ($a: expr, $b: expr, $($tail:expr,)*) => { $crate::gpio_component_helper_max_pin! (max ($a, $b), $($tail,)*) };
-    ($a: expr,) => { $a };
+    ($a:expr, $b:expr, $($tail:expr),* $(,)?) => { $crate::gpio_component_helper_max_pin! (max ($a, $b), $($tail,)*) };
+    ($a:expr $(,)?) => { $a };
 }
 
 #[macro_export]
@@ -64,7 +64,7 @@ macro_rules! gpio_component_helper_max_pin {
 macro_rules! gpio_component_helper {
     (
         $Pin:ty,
-        $($nr:literal => $pin:expr),*
+        $($nr:literal => $pin:expr),* $(,)?
     ) => {{
         use kernel::count_expressions;
         use kernel::hil::gpio::InterruptValueWrapper;
@@ -91,7 +91,7 @@ macro_rules! gpio_component_helper {
 
 #[macro_export]
 macro_rules! gpio_component_buf {
-    ($Pin:ty) => {{
+    ($Pin:ty $(,)?) => {{
         use capsules::gpio::GPIO;
         use core::mem::MaybeUninit;
         static mut BUF: MaybeUninit<GPIO<'static, $Pin>> = MaybeUninit::uninit();
--- a/kernel/src/common/utils.rs
+++ b/kernel/src/common/utils.rs
@@ -15,7 +15,7 @@
 /// destructor.
 #[macro_export]
 macro_rules! static_init {
-    ($T:ty, $e:expr) => {{
+    ($T:ty, $e:expr $(,)?) => {{
         let mut buf = $crate::static_buf!($T);
         buf.initialize($e)
     }};
@@ -45,7 +45,7 @@ macro_rules! static_init {
 /// possible.
 #[macro_export]
 macro_rules! static_buf {
-    ($T:ty) => {{
+    ($T:ty $(,)?) => {{
         // Statically allocate a read-write buffer for the value, write our
         // initial value into it (without dropping the initial zeros) and
         // return a reference to it.
@@ -137,7 +137,7 @@ impl<T> StaticUninitializedBuffer<T> {
 /// The static buffer must be passed in.
 #[macro_export]
 macro_rules! static_init_half {
-    ($B:expr, $T:ty, $e:expr) => {
+    ($B:expr, $T:ty, $e:expr $(,)?) => {
         {
             use core::mem::MaybeUninit;
             let buf: &'static mut MaybeUninit<$T> = $B;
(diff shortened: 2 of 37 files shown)


boards/components/src/gpio.rs
```diff
@@ -51,8 +51,8 @@ use kernel::static_init_half;
 #[macro_export]
 macro_rules! gpio_component_helper_max_pin {
     () => { 0usize };
-    ($a: expr, $b: expr, $($tail:expr,)*) => { $crate::gpio_component_helper_max_pin! (max ($a, $b), $($tail,)*) };
-    ($a: expr,) => { $a };
+    ($a:expr, $b:expr, $($tail:expr),* $(,)?) => { $crate::gpio_component_helper_max_pin! (max ($a, $b), $($tail,)*) };
+    ($a:expr $(,)?) => { $a };
```
lschuermann, line 55: Quite interestingly, here a trailing comma was actually _required_ in the arguments. :)

kernel/src/common/utils.rs
```diff
@@ -210,6 +210,6 @@ macro_rules! create_capability {
 #[macro_export]
 macro_rules! count_expressions {
     () => (0usize);
-    ($head:expr) => (1usize);
-    ($head:expr, $($tail:expr),*) => (1usize + count_expressions!($($tail),*));
+    ($head:expr $(,)?) => (1usize);
+    ($head:expr, $($tail:expr),* $(,)?) => (1usize + count_expressions!($($tail),*));
```
ppannuto, line 214: I don't speak Rust macro that well, but it looks like this might also be a case where the trailing comma is currently required (as part of the `$($tailExpr),*` pattern?) and the last `$(,)?` isn't actually ever going to match?

ppannuto (approved):
This is a great ergonomics improvement -- thanks for tackling all of these!

kernel/src/common/utils.rs
lschuermann: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4c2e7eb72007eda40ad5b974a54d0dc8

I have to admit, it's a quite confusing syntax. :laughing: However it's still required, since the repeated match (is there a more precise name for this?) only specifies the delimiter _between_ the items, so any trailing comma needs to be matched separately.

lschuermann:
@ppannuto What do you think about the places listed in _TODO or Help Wanted_? Is it fine for me to change `chips` and `capsules` too or are these the responsibilities of the respective authors / maintainers? I guess those macros aren't used as extensively around Tock, but it's a mechanical change I could do for consistency's sake.

hudson-ayers (approved):
Thanks for taking this on!

hudson-ayers:
> @ppannuto What do you think about the places listed in _TODO or Help Wanted_? Is it fine for me to change `chips` and `capsules` too or are these the responsibilities of the respective authors / maintainers? I guess those macros aren't used as extensively around Tock, but it's a mechanical change I could do for consistency's sake.

I think it would be fine to change the macros in `chips` and `capsules`, but it is not necessary if you don't feel like it. It is probably best to leave `enum_primitive` alone IMO.

lschuermann:
I'll add the commits for `chips` and `capsules` shortly.

kernel/src/common/utils.rs
```diff
@@ -15,7 +15,7 @@
 /// destructor.
 #[macro_export]
 macro_rules! static_init {
-    ($T:ty, $e:expr) => {{
+    ($T:ty, $e:expr $(,)?) => {{
```
bradjc, line 18: I understand the cases where there are lists, but why would you want to add a comma after the second argument to something like static_init?

bradjc (changes requested):
I understand the use cases for lists, but not for where macros take a fixed number of arguments. For example `align4!(0x6019,)` would look weird.

lschuermann:
@bradjc It looks weird when you write it out as one line, but it's valid Rust function invocation syntax.

_Especially_ with `static_init!`, where you have two (often) very long parameters, writing out as two lines is common. In this notation, a trailing comma is (subjectively speaking) more natural:

```rust
let hardware_timer = static_init!(
    rv32i::machine_timer::MachineTimer,
    rv32i::machine_timer::MachineTimer::new(e310x::timer::MTIME_BASE),
);
```

This is even more true for macros with a larger number of parameters, where you'd write each parameter on a separate line in any case.

hudson-ayers:
> @bradjc It looks weird when you write it out as one line, but it's valid Rust function invocation syntax.
> 
> _Especially_ with `static_init!`, where you have two (often) very long parameters, writing out as two lines is common. In this notation, a trailing comma is (subjectively speaking) more natural:
> 
> ```rust
> let hardware_timer = static_init!(
>     rv32i::machine_timer::MachineTimer,
>     rv32i::machine_timer::MachineTimer::new(e310x::timer::MTIME_BASE),
> );
> ```
> 
> This is even more true for macros with a larger number of parameters, where you'd write each parameter on a separate line in any case.

Playground example showing a trailing comma is valid in functions that take a single parameter: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8f64b3e50f69e8bf36b0d24e0c9efd71

lschuermann:
Whether one should do this is, of course, a subjective question. [This issue on rustfmt](https://github.com/rust-dev-tools/fmt-rfcs/issues/42) seems to have a lot of subjective comments and some evaluation of the advantages / disadvantages. What led me to implement it this way is that I wanted the macro invocations to closely follow the Rust function call syntax, wherever possible and reasonable.

lschuermann:
How do we continue here? I'd rather not make the additional changes to `chips` and `capsules` if this has no chance of being merged.

bradjc (approved):
That's a reasonable argument. And I support deferring to whatever rustfmt does.

PR #2214, opened 2020-11-22
tock-cells: make TakeCell constructor a const fn

lschuermann:
### Pull Request Overview

This pull request changes the TakeCell to be a `const fn`, such that it can be used in other `const fn`s (as used in PR #2203, port to LiteX SoCs).

This requires the `const_mut_refs` feature, which is already enabled in the `tock-cells` crate.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/libraries/tock-cells/src/take_cell.rs
+++ b/libraries/tock-cells/src/take_cell.rs
@@ -23,7 +23,7 @@ impl<'a, T: ?Sized> TakeCell<'a, T> {
     }
 
     /// Creates a new `TakeCell` containing `value`
-    pub fn new(value: &'a mut T) -> TakeCell<'a, T> {
+    pub const fn new(value: &'a mut T) -> TakeCell<'a, T> {
         TakeCell {
             val: Cell::new(Some(value)),
         }


hudson-ayers:
Can you give a pointer to where in #2203 requires this change? I would have thought that with the peripheral instantiation changes moving away from the use of global variables there should not be much need for const cell constructors (in fact, part of the original motivation for those changes was to allow us to reduce our reliance on various portions of the `const_fn` unstable feature).

lschuermann:
@hudson-ayers You're right! The `LiteEth` driver constructor used to be required `const` due to the way I've done peripheral instantiation prior to integrating your solution. As you've actually noted in https://github.com/rust-lang/rust/issues/57349 we're not going to get around using `const_mut_refs` in `tock-cells`, so what this adds will probably stabilize at the same time as the other uses. Would this still be worth adding then? Or should we refrain from adding this?

hudson-ayers:
I would prefer not to add it if there is not a use case that requires it. Somewhere in my work backlog is a plan to remove const constructors from `OptionalCell` and `MapCell` as part of reducing our reliance on the various sub-parts of `const_fn`.

PR #2215, opened 2020-11-23
remove all uses of unstable const_in_array_repeat_expressions feature

hudson-ayers:
### Pull Request Overview

This pull request removes all uses of the unstable `const_in_array_repeat_expressions` feature. It was originally necessary in order to have ergonomic scheduler components, but a subset of the feature that is sufficient for our use was accidentally stabilized, and the Rust maintainers are moving to make that stabilization permanent (see https://github.com/rust-lang/rust/pull/79270). So this is an easy way to get rid of an unstable feature.

### Testing Strategy

This pull request was tested by compiling and flashing several apps on an Imix.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/acd52832/src/main.rs
+++ b/boards/acd52832/src/main.rs
@@ -4,7 +4,6 @@
 // Disable this attribute when documenting, as a workaround for
 // https://github.com/rust-lang/rust/issues/62184.
 #![cfg_attr(not(doc), no_main)]
-#![feature(const_in_array_repeat_expressions)]
 #![deny(missing_docs)]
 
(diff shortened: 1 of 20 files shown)


lschuermann (approved):
Reading the Rust PR it seems like this is only a proposal to acknowledge that `[CONST; N]` is stable, not a commitment. Nonetheless this change can't hurt and it seems much closer to stable (well, actually is) than `const_in_array_repeat_expression`.

bradjc (approved):

alistair23 (approved):

PR #2216, opened 2020-11-23
add 15.4 and ble to nano33ble

hudson-ayers:
### Pull Request Overview

This pull request adds the 15.4 and BLE drivers to the nano33ble, rather than leaving support as commented out, as the comments had already fallen out-of-date.

### Testing Strategy

BLE was tested using the `ble_advertising` and `ble_passive_scanning` apps in `libtock-c`.

15.4 was tested using the `radio_tx` and `radio_rx` apps in `libtock-c` and sending messages back and forth with an nrf52840-dk.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nano33ble/src/main.rs
+++ b/boards/nano33ble/src/main.rs
@@ -322,7 +322,7 @@ pub unsafe fn reset_handler() {
             .finalize(());
 
     let serial_num = nrf52840::ficr::FICR_INSTANCE.address();
-    let serial_num_bottom_16 = serial_num[0] as u16 + ((serial_num[1] as u16) << 8);
+    let serial_num_bottom_16 = u16::from_le_bytes([serial_num[0], serial_num[1]]);
     let (ieee802154_radio, _mux_mac) = components::ieee802154::Ieee802154Component::new(
         board_kernel,
         &base_peripherals.ieee802154_radio,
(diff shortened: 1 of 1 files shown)


boards/nano33ble/src/main.rs
```diff
@@ -314,16 +317,23 @@ pub unsafe fn reset_handler() {
...
 
-    // let (ieee802154_radio, _) = Ieee802154Component::new(
-    //     board_kernel,
-    //     &base_peripherals.ieee802154_radio,
-    //     PAN_ID,
-    //     SRC_MAC,
-    // )
-    // .finalize(());
+    let serial_num = nrf52840::ficr::FICR_INSTANCE.address();
+    let serial_num_bottom_16 = serial_num[0] as u16 + ((serial_num[1] as u16) << 8);
```
lschuermann, line 325:
```suggestion
    let serial_num_bottom_16 = u16::from_le_bytes([serial_num[0], serial_num[1]]);
```

lschuermann (approved):
Looks great.

alistair23 (approved):

bradjc:
Should we only add 15.4? I would really like to see #2036 finished and used instead for BLE. Or is it worth it to have advertisements work in the short term, and change the syscall interface later?

hudson-ayers:
> Should we only add 15.4? I would really like to see #2036 finished and used instead for BLE. Or is it worth it to have advertisements work in the short term, and change the syscall interface later?

It seems that #2036 still has a ways to go. And once #2036 works we are going to have to change the syscall interface later for all the other nrf boards that currently provide only basic BLE advertisement support (nrf52840dk, nrf52840_dongle, nrf52dk), so my thought is that this just puts the nano33ble on par with what we will have to do for those other boards. I think it is nice to have advertisements work in the short term

bradjc (approved):

PR #2218, opened 2020-11-24
Allow processes to read from buffers shared with read-only allow

jrvanwhy:
### Pull Request Overview

This pull request changes the 2.0 syscalls TRD to allow processes that have shared a buffer with the kernel to read from that buffer.

The benefit of this change is it allows us to use `&[u8]` in `libtock-rs`' system call APIs to pass allow buffers. Without this change, `libtock-rs` would need to require processes to instantiate a custom type that tracks whether a buffer has been shared and mediates access to it, which is unergonomic to use.

### Testing Strategy

Ran `make prepush`, proofread changes.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/doc/reference/trd-syscalls.md
+++ b/doc/reference/trd-syscalls.md
@@ -459,9 +459,10 @@ again to re-allow it with a different size.
 ---------------------------------
 
 The Read-Only Allow class is identical to the Read-Write Allow class
-with one exception: the buffer it passes to the kernel is
-read-only. The kernel cannot write to it. It semantics and calling
-conventions are otherwise identical to Read-Write Allow.
+with two exceptions: the buffer it passes to the kernel is read-only, and the
+process retains read access to the buffer. The kernel cannot write to the
+buffer. Read-Only Allow's semantics and calling conventions are otherwise
+identical to Read-Write Allow.
 
 The Read-Only Allow class exists so that userspace can pass references to constant data
 to the kernel. Often, constant data is stored in flash rather than RAM. Constant data


lschuermann (approved):
I think this is definitely one of the desired effects of introducing the read only allow.

Also, I'm not really concerned about any safety and soundness impact, because in the kernel we must expect and tolerate or handle a much worse scenario anyways: that user-space actually modifies the buffer and hence does not obey to any of these rules.

Let's hear what @phil-levis says.

phil-levis (approved):

lschuermann:
@phil-levis I'll integrate the CI fix into the branch, which should cause CI to work again.

lschuermann:
I have cherry-picked the commits of #2212 onto the tock-2.0-dev branch, which should fix the CI. Because @phil-levis already approved and tried to merge, and this targets only the tock-2.0-dev branch, I'm taking the liberty to try merging.

bors r+

hudson-ayers:
hmm, I thought that bors would run the tests on the merge commit of master and the head of the PR, so I am surprised that a rebase/cherry-pick was necessary to pass the tests.

jrvanwhy:
> hmm, I thought that bors would run the tests on the merge commit of master and the head of the PR, so I am surprised that a rebase/cherry-pick was necessary to pass the tests.

This PR was merged into `tock-2.0-dev`, not `master`. Leon cherrypicked the commits into `tock-2.0-dev`, then the CI ran on the merge commit between this PR and `tock-2.0-dev`.

hudson-ayers:
ohhh, that makes sense. I didn't realize the target branch was different.

PR #2220, opened 2020-11-24
tock-registers: remove duplicate code, make local register copy read-write

berkus:
### Pull Request Overview

* Added macros to replace some repetitive code. Add missing implementations of said code for u128.

The task was reducing cognitive complexity of a) repeated code with slight variations, b) too much unrelated code smashed under a single macro. Based on feedback from @lschuermann [elsewhere](https://github.com/tock/tock/pull/2210#issuecomment-733025295) I believe this is balanced enough to keep compilation still fast and code DRY.

* Add more aggressive inlining markers.

* Make LocalRegisterCopy modifiable. I found in the usage (and eventual abusage) of LocalRegsiterCopy that I actually use it to mimic structured bitfield accesses in a very convenient form on local variables, and for this original LocalRegisterCopy was a bit limited. 

* Don't consume self in several FieldValue<> accessors. Turns out I still need FieldValue after accessing its mask(). This makes these methods more akin to impls for other structs, too.

### Testing Strategy

* [x] Added test for u128 field.

### TODO or Help Wanted

* [x] Should be good to go.

### Documentation Updated

I believe nothing is needed to be updated there.

### Formatting

- [x] Ran `make prepush`.

I didn't, but running `rustfmt` on the only touched file (registers.rs) yields no changes.

### DIFF
--- a/libraries/tock-register-interface/src/registers.rs
+++ b/libraries/tock-register-interface/src/registers.rs
@@ -694,6 +694,9 @@ mod tests {
             let field64 = Field::<u64, ()>::new(0x12345678_9abcdef0, 1);
             assert_eq!(field64.mask, 0x12345678_9abcdef0_u64);
             assert_eq!(field64.shift, 1);
+            let field128 = Field::<u128, ()>::new(0x12345678_9abcdef0_0fedcba9_87654321, 1);
+            assert_eq!(field128.mask, 0x12345678_9abcdef0_0fedcba9_87654321_u128);
+            assert_eq!(field128.shift, 1);
         }
 
         #[test]
--- a/libraries/tock-register-interface/src/registers.rs
+++ b/libraries/tock-register-interface/src/registers.rs
@@ -79,31 +79,21 @@ pub trait IntLike:
     fn zero() -> Self;
 }
 
-impl IntLike for u8 {
-    fn zero() -> Self {
-        0
-    }
+macro_rules! IntLike_impl_for {
+    ($type:ty) => {
+        impl IntLike for $type {
+            fn zero() -> Self {
+                0
+            }
+        }
+    };
 }
-impl IntLike for u16 {
-    fn zero() -> Self {
-        0
-    }
-}
-impl IntLike for u32 {
-    fn zero() -> Self {
-        0
-    }
-}
-impl IntLike for u64 {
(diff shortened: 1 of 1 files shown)


bradjc (approved):
I guess the macros are reasonable stand-ins until presumably rust updates some day and things can use generic types.

Also seems reasonable to me to make local register copy a mutable copy.

berkus:
@bradjc Yep, I'm keeping eye on rustc tracking issue for that, will rewrite this code as soon as rustc is ready.

lschuermann (approved):

berkus:
Is there something specific we're waiting for here?

lschuermann:
> Is there something specific we're waiting for here?

Not that I would know of. PRs are usually kept open for a couple of days for maintainers to take a look. If there are no other change requests incoming, I'd expect this to get merged in a couple of days. Because of the holidays in the US it might take a bit longer.

berkus:
@lschuermann got it, thanks!

brghena (approved):

hudson-ayers (approved):

alistair23 (approved):

berkus:
Thanks, folks!

berkus:
Hi, any plans to release this as `tock-registers` update?

lschuermann:
Out of experience I'd say normally it takes a bit longer until a new release is made which then incorporates more improvements (especially given that the previous release was tagged a mere 22 days ago).

I suppose if you need these changes in another crate, you could probably [reference the crate directly from the Tock repository](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories) (which will pull in a substantial amount of submodules) or a new minor release could be tagged. What do you think @ppannuto?

berkus:
I'd be very happy if you manage to pull tock-registers release soon, but if not - no worries, I can live with my current configuration for a while. Especially that I'm enjoying a new M1 rustc configuration fun at the moment (random SIGKILLS in rustc generated binaries :P)

lschuermann:
I've been told that it's unlikely we'll see a new `tock-registers` release this year (prior to the 18th), I'd guess that we could have a new release early next year. I hope that works for you.

berkus:
@lschuermann got it, sad but will survive!

PR #2221, opened 2020-11-24
usb-cdc: fake successful uart transmissions so that blocking debug calls resume

hudson-ayers:
### Pull Request Overview

Currently, `libtock-c` apps that contain `printf()` calls hang indefinitely after calling `printf()` on boards that use CDC-based console output, until a user calls `tockloader listen`. This can be observed by adding a print statement to the `blink` app.

As a fix for this issue, this PR modifies the CDC stack to deliver `SUCCESS` callbacks even when no CDC host is connected (basically pretending the data was sent). However, it only does so after a 10 second delay -- which ensures that queued debug statements made within the first 10 seconds after boot are not dropped in the time it might take a user to call `tockloader listen`. 

### Testing Strategy

This pull request was tested by adding a `printf()` call to `blink`, and running that app. With this change, the light begins blinking after 10 seconds if `tockloader listen` is not called. If `tockloader listen` is called after 5 seconds, all debug output is printed, including the initial "Initialization complete..." message.

### TODO or Help Wanted

This solution is a little hacky, but was my favorite of the options I thought of. Alternatives include:

- remove all blocking debug output from the kernel, `libtock-c` and `libtock-rs`. This would eventually lead to debug buffer overflow warnings, but perhaps those warnings could be suppressed somehow (or we could accept the wasted cycles?)

- Indicate failure rather than success, and modify the console driver to appropriately handle a specific error code as an indication that no callback will be delivered, and blocking should not occur. This is also probably more involved.

### Documentation Updated

- [x] No updates are required (I think? Is there somewhere I should document this?)

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/clue_nrf52840/src/main.rs
+++ b/boards/clue_nrf52840/src/main.rs
@@ -81,7 +81,11 @@ static mut PROCESSES: [Option<&'static dyn kernel::procs::ProcessType>; NUM_PROC
 
 static mut CHIP: Option<&'static nrf52840::chip::NRF52<Nrf52840DefaultPeripherals>> = None;
 static mut CDC_REF_FOR_PANIC: Option<
-    &'static capsules::usb::cdc::CdcAcm<'static, nrf52::usbd::Usbd>,
+    &'static capsules::usb::cdc::CdcAcm<
+        'static,
+        nrf52::usbd::Usbd,
+        capsules::virtual_alarm::VirtualMuxAlarm<'static, nrf52::rtc::Rtc>,
+    >,
 > = None;
 
 /// Dummy buffer that causes the linker to reserve enough space for the stack.
@@ -283,8 +287,13 @@ pub unsafe fn reset_handler() {
         0x239a,
         0x8071,
         strings,
+        mux_alarm,
+        dynamic_deferred_caller,
     )
-    .finalize(components::usb_cdc_acm_component_helper!(nrf52::usbd::Usbd));
+    .finalize(components::usb_cdc_acm_component_helper!(
+        nrf52::usbd::Usbd,
+        nrf52::rtc::Rtc
+    ));
     CDC_REF_FOR_PANIC = Some(cdc); //for use by panic handler
 
     // Create a shared UART channel for the console and for kernel debug.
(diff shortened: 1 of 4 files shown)


lschuermann:
Hm, that's an interesting problem. I agree, this is a good non-invasive solution, which fixed the currently experienced issues. However waiting for some interval and depending data was sent when it really wasn't are things I'd intuitively associate with race conditions.

Would it help for Tock 2.0 to be honest to the userspace application, possibly by introducing a return value along the lines of "it's queued, but depending on the rest of the system behavior the message might not actually be sent" (depending on the buffer space, etc.)?

bradjc (dismissed):
Very nice.

hudson-ayers:
> Would it help for Tock 2.0 to be honest to the userspace application, possibly by introducing a return value along the lines of "it's queued, but depending on the rest of the system behavior the message might not actually be sent" (depending on the buffer space, etc.)?

I think that is already effectively what `SUCCESS` means for any virtualized resource.  I think any changes to improve on this solution would be console specific, and not rely on any ABI changes.

alistair23 (dismissed):

hudson-ayers:
ah, wait, this is gonna need another rebase once the 15.4 PR passes bors

bradjc (approved):

lschuermann:
> I think that is already effectively what `SUCCESS` means for any virtualized resource. I think any changes to improve on this solution would be console specific, and not rely on any ABI changes.

Agreed, and I'm not talking about ABI changes. While Tock 2.0 certainly is a good chance to get the ABI right, the capsule APIs also have some stability (at least between minor releases), right? However if you think that having a call succeed on a virtualized resource will only indicate whether the request was queued and not necessarily whether it was executed (which is a very reasonable assumption), that's of course fine.

hudson-ayers:
> Agreed, and I'm not talking about ABI changes. While Tock 2.0 certainly is a good chance to get the ABI right, the capsule APIs also have some stability (at least between minor releases), right? However if you think that having a call succeed on a virtualized resource will only indicate whether the request was queued and not necessarily whether it was executed (which is a very reasonable assumption), that's of course fine.

For virtualized devices it is impossible to synchronously return errors beyond whether a request was queued. Errors can be returned asynchronously, and that is in fact possible with UART, but the concern here is that the console and userspace interactions with the console probably don't correctly handle those failures, so just returning SUCCESS to imitate how UART works on every other board is less likely to break things.

PR #2224, opened 2020-11-28
Port log tests from imix to nano33ble

kevtan:
### Pull Request Overview

I ported the persistent log tests (`linear_log_test.rs` and `log_test.rs`) from the imix to the nano33ble so that all the tests can be in place before we virtualize the log. This way, it'll hopefully be easier to test the correctness of the log virtualization.

### Testing Strategy

1. Compile and flash the kernel.
2. Retrieve debug output tockloader.
3. Manually verify that the debug output is as expected.

The debug output of `linear_log_test.rs` can be found [here](https://gist.github.com/kevtan/1b862aa166906dd3fe5189c5b5b0d215) and the debug output of `log_test.rs` can be found [here](https://gist.github.com/kevtan/78d9a567028369ee4a050895e92d4a4a).

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nano33ble/src/test/linear_log_test.rs
+++ b/boards/nano33ble/src/test/linear_log_test.rs
@@ -82,23 +82,23 @@ pub unsafe fn run(
 static TEST_OPS: [TestOp; 9] = [
     TestOp::Read,
     // Write to first page.
-    TestOp::Write(8),
-    TestOp::Write(300),
+    TestOp::Write(64),
+    TestOp::Write(2400),
     // Write to next page, too large to fit on first.
-    TestOp::Write(304),
+    TestOp::Write(2432),
     // Write should fail, not enough space remaining.
-    TestOp::Write(306),
+    TestOp::Write(2448),
     // Write should succeed, enough space for a smaller entry.
-    TestOp::Write(9),
+    TestOp::Write(72),
     // Read back everything to verify and sync.
     TestOp::Read,
     TestOp::Sync,
     // Write should still fail after sync.
-    TestOp::Write(308),
+    TestOp::Write(2464),
 ];
 
 // Buffer for reading from and writing to in the log tests.
-static mut BUFFER: [u8; 310] = [0; 310];
+static mut BUFFER: [u8; 2480] = [0; 2480];
 // Time to wait in between log operations.
 const WAIT_MS: u32 = 3;
 
--- a/boards/nano33ble/src/test/linear_log_test.rs
+++ b/boards/nano33ble/src/test/linear_log_test.rs
@@ -1,0 +1,326 @@
+//! Tests the log storage interface in linear mode. For testing in circular mode, see
+//! log_test.rs.
+//!
+//! The testing framework creates a non-circular log storage interface in flash and performs a
+//! series of writes and syncs to ensure that the non-circular log properly denies overly-large
+//! writes once it is full. For testing all of the general capabilities of the log storage
+//! interface, see log_test.rs.
+//!
+//! To run the test, add the following line to the imix boot sequence:
+//! ```
+//!     test::linear_log_test::run(mux_alarm, dynamic_deferred_caller);
+//! ```
+//! and use the `USER` and `RESET` buttons to manually erase the log and reboot the imix,
+//! respectively.
+
+use capsules::log;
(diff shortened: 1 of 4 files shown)


boards/nano33ble/src/test/linear_log_test.rs
```diff
@@ -0,0 +1,326 @@
...
+use kernel::static_init;
+use kernel::storage_volume;
+use kernel::ReturnCode;
+use nrf52840::{
+    nvmc::{NrfPage, Nvmc},
+    rtc::Rtc,
+};
+
+// Allocate 4 KiB volume for log storage (the nano33ble page size is 4 KiB).
+storage_volume!(LINEAR_TEST_LOG, 4);
```
hudson-ayers, line 34: several of the tests are specifically designed to test writes across multiple pages, so it might be better if this was 8?

boards/nano33ble/src/test/linear_log_test.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+static TEST_OPS: [TestOp; 9] = [
+    TestOp::Read,
+    // Write to first page.
+    TestOp::Write(8),
+    TestOp::Write(300),
+    // Write to next page, too large to fit on first.
+    TestOp::Write(304),
+    // Write should fail, not enough space remaining.
```
hudson-ayers, line 83: With the different page size, these comments no longer seem right.

hudson-ayers (commented):
Thanks, always good to have more tests. Does this work as-is today? Or is this somehow blocked on your virtual log PR?

As an aside, this is another good example of why Tock would benefit from having a way to write and instantiate tests capsule-level tests that are generic across boards.

boards/nano33ble/src/test/linear_log_test.rs
kevtan: Good point! The imix had page size 512 so 1 KiB was 2 pages. Since the nano33ble has page size 4096, I've allocated 8 KiB which is also 2 pages.

boards/nano33ble/src/test/linear_log_test.rs
kevtan: Updated! Thanks for bringing that to my attention.

kevtan:
@hudson-ayers This PR works as-is and is not blocked by the virtual log PR. In fact, I'd like to get this PR merged before the virtual log so that the virtual log can build on some basic log testing infrastructure for the nano 33!

boards/nano33ble/src/test/linear_log_test.rs
```diff
@@ -0,0 +1,339 @@
...
+    TestOp::Read,
+    TestOp::Sync,
+    // Write should still fail after sync.
+    TestOp::Write(2464),
+];
+
+// Buffer for reading from and writing to in the log tests.
+static mut BUFFER: [u8; 2480] = [0; 2480];
+// Time to wait in between log operations.
+const WAIT_MS: u32 = 3;
```
hudson-ayers, line 103: I would keep this a little higher given the larger writes in this test.

boards/nano33ble/src/test/log_test.rs
```diff
@@ -0,0 +1,659 @@
...
+    fn fired(&self) {
+        // Erase log.
+        self.state.set(TestState::Erase);
+        self.erase();
+    }
+}
+
+fn entry_id_to_test_value(entry_id: usize) -> u64 {
+    // Page and entry header sizes for log storage.
+    const PAGE_SIZE: usize = 512;
```
hudson-ayers, line 652: This needs to be updated to reflect the nrf52 page size

hudson-ayers (commented):
A couple small changes. Otherwise this looks good, assuming both tests pass when using the pin to manually erase the log as the instructions specify

boards/nano33ble/src/test/log_test.rs
kevtan: Thanks for catching that!

bradjc:
Looks like a build error.

kevtan:
@bradjc The CI system was complaining about a snake case warning. I've fixed this!

bradjc (approved):

PR #2225, opened 2020-11-30
arch/cortex-m: change more llvm_asm! to asm!

JOE1994:
### Pull Request Overview

This PR changes 7 invocations of `llvm_asm!` to `asm!`.
Like #2092, this PR intends to make progress towards finishing #1923 .
I made the changes after reading through the [inline-asm RFC](https://github.com/Amanieu/rfcs/blob/inline-asm/text/0000-inline-asm.md#rules) .

#### A couple of notes on the changes
*  The first two `llvm_asm!` calls with `cc` clobbers are translated to `asm!` calls that don't explicitly state `cc` clobbers.
    In `asm!`, `cc` clobbers are assumed by default. (Option `preserve_flags` can be used when `cc` clobbers don't happen)
    * reference: [Rust Zulip chat stream 'project-inline-asm'](https://rust-lang.zulipchat.com/#narrow/stream/216763-project-inline-asm/topic/Moving.20forward.20with.20the.20RFC/near/185306638)
* `volatile` is assumed by default in `asm!`, so explicit `volatile` keywords were removed.
   (Option `pure` can be used to get rid of the `volatile` semantics)
   * reference: [Link to Q&A on `volatile` from Rust Zulip chat](https://rust-lang.zulipchat.com/#narrow/stream/216763-project-inline-asm/topic/volatile.20question/near/208320361)
*  `asm!` disallows use of `bool` variables as outputs. I replaced the output `bool` variables with `u32` variables.
```shell
   error: cannot use value of type `bool` for inline assembly
   --> arch/cortex-m/src/lib.rs:409:19
    |
409 |         out("r1") kernel_stack,
    |                   ^^^^^^^^^^^^
    |
    = note: only integers, floats, SIMD vectors, pointers and function pointers can be used as arguments for inline assembly
```

### Testing Strategy
I tested this PR by building the kernel and flashing to a `Hail` board with a blinky app.
Didn't see any erroneous behavior when pressing the reset button...

### TODO or Help Wanted
I somewhat naively replaced `bool` variables with `u32` variables.
Would it be better to replace the `bool` variables with `usize` instead?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### Thank you for reviewing this PR :crab:

### DIFF
--- a/arch/cortex-m/src/lib.rs
+++ b/arch/cortex-m/src/lib.rs
@@ -395,44 +395,42 @@ unsafe fn kernel_hardfault_arm_v7m(faulting_stack: *mut u32) {
 #[naked]
 pub unsafe extern "C" fn hard_fault_handler_arm_v7m() {
     let faulting_stack: *mut u32;
-    let kernel_stack: bool;
+    let kernel_stack: u32;
 
     // First need to determine if this a kernel fault or a userspace fault.
-    llvm_asm!(
-    "
-    mov    r1, 0     /* r1 = 0 */
-    tst    lr, #4    /* bitwise AND link register to 0b100 */
-    itte   eq        /* if lr==4, run next two instructions, else, run 3rd instruction. */
-    mrseq  r0, msp   /* r0 = kernel stack pointer */
-    addeq  r1, 1     /* r1 = 1, kernel was executing */
-    mrsne  r0, psp   /* r0 = userland stack pointer */"
-    : "={r0}"(faulting_stack), "={r1}"(kernel_stack)
-    :
-    : "cc"
-    : "volatile" );
+    asm!(
+        "mov    r1, 0     /* r1 = 0 */",
+        "tst    lr, #4    /* bitwise AND link register to 0b100 */",
+        "itte   eq        /* if lr==4, run next two instructions, else, run 3rd instruction. */",
+        "mrseq  r0, msp   /* r0 = kernel stack pointer */",
+        "addeq  r1, 1     /* r1 = 1, kernel was executing */",
+        "mrsne  r0, psp   /* r0 = userland stack pointer */",
+        out("r0") faulting_stack,
+        out("r1") kernel_stack,
+        options(nomem, nostack)
+    );
 
-    if kernel_stack {
+    if kernel_stack != 0 {
         // Need to determine if we had a stack overflow before we push anything
         // on to the stack. We check this by looking at the BusFault Status
         // Register's (BFSR) `LSPERR` and `STKERR` bits to see if the hardware
         // had any trouble stacking important registers to the stack during the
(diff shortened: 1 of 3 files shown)


arch/cortex-m/src/lib.rs
```diff
@@ -395,44 +395,42 @@ unsafe fn kernel_hardfault_arm_v7m(faulting_stack: *mut u32) {
...
+        "tst    lr, #4    /* bitwise AND link register to 0b100 */",
+        "itte   eq        /* if lr==4, run next two instructions, else, run 3rd instruction. */",
+        "mrseq  r0, msp   /* r0 = kernel stack pointer */",
+        "addeq  r1, 1     /* r1 = 1, kernel was executing */",
+        "mrsne  r0, psp   /* r0 = userland stack pointer */",
+        out("r0") faulting_stack,
+        out("r1") kernel_stack,
+        options(nomem, nostack)
+    );
+
```
lschuermann, line 412: As you outlined, we can't use `bool` for outputs anymore which makes sense. Maybe it'd be helpful to insert either a comment on the `kernel_stack` variable stating that this is storing a boolean value, or adding a `debug_assert!(kernel_stack >= 1)`?

Same would go for the other `bool` to `u32` transitions.

lschuermann (commented):
Thanks for these changes! It's unfortunate that the fact that `volatile` is assumed by default is only documented in some chats, I couldn't find other references on it.

arch/cortex-m/src/scb.rs
```diff
@@ -300,8 +300,7 @@ pub unsafe fn disable_fpca() {
     SCB.cpacr
         .modify(CoprocessorAccessControl::CP10::CLEAR + CoprocessorAccessControl::CP11::CLEAR);
 
-    llvm_asm!("dsb");
-    llvm_asm!("isb");
+    asm!("dsb", "isb");
```
hudson-ayers, line 303: should this also have `options(nomem, nostack, preserves_flags)`? It seems that these barrier instructions should not be touching memory or stack or condition codes

hudson-ayers (commented):
Thanks for contributing these!

arch/cortex-m/src/lib.rs
JOE1994: Thank you for your feedback! :+1: I added a explicit comment stating that the variable is storing a boolean value.

arch/cortex-m/src/scb.rs
JOE1994: Thank you for your feedback :+1: I added the options to the `asm!` invocation :)

lschuermann (approved):
Looks good, thanks for the contribution!

hudson-ayers (approved):

bradjc (approved):

PR #2229, opened 2020-12-02
kernel: Fix potential unsoundness in dynamic grant allocation

alistair23:
### Pull Request Overview

This is a rebase of the original PR: https://github.com/tock/tock/pull/2052

In the original code, dynamic grant allocation directly returns
`Owned`. This is problematic, as if a capsule stores these Owned values
outside of another app-specific grant, then it could keep it around
after the app has crashed/restarted, and then access deallocated
memory.

This fixes that by instead returning a new `DynamicGrant` type, which
requires an `enter` call to access.

Signed-off-by: David Ross <David.Ross@wdc.com>

### Testing Strategy

CI

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -1,13 +1,12 @@
 //! Data structure to store a list of userspace applications.
 
+use crate::callback::AppId;
+use crate::process::{Error, ProcessType};
+use crate::sched::Kernel;
 use core::marker::PhantomData;
 use core::mem::{align_of, size_of};
 use core::ops::{Deref, DerefMut};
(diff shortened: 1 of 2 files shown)


alistair23:
Ping!

hudson-ayers (approved):
This conflicts with the changes in #2137 , but we still don't have consensus on what to do for that PR, so maybe we should just merge this for now?

cc @alevy

alistair23:
Ping!

lschuermann (approved):
LGTM.

alevy (approved):

PR #2232, opened 2020-12-03
Makefile: Don't require root permissions to install npm modules

alistair23:
### Pull Request Overview

Removes the requirement for root permissions when running the CI.

### Testing Strategy

Running CI.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -334,20 +334,20 @@ ci-job-clippy:
 
 define ci_setup_markdown_toc
 	$(call banner,CI-Setup: Install markdown-toc)
-	npm install -g markdown-toc
+	npm install markdown-toc
 endef
 
 .PHONY: ci-setup-markdown-toc
 ci-setup-markdown-toc:
 	$(call ci_setup_helper,\
-		command -v markdown-toc,\
-		npm install -g markdown-toc,\
+		PATH="node_modules/.bin:${PATH}" command -v markdown-toc,\
+		npm install markdown-toc,\
 		ci_setup_markdown_toc,\
 		CI_JOB_MARKDOWN)
 
 define ci_job_markdown_toc
 	$(call banner,CI-Job: Markdown Table of Contents Validation)
-	@CI=true tools/toc.sh
+	@CI=true PATH="node_modules/.bin:${PATH}" tools/toc.sh
 endef
 
 .PHONY: ci-job-markdown-toc
--- a/tools/toc.sh
+++ b/tools/toc.sh
@@ -11,7 +11,7 @@
 let ERROR=0
 
 # Find all markdown files
-for f in $(find * -name "*.md"); do
+for f in $(find . -path ./node_modules -prune -false -o -name "*.md"); do
 
 	# Only use ones that include a table of contents
 	grep '<!-- toc -->' $f > /dev/null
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -334,20 +334,20 @@ ci-job-clippy:
...
-	npm install -g markdown-toc
+	npm install markdown-toc
 endef
 
 .PHONY: ci-setup-markdown-toc
 ci-setup-markdown-toc:
 	$(call ci_setup_helper,\
-		command -v markdown-toc,\
-		npm install -g markdown-toc,\
+		PATH="node_modules/.bin:${PATH}" command -v markdown-toc,\
```
lschuermann, line 343: I'm not an expert on the Tock make-based build system at all, but I'd try to avoid modifying $PATH wherever possible. Maybe it could be a better idea to just call the binary with the full path?

alistair23: It doesn't modify `PATH` outside of the command, just for this individual call. It's also similar to what we already do to run QEMU.

lschuermann: Right. It's fine, but I'd preferably just specify the commands by their full path.

alistair23: That would require editing the `tools/toc.sh` script though, which would then break people who want to use a global install.

lschuermann: Right, that should be fine then.

Makefile
```diff
@@ -334,20 +334,20 @@ ci-job-clippy:
 
 define ci_setup_markdown_toc
 	$(call banner,CI-Setup: Install markdown-toc)
-	npm install -g markdown-toc
+	npm install markdown-toc
```
lschuermann, line 337: This needs to be installed outside of the Tock repository, as having either the `node_modules` directory present breaks `markdown-toc`:

```
********************************************************************************
* CI-Job: Markdown Table of Contents Validation                                *
********************************************************************************

/home/runner/work/tock/tock/node_modules/markdown-toc/lib/insert.js:40
    throw new Error('markdown-toc only supports one Table of Contents per file.');
    ^

Error: markdown-toc only supports one Table of Contents per file.
    at Function.insert (/home/runner/work/tock/tock/node_modules/markdown-toc/lib/insert.js:40:11)
    at /home/runner/work/tock/tock/node_modules/markdown-toc/cli.js:59:27
```

lschuermann (changes requested):

Makefile
lschuermann: Installing one directory up seems to work fine on my local machine.

alistair23: Fixed!

alistair23:
Ping!

hudson-ayers (approved):

tools/toc.sh
```diff
@@ -11,7 +11,7 @@
 let ERROR=0
 
 # Find all markdown files
-for f in $(find * -name "*.md"); do
+for f in $(find . -path ./node_modules -prune -false -o -name "*.md"); do
```
lschuermann, line 14: That expression seems interesting :smile:. It will also catch directories, so it'll probably break once we have a directory named `*.md`. It might be a good idea to add `-type f` after `-o`.

Also have you tested it on BSD `find` / macOS? I after checking the CI run and reading the macOS `find` manpage I _think_ it should work, but I don't have a machine to test on.

lschuermann (approved):

PR #2233, opened 2020-12-03
BLE Capsule based on Rubble

alistair23:
### Pull Request Overview

This PR is a rebase of the original Rubble PR, which is here: https://github.com/tock/tock/pull/2036

I'm not going to repeat the original description. I have rebased this on master and addresses all comments from the original PR.

The last discussion from #2036 was 

> However, it should be the case that out-of-tree boards can depend on the Kernel (e.g. chips / capsules / kernel / libraries) without having to download the dependency

I just checked and boards that aren't using the Rubble dependency do not list Rubble in their dependency tree (checked with `cargo-tree tree`).

### Testing Strategy

Unfortunately I don't have a board to test on. I'm going to get one, but until then this is just build tested. I'm hoping that this currently is in the same working state as https://github.com/tock/tock/pull/2036

### TODO or Help Wanted

I'm not sure...

This is currently a large PR. I would like to try and get it merged before working on expanding features. The main issue now seems to be agreement on how to ingrate Rubble.

If we could agree on next steps forward that would be great. As well as that I need to test this.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


hudson-ayers (commented):
I just tested this again on my nano33 and advertisements do work! Not sure why I didn't see them earlier, but could be an issue with my scanner app.

So this is at least close to the level of functionality we have in the non-rubble stack, with a much easier path forward to adding more functionality.

I'll try to do a full review soon, but personally I think that with a full code review this is probably suitable to merge as-is, and I think we should make this the default ble stack for at least one of the nrf52 boards (probably the nano33).

boards/nordic/nrf52_rubble/README.md
```diff
@@ -0,0 +1,87 @@
...
+Finally, add code to initialize the rubble driver to the `reset_handler` function:
+
+```rust
+
+#[no_mangle]
+pub unsafe fn reset_handler() {
+    // ...
+
+    let ble_radio =
+        nrf52_rubble::RubbleComponent::new(board_kernel, &nrf52840::ble_radio::RADIO, mux_alarm)
```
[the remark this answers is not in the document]
alistair23: Fixed!

alistair23:
Thanks for testing!

With this working I agree with it being ready to merge. When I get my boards I will work on enabling it by default (I'm not getting an Arudino though, I was thinking the nordic boards), updating the Rubble package and then adding more functionality will be next.

hudson-ayers:
Is this still blocked on #2229 ?

alistair23:
I don't think it's blocked on that, the original wasn't.

bradjc:
Do you mind if I add commits to this PR? There are some organizational and documentation things I would like to address. Or would comments be better? Overall the code looks good.

alistair23:
Go ahead!

bradjc:
I tried this PR on the nano33 and can only see the advertisements in iOS. On Android I never see advertisements, either based on the peripheral name or address.

alistair23:
That's not ideal. I'm getting a Nano33 BLE but I don't have an Android or iOS device so I won't be able to test either of those. Are you able to look into it more Brad?

I checked the Rubble issues and this might be the problem: https://github.com/jonas-schievink/rubble/issues/56

hudson-ayers (approved):

bradjc:
I manually changed the first byte of the adv buffer in the nrf52 driver (so right before the advertisement is transmitted) from 0x40 to 0x42 and i can see it on android now.

alistair23:
CI is passing now :)

alistair23:
Hopefully netlify is passing, I had to edit the Makefile:

> When two crates are included but are different versions (such as when
> including Rubble) cargo doc will print warning. This is a bug and
> documented here: rust-lang/cargo#6313
> 
> To avoid that causing netlify failures let's ignore that failure.

boards/support/tock_rubble/src/tock_rubble.rs
```diff
@@ -0,0 +1,449 @@
...
+            tx,
+            rx,
+        );
+        res.map(|next_update| match next_update {
+            rubble::link::NextUpdate::Keep => hil::rubble::types::NextUpdate::Keep,
+            rubble::link::NextUpdate::Disable => hil::rubble::types::NextUpdate::Disable,
+            rubble::link::NextUpdate::At(time) => hil::rubble::types::NextUpdate::At(
+                hil::rubble::types::Instant::from_raw_micros(time.raw_micros()),
+            ),
+        })
```
bradjc, line 441: I don't understand why the tock-rubble layer pushes the timer loop logic back down to the caller (in this case the Driver capsule). It seems like this interface layer should have to handle managing the timers for the Rubble library, and not push them though making every kernel use of Rubble have to deal with it.

boards/support/tock_rubble/src/tock_rubble.rs
```diff
@@ -0,0 +1,449 @@
...
+            let ad = rubble::link::ad_structure::AdStructure::from_bytes(&mut ad_byte_reader)?;
+            ad_structures[num_ads] = ad;
+            num_ads += 1;
+            if num_ads > 16 {
+                return Err(rubble::Error::InvalidLength);
+            }
+        }
+        let ad_structures = &ad_structures[..num_ads];
+
+        let res = self.0.start_advertise(
```
bradjc, line 428: This doesn't work on android if the device is only advertising without supporting connections.

https://github.com/jonas-schievink/rubble/blob/6518ff8e44b45a70b85468c6ae330848a2347b71/rubble/src/link/mod.rs#L259-L280

The `start_advertise()` function directly calls `PduBuf::discoverable()` which is the wrong type since we don't respond to scan requests.

Rubble only seems to expose that one start advertising function, so I'm not sure how to fix this.

bradjc (changes requested):
I'm testing on nano33. The code transmits packets for a while...and then just stops. My debug messages no longer print, and interrupts from the usb cdc stack don't seem to be handled. The BLE advertisements are no longer sent. I'm not sure what is going wrong, but I suspect the issue goes beyond adding the Rubble library.

boards/support/tock_rubble/src/tock_rubble.rs
brghena: Just a note: it's likely the scan responses that need to be supported. The android scanning application wouldn't be able to notice if a transmission was lying about supporting connections.

boards/support/tock_rubble/src/tock_rubble.rs
alistair23: I'm not sure what you mean here, can you point to the code?

boards/support/tock_rubble/src/tock_rubble.rs
alistair23: Just to confirm, do you mean support? https://github.com/jonas-schievink/rubble/blob/ed59fd68d318fa7a6febbcbc6425463db37c6a05/rubble/src/link/advertising.rs#L617

alistair23:
I just tested this on a Nano33 Sense and I can see "Rubble on Tock" from two different smartphones and a Windows machine.

boards/support/tock_rubble/src/tock_rubble.rs
hudson-ayers: I think he means this: https://github.com/jonas-schievink/rubble/blob/ed59fd68d318fa7a6febbcbc6425463db37c6a05/rubble/src/link/advertising.rs#L624

kernel/src/hil/rubble/rubble.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// This will error if the advertising data is incorrectly formatted, or if
+    /// it's too long.
+    fn start_advertise(
+        &mut self,
+        interval: Duration,
+        data: &[u8],
+        transmitter: &mut R::BleRadio,
+        tx: <R::PacketQueue as RubblePacketQueue>::Consumer,
+        rx: <R::PacketQueue as RubblePacketQueue>::Producer,
+    ) -> Result<NextUpdate, Self::Error>;
```
bradjc, line 88: Here is the interface definition for the interface tock capsules use for the Rubble stack to start advertising. This returns `NextUpdate` which the caller has to then keep track of make sure there is a timer set to fire at that time. Having operations return the time to do the next operation is how Rubble organizes events, but that seems like it should be contained in the tock-rubble adapter, and hidden from the tock capsules.

alistair23: Ah I see.

I'm guessing it's because the tock_rubble support has a few different layers (as different structs/traits) so there is no single alarm instance. In this case I think it is cleaner to just pass the `NextUpdate` to the capsule where it can all be managed in one place.

As well as that doesn't it match other capsules in Tock where the capsule is in charge of setting alarms? It also means a different capsule could in theory do different operations with the `NextUpdate`.

bradjc: Why is that managed in one place? That is very much a Rubble-only concept, and the Rubble management is then split between the tock-rubble crate and any user of BLE (which could be several capsules). It's even worse if multiple kernel capsules want to use BLE (say for app updates and to provide BLE to userspace).

alistair23: Managed in one place in the sense that there is a single alarm function.

An alarm for a BLE implementation is common and not just a Rubble concept. I would expect any BLE capsule to use an alarm.

I see what you are saying though, if we could move it out of the capsule we can have less duplication between multiple BLE related capsules.

Maybe that can be changed later? These is still lots missing from this implementation so I we'll still need more work here.

alistair23:
Ping! @bradjc do you think the current issues are blocking this PR or can they be improved in the future? I would like to get this in soon so we have a base line to work from. I am currently working on connections but will also need to work on the 2.0 port. I would like to see this merged before that though.

bradjc:
I played around with this a little bit today, and I don't think this PR in its current form is something we should merge. I'll try to explain why, and I think we have two options.

This PR essentially makes it possible to use Rubble directly inside of Tock. The bulk of the code is just allowing for renames between Tock types and Rubble types and adapting alarms so that a capsule in `/capsules` can use Rubble without actually depending on the external dependency. I suppose this shouldn't be surprising, as that was the goal and what the PR description states.

My concern is that the Rubble interface and structure is quite a bit different from how Tock works, and to use Rubble a lot of the Rubble structure has to be handled in the capsule. This results in Rubble not just being responsible for the BLE stack implementation, but also dictating a pretty substantial kernel interface. I'm not sure we want to design the Tock BLE interface around the Rubble interface. However, I do want to make progress on BLE, and we shouldn't wait for the correct interface to come along someday before we can make progress on BLE.

### Ideas on how to proceed

1. I don't think there is anything wrong with directly using Rubble on top of Tock, even in the kernel. However, the effort to rename all of the interfaces just so that capsule can go in `/capsules` seems unnecessary.
    
    So this option would be to just move the capsule to `/boards/support/tock-rubble`. Boards could still use it just as they do now. The implementation wouldn't really change. We could remove a lot of the interfaces in `kernel/hil` since `/capsules` would no longer need to use them. We also wouldn't need as much code which just renames from rubble->tock.

    This would still allow us to use Rubble, and to provide userspace with a BLE interface, and to get this going quickly. It wouldn't, however, add the Rubble interface to Tock, avoiding the confusion of having the external library's interface as a kernel interface.

2. The other option I can think of would be to work more top-down. That is, we need a syscall interface for BLE, and that interface necessarily has to be separate from the Rubble interface since it has to go through syscalls. If we could establish that interface in a capsule, then we could create a `/hil` interface that would be sufficient to implement the syscalls. Then the new BLE capsule would provide the syscall interface while only using the new high-level BLE HIL. The new BLE HIL would then be interfaced to Rubble in `/boards/support/tock-rubble`. This would also remove the need to add the Rubble interfaces to the kernel.

    The other benefit of this approach is the new HIL interface would provide a clear "upper" interface for BLE operations. Capsules like a over-the-air app updating could easily use it. Also, other BLE stack implementations (like the BLE serialization stack used by Hail and Imix) could implement it.

    We would end up with a coherent kernel stack like:

    ```
   | New BLE capsule

    hil::ble::UpperInterface

   | BLE stack (rubble or otherwise)

    hil::ble::LowerInterface

   | BLE hardware (nrf52, nrf51-over-serial, etc.
    ```

### My take

Either of these options seem good to me. @alistair23 I know you have been working on BLE connections (presumably services/attributes), and so I'm assuming that comes with a syscall interface.

One option is to go with strategy 2 with the code that currently exists. I think that would be pretty straightforward, and I could make a "hil::ble::UpperInterface" for what we have, and move the capsule over to use that.

We can also go with option 1, develop the syscall interface outside of the main capsules folder, and then move it over at some point later. I think that would get BLE up and running sooner.

alistair23:
I'm not sure I follow.

You are worried that the Rubble stack has too much impact on the Rubble HIL inside the kernel?

Tock currently has a very small BLE interface that only works for advertising on a single SoC. So I don't think the current BLE interface inside Tock is 100% either.

This PR adds a Rubble BLE interface, but I don't think it really changes the current one. I'm assuming your worry is that now we have a specific BLE related interface that is directly tied to Rubble and not generic? I see your concern, but right now Rubble is the only BLE stack we are looking at having. It seems silly to complicate everything to have a different HIL for the possibility that one day we will add a second BLE stack. I haven't been able to find any other project that supports multiple Bluetooth stacks, so I really don't think it's something we are going to want to do. We can also keep the current Tock BLE HIL and extend that as needed.

I have been working on allowing connections to Tock/Rubble. It so far doesn't really involve much interface/syscall work, mostly just changes to the driver to try and get transmit/receive working at the same time.

I do see advantages with option 2 above and I think that would be a better design. I'm worried that this will just stall trying to get the best design and never be accepted. I unfortunately cannot justify too much more time on this so I won't be able to do the port.

It sounds like you are happy to port this over to option 2 above. In that case go for it! :) I can help test and review changes.

alexandruradovici:
Sorry for jumping into this, I am trying to use Tock for one of my IoT clases and having a working BLE stack would be very important for me.

If there is no one working on this already, I could try to make some time and try rewrite this, probably with some help from @alistair23. 

Personally, I prefer option 2 as it seems more generic.

bradjc:
As for using the stack (in userspace at least) I think the main missing piece is a syscall interface that supports more than just advertisements. And then implementing the generic transmit function for the nrf52.

As for the interface, if we want to use the Rubble interface I think we should just use it directly (aka in the /boards directory). Rubble is only on version 0.0.3, and including its interface it in the Tock kernel seems premature. We also need a syscall interface, and it's not clear that the Rubble interface maps well to Tock syscalls. If it does, and the rubble interface simplifies the userspace interface, then I think that would be a good sign Tock should use the Rubble interface (or a very similar version).

alistair23:
I made some progress on allowing connections. It needed minimal changes to the BLE stack but a few changes to the nRF driver. I can share them somewhere if you are interested. I didn't get them fully working though, but I think it is close.

As for porting to option 2 I was under the impression that @bradjc was going to do that. If you would like to @alexandruradovici  then go for it.

alexandruradovici:
> As for porting to option 2 I was under the impression that @bradjc was going to do that. If you would like to @alexandruradovici then go for it.

If @bradjc has time, that is great. I am just offering to help if needed to get this done as faster :)

bradjc:
I'd love to see the BLE stack working, but don't block on me to write a new BLE HIL or syscall interface.

alistair23:
If we don't want to block on anyone is it worth merging this then?

alistair23:
Ping!

boards/Makefile.common
```diff
@@ -245,7 +245,7 @@ debug-lst:  $(TARGET_PATH)/debug/$(PLATFORM).lst
 doc: | target
 	@# This mess is all to work around rustdoc giving no way to return an
 	@# error if there are warnings. This effectively simulates that.
-	$(Q)RUSTDOCFLAGS='-Z unstable-options --document-hidden-items -D warnings' $(CARGO) --color=always doc $(VERBOSE) --release --package $(PLATFORM) --target-dir=$(TARGET_DIRECTORY) 2>&1 | grep -C 9999 warning && (echo "Warnings detected during doc build" && if [[ $$CI == "true" ]]; then echo "Erroring due to CI context" && exit 33; fi) || if [ $$? -eq 33 ]; then exit 1; fi
+	$(Q)RUSTDOCFLAGS='-Z unstable-options --document-hidden-items -D warnings' $(CARGO) --color=always doc $(VERBOSE) --release --package $(PLATFORM) --target-dir=$(TARGET_DIRECTORY) 2>&1 | grep warning | grep -v "output filename collision" && (echo "Warnings detected during doc build" && if [[ $$CI == "true" ]]; then echo "Erroring due to CI context" && exit 33; fi) || if [ $$? -eq 33 ]; then exit 1; fi
```
ppannuto, line 248: Should we not fix duplicate filenames rather than just silencing this? Especially as this is a global check, not a silencing unique to the issue here.

ppannuto (changes requested):
I'm catching up on a lot Tock, but from a read over of this, I think I share the concerns of being protective of the `kernel::hil` namespace. I don't think it makes sense to have a `hil::rubble`, but maybe I'm missing why that is necessary.

I do think that (eventually) a generic BLE syscall would be great, but that can hold off. The fastest path I see to merging this is just pulling more of it into the support folder -- it lets folks opt-in to a working BLE instance today, but makes it clear that we aren't really committed to this interface.

skannan89:
Any updates? Would love to see BLE support.

PR #2234, opened 2020-12-04
nano 33 ble & adafruit clue increase the number of deferred calls slots

alexandruradovici:
### Pull Request Overview

This pull request fixes the number of deferred calls slots for Nano 33 BLE and CLUE. Boards panic without this due to lack of enough slots.

### Testing Strategy

This pull request was tested using the Nano 33 BLE and CLUE boards.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/clue_nrf52840/src/main.rs
+++ b/boards/clue_nrf52840/src/main.rs
@@ -246,7 +246,7 @@ pub unsafe fn reset_handler() {
     //--------------------------------------------------------------------------
 
     let dynamic_deferred_call_clients =
-        static_init!([DynamicDeferredCallClientState; 2], Default::default());
+        static_init!([DynamicDeferredCallClientState; 3], Default::default());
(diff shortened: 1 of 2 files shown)


lschuermann (approved):
Right, the additional deferred call client was probably introduced by #2221.

hudson-ayers (approved):

alexandruradovici:
Yes, most probably.

hudson-ayers:
Yep. I should have increased it. Would be nice if DeferredCall worked such that it was impossible to not have enough slots. 

bors r+

lschuermann:
> Would be nice if DeferredCall worked such that it was impossible to not have enough slots.

I actually thought of this while designing it but couldn't come up with a solution that would also provide good ergonomics otherwise. Passing clients statically from the array etc. all have their own issues unfortunately.

PR #2237, opened 2020-12-04
Revise the description of Subscribe's semantics to match its implementation.

jrvanwhy:
The implementation of `subscribe` in the kernel is unable to guarantee that a callback is not called after a subsequent `subscribe` call. Instead, it works by passing drivers non-copyable Callback types which drivers can use to invoke callbacks. This updates the TRD to match the implementation.

The new wording highlights the difference between what drivers *can* do and what they are *supposed to* do. This is important for memory safety in the context of Tock's threat model: an app must not encounter undefined behavior if a driver misbehaves, although it is okay for driver misbehavior to result in denial of service.

This change doesn't change the behavior of a "well-behaved" driver; it does require apps to be somewhat tolerant to non-well-behaved drivers.

IIRC, there is an ongoing discussion about whether or not capsules should be able to create instances of the Null Callback. I don't have an opinion on that topic, although I did allow for that in the wording of this PR. I'm open to changing the wording if we decide that capsules should not be able to construct Null Callback handles.

When we reach consensus on this PR, I will open a similar PR to update the description of the allow system calls.

### Testing Strategy

No testing done (`make prepush` threw unrelated errors)

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/doc/reference/trd-syscalls.md
+++ b/doc/reference/trd-syscalls.md
@@ -309,12 +309,23 @@ memory and is not the Null Callback described below), the kernel MUST
 NOT invoke the requested driver and MUST immediately return a failure
 with a return code of EINVAL.
 
-A passed callback MUST be valid until the next invocation of `subscribe`
-with the same syscall and driver identifier. When userspace invokes
-subscribe, the kernel MUST cancel all pending callbacks for that driver
-and subscribe identifier: it MUST NOT invoke the previous callback after
-the call to subscribe, and MUST NOT invoke the new callback for events
-that occurred before the call to subscribe. 
+When the kernel invokes a driver, it passes a handle representing the passed
+callback to the driver. The driver is then responsible for storing the handle,
+and MUST return a handle, which is returned to the caller of `subscribe`. With
+the exception of a handle to the Null Callback, capsules cannot create or
+duplicate these handles.
+
+A passed callback MUST remain valid until all handles to that callback have been
+returned to userspace. When userspace invokes subscribe, the kernel MUST cancel
+all pending callbacks for that driver and subscribe identifier: it MUST NOT
+invoke the returned callback after the call to subscribe.
+
+Drivers SHOULD implement swapping semantics for the subscribe handles for each
+subscribe identifier. Conceptually, on each `subscribe` call drivers SHOULD keep
+the new handle and return the previous handle for the subscribe identifier.
+Drivers SHOULD start with their handles initialized to the Null Callback.
+Drivers SHOULD NOT use a callback handle in response to an event that occurred
+before the handle was created.
 
 Note that these semantics create a period over which callbacks might
 be lost: any callbacks that were pending when `subscribe` was called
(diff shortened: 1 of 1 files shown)


lschuermann:
I'm not sure whether we've reached a stage at which we've reached consensus on whether the kernel *will* or *will not* guarantee (verify) that a capsule passes back the correct (swapped) pointer and userdata to a Subscribe call, by comparing against stored `(process id, driver num, subscribe num, pointer)` tuples.

*If* we decide that we want to verify this for Allow system calls and find a sufficiently efficient implementation, it's not unreasonable to also enforce correct capsule behavior at runtime for Subscribe system calls, making this a guarantee instead of a notion of correct behavior.

It's unfortunate that we couldn't discuss this on this week's call. I'll be sure (and check with @alevy) that it lands on next week's agenda. I'm currently trying different implementation strategies for such Allow/Subscribe tables and analyze their complexity / RAM impact.

lschuermann:
I think I found a solution which would allow us to *guarantee* that any Subscribe system call will either return the null callback on the first call for a given `(process,capsule,subscribenum)` combination and return the previous values on any subsequent successful calls. This would have a memory overhead of 64 bit per Callback-struct (32 bit if we'd limit the capsule num to 24 bit and the subscribenum to 8 bit) and a negligible compute complexity. Making this guarantee for Subscribe is much easier than it is for Allow, as we don't need a linearly searchable index of Callbacks. I'll try to build a proof of concept this week.

jrvanwhy:
I didn't realize this aspect of the design was a work in progress!

lschuermann:
> I didn't realize this aspect of the design was a work in progress!

I'm not certain if @phil-levis will agree with my reasoning here either, but from what I know we are at a point where can give a weak guarantee that the interfaces of traits and types with which capsules deal are somewhat stable (and should they require changes we're probably the ones to clean up the mess :smile:). Also, we are pretty sure that the semantics of what capsules are allowed to do, and more importantly what they aren't (but is technically possible in the type system) is quite clear (e.g. this exact issue).

The implementation of the types themselves though might still be subject to change (without touching the public interface). The buffer-aliasing protection for instance is not implemented on the branch yet and the design is still in the works. Depending on the design outcomes (and the potential performance impacts) we would decide what the kernel can actually guarantee vs. what it has to trust capsules on. Hope that makes sense!

doc/reference/trd-syscalls.md
```diff
@@ -309,12 +309,23 @@ memory and is not the Null Callback described below), the kernel MUST
...
+and MUST return a handle, which is returned to the caller of `subscribe`. With
+the exception of a handle to the Null Callback, capsules cannot create or
+duplicate these handles.
+
+A passed callback MUST remain valid until all handles to that callback have been
+returned to userspace. When userspace invokes subscribe, the kernel MUST cancel
+all pending callbacks for that driver and subscribe identifier: it MUST NOT
+invoke the returned callback after the call to subscribe.
+
+Drivers SHOULD implement swapping semantics for the subscribe handles for each
```
phil-levis, line 323: This should be a MUST, right? We can't statically check it, any more than we can statically check passed callbacks being valid. But we can state what the expectations are. A SHOULD means that it's OK if an implementation doesn't, as long as they really think they know what they're doing.

phil-levis (changes requested):

doc/reference/trd-syscalls.md
lschuermann: Yes, if we're following the precise definitions of RFC 2119 a "SHOULD" is a too weak statement here and "MUST" seems more appropriate.

lschuermann:
Based on the core team call on 2020/12/11, there was a strong push towards checking these constraints in the kernel. I found that for `Callback`, if we remove the current `Copy` implementation, that is possible without any memory impact and minimal complexity, so my feeling is that we're most likely going to check this on every board (i.e. make this a hard check in the kernel).

If we change this PR to say that capsules MUST behave this way, this is in line with these changes. Then, per definition, every capsule not behaving this way is incorrect and MAY be blocked by the kernel or cause a kernel panic if we are to implement a check. Therefore, I see this PR as a valuable addition nonetheless.

jrvanwhy:
I think this is obsoleted by #2405 -- and regardless this discussion is still ongoing.

PR #2239, opened 2020-12-06
net/ipv6: remove stale module, remove redundant module hierarchy

lschuermann:
### Pull Request Overview

This pull request
- removes the file `capsules/src/net/ip.rs`, as it seems to be an older version of the actively used `capsules/src/net/ipv6/ipv6.rs` (`ip.rs` was not referenced anywhere)
- moves the contents of `capsules/src/net/ipv6/ipv6.rs` to `capsules/src/net/ipv6/mod.rs`, to remove the redundant module specification `ipv6::ipv6`
- do the same for `udp`
- do the same for `icmpv6`

### Testing Strategy

### TODO or Help Wanted

The redundant module specification appears to be a common pattern in the networking stack. If this change is deemed good, I'll make sure to add commits changing the other respective modules in the networking stack accordingly.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


lschuermann:
The failing netlify build does not seem to be caused by this PR and is solved in #2240. The merge commit should resolve the failing build.

hudson-ayers (dismissed):
Thanks, this was overdue. The `ip.rs` vs `ipv6.rs` duplication in particular always bothered me but I never looked into it.

jrvanwhy:
> This is definitely something I don't understand about Rust. Why is it advantageous to put code in files called mod.rs, and then have many different mod.rs (or lib.rs) files? I think it is easier to find code in a file called "ipv6.rs". I'm also not a fan of moving code since it breaks github history, but I think that is just a me thing.

Rust made [made some changes that allow you to avoid the `mod.rs` name](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html#no-more-modrs) a couple years ago. I have been using `mod.rs` to be consistent with the existing codebase but I personally prefer the new naming scheme. A simple rename shouldn't break things too bad in history (assuming Git recognizes the rename).

However, you're still forced to put nested submodules into directories, which is annoying.

hudson-ayers:
Unfortunately, git seems not to be recognizing the rename here :/

lschuermann:
> Rust made [made some changes that allow you to avoid the `mod.rs` name](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html#no-more-modrs) a couple years ago.

Subjectively speaking this makes it even worse, as now code in a module is in part contained in a file with the module's name, and in part in a subdirectory of the same name. Sure, the `mod.rs` name may be unfortunate, but it's all grouped in a directory then. I'm fine writing code either way though, so if you like this better @bradjc I'll change it accordingly.

> Unfortunately, git seems not to be recognizing the rename here :/

That is probably because I needed to delete the line importing the nested module (as it's now contained one-level up). I'll be sure to rewrite the history in a way that this is two commits, which should allow Git to track the rename accordingly (don't want to steal anyone's credit / take anyone's git-blame). The intermediate commit will not compile, but the patch set in it's entirety will (which is what's important).

alistair23 (dismissed):

lschuermann:
@hudson-ayers Taking yet another trip into the world of Git, it seems as if it does not store renames in any special way but rather detects them when doing `git diff` or similar operations. There's a threshold to doing so (when it's not just a plain rename but a few lines have changed too), but that only effects the diffs git shows, not at all how a commit is stored or processed.

More importantly, even if a file is renamed in Git, `git blame` will still show the renaming commit as affecting the entire file. Which makes sense, after all not showing the rename-commit to affect the file would be a lie, as the file path can have semantic meaning in the code beyond just the place where it's stored.

What does this mean for this PR though? Sure, it's not elegant to "steal credit" (at least in terms of git blame), however those measures are highly arbitrary anyways and the original author and history changes are still contained in the branch history. A decision is more far-reaching than this simple module restructuring and influences what contributions to Tock are welcome vs. what aren't (which I'd adhere to). What do you think @bradjc?

hudson-ayers:
To be clear, I am not worried at all about stolen credit, its just nice when the GitHub UI can be used to track down the PR associated with a change. When trying to understand the justification behind some of the code in `grant.rs` I found that very helpful. But personally I am not super worried about that for these files.

bradjc:
I don't see the benefit to moving a bunch of code around, when just adding `pub use` can do the same thing without a large diff. I found that approach to be useful like here: https://github.com/tock/tock/pull/2233/files#diff-3ba4ca48885bffab15d8e3adc8fca89da108cc0a321c6577043e904037124558R35

lschuermann:
> I don't see the benefit to moving a bunch of code around, when just adding `pub use` can do the same thing without a large diff.

I think that should work here :+1:. Will change the PR accordingly. Here the filesystem paths do make sense, so a rename is not strictly required, however the module system should still be fixed.

lschuermann:
@bradjc Updated this to reexport the exports of the nested modules using `pub use`. I agree this is the best approach here, as the files still have expressive names (matching those of the outer modules).

hudson-ayers (approved):
Thanks again

bradjc (approved):

PR #2244, opened 2020-12-07
Adafruit Clue SHT3x temperature and humidity sensor and I2C fix

cosmindanielradu19:
### Pull Request Overview

This pull request adds support for SHT3x sensor to be used for temperature and humidity by Adafruit Clue.
This pull request fixes the I2C used by this board.
This pull request adds humidity component.

### Testing Strategy

This pull request was tested by using an Adafruit Clue.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/clue_nrf52840/Makefile
+++ b/boards/clue_nrf52840/Makefile
@@ -10,7 +10,6 @@ ifdef PORT
   FLAGS += -p $(PORT)
 endif
 
-APP=../../../libtock-c/examples/sensors/build/cortex-m4/cortex-m4.tbf
 KERNEL=$(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
 KERNEL_WITH_APP=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/debug/$(PLATFORM)-app.elf
 
@@ -18,6 +17,9 @@ KERNEL_WITH_APP=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/debug/$(PLATFORM)-app.el
 .PHONY: program program-apps
 
 program: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).bin
+	ifndef PORT
+		$(error Please specify the serial port using the PORT variable) 
+	endif
 	adafruit-nrfutil dfu genpkg --dev-type 0x0052 --sd-req 0x00B6 --application $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).bin $(KERNEL).zip
 	echo "Trying to reset device"
 	stty -F $(PORT) 1200 && sleep 0.5 && stty -F $(PORT) 1200
@@ -25,6 +27,12 @@ program: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).bin
 
 # Upload the kernel and apps using nrfutil
 program-apps: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
+	ifndef PORT
+		$(error Please specify the serial port using the PORT variable) 
+	endif
+	ifndef APP
+		$(error Please define the APP variable with the TBF file to flash an application) 
+	endif
 	arm-none-eabi-objcopy --update-section .apps=$(APP) $(KERNEL) $(KERNEL_WITH_APP)
 	arm-none-eabi-objcopy --output-target=binary -S $(KERNEL_WITH_APP) $(KERNEL_WITH_APP).bin
 	adafruit-nrfutil dfu genpkg --dev-type 0x0052 --sd-req 0x00B6 --application $(KERNEL_WITH_APP).bin $(KERNEL_WITH_APP).zip
--- a/boards/clue_nrf52840/Makefile
+++ b/boards/clue_nrf52840/Makefile
@@ -10,7 +10,7 @@ ifdef PORT
   FLAGS += -p $(PORT)
 endif
 
-APP=../../../libtock-c/examples/screen/build/cortex-m4/cortex-m4.tbf
+APP=../../../libtock-c/examples/sensors/build/cortex-m4/cortex-m4.tbf
 KERNEL=$(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
 KERNEL_WITH_APP=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/debug/$(PLATFORM)-app.elf
 
--- a/boards/clue_nrf52840/README.md
+++ b/boards/clue_nrf52840/README.md
@@ -94,7 +94,7 @@ and the physical elements on the CLUE nRF52480 board.
 
 The following sensors are working so far
 
-| Sensorr | Physical Element    |
+| Sensor | Physical Element    |
 |-------------------|---------------------|
 | Proximity           | APDS9960              |
 | Temperature & Humidity | SHT31 |
--- a/boards/clue_nrf52840/README.md
+++ b/boards/clue_nrf52840/README.md
@@ -55,12 +55,12 @@ This previous step will create a TAB (`.tab` file) that normally tockloader
 would use to program on the board. However, tockloader is currently not
 supported.
 
-Define the `APP` variable in to point towards the TBF file.
+Define the `APP` variable to point towards the TBF file.
 
 At this point you should be able to simply run `make program-apps` in this directory
 to install a fresh kernel with the app(s).
 
-You will need to specify the port and the app as shown in thhe example:
+You will need to specify the port and the app as shown in the example:
 
 ```
 $ make program-apps PORT=<serial port path> APP=<path to TBF>
--- a/boards/clue_nrf52840/README.md
+++ b/boards/clue_nrf52840/README.md
@@ -55,20 +55,15 @@ This previous step will create a TAB (`.tab` file) that normally tockloader
 would use to program on the board. However, tockloader is currently not
 supported.
 
-Modify the `APP` variable in the `Makefile` to point towards the TBF file.
-
-```makefile
-APP=../../../libtock-c/examples/screen/build/cortex-m4/cortex-m4.tbf
-
-```
+Define the `APP` variable in to point towards the TBF file.
 
 At this point you should be able to simply run `make program-apps` in this directory
 to install a fresh kernel with the app(s).
 
-You will need to specify the port like so:
+You will need to specify the port and the app as shown in thhe example:
 
 ```
-$ make program-apps PORT=<serial port path>
+$ make program-apps PORT=<serial port path> APP=<path to TBF>
 ```
 
 ### Userspace Resource Mapping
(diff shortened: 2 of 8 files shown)


boards/clue_nrf52840/README.md
```diff
@@ -90,6 +90,16 @@ and the physical elements on the CLUE nRF52480 board.
 | LED[0]            | Red LED Red         |
 | LED[1]            | White LEDs          |
 
+### Sensors
+
+The following sensors are working so far
+
+| Sensorr | Physical Element    |
```
bradjc, line 97:
```suggestion
| Sensor | Physical Element    |
```

boards/clue_nrf52840/src/main.rs
```diff
@@ -355,6 +361,28 @@ pub unsafe fn reset_handler() {
...
+        ),
+    );
+    sht3x.reset();
+
+    let temperature = static_init!(
+        capsules::temperature::TemperatureSensor<'static>,
+        capsules::temperature::TemperatureSensor::new(sht3x, board_kernel.create_grant(&grant_cap))
+    );
+
+    kernel::hil::sensors::TemperatureDriver::set_client(sht3x, temperature);
```
bradjc, line 377: We do have a temperature sensor capsule if you want to use that.

capsules/src/sht3x.rs
```diff
@@ -0,0 +1,308 @@
+//! Driver for SHT3x Temperature and Himidity Sensor
```
bradjc, line 1:
```suggestion
//! Driver for SHT3x Temperature and Humidity Sensor
```

boards/clue_nrf52840/src/main.rs
cosmindanielradu19: we use it

hudson-ayers: I am guessing Brad meant to say "component" not "capsule"

cosmindanielradu19: I fixed this and I added humidity component also.

boards/clue_nrf52840/Makefile
```diff
@@ -10,7 +10,7 @@ ifdef PORT
   FLAGS += -p $(PORT)
 endif
 
-APP=../../../libtock-c/examples/screen/build/cortex-m4/cortex-m4.tbf
+APP=../../../libtock-c/examples/sensors/build/cortex-m4/cortex-m4.tbf
```
lschuermann, line 13: You should probably avoid updating this variable here?
It's probably best to not hardcode any `APP` variable here at all, but to just take it from the environment as [earlgrey-nexysvideo](https://github.com/tock/tock/blob/b939473ad41f9723605155add18e92cf3d78cd53/boards/earlgrey-nexysvideo/README.md#programming-apps) does.

lschuermann (changes requested):

boards/clue_nrf52840/Makefile
lschuermann: Thanks!

bradjc (approved):

capsules/src/sht3x.rs
```diff
@@ -0,0 +1,308 @@
...
+            read_hum: Cell::new(false),
+            alarm: alarm,
+        }
+    }
+
+    pub fn clear_status(&self) {
+        self.buffer.take().map(|buffer| {
+            self.i2c.enable();
+
+            buffer[0] = ((Registers::CLEARSTATUS as u16) >> 8) as u8;
```
lschuermann, line 106: This could probably made into a

```rust
let [high, low] = u16::to_be_bytes(Registers::CLEARSTATUS);
```

which I find more expressive, but either way is fine. :+1:

capsules/src/sht3x.rs
```diff
@@ -0,0 +1,308 @@
...
+            }
+
+            self.i2c.write(buffer, 2);
+
+            self.state.set(State::Reset);
+        });
+    }
+
+    pub fn is_heater_enabled(&self) -> bool {
+        false
```
lschuermann, line 155: This does not seem like it returns the current state, as there is a function to turn the heater on?

capsules/src/sht3x.rs
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    pub fn read_humidity(&self) -> ReturnCode {
```
lschuermann, line 158: Should these functions be public? I'm guessing one is supposed to access them through the HILs?

lschuermann (changes requested):

capsules/src/sht3x.rs
cosmindanielradu19: I fixed this, but still don't really know where I should return the value of the HEATER register

capsules/src/sht3x.rs
cosmindanielradu19: You are right, I updated the code so the functions are not public anymore

capsules/src/sht3x.rs
lschuermann: That's a good question! The datasheet indicates that the heater is for plausibility checking only, so extending the temperature HIL to incorporate such functionality (with an additional callback) is probably not a good idea.

If you want to support heater operation, there are two different approaches:
- store the heater state in an internal variable to make reporting its state a synchronous operation
- have a callback method to call

Given that there seem to be other operations (such as a reset) which are carried out asynchronously, it might be useful for a user of the device to know when they're finished. Otherwise the users would need to spin-wait on the result or approximate it using timers (and then check accordingly). Therefore I would likely go the route of adding a device-specific client (e.g. `&'a dyn SHT3xClient`) having methods to signal the heater state changed, what the current heater state is and that the device was reset.

Implement this at your discretion. I'd also be fine if the SHT3x couldn't be reset and the heater turned on.

cosmindanielradu19: My suggestion is to leave this PR without reset and heater, and I will come back later with another PR that will implement the client for SHT3x.

lschuermann: Sure, that works for me.

capsules/src/sht3x.rs
```diff
@@ -0,0 +1,242 @@
...
+                self.read_temp_hum()
+            } else {
+                self.read_temp.set(true);
+                ReturnCode::SUCCESS
+            }
+        }
+    }
+
+    fn read_temp_hum(&self) -> ReturnCode {
+        self.buffer.take().map_or(ReturnCode::ENOMEM, |buffer| {
```
lschuermann, line 129: From how I understand the driver, the buffer should only be `None` when an I2C operation is currently running. Wouldn't `EBUSY` be a more appropriate `ReturnCode` then?

If you never expect this function to be called directly (always using `read_temperature` and `read_humidity` instead), and the buffer should therefore never be taken on this function invocation, I think it's okay to panic here (as there's a bug in the driver). This could be done using something like `.map_or_else(|| panic("your message"), |buffer| {...})`.

What option you choose is up to you.

boards/clue_nrf52840/README.md
```diff
@@ -55,20 +55,15 @@ This previous step will create a TAB (`.tab` file) that normally tockloader
 would use to program on the board. However, tockloader is currently not
 supported.
 
-Modify the `APP` variable in the `Makefile` to point towards the TBF file.
-
-```makefile
-APP=../../../libtock-c/examples/screen/build/cortex-m4/cortex-m4.tbf
-
-```
+Define the `APP` variable in to point towards the TBF file.
```
lschuermann, line 58: Remove the "in".

boards/clue_nrf52840/README.md
```diff
@@ -55,20 +55,15 @@ This previous step will create a TAB (`.tab` file) that normally tockloader
...
 
 At this point you should be able to simply run `make program-apps` in this directory
 to install a fresh kernel with the app(s).
 
-You will need to specify the port like so:
+You will need to specify the port and the app as shown in thhe example:
```
lschuermann, line 63: "thhe" -> "the"

capsules/src/sht3x.rs
cosmindanielradu19: The update is done, everything is ready to go

lschuermann (commented):
Unfortunately I don't have hardware to test this on, but this seems very close! Added some minor comments / improvements.

cosmindanielradu19:
> Unfortunately I don't have hardware to test this on, but this seems very close! Added some minor comments / improvements.

I tested it by myself with sensors example from libtock-c

lschuermann (dismissed):
LGTM.

boards/components/src/sht3x.rs
```diff
@@ -0,0 +1,100 @@
...
+//! let sht3x = components::sht3x::SHT3xComponent::new(sensors_i2c_bus, mux_alarm).finalize(
+//!         components::sht3x_component_helper!(nrf52::rtc::Rtc<'static>),
+//!     );
+//! sht3x.reset();
+//! ```
+//!
+//! With a specified i2c address
+//! ```rust
+//! let sht3x = components::sht3x::SHT3xComponent::new(sensors_i2c_bus, mux_alarm).finalize(
+//!         components::sht3x_component_helper!(nrf52::rtc::Rtc<'static>, capsules::sht3x::BASE_ADDR << 1),
```
hudson-ayers, line 19: is the address shift in this example still needed?

capsules/src/sht3x.rs
```diff
@@ -0,0 +1,245 @@
...
+                                self.humidity_client.map(|cb| cb.callback(usize::MAX));
+                            }
+                        }
+                        self.buffer.replace(buffer);
+                        self.state.set(State::Idle);
+                    }
+                    State::Read => {
+                        self.buffer.replace(buffer);
+                        let interval = A::ticks_from_ms(20);
+                        self.alarm.set_alarm(self.alarm.now(), interval);
```
hudson-ayers, line 206: can you explain why you set an alarm here, rather than beginning the next operation immediately?

boards/components/src/sht3x.rs
cosmindanielradu19: Yes, in the main is used in the same way, because i2c address must be fixed to use the proper address.

capsules/src/sht3x.rs
cosmindanielradu19: The sensor needs 20 ms between the measurement command and the read command.

hudson-ayers (dismissed):

boards/components/src/sht3x.rs
lschuermann: Are you sure that still applies after #2272?

cosmindanielradu19: I didn't see #2272, now I rebased and there shouldn't be any conflicts.

hudson-ayers (dismissed):

boards/components/src/sht3x.rs
lschuermann: It seems you've adjusted the main accordingly (to not require a shift), although the comment is still out of date?

cosmindanielradu19: Sorry, I forgot about the comment. I solved it now.

hudson-ayers (approved):

lschuermann (approved):

bradjc (approved):

Issue #2235, opened 2020-12-04
Porting system call capsules to Tock 2.0 system call API


phil-levis:
We've now reached a point when we're almost ready to port all of the system call capsules to the new Tock 2.0 system call API. All of this work is going on in the [tock-2.0-dev branch](https://github.com/tock/tock/tree/tock-2.0-dev).

This issue is tracking who is porting which capsules and their status. I've clustered them by complexity. Please sign up to port and check off when done and tested. We have a [porting guide](https://github.com/tock/tock/blob/tock-2.0-dev/doc/Porting_Capsules_2.md) that you are free to update/refine as you learn new things.

### Complex/Tricky
- [x]  adc (@alexandruradovici): tricky because of the buffering/semantics of high speed sampling, plus there are two implementations (virtualized and dedicated)
- [x] alarm (@phil-levis): tricky because virtualization is not FIFO, traversing app states in a complex way
- [x] app_flash_driver (@lschuermann): tricky because there's careful buffer management/copies
- [x] ble_advertisement (@alevy): tricky because there's a complex state machine and buffer management
- [x] console (@phil-levis): tricky due to complex reads/buffers and virtualization behavior
- [x] ctap (@alistair23 ): tricky because there are many different operations and buffer management -- **Needs testing before merging into master.**
- [x] ~~hd44780~~ text_screen (@alexandruradovici ): complex protocols with the display
- [x] i2c_master_slave_driver (@phil-levis ):  Handles many different I2C cases, not commonly used
- [x] ieee802154/driver (@hudson-ayers): RX and TX paths, 3 buffers, and a large number of commands 
- [x] ipc (@alevy): Complex buffer management and data access.
- [x] net/udp/driver (@hudson-ayers): Three buffers, and large/complex command set
- [x] spi_controller (@phil-levis): Two buffers in concurrent use (bounds checks), one of which is optional
- [x] spi_peripheral (@phil-levis): Two buffers in concurrent use (bounds checks), one of which is optional.

### Medium/Non-trivial
- [x] button (@alevy): Needs a bunch of extra checks and issues of pending events
- [x] crc (@phil-levis ): has buffers
- [x] gpio (@phil-levis ): nothing super-complicated but has all of the moving parts and lots of operations
- [x] gpio_async (@hudson-ayers): nothing super-complicated but has all of the moving parts and lots of operations
- [x] hmac (@alistair23 ): buffers and some complex logic -- **Needs testing before merging into master.**
[...]

PR #2245, opened 2020-12-07
Update analog comparator to tock 2.0 syscalls

hudson-ayers:
### Pull Request Overview

This pull request ~~adds a helper function for converting `ReturnCode` to `CommandResult`, and~~ updates the analog comparator to use the Tock 2.0 system calls.

~~Note the somewhat unusual approach to swapping the passed callback, which I elaborate on in #2235.~~

### Testing Strategy

This pull request was tested by flashing the analog comparator test app on the `tock-2.0-dev` branch of `libtock-c`.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/analog_comparator.rs
+++ b/capsules/src/analog_comparator.rs
@@ -39,7 +39,7 @@ pub const DRIVER_NUM: usize = driver::NUM::AnalogComparator as usize;
 
 use core::cell::Cell;
 use kernel::hil;
-use kernel::{AppId, Callback, LegacyDriver, ReturnCode};
+use kernel::{AppId, Callback, CommandResult, Driver, ErrorCode, ReturnCode};
 
 pub struct AnalogComparator<'a, A: hil::analog_comparator::AnalogComparator<'a> + 'a> {
     // Analog Comparator driver
@@ -47,7 +47,7 @@ pub struct AnalogComparator<'a, A: hil::analog_comparator::AnalogComparator<'a> 
     channels: &'a [&'a <A as hil::analog_comparator::AnalogComparator<'a>>::Channel],
 
     // App state
-    callback: Cell<Option<Callback>>,
+    callback: Cell<Callback>,
 }
 
 impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> AnalogComparator<'a, A> {
@@ -61,7 +61,7 @@ impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> AnalogComparator<'a, A
             channels,
 
             // App state
-            callback: Cell::new(None),
+            callback: Cell::new(Callback::default()),
         }
     }
 
@@ -104,7 +104,7 @@ impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> AnalogComparator<'a, A
     }
 }
 
-impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> LegacyDriver for AnalogComparator<'a, A> {
+impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> Driver for AnalogComparator<'a, A> {
     /// Control the analog comparator.
     ///
     /// ### `command_num`
@@ -119,8 +119,8 @@ impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> LegacyDriver for Analo
     /// - `3`: Stop interrupt-based comparisons.
     ///        Input x chooses the desired comparator ACx (e.g. 0 or 1 for
     ///        hail, 0-3 for imix)
-    fn command(&self, command_num: usize, channel: usize, _: usize, _: AppId) -> ReturnCode {
-        match command_num {
+    fn command(&self, command_num: usize, channel: usize, _: usize, _: AppId) -> CommandResult {
+        let retcode = match command_num {
             0 => ReturnCode::SuccessWithValue {
                 value: self.channels.len() as usize,
             },
@@ -132,24 +132,25 @@ impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> LegacyDriver for Analo
             3 => self.stop_comparing(channel),
 
             _ => ReturnCode::ENOSUPPORT,
-        }
+        };
+        retcode.into()
     }
 
     /// Provides a callback which can be used to signal the application
     fn subscribe(
         &self,
         subscribe_num: usize,
-        callback: Option<Callback>,
+        mut callback: Callback,
         _app_id: AppId,
-    ) -> ReturnCode {
+    ) -> Result<Callback, (Callback, ErrorCode)> {
         match subscribe_num {
             // Subscribe to all interrupts
             0 => {
-                self.callback.set(callback);
-                ReturnCode::SUCCESS
+                self.callback.swap(Cell::from_mut(&mut callback));
+                Ok(callback)
             }
             // Default
-            _ => ReturnCode::ENOSUPPORT,
+            _ => Err((callback, ErrorCode::NOSUPPORT)),
         }
     }
 }
@@ -159,8 +160,6 @@ impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> hil::analog_comparator
 {
     /// Callback to userland, signaling the application
     fn fired(&self, channel: usize) {
-        self.callback
-            .get()
-            .map_or_else(|| false, |mut cb| cb.schedule(channel, 0, 0));
+        self.callback.get().schedule(channel, 0, 0);
     }
 }
--- a/capsules/src/analog_comparator.rs
+++ b/capsules/src/analog_comparator.rs
@@ -67,17 +67,15 @@ impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> AnalogComparator<'a, A
     }
 
     // Do a single comparison on a channel
-    fn comparison(&self, channel: usize) -> ReturnCode {
+    fn comparison(&self, channel: usize) -> Result<bool, ErrorCode> {
         if channel >= self.channels.len() {
-            return ReturnCode::EINVAL;
+            return Err(ErrorCode::INVAL);
         }
         // Convert channel index
         let chan = self.channels[channel];
         let result = self.analog_comparator.comparison(chan);
 
-        ReturnCode::SuccessWithValue {
-            value: result as usize,
-        }
+        Ok(result)
     }
 
     // Start comparing on a channel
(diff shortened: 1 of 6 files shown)


kernel/src/driver.rs
```diff
@@ -149,6 +150,16 @@ impl CommandResult {
     }
 }
 
+impl From<ReturnCode> for CommandResult {
```
phil-levis, line 153: This is a bit tricky. One of the requirements of commands in 2.0 is that they have a single failure return type and a single success return type. Using this `From` means one could easily inadvertently violate this requirement. I'm worried about this because there are HILs which don't specify whether `ReturnCode::Success` or `ReturnCode::SuccessWithValue` is returned, so there could be implementations return both in different circumstances.

phil-levis (commented):
Everything looks good except for the new `From`: I'm worried it will let ambiguities propagate and make capsules less careful about following the 2.0 restrictions on command return types.

kernel/src/driver.rs
hudson-ayers: Ah, I didn't realize that was a requirement -- just to clarify, for a given driver there can be multiple success types returned by a command, but for a given command_num only a single return type is possible? 

e.g. LedDriver::command() can return both `CommandResult::success()` and `CommandResult::success_u32()`, but only for different command numbers?

hudson-ayers: I still feel it is important to have something like this in the kernel, rather than every driver have to reimplement this same logic -- there should only be one correct way to convert a `ReturnCode` to a `CommandResult`, depending on the allowable type of `ReturnCode` for the given command_num. 

It seems odd to me that otherwise every single capsule should have to reimplement the logic in your `console` example, for instance. How about adding the following to `CommandResult`?

```rust
    /// Converts a `ReturnCode` to either `CommandResult::Success` or `CommandResult::Failure`.
    /// Used to ensure that only a single success type can be returned from a given
    /// `command_num`. If `ReturnCode::SuccessWithValue` is passed, the value will be dropped.
    pub fn from_return_code_no_success_with_value(rc: ReturnCode) -> Self {
        match rc {
            ReturnCode::SuccessWithValue { .. } => Self::success(),
            ReturnCode::SUCCESS => Self::success(),
            _ => Self::failure(ErrorCode::try_from(rc).unwrap()),
        }
    }

    /// Converts a `ReturnCode` to either `CommandResult::SuccessU32` or `CommandResult::Failure`.
    /// Used to ensure that only a single success type can be returned from a given
    /// `command_num`. If ReturnCode::Success is passed, this function will panic.
    pub fn from_return_code_success_only_with_value(rc: ReturnCode) -> Self {
        match rc {
            ReturnCode::SuccessWithValue { value } => Self::success_u32(value as u32),
            ReturnCode::SUCCESS => panic!("Cannot create CommandResult"),
            _ => Self::failure(ErrorCode::try_from(rc).unwrap()),
        }
    }
```

lschuermann: I'm wary of these wrappers as they abstract a vital part of encoding & communicating successes and errors to the userspace process, which is then no longer explicit but hidden behind a wrapper.

Really the only reason why we have this mechanical conversion to `CommandResult` at the end (which subjectively speaking I find somewhat misleading) is that we are currently in Stage 1 - so just mechanically porting capsules to the new API. In a Stage 2, we would be optimizing these capsules to make full use of the new ABI, so essentially returning a fully built `CommandResult` from each match arm. E.g. a "get current time" will have a completely different method of encoding it's return value to userspace than a "reset timer" (using different variants of the `CommandResult`).

It really doesn't make any sense to take two vastly independent operations (such as two different command numbers) and try to apply common operations on their return values, if that works it's going to be a coincidence.

hudson-ayers: I don't want to use these to apply common operations on the return values from different command numbers, I want to use these to convert values within command numbers (I realize the example in this PR does not reflect that, I suggested these after learning about the new restrictions). Take the following example, from the analog comparator:

```rust
        match command_num {
            1 => match ErrorCode::try_from(self.comparison(channel)) {
                Err(_) => CommandResult::success(),
                Ok(e) => CommandResult::failure(e),
            },

            2 => match ErrorCode::try_from(self.start_comparing(channel)) {
                Err(_) => CommandResult::success(),
                Ok(e) => CommandResult::failure(e),
            },

            3 => match ErrorCode::try_from(self.stop_comparing(channel)) {
                Err(_) => CommandResult::success(),
                Ok(e) => CommandResult::failure(e),
            },
    }
```

```rust
        match command_num {
            1 => ErrorCode::from_return_code_no_success_with_value(self.comparison(channel)),

            2 => ErrorCode::from_return_code_no_success_with_value(self.start_comparing(channel)),

            3 => ErrorCode::from_return_code_no_success_with_value(self.stop_comparing(channel)),
        }
```

These do the same thing - wrap a subcommand where it is expected that the `ReturnCode` will not be `SuccessWithValue`. But in my opinion the second make it much more explicit what is happening, and has a lot less repeated boilerplate.

We don't yet have an example of the appropriate way to handle a ReturnCode that should be `SuccessWithValue` but ends up being `Success`, so we can a) let each capsule author decide what to do (make up a value? panic?) or we can provide a helper function that makes the decision uniform. 

I understand that in stage 2 this will not be necessary, but I would not be surprised if many capsules are never udpated to "stage 2", or not updated for a very long time.

capsules/src/analog_comparator.rs
```diff
@@ -132,24 +132,25 @@ impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> LegacyDriver for Analo
...
         _app_id: AppId,
-    ) -> ReturnCode {
+    ) -> Result<Callback, (Callback, ErrorCode)> {
         match subscribe_num {
             // Subscribe to all interrupts
             0 => {
-                self.callback.set(callback);
-                ReturnCode::SUCCESS
+                self.callback.swap(Cell::from_mut(&mut callback));
+                Ok(callback)
```
bradjc, line 150: Wait why is this not

```rust
let old_callback = self.callback.replace(callback);
Ok(old_callback)
```

hudson-ayers: it should be, phil suggested the same in #2235 . I didn't realize the swapping itself was not important.

kernel/src/driver.rs
bradjc: Only 16 capsules use `SuccessWithValue`, and at least a few of those only exist in command=0 to return like the number of LEDs or something. I'm not sure it makes sense to add all of this boilerplate:

```
            1 => match ErrorCode::try_from(self.comparison(channel)) {
                Err(_) => CommandResult::success(),
                Ok(e) => CommandResult::failure(e),
            },
```
for the relatively rare case, particularly when that boilerplate is quite confusing (an Err returns success??). It also makes the capsule hard to read, since the actual operation (doing a comparison) is buried in a lot of irrelevant text that a capsule reader doesn't need to understand.

Doing:
```
1 => ErrorCode::from_return_code(self.comparison(channel)),
```
is much more straightforward. `from_return_code` can just always assume that SuccessWithValue is not used, and we can special case the conversion in the capsules where `SuccessWithValue` is used.

phil-levis: *shrug* SuccessWithValue was a mistake, we're now paying a cost for it. :(

I agree that your code is much more straightforward, but it's also incorrect. What if the `ReturnCode` is a success? `ErrorCode` does not encode success. It's an error code. This is a security-focused embedded kernel: we should not sacrifice correctness for code brevity.

In the particular case I was bringing up, the ambiguity of under-specified APIs could easily propagate to the syscall API and cause capsules to unknowingly violate the syscall API specification. Not good. The right approach is to have syscall drivers clean up possibly sloppy code, not cause that sloppiness to spread to userspace.

That being said. one reasonable way forward is to cut out SuccessWithValue. We can transition the capsules/traits that use it to use an alternative approach. I 100% agree that the current approach is verbose and am in favor of finding a cleaner way. But we need to do so without introducing errors.

hudson-ayers: I'm not convinced that my latest proposed solution (see a couple comments up) is worse in any aspect than the example in the updates console or temperature drivers. The code in each the console and temperature drivers, if presented with a SuccessWithValue, would silently transform the SuccessWithValue to Success -- this in no way ensures that whoever ported the code actually traced all the way back to verify that a SuccessWithValue in that location was impossible. I don't see why putting the same functionality behind a single function call, explicitly labeled "no_value", somehow makes this mistake more or less likely. But if we really wanna make sure the code is correct, perhaps we should panic if a SuccessWithValue is present -- this is much more likely to be noticed during testing!

All for getting rid of SuccessWithValue, maybe those are the HILs that should first be converted to Result.

phil-levis: > I'm not convinced that my latest proposed solution (see a couple comments up) is worse in any aspect than the example in the updates console or temperature drivers.

Your latest proposed solution seems reasonable. The big issue, of course, are the `panic!` calls. IMO, what's important is that the caller (e.g., the system call driver capsule) explicitly states what it expects, rather than doing an oblivious transformation. While I think the current behavior and invocation approach is correct, I agree its ergonomics are not great. One of the goals of this process is to figure out what the best ergonomics are and write them down, so we can translate the porting guide into a "writing syscall drivers guide".

hudson-ayers: I realize this PR isn't really the right place for this discussion, and that I am maybe repeating myself a little, so I'm sorry for dragging it on.

But IMO this:
```rust
                    let rcode = self.driver.read_temperature();
                    let eres = ErrorCode::try_from(rcode);
                    match eres {
                        Ok(ecode) => CommandResult::failure(ecode),
                        _ => CommandResult::success(),
                    }
```

does not make it very clear `read_temperature()` is not allowed to return `ReturnCode::SuccessWithValue`.

hudson-ayers: There are actually 0 uses of `SuccessWithValue` outside of userspace drivers (if you count IPC and memop as userspace drivers), so I think once all userspace drivers using it are transitioned to the new syscall API, removing `SuccessWithValue` will be really easy.

IMO, we should just make removing `SuccessWithValue` a blocking item for Tock 2.0, which lets everyone not worry about the possibility of its presence when updating each driver (and lets us provide an intuitive / easy to use `CommandResult::from(ReturnCode)`). Looking at the uses, I believe that completely removing `SuccessWithValue` will take only a couple hours once all 16 drivers using it are updated to the new syscall interface.

hudson-ayers: Ok, I fleshed out my proposal in #2252, which is hopefully a more appropriate place for continued discussion. Marking this conversation as resolved, as the relevant code is no longer present in this PR anyway.

phil-levis (approved):

hudson-ayers:
rebased.

bors r+

bradjc:
Now that #2252 is merged, can this be updated? Or is this still the latest version.

hudson-ayers:
Oh, yes, this can be updated, but I suppose I will need to submit another PR.

hudson-ayers:
See #2269

Issue #2246, opened 2020-12-08
Flash Controller Failed tests on OpenTitan 


KaisserbenDll:
The initial support for the OpenTitan Flash Controller, as documented here: https://docs.opentitan.org/hw/ip/flash_ctrl/doc/index.html is failing. 

I used the non volatile storage capsule provided by Tock (latest version in master). I tried to run  the libtock-c application in ` examples/tests/nonvolatile_storage`  and i expected the following output `All tests succeeded`. Instead, the test fails outputting: 
Inconsistency between data written and read at index 1. 
 
## Used Components: 

1) I used the master branch of the repository of lowrisc/opentitan (commit: 23b882050e91bd9ef392bde871d2cdd64db30646)
 to build the opentitan bitstream to try the non_volatile_storage capsule of Tock.  Note that the latest version of Opentitan has a changed  address of the Flash Controller. I receive the same failed test. 

2) I tested it also with the bitstream provided in the Makefile tock offers in the ci_setup_qemu_opentitan target (https://github.com/tock/tock/blob/master/Makefile#L517) 

@towoe 
## Log
```
I00438 boot_rom.c:61] Boot ROM initialisation has completed, jump into flash!
OpenTitan initialisation complete. Entering main loop
[Nonvolatile Storage] Test App
                              Have 32768 bytes of nonvolatile storage
                                                                     Test with size 256 ...
                Finished write! 14
                                        Finished read! 14
                                                                Inconsistency between data written and read at index 1
                                      Failed with code -1
```
## Meta
Hardware platform:  Earlgrey on Nexys Video

alistair23:
Thanks for raising this.

First, I have seen some issues using the OpenTitan Flash controller with the latest bitstreams. It seems like there are a few regression with the latest bitstreams.
That is mostly related to interrupts not occurring though. I don't think I have seen data loss with the latest master commit of Tock. I did push a fix to Tock recently that fixes some issues with the newer bitstreams.
From my understanding the failing case would need to be replicated with the lowRISC embedded software in order to have them investigated or fixed. From memory that isn't interrupt driven though, so replicating bugs will be difficult.
I am pretty sure I ran these tests when I wrote the flash controller, so I am guessing this is a hardware bug, in which case you might want to look at replicating it with the OpenTitan baremetal software.

It could also be a Tock bug though ;)
Can you provide more logs? It would be nice to see the actual data being written and read from flash. Is the data not being written to flash or are the reads not working correctly?

[...]

PR #2248, opened 2020-12-09
hil: Flash: Re-write the flash HIL

alistair23:
### Pull Request Overview

This PR changes the Flash HIL.

It does this for a few reasons:
 * Allow offset reads, some flash controllers allow offset reads in a page. In this case let's allow that in the HIL.
 * Allow writing to addresses instead of regions. This allows us to write smaller areas then an entire region
 * Clearly state that the erase has to happen separately from write, see https://github.com/tock/tock/issues/2246#issuecomment-741219570
 * Use `Result` and `ReturnCode` instead of a custom error type
 * Use const_generics to make the buffer handling simpler

### Testing Strategy

Pretty much none. This breaks all current flash users. Once the HIL is approved I can work on updating the users.

### TODO or Help Wanted

Comments and updating the flash HIL users.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.


hudson-ayers:
Going to point out that there was a lot of discussion on improvements for the flash HIL here: https://github.com/tock/tock/issues/1459 . Those changes are much further reaching than those in this PR, but maybe this is a reasonable intermediate change that is less controversial

alistair23:
Thanks @hudson-ayers

From a quick look this PR seems to address most of the issues besides the problem or writing/erasing a flash bank that we are also executing from, as in how do we know if it's ok to write/erase?

hudson-ayers:
> Thanks @hudson-ayers
> 
> From a quick look this PR seems to address most of the issues besides the problem or writing/erasing a flash bank that we are also executing from, as in how do we know if it's ok to write/erase?

This PR also doesn't open the can of worms of whether we should have a separate HIL for off-chip flash, and use a blocking interface for on-chip flash, but I think that's okay.

alistair23:
I pushed some updates that should address more concerns raised in https://github.com/tock/tock/issues/1459

krady21:
I am really excited to see these type of changes to the flash hil. Not depending on a fixed page size struct would be really helpful for chips that have variable page sizes (stm32f4xx for example). I remember that while working on the stm32f3 flash controller I really wished for a way to return a specific error through the Error struct (now ReturnCode). Would you see any benefits in adding a ReturnCode field in the form of EFLASH(&'static str) or something similar?

alistair23:
Do you mean the ability to return a string in the `ReturnCode`? I don't think that's a great idea. I think it would be better to just add more `ReturnCode`s if we are missing anything.

krady21:
Yes. That is mainly what I was thinking about. An example would be:
`
self.client.map(|client| {
    self.buffer.take().map(|buffer| {
        client.write_complete(buffer, ReturnCode::EFLASH("error name");
    });
});
`
It would allow chip maintainers to be more specific about the type of error they send to the client without adding a new ReturnCode for every chip specific flash error . For example, stm32f3 has two types of flash errors (write protection error and programming error) and the client tell them apart when it receives callbacks. I don't think this feature is super important, rather just something that I would've liked to see added and is somewhat related to the flash hil.

alistair23:
That is a big change for `ReturnCode` (which is currently just a usize). If you want you can propose that in a separate PR.

phil-levis:
I agree at a high level that we should revisit the flash HIL. It's pretty old, and our understandings of Rust and Tock have evolved a lot. 

Let's start with documentation first. We agreed 8 or so months ago that we would do complete overhauls of HILs rather than piecemeal. So let's redesign the flash HIL so it does everything we need it to do. A TRD would be the right place to start.

I'd like to hear from Amit why there's the `Page` abstraction in the old approach; that might have been an arbitrary decision or driven by some technical concerns.

No way we are embedded static strings in return codes. Our static strings are already a 40% overhead on code size.

Incorporating some of the thoughts in #1459 would be good. It's also worth noting that in most NAND chips the write unit and erase unit are *not* the same: you can write on a page basis but must erase on a block basis. We need to do a survey of not just NOR/NAND but also FRAM/MRAM/etc.

alistair23:
Phil has some good points. I have changed this so that write/erase pages can be different lengths and so we always write an entire page.

hudson-ayers:
> Incorporating some of the thoughts in #1459 would be good. It's also worth noting that in most NAND chips the write unit and erase unit are _not_ the same: you can write on a page basis but must erase on a block basis. We need to do a survey of not just NOR/NAND but also FRAM/MRAM/etc.

In my opinion, while it might be worth considering FRAM/MRAM if we change the nonvolatile_storage HIL, for the flash HIL I think it is reasonable to exclude them -- FRAM/MRAM have substantially different properties from Flash (FRAM, for example, is much faster, does not require erase-before-write, does not require wear-leveling, is byte addressable and programmable, etc.), and probably would deserve their own HIL.

phil-levis:
> In my opinion, while it might be worth considering FRAM/MRAM if we change the nonvolatile_storage HIL, for the flash HIL I think it is reasonable to exclude them -- FRAM/MRAM have substantially different properties from Flash (FRAM, for example, is much faster, does not require erase-before-write, does not require wear-leveling, is byte addressable and programmable, etc.), and probably would deserve their own HIL.

That seems reasonable. If after looking at how MCUs provide FRAM/MRAM or other NVRAM the conclusion is they are too different and should not be generalized here, that's a totally valid conclusion. But we should check. My experience with them 10 or so years ago supports your conclusion, but it was 10 years ago, and I don't know what MCU interfaces to them look like today.

alistair23:
Ping!

kernel/src/hil/flash.rs
```diff
@@ -4,141 +4,150 @@
...
+///      in an operation.
+/// `E`: Should be the minimum number of bytes that can be erased
+///      in an operation.
+pub trait Flash<const W: usize, const E: usize> {
+    /// Read data from flash into a buffer.
+    ///
+    /// This function will read data stored in flash at `address` and
+    /// `length` into the buffer `buf`.
+    ///
+    /// Not all implementations support this, in that case `read_page`
```
ppannuto, line 94: Was `read_page` removed intentionally? Unless I'm misreading this diff, it looks gone?

ppannuto, line 94: "Not all implementations support this" -- should we implement this a layered traits, similar to the `UartAdvanced` or `ScreenAdvanced`? The assumption I'm making is that all flash controllers support page-style accesses and some will support more granular access? 

That might also make it easier/more reasonable to parameterize the 'advanced' read trait on the read alignment (I would put money down that some chips allow byte access, while others only allow word access, but can't name examples off the top of my head).

kernel/src/hil/flash.rs
```diff
@@ -4,141 +4,150 @@
...
-    fn write_page(
+    /// Write data from a buffer to flash.
+    ///
+    /// This function will write the buffer `buf` to the `address` specified
+    /// in flash.
+    ///
+    /// `address` should be aligned to `W`.
+    ///
+    /// This function will not erase the page first. The user of this function
+    /// must ensure that a page is erased before writing.
```
ppannuto, line 114: Should we more precisely encode what it means to erase versus write in this HIL? It's probably meaningful to encode that writes allow 1->0 but not 0->1 explicitly

kernel/src/hil/flash.rs
```diff
@@ -4,141 +4,150 @@
...
+        buf: &'static mut [u8; W],
+    ) -> Result<(), (ReturnCode, &'static mut [u8; W])>;
 
-    /// Erase a page of flash by setting every byte to 0xFF.
-    fn erase_page(&self, page_number: usize) -> ReturnCode;
+    /// Erase a page of flash, setting every byte to 0xFF.
+    ///
+    /// On success returns nothing
+    /// On failure returns a `ReturnCode`.
+    fn erase_page(&self, page_number: usize) -> Result<(), ReturnCode>;
```
ppannuto, line 128: I think it's awkward to have some interfaces use addresses and others use page numbers. Especially once this HIL is generic across page size, I think it would be valuable for the page/address translation to in some ergonomic way be captured in this interface (`get_page_from_address`?).

The easier change might be to have `erase_page` (`erase_pages`?) take and address and length, like the other interfaces, and simply `EINVAL` addr/len pairs that aren't clean pages.

kernel/src/hil/flash.rs
alistair23: Yes, it has just been replaced with a `read()` function.

alistair23: The comment is out of date and I should remove it.

I was really torn about this. I was thinking of adding a `read_page()` function that capsules should use as a fallback. The problem is this then requires another const generic to specify the page size, which means we would then have three different sizes that need to be specified. As well as that it need to work well with the read callback. I was thinking instead that the driver could handle a read operation by splitting it as required.

For example is reading across pages aren't allowed the driver could read the page twice and then return the data.

This offloads the complexity to the code interacting with the hardware. The disadvantage is obviously that then we could duplicate the handling, but from what I have seen most hardware is pretty lenient with flash reads so I don't think this will be a huge problem.

kernel/src/hil/flash.rs
alistair23: Yeah, I see what you mean. But specifying an address and length to erase seems strange as well. Flash pretty much only ever erases and entire page (while generally allows more fine grained read/write). Allowing the caller to specify an address just seems confusing. People would then expect the erase to erase that address and length, even when it can't.

kernel/src/hil/flash.rs
alistair23: I think users of flash would know that, but I'll add something,

kernel/src/hil/flash.rs
ppannuto: I wonder if all of these problems are ameliorated by the layered HIL approach. The "basic flash" HIL is all page-based, it exposes only `page_read`, `page_write`, `page_erase` and takes page indicies. It's generic across page size.

The "advanced flash" HIL is address-based, it exposes `read` and `write` which take address/length, and is generic across the minimum alignment.

alistair23: So the "basic flash" HIL will be what we currently have and there will be a more advanced one (which is this PR).

How do capsules work with the two though? Won't capsules just end up using one and we either loose functionality or don't support all hardware?

phil-levis:
A couple of comments from the core call:
  1. Can we have better typing on addresses? E.g., have a way to pass in addresses to the HIL to produce a (checked) type that is then passed for read/write operations. 
  2. Can we somehow unify/simplify the distinction between addresses and pages? E.g., provide helpers that map addresses to page  numbers?
  3. The interface may want to support writes longer than W; otherwise large writes require a long series of small writes (with callbacks, etc.)
  4. We may want to separate out read/write/erase into separate traits, following the principle of least privilege. There can always be composite types so clients can do all three with a single reference.

alistair23:
Thanks for summarising that Phil.

I have changed the PR so it doesn't break all current flash users, but instead marks the current flash HIL as deprecated.

1. I'm not clear on what this would look like. I can see something like this:
```rust
/// A checked flash region
struct FlashReadRegion, {
    address: u64,
    length: u32,
}

fn get_read_region(&self, address: u64, length: u32) -> Result<FlashReadRegion, ReturnCode>;

fn read(&self, region: &FlashReadRegion, buf: &'static mut [u8], ) -> Result<(), (ReturnCode, &'static mut [u8])>;
```
But I don't see what that gets us `get_read_region()` would just do the same checks we would otherwise do in `read()`.

2. I have removed all mention of pages, hopefully that helps.
3. Yep, fixed.
4. I'm not sold on this idea. If everyone else is onboard I can do that, but for the HIL I don't think it's a great idea. All flash implementations will support all 3, this should be split at a higher level (where the app interfaces with the kernel for example)

phil-levis:
I'll start working on a TRD for this. It's a complicated problem.

alistair23:
Ping!

lschuermann:
> 1. I'm not clear on what this would look like. I can see something like this:
> 
> ```rust
> /// A checked flash region
> struct FlashReadRegion, {
>     address: u64,
>     length: u32,
> }
> 
> fn get_read_region(&self, address: u64, length: u32) -> Result<FlashReadRegion, ReturnCode>;
> 
> fn read(&self, region: &FlashReadRegion, buf: &'static mut [u8], ) -> Result<(), (ReturnCode, &'static mut [u8])>;
> ```

Yes, that's exactly what I thought of. You're right in that it does do the exact same checks as would be done otherwise, but it's not the sanity checks in the flash implementation itself I'm worried about: my issue is that a `write` and `erase` only works on a certain granularity, and issuing a `read` or `erase` operation by specifying an address won't always exactly match the address boundaries but might `erase` more than the consumer expected. Doing this in two stages gives the consumer the chance to see what the flash would be doing, prior to actually committing that operation, and can hence verify that this is indeed desired and/or depending on that result perform additional actions (such as re-write bytes outside of the range requested).

This arose from the fact that we had (have?) the issue of some methods taking an address, while others would take pages. Doing this two-stage commit would give us the chance to always talk to the flash in terms of an address first (which has the advantage of the client not requiring to know the precise flash limitations w.r.t. read / erase pages), resolve that to pages, check which bytes will _actually_ be affected by the operation, and then be able to commit it. Naturally, this would be a low-level flash HIL. A higher-level HIL might want to use this low-level HIL to then present the user an opaque byte-addressable block device.

alistair23:
Rebased this and added a `Page` per Leons suggestion

kernel/src/hil/flash.rs
```diff
@@ -142,3 +164,111 @@ pub trait Client<F: Flash> {
...
+    // should be used instead.
+    fn read(
+        &self,
+        region: &Page,
+        buf: &'static mut [u8],
+    ) -> Result<(), (ErrorCode, &'static mut [u8])>;
+
+    /// Create a `Page` from the provided `address` and `length`.
+    ///
+    /// The `Page->address` and Page->length` will be modified to match
```
dcz-self, line 200: There should be an elaboration on what modification exactly happens, e.g.:
```suggestion
    /// The resulting `Page` will cover the requested area, adjusting the provided values according to
```

kernel/src/hil/flash.rs
```diff
@@ -142,3 +164,111 @@ pub trait Client<F: Flash> {
...
+    /// the flash hardware requirements.
+    fn get_read_region(&self, address: u64, length: u32) -> Result<Page, ErrorCode>;
+
+    /// Write data from a buffer to flash.
+    ///
+    /// This function will write the buffer `buf` to the `address` specified
+    /// in flash.
+    ///
+    /// `address` must be aligned to `W`.
+    /// The length of `buf` must be aligned to `W`.
```
dcz-self, line 210: What happens if it's not?

kernel/src/hil/flash.rs
```diff
@@ -142,3 +164,111 @@ pub trait Client<F: Flash> {
...
+
+    /// Create a `Page` from the provided `address` and `length`.
+    ///
+    /// The `Page->address` and Page->length` will be modified to match
+    /// the flash hardware requirements.
+    fn get_read_region(&self, address: u64, length: u32) -> Result<Page, ErrorCode>;
+
+    /// Write data from a buffer to flash.
+    ///
+    /// This function will write the buffer `buf` to the `address` specified
```
dcz-self, line 206:
```suggestion
    /// This function will write the buffer `buf` to the `region.address` specified
```

kernel/src/hil/flash.rs
```diff
@@ -142,3 +164,111 @@ pub trait Client<F: Flash> {
...
+    /// in flash.
+    ///
+    /// `address` must be aligned to `W`.
+    /// The length of `buf` must be aligned to `W`.
+    ///
+    /// This function will not erase the page first. The user of this function
+    /// must ensure that a page is erased before writing.
+    /// Writes to flash can only turn a `1` to a `0`. To change a `0` to a `1`
+    /// the region must be erased.
```
dcz-self, line 215: I think this is slightly misleading, as there are sometimes additional restrictions depending on the hardware (e.g. can only do that twice, and I'm not sure if NAND supports that kind of stuff at all. Not to mention flash with FTL). IMO this should either send users to the manual for the specific hardware, or ask to erase before every overwrite.

kernel/src/hil/flash.rs
```diff
@@ -142,3 +164,111 @@ pub trait Client<F: Flash> {
...
+    /// Note that some hardware only allows a limited number of writes before
+    /// an erase. If that is the case the implementation MUST return an error
+    /// `ErrorCode::NOMEM` when this happens, even if the hardware silently
+    /// ignores the write.
+    ///
+    /// On success returns nothing
+    /// On failure returns a `ErrorCode` and the buffer passed in.
+    fn write(
+        &self,
+        region: &Page,
```
dcz-self, line 226: I'm not sure that Page is the right choice here, because it opens the possibility that the buffer is of different length than the page.

kernel/src/hil/flash.rs
```diff
@@ -142,3 +164,111 @@ pub trait Client<F: Flash> {
...
+    /// Create a `Page` from the provided `address` and `length`.
+    ///
+    /// The `Page->address` and Page->length` will be modified to match
+    /// the flash hardware requirements.
+    fn get_write_region(&self, address: u64, length: u32) -> Result<Page, ErrorCode>;
+
+    /// Erase a page/pages of flash, setting every byte to 0xFF.
+    ///
+    /// This will erase all pages starting from `address` for the `length`.
+    /// `address` and `length must allign with `E`.
```
dcz-self, line 239:
```suggestion
    /// `address` and `length must align with `E`.
```

dcz-self (commented):
Is it a good idea to rely on specific erase behaviour?

Without specifying the erase value, the Flash HIL could be useful for generic block storage, e.g. flash with a HIL, like eMMC storage, or, with minimal adjustments, SD cards.

kernel/src/hil/flash.rs
```diff
@@ -142,3 +164,111 @@ pub trait Client<F: Flash> {
...
+///
+/// `W`: Should be the minimum number of bytes that can be written
+///      in an operation.
+/// `E`: Should be the minimum number of bytes that can be erased
+///      in an operation.
+pub trait Flash<const W: usize, const E: usize> {
+    /// Read data from flash into a buffer.
+    ///
+    /// This function will read data stored in flash at `address` and
+    /// `length` into the buffer `buf`.
```
dcz-self, line 184: What if the buffer is smaller than the requested range? Will the buffer get filled multiple times?

What if the buffer is slightly larger than a page size? Does it make sense to require the driver to match the neighboring pages if the hardware only allows reading one page at a time?

Easy readout may be the point of this HIL, but I also think there's an intermediate level of complexity which separates write and erase size but doesn't care about comfortable reads like that.

dcz-self (commented):
More comments pending.

dcz-self:
I'm not sure if it's a good idea to make Page fields public. This leads to problems described above where the addresses are not really checked any more.

I think one way out of it is to create separate checked stuctures for the 3 use cases: erase, write, and read, each checked in the appropriate way, and convertible to each other, e.g.

```
struct EraseRegion<const E: u32> {
  pub start_block: u32, pub length_blocks: u32
}

struct ReadRegion {
  pub start_address: u64, pub length_bytes: u32,
}

impl ReadRegion {
  fn try_into_erase<E>(self) -> Option<EraseRegion<E>> {...}
  fn expand_into_erase<E>(self) -> EraseRegion<E> {...}
}
```

In addition, the indices of erase/write blocks could get a newtype each, just to avoid mixups.

I think I like it overall, and is an improvement over the old HIL.

dcz-self:
While writing an implementation, two things come to mind:

- there should be a way to get the total size of the device
- is this supposed to accommodate for non-uniform page sizes? If not, page checking logic becomes somewhat simpler, or at least reusable.

phil-levis:
@alistair23 Should we move forward on this or close it?

alistair23:
I think this is still worth moving forward on.

https://github.com/tock/tock/pull/2993 doesn't end up replacing this. Although if https://github.com/tock/tock/pull/2993 morphs into replacing this that is also fine

I do need some involvement before I can spend more time on it though. This PR received no feedback for 4 months before @dcz-self. If I rebase this and address comments will someone review it this time?

dcz-self:
When I wrote https://github.com/tock/tock/pull/2993 , I was actually building on top of this. I consider #2993 a  continuation of this work, even if it doesn't cover the same use cases yet. Therefore, although that might be controversial, I think the block_storage API should be generalized, rather than build fully-blown storage-specific HILs.

alistair23:
> I think the block_storage API should be generalized, rather than build fully-blown storage-specific HILs

Fine with me as well!

I should have bee a little clearer. I just meant that https://github.com/tock/tock/pull/2993 in it's current form doesn't replace this. So this PR is still relevant. If the changes go into the `block_device` that's fine with me

alevy:
@alistair23 is this something you think should be revived?

alevy:
@alistair23 OK, understood. I'll see where this stands for relevant folks w.r.t. #2993 and independently and either find a way to push this, or find a way to meaningfully "bookmark for later"

alevy:
Closing without prejudice.

alistair23:
So no updates for where this this stands with relevant folks? It would be nice to improve the flash HIL

dcz-self:
Sorry, I concluded that developing Tock takes too much time and I'm starting over with embedded-hal. For the foreseeable future, I won't be coming back to this

PR #2249, opened 2020-12-09
hil: Flash: Add blocking flash operations

alistair23:
### Pull Request Overview

This PR builds on top of: https://github.com/tock/tock/pull/2248

This should address most issues raised in: https://github.com/tock/tock/issues/1459

Most flash chips require a blocking write/erase to the flash bank that is being executed from. In this case the CPU will stop executing as it is also trying to read instructions from the same flash and is unable to do both at the same time. Some chips will helpfully just halt CPU execution, which makes this easy with a DMA engine. Others are more complex and will require code in RAM to perform the operation.

In this case Tock's async model falls apart as doing an async operation will either stall the CPU or in some cases can result in a broken state. For example, what happens if an interrupt occurs during the write process?

This PR adds blocking HIL flash operations. The idea is that a driver can return an error via the async calls and then the user can fallback to the blocking calls. Implementations can then have their own blocking implementation.

#### Documentation

The STM series ([Datasheet, section 2.5, page 11](https://www.st.com/resource/ja/programming_manual/cd00233952-stm32f205-215-stm32f207-217-flash-programming-manual-stmicroelectronics.pdf)) of chips will result in a bus stall on all flash reads while an erase or write is occuring. "This means that code or data fetches cannot be performed while a write/erase operation is ongoing."

The Apollo3 ([Datasheet, section 3.10.0, page 164](https://ambiq.com/wp-content/uploads/2020/10/Apollo3-Blue-MCU-Datasheet.pdf)) says: "When  erase  orprogramming operations are active, instructions cannot be fetched for execution from the Flash memory,so the on-chip SRAM would have to be used for code execution"

The nRF ([Datasheet, section 11.1, page 28](https://infocenter.nordicsemi.com/pdf/nRF52840_OPS_v0.5.pdf)) will "The CPU is halted if the CPU executescode from the flash while the NVMC is writing to the flash."

### Testing Strategy

None

### TODO or Help Wanted

The actual implementation.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/kernel/src/hil/flash.rs
+++ b/kernel/src/hil/flash.rs
@@ -4,78 +4,59 @@
 //! `page`. Here is an example of a page type and implementation of this trait:
 //!
 //! ```rust
+//! #![feature(min_const_generics)]
 //! use core::ops::{Index, IndexMut};
-//!
-//! use kernel::hil;
+//! use kernel::hil::flash;
 //! use kernel::ReturnCode;
 //!
 //! // Size in bytes
 //! const PAGE_SIZE: u32 = 1024;
 //!
-//! struct NewChipPage(pub [u8; PAGE_SIZE as usize]);
+//! struct NewChipStruct;
 //!
-//! impl Default for NewChipPage {
-//!     fn default() -> Self {
-//!         Self {
-//!             0: [0; PAGE_SIZE as usize],
-//!         }
-//!     }
-//! }
-//!
-//! impl NewChipPage {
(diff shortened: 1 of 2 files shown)


bradjc:
As Tock is async I don't think we want to add blocking APIs to core HILs.

alistair23:
That doesn't always apply to flash though, where the hardware requires the operation to be synchronous for some select operations (like OTA updates of flash)

PR #2252, opened 2020-12-10
Tock 2.0: Deprecate SuccessWithValue

hudson-ayers:
### Pull Request Overview

This PR removes all uses of `ReturnCode::SuccessWithValue` aside from those contained directly within an implementation of `LegacyDriver::command()`.

It also updates the porting guide to reflect that part of porting requires removing all uses of `SuccessWithValue` when moving to Driver.

This should make it extremely easy to remove `SuccessWithValue` entirely once all capsules have been ported.

Given that `SuccessWithValue` is no longer a consideration, this PR also adds a helper function to the kernel for converting a `ReturnCode` to a `CommandResult`.

There were 5 locations outside `LegacyDriver::command()` implementations that constructed `SuccessWithValue`:

1. UDP driver. I modified this code to ensure that SuccessWithValue is only constructed within `command()`, and not in any helper functions, by replacing its use with `Result`.

2. Hd44780 capsule returned `SuccessWithValue` from `allow()`, but allow() cannot return values in the case of success in the Tock 2.0 API. I modified it to just return Success, and made the previously returned value accessible via a new command. This will require userspace updates to work, but was a necessary change irrespective of deprecating `SuccessWithValue`.

3. IPC also returns `SuccessWithValue` from `allow()` -- I did not touch this, as I am told @alevy is working on porting IPC anyway, which will inevitably remove this use.

4. Memop liberally uses `SuccessWithValue`, but all of these will naturally be replaced as part of the switch to Tock 2.0

5. The analog comparator driver constructs `SuccessWithValue` in a helper function -- this has already been removed in #2245 , so I did not duplicate it here.

### Testing Strategy

This pull request has not been tested.

### TODO or Help Wanted

This pull request still needs thoughts on the `CommandResult::from(ReturnCode)` addition, and the changes to the porting guide.

### Documentation Updated

- [x] Updated the porting doc, but other reference to SuccessWithValue may still exist

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/driver.rs
+++ b/kernel/src/driver.rs
@@ -149,6 +149,19 @@ impl CommandResult {
     }
 }
 
+use core::convert::TryFrom;
+impl From<ReturnCode> for CommandResult {
+    fn from(rc: ReturnCode) -> Self {
+        match rc {
+            ReturnCode::SUCCESS => CommandResult::success(),
+            ReturnCode::SuccessWithValue { .. } => {
+                panic!("SuccessWithValue is deprecated");
+            } //TODO: delete before Tock 2.0
+            _ => CommandResult::failure(ErrorCode::try_from(rc).unwrap()),
+        }
+    }
+}
+
 #[allow(unused_variables)]
 pub trait Driver {
     fn subscribe(
--- a/capsules/src/hd44780.rs
+++ b/capsules/src/hd44780.rs
@@ -130,7 +130,7 @@
 
 use crate::driver;
 use core::cell::Cell;
-use kernel::common::cells::TakeCell;
+use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::hil::gpio;
 use kernel::hil::time::{self, Alarm};
 use kernel::{AppId, AppSlice, Callback, Grant, LegacyDriver, ReturnCode, SharedReadWrite};
@@ -256,6 +256,7 @@ pub struct HD44780<'a, A: Alarm<'a>> {
     lcd_after_command_status: Cell<LCDStatus>,
     lcd_after_delay_status: Cell<LCDStatus>,
     command_to_finish: Cell<u8>,
+    bytes_written: OptionalCell<usize>, //saves number of bytes written after an allow.
 }
 
 impl<'a, A: Alarm<'a>> HD44780<'a, A> {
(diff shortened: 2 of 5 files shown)


phil-levis:
I'll work on updating memop tomorrow.

alistair23 (approved):

kernel/src/driver.rs
```diff
@@ -149,6 +149,19 @@ impl CommandResult {
     }
 }
 
+use core::convert::TryFrom;
+impl From<ReturnCode> for CommandResult {
```
lschuermann, line 153: I see the point in adding a helper method for the common case, and especially for the first transition from the old to the new system call ABI. However, this is not the only way to encode a `ReturnCode` (and later `ErrorCode`) as a `CommandResult`. I'd like to emphasize this fact, and I believe that the generic `From` trait -- although common in the Rust ecosystem -- is poorly suited because of its semantic meaning elsewhere in the Rust ecosystem.

If we add a helper, I'd appreciate a more explicitly method on `CommandResult` (even `from_return_code`) to emphasize the fact that this is one particular way and helper to convert a `ReturnCode` to a `CommandResult` and by no means the only one (especially when other branches of a command would return other variants, given that each command is supposed to have only one success and failure variant).

hudson-ayers: One `SuccessWithValue` is removed, how else would a `ReturnCode` be converted to a `CommandResult`? Or do you just mean that there might be places in Tock like this:

```rust
fn command(...) -> CommandResult {
    match command_num {
        0 => {
            let retcode = capsule_helper_func();
            let mut some_u32 = 0;
            if retcode == ReturnCode::SUCCESS {
                some_u32 = other_helper_func();
            }
            match retcode {
                ReturnCode::SUCCESS => CommandResult::success_u32(some_u32),
                _ => CommandResult::failure(),
            }
        }
    }
}
```
where because you want to return additional information, you have to construct the `CommandResult` differently?

hudson-ayers: Or is this a concern about `?` automatically coercing when `From` is implemented? I don't think that can be an issue here either, because `CommandResult` is not actually a `Result<>` type?

phil-levis: bors r+

PR #2255, opened 2020-12-10
boards: BBC Micro:bit v2 (nRF52833)

alexandruradovici:
### Pull Request Overview

This pull request adds the BBC Micro:bit v2 board to Tock. The following work:
  - GPIO
  - Alarm
  - I2C
  - UART Console (via integrated debugger)
  - External Interrupts
  - Buzzer
  - Buttons
  - LED matrix (driver added)
  - BLE (advertise and scan)
  - LSM303AGR sensor
  - ADC
  - Temperature Sensor
  - App Flash

Additionally this PR:
  - adds LED matrix driver
  - adds support for the LSM303AGR sensor.
  - fixes a bug in LSM303DLHC that prevents the I2C bus from starting

Adding this board makes sense as it is very cheap (~15 USD), widely available and has an integrated debugger.

### Testing Strategy

This pull request was tested using a BBC Micro:bit v2 with the LSM303AGR sensor.

### TODO or Help Wanted

Some boards (undefined how to identify them) have an FXOS8700CQ sensor instead of LSM303AGR. If anyone has one of these boards, it would be great if we could test the FX support.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


brghena:
This board is incredibly exciting to me. It doesn't look like I can buy it anywhere until early next year though.

alexandruradovici:
In Europe, Pimoroni is selling it. I got one last week. I think they just launched it.

[1] https://shop.pimoroni.com/products/new-micro-bit-v2

bradjc:
This is a pretty interesting board...and one that could be a good "standard" tock board. One thought though, is I believe right now flashing the blink app on this board would result in the board not doing anything since the standard led syscall is not included. That might be confusing. Is there a way to still support the normal led commands from userspace while also intergrating the matrix?

alexandruradovici:
The LED Matrix exports the LED api, so right now the LEDs work as expected, numbered 0 to 24, left to right, top bottom.

LED 0 is set as the panic LED.

My plan is to add support for the screen driver for more graphics. A 5x5 1 bit screen.

alexandruradovici:
I have updated the nrf52 `ADC` driver in the following way:
  - set it to 12 bits instead of 14, as 14 bits requires oversampling
  - the ADC returns negative values as well, this actually making it an ADC on 12+1 bits. As most of the devices in user space expect values from 0 to 2^n-1 I have put a lower cap on 0. What actually happened before this change was that the users space got overflowed values. I was basically reading 3950 for GND.

Now here comes the question: how should this be handled? In 99% of the cases I am sure the user space expects a value of 0 or greater. I have two possible solutions so far:
 1. Adding a function to the HIL to report the ADC value at the reference point? In this way, the user space would know how to use the values. 
  2. Adding a function that returns a signed value?

bradjc:
The ADC still only has 12 bits of resolution though, right? The HIL specifies that ADC values should be left aligned in 16 bits, so as long as the reading isn't more than 16 bits it should be fine with the current HIL. The chip driver will just have to adjust the reading to be in the range 0 - 2^16-1.

alexandruradovici:
> The ADC still only has 12 bits of resolution though, right? The HIL specifies that ADC values should be left aligned in 16 bits, so as long as the reading isn't more than 16 bits it should be fine with the current HIL. The chip driver will just have to adjust the reading to be in the range 0 - 2^16-1.

Ok, that makes sense. Still I have a problem with the overflow, for GND I get almost every time negative values (even after calibration). Left shifting does not solve this problem.

alexandruradovici:
I reverted the nrf52 ADC and shifted the bits left. I did leave the 0 cap.

boards/microbit_v2/src/main.rs
```diff
@@ -0,0 +1,495 @@
...
+    };
+
+    let chip = static_init!(
+        nrf52833::chip::NRF52<Nrf52833DefaultPeripherals>,
+        nrf52833::chip::NRF52::new(nrf52833_peripherals)
+    );
+    CHIP = Some(chip);
+
+    // Need to disable the MPU because the bootloader seems to set it up.
+    chip.mpu().clear_mpu();
```
bradjc, line 446: Is this true on this board? It would be good to double check now (if you haven't) so we don't copy this line from the nano33 unnecessarily.

bradjc (approved):

boards/microbit_v2/src/main.rs
alexandruradovici: I don't know, let me check.

alexandruradovici: Deleted

bradjc (approved):

capsules/src/led_matrix.rs
```diff
@@ -0,0 +1,247 @@
...
+            buffer: TakeCell::new(buffer),
+            alarm,
+            col_activation: col_activation,
+            row_activation: row_activation,
+            current_row: Cell::new(0),
+            timing: (1000 / (refresh_rate * rows.len())) as u8,
+        }
+    }
+
+    pub fn init(&self) {
```
hudson-ayers, line 123: If I understand correctly, it seems that this function toggles all the LEDs in the matrix in some pattern at some refresh rate? Should this be named something like `patterned_blink()`? Is anything in this function actually necessary to initialize the LED matrix?

chips/nrf52833/src/interrupt_service.rs
```diff
@@ -0,0 +1,36 @@
+use crate::deferred_call_tasks::DeferredCallTask;
+use nrf52::chip::Nrf52DefaultPeripherals;
+
+/// This struct, when initialized, instantiates all peripheral drivers for the nrf52840.
+/// If a board wishes to use only a subset of these peripherals, this
+/// should not be used or imported, and a modified version should be
+/// constructed manually in main.rs.
+pub struct Nrf52833DefaultPeripherals<'a> {
+    pub nrf52: Nrf52DefaultPeripherals<'a>,
+    // put additional 52832 specific peripherals here
```
hudson-ayers, line 10:
```suggestion
    // put additional 52833 specific peripherals here
```

boards/microbit_v2/src/main.rs
```diff
@@ -0,0 +1,491 @@
+//! Tock kernel for the Micro:bit v2.
+//!
+//! It is based on nRF52833 SoC (Cortex M4 core with a BLE).
+
+#![no_std]
+// Disable this attribute when documenting, as a workaround for
+// https://github.com/rust-lang/rust/issues/62184.
+#![cfg_attr(not(doc), no_main)]
+#![feature(const_in_array_repeat_expressions)]
```
hudson-ayers, line 9: this should no longer be needed (see https://github.com/tock/tock/pull/2215)

hudson-ayers (dismissed):
Looks good, a few small suggestions

chips/nrf52833/src/interrupt_service.rs
alexandruradovici: Done

boards/microbit_v2/src/main.rs
alexandruradovici: Done

capsules/src/led_matrix.rs
alexandruradovici: No, this function inits the rows and columns to 0 and/or 1 (depending on the settings). It basically turns off the matrix at the beginning.

hudson-ayers: Ah, why does that require a timer?

alexandruradovici: I thought you were referring to only to the init function. The driver does require a timer as the matrix can "display" one line at a time. The call to `next_row` just starts the marix functioning.

hudson-ayers: Ah, I understand, thanks for explaining

hudson-ayers (approved):

bradjc (approved):

jaustin:
Hello from the Micro:bit Educational Foundation 👋. Sorry we're a bit late to the party on helping out with things like the FXOS-versions of the device, etc...

At the moment, there are no non-prototype micro:bit V2s with an FXOS8700CQ, and we don't currently plan to make any. We have an 'Editors, Devices and DAL' mailing list https://microbit.us14.list-manage.com/subscribe?u=e1c30f24b90ff3d70275cfff2&id=25403c7650 which we use to share about hardware revisions and updates, if we were to change that.

In general, we're super-keen for people to adopt micro:bit V2 (I mean, that's what we exist for :D) and the idea of it being a standard platform for tock is really exciting - we'd like to understand how we can help… Seems like you’ve already found tech.microbit.org and so perhaps the microbit-community slack, but help@microbit.org is staffed by excellent humans and we’re around on GitHub too.

alexandruradovici:
Hello Jonathan, thank you for reaching out to us. Its exciting to for us as well to hear that the Micro:bit foundation is interested in the Tock port.

I think that, together with Nano 33, the Micro:bit is one of the best supported ARM boards. We still lack a BLE stack, but that is a work in progress. What we could not clearly figure out so far is the Radio. My understanding is that it might be an nRF proprietary protocol? Which one of your communication channels (slack, help@microbit.org) would be the best one to ask about it?

Personally, I am trying to use the Micro:bit with Tock throughout my operating systems and IoT university courses this year, both of them focusing on secure programming. I suppose your age target is a little lower, but if this might be o interest to you, I can gladly share the materials and my experience so far. I think this might be a good opportunity to showcase a more professional usage of the Micro:bit and the advantages of Tock.

jaustin:
Hi @alexandruradovici - it'd be great it you want to put a reference to any of your research on https://tech.microbit.org/labs/ 

For the micro:bit Radio, in case you haven't already found these links, there's a top level documentation about it here
https://lancaster-university.github.io/microbit-docs/ubit/radio/

However, MakeCode has added some additional fileds on top of that, which are documented here https://ukbaz.github.io/howto/ubit_radio.html (not by us, but by other helpful community members!).

If your aim is to be compatible with the MakeCode radio, then I think Slack is best.

alexandruradovici:
@jaustin Do you have a suggestion as of where to put a reference? I suppose the Software category would be the place, I'm not sure whether to write it under Programming Languages or make a separate entry in the menu for Tock or maybe RTOSes ?

alexandruradovici:
Hi @jaustin, we just published a book about using Tock and micro:bit [Getting Started with Secure Embedded Systems](https://link.springer.com/book/10.1007/978-1-4842-7789-8#toc). It handles the development of embedded systems at a professional level. I was wondering if it would be of interest for the Micro:bit Educational Foundation and maybe post a mention about it?

PR #2258, opened 2020-12-11
Update touch and ft60x6 drivers to Tock 2.0

alexandruradovici:
### Pull Request Overview

This pull request updates the `touch` and `ft60x6` drivers to Tock 2.0 syscalls.

It adds an `is_set` function to `Callback` to be able to determine if a callback has been set.

### Testing Strategy

This pull request was tested using an STM32F412G Discovery board.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


kernel/src/callback.rs
```diff
@@ -216,6 +216,10 @@ impl Callback {
             ),
         }
     }
+
+    pub fn is_set(&self) -> bool {
```
phil-levis, line 220: I'd like to understand this better. Whether an application has a callback set or not shouldn't provide much information to the kernel: it could be that the callback does nothing. The kernel provides a more efficient way to do this with the Null Callback, not even bothering to invoke userspace. But, at a basic level, there is always a Callback, just some of them do nothing.

It looks like this method is needed for when you enable operations, you want to see if a non-Null Callback has been registered. Why? It might be removed right after you make one of those calls.

phil-levis (commented):
I think we don't want a Callback::is_set, but want to understand your use case better to be sure.

kernel/src/callback.rs
alexandruradovici: > I'd like to understand this better. Whether an application has a callback set or not shouldn't provide much information to the kernel: it could be that the callback does nothing. The kernel provides a more efficient way to do this with the Null Callback, not even bothering to invoke userspace. But, at a basic level, there is always a Callback, just some of them do nothing.
> 
> It looks like this method is needed for when you enable operations, you want to see if a non-Null Callback has been registered. Why? It might be removed right after you make one of those calls.

As discussed here (https://github.com/tock/tock/pull/1986#discussion_r452421535), the touch panel is enabled and disabled automatically based on whether there is or not an app that has subscribed. Using the new `Callback`, there does not seem to be a way to know if the app actually submitted or not a callback.

phil-levis: Ah, OK.

> As discussed here ([#1986 (comment)](https://github.com/tock/tock/pull/1986#discussion_r452421535)), the touch panel is enabled and disabled automatically based on whether there is or not an app that has subscribed. Using the new `Callback`, there does not seem to be a way to know if the app actually submitted or not a callback.

Ah, OK. In the 2.0 calling semantics, you want to not use this approach. A subscribe is just that: registering a callback. Early Tock code would overload subscribe and allow calls to also trigger operations, and experience showed this was bug prone: you want to separate concerns. E.g., I believe an early UART driver would start a transfer when you allowed a buffer. You're effectively overloading whether a process receives callbacks with whether the device is active. While this can make sense from a user-library standpoint, the kernel driver should provide explicit control. I.e., have enable/disable commands. Put another way, calls to allow or subscribe should not change the power or operational state of the kernel: they are merely for passing buffers or registering callbacks.

alexandruradovici: That was my initial idea, but the argument of not doing that was that one app should not be able to disable the touch pad, so then I changed to "automatic". I can add the enable and disable commands.

I am not exactly sure how to proceed here.

hudson-ayers: I think there should be `enable`/`disable` commands, but the driver should only actually disable the underlying hardware if there are no remaining apps that have called `enable` without calling `disable`. This would involve adding an `enable` bool to the `App` struct, then iterating through all of them before deciding whether to actually disable the underlying hardware.

It also probably makes sense to check if an app has "enabled" the screen before trying to deliver a callback to that app, as it is easy to imagine an app calling `disable` but not removing its registered callback... @phil-levis or should we assert that the only way to avoid receiving callbacks is by passing a null pointer?

phil-levis: You can definitely enable/disable callbacks. The point is that if you have callbacks enabled but have subscribed with the Null Callback they'll be generated but nothing is called. Just like a button in a UI still generates callbacks when it's clicked even if none are registered, it just does nothing.

The right way to do this is through virtualization, as Hudson indicates. Each application has a flag of whether it has them enabled. The main driver maintains a count. When the count > 0, they are enabled below. When the count == 0, the system can be disabled.

alexandruradovici: Solved

phil-levis (approved):

hudson-ayers (approved):
bors r+

PR #2259, opened 2020-12-11
Create a debug chip layout for adequate kernel ROM

kevtan:
### Pull Request Overview

Create a debug linker script so that `flash-debug` works more or less out of the box. Previously, running `make flash-debug` would give you this error: https://gist.github.com/kevtan/40239f93b11393c44aec8235c2d7c2ad.

### Testing

If you uncomment the debug linker script, you don't get the issue anymore.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/stm32f3discovery/README.md
+++ b/boards/stm32f3discovery/README.md
@@ -7,13 +7,20 @@ website](https://www.st.com/en/evaluation-tools/stm32f3discovery.html).
 ## Flashing the kernel
 
 The kernel can be programmed using OpenOCD. `cd` into `boards/stm32f3discovery`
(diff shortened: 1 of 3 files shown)


bradjc:
Are you using the debug builds?

bradjc:
I think this is a nice implementation of this change, but I'm not aware of anyone using the debug builds, and I think it would be helpful to either document when you would want to switch to debug, or at least document here what issue you were running into so we can point future developers in this direction if they have similar issues.

alevy:
@kevtan Closing for now, re-open if you get back to this.

PR #2262, opened 2020-12-11
nrf52: nvmc: remove global static

bradjc:
### Pull Request Overview

I think that with the new static init scheme for peripherals we shouldn't have these statics anymore.

I found this when using an old capsule that uses the nvmc and the state variable was somehow changing mysteriously! Turns out I had two copies of NVMC.

### Testing Strategy

Implementing bootloader on nrf52840dk over USB.

### TODO or Help Wanted

I don't think this will compile yet. Consider this a bug report.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/acd52832/src/main.rs
+++ b/boards/acd52832/src/main.rs
@@ -142,11 +142,11 @@ pub unsafe fn reset_handler() {
 
     // Make non-volatile memory writable and activate the reset button
     let uicr = nrf52832::uicr::Uicr::new();
-    nrf52832::nvmc::NVMC.erase_uicr();
-    nrf52832::nvmc::NVMC.configure_writeable();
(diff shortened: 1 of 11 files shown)


alistair23 (dismissed):

lschuermann (dismissed):

hudson-ayers:
I pushed a commit that removed the remaining users of the global NVMC, and also noticed that the global CLOCK variable somehow survived my initial changes as well, so I removed it (and uses of it) too.

hudson-ayers (approved):

hudson-ayers:
there was an implicit conflict because the added microbit board used the global CLOCK (yet another good reason to use less `static mut`)

hudson-ayers (approved):

PR #2266, opened 2020-12-14
Update UDP driver to Tock 2.0 system calls

hudson-ayers:
### Pull Request Overview

This pull request updates the UDP driver to use Tock 2.0 system calls. It also increases the stack size on the nrf52840dk. 

### Testing Strategy

This pull request was tested by flashing an updated version of the UDP apps onto an Imix and an nrf52840dk and exchanging UDP packets between the two. The updated apps work but I noticed that when `Ok()` is returned by subscribe from the kernel the apps treat it as an unknown return value. I assume this is because the error codes still need updating in `libtock-c`.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nordic/nrf52840dk/src/main.rs
+++ b/boards/nordic/nrf52840dk/src/main.rs
@@ -138,7 +138,7 @@ static mut CHIP: Option<&'static nrf52840::chip::NRF52<Nrf52840DefaultPeripheral
 /// Dummy buffer that causes the linker to reserve enough space for the stack.
 #[no_mangle]
 #[link_section = ".stack_buffer"]
-pub static mut STACK_MEMORY: [u8; 0x1000] = [0; 0x1000];
+pub static mut STACK_MEMORY: [u8; 0x2000] = [0; 0x2000];
 
 /// Supported drivers by the platform
 pub struct Platform {
@@ -194,7 +194,7 @@ impl kernel::Platform for Platform {
             capsules::nonvolatile_storage_driver::DRIVER_NUM => {
                 f(Some(Err(self.nonvolatile_storage)))
             }
-            capsules::net::udp::DRIVER_NUM => f(Some(Err(self.udp_driver))),
+            capsules::net::udp::DRIVER_NUM => f(Some(Ok(self.udp_driver))),
             capsules::tock2_test::DRIVER_NUM => f(Some(Ok(self.tock2_test_driver))),
             kernel::ipc::DRIVER_NUM => f(Some(Err(&self.ipc))),
             _ => f(None),
--- a/capsules/src/net/udp/driver.rs
+++ b/capsules/src/net/udp/driver.rs
@@ -17,6 +17,7 @@ use crate::net::udp::udp_send::{UDPSendClient, UDPSender};
 use crate::net::util::host_slice_to_u16;
 use core::cell::Cell;
 use core::convert::TryFrom;
+use core::mem::size_of;
 use core::{cmp, mem};
 use kernel::capabilities::UdpDriverCapability;
 use kernel::common::cells::MapCell;
@@ -48,7 +49,7 @@ impl UDPEndpoint {
     /// This function returns the new offset into the buffer wrapped in an
     /// SResult.
     pub fn encode(&self, buf: &mut [u8], offset: usize) -> SResult<usize> {
-        stream_len_cond!(buf, mem::size_of::<UDPEndpoint>() + offset);
+        stream_len_cond!(buf, size_of::<UDPEndpoint>() + offset);
 
         let mut off = offset;
         for i in 0..16 {
@@ -253,15 +254,15 @@ impl<'a> UDPDriver<'a> {
 
     #[inline]
     fn parse_ip_port_pair(&self, buf: &[u8]) -> Option<UDPEndpoint> {
-        if buf.len() != mem::size_of::<UDPEndpoint>() {
+        if buf.len() != size_of::<UDPEndpoint>() {
             debug!(
                 "[parse] len is {:?}, not {:?} as expected",
                 buf.len(),
-                mem::size_of::<UDPEndpoint>()
+                size_of::<UDPEndpoint>()
             );
             None
         } else {
-            let (a, p) = buf.split_at(mem::size_of::<IPAddr>());
+            let (a, p) = buf.split_at(size_of::<IPAddr>());
             let mut addr = IPAddr::new();
             addr.0.copy_from_slice(a);
 
@@ -280,9 +281,9 @@ impl<'a> Driver for UDPDriver<'a> {
     /// ### `allow_num`
     ///
     /// - `0`: Read buffer. Will contain the received payload.
-    /// - `2`: Config buffer. Used to contain miscellaneous data associated with
+    /// - `1`: Config buffer. Used to contain miscellaneous data associated with
     ///        some commands, namely source/destination addresses and ports.
-    /// - `3`: Rx config buffer. Used to contain source/destination addresses
+    /// - `2`: Rx config buffer. Used to contain source/destination addresses
     ///        and ports for receives (separate from `2` because receives may
     ///        be waiting for an incoming packet asynchronously).
     fn allow_readwrite(
@@ -298,15 +299,15 @@ impl<'a> Driver for UDPDriver<'a> {
                     mem::swap(&mut app.app_read, &mut slice);
                     Ok(())
                 }
-                2 => {
+                1 => {
                     mem::swap(&mut app.app_cfg, &mut slice);
                     Ok(())
                 }
-                3 => {
+                2 => {
                     mem::swap(&mut app.app_rx_cfg, &mut slice);
                     Ok(())
                 }
-                _ => Err(ErrorCode::INVAL),
+                _ => Err(ErrorCode::NOSUPPORT),
             })
             .map_err(ErrorCode::from);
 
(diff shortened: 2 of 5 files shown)


capsules/src/net/udp/driver.rs
```diff
@@ -261,32 +190,29 @@ impl<'a> UDPDriver<'a> {
...
-                                Err(mut buf) => {
-                                    buf.reset();
-                                    self.kernel_buffer.replace(buf);
-                                    ReturnCode::FAIL
-                                }
+            let result = app.app_write.map_or(ReturnCode::ENOMEM, |payload| {
+                self.kernel_buffer
+                    .take()
+                    .map_or(ReturnCode::ENOMEM, |mut kernel_buffer| {
+                        kernel_buffer[0..payload.len()].copy_from_slice(payload.as_ref());
```
lschuermann, line 197: What will happen here if the app allows a slice longer than the `kernel_buffer`? (It's limited by `max_tx_pyld_len`, is that checked to be in bounds of the `kernel_buffer` length?)

capsules/src/net/udp/driver.rs
```diff
@@ -466,154 +436,176 @@ impl<'a> LegacyDriver for UDPDriver<'a> {
...
+                            let iface_size = mem::size_of::<IPAddr>();
+                            for i in 0..n_ifaces_to_copy {
+                                cfg[i * iface_size..(i + 1) * iface_size]
+                                    .copy_from_slice(&self.interface_list[i].0);
+                            }
+                            // Returns total number of interfaces
+                            Ok(self.interface_list.len())
+                        })
+                    })
+                    .unwrap_or_else(|err| Err(err.into()));
```
lschuermann, line 471: Why do you have the `match` below? I think you could just return the `CommandResult` from the `.enter(` closure and `.unwrap_or_else` fallback?

lschuermann (commented):
Looks pretty good in general, although I'm by no means an expert for the UDP driver.

One think I noticed in the migration quite commonly is the fairly mechanical 

```rust
match res {
    Ok(v) => CommandResult::success_u32(v),
    Err(e) => CommandResult::failure(e),
}
```
(or similar) not only here but in other Tock 2.0 migrations as well. Is there any advantage to just creating the `CommandResult` inside the `Grant::enter` closure and having `.unwrap_or_else(|err| CommandResult::failure(e))` instead?

boards/nordic/nrf52840dk/src/main.rs
```diff
@@ -138,7 +138,7 @@ static mut CHIP: Option<&'static nrf52840::chip::NRF52<Nrf52840DefaultPeripheral
 /// Dummy buffer that causes the linker to reserve enough space for the stack.
 #[no_mangle]
 #[link_section = ".stack_buffer"]
-pub static mut STACK_MEMORY: [u8; 0x1000] = [0; 0x1000];
+pub static mut STACK_MEMORY: [u8; 0x2000] = [0; 0x2000];
```
phil-levis, line 141: That's a huge stack: 8kB. What needs it?

capsules/src/net/udp/driver.rs
phil-levis: This is a good point. I should update the porting guide on how to deal with slices/buffers of different sizes, especially when they might disappear from under you.

capsules/src/net/udp/driver.rs
```diff
@@ -364,34 +289,64 @@ impl<'a> LegacyDriver for UDPDriver<'a> {
...
         appid: AppId,
         allow_num: usize,
-        slice: Option<AppSlice<SharedReadWrite, u8>>,
-    ) -> ReturnCode {
-        match allow_num {
-            0 | 1 | 2 | 3 => self.do_with_app(appid, |app| {
-                let mut success = true;
-                match allow_num {
-                    0 => app.app_read = slice,
-                    1 => match slice {
```
phil-levis, line 374: I'd suggest compacting the identifier space.

capsules/src/net/udp/driver.rs
```diff
@@ -466,154 +436,176 @@ impl<'a> LegacyDriver for UDPDriver<'a> {
...
-                }
-                // Returns total number of interfaces
-                ReturnCode::SuccessWithValue {
-                    value: self.interface_list.len(),
+            1 => {
+                let res = self
+                    .apps
+                    .enter(appid, |app, _| {
+                        app.app_cfg.mut_map_or(Err(ErrorCode::INVAL), |cfg| {
+                            if cfg.len() != arg1 * mem::size_of::<IPAddr>() {
```
phil-levis, line 458: On a tangential note, it would be good to document what error codes are expected when. This could just as easily be an ESIZE.

capsules/src/net/udp/driver.rs
```diff
@@ -364,34 +289,64 @@ impl<'a> LegacyDriver for UDPDriver<'a> {
...
+        appid: AppId,
+        allow_num: usize,
+        mut slice: ReadOnlyAppSlice,
+    ) -> Result<ReadOnlyAppSlice, (ReadOnlyAppSlice, ErrorCode)> {
+        let res = match allow_num {
+            1 => self
+                .apps
+                .enter(appid, |app, _| {
+                    if slice.len() > self.max_tx_pyld_len {
+                        Err(ErrorCode::INVAL) // passed buffer too long
```
phil-levis, line 336: Not SIZE?

capsules/src/net/udp/driver.rs
```diff
@@ -466,154 +436,176 @@ impl<'a> LegacyDriver for UDPDriver<'a> {
...
+                        if app.pending_tx.is_some() {
+                            // Cannot support more than one pending tx per process.
+                            return Err(ErrorCode::BUSY);
+                        }
+                        if app.bound_port.is_none() {
+                            // Currently, apps need to bind to a port before they can send from said port
+                            return Err(ErrorCode::RESERVE);
                         }
+                        let next_tx = app.app_cfg.map_or(None, |cfg| {
+                            if cfg.len() != 2 * mem::size_of::<UDPEndpoint>() {
```
phil-levis, line 492: This seems like a common idiom (2 * mem::size_of::); can it be abstracted into a method?

phil-levis (commented):
Overall looks good -- there's just Leon's point about the slice lengths. I'll add some notes to the porting guide.

boards/nordic/nrf52840dk/src/main.rs
hudson-ayers: I got a stack overflow when running the UDP example receive application with a 4kB stack -- I think somewhere within the networking code an MTU-sized buffer is being passed on the stack, as I remember running into this issue on Imix as well. I just haven't had time to look into it.

Imix already uses an 8kB stack today, so I changed this to match.

capsules/src/net/udp/driver.rs
hudson-ayers: I was just following the porting guide example, but you're right that's much better. I'll make that change throughout here and submit a PR to update the porting guide to do the same.

lschuermann: I'll happily create the porting guide PR if you like.

capsules/src/net/udp/driver.rs
hudson-ayers: This will panic if the app allows a slice longer than the `kernel_buffer`. The UDP component checks that the `kernel_buf` is longer than `max_tx_pyld_len`, so as long as the UDP capsule is constructed via the component, it should be impossible for an app to do so given the checks in `allow()`.

Actually, after checking, the UDP component mistakenly set the `kernel_buf` length to 8 bytes smaller `max_tx_pyld_len`, so I have fixed that as well.

I am confused by "they might disappear from under you" comment -- isn't this covered by the `map_or()`?

capsules/src/net/udp/driver.rs
hudson-ayers: That would be great!

capsules/src/net/udp/driver.rs
hudson-ayers: we use both `2*mem::size_of::` and `mem::size_of`... we could abstract it into a method like `const fn size_of()` on `UDPEndpoint`, but that feels not worth it to me.

I did add a `use core::mem::size_of` so that each `2*mem::size_of<T>()` is now just `2*size_of<T>()`

capsules/src/net/udp/driver.rs
hudson-ayers: This is fixed in this PR now.

capsules/src/net/udp/driver.rs
hudson-ayers: done.

capsules/src/net/udp/driver.rs
hudson-ayers: I added a couple more error code docs to what was already there -- completely fleshing them out requires looking at the possible return codes that can be created all the way down the networking stack, but I think that is outside the scope of this PR.

capsules/src/net/udp/driver.rs
hudson-ayers: Changed.

hudson-ayers:
I believe all comments have been addressed.

capsules/src/net/udp/driver.rs
phil-levis: Sorry, I was writing imprecisely.

I'm generally worried about capsule code assuming things about the length of application slices: they can change in length between invocations. In this particular case, what happens if the application slice is longer than the kernel buffer? An application could provide an arbitrary length AppSlice. I see this is checked in allow(), but this checks against `max_tx_payld_len`, not the actual length of `kernel_buf`. So someone accidentally passes the wrong value when creating the driver, then in some weird cases suddenly the kernel panics when an application sends a long buffer.

I think this code would be more robust/safter if it explicitly looked at the length of `kernel_buffer`.

phil-levis (approved):

capsules/src/net/udp/driver.rs
hudson-ayers: I added an extra check.

lschuermann (approved):
Without knowing too much about the UDP driver, this looks good, especially in terms of the 2.0 Driver interface.

PR #2273, opened 2020-12-15
kernel: hil: Add a KV Store HIL

alistair23:
### Pull Request Overview

@bradjc mentioned he would be happier with https://github.com/tock/tock/pull/2156 and https://github.com/tock/tock/pull/2205 if we had a HIL to use. So this is that HIL.

If this is accepted then I can update the TicKV capsule to use this.

### Testing Strategy

None

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/kernel/src/hil/kv_system.rs
+++ b/kernel/src/hil/kv_system.rs
@@ -1,0 +1,203 @@
+//! Low level interface for Key-Value (KV) Stores
+//!
+//! The KV store implementation in Tock has three levels, described below.
+//!
+//! 1 - Hardware Level:
+//! This level is the interface that writes a buffer to the hardware. This will
+//! generally be writing to flash, although in theory it would be possible to
+//! write to other mediums.
+//!
+//! An example of the HIL used here is the Tock Flash HIL.
+//!
+//! 2 - KV System Level:
+//! This level can be thought of like a file system. It is responsible for
+//! taking save/load operations and generating a buffer to pass to level 1
+//! This level is also in charge of generating hashes and checksums.
+//!
+//! This level allows generating a key hash but otherwise operates on
+//! hashed keys. This level is not responsible for permission checks.
+//!
+//! This file describes the HIL for this level.
+//!
+//! 3 - KV Store:
+//! This is a user friendly high level HIL. This HIL is used inside the kernel
+//! and exposed to applications to allow KV operations. The API from this level
+//! should be high level, for example set/get/delete on unhashed keys.
+//! This level is in charge of enforcing permissions. It is expected that
+//! this level will combine the user data with a header and pass that to the
+//! level 2 system HIL.
+//!
+//! This level is also in charge of generating the key hash by calling into
+//! level 2.
+//!
+//! There is currently no implementation of this HIL in Tock.
+//!
+//! The expected setup inside Tock will look like this:
+//! +-----------------------+
+//! |                       |
+//! |  Capsule using K-V    |
+//! |                       |
+//! +-----------------------+
+//!
+//!    hil::kv_store (not written yet)
+//!
+//! +-----------------------+
+//! |                       |
+//! |  K-V in Tock          |
+//! |                       |
+//! +-----------------------+
+//!
+//!    hil::kv_system (this PR)
+//!
+//! +-----------------------+
+//! |                       |
+//! |  K-V library          |
+//! |                       |
+//! +-----------------------+
+//!
+//!    hil::flash
+
+use crate::returncode::ReturnCode;
+
+/// The type of keys, this should define the output size of the digest
+/// operations.
+pub trait KeyType: Eq + Copy + Clone + Sized + AsRef<[u8]> + AsMut<[u8]> {}
+
+/// Implement this trait and use `set_client()` in order to receive callbacks.
+pub trait Client<'a, K: KeyType> {
+    /// This callback is called when the append_key operation completes
+    ///
+    /// `result`: Nothing on success, 'ReturnCode' on error
+    /// `unhashed_key`: The unhashed_key buffer
+    /// `key_buf`: The key_buf buffer
+    fn generate_key_complete(
+        &'a self,
+        result: Result<(), ReturnCode>,
+        unhashed_key: &'static [u8],
+        key_buf: &'static K,
+    );
+
+    /// This callback is called when the append_key operation completes
+    ///
+    /// `result`: Nothing on success, 'ReturnCode' on error
+    /// `key`: The key buffer
+    /// `value`: The value buffer
+    fn append_key_complete(
+        &'a self,
+        result: Result<(), ReturnCode>,
+        key: &'static mut K,
+        value: &'static mut [u8],
+    );
+
+    /// This callback is called when the get_value operation completes
+    ///
+    /// `result`: Nothing on success, 'ReturnCode' on error
+    /// `key`: The key buffer
+    /// `ret_buf`: The ret_buf buffer
+    fn get_value_complete(
+        &'a self,
+        result: Result<(), ReturnCode>,
+        key: &'static mut K,
+        ret_buf: &'static mut [u8],
+    );
+
+    /// This callback is called when the invalidate_key operation completes
+    ///
+    /// `result`: Nothing on success, 'ReturnCode' on error
+    /// `key`: The key buffer
+    fn invalidate_key_complete(&'a self, result: Result<(), ReturnCode>, key: &'static mut K);
+
+    /// This callback is called when the garbage_collect operation completes
+    ///
+    /// `result`: Nothing on success, 'ReturnCode' on error
+    fn garbage_collect_complete(&'a self, result: Result<(), ReturnCode>, key: &'static mut K);
+}
+
+pub trait KVSystem {
+    /// The type of the hashed key. For example '[u8; 64]'.
+    type K: KeyType;
+
+    /// Set the client
+    fn set_client(&self, client: &dyn Client<Self::K>);
+
+    /// Generate key
+    ///
+    /// `unhashed_key`: A unhashed key that should be hashed.
+    /// `key_buf`: A buffer to store the hashed key output.
+    ///
+    /// On success returns nothing.
+    /// On error the unhashed_key, key_buf and `ReturnCode` will be returned.
+    fn generate_key(
+        &self,
+        unhashed_key: &'static [u8],
+        key_buf: &'static Self::K,
+    ) -> Result<(), (&'static [u8], &'static Self::K, ReturnCode)>;
+
+    /// Appends the key/value pair.
+    ///
+    /// `key`: A hashed key. This key will be used in future to retrieve
+    ///        or remove the `value`.
+    /// `value`: A buffer containing the data to be stored to flash.
+    ///
+    /// On success nothing will be returned.
+    /// On error the key, value and a `ReturnCode` will be returned.
+    ///
+    /// The possible `ReturnCode`s are:
+    ///    `EBUSY`: An operation is already in progress
+    ///    `EINVAL`: An invalid parameter was passed
+    ///    `ENODEVICE`: No KV store was setup
+    ///    `ENOSUPPORT`: The key could not be added due to a collision.
+    ///    `ENOMEM`: The key could not be added due to no more space.
+    fn append_key(
+        &self,
+        key: &'static Self::K,
+        value: &'static [u8],
+    ) -> Result<(), (&'static Self::K, &'static [u8], ReturnCode)>;
+
+    /// Retrieves the value from a specified key.
+    ///
+    /// `key`: A hashed key. This key will be used to retrieve the `value`.
+    /// `ret_buf`: A buffer to store the value to.
+    ///
+    /// On success nothing will be returned.
+    /// On error the key, ret_buf and a `ReturnCode` will be returned.
+    ///
+    /// The possible `ReturnCode`s are:
+    ///    `EBUSY`: An operation is already in progress
+    ///    `EINVAL`: An invalid parameter was passed
+    ///    `ENODEVICE`: No KV store was setup
+    ///    `ENOSUPPORT`: The key could not be found.
+    fn get_value(
+        &self,
+        key: &'static Self::K,
+        ret_buf: &'static mut [u8],
+    ) -> Result<(), (&'static Self::K, &'static [u8], ReturnCode)>;
+
+    /// Invalidates the key in flash storage
+    ///
+    /// `key`: A hashed key. This key will be used to remove the `value`.
+    ///
+    /// On success nothing will be returned.
+    /// On error the key and a `ReturnCode` will be returned.
+    ///
+    /// The possible `ReturnCode`s are:
+    ///    `EBUSY`: An operation is already in progress
+    ///    `EINVAL`: An invalid parameter was passed
+    ///    `ENODEVICE`: No KV store was setup
+    ///    `ENOSUPPORT`: The key could not be found.
+    fn invalidate_key(&self, key: &'static Self::K) -> Result<(), (&'static Self::K, ReturnCode)>;
+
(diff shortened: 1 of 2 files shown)


kernel/src/hil/kv_store.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// `key`: A unhashed key. This will be hashed internally.
+    ///
+    /// On success nothing will be returned.
+    /// On error the key and a `ReturnCode` will be returned.
+    ///
+    /// If a power loss occurs before success is returned the data is
+    /// assumed to be lost.
+    fn invalidate_key(&self, key: &'static [u8]) -> Result<(), (&'static [u8], ReturnCode)>;
+
+    /// Perform a garbage collection on the KV Store
```
lschuermann, line 76: Maybe it's worth adding a comment that on a key-value store which does not require garbage collection, this can be implemented as a `nop` and returning `Ok(0)`?

kernel/src/hil/kv_store.rs
```diff
@@ -0,0 +1,81 @@
...
+        value: Option<&'static mut [u8]>,
+        buf: Option<&'static mut [u8]>,
+    );
+}
+
+pub trait KVStore {
+    /// Appends the key/value pair.
+    ///
+    /// `key`: A unhashed key. This will be hashed internally. This key
+    ///        will be used in future to retrieve or remove the `value`.
```
lschuermann, line 38: This makes some assumptions about hashing the key, whereas I can imagine this would normally be internal to the KV store? If there is a collision, an appropriate ReturnCode variant will be returned.

kernel/src/hil/kv_store.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// `value`: A buffer containing the data to be stored to flash.
+    ///
+    /// On success nothing will be returned.
+    /// On error the key, value and a `ReturnCode` will be returned.
```
lschuermann, line 42: You should fix some `ReturnCode`s for known error cases, at least for the key collision case (as I imagine that's a quite common and prominent error here). That also holds for `ReturnCode`s which are returned as part of the asynchronous callback.

kernel/src/hil/kv_store.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// Retrieves the value from a specified key.
+    ///
+    /// `key`: A unhashed key. This will be hashed internally.
+    /// `buf`: A buffer to store the value to.
+    ///
+    /// On success nothing will be returned.
+    /// On error the key, buf and a `ReturnCode` will be returned.
+    ///
+    /// If a power loss occurs before success is returned the data is
+    /// assumed to be lost.
```
lschuermann, line 58: I don't quite follow. Why should a `get_key(` operation cause any data loss? Are you referring to the data in RAM?

kernel/src/hil/kv_store.rs
```diff
@@ -0,0 +1,81 @@
...
+    ///
+    /// 'operation': The operation that completed
+    /// 'result': Nothing on success, 'ReturnCode' on error
+    /// 'key': A key buffer is one exists
+    /// 'value': A value buffer is one exists
+    /// 'buf': A buf buffer is one exists
+    fn operation_complete(
+        &'a self,
+        operation: Operation,
+        result: Result<(), ReturnCode>,
```
lschuermann, line 27: This might look unnecessary now, but since in Tock 2.0 `ErrorCode`s we don't have a `SUCCESS` variant, this is exactly the way to go. :+1:

lschuermann (dismissed):
A HIL for a KV store sounds like a great idea! Added some comments.

lschuermann:
It'd be great if we could make the key an associated type instead of a `'static [u8]` slice. The associated types would require some trait implementations that make sense from a KV-store and user perspective. That could allow for much simpler KV store implementations, using a fixed size integer as a key without internal hashing. Going along with that idea, it might be helpful to add a `next_free_key` method, for users which are not interested in having semantic meaning in their keys or want to work around a collision.

kernel/src/hil/kv_store.rs
alistair23: There are lots of different possible KV store specific errors that can happen. I don't really want to list them all out in `ReturnCode` as `ReturnCode` is generic.

kernel/src/hil/kv_store.rs
alistair23: I have changes this to a hashed key.

The reason to pass the KV store an unhashed key is to allow the store to internally hash it. Now the capsule will have to perform the hash. This has the disadvantage that the implementation doesn't control the hash, but hashing internally in the KV store ends up being tricky to get right.

kernel/src/hil/kv_store.rs
```diff
@@ -0,0 +1,84 @@
...
+        key: Option<&'static mut K>,
+        value: Option<&'static mut [u8]>,
+        buf: Option<&'static mut [u8]>,
+    );
+}
+
+pub trait KVStore<K: KeyType> {
```
lschuermann, line 40: Using generics is probably fine, but I think that it's more idiomatic in the Rust universe to use an associated type in such a case. It seems close to the examples from the [_Rust by example_ section](https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html).

As per this [StackOverflow answer](https://stackoverflow.com/a/32065644) it also makes sense to ensure that there can be only one implementation of `KVStore` on a given type, not multiple (which an associated type would assure). I quite like that you don't have to write out the full type specification every time (as it is unambiguous then).

kernel/src/hil/kv_store.rs
lschuermann: I see your point, but an extensive listing is common in other HILs, like for instance UART:
https://github.com/tock/tock/blob/9f0da22661dbfb04319edad93ac37f50cc4013b1/kernel/src/hil/uart.rs#L82-L95

I think it'd would still be worth covering the common cases. If something bad happens in the KV Store on-flash architecture, memory corruption, device failure or other unforeseen edge cases, it'd probably still be fine to return an error not listed there, or simply `EFAIL`. A HIL device the contract between implementer and consumer, and I'd argue error handling should be a part of that?

kernel/src/hil/kv_store.rs
alistair23: Changed.

kernel/src/hil/kv_store.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// The type of the hashed key. For example '[u8; 64]'.
+    type KeyType;
+
```
lschuermann, line 41: I guess the `set_client` method could be part of the HIL, as done with the other HILs (such as UART).

kernel/src/hil/kv_store.rs
hudson-ayers: Now this seems to use both generics and an associated type -- I think you want to have this be just 
```rust
pub trait KVStore {
    type K: KeyType;
    //...
}
```

lschuermann: Yes, it shouldn't be mixed in one trait. Although I think it's good that the client still uses generics. A consumer implementing it might want to access multiple key-value stores with different `KeyType`s, which will be possible with generics.

alistair23: Whoops, fixed.

kernel/src/hil/kv_store.rs
alistair23: Added

kernel/src/hil/kv_store.rs
```diff
@@ -0,0 +1,112 @@
+//! Interface for KV Stores
```
bradjc, line 1:
```suggestion
//! Interface for Key-Value Stores
```

kernel/src/hil/kv_store.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+pub trait KVStore {
+    /// The type of the hashed key. For example '[u8; 64]'.
+    type K: KeyType;
+
+    /// Set the client
+    fn set_client(&self, client: & dyn Client<Self::K>);
+
+    /// Appends the key/value pair.
```
bradjc, line 45: What happens if the key is already used?

kernel/src/hil/kv_store.rs
```diff
@@ -0,0 +1,112 @@
...
+
+/// Implement this trait and use `set_client()` in order to receive callbacks.
+pub trait Client<'a, K: KeyType> {
+    /// This callback is called when any operation is complete.
+    ///
+    /// `operation`: The operation that completed
+    /// `result`: Nothing on success, 'ReturnCode' on error
+    /// `key`: A key buffer is one exists
+    /// `value`: A value buffer is one exists
+    /// `buf`: A buf buffer is one exists
```
bradjc, line 27:
```suggestion
    /// `key`: A key buffer if one exists
    /// `value`: A value buffer if one exists
    /// `buf`: A buf buffer if one exists
```

kernel/src/hil/kv_store.rs
```diff
@@ -0,0 +1,112 @@
...
+    ///
+    /// On success nothing will be returned.
+    /// On error the key, buf and a `ReturnCode` will be returned.
+    ///
+    /// The possible `ReturnCode`s are:
+    ///    `EBUSY`: An operation is already in progress
+    ///    `EINVAL`: An invalid parameter was passed
+    ///    `ENODEVICE`: No KV store was setup
+    ///    `ENOSUPPORT`: The key could not be found.
+    fn get_key(
```
bradjc, line 79:
```suggestion
    fn get_value(
```
?

kernel/src/hil/kv_store.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// `key`: A key buffer is one exists
+    /// `value`: A value buffer is one exists
+    /// `buf`: A buf buffer is one exists
+    fn operation_complete(
+        &'a self,
+        operation: Operation,
+        result: Result<(), ReturnCode>,
+        key: Option<&'static mut K>,
+        value: Option<&'static mut [u8]>,
+        buf: Option<&'static mut [u8]>,
```
bradjc, line 34: What is `buf`?

bradjc (commented):
A couple questions, but overall looks good to me.

kernel/src/hil/kv_store.rs
alistair23: That is up to the implementation, maybe some implementations allow that.

kernel/src/hil/kv_store.rs
alistair23: It is the buffer passed into `get_value()`, I'll fix this.

kernel/src/hil/kv_store.rs
```diff
@@ -0,0 +1,112 @@
...
+/// Implement this trait and use `set_client()` in order to receive callbacks.
+pub trait Client<'a, K: KeyType> {
+    /// This callback is called when any operation is complete.
+    ///
+    /// `operation`: The operation that completed
+    /// `result`: Nothing on success, 'ReturnCode' on error
+    /// `key`: A key buffer if one exists
+    /// `value`: A value buffer if one exists
+    /// `ret_buf`: A ret_buf buffer if one exists
+    fn operation_complete(
```
lschuermann, line 28: I'd think most other HILs would use separate methods in the client for each type of callback (each kind of operation that is completed). Take for instance the nonvolatile storage driver:
https://github.com/tock/tock/blob/2762b10452a883a1a5d5a28c0d0831f6aee2c69d/kernel/src/hil/nonvolatile_storage.rs#L21-L32

That would have an advantage in that you can more precisely specify what values are returned for a specific operation (instead of wrapping them all in an `Option`) and thus reduces the potential for runtime errors and the necessity for checks.

I think this might be fine too. Is there a specific reason you designed it this way? Maybe because of code bloat, since most HILs probably don't have that many different callbacks?

bradjc: I had this same question. The `Option<>` seems like extra work for the user that separate functions would make simpler.

kernel/src/hil/kv_store.rs
bradjc: But what does the user expect to happen? Do they get the last value appended? All of the values appended? Is this how values are updated?

alistair23: They would either get an error or a success value at this call.

When they call `get_value()` I guess that would depend on the implementation.

kernel/src/hil/kv_store.rs
alistair23: I had combined them to allow other operations such as init to have callabacks. I have split them out though.

kernel/src/hil/kv_store.rs
bradjc: I'm not sure I understand what the purpose of this interface is if how it works is left to a particular implementation. If how a key's value is updated isn't specified then it will be hard for any upper layers to use this, won't it?

alistair23: I don't think so.

An upper layer attempts to add a key. Implementations can decide what they do with that, maybe they support collisions, maybe they don't and return either Success or Error.

Just a note though, that you can't really support collisions when passing in the hashed key, so you pretty much will always have an error. I'll add a function to generate keys to allow custom key types.

bradjc: How does a client update a key with a new value?

bradjc:
Thinking about this more, this interface is I think fairly low level. Wouldn't we be better off having the HIL interface be easier to use, and forcing the implementation to deal with the complexity?

The interface could be:
- `set_value(key: &'static [u8], value: &'static [u8])`
- `get_value(key: &'static [u8])`

Converting the key to a hashed key could be done internally, as could any potential retries necessary to update a value.

Otherwise, all users have to implement the boilerplate of key conversion, and have to reason about why there are two keys and what they mean.

Although, maybe it would be useful to allow users to operate with the hashed key if they already have it? I could see this HIL exposing two sets of functions, the higher level easy-to-use versions, and the lower level hashed-key versions.

alistair23:
I'm not clear on what you mean.

The goal of this interface is to address concerns with the TicKV implementation. This HIL is for a KVStore that interfaces with the filesystem/flash to implement. The boiler plate will implemented in a capsule so that the design is more Tock friendly.

If we made a simple get/set/delete interface we would be back to where we started with an implementation like this: https://github.com/tock/tock/pull/2205

We could add a second HIL that is more user friendly, but that seems like a issue.

bradjc:
I think this is a rough sketch of the proposed KV stack in Tock:

```
+-----------------------+
|                       |
|  Capsule using K-V    |
|                       |
+-----------------------+

   hil::kv::upper (not written yet)

+-----------------------+
|                       |
|  K-V in Tock          |
|                       |
+-----------------------+

   hil::kv::lower (this PR)

+-----------------------+
|                       |
|  K-V library          |
|                       |
+-----------------------+

   hil::flash
```

Which looks good to me. I think for this PR the main thing is we need some sort of names for the two K-V interfaces.

alistair23:
Just pushed an update, I used your nice layout @bradjc

kernel/src/hil/kv_system.rs
```diff
@@ -0,0 +1,203 @@
...
+    /// `key`: The key buffer
+    fn invalidate_key_complete(&'a self, result: Result<(), ReturnCode>, key: &'static mut K);
+
+    /// This callback is called when the garbage_collect operation completes
+    ///
+    /// `result`: Nothing on success, 'ReturnCode' on error
+    fn garbage_collect_complete(&'a self, result: Result<(), ReturnCode>, key: &'static mut K);
+}
+
+pub trait KVStore {
```
bradjc, line 116:
```suggestion
pub trait KVSystem {
```

bradjc (approved):
I think this interface looks good, and we can always update it if needed. This should be sufficient for writing a shim capsule between TickV and this interface.

PR #2274, opened 2020-12-16
Update IEEE-802.15.4 driver to Tock 2.0

hudson-ayers:
### Pull Request Overview

This pull request updates the 15.4 driver to Tock 2.0. It also fixes a place where it was possible that callbacks would be delivered during a downcall by using dynamic_deferred_call.

This driver raised one question about the Tock 2.0 interface for me: how should capsules idiomatically ensure that they do not
overwrite buffers in use by userspace?

Before this PR, the 15.4 driver required a new call to `subscribe()` and a new call to `allow()` for every requested callback and for every time the driver might want to receive a packet. This was ensured by calling `take()` on the `Callback`/`AppSlice` that was being manipulated in an upcall to deliver a packet to an app. This design ensured that once a callback was registered, an app was guaranteed exclusive access to read/copy/modify the packet associated with that callback, contained within the buffer that it allowed the kernel when it called `receive()`. In Tock 2.0, the kernel can no longer un-allow an AppSlice itself -- only apps can -- so to provide the same guarantee, a capsule would have to internally track state indicating that it should not write to that `AppSlice` until the application has somehow indicated to the capsule that it is done reading whatever was in the `AppSlice`.
This design is a lot messier, and I suspect that many capsules will simply not consider this possibility, making it possible for the kernel to modify a buffer that a userspace app is only partially finished with reading.

I did not write code to address this issue yet because I am not sure how best to handle it.

### Testing Strategy

This pull request has been tested with libtock-c.

### TODO or Help Wanted

Answer to my exclusive AppSlice access question

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


lschuermann:
@hudson-ayers We talked about this issue (in another context) on Slack for a bit. Going along with [Phil's comments on another PR](https://github.com/tock/tock/pull/2258#discussion_r543522910), we should not introduce any side effects on an `allow` or `subscribe` operation, and in turn should probably also avoid any `command` causing modifications to the state `AppSlice`s or `Callback`s. In this regard, the Tock 2.0 interface in part forces you to write cleaner code.

I discovered this issue to affect the `UDP` driver too, so handling this was an issue also prior to the Tock 2.0 switch.

One quick solution to this issue is that userspace will acknowledge a reception by an explicit command, which tells the kernel that it may overwrite the shared buffer again. It does not scale particularly well though, especially when userspace is seldomly scheduled. It should not have any overhead or performance impact compared to the previous approach though.

In the future we're going to want to use a common architecture for this, the UDP driver and potentially others (such as IEEE802 TAP). It's probably best to publish our discussion results in an issue and get feedback from others. For now, as part of the mechanical Tock 2.0 migration, a simple change such as to require explicit ACKs seems like a good idea.

phil-levis:
Note that you technically aren't supposed to access a read/write application since until you request it back. But coupling a callback to a command (e.g., a `receive` call) means an application can receive a callback, either allow a new buffer and get the filled-in one back, or allow a zero-length and copy it out and call allow to give it back, then call the command. So you tie the next reception to the command.

capsules/src/ieee802154/driver.rs
```diff
@@ -523,33 +502,55 @@ impl framer::KeyProcedure for RadioDriver<'_> {
...
         match allow_num {
-            0 | 1 | 2 => self.do_with_app(appid, |app| {
-                match allow_num {
-                    0 => app.app_read = slice,
-                    1 => app.app_write = slice,
-                    2 => app.app_cfg = slice,
-                    _ => {}
+            0 | 1 => {
+                let res = self.apps.enter(appid, |app, _| match allow_num {
+                    0 => mem::swap(&mut app.app_read, &mut slice),
```
phil-levis, line 523: A nice side-effect of the swap approach is each case can be simple like this.

phil-levis (approved):

phil-levis:
I tried to resolve the conflict on `imix/src/main.rs`, now it seems like there are lots of compilation errors?

hudson-ayers:
> I tried to resolve the conflict on `imix/src/main.rs`, now it seems like there are lots of compilation errors?

Oh, I think Leon merged tock-2.0-dev with master to pull in the past few weeks of changes. I think that those changes included adding 15.4 support for several boards, so I will need to update the main file of each to reflect the changes to the 15.4 component.

lschuermann:
Yes, I pulled in master to avoid the Netlify CI breakage, include your `make prepush` fixes, etc. It's probably best to avoid diverging too far from master, so pulling it in every now and then seems like a good option. This also keeps the merge commits small and easy to check. Sorry for breaking your PR.

hudson-ayers:
No worries at all! This was inevitable once we merged anyway, and I'm all for frequent merges

hudson-ayers:
> Note that you technically aren't supposed to access a read/write application slice until you request it back. But coupling a callback to a command (e.g., a `receive` call) means an application can receive a callback, either allow a new buffer and get the filled-in one back, or allow a zero-length and copy it out and call allow to give it back, then call the command. So you tie the next reception to the command.

It doesn't seem that any of the uses of `getnstr_async()` in `libtock-c` do this today, even after the 2.0 update -- they just read from the allowed buffer once a callback arrives.

hudson-ayers:
Updated to resolve conflicts. This has now been tested and verified to work alongside the corresponding libtock-c PR.

PR #2275, opened 2020-12-16
nrf52: usb: read size after EPDATA

bradjc:
### Pull Request Overview

While working on the tock bootloader on top of CDC/USB I was running into issues where when the host was transferring a lot of data to the board (like doing an app load) the CDC software stack was not receiving all of the sent data. I eventually tracked this down to the nrf52 usb driver reporting a length of 0 to the upper USB layers, even when 64 bytes had been received (checked by printing bytes of the endpoint buffer). 

It seems that the endpoint out SIZE register can get overwritten before software has a chance to read it if SIZE is checked after the DMA finishes. So, I moved the SIZE read earlier in the transaction and now I'm receiving all of the data.

I was also running into issues where it seemed like interrupts were being missed, and I removed the `SIZE.set()` line that was happening after an OUT transaction has finished. According to the datasheet this is not needed and the DMA finishing is sufficient to re-enable the endpoint.

### Testing Strategy

This pull request was tested by running the Tock bootloader on top of the USB CDC stack and transferring 512 bytes quickly over USB to the device.

### TODO or Help Wanted

I need some help with the Rust. I need somewhere to store the actual size value, and I added it to an enum. However, it is pretty ugly getting it back out, and I couldn't figure out how to keep the asserts. Surely there is a more obvious way to do this.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/nrf52/src/usbd.rs
+++ b/chips/nrf52/src/usbd.rs
@@ -671,10 +671,12 @@ pub enum BulkOutState {
     // There is a pending OUT packet in this endpoint's buffer, to be read by
     // the client application.
     OutDelay,
-    // There is a pending EPDATA to reply to.
-    OutData,
-    // There is a pending DMA transfer on this OUT endpoint.
-    OutDma,
+    // There is a pending EPDATA to reply to. Store the size right after the
+    // EPDATA event.
+    OutData { size: u32 },
+    // There is a pending DMA transfer on this OUT endpoint. Still need to keep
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):
I tried to address the help wanted, let me know if that's good enough - getting the size is still a little unergonomic, using a match might be a little better, but I think it makes little difference.

bradjc:
Looks pretty good to me!

PR #2276, opened 2020-12-16
Update Gpio-async driver to 2.0

hudson-ayers:
### Pull Request Overview

This pull request updates the GPIO async driver to 2.0. It was very straightforward.

### Testing Strategy

This can only be tested on an acd52832 board, which I do not have, but it was such a straightforward port I am not very concerned anything would break.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


capsules/src/gpio_async.rs
```diff
@@ -25,27 +25,27 @@
...
 }
 
 impl<'a, Port: hil::gpio_async::Port> GPIOAsync<'a, Port> {
     pub fn new(ports: &'a [&'a Port]) -> GPIOAsync<'a, Port> {
         GPIOAsync {
-            ports: ports,
-            callback: OptionalCell::empty(),
-            interrupt_callback: OptionalCell::empty(),
+            ports,
+            callback: Cell::new(Callback::default()),
```
lschuermann, line 47: Hm, this only works because `Callback: Copy`, which I think we wanted to remove (at least to keep the Tock 2.0 callback swap protections in check). I'd read up on the core notes, but it seems the PR is not up yet.

Having `Callback: Copy` makes checking against capsule swaps substantially harder and increases overhead in the options I considered. Maybe we should remove that (with the risk of having to change a lot of code)?

That is a general question though, which should've been considered prior to giving the green light on porting. :smile: I'll publish my current approach on the anti-swap checks tomorrow, which `Copy` would break...

lschuermann (approved):
Looks good other than my comment on `Callback` implementing `Copy`, but that's a general note/question/concern, not specific to this PR.

capsules/src/gpio_async.rs
hudson-ayers: We can still use a `Cell` even without `Callback: Copy`. The only method I use that requires `Callback: Copy` is the `Cell::get()`, which could be replaced with `Cell::get_mut()`.

FWIW my analog_comparator to 2.0 PR used this same approach

phil-levis (approved):

capsules/src/gpio_async.rs
lschuermann: Yes, `get_mut` works without `Copy`, but requires a mutable reference of the `Cell`. I've identified `RefCell` to work in the case where we used to have an `OptionalCell<Callback>` in Tock 1.x, see #2282.

PR #2282, opened 2020-12-17
Tock 2.0: implement subscribe Callback swapping restrictions, part 1

lschuermann:
### Pull Request Overview

This pull request contains various commits, performing changes required to implement the Tock 2.0 subscribe system call, including the restrictions on capsules illegally swapping `Callback`s of other subscribe nums, different drivers or other processes.

This is quite a large changeset. I tried keeping the commits themselves clean and can suggest going through them individually. The "Files changes" view on GitHub is cluttered with mechanical capsule updates.

The changes are:

- `0807887`: removing a double (indirect) pointer inside of `Callback`.

  This was likely introduced due to the (slightly confusing) fact that `NonNull` is already a "pointer type", and as such does not need to have the pointer specified as part of it's generic type.

- `3885d20`: make the contained pointer in `ProcessCallback` optional (`NonNull<()>` -> `Option<NonNull>`).

  This allows us to differentiate between default instances of a `Callback` (ones which don't contain a `ProcessCallback` at all) vs. _null callbacks_, so explicitly set callbacks from a specific process which do not fire.

  This is required because we cannot check the `AppId` and `CallbackId` (driver num + subscribe num) on a default instance of a Callback, whereas we must check them on an explicit _null callback_ to ensure that the capsule has not swapped this `Callback`. Furthermore, the _null callback_ is allowed to contain `appdata` and this must not be leaked to other processes, which this change allows to assure.

- `2466a3a`, `064bb56`: remove the `Copy` trait implementation on `Callback`.

  All of these checks will ultimately rely on the fact that one cannot copy or clone a `Callback` type. With a `Copy` implementation a capsule might be able to hold onto a process-specific `Callback` and return the same (technically matching for this driver + subscribe num combination) instance on each call. Furthermore, a capsule might be able to schedule a callback which userspace believes has been unsubscribed from.

  The first commit includes all kernel changes, including `IPC`. I'd appreciate if someone more knowledgeable can confirm that I didn't break anything. @alevy?

  The second commit changes all of the capsules. Most capsules required a mechanical change from `OptionalCell` to `MapCell`, or taking a mutable reference instead of moving (copying).

- `39c4ade`: This rewrites the system call handler for `subscribe`.

  The subscribe system call is now handled as a method on `dyn ProcessType`, similar to the allow system calls.

  Checks have been introduced that a capsule must return the previous `Callback` or a default instance in the success case, and the passed parameter in the error case.

   This is not yet sufficient to guarantee that a capsule cannot hold onto a `Callback`, but severely limits the ways it can misbehave w.r.t. the Tock 2.0 subscribe semantics. Most importantly, this prevents leaking data from one process to another implicity through   the function pointer and appdata of a `Callback` instance.

All of these changes should have minimal runtime complexity and memory overhead. Removing the `Copy` trait will cause some inefficiencies due to the use of `MapCell` and `RefCell`, but this appears strictly necessary if we want to avoid a central lookup table for valid Callbacks.

Notably, this does not introduce any memory overhead in the `Callback` type and requires no external database keeping track of `Callback` instances.

### Testing Strategy

This pull request was tested by compiling without the `touch` driver.

### TODO or Help Wanted

This pull request still needs
- changing the touch driver to use `MapCell`

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/callback.rs
+++ b/kernel/src/callback.rs
@@ -172,7 +172,7 @@ struct ProcessCallback {
     app_id: AppId,
     callback_id: CallbackId,
     appdata: usize,
-    fn_ptr: NonNull<*mut ()>,
+    fn_ptr: NonNull<()>,
 }
 
 #[derive(Clone, Copy, Default)]
@@ -185,7 +185,7 @@ impl Callback {
         app_id: AppId,
         callback_id: CallbackId,
         appdata: usize,
-        fn_ptr: NonNull<*mut ()>,
+        fn_ptr: NonNull<()>,
     ) -> Callback {
         Callback {
             cb: Some(ProcessCallback::new(app_id, callback_id, appdata, fn_ptr)),
@@ -223,7 +223,7 @@ impl ProcessCallback {
         app_id: AppId,
         callback_id: CallbackId,
         appdata: usize,
-        fn_ptr: NonNull<*mut ()>,
+        fn_ptr: NonNull<()>,
     ) -> ProcessCallback {
         ProcessCallback {
             app_id,
--- a/kernel/src/callback.rs
+++ b/kernel/src/callback.rs
@@ -172,7 +172,6 @@ pub struct CallbackId {
 /// In contrast to the contained [`ProcessCallback`], this type does
 /// not actually have to point to a userspace process. This is the
 /// case in the [default instances](<Callback as Default>::default).
-#[derive(Clone, Copy)]
 pub struct Callback(Option<ProcessCallback>);
 
 impl Callback {
@@ -228,7 +227,7 @@ impl Callback {
     ///
     /// Otherwise, this function returns `true`.
     pub fn schedule(&mut self, r0: usize, r1: usize, r2: usize) -> bool {
-        self.0.map_or(true, |mut cb| cb.schedule(r0, r1, r2))
+        self.0.as_mut().map_or(true, |cb| cb.schedule(r0, r1, r2))
     }
 
     pub(crate) fn into_subscribe_success(self) -> GenericSyscallReturnValue {
@@ -305,7 +304,6 @@ impl Default for Callback {
 /// A [`ProcessCallback`] may be a _null callback_, not pointing to a
 /// valid function. In this case, the callback won't actually be
 /// called in userspace.
-#[derive(Clone, Copy)]
 struct ProcessCallback {
     app_id: AppId,
     callback_id: CallbackId,
--- a/kernel/src/callback.rs
+++ b/kernel/src/callback.rs
@@ -8,8 +8,6 @@ use crate::config;
 use crate::debug;
 use crate::process;
 use crate::sched::Kernel;
-use crate::syscall::GenericSyscallReturnValue;
-use crate::ErrorCode;
 
 /// Userspace app identifier.
 ///
@@ -160,8 +158,8 @@ impl AppId {
 /// This contains the driver number and the subscribe number within the driver.
 #[derive(Copy, Clone, PartialEq, Debug)]
 pub struct CallbackId {
-    pub driver_num: usize,
-    pub subscribe_num: usize,
+    pub driver_num: u32,
+    pub subscribe_num: u32,
 }
 
 /// A callback to userspace
@@ -169,9 +167,12 @@ pub struct CallbackId {
 /// This is essentially a wrapper around a function pointer with
 /// additional data.
 ///
-/// In contrast to the contained [`ProcessCallback`], this type does
+/// In contrast to the contained `ProcessCallback`, this type does
 /// not actually have to point to a userspace process. This is the
-/// case in the [default instances](<Callback as Default>::default).
+/// case in the [default instances](Callback::default).
+///
+/// A default instance will never schedule an actual callback to
+/// userspace.
 pub struct Callback(Option<ProcessCallback>);
 
 impl Callback {
@@ -195,6 +196,14 @@ impl Callback {
             appdata,
             fn_ptr,
         )))
+    }
+
+    /// Get the contained [`ProcessCallback`] struct
+    ///
+    /// If the [`Callback`] refers to a process, this returns the
+    /// contained [`ProcessCallback`] struct.
+    pub(crate) fn into_inner(self) -> Option<ProcessCallback> {
+        self.0
     }
 
     /// Attempt to trigger the callback.
@@ -229,57 +238,6 @@ impl Callback {
     pub fn schedule(&mut self, r0: usize, r1: usize, r2: usize) -> bool {
         self.0.as_mut().map_or(true, |cb| cb.schedule(r0, r1, r2))
     }
-
-    pub(crate) fn into_subscribe_success(self) -> GenericSyscallReturnValue {
-        match self.0 {
-            None => {
-                // Default instance of a [`Callback`]
-                GenericSyscallReturnValue::SubscribeSuccess(0 as *mut u8, 0)
-            }
-            Some(cb) => {
-                // Process callback
-                match cb.fn_ptr {
-                    Some(ptr) => {
-                        // Valid callback
-                        GenericSyscallReturnValue::SubscribeSuccess(
-                            ptr.as_ptr() as *const u8,
-                            cb.appdata,
-                        )
-                    }
-                    None => {
-                        // Null callback
-                        GenericSyscallReturnValue::SubscribeSuccess(0 as *mut u8, 0)
-                    }
-                }
-            }
-        }
-    }
-
-    pub(crate) fn into_subscribe_failure(self, err: ErrorCode) -> GenericSyscallReturnValue {
-        match self.0 {
-            None => {
-                // Default instance of a [`Callback`]
-                GenericSyscallReturnValue::SubscribeFailure(err, 0 as *mut u8, 0)
-            }
-            Some(cb) => {
-                // Process callback
-                match cb.fn_ptr {
-                    Some(ptr) => {
-                        // Valid callback
-                        GenericSyscallReturnValue::SubscribeFailure(
-                            err,
-                            ptr.as_ptr() as *const u8,
-                            cb.appdata,
-                        )
-                    }
-                    None => {
-                        // Null callback
-                        GenericSyscallReturnValue::SubscribeFailure(err, 0 as *mut u8, 0)
-                    }
-                }
-            }
-        }
-    }
 }
 
 impl Default for Callback {
@@ -304,11 +262,11 @@ impl Default for Callback {
 /// A [`ProcessCallback`] may be a _null callback_, not pointing to a
 /// valid function. In this case, the callback won't actually be
 /// called in userspace.
-struct ProcessCallback {
-    app_id: AppId,
-    callback_id: CallbackId,
-    appdata: usize,
-    fn_ptr: Option<NonNull<()>>,
+pub(crate) struct ProcessCallback {
+    pub(crate) app_id: AppId,
+    pub(crate) callback_id: CallbackId,
+    pub(crate) appdata: usize,
+    pub(crate) fn_ptr: Option<NonNull<()>>,
 }
 
 impl ProcessCallback {
--- a/kernel/src/callback.rs
+++ b/kernel/src/callback.rs
@@ -164,58 +164,153 @@ pub struct CallbackId {
     pub subscribe_num: usize,
 }
 
-/// Type for calling a callback in a process.
+/// A callback to userspace
 ///
-/// This is essentially a wrapper around a function pointer.
+/// This is essentially a wrapper around a function pointer with
+/// additional data.
+///
+/// In contrast to the contained [`ProcessCallback`], this type does
+/// not actually have to point to a userspace process. This is the
+/// case in the [default instances](<Callback as Default>::default).
+#[derive(Clone, Copy)]
+pub struct Callback(Option<ProcessCallback>);
+
+impl Callback {
+    /// Construct a new [`Callback`] pointing to a process
+    ///
+    /// This will create a [`ProcessCallback`] internally, hence
+    /// pointing to a userspace process, in constrast to the [default
+    /// instances](<Callback as Default::default).
+    ///
+    /// The passed pointer may be `None`, in which case the Callback
+    /// will not actually be scheduled (creating a _null callback_).
+    pub(crate) fn new(
+        app_id: AppId,
+        callback_id: CallbackId,
+        appdata: usize,
+        fn_ptr: Option<NonNull<()>>,
+    ) -> Self {
+        Callback(Some(ProcessCallback::new(
+            app_id,
+            callback_id,
+            appdata,
+            fn_ptr,
+        )))
+    }
+
+    /// Attempt to trigger the callback.
+    ///
+    /// If this callback refers to an actual function pointer (not a
+    /// _null callback_) and the process is valid, this will queue the
+    /// [`Callback`] for the associated process.
+    ///
+    /// ## Arguments
+    /// The arguments (`r0-r2`) are the values passed back to the
+    /// process and are specific to the individual `Driver`
+    /// interfaces.
+    ///
+    /// ## Return value
+    /// The following conditions are checked in this order to
+    /// determine the return value:
+    ///
+    /// 1. If this [`Callback`] is a default instance (does not refer
+    ///    to a process), it returns `true` and behaves as if the
+    ///    callback was scheduled.
+    ///
+    /// 2. This function returns `false` if the process is no longer
+    ///    active.
+    ///
+    /// 3. For a _null callback_, this function will return `true` and
+    ///    behave as if the callback was scheduled.
+    ///
+    /// 4. This function returns `false` if the queue for the process
+    ///    is full and the callback could not be scheduled.
+    ///
+    /// Otherwise, this function returns `true`.
+    pub fn schedule(&mut self, r0: usize, r1: usize, r2: usize) -> bool {
+        self.0.map_or(true, |mut cb| cb.schedule(r0, r1, r2))
+    }
+
+    pub(crate) fn into_subscribe_success(self) -> GenericSyscallReturnValue {
+        match self.0 {
+            None => {
+                // Default instance of a [`Callback`]
+                GenericSyscallReturnValue::SubscribeSuccess(0 as *mut u8, 0)
+            }
+            Some(cb) => {
+                // Process callback
+                match cb.fn_ptr {
+                    Some(ptr) => {
+                        // Valid callback
+                        GenericSyscallReturnValue::SubscribeSuccess(
+                            ptr.as_ptr() as *const u8,
+                            cb.appdata,
+                        )
+                    }
+                    None => {
+                        // Null callback
+                        GenericSyscallReturnValue::SubscribeSuccess(0 as *mut u8, 0)
+                    }
+                }
+            }
+        }
+    }
+
+    pub(crate) fn into_subscribe_failure(self, err: ErrorCode) -> GenericSyscallReturnValue {
+        match self.0 {
+            None => {
+                // Default instance of a [`Callback`]
+                GenericSyscallReturnValue::SubscribeFailure(err, 0 as *mut u8, 0)
+            }
+            Some(cb) => {
+                // Process callback
+                match cb.fn_ptr {
+                    Some(ptr) => {
+                        // Valid callback
+                        GenericSyscallReturnValue::SubscribeFailure(
+                            err,
+                            ptr.as_ptr() as *const u8,
+                            cb.appdata,
+                        )
+                    }
+                    None => {
+                        // Null callback
+                        GenericSyscallReturnValue::SubscribeFailure(err, 0 as *mut u8, 0)
+                    }
+                }
+            }
+        }
+    }
+}
+
+impl Default for Callback {
+    /// Construct a new default [`Callback`]
+    ///
+    /// A default [`Callback`] instance will not point to any actual
+    /// userspace process. Therefore, no actual callbacks will be
+    /// scheduled by this instance.
+    fn default() -> Self {
+        Callback(None)
+    }
+}
+
+/// [`Callback`] referring to an actual process
+///
+/// This is essentially a wrapper around a function pointer with
+/// additional data.
+///
+/// The [`ProcessCallback`], in contrast to [`Callback`], is always
+/// explicity created by and for a specific process.
+///
+/// A [`ProcessCallback`] may be a _null callback_, not pointing to a
+/// valid function. In this case, the callback won't actually be
+/// called in userspace.
 #[derive(Clone, Copy)]
 struct ProcessCallback {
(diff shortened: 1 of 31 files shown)


kernel/src/callback.rs
```diff
@@ -232,54 +283,70 @@ impl ProcessCallback {
...
+                            task_res,
+                        );
+                    }
+
+                    task_res
+                } else {
+                    // This is a null callback, behave as if the
+                    // callback was scheduled
+                    true
+                }
```
hudson-ayers, line 349: not printing that a null callback was scheduled when trace_syscalls is enabled seems like it could lead to confusion when a userspace bug leads to accidentally passing a null callback

kernel/src/callback.rs
```diff
@@ -160,70 +158,123 @@ impl AppId {
...
+impl Default for Callback {
+    /// Construct a new default [`Callback`]
+    ///
+    /// A default [`Callback`] instance will not point to any actual
+    /// userspace process. Therefore, no actual callbacks will be
+    /// scheduled by this instance.
+    fn default() -> Self {
+        Callback(None)
     }
 }
```
hudson-ayers, line 252: why manually implement default when the derived version would be the same?

hudson-ayers (commented):
I am still somewhat confused by the need for seperate "None Callbacks" and "null callbacks", and why we can't just have the kernel convert null callbacks to none callbacks (or default callbacks as you call them) when such a subscribe is made.

I am guessing you explained this on a call and I missed it, but I think it would be good for the justification to live in this PR somewhere anyway.

libraries/tock-cells/src/map_cell.rs
```diff
@@ -86,6 +86,23 @@ impl<T> MapCell<T> {
...
     }
 
+    /// Sets the `MapCell` value.
+    ///
+    /// If the passed value is a `Some(T)`, the MapCell will be set to
+    /// be occupied with this value. If the passed value is a `None`,
+    /// the `MapCell` will not be occupied.
+    ///
+    /// The previous value will be returned.
+    pub fn set(&self, val: Option<T>) -> Option<T> {
```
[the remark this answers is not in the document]
lschuermann: I guess that would work, but it's not going to be a single-method replacement of `OptionalCell::insert` as previously used, but rather requires the expression `if let Some(cb) = callback { self.callback.replace(cb) } else { self.callback.take() }`. This is really only required for the remaining Tock 1.x capsules, as in 2.0 those `OptionalCell`s would be replaced by `RefCell`s accordingly. Would you be fine integrating this temporarily, potentially under another name?

kernel/src/callback.rs
lschuermann: Agreed. We should introduce a trace message there as well.

kernel/src/callback.rs
lschuermann: I'm implementing manually for the documentation comment (as this is a publicly exposed API), which will be visible in the generated rustdoc. I can switch to the derive macro if you want.

lschuermann:
> I am still somewhat confused by the need for seperate "None Callbacks" and "null callbacks"

I tried to explain this in my comment on `017f368` of the PR message. Essentially, for default callbacks we can't enforce any constraints as they contain no information on the process, driver num and subscribe num they're created over (or actually, aren't).

However, an explicitly created _null callback_ is something different: it's a valid callback for userspace to register, along with a passed `userdata`.

We should differentiate between the two for two reasons:
- Not differentiating would mean that we'd have to move the `userdata` from `ProcessCallback` into `Callback`. Because we can't enforce constraints on the null (and default) callbacks then - as a capsule might rightfully return the default callback on the first subscribe num invocation, which wouldn't have matching process id, driver num and subscribe num - a capsule could technically leak the `userdata` of one app into the other.

- Being able to identify a callback as a default callback (vs. a null callback), along with later introducing a capability-based constraint on creating these default instances, will limit the ability for capsules to arbitrarily create and return default instances, where it should have returned a previous non-default instance instead. If we can't distinguish between the two, we're going to have a hard time enforcing such constraints without a central table (which I'd like to avoid to minimize the overhead).

This is my fourth iteration on the implementation of these checks, after some headaches. I hope my explanation is somewhat clear, but it could also be my reasoning is wrong here. But you're absolutely right, this PR should contain an expressive description of why this is required -- which I hope this satisfies.

kernel/src/process.rs
```diff
@@ -1212,6 +1255,125 @@ impl<C: Chip> ProcessType for Process<'_, C> {
...
+            Ok(returned_callback) => {
+                // The capsule indicated that the subscribe operation succeeded
+                match returned_callback.into_inner() {
+                    None => {
+                        // The capsule returned a default callback, we
+                        // have to believe this to be correct
+                        //
+                        // TODO: Limit the default constructor on
+                        // Callback to require a capability for Grant
+                        // initializations
```
jrvanwhy, line 1346: I don't see how using capabilities to limit calls to the default constructor can prevent capsules from storing callbacks across multiple `subscribe` calls.

Consider a capsule that contains a `[Callback; 2]`. The first time `subscribe` is called, it swaps the app's callback with the first entry in the array. The second time, it swaps the app's callback with the second entry in the array. Now the capsule has both the old and new callback, and the app is unaware the callback is stored.

lschuermann: I suppose you're right, there's always going to be a way for a capsule to hold onto `Callback`s for capsules, if they are allowed to return a default Callback. I was also concerned on how to create default instances requiring a capability for `Callback`s not stored in Grants, that would've required passing them in from the `main.rs`.

lschuermann:
Thanks for the reviews @hudson-ayers and @jrvanwhy. I see that limiting the default constructor on `Callback` is not the way to go here - even if we require a capability in this constructor, a capsule can still hold onto `Callback`s and we can't really avoid that. 

Going to plan B: What if we were to enforce that the subscribe numbers are densely packed per driver, and the kernel (possibly via the platform trait) would know this max subscribe number. We could then, between the Grant initialization and the process start, call out to the driver and subscribe an explicit _null callback_ (with appdata `0` such that we conform to the TRD) for this process, driver and subscribe number, for all subscribe numbers in `0..SUBSCRIBE_COUNT`. This way, we can enforce that in the initial subscribe calls, the driver **must** return a _default callback_, whereas on subsequent subscribe system calls the driver **must** return a matching _null callback_. @jrvanwhy Would that work?

This has the advantage of still avoiding a central registry and not adding any memory requirements, just a one-time cost per registered driver & process start.

jrvanwhy:
> Going to plan B: What if we were to enforce that the subscribe numbers are densely packed per driver, and the kernel (possibly via the platform trait) would know this max subscribe number. We could then, between the Grant initialization and the process start, call out to the driver and subscribe an explicit _null callback_ (with appdata `0` such that we conform to the TRD) for this process, driver and subscribe number, for all subscribe numbers in `0..SUBSCRIBE_COUNT`. This way, we can enforce that in the initial subscribe calls, the driver **must** return a _default callback_, whereas on subsequent subscribe system calls the driver **must** return a matching _null callback_. Given that we introduce at most one valid instance for each process per (driver, subscribe_num) combination, and during runtime enforce that we always exchange one for another, I don't see any way for a driver to hold onto a `Callback`. @jrvanwhy Would that work?

I believe that would work, yes.

No comment how expensive it is or whether the trade-off (which is avoiding checks in synchronous userspace code) is worth it.

kernel/src/callback.rs
hudson-ayers: no, that makes sense

libraries/tock-cells/src/map_cell.rs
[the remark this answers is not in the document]
hudson-ayers: I am fine with it temporarily under a different name -- maybe `replace_opt()`?

lschuermann:
Sorry it took me a while to get back to this.

> I have no comment on how expensive it is or whether the trade-off (which is avoiding checks in synchronous userspace code) is worth it.

@jrvanwhy I'm sorry if my previous comments seemed like I wanted to force a statement in this direction. While I do try to optimize regarding these criteria, I'd say that it only makes sense to determine the resource consumption and therefore also whether this check is worth doing after we have an implementation.

Nonetheless, I think that the approach to "swap out" the default callbacks on a process start (restart) is worth trying. In a crude proof of concept this seems to be working, although some changes to the `Platform` trait might be necessary. This patch set does not include that component, but rather only the more "uncontroversial" changes, which also serve as a basis for that.

Even if we decide that this is not worth doing, this patchset ensures that future capsules will be written with the constraint of `Callback` not implementing `Copy` (which is required for virtually all ownership-based restrictions that could also be introduced in the future, and I think was also the original intention here). Also it does limited the number of ways a capsule can misbehave w.r.t. the Tock 2.0 subscribe semantics, while being (in terms of memory) essentially free. Therefore, I'd appreciate if we can merge this nonetheless. In the worst case, a revert of a selection of the changes should work.

libraries/tock-cells/src/map_cell.rs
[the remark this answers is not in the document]
lschuermann: I have removed the offending commit from the patchset. Since there were not too many capsules still using `LegacyDriver`, migrating these to the `if let` expression wasn't too bad.

jrvanwhy:
> > I have no comment on how expensive it is or whether the trade-off (which is avoiding checks in synchronous userspace code) is worth it.
> 
> @jrvanwhy I'm sorry if my previous comments seemed like I wanted to force a statement in this direction. While I do try to optimize regarding these criteria, I'd say that it only makes sense to determine the resource consumption and therefore also whether this check is worth doing after we have an implementation.

No worries, I only wrote that because I didn't want my "I believe that would work, yes" comment to be misinterpreted as "I would prefer that solution".

kernel/src/callback.rs
lschuermann: I've added an explicit message that a null callback was (not) scheduled.

hudson-ayers (commented):
I have looked through the most recent code, and I don't have any issues with it. But I am wary of merging it when we do not yet know for sure that it will work as a basis for callback swapping restrictions, as I think there is a decent amount of added complexity here. I am going to lay out my current understanding of the pros/cons:

Pros:
- lays groundwork for a low-memory-overhead approach to ensure that a capsule can never call a callback that userspace believes has been revoked
- some additional checks on capsules correctly using subscribe (capsules cannot pretend subscribe failed but keep the callback, can't schedule on inactive process, etc.)

Cons: 
- As standalone changes, these offer little in the way of guarantees for userspace, and thus are dependent on a future PR to be useful -- but it's hard to know for sure they will work until that future PR arrives.
- non-virtualized drivers now have to use `RefCell` / `MapCell` to hold callbacks.
- additional opportunities for capsule code to cause kernel panics (when a capsule returns the wrong callback)
-  capsules cannot copy callbacks, which can complicate their use in closures (see changes to 15.4 driver)
- lots of changes to existing capsules which won't be needed once those capsules are ported to 2.0, and some merge conflicts already (adc)
- reasoning about "Default Callbacks" vs. "null callbacks" is confusing, and leads to a decent amount of error checking code that is just an artifact of this particular approach
- several hundred lines of changed code in `sched.rs`, `process.rs`, and `callback.rs`

Please let me know if I missed anything. I would like for others to weigh in, but I think it might make sense to wait on merging this.

hudson-ayers:
Question:

If we have to enforce that subscribe numbers are densely packed for each driver, should capsules directly manage callbacks at all? My understanding is that the reason that capsules currently manage callbacks is so that callbacks can be stored in the grant region -- otherwise scaling the number of callbacks based on the number of processes is hard.

But if the subscribe numbers are densely packed, could we just have the kernel store callbacks in some data structure with `SomeDriver::SUBSCRIBE_COUNT * NUM_PROCS` callback slots per driver? I suppose this would still use memory for inactive processes, which the current approach does not...

lschuermann:
I tend to disagree regarding some of the Cons you mentioned and would like to explain my reasoning.

> reasoning about "Default Callbacks" vs. "null callbacks" is confusing, and leads to a decent amount of error checking code that is just an artifact of this particular approach

I'd argue that this is not just an artifact of this particular approach, but rather a basic concept of the Tock 2.0 subscribe semantics. A capsule registering a null callback is an explicit operation, exactly as registering a non-null callback is. The fact that we need a default Callback at all is due to the fact that @phil-levis didn't want to wrap it in an `Option<>` explicitly, which I find works great. However, a default callback is _not_ a null callback, as a null callback (and any other callback for that matter) has additional values (appdata) assigned and therefore fundamentally belongs to a process. If we wouldn't distinguish, than we'd have to include an appdata field on every callback (also the default one). Also, we'd have to store the `AppId`, as we can't risk leaking the appdata of one process into another as part of it's return value. If we do that, we'd have essentially the same construct as I'm introducing in this PR (possibly with a different type layout).

> non-virtualized drivers now have to use `RefCell` / `MapCell` to hold callbacks

I think this will have to change anyways. Given that we can't return a `Callback` of one process to another (that would be leaking the appdata as part of the return value, we agreed on checking this on a call), we can't have "one callback" per capsule, but should rather -- also in non-virtualized drivers -- store them in a process grant structure and simply mark which process is the active client. This also makes sense from a semantic point of view: a callback is a process resource, and thus should be kept in the Grant, which will also make it "vanish" when the process dies.

> As standalone changes, these offer little in the way of guarantees for userspace

They offer the very significant guarantee that a capsule can't implicitly leak process data to another as part of a Callback, as stated above.

> additional opportunities for capsule code to cause kernel panics

I agree, although I think it's the best approach here (given that we can't seem to check this statically at compile time). I can offer to disable these checks in the kernel, but the discussions in the call seemed in favor of the kernel being responsible for checking the capsule correctness regarding subscribe semantics. We can also introduce mechanisms to block a capsule's operations (as it is clearly incorrect) as an alternative to panicing.

> lots of changes to existing capsules which won't be needed once those capsules are ported to 2.0

Can you explain how that is an issue? I've done the work, and they're going to be ported anyways. I'm keeping this PR up to date by rebasing. By doing this I'd like to enable getting these checks in _while_ we're porting capsules and while I've got time. I don't want 2.0 to be blocked if we only start caring about this once all capsules are already ported.

> several hundred lines of changed code in sched.rs, process.rs, and callback.rs

I can try to make this changeset smaller, but prefer keeping it readable and refactor where it makes sense. I was under the impression we were okay with larger changes to these core components as part of 2.0 if they are justified. We've already been touching these files quite a bit.

hudson-ayers:
> I'd argue that this is not just an artifact of this particular approach, but rather a basic concept of the Tock 2.0 subscribe semantics. A capsule registering a null callback is an explicit operation, exactly as registering a non-null callback is (in fact, from an app's view we don't distinguish). The fact that we need a default Callback at all is due to the fact that @phil-levis didn't want to wrap it in an `Option<>` explicitly, which I find works great. However, a default callback is _not_ a null callback, as a null callback (and any other callback for that matter) has additional values (appdata) assigned and therefore fundamentally belongs to a process. If we wouldn't distinguish, than we'd have to include an appdata field on every callback (also the default one). Also, we'd have to store the `AppId`, as we can't risk leaking the appdata of one process into another as part of it's return value. If we do that, we'd have essentially the same construct as I'm introducing in this PR (possibly with a different type layout).

It seems to me that an implementation using a lookup table would not need to differentiate between default and null callbacks. Everywhere that we currently construct a null callback could just use a default callback instead. Maybe there is some fundamental aspect here I am missing.

> I think this will have to change anyways. Given that we can't return a `Callback` of one process to another (that would be leaking the appdata as part of the return value, we agreed on checking this on a call), we can't have "one callback" per capsule, but should rather -- also in non-virtualized drivers -- store them in a process grant structure and simply mark which process is the active client. This also makes sense from a semantic point of view: a callback is a process resource, and thus should be kept in the Grant, which will also make it "vanish" when the process dies.

This makes sense. If Tock 2.0 requires that all apps use Grant regions to enforce exclusive access of non-virtualized drivers, should that be added to the porting guide?

> They offer the very significant guarantee that a capsule can't implicitly leak process data to another as part of a Callback, as stated above.

I thought that this guarantee is not offered without part 2, which we don't have yet? Such as in the example that Johnathan gave?

> I agree, although I think it's the best approach here (given that we can't seem to check this statically at compile time). I can offer to disable these checks in the kernel, but the discussions in the call seemed in favor of the kernel being responsible for checking the capsule correctness regarding subscribe semantics. We can also introduce mechanisms to block a capsule's operations (as it is clearly incorrect) as an alternative to panicing.

I agree that panicking is the best way to handle this for now. I am just pointing out that until we have part 2, this adds panics (a downside) without the upside of the leak guarantees for userspace.

> Can you explain how that is an issue? I've done the work, and they're going to be ported anyways. I'm keeping this PR up to date by rebasing. By doing this I'd like to enable getting these checks in _while_ we're porting capsules and while I've got time. I don't want 2.0 to be blocked if we only start caring about this once all capsules are already ported.

I just mean that if people have begun porting capsules already, then this is merged, it will be a merge conflict for those people. Not a huge deal but I already noticed it happened with the adc to 2.0 PR. With the adc capsule, for example, the port to 2.0 makes the changes to the adc in this PR unneeded.

> I can try to make this changeset smaller, but prefer keeping it readable and refactor where it makes sense. I was under the impression we were okay with larger changes to these core components as part of 2.0 if they are justified. We've already been touching these files quite a bit.

I don't think you need to make the changeset smaller. My concern here is that we make a bunch of changes to the kernel, then find that implementing part 2 does not work quite as expected, and reveals some of these changes to be unnecessary. It just seems easier to evaluate these changes when we can see exactly how they enable part 2, but if we merge now people are unlikely to revisit this already-merged PR whenever part 2 arrives.

lschuermann:
Gotcha. We can wait until we reach a point at which we're fairly certain what direction we're going in, prior to merging this changeset. I still have a strong feeling that we're going to want almost all of this as long as we're not drastically changing the way Subscribe is architected in Tock (see below), but that's fine. I just don't have the time to keep rebasing this with the changes in `tock-2.0-dev`, so closing this for now (as it has no chance of being merged).

> It seems to me that an implementation using a lookup table would not need to differentiate between default and null callbacks. 

That might be true. You raise an interesting question about using a lookup table here too (as for allow, that currently seems like the best approach). I've build PoCs using lookup tables and always found them to be somewhat expensive, as a capsule will always have to hold onto at least an `AppId` or similar (which could be retrieved by iterating over a Grant), invoke the kernel with that, go through some references and arrive at the Callback data in the table. This might be worth pursuing, though is a very drastic change in the way the Subscribe is mapped to semantics within capsules and the kernel at large. However, even for allow, for efficiency reasons my current approach is to _also_ keep the pointer and length in the AppSlice itself to avoid going through a chain of references and (if required, expensive) table searches. I can try to further explore in this direction.

> > They offer the very significant guarantee that a capsule can't implicitly leak process data to another as part of a Callback, as stated above.
> 
> I thought that this guarantee is not offered without part 2, which we don't have yet? Such as in the example that Johnathan gave?

Actually this guarantee is offered: this changeset is in itself sufficient to guarantee that if a capsule returns any type of `Callback` expect for default instances, it _must_ be a `Callback` of the corresponding `driver_num`, `subscribe_num`, and matching `AppId`. Therefore, it prevents information leaks from one process to another through `Callback` instances. It's not yet sufficient to guarantee that a capsule cannot hold on to a `Callback` and schedule it after userspace has revoked it, neither can it guarantee that userspace will get the exact `Callback` back which it subscribed initially (as returning a default `Callback` is always allowed). This explains the necessity for null vs. default callbacks.

Because we can make all of these guarantees already, because on the call @phil-levis mentioned that we're likely going to remove `Copy`, and it was mentioned that exactly these guarantees (against information leakage) are the basic checks we're going to want to do anyways, I opened this "firt half" of the changeset in the first place. (As stated below) this PR would rightfully break all ported non-virtualized drivers, because they're leaking information in the way they're currently architected.

> This makes sense. If Tock 2.0 requires that all apps use Grant regions to enforce exclusive access of non-virtualized drivers, should that be added to the porting guide?

To be honest, I only noticed this implication while developing these protection mechanisms. In hindsight, it should've been added and my suspicion is that we're going to have to change the non-virtualized drivers accordingly (as currently, if I'm not mistaken, they are incorrect regardless of how we're going to enforce it). If we're now waiting on this to settle, we can probably also fix these ported capsules after we've reached a solution here.

hudson-ayers:
> Gotcha. We can wait until we reach a point at which we're fairly certain what direction we're going in, prior to merging this changeset. I still have a strong feeling that we're going to want almost all of this as long as we're not drastically changing the way Subscribe is architected in Tock (see below), but that's fine. I just don't have the time to keep rebasing this with the changes in `tock-2.0-dev`, so closing this for now (as it has no chance of being merged).

I don't mean to suggest this would never be merged, I don't speak for the whole core team! I was just trying to voice my concerns based on my current understanding. 
 
> That might be true. You raise an interesting question about using a lookup table here too (as for allow, that currently seems like the best approach). I've build PoCs using lookup tables and always found them to be somewhat expensive, as a capsule will always have to hold onto at least an `AppId` or similar (which could be retrieved by iterating over a Grant), invoke the kernel with that, go through some references and arrive at the Callback data in the table. This might be worth pursuing, though is a very drastic change in the way the Subscribe is mapped to semantics within capsules and the kernel at large. However, even for allow, for efficiency reasons my current approach is to _also_ keep the pointer and length in the AppSlice itself to avoid going through a chain of references and (if required, expensive) table searches. I can try to further explore in this direction.

I also don't want to say you need to explore this more now, I just mean that until we know for sure part 2 works as expected it is hard to know all portions of this PR will be needed.

> Actually this guarantee is offered: this changeset is in itself sufficient to guarantee that if a capsule returns any type of `Callback` expect for default instances, it _must_ be a `Callback` of the corresponding `driver_num`, `subscribe_num`, and matching `AppId`. Therefore, it prevents information leaks from one process to another through `Callback` instances. It's not yet sufficient to guarantee that a capsule cannot hold on to a `Callback` and schedule it after userspace has revoked it, neither can it guarantee that userspace will get the exact `Callback` back which it subscribed initially (as returning a default `Callback` is always allowed). This explains the necessity for null vs. default callbacks.

Oh, this is a very good point, and something I misunderstood.

> Because we can make all of these guarantees already, because on the call @phil-levis mentioned that we're likely going to remove `Copy`, and it was mentioned that exactly these guarantees (against information leakage) are the basic checks we're going to want to do anyways, I opened this "firt half" of the changeset in the first place. (As stated below) this PR would rightfully break all ported non-virtualized drivers, because they're leaking information in the way they're currently architected.
> 
> > This makes sense. If Tock 2.0 requires that all apps use Grant regions to enforce exclusive access of non-virtualized drivers, should that be added to the porting guide?
> 
> To be honest, I only noticed this implication while developing these protection mechanisms. In hindsight, it should've been added and my suspicion is that we're going to have to change the non-virtualized drivers accordingly (as currently, if I'm not mistaken, they are incorrect regardless of how we're going to enforce it). If we're now waiting on this to settle, we can probably also fix these ported capsules after we've reached a solution here.

Well, it only breaks ported non-virtualized drivers if multiple apps try to use them, right? In which case they would have worked incorrectly anyway (though now the kernel will panic, when before there might have just been missing callbacks or something)

PR #2284, opened 2020-12-18
Proximity driver to Tock 2.0 system-call interface

alevy:
### Pull Request Overview

This pull request Port the proximity driver to the Tock 2.0 system-call Driver interface.

I simplified `enqueue_command` a bit by moving the logic into the enter closure.

### Testing Strategy

So far, only compiled-tested. I don't have a board with the right peripheral (the nanoble-sense or the clue_nrf52840).

### TODO or Help Wanted

Can @arjundeopujari (original author of the driver) maybe test this quickly? Or else someone with one of the two boards that have this driver?

### Documentation Updated

- [] ~~Updated the relevant files in `/docs`, or no updates are required.~~

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/clue_nrf52840/src/main.rs
+++ b/boards/clue_nrf52840/src/main.rs
@@ -136,7 +136,7 @@ impl kernel::Platform for Platform {
     {
         match driver_num {
             capsules::console::DRIVER_NUM => f(Some(Ok(self.console))),
-            capsules::proximity::DRIVER_NUM => f(Some(Err(self.proximity))),
+            capsules::proximity::DRIVER_NUM => f(Some(Ok(self.proximity))),
             capsules::gpio::DRIVER_NUM => f(Some(Err(self.gpio))),
             capsules::alarm::DRIVER_NUM => f(Some(Err(self.alarm))),
             capsules::led::DRIVER_NUM => f(Some(Ok(self.led))),
--- a/boards/nano33ble/src/main.rs
+++ b/boards/nano33ble/src/main.rs
@@ -124,7 +124,7 @@ impl kernel::Platform for Platform {
     {
         match driver_num {
             capsules::console::DRIVER_NUM => f(Some(Ok(self.console))),
-            capsules::proximity::DRIVER_NUM => f(Some(Err(self.proximity))),
+            capsules::proximity::DRIVER_NUM => f(Some(Ok(self.proximity))),
             capsules::gpio::DRIVER_NUM => f(Some(Err(self.gpio))),
             capsules::alarm::DRIVER_NUM => f(Some(Err(self.alarm))),
             capsules::led::DRIVER_NUM => f(Some(Ok(self.led))),
--- a/capsules/src/proximity.rs
+++ b/capsules/src/proximity.rs
@@ -49,9 +49,10 @@
 //! ```
 
 use core::cell::Cell;
+use core::mem;
 use kernel::hil;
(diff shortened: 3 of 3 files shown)


hudson-ayers (approved):
These changes look good.

I did notice while reviewing that `run_next_command()` calls `find_thresholds()` internally which triggers unsoundness per #2135 , but that has nothing to do with the changes here.

lschuermann (approved):
This looks pretty good.

It appears I'm unable to comment on unchanged regions, but the comments regarding `subscribe` in the top of the file seem partially outdated. I don't think such a description is required any more, given that encoding of `allow` and `subscribe` is now centrally done in the kernel.

arjundeopujari:
@hudson-ayers @lschuermann Did one of you get a chance to test this with the aforementioned hardware (nano33ble or clue_nrf52840)?

hudson-ayers:
> @hudson-ayers @lschuermann Did one of you get a chance to test this with the aforementioned hardware (nano33ble or clue_nrf52840)?

I did not

lschuermann:
@arjundeopujari  Unfortunately not, those boards are still on my wishlist.

arjundeopujari:
@hudson-ayers @lschuermann Okay. I will test them quickly after the call today on one of my nano33ble's

arjundeopujari:
@alevy Ok.  I tested with the original driver and the 2.0 driver and the 2.0 driver consistently generates a data access fault.  See below serial output from tockloader.
=======

---| Fault Status |---
Data Access Violation:              true
Faulting Memory Address:            0x00000004
Fault Status Register (CFSR):       0x00000082
Hard Fault Status Register (HFSR):  0x40000000

---| App Status |---
App: proximity   -   [Fault]
 Events Queued: 0   Syscall Count: 6   Dropped Callback Count: 0
 Restart Count: 0
 Last Syscall: Some(ReadOnlyAllow { driver_number: 1, subdriver_number: 2, allow_address: 0x20004800, allow_size: 0 })

========
For more info on CSFR and HFSR reg-fields: https://interrupt.memfault.com/blog/cortex-m-fault-debug#cfsr

Steps to recreate: 
Flash the app using `bossac -i -e -o 0x40000 -w build/cortex-m4/cortex-m4.tbf -R` from `/examples/tests/proximity` directory on a nano33ble sense board. 

Comments:  It consistently seems to consistently fault at the same address.  Not sure if this could be caused by @hudson-ayers 's claim on the Grant API unsoundness?  Original one still works perfectly well.  Not sure what is going on here.

arjundeopujari:
@bradjc Maybe you can double-check this since it seems it should work and you have the board with you.

phil-levis:
I have seen a data access fault as well. I tracked mine down to `__sfp`, which is part of finding open FILE structures for `fopen`, which should not be invoked (it should not be in the binary! newlib is killing me). So something is being corrupted. My best guess is  something is wrong with memop. I'm going to try to reproduce the error and track down its cause.

hudson-ayers:
I saw a data access fault on the UDP example as well once I rebased libtock-c and updated to the new system calls. Haven't had time to dig into it yet

phil-levis:
I think I found it. crt0.c has inline assembly for invoking `memop` and it was still using 1.x syscall invocations. I pushed the fix to tock-2.0-dev.

hudson-ayers:
Pulling in that change fixed my bug as well, thanks!

PR #2288, opened 2020-12-18
hil: Sound Pressure

alexandruradovici:
### Pull Request Overview

This pull request adds a Sound Pressure HIL and driver. This uses an ADC microphone.

Additionally, the pull request:
  - adds microphone support for Microbit v2
  - adds a `adc_microphone` capsule for sampling ADC microphones.
  - fixes an issue with the `virtual_adc` capsule that prevents requesting another sample from' withing the `sample_ready`
  - adds the ability to setup the nrf52 adc channels individually by setting the gain, bias and sampling time
  - ~adds the [micromath](https://github.com/NeoBirth/micromath) library to Tock (it is Apache 2.0 and MIT). This is used here for computing the sound level in dB (uses `log10`)~
  - adds a panic and kernel debug LED (uses the microphone LED)
  - adds drive type public function to nrf52 gpio

I opened this as a separate pull request as it adds a new library and might require some more reviews.

### Testing Strategy

This pull request was tested using a Microbit v2 board.

### TODO or Help Wanted

Requires merge of #2255 

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/adc_microphone.rs
+++ b/capsules/src/adc_microphone.rs
@@ -99,7 +99,8 @@ impl<'a, P: gpio::Pin> adc::Client for AdcMicrophone<'a, P> {
             }) {
                 // self.enable_pin.map (|pin| pin.clear ());
                 let spl = self.compute_spl();
-                self.spl_client.map(|client| client.callback(spl));
+                self.spl_client
+                    .map(|client| client.callback(ReturnCode::SUCCESS, spl));
             }
         }
     }
--- a/capsules/src/adc_microphone.rs
+++ b/capsules/src/adc_microphone.rs
@@ -1,71 +1,10 @@
 use core::cell::Cell;
-use core::f32;
 use kernel::common::cells::{OptionalCell, TakeCell};
+use kernel::common::math;
 use kernel::hil::adc;
 use kernel::hil::gpio;
 use kernel::hil::sensors::{SoundPressure, SoundPressureClient};
 use kernel::ReturnCode;
-
-// f32 log10 function adapted from [micromath](https://github.com/NeoBirth/micromath)
-const EXPONENT_MASK: u32 = 0b01111111_10000000_00000000_00000000;
-const EXPONENT_BIAS: u32 = 127;
-
-fn abs(n: f32) -> f32 {
-    f32::from_bits(n.to_bits() & 0x7FFF_FFFF)
-}
-
-fn extract_exponent_bits(x: f32) -> u32 {
-    (x.to_bits() & EXPONENT_MASK).overflowing_shr(23).0
-}
-
-fn extract_exponent_value(x: f32) -> i32 {
-    (extract_exponent_bits(x) as i32) - EXPONENT_BIAS as i32
-}
-
-fn ln_1to2_series_approximation(x: f32) -> f32 {
-    // idea from https://stackoverflow.com/a/44232045/
-    // modified to not be restricted to int range and only values of x above 1.0.
-    // and got rid of most of the slow conversions,
-    // should work for all positive values of x.
-
-    //x may essentially be 1.0 but, as clippy notes, these kinds of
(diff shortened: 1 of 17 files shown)


alexandruradovici:
Rebased it.

bradjc:
The new HIL and capsule looks good, but I'm not sure about adding an entire library for just the one `log10` function.

alexandruradovici:
The library's license is Apache 2.0 and MIT, I think it should be ok to just strip out that function and specify it's copyright.

On the other hand, math functions might be useful for other capsules too. If not used, I don't think it increases the kernel's size.

hudson-ayers:
Even if it doesn't increase the kernel's size when not used, there is some maintenance burden as we might eventually update to newer rustc versions that could surface warnings or errors in the vendored code.

Either way I think that vendoring in external libraries should probably happen in standalone PRs.

alexandruradovici:
> Even if it doesn't increase the kernel's size when not used, there is some maintenance burden as we might eventually update to newer rustc versions that could surface warnings or errors in the vendored code.
> 
Agree.

> Either way I think that vendoring in external libraries should probably happen in standalone PRs.
The solution would then be to copy in place the requested function and write in a comment the copyright. Would that work?

alexandruradovici:
I deleted the `micromath` library and added the log10 function to the `sound_pressure` driver.

capsules/src/adc_microphone.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+}
+
+fn log10_ln_approx(x: f32) -> f32 {
+    //using change of base log10(x) = ln(x)/ln(10)
+    let ln10_recip = f32::consts::LOG10_E;
+    let fract_base_ln = ln10_recip;
+    let value_ln = ln_1to2_series_approximation(x);
+    value_ln * fract_base_ln
+}
```
hudson-ayers, line 68: Thanks for extracting out the used function for now, but could this go in `kernel/src/common/math.rs`? There are other similar log functions there.

alexandruradovici: Done

bradjc (approved):

hudson-ayers (approved):

kernel/src/hil/sensors.rs
```diff
@@ -112,3 +112,26 @@ pub trait NineDofClient {
     /// over the syscall interface to an application.
     fn callback(&self, arg1: usize, arg2: usize, arg3: usize);
 }
+
+/// Basic Interface for Sound Pressure
+pub trait SoundPressure<'a> {
```
phil-levis, line 117: Please add comments that explain the state machine. For example, what happens if you call `disable()` then `read_sound_pressure()`? What happens if you call `read_sound_pressure()` then `disable()`? Specify what the valid ReturnCode values are for each call and what they mean. 

The callback needs to include return a ReturnCode. Any API where there is a callback can have a situation where the initial request succeeds, but acquisition fails due to a later error. E.g., if this API is virtualized, it can be that issuing a read just puts the request on a queue (or waits for access to a bus), but then when it's pulled off the queue the request fails.

phil-levis (dismissed):
Detailed comment is inline. The callback needs to provide a ReturnCode for deferred failures.

kernel/src/hil/sensors.rs
alexandruradovici: > The callback needs to include return a ReturnCode. Any API where there is a callback can have a situation where the initial request succeeds, but acquisition fails due to a later error. E.g., if this API is virtualized, it can be that issuing a read just puts the request on a queue (or waits for access to a bus), but then when it's pulled off the queue the request fails.

I agree, but Temperature, Humidity and others do not have one. I will add it here, should I add a PR with the ReturnCode for the others?

hudson-ayers (approved):

bradjc (approved):

bradjc:
@phil-levis

PR #2299, opened 2020-12-23
Add CSR support for 64 Bit RISC-V

bradjc:
### Pull Request Overview

Adaptation of #2041.

This pull request takes only the CSR updates from #2041 so that the CSR code can be compiled for both 32 bit and 64 bit RISC-V. This is mainly just converting `u32` -> `usize`, and then adding arch `cfg`s where the 32 bit and 64 bit architectures don't quite match.

I believe this is the non-controversial changes from #2041, and is a step towards supporting a 64 bit RISC-V platform. The other changes needed to fully support rv64 involve changing the low-level assembly since there is no "register size load/store" in RISC-V, and updating the PMP code. I think we need to come up with some strategy to support full rv64 without hopefully duplicating a lot of very delicate code.

Also, yes, it pains me to add so many `cfg` statements. However, they are all based on the architecture as specified by the RISC-V spec, and not part of the software architecture in Tock. So they should be reasonably well documented, and any confusion can be resolved by referring to the RISC-V spec.

### Testing Strategy

I tried the arty-e21 board and it still works.

### TODO or Help Wanted

n/a

### Documentation Updated

I did a grep in the docs and I didn't see anywhere we talked about CSR sizes.

### Formatting

- [x] Ran `make prepush`.

Sidenote: enabling automatic rustfmt in your text editor is absolutely amazing and you should absolutely do it.


hudson-ayers:
Also, renaming lots of files can be a challenge when maintaining 2 separate branches -- it makes it easy for commits on one branch to not make it into the final merged product. Before this is merged, we should probably merge `master` into `tock-2.0-dev`, then merge this PR into `master`, then also apply the changes in this PR directly to `tock-2.0-dev`

bradjc:
Removed the register macro change commit. Still waiting on the register `usize` PR.

alistair23 (approved):

lschuermann:
> Before this is merged, we should probably merge `master` into `tock-2.0-dev`, then merge this PR into `master`, then also apply the changes in this PR directly to `tock-2.0-dev`

Thanks for looking out, this seems like a good strategy. I've merged `master` into `tock-2.0-dev`, as that was due anyways. Once this is merged, I can get the changes into `tock-2.0-dev`, carefully watching out for any changes to those respective files.

hudson-ayers (approved):
I didn't check every single change, there is a lot. But looking at a subset it seems this is pretty unlikely to break anything on 32 bit, and we can catch anything that doesn't work on 64 bit once we actually have a platform to test on.

bors r+

PR #2303, opened 2020-12-23
llvm_asm -> asm: arty-e21

bradjc:
### Pull Request Overview

Convert to asm macro.

#1923

### Testing Strategy

Running apps on arty-e21.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/support.rs
+++ b/arch/rv32i/src/support.rs
@@ -8,7 +8,7 @@ use core::ops::FnOnce;
 /// NOP instruction
 pub fn nop() {
     unsafe {
-        asm!("nop", options(nomem, nostack));
+        asm!("nop", options(nomem, nostack, preserves_flags));
     }
 }
 
--- a/arch/rv32i/src/support.rs
+++ b/arch/rv32i/src/support.rs
@@ -8,7 +8,7 @@ use core::ops::FnOnce;
 /// NOP instruction
 pub fn nop() {
     unsafe {
-        llvm_asm!("nop" :::: "volatile");
+        asm!("nop", options(nomem, nostack));
     }
 }
 
@@ -16,7 +16,7 @@ pub fn nop() {
 #[inline(always)]
 /// WFI instruction
 pub unsafe fn wfi() {
-    llvm_asm!("wfi" :::: "volatile");
+    asm!("wfi", options(nomem, nostack));
 }
 
 pub unsafe fn atomic<F, R>(f: F) -> R
--- a/chips/arty_e21_chip/src/chip.rs
+++ b/chips/arty_e21_chip/src/chip.rs
@@ -16,6 +16,7 @@ pub struct ArtyExx<'a, I: InterruptService<()> + 'a> {
     pmp: pmp::PMP,
     userspace_kernel_boundary: rv32i::syscall::SysCall,
     clic: rv32i::clic::Clic,
+    machinetimer: &'a rv32i::machine_timer::MachineTimer<'a>,
     interrupt_service: &'a I,
 }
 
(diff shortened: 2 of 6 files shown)


hudson-ayers (commented):
Nice to get rid of  `cfg[]` + assembly in a couple places!

arch/rv32i/src/support.rs
```diff
@@ -8,15 +8,15 @@ use core::ops::FnOnce;
...
+        asm!("nop", options(nomem, nostack));
     }
 }
 
 #[cfg(all(target_arch = "riscv32", target_os = "none"))]
 #[inline(always)]
 /// WFI instruction
 pub unsafe fn wfi() {
-    llvm_asm!("wfi" :::: "volatile");
+    asm!("wfi", options(nomem, nostack));
```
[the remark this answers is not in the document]
lschuermann: I don't think so. When we're catching an interrupt in kernel mode and calling the trap handler, which do push some caller-saved registers to the stack. Furthermore, the actual interrupt handler itself might further use the stack. However, prior to continuing execution at the previous location (saved in `mepc`), we're restoring the stack to its original state. So from the point of view of this instruction, the stack won't be modified.

Furthermore, an interrupt can come in at virtual any instruction. A `wfi` is just an explicit way to allow the processor to go into a low-power mode until an interrupt occurs and thereby doesn't behave differently with respect to side-effects compared to a `nop` (which, if I'm not mistaken, is actually a legal implementation of `wfi`).

chips/arty_e21_chip/src/chip.rs
```diff
@@ -134,11 +121,8 @@ impl<'a, I: InterruptService<()> + 'a> ArtyExx<'a, I> {
             addi t0, t0, %lo(_start_trap)
```
lschuermann, line 121: I think a `li t0, _start_trap`, which will expand to a `lui` and `addi`, is a bit nicer.

chips/arty_e21_chip/src/chip.rs
```diff
@@ -134,11 +121,8 @@ impl<'a, I: InterruptService<()> + 'a> ArtyExx<'a, I> {
...
             ori  t0, t0, 0x02 // Set CLIC direct mode
             csrw 0x305, t0    // Write the mtvec CSR.
-        "
-        :
-        :
-        :
-        : "volatile");
+            "
```
lschuermann, line 124: This is more a general question: shouldn't we mark the register `t0` as clobbered here (as described [here](https://doc.rust-lang.org/beta/unstable-book/library-features/asm.html#clobbered-registers))?

Given that this entire function is a large assembly block, it's probably fine, but I imagine the compiler could potentially do some optimizations or perform inlining which could break the assumptions made about registers not being clobbered in the future?

arch/rv32i/src/support.rs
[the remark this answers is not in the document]
hudson-ayers: Hmm, I think you are right. For cortex-m we did not mark `wfi` as `nostack` -- but we did mark it as `preserves_flags`. We should probably make them match. -- I think that both `nostack` and `preserves_flags` both seem right to include.

chips/arty_e21_chip/src/chip.rs
hudson-ayers: I agree, t0 should be marked as clobbered.

hudson-ayers (approved):

lschuermann (approved):

alistair23 (approved):

PR #2307, opened 2020-12-24
Updating Alarm syscall driver to 2.0 syscall API

phil-levis:
### Pull Request Overview

This pull requests updates the alarm system call driver to use the 2.0 syscall API.

### Testing Strategy

This pull request was tested by running two copies of multi_alarm_test in parallel on the imix platform. I found a few bugs while doing so, which led me to update libtock-c's alarm code to use the new (reference, dt) system calls. It had still be using just an absolute time, which if there are long delays due to printfs can cause the hang-on-alarm-in-the-past bug.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/alarm.rs
+++ b/capsules/src/alarm.rs
@@ -3,10 +3,10 @@
 
 use core::cell::Cell;
 use core::mem;
-use kernel::hil::time::{self, Alarm, Frequency, Ticks, Ticks32};
-use kernel::{AppId, Callback, CommandResult, Driver, ErrorCode, Grant};
-use kernel::{ReturnCode};
 use kernel::debug;
+use kernel::hil::time::{self, Alarm, Frequency, Ticks, Ticks32};
+use kernel::ReturnCode;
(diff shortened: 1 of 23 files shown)


capsules/src/alarm.rs
```diff
@@ -227,6 +242,7 @@ impl<'a, A: Alarm<'a>> LegacyDriver for AlarmDriver<'a, A> {
                         let future_time = data;
                         let dt = future_time.wrapping_sub(reference);
                         // if previously unarmed, but now will become armed
+                        debug!("Alarm.rs: setting to {} + {}", reference, dt);
```
hudson-ayers, line 245:
```suggestion
```

hudson-ayers (approved):
Looks good except for the leftover debug statement

capsules/src/alarm.rs
arjundeopujari: Yes. Looks good.

PR #2310, opened 2020-12-28
update rust to nightly-2020-12-28

hudson-ayers:
### Pull Request Overview

This pull request updates the rust nightly version to 2020-12-28. This was motivated by the stabilization of `min_const_generics` ( https://github.com/rust-lang/rust/pull/79135 ) -- by using this nightly, we will be able to use const generics in Tock without adding an unstable feature.

### Testing Strategy

compiling

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers:
It looks like a lot of what we are doing with inline assembly in naked functions now generates warnings. I didn't realize, I'm gonna mark this as draft until I have fixes.

One particularly thorny issue is that it is illegal to use anything other than a single `asm!()` block in any naked function. But we pretty extensively mix rust code and insline asm in several functions like `hard_fault_handler_arm_v7m()` which are marked `#[naked]`.  The reason this generates a warning, I believe, is because rust code relies on the compiler generated prelude in order to correctly use stack-allocated variables, but `#[naked]` prevents generation of that prelude.

hudson-ayers:
#2317 is the first step towards merging this.

hudson-ayers:
closed in favor of #2334

PR #2311, opened 2020-12-28
Board based instantiation of chip drivers and interrupt mappings: imxrt10xx

mciantyre:
### Pull Request Overview

The PR refactors the `imxrt10xx` chip and boards to support board-based initialization (first proposed in #2069). It builds on the work of #1918 and #2200, where we noted the effort as a TODO. After this PR, you may instantiate and configure i.MX RT peripherals in a board, or use the default peripherals provided by the chip.

Most peripherals were easily transitioned to the new API. The exception was the GPIO driver, which referenced `static` GPIO ports and pins throughout the code. I may have found some issues in the driver, so I took the refactoring opportunity to update the implementation, trying to follow [this suggestion](https://github.com/tock/tock/blob/master/chips/stm32f4xx/src/gpio.rs#L595). I'll leave more details in review comments.

Other changes include

- fix a GPT clock gating bug introduced in #2200
- move UART root clock initialization out of the `configure()` method, and into CCM peripheral setup

### Testing Strategy

Tested `boards/teensy40` on a Teensy 4.0 using the tests from #2200.

`boards/imxrt1050-evkb` continues to compile, but I don't have the hardware to test examples.

### TODO or Help Wanted

Looking for feedback on the refactor. Let me know if it's deviating too much from the other chips.

The new GPIO driver relies on the `min_const_generics` feature. From my understanding, this feature will be stabilized in 1.51. If we don't want to adopt that feature right now, I'm happy to reconsider.

Holding as draft to

- [ ] figure out a way to remove the remaining `static` peripheral in `iomuxc_snvs`
- [ ] remove the `std` dependency in some new unit tests, avoiding the conditional `no_std` in the crate
- [ ] leave thoughts on GPIO changes
- [ ] clean up commit messages

### Documentation Updated

- [x] ~~Updated the relevant files in `/docs`, or~~ no updates are required.

### Formatting

- [ ] Ran `make prepush`.


chips/imxrt10xx/src/gpio.rs
```diff
@@ -436,193 +91,213 @@ enum_from_primitive! {
...
-        registers: GPIO3_BASE,
-        clock: PortClock(ccm::PeripheralClock::CCGR1(ccm::HCLK1::GPIO1)),
-    },
-    Port {
-        registers: GPIO4_BASE,
-        clock: PortClock(ccm::PeripheralClock::CCGR1(ccm::HCLK1::GPIO1)),
-    },
-    Port {
-        registers: GPIO5_BASE,
-        clock: PortClock(ccm::PeripheralClock::CCGR1(ccm::HCLK1::GPIO1)),
```
mciantyre, line 621: If reverting to this change, update the `PeripheralClock`s. Before this PR, GPIO2 through 5 would control GPIO1's clock gate.

chips/imxrt10xx/src/gpio.rs
```diff
@@ -732,1230 +535,133 @@ impl ClockInterface for PortClock {
...
+        };
+        self.registers.gdir.set(gdir);
     }
 
     fn set_output_high(&self) {
-        let port = self.pinid.get_port();
-
-        match self.pinid.get_pin_number() {
-            0b000000 => {
-                port.registers.dr.write(DR::DR0::SET);
```
mciantyre, line 1241: From my understanding, a `write` usage will set all non-specified fields to zero. These calls on `dr` may need to be a modify, otherwise setting one pin high will set all other pins low.

Alternatively, write a high bit `dr_set`, which will set the pin high.

hudson-ayers (dismissed):
This looks good to me, thanks for taking this on!

I think that #2310 may take awhile, as we need to rewrite a decent chunk of assembly code to get rid of warnings. But I think that we are ok accepting PRs with `min_const_generics` now that it is stabilized, and we can just remove the feature flag once #2310 is merged -- so feel free to mark this as ready for review if you do not have anything else you plan to change.

hudson-ayers (dismissed):

hudson-ayers (approved):

bradjc (approved):

PR #2315, opened 2020-12-30
Tock 2.0: update multi touch

alexandruradovici:
### Pull Request Overview

This pull request fixes:
 - a multi touch report bug for the touch driver by adding the ack command and updating the touch callbacks
 - the ft6x06 driver to report the touch id

### Testing Strategy

This pull request was tested using an STM32F412G Discovery board.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/ft6x06.rs
+++ b/capsules/src/ft6x06.rs
@@ -139,6 +139,7 @@ impl<'a> i2c::I2CClient for Ft6x06<'a> {
                     let status = match buffer[touch_event * 8 + 2] >> 6 {
                         0x00 => TouchStatus::Pressed,
                         0x01 => TouchStatus::Released,
+                        0x02 => TouchStatus::Moved,
                         _ => TouchStatus::Released,
                     };
                     let x = (((buffer[touch_event * 8 + 2] & 0x0F) as u16) << 8)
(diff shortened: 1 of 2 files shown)


capsules/src/ft6x06.rs
```diff
@@ -147,12 +148,13 @@ impl<'a> i2c::I2CClient for Ft6x06<'a> {
                         + (buffer[touch_event * 8 + 5] as u16);
                     let pressure = Some(buffer[touch_event * 8 + 6] as u16);
```
phil-levis, line 149: Ugh these embedded constants! I don't want to block a 2.0 update on fixing cruft like this, but fixing it seems pretty important from a maintenance perspective.

phil-levis (approved):
This driver definitely needs its data parsing code cleaned up -- hard coded constants should be replaced by structures, etc.

capsules/src/ft6x06.rs
alexandruradovici: I agree, but I'm not sure how to fix this. I can either:
- declare some consts instead of the hard coded numbers.
- have a struct instead of a buffer, but I'm not sure I can convert the struct to the array with some kind of `memcpy` that would be considered safe

Any suggestion is welcome.

PR #2317, opened 2020-12-30
ARM Cortex-m hardfault handler: remove UB and fix warnings on newer nightlies

hudson-ayers:
### Pull Request Overview

This pull request modifies the cortex-m hardfault handler to contain only a single, no-return asm!() block, rather than mixing rust code and inline assembly, which is undefined behavior in `#[naked]` functions (see: https://internals.rust-lang.org/t/idea-naked-functions-2-0/11637/25 ).
On recent nightlies, because of https://github.com/rust-lang/rust/pull/79653, rustc now warns that our mixing inline assembly and rust code, as well as our use of `llvm_asm` is deprecated and will be a hard error in future versions of rust. In order to update to newer versions of the rust compiler, we will have to fix this.

The primary trick here is that the only reason (I think) the hard fault handler needed to be a naked function is so that the stack pointer can be read unmodified, and so that calling the handler does not push to the stack (which could lead to an immediate double fault when the fault was a stack overflow). Therefore the only asm that needs to be in a naked function is the assembly code that:
- stores the appropriate stack pointer
- determines if a stack overflow occurred
- updates the stack pointer in the case of a stack overflow

The rest of the hard fault handler can just be placed in a non-naked function, which is called from the inline assembly in the naked function. This involves placing the parameters in the appropriate registers and then branching to the correct function. This allows for the convenience of mixing rust code and assembly in the rest of the hard fault handler.

This requires changing `_estack` to be a `static` rather than a `fn()` type, but I think it was only a function type because it is passed as a function in other files (`chips/sam4l/src/lib.rs`), which I think were copied to this file. This change compiles fine and better represents the actual type of `_estack`.

### Testing Strategy

This pull request was tested by:
-  manually inserting a stack overflow in the kernel, and observing that "kernel stack overflow" is printed along with the panic message. 
- manually inserting a read to an invalid memory address, and observing a normal panic message is printed.

### TODO or Help Wanted

I do not fully understand the ARM calling convention -- do I need to be validating the stack pointer is double-word aligned before calling the non-naked function?

Also, is my understanding correct that noreturn inline asm does not need to mark its clobbers? It seems that the newer compiler tries to prevent me from doing so.

This PR does not fix all warnings surfaced in newer nightlies -- just the warnings surfaced by the hardfault handler.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/lib.rs
+++ b/arch/cortex-m/src/lib.rs
@@ -392,7 +392,10 @@ unsafe fn kernel_hardfault_arm_v7m(faulting_stack: *mut u32) -> ! {
 }
 
 #[cfg(all(target_arch = "arm", target_os = "none"))]
-unsafe extern "C" fn hard_fault_handler_arm_v7m_non_naked(
+/// Continue the hardfault handler. This function is not `#[naked]`, meaning we can mix
+/// `asm!()` and Rust. We separate this logic to not have to write the entire fault
+/// handler entirely in assembly.
+unsafe extern "C" fn hard_fault_handler_arm_v7m_continued(
     faulting_stack: *mut u32,
     kernel_stack: u32,
     stack_overflow: u32,
@@ -470,9 +473,9 @@ pub unsafe extern "C" fn hard_fault_handler_arm_v7m() {
         "ite   ne               /* check if the result of that bitwise AND was not 0 */",
         "movne r2, #1           /* BFSR & 0b00110000 != 0; r2 = 1 */",
         "moveq r2, #0           /* BFSR & 0b00110000 == 0; r2 = 0 */",
-        "and r5, r1, r2 /* bitwise and r2 and r1, store in r5 */ ",
-        "cmp  r5, #1 /*  update condition codes to reflect if r2 == 1 && r1 == 1 */",
-        "itt  eq /* if r5==1 run the next 2 instructions, else skip to branch */",
+        "and r5, r1, r2         /* bitwise and r2 and r1, store in r5 */ ",
+        "cmp  r5, #1            /*  update condition codes to reflect if r2 == 1 && r1 == 1 */",
+        "itt  eq                /* if r5==1 run the next 2 instructions, else skip to branch */",
         // if true, The hardware couldn't use the stack, so we have no saved data and
         // we cannot use the kernel stack as is. We just want to report that
         // the kernel's stack overflowed, since that is essential for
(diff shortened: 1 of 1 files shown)


arjundeopujari:
> I do not fully understand the ARM calling convention -- do I need to be validating the stack pointer is double-word aligned?
For cortex-m0, the SP is automatically aligned to 8-byte addresses when executing exceptions/faults. You can also have this automatically done on other cortex m processors by setting STKALIGN in the CCR register.  This is recommended by ARM itself for binary compatibility among processors.  I am not sure if Rust compiler assumes a 8 byte aligned SP like C/C++ does however.  Did I answer your question? @hudson-ayers 

Sources: https://developer.arm.com/documentation/ddi0432/c/programmers-model/instruction-set-summary/binary-compatibility-with-other-cortex-processors
https://community.arm.com/developer/ip-products/processors/f/cortex-m-forum/6344/what-is-the-meaning-of-a-64-bit-aligned-stack-pointer-address

bradjc (approved):

hudson-ayers:
> For cortex-m0, the SP is automatically aligned to 8-byte addresses when executing exceptions/faults. You can also have this automatically done on other cortex m processors by setting STKALIGN in the CCR register. This is recommended by ARM itself for binary compatibility among processors. I am not sure if Rust compiler assumes a 8 byte aligned SP like C/C++ does however. Did I answer your question? @hudson-ayers
> 
> Sources: https://developer.arm.com/documentation/ddi0432/c/programmers-model/instruction-set-summary/binary-compatibility-with-other-cortex-processors
> https://community.arm.com/developer/ip-products/processors/f/cortex-m-forum/6344/what-is-the-meaning-of-a-64-bit-aligned-stack-pointer-address

Thanks for finding these links -- it seems that STKALIGN defaults to 1 on cortex-m3, cortex-m4, and cortex-m7, so I think this should be fine

bradjc (approved):

alistair23:
This PR (specifically commit fa691268679fdc880e68558555c7eb595a3e0e31) breaks Tock on the nRF52840dk. Tock prints `fix hard fault handler assembly to not use stack in naked_fn` but then resets so the board is stuck in an infinite reboot loop.

hudson-ayers:
Can you explain how you surface this breakage? 

If I flash current master (just the kernel) it works. Same if I add the blink app.
If I then flash a faulting app, I get the correct panic output for that faulted app.
If I add a kernel stack overflow, I get a panic output telling me there was a kernel stack overflow.

@alistair23

alistair23:
It seems to happen with a fault. When I get a hard fault the board resets instead of printing debug information.

alistair23:
This diff on master (partial revert) fixes the problem

```diff
diff --git a/arch/cortex-m/src/lib.rs b/arch/cortex-m/src/lib.rs
index e46627eb0..08dd5652c 100644
--- a/arch/cortex-m/src/lib.rs
+++ b/arch/cortex-m/src/lib.rs
@@ -403,7 +403,7 @@ unsafe extern "C" fn hard_fault_handler_arm_v7m_continued(
     faulting_stack: *mut u32,
     kernel_stack: u32,
     stack_overflow: u32,
-) -> ! {
+) {
     if kernel_stack != 0 {
         if stack_overflow != 0 {
             // Panic to show the correct error.
@@ -415,8 +415,8 @@ unsafe extern "C" fn hard_fault_handler_arm_v7m_continued(
     } else {
         // Hard fault occurred in an app, not the kernel. The app should be
         // marked as in an error state and handled by the kernel.
-        asm!(
-            "
+        llvm_asm!(
+        "
         /* Read the relevant SCB registers. */
         ldr r0, =SCB_REGISTERS  /* Global variable address */
         ldr r1, =0xE000ED14     /* SCB CCR register address */
@@ -443,9 +443,8 @@ unsafe extern "C" fn hard_fault_handler_arm_v7m_continued(
         isb
 
         movw LR, #0xFFF9
-        movt LR, #0xFFFF",
-            options(noreturn)
-        );
+        movt LR, #0xFFFF"
+        : : : "r1", "r0", "r2", "memory" : "volatile" );
     }
 }
```

hudson-ayers:
Thanks for finding that, I believe that #2350 is an appropriate fix.

PR #2324, opened 2021-01-05
LED: remove need for `&mut self`

bradjc:
### Pull Request Overview

When adding an LED to the bootloader I noticed the LED trait requires mutable references. Requiring `mut` requires changing a lot of other code to `mut`, and this is not consistent with other Tock code.

This was a quick PR to implement, but I'm not sure why `Led` used mutable references. 

### Testing Strategy

travis

### TODO or Help Wanted

If there is a need for the mutable self references we can close this.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/led.rs
+++ b/capsules/src/led.rs
@@ -61,11 +61,11 @@ pub const DRIVER_NUM: usize = driver::NUM::Led as usize;
 /// Holds the array of LEDs and implements a `Driver` interface to
 /// control them.
 pub struct LedDriver<'a, L: led::Led> {
-    leds: TakeCell<'a, [&'a mut L]>,
+    leds: TakeCell<'a, [&'a L]>,
 }
 
 impl<'a, L: led::Led> LedDriver<'a, L> {
-    pub fn new(leds: &'a mut [&'a mut L]) -> Self {
+    pub fn new(leds: &'a mut [&'a L]) -> Self {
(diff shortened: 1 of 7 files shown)


lschuermann (approved):
I was actually pleasantly surprised to see `Led` requiring a mutable reference to modify its state, but as you state that does not fit nicely within the rest of the Tock ecosystem.

capsules/src/led.rs
```diff
@@ -61,11 +61,11 @@ pub const DRIVER_NUM: usize = driver::NUM::Led as usize;
...
 pub struct LedDriver<'a, L: led::Led> {
-    leds: TakeCell<'a, [&'a mut L]>,
+    leds: TakeCell<'a, [&'a L]>,
 }
 
 impl<'a, L: led::Led> LedDriver<'a, L> {
-    pub fn new(leds: &'a mut [&'a mut L]) -> Self {
+    pub fn new(leds: &'a mut [&'a L]) -> Self {
         // Initialize all LEDs and turn them off
         for led in leds.iter_mut() {
```
hudson-ayers, line 70:
```suggestion
        for led in leds.iter() {
```

hudson-ayers (approved):

alistair23 (approved):

alevy (approved):

PR #2327, opened 2021-01-08
fix bug in `_start_trap()` assembly

hudson-ayers:
### Pull Request Overview

@vsukhoml found a bug in #2073 after it was merged, this PR fixes it. That PR added code to save `mtval` to the stored state struct, but overwrote the value in t0 (the saved `mcause` value) in the process. `t0` is used after that under the assumption it still holds `mcause`. This PR just moves the code to save `mtval` above the code that saves `mcause`.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

- [x] N/A

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/lib.rs
+++ b/arch/rv32i/src/lib.rs
@@ -337,10 +337,10 @@ pub extern "C" fn _start_trap() {
             sw   t0, 1*4(s0)  // Save the app sp to the stored state struct
             csrr t0, 0x341    // CSR=0x341=mepc
             sw   t0, 31*4(s0) // Save the PC to the stored state struct
-            csrr t0, 0x342    // CSR=0x342=mcause
-            sw   t0, 32*4(s0) // Save mcause to the stored state struct
             csrr t0, 0x343    // CSR=0x343=mtval
             sw   t0, 33*4(s0) // Save mtval to the stored state struct
+            csrr t0, 0x342    // CSR=0x342=mcause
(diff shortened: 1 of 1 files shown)


arch/rv32i/src/lib.rs
```diff
@@ -337,10 +337,10 @@ pub extern "C" fn _start_trap() {
             sw   t0, 1*4(s0)  // Save the app sp to the stored state struct
             csrr t0, 0x341    // CSR=0x341=mepc
             sw   t0, 31*4(s0) // Save the PC to the stored state struct
-            csrr t0, 0x342    // CSR=0x342=mcause
-            sw   t0, 32*4(s0) // Save mcause to the stored state struct
             csrr t0, 0x343    // CSR=0x343=mtval
             sw   t0, 33*4(s0) // Save mtval to the stored state struct
+            csrr t0, 0x342    // CSR=0x342=mcause
```
lschuermann, line 342: Good catch @vsukhoml! I think it might be worth adding an additional comment here which indicates why exactly we're following this order. I'd argue in such a section of mechanical CSR saves this is really easy to miss and likely why the issue was introduced in the first place.

```suggestion

            // Save mcause last, as we depend on it being loaded in t0 below
            csrr t0, 0x342    // CSR=0x342=mcause
```

lschuermann (approved):

bradjc (approved):
Assembly is hard :)

PR #2328, opened 2021-01-08
Core notes from 2020-12-18

brghena:
### Pull Request Overview

Notes from our call December 18th. I went through the audio recording to make sure errorcode and returncode are properly placed, I hope.

### Testing Strategy

Reading and listening.

### TODO or Help Wanted

None

### Documentation Updated

- [X ] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [N/A] Ran `make prepush`.

### DIFF
--- a/doc/wg/core/notes/core-notes-2020-12-18.md
+++ b/doc/wg/core/notes/core-notes-2020-12-18.md
@@ -1,0 +1,133 @@
+# Tock Core Notes 2020-12-18
+
+## Attending
+ * Branden Ghena
+ * Amit Levy
+ * Johnathan Van Why
+ * Alistair
+ * Leon Schuermann
+ * Brad Campbell
+ * Arjun
+ * Philip Levis
+ * Hudson Ayers
+ * Pat Pannuto
+ * Vadim Sukhomlinov
+
+## Updates
+### Libtock runtime assembly
+ * Johnathan: Working on libtock runtime, the new runtime for libtock-rs. Trying to get it working with the stable asm! feature. Requires standalone assembly for entry point. Might end up having to rely on another unstable nightly feature, but we'll see. You need either naked functions or global ASM, or external assembly to make this work. But toolchain is a pain for external assembly.
+ * Phil: I'm a fan of external assembly
+ * Amit: Does LLVM not have an assembler?
+ * Johnathan: I'm not sure, more importantly I'm not sure if it comes with the Rust toolchain.
+ * Johnathan: We might also lose out on inline assembly optimizations with external assembly. It seems like inline is on the track towards stabilization.
+ * Amit: This is the low-level runtime, right? (yes)
+ * Hudson: Rust core team has stated they hope to have inline assembly stable for 2021 rust release.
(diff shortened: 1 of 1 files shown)


lschuermann (approved):
Looks good. Some of my comments on the `ReturnCode`/`ErrorCode` discussion have been quite confusing but this is a great summary. :smile:

doc/wg/core/notes/core-notes-2020-12-18.md
```diff
@@ -0,0 +1,133 @@
...
+ * Johnathan: For our use case, naked functions are the best option. 
+ * Amit: It's possible that it's not an issue for other embedded projects, much like it isn't for the Tock kernel, because the entry point follows C calling conventions. Guessing here. So it could be a userspace problem.
+ * Johnathan: Arguably we could make entry follow the C calling conventions, which would solve this. We are just missing a valid stack pointer, I think.
+ * Amit: Doesn't changing the stack pointer in inline assembly mess with things though?
+ * Johnathan: I think it would be okay. As long as the assembly block never returns.
+ * Amit: So we'd have a trampoline function as the entry point. But we couldn't do what we're doing now, initializing sections of the process in Rust, it would have to be in assembly.
+ * Johnathan: Yeah, it's probably undefined behavior to have it done in Rust. I think if we wanted to move to stable without external assembly, we could push on "nakedfunc", I forget the exact feature name, because there are questions about arguments, but there should be a simple to stabilize version without arguments.
```
hudson-ayers, line 32: @jrvanwhy I didn't really follow this during the call, but now reading it it seems pretty similar to some of the changes in #2317 . I don't see where the entry point initializes sections of the process in Rust, though -- it looks like the naked function just jumps to an `extern 'C'` function containing rust code, which I thought is fine. In fact it looks like the current implementation is fine except that the `llvm_asm!()` block takes the arguments to the function as inputs, rather than accessing them in the assembly code according to the calling convention. Maybe I am missing something though.

Either way, IMO it seems likely that `naked_functions` and `asm` will be stabilized around the same time.

hudson-ayers (approved):

doc/wg/core/notes/core-notes-2020-12-18.md
jrvanwhy: For the Rust userspace runtime, assembly code needs to do the following before executing Rust code:

1. Initialize the stack (i.e. set the stack pointer)
2. Initialize `.data` and `.bss`

(There is currently a soundness issue in `libtock-rs` because the assembly only does #1, not #2.

Do you have a source for `naked_functions` being stabilized around the same time as `asm`? I was under the impression that `naked_functions` is far from stabilization. If my impression is wrong, then I would appreciate that, as I am in the middle of putting together the tooling to supporting using external assembly. My goal is for `libtock-rs` to move to a stable compiler when `asm` is stabilized, but I'm fine with waiting for `naked_functions` if it is on a similar timeline.

hudson-ayers: I thought had read something to that effect, but looking around all I could find was: [this](https://github.com/rust-lang/rust/issues/32408#issuecomment-484726301) statement, which is not that strong. 

So I probably conveyed more confidence than I should have.

PR #2321, opened 2021-01-01
Tock 2.0: update hd44780 driver to 2.0 api

alexandruradovici:
### Pull Request Overview

This pull request updates the hd44780 driver to use the 2.0 API.

NOTE: this driver is somehow deprecated, I think we should merge #2079. I will make the changes to adapt it for 2.0 and submit a new PR with it.

### Testing Strategy

This has been tested using a Nucelo 429ZI board.

### TODO or Help Wanted

 - [x] @teonaseverin please test this PR using https://github.com/tock/libtock-c/pull/152

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make prepush`.

PR #2335, opened 2021-01-08
Tock 2.0: update text_screen driver to 2.0 API

alexandruradovici:
### Pull Request Overview

This pull request updates the text_screen driver to the 2.0 API.

This should replace #2321.

### Testing Strategy

This pull request was tested using a Nucelo STM429ZI board and an HD44780 LCD.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/syscalls/README.md
+++ b/doc/syscalls/README.md
@@ -125,3 +125,12 @@ _Note:_ GPIO is slated for re-numbering in Tock 2.0.
 |   | 0x80002       | PCA9544A         | I2C address multiplexing                   |
 |   | 0x80003       | GPIO Async       | Asynchronous GPIO pins                     |
 |   | 0x80004       | nRF51822         | nRF serialization link to nRF51822 BLE SoC |
+
+### Miscellaneous
+
+|2.0| Driver Number | Driver                                  | Description                                |
+|---|---------------|-----------------------------------------|--------------------------------------------|
+|   | 0x90000       | Buzzer                                  | Buzzer                                     |
+|   | 0x90001       | [Screen](90001_screen.md)               | Graphic Screen                             |
+|   | 0x90002       | [Touch](90002_touch.md)                 | Multi Touch Panel                          |
+|   | 0x90003       | [Text Screen](90003_text_screen.md)     | Text Screen                                |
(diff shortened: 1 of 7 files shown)


kernel/src/hil/text_screen.rs
```diff
@@ -18,7 +18,11 @@ pub trait TextScreen<'a> {
     /// Return values:
     /// - `SUCCESS`: The write command is valid and will be sent to the driver.
     /// - `EBUSY`: The driver is busy with another command.
-    fn print(&self, buffer: &'static mut [u8], len: usize) -> ReturnCode;
+    fn print(
+        &self,
+        buffer: &'static mut [u8],
+        len: usize,
+    ) -> Result<(), (ReturnCode, &'static mut [u8])>;
```
hudson-ayers, line 25: Now that we have `ErrorCode`, if a HIL is going to return a `Result<>`, it makes more sense to return a `Result<(), ErrorCode>` than a `Result<(), ReturnCode>`.

In the long term the goal is to move all error handling to the former, and deprecate `ReturnCode`.

doc/syscalls/README.md
```diff
@@ -125,4 +125,3 @@ _Note:_ GPIO is slated for re-numbering in Tock 2.0.
 |   | 0x80002       | PCA9544A         | I2C address multiplexing                   |
 |   | 0x80003       | GPIO Async       | Asynchronous GPIO pins                     |
 |   | 0x80004       | nRF51822         | nRF serialization link to nRF51822 BLE SoC |
-|   | 0x80005       | [HD44780](80005_hd44780.md)          | LCD HD44780 capsule                        |
```
hudson-ayers, line 128: Can you add the text screen to this doc somewhere?

hudson-ayers (approved):
This looks right to me, thanks! I have one comment on the HIL change.

kernel/src/hil/text_screen.rs
alexandruradovici: If `ReturnCode` will be deprecated, functions HIL should return `Result<_, ErrorCode>`?

hudson-ayers: That is the goal, though we are not enforcing it at the 2.0 switch because it will be *a lot* of refactoring work. But generally it does not make sense that it could be possible to return a `Success` variant in an `Err()`, so `ErrorCode` makes more sense than `ReturnCode` when used in a `Result`

doc/syscalls/README.md
alexandruradovici: I added the documentation for the touch panel and the two screen drivers.

kernel/src/hil/text_screen.rs
alexandruradovici: I'd rather have this merged, as we do need to rewrite in a similar way the screen driver and it is easier to have both of them rewritten at the same time.

hudson-ayers: ok, that's fine with me.

PR #2339, opened 2021-01-11
Apollo3 llvm_asm -> asm

bradjc:
### Pull Request Overview

#1923 

Remove earlgray feature, update apollo3.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/apollo3/src/lib.rs
+++ b/chips/apollo3/src/lib.rs
@@ -2,7 +2,7 @@
 
 #![crate_name = "apollo3"]
 #![crate_type = "rlib"]
-#![feature(llvm_asm, const_fn)]
+#![feature(asm, const_fn)]
 #![no_std]
 
 // Peripherals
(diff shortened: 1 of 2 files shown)


chips/apollo3/src/lib.rs
```diff
@@ -104,7 +104,7 @@ pub unsafe fn init() {
     scb::disable_fpca();
 
     // This ensures the FPU is actually disabled
-    llvm_asm!("svc 0xff" : : : "r0","r1","r2","r3","r12" : "volatile" );
+    asm!("svc 0xff", out("r0") _, out("r1") _, out("r2") _, out("r3") _, out("r12") _);
```
hudson-ayers, line 107:
```suggestion
    asm!("svc 0xff", out("r0") _, out("r1") _, out("r2") _, out("r3") _, out("r12") _, options(preserves_flags));
```

hudson-ayers: Isn't this going to run the svc handler? I am confused why this doesn't cause us to cross the userspace/kernel boundary when this is exactly what we use to cross the boundary in `switch_to_user`.

hudson-ayers: I guess that is really a question for @alistair23

alistair23: Yes, it should call `svc_handler()`.

`svc_handler()` can also trap back to the kernel, which is what happens in this case.

I'm not really sure why we need this here, but I remember that it didn't work otherwise.

alistair23 (approved):

hudson-ayers (approved):

PR #2340, opened 2021-01-11
const generics instead of macro for pmp configuration

hudson-ayers:
### Pull Request Overview

This pull request replaces the use of a giant macro for configuring the risc-v PMP with use of const generics so that the struct is generic over the number of regions available. Unfortunately, one shortcoming of `min_const_generics` (which is now stable) is that arbitrary expressions involving generic constants are not yet supported when those constants are used in the definition of types. Because PMP has some array types of size `NUM_REGIONS`, and some of size `NUM_REGIONS / 2`, both of these values actually have to be separate generic parameters.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -18,22 +18,12 @@
 //! avoid this wasted memory we use TOR and each memory region uses two physical
 //! PMP regions.
 
-/// Instantiate a PMP configuration.
-///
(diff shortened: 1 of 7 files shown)


bradjc (approved):
So cool

alistair23 (approved):

PR #2341, opened 2021-01-11
Use const generics for nrf gpio pins

hudson-ayers:
### Pull Request Overview

This pull request finishes porting the nrf52 chips to the new peripheral model by using const generics to store GPIO ports with different numbers of pins in the same type. It also ports the pwm0 peripheral over to the non-`static mut` approach -- it was missed in the initial PR.

### Testing Strategy

This pull request was tested by running the `buttons` app on an nrf52840dk.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


chips/nrf52840/src/lib.rs
```diff
@@ -1,5 +1,4 @@
 #![no_std]
```
bradjc, line 1:
```suggestion
#![no_std]

```

bradjc (approved):

bradjc (approved):

alistair23 (approved):

hudson-ayers:
fixed merge conflict with sound pressure PR

alevy (approved):

alistair23:
This PR (specifically `port nrf52 gpio to new peripheral model using const generics`) causes a hardfault on the nRF5280DK to enter a continuous reboot loop.

This is the same problem that https://github.com/tock/tock/pull/2350 fixed from a different patch. Reverting d32bd49af49b8dc333e43bcc6184bd5e65f62753 on current master fixes the issue.

hudson-ayers:
that's surprising..I'll look into it. Sorry for burning you twice on the same issue

alistair23:
No worries. It's easy to debug because I was already looking for it :)

hudson-ayers:
On current master, I don't see the issue that #2350 fixed. (i.e. if an app hard faults, I get the full panic printout, and if the kernel faults, I get a full panic printout). Tested on an nrf52840dk. Wondering if you have additional changes that could be interacting with these to cause that?

alistair23:
I don't have any local changes, but I am having trouble reproducing this. It seems to not happen all of the time. I'll keep an eye out and see if I can get a better reproduction case.

hudson-ayers:
I wonder if the bug fixed in #2359 was somehow contributing to this. That might explain some of the difficulty reproducing.

PR #2342, opened 2021-01-11
use const generics to tie number of IPC regions to number of processes on board

### LINKED ISSUE #1583: Tracking: Couple IPCData array sizes with number of apps on board
Currently, the IPCData array in `kernel/src/ipc.rs` is hardcoded to hold 8 elements, and whenever an application registers as an IPC client, the callback and shared memory for that client is stored at that place in the array. Currently the Hail board is initialized with support for up to 20 apps. If an app with an index greater than 7 attempts to use IPC, IPC will fail.

Ideally we would tie the size of these arrays to the size of the PROCESSES array, though I am not yet sure exactly how that should be done.

hudson-ayers:
### Pull Request Overview

This pull request fixes #1583 . Now, IPC works even if one of the apps has an index higher than 7. Further, when boards are compiled with support for less than 8 processes, `(8-NUM_PROCS)*48` bytes of memory are saved.

### Testing Strategy

This pull request was tested by flashing the IPC apps on an nrf52840dk.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/acd52832/src/main.rs
+++ b/boards/acd52832/src/main.rs
@@ -70,7 +70,7 @@ pub struct Platform {
     >,
     rng: &'static capsules::rng::RngDriver<'static>,
     temp: &'static capsules::temperature::TemperatureSensor<'static>,
-    ipc: kernel::ipc::IPC,
+    ipc: kernel::ipc::IPC<NUM_PROCS>,
     alarm: &'static capsules::alarm::AlarmDriver<
         'static,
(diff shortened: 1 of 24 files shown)


bradjc (approved):
Works great for me on hail.

alistair23 (approved):

alevy (approved):

PR #1010, opened 2018-06-20
Add function that enables FPU for cortex-m4 MCUs

eupn:
### Pull Request Overview

This pull request adds a function that enables Cortex-M4's Floating Point Unit (FPU).

As said in the [User Guide](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0553a/BEHBJHIG.html), [here](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dai0298a/CEGJEIGH.html) and also [here](https://www.silabs.com/community/mcu/32-bit/knowledge-base.entry.html/2014/04/16/how_to_enable_hardwa-vM9u), firmware must manually enable FPU to perform floating-point operations, so it would be nice if that function is available to user.

### Testing Strategy

This pull request was tested by `code compilation`.

### TODO or Help Wanted

This pull request adds a function that still needs a some `#[cfg()]` gate to make sure this function available to use only on Cortex-M4 MCUs, since `scb.rs` module is moved from `cortex-m4` into generic `cortex-m` module.

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [ ] Ran `make formatall`.

PR #2348, opened 2021-01-12
Initial commit of the Cortex-M4F arch crate

alistair23:
### Pull Request Overview

Following up no discussions here: https://github.com/tock/tock/pull/1010#issuecomment-400824717 this PR creates a Cortex-M4F arch crate that can be used for Cortex-M4F boards/chips.

Currently this just creates the crate and adds function to enable the FPU. This doesn't actually enable anything and should have no functional change.

### Testing Strategy

Build testing and running on some of the Nordic boards.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
     "arch/cortex-m0",
     "arch/cortex-m3",
     "arch/cortex-m4",
+    "arch/cortex-m4f",
     "arch/cortex-m7",
     "arch/riscv",
     "arch/rv32i",
--- a/arch/cortex-m/src/scb.rs
+++ b/arch/cortex-m/src/scb.rs
@@ -8,7 +8,7 @@ use kernel::common::StaticRef;
 register_structs! {
     /// In an ARMv7-M processor, a System Control Block (SCB) in the SCS
     /// provides key status information and control features for the processor.
-    ScbRegisters {
+    pub ScbRegisters {
         /// CPUID Base Register
         (0x00 => cpuid: ReadOnly<u32, CpuId::Register>),
 
@@ -58,7 +58,7 @@ register_structs! {
         (0x80 => _reserved1),
 
         /// Coprocessor Access Control Register
-        (0x88 => cpacr: ReadWrite<u32, CoprocessorAccessControl::Register>),
+        (0x88 => pub cpacr: ReadWrite<u32, CoprocessorAccessControl::Register>),
 
         /// 0xE000ED8C, Reserved.
         (0x8c => _reserved2),
@@ -251,7 +251,7 @@ register_bitfields![u32,
         ADDRESS         OFFSET(0)   NUMBITS(32)
     ],
 
-    CoprocessorAccessControl [
+    pub CoprocessorAccessControl [
         CP11            OFFSET(22)  NUMBITS(2),
         CP10            OFFSET(20)  NUMBITS(2),
         CP7             OFFSET(14)  NUMBITS(2),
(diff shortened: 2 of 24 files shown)


bradjc:
Looks good. I assume you are planning to add the syscall implementation?

alistair23:
Yeah, I'm trying to, but pretty stuck.

lschuermann:
As much as I like seeing hard-float support in Tock, I think it might be better to wait on an initial Tock 2.0 release here? My reasoning being that we currently only specify the ABI for 32-bit soft-float, and this would effectively introduce a new, different ABI (as would 64-bit support). This has not only hypothetical consequences, for instance in userspace assembly we currently rely on the fact that hard-float registers (which we don't current have / use) don't get clobbered. I think it might be advantageous to wait after an initial 2.0 release to keep the complexity of the changes low? What do you think @alistair23 @bradjc?

bradjc:
> As much as I like seeing hard-float support in Tock, I think it might be better to wait on an initial Tock 2.0 release here? My reasoning being that we currently only specify the ABI for 32-bit soft-float, and this would effectively introduce a new, different ABI (as would 64-bit support). This has not only hypothetical consequences, for instance in userspace assembly we currently rely on the fact that hard-float registers (which we don't current have / use) don't get clobbered. I think it might be advantageous to wait after an initial 2.0 release to keep the complexity of the changes low? What do you think @alistair23 @bradjc?

This suggests to me that some of the syscall code refactoring updates should be merged into 1.x. Since we can't specify how all future ABIs and architectures will work, specific implementations will have to define individually exactly how the ABI works for that architecture. Since 2.0 isn't passing more values between kernel and userspace, a 1.x implementation of the ABI should map well to the 2.0 interface.

alistair23:
Also worth noting that this PR doesn't enable the FPU, it just creates a cortex-m4f crate and prepares to enable the FPU. More work is still required before it can be turned on.

lschuermann:
> Since we can't specify how all future ABIs and architectures will work, specific implementations will have to define individually exactly how the ABI works for that architecture.

I wasn't trying to impl that we can't have ABIs following the precise infrastructure in the kernel (which is why we have the `UserspaceKernelBoundary` trait in the first place, right?). Rather, I was wary of some added complexity to reason about in all of the changes (of especially libtock-rs) with respect to the side-effects of the system call. It wasn't an objection to include this - and I think it'll work just fine - it might just make the diff a bit larger and requires careful consideration of the different clobbers and flags in the userspace libraries. I recall @jrvanwhy asking in this direction in Slack.

alistair23:
Ping!

hudson-ayers:
The code in this PR looks fine, but merging it as-is would put us in the strange position of having an "enable_fpu" function that looks like it should work (look, there is even a stand-alone architecture crate just for floating point support!), but if called, breaks userspace.

I think Pat's comment on #1010 still applies: "As it stands, if someone playing around with Tock called this function, userland would break in fairly cryptic fashion with very little introspection into why* (and somewhat far removed from this function call as the error would not appear until the next context switch)."

This PR also adds some code duplication / additional abstraction with no real benefit until the UKB implementation is done.

bradjc:
I agree that having a cortex-m4f crate without actually having support for m4f would be confusing.

ppannuto:
One particularly interesting question for hard/soft float in ARM is userspace awareness. In particular, I _think_ it should be possible to run apps compiled for soft float in hard float mode without any modification -- however, if that's wrong, it could put the kernel in the very interesting space of needing to enable and disable hard float depending on which application is running.

The related question is whether the kernel should be compiled with hard float enabled or disabled, as it does incur non-trivial additional overhead when it's enabled but not used.

hudson-ayers:
Another consideration here is that it probably does not make sense to enable the FPU within the kernel. As I understand it, Linux requires calling `kernel_fpu_begin()` in order to safely use hardware floating point within the kernel. As a result, Linux only has to save/restore floating point state when switching between userspace processes that both want to use the FPU. This might be a good design for Tock as well, to prevent adding a ton of extra context switch overhead to all context switches (in much the same way that we only update the MPU configuration when switching to a different app).

We could even take this a step further and have tockloader auto-detect if hard-float is enabled by using `objdump` or `readelf` and then marking as such in the tbf header. Then the kernel would only have to enable the FPU for apps that actually use it

PR #2351, opened 2021-01-13
Yield 2.0

phil-levis:
### Pull Request Overview

This pull request is a draft for `yield` in Tock 2.0. It implements `yield-no-wait` following the approach proposed by @jrvanwhy of passing a pointer to store the result. It has `yield-no-return` (i.e., exit) unimplemented for now. It also updates the the syscall TRD with the arguments and their values.

### Testing Strategy

This pull request has been tested with the `yield` test application in libtock-c. The two yield variants work as intended.

### TODO or Help Wanted

None.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


kernel/src/process.rs
```diff
@@ -1387,6 +1405,14 @@ impl<C: Chip> ProcessType for Process<'_, C> {
         }
     }
 
+    fn yield_wait_field(&self, addr: *mut u8) -> Option <&mut u8> {
+        if self.in_app_owned_memory(addr, 1) {
+            unsafe {Some(&mut *addr)}
+        } else {
+            None
+        }
+    }
```
bradjc, line 1414: I think rather than using unsafe to return a reference, this would be better as `fn set_yield_flag(&self, addr: *mut u8, value: u8)`. Then accessing process memory is encapsulated in process.rs (which could be useful for host side testing and different Process implementations).

doc/reference/trd-syscalls.md
```diff
@@ -260,24 +260,51 @@ call to invoke. The kernel invokes callbacks only in response to Yield
...
-If no callbacks are pending it returns immediately. The `yield-no-wait` system
-call returns `Success` if a callback executed and `Failure` if no callback
-executed.
+The second call, `yield-no-wait`, executes a single callback if any is pending.
+If no callbacks are pending it returns immediately. 
 
-The return values for Yield system calls are `Success` and `Failure`. 
+The third call, `yield-no-return` never returns and executes no callbacks. It
+causes the process to terminate. The kernel can reclaim any resources held by
+a process after it invokes `yield-no-return`.
```
jrvanwhy, line 280: The name `yield-no-return` doesn't suggest to me that the process can end and be reclaimed. Is there a reason you picked `yield-no-return` over `yield-exit`?

arjundeopujari: @jrvanwhy @phil-levis Yes, I think `yield-exit` makes it more clear that this system call would invoke the kernel to terminate the app and reclaim the app memory.  `yield-no-return` seems a bit more ambiguous.

kernel/src/sched.rs
```diff
@@ -779,14 +782,31 @@ impl Kernel {
                 }
                 process.set_syscall_return_value(rval);
             }
-            Syscall::Yield => {
+            Syscall::Yield { wait, terminate, address} => {
                 if config::CONFIG.trace_syscalls {
-                    debug!("[{:?}] yield", process.appid());
+                    debug!("[{:?}] yield {}", process.appid(), wait);
```
hudson-ayers, line 787:
```suggestion
                    debug!("[{:?}] yield. wait: {}", process.appid(), wait);
```

kernel/src/process.rs
hudson-ayers: agreed. This function is not currently marked `unsafe`, but it would be easy to create two mutable references to the same memory location by calling this twice, which is UB.

phil-levis:
OK, transitioning this out of draft. Once the current design gets a thumbs up, I'll start testing.

doc/reference/trd-syscalls.md
```diff
@@ -264,26 +264,51 @@ call to invoke. The kernel invokes callbacks only in response to Yield
...
+| Yield identifer        | r0       |
+| No wait field          | r1       |
+
+
+The yield identifier specifies which call is invoked.
+
+| System call     | Yield identifier value |
+|-----------------|------------------------|
+| yield-no-wait   |                      0 |
+| yield-wait      |                      1 |
```
bradjc, line 300: Since `yield-wait` is the common case, I expected that it would be 0, and the unlikely variants would be numbered from there.

doc/reference/trd-syscalls.md
```diff
@@ -264,26 +264,51 @@ call to invoke. The kernel invokes callbacks only in response to Yield
...
+| yield-exit      |                      2 |
+
+
+The no wait field is only used by `yield-no-wait`. It contains the
```
bradjc, line 304: I would still like to see processes be able to include an argument when they call `exit`, and have a convention where the process can indicate to the kernel if it would like to be restarted.

kernel/src/syscall.rs
```diff
@@ -64,7 +64,11 @@ pub enum Syscall {
     /// interrupts and callbacks.
     ///
     /// System call class ID 0
-    Yield,
+    Yield {
+        wait: bool,
+        terminate: bool,
+        address: *mut u8,
+    },
```
bradjc, line 71: Splitting the parsing of yield into two places (once when this enum is created, and then again in sched) makes this a bit difficult to follow. Following memop's lead, this could just store values, and leave the interpretation of those values when the syscall is handled.

doc/reference/trd-syscalls.md
lschuermann: Regarding the exit argument and restart policy, could it make sense to have a separate syscall for that (essentially a "process control system call")? That way the process could set a return value / set the restart policy manually using this command, and then call yield-exit (with the usual yield semantics, but just stopping execution indefinitely)?

bradjc: Also since Linux came up on Friday, there is an explicit exit syscall (https://man7.org/linux/man-pages/man2/_exit.2.html) in Linux which just [exits](https://code.woboq.org/userspace/glibc/sysdeps/unix/sysv/linux/_exit.c.html#_exit), and is separate from [yield](https://man7.org/linux/man-pages/man2/sched_yield.2.html). I still think yielding and exiting are separate, but `yield-exit` is pretty specific and clear. But even though `exit()` in Linux does do other things, it does eventually call the exit syscall.

alistair23: +1 to exit having a return value.

kernel/src/syscall.rs
phil-levis: Makes sense. This two-place parsing was from when address was an Option.

doc/reference/trd-syscalls.md
phil-levis: Exit has a return value or an argument?

If exit has an argument, we should specify what the argument means. We need to specify its values and their semantics.

I have a different proposal: there should be `yield-exit` and `yield-exit-restart`. We can extend different exit cases by adding more system calls.

doc/reference/trd-syscalls.md
phil-levis: I thought "0 is false, 1 is true." It also means that 0 does not block, all other values block (some terminally).

arjundeopujari: I think its fine as it is.  It makes more sense to me for it to be structured like this...

0  - no waiting
1 - wait/block until callback
2 - wait/block forever (exit)  

I dont know if one should focus on common case.  It seems the numbering, as it is, correlates with the "waiting" period.

doc/reference/trd-syscalls.md
arjundeopujari: >  have a different proposal: there should be yield-exit and yield-exit-restart. We can extend different exit cases by 
> adding more system calls.

Wouldn't adding more syscalls for every exit condition complicate the interface a bit?  Adding an "exit status" argument could make it more similar to Linux's `void _exit(int status)` call.

phil-levis: Now that we have two exit cases, I think I'm on board with Brad's suggestion. We should have a yield with two variants (-wait and -no-wait), and an exit with two variants (-terminate and -restart).

phil-levis: I've factored out exit from yield. I think they are semantically different: exit never returns and doesn't execute syscalls. If we need to expand the set of cases for either we'll want them distinct, in terms of shared functionality. Otherwise I'm worried about a fragmented identifier space (e.g., 0, 1, 4, 5 are yields, 2, 3, 6 are exits, etc.).

kernel/src/syscall.rs
phil-levis: Addressed.

phil-levis:
I've cut out exit, but written it up in the design document. Assuming the design document and yield implementations are good, I'll do a later PR for the exit implementations (building on earlier PRs from others). I'm keeping exit in the design document in this PR because I think the split between them is part of yield's design.

phil-levis:
OK, I think this is ready to go. I'll do an exit() PR soon.

kernel/src/process.rs
```diff
@@ -416,6 +420,11 @@ pub trait ProcessType {
             -> Result<ReadOnlyAppSlice, (ReadOnlyAppSlice, ErrorCode)>,
     ) -> GenericSyscallReturnValue;
 
+    /// Set a single byte within the process address space at
+    /// `addr` to `value`. Return true if `addr` is a valid
+    /// address and the value was written, false otherwise.
```
lschuermann, line 425: In `process.rs` we have to be careful about what guarantees and preconditions the various methods enforce, to not risk writing to some other (potentially kernel-reserved) memory region. Maybe it's a good idea to copy parts of the documentation of `in_app_owned_memory` to also apply here?:

```suggestion
    /// Set a single byte within the process address space at
    /// `addr` to `value`. Return true if `addr` is within the RAM
    /// bounds currently exposed to the process (thereby writable
    /// by the process itself) and the value was set, false otherwise.
```

lschuermann (approved):
Looks good to me, other than some minor documentation nitpicks.

alistair23 (changes requested):

doc/reference/trd-syscalls.md
```diff
@@ -539,8 +585,45 @@ are 12:
...
+restarts the application, it MUST assign it a new process identifier. The
+kernel MAY reuse existing process resources (e.g., RAM regions) or MAY
+allocate new ones.
+
+The register arguments for Exit system calls are as follows. The registers
+r0-r3 correspond to r0-r3 on CortexM and a0-a3 on RISC-V.
+
+| Argument          | Register |
+|-------------------|----------|
+| Exit identifer    | r0       |
```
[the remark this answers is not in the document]
jrvanwhy: I wonder if an exit code could replace `LowLevelDebug`'s [predefined alert codes](https://github.com/tock/tock/blob/master/doc/syscalls/00008_low_level_debug.md#predefined-alert-codes) mechanism? Both alert codes that currently exist indicate conditions where the process should terminate.

On another note: when I defined the `LowLevelDebug` alert codes, I made them global across all runtimes. In retrospect, I think those alert codes should belong to the runtimes (i.e. `libtock-c` and `libtock-rs` would each have their own alert code tables).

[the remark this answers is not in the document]
phil-levis: In the exit PR, it can return a return value, which is always Failure.

PR #2354, opened 2021-01-14
kernel: syscall: use `from` fn

bradjc:
Rather than a separate helper function, use a `From` pattern. This is more rust-y.

Backported from 2.0 branch.

Why? This will make the 2.0 diff easier to read, because the relevant code will show an inline diff, rather than just seeing one block of code deleted and a different block added.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers (approved):

lschuermann (approved):

alistair23 (approved):

bradjc:
Good to go?

PR #2355, opened 2021-01-14
kernel: add in_app_owned_flash fn

bradjc:
Add `in_app_owned_flash()` to `Process`. This mirrors the RAM version.

Backported from 2.0 branch.

It didn't occur to me until after that this would raise the dead code warning, so maybe this isn't worth it? It just makes reviewing the 2.0 changes slightly easier since we can look at the check now, rather than with all of the other changes.

### Testing Strategy

travis / 2.0 branch

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -2222,8 +2222,8 @@ impl<C: 'static + Chip> Process<'_, C> {
     }
 
     /// Checks if the buffer represented by the passed in base pointer and size
-    /// are within the memory bounds currently exposed to the processes (i.e.
-    /// ending at `app_break`. If this method returns true, the buffer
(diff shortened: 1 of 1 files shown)


alistair23 (approved):

hudson-ayers (approved):

bradjc:
Anyone else want to double check the bounds checking is correct?

bradjc:
What about my fix to the other comment?

hudson-ayers:
IMO at the point it might be better to just close this and let it take care of itself when #2446 is merged, especially since this adds the allow() for the dead code warning that is going to have to be removed as soon as #2446 is merged anyway.

PR #2358, opened 2021-01-14
kernel: move tbf code to its own library

bradjc:
### Pull Request Overview

As part of the effort to remove `unsafe` from the TBF parsing code (#1465), the last goal was always to move all of the TBF parsing code to its own library since by definition the kernel is not the only thing that needs to handle TBFs. Rather than have TBF code re-implemented, this would allow there to be a single TBF parsing crate that can be used by the kernel and also tools like elf2tab.

Turns out, moving this to a library is quite easy. I didn't want to do this earlier in case there were fixes or updates needed to the earlier tbf parsing re-write from #1465. But since the code has been working now seems like a reasonable time to move it to a library.

Apologies to @alistair23 for moving code, but I can help update if git doesn't do it automatically.

### Testing Strategy

This pull request was tested by running kernels on nano33 and hail, but really this doesn't change any logic.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/libraries/tock-tbf/Cargo.toml
+++ b/libraries/tock-tbf/Cargo.toml
@@ -1,0 +1,5 @@
+[package]
+name = "tock-tbf"
+version = "0.1.0"
+authors = ["Tock Project Developers <tock-dev@googlegroups.com>"]
+
--- a/libraries/tock-tbf/Cargo.toml
+++ b/libraries/tock-tbf/Cargo.toml
@@ -2,4 +2,4 @@
 name = "tock-tbf"
 version = "0.1.0"
 authors = ["Tock Project Developers <tock-dev@googlegroups.com>"]
-
+edition = "2018"
--- a/libraries/tock-tbf/src/lib.rs
+++ b/libraries/tock-tbf/src/lib.rs
@@ -4,4 +4,5 @@
 #![forbid(unsafe_code)]
 #![no_std]
 
-pub mod tbfheader;
+pub mod parse;
+pub mod types;
--- a/libraries/tock-tbf/src/lib.rs
+++ b/libraries/tock-tbf/src/lib.rs
@@ -3,6 +3,5 @@
 // Parsing the headers does not require any unsafe operations.
 #![forbid(unsafe_code)]
 #![no_std]
-#![feature(try_trait)]
 
 pub mod tbfheader;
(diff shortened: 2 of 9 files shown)


hudson-ayers (dismissed):
This is great. Looks like git tracked the moved files correctly. 

Nice to remove another nightly feature from the kernel crate (now people can't use it elsewhere in the kernel).

alistair23 (approved):

libraries/tock-tbf/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+//! Tock Binary Format (TBF) header parsing library.
+
+// Parsing the headers does not require any unsafe operations.
+#![forbid(unsafe_code)]
+#![no_std]
+#![feature(try_trait)]
```
hudson-ayers, line 6:
```suggestion
```

This feature appears unused!

hudson-ayers (dismissed):

alistair23 (dismissed):

libraries/tock-tbf/Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[package]
+name = "tock-tbf"
+version = "0.1.0"
+authors = ["Tock Project Developers <tock-dev@googlegroups.com>"]
```
jrvanwhy, line 4: Should tock-tbf be on the 2018 edition?

libraries/tock-tbf/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+//! Tock Binary Format (TBF) header parsing library.
+
+// Parsing the headers does not require any unsafe operations.
+#![forbid(unsafe_code)]
+#![no_std]
+
+pub mod tbfheader;
```
jrvanwhy, line 7: This namespaces everything under `tock_tbf::tbfheader`. Was that done to provide for new modules in the future? If not, I think everything should be in the `tock_tbf` namespace rather than a submodule.

bradjc: I didn't think about it, specifically, but doing this was convenient because it meant almost zero changes to process.rs.

I suppose we do need to think about how tock-tbf is organized. I can work on that.

libraries/tock-tbf/Cargo.toml
bradjc: Yes. Copied from tock-cells, so that library should be updated too.

bradjc:
I added some structure to the tbf library crate. I moved the types into their own file, and all of the parsing code into its own file. This makes it much more straightforward to add code to create tbfs in the future.

hudson-ayers (approved):

alistair23 (approved):

jrvanwhy (approved):

PR #2362, opened 2021-01-17
Tock 2.0: port nonvolatile storage & app flash driver

lschuermann:
### Pull Request Overview

This pull request ports both the nonvolatile storage driver and the app flash driver to the Tock 2.0 system call interface.

I first attempted to port the nonvolatile storage driver exclusively, but that appeared to be pretty complicated as we'd have to deal with converting between `ReturnCode` and `ErrorCode` quite a bit. It seemed more elegant to directly migrate the `NonvolatileStorage` HIL to use `Result<_, ErrorCode>` for its return values and change the rest of the infrastructure accordingly. This is also the reason why I've taken on the app flash driver as well. Hope that's okay with @ppannuto, who wanted to do that originally.

The commits changing the HIL (5600e76 .. 9c3097e) and the ones changing the ABI (LegacyDriver -> Driver, 619ed72 .. 7555d30) are separated such that they can be reviewed as a single PR, but (as long as we also backport ErrorCode) the HIL changes can be easily backported to master afterwards.

### Testing Strategy

This pull request was tested by running the `nonvolatile_storage` and `app_state` tests in `libtock-c` on the accompanying userspace PR.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/app_flash_driver.rs
+++ b/capsules/src/app_flash_driver.rs
@@ -23,10 +23,11 @@
 //! ```
 
 use core::cmp;
+use core::mem;
 use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::hil;
 use kernel::ErrorCode;
-use kernel::{AppId, AppSlice, Callback, Grant, LegacyDriver, ReturnCode, SharedReadWrite};
+use kernel::{AppId, Callback, CommandResult, Driver, Grant, Read, ReadOnlyAppSlice};
 
 /// Syscall driver number.
 use crate::driver;
@@ -34,8 +35,8 @@ pub const DRIVER_NUM: usize = driver::NUM::AppFlash as usize;
 
 #[derive(Default)]
 pub struct App {
-    callback: Option<Callback>,
-    buffer: Option<AppSlice<SharedReadWrite, u8>>,
+    callback: Callback,
+    buffer: ReadOnlyAppSlice,
     pending_command: bool,
     flash_address: usize,
 }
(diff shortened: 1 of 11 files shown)


capsules/src/nonvolatile_storage_driver.rs
```diff
@@ -526,46 +586,54 @@ impl LegacyDriver for NonvolatileStorage<'_> {
...
-                ReturnCode::SUCCESS
+            0 /* This driver exists. */ => {
+                CommandResult::success()
             }
 
-            // How many bytes are accessible from userspace.
-            1 => ReturnCode::SuccessWithValue {
-                value: self.userspace_length,
+            1 /* How many bytes are accessible from userspace */ => {
+		// TODO: Would break on 64-bit platforms
```
hudson-ayers, line 602: well, only if more than 4 GB are reserved as accessible from userspace, which seems like a platform pretty far outside Tock's wheelhouse for now

capsules/src/app_flash_driver.rs
```diff
@@ -132,20 +132,23 @@ impl hil::nonvolatile_storage::NonvolatileStorageClient<'static> for AppFlash<'_
...
                                     *c = d[i];
                                 }
 
-                                self.driver.write(buffer, flash_address, length)
-                                    == ReturnCode::SUCCESS
+                                if let Ok(()) = self.driver.write(buffer, flash_address, length) {
+                                    true
+                                } else {
+                                    false
+                                }
```
hudson-ayers, line 151:
```suggestion
                                self.driver.write(buffer, flash_address, length).is_ok()
```

hudson-ayers (approved):
This looks good to me! A few small comments.

phil-levis (approved):
Looks good to me. Agree with Hudson on deprecation of SuccessWithValue.

PR #2363, opened 2021-01-18
rv32i: transition llvm_asm! to asm!

lschuermann:
### Pull Request Overview

This pull request transitions the remaining `llvm_asm!` snippet of the `rv32i` towards the new Rust `asm!` macro (see #1923 for the respective tracking issue). It's split up into commits that represent atomic changes respectively.

In addition to the raw transition, this changes two things

- it removes the unused `abort()` / `_abort` fuction. It was previously public & safe, such that untrusted code could simply call to it.

- in `switch_to_process`, it forces the `Riscv32iStoredState` pointer to be passed in a specific register. This is important since, while we don't clobber any registers in the assembly block there, we do write to `t0` after we've saved it to the stack. Later, we access the stored state pointer (from some unspecified register). If the compiler were to put this information in `t0`, we would access the overwritten contents instead. By fixing the state pointer to always be passed in `a0` (which the compiler has done up to now), we can rely on being able to safely overwrite `t0` after it's been saved on the stack.

  Notably, knowing the register where the stored state is passed in potentially enables an optimization where we don't need to save it to the stack twice (and hence save a memory write & read). However, implementing that made the ASM quite a bit more confusing, so I might implement that as a follow up PR, once I find the time to write it in a clean fashion.

All in all, the abort function assembly is removed
```diff
- 00015210 <abort>:
-    15210:	df1ea06f          	j	0 <_start>
-    15214:	00008067          	ret
```
and the  `switch_to_process` function seems to generate the exact same assembly.

One weird thing about `asm!()`: I tried to replace the `lui` + `addi` based immediate load by a `li` pseudo-instruction, however the compiler complained that `_return_to_kernel` was not a valid immediate value there, whereas it continues to work with the `%hi` / `%lo` accessors. The solution™ is to insert the immediate value as a constant, however we can't refer to an assembly label there. Hence leaving it the way it is.

### Testing Strategy

This pull request was tested by running a userspace app in the LiteX simulator and on the LiteX Arty board.

### TODO or Help Wanted

This pull request still needs careful review. I can recommend stepping through the commits individually such that the changes can be reviewed separately.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/lib.rs
+++ b/arch/rv32i/src/lib.rs
@@ -2,7 +2,7 @@
 
 #![crate_name = "rv32i"]
 #![crate_type = "rlib"]
-#![feature(llvm_asm, asm, const_fn, naked_functions)]
+#![feature(asm, const_fn, naked_functions)]
 #![no_std]
 
 use core::fmt::Write;
--- a/arch/rv32i/src/lib.rs
+++ b/arch/rv32i/src/lib.rs
@@ -384,23 +384,6 @@ pub extern "C" fn _start_trap() {
     }
 }
 
-/// Ensure an abort symbol exists.
-#[cfg(all(target_arch = "riscv32", target_os = "none"))]
-#[link_section = ".init"]
-#[export_name = "abort"]
-pub extern "C" fn abort() {
-    unsafe {
-        llvm_asm! ("
-            // Simply go back to the start as if we had just booted.
-            j    _start
-        "
-        :
-        :
-        :
-        : "volatile");
-    }
-}
-
 /// Print a readable string for an mcause reason.
 pub unsafe fn print_mcause(mcval: csr::mcause::Trap, writer: &mut dyn Write) {
     match mcval {
--- a/arch/rv32i/src/syscall.rs
+++ b/arch/rv32i/src/syscall.rs
@@ -245,7 +245,7 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
             // Store process state pointer on stack as well. We need
             // to have this available for after the app returns to the
             // kernel so we can store its registers.
-            "sw   {state_struct}, 1*4(sp)",
+            "sw   a0, 1*4(sp)",
 
             // From here on we can't allow the CPU to take interrupts
             // anymore, as that might result in the trap handler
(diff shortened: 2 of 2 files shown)


arch/rv32i/src/syscall.rs
```diff
@@ -130,221 +130,226 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
...
+            //  5*4(sp): x4
+            //  4*4(sp): x3
+            //  3*4(sp): x1
+            //  2*4(sp): _return_to_kernel (address to resume after trap)
+            //  1*4(sp): *state   (Per-process StoredState struct)
+            //  0*4(sp): app s0   <- new stack pointer
+            // ```
+
+            // Move the stack pointer down to make room.
+            "addi sp, sp, -34*4",
```
bradjc, line 182: Why add all of the `"",` on each line?

arch/rv32i/src/lib.rs
```diff
@@ -388,17 +388,10 @@ pub extern "C" fn _start_trap() {
...
-            // Simply go back to the start as if we had just booted.
-            j    _start
-        "
-        :
-        :
-        :
-        : "volatile");
-    }
+#[naked]
+pub unsafe extern "C" fn abort() {
```
bradjc, line 392: I think even better we can just remove this entire function. I'm not sure if it was needed before or just a bad copy-and-paste on my part.

arch/rv32i/src/syscall.rs
lschuermann: I've seen some examples of other people using `asm!()` doing that. It has the added benefit of my editor and rustfmt aligning the instructions correctly and detecting whether I'm using the correct, spaces-based indentation. I think either would be fine with me though.

lschuermann: [The Rust unstable book seems to prefer this syntax over multiline strings](https://doc.rust-lang.org/nightly/unstable-book/library-features/asm.html#inputs-and-outputs), for what it's worth.

arch/rv32i/src/lib.rs
lschuermann: Right, it really doesn't seem to be referenced anywhere. Given it's side effects when called are pretty significant, it's probably best to remove if we don't use it. Done.

arch/rv32i/src/syscall.rs
bradjc: I do wish they would add more examples to the book. The longest assembly block there is just two instructions.

lschuermann: I agree that the documentation is very sparse. Reading through the documentation again seems that using multiple strings was chosen exactly for the benefits I've noticed while using it. To cite them: _"First, we can see that asm! allows multiple template string arguments; each one is treated as a separate line of assembly code, as if they were all joined together with newlines between them. This makes it easy to format assembly code."_

As I said both is fine with me. Should I change it to be a multiline string?

alistair23 (dismissed):

hudson-ayers (approved):

arch/rv32i/src/syscall.rs
bradjc: I'm typically on the side of smaller commits are better. However, I'm also a believer that if rustfmt doesn't enforce it then it's completely up to the author. I am worried about a collision with my stack pointer PR however :)

bradjc:
I did the rebase needed because of my stack pointer changes, and I simplified the PR to just the changes needed which makes the diff much easier to compare.

alistair23 (approved):

hudson-ayers (approved):

lschuermann:
> simplified the PR to just the changes needed which makes the diff much easier to compare

Fair enough. I have a slight preference over the other syntax, but in the end what really matters is that we have a consistent style around the Tock kernel. If we were to change it to the multi-string syntax (which the book seems to have a slight preference over) we should do so throughout the kernel, in a separate commit.

bradjc:
There is certainly discussion about this: https://github.com/rust-dev-tools/fmt-rfcs/issues/152. But I assume rustfmt will not actually add the `"",` on each line, and I don't think we will actually be able to keep it consistent. But that doesn't mean we can't try.

lschuermann:
> I don't think we will actually be able to keep it consistent. But that doesn't mean we can't try.

I didn't want to imply we _must_ switch, and my personal preference probably doesn't matter much for the project. I haven't seen the rustfmt discussion, thanks for sharing the link. Maybe the best approach is to let dust settle there and see what the most commonly accepted and preferred syntax is in a few months?

PR #2367, opened 2021-01-19
ctap: Update to use Tock2.0 Driver

alistair23:
### Pull Request Overview

Update CTAP capsule to use the new Tock2.0 Driver.

### Testing Strategy

Unfortunately libtock-rs doesn't seem to work with 2.0, so no testing.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/ctap.rs
+++ b/capsules/src/ctap.rs
@@ -29,27 +29,30 @@
 
 use core::cell::Cell;
 use core::marker::PhantomData;
+use core::mem;
 use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::hil::usb_hid;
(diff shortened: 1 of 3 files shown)


phil-levis:
Can you test with libtock-c?

hudson-ayers (approved):

phil-levis:
I think we're kind of at an impasse, then -- we should not be committing untested code to mainline Tock.

phil-levis:
Discussion on the core call today reached consensus that it's OK to merge this untested with the -dev branch, but will need testing before merging -dev into master. This particular PR/syscall set is tricky due to CTAP depending on userspace and libtock-rs using the ctap crate, but we will cross that bridge when we get to merging with -master. In the worst case, this capsule will be marked "untested" or "dev" or something (but remain in the repository).

Issue #2320, opened 2021-01-01
Tock 2.0: callbacks, return code, and errorcode


hudson-ayers:
Tock 2.0 changes system calls such that success vs. failure is now passed in a separate register from the register used to communicate error codes in the case of a failure. This works fine because we increased the number of registers used for system calls, and meshes well with our desire to eventually remove `ReturnCode` from the Tock kernel in favor of `Result<T, ErrorCode>`.

However, callbacks still use `ReturnCode` to communicate errorcodes + the possibility of success in a single register. And the number of registers used to communicate values in callbacks has not changed for Tock 2.0, so using the same approach we do for `Command` is not feasible -- several callbacks exist that already use all of the available registers. Callbacks are generally a contract between a capsule and a corresponding userspace driver, so it is okay for callbacks to still use ReturnCode or some similar format for communicating errors. But I anticipate that there will be some confusion as some capsules start to report ErrorCode's to communicate errors while others still use ReturnCode's -- especially because the values used to represent the same errors do not match.

After some discussion with @lschuermann , I believe the best solution would be to change `ReturnCode` to represent errors using the same underlying values as `ErrorCode` does, after changing `ErrorCode` to not represent any errors using `0`. This would mean `ReturnCode` and `ErrorCode` would be identical types, except that `ReturnCode` would also have a `Success: 0` variant. Even once `ReturnCode` is deprecated, we could have a function like the following:

```rust
impl From<Option<ErrorCode>> for u32 {
    ...
}
```
so that a `Result<T, ErrorCode>` can easily be communicated as success/failure in a callback via:

```rust
self.callback.schedule(e.ok().into(), something as u32, something as u32)
```

Unfortunately, this would require changing a substantial amount of error handling code in `libtock-c` apps which relies on the fact that, in callbacks, errors are negative and success is positive.

Technically, there is no reason we have to make this change as part of 2.0 -- as the types of values passed in callbacks are not part of the ABI contract. Is this a change we should try to make before delviering 2.0, or schedule for sometime afterwards (perhaps whenever `ReturnCode` is actually deprecated in the kernel?)

phil-levis:
[...]

PR #2368, opened 2021-01-19
hmac: Update to use Tock2.0 Driver

alistair23:
### Pull Request Overview

Update the HMAC capsule and HIL to match the Tock2.0 design.

### Testing Strategy

Non, libtock-rs doesn't yet work with 2.0.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/hmac.rs
+++ b/capsules/src/hmac.rs
@@ -30,11 +30,15 @@ pub const DRIVER_NUM: usize = driver::NUM::Hmac as usize;
 use core::cell::Cell;
 use core::convert::TryInto;
 use core::marker::PhantomData;
+use core::mem;
 use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::common::leasable_buffer::LeasableBuffer;
 use kernel::hil::digest;
 use kernel::hil::digest::DigestType;
-use kernel::{AppId, AppSlice, Callback, Grant, LegacyDriver, ReturnCode, SharedReadWrite};
+use kernel::{
+    AppId, Callback, CommandResult, Driver, ErrorCode, Grant, Read, ReadWrite, ReadWriteAppSlice,
+    ReturnCode,
+};
 
 pub struct HmacDriver<'a, H: digest::Digest<'a, T>, T: 'static + DigestType> {
     hmac: &'a H,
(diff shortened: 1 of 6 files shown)


capsules/src/hmac.rs
```diff
@@ -246,26 +236,23 @@ impl<'a, H: digest::Digest<'a, T> + digest::HMACSha256, T: DigestType> digest::C
...
-                    app.callback.map(|cb| match result {
-                        Ok(_) => cb.schedule(0, pointer as usize, 0),
-                        Err(e) => cb.schedule(usize::from(e), pointer as usize, 0),
+                    app.data.mut_map_or((), |dest| {
+                        dest.as_mut().copy_from_slice(digest.as_ref());
                     });
 
+                    match result {
+                        Ok(_) => app.callback.schedule(0, pointer as usize, 0),
+                        Err(e) => app.callback.schedule(usize::from(e), pointer as usize, 0),
```
hudson-ayers, line 253: This looks like exactly the error I was hoping to prevent with the (as of now unresolved) discussion in #2320 . Because `ErrorCode` and `ReturnCode` use different values for the same errors, this now passes different error values to userspace than before. Which would require as-of-yet nonexistent changes in userspace to properly support. So for now, this should probably convert the error code to a return code, then to a usize, in order to not break any existing error handling code.

capsules/src/hmac.rs
```diff
@@ -447,39 +451,39 @@ impl<'a, H: digest::Digest<'a, T> + digest::HMACSha256, T: DigestType> LegacyDri
...
             pending_run_app: None,
-            key: None,
-            data: None,
-            dest: None,
+            key: ReadWriteAppSlice::default(),
+            data: ReadWriteAppSlice::default(),
+            dest: ReadWriteAppSlice::default(),
         }
     }
 }
```
hudson-ayers, line 489: I think this could just derive `Default`, which would cut some unneeded lines of code.

hudson-ayers (commented):
This looks good, but I think the callback needs to use a ReturnCode for now until #2320 is resolved.

phil-levis:
Can you test with libtock-c?

capsules/src/hmac.rs
```diff
@@ -227,9 +220,7 @@ impl<'a, H: digest::Digest<'a, T> + digest::HMACSha256, T: DigestType> digest::C
                         self.hmac.clear_data();
                         self.appid.clear();
 
-                        app.callback.map(|cb| {
-                            cb.schedule(usize::from(e.0), 0, 0);
-                        });
+                        app.callback.schedule(usize::from(e.0), 0, 0);
```
hudson-ayers, line 223: Sorry, I missed this one on my first review, but this also needs to do the `ReturnCode` converesion

hudson-ayers (approved):

phil-levis:
Needs testing.

phil-levis:
Discussion on the core call today reached consensus that it's OK to merge this untested with the -dev branch, but will need testing before merging -dev into master.

@lschuermann Volunteered to write a libtock-c library/test for HMAC.

PR #2370, opened 2021-01-20
Rework the initial stack pointer and memory break the kernel gives to process.

jrvanwhy:
### Pull Request Overview

Previously, the kernel gave the app 3 KiB of usable memory and initialized the stack pointer to the top of that memory. This was a backwards-compatibility measure to support Tock 1.0 process binaries. This commit makes the `UserspaceKernelBoundary` responsible for picking the initial stack pointer and process break, and reduces the guarantees the kernel provides to userspace.

In particular:

1. The memory break is now set as small as possible. Process initialization should move the break upwards before setting the stack pointer.
2. The stack pointer is set to the memory break on ARM and set to an unspecified value (currently 0) on RISC-V

This has the following benefits over the previous setup:

1. It supports processes that use less than 3KiB of memory.
2. It makes process runtime setup easier on ARM, as the memory break should unconditionally be set before the stack pointer, whereas this previously depended on the process' stack size and desired break.

I am unsure how much stack space ARM needs at startup, so there is currently a placeholder value. I intend to set that value correctly before merging this.

### Testing Strategy

This PR is untested. I do not have physical hardware to test on. I could test in QEMU on HiFive using `libtock-rs`, but I would need to write the entry point to do so.

### TODO or Help Wanted

I need help identifying the correct stack size to use on ARM. I also need help (ideas?) testing this.

### Documentation Updated

- [X] Updated the relevant files in `/docs`, or no updates are required. *I don't think updates are required*

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/syscall.rs
+++ b/arch/cortex-m/src/syscall.rs
@@ -63,9 +63,7 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
         // interrupt executes, so we need to give the process an initial usable
         // stack. We set the initial break at the top of the stack -- processes
         // must move the break upwards in their runtime implementation.
-        // TODO: 56 is a placeholder -- how large a space do we need to allocate
-        // here?
-        let original_stack_pointer = process_memory_start.wrapping_add(56);
+        let original_stack_pointer = process_memory_start.wrapping_add(0x20);
         kernel::syscall::InitialProcessLayout {
             original_break: original_stack_pointer,
             original_stack_pointer,
--- a/arch/cortex-m/src/syscall.rs
+++ b/arch/cortex-m/src/syscall.rs
@@ -58,16 +58,13 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
     fn initial_process_layout(
         &self,
         process_memory_start: *const u8,
-    ) -> kernel::syscall::InitialProcessLayout {
+    ) -> (*const u8, *const u8) {
         // ARM uses the userspace stack to store register values when an
         // interrupt executes, so we need to give the process an initial usable
         // stack. We set the initial break at the top of the stack -- processes
         // must move the break upwards in their runtime implementation.
         let original_stack_pointer = process_memory_start.wrapping_add(0x20);
-        kernel::syscall::InitialProcessLayout {
-            original_break: original_stack_pointer,
-            original_stack_pointer,
-        }
+        (original_stack_pointer, original_stack_pointer)
     }
 
     unsafe fn initialize_process(
--- a/arch/cortex-m/src/syscall.rs
+++ b/arch/cortex-m/src/syscall.rs
@@ -55,6 +55,23 @@ impl SysCall {
 impl kernel::syscall::UserspaceKernelBoundary for SysCall {
     type StoredState = CortexMStoredState;
 
+    fn initial_process_layout(
+        &self,
+        process_memory_start: *const u8,
+    ) -> kernel::syscall::InitialProcessLayout {
+        // ARM uses the userspace stack to store register values when an
+        // interrupt executes, so we need to give the process an initial usable
+        // stack. We set the initial break at the top of the stack -- processes
+        // must move the break upwards in their runtime implementation.
+        // TODO: 56 is a placeholder -- how large a space do we need to allocate
+        // here?
+        let original_stack_pointer = process_memory_start.wrapping_add(56);
+        kernel::syscall::InitialProcessLayout {
+            original_break: original_stack_pointer,
+            original_stack_pointer,
+        }
+    }
+
     unsafe fn initialize_process(
         &self,
         stack_pointer: *const usize,
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -1805,8 +1805,6 @@ fn exceeded_check(size: usize, allocated: usize) -> &'static str {
 }
 
 impl<C: 'static + Chip> Process<'_, C> {
-    const INITIAL_APP_MEMORY_SIZE: usize = 3 * 1024;
-
     // Memory offset for callback ring buffer (10 element length).
     const CALLBACK_LEN: usize = 10;
     const CALLBACKS_OFFSET: usize = mem::size_of::<Task>() * Self::CALLBACK_LEN;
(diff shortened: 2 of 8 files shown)


jrvanwhy:
I just looked at `libtock-c`'s entry point and it *appears* that it is compatible with this PR.

kernel/src/process.rs
```diff
@@ -1923,10 +1921,6 @@ impl<C: 'static + Chip> Process<'_, C> {
         let initial_kernel_memory_size =
             grant_ptrs_offset + Self::CALLBACKS_OFFSET + Self::PROCESS_STRUCT_OFFSET;
 
-        if min_app_ram_size < Self::INITIAL_APP_MEMORY_SIZE {
-            min_app_ram_size = Self::INITIAL_APP_MEMORY_SIZE;
-        }
```
bradjc, line 1928: I think we still need this check (or a version of it). If the TBF sets the requested RAM to 0 (or less than arch.syscall says is the minimum), then we might put the stack pointer inside of the kernel reserved memory.

kernel/src/process.rs
```diff
@@ -2034,8 +2027,9 @@ impl<C: 'static + Chip> Process<'_, C> {
         }
 
         // Set the initial process stack and memory to 3072 bytes.
```
bradjc, line 2029:
```suggestion
        // Get the architecture-specific initial process-accessible memory size.
        // Ideally, the kernel intends to not make any assumptions about process memory layout,
        // and leave the setup of the stack, heap, and other data entirely to the process. However,
        // on some architectures context-switching requires using the process's stack, so on those
        // architectures we must have at least a minimal stack setup.
```

kernel/src/syscall.rs
```diff
@@ -389,6 +395,11 @@ pub trait UserspaceKernelBoundary {
     /// initialization must happen in the `initialize_process()` function.
     type StoredState: Default;
 
+    /// Called by the kernel during process creation. Allows for
+    /// architecture-specific process layout decisions, such as stack pointer
+    /// initialization.
+    fn initial_process_layout(&self, process_memory_start: *const u8) -> InitialProcessLayout;
```
bradjc, line 401:
```suggestion
    /// Called by the kernel during process creation. Allows for
    /// architecture-specific process layout decisions, such as stack pointer
    /// initialization.
    ///
    /// Returns two pointers. First, the address where the kernel should set the start of the
    /// process's stack, and second the address where the kernel should set the app_brk.
    fn initial_process_layout(&self, process_memory_start: *const u8) -> (*const u8, *const u8);
```

bradjc (commented):
I think this is the right way to go.

Having Process keep track of the stack pointer at all (beyond the debugging case), however, is at odds with the idea that process's can manage their own memory. Process only tracks the stack pointer for context switches, so really only the arch/syscall code needs to know about the stack pointer. I would like to append to this PR to separate stack pointer handling and make that architecture-specific.

Then, we still need `initial_process_layout`, but all it needs to do is tell the kernel where it should set the app_brk (and therefore the MPU) when running the process for this first time.

arch/cortex-m/src/syscall.rs
```diff
@@ -55,6 +55,23 @@ impl SysCall {
...
 
+    fn initial_process_layout(
+        &self,
+        process_memory_start: *const u8,
+    ) -> kernel::syscall::InitialProcessLayout {
+        // ARM uses the userspace stack to store register values when an
+        // interrupt executes, so we need to give the process an initial usable
+        // stack. We set the initial break at the top of the stack -- processes
+        // must move the break upwards in their runtime implementation.
+        // TODO: 56 is a placeholder -- how large a space do we need to allocate
```
alistair23, line 66: I think without FPU this should be 0x20

jrvanwhy:
> Having Process keep track of the stack pointer at all (beyond the debugging case), however, is at odds with the idea that process's can manage their own memory. Process only tracks the stack pointer for context switches, so really only the arch/syscall code needs to know about the stack pointer. I would like to append to this PR to separate stack pointer handling and make that architecture-specific.

I'm hesitant to make these changes myself, as I do not currently have a way to test Tock kernel changes (especially on ARM!). I gave you access to push commits to this PR, in case you want to make those changes yourself.

kernel/src/process.rs
jrvanwhy: I added a check to verify the initial process memory region and initial kernel memory region do not overlap. I decided to make it an error rather than quietly increasing the process memory region's size for two reasons:

A. This likely indicates something is wrong with the `min_ram_size` calculation.
B. It is very unlikely the process will be happy with only 32 bytes of usable memory, and it would not be able to increase its break.

kernel/src/process.rs
jrvanwhy: Done, thanks!

kernel/src/syscall.rs
jrvanwhy: I used your new comment. Is there a reason you suggested removing the `InitialProcessLayout` struct? Is that to prepare for changes that make this only return the initial break?

arch/cortex-m/src/syscall.rs
jrvanwhy: I set it to `0x20`, thanks.

kernel/src/syscall.rs
bradjc: Oh right I meant to comment on that. I think the return type is fine in general, it just is a little inconsistent with the remainder of the trait, specifically the return value for switch_to_process().

jrvanwhy: Okay, making it more consistent with the rest of the trait seems good to me. I'll go ahead and make that change.

bradjc: There's actually another problem: we can't actually return a pointer from this new init function. We have a bit of a dependency chain: the syscall impl can't setup any process memory until memory has actually been allocated for that process, but Process cannot allocate any memory for a process until it asks the MPU what it can actually protect, and Process cannot call the MPU functions before it knows how much RAM the process needs.

So, Process has to ask syscall for a number of bytes, use that to calculate the requirements for the process, then ask the MPU, then create the actual pointers/addresses to initial process memory.

I'm working on a 1.x implementation to work all of this out.

bradjc:
I have to look through this a lot more and test (hence no PR yet), but here are my changes to remove stack pointer tracking from process.rs: https://github.com/tock/tock/compare/process-no-stack-pointer?expand=1

This has some nice benefits:
- The `Process` struct is smaller, since we do not track the stack pointer and some other state.
- I'm actually checking that the stack pointer is valid before using it in the cortex-m UKB implementation.
- process.restart() is ever so closer to create(), making it easier to refactor Process initialization.

Todo:
- Need to figure out how (or if we still should) to track maximum stack depth in process.rs for debugging.

jrvanwhy:
I believe this is obsoleted by #2383 .

bradjc:
We still need a change for the 2.0 branch, it's just deleting the 3kB code I left in the syscall implementations.

PR #2371, opened 2021-01-20
Migrate the LowLevelDebug capsule to the Tock 2.0 syscall API.

jrvanwhy:
### Testing Strategy

Ran `make prepush`. I don't currently have the ability to test `LowLevelDebug` on hardware, unfortunately.

*I don't see how these changes could go wrong...*

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/low_level_debug/mod.rs
+++ b/capsules/src/low_level_debug/mod.rs
@@ -5,7 +5,7 @@ mod fmt;
 
 use core::cell::Cell;
 use kernel::hil::uart::{Transmit, TransmitClient};
-use kernel::{AppId, Grant, ReturnCode};
+use kernel::{AppId, CommandResult, ErrorCode, Grant, ReturnCode};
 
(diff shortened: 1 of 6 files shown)


capsules/src/low_level_debug/mod.rs
```diff
@@ -5,7 +5,7 @@ mod fmt;
 
 use core::cell::Cell;
 use kernel::hil::uart::{Transmit, TransmitClient};
-use kernel::{AppId, Grant, ReturnCode};
+use kernel::{AppId, CommandResult, ErrorCode, Grant, ReturnCode};
```
alistair23, line 8: Do we still need `ReturnCode`?

hudson-ayers (approved):

capsules/src/low_level_debug/mod.rs
hudson-ayers: It is still used by the UART HIL which this capsule implements

PR #2376, opened 2021-01-22
MicroBit v2 bootloader

alexandruradovici:
### Pull Request Overview

This pull request adds the Tock Bootloader for MicroBit v2.

Kernel and applications can now be programmed using tockloader or flashed using openocd.

The bootloader can be flashed using the MicroBit MSD drive with a simple dran and drop.

### Testing Strategy

This pull request was tested using a MicroBit v2 board.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


lschuermann:
I was actually pretty surprised to see bootloader binaries to be kept in Tock. I don't think that's a particularly great idea, as Git and its workflows were not designed for binary artifacts. I'd much rather host these files somewhere else, although it has to be reliable. I imagine GitHub releases (and in addition some mirrors or content-addressable storage like IPFS) could be a good fit. This is no criticism of this particular PR, but a comment on the current situation.

More worrisome for me is that I can't seem to be able to reproduce this binary with https://github.com/tock/tock-bootloader/pull/21 and latest Tock master. I'm not entirely certain about the state of reproducibility of Tock (and Tock Bootloader), but I'm hesitant to simply trust a binary someone else has put out on the web. Could you post the toolchain version and Tock + Tock bootloader commit you've been building this on so I can confirm the integrity? I think it's crucial to have a few core team members confirm this (along with other contributors such as myself).

I'd be willing to build the required infrastructure for downloading artifacts not included in Tock, if that's something we want as an alternative to binaries in Git, which would include
1. checking if the binary with a pinned hash is present
2. if not, prompting the user whether they want to download it, along with the source URL (list of mirrors), destination path and hash
3. downloading the binary & subsequent integrity check against the pinned hash

alexandruradovici:
> I was actually pretty surprised to see bootloader binaries to be kept in Tock. I don't think that's a particularly great idea, as Git and its workflows were not designed for binary artifacts. I'd much rather host these files somewhere else, although it has to be reliable. I imagine GitHub releases (and in addition some mirrors or content-addressable storage like IPFS) could be a good fit. This is no criticism of this particular PR, but a comment on the current situation.
> 
> More worrisome for me is that I can't seem to be able to reproduce this binary with [tock/tock-bootloader#21](https://github.com/tock/tock-bootloader/pull/21) and latest Tock master. I'm not entirely certain about the state of reproducibility of Tock (and Tock Bootloader), but I'm hesitant to simply trust a binary someone else has put out on the web. Could you post the toolchain version and Tock + Tock bootloader commit you've been building this on so I can confirm the integrity? I think it's crucial to have a few core team members confirm this (along with other contributors such as myself).
> 

I did some small changes, not affecting the bootloader, in the tock-bootloader after writing this binary, I think this is why you cannot reproduce it. I will build a new bin file and add has file next to it, adding the toolchain version and all for it.

> I'd be willing to build the required infrastructure for downloading artifacts not included in Tock, if that's something we want as an alternative to binaries in Git, which would include
> 
>     1. checking if the binary with a pinned hash is present
> 
>     2. if not, prompting the user whether they want to download it, along with the source URL (list of mirrors), destination path and hash
> 
>     3. if user confirmed, downloading the binary & subsequent integrity check against the pinned hash

bradjc:
> I was actually pretty surprised to see bootloader binaries to be kept in Tock. I don't think that's a particularly great idea, as Git and its workflows were not designed for binary artifacts. I'd much rather host these files somewhere else, although it has to be reliable. I imagine GitHub releases (and in addition some mirrors or content-addressable storage like IPFS) could be a good fit. This is no criticism of this particular PR, but a comment on the current situation.
> 
> More worrisome for me is that I can't seem to be able to reproduce this binary with [tock/tock-bootloader#21](https://github.com/tock/tock-bootloader/pull/21) and latest Tock master. I'm not entirely certain about the state of reproducibility of Tock (and Tock Bootloader), but I'm hesitant to simply trust a binary someone else has put out on the web. Could you post the toolchain version and Tock + Tock bootloader commit you've been building this on so I can confirm the integrity? I think it's crucial to have a few core team members confirm this (along with other contributors such as myself).
> 
> I'd be willing to build the required infrastructure for downloading artifacts not included in Tock, if that's something we want as an alternative to binaries in Git, which would include
> 
>     1. checking if the binary with a pinned hash is present
> 
>     2. if not, prompting the user whether they want to download it, along with the source URL (list of mirrors), destination path and hash
> 
>     3. if user confirmed, downloading the binary & subsequent integrity check against the pinned hash

I see why this would be more elegant, but we should also think about our users as well. Many boards come with bootloader firmware which is completely unknown and not even open source. Getting a board running should be as frictionless as possible, because no one wants to get new hardware and then struggle to actually get it to do anything. Committing the binaries is an easy solution that reduces failure points, even if it is harder to track where the binary comes from. I agree we could probably do a better job of specifying exactly how a bootloader was built.

alexandruradovici:
Turns out the bin file was the latest, I the build branches were different. I added a version file with all the information.

lschuermann:
> but we should also think about our users as well. [...] Getting a board running should be as frictionless as possible

I couldn't agree more; part of what got me onto Tock quickly was that it "just worked". Hence I'm _not_ proposing to make the users go to great lengths, such as compiling the bootloader itself or point to some URL to download. Still, I'm not super happy with binary files in the repository. All binary files (images, bootloaders, pdfs) already make up 13MB (>35% of the checked out size) in current master alone, not accounting for the history, are mostly not compressible and will stay in the history forever. I don't think GitHub release artifacts are particularly unreliable and imagine a simple `make flash-bootloader` saying "I need to fetch the binary from one of the following mirrors. Are you okay with that? [Y/n]". I don't want to block this PR, but if this would be a valid strategy, I'd happy introduce the infrastructure for that.

alexandruradovici:
> > but we should also think about our users as well. [...] Getting a board running should be as frictionless as possible
> 
> I couldn't agree more; part of what got me onto Tock quickly was that it "just worked". Hence I'm _not_ proposing to make the users go to great lengths, such as compiling the bootloader itself or point to some URL to download. Still, I'm not super happy with binary files in the repository. All binary files (images, bootloaders, pdfs) already make up 13MB (>35% of the checked out size) in current master alone, not accounting for the history, are mostly not compressible and will stay in the history forever. I don't think GitHub release artifacts are particularly unreliable and imagine a simple `make flash-bootloader` saying "I need to fetch the binary from one of the following mirrors. Are you okay with that? [Y/n]". I don't want to block this PR, but if this would be a valid strategy, I'd happy introduce the infrastructure for that.

What about having a manually started action in `tock-bootloader` that would publish a release of a certain bootloader? I don't think this bootloader will be updated very frequently. I can help set up that action.

boards/microbit_v2/README.md
```diff
@@ -17,7 +17,83 @@ following sensors:
...
+### Using the MicroBit USB Drive
+
+Connect then MicroBit to the computer. A USB drive labeled `MICROBIT` should show up. 
+
+Drag and drop the `bootloader/tock-bootloader.microbit_v2.v1.1.0.bin` to the `MICROBIT` drive and wait for a few seconds.
+
+The board will reset and the bootloader should be running on it. To check whether it's working, press and hold the Button A while pressing the reset button. The Microphone LED should light up.
+
+### Using openocd 
+Use the `make flash-bootloader` command to flash [Tock Bootloader](https://github.com/tock/tock-bootloader) to the board.
```
alistair23, line 39: Can you describe how to get back to the USB drive mode?

boards/microbit_v2/README.md
```diff
@@ -17,7 +17,83 @@ following sensors:
...
+
+```bash
+$ make flash-bootloader
```
alistair23, line 42: Can you also describe how to build the bootloader yourself?

As well as that I'm not sure if we want binary blobs in Tock.

alexandruradovici: > Can you also describe how to build the bootloader yourself?
> 
It never goes away, the USB drive will always appear, it is provided by the debugger not the nrf52 chip.

> As well as that I'm not sure if we want binary blobs in Tock.

I just duplicated Brad's Nano 33 approach :)

boards/microbit_v2/README.md
alexandruradovici: I never goes away as it is provided by the debugger chip.

alistair23: Cool! Maybe just clarify that? I expect some users (I would be) would be worried about breaking their boot loader.

alexandruradovici:
I sent the PR with the release workflow https://github.com/tock/tock-bootloader/pull/22

boards/microbit_v2/README.md
alexandruradovici: Done

boards/microbit_v2/Makefile
```diff
@@ -21,8 +23,16 @@ flash-debug: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
...
-.PHONY: program
-program: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).bin
+.PHONY: flash-app
+flash-app: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).bin
 	arm-none-eabi-objcopy --update-section .apps=$(APP) $(KERNEL) $(KERNEL_WITH_APP)
 	$(OPENOCD) $(OPENOCD_OPTIONS) -c "program $(KERNEL_WITH_APP); verify_image $(KERNEL_WITH_APP); reset; shutdown;"
 
+.PHONY: program
+program: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).bin
+	$(TOCKLOADER) flash $< --page-size 512
```
bradjc, line 33: I think this should specify the address.

boards/microbit_v2/src/main.rs
```diff
@@ -71,7 +71,7 @@ static mut CHIP: Option<&'static nrf52833::chip::NRF52<Nrf52833DefaultPeripheral
 /// Dummy buffer that causes the linker to reserve enough space for the stack.
 #[no_mangle]
 #[link_section = ".stack_buffer"]
-pub static mut STACK_MEMORY: [u8; 0x1500] = [0; 0x1500];
+pub static mut STACK_MEMORY: [u8; 0x2000] = [0; 0x2000];
```
bradjc, line 74: Revert?

boards/microbit_v2/bootloader/version
```diff
@@ -0,0 +1,12 @@
+tock-kernel: 96fceb1a669b7c780b3edf9853882c02e4fcdefb
+tock-bootloader: 885883c08e1099374c532d03c621e3a5a8e2fe29
```
bradjc, line 2: Commenting on the binary: we should probably be as pedantic as we can to not confuse ourselves in the future. If we change the bootloader to support this board we should change the micro number of the version too.

bradjc (changes requested):
We should merge the board to tock-bootloader first.

Otherwise, this looks good to me.

boards/microbit_v2/bootloader/version
alexandruradovici: Are you referring to change the version from 1.1.0 to 1.1.1 ?

boards/microbit_v2/Makefile
alexandruradovici: Doesn't the flash start at offset 0 after the bootloader by default?

boards/microbit_v2/src/main.rs
alexandruradovici: Rebased

boards/microbit_v2/bootloader/version
bradjc: Yes.

boards/microbit_v2/Makefile
bradjc: No, I think tockloader defaults to 0x30000 actually.

boards/microbit_v2/bootloader/version
alexandruradovici: Done

boards/microbit_v2/Makefile
alexandruradovici: Done.

alexandruradovici:
The bootloader is updated to v1.1.1. It has a version file describing the kernel and bootloader versions used. I added two attributes to the bootloader boothash and kernhash.

boards/microbit_v2/README.md
alexandruradovici: Done.

bradjc:
Now that you have done all of this...we arrive at a decision. Our options:

1. Use the serial bootloader.
2. Use openocd but make it work with tockloader better.
3. Combination of 1 and 2. Load the bootloader, but use openocd.

Number 1.
- Pro: The pro for using the bootloader is it is very "tock-y", specifically, we get the bootloader attributes that tockloader can use to customize things.
- Con: the bootloader is much slower than openocd.

Number 2.
- Pro: Fast. Need openocd to load the bootloader anyway.
- Con: You normally have to type `tockloader install --board microbit_v2 --openocd` every time.

Number 3.
- Pro: Get attributes and is fast.
- Con: Might be confusing? tockloader command verbosity still applies.

As for the tockloader con, I've been working on a patch that will enable tockloader to auto detect at least some boards that use Jlinkexe or openocd. For example, on master this now works with the nrf52840dk:

```
tockloader list
[INFO   ] Using settings from KNOWN_BOARDS["nrf52dk"]
┌──────────────────────────────────────────────────┐
│ App 0                                            |
└──────────────────────────────────────────────────┘
  Name:                  hello_loop
  Enabled:               True
  Sticky:                False
  Total Size in Flash:   8192 bytes


[INFO   ] Finished in 1.387 seconds
```

I also flashed the nano33ble bootloader to the nrf52840dk so this works too:

```
tockloader list-attributes
[INFO   ] Using settings from KNOWN_BOARDS["nrf52dk"]
[STATUS ] Listing attributes...
00:    board = nano33ble
01:     arch = cortex-m4
02:  appaddr = 0x50000
03:
04:
05:
06:
07:
08:
09:
10:
11:
12:
13:
14:
15:
[INFO   ] Finished in 0.925 seconds
```

The end result is I think we should encourage users to flash the bootloader, but we should take advantage of the performance of openocd for the microbitv2 since 1) we can, and 2) we need openocd to load the bootloader anyway.

bradjc:
On a related note, can you run

```
$ openocd -c "interface cmsis-dap"
```
with the microbitv2 connected? I need to see the output to make the autodetect feature work for the board.

alexandruradovici:
> On a related note, can you run
> 
> ```
> $ openocd -c "interface cmsis-dap"
> ```
> 
> with the microbitv2 connected? I need to see the output to make the autodetect feature work for the board.

```
Open On-Chip Debugger 0.10.0
Licensed under GNU GPL v2
For bug reports, read
	http://openocd.org/doc/doxygen/bugs.html
Info : CMSIS-DAP: SWD  Supported
Error: CMSIS-DAP: JTAG not supported
```

alexandruradovici:
I fully understand and agree with your point. The bootloader is terribly slow. I do have another use case for it as I am trying to make the microbit tock experience similar to makecode, that is using a browser. I'm not there yet, but the bootloader offers a nice option to flash apps.

I tried the DAPjs that allows flashing using the debugger via WebUSB, but I am not able to flash only apps, so the bootloader seems now the only viable option in this direction.

bradjc:
> I fully understand and agree with your point. The bootloader is terribly slow. I do have another use case for it as I am trying to make the microbit tock experience similar to makecode, that is using a browser. I'm not there yet, but the bootloader offers a nice option to flash apps.
> 
> I tried the DAPjs that allows flashing using the debugger via WebUSB, but I am not able to flash only apps, so the bootloader seems now the only viable option in this direction.

Ok awesome. So having both options seems desirable then. That way if you want to us tockloader directly it can use openocd, but if you want to use the bootloader it will work as well.

Given you are trying to use a browser, what are your thoughts on bootloader entry? Right now it requires a manual button press. Would it be better to make it automatic? If so, any thoughts on how to do that?

Also, thanks for running that command. In the latest tockloader (on master branch) you should be able to run tockloader commands and use openocd without any command line arguments.

bradjc:
Ok one last update to the newest bootloader and lets merge this. We can revisit the bootloader entry question later if so desired.

alexandruradovici:
> Ok one last update to the newest bootloader and lets merge this. We can revisit the bootloader entry question later if so desired.

Should I build one locally or can you use the release action to build a release?

bradjc:
I did a release, is there anything else I'm supposed to do?

alexandruradovici:
The release you did does not seem to have the bin and version files? Did you run the workflow to make the release or did you do it manually?

bradjc:
I did this https://github.com/tock/tock-bootloader/actions/runs/582039351. Honestly, I don't know how this works. Can you add instructions to README.md? I thought this would just work.

hudson-ayers:
I think the point of confusion is that there were two releases. The relevant release with the bin and version files is this one: https://github.com/tock/tock-bootloader/releases/tag/microbit_v2-vv1.1.1

alexandruradovici:
Sorry for the delay, I just catching up with Tock again. I Deleted the binary bootloader and changed the Makefile to directly download the bootloader release.

boards/microbit_v2/Makefile
```diff
@@ -25,8 +27,18 @@ flash-debug: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
 flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
 	$(OPENOCD) $(OPENOCD_OPTIONS) -c "program $<; verify_image $<; reset; shutdown;"
 
-.PHONY: program
-program: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
+.PHONY: flash-app
+flash-app: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
 	arm-none-eabi-objcopy --update-section .apps=$(APP) $(KERNEL) $(KERNEL_WITH_APP)
 	$(OPENOCD) $(OPENOCD_OPTIONS) -c "program $(KERNEL_WITH_APP); verify_image $(KERNEL_WITH_APP); reset; shutdown;"
```
hudson-ayers, line 33: do we still need this command if there is tockloader support? Wouldn't it make more sense to just install apps using tockloader?

boards/microbit_v2/README.md
```diff
@@ -55,7 +132,7 @@ APP=../../../libtock-c/examples/c_hello/build/cortex-m4/cortex-m4.tbf
 KERNEL=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/debug/$(PLATFORM).elf
 KERNEL_WITH_APP=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/debug/$(PLATFORM)-app.elf
 
-.PHONY: program
+.PHONY: flash-app
```
hudson-ayers, line 135: Here also, shouldn't we just refer users to the normal, tockloader based approach for installing apps?

boards/microbit_v2/Makefile
bradjc: I agree, particularly since board autodetection has improved significantly in the next release of tockloader.

alexandruradovici: I modified the Makefile. I am still having trouble using tockloader directly without the serial bootloader. With tockloader 1.6.0 I still have to specify the `--openocd`  parameter and the `-- board` parameter.

boards/microbit_v2/README.md
alexandruradovici: Done

bradjc (approved):

bradjc:
Do you care if we squash these commits? That way bootloader binary isn't added and then removed.

alexandruradovici:
Not at all, go ahead.

PR #1624, opened 2020-02-23
Add an exit syscall (was: Add a syscall to notify the kernel that the app panicked)

torfmaster:
### Pull Request Overview

This pull request is an attempt to address https://github.com/tock/libtock-rs/issues/133 by adding a syscall to allow an app to notify the kernel that an app has panicked. ~The app will not be resumed anymore neither will there be callbacks scheduled~.  The app will be stopped/restarted depending on the first on only parameter of the syscall (0=stop, 1=restart).

### Testing Strategy

This pull request was tested by flashing the `panic` example of https://github.com/torfmaster/libtock-rs/tree/feature/panic-syscall to verify that the syscall is triggered and the app is not resumed.

### TODO or Help Wanted

The implementation is quite raw and there are some questions to be clarified.
Open questions are:
 
 * How to handle panicked apps? I'm in favour of letting the app die and maybe even let the kernel panic because no possibility to recover an app from a "panicked" state than restarting the whole system
 * ~Do we need parameters? I find it hard to squeeze the cause for a panic into some bytes so I would not pass parameters to the syscall~

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make formatall`.

PR #2378, opened 2021-01-23
Add Exit system call class to Tock 2.0.

phil-levis:
### Pull Request Overview

This pull request adds the Exit system call class to Tock 2.0, including `exit-terminate` and `exit-restart`. It refactors some of the process restart code to separate policy from mechanism. It builds on the `yield_2.0` branch/PR #2351 . This borrows heavily from @torfmaster's #1624. 

### Testing Strategy

This pull request was tested by running the `restart` and `terminate` test apps in libtock-c. Both behave as expected, and work correctly with process console.

### TODO or Help Wanted

This pull request doesn't need anything, but process console could probably benefit from a command to reboot a process as well as one to start a terminated process.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


doc/reference/trd-syscalls.md
```diff
@@ -539,8 +587,60 @@ are 12:
...
+the application asks from the kernel. With `exit-terminate`, the application
+tells the kernel that it considers itself completed and does not need to run
+again. With `exit-restart`, it tells the kernel that it would like to be
+rebooted and run again. For example, `exit-terminate` might be used by a 
+process that stores some one-time data on flash, while `exit-restart` might
+be used if the process runs out of memory.
+
+The completion code is an unsigned 32-bit number which indicates status. This
+information can be stored in the kernel and used in management or policy decisions.
+The definition of these status codes is outside the scope of this document.
```
alistair23, line 640: Why not at least say a completion code of 0 is success?

alistair23 (changes requested):

doc/reference/trd-syscalls.md
phil-levis: Because that's an aspect of POSIX semantics which I'm not certain we want to borrow. We very well might, but we haven't fully thought this through. Rather than block this document on a design for process exit codes, I thought it better to just defer to another document. Nothing depends on or uses this yet, so binding it so early seems like a mistake.

alistair23: We should bind it at some point, otherwise different apps are going to return different values and it will become fragmented.

I don't think returning 0 for success is just a POSIX thing, doesn't Windows and everything else do that as well?

Just to clarify I'm not saying specify the meanings, just something along the lines of negative numbers mean failure, 0 means success (or something like that).

alistair23: Also, we could copy the low level debug meanings here and then remove it and replace it with this instead.

doc/reference/trd-syscalls.md
```diff
@@ -200,19 +200,20 @@ are additional error codes to include errors related to userspace.
 | 13    | NOACK       | The packet transmission was sent but not acknowledged.                                  |
 | 1024  | BADRVAL     | The variant of the return value did not match what the system call should return.       |
 
-Any value not specified in the above table is reserved.
+Any value not specified in the above table is reserved and it
+MUST NOT be used.
```
jrvanwhy, line 204: I personally found the wording "Any value not specified in the above table is reserved." a bit ambiguous. The way I read it, it was "reserved for future use" -- a future kernel version could use a different values, and processes must be ready to handle the unexpected value.

The way I read your new wording, you are saying that a Tock 2.0 kernel will *never* return a different value, which allows for a different `libtock-rs` implementation.

I do not expect that you intended this to be a semantics change, because if it is a semantics change it seems orthogonal to this PR. Should this be in a different PR?

jrvanwhy: Oh, this was actually part of #2351 .

doc/reference/trd-syscalls.md
jrvanwhy: I kinda think the low level debug alert codes should have been runtime-specific from the start, so I'm not sure I agree.

phil-levis: I agree we should bind it at some point (even very soon). I just think that this isn't the place to do it. It's going to be part of discussing, designing, and specifying the lifetime of a process in Tock.

doc/reference/trd-syscalls.md
```diff
@@ -624,6 +627,20 @@ The exit identifier specifies which call is invoked.
...
+rebooted and run again. For example, `exit-terminate` might be used by a 
+process that stores some one-time data on flash, while `exit-restart` might
+be used if the process runs out of memory.
+
+The completion code is an unsigned 32-bit number which indicates status. This
+information can be stored in the kernel and used in management or policy decisions.
+The definition of these status codes is outside the scope of this document.
+
+If an exit syscall is successful, it does not return. Therefore, the return
+value of an exit syscall is always _Failure_.
```
jrvanwhy, line 643: If I'm reading the code correctly, `exit-terminate` and `exit-restart` never return regardless of whether they fail. Is that correct?

Do we want to guarantee that behavior for `exit-terminate`? It would make the entry point logic slightly simpler.

phil-levis: `exit-terminate` and `exit-restart` always succeed and so should never return. We should guarantee it for both of them. I think going down the "conditional restart" path is probably not helpful (e.g., "well exit restart only if you'll restart me").

phil-levis: I've updated the document to say that that `exit-restart` and `exit-terminate` MUST succeed (and so never return).

alistair23 (approved):

bradjc:
This PR will be hard to reconcile with master. Can you move the restart function back to where it was (at least in git's view) so the diff is smaller?

phil-levis:
> 
> 
> This PR will be hard to reconcile with master. Can you move the restart function back to where it was (at least in git's view) so the diff is smaller?

I don't think so, because now it's inside the `ProcessType` trait (is invoked from sched.c)? My thought is that this mechanism (the policy is in `try_restart`) should be available through the kernel. The policy in `try_restart` is in response to a userspace request. A command from `pconsole` might be handled differently.

phil-levis:
Let's discuss this on the core call tomorrow? Since it's a major change I'd like a little discussion before merging.

alevy (approved):

phil-levis:
OK, I moved the restart mechanism out of the `ProcessType` trait so that all attempts to restart a process must flow through the policy implemented in `try_restart.` I will think about a later PR that separates this out with the explanation why it is necessary to have extensible policies within the kernel.

bradjc (approved):

doc/reference/trd-syscalls.md
jrvanwhy: Phil, is there a merge or rebase that should be done to remove changes than have already been merged?

PR #2379, opened 2021-01-24
microbit_v2: fix kernel stack overflow

alexandruradovici:
### Pull Request Overview

This pull request fixes a kernel stack overflow for microbit v2.

Release: 0x1000
Debug: 0x2000

### Testing Strategy

This pull request was tested using a MicroBit v2.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/microbit_v2/Makefile
+++ b/boards/microbit_v2/Makefile
@@ -10,8 +10,8 @@ OPENOCD=openocd
 OPENOCD_OPTIONS=-f openocd.cfg
 
 APP=../../../libtock-c/examples/sensors/build/cortex-m4/cortex-m4.tbf
-KERNEL=$(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
-KERNEL_WITH_APP=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/debug/$(PLATFORM)-app.elf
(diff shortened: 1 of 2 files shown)


alexandruradovici:
After comparing branches, but I might be wrong, I think this comes from the new TBF library, that being the only significant change.

hudson-ayers (dismissed):

bradjc:
Nothing about the TBF header code changed, just where it's located.

Needing to make this change is concerning.

alexandruradovici:
This seems to be the only significant change between the last working master that I had this one.

bradjc:
Are you running any userspace apps? The nano33 has a 0x1000 stack, and I just tried the ble test app and that seemed to work.

If we need to make the stack bigger that is fine, but if there is a underlying bug it would be better to find that first.

alexandruradovici:
> Are you running any userspace apps? The nano33 has a 0x1000 stack, and I just tried the ble test app and that seemed to work.
> 

kernel (release) - works
kernel (debug) - faults
kernel + apps (release) - faults
kernel + app (debug) - faults

> If we need to make the stack bigger that is fine, but if there is a underlying bug it would be better to find that first.

I'll try to start it with less drivers and see what happens.

hudson-ayers (approved):
That makes sense -- I suspect many boards in Tock will require more stack space when debug mode is used.

bradjc (approved):

Issue #2366, opened 2021-01-19
vDSO style syscalls


alistair23:
Many modern operations systems use something similar to Linux's [vDSO](https://man7.org/linux/man-pages/man7/vdso.7.html) style syscalls.

That is for some common syscall operations instead of requiring userspace to trap into the kernel, which can take a long time, they can just read the value from memory.

The Linux kernel documenation describes this well:

```
       Why does the vDSO exist at all?  There are some system calls the
       kernel provides that user-space code ends up using frequently, to
       the point that such calls can dominate overall performance.  This
       is due both to the frequency of the call as well as the context-
       switch overhead that results from exiting user space and entering
       the kernel.
```

I think it's worth adding support to Tock for vDSO style syscalls. The general idea being:

 1. All applications are made aware of a memory address used for vDSO (where should this be?)
 2. When the kernel returns to an app it writes all of the configured values to the vDSO memory location
 3. The applications libtock library can then check the existence of vDSO values before performing a syscall. This should be handled in the library, not by the users application code.

The values in memory can be stored in simple key/value pairs, where the key matches with the existing syscall values. That way an application can easily determine what vDSO syscalls are supported.

Two good vDSO syscall candidates would be the current time and are any callbacks pending. Capsules can add their own vDSO syscalls as well. For example maybe the ADC capsule allows an application to setup the ADC value in a vDSO syscall, then allowing the application to read the value without trapping.

bradjc:
This could be global or per-process. Global would be simpler to setup, but putting the memory location inside of the process's already accessible memory would be easier to make work with the MPU.

bradjc:
> For example maybe the ADC capsule allows an application to setup the ADC value in a vDSO syscall, then allowing the application to read the value without trapping.

I think this is the same thing as `allow`.

alistair23:
I think this would need to be local, to ensure we don't leak information. We could require the app to call `allow` to get a region of memory or always use a constant offset from the start of the apps memory.

> > For example maybe the ADC capsule allows an application to setup the ADC value in a vDSO syscall, then allowing the application to read the value without trapping.
> 
> I think this is the same thing as `allow`.

Yes, this is pretty much the same as allow without a callback. There might be some occasions where we want this to update, even if an interrupt hasn't triggered (such as writing the current time) in which case it ends up a little different to allow.

lschuermann:
[...]

PR #2381, opened 2021-01-25
kernel: Read Only State

alistair23:
### Pull Request Overview

Based on the proposal here: https://github.com/tock/tock/issues/2366 this is a vDSO style read only syscall where userspace can quickly read values from memory instead of performing an entire syscall.

This allows user space to implement a `quick_yield()` where we only perform a syscall if there is a pending task. This also allows userspace to easily read the current time without performing a syscall.

In order to allow ROS, this PR also adds shared allow syscalls, based on https://github.com/tock/tock/pull/2590#issuecomment-854259226.

Some of the TRD update is borrowed from: https://github.com/tock/tock/blob/trd104-buffer-reading/doc/reference/trd104-syscalls.md

The idea of these syscalls is to handle capsules that expect userspce to access a buffer while the kernel has access.

Add a shared allow that is similar to read/write allow except that the app is expected to be able to access the buffer.

This is different to allow_readwrite() in that the app is allowed to read/write the buffer once it has been passed to the kernel.
For more details on how this can be done safely see TRD104.

### Testing Strategy

Tested on the nRF with libtock-c.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


phil-levis:
Do we have any use cases when this has been a problem?

alistair23:
Yes. If you are doing work in a loop and want to check if you should break out of the loop.

It's similar to the original yield-no-wait problem/solution, except in a tight loop yield-no-wait has too much overhead as you have to enter the kernel on every loop. For a small loop you can spend more time in the context switch then actually doing work.

alexandruradovici:
What about making this more generic, like for instance allow other drivers to register values to this ROSDriver?

The required buffer would be something like:

```
|-------------------------|
|   Item 1                 |
|-------------------------|
|                              |
|     Item2                |
|-------------------------|
| ...                          |
```

Each registered item would have a unique id (driver_id / minor?). We would have another command system call that would get the position and size of the item in the buffer based on the unique id.

I can draft the changes for this.

alistair23:
I thought about using a driver id with the value, the thought there was to make this more of a key/value pair to indicate different information and then the data could be more dynamic. Possibly even updated from other capsules.

At that point though it probably makes more sense to just add support in the other capsules. For example instead of indicating that an alarm has expired here, the Alarm capsule could be extended to have an extra `allow` call similar to this to pass information back.

My thinking with this driver was to support only with system information (I count current time as that) and other capsule specific information could be handled by their capsules.

alexandruradovici:
> I thought about using a driver id with the value, the thought there was to make this more of a key/value pair to indicate different information and then the data could be more dynamic. Possibly even updated from other capsules.
> 
> At that point though it probably makes more sense to just add support in the other capsules. For example instead of indicating that an alarm has expired here, the Alarm capsule could be extended to have an extra `allow` call similar to this to pass information back.
> 

I thought about this, but there is a possible race problem. The app could be preempted right in the middle of a read from the allowed buffer. Unless data is u32, this might be a problem. I think this applies to your timer ticks implementation that is u64.

What about adding a u32 "version" field that gets incremented after every write. The app could than run:

```c
do
{
  int version = read_version ();
  // read the actual data
} while (version != read_version())
```
 
> My thinking with this driver was to support only with system information (I count current time as that) and other capsule specific information could be handled by their capsules.

alistair23:
Good point about the possible race condition of a 64-bit value. That is a risk.

I would prefer to leave it as a 64-bit time because I feel that a 32-bit timer is not long enough.

I have changed the userspace implementation to read the high/low values in a loop to hopefully avoid atomic issues.

alistair23:
I have pushed a userspace implementation: https://github.com/tock/libtock-c/pull/170

lschuermann:
While I do think vDSO-style system calls could be pretty useful (for instance, I constantly need the current time in a userspace network stack), I'm not quite sure about the additional complexity & kernel code introduced by this particular implementation strategy. Is there a particular reason why this can't be a simple extension to a `Driver` trait (something like `ProactiveDriver: Driver`), which is called prior to switching to a particular process, and uses simple `AppSlice`s for that? I can write up a small PoC.

This way, we don't introduce a lot of specialized infrastructure and have proper separation of concern by letting each individual driver manage it's proactive memory region shared with a particular process. With something like #2391 we can easily add a method that iterates over all `ProactiveDriver`s on a platform and invokes them with the AppId being switched to.

alistair23:
Just to make sure I understand, you want to extend every capsule with a `ProactiveDriver` and iterate over all implementations of `ProactiveDriver` when switching to an application?

I'm not completely opposed to that, but I don't see the benefit. There are only a handful of event types that aren't interrupt driver (like pending syscalls or ticks) and I think they should be implemented in ROS. Everything else can be handled by the specific capsule updating the allow memory when an interrupt occurs.

lschuermann:
This is just a possible implementation strategy, but it might be good to think about the different options, hence mentioning it.

> you want to extend every capsule with a `ProactiveDriver`

I want to extend every capsule to be a `ProactiveDriver` (make `Driver` a supertrait of `ProactiveDriver`) which may want / need to communicate values proactively to userspace upon a context switch. For example, `Alarm` would be one such driver. A method such as `switch_process(&self, process: AppId)` would be invoked and the regular mechanisms of a driver to convey information to userspace would be used. The platform would feature an `iter_proactive_drivers<F: Fn(&dyn ProactiveDriver)>(&self, f: F)` method to retrieve the proactive drivers for a given platform.

> I also don't understand why that is less complex then this implementation

"complex" may have been the wrong expression. It's technically not too different, however I'd avoid adding code to the kernel crate when it can be mostly implemented using the regular capsule mechanisms. I'd also be wary to add a distinct special ROS mechanism when we can reuse existing infrastructure by slightly extending the scope of what a capsule can do (optionally, that is!). Furthermore I like the idea of having separation of concerns here, such that a capsule that is responsible for conveying the information via regular system calls to userspace anyways would also be responsible for the vDSO-style of system calls.

alistair23:
How would you communicate pending syscalls to processes with the `ProactiveDriver`?

alistair23:
Any more comments on this?

alistair23:
Ping!

alistair23:
Rebased on 2.0, this is ready to merge.

alistair23:
Ping!

lschuermann:
Sorry for being unresponsive w.r.t. this discussion for some time. Looking at the design as it is now, I'm still not convinced that this approach is the right way to go. I do think that "read-only" / vDSO-style system calls are a great way to communicate certain information to userspace, but to limit it to this one `ROSDriver` which really only sets the current time is quite restricting.

As a concrete example, I have an application transmitting and receiving Ethernet frames. I'd very much like the driver to provide my application interface statistics without requiring use of an explicit system call for this.

I don't think the Tock kernel should be opinionated about what types of information a userspace app on a specific board may need, and restrict introduced mechanisms to this. While the mechanism in general makes sense, a board should be able to hook custom data sources into this mechanism.

I see two possibilities for implementing this:
- either change ROSDriver to not be a `struct`, but a `trait ROSDriver: Driver` (which also requires the `update_values` method), such that a board at the very least has the option to convey custom values through this mechanism

  We can then just ship the `ROSTimeDriver` as a capsule, implementing this trait.
- or allow capsules to statically or dynamically hook into this, by - for instance - a ROSDriver taking a slice of capsule references, and calling `update_values` on each of them.

I have a preference for the latter, because it makes it the capsule's responsibilities to communicate the data in the way they see fit (given that they already control the capsule ABI, which would be the alternative way to get the information). The first approach would allow me to define my own behavior on out-of-tree boards, and hence would technically satisfy my requirements as well.

alistair23:
For your example for Ethernet statistics isn't that available via a standard `allow` syscall? Then the statistics can just be updated when Ethernet packets are received/transmitted?

I don't think a board should be able to change this. If every board has it's own implementation then apps can't be built to run on any Tock board as the kernel interface is now board dependent. That seems against the general design of Tock and makes the userspace code more confusing.

Capsules can hook into this by just creating a read only `allow` syscall. The only thing missing is a callback. Adding a general callback is something that could be added in future if required. Although generally I expect that the values would just be updated when events occur (via interrupts) in the capsule. There are only a handful of exceptions to this, like the current time and kernel statistics.

alistair23:
Any more thoughts or comments?

lschuermann:
> For your example for Ethernet statistics isn't that available via a standard `allow` syscall? Then the statistics can just be updated when Ethernet packets are received/transmitted?

For my homebrew implementation, I'm querying them explicitly from the device. I'd like to only do this when the relevant app is scheduled. I imagine there are more scenarios where this would come in handy.

> I don't think a board should be able to change this. If every board has it's own implementation then apps can't be built to run on any Tock board as the kernel interface is now board dependent. That seems against the general design of Tock and makes the userspace code more confusing.

Fair, I guess? As a board implementer I'll tailor the kernel to my needs and as such I can change the kernel to my liking. I don't think that should be the general spirit though, especially if there is an arguably good way to give users more freedom while not breaking or complicating the vanilla Tock boards/ecosystem.

Also, I'm strictly against making the implementation _custom_ for every board. I'm just saying that ultimately, the driver that should decide over the contents and the semantics of the "read only system call" should be the driver usually managing this peripheral (e.g. for the current system time, the alarm capsule). If it can hook into this system (statically, at compile time), it can have it's very own allow-region and set the current system time, just as you do now. If I now have a second driver with a usecase for this, I can hook it into the system as well, and it'll get a call just before switching to an app to do it's business. Those two will be independent of each other, and changing / excluding one interface won't break another driver. It mirrors the responsibilities of the capsules & userspace libtock-c implementations only this subsystem 1-to-1. If I change a driver (including the ROS system of that driver, if it has one), I'm going to have to adjust the userspace driver as well.

Also, I'm not worried about performance either. I don't want to implement this mapping dynamically, but rather explicitly choose the drivers which should be hooked into the "read only system call" kernel subsystem in the board crate, and pass a static slice of these drivers in the kernel. In the common case, it'll work just as performant as the current proposal.

alistair23:
Ping!

@lebakassemmerl after our discussion do you have any more thoughts on this?

alistair23:
Ping!

alistair23:
Ping!

phil-levis:
> Good point about the possible race condition of a 64-bit value. That is a risk.
> 
> I would prefer to leave it as a 64-bit time because I feel that a 32-bit timer is not long enough.
> 
> I have changed the userspace implementation to read the high/low values in a loop to hopefully avoid atomic issues.

This will avoid but not solve the problem. The read high/low problem works for roll-over but not atomicity.

alistair23:
> > Good point about the possible race condition of a 64-bit value. That is a risk.
> > I would prefer to leave it as a 64-bit time because I feel that a 32-bit timer is not long enough.
> > I have changed the userspace implementation to read the high/low values in a loop to hopefully avoid atomic issues.
> 
> This will avoid but not solve the problem. The read high/low approach works for roll-over but not atomicity.

Can you point out why it doesn't solve the problem?

1. Userspace reads high timer value
2. An interrupt occurs and userspace stops running
3. Eventually userspace is resumed and now reads the low value
4. Userspace checks to make sure the high value it read matches the current high value. If they match all good, if not repear the process

1. Userspace reads high and low value
2. An interrupt occurs and userspace stops running
3. Eventually userspace is resumed and now compares high to value it has with current high
4. If they match userspace could still have an outdated time, but a u32 would also have the same problem here.

phil-levis:
Sorry, I was too terse.

The problem is that high/low works fine for rollover on a counter but isn't generalizable to consistent reads of shared variables. This is presented as a generic mechanism, but is really only feasible for counters. One approach is to include a version number at the head of the region: you can use this in the same way you do high for 64-bit reads. But requiring each data item to include its own atomicity mechanisms seems problematic to me.

phil-levis:
If we move forward with this, we will need to change the tentative text for TRD 104, which lays out rules for when an allowed buffer can be read from userspace.

alistair23:
Ah, ok.

So if we add more fields that require atomicity we can add a counter region as well like you are describing. Currently I can't forsee a use case of where we will need it. Indiviual regions don't need to be atomic, for example it doesn't matter if you read the time then receive an interrupt then read the number of pending calls.

So we would only hit an issue when we need to read values larger then 32-bits that also aren't counters, which seems unlikely.

As for the TRD change I think https://github.com/tock/tock/pull/2590 should cover it

phil-levis:
> As for the TRD change I think #2590 should cover it

I think it will explicitly not allow this. For userspace to be able to read an allowed buffer it must be not in use, which means the kernel can't write to it. I think the text doesn't say that explicitly, but it's intended (otherwise the state machine is meaningless).

It means the 104 text will need to say that it's possible for userspace to read from a buffer while it is in use, but there must be mechanisms defined in the system call TRD that specify how you do so atomically. It also means that it should be possible for a buffer to start (and stay) in the in use state.

Alternatively, we can make a specialized carve-out in 104, or say that other TRDs can specify other mechanisms.

alistair23:
Userspace can always read a buffer, so it will have to be possible. I agree that userspace reading a buffer in use should be discouraged (but not enforced as we can't) as otherwise the data could be incomplete.

In the case of ROS though whenever userspace is running the buffers are not in use (in the sense that the kernel will never be mid-operation). It is not that dissimilar to an app reading a buffer after the complete callback.

alistair23:
@lschuermann and @phil-levis this now uses the new shared allow syscall

lschuermann:
I have taken a more in-depth look into the code and I'd say that in general, this looks good.

However, I'm still wary of the way the kernel is now involved with the time and alarm subsystem as presented to userspace. I agree that there are only limited use cases for the ROS mechanism as introduced here. Nonetheless, I think that the job of the core kernel is to, for these extraordinary use cases, provide hooks such that they can be built if so desired. I'm of the strong opinion that the core kernel should not be involved in providing these mechanisms directly. I don't think that having the kernel involved with the time and alarm subsystem (apart from purely kernel-internal infrastructure, such as scheduling) is a particular good design, especially when it comes to separation of concerns and presenting a unified interface to userspace. With this PR, there would now be two system call drivers communicating time and alarm information to userspace. To me, this is emphasized by the fact that we now _need_ some `T: Time` to construct the kernel, even if we don't use ROS.

Instead, I think that this mechanism is much better incorporated with the existing system call drivers. The alarm capsule provides a unified interface for all things time & clock related, and as such can incorporate the ROS mechanism for communicating the current clock ticks to userspace. Furthermore, by using the same driver, userspace can assume that querying the alarm driver through a regular Command system call, or through vDSO style system calls refers to the same timebase. When we have two drivers, a board author can easily hook up different clocks to each of them.

I'm sufficiently convinced that the use cases for this are incredibly narrow as to not require multiple ROS implementations. Still, for the reasons outlined above, especially to keep the core kernel's involvement with the time / alarm subsystem minimal, I'm of the opinion that a trait-based ROSDriver is better suited here. As a bonus -- if board authors so desire and are aware of the performance penalties -- they _could_ do a manual fanout of ROSDriver calls to multiple subsystems.

To make sure we're on the same page, I've [taken a stab at implementing this here](https://github.com/lschuermann/tock/commit/8fd3d7873824ae8a127ddd2c546ee52bddffbb4d). It's very similar to what this PR proposes. It doesn't add any flash size overhead (actually is slightly smaller) when disabled. Both mechanisms currently have in common that they add some flash overhead even if they're disabled. With the trait based approach, it should be relatively easy to avoid inclusion of this in the binary at all with some clever generics tricks.

alistair23:
I see lots of issues with that implementation.

First like I have said before, I don't see any reason why we should allow the ROS area to be flexible. Allowing board authors to change the region goes against the general design of Tock that an app can run on any board. By allowing boards to change the regions ROS now becomes a board specific format. You describe that as a bonus but I see that as a disadvantage.

Secondly each capsule now needs to implement it's own count and own format. That increases the overhead of the implementation in terms of buffer size and operations performed on context switch.

Your implementation also doesn't have the pending tasks call, so there is going to have to be some core implementation of the ROS trait anyway.

I agree that not requiring `T: Time` to create the kernel with the trait method is an advantage, but all Tock boards must have an alarm anyway, so I don't see it as too much of a burden.

> Furthermore, by using the same driver, userspace can assume that querying the alarm driver through a regular Command system call, or through vDSO style system calls refers to the same timebase. When we have two drivers, a board author can easily hook up different clocks to each of them.

+1 That is a good point, but I don't think it outweighs the disadvantages above.

Although your proposal is more flexible, I don't think we should support that sort of flexibility. Although capsules can already DOS the app, I don't think we should allow them to execute on every context switch as controlled by the board.

lschuermann:
> I don't see any reason why we should allow the ROS area to be flexible.

My point is that there is no "_the_ ROS area". In my example, alarm implements the ROS mechanism and hence incorporates _an_ ROS-style buffer as part of its set of Allow-buffers.

> Secondly each capsule now needs to implement it's own count and own format.

Correct, but given we probably agree on the use-cases of this being very narrow, that should only have a single implementation in upstream Tock anyways?

> Your implementation also doesn't have the pending tasks call, so there is going to have to be some core implementation of the ROS trait anyway.

I don't think I understand the utility or the motivation for this anyways. Why does userspace suddenly have access to this? Can it recover this information without using ROS / vDSO semantics? I'd be wary to supply such information to userspace without thinking about the consequences this has for future (re)architectures of the kernel. I don't see this discussed here yet, it's something worth to put some thoughts into.

---

I don't have much more to bring to the table. I don't particularly like the design with a single buffer / region, for the (IMHO) valid reasons outlined above. On the other hand, there admittedly are some benefits; for example when it comes to including diverse data sources in this single buffer (as we then only have to access a single Grant).

I'm going to let others decide on what approach is better suited for Tock. Still, even if we have a single `ROSDriver` as in your proposal, please use a trait to hook it in the kernel and put it into `capsules`. This at the very least relaxes the requirement for `T: Time` on the kernel, as we can pass a `()` (unit) type to disable the mechanism and avoids compiling the types into the kernel (because, even if you pass a `None`, this `Option` is still generic over `ROSDriver`).

alistair23:
Any comments from anyone else?

phil-levis:
I will take a look.

phil-levis:
I think I've said before, we should separate this from 104? Two reasons: it would be a good example of how to write a TRD to extend the system calls, and it also doesn't block 104's finalization on work/experience with this system call. Note how 104 actually saw some significant changes after many months, once we had more experience.

phil-levis (changes requested):
Overall, I think this decomposition is good.

I think we should decouple this from 104. It can be a draft that becomes 105. That will give us time to play with it and get experience before finalizing it. We should add this right after 2.0.

I think that the documentation needs to explain more. E.g., RISC-V shows actual sample assembly code on how to deal with wraparound of mtimer on a 32-bit architecture. Similarly, the 10001 should show sample code of how to safely access this shared state. But this is just documentation and can be easy/quick to fix.

doc/reference/trd105-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
+Shared Allow System Call
+========================================
+
+**TRD:** 105 <br/>
```
phil-levis, line 4: Technically, we don't assign a number until it's in its final stages (to avoid fragmentation). So you can just call  this trd-shared-allow-syscalls.md.

doc/reference/trd105-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+presence of arbitrary userspace reads is too great a programming
+burden for an unintended use case.
+
+However, there can be cases when it is necessary for userspace to be
+able to read a buffer without first revoking it from the kernel with a
+Read-Write Allow. These cases are situations when the cost of a
+Read-Write Allow system call is an unacceptable overhead for
+accessing the data.
+
+This allows capsules that support the shared allow call to communicte with
```
phil-levis, line 40: communicate

doc/syscalls/10001_ros.md
```diff
@@ -0,0 +1,85 @@
...
+  |-------------------------|
+  |                         |
+  |     Time Ticks (u64)    |
+  |-------------------------|
+
+Example C code to safely read the 64-bit timer value is included below,
+where `ptr` is a `uint32_t*` pointing to the ROS memory region.
+
+```c
+  uint32_t high, low;
```
phil-levis, line 39: I don't think this code is correct. You need a volatile.

doc/reference/trd105-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+=================================
+
+2.1 Shared Allow (Class ID: 7)
+---------------------------------
+The shared allow syscall follows the same expectations and requirements
+as described for the Read-Write syscall in section TRD104 4.4, with the
+exceptionthat apps are explicity allowed to read/write buffers that have
+been passed to the kernel.
+
+The standard calling pattern for reading data from the Tock kernel is to
```
phil-levis, line 57: Why is this text here? Explaning the standard ReadWriteAllow approach probably isn't needed given this is a new allow.

doc/reference/trd105-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+  1. use `subscribe` to register a callback,
+  2. make a Read-Write Allow call to share a buffer with the kernel,
+  3. call a `command` to start an operation that writes the allowed buffer,
+  4. in the upcall that signals the operation completes, make another
+  Read-Write Allow call to reclaim the buffer shared with the kernel.
+
+As simultaneous accesses to a buffer from both userspace and the kernel can
```
phil-levis, line 64: It would be good to explain these issues. Pro-actively explaining what can go wrong will save us lots of headaches trying to explain this to authors later.

doc/reference/trd105-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+cause issues if not implemented properly, each capsule using the Shared Allow
+mechanism MUST document, in a Draft or Final Documentary TRD, how consistency
+of the data exchanged through this shared memory region is achieved.
+Examples for mechanisms to ensure consistent data reads/writes can be driver state
+machines, precisely specifying when writes will be issued to the buffer by the
```
phil-levis, line 69: Fleshing out examples of each of these would be good. A monotonic counter, for example, can be pulled from the specific libtock example below.

phil-levis (changes requested):
This is proposing adding a new system call class; we want to explain it fully and guide users on how and when it can be safely used, so I think the TRD needs to be fleshed out a bit. Once that's done and 2.0 is released, I think this is about ready for an initial merge, so we can use it and get to know it better before finalizing it.

alistair23:
Now that 2.0 is out this is ready to merge

alistair23:
I have updated this to not change the `kernel_loop()` args and instead exposes a trait via `KernelResources`.

That should address your concerns @lschuermann

doc/reference/trd-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+2.1 Shared Allow (Class ID: 7)
+---------------------------------
+The shared allow syscall follows the same expectations and requirements
+as described for the Read-Write syscall in section TRD104 4.4, with the
+exceptionthat apps are explicity allowed to read/write buffers that have
+been passed to the kernel.
+
+As simultaneous accesses to a buffer from both userspace and the kernel can
+cause issues if not implemented properly. For example a userspace app could
+read partially written data or the kernel could act on partially written data.
```
phil-levis, line 59: This suggests that it allows userspace to write to the buffer?

doc/reference/trd-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+each capsule using the Shared Allow mechanism MUST document, in a Draft or
+Final Documentary TRD, how consistency of the data exchanged through this
+shared memory region is achieved.
+
+Examples for mechanisms to ensure consistent data reads/writes can be driver state
+machines, precisely specifying when writes will be issued to the buffer by the
+kernel, or monotonically increasing counters along the data for userspace to ensure
+a read operation yielded consistent data.
+
+For an example of a monotomic counter implementation look at the
```
phil-levis, line 70: This document should walk through the examples and how to write them.

doc/reference/trd-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+`doc/syscalls/10001_ros.md` description.
+
+3 libtock-c Userspace Library Methods
```
phil-levis, line 73: Specify the allowed arguments and return types. This should be a full specification of the system call.

doc/reference/trd-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+been passed to the kernel.
+
+As simultaneous accesses to a buffer from both userspace and the kernel can
+cause issues if not implemented properly. For example a userspace app could
+read partially written data or the kernel could act on partially written data.
+This would result in obscure timing bugs that are hard to detect. Due to this
+each capsule using the Shared Allow mechanism MUST document, in a Draft or
+Final Documentary TRD, how consistency of the data exchanged through this
+shared memory region is achieved.
+
```
phil-levis, line 64: This needs to say exactly what can be done with this system call, and can't. I.e., is it that userspace MAY read the contents of a buffer shared, but MUST NOT read to it?

phil-levis (changes requested):
Please fill in the details on the semantics of the system call, what uses are permissible (and not), the call arguments and return types, and consistency mechanisms.

doc/reference/trd-shared-allow-syscalls.md
alistair23: That was the intent

phil-levis: Moved the discussion to #2803 . I thought we had agreed on a userspace-readable/kernel-writeable buffer for your use case, not arbitrary shared memory.

capsules/src/ros.rs
```diff
@@ -0,0 +1,142 @@
...
+//!
+//! The layout of the read only syscalls in the allow region depends on the
+//! version. Userspace can use `command 0` to get the version information.
+//!
+//! Versions are backwards compatible, that is new versions will only add
+//! fields, not remove existing ones or change the order.
+//!
+//! Version 1:
+//!   |-------------------------|
+//!   |       Count (u32)       |
```
ppannuto, line 21: Count of what?

capsules/src/ros.rs
```diff
@@ -0,0 +1,142 @@
...
+use kernel::process::{self, ProcessId};
+use kernel::processbuffer::{SharedProcessBuffer, WriteableProcessBuffer};
+use kernel::syscall::{CommandReturn, SyscallDriver};
+use kernel::ErrorCode;
+
+/// Syscall driver number.
+pub const DRIVER_NUM: usize = 0x10001;
+const VERSION: u32 = 1;
+
+pub struct ROSDriver<'a, T: Time> {
```
ppannuto, line 43:
```suggestion
pub struct ReadOnlySyscallDriver<'a, T: Time> {
```

This is a partially symptom of a different project I'm working on, but I can't think of anything but [robots](https://www.ros.org/) when I see ROS. I think it's worth the characters / clarity to just spell this out everywhere.

capsules/src/ros.rs
```diff
@@ -0,0 +1,142 @@
...
+                    }
+                    if buf.len() >= 16 {
+                        let now = self.timer.now().into_usize() as u64;
+                        buf[8..16].copy_from_slice(&now.to_le_bytes());
+                    }
+                });
+            })
+            .unwrap();
+
+        self.count.set(count.wrapping_add(1));
```
ppannuto, line 83: Is this a duplicate of the count that's in the process debug info?

doc/reference/trd-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
+Shared Allow System Call
+========================================
+
+**TRD:** 105 <br/>
```
ppannuto, line 4:
```suggestion
**TRD:** XXX <br/>
```

doc/reference/trd-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+**Draft-Discuss:** tock-dev@googlegroups.com</br>
+
+Abstract
+-------------------------------
+
+This document describes the shared allow system call application binary
+interface (ABI) between user space processes and the Tock kernel for 32-bit
+ARM Cortex-M and RISC-V RV32I platforms.
+
+This is an extension on the allow calls specified in TRD 105.
```
ppannuto, line 21:
```suggestion
This is an extension on the allow calls specified in [TRD 104](trd104-syscalls.md).
```

doc/reference/trd-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+presence of arbitrary userspace reads is too great a programming
+burden for an unintended use case.
+
+However, there can be cases when it is necessary for userspace to be
+able to read a buffer without first revoking it from the kernel with a
+Read-Write Allow. These cases are situations when the cost of a
+Read-Write Allow system call is an unacceptable overhead for
+accessing the data.
+
+This allows capsules that support the shared allow call to communicate with
```
ppannuto, line 40:
```suggestion
Instead, capsules that support the shared allow call can communicate with
```

doc/reference/trd-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+write statistics to the buffer and at any time the app can read the statistics
+from the buffer.
+
+2 System Call API
+=================================
+
+2.1 Shared Allow (Class ID: 7)
+---------------------------------
+The shared allow syscall follows the same expectations and requirements
+as described for the Read-Write syscall in section TRD104 4.4, with the
```
ppannuto, line 53:
```suggestion
as described for the Read-Write syscall in [TRD104 Section 4.4](trd104-syscalls.md#44-read-write-allow-class-id-3), with the
```

doc/reference/trd-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+exceptionthat apps are explicity allowed to read/write buffers that have
```
ppannuto, line 54:
```suggestion
exception that apps are explicitly allowed to read/write buffers that have
```

doc/reference/trd-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+been passed to the kernel.
+
+As simultaneous accesses to a buffer from both userspace and the kernel can
```
ppannuto, line 57:
```suggestion
Simultaneous accesses to a buffer from both userspace and the kernel can
```

doc/reference/trd-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+cause issues if not implemented properly. For example a userspace app could
+read partially written data or the kernel could act on partially written data.
+This would result in obscure timing bugs that are hard to detect. Due to this
+each capsule using the Shared Allow mechanism MUST document, in a Draft or
+Final Documentary TRD, how consistency of the data exchanged through this
+shared memory region is achieved.
+
+Examples for mechanisms to ensure consistent data reads/writes can be driver state
+machines, precisely specifying when writes will be issued to the buffer by the
+kernel, or monotonically increasing counters along the data for userspace to ensure
```
ppannuto, line 67:
```suggestion
kernel, or monotonically increasing counters along with the data for userspace to ensure
```

doc/reference/trd-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+
+```c
+typedef struct {
+  bool success;
+  void* ptr;
+  size_t size;
+  tock_error_t error;
+} allow_shared_return_t;
+
+allow_shared_return_t allow_shared(uint32_t driver, uint32_t allow, void* ptr, size_t size);
```
ppannuto, line 89:
```suggestion
read_write_allow_return_t read_write_allow(uint32_t driver, uint32_t allow, void* ptr, size_t size);
```

?

It seems odd to me for the syscall and the interface to the syscall to use different names. That might invite confusion. I'm reasonably sure that our syscalls and interfaces use consistent names elsewhere.

doc/reference/trd-shared-allow-syscalls.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+The `success` field indicates whether the call succeeded.
+If it failed, the error code is stored in `error`. If it succeeded,
+the value in `error` is undefined. `ptr` and `size` contain the pointer
+and size of the passed buffer.
```
ppannuto, line 95: This should also explain what the `allow` parameter is [even if it's just a duplication of some of the 104 §4.4 text, locality of reference is valuable here]

doc/syscalls/10001_ros.md
```diff
@@ -0,0 +1,85 @@
+---
+driver number: 0x10001
+---
+
+# Read Only System
```
ppannuto, line 5:
```suggestion
# Read Only Syscall
```
?

doc/syscalls/10001_ros.md
```diff
@@ -0,0 +1,85 @@
...
+  volatile uint32_t high, low;
+
+  do {
+    // Set the high bytes the value in memory
+    high = ptr[3];
+    // Read the low bytes
+    low = ptr[2];
+    // If the high bytes don't match what is still in memory re-try
+    // the load
+  } while (high != ptr[3]);
```
ppannuto, line 48: What are "high" and "low" here? Can these names match the names in the ASCII art block above?

doc/syscalls/10001_ros.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Data Consistency
+
+Before returning from the kernel to userspace all ROS fields will be updated.
+This ensures that at all times when an application is running, the ROS fields
+are valid and consistent.
+
+If userspace wants to read multiple values and ensure consistency between reads
+the `count` field should be used to ensure that the kernel didn't run between
+reads.
```
ppannuto, line 59: Is `count` updated on every context switch, or only when the actual contents of the ROS data structure update? (this may answer my 'duplicate of debug info' question above)

doc/syscalls/10001_ros.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Allow Shared
+
+  * Description allow_shared() is used to setup the read only buffer:
+
+  * ### Allow Read/Write Number: 0
+
+    **Description**: Memory Buffer.
+
+    **Argument 1**: Slice into which the read only syscal values should be stored
```
ppannuto, line 69:
```suggestion
    **Argument 1**: Slice into which the read only syscall values should be stored.
```

doc/syscalls/10001_ros.md
```diff
@@ -0,0 +1,85 @@
...
+
+  * ### Command Number: 0
+
+    **Description**: Get Version.
+
+    **Argument 1**: Unused
+
+    **Argument 2**: Unused
+
+    **Returns**: The latest version of ROS supported by the kernel.
```
ppannuto, line 85:
```suggestion
    **Returns**: The latest version of ROS supported by the kernel.
```

The latest version, or the version currently in use?

capsules/src/ros.rs
alistair23: It's the count of context switchs. I have updated this

capsules/src/ros.rs
alistair23: I suspect so. This is something that @phil-levis wanted

alistair23: Just to clarify. The idea is that count can be used for userspace to check synchronisation with the kernel data. User space doesn't use it because we can check the high/low of the timer instead, but I think the idea is that it could be useful in the future.

doc/reference/trd-shared-allow-syscalls.md
alistair23: Fixed!

capsules/src/lib.rs
```diff
@@ -63,6 +63,7 @@ pub mod proximity;
 pub mod rf233;
 pub mod rf233_const;
 pub mod rng;
+pub mod ros;
```
ppannuto, line 66:
```suggestion
pub mod read_only_syscall;
```

I think the same idea holds for the filename on spelling this out

doc/syscalls/10001_ros.md
```diff
@@ -0,0 +1,88 @@
...
+
+  do {
+    // Set the high bytes the value in memory
+    time_ticks_high = ptr[3];
+    // Read the low bytes
+    time_ticks_low = ptr[2];
+    // If the high bytes don't match what is still in memory re-try
+    // the load
+  } while (time_ticks_high != ptr[3]);
+```
```
ppannuto, line 49:
```suggestion
```c
  volatile uint32_t* ptr; // Points to ROS memory region
  uint32_t time_ticks_high, time_ticks_low;
  uint32_t switch_count = 0;

  do {
    // Start our attempt at an atomic read
    switch_count = ptr[0];
    // Set the high bytes the value in memory
    time_ticks_high = ptr[3];
    // Read the low bytes
    time_ticks_low = ptr[2];
    // Check if there was a context switch while reading this value out, if so, retry
  } while (switch_count != ptr[0]);
```
```

I think this is a more accurate code snippet to read according to the mechanisms outlined here?

**Review by ppannuto — COMMENTED** [review]
Pretty much looks good to me; one consistency note on the actual filename and I think the code example had fallen out of date, which I hopefully fixed correctly.

**Review by hudson-ayers — APPROVED** [review]

**Review by hudson-ayers — DISMISSED** [review]

**Review by ppannuto — APPROVED** [review]

File: capsules/src/driver.rs [hunk 30]
```diff
@@ -19,6 +19,7 @@ pub enum NUM {
 
     // Kernel
     Ipc                   = 0x10000,
+    ReadOnlySysCalls      = 0x10001,
```
bradjc, line 22: Does this still fall in the "kernel" category?

bradjc (changes requested):
I think this is basically good to go.

I have several comments on naming. Since this is in a kernel trait I think it is worth the tedium to make the names consistent. Also, using read only "Syscalls" is legacy at this point, and I think should be changed before merging.

Blocking on the implementation of `count` in the capsule. Right now it provides the total number of context switches ever across all processes, which seems bad. Maybe that is what we want and is OK, but I want to check, and we should make it clear that is what "count" means in documentation.




It does, however,

capsules/src/read_only_syscall.rs
```diff
@@ -0,0 +1,142 @@
...
+use kernel::ErrorCode;
+
+/// Syscall driver number.
+pub const DRIVER_NUM: usize = 0x10001;
+const VERSION: u32 = 1;
+
+pub struct ReadOnlySyscallsDriver<'a, T: Time> {
+    timer: &'a T,
+
+    count: Cell<u32>,
```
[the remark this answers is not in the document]
alistair23: Good catch! I originally thought the shared count would be fine, but it does give away information about other apps. I have fixed this

alistair23:
Fixed the side channel leak from the shared count. Also addressed all of @bradjc 's other comments, including renaming from syscalls to state

lschuermann (changes requested):
This looks pretty good and eliminates most of my concerns. Not having it implemented in the kernel but using a simple capsule and standard `SyscallDriver` is nice, and downstream boards can still add entirely different functions before the context switch if they so desire.

doc/syscalls/10001_ros.md
```diff
@@ -0,0 +1,96 @@
+---
```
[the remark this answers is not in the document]
lschuermann: FWIW the collision was probably because the assigned driver number for the low level debug capsule wasn't listed in `driver::NUM`, #2827 fixes this.

ppannuto (dismissed):

doc/syscalls/README.md
```diff
@@ -57,6 +57,7 @@ stabilized or not (a "✓" indicates stability) in the Tock 2.0 release.
 |2.0| Driver Number | Driver           | Description                                |
 |---|---------------|------------------|--------------------------------------------|
 |   | 0x10000       | IPC              | Inter-process communication                |
+|   | 0x10001       | ROS              | Read Only System, access system information|
```
bradjc, line 60: Need to update this.

doc/syscalls/00009_ros.md
```diff
@@ -0,0 +1,96 @@
+---
+driver number: 0x10001
```
bradjc, line 2: Update

bradjc (commented):
Just need to update the driver num in a couple places.

Also: shouldn't some board actually use this?

alistair23:
Fixed the last few driver numbers.

No, the agreement for having it merged was that it would be disabled by default

bradjc (approved):

hudson-ayers (approved):

ppannuto (approved):

lschuermann (approved):

phil-levis (approved):

PR #2382, opened 2021-01-26
tsl12561: Update to use Tock 2.0 Driver

ppannuto:
### Pull Request Overview

This pull request updates the tsl12561 to Tock 2.0 interface.

### Testing Strategy

Compiling -- I'm not actually sure what hardware platform has this sensor (if any remain).

### TODO or Help Wanted

I'm late to the 2.0 conversion party, but I don't think I messed anything up.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/tsl2561.rs
+++ b/capsules/src/tsl2561.rs
@@ -14,10 +14,10 @@
 //! > using an empirical formula to approximate the human eye response.
 
 use core::cell::Cell;
-use kernel::common::cells::{OptionalCell, TakeCell};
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):
looks right to me!

the original PR that added this capsule says it was "for the BLEES board"...and I do not know what that is

brghena:
nRF51822 board from lab11: https://github.com/lab11/blees So that's not going to be a valid platform to test on...

Issue #1769, opened 2020-04-17
Remove guarantees on memory size and stack pointer value at app startup (Tock 2.0)


jrvanwhy:
The Tock kernel currently implements the following:

1. Apps receive at least [`min_ram_size`](https://github.com/tock/tock/blob/master/doc/TockBinaryFormat.md#1-main) at startup.
1. Apps receive at least 3 KiB of RAM at startup.
1. The stack pointer is set 3 KiB after the start of the app's memory region at startup.

We should allow apps to use less than 3 KiB of RAM. However, doing so would break apps that rely on the above behaviors, and therefore needs to wait until Tock 2.0.

jrvanwhy:
I propose we replace the replace the existing behavior (at outlined in the issue behavior) with the following behavior:

1. Apps receive the minimum RAM they can receive that is greater than or equal to their `min_ram_size` value (subject to MPU constraints).
1. Apps are started with the highest valid stack pointer value (i.e. just before the initial app break, subject to alignment constraints).

Keeping the RAM allocation size close to `min_ram_size` bytes will help diagnose apps that give a too-small value for `min_ram_size`. Setting the stack pointer to the top of the app's RAM region gives room for interrupts to push frames onto the stack (I know this is required for ARM, I don't know about RISC-V). We would expect the userspace runtimes to perform their own (better) stack setup.

ppannuto:
I might suggest that apps start with an undefined stack pointer.

I don't think either the Rust or C runtime currently have any expectations or requirements on the initial stack pointer, I will speak authoritatively that [the C runtime starts with assembly that sets `sp` before doing anything else](https://github.com/tock/libtock-c/blob/dc87cc2d76cd7e528cc85d52d2dec990995ae4b1/libtock/crt0.c#L58).

alistair23:
The Rust assembly is basically the same as the C assembly (at least for RV32) and I'm pretty sure we also have no expectations on the stack pointer being set up.

bradjc:
> I might suggest that apps start with an undefined stack pointer.
> 
> I don't think either the Rust or C runtime currently have any expectations or requirements on the initial stack pointer, I will speak authoritatively that [the C runtime starts with assembly that sets `sp` before doing anything else](https://github.com/tock/libtock-c/blob/dc87cc2d76cd7e528cc85d52d2dec990995ae4b1/libtock/crt0.c#L58).
> 
[...]

PR #2383, opened 2021-01-27
kernel: Remove setting/tracking the stack pointer in process.rs

bradjc:
### Pull Request Overview

Background: Because cortex-m architectures use the stack to pass service call state between privileged and user modes, and tock was originally written for the cortex-m4, the core kernel has historically tracked the stack pointer of each process so the kernel knows where the svc state is and can use it during context switches. When the UserlandKernelBoundary trait was added, it was not obvious that this is the only reason the kernel tracks the stack pointer, and it was not factored out.

With more experience with RISC-V syscalls, and more thinking about what the interface should be between userspace apps and the kernel, it's more clear now that the core kernel should not have any expectations about a process's stack pointer. Process-accessible memory should be entirely managed by that process, and process.rs has no need to track something like the stack pointer.

Additionally, since the kernel only knows what the process's stack pointer is from the process telling it, we need to do a better job of treating the stack pointer as untrusted state. That is, we should be verifying that it is in fact within a process's accessible RAM region before dereferencing the stack pointer.

#### This PR

This refactors process.rs to remove keeping track of the stack pointer in process.rs (with one exception). To do this, the UKB interface changes to require each function to take in the memory bounds of the process. If the UKB implementation wants/needs to use the stack it can keep track of the stack pointer in the syscall stored_state for the process. The cortex-m implementation now checks that the stack pointer is valid before using it, as well.

This also copies from #2370 to add a function to the UKB trait so that each architecture can inform the kernel of the minimum amount of RAM space the process needs to context switch to it the first time. However, unlike #2370, this PR is still 1.x compatible.

The exception is that for debugging the kernel will still keep track of the lowest it sees the stack pointer if the UKB implementation shares the new stack pointer. Again, this is only for debugging when `panic!()` runs.

Lastly, this PR refactors the restart() function a bit, making it closer to process.new(). That means we save less state when the process starts, and re-compute more when the process finishes.

This helps with #1769.

### Testing Strategy

This pull request was tested by running various apps on hail and nano33ble. I also tried restarting apps.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/kernel/src/syscall.rs
+++ b/kernel/src/syscall.rs
@@ -63,6 +63,13 @@ pub enum ContextSwitchReason {
 /// This trait must be implemented by the architecture of the chip Tock is
 /// running on. It allows the kernel to manage switching to and from processes
 /// in an architecture-agnostic manner.
+///
+/// Since exactly how callbacks and return values are passed between kernelspace
+/// and userspace is implementation specific, and may use process memory to
+/// store state when switching, functions in this trait are passed the bounds of
+/// process-accessible memory so that the implementation can verify it is
+/// reading and writing memory that the process has valid access to. These
+/// bounds are passed through `memory_start` and `app_brk` pointers.
 pub trait UserspaceKernelBoundary {
     /// Some architecture-specific struct containing per-process state that must
     /// be kept while the process is not running. For example, for keeping CPU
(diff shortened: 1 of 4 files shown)


kernel/src/syscall.rs
```diff
@@ -73,36 +80,63 @@ pub trait UserspaceKernelBoundary {
...
+    /// switch is possible.
+    ///
+    /// Some architectures may not need any allocated memory, and this should
+    /// return 0. In general, implementations should try to pre-allocate the
+    /// minimal amount of process-accessible memory (i.e. return as close to 0
+    /// as possible) to provide the most flexibility to the process. However,
+    /// the return value will be nonzero for architectures where values are
+    /// passed in memory between kernelspace and userspace during syscalls or a
+    /// stack needs to be setup.
+    fn initial_process_app_brk_size(&self) -> usize;
```
jrvanwhy, line 99: I suspect there should be an alignment-related statement here. Something like "The memory will be aligned to at least the smallest alignment supported by the MPU."

This might be a bit pedantic, because I don't expect there to be an architecture where MPU regions are not at least word-aligned.

hudson-ayers (commented):
I think this looks good, with the exception of the UB change I suggested.

I spent awhile trying to reason about if the stack pointer debug logic is changed now, and couldn't convince myself one way or the other, but I don't think its that important if the stack pointer debug info doesn't help right after a process starts anyway.

bradjc:
> I spent awhile trying to reason about if the stack pointer debug logic is changed now, and couldn't convince myself one way or the other, but I don't think its that important if the stack pointer debug info doesn't help right after a process starts anyway.

You are correct that it has changed, but I agree it's not significant, and also it is _probably_ better after this change. Since process.rs does not assign the initial stack pointer, it is unable to store any debug information right after the process is created w.r.t. the stack pointer. Before it did set the initial stack pointer and therefore could save it. So, between when a process is created and before the process calls memop to tell the kernel where it placed its stack, we have less debug info than we did before. However, since we don't want the kernel to manage the process's stack at all, this is probably a good thing.

kernel/src/syscall.rs
bradjc: If you are thinking about alignment for the MPU, process.rs is not responsible for aligning memory for the MPU. process.rs doesn't know what the MPU alignment requirements are just like it doesn't know the required min app_brk size, since both are platform-specific.

Yes, the platform will want the memory region of at least this sized aligned for the MPU, but it's up to the platform's MPU implementation to enforce that.

That being said, process.rs does decide where inside of the entire memory allocation for a process to allocate the number of bytes reserved from this function. I think it is established that the "process accessible" region is in the lower addresses of the process's overall memory region. I'm not sure if we explicitly say that the kernel will start the process accessible region from the very beginning of the overall region, however. Although it is certainly implied in this overall interface.

jrvanwhy: > That being said, process.rs does decide where inside of the entire memory allocation for a process to allocate the number of bytes reserved from this function. I think it is established that the "process accessible" region is in the lower addresses of the process's overall memory region. I'm not sure if we explicitly say that the kernel will start the process accessible region from the very beginning of the overall region, however. Although it is certainly implied in this overall interface.

Ah, I misunderstood how that works. I suppose that means that `process.rs` is responsible for suitably aligning the region?

hudson-ayers (dismissed):

kernel/src/syscall.rs
bradjc: process.rs can start the process's accessible memory region wherever it wants inside of the region chosen by the MPU code, so yes it does need to think about alignment. Should we say that accessible memory start will be word aligned?

bradjc: Alright I added it to the `initialize_process` comment. I made a comment in process.rs that we have to respect this as well.

jrvanwhy (approved):

hudson-ayers (approved):

PR #2387, opened 2021-01-27
Transition I2CMaster system call driver to 2.0 syscall API

phil-levis:
### Pull Request Overview

This pull request adds pulls I2CMaster fully into the Tock 2.0 syscall API, also removing its dependence on an I2C by making it a `dyn I2CMaster` reference (in line with more modern capsule designs).

### Testing Strategy

This pull request was tested by adding support in libtock-c, writing a test application, and testing it along with a slave in the Master/Slave test app.

### TODO or Help Wanted

Nothing.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/earlgrey-nexysvideo/src/main.rs
+++ b/boards/earlgrey-nexysvideo/src/main.rs
@@ -75,7 +75,7 @@ struct EarlGreyNexysVideo {
         'static,
         capsules::virtual_uart::UartDevice<'static>,
     >,
-    i2c_master: &'static capsules::i2c_master::I2CMasterDriver<lowrisc::i2c::I2c<'static>>,
+    i2c_master: &'static capsules::i2c_master::I2CMasterDriver<'static>,
     nonvolatile_storage: &'static capsules::nonvolatile_storage_driver::NonvolatileStorage<'static>,
 }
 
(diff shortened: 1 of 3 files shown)


hudson-ayers:
Is there something about using generics that is incompatible with 2.0? I'm kinda bummed by this change, it seems to just add overhead where none is needed. The driver is never going to reference more than one implementation of the underlying driver at runtime. I realize that generics aren't great for code readability, but this just adds unneeded v-table lookups on every access to the underlying driver.

phil-levis:
I think it's entirely a readability question. I changed it because I thought this was the approach we were moving towards adopting, except perhaps for modules that have lots and lots of calls (e.g., Alarm). I'm happy to follow consensus on this, whatever it is. Perhaps let's bring this up in the call tomorrow?

hudson-ayers:
The size-diff action did not run on this PR because it only runs on master (my bad), but this PR increases code size on Imix by 704 bytes.

phil-levis:
That is a lot (imagine x10 capsules). I will change it back. We should look into quantifying this for all capsules and transition them if the savings are significant.

hudson-ayers (approved):

hudson-ayers:
This really doesn't seem to have anything 2.0 specific about it, should it just be merged into master?

phil-levis:
Ah, it looks like Amit's and my PRs overlapped. I picked this up trying to take some load off Amit's set of capsules. His PR to update this capsule to the 2.0 API was merged. So the change-to-2.0 parts are not diffs to the tock-2.0-dev branch.

PR #2388, opened 2021-01-28
capsules: mlx90614: Update to the Tock 2.0 Driver

alistair23:
### Testing Strategy

Run on Apollo3

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/mlx90614.rs
+++ b/capsules/src/mlx90614.rs
@@ -22,7 +22,7 @@ use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::common::registers::register_bitfields;
 use kernel::hil::i2c::{self, Error};
 use kernel::hil::sensors;
-use kernel::{AppId, Callback, LegacyDriver, ReturnCode};
+use kernel::{AppId, Callback, CommandResult, Driver, ErrorCode, ReturnCode};
 
 /// Syscall driver number.
 pub const DRIVER_NUM: usize = driver::NUM::Mlx90614 as usize;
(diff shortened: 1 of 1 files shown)


capsules/src/mlx90614.rs
```diff
@@ -155,55 +151,60 @@ impl<'a> i2c::I2CClient for Mlx90614SMBus<'a> {
...
+        callback: Callback,
         _app_id: AppId,
-    ) -> ReturnCode {
-        match subscribe_num {
+    ) -> Result<Callback, (Callback, ErrorCode)> {
+        let res = match subscribe_num {
             0 /* set the one shot callback */ => {
-				self.callback.insert(callback);
-				ReturnCode::SUCCESS
+                self.callback.replace(callback);
```
phil-levis, line 198: I think you want a `mem::swap` here? Otherwise you return the passed callback, rather than the replaced one? You could assign to `callback` and that would work.

phil-levis (changes requested):

capsules/src/mlx90614.rs
ppannuto: I think this is right as-is, [replace returns the old contents of the Cell](https://doc.rust-lang.org/std/cell/struct.Cell.html#method.replace). If this is wrong, I did the same thing in #2382

hudson-ayers (approved):

capsules/src/mlx90614.rs
phil-levis: Sure, replace does that, but the return value is discarded (there's a semicolon there, and the next line returns `Ok(())`).

phil-levis: Yours is correct -- you return the result if `replace` in an `Ok()`. Using `replace` is fine in general; the challenge here is that callback is inserted in the bottom `match`.

phil-levis (changes requested):

capsules/src/mlx90614.rs
ppannuto: Oh, yeah, I see that now. I think the whole bottom match and intermediate `res` variable can/should go away here, very much like #2382 probably.

alistair23:
Fixed

phil-levis (approved):

PR #2390, opened 2021-01-28
MSP432 I2C driver

lebakassemmerl:
### I2C driver support for the MSP432

This PR adds support for I2C on the MSP432. The implementation is done without the usage of DMA since the MSP432 has only 8 DMA channel and 1 I2C bus would require 2 channels (TX and RX). Due to the fact that in most of the cases only a few bytes are transmitted over I2C I thought it would be better to handle the complete implementation via interrupts and save the DMA channels for other peripherals.

### Testing Strategy

Tested on the evaluation board using a [DS1621 temperature sensor](https://pdfserv.maximintegrated.com/en/ds/DS1621.pdf) as slave and the sensor example as application. 

### TODO or Help Wanted

* Is the no-DMA approach ok/fine? Or should I rewrite the driver to use DMA channels?
* For testing I wrote a capsule driver for the DS1621 chip. It's a very simple temperature sensor which works with I2C. Also not all functionalities are implemented but it works well as a standard temperature sensor. Is there an interest for this driver-support? If yes, I could create another PR for it.

### Documentation Updated

### Formatting

- [x] Ran `make ci-nosetup`.

### DIFF
--- a/chips/msp432/src/i2c.rs
+++ b/chips/msp432/src/i2c.rs
@@ -17,6 +17,8 @@ pub enum Speed {
 
 #[derive(Copy, Clone, PartialEq)]
 enum OperatingMode {
+    Unconfigured,
+    Disabled,
     Idle,
     Write,
     Read,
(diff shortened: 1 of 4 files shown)


chips/msp432/src/i2c.rs
```diff
@@ -0,0 +1,374 @@
...
+            + usci::UCBxCTLW1::UCSWACK::HardwareTriggered
+            // Set glitch filtering to 50ns (according to I2C standard)
+            + usci::UCBxCTLW1::UCGLIT::_50ns,
+        );
+
+        // Don't clear the module reset here since we set the state to Disabled
+        self.mode.set(OperatingMode::Disabled);
+    }
+
+    pub fn set_speed(&self, speed: Speed) {
```
bradjc, line 170: Does this need to be `pub`?

bradjc (approved):

chips/msp432/src/i2c.rs
lebakassemmerl: Since the HIL traits don't offer any possibility to set the speed it's done via a seperate function. At least at the STM32F4 discovery board (which is where I got the idea from) it's done the same way.

I could also set it to a default value if this would be a better solution.

PR #2391, opened 2021-01-28
Add platform_helper! macro

lschuermann:
### Pull Request Overview

This pull request adds a `platform_helper!` macro, which generates a struct of all drivers of a board and automatically implements the `kernel::Platform` trait on it. The existing boards are updated to use this macro.

#### Motivation

The motivation for this macro originated from an implementation to detect and prevent swapping of AppSlices and Callbacks in the Tock 2.0 driver interface. This implementation requires iterating over all installed drivers of a particular platform, which would require adding a new method to the platform trait. Given that the Platform structs and this trait implementation are rather mechanical, it seemed to make sense to create a macro for that.

#### Advantages

The primary advantage of this PR is the decoupling of the semantic information of a Platform struct from the trait implementation itself. While the `kernel::Platform` implementation determines how the information of the installed capsules is conveyed to the kernel, the `platform_helper!` macro determines a syntax which encodes this information in a well-defined way. This enables changing or extending the `kernel::Platform` trait (for instance, adding a method to iterate over the capsules) with minimal impact, by only updating the macro.

It could also be useful for implementing vDSO style system calls (#2366) with capsules, for instance by adding a `ProactiveDriver: Driver` trait and adding a method to invoke every registered `ProactiveDriver` prior to switching to an app.

### Testing Strategy

This pull request was not tested.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


bradjc:
Just for my own understanding, can you briefly explain or send a pointer to what swapping an AppSlice is and why that is bad?

bradjc:
I can say upfront that I am macro adverse. Macros can really hide what code actually exists, and can make it very difficult to debug or understand what code is actually being compiled. They don't always play nice with IDE-esque tools as well.

Having main.rs be largely macro-free is in my opinion a big win up to this point. We do have `static_init`, but that is at least fairly contained and doesn't have any control-flow.

I'm curious what others think, but I kind of view this change as a "last resort". We do want to ensure safety and reliability in Tock, but we also have to consider usability and maintainability. It would also help to see what the iterate-all-drivers code looks like.

alistair23:
I agree with Brad. I think Rust macros are very confusing and being in `main.rs` means that lots of developers are going to have to use them.

In saying that it seems pretty straight forward to just copy and paste an existing macro and edit that, so it shouldn't be too hard to use.

The advantage of not having to define each capsule twice (`struct` and `kernel::Platform` implemtantion) seems really nice though.

I would say I'm more of a meh about this PR then Brad's last resort. There is good and bad in using macros here.

lschuermann:
> I can say upfront that I am macro adverse. Macros can really hide what code actually exists, and can make it very difficult to debug or understand what code is actually being compiled. They don't always play nice with IDE-esque tools as well.

I think that's a good position to start from. I'm also very hesitant about using macros. The hygiene guarantees Rust provides are a great advantage over macros in other languages, but the issues you mentioned will remain.

The primary argument for using macros here is that in the case of a Platform trait, for each driver, the `(field_name, driver_num, type)` 3-tuple is tightly coupled for each entry, but when written explicitly split out over the struct definition and the `Platform` trait implementation. It's easy to make a mistake here and let a `driver_num` point to the wrong field. I imagined these interdependencies getting harder to manage in case the Platform trait would be extended by a method to iterate over all drivers, a possible implementation of which could look like:

```rust
fn iter_drivers<F>(&self, f: F)
where
    F: Fn(usize, &dyn kernel::Driver),
{
    f(
        capsules::console::DRIVER_NUM,
        self.console as &dyn kernel::Driver,
    );
    f(
        capsules::alarm::DRIVER_NUM,
        self.alarm as &dyn kernel::Driver,
    );
}
```

I would've hoped that this macro might manage to represent the 3-tuple of `(field_name, driver_num, type)` in an abstract way, mostly independent of it's precise implementation and the interactions as part of the `Platform` trait. On the other hand I see the drawbacks this introduces (for one, rustfmt doesn't like the syntax too much).

hudson-ayers:
I am torn on this. I agree with Brad that macros hide code and make tracing stuff harder. I have definitely noticed students that are new to Rust/Tock have the most trouble understanding the component macros when they try to understand board bring up. The macro introduced in this PR is especially hard to follow.

On the other hand, this is pretty cool. Right now the whole Platform trait implementation has a ton of unnecessary noise. All a "board maintainer" cares about is: which userspace drivers do I want to include. To get there, they have to list the specific types of all these drivers, specify a bunch of lifetimes, specify a bunch of generic types (which change if you add virtualization!) specify the corresponding driver numbers, etc. There ends up being a lot of code duplication as a result. This cuts out some of that, which I like. But ultimately probably not enough to be worthwhile.

I suspect this could be taken further if we made DRIVER_NUM an associated constant on the `Driver` trait. Then there would be no need to list the driver_num for each capsule at all - just the type. We could maybe even derive the field names from the capsule names! I could imagine defining the Platform struct and trait looking like this:

```rust
create_userspace_drivers!(
        capsules::console::Console<'static>,
        capsules::gpio::GPIO<'static, arty_e21_chip::gpio::GpioPin<'static>>,
        capsules::alarm::AlarmDriver<
            'static,
            VirtualMuxAlarm<'static, rv32i::machine_timer::MachineTimer<'static>>,
        >,
        capsules::led::LedDriver<
            'static,
            hil::led::LedHigh<'static, arty_e21_chip::gpio::GpioPin<'static>>,
        >,
        capsules::button::Button<'static, arty_e21_chip::gpio::GpioPin<'static>>,
);
```
(legacy_driver omitted bc it will be gone post-2.0. board name ommitted because calling it Platform everywhere is fine imo)

Then the macro would actually represent what users are doing: they are listing the types of userspace drivers they want.

But I don't want to ask Leon to do more work on this if we are unlikely to include it anyway.

lschuermann:
> Just for my own understanding, can you briefly explain or send a pointer to what swapping an AppSlice is and why that is bad?

In Tock 2.0, a `subscribe` or `allow` operations takes an AppSlice/Callback and returns another, through Rust move semantics. The information contained in these structs are returned to the process issuing this system call. A userspace process expects that, for a given subdriver number and a given driver, an allow or subscribe operation will always return the previous values of this (driver, subdriver) combination. In case of a buffer, a Rust userspace might convert it back to a slice and pass it back to the caller. In case of a subscribe operation, the appdata might refer to a location in memory that holds state information which might want to be `free()`d. Currently, this is by convention, but not enforced in the kernel. I'm working towards solutions that will be able to catch misbehaving capsules at runtime. Hope that gives an idea. This PR is largely independent of that and should really have been an RFC.

lschuermann:
> I suspect this could be taken further if we made DRIVER_NUM an associated constant on the `Driver` trait.

That's a good idea and what I was originally aiming for. However, having associated constants on `Driver`s will make them not be object safe anymore, which we need for the rest of the kernel infrastructure to work, so that unfortunately is not an option. I'd be happy to explore more in this direction though and willing to invest the time, if we find ways to make that (or an approach with similar semantics) work.

> Then the macro would actually represent what users are doing: they are listing the types of userspace drivers they want.

That would be a nice goal to work towards. Even if this PR in it's current state isn't what we're aiming for, I suppose it's still a good survey that, in a more polished variant, it might be something we want to have?

hudson-ayers:
> That's a good idea and what I was originally aiming for. However, having associated constants on `Driver`s will make them not be object safe anymore, which we need for the rest of the kernel infrastructure to work, so that unfortunately is not an option. I'd be happy to explore more in this direction though and willing to invest the time, if we find ways to make that (or an approach with similar semantics) work. I could imagine something being possible along the lines of const trait fns maybe?

Ah, didn't think about object safety. const trait fns probably wouldn't work either for the same reason, and will be unstable for awhile. There are hacky ways to pattern match on non-constants that could work, but we would lose the compile time checking that no driver number is used twice, which is a bit of a shame. It would probably be possible to move that check to runtime.

> That would be a nice goal to work towards. Even if this PR in it's current state isn't what we're aiming for, I suppose it's still a good survey that, in a more polished variant, it might be something we want to have?

Yeah my position is I would want this if it could go that far but in its current state I don't know that it's worth the indirection.

bradjc:
> I suspect this could be taken further if we made DRIVER_NUM an associated constant on the `Driver` trait.

We have intentionally made it so that main.rs has to define the driver num mapping for the syscalls it exports. We've always viewed that as a kernel configuration operation that boards have to be explicit about. I'm not sure if making it an associated constant would make the driver number fixed or not, but wanted to point this out.

hudson-ayers:
> We have intentionally made it so that main.rs has to define the driver num mapping for the syscalls it exports. We've always viewed that as a kernel configuration operation that boards have to be explicit about. I'm not sure if making it an associated constant would make the driver number fixed or not, but wanted to point this out.

I was imagining it as making a driver number fixed, so that is a good point.

bradjc:
> > Just for my own understanding, can you briefly explain or send a pointer to what swapping an AppSlice is and why that is bad?
> 
> In Tock 2.0, a `subscribe` or `allow` operations takes an AppSlice/Callback and returns another, through Rust move semantics. The information contained in these structs are returned to the process issuing this system call. A userspace process expects that, for a given subdriver number and a given driver, an allow or subscribe operation will always return the previous values of this (driver, subdriver) combination. In case of a buffer, a Rust userspace might convert it back to a slice and pass it back to the caller. In case of a subscribe operation, the appdata might refer to a location in memory that holds state information which might want to be `free()`d. Currently, this is by convention, but not enforced in the kernel. I'm working towards solutions that will be able to catch misbehaving capsules at runtime. Hope that gives an idea. This PR is largely independent of that and should really have been an RFC.

Hudson make this nice figure:

image (1)

If I'm understanding everything, the issue boils down to calculating S_n so we know how much space to allocate?

What about storing whatever book keeping is needed in the grant region? Since we have to create grants anyway, when a grant is created we could say how much space we need to enforce allows and subscribes. If there is not enough space, we return an error. Also, the space is then only allocated if the driver is actually used by a process.

lschuermann:
Hm, I suppose this doesn't necessarily need to be open anymore. It's an interesting approach (also in light of #2462 requiring driver nums on multiple occasions, which this could work around), but it also has a lot of issues. We can always reopen this if we need it later on.

PR #2393, opened 2021-01-29
Port NRF Serialization to 2.0

ppannuto:
### Pull Request Overview

This updates the nrf serialization to the 2.0 interface.

### Testing Strategy

Tested on imix with the `ble-uart` test application and the nRF BLE sniffer app for an iPhone. The device advertises with an interval of 505ms. 

### TODO or Help Wanted

Not required, but figuring out why the `ble-uart` application throws an error, code 16835, would be good. The code in master throws this as well, so it's not blocking on this PR.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


capsules/src/nrf51822_serialization.rs
```diff
@@ -181,33 +217,33 @@ impl LegacyDriver for Nrf51822Serialization<'_> {
...
+    fn command(&self, command_type: usize, _: usize, _: usize, appid: AppId) -> CommandResult {
         match command_type {
-            0 /* check if present */ => ReturnCode::SUCCESS,
+            0 /* check if present */ => CommandResult::success(),
 
             // Send a buffer to the nRF51822 over UART.
             1 => {
                 self.apps.enter(appid, |app, _| {
-                    app.tx_buffer.take().map_or(ReturnCode::FAIL, |slice| {
+                    app.tx_buffer.map_or(CommandResult::failure(ErrorCode::FAIL), |slice| {
```
phil-levis, line 227: Should this be a NOMEM or RESERVE?

capsules/src/nrf51822_serialization.rs
```diff
@@ -244,7 +278,7 @@ impl uart::ReceiveClient for Nrf51822Serialization<'_> {
 
         self.active_app.map(|appid| {
             let _ = self.apps.enter(*appid, |app, _| {
-                app.rx_buffer = app.rx_buffer.take().map(|mut rb| {
+                app.rx_buffer.mut_map(|rb| {
```
phil-levis, line 281: I think you do want a `mut_map_or` here, because you want to note the case of when there wasn't a receive buffer, and when there is, how much was copied. In particular, you should be signaling `max_len`, not `rx_len` in the callback. So you need to pass this out of the `mut_map_or`.

phil-levis (changes requested):
I can imagine that we might want a `map` and `mut_map`, but I think in this case you want to use `mut_map_or` in order to pass a useful value to userspace. With the current implementation, it's possible userspace passes a smaller buffer than what's received and you'll pass the complete receive length rather than what's copied. This will likely lead userspace to run off the end of the buffer.

phil-levis:
Working on this now.

phil-levis:
I've updated this to use `mut_map_or` and pass lengths properly back to userspace. 

I've also updated the libtock-c support to use the new APIS: https://github.com/tock/libtock-c/tree/nrf_serialization_2.0

However, it seems to fail immediately (`simple_ble_init` returns an error), and this occurs when registering a handler through `SOFTDEVICE_HANDLER_INIT`. I'd appreciate it if someone who is at last familiar with the NRF softdevices takes a look.

phil-levis (dismissed):
This still needs to be tested -- can't get the userspace library working.

phil-levis (commented):
Needs testing.

bradjc:
I had to make one change, removing the buffer length from the rx syscall. Otherwise I was able to make this interface work with libtock-c.

hudson-ayers (approved):
LGTM, I don't think Tock 2.0 requires that apps be able to specify a receive length

capsules/src/nrf51822_serialization.rs
```diff
@@ -104,38 +107,67 @@ impl LegacyDriver for Nrf51822Serialization<'_> {
...
+    }
+
+    /// Pass application space memory to this driver.
+    ///
+    /// This also sets which app is currently using this driver. Only one app
+    /// can control the nRF51 serialization driver.
+    ///
+    /// ### `allow_num`
+    ///
+    /// - `1`: Provide a TX buffer.
```
ppannuto, line 146:
```suggestion
    /// - `0`: Provide a TX buffer.
```

capsules/src/nrf51822_serialization.rs
```diff
@@ -244,24 +281,26 @@ impl uart::ReceiveClient for Nrf51822Serialization<'_> {
 
         self.active_app.map(|appid| {
             let _ = self.apps.enter(*appid, |app, _| {
-                app.rx_buffer = app.rx_buffer.take().map(|mut rb| {
+                let len = app.rx_buffer.mut_map_or(0, |rb| {
```
ppannuto, line 284: I believe the `len` variable and the `mut_map_or` and later `map_or` are wrong here.

The comment below says `Note: This indicates how many bytes were received by hardware, regardless of how much space (if any) was available in the buffer provided by the app.`

But with the current code, if the `rx_buffer` is missing it will return `0` instead, and if it's too short, the later `max_len` will return the app's `rx_buffer` length, not the returned hardware length.

I think the fix is simply to remove the `len` variable, remove all of the `_or`'s here, and pass `rx_len` in the callback below. I'm not going to push that fix without consensus though, as I think we'd be going back and forth if I did that..

ppannuto (commented):
image

_Okay github._

capsules/src/nrf51822_serialization.rs
phil-levis: Good point -- then we should pass both. I assume these semantics are there for the peculiarities of the nRF serialization library? I'd therefore recommend this become

```rust
app.callback.schedule(4, rx_len, len);
```

Having implemented a bunch of these capsules, I've come to think that a simple `map` is dangerous; the situations in which you want to plug away when there's nothing there is pretty narrow. The API you describe, in which the byte count is the number received by the hardware, rather than what's in the buffer, is fine as long as there is always a buffer (of known length) there. But this `map` indicates that there might not be; calling `map` prevents you from being able to detect an error case. It might be that we do not want to dig into the nRF code to handle this case right now, but I think it doesn't speak well for the kernel if we make it impossible to handle. `map_or` has the advantage that it forces you to think about this -- what if the buffer isn't there?

phil-levis:
OK, this is now working.

kernel/src/mem.rs
```diff
@@ -50,6 +50,19 @@ pub trait Read {
...
+    /// If the process is no longer alive and the memory has been
+    /// reclaimed, this method must return the default value.
+    ///
+    /// # Default AppSlice
+    ///
+    /// A default instance of an AppSlice must not execute the closure.
+    fn map<F>(&self, fun: F)
+    where
+        F: FnOnce(&[u8]);
+
```
hudson-ayers, line 65: All the added functions in this file are no longer used, but still present here

alevy:
This is great! (Modulo @hudson-ayers's comments)

Can we please squash these nine commits and rewrite a single descriptive commit message? They are kind of all over the place and ultimately the changes in the final diff are fairly self contained.

Otherwise, I think, good to merge

kernel/src/mem.rs
phil-levis: Done.

phil-levis:
Ugh. Because I pulled tock-2.0-dev into it back on the 9th, squashing is beyond by git abilities.

PR #2394, opened 2021-01-29
trd: syscalls, 64 bit types in AAPCS friendly way

ppannuto:
### Pull Request Overview

Given the goal of minimizing register moving and churn, change the
syscall return for 64-bit values to something that can be expressed
in ARM. From the AAPCS:

> A double-word sized type is passed in two consecutive registers
> (e.g., r0 and > r1, or r2 and r3). The content of the registers
> is as if the value had been loaded from memory representation with
> a single LDM instruction.
>
> If the argument requires double-word alignment (8-byte), the NCRN is rounded
> up to the next even register number.

The point: A function signature of the form `f(int x, long long y);` will map
`x -> r0` and `y -> r2,r3` on ARM.

This change will avoid the need to move 64-bit values.

### Testing Strategy

N/A

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/doc/reference/trd-syscalls.md
+++ b/doc/reference/trd-syscalls.md
@@ -149,7 +149,7 @@ for CortexM they are `r0`-`r3` and for RISC-V they are `a0`-`a3`.
 | Success                    | 128  |                    |                    |                    |
 | Success with u32           | 129  | Return Value 0     |                    |                    |
 | Success with 2 u32         | 130  | Return Value 0     | Return Value 1     |                    |
-| Success with u64           | 131  | Return Value 0 LSB | Return Value 0 MSB |                    |
+| Success with u64           | 131  |                    | Return Value 0 LSB | Return Value 0 MSB |
 | Success with 3 u32         | 132  | Return Value 0     | Return Value 1     | Return Value 2     |
 | Success with u32 and u64   | 133  | Return Value 0     | Return Value 1 LSB | Return Value 1 MSB |
 


hudson-ayers (approved):
This also seems to improve performance for risc-v, from https://riscv.org/wp-content/uploads/2015/01/riscv-calling.pdf :

>When primitive arguments twice the size of a pointer-word are passed on the stack, they are
naturally aligned. When they are passed in the integer registers, they reside in an aligned even-odd register pair, with the even register holding the least-significant bits. In RV32, for example, the function void foo(int, long long) is passed its first argument in a0 and its second in a2 and a3. Nothing is passed in a1.

alistair23 (approved):

hudson-ayers:
should this PR go ahead and update the syscall encoding in `kernel/src/syscalls.rs`, as well?

ppannuto:
Those tables didn't seem to quite line up the TRD already, so I'd assumed they were possibly in a holding pattern (e.g. `syscalls.rs` currently has entry _131_ as `SuccessU32U32U32` and entry _132_ as `SuccessU64`). I can update all of the entries to match this table, but that feels like possibly a different issue.

phil-levis:
Let's wait on this a bit? I want to think about it.. Also, I am wary of specification changes without an implementation, etc.

I am trying to understand how this change would manifest in better assembly code. We aren't using the calling convention but rather just inline assembly.

phil-levis (changes requested):
I'm open to this change but would like to discuss it: I'm trying to understand whether this actually changes any assembly anywhere.

phil-levis:
> 
> 
> This also seems to improve performance for risc-v, from https://riscv.org/wp-content/uploads/2015/01/riscv-calling.pdf :
> 
> > When primitive arguments twice the size of a pointer-word are passed on the stack, they are
> > naturally aligned. When they are passed in the integer registers, they reside in an aligned even-odd register pair, with the even register holding the least-significant bits. In RV32, for example, the function void foo(int, long long) is passed its first argument in a0 and its second in a2 and a3. Nothing is passed in a1.

Where does this stack alignment play out? My understanding of the APIs is that since it's a return value (not an argument), I'm not sure how this plays out. The API

ppannuto:
I don't think it would directly affect the bottom-layer syscall, but I would imagine that the syscall implementation will map return values from registers onto variables, which may then end up in function calls. 

Moving it over a register costs nothing and may help in certain situations, so it seemed sensible.

phil-levis:
> I don't think it would directly affect the bottom-layer syscall, but I would imagine that the syscall implementation will map return values from registers onto variables, which may then end up in function calls.
> 
> Moving it over a register costs nothing and may help in certain situations, so it seemed sensible.

But this isn't how the syscalls work. I'm kind of worried about this suggested change, because it's a "seems like a good idea" that is quite divested from the actual call patterns. This is a return type, not an argument to a function. So the part of AACPS you cite isn't particularly relevant. Instead, this one is:

 >   A Composite Type larger than 4 bytes, or whose size cannot be determined statically by both caller and callee, is stored in memory at an address passed as an extra argument when the function was called (Parameter Passing, rule A.4). The memory to be used for the result may be modified at any point during the function call.

So this means it's living on the stack. So for the case you're considering to happen, some other call must have somehow mapped it into registers? The one place this can kind-of-sort-of-happen is if you try to take the structure and then cast it to having a `u64`. But this is super-dicey; you're basically trusting that the layout matches the C structure layout. You're also allocating a structure that has 8 bytes of waste, holding the `type` field and the padding. Put another way, if you are calling functions with the 64-bit value, you should no longer have it stored in the full return type structure and so register allocation on calls is not an issue.

So this has less to do with register allocation in AACPS and instead C structure layout. I think it boils down to whether we want to bind the layout of the system return type to C structure layouts. I.e., should this be able to work:

``` c
typedef struct {
  syscall_rtype_t type;
  uint32_t data[3];
} syscall_return_t;

typedef struct {
  syscall_rtype_t type;
  uint64_t val64;
} syscall_success_u84_t;

syscall_return_t rval = command2(...);
if (rval.type == TOCK_SYSCALL_SUCCESS_U64) {
    syscall_success_u84_t* success = (syscall_success_u64_t*)&rval; 
    printf("%llu\m", success->val64); // Does this get the stored value?
}
```

As opposed to
```
uint64_t val = (uint64_t) rval[0] | (uint64_t)rval[1] << 32;
```

bradjc:
Since we're only returning 64 bit values from syscalls, and the low-level syscall needs to return the error value as well, I don't think we can optimize the function return registers in this way.

Phil's point is good, but I think just adding a `packed` attribute to the `syscall_success_u64_t` struct should get the same effect? That is, if we do want to be able to directly copy registers into a C struct we still can, it just has to be a packed struct. And since we aren't on a 64 bit platform having a packed struct shouldn't matter. So, the net effect is the same, but we wouldn't have to make a change at this stage.

bradjc:
Closing this but could revisit if we have a use case where this would help.

PR #2334, opened 2021-01-08
update rust january 7 2021

hudson-ayers:
### Pull Request Overview

This pull request updates the rust nightly version to 2021-01-07. This was motivated by the stabilization of `min_const_generics` ( https://github.com/rust-lang/rust/pull/79135 ) -- by using this nightly, we will be able to use const generics in Tock without adding an unstable feature.

### Testing Strategy

compiling

### TODO or Help Wanted

N/A

### Documentation Updated

- [x] Updated the relevant files in `/docs`.

### Formatting

- [x] Ran `make prepush`.

PR #2223, opened 2020-11-25
Stack size update for STM32F4 boards

alexandruradovici:
### Pull Request Overview

This pull request fixes the stack size requirement for STM32F4 boards. Due to the new board instantiation #2188, the required stack size has to be doubled as @hudson-ayers suggested in https://github.com/tock/tock/pull/2188#discussion_r529676461.

### Testing Strategy

This pull request was tested using an STM32f412GDiscovery board.

### TODO or Help Wanted

@krady21, @cosmindanielradu19, @teonaseverin  please test using the ~STM32F3Discovery and~ Nucleo. Running the blink app is just enough.

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make prepush`.

PR #2400, opened 2021-02-02
Double stack buffer for i.MX RT-based boards, prune panic! allocations

mciantyre:
### Pull Request Overview

This pull request doubles the stack buffer for boards that use i.MX RT chips. #2311 added board-based peripheral initialization for i.MX RT boards. We use more stack space (at least ~2KiB, most memory for GPIO pins) to initialize the static peripherals, which might exceed the stack buffer once the kernel starts running. Doubling the stack size seems to suffice for running the kernel and two applications on a Teensy 4.0. This approach is similar to #2223, which noted the issue in STM32 boards.

I first bisected the issue down to a commit in #2334. After that compiler update, I couldn't run the Tock kernel on the Teensy 4.0. But it turns out that, after #2311 and before the updated compiler, a `panic!()` would overflow the stack. So this seems to be a latent issue that manifested with a compiler update. The boards' panic handlers were allocating large GPIO ports structures just to access a single LED. This PR fixes that by exposing a constructor for individual GPIO pins, and using that constructor in the panic handler.

### Testing Strategy

This pull request was tested by rebasing onto trunk, compiling the kernel for a Teensy 4.0, and using the kernel to run two `libtock-c` example apps. Then, I added a `panic!()` just before the main kernel loop, and showed that the panic handler worked as expected.

### TODO or Help Wanted

There may be other boards that allocate large objects (GPIO ports and pins) in the panic handler. See the [stm32f3discovery board](https://github.com/tock/tock/blob/86079650ae9d2e5b7287571a2219050c76aa3f1c/boards/stm32f3discovery/src/io.rs#L72) and the [stm32f412gdiscovery board](https://github.com/tock/tock/blob/86079650ae9d2e5b7287571a2219050c76aa3f1c/boards/stm32f412gdiscovery/src/io.rs#L74) for two examples. Are we aware of this behaviors in these boards, or in other boards?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/teensy40/src/io.rs
+++ b/boards/teensy40/src/io.rs
@@ -48,8 +48,8 @@ impl Write for Writer<'_> {
 #[panic_handler]
 unsafe fn panic_handler(panic_info: &core::panic::PanicInfo) -> ! {
     let ccm = crate::imxrt1060::ccm::Ccm::new();
-    let ports = imxrt10xx::gpio::Ports::new(&ccm);
-    let led = &mut led::LedHigh::new(ports.pin(gpio::PinId::B0_03));
+    let pin = crate::imxrt1060::gpio::Pin::from_pin_id(gpio::PinId::B0_03);
+    let led = &mut led::LedHigh::new(&pin);
     let mut lpuart2 = lpuart::Lpuart::new_lpuart2(&ccm);
     let mut writer = Writer::new(&mut lpuart2);
     debug::panic(
--- a/boards/imxrt1050-evkb/src/io.rs
+++ b/boards/imxrt1050-evkb/src/io.rs
@@ -63,9 +63,8 @@ impl IoWrite for Writer {
 #[panic_handler]
 pub unsafe extern "C" fn panic_fmt(info: &PanicInfo) -> ! {
     // User Led is connected to AdB0_09
-    let ccm = crate::imxrt1050::ccm::Ccm::new();
-    let ports = imxrt1050::gpio::Ports::new(&ccm);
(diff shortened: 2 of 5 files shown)


boards/teensy40/src/io.rs
```diff
@@ -48,8 +48,8 @@ impl Write for Writer<'_> {
 #[panic_handler]
 unsafe fn panic_handler(panic_info: &core::panic::PanicInfo) -> ! {
     let ccm = crate::imxrt1060::ccm::Ccm::new();
-    let ports = imxrt10xx::gpio::Ports::new(&ccm);
-    let led = &mut led::LedHigh::new(ports.pin(gpio::PinId::B0_03));
+    let pin = crate::imxrt1060::gpio::Pin::from_pin_id(gpio::PinId::B0_03);
+    let led = &mut led::LedHigh::new(&pin);
```
hudson-ayers, line 52: this is an important change that we should definitely extend to other boards

hudson-ayers (approved):
This all looks right to me. I still mean to get around to finding a way to initialize peripherals incrementally, so initialization does not require so much stack space (the stack allocation should only be temporary anyway, the actual objects end up stored in lazily initialized globals).

bradjc:
We should merge this so the board works, but it's a bummer to double the stack RAM just for bootup.

hudson-ayers:
An RFC for approaches to fix the issue that necessitated this is here: https://github.com/tock/tock/issues/2425

Issue #2270, opened 2020-12-14
subscribe: validate callback address boundaries?


lschuermann:
While developing the swap constraints of the `subscribe` operation for the Tock 2.0 system call interface, I noticed that a subscribe operation does not validate the address boundaries of the callback in any way.

## Example

A simple dummy application
```c
int main(void) {
  putnstr_async(hello, strlen(hello), (subscribe_cb *) 0x00000004, NULL);
  return 0;
}
```

will actually cause the CPU to actually jump to the specified address (`0x00000004`) in usermode, which in turn panics the application as this is catched by the MPU accordingly:

```
---| Fault Status |---
Instruction Access Violation:       true
Forced Hard Fault:                  true
Fault Status Register (CFSR):       0x00000001
Hard Fault Status Register (HFSR):  0x40000000

---| App Status |---
App: c_hello   -   [StoppedFaulted]
 Events Queued: 0   Syscall Count: 7   Dropped Callback Count: 0
 Restart Count: 5
 Last Syscall: Some(YIELD)

  [...]

  R0 : 0x0000000E    R6 : 0x0003030C
  R1 : 0x00000000    R7 : 0x20004000
  R2 : 0x00000000    R8 : 0x00000000
  R3 : 0x00000000    R10: 0x00000000
  R4 : 0x0003002C    R11: 0x00000000
  R5 : 0x20004000    R12: 0x00000000
  R9 : 0x20004800 (Static Base Register)
  SP : 0x200047D0 (Process Stack Pointer)
  LR : 0x00030141
  PC : 0x00000004
 YPC : 0x00030140

```

## Safety

For chips that have proper memory protection support this is not a safety/security issue, but much rather a question of whether the kernel should catch such invalid values beforehand. I have a strong feeling that returning something like `EINVAL` would be much more appropriate than simply jumping to the specified address, no questions asked. It could be beneficial for a userspace application to be able to handle these errors gracefully, instead of just crashing as soon as the callback is scheduled.

## Outlook for Tock 2.0

Tock 2.0 requires some handling in order to prevent swapping of Callback structs anyways. My question is whether the kernel should sanity-check the values received from userspace and only accept (and invoke the `subscribe` methods) if the callbacks are located in process flash or RAM, or whether we should continue to blindly accept any value passed from userspace here.

If we want to handle this, I can simply add checks as to whether the passed address is in process accessible memory as part of the Tock 2.0 callback swap constraint checks.

alistair23:
[...]

PR #2405, opened 2021-02-03
Tock 2.0: clarify subscribe behavior w.r.t. invalid/refused callback

### LINKED ISSUE #2396: trd-syscall: subscribe callback cancellation ambiguity
The juxtaposition of `immediately return` and `must not invoke the previous callback after the call to subscribe` is possibly ambiguous:

> If the passed callback is not valid (is outside process executable
> memory and is not the Null Callback described below), the kernel MUST
> NOT invoke the requested driver and MUST immediately return a failure
> with a return code of EINVAL.
> 
> A passed callback MUST be valid until the next invocation of `subscribe`
> with the same syscall and driver identifier. When userspace invokes
> subscribe, the kernel MUST cancel all pending callbacks for that driver
> and subscribe identifier: it MUST NOT invoke the previous callback after
> the call to subscribe, and MUST NOT invoke the new callback for events
> that occurred before the call to subscribe.

What is the expected behavior in the case that a second call to subscribe passes an invalid callback? Is the original callback cancelled (and replaced with the null callback?) or is the original callback still active?

lschuermann:
### Pull Request Overview

The description of the Subscribe system call semantics was possibly contradicting itself by stating that, in case of passing an invalid Callback, both

- the driver would never be invoked and
- the previously registered callback must not be invoked after the call to subscribe.

This resolves this issue by clearly stating that only in case of "accepted" callbacks (i.e. the kernel has invoked the driver's `subscribe`method AND the driver has returned the old callback through the `Ok` variant) this guarantee is made.

Fixes #2396
Reported-By: @ppannuto 

### Testing Strategy

N/A

### TODO or Help Wanted

It's going to be very interesting to see how this is actually going to be implemented, as it implicitly imposes some requirements about the precise ordering of when a callback is deemed "valid", when the driver is invoked and when the callback queue is cleared. A valid implementation strategy could look like:
- the kernel checks whether a callback is "valid" (optional, see #2270 for a discussion on whether that's necessary / a good idea)
- the kernel invokes the driver's `subscribe` method
- _if_ the callback has been swapped successfully, the callback queue would be cleared

This would have two major consequences:
- we would need to guarantee that, because during the invocation of `subscribe` the driver has access to both the new and old Callbacks, it cannot queue a callback on the new address and at the same time revoke it
- by clearing the callback queue _after_ the driver's `subscribe` was invoked, any callbacks which were queued while executing `subscribe` would be removed as well

Nonetheless, having a clear definition in the TRD is important. All other ways I can think of to implement something like this would come with similar headaches.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/doc/reference/trd-syscalls.md
+++ b/doc/reference/trd-syscalls.md
@@ -351,14 +351,18 @@ unmodified.
 If the passed callback is not valid (is outside process executable
 memory and is not the Null Callback described below), the kernel MUST
 NOT invoke the requested driver and MUST immediately return a failure
-with a return code of EINVAL.
+with a return code of EINVAL. The currently registered callback
+remains registered and the kernel does not cancel any pending invocations
+of the existing callback.
 
-A passed callback MUST be valid until the next invocation of `subscribe`
-with the same syscall and driver identifier. When userspace invokes
-subscribe, the kernel MUST cancel all pending callbacks for that driver
-and subscribe identifier: it MUST NOT invoke the previous callback after
-the call to subscribe, and MUST NOT invoke the new callback for events
-that occurred before the call to subscribe.
+Any callback passed from a process MUST remain valid until the next successful invocation of
+`subscribe` by that process with the same syscall and driver identifier. When
+a process makes a successful subscribe system call (one which results
+in the `Success with 2 u32` return variant), the kernel MUST cancel
+all pending callbacks on that process for that driver and subscribe identifier: it
+MUST NOT invoke the previous callback after the call to `subscribe`, and
+MUST NOT invoke the new callback for events that the kernel handled before the
+call to `subscribe`.
 
 Note that these semantics create a period over which callbacks might
 be lost: any callbacks that were pending when `subscribe` was called


doc/reference/trd-syscalls.md
```diff
@@ -351,14 +351,17 @@ unmodified.
...
-A passed callback MUST be valid until the next invocation of `subscribe`
-with the same syscall and driver identifier. When userspace invokes
-subscribe, the kernel MUST cancel all pending callbacks for that driver
-and subscribe identifier: it MUST NOT invoke the previous callback after
-the call to subscribe, and MUST NOT invoke the new callback for events
-that occurred before the call to subscribe.
+with a return code of EINVAL. The currently registered callback
+remains unchanged.
+
+A passed callback MUST be valid until the next invocation of
```
jrvanwhy, line 357: I think "valid" should be added after "A passed callback MUST be valid until the next", because an existing callback must remain valid after a failed `subscribe` call.

phil-levis:
Nice catch.

My initial thought is to agree with these semantics. A call to `subscribe` that fails has no effect. Callbacks are not affected/cleared. 

I don't think the semantics will be that difficult. If the kernel decides to reject the call, it doesn't change any pending callbacks. If the call into the driver succeeds, then the kernel has the old callback, and revokes any pending calls on it. Or am I missing something?

doc/reference/trd-syscalls.md
lschuermann: I agree. I've actually changed it so say "accepted" instead of "valid", as it not only matters whether the kernel deems this to be a valid callback, but also whether the capsule accepts the subdriver number, and swaps the callbacks internally.

lschuermann:
> I don't think the semantics will be that difficult. If the kernel decides to reject the call, it doesn't change any pending callbacks. If the call into the driver succeeds, then the kernel has the old callback, and revokes any pending calls on it. Or am I missing something?

The issue that that currently, we seem to currently identify callbacks purely based on their driver and subdriver number:

https://github.com/tock/tock/blob/17e698e8fb2c9628624398435f75db0f5a50dbcd/kernel/src/sched.rs#L819-L831

This way, two things can happen:

- if we invoke a driver's `subscribe` method and it succeeds (i.e. returns `Ok(old_callback)`) we'd have to clear all callbacks (which implies also callbacks scheduled on the new `Callback` instance passed in to subscribe and scheduled during the call, hence would make scheduling a Callback while it is being registered impossible)

- if we invoke a driver's `subscribe` method and it fails (i.e. returns `Err((new_callback, some_errorcode))`), we still MUST clear all callbacks, as we'd otherwise having callbacks scheduled on the new instance, as the driver had access to it in the `subscribe` invocation

Really, the only viable fix seems to be to track callbacks by their `(driver_num, subdriver_num, address, userdata)` 4-tuple, since we can then effectively remove pending callbacks for the old and new instance individually. This previously would not have been possible but is now, since we're getting the old callback back from the driver as part of it's return value and thereby always know the respective 4-tuple to remove from the queue. I'll revise the `subscribe` implementation and integrate these new semantics, as well as (hopefully) preventing swapping of Callbacks in a follow-up PR.

lschuermann:
> Nice catch.

The credit goes to @ppannuto in #2396 though, which I've indicated through the `Reported-by` Git trailer.

alistair23 (approved):

phil-levis:
> Really, the only viable fix seems to be to track callbacks by their `(driver_num, subdriver_num, address, userdata)` 4-tuple, since we can then effectively remove pending callbacks for the old and new instance individually. This previously would not have been possible but is now, since we're getting the old callback back from the driver as part of it's return value and thereby always know the respective 4-tuple to remove from the queue. I'll revise the `subscribe` implementation and integrate these new semantics, as well as (hopefully) preventing swapping of Callbacks in a follow-up PR.

If callbacks could be enqueued preemptively (in interrupt context) then I think this is true. But given that a callback will not be registered while we are doing this operation, why couldn't the code be

``` rust
let res = d.subscribe(subdriver_number, callback, process.appid());
match res {
    Ok(newcb) => { 
        process.remove_pending_callbacks(callback_id);
        newcb.into_subscribe_success(),
    }
    Err((newcb, err)) => newcb.into_subscribe_failure(err),
}
```

Am I missing an edge case?

doc/reference/trd-syscalls.md
```diff
@@ -351,14 +351,17 @@ unmodified.
...
-with a return code of EINVAL.
-
-A passed callback MUST be valid until the next invocation of `subscribe`
-with the same syscall and driver identifier. When userspace invokes
-subscribe, the kernel MUST cancel all pending callbacks for that driver
-and subscribe identifier: it MUST NOT invoke the previous callback after
-the call to subscribe, and MUST NOT invoke the new callback for events
-that occurred before the call to subscribe.
+with a return code of EINVAL. The currently registered callback
+remains unchanged.
```
phil-levis, line 355:
```suggestion
remains registered and the kernel does not cancel any pending invocations
of the existing callback.
```

doc/reference/trd-syscalls.md
```diff
@@ -351,14 +351,17 @@ unmodified.
...
+
+A passed callback MUST be valid until the next accepted invocation of
```
phil-levis, line 357:
```suggestion
Any callback passed from a process MUST remain valid until the next successful invocation of
```

doc/reference/trd-syscalls.md
```diff
@@ -351,14 +351,17 @@ unmodified.
...
+`subscribe` with the same syscall and driver identifier. When
```
phil-levis, line 358:
```suggestion
`subscribe` by that process with the same syscall and driver identifier. When
```

doc/reference/trd-syscalls.md
```diff
@@ -351,14 +351,17 @@ unmodified.
...
+userspace issues an accepted subscribe system call (one which results
```
phil-levis, line 359:
```suggestion
a process makes a successful subscribe system call (one which results
```

doc/reference/trd-syscalls.md
```diff
@@ -351,14 +351,17 @@ unmodified.
...
+in the `Success with 2 u32` return variant), the kernel MUST cancel
+all pending callbacks for that driver and subscribe identifier: it
```
phil-levis, line 361:
```suggestion
all pending callbacks on that process for that driver and subscribe identifier: it
```

doc/reference/trd-syscalls.md
```diff
@@ -351,14 +351,17 @@ unmodified.
...
+MUST NOT invoke the previous callback after the call to subscribe, and
```
phil-levis, line 362:
```suggestion
MUST NOT invoke the previous callback after the call to `subscribe`, and
```

doc/reference/trd-syscalls.md
```diff
@@ -351,14 +351,17 @@ unmodified.
...
+MUST NOT invoke the new callback for events that occurred before the
```
phil-levis, line 363:
```suggestion
MUST NOT invoke the new callback for events that the kernel handled before the
```

doc/reference/trd-syscalls.md
```diff
@@ -351,14 +351,17 @@ unmodified.
...
+call to subscribe.
```
phil-levis, line 364:
```suggestion
call to `subscribe`.
```

phil-levis (commented):
A bunch of further polishing. I'm putting this as a comment so you can apply/tweak/discard/improve as you see fit and not block on me.

phil-levis:
Brief note -- we shouldn't do this now, but I think that it might be useful to distinguish callbacks within the kernel with callbacks to userspace; we could call the former callbacks and the latter upcalls. So we'd change `Callback` to `Upcall`. Worth discussing at some point.

lschuermann:
Thanks for the improvements @phil-levis! I fully agree. I've taken the liberty to squash the 8 single-line GitHub-generated commits and squashed them into one, making you co-author to keep the history clean. Hope that's fine!

ppannuto (approved):
I think this looks good, I don't see any remaining ambiguity.

It's perhaps a bit unfortunate that there's no way to swap callbacks without (possibly) dropping events, but I can't really think of a non-contrived use case that would demand a "hot" swap.

hudson-ayers (approved):

phil-levis:
> It's perhaps a bit unfortunate that there's no way to swap callbacks without (possibly) dropping events, but I can't really think of a non-contrived use case that would demand a "hot" swap.

I think the document does talk about this. If you need a hot swap like this, then the thing to do is do it in userspace. I.e., have your upcall function call a hot-swappable function pointer.

PR #2407, opened 2021-02-03
Tock 2.0: port #2383 (kernel: Remove setting/tracking the stack pointer in process.rs)

lschuermann:
### Pull Request Overview

This pull request ports the changes of #2383 ("kernel: Remove setting/tracking the stack pointer in process.rs", thanks @bradjc!) to `tock-2.0-dev`.

This is a draft PR, because
- the changes are not yet tested
- I believe it would be a good idea to rework the stack management as already proposed in the `syscall.rs` implementations for `rv32i` and `cortex-m` in this PR, such that processes start preallocated with the smallest stack possible. I'd try to implement this in the coming days for `libtock-c` and update this PR accordingly.
- ...with that implemented, the documentation can be updated as part of this PR as well.

I'm opening this PR now to indicate that at least someone is working on it :smile:.

I'd be happy about feedback regarding the merge result.

### Testing Strategy

This pull request *is not yet tested*.

### TODO or Help Wanted

This pull request still needs actually changing the initial stack allocation for `rv32i` and `cortex-m`, along with userspace support and testing.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/syscall.rs
+++ b/arch/rv32i/src/syscall.rs
@@ -52,12 +52,29 @@ impl SysCall {
 impl kernel::syscall::UserspaceKernelBoundary for SysCall {
     type StoredState = Riscv32iStoredState;
 
+    fn initial_process_app_brk_size(&self) -> usize {
+        // TOCK 1.X
+        //
+        // We do not need any memory to start with, but the 1.x Tock kernel
+        // allocates at least 3 kB to processes, and we need to ensure that
+        // happens as userspace may expect it.
+        3 * 1024
+
+        // TOCK 2.0
+        //
+        // The RV32I UKB implementation does not use process memory for any
+        // context switch state. Therefore, we do not need any process-accessible
+        // memory to start with to successfully context switch to the process the
+        // first time.
+        //0
+    }
+
     unsafe fn initialize_process(
         &self,
-        stack_pointer: *const usize,
-        _stack_size: usize,
+        accessible_memory_start: *const u8,
+        _app_brk: *const u8,
         state: &mut Self::StoredState,
-    ) -> Result<*const usize, ()> {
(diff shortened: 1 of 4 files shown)


hudson-ayers:
@lschuermann if I review and test this as-is, are you okay with merging it in now and putting off your second two bullets for another PR (when you or anyone else has time)? I would like to merge master into tock-2.0-dev again now that we are finishing up all the system call ports, and this is the primary blocker on that

arch/rv32i/src/syscall.rs
```diff
@@ -52,32 +52,56 @@ impl SysCall {
...
         state.regs.iter_mut().for_each(|x| *x = 0);
         state.pc = 0;
         state.mcause = 0;
 
         // The first time the process runs we need to set the initial stack
         // pointer in the sp register.
-        state.regs[R_SP] = stack_pointer as u32;
+        //
+        // TOCK 1.X
+        state.regs[R_SP] = accessible_memory_start.add(3 * 1024) as u32;
```
hudson-ayers, line 87: Until we switch to the tock 2.0 version, we technically need a check here that there is enough stack space for this to be safe I think. But maybe that doesn't matter for the 2.0-dev branch, because we should be switching to the 2.0 version before this will actually be used for anything

hudson-ayers (approved):
I looked through the merge and it looks right to me, though this is pretty complex code with lots of subtle interactions.

I tested this code on Imix and successfully ran:
blink + hello_loop + stack_size_test01 + stack_size_test02
mpu_stack_growth
mpu_walk_region

bradjc:
This is not going to get easier to merge, so let's get this in soon. Thanks @hudson-ayers for looking this over. Marking last call now. We can still do any cleanup in the 2.0 process.

PR #2409, opened 2021-02-05
Port MAX17205 system call capsule to 2.0

alevy:
### Pull Request Overview

This pull request ports the MAX17205 "Fuel Guage" system call capsule to the new 2.0 driver interface.

This capsule is _not_ multi-process friendly, so this port is a tad bit
kludgey (in particular, need to use `MapCell` to store the callback,
unlike with other capsules).

I've also left ReturnCode in place for the internal interface since the
client/driver interface can probably generally be improved with the
removal of ReturnCode, but it's a fairly breaking change.

### Testing Strategy

This driver is only currently used (as far as we know) on Signpost. Luckily, @ppannuto has volunteered to help test!

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/max17205.rs
+++ b/capsules/src/max17205.rs
@@ -38,9 +38,9 @@
 //! ```
 
 use core::cell::Cell;
-use kernel::common::cells::{OptionalCell, TakeCell};
+use kernel::common::cells::{MapCell, OptionalCell, TakeCell};
 use kernel::hil::i2c;
-use kernel::{AppId, Callback, LegacyDriver, ReturnCode};
+use kernel::{AppId, Callback, CommandResult, Driver, ErrorCode, ReturnCode};
 
 /// Syscall driver number.
 use crate::driver;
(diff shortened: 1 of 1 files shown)


capsules/src/max17205.rs
```diff
@@ -419,17 +419,23 @@ impl LegacyDriver for MAX17205Driver<'_> {
...
-                self.callback.insert(callback);
-                ReturnCode::SUCCESS
+                if let Some(prev) = self.callback.replace(callback) {
+                    Ok(prev)
+                } else {
+                    // TODO(alevy): This should never happen because we start with a full MapCell
+                    // and only ever replace it. This is just defensive until this module becomes
+                    // multi-user, which will preclude the need for a MapCell in the first place.
+                    Ok(Callback::default())
+                }
```
lschuermann, line 434: Given we initialize the `MapCell` with a `Callback` and only ever swap one for another, this case should -- as you state -- never happen. I would've expected to just panic (`.expect`) instead of failing silently, as something is clearly off then. But I don't have any strong opinion about it, and this is a fair documentation placeholder :smile:.

lschuermann (approved):
Looks good, although I didn't test it. There might be some changes required to how non-virtualized capsules work w.r.t. handling their capsules in the future, but MapCell is the way to go currently.

capsules/src/max17205.rs
alevy: Having the fallback case seemed more resilient to me. In particular, I can imagine someone might come back to this capsule at some point and notice that the `MapCell` is initialized to some strange default value rather than `empty` the way it's used everywhere else. If they decide to change that without carefully reasoning about the particular semantics of this capsule, and none of us catch it... everything will still work exactly as intended because of this fallback case.

hudson-ayers (approved):

phil-levis:
As per discussions in the core call, we'll merge the Signpost drivers and test them when we get to merging/testing time.

PR #2411, opened 2021-02-05
Port BLE advertising to new 2.0 driver interface

alevy:
### Pull Request Overview

This pull request ports the BLE advertising system call capsule to the new 2.0 driver interface. Most of the port is straight forward translation of `ReturnCode` to `CommandResult` or Result<T, (T, ErrorCode)>, with two noteworthy points:

1. I moved the advertising data to allow_readonly insead of
   allow_readwrite, since it does _not_ need to be writable (and indeed
   putting advertising data in the code segment makes sense). This is a
   breaking change for the userland API, but we're breaking the userland
   API anyway for 2.0.

 2. I did _not_ change the BLE advertising HIL, which includes a bunch
    of ReturnCode's. So ReturnCode is still in there.

### Testing Strategy

Tested with libtock-c changes in tock/libtock-c#173

### TODO or Help Wanted

Port libtock-c counterpart

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/ble_advertising_driver.rs
+++ b/capsules/src/ble_advertising_driver.rs
@@ -10,13 +10,14 @@
 //! Data payloads are limited to 31 bytes since the maximum advertising channel
 //! protocol data unit (PDU) is 37 bytes and includes a 6-byte header.
 //!
-//! ### Allow system call
+//! ### Allow system calls
 //!
-//! The allow systems calls are used for buffers from allocated by userland
+//! There is one ReadWrite and one ReadOnly allow buffers, both at index `0`.
 //!
-//! There are two different buffers:
-//! * 0: Advertising data
-//! * 1: Passive scanning buffer
+//! * ReadOnly: Advertising data, containing the full _payload_ (i.e. excluding the header) the
+//!             process wishes to advertise.
+//! * ReadWrite: Passive scanning buffer, which is populated during BLE scans with complete (i.e.
+//!              including headers) advertising packets received on channels 37, 38 and 39.
 //!
 //! The possible return codes from the 'allow' system call indicate the following:
 //!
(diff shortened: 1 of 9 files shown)


capsules/src/ble_advertising_driver.rs
```diff
@@ -452,20 +455,16 @@ where
...
-                        .scan_buffer
-                        .as_mut()
-                        .map(|userland| {
-                            for (dst, src) in userland.iter_mut().zip(buf[0..len as usize].iter()) {
-                                *dst = *src;
-                            }
-                        })
-                        .is_some();
+                    let success = app.scan_buffer.mut_map_or(false, |userland| {
+                        for (dst, src) in userland.iter_mut().zip(buf[0..len as usize].iter()) {
```
niklasad1, line 459: I think we could use https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice instead of the loop with manual copying.

`userland[0..len as usize].copy_from_slice(&buf[0..len as usize])`

lschuermann: Agreed. For some edge cases, where this doesn't work (when you don't have raw access to the slice), I've verified that `.for_each(|(dst, src)| *dst = *src)` will be optimized to essentially a memcpy, but the Rust board-utils seem preferable  where applicable.

capsules/src/ble_advertising_driver.rs
```diff
@@ -630,75 +629,95 @@ where
...
+
+    fn allow_readwrite(
+        &self,
+        appid: kernel::AppId,
+        allow_num: usize,
+        mut slice: ReadWriteAppSlice,
+    ) -> Result<ReadWriteAppSlice, (ReadWriteAppSlice, ErrorCode)> {
+        let res = match allow_num {
             // Passive scanning buffer
             1 => self
```
lschuermann, line 680: We might want to change the index of the `allow_readwrite` buffers to start at `0` as well.

lschuermann (commented):
LGTM, except for the nitpicks mentions. Yet to test it though.

capsules/src/ble_advertising_driver.rs
alevy: Agree, good catch

alevy:
Addressed @niklasad1 and @lschuermann's comments

lschuermann (approved):

niklasad1 (commented):
LGTM but I haven't follow the `tock 2.0 design` and don't feel comfortable signing off on it :)

hudson-ayers (approved):

phil-levis:
Is this tested? Would be good to have libtock-c updated as well.

alevy:
@phil-levis doing all of the today

alistair23 (approved):

alevy:
Tested with libtock-c changes in tock/libtock-c#173

PR #2420, opened 2021-02-09
RISC-V: Add support for detecting locked PMP regions, no need to sort

alistair23:
### Pull Request Overview

Add support for dynamically detecting the number of PMP regions and determining if they are locked.

We also no longer sort the PMP regions:

The RISC-V spec describes TOR like this:

```
If TOR is selected, the associated address register forms the top of the address range, and the
preceding PMP address register forms the bottom of the address range. If PMP entry i’s A field is
set to TOR, the entry matches any address y such that pmpaddri−1 ≤ y < pmpaddri (irrespective of
the value of pmpcfgi−1 ). If PMP entry 0’s A field is set to TOR, zero is used for the lower bound,
and so it matches any address y < pmpaddr0.
```

The inital PMP implementation mis-interpreted this and required sorting
the regions and specifically made the previous region match AND then
disabled read/write/access.

This PR changes that operation. Now instead we do the following.

For a region i, we set pmpaddr[i-1] as the start address and pmpaddr[i]
as the end address (same as before). We then set pmpcfg[i-1] to be
disabled, so there is no region there. We then set pmpcfg[i] as TOR and
it matches between pmpaddr[i-1] and pmpaddr[i].

The advantage with this new approach is we no longer have to sort the
regions (which can take time). The main advantage of sorting the regions
is that we could share addresses, for example if a start and end address
are the same we can save a PMP regions by re-using those two addresses.

In reality though this doesn't happen for apps (as flash and memory are
split) and we don't support it anyway.

Sorting the regions will also break if we have a locked region in the
middle of the list.

### Testing Strategy

Running an app on OpenTitan and QEMU.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -255,47 +255,6 @@ impl<const NUM_REGIONS_OVER_TWO: usize> PMPConfig<NUM_REGIONS_OVER_TWO> {
         }
         None
     }
-
-    fn sort_regions(&mut self) {
-        // Get the app region address
-        let app_addres = if self.app_memory_region.is_some() {
-            Some(
-                self.regions[self.app_memory_region.unwrap_or(0)]
-                    .unwrap()
-                    .location
-                    .0,
-            )
-        } else {
-            None
-        };
-
-        // Sort the regions
-        self.regions.sort_unstable_by(|a, b| {
-            let (a_start, _a_size) = match a {
-                Some(region) => (region.location().0 as usize, region.location().1),
-                None => (0xFFFF_FFFF, 0xFFFF_FFFF),
-            };
-            let (b_start, _b_size) = match b {
-                Some(region) => (region.location().0 as usize, region.location().1),
-                None => (0xFFFF_FFFF, 0xFFFF_FFFF),
-            };
-            a_start.cmp(&b_start)
-        });
-
-        // Update the app region after the sort
-        if app_addres.is_some() {
-            for (i, region) in self.regions.iter().enumerate() {
-                match region {
-                    Some(reg) => {
-                        if reg.location.0 == app_addres.unwrap() {
-                            self.app_memory_region.set(i);
-                        }
-                    }
-                    None => {}
-                }
-            }
-        }
-    }
 }
 
 impl<const NUM_REGIONS_OVER_TWO: usize> kernel::mpu::MPU for PMP<NUM_REGIONS_OVER_TWO> {
@@ -414,8 +373,6 @@ impl<const NUM_REGIONS_OVER_TWO: usize> kernel::mpu::MPU for PMP<NUM_REGIONS_OVE
         config.regions[region_num] = Some(region);
         config.is_dirty.set(true);
 
-        config.sort_regions();
-
         Some(mpu::Region::new(start as *const u8, size))
     }
 
@@ -486,8 +443,6 @@ impl<const NUM_REGIONS_OVER_TWO: usize> kernel::mpu::MPU for PMP<NUM_REGIONS_OVE
 
         config.app_memory_region.set(region_num);
 
-        config.sort_regions();
-
         Some((region_start as *const u8, region_size))
     }
 
@@ -524,8 +479,6 @@ impl<const NUM_REGIONS_OVER_TWO: usize> kernel::mpu::MPU for PMP<NUM_REGIONS_OVE
         config.regions[region_num] = Some(region);
         config.is_dirty.set(true);
 
-        config.sort_regions();
-
         Ok(())
     }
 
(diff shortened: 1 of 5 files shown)


bradjc (commented):
Didn't test but looks good. Question: why remove `NUM_REGIONS`? I know it's a bit redundant, but from a chip setup perspective the chip knows the number of PMP regions, and not necessarily the PMP implementation. If we wanted to change the PMP implementation in the future it might make sense to configure the total number of regions.

chips/earlgrey/src/chip.rs
```diff
@@ -16,7 +16,7 @@ use crate::plic::PLIC;
 
 pub struct EarlGrey<'a, A: 'static + Alarm<'static>, I: InterruptService<()> + 'a> {
     userspace_kernel_boundary: SysCall,
-    pmp: PMP<4, 2>,
+    pmp: PMP<8>,
```
hudson-ayers, line 19: did we just have the wrong value here before?

arch/rv32i/src/pmp.rs
```diff
@@ -47,19 +38,75 @@ register_bitfields![u8,
...
+            // Flip R, W, X bits
+            let pmpcfg_new = pmpcfg_og ^ (3 << ((i % 4) * 8));
+            csr::CSR.pmpcfg[i / 4].set(pmpcfg_new);
+
+            // Check if the bits are set
+            let pmpcfg_check = csr::CSR.pmpcfg[i / 4].get();
+
+            if pmpcfg_check == pmpcfg_og {
+                // We didn't correctly flip all of the bits
+                if pmpcfg_og & ((1 << 7) << ((i % 4) * 8)) > 0 {
```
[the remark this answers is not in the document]
alistair23: This checks if the bit is locked

chips/earlgrey/src/chip.rs
alistair23: No, 4 is the number of regions that OT currently has.

The idea of bumping this up to 16 is that we can auto-detect the number between 0 and 16.

arch/rv32i/src/pmp.rs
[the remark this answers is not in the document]
bradjc: Isn't that what the `if` on 91 does?

chips/earlgrey/src/chip.rs
hudson-ayers: Oh, I see. I could see this being misleading to people who might look at this file and assume OT has 16 regions. Perhaps we should rename that const generic variable?

arch/rv32i/src/pmp.rs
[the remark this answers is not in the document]
alistair23: There are two checks.

The first checks if all of the bits we set stuck. If they didn't stick it could be because the region is locked. or because the reason isn't implemented. So this second check tells if it's locked.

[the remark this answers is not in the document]
alistair23: I have updated the comments to make this clear

chips/earlgrey/src/chip.rs
alistair23: The comment in PMP should describe this

Any ideas on a better name?

hudson-ayers: perhaps `MAX_AVAILABLE_REGIONS_OVER_TWO`? I realize that is a bit long.

alistair23: Done!

bradjc: The hardware should be fixed, and I think adding uncertainty in the name makes it confusing. OT is a unique case for now, but eventually the hardware will be fixed.

alistair23: I don't think it should be though.

The RISC-V spec specifically allows any number of PMP regions. It's entirely possible there could be different chips in the same family with different number of regions. Tock can determine the number available at run time.

There are also a few FPGA implementations supported in Tock which can be adjusted and also detected at run time.

arch/rv32i/src/pmp.rs
```diff
@@ -47,19 +38,84 @@ register_bitfields![u8,
...
 /// Main PMP struct.
-pub struct PMP<const NUM_REGIONS: usize, const NUM_REGIONS_OVER_TWO: usize> {
+///
+/// Tock will ignore locked PMP regions. Note that Tock will not make any
+/// attempt to avoid access faults from locked regions.
+///
+/// `MAX_AVAILABLE_REGIONS_OVER_TWO`: The number of PMP regions divided by 2.
+///  The RISC-V spec mandates that there must be either 0, 16 or 64 PMP
+///  regions implemented. If you are using this PMP struct we are assuming
+///  there is more then 0 implemented. So this value should be either 8 or 32.
```
bradjc, line 48:
```suggestion
///  there are more than 0 implemented. So this value should be either 8 or 32.
```

bradjc (approved):

alistair23:
I didn't remove the concept of `NUM_REGIONS`.

Ideally chips would either specify 16 or 64 regions (as defined in the RISC-V spec) and Tock can dynamically determine how many regions are available (that is what we do on OpenTitan after this PR, as OT currently has 4, but plans to have 16).

This does unfortunately waste space as we then allocate elements in an array to handle those regions. So `Chips` that know they will only ever have some smaller number can specify that (just like `NUM_REGIONS`).

bradjc (approved):

PR #2423, opened 2021-02-10
RISC-V: Add support for ePMP

alistair23:
### Pull Request Overview

This PR adds support for [ePMP](https://docs.google.com/document/d/1Mh_aiHYxemL0umN3GTTw8vsbmzHZ_nxZXgjgOUzbvc8/edit#) to RISC-V.

ePMP can be used to apply memory protections to the kernel.

The expectation is that for OpenTitan ePMP will be enabled before Tock runs, so we must support it.

### Testing Strategy

Currently it's not connected to anything, but I have tested it on the QEMU implementation of ePMP.

libtock-c tests:
 - [X] mpu_flash_end correctly panics
 - [x] mpu_ram_end fails on both PMP and ePMP
 - [X] mpu_stack_growth overflows the stack
 - [X] mpu_walk_region (stack needs to be increased to 2048)
     - Doesn't panic if overrun isn't set
     - panics at the correct flash and ram addresses if overrun is set

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


silvestrst:
I didn't have chance to review it properly yet, but does this change support ePMP `MMWP` mode (deny by default)?

alistair23:
Tock sets `MMWP`. I haven't tested if MMWP is already set, but that should be a pretty simple change.

alistair23:
Rebased and ready to merge!

hudson-ayers (approved):
This mostly looks fine. To anyone reviewing, I recommend reviewing locally using
```bash
$ git diff --no-index pmp.rs epmp.rs
```

as `epmp.rs` is largely a duplicate of `pmp.rs`.

I do worry some that `epmp.rs` being mostly a duplicate will make it easy for these two files to become out of sync, much like we saw happen for `cortexm7/mpu.rs` when it was a separate file. But I am not familiar enough with PMP /ePMP to know whether removing that duplication is feasible.

Another comment, in light of the recent discussion on system call overhead: this does add O(N/2) cycles to every system call where N == number of PMP regions.

bradjc:
I think we should use the epmp somewhere if we are going to merge it. I would also like to see the type names changed if it is going to be a separate implementation, as otherwise it makes it very easy for someone to look at the wrong code and not realize it.

This seems difficult to maintain, and inevitably bugs will be fixed in one implementation but not the other. Are we ok with that?

alistair23:
ePMP is not yet supported in the Ibex core (at least from what I can tell) so I can't enable it yet. I agree that it's not ideal.

As for sharing a lot of the code, there are enough differences that I couldn't see how we could re-use the PMP implementation and just add extra functionality. The main way I can think of is using `cfg`, but that's a little messy.

alistair23:
I ran some libtock-c tests and this passes all the same tests as the current PMP implementation.

alistair23:
Ping!

bradjc (approved):

PR #2424, opened 2021-02-10
kernel: sched: Allow Platform to handle app faults

alistair23:
### Pull Request Overview

This commit adds a function to the Platform trait that can be used to allow 
a Platform to handle an app fault. This avoids terminating or restarting and
app and instead allows it to continue execution.

This is an optional function that can be implemented by `Chip`s that
allows the chip to handle the app fault and not terminate or restart
the app.

If `Ok(())` is returned by this function then the kernel will not
terminate or restart the app, but instead allow it to continue
running. NOTE in this case the chip must have fixed the underlying
reason for fault otherwise it will re-occur.

This can not be used to get around Tock's protections. If for example
this function just ignored the error and allowed the app to continue
the fault would continue to occur,

If `Err(())` is returned then the kernel will set the app as faulted
and follow the `FaultResponse` protocol.

It is unlikey a `Chip` will need to implement this. This should be used
for only a handul of use cases. Possible use cases include:
- Allowing the kernel to emulate un-implemented instructions
  This could be used to allow apps to run on hardware that doesn't
  implement some instructions, for example atomics.
- Allow the kernel to handle hardware faults, triggered by the app
  This can allow an app to continue running if it triggers certain
  types of faults. For example if an app triggers a memory parity
  error the kernel can handle the error and allow the app to
  continue (or not).
- Allow an app to execute from external QSPI
  This could be used to allow an app to execute from external QSPI
  where access faults can be handled by the `Chip` to ensure the
  QPSI is mapped correctly.

### Testing Strategy

Running in the CI

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/kernel/src/platform/mod.rs
+++ b/kernel/src/platform/mod.rs
@@ -62,6 +62,43 @@ pub trait Platform {
     ) -> Result<(), errorcode::ErrorCode> {
         Ok(())
     }
+
+    /// This function is called when an app faults.
+    ///
+    /// This is an optional function that can be implemented by `Platform`s that
+    /// allows the chip to handle the app fault and not terminate or restart
+    /// the app.
+    ///
+    /// If `Ok(())` is returned by this function then the kernel will not
+    /// terminate or restart the app, but instead allow it to continue
+    /// running. NOTE in this case the chip must have fixed the underlying
+    /// reason for fault otherwise it will re-occur.
+    ///
+    /// This can not be used for apps to circumvent Tock's protections. If
+    /// for example this function just ignored the error and allowed the app
+    /// to continue the fault would continue to occur.
+    ///
+    /// If `Err(())` is returned then the kernel will set the app as faulted
+    /// and follow the `FaultResponse` protocol.
+    ///
+    /// It is unlikey a `Platform` will need to implement this. This should be used
+    /// for only a handul of use cases. Possible use cases include:
+    ///    - Allowing the kernel to emulate unimplemented instructions
+    ///      This could be used to allow apps to run on hardware that doesn't
+    ///      implement some instructions, for example atomics.
+    ///    - Allow the kernel to handle hardware faults, triggered by the app.
+    ///      This can allow an app to continue running if it triggers certain
+    ///      types of faults. For example if an app triggers a memory parity
+    ///      error the kernel can handle the error and allow the app to
+    ///      continue (or not).
+    ///    - Allow an app to execute from external QSPI.
+    ///      This could be used to allow an app to execute from external QSPI
+    ///      where access faults can be handled by the `Platform` to ensure the
+    ///      QPSI is mapped correctly.
+    #[allow(unused_variables)]
+    fn process_fault_hook(&self, process: &dyn process::ProcessType) -> Result<(), ()> {
+        Err(())
+    }
 }
 
 /// Interface for individual MCUs.
--- a/kernel/src/sched.rs
+++ b/kernel/src/sched.rs
@@ -627,8 +627,12 @@ impl Kernel {
                     // why and handle the process as appropriate.
                     match context_switch_reason {
                         Some(ContextSwitchReason::Fault) => {
-                            // Let process deal with it as appropriate.
-                            process.set_fault_state();
+                            // The app faulted, check if the chip wants to
+                            // handle the fault.
+                            if platform.process_fault_hook(process).is_err() {
+                                // Let process deal with it as appropriate.
+                                process.set_fault_state();
+                            }
                         }
                         Some(ContextSwitchReason::SyscallFired { syscall }) => {
                             self.handle_syscall(platform, process, syscall);
--- a/kernel/src/syscall.rs
+++ b/kernel/src/syscall.rs
@@ -400,6 +400,9 @@ pub enum ContextSwitchReason {
     /// Process called a syscall. Also returns the syscall and relevant values.
     SyscallFired { syscall: Syscall },
     /// Process triggered the hardfault handler.
+    /// The implementation should still save registers in the event that the
+    /// `Platform` can handle the fault and allow the app to continue running.
+    /// For more details on this see `Platform::process_fault_hook()`.
     Fault,
     /// Process interrupted (e.g. by a hardware event)
     Interrupted,


bradjc:
1. Why in `Chip` and not `Platform`?

2. It seems like a lot of tock layers would be involved to potentially handle a fault in some meaningful way. Is it viable to actually implement a fault fixup handler in Tock? Or would it be too messy / difficult to reason about?

3. Does the UKB interface work if this function returns `Ok()`? As in, does the app run correctly again?

alistair23:
1. I think this would be more of a `Chip` specific feature. For example an SoC would need to use this for one of the reasons above. It doesn't seem as much of a `Platform` requirement. `Chip` also handles all of the other types of faults, so it feels more at home there. I don't really mind either way, I can see a justification for either.

2. I don't think it would be too hard. The `Chip` can figure out the reason (the same way the panic handler currently does) and then address the issue. For example for the QSPI justification it could call a function in the QPSI driver to allocate that memory and then let the app continue running.

3. Yes. The app will pickup from where it left off and the sched loop in the kernel will have setup the MPU/PMP and everything else as it always does.

phil-levis:
>     1. I think this would be more of a `Chip` specific feature. For example an SoC would need to use this for one of the reasons above. It doesn't seem as much of a `Platform` requirement. `Chip` also handles all of the other types of faults, so it feels more at home there. I don't really mind either way, I can see a justification for either.

I think it makes a lot more sense in Platform. It's something that's bound to a particular build of a kernel for an SoC, not for a particular SoC. 

It does seem strange for a Platform to be able to do this though, because in many cases the decision to do so, or what to do, might be application-specific. The analogy to this is handling SIGSEGV in POSIX. There are programs that do this (e.g., some runtimes with managed memory/garbage collection).

Before we add a feature like this that changes some core aspects of the kernel, I'd like to know what the exact problems it's solving are. Adding lots of features because they could be useful in some cases complicates the kernel. It could be that the exact set of problems we're worried about could be solved in a different way that's less invasive. I worry about a kernel inadvertently allowing faulting applications to continue and entering an infinite fault loop.

ppannuto:
I agree that the choice to wire up this handling is probably a `Platform` question, but can also see where it may be likely that the actual implementation lives in a `Chip`.

QSPI is perhaps a good example of this, though I'm not sure I quite understand the situation it presents: is the thinking that application code will be (partially?) in external memory, and that if the application jumps there the kernel should (transparently?) remap the external memory into the application-accessible space? Why would an application not explicitly ask the QSPI driver to do this mapping as part of the application startup/setup code?

alistair23:
I don't think this should be exposed to apps. That is an app should have no idea if this is happening or have any control over it.

I'm happy to move it to `Platform` if that is what everyone thinks.

For the QSPI example the idea is that an application will either be running partially or fully from QSPI. When the app loads some of it's data from QSPI there could be a fault as the memory isn't mapped. The kernel then changes some hardware registers to map the data and continues the app. I have heard there are chips that do this, but I wasn't able to find any though. The application wouldn't ask the kernel for anything, because it is just executing out of what it thinks is flash. We don't want the app to be custom written to run from QSPI.

The better example for me is a parity error. OT has SRAM parity. If an app accesses it's valid memory, but there is a parity error the kernel should be able to handle it. Maybe the error can be corrected by the kernel allowing the app to continue running, maybe it can't be in which case the kernel can terminate the app.

ppannuto:
The parity example makes a lot of sense; I can start to imagine other neat optimizations this might enable (e.g. powering off a peripheral an application hasn't used in a while and silently powering it back on).

As this is essentially an exception handling mechanism, I think we'll need to encode some notion of synchronous / asynchronous exception (to borrow the ARM lingo). I.e., when a `svc` triggers a "fault", we return to userspace at the _next_ instruction, after the `svc`. If something like a memory read causes a fixable fault, we'll need to return to userspace _at_ the faulting instruction so that it can be successfully retried. I think encoding the "where to return to" question will likely have to integrate with the UKB interface, though I'm not sure I have an immediate instinct for the right way to do that.

alistair23:
The peripheral use case only works for things not accesses via drivers. But a good example would also be powering off memory banks when not in use. The Apollo3 can disable SRAM banks to reduce power usage.

My thinking about "where to return" was just to not touch the PC. That is the idea is we are fixing something that causes a fault and re-running the instruction. We aren't skipping the instruction. In saying that we are passing the `ProcessType` down to this function, which has access to the `stored_state` so we could increase the PC.

alistair23:
I have updated this to use `Platform` instead of `Chip`

phil-levis:
OK, this makes a lot more sense to me now. The use case is that the kernel should handle faults and be able to decide what to do with them. In most cases it will terminate the application, but you want to be able to specify what other additional handling you might want.

@ppannuto I think the hardware handles the next instruction question for us? The `ReturnAddress()` for `svc` is `NextInstrAddr()` while for a synchronous hard fault it is `ThisInstrAddr()`. Is there a case when a process could cause a fault in which we would want it to behave differently than the ARM spec?

phil-levis (approved):

ppannuto:
I think the canonical example is an asychronous bus fault (i.e. writing to a MMIO peripheral that buffers or delays the write on the peripheral end, delaying the exception), but that's not a common case; it's probably premature to worry about doing anything special prior to finding a use case that demands it.

kernel/src/sched.rs
```diff
@@ -623,8 +623,12 @@ impl Kernel {
...
                     match context_switch_reason {
                         Some(ContextSwitchReason::Fault) => {
-                            // Let process deal with it as appropriate.
-                            process.set_fault_state();
+                            // The app faulted, check if the chip wants to
+                            // handle the fault.
+                            if platform.process_app_fault(process).is_err() {
+                                // Let process deal with it as appropriate.
+                                process.set_fault_state();
+                            }
```
ppannuto, line 631: Possibly more of a style question than a correctness one, should this include a `continue` here in the case where the exception was handled by the platform? We include (effectively unused, since the match is the last statement in this loop) `continue`s in all of the match/branches below. I _think_ the intent is to be specific about 'this case means keep running this process (`continue`) or stop (`break`).

kernel/src/platform/mod.rs
```diff
@@ -60,6 +60,43 @@ pub trait Platform {
     ) -> Result<(), returncode::ReturnCode> {
         Ok(())
     }
+
+    /// This function is called when an app faults
+    ///
+    /// This is an optional function that can be implemented by `Chip`s that
```
bradjc, line 66:
```suggestion
    /// This is an optional function that can be implemented by `Platform`s that
```

kernel/src/platform/mod.rs
```diff
@@ -60,6 +60,43 @@ pub trait Platform {
...
+    /// reason for fault otherwise it will re-occur.
+    ///
+    /// This can not be used to get around Tock's protections. If for example
+    /// this function just ignored the error and allowed the app to continue
+    /// the fault would continue to occur,
+    ///
+    /// If `Err(())` is returned then the kernel will set the app as faulted
+    /// and follow the `FaultResponse` protocol.
+    ///
+    /// It is unlikey a `Chip` will need to implement this. This should be used
```
bradjc, line 82:
```suggestion
    /// It is unlikey a `Platform` will need to implement this. This should be used
```

kernel/src/platform/mod.rs
```diff
@@ -60,6 +60,43 @@ pub trait Platform {
...
+    ///      This can allow an app to continue running if it triggers certain
+    ///      types of faults. For example if an app triggers a memory parity
+    ///      error the kernel can handle the error and allow the app to
+    ///      continue (or not).
+    ///    - Allow an app to execute from external QSPI
+    ///      This could be used to allow an app to execute from external QSPI
+    ///      where access faults can be handled by the `Chip` to ensure the
+    ///      QPSI is mapped correctly.
+    #[allow(unused_variables)]
+    fn process_app_fault(&self, process: &dyn process::ProcessType) -> Result<(), ()> {
```
bradjc, line 97: This name is a little confusing. Is process a noun or a verb here? Maybe `process_fault_hook()`?

alistair23: It was a verb, I have changed it to `process_fault_hook()`

ppannuto (approved):

kernel/src/platform/mod.rs
```diff
@@ -60,6 +60,43 @@ pub trait Platform {
     ) -> Result<(), returncode::ReturnCode> {
         Ok(())
     }
+
+    /// This function is called when an app faults
```
bradjc, line 64:
```suggestion
    /// This function is called when an app faults.
```

kernel/src/platform/mod.rs
```diff
@@ -60,6 +60,43 @@ pub trait Platform {
...
+    ///
+    /// This is an optional function that can be implemented by `Platform`s that
+    /// allows the chip to handle the app fault and not terminate or restart
+    /// the app.
+    ///
+    /// If `Ok(())` is returned by this function then the kernel will not
+    /// terminate or restart the app, but instead allow it to continue
+    /// running. NOTE in this case the chip must have fixed the underlying
+    /// reason for fault otherwise it will re-occur.
```
bradjc, line 73: I think we need to update the comments for `ContextSwitchReason` to clarify that when returning the fault case the UKB implementation should leave things in a state where the kernel can re-run the last instruction.

kernel/src/platform/mod.rs
```diff
@@ -60,6 +60,43 @@ pub trait Platform {
...
+    ///
+    /// This can not be used to get around Tock's protections. If for example
+    /// this function just ignored the error and allowed the app to continue
+    /// the fault would continue to occur,
```
bradjc, line 77:
```suggestion
    /// the fault would continue to occur.
```

bradjc, line 77: This is maybe too strong of a claim. A UKB implementation could somewhat easily end up running the next instruction after the one that caused the fault if they are not careful. Do we need to specify this?

kernel/src/platform/mod.rs
```diff
@@ -60,6 +60,43 @@ pub trait Platform {
...
+    ///
+    /// If `Err(())` is returned then the kernel will set the app as faulted
+    /// and follow the `FaultResponse` protocol.
+    ///
+    /// It is unlikey a `Platform` will need to implement this. This should be used
+    /// for only a handul of use cases. Possible use cases include:
+    ///    - Allowing the kernel to emulate un-implemented instructions
```
bradjc, line 84:
```suggestion
    ///    - Allowing the kernel to emulate unimplemented instructions.
```

kernel/src/platform/mod.rs
```diff
@@ -60,6 +60,43 @@ pub trait Platform {
...
+    ///      This could be used to allow apps to run on hardware that doesn't
+    ///      implement some instructions, for example atomics.
+    ///    - Allow the kernel to handle hardware faults, triggered by the app
```
bradjc, line 87:
```suggestion
    ///    - Allow the kernel to handle hardware faults, triggered by the app.
```

kernel/src/platform/mod.rs
```diff
@@ -60,6 +60,43 @@ pub trait Platform {
...
+    ///      This can allow an app to continue running if it triggers certain
+    ///      types of faults. For example if an app triggers a memory parity
+    ///      error the kernel can handle the error and allow the app to
+    ///      continue (or not).
+    ///    - Allow an app to execute from external QSPI
```
bradjc, line 92:
```suggestion
    ///    - Allow an app to execute from external QSPI.
```

kernel/src/platform/mod.rs
alistair23: I have clarified this so that apps can't get around the protections. The `Platform` can avoid the protections

bradjc (dismissed):

alistair23:
This should be ready to be merged

alistair23:
Now that 2.0 is merged this is ready to go!

ppannuto (approved):

alistair23:
Ping!

lschuermann (approved):
I like the idea of being able to supplement architecture / chip features with this. Looks good to me, in particular I don't see how this would hurt anyone.

hudson-ayers (approved):

bradjc (approved):

PR #2431, opened 2021-02-15
Polishing pass on Tock 2.0 system call TRD.

phil-levis:
### Pull Request Overview

This pull request is where we're doing the final polishing on the syscall TRD. This will be where we assign it a number, etc.

### Testing Strategy

N/A

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


jrvanwhy:
Is there a particular timeline you are looking for? I will be swamped the next few weeks.

phil-levis:
> Is there a particular timeline you are looking for? I will be swamped the next few weeks.

I think when we are ready for release. Assuming that all we are doing is better/more precisely stating how everything works, this can be the last thing we do.

hudson-ayers:
This branch had merges with tock-2.0-dev included in the commit history which made rebasing on top of the rebased version of tock-2.0-dev that was merged into master challenging.

So I squashed the changes on this PR and cherry-picked them as a single commit onto master. Hope that is ok

doc/reference/trd104-syscalls.md
```diff
@@ -235,57 +245,60 @@ The 6 classes are:
...
-is signaled by a callback (see the Subscribe call in 4.2).
+All of the system call classes except Yield and Exit are
+non-blocking. When a userspace process calls a Subscribe, Command,
+Allow, Read-Only Allow, or Memop syscall, the kernel will not put it
+on a wait queue. Instead, it will return immediately upon
+completion. The kernel scheduler may decide to suspend the process due
+to a timeslice expiration or the kernel thread being runnable, but the
+system calls themselves do not block. If an operation is long-running
+(e.g., I/O), its completion is signaled by an upcall (see the
+Subscribe call in 4.2).
```
bradjc, line 256: I would have written this that system calls themselves _do_ block. For example, every memop blocks the process from running until the entire memop operation has completed. Only then will the kernel reschedule the process.

Also, with the new scheduler trait, there is no guarantee that a process will not be put on a wait queue after calling a syscall.

doc/reference/trd104-syscalls.md
```diff
@@ -235,57 +245,60 @@ The 6 classes are:
...
-arguments are unsigned 32-bit integers. For example, the
-Console driver has driver identifier `0x1` and a Command to the console driver with
+System calls implemented by system call drivers (Subscribe, Command,
+Allow, Read-Only Allow) all include two arguments, a driver identifier
+and a syscall identifier. The driver identifier specifies which system
+call driver to invoke. The syscall identifier (which is different than
+the Syscall Class ID in the table above) specifies which instance of
+that system call on that driver to invoke. Both arguments are unsigned
+32-bit integers. For example, the Console system call driver has
+driver identifier `0x1` and a Command to the console driver with
```
bradjc, line 267:
```suggestion
driver identifier `0x1` by default and a Command to the console driver with
```

doc/reference/trd104-syscalls.md
```diff
@@ -423,22 +437,22 @@ start at address 0x0. However, even if they do begin at 0x0, the
...
 In the case that 0x0 is valid application code and where the
-linker places a callback function, the first instruction of the function
+linker places an upcall function, the first instruction of the function
 should be a no-op and the address of the second instruction passed
 instead.
 
 If a userspace process invokes subscribe on a driver ID that is not
 installed in the kernel, the kernel MUST return a failure with an
-error code of NOSUPPORT and a callback of the Null Callback.
+error code of NOSUPPORT and an upcall of the Null Upcall.
```
bradjc, line 446:
```suggestion
error code of `NODEVICE` and an upcall of the Null Upcall.
```

Per line 272? Although per the code it's `NOSUPPORT`.

doc/reference/trd104-syscalls.md
phil-levis: That's not traditionally what blocking means. Non-blocking I/O calls, for example, still return a return value. Usually blocking means that calling the function can cause the process to be descheduled and put on a wait queue, or forced to wait in some way.

doc/reference/trd104-syscalls.md
phil-levis: You can change the mappings? This seems extremely dubious. If you can change the mappings then we shouldn't have standard ones/a registry.

doc/reference/trd104-syscalls.md
phil-levis: Good catch.

doc/reference/trd104-syscalls.md
```diff
@@ -459,7 +473,9 @@ Command instance (combination of major and minor identifier) specifies
 its failure variant and success variant. If userspace invokes a
 command on a peripheral that is not installed, the kernel returns a
 failure variant of `Failure`, with an associated error code of
-`NOSUPPORT`.
+`NOSUPPORT`. Therefore, command invocations that need to
```
bradjc, line 476: There are a few references to `NOSUPPORT`, including the current implementation:

https://github.com/tock/tock/blob/38c12029d76530847591aded6afe14008f49668e/kernel/src/driver.rs#L186-L193

doc/reference/trd104-syscalls.md
```diff
@@ -649,61 +685,168 @@ information can be stored in the kernel and used in management or policy decisio
...
+```c
+void tock_exit(uint32_t completion_code);
+void tock_restart(uint32_t completion_code);
+```
+
+Since these two variants of Exit never return, they have
+no return value.
+
+
+5.2 libtock-rs
```
jrvanwhy, line 844: `libtock-rs` will have a more complex syscalls API than `libtock-c`, as there will be multiple variations of Subscribe, Read-Only Allow, and Read-Write Allow with different lifetime semantics. I also expect it to keep changing for a while.

Should we get rid of the `libtock-rs` heading?

doc/reference/trd104-syscalls.md
phil-levis: Yeah, I think there's a discrepancy between the code and the error values. The kernel should return `ErrorCode::NODEVICE` if there is no `Driver` implementation for the device ID, but a system call driver should return 'ErrorCode::NOSUPPORT` if it's an invalid system call ID.

At least, this is what the ErrorCode/ReturnCode values seem to say to me, unless we want to interpret a "device" as something else. Or we could just tweak this text and come up with a definition of NODEVICE that's distinct from NOSUPPORT and UNINSTALLED.

doc/reference/trd104-syscalls.md
phil-levis: Sure -- the goal here was to show examples. Having a follow-up TRD that specifies libtock-rs would be fine.

doc/reference/trd104-syscalls.md
alevy: I agree with @phil-levis  that "blocking" implies blocking on I/O (or some other long operation).

@bradjc is right that there is no type- or implementation- guarantee, but this _is_ the semantics of system calls, and it _should_ be implemented by any scheduler (unless of course that scheduler is exploring ways in which different semantics might be better). In other words, it is the semantic that a Tock kernel promises to processes.

doc/reference/trd104-syscalls.md
alevy: I think this is a persistent confusion that ought to be clarified, but probably not here. There are several notions of what "Tock" is, at least: a kernel crate that provides a particular execution model, interfaces for capsules, low-level system call semantics (the 2.0 system call interface defined in this document), etc; also an operating system implementation that uses that kernel and provides certain basic services (a console, timers, etc), with particular standardized system call interfaces.

In the latter case, the Console driver has driver identifier `0x1`. In former case, there may or may not be a console driver at all, it might be at a different system call number, and may have a totally different interface. Maybe it's used to store beanie baby futures.

I think this document is defining the latter---the low level system call interface that the kernel enforces, not the standard set of services provided via the system call interface to boards by convention.

Also, maybe this particular use of language doesn't really matter here, as it's just an example, and the example is correct. Would using "by convention" rather than "by default" be OK with everyone (which I think is technically more precise, though at the risk of being unnecessarily verbose)?

doc/reference/trd104-syscalls.md
```diff
@@ -235,57 +257,64 @@ The 6 classes are:
...
-is signaled by a callback (see the Subscribe call in 4.2).
+All of the system call classes except Yield and Exit are
+non-blocking. When a userspace process calls a Subscribe, Command,
+Allow, Read-Only Allow, or Memop syscall, the kernel will not put it
+on a wait queue. Instead, it will return immediately upon
+completion. The kernel scheduler may decide to suspend the process due
+to a timeslice expiration or the kernel thread being runnable, but the
+system calls themselves do not block. If an operation is long-running
+(e.g., I/O), its completion is signaled by an upcall (see the
+Subscribe call in 4.2).
```
bradjc, line 268:
```suggestion
All of the system call classes except Yield and Exit are
non-blocking. When a userspace process calls a Subscribe, Command,
Allow, Read-Only Allow, or Memop syscall, the kernel will not put the process
on a wait queue while handling the syscall. Instead, the kernel will complete the
syscall and prepare the return value for the syscall immediately.
The kernel scheduler may not, however, run the process immediately after
handling the syscall, and may instead decide to suspend the process due
to a timeslice expiration or the kernel thread being runnable.
If an operation is long-running
(e.g., I/O), its completion is signaled by an upcall (see the
Subscribe call in 4.2).
```

doc/reference/trd104-syscalls.md
```diff
@@ -235,57 +257,64 @@ The 6 classes are:
...
-arguments are unsigned 32-bit integers. For example, the
-Console driver has driver identifier `0x1` and a Command to the console driver with
+System calls implemented by system call drivers (Subscribe, Command,
+Allow, Read-Only Allow) all include two arguments, a driver identifier
+and a syscall identifier. The driver identifier specifies which system
+call driver to invoke. The syscall identifier (which is different than
+the Syscall Class ID in the table above) specifies which instance of
+that system call on that driver to invoke. Both arguments are unsigned
+32-bit integers. For example, the Console system call driver has
+driver identifier `0x1` and a Command to the console driver with
```
bradjc, line 279:
```suggestion
driver identifier `0x1`, by convention, and a Command to the console driver with
```

bradjc:
So

https://github.com/tock/tock/blob/c46e0cc5516ab9e4d55a2167e19ba7b70f4ce41e/kernel/src/sched.rs#L886

should be `NODEVICE`?

phil-levis:
> So
> 
> https://github.com/tock/tock/blob/c46e0cc5516ab9e4d55a2167e19ba7b70f4ce41e/kernel/src/sched.rs#L886
> 
> should be `NODEVICE`?

Yes, based on the description of the errors. Let's talk about it in the call today.

alistair23 (approved):

alistair23 (commented):
`NODEVICE` and `NOSUPPORT` changes as discussed in weekly call

alistair23 (dismissed):

phil-levis:
Alright, I think we are almost there and we're ready for last call. I'd like it if someone who isn't an author could do a proofreading/editorial pass. @alevy @ppannuto ?

bradjc (dismissed):

hudson-ayers (approved):

doc/reference/trd104-syscalls.md
```diff
@@ -164,19 +165,38 @@ value so the pointer can be passed back.
...
-well with Rust's `Result` type.
+operation might have multiple success return variants or failure return
+variants, then it MUST be split into multiple system calls.
+
+This requirement of a single failure variant and a single success
+variant is to simplify userspace implementations and preclude them
+from having to handle many different cases.  The presence of many
+difference cases suggests that the operation should be split up, as
+there is non-determinism in its execution or its meaning is
+overloaded. The requirement of a single failure anda single success
```
brghena, line 176:
```suggestion
overloaded. The requirement of a single failure and a single success
```

doc/reference/trd104-syscalls.md
```diff
@@ -193,35 +213,37 @@ are additional error codes to include errors related to userspace.
...
-the platform has and which have drivers installed in the kernel.
+Tock has 7 classes or types of system calls. When a system call is
+invoked, the class is encoded as the Syscall Class ID. Some system
+call classes are implemented by the core kernel and so the supported
+calls are the same across kernels. Others are implemented by system
+call drivers, which can be added and removed in different kernel
+builds. The full set of valid system calls a kernel supports therefore
+depends on what system call drivers it has installed.
 
 The 6 classes are:
```
brghena, line 248: Right above the table it says "The 6 classes are:" when there are 7 classes.

doc/reference/trd104-syscalls.md
```diff
@@ -235,57 +257,66 @@ The 6 classes are:
...
-All of the system call classes except Yield and Exit are non-blocking. When a userspace
-process calls a Subscribe, Command, Allow, Read-Only Allow, or Memop syscall,
-the kernel will not put it on a wait queue. Instead, it will return immediately
-upon completion. The kernel scheduler may decide to suspend the process due to
-a timeslice expiration or the kernel thread being runnable, but the system calls
-themselves do not block. If an operation is long-running (e.g., I/O), its completion
-is signaled by a callback (see the Subscribe call in 4.2).
+All of the system call classes except Yield and Exit are
+non-blocking. When a userspace process calls a Subscribe, Command,
+Allow, Read-Only Allow, or Memop syscall, the kernel will not put the process
```
brghena, line 262:
```suggestion
Read-Write Allow, Read-Only Allow, or Memop syscall, the kernel will not put the process
```

doc/reference/trd104-syscalls.md
```diff
@@ -235,57 +257,66 @@ The 6 classes are:
...
 
-Peripheral driver-specific system calls (Subscribe, Command, Allow, Read-Only Allow)
-all include two arguments, a driver identifier and a syscall identifier. The driver identifier
-specifies which peripheral system call driver to invoke. The syscall identifier (which
-is different than the Syscall Class ID in the table above)
-specifies which instance of that system call on that driver to invoke. Both
-arguments are unsigned 32-bit integers. For example, the
-Console driver has driver identifier `0x1` and a Command to the console driver with
+System calls implemented by system call drivers (Subscribe, Command,
+Allow, Read-Only Allow) all include two arguments, a driver identifier
```
brghena, line 275:
```suggestion
Read-Write Allow, Read-Only Allow) all include two arguments, a driver identifier
```

brghena (commented):
Some typos that should be fixed

doc/reference/trd104-syscalls.md
phil-levis: Fixed

hudson-ayers (approved):

PR #2433, opened 2021-02-15
use const generics to remove duplicate code for cortex-m MPU

hudson-ayers:
### Pull Request Overview

This pull request uses const generics to remove duplicate code for the Cortex-M MPU. It also moves mpu.rs into `arch/cortex-m`,
rather than having it in `arch/cortex-m3` and using symbolic links elsewhere. Each specific architecture re-exports the generic type as a concrete type with the correct number of regions for that architecture. I think this is more in line with how we generally handle shared code in other places. I realize that now `cortex-m/mpu.rs` would not work for all `cortex-m` platforms (e.g. `cortex-m0`), but architectures for which this does not work should just not re-export the functionality, and instead present their own MPU type. `cortex-m/src/mpu.rs` clearly indicates in the comments that it is only for use on m3/m4/m7.

We need generics because the cortex-m7 MPU has 16 regions instead of 8. The cortex-m7 MPU has already fallen out of date with the cortex-m3 MPU implementation because #2120 only updated the m3 implementation, and not the mostly duplicate code used in m7.

I used the `pub mod MPU {}` wrapper in each `lib.rs` to avoid having to modify every chip to update the path, but it could be removed.

### Testing Strategy

This pull request was tested by compiling and running blink on an Imix.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


bradjc (approved):

arch/cortex-m/src/mpu.rs
```diff
@@ -150,35 +150,31 @@ impl MPU {
...
-                CortexMRegion::empty(2),
-                CortexMRegion::empty(3),
-                CortexMRegion::empty(4),
-                CortexMRegion::empty(5),
-                CortexMRegion::empty(6),
-                CortexMRegion::empty(7),
-            ],
+impl<const NUM_REGIONS: usize> Default for CortexMConfig<NUM_REGIONS> {
+    fn default() -> Self {
+        // a bit of a hack to initialize array without unsafe
```
ppannuto, line 165: Is there any way to impl for a specific value? i.e. create impl's for `CortexMConfig<8>` and `CortexMConfig<16>`, which covers our use cases, and the would throw a 'missing impl' error for any other value of `NUM_REGIONS`? Could address this.

ppannuto (approved):
Looks good to me; the symlink thing was always a bit messy :/

alistair23 (approved):

arch/cortex-m/src/mpu.rs
hudson-ayers: Unfortunately I don't think we can get this to work using just `min_const_generics`. Implementing for just 8 and 16 leads to the following error on Imix (which uses `NUM_REGIONS = 8`):
```
error[E0277]: the trait bound `CortexMConfig<NUM_REGIONS>: Default` is not satisfied
   --> arch/cortex-m/src/mpu.rs:392:5
    |
392 |     type MpuConfig = CortexMConfig<NUM_REGIONS>;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Default` is not implemented for `CortexMConfig<NUM_REGIONS>`
    |
   ::: kernel/src/platform/mpu.rs:94:21
    |
94  |     type MpuConfig: Default + Display;
    |                     ------- required by this bound in `kernel::mpu::MPU::MpuConfig`
    |
    = help: the following implementations were found:
              <CortexMConfig<16_usize> as Default>
              <CortexMConfig<8_usize> as Default>
```

Adding a blanket implementation on top of those two that panics gives the following error:

```
error[E0119]: conflicting implementations of trait `core::default::Default` for type `mpu::CortexMConfig<8_usize>`:
   --> arch/cortex-m/src/mpu.rs:169:1
    |
163 | impl<const NUM_REGIONS: usize> Default for CortexMConfig<NUM_REGIONS> {
    | --------------------------------------------------------------------- first implementation here
...
169 | impl Default for CortexMConfig<8> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `mpu::CortexMConfig<8_usize>`
```

and I do not think that min_const_generics offers a way to `impl<const NUM_REGIONS: usize> ... where NUM_REGIONS != 8`

ppannuto: Ah, yeah, that makes sense. The compiler can't know at that point which versions it will eventually need. Perhaps someday as const generics evolves that can be optimized, but I don't think a few extra cycles on startup is going to hurt much for today.

bradjc (approved):

Issue #1993, opened 2020-07-01
IPC: Return error if shared memory region cannot be protected due to MPU alignment issues


bradjc:
Right now the IPC `allow()` call can succeed, and a notify callback can be triggered, even if the `expose_to()` call fails.

bradjc:
If we re-implement IPC to use message passing rather than shared memory, we can revisit the MPU implementation (re: #1532).

alexandruradovici:
I was playing with the IPC API and noticed that applications fault due to memory protection if the shared buffers are not multiple of 64. The buffer's memory region does not seem to be mapped.

I cannot figure out the reason for this, all the API calls seem to return success.

If this is a requirement, maybe we should add this to the kernel and refuse the buffer?

lschuermann:
While debugging tock/tock#2906 I've noticed a more intricate issue of app load order and, what appears to me as a racy relationship of the IPC service registering its callback and the client notifying the service. This is in the context of a failing `rot13_{service,client}` CI test. When debugging these issues I used to change the app's order in the TBF list and inserted `printf` statements. With the `rot13_service` and `rot13_client` installed in this order, what seems to be happening in the normal (good) case is:
1. `rot13_client` performs initialization and all of its IPC calls including `ipc_notify_service` in one go, which will enqueue a task in the target process' queue
2. `rot13_service` performs initialization and the call to `ipc_register_service_callback`, which will register its callback
3. the kernel processes the task enqueued in (1) which calls back to `IPC::schedule_upcall`, which will ultimately schedule the proper upcall on the `rot13_service` process
4. the IPC service callback scheduled in (3) is called by the kernel scheduler

Now, if the kernel were to execute (3) before (2), the IPC task is converted into a callback (`FunctionCall`) _before_ the service had a chance to subscribe. When service subscribes, this pending upcall is removed from its queue. It appears one can trigger this behavior through a combination of load-order and scheduler behavior, but it is most obvious when inserting a simple `printf` statement at the beginning of `ipc_service`'s main.

We currently have no way to tell whether notifying an IPC service (i.e. scheduling a callback in the service process) actually worked. Whereas the source code used to indicate that we might have a method of identifying a null-callback (`EOFF`), `Process::enqueue_task` does not actually provide us any information about whether the enqueued IPC task will result in a callback being scheduled or not. This usually makes sense and is a logical consequence of our architecture based around the fact that a null-callback should be treated exactly the same as a proper "do nothing" callback, but here it's not good.

[...]

PR #2434, opened 2021-02-16
Tock 2.0 IPC

alevy:
### Pull Request Overview
This pull request ports the IPC driver to the Tock 2.0 driver interface.

This is a mostly fairly straightforward port. However, the old driver API relied on `allow` to return a discovered IPC endpoint upon success, which is no longer possible in the 2.0 driver interface (allow cannot return a value).

Recall that IPC works as follows:

1. A client "discovers" an IPC service by supplying the kernel with the service's name (taken from the process's TBF name header)
2. If such a service exists, the kernel supplies the client process with a 32-bit non-zero descriptor to reference the service (implemented as the AppId + 1, but that's not guaranteed).
3. The client can use this descriptor to share memory with the service (using allow) and to `notify` it (using command).

For discovery, `allow(IPC, 0)` had been used to both share a buffer containing the service name _and_ request that the kernel perform discovery, returning a descriptor on success.

_This is no longer possible_, and it also not canonical driver semantics.

With this PR, discovery is split into an `allow_readonly` (which is nice anyway so
package names can be hardcoded in ROM rather than copied to
memory) followed by a `command(IPC, 1)` to actually perform the discovery and return a descriptor.

To make this cleaner, I also rejiggerd the arguments in `command`, which now has to mean more than _just_ notify.

Previously, the arguments were:

1. Descriptor (as opposed to the normal command number)
2. Service-or-client-notify (which is actually kind of like a command number)
3. Unused

Now, the arguments are:

1. Command number (can be "discover," "service notify," or "client notify")
2. Target descriptor
3. Unused

There are a number of other changes I believe _could_ make sense to make to IPC before releasing 2.0, but I think they should be decoupled from this pull request, which merely adapts the interface very little to get it working. These are just ideas, and we should discuss all these _separately_:

- Allow both read-only and read-write buffer sharing between client/service
- Use borrow semantics for memory ownership
- Let client subscribe to be notified when a service is available
- Probably more...

### Testing Strategy

Tested using the IPC tutorial in libtock-c (examples/tutorial/05-ipc) after applying changes from tock/libtock-c#176

### TODO or Help Wanted

~~Still need to adapt libtock-c userspace to properly test.~~ (Done in tock/libtock-c#176)

Still need to update documentation

Please review not only the implementation but the change in interface described above. As I said, there are a variety of changes that might make sense for IPC that go beyond these minimal changes. I ask that we get something in that reliably reproduces the existing high-level semantics for IPC and after (or maybe concurrently) decide whether other changes should happen (either before or after 2.0).

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/ipc.rs
+++ b/kernel/src/ipc.rs
@@ -5,12 +5,11 @@
 
 use crate::callback::{AppId, Callback};
 use crate::capabilities::MemoryAllocationCapability;
-use crate::driver::LegacyDriver;
 use crate::grant::Grant;
-use crate::mem::legacy::{AppSlice, SharedReadWrite};
+use crate::mem::Read;
 use crate::process;
-use crate::returncode::ReturnCode;
 use crate::sched::Kernel;
+use crate::{CommandResult, Driver, ErrorCode, ReadOnlyAppSlice, ReadWriteAppSlice};
 
 /// Syscall number
 pub const DRIVER_NUM: usize = 0x10000;
@@ -30,22 +29,23 @@ pub enum IPCCallbackType {
 struct IPCData<const NUM_PROCS: usize> {
     /// An array of app slices that this application has shared with other
     /// applications.
-    shared_memory: [Option<AppSlice<SharedReadWrite, u8>>; NUM_PROCS],
+    shared_memory: [ReadWriteAppSlice; NUM_PROCS],
+    search_slice: ReadOnlyAppSlice,
     /// An array of callbacks this process has registered to receive callbacks
     /// from other services.
-    client_callbacks: [Option<Callback>; NUM_PROCS],
+    client_callbacks: [Callback; NUM_PROCS],
     /// The callback setup by a service. Each process can only be one service.
-    callback: Option<Callback>,
+    callback: Callback,
 }
 
 impl<const NUM_PROCS: usize> Default for IPCData<NUM_PROCS> {
     fn default() -> IPCData<NUM_PROCS> {
-        // need this until const_in_array_repeat_expressions is stable
-        const NONE_APPSLICE: Option<AppSlice<SharedReadWrite, u8>> = None;
+        const DEFAULT_RW_APP_SLICE: ReadWriteAppSlice = ReadWriteAppSlice::const_default();
         IPCData {
-            shared_memory: [NONE_APPSLICE; NUM_PROCS],
-            client_callbacks: [None; NUM_PROCS],
-            callback: None,
+            shared_memory: [DEFAULT_RW_APP_SLICE; NUM_PROCS],
+            search_slice: ReadOnlyAppSlice::default(),
+            client_callbacks: [Callback::default(); NUM_PROCS],
+            callback: Callback::default(),
         }
     }
 }
@@ -67,66 +67,72 @@ impl<const NUM_PROCS: usize> IPC<NUM_PROCS> {
     /// scheduler loop if an IPC task was queued for the process.
     pub(crate) unsafe fn schedule_callback(
         &self,
-        appid: AppId,
-        otherapp: AppId,
+        schedule_on: AppId,
+        called_from: AppId,
         cb_type: IPCCallbackType,
-    ) {
+    ) -> Result<(), process::Error> {
         self.data
-            .enter(appid, |mydata, _| {
-                let callback = match cb_type {
+            .enter(schedule_on, |mydata, _| {
+                let mut callback = match cb_type {
                     IPCCallbackType::Service => mydata.callback,
-                    IPCCallbackType::Client => match otherapp.index() {
-                        Some(i) => *mydata.client_callbacks.get(i).unwrap_or(&None),
-                        None => None,
+                    IPCCallbackType::Client => match called_from.index() {
+                        Some(i) => *mydata
+                            .client_callbacks
+                            .get(i)
+                            .unwrap_or(&Callback::default()),
+                        None => Callback::default(),
                     },
                 };
-                callback.map_or((), |mut callback| {
-                    self.data
-                        .enter(otherapp, |otherdata, _| {
-                            // If the other app shared a buffer with us, make
-                            // sure we have access to that slice and then call
-                            // the callback. If no slice was shared then just
-                            // call the callback.
-                            match appid.index() {
-                                Some(i) => {
-                                    if i >= otherdata.shared_memory.len() {
-                                        return;
-                                    }
-
-                                    match otherdata.shared_memory[i] {
-                                        Some(ref slice) => {
-                                            slice.expose_to(appid);
-                                            callback.schedule(
-                                                otherapp.id() + 1,
-                                                slice.len(),
-                                                slice.ptr() as usize,
-                                            );
-                                        }
-                                        None => {
-                                            callback.schedule(otherapp.id() + 1, 0, 0);
-                                        }
-                                    }
-                                }
-                                None => {}
+                self.data.enter(called_from, |called_from_data, _| {
+                    // If the other app shared a buffer with us, make
+                    // sure we have access to that slice and then call
+                    // the callback. If no slice was shared then just
+                    // call the callback.
+                    match schedule_on.index() {
+                        Some(i) => {
+                            if i >= called_from_data.shared_memory.len() {
+                                return;
                             }
-                        })
-                        .unwrap_or(());
-                });
+
+                            match called_from_data.shared_memory.get(i) {
+                                Some(slice) => {
+                                    self.data
+                                        .kernel
+                                        .process_map_or(None, schedule_on, |process| {
+                                            process.add_mpu_region(
+                                                slice.ptr(),
+                                                slice.len(),
+                                                slice.len(),
+                                            )
+                                        });
+                                    callback.schedule(
+                                        called_from.id() + 1,
+                                        crate::mem::Read::len(slice),
+                                        crate::mem::Read::ptr(slice) as usize,
+                                    );
+                                }
+                                None => {
+                                    callback.schedule(called_from.id() + 1, 0, 0);
+                                }
+                            }
+                        }
+                        None => {}
+                    }
+                })
             })
-            .unwrap_or(());
+            .and_then(|x| x)
     }
 }
 
-// TODO: Write a Tock 2.0 driver implementation
-impl<const NUM_PROCS: usize> LegacyDriver for IPC<NUM_PROCS> {
+impl<const NUM_PROCS: usize> Driver for IPC<NUM_PROCS> {
     /// subscribe enables processes using IPC to register callbacks that fire
     /// when notify() is called.
     fn subscribe(
         &self,
         subscribe_num: usize,
-        callback: Option<Callback>,
+        mut callback: Callback,
         app_id: AppId,
-    ) -> ReturnCode {
+    ) -> Result<Callback, (Callback, ErrorCode)> {
         match subscribe_num {
             // subscribe(0)
             //
@@ -139,10 +145,10 @@ impl<const NUM_PROCS: usize> LegacyDriver for IPC<NUM_PROCS> {
             0 => self
                 .data
                 .enter(app_id, |data, _| {
-                    data.callback = callback;
-                    ReturnCode::SUCCESS
+                    core::mem::swap(&mut data.callback, &mut callback);
+                    callback
                 })
-                .unwrap_or(ReturnCode::EBUSY),
+                .map_err(|e| (callback, e.into())),
 
             // subscribe(>=1)
             //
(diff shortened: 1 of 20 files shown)


alevy:
Please do not merge yet, I have not tested properly and have not updated all the documentation. However, I believe it's more or less complete, so worth reviewing.

I've separated into two commits, one that changes the IPC driver and one (with a one-line change in a ton of files) that updates references to IPC in all boards using it. I suggest looking primarily at the first commit, which isn't actually that big.

hudson-ayers:
This increases code size by enough that the kernel no longer fits into flash on Imix (hence the CI failure). Looking at hail, it increased size by 1264 bytes.

alevy:
> 1264 bytes

I'll look into it, but that's _got_ to be an alignment thing (e.g., it maybe increased by enough of a margin to just push it over the alignment boundary or something), right?

kernel/src/ipc.rs
```diff
@@ -187,33 +206,104 @@ impl<const NUM_PROCS: usize> LegacyDriver for IPC<NUM_PROCS> {
...
         appid: AppId,
-    ) -> ReturnCode {
-        let cb_type = if client_or_svc == 0 {
-            IPCCallbackType::Service
-        } else {
-            IPCCallbackType::Client
-        };
+    ) -> CommandResult {
+        match command_number {
+            0 => {
```
hudson-ayers, line 215:
```suggestion
            0 =>
            /* Discover */
            {
```

kernel/src/ipc.rs
```diff
@@ -187,33 +206,104 @@ impl<const NUM_PROCS: usize> LegacyDriver for IPC<NUM_PROCS> {
...
+                            |slice| {
+                                self.data
+                                    .kernel
+                                    .process_until(|p| {
+                                        let s = p.get_process_name().as_bytes();
+                                        // are slices equal?
+                                        if s.len() == slice.len()
+                                            && s.iter().zip(slice.iter()).all(|(c1, c2)| c1 == c2)
+                                        {
+                                            Some(CommandResult::success_u32(p.appid().id() as u32))
```
hudson-ayers, line 229: Based on your description in this PR, and how this used to work when it was part of `allow()`, shouldn't this be `p.appid().id() + 1`?

kernel/src/ipc.rs
```diff
@@ -187,33 +206,104 @@ impl<const NUM_PROCS: usize> LegacyDriver for IPC<NUM_PROCS> {
...
+                            },
+                        )
                     })
-            })
+            }
+            _ => CommandResult::failure(ErrorCode::NOSUPPORT),
+        }
+    }
+
+    fn allow_readonly(
```
hudson-ayers, line 290:
```suggestion

    /// If allow_readonly is called with target_id == 0, it is an IPC service discover
    /// call. The contents of the slice should be the string name of the IPC
    /// service. The IPC driver will store this name for use by the service discover
    /// command.
    fn allow_readonly(
```

kernel/src/ipc.rs
```diff
@@ -233,54 +323,35 @@ impl<const NUM_PROCS: usize> LegacyDriver for IPC<NUM_PROCS> {
         &self,
         appid: AppId,
         target_id: usize,
```
hudson-ayers, line 325: I can't comment on the lines above this, but the doc comments are now out of date.

kernel/src/ipc.rs
```diff
@@ -233,54 +323,35 @@ impl<const NUM_PROCS: usize> LegacyDriver for IPC<NUM_PROCS> {
...
-        slice: Option<AppSlice<SharedReadWrite, u8>>,
-    ) -> ReturnCode {
+        mut slice: ReadWriteAppSlice,
+    ) -> Result<ReadWriteAppSlice, (ReadWriteAppSlice, ErrorCode)> {
         if target_id == 0 {
```
hudson-ayers, line 328: I am guessing this is the original reason that the target id was app_id + 1? Now that allow does not use target_id == 0, it seems we could go back to target_id = app_id. But maybe there is a benefit in keeping them separate in that it forces apps not to assume they are equal, which is something the kernel does not guarantee

hudson-ayers (commented):
Looks mostly good to me, I have a few comments.

hudson-ayers:
> > 1264 bytes
> 
> I'll look into it, but that's _got_ to be an alignment thing (e.g., it maybe increased by enough of a margin to just push it over the alignment boundary or something), right?

Using Phil's size tool, the code size required for functions in the `kernel::ipc::IPC::* ` namespace went from using 1006 bytes before this PR to using 2392 bytes afterwards. I think the size tool counts code space wasted due to alignment separately. Unfortunately the tool does not provide any resolution on where the size is going within individual methods on the struct. The one place in the changes where it is clear size might have changed is `command()` -- before, the code for service notify and client notify was almost entirely shared, and now it is duplicated across the two match arms. Still it makes no sense how that could over double the code size used..

kernel/src/ipc.rs
```diff
@@ -139,10 +153,10 @@ impl<const NUM_PROCS: usize> LegacyDriver for IPC<NUM_PROCS> {
             0 => self
```
phil-levis, line 153: 0 has to be the "exists" command; you should bump all the command IDs by 1:

https://github.com/tock/tock/blob/tock-2.0-dev/doc/reference/trd-syscalls.md#431-command-identifier-0

kernel/src/ipc.rs
```diff
@@ -187,33 +206,104 @@ impl<const NUM_PROCS: usize> LegacyDriver for IPC<NUM_PROCS> {
...
         _: usize,
         appid: AppId,
-    ) -> ReturnCode {
-        let cb_type = if client_or_svc == 0 {
-            IPCCallbackType::Service
-        } else {
-            IPCCallbackType::Client
-        };
+    ) -> CommandResult {
+        match command_number {
```
phil-levis, line 214: 0 has to be the "exists" command: you should bump all commands by 1:

https://github.com/tock/tock/blob/tock-2.0-dev/doc/reference/trd-syscalls.md#431-command-identifier-0

phil-levis:
>     * Use borrow semantics for memory ownership

This is interesting; fundamentally the idea of this shared memory goes against Rust's ownership model, so does need to be protected by some sort of lock. If we do decide to follow the borrow semantics, though, this becomes message passing (except if it's read-only on both sides)?

kernel/src/ipc.rs
alevy: Yes! Missed that thanks

hudson-ayers:
> This is interesting; fundamentally the idea of this shared memory goes against Rust's ownership model, so does need to be protected by some sort of lock. If we do decide to follow the borrow semantics, though, this becomes message passing (except if it's read-only on both sides)?

I suspect that the way we use shared memory now would be difficult to use without undefined behavior for a rust userspace? I realize libtock-rs does not currently support IPC.

I am not sure what exactly defines "message passing", but it seems that the ability to 

- take a lock
- modify 1 byte of the shared memory 
- release the lock

is lower overhead than having to pass the entire shared memory region back as a message. Maybe this is more of a semantic difference

kernel/src/ipc.rs
```diff
@@ -30,22 +29,33 @@ pub enum IPCCallbackType {
...
+                ptr_i.write(ReadWriteAppSlice::default());
+                ptr_i = ptr_i.add(1);
+            }
+
+            IPCData {
+                shared_memory: array_hack.assume_init(),
+                search_slice: ReadOnlyAppSlice::default(),
+                client_callbacks: [Callback::default(); NUM_PROCS],
+                callback: Callback::default(),
+            }
```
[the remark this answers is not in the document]
alevy: Unfortunately, the problem is not that it's not const, it's that `ReadWriteAppSlice` is not marked `Copy`, so without `const_in_array_repeat_expressions` (which is now not even an option with unstable) I don't think there are any tricks we can do.

If we're happy to make `ReadWriteAppSlice` `Copy`, we don't need unsafe, but I believe that could result in mutable aliases.

phil-levis:
> This increases code size by enough that the kernel no longer fits into flash on Imix (hence the CI failure). Looking at hail, it increased size by 1264 bytes.

I've been banging up on this a bit with imix. We should do a spring cleaning and see where we could reclaim some space. ADC for example is 6.1k!

kernel/src/ipc.rs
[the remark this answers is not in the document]
hudson-ayers: `const_in_array_repeat_expressions` was removed in favor of requiring explicit promotion of constants in constant blocks (see the discussion around [here](https://github.com/rust-lang/rust/issues/49147#issuecomment-736441091)). So as long as there is a const constructor, and you declare the const outside of the actual array instantiation, this works fine. This is the same reason the previous implementation works. Eventually this will get a little cleaner once [inline_const](https://github.com/rust-lang/rust/issues/76001) is stable -- then the `const` could be declared at the point of instantiation.

I tested the code changes above and they compile. Happy to push the commit if you prefer it, or drop it if you don't

alevy:
@phil-levis:

> I've been banging up on this a bit with imix. We should do a spring cleaning and see where we could reclaim some space. ADC for example is 6.1k!

I totally agree, but also I'm not sure how reliable those numbers are at that granularity (the tool is great! don't get me wrong). For example, removing the nrf51 capsule from Hail _adds_ ~1kB to the kernel, and the scheduler module occupies ~10kB on Hail, but only ~5kB on the Nano BLE, and ~8kB on the nrf52840DKs

The code changes really _shouldn't_ require adding much code space at all, so I'll debug this enough to figure out what's going on (though again, removing things sometimes _increases_ the size, so it's pretty brittle) and get imix inline.

But agree that we should do a deeper more comprehensive dive into this. These modules are all far bigger than they need to be.

phil-levis:
> the scheduler module occupies ~10kB on Hail, but only ~5kB on the Nano BLE, and ~8kB on the nrf52840DKs

I think this is inlining: modules higher in the call tree can take on code from lower ones. Happens less in the opposite direction.

kernel/src/ipc.rs
```diff
@@ -70,63 +80,67 @@ impl<const NUM_PROCS: usize> IPC<NUM_PROCS> {
...
-                                        }
-                                    }
+                            match otherdata.shared_memory.get(i) {
+                                Some(slice) => {
+                                    self.data.kernel.process_map_or(None, otherapp, |process| {
+                                        process.add_mpu_region(
+                                            slice.ptr(),
+                                            slice.len(),
+                                            slice.len(),
+                                        )
```
bradjc, line 114: We should probably look at https://github.com/tock/tock/issues/1993 since we are here. What happens if `add_mpu_region` fails?

kernel/src/ipc.rs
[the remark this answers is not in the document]
alevy: Aha! I didn't realize the subtlety about declaring the const ouside the array initialization so when I tried it it didn't work. This seems much much much better than what I have, will change.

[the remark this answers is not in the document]
alevy: Coincidentally also fixes the imix code size issue

kernel/src/ipc.rs
alevy: Right, I hadn't thought through all the implications so just left it as is for now. After porting libtock-c, I'll see if we can just elide this transformation. I don't think there is a significant benefit to obscuring this intentionally from apps if it doesn't serve another purpose (the semantics should still not promise anything about the meaning of the numbers though)

hudson-ayers (approved):

kernel/src/sched.rs
```diff
@@ -326,22 +326,22 @@ impl Kernel {
     /// `FAIL`. That is, if the closure returns any other return code than
     /// `FAIL`, that value will be returned from this function and the iteration
     /// of the array of processes will stop.
```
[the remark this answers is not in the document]
hudson-ayers: At some point in the force pushing this change seems to have been dropped

alevy:
Please still don't merge, I'm currently testing with libtock-c and still need to update documentation

bradjc:
I was playing around with IPC today and I found some issues. https://github.com/tock/tock/issues/1993 Might not be relevant for this PR but wanted to make a note of it.

alevy:
I pushed a commit (which I will squash before we merge) that fixes a bug I had introduced in `schedule_callback`. Because the bug was subtle and hard to catch, I also updated the names of the arguments. While after testing I'm fairly certain that function is now correct, it would be very useful for someone to quickly review my changes (just in that last commit should be fine) and verify that what's going on in that function kind of actually makes sense.

alevy:
Testing complete. Just going to update documentation, then I think we should be good (modulo squashing the last commit with the first one)

alevy:
Documentation updated

hudson-ayers (approved):
The most recent commit looks correct to me, it seems to match the behavior from when `expose_to()` was used, and the updated names are helpful

alevy:
Squashed

kernel/src/mem.rs
```diff
@@ -131,6 +131,14 @@ impl ReadWriteAppSlice {
     pub(crate) fn consume(self) -> (*mut u8, usize) {
         (self.ptr, self.len)
     }
+
+    pub(crate) const fn const_default() -> Self {
```
bradjc, line 135: Can you capture some wisdom on what this is?

kernel/src/sched.rs
```diff
@@ -683,7 +682,13 @@ impl Kernel {
                                         );
                                     },
                                     |ipc| {
-                                        ipc.schedule_callback(process.appid(), otherapp, ipc_type);
+                                        // TODO(alevy): this could error for a variety of reasons.
+                                        // Should we communicate the error somehow?
```
bradjc, line 686: Are we adding this to #1993?

alevy: Yes

bradjc (approved):

phil-levis:
Hurray!

PR #2436, opened 2021-02-17
Defining high and low BLE HILs

alexandruradovici:
### Pull Request Overview

Following the discussion for #2233, I tried to define a BLE lower and upper HILs. This is far from being ready, I opened this PR to get feedback and see if I am on the right track. The premises were the following:

BLE Lower Interface
 - sends and receives packets using the hardware device

BLE Higher interface
 - uses the GAP and GATT structure

### Testing Strategy

N/A

### TODO or Help Wanted

I would love to get feedback about this to make sure this is getting in the right direction.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/kernel/src/hil/ble.rs
+++ b/kernel/src/hil/ble.rs
@@ -1,14 +1,14 @@
 //! BLE Upper and Lower level HILs
 //!
 //!
-//!       BLE Capsule
+//!       BLE Syscall Capsule
 //!
 //! +-------------------+  +--------------------+
 //! |   BleGattServer   |  |   BLEGattClient    |
 //! +-------------------+  +--------------------+
 //!
 //!
-//!        BLE Stack (eq: Rubble)
+//!        BLE Stack (eg: Rubble)
 //!
 //!
 //! +-------------------------------------------+
@@ -17,6 +17,52 @@
 //!
 
 use crate::ReturnCode;
+
+// *************************************************************************
+//   High Level HIL
+// *************************************************************************
+
+pub enum AttributeType {
+    Uuid16 (u16),
+    Uuid128 (u128)
+}
+
+pub const PRIMARY_SERVICE:AttributeType = AttributeType::Uuid16 (0x2800);
+pub const SECINDARY_SERVICE:AttributeType = AttributeType::Uuid16 (0x2801);
+pub const INCLUDE:AttributeType = AttributeType::Uuid16 (0x2802);
+pub const CHARACTERISTIC:AttributeType = AttributeType::Uuid16 (0x2802);
+
+/// Attribute format types from https://www.bluetooth.com/specifications/assigned-numbers/format-types/
+#[repr(u16)]
+pub enum AttributeFormatType {
+    Boolean = 0x01,
+    Bit2 = 0x02,
+    Nibble = 0x03,
+    Uint8 = 0x04,
+    Uint12 = 0x05,
+    Uint16 = 0x06,
+    Uint24 = 0x07,
+    Uint32 = 0x08,
+    Uint48 = 0x09,
+    Uint64 = 0x0a,
+    Uint128 = 0x0b,
+    Sint8 = 0x0c,
+    Sint12 = 0x0d,
+    Sint16 = 0x0e,
+    Sint24 = 0x0f,
+    Sint32 = 0x10,
+    Sint48 = 0x11,
+    Sint64 = 0x12,
+    Sint128 = 0x13,
+    Float32 = 0x14,
+    Float64 = 0x15,
+    SFloat = 0x16,
+    Float = 0x17,
+    DUint16 = 0x18,
+    UTF8s = 0x19,
+    UTF16S = 0x1a,
+    Struct = 0x1b
+}
 
 pub enum BleGattAccessType {
     None,
@@ -29,14 +75,17 @@ pub enum BleGattAccessType {
     ReadWriteNotify
 }
 
-pub enum BleGattService<'a> {
-    Bit16(u16, &'a [BleGattCharacteristic]),
-    Bit128(u128, &'a [BleGattCharacteristic]),
+pub trait BleGattService<'a> {
+    type Characteristic: BleGattCharacteristic;
+
+    fn get_uuid (&self) -> AttributeType;
+    fn get_characteristics(&self) -> &'a [Self::Characteristic];
 }
 
-pub enum BleGattCharacteristic {
-    Bit16(u16, BleGattAccessType),
-    Bit128(u128, BleGattAccessType),
+pub trait BleGattCharacteristic {
+    fn get_uuid () -> AttributeType;
+    fn get_format_type () -> AttributeFormatType;
+    fn get_access () -> BleGattAccessType;
 }
 
 pub enum BleGattError {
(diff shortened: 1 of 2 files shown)


kernel/src/hil/ble.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// 
+    /// The packet data an opaque buffer
+    fn receive_packet(&self, buf: &'a mut [u8], len: usize, channel: RadioChannel);
+
+    fn set_receive_client(&self, client: &'a dyn BleRxClient);
+    fn set_transmit_client(&self, client: &'a dyn BleTxClient);
+}
+
+pub trait BleRxClient<'a> {
+    fn receive_event(&self, buf: &'a mut [u8], len: u8, result: ReturnCode);
```
bradjc, line 152:
```suggestion
    fn receive_complete(&self, buf: &'a mut [u8], len: u8, result: ReturnCode);
```
?

kernel/src/hil/ble.rs
```diff
@@ -0,0 +1,247 @@
...
+}
+
+pub trait BleTxClient<'a> {
+    fn transmit_event(&self, buf: &'a mut [u8], result: ReturnCode);
```
bradjc, line 156:
```suggestion
    fn transmit_complete(&self, buf: &'a mut [u8], result: ReturnCode);
```
?

kernel/src/hil/ble.rs
```diff
@@ -0,0 +1,247 @@
...
+        name: &'a [u8],
+        services: &'a [Self::Service],
+        connect_interval_ms: u32,
+    ) -> ReturnCode;
+
+    // Disconnect GATT Central
+    fn disconnect(&self) -> ReturnCode;
+
+    // Start GAP Advertisement
+    fn start_advertisement(&self) -> ReturnCode;
```
bradjc, line 115: How does the advertisement contents get set?

alexandruradovici: Using the `configure` function

bradjc: But there are more advertising parameters. We don't need to support them all initially necessarily, but having a function for future expansion would be helpful.

alexandruradovici: I haven't seen anything else in the GAP advertising, but I'm not an expert in BLE. What parameters are you referring to?

l162:
What do you all think about having more specific abstractions for the lower level network interface (i.e. link layer)? 

In the BLE example, this might mean having separate ble hil and link hil, where ble is defined as in the high level interface in this PR and link defines traits like those in Rubble link module. 

Still fairly new here so apologies if missing something obvious or this is discussed elsewhere.

alexandruradovici:
> This may not be the right place to ask this, but, what do you all think about having more specific abstractions for the lower level network interface (i.e. link layer)?
> 
> In the BLE example, this might mean having separate ble hil and link hil, where ble is defined as in the high level interface in this PR and link defines traits like those in Rubble link module.
> 
I think this is the purpose of this PR. Can you suggest some changes in the code? I am not sure I understand what you mean.
> Still fairly new here so apologies if missing something obvious or this is discussed elsewhere.

kernel/src/hil/ble.rs
bradjc: For example having manufacturer specific data in the advertisement, whether the device is connectable or not, long or short service UUIDs. There are a lot of advertisement options.

l162:
>  Can you suggest some changes in the code?

Happy to give this a short, but thought it might be better to discuss first since link layer exists across a variety of network protocols and I am not familiar enough with this project to understand well whether making that sort of design decision fits goals here (i.e. it maybe better to not add additional layer of complexity and just continue to have `radio`, `ethernet` hils). 

The basic idea though is to have traits like `Transmitter`, `Receiver`, and `Config`. Also possible to extend to general purpose network infrastructure like `PDU`, `DeviceAddress`, etc. 

For Rubble analogy, this means L2CAP is implemented in capsule as glue between two HALs.

hudson-ayers:
> Happy to give this a shot, but thought it might be better to discuss first since link layer exists across a variety of network protocols and I am not familiar enough with this project to understand well whether making that sort of design decision fits goals here (i.e. it maybe better to not add additional layer of complexity and just continue to have `radio`, `ethernet` hils).
> 
> The basic idea though is to have traits like [`Transmitter`](https://github.com/jonas-schievink/rubble/blob/d965126dc5d5ed1ae665cda3f1db25c1fa7e9659/rubble/src/link/mod.rs#L539), `Receiver`, and [`Config`](https://github.com/jonas-schievink/rubble/blob/d965126dc5d5ed1ae665cda3f1db25c1fa7e9659/rubble/src/config.rs#L15). Also possible to extend to general purpose network infrastructure like `PDU`, `DeviceAddress`, etc.
> 
> For Rubble analogy, this means L2CAP is implemented in capsule as glue between two hil impls.

I suspect that creating a generic link layer abstraction that works across BLE / ethernet / 802.15.4 will be quite difficult. I assume the idea here is some interface that is usable by an IP layer above, but the way that IP interacts with the lower layers in BLE is pretty different from how it interacts with the lower layers in 802.15.4 (e.g. via 6LoWPAN). Also, as of today we only have a minimal BLE stack, no ethernet stack, and an incomplete IP implementation -- so it would be difficult to even understand the requirements of all 3 when creating such an interface. Do you know of any other OS with a generic link layer abstraction of this sort?

l162:
> creating a generic link layer abstraction that works across BLE / ethernet / 802.15.4 will be quite difficult

It seems like you are describing a type class for networked systems
```rust

trait LinkLayer {
    type Frame: Frame;
    type Delivery: Config;
    type Link: Link;
}
```
This is one interpretation of a link hil, but, is likely the most restrictive. By contrast, a minimal interface might consist purely of marker types. 

A hybrid might be advancing ad-hoc. For this PR, that would mean splitting high level and low level into separate hils; but, in other instances, it may be desirable to add more to the link hil, which would necessarily mean retrofitting BLE. 

```Rust

trait Link {
    fn set_transmit_client(&self, client: &'a dyn TxClient); 	
    fn set_receive_client(&self, client: &'a dyn RxClient);
}

trait RxClient {
    fn receive_complete(&self, buf: …);
}

trait TxClient {
    fn transmit_complete(&self, buf: …);
}
```

The thinking is that the higher up-front cost will make it easier to reason about networking generally. As an example, having separate hils should set up a one-to-one correspondence with capabilities. Wanting to have fine-grained network capabilities without separate high-level/low-level hils means either requiring both high level or low level networking capabilities (which may be undesirable), or having protocol-specific capability in favor of general networking capability (which may be desirable).

Even stepping back, if wanting to evaluate UDP capability model (perhaps to inform above), might look to existing BLE stack and see whether using something like `BLEDriverCapability` would work with existing hil. Since there are no BLE capabilities, though, probably look to UDP which does have them; but, since UDP does not have hil, not really clear how this might translate to BLE hil? 

So an ad-hoc link hil may be a good place to specify the requirements for general network model, but is likely not be a good way to model specific network behavior.

skannan89:
@alexandruradovici Any updates on this PR? Would love to see BLE support.

phil-levis:
> I suspect that creating a generic link layer abstraction that works across BLE / ethernet / 802.15.4 will be quite difficult.

Yes. Probably not worth it. Doing this even for different MAC layers of a given link layer is a research topic (Polastre et al., https://dl.acm.org/doi/10.1145/1098918.1098928 )

phil-levis:
Should we close this, or is there an intention to continue work on it?

PR #2439, opened 2021-02-17
Tock 2.0: Remove userspace IEEE 802.15.4 driver from Imix platform

hudson-ayers:
### Pull Request Overview

This pull request removes the userspace 15.4 driver from Imix. The primary motivation is to reduce code size, as tock 2.0 development has been hung up several times by code size issues on Imix. Most recently, my attempt to merge master into tock-2.0-dev failed because the kernel could not fit in flash.

I chose to remove the userspace 15.4 driver because we now have the higher level UDP interface, which should be used instead of this one anyway. In fact, the presence of the 15.4 driver breaks many of the security guarantees of the UDP stack because a second app could just listen-in directly on the contents of 15.4 packets.

 Eventually the component for 15.4 should be factored out into two components, one for creating the 15.4 mux and one for creating the userspace driver, but I prefer to leave that task for the future, as dead code elimination gets us far enough here.

### Testing Strategy

This pull request was tested by compiling. It reduced code size by about 6kB.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/imix/src/main.rs
+++ b/boards/imix/src/main.rs
@@ -129,7 +129,6 @@ struct Imix {
     >,
     ipc: kernel::ipc::IPC<NUM_PROCS>,
     ninedof: &'static capsules::ninedof::NineDof<'static>,
-    radio_driver: &'static capsules::ieee802154::RadioDriver<'static>,
     udp_driver: &'static capsules::net::udp::UDPDriver<'static>,
     crc: &'static capsules::crc::Crc<'static, sam4l::crccu::Crccu<'static>>,
     usb_driver: &'static capsules::usb::usb_user::UsbSyscallDriver<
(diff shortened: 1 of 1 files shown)


bradjc (approved):

bradjc:
Pragmatic fix for now we that can re-evaluate in the future if need be, but it seems like UDP is the more desirable interface anyway.

bradjc:
Wait why not merge this into master?

phil-levis:
As we remove capsules from the main imix board, I'd like to propose we have a second board that includes just the elided capsules. Otherwise testing them becomes extremely onerous. We can call this board imix_capsule_testing or something.

phil-levis:
I agree with Brad, this should be on master. We are going to merge master into tock-2.0-dev soon.

hudson-ayers:
I updated this to be a change on master.

I have a completed merge of master into tock-2.0-dev that is only blocked on this.

hudson-ayers:
> As we remove capsules from the main imix board, I'd like to propose we have a second board that includes just the elided capsules. Otherwise testing them becomes extremely onerous. We can call this board imix_capsule_testing or something.

One thing to point out in this specific case is that the nrf52840dk still includes the raw 15.4 driver

ppannuto (approved):
bors r+

Agreed, we need a more comprehensive understanding of how capsule code is tested generally I think. Right now it's a somewhat ad-hoc reliance on 'at least one board is using each capsule' I think. Perhaps something that can be added to Brad's proposed hack day on testing infrastructure.

hudson-ayers:
bors passed but could not do the merge because netlify-preview did not run because I changed the base branch as the last action on this PR, so merging manually

phil-levis:
Oooof. This broke the imix test application.

hudson-ayers:
I think the breakage should be fixed by https://github.com/tock/libtock-c/pull/190 ? Master just needs to be merged into the 2.0 branch probably.

PR #2440, opened 2021-02-17
RFC: Add `install` make target for boards

bradjc:
Tock has had a long standing convention of two makefile targets for boards to load a kernel onto a board.

1. flash: Uses JTAG to flash the kernel.
1. program: Uses UART and a bootloader to flash the kernel.

This is helpful because some boards support both but require different hardware, and it gives the user some idea of what tools will be needed to load the kernel.

However, new users must figure out which is appropriate if a board supports both. Often there is one target that experienced users just know to use, and we should make this more clear to new users. Also, having different targets on different boards complicates writing a common tutorial guide for multiple platforms.

This patch adds a new target `make install` that just uses the default for the platform. The expectation is that most users can install the kernel using `make install`. The other targets are of course there for users who need something more specific.

### Testing Strategy

This pull request was tested by using `make install` to flash the kernel on hail.

### TODO or Help Wanted

Thoughts?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/weact_f401ccu6/Makefile
+++ b/boards/weact_f401ccu6/Makefile
@@ -9,6 +9,10 @@ include ../Makefile.common
 OPENOCD=openocd
 OPENOCD_OPTIONS=-f openocd.cfg
 
+# Default target for installing the kernel.
+.PHONY: flash
+install: program
+
 flash: $(KERNEL)
 	$(OPENOCD) $(OPENOCD_OPTIONS) -c "init; reset halt; flash write_image erase $<; verify_image $<; reset; shutdown"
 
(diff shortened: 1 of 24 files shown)


boards/weact_f401ccu6/Makefile
```diff
@@ -9,6 +9,10 @@ include ../Makefile.common
 OPENOCD=openocd
 OPENOCD_OPTIONS=-f openocd.cfg
 
+# Default target for installing the kernel.
+.PHONY: flash
```
hudson-ayers, line 13:
```suggestion
.PHONY: install
```

boards/weact_f401ccu6/Makefile
```diff
@@ -9,6 +9,10 @@ include ../Makefile.common
...
+install: program
```
hudson-ayers, line 14:
```suggestion
install: flash
```

hudson-ayers (commented):
I like this change

alistair23 (approved):

hudson-ayers (approved):

ppannuto (approved):

alistair23 (approved):

PR #2441, opened 2021-02-18
Change process setup to minimize the initially-accessible amount of memory.

jrvanwhy:
### Pull Request Overview

This was discussed at https://github.com/tock/tock/issues/1769. This change removes the Tock 1.0 behavior of always giving processes at least 3 kB of usable memory, and instead provides processes the minimum-possible amount of accessible memory to start. This simplifies the C and Rust runtimes, as the runtime initialization assembly can now unconditionally adjust the process break before setting the stack pointer.

The Tock 2.0 rewrite of `libtock-rs` already uses the new semantics on RISC-V, and will use these new semantics on ARM when I merge my work-in-progress code upstream.

### Testing Strategy

Running the tock-on-titan `flash_test` app on H1 (a Cortex-M3 chip) using these changes.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/syscall.rs
+++ b/arch/cortex-m/src/syscall.rs
@@ -61,17 +61,9 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
     type StoredState = CortexMStoredState;
 
     fn initial_process_app_brk_size(&self) -> usize {
-        // TOCK 1.X
-        //
-        // The 1.x Tock kernel allocates at least 3 kB to processes, and we need
-        // to ensure that happens as userspace may expect it.
-        3 * 1024
-
-        // TOCK 2.0
-        //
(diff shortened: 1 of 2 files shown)


jrvanwhy:
Huge thank you to Brad for #2383 !

Also, I was unable to test this change on RISC-V. The only RISC-V test environment I have available is QEMU's simulated Hifive1, and the latest `tock-2.0-dev` hifive1 kernel is not booting (with or without these changes). I don't have the time to debug it, unfortunately.

hudson-ayers (approved):

bradjc (approved):

lschuermann:
@jrvanwhy Thanks for doing this! I didn't do it as part of #2407 as I didn't have a chance to test the new behavior at that moment. Great that this is integrated now.

PR #2442, opened 2021-02-18
board: Raspberry Pi Pico Tock port

ioanaculic:
### Pull Request Overview

This pull request adds support for the RP2040 chip and the Raspberry Pi Pico board. This is a work in progress, I opened the PR to track the changes.

 - [X] Cortex-M0+
 - [X] GPIO
 - [X] External Interrupts
 - [X] UART
 - [X] Timer
 - [X] ADC
 - [X] Temperature Sensor

This PR adds the ability to *software* mask NVIC pending interrupts. This is required as the RP2040 chip has two cores, and one of the pending interrupts (SIO_IRQ_PROC1) cannot be cleared from the first core. The second core has to be started so that it can clear that pending interrupt.

Apps can be uploaded only into RAM due to an openocd issue (https://github.com/raspberrypi/openocd/issues/25) that the Raspberry Pi Foundation is working on.

### Testing Strategy

This was tested using a Raspberry Pi Pico connected to a Raspberry Pi 2B, 3B and 4B.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers:
This is exciting, having support for the Pico in Tock would be really cool!

> I saw that the cortex-m0 crate is deprecated, but this MCU uses the cortex-m0+, any feedback on how to organize this would be appreciated.

The m0 code is deprecated because we no longer have boards in upstream Tock using an m0 chip -- a major reason for this is that m0 boards we used previously did not have an MPU. I suspect that if this board makes it into upstream Tock we would be happy to un-deprecate `cortex-m0`. The fact that the RP2040 has an MPU is a big help toward this.

Regarding structuring your code, I think it would be better to have a cortex-m0p crate depend on the existing cortex-m0 crate, rather than duplicate all that code. An example of how that might look can be found in this PR for adding a cortex-m4f crate: https://github.com/tock/tock/pull/2348/files . Basically the m0 crate should just contain code that applies to all m0 chips, and m0p crate should contain any code (e.g. MPU code) that only applies to m0+ chips, and otherwise re-export stuff from m0. Happy to explain further if this is still confusing.

Other thoughts:
- the way we include MPU code across cortex-m chips just changed a few hours ago, you may want to check out this PR: https://github.com/tock/tock/pull/2433 to see the new approach

- Tock does not have multi-core support, and attempting to run the kernel on multiple cores is likely to break many of Tock's safety/soundness assumptions. So I suspect that for the initial commit of this code to Tock, we will want the Pico to be configured to run code on one core only.

alexandruradovici:
We updated the m0 support, but this needs some review for the asm part. 

One problem that we face is that we move out the `switch_to_user_arm_v7m` function from the `cortex-m` crate as the compiler complained about illegal thumb instructions `ldmia` and `stmia` even tough this function is not used. We copied the function to the `cortex-m3` and `cortex-m4` crates. Should `cortex-m4` import it from `cortex-m3`?

arch/cortex-m/src/lib.rs
```diff
@@ -210,52 +210,9 @@ pub unsafe extern "C" fn unhandled_interrupt() {
     panic!("Unhandled Interrupt. ISR {} is active.", interrupt_number);
 }
 
-/// Assembly function called from `UserspaceKernelBoundary` to switch to an
-/// an application. This handles storing and restoring application state before
-/// and after the switch.
-#[cfg(all(target_arch = "arm", target_os = "none"))]
```
ppannuto, line 216: From `rustc --print target-list`

Can this `cfg` maybe be an `any` of 

```
thumbv7a-pc-windows-msvc
thumbv7a-uwp-windows-msvc
thumbv7em-none-eabi
thumbv7em-none-eabihf
thumbv7m-none-eabi
thumbv7neon-linux-androideabi
thumbv7neon-unknown-linux-gnueabihf
thumbv7neon-unknown-linux-musleabihf
thumbv8m.base-none-eabi
thumbv8m.main-none-eabi
thumbv8m.main-none-eabihf
```

(or possibly not even those last three)?

alexandruradovici: We tried `#[cfg(all(target_arch = "thumbv7em", target_os = "none"))]` and it ignores it when we compile for the m4 (TARGET=thumbv7em-none-eabi)

jrvanwhy: `target_arch` is `arm` here, not `thumbv7em` (it is different from the first part of the target triple).

I haven't looked at this PR to see what you are trying to achieve, but I don't think you can access a target triple from `cfg`. You can access it from `build.rs` (as the `CARGO_BUILD_TARGET` environment flag) if you really need to know.

alexandruradovici: I don't think using build.rs is useful, we need to make the compiler to ignore a function that contains armv7 code (thumb2).

hudson-ayers: yeah, I think if we wanted to use `cfg` we would need to create a custom attribute which is set differently for thumbv6 boards than for thumbv7. I imagine 4 possible solutions here:

1. custom cfg options for thumbv7 vs thumbv6, with the correct cfg set in build.rs for each cortex-m board.
2. Avoid the problem for now by replacing `ldmia` and `stdmia` with multiple instructions that accomplish the same thing (I think this is possible but didn't look deeply into it)
3. move the function to cortex-m3 and import from m4 (I do not like this, as the m4 crate should not depend on the m3 crate)
4. Create `thumbv6m` and `thumbv7m` crates and integrate them into the existing hierarchy somehow

I probably prefer (2) for now but am curious what others think

bradjc: It seems like we need to fix rust's `target_arch` cfg for them, as "arm" is insufficient (arg!!). Could we define `target_arch_tock` in our build system to be more specific?

bradjc: That is, I prefer (1) (since that fixes the underlying problem that rust has created, and might be useful in the future since we should be able to cfg on an _actual_ architecture, not just a family of architectures. Perhaps rust will fix this in the future.).

I guess we could do (4), but I find "cortex-mX" more user friendly than "thumbvXX" since humans are exposed to marketing names, for better or worse. I think we could balance this with yet-more-crates, but that fragments code even more, and makes porting fixes to all relevant platforms more challenging.

bradjc: It looks like rust does give us a way to do this already: https://github.com/tock/tock/pull/2452

bradjc: We are now more pedantic about exactly which arm architecture, so this problem should be avoided now that #2452 has been merged.

.vscode/settings.json
```diff
@@ -1,4 +1,4 @@
 {
     "editor.formatOnSave": true,
     "rust-client.channel": "nightly-2021-03-19",
-}
+}
```
alistair23, line 4: Don't change this file

arch/cortex-m0/src/lib.rs
```diff
@@ -282,18 +282,41 @@ unsafe fn kernel_hardfault(faulting_stack: *mut u32) {
...
-         b 1b
-         "
-    );
+    //     asm!(
+    //         "
+    //          bkpt
+    // 1:
+    //          b 1b
+    //          "
+    //     );
```
alistair23, line 319: If this doesn't work it should just be removed.

boards/raspberry_pi_pico/Makefile
```diff
@@ -0,0 +1,28 @@
...
+
+include ../Makefile.common
+
+OPENOCD=openocd
+OPENOCD_OPTIONS=-f openocd.cfg
+
+APP=../../../libtock/libtock-c/examples/tests/gpio/build/cortex-m0/cortex-m0.tbf
+KERNEL=$(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
+KERNEL_WITH_APP=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/debug/$(PLATFORM)-app.elf
+
```
alistair23, line 15: This is missing an `install` target

arch/cortex-m0/src/lib.rs
alexandruradovici: I added the `panic!` call here, but I am not sure what the original author meant by wrong PanicInfo. This is why I did not delete this yet.

.vscode/settings.json
ioanaculic: My editor deleted the last empty line. I put back the original file.

boards/raspberry_pi_pico/Makefile
ioanaculic: Added

alistair23 (changes requested):
Overall looks good!

arch/cortex-m/src/nvic.rs
```diff
@@ -13,6 +13,24 @@
 use kernel::common::registers::{register_bitfields, register_structs, ReadOnly, ReadWrite};
 use kernel::common::StaticRef;
 
+#[macro_export]
+macro_rules! interrupt_mask {
+    ($($interrupt: expr),+) => {{
+        let mut high_interrupt: u128 = 0;
+        let mut low_interrupt: u128 = 0;
```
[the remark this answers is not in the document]
ppannuto: I'm a sucker for a question like this, and diving a little down the rabbit hole, it looks like no:

```diff
+#[no_mangle]
pub unsafe fn has_pending_with_mask(mask: (u128, u128)) -> bool {
    NVIC.ispr
        .iter()
        .take(number_of_nvic_registers())
        .enumerate()
        .fold(0, |i, (block, ispr)| {
            let interrupt_mask = if block < 4 { mask.1 } else { mask.0 };
            (ispr.get() & !((interrupt_mask >> (32 * block % 4)) as u32)) | i
        })
        != 0
}
```

```
10005a50 <has_pending_with_mask>:
10005a50: f8 b5        	push	{r3, r4, r5, r6, r7, lr}
10005a52: 04 af        	add	r7, sp, #16
10005a54: 00 90        	str	r0, [sp]
10005a56: 01 21        	movs	r1, #1
10005a58: 49 02        	lsls	r1, r1, #9
10005a5a: 0c 4a        	ldr	r2, [pc, #48]
10005a5c: 13 68        	ldr	r3, [r2]
;         self as *const [T] as *const T
10005a5e: 5a 18        	adds	r2, r3, r1
;     unsafe { intrinsics::volatile_load(src) }
10005a60: 59 68        	ldr	r1, [r3, #4]
10005a62: 0f 23        	movs	r3, #15
;             fn bitand(self, rhs: $t) -> $t { self & rhs }
10005a64: 0b 40        	ands	r3, r1
10005a66: 58 1c        	adds	r0, r3, #1
10005a68: bc 68        	ldr	r4, [r7, #8]
10005a6a: 00 25        	movs	r5, #0
10005a6c: 29 46        	mov	r1, r5
10005a6e: 04 2d        	cmp	r5, #4
10005a70: 26 46        	mov	r6, r4
10005a72: 00 d3        	blo	#0 <has_pending_with_mask+0x26>
10005a74: 00 9e        	ldr	r6, [sp]
;     unsafe { intrinsics::volatile_load(src) }
10005a76: 08 ca        	ldm	r2!, {r3}
10005a78: b3 43        	bics	r3, r6
10005a7a: 19 43        	orrs	r1, r3
;                 count += 1;
10005a7c: 6d 1c        	adds	r5, r5, #1
;             accum = f(accum, x)?;
10005a7e: a8 42        	cmp	r0, r5
10005a80: f5 d1        	bne	#-22 <has_pending_with_mask+0x1e>
10005a82: 48 1e        	subs	r0, r1, #1
10005a84: 81 41        	sbcs	r1, r0
10005a86: 08 46        	mov	r0, r1
10005a88: f8 bd        	pop	{r3, r4, r5, r6, r7, pc}
10005a8a: c0 46        	mov	r8, r8


10005a8c <$d.230>:
10005a8c: 5c de        	udf	#92
10005a8e: 00 10        	asrs	r0, r0, #32

10005a90 <cortexm::nvic::Nvic::enable::hf2b12544582bd8b5>:
...
```

For this kind of bitmask manipulation, the compiler can figure out the right thing to do here, which makes sense ot me.

boards/kernel_layout.ld
```diff
@@ -229,11 +229,16 @@ SECTIONS
         . = ALIGN(4);
         _sapps = .;
 
-        /* Include a placeholder byte in this section so that the linker
+        /* Include a placeholder byte(s) in this section so that the linker
          * includes a segment for it. Otherwise the section will be empty and
          * the linker will ignore it when defining the segments.
+         * If less then 4 bytes, some linkers set this section to size 0
+         * and openocd fails to write it.
```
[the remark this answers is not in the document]
alexandruradovici: The linker used to build Tock for ARM I guess. I think it is the same for all Cortex-M chips. For some reason, after building the elf file, the apps section has size 0 if only a 0 byte is placed there. Somehow gdb sends the vEraseFlash command with size 0, which means erase all page. This fails as openocd (al the least the Raspberry Pi Pico port) requires this to be at the beginning of the flash.

This is the issue that we reported to Raspberry PI https://github.com/raspberrypi/openocd/issues/25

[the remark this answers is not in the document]
alexandruradovici: > Hmm... Have you triggered a panic!, does it have the correct information in it?

I think so, it is similar to the the M3 information, but my knowledge is limited here, so I would appreciate someone double checking it.

boards/raspberry_pi_pico/src/main.rs
```diff
@@ -0,0 +1,424 @@
...
+    // CLK PERI = clk_sys. Used as reference clock for Peripherals. No dividers so just select and enable
+    // Normally choose clk_sys or clk_usb
+    peripherals
+        .clocks
+        .configure_peripheral(PeripheralAuxiliaryClockSource::System, 125000000);
+}
+
+/// Entry point in the vector table called on hard reset.
+#[no_mangle]
+
```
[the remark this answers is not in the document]
ioanaculic: Done

boards/raspberry_pi_pico/src/main.rs
```diff
@@ -0,0 +1,424 @@
...
+            capsules::led::DRIVER_NUM => f(Some(self.led)),
+            kernel::ipc::DRIVER_NUM => f(Some(&self.ipc)),
+            capsules::adc::DRIVER_NUM => f(Some(self.adc)),
+            capsules::temperature::DRIVER_NUM => f(Some(self.temperature)),
+            _ => f(None),
+        }
+    }
+}
+
+/// Entry point used for debuger
```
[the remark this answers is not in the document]
ioanaculic: Done

boards/kernel_layout.ld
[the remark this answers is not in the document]
alexandruradovici: The panic message seems fine, I deleted the code that stops at a break point.

arch/cortex-m0/src/lib.rs
alexandruradovici: The panic message seems fine, I deleted the code that stops at a break point.

boards/kernel_layout.ld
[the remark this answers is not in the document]
alistair23: Can you link to the issue in this comment?

chips/rp2040/README.md
```diff
@@ -0,0 +1 @@
+# RP2040
```
alistair23, line 1: Probably worth adding some more details here.

boards/kernel_layout.ld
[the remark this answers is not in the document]
alexandruradovici: Done

ioanaculic:
I think this is ready to go.

chips/rp2040/src/chip.rs
```diff
@@ -0,0 +1,196 @@
...
+use crate::xosc::Xosc;
+use cortexm0p::interrupt_mask;
+
+#[repr(u8)]
+pub enum Processor {
+    Processor0 = 0,
+    Processor1 = 1,
+}
+
+pub struct Rp2040<'a, I: InterruptService<DeferredCallTask> + 'a> {
```
lschuermann, line 26: Given that this itself chip doesn't actually need any deferred calls, you can throw the `deferred_call_tasks` module out and use a unit `()` for this type parameter as a placeholder instead, like done in `e310x`: https://github.com/tock/tock/blob/1b3c6a14cf22d3ec2959a22c69b69e336aac7903/chips/e310x/src/chip.rs#L18

chips/rp2040/src/chip.rs
```diff
@@ -0,0 +1,196 @@
...
+                self.pins.handle_interrupt();
+                true
+            }
+            _ => false,
+        }
+        // true
+    }
+
+    unsafe fn service_deferred_call(&self, _task: DeferredCallTask) -> bool {
+        // match task {
```
lschuermann, line 190: This comment can be removed.

chips/rp2040/src/chip.rs
```diff
@@ -0,0 +1,196 @@
...
+                self.adc.handle_interrupt();
+                true
+            }
+            interrupts::IO_IRQ_BANK0 => {
+                self.pins.handle_interrupt();
+                true
+            }
+            _ => false,
+        }
+        // true
```
lschuermann, line 186: Please remove.

chips/rp2040/src/clocks.rs
```diff
@@ -0,0 +1,1399 @@
...
+        /// Interrupt Enable
+        (0x0BC => inte: ReadWrite<u32>),
+        /// Interrupt Force
+        (0x0C0 => intf: ReadWrite<u32>),
+        /// Interrupt status after masking & forcing
+        (0x0C4 => ints: ReadWrite<u32>),
+        (0x0C8 => @END),
+    },
+    PllRegisters {
+        /// Control and Status\n
```
lschuermann, line 100: Nit: Can you remove the line feed control sequences in these comments? Applies throughout the file.

lschuermann (commented):
This pretty cool, just ordered my Pico! I did a first pass over the files and have some (partly stylistic and nitpicking) comments. Going to have a closer look when I have the HW in front of me. Thanks for the port!

arch/cortex-m/src/lib.rs
```diff
@@ -239,7 +239,7 @@ pub unsafe extern "C" fn initialize_ram_jump_to_main() {
     ldr r0, ={sbss}     // r0 = first address of .bss
     ldr r1, ={ebss}     // r1 = first address after .bss
 
-    mov r2, #0          // r2 = 0
+    movs r2, #0          // r2 = 0
```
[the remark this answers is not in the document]
alexandruradovici: It seems that armv6/thumbv1 does not have the mov instruction. It only has movs (mov and set flags).

boards/components/src/temperature_rp2040.rs
```diff
@@ -0,0 +1,59 @@
+use capsules::temperature_rp2040::TemperatureRp4020;
```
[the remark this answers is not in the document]
alexandruradovici: > Not necessarily opposed to this, but given components are made to foster code reuse across board definitions, is this component required? Are there going to be a multitude of boards using the RP2040?
> 

Several boards use RP2040 like the Sparkfun RP2040 family, Feather RP2040 family and the Ardunio RP2040 board.

> Seeing that this is very similar compared to the `temperature_stm` component, maybe we can instead try to combine these two into a `temperature_adc` component? I realize this might be difficult with how the types work currently.

I agree with this, I suggest we leave this for a future PR as it requires a bit more work.

boards/components/src/temperature_rp2040.rs
```diff
@@ -0,0 +1,59 @@
...
+use kernel::static_init_half;
+
+#[macro_export]
+macro_rules! temperaturerp4020_adc_component_helper {
+    ($A:ty, $channel:expr, $adc_mux:expr $(,)?) => {{
+        use capsules::temperature_rp2040::TemperatureRp4020;
+        use capsules::virtual_adc::AdcDevice;
+        use core::mem::MaybeUninit;
+        use kernel::hil::adc::Adc;
+        let mut temperature_stm_adc: &'static capsules::virtual_adc::AdcDevice<'static, $A> =
```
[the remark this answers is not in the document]
alexandruradovici: Yes. @cosmindanielradu19  please fix this.

boards/raspberry_pi_pico/Makefile
```diff
@@ -0,0 +1,32 @@
...
+TOCK_ARCH=cortex-m0p
+TARGET=thumbv6m-none-eabi
+PLATFORM=raspberry_pi_pico
+
+include ../Makefile.common
+
+OPENOCD=openocd
+OPENOCD_OPTIONS=-f openocd.cfg
+
+APP=../../../libtock-c/examples/blink/build/cortex-m0/cortex-m0.tbf
```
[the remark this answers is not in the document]
lschuermann: I see this is done in some other boards as well, preparing a PR...

chips/rp2040/src/timer.rs
```diff
@@ -0,0 +1,266 @@
...
+        if !now.within_range(reference, expire) {
+            expire = now;
+        }
+
+        if expire.wrapping_sub(now) < self.minimum_dt() {
+            expire = now.wrapping_add(self.minimum_dt());
+        }
+
+        self.registers.alarm0.set(expire.into_u32());
+        self.enable_timer_interrupt();
```
hudson-ayers, line 238: why is this needed? Is the other `enable_interrupt` insufficient?

chips/rp2040/src/timer.rs
```diff
@@ -0,0 +1,266 @@
...
+                n.enable();
+            })
+        }
+    }
+
+    fn disable_timer_interrupt(&self) {
+        unsafe {
+            cortexm0p::nvic::Nvic::new(TIMER_IRQ_0).disable();
+        }
+    }
```
hudson-ayers, line 204: flagging these as well as hopefully not needed

chips/rp2040/src/uart.rs
```diff
@@ -0,0 +1,733 @@
...
+    }
+
+    fn uart_is_writable(&self) -> bool {
+        !self.registers.uartfr.is_set(UARTFR::TXFF)
+    }
+
+    pub fn send_byte(&self, data: u8) {
+        while !self.uart_is_writable() {}
+        self.registers.uartdr.write(UARTDR::DATA.val(data as u32));
+    }
```
hudson-ayers, line 462: This type of blocking interface is not favored in the tock kernel, see https://github.com/tock/tock/blob/master/doc/Design.md#split-phase-operation . While some exceptions exist, they are generally only for cases where blocking is expected to be very short

chips/rp2040/src/uart.rs
```diff
@@ -0,0 +1,733 @@
...
+        self.registers
+            .uartcr
+            .modify(UARTCR::UARTEN::SET + UARTCR::TXE::SET + UARTCR::RXE::SET);
+
+        self.registers
+            .uartdmacr
+            .write(UARTDMACR::TXDMAE::SET + UARTDMACR::RXDMAE::SET);
+
+        let n = unsafe { cortexm0p::nvic::Nvic::new(self.interrupt) };
+        n.enable();
```
hudson-ayers, line 651: flagging this as also hopefully not needed

hudson-ayers (changes requested):
This overall looks quite good, I do have several changes though

chips/rp2040/src/uart.rs
alexandruradovici: This is used only for panic

chips/rp2040/src/timer.rs
alexandruradovici: From our tests that does not seem to be enough, I'll dig a little further and get back.

chips/rp2040/src/adc.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    fn enable_interrupt(&self) {
+        let n = unsafe { cortexm0p::nvic::Nvic::new(interrupts::ADC_IRQ_FIFO) };
+        n.enable();
+    }
+
+    fn disable_interrupt(&self) {
+        let n = unsafe { cortexm0p::nvic::Nvic::new(interrupts::ADC_IRQ_FIFO) };
+        n.disable();
+    }
```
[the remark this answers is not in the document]
alexandruradovici: From our tests that does not seem to be enough, I'll dig a little further and get back.

boards/raspberry_pi_pico/README.md
```diff
@@ -0,0 +1,76 @@
...
+The [Raspberry Pi Pico](https://www.raspberrypi.org/products/raspberry-pi-pico/) is a
+board developed by the Raspberry Pi Foundation and is based on the RP2040 chip.
+
+## Getting Started
+
+First, follow the [Tock Getting Started guide](../../../doc/Getting_Started.md)
+
+## Flashing the kernel
+
+The kernel can be programmed using a second Raspberry Pi device and OpenOCD. OpenOCD needs to be installed on the Regular Raspberry Pi device.
```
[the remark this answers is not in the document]
alexandruradovici: It uses a custom version of `openocd` [1] that does bit-banging over the (standard) Raspberry Pi pins. Another solution is to use another Pico as a (bit banging) programmer. @ioanaculic can you document this?

[1] https://github.com/raspberrypi/openocd

chips/rp2040/src/uart.rs
hudson-ayers: Oh, thanks, sorry I didn't trace through and realize that. In that case it is fine.

boards/raspberry_pi_pico/README.md
```diff
@@ -0,0 +1,76 @@
...
+(wait to connect)
+
+$ openocd -f interface/raspberrypi-swd.cfg -f target/rp2040.cfg
+```
+You can also open a serial console on the Raspberry Pi for debug messages.
+```bash
+$ sudo apt install minicom
+$ minicom -b 115200 -o -D /dev/serial0
+```
+
```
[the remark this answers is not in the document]
alexandruradovici: At least openocd has to run on the Pi. Of course, using a serial TTL adapter works for using the serial port on the computer. But since most users will have a Raspberry Pi, this was an easier way, by simply adding three cables.

boards/raspberry_pi_pico/Makefile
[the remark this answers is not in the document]
ioanaculic: I changed this, but when running make it still asks for the APP variable to be defined. Am I missing something? I placed below the command I tried to run.

```
$ APP=file make program
    Finished dev [optimized + debuginfo] target(s) in 0.04s
   text    data     bss     dec     hex filename
 106500       0    8192  114692   1c004 /Users/bender/software/tock/target/thumbv6m-none-eabi/debug/raspberry_pi_pico
96949ef6e63c284c6febcd11bb109b1960fdc6d3dea76044bee2c97194f400ad  /Users/bender/software/tock/target/thumbv6m-none-eabi/debug/raspberry_pi_pico.bin
Makefile:29: *** Please define the APP variable with the TBF file to flash an application.  Stop.
```

boards/README.md
```diff
@@ -6,6 +6,7 @@ that Tock supports.
 
 | Board                                                                | Architecture    | MCU            | Interface  | App deployment | QEMU Support? |
 |----------------------------------------------------------------------|-----------------|----------------|------------|----------------|---------------|
+| [Raspberry Pi Pico](raspberry_pi_pico/README.md)                     | ARM Cortex-M0+  | RP2040         | openocd    | openocd        | No            |
```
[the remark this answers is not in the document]
ioanaculic: I noticed that the boards were ordered based on the alphabetical order of the chip. If this is different, I can move it to the bottom.

arch/cortex-m/src/nvic.rs
```diff
@@ -134,6 +152,26 @@ pub unsafe fn next_pending() -> Option<u32> {
...
 
+pub unsafe fn next_pending_with_mask(mask: (u128, u128)) -> Option<u32> {
+    for (block, ispr) in NVIC
+        .ispr
+        .iter()
+        .take(number_of_nvic_registers())
+        .enumerate()
+    {
+        let interrupt_mask = if block < 4 { mask.1 } else { mask.0 };
+        let ispr = ispr.get() & !((interrupt_mask >> (32 * block % 4)) as u32);
```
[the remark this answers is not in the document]
ioanaculic: done

boards/components/src/temperature_rp2040.rs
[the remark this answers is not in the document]
ioanaculic: done

chips/rp2040/src/chip.rs
ioanaculic: done

chips/rp2040/src/chip.rs
ioanaculic: done

chips/rp2040/src/clocks.rs
ioanaculic: done

boards/raspberry_pi_pico/README.md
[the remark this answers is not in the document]
ioanaculic: done

boards/raspberry_pi_pico/README.md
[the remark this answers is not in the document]
ioanaculic: done

chips/rp2040/src/adc.rs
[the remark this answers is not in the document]
ioanaculic: done

chips/rp2040/src/deferred_call_tasks.rs
```diff
@@ -0,0 +1,27 @@
+//! Definition of Deferred Call tasks.
```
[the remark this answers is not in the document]
ioanaculic: done

chips/rp2040/src/timer.rs
ioanaculic: We tested it and it does not work without it. I have commented this in the source.

chips/rp2040/src/timer.rs
ioanaculic: We tested it and it does not work without these. I have commented these in the source.

chips/rp2040/src/uart.rs
ioanaculic: done

capsules/src/temperature_rp2040.rs
```diff
@@ -0,0 +1,66 @@
+//! Driver for STM ADC MCU temperature sensor
```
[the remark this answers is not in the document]
ioanaculic: done

chips/rp2040/src/chip.rs
ioanaculic: done

boards/raspberry_pi_pico/src/io.rs
```diff
@@ -0,0 +1,74 @@
...
+#[no_mangle]
+#[panic_handler]
+pub unsafe extern "C" fn panic_fmt(pi: &PanicInfo) -> ! {
+    // LED is conneted to GPIO 25
+    let led_kernel_pin = &RPGpioPin::new(RPGpio::GPIO25);
+    let led = &mut LedHigh::new(led_kernel_pin);
+    let writer = &mut WRITER;
+    let uart0 = &Uart::new_uart0();
+
+    if !uart0.is_configured() {
```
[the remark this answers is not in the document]
alexandruradovici: I changed it so that the board can setup which uart to use for the panic print. If no uart is set, it will use uart0 as default.

arch/cortex-m/src/nvic.rs
```diff
@@ -134,6 +152,26 @@ pub unsafe fn next_pending() -> Option<u32> {
     None
 }
 
+pub unsafe fn next_pending_with_mask(mask: (u128, u128)) -> Option<u32> {
```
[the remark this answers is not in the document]
alexandruradovici: Done

arch/cortex-m/src/nvic.rs
```diff
@@ -14,6 +14,24 @@ use kernel::common::registers::interfaces::{Readable, Writeable};
 use kernel::common::registers::{register_bitfields, register_structs, ReadOnly, ReadWrite};
 use kernel::common::StaticRef;
 
+#[macro_export]
```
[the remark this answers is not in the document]
alexandruradovici: Done

boards/raspberry_pi_pico/src/flash_bootloader.rs
```diff
@@ -0,0 +1,19 @@
+/// padded bootloader used to boot from flash
```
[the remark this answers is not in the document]
alexandruradovici: I added the link to the datasheet explaining the bootloader and to the source files.

boards/README.md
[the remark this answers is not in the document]
alexandruradovici: @bradjc should this be placed in alphabetical order or at the bottom?

chips/rp2040/src/timer.rs
```diff
@@ -0,0 +1,282 @@
...
+
+        if expire.wrapping_sub(now) < self.minimum_dt() {
+            expire = now.wrapping_add(self.minimum_dt());
+        }
+
+        self.registers.alarm0.set(expire.into_u32());
+        self.enable_timer_interrupt();
+        self.enable_interrupt();
+
+        // panic!("{} now {:?}", self.registers.alarm0.get(), self.now());
```
alistair23, line 251: This should be removed

alistair23 (approved):

chips/rp2040/src/timer.rs
ioanaculic: done

boards/README.md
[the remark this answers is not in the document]
hudson-ayers: I think you mean by alphabetical order of the arch? I think that the boards are grouped by arch, but the order ending up being alphabetical is mostly just luck because there are really only 3 architectures so far. I think putting this at the end of the ARM boards makes sense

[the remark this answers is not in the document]
alexandruradovici: Done :)

hudson-ayers (approved):

alexandruradovici:
Ping?

boards/raspberry_pi_pico/src/main.rs
```diff
@@ -0,0 +1,431 @@
...
+/// by default. Without this function, gdb sets the PC to the
+/// beginning of the flash. This is not correct, as the RP2040
+/// has a more complex boot process.
+///
+/// This function is set to be the entry point for gdb and is used
+/// to send the RP2040 back in the bootloader so that all the boot
+/// sqeuence is performed.
+#[no_mangle]
+#[naked]
+pub unsafe extern "C" fn reset() {
```
bradjc, line 103:
```suggestion
pub unsafe extern "C" fn jump_to_bootloader() {
```
?
Or is the name `reset` required?

bradjc (approved):

boards/raspberry_pi_pico/src/main.rs
ioanaculic: done

bradjc (approved):

hudson-ayers (approved):

lschuermann:
Looks good, although I unfortunately didn't manage to get access to the hardware to test it in time.

One question: this adds a bunch of in-development commits. Might it make sense to squash-merge here?

alistair23 (approved):

PR #2445, opened 2021-02-19
Tock 2.0: implement Callback swapping restrictions

lschuermann:
### Pull Request Overview

This pull request is a follow-up to #2282, which was especially criticized for being only a part of the changes required to uphold all safety requirements we'd like to have with regards to Callback swapping restrictions. It's one implementation strategy out of many.

#### Issue

Essentially, the protections introduced in this PR are required because in the Tock 2.0 system call architecture, with every call to `subscribe`, the previous `Callback` provided by the application must be returned. If we were to not enforce any constraints here, this would have serious side effects:

- a `Callback` subscribed by process A could be handed back to process B.

  This leaks both the callback pointer and the `appdata` field of the `Callback` to another application.

- a `Callback` from one capsule could be given to another capsule and returned as part of a call to subscribe there.
- a `Callback`, passed to a driver under subscribe (subdriver) number `x` could be returned as part of a call to `subscribe` to the same driver under subscribe (subdriver) number `y`, where `x != y`.
- a capsule could pass back a _null callback_ (with callback pointer and appdata being 0), where a process has actually subscribed a `Callback` with non-null values before.

  All of these cases could cause inconsistencies and undefined behavior in userspace, if a process were to rely on the fact that the returned `Callback` is the one previously passed to the capsule.

  It was indicated that userspace libraries might want to rely on this behaviour for stack-allocated callbacks (especially in libtock-rs).

#### Guarantees made by this PR 

As a whole, this PR enables the kernel to check for these various cases

- a capsule cannot return a `Callback` originally passed to a different capsule in a `subscribe` invocation
- a capsule cannot return, as a result of a `subscribe` invocation, a callback which was provided by a different process
- a capsule cannot return, as a result of a `subscribe` invocation, a callback which was provided by the same process under a different subscribe / subdriver number
- a capsule cannot return a _null callback_ (e.g. a default Callback instance) where a process had subscribe a non-null `Callback` before

#### Why do we need a _default_ AND a _null_ callback?

As callbacks are commonly stored in Grant regions which must implement the `Default` trait, there must be some instance of a `Callback` which is also instantiable either through the `Default` trait, or without involving other components such as the core kernel. It is impractical to change Grant regions to have an alternative `Default`-like trait implementation for constructing them with a context.

However, if a capsule can construct an arbitrary number of `Callback` instances, this poses a problem: how can we know whether a `Callback` returned as a result of a `subscribe` invocation is an arbitrary default instance, or one which has been set by the process explicitly and how can we know whether this is the first `subscribe` invocation for this `(driver, subdriver, process)` combination, where it is legal to return a default callback instance?

This PR proposes a particular solution to this issue: by distinguishing between _default_ and _null_ callbacks (where _null_ callbacks are always bound to a specific process), and replacing all _default_ callbacks with process-bound callbacks upon process initialization, we can be sure that if a capsule returns a _null_ callback, it must belong to a specific process and couldn't have been created arbitrarily. Furthermore, `Callback`s do not implement `Clone` any longer, such that for each `(driver, subdriver, process)` combination only a single `Callback` instance can exist in the kernel, except for the time where a subscribe operation is being processed by the kernel.

Semantically, `struct Callback(Option<ProcessCallback>)`, where `Callback(None)` is a default callback and `Callback(ProcessCallback { appid: some_app, appdata: 0, fn_ptr: 0x0 })` is a process-bound null callback, is equivalent to just using `Option<ProcessCallback>` in the capsules directly. In my opinion that's also a bit cleaner, as it expresses the need to create a process-bound callback through the type system itself. However, I didn't want to do extensive changes to every capsule in this PR. it's already sufficiently long.

#### Implications

- This PR removes `Copy, Clone` trait implementations from `Callbacks`. I don't see any solution which wouldn't heavily modify larger parts of the OS to get around this. Most approaches I've come up with use the fact that through ownership and move semantics, we can control & limit the lifetime of `Callback`s effectively and at compile time.

- This PR adds another method to the `Driver` trait, which is called upon process initialization. This gives capsules a chance to replace all of their `Callback`s with process-bound _null_ `Callback`s by use of a `ProcessCallbackFactory` which guarantees that for each `(driver, subdriver, process)` combination only a single null `Callback` can be initialized.

  This can even be extended through an `AppSliceFactory` later, which would also register how many `AppSlice`s a specific capsule for a given process holds, such that the required AppSlice table in the kernel can be sized sufficiently.

  However, it introduces another hook where capsules are called and handed control in the system.

- This PR adds another method to the `Platform` trait, which can be used to invoke a closure on all installed drivers on the system. This is used to message global events to all capsules, such as process initialization.

  One could think about using this infrastructure to also communicate process exits / dies. This can, for instance, be used to turn off wireless radios when no process uses the hardware any more.

- This implementation strategy, in contrast to some others, does not require any central registry or other data structures. By tightly controlling the instances of the `Callback` type at compile time, we can uphold the desired guarantees without using additional data fields.

### Testing Strategy

This pull request was tested by trying to swap callbacks on a Hail board. It need more testing.

### TODO or Help Wanted

This pull request has the required changes to capsules and platforms implemented for only a small subset capsules and platforms respectively. I'd like to get feedback on this first prior to putting in even more work.

Currently, this PR only prints debug messages when a capsule misbehaves. This is done for compatibility and easier testing. Different handling strategies could be implemented later, such as "blocking" the capsule from further operation or panicing the system.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/boards/hail/src/main.rs
+++ b/boards/hail/src/main.rs
@@ -106,6 +106,28 @@ impl Platform for Hail {
             _ => f(None),
         }
     }
+
+    fn iter_drivers<F>(&self, f: F)
+    where
+        F: Fn(usize, &dyn kernel::Driver),
+    {
+        f(capsules::console::DRIVER_NUM, self.console);
+        f(capsules::gpio::DRIVER_NUM, self.gpio);
+        f(capsules::alarm::DRIVER_NUM, self.alarm);
+        f(capsules::spi_controller::DRIVER_NUM, self.spi);
+        // TODO: nrf51822_serialization
+        f(capsules::ambient_light::DRIVER_NUM, self.ambient_light);
+        f(capsules::adc::DRIVER_NUM, self.adc);
+        f(capsules::led::DRIVER_NUM, self.led);
+        f(capsules::button::DRIVER_NUM, self.button);
+        f(capsules::humidity::DRIVER_NUM, self.humidity);
+        f(capsules::temperature::DRIVER_NUM, self.temp);
+        f(capsules::ninedof::DRIVER_NUM, self.ninedof);
+        f(capsules::rng::DRIVER_NUM, self.rng);
+        f(capsules::crc::DRIVER_NUM, self.crc);
+        f(capsules::dac::DRIVER_NUM, self.dac);
+        // TODO: ipc
+    }
 }
 
 /// Helper function called during bring-up that configures multiplexed I/O.
--- a/boards/hail/src/main.rs
+++ b/boards/hail/src/main.rs
@@ -489,6 +489,7 @@ pub unsafe fn reset_handler() {
     kernel::procs::load_processes(
         board_kernel,
         chip,
+        &hail,
         core::slice::from_raw_parts(
             &_sapps as *const u8,
             &_eapps as *const u8 as usize - &_sapps as *const u8 as usize,
--- a/boards/litex/sim/src/main.rs
+++ b/boards/litex/sim/src/main.rs
@@ -142,6 +142,14 @@ impl Platform for LiteXSim {
             _ => f(None),
         }
     }
+
+    fn iter_drivers<F>(&self, f: F)
+    where
+        F: Fn(usize, &dyn kernel::Driver),
+    {
+        f(capsules::console::DRIVER_NUM, self.console);
+        f(capsules::alarm::DRIVER_NUM, self.alarm);
+    }
 }
 
 /// Reset Handler.
--- a/boards/litex/sim/src/main.rs
+++ b/boards/litex/sim/src/main.rs
@@ -404,6 +404,7 @@ pub unsafe fn reset_handler() {
     kernel::procs::load_processes(
         board_kernel,
         chip,
+        &litex_sim,
         core::slice::from_raw_parts(
             &_sapps as *const u8,
             &_eapps as *const u8 as usize - &_sapps as *const u8 as usize,
--- a/capsules/src/adc.rs
+++ b/capsules/src/adc.rs
@@ -1187,9 +1187,8 @@ impl<A: hil::adc::Adc + hil::adc::AdcHighSpeed> Driver for AdcDedicated<'_, A> {
             // subscribe to ADC sample done (from all types of sampling)
             0 => {
                 // set callback
-                self.appid.map_or(Err((callback, ErrorCode::FAIL)), |id| {
-                    let res = self
-                        .apps
+                let res = self.appid.map_or(Err(ErrorCode::FAIL), |id| {
+                    self.apps
                         .enter(*id, |app, _| {
                             mem::swap(&mut app.callback, &mut callback);
                         })
@@ -1200,13 +1199,14 @@ impl<A: hil::adc::Adc + hil::adc::AdcHighSpeed> Driver for AdcDedicated<'_, A> {
                                 self.appid.clear();
                             }
                             ErrorCode::from(err)
-                        });
-                    if let Err(err) = res {
-                        Err((callback, err))
-                    } else {
-                        Ok(callback)
-                    }
-                })
+                        })
+                });
+
+                if let Err(err) = res {
+                    Err((callback, err))
+                } else {
+                    Ok(callback)
+                }
             }
 
             // default
--- a/capsules/src/alarm.rs
+++ b/capsules/src/alarm.rs
@@ -16,7 +16,6 @@ enum Expiration {
     Enabled { reference: u32, dt: u32 },
 }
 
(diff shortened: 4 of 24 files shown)


phil-levis:
> Semantically, `struct Callback(Option<ProcessCallback>)`, where `Callback(None)` is a default callback and `Callback(ProcessCallback { appid: some_app, appdata: 0, fn_ptr: 0x0 })` is a process-bound null callback, is equivalent to just using `Option<ProcessCallback>` in the capsules directly. In my opinion that's also a bit cleaner, as it expresses the need to create a process-bound callback through the type system itself. However, I didn't want to do extensive changes to every capsule in this PR. it's already sufficiently long.

The current semantics of subscribe in the TRD indicate that the Null Upcall **can** have user data with it, which will be correctly stored and returned on a subsequent call to `subscribe`. Should we restrict this such that the `appdata` field must also be 0? I am tempted to say no, that `appdata` acts normally.

phil-levis:
>     * This PR adds another method to the `Platform` trait, which can be used to invoke a closure on all installed drivers on the system. This is used to message global events to all capsules, such as process initialization.

Does this mean that all capsules which store a Callback in a grant region will allocate their grant region? We've so far been relying on lazy grant allocation, such that a kernel allocates grant regions only for capsules that use them. With this, we'll have to always have a grant region large enough for every possible grant?

phil-levis:
>     * This implementation strategy, in contrast to some others, does not require any central registry or other data structures. By tightly controlling the instances of the `Callback` type at compile time, we can uphold the desired guarantees without using additional data fields.

+1

lschuermann:
> Does this mean that all capsules which store a Callback in a grant region will allocate their grant region?

Yes, that is an unfortunate consequence of this PR. Enforcing the desired invariants while not increasing memory demands significantly and providing sufficiently good developer ergonomics is a constrained optimization problem I've spent a lot of work on, and each time I seem to miss one aspect. :disappointed: 

My reverse ask would be: how large is the impact of the lazy grant allocation, really? For sure, apps generally don't use every capsule installed on a board, although I expect reasonably complex processes to interact with a bunch. Furthermore, is there a way around this (for instance, could we force processes to initialize drivers prior to first use, so force the process to trigger the call to `Driver::process_init`)? I fear something like that might change the semantics around Tock apps quite a bit.

Without trying to mess up Tock's core ideas too much, could one think about making Grant allocation more explicit? I used to not know about this and took their seemingly static existence for granted (no pun intended). Still today I forget about it sometimes (like here). If we could think about an explicit `.allocate()`, it could map well to a process explicitly requesting/initializing a driver, which seems to be common around other OSes as well.

hudson-ayers:
> My reverse ask would be: how large is the impact of the lazy grant allocation, really?

I looked into this briefly: on Imix today, allocating every grant for every process requires about 1100 bytes of memory per process. For our default of 4 processes per board, that is 4400 bytes of memory for grants. 

> Furthermore, is there a way around this (for instance, could we force processes to initialize drivers prior to first use, so force the process to trigger the call to Driver::process_init)?

I think there are a lot of reasons this would be nice, but one thing to consider is that dynamically loadable apps are a long-standing goal of Tock, and we probably don't want to commit to a kernel design that makes dynamically loaded apps impossible.

jrvanwhy:
I have been trying to support extremely-small apps in libtock-rs. I shrunk one of the tock-on-titan apps down to 648 bytes of RAM, not counting the grant region.

hudson-ayers:
One idea: would it be possible to construct the default callbacks as part of the existing lazy initialization? 

The lazy initialization is in `kernel/src/grant.rs`, in `AppliedGrant::get_or_allocate()`. Could we call `process_init()` from within that function, after creating the default version of `T`, but before creating the `AppliedGrant`?

That also might make it possible to re-use the `grant_index` to uniquely identify grant regions, rather than using `driver_num`, which is kinda nice because that means it is still possible for a capsule to have multiple grant regions if that was desired for some reason.

bradjc:
I don't quite understand how this approach works. Is there a step-by-step explanation of exactly what the procedure is that the kernel goes though to implement the callback swap restriction? That would be good to have in a comment somewhere.

----

Pre-allocating every grant effectively removes the grant mechanism entirely. The net effect is we no longer do per-process dynamic memory allocation, and instead just statically allocate enough memory for every possible kernel operation, whether a process cares about it or not. I don't think we should pursue this, but if we do we should probably re-write the grant code since it no longer would need to do any dynamic allocations.

lschuermann:
Superseeded by #2462 because of the obvious drawbacks regarding Grant region allocation mentioned by @phil-levis and others. Thanks for the feedback!

PR #2448, opened 2021-02-20
Better method parsing for accounting for code size.

phil-levis:
### Pull Request Overview

This pull request improves the `print_tock_memory_usage` tool. It parses method names inside traits, so you can now account for individual methods. It also allows you to order symbol groups by size.

It changes the `-s` option to be sort by size; the previous `-s`, show waste, is now `-w`.

### Testing Strategy

This pull request was tested by running it on an imix ELF file.

### TODO or Help Wanted

This pull request still needs some use, make sure it's not missing some cases that imix doesn't trigger.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/tools/github_actions_size_changes.sh
+++ b/tools/github_actions_size_changes.sh
@@ -16,7 +16,7 @@ make allboards > /dev/null 2>&1
 for elf in $(find . -maxdepth 8 | grep 'release' | egrep '\.elf$' | grep -v 'riscv'); do
     tmp=${elf#*release/}
     b=${tmp%.elf}
-    ./tools/print_tock_memory_usage.py -s ${elf} > current-benchmark-${b}
+    ./tools/print_tock_memory_usage.py -w ${elf} > current-benchmark-${b}
 done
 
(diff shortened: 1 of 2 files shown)


bradjc (dismissed):

hudson-ayers (approved):
I added a commit to update the github actions size tooling to reflect the change to the CLI options.

Looking at the output of the "benchmarks" github action confirms that the tool still runs successfully for every upstream Tock board with the default options, so that is a good sign!

One comment for the future: Rust will soon be switching to a new symbol mangling approach (tracked [here](https://github.com/rust-lang/rust/issues/60705)), which will surely break the cxxfilt-based demangling this tool uses today. To get around this issue, and the various issues this PR surfaces where manual manipulation of symbols is needed, we will probably want to update this tool to use the C-api to https://github.com/alexcrichton/rustc-demangle, which will always be kept up-to-date with the current rust mangling approach, and which will be capable of demangling symbols from both the old and new scheme.

hudson-ayers:
ah, my commit actually causes the benchmarks PR to fail, because it runs size measurements for current master and compares against the head of the PR, but on current master the `-w` option does not exist. That issue will resolve itself just by merging this PR.

bradjc (approved):

PR #2449, opened 2021-02-22
arch: cortex-m: update to asm!

bradjc:
### Pull Request Overview

This pull request switches the last llvm_asm in cortex-m to asm.

This was somewhat straightforward, except that Rust will not let me mark r6 as clobbered (`out("r6") _`) like the other registers. I get:

```
error: invalid register `r6`: r6 is used internally by LLVM and cannot be used as an operand for inline asm
   --> arch/cortex-m/src/lib.rs:254:31
    |
254 |     out("r4") _, out("r5") _, out("r6") _, out("r8") _, out("r9") _,
    |                               ^^^^^^^^^^^
```

I found that if I manually saved r6 in r3 the kernel still works.

#1923

### Testing Strategy

Running hail app on hail.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/lib.rs
+++ b/arch/cortex-m/src/lib.rs
@@ -221,11 +221,14 @@ pub unsafe extern "C" fn switch_to_user_arm_v7m(
 ) -> *const usize {
     asm!(
     "
-    // Rust `asm!()` macro (as of Feb 2021) will not let us mark r6 as a
-    // clobber. However, in the process of restoring and saving the process's
-    // registers, we do in fact clobber r6. So, we work around this by doing our
-    // own manual saving of r6 using r3, and then mark r3 as clobbered.
-    mov r3, r6
+    // Rust `asm!()` macro (as of Feb 2021) will not let us mark r6 or r7 as
+    // clobbers. r6 is used internally by LLVM, and r7 is used for the frame
+    // pointer. However, in the process of restoring and saving the process's
+    // registers, we do in fact clobber r6 and r7. So, we work around this by
+    // doing our own manual saving of r6 using r2 and r7 using r3, and then mark
+    // both as clobbered.
+    mov r2, r6
+    mov r3, r7
 
     // The arguments passed in are:
     // - `r0` is the bottom of the user stack
@@ -255,13 +258,14 @@ pub unsafe extern "C" fn switch_to_user_arm_v7m(
     // application has executed.
     mrs r0, PSP   // r0 = PSP
 
-    // Need to restore r6 since we clobbered it when switching to and from the
-    // app.
-    mov r6, r3
+    // Need to restore r6 and r7 since we clobbered them when switching to and
+    // from the app.
+    mov r6, r2
+    mov r7, r3
     ",
     inout("r0") user_stack,
     in("r1") process_regs,
-    out("r3") _, out("r4") _, out("r5") _, out("r8") _, out("r9") _,
+    out("r2") _, out("r3") _, out("r4") _, out("r5") _, out("r8") _, out("r9") _,
     out("r10") _, out("r11") _);
 
     user_stack
--- a/arch/cortex-m/src/lib.rs
+++ b/arch/cortex-m/src/lib.rs
@@ -2,7 +2,6 @@
 
 #![crate_name = "cortexm"]
 #![crate_type = "rlib"]
-#![feature(llvm_asm)]
 #![feature(asm)]
 #![feature(naked_functions)]
 #![no_std]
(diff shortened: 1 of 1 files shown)


arch/cortex-m/src/lib.rs
```diff
@@ -220,18 +219,25 @@ pub unsafe extern "C" fn switch_to_user_arm_v7m(
...
 ) -> *const usize {
-    llvm_asm!(
-        "
+    asm!(
+    "
+    // Rust `asm!()` macro (as of Feb 2021) will not let us mark r6 as a
+    // clobber. However, in the process of restoring and saving the process's
+    // registers, we do in fact clobber r6. So, we work around this by doing our
+    // own manual saving of r6 using r3, and then mark r3 as clobbered.
+    mov r3, r6
```
hudson-ayers, line 228: This makes me a little uneasy. I am not sure if it is valid to write to a reserved register in this way (r6, which LLVM uses as the base pointer on ARM). From the Rust inline asm docs:

> In some cases LLVM will allocate a "reserved register" for reg operands even though this register cannot be explicitly specified. Assembly code making use of reserved registers should be careful since reg operands may alias with those registers. Reserved registers are:
>
> - The frame pointer on all architectures.
> - r6 on ARM.

Looking into this a little further, I found [this](https://github.com/rust-lang/rust/issues/73450#issuecomment-650463347) discussion, which seems quite similar. It sounds like saving the value on the stack is the recommended approach -- I also see that approach recommended [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=11807#c25) in the gcc issue discussing how to mark reserved registers as clobbered.

bradjc: We can save the register to the stack, but then what? Don't we have to write it back to r6 still?

hudson-ayers: Yes that's true, we would have to pop it back into the register. and I don't really know why that would be any different than what you suggested here. This just seems like somewhere where inline assembly can be a bit of a minefield of confusing or unexpected interactions with the compiler. I don't totally understand the warnings here or how LLVM will use r6.

bradjc: Also, we are storing r6 on the stack via r3 and the svc call, correct? r0-r3 have to be stored somewhere, right?

arch/cortex-m/src/lib.rs
```diff
@@ -243,14 +249,21 @@ pub unsafe extern "C" fn switch_to_user_arm_v7m(
...
-    : "r4","r5","r6","r8","r9","r10","r11" : "volatile" );
+    mrs r0, PSP   // r0 = PSP
+
+    // Need to restore r6 since we clobbered it when switching to and from the
+    // app.
+    mov r6, r3
+    ",
+    inout("r0") user_stack,
+    in("r1") process_regs,
+    out("r3") _, out("r4") _, out("r5") _, out("r8") _, out("r9") _,
```
ppannuto, line 264: It wasn't here before either.., but why is there no `r7`?

ppannuto (commented):
This looks like a reasonable solution of this problem to me. It's my understanding that compilers generally treat assembly blocks as blind copy/paste segments, what matters is that the input and output invariants are accurate.

arch/cortex-m/src/lib.rs
bradjc: I believe r7 is for the frame pointer, and returns the same error as trying to mark r6 as a clobber.

ppannuto: Do we need to be doing the same thing (under the same premise that userspace can/will likely clobber it)?

arch/cortex-m/src/lib.rs
hudson-ayers: yeah, I guess so. I'm a bit out of my depth here, so I'll resolve this.

arch/cortex-m/src/lib.rs
bradjc: I would say no, since we haven't been so far with no negative consequences.

hudson-ayers: https://github.com/rust-lang/rust/issues/81092#issuecomment-784524675 suggests that it might be safest to do so, though I did not ask explicitly about `r7`

ppannuto: That feels more like we might be getting lucky the being correct. Assuming nothing has changed since my compilers class project in 2012, all else being equal, llvm will allocate registers following a lowest-number-first policy. In practice, it's likely `r7` simply hasn't been live in a meaningful way for kernel code across syscalls, but that doesn't it couldn't be. We should probably save `r7` as well.

bradjc: I don't think we have just been getting lucky, but I think it makes sense to be defensive since we have another register (r2) to spare. I added a commit which saves r7.

hudson-ayers (approved):

ppannuto (approved):

PR #2451, opened 2021-02-23
cfg: add `target_arch_specific`

bradjc:
### Pull Request Overview

Right now Rust only gives us `target_arch`, which is a family, not a specific architecture. This uses our build system to add `target_arch_specific` which allows us to cfg on an actual architecture.

Inspired by #2442.

I view `target_arch` being so vague as a failing of Rust that Rust should remedy either by adding another cfg flag (like this PR), or making `target_arch` actually point to specific architectures. I like this approach because it fills in a Rust gap that Rust may someday take care of for us.

### Testing Strategy

Compiling the cortex-m crate.

### TODO or Help Wanted

Thoughts?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/lib.rs
+++ b/arch/cortex-m/src/lib.rs
@@ -214,7 +214,11 @@ pub unsafe extern "C" fn unhandled_interrupt() {
 /// Assembly function called from `UserspaceKernelBoundary` to switch to an
 /// an application. This handles storing and restoring application state before
 /// and after the switch.
-#[cfg(all(target_arch = "arm", target_os = "none"))]
+#[cfg(all(
+    target_arch = "arm",
+    target_arch_specific = "thumbv7em",
+    target_os = "none"
+))]
 #[no_mangle]
 pub unsafe extern "C" fn switch_to_user_arm_v7m(
     mut user_stack: *const usize,
@@ -254,7 +258,11 @@ pub unsafe extern "C" fn switch_to_user_arm_v7m(
     user_stack
 }
 
-#[cfg(all(target_arch = "arm", target_os = "none"))]
+#[cfg(all(
+    target_arch = "arm",
+    target_arch_specific = "thumbv7em",
+    target_os = "none"
+))]
 #[inline(never)]
 unsafe fn kernel_hardfault_arm_v7m(faulting_stack: *mut u32) -> ! {
     let stacked_r0: u32 = *faulting_stack.offset(0);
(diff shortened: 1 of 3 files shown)


ppannuto (approved):
The lack of arch granularity has been obnoxious more than once, I think this will prove to be a useful addition.

Do we have any evidence that such a thing will ever reach upstream Rust, though? Is this even a problem anyone is aware of?

bradjc:
> Do we have any evidence that such a thing will ever reach upstream Rust, though? Is this even a problem anyone is aware of?

I do not, at least.

hudson-ayers (approved):
This looks good to me. Hopefully eventually `target_feature` will be mature / stable enough to replace this.

bradjc:
#2451 vs #2452, only one can be merged!

jrvanwhy:
Is there a reason you set this via Makefile rather than via a cargo build script?

jrvanwhy:
Ah, Brad sniped me. I prefer the approach in #2452 .

phil-levis:
> 
> 
> Ah, Brad sniped me. I prefer the approach in #2452 .

#2452 requires a nightly feature?

bradjc:
> > Ah, Brad sniped me. I prefer the approach in #2452 .
> 
> #2452 requires a nightly feature?

Yes. The thinking is that this PR is an existence proof of how to remove that feature should every other feature be stabilized before this one.

jrvanwhy:
> > Ah, Brad sniped me. I prefer the approach in #2452 .
> 
> #2452 requires a nightly feature?

I didn't notice the nightly feature requirement when I wrote that. I'm unsure whether that changes my opinion of #2452.

I think doing this in `build.rs` by reading `TARGET` would be better than either this or #2452. That would avoid both the need for a new nightly feature and for changes in downstream build systems (e.g. tock-on-titan's).

bradjc:
The problem with using build.rs is we have ~20 build.rs files which would be a lot of duplicated code. Maybe there is a way to  consolidate it, however.

bradjc:
I think the feature is #2452 is inline with our feature policy. It may or may not get stabilized, but even if it doesn't we have at least one method to remove it should `asm`, `naked_functions`, `const_fn`, and `intrinsics_atomic` all become stabilized first. I imagine this feature is also relatively low risk, as it only relates to including/not including code, and not new logic paths in the compiler.

bradjc:
#2452 is the correct way to do this.

Issue #1736, opened 2020-04-06
single-byte console writes never complete


bcantrill:
Currently, the <a href="https://github.com/tock/libtock-rs/blob/master/examples/timer.rs">timer example from libtock-rs</a> hangs, after emitting a single `[`:

```
Version:    opentitan-snapshot-20191101-1-887-ge849d530
Build Date: 2020-04-03, 11:01:10
I00000 boot_rom.c:44] Boot ROM initialisation has completed, jump into flash!
OpenTitan initialisation complete. Entering main loop
[
```

I looked at a trace under <a href="https://github.com/oxidecomputer/tockilator">Tockilator</a> (and because it's instructive, I have also pushed everything needed to reproduce this to the <a href="https://github.com/oxidecomputer/tockilator/tree/master/example/libtock-rs/timer">example/libtock-rs/timer subdirectory</a>).

For starters, here is the output showing just system calls:

```
% tockilator ./trace_core_00000000.log
169204   => SYSCALL MEMOP operand=0 (brk) arg0=0x10003004
187307   <= SYSCALL MEMOP operand=0 (brk) arg0=0x10003004
187310   => SYSCALL MEMOP operand=10 (update-stack-start) arg0=0x10003400
202583   <= SYSCALL MEMOP operand=10 (update-stack-start) arg0=0x10003400
202586   => SYSCALL MEMOP operand=11 (update-heap-start) arg0=0x10003004
217281   <= SYSCALL MEMOP operand=11 (update-heap-start) arg0=0x10003004
217358   => SYSCALL MEMOP operand=0 (brk) arg0=0x10003404
234333   <= SYSCALL MEMOP operand=0 (brk) arg0=0x10003404
235019   => SYSCALL COMMAND driver=0 (timer) subdriver=0 arg0=0x0 arg1=0x0
250366   <= SYSCALL COMMAND driver=0 (timer) subdriver=0 arg0=0x0 arg1=0x0
250373   => SYSCALL COMMAND driver=0 (timer) subdriver=1 arg0=0x0 arg1=0x0
265330   <= SYSCALL COMMAND driver=0 (timer) subdriver=1 arg0=0x0 arg1=0x0
265339   => SYSCALL SUBSCRIBE driver=0 (timer) subdriver=0 cb=0x20032006 data=0x1000331c
281047   <= SYSCALL SUBSCRIBE driver=0 (timer) subdriver=0 cb=0x20032006 data=0x1000331c
[...]

PR #2454, opened 2021-02-26
Transitions Callback to Upcall for 2.0. In a seprate PR for review sanity.

phil-levis:
### Pull Request Overview

This pull request changes `kernel::callback::Callback` to `kernel::upcall::Upcall`. It updates the rest of the codebase to use this, and also tries to change all of the comments to distinguish general callbacks from upcalls.

Given that this touches so many files, I suggest that we merge this after the major code review for 2.0.

### Testing Strategy

### TODO or Help Wanted

I think the terminology is 95% right -- given the scope of these changes, I think that simply fixing the last few when we encounter them will be sufficient.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/litex/src/uart.rs
+++ b/chips/litex/src/uart.rs
@@ -237,7 +237,7 @@ impl<'a, R: LiteXSoCRegisterConfiguration> LiteXUart<'a, R> {
 
     // This is either called as a deferred call or by a
     // hardware-generated interrupt, hence it is guaranteed to be an
-    // upcall
+    // callback
     fn resume_tx(&self) {
         let len = self.tx_len.get();
         let mut progress = self.tx_progress.get();
@@ -344,12 +344,12 @@ impl<'a, R: LiteXSoCRegisterConfiguration> uart::Transmit<'a> for LiteXUart<'a, 
         //
         // If it does not fill the FIFO, an
         // interrupt will _not_ be generated and hence we have to
-        // perform the upcall using a deferred call.
+        // perform the callback using a deferred call.
         //
         // If we fill up the FIFO, an interrupt _will_ be
         // generated. We can transmit the rest using `resume_tx` and
         // directly call the callback there, as we are guaranteed to
-        // be in an upcall.
+        // be in a callback.
         let mut fifo_full: bool;
         let mut progress: usize = 0;
         while {
@@ -370,7 +370,7 @@ impl<'a, R: LiteXSoCRegisterConfiguration> uart::Transmit<'a> for LiteXUart<'a, 
         // _must_ have been written to the device
         //
         // In this case, we must request a deferred call for the
-        // upcall
+        // callback
         if !fifo_full {
             assert!(progress == tx_len);
 
--- a/doc/Syscalls.md
+++ b/doc/Syscalls.md
@@ -43,7 +43,7 @@ modes](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0553a/CHDI
 
 Second, context switching to the kernel allows it to do other resource handling
 before returning to the application. This could include running other
-applications, servicing queued callbacks, or many other activities. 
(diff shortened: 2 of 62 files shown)


chips/litex/src/uart.rs
```diff
@@ -344,12 +344,12 @@ impl<'a, R: LiteXSoCRegisterConfiguration> uart::Transmit<'a> for LiteXUart<'a,
...
         // If it does not fill the FIFO, an
         // interrupt will _not_ be generated and hence we have to
-        // perform the upcall using a deferred call.
+        // perform the callback using a deferred call.
         //
         // If we fill up the FIFO, an interrupt _will_ be
         // generated. We can transmit the rest using `resume_tx` and
         // directly call the callback there, as we are guaranteed to
-        // be in an upcall.
+        // be in a callback.
```
hudson-ayers, line 352: Shouldn't this stay as upcall? for tock's rules about scheduling callbacks, I think what matters is that the callback is only called from within an upcall

doc/Syscalls.md
```diff
@@ -61,13 +61,13 @@ In Tock, a process can be in one of seven states:
...
- in the Running state until it explicitly yields. Callbacks from other kernel
- operations are not delivered to Running processes (i.e. callbacks do not
+ in the Running state until it explicitly yields. Upcalls from other kernel
+ operations are not delivered to Running processes (i.e. upcalls do not
  interrupt processes), rather they are enqueued until the process yields.
  - **Yielded**: Suspended operation. A Yielded process will not be scheduled by
  Tock. Processes often yield while they are waiting for I/O or other operations
  to complete and have no immediately useful work to do. Whenever the kernel issues
- a callback to a Yielded process, the process is transitioned to the Running state.
+ a upcall to a Yielded process, the process is transitioned to the Running state.
```
hudson-ayers, line 70:
```suggestion
 an upcall to a Yielded process, the process is transitioned to the Running state.
```

doc/Syscalls.md
```diff
@@ -156,64 +156,70 @@ in the libtock-c syscall code for more information about Yield.
...
+causes the trap handler to execute. The trap handler checks
+`mscratch`, and if the value is nonzero then it contains the stack
+pointer of the kernel and this trap must have happened while the
+system was executing an application. Then, the kernel stack pointer
+from `mscratch` is used to find the pointer to the stored state
+struct, and all process registers are saved. The trap handler also
+saves the process PC from the `mepc` CSR and the `mcause` CSR. It then
+loads the kernel address of where to resume the context switching code
+to `mepc` and calls `mret` to exit the trap handler. Back in the
+context switching code, the kernel restores its registers from its
```
hudson-ayers, line 192: This whole diff seems unrelated to this PR?

hudson-ayers (approved):
LGTM, but a few comments

chips/litex/src/uart.rs
lschuermann: I think callback is better here, given that this references an "upcall" from the scheduler, right? That shouldn't be confused with an upcall to userspace. The important thing here is that we're in a different stack frame to be able to safely call `resume_tx`, such that it can enter it's resources again. This is actually the exact issue that confused me in the initial discussion for this in the first place. I suppose saying "upcall from the scheduler" could be a valid option? For reference, this is @alevy's comment which made me call this an "upcall": https://github.com/tock/tock/issues/1736#issuecomment-610015515 :smile:

lschuermann (commented):
This looks pretty good, apart from Hudson's comments! It needs a rebase though.

lschuermann (approved):

alistair23 (approved):

doc/Syscalls.md
phil-levis: It's just a formatting/line wrapping change.

Issue #2369, opened 2021-01-20
Process::memory violates aliasing requirements (undefined behavior)


jrvanwhy:
[`Process::memory`](https://github.com/tock/tock/blob/2a729631fca67e9eb7c1b6dac3ec0a89c9ff92b5/kernel/src/process.rs#L821) is a `&'static mut [u8]` that points to a process' memory region. Although Rust's memory model is not yet 100% formally specified, it seems very unlikely that holding a `&mut [u8]` reference into memory that is mutated by the userspace process is not undefined behavior. It is certainly undefined behavior under the stacked borrows model.

hudson-ayers:
Looking at this briefly, `Process::memory` is only ever used via calls to `self.memory.as_ptr()` and `self.memory.len()`, so it should be pretty straightforward to instead store a pointer and length (in fact, I wonder if we used to do exactly that).

bradjc:
To the untrained eye, storing a pointer and length is exactly what a slice does. Can we keep a reference to a non-mutable slice?

jrvanwhy:
> To the untrained eye, storing a pointer and length is exactly what a slice does. Can we keep a reference to a non-mutable slice?

They are structurally equivalent but not semantically equivalent.

If we store a non-`mut` slice, then we would generate undefined behavior when we produce a `&mut [u8]` as a result of an `allow` system call (I think the kernel does that). Under the stacked borrows model, if we never generate a `&mut [u8]` pointing into application memory, the app mutating that memory still causes undefined behavior.

Using a pointer + length does not carry the same semantics around mutation.

lschuermann:
Ha! Just wanted to open this issue. Agree with the proposed solution of storing a pointer and length. In general I feel we should really enforce only a single mutable slice reference to any process-available memory section existing in the kernel. Given a capsule can deference an AppSlice at any time, it's hard to uphold that guarantee if we also store slices in the kernel.

bradjc:
We also have a mutable slice in main.rs. Is that OK?

jrvanwhy:
> We also have a mutable slice in main.rs. Is that OK?

I believe that is problematic as well, as it could overlap with mutable references obtained via `AppSlice`.

That said, I don't expect it to result in bugs in the near future as long as the slice isn't used after processes start executing -- the compiler doesn't really have a reason to "notice" the presence of undefined behavior.

bradjc:
[...]

PR #2458, opened 2021-03-02
RFC: kernel: change `unsafe` handling

bradjc:
sched.rs and process.rs had several functions marked entirely `unsafe`, resulting in hundreds of lines of code where unsafe operations were allowed and no clear "ownership" of where the unsafe operations were actually being checked.

This PR:

1. Significantly reduces the LoC where `unsafe` is implicitly allowed, and 
2. Tries to clarify where we are doing checks to ensure the unsafe operations are OK.

RFC:

In Tock we try to define `unsafe` as strictly "memory unsafe", not "sensitive function that should be called carefully". I argue that the process.X functions that used to be marked `unsafe` are doing the required checks to ensure that they are memory safe, and therefore shouldn't be marked `unsafe`. I think I was the one who asked for `set_byte()` to be marked `unsafe`, but I think I was wrong about that, since the set_byte implementation does the required check.

Does this seem like the approach we want to take? Should we do something else?

### Testing Strategy

travis

### TODO or Help Wanted

Comments?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -451,7 +451,14 @@ pub trait ProcessType {
     /// `addr` to `value`. Return true if `addr` is within the RAM
     /// bounds currently exposed to the process (thereby writable
     /// by the process itself) and the value was set, false otherwise.
-    fn set_byte(&self, addr: *mut u8, value: u8) -> bool;
+    ///
+    /// ### Safety
+    ///
+    /// This function verifies that the byte to be written is in the process's
+    /// accessible memory. However, to avoid undefined behavior the caller needs
+    /// to ensure that no other references exist to the process's memory before
+    /// calling this function.
+    unsafe fn set_byte(&self, addr: *mut u8, value: u8) -> bool;
 
     /// Get the first address of process's flash that isn't protected by the
     /// kernel. The protected range of flash contains the TBF header and
@@ -1448,16 +1455,12 @@ impl<C: Chip> ProcessType for Process<'_, C> {
         }
     }
 
-    // This function verifies that the dereference is valid and safe.
-    #[allow(clippy::not_unsafe_ptr_arg_deref)]
-    fn set_byte(&self, addr: *mut u8, value: u8) -> bool {
+    unsafe fn set_byte(&self, addr: *mut u8, value: u8) -> bool {
         if self.in_app_owned_memory(addr, 1) {
-            // This is safe because we have verified that the address we are
-            // writing to is in fact within the process's process-accessible
-            // memory.
-            unsafe {
-                *addr = value;
-            }
+            // We verify that this will only write process-accessible memory,
+            // but this can still be undefined behavior if something else holds
+            // a reference to this memory.
+            *addr = value;
             true
         } else {
             false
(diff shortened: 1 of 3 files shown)


kernel/src/process.rs
```diff
@@ -1465,9 +1465,16 @@ impl<C: Chip> ProcessType for Process<'_, C> {
...
 
-    unsafe fn set_byte(&self, addr: *mut u8, value: u8) -> bool {
+    // This function verifies that the dereference is valid and safe.
+    #[allow(clippy::not_unsafe_ptr_arg_deref)]
+    fn set_byte(&self, addr: *mut u8, value: u8) -> bool {
         if self.in_app_owned_memory(addr, 1) {
-            *addr = value;
+            // This is safe because we have verified that the address we are
+            // writing to is in fact within the process's process-accessible
+            // memory.
```
hudson-ayers, line 1474: is this check really sufficient for memory safety? I think it is only sufficient if we can guarantee that no other reference to this byte exists anywhere in the kernel. Which I think currently is not true (https://github.com/tock/tock/issues/2369).

kernel/src/sched.rs
```diff
@@ -602,7 +602,8 @@ impl Kernel {
             }
 
             // Check if the scheduler wishes to continue running this process.
-            if !scheduler.continue_process(process.appid(), chip) {
+            let continue_process = unsafe { scheduler.continue_process(process.appid(), chip) };
```
hudson-ayers, line 605: This maybe should not be an unsafe function at all, to be consistent with the other changes here.

hudson-ayers (commented):
This generally seems like a step in the right direction to me. We may want to explore deeper into some of the functions that still require unsafe blocks here to make sure we have very precisely defined the requirements for how to safely call those functions as well.

hudson-ayers:
> In Tock we try to define unsafe as strictly "memory unsafe", not "sensitive function that should be called carefully". 

Really I think this is how Rust defines unsafe, so it is good we define it that way as well

kernel/src/process.rs
bradjc: On cortex-m every function in this PR writes to process memory eventually. We can take care of #2369 pretty quickly, right?

hudson-ayers: Yeah I think we can take care of #2369 pretty quickly. But I think we should clearly document that this function is only safe if no references to the byte being modified exist anywhere in the kernel. For example it might still be unsafe to call this function early in the board setup process when we still have references to process memory.

bradjc: If `set_byte` cannot verify that this is safe, and there are no "tock level" restrictions (convention or otherwise) that would guarantee this is safe, then I think we should leave this function marked `unsafe` because the caller then needs to do additional checks.

However, I would rather have rules around when a reference to process memory can exist than have to consider the implications of what if the reference does exist for every function in process and every use of those functions.

hudson-ayers: Yes, I think having rules around when a reference to process memory can exist would be a good solution here.

kernel/src/sched.rs
bradjc: Right, that should be a separate PR.

bradjc:
Any other thoughts?

ppannuto (approved):
I think this is a good set of changes, and generally always supportive of scoping down use of `unsafe`.

hudson-ayers (commented):
This looks good except that I still think we should leave `set_byte()` as `unsafe`

bradjc:
I reverted the set_byte change, but added more clarification on what that `unsafe` means.

hudson-ayers (approved):

Issue #2222, opened 2020-11-25
RAM initialization code violates pointer provenance


jamesmunns:
RAM initialization currently passes pointers to these statics:

https://github.com/tock/tock/blob/master/chips/stm32f4xx/src/lib.rs#L65-L79

to `tock-rt0`:

https://github.com/tock/tock/blob/master/libraries/tock-rt0/src/lib.rs#L14-L34

But since r0 will fill the entire memory range between the pointers, it ends up calling offset with values that create pointers well beyond the u32 the static is declared to contain. According to the [docs of `offset`](https://doc.rust-lang.org/std/primitive.pointer.html#method.offset), this is 

> Both the starting and resulting pointer must be either in bounds or one byte past the end of the same allocated object. Note that in Rust, every (stack-allocated) variable is considered a separate allocated object.

(every `static` also counts as its own "allocated object", I believe)

Moreover, the `ptr::write` and `ptr::write_volatile` calls in r0 also cause UB, because this invariant is 

> dst must be valid for writes.

Where the [validity](https://doc.rust-lang.org/std/ptr/index.html#safety) of a pointer requires this:

> For a pointer to be valid, it is necessary, but not always sufficient, that the pointer be dereferenceable: the memory range of the given size starting at the pointer must all be within the bounds of a single allocated object. Note that in Rust, every (stack-allocated) variable is considered a separate allocated object.

...but again, we are crossing between potentially many allocated objects (every static in the .data/.bss section).

Furthermore, writing to a static through a pointer not derived from that static might violate aliasing rules, but it isn't yet clear if this is the case today (I've asked the folks working on the unsafe code guidelines for clarification / guidance).

The most robust solution for this issue is to write the RAM init code in assembly instead of Rust, and run no Rust code at all until RAM is initialized.

(this issue was recently discovered to affect most -rt crates in the ecosystem; we do not believe it to cause practical issues at the moment)

jrvanwhy:
> Furthermore, writing to a static through a pointer not derived from that static might violate aliasing rules, but it isn't yet clear if this is the case today (I've asked the folks working on the unsafe code guidelines for clarification / guidance).

Do you have a link to this discussion, or is it a private communication?

`libtock-rs`' runtime performs memory setup by casting `usize` values into `*mut u8` headers, then writing the memory using `core::ptr::{copy_nonoverlapping, write_bytes}`. I think that whether `libtock-rs`' runtime is sound depends on the answer to your question.

jamesmunns:
[...]

PR #2446, opened 2021-02-19
Tock 2.0 Alpha1

bradjc:
### Pull Request Overview

This pull request updates the Tock kernel to the 2.0 kernel<->userspace interface.

The current plan is to merge this on Monday, March 22, once @lschuermann has cleaned up the commit log.

**This is an API breaking change!**

Merging this PR will break the Tock 1.0 agreement with userspace. When this PR is merged, Tock is in "release mode" towards 2.0. The major version number change signifies the breaking change for userspace. We expect that merging this PR will create ~~v2.0-rc1~~ v2.0-alpha1, and further development towards v2.0 will occur on the master branch until the release.

However, merging this PR only signifies breaking the 1.0 API, it does not imply the 2.0 interface is stabilized. We expect more development before settling on the final 2.0 API.

Here are the three major items that we have to do before merging the `tock-2.0-dev` branch into master: 
1. [x] Completely wrap up the system call API/ABI updates (I think this is all that's left: https://github.com/tock/libtock-c/pull/177; we might need a follow-up one for TOCK_EBADRVAL, but this is not blocking)
2. [x] Update documentation, so it is consistent with the code: Markdown files (Syscalls.md, Userland.md)
3. [x] Update documentation, so it is consistent with the code: the Tock book https://github.com/tock/book/pull/13
4. [x] Update documentation, so it is consistent with the code: courses #2455 
5. [x] Overall code review (this PR)
6. [x] Change Callback to Upcall #2454 
7. [x] Change GenericSyscallReturnValue to SyscallReturn
8. [x] Clean up commit log (@lschuermann )
9. [x] libtock-c testing (@phil-levis ), https://github.com/tock/libtock-c/pull/195

Then, before we can tag an RC1, we need to complete everything in #2429.

### Merge Sign-Off

This is a big change, and API breaking. It would be good to get consensus. Can you please check your name when you are ready for this PR to be merged? Doing it this way 1) allows me to participate and 2) doesn't change as new commits are added.

- ~[ ] Niklas Adolfsson, @niklasad1~ (#2495)
- [x] Hudson Ayers, @hudson-ayers
- [x] Brad Campbell, @bradjc
- [x] Branden Ghena, @brghena
- [x] Philip Levis, @phil-levis
- [x] Amit Levy, @alevy
- [x] Pat Pannuto, @ppannuto 
- [x] Johnathan Van Why, @jrvanwhy

### Testing Strategy

[...]

PR #2461, opened 2021-03-03
Implement RT0 in assembly

bradjc:
### Pull Request Overview

According to #2222 we shouldn't try to implement RAM init functions in Rust. This PR (starts to) implement rt0 in assembly for cortex-m and rv32.

I'd like to make sure we are happy with this and the names before switching boards and chips.

### Testing Strategy

in progress

### TODO or Help Wanted

Thoughts?

Also it would be great to have more eyes on my assembly.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


boards/hail/src/main.rs
```diff
@@ -169,14 +169,12 @@ unsafe fn set_pin_primary_functions(peripherals: &Sam4lDefaultPeripherals) {
...
 ///
-/// This symbol is loaded into vector table by the SAM4L chip crate.
-/// When the chip first powers on or later does a hard reset, after the core
-/// initializes all the hardware, the address of this function is loaded and
-/// execution begins here.
+/// This is called from the reset handler after memory initialization is
+/// complete.
 #[no_mangle]
-pub unsafe fn reset_handler() {
+pub unsafe fn main() {
```
ppannuto, line 177:
```suggestion
pub unsafe fn rust_entry_point() {
```

Maybe this is a more direct name. `main` is a bit overloaded in what it means, but I believe the term "entry point" to be fairly standardized as the 'first code that is run', so calling this the "rust entry point" is perhaps a more clear term for folks reading this?

ppannuto (commented):
I reviewed the ARM assembly pretty carefully, and it looks correct. I also double checked that all of the instructions used are valid in v6m, so this should work without modification if one of the Cortex M0-based platforms comes back into tree (interestingly, `STM <rn>!, {regs}` is valid v6m, while `STM <rn>, {regs}` is not; i.e. writeback is required)

boards/hail/src/main.rs
bradjc: I thought main was pretty clear, since it's where all of the interesting work starts but is not the _first_ thing that runs.

`rust_entry_point` is more grepable, however.

hudson-ayers:
This generally looks good, though it would be nice to have some confirmation that using a naked function containing only a single inline asm block is a safe use of "rust" before memory is initialized. I only say this because I noticed that the fix to https://github.com/rust-embedded/cortex-m-rt/issues/300 used external asm, though it is possible that is just because naked_fn is unstable.

From trying to follow https://rust-lang.zulipchat.com/#narrow/stream/136281-t-lang.2Fwg-unsafe-code-guidelines/topic/The.20least.20incorrect.20init.20code.20.3A) I think that a naked inline asm function should be fine...but this is all a bit beyond me. @jonas-schievink any chance you know if this is safe? Or as safe as using external asm?

jonas-schievink:
I think that's fine, yes.

jonas-schievink:
(as safe as external asm, that is)

alistair23 (approved):
I think this looks good. I like using `main()` as I think that is what people expect to run after some setup.

lschuermann:
These changes look good. I'm certainly no expert on ARM assembly but the
RISC-V changes look correct to me. I think `main` is fine. A grep for
"fn main()" shouldn't yield too many results. :)

alistair23 (dismissed):

bradjc:
Ok I think this is now everything that needs to be changed to do this. Recent updates:

- Updated all boards to `main()` rather than `reset_handler()`.
- Changed references to `reset_handler` to `main`.
- Removed libraries/tock-rt0

ppannuto (approved):

bradjc:
@tock/core-wg This does change the boot process for all tock boards, and is probably worth a quick thought.

phil-levis (dismissed):

hudson-ayers:
Given the 72 files changed, I would probably recommend merging this after the 2.0 alpha 1 PR.

alistair23 (dismissed):

alevy (dismissed):
I hate that we have to do this, but it seems like we have to do this. Thanks @bradjc

lschuermann (approved):

alevy:
are we blocking on #2446 for merging this?

hudson-ayers:
yes I think so

bradjc:
Rebased and updated the new board.

hudson-ayers (approved):

alistair23 (approved):

ppannuto (approved):
bors r+

PR #2462, opened 2021-03-03
Tock 2.0: implement Callback swapping restrictions (v3)

lschuermann:
### Pull Request Overview

This pull request is a follow-up to #2282 and #2445. This was done jointly together with @hudson-ayers. Thanks Hudson!

#### Issue

Essentially, the protections introduced in this PR are required because in the Tock 2.0 system call architecture, with every call to `subscribe`, the previous `Callback` provided by the application must be returned. If we were to not enforce any constraints here, this would have serious side effects:

- a `Callback` subscribed by process A could be handed back to process B.

  This leaks both the callback pointer and the `appdata` field of the `Callback` to another application.

- a `Callback` from one capsule could be given to another capsule and returned as part of a call to subscribe there.
- a `Callback`, passed to a driver under subscribe (subdriver) number `x` could be returned as part of a call to `subscribe` to the same driver under subscribe (subdriver) number `y`, where `x != y`.
- a capsule could pass back a _null callback_ (with callback pointer and appdata being 0), where a process has actually subscribed a `Callback` with non-null values before.

  All of these cases could cause inconsistencies and undefined behavior in userspace, if a process were to rely on the fact that the returned `Callback` is the one previously passed to the capsule.

  It was indicated that userspace libraries might want to rely on this behaviour for stack-allocated callbacks (especially in libtock-rs).

#### Guarantees made by this PR 

This PR enables the kernel to guarantee:

- a capsule cannot return a `Callback` originally passed to a different capsule in a `subscribe` invocation
- a capsule cannot return, as a result of a `subscribe` invocation, a callback which was provided by a different process
- a capsule cannot return, as a result of a `subscribe` invocation, a callback which was provided by the same process under a different subscribe / subdriver number

#### Approach

This PR changes Grants to use a `GrantDefault` trait for the default initialization of contained structures, where currently `Default` is used. Using a custom trait here allows us to pass in futher data into the default implementation of said contained structures.

One of the things passed into the `grant_default` constructor is a `ProcessCallbackFactory`, which can be used to obtain `Callback`s that are always bound to a specific `(driver, subdriver, process)` combination. Furthermore, only one Callback for a given `(driver, subdriver, process)` combination may exist in the kernel at any given time, except when a subscribe system call is invoked (to swap out one `Callback` for another).

This approach requires making sure that each driver can only hold a single grant region, to ever only get a `ProcessCallbackFactory` once per process.

#### Implications

- This PR removes `Copy, Clone` trait implementations from `Callbacks`. I don't see any solution which wouldn't heavily modify larger parts of the OS to get around this. Most approaches I've come up with use the fact that through ownership and move semantics, we can control & limit the lifetime of `Callback`s effectively and at compile time.

- This implementation strategy, in contrast to some others, does not require any central registry or other data structures. By tightly controlling the instances of the `Callback` type at compile time, we can uphold the desired guarantees without using additional data fields. There is one exception to this -- we now track the correspondence of `grant_num` --> `driver_num`, which requires an array for this tracking. Currently, we set this array to be of size `MAX_DRIVERS`, where `MAX_DRIVERS` is a field in the kernel `CONFIG` struct. This could be moved into `main.rs` for every board to allow boards to set `MAX_DRIVERS` independently, but we wanted to keep the diff as small as possible here.

- This introduces a procedural macro, to be able to `#[derive(GrantDefault)]`. This is optional and can be removed easily, at the cost of a little more verbosity in capsules (each capsule will have to manually derive GrantDefault, even if before it used `#[derive(Default)]`. Using procedural macros pulls in the external crates `proc-macro`, `syn` and `proc-macro2`.

- Non-virtualized capsules need to be changed to work with this PR. Because `Callback`s are a process resource, they need to be kept in their respective Grant region. This also prevents accidentally leaking a `Callback` of one process to another, which many non-virtualized capsules do as of their current implementation. Instead, they should store which process is currently active and using them.

### Testing Strategy

This pull request was tested by trying to swap callbacks on a Hail board using an example capsule. It need more testing.

### TODO or Help Wanted

This pull request has the required changes to all capsules except those capsules which are not virtualized across multiple apps (and thus do not have grant regions). We will update those capsules to use grant regions to store callbacks if the general design in this PR is approved.

Given that these updates might require to touch the logic of the respective capsules, they are carried out in individual PRs. If you want to change a driver to use Grants as done in #2521, please add yourself to this list:
- [X] `src/analog_comparator.rs` (@bradjc, #2521) 
- [x] `src/max17205.rs` (@hudson-ayers, #2553)
- [x] `src/spi_peripheral.rs` (@lschuermann, #2559)
- [x] `src/pca9544a.rs` (@hudson-ayers, #2554)
- [x] `src/l3gd20.rs` (@phil-levis, #2572)
- [x] `src/lsm303dlhc.rs` (@hudson-ayers @lschuermann, #2518)
- [x] `src/i2c_master_slave_driver.rs`(@hudson-ayers, #2547)
- [x] `src/mlx90614.rs` (@hudson-ayers, #2555)
- [x] `src/sdcard.rs` (@lschuermann, #2560)
- [x] `src/lsm303agr.rs` (@phil-levis, #2574)
- [x] `src/lps25hb.rs` (@hudson-ayers, #2556)
- [x] `src/gpio_async.rs` (@ppannuto, #2562)
- [x] `src/ltc294x.rs` (@phil-levis, #2575)
- [x] `src/spi_controller.rs` (@lschuermann, #2558)
- [x] `src/tsl2561.rs` (@hudson-ayers, #2557)

(list obtained through `grep -rL "Grant" | xargs grep -rl "Driver" | xargs grep -rl "Upcall\|AppSlice"`)

This PR mostly panics when capsules misbehave, but other approaches could be imagined in the future to block those capsules indefinitely without faulting the kernel.

### Documentation Updated

- [ ] TODO.

### Formatting

- [x] Ran `make prepush`.


hudson-ayers:
As mentioned by Leon, I worked with him on this updated design and am pretty happy with how it turned out. The diff may seem large, but it is mostly one-time changes to capsules and board files. To initially review it, I recommend looking at one changed capsule, one changed main.rs, and one changed component, and then looking at the changes in the kernel crate.

I know one question on #2445 was to clearly explain how these changes enable the kernel to enforce these restrictions. To summarize, this PR achieves this by making it so that the `Callback` struct can only be created by the kernel as part of grant initialization. Further, this PR makes it so that `Callback`s created during grant initialization can only be created using a `ProcessCallbackFactory` which enforces that for a given `driver_num`, at most one `Callback` can exist for a given `subdriver_num`. Thus, combined with the removal of `Copy` and `Clone`, no capsule can hold two `Callback`s with the same `driver_num` and `subdriver_num`. Therefore, it is trivial for the kernel to confirm that when a newly constructed `Callback` is passed in for a given `subdriver_num`, the returned `Callback` is the `Callback` userspace expects to receive.

bradjc:
1. So this doesn't prevent a capsule from returning the wrong callback, but rather it makes it possible for the core kernel to check that the proper callback was returned?
2. Why is a capsule with two grants problematic? Is Kernel.grant_num_mapping only to prevent multiple grants in one capsule?
3. In my opinion needing the external macro crates is problematic.

lschuermann:
> 1. So this doesn't prevent a capsule from returning the wrong callback, but rather it makes it possible for the core kernel to check that the proper callback was returned?

Yes. We thought about ways to do the entire check statically (e.g. through const generics, etc.) but ultimately couldn't find a way to make that work. The compiler can only work with the type information it has, so pretty much every approach relying on this would generate a bunch of types at compile time, require use of generics and monomorphisation would increase code size.

> 3. In my opinion needing the external macro crates is problematic.

That's unfortunate, though not an issue. The only crate _absolutely required_ to build macros is `proc-macro`, which is built by the Rust compiler team (the compiler essentially "links" against that interface defined there, as procedural macros are extensions to the compiler). `syn` and `proc-macro2` make build macros actually feasible and are widely used throughout the entire Rust ecosystem.

If we decide against including these crates, we can just throw the macros out. It's going to get a bit more verbose then, but it's no big deal.

bradjc:
I'm going to try a different take on why we do not need to check for duplicate grants.

Capsules do not allocate grants. Any capsule can require 0 or more grants and use them however the capsule sees fit. A capsule cannot, however, create its own grant region. Because we require a capability for grant creation, only the board's crate can create a grant, and then decide whether to give it to the capsule (or not). Therefore, it is not actually the fault of the capsule if system threat model guarantees are violated if it needs two grants, but rather it is the fault of the board, which is trusted code. As such, we do _not_ need to verify that a capsule does not use multiple grant regions, but rather we need to verify that a board does not give a capsule multiple grant regions. Since that requires a capability, we expect that to happen during code review already.

hudson-ayers:
I find that argument pretty convincing actually. I didn't consider that creating two grants would consider passing both in in main.rs. Without this check, however, we will also need to verify that no two drivers create a grant using the same passed driver_num. Which is probably much easier to miss in code review.

bradjc:
Why do we need `driver_num` included at all? The `(process, subdriver_num)` tuple would not be globally unique. But how could a capsule get a callback from a different grant region?

If we don't need driver_num to check correctness, we would only need it for CallbackId. Do we need the driver number in Callback?

lschuermann:
@bradjc Supposedly, two drivers could accidentally or deliberately create a channel over which the `Callback` could be exchanged, which would then result in not being able to guarantee correctness of the returned callback in every case. In practice I find that to be somewhat unlikely, but given two mischievous capsules -- in general -- possible. The threat model is however pretty clear that capsule code is untrusted in this regard.

bradjc:
> @bradjc Supposedly, two drivers could accidentally or deliberately create a channel over which the `Callback` could be exchanged, which would then result in not being able to guarantee correctness of the returned callback in every case. In practice I find that to be somewhat unlikely, but given two mischievous capsules -- in general -- possible. The threat model is however pretty clear that capsule code is untrusted in this regard.

How would they get a reference to each other, or otherwise create the channel? I think that would have to come from main.rs, so using my logic from above, that is a trusted board code bug, not a capsule threat model violation.

bradjc:
@alevy @jrvanwhy thoughts?

hudson-ayers:
> How would they get a reference to each other, or otherwise create the channel? I think that would have to come from main.rs, so using my logic from above, that is a trusted board code bug, not a capsule threat model violation.

I am not sure about this interpretation. In this case the "trusted board code bug" is that the board code gave two capsules a reference to one another. But we do this all the time with virtualizers! I guess you could take a more strict interpretation and say that the bug is giving two userspace drivers a reference to one another, but it is not immediately clear why this should be forbidden, or how one would easily ensure this doesn't happen. If the bug is "gave two capsules a reference to each other in a way that makes it possible for them to share a callback", that means that auditing board code to ensure userspace is sound requires analyzing the code of the capsules being used, which is explicitly something the threat model wants to avoid having to guarantee. 

Say I am writing a board and I find two Tock capsules posted online somewhere - lets say, an 802.15.4_driver and an aes_driver. 15.4 uses AES encrpytion for frames, so it makes sense they are released by the same author.
I include them both in my main.rs using the provided component that I also found online. That components gives the two drivers a reference to each other, presumably so the 15.4 driver can easily use some of the code in the AES driver when it needs to encrypt frames. But secretly, the capsules reference each other in such a way that they are able to swap callbacks (maybe one is allowed to execute a closure on the other, or something). Now, by simply including these drivers I found online in my board, my entire userspace runtime is unsound because of callback swapping that the kernel promised would not happen. My understanding is that we want to avoid board authors having to manually audit capsules to verify userspace soundness.

jrvanwhy:
+1 to Hudson's explanation. My shortened version is that yes we technically could make it the responsibility of the board main file, but in practice it would be easy to make mistakes in the board main file if we do so.

bradjc:
OK next attempt.

What about having callback store a `Option<DriverNum>`? The default callback created at initialization will have `None`. But, any callback passed in will have `Some(DriverNum)`. If the same <process, driver, subscribe> is called again, then the `DriverNum` has to match. For the first call to <process, driver, subscribe>, the driver number would be None and that would just match.

hudson-ayers:
> OK next attempt.
> 
> What about having callback store a `Option<DriverNum>`? The default callback created at initialization will have `None`. But, any callback passed in will have `Some(DriverNum)`. If the same <process, driver, subscribe> is called again, then the `DriverNum` has to match. For the first call to <process, driver, subscribe>, the driver number would be None and that would just match.

That is pretty close to Leon's first approach. The problem is there is no way for the kernel to know the same <process, driver, subscribe> has been called before. Even if it is only possible for trusted kernel code to modify the field containing `Option<DriverNum>`, consider this scenario:

Capsule A and Capsule B each have 2 callbacks (subdriver num 0 and 1 for both). Thus there are 4 callbacks total: A0, A1, B0, B1. These capsules have a reference to each other and cooperate maliciously with the goal of violating the kernel guarantees provided to userspace. At boot the callbacks look like this:

- None / 0
- None / 1
- None / 0
- None / 1

the app calls subscribe (A0) and the capsules behave as expected:

- Some(A) / 0
- None / 1
- None / 0
- None / 1

the app calls subscribe (A0) again, but capsule A communicates with capsule B and returns the None / 0 callback that B still holds. The kernel is none the wiser, because it does not remember that A0 was already called. But now capsule A has held onto a callback longer than allowed.

bradjc:
OK, what about deciding on a policy that two grants for the same process cannot be entered simultaneously? I'm fairly sure this doesn't happen in any existing capsules. This might also eliminate the concern about a capsule having multiple grants.

I've come up with one strawman example of why one capsule might need to interact with another capsule's grants. Imagine a SynchronousSensor capsule that holds references to a couple other sensors and samples them simultaneously. Each sensor driver when the measurement finishes sends the result back to the process via upcall. The SynchronousSensor capsule then needs to mark in the grant region for each sensor capsule that this process wants the result of the current measurement.

However, I don't think the SynchronousSensor capsule needs to enter the grants _simultaneously_. Rather, it just needs to enter the grant of each connected sensor and set the flag.

This policy need not be promised for stability. It would just be a mechanism to enforce the callback swapping rules, and should a better callback swapping prevention implementation be discovered the rule could be removed (just like the one-grant-per-Driver policy proposed here).

And, I think this can be implemented with an additional check in grant.enter() to make sure no other grants are entered for this process.

hudson-ayers:
> OK, what about deciding on a policy that two grants for the same process cannot be entered simultaneously? I'm fairly sure this doesn't happen in any existing capsules. This might also eliminate the concern about a capsule having multiple grants.
> 
> I've come up with one strawman example of why one capsule might need to interact with another capsule's grants. Imagine a SynchronousSensor capsule that holds references to a couple other sensors and samples them simultaneously. Each sensor driver when the measurement finishes sends the result back to the process via upcall. The SynchronousSensor capsule then needs to mark in the grant region for each sensor capsule that this process wants the result of the current measurement.
> 
> However, I don't think the SynchronousSensor capsule needs to enter the grants _simultaneously_. Rather, it just needs to enter the grant of each connected sensor and set the flag.
> 
> This policy need not be promised for stability. It would just be a mechanism to enforce the callback swapping rules, and should a better callback swapping prevention implementation be discovered the rule could be removed (just like the one-grant-per-Driver policy proposed here).
> 
> And, I think this can be implemented with an additional check in grant.enter() to make sure no other grants are entered for this process.

I think the above attack would still work even with this restriction. The second callback swap only requires entering the grant for capsule B, not entering both at once.

bradjc:
The issue is that the `Callback` provided in the `subscribe()` arguments can be used as a holding spot to facilitate arbitrary callback swapping? In that case, it seems like any solution will require an additional identifier in the Callback object itself.

hudson-ayers:
Yeah I think so. Happy to take any ideas about better ways for the kernel to map `grant_num`s to `driver_num`s that require less changes in `main.rs`. For example, I think if the kernel had a way to take a `driver_num` and get the `grant_num` associated with that driver, that would be enough. But I can't think of a good way to do this without either extending the `Driver` trait or doing what we do here.

lschuermann:
Working on rebasing this, ETA 1-2 days.

hudson-ayers:
I merged in the master branch, and I believe this is now ready for review. There were *a lot* of changes in the kernel crate between when this was first submitted and now, so those files could definitely use a second set of eyes to be sure the merge didn't mess anything up.

capsules/src/console.rs
```diff
@@ -63,6 +64,21 @@ pub struct App {
     read_len: usize,
 }
 
+impl GrantDefault for App {
+    fn grant_default(_process_id: ProcessId, cb_factory: &mut ProcessUpcallFactory) -> Self {
+        App {
+            write_callback: cb_factory.build_upcall(1).unwrap(),
```
alistair23, line 70: Shouldn't this start at 0?

capsules/src/i2c_master.rs
```diff
@@ -4,19 +4,28 @@ use enum_primitive::enum_from_primitive;
...
-#[derive(Default)]
 pub struct App {
     callback: Upcall,
     slice: ReadWriteAppSlice,
 }
 
+impl GrantDefault for App {
+    fn grant_default(_process_id: ProcessId, cb_factory: &mut ProcessUpcallFactory) -> Self {
+        App {
+            callback: cb_factory.build_upcall(1).unwrap(),
```
alistair23, line 23: Same here?

boards/acd52832/src/main.rs
```diff
@@ -201,6 +202,7 @@ pub unsafe fn main() {
     //
     let button = components::button::ButtonComponent::new(
         board_kernel,
+        capsules::button::DRIVER_NUM as u32,
```
bradjc, line 205: Can we just decide on if these are u32 or usize? Then we don't need to have `as u32` everywhere.

bradjc (commented):
I have two things:
1. Do we need the kernel to check for double grants?
2. Why is Process involved with handling subscribe calls now?

Otherwise looks good.

kernel/src/sched.rs
```diff
@@ -181,6 +181,7 @@ impl Kernel {
             process_identifier_max: Cell::new(0),
             grant_counter: Cell::new(0),
             grants_finalized: Cell::new(false),
+            grant_num_mapping: [None; config::CONFIG.max_drivers],
```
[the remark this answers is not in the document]
hudson-ayers: We could get rid of this check. If we do so, it makes it possible for a board to introduce UB in userspace if a capsule creates 2 grants, and is intentionally written so as to maliciously swap callbacks stored in those two grants.

I don't think there is any other potential attack vector enabled by removing this check -- @lschuermann do you agree?

I am fine with removing the check. If we do remove the check, I wonder if we should mark `create_grant()` as `unsafe` -- misusing it cannot create unsoundness in the kernel, but could lead to it in the rust userspace. I suppose the answer is still probably "no".

hudson-ayers:
> I have two things:
> 
> 1. Do we need the kernel to check for double grants?

I responded directly on your comment on the file.

> 2. Why is Process involved with handling subscribe calls now?

@lschuermann I think I recall you making/being in favor of that change, but I can't remember the justification

capsules/src/console.rs
hudson-ayers: These values correctly represent the `subdriver_num` associated with each of those callbacks today. This PR does not require that subscribe numbers start at 0, only that they be declared in increasing order in this initialization function.

capsules/src/i2c_master.rs
hudson-ayers: See other comment, this value is correct

boards/acd52832/src/main.rs
hudson-ayers: I feel like there was already some discussion on this based on the possibility of one day supporting 64-bit platforms, but I do not know what came of it

bradjc: I recall deciding that there are not going to be >= 2^32+1 drivers, so we can always say the driver num is 32 bits. But I don't know that for sure.

lschuermann:
Sorry, just catching up on this.

> @lschuermann I think I recall you making/being in favor of that change, but I can't remember the justification

`process.rs` being involved in this is an old artifact and should be changed. This mirrored the way subscribe/allow were implemented on `tock-2.0-dev` temporarily (I've been working on this PR since December 2020 :smile:), but was ultimately changed prior to being merged into master. Will change / refactor this tomorrow.

Given that this had master merged in, I'd rather not do a rebase but just add a commit on top fixing this.

boards/acd52832/src/main.rs
lschuermann: > I recall deciding that there are not going to be >= 2^32+1 drivers, so we can always say the driver num is 32 bits. But I don't know that for sure.

I can confirm that. I have this on my todo list, but would rather not try to sneak this change in here (at least to existing code) as I feel it justifies a separate PR. That said, the new interfaces use `u32`s already, hence these conversions.

kernel/src/process.rs
```diff
@@ -313,6 +313,51 @@ pub trait Process {
 
     // additional memop like functions
 
+    /// Process a `Subscribe` request for a given process.
```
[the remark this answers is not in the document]
lschuermann: Will change, see https://github.com/tock/tock/pull/2462#issuecomment-844443207.

kernel/src/sched.rs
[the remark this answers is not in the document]
lschuermann: This topic seems sufficiently complex to contain hidden 

> I don't think there is any other potential attack vector enabled by removing this check

I'd carefully agree with this, but the interplay of all parts involved makes this rather hard to be absolutely sure we're not creating a loophole here :smile:.

> If we do remove the check, I wonder if we should mark `create_grant()` as `unsafe` -- misusing it cannot create unsoundness in the kernel, but could lead to it in the rust userspace. I suppose the answer is still probably "no".

I don't think unsafe applies here, as it's not behavior Rust considers to be unsafe, at least not with respect to the current application (i.e. the kernel). I think the kernel giving guarantees about certain behavior to userspace is rather related to semantically correct kernel code.

hudson-ayers:
commit on top sounds great, sorry for merging instead of rebasing

boards/acd52832/src/main.rs
hudson-ayers: I also would prefer not to block this PR on that change. It's been open for months and handling collisions with other PRs is difficult enough as-is.

kernel/src/sched.rs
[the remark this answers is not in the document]
hudson-ayers: @jrvanwhy do you have any thoughts on this?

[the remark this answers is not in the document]
hudson-ayers: After some more discussion with Brad, the attack vector of "one capsule being initialized with the same driver number as another capsule" still exists, if both capsules collude, and one of the capsules is never called. This is a more subtle, harder to catch error, so I am in favor of keeping the check.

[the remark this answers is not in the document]
lschuermann: After I quick discussion with @hudson-ayers, I still think this check could be useful but
- we don't create unsoundness in the kernel by not checking this
- given I'm more worried about out-of-tree boards, assuming the in-tree boards don't assign multiple grants per driver or mess this up another way, we'd still check that there are no accidental `Upcall` swaps by having this mechanism (aka. validate in-tree capsule correctness). That doesn't hold for out-of-tree capsules though.

However, @hudson-ayers' points are valid, so in general I'm indifferent about this. I do think that we could remove this check if it turns out to hurt performance or ergonomics significantly.

lschuermann:
Hudson Ayers ***@***.***> writes:
> commit on top sounds great, sorry for merging instead of rebasing

Should be done. The local diff looks fine, but I'm on 2G/GPRS internet
so confirming on GitHub is difficult. :)

It introduces a few new u32 -> usize casts for driver_num and
subdriver_num. I can create an RFC PR to discuss and ultimately change
the integer widths to be consistent across the codebase after this is
merged.

kernel/src/sched.rs
```diff
@@ -881,20 +883,99 @@ impl Kernel {
...
+
+                                    // TODO: The capsule might have already scheduled
+                                    // upcalls on the new instance, and we must not
+                                    // clear upcalls on the previous instance in
+                                    // this branch (capsule refused the subscribe
+                                    // operation). However, if the two upcalls are
+                                    // identical, we can't distinguish the two and
+                                    // will cancel upcalls on _both_ the previous
+                                    // and the new instance.
+                                    process.remove_pending_upcalls(returned_upcall.upcall_id);
```
hudson-ayers, line 940: I will take a stance on this and say that I think it is fine if we clear upcalls on the previous instance in this branch in the case where the rejected upcall was identical to an already held upcall. If a process calls `subscribe()` for some `(driver_num, subscribe_num)` and the capsule rejects, I think it is reasonable for the app to expect no future upcalls associated with that tuple. Can we just edit the subscribe TRD to state as such?

lschuermann: I'm still somewhat convinced we can work around that, without too many hacks and good ergonomics. There are some TODOs in there still, for instance currently a capsule can reject subscribe returning the old upcall, and accept subscribe returning the new upcall -- both of these should be prohibited. By some clever type wrapping the passed upcall and enforcing these invariants by wrapping the `swap` method, we can effectively prevent all of this wrong behavior. Can we give this a shot first, and not immediately declare this as "works as intended"? We should be able to quickly decide on either approach once this is merged.

boards/acd52832/src/main.rs
bradjc: Then can we move the cast and have create grant take `usize`? If we don't do it now, it might not ever happen.

kernel/src/sched.rs
```diff
@@ -146,6 +146,8 @@ pub struct Kernel {
     /// created and the data structures for grants have already been
     /// established.
     grants_finalized: Cell<bool>,
+
+    grant_num_mapping: [Option<u32>; config::CONFIG.max_drivers],
```
bradjc, line 150: doc

kernel/src/sched.rs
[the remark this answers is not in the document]
bradjc: I've changed my mind, because if two capsules set `DRIVER_NUM` in their files to the same number, and a board blindly uses it, then they could collude to return bad upcalls. This is still a board error, but would be a pretty subtle error. So I think it's better to have an explicit check.

kernel/src/sched.rs
hudson-ayers: Ok, I am fine to leave this as a TODO. But is it a blocking item for 2.0?

lschuermann: It's supposedly a quick and small change, and I think the implications of upcalls being cleared if subscribe is accepted, but the with the new upcall returned can lead to inconsistencies in user space, if not unsoundness. So I think it's important that we enforce that the subscribe return values are actually consistent.

boards/acd52832/src/main.rs
hudson-ayers: Sure, but it means all the components that take the value as a u32 have to be changed to take a usize, only to be changed back in the next PR. I went ahead and did this though.

kernel/src/sched.rs
hudson-ayers: done

jrvanwhy:
I haven't had the time to do a detailed review of the PR, but I looked hard enough to convince myself the overall design is sound.

kernel/src/sched.rs
```diff
@@ -384,18 +391,33 @@ impl Kernel {
...
+        let driver_num = driver_num as u32;
         if self.grants_finalized.get() {
             panic!("Grants finalized. Cannot create a new grant.");
         }
 
         // Create and return a new grant.
         let grant_index = self.grant_counter.get();
         self.grant_counter.increment();
-        Grant::new(self, grant_index)
+        if self.grant_num_mapping[grant_index].is_none() {
```
bradjc, line 408:
```suggestion
        // Verify that this new grant will have a unique `driver_num`.
        // We must ensure that two grants do not have the same `driver_num`
        // to correctly implement the kernel's promised upcall subscribe semantics.
        if self.grant_num_mapping[grant_index].is_none() {
```

kernel/src/sched.rs
```diff
@@ -867,37 +889,99 @@ impl Kernel {
...
+                                    || returned_upcall.appdata != appdata
+                                    || returned_upcall.fn_ptr != fn_ptr
+                                {
+                                    // The capsule did not return the Upcall passed in
+                                    //
+                                    // TODO: How to handle this?
+                                    panic!(
+                                        "Driver {}, subscribe num {}: Upcall swapped in error case",
+                                        driver_number, subdriver_number
+                                    );
```
bradjc, line 934: This seems like an important todo. It's a shame if after all this work buggy capsules can still cause kernel panics.

bradjc (changes requested):
I am pretty sold except for the new `panic!()`s in the kernel. Generally we say panics are ok because they either happen immediately or on some obscure hw issue. But these panics seems like they could happen at any time due to bad capsule logic.

If these panics are only supposed to happen immediately when a capsule is first used, then couldn't we skip all of the work in this PR and just test each capsule statically (ie call all of the subscribe calls and verify the correct callback is returned), likely before a release?

bradjc:
One outcome of discussions: terminating the process that called subscribe where the capsule violated the upcall swap semantics is still not the best solution, but I would prefer that to adding a panic.

bradjc:
And a more radical idea: we get rid of `Driver.subscribe()` entirely. Rather than having every capsule implement subscribe (which I believe per the TRD the implementation can essentially _only_ save the new upcall and return the old one), all upcall management would be handled by the core kernel. The kernel would store the upcall on behalf of the capsule, and handle the swap itself. The capsule would then schedule an upcall with something like `self.grant.schedule_upcall(upcall_id, params)`.

This would be implemented by extending the `create_grant()` interface to take the number of upcalls the capsule wants. Then, when a grant is allocated for a process, the kernel would allocate its own structure in the process's grant region to store the upcalls.

A lot of the machinery to do this has already been implemented by this PR.

#### Effects

- Capsules can't violate upcall swap semantics because they don't actually manage any upcall objects.
- Grant memory _might_ be _reduced_, because we wouldn't need to store the entire current `Upcall` struct (particularly the process ID) for each upcall.
- Code size might go down, because each capsule doesn't have to implement the redundant upcall handling.

and:

- Capsules would not be able to number their upcalls, except starting at 0 and incrementing sequentially. Most do this already, but, it is a restriction.
- I'm not entirely sure how grant creation would be told how many upcalls to hold.

hudson-ayers:
> And a more radical idea: we get rid of `Driver.subscribe()` entirely. Rather than having every capsule implement subscribe (which I believe per the TRD the implementation can essentially _only_ save the new upcall and return the old one), all upcall management would be handled by the core kernel. The kernel would store the upcall on behalf of the capsule, and handle the swap itself. The capsule would then schedule an upcall with something like `self.grant.schedule_upcall(upcall_id, params)`.
> 
> This would be implemented by extending the `create_grant()` interface to take the number of upcalls the capsule wants. Then, when a grant is allocated for a process, the kernel would allocate its own structure in the process's grant region to store the upcalls.
> 
> A lot of the machinery to do this has already been implemented by this PR.
> 
> #### Effects
> * Capsules can't violate upcall swap semantics because they don't actually manage any upcall objects.
> * Grant memory _might_ be _reduced_, because we wouldn't need to store the entire current `Upcall` struct (particularly the process ID) for each upcall.
> * Code size might go down, because each capsule doesn't have to implement the redundant upcall handling.
> 
> and:
> 
> * Capsules would not be able to number their upcalls, except starting at 0 and incrementing sequentially. Most do this already, but, it is a restriction.
> * I'm not entirely sure how grant creation would be told how many upcalls to hold.

I find this idea very appealing, and agree that it would reduce code size. It would also make many undesirable aspects of this PR (like the multiple grant detection) unneeded. One other downside is that capsules can no longer name their upcalls (e.g. `app_rx_buffer` is going to just become something only accessed via `grant.get_upcall(0)`, though I suppose named constants in the capsule can mitigate that).

Re: how grant creation would be told how many upcalls to hold -- I suppose we would make that an associated const on the `Driver` trait?

I think the primary thing that would discourage me from trying this is that it seems pretty tricky to implement. Leon and I ran into a number of "gotchas" when designing what is currently in this PR, and this would be somewhat of a clean-slate take

bradjc:
> Re: how grant creation would be told how many upcalls to hold -- I suppose we would make that an associated const on the `Driver` trait?

Can we do something like `Grant<T, NUM_UPCALLS>`? I don't really know what's possible.

lschuermann:
> Can we do something like `Grant<T, NUM_UPCALLS>`? I don't really know what's possible.

AFAIK generics, whether over types or constants are prone to code bloat due to monomorphization. Also, it'd make the types incompatible to each other.

I did think about this approach in the past but did not pursue it given this implementation was much easier conceptually. Also we've established that we'd rather have something working quickly(ish, this was planned originally for early 2021), and then potentially work out a more efficient and/or cleaner implementation later, possibly after Tock 2.0. As this "simpler" approach (as less about the overall semantics was changed and the kernel diff is also comparatively small) already took three tries and a ton of discussions, I'm somewhat opposed to throwing it over board again. I agree it might be nice, though I have a hard time thinking about a precise implementation strategy. If it works and the implementation is reasonable we could always change it later (the userspace API won't change). I'll keep thinking about it, but these are my 2ct on the issue.

hudson-ayers:
> > Can we do something like `Grant<T, NUM_UPCALLS>`? I don't really know what's possible.
> 
> AFAIK generics, whether over types or constants are prone to code bloat due to monomorphization. Also, it'd make the types incompatible to each other.

There is one grant per `T` already, so I don't think this would lead to any code bloat beyond what already exists. I.e., for a given `T`, there would not be multiple grants with different numbers of upcalls.

hudson-ayers:
One other concern with Brad's approach is that it makes it so capsules can no longer reject `susbcribe()` calls. It probably could be amended to make this possible, if needed, but we would lose some of the benefits (i.e. each capsule might still need to implement `subscribe()`, and some of the code size savings would be reduced)

jrvanwhy:
> One other concern with Brad's approach is that it makes it so capsules can no longer reject `susbcribe()` calls. It probably could be amended to make this possible, if needed, but we would lose some of the benefits (i.e. each capsule might still need to implement `subscribe()`, and some of the code size savings would be reduced)

Do we have a valid use case for rejecting subscribe calls?

hudson-ayers:
> > One other concern with Brad's approach is that it makes it so capsules can no longer reject `susbcribe()` calls. It probably could be amended to make this possible, if needed, but we would lose some of the benefits (i.e. each capsule might still need to implement `subscribe()`, and some of the code size savings would be reduced)
> 
> Do we have a valid use case for rejecting subscribe calls?

Brad asked the same question -- I can't think of one, but I was not as involved with designing the original interface. @phil-levis ?

kernel/src/sched.rs
hudson-ayers: Replaced with terminating the process.

hudson-ayers:
I updated this to not panic in response to a misbehaving capsule, but we can still discuss the other approach

phil-levis:
> Brad asked the same question -- I can't think of one, but I was not as involved with designing the original interface. @phil-levis ?

I don't think we have a good use case for rejecting subscribe calls. It's important that a subscribe can return an error (e.g., in case there is no driver, the pointer is invalid, etc.). But if a subscribe hits a capsule then I think it's OK to say it must be handled. Not allowing a client to change a callback is a strange behavior. Correctness issues (e.g., don't lose outstanding upcalls) are best handled by doing additional dispatch in the upcall (in userspace) rather than adding complexity to when the upcall can change.

lschuermann:
I thought to have responded to this some time ago, but it must've gotten lost somehow...

I'm not convinced we _need_ to able to refuse Subscribe operations. I think that this interface was designed in a similar fashion to the Allow interface, where refusing an Allow operation actually has some advantages: while a process can always fault and memory be reclaimed between two subsequent capsule invocations, a process may choose to reject an Allow call while an important and/or uninterruptable task is being executed. While an Allowed buffer may vanish because the process restarts/faults at any time, and the capsule must handle this gracefully, in this case it does not have to _also_ present a consistent and intact interface to this process again. When a potentially long-running operation can be interrupted by an Allow call from a process, and the capsule has no way to intervene, this handling this may require additional code paths.

Now, circling back to Subscribe/Upcalls: while this reasoning might in general apply to the case of Upcalls as well, given that the only interactions a capsule can do with those is (unconditionally) schedule a call, and it's the process' fault to not be notified of Upcall-based events, there does not seem to be a good reason to refuse Subscribe operations.

bradjc:
Feature merged in #2639.

PR #2464, opened 2021-03-04
Tock 2.0 - AppSlice: fix unsoundness with arbitrary pointers in slices

lschuermann:
### Pull Request Overview

Rust has strict pointer validity requirements, which -- in part -- also apply to accesses of zero length. In Tock 2.0, `AppSlice`s are
allowed to have arbitrary pointers passed in by userspace as long as their length is set to be 0. However, it is still not automatically safe & sound to construct Rust slices from these pointers.

This change ensures that if an AppSlice of length 0 is used (i.e. a Rust slice is created), the pointer is set to the value returned by `NonNull::dangling()`, which is guaranteed to meet the alignment & other requirements for valid pointers posed by Rust, while not requiring (or ensuring) that the memory which the pointer points to is valid or allocated. Therefore, this is only safe for zero-length accesses (e.g. for zero-length slices).

This duplicates a bunch of documentation. However this is a very important invariant we must uphold and as such, I feel it justifies the additional verbosity and explicit explanation on each of the three occurrences.

### Testing Strategy

TODO! This pull request needs testing.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/mem.rs
+++ b/kernel/src/mem.rs
@@ -13,11 +13,73 @@
 //! [`Read`] and [`ReadWrite`] traits, implemented on the
 //! AppSlice-structs.
 
-use core::ptr::NonNull;
-use core::slice;
-
 use crate::capabilities;
 use crate::AppId;
+
+/// Convert an AppSlice's internal representation to a Rust slice.
+///
+/// This function will automatically convert zero-length AppSlices
+/// into valid zero-sized Rust slices regardless of the value of
+/// `ptr`.
+///
+/// # Safety requirements
+///
+/// In the case of `len != 0`, the memory `[ptr; ptr + len)` must be
+/// within a single process' address space, and `ptr` must be
+/// nonzero. This memory region must be mapped as _readable_, and
+/// optionally _writable_ and _executable_. It must be allocated
+/// within a single process' address space for the entire lifetime
+/// `'a`. No mutable slice pointing to an overlapping memory region
+/// may exist over the entire lifetime `'a`.
+unsafe fn raw_appslice_to_slice<'a>(ptr: *const u8, len: usize) -> &'a [u8] {
+    use core::ptr::NonNull;
+    use core::slice::from_raw_parts;
+
+    // Rust has very strict requirements on pointer validity[1] which
+    // also in part apply to accesses of length 0. We allow an
+    // application to supply arbitrary pointers if the buffer length is
+    // 0, but this is not allowed for Rust slices. For instance, a null
+    // pointer is _never_ valid, not even for accesses of size zero.
+    //
+    // To get a pointer which does not point to valid (allocated) memory, but
+    // is safe to construct for accesses of size zero, we must call
+    // NonNull::dangling(). The resulting pointer is guaranteed to be well-aligned
+    // and uphold the guarantees required for accesses of size zero.
+    //
+    // [1]: https://doc.rust-lang.org/core/ptr/index.html#safety
+    match len {
+        0 => from_raw_parts(NonNull::<u8>::dangling().as_ptr(), 0),
+        _ => from_raw_parts(ptr, len),
+    }
+}
+
+/// Convert an AppSlice's internal representation to a mutable Rust
+/// slice.
+///
+/// This function will automatically convert zero-length appslices
+/// into valid zero-sized Rust slices regardless of the value of
+/// `ptr`.
+///
+/// # Safety requirements
+///
+/// In the case of `len != 0`, the memory `[ptr; ptr + len)` must be
+/// within a single process' address space, and `ptr` must be
+/// nonzero. This memory region must be mapped as _readable_ and
+/// _writable_, and optionally _executable_. It must be allocated
+/// within a single process' address space for the entire lifetime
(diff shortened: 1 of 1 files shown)


lschuermann:
Also, to show that this as _very_ drastic implications, take this example:

```rust
use core::ptr::NonNull;
use core::mem::size_of;
use core::slice;

fn main() {
    println!("size_of ptr: {},  Option<ptr>: {}",
        size_of::<&[u8]>(),
        size_of::<Option<&[u8]>>()
    );

    let super_unsound = Some(unsafe {
        slice::from_raw_parts_mut(0x0 as *mut u8, 0)
    });
    
    if let Some(_) = super_unsound {
        println!("All fine.");
    } else {
        println!("Houston, we have a problem!")
    }
    
    let hopefully_sound = Some(unsafe {
        slice::from_raw_parts_mut(NonNull::<u8>::dangling().as_ptr(), 0)
    });

    if let Some(slice) = hopefully_sound {
        println!(
            "NonNull::dangling works as expected, returned ptr: {:p}",
            slice.as_ptr()
        );
    }
}
```

which produces

```
size_of ptr: 16,  Option<ptr>: 16
Houston, we have a problem!
NonNull::dangling works as expected, returned ptr: 0x1
```

hudson-ayers (approved):
The code changes look right to me after reading the reference you provided, great catch.

I will leave it to other to decide whether we should duplicate the documentation in the way you have here or use references to that documentation in a single place.

alistair23 (approved):

kernel/src/mem.rs
```diff
@@ -178,7 +224,24 @@ impl ReadWrite for ReadWriteAppSlice {
                 // memory that the process has `allow`ed to the kernel, and will not permit
                 // the process to free any memory after it has been `allow`ed. This guarantees
                 // that the buffer is safe to convert into a slice here.
-                let slice = unsafe { slice::from_raw_parts_mut(self.ptr, self.len) };
+                //
```
jrvanwhy, line 227: Question tangential to this PR:

How do we prevent a capsule from invoking `mut_map_or` recursively on the same `ReadWriteAppSlice` to get two `&mut` references to a single buffer? In other words: what happens if `fun` calls `mut_map_or` again? Will `process_map_or` not call the closure on the second invocation?

jrvanwhy:
I read through the comments and the code and they look technically correct.

One option to consider would be to extract the `if len == 0 { ... } else { ... }` logic into separate functions, and use a single comment to describe both. Something like:

```rust
// appslice_parts_to_slice and appslice_parts_to_slice_mut convert an AppSlice's
// internal (ptr, len) representation into a slice. This will automatically
// convert zero-length appslices into valid zero-sized slices regardless of the
// value of ptr.
//
// Safety: The memory [ptr, ptr + len) must be within a single process' address
// space. If len is nonzero then ptr must be nonzero as well.
fn raw_appslice_to_slice(ptr: *const u8, len: usize) -> &[u8] {
    use core::ptr::NonNull;
    use core::slice::from_raw_parts;
    match len {
        0 => unsafe { from_raw_parts(NonNull::dangling().as_ptr(), len) },
        _ => unsafe { from_raw_parts(ptr, len) },
    }
}

// See docs for appslice_parts_to_slice.
fn raw_appslice_to_slice_mut(ptr: *mut u8, len: usize) -> &mut [u8] {
    use core::ptr::NonNull;
    use core::slice::from_raw_parts;
    match len {
        0 => unsafe { from_raw_parts_mut(NonNull::dangling().as_ptr(), len) },
        _ => unsafe { from_raw_parts_mut(ptr, len) },
    }
}
```

That would remove some of the redundancy.

jrvanwhy (approved):

kernel/src/mem.rs
lschuermann: I don't think we currently protect against this, which is pretty bad. I _think_ the usual thing to do in Rust would be to require a mutable borrow of the "parent" object here, but I'm not sure how that will collide with the rest of the ecosystem (i.e. expectations by capsule code regarding non-mutable reference to ReadWriteAppSlice). I'll analyze and create a PR accordingly, if you don't mind. Memory safety is hard. :)

hudson-ayers: TakeCell style semantics might be easier to integrate

lschuermann: It turns out that requiring a mutable borrow of the `ReadWriteAppSlice` itself only breaks two capsules (which I've just mechanically fixed so the borrow-checker is happy). Should be fixed in #2465. Thanks for reporting this @jrvanwhy!

lschuermann:
@jrvanwhy Thanks for those recommendations. As suggested, I've split out the actual creation of the Rust slices based on AppSlice information into separate unsafe functions, which also gives us a chance to specify the required safety requirements in a more explicit, encapsulated way.

bradjc (commented):
Looks good, but took me a while to figure out where `from_raw_parts` is defined.

ppannuto (approved):

jrvanwhy (approved):

PR #2467, opened 2021-03-05
ci: Automatically add fmt commit if needed.

bradjc:
### Pull Request Overview

I found this https://github.com/mbrobbel/rustfmt-check and copied the example workflow. Does this work? I don't know! Would it be cool if it did? Yes!

I figure we can see if this works before changing our other CI formatting rules.

### Testing Strategy

yolo

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/.github/workflows/rustfmt.yml
+++ b/.github/workflows/rustfmt.yml
@@ -1,0 +1,17 @@
+# This workflow automatically adds a Rustfmt commit if needed to a PR.
+
+on: pull_request
+
+name: Rustfmt
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+            components: rustfmt
+      - uses: mbrobbel/rustfmt-check@master
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/boards/hail/src/main.rs
+++ b/boards/hail/src/main.rs
@@ -15,10 +15,10 @@ use capsules::virtual_spi::VirtualSpiMasterDevice;
 use kernel::capabilities;
 use kernel::common::dynamic_deferred_call::{DynamicDeferredCall, DynamicDeferredCallClientState};
 use kernel::component::Component;
+use kernel::hil;
 use kernel::hil::i2c::I2CMaster;
 use kernel::hil::led::LedLow;
 use kernel::hil::Controller;
-use kernel::hil;
 use kernel::Platform;
 #[allow(unused_imports)]
 use kernel::{create_capability, debug, debug_gpio, static_init};
@@ -129,10 +129,10 @@ unsafe fn set_pin_primary_functions(peripherals: &Sam4lDefaultPeripherals) {
     peripherals.pa[14].configure(None); //... BLUE_LED
     peripherals.pa[15].configure(None); //... GREEN_LED
     peripherals.pa[16].configure(None); //... BUTTON - User Button
-          peripherals.pa[17].configure(None); //... !NRF_RESET - Reset line for nRF51822
+    peripherals.pa[17].configure(None); //... !NRF_RESET - Reset line for nRF51822
     peripherals.pa[18].configure(None); //... ACC_INT2 - FXOS8700CQ Interrupt 2
-    peripherals.pa[19].configure(None   ); //... unused
-    peripherals.pa[20].configure(  None); //... !LIGHT_INT - ISL29035 Light Sensor Interrupt
+    peripherals.pa[19].configure(None); //... unused
+    peripherals.pa[20].configure(None); //... !LIGHT_INT - ISL29035 Light Sensor Interrupt
                                         // SPI Mode
     peripherals.pa[21].configure(Some(A)); // D3 - SPI MISO
     peripherals.pa[22].configure(Some(A)); // D2 - SPI MOSI
--- a/boards/hail/src/main.rs
+++ b/boards/hail/src/main.rs
@@ -15,10 +15,10 @@ use capsules::virtual_spi::VirtualSpiMasterDevice;
 use kernel::capabilities;
 use kernel::common::dynamic_deferred_call::{DynamicDeferredCall, DynamicDeferredCallClientState};
 use kernel::component::Component;
-use kernel::hil;
 use kernel::hil::i2c::I2CMaster;
 use kernel::hil::led::LedLow;
 use kernel::hil::Controller;
+use kernel::hil;
 use kernel::Platform;
 #[allow(unused_imports)]
 use kernel::{create_capability, debug, debug_gpio, static_init};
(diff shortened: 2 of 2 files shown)


lschuermann:
Would you require the author to rebase (and squash) their changes if a fmt commit is added? My comment applies if that is not the case:

To be honest with you, I don't think this is a good idea at all.

If this makes it in, we've got a lot of just "formatting" commits there, which will (if one relies on these CI-generated commits) make something like Git blame and other tools pretty much unusable. Furthermore, it clutters our Git history with commits that have no relevance regarding the actual functionality and should really be squashed into the commits they should be fixing.

I know that Tock has a very relaxed policy around commit messages, although there technically [are guidelines documented](https://github.com/tock/tock/blob/7995e8b42e57898b509fc4312dd34f3688a6194a/.github/CONTRIBUTING.md#step-3-commit). I think we could do significantly better here, and this would be a step in the opposite direction.

ppannuto:
@lschuermann What if we were to [autosquash](https://github.com/tibdex/autosquash) the format commit?

It messes with people's git history / branches, but not in too obtrusive of a manner.. we could also add something to the PR template that warns folks that an auto-format will run on pushed PR branches.

bradjc:
I'm a huge proponent of clean PR commit histories. https://mystor.github.io/git-revise.html is a marvelous tool, and I've spent longer than I care to admit rebasing and shuffling commits, and using -p and everything else so that if I need a git bisect the commits make sense. A few years ago we discussed this, and we experimented a little bit with "squash and merge" on github. This also I think motivated some of the text in our contributing doc (which I still think is good guidance in any case, and with few exceptions guidance I personally follow, so I support it still being there).

However, over time it just doesn't seem worth it to care. `git rebase` is practically impossible for anyone new to git to use, and expecting PR authors to use it is unreasonable. Today, we have "formatting" commits on many PRs, so with or without this we will end up with formatting PRs (assuming we don't remove rustfmt). This, however, removes the burden on the developer who may not be an experienced Rust developer. I think it also makes contributing a little more friendly, because some PRs will get a green check rather than a red x when the only issue is some spacing is wrong. Also, PRs that don't pass travis _tend_ to get ignored or not reviewed compared to those that do. I'd like to see us sacrifice some commit history messiness for a slightly improved contributing experience.

lschuermann:
> However, over time it just doesn't seem worth it to care.

It's sad that ultimately tools like GitHub have managed to shift the focus away from Git to their workflows that we have come to this point. I see your point and made some of the same experiences when working on non-Tock projects. Maybe this is really just me, but I'm having a hard time giving up on this and see the value which a clean & concise Git history brings us to be almost immeasurable, compared to the effort it'd take to rebase a change locally. Also, I'd personally see it as a huge inconvenience and annoyance if something would generate automatic commits on top of my tree.

Where I don't quite follow: `rebase` is "hard", but wouldn't this cause people to use it even more? If people don't rebase locally as a response to PR feedback, then they're going to make commits on top of their GitHub branch ... which will contain the formatting commits now? So at least a `pull --rebase` would be required constantly, and formatting changes do have the tendency to break automatic merges. And given that they don't apply to individual commits but the whole pushed set at once, resolving those conflicts would be trickier than just me doing the formatting (and proper rebase) locally.

ppannuto:
For the common-case, I think I would pretty happily revisit defaulting to squash & merge for PRs. I think [this post linked from the autosquash bot documentation](https://zachholman.com/posts/git-commit-history/) summarizes the reality well -- the 'unit of change' for most things in Tock is the PR. I _occasionally_ look at the series of commits in a PR, but the reality is that I primarily review via the 'Files Changed' tab.

---

Re: local history and non-expert git users. That's a pretty compelling use case against an autosquash for format commits, because it will create divergence that's hard for users to navigate. I imagine that the average user we're worried about here won't know of or use the `--rebase` flag for pull, rather they will just create a local merge (the default behavior -- and thus implicit preference -- of git). The addition of all these meaningless merges would further support squashing the final PR perhaps.

alistair23:
I'm 100% with Leon here. The GitHub workflow is already awful and adding this would make it even worse. We should be making commits standalone (including formatting) instead of lumping a automated format change on the authors commits.

I also don't think squashing the PR into a single commit is a good idea. PRs end up being big and a lot of effort is (and should be) spent in splitting up PRs to small easy to review commits and adding detailed and useful commit messages. Reviewing is hard enough and time consuming enough already, we shouldn't make it even harder.

Tock has a very messy history already and is hard to impossible to bisect and follow. I think we should be moving to improve on that and align ourselves with other open source projects (glibc, Linux, QEMU ect.) instead of making the history more confusing by adding automated commits.

This will also be a pain to deal with when pushing changes as people don't rebase their local branches from their own remote, so the auto-formated changes will be lost when a user makes a change.

bradjc:
This PR is only about automatically adding a fmt commit if needed, not changing PR merge policies.

For everyone who runs rustfmt automatically or before issuing a PR, this will have no effect. So I don't think this affects PRs from those commenting here :).

----

Re: changing the remote branch but not changing a user's local branch: this is a valid concern. I still the net benefit is there for two reasons: 1) It still gets the PR a green check, which is better for getting the PR reviewed, and 2) suggested changes in PRs have the same effect so this is going to become more common, not less.

If a user does add commits, but does not pull the changes, then there are few things that can happen:

- They force push. The fmt commit gets re-added. No rebase, no issue.
- They merge/rebase locally, and everything goes well. They push. No manual rebase, no issue.
- They merge/rebase locally, and it does not work. They will have to resolve the merge somehow. This isn't great, but I don't think it will be that common. And, this is a problem with or without this PR.

alistair23:
How many people would this help though. When do we have an issue where a PR doesn't pass formatting and the author doesn't fix it up?

I don't see much benefit, just lots of downsides of messing around and auto adding commits. Every commit in a PR should build and pass formatting, not just the last commit, which this is then encouraging.

bradjc:
This would mainly help first time contributors, of which we have had a few PRs where CI didn't pass because of formatting. I don't have the time or patience to get a number, but it certainly has happened.

Authors (or reviewers sometimes) do fix the issue, by adding a formatting commit. The net effect is the same, but since CI can do this for us, why not take advantage of that? We could do all of the things CI does manually, but why not let robots help us? I guess this is why I don't see any downside, as I don't think this changes anything. If a PR author wants to have well-organized commits, and some formatting issue occurs and this auto commits, they would just remove or squash the auto commit. For the majority of users, they would just create the exact same commit anyway, but now it happens automatically.

bradjc:
Since it is hard to remember when this PR would be useful, I want to note #2760 where this functionality would have both been helpful and resulted in the same outcome.

image

phil-levis (commented):
I'd personally prefer this. To Alistair's point, I often make commits that do not pass formatting or CI, e.g., when transitioning between machines or because the `make format` warns you if you try to format with uncommitted changes. But I'm fine with how things are now.

PR #2470, opened 2021-03-08
riscv: Use macros to create non-PMP CSRs

alistair23:
### Pull Request Overview

The current giant switch table for RISC-V CSRs is a pain to edit and extend.

With more RISC-V extensions on the way and with some RISC-V implementations using CSRs to modify the system the list of CSRs is only growing.

In an effort to try and cleanup and improve the RISC-V CSRs (which are also some of the largest functions in the Tock kernel) I have added a macro that generates RISC-V CSR access functions.

Unfortunately I can't figure out a way to do with for the array of PMP CSRs.

### Testing Strategy

CI

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/arch/riscv/src/csr/mod.rs
+++ b/arch/riscv/src/csr/mod.rs
@@ -1,6 +1,15 @@
 //! Tock Register interface for using CSR registers.
 
-use riscv_csr::csr::ReadWriteRiscvCsr;
+use core::marker::PhantomData;
+use tock_registers::registers::{
+    Field, FieldValue, IntLike, LocalRegisterCopy, RegisterLongName, TryFromValue,
+};
+
+use riscv_csr::csr::{
+    ReadWriteRiscvPmpCsr, MCAUSE, MCYCLE, MCYCLEH, MEPC, MIE, MINSTRET, MINSTRETH, MIP, MSCRATCH,
+    MSTATUS, MTVAL, MTVEC, STVEC, UTVEC,
+};
+use riscv_csr::riscv_csr;
 
 pub mod mcause;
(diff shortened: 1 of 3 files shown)


hudson-ayers:
Awhile back I suggested using const generics to pass the addresses of the non-PMP CSRs, as a substitute for the messy code in ... . A link to a commit that achieved this is here: https://github.com/hudson-ayers/tock/commit/ab4965cea5270ee3b5515513669b118a357ee7b6 , though it is of course outdated. Now that `min_const_generics` is stable I think it would be reasonable to consider that approach. Do you think that would be preferable to what you have here? I haven't fully paged all this back in, but I think it would allow for the interface we want without using macros?

alistair23:
I had a quick look at that, it's not that simple though as I get this error:

```
error: type parameters with a default must be trailing
   --> libraries/riscv-csr/src/csr.rs:106:42
    |
106 | pub struct ReadWriteRiscvCsr<T: IntLike, R: RegisterLongName = (), const V: usize> {
    |                                          ^
    |
    = note: using type defaults and const parameters in the same parameter list is currently not permitted
```

hudson-ayers:
It seems like this PR doesn't help with the array problem either though? Your other recent PR does, but the costs of dynamic dispatch are probably too high anyway

alistair23:
Yep, it doesn't either. I was originally hoping to get this working the the array (see https://github.com/tock/tock/pull/2471) but the overhead is too high.

I just got the const generic version working (good idea :) ) so I'm going to close this and use that instead.

PR #2478, opened 2021-03-10
Cortex-m: document that userspace stack pointer will always be aligned on entry to the kernel

hudson-ayers:
### Pull Request Overview

~~This pull request checks that the process stack pointer is aligned upon return from userspace. This is needed because Rust requires references be correctly aligned, and `set_syscall_return_value()` creates references to items on the stack offset from the stack pointer value.~~
This PR updates code comments.

### Testing Strategy

~~This pull request was tested by running blink, stack_size_test01, and stack_size_test02 on Imix. I also tried to make a test that would intentionally unalign the stack pointer from userspace to verify these changes would catch it (I used inline asm to decrement the stack pointer by 1), but the process faults before returning to the kernel. So we have to take this on faith for now unless someone can make a better test.~~
None needed

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/syscall.rs
+++ b/arch/cortex-m/src/syscall.rs
@@ -113,11 +113,13 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
         let sp = state.psp as *mut u32;
         let (r0, r1, r2, r3) = (sp.offset(0), sp.offset(1), sp.offset(2), sp.offset(3));
 
-        // TODO: Are these guarantees _always_ satisfied? E.g. must
-        // the stack_pointer always be properly aligned?
-
         // These operations are only safe so long as
-        // - the pointers are properly aligned
+        // - the pointers are properly aligned. This is guaranteed because the
+        //   pointers are all offset multiples of 4 bytes from the stack
+        //   pointer, which is guaranteed to be properly aligned after
+        //   exception entry on Cortex-M. See
+        //   https://github.com/tock/tock/pull/2478#issuecomment-796389747
+        //   for more details.
         // - the pointer is dereferencable, i.e. the memory range of
         //   the given size starting at the pointer must all be within
         //   the bounds of a single allocated object


arch/cortex-m/src/syscall.rs
```diff
@@ -191,9 +191,10 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
         // within process accessible memory.
```
bradjc, line 191:
```suggestion
        // within process accessible memory, and that the stack pointer
        // is properly aligned.
```

bradjc (approved):

jrvanwhy:
I tried to test this on H1 (a Cortex-M3) by adding 2 or 1 to the stack pointer and making a system call. The check did not trigger. The system call returned and the process continued executing.

I then had the app fault with a `udf` instruction, and the stack pointer the kernel printed out was aligned. I suspect the hardware is dropping the lower bits of the stack pointer when a system call is made.

hudson-ayers:
> > but the process faults before returning to the kernel.
> 
> Presumably because the stack is unaligned?

The fault I get is an invalid state usage fault, not an unaligned access fault. Not exactly sure why I get that. If I add 1 to the stack pointer instead of subtracting 1 I see what Johnathan sees -- the hardware immediately drops the lower bits and leaves the stack pointer unchanged. So I think when I was subtracting 1 the hardware was actually decrementing the stack by 4 bytes, and that just happened to cause a fault, probably because of the compiler expecting something on the stack when returning from main().

hudson-ayers:
So...will all Arm cortex-m hardware refuse to set the stack pointer to an unaligned value? Should we keep this check just in case?

ppannuto:
Huh, TIL: Refuse to set isn't quite the right thing -- it will allow an unaligned stack pointer during normal operation, but on exception entry it will correct it:

image
image

with that implementation in mind, we can safely know (for CortexM at least) that things will always be aligned by the time we hit the kernel.

bradjc:
So just update this PR to a comment noting this?

hudson-ayers:
Updated.

bradjc (approved):

alistair23 (approved):

PR #2479, opened 2021-03-10
Remove ExitCall type, don't specify representation of YieldCall type

hudson-ayers:
### Pull Request Overview

This pull request removes the completely unused `ExitCall` type, and removes the `#[repr(u8)]` from `YieldCall`, as `YieldCall` is not used across the system call boundary, so there is no need to specify its representation.

### Testing Strategy

This pull request was tested by compiling

### TODO or Help Wanted

N/A

### Documentation Updated

- [x] No updates are required.

### Formatting

- [x] Ran `make prepush`.

PR #2480, opened 2021-03-10
Tock 2.0: Use ExitCall Type

bradjc:
### Pull Request Overview

This pull request uses the ExitCall type. Rebuttal to #2479 :). There might be a nicer way to implement this.

I'm not particularly attached to either approach, but now we can compare, I guess.

### Testing Strategy

travis

### TODO or Help Wanted

Might be a cleaner implementation.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/sched.rs
+++ b/kernel/src/sched.rs
@@ -30,7 +30,7 @@ use crate::platform::watchdog::WatchDog;
 use crate::platform::{Chip, Platform};
 use crate::process::{self, Task};
 use crate::syscall::{ContextSwitchReason, SyscallReturn};
-use crate::syscall::{Syscall, YieldCall};
+use crate::syscall::{ExitCall, Syscall, YieldCall};
 
 /// Threshold in microseconds to consider a process's timeslice to be exhausted.
 /// That is, Tock will skip re-scheduling a process if its remaining timeslice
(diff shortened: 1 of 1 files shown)


kernel/src/sched.rs
```diff
@@ -989,15 +989,31 @@ impl Kernel {
...
+                        // system call class.
+                        process
+                            .set_syscall_return_value(SyscallReturn::Failure(ErrorCode::NOSUPPORT));
+                    },
+                    |exit| match exit {
+                        ExitCall::Terminate => process.terminate(completion_code as u32),
+                        ExitCall::Restart => process.try_restart(completion_code as u32),
+                    },
+                );
+            }
```
hudson-ayers, line 1016:
```suggestion
            } => match which {
                // The process called the `exit-terminate` system call.
                ExitCall::Terminate as u32 => process.terminate(completion_code as u32),
                // The process called the `exit-restart` system call.
                ExitCall::Restart as u32 => process.try_restart(completion_code as u32),
                // The process called an invalid variant of the Exit
                // system call class.
                _ => process.set_syscall_return_value(SyscallReturn::Failure(ErrorCode::NOSUPPORT)),
            },
```

I think this is a less convoluted use, even if it doesn't enforce handling each variant of the exit system call class.

bradjc:
```
error: expected one of `!`, `(`, `...`, `..=`, `..`, `::`, `=>`, `if`, `{`, or `|`, found keyword `as`
   --> kernel/src/sched.rs:995:41
    |
995 |                     ExitCall::Terminate as u32 => process.terminate(completion_code as u32),
    |                                         ^^ expected one of 10 possible tokens

warning: unused import: `ExitCall`
  --> kernel/src/sched.rs:33:22
   |
33 | use crate::syscall::{ExitCall, Syscall, YieldCall};
   |                      ^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

error: aborting due to previous error; 1 warning emitted

error: could not compile `kernel`
```

hudson-ayers (approved):
I am fine with this approach

phil-levis:
I agree with this PR in principle, but given how much complex it makes the code it seems like a step backwards.

phil-levis (changes requested):
This increases the code from 8 to 24 lines... while I agree using the type is good in principle, I don't think it's worth tripling code size.

Is there an easy way to see how the resulting binary size changes?

ppannuto:
> weact-f401ccu6: flash use +3988 bytes (+5.07%), RAM use +0 bytes(+0.00%)
>
> _From the "Benchmarks" CI check:_ https://github.com/tock/tock/pull/2480/checks?check_run_id=2080253776

..which is.. surprisingly large given the change here :/

bradjc:
That seems large, doesn't it. I think that is comparing to master. When I build the weact locally I get the exact same numbers on this branch and on tock-2.0-dev when size runs.

hudson-ayers:
yeah, sorry, the Benchmarks CI check is only setup to run on master and the results are incorrect on other branches

PR #2481, opened 2021-03-12
Initial support for SweRV EH1 and SweRVolf

alistair23:
### Pull Request Overview

This PR adds support for running Tock on the SweRVolf EH1 Verilator simulator.

This is useful as we can run a RISC-V Tock model fully in software by just using the make scripts. My plan is to use this in CI eventually as well (as long as it isn't too slow).

#### SweRV EH1

> The Western Digital SweRV Core EH1 is a 32-bit, 2-way superscalar, 9 stage pipeline core.
> 
> The SweRV Cores will be used in a variety of Western Digital products in the near future. All three cores have been open sourced in CHIPS Alliance github. CHIPS Alliance is an organization that develops open source hardware and open source software design tools. It is a barrier free, collaborative environment which lowers the cost of RTL development. The SweRV Core EH1 is available at https://github.com/chipsalliance/Cores-SweRV

From: https://www.westerndigital.com/company/innovations/risc-v

#### SweRVolf

SweRVolf is a FuseSoC-based SoC for the SweRV RISC-V core.

### New Platform Checklist

- [X] Hardware is widely available. (It's a simulation!)
- [X I can support the platform, which includes release testing for the platform, at least initially.
  - [X] `Console`, including `debug!()` and userspace `printf()`.
  - [X] Timers.
  - [ ] GPIO with interrupts. (There isn't really a GPIO)

### Testing Strategy

Running on the SweRVolf EH1 simulation.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


alevy:
A very initial review, just trying to run this locally:

- Can you please add `.gitignore`s for all the build artifacts? For me these are:
  - `binary.hex`
  - `build/`
  - `fusesoc.conf` 
  - `fusesoc_libraries/`
  - `makehex.py`
- The board builds and the simulator runs for me, but there is no output after "Releasing reset" after a few minutes of pinning one of my CPUs (no worries, though, I have plenty of those).
- I have no idea why the size tool is borking on the build...

hudson-ayers:
the size tool fails whenever a new board is added, because it fails when trying to build the board on current master. It won't block merging though

alistair23:
I have updated the `.gitignore`.

Strange, I'm not sure why it's not working. I see prints within about 20 seconds

alistair23:
Do you have a new enough version of Verilator?

alevy:
> Do you have a new enough version of Verilator?

you tell me!

```
Verilator 4.108 2021-01-10 rev v4.106-158-g484b76e5b
```

In any case, it's probably an issue with my setup, just flagging it. Since obviously you won't be able to reproduce (it's working for you), we can maybe sync online to try and resolve.

alistair23:
That looks right. That's strange. I'm on Slack to help debug if you want (or Matrix :) )

chips/swerv/src/eh1_pic.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub fn disable_all(&self) {
+        for enable in self.registers.meie.iter() {
+            enable.write(MEIE::INTEN::DISABLE);
+        }
+    }
+
+    /// Get the index (0-256) of the lowest number pending interrupt, or `None` if
+    /// none is pending. RISC-V PIC has a "claim" register which makes it easy
+    /// to grab the highest priority pending interrupt.
+    #[cfg(all(target_arch = "riscv32", target_os = "none"))]
```
hudson-ayers, line 166: also seems like this should not need a `cfg`?

chips/swervolf-eh1/src/syscon.rs
```diff
@@ -0,0 +1,198 @@
...
+use kernel::ReturnCode;
+
+/// 100Hz `Frequency`
+#[derive(Debug)]
+pub struct Freq100Hz;
+impl time::Frequency for Freq100Hz {
+    fn frequency() -> u32 {
+        10
+    }
+}
```
hudson-ayers, line 17:
```suggestion
/// 100Hz `Frequency`
#[derive(Debug)]
pub struct Freq100Hz;
impl time::Frequency for Freq100Hz {
    fn frequency() -> u32 {
        100
    }
}
```
?

alistair23:
@hudson-ayers Fixed your requests

hudson-ayers:
I can confirm that I see prints on my machine (Ubuntu 20.04, only thing I changed was running `sudo apt install verilator` instead of `pacman`). Very cool!

Can you add the new platform checklist to the PR description? It seems that this PR does not support GPIO, which maybe is fine given it is meant to be used primarily as a simulation platform anyway, but wanted to flag that for discussion.

alevy:
> I can confirm that I see prints on my machine

I must be losing my edge

image

alevy:
Works for me as well in a Ubuntu container:

```
Loading RAM contents from /tock/boards/swervolf/binary.hex
Releasing reset
SweRVolf initialisation complete.
Entering main loop
```

So this is definitely a toolchain thing, hrm....

alevy:
And frustratingly, I'm able to run the example `.vh` (hex) files in the fusesoc_libraries repo (e.g. `zephyr_hello`) and get output.

The only difference I can identify between my working Ubuntu version and non-working NixOS version is that on Ubuntu I wasn't using Python's `venv` to install/use `fusesoc` while on NixOS I am.

Tested now with several versions of verilator (default on NixOS is more recent than on Ubuntu), including exactly the one on Ubuntu. I have no clue what might be different. Maybe another NixOS user (@lschuermann?) or a Mac user (for some more toolchain diversity) can try and get this working?

hudson-ayers:
> The only difference I can identify between my working Ubuntu version and non-working NixOS version is that on Ubuntu I wasn't using Python's `venv` to install/use `fusesoc` while on NixOS I am.

I also used `venv`, though I hope that would not make a difference

alistair23:
I updated the description with the checklist.

As for the Nix problem I'm not really sure what to do. Maybe file an issue with FuseSoC?

boards/swervolf/README.md
```diff
@@ -0,0 +1,75 @@
+SweRVolf
+========
+
+SweRVolf is a FuseSoC-based SoC for the SweRV RISC-V core.
+
+This can be used to run the RISC-V compliance tests, Zephyr OS
+or other software in simulators or on FPGA boards. Focus is on
+portability, extendability and ease of use; to allow SweRV users
+to quickly get software running, modify the SoC to their needs or
+port it to new target devices.
```
bradjc, line 10: Can you add a link to a product/project page or something similar?

hudson-ayers (approved):

lschuermann (commented):
Looks pretty good. Great to see another FPGA / Verilated board in Tock! I think this is reasonably similar to what LiteX can do, so we can probably use all CI infrastructure with both of them! :)

Outlined some minor change requests. Edit: ugh, looks like the GitHub UI bugged out again and lost my review comments.

boards/swervolf/Makefile
```diff
@@ -0,0 +1,20 @@
...
+include ../Makefile.common
+
+# Default target for installing the kernel.
+.PHONY: install
+install: sim
+
+setup-sim:
+	fusesoc library add fusesoc-cores https://github.com/fusesoc/fusesoc-cores
+	fusesoc library add swervolf https://github.com/chipsalliance/Cores-SweRVolf
+	wget https://raw.githubusercontent.com/chipsalliance/Cores-SweRVolf/master/sw/makehex.py
```
lschuermann, line 15: Can you add a notice that this downloads stuff from the internet _and_ add a checksum verification here? I wouldn't want to have anything downloaded from another repository as part of the Makefile Infrastructure, even worse it being executable code.

.gitignore
```diff
@@ -65,3 +65,10 @@ package-lock.json
 
 # Local Netlify folder
 .netlify
+
+# FuseSoC build files
```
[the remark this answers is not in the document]
alistair23: Fixed

boards/swervolf/Makefile
alistair23: The file is tiny so I just checked it in.

boards/swervolf/README.md
```diff
@@ -0,0 +1,77 @@
...
+Running in simulation with EH1 core
+-----------------------------------
+For full details in setting up the FuseSoC simulator see: https://github.com/chipsalliance/Cores-SweRVolf#prerequisites
+
+The quick steps to setup are shown below:
+
+First install verilator
+
+```shell
+sudo pacman -S verilator
```
lschuermann, line 31: Can we change this to not be Arch Linux (or any other pacman-based distribution) specific? I imagine just specifying the required dependencies and their tested version would be great.

boards/swervolf/README.md
```diff
@@ -0,0 +1,77 @@
...
+functionality impact.
+
+```diff
+diff --git a/arch/rv32i/src/lib.rs b/arch/rv32i/src/lib.rs
+index 5b6a23615..5e207a423 100644
+--- a/arch/rv32i/src/lib.rs
++++ b/arch/rv32i/src/lib.rs
+@@ -88,7 +88,7 @@ pub extern "C" fn _start() {
+ /// This moves the data segment from flash to RAM and zeros out the BSS section.
+ pub unsafe fn init_memory() {
```
lschuermann, line 70: Just commenting so that we're aware #2461 is in last-call and this documentation might be outdated very soon.

boards/swervolf/README.md
alistair23: I would prefer to leave it as is. Tock has lots of distro/OS specific steps (like OSX and Ubuntu) so I don't think it's strange to include this. It also makes it easier for people to get started and for me to remember how to set it up :P

lschuermann: Fair enough. Care to include the version for this and the other packages at least? It's not unlikely to have ancient packages (looking at you, Debian). Having a version reference (at least some lower boundary) rules out a lot of errors.

alistair23: The full SweRVolf contains that type of information. I would rather leave it there instead of duplicating it here.

lschuermann: Fair enough. I think this will be the place where people are going to look, but it's up to you.

hudson-ayers (approved):

bradjc (approved):

alevy (commented):
I don't really have much to add here. I'm able to build and run the simulator in a Ubuntu docker container, but not natively in NixOS (I can build the board but the simulator just hangs)---a bummer, since it means realistically I won't test it and also Nix is our best documentation (the Makefiles are also good for general tools, but not board specific ones) of which specific versions of different tools in the chain actually work. I am, conversely able to run pre-compiled bare metal example binaries from the fusesoc repo, so it's not as though the toolchain is entirely broken on my end.

I'm fine with merging this. It's about the state of Open titan in terms of runnability (some people seem to be able to) and if I hadn't happened to try it out, we would have likely merged it without waiting for my comment.

PR #2487, opened 2021-03-16
allow: move allow system calls handling to sched.rs

lschuermann:
### Pull Request Overview

As discussed on the core team call of 2021-03-12 and requested by @bradjc.

This moves the handling of allow-style system calls (both read-only and read-write) allow from `process.rs` to `sched.rs`, given that the scheduler should have the ultimate authority of interaction between processes and capsules and implement the system call ABI. This also mirrors how other system calls are currently handled in Tock.

The logic for checking whether an allow request is valid w.r.t. to the process' memory regions is still implemented in process.rs. It doesn't make sense to change this, because checking whether an allow request should be granted (whether the buffer is in the process-accessible memory region) is a responsibility of the `ProcessType` anyways.

### Testing Strategy

This pull request requires testing.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -386,66 +386,48 @@ pub trait ProcessType {
     // additional memop like functions
 
     /// Creates a `ReadWriteAppSlice` from the given offset and size
-    /// in process memory and invokes `allow_readwrite` on a `Driver`
-    /// implementation to invoke its system call.
-    ///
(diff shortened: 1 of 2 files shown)


hudson-ayers (approved):

bradjc (commented):
Looks good. Reviewing in more depth context of #2446.

PR #2494, opened 2021-03-22
kernel: remove memory slice from Process struct

### LINKED ISSUE #2369: Process::memory violates aliasing requirements (undefined behavior)

bradjc:
### Pull Request Overview

This pull request removes the mutable slice to process memory that the Process object holds. This helps avoid undefined behavior since there are many cases (like AppSlices) where we generate new slices to the same process memory.

I also was able to remove an `unsafe` by changing `add`->`wrapping_add`.

Fixes #2369.

### Testing Strategy

Running an app on Hail. The change is pretty mechanical.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -124,8 +124,12 @@ impl fmt::Debug for ProcessLoadError {
 /// Processes are found in flash starting from the given address and iterating
 /// through Tock Binary Format (TBF) headers. Processes are given memory out of
 /// the `app_memory` buffer until either the memory is exhausted or the
-/// allocated number of processes are created. A reference to each process is
-/// stored in the provided `procs` array. How process faults are handled by the
+/// allocated number of processes are created. This buffer is a non-static slice,
+/// ensuring that this code cannot hold onto the slice past the end of this function
+/// (instead, processes store a pointer and length), which necessary for later
+/// creation of `AppSlice`'s in this memory region to be sound.
+/// A reference to each process is stored in the provided `procs` array.
+/// How process faults are handled by the
 /// kernel must be provided and is assigned to every created process.
 ///
 /// This function is made `pub` so that board files can use it, but loading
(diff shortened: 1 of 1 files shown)


hudson-ayers (dismissed):
I added a commit that goes a little further and makes `load_processes()` not require a `'static` reference to `app_memory`. This makes it easy to verify that `process.rs` cannot hold onto a slice to app memory without using unsafe to construct it.

lschuermann:
Whoops, I suppose this wasn't supposed to happen. @bradjc Can you reopen this by pointing to `master` as the target branch?

hudson-ayers (approved):

lschuermann (approved):

alistair23 (dismissed):

jrvanwhy:
> I added a commit that goes a little further and makes `load_processes()` not require a `'static` reference to `app_memory`. This makes it easy to verify that `process.rs` cannot hold onto a slice to app memory without using unsafe to construct it.

I think this change deserves an explanatory comment in the code. I agree with the change, but until I read you (Hudson's) comment here I was planning to add a review comment indicating that using a `'static` reference seems more correct.

hudson-ayers (approved):

jrvanwhy (approved):

ppannuto (approved):
bors r+

PR #2499, opened 2021-03-22
SweRVolf: Add support for the platform timers

alistair23:
### Pull Request Overview

This PR adds support for the two internal timers. This means we now don't have to mux the RISC-V mtimer.

### Testing Strategy

Running in simulation.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


chips/swerv/src/eh1_timer.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    fn minimum_dt(&self) -> Self::Ticks {
+        Self::Ticks::from(1 as u32)
+    }
+}
+
+impl kernel::SchedulerTimer for Timer<'_> {
+    fn start(&self, us: u32) {
+        let now = self.now();
+        let tics = Self::ticks_from_us(us);
```
hudson-ayers, line 186: note that `ticks_from_us()` rounds down, so if us is less than 10000 this will set an alarm for `now`

chips/swerv/src/eh1_timer.rs
```diff
@@ -0,0 +1,213 @@
...
+            mitcnt1: ReadWriteRiscvCsr::new(),
+            mitb0: ReadWriteRiscvCsr::new(),
+            mitb1: ReadWriteRiscvCsr::new(),
+            mitctl0: ReadWriteRiscvCsr::new(),
+            mitctl1: ReadWriteRiscvCsr::new(),
+        }
+    }
+
+    pub fn handle_interrupt(&self) {
+        self.stop();
```
hudson-ayers, line 68: Why stop the timer here? Could it not just be allowed to continue?

chips/swerv/src/eh1_timer.rs
```diff
@@ -0,0 +1,213 @@
...
+        BOUND OFFSET(0) NUMBITS(32) []
+    ],
+    MITCTL [
+        ENABLE OFFSET(0) NUMBITS(1) [],
+        HALT_EN OFFSET(1) NUMBITS(1) [],
+        PAUSE_EN OFFSET(2) NUMBITS(1) [],
+    ],
+];
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
alistair23: Ah, it's not needed anymore

chips/swerv/src/eh1_timer.rs
```diff
@@ -0,0 +1,213 @@
...
+use riscv_csr::csr::ReadWriteRiscvCsr;
+
+/// 100Hz `Frequency`
+#[derive(Debug)]
+pub struct Freq100Hz;
+impl time::Frequency for Freq100Hz {
+    fn frequency() -> u32 {
+        100
+    }
+}
```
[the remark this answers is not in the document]
alistair23: This isn't actually right either. I just picked this and forgot to come back and adjust it.

In a simulation timing is a little strange, but the hardware shows a frequency of 50MHz: https://github.com/chipsalliance/Cores-SweRVolf/blob/2e1802e839923f593d49df18c8c46572b48ff806/rtl/swervolf_core.v#L214

chips/swerv/src/eh1_timer.rs
alistair23: It can, but then you will get another interrupt when it ticks over the compare register again. The hardware automatically rests the counter when an interrupt occurs.

hudson-ayers: oh, makes sense.

alistair23:
I don't have the FPGA board so I haven't tested this on the FPGA. The Tock board readme only talks about the simulation.

Apparently the FPGA and simulation have the same frequency though (if I'm reading the system verilog correctly).

hudson-ayers:
I think the issue was that using the same frequency in verilator made running anything meaningful unbearably slow. But if this seems to work fine in practice, sounds good to me.

bradjc (approved):

chips/swerv/src/eh1_timer.rs
hudson-ayers: Resolved by the change to higher frequency

hudson-ayers (approved):
bors r+

PR #2500, opened 2021-03-22
Fix ADC 2.0 Bug

brghena:
The divide by two was left off in #2312 when updating to Tock 2.0

### Pull Request Overview

`app_buf_offset` is the buffer offset in bytes. `skip_amt` is used below as the number of `chunks_mut(2)` that get skipped, so it should be in units of "samples" (each of which are two bytes), which is why it should be set to `app_buf_offset / 2`. 

Later, in the code we call `app.app_buf_offset.set(app.app_buf_offset.get() + length * 2);`, which reinforces that `app_buf_offset` is a number of bytes, not a number of samples.

@hudson-ayers took a look at this and agreed it was a mistake. @alexandruradovici do you remember any reason why this would be intentional and I'm incorrect?

### Testing Strategy

Looking at it very carefully.

### TODO or Help Wanted

Should be tested on actual hardware.

### Documentation Updated

- [N/A] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/adc.rs
+++ b/capsules/src/adc.rs
@@ -938,7 +938,7 @@ impl<A: hil::adc::Adc + hil::adc::AdcHighSpeed> hil::adc::HighSpeedClient for Ad
                             });
                         }
 
-                        let skip_amt = app.app_buf_offset.get();
+                        let skip_amt = app.app_buf_offset.get() / 2;
 
                         {
                             let app_buf;


hudson-ayers (approved):

alexandruradovici:
I think you are right, this is a bug. When porting the capsule, I wrongly assumed chunks API did not exist anymore (I thought they were declared in the Grant API, turned out they were a Rust library function) and modified the code accordingly. When I realized that I could still use chunks, I modified the back the code. Most probably this one slipped.

Sorry for that :(

ppannuto (approved):
bors r+

brghena:
Definitely not your fault @alexandruradovici! It was great that you put in the effort to update the ADC. I'm glad you can confirm it wasn't intentional.

PR #2502, opened 2021-03-24
arch: Update arm cortex-m0, add cortex-m0p

alexandruradovici:
### Pull Request Overview

This pull request:
 * fixes some typos in `cortex-m0` asm!
 * updates `cortex-m` to mark functions as arm_v7m 
 * adds the `cortex-m0p` crate for Cortex M0+

### Testing Strategy

This pull request was tested using a Raspberry Pi Pico initial build

### TODO or Help Wanted

As my arm assembly knowledge is somehow limited, any feedback would be great.

Some more testing using different devices would be helpful

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m0p/src/lib.rs
+++ b/arch/cortex-m0p/src/lib.rs
@@ -22,17 +22,6 @@ pub use cortexm0::hard_fault_handler;
 pub use cortexm0::svc_handler;
 pub use cortexm0::systick_handler;
 
-extern "C" {
-    // _estack is not really a function, but it makes the types work
-    // You should never actually invoke it!!
-    fn _estack();
-    static mut _sstack: u32;
-    static mut _szero: u32;
-    static mut _ezero: u32;
-    static mut _etext: u32;
-    static mut _srelocate: u32;
-    static mut _erelocate: u32;
-}
 
 // Mock implementation for tests on Travis-CI.
 #[cfg(not(any(target_arch = "arm", target_os = "none")))]
--- a/arch/cortex-m0p/src/lib.rs
+++ b/arch/cortex-m0p/src/lib.rs
@@ -1,0 +1,45 @@
+//! Shared implementations for ARM Cortex-M0+ MCUs.
+
+#![crate_name = "cortexm0p"]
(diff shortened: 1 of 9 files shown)


arch/cortex-m0p/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+    // _estack is not really a function, but it makes the types work
+    // You should never actually invoke it!!
+    fn _estack();
+    static mut _sstack: u32;
+    static mut _szero: u32;
+    static mut _ezero: u32;
+    static mut _etext: u32;
+    static mut _srelocate: u32;
+    static mut _erelocate: u32;
+}
```
bradjc, line 36:
```suggestion
```

bradjc (commented):
I think those two deletions are ok.

alexandruradovici:
I'm not sure why the benchmark tests are failing. Is there any way I can run them locally?

hudson-ayers:
I think the benchmark CI is failing for the same reason the build CI is failing, because the benchmark PR relies on being able to build all boards. getting `make allboards` to pass locally should hopefully fix both -- currently it looks like at least the redboard_artemis_nano build fails

alexandruradovici:
> I think the benchmark CI is failing for the same reason the build CI is failing, because the benchmark PR relies on being able to build all boards. getting `make allboards` to pass locally should hopefully fix both -- currently it looks like at least the redboard_artemis_nano build fails

There was a typo in `cortex-m`, strangely `make prepush` passed locally which makes me think that it does not clean before rebuilding.

bradjc (approved):

hudson-ayers:
> There was a typo in `cortex-m`, strangely `make prepush` passed locally which makes me think that it does not clean before rebuilding.

`make prepush` runs `make format` and `make clippy` basically. `make clippy` does build all files, but there is an outstanding bug in clippy where it does not always catch errors if it has already been run once (https://github.com/rust-lang/rust-clippy/issues/4612) though it looks like if we update to a newer version of clippy that may now be fixed!

PR #2506, opened 2021-03-26
kernel: Reorganize process.rs

bradjc:
### Pull Request Overview

This PR splits up process.rs into:

- `kernel/src/`
  - `process.rs`: `trait Process` (renamed from `ProcessType`), and `struct ProcessId` (will be called `AppId` for now). Probably also the errors and other types.
  - `process_standard.rs`: `struct ProcessStandard` (renamed from `Process`).
  - `process_utilities.rs`: `load_process()`. Will likely include a non linked-list version of process discovery soon.
  - `process_policies.rs`: `ProcessRestartPolicy`. Will need to include a global restart policy at some point.

The only non-compiler-enforced changes beyond the process.rs reorg are:

- added some file-level comments
- moved AppId to process.rs (from upcall.rs)

~~I don't like this PR, but I'm looking at #2184 which is pretty much impossible to rebase/cherrypick so likely needs to be re-done, and I think this is the first step. If we are going to have `ProcessId`, it should be in process.rs, not upcall.rs.~~

~~All this PR does is copy code from one place to the other, and update all of the imports.~~

### Testing Strategy

travis

### TODO or Help Wanted

We can merge this as part of the full switch, but it is easier to look at in this form.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers:
Could this just be in its own file entirely? Maybe process_id.rs? process.rs is already too big and makes merges a nightmare

bradjc:
I don't like it, but I do agree. How far should we go? What about

- /kernel
  - /process
    - lib.rs (mostly empty)
    - process.rs: `ProcessType` renamed to `Process`, and `ProcessId`
    - userland_process.rs: `Process` renamed to `UserlandProcess`
    - support.rs: `load_processes`

And then `ProcessRestartPolicy` has to go somewhere.

hudson-ayers:
>   * /process
>     
>     * lib.rs (mostly empty)
sounds good
>     * process.rs: `ProcessType` renamed to `Process`, and `ProcessId`
+1 to trait and implementation in separate files. Indifferent about renaming `ProcessType`. I still think `ProcessId` could be its own file (just like it is now).
>     * userland_process.rs: `Process` renamed to `UserlandProcess`
Sounds good, still indifferent about renaming.
>     * support.rs: `load_processes`
+1
> And then `ProcessRestartPolicy` has to go somewhere.
`restart_policy.rs` sounds good to me.

phil-levis:
I'm kind of against creating a new package. When you have files that are tightly entwined, keeping them in a single directory is good. Otherwise you're moving back and forth between directories a lot and typing a lot of paths.

I like Brad's separation of the functionality. `process.rs` can have all of the trait and type definitions. And stuff like process restart policy. Another file (more below) can have our current process implementation. A third file, `loader` (or `process_loader`), can have `load_processes`. It's the code that turns a TBF into an in-memory process.

For the file that has the current process implementation, I'd shy away from userland, unless we expect there to be a kernel process type at some point in the future. Let's use an adjective that can be a counterpoint to what we might expect other processes to be. If we don't have a great sense of this then do something neutral, like `StandardProcess` or `BasicProcess`.

lschuermann:
`process.rs` being pretty huge got me more than once, and the separation proposed by @bradjc with @phil-levis makes a lot of sense to me. Thank you @bradjc for taking this one. I was already working on it when I saw you have recreated my changes of #2184 (which I agree, would've had to be redone completely anyways).

bradjc:
Updated plan:

- `kernel/src/`
  - `process.rs`: `trait Process` (renamed from `ProcessType`), and `struct ProcessId` (will be called `AppId` for now). Probably also the errors and other types.
  - `process_standard.rs`: `struct ProcessStandard` (renamed from `Process`).
  - `process_utilities.rs`: `load_process()`. Will likely include a non linked-list version of process discovery soon.
  - `process_policies`.rs`: `ProcessRestartPolicy`. Will need to include a global restart policy at some point.

When we settle on a structure and relevant PRs are merged or closed I can redo this PR, but it needs to be a quick turnaround due to how large of a diff this PR will be.

phil-levis:
Which PRs do you want to settle down before you move forward with this?

bradjc:
> Which PRs do you want to settle down before you move forward with this?

#2510

bradjc:
Ok! New version is here, take a look! I resisted any urge to change anything other than moving code, so there shouldn't be any functionality changes.

alevy (approved):

ppannuto (approved):

lschuermann (approved):
The only two comments I have are
- `ProcessStandard` might sound a bit like standard in the sense of norm, but I don't have a better name, and this really is a nitpick and I'm happy with the name.
- I'm wondering why the files aren't nested in a `process` module (e.g. `process::support`). This really doesn't make a different and so is also totally fine.

All in all, I think these changes help a lot and make (what previously was) `process.rs` much more manageable.

hudson-ayers (approved):

PR #2508, opened 2021-03-30
Replace `ReturnCode` with `Result<(), ErrorCode>`

hudson-ayers:
### Pull Request Overview

*Note: This description has been updated to reflect the new state of the PR*

This pull request removes the `ReturnCode` implementation from the Tock kernel and replaces it with `Result<(), ErrorCode>`.

It updates all uses of `ReturnCode` accordingly, e.g. `ReturnCode::SUCCESS` --> `Ok(())` and `ReturnCode::ENOMEM` --> `Err(ErrorCode::NOMEM)`.

It removes the `From` implementation which allowed for converting a `ReturnCode` into an isize/usize, and instead provides a function `retcode_into_usize(rc: Result<(), ErrorCode>)` for this purpose (converting to the negative values used previously to represent `ReturnCode`s. For now, it does not change the values passed in callbacks, but because all callbacks now use this function, it should be easy enough to automate the transition to passing positive `ErrorCode` values instead of negative `ReturnCode` values.

There are several next steps, discussed below. For now this PR just contains the changes that I was able to automate.

#### Next Steps: Questions
~~1. Should we keep the type alias, or should I find/Replace `ReturnCode` with `Result<(), ErrorCode>`? I kept the alias for now because changing it will add thousands of more changes to the PR.~~
~~2. Should I switch over to passing `ErrorCode` values in callbacks in this PR, or a follow-up PR?~~ Done in #2511 
~~3. In several places, we now have, basically, `Result<(), Result<(), ErrorCode>>` as a result of places that previously passed `Result<(), ReturnCode>`. Should I fix those now, or save all non-automated fixes for a separate PR? Some of these are in HILs and will require a decent amount of code changes on either side of the HIL.~~ Leaving for later.
~~4. To get this PR to pass CI, I had to pass `-A unused_must_use` to rustc in `Makefile.common`. Otherwise there are hundreds of warnings for unused `Result` types. Eventually I think we will want to require callers to explicitly use or drop these `Results`, but in some cases doing so will require large scale changes. I think we basically have two options: keep the allow flag passed to rustc, or add an explicit drop everywhere this warning is currently generated (e.g. `let _ = function_returning_result()`). The latter encourages better behavior in future code, but clutters existing code some.~~
5. Now the Tock kernel contains lots of non-idiomatic `Result` handling (e.g. manual handling of `Result` instead of using `?`). Fixing all this is not really feasible in the short term, but should we add a note somewhere about this?

#### Mechanism (For those curious)

The vast majority of these changes were made using a series of bash scripts. The workflow was basically as follows:
1. Replace `ReturnCode` with type alias.
2. `s/ReturnCode::ENOMEM/Err(ErrorCode::NOMEM)/g` (repeat for all ReturnCode variants)
3. Use another find/replace to fixup places where this led to `kernel::Ok(())` instead of just `Ok(())`, etc.
3. Run `cargo build` in top-level workspace, and then use bash scripts that parse the output to fixup imports. I had to rerun these commands several times, as each time would allow the compiler to get further through compilation, surfacing more errors that the scripts could then fix.
  - Find all errors for missing `ErrorCode` import, and insert the import on the line after the first import in that file
```bash
  cargo b 2> /tmp/err.txt && cat /tmp/err.txt | rg 'error\[E0433\].*ErrorCode' -A 1 | rg -v 'error\[E0433\].*ErrorCode' | rg -v '^\-\-$' | awk '{print $2}' | cut -f1 -d":" | awk '!seen[$0]++' > ~/files4.txt && cat ~/files4.txt | xargs sed -i -e '0,/^\(use .*;\)$/s//\1\nuse kernel::ErrorCode;/'
```
  - Replace all errors resulting from the now-missing `usize::From(...)` trait implementation by chasing error 0277 and find/replacing on those lines with an invocation of `retcode_into_usize`:
```bash
$ cat /tmp/err.txt | rg 'error\[E0277\].*usize' -A 1 | rg -v 'error\[E0277\].*ErrorCode' | rg -v '^\-\-$' | awk '{print $2}' | cut -f1,2 -d":" | awk '!seen[$0]++' > ~/files5.txt && cat ~/files5.txt | xargs -l ~/script.sh
```
The contents of `script.sh`:
```bash
#!/usr/bin/env bash

IN=$1
FILE=`echo $IN | cut -f1 -d":"`
LINE=`echo $IN | cut -f2 -d":"`

# basic Ok(()) case
sed -i "${LINE}s/\(usize\|From\)::from(Ok(()))/kernel::retcode_into_usize(Ok(()))/g" $FILE

# basic Err case, skip any with parens inside Err()
sed -i "${LINE}s/\(usize\|From\)::from(Err(\([^()]*\)))/kernel::retcode_into_usize(Err(\2))/g" $FILE

# variable case - allow anything without opening paren, bc if another opening paren
# then we will have balance issues
sed -i "${LINE}s/\(usize\|From\)::from(\([^(]*\?\))/kernel::retcode_into_usize(\2)/g" $FILE

# Ok(()).into()
sed -i "${LINE}s/Ok(())\.into()/kernel::retcode_into_usize(Ok(()))/g" $FILE

# Err(thing).into()
sed -i "${LINE}s/Err(\([^()]*\))\.into()/kernel::retcode_into_usize(Err(\1))/g" $FILE

#Anything .into()
sed -i "${LINE}s/\([a-zA-Z0-9_]*\)\.into()/kernel::retcode_into_usize(\1)/g" $FILE
```
After this I just manually chased a few scattered warnings and errors that weren't quite covered by the scripts, and ran `rustfmt`.

### Testing Strategy

### TODO or Help Wanted

This pull request still needs the above questions answered.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


bradjc:
> * Should we keep the type alias, or should I find/Replace `ReturnCode` with `Result<(), ErrorCode>`? I kept the alias for now because changing it will add thousands of more changes to the PR.

I'm still surprised Rust has `type` at all. I think that returning `Ok()` for a type of `ReturnCode` is unnecessarily confusing, and it is better to use the idiomatic `Result<>`. So, I say you should do the replace and that change should happen in this PR.


> * Should I switch over to passing `ErrorCode` values in callbacks in this PR, or a follow-up PR?

No? Is this for upcalls? If `ReturnCode` is gone, what is being passed now?

> * In several places, we now have, basically, `Result<(), Result<(), ErrorCode>>` as a result of places that previously passed `Result<(), ReturnCode>`. Should I fix those now, or save all non-automated fixes for a separate PR? Some of these are in HILs and will require a decent amount of code changes on either side of the HIL.

I agree this is weird, but doing number 1 will make it clear that it is weird, and can be fixed later.

> * To get this PR to pass CI, I had to pass `-A unused_must_use` to rustc in `Makefile.common`. Otherwise there are hundreds of warnings for unused `Result` types. Eventually I think we will want to require callers to explicitly use or drop these `Results`, but in some cases doing so will require large scale changes. I think we basically have two options: keep the allow flag passed to rustc, or add an explicit drop everywhere this warning is currently generated (e.g. `let _ = function_returning_result()`). The latter encourages better behavior in future code, but clutters existing code some.

I think that switching to result means dealing with the implications of that decision. So I don't think we should merge the compiler flag, and instead need to add `let _ =`.

> * Now the Tock kernel contains lots of non-idiomatic `Result` handling (e.g. manual handling of `Result` instead of using `?`). Fixing all this is not really feasible in the short term, but should we add a note somewhere about this?

Eh, I'm no rust expert, and I find dealing with `Result` is always challenging. https://doc.rust-lang.org/core/result/enum.Result.html just has too many functions, the examples are too terse, and there isn't a "So you want to..." guide (unless there is and I don't know about it). My take is that we could try to standardize and improve, but it will always be an uphill battle and we should instead focus that effort on subjective things related to tock, not rust. So, we can improve the result handling as we go, but no need to add another thing to the todo list.

hudson-ayers:
> I'm still surprised Rust has `type` at all. I think that returning `Ok()` for a type of `ReturnCode` is unnecessarily confusing, and it is better to use the idiomatic `Result<>`. So, I say you should do the replace and that change should happen in this PR.

Done.

> > * Should I switch over to passing `ErrorCode` values in callbacks in this PR, or a follow-up PR?
> 
> No? Is this for upcalls? If `ReturnCode` is gone, what is being passed now?

I am talking about #2320 basically. Right now libtock-c still expects negative values for errors, and this PR keeps it that way, by calling `retcode_into_usize(Result<(), ErrorCode>)` everywhere where a `ReturnCode` was previously passed in a callback.

hudson-ayers:
Note: per the size report, this reduces code size by ~100 bytes on each board

bradjc (dismissed):
This is a fantastic PR. We've had it the back of our minds for a while that we really should be using the Rust convention for error/success, and that our C backgrounds and starting Tock pretty early lead us to roll our own. However, no one _really_ wanted to start that undertaking other than trying to use Result appropriately going forward. But it turns out that just mechanically swapping `ReturnCode` for `Result<(), ErrorCode>` is a huge step in the right direction, and while a large change set, it is pretty straightforward.

While there is almost certainly numerous places where `Result<(), ErrorCode>` can be handled better with the rust machinery `Result` provides, this makes it a lot easier to make those changes going forward.

lschuermann (approved):
Skimming over the code this looks pretty good! It's pretty huge, but the changes seem to be very straightforward.

I've been pushing for this for a long time and I think the biggest advantage of this is that people get introduced to using `Result<_, ErrorCode>` and refrain from using `ReturnCode` (or can't any longer) which was one of the issues we had with the transition.

phil-levis:
I think that this, in a couple of cases, makes HILs not what we want, particularly those that had conditional buffer swaps. E.g.:

```rust
    fn transmit(
        &self,
        spi_buf: &'static mut [u8],
        frame_len: usize,
    ) -> (Result<(), ErrorCode>, Option<&'static mut [u8]>);
```

This should be

```rust
    fn transmit(
        &self,
        spi_buf: &'static mut [u8],
        frame_len: usize,
    ) -> (Result<(), (ErrorCode,  &'static mut [u8])>;
```

I'm OK with interfaces being not-quite-right and we'll fix them up, except for the HILs. I think we should fix the HILs in this PR (and update TRDs accordingly).

phil-levis:
I think the offending HIL methods are:

- [ ] `radio::RadioData::transmit`
- [ ] `adc::AdcHighSpeed::provide_buffer`
- [ ] `flash::Flash::read_page`
- [ ] `flash::Flash::write_page`
- [ ] `log::LogRead::read`
- [ ] `log::LogWrite::append`
- [ ] `symmetric_encryption::AES128::crypt`
- [ ] `text_screen::TextScreen::print`
- [ ] `uart::Transmit::transmit_buffer`
- [ ] `uart::receive::receive_buffer`, 

Additionally (but not to block this PR), these APIs do not correctly handle buffer swapping:  `bus8080::Bus8080`, `spi::SpiMaster::read_write_bytes`, `spi::SpiMasterDevice::read_write_bytes`, `spi::SpiSlave::read_write_bytes`,  `spi::SpiSlaveDevice::read_write_bytes`.

lschuermann:
>  I think that this, in a couple of cases, makes HILs not what we want, particularly those that had conditional buffer swaps.

Yes, I totally agree. I don't think we're technically making the situation worse with this, but using `Result` over `ReturnCode` shows some obvious weaknesses with the previous approach (where a failure is not necessarily coupled to actually return the passed buffer back). Great find!

I've taken the liberty to try and change one prominent HIL (UART) to use the proposed mechanism. Overall it's not too bad to work through and the changes are very mechanical. Whether to do it in this or another PR, I don't have strong feelings about :smile:. Feel free to pull [this head](https://github.com/lschuermann/tock/tree/dev/remove-returncode) in, I'll happily volunteer for some of the other HILs if my approach is deemed correct.

While porting this, I also noticed that `apollo3`, `sifive` and `swervolf-eh1`'s `uart::Receive` implementations (all of which are mock implementations) will silently drop the passed in buffer _and_ return an error. That's not possible anymore, so I've changed these to actually give the buffer back. Furthermore, there are a lot of places (in capsules) where errors of the UART HIL aren't handled correctly; though that's a disparate issue.

hudson-ayers:
@lschuermann When you posted this I had already made changes to a bunch of the HILs, including UART, using a semi-automated method that worked across a bunch at once. I did look through your commits to find places where our approach differed, and stole some of your better mechanisms in places.

My latest commit addresses all of the HILs that @phil-levis mentioned except `text_screen` and `flash`. **If anyone else can volunteer to take on those two that would be great!** 

My commit *does* update in-line code comments to reflect the new return values. However, it *does not* update the TRDs -- for both ADC and radio, the two relevant capsules with TRDs, the TRDs were so out of date that fixing them would have required rewriting much of the TRD to reflect the many code changes to those HILs since the TRD was last updated. I consider that outside the scope of this PR.

lschuermann:
@hudson-ayers Thanks for taking this on. Unfortunate I spent the time but glad to hear some of it helped. I did skim through the changes and they make sense, and again seem to be rather mechanical. I'll look into `text_screen` and `flash` and will post if I take them on. By the way: in these cases I do try to give proper attribution through the appropriate git trailers such as `Co-authored-by` & others (as done [here](https://github.com/lschuermann/tock/commit/06c3bfa91029c7a020db095898f302f970eae947), a [list of common trailers](https://git.wiki.kernel.org/index.php/CommitMessageConventions)).

hudson-ayers:
@lschuermann Done! Hope I got the syntax right.

I also added a commit to switchover documentation referencing EBUSY to instead reference BUSY, etc. It's imperfect, in some places it would be more correct to have said `Err(BUSY)`, but it is reasonably easy to understand either way, and find/replacing `EBUSY --> Err(BUSY)` also is wrong in some places where the Result returns additional items in the error case, or where the code is not being referenced as part of a result. If anyone wants to fix this up further, feel free.

bradjc:
@alevy help?

hudson-ayers:
I think all requested changes have now been addressed.

phil-levis (approved):

phil-levis:
Let's discuss this on the core call this week before merging.

bradjc (approved):

hudson-ayers:
Merged instead of rebasing because:

1. There are a lot of commits in this PR, rebasing would be tough
2. If people have partially reviewed, I don't want to completely lose their progress in the github UI

bradjc (approved):
We can nominate this PR for the "recovering C programmers" award.

lschuermann (approved):
Looked through most of the files, don't find anything looking weird or suspicious haha. It looks very mechanical, and about how I would've done it. :+1:

PR #2510, opened 2021-03-30
Rewrite the interface between process.rs and grant.rs

bradjc:
### Pull Request Overview

This PR is motivated by:

1. The interface for grants between process.rs (which owns the actual memory for the grant region) and grant.rs (which manages the interface for grants presented to capsules) is full of `unsafe` functions, and requires both sides to be very careful to ensure that all the unsafe calls are actually OK. Also the interface that process.rs exposes is very low level, and essentially provides direct memory access to a process's memory.
2. The protections to avoid entering a grant twice had a hole, and doing `app = grant.iter()` -> `grant.enter(appid)` -> `app.enter()` would still allow a grant to be entered twice.

The new interface encapsulates the unsafety and makes it clear what the expectations are. Process now better manages the grant memory and verifies that the memory is being handled correctly.

This turned into really a full re-write of the interface.

#### Specific Notable Changes

- Double grant entry is checked on `.enter()` and not when the grant is allocated. This makes the logic more robust and avoids loopholes like the one described above. To do this, process.rs now handles enter/leave tracking.
- `.iter_skip_unentered()` has been replaced with `.enter_if_unentered()`. Since double entry is tracked on enter, the iterator cannot guarantee it is providing all unentered grants. Note: it is very unlikely anyone should use `.enter_if_unentered()`. If you notice it in a PR you should ask for an explanation and a comment if one isn't there.
- `Borrowed` no longer has a (random) `.appid()` interface. It seemed strange to me that the actual grant memory had this bonus function, rather than just the AppliedGrant. This change required changing `.each()` to use the closure `fn (AppId, Borrowed)` so that capsules can still get the appid.
- Dynamic grants are managed by Process as well. Before, dynamic grants stored the raw pointer and an AppId. Having the AppId meant that grant.rs could check the process still exists before using the pointer. However, it could not check that the dynamic grant had not been free()d or if the process memory is still valid. Now, using a dynamic grant requires going through Process, so all relevant checks can occur.

### Testing Strategy

Running various apps.

I also created a capsule test which verifies that in three different cases the board panics if a double grant is entered.

### TODO or Help Wanted

I need help with Rust on arrays of dynamic grants. You can see it commented out in the diff. I couldn't figure out how to get the pointer casts to work correctly.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -30,7 +30,7 @@
 //! operation, they can use an `Allocator` to request additional memory from
 //! the process's grant region.
 //!
-//! ```
+//! ```skip
 //!                            ┌──────────────────┐
 //!                            │                  │
 //!                            │ Capsule          │
@@ -79,7 +79,7 @@
 //! Here is an overview of the types used by grant.rs to implement the Grant
 //! functionality in Tock:
 //!
-//! ```
+//! ```skip
 //!                         ┌──────────────────────────┐
 //!                         │ struct Grant<T> {        │
 //!                         │   number: usize          │
@@ -333,7 +333,7 @@ impl<'a, T: Default> ProcessGrant<'a, T> {
     /// bug in the code and not a condition that should be handled. For example,
     /// this benign looking code is wrong:
     ///
-    /// ```
+    /// ```skip
     /// self.apps.enter(thisapp, |app_grant, _| {
     ///     // Update state in the grant region of `thisapp`. Also, mark that
     ///     // `thisapp` needs to run again.
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -719,43 +719,42 @@ impl GrantRegionAllocator {
         Ok(CustomGrant::new(custom_grant_identifier, self.appid))
     }
 
-    // /// Allocates a slice of n instances of a given type. Each instance is
-    // /// initialized using the provided function.
-    // ///
-    // /// The provided function will be called exactly `n` times, and will be
-    // /// passed the index it's initializing, from `0` through `num_items - 1`.
-    // ///
-    // /// # Panic Safety
-    // ///
-    // /// If `val_func` panics, the freshly allocated memory and any values
-    // /// already written will be leaked.
-    // pub fn alloc_n_with<T, F>(
-    //     &mut self,
-    //     num_items: usize,
-    //     mut init: F,
-    // ) -> Result<CustomGrant<[T]>, Error>
-    // where
-    //     F: FnMut(usize) -> T,
-    // {
-    //     let (custom_grant_identifier, typed_ptr) = self.alloc_n_raw::<T>(num_items)?;
+    /// Allocates a slice of n instances of a given type. Each instance is
+    /// initialized using the provided function.
+    ///
+    /// The provided function will be called exactly `n` times, and will be
+    /// passed the index it's initializing, from `0` through `NUM_ITEMS - 1`.
+    ///
+    /// # Panic Safety
+    ///
+    /// If `val_func` panics, the freshly allocated memory and any values
+    /// already written will be leaked.
+    pub fn alloc_n_with<T, F, const NUM_ITEMS: usize>(
+        &mut self,
+        mut init: F,
+    ) -> Result<CustomGrant<[T; NUM_ITEMS]>, Error>
+    where
+        F: FnMut(usize) -> T,
+    {
+        let (custom_grant_identifier, typed_ptr) = self.alloc_n_raw::<T>(NUM_ITEMS)?;
 
-    //     for i in 0..num_items {
-    //         // # Safety
-    //         //
-    //         // The allocate function guarantees that `ptr` points to memory
-    //         // large enough to allocate `num_items` copies of the object.
-    //         unsafe {
-    //             write(typed_ptr.as_ptr().add(i), init(i));
-    //         }
-    //     }
+        for i in 0..NUM_ITEMS {
+            // # Safety
+            //
+            // The allocate function guarantees that `ptr` points to memory
+            // large enough to allocate `num_items` copies of the object.
+            unsafe {
+                write(typed_ptr.as_ptr().add(i), init(i));
+            }
+        }
 
-    //     // // convert `NonNull<T>` to a fat pointer `NonNull<[T]>` which includes
-    //     // // the length information. We do this here as initialization is more
-    //     // // convenient with the non-slice ptr.
-    //     // let slice_ptr = NonNull::new(slice_from_raw_parts_mut(typed_ptr.as_ptr(), num_items)).unwrap();
+        // // convert `NonNull<T>` to a fat pointer `NonNull<[T]>` which includes
+        // // the length information. We do this here as initialization is more
+        // // convenient with the non-slice ptr.
+        // let slice_ptr = NonNull::new(slice_from_raw_parts_mut(typed_ptr.as_ptr(), num_items)).unwrap();
 
-    //     Ok(CustomGrant::new(custom_grant_identifier, self.appid))
-    // }
+        Ok(CustomGrant::new(custom_grant_identifier, self.appid))
+    }
 
     /// Allocates uninitialized grant memory appropriate to store a `T`.
     ///
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -30,7 +30,7 @@
 //! operation, they can use an `Allocator` to request additional memory from
 //! the process's grant region.
 //!
-//! ```ignore
+//! ```text,ignore
 //!                            ┌──────────────────┐
 //!                            │                  │
 //!                            │ Capsule          │
@@ -79,7 +79,7 @@
 //! Here is an overview of the types used by grant.rs to implement the Grant
 //! functionality in Tock:
 //!
-//! ```ignore
+//! ```text,ignore
 //!                         ┌──────────────────────────┐
 //!                         │ struct Grant<T> {        │
 //!                         │   number: usize          │
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -302,7 +302,8 @@ impl<'a, T: Default> ProcessGrant<'a, T> {
         self.process.appid()
     }
 
-    /// Run a function with access to the contents of the grant region.
+    /// Run a function with access to the memory in the related process for the
+    /// related Grant.
     ///
     /// This is "entering" the grant region, and the _only_ time when the
     /// contents of a grant region can be accessed.
@@ -322,9 +323,9 @@ impl<'a, T: Default> ProcessGrant<'a, T> {
         self.access_grant(fun, true).unwrap()
     }
 
-    /// Run a function with access to the contents of the grant region only if
-    /// that grant region is not already entered. If the grant is already
-    /// entered silently skip it.
+    /// Run a function with access to the memory in the related process for the
+    /// related Grant only if that grant region is not already entered. If the
+    /// grant is already entered silently skip it.
     ///
     /// **You almost certainly should use `.enter()` rather than
     /// `.try_enter()`.**
@@ -381,8 +382,9 @@ impl<'a, T: Default> ProcessGrant<'a, T> {
         self.access_grant(fun, false)
     }
 
-    /// Run a function with access to the contents of the grant region and an
-    /// allocator for allocating additional memory in a process's grant region.
+    /// Run a function with access to the memory in the related process for the
+    /// related Grant. Also provide this function with an allocator for
+    /// allocating additional memory in the process's grant region.
     ///
     /// This is "entering" the grant region, and the _only_ time when the
     /// contents of a grant region can be accessed.
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -629,7 +629,7 @@ impl<T> CustomGrant<T> {
         }
     }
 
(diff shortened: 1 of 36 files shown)


kernel/src/grant.rs
```diff
@@ -72,190 +89,384 @@ impl<'a, T: Default> AppliedGrant<'a, T> {
...
+    /// in certain cases.
+    ///
+    /// Therefore, only use `enter_if_unentered()` if you are sure you want to
+    /// skip the already entered grant. Cases for this are rare.
+    ///
+    /// ## Return
+    ///
+    /// Returns `None` if the grant is already entered. Otherwise returns
+    /// `Some(fun())`.
+    pub fn enter_if_unentered<F, R>(self, fun: F) -> Option<R>
```
phil-levis, line 258: Perhaps `try_enter`? That is a common terminology for locked objects.

kernel/src/grant.rs
```diff
@@ -72,190 +89,384 @@ impl<'a, T: Default> AppliedGrant<'a, T> {
...
+    ///
+    /// This is "entering" the grant region, and the _only_ time when the
+    /// contents of a grant region can be accessed.
+    ///
+    /// Note, a grant can only be entered once at a time. Attempting to call
+    /// `.enter()` on a grant while it is already entered will result in a
+    /// panic!()`. See the comment in `access_grant()` for more information.
     pub fn enter<F, R>(self, fun: F) -> R
     where
         F: FnOnce(&mut Borrowed<T>, &mut Allocator) -> R,
```
phil-levis, line 196: What was the design rationale for passing `Allocator`? It seems like every use case is `|app, _|`.

kernel/src/grant.rs
```diff
@@ -3,29 +3,26 @@
...
+use core::ptr::{write, NonNull};
 
-use crate::process::{Error, ProcessType};
+use crate::process::{Error, ProcessDynamicGrantIdentifer, ProcessType};
 use crate::sched::Kernel;
 use crate::upcall::AppId;
 
-/// Type that indicates a grant region has been entered and borrowed.
-/// This is passed to capsules when they try to enter a grant region.
+/// This type provides access to the memory in a process's grant.
```
phil-levis, line 12: It seems like the file is missing a header/top comment that describes what the file contains and its role in the OS.

kernel/src/grant.rs
```diff
@@ -3,29 +3,26 @@
...
+///
+/// The grant must be entered to retrieve a `Borrowed` object. This object is
```
phil-levis, line 14: It seems like this should describe the type, rather than how you obtain it? Something like "A `Borrowed` object provides access to memory with a process's grant region. Borrow objects are typed by what a particular grant stores. Kernel code gains access to a `Borrowed` object by calling `Grant::enter`."

kernel/src/grant.rs
```diff
@@ -3,29 +3,26 @@
...
-    fn new(data: &'a mut T, appid: AppId) -> Borrowed<'a, T> {
-        Borrowed {
-            data: data,
-            appid: appid,
-        }
-    }
-
-    pub fn appid(&self) -> AppId {
-        self.appid
+    /// Create a `Borrowed` object to provide to a capsule. Only one can be
```
phil-levis, line 22: Not necessarily only to a capsule? Could it be that kernel services create grants too?

kernel/src/grant.rs
```diff
@@ -42,28 +39,48 @@ impl<'a, T: 'a + ?Sized> DerefMut for Borrowed<'a, T> {
     }
 }
 
+/// A grant applied to a particular process.
```
phil-levis, line 42: An instance of a memory grant for a particular process.

kernel/src/grant.rs
```diff
@@ -42,28 +39,48 @@ impl<'a, T: 'a + ?Sized> DerefMut for Borrowed<'a, T> {
...
+///
+/// This is created from a `Grant` to setup memory for that grant in a specific
+/// process's grant region. An `AppliedGrant` guarantees that the memory for the
+/// grant has been allocated in the process's memory.
```
phil-levis, line 46: We probably want to precisely define the difference between a "grant", a "grant region", and what's current an `AppliedGrant`. Perhaps grant region, grant, and grant instance?

kernel/src/grant.rs
```diff
@@ -42,28 +39,48 @@ impl<'a, T: 'a + ?Sized> DerefMut for Borrowed<'a, T> {
...
+
+    /// Used to keep the Rust type of the grant.
     _phantom: PhantomData<T>,
 }
 
 impl<'a, T: Default> AppliedGrant<'a, T> {
-    fn get_or_allocate(grant: &Grant<T>, process: &'a dyn ProcessType) -> Result<Self, Error> {
-        // Here is an example of how the grants are laid out in a
-        // process's memory:
+    /// Create an `AppliedGrant` for the given Grant and apply it to the given
```
phil-levis, line 59: What does it mean to "apply" a grant? Can you create a grant and not "apply" it?

kernel/src/grant.rs
```diff
@@ -72,190 +89,384 @@ impl<'a, T: Default> AppliedGrant<'a, T> {
...
+        self.process.leave_grant(self.grant_num);
+
+        Some(res)
     }
 }
 
-/// Grant which was dynamically allocated in a particular app's memory.
-pub struct DynamicGrant<T: ?Sized> {
-    data: NonNull<T>,
+/// Grant which was dynamically allocated from the kernel-owned grant region in
```
phil-levis, line 389: It might be useful to specify that regular grants are static, or call them StaticGrant?

phil-levis (commented):
Overall, I like these changes, they clean things up a lot.

However, I think there needs to be a cleanup/rewrite of the documentation. There's a lot of ambiguity and fuzziness on grant vs. grant region vs. a particular grant in a process grant region. There are lots of explanations in contrast to a regular grant (e.g., DynamicGrant) but no clear statement of what a regular (static) grant is or how it's allocated.

Finally, it would be good to benchmark how long it takes to enter a grant.

kernel/src/grant.rs
bradjc: Yes! This is very confusing.

kernel/src/grant.rs
bradjc: Apply currently means actually allocate memory from a specific process to "apply" that grant to that process.

kernel/src/grant.rs
bradjc: I avoided that because _all_ grants are dynamically allocated, and any grant allocation (AppliedGrant or DynamicGrant) can fail due to lack of memory.

kernel/src/grant.rs
phil-levis: Ah, yeah, that makes sense. I'd suggest `.enter_with_allocator`. If separating the two cases issues I think it's OK to pass the allocator; it would be useful if the documentation talked about it and when/how it is used.

kernel/src/grant.rs
phil-levis: The comment says "grant which was dynamically allocated", which implies there are grants which are **not** dynamically allocated? It seems like we need to figure out the terminology here.

bradjc:
I added some overview documentation to try to clarify some of the language. I haven't propagated those changes to the entire file (or made the other changes) but I will do so soon.

kernel/src/process.rs
```diff
@@ -920,6 +986,12 @@ pub struct Process<'a, C: 'static + Chip> {
     /// The process's memory.
     memory: &'static mut [u8],
 
+    /// Reference to the slice of pointers stored in the process's memory
+    /// reserved for the kernel. These pointers are null if the grant region has
+    /// not been allocated. When the grant region is allocated these pointers
+    /// are updated to point to the allocated memory.
+    grant_pointers: MapCell<&'static mut [*mut u8]>,
+
```
hudson-ayers, line 994:
```suggestion
    /// Reference to the slice of pointers stored in the process's memory
    /// reserved for the kernel. These pointers are null if the grant region has
    /// not been allocated. When the grant region is allocated these pointers
    /// are updated to point to the allocated memory. No other reference to these
    /// pointers exists in the Tock kernel.
    grant_pointers: MapCell<&'static mut [*mut u8]>,

```

hudson-ayers (commented):
A few nits. I have not reviewed the entirety of the changes in depth but overall this looks like a really desirable change that makes the interface much easier to understand (and much better documented)

bradjc:
I committed your suggestions.

Did you find the terminology sensible?

hudson-ayers:
I think the terminology is an improvement over the quite confusing terminology that exists now.

That said, I have a few concerns:

1. The difference between `Grant` and `ProcessGrant` is a little hazy from just the docs. I suggested an improvement in my commit below.
2. `CustomGrant` is a little weird as a name. First, `CustomGrant` is a per-process `Grant`, so `CustomProcessGrant` might be a better name to appropriately parallel our other grant type for which there is one-per-process. 
3. In general it is still not easy to determine the difference between a `ProcessGrant` and a `CustomGrant`. As far as I can tell, the difference is that each capsule gets a single `Grant` which allows that capsule to allocate (dynamically) exactly one `ProcessGrant` per process, and each `ProcessGrant` has to wrap the same type (which is determined by the `Grant` that generated it). Thus, without `CustomGrant`, a single `Grant` can only be used for a single one-time allocation per process. However, a `Grant` can create multiple `CustomGrant`s per process, and can create objects of different types in each `CustomGrant`, with those types not being restricted by the template type of the `Grant` that produced the `CustomGrant`. I am not sure the docs (or names) currently express this difference. I don't really have any better ideas for names, so just adding basically the text of this bullet might help.

I have a commit with some suggested improvements [here](https://github.com/tock/tock/commit/88cf0b0e31ca8a44ee70a4d4924d5cf6d7c317db), take them if you like.

One change I especially like is that this PR separates out use of the allocator into separate functions. In the future I think it might make sense to gate using those functions behind a capability - allowing a capsule to consume an arbitrary amount of memory from any process that uses it seems dangerous. Requiring a capability for that makes sense to me.

bradjc:
> A single `Grant` can only be used for a single one-time allocation per process.

Yes that is exactly correct.

> However, a `Grant` can create multiple `CustomGrants` per process.

`CustomGrant`s are not related to the `Grant` type, other than the way capsules get access to the allocator to create a custom grant is through calling functions implemented for `Grant`. That is, if I told you about a custom grant, you would have no way of determining which grant was originally used to get the allocator.

I know that `DynamicGrant` is a confusing name, but I don't have any great ideas either. I picked a name, but something better would be better.

hudson-ayers:
Also: if you still want to bring back `alloc_n_with`, I have a commit that does so using const generics here: https://github.com/hudson-ayers/tock/commit/ff19dc64dbc6dfb11757c73b2ba7a05440635be3 .

bradjc:
> Also: if you still want to bring back `alloc_n_with`, I have a commit that does so using const generics here: [hudson-ayers@ff19dc6](https://github.com/hudson-ayers/tock/commit/ff19dc64dbc6dfb11757c73b2ba7a05440635be3) .

My hero!

bradjc:
I think there might be two outstanding issues:

1. The name of `ProcessGrant`. Taking suggestions.
2. Whether to say "capsules" in the docs. I decided that trying avoid saying capsules makes the documentation unnecessarily obtuse, and I think saying "capsules" is actually consistent with @alevy's broad definition of "capsule".

bradjc:
Fixed merge conflict.

phil-levis:
Looks like it still has a merge conflict. Happy to approve it once that's resolved.

bradjc:
Fixed merge conflict after ReturnCode -> Result changes.

bradjc:
Any other comments?

hudson-ayers (dismissed):

phil-levis (approved):

bradjc:
Rebased to include changes to analog comparator.

kernel/src/grant.rs
```diff
@@ -1,266 +1,694 @@
...
+            .ok();
+
+        // # `unwrap()` Safety
+        //
+        // Only `unwrap()` if some, otherwise return early.
+        let grant_ptr = if optional_grant_ptr.is_some() {
+            optional_grant_ptr.unwrap()
+        } else {
+            return None;
+        };
```
lschuermann, line 498: If we wanted to get rid of the `unwrap()` here, using an if-let should be equivalent.
```suggestion
        let grant_ptr =
            if let Some(ptr) = optional_grant_ptr {
                ptr
            } else {
                return None;
            };
```

kernel/src/process.rs
```diff
@@ -261,6 +261,24 @@ pub fn load_processes<C: Chip>(
...
+/// memory later.
+///
+/// We use this type rather than a direct pointer so that any attempt to access
+/// can ensure the process still exists and is valid, and that the custom grant
+/// has not been freed.
+///
+/// The fields of this struct are private so only Process can create this
+/// identifier.
+#[derive(Copy, Clone)]
+pub struct ProcessCustomGrantIdentifer {
```
lschuermann, line 278: In the future (especially with the reorganization of `kernel::process`) one might want to look into making this a type parameter on the Process trait, but it seems totally fine to have that type used for all processes right now.

lschuermann (approved):
I like this architecture and the separation it provides. Also improves the naming and documentation a lot! My comments are just things I've stumbled over while looking through it, don't need to be addressed.

ppannuto (approved):

hudson-ayers (approved):

PR #2511, opened 2021-03-31
Define StatusCode which uses ErrorCode's numeric encoding

bradjc:
This switches error handling between the kernel and userspace to use the same number <-> error mapping as ErrorCode uses.

Part of #2320 and #2449

#### libtock-c changes

Userspace drivers need to be updated to use the new error numbers.

- [x] ble_advertising_driver
- [x] console
- [x] crc
- [x] hmac
- [x] ieee802154
- [x] max17205
- [x] udp
- [x] screen
- [x] text_screen

### Testing Strategy

todo

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/ble_advertising_driver.rs
+++ b/capsules/src/ble_advertising_driver.rs
@@ -469,7 +469,7 @@ where
 
                     if success {
                         app.scan_callback.schedule(
-                            kernel::retcode_into_usize(result),
+                            kernel::into_returncode(result),
                             len as usize,
                             0,
                         );
--- a/capsules/src/ble_advertising_driver.rs
+++ b/capsules/src/ble_advertising_driver.rs
@@ -469,7 +469,7 @@ where
 
                     if success {
                         app.scan_callback.schedule(
-                            kernel::into_returncode(result),
+                            kernel::into_statuscode(result),
                             len as usize,
                             0,
                         );
--- a/capsules/src/console.rs
+++ b/capsules/src/console.rs
@@ -338,7 +338,7 @@ impl uart::TransmitClient for Console<'_> {
                         app.write_remaining = 0;
                         app.pending_write = false;
                         app.write_callback
-                            .schedule(kernel::retcode_into_usize(return_code), 0, 0);
+                            .schedule(kernel::into_returncode(return_code), 0, 0);
                     }
                 }
             })
@@ -359,7 +359,7 @@ impl uart::TransmitClient for Console<'_> {
                                 app.write_remaining = 0;
                                 app.pending_write = false;
                                 app.write_callback.schedule(
-                                    kernel::retcode_into_usize(return_code),
+                                    kernel::into_returncode(return_code),
                                     0,
                                     0,
                                 );
@@ -440,7 +440,7 @@ impl uart::ReceiveClient for Console<'_> {
                                 };
 
                                 app.read_callback.schedule(
-                                    kernel::retcode_into_usize(ret),
+                                    kernel::into_returncode(ret),
                                     received_length,
                                     0,
                                 );
@@ -448,7 +448,7 @@ impl uart::ReceiveClient for Console<'_> {
                             _ => {
                                 // Some UART error occurred
                                 app.read_callback.schedule(
-                                    kernel::retcode_into_usize(Err(ErrorCode::FAIL)),
+                                    kernel::into_returncode(Err(ErrorCode::FAIL)),
                                     0,
                                     0,
                                 );
--- a/capsules/src/console.rs
+++ b/capsules/src/console.rs
@@ -338,7 +338,7 @@ impl uart::TransmitClient for Console<'_> {
                         app.write_remaining = 0;
                         app.pending_write = false;
                         app.write_callback
-                            .schedule(kernel::into_returncode(return_code), 0, 0);
+                            .schedule(kernel::into_statuscode(return_code), 0, 0);
                     }
                 }
             })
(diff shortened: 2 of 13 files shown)


hudson-ayers (commented):
This looks right to me. Maybe others can comment on whether `into_returncode()` is a good name.

phil-levis (approved):
Seems good to revisit the ReturnCode debate...

alistair23 (approved):

bradjc:
In an attempt to articulate how ErrorCode relates to ReturnCode relates to libtock-c return values, I added a section to the docs (copied below). The takeaway is `ReturnCode` is gone from the kernel. Since what returncode had become isn't returned from a function, it is now called `StatusCode` which is `ErrorCode`+Success. Anything named `*Code` is only a single usize mapping type. 

This frees up the name "ReturnCode" to be used in libtock-c (where it will be returned from functions).

added to docs:

### Naming Conventions

- `*Code` (e.g. `ErrorCode`, `StatusCode`): These types are mappings between
  numeric values and semantic meanings. These can always be encoded in a
  `usize`.
- `*Return` (e.g. `SyscallReturn`): These are more complex return types that can
  include arbitrary values, errors, or `*Code` types.

### Type Descriptions

- `*Code`s:
  - `ErrorCode`: A standard set of errors and their numeric representations in
    Tock.  This is used to represent errors for syscalls, and elsewhere in the
    kernel.
  - `StatusCode`: All errors in `ErrorCode` plus a Success value (represented by
    0). This is used to pass a success/error status between the kernel and
    userspace.

    `StatusCode` is a pseudotype that is not actually defined as a concrete Rust
    type. Instead, it is always encoded as a `usize`. Even though it is not a
    concrete type, it is useful to be able to return to it conceptually, so we
    give it the name `StatusCode`.

    The intended use of `StatusCode` is to convey success/failure to userspace
    in upcalls. To try to keep things simple, we use the same numeric
    representations in `StatusCode` as we do with `ErrorCode`.

- `*Return`s:
  - `SyscallReturn`: The return type for a syscall. Includes whether the syscall
    succeeded or failed, optionally additional data values, and in the case of
    failure an `ErrorCode`.

phil-levis:
> In an attempt to articulate how ErrorCode relates to ReturnCode relates to libtock-c return values, I added a section to the docs (copied below). 

This is a good formulation

phil-levis (approved):

lschuermann (approved):
This looks correct to me, and matches the userspace PR (https://github.com/tock/libtock-c/pull/198).

hudson-ayers (approved):
I rebased and updated the target branch to master

ppannuto (approved):

lschuermann (approved):

hudson-ayers:
https://github.com/tock/libtock-c/pull/198 is merging now, so...

bors r+

hudson-ayers:
merging manually because the netlify CI did not re-run when I changed the base branch. bors passed, so the netlify check did succeed on the merged commit.

jettr:
I am just checking out this change now, and I do like the use of Result<_,_> in kernel instead of ReturnCode. It feels much more rusty!

I want to ensure I am understanding this correctly: Success is 0, and failures are positive integers returned over the syscall interface to applications, right?

If so, is there a reason we did not alias the errors as negative numbers? It is a common convention to map errors to the negative number space, and it also allows for a APIs to return a meaning positive value, e.g. length of a read operation, without having to use another register to pass data across the syscall interface.

lschuermann:
> It feels much more rusty!

Don't you mean [Rustastic](https://users.rust-lang.org/t/name-for-idiomatic-rust/2772)? :smile: 

> Success is 0, and failures are positive integers returned over the syscall interface to applications, right?

Correct, that is the description of `StatusCode`, a concept not contained in a type in the kernel, but used to communicate these values over the system call boundary in the case of `Upcall`s.

> If so, is there a reason we did not alias the errors as negative numbers?

_Please don't view the statements below as authoritative, though this is what I remember from discussions about this issue._

Given the restructuring of the system call ABI, especially considering the [encoding of return values](https://github.com/tock/tock/blob/4eafc24b9533504a418495592809697d9084133e/doc/reference/trd104-syscalls.md#32-return-values) using 4 registers, we are explicitly using a separate register for encoding the return value. While previous versions of Tock did combine data and error encodings using positive and negative number spaces of `usize` types respectively, that can cause various issues. This change appears to have been initially motivated by [this](https://groups.google.com/g/tock-dev/c/jicoPcapHNU) `tock-dev` ML thread. In particular, by combining sucess values and errors in a single integer, you're effectively reducing representable data to `mem::size_of<usize>() * 8 - 1` bits (so for a 32-bit platform, 31-bit of space available). Unfortunately, there is no proper data type for a, say, 31-bit integer type in Rust. Combining an error with success values will therefore lead to an implicit or explicit cut off or error if the success data were to be larger than 31 bits. For these reasons, and the increased complexity in handling this between two code bases over a system call boundary, it has been decided to not combine success values and errors in a single register over the system call boundary. Using a separate register enables us to always use the full register width for success data.

Following that argument, the reason as to not use negative integers for encoding errors is more or less "because it isn't required". Using negative integers if we shan't use the positive number space hardly brings us any benefits, but makes us think about things such as sign extension, requires to convert the values back when debugging system calls, etc. As Rust userspace will represent success/error cases as `enum`s anyways, the only case where it might make sense to convert it back to the idiomatic C representation seems to be in a C-userspace, and negating a two's complement number is easy enough.

jettr:
Thank you for the detailed response and links! I do agree that having only 31 bits of meaningful success values has been a landmine in the code a few times. I can get behind separating the error codes from success now that we have more registers to pass information back from the kernel (I forgot that 2.0 expanded from 1 to 4 registers).

Good to know that Rustastic is the authoritative idiomatic rust term :)

PR #2512, opened 2021-04-01
rv32i: use atomic CSR instructions for rv32i::support::atomic

lschuermann:
### Pull Request Overview

This pull request contains a bunch of cleanly separated commits, enabling us to switch the `rv32i::support::atomic` to use RISC-V atomic CSR instructions.

Previously, the mstatus CSR was first read and then written to clear the machine mode interrupts. If an interrupt would arrive between these instructions, it could have changed mstatus, whereas this function would write back the previously read and modified register contents. Furthermore, if an interrupt handler would have cleared MIE between reading and writing the CSR, this function would set MIE again after execution of the passed in function.

As far as I can tell, this is not a real issue currently, due to the way interrupt handlers work and the fact that rv32i::support::atomic is already executed when the CPU is already in machine mode. However, it's best not to rely on this and use proper RISC-V instructions.

Additionally, this eliminates 7 instructions and one branch on rustc 1.52.0-nightly.

The other commits leading up to this are:

- riscv-csr: support atomic read/write & bit-field-set/clear ops

  Add atomic_replace, read_and_set_bits and read_and_clear_bits methods to the RISC-V CSR register wrappers to support the corresponding CSRRW, CSRRS and CSRRC RISV-C instructions.

  Using the atomic instructions can be required to prevent race conditions in interrupt contexts and context switches. Furthermore, using a combined read and write can often save instructions compared to a sequential read and write.

  The CSRR rd, csr and CSRW csr, rs1 instructions used in the RISC-V CSR register wrapper currently are assembler pseudo-instructions encoded as CSRRS rd, csr, x0 and CSRRW x0, csr, rs1 respectively (see RISC-V Specification v2.2, Volume I, 2.8 Control and Status Register Instructions).

- tock-registers: make Field::mask field public

  This partially reverts commit fed7a58 ("Make the Field::mask and FieldValue::mask fields private.") introduced through GitHub pull request tock#1939.

  In order to be able to implement atomic RISC-V CSR bit-field-set and bit-field-clear operations which operate on tock-register's Fields, access to the Field::mask field is required.

- tock-registers: fix Copy and Clone implementation on Field

  A Field contains a PhantomData over a generic R: RegisterLongName, and a PhantomData<R> is copyable regardless of whether R: Copy. However, using #[derive(Copy, Clone)] will generate

      #[automatically_derived]
      #[allow(unused_qualifications)]
      impl<T: ::core::marker::Copy + IntLike,
           R: ::core::marker::Copy + RegisterLongName>
              ::core::marker::Copy for Field<T, R> {}

  and hence Field will only implement Copy when R: Copy.

  Manually implementing Clone and Copy works around this issue. See rust-lang/rust#26925 for more information on this issue.

- riscv-csr: provide atomic read and set/clear field methods

  Add atomic read and set/clear methods for Fields on RISC-V CSR register wrappers. This wraps the read_and_set_bits and read_and_clear_bits into a more developer-friendly method that is compatible with the tock-registers types.

### Testing Strategy

This pull request was tested by compiling and looking at the generated assembly, as well as running litex/sim in Verilator and looking at the instruction trace.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/support.rs
+++ b/arch/rv32i/src/support.rs
@@ -23,16 +23,24 @@ pub unsafe fn atomic<F, R>(f: F) -> R
 where
     F: FnOnce() -> R,
 {
-    let original_mstatus = CSR.mstatus.extract();
-    if original_mstatus.is_set(mstatus::mie) {
-        CSR.mstatus
-            .modify_no_read(original_mstatus, mstatus::mie::CLEAR);
-    }
+    // Read the mstatus MIE field and disable machine mode interrupts
+    // atomically
+    //
+    // The result will be the original value of [`mstatus::mie`],
+    // shifted to the proper position in [`mstatus`].
+    let original_mie: usize = CSR
+        .mstatus
+        .read_and_clear_bits(mstatus::mie.mask << mstatus::mie.shift)
+        & mstatus::mie.mask << mstatus::mie.shift;
+
+    // Machine mode interrupts are disabled, execute the atomic
+    // (uninterruptible) function
     let res = f();
-    if original_mstatus.is_set(mstatus::mie) {
-        CSR.mstatus
-            .modify_no_read(original_mstatus, mstatus::mie::SET);
-    }
+
+    // If [`mstatus::mie`] was set before, set it again. Otherwise,
+    // this function will be a nop.
+    CSR.mstatus.read_and_set_bits(original_mie);
+
     res
 }
 
--- a/libraries/riscv-csr/src/csr.rs
+++ b/libraries/riscv-csr/src/csr.rs
@@ -212,4 +212,119 @@ impl<R: RegisterLongName, const V: usize> ReadWriteRiscvCsr<usize, R, V> {
     pub fn matches_all(&self, field: FieldValue<usize, R>) -> bool {
         field.matches_all(self.get())
     }
+
+    // Special methods only available on RISC-V CSRs, not found in the
+    // usual tock-registers interface
+
+    /// Atomically swap the contents of a CSR
+    ///
+    /// Reads the current value of a CSR and replaces it with the
+    /// specified value in a single instruction, returning the
+    /// previous value.
+    ///
+    /// This method corresponds to the RISC-V `CSRRW rd, csr, rs1`
+    /// instruction where `rs1 = in(reg) value_to_set` and `rd =
+    /// out(reg) <return value>`.
+    #[cfg(all(
+        any(target_arch = "riscv32", target_arch = "riscv64"),
+        target_os = "none"
+    ))]
+    #[inline]
+    pub fn atomic_replace(&self, val_to_set: usize) -> usize {
+        let r: usize;
+        unsafe {
+            asm!("csrrw {rd}, {csr}, {rs1}",
+                 rd = out(reg) r,
+                 csr = const V,
+                 rs1 = in(reg) val_to_set);
+        }
+        r
+    }
+
+    /// Atomically swap the contents of a CSR
+    ///
+    /// Reads the current value of a CSR and replaces it with the
+    /// specified value in a single instruction, returning the
+    /// previous value.
+    ///
+    /// This method corresponds to the RISC-V `CSRRW rd, csr, rs1`
+    /// instruction where `rs1 = in(reg) value_to_set` and `rd =
+    /// out(reg) <return value>`.
+    // Mock implementations for tests on Travis-CI.
+    #[cfg(not(any(target_arch = "riscv32", target_arch = "riscv64", target_os = "none")))]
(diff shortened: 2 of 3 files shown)


bradjc (dismissed):

lschuermann:
I feel like `atomic_replace`, `read_and_set_bits` and `read_and_clear_bits` justify a documentation comment. However, with the `#[cfg` attributes I wasn't sure which method to document: the RISC-V ones or the mock implementations? Documenting the riscv-ones will probably never generate actual rustdoc, whereas the mock implementations don't make sense to document.

arch/rv32i/src/support.rs
```diff
@@ -23,16 +23,20 @@ pub unsafe fn atomic<F, R>(f: F) -> R
...
+    // (uninterruptible) function
     let res = f();
-    if original_mstatus.is_set(mstatus::mie) {
-        CSR.mstatus
-            .modify_no_read(original_mstatus, mstatus::mie::SET);
+
+    // If the mstatus MIE field was set prior to clearing it, reenable
+    // machine mode interrupts
+    if original_mie != 0 {
+        CSR.mstatus.read_and_set_field(mstatus::mie);
```
lschuermann, line 37: We could also potentially get rid of this branch, by just using the `original_mie` shifted by mstatus::mie.shift as the bitmask to `read_and_set_bits` directly. Is that worth it, or should we keep using the `read_and_set_field` here (which looks a bit nicer)?

libraries/riscv-csr/src/csr.rs
```diff
@@ -212,4 +212,105 @@ impl<R: RegisterLongName, const V: usize> ReadWriteRiscvCsr<usize, R, V> {
...
+
+    #[cfg(all(
+        any(target_arch = "riscv32", target_arch = "riscv64"),
+        target_os = "none"
+    ))]
+    #[inline]
+    pub fn atomic_replace(&self, val_to_set: usize) -> usize {
+        let r: usize;
+        unsafe {
+            asm!("csrrw {rd}, {csr}, {rs1}",
```
hudson-ayers, line 227: I am not super familiar with the RISC-V ISA, but it is somewhat surprising to me that this compiles on earlgray despite it not having the A extension. Looking at the spec, I guess it is the case that the base instruction set includes atomic instructions for modifying CSRs, but not for general atomic operations?

lschuermann: I think this is because CSRs are core-local, so atomicity is much easier to enforce here, vs. on memory in general. But yes, as far as I understand every compliant rv32i core has these instructions, and the "regular" CSR reads & writes are just pseudo instructions for these atomic ones.

lschuermann: I also tested the csrrw, csrrs and csrrc on the VexRiscv without any extensions (just rv32i), but not this PR in its entirety yet, will do in a few hours.

hudson-ayers:
> I feel like `atomic_replace`, `read_and_set_bits` and `read_and_clear_bits` justify a documentation comment. However, with the `#[cfg` attributes I wasn't sure which method to document: the RISC-V ones or the mock implementations? Documenting the riscv-ones will probably never generate actual rustdoc, whereas the mock implementations don't make sense to document.

The approach taken in the `cortex-m` crate has generally been mixed, but I prefer the approach of including the same doc comment on both. This risks the two becoming out-of-sync, but at least includes the documentation in rustdoc and on the actual function

lschuermann:
Okay, I've added a documentation comment to all introduced functions. Furthermore, I've changed `rv32i::support::atomic` to directly use the `ReadWriteRiscvCsr::{read_and_set_bits, read_and_clear_bits}` functions instead of the convenience functions operating on `Field`s. This allows us to prevent unnecessary bitshifts and branches, where it's up to the compiler to optimize it (and it currently doesn't fully). Instead now the generated assembly uses immediates everywhere and has no conditional branches in the entire `rv32::support:atomic`, which is as efficient as we're going to get.

I've verified that this works, even on a CPU which does not have the RISC-V _A_ extension:
```
   0000e490     00800513        addi         x10,x0,8             x0:0x00000000 x10=0x00000008
   0000e494     30053473        csrrc        x8,mstatus,x10	 x10:0x00000008  x8=0x00000088
   <... the actual atomic inlined function>
   0000e4d8     00847513        andi         x10,x8,8             x8:0x00000088 x10=0x00000008
   0000e4dc     300525f3        csrrs        x11,mstatus,x10     x10:0x00000008 x11=0x00000080
```
(where 8 is the bitmask for the mstatus::mie bit)

alistair23 (approved):

bradjc (approved):

PR #2513, opened 2021-04-02
Add i.MX RT clock configuration

mciantyre:
### Pull Request Overview

This pull request adds ARM clock configuration for i.MX RT processors. The ARM clock speed can be configured using the CCM and CCM_ANALOG peripherals. As of this PR, the Teensy 4 board should run the ARM processor at 600MHz. The IPG clock should run at 25% of that speed.

To support variable processor speed, this PR also introduces a DCDC driver for the i.MX RT. See the data sheet for recommended voltage levels at each processor frequency.

### Testing Strategy

This pull request was tested in follow-on work. Based on this PR, we use the IPG clock as the GPT input. Using the periodic clock and GPT dividers, we divide the IPG clock to maintain a target `hil::time::Frequency` frequency. We observe a toggling GPIO at the expected rate. This should indicate that the IPG clock is running as configured, and that the ARM core is running as expected.

### TODO or Help Wanted

This PR does not apply the new features to the `imxrt1050-evk` board. The 1050 evaluate board could benefit from this work at a later time.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


chips/imxrt10xx/src/ccm.rs
```diff
@@ -1,44 +1,60 @@
...
-    ccgr4: ReadWrite<u32, CCGR4::Register>,
-    // clock gating register 5
-    ccgr5: ReadWrite<u32, CCGR5::Register>,
-    _reserved6: [u8; 8],
+register_structs! {
+    /// Clock Controller Module
+    CcmRegisters {
+        /// CCM Control Register
+        (0x000 => ccr: ReadWrite<u32, CCR::Register>),
+        (0x004 => _reserved0),
```
lschuermann, line 10: I think the idea behind the `(<address> => <type>)` syntax is that we don't need the reserved fields here.

chips/imxrt10xx/src/ccm_analog.rs
```diff
@@ -0,0 +1,625 @@
...
+        (0x000 => pll_arm: RegisterSCT<PLL_ARM::Register>),
+        /// Analog USB1 480MHz PLL Control Register
+        (0x010 => pll_usb1: RegisterSCT<PLL_USB1::Register>),
+        /// Analog USB2 480MHz PLL Control Register
+        (0x020 => pll_usb2: RegisterSCT<PLL_USB2::Register>),
+        /// Analog System PLL Control Register
+        (0x030 => pll_sys: RegisterSCT<PLL_SYS::Register>),
+        /// 528MHz System PLL Spread Spectrum Register
+        (0x040 => pll_sys_ss: ReadWrite<u32, PLL_SYS_SS::Register>),
+        (0x044 => _reserved0),
```
lschuermann, line 52: Same here.

lschuermann (commented):
This looks pretty good, thanks for the contribution!

I have an i.MX RT1060 Evaluation Kit lying around which I wanted to get up & running for some time now, and have access to it in a few days. If you'd like someone else to confirm these changes I can offer to test it there (assuming I get it working), but we don't need to block on this.

chips/imxrt10xx/src/ccm_analog.rs
alexandruradovici: @lschuermann are you sure about this? I don't think the macro will output the reserved space without this.

```
(0x048 => clk_peri_ctrl: ReadWrite<u32, CLK_PERI_CTRL::Register>),
// (0x04C => _reserved0),
/// Indicates which src is currently selected (one-hot)
(0x050 => clk_peri_selected: ReadOnly<u32, CLK_PERI_SELECTED::Register>),
```
produces
```
#[doc =
    r" Clock control, can be changed on-the-fly (except for auxsrc)"]
clk_peri_ctrl: ReadWrite<u32, CLK_PERI_CTRL::Register>,
#[doc = r" Indicates which src is currently selected (one-hot)"]
clk_peri_selected: ReadOnly<u32, CLK_PERI_SELECTED::Register>,
```

lschuermann: I am not actually, I've never used the syntax myself! I actually didn't see the point in the syntax if it _wouldn't_ work, but given the macro output you're obviously right. Thanks for the hint!

lschuermann: Yeah, sorry, I've just now realized my mistake: the syntax acutally does make sense, because it automatically sizes the reserved fields correctly, though doesn't get rid of them entirely. `proc_macros` probably aren't smart enough to autogenerate the reserved field names (except maybe with some really clever ident concat hacks which we probably don't really want to use anyways) :smile:.

chips/imxrt10xx/src/ccm.rs
lschuermann: I thought wrong! :smile:

mciantyre:
Thanks for the look over. Having another board implementation for the 1060 EVK would be great. Let me know if you need any help with bring-up.

bradjc (approved):

hudson-ayers (approved):
bors r+

PR #2516, opened 2021-04-02
Add QEMU riscv32 "virt" board

lschuermann:
### Pull Request Overview

This pull request adds support for the `qemu-system-riscv32` "virt" machine type, ~as well as initial support for memory mapped VirtIO devices (VirtIO over MMIO) with split virtqueues~ (VirtIO support will be provided in a separate follow-up PR).

The `qemu-system-riscv32 -machine virt` is an interesting target, because it
- does not try to mimic any real-world device, but is created solely for virtualization
- therefore, does not have the virtually imposed hardware limitations and inflexible resource allocation and memory layout as other targets (such as the `hifive1`)
- exposes VirtIO devices, either via the emulated PCI bus or via VirtIO over MMIO transports

VirtIO support in Tock can be interesting because it supports various device types that can be used for testing larger parts of the kernel and/or userspace in CI environments, or for development. While QEMU does not emulate targets as precise as other simulators such as Renode or emulate actual hardware through tools like Verilator, it is very fast and convenient for development. Specifically, devices such as a network card or a block device can be used for testing the respective capsules, virtualizers and userspace applications.

The `qemu_rv32_virt` board and `qemu_rv32_virt_chip` are very similar to `hifive` and `e310x` respectively, because the QEMU `virt` target uses much of the infrastructure for `sifive_e` chips. It's unfortunate that this introduces some duplicated code (especially `plic.rs`), but I expect these to deviate further in the future. Also, because `virt` is the officially recommended target for running Linux on QEMU RISC-V, I expect it to receive better and longer support than `hifive1`.

### Testing Strategy

This pull request was tested by running
```
qemu-system-riscv32
  -machine virt \
  -bios $TOCK_DIR/target/riscv32imac-unknown-none-elf/release/qemu_rv32_virt.bin
  -global virtio-mmio.force-legacy=false
  -netdev tap,id=n0,script=no,downscript=no -device virtio-net-device,netdev=n0
  -device virtio-rng-device
```
on QEMU v5.2. The libtock-c example `tests/rng` seems to work fine. The `VirtIONet` dev works in my various applications just fine.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


lschuermann:
Rebased to incorporate changes to the rv32i mtimer implemented in the CLINT of PR #2486.

boards/qemu_rv32_virt/src/io.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/// Panic handler.
+#[cfg(not(test))]
+#[no_mangle]
+#[panic_handler]
+pub unsafe extern "C" fn panic_fmt(pi: &PanicInfo) -> ! {
+    let writer = &mut WRITER;
+
+    debug::panic::<DummyLed, _, _>(&mut [], writer, pi, &rv32i::support::nop, &PROCESSES, &CHIP)
```
alistair23, line 59: You should just do what SweRVolf does here and not require the dummy LEDs: https://github.com/tock/tock/blob/master/boards/swervolf/src/io.rs#L40

boards/qemu_rv32_virt/src/main.rs
```diff
@@ -0,0 +1,431 @@
...
+// impl capsules::virtio::devices::virtio_net::VirtIONetClient for VirtIONetDummyClient {
+//     fn packet_received(&self, id: usize, buffer: &'static mut [u8], len: usize) {
+//         debug!("Packet received from virtio-net if {}, len {}.", id, len);
+//         self.0.return_rx_buffer(buffer);
+//     }
+
+//     fn packet_sent(&self, _id: usize, _buffer: &'static mut [u8]) {
+//         unimplemented!("VirtIONetDummyClient packet_sent");
+//     }
+// }
```
alistair23, line 56: Can this just be removed? Commented out code is prone to bit rotting.

chips/qemu_rv32_virt_chip/src/plic.rs
```diff
@@ -0,0 +1,147 @@
...
+    pending: [ReadWrite<u32>; 2],
+    _reserved2: [u8; 4088],
+    /// Interrupt Enable Register
+    enable: [ReadWrite<u32>; 2],
+    _reserved3: [u8; 2088952],
+    /// Priority Threshold Register
+    threshold: ReadWrite<u32, priority::Register>,
+    /// Claim/Complete Register
+    claim: ReadWrite<u32>,
+}
```
alistair23, line 29: You can't copy the PLIC from the SiFive chip?

chips/qemu_rv32_virt_chip/src/uart.rs
```diff
@@ -0,0 +1,468 @@
+//! QEMU's memory mapped 16550 UART
```
alistair23, line 1: Tock already has a 16550 compatible UART, can you not use that?

alistair23 (commented):
I had a quick look and overall this looks good.

Maybe it would be worth splitting out the QEMU machine and the virtIO drivers into separate PRs. That way we can focus on the board first and then the other stuff second.

boards/qemu_rv32_virt/Makefile
```diff
@@ -0,0 +1,7 @@
+# Makefile for building the Tock kernel for the qemu-system-riscv32
+# virt machine type
+
+TARGET=riscv32imac-unknown-none-elf
+PLATFORM=qemu_rv32_virt
+
+include ../Makefile.common
```
alistair23, line 7: Can you add the commands to run this here as well?

chips/qemu_rv32_virt_chip/src/chip.rs
```diff
@@ -0,0 +1,279 @@
...
+use rv32i;
+use rv32i::csr::{mcause, mie::mie, mip::mip, CSR};
+use rv32i::pmp::PMP;
+
+use crate::interrupts;
+use crate::plic::Plic;
+use crate::plic::PLIC;
+use kernel::InterruptService;
+
+type QemuRv32VirtPMP = PMP<4>;
```
alistair23, line 19: QEMU has 16 PMP regions

boards/qemu_rv32_virt/src/io.rs
lschuermann: I'm not sure about this. On the one hand, the SweRVolf implementation already differs from the `kernel::panic` implementation, at least in non-substantial ways, for instance by not flushing the debug buffer (if I'm not mistaken). Further, I'd rather not have this deviate further from what the kernel does. Maybe it's best to have a `panic` and a `panic_returning` where one does not produce the infinite loop blinking?

boards/qemu_rv32_virt/src/main.rs
lschuermann: I don't expect it to stay in here long. It's useful if you just want to test the network interface's basic functionality until we have a proper HIL and driver/stack.

chips/qemu_rv32_virt_chip/src/plic.rs
lschuermann: To be sure I understand what you're proposing. I should add the sifive chip as a dependency to this and then just use it from there? Sure, I can do that. It's technically slightly weird to have the `qemu_rv32_virt_chip` dependent on this crate but QEMU seems to have a similar-ish architecture (by just reusing the SiFive PLIC implementation).

boards/qemu_rv32_virt/src/io.rs
lschuermann: Opened #2534 about this. Let's see what comes out of that.

chips/qemu_rv32_virt_chip/src/plic.rs
alistair23: Yep, that's what I'm proposing. I don't think that it's weird. That is why the RISC-V chips are all split up, to allow sharing vendor IP between chips.

QEMU uses the SiFive PLIC internally (which is now almost a RISC-V standard) so it makes sense to reuse it.

boards/qemu_rv32_virt/src/io.rs
lschuermann: Given that #2534 is merged I'll throw the `DummyLed` out. Thanks for the hint!

chips/qemu_rv32_virt_chip/src/plic.rs
lschuermann: Okay, I know remember why I've copied it over: this implementation is currently in the `e310x` chip-crate, not in `sifive`, and so I've mirrored that setup. I will move it to the `sifive` crate accordingly and change both chips to use that instead.

chips/qemu_rv32_virt_chip/src/chip.rs
lschuermann: Changed accordingly. Can you tell me where that is defined though? I couldn't figure it out, at some point literally grepped for `16` in the (slightly outdated) QEMU source tree.

chips/qemu_rv32_virt_chip/src/uart.rs
lschuermann: Huh, that's interesting. It even took me a while to find the file you're referring to. I didn't try to reinvent the wheel, rather parts of this code have already been written >6 months ago collecting dust and so this was developed entirely independent. I think both implementations have some distinct advantages (for instance, this has extensive `register_bitfields!` and less bitmasks, whereas yours uses `register_structs!`) and are in general very similar. I'd suggest I try merging them and putting them in a common place (given that they aren't bound to a specific chip), if that's okay with you? I'd also be integrating read as I'm close to have that working locally.

alistair23: Yep, that sounds good to me

chips/qemu_rv32_virt_chip/src/chip.rs
alistair23: `MAX_RISCV_PMPS` is what you are looking for. It's in `cpu.h`

boards/qemu_rv32_virt/Makefile
lschuermann: Done.

chips/qemu_rv32_virt_chip/src/uart.rs
lschuermann: Okay, I spent some significant time on getting the SweRVolf running (wouldn't work on Nix and requires a source-compiled Verilator on Debian as the package is too outdated) and merging these two. It's significantly more difficult than I expected (without ending up with a barely usable interface) because while these are very similar, the memory mapped interfaces aren't the same: QEMU presents the (usually exposed on an x86 I/O bus) "registers" as a byte-packed array in memory, whereas the SweRVolf puts them in 32-bit wide memory locations. I solved that issue by treating them all like `u8` registers and having a configurable const-generic based offset between the registers.

... but it still doesn't work. I can't trace down as to why exactly, but also didn't spent much time on debugging yet. I'd say that it might be much more reasonable to keep these separate for now and care about merging them later? In a separate PR we can then also resolve the (for now unanswered) question as to where shared peripherals between chips across vendors are stored.

lschuermann:
> Maybe it would be worth splitting out the QEMU machine and the virtIO drivers into separate PRs. That way we can focus on the board first and then the other stuff second.

Thanks for the review! Back when I started working on this I asked on a call whether people would be interested to see this immediately or I should rather provide it integrated into an out-of-the-box runnable board and was specifically told to open a PR as a single package. I _can_ split it up again, but it's already cleanly separated into different commits and quite frankly, the `virt` machine doesn't really make much sense without VirtIO IMHO.

The VirtIO implementation is only a start, but already working quite well (used it for a few days as an Ethernet crossover cable on emulated an x86-64 with >100Mbit/s throughput and >100GiB transmitted). I'll add some documentation for it in the next few days which should help reviewing and transition this PR out of the draft phase.

chips/qemu_rv32_virt_chip/src/uart.rs
alistair23: Hmmm... That's annoying that they are treated differently. I agree generics sound like the best way to handle this.

I don't see why we need two 16550 compatible UARTs in Tock though. 16550 compatible UART is pretty common so including one that can be used by multiple chips/boards seems really useful. It seem reasonable to have a "generic" chip with support for the UART.

lschuermann:
This PR is not abandoned, but I'm currently spending my time I have for Tock on working towards 2.0.

lschuermann:
Alright, finally got back to this. :)
I've rebased on latest master, everything seems to work still. I'll write some documentation and clean up the codebase, possibly also avoiding a reimplementation of the 16550 by reconciling it with the SweRVolf implementation. When everything's ready, I'll split out the VirtIO implementation, which will make this PR a lot simpler to review (it's very similar to the HiFive1 / E310 implementation then).

boards/qemu_rv32_virt/src/io.rs
```diff
@@ -0,0 +1,45 @@
...
+    let writer = &mut WRITER;
+
+    debug::panic_print::<_, _>(writer, pi, &rv32i::support::nop, &PROCESSES, &CHIP);
+
+    // The system is no longer in a well-defined state, so loop
+    // endlessly.
+    //
+    // If possible we might want to exit the LiteX simulation in the
+    // future instead.
+    loop {}
```
alistair23, line 44: We could use semi-hosting calls here and exit QEMU

bradjc:
Why did this not make it across the finish line? Should we still pursue this?

lschuermann:
> Why did this not make it across the finish line? Should we still pursue this?

@bradjc Both this and the SweRV chips use a 16550 "compatible" UART, however there are some differences with respect to the implementation's behavior and how they are memory mapped. Subjectively speaking, this PR contains a more complete implementation of the 16550 UART and I wanted to get this working on the SweRV Verilated board as well, but really don't know how to best debug this (spent a few days on it but couldn't get it to work in the end).

If we're fine with having some code duplication for a similar peripheral between two chips, we can easily get this merged (first without the VirtIO drivers) and then introduce VIrtIO support in a second PR (that needs a good look still, but was generally rather feature-complete).

I still think that VirtIO support would drastically improve the virtual Tock story and is a sensible transport also for other FPGA targets (which can use the memory-mapped VirtIO by attaching to the emulated target's memory bus).

alistair23:
In the short run some code duplication is ok. But I think we should really push to have a single 16550 compatible UART. I have seen a few other MCUs with 16550 UARTs and it would be great if we can just expose a single working driver, instead of duplication's everywhere.

I'm happy to merge this without the VirtIO drivers though. If we do can we open a tracking issue for the 16550 UART.

Improvements to virtual Tock would be great! The OpenTitan board gets us a fair step of the way, but the more boards we can run on the better. If we could get a virt arm board as well we can finally CI test both architectures!

lschuermann:
I removed the VirtIO support in this PR and will provide a follow-up for this. Furthermore, I suggest we merge this despite the duplicate 16550 UART driver. The PR should be pretty uncontroversial now, as the board introduced is _very_ similar to the included HiFive1 already :slightly_smiling_face:.

boards/qemu_rv32_virt/Makefile
```diff
@@ -0,0 +1,60 @@
...
+	qemu-system-riscv32 \
+	  -machine virt \
+	  -bios $^ \
+	  -global virtio-mmio.force-legacy=false \
+	  -device virtio-rng-device \
+	  -nographic
+	  @# attaching a TAP network device requires proper permissions
+	  @# to create a tuntap device or access to an existing device
+	  @# -netdev tap,id=n0,script=no,downscript=no
+	  @# -device virtio-net-device,netdev=n0
```
alistair23, line 43: We can do networking without TAP, but let's leave that for the future

alistair23 (dismissed):

bradjc (commented):
Looks good, just needs a board readme and a description on the chip readme.

chips/sifive/src/plic.rs
```diff
@@ -0,0 +1,143 @@
+//! Platform Level Interrupt Control peripheral driver.
```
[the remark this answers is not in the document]
lschuermann: I can see that argument and agree that we should in the long run move it to the `arch/riscv` crate, but I'd rather move it once we're confident it works and is integrated with OpenTitan.

boards/qemu_rv32_virt/Makefile
lschuermann: Right, for outgoing connections that works fine (using QEMU's userspace networking functionality, right?) but for incoming connections it wouldn't. My dummy/demo usecase for this is going to be running a webserver, which is great because you have some static code running in the QEMU machine and can interact from within your own OS / browser. This is why I decided to go with TAP at first, in addition to it being easier to debug.

boards/README.md
```diff
@@ -33,6 +33,7 @@ that Tock supports.
 | [LiteX on Digilent Arty A-7](litex/arty/README.md)                   | RISC-V RV32I    | LiteX+VexRiscV | custom     | custom         | No            |
 | [Verilated LiteX Simulation](litex/sim/README.md)                    | RISC-V RV32I    | LiteX+VexRiscv | custom     | custom         | No            |
 | [ESP32-C3-DevKitM-1](esp32-c3-devkitM-1/README.md)                   | RISC-V-ish RV32I| ESP32-C3       | custom     | custom         | No            |
+| [QEMU RISC-V 32 bit `virt` platform](qemu_rv32_virt/README.md)       | RISC-V RV32IMAC | QEMU           | custom     | custom         | Yes (duh)     |
```
lschuermann, line 36: Meh, I thought that we were in a good enough state to just say "yeah, sure, QEMU works with this board", but looking that the [downstream distribution situation](https://repology.org/project/qemu/versions) we really are better off specifing a version here. I'm still unsure though, is this supposed to be a minimum-supported or tested version? We should make the expectations clear, the `virt` target introduced here worked without issues on v5.2.0 and still works on v7.0.0. The Makefile contains a "tested" version and prints that, but that does not necessarily correspond to the minimum supported one. @alistair23 what do you think?

boards/qemu_rv32_virt/Makefile
alistair23: You can setup port forwarding with the userspace networking. So you should still be able to host a webserver (ICMP doesn't work though)

boards/README.md
alistair23: I don't think we should worry about specifying a version here.

As long as QEMU has RISC-V support (which it has for years) then the virt machine will work. In the future I expect the virt machine itself to be versioned, and we can then specify version there. But I don't think we need to say a minimum or maximum QEMU version

lschuermann: Okay, cool, then I'll just leave the README like that.

boards/qemu_rv32_virt/Makefile
lschuermann: That's good to know. Given this doesn't add any networking in this PR, I'd say we change the comment once VirtIO is introduced proper.

hudson-ayers (approved):
bors r+

PR #2517, opened 2021-04-04
[RFC] tock-registers: use traits for accessing / manipulating registers

lschuermann:
### Pull Request Overview

**This includes a lot of mechanical changes (adding imports). I suggest reviewing this commit first: https://github.com/tock/tock/commit/900daa36e788aa**

Replace the individual implementations of common methods on registers (e.g. `get`, `set`, `read`, `modify`) by traits which represent the respective access level (`Readable`, `Writeable`, `ReadWriteable`).

This significantly reduces code duplication found in tock-registers by providing default implementations for most methods, as well as allowing to implement custom register types adhering to the tock-registers interface. Furthermore, generic implementations of peripherals can be designed to take trait objects instead of concrete register types, to be agnostic over their actual register (e.g. some register mapped to either a RISC-V CSR or a memory location).

The `Readable`, `Writeable` and `ReadWriteable` manipulation traits have associated type parameters specifying the underlying integer type and `RegisterLongName`. Associated type parameters are used, as every monomorphized register over precise `IntLike` and `RegisterLongName` types should have at most a single implementation of the respective manipulation traits. If generic traits were used, multiple manipulation traits could be implemented on a single monomorphized register type.

Every readable register type implements `Readable`, which requires the `get`-method to be implemented manually. Every writeable register type implements `Writeable`, which requires the `set`-method to be implemented manually. Registers which implement both `Readable` and `Writeable`, and have the same associated `IntLike` and `RegisterLongName` types for both of their Readable and Writeable implementations also implement `ReadWriteable`. For Aliased registers (where read and write operations correspond to different registers), `ReadWriteable` is therefore in general not implemented.

Potential drawbacks of this approach include

- requiring to import `Readable`, `Writeable` and `ReadWriteable` to be able to use the respective methods

- having a standard interface, which does not allow deviation from (I suppose this is also a big advantage)

  For instance, LocalRegisterCopy used to require a mutable reference for methods modifying its content. This does not match the interface of other registers. Hence, it now uses interior mutability through a Cell<T>. This has the additional implication of not supporting the Copy trait any longer.

As an example of how this comes in handy in practice, the RISC-V CSR library is changed to use these traits instead, and hence now automatically implement most methods, and are guaranteed to stay in sync with the tock-registers API described by those traits. Further usecases in and around Tock exist, for instance the LiteX register interface, using tock-registers on `u8`-slices in transport protocol headers, and with other projects using the library, being able to easily implement their own register types with different read / write strategies.

### Testing Strategy

This pull request was tested by compiling and running in a LiteX simulation (Verilator). Also, the binary size of boards does not seem to change.

### TODO or Help Wanted

This pull request still needs comments about whether we want this.

If this were merged, it'd be a breaking API change for tock-registers, which (assuming we follow SemVer) require a major version bump in the next release.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


bradjc:
I'm indifferent about this PR.

alistair23:
Having to import `Readable`, `Writeable` and `ReadWriteable` seems like a hassle and with no size reduction I wouldn't say it's worth it.

hudson-ayers:
I think that the reduced code duplication is pretty nice, and think this represents a good "rusty" way to handle that duplication. On the other hand I agree its a shame to have to add all those imports, and think `LocalRegisterCopy` using a `Cell<>` internally is a bit of a shame. So in general I am also somewhat indifferent here.

bradjc:
I agree on the imports, it's not user friendly to have to explicitly include something which is never referenced, and even worse when the compiler tells you _exactly how to fix it_ yet doesn't just do it for you. But, that is a price of using Rust, and not something we really control.

lschuermann:
Yeah, the imports aren't nice but an artifact I don't one can get around when using methods in traits. I tried to give a complete overview in the PR comment above, but to emphasize my primary motivation with this: tock-registers can already be downloaded on crates.io and the `register_bitfields!` (and related) infrastructure provide great ergonomics and are rather useful. While it is technically possible to use these macros and the generated code (i.e. `Field`s, `FieldValue`s) in other code without these changes, having to essentially reimplement the register structures along with all of their accessors and manipulators is a bummer. It's pretty cool to just have to implement either one or two methods (depending on access level) on a custom register type, and get the rest of the infrastructure plus consistency with future versions derived for free.

lschuermann:
Huh, I just realize I wasn't the only one thinking this might be a good idea. @andre-richter seems to have done essentially the same thing in https://github.com/rust-embedded/register-rs/, right? (It's very similar, but doesn't split it up into three traits for _Read_, _Write_ and _ReadWrite_, which has the drawback that one cannot treat a _ReadWrite_ type as a type being _Read_ or _Write_).

I'm happy to see this, but it also seems slightly obscure to have a library defining some register types, and yet another library using this library and defining traits _mimicking_ this interface, which are however _technically incompatible_ from a trait / interface perspective (i.e. a `ReadWrite<A, B>` doesn't implement `RegisterReadWrite<A, B>` from the registers-rs crate).

I think it'd make much more sense to integrate this into `tock-registers` directly, which would likely obsolete registers-rs (but could live on for backwards compatibility). It would make this a proper general-purpose register interface library, with some templates for register abstractions and the possibility for someone to define their very own types. @andre-richter, do you have an opinion on this?

In an ideal world this library would then live in it's own repository. I don't however know how people in Tock feel about this.

andre-richter:
I‘m happy to have this included here. In fact, my very first pitch read almost like what you’ve written, and I’ve asked back then if we can have this in tock registers.

I wanted a common interface for MMIO and architecture-specific CPU registers to abstract over. So code dealing with either kind of register looks and reads the same. Can’t find it now because I am on mobile.

The answer I got back then was that tock is focused on Microcontrollers and barely reads any CPU registers, so they’d rather not include it directly. This was the birth of register-rs.

If that stance has changed now, even better :)

lschuermann:
Whether this has a stance now isn't up to me, this RFC should be a good place to figure that out. The fact that we essentially re-implement a tock-registers type for CPU registers in Tock right now (RISC-V CSRs), for me at least, is a good indications that one might want to consider this.

We won't be able to do anything about the imports, but we can fix the `LocalRegisterCopy` not implementing the `Copy` trait. A solution could be to have `LocalRegisterCopy` as the only special type not implementing the traits proposed here. That sounds like a very reasonable thing to do, since `LocalRegisterCopy` is by definition only a modifiable copy (snapshot) of a register, and not an actual register type. For situations in which we want to have a register in memory, but implementing the traits, we already have `InMemoryRegister` (which can then be passed into methods taking a `&dyn {Readable, Writeable, ReadWriteable}`.

bradjc (approved):

lschuermann:
I have rebased on latest master. Furthermore, I have changed `LocalRegisterCopy` back to _not_ use a Cell and hence not implement the `Readable` and `Writeable` interface (as it now requires `&mut self` references again). As stated above I think this is fine, because it's by definition not a "real" register type. For a proper register type which is simply stored at a regular memory location we have `InMemoryRegister` anyways, which implements the proper interface.

I hope this makes this somewhat less controversial. I've taken the liberty to transfer this out of the RFC phase. The only remaining question I have is whether `Readable`, `Writeable` and `ReadWriteable` are good names for traits.

alistair23 (dismissed):

bradjc:
After some time to simmer on this, this trait-based implementation with less code re-use is a big improvement. Particularly because RISC-V CSRs use the same interface, and other HW features may benefit from using this software interface. And also because it just seems like the Rusty way to do this.

My hesitancy comes from two things:
1. The registers crate now provides `Readable` and `ReadOnly`, as well as `ReadWrite` and `ReadWriteable`. This is on top of `kernel::Read` and `kernel::ReadWrite` which exist in Tock for appslices. If we didn't have to import them then this wouldn't matter, but Rust's namespacing (or failure to enforce namespacing) is letting us down here.
2. For better or worse, I just find

    ```
    use kernel::common::math;
    use kernel::common::registers::{
        register_bitfields, FieldValue, ReadOnly, ReadWrite, Readable, Writeable,
    };
    ```

    with imports wrapped to their own indented line aesthetically unappealing. It is probably the rustfmt decision I most disagree with.

My proposal is we can resolve both of these by namespacing the traits in a `traits` module. So, capsules would have:

```
use kernel::common::registers::{register_bitfields, FieldValue, ReadOnly, ReadWrite};
use kernel::common::registers::traits::{Readable, Writeable};
```

Is this worth it? I don't know. I think it provides more clarity on what the long list of imports actually are, and to some extent helps explain why `Readable, Writeable` are imported but then never used.

I also want to change the kernel crate to not directly export something called `Read`.

lschuermann:
> The registers crate now provides `Readable` and `ReadOnly`, as well as `ReadWrite` and `ReadWriteable`. This is on top of `kernel::Read` and `kernel::ReadWrite` which exist in Tock for appslices.

I agree, this is terrible.

> My proposal is we can resolve both of these by namespacing the traits in a traits module. 

That seems fine. It manages to distinguish the types from traits, which is great. The only thing bugging me about this is that the types are now technically on a higher level in the module hierarchy than the traits, whereas the traits are usually seen as governing the behavior of types implementing them. But that might just be me.

> I don't know. I think it provides more clarity on what the long list of imports actually are, and to some extent helps explain why Readable, Writeable are imported but then never used.

Yes, I like that.

> I also want to change the kernel crate to not directly export something called Read.

Sounds like a good idea. I'm going to implement your proposed changes, but does anyone of you (native speakers) have better names than `Readable` and the likes, which are at the same time not too verbose? I had a hard time finding something good and in the end just wanted to get this proposal out, not taking hours to come up with good names. :smile:

gendx:
I made a similar proposal in https://github.com/tock/tock/pull/1766, which at the time didn't get much traction. So I'm definitely supportive :)

lschuermann:
Okay, I changed the traits to be exported under `kernel::common::registers::interfaces`. I didn't use `traits`, given I'm a bit wary of potential issues or confusing arising around using a language primitive name as a module name.

This does not yet modify the internal structure of `tock-registers`. I suppose this should make it pretty easy to split up the interfaces, actual registers and helper types (such as `Field`, `FieldValue` and `LocalRegisterCopy`) into different modules. Furthermore, one could think about adding a feature disabling the registers shipped here, such that it's usable as a common-sense library for having interoperable interfaces around registers, without also pulling in the registers themselves.

I propose I do this in a follow-up PR, as it won't change anything about the way `tock-registers` is used in Tock itself. As long as were happy with the terminology used, I consider this PR ready.

ppannuto (approved):
I think this is a good change. We should probably flag it for short, final overview/discussion/notification on this Friday's call, but I suspect it's one that we can come to consensus and merge during the call.

----

> In an ideal world this library would then live in it's own repository. I don't however know how people in Tock feel about this.

We (I) have been somewhat hesitant about this for two reasons:

1. Tock loses 'control' of the interfaces. As a library embedded in Tock, we held the right to change the interface however Tock might need, without worrying about how it would impact other users (though, in practice, we did start flagging @andre-richter whenever we were considering changes once this library became effectively exported). This allowed us to explore and modify things. This was more important when these interfaces with in their nascence. With sufficient time and use, we seem largely happy with these interfaces (indeed, we are now codifying the interfaces with this PR). Still, it means that any change that Tock may wish to make to these in the future is no longer just an internal decision of the Tock project, but a change to a general, public library.

2. It becomes an "external" dependency. Now, it's a dependency owned and managed by the Tock developers, but nonetheless it would be the first non-Rust-core crate depended on, and I believe the first code not directly in the `tock` repository that's built in to the kernel. It's probably a good opportunity to figure out how the Tock team wants to manage a "trusted" dependency (audit / approval process of version-pinned reference to the library?), but this should be part of a longer discussion.

I don't think either of these should affect this PR, but I was thinking about them as next steps, so wanted to write them down while thoughts were fresh.

andre-richter:
Actually, it’s kind of a Schrödinger’s crate in reality, isn’t it? It is in crates.io already.

Internal and external dependency at the same time 🤓

bradjc (approved):

Issue #1082, opened 2018-07-06
Tracking: Allow Apps to Fault and Restart Successfully


bradjc:
Having apps separate from the kernel and able to fail without crashing the system is a core design goal for Tock. Currently, however, faulty apps do crash the entire board. This tracking issue outlines what needs to be done to allow apps to fail and have the kernel restart them successfully.

- [x] Implement the logic for `FaultResponse::Restart` in the kernel. (#941)
  - [x] Update to match changes (#1529)
  - [x] Separate process array index from identifier (#1566)
  - [x] Implement a process identifier allocation scheme so that restarted processes get different identifiers. (#1588)
- [x] Add an app fault option that just stops the app on a fault. (#1238)
- [x] Create a test capsule that can artificially restart otherwise working apps. (#957)
  - [x] Implement capabilities so not just any capsule can restart apps. (#975)
- [x] Verify that all capsules respond correctly to an app being restarted (and having its grant region reset).
  - [x] Blink (timer + LED)
  - [x] hello_loop (timer + console)
  - [x] adc: **fails** (#1525, #1609)
  - [x] ble serialization: **fails** (#1612)
  - [ ] If an IPC service restarts, the client(s) do not know and do not reconfigure themselves.
- [x] Implement tracking and a policy for apps that fail too quickly or too often. (#1565)
- [x] Add a `fault` command to process console to specifically fault certain processes.

bradjc:
I've had a branch https://github.com/tock/tock/tree/adc-grant-apps with a re-write of the ADC for a while now, but I got too ambitious and tried to refactor it as well, but all of the buffers and possibilities make it difficult without breaking backwards compatibility.

bradjc:
If someone is willing to add grant region support to the ADC capsule driver that would really help with this effort.

bradjc:
I think this one is getting pretty close, just needs some PRs to be merged, and some code in branches to become PRs.

Also, I'm planning on scoping this issue to enabling apps to restart without the kernel being in some odd state when they do. This does not mean that all apps will restart and behave exactly as they did when the board first booted. For example, the ADC capsule only supports one app at a time. So, if an app restarts and tries to use the ADC capsule, the restarted app will look the same as if a second app is trying to use the capsule, and the capsule will return an error. The kernel is in an expected state, and if the app waits and tries again to use the capsule it should be fine. The root cause is that the capsule is not virtualized properly, which is a separate issue.

bradjc:
[...]

PR #2518, opened 2021-04-04
Lsm303dlhc: Enforce use by a single process, move callback to Grant

hudson-ayers:
### Pull Request Overview

This pull request is the first in a series of PRs to enforce that all non-virtualized capsules cannot be used by multiple processes.

It adds a method to be able to check whether a process identifier is still valid (whether the same process _instance_ is still running, e.g. it has not restarted).

It adds helper infrastructure for reserving nonvirtualized drivers, in a new folder, `capsules/support`.

Finally, it enforce that only a single process may (implicitly) reserve the nonvirtualized userspace driver for the lsm303dlhc.
It also moves the `Upcall` type into the process `Grant` region, as a preparation for the upcoming PR implementing restrictions on callback swapping #2462 .

This PR was co-authored by @lschuermann .

### Testing Strategy

This pull request was tested by compiling, as we do not have hardware available for this driver. That said, the changes are pretty straightforward.

### TODO or Help Wanted

This pull request still needs testing.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/lsm303dlhc.rs
+++ b/boards/components/src/lsm303dlhc.rs
@@ -54,11 +54,13 @@ macro_rules! lsm303dlhc_i2c_component_helper {
     };};
 }
 
-pub struct Lsm303dlhcI2CComponent {}
+pub struct Lsm303dlhcI2CComponent {
+    board_kernel: &'static kernel::Kernel,
+}
 
 impl Lsm303dlhcI2CComponent {
-    pub fn new() -> Lsm303dlhcI2CComponent {
-        Lsm303dlhcI2CComponent {}
+    pub fn new(board_kernel: &'static kernel::Kernel) -> Lsm303dlhcI2CComponent {
+        Lsm303dlhcI2CComponent { board_kernel }
     }
 }
 
@@ -72,10 +74,17 @@ impl Component for Lsm303dlhcI2CComponent {
     type Output = &'static Lsm303dlhcI2C<'static>;
 
     unsafe fn finalize(self, static_buffer: Self::StaticInput) -> Self::Output {
+        let grant_cap =
+            kernel::create_capability!(kernel::capabilities::MemoryAllocationCapability);
         let lsm303dlhc = static_init_half!(
             static_buffer.3,
             Lsm303dlhcI2C<'static>,
-            Lsm303dlhcI2C::new(static_buffer.0, static_buffer.1, static_buffer.2)
+            Lsm303dlhcI2C::new(
+                static_buffer.0,
+                static_buffer.1,
+                static_buffer.2,
+                self.board_kernel.create_grant(&grant_cap),
+            )
         );
         static_buffer.0.set_client(lsm303dlhc);
         static_buffer.1.set_client(lsm303dlhc);
--- a/boards/stm32f3discovery/src/main.rs
+++ b/boards/stm32f3discovery/src/main.rs
@@ -623,7 +623,7 @@ pub unsafe fn main() {
         components::i2c::I2CMuxComponent::new(&peripherals.i2c1, None, dynamic_deferred_caller)
             .finalize(components::i2c_mux_component_helper!());
 
-    let lsm303dlhc = components::lsm303dlhc::Lsm303dlhcI2CComponent::new()
+    let lsm303dlhc = components::lsm303dlhc::Lsm303dlhcI2CComponent::new(board_kernel)
         .finalize(components::lsm303dlhc_i2c_component_helper!(mux_i2c));
 
     lsm303dlhc.configure(
--- a/capsules/src/lsm303dlhc.rs
+++ b/capsules/src/lsm303dlhc.rs
@@ -79,12 +79,13 @@
 #![allow(non_camel_case_types)]
 
 use core::cell::Cell;
+use core::mem;
 use enum_primitive::cast::FromPrimitive;
 use enum_primitive::enum_from_primitive;
 use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::hil::i2c::{self, Error};
 use kernel::hil::sensors;
-use kernel::{CommandReturn, Driver, ErrorCode, ProcessId, Upcall};
+use kernel::{CommandReturn, Driver, ErrorCode, Grant, ProcessId, Upcall};
 
 use crate::lsm303xx::{
     AccelerometerRegisters, Lsm303AccelDataRate, Lsm303MagnetoDataRate, Lsm303Range, Lsm303Scale,
@@ -134,7 +135,6 @@ pub struct Lsm303dlhcI2C<'a> {
     config_in_progress: Cell<bool>,
     i2c_accelerometer: &'a dyn i2c::I2CDevice,
     i2c_magnetometer: &'a dyn i2c::I2CDevice,
-    callback: Cell<Upcall>,
     state: Cell<State>,
     accel_scale: Cell<Lsm303Scale>,
     mag_range: Cell<Lsm303Range>,
@@ -146,6 +146,20 @@ pub struct Lsm303dlhcI2C<'a> {
     buffer: TakeCell<'static, [u8]>,
     nine_dof_client: OptionalCell<&'a dyn sensors::NineDofClient>,
     temperature_client: OptionalCell<&'a dyn sensors::TemperatureClient>,
+    current_process: OptionalCell<ProcessId>,
+    apps: Grant<App>,
+}
+
+pub struct App {
+    callback: Upcall,
+}
+
+impl Default for App {
+    fn default() -> Self {
+        Self {
+            callback: Upcall::default(),
+        }
+    }
 }
 
 impl<'a> Lsm303dlhcI2C<'a> {
@@ -153,13 +167,13 @@ impl<'a> Lsm303dlhcI2C<'a> {
         i2c_accelerometer: &'a dyn i2c::I2CDevice,
         i2c_magnetometer: &'a dyn i2c::I2CDevice,
         buffer: &'static mut [u8],
+        grant: Grant<App>,
     ) -> Lsm303dlhcI2C<'a> {
         // setup and return struct
         Lsm303dlhcI2C {
             config_in_progress: Cell::new(false),
             i2c_accelerometer: i2c_accelerometer,
             i2c_magnetometer: i2c_magnetometer,
-            callback: Cell::new(Upcall::default()),
             state: Cell::new(State::Idle),
             accel_scale: Cell::new(Lsm303Scale::Scale2G),
             mag_range: Cell::new(Lsm303Range::Range1G),
@@ -171,6 +185,8 @@ impl<'a> Lsm303dlhcI2C<'a> {
             buffer: TakeCell::new(buffer),
             nine_dof_client: OptionalCell::empty(),
             temperature_client: OptionalCell::empty(),
+            current_process: OptionalCell::empty(),
+            apps: grant,
         }
     }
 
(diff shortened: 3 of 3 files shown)


lschuermann:
To explain the reasoning behind these changes:
- To not change the semantics of the nonvirtualized userspace driver's system call ABI too much, this implicitly reserves the driver as soon as the first process accesses it. Every allow, subscribe and command system call (except `command_num == 0`) counts as an access.

  Once a process has reserved the driver, no other process can access it afterwards. To support process restarts, if a the process previously holding a reservation is no longer valid (e.g. has exited / restarted), the driver can be reserved again.

  Only the `Upcall`s/`AppSlice`s are moved into the `Grant` region, even though arguably much more state should be process-local. This is to avoid large changes to the driver logic, as otherwise on a new reservation (i.e. switching to a different, default-initialized `Grant`), the `Grant`-held state might no longer match the actual hardware configuration / state. This does does have the possibility for data leakage between processes, but isn't any worse compared to how the drivers worked previously.

- If another process tries to access a reserved driver, the kernel panics. The reasoning behind this is that, if a system has two different applications trying to access the same nonvirtualized userspace driver, something is obviously wrong.

- Adding an `.is_valid() -> bool` method on `AppId` (soon `ProcessId`) seems moderately uncontroversial, given that one can pretty reliably query this information by trying to access a process' `AppSlice` already.

These PRs don't intend to fix everything about the nonvirtualized userspace drivers currently in Tock. Our goals were to make them compatible to #2462, as well as not make the situation worse than it already is :smile:.

bradjc:
Why not just implement ninedof and temperature for this chip and use those Driver implementations?

Non-virtualized drivers are easy to write, but work poorly with Tock's multi-process model. I know we have them, and probably will in the future, but I'm not sure about adding support code to make it easier to write non virtualized drivers. I'm also opposed to setting the precedent of capsules panic-ing if multiple apps use them. The other app should just get an error and not crash the system.

Also, I'm not sure about `AppId.is_valid()`. That suggests that users of AppId should call it, and places the burden of deciding when on capsule authors. Instead, I thought we had centralized that check any time an AppId is used, and created the necessary errors to handle if a user tries to use an AppId that happens to be invalid.

hudson-ayers:
> Why not just implement ninedof and temperature for this chip and use those Driver implementations?

That might be a fine approach, but seems like something that would have come up when this driver was first merged. This capsule does provide a number of configuration commands that are not exposed through the generic ninedof and temperature HILs, so that may be one reason. Our main goal here is to have a mechanism for somewhat easily making non-virtualized drivers safe / correct in the event that multiple apps try to use them, so that we can proceed with #2462. 

> Non-virtualized drivers are easy to write, but work poorly with Tock's multi-process model. I know we have them, and probably will in the future, but I'm not sure about adding support code to make it easier to write non virtualized drivers. 

The support code is more so intended to make it easier for us to switch the existing set of non-virtualized capsules over. It is, however, an optional part of this PR. My goal, once we have done this transition for all of the capsules, was to move all of the non-virtualized capsules into a `non_virtualized` subfolder with a README containing a disclaimer that this approach is not recommended.

> I'm also opposed to setting the precedent of capsules panic-ing if multiple apps use them. The other app should just get an error and not crash the system.

We could return an error instead. We originally started to take that approach. One reason that doing so is unfortunate is the Tock 2.0 requirement that return types for a given subdriver number always be the same -- if command(5) sometimes returns a `failure_u32()` we need to always return a `failure_u32()`, rather than sometimes return a `failure()` in the event that the capsule is already reserved. We would have to manually identify the places where this happens, or be careful to check that `failure_u32()` return values are not added when these capsules are modified in the future,

> Also, I'm not sure about `AppId.is_valid()`. That suggests that users of AppId should call it, and places the burden of deciding when on capsule authors. Instead, I thought we had centralized that check any time an AppId is used, and created the necessary errors to handle if a user tries to use an AppId that happens to be invalid.

The reason for adding `AppId.is_valid()` is that it seems like the easiest way for us to support apps restarting. Currently the kernel provides no mechanisms for tracking apps across restarts, so the easiest technique was for apps to implicitly release their lock on a driver once the app is no longer valid. Then the restarted app can use the driver again. One solution here could be to move `NonVirtualizedProcessGate` into the `kernel` crate, and make `AppId::is_valid()` `pub(crate)`. This way only the helper type would be able to use that function.

bradjc:
> > Why not just implement ninedof and temperature for this chip and use those Driver implementations?
> 
> That might be a fine approach, but seems like something that would have come up when this driver was first merged. This capsule does provide a number of configuration commands that are not exposed through the generic ninedof and temperature HILs, so that may be one reason. Our main goal here is to have a mechanism for somewhat easily making non-virtualized drivers safe / correct in the event that multiple apps try to use them, so that we can proceed with #2462.

Ok, it could probably do both still, but, you're right that is a bit of a separate issue.

> > Non-virtualized drivers are easy to write, but work poorly with Tock's multi-process model. I know we have them, and probably will in the future, but I'm not sure about adding support code to make it easier to write non virtualized drivers.
> 
> The support code is more so intended to make it easier for us to switch the existing set of non-virtualized capsules over. It is, however, an optional part of this PR. My goal, once we have done this transition for all of the capsules, was to move all of the non-virtualized capsules into a `non_virtualized` subfolder with a README containing a disclaimer that this approach is not recommended.

We have several such non-virtualized drivers that did not require a support module to implement. I'm not sure why it's needed now, particularly if we want to discourage this.

> > I'm also opposed to setting the precedent of capsules panic-ing if multiple apps use them. The other app should just get an error and not crash the system.
> 
> We could return an error instead. We originally started to take that approach. One reason that doing so is unfortunate is the Tock 2.0 requirement that return types for a given subdriver number always be the same -- if command(5) sometimes returns a `failure_u32()` we need to always return a `failure_u32()`, rather than sometimes return a `failure()` in the event that the capsule is already reserved. We would have to manually identify the places where this happens, or be careful to check that `failure_u32()` return values are not added when these capsules are modified in the future.

I see that is a ramification of 2.0, but we can just return the same error type? It only comes up in a couple places.

> 
> > Also, I'm not sure about `AppId.is_valid()`. That suggests that users of AppId should call it, and places the burden of deciding when on capsule authors. Instead, I thought we had centralized that check any time an AppId is used, and created the necessary errors to handle if a user tries to use an AppId that happens to be invalid.
> 
> The reason for adding `AppId.is_valid()` is that it seems like the easiest way for us to support apps restarting. Currently the kernel provides no mechanisms for tracking apps across restarts, so the easiest technique was for apps to implicitly release their lock on a driver once the app is no longer valid. Then the restarted app can use the driver again. One solution here could be to move `NonVirtualizedProcessGate` into the `kernel` crate, and make `AppId::is_valid()` `pub(crate)`. This way only the helper type would be able to use that function.

We do this, for instance in adc, without is_valid. Capsules simply have to try to use an AppId and then handle the error when that AppId is no longer valid. This came up when working on https://github.com/tock/tock/issues/1082.

But, it's still not clear when a capsule should call .is_valid(). Everytime before using an AppId? Only sometimes? Only the first time? What happens if they forget? I think it is easier to just assume the common case (the process has not disappeared) and handle the error case if it arises. This was reasonably straightforward to allow restartable processes.

hudson-ayers:
> > > Non-virtualized drivers are easy to write, but work poorly with Tock's multi-process model. I know we have them, and probably will in the future, but I'm not sure about adding support code to make it easier to write non virtualized drivers.
> > 
> > 
> > The support code is more so intended to make it easier for us to switch the existing set of non-virtualized capsules over. It is, however, an optional part of this PR. My goal, once we have done this transition for all of the capsules, was to move all of the non-virtualized capsules into a `non_virtualized` subfolder with a README containing a disclaimer that this approach is not recommended.
> 
> We have several such non-virtualized drivers that did not require a support module to implement. I'm not sure why it's needed now, particularly if we want to discourage this.

Its not needed, we just thought it was helpful. It could be removed.

> 
> > > I'm also opposed to setting the precedent of capsules panic-ing if multiple apps use them. The other app should just get an error and not crash the system.
> > 
> > 
> > We could return an error instead. We originally started to take that approach. One reason that doing so is unfortunate is the Tock 2.0 requirement that return types for a given subdriver number always be the same -- if command(5) sometimes returns a `failure_u32()` we need to always return a `failure_u32()`, rather than sometimes return a `failure()` in the event that the capsule is already reserved. We would have to manually identify the places where this happens, or be careful to check that `failure_u32()` return values are not added when these capsules are modified in the future.
> 
> I see that is a ramification of 2.0, but we can just return the same error type? It only comes up in a couple places.

Yeah, I think that's fine.

> > > Also, I'm not sure about `AppId.is_valid()`. That suggests that users of AppId should call it, and places the burden of deciding when on capsule authors. Instead, I thought we had centralized that check any time an AppId is used, and created the necessary errors to handle if a user tries to use an AppId that happens to be invalid.
> > 
> > 
> > The reason for adding `AppId.is_valid()` is that it seems like the easiest way for us to support apps restarting. Currently the kernel provides no mechanisms for tracking apps across restarts, so the easiest technique was for apps to implicitly release their lock on a driver once the app is no longer valid. Then the restarted app can use the driver again. One solution here could be to move `NonVirtualizedProcessGate` into the `kernel` crate, and make `AppId::is_valid()` `pub(crate)`. This way only the helper type would be able to use that function.
> 
> We do this, for instance in adc, without is_valid. Capsules simply have to try to use an AppId and then handle the error when that AppId is no longer valid. This came up when working on #1082.
> 
> But, it's still not clear when a capsule should call .is_valid(). Everytime before using an AppId? Only sometimes? Only the first time? What happens if they forget? I think it is easier to just assume the common case (the process has not disappeared) and handle the error case if it arises. This was reasonably straightforward to allow restartable processes.

If `NonVirtualizedProcessGate` is moved into the kernel, capsules will never call `is_valid()`. The only function exposed to them will be `try_reserve()`, which we imagine would be called at the beginning of every system call.

I agree that instead we could take the logic from ADC and use it in every non-virtualized capsule. I think your suggestion amounts to placing this code block at the beginning of every system call for every remaining non-virtualized capsule:

```rust
        // Return true if this app already owns the ADC capsule, if no app owns
        // the ADC capsule, or if the app that is marked as owning the ADC
        // capsule no longer exists.
        let match_or_empty_or_nonexistant = self.appid.map_or(true, |owning_app| {
            // We have recorded that an app has ownership of the ADC.

            // If the ADC is still active, then we need to wait for the operation
            // to finish and the app, whether it exists or not (it may have crashed),
            // still owns this capsule. If the ADC is not active, then
            // we need to verify that that application still exists, and remove
            // it as owner if not.
            if self.active.get() {
                owning_app == &appid
            } else {
                // Check the app still exists.
                //
                // If the `.enter()` succeeds, then the app is still valid, and
                // we can check if the owning app matches the one that called
                // the command. If the `.enter()` fails, then the owning app no
                // longer exists and we return `true` to signify the
                // "or_nonexistant" case.
                self.apps
                    .enter(*owning_app, |_, _| owning_app == &appid)
                    .unwrap_or(true)
            }
        });
        if match_or_empty_or_nonexistant {
            self.appid.set(appid);
        } else {
            return CommandReturn::failure(ErrorCode::NOMEM);
        }
```

Contrasting that with what we use in our PR, it is (IMO) much easier to understand what is happening:
```rust
        if self.owning_process.try_reserve(process_id) == ProcessReservationResult::Fail
        {
            return CommandReturn::failure(ErrorCode::NOMEM);
        }
```

But I think it is fine to use the ADC approach everywhere if there is significant concern that making this process easier to understand will encourage more people to do it.

lschuermann:
I'm going to change this to check for process reservations the way that #2521 does. Will push as soon as #2531 is in (and I'm awake again :smile:).

lschuermann:
I changed this to match the mechanism introduced by @bradjc in #2521. In an effort to reduce confusion I inverted the boolean logic used in determining whether this driver is already used by another process. I'm pretty sure this didn't really help, so I'm happy to change it back for consistency's sake. If we duplicate the logic in every capsule I fear it might deviate over time anyways.

hudson-ayers:
Thanks Leon! Those changes look good to me

bradjc (dismissed):

lschuermann:
I've changed this to return `NOMEM` in case the driver is reserved, and reverted the flipped boolean logic back to essentially a literal copy of what's in #2521.

bradjc (dismissed):

bradjc (approved):

Issue #2490, opened 2021-03-17
Consolidate `mtime` and `mtimecmp` implementation


alistair23:
Every RISC-V chip has it's own memory map of `mtime` and `mtimecmp`. This means that the logic to implement the timer is repeated in all chips.

Instead we should create a generic instance in the RISC-V arch that can be used to access these two register and use that.

PR #2519, opened 2021-04-07
RISC-V: Update the timer implementations to use the arch mtimer

alistair23:
### Pull Request Overview

This fixes https://github.com/tock/tock/issues/2490 by using the generic RISC-V architecture version of mtimer for all of the RISC-V chips.

### Testing Strategy

CI

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/chips/earlgrey/src/timer.rs
+++ b/chips/earlgrey/src/timer.rs
@@ -57,14 +57,21 @@ pub struct RvTimer<'a> {
     registers: StaticRef<TimerRegisters>,
     alarm_client: OptionalCell<&'a dyn time::AlarmClient>,
     overflow_client: OptionalCell<&'a dyn time::OverflowClient>,
+    mtimer: MachineTimer<'a>,
 }
 
 impl<'a> RvTimer<'a> {
-    pub const fn new() -> RvTimer<'a> {
-        RvTimer {
+    pub fn new() -> Self {
+        Self {
             registers: TIMER_BASE,
             alarm_client: OptionalCell::empty(),
             overflow_client: OptionalCell::empty(),
+            mtimer: MachineTimer::new(
+                &TIMER_BASE.compare_low,
+                &TIMER_BASE.compare_high,
+                &TIMER_BASE.value_low,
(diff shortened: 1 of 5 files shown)


chips/earlgrey/src/timer.rs
```diff
@@ -94,19 +94,14 @@ impl time::Time for RvTimer<'_> {
...
-            // Wraparound
-            high = self.registers.value_high.get();
-        }
-        Ticks64::from(((high as u64) << 32) | second_low as u64)
+        let mtimer = MachineTimer::new(
+            &self.registers.compare_low,
+            &self.registers.compare_high,
+            &self.registers.value_low,
+            &self.registers.value_high,
+        );
```
hudson-ayers, line 102: Would it make more sense to just call `MachineTimer::new()` once and save it as a field on `rv_timer`, which is then accessed from the other functions?

hudson-ayers (commented):
Generally looks good, but seems like we could cut out some boilerplate

chips/earlgrey/src/timer.rs
alistair23: Yes it would, but I couldn't figure out how to do that while keeping lifetimes happy.

hudson-ayers:
Sorry, I didn't mean to push that directly to your branch, so I pushed it back. https://github.com/hudson-ayers/tock/commit/6844c4ec09c8c35d945341d6151a214508f0ba8a is a commit that constructs the mtimer only once, if you are happy with it. It makes the constructors no longer `const`, but we don't need them to be `const` anymore anyway

alistair23:
Ah, I didn't think about using the static address to do it!

Thanks for that. Do you mind fixing up the [SiFive CLINT](https://github.com/tock/tock/blob/master/chips/sifive/src/clint.rs) as well?

Feel free to push the changes to this branch/PR.

hudson-ayers (approved):

hudson-ayers (approved):

bradjc (approved):

PR #2521, opened 2021-04-08
capsules: Convert analog comparator to use grant

bradjc:
### Pull Request Overview

As part of #2462 this switches the analog comparator capsule to use a grant.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/analog_comparator.rs
+++ b/capsules/src/analog_comparator.rs
@@ -37,31 +37,38 @@
 use crate::driver;
 pub const DRIVER_NUM: usize = driver::NUM::AnalogComparator as usize;
 
-use core::cell::Cell;
+use core::mem;
+
+use kernel::common::cells::OptionalCell;
 use kernel::hil;
-use kernel::{AppId, CommandReturn, Driver, ErrorCode, Upcall};
+use kernel::{AppId, CommandReturn, Driver, ErrorCode, Grant, Upcall};
 
 pub struct AnalogComparator<'a, A: hil::analog_comparator::AnalogComparator<'a> + 'a> {
     // Analog Comparator driver
     analog_comparator: &'a A,
     channels: &'a [&'a <A as hil::analog_comparator::AnalogComparator<'a>>::Channel],
 
-    // App state
-    callback: Cell<Upcall>,
+    grants: Grant<App>,
+    current_process: OptionalCell<AppId>,
+}
+
+#[derive(Default)]
+pub struct App {
+    callback: Upcall,
 }
 
 impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> AnalogComparator<'a, A> {
     pub fn new(
         analog_comparator: &'a A,
         channels: &'a [&'a <A as hil::analog_comparator::AnalogComparator<'a>>::Channel],
+        grant: Grant<App>,
     ) -> AnalogComparator<'a, A> {
         AnalogComparator {
             // Analog Comparator driver
             analog_comparator,
             channels,
-
-            // App state
-            callback: Cell::new(Upcall::default()),
+            grants: grant,
+            current_process: OptionalCell::empty(),
         }
     }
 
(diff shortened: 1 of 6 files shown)


capsules/src/analog_comparator.rs
```diff
@@ -117,7 +124,19 @@ impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> Driver for AnalogCompa
...
+                .enter(*current_process, |_, _| current_process == &appid)
+                .unwrap_or(true)
+        });
+        if match_or_empty_or_nonexistant {
+            self.current_process.set(appid);
+        } else {
+            return CommandReturn::failure(ErrorCode::NOMEM);
+        }
+
         match command_num {
```
hudson-ayers, line 140: One concern with this approach: it makes it possible for the vale returned by the "driver exists" check (aka command 0) to change over time, if another process claims the driver but later exits.

I guess this is just part of dealing with the existence of non-virtualized drivers. Anyone running multiple apps trying to use one of these capsules is simply doing something wrong.

hudson-ayers (approved):

bradjc:
Good point. I think that is worth an extra check.

hudson-ayers (approved):

hudson-ayers (dismissed):

alistair23 (approved):

capsules/src/analog_comparator.rs
```diff
@@ -117,7 +124,24 @@ impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> Driver for AnalogCompa
...
+        // Check if this driver is free, or already dedicated to this process.
+        let match_or_empty_or_nonexistant = self.current_process.map_or(true, |current_process| {
+            self.grants
+                .enter(*current_process, |_, _| current_process == &appid)
+                .unwrap_or(true)
+        });
+        if match_or_empty_or_nonexistant {
+            self.current_process.set(appid);
+        } else {
+            return CommandReturn::failure(ErrorCode::NOMEM);
```
lschuermann, line 142: Intuitively I would've expected another ErrorCode, something like `BUSY`? `NOMEM` kind of fits here I guess?

lschuermann:
I've just run a minimal example of how this PR checks whether a nonvirtualized driver can be reserved by a process, compared to the implementation in #2518. This was only for curiosity's sake, and to see whether entering the Grant would have significant impact. These numbers are obtained on a `rv32imc` build with `rustc 1.52.0-nightly (1705a7d64 2021-03-18)` and [this minimal example](https://gist.github.com/lschuermann/6b3e7f0cf8de9a6ce09d44e175b55ac6). I measured the system call end-to-end instruction count, that is # of instructions executed from the trap handler, until the `mret` returning to the process.
- Using this approach, on the first command syscall (cmd_num = 1) I measure `903`, on the second `1387`, and on the third and subsequent calls `1034` instructions.
- Using the approach of #2518, the first command syscall takes `890`, and the second `887`.

This is probably fine for non-virtualized drivers (given they aren't Tock's idiomatic way to write drivers). It does however seem to tell us that entering a Grant has a cost which one might want to look into optimizing, along with system call routing and switching to a process. Here is a [more thorough investigation](https://groups.google.com/g/tock-dev/c/FPTmNe4BAq0) of the instructions spent on syscall routing.

capsules/src/analog_comparator.rs
bradjc: Historically that is the error that has been returned because there was no memory available to store the new process's request when grants were not being used.

Maybe we should use `ErrorCode::RESERVE`? I'm not that worried either way as I don't think userspace will actually use these and they should be virtualized anyway.

lschuermann: The explanation behind `NOMEM` makes sense, I'm happy to use that. I went through the other error codes and I agree, no other code seems to fit well.

I'll update #2518 for consistency's sake.

lschuermann (dismissed):

hudson-ayers (approved):

lschuermann (approved):

PR #2522, opened 2021-04-09
Update stack-analysis script: reset_handler is now `main()`

bradjc:
### Pull Request Overview

This should have happened in the re-write of memory initialization.

### Testing Strategy

Should be fine.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers (approved):
This might have more false positives (other function names could include "main") but this is just a helper tool so I think that is fine

lschuermann (approved):

PR #2523, opened 2021-04-09
TRD for HIL design

phil-levis:
### Pull Request Overview

This pull request adds a TRD that provides 9 rules for designing HILs in Tock.

### Testing Strategy

This was passed around the core group and the -dev mailing list for review and comments.

### TODO or Help Wanted

This pull request still needs more feedback and discussion.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


doc/reference/trd-hil-design.md
```diff
@@ -0,0 +1,655 @@
...
+In this approach, when the `ReturnCode` is `SUCCESS`, the `Option` is always supposed
+to be `None`; it the `ReturnCode` has an error value, the `Option` contains the passed
+buffer. This invariant, however, cannot be checked. Transitioning to using `Result`
+both makes Tock more in line with standard Rust code and enforces the invariant.
+
+
+Rule 4: Always Pass a Mutable Reference to Buffers
+===============================
+
+Suppose you are desiging a trait to write some text to an LCD screen. The trait
```
lschuermann, line 338:
```suggestion
Suppose you are designing a trait to write some text to an LCD screen. The trait
```

doc/reference/trd-hil-design.md
```diff
@@ -0,0 +1,655 @@
...
+  fn received_buffer(&self, buf: &'static mut [u8]) {
+    self.lcd.display_text(buf); // discards mut
+  }
+}
+```
+
+The problem is in this last line. `TypeToText` needs a mutable
+reference so it can read into it. But once it passes the reference
+to `LcdTextDisplay`, it discards mutability and cannot get it back:
+`text_displayed` provdes an immutable reference, which then cannot
```
lschuermann, line 404:
```suggestion
`text_displayed` provides an immutable reference, which then cannot
```

doc/reference/trd-hil-design.md
```diff
@@ -0,0 +1,655 @@
...
+There are cases when operations are synchronous *sometimes*. The
+random number generator in Rule 1 is an example. If random bits are
+cached, then a call to request random bits can somtimes retun those
+bits synchronously.  If the random number generator needs to engage
+the underlying AES engine, then the random bits have to be
+asyncronous. As Rule 1 goes into, even operations that *could* be
+synchronous should have a callback that executes asynchronously.
+
+Having a conditional synchronous operation and an asynchronous backup
+is a poor solution. While it might seem to make the synchronous cases
```
lschuermann, line 613: It might be worth mentioning that this _might_ be worth it if the operation is synchronous most of the time, and for efficiency reasons it is undesirable to always have to issue a deferred call? I feel like this might be justified in some cases.

lschuermann (commented):
This is pretty awesome, a great guide for people to get to know the concepts of HILs in Tock (which deviate quite a bit from the way one would do things in other Rust codebases). I also think this is a valuable reference to refer to when discussing HIL designs.

I attached some comments, most of which are simple spell corrections. I'd suggest we use the GitHub commit-apply feature and do a final Markdown-format and squash before merge? Also tried not to nitpick too much. :)

doc/reference/trd-hil-design.md
```diff
@@ -0,0 +1,655 @@
...
+**Draft-Modified:** April 9, 2021<br/>
+**Draft-Version:** 3<br/>
+**Draft-Discuss:** tock-dev@googlegroups.com</br>
+
+Abstract
+-------------------------------
+
+This document describes design rules hardware interface layers (HILs)
+in the Tock operating system. HILs are Rust traits that provide a
+standard interface to a hardware resource, such as a sensor, a flash
```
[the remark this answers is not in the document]
phil-levis: A virtualizer is still on top of that hardware resource. But I think we could say "software abstraction" and that would work too.

doc/reference/trd-hil-design.md
```diff
@@ -0,0 +1,655 @@
...
+Consider the following API for requesting 32 bits of randomness:
+
+```rust
+trait Random {
+  fn random(&self) -> Result<(), ErrorCode>;
+  fn set_client(&self, client: &'static Client);
+}
+
+trait Client {
+  fn random_ready(&self, bits: u32, result: Result<(), ErrorCode>);
```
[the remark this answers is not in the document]
phil-levis: Agreed! I just quickly wrote this up, your API is better.

doc/reference/trd-hil-design.md
phil-levis: It definitely is. But as the introduction says, 

"While these are design rules, they are not sarosanct. There are reasons or 
edge cases why a particular HIL might need to break
one (or more) of them. In such cases, be sure to
understand the reasoning behind the rule; if those considerations
don't apply in your use case, then it might be OK to break it. But it's 
important to realize the exception is true for *all*
implementations of the HIL, not just yours; a HIL is intended to be a
general, reusable API, not a specific implementation."

doc/reference/trd-hil-design.md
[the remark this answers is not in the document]
phil-levis: Although, there is the H in **H**IL...

doc/reference/trd-hil-design.md
```diff
@@ -0,0 +1,655 @@
...
+and ownership.
+
+This document describes these requirements and describes a set of
+design rules for HILs. They are:
+
+1. Do not issue synchronous callbacks.
+2. Split-phase operations return a synchronous `Result` type which 
+   includes an error code in its `Err` value.
+3. Split-phase operations with a buffer parameter return a tuple in their error 
+   result, which includes the passed buffer as an element.
```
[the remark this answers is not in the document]
phil-levis: I agree my wording wasn't great, but found yours  a little hard to parse. How about

"Error results of split-phase operations with a buffer parameter include a reference to passed buffer. This returns the buffer to the caller."

alistair23 (dismissed):

doc/reference/trd-hil-design.md
```diff
@@ -0,0 +1,685 @@
...
+
+1. Do not issue synchronous callbacks.
+2. Split-phase operations return a synchronous `Result` type which 
+   includes an error code in its `Err` value.
+3. For split-phase operations, `Ok` means a callback will occur
+   while `Err` means one won't.
+4. Error results of split-phase operations with a buffer parameter 
+   include a reference to passed buffer. This returns the buffer 
+   to the caller.
+5. Split-phase operrations with a buffer parameter take a mutable reference
```
jrvanwhy, line 67: "operations" is typo'd here.

lschuermann: Haha, I already suggested that above but GitHub has conveniently hidden it. :laughing:

doc/reference/trd-hil-design.md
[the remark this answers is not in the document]
lschuermann: Yeah, it does make sense. HILs can be first and foremost designed for hardware abstraction. If that particular "hardware" is a piece of software behaving like a hardware component, so be it. Thanks for the explanation.

doc/reference/trd-hil-design.md
lschuermann: Sure. I suppose I was still mentally applying the statement of the previous section ("This rule, however, is more precise and strict.") here. In general this rule does make sense.

doc/reference/trd-hil-design.md
```diff
@@ -0,0 +1,685 @@
...
+  
+fn random_ready(&self, bits: u32, result: Result<(), ErrorCode>) {
+  match result {
+     Ok(()) => {
+	   // Use the random bits
+	   self.state.set(State::Idle);
+     },
+	 Err(e) => {
+	   self.state.set(State::Error);
+	 }
```
jrvanwhy, line 168: Something is off with this code block's formatting. It seems to mix indentations widths of 2, 5, and 8.

doc/reference/trd-hil-design.md
```diff
@@ -0,0 +1,685 @@
...
+a HIL defines its own error codes. The I2C HIL, for example, defines an 
+`i2c::Error` enumeration for cases such as address and data negative
+acknowledgements, which can occur in I2C.
+
+If a method doesn't return a synchronous error, there is no way for a caller
+to know if the operation succeeded. This is especially problematic for
+split-phase calls: whether the operation succeeds indicates whether
+there will be a callback.
+
+Rule 3: Split-phase `Result` Values Indicate Whether a Callback Will Occur
```
jrvanwhy, line 282: This section seems incomplete. I assume it was supposed to indicate that `Ok` indicates a callback will occur and `Err` indicates a callback will not.

doc/reference/trd-hil-design.md
lschuermann: Yes, I think we should format & squash prior to merge, as there's trailing whitespaces in the doc and the 80 character limit is commonly violated when using the GitHub suggestions.

doc/reference/trd-hil-design.md
phil-levis: Working on it -- current text is a placeholder, it's not very clear. Will flesh it out tomorrow.

jrvanwhy: Sorry, I didn't mean to push. I forgot this doc is a work in progress.

doc/reference/trd-hil-design.md
```diff
@@ -0,0 +1,713 @@
...
+
+Rule 5: Always Pass a Mutable Reference to Buffers
+===============================
+
+Suppose you are designing a trait to write some text to an LCD screen. The trait
+takes a buffer of ASCII characters, which it puts on the LCD:
+
+```rust
+trait LcdTextDisplay {
+  fn display_text(&self, text: &'static [u8]) -> Result<(), ErrorCode>;
```
ppannuto, line 401:
```suggestion
trait LcdTextDisplay {
  // This is an anti-pattern: the `text` buffer should be `mut`, for reasons explained below
  fn display_text(&self, text: &'static [u8]) -> Result<(), ErrorCode>;
```

I think it's useful when including "wrong" code examples to tie a comment directly to the wrong code, to help ensure skimming readers or copy-pasters have locality indicating that this is a negative example.

doc/reference/trd-hil-design.md
```diff
@@ -0,0 +1,713 @@
...
+
+Capsules implemented on a synchronous HIL only work for
+implementations with synchronous behavior. Such a HIL limits
+reuse. For example, a storage system built on top of this synchronous
+API can only work on the same flash bank instructions are stored on:
+otherwise, the operations will be split-phase.
+
+There are use cases when splitting HILs in this way is worth it. For
+example, straightline code can often be shorter and simpler than
+event-drive systems.  By providing a synchronous API for the subset of
```
ppannuto, line 694:
```suggestion
event-driven systems.  By providing a synchronous API for the subset of
```

ppannuto (approved):
This is a _fantastic_ document

doc/reference/trd-hil-design.md
phil-levis: No worries -- I had forgotten I hadn't finished it. :)

phil-levis:
OK, I think this is in a good enough state to merge as a draft. In a few months, based on experiences and reflection, we can finalize it.

doc/reference/trd-hil-design.md
phil-levis: I think I have fixed the code blocks, I will check them.

lschuermann (approved):
I finally figured out how to use flyspell-mode in Emacs and corrected a few remaining spelling mistakes :smile:. I've run every code block through rustfmt and took the liberty to remove some trailing whitespace.

I really like this document and I think it will be a valuable reference! Thanks @phil-levis!

doc/reference/trd-hil-design.md
lschuermann: The blocks were mostly fixed, I've run them all through rustfmt to check and applied the very few remaining fixes.

doc/reference/trd-hil-design.md
lschuermann: Has been fixed in 730b2dbcda0c00a977fe187fd23d921cbd141550.

hudson-ayers (approved):

doc/reference/trd-hil-design.md
```diff
@@ -0,0 +1,728 @@
...
+fn send(&self, buf: &'static mut [u8]) -> (ReturnCode, Option<&'static mut [u8]>);
+```
+
+In this approach, when the `ReturnCode` is `SUCCESS`, the `Option` is always supposed
+to be `None`; it the `ReturnCode` has an error value, the `Option` contains the passed
+buffer. This invariant, however, cannot be checked. Transitioning to using `Result`
+both makes Tock more in line with standard Rust code and enforces the invariant.
+
+
+Rule 5: Always Pass a Mutable Reference to Buffers
```
jrvanwhy, line 402: Musing: I wonder if we should have some sort of `UniqueROBuffer` that provided immutable access to a reference, but which is guaranteed to be the only borrow of a given buffer. Something like:

```rust
pub struct UniqueROBuffer { buffer: &'static mut [u8] }

impl From<&'static mut [u8]> for UniqueROBuffer {
    fn from(buffer: &'static mut [u8]) -> UniqueROBuffer { UniqueROBuffer { buffer } }
}

impl From<UniqueROBuffer> for &'static mut [u8] {
    fn from(buffer: UniqueROBuffer) -> &'static mut [u8] { buffer.buffer }
}

impl core::ops::Deref for UniqueROBuffer {
    type Target = [u8];
    fn deref(&self) -> &[u8] { &self.buffer }
}
```

jrvanwhy (approved):

doc/reference/trd-hil-design.md
jrvanwhy: The only benefit of `UniqueROBuffer` over `&'static mut [u8]` is it provides immutable rather than mutable access, so it might not be worth the cognitive overhead the extra type and conversions.

hudson-ayers: This sounds somewhat similar to `LeasableBuffer` (https://github.com/tock/tock/blob/master/kernel/src/common/leasable_buffer.rs), which is used for passing a portion of a buffer to a lower-layer call while still getting the entire buffer back in the callback.

There was not really enough momentum to motivate converting existing code that could have used this, so it remains mostly used by the networking stack,

phil-levis: > Musing: I wonder if we should have some sort of `UniqueROBuffer` that provides immutable access to a reference, but which is guaranteed to be the only borrow of a given buffer. Something like:

As Hudson pointed out, parts of Tock have tried doing something like this where it was particularly important/useful. I think types such as these aren't incompatible with this rule. You're no longer passing a buffer, but rather a type that provides more refined access semantics.

PR #2527, opened 2021-04-09
kernel: Add StopWithDebug Process Fault Response

bradjc:
### Pull Request Overview

The `FaultResponse::StopWithDebug` is the same as `Stop`, but the kernel also prints a debug message letting the user know the process faulted and was stopped.

This PR also makes that the default on Nano 33 BLE because if a process fails immediately, the USB/CDC stack doesn't have time to setup, and the panic print message doesn't work. This avoids that issue by having the kernel just stop the process and notify the user via a debug message that it faulted. Then the user can use process console to panic the kernel and see the debug message if they wish.

### Testing Strategy

This pull request was tested by using my `crash-immediately-syscall` app and the pconsole panic command.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nano33ble/src/main.rs
+++ b/boards/nano33ble/src/main.rs
@@ -69,13 +69,20 @@ const PAN_ID: u16 = 0xABCD;
 /// UART Writer for panic!()s.
 pub mod io;
 
-// State for loading and holding applications.
-// How should the kernel respond when a process faults.
-const FAULT_RESPONSE: kernel::procs::FaultResponse = kernel::procs::FaultResponse::Panic;
+// How should the kernel respond when a process faults. Stop but do not
+// immediately panic.
+//
+// We choose this fault response for the nano33ble because it does not have
+// dedicated UART-to-USB hardware. If a process faults very early after boot,
+// the USB stack does not have enough time for initialization, and a `panic!()`
+// leads to the USB-CDC stack to put the serial port in a bad state. By only
+// stopping the app, the user can request a panic later to avoid the issue.
+const FAULT_RESPONSE: kernel::procs::FaultResponse = kernel::procs::FaultResponse::StopWithDebug;
 
 // Number of concurrent processes this platform supports.
 const NUM_PROCS: usize = 8;
 
+// State for loading and holding applications.
 static mut PROCESSES: [Option<&'static dyn kernel::procs::ProcessType>; NUM_PROCS] =
     [None; NUM_PROCS];
 
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -789,6 +789,10 @@ pub enum FaultResponse {
 
     /// Stop the app by no longer scheduling it to run.
     Stop,
+
+    /// Stop the app by no longer scheduling it to run, leave the app in a fault
+    /// state, and print a debug message indicating what happened.
+    StopWithDebug,
 }
 
 /// Tasks that can be enqueued for a process.
@@ -1103,6 +1107,13 @@ impl<C: Chip> ProcessType for Process<'_, C> {
                 // this app as well.
                 self.terminate(COMPLETION_FAULT);
                 self.state.update(State::Faulted);
+            }
+            FaultResponse::StopWithDebug => {
+                // This is the same as `Stop`, but also prints a debug message
+                // noting that the process faulted.
+                self.terminate(COMPLETION_FAULT);
+                self.state.update(State::Faulted);
+                debug!("Process {} faulted and was stopped.", self.process_name);
             }
         }
     }


phil-levis:
I'm a bit wary of this change. We're changing a core kernel abstraction (the fault cases) due to an edge case on one board. I'd much much rather design and define the process fault model.

alistair23 (approved):

ppannuto:
> the USB/CDC stack doesn't have time to setup

Is this a literal time question, i.e. adding a longer delay to `panic_begin` would also address this, or a case where more things in the USB subsystem need to run?

bradjc:
> > the USB/CDC stack doesn't have time to setup
> 
> Is this a literal time question, i.e. adding a longer delay to `panic_begin` would also address this, or a case where more things in the USB subsystem need to run?

I'm not sure. I suspect that simply waiting longer isn't sufficient, as the stack gets in to a strange state that even the host has a hard time dealing with.

hudson-ayers:
> I'm a bit wary of this change. We're changing a core kernel abstraction (the fault cases) due to an edge case on one board. I'd much much rather design and define the process fault model.

I can definitely see this being useful aside from the edge case of the nano33. It doesn't seem hard to imagine use cases where one would want a debug message when an app faults, but wouldn't want the kernel to panic. In fact, I would probably be in favor of making this (or perhaps a similar restart-with-debug FaultResponse) the default on all boards. I think the current approach of panicking the kernel is probably more an artifact of the fact that panicking is currently the only way to get debug info about failed apps.

bradjc:
> That said, it would be nice if this printed out all of the app information that a panic currently provides.

Agreed, but that is more or less impossible right now because of how large the panic debug info is, and how large of a buffer `kernel::debug` would need to hold it all (because the debug functions are synchronous).

bradjc:
An alternative design might look like:

```rust
process.rs:

enum ProcessFaultAction {
  Panic, Restart, Stop
}

trait ProcessFaultPolicy {
  fn action(&self, process: &ProcessType) -> ProcessFaultAction;
}

struct Process {
  ...
  fault_policy: &dyn ProcessFaultPolicy,
  ...
}




impl Process {
  fn set_fault_state () {
    let action = self.fault_policy(self);

    match action {
      ProcessFaultAction::Panic => panic!("process faulted"),
      ProcessFaultAction::Restart => self.restart(),
      ProcessFaultAction::Stop => self.terminate(),
    }
  }
}
```

Perhaps there are more actions than panic, stop, restart, but those seem like the obvious ones. process.rs needs to know what action to take so it can handle the action correctly. However, it doesn't need to know how the policy decided on the action, or if it decided to use a `debug!()` message, or if it wants to restart sometimes and stop other times.

I think this should be somewhat straightforward change. Should we go this route? I suspect @phil-levis is in favor.

lschuermann:
I agree with @hudson-ayers in that this is probably very useful, and panic'ing because of a process fault is a good default behavior for development, but stopping the process and printing a message might be more sensible. It also better conveys the fact to people new to Tock that processes are indeed properly isolated, as panicing the system from a process fault might given the initial impression that a process fault is necessarily coupled to a system panic.

That said -- if I understand it correctly -- I like @bradjc's proposal to allow custom implementations of a `ProcessFaultAction` a lot! In particular, the fact that a `dyn ProcessFaultAction::fault_policy` is passed both the process and a `&self` reference allows it to perform arbitrary complex logic and have associated state, for instance to log details about a process fault to persistent storage. While such a hook brings potential for "abuse", it's gives the developers freedom to flexibly define their process fault behavior without intrusive changes in the kernel, which I really like.

phil-levis:
@bradjc and I chatted about this in a recent call. My major concern related to `FaultResponse` being an enum, such that any additional algorithms or policies require modifying `process.rs`. The proposed approach of making it a trait solves this. Two suggestions:

1. Name it `fault` rather than `set_fault_state` or `action`; this isn't setting a state, and it might not actually cause the system to enter a state (e.g., panic). 
2. The enumeration of possible kernel responses is excellent: it crisply defines how the kernel can respond (the mechanism). Should there be a `Resume` action, which just tells the kernel to resume the process? This does start going down the same path as POSIX signals, where some faults are recoverable/maskable, others aren't. It could be that we don't have any systems or policies which would need this now. But is it something we want to preclude or prohibit? My initial thought is that we don't want to prohibit it, but I am not sure and could be convinced otherwise.

bradjc:
> @bradjc and I chatted about this in a recent call. My major concern related to `FaultResponse` being an enum, such that any additional algorithms or policies require modifying `process.rs`. The proposed approach of making it a trait solves this. Two suggestions:
> 
>     1. Name it `fault` rather than `set_fault_state` or `action`; this isn't setting a state, and it might not actually cause the system to enter a state (e.g., panic).

This is a bit of a larger change, as many `Process` fns are named `set_X_state()`. This can be a separate change if needed.

>     2. The enumeration of possible kernel responses is excellent: it crisply defines how the kernel can respond (the mechanism). Should there be a `Resume` action, which just tells the kernel to resume the process? This does start going down the same path as POSIX signals, where some faults are recoverable/maskable, others aren't. It could be that we don't have any systems or policies which would need this now. But is it something we want to preclude or prohibit? My initial thought is that we don't want to prohibit it, but I am not sure and could be convinced otherwise.

How would this work with `platform.process_fault_hook()`? I think if a board wants to try continuing a process, it can return `Ok()` from that function.

bradjc:
#2537 is the new implementation.

phil-levis:
Close this?

bradjc:
#2537 merged, no longer need this open for reference to the discussion.

PR #2528, opened 2021-04-09
Add a "panic" command to process console

bradjc:
### Pull Request Overview

This pull request is supposed to be just adding a "panic" command to process console so that users can force a panic, and more importantly the debug information that comes along with it. This is a stop-gap if we can instead use the process console to display the per-process debug info without a full kernel panic someday.

Unfortunately, on the board I want to use this on (the nano33ble), just doing a normal `panic!()` in process console doesn't work. The issue is that process console commands are parsed in a UART TX done callback (which on the nano33 originates from the CDC/USB stack). Since panic changes the control flow immediately, the TX done callback never finishes returning down the callback chain, and the USB driver's state is effectively corrupted.

So, this implementation uses a deferred call so that the panic runs on a different callback. This makes it work on the nano33ble, but adds a deferred call for every process console user.

### Testing Strategy

Using it on the nano33ble.

### TODO or Help Wanted

I don't like this solution (hence why no other boards compile since I only fixed up the nano33ble). Any thoughts on a better way to handle this? Some possible options:

1. Use this approach, and implement for all boards that use process console.
2. Make the deferred call an `Option<>`, and just immediately panic for boards that use a `None`.
3. Try to hack the panic implementation in nano33ble/io.rs to cleanup the state that wasn't cleaned up if the panic happens in a UART callback.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/process_console.rs
+++ b/capsules/src/process_console.rs
@@ -12,6 +12,7 @@
 //!  - 'stop n' stops the process with name n
 //!  - 'start n' starts the stopped process with name n
 //!  - 'fault n' forces the process with name n into a fault state
+//!  - 'panic' causes the kernel to run the panic handler
 //!
 //! ### `list` Command Fields:
 //!
@@ -210,7 +211,7 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
                         let clean_str = s.trim();
                         if clean_str.starts_with("help") {
                             debug!("Welcome to the process console.");
-                            debug!("Valid commands are: help status list stop start fault");
+                            debug!("Valid commands are: help status list stop start fault panic");
                         } else if clean_str.starts_with("start") {
                             let argument = clean_str.split_whitespace().nth(1);
                             argument.map(|name| {
@@ -290,6 +291,8 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
                                 "Timeslice expirations: {}",
                                 info.timeslice_expirations(&self.capability)
                             );
+                        } else if clean_str.starts_with("panic") {
+                            panic!("ProcessConsole forced a kernel panic.");
                         } else {
                             debug!("Valid commands are: help status list stop start fault");
                         }


phil-levis:
> TX done callback never finishes returning down the callback chain, and the USB driver's state is effectively corrupted.

This sounds like a bug in the USB driver? The stack should be in a well-defined state in the callback. I'll try to take a look.

alistair23:
I'm not a fan of a userspace process being able to panic the kernel

hudson-ayers:
> I'm not a fan of a userspace process being able to panic the kernel

The process console is not a userspace process, it is a privileged (and optional) kernel capsule that directly exposes a console interface over a serial connection. It is present even if no userspace processes exist on the board. It also provides the ability for a user to start/stop/fault arbitrary processes, which seems about as dangerous as being able to panic the kernel

alistair23:
Ah, you are right

In which case option 3 might be the best option. I feel like Tock should be able to handle a panic call anywhere, even inside the USB stack. If that isn't feasible though then option 2 seems like the next best bet.

hudson-ayers:
I tested this out some, and even with the changes in this PR I do not always see panic prints complete -- they often stop after 3 or 4 buffers. Somewhat separately, I think that the `enable_interrupts()` and `disable_interrupts()` lines in `usbd.rs::handle_interrupt()` should be removed (https://github.com/tock/tock/blob/1c2bcbe177b9376d5ac87a1af0b0b36a772aedec/chips/nrf52/src/usbd.rs#L1133). Given how Tock handles interrupts, I do not think there is any benefit to disabling interrupts in an interrupt handler, as the interrupts of concern have already been disabled at the NVIC granularity in https://github.com/tock/tock/blob/1c2bcbe177b9376d5ac87a1af0b0b36a772aedec/arch/cortex-m/src/lib.rs#L150

bradjc:
This works ok on the nano33ble if no processes are using the console, however, it seems to not print all of the panic information reliably if I have, say, the `hello_loop` test installed. So, the complexity of the deferred call is probably not worth it if it doesn't even work reliably on the board that inspired it.

hudson-ayers:
I was testing with hello_loop installed as well if I remember correctly. That sounds like something that might be good to dig into separately -- it seems possible the same issue could manifest for a normal kernel panic, depending on the state of the USB stack.

hudson-ayers:
Should we close this then?

bradjc:
I still would like to add the panic command.

bradjc:
Ok now this PR just adds the "panic" command.

lschuermann (approved):

hudson-ayers (approved):

PR #2531, opened 2021-04-12
tock-cells/OptionalCell: add get-method for retrieving inner Option

lschuermann:
### Pull Request Overview

Adds an OptionalCell::get() method which returns a copy of the
contained Option<T>, which is either None or Some. This allows a user
to treat an OptionalCell<T> as a Cell<Option<T>> in most cases, while
still providing the benefits of a regular OptionalCell<T>.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

- [X] ~Updated the relevant files in `/docs`, or~ no updates are required.

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/libraries/tock-cells/src/optional_cell.rs
+++ b/libraries/tock-cells/src/optional_cell.rs
@@ -148,6 +148,11 @@ impl<T> OptionalCell<T> {
 }
 
 impl<T: Copy> OptionalCell<T> {
+    /// Returns a copy the contained [`Option`].
+    pub fn get(&self) -> Option<T> {
+        self.value.get()
+    }
+
     /// Returns the contained value or panics if contents is `None`.
     pub fn expect(&self, msg: &str) -> T {
         self.value.get().expect(msg)


hudson-ayers (approved):
This should have been included from the start, IMO

jrvanwhy (approved):

bradjc:
What is the use case for this? The reason .get() wasn't implemented is
after calling get you are in the same position, you still have to use map
or something else to handle the None case.

hudson-ayers:
sorry if I merged this prematurely. Let me know if I should back it out.

Generally the case where this is useful is anywhere you are passing the internal value to something that takes in an `Option`. I have run into this a few times before and meant to make a PR like this in the past. Then, instead of

```rust
my_optional_cell.map_or_else(|| {
        f(None)
    }, 
    |contained| {
        f(Some(contained))
});
```
you just have

```rust
f(my_optional_cell.get());
```

Having this method also lets you do 

```rust
if let Some(contained) = my_optional_cell {

}
```
style binding, and lets you `match` on the contained value. the ability to use match lets you use pattern matching directly, wheras with `map_or` you would have to put the pattern matching internally:

```rust
match my_optional_cell.get() {
    Some(Enum::TheVariantIExpect) => true,
    _ => panic!("unexpected!")
}
```
vs.
```rust
my_optional_cell.map_or_else(|| {
    panic!("unexpected")
    }, |contained| {
    match contained {
        Enum::TheVariantIExpect => true,
        _ => panic!("unexpected"),
    }
})
```

Leon mentioned that it could make https://github.com/tock/tock/pull/2521/ (and transitioning other capsules) cleaner as well, though I don't know the exact technique he planned to use.

lschuermann:
@hudson-ayers is spot on. Besides the issues he mentioned, this also increases flexibility for developers to express statements as they see fit. Closures are a great tool in Rust but can also make your code quite a bit more complicated than it ought to be.

Furthermore, they tend to cause you biting your own tail when it comes to capturing the environment: passing references isn't always possible due to lifetime constraints, and a `move` closure will move values out of the outer scope. Compare this to an `if-let` or `match`, where the compiler (borrow checker) can deduce that only a single branch will be entered and hence you're not trying to do anything potentially dangerous (e.g. as you see in Tock 2.0 drivers' allow method implementations, where the `AppSlice` can't be passed into the Grant enter closure, and into an `.unwrap_or_else(` if entering the Grant fails, as such requiring an additional `match` below).

As a workaround I've used `Cell::<Option<T>>::get(mycell)`, or `OptionalCell::<T>::map(mycell, |val| val)`. Given that this can be implemented without any impacts on safety, those workarounds seem unnecessary.

I do however understand your point, and if from an API design standpoint this isn't desired, I'm more than happy to have this reverted and just fall back to the aforementioned workarounds. It's not a big deal.

bradjc:
You've convinced me being able to pull the option out is useful, but is there a different name we can use? I tend to think of `OptionalCell` and `Cell` the same, and will sometimes call `.get()` thinking the optional cell is just a cell. But since .get() doesn't exist, it's clear that I cannot just get the value and have to handle the None case. Without .get() being unimplemented, it would be easy to write:

```
ocell.get().map(|v| { ... });
```

which is clunky.

I don't want to stray too far from Rust convention, so maybe .get() is the only reasonable option, but I think that if get is not `OptionalCell<T>.get() -> T`, then it's not really a "get". I was thinking `.option()` or `.extract()`.

lschuermann:
I'm absolutely open to using a different name, that was also a concern of mine. I like `option()`, although I'm not sure that's expressive enough. What would you say about `into_option()`? (I think using `into_<type>` is a common idiom in the Rust ecosystem) I'll happily create a follow-up PR.

kupiakos:
> but I think that if get is not `OptionalCell<T>.get() -> T`, then it's not really a "get"

I disagree; when I pulled up this type the first question I asked was "where's the 'get' method"? For something that has the name "optional", it's only natural that "getting" the value returns an `Option`.

Other examples of `Type<T>::get()` returning an `Option<T>` in the standard library include:
- `[T]::get`
- `HashMap<T>::get`
- This one is new and exactly matches our case: [`OnceCell<T>::get`](https://doc.rust-lang.org/core/cell/struct.OnceCell.html#method.get), which returns `None` if the cell is empty or being initialized

kupiakos:
> Without .get() being unimplemented, it would be easy to write:
> 
> ```
> ocell.get().map(|v| { ... });
> ```
> 
> which is clunky.

I find this clunky in the same way `some_option.map(|x| do_thing(x)).map_err(|e| log_err!(e))` is clunky - it could be written more succinctly as `some_option.map_or_else(|e| log_err!(e), |x| do_thing(x))`. The writer in the `ocell.get().map()` example is not taking advantage of a [`fn map<R>(&self, f: impl FnOnce(T) -> R) -> Option<R>` convenience method](https://github.com/tock/tock/pull/3535) with the same semantics, when the purpose of `OptionalCell<T>` is to be a convenient shorthand for `Cell<Option<T>>`. This is a readability issue - one that I believe is outweighed by the more obscure name of `extract`.

Also, when `get().map()` is able to compile, a field can be migrated from `Cell<Option<T>>` to `OptionalCell<T>` and still have `.get().map()` work - as well as any other method chained after `.get()`. This could be seen as an advantage or disadvantage, though I lean to the former. It's an advantage because:
1. It allows for incremental migration of code from `Cell<Option<T>>` to `OptionalCell<T>`.
2. This is similar to how `take` and `replace` on `OptionalCell<T>` behave with respect to `Cell<Option<T>>` - no breaking change for `ocell.take().foo()` and `ocell.replace(todo!()).foo()`.
3. Solely-readability concerns with writing `get().foo()` in working code can be fixed during targeted refactors.

It's a disadvantage because:
1. One method of refactoring is to break all callsites that should be affected by the change, and this makes that easy - the `get` method isn't there.
2. Allowing known-clunky code to compile can be seen as a form of tech debt accumulation.
3. Most of the convenience functions on `OptionalCell` are shorthand for `get().foo()`. That's a distinction with `take` and `replace`.

PR #2534, opened 2021-04-13
kernel: introduce debug::panic_print for boards without LEDs

lschuermann:
### Pull Request Overview

Introduce an alternative to `debug::panic` which does not require an LED type parameter and returns instead of blinking an LED in an infinite loop.

This is useful for boards which do not feature any GPIO or LEDs, such as the SweRVolf and LiteX sim boards running inside Verilator, and other virtualized boards (for example, QEMU riscv32-virt as currently developed in #2516) in the future.

Consequently, the SweRVolf and LiteX sim boards have been changed to use this function instead.

### Testing Strategy

This pull request was tested by issuing a panic after loading processes on the LIteX simulation. It requires testing on SweRVolf and some board using the regular `debug::panic` function.

### TODO or Help Wanted

- I'm not sure enough boards can benefit from this justify splitting it into two functions in `debug!`?

  Alternatives include a mock LED type implementation as done in LiteX sim right now, or duplicating the `debug::panic` method body as done in SweRVolf right now.

  I'll happily continue to use one of those approaches if this isn't deemed worth it.

- ~`debug::panic_noloop` might be a really bad name for this.~

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/debug.rs
+++ b/kernel/src/debug.rs
@@ -88,6 +88,32 @@ pub trait IoWrite {
 ///////////////////////////////////////////////////////////////////
 // panic! support routines
 
+/// Tock panic routine, without the infinite LED-blinking loop.
+///
+/// This is useful for boards which do not feature LEDs to blink or
+/// want to implement their own behaviour. This method returns after
+/// performing the panic dump.
+///
+/// After this method returns, the system is no longer in a
+/// well-defined state. Care must be taken on how one interacts with
+/// the system once this function returns.
+///
+/// **NOTE:** The supplied `writer` must be synchronous.
+pub unsafe fn panic_print<W: Write + IoWrite, C: Chip>(
+    writer: &mut W,
+    panic_info: &PanicInfo,
+    nop: &dyn Fn(),
+    processes: &'static [Option<&'static dyn ProcessType>],
+    chip: &'static Option<&'static C>,
+) {
+    panic_begin(nop);
+    panic_banner(writer, panic_info);
+    // Flush debug buffer if needed
+    flush(writer);
+    panic_cpu_state(chip, writer);
+    panic_process_info(processes, writer);
+}
+
 /// Tock default panic routine.
 ///
 /// **NOTE:** The supplied `writer` must be synchronous.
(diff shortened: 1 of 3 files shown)


kernel/src/debug.rs
```diff
@@ -88,23 +88,51 @@ pub trait IoWrite {
...
+/// performing the panic dump.
+///
+/// After this method returns, the system is no longer in a
+/// well-defined state. Care must be taken on how one interacts with
+/// the system once this function returns.
 ///
 /// **NOTE:** The supplied `writer` must be synchronous.
-pub unsafe fn panic<L: hil::led::Led, W: Write + IoWrite, C: Chip>(
-    leds: &mut [&L],
+pub unsafe fn panic_noloop<W: Write + IoWrite, C: Chip>(
```
hudson-ayers, line 102: I think a better name might be `panic_print_all` or something.

hudson-ayers (dismissed):
This looks fine to me. Its nice this lets the swervolf still exit the simulation if it wants to. I do wonder if the DummyLed approach could have been replaced by just passing an empty slice, which would serve half the use case here.

alistair23 (approved):

kernel/src/debug.rs
bradjc, line 102:
```suggestion
pub unsafe fn panic_print<W: Write + IoWrite, C: Chip>(
```

lschuermann: Looks good to me. Applied using Git to update it everywhere.

hudson-ayers (approved):

kernel/src/debug.rs
lschuermann: Applied the suggestion from Brad below, which is essentially the same. Hope that's okay.

bradjc (approved):

alistair23 (approved):

lschuermann:
> I do wonder if the DummyLed approach could have been replaced by just passing an empty slice, which would serve half the use case here.

Sorry, overlooked that question. That''s actually what I've been doing in LiteX sim: https://github.com/tock/tock/blob/4a29f3829a071e16459eb6e77e0a63da66af40a8/boards/litex/sim/src/io.rs#L58-L65

Unfortunately, a proper type implementing `hil::led::Led` is still required for the monomorphization of the `debug::panic` function to work, Rust sadly can't invent one out of thin air. :smile:

PR #2535, opened 2021-04-14
Proximity capsule: don't double enter grant

alevy:
### Pull Request Overview

This pull request fixes a bug in the proximity driver introduced by #2137. The proximty driver re-enters an entered grant while enqueuing a command by iterating through _all_ grants to determine if any have outstanding commands.

Instead of iterating through grants to determine whether there are outstanding commands, this PR simply uses the outstanding command field.

### Testing Strategy

Tested using the `sensors` app in libtock-c. Without this PR, the app results in a kernel panic, with this change it works (sort of, see TODO)

### TODO

The apds9960 driver, which implements the `ProximityDriver` hil for the nano33ble sense does not currently implement the startup sequence properly. As a result, the chip is, by default, (probably) left in either a low power mode or some other unintended state.

Several restarts of the board after programming or power up seem to eventually put the chip into the right state.

### Documentation Updated

- [X] ~Updated the relevant files in `/docs`, or no updates are required.~

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/proximity.rs
+++ b/capsules/src/proximity.rs
@@ -133,7 +133,13 @@ impl<'a> ProximitySensor<'a> {
                 if (self.command_running.get() == ProximityCommand::ReadProximityOnInterrupt)
                     && (command == ProximityCommand::ReadProximityOnInterrupt)
                 {
-                    let t: Thresholds = self.find_thresholds();
+                    let mut t: Thresholds = self.find_thresholds();
+                    if t.lower < app.lower_proximity {
+                        t.lower = app.lower_proximity;
+                    }
+                    if t.upper > app.upper_proximity {
+                        t.upper = app.upper_proximity;
+                    }
                     let _ = self.driver.read_proximity_on_interrupt(t.lower, t.upper);
                     self.command_running
                         .set(ProximityCommand::ReadProximityOnInterrupt);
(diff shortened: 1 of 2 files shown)


hudson-ayers (changes requested):

capsules/src/proximity.rs
```diff
@@ -151,19 +151,20 @@ impl<'a> ProximitySensor<'a> {
...
+                            let _ = self.driver.read_proximity();
+                            self.command_running.set(ProximityCommand::ReadProximity);
                         }
-                    });
-                }
-
-                if num_commands == 1 {
-                    let _ = self.run_next_command();
+                        ProximityCommand::ReadProximityOnInterrupt => {
+                            let t: Thresholds = self.find_thresholds();
```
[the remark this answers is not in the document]
alevy: You're right. `examples/tests/proximity` in libtock-c triggers this.

I actually don't believe any of this has to happen from within the entered grant. We can just determine whether to run a command. But `iter_unentered_grants` works for now.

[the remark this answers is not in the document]
alevy: Fixed in 8b13b0c

capsules/src/proximity.rs
```diff
@@ -151,19 +151,20 @@ impl<'a> ProximitySensor<'a> {
                     return CommandReturn::success();
```
bradjc, line 151: On line 136, what happens if the current app being enqueued has the most extreme thresholds?

hudson-ayers: yeah, in both places actually, is using unentered_grants() then you need to manually check the thresholds of the app already entered

bradjc: It's a very difficult API to use correctly.

alevy: fixed in 13f8fd6

hudson-ayers (dismissed):
I added a commit to fix a second place where the thresholds could be calculated incorrectly. Marking as last call because

a) I looked over this pretty closely this time
b) the sensors/ example not working on the nano33 is a serious bug we should try to fix quickly

hudson-ayers (approved):

PR #2537, opened 2021-04-16
kernel: Change ProcessFaultPolicy to a trait, add StopWithDebug policy

bradjc:
### Pull Request Overview

This pull request changes how per-process fault policies are handled in the kernel. Now, each process has a reference to a policy with trait:

```rust
pub trait ProcessFaultPolicy {
    fn action(&self, process: &dyn Process) -> process::FaultAction;
}
```

and all that policy has to do is return one action out of `Panic`, `Stop`, `Restart`. _How_ it decides on the action is completely up to the policy.

I also added a new policy called `StopWithDebug`, which prints a debug message if a process faults and then stops the process.

Each board is updated to the new policy mechanism.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/acd52832/src/main.rs
+++ b/boards/acd52832/src/main.rs
@@ -41,7 +41,7 @@ pub mod io;
 
 // State for loading and holding applications.
 // How should the kernel respond when a process faults.
-const FAULT_RESPONSE: kernel::procs::FaultResponse = kernel::procs::FaultResponse::Panic;
+const FAULT_RESPONSE: kernel::procs::PanicFaultPolicy = kernel::procs::PanicFaultPolicy {};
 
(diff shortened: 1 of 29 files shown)


hudson-ayers (approved):
Looks good, though I remain a little worried about the continued spread of dynamic dispatch throughout the core kernel. I don't really see a better alternative here that would also address the maintainability concerns that prompted this change

lschuermann (approved):

bradjc:
If your kernel is spending too many instructions deciding what to do when a process crashes then you probably have bigger problems. :)

PR #2538, opened 2021-04-17
Make I2C trait return buffers upon request error

alexandruradovici:
### Pull Request Overview

This pull request modifies the I2C HIL so that request functions return the given buffer upon error. Without this, the i2c caller will loose the buffers if the I2C device is busy or is not able to perform the request. I noticed that the SMBus already did implement this.

This PR also:
 - modifies the error the the SMBus returns so that it follows the StatusCode (extended - as buffers are also returned) way.
 - adds checking for an outstanding request for the I2C Virtual device

A small problem arises when the I2C Virtualization is used. Request functions may fail due to two reasons:
1. The virtual device already has an outstanding request, in which case it immediately returns an error and the buffers
2. The virtual device accept the request, but later on the Mux is not able to perform it, in which case the error should be returned asynchronously via `read_write_done`. As the I2C Error enum did not provide a suitable option for this error, this PR adds a new `Result (ErrorCode)` item.

This PR does not yet modify drivers as I am waiting for feedback first.

### Testing Strategy

### TODO or Help Wanted

While I think this is useful as it corrects some corner cases, I am not sure that this approach is the best. Function definitions become a little for complicated. Any feedback would be greatly appreciated.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/capsules/src/fxos8700cq.rs
+++ b/capsules/src/fxos8700cq.rs
@@ -208,7 +208,7 @@ impl<'a> Fxos8700cq<'a> {
             buf[0] = Registers::CtrlReg4 as u8;
             buf[1] = 1; // CtrlReg4 data ready interrupt
             buf[2] = 1; // CtrlReg5 drdy on pin 1
-                        // TODO verify errors
+            // TODO verify errors
             let _ = self.i2c.write(buf, 3);
             self.state.set(State::ReadAccelSetup);
         });
@@ -258,7 +258,7 @@ impl I2CClient for Fxos8700cq<'_> {
             });
             return;
         }
-        match self.state.get() {
+        match self.state.get() { 
             State::ReadAccelSetup => {
                 // Setup the interrupt so we know when the sample is ready
                 self.interrupt_pin1
@@ -298,7 +298,7 @@ impl I2CClient for Fxos8700cq<'_> {
                 // Now put the chip into standby mode.
                 buffer[0] = Registers::CtrlReg1 as u8;
                 buffer[1] = 0; // Set the active bit to 0.
-                               // TODO verify errors
+                // TODO verify errors
                 let _ = self.i2c.write(buffer, 2);
                 self.state
                     .set(State::ReadAccelDeactivating(x as i16, y as i16, z as i16));
--- a/capsules/src/fxos8700cq.rs
+++ b/capsules/src/fxos8700cq.rs
@@ -208,7 +208,8 @@ impl<'a> Fxos8700cq<'a> {
             buf[0] = Registers::CtrlReg4 as u8;
             buf[1] = 1; // CtrlReg4 data ready interrupt
             buf[2] = 1; // CtrlReg5 drdy on pin 1
-            self.i2c.write(buf, 3);
+                        // TODO verify errors
+            let _ = self.i2c.write(buf, 3);
             self.state.set(State::ReadAccelSetup);
         });
     }
@@ -220,7 +221,8 @@ impl<'a> Fxos8700cq<'a> {
             buf[0] = Registers::MCtrlReg1 as u8;
             // Enable both accelerometer and magnetometer, and set one-shot read.
             buf[1] = 0b00100011;
-            self.i2c.write(buf, 2);
+            // TODO verify errors
+            let _ = self.i2c.write(buf, 2);
             self.state.set(State::ReadMagStart);
         });
     }
@@ -234,7 +236,8 @@ impl gpio::Client for Fxos8700cq<'_> {
             // When we get this interrupt we can read the sample.
             self.i2c.enable();
             buffer[0] = Registers::OutXMsb as u8;
-            self.i2c.write_read(buffer, 1, 6); // read 6 accel registers for xyz
+            // TODO verify errors
+            let _ = self.i2c.write_read(buffer, 1, 6); // read 6 accel registers for xyz
             self.state.set(State::ReadAccelReading);
         });
     }
@@ -264,7 +267,8 @@ impl I2CClient for Fxos8700cq<'_> {
                 // Enable the accelerometer.
                 buffer[0] = Registers::CtrlReg1 as u8;
                 buffer[1] = 1;
-                self.i2c.write(buffer, 2);
+                // TODO verify errors
+                let _ = self.i2c.write(buffer, 2);
                 self.state.set(State::ReadAccelWait);
             }
             State::ReadAccelWait => {
@@ -272,7 +276,8 @@ impl I2CClient for Fxos8700cq<'_> {
                     // Sample is already ready.
                     self.interrupt_pin1.disable_interrupts();
                     buffer[0] = Registers::OutXMsb as u8;
-                    self.i2c.write_read(buffer, 1, 6); // read 6 accel registers for xyz
+                    // TODO verify errors
+                    let _ = self.i2c.write_read(buffer, 1, 6); // read 6 accel registers for xyz
                     self.state.set(State::ReadAccelReading);
                 } else {
                     // Wait for the interrupt to trigger
@@ -293,7 +298,8 @@ impl I2CClient for Fxos8700cq<'_> {
                 // Now put the chip into standby mode.
                 buffer[0] = Registers::CtrlReg1 as u8;
                 buffer[1] = 0; // Set the active bit to 0.
-                self.i2c.write(buffer, 2);
+                               // TODO verify errors
+                let _ = self.i2c.write(buffer, 2);
                 self.state
                     .set(State::ReadAccelDeactivating(x as i16, y as i16, z as i16));
             }
@@ -309,7 +315,8 @@ impl I2CClient for Fxos8700cq<'_> {
                 // One shot measurement taken, now read result.
                 buffer[0] = Registers::MOutXMsb as u8;
                 self.state.set(State::ReadMagValues);
-                self.i2c.write_read(buffer, 1, 6);
+                // TODO verify errors
+                let _ = self.i2c.write_read(buffer, 1, 6);
             }
             State::ReadMagValues => {
                 let x = (((buffer[0] as u16) << 8) | buffer[1] as u16) as i16;
--- a/capsules/src/fxos8700cq.rs
+++ b/capsules/src/fxos8700cq.rs
@@ -25,7 +25,7 @@ use core::cell::Cell;
 use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::hil;
 use kernel::hil::gpio;
-use kernel::hil::i2c::{I2CClient, I2CDevice};
+use kernel::hil::i2c::{Error, I2CClient, I2CDevice};
 use kernel::ErrorCode;
 
 pub static mut BUF: [u8; 6] = [0; 6];
@@ -241,7 +241,7 @@ impl gpio::Client for Fxos8700cq<'_> {
 }
 
 impl I2CClient for Fxos8700cq<'_> {
-    fn command_complete(&self, buffer: &'static mut [u8], status: Result<(), ErrorCode>) {
+    fn command_complete(&self, buffer: &'static mut [u8], status: Result<(), Error>) {
         // If there's an I2C error, just reset and issue a callback
         // with all 0s. Otherwise, if there's no sensor attacherd,
         // it's possible to have nondeterminstic behavior, where
--- a/capsules/src/fxos8700cq.rs
+++ b/capsules/src/fxos8700cq.rs
@@ -208,6 +208,7 @@ impl<'a> Fxos8700cq<'a> {
             buf[0] = Registers::CtrlReg4 as u8;
             buf[1] = 1; // CtrlReg4 data ready interrupt
             buf[2] = 1; // CtrlReg5 drdy on pin 1
+
             // TODO verify errors
             let _ = self.i2c.write(buf, 3);
             self.state.set(State::ReadAccelSetup);
@@ -258,7 +259,7 @@ impl I2CClient for Fxos8700cq<'_> {
             });
             return;
         }
-        match self.state.get() { 
+        match self.state.get() {
             State::ReadAccelSetup => {
                 // Setup the interrupt so we know when the sample is ready
                 self.interrupt_pin1
@@ -298,6 +299,7 @@ impl I2CClient for Fxos8700cq<'_> {
                 // Now put the chip into standby mode.
                 buffer[0] = Registers::CtrlReg1 as u8;
                 buffer[1] = 0; // Set the active bit to 0.
+
                 // TODO verify errors
                 let _ = self.i2c.write(buffer, 2);
                 self.state
--- a/capsules/src/fxos8700cq.rs
+++ b/capsules/src/fxos8700cq.rs
@@ -25,7 +25,7 @@ use core::cell::Cell;
 use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::hil;
 use kernel::hil::gpio;
-use kernel::hil::i2c::{Error, I2CClient, I2CDevice};
+use kernel::hil::i2c::{I2CClient, I2CDevice};
 use kernel::ErrorCode;
 
 pub static mut BUF: [u8; 6] = [0; 6];
@@ -241,13 +241,13 @@ impl gpio::Client for Fxos8700cq<'_> {
 }
 
 impl I2CClient for Fxos8700cq<'_> {
-    fn command_complete(&self, buffer: &'static mut [u8], error: Error) {
+    fn command_complete(&self, buffer: &'static mut [u8], status: Result<(), ErrorCode>) {
         // If there's an I2C error, just reset and issue a callback
         // with all 0s. Otherwise, if there's no sensor attacherd,
         // it's possible to have nondeterminstic behavior, where
         // sometimes you get callbacks and sometimes you don't, based
         // on whether a floating interrupt line triggers. -pal 3/19/21
-        if error != Error::CommandComplete {
+        if status != Ok(()) {
             self.state.set(State::Disabled);
             self.buffer.replace(buffer);
             self.callback.map(|cb| {
--- a/capsules/src/hts221.rs
+++ b/capsules/src/hts221.rs
@@ -117,14 +117,16 @@ impl<'a> Hts221<'a> {
                 match self.state.get() {
                     State::Reset => {
                         buffer[0] = REG_AUTO_INCREMENT | CALIB_REG_1ST;
-                        self.i2c.write_read(buffer, 1, 16);
+                        // TODO verify errors
+                        let _ = self.i2c.write_read(buffer, 1, 16);
                         self.state.set(State::Calibrating);
                     }
                     State::Idle(calibration_data, _, _) => {
                         buffer[0] = REG_AUTO_INCREMENT | CTRL_REG1;
                         buffer[1] = 1 << 2 | 1 << 7; // BDU + PD
                         buffer[2] = 1; // ONE SHOT
-                        self.i2c.write(buffer, 3);
+                                       // TODO verify errors
+                        let _ = self.i2c.write(buffer, 3);
                         self.state.set(State::InitiateReading(calibration_data));
                     }
                     _ => {} // Should really never happen since we only have `buffer` available in the above two states
@@ -202,7 +204,8 @@ impl<'a> I2CClient for Hts221<'a> {
                         buffer[0] = REG_AUTO_INCREMENT | CTRL_REG1;
                         buffer[1] = 1 << 2 | 1 << 7; // BDU + PD
                         buffer[2] = 1; // ONE SHOT
-                        self.i2c.write(buffer, 3);
+                                       // TODO verify errors
+                        let _ = self.i2c.write(buffer, 3);
                         self.state.set(State::InitiateReading(CalibrationData {
                             temp_slope,
                             temp_intercept,
@@ -212,17 +215,20 @@ impl<'a> I2CClient for Hts221<'a> {
                     }
                     State::InitiateReading(calibration_data) => {
                         buffer[0] = STATUS_REG;
-                        self.i2c.write_read(buffer, 1, 1);
+                        // TODO verify errors
+                        let _ = self.i2c.write_read(buffer, 1, 1);
                         self.state.set(State::CheckStatus(calibration_data));
                     }
                     State::CheckStatus(calibration_data) => {
                         if buffer[0] & 0b11 == 0b11 {
                             buffer[0] = REG_AUTO_INCREMENT | HUMID0_REG;
-                            self.i2c.write_read(buffer, 1, 4);
+                            // TODO verify errors
+                            let _ = self.i2c.write_read(buffer, 1, 4);
                             self.state.set(State::Read(calibration_data));
                         } else {
                             buffer[0] = STATUS_REG;
-                            self.i2c.write_read(buffer, 1, 1);
+                            // TODO verify errors
+                            let _ = self.i2c.write_read(buffer, 1, 1);
                         }
                     }
                     State::Read(calibration_data) => {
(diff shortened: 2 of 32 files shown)


kernel/src/hil/i2c.rs
```diff
@@ -27,19 +28,27 @@ pub enum Error {
 
     /// No error occured and the command completed successfully.
     CommandComplete,
+
+    /// Error reported when a virtual device fails synchronously
+    Request(ErrorCode),
```
hudson-ayers, line 33: I am not sure what "request" means here, could this be something like `VirtualizationErr`?

hudson-ayers (commented):
Correctly returning the buffers in the case of failure is a needed change. But I don't really like that now the synchronous HIL methods return `ErrorCode`, while `command_complete()` returns an `I2C::Error`. This means that identical failures would be returned as different error types depending on if a request was executed immediately or after being queued.

I think it would probably be better to return `I2C::Error` everywhere -- especially since that type can now wrap an `ErrorCode` as well -- since it is more expressive.

kernel/src/hil/i2c.rs
alexandruradovici: That would be an error returned if the actual request to the I2C would not be successful. Even though it is only in the case of virtualization, it is not an actual virtualization error. At first, I named it ErrorCode (`ErrorCode(ErrorCode)`). Maybe CommandError ?

alexandruradovici:
> Correctly returning the buffers in the case of failure is a needed change. But I don't really like that now the synchronous HIL methods return `ErrorCode`, while `command_complete()` returns an `I2C::Error`. This means that identical failures would be returned as different error types depending on if a request was executed immediately or after being queued.
> 
> I think it would probably be better to return `I2C::Error` everywhere -- especially since that type can now wrap an `ErrorCode` as well -- since it is more expressive.

The only difference is that `I2C::Error` returns transmission errors, while commands fail for totally different reasons, as they do not even start the transmission when returning an error.

Additionally, a command should not return for any reason `CommandComplete`.

hudson-ayers:
> The only difference is that `I2C::Error` returns transmission errors, while commands fail for totally different reasons, as they do not even start the transmission when returning an error.
> 
> Additionally, a command should not return for any reason `CommandComplete`.

I probably should not have said "everywhere" -- I guess this applied just for `transmit()` then. I did not realize `I2C::Error` does not have errors for reception. But if a transmission executes synchronously, any errors will have to be returned via an `ErrorCode` with the current design (or via the `command_complete()` callback via a deferred call -- is that what happens?). If the same transmission executes asynchronously (queued by the virtualizer), errors will be returned as `I2C::Error` via the callback. 

Am I misunderstanding how this works?

kernel/src/hil/i2c.rs
```diff
@@ -27,19 +28,27 @@ pub enum Error {
 
     /// No error occured and the command completed successfully.
     CommandComplete,
```
hudson-ayers, line 30: I agree with your ML comment that having `CommandComplete` here no longer makes sense -- it might be better to remove this enum variant and instead have the `command_complete()` callback pass a `Result<(), Error>`

alexandruradovici:
> > The only difference is that `I2C::Error` returns transmission errors, while commands fail for totally different reasons, as they do not even start the transmission when returning an error.
> > Additionally, a command should not return for any reason `CommandComplete`.
> 
> I probably should not have said "everywhere" -- I guess this applied just for `transmit()` then. I did not realize `I2C::Error` does not have errors for reception. But if a transmission executes synchronously, any errors will have to be returned via an `ErrorCode` with the current design (or via the `command_complete()` callback via a deferred call -- is that what happens?). If the same transmission executes asynchronously (queued by the virtualizer), errors will be returned as `I2C::Error` via the callback.
> 
> Am I misunderstanding how this works?

My choice of words was poor. I meant that `I2CError` returns only errors related to the actual data transfer. It does not cover cases where a transmission cannot even be started for reasons like busy device, invalid buffer size etc.  

Whether the data transfer is performed in a synchronous way or not should should not be relevant. The HIL commands (functions) return whether the data transfer can or cannot start. If the transfer starts, it returns Ok (()), if not, it returns an Err(ErrorCode).

A special case, not covered in the HIL TRD, is what happens when a command is not issued immediately, as it waits in a queue. Then, the actual error has to be returned via the callback. For that special case only, I added the Request(ErrorCode).

This might be a problem for other HILs, like SpiMaster, as its client does not define an error in the callback.

kernel/src/hil/i2c.rs
alexandruradovici: This make much more sense. On the other hand, as you wrote in the mailing list, if some HILs return different, that might be a problem with converting errors.

hudson-ayers:
> My choice of words was poor. I meant that `I2CError` returns only errors related to the actual data transfer. It does not cover cases where a transmission cannot even be started for reasons like busy device, invalid buffer size etc.
> 
> Whether the data transfer is performed in a synchronous way or not should should not be relevant. The HIL commands (functions) return whether the data transfer can or cannot start. If the transfer starts, it returns Ok (()), if not, it returns an Err(ErrorCode).
> 
> A special case, not covered in the HIL TRD, is what happens when a command is not issued immediately, as it waits in a queue. Then, the actual error has to be returned via the callback. For that special case only, I added the Request(ErrorCode).
> 
> This might be a problem for other HILs, like SpiMaster, as its client does not define an error in the callback.

Ah, I see what you are going for. I was imagining that with the added `Request(ErrorCode)` variant, `I2CError` could be used to cover cases where a transmission cannot even be started (for reasons like busy device, etc.) -- those would be passed as an ErrorCode in that variant. Then you could use `Result<(), I2CError>` everywhere that currently returns `Result<(), ErrorCode>` or `I2CError` (combined with removing `I2CError::CommandComplete`). The added advantage here is that in the case where an I2C transmission is not queued, but executed immediately, then has an `AddressError` or similar, you can get a specific I2CError immediately rather than having to schedule that error to be returned in a callback through a `DeferredCall`.

I am not sure which approach is better, and would love for others to weigh in (cc @phil-levis )

alistair23:
I agree with Hudson, I think that the callback and the caller should both use the same error type.

alexandruradovici:
Based on what the HIL TRD states, the returned error should be the same for synchronous calls and asynchronous calls. 

What about returning something like this:

```rust
pub trait I2CHwMasterClient {
    /// Called when an I2C command completed. The `error` denotes whether the command completed
    /// successfully or if an error occured.
    fn command_complete(
                        &self, buffer: &'static mut [u8], 
                        error: Result<(), (error_code: ErrorCode, specific_error: Option<i2c::Error>)>
    );
}
```

In this way the HIL returns an `ErrorCode` and potentially a more fined grained error depending on the trait.

lschuermann:
> In this way the HIL returns an `ErrorCode` and potentially a more fined grained error depending on the trait.

I don't think that is the intended effect of the HIL changes, and will likely lead to more confusion in the error handling (how are the first and the second error coupled, and how to deal with inconsistencies?). As far as I understand the changes as merged [here](https://github.com/tock/tock/pull/2550/commits/f743c2ce01061e22fac81a9aa85697c6b2d50266), the intended way to handle errors is by either (preferably) using an `ErrorCode` in both synchronous and asynchronous cases, or (if there is a good reason to do so, such as here) just use a custom enum, also in both synchronous and asynchronous cases. Hence, the custom error enum should cover all error cases, so all errors that can arise synchronously and asynchronously.

alexandruradovici:
> > In this way the HIL returns an `ErrorCode` and potentially a more fined grained error depending on the trait.
> 
> I don't think that is the intended effect of the HIL changes, and will likely lead to more confusion in the error handling (how are the first and the second error coupled, and how to deal with inconsistencies?). As far as I understand the changes as merged [here](https://github.com/tock/tock/pull/2550/commits/f743c2ce01061e22fac81a9aa85697c6b2d50266), the intended way to handle errors is by either (preferably) using an `ErrorCode` in both synchronous and asynchronous cases, or (if there is a good reason to do so, such as here) just use a custom enum, also in both synchronous and asynchronous cases. Hence, the custom error enum should cover all error cases, so all errors that can arise synchronously and asynchronously.

I am trying to avoid using a custom error as this will make error handling more difficult. I think many HILs will actually have to use a custom error, as ErrorCode only covers generic errors. 

I can think of two solutions that would avoid this:

##### Modify ErrorCode (far fetched)
I know this might sound crazy and requires a lot of changes, but what about adding a payload to ErrorCode::FAIL. This is very general and I think this is the case when a callback returns an error.

```rust
enum ErrorCode<E> {
   FAIL (E)
   // ...
}
```
##### Use ErrorCode
We could use ErrorCode and maybe add another function like `get_last_specific_error(&self) -> i2c:Error' to return a more fine grained error.

lschuermann:
> ##### Modify ErrorCode (far fetched)
> 
> I know this might sound crazy and requires a lot of changes, but what about adding a payload to ErrorCode::FAIL. This is very general and I think this is the case when a callback returns an error.

I'm not sure about this. We'd be losing some of the nice guarantees provided by the rather simple (essentially C-style) enum which `ErrorCode` currently is. For example, we can currently losslessly convert it to an integer, which we use for providing errors and status-fields (in the case of upcalls) to userspace. Furthermore, having a generic field in `ErrorCode` will cause the different ErrorCodes to be incompatible to each other, and each will be monomorphorized into its own distinct type at compile time, opening the potential for bloating the text section. I'm not absolutely opposed to it though.

> ##### Use ErrorCode
> 
> We could use ErrorCode and maybe add another function like `get_last_specific_error(&self) -> i2c:Error' to return a more fine grained error.

To me this ultimately sounds more confusing than just having a custom type. I think that you either need a custom, more descriptive error type, or can just adopt `ErrorCode` somehow. Returning both a coarse error description in the form of an `ErrorCode` and _then_ also providing more information sounds redundant and confusing.

> I am trying to avoid using a custom error as this will make error handling more difficult. I think many HILs will actually have to use a custom error, as ErrorCode only covers generic errors.

Error handling isn't easy, in particular since in Rust you usually have to handle each error that can come up. "Easy" error handling made possible by using the same error enumeration may look more elegant, but often can be straight up wrong: just passing some returned error up to the caller without filtering may likely violate the contract of your function in terms of the errors you return. In this way I'd argue, that it'd probably be "better" to always have precise error enumerations which can describe exactly what can go wrong. But for the reasons outlined above, in Tock we have an `ErrorCode`. This still doesn't spare us from handling each error case individually and making sure our functions return what they're supposed to -- if we want our code to be correct. As such, I don't think a custom error enumeration will hurt in these cases.

alexandruradovici:
Based in the discussion here, I modified the HIL so that it only returns `ErrorCode`. There are a few things where I am not sure how to continue:

1. I mapped the `ArbitrationLost` to `ErrorCode::FAIL`, I'm not sure if the is the best value
2. apollo3 uses a synchronous callback to report an error, I modified it to return an error, @alistair23 can you take a look if that is ok?
3. I'm not sure that the code for saml4 I2CSlave is correct, @bradjc it would be great if you can take a look at the modifications here. I think that the read and write slave functions might have to return another error at `if interrupts.is_set(StatusSlave::SAM) && !status.is_set(StatusSlave::TRA)`. I think it should return `Ok(())` only from inside the if.
4. lowrisc does not seem to make any checks whether it can actually set up an I2C transfer. As I don't have the actual board, it would be very helpful if @alistair23 you could take a look and confirm this.
5. Now that the I2C returns a `Result`, there are a lot of usage warnings. As quick fix would be either to use `let _ =` or `unwrap()`. I'm not sure which one should be preferred.

chips/apollo3/src/iom.rs
```diff
@@ -563,37 +567,42 @@ impl<'a> Iom<'_> {
...
-            });
-            return;
-        }
-
-        // Clear and enable interrupts
-        regs.intclr.set(0xFFFF_FFFF);
-        regs.inten.set(0xFFFF_FFFF);
+            // self.master_client.map(|client| {
+            //     client.command_complete(self.buffer.take().unwrap(), Err(ErrorCode::NOSUPPORT));
+            // });
```
alistair23, line 574: Can you remove this?

alistair23 (commented):
The apollo3 and lowriscv changes look good to me

chips/apollo3/src/iom.rs
alexandruradovici: Yes, just wanted confirmation from you.

kernel/src/hil/i2c.rs
```diff
@@ -55,9 +14,25 @@ pub trait I2CMaster {
...
-    fn write_read(&self, addr: u8, data: &'static mut [u8], write_len: u8, read_len: u8);
-    fn write(&self, addr: u8, data: &'static mut [u8], len: u8);
-    fn read(&self, addr: u8, buffer: &'static mut [u8], len: u8);
+    fn write_read(
+        &self,
+        addr: u8,
+        data: &'static mut [u8],
+        write_len: u8,
+        read_len: u8,
+    ) -> Result<(), (ErrorCode, &'static mut [u8])>;
```
phil-levis, line 23: I thought the `i2c::Error` type was actually very useful, because it allowed you to easily figure out what had gone wrong, e.g., when using new hardware. I'm mostly in favor of keeping it, but enlarging it to include other general error conditions to support virtualization (e.g., OFF).

If the trait does transition to `ErrorCode`, then I think the calls need to map I2C error conditions to `ErrorCodes` and document that mapping. E.g., `ErrorCode::NOACK` could be mean either address NAK or data NAK, `ErrorCode::RESERVE` could mean arbitration lost, etc.

phil-levis (changes requested):

kernel/src/hil/i2c.rs
phil-levis: (Sorry I am just getting to this now)

It seems like the issue we're wrestling with is neither the original Error nor ErrorCode can capture all of the error cases we care about, and we don't want to lose valuable information due to this. I'd suggest expanding Error to include relevant more general ErrorCodes, and writing a `from` so you can map it to an ErrorCode, with I2C-specific errors mapping as best we can.

lschuermann: Indeed, that sounds like a very reasonable approach. I agree the custom type adds some value here, and the fact that we're going to have to convert it to an application specific `ErrorCode` down the line is just a consequence of Rust's approach to error handling, but one which is necessary in order to get correct and dependable interfaces.

I think this discussion has brought a lot of valuable insights on how we want to do error handling in these potentially more complex scenarios. I hope @alexandruradovici won't mind changing the error handling story again.

alexandruradovici: Don worry, I don't mind, this is why I opened the discussion to get this discussion going. Just to sum up, to see if I got this right:
 - put back `i2c:Error` and make the HIL return this
 - add a From trait implementation to `ErrorCode` to be able to convert `i2c:Error` to `ErrorCode`

Another item is the is not the mapping of `i2c::Error` to `ErrorCode`. Based on the discussion here, this might be the following:

| i2c::Error | ErrorCode |
|----------|------------|
| DataNak | NOACK |
| AddressNak | NOACK |
| ArbitrationLost | REVERVE |
| Overrun | ? |
| Not Supported | NOSUPPORT |

hudson-ayers: I might suggest `SIZE` for Overrun, though it is not a perfect match

phil-levis: I agree it is not a perfect match, but as long as we clearly state the mapping, that is OK.

lschuermann: Right. And IMHO it wouldn't even hurt if we didn't have an `impl From<i2c::Error> for ErrorCode`. Doing the conversion at the client ensures that the errors are mapped exactly according to its desired interface. But having a conversion to `ErrorCode` using the `From` trait is of course also fine, it's good to me either way!

alexandruradovici: Another quick question: what error should be returned if the device has another operation inflight (the former `ErrorCode::BUSY`)? I was thinking of `Error::ArbitrationLost` but I am not sure if this is the best one.

Maybe we should add another item to `i2c::Error` for this case?

alexandruradovici:
I reverted the error back to `i2c::Error`.

As there was not variant for `BUSY` I added a `Busy` variant to `i2c:Error`.

There are a few things that still need to be cleared for `saml4`:

- there is a (probably theoretical) situation when the DMA might not be available. I'm not sure what error to return. For now I used `NotSupported`
- if the slave address is not set, the error returned now is `AddressNak`. I'm not sure that this is the best approach
- if the (slave) device is off, I used `NotSupported`

As soon as we clear out these things, I'll with drivers and error handling.

hudson-ayers:
> I reverted the error back to `i2c::Error`.
> 
> As there was not variant for `BUSY` I added a `Busy` variant to `i2c:Error`.

Sounds good to me

> There are a few things that still need to be cleared for `saml4`:
> 
> * there is a (probably theoretical) situation when the DMA might not be available. I'm not sure what error to return. For now I used `NotSupported`

After looking at the code I didn't immediately see where this error could occur. If it is a situation where "DMA not being available" means the dma reference was never passed to the I2C driver I think `NotSupported` is good. If the situation is instead the DMA dynamically rejecting the request (not sure that is possible on the sam4l anyway), I think `Busy` is probably more appropriate.

> * if the slave address is not set, the error returned now is `AddressNak`. I'm not sure that this is the best approach

Seems fine to me.

> * if the (slave) device is off, I used `NotSupported`

I think `AddressNak` is also appropriate here -- how do you distinguish between the device being off and the wrong address being broadcast on the bus?

alexandruradovici:
> > I reverted the error back to `i2c::Error`.
> > As there was not variant for `BUSY` I added a `Busy` variant to `i2c:Error`.
> 
> Sounds good to me
> 
> > There are a few things that still need to be cleared for `saml4`:
> > 
> > * there is a (probably theoretical) situation when the DMA might not be available. I'm not sure what error to return. For now I used `NotSupported`
> 
> After looking at the code I didn't immediately see where this error could occur. If it is a situation where "DMA not being available" means the dma reference was never passed to the I2C driver I think `NotSupported` is good. If the situation is instead the DMA dynamically rejecting the request (not sure that is possible on the sam4l anyway), I think `Busy` is probably more appropriate.
> 
I think this applies only to the first situation when the dma reference was not passed.
> > * if the slave address is not set, the error returned now is `AddressNak`. I'm not sure that this is the best approach
> 
> Seems fine to me.
> 
> > * if the (slave) device is off, I used `NotSupported`
> 
> I think `AddressNak` is also appropriate here -- how do you distinguish between the device being off and the wrong address being broadcast on the bus?

It just verifies if the slave is enabled (`if self.slave_enabled.get()`). This seems to happen only if `enable(&self)` was not called.

kernel/src/hil/i2c.rs
alexandruradovici: This has been settled.

capsules/src/virtual_i2c.rs
```diff
@@ -244,22 +265,42 @@ impl i2c::I2CDevice for I2CDevice<'_> {
...
+        read_len: u8,
+    ) -> Result<(), (Error, &'static mut [u8])> {
+        if self.operation.get() == Op::Idle {
+            self.buffer.replace(data);
+            self.operation.set(Op::WriteRead(write_len, read_len));
+            self.mux.do_next_op();
+            Ok(())
+        } else {
+            Err((Error::ArbitrationLost, data))
+        }
```
hudson-ayers, line 281: Just flagging this because it is clearly a change in logic -- the idea here is that if a node is not currently idle we will always reject new requests does make sense to me.

hudson-ayers (approved):
Overall looks great -- just a few comments

capsules/src/virtual_i2c.rs
alexandruradovici: I think this was an error before, the mux would accept new requests wile having one in progress. I think this would have lead to buffer leakage and unexpected results.

hudson-ayers: That makes sense to me

capsules/src/fxos8700cq.rs
```diff
@@ -208,7 +208,8 @@ impl<'a> Fxos8700cq<'a> {
             buf[0] = Registers::CtrlReg4 as u8;
             buf[1] = 1; // CtrlReg4 data ready interrupt
             buf[2] = 1; // CtrlReg5 drdy on pin 1
-            self.i2c.write(buf, 3);
+                        // TODO verify errors
```
alistair23, line 211: Wrong alignment here.

capsules/src/fxos8700cq.rs
```diff
@@ -293,7 +298,8 @@ impl I2CClient for Fxos8700cq<'_> {
                 // Now put the chip into standby mode.
                 buffer[0] = Registers::CtrlReg1 as u8;
                 buffer[1] = 0; // Set the active bit to 0.
-                self.i2c.write(buffer, 2);
+                               // TODO verify errors
```
alistair23, line 301: and here

capsules/src/hts221.rs
```diff
@@ -117,14 +117,16 @@ impl<'a> Hts221<'a> {
...
+                        // TODO verify errors
+                        let _ = self.i2c.write_read(buffer, 1, 16);
                         self.state.set(State::Calibrating);
                     }
                     State::Idle(calibration_data, _, _) => {
                         buffer[0] = REG_AUTO_INCREMENT | CTRL_REG1;
                         buffer[1] = 1 << 2 | 1 << 7; // BDU + PD
                         buffer[2] = 1; // ONE SHOT
-                        self.i2c.write(buffer, 3);
+                                       // TODO verify errors
```
alistair23, line 128: and here

alistair23 (changes requested):

capsules/src/i2c_master.rs
```diff
@@ -72,18 +72,24 @@ impl<'a, I: 'a + i2c::I2CMaster> I2CMasterDriver<'a, I> {
                         }
                         self.tx.put(Transaction { app_id, read_len });
 
-                        match command {
-                            Cmd::Ping => (), // Unexpected, shouldn't get here (was Err(ErrorCode::INVAL))
+                        match match command {
```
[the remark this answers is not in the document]
alexandruradovici: Yes :) Should I split this into two matches?

capsules/src/fxos8700cq.rs
alexandruradovici: Sorry about that, updated.

capsules/src/hts221.rs
alexandruradovici: It seems that this is a formatter error. I fixed them and afterwards format fails. I added a newline before some of the comments.

capsules/src/i2c_master.rs
```diff
@@ -52,16 +52,16 @@ impl<'a, I: 'a + i2c::I2CMaster> I2CMasterDriver<'a, I> {
...
         rlen: u8,
-    ) {
+    ) -> Result<(), ErrorCode> {
         // TODO(alevy) this function used to try and return Result<(), ErrorCode>s, but would always return
         // ENOSUPPORT and all call-sites simply ignore the return value. Nonetheless, some error
         // handling is probably useful. Comments inline where there used to be non-success results.
         self.apps
             .enter(app_id, |_| {
                 // TODO(alevy): if app.slice.map doesn't have a slice, we would have returned
                 // INVAL here. I.e., the driver is attempting an operation without sharing memory.
```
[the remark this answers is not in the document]
alexandruradovici: Done

capsules/src/i2c_master.rs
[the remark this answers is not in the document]
alexandruradovici: Done

hudson-ayers (approved):

hudson-ayers:
Marked as last-call. Any last comments? All requested changes have been addressed and this change brings the I2C HIL in-line with the recommended structure in TRD-hil-design, I would like to merge this soon so we can review the PRs blocked on it

alistair23 (approved):

PR #2539, opened 2021-04-17
[RFC] multicore support

alexandruradovici:
### Pull Request Overview

Several MCUs provide multiple processing core within the ARM Cortex. One example is the Raspberry Pi Pico that is a dual core M0+. Another example is Arduino Portenta H7 (STM32 with two cores, an M4 and an M7).

I am not sure how this should be handled by Tock so I am opening this PR to get some feedback.

I think there might be several approaches to this, with pros and cons for each of them.

#### SMP
This seems to be the most intuitive approach. The kernel runs and schedules all the processes on any of the available cores.
##### Pros
 - easy to understand and use for use space developer perspective, no changes are needed
 - allows faster processing no matter what actions the apps perform
##### Cons
 - breaks the idea that drivers run uninterrupted on a single thread, requires drivers rewrite
 - make the kernel scheduler more complex as it needs to take into account processes that might have already been scheduled by the kernel running on another core

#### Dedicated core for a specific process
The idea is to use the additional cores for one single user space process. This process:
  - should only perform processing task
  - should never be preempted
  - may only use system calls to a specific set of drivers that are never used by the main core kernel
 
##### Pros
  - allows real time processing in user space
  - rather simple implementation in the kernel, no scheduling required
##### Cons
  - user space programming is different, due to limited API that is usable
  - might not use all the processing power available, as the core is used only by one single process

#### Separated kernels
The idea is to run a separate kernel on each core. Each kernel:
  - has its own memory space
  - loads a different subset of applications
  - loads a different subset of drivers
 
##### Pros
 - easy to implement, requires minimal kernel changes
 - should work with different core types (Arduino Protenta H7 has an M4 and an M7)
##### Cons
  - might waste memory resources
  - requires some driver changes to be able to share peripherals among the kernels

### Testing Strategy

N/A

### TODO or Help Wanted

As this requires several changes, I am looking for a lot of feedback.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/chips/apollo3/src/chip.rs
+++ b/chips/apollo3/src/chip.rs
@@ -88,6 +88,7 @@ impl<I: InterruptService<()> + 'static> Chip for Apollo3<I> {
     type UserspaceKernelBoundary = cortexm4::syscall::SysCall;
     type SchedulerTimer = cortexm4::systick::SysTick;
     type WatchDog = ();
+    type Core = ();
 
     fn service_pending_interrupts(&self) {
         unsafe {
@@ -143,5 +144,9 @@ impl<I: InterruptService<()> + 'static> Chip for Apollo3<I> {
 
     unsafe fn print_state(&self, write: &mut dyn Write) {
         cortexm4::print_cortexm4_state(write);
+    }
+
+    fn current_core(&self) -> &Self::Core {
+        &()
     }
 }
--- a/chips/arty_e21_chip/src/chip.rs
+++ b/chips/arty_e21_chip/src/chip.rs
@@ -146,6 +146,7 @@ impl<'a, I: InterruptService<()> + 'a> kernel::Chip for ArtyExx<'a, I> {
     type UserspaceKernelBoundary = rv32i::syscall::SysCall;
     type SchedulerTimer = ();
     type WatchDog = ();
+    type Core = ();
 
     fn mpu(&self) -> &Self::MPU {
         &self.pmp
@@ -196,6 +197,10 @@ impl<'a, I: InterruptService<()> + 'a> kernel::Chip for ArtyExx<'a, I> {
 
     unsafe fn print_state(&self, write: &mut dyn Write) {
         rv32i::print_riscv_state(write);
+    }
+
+    fn current_core(&self) -> &Self::Core {
+        &()
     }
 }
 
--- a/chips/e310x/src/chip.rs
+++ b/chips/e310x/src/chip.rs
@@ -109,6 +109,7 @@ impl<'a, A: 'static + Alarm<'static>, I: InterruptService<()> + 'a> kernel::Chip
     type UserspaceKernelBoundary = rv32i::syscall::SysCall;
     type SchedulerTimer = kernel::VirtualSchedulerTimer<A>;
     type WatchDog = ();
+    type Core = ();
 
     fn mpu(&self) -> &Self::MPU {
         &self.pmp
@@ -168,6 +169,10 @@ impl<'a, A: 'static + Alarm<'static>, I: InterruptService<()> + 'a> kernel::Chip
 
     unsafe fn print_state(&self, writer: &mut dyn Write) {
         rv32i::print_riscv_state(writer);
+    }
+
+    fn current_core(&self) -> &Self::Core {
+        &()
     }
 }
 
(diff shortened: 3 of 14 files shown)


brghena:
Great to see people thinking about this! The nRF53 series is also dual-core (one for applications and one for managing the radio), and I expect this scenario to become common on top-tier microcontrollers moving forward.

My thoughts:

We had discussed how to do multi-core Tock before in the context of multi-microcontroller platforms (PCBs with multiple microcontrollers and a shared bus between them). We were mostly focused on the "Separated Kernels" approach you pointed out above. It seems that system calls should be able to be routed externally between the microcontrollers to make such a system work. Conceptually, applications should be able to run on either core, and just pass system calls back-and-forth. That would mean that drivers themselves wouldn't have to change, I think.

For the "SMP" approach, I'd be concerned that it would be fundamental change for Tock to have multiple simultaneous threads. I think it could be a significant amount of effort to add.

For the "Dedicated Core" method, it's also not intuitive to me how multi-microcontroller boards or the nRF53 would work. For both scenarios, there are some peripherals only accessible from one of the cores, which implies to me that you need at least part of an OS kernel running on each microcontroller.

hudson-ayers:
+1 to Brandon's comments.

I know @TPackard is working on a Tock port for the nrf53840, and is pursuing the "separate kernel" approach.

I think that SMP would be a lot of work to get right, especially in the case of heterogeneous cores. I also think it would be very difficult to modify the kernel to support this in a way that is 0-cost to boards with only a single core.

Re: "dedicated core", I don't understand how the single process on the dedicated core would issue system calls to standalone drivers without having another kernel running on that core. If it is issuing system calls to the kernel running on the other core, then it seems that process does in fact need to be scheduled alongside other processes to some extent (there will be decisions to be made about when to service system calls from this second core, and when to deliver results back). I suppose this would require some custom capsule on the board to expose a fully asynchronous system call interface to the app running on the second core?

lschuermann:
I don't have much to add to @brghena's and @hudson-ayers's comments. I really like the idea, though I think SMP would be particular challenging for Tock because many of the abstractions have been carefully designed specifically with the constraint that there may only ever be a single thread of execution, at least in the kernel (and _if_ I understand the "dedicated core" approach correctly, this might apply too).

Apart from the obvious cases where Rust would actively help us hunt the parts down which aren't synchronized (e.g. using `Cell` which is not thread-safe), there are also the subtle interactions between the kernel and processes over the system call layer making this tricky. Take for example an allowed slice: currently, we construct a slice from a pointer when we dereference the `AppSlice` in a capsule, and that is only sound Rust because we (1) are not persistently storing the `&mut [u8]` slice type in the kernel (only a `*mut u8` pointer) when switching to an app and (2) an app cannot modify the buffer while the kernel is accessing it (because it can't currently run in parallel). If a process would be running on one core, and a piece of kernel code on the other, we'd therefore need to prevent a capsule from dereferencing any `AppSlice` belonging to a running process and can't schedule processes where an `AppSlice` is currently dereferenced to a `&mut [u8]`. This is just an arbitrary example to outline why this might be tricky -- but it'd be super cool if we could make it work _and_ not have the single core be less efficient.

alexandruradovici:
Thank you for the comments. I took a closer look to Tock's kernel, and I was thinking about the following approach that seems to come close to SMP:

1.There can only be once instance of the kernel running at once, regardless on which core it runs. This would not break Tock assumptions, except grants. The kernel would use a spinlock for this.
2. When accessing a process' AppSlice, if the process that shared that slice is currently running, ask the other core to preempt it using an interrupt (this might be a little tricky) and spinlock (with a timeout?) until this happens.
3. Route all interrupts (except the preempt ones) to one core.

This is might be a little less efficient than having two separate kernels running, but does make it usable for any number of cores.

I think I covered most of the situations, but I might be missing some aspects.

Acquire before starting the loop
```rust
chip.watchdog().setup();
        spinlock.acquire ();
        loop {
            chip.watchdog().tickle();
```

Release before switching to a process, acquire when getting back
```rust
spinlock.release ();
scheduler_timer.arm();
let context_switch_reason = process.switch_to();
scheduler_timer.disarm();
spinlock.acquire ();
```
Release before sleeping, acquire when waking
```rust
spinlock.release();
chip.watchdog().suspend();
chip.sleep();
chip.watchdog().resume();
spinlock.acquire();
```

jrvanwhy:
Re: https://github.com/tock/tock/pull/2539#issuecomment-822796231 (Alexandru's most recent comment)

Tock design question: does Tock want to support shared-memory IPC, where two processes simultaneously have RW access to the same region of RAM? If yes, then we will need additional design work.

bradjc:
Moved to #2873.

PR #2540, opened 2021-04-18
MSP432 SPI driver

LosDeiblos:
### Pull Request Overview

This PR adds support for SPI on the MSP432. The SPI related registers in the USCI modules are identical, but the usage of certain bits is different to UART and I2C mode. As a solution I simply cast the concerned registers to UCSPIx-registers instead of using the UCAx/UCBx-registers. With this trick the references of UCAx/UCBx-registers can be passed to the SPI-constructor.

### Testing Strategy

Tested on the evaluation board using a the spi_controller_transfer example as application.

### TODO or Help Wanted

* Is the approach of casting the UCAx/UCBx-registers to a common type ok/fine? Or should I rewrite the driver to use a different approach?
* I tested the SPI driver also using a [Waveshare 2.7inch E-Ink display](https://www.waveshare.com/w/upload/2/2d/2.7inch-e-paper-Specification.pdf) as a slave. Is there an interest for a driver-support for this E-Ink display? If yes, I could create a new PR for it.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/msp_exp432p401r/Makefile
+++ b/boards/msp_exp432p401r/Makefile
@@ -19,7 +19,7 @@ flash-debug: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
 	$(OPENOCD) $(OPENOCD_OPTIONS) -c "init; reset halt; flash write_image erase $<; verify_image $<; reset; shutdown"
 
 .PHONY: flash
-flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
+flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
 	$(OPENOCD) $(OPENOCD_OPTIONS) -c "init; reset halt; flash write_image erase $<; verify_image $<; reset; shutdown"
 
 .PHONY: flash-app
--- a/chips/msp432/src/spi.rs
+++ b/chips/msp432/src/spi.rs
@@ -1,3 +1,10 @@
+//! Serial Peripheral Interface (SPI)
+//!
+//! Implementation of DMA-based SPI master communication for the MSP432.
+//! Both eUSCI_A and eUSCI_B can be used for SPI communication, but the
+//! the registers have a different meaning than in UART/I2C mode.
+//!
+
 use crate::dma;
 use crate::usci::{self, UCSPIxCTLW0, UCSPIxIE, UCSPIxIFG, UCSPIxSTATW};
 use core::cell::Cell;
--- a/chips/msp432/src/spi.rs
+++ b/chips/msp432/src/spi.rs
@@ -1,0 +1,469 @@
+use crate::dma;
+use crate::usci::{self, UCSPIxCTLW0, UCSPIxIE, UCSPIxIFG, UCSPIxSTATW};
+use core::cell::Cell;
+use kernel::common::{
+    cells::{OptionalCell, TakeCell},
+    registers::{ReadOnly, ReadWrite},
+    StaticRef,
+};
+use kernel::hil::gpio::Pin;
+use kernel::hil::spi;
+use kernel::ErrorCode;
+
+/// The SPI related registers are identical between the USCI_A and the USCI_B module, but the usage
+/// of the certain bits is different to UART and I2C mode. Thus we simply cast the concerned
+/// registers to a UCSPIx-registers instead of UCAx or UCBx registers. With this trick we can pass
+/// references of USCI_A and USCI_B modules to the Spi-constructor.
+pub trait UsciSpiRef {
+    fn ctlw0(&self) -> &ReadWrite<u16, UCSPIxCTLW0::Register>;
+    fn brw(&self) -> &ReadWrite<u16>;
+    fn statw(&self) -> &ReadWrite<u16, UCSPIxSTATW::Register>;
+    fn rxbuf(&self) -> &ReadOnly<u16>;
+    fn txbuf(&self) -> &ReadWrite<u16>;
+    fn ie(&self) -> &ReadWrite<u16, UCSPIxIE::Register>;
+    fn ifg(&self) -> &ReadWrite<u16, UCSPIxIFG::Register>;
+}
+
+impl UsciSpiRef for StaticRef<usci::UsciBRegisters> {
+    fn ctlw0(&self) -> &ReadWrite<u16, usci::UCSPIxCTLW0::Register> {
+        unsafe {
+            &*(&self.ctlw0 as *const ReadWrite<u16, usci::UCBxCTLW0::Register>
+                as *const ReadWrite<u16, usci::UCSPIxCTLW0::Register>)
+        }
+    }
+
+    fn brw(&self) -> &ReadWrite<u16> {
+        &self.brw
+    }
+
+    fn statw(&self) -> &ReadWrite<u16, usci::UCSPIxSTATW::Register> {
+        unsafe {
+            &*(&self.statw as *const ReadWrite<u16, usci::UCBxSTATW::Register>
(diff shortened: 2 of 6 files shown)


boards/msp_exp432p401r/Makefile
```diff
@@ -19,7 +19,7 @@ flash-debug: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).elf
 	$(OPENOCD) $(OPENOCD_OPTIONS) -c "init; reset halt; flash write_image erase $<; verify_image $<; reset; shutdown"
 
 .PHONY: flash
-flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
+flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
```
bradjc, line 22: Is this change needed?

chips/msp432/src/spi.rs
```diff
@@ -0,0 +1,469 @@
+use crate::dma;
```
bradjc, line 1: Please add `//!` comment.

bradjc (approved):
This looks pretty straightforward to me. I do have a question about the Makefile change.

bradjc:
>     * I tested the SPI driver also using a [Waveshare 2.7inch E-Ink display](https://www.waveshare.com/w/upload/2/2d/2.7inch-e-paper-Specification.pdf) as a slave. Is there an interest for a driver-support for this E-Ink display? If yes, I could create a new PR for it.

Yes! I was thinking about wanting an e-ink display last week, in fact.

chips/msp432/src/usci.rs
```diff
@@ -102,6 +102,147 @@ register_structs! {
...
+            /// Enabled. eUSCI logic held in reset state
+            Enabled = 1
+        ],
+        /// STE mode select in master mode
+        UCSTEM OFFSET(1) NUMBITS(1) [
+            /// STE pin is used to prevent conflicts with other masters
+            PreventConflicts = 0,
+            /// STE pin is used to generate the enable signal for a 4-wire slave
+            ChipSelect = 1
+        ],
```
hudson-ayers, line 119: I am trying to verify the safety of the transmute used in `spi.rs`. I am confused, however, because this bitfield is clearly different than the bit described at offset 1 in `UCBxCTLW0`. I suspect it is because I do not understand how these registers work on the MSP432. Is the idea that these registers have a totally different meaning depending on if the module is configured for SPI vs. if it is configured for I2C? And thus the bitfield descriptions are different in that case?

boards/msp_exp432p401r/Makefile
LosDeiblos: During testing I noticed that flashing the kernel using `make flash` overwrites the .apps section, which requires to also flash the application again. This is probably the case, because the elf file contains a 0-byte as a placeholder in section .apps. As far as I understand the .bin file doesn't contain this placeholder. Therefore I changed the Makefile to flash the binary instead of the elf file, which allows to flash the kernel independently from the application.

chips/msp432/src/spi.rs
LosDeiblos: Added the comments now :)

LosDeiblos:
> > ```
> > * I tested the SPI driver also using a [Waveshare 2.7inch E-Ink display](https://www.waveshare.com/w/upload/2/2d/2.7inch-e-paper-Specification.pdf) as a slave. Is there an interest for a driver-support for this E-Ink display? If yes, I could create a new PR for it.
> > ```
> 
> Yes! I was thinking about wanting an e-ink display last week, in fact.

Great! Then I will add it hopefully very soon. I assume that I should use the screen trait for it?

chips/msp432/src/spi.rs
```diff
@@ -0,0 +1,476 @@
...
+/// references of USCI_A and USCI_B modules to the Spi-constructor.
+pub trait UsciSpiRef {
+    fn ctlw0(&self) -> &ReadWrite<u16, UCSPIxCTLW0::Register>;
+    fn brw(&self) -> &ReadWrite<u16>;
+    fn statw(&self) -> &ReadWrite<u16, UCSPIxSTATW::Register>;
+    fn rxbuf(&self) -> &ReadOnly<u16>;
+    fn txbuf(&self) -> &ReadWrite<u16>;
+    fn ie(&self) -> &ReadWrite<u16, UCSPIxIE::Register>;
+    fn ifg(&self) -> &ReadWrite<u16, UCSPIxIFG::Register>;
+}
```
phil-levis, line 32: This seems like a reasonable approach, but could perhaps be explained better. You aren't casting, you are encapsulating. I.e., the two actual bit mappings are different, so you define a common abstraction (`UsciSpiRef`, which should perhaps be `UsciSpiRegisters`), such that the SPI implementation can operate on top of either one.

I would suggest having the Spi type be a type parameter, rather than a `dyn`, though. We've found that this reduces code size and reduces overheads.

chips/msp432/src/spi.rs
```diff
@@ -0,0 +1,476 @@
...
+    tx_dma_src: u8,
+
+    rx_dma: OptionalCell<&'a dma::DmaChannel<'a>>,
+    pub(crate) rx_dma_chan: usize,
+    rx_dma_src: u8,
+}
+
+impl<'a> Spi<'a> {
+    pub fn new(
+        registers: &'static dyn UsciSpiRef,
```
phil-levis, line 189: I'd suggest parameterizing Spi by the `UsciSpiRef` implementation rather than having a `dyn`.

chips/msp432/src/spi.rs
hudson-ayers: The underlying type is in fact transmuted in the implementation of this trait.

chips/msp432/src/spi.rs
```diff
@@ -0,0 +1,476 @@
...
+    fn ifg(&self) -> &ReadWrite<u16, UCSPIxIFG::Register>;
+}
+
+impl UsciSpiRef for StaticRef<usci::UsciBRegisters> {
+    fn ctlw0(&self) -> &ReadWrite<u16, usci::UCSPIxCTLW0::Register> {
+        unsafe {
+            &*(&self.ctlw0 as *const ReadWrite<u16, usci::UCBxCTLW0::Register>
+                as *const ReadWrite<u16, usci::UCSPIxCTLW0::Register>)
+        }
+    }
```
hudson-ayers, line 40: This makes me somewhat uncomfortable. I understand the approach and why it works, but this is basically a `transmute` and my experience is that proving a transmute like this will not cause UB is not always as straightforward as it seems.

At the very least I think there should be a `// SAFETY: ` comment above this describing why it is a safe operation. If @jrvanwhy has a chance to weigh in that might also be nice.

bradjc (changes requested):
I had approved, but I didn't look closely enough for the unsafe use. That warrants a closer look because 1) problems from unsafe are often very subtle, 2) unsafe can be difficult to remove later and we want to avoid setting examples that others then copy.

lschuermann:
I'm wondering whether it might be a good idea to have a wrapper around a `tock-registers` register, which can be used to treat a register as having a different `RegisterLongName`? From a safety perspective that should be totally fine, given that on registers we can simply set a raw value directly. Especially with #2517, we should be able to rather seamlessly integrate something like this and then derive the accessor traits accordingly. I'll create a PR once #2517 makes it in.

Depending on how long #2517 will take and whether such an "RegisterAlias" wrapper-type is actually desired, I'd vouch for not using `transmute`, but instead work around this by constructing a `LocalRegisterCopy` to use the `tock-registers` interface and read/write raw values using `get()` / `set()`.

phil-levis:
This has been waiting for 8 months -- closing it.

PR #2541, opened 2021-04-19
Fix NRF52 802.15.4 radio bug on CCA busy + add UDP/802154 driver to Nano 33

alevy:
### Pull Request Overview

This pull request accomplishes two main things:

1. Add 802.15.4 & the UDP driver to the Nano 33 Sense board (thanks @hudson-ayers!)
2. Fixes a bug in the nrf52 ieee802154 driver that was causing a state machine failure (i.e. the driver just got stuck) upon frame transmission when CCA fails.

#### What was the bug and how did you fix it?

Very simply, if ever in preparation for a transmission, the radio performed a CCA check and it failed (i.e. someone else is already transmitting on that channel), it's supposed to wait a specified period of time then retry. What was happening prior to this PR is there would never be a retry. So, if there was ever a CCA check failure, there could be no more transmission/receptions.

The problem was that while the timer was setup properly, the radio never got reconfigured in response to the timer expiring. I don't _actually_ know the exact reason the strategy coded in the previous coding wasn't working, however, I just went with a more straight forward (though in the long run, worse) approach. At least it works.

Instead of using the Prorgammable Peripheral Interconnect on the NRF52's
ieee802154 radio driver to retry CCA after a timer expires when it
fails, use simple software logic instead. PPI invokes tasks in one
peripheral (the radio) in response to an event in another (the timer),
without requiring an interrupts serviced in software.

The best thing to do would be to use PPI as well as shortcuts to
simplify the radio logic significantly, as well as likely improve
transition time performance. I'll leave that for another adventure.

#### Why is this PR touching so many files?

Most of the changes are small refactoring in the NRF52-based boards. The NRF52's ieee802154 driver no longer requires a PPI, which had been passed down all the way from the board (it's not clear this was ever a good thing, this was required even for boards that used an nrf52 with no ieee802154 driver). So now, each board that relies on any nrf52 chip has a couple line changes removing the PPI initializing.

### Testing Strategy

If anyone has ideas on how to test this reliably, please share. Meanwhile, I tested this by running the `examples/tests/udp/udp_send` app from `libtock-c` with a much smaller interval than the default (I used 50 ms) and waiting for a CCA to fail (probably because randomly some other 2.4GHz device was transmitting---maybe those Philips Hues I have are not completely useless after all). Without this change eventually transmissions stop. With the changes, they continue.

To help my debugging I added some debug statements in the kernel to identify when a CCA failed, confirming that (a) this was the problem and (b) that my fix addresses the problem. You might consider adding similar debug statements if you want to test---again, it's otherwise pretty hard to tell whether you got a collision.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nano33ble/src/main.rs
+++ b/boards/nano33ble/src/main.rs
@@ -77,7 +77,7 @@ pub mod io;
 
 // State for loading and holding applications.
 // How should the kernel respond when a process faults.
-const FAULT_RESPONSE: kernel::procs::FaultResponse = kernel::procs::FaultResponse::Stop;
+const FAULT_RESPONSE: kernel::procs::FaultResponse = kernel::procs::FaultResponse::Panic;
 
 // Number of concurrent processes this platform supports.
 const NUM_PROCS: usize = 8;
--- a/boards/nano33ble/src/main.rs
+++ b/boards/nano33ble/src/main.rs
@@ -159,11 +159,10 @@ impl kernel::Platform for Platform {
 /// these static_inits is wasted.
 #[inline(never)]
 unsafe fn get_peripherals() -> &'static mut Nrf52840DefaultPeripherals<'static> {
-    let ppi = static_init!(nrf52840::ppi::Ppi, nrf52840::ppi::Ppi::new());
     // Initialize chip peripheral drivers
     let nrf52840_peripherals = static_init!(
         Nrf52840DefaultPeripherals,
-        Nrf52840DefaultPeripherals::new(ppi)
+        Nrf52840DefaultPeripherals::new()
     );
 
     nrf52840_peripherals
--- a/boards/nano33ble/src/main.rs
+++ b/boards/nano33ble/src/main.rs
@@ -63,15 +63,21 @@ const I2C_PULLUP_PIN: Pin = Pin::P1_00;
 /// Interrupt pin for the APDS9960 sensor.
 const APDS9960_PIN: Pin = Pin::P0_19;
 
+// Constants related to the configuration of the 15.4 network stack
 /// Personal Area Network ID for the IEEE 802.15.4 radio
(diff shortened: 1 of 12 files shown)


boards/nano33ble/src/main.rs
```diff
@@ -63,15 +63,21 @@ const I2C_PULLUP_PIN: Pin = Pin::P1_00;
...
+const DEFAULT_CTX_PREFIX_LEN: u8 = 8; //Length of context for 6LoWPAN compression
+const DEFAULT_CTX_PREFIX: [u8; 16] = [0x0 as u8; 16]; //Context for 6LoWPAN Compression
 
 /// UART Writer for panic!()s.
 pub mod io;
 
 // State for loading and holding applications.
 // How should the kernel respond when a process faults.
-const FAULT_RESPONSE: kernel::procs::FaultResponse = kernel::procs::FaultResponse::Panic;
+const FAULT_RESPONSE: kernel::procs::FaultResponse = kernel::procs::FaultResponse::Stop;
```
hudson-ayers, line 80: Did you mean to include this? I think it makes sense for the nano33 given that early panics can bork the USB stack, but it maybe should be a separate PR (or at least called out in the description)

hudson-ayers (approved):
This generally looks good to me! Not passing the ppi for all the nordic chips is an added bonus.

boards/nano33ble/src/main.rs
alevy: I did not! Good catch!

alevy:
```suggestion
const FAULT_RESPONSE: kernel::procs::FaultResponse = kernel::procs::FaultResponse::Panic;
```

hudson-ayers (approved):

PR #2542, opened 2021-04-19
Change the Nano33BLE to the StopWithDebug fault response

bradjc:
### Pull Request Overview

This PR makes the `StopWithDebug` the default fault policy on Nano 33 BLE because if a process fails immediately, the USB/CDC stack doesn't have time to setup, and the panic print message doesn't work. This avoids that issue by having the kernel just stop the process and notify the user via a debug message that it faulted. Then the user can use process console to panic the kernel and see the debug message if they wish.

Replaces #2527.

~~Blocked on #2537.~~

### Testing Strategy

Trying crash apps on the nano33ble.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nano33ble/src/main.rs
+++ b/boards/nano33ble/src/main.rs
@@ -75,13 +75,17 @@ const DEFAULT_CTX_PREFIX: [u8; 16] = [0x0 as u8; 16]; //Context for 6LoWPAN Comp
 /// UART Writer for panic!()s.
 pub mod io;
 
-// State for loading and holding applications.
-// How should the kernel respond when a process faults.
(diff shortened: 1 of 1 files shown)


lschuermann (approved):

alistair23 (approved):

lschuermann:
This seems like a very uncontroversial change, especially now that `StopWithDebugFaultPolicy` is already in the kernel.
Also it has 3 approvals and been open for 4 days, so I'm taking the liberty to try my first

bors r+

Hope this works and doesn't break :)

PR #2543, opened 2021-04-19
Update micro:bit readme

alexandruradovici:
### Pull Request Overview

This pull request updates the microbit's Readme
 - fixes the memory layout description
 - updates the tockloader command used without the bootloader 

### Testing Strategy

N/A

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/microbit_v2/README.md
+++ b/boards/microbit_v2/README.md
@@ -90,7 +90,7 @@ The kernel memory layout is different if there is no bootloader. Change the `lay
 MEMORY
 {
   # with bootloader
-  # rom (rx)  : ORIGIN = 0x00010000, LENGTH = 192K
+  # rom (rx)  : ORIGIN = 0x00008000, LENGTH = 192K
   # without bootloader
   rom (rx)  : ORIGIN = 0x00000000, LENGTH = 256K
   prog (rx) : ORIGIN = 0x00040000, LENGTH = 256K
@@ -123,5 +123,7 @@ $ tockloader install app.tab
 
 Without bootloader
 ```bash
-$ tockloader --openocd --board microbit_v2 install app.tab
+$ tockloader --openocd --board microbit_v2 --bundle-apps install app.tab
 ```
+
+> `--bundle-apps` seems to be needed due to an [openocd issue](https://github.com/tock/tockloader/issues/67)
--- a/boards/microbit_v2/README.md
+++ b/boards/microbit_v2/README.md
@@ -48,12 +48,6 @@ Use the `make flash-bootloader` command to flash [Tock Bootloader](https://githu
 $ make flash-bootloader
 ```
 
-> If you are using an `openocd` 0.11 or newer comment out the following lines in `openocd.cfg`
-> ```
-> # flash bank $_CHIPNAME.flash nrf51 0x00000000 0 1 1 $_TARGETNAME
-> # flash bank $_CHIPNAME.uicr nrf51 0x10001000 0 1 1 $_TARGETNAME
-> ```
-
 ## Uploading the kernal
 
 Make sure you have flashed [Tock Bootloader](https://github.com/tock/tock-bootloader) to the board.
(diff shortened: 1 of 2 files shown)


boards/microbit_v2/README.md
```diff
@@ -90,7 +90,7 @@ The kernel memory layout is different if there is no bootloader. Change the `lay
 MEMORY
 {
   # with bootloader
-  # rom (rx)  : ORIGIN = 0x00010000, LENGTH = 192K
+  # rom (rx)  : ORIGIN = 0x00008000, LENGTH = 192K
```
lschuermann, line 93: This seems like it's going to be outdated again in no time. Do we need to replicate these values here or could it be easier to just point to the commented out lines in `layout.ld` directly, perhaps using a Markdown link?

lschuermann (dismissed):

boards/microbit_v2/README.md
alexandruradovici: Putting a link would be ok, just that people tend not to follow it. It might make it more obvious this way.

lschuermann: Sure, works for me.

boards/microbit_v2/openocd.cfg
```diff
@@ -6,6 +6,8 @@ set WORKAREASIZE 0x40000
 
 $_TARGETNAME configure -work-area-phys 0x20000000 -work-area-size $WORKAREASIZE -work-area-backup 0
 
+# comment the following lines if you are using openocd 0.11 or newer
```
brghena, line 9: I think we can use if statements in cfg files. Is there a way to automatically include these lines based on version number?

Alternatively, should they be disabled by default, since new people will install openocd version 0.11.

brghena: Also, why are we referencing the nrf51 here anyways?

alexandruradovici: I don't know, this seems to work. I used the configuration file provided by the Arduino IDE plugin for the Micro:bit.

> I think we can use if statements in cfg files. Is there a way to automatically include these lines based on version number?
> 
Still looking into this, but I don't seem to find a lot of documentation.
> Alternatively, should they be disabled by default, since new people will install openocd version 0.11.
It was released in March, Ubuntu still provides 0.10. The 0.11 si in rc-2.

alexandruradovici: > Also, why are we referencing the nrf51 here anyways?

I used the configuration file from the Arduino IDE plugin for Micro:bit. This seems to work.

alexandruradovici: Fixed it, I placed a catch around the two lines that show errors. It will not print an error, but work. I suggest we delete them after a while, when 0.11 becomes more popular.

boards/microbit_v2/openocd.cfg
```diff
@@ -6,8 +6,9 @@ set WORKAREASIZE 0x40000
...
 $_TARGETNAME configure -work-area-phys 0x20000000 -work-area-size $WORKAREASIZE -work-area-backup 0
 
-# comment the following lines if you are using openocd 0.11 or newer
+# catch is necessary to be backward compatible with openocd 0.10
+# this should be removed when 0.11 becomes more widely used
 
-flash bank $_CHIPNAME.flash nrf51 0x00000000 0 1 1 $_TARGETNAME
-flash bank $_CHIPNAME.uicr nrf51 0x10001000 0 1 1 $_TARGETNAME
+# catch { flash bank $_CHIPNAME.flash nrf51 0x00000000 0 1 1 $_TARGETNAME } err
+# catch { flash bank $_CHIPNAME.uicr nrf51 0x10001000 0 1 1 $_TARGETNAME } err
```
brghena, line 13: Did you mean to leave these commented out?

boards/microbit_v2/openocd.cfg
brghena: Awesome. That's exactly what I was thinking.

boards/microbit_v2/openocd.cfg
alexandruradovici: Of course :)

brghena (approved):

brghena:
I confirm that this works with a microbit on openocd 0.10.0.

I also _think_ it works with 0.11.0. The output is a little different, but I think it's flashing code fine. The `catch` statements definitely work.

PR #2545, opened 2021-04-21
Kernel "platform" module reorganization

bradjc:
### Pull Request Overview

The structure of `/kernel` is ad-hoc, and has slowly been amended such that:

1. Some things are exported as `kernel::Thing`, and others `kernel::module::Thing`, without really a reason why.
2. The organization is not particularly descriptive.

While I am normally in favor of leaving things alone, since I already split up process.rs into several files, 2.0 seems like the time to restructure `/kernel` in a way that makes sense and we can build on in the future.

Doing this restructure in one PR is scary because it is likely to take a while to agree on the right naming and structure, and it often requires changing a lot of files to change `use` commands. So this is going to be a bit piecemeal in mechanism, but hopefully single-shot in terms of the new organization and plan.

This PR looks at the `platform` module in kernel. The changes I made:

- See #2551 for the plan.

### Testing Strategy

travis. No code changes, just reorganizations.

### TODO or Help Wanted

Todo:

- Fixup other files. So far only hail compiles.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alistair23 (dismissed):

hudson-ayers:
These changes generally look good to me. I would be in favor of the proposed `Platform` --> `Board` rename, as it is not clear to me how they are different.

I think `traits` is fine, though `interfaces` might make more sense to a non-Rust person perusing the codebase. Don't feel strongly about this.

bradjc:
> Given that the things in `platform` are re-used traits, and we have traits that are not in `platform`, why not just move these into the main kernel crate? Or is the point that these are APIs for the kernel to interact with a board? A package that has 3 traits in it seems a little overkill. Or is the intention that through the directory structure one shows what a board has to implement in order for the kernel to work properly?

I think the grouping you described is correct: platform contains interfaces between the kernel crate and other crates that other crates must implement, whereas the root crate has interfaces/code used internally or that are implemented by kernel and exposed to other crates.

bradjc:
I've updated this PR with the discussion from #2551. It needs to be ported to all other chips/boards, but I want to give one more chance to look at the names and structure of `KernelResources`.

bradjc:
This is really the heart of this PR, so probably the place to start:

```rust
pub trait KernelResources<C: Chip> {
    type SystemCallDispatch: SystemCallDispatch;
    type SystemCallFilter: SystemCallFilter;
    type ProcessFault: ProcessFault;
    type Scheduler: Scheduler<C>;
    type SchedulerTimer: scheduler_timer::SchedulerTimer;
    type WatchDog: watchdog::WatchDog;

    fn system_call_dispatch(&self) -> &Self::SystemCallDispatch;
    fn system_call_filter(&self) -> &Self::SystemCallFilter;
    fn process_fault(&self) -> &Self::ProcessFault;
    fn scheduler(&self) -> &Self::Scheduler;
    fn scheduler_timer(&self) -> &Self::SchedulerTimer;
    fn watchdog(&self) -> &Self::WatchDog;
}
```

Boards implement `KernelResources` and pass it to `kernel_loop`. You can see an example as I have implemented it for hail.

hudson-ayers:
I really like the `KernelResources` approach and think the implementation for Hail looks good.

phil-levis:
This looks good to me.

bradjc:
Replaced by #2659

PR #2547, opened 2021-04-22
capsules/i2c_master_slave_driver: use grant, enforce single process

hudson-ayers:
### Pull Request Overview

This pull request enforces that only a single process may (implicitly) reserve the nonvirtualized userspace driver. Also, the driver now store callbacks and appslices in grant regions.

This is one of the remaining capsules blocking #2462 .

One observation: repeatedly handling the possibility of a grant entrance (within a single function) failing is a pain. But I do not think it is possible for a grant entrance to fail after the first entrance -- the grant is already allocated at that point, and the process is present. So I omitted the error handling in `command()` after the first grant entrance.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/i2c_master_slave_driver.rs
+++ b/capsules/src/i2c_master_slave_driver.rs
@@ -13,10 +13,10 @@
 
 use core::cell::Cell;
 use core::cmp;
-use kernel::common::cells::{MapCell, TakeCell};
+use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::hil;
 use kernel::{CommandReturn, ProcessId, Upcall};
-use kernel::{Driver, ErrorCode, Read, ReadWrite, ReadWriteAppSlice};
+use kernel::{Driver, ErrorCode, Grant, Read, ReadWrite, ReadWriteAppSlice};
 
 pub static mut BUFFER1: [u8; 256] = [0; 256];
 pub static mut BUFFER2: [u8; 256] = [0; 256];
@@ -49,7 +49,8 @@ pub struct I2CMasterSlaveDriver<'a> {
     master_buffer: TakeCell<'static, [u8]>,
     slave_buffer1: TakeCell<'static, [u8]>,
     slave_buffer2: TakeCell<'static, [u8]>,
-    app: MapCell<App>,
+    app: OptionalCell<ProcessId>,
+    apps: Grant<App>,
 }
 
 impl<'a> I2CMasterSlaveDriver<'a> {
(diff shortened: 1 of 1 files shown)


bradjc (approved):

capsules/src/i2c_master_slave_driver.rs
```diff
@@ -222,73 +237,133 @@ impl hil::i2c::I2CHwSlaveClient for I2CMasterSlaveDriver<'_> {
...
+            self.apps
+                .enter(*current_process, |_| current_process == &process_id)
+                .unwrap_or(true)
+        });
+        if match_or_empty_or_nonexistant {
+            self.app.set(process_id);
+        } else {
+            return Err((slice, ErrorCode::NOMEM));
+        }
+        let app = self.app.expect("cant fail");
```
lschuermann, line 256: Isn't `app` always equivalent to `process_id` here, given the logic above? Then I suppose we could just use that and get rid of the `expect`.

capsules/src/i2c_master_slave_driver.rs
```diff
@@ -222,73 +237,133 @@ impl hil::i2c::I2CHwSlaveClient for I2CMasterSlaveDriver<'_> {
...
-            }
-            // Pass in a buffer for handling a write issued by another I2C master.
-            3 => {
-                self.app.map(|app| {
-                    core::mem::swap(&mut app.slave_rx_buffer, &mut slice);
-                });
-                Ok(slice)
-            }
-            _ => Err((slice, ErrorCode::NOSUPPORT)),
+        // Check if this non-virtualized driver is already in use by
```
lschuermann, line 244: The other drivers seem to only do the reservation checks on a successful `command` invocation. Do we want to do this everywhere (i.e. subscribe, allow_read{only,write}). This _might_ be beneficial in that we're not necessarily entering the process' Grant region (and thus allocating its Grant region) when we're refusing the operation.

If so, we might want to be consistent and update the other changed drivers (#2518, #2521) as well.

lschuermann (commented):
Overall this looks good, just two minor issues/questions.

capsules/src/i2c_master_slave_driver.rs
hudson-ayers: good call, done

capsules/src/i2c_master_slave_driver.rs
hudson-ayers: Good catch here as well. Based on the discussion on the core call, I removed those extra checks from this driver.

alistair23 (approved):

capsules/src/i2c_master_slave_driver.rs
lschuermann: Did you push this already?

capsules/src/i2c_master_slave_driver.rs
```diff
@@ -222,73 +237,107 @@ impl hil::i2c::I2CHwSlaveClient for I2CMasterSlaveDriver<'_> {
...
-            3 => {
-                self.app.map(|app| {
-                    core::mem::swap(&mut app.slave_rx_buffer, &mut slice);
-                });
-                Ok(slice)
-            }
-            _ => Err((slice, ErrorCode::NOSUPPORT)),
+        let res = self
+            .apps
+            .enter(app, |app| {
```
lschuermann, line 246: Nit: this technically works, though the `ProcessId` and `Grant` region are related but different things and I fear that shadowing the `app` variable might be confusing. Feel free to not change this if not considered an issue.

bradjc (approved):

Issue #1088, opened 2018-07-06
Tracking: Revisit what is exported from the kernel crate


bradjc:
The kernel crate contains many unsafe and "dangerous" (i.e. they must be used carefully or they invalidate Tock safety guarantees) data structures. We need to be careful with exactly how these are exposed outside of the kernel crate.

- [x] Restrict which modules are public. (#955)
- [x] Make internal use only functions limited to just the kernel crate (using the `crate` modifier).
- [ ] Evaluate and justify that functions which wrap a single `unsafe` call are safe.
- [ ] Mark only functions which violate Rust type safety as `unsafe`.
- [ ] Require a capability to call all "dangerous" functions.
  - [x] Add capabilities feature.

bradjc:
I came across an instance in Tock where this might be complicated: capsules can get access to `process` objects by calling `kernel.process_each_capability()`. Now, iterating the list of processes requires a capability, but once they have the process object they can call any function in the `ProcessType` trait (that isn't marked unsafe).

phil-levis:
> I came across an instance in Tock where this might be complicated: capsules can get access to `process` objects by calling `kernel.process_each_capability()`. Now, iterating the list of processes requires a capability, but once they have the process object they can call any function in the `ProcessType` trait (that isn't marked unsafe).

This was intentional: the idea was that the ProcessManagementCapability meant a (non-crate) module could access and manage processes. If we want to have finer grained control, we can add individual capabilities to certain `ProcessType` operations, or break up the trait into smaller traits with narrower functionality. The point was to make it possible for capsules (e.g., the process console) to be able to access and interact with processes.

bradjc:
2.0 seems like an opportunity to actually handle this issue.

Proposal: we remove all `pub use crate::` lines from kernel/src/lib.rs and use pub and pub(crate) modifiers in each file to determine visibility. We make special exceptions for objects that we don't want to be namespaced (perhaps `Kernel`).

https://github.com/tock/tock/blob/fb130b4b668a1e60514da85fc4a94a05cd840797/kernel/src/lib.rs#L112

hudson-ayers:
Why do we have to remove the `pub use` declarations for that? If you tried to mark `AppId` as `pub(crate)` instead of the current `pub` it would fail to compile (you can't `pub use` a non-`pub` struct).

And I don't see a ton of benefit to exposing the internal kernel file structure to capsules, it just makes it a pain to rename files within the kernel, and adds verbosity to imports.

[...]

Issue #1998, opened 2020-07-03
Allow SysTick trait to be configured from boards crate


rajivr:
As mentioned [here](https://github.com/tock/tock/pull/1990#issuecomment-652218550), with SysTick trait is there seems to an assumption on being able to accurately get remaining time-slice in micro-seconds (in `get_remaining_us`). The way SysTick reference clock (STCLK) gets configured can vary depending on the tune-able parameters that chip vendor provides. Currently SysTick is initialized in `chip.rs` in the chip crate.

Clocks are usually setup in the board crate. We might have to consider some way allowing SysTick calibration to be done from board crate.

hudson-ayers:
Whatever mechanism is ultimately used to address this should also be used for the watchdog.

PR #2551, opened 2021-04-23
kernel: Reorganize crate exports

bradjc:
### Pull Request Overview

This pull request is a discussion point for addressing #1088.

Initial idea:

```rust
// Kernel crate resources exposed as `kernel::module::Type`.
pub mod appslice;
pub mod capabilities;
pub mod common;
pub mod component;
pub mod debug;
pub mod driver;
pub mod errorcode;
pub mod grant;
pub mod hil;
pub mod introspection;
pub mod ipc;
pub mod schedulers;
pub mod syscall;
pub mod traits;
pub mod upcall;

mod config;
mod memop;
mod process;
mod process_policies;
mod process_standard;
mod process_utilities;

// Core resources exposed as `kernel::Type`.
pub use crate::driver::Driver;
pub use crate::errorcode::ErrorCode;
pub use crate::kernel::Kernel;
pub use crate::process::ProcessId;
pub use crate::scheduler::Scheduler;

// Export only select items from the process module. To remove the name conflict
// this cannot be called `process`, so we use a shortened version. These
// functions and types are used by board files to setup the platform and setup
// processes.
/// Publicly available process-related objects.
pub mod procs {
    pub use crate::process::{
        Error, FaultAction, FunctionCall, FunctionCallSource, Process, State, Task,
    };
    pub use crate::process_policies::{
        PanicFaultPolicy, ProcessFaultPolicy, RestartFaultPolicy, StopFaultPolicy,
        StopWithDebugFaultPolicy, ThresholdRestartFaultPolicy,
        ThresholdRestartThenPanicFaultPolicy,
    };
    pub use crate::process_standard::ProcessStandard;
    pub use crate::process_utilities::{load_processes, ProcessLoadError};
}
```

Basically everything is exposed through the module (file) rather than being `kernel::Type` except for a few key types that are so essential seem to make sense as `kernel::X` types.

Other changes:

- `/platform` is now `/traits`.
  - Added `chip.rs` and `platform.rs` to `/traits`.
- `mem.rs` -> `appslice.rs`.
- `/sched` -> `/schedulers`.

Questions:

- Right now we have `kernel::procs::`. That's not my favorite. Should we do `kernel::processes::`?
- `trait Platform` doesn't match that it is implemented by what we call boards. Should we change it to `trait Board`?
- `trait Driver` is a very generic name which has many meanings in embedded code. Should we change it?

### Testing Strategy

travis

### TODO or Help Wanted

Discussion.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/kernel/src/lib.rs
+++ b/kernel/src/lib.rs
@@ -89,45 +89,36 @@
 #![warn(unreachable_pub)]
 #![no_std]
 
+// Kernel crate resources exposed as `kernel::module::Type`.
+pub mod appslice;
 pub mod capabilities;
 pub mod common;
 pub mod component;
 pub mod debug;
+pub mod driver;
+pub mod errorcode;
+pub mod grant;
 pub mod hil;
 pub mod introspection;
 pub mod ipc;
+pub mod schedulers;
 pub mod syscall;
+pub mod traits;
+pub mod upcall;
 
 mod config;
-mod driver;
-mod errorcode;
-mod grant;
-mod mem;
 mod memop;
-mod platform;
 mod process;
 mod process_policies;
 mod process_standard;
 mod process_utilities;
-mod sched;
-mod upcall;
 
-pub use crate::driver::{CommandReturn, Driver};
-pub use crate::errorcode::into_statuscode;
+// Core resources exposed as `kernel::Type`.
+pub use crate::driver::Driver;
 pub use crate::errorcode::ErrorCode;
-pub use crate::grant::{Grant, ProcessGrant};
-pub use crate::mem::{Read, ReadOnlyAppSlice, ReadWrite, ReadWriteAppSlice};
-pub use crate::platform::scheduler_timer::{SchedulerTimer, VirtualSchedulerTimer};
-pub use crate::platform::watchdog;
-pub use crate::platform::{mpu, Chip, InterruptService, Platform};
-pub use crate::platform::{ClockInterface, NoClockControl, NO_CLOCK_CONTROL};
+pub use crate::kernel::Kernel;
 pub use crate::process::ProcessId;
-pub use crate::sched::cooperative::{CoopProcessNode, CooperativeSched};
-pub use crate::sched::mlfq::{MLFQProcessNode, MLFQSched};
-pub use crate::sched::priority::PrioritySched;
-pub use crate::sched::round_robin::{RoundRobinProcessNode, RoundRobinSched};
-pub use crate::sched::{Kernel, Scheduler};
-pub use crate::upcall::Upcall;
+pub use crate::scheduler::Scheduler;
 
 // Export only select items from the process module. To remove the name conflict
 // this cannot be called `process`, so we use a shortened version. These


hudson-ayers:
> Questions:
>
> Right now we have kernel::procs::. That's not my favorite. Should we do kernel::processes::?

I think procs is fine

> trait Platform doesn't match that it is implemented by what we call boards. Should we change it to trait Board?

I vote yes

> trait Driver is a very generic name which has many meanings in embedded code. Should we change it?

I vote `UserDriver`

#### Other comments

- One thing that is a little weird about `traits/` is how many traits are defined in `kernel/` that are not in `traits/`  -- for example, `Process`, `Scheduler`, `Debug`, `Driver`, `Component`, etc. What really distinguishes the traits in `platform/` today is that they are *hardware* interfaces that are not in `hil/`. I think that is partially to denote these interfaces as being higher importance than those in `hil/` because they are traits that are required to be implemented in order to construct a kernel. Thoughts on `platform_traits` or `platform_hil`? It is pretty long, but I think that combined with renaming `Platform` --> `Board` is maybe the most clear.

- Other than that, I generally think this looks great. If a PR implementing exactly what is described was implemented I would be happy to approve it and consider it a big improvement on what we have now

lschuermann:
> One thing that is a little weird about `traits/` is how many traits are defined in `kernel/` that are not in `traits/`  -- for example, `Process`, `Scheduler`, `Debug`, `Driver`, `Component`, etc. What really distinguishes the traits in `platform/` today is that they are _hardware_ interfaces that are not in `hil/`. I think that is partially to denote these interfaces as being higher importance than those in `hil/` because they are traits that are required to be implemented in order to construct a kernel. Thoughts on `platform_traits` or `platform_hil`? It is pretty long, but I think that combined with renaming `Platform` --> `Board` is maybe the most clear.

I agree on _traits_ being a bit tricky. In general, I fear naming a module after a language construct might cause various issues, if not technically at least semantically.

> > trait Platform doesn't match that it is implemented by what we call boards. Should we change it to trait Board?
> 
> I vote yes

While we have "boards" in the upstream repository, I imagine Tock may not run on _physical boards_ in production settings, but rather as a small part of a larger system (think a platform co-processor). I agree that Platform isn't a great name, but "boards" might be attached to the types of devices we do development on -- not necessarily what Tock would run on in production?

bradjc:
>     * One thing that is a little weird about `traits/` is how many traits are defined in `kernel/` that are not in `traits/`  -- for example, `Process`, `Scheduler`, `Debug`, `Driver`, `Component`, etc. What really distinguishes the traits in `platform/` today is that they are _hardware_ interfaces that are not in `hil/`. I think that is partially to denote these interfaces as being higher importance than those in `hil/` because they are traits that are required to be implemented in order to construct a kernel. Thoughts on `platform_traits` or `platform_hil`? It is pretty long, but I think that combined with renaming `Platform` --> `Board` is maybe the most clear.

Maybe leaving it at `platform` is the way to go. We can say "Platform" refers to the comprehensive hardware stack (everything from the arch to the chip to the board) that tock is running on. So in the kernel crate, "/platform" refers to the stuff that the kernel needs to know about or have control over that is hardware-specific.

That would, then, make it important to change `trait Platform`.





> > > trait Platform doesn't match that it is implemented by what we call boards. Should we change it to trait Board?
> > 
> > 
> > I vote yes
> 
> While we have "boards" in the upstream repository, I imagine Tock may not run on _physical boards_ in production settings, but often as a small part of a larger system (think a platform co-processor). I agree that Platform isn't a great name, but "boards" might be attached to the types of devices we do development on -- not necessarily what Tock would run on in production?

I'm happy to define "board" as the collection of hardware connected to the MCU.

Looking at `trait Platform`, it really has nothing to do with hardware. It's actually more of a configuration for the particular "distribution" of Tock.

```rust
pub trait Platform {
    fn with_driver<F, R>(&self, driver_num: usize, f: F) -> R
    where
        F: FnOnce(Option<&dyn Driver>) -> R;

    fn filter_syscall(
        &self,
        _process: &dyn process::Process,
        _syscall: &syscall::Syscall,
    ) -> Result<(), errorcode::ErrorCode> {
        Ok(())
    }

    fn process_fault_hook(&self, process: &dyn process::Process) -> Result<(), ()> {
        Err(())
    }
}
``` 


`with_driver()` is related to hardware, but there can be software drivers, and a "distribution" is under no obligation to expose all or any particular hardware feature. The other two APIs have little to do with the actual hardware running Tock.

Renaming `trait Platform` to `trait Board` because we implement it in `/boards` makes less sense to me. I'll keep thinking about what name might make sense.

bradjc:
> > trait Driver is a very generic name which has many meanings in embedded code. Should we change it?
> 
> I vote `UserDriver`

What about `SyscallDriver`?

hudson-ayers:
> > > trait Driver is a very generic name which has many meanings in embedded code. Should we change it?
> > 
> > 
> > I vote `UserDriver`
> 
> What about `SyscallDriver`?

works for me!

bradjc:
Let's look at `kernel::procs`:

```rust
pub mod procs {
    pub use crate::process::{
        Error, FaultAction, FunctionCall, FunctionCallSource, Process, State, Task,
    };
    pub use crate::process_policies::{
        PanicFaultPolicy, ProcessFaultPolicy, RestartFaultPolicy, StopFaultPolicy,
        StopWithDebugFaultPolicy, ThresholdRestartFaultPolicy,
        ThresholdRestartThenPanicFaultPolicy,
    };
    pub use crate::process_standard::ProcessStandard;
    pub use crate::process_utilities::{load_processes, ProcessLoadError};
}
```


Things I don't like:
- Many things in the same `kernel::procs::` namespace.
- The name "procs".
- That `FunctionCall` (what is that??) and `FunctionCallSource` are publicly exported.

Things I'm not keen on:
- Having just `procs::Error` _and_ `procs::ProcessLoadError`.

My proposal:
- We just remove `FunctionCallSource`. I think it was added for unsubscribe, but it adds a fair bit of generality where it's not clear it is needed. `FunctionCall`s mainly come from upcalls, but can also come from the kernel when it first calls the process's init function. Do we imagine function calls originating from any other source? Having `FunctionCall` store an `Option<UpcallId>` seems like a much more direct solution, since that is (I think) all we actually need to remove `FunctionCall`s after an unsubscribe.
- Rename `FunctionCall` to `ProcessFunctionCall`. Or something else?
- I thought we could keep `FunctionCall` internal, but we need it to pass to a `UserlandKernelBoundary` implementation. So it stays public.
- We update the process<->upcall interface. Instead of `process.enqueue_task()`, we have the more narrow `process.enqueue_upcall(Upcall, r0, r1, r2)`. That way how Process implementations handle upcalls doesn't leak into the upcall implementation.
- We try namespacing all of the policies in `kernel::procs::policies::`.
- We rename `kernel::procs` to `kernel::processes`.

hudson-ayers:
> * We just remove `FunctionCallSource`. I think it was added for unsubscribe, but it adds a fair bit of generality where it's not clear it is needed. `FunctionCall`s mainly come from upcalls, but can also come from the kernel when it first calls the process's init function. Do we imagine function calls originating from any other source? Having `FunctionCall` store an `Option<UpcallId>` seems like a much more direct solution, since that is (I think) all we actually need to remove `FunctionCall`s after an unsubscribe.

I think the current approach is easier to understand -- just storing an `Option<UpcallId>` relies on code comments to indicate that the ID not being present means the `FunctionCall` came from the kernel. It is unfortunate that all these types have to be public to enable external implementations of process, but that sort of feels like a cost we just have to pay.

> * Rename `FunctionCall` to `ProcessFunctionCall`. Or something else?

This is an improvement. Maybe `UserspaceFunctionCall` makes it a bit clearer that this is a function pointer to userspace code, which is why it is different from other function pointers in the kernel which may or may not be associated with a given process?

> * I thought we could keep `FunctionCall` internal, but we need it to pass to a `UserlandKernelBoundary` implementation. So it stays public.

I think that the need to make `Task` public for external process implementations also means that the variants of `Task` need to be public.

> * We update the process<->upcall interface. Instead of `process.enqueue_task()`, we have the more narrow `process.enqueue_upcall(Upcall, r0, r1, r2)`. That way how Process implementations handle upcalls doesn't leak into the upcall implementation.

We would also need to add a `process.enqueue_ipc_operation()` as well. I think doing that makes for a more sensible interface, but underneath it makes sense for `ProcessStandard::enqueue_upcall()` and `ProcessStandard::enqueue_ipc_operation()` to both call the existing `enqueue_task()` (which we would just make a private method on `ProcessStandard`). 

> * We try namespacing all of the policies in `kernel::procs::policies::`.

+1

> * We rename `kernel::procs` to `kernel::processes`.

I am indifferent on this.

ppannuto:
> > * We just remove `FunctionCallSource`. I think it was added for unsubscribe, but it adds a fair bit of generality where it's not clear it is needed. `FunctionCall`s mainly come from upcalls, but can also come from the kernel when it first calls the process's init function. Do we imagine function calls originating from any other source? Having `FunctionCall` store an `Option<UpcallId>` seems like a much more direct solution, since that is (I think) all we actually need to remove `FunctionCall`s after an unsubscribe.
> 
> I think the current approach is easier to understand -- just storing an `Option<UpcallId>` relies on code comments to indicate that the ID not being present means the `FunctionCall` came from the kernel. It is unfortunate that all these types have to be public to enable external implementations of process, but that sort of feels like a cost we just have to pay.

+1 to this view, explicit source names is more readable than implicit implication of empty `Option`.

> > * Rename `FunctionCall` to `ProcessFunctionCall`. Or something else?
> 
> This is an improvement. Maybe `UserspaceFunctionCall` makes it a bit clearer that this is a function pointer to userspace code, which is why it is different from other function pointers in the kernel which may or may not be associated with a given process?

Perhaps `UserspaceUpcallRef`? I find the `Call` part confusing as it's not something any kernel code can ever call, rather it's a reference to a method that will be injected into userspace execution at some point.

I like the explicit `Userspace` no matter what we do with the rest.

> > * We update the process<->upcall interface. Instead of `process.enqueue_task()`, we have the more narrow `process.enqueue_upcall(Upcall, r0, r1, r2)`. That way how Process implementations handle upcalls doesn't leak into the upcall implementation.
> 
> We would also need to add a `process.enqueue_ipc_operation()` as well. I think doing that makes for a more sensible interface, but underneath it makes sense for `ProcessStandard::enqueue_upcall()` and `ProcessStandard::enqueue_ipc_operation()` to both call the existing `enqueue_task()` (which we would just make a private method on `ProcessStandard`).

+1

> > * We try namespacing all of the policies in `kernel::procs::policies::`.
> 
> +1

+1

> > * We rename `kernel::procs` to `kernel::processes`.
> 
> I am indifferent on this.

I'm also pro `processes`; we spell things out everywhere else, this was always an odd anachronism due to Rust limitations.

bradjc:
> > > * We just remove `FunctionCallSource`. I think it was added for unsubscribe, but it adds a fair bit of generality where it's not clear it is needed. `FunctionCall`s mainly come from upcalls, but can also come from the kernel when it first calls the process's init function. Do we imagine function calls originating from any other source? Having `FunctionCall` store an `Option<UpcallId>` seems like a much more direct solution, since that is (I think) all we actually need to remove `FunctionCall`s after an unsubscribe.
> > 
> > 
> > I think the current approach is easier to understand -- just storing an `Option<UpcallId>` relies on code comments to indicate that the ID not being present means the `FunctionCall` came from the kernel. It is unfortunate that all these types have to be public to enable external implementations of process, but that sort of feels like a cost we just have to pay.
> 
> +1 to this view, explicit source names is more readable than implicit implication of empty `Option`.

Custom types and names come with their own readability overhead, where a reader has to now find the custom type and understand its meaning and usage. Currently that is:

```rust
/// Enumeration to identify whether a function call for a process comes directly
/// from the kernel or from a upcall subscribed through a `Driver`
/// implementation.
///
/// An example of a kernel function is the application entry point.
#[derive(Copy, Clone, Debug)]
pub enum FunctionCallSource {
    /// For functions coming directly from the kernel, such as `init_fn`.
    Kernel,
    /// For functions coming from capsules or any implementation of `Driver`.
    Driver(UpcallId),
}
```

It also adds complexity in its generality, which I would be supportive of if we had even a thought on what other function call sources can exist, or why we would even care what triggered the function call to be enqueued for a process. Tracking the source is only used to cancel pending upcalls in response to a (un)subscribe.

> 
> > > * Rename `FunctionCall` to `ProcessFunctionCall`. Or something else?
> > 
> > 
> > This is an improvement. Maybe `UserspaceFunctionCall` makes it a bit clearer that this is a function pointer to userspace code, which is why it is different from other function pointers in the kernel which may or may not be associated with a given process?
> 
> Perhaps `UserspaceUpcallRef`? I find the `Call` part confusing as it's not something any kernel code can ever call, rather it's a reference to a method that will be injected into userspace execution at some point.

Well, the whole issue with `FunctionCallSource` stems from function calls _not_ always being upcalls.

ppannuto:
Ah, okay, I've read the rest of the surrounding context here and am now caught up on this, sorry. On deeper read this does seem like a type that we can eliminate -- and likely eliminate confusion by renaming `source` in the `FunctionCall` structure to something like `originating_driver: Option<UpcallId>` to help this bit a touch more self-documenting

hudson-ayers:
Fair enough, I see both sides of it. I certainly won't block that change

phil-levis:
>     * We just remove `FunctionCallSource`. I think it was added for unsubscribe, but it adds a fair bit of generality where it's not clear it is needed. `FunctionCall`s mainly come from upcalls, but can also come from the kernel when it first calls the process's init function. Do we imagine function calls originating from any other source? Having `FunctionCall` store an `Option<UpcallId>` seems like a much more direct solution, since that is (I think) all we actually need to remove `FunctionCall`s after an unsubscribe.

Storing an `Option<UpcallId>` makes sense. 

>     * Rename `FunctionCall` to `ProcessFunctionCall`. Or something else?

It's really not a function call (it's not on the same stack!) so I think this is not a good name. `main` is a special case. How about `UpcallInvocation` or `UpcallCall`?
 
>     * We update the process<->upcall interface. Instead of `process.enqueue_task()`, we have the more narrow `process.enqueue_upcall(Upcall, r0, r1, r2)`. That way how Process implementations handle upcalls doesn't leak into the upcall implementation.

This seems to be a separate question, redefining the process interface. Seems suitable for separate PR.
 
>     * We rename `kernel::procs` to `kernel::processes`.

Why not `kernel::process`?

phil

phil-levis:
> I haven't come up with anything great. So far, my best idea is `KernelExtensions`.

It seems like a problematic trait, because it's combining very different pieces of functionality. The reason they are coupled are because a board has to implement them, not because they are semantically linked. It would make more sense to break these into separate traits.  We could use a composite trait if there's an issue if the number of references held. 

```
trait SystemCallDispatch {...}
trait SystemCallFilter {...}
trait ProcessFault {...}
```

ppannuto:
On the one hand, splitting traits to semantic meaning is nice, but on the other hand, it creates a bit of a usability crunch. 

Would we then want to define a supertrait that ensure folks pass everything that's needed to the kernel?
`BoardInstantiation: SystemCallDispatch + SystemCallFilter + ProcessFault`
..or `TockInstance`, `KernelRequirements`, or `ThisIsWhatYourTockReallyIsAtTheEndOfTheDay`?

Feels like we'd still want/need the name :/

phil-levis:
> On the one hand, splitting traits to semantic meaning is nice, but on the other hand, it creates a bit of a usability crunch.
> 
> Would we then want to define a supertrait that ensure folks pass everything that's needed to the kernel?
> `BoardInstantiation: SystemCallDispatch + SystemCallFilter + ProcessFault`
> ..or `TockInstance`, `KernelRequirements`, or `ThisIsWhatYourTockReallyIsAtTheEndOfTheDay`?
> 
> Feels like we'd still want/need the name :/

The issue is less about semantic meaning and more about decomposition. I.e., if the methods are all in a single trait, you can't implement them independently, as composable parts. But if they are separate traits, aggregated into a composite trait requires fewer references and is a simpler configuration call. We can easily have a dispatcher that provides all the traits and just passes through to implementations, or provides default implementations.

bradjc:
Updated version:

```rust
// Kernel crate resources exposed as `kernel::module::Type`.
pub mod appslice;
pub mod capabilities;
pub mod common;
pub mod component;
pub mod debug;
pub mod driver;
pub mod errorcode;
pub mod grant;
pub mod hil;
pub mod introspection;
pub mod ipc;
pub mod schedulers;
pub mod syscall;
pub mod platform;
pub mod process;
pub mod upcall;

mod config;
mod memop;
mod process_policies;
mod process_standard;
mod process_utilities;

// Core resources exposed as `kernel::Type`.
pub use crate::driver::Driver;
pub use crate::errorcode::ErrorCode;
pub use crate::kernel::Kernel;
pub use crate::process::ProcessId;
pub use crate::scheduler::Scheduler;
```


And in process.rs:

```
// Export all process related types via `kernel::process::`.
pub use crate::process_policies::{
    PanicFaultPolicy, ProcessFaultPolicy, RestartFaultPolicy, StopFaultPolicy,
    StopWithDebugFaultPolicy, ThresholdRestartFaultPolicy, ThresholdRestartThenPanicFaultPolicy,
};
pub use crate::process_standard::ProcessStandard;
pub use crate::process_utilities::{load_processes, ProcessLoadError};
```

bradjc:
Right now the last thing main.rs does is call kernel_loop, which has this signature:

```rust
pub fn kernel_loop<P: Platform, C: Chip, SC: Scheduler<C>, const NUM_PROCS: usize>(
        &self,
        platform: &P,
        chip: &C,
        ipc: Option<&ipc::IPC<NUM_PROCS>>,
        scheduler: &SC,
        _capability: &dyn capabilities::MainLoopCapability,
    )
```

There are three main parameters which boards have to provide: `chip`, `platform`, and `scheduler`. Conceptually, we are providing two things: 1) hardware resources including a systick timer and MPU, and 2) configuration/settings including the syscall filter, fault fixup handler, and scheduler.

First, I propose we provide the scheduler through this new `KernelResources` (name pending) trait. That way we keep a consistent way we provide extensions to the kernel. (This `KernelResources` would just be a meta trait and would simply provide references to associated traits.)

The hardware (`Chip`) parameter is a little trickier. First, there is some tension about where things like the systick timer and watchdog should come from (see #1998). Currently the `Chip` implementation provides all of the hardware resources, but a board, could theoretically want to use a different systick timer (if there are multiple options), or use or not use a watchdog.

Current structure:
```rust
pub trait Chip {
  // these seem very chip specific
    fn service_pending_interrupts(&self);
    fn has_pending_interrupts(&self) -> bool;
    fn sleep(&self);
    unsafe fn atomic<F, R>(&self, f: F) -> R
    where
        F: FnOnce() -> R;
    unsafe fn print_state(&self, writer: &mut dyn Write);

  // these seems fairly reasonable for a chip to implement
    fn mpu(&self) -> &Self::MPU;
    fn userspace_kernel_boundary(&self) -> &Self::UserspaceKernelBoundary;

  // these are the potential issues
    fn scheduler_timer(&self) -> &Self::SchedulerTimer;
    fn watchdog(&self) -> &Self::WatchDog;
}
```

Should we rework this somehow? Maybe pull `scheduler_timer()` and `watchdog()` out of `Chip` and add them to `KernelResources` (name)?

hudson-ayers:
> First, I propose we provide the scheduler through this new `KernelResources` (name pending) trait. That way we keep a consistent way we provide extensions to the kernel. (This `KernelResources` would just be a meta trait and would simply provide references to associated traits.)

+ 1

> The hardware (`Chip`) parameter is a little trickier. First, there is some tension about where things like the systick timer and watchdog should come from (see #1998). Currently the `Chip` implementation provides all of the hardware resources, but a board, could theoretically want to use a different systick timer (if there are multiple options), or use or not use a watchdog.
> 
> Current structure:
> 
> ```rust
> pub trait Chip {
>   // these seem very chip specific
>     fn service_pending_interrupts(&self);
>     fn has_pending_interrupts(&self) -> bool;
>     fn sleep(&self);
>     unsafe fn atomic<F, R>(&self, f: F) -> R
>     where
>         F: FnOnce() -> R;
>     unsafe fn print_state(&self, writer: &mut dyn Write);
> 
>   // these seems fairly reasonable for a chip to implement
>     fn mpu(&self) -> &Self::MPU;
>     fn userspace_kernel_boundary(&self) -> &Self::UserspaceKernelBoundary;
> 
>   // these are the potential issues
>     fn scheduler_timer(&self) -> &Self::SchedulerTimer;
>     fn watchdog(&self) -> &Self::WatchDog;
> }
> ```
> 
> Should we rework this somehow? Maybe pull `scheduler_timer()` and `watchdog()` out of `Chip` and add them to `KernelResources` (name)?

I think that making `scheduler_timer` and `watchdog` configurable from `main.rs` is important. I am not sure whether it is better to do that by putting them in `KernelResources` or by trying to make them generics passed to `Chip` (as opposed to associated types, which they are now).

phil-levis:
I agree with all of this, with one exception: `common`.

I get that there's a grab-bag of stuff in there, but `common` doesn't seem to communicate this. It seems to me that there's actually four different types of code in here:
  1. data structures, such as leasable_buffer, list, queue, ring_buffer, and static_ref,
  2. math operations (math.rs)
  3. peripheral management (only actually used in msp432/cs, sam4l/i2c, sam4l/spi and sam4l/eic),
  4. deferred procedure calls

The module also exports cells and registers.

I'd like to suggest that:
  - common should really just be data structures, and renamed accordingly,
  - peripheral management and DPC should be in the kernel crate directly, since they are ways that capsules and chips interact with the kernel, and
  -  cells and registers should be directly exported by the kernel crate.

I'm not quite sure what do with math. I suppose it could either go into the kernel crate, or stay with the data structures. But IIRC it's used mostly in core kernel functionality. I think it would be OK to keep it with the data structures, though, as long as the crate name reflects this.

bradjc:
Proposed update to `common` folder:
```
kernel/
   deferred_call.rs
   dynamic_deferred_call.rs

   utilities/
      math.rs
      static_init.rs
      storage_volume.rs
      leasable_buffer.rs
      static_ref.rs
      peripheral_management.rs

      ::cells
      ::registers

   collections/
      list, queue, ring buffer
```

- Move the generic data structures to `collections` (seems to be the standard name for those in rust and other libraries).
- Move deferred calls to the root, particularly since those have static variables.
- Basically rename "common" to more specific "utilities". This would then contain (and/or export) all code that the kernel crate provides to make writing other parts of tock easier (or possible), but that the core kernel does not use directly. I believe everything I have listed under utilities is not used in the core kernel itself.
  - Split `utils.rs` into better grouped files.

lschuermann:
I'm generally in favor of the `common` folder restructuring. However, I'd like to raise the question whether it could make sense to move `cells` and `registers` to the top level crate exports?

My reasoning being that these modules don't have anything to do with the kernel in particular, and therefore incorporating them into the kernel module organization doesn't seem to make much sense to me. Instead, these are (more or less) just crate re-exports. Keeping these makes sense, as these crates are used all throughout the Tock ecosystem; having the `kernel`-crate reexport that functionality spares us from explicitly importing these in various places and can prevent version conflicts. However, the fact that they are reexported through the `kernel` crate seems to be solely justified by the fact that `kernel` is used by essentially all other crates already, and not by some semantic relationship between those crates.

(The `tock-registers` crate is not simply reexported, but the module-paths are slightly rewritten. I'm working on fixing that as part of #2579, such that in the future this can be a simple reexport.)

bradjc:
> I'm generally in favor of the `common` folder restructuring. However, I'd like to raise the question whether it could make sense to move `cells` and `registers` to the top level crate exports?
> 
> My reasoning being that these modules don't have anything to do with the kernel in particular, and therefore incorporating them into the kernel module organization doesn't seem to make much sense to me. Instead, these are (more or less) just crate re-exports. Keeping these makes sense, as these crates are used all throughout the Tock ecosystem; having the `kernel`-crate reexport that functionality spares us from explicitly importing these in various places and can prevent version conflicts. However, the fact that they are reexported through the `kernel` crate seems to be solely justified by the fact that `kernel` is used by essentially all other crates already, and not by some semantic relationship between those crates.

Historically they are re-exported because they were originally implemented in the kernel crate, and when we moved them to their own crates it didn't seem worth it to change all of the existing chip and capsule files to the new `use` path. So I agree that now is the time to revisit that decision since everything is changing anyway.

The reason I proposed leaving them as exported via `kernel::utilities::` is registers and cells are essential to developing tock, and having the kernel export them allows other crates to use `kernel` and get all of the important tock resources. Also, I couldn't come up with any reasons why other tock crates would benefit from including the dependency directly.

bradjc:
I've put implementing this on hold until we decide on #2642.

phil-levis:
> I've put implementing this on hold until we decide on #2642.

What is #2642?

bradjc:
Replaced by #2659

PR #2562, opened 2021-04-27
capsules/gpio_async: enforce grant/single-process

ppannuto:
### Pull Request Overview

This is one of the capsules blocking #2462.

In contrast to the other PRs in the #2462 series, there's not really so much
meaning to "owning" an Async GPIO peripheral, so this PR effectively
just virtualizes the peripheral rather than enforcing single-process semantics.

Current GPIO HIL semantics notify *all* processes of interrupts to any
pin with interrupts configured (and it's left to higher layers to filter
activity based on which pin generated their interrupt). For Async GPIO,
this is awkward to virtualize (or not), as there is no owning process of
a pin for activity interrupts, but there is for configuration result
interrupts.

Note, however, that the underlying hardware likely can't handle multiple
concurrent configurations from multiple apps.
For now then, this capsule tracks a configuration while it is in flight
and notifies the correct process that their configuration has succeeded.
It does not track pin "owners", and notifies all apps of events on pins
(which is the same behavior as the regular gpio capsule).
In the rare case where two apps attempt concurrent configuration
requests, the later app will receive `EBUSY`. A retry loop should be
sufficient for most apps to handle this case.

### Testing Strategy

Compiling.

### TODO or Help Wanted

Not sure I love the 'notify all apps of all GPIO events' behavior, but
I believe we made that call because assigning owners to each pin
would incur non-trivial memory overhead. Might be worth thinking
through whether the same argument holds for Aysnc GPIO, but this
PR probably isn't the best place for that discussion.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/acd52832/src/main.rs
+++ b/boards/acd52832/src/main.rs
@@ -343,7 +343,10 @@ pub unsafe fn main() {
     // `gpio_async` is the object that manages all of the extenders.
     let gpio_async = static_init!(
         capsules::gpio_async::GPIOAsync<'static, capsules::mcp230xx::MCP230xx<'static>>,
-        capsules::gpio_async::GPIOAsync::new(async_gpio_ports)
+        capsules::gpio_async::GPIOAsync::new(
+            async_gpio_ports,
+            board_kernel.create_grant(&memory_allocation_capability)
+        ),
     );
     // Setup the clients correctly.
     for port in async_gpio_ports.iter() {
--- a/capsules/src/gpio_async.rs
+++ b/capsules/src/gpio_async.rs
@@ -25,10 +25,9 @@
 //! }
 //! ```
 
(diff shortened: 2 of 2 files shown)


lschuermann (approved):
This is great. I think that this is a really good compromise between a peripheral somewhat hard to virtualize, and a driver which fits the Tock ecosystem. There might be potential edge-cases where the fact that one application can influence the callbacks delivered to another app due to the interrupt configuration, but it's already much better than the previous non-virtualized solution.

I think it's fine that an application might get more callbacks than it asked for, given it can ignore the ones not requested, or just unsubscribe entirely.

bradjc (approved):

hudson-ayers (approved):
bors r+

PR #2567, opened 2021-05-03
Led HIL for LedMarix

alexandruradovici:
### Pull Request Overview

This pull request adds the ability to use single LEDs from the LedMatrixDriver with the Led HIL.

This is useful for drivers that want to control LEDs that are within an Led matrix.

### Testing Strategy

This pull request was tested using a micro:bit v2 board.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/led_matrix.rs
+++ b/boards/components/src/led_matrix.rs
@@ -27,6 +27,20 @@
 //!     nrf52::rtc::Rtc<'static>
 //! ));
 //! ```
+//!
+//! Single LED usage
+//! ----------------
+//!
+//! ```rust
+//! let led = components::led_matrix_led!(
+//!     nrf52::gpio::GPIOPin<'static>,
+//!     capsules::virtual_alarm::VirtualMuxAlarm<'static, nrf52::rtc::Rtc<'static>>,
+//!     led,
+//!     1,
+//!     2
+//! );
+//! ```
+//!
 
 use capsules::led_matrix::LedMatrixDriver;
 use capsules::virtual_alarm::{MuxAlarm, VirtualMuxAlarm};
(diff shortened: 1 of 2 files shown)


bradjc (dismissed):

hudson-ayers (dismissed):

lschuermann (approved):
Looks good in general, with two minor comments:

- it might make sense to use the introduced helper methods not only in the `LedMatrixLed` but also in the `kernel::Driver` implementation, to avoid duplicating the logic there.
- I'm entirely convinced we'd need a macro in `components` to construct an `LedMatrixLed` (since it's just a wrapper around a static-init'ed `LedMatrixLed::new`), but I'm also not opposed to it.

alistair23 (dismissed):

hudson-ayers:
> * I'm not entirely convinced we'd need a macro in `components` to construct an `LedMatrixLed` (since it's just a wrapper around a static-init'ed `LedMatrixLed::new`), but I'm also not opposed to it.

When I'm looking for how to add a capsule to a board I will often go straight to the component, so I consider this helpful

alexandruradovici:
> Looks good in general, with two minor comments:
> 
>     * it might make sense to use the introduced helper methods not only in the `LedMatrixLed` but also in the `kernel::Driver` implementation, to avoid duplicating the logic there.
> 
Can you please be more specific, I'm not sure what you are referring to.

>     * I'm not entirely convinced we'd need a macro in `components` to construct an `LedMatrixLed` (since it's just a wrapper around a static-init'ed `LedMatrixLed::new`), but I'm also not opposed to it.

lschuermann:
If I'm not mistaken you're implementing functions such as `read` and `toggle`, which might also be usable as part of the `command` implementation. Here it's not a big deal, it's fine as it is now.

alexandruradovici:
> If I'm not mistaken you're implementing functions such as `read` and `toggle`, which might also be usable as part of the `command` implementation. Here it's not a big deal, it's fine as it is now.

You are right, I modified it.

hudson-ayers (approved):
bors r+

PR #2572, opened 2021-05-04
capsules/l3gd20: migrating non-virtualized driver to Grants

phil-levis:
### Pull Request Overview

As part of #2462, this migrates the l3gd20 nonvirtualized userspace driver to keep process state in a Grant region. It adopts the mechanism to reserve a driver as introduced in #2521.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/l3gd20.rs
+++ b/capsules/src/l3gd20.rs
@@ -103,10 +103,11 @@
 //!
 
 use core::cell::Cell;
+use core::mem;
 use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::hil::sensors;
 use kernel::hil::spi;
-use kernel::{CommandReturn, Driver, ErrorCode, ProcessId, Upcall};
+use kernel::{CommandReturn, Driver, ErrorCode, Grant, ProcessId, Upcall};
 
 use crate::driver;
 pub const DRIVER_NUM: usize = driver::NUM::L3gd20 as usize;
@@ -174,6 +175,18 @@ enum L3gd20Status {
 //     Idle,
 // }
 
+pub struct App {
+    upcall: Upcall,
+}
+
+impl Default for App {
+    fn default() -> App {
+        App {
+            upcall: Upcall::default()
+        }
+    }
+}
+
 pub struct L3gd20Spi<'a> {
     spi: &'a dyn spi::SpiMasterDevice,
     txbuffer: TakeCell<'static, [u8]>,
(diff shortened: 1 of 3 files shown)


capsules/src/l3gd20.rs
```diff
@@ -174,6 +175,18 @@ enum L3gd20Status {
...
+    upcall: Upcall,
+}
+
+impl Default for App {
+    fn default() -> App {
+        App {
+            upcall: Upcall::default(),
+        }
+    }
+}
```
hudson-ayers, line 188: This could be just a `#[derive(Default)]` annotation

capsules/src/l3gd20.rs
```diff
@@ -377,16 +408,19 @@ impl Driver for L3gd20Spi<'_> {
...
     ) -> Result<Upcall, (Upcall, ErrorCode)> {
         match subscribe_num {
             0 /* set the one shot callback */ => {
-                callback = self.callback.replace (callback);
-                Ok (callback)
+                let _ = self
+                    .grants.enter(process_id, |grant| {
+                    mem::swap(&mut grant.upcall, &mut upcall);
+                });
+                Ok(upcall)
```
hudson-ayers, line 420: This assumes entering the grant does not fail, but if entering the grant does fail, this will incorrectly return `Ok()` to the app despite the fact that the callback will never be called.

capsules/src/l3gd20.rs
lschuermann: I think this is fine, in particular since #2462 does not include the proc-macro-based derive `GrantDefault` implementations, so this will make our lives slightly easier. :smile:

capsules/src/l3gd20.rs
phil-levis: Ah, right. I was operating under the assumption that one could only fail to enter the grant if the process has died (it therefore wouldn't matter if you returned Ok()). But it could also be that the grant region is exhausted. 

One note: I copied this code from the `spi_controller` capsule, so I suspect there are others that aren't handling this correctly.

phil-levis: I copied this from `spi_controller`. If the only failure is a process fails, I think this is OK. But since it can also fail if the grant region is depleted, I agree, we want to pass the error to userspace.

lschuermann: Good catch! I suspect I was irritated because of the fact that issuing a command will indeed release the reservation once entering a Grant fails. Will issue fixes accordingly.

hudson-ayers (approved):

lschuermann (approved):

capsules/src/l3gd20.rs
lschuermann: Fixed in #2577.

PR #2582, opened 2021-05-11
document compiler fence before switching to userspace

lschuermann:
### Pull Request Overview

We need to ensure that all writes issued in the kernel are committed to memory prior to switching to userspace, in order to provide a consistent view on application-accessible buffers.

This adds comments outlining why we're guaranteed to not perform memory-reordering across context switches.

### Testing Strategy

None.

### TODO or Help Wanted

A draft PR because
- [ ] I don't know whether the ARM implementation is guaranteed to not have memory-reordering happen across context switches. I'm confused by the call through the C-style `switch_to_user` symbols (it in general seems like it would be more cleanly encapsulated through generics).

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/syscall.rs
+++ b/arch/rv32i/src/syscall.rs
@@ -166,6 +166,16 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
         _app_brk: *const u8,
         state: &mut Riscv32iStoredState,
     ) -> (ContextSwitchReason, Option<*const u8>) {
+        // We need to ensure that the compiler does not reorder
+        // kernel memory writes to after the userspace context switch
+        // to ensure we provide a consistent memory view of
+        // application-accessible buffers.
+        //
+        // The compiler will not be able to reorder memory accesses
+        // beyond this point, as the "nomem" option on the asm!-block
+        // is not set, hence the compiler has to assume the assembly
+        // will issue arbitrary memory accesses (acting as a compiler
+        // fence).
         asm!("
           // Before switching to the app we need to save the kernel registers to
           // the kernel stack. We then save the stack pointer in the mscratch


jrvanwhy:
I think that `asm!` already gives is this guarantee, because we are not passing the `nomem` [option](https://doc.rust-lang.org/unstable-book/library-features/asm.html#options-1). IIRC app memory is already in a global variable in the boards.

lschuermann:
Yes, that makes sense. I think it might not be the fact that app memory is already in a global variable (because consistency in that alone would only make app writes visible to the kernel), but rather that the compiler must assume arbitrary memory access and hence provide a consistent view on memory in the assembly block, issuing all pending writes accordingly.

My question is: do we want to rely on this (more or less) implicit behavior? Or could it help to have this compiler fence added, if only for clarity?

jrvanwhy:
> My question is: do we want to rely on this (more or less) implicit behavior? Or could it help to have this compiler fence added, if only for clarity?

"The compiler can't migrate memory operations across an `asm!` block unless it has the `nomem` option" seems like a reasonable behavior to rely upon implicitly. I think it would be better to have a comment than an compiler fence.

lschuermann:
I agree. Will change to a comment. Thanks for your advice @jrvanwhy!

lschuermann:
Given we've established the RISC-V side of things is safe, I'd still like to explicitly confirm the same holds true for Cortex-M, as switching to userspace works differently there.

With all Cortex-M sub-architectures, the generic `cortexm::syscall::SysCall`'s (being the `UserspaceKernelBoundary`) `switch_to_process` is the entry point for switching to userspace. This function immediately calls into a `switch_to_user` symbol using the C FFI.

https://github.com/tock/tock/blob/90147753c8821361ccb196a395f8e0e2624b6b70/arch/cortex-m/src/syscall.rs#L181-L187

This symbol is defined in Rust in the respective `cortexm{0,0p,3,4,7}` crates representing the precise Cortex-M sub-architecture of the chip. From there on, either directly or through another common function of the `cortexm` crate, an inline assembly block is invoked which does not have the "nomem" option set:

https://github.com/tock/tock/blob/90147753c8821361ccb196a395f8e0e2624b6b70/arch/cortex-m4/src/lib.rs#L30-L35

https://github.com/tock/tock/blob/90147753c8821361ccb196a395f8e0e2624b6b70/arch/cortex-m/src/lib.rs#L298-L302

What worries me is that presumably the compiler should be free to assume that certain memory regions (e.g. behind a mutable slice reference, etc.), which are not passed as a `*const`/`*mut` must not be read/changed by the FFI function [1] (i.e. _not_ inserting a compiler fence when calling an FFI function in general and being able to perform memory-reordering across it). 

Naturally, the compiler cannot reason beyond a FFI call, and as such wouldn't be able to infer anything about the fact that the code executed beyond that call will be able to access the app memory.

I might very well be missing something again, so please correct me if I'm wrong. Still, I think it's valuable to explicitly reason why what we're doing here is actually sound. In any case, based on [a discussion on the Rust forum](https://users.rust-lang.org/t/will-rustc-emit-memory-barriers-for-calling-into-an-ffi-function/57903/7), I'd deduce that at least under some circumstances the compiler will perform memory-reordering (or elide writes) across FFI calls, so we might want to be careful here.

[1]: I know this reasoning is fuzzy, as a pointer doesn't incorporate a length, but passing a pointer of some Rust object down a C FFI might be viewed by the compiler as only providing access to that single allocated object, if following [Rust's pointer validity rules](https://doc.rust-lang.org/std/ptr/index.html#safety).

jrvanwhy:
First: I will assume that process memory is defined as a `static` variable in the kernel. I don't really know how Rust handles access to memory that is not part of a `static` variable; I think that is a not-yet-specified part of Rust's memory model.

Based on my reading of https://github.com/rust-lang/unsafe-code-guidelines/issues/215, I think we just need to pass a pointer derived from the process memory variable to the process' entry point. That should grant the `asm!` blocks read/write access to that memory.

Constraints around `&`/`&mut` references and aliasing are an additional, but distinct, concern. We can't have a `&mut` reference pointing at data when that data is mutated by a userspace process, even if we passed a pointer to that data to that process. As such, we need to destroy all such references before calling `switch_to_process`.

lschuermann:
> Constraints around `&`/`&mut` references and aliasing are an additional, but distinct, concern. We can't have a `&mut` reference pointing at data when that data is mutated by a userspace process, even if we passed a pointer to that data to that process. As such, we need to destroy all such references before calling `switch_to_process`.

Agreed. That's a different story. Here I'm specifically concerned about memory reordering across context switches to userspace. It's also clear that a compiler fence won't possibly allow us to circumvent any of Rust's requirements towards references.

arch/rv32i/src/syscall.rs
```diff
@@ -166,6 +166,16 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
         _app_brk: *const u8,
         state: &mut Riscv32iStoredState,
     ) -> (ContextSwitchReason, Option<*const u8>) {
+        // We need to ensure that all writes issued in the kernel are
+        // committed to memory prior to switching to userspace, in
+        // order to provide a consistent view on
+        // application-accessible buffers.
```
bradjc, line 172:
```suggestion
        // We need to ensure that the compiler does not reorder
        // kernel memory writes to after the userspace context switch
        // to ensure we provide a consistent memory view of
        // application-accessible buffers.
```

I think this clarifies that this is about compiler instruction ordering and not pipelining in the microarchitecture.

hudson-ayers:
> Given we've established the RISC-V side of things is safe, I'd still like to explicitly confirm the same holds true for Cortex-M, as switching to userspace works differently there.
> 
> With all Cortex-M sub-architectures, the generic `cortexm::syscall::SysCall`'s (being the `UserspaceKernelBoundary`) `switch_to_process` is the entry point for switching to userspace. This function immediately calls into a `switch_to_user` symbol using the C FFI.
> 
> https://github.com/tock/tock/blob/90147753c8821361ccb196a395f8e0e2624b6b70/arch/cortex-m/src/syscall.rs#L181-L187
> 
> This symbol is defined in Rust in the respective `cortexm{0,0p,3,4,7}` crates representing the precise Cortex-M sub-architecture of the chip. From there on, either directly or through another common function of the `cortexm` crate, an inline assembly block is invoked which does not have the "nomem" option set:
> 
> https://github.com/tock/tock/blob/90147753c8821361ccb196a395f8e0e2624b6b70/arch/cortex-m4/src/lib.rs#L30-L35
> 
> https://github.com/tock/tock/blob/90147753c8821361ccb196a395f8e0e2624b6b70/arch/cortex-m/src/lib.rs#L298-L302
> 
> What worries me is that presumably the compiler should be free to assume that certain memory regions (e.g. behind a mutable slice reference, etc.), which are not passed as a `*const`/`*mut` must not be read/changed by the FFI function [1] (i.e. _not_ inserting a compiler fence when calling an FFI function in general and being able to perform memory-reordering across it).
> 
> Naturally, the compiler cannot reason beyond a FFI call, and as such wouldn't be able to infer anything about the fact that the code executed beyond that call will be able to access the app memory.
> 
> I might very well be missing something again, so please correct me if I'm wrong. Still, I think it's valuable to explicitly reason why what we're doing here is actually sound. In any case, based on [a discussion on the Rust forum](https://users.rust-lang.org/t/will-rustc-emit-memory-barriers-for-calling-into-an-ffi-function/57903/7), I'd deduce that at least under some circumstances the compiler will perform memory-reordering (or elide writes) across FFI calls, so we might want to be careful here.
> 
> [1]: I know this reasoning is fuzzy, as a pointer doesn't incorporate a length, but passing a pointer of some Rust object down a C FFI might be viewed by the compiler as only providing access to that single allocated object, if following [Rust's pointer validity rules](https://doc.rust-lang.org/std/ptr/index.html#safety).

I think I understand this concern now. The only way this can be problematic is if Rust were to re-order a write to an `AppSlice` across invocations of the scheduler loop. However, this would require Rust to re-order a write to memory past the point where the object giving access to that memory is dropped (as AppSlices are just stored as a pointer/length combo outside of where they are used). Is that actually possible?

jrvanwhy:
If a `&mut [u8]` exists that points into process memory when `switch_to_user` is called, and the process mutates that memory, we get undefined behavior in the kernel. This is because `&mut [u8]` implies exclusive access: the compiler assumes nothing else can access the value. UB also occurs if a `&[u8]` exists in the kernel to memory mutated by a process because `&[u8]` implies no usable `mut` bindings exist to the same memory. UB does not occur if the reference is a `&[UnsafeCell<u8>]`, because another "immutable" binding could exist that is allowed to mutate the value.

So if `AppSlice`-like types contain `&mut [u8]` or `&[u8]` references, they will need to be dropped before calling `switch_to_user`.

Raw pointers don't have that same issue, as they do not imply any sort of exclusive access.

lschuermann:
@jrvanwhy (from an older response)
> First: I will assume that process memory is defined as a `static` variable in the kernel. I don't really know how Rust handles access to memory that is not part of a `static` variable; I think that is a not-yet-specified part of Rust's memory model.
> 
> Based on my reading of [rust-lang/unsafe-code-guidelines#215](https://github.com/rust-lang/unsafe-code-guidelines/issues/215), I think we just need to pass a pointer derived from the process memory variable to the process' entry point. That should grant the `asm!` blocks read/write access to that memory.

I don't think process memory is defined as a `static` variable in the kernel. It seems that on upstream boards, we simply define the `app_memory` region in our linker scripts to be located in RAM and provide the `_sappmem` and `_eappmem` symbols respectively. `load_processes` takes a temporary slice reference of that memory region. Hence, we are technically passing a pointer to that memory region down to the process (through `initialize_process` in cortex-m, which sets `state.psp`, which will as part of `switch_to_process` get passed to the `switch_to_user` FFI function). However, given it doesn't reference a static variable in the kernel (stems from a temporary unsafe-constructed slice), I'm really not sure whether passing down this pointer to `app_memory` will make the compiler guarantee that all writes to that memory region will not be reordered across that FFI call.

> So if AppSlice-like types contain &mut [u8] or &[u8] references, they will need to be dropped before calling switch_to_user.

They do not. You're correct; if they would, we'd have the issue of the process mutating that memory, as well as the kernel itself potentially mutating shared memory through the other pointer writes as part of the `UserspaceKernelBoundary`.

jrvanwhy:
> @jrvanwhy (from an older response)
> 
> > First: I will assume that process memory is defined as a `static` variable in the kernel. I don't really know how Rust handles access to memory that is not part of a `static` variable; I think that is a not-yet-specified part of Rust's memory model.
> > Based on my reading of [rust-lang/unsafe-code-guidelines#215](https://github.com/rust-lang/unsafe-code-guidelines/issues/215), I think we just need to pass a pointer derived from the process memory variable to the process' entry point. That should grant the `asm!` blocks read/write access to that memory.
> 
> I don't think process memory is defined as a `static` variable in the kernel. It seems that on upstream boards, we simply define the `app_memory` region in our linker scripts to be located in RAM and provide the `_sappmem` and `_eappmem` symbols respectively. `load_processes` takes a temporary slice reference of that memory region. Hence, we are technically passing a pointer to that memory region down to the process (through `initialize_process` in cortex-m, which sets `state.psp`, which will as part of `switch_to_process` get passed to the `switch_to_user` FFI function). However, given it doesn't reference a static variable in the kernel (stems from a temporary unsafe-constructed slice), I'm really not sure whether passing down this pointer to `app_memory` will make the compiler guarantee that all writes to that memory region will not be reordered across that FFI call.

As I understand it, producing the memory reference that way puts you in "that part of the language hasn't been figured out" territory. I think the unsafe code guidelines WG wants it to be defined behavior, so I'm okay with relying on it for now.

lschuermann:
> I think the unsafe code guidelines WG wants it to be defined behavior, so I'm okay with relying on it for now.

Fair. Do you think it might be worth opening a tracking issue for this?

Given we're only documenting the RISC-V side of things here, I think this PR is ready.

bradjc (approved):

hudson-ayers (approved):

jrvanwhy:
> > I think the unsafe code guidelines WG wants it to be defined behavior, so I'm okay with relying on it for now.
> 
> Fair. Do you think it might be worth opening a tracking issue for this?

I opened https://github.com/rust-lang/unsafe-code-guidelines/issues/285, it seems good to have a tracking issue here to watch it.

jrvanwhy (approved):

PR #2585, opened 2021-05-17
boards/Makefile: remove relative libtock-c path assumptions

lschuermann:
### Pull Request Overview

Removes any assumptions about application TBFs or `libtock-c` trees using relative paths outside of the Tock repository. In the cases where Tockloader is not supported (yet), the TBF file to include should be passed using the `APP` variable instead, as done in `clue_nrf52840`.

### Testing Strategy

This pull request was tested by running make targets.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/boards/imxrt1050-evkb/Makefile
+++ b/boards/imxrt1050-evkb/Makefile
@@ -1,16 +1,7 @@
 # Makefile for building the tock kernel for the i.MX RT1052 platform
-#
-APP=../../../libtock-c/examples/c_hello/build/cortex-m7/cortex-m7.tbf
-# APP=../../../libtock-rs/target/thumbv7em-none-eabi/tab/imxrt1050/hello_world/cortex-m7.tbf
+
 TARGET=thumbv7em-none-eabi
 PLATFORM=imxrt1050-evkb
-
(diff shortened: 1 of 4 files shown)


hudson-ayers:
Looks good, look forward to the remaining change

lschuermann:
To give an update why this has been pending for so long: I'm not really confident to make these changes to the `imxrt1050-evkb` board without testing it first (for instance, will the tools still work if there's an `elf` extension instead of `axf`, etc.?). Unfortunately NXP's user experience for downloading their MCUXpresso software is entirely broken, and after what have been at least 15 attempts I've managed to start a download with 150kbit/s of a 800MB file. I'll try to have it done by tomorrow.

lschuermann:
I couldn't test the actual flashing process on the `imxrt1050-evkb` unfortunately as I have a slightly different chip than the one supported here, but I did manage to get NXP's flashing tool to complain about the chip and not the input file, so I'm positive these changes should work. Updated the documentation accordingly.

hudson-ayers (approved):

Issue #1995, opened 2020-07-01
Tracking: Support reduced overhead of debugging in `ProcessStandard`


bradjc:
## Background

To help with debugging processes, the `Process` struct currently includes a `Debug` struct that holds different values to help with debugging a process if it crashes. We intentionally sectioned off the `Debug` items with the thought that eventually we wouldn't want to make including debug state mandatory for a userspace process.

## What's New

There is interest in changing how process debugging works on two fronts:
1. Removing it to reduce overhead (see #1830).
2. Increasing it to make debugging even easier (see #1698). Also, I would like to add a "logging" feature that records when certain things fail (for example an sbrk call returns an error).

## Compile-time configuration

#1830

## Dynamic Proposal

At a high-level, I'm imagining a `ProcessDebugger` trait, with different implementations that could (optionally) be passed to `Process::create()` and stored as a reference in the `Process` struct.

```rust
struct Process {
...
    debugger: Option<&ProcessDebugger>
...
}
```

The trait would specify all of the hooks that `Process` calls if the debugger reference is present.

```rust
trait ProcessDebugger {
    fn syscall_called(&self, syscall);
...
}
```

The ProcessDebugger implementation could then decide what to store, and how to display it on a panic.

I don't know exactly how the `Process` trait would change.

I also don't know if this is actually possible. But I'm looking for something that gives us the flexibility to permit the maximum amount of debugging possible, while also allowing us to use less or none at all if desired.

hudson-ayers:
I think that this approach sounds good if we can use generics rather than trait objects to store the `ProcessDebugger` in the `Process`. Otherwise the compiler will not be able to optimize away the function calls in the case of the minimal `ProcessDebugger` being used so this will add a baseline overhead that the proposal in #1830 will not.

However I am concerned that `ProcessType` already being stored behind a trait object may make this difficult

bradjc:
Not much progress here, this might get revisited from a code size perspective.

PR #1735, opened 2020-04-05
[RFC] Driver registry / driver lookup mechanism

lschuermann:
### Pull Request Overview
This PR adds a _driver registry_, which assigns each driver a possibly non-deterministic driver id to use in syscalls. In addition to that, it provides a lookup mechanism to determine the driver id for a given instance of a particular driver. For looking up the driver itself, the existing driver ids are used.

This is very much work in progress and a simple proof of concept. The code is primarily a demonstration for my solution approach to the problem outlined in _Motivation_.

Looking up the driver id (syscall argument) once may have a high cost, but I tried to keep the overhead during runtime as low as possible. This is using array indices to directly map from a driver id to a driver. I have not looked at the generated assembly yet.

As this approach does not make sense for drivers without the need for multiple instances, those can still be addressed the usual way. The pseudorandom ids can be generated after an offset specified in the constructor, dividing the driver ids into two regions: Static, and generated during runtime. Per driver, one mechanism must be decided.

As such an approach incurs significant changes to driver addressing from userspace, it'd be something to work towards for Tock 2.0 #1607.

Feedback is of course welcome, in the likely case that we don't want to have this this PR can be closed. :)

### Motivation
_I'd like to apologize for the confusing terminology, this is something I still have to work on. Specifically, this introduces a separation of what's traditionally known as the driver id, and the id used in syscalls._

Not all drivers offer the possibility to manage multiple devices. While that does make sense for something like LEDs, this is impractical for more complex drivers as the overhead for implementing multi-device support would quickly outrun the benefits. In such cases, when a person new to Tock asks "and how can I have multiple of those?" the standard answer has been to add another driver instance to the board.

That's fine, but suddenly the new driver is reachable under a different syscall number as [originally intended / assigned](https://github.com/tock/tock/blob/master/doc/syscalls/README.md). As this may be hardcoded in userspace, it requires substantial changes and the userspace library will be incompatible.

A possible solution for this issue is to reserve for each driver a few driver ids, taking the offset to the base driver id as an argument in the userspace library. That only postpones the issue until someone reaches the `n` instances limit, and feels more like a hack than a solution.

This pull request takes another approach. The driver numbers are assigned once at boot in a possibly non-deterministic way through a `DriverRegistry` component. That component itself provides a driver to userspace, exposing functionality to lookup the id to use for the syscall. Multiple instances of a single driver can be registered with the `DriverRegistry`, providing an individual _instance identifer_.

[...]

PR #1375, opened 2019-09-06
arch/cortex-m0: Add `.align 4` for `svc_handler()`

rajivr:
Alignment issue came up while trying to set a breakpoint and stepping through `svc_handler()`.

PR #2587, opened 2021-05-19
New ProcessConsole features

maltyam:
### Pull Request Overview

This pull request builds out the process console.  It contains an implementation of the console writer which
prints out console outputs without dropping packets.  The writer also supports a state machine to allow for 
larger prints to be managed with a smaller queue.  These changes were necessary in the implementation of two 
more commands to the process console.  The process command prints out the process memory map of the 
specified running process and the kernel command prints out the kernel version, a list of the drivers, and the 
kernel memory map.  

### Testing Strategy

This pull request was tested by Andrew Malty on the IMIX board.

### TODO or Help Wanted

This pull request still needs to be tested on other boards.

### Documentation Updated

No updates required

### Formatting

Ran `make prepush`.


bradjc:
Can you rebase this onto the master branch? There are a lot of extra changes from other PRs and the diff is hard to read.

alevy:
@maltyam This seems generally very good. Could you please extend the PR description with an explanation of what architectural changes you made and maybe some guidance on how to read the code? That would help tremendously with reviewing _in particular_ since this touches two files in the kernel crate.

kernel/src/process.rs
```diff
@@ -731,3 +737,1611 @@ pub struct FunctionCall {
...
+    /// long.
+    dropped_callback_count: usize,
+
+    /// How many times this process has been paused because it exceeded its
+    /// timeslice.
+    timeslice_expiration_count: usize,
+}
+
+/// A type for userspace processes in Tock.
+pub struct Process<'a, C: 'static + Chip> {
```
bradjc, line 784: These changes need to be integrated to ProcessStandard at some point.

kernel/src/introspection.rs
```diff
@@ -23,6 +23,23 @@ pub struct KernelInfo {
...
+    static _estack: u32;
+    static _sstack: u32;
+    static _stext: u32;
+    static _etext: u32;
+    static _srodata: u32;
+    static _erodata: u32;
+    static _srelocate: u32;
+    static _erelocate: u32;
+    static _szero: u32;
+    static _ezero: u32;
```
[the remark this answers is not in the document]
phil-levis: I agree with this point conceptually, but these seem like things we do actually depend on. These are the basic memory regions of executables. Unless there's no relocation (boot from RAM)? It seems strange for the kernel to not specify its memory layout.

phil-levis:
> (4) I'm hesitant about, because it adds a macro to main.rs.

I agree -- Andrew and I went back and forth a lot on this. We tried coming up with ways that wouldn't, and couldn't. But if there is one, that would be a great improvement.

lschuermann:
Two thoughts which popped up during the call:
- the added macro is at least somewhat similar to #2391, so we can (technically) extend that to reduce required mechanical code for implementing the Platform trait. I know it's gotten its share of pushback, but I think it might be worth mentioning, as this here is similar.
- as part of #1375, I've proposed introducing an optional supertrait of `Driver` [called `DriverInfo`](https://github.com/tock/tock/pull/1735/files#diff-25c0ac2fde905b8f9dca148b082434210950d079ee05367682e3c297b99b80ebR18-R27). If more info about drivers than just the name or type should be printed (one could think about a driver API version or instance identifier), that might be a way to go.

Thanks for the presentation @maltyam!

hudson-ayers:
> > ```
> > 3. process memory map can be printed by process console, and is event driven.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (3) I need to think about a little more, because I'm a little concerned about the portability of the approach and the maintainability.
> 
> One option I can think of is to implement the event driven process memory print in ProcessStandard rather than process console. I could imagine an API in `trait Process`:
> 
> ```rust
> // How many times need to call `print_memory_map()` to print full memory map.
> fn get_number_memory_map_iterations() -> usize;
> 
> // Print a section of the memory map.
> fn print_memory_map(writer, iteration: usize);
> ```
> 
> process_console could call print_memory_map in an event driven manner, and panic could just call it in a loop. This would remove the duplication.
> 
> It wouldn't be the most pretty code in process_standard.rs, but it's already pretty messy. But I think the improved debugability would be worth it.

Given the observation in #2594 that uncalled functions on the process_standard will be included in every binary, I don't think this is a good idea.

I also think that it is a little strange to make this part of the `Process` trait, regardless.

bradjc:
> Given the observation in #2594 that uncalled functions on the process_standard will be included in every binary, I don't think this is a good idea.
> 
> I also think that it is a little strange to make this part of the `Process` trait, regardless.

Why? We already have:

```rust
    /// Print out the memory map (Grant region, heap, stack, program
    /// memory, BSS, and data sections) of this process.
    fn print_memory_map(&self, writer: &mut dyn Write);

    /// Print out the full state of the process: its memory map, its
    /// context, and the state of the memory protection unit (MPU).
    fn print_full_process(&self, writer: &mut dyn Write);
```

So it's just updating the API, not adding anything new. But rather than having a blocking interface, it would be more async compatible.

hudson-ayers:
> Why? We already have:
> 
> ```rust
>     /// Print out the memory map (Grant region, heap, stack, program
>     /// memory, BSS, and data sections) of this process.
>     fn print_memory_map(&self, writer: &mut dyn Write);
> 
>     /// Print out the full state of the process: its memory map, its
>     /// context, and the state of the memory protection unit (MPU).
>     fn print_full_process(&self, writer: &mut dyn Write);
> ```
> 
> So it's just updating the API, not adding anything new. But rather than having a blocking interface, it would be more async compatible.

It sounds like this would be in addition to those, rather than replacing them? At least that was how I interpreted your proposal. And we need the sync version for panic.

In general, though, it sort of feels like all of those functions should be in a separate `ProcessDebug` trait or similar -- though I realize that is somewhat besides the point here.

bradjc:
> It sounds like this would be in addition to those, rather than replacing them? At least that was how I interpreted your proposal. And we need the sync version for panic.

No the intent is that it would replace them, and a synchronous user simply iterates in a for loop.

> 
> In general, though, it sort of feels like all of those functions should be in a separate `ProcessDebug` trait or similar -- though I realize that is somewhat besides the point here.

Yeah: #1995

phil-levis:
> > ```
> > 3. process memory map can be printed by process console, and is event driven.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (3) I need to think about a little more, because I'm a little concerned about the portability of the approach and the maintainability.
> 
> One option I can think of is to implement the event driven process memory print in ProcessStandard rather than process console. I could imagine an API in `trait Process`:
> 
> ```rust
> // How many times need to call `print_memory_map()` to print full memory map.
> fn get_number_memory_map_iterations() -> usize;
> 
> // Print a section of the memory map.
> fn print_memory_map(writer, iteration: usize);
> ```
> 
> process_console could call print_memory_map in an event driven manner, and panic could just call it in a loop. This would remove the duplication.
> 
> It wouldn't be the most pretty code in process_standard.rs, but it's already pretty messy. But I think the improved debugability would be worth it.

It seems weird that printing out process state is defined in a process-specific way. It would kind of be like if `top` somehow asked the kernel how to print out process state. It also mean that if you had multiple kinds of processes, then a great deal of this code would be replicated. It also seems a bit strange to extend the Process trait with some functionality that's intended to support one specific capsule. If you want to change how the data is printed out, you need to either change the kernel code or add a new Process type. I.e., you can't write a new version of process console that prints out the data in a different way.

bradjc:
> I'm pretty strongly against this idea: I think it's overly prioritizing implementation cleanliness over abstraction cleanliness.
> 
> It seems weird that printing out process state is defined in a process-specific way. It would kind of be like if `top` somehow asked the kernel how to print out process state. It also means that if you had multiple kinds of processes, then a great deal of this code would be replicated. It also seems a bit strange to extend the Process trait with some functionality that's intended to support one specific capsule. If you want to change how the data is printed out, you need to either change the kernel code or add a new Process type. I.e., you can't write a new version of process console that prints out the data in a different way.

If the system was severely code-size constrained, I think top (and any other tool) would ask the kernel to print debug information about processes, because the alternative is duplicating the code multiple times.

But in any case, we already have essentially this API today, except `get_number_memory_map_iterations()` is hardcoded to 1, and `print_memory_map()` is always called with `0`.

phil-levis:
> If the system was severely code-size constrained, I think top (and any other tool) would ask the kernel to print debug information about processes, because the alternative is duplicating the code multiple times.

Right -- and the problem is they have to use the way the kernel does it, as opposed to the way that they want to do it. If you want to change it, you have to change the kernel? The only reason process console currently prints things out similarly to the kernel is we weren't sure what it should look like (not having substantive use cases) so I suggested just mostly copying the kernel output.

> But in any case, we already have essentially this API today, except `get_number_memory_map_iterations()` is hardcoded to 1, and `print_memory_map()` is always called with `0`.

How do you divide the number of iterations? What happens if the debug buffer isn't big enough for an iteration? 

That being said, I think the idea of pulling this out into a `ProcessStatePrint` or `ProcessDebug` trait makes sense. The key thing, though is that it should take a Process as a parameter, rather than be necessarily implemented by a `Process` type. That way, you can have an implementation that operates on any `Process` type, rather than coupling it to a specific implementation. 

The fact that this is implemented in `Process` seems mostly a historical artifact and not a design decision?

bradjc:
> How do you divide the number of iterations?

This PR demonstrates one possible division.

> What happens if the debug buffer isn't big enough for an iteration?

It truncates, just as debug!() does now.

phil-levis:
We discussed this on the call today. 
  1. There was agreement to keep ProcessConsole having its own printing.
  2. There was agreement that the capsule listing macro should be at least optional, and that if needs to be removed altogether that would be fine: this is up to Brad on whether it's a hard line.
  3. The concern about kernel symbols was unresolved and we said we'd resolve it in this discussion.

boards/imix/src/main.rs
```diff
@@ -43,6 +43,7 @@ use components::isl29035::AmbientLightComponent;
 use components::led::LedsComponent;
 use components::nrf51822::Nrf51822Component;
 use components::process_console::ProcessConsoleComponent;
+
```
phil-levis, line 46: Remove

boards/imix/src/main.rs
```diff
@@ -102,6 +103,7 @@ static mut CHIP: Option<&'static sam4l::chip::Sam4l<Sam4lDefaultPeripherals>> =
 #[link_section = ".stack_buffer"]
 pub static mut STACK_MEMORY: [u8; 0x2000] = [0; 0x2000];
 
+/// Supported drivers by the platform
```
phil-levis, line 106: Remove

phil-levis (approved):

boards/imix/src/main.rs
maltyam: I can't remove this because every struct requires a comment before it to define the struct.

kernel/src/introspection.rs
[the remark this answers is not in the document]
bradjc: It's more that these are symbols from a linker script which may not be the correct names for something like host-side testing.

bradjc:
>There was agreement that the capsule listing macro should be at least optional, and that if needs to be removed altogether that would be fine: this is up to Brad on whether it's a hard line.

I am always very wary of macros. When they behave like a function I think they work well. `static_init!()` for example essentially takes specific arguments, and more or less "returns" something which is assigned to a name. `storage_volume!()` also takes specific "arguments". `debug_gpio!()` is very much like a function.

When macros get more complicated and start altering the code blocks that are actually compiled, reading code becomes harder. It also becomes much more difficult to track down issues when things go wrong, and IDE tools don't work well with macros.

main.rs is arguably the most visible file of Tock to new users, and I'm more hesitant to add complexity to this file than I might be for other parts of the kernel.

But, macros represent a tradeoff, as they enable functionality that is not always possible in a language. I prefer to give up on functionality in exchange for long-term code readability and new user friendliness. However, I'm just one developer.

bradjc:
>     1. There was agreement to keep ProcessConsole having its own printing.

Does this mean replacing `debug!()` or having the memory map strings in process console as well as process_standard.rs, or both?

In light of https://github.com/tock/tock/issues/2594 and other efforts to reduce code size, duplicating a rather large block of code seems undesirable.

boards/imix/src/main.rs
hudson-ayers: There wasn't a comment before this PR, but the formatter must have passed for that code to reach master

phil-levis:
> 
> 
> > ```
> > 1. There was agreement to keep ProcessConsole having its own printing.
> > ```
> 
> Does this mean replacing `debug!()` or having the memory map strings in process console as well as process_standard.rs, or both?

Both. Replacing `debug!()` because of its ordering issues with echo output (there should be a single output stream), and so that it doesn't overrun the debug buffer.  Keeping separate strings because process console will very possibly want different outputs in the future.
> 
> In light of #2594 and other efforts to reduce code size, duplicating a rather large block of code seems undesirable.
> 
> https://github.com/tock/tock/pull/2587/checks?check_run_id=2738210175#step:6:308:
> 
> ```
> imix: flash use +6944 bytes (+3.95%), RAM use -4 bytes(-0.01%)
> ```

@alevy can correct me if I'm wrong, but his comment was that the systems concerned with code size probably aren't running process console. Also, process console is likely not going to want the same output as a panic dump: it's just there as an example. 

If we later refactor the code to reduce redundancy that would be fine. But @maltyam literally graduates tomorrow, so I don't think we can expect him to substantively refactor his code between then and now.

phil-levis:
> When macros get more complicated and start altering the code blocks that are actually compiled, reading code becomes harder. It also becomes much more difficult to track down issues when things go wrong, and IDE tools don't work well with macros.

This all makes a lot of sense to me. It's also the case that one macro is not a big deal, but lots of macros can be. So deeply important macros like `static_init` are worth it, but adding more for specific services is a bit less compelling. @maltyam removed it, just because this does seem like not that compelling. It's super-useful for process console, but increasing the complexity of main.rs for process console seems iffy.

boards/imix/src/main.rs
phil-levis: Maybe the formatted updated? That happens...

phil-levis: It's fine to keep it.

capsules/src/process_console.rs
```diff
@@ -183,12 +256,458 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
...
+        }
+    }
+    //defines the behavior at each state
+    fn create_state_buffer(&self, state: WriterState, process_id: Option<ProcessId>) {
+        match state {
+            WriterState::KernelBss => {
+                let mut console_writer = ConsoleWriter::new();
+                let kernel_info = KernelInfo::new(self.kernel);
+                let init_bottom: usize = kernel_info.get_kernel_init_end() as usize;
+                //let bss_start: usize = kernel_info.get_kernel_bss_start() as usize;
```
bradjc, line 306: I just want to check if I'm missing something, why isn't bss_start used here?

maltyam: init_bottom is the same as bss_start so you only need one of them.  I only created both functions for the sake of completeness.

bradjc:
Ok, summary of my changes:

1. Address symbols are in the component rather than introspection. This move has another benefit, because now we don't have to comment on why the `unsafe` needed to access those symbols is ok, and we don't have to think about what would happen if they were wrong for some reason.
2. I had to refactor process_console.rs a little to get rustfmt to work. Turns out rustfmt doesn't handle long lines inside of `match` very well.
3. I updated the new functions to the `Process` trait and commented them.

phil-levis (approved):

PR #2590, opened 2021-05-21
RFC: update trd104

hudson-ayers:
### Pull Request Overview

This pull request modifies the text of trd104 to reflect that the kernel cannot prevent applications from accessing buffers that have been "allowed". It also modifies the text to no longer forbid applications from reading from allowed buffers -- given that the kernel cannot stop apps from doing so, I do not really see the benefit of this restriction. It gains us nothing in terms of soundness in the kernel, and can't be relied on as a security policy to prevent apps from accessing kernel data. 

### Testing Strategy

N/A

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/reference/trd104-syscalls.md
+++ b/doc/reference/trd104-syscalls.md
@@ -510,13 +510,14 @@ kernel returns `Failure` with an error code of `NODEVICE`.
 
 The Read-Write Allow system call class is how a userspace process
 shares buffer with the kernel that the kernel can read and write. When
-userspace shares a buffer, it can no longer access it. Calling a
+userspace shares a buffer, it should no longer modify it, though the kernel
+will not stop the app from doing so. Calling a
 Read-Write Allow system call returns a buffer (address and
 length).  On the first call to a Read-Write Allow system call, the
-kernel returns a zero-length buffer. Subsequent successful calls to 
-Read-Write Allow return the previous buffer passed. Therefore, to 
-regain access to a passed buffer, the process must call the same 
-Read-Write Allow system call again. It can do so with a zero-length 
+kernel returns a zero-length buffer. Subsequent successful calls to
+Read-Write Allow return the previous buffer passed. Therefore, to
+regain the ability to write to a passed buffer, the process must call the same
+Read-Write Allow system call again. It can do so with a zero-length
 buffer if it wishes to pass no memory to the kernel.
 
 The register arguments for Read-Write Allow system calls are as
@@ -545,28 +546,26 @@ support multiple allowed buffers.
 The Tock kernel MUST check that the passed buffer is contained within
 the calling process's writeable address space. Every byte of the
 passed buffer must be readable and writeable by the
-process. Zero-length buffers may therefore have abitrary addresses. If
+process. Zero-length buffers may therefore have arbitrary addresses. If
 the passed buffer is not complete within the calling process's
 writeable address space, the kernel MUST return a failure result with
 an error code of `INVALID`.
 
-Because a process relinquishes access to a buffer when it makes a
-Read-Write Allow call with it, the buffer passed on the subsequent
-Read-Write Allow call cannot overlap with the first passed buffer.
-This is because the application does not have access to that
-memory. If an application needs to extend a buffer, it must first call
+When a process passes a buffer to
+Read-Write Allow call with, buffers passed on subsequent
+Read-Write Allow calls cannot overlap with the first passed buffer.
+If an application needs to extend a buffer, it must first call
 Read-Write Allow to reclaim the buffer, then call Read-Write Allow
 again to re-allow it with a different size. If userspace passes
 an overlapping buffer, the kernel MUST return a failure result with
 an error code of `INVALID`.
- 
+
 4.5 Read-Only Allow (Class ID: 4)
 ---------------------------------
 
 The Read-Only Allow class is very similar to the Read-Write Allow class.
-It differs in tow ways: the buffer it passes to the kernel is read-only,
-and the process retains read access to the buffer. The kernel cannot
-write to the buffer. The semantics and calling conventions of
+It differs in one way: the buffer it passes to the kernel is read-only,
+the kernel cannot write to the buffer. The semantics and calling conventions of
 Read-Only Allow are otherwise identical to Read-Write Allow.
 
 The Read-Only Allow class exists so that userspace can pass references
--- a/doc/reference/trd104-syscalls.md
+++ b/doc/reference/trd104-syscalls.md
@@ -510,8 +510,7 @@ kernel returns `Failure` with an error code of `NODEVICE`.
 
 The Read-Write Allow system call class is how a userspace process
 shares buffer with the kernel that the kernel can read and write. When
-userspace shares a buffer, it should no longer modify it, though the kernel
-will not stop the app from doing so. Calling a
+userspace shares a buffer, it must no longer modify it. Calling a
 Read-Write Allow system call returns a buffer (address and
 length).  On the first call to a Read-Write Allow system call, the
 kernel returns a zero-length buffer. Subsequent successful calls to
@@ -552,7 +551,7 @@ writeable address space, the kernel MUST return a failure result with
 an error code of `INVALID`.
 
 When a process passes a buffer to
-Read-Write Allow call with, buffers passed on subsequent
+a Read-Write Allow call, buffers passed on subsequent
 Read-Write Allow calls cannot overlap with the first passed buffer.
 If an application needs to extend a buffer, it must first call
 Read-Write Allow to reclaim the buffer, then call Read-Write Allow
@@ -565,7 +564,7 @@ an error code of `INVALID`.
 
 The Read-Only Allow class is very similar to the Read-Write Allow class.
 It differs in one way: the buffer it passes to the kernel is read-only,
-the kernel cannot write to the buffer. The semantics and calling conventions of
+so the kernel cannot write to the buffer. The semantics and calling conventions of
 Read-Only Allow are otherwise identical to Read-Write Allow.
 
 The Read-Only Allow class exists so that userspace can pass references


doc/reference/trd104-syscalls.md
```diff
@@ -545,28 +546,26 @@ support multiple allowed buffers.
...
 writeable address space, the kernel MUST return a failure result with
 an error code of `INVALID`.
 
-Because a process relinquishes access to a buffer when it makes a
-Read-Write Allow call with it, the buffer passed on the subsequent
-Read-Write Allow call cannot overlap with the first passed buffer.
-This is because the application does not have access to that
-memory. If an application needs to extend a buffer, it must first call
+When a process passes a buffer to
+Read-Write Allow call with, buffers passed on subsequent
```
alevy, line 555:
```suggestion
a Read-Write Allow call, buffers passed on subsequent
```

doc/reference/trd104-syscalls.md
```diff
@@ -545,28 +546,26 @@ support multiple allowed buffers.
...
+
 4.5 Read-Only Allow (Class ID: 4)
 ---------------------------------
 
 The Read-Only Allow class is very similar to the Read-Write Allow class.
-It differs in tow ways: the buffer it passes to the kernel is read-only,
-and the process retains read access to the buffer. The kernel cannot
-write to the buffer. The semantics and calling conventions of
+It differs in one way: the buffer it passes to the kernel is read-only,
+the kernel cannot write to the buffer. The semantics and calling conventions of
```
alevy, line 568:
```suggestion
so the kernel cannot write to the buffer. The semantics and calling conventions of
```

doc/reference/trd104-syscalls.md
```diff
@@ -510,13 +510,14 @@ kernel returns `Failure` with an error code of `NODEVICE`.
 
 The Read-Write Allow system call class is how a userspace process
 shares buffer with the kernel that the kernel can read and write. When
-userspace shares a buffer, it can no longer access it. Calling a
+userspace shares a buffer, it should no longer modify it, though the kernel
+will not stop the app from doing so. Calling a
```
lschuermann, line 514:
```suggestion
userspace shares a buffer, it should no longer modify it.
```

I don't think that comment is particularly useful, and if in the future we had a platform where this was possible to properly prevent, it could still be beneficial in kernel code?

phil-levis: Yes. The point is that the kernel MAY but is not responsible for doing so.

doc/reference/trd104-syscalls.md
```diff
@@ -510,13 +510,13 @@ kernel returns `Failure` with an error code of `NODEVICE`.
 
 The Read-Write Allow system call class is how a userspace process
 shares buffer with the kernel that the kernel can read and write. When
-userspace shares a buffer, it can no longer access it. Calling a
+userspace shares a buffer, it must no longer modify it. Calling a
```
phil-levis, line 513: I think we want to say access. The kernel can write to the buffer. We don't want userspace to view intermediate writes. So I think "it must no longer access it"?

doc/reference/trd104-syscalls.md
```diff
@@ -510,13 +510,13 @@ kernel returns `Failure` with an error code of `NODEVICE`.
...
 Read-Write Allow system call returns a buffer (address and
 length).  On the first call to a Read-Write Allow system call, the
-kernel returns a zero-length buffer. Subsequent successful calls to 
-Read-Write Allow return the previous buffer passed. Therefore, to 
-regain access to a passed buffer, the process must call the same 
-Read-Write Allow system call again. It can do so with a zero-length 
+kernel returns a zero-length buffer. Subsequent successful calls to
+Read-Write Allow return the previous buffer passed. Therefore, to
+regain the ability to write to a passed buffer, the process must call the same
```
phil-levis, line 518: The same as before. I think we want to use the word "access."

doc/reference/trd104-syscalls.md
```diff
@@ -545,28 +545,26 @@ support multiple allowed buffers.
...
 the passed buffer is not complete within the calling process's
 writeable address space, the kernel MUST return a failure result with
 an error code of `INVALID`.
 
-Because a process relinquishes access to a buffer when it makes a
-Read-Write Allow call with it, the buffer passed on the subsequent
-Read-Write Allow call cannot overlap with the first passed buffer.
-This is because the application does not have access to that
-memory. If an application needs to extend a buffer, it must first call
+When a process passes a buffer to
```
phil-levis, line 553: Why remove the explanatory clause? I'm open to removing it, but I had included it because I thought it was useful to explain how overlapping would implicitly violate the lack of access (it means userspace had access before).

doc/reference/trd104-syscalls.md
```diff
@@ -545,28 +545,26 @@ support multiple allowed buffers.
...
+
 4.5 Read-Only Allow (Class ID: 4)
 ---------------------------------
 
 The Read-Only Allow class is very similar to the Read-Write Allow class.
-It differs in tow ways: the buffer it passes to the kernel is read-only,
-and the process retains read access to the buffer. The kernel cannot
-write to the buffer. The semantics and calling conventions of
+It differs in one way: the buffer it passes to the kernel is read-only,
+so the kernel cannot write to the buffer. The semantics and calling conventions of
```
phil-levis, line 567: I think we want to keep both.

doc/reference/trd104-syscalls.md
hudson-ayers: It seems possible there are cases where userspace is willing to accept the possibility of intermediate writes in exchange for a reduction in overhead.

As written, apps are required to unallow and reallow buffers every time they want to read from them. But for many capsules, this is unnecessary - for example, an ADC capsule can promise not to write to an allowed buffer after issuing a sampling_complete() callback. This would allow an application reading results out of that allowed buffer to save the overhead of two system calls to unallow and re-allow the buffer.

Of course, in the general case this is not good practice, and application authors should be careful to investigate their particular use case before trying this. But for a given capsule it is possible to make it completely safe for userspace to do this, and I don't see why we should forbid it in documentation. As the documentation is written, it seems like we would be compelled to request changes on any userspace library submitted to `libtock-c` that relies on this.

doc/reference/trd104-syscalls.md
hudson-ayers: I removed the explanatory clause because I removed the word "access" in favor of "modify" in other places, so it no longer made sense. I am happy to restore it if others agree that "access" is actually what we want here

doc/reference/trd104-syscalls.md
phil-levis: I agree that allow-based swapping is a significant overhead, and we should avoid when it's safe to do so. But I think the problem is that there's a coupling between the liveness of the slice in the kernel and particular calling semantics of commands. To take your ADC example, how do we know that the ADC capsule "promises" that? It can't be that a *capsule* does -- that's an implementation. It has to be that it's specified in the device syscall API that an implementer of the ADC system calls won't do so. 

But historically, our ability or inclination to precisely specify system call APIs has been not great. It's pretty much "this is what the capsule implements, here's the libtock wrapper for it". Since we've been unable to precisely specify the liveness of buffers or safety of operations in system call APIs, we have two choices:
  1. Specify the liveness/ownership of buffers at the system call class level (this is what TRD 104 does).
  2. Change course and precisely specify the system call APIs of different device drivers and then check that our particular implementations follow those specifications.

I took the approach of 1. in 104 because I thought it was wishful thinking to achieve 2. But if we can do 2, then I think it would be totally reasonable for 104 to lay out the criteria for when allowed slices can be accessed by userspace, as well as how a system call API should specify this.

doc/reference/trd104-syscalls.md
phil-levis: See my longer comment above. I think that if we want userspace to be able to access allowed buffers without calling allow again, we will need to require some coupling between allow buffer numbers and callbacks/commands. I.e., allow buffer X can be accessed in the period between callback Y being issued and command Z being called.

doc/reference/trd104-syscalls.md
hudson-ayers: > To take your ADC example, how do we know that the ADC capsule "promises" that? It can't be that a capsule does -- that's an implementation. It has to be that it's specified in the device syscall API that an implementer of the ADC system calls won't do so.

I think it is fine for a device syscall API to specify something like that if needed. I also think that in many real-world use cases of Tock the author of the capsule will be the same as the author of the userspace library that interacts with the capsule, and those users are happy to rely on the implementation as the "promise". Even if we cannot rely on that promise for capsules/userspace code submitted  upstream, we should not have documentation that suggests those users should not make optimizations which are in reality perfectly safe to make if done correctly.

> But historically, our ability or inclination to precisely specify system call APIs has been not great. It's pretty much "this is what the capsule implements, here's the libtock wrapper for it". Since we've been unable to precisely specify the liveness of buffers or safety of operations in system call APIs, we have two choices:
>
> 1. Specify the liveness/ownership of buffers at the system call class level (this is what TRD 104 does).
> 2. Change course and precisely specify the system call APIs of different device drivers and then check that our particular implementations follow those specifications.

I would suggest something like:

3. Specify that device drivers lose access to allowed buffers unless the system call API of that particular device driver promises safe access to allowed buffers under certain conditions.

This prevents us from closing the door on these optimizations in the future, without forcing us to undertake precisely specifying every system call API (which I agree is probably wishful thinking)

phil-levis:
@hudson-ayers @lschuermann and I chatted for an hour or so yesterday. The brief summary is that there might be cases when userspace is allowed *read-only* access to a buffer that has been shared with a read-write allow, because the cost of two allow calls (one to revoke the buffer, the second to re-allow it) is too large. We mostly coalesced on a framing in which this access is allowed only while a kernel operation is not ongoing: there is no command whose completion is pending, e.g., through an upcall.

phil-levis:
https://github.com/tock/tock/blob/trd104-buffer-reading/doc/reference/trd104-syscalls.md

Just read 4.4 (and the new 4.4.1). This is just a first try.

hudson-ayers:
I read through it, it mostly looks good. Feel free to overwrite the changes in this PR with those. One comment:

> Userspace MUST NOT write a buffer that has been shared with the kernel with a Read-Write Allow call. The writing restriction is because changing the buffer mid-operation might violate invariants within the kernel and cause a panic. Rust type safety depends on there not being arbitrary concurrent writers.

It is unacceptable for type safety in the kernel to be dependent on the behavior of userspace apps, and that is not true today -- whenever a userspace app is running, `AppSlice` is not stored as a reference in the kernel, but as a pointer and a length. So an app writing to a shared buffer does not violate type safety in the kernel. However, it may lead to incorrect capsule behavior if a capsule chooses to manually cache values read from allowed buffers across invocations of the scheduler, under the assumption that an allowed buffer will not change.

phil-levis:
> It is unacceptable for type safety in the kernel to be dependent on the behavior of userspace apps, and that is not true today -- whenever a userspace app is running, `AppSlice` is not stored as a reference in the kernel, but as a pointer and a length. So an app writing to a shared buffer does not violate type safety in the kernel. However, it may lead to incorrect capsule behavior if a capsule chooses to manually cache values read from allowed buffers across invocations of the scheduler, under the assumption that an allowed buffer will not change.

Yes, exactly. But I think your first sentence and last sentence are mutually exclusive, unless we want to manually inspect all code that touches an `AppSlice` to check that it doesn't cache values. I used the word "invariants" to try to express this; I agree that the "type safety" sentence is incorrect and should be removed.

hudson-ayers:
> Yes, exactly. But I think your first sentence and last sentence are mutually exclusive, unless we want to manually inspect all code that touches an `AppSlice` to check that it doesn't cache values. I used the word "invariants" to try to express this; I agree that the "type safety" sentence is incorrect and should be removed.

I would recommend the following text: 

"Userspace MUST NOT write a buffer that has been shared with the kernel with a Read-Write Allow call. The writing restriction is because changing the buffer mid-operation might violate invariants within capsules, and thus cause a panic or otherwise incorrect behavior."

jrvanwhy:
The threat model says that processes shouldn't be allowed to deny service to other processes. Wouldn't a capsule that panics if a process writes memory at the wrong time violate that?

hudson-ayers:
> The threat model says that processes shouldn't be allowed to deny service to other processes. Wouldn't a capsule that panics if a process writes memory at the wrong time violate that?'

Yes, a capsule written in this way would violate the threat model. Unfortunately, we don't have a good solution for ensuring that capsules can never panic on the basis of user input (not just about buffers). We basically just rely on manual review of capsule code for this. Until we have a good solution to that issue, guidelines like this one can be helpful in encouraging users to write apps in such a way that is less likely to cause issues.

One thing that bothers me about using "MUST NOT" in the way this document does is that some readers might interpret that as saying the kernel can rely on these things always being true, which is not the case. Instead, I believe "MUST NOT" is used more as "if you write an app that does this, the capsule you are interacting with is not guaranteed to behave in the way you expect".

jrvanwhy:
How about the following?

"Userspace MUST NOT write a buffer that has been shared with the kernel with a Read-Write Allow call. The writing restriction is because changing the buffer mid-operation might allow the kernel to observe a partially-written state, which is complex to handle correctly. Conversely, the kernel MUST limit the disruption caused by a process changing a shared buffer to that process."

My intention is to ban the behavior so we can rely on it for backwards-compatibility while indicating that capsules cannot panic if processes ignore the ban.

phil-levis:
> 
> 
> The threat model says that processes shouldn't be allowed to deny service to other processes. Wouldn't a capsule that panics if a process writes memory at the wrong time violate that?

I think I'd agree with Hudson on this. For example, if a capsule is carelessly written and can panic in response to process input, is it the process that denied service or the capsule?

phil-levis:
> 
> 
> How about the following?
> 
> "Userspace MUST NOT write a buffer that has been shared with the kernel with a Read-Write Allow call. The writing restriction is because changing the buffer mid-operation might allow the kernel to observe a partially-written state, which is complex to handle correctly. Conversely, the kernel MUST limit the disruption caused by a process changing a shared buffer to that process."
> 
> My intention is to ban the behavior so we can rely on it for backwards-compatibility while indicating that capsules cannot panic if processes ignore the ban.

I agree with the sentiment of this but I think "limit the disruption" is too vague for a MUST. Can we make it more precise?

phil-levis:
How about

"Userspace MUST NOT write a buffer that has been shared with the kernel with a Read-Write Allow call. The writing restriction is because changing the buffer mid-operation might allow the kernel to observe a partially-written state, which is complex to handle correctly.  At the same time, the kernel MUST NOT read data from a buffer in a non-atomic way such that a bug in a userspace program that writes the buffer could cause the kernel to panic."

jrvanwhy:
> I think I'd agree with Hudson on this. For example, if a capsule is carelessly written and can panic in response to process input, is it the process that denied service or the capsule?

The way I see it, the capsule has a denial of service vulnerability and the process exploited it.

> How about
>
> "Userspace MUST NOT write a buffer that has been shared with the kernel with a Read-Write Allow call. The writing restriction is because changing the buffer mid-operation might allow the kernel to observe a partially-written state, which is complex to handle correctly. At the same time, the kernel MUST NOT read data from a buffer in a non-atomic way such that a bug in a userspace program that writes the buffer could cause the kernel to panic."

That sounds good to me.

phil-levis:
> That sounds good to me.

Pushed. I moved it into 4.4 since it's about all cases, not just the reading case.

alistair23:
Can we add something as well to say that userspace can write to a Read Allow call. For example to send data an app can fill a buffer, trigger a send and then once complete fill the buffer again and repeat.

As well as that apps should be able to read buffers at any time. That will allow https://github.com/tock/tock/pull/2381 but also it's not something that we can stop, although we don't always guarantee the integrity of the data being read.

phil-levis:
> 
> 
> Can we add something as well to say that userspace can write to a Read Allow call. For example to send data an app can fill a buffer, trigger a send and then once complete fill the buffer again and repeat.

When is this a problem? It comes up on the opposite direction because userspace can't control the rate of incoming events.

> As well as that apps should be able to read buffers at any time. 

Please follow the discussion. This is not OK. It's not a good idea to completely redefine the buffer sharing abstractions and whole memory access guarantees of the operating systems for one specific use case related to needing to get time.

alistair23:
It's hard to follow along as it seems like a lot of the discussion happens in separate calls.

> When is this a problem? It comes up on the opposite direction because userspace can't control the rate of incoming events.

This PR says `Therefore, to regain the ability to write to a passed buffer, the process must call the same Read-Write Allow system call again.` 

This adds a lot of syscall overhead for an app to send data. For example if there is a capsule to send data over USB or Bluetooth.  The app can create a 32-byte buffer and write the data it wants to send. It then calls `allow` to pass it to the kernel. Once the operation is complete (after a callback) we shouldn't require the app to perform two more allow syscalls to re-use that buffer. Instead the app should just be able to copy data into the buffer and call the `command` again.

This would only apply to Read allow calls though. It's hard to tell if this is allowed or not from the current writing (it doesn't seem to be).

> Please follow the discussion. This is not OK. It's not a good idea to completely redefine the buffer sharing abstractions and whole memory access guarantees of the operating systems for one specific use case in one application related to needing to get time. There are a lot of better ways to do this that don't have far-reaching implications to all system call code.

Like I said it's really hard to follow this when there are separate calls happening. From what I can see here this is what Hudson suggested earlier and I haven't seen anyone saying it's not acceptable.

Hudon said this before `Specify that device drivers lose access to allowed buffers unless the system call API of that particular device driver promises safe access to allowed buffers under certain conditions.` and I don't see anyone say that it's not an option.

We (as in the Tock kernel) have no way of stopping this, so I don't see how we can really say it's not allowed. There are lot of uses cases for this, such as Hudson's ADC example above. For reading a single byte ADC value buffer the overhead of having to call re-allow a buffer is large compared to the impact of getting some stale data.

hudson-ayers:
> This has been part of the core of syscall semantics for Tock 2.0; we've been talking about it for about 7 months, as well as general issues of AppSlice aliasing/soundness.

I don't think we should state that this is an issue of aliasing/soundness. If an app doing something can cause unsoundness in the kernel, that is a bug in the kernel. For the particular issue of a user application modifying a buffer passed via a `ReadOnlyAllow` call, it is the case today that this does not cause unsoundness (when userspace is running, the kernel holds no *references* to app memory. it only holds a pointer/length combo).

I share Alistair's concern that wording the document the way it is now may mistakenly lead capsule authors to rely on this guarantee that cannot exist (at least today).

@alistair23 one aspect of this document that we discussed on our out-of-band call is that it seems possible future hardware might support far more MPU regions with finer granularity, such that this *could* in fact be enforced by the kernel. In that case, it might be possible to substantially simplify the in-kernel `AppSlice` code to hold references directly. But if we promise today that apps *can* do this, we cannot retract that promise without breaking backwards compatibility.

phil-levis:
> We (as in the Tock kernel) have no way of stopping this, so I don't see how we can really say it's not allowed. There are lot of uses cases for this, such as Hudson's ADC example above. For reading a single byte ADC value buffer the overhead of having to call re-allow a buffer is large compared to the impact of getting some stale data.

If your concern is efficiency, don't read a single ADC value. Or write an ADC driver that returns a single sample in a callback.

There are three reasons you can be concerned about the cycle count for the allow semantics:
  1. they limit throughput because callbacks are at a high rate and the cost of calling allow limits this,
  2. they are an energy overhead because they are a significant fraction of CPU cycles for ultra-low duty cycle applications, or
  3. they add latency for low-latency applications.

I don't think we're very worried about 2. or 3.; for 2. the energy costs of those cycles are going to be pretty tiny compared to the I/O, while for 3. we are talking about pretty small latencies (maybe 10us tops), and if that's an issue a scheduling quantum or an interrupt is a much bigger concern.

So we're talking about 1. Generally speaking, Tock was never really designed for high throughput applications. There really aren't that many cases of it: in most of them (e.g., Hudson's ADC example), we have abstractions to try to help with this, such as large buffer high-speed ADC sampling. If Tock did prioritize these workloads, we would have a different I/O architecture, interrupt handling architecture, and better support for memory movement within the kernel.

We can say it's not allowed in that if you do it, it can lead to undefined behavior. Of course an assembly program can do it; we're just saying that if it does, things can go kaboom. This is the same as saying you're not allowed to dereference the null pointer. If you language and libraries can actually prevent you from doing this (e.g., when you pass a buffer to the kernel through allow, you lose ownership of it), then that's great. But it's important to clearly delineate who can write what memory when, otherwise how do you enforce memory safety.

> I don't think we should state that this is an issue of aliasing/soundness.

I'm looking at it from the other direction, if the kernel can arbitrarily write to a userspace buffer. Unlike the kernel, which can make some assumptions on read/write atomicity, userspace can't. In my experience, forcing applications define their data consistency models without much support from the kernel leads to incorrect applications. Atomicity and consistency are hard problems in the presence of concurrency. There might be some trivial use cases when it's not a big deal, but that's not a good reason to not support more complex use cases. E.g., sure, `O_APPEND` is great, but you still need `fsync`.

hudson-ayers:
> > I don't think we should state that this is an issue of aliasing/soundness.
> 
> I'm looking at it from the other direction, if the kernel can arbitrarily write to a userspace buffer. Unlike the kernel, which can make some assumptions on read/write atomicity, userspace can't. In my experience, forcing applications define their data consistency models without much support from the kernel leads to incorrect applications. Atomicity and consistency are hard problems in the presence of concurrency. There might be some trivial use cases when it's not a big deal, but that's not a good reason to not support more complex use cases. E.g., sure, `O_APPEND` is great, but you still need `fsync`.

The kernel can't write to buffers allowed via `ReadOnlyAllow`, so I don't think this applies

phil-levis:
> The kernel can't write to buffers allowed via `ReadOnlyAllow`, so I don't think this applies

I was saying that we have been talking about the allow semantics for 7 months, as well as general issues of aliasing/soundness, not that the particular case of sending data had issues of aliasing/soundness. I was responding to

> As well as that apps should be able to read buffers at any time.

bradjc:
> there might be cases when userspace is allowed _read-only_ access to a buffer that has been shared with a read-write allow, because the cost of two allow calls (one to revoke the buffer, the second to re-allow it) is too large. We mostly coalesced on a framing in which this access is allowed only while a kernel operation is not ongoing: there is no command whose completion is pending, e.g., through an upcall.

At the risk of wading into a thorny discussion, why do we not have a `WriteOnly` allow() that provides these semantics? This would I think be intuitive for things like ADCs, microphones, and RNGs.

In this hypothetical future, where userspace app slice access restrictions are enforced by hardware, apps have to be very careful how they use this access, because the kernel traps the access fault, not the app itself. So it seems that if an app _sometimes_ has access to an allowed buffer and sometimes not, app implementations would likely be conservative and request the buffer back rather than chance an access which could fault. By moving this feature to its own type of allow we might be able to provide a clearer guarantee.

phil-levis:
> At the risk of wading into a thorny discussion, why do we not have a `WriteOnly` allow() that provides these semantics? This would I think be intuitive for things like ADCs, microphones, and RNGs.

What benefit does this bring? The issue is reading an intermediate state as the kernel writes it. Precluding the kernel from reading the data wouldn't help. You need mechanisms so that userspace knows when data is in a consistent state.

phil-levis:
Leon, Hudson and I discussed this some more today. We agreed with a framing that covers two points:
  1. Capsules can't assume that userspace doesn't change or read an allowed buffer. Even if the capsule prevents a process from revoking a buffer, the process might crash or the process might modify the bytes of data. So capsules need to be robust to this.
  2. At the same time, capsules can't assume that userspace libraries modify allowed buffers. If a capsule API wants to allow userspace to access an allowed buffer (without revoking it), this is an *optimization* and can't be required. This is to ensure that capsules support libtock implementations which do not allow access to allowed buffers. This means, for example, we would not accept a system call API which depends on userspace just looking at the buffer (rather than an upcall to signal when it should be revoked).

phil-levis:
https://github.com/tock/tock/blob/trd104-buffer-reading/doc/reference/trd104-syscalls.md

has another try.

hudson-ayers:
> In these cases, it can be acceptable for a system call driver to allow userspace to read or write a buffer passed with a Read-Write Allow.

I recommend "current shared via a Read-Write Allow"

> A process MUST NOT write to a buffer shared with the kernel through a Read-Only Allow.

On our call, I thought we had decided that it would be permitted subject to the same rules as userspace reading from an allowed buffer?

lschuermann:
> > A process MUST NOT write to a buffer shared with the kernel through a Read-Only Allow.
> 
> On our call, I thought we had decided that it would be permitted subject to the same rules as userspace reading from an allowed buffer?

I was under the impression we did want to permit reads in certain cases for R/W allows, but never writes to all shared buffers. The reasoning for this being that, if in the future we had memory protection that would allow to enforce this, we could avoid potentially complex and expensive consistency and runtime checks at every invocation of driver code (as the buffer might changed in between).

hudson-ayers:
I think there was a misunderstanding during the call then, that's not how I recall our conclusion. No worries though, I may be the one that misunderstood.

To lay out my feelings on that in text, for others to see:

- Capsules today should already be designed to follow the threat model in the presence of userspace writes to buffers shared via `ReadOnlyAllow`. So I don't think that saying "apps MUST NOT do this" represents any sort of safety win.
- I think that a future where most MCU's that Tock supports have such high-performance memory protection is likely far away, and if it ever exists will likely inspire a number of other changes to the system call interface. Designing for that future would require the Tock kernel to either deprecate all of our currently supported platforms, or design kernel code such that AppSlice is an entirely different type depending on the MPU support available. Thus it seems likely (to me) that a major version bump would be required to optimize for this sort of hardware anyway.
- Even in a future where MPUs offer much greater granularity and numbers of regions, it seems likely that the cost of reconfiguring the MPU for these buffers on every allow might be comparable to the cost of performing these checks today. Adding and removing RISC-V PMP regions today requires iterating through every other region and checking for overlap.
- There are performance optimizations available *today* for allowing a userspace buffer to write to a buffer shared via a `ReadOnlyAllow` (e.g. the packet sending example Alistair mentioned). In my opinion, the benefits of allowing these optimizations outweigh the costs of some hypothetical future where MPU hardware is greatly different than what is available today.

alistair23:
From a RISC-V perspective for the MPU/PMP. There isn't a visible future where hardware can apply enforcement for apps to access allowed buffers where the buffers are allocated inside app memory.

There are a few reasons for this. The first is that although some CPUs might provide enough regions (think 32 or more PMP regions) we probably will never get to a point where all the CPUs Tock aims to support ship this number (OpenTitan only has 16). As Hudson mentions the overhead of loading and restoring all of those regions will be large. As well as that even with a large number such as 32 PMP regions, that really only allows around 8 allowable buffers per app. It seems entirely possible that some apps will have more then that.

lschuermann:
@hudson-ayers I agree with most of your observations and don't have a strong opinion on this particular issue. I think that both allowing and forbidding userspace to write to shared memory is a workable solution, with different tradeoffs.

> Capsules today should already be designed to follow the threat model in the presence of userspace writes to buffers shared via `ReadOnlyAllow`. So I don't think that saying "apps MUST NOT do this" represents any sort of safety win.

It's correct that capsules must be tolerant to this behavior (at least to not cause any inherent safety, data leakage or denial of service if such behavior is observed), _if_ we can't enforce in hardware for this to not be able to happen. Still, under the premise of that being possible through hardware in the future (which may well be incorrect!), I believe ["MUST NOT" as defined by RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119#section-2) would be the correct phrasing for such a requirement. It seems fine to declare some behavior as invalid, although it is technically possible (for example, I could build a TCP implementation which does not understand the Maximum Segment Size option, however it would make it non-compliant to RFC 793). If we wanted to, in the future, prevent writes to shared buffers and we don't declare that behavior as incorrect now (e.g. through a _MUST NOT_), this would mean a breaking ABI change.

> There are performance optimizations available _today_ for allowing a userspace buffer to write to a buffer shared via a `ReadOnlyAllow` (e.g. the packet sending example Alistair mentioned). In my opinion, the benefits of allowing these optimizations outweigh the costs of some hypothetical future where MPU hardware is greatly different than what is available today.

I think that is true, in particular when looking a producer-consumer ring buffers from userspace to the kernel. However, if we were to define this acceptable behavior, we're going to want to have an exact per-capsule specification of when this allowed (regardless of the fact that it's technically always possible), as established for reading a shared R/W allow.

hudson-ayers:
> It's correct that capsules must be tolerant to this behavior (at least to not cause any inherent safety, data leakage or denial of service if such behavior is observed), _if_ we can't enforce in hardware for this to not be able to happen. Still, under the premise of that being possible through hardware in the future (which may well be incorrect!), I believe ["MUST NOT" as defined by RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119#section-2) would be the correct phrasing for such a requirement. It seems fine to declare some behavior as invalid, although it is technically possible (for example, I could build a TCP implementation which does not understand the Maximum Segment Size option, however it would make it non-compliant to RFC 793, and I must not be able to cause safety issues on my peer's side with that). If we wanted to, in the future, prevent writes to shared buffers and we don't declare that behavior as incorrect now (e.g. through a _MUST NOT_), this would mean a breaking ABI change.

Yeah, agreed, I just wanted to make the point that the only justification for including a _MUST NOT_ today seems to be for future proofing (as a set up for my next argument that this future proofing seems unlikely to ever pay dividends)

> I think that is true, in particular when looking a producer-consumer ring buffers from userspace to the kernel. However, if we were to define this acceptable behavior, we're going to want to have an exact per-capsule specification of when this allowed (regardless of the fact that it's technically always possible), as established for reading a shared R/W allow.

Yes, my understanding was that we were going to impose the same requirements as we did for reading a shared R/W allow

phil-levis:
> 
> 
> I think there was a misunderstanding during the call then, that's not how I recall our conclusion. No worries though, I may be the one that misunderstood.
> 
> To lay out my feelings on that in text, for others to see:
> 
>     * Capsules today should already be designed to follow the threat model in the presence of userspace writes to buffers shared via `ReadOnlyAllow`. So I don't think that saying "apps MUST NOT do this" represents any sort of safety win.

I'd argue that it doesn't provide a safety win in theory, but it does in practice.

**If** all capsule code perfectly follows the requirement that it not assume a buffer might change under it, and so checks everything and is always perfectly careful, then you're right, we don't need this restriction. It's fine to write capsules that expect userspace to write to Read-Only allowed buffers. So in theory, it doesn't help.

But I don't think this is the case. I think instead we should be paranoid and say that you don't write to a buffer allowed to the kernel. Of course, a bug might cause you to. But the point is that you should not go around designing APIs that do this. That being said, I think that the sentence is in this way mis-targeted. It should instead say "System call APIs MUST NOT include or depend on userspace writing to a buffer that has been given to the kernel with a Read-Only allow."

>     * There are performance optimizations available _today_ for allowing a userspace buffer to write to a buffer shared via a `ReadOnlyAllow` (e.g. the packet sending example Alistair mentioned). In my opinion, the benefits of allowing these optimizations outweigh the costs of some hypothetical future where MPU hardware is greatly different than what is available today.

I think this performance benefit is pretty inconsequential. For it to be significant, all of these would have to hold:
  1. You have a low clock speed  to bus/link speed ratio.
  2. You are allowing tiny buffers and can't do any aggregation.
  3. You are trying to send very fast and care about messages/second or throughput.

I'd argue that Tock was never designed for these circumstances, and if it were, we would do I/O very differently. Can someone present a concrete example where this overhead is a problem?

This is distinct from the receive (Read-Write allow) case because those messages/operations come from an external source. So you cannot control their arrival frequency. E.g., another device might, due to its own I/O scheduling, send a brief burst of 3 small packets and you want to be able to handle that.

phil-levis:
AND WHY THE HELL IS THE CLOSE BUTTON RIGHT NEXT TO THE COMMENT BUTTON!?!?!? :)

hudson-ayers:
> I'd argue that it doesn't provide a safety win in theory, but it does in practice.
> 
> **If** all capsule code perfectly follows the requirement that it not assume a buffer might change under it, and so checks everything and is always perfectly careful, then you're right, we don't need this restriction. It's fine to write capsules that expect userspace to write to Read-Only allowed buffers. So in theory, it doesn't help.
> 
> But I don't think this is the case. I think instead we should be paranoid and say that you don't write to a buffer allowed to the kernel. Of course, a bug might cause you to. But the point is that you should not go around designing APIs that do this. 

This is fair, but I think if the reason we say that capsules can't do this is "designing and using such an API correctly is hard" we should at least be honest about that in the text. Otherwise someone could come to this document in a few years and struggle to figure out why this is forbidden, or if there is some deeper technical reason that it is impossible to allow this.

> That being said, I think that the sentence is in this way mis-targeted. It should instead say "System call APIs MUST NOT include or depend on userspace writing to a buffer that has been given to the kernel with a Read-Only allow."

I prefer this wording

> I think this performance benefit is pretty inconsequential. For it to be significant, all of these would have to hold:
> 
> 1. You have a low clock speed  to bus/link speed ratio.
> 2. You are allowing tiny buffers and can't do any aggregation.
> 3. You are trying to send very fast and care about messages/second or throughput.
> 
> I'd argue that Tock was never designed for these circumstances, and if it were, we would do I/O very differently. Can someone present a concrete example where this overhead is a problem?
>
> This is distinct from the receive (Read-Write allow) case because those messages/operations come from an external source. So you cannot control their arrival frequency. E.g., another device might, due to its own I/O scheduling, send a brief burst of 3 small packets and you want to be able to handle that.

This is a reasonable point. I am fine with leaving the RO allow the way it is now, modulo the change to the text you suggested above, if no one can present a use case that meets all 3 of the requirements you mentioned.

phil-levis:
I pushed some updates based on the feedback.

lschuermann:
@phil-levis

> An important invariant is that userspace does use buffer data read while an operation that modifies the buffer is ongoing.

I think that should be a "does _not_".

> In these cases, it can be acceptable for a system call driver to allow userspace to read or write a buffer passed with a Read-Write Allow. An important invariant is that userspace does not use buffer data read while an operation that modifies the buffer is ongoing. Instead, userspace only uses data collected from a buffer before or after an operation completes: data reads and operations MUST be atomic with respect to one another.
> 
> For a system call API to allow userspace to read allowed buffers, the mechanism that ensures data reads are atomic MUST be documented in a Draft or Final Documentary TRD. If this conditions are met then userspace MAY read the buffer under the conditions specified in the TRD.
> 
> One example mechanism for atomic access is to specify a state machine for that buffer as described below. Other mechanisms (e.g., based on monotonic counters) can be acceptable as well.

I'm pretty happy with this text in general. I think it's important to distinguish between what technically can happen once a buffer is RW-allowed (the raw ABI contract), and how we define individual capsules' API contracts. I'm under the impression this specific text does a good job of limiting the ways people can misuse this interface accidentally, but still leaving great potential for optimizations.

If I understand it's intentions correctly, it on the one hand enables us to ship pre-built mechanisms for efficient userspace-kernel data exchange (such as [lock-free ring buffers](https://ferrous-systems.com/blog/lock-free-ring-buffer/)), which are thoroughly verified to enable safe & consistent data exchange, have formally defined semantics through TRDs, and be ready to use by capsules. On the other hand, it prevents capsules from simply rolling their own optimizations, at least without formal specification and careful consideration.

Without trying to get to pedantic, some minor questions:
- this invariant as documented will prevent userspace to read any region of the buffer, even if there is a formally documented mechanism that clearly partitions the buffer into regions which are subject to change as a result of a running operation, and regions which are safe to read. Do we want to restrict the text to say "userspace does not use buffer data read while an operation that modifies this data is ongoing"? That would allow us to build data structures such as the aforementioned lock-free ring buffers.
- the TRD talks about it being acceptable to "read or write a buffer passed with a Read-Write Allow". After hearing @alistair23's and @hudson-ayers's feedback I think this is good; both should be potentially allowed _if and only if_ there is a formally documented exception through a TRD. However, the section below is only concerned with reads. Should we also extend this to include writes?

phil-levis:
>     * the TRD talks about it being acceptable to "read or write a buffer passed with a Read-Write Allow". After hearing @alistair23's and @hudson-ayers's feedback I think this is good; both should be potentially allowed _if and only if_ there is a formally documented exception through a TRD. However, the section below is only concerned with reads. Should we also extend this to include writes?

See the discussion above. I agree that we need it in the read case because of the arrival of external events. But needing it in the write case is based on a pretty narrow set of conditions. If we have a throughput/write limited application, then single buffer operations are not the way to go (it's a stop-and-go protocol). I'd argue that instead we should design a separate API that's designed for such circumstances.

phil-levis:
> * this invariant as documented will prevent userspace to read any region of the buffer while an operation is ongoing, even if there is a formally documented mechanism that clearly partitions the buffer into regions which are subject to change as a result of a running operation, and regions which are safe to read. Do we want to restrict the text to say "userspace does not use buffer data read while an operation that modifies this data is ongoing"? That would allow us to build data structures such as the aforementioned lock-free ring buffers.

I tried to word this carefully but I think I missed. My *intention* was that it would allow this. But I think the text doesn't quite say that. What the text was trying to say is that you can **read** the data while an operation is ongoing (literally, you can perform a memory load) but that you can't **use** data that is inconsistent. I've updated the text to try to refine this:

> An important invariant is that userspace MUST NOT use buffer data  that is read while that data is in an inconsistent state. It MAY read the  data, detect it is inconsistent, and discard it. Instead, userspace only uses data collected from a buffer *before* or *after* an operation completes: data reads and operations MUST be atomic with respect to one another. 

Is this clearer?

lschuermann:
> Is this clearer?

Yes, I think this makes it obvious as to how an app must behave.

> I'd argue that instead we should design a separate API that's designed for such circumstances.

I think that's fair, for efficient bi-directional data transfer we clearly want to have a carefully designed & formally specified API, not just something which appears to "just work".

My remaining question is: _can_ we potentially build this mechanism on top of a R/W allow in the future, or would we need to use an entirely different syscall class? Given the text says "it can be acceptable for a system call driver to allow userspace to read or write a buffer passed with a Read-Write Allow", I believe the former to be the case. If that is indeed the intention, I think the TRD is sufficiently open for now as to give the ability to build something in the future, while not providing a clear statement on how this can be done today. I'd be happy with this.

hudson-ayers:
I am happy with the most recent version of the text that Phil shared

jrvanwhy:
I intended for `libtock-rs`' read-write allow function to have a signature similar to:

```rust
fn allow_rw_static(driver_id: u32, buffer_id: u32, buffer: &'static mut [u8])
    -> Result<&'static mut [u8], (ErrorCode, &'static mut [u8])>
```

On its own, this is incompatible with simultaneous use by the process binary and kernel, because there is no safe way for the process binary to mutate the buffer after a successful `allow_rw_static` call. The benefit of this incompatibility is it works with `&'static mut [u8]` references, which are a very natural type to work with in Rust.

If we want to allow for simultaneous use by the process binary and kernel, we would need something like:

```rust
struct VolatileByte { data: UnsafeCell<u8> }
impl VolatileByte { } // Useful getters/setters

fn allow_rw_static_shared(driver_id: u32, buffer_id: u32, buffer: &'static [VolatileByte])
    -> Result<&'static [VolatileByte], (ErrorCode, &'static [VolatileByte])>
```

Then calling code can still access the buffer after it has been shared. However, the interface is far less pleasant to work with.

## The Conflict

`libtock-rs` can only expose one of the above methods. If `libtock-rs` were to expose both `allow_rw_static` and `allow_rw_static_shared`, then safe code could invoke undefined behavior:

```rust
fn main() {
    static BUFFER: [VolatileByte; 1] = [VolatileByte::new(0)];
    allow_rw_static_shared(1, 1, &BUFFER);
    let buffer_ref: &'static mut [u8] = allow_rw_static(1, 1, &[]).unwrap();
    BUFFER[0].set(1);  // Undefined behavior: mutated BUFFER while it is borrowed by buffer_ref
}
```

I don't think that `allow_rw_static` should be `unsafe`, because it is a commonly-used function (I expect nearly every syscall API interface to use it). Making `allow_rw_static_shared` `unsafe` is not sufficient, because as long as `allow_rw_static` is safe then any use of `allow_rw_static_shared` allows safe code to retrieve the buffer as a `&'static mut [u8]`.

## My suggested solution

I suggest we make shared-access Allow its own syscall class, separate from read-write allow. That way userspace code can't mix buffers that should have different types (`VolatileByte` versus `u8`).

phil-levis:
> I suggest we make shared-access Allow its own syscall class, separate from read-write allow. That way userspace code can't mix buffers that should have different types (`VolatileByte` versus `u8`).

Given how complicated this is getting, this seems like a very good approach. @alistair23 and @lschuermann can come up with an API that fits both of their use cases (Ethernet packet reception and scheduler time hooks). This shouldn't require us to bump a major version. We can write a separate TRD for the new system call (Allow-Shared?).

jrvanwhy:
> > I suggest we make shared-access Allow its own syscall class, separate from read-write allow. That way userspace code can't mix buffers that should have different types (`VolatileByte` versus `u8`).
> 
> Given how complicated this is getting, this seems like a very good approach. @alistair23 and @lschuermann can come up with an API that fits both of their use cases (Ethernet packet reception and scheduler time hooks). This shouldn't require us to bump a major version. We can write a separate TRD for the new system call (Allow-Shared?). Thoughts?

I agree that adding a syscall class after 2.0 would not require a major version bump. Putting shared Allow into a separate TRD sounds good to me.

alistair23:
Would we allow read and write access from the app, or just read?

alistair23:
I had a first stab at what I expect the implementation to look like: https://github.com/tock/tock/pull/2610

I've left the TRD changes to this PR, let me know if you want me to write the changes

hudson-ayers:
This is now obsoleted by a combination of #2610 and #2617

PR #2595, opened 2021-05-24
OpenTitan: Initial support for OTBN

alistair23:
### Pull Request Overview

This is the first step to supporting the [OpenTitan Big Number Accelerator (OTBN)](https://docs.opentitan.org/hw/ip/otbn/doc/index.html).

This PR adds full support for the OTBN including virtuliser support. The design no longer user the accelerator HIL.

### Testing Strategy

Running on the OT FPGA.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


kernel/src/hil/accel.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// data supplied to `load_binary()`.
+    fn binary_load_done(&'a self, result: Result<(), ErrorCode>, input: &'static mut [u8]);
+
+    /// This callback is called when a operation is computed.
+    /// On error or success `output` will contain a reference to the original
+    /// data supplied to `run()`.
+    fn op_done(&'a self, result: Result<(), ErrorCode>, output: &'static mut [u8; T]);
+}
+
+/// A generic accellerator. This can be used to accelerate any type of
```
hudson-ayers, line 19:
```suggestion
/// A generic accelerator. This can be used to accelerate any type of
```

kernel/src/hil/accel.rs
```diff
@@ -0,0 +1,52 @@
+//! Interface for Accelerators
```
hudson-ayers, line 1:
```suggestion
//! Interface for Hardware accelerators
```

kernel/src/hil/accel.rs
```diff
@@ -0,0 +1,52 @@
...
+    ) -> Result<(), (ErrorCode, &'static mut [u8])>;
+
+    /// Run the acceleration operation.
+    /// This doesn't return any data, instead the client needs to have
+    /// set a `op_done` handler to determine when this is complete.
+    /// On error the return value will contain a return code and the original data
+    /// If there is data from the `load_binary()` command asyncrously waiting to
+    /// be written it will be written before the operation starts.
+    fn run(&'a self, output: &'static mut [u8; T])
+        -> Result<(), (ErrorCode, &'static mut [u8; T])>;
```
hudson-ayers, line 46: Why pass a sized array instead of a slice? It seems possible some accelerators would want to support variable length inputs

kernel/src/hil/accel.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// configure the accelerator.
+    /// This function can be called multiple times if multiple binary blobs
+    /// are required.
+    /// There is no guarantee the data has been written until the `binary_load_done()`
+    /// callback is fired.
+    /// On error the return value will contain a return code and the original data
+    fn load_binary(
+        &self,
+        input: LeasableBuffer<'static, u8>,
+    ) -> Result<(), (ErrorCode, &'static mut [u8])>;
```
hudson-ayers, line 37: It seems strange to pass in a `LeasableBuffer` but get back a slice in the error case

kernel/src/hil/accel.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use crate::common::leasable_buffer::LeasableBuffer;
+use crate::ErrorCode;
+
+/// Implement this trait and use `set_client()` in order to receive callbacks.
+pub trait Client<'a, const T: usize> {
+    /// This callback is called when the binary data has been loaded
+    /// On error or success `input` will contain a reference to the original
+    /// data supplied to `load_binary()`.
+    fn binary_load_done(&'a self, result: Result<(), ErrorCode>, input: &'static mut [u8]);
```
hudson-ayers, line 11: This interface seems somewhat OTBN specific -- there are plenty of hardware accelerators that don't load a binary in order to be used.

hudson-ayers (commented):
I sort of feel that a "generic accelerator" might be the wrong level of abstraction for a HIL. There are all sorts of different hardware accelerators that might be best used in many different ways. "Crypto accelerator" might be closer to what we want, but even there the OTBN docs state "OTBN is not designed to be compatible with other cryptographic accelerators". At the very least I think the HIL should attempt to describe the class of accelerators for which this interface is well suited, with more than one example

kernel/src/hil/accel.rs
alistair23: This is an output, the idea is that the output is fixed by the hardware

kernel/src/hil/accel.rs
alistair23: If no binary data is required then this doesn't have to be called. So I don't see that it would break non OT users

kernel/src/hil/accel.rs
alistair23: This is the same thing we do with the HMAC. The idea is to pass in a buffer and a length and then we just return the buffer. In this case though we skip the length argument by using the `LeasableBuffer`

alistair23:
After a discussion this morning I realised that the accelerator shouldn't be exposed to userspace (maybe just for a PoC, but not for real world use).

In light of that we probably don't need a generic HIL for accelerators so we might not want to go down this route. In saying that I think a HIL is still really useful and would allow a virtual_accel.rs capsule that we can use to virtualise the accelerator. This way we can build multiple kernel capsules on top of it.

I have tried to design the HIL to be pretty flexible to allow future platforms to use it as well.

kernel/src/hil/accel.rs
hudson-ayers: My mistake. Still, it seems possible some accelerators would size their outputs based on the size of the input

alistair23: My worry with an output buffer is what if a user passes a too small buffer.

For example if the output of the accelerator is 64 bytes, but a user passes a `[u8; 32]` we will throw a `NOMEM` error (instead of only half filling it), but then the user has no way of knowing how big the buffer really should be. My expectation is that the output buffer size of a accelerator is constant (although the useful information will change based on protocol).

The other way of handling it could be just to trust the caller and if they pass a smaller buffer just fill the small buffer as they might not want the entire output. That seems prone to errors though.

If I change it to a slice what should we do for length mismatches?

hudson-ayers: That's a reasonable point. If its a slice I think that returning an Error on length mismatches would be fine -- it could be up to the caller to pass an appropriate length buffer. I suppose it is fine to leave it as fixed size until we have a concrete example of an accelerator that does not provide a fixed size output

alistair23:
@bradjc and others. I have updated this to no longer use a HIL. It's now only exposed to OT itself. Let me know what you think

bradjc (approved):
Would be good to change the virtualizer name to VirtualOtbn or something like that.

alistair23:
Changed the names

bradjc (approved):

PR #2599, opened 2021-05-26
Add integration test support in the kernel, create tests for OpenTitan

alistair23:
### Pull Request Overview

Although Tock does currently have some automated tests they currently rely on running the binary somewhere (either QEMU or hardware) and scraping the output. This works fine for some things, but it doesn't allow us to easily check for expected failures or run small unit tests without code modification.

This PR extends the OpenTitan board to use rust's built in test infrastructure. With this we can now run `make test` which under the hood calls `cargo test` (with some other Tock arguments) to run the Tock kernel in test mode. This allows us to write until tests which we can run to test the kernel or other hardware devices.

This PR currently only adds support for running on QEMU, but I have been able to run the tests on both QEMU and hardware.

The idea here is to make development, OT sha updates and other work easier by automating even more testing.

By running in the kernel loop we can run more then just simple tests and can actually run full hardware interaction tests. I have test cases working for both OTBN and HMAC (PRs will come after this PR).

We use [semihosting](https://github.com/riscv/riscv-semihosting-spec) to exit QEMU. In order to do that though we need the [exact 3 instructions](https://github.com/riscv/riscv-semihosting-spec/blob/main/riscv-semihosting-spec.adoc#11-semihosting-trap-instruction-sequence):

```asm
slli x0, x0, 0x1f       # 0x01f01013    Entry NOP
ebreak                  # 0x00100073    Break to debugger
srai x0, x0, 7          # 0x40705013    NOP encoding the semihosting call number 7
```

Using Rust's inline asm
```rust
       asm!(
          "slli x0, x0, 0x1f
          ebreak
          srai x0, x0, 7"
      );
```
incorrectly generates the compressed `9002` form of ebreak as the assembler will compress the instructions from `ebreak` to `c.ebreak`. This is the correct behaviour as the assembler should try to compress the instructions if it can.

In order to work with semihosting we include the assembly here where we are able to disable compressed instruction support. This follows the example in the RISC-V semihosting spec.

### Testing Strategy

Run `make test`

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


boards/earlgrey-nexysvideo/build.rs
```diff
@@ -1,4 +1,15 @@
...
     println!("cargo:rerun-if-changed=../kernel_layout.ld");
+
+    let out_dir = env::var("OUT_DIR").unwrap();
+
+    println!("cargo:rustc-link-search={}/../../../", out_dir);
+    println!("cargo:rerun-if-changed=libriscv32imc-unknown-none-elf.a");
+    println!(
+        "cargo:rustc-link-lib=static={}",
+        "riscv32imc-unknown-none-elf"
+    );
```
hudson-ayers, line 14: am I correct in understanding that the remaining fix needed is a way to only include these lines for test builds?

alistair23: Yep

boards/earlgrey-nexysvideo/.cargo/config
```diff
@@ -0,0 +1,2 @@
+[target.'cfg(any(target_arch = "arm", target_arch = "riscv32"))']
```
hudson-ayers, line 1: arm?

arch/rv32i/src/asm.S
```diff
@@ -0,0 +1,5 @@
+.global semihost_command
+semihost_command:
+	slli x0, x0, 0x1f       # 0x01f01013   		Entry NOP
+	.word 0x00100073 		# ebreak	    	Break to debugger
+	.word 0x40705013		# srai x0, x0, 7    NOP encoding the semihosting call number 7
```
hudson-ayers, line 5: some weird alignment here

boards/earlgrey-nexysvideo/Makefile
```diff
@@ -36,3 +36,12 @@ flash-app: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
 	$(RISC_PREFIX)-objcopy --update-section .apps=$(APP) $^ $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.elf
 	$(RISC_PREFIX)-objcopy --output-target=binary $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.elf $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.bin
 	$(OPENTITAN_TREE)/build-out/sw/host/spiflash/spiflash $(FLASHID) --input=$(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.bin
+
+test:
+	mkdir -p $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/deps
+	riscv64-linux-gnu-gcc -ggdb3 -c -mabi=ilp32 -march=rv32imc ../../arch/rv32i/src/asm.S -o $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/asm.o
```
hudson-ayers, line 42: it seems like a bit of a shame to require the ability to build the asm just to run these tests, especially since the asm should never really change. Could we just include a prebuilt .o file in boards/earlgrey-nexysvideo somewhere, and have a second target for re-generating that file from the .S file? It would be nice to add these tests to CI, without adding dependencies that can't be installed via rustup.

boards/earlgrey-nexysvideo/src/main.rs
```diff
@@ -6,6 +6,9 @@
 // Disable this attribute when documenting, as a workaround for
 // https://github.com/rust-lang/rust/issues/62184.
 #![cfg_attr(not(doc), no_main)]
+#![feature(custom_test_frameworks)]
```
hudson-ayers, line 9: Flagging this nightly feature for others to see -- nightly features in `boards/` are generally less problematic than those in other places, but we should consider that this feature will likely become a requirement for our CI to keep working if we incorporate these tests.

alistair23:
Fixed!

boards/earlgrey-nexysvideo/src/tests.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// This creates a second UART peripheral.
+// This should ONLY be used by test cases running on QEMU.
+static mut WRITER: Writer = Writer {
+    uart: earlgrey::uart::Uart::new(
+        earlgrey::uart::UART0_BASE,
+        earlgrey::chip_config::CONFIG.peripheral_freq,
+    ),
+};
```
hudson-ayers, line 40: It seems possible for there to be weird behavior in the tests if they call any code in the kernel containing debug statements, once there are two writers. I assume the issue is that you need a synchronous writer here?

hudson-ayers (commented):
This is cool, though I think the fact that it never can run the scheduler makes it pretty inherently limited compared to hardware tests or even the existing QEMU tests.

boards/earlgrey-nexysvideo/Makefile
alistair23: I have added a prebuilt test as well that users can run

alistair23: Ok, it now uses a prebuilt binary and there is a make target to rebuild the binary.

boards/earlgrey-nexysvideo/src/tests.rs
alistair23: Yes, there could be. The problem is because we don't go into the kernel loop we can't use the usual `debug!`. That actually is something we could maybe improve in the future, by maybe running the tests as part of the loop (or replacing the loop with a custom test loop)

alistair23:
Woo! Just pushed an update that now runs the tests in the main kernel loop. Hopefully now it can be used more for testing hardware devices!

kernel/src/sched.rs
```diff
@@ -524,6 +520,56 @@ impl Kernel {
         }
     }
 
+    /// Main loop of the OS.
+    ///
+    /// Most of the behavior of this loop is controlled by the `Scheduler`
+    /// implementation in use.
+    #[cfg(not(test))]
```
hudson-ayers, line 527: why is this needed? Can't the test code just not call this?

kernel/src/sched.rs
```diff
@@ -524,6 +520,56 @@ impl Kernel {
...
+        chip.watchdog().setup();
+        loop {
+            self.loop_op(platform, chip, ipc, scheduler, capability, false);
+        }
+    }
+
+    /// Main loop of the OS.
+    ///
+    /// Most of the behavior of this loop is controlled by the `Scheduler`
+    /// implementation in use.
```
hudson-ayers, line 545: This should document that this version is only used for running integration tests

hudson-ayers (commented):
This is cool, but historically modifying the core kernel for testing purposes has been unpopular -- let's see what others think

alistair23:
This is a pretty small modification and we are using the standard Rust test framework. It also doesn't result in any modifications when not running as a test.

phil-levis (commented):
We talked about this on the OT call. I'm of the opinion that testing is perhaps the *best* reason to change the kernel (if you have to change it). The way this factors this out is better than a level of indirection of policy.

hudson-ayers (approved):
LGTM, but I think we should add this as an agenda item for the core call, at least to bring attention.

boards/earlgrey-nexysvideo/.cargo/config
```diff
@@ -0,0 +1,2 @@
+[target.'cfg(target_arch = "riscv32")']
+runner = "qemu-system-riscv32 -M opentitan -bios ../../tools/qemu-runner/opentitan-boot-rom.elf -nographic -serial stdio -monitor none -semihosting -kernel "
```
ppannuto, line 2: Does this need to be in `.cargo`? As a rule, we have avoided using hidden configuration in Tock (this is one of the underlying reasons for our make-based design). Can we drive these with flags from the invocation instead?

fwiw @jrvanwhy opened an upstream issue about this a little while ago: https://github.com/rust-lang/cargo/issues/8244, but it does not seem to have gained much traction :/

arch/rv32i/src/asm.S
```diff
@@ -0,0 +1,5 @@
+.global semihost_command
+semihost_command:
+	slli x0, x0, 0x1f		# 0x01f01013   		Entry NOP
+	.word 0x00100073 		# ebreak	    	Break to debugger
+	.word 0x40705013		# srai x0, x0, 7    NOP encoding the semihosting call number 7
```
ppannuto, line 5: This confuses me a bit, we use inline asm everywhere else in Tock, and I'm reasonably confident that it's treated as an opaque block of assembly by the compiler. There was some explanation in the top-level PR description, but if this truly cannot be done with inline asm, I think a detailed explanation of why needs to be in this file here.

boards/earlgrey-nexysvideo/Makefile
```diff
@@ -36,3 +36,17 @@ flash-app: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
 	$(RISC_PREFIX)-objcopy --update-section .apps=$(APP) $^ $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.elf
 	$(RISC_PREFIX)-objcopy --output-target=binary $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.elf $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.bin
 	$(OPENTITAN_TREE)/build-out/sw/host/spiflash/spiflash $(FLASHID) --input=$(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.bin
+
+test-rebuild:
+	mkdir -p $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/deps
+	riscv64-linux-gnu-gcc -ggdb3 -c -mabi=ilp32 -march=rv32imc ../../arch/rv32i/src/asm.S -o $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/asm.o
```
ppannuto, line 42:
```suggestion
	$(RISC_PREFIX)-gcc -ggdb3 -c -mabi=ilp32 -march=rv32imc ../../arch/rv32i/src/asm.S -o $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/asm.o
```
?

boards/earlgrey-nexysvideo/Makefile
```diff
@@ -36,3 +36,17 @@ flash-app: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
...
+	ar crs $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/libriscv32imc-unknown-none-elf.a $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/asm.o
```
ppannuto, line 43:
```suggestion
	$(AR) crs $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/libriscv32imc-unknown-none-elf.a $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/asm.o
```

boards/earlgrey-nexysvideo/Cargo.toml
```diff
@@ -28,3 +28,6 @@ lowrisc = { path = "../../chips/lowrisc" }
 #      OpenTitan SoC design simulated in Verilator.
 fpga_nexysvideo = ["earlgrey/config_fpga_nexysvideo"]
 sim_verilator = ["earlgrey/config_sim_verilator"]
+# This is used to indicate that we should include the prebuilt
+# RISC-V assmebly. This is currently just used to run RISC-V tests
+asm = []
```
ppannuto, line 33: This is a bit of a generic name, "asm" could have many meanings and side-effects. Could we call this something more descriptive?

boards/earlgrey-nexysvideo/src/io.rs
```diff
@@ -1,12 +1,17 @@
...
 use kernel::debug;
 use kernel::debug::IoWrite;
+#[cfg(not(test))]
 use kernel::hil::gpio::Configure;
+#[cfg(not(test))]
 use kernel::hil::led;
 
+#[cfg(not(test))]
 use crate::CHIP;
+#[cfg(not(test))]
```
ppannuto, line 14: maybe: can we move these imports into the the `cfg` guard on line 41? Basically tie them to the `panic_fmt` where they are used rather than a whole bunch of on/offs up here?

boards/earlgrey-nexysvideo/src/main.rs
ppannuto: I think we the fairly federated CI infrastructure this is okay. We should be able to slice out this test in the future if needed relatively easily.

kernel/src/sched.rs
```diff
@@ -524,6 +520,55 @@ impl Kernel {
...
+        loop {
+            self.loop_op(platform, chip, ipc, scheduler, capability, false);
+        }
+    }
+
+    /// Main test loop of the OS.
+    ///
+    /// This is similar to `kernel_loop()` but also calls tests passed in via
+    /// the `tests` argumet. This should only be used for running tests.
+    pub fn test_kernel_loop<P: Platform, C: Chip, SC: Scheduler<C>, const NUM_PROCS: usize>(
```
ppannuto, line 545: Not that I'm one to advocate for more `cfg`s, but at the same time, this function should _only_ ever be used in a CI/test setup, right? Feels like we should indicate that.

This does also break our prior policy of only `cfg`ing arch-specific stuff. It sounds like the tradeoffs here have been discussed a fair bit by the OT-WG, and `cfg`ing test harnesses is pretty well-supported by Rust and is a generally supported notion in the ecosystem, but would like to understand whether this really needs to be a separate function, or if all we _really_ need is to expose a sleep / don't sleep parameter?

kernel/src/sched.rs
```diff
@@ -524,6 +520,55 @@ impl Kernel {
...
+        // Run all of the tests
+        for test in tests {
+            for _ in 0..20 {
+                self.loop_op(platform, chip, ipc, scheduler, capability, true);
+            }
+            test();
+            for _ in 0..100 {
+                self.loop_op(platform, chip, ipc, scheduler, capability, true);
+            }
+        }
```
ppannuto, line 564: From a design point, I don't really understand why this loop is here? In particular, the loops of 20 and 100 seem somewhat odd and arbitrary. It feels like this loop should really be the responsibility of the test harness runner.

ppannuto (changes requested):
I think I'm missing something in the overall design here, but it seems like what's needed is not a separate `kernel_loop` function, but rather the ability to pass a sleep / not-sleep parameter to the scheduler.

The `kernel_loop` method is already `pub`, and could (should?) be invoked by the test harness directly. That would then let all of the test-specific logic (the loops before/after running the `test`) be lifted out of the kernel core and into the test harness.

It would also eliminate the need to add a separate loop function (and the associated `cfg`) to the kernel core.

We may also consider adding something like an `InhibitSleepCapability` to protect this interface; though that might be redundant with the main loop capability.

boards/earlgrey-nexysvideo/.cargo/config
alistair23: Yeah, it does have to be. I couldn't get it to work anywhere else.

alistair23:
> I think I'm missing something in the overall design here, but it seems like what's needed is not a separate `kernel_loop` function, but rather the ability to pass a sleep / not-sleep parameter to the scheduler.

So we need two things. First we need a way to run the kernel loop only a number of times. This way it can complete and we can return success of failure. Secondly we want to ensure that we don't sleep, that way we can ensure that if we run out of things to do we will still complete the loop and return.

> The `kernel_loop` method is already `pub`, and could (should?) be invoked by the test harness directly. That would then let all of the test-specific logic (the loops before/after running the `test`) be lifted out of the kernel core and into the test harness.

Do you mean just move the test loop to the board and call it there? The idea of keeping it in the kernel means that all boards run the same harness (although the tests will be board specific).

arch/rv32i/src/asm.S
alistair23: It turns out that the `asm!` blocks are not treated as raw assembly.

The output of this method (linking with a block of code) results in these instructions being in the final binary:

```asm
2000cc64:       01f01013                slli    zero,zero,0x1f
2000cc68:       00100073                ebreak
2000cc6c:       40705013                srai    zero,zero,0x7
```
Which correctly matches the semihosting requirements.

While if instead if I do this in Rust code

```rust
        asm!(
            "slli x0, x0, 0x1f
            ebreak
            srai x0, x0, 7"
        );
```

The final binary contains:

```asm
20009ade:       01f01013                slli    zero,zero,0x1f
20009ae2:       9002                    ebreak
20009ae4:       40705013                srai    zero,zero,0x7
```

As you can see the compiler will compress the `ebreak` instruction, which then doesn't work with the RISC-V semihosting spec. The assembly block on the other hand can use the full hex of the instruction, which avoids any changes to the output.

alistair23: I have added a comment to the assembly file as well

kernel/src/sched.rs
alistair23: I originally had this function protected by a `#[cfg(not(test))]`, but Hudson preferred that to be removed.

I know Tock doesn't like `cfg`, but `cfg(test)` is pretty core to Rust. It's support by all of the Rust tools and used by most Rust projects so in this case I don't think we are doing anything too confusing here.

Like I mentioned in my other reply, we want to run the tests as part of the loop AND we also want the kernel loop to end, so we can then report success. That's why there is a change to the kernel loop.

kernel/src/sched.rs
alistair23: The numbers are arbitrary. As long as the loop runs enough times that the test can complete and all debug output is printed (but not an unlimited number of times) it doesn't really matter. We could move this to the test harness (in the board), but this should be the same across boards and this way we aren't exposing the internal kernel loop logic to boads

hudson-ayers: It does seem to be the case that different tests might require different numbers of iterations to complete, and once some longer running tests are included it will make sense to not make every test wait the length of time required by the longest test. Though maybe none of our tests will ever be long enough for that to matter

kernel/src/sched.rs
```diff
@@ -454,49 +454,45 @@ impl Kernel {
...
-                                        scheduler,
-                                        process,
-                                        ipc,
-                                        timeslice_us,
-                                    );
-                                    scheduler.result(reason, time_executed);
-                                });
-                            }
-                            SchedulingDecision::TrySleep => {
+        no_sleep: bool,
```
bradjc, line 464:
```suggestion
```

kernel/src/sched.rs
```diff
@@ -454,49 +454,45 @@ impl Kernel {
...
+                                    scheduler,
+                                    process,
+                                    ipc,
+                                    timeslice_us,
+                                );
+                                scheduler.result(reason, time_executed);
+                            });
+                        }
+                        SchedulingDecision::TrySleep => {
+                            if !no_sleep {
```
bradjc, line 495:
```suggestion
                            // For testing, it may be helpful to
                            // disable sleeping the chip in case
                            // the running test does not generate
                            // any interrupts.
                            if !no_sleep {
```

kernel/src/sched.rs
```diff
@@ -524,6 +520,55 @@ impl Kernel {
...
+        &self,
+        platform: &P,
+        chip: &C,
+        ipc: Option<&ipc::IPC<NUM_PROCS>>,
+        scheduler: &SC,
+        capability: &dyn capabilities::MainLoopCapability,
+    ) -> ! {
+        chip.watchdog().setup();
+        loop {
+            self.loop_op(platform, chip, ipc, scheduler, capability, false);
```
bradjc, line 537:
```suggestion
            self.kernel_loop_operation(platform, chip, ipc, scheduler, false, capability);
```

kernel/src/sched.rs
bradjc: I think people hacking around might want to be able to run the test version without having to go through layers of rust testing cfgs and infrastructure.

kernel/src/sched.rs
```diff
@@ -524,6 +520,55 @@ impl Kernel {
...
+        ipc: Option<&ipc::IPC<NUM_PROCS>>,
+        scheduler: &SC,
+        capability: &dyn capabilities::MainLoopCapability,
+        tests: &[&dyn Fn()],
+    ) {
+        chip.watchdog().setup();
+        // Run all of the tests
+        for test in tests {
+            for _ in 0..20 {
+                self.loop_op(platform, chip, ipc, scheduler, capability, true);
```
bradjc, line 558:
```suggestion
                self.kernel_loop_operation(platform, chip, ipc, scheduler, true, capability);
```

kernel/src/sched.rs
```diff
@@ -524,6 +520,55 @@ impl Kernel {
...
+            }
+            test();
+            for _ in 0..100 {
+                self.loop_op(platform, chip, ipc, scheduler, capability, true);
```
bradjc, line 562:
```suggestion
                self.kernel_loop_operation(platform, chip, ipc, scheduler, true, capability);
```

kernel/src/sched.rs
```diff
@@ -524,6 +520,55 @@ impl Kernel {
...
+            }
+        }
+
+        // Run the loop a few more times to see all the messaages
+        for _ in 0..100 {
+            self.loop_op(platform, chip, ipc, scheduler, capability, true);
```
bradjc, line 568:
```suggestion
            self.kernel_loop_operation(platform, chip, ipc, scheduler, true, capability);
```

kernel/src/sched.rs
```diff
@@ -524,6 +520,55 @@ impl Kernel {
...
+        &self,
+        platform: &P,
+        chip: &C,
+        ipc: Option<&ipc::IPC<NUM_PROCS>>,
+        scheduler: &SC,
+        capability: &dyn capabilities::MainLoopCapability,
+        tests: &[&dyn Fn()],
+    ) {
+        chip.watchdog().setup();
+        // Run all of the tests
```
bradjc, line 555:
```suggestion
        // Run all of the tests.
        //
        // The kernel loop work function is run in a loop which should
        // be long enough to complete each test. However, this has
        // not been robustly verified, and the loop iteration counts
        // may have to be adjusted for newer or more complex tests.
```

boards/earlgrey-nexysvideo/src/main.rs
```diff
@@ -404,12 +411,57 @@ pub unsafe fn main() {
...
+}
+
+/// Main function.
+///
+/// This function is called from the arch crate after some very basic RISC-V
+/// setup and RAM initialization.
+#[no_mangle]
+pub unsafe fn main() {
+    #[cfg(test)]
+    test_main();
```
bradjc, line 424: Where does this come from and what code is actually running?? This is why cfgs/macros can be so hard for new users. I'm searching for "fn test_main" in the changed files and not finding anything. This normally always works in Rust because `fn` has to be the last keyword before the function name. But either I'm missing something or rust testing breaks that primitive.

bradjc:
I don't understand the need for the assembly file. Why is it needed for testing, and why do we need an external .S file when we haven't needed to resort to that complexity otherwise?

Also the .cargo config file is a MAJOR bummer. But if we can't hack it with the makefile then I don't know what we can do.

arch/rv32i/src/asm.S
ppannuto: Interesting... ARM opcodes resolve the compression of ambiguous instructions with option width specifiers (`mov.n r0 r0 r0` is 16-bit, `mov.w r0 r0 r0` is 32-bit; narrow/wide).

It looks like RISCV sort of does this; the `9002` instruction is more precisely `c.ebreak` (compressed ebreak).

Unfortunately, I can't find an equivalent instruction in RISCV to force the wider instruction choice.

Digging into the RISCV manual, I think the comment as-written is incorrect. If a machine supports compressed instructions, then the assembler is _supposed_ replace `ebreak` with `c.ebreak`. IIUC, the real reason this is working is because the machine model that's passed to the assembler line is _different_ than the true underlying machine model, and claims the architecture doesn't support compression. If I've got this all right, this is pretty subtle and a bit of a (necessary? :/) hack. We should highlight the significance of the "incorrect" machine model given to the assembler in some way though.

It would also be good to verify that there's no way to simply ask the assembler to emit the preferred form of `ebreak`. Following this:
image
I tried `g.ebreak`, but that got rejected by the assembler on my machine :/

alistair23: I have updated the comment to try and be more clear that we are disabling compressed instructions as we need the uncompressed version.

boards/earlgrey-nexysvideo/src/main.rs
alistair23: This comes from the `reexport_test_harness_main` at the top of the file. There are some more details here: https://os.phil-opp.com/testing/#custom-test-frameworks

It seems to be part of Rust's test infrastructure.

alistair23:
I just pushed an update that has the test case call the kernel operations instead of a test specific loop in the kernel.

This means we don't need changes to the kernel loop, we still need to expose the loop and the sleep disable option though. The other main advantage here is that tests can now check if an event has occurred which should help to catch callbacks that don't occur.

ppannuto (approved):
Thanks for pulling the loop out, I think the changes to the core sched.rs are a lot cleaner now.

The revised explanation of the assembly looks good too. Thanks for bearing with me on that one. I may ping the RISC-V mailing list once to see if there's a way to force the uncompressed version to be emitted. It would be nice to get rid of the external assembly file and extra tool invocation/linking if we can, but that doesn't need to stand in the way of this PR right now. 

I can see where some of what's in the boards/nexys/tests might someday in the future get pulled into a more generic, used-by-all-boards system, but I think our process of experimenting with new stuff in board-specific is good. It also mitigates some of the curmudgeonly-ness that I share with @bradjc regarding the magic-ness of the Rust testing framework, but I think that's probably a me problem :).

ppannuto:
Well, I asked the RISC-V folks: https://groups.google.com/u/1/a/groups.riscv.org/g/isa-dev/c/XKkYacERM04 -- maybe we'll be able to ditch the dedicated assembly file.

ppannuto:
Great. The suggestion from Palmer does work, and we can get rid of the separate assembly. I've confirmed with these patches that the resulting object file includes the assembly verbatim as desired:

    20008472: 13 10 f0 01   slli    zero, zero, 31
    20008476: 73 00 10 00   ebreak
    2000847a: 13 50 70 40   srai    zero, zero, 7

It may be possible to further simplify the build step to do less, but I'm not 100% confident in what all of the semihosting separate build is doing.

hudson-ayers:
@ppannuto that commit seems to break CI

ppannuto:
Oops, fixed, sorry.

hudson-ayers (approved):

boards/earlgrey-nexysvideo/Makefile
```diff
@@ -36,3 +36,14 @@ flash-app: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
 	$(RISC_PREFIX)-objcopy --update-section .apps=$(APP) $^ $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.elf
 	$(RISC_PREFIX)-objcopy --output-target=binary $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.elf $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.bin
 	$(OPENTITAN_TREE)/build-out/sw/host/spiflash/spiflash $(FLASHID) --input=$(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.bin
+
+test-rebuild:
+	mkdir -p $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/deps
+	cp $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/libriscv32imc-unknown-none-elf.a ./
```
bradjc, line 42: We don't need this anymore?

alistair23:
Ah cool! Thanks for that Pat. I didn't realise we could do that in the rust `asm`. I just pushed an update that should remove all of the infrastructure for including the binary (I squashed it into your commit).

phil-levis (commented):
I think this is a *much* better decomposition. It gives tests the ability to control invocation of the kernel loop, which allows them to more clearly specify their expectations. I'm holding off on Approve until I do a more detailed read of the code, but this seems like the right architecture.

bradjc (dismissed):

hudson-ayers (dismissed):

arch/rv32i/src/lib.rs
```diff
@@ -425,6 +425,39 @@ pub extern "C" fn _start_trap() {
...
+/// for more details on the three insturctions.
+///
+/// In order to work with semihosting we include the assembly here
+/// where we are able to disable compressed instruction support. This
+/// follows the example used in the Linux kernel:
+/// https://elixir.bootlin.com/linux/v5.12.10/source/arch/riscv/include/asm/jump_label.h#L21
+/// as suggested by the RISC-V developers:
+/// https://groups.google.com/a/groups.riscv.org/g/isa-dev/c/XKkYacERM04/m/CdpOcqtRAgAJ
+#[cfg(all(target_arch = "riscv32", target_os = "none"))]
+pub unsafe fn semihost_command(_command: usize, _arg0: usize, _arg1: usize) -> ! {
```
lschuermann, line 440: I don't quite understand this part. The parameters don't seem to be used so I guess this will always issue the particular semihosting command to exit the emulation?

In that case, I'd vouch to remove the arguments and make the function specific for exiting emulation.

Otherwise, if this is supposed to be a generic semihosting interface, I don't think we can / should mark this function as non-returning, because presumably some semihosting commands will continue execution at some point? In short, I see a mismatch between having the generic arguments, and the return value being specific for one particular semihosting command.

boards/earlgrey-nexysvideo/src/io.rs
```diff
@@ -56,3 +61,14 @@ pub unsafe extern "C" fn panic_fmt(pi: &PanicInfo) -> ! {
...
+
+#[cfg(test)]
+#[no_mangle]
+#[panic_handler]
+pub unsafe extern "C" fn panic_fmt(pi: &PanicInfo) -> ! {
+    let writer = &mut WRITER;
+
+    let _ = write!(writer, "{}", pi);
+    // Exit QEMU with a return code of 1
+    rv32i::semihost_command(0x18, 1, 0);
```
lschuermann, line 73: AFAIK earlgrey-nexysvideo also runs on an FPGA board. How does this behave here?

lschuermann (commented):
Looks good in general. Some questions though.

arch/rv32i/src/lib.rs
lschuermann: I see that there are indeed some arguments passed into this function. If these are used because they're loaded into registers, we definitely want to use a defined (C) ABI here. If this works because the arguments are currently loaded in the correct registers, but it's not marked `extern "C"`, that is probably pure coincidence.

alistair23: Good catch, the `extern "C"` was lost in the conversion to inline asm and it continued to work. I have added it back.

I also changed `semihost_command` to not be `-> !` as you are right, some commands could return. I then added non returning helper functions for the exit commands.

lschuermann: Great, this looks much better :+1:. I still think we need to remove the `noreturn` `asm!` option, otherwise this looks good to me.

hudson-ayers (approved):

bradjc (approved):

PR #2601, opened 2021-05-28
Change qemu-path to relative path.

jack007xx:
### Pull Request Overview

This pull request changes qemu-path to relative path.
`make ci-setup-qemu` doesn't set `qemu-system-riscv32` to PATH.
So, define path/to/qemu as a var, and use it.

### Testing Strategy

### TODO or Help Wanted

Or, is there some reason using local `qemu-system-riscv32` ?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/earlgrey-nexysvideo/Makefile
+++ b/boards/earlgrey-nexysvideo/Makefile
@@ -5,7 +5,7 @@ TARGET=riscv32imc-unknown-none-elf
 PLATFORM=earlgrey-nexysvideo
 FLASHID=--dev-id="0403:6010"
 RISC_PREFIX = riscv64-elf
-QEMU=../../tools/qemu/build/qemu-system-riscv32
+QEMU ?= ../../tools/qemu/build/qemu-system-riscv32
 
 
 include ../Makefile.common
--- a/boards/earlgrey-nexysvideo/Makefile
+++ b/boards/earlgrey-nexysvideo/Makefile
@@ -5,6 +5,7 @@ TARGET=riscv32imc-unknown-none-elf
 PLATFORM=earlgrey-nexysvideo
 FLASHID=--dev-id="0403:6010"
 RISC_PREFIX = riscv64-elf
+QEMU=../../tools/qemu/build/qemu-system-riscv32
 
 
 include ../Makefile.common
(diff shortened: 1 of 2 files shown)


bradjc (approved):

lschuermann (changes requested):
As far as I understand it, this will (more or less) force you to use the shipped QEMU version. Many distributions do ship current versions of QEMU however, which at least for the `hifive1` board will work just fine. Furthermore, they are prebuilt and in general have less trust issues as they're built by the distribution vendor.

I think we should either allow to optionally use the "shipped" QEMU, or just make people add it to their path.

jack007xx:
If that's the case, it might be better to change the README.

The question is which one should be the default.
I thought it would be better to use relative paths for the make file, as it is difficult to understand for beginners.
(Of course, it should be mentioned in the README that you can use the original QEMU by rewriting the variables.)

lschuermann:
Shotaro Tsuboi ***@***.***> writes:
> If that's the case, it might be better to change the README.
>
> The question is which one should be the default.
> I thought it would be better to use relative paths for the make file,
> as it is difficult to understand for beginners.

Well, given the earlgrey-nexysvideo board's README already states that
the QEMU version shipped in Tock is really just there such that
distributions shipping older versions can work as well, I don't think we
should make people go through setting options just to use their system's
QEMU:

> Unfortunately you need QEMU 5.2, which at the time of writing is
> unlikely to be avaliable in your distro. Luckily Tock can build QEMU
> for you. [...]

> (Of course, it should be mentioned in the README that you can use the
> original QEMU by rewriting the variables.)

I agree it's good to just document the additional required step to add
it to PATH. If you want you can probably add some Makefile-logic to make
the used QEMU overridable, but I don't think that's worth the trouble
given this issue is going to get better over time: for instance, Ubuntu
21.04 already ships QEMU 5.2 [1].

[1]: https://packages.ubuntu.com/hirsute/qemu

alistair23:
The problem though is that we keep needing newer versions (we actually need version 6.0 now, I'll update that). As the OT bitstreams change the QEMU model has to change as well. Eventually it will stabalise though

bradjc:
If we need a very specific version (presumably to include the latest OT patches) then it seems easier to default to using the tock-provided version. Someone who wants to try a different version can override the `QEMU` variable.

boards/earlgrey-nexysvideo/Makefile
```diff
@@ -5,6 +5,7 @@ TARGET=riscv32imc-unknown-none-elf
 PLATFORM=earlgrey-nexysvideo
 FLASHID=--dev-id="0403:6010"
 RISC_PREFIX = riscv64-elf
+QEMU=../../tools/qemu/build/qemu-system-riscv32
```
alistair23, line 8: Can we use `?=` so this can be overridden via the command line

lschuermann:
> If we need a very specific version (presumably to include the latest OT patches) then it seems easier to default to using the tock-provided version.

Fair. What should we do about the HiFive board then? I haven't had any issued with QEMU v5.2, which is already pretty common across distributions. Should we always hint users at building their own QEMU or decide on a board-by-board basis, given the assumptions the Tock code makes about the device model?

alistair23:
Good point. I would drop the change from the HiFive board

alistair23 (approved):

lschuermann (approved):

lschuermann:
Thanks @jack007xx!

PR #2605, opened 2021-06-01
allow size printing tool to be used on non-arm architectures

hudson-ayers:
### Pull Request Overview

This pull request removes the restriction that the size tool only be used on arm architectures. As far as I can tell, this restriction only existed because the tool used to use `arm-none-eabi-objdump`. Now that the tool uses `llvm-objdump` instead, there is no need for an architecture restriction.

It also updates the code size benchmarking tool to no longer ignore riscv boards, as that is no longer necessary. 

### Testing Strategy

This pull request was tested by running the tool on earlgrey-nexysvideo.elf and observing that the output looks sensible.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/tools/github_actions_size_changes.sh
+++ b/tools/github_actions_size_changes.sh
@@ -13,7 +13,7 @@ set -e
 
 # Bench the current commit that was pushed. Requires navigating back to build directory
 make allboards > /dev/null 2>&1
-for elf in $(find . -maxdepth 8 | grep 'release' | egrep '\.elf$' | grep -v 'riscv'); do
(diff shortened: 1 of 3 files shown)


hudson-ayers:
The benchmarks PR only fails on this because when it checks out the master repo, the old version of the repo does not benchmark the riscv architectures, so the comparison fails.

lschuermann (approved):
Confirmed the tool works for me with Nix's `llvm-objdump` on RISC-V boards and the output seems good.

bradjc (approved):

alistair23 (approved):

PR #2606, opened 2021-06-01
build with codegen-units = 1 to reduce code size

hudson-ayers:
### Pull Request Overview

This pull request modifies `Cargo.toml` to use `codegen-units = 1` for release builds. By default, Cargo uses 16 codegen units for each crate and LLVM processes them in parallel. This speeds up compilation but prevents LLVM from applying some optimizations which have to look at all of the units at once.

For Imix, adding this option reduces code size by 1.5 kB at the expense of compilation taking about 2 seconds longer.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,3 +86,4 @@ panic = "abort"
 lto = true
 opt-level = "z"
 debug = true
+codegen-units = 1


ppannuto (approved):
This seems like a nice and easy win; any reason to not just merge now?

lschuermann (approved):
I did not get to test the impacts of this thoroughly yet, though I agree with @ppannuto. It's an official feature, I don't see what could go wrong. Also there seems to be a consistent decrease in size across all boards.

alistair23 (approved):

PR #2610, opened 2021-06-04
kernel: Add a shared allow syscall

alistair23:
### Pull Request Overview

This is a look at what a shared allow syscall would look like. This is based on https://github.com/tock/tock/pull/2590#issuecomment-854259226

Some of the TRD update is borrowed from: https://github.com/tock/tock/blob/trd104-buffer-reading/doc/reference/trd104-syscalls.md

The idea of these syscalls is to handle capsules that expect userspce to access a buffer while the kernel has access. [ROS](https://github.com/tock/tock/pull/2381) would use this (it needs to be rebased once this is merged) and there are some other users.

Add a shared allow that is similar to read/write allow except that the
app is expected to be able to access the buffer.

This is different to `allow_readwrite()` in that the app is allowed
to read/write the buffer once it has been passed to the kernel.
For more details on how this can be done safely see TRD104.

### Testing Strategy

None

### TODO or Help Wanted

This needs to be tested, TRD updated and libtock-c supported added.

The other question is do we allow userspace write access or is it only read access?

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/doc/reference/trd104-syscalls.md
+++ b/doc/reference/trd104-syscalls.md
@@ -559,7 +559,7 @@ Read-Write Allow to reclaim the buffer, then call Read-Write Allow
 again to re-allow it with a different size. If userspace passes
 an overlapping buffer, the kernel MUST return a failure result with
 an error code of `INVALID`.
- 
+
 4.5 Read-Only Allow (Class ID: 4)
 ---------------------------------
 
@@ -689,6 +689,50 @@ The definition of these status codes is outside the scope of this document.
 If an exit syscall is successful, it does not return. Therefore, the return
 value of an exit syscall is always `Failure`. `exit-restart` and 
 `exit-terminate` MUST always succeed and so never return. 
+
+4.8 Shared Allow (Class ID: 7)
+---------------------------------
+The shared allow syscall follows the same expectations and requirements
+as described for the Read-Write syscall in section 4.4, with the exception
+that apps are explicity allowed to read/write buffers that have been passed
+to the kernel.
+
+The standard calling pattern for reading data from the Tock kernel is to
+  1. use `subscribe` to register a callback,
+  2. make a Read-Write Allow call to share a buffer with the kernel,
+  3. call a `command` to start an operation that writes the allowed buffer,
+  4. in the upcall that signals the operation completes, make another
+  Read-Write Allow call to reclaim the buffer shared with the kernel.
+
+In normal use, userspace does not access a buffer that
+has been shared with the kernel with a Read-Write Allow call. This
+reading restriction is because the contents of the buffer may be in an
+intermediate state and so not consistent with expected data
+models. Ensuring every system call driver maintains consistency in the
+presence of arbitrary userspace reads is too great a programming
+burden for an unintended use case.
+
+However, there can be cases when it is necessary for userspace to be
+able to read a buffer without first revoking it from the kernel with a
+Read-Write Allow. These cases are situations when the cost of a
+Read-Write Allow system call is an unacceptable overhead for
+accessing the data.
+
+This allows capsules that support the shared allow call to communicte with
+applications without buffers needing to be re-allowed. For example a capsule
+might want to report statistics to a userspace app. It could do this by letting
+the app perform a shared allow call to allocate a buffer. Then the capsule can
+write statistics to the buffer and at any time the app can read the statistics
+from the buffer.
+
+As simultaneous accesses to a buffer from both userspace and the kernel can
+cause issues if not implemented properly, each capsule using the Shared Allow
+mechanism MUST document, in a Draft or Final Documentary TRD, how consistency
+of the data exchanged through this shared memory region is achieved.
+Examples for mechanisms to ensure consistent data reads/writes can be driver state
+machines, precisely specifying when writes will be issued to the buffer by the
+kernel, or monotonically increasing counters along the data for userspace to ensure
+a read operation yielded consistent data.
 
 5 libtock-c Userspace Library Methods
 =================================
--- a/kernel/src/driver.rs
+++ b/kernel/src/driver.rs
@@ -16,6 +16,9 @@
 //!   * `allow_readwrite` provides the driver read-write access to an
 //!   application buffer.
 //!
+//!   * `allow_shared` provides the driver read-write access to an
+//!   application buffer that is still shared with the app.
+//!
 //!   * `allow_readonly` provides the driver read-only access to an
 //!   application buffer.
 //!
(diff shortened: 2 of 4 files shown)


phil-levis:
I'd like to factor this out from TRD104. We're basically blocked on that for releasing 2.0. This idea of allowing access to shared buffers was raised pretty recently. After all of our back and forth, it seems pretty clear to me that it's not a simple problem. I'd much rather release 2.0 and then have this as work-in-progress in master.

alistair23:
We don't have to block 2.0 on this as it is backwards compatible (we are only adding a syscall). We also aren't that close to 2.0 yet though, as there are still at least a few capsules that don't work (https://github.com/tock/tock/pull/2586 for example isn't merged).

In saying that it seems useful to have. This is conceptually pretty simple. https://github.com/tock/tock/pull/2590 was trying to achieve the same thing, but this method makes it cleaner for Rust userspace. This wasn't really raised that recently as the ROS PR has been open since January. Granted I probably should have caught this sooner, but there is a lot in the 2.0 changes.

The back and forth so far is about how to expose this to userspace so that userspace always sees valid data. I would say we can offload that to be capsule specific, via state machine or some other mechanism. Otherwise I think everyone is in general agreement here that it's something worth having, maybe I misread the room though.

PS: What I'm not clear on though is if we should allow apps to write to buffers once given to the kernel. I don't see a reason to restrict it, unless anyone else does.

phil-levis:
I am not super-happy with the lack of type-checking, but it looks like only lowrisc supports this trait, I think we can punt on it for now. If it turns out we start having drivers that only support some subset of the widths we can revisit it; @alistair23 's point about additional code is very valid.

phil-levis (dismissed):

doc/reference/trd104-syscalls.md
```diff
@@ -690,6 +690,41 @@ If an exit syscall is successful, it does not return. Therefore, the return
 value of an exit syscall is always `Failure`. `exit-restart` and 
 `exit-terminate` MUST always succeed and so never return. 
 
+4.8 Shared Allow (Class ID: 7)
+---------------------------------
+The shared allow syscall follows the same expectations and requirements
+as described for the Read-Write syscall in section 4.4. With the exception
```
lschuermann, line 696:
```suggestion
as described for the Read-Write syscall in section 4.4, with the exception
```

doc/reference/trd104-syscalls.md
```diff
@@ -690,6 +690,41 @@ If an exit syscall is successful, it does not return. Therefore, the return
...
+Read-Write Allow system call is an unacceptable overhead for
+accessing the data.
+
+This allows capsules that support the shared allow call to communicte with
+applications without buffers needing to be re-allowed. For example a capsule
+might want to report statistics to a userspace app. It could do this by letting
+the app perform a shared allow call to allocate a buffer. Then the capsule can
+write statistics to the buffer and at any time the app can read the statistics
+from the buffer.
+
```
lschuermann, line 727:
```suggestion

As simultaneous accesses to a buffer from both userspace and the kernel can
cause issues if not implemented properly, each capsule using the Shared Allow
mechanism MUST document, in a Draft or Final Documentary TRD, how consistency
of the data exchanged through this shared memory region is achieved.
Examples for mechanisms to ensure consistent data reads/writes can be driver state
machines, precisely specifying when writes will be issued to the buffer by the
kernel, or monotonically increasing counters along the data for userspace to ensure
a read operation yielded consistent data.
```

lschuermann (changes requested):
In general, I agree that adding a _Shared Allow_ system call makes sense. It's a good tradeoff between making the simple case simple (_Read/Write Allow_), but still allowing careful optimizations, if these are warranted & justified.

As discussed on the call of 2021-06-04, I'm still hesitant to just introduce this mechanism without any constraints. I like the basic requirement of explicitly & formally documenting any usage of this advanced mechanism (as proposed in #2590), for instance as a TRD. This is because with a _Shared Allow_ we'd be explicitly putting the burden to ensure data consistency on the user. Having the requirement of formal specification, at the very least, requires to put care into the design of the data consistency mechanism and can be easier to understand than the proper implementation. Naturally, this is only enforceable on upstream capsules.

lschuermann:
Re @alexandruradovici's question whether we also require a _Read-Only Shared Allow_ mechanism:
tl;dr: I'm not convinced this is required for now.

One of the primary motivations for featuring a _Read-Only Allow_ in the first place was the ability to share non-writeable memory regions with the kernel (e.g. flash). Having a flash-backed _Read-Only Shared Allow_ to me does not make much sense, as the whole point of the _Shared Allow_ concept is to have (at least) one party _write_ to the buffer while the other still has access. With flash, that wouldn't be possible anyways.

Now, there's of course also the benefit that with a _Read-Only Allow_ you're guaranteed that a capsule won't be able to write to allowed memory. Therefore, the mechanism _might_ be useful for, say, an application sending packets over a network interface. This is already possible through a _Read-Write Shared Allow_, minus the kernel-given guarantee that a capsule can't write to the allowed memory region. However, given that a _Shared Allow_-capsule must have explicitly & formally defined semantics as to when and how it is able to access the buffer, I think that this is not a pressing issue.

alexandruradovici:
> Re @alexandruradovici's question whether we also require a _Read-Only Shared Allow_ mechanism:
> tl;dr: I'm not convinced this is required for now.
> 
> One of the primary motivations for featuring a _Read-Only Allow_ in the first place was the ability to share non-writeable memory regions with the kernel (e.g. flash). Having a flash-backed _Read-Only Shared Allow_ to me does not make much sense, as the whole point of the _Shared Allow_ concept is to have (at least) one party _write_ to the buffer while the other still has access. With flash, that wouldn't be possible anyways.
> 
> Now, there's of course also the benefit that with a _Read-Only Allow_ you're guaranteed that a capsule won't be able to write to allowed memory. Therefore, the mechanism _might_ be useful for, say, an application sending packets over a network interface. This is already possible through a _Read-Write Shared Allow_, minus the kernel-given guarantee that a capsule can't write to the allowed memory region. However, given that a _Shared Allow_-capsule must have explicitly & formally defined semantics as to when and how it is able to access the buffer, I think that this is not a pressing issue.

I understand your point and it makes sense. Another use case for ReadOnly Shared Allow is for instance printing. An application might allow (read only) a larger buffer and then send commands to the driver to print part of them. I am not sure if printf actually does this right now. At least this is how I would have done it when I started out with Tock.

I agree with you though that it is not a pressing issue.

phil-levis:
I think overall this looks pretty reasonable. I think it would make sense to couple the Shared Allow TRD with the use cases we need. That way, we're not juggling two separate documents, and it would also give examples.

I do think we want to distinguish the three cases, though: User Read/Kernel Write (Alistair's use cases), User Write/Kernel Read (packet buffers have been discussed as a possible need) and User Write/Kernel Write.

I think there is strong evidence that some applications need the first of these. The second and third I am less convinced of. The third is especially tricky, it's basically shared memory. So I'd personally prefer narrowly tailoring a new TRD for the first case, using timestamps and fast-yield as two example implementations/use cases. Then, if concrete use cases where the second and third cases are *needed* come up, we can design based on those experiences.

alistair23:
Ok, it seems like everyone is generally on board with this then. I am closing this PR and have updated https://github.com/tock/tock/pull/2381 to add the shared allow

PR #2614, opened 2021-06-07
OpenTitan: Add SHA support

alistair23:
### Pull Request Overview

This PR adds SHA support to capsules and OpenTitan. It also allows full SHA and HMAC virtualisation

### Testing Strategy

Run 1 app with SHA operations and another app using HMAC at the same time on the FPGA.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/virtual_sha.rs
+++ b/capsules/src/virtual_sha.rs
@@ -1,0 +1,178 @@
+//! Virtualize the SHA interface to enable multiple users of an underlying
+//! SHA hardware peripheral.
+
+use core::cell::Cell;
+use kernel::common::cells::OptionalCell;
+use kernel::common::leasable_buffer::LeasableBuffer;
+use kernel::common::{ListLink, ListNode};
+use kernel::hil::digest;
+use kernel::ErrorCode;
+
+pub struct VirtualMuxSha<'a, A: digest::Digest<'a, L>, const L: usize> {
+    mux: &'a MuxSha<'a, A, L>,
+    next: ListLink<'a, VirtualMuxSha<'a, A, L>>,
+    client: OptionalCell<&'a dyn digest::Client<'a, L>>,
+    id: u32,
+}
+
+impl<'a, A: digest::Digest<'a, L>, const L: usize> ListNode<'a, VirtualMuxSha<'a, A, L>>
+    for VirtualMuxSha<'a, A, L>
+{
+    fn next(&self) -> &'a ListLink<VirtualMuxSha<'a, A, L>> {
+        &self.next
+    }
+}
+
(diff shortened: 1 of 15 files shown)


chips/lowrisc/src/hmac.rs
```diff
@@ -293,3 +290,28 @@ impl hil::digest::HMACSha512 for Hmac<'_> {
...
+        // Ensure the SHA is setup
+        regs.cfg.write(
+            CFG::HMAC_EN::CLEAR + CFG::SHA_EN::SET + CFG::ENDIAN_SWAP::SET + CFG::DIGEST_SWAP::SET,
+        );
+
+        Ok(())
+    }
+}
+
+impl hil::digest::Sha384 for Hmac<'_> {
```
phil-levis, line 307: Do we eventually expected SHA384 and SHA512 support? If not, perhaps the virtualizer should not include them for now, since this is the only implementation.

alistair23: I do expect SHA384 and SHA512 in the future

capsules/src/virtual_sha.rs
```diff
@@ -0,0 +1,178 @@
...
+        } else {
+            Err(ErrorCode::BUSY)
+        }
+    }
+}
+
+impl<'a, A: digest::Digest<'a, L> + digest::Sha384, const L: usize> digest::Sha384
+    for VirtualMuxSha<'a, A, L>
+{
+    fn set_mode_sha384(&self) -> Result<(), ErrorCode> {
```
phil-levis, line 131: I don't think this is right: it's not virtualizing, it's exposing the underlying mux and failing if the current client can't gain control. This is the exact opposite of virtualization? The configuration needs to be stored in the VirtualMuxSha then applied when this particular virtual instance is allowed to perform an operation.

But more generally, how are requests queued? This is a key element of virtualization.

phil-levis (changes requested):

alistair23:
I have updated the virtulisation to queue events

bradjc:
@phil-levis

phil-levis (approved):

alistair23:
I just rebased this on master

alistair23:
This is also required to HMAC to work as part of the 2.0 release, it would be good to get this in soon.

bradjc (approved):

lschuermann:
For the future, can we try to make the PR title / description better represent what's really changed or added?

I think "OpenTitan: Add SHA support" is not necessarily incorrect, although really understates what's introduced here: adding a SHA driver, virtualizer and quite significant changes to HMAC. Probably the fault's on me for judging based on the title whether I can afford to do a review in my (unfortunately) currently limited time, but I would've liked to a closer look at a SHA userspace driver.

Issue #2344, opened 2021-01-11
Tracking: LoRa Support


bradjc:
Tock needs LoRa support, and this issue is a marker that we would like to see support for a stack.

Generally there are two versions:

1. Implement a full LoRaWAN stack in Tock (in kernel or userspace). #1760 is a start on this.
2. Use an external chip with a full LoRaWAN stack already, and send it AT commands.

bradjc:
Another LoRa driver is here: #2615

bradjc:
We merged #3360

alistair23:
I think this is now fully supported with: https://github.com/tock/tock/tree/master/boards/apollo3/lora_things_plus#radiolib

alistair23:
If anyone is interested this is another promising development environment

https://wiki.seeedstudio.com/Wio-WM1110_Dev_Kit_Hardware_Overview/

It's the nRF52840, which [Tock supports](https://github.com/tock/tock/blob/master/boards/nordic/nrf52840dk/README.md) with a LR1110. Unfortunately RadioLib doesn't support the [LR1110 yet](https://github.com/jgromes/RadioLib/issues/679)

Seeed have announced a product based on the development kit as well: https://www.kickstarter.com/projects/seeed/sensecap-t1000-track-what-you-care-with-no-boundaries

brghena:
Wow, this board has a lot of features for pretty cheap: https://www.seeedstudio.com/Wio-WM1110-Dev-Kit-p-5677.html

 * LoRaWAN
 * GPS
 * Antennas for LoRa and GPS
 * WiFi AP Scanning (detects nearby AP MAC addresses for positioning)
 * nRF52840 (so BLE/Thread)
 * Temp/Humidity and Accelerometer over I2C
 * AA battery holder AND a solar panel

bradjc:
> Wow, this board has a lot of features for pretty cheap: https://www.seeedstudio.com/Wio-WM1110-Dev-Kit-p-5677.html

Wow! Is this the future of Tock boards??

The issue is how to program it. Seems like by default USB is attached to UART (yuck):

<img width="890" alt="image" src="https://github.com/tock/tock/assets/1467890/865058ce-2782-4b0a-a258-564b1c41abd1">

We _can_ work with that, but we need some way to install the bootloader the first time.

bradjc:
Also, it doesn't seem to have actual GPS, but it can save GPS info which can be later processed to determine location.

PR #2615, opened 2021-06-08
Lora driver first steps

oliviaweng:
### Context 

This PR is a first attempt at building a LoRa driver in Tock. It is very much a work-in-progress and has a lot of comments/commented out code on unformed thoughts/ideas as to where further development might go, so apologies for the messiness. Nevertheless, let me explain what this LoRa driver is and currently does.

Since Tock currently has no support for LoRa, I start with assuming that I have a peripheral board that has a SX12xx LoRa transceiver and can use the LoRa-MAC-in-C library. I then have a separate controller board, which is the one running Tock that will communicate LoRa commands specified by user syscalls over I2C to the peripheral board, which is running an LMIC application sketch that is set up to receive the commands over I2C and make the relevant LMIC API calls to the actual LoRa radio. This setup is depicted here:

[hw-overview.pdf](https://github.com/tock/tock/files/6617187/hw-overview.pdf)

The far-off goal for this LoRa driver is that it will have multiplexing support such that it can support multiple user applications each potentially on their own LoRa network. With this in mind, here is a rough outline of what this might look like in Tock:

[sw-overview.pdf](https://github.com/tock/tock/files/6617248/sw-overview.pdf)

The blue box is the scope of this project. I've built a toy user application and the start of a LoRa syscall library within libtock-c, but that should probably be in a PR for that repo.

### What has been implemented so far  

I've started development on this driver for the Nordic nrf52840dk board. This PR demonstrates a working end-to-end example of receiving a `command()` to send some bytes over LoRa and then writing those bytes to the I2C bus that the peripheral board will then receive and send over the LoRa network it is connected to. I assume that the peripheral LoRa board handles its own network connection for now. 

With regards to the figure above depicting the new LoRa driver capsules, I have implemented the minimum necessary for the LoRa (syscall controller) capsule and the LMIC I2C capsule to achieve this working example. I haven't added the Virtual LMIC capsule yet. I'm hoping that once the LoRa syscall capsule and LMIC I2C are more complete (have more capabilities other than just sending a few bytes), then work can start on the Virtual LMIC capsule. 

*Note:* There is code for an LMIC SPI capsule (as well as some commented out code initializing it) because initially I had wanted to to use SPI to communicate with the peripheral LoRa board; however, the LoRa board I was using was unable to act as a SPI peripheral, so I switched to using I2C instead. I still included it because I could using the LMIC SPI capsule instead of the LMIC I2C capsule to send bytes over SPI, so the LMIC SPI capsule is in fact functional (I could see using a Saleae Logic Analyzer that the bytes that the user wanted to send went through the SPI bus). I just can't use it to test it with the peripheral LoRa board, since as stated it cannot receive bytes as SPI peripheral. Development on the LMIC SPI capsule will probably end here for now. 

### Testing Strategy

I tested this on the Nordic nRF52840dk board. It is connected to a B-L072Z0KRWAN1 Discovery board, which has the LoRa radio, via I2C. I was able to verify that the user's specified packet was sent over I2C correctly using a Saleae Logic Analyzer to examine the bytes. I then confirmed that these bytes are then sent to a LoRa hotspot in the Helium network by using the Helium console to see that a hotspot had indeed received this packet. 

In my test however, I can only send bytes over I2C twice and then the kernel effectively ``segfaults'' with an error that I haven't called `replace()` one of the kernel buffers in the LoRa syscall controller properly. I'm still not exactly sure what I can use to `replace()` it. 

### TODO or Help Wanted

This pull request still needs a lot of work. I appreciate any feedback on what I have so far as well as input on how to work out the rest of the LMIC HIL/rest of the driver.

### Documentation Updated

I haven't updated any documentation yet, since this is still incomplete. 

### Formatting

- [x] Ran `make prepush`.


lschuermann:
Welcome @oliviaweng! It's great to see movement in making Tock speak LoRa! This definitely seems to be interesting for Tock, and there's already a tracking issue for it: #2344.

I think the approach of using a second MCU running LMIC is rather interesting, and looks at least similar to what we've been doing with nRF5x serialization on the Hail and Imix boards. I'm really interested to see how this compares to the architecture of #1760, which AFAIK tried to directly speak to the SX127X LoRa modem. Maybe this PR is indeed a more workable architecture, at least for the short term.

bradjc:
Looks like a good start! Some comments that might be helpful as you work on this, particularly to help the upstreaming/review process.

- It's fine to change the `boards/nrf52840dk` main.rs file during development, but eventually we will want to undo those changes, since not everyone using that development board will have a lora radio attached.
- Adding a new HIL might be too much overhead for this use case (and will certainly complicate/slow down the merge process). I would suggest moving the trait to the capsules crate, where it can serve the same purpose but without becoming a standard interface in tock (and incurring the requisite level of review).
- It might not be worth supporting both SPI and I2C. I would just use SPI for now. I2C can always be added later. That said, if you want to develop and test both then by all means do.
- Lastly, the tricking thing with LoRa is all of the different interfaces and software stacks and possible combinations of chips. Therefore, naming is tricky but important. I think the stack you are working on has much of the logic in userspace, and the kernel is simply a pass through to the hardware (although that may not be entirely the case). Naming the capsule files so that this structure is clear (and won't conflict with a different future stack that may be implemented more in the kernel) will be helpful.

bradjc:
Closing for now, but noted here: #2344. Please re-open if any new progress is made!

PR #2617, opened 2021-06-08
Update TRD104 to more clearly state the expectations for allowed buffers.

phil-levis:
### Pull Request Overview

This pull request grew out of #2590.

It does a bunch of cleanup on TRD104. The most notable change is a clearer definition of the expectations on allowed buffers. The prior text said that userspace cannot access a shared buffer. These changes update that text to be clearer and also state the assumptions that the kernel can make. Specifically:

1. A userspace API MUST NOT depend on or assume userspace accessing an allowed buffer.
2. The kernel MUST NOT assume that userspace does not access a buffer.

The first wording captures the intent. We cannot state that userspace MUST NOT access the buffer because there is no reasonable way to check or enforce this either through software or code review: a userspace bug could cause it to inadvertently access a buffer. The point is not that userspace APIs should not be designed such that they expect or use concurrent access. Therefore, if we in the future use an MPU to protect an allowed buffer we won't break userspace APIs.

The second wording captures that buffers can disappear unexpectedly. While userspace shouldn't be touching shared buffers, those buffers can change at any time. Userspace might revoke the buffer through another allow, or the process might exit (removing access to the buffer).

After a lot of discussion back and forth about concurrent kernel/user access to allowed buffers, we concluded this should be a different system call, which we will specify in a separate document.

### Testing Strategy

This pull request was tested by a huge number of discussions and debates.

### TODO or Help Wanted

This pull request still needs lots more careful reads.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alistair23 (approved):

doc/reference/trd104-syscalls.md
```diff
@@ -553,21 +556,104 @@ an error code of `INVALID`.
...
+to read or write to the buffer, it MUST first regain access to it by
+calling the corresponding Read-Write Allow.
+
+
+4.4.1 Buffers Can Change
+---------------------------------
+The standard use of Read-Write Allow requires that userspace does not
+access a buffer once it has been allowed. However, the kernel MUST NOT
+assume that an allowed buffer does not change: there could be a bug,
+compromise, or other error in the userspace code. The fact that the
```
lschuermann, line 578: I'm not entirely sure what _compromise_ means in this context?

lschuermann (commented):
I think these changes look good! One minor question though.

lschuermann (dismissed):

doc/reference/trd104-syscalls.md
```diff
@@ -553,21 +556,104 @@ an error code of `INVALID`.
...
+identical to Read-Write Allow: a userspace API MUST NOT depend on writing 
+to a shared buffer and the kernel MUST NOT assume the buffer does 
+not change.
+
+This restriction on writing to buffers is to limit the complexity of
+code review in the kernel. If a userspace library relies on writes to
+shared buffers, then kernel code correspondingly relies on them. This
+sort of concurrent access can have unforeseen edge cases which cause
+the kernel to panic, e.g., because values changed between method
+calls.
```
hudson-ayers, line 656: A second consideration here is also that a userspace API which requires this interface would complicate the libtock-rs API considerably -- that might be worth mentioning here as well.

hudson-ayers (dismissed):
Looks great, I have one (optional) suggestion

jrvanwhy:
Leon asked me a question that made me realize that I have not fully thought through `libtock-rs`' Allow API. To date, I have spent most of my time thinking about Read-Write Allow, and have assumed that Read-Only Allow is strictly easier. Unfortunately, I now realize that Read-Only Allow has its own challenge.

# The Challenge

To start, consider what a naive Allow API would look like in `libtock-rs`:

```rust
fn allow_ro_static(driver_id: u32, buffer_id: u32, new_buffer: &'static [u8])
    -> Result<&'static [u8], (ErrorCode, &'static [u8])>;

fn allow_rw_static(driver_id: u32, buffer_id: u32, new_buffer: &'static mut [u8])
    -> Result<&'static mut [u8], (ErrorCode, &'static mut [u8])>;
```

`allow_rw_static` looks good. It takes a `&mut` reference, which guarantees that there is no other reference to the buffer. It returns a `&mut` reference so that the code that owns the buffer can re-use it. That code would use a type similar to `TakeCell` to store the buffer, checking availability at runtime.

`allow_ro_static` is problematic, however. It is supposed to support buffers that are written by the app and read by the kernel -- but you can't build a `TakeCell`-like tool to track mutable and immutable access. If you try, you end up with something unsound:

```rust
impl TakeCell {
    // Retrieves a mutable reference if no other reference exists.
    fn take_mut(&self) -> Option<&'static mut [u8]>;

    // Returns the mutable reference.
    fn return_mut(&self, val: &'static mut [u8]);

    // Retrieves a shared reference if no other reference exists.
    fn take_ref(&self) -> Option<&'static [u8]>;

    // Returns the shared reference
    fn return_ref(&self, val: &'static [u8]);
}

fn main() {
    let mut variable = [0];
    let take_cell = TakeCell::new(&mut variable);
    let ref1: &[u8] = take_cell.take_ref().unwrap();
    let ref2: &[u8] = ref1; // Copies the reference.
    take_cell.return_ref(ref1);
    let mut_ref: &mut [u8] = take_cell.take_mut().unwrap();
    // mut_ref and ref2 are both alive here, which is undefined behavior
}
```

The problem is that `&` references are `Copy`, so there's no way for `RoTakeCell` to determine that every shared reference has been returned.

One apparent solution is to have `allow_ro_static` take a `&mut` reference instead for uniqueness:

```rust
fn allow_ro_static(driver_id: u32, buffer_id: u32, new_buffer: &'static mut [u8])
    -> Result<&'static mut [u8], (ErrorCode, &'static mut [u8])>;

fn allow_rw_static(driver_id: u32, buffer_id: u32, new_buffer: &'static mut [u8])
    -> Result<&'static mut [u8], (ErrorCode, &'static mut [u8])>;
```

but now `allow_ro_static` doesn't work with references that are pointing to read-only memory (i.e. truly immutable variables). The same result happens if we try to use `&[Cell<u8>]` instead: we can't create a `[Cell<u8>]` in readonly memory, forcing everything into RAM. One of the primary motivations for Read-Only Allow was to keep read-only buffers in immutable storage, so this is a deal-breaker.

# Idea: Move-Only Immutable Reference

Instead of using `&` references for Read-Only Allow, we can introduce our own reference type that is equivalent to `&` except it is non-copyable:

```rust
struct AllowRef { slice: core::ptr::NonNull<[u8]> }

impl std::ops::Deref for AllowRef {
    type Target = [u8];
    fn deref(&self) -> &[u8] { /* omitted */ }
}
```

Then we make Read-Only Allow use `AllowRef`:

```rust
fn allow_ro_static(driver_id: u32, buffer_id: u32, new_buffer: AllowRef)
    -> Result<AllowRef, (ErrorCode, AllowRef)>;

fn allow_rw_static(driver_id: u32, buffer_id: u32, new_buffer: &'static mut [u8])
    -> Result<&'static mut [u8], (ErrorCode, &'static mut [u8])>;
```

Then we need two `TakeCell`-like types: one for buffers in read-only memory, and a second for buffers in read-write memory. The read-only version can return multiple `AllowRef`s, while the read-write one can return at most one:

```rust
struct RoAllowBuffer<const LEN: usize> { buffer: [u8; LEN] }

impl<const LEN: usize> RoAllowBuffer<LEN> {
    // Does not need to perform runtime checks, as there can't be unique references to the buffer
    pub fn get(&self) -> AllowRef { /* omitted */ }
}

struct RwAllowBuffer<const LEN: usize> {
    borrowed: Cell<bool>,
    buffer: UnsafeCell<[u8; LEN]>,
}

impl<const LEN: usize> RwAllowBuffer<LEN> {
    // Returns a reference if not already borrowed.
    pub fn get_mut(&self) -> Option<&'static mut [u8]> { /* omitted */ }

    // Undoes the borrow. Must check that buffer == self.buffer at runtime.
    pub fn return_mut(&self, buffer: &'static mut [u8]) { /* omitted */ }

    // Returns a read-only reference if not already borrowed.
    pub fn get(&self) -> Option<AllowRef> { /* omitted */ }

    // Undoes the borrow. Must check that buffer == self.buffer at runtime.
    pub fn return_ref(&self, buffer: AllowRef) { /* omitted */ }
}
```

As an alternative, we could modify `RwAllowBuffer` to support multiple
simultaneous immutable borrows by having it track the number of `AllowRef`s
present.

# Other options?

Every other sound Allow API I can think of has considerably more overhead, so I think this is what `libtock-rs`' Allow API will need to look like.

For @phil-levis and @lschuermann: I do not think this impacts the TRD. :-)

phil-levis:
OK, I have rewritten the text so that it does not require the kernel to return `INVALID` on an overlapping Read-Write Allow call, and that the kernel has be safe if/when this happens.

jrvanwhy (approved):

hudson-ayers (approved):

alistair23 (approved):

lschuermann (approved):
This looks good to me.

lschuermann:
Awesome! As documented in the TRD, the kernel's AppSlice infrastructure as of now does not match the specification. I'll create a PR proposal with the slice-of-cells based mechanism @hudson-ayers and I worked on, as proposed on the `tock-dev` ML-thread.

Issue #2579, opened 2021-05-08
Tracking: Release new version of tock-register-interface


andre-richter:
Now that https://github.com/tock/tock/pull/2517 has landed, please consider releasing a new version on crates.io.

This will allow me to deprecate https://github.com/rust-embedded/register-rs and refer people to `tock-register-interface` for future use.

Thanks in advance,
Andre

@ppannuto @lschuermann

lschuermann:
Thanks for this tracking issue! Releasing a new version is definitely warranted due to this change, so it has been on my list for some time (rather, poking someone to do it :smile:).

That being said, I still have a few short term todos which I think should be done prior to a release:
- [ ] splitting up the internal structure, such that the interfaces, register types, and helpers are cleanly and semantically separated in different modules (that is already somewhat reflected in Tock's `kernel`-crate re-exports of this library, but should really be done in the library itself)
- [ ] adding a feature-flag to exclude the "default" register types (`ReadOnly`, `ReadWrite`, etc.), allowing this to be used as an interface-only library (following the previous change, essentially just disabling importing the module containing these types). A major advantage is that enabling this feature should then remove all unsafe code from the library. 
- [ ] doing a final pass over the doc-comments, and adding some documentation tests and more usage examples

I have some changes ready locally, and I expect to have all the boxes ticked in 1-2 weeks.

Thanks for your input on this, @andre-richter!

ppannuto:
Acking this thread, and eager for @lschuermann's updates! 

1 and 3 here are definite wins. I felt like there was possibly some unresolved discussion around 2 in the call last Friday, but we can re-examine those questions where there is a concrete proposal to discuss.

andre-richter:
@lschuermann please see https://github.com/tock/tock/pull/1489

hnj2:
Hello, I was wondering what the status of the release is?
According to #2618 it is nearly ready, but if that isn't the case a small version bump of `tock-registers` to `0.6.1` to make #2589 available might be nice for nightly users?

lschuermann:
@hnj2 Sorry for the wait! It seems that the only blocking PR for a release is #2551 and writing a Changelog, and I would've hoped that this is a rather uncontroversial set of changes. However, there seems to be some (entirely valid) discussion around this.

I'm not sure. Depending on what @ppannuto and @bradjc think a point release may be justified as the nightly breakage seems to be a rather pressing issue for the wider community.

[...]

PR #1489, opened 2019-11-28
register_structs!: Fix breakage of custom_test_frameworks

andre-richter:
The current version of the `register_structs!` macro is not usable in `no_std`
libraries or binaries that implement `#![feature(custom_test_frameworks)]`.

Current `register_structs!` definitions will break the test execution because
they emit unit tests with the `#[test]` attribute macro, which depends on the
`std` test crate which isn't there and can't be used in `no_std`:

```
error[E0463]: can't find crate for `test`
   | 
  ::: <::core::macros::builtin::test macros>:1:1
   |
1  |    ($ item : item) => { }
   |    ---------------------- in this expansion of `#[test]`
  --> <::tock_registers::macros::register_structs macros>:11:49
   |
1  |  / { $ ($ (# [$ attr : meta]) * $ name : ident { $ ($ fields : tt) * }), * } =>
2  |  | {
3  |  |     $
4  |  |     ($ crate :: register_fields !
...   |
11 |  |              use super :: super :: * ; # [test] fn test_offsets ()
   |  |________________________________________--------_^
   | ||                                        |
   | ||                                        in this macro invocation
12 | ||              { $ crate :: test_fields ! (@ root $ name $ ($ fields) *) }
   | ||________________________________________________________________________^ can't find crate
13 |  |          }) *
14 |  |     }
15 |  | } ;
   |  |___- in this expansion of `register_structs!`
```

More literature at: https://os.phil-opp.com/testing

Fix this by making the `std` unit tests opt-in through a feature. I went for opt-in because tock-registers is `no_std` by nature.

CC @ppannuto @gendx

PR #2618, opened 2021-06-09
tock-register-interface: reorganize & document

lschuermann:
### Pull Request Overview

This pull request refactors the `tock-register-interface` crate. It implements the remaining todos of #2579, and should enable us to release a new `tock-registers` release on crates.io soon.

### Motivation

#2517 introduced traits to access / manipulate `tock-registers` register types. One large benefit of this is that other libraries can now use the interfaces, types and macros provided by `tock-registers` on their own register types. By implementing these traits, the register types will be fully compatible with the register wrappers shipped as part of `tock-registers`. The utility of this is proven by our very own `riscv-csr` crate, which does exactly that: implement some custom register types, which implement the `Readable` and `Writeable` traits as exported by `tock-registers` and thereby can integrate with other useful macros and types, such as `Field` and `register_bitfields!`.

However, this also emphasized some differences between types in `tock-registers`. For instance, a `LocalRegisterCopy` is a fundamentally different thing than a `ReadWrite`/`InMemoryRegister`. It's not a register per se, implements the Copy trait and requires a mutable borrow to be mutated, hence not implementing the usual `Readable`/`Writeable`/`ReadWriteable` traits.

Furthermore, the ability to utilize the `tock-registers` infrastructure by types defined outside of this crate emphasizes the fact that types such as `Field`, `FieldValue`, and traits like `IntLike` and `RegisterLongName` aren't only related or usable by the included register wrappers.

Previously, the crate shipped all of these types and traits in the `registers` module, and all macros in the `macros` module (which get exported as top-level crate exports automatically). After the changes mentioned above, this seemed rather arbitrary.

This PR is an attempt to reorganize the `tock-registers` structure. Furthermore, it adds a lot of Rustdoc documentation and implicitly adds integration tests through the included code examples, which get compiled and run by `cargo test`.

Not every user will need the shipped register types. Therefore, this PR introduces a feature-flag to enable building the `crate::registers` module, which is part of the default feature set. A significant advantage is that, by disabling this feature flag, the crate does not use any unsafe code. In fact, this is enforced by setting `forbid(unsafe_code)` if `register_types` is not enabled. This is really motivated by usages of this crate external to Tock, I'm not proposing to necessarily utilize this feature in Tock.

### Approach

The final crate exports look like this:

```rust
crate tock-registers {
    // Include IntLike and RegisterLongName as part of the top-level crate exports
    // as they represent basic building blocks of all types and macros included
    // in this crate.
    pub trait IntLike;
    pub trait RegisterLongName;

    // Top-level as it's not really a register, but a common and useful type, relating
    // to all registers implementing the crate::interfaces::* traits
    pub struct LocalRegisterCopy;

    pub mod fields {
        pub struct Field;
        pub struct FieldValue;

        // Is only used by fields and the corresponding bitfields macros
        pub trait TryFromValue;

        // Also contains all macros and tests relating to bitfields
        // although macros are exported as top-level crate items by Rust
    };

    pub mod interfaces {
        pub trait Readable;
        pub trait Writeable;
        pub trait ReadWriteable;
    };

    // Only included when `register_types` feature is enabled, part of
    // the default features
    pub mod registers {
        pub struct ReadWrite;
        pub struct ReadOnly;
        pub struct WriteOnly;
        pub struct Aliased;
    };

    // Contains the register_structs macros, don't quite know where to put these
    mod macros;
}
```

This structure seems pretty natural, both in terms of usage and dividing the actual code / implementation. I'm not 100% convinced on putting `LocalRegisterCopy` in the top-level module, and I'd really like to find a better name for the `register_structs!` macros other than _macros_.

Of course, this structure is somewhat subjective. Excited to hear what others think.

### Testing Strategy

### TODO or Help Wanted

- [ ] This pull request possibly still needs update of the `README.md` once the final structure is settled on.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


lschuermann:
One other thing came up during the development: what is the utility of `no_std_unit_tests`?

I see that the `tests`-module (now as a submodule of `crates::fields`) is not present when this feature is set. However, we can't use the standard library in there regardless (the whole crate is always `no_std` currently):

```
   Compiling tock-registers v0.6.0 ($TOCK/libraries/tock-register-interface)
error: cannot find macro `println` in this scope
   --> libraries/tock-register-interface/src/fields.rs:474:6
    |
474 |         println!("Test!");
    |         ^^^^^^^

```

Furthermore, as documented in the [feature unification section of the cargo documentation](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification), features should always be additive. `no_std_unit_tests` clearly violates that specification by disabling certain functionality. As such -- were we ever to introduce tests that relied on the standard library being present -- we should have a feature `std_tests` instead which explicitly enables the standard library and is not in the default feature set, and not the other way around.

Therefore, if there is no good reason to keep `no_std_unit_tests` around, I'd vouch to remove it.

andre-richter:
@lschuermann please see https://github.com/tock/tock/pull/1489

bradjc:
+1 on interfaces.rs.

I don't understand why not have macros.rs any more. The amazingly complicated register definition macros are a key piece of the library, and it was easy to find them in macros.rs. Fields is very generic.

Why does so much code have to move to lib.rs? lib.rs is the least descriptive name. For example, why is a local register copy a lib, and not still in the registers.rs file?

lschuermann:
@bradjc 

> I don't understand why not have macros.rs any more. The amazingly complicated register definition macros are a key piece of the library, and it was easy to find them in macros.rs. Fields is very generic.

There is still `macros.rs`, but for the register structs. The other macros were related to bitfields. These have been moved into a common module along the `Field`s and `FieldValue`s as they are closely related. The concept of a `macros` module is new to me, in other codebases I tend to see macros placed next to the types on which they operate, as there is a strong coupling there. I'm not particularly invested in this and I'll happily move it back if you don't agree.

> Why does so much code have to move to lib.rs? lib.rs is the least descriptive name. For example, why is a local register copy a lib, and not still in the registers.rs file?

I agree that `lib.rs` is not a particularly descriptive name, though I'd rather think of it as the top-level crate module (from a user POV, that is `use tock_registers::{...}`.

As far as I know it's a common concept to have very generic types, and especially types which are used throughout the library, to be placed in the top-level module. This is why `IntLike` and `LocalRegisterCopy` have been moved there, as these are literally some of the basic building blocks of the entire tock-registers infrastructure.

Now, `LocalRegisterCopy` is a tricky beast: it's really not a register. It implements `Copy`, it requires mutable `&self` references, it can only be constructed _from_ a register and it does not use the `interfaces::*` traits, all of which deviate from proper registers. As they name says, it's really only the copy of some register, whatever type that source register has. Therefore, it is relevant to both register types shipped in this library, and register types created elsewhere (e.g. `riscv-csr`s).

Because of this distinction, I've moved it to the top-level crate exports. I'm not invested on this location. It did have the benefit of enabling other crates (such as `riscv-csr`) to disable the `register_types` feature, which would simply disable including the `tock_registers::registers` module, making this an interface-only library. Furthermore, this removes all `unsafe` code, which is a potential criterium for other users interested in this crate (we as Tock are concerned with that as well, right?).

I could move it back to `registers` if you'd like, and simply make including the types themselves conditional on this feature flag.

hnj2:
As far as I can tell the `README.md` doesn't need to be changed :+1: I've read through it and did not find any interface changes that need to be reflected there.
Are there other **Help Wanted/TODO** items?

andre-richter:
re readme:
There is this one paragraph starting with “WARNING” which references an already merged PR. What about this one?

lschuermann:
@hnj2, @andre-richter Thanks! Removed the warning from the README in de9ddcc, as indeed the offsets in the register structs seem to be properly tested in CI.

ppannuto:
Overall I think this is pretty good.

I think that moving the macros for fields to be next to the fields is fine, don't really have strong feelings either way there.

The `lib.rs` is bit trickier. I really think Rust did the world a bit of a disservice there by not setting out whether it's an interfaces file or a file intended to hold code. I don't think we are so consistent, but are moving towards `lib.rs` / `mod.rs` as interface descriptors / re-export files, but not holding code themselves. I'd be inclined to go ahead and move the code into something like `local_register.rs` and just re-export via `lib.rs`.

hnj2:
Concerning the Documentation: The traits `interfaces::*` enable the implementation of non-mmio registers, and this is nicely documented in the `interfaces` submodule.
But a reference to this functionality is missing from the crates root documentation and the README.

Additionally i think the `DummyRegister` example could be replaced by something concrete like an ARM SPSR register:
```
```

lschuermann:
@hnj2 Thanks for the great feedback, will incorporate! I haven't looked at the README in detail after these changes yet, will do a final pass. Also (as a reminder for myself) we're going to want to update the changelog.

bradjc (dismissed):

lschuermann:
Will incorporate the remaining changes this evening.

lschuermann:
I think this should be ready now.

@ppannuto 
>  I don't think we are so consistent, but are moving towards `lib.rs` / `mod.rs` as interface descriptors / re-export files, but not holding code themselves. I'd be inclined to go ahead and move the code into something like `local_register.rs` and just re-export via `lib.rs`.

Makes sense, done.

@hnj2 
> But a reference to this functionality is missing from the crates root documentation and the README.

I tried to hint at it in the README, and also at the fact that one can now use this library to wrap CPU registers and other exciting things. However I've always found the README to be a good starting point for `tock-registers`, where for more in-depth knowledge one would look at the Rustdoc. Therefore, I wouldn't make this implementation detail too prominent there. I suppose we could add a sentence to the root documentation...

> Additionally i think the `DummyRegister` example could be replaced by something concrete like an ARM SPSR register

I agree that would be nice. However, I'm using the `DummyRegister` documentation as a doctest, which IMHO is a rather valuable integration test given how much of the library it ties together in one example. Using some more realistic but architecture-dependent code snippet would make this more difficult, if not impossible.

@ppannuto assuming this is close to being merged, would we create a follow-up PR bumping the version and updating the changelog or do that here?

ppannuto (dismissed):
This is great, thank you for doing all the heavy lifting here.

Regarding the documentation question; perhaps add an explicit link to https://docs.tockos.org/tock_registers/index.html somewhere in the README might be the best of both worlds?

ppannuto:
I think a follow-up PR with just changelog and version is our usual approach / best.

ppannuto:
@lschuermann that netlify build error looks real:

image
^I think we fixed the permissions for any to view logs, but just in case

lschuermann:
Ah, sorry. Forgot to run `cargo doc` on the latest changes. I knew I was forgetting something! Will fix shortly.

bradjc (dismissed):

lschuermann:
> that netlify build error looks real

Fixed! Let's see what the CI says now. :)

ppannuto (approved):
bors r+

🎉

PR #2619, opened 2021-06-10
Add I2C slave support for nRF

alistair23:
### Pull Request Overview

This builds on top of https://github.com/tock/tock/pull/2612 by adding nRF I2C `master_slave_driver` support.

### Testing Strategy

Running I2C communication on the nRF

### TODO or Help Wanted

Nothing

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/i2c_master_slave_driver.rs
+++ b/capsules/src/i2c_master_slave_driver.rs
@@ -16,7 +16,7 @@ use core::cmp;
 use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::hil;
 use kernel::{CommandReturn, ProcessId, Upcall};
-use kernel::{Driver, ErrorCode, Grant, Read, ReadWrite, ReadWriteAppSlice};
+use kernel::{Driver, ErrorCode, Grant, Read, ReadOnlyAppSlice, ReadWrite, ReadWriteAppSlice};
 
 pub static mut BUFFER1: [u8; 256] = [0; 256];
 pub static mut BUFFER2: [u8; 256] = [0; 256];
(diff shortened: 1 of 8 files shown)


capsules/src/i2c_master_slave_driver.rs
```diff
@@ -130,7 +130,7 @@ impl hil::i2c::I2CHwMasterClient for I2CMasterSlaveDriver<'_> {
                         // than the buffer could lead to array overrun errors in
                         // userspace. The I2C syscall API should pass back lengths.
                         // -pal 3/5/21
-                        app.master_tx_buffer.mut_map_or(0, move |app_buffer| {
+                        app.master_rx_buffer.mut_map_or(0, move |app_buffer| {
```
hudson-ayers, line 133: Was this a bug that it used `master_tx_buffer` before? Or is the idea that `WriteRead` ops should send from the tx buffer but read into the rx buffer?

Should this be accompanied by a TRD update or at least a libtock-c update to reflect which buffer is used here?

hudson-ayers (approved):
Looks good, though one question about a buffer change in `capsules`

capsules/src/i2c_master_slave_driver.rs
alistair23: The libtock-c documentation and interface isn't clear, but this seems like a bug.

The `MasterAction::Read` will copy the read data to the `app.master_rx_buffer` buffer while the read action from `MasterAction::WriteRead` copies data to `app.master_tx_buffer`. This changes all read operations to store data to `app.master_rx_buffer`. I'll add a comment to the libtock-c interface to make this clear

hudson-ayers: That sounds right to me, thanks

alistair23:
This is ready to go

bradjc (approved):

PR #2623, opened 2021-06-12
imxrt10xx: Add DMA peripheral, integrate DMA into LPUART

mciantyre:
### Pull Request Overview

This pull request

- adds a DMA driver for i.MX RT chips.
- extends the LPUART driver to optionally support DMA transfers and receives.
- demonstrates a DMA-driven LPUART peripheral on the Teensy 4.

### Testing Strategy

Tested on a Teensy 4.0 with these Tock examples:

- console
- console_recv_short
- console_recv_long

Ensured that interrupt-driven LPUART continues to work, and that DMA-driven LPUART works, too. See the latest commit's message for specifics.

### TODO or Help Wanted

We could be clever about when we use DMA- or interrupt-driven I/O in the LPUART driver. The LPUART peripheral has small TX and RX queues: four words in each direction. We're not using them now (corrected the comment), but a future implementation could use them to quickly move small buffers without a DMA transfer.

The DMA implementation exposes 32 DMA channels. Nearly all of the i.MX RT 10xx MCUs have 32 DMA channels, except for the 1011. Accessing DMA channels 16 through 31 on that system would index into undocumented / reserved peripheral memory. There's no imxrt1011 board in tree, so this PR targets the 1050 and 1060 boards which are supported in tree.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/teensy40/src/main.rs
+++ b/boards/teensy40/src/main.rs
@@ -56,6 +56,32 @@ impl kernel::Platform for Teensy40 {
     }
 }
 
+/// Static configurations for DMA channels.
+///
+/// All DMA channels must be unique.
+mod dma_config {
+    use super::imxrt1060::nvic;
+
+    /// DMA channel for LPUART2_RX (arbitrary).
+    pub const LPUART2_RX: usize = 7;
+    /// DMA channel for LPUART2_TX (arbitrary).
+    pub const LPUART2_TX: usize = 8;
+
+    /// Add your DMA interrupt vector numbers here.
+    const DMA_INTERRUPTS: &[u32] = &[nvic::DMA7_23, nvic::DMA8_24];
+
+    /// Enable DMA interrupts for the selected channels.
+    #[inline(always)]
+    pub fn enable_interrupts() {
+        DMA_INTERRUPTS
+            .iter()
+            .copied()
+            // Safety: creating NVIC vector in platform code. Vector is valid.
+            .map(|vector| unsafe { cortexm7::nvic::Nvic::new(vector) })
+            .for_each(|intr| intr.enable());
+    }
+}
+
 /// This is in a separate, inline(never) function so that its stack frame is
 /// removed when this function returns. Otherwise, the stack space used for
 /// these static_inits is wasted.
@@ -162,8 +188,17 @@ pub unsafe fn main() {
         peripherals.ccm.perclk_divider(),
     );
 
+    peripherals.dma.clock().enable();
+    peripherals.dma.reset_tcds();
+    peripherals
+        .lpuart2
+        .set_rx_dma_channel(&peripherals.dma.channels[dma_config::LPUART2_RX]);
+    peripherals
+        .lpuart2
+        .set_tx_dma_channel(&peripherals.dma.channels[dma_config::LPUART2_TX]);
+
     cortexm7::nvic::Nvic::new(imxrt1060::nvic::GPT1).enable();
-    cortexm7::nvic::Nvic::new(imxrt1060::nvic::LPUART2).enable();
+    dma_config::enable_interrupts();
 
     let chip = static_init!(Chip, Chip::new(peripherals));
     CHIP = Some(chip);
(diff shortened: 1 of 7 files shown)


bradjc (dismissed):

boards/teensy40/src/main.rs
```diff
@@ -162,8 +162,18 @@ pub unsafe fn main() {
         peripherals.ccm.perclk_divider(),
     );
 
+    peripherals.dma.clock().enable();
+    peripherals.dma.reset_tcds();
+    peripherals
+        .lpuart2
+        .set_rx_dma_channel(&peripherals.dma.channels[7]);
```
ppannuto, line 169: Are the DMA channels arbitrary or mapped to specific peripherals? `7` here and `8` below feel a bit magic-numbery, which might be fine, but if they could be replaced with a meaningful enum or something that might help readability.

chips/imxrt10xx/src/dma.rs
```diff
@@ -0,0 +1,784 @@
...
+    ]
+];
+
+#[repr(C, align(32))]
+struct TransferControlDescriptor {
+    saddr: ReadWrite<u32>,
+    soff: ReadWrite<u16>, // Signed number
+    attr: ReadWrite<u16, TransferAttributes::Register>,
+    nbytes: ReadWrite<u32>, // Assumes minor loop mapping is disabled (EMLM = 0)
+    slast: ReadWrite<u32>,  // Signed number
```
ppannuto, line 94: These comments are a bit confusing -- is there a reason we can't use `<i32>` here?

chips/imxrt10xx/src/dma.rs
```diff
@@ -0,0 +1,784 @@
...
+
+/// Describes a type that can be transferred via DMA.
+pub trait DmaElement: private::Sealed {
+    /// An identifier describing the data transfer size
+    ///
+    /// See TCD\[SSIZE\] and TCD\[DSIZE\] for more information.
+    const DATA_TRANSFER_ID: u16;
+}
+
+mod private {
```
ppannuto, line 410: What is this? Can we document this?

chips/imxrt10xx/src/dma.rs
hudson-ayers: cc @lschuermann -- shouldn't this be pretty easy to add now, with the new traits? Is there any reason not to do this?

lschuermann: It's probably easy to add, though we must be careful on how we interact with bitfields then. We can't implement it outside of the library as either the trait (`IntLike`) or the type needs to be in the local crate for that to work. We might want to create a supertrait of `IntLike` which then explicitly enables bitfield-style manipulations on unsigned integers only, because semantically that might not make a ton of sense with signed integers.

Whether or not this is useful in general I don't know. Not being able to use bitfields for that takes away a major chunk of Tock-register's utility. I'm not strictly opposed, but casting seems easy enough if it's just rare cases.

ppannuto: @mciantyre short term, I think leaving as `u16` with comments is fine. Long term, we should probably consider this part of the registers interface, but that's orthogonal to this PR.

boards/teensy40/src/main.rs
ppannuto: Mhmm, I think coalescing interrupts is probably a premature optimization at this stage. Capturing some of this design tradeoff in comments would be nice, but otherwise leaving things as they are should be fine for now.

mciantyre:
- Rebase
- Add DMA channel considerations in `dma` module documentation
- Add notes about sealed trait pattern in `dma` module
- Favor constants when identifying DMA channels in `boards/teensy40`

ppannuto (approved):

bradjc (approved):
I don't understand why the DMA config is in the board file rather than the chip (example: https://github.com/tock/tock/blob/master/chips/sam4l/src/chip.rs#L117), but it's probably not a blocking issue.

PR #2624, opened 2021-06-15
Fixes three printing bugs in `print_tock_memory_usage.py`

phil-levis:
### Pull Request Overview

Fixes three bugs in the memory printing tool.

The first was that many "::" were being incorrectly parsed into ":" in symbol demangling, making grepping the output harder. Now all are properly demangled as "::".

The second is that the number of symbols in a group was not being correctly calculated due to a scoping bug; only the last symbol group's size was considered and applied to all groups. 

The third bug relates to printing whether a group has multiple elements under it; single element groups were truncated names. Changed to show a * if the group is not a complete name, even if there is only one element underneath.
### Testing Strategy

### TODO or Help Wanted

All good.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/tools/print_tock_memory_usage.py
+++ b/tools/print_tock_memory_usage.py
@@ -103,7 +103,7 @@ escape_sequences = [
     ["$u27$",   "\'"],
     ["$u5b$",   "["],
     ["$u5d$",   "]"],
-    ["..",      ":"],
+    ["..",      "::"],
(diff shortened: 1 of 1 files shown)


ppannuto (approved):
bors r+

Looks an easy / small maintenance fix

PR #2628, opened 2021-06-22
kernel: add common KernelResult type

jettr:
Add a helper type that we can use instead of `Result<(), ErrorCode>` in most places in the kernel.

### Testing Strategy

Used new `KernelResult` in downstream code with success.

### TODO

- If this seems good, I will go through and change `Result<(), ErrorCode> ` to `KernelResult` in the existing code before submitting PR

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/lib.rs
+++ b/kernel/src/lib.rs
@@ -129,6 +129,9 @@ pub use crate::sched::round_robin::{RoundRobinProcessNode, RoundRobinSched};
 pub use crate::sched::{Kernel, Scheduler};
 pub use crate::upcall::Upcall;
 
+/// Standard kernel result that uses [`ErrorCode`] as the error.
+pub type KernelResult<T = ()> = Result<T, ErrorCode>;
+
 // Export only select items from the process module. To remove the name conflict
 // this cannot be called `process`, so we use a shortened version. These
 // functions and types are used by board files to setup the platform and setup


jrvanwhy:
IIRC this was a topic of discussion among the core WG for several months out of the past year :-)

I recommend syncing with @hudson-ayers on this.

alistair23:
Although I think `Result<(), ErrorCode>` is used a fair bit inside the kernel I am more in favour or everything being contained within the return. So for example if a buffer could be returned the return value should be: `Result<usize, (ErrorCode, &'static mut [u8])>`. In which case something like `KernelResult` might not be as useful as lots of HIL functions take a buffer (but not all).

In saying that there are still lots of places that don't do that, and I'm not sure if it's agreed on which is the correct way forward.

lschuermann:
I agree with @alistair23. Using proper `Result<T, E>` types has the significant advantage of being able to precisely express what information can be conveyed in either the error, or the success case. I know that historically `ReturnCode` had seen significant use in error handling in Tock, and that is conceptually similar to `Result<(), ErrorCode>`, I don't think we are going to want to motivate using this for new code by making it easier to express, for two reasons:

- most often one does want to convey additional data in both the success and error cases. Using a `KernelResult` in an API will therefore require additional fields for success data, which don't contain meaningful data in the error case. This should ideally all be encapsulated in the `T` and `E` of `Result`, such that code using these values if forced to check for errors. I know the pattern of having success fields and an error indicator is still somewhat common in the Tock codebase, for example in HILs, but having `ErrorCode` and `Result` will -- for new APIs -- foster more meaningful integration of these discriminated union types.

- `ErrorCode` often is not a particularly concise notion of the errors which can occur over a given interface. Granted, it's useful for conveying such information to userspace and is a well-defined type across the codebase. However, if for a given interface it does not work out, we're going to want to motivate people to use the flexibility provided by `Result` in that one can freely choose `E`. For HILs specifically, this is documented in [this revision of the draft TRD about HIL design](https://github.com/tock/tock/blob/73f240814e800abe3b9946a604fe1c7d689fb164/doc/reference/trd-hil-design.md#rule-2-return-synchronous-errors).

That being said, I'm always in favor of typing less, so this is appreciated! However, for the reasons outlined above, I have a preference for explicitly expressing the aforementioned concept -- if it makes sense to use -- as `Result<(), ErrorCode>`, because it makes one think about what information is only conveyed in the success and error cases.

jettr:
Thanks for the quick feedback and link to the hil design guide (it was a good read)!

I agree that not all code should use this proposed `KernelResult`. I can definite add comments to that effect saying that creating your own error enum for your hil is typically the correct approach (and passing an buffers back in the error case).

I do still think that `Result<(), ErrorCode>` will be used in the code quite a bit (I found 859 instances of it the current code), and having a nicer alias would help out there in my opinion.

Also maybe we should call it `UpcallResult` instead of `KernelResult` to encourage people to think of it as a upcall boundary result. Then hopefully people will not mistake it for something they should use deep in kernel code but only translate to near the boundary (and if you really have nothing else useful to add, you can use it deeper in the kernel). Thoughts?

bradjc:
I prefer to keep it simple with the default Rust `Result<>` syntax. Otherwise there is another layer of indirection for people reading the code. Also, we have custom Tock types for return values in certain places that indicate we needed something custom for a very specific reason (often related to the syscall interface), and it signals to readers that there is some subtle complexity they have to handle. Using a custom type just as shorthand would break this signaling mechanism.

jettr:
Sounds good. Abandoning change. Thank you for the feedback.

PR #2629, opened 2021-06-23
OpenTitan: AES: Update the AES implementation

alistair23:
### Pull Request Overview

This updates the implemention to remove all infinite loops, use callbacks and update to the current OpenTitan bitstream addresses and registers.

### Testing Strategy

Running the Tock AES tests.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/chips/earlgrey/src/aes.rs
+++ b/chips/earlgrey/src/aes.rs
@@ -2,13 +2,13 @@
 //!
 //! <https://docs.opentitan.org/hw/ip/aes/doc/>
 
+use core::cell::Cell;
 use kernel::common::cells::{OptionalCell, TakeCell};
 use kernel::common::registers::interfaces::{Readable, Writeable};
 use kernel::common::registers::{
     register_bitfields, register_structs, ReadOnly, ReadWrite, WriteOnly,
 };
 use kernel::common::StaticRef;
-use kernel::debug;
 use kernel::hil;
 use kernel::hil::symmetric_encryption;
 use kernel::hil::symmetric_encryption::{AES128_BLOCK_SIZE, AES128_KEY_SIZE};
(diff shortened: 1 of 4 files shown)


chips/earlgrey/src/aes.rs
```diff
@@ -318,32 +392,116 @@ impl<'a> hil::symmetric_encryption::AES128<'a> for Aes<'a> {
                 }
             }
         }
+
+        if self.deferred_call.get() {
+            panic!("Operation already in progress");
```
bradjc, line 397: This can't just return an error?

chips/earlgrey/src/aes.rs
```diff
@@ -318,32 +392,116 @@ impl<'a> hil::symmetric_encryption::AES128<'a> for Aes<'a> {
...
+        self.registers.ctrl.write(ctrl);
     }
 }
 
 impl kernel::hil::symmetric_encryption::AES128ECB for Aes<'_> {
     fn set_mode_aes128ecb(&self, encrypting: bool) {
-        self.configure(encrypting);
+        if !self.idle() {
+            panic!("AES is not idle");
+        }
```
bradjc, line 452: This is unfortunate.

alistair23: Yeah, we can't return an error so I'm not sure what else to do

alistair23: I'm planning on working on capsules as well, if I end up changing the HIL I can fix this up and remove these

ppannuto (approved):
It would be good to fix the HIL, but in the now, I guess the panics are better than incorrect behavior.

bradjc (approved):
https://github.com/tock/tock/issues/2635

alistair23:
This is ready to go

PR #2632, opened 2021-06-25
Support aliasing of process-allowed memory

lschuermann:
### Pull Request Overview

As of 8f66d98 ("Merge #2617"), it is explicitly permitted for Tock applications to _Allow_ overlapping memory regions to multiple Allow slots in both the same and different capsules simultaneously. This has the implication of making the current interface for accessing process memory (`AppSlice`) unsound.

This commit rewrites the interface for interacting with process memory, such that it is valid for the kernel to have multiple process buffers which can be both writeable and overlapping. This is achieved by developing an interface around `&[Cell<u8>]` (slice of Cells), which does permit interior mutability due to the usage of `UnsafeCell<u8>`.

It further adapts a consistent naming scheme:

- App -> Process, as potentially multiple processes of the same application can run simultaneously.

- AppSlice -> ProcessBuffer, as this is not really a slice, but rather represents a bounded memory region (buffer) within some process memory.

- introduction of new `ProcessSlice`s, which are `#[repr(transparent)]` wrappers around slices of Cells, providing convenient APIs and, in the case of read-only Allowed memory, limiting to read-only accesses.

- introduction of a `ReadableProcessByte`, which is a `#[repr(transparent)]` wrapper around a `Cell<u8>`, limiting to read-only accesses.

In general, this manages to provide a convenient interface for accessing userspace memory, while not directly operating on Rust slices of `u8`s with their inherent aliasing restrictions.

However, the `ProcessSlices` obtained are incompatible with regular Rust slices, which sometimes makes copying over an intermediate buffer necessary.

Usually, Rust slices only feature a `copy_from_slice` method, taking a mutable `&mut self` reference. However, given that a Rust slice cannot copy from a `ProcessSlice`, these feature an inverse method `copy_to_slice`, which can write the contents to a regular Rust slice.

### Testing Strategy

This pull request was tested by compiling. It needs more testing.

### TODO or Help Wanted

This pull request is a work on progress. Feedback in general, and especially regarding `mem.rs`'s contents is very welcome!

In particular, the CRC capsule is rather hard to port to this interface. This is because the HIL in it's current form is unsound if a DMA / asynchronous hardware is used, as it takes a slice reference with an anonymous lifetime. It will need to be fixed in a separate PR, as a precondition to merging this.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/driver.rs
+++ b/kernel/src/driver.rs
@@ -70,7 +70,7 @@
 //! encoding these types into the Tock system call ABI specification.
 
 use crate::errorcode::ErrorCode;
-use crate::mem::{ReadOnlyAppSlice, ReadWriteAppSlice};
+use crate::mem::{ReadOnlyProcessBuffer, ReadWriteProcessBuffer};
 use crate::process;
 use crate::process::ProcessId;
 use crate::syscall::SyscallReturn;
@@ -188,7 +188,7 @@ pub trait Driver {
         CommandReturn::failure(ErrorCode::NOSUPPORT)
     }
 
-    /// System call for a process to pass a buffer (a ReadWriteAppSlice) to
+    /// System call for a process to pass a buffer (a ReadWriteProcessBuffer) to
     /// the kernel that the kernel can either read or write. The kernel calls
     /// this method only after it checks that the entire buffer is
     /// within memory the process can both read and write.
@@ -196,13 +196,13 @@ pub trait Driver {
         &self,
         app: ProcessId,
         which: usize,
-        slice: ReadWriteAppSlice,
-    ) -> Result<ReadWriteAppSlice, (ReadWriteAppSlice, ErrorCode)> {
+        slice: ReadWriteProcessBuffer,
+    ) -> Result<ReadWriteProcessBuffer, (ReadWriteProcessBuffer, ErrorCode)> {
         Err((slice, ErrorCode::NOSUPPORT))
     }
 
     /// System call for a process to pass a read-only buffer (a
-    /// ReadOnlyAppSlice) to the kernel that the kernel can read.
+    /// ReadOnlyProcessBuffer) to the kernel that the kernel can read.
     /// The kernel calls this method only after it
     /// checks that that the entire buffer is within memory the
     /// process can read. This system call allows a process to pass
@@ -211,8 +211,8 @@ pub trait Driver {
         &self,
         app: ProcessId,
         which: usize,
-        slice: ReadOnlyAppSlice,
-    ) -> Result<ReadOnlyAppSlice, (ReadOnlyAppSlice, ErrorCode)> {
+        slice: ReadOnlyProcessBuffer,
+    ) -> Result<ReadOnlyProcessBuffer, (ReadOnlyProcessBuffer, ErrorCode)> {
         Err((slice, ErrorCode::NOSUPPORT))
     }
 
--- a/kernel/src/mem.rs
+++ b/kernel/src/mem.rs
@@ -8,19 +8,29 @@
 //! allow_read_only() system calls.
 //!
 //! A read-write and read-only call is mapped to the high-level Rust
-//! types [`ReadWriteAppSlice`] and [`ReadOnlyAppSlice`]
+//! types [`ReadWriteProcessBuffer`] and [`ReadOnlyProcessBuffer`]
 //! respectively. The memory regions can be accessed through the
-//! [`Read`] and [`ReadWrite`] traits, implemented on the
-//! AppSlice-structs.
+//! [`ReadableProcessBuffer`] and [`WriteableProcessBuffer`] traits,
+//! implemented on the process buffer structs.
+//!
+//! Each access to the buffer structs requires a liveness check to ensure that the
+//! process memory is still valid. For a more traditional interface, users can convert
+//! buffers into [`ReadableProcessSlice`] or [`WriteableProcessSlice`] and use these
+//! for the lifetime of their operations. Users cannot hold live-lived references to
+//! these slices, however.
+
+use core::cell::Cell;
+use core::ops::{Index, Range, RangeFrom, RangeTo};
 
 use crate::capabilities;
-use crate::process::ProcessId;
+use crate::process::{self, ProcessId};
 
-/// Convert an AppSlice's internal representation to a Rust slice.
+/// Convert a process buffer's internal representation to a
+/// ReadableProcessSlice.
 ///
-/// This function will automatically convert zero-length AppSlices
-/// into valid zero-sized Rust slices regardless of the value of
-/// `ptr`.
+/// This function will automatically convert zero-length process
+/// buffers into valid zero-sized Rust slices regardless of the value
+/// of `ptr`.
 ///
 /// # Safety requirements
 ///
@@ -29,36 +39,49 @@ use crate::process::ProcessId;
 /// nonzero. This memory region must be mapped as _readable_, and
 /// optionally _writable_ and _executable_. It must be allocated
 /// within a single process' address space for the entire lifetime
-/// `'a`. No mutable slice pointing to an overlapping memory region
-/// may exist over the entire lifetime `'a`.
-unsafe fn raw_appslice_to_slice<'a>(ptr: *const u8, len: usize) -> &'a [u8] {
-    use core::ptr::NonNull;
-    use core::slice::from_raw_parts;
-
-    // Rust has very strict requirements on pointer validity[1] which
-    // also in part apply to accesses of length 0. We allow an
-    // application to supply arbitrary pointers if the buffer length is
-    // 0, but this is not allowed for Rust slices. For instance, a null
-    // pointer is _never_ valid, not even for accesses of size zero.
-    //
-    // To get a pointer which does not point to valid (allocated) memory, but
-    // is safe to construct for accesses of size zero, we must call
-    // NonNull::dangling(). The resulting pointer is guaranteed to be well-aligned
-    // and uphold the guarantees required for accesses of size zero.
-    //
-    // [1]: https://doc.rust-lang.org/core/ptr/index.html#safety
-    match len {
-        0 => from_raw_parts(NonNull::<u8>::dangling().as_ptr(), 0),
-        _ => from_raw_parts(ptr, len),
-    }
+/// `'a`.
+///
+/// It is sound for multiple overlapping [`ReadableProcessSlice`]s or
+/// [`WriteableProcessSlice`]s to be in scope at the same time.
+unsafe fn raw_processbuf_to_roprocessslice<'a>(
+    ptr: *const u8,
+    len: usize,
+) -> &'a ReadableProcessSlice {
+    // Transmute a reference to a slice of Cell<u8>s into a reference
+    // to a ReadableProcessSlice. This is possible as
+    // ReadableProcessSlice is a #[repr(transparent)] wrapper around a
+    // [ReadableProcessByte], which is a #[repr(transparent)] wrapper
+    // around a [Cell<u8>], which is a #[repr(transparent)] wrapper
+    // around an [UnsafeCell<u8>], which finally #[repr(transparent)]
+    // wraps a [u8]
+    core::mem::transmute::<&[u8], &ReadableProcessSlice>(
+        // Rust has very strict requirements on pointer validity[1]
+        // which also in part apply to accesses of length 0. We allow
+        // an application to supply arbitrary pointers if the buffer
+        // length is 0, but this is not allowed for Rust slices. For
+        // instance, a null pointer is _never_ valid, not even for
+        // accesses of size zero.
+        //
+        // To get a pointer which does not point to valid (allocated)
+        // memory, but is safe to construct for accesses of size zero,
+        // we must call NonNull::dangling(). The resulting pointer is
+        // guaranteed to be well-aligned and uphold the guarantees
+        // required for accesses of size zero.
+        //
+        // [1]: https://doc.rust-lang.org/core/ptr/index.html#safety
+        match len {
+            0 => core::slice::from_raw_parts(core::ptr::NonNull::<u8>::dangling().as_ptr(), 0),
+            _ => core::slice::from_raw_parts(ptr, len),
+        },
+    )
 }
 
-/// Convert an AppSlice's internal representation to a mutable Rust
-/// slice.
+/// Convert an process buffers's internal representation to a
+/// WriteableProcessSlice.
 ///
-/// This function will automatically convert zero-length appslices
-/// into valid zero-sized Rust slices regardless of the value of
-/// `ptr`.
+/// This function will automatically convert zero-length process
+/// buffers into valid zero-sized Rust slices regardless of the value
+/// of `ptr`.
 ///
 /// # Safety requirements
 ///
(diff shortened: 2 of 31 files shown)


lschuermann:
Opening this draft PR even though the CI currently fails because of the CRC module in caspules being disabled.

It'd be great to get feedback, in particular about the [new proposed `mem.rs` version](https://github.com/tock/tock/blob/c7d39661c7dd1b638ba14c45aa39b5b333e38cca/kernel/src/mem.rs). This should be very close to what we have discussed on 2021-06-18, including switching to the `map_or` closure approach also used with `AppSlices` and the naming. I'll fix the CRC HIL and implementations in the next few days.

lschuermann:
Requested changes:

- `ProcessBuf` -> `ProcessBuffer`
- `map_or` -> `enter`

kernel/src/mem.rs
```diff
@@ -8,19 +8,23 @@
 //! allow_read_only() system calls.
 //!
 //! A read-write and read-only call is mapped to the high-level Rust
-//! types [`ReadWriteAppSlice`] and [`ReadOnlyAppSlice`]
+//! types [`ReadWriteProcessBuf`] and [`ReadOnlyProcessBuf`]
 //! respectively. The memory regions can be accessed through the
-//! [`Read`] and [`ReadWrite`] traits, implemented on the
-//! AppSlice-structs.
+//! [`ReadableProcessBuf`] and [`WriteableProcessBuf`] traits,
+//! implemented on the process buffer structs.
```
ppannuto, line 14:
```suggestion
//! implemented on the process buffer structs.
//!
//! Each access to the buffer structs requires a liveness check to ensure that the
//! process memory is still valid. For a more traditional interface, users can convert
//! buffers into [`ReadableProcessSlice`] or [`WriteableProcessSlice`] and use these
//! for the lifetime of their operations. Users cannot hold live-lived references to
//! these slices, however.
```

kernel/src/mem.rs
```diff
@@ -69,213 +82,199 @@ unsafe fn raw_appslice_to_slice<'a>(ptr: *const u8, len: usize) -> &'a [u8] {
...
+/// Read-only buffer shared by a userspace process
+///
+/// This struct is provided to capsules when a process Allows a
+/// particular section of its memory to the kernel and gives the
+/// kernel read access to this memory.
+///
+/// It can be used to obtain a [`ReadableProcessSlice`], which is
+/// based around a slice of [`Cell`]s. This is because a userspace can
+/// Allow overlapping sections of memory into different
+/// [`ReadableProcessSlice`]. Having at least one mutable Rust slice
```
ppannuto, line 196:
```suggestion
/// allow overlapping sections of memory into different
/// [`ReadableProcessSlice`]s. Having at least one mutable Rust slice
```

kernel/src/mem.rs
```diff
@@ -425,8 +453,326 @@ impl Read for ReadOnlyAppSlice {
...
+    }
+}
+
+/// Readable and accessible slice of memory of a process buffer
+///
+///
+/// The only way to obtain this struct is through a
+/// [`ReadWriteProcessBuf`] or [`ReadOnlyProcessBuf`].
+///
+/// TODO: document
```
ppannuto, line 520:
```suggestion
/// Readable and accessible slice of memory of a process buffer
///
/// The only way to obtain this struct is through a
/// [`ReadWriteProcessBuf`] or [`ReadOnlyProcessBuf`].
///
/// Slices provide a more convenient, traditional interface to process memory.
/// These slices are transient, as the underlying buffer must be checked each time
/// a slice is created. This is usually enforced by the anonymous lifetime defined
/// by the creation of the slice.
```

ppannuto (commented):
Added a few doc comments that I think capture what we talked about on the call.

bradjc:
How do we handle this todo?

https://github.com/tock/tock/blob/94172ecf293e6df598192111b87db9dbbcaad178/kernel/src/sched.rs#L1074-L1075

I'm also wondering about this sequence of events:

1. A process allows a buffer at a reasonably high memory location.
2. A process unallows that buffer, but a malicious capsule holds onto a copy of the `ProcessBuffer` via some capsule-collusion-swapping trick.
3. A hypothetical implementation of `Process` sees the unallow and permits the process to move its app brk down.
4. The process then has the kernel allocate things in the grant region.
5. The capsule that holds the old `ProcessBuffer` then uses it, and gains access to other kernel memory saved in the grant.

There may be other permutations of this type of issue, but the general idea is since the Process implementation is not consulted when `ProcessBuffer.enter()` is called, I think we need a guarantee where once a ProcessBuffer is created the underlying memory must be process-accessible for as long as that process exists, or have some convention where the process implementation knows exactly when a specific ProcessBuffer can no longer exist within the kernel and thus does not need to worry about a capsule trying to use it.

lschuermann:
@bradjc I think those are great points! Given the scenario you described it seems the most reasonably (given it essentially prevents all of these issues) to guarantee that the kernel (capsules) won't / can't swap process buffers, as with upcalls. My (@hudson-ayers our?) initial plan was to apply those same mechanisms we'd introduce with #2462 also to process buffers -- it should work analog to that.

I think this really starts to get challenging -- as you described -- when we allow moving the app brk down. If that's not the case, the only real issue is that applications could be handed the wrong buffer back. AppSlices today at least aren't `Copy`, which makes it slightly better than the situation for `Upcall`s. That being said, this is still bad, and for the reasons we discussed today as part of upcalls I think also justifies protection by the kernel.

hudson-ayers:
Agreed that our initial plan was to reuse the mechanisms for callbacks for AppSlices. It still would be nice to get those guarantees.

That said, I think that as of today (where we don't allow processes to move the app brk down), swapping AppSlices, especially with this PR, is not as bad as Callback swapping. Callback swapping can lead to the kind of infinite recursion cases that @jrvanwhy mentioned today. I don't think a rust userspace will not become unsound if capsules secretly swap an AppSlice, as userspace only regains access to the buffer that is returned by the kernel (and shared simultaneous access will not be permitted, per TRD104).

I think it is ok if we release Tock 2.0 without addressing AppSlice swapping, and then just address it alongside any proposal to allow moving an app brk down. 

Also, wouldn't moving the app brk down include adjusting the MPU boundaries such that any access by an app to those addresses would lead to a fault?

lschuermann:
@hudson-ayers I agree. I think this is resolvable after Tock 2.0, in particular we can simply state to not make that guarantee now and introduce it later, as it's governed by capsules and can't be exploited by userspace alone; hence userspace can't rely on this implementation detail. Furthermore, checks in userspace now don't hurt when protections are introduced, they perhaps simply become unnecessary.

kernel/src/mem.rs
lschuermann: I have taken care to (up to now) capitalize _allow_ explicitly, as I wanted to emphasize I was talking about Tock's mechanism and now "allow" aka. "permit". Is that even a correct thing to do in English? Happy to change it.

bradjc:
> Also, wouldn't moving the app brk down include adjusting the MPU boundaries such that any access by an app to those addresses would lead to a fault?

Yes. I think the only potential issue is violating memory safety within the kernel.

kernel/src/mem.rs
lschuermann: Thinking about it, it probably makes more sense to distinguish those through a Markdown inline-code syntax: "`allow` overlapping"

kernel/src/mem.rs
```diff
@@ -31,34 +35,43 @@ use crate::process::ProcessId;
...
+        // is safe to construct for accesses of size zero, we must call
+        // NonNull::dangling(). The resulting pointer is guaranteed to be well-aligned
+        // and uphold the guarantees required for accesses of size zero.
+        //
+        // [1]: https://doc.rust-lang.org/core/ptr/index.html#safety
+        match len {
+            0 => core::slice::from_raw_parts(core::ptr::NonNull::<u8>::dangling().as_ptr(), 0),
+            _ => core::slice::from_raw_parts(ptr, len),
+        },
+    )
```
jrvanwhy, line 72: I don't think `transmute` is needed here. I think something like the following will work:

```rust
// addr should be inferred as a *const ReadableProcessSlice
let addr = match len {
    0 => core::ptr::NonNull::dangling().as_ptr(),
    _ => ptr as *const _,
};
core::slice::from_raw_parts(addr, len)
```

The same applies to `raw_processbuf_to_rwprocessslice`.

bradjc:
Ok so today we only have `Process.build_readonly_appslice()` and `Process.build_readwrite_appslice()` APIs, so any process implementation has no way of knowing if a particular process buffer is no longer in use. If we ever added some sort of unallow API to Process then we would have to figure out how to deal with this condition, but today a correct Process implementation _must_ assume all allowed buffers ever created still exist, and therefore could never use re-use any allowed memory for the grant region.

kernel/src/mem.rs
```diff
@@ -69,213 +82,199 @@ unsafe fn raw_appslice_to_slice<'a>(ptr: *const u8, len: usize) -> &'a [u8] {
...
+/// based around a slice of [`Cell`]s. This is because a userspace can
+/// Allow overlapping sections of memory into different
+/// [`ReadableProcessSlice`]. Having at least one mutable Rust slice
+/// along with read-only slices to overlapping memory in Rust violates
+/// Rust's aliasing rules. A slice of [`Cell`]s avoids this issue by
+/// explicitly supporting interior mutability. Still, a memory barrier
+/// prior to switching to userspace is required, as the compiler is
+/// free to reorder reads and writes, even through [`Cell`]s.
+pub struct ReadOnlyProcessBuf {
+    ptr: *const u8,
```
jrvanwhy, line 209: Should `ptr` be a `core::ptr::NonNull` so that `Option<ReadOnlyProcessBuf>` benefits from niche optimizations? That would require changing the `new_external` safety invariants somewhat, but I think it would shrink grant sizes significantly.

The same applies to `ReadWriteProcessBuf`.

lschuermann: I suppose the issue with this is that the process is free to Allow a buffer with a null-pointer, so long as the name is zero. Now, that is never valid as a Rust slice reference, which is why were using `NonNull::dangling()` in this case, but we're going to need to provide the process' Allow data back unmodified, which is why presumably we can't use `NonNull` here.

However, `struct ProcessId` starts with a `&'static Kernel`, which is guaranteed to never be null. By making this the first field, we should still benefit from these optimizations, right?

jrvanwhy: Oh, right, `ReadOnlyProcessBuf`/`ReadWriteProcessBuf` are used to return the correct slice back to userspace, not just to access it from capsules.

lschuermann:
I have addressed the requested changes, namely
- renaming `ProcessBuf` to `ProcessBuffer` and
- changing the access-methods on the process buffers to be named `enter` and `mut_enter`, and use the well-established `Grant::enter` semantics.

In general I think that this is a step in the right direction. Returning a `Result<R, process::Error>` on a `ProcessBuffer::{mut_,}enter` makes enables chaining of Results through methods such as `map_err` and makes Rust complain about unhandled `Result`s. However, I suppose some of the capsule code would have been written differently if these semantics were available from the start. Given all of the mechanical changes over the course of developing Tock 2.0, it might make sense to take a closer look at individual capsules and fix some code which is no longer idiomatic Rust.

I've taken great care that I don't change any functionality of any capsule in this PR, such that the changes are easy to verify we can focus the review efforts on `mem.rs`.

kernel/src/mem.rs
lschuermann: I'm afraid I don't quite understand. If we were to pass in a `*const ReadableProcessSlice` in `core::slice::from_raw_parts`, wouldn't we get a slice of `ReadableProcessSlice`s? This is also what the compiler appears to do:

```
  --> kernel/src/mem.rs:73:2
   |
47 | ) -> &'a ReadableProcessSlice {
   |      ------------------------ expected `&'a ReadableProcessSlice` because of return type
...
73 |     core::slice::from_raw_parts(addr, len)
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `ReadableProcessSlice`, found slice
   |
   = note: expected reference `&'a ReadableProcessSlice`
              found reference `&[_]`
```

I think we need to use `transmute` here, as we have a proper Rust slice and want to convert it to a type which -- under the hood -- is still a proper Rust slice, but is a newtype and therefore can't be treated as one.

jrvanwhy: Oh, I see what's happening. Yes, you're right, I think you need `transmute` here.

hudson-ayers:
I think this can be marked as ready-for-review if the only blocker is the CRC update

kernel/src/syscall.rs
```diff
@@ -222,7 +222,7 @@ pub enum SyscallReturnVariant {
 /// method.
 ///
 /// Capsules do not use this struct. Capsules use higher level Rust types
-/// (e.g. [`ReadWriteAppSlice`](crate::ReadWriteAppSlice) and
+/// (e.g. [`ReadWriteProcessBuf`](crate::ReadWriteProcessBuf) and
```
bradjc, line 225:
```suggestion
/// (e.g. [`ReadWriteProcessBuffer`](crate::ReadWriteProcessBuffer) and
```

kernel/src/process_standard.rs
```diff
@@ -612,14 +612,14 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
             // memory (verified using `in_app_owned_memory` or
             // `in_app_flash_memory`) and respect alignment and other
             // constraints of the Rust references created by
-            // ReadWriteAppSlice.
+            // ReadWriteProcessBuf.
```
bradjc, line 615:
```suggestion
            // ReadWriteProcessBuffer.
```

kernel/src/process_standard.rs
```diff
@@ -612,14 +612,14 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
...
             //
             // ### Safety
             //
             // We encapsulate the unsafe here on the condition in the TODO
-            // above, as we must ensure that this `ReadOnlyAppSlice` will be
+            // above, as we must ensure that this `ReadOnlyProcessBuf` will be
```
bradjc, line 620:
```suggestion
            // above, as we must ensure that this `ReadOnlyProcessBuffer` will be
```

kernel/src/process_standard.rs
```diff
@@ -584,9 +584,9 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
             // ### Safety
             //
             // We specific a zero-length buffer, so the implementation of
-            // `ReadOnlyAppSlice` will handle any safety issues. Therefore, we
+            // `ReadOnlyProcessBuf` will handle any safety issues. Therefore, we
```
bradjc, line 587:
```suggestion
            // `ReadOnlyProcessBuffer` will handle any safety issues. Therefore, we
```

kernel/src/process_standard.rs
```diff
@@ -540,14 +540,14 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
...
             // memory (verified using `in_app_owned_memory`) and
             // respect alignment and other constraints of the Rust
-            // references created by ReadWriteAppSlice.
+            // references created by ReadWriteProcessBuf.
             //
             // ### Safety
             //
             // We encapsulate the unsafe here on the condition in the TODO
-            // above, as we must ensure that this `ReadWriteAppSlice` will be
+            // above, as we must ensure that this `ReadWriteProcessBuf` will be
```
bradjc, line 548:
```suggestion
            // above, as we must ensure that this `ReadWriteProcessBuffer` will be
```

kernel/src/process_standard.rs
```diff
@@ -540,14 +540,14 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
             // we make sure that we're pointing towards userspace
             // memory (verified using `in_app_owned_memory`) and
             // respect alignment and other constraints of the Rust
-            // references created by ReadWriteAppSlice.
+            // references created by ReadWriteProcessBuf.
```
bradjc, line 543:
```suggestion
            // references created by ReadWriteProcessBuffer.
```

kernel/src/process_standard.rs
```diff
@@ -515,9 +515,9 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
             // ### Safety
             //
             // We specific a zero-length buffer, so the implementation of
-            // `ReadWriteAppSlice` will handle any safety issues. Therefore, we
+            // `ReadWriteProcessBuf` will handle any safety issues. Therefore, we
```
bradjc, line 518:
```suggestion
            // `ReadWriteProcessBuffer` will handle any safety issues. Therefore, we
```

kernel/src/mem.rs
```diff
@@ -69,213 +90,205 @@ unsafe fn raw_appslice_to_slice<'a>(ptr: *const u8, len: usize) -> &'a [u8] {
 /// within a single process' address space for the entire lifetime
 /// `'a`. No other mutable or immutable slice pointing to an
 /// overlapping memory region may exist over the entire lifetime `'a`.
```
bradjc, line 92: This is no longer true, correct?

kernel/src/mem.rs
```diff
@@ -413,20 +463,359 @@ impl Read for ReadOnlyAppSlice {
...
+                .kernel
+                .process_map_or(Err(process::Error::NoSuchApp), pid, |_| {
+                    // Safety: `kernel.process_map_or()` validates that
+                    // the process still exists and its memory is still
+                    // valid. `Process` tracks the "high water mark" of
+                    // memory that the process has `allow`ed to the
+                    // kernel, and will not permit the process to free any
+                    // memory after it has been `allow`ed. This guarantees
+                    // that the buffer is safe to convert into a slice
+                    // here.
```
bradjc, line 482:
```suggestion
                    // Safety: `kernel.process_map_or()` validates that
                    // the process still exists and its memory is still
                    // valid. Once a `ProcessBuffer` is created in a process
                    // there is no API for removing it, so any `Process`
                    // implementation must assume the ProcessBuffer always
                    // exists and will not permit the process to free any
                    // memory after it has been `allow`ed. This guarantees
                    // that the buffer is safe to convert into a slice
                    // here.
```

kernel/src/mem.rs
```diff
@@ -413,20 +463,359 @@ impl Read for ReadOnlyAppSlice {
...
```
bradjc, line 508:
```suggestion
                    // Safety: `kernel.process_map_or()` validates that
                    // the process still exists and its memory is still
                    // valid. Once a `ProcessBuffer` is created in a process
                    // there is no API for removing it, so any `Process`
                    // implementation must assume the ProcessBuffer always
                    // exists and will not permit the process to free any
                    // memory after it has been `allow`ed. This guarantees
                    // that the buffer is safe to convert into a slice
                    // here.
```

kernel/src/mem.rs
```diff
@@ -289,117 +302,154 @@ impl Read for ReadWriteAppSlice {
...
```
bradjc, line 321:
```suggestion
                    // Safety: `kernel.process_map_or()` validates that
                    // the process still exists and its memory is still
                    // valid. Once a `ProcessBuffer` is created in a process
                    // there is no API for removing it, so any `Process`
                    // implementation must assume the ProcessBuffer always
                    // exists and will not permit the process to free any
                    // memory after it has been `allow`ed. This guarantees
                    // that the buffer is safe to convert into a slice
                    // here.
```

kernel/src/driver.rs
```diff
@@ -201,21 +201,21 @@ pub trait Driver {
         CommandReturn::failure(ErrorCode::NOSUPPORT)
     }
 
-    /// System call for a process to pass a buffer (a ReadWriteAppSlice) to
+    /// System call for a process to pass a buffer (a ReadWriteProcessBuf) to
```
bradjc, line 204:
```suggestion
    /// System call for a process to pass a buffer (a ReadWriteProcessBuffer) to
```

kernel/src/driver.rs
```diff
@@ -201,21 +201,21 @@ pub trait Driver {
...
-        slice: ReadWriteAppSlice,
-    ) -> Result<ReadWriteAppSlice, (ReadWriteAppSlice, ErrorCode)> {
+        slice: ReadWriteProcessBuffer,
+    ) -> Result<ReadWriteProcessBuffer, (ReadWriteProcessBuffer, ErrorCode)> {
         Err((slice, ErrorCode::NOSUPPORT))
     }
 
     /// System call for a process to pass a read-only buffer (a
-    /// ReadOnlyAppSlice) to the kernel that the kernel can read.
+    /// ReadOnlyProcessBuf) to the kernel that the kernel can read.
```
bradjc, line 218:
```suggestion
    /// ReadOnlyProcessBuffer) to the kernel that the kernel can read.
```

kernel/src/mem.rs
hudson-ayers: I believe Brad is correct. It is still true that no mutable slice (i.e. `&mut [u8]`) may exist pointing to a portion of the memory region, but an immutable slice pointing to that region should be fine?

lschuermann: Yes, this comment is outdated, as it was intended to say that no other mutable or immutable _AppSlice_ (now ProcessBuffer) must exist while this is in scope. This is not the case any longer.

However, viewed from @hudson-ayers's perspective (when talking about Rust primitive slices of `u8`), this is still correct, both for immutable and mutable slices. This is the case for all process memory. When another immutable or mutable slice over said memory region is in scope, both the kernel changing memory through a slice of `Cell`s, as well as the process changing this memory will lead to unsoundness in the kernel.

I'll update the comment to incorporate these required invariants more precisely.

kernel/src/mem.rs
lschuermann: I understand the motivation behind this change, and this seems correct to me. However, I fear that the rephrased comment might be more confusing given it does not clearly express the relation between these safety considerations and the process' high water mark any longer. Specifically, I think that "must assume the ProcessBuffer always exists" could be confusing as, from a capsule developer perspective, this struct is passed back to the kernel on a call to `allow_read{only,write}`. Do you mind if I try rephrasing?

bradjc: Yeah I have no attachment to this text. What I would like to avoid is tying the safety of process buffers to a particular implementation detail of `trait Process` as there is no guarantee that process buffers will only be used with ProcessStandard.

kernel/src/syscall.rs
```diff
@@ -222,8 +222,8 @@ pub enum SyscallReturnVariant {
 /// method.
 ///
 /// Capsules do not use this struct. Capsules use higher level Rust types
-/// (e.g. [`ReadWriteAppSlice`](crate::ReadWriteAppSlice) and
-/// `UpcallMemory`) or wrappers around this struct
+/// (e.g. [`ReadWriteProcessBuffer`](crate::ReadWriteProcessBuffer) and
+/// [`UpcallMemory`]) or wrappers around this struct
```
bradjc, line 226: Unfortunately I don't think this will work as UpcallData is not public.

kernel/src/process.rs
```diff
@@ -317,13 +317,13 @@ pub trait Process {
     /// process memory brk.
     fn app_memory_break(&self) -> *const u8;
 
-    /// Creates a `ReadWriteAppSlice` from the given offset and size
+    /// Creates a `ReadWriteProcessBuf` from the given offset and size
```
bradjc, line 320:
```suggestion
    /// Creates a `ReadWriteProcessBuffer` from the given offset and size
```

kernel/src/process.rs
```diff
@@ -317,13 +317,13 @@ pub trait Process {
...
     /// in process memory.
     ///
     /// ## Returns
     ///
     /// In case of success, this method returns the created
-    /// [`ReadWriteAppSlice`].
+    /// [`ReadWriteProcessBuf`].
```
bradjc, line 326:
```suggestion
    /// [`ReadWriteProcessBuffer`].
```

kernel/src/process.rs
```diff
@@ -337,15 +337,15 @@ pub trait Process {
         &self,
         buf_start_addr: *mut u8,
         size: usize,
-    ) -> Result<ReadWriteAppSlice, ErrorCode>;
+    ) -> Result<ReadWriteProcessBuffer, ErrorCode>;
 
-    /// Creates a [`ReadOnlyAppSlice`] from the given offset and size
+    /// Creates a [`ReadOnlyProcessBuf`] from the given offset and size
```
bradjc, line 342:
```suggestion
    /// Creates a [`ReadOnlyProcessBuffer`] from the given offset and size
```

kernel/src/process.rs
```diff
@@ -337,15 +337,15 @@ pub trait Process {
...
     /// in process memory.
     ///
     /// ## Returns
     ///
     /// In case of success, this method returns the created
-    /// [`ReadOnlyAppSlice`].
+    /// [`ReadOnlyProcessBuf`].
```
bradjc, line 348:
```suggestion
    /// [`ReadOnlyProcessBuffer`].
```

kernel/src/syscall.rs
lschuermann: Right, I just realized.

lschuermann: Can't we just reference the `GrantUpcallTable` type here? After all, that's what capsules are confronted with?

bradjc: Yeah that's the type I meant. I tried before and it didn't work, but I could have missed something or it could have changed.

lschuermann: Oh, sorry. I was confused, as we can't have a private struct in a public interface, but there's nothing stopping us from just not exporting that.

I suppose this leads to a different question, which is perhaps best answered in #2551: should we export these types? It makes cross-referencing them in documentation impossible, and more importantly does not allow the user to introspect their public API. For example, take a look at the `Grant::enter` method [as it is currently documented](https://docs.tockos.org/kernel/struct.grant#method.enter). It's a bummer I can't click on the `GrantUpcallTable` and see that I have to call it's `schedule_upcall` method, but instead have to look into the source.

bradjc: Ack Rust tools are always fighting against safety! The problem is making things public can lead to soundness issues (which used to be a much bigger issue than it is today, and is what lead to Capabilities) and it's often hard to reason about 1) why a thing in kernel/ is public at all and 2) if it is public is it safe (or not).

I also don't understand why this would be considered a tock bug and not a cargo doc bug. If code in a different crate can use the api just fine, then why shouldn't cargo doc document it.

lschuermann: Agreed. Will ask Rust folks or open an issue.

phil-levis:
I have a compiling version of CRC in https://github.com/lschuermann/tock/tree/dev/crc-hil-redesign

I'm going to start testing and debugging it now. But now that it's compiling I think it's fine to move forward.

bradjc (approved):

hudson-ayers (approved):

kernel/src/mem.rs
```diff
@@ -289,117 +315,162 @@ impl Read for ReadWriteAppSlice {
...
+		    // called once a `ProcessBuffer` has been passed back into
+		    // the kernel, a given `Process` implementation must assume
+		    // that the memory described by a once-allowed
+		    // `ProcessBuffer` is still in use, and thus will not
+		    // permit the process to free any memory after it has
+		    // been `allow`ed to the kernel once. This guarantees
+                    // that the buffer is safe to convert into a slice
+                    // here. For more information, refer to the
+		    // comment and subsequent discussion on tock/tock#2632:
+		    // https://github.com/tock/tock/pull/2632#issuecomment-869974365
```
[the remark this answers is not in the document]
lschuermann: Yeah, this is weird. Once #2636 is upstream, I'll rebase these changes and take a final look to make sure the formatting works out. Sometimes it seems rustfmt just doesn't touch certain code sections, whereas others it formats a little but produces output as seen here.

kernel/src/sched.rs
```diff
@@ -1183,46 +1180,44 @@ impl Kernel {
...
+                                    ro_pbuf,
                                 ) {
-                                    Ok(returned_appslice) => {
+                                    Ok(returned_pbuf) => {
                                         // The capsule has accepted the allow
                                         // operation. Pass the previous buffer
                                         // information back to the process.
                                         //
-                                        // TODO: Prevent swapping of AppSlices by
+                                        // TODO: Prevent swapping of ProcessBuffers by
```
bradjc, line 1200: What is the status of these TODOs?

kernel/src/mem.rs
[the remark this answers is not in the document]
lschuermann: Should be resolved.

kernel/src/sched.rs
lschuermann: Citing @hudson-ayers 

> I think it is ok if we release Tock 2.0 without addressing AppSlice swapping, and then just address it alongside any proposal to allow moving an app brk down.

I'd still say that we might want to address this at some point, when convenient, possibly when we arrive at a `trait Process` implementation which supports moving `app_brk` down. Hence, I'd vouch to deliberately leave this in.

lschuermann:
I've rebased on latest master and went through the changes again. It's easy to slip up in these mass refactors, but to my eyes this looks good so far.

I suppose this is a final chance to answer some remaining questions, just to be sure we've settled on the correct design:
- we do use `mem::transmute` here to convert between types. As far as I understand, what we're doing here is entirely sound. In particular, I'm pretty sure we can treat a `#[repr(transparent)]` type over a slice as just a slice. However, I'd also be open to asking some Rust folks prior to releasing 2.0 whether what we're doing here is actually properly sound. After all, this doesn't effect only some single capsule but all drivers.
- this will -- by design -- allow multiple "mutable borrows" of a mutable ProcessBuffer, something which wasn't possible before. We _could_ require a `&mut` reference in `mut_enter`, but there's  no technical reason to do so. It might have the advantage of having only a single mutable borrow in scope, so the cannot be two mutable references to _the same_ ProcessBuffer at the same time. However, a reference to _a different_ ProcessBuffer can still influence others, as they can overlap. Therefore, I think this is fine.
- this uses `{mut_,}enter`, returning a `Result`. I think this works out nice, especially in new code, and makes error handling more idiomatic. Now is the last chance to nitpick whether it should really be called `enter_mut`! :smile:

kernel/src/sched.rs
bradjc: Gotcha, do we have a decent reference we can point to about that todo, like an issue? I'm worried we'll forget what the intent is and why we didn't do it for 2.0.

capsules/diffd
```diff
@@ -0,0 +1,1164 @@
+diff --git a/boards/components/src/crc.rs b/boards/components/src/crc.rs
```
bradjc, line 1: I assume this wasn't intentional.

lschuermann: Lesson learned: keep your repository clean. Thanks @bradjc.

bradjc (dismissed):

kernel/src/sched.rs
lschuermann: I'll point to this issue, specifically your comment about this? I think that's a pretty decent reference.

lschuermann: Done. Feel free to resolve if this works for you.

bradjc (approved):

kernel/src/sched.rs
lschuermann: As discussed on the call, I'll remove this comment and create a feature-request issue, plus PR an update to TRD104.

hudson-ayers:
Why is this now "blocked"?

lschuermann:
> Why is this now "blocked"?

I wanted to remove the TODO comments as discussed on the call, but couldn't on mobile. Not sure whether the "blocked" tag is the right thing to use here, but I wanted to mark that there is still one thing missing.

bradjc (dismissed):

hudson-ayers (approved):

bradjc:
Arg needs rebase!

lschuermann:
Will do.

bradjc (approved):

bradjc:
Ok! Let's make this happen.

bors r+

lschuermann:
Tagging @ppannuto. Is there any reason why Netlify doesn't report a status?

lschuermann:
In the meantime, trying one more time:

bors r+

ppannuto:
https://www.netlifystatus.com/

looks like netlify is down; relying on local testing of `alldoc` and moving forward with manual merge

lschuermann:
Thanks @ppannuto!

PR #2636, opened 2021-06-29
Redesign CRC HIL & fix SAM4L implementation

lschuermann:
### Pull Request Overview

The current version of the CRC HIL fostered unsound implementations, in that it relies on asynchronous processing of input data, but only takes a read-only reference with an anonymous lifetime of the respective data buffer. When this buffer is used for a DMA operation, it may already have gone out of scope (for instance, when allocating on the stack). Therefore, this PR redesigns the CRC HIL interface in accordance with other current HILs and best practices as documented in the TRD for HIL design.

This is, perhaps most importantly, a prerequisite for trivially porting the changes of #2632 to the CRC driver.

### Testing Strategy

This pull request was not tested. It's not ready yet, only the HIL is done. I'd appreciate people testing with SAM4L boards as soon as this is ready.

### TODO or Help Wanted

This pull request still needs
- [x] porting the HIL changes to the SAM4L code
- [ ] porting the HIL changes to the capsule

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/boards/components/src/crc.rs
+++ b/boards/components/src/crc.rs
@@ -1,17 +1,18 @@
-//! Component for CRC syscall interface.
+//! Component for Crc syscall interface.
 //!
 //! This provides one Component, `CrcComponent`, which implements a
-//! userspace syscall interface to the CRC peripheral.
+//! userspace syscall interface to the Crc peripheral.
 //!
 //! Usage
 //! -----
 //! ```rust
-//! let crc = components::crc::CrcComponent::new(board_kernel, &sam4l::crccu::CRCCU)
+//! let crc = components::crc::CrcComponent::new(board_kernel, &sam4l::crccu::CrcCU)
 //!     .finalize(components::crc_component_helper!(sam4l::crccu::Crccu));
 //! ```
 
 // Author: Philip Levis <pal@cs.stanford.edu>
-// Last modified: 6/20/2018
+// Author: Leon Schuermann  <leon@is.currently.online>
+// Last modified: 6/2/2021
 
 use core::mem::MaybeUninit;
 
@@ -19,8 +20,8 @@ use capsules::crc;
 use kernel::capabilities;
 use kernel::component::Component;
 use kernel::create_capability;
-use kernel::hil;
-use kernel::static_init_half;
+use kernel::hil::crc::Crc;
+use kernel::{static_init, static_init_half};
 
 // Setup static space for the objects.
 #[macro_export]
@@ -28,18 +29,18 @@ macro_rules! crc_component_helper {
     ($C:ty $(,)?) => {{
         use capsules::crc;
         use core::mem::MaybeUninit;
-        static mut BUF: MaybeUninit<crc::Crc<'static, $C>> = MaybeUninit::uninit();
+        static mut BUF: MaybeUninit<crc::CrcDriver<'static, $C>> = MaybeUninit::uninit();
         &mut BUF
     };};
 }
 
-pub struct CrcComponent<C: 'static + hil::crc::CRC<'static>> {
+pub struct CrcComponent<C: 'static + Crc<'static>> {
     board_kernel: &'static kernel::Kernel,
     driver_num: usize,
     crc: &'static C,
 }
 
-impl<C: 'static + hil::crc::CRC<'static>> CrcComponent<C> {
+impl<C: 'static + Crc<'static>> CrcComponent<C> {
     pub fn new(
         board_kernel: &'static kernel::Kernel,
         driver_num: usize,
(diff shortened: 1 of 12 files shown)


lschuermann:
Question: We currently support 5 CRC algorithms:

- Crc32, polynomial 0x04C11DB7, output reversed then inverted ("CRC-32")
- Crc32C, polynomial 0x1EDC6F41, output reversed then inverted ("CRC-32C" / "Castagnoli")
- Sam4L16, polynomial 0x1021, no output post-processing
- Sam4L32, polynomial 0x04C11DB7, no output post-processing
- Sam4L32C, polynomial 0x1EDC6F41, no output post-processing

The utility of the first two is obvious. Third (Sam4L16) I don't really get. I don't have the hardware to test, but concerning the manual this really looks like a regular CRC16-CCITT (might need to be reversed though)? Finally, given the last two are just essentially a result of how the CRC engine is implemented in the SAM4L (by not doing the inversion in hardware), I wonder what the utility of exposing those "functions" is, considering they really seem to be entirely hardware-specific.

I suppose this is a more general question which boils down to: do we want HILs to reflect specific characteristics and features of certain hardware, or just support the established standards? A blanket statement might be difficult, though a rough indication can help.

ppannuto:
Issues along these lines have come up before, and I think our answer has been pretty consistently that things in HILs should be things that can be reasonably expected to be implemented by a variety of hardware. If these `SAM4L*` things are truly chip-specific, then they should be public exports of the SAM4L chip implementation, but probably not part of the HIL.

If some of them are (or are near) a CRC standard, we should rename the HIL to match the standard, and possibly add the reversing to SAM4L implementation if that's missing presently.

bradjc:
I don't see why we would have literal "Sam4l" modes in the HIL.

Also, the CRC HIL touches very little code (it has only ever been implemented once). I say we move quickly to get this in to make process on process buf aliasing.

lschuermann:
Agreed. I have the capsule almost ready and will push either today or tomorrow. I wanted to do this "right", but we shouldn't block for longer than the next two days on this.

lschuermann:
Hm, I can push what I have now but the capsule changes are pretty straightforward, I just need to get them done. Give me another few minutes.

phil-levis:
> Hm, I can push what I have now but the capsule changes are pretty straightforward, I just need to get them done. Give me another few minutes.
> 
> Edit: @phil-levis I've just pushed my WIP capsule code. Feel free to look at it, but I can also offer to get it done tomorrow. Unfortunately won't meet the today-deadline.

OK, got it. I had started working on this, actually. It turns out our edits were nearly identical except I was putting logic in `input_done` and you were doing `crc_done`.

I'll work on this over the weekend. I'm flying tomorrow, if there's enough space on the plane I'll work on it then.

lschuermann:
@phil-levis Awesome, thanks for working on this! Glad to hear the CRC tests pass, I knew there must've been some things in that implementation which won't work on the real hardware. :smile: 

There's still one small discrepancy in the way I imagined the HIL and the SAM4L implementation works, and the CRC capsule:
https://github.com/tock/tock/blob/5c6c6ab4e0520ec1270292a550ac37a12cd3992e/capsules/src/crc.rs#L422-L426

This code assumes that if the returned leasable buffer is of length 0, the operation completed. This effectively means that the CRC would need to reslice the buffer such that it contains the remaining bytes. However, that would make imposing a length limit on the passed buffer through the LeasableBuffer impossible. Instead, the SAM4L implementation takes a leasable buffer, reslices it if it was too large and passes that resliced buffer back.

Nonethless, as far as I can tell this implementation is at least sound, so I'd be okay with merging now and fixing the capsule prior to the 2.0 release, possibly during testing.

hudson-ayers:
+1 for merging a version of CRC that compiles, and documenting the remaining bugs in an issue that can be fixed later

lschuermann:
@hudson-ayers Currently rebasing, will document at the top of the CRC capsule that this may or may not work in its current state.

bradjc (dismissed):

lschuermann:
Now that one can see the Netlify errors it feels like Rustdoc is much more difficult to get right. :smile:

hudson-ayers (approved):

hudson-ayers:
Update from Phil:

>I have everything fully working (system call driver, tests, etc). I need to test some more edge cases (large userspace buffers) but the test_crc works. I couldn’t push due to Internet issues will push later tonight. Will also add tests for large buffers. The tricky bit is there are two layers of sequencing: one is user buffers onto kernel buffers, the other is kernel buffers onto the maximum 64 byte operation through the CRCCU.

hudson-ayers:
bors r+

@phil-levis Hopefully this being merged is not too inconvenient, and your remaining changes can be submitted as a follow up PR. This will allow us to try to get #2632 done by the meeting tomorrow.

PR #2638, opened 2021-06-29
Rename Driver trait parameters for consistency

jettr:
Use `process_id` instead of app or app_id
Use `which` for the id parameter for all three functions

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/driver.rs
+++ b/kernel/src/driver.rs
@@ -194,7 +194,7 @@ pub trait Driver {
     /// within memory the process can both read and write.
     fn allow_readwrite(
         &self,
-        app: ProcessId,
+        process_id: ProcessId,
         which: usize,
         slice: ReadWriteProcessBuffer,
     ) -> Result<ReadWriteProcessBuffer, (ReadWriteProcessBuffer, ErrorCode)> {
@@ -209,7 +209,7 @@ pub trait Driver {
     /// read-only data (e.g., in flash) to the kernel.
     fn allow_readonly(
         &self,
-        app: ProcessId,
+        process_id: ProcessId,
         which: usize,
         slice: ReadOnlyProcessBuffer,
     ) -> Result<ReadOnlyProcessBuffer, (ReadOnlyProcessBuffer, ErrorCode)> {
--- a/kernel/src/driver.rs
+++ b/kernel/src/driver.rs
@@ -184,7 +184,13 @@ pub trait Driver {
     /// is signaled with an upcall). Command 0 is a reserved command to
     /// detect if a peripheral system call driver is installed and must
     /// always return a CommandReturn::Success.
-    fn command(&self, which: usize, r2: usize, r3: usize, caller_id: ProcessId) -> CommandReturn {
+    fn command(
+        &self,
+        command_num: usize,
+        r2: usize,
+        r3: usize,
+        caller_id: ProcessId,
+    ) -> CommandReturn {
         CommandReturn::failure(ErrorCode::NOSUPPORT)
     }
 
@@ -195,7 +201,7 @@ pub trait Driver {
     fn allow_readwrite(
         &self,
         process_id: ProcessId,
-        which: usize,
+        allow_num: usize,
         slice: ReadWriteProcessBuffer,
     ) -> Result<ReadWriteProcessBuffer, (ReadWriteProcessBuffer, ErrorCode)> {
         Err((slice, ErrorCode::NOSUPPORT))
@@ -210,7 +216,7 @@ pub trait Driver {
     fn allow_readonly(
         &self,
         process_id: ProcessId,
-        which: usize,
+        allow_num: usize,
         slice: ReadOnlyProcessBuffer,
     ) -> Result<ReadOnlyProcessBuffer, (ReadOnlyProcessBuffer, ErrorCode)> {
         Err((slice, ErrorCode::NOSUPPORT))
@@ -305,5 +311,5 @@ pub trait Driver {
     // mechanism may find more uses in the future if the kernel needs to store
     // additional state on a per-driver basis and therefore needs a mechanism to
     // force a grant allocation.
-    fn allocate_grant(&self, appid: ProcessId) -> Result<(), crate::process::Error>;
+    fn allocate_grant(&self, process_id: ProcessId) -> Result<(), crate::process::Error>;
 }
--- a/kernel/src/driver.rs
+++ b/kernel/src/driver.rs
@@ -190,7 +190,7 @@ pub trait Driver {
         command_num: usize,
         r2: usize,
         r3: usize,
-        caller_id: ProcessId,
+        process_id: ProcessId,
     ) -> CommandReturn {
         CommandReturn::failure(ErrorCode::NOSUPPORT)
     }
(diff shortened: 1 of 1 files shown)


lschuermann:
I think this is a rather uncontroversial set of changes. What bugs me a lot is that we have the order inconsistent, though that might be solved by the Upcall swap prevention mechanism we're currently working on.

One question: do you think `which` is a good choice here? It's probably sufficiently clear within the method signature, but we've been talking about either a `{subscribe,allow} numbers` or `subdriver numbers`, as this terminology also carries meaning without context.

bradjc:
We should hold off on the change to subscribe() until we resolve the callback swapping PRs.

hudson-ayers (dismissed):
The ordering of `which` vs. `process_id` being different for subscribe vs. allow is also painful and would be nice to fix..

Agreed this should wait until after the other PRs touching all the capsules

jettr:
I also wanted to change the order of `which` and `process_id`, too! I can go through and make that larger clean up as well if we are all in agreement of the following order:

`which`/`id`/`xxx_id`,
other parameters.
`process_id`

As for the name of `which`, I prefer just `id` to `which`, but `subscribe_id`, `allow_id`, and `cmd_id` work for me as well.

I am guessing you are talking waiting for #2462 to be resolved first, then land these clean up changes, right?

alexandruradovici:
> I also wanted to change the order of `which` and `process_id`, too! I can go through and make that larger clean up as well if we are all in agreement of the following order:
> 
> `which`/`id`/`xxx_id`,
> other parameters.
> `process_id`
> 
> As for the name of `which`, I prefer just `id` to `which`, but `subscribe_id`, `allow_id`, and `cmd_id` work for me as well.
> 
As the docs call them `numbers`, shouldn't we maybe rename them to `subscribe_number`, `allow_number` and `command_number` or the shorter version `subscribe_num`, etc?

> I am guessing you are talking waiting for #2462 to be resolved first, then land these clean up changes, right?

jettr:
`subscribe_num`, `allow_num`, and `cmd_num` sound good to me.

hudson-ayers:
> I am guessing you are talking waiting for #2462 to be resolved first, then land these clean up changes, right?

Yeah, either #2462 or an upcoming PR based on the WIP https://github.com/tock/tock/tree/kernel-grant-upcalls-2 (that WIP is the reason #2462 has not been merged, as it has a number of possible advantages over the approach in #2462)

ppannuto:
> subscribe_num, allow_num, and cmd_num sound good to me.

If we're going to rename, I think we should rename to full names, i.e. `subscribe_number`, `allow_number`, `command_number`

jrvanwhy:
TRD 104 uses "identifier" to describe these numbers in several places, so what about `subscribe_id`, `ro_allow_id`, `rw_allow_id`, `command_id`?

The advantage relative to the `_number` names is it would reduce the amount of line wrapping, which helps with readability.

jettr:
Should `ro_allow_` and `rw_allow_` both be `allow_`? Do they share the same number space? Can I share the same logical buffer as RO or RW depending on how the command needs to use it?

Having both `ro_allow_` and `rw_allow_` makes me thing that the value of 2, for example, for each would represent two different logical buffers instead of the same logical buffer, just shared differently.

jrvanwhy:
> Should `ro_allow_` and `rw_allow_` both be `allow_`? Do they share the same number space?

No, they are separate number spaces. That allows `libtock-rs` to have an API that looks like (ignoring error handling):

```rust
fn ro_allow(driver_id: u32, ro_allow_id: u32, new_buffer: &[u8]) -> &[u8];
fn rw_allow(driver_id: u32, rw_allow_id: u32, new_buffer: &mut [u8]) -> &mut [u8];
```

without providing a way to transmute a `&[u8]` into a `&mut [u8]`, which allows them to be safe.

bradjc:
> TRD 104 uses "identifier" to describe these numbers in several places, so what about `subscribe_id`, `ro_allow_id`, `rw_allow_id`, `command_id`?

In the kernel code, we use "identifier" to denote a specific upcall/process buffer/command, that is the tuple of (driver_num, subscribe_num/allow_num/command_num). "number" is then for each of the individual fields. For example:

https://github.com/tock/tock/blob/ab5bd71d1d19f12e877df1c5342808e08a935ac3/kernel/src/upcall.rs#L12-L19

phil-levis:
On a tangent, I think at some point we will want to clean up our use of ProcessId. The issue is that it's more than an Id: it's a struct with 3 fields (12 bytes): the actual ID, the index into the process array, and a reference to the kernel. There are lots of cases when we just want to store/use the ID, but end up storing all three.

phil-levis:
> > TRD 104 uses "identifier" to describe these numbers in several places, so what about `subscribe_id`, `ro_allow_id`, `rw_allow_id`, `command_id`?
> 
> In the kernel code, we use "identifier" to denote a specific upcall/process buffer/command, that is the tuple of (driver_num, subscribe_num/allow_num/command_num). "number" is then for each of the individual fields. For example:
> 
> https://github.com/tock/tock/blob/ab5bd71d1d19f12e877df1c5342808e08a935ac3/kernel/src/upcall.rs#L12-L19

Let's change the 104 text to agree with this terminology, then.

bradjc:
> On a tangent, I think at some point we will want to clean up our use of ProcessId. The issue is that it's more than an Id: it's a struct with 3 fields (12 bytes): the actual process number (a process ID in POSIX), the index into the process array, and a reference to the kernel. There are lots of cases when we just want to store/use the number, but end up storing all three.

Agreed, today's `ProcessId` type is more precisely `ProcessHandle`, or a freely shareable and copyable handle that is always checked for validity before being used. For uses within the core kernel, we do not always need this functionality. We might also be better off just storing a `&dyn Process` directly in many cases inside of the core kernel.

phil-levis:
> Agreed, today's `ProcessId` type is more precisely `ProcessHandle`, or a freely shareable and copyable handle that is always checked for validity before being used. For uses within the core kernel, we do not always need this functionality. We might also be better off just storing a `&dyn Process` directly in many cases inside of the core kernel.

Right, and it seems like we pass it around by value a lot, which can lead to copies for a 12-byte structure. Because it's called `ProcessId` I think people think it's lightweight, when it isn't.

bradjc:
I resolved the merge conflict and updated to the _num naming.

I also updated a comment along the way.

kernel/src/driver.rs
```diff
@@ -184,7 +185,13 @@ pub trait Driver {
     /// is signaled with an upcall). Command 0 is a reserved command to
     /// detect if a peripheral system call driver is installed and must
     /// always return a CommandReturn::Success.
-    fn command(&self, which: usize, r2: usize, r3: usize, caller_id: ProcessId) -> CommandReturn {
+    fn command(
+        &self,
+        command_num: usize,
+        r2: usize,
+        r3: usize,
+        caller_id: ProcessId,
```
jettr, line 193: Can we call this process_id

(I can make the change. I don't know whose court the ball is in since you pushed the change for me; just let me know)

bradjc: Yes good catch.

bradjc, line 193:
```suggestion
        process_id: ProcessId,
```

bradjc (approved):

hudson-ayers (approved):

PR #2639, opened 2021-06-29
Tock 2.0: implement Upcall swapping restrictions (v4)

bradjc:
### Pull Request Overview

This is an alternative approach to #2462 that ensures apps always get back the previous upcall on a subscribe call.

This implementation works by handling all upcall storage and swapping entirely in the core kernel. The core kernel creates its own data structure as a process grant and stores upcalls there. Since the core kernel is entirely responsible, we can ensure that upcall swapping is always handled correctly.

Capsules now no longer store upcalls. Instead they use an API like:

```rust
self.grants.enter(processid, |process_grant, upcalls| {
    ... // use the grant

    upcalls.schedule_upcall(upcall_id, arg0, arg1, arg2);
}
```

Other ramifications of this change:

- `Driver` now includes a `allocate_grant()` function that capsules have to implement. This provides the core kernel a method for forcing a grant allocation so that it can store an upcall if the process calls subscribe before the grant has been allocated.
- Grants need to be identified with the driver number of the capsule that is going to use them. This allows the kernel to match the correct upcall on a subscribe call.
- Grant<T> is now `Grant<T, NUM_UPCALLS>` so capsules can tell the core kernel how many upcalls they need stored.

### Testing Strategy

Tested apps on hail and imix.

### TODO or Help Wanted

- [x] Move pointer manipulation code out of process_standard.rs
- [x] Unsafe documentation
- [x] Cleanup
- [x] Fix IPC to make `NUM_CALLBACKS` tied to `NUM_PROCESSES` again, and to address that IPC did not follow TRD 104 (additional actions were taken in response to a subscribe)

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alexandruradovici:
Nice implementation, from the capsule writer point of view this seems to make driver writing easier.

I have two concerns though.

What happens if a capsule has two or more grants? My assumption based on the `driver_num_to_grant_num` implementation is that the first one created (or the first that is actually allocated if the capsule only allocates some of them) is used for upcalls.

As the driver number is now used in two places, the `with_driver` function and the `create_grant`, it is easy to make a mistake and supply two different driver numbers, which would result in upcalls silently failing and registering upcalls for another capsule.

hudson-ayers:
Currently, this PR reduce code size on Imix by ~300 bytes.

hudson-ayers:
> I have two concerns though.
> 
> What happens if a capsule has two or more grants? My assumption based on the `driver_num_to_grant_num` implementation is that the first one created (or the first that is actually allocated if the capsule only allocates some of them) is used for upcalls.

Good catch -- we need to bring back the check from #2462 that there is only one grant per driver. That is done now. I think we might be able to get away with just doing a check when each grant is allocated, rather than having to use a separate table, but using the existing solution for now.

> As the driver number is now used in two places, the `with_driver` function and the `create_grant`, it is easy to make a mistake and supply two different driver numbers, which would result in upcalls silently failing and registering upcalls for another capsule.

Yeah, this problem also exists for #2462 . But I just added some checks to this PR so that this mistake will not lead to registering upcalls for another capsule, and instead lead to an error. We could replace this with a panic for it not to be silent

alexandruradovici:
> > I have two concerns though.
> > What happens if a capsule has two or more grants? My assumption based on the `driver_num_to_grant_num` implementation is that the first one created (or the first that is actually allocated if the capsule only allocates some of them) is used for upcalls.
> 
> Good catch -- ~we need to bring back the check from #2462 that there is only one grant per driver. That is done now. I think we might be able to get away with just doing a check when each grant is allocated, rather than having to use a separate table, but using the existing solution for now.~
> I changed this to check at the time of grant allocation if there are multiple grants with the same driver_num.
> 
> > As the driver number is now used in two places, the `with_driver` function and the `create_grant`, it is easy to make a mistake and supply two different driver numbers, which would result in upcalls silently failing and registering upcalls for another capsule.
> 
> Yeah, this problem also exists for #2462 . But I just added some checks to this PR so that this mistake will not lead to registering upcalls for another capsule, and instead lead to an error. We could replace this with a panic for it not to be silent.

You are right, but it still bugs me that the driver number has to be supplied in two places. Definitely the check helps.

bradjc:
> You are right, but it still bugs me that the driver number has to be supplied in two places. Definitely the check helps.

Unfortunately I think having a one-to-one mapping of grant to driver_num is fundamental, and the `Grant` type needs to know the driver_num. If we changed the design to remove the `with_driver()` mapping function we could hard-code driver numbers for capsules, but then we restrict how board authors can customize their kernel.

kernel/src/upcall.rs
```diff
@@ -15,109 +15,78 @@ use crate::ErrorCode;
 #[derive(Copy, Clone, PartialEq, Debug)]
 pub struct UpcallId {
     pub driver_num: usize,
-    pub subscribe_num: usize,
+    pub subscribe_num: usize, //TODO: use u32?
```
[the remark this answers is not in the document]
hudson-ayers: In #2462 these were ported to u32

kernel/src/grant.rs
```diff
@@ -721,22 +1006,22 @@ impl GrantRegionAllocator {
     /// initialized using the provided function.
     ///
     /// The provided function will be called exactly `n` times, and will be
-    /// passed the index it's initializing, from `0` through `NUM_ITEMS - 1`.
+    /// passed the index it's initializing, from `0` through `NUM_UPCALLS_ITEMS - 1`.
```
bradjc, line 1009: Are these changes supposed to be here?

bradjc:
The hail app now works on hail!

kernel/src/grant.rs
hudson-ayers: nope, thats a find/replace mistake, sorry

kernel/src/sched.rs
```diff
@@ -902,37 +903,101 @@ impl Kernel {
...
+                        // allocate the grant.
+                        match crate::grant::subscribe(process, upcall) {
+                            Ok(old_upcall) => old_upcall.into_subscribe_success(),
+                            Err((new_upcall, _err)) => {
+                                // If we get here the subscribe swap failed. We try
+                                // again after asking the capsule to allocate the
+                                // grant.
+                                platform.with_driver(driver_number, |driver| match driver {
+                                    Some(d) => {
+                                        if d.allocate_grant(process.processid()).is_err() {
```
hudson-ayers, line 959: we should probably check if the grant is allocated before calling this

bradjc: We do on line 915.

kernel/src/sched.rs
```diff
@@ -902,37 +903,101 @@ impl Kernel {
                 upcall_ptr,
                 appdata,
             } => {
```
bradjc, line 905: @phil-levis Can you read through the subscribe match block, and check that the logic flow and comments match what TRD104 promises?

kernel/src/sched.rs
hudson-ayers: Do we? This is what I see for line 915:
```
                // First check if `upcall_ptr` is null. A null upcall pointer
                // represents the special "unsubscribe" operation.
                let ptr = NonNull::new(upcall_ptr);
```

bradjc: Ha that's what I get for doing it on my phone. That should be "951". Basically we first try the subscribe, and only if it fails do we try to do the grant allocation and eventually retry the subscribe.

kernel/src/upcall.rs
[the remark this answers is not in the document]
alevy: yeah... `usize` if they are register sized and `u32` if they are 4-byte words (which happen to be the same on all of our platforms). Since an upcall is something that models the interaction over a system call boundary, it seems to me it should be a `usize`. E.g., "how many subscribe numbers can there be?" answer: "the max value that can fit in a CPU register", not, "2^32".

bradjc:
Ok I believe all in code TODOs have been resolved, and all new `unsafe` usages are documented. I updated the new boards changes to use the new create_grant signature.

The remaining issue is that IPC is broken because it can no longer do any operations as a result of a subscribe.

hudson-ayers:
I pushed the IPC fix. The main breakage was that the number of upcalls now needs to be passed as a const, but for IPC the number of upcalls is `NUM_PROCS + 1`, and `min_const_generics` does not let you do arithmetic on passed consts. So for now we need another generic parameter, unfortunately.

I investigated everything that the existing subscribe implementation did, and it was actually just buffer swapping with some extra checks (which allowed the driver to reject subscribes). As far as I can tell the only check that has been removed by letting the kernel handle all this is a call to `lookup_app_by_identifier()`, called when an app subscribes to a service by registering an upcall in the slot associated with that service. This check was used to verify a process actually existed in that slot before trying to register an upcall for that service. Now, if an app tries to register an upcall for a service corresponding to a non-existent process, it will not get an error. I don't think this is a huge deal, personally.

bradjc:
Ok, the IPC issues have been resolved, and I updated the capsule porting guide. I think this is now a viable PR to implement the upcall swapping restrictions.

hudson-ayers:
I have tested this PR with blink and with the IPC apps on Imix

hudson-ayers:
Also, the final version of this PR reduces code size on Imix by 1500 bytes!

kernel/src/sched.rs
phil-levis: Will do.

bradjc:
One issue that comes up with this approach, or any approach that tags a Grant with a driver_num, is that it is possible to have a mismatch between the syscall driver_num and the grant_driver num, and depending on what userspace does this could manifest in some very strange behavior for userspace processes that would not be immediately obvious to developers.

One thought is to add a `panic()` if the kernel detects a mismatch. This would be palatable from a tock perspective because this would be the fault of a faulty main.rs file, which we consider trusted code anyway. However, I do not think this is possible, because the mismatch case would present identically to the case where a capsule maliciously implements `Driver.allocate_grant()` (i.e. does not allocate its grant when asked). In both cases, the kernel has to handle a subscribe call where it does not find a matching grant to store it in. But if we panic, then a capsule can cause a kernel panic entirely by how it implements one function (allocate_grant()), in a way that a capsule is technically allowed to do.

Another thought is to make the mistake harder to make by embedding the default driver_num in components themselves. Since `with_driver()` is in our experience always implemented correctly and using the default DRIVER_NUMs, this would help ensure that the same driver_num is used in the syscall mapping and in the create_grant() call. However, by hiding the fact that the driver_num is stored in the capsule, this could cause other problems when developers aren't aware the driver_num is in the capsule and they must ensure it matches. Also, the version where driver_num is passed to components has been around for a while in PRs. Ultimately, we can change our mind on how to minimize the chance of mixing up the driver_num, but it seems like it is worth seeing if this is a common bug before trying too hard to solve it.

kernel/src/syscall.rs
```diff
@@ -277,10 +277,10 @@ pub enum SyscallReturn {
 
     /// Subscribe success case, returns the previous upcall function
     /// pointer and application data.
-    SubscribeSuccess(*const u8, usize),
+    SubscribeSuccess(*mut (), usize),
```
phil-levis, line 280: Why is this `mut` rather than const?

kernel/src/syscall.rs
```diff
@@ -277,10 +277,10 @@ pub enum SyscallReturn {
...
     /// Subscribe failure case, returns the passed upcall function
     /// pointer and application data.
-    SubscribeFailure(ErrorCode, *const u8, usize),
+    SubscribeFailure(ErrorCode, *mut (), usize),
```
phil-levis, line 283: Why is this `mut` rather than const?

lschuermann:
I agree we want to get this in quickly, but should this be `last-call` already? It's a major change to the core kernel, I would've liked to have an in depth look over the weekend.

kernel/src/process_standard.rs
```diff
@@ -1350,7 +1409,7 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
         // sure we allocate enough memory just for that.
 
         // Make room for grant pointers.
-        let grant_ptr_size = mem::size_of::<*const usize>();
+        let grant_ptr_size = mem::size_of::<(usize, *mut u8)>();
```
phil-levis, line 1412: `*const`

jrvanwhy:
> I agree we want to get this in quickly, but should this be `last-call` already? It's a major change to the core kernel, I would've liked to have an in depth look over the weekend.

We were planning to wait until at least Monday to merge.

kernel/src/grant.rs
```diff
@@ -153,6 +157,159 @@ impl<'a, T: 'a + ?Sized> DerefMut for GrantMemory<'a, T> {
...
+/// This UpcallMemory object provides a handle to access Upcalls stored on
+/// behalf of a particular grant/driver.
+///
+/// Capsules gain access to a UpcallMemory object by calling `Grant::enter()`.
+/// From there, they can schedule upcalls. No other access to upcalls is
+/// provided.
+///
+/// It is expected that this type will only exist as a short-lived stack
+/// allocation, so its size is not a significant concern.
+pub struct UpcallMemory<'a> {
```
phil-levis, line 169: `UpcallTable`

doc/Porting_v1_Capsules_to_v2.md
```diff
@@ -272,70 +273,44 @@ method swaps the passed `ReadOnlyAppSlice` and the one in the `App` region,
 returning the one that was in the app region. It then returns `slice`,
 which is either the passed slice or the swapped out one.
 
-### Example of `subscribe`
+### The new subscription mechanism
 
-A call to `subscribe` has a similar structure to `allow`. Here is
-an example from `console`:
+To use upcalls a capsule now provides the number of upcalls as a templated value
```
phil-levis, line 278: Explain why, how V2 subscribe is different.

phil-levis (approved):

kernel/src/grant.rs
```diff
@@ -382,14 +710,15 @@ impl<'a, T: Default> ProcessGrant<'a, T> {
     /// `Some(fun())`.
     pub fn try_enter<F, R>(self, fun: F) -> Option<R>
     where
-        F: FnOnce(&mut GrantMemory<T>) -> R,
+        F: FnOnce(&mut GrantMemory<T>, &mut UpcallMemory) -> R,
```
phil-levis, line 713: From call:

GrantMemory -> GrantData
UpcallMemory -> GrantUpcallTable

kernel/src/process_standard.rs
bradjc: This is a pointer to mutable memory, as it points to where we store the upcalls and grant object.

kernel/src/grant.rs
bradjc: Changed UpcallMemory, other change is in 2651.

doc/Porting_v1_Capsules_to_v2.md
bradjc: Updated.

kernel/src/process_standard.rs
phil-levis: Ah, right. Sorry, my pattern-match wasn't selective enough.

kernel/src/grant.rs
```diff
@@ -488,33 +836,73 @@ impl<'a, T: Default> ProcessGrant<'a, T> {
...
+        let saved_upcalls_ptr = unsafe { grant_ptr.add(size_of::<usize>()) };
+        // # Safety
+        //
+        // Creating this slice is safe because the pointer is in the grant
+        // allocation that is guaranteed to still exist, a grant can only be
+        // entered once at a time so this is the only mutable reference to this
+        // slice, and the slice has valid SavedUpcalls which are guaranteed to
+        // be initialized.
+        let saved_upcalls_slice = unsafe {
+            slice::from_raw_parts_mut(saved_upcalls_ptr as *mut SavedUpcall, NUM_UPCALLS)
```
lschuermann, line 879: Can this be a `slice::from_raw_parts_mut` here?

kernel/src/grant.rs
```diff
@@ -318,7 +664,7 @@ impl<'a, T: Default> ProcessGrant<'a, T> {
     /// panic!()`. See the comment in `access_grant()` for more information.
     pub fn enter<F, R>(self, fun: F) -> R
     where
-        F: FnOnce(&mut GrantMemory<T>) -> R,
+        F: FnOnce(&mut GrantMemory<T>, &mut GrantUpcallTable) -> R,
```
lschuermann, line 667: Why do we pass a mutable reference to `GrantUpcallTable`? My fear is that users are going to be able to `mem::swap` this potentially, though this probably is (a) not possible because of non-reentrant Grants and (b) not a concern given the passed `GrantUpcallTable` reference is dropped afterwards. Still, if there's no good reason for it to be mutable, it shouldn't be.

kernel/src/grant.rs
```diff
@@ -560,12 +948,39 @@ impl<'a, T: Default> ProcessGrant<'a, T> {
...
+        let saved_upcalls_ptr = unsafe { grant_ptr.add(size_of::<usize>()) };
+        // # Safety
+        //
+        // Creating this slice is safe because the pointer is in the grant
+        // allocation that is guaranteed to still exist, a grant can only be
+        // entered once at a time so this is the only mutable reference to this
+        // slice, and the slice has valid SavedUpcalls which are guaranteed to
+        // be initialized.
+        let saved_upcalls_slice = unsafe {
+            slice::from_raw_parts_mut(saved_upcalls_ptr as *mut SavedUpcall, NUM_UPCALLS)
```
lschuermann, line 981: Here as well.

lschuermann (dismissed):
These changes look good and seem to improve semantics w.r.t. driver implementations. I had a very close look at the kernel code and could not find any unsafety / unsoundness. I have not tested this at runtime, nor its performance impact.

kernel/src/grant.rs
```diff
@@ -153,6 +169,166 @@ impl<'a, T: 'a + ?Sized> DerefMut for GrantMemory<'a, T> {
...
+    }
+
+    /// Schedule the specified upcall for the process with r0, r1, r2 as
+    /// provided values.
+    ///
+    /// Capsules call this function to schedule upcalls, and upcalls are
+    /// identified by the `subscribe_num`, which must match the subscribe number
+    /// used when the upcall was originally subscribed by a process.
+    /// `subscribe_num`s are indexed starting at zero.
+    pub fn schedule_upcall(&self, subscribe_num: usize, r0: usize, r1: usize, r2: usize) -> bool {
```
ppannuto, line 217: Should this return a `Result<(), Error>` rather than a bool?

It's a bit annoying from a verbosity perspective, as it results in a boilerplate `upcalls.schedule_upcall(..)` -> `upcalls.schedule_upcall(..).ok()` for most callers to silence the ignored `Result`, but it does seem clearer than the `bool`, and there is some benefit to asking capsule authors to consider briefly whether they care about the process not actually being notified.

kernel/src/grant.rs
```diff
@@ -258,19 +547,74 @@ impl<'a, T: Default> ProcessGrant<'a, T> {
                 //    as the process does. The grant is only accessible while
                 //    the process is still valid.
                 //
-                // 2. The pointer is correct aligned. The `allocate_grant()`
-                //    function ensures the pointer is correctly aligned.
+                // 2. The pointer is correct aligned because we calculated the
```
ppannuto, line 550:
```suggestion
                // 2. The pointer is correctly aligned because we calculated the
```

kernel/src/grant.rs
```diff
@@ -258,19 +547,74 @@ impl<'a, T: Default> ProcessGrant<'a, T> {
...
+                    // Initialize the SavedUpcalls in an explicit loop. We do
+                    // not use a slice because before this runs the SavedUpcalls
+                    // are not initialized and creating a slice to uninitialized
+                    // memory is not safe.
+                    for i in 0..NUM_UPCALLS {
+                        // # Safety
+                        //
+                        // This is safe because we have allocated enough space
+                        // for `NUM_UPCALLS` and that each SavedUpcall is at
+                        // least aligned to 4 bytes.
```
ppannuto, line 572:
```suggestion
                        // least aligned to `usize` bytes.
```
?

ppannuto (commented):
Read over `kernel/` and `doc/` pretty carefully. I think this all looks good; a few small doc cleanups and one interface question.

kernel/src/grant.rs
```diff
@@ -81,7 +90,7 @@
 //!
 //! ```text,ignore
 //!                         ┌──────────────────────────┐
-//!                         │ struct Grant<T> {        │
+//!                         │ struct Grant<T, NUM_UP> {│
 //!                         │   number: usize          │
```
alexandruradovici, line 94:
```suggestion
//!                         │   driver_num: usize          │
//!                         │   grant_num: usize          │
```

kernel/src/sched.rs
```diff
@@ -902,37 +921,101 @@ impl Kernel {
...
+                                            // could not be created and return an
+                                            // error to userspace.
+                                            new_upcall.into_subscribe_failure(ErrorCode::NOMEM)
+                                        } else {
+                                            // Now we try again. It is possible that
+                                            // the capsule did not actually allocate
+                                            // the grant, at which point this will
+                                            // fail again and we return an error to
+                                            // userspace.
+                                            match crate::grant::subscribe(process, new_upcall) {
```
alexandruradovici, line 989: It might be a good idea to show a debug message if the driver wrongly allocates a wrong grant. https://github.com/tock/tock/pull/2654

alexandruradovici (commented):
I have a small documentation update and a debug suggestion.

kernel/src/grant.rs
hudson-ayers: I'm not sure what you mean -- it already is a `slice::from_raw_parts_mut` here? Is the concern that using it this way is unsafe?

kernel/src/grant.rs
hudson-ayers: I agree this is not needed, and pushed a commit changing it to a non-mutable reference

kernel/src/grant.rs
hudson-ayers: I just ported the existing interface for `schedule()`. I think you're right that a `Result` would be better, but I might prefer for that to be submitted as a follow up PR, since it would require updating all the capsules again, and the current content of this PR has been pretty heavily reviewed.

kernel/src/grant.rs
lschuermann: Ah, sorry! Should've been "can this be a `slice::from_raw_parts`" (i.e. non-mutable) here? Not a safety concern per se.

kernel/src/grant.rs
bradjc: Does a capsule have any recourse if a schedule() fails? I'm guessing that is why this has escaped the resultifying this long, I'm not sure what a capsule could possibly do that would be worth it. Has any capsule ever even handled the bool?

I'm not saying we want to stick with the c convention, but we also probably don't want `upcalls.schedule().map_err(debug!("i have no idea what happened"))`.

bradjc:
Addressed all of the comments to this point.

kernel/src/grant.rs
ppannuto: I don't have a compelling in-tree example I don't think, but my guess was that it was more likely that some capsule might need to do something extra if it thinks the process has died (e.g. check in some other way if the process is truly dead after the subscribe fails and choose to power off the associated peripheral).

It doesn't need the whole `map_err` bit, literally appending `.ok()` as shown is sufficient; that is a "Rust trick" I read somewhere along the way that just turns a `Result` into an `Option` (you lose the `Error` when you do this; Success -> `Some`, any error -> `None`), and the `Option` can be dropped without the compiler yelling.

@hudson-ayers follow-up PR seems fine; should be a pretty mechanical change.

ppannuto (approved):

kernel/src/grant.rs
bradjc: Thinking about this a little more, it's impossible for .schedule() to detect a dead process, because to enter a grant the process must be valid and while the grant is entered the process must stay valid. So schedule could fail because:

1. the `subscribe_num` was invalid (i.e. too big)
2. the upcall is null, i.e. the process never called subscribe
3. there is no room in the tasks array for the process

We don't need a persistent runtime check for 1, number 2 is normal operation that capsules aren't expected to defend against (I'm pretty sure), and for 3 its not clear what the capsule should do. Perhaps encapsulating these three cases in an Error type would spur someone to think about this any maybe come up with something.

And `ok()` is the clear tool for capsule authors to use, but I was just suggesting that if schedule() returns a proper result and a capsule author is thinking about what to do with the error case (i.e. "why does the kernel return me an error if I am not expected to do anything with it?"), they likely won't come up with anything and might write a debug statement instead.

hudson-ayers (approved):

lschuermann (approved):

hudson-ayers:
I have tested the current head of this PR using several apps on Imix, including the `Imix`, `blink`, and `hello_loop` apps simultaneously.

bors r+

kernel/src/grant.rs
ppannuto: Opened #2656

phil-levis: If `schedule` cannot detect a dead process, what are the error cases?

It seems like this should return nothing. A `Result` is just going to add a lot of boilerplate that doesn't do anything.

PR #2642, opened 2021-06-30
Tock reg UIntLike

### LINKED ISSUE #2641: Quick RFC: Tock Registers: `IntLike` to `UIntLike`
#2623 highlighted that really our register interface lets us turn register operations into operations on well-defined unsigned numbers. We called this trait magic `IntLike`, presumably as we were saying that the underlying data could be manipulated like numbers, but I think the implementation is more accurately `UIntLike` (indeed, all of the example types are `u8`, `u16`, ...).

Since we're going to have some churn with this upcoming release, and since it will hopefully be a 1.0 interface, I'd propose we rename `IntLike` to `UIntLike`, both to more accurately reflect what's happening, and leave the door open to doing something intelligent in a non-breaking manner with registers that hold signed integer -like values in the future.

This should be a quick `s/IntLike/UIntLike` if we opt to make the change.

Thoughts? @lschuermann @andre-richter @bradjc @brghena

ppannuto:
### Pull Request Overview

tock-registers: Rename `IntLike` to `UIntLike`

This better captures the semantics of this interfaces.

Closes #2641.

Depends on #2618.

### Testing Strategy

### TODO or Help Wanted

N/A.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/litex/src/event_manager.rs
+++ b/chips/litex/src/event_manager.rs
@@ -4,7 +4,7 @@
 //! behave differently, can be found in the LiteX repository under
 //! [`litex/soc/interconnect/csr_eventmanager.py`](https://github.com/enjoy-digital/litex/blob/master/litex/soc/interconnect/csr_eventmanager.py).
 
-use crate::litex_registers::{IntLike, Read, ReadWrite};
+use crate::litex_registers::{Read, ReadWrite, UIntLike};
 use core::marker::PhantomData;
 
 /// LiteX event manager abstraction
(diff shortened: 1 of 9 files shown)


hudson-ayers (approved):

alistair23 (approved):

lschuermann:
Currently rebasing this on the latest changes of #2618.

bradjc (approved):

lschuermann:
This seems like a pretty minor and uncontroversial change, thus

bors r+

I guess by far the most tricky thing must've been understanding the weird `litex_registers` module, which I will fix to use the `interfaces` trais, now that all the changes have landed properly. Thanks @ppannuto!

PR #2648, opened 2021-07-02
Move kernel code out of monomorphized functions

hudson-ayers:
### Pull Request Overview

This commit applies the non-generic-inner-functions approach
(https://www.possiblerust.com/pattern/non-generic-inner-functions)
to several functions in grant.rs and sched.rs. None of these changes
should impact functionality in any way (assuming I got these changes
right). At the expense of the code being a little less readable,
these changes reduce the code size of Imix by 6528 (!) bytes.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/sched.rs
+++ b/kernel/src/sched.rs
@@ -230,6 +230,27 @@ impl Kernel {
         self.work.get() == 0
     }
 
+    /// Helper function that moves all non-generic portions of process_map_or
+    /// into a non-generic function to reduce code bloat from monomorphization.
+    pub(crate) fn get_process(&self, processid: ProcessId) -> Option<&dyn process::Process> {
+        // We use the index in the `appid` so we can do a direct lookup.
+        // However, we are not guaranteed that the app still exists at that
+        // index in the processes array. To avoid additional overhead, we do the
+        // lookup and check here, rather than calling `.index()`.
+        match self.processes.get(processid.index) {
+            Some(Some(process)) => {
+                // Check that the process stored here matches the identifier
+                // in the `appid`.
+                if process.processid() == processid {
+                    Some(*process)
+                } else {
+                    None
+                }
+            }
+            _ => None,
+        }
+    }
+
     /// Run a closure on a specific process if it exists. If the process with a
     /// matching `ProcessId` does not exist at the index specified within the
     /// `ProcessId`, then `default` will be returned.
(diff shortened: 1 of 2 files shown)


lschuermann:
This is great. I'm a bit wary of both `#[inline(never)]` and `#[inline(always)]` as in general the compiler is able to do a good job to figure these out (at least w.r.t. it's optimization goals), but here they do seem justified such that we're guaranteed to achieve the desired effects.

Have you measured the performance impact of this yet? I could throw it into the LiteX simulator to see what changes, if you'd like. I don't _expect_ much of an effect, but it might be useful to check.

ppannuto (approved):
The changes themselves looks good (n.b. easier viewed with whitespace ignored).

One quick Q, I'm assuming you tested this empirically, but the linked blog post doesn't include the `inline` directives -- any sense of why we need them and they didn't (or perhaps just a simplified example there?)

hudson-ayers:
> One quick Q, I'm assuming you tested this empirically, but the linked blog post doesn't include the `inline` directives -- any sense of why we need them and they didn't (or perhaps just a simplified example there?)

Yeah, I noticed one place where without the directive about 50 bytes of code were added back. I didn't really look into it further. I generally agree that the compiler is pretty good about making inlining decisions, but since I noticed a small difference in one instance I went ahead and added the hints everywhere.

alistair23 (approved):

hudson-ayers:
After rebasing this on top of the callback swapping PR, I found that using the inline directives can actually hurt in some cases. Given that, I decided to remove them -- the compiler seems to usually come close to the best case without them, while including them when they should not be can lead to a significant hit.

kernel/src/sched.rs
```diff
@@ -230,6 +230,28 @@ impl Kernel {
         self.work.get() == 0
     }
 
+    /// Helper function that moves all non-generic portions of process_map_or
+    /// into a non-generic function to reduce code bloat from monomorphization.
+    #[inline(never)]
+    pub fn get_process(&self, processid: ProcessId) -> Option<&dyn process::Process> {
```
bradjc, line 236:
```suggestion
    pub(crate) fn get_process(&self, processid: ProcessId) -> Option<&dyn process::Process> {
```

kernel/src/sched.rs
```diff
@@ -240,34 +262,15 @@ impl Kernel {
     /// different index in the processes array. Note that a match _will_ be
     /// found if the process still exists in the correct location in the array
     /// but is in any "stopped" state.
+    #[inline(always)]
```
bradjc, line 265: This is still here?

hudson-ayers: fixed. That did cost us a few dozen bytes

bradjc (approved):

brghena (approved):

PR #2651, opened 2021-07-02
kernel: grant: rename `GrantMemory` to `GrantData`

bradjc:
### Pull Request Overview

`GrantData` better captures what the type does, it provides a mechanism for accessing the actual data stored by a capsule in a process's grant region.

This changed based on discussions on what to call the version of the same type that provides capsules access to upcalls stored in the grant region.

### Testing Strategy

Just a name change.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers (approved):

ppannuto (approved):
This was discussed on the call so I don't think it needs a long review period here.

PR #2654, opened 2021-07-05
Warn about driver allocating a wrong grant

alexandruradovici:
### Pull Request Overview

This pull request tries to show a warning or an error if a driver does not allocate a grant or seems to allocate the wrong grant through `allocate_grant`.

The idea is simple: count the number of allocated grants before calling `allocate_grant`. If `allocate_grant` returns success but subscribe does not work, the reason is one of the two:
  1. the driver did not allocate a grant, in this case the number of allocated grants stays the same
  2. the driver allocated the wrong grant, a grant having a different driver number, in this case the number of grants is grater the the original number of grants

These actions are performed only If `trace_syscalls` is set to `true`.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/sched.rs
+++ b/kernel/src/sched.rs
@@ -1009,7 +1009,12 @@ impl Kernel {
                                                 Ok(old_upcall) => {
                                                     old_upcall.into_subscribe_success()
                                                 }
-                                                Err((new_upcall, err)) => {
+                                                Err((new_upcall, ErrorCode::NOMEM)) => {
+                                                    // Special case the handling of ErrorCode::NOMEM,
+                                                    // as it indicates a missing Grant allocation.
+                                                    // Depending on the kernel configuration, we
+                                                    // inform the user about the root cause of this
+                                                    // issue.
                                                     if config::CONFIG.trace_syscalls {
                                                         // It appears the Grant is still not allocated.
                                                         // Based on whether the number of allocated
@@ -1032,6 +1037,13 @@ impl Kernel {
                                                                 }
                                                             }
                                                     }
+                                                    new_upcall.into_subscribe_failure(ErrorCode::NOMEM)
+                                                }
+                                                Err((new_upcall, err)) => {
+                                                    // Handler for all errors other than
+                                                    // ErrorCode::NOMEM, for example when
+                                                    // the subscribe number exceeds the
+                                                    // number of Upcalls in the Grant region.
                                                     new_upcall.into_subscribe_failure(err)
                                                 }
                                             }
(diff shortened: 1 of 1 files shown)


kernel/src/sched.rs
```diff
@@ -994,6 +1007,17 @@ impl Kernel {
                                                     old_upcall.into_subscribe_success()
                                                 }
                                                 Err((new_upcall, err)) => {
+                                                    if config::CONFIG.trace_syscalls {
+                                                        if let Some(currently_allocated_grants_count) = process.grant_allocated_count() {
+                                                            if currently_allocated_grants_count > allocated_grants_count {
+                                                                debug! ("[{:?}] error driver {} allocated wrong grant for upcalls", process.processid(), driver_number);
```
hudson-ayers, line 1013: It definitely seems like a rustfmt bug that this passes formatting with such long line length.

hudson-ayers (approved):
I like the approach of a debug message that only shows up if a config option is toggled. Grouping this in with `trace_syscalls` seems fine, though I could see it eventually being gated behind a different debugging config.

kernel/src/sched.rs
alexandruradovici: It is a bug, I tried formatting it in several ways, no luck, I formatted it manually.

lschuermann (changes requested):
Looks good in general, although as far as I understand this will print the debug message regardless of which actual error occurred in `kernel::grant::subscribe`. It should only be printed on an `ErrorCode::NOMEM`. Opened https://github.com/WyliodrinEmbeddedIoT/tock/pull/7 accordingly (that also make the ci-format check pass).

kernel/src/sched.rs
lschuermann: Yeah, I've noticed that too. https://github.com/WyliodrinEmbeddedIoT/tock/pull/7/commits/9eb3e2ffcfc433cacc17a13d4689df30a854109f seems like a reasonable formatting + makes the CI work...

alexandruradovici:
> Looks good in general, although as far as I understand this will print the debug message regardless of which actual error occurred in `kernel::grant::subscribe`. It should only be printed on an `ErrorCode::NOMEM`. Opened [WyliodrinEmbeddedIoT#7](https://github.com/WyliodrinEmbeddedIoT/tock/pull/7) accordingly.

Thank you @lschuermann for the patch

lschuermann (approved):

bradjc:
Merged manually because I couldn't push to this PR.

PR #2656, opened 2021-07-07
kernel: schedule upcall, bool -> Result

ppannuto:
### Pull Request Overview

Following up on https://github.com/tock/tock/pull/2639#discussion_r664131701, this looks at changing the signature of scheduling upcalls to return a `Result` rather than a simple `bool`.

To start, this just looks at the kernel interfaces (i.e. this introduces 110 "unused Result" warnings in capsule code right now).

One interesting thing surfaced is that IPC does use the result of upcall scheduling to report to the IPC sender whether the receiver was scheduled successfully or not. Previously, this treated scheduling the null upcall as a failure. I'm not 100% sure that's right though -- a receiver has every right (although it would be odd) to ignore notifications. I used the new error code granularity to separate out this case right now; though maybe the best solution is to simply bubble up the error all the way to userspace for IPC.

---

This commit just changes the kernel interfaces, it does not address
all of the new unused Result warnings at all of the capsule scheduling
call sites.

### Testing Strategy

N/A yet [compiling]

### TODO or Help Wanted

Do these error code assignments make sense?

Should IPC treat scheduling a null upcall as an error or success?

Should I move forward with the rest of the capsules?

### Documentation Updated

### Formatting

- [] Ran `make prepush`.

### DIFF
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -129,7 +129,7 @@ use core::slice;
 
 use crate::process::{Error, Process, ProcessCustomGrantIdentifer, ProcessId};
 use crate::sched::Kernel;
-use crate::upcall::{Upcall, UpcallId};
+use crate::upcall::{Upcall, UpcallError, UpcallId};
 use crate::ErrorCode;
 
 /// This GrantData object provides access to the memory allocated for a grant
@@ -215,11 +215,17 @@ impl<'a> GrantUpcallTable<'a> {
     /// identified by the `subscribe_num`, which must match the subscribe number
     /// used when the upcall was originally subscribed by a process.
     /// `subscribe_num`s are indexed starting at zero.
-    pub fn schedule_upcall(&self, subscribe_num: usize, r0: usize, r1: usize, r2: usize) -> bool {
+    pub fn schedule_upcall(
+        &self,
+        subscribe_num: usize,
+        r0: usize,
+        r1: usize,
+        r2: usize,
+    ) -> Result<(), UpcallError> {
         // Implement `self.upcalls[subscribe_num]` without a chance of a panic.
-        self.upcalls
-            .get(subscribe_num)
-            .map_or(false, |saved_upcall| {
+        self.upcalls.get(subscribe_num).map_or(
+            Err(UpcallError::InvalidSubscribeNum),
+            |saved_upcall| {
                 // We can create an `Upcall` object based on what is stored in
                 // the process grant and use that to add the upcall to the
                 // pending array for the process.
@@ -233,7 +239,8 @@ impl<'a> GrantUpcallTable<'a> {
                     saved_upcall.fn_ptr,
                 );
                 upcall.schedule(self.process, r0, r1, r2)
-            })
+            },
+        )
     }
 }
 
--- a/kernel/src/ipc.rs
+++ b/kernel/src/ipc.rs
@@ -126,20 +126,25 @@ impl<const NUM_PROCS: usize, const NUM_UPCALLS: usize> IPC<NUM_PROCS, NUM_UPCALL
                                                 )
                                             },
                                         );
-                                        my_upcalls.schedule_upcall(
-                                            to_schedule,
-                                            called_from.id() + 1,
-                                            crate::mem::ReadableProcessBuffer::len(slice),
-                                            crate::mem::ReadableProcessBuffer::ptr(slice) as usize,
-                                        );
+                                        my_upcalls
+                                            .schedule_upcall(
+                                                to_schedule,
+                                                called_from.id() + 1,
+                                                crate::mem::ReadableProcessBuffer::len(slice),
+                                                crate::mem::ReadableProcessBuffer::ptr(slice)
+                                                    as usize,
+                                            )
+                                            .ok();
                                     }
                                     None => {
-                                        my_upcalls.schedule_upcall(
-                                            to_schedule,
-                                            called_from.id() + 1,
-                                            0,
-                                            0,
-                                        );
+                                        my_upcalls
+                                            .schedule_upcall(
+                                                to_schedule,
+                                                called_from.id() + 1,
+                                                0,
+                                                0,
+                                            )
+                                            .ok();
                                     }
                                 }
                             }
--- a/kernel/src/ipc.rs
+++ b/kernel/src/ipc.rs
@@ -226,12 +226,14 @@ impl<const NUM_PROCS: usize, const NUM_UPCALLS: usize> Driver for IPC<NUM_PROCS,
                             CommandReturn::failure(ErrorCode::INVAL),
                             otherapp,
                             |target| {
-                                let ret = target
-                                    .enqueue_task(process::Task::IPC((appid, cb_type)))
-                                    .is_ok();
+                                let ret = target.enqueue_task(process::Task::IPC((appid, cb_type)));
                                 match ret {
-                                    true => CommandReturn::success(),
-                                    false => CommandReturn::failure(ErrorCode::FAIL),
+                                    Ok(()) => CommandReturn::success(),
+                                    Err(e) => match e {
+                                        // The other side has a null upcall, choosing to ignore
+                                        ErrorCode::OFF => CommandReturn::success(),
+                                        _ => CommandReturn::failure(e),
+                                    },
                                 }
                             },
                         )
@@ -251,12 +253,14 @@ impl<const NUM_PROCS: usize, const NUM_UPCALLS: usize> Driver for IPC<NUM_PROCS,
                             CommandReturn::failure(ErrorCode::INVAL),
                             otherapp,
                             |target| {
-                                let ret = target
-                                    .enqueue_task(process::Task::IPC((appid, cb_type)))
-                                    .is_ok();
+                                let ret = target.enqueue_task(process::Task::IPC((appid, cb_type)));
                                 match ret {
-                                    true => CommandReturn::success(),
-                                    false => CommandReturn::failure(ErrorCode::FAIL),
+                                    Ok(()) => CommandReturn::success(),
+                                    Err(e) => match e {
+                                        // The other side has a null upcall, choosing to ignore
+                                        ErrorCode::OFF => CommandReturn::success(),
+                                        _ => CommandReturn::failure(e),
+                                    },
                                 }
                             },
                         )
--- a/kernel/src/ipc.rs
+++ b/kernel/src/ipc.rs
@@ -226,7 +226,9 @@ impl<const NUM_PROCS: usize, const NUM_UPCALLS: usize> Driver for IPC<NUM_PROCS,
                             CommandReturn::failure(ErrorCode::INVAL),
                             otherapp,
                             |target| {
-                                let ret = target.enqueue_task(process::Task::IPC((appid, cb_type)));
+                                let ret = target
+                                    .enqueue_task(process::Task::IPC((appid, cb_type)))
+                                    .is_ok();
                                 match ret {
                                     true => CommandReturn::success(),
                                     false => CommandReturn::failure(ErrorCode::FAIL),
@@ -249,7 +251,9 @@ impl<const NUM_PROCS: usize, const NUM_UPCALLS: usize> Driver for IPC<NUM_PROCS,
                             CommandReturn::failure(ErrorCode::INVAL),
                             otherapp,
                             |target| {
-                                let ret = target.enqueue_task(process::Task::IPC((appid, cb_type)));
+                                let ret = target
+                                    .enqueue_task(process::Task::IPC((appid, cb_type)))
+                                    .is_ok();
                                 match ret {
                                     true => CommandReturn::success(),
                                     false => CommandReturn::failure(ErrorCode::FAIL),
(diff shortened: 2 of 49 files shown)


lschuermann:
> One interesting thing surfaced is that IPC does use the result of upcall scheduling to report to the IPC sender whether the receiver was scheduled successfully or not. Previously, this treated scheduling the null upcall as a failure.

I'm reasonably confident that we, at least somewhere in Tock 2.0 development, decided to not tell capsules whether an `Upcall` was properly scheduled at all. As far as I remember, the boolean return value depended only on whether the process was still alive, and not on the fact whether the upcall in question is currently a null-upcall. @phil-levis is that correct?

Currently, `schedule_upcall` will return `false` if this is a null-upcall. Maybe this confusion around error handling (and it having changed over time) led to these implementation issues?

bradjc:
For reference, reasons schedule() can fail:
1. The process is not valid
2. The upcall is null
3. The subscribe num is invalid
4. The tasks array in the process was taken and not replaced
5. There is no room in the tasks queue

In practice, (1) cannot happen, because the grant.enter() would fail first, and the process has to remain valid while the grant is entered, and (4) won't happen, because that would be a kernel bug that must be fixed and everything would break immediately.

For (3), subscribe_num is almost certainly going to be a hardcoded value (all instances today are hardcoded), and I don't see any value in a runtime check.

For (2) I agree capsules shouldn't do anything in response to a null upcall.

Then (5) is the only case I could see returning an error, but I don't see what a capsule could do with that information.

ppannuto:
> 5. There is no room in the tasks queue
> Then (5) is the only case I could see returning an error, but I don't see what a capsule could do with that information.

Here again I think IPC might be the interesting kid on the block. In particular, if process A does something, and wants to notify process B via IPC, it does so, enqueues a task and all is happy. If A keeps running, and during this same quanta (or just generally before B gets to run, due to process priorities, etc) and does another something which requires notifying B via IPC, then [eventually] B will need to run and drain IPC events. If A finds it is generating many events that B never responds to, it may wish to `yield` or otherwise ask the scheduler to lower its priority so that B can run.

lschuermann:
@bradjc I generally agree with your assessment, although I'm not sure about (3). You're right in that the subscribe number is hardcoded in current capsules, but given we're providing interfaces to downstream developers and this parameter could well be determined at runtime -- regardless of whether there is a good reason to do so -- I'm of the opinion we should return an error in this case.

If we are returning an error in one case, we might as well inform the capsule about all cases except (4), as that would indicate a serious kernel bug and might even justify a panic, and (2), which is to say that a null-callback is defined as a valid callback which is simply never delivered or from the capsules point of view ignored by the process.

bradjc:
I can certainly see the argument "wait I passed in an invalid subscribe_num and didn't get an error?!?". I propose then returning `Result<(), UpcallError>`:

```rust
enum UpcallError {
  InvalidSubscribeNum,
  QueueFull,
  // Process is inactive (can never happen) or process.tasks was empty (won't ever happen).
  // If either _did_ happen, well that would be a kernel error.
  // We do this in ProcessError to solve this same problem, and make it clear
  // that users of the error type shouldn't try to handle this case.
  KernelError,
}
```

lschuermann:
@bradjc I like that approach. I've taken at stab at implementing it here: ce9c08498a22fb341c91945797dc7e4a9f59c60a. If @ppannuto doesn't mind I can push it to this branch.

kernel/src/process_standard.rs
```diff
@@ -213,26 +213,27 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
...
-            });
-        } else {
-            self.kernel.increment_work();
-        }
-
-        ret
+        self.tasks.map_or_else(
+            |tasks| {
+                tasks.enqueue(task);
+                self.kernel.increment_work();
```
lschuermann, line 226: This discards the result of `tasks.enqueue` and always increments the work counter, which -- if I understand correctly -- effectively prevent the chip from going to sleep once a task has failed to be enqueued.

alexandruradovici:
> Then (5) is the only case I could see returning an error, but I don't see what a capsule could do with that information.

The `touch` driver receives a lot of callbacks from the underlying hardware driver as a result of the movement of fingers on the touch panel's surface. If the application is not processing callbacks fast enough, the task queue will be filled. The driver counts the _dropped_ upcalls and reports them to the application. In turn, the application can decide to prioritize the callbacks.

bradjc:
> @bradjc I like that approach. I've taken at stab at implementing it here: [ce9c084](https://github.com/tock/tock/commit/ce9c08498a22fb341c91945797dc7e4a9f59c60a). If @ppannuto doesn't mind I can push it to this branch.

I say push, we can always remove the commit.

> I wasn't sure whether `UpcallError` should be in `kernel::grant` or `kernel::upcall`, both seemed equally fitting giving that `GrantUpcallTable` is in `kernel::grant`. Furthermore, we need to export this enum, as otherwise user's will have no way to properly use it, for instance in a `match` expression.

I would say upcall.rs, and that upcall things being in grant.rs is the exception we would otherwise like to avoid.

bradjc:
I'm marking this 2.0 because I think we either do it now while we are thinking about it and change capsules, or we leave it `-> bool`.

lschuermann:
> I say push, we can always remove the commit.

Pushed my changes. This still does not address the unused `Result`s. @ppannuto happy to fix if people are otherwise happy with the interface.

kernel/src/grant.rs
```diff
@@ -170,6 +170,46 @@ impl<'a, T: 'a + ?Sized> DerefMut for GrantData<'a, T> {
...
+    /// `GrantUpcallTable::schedule_upcall` had no observable effects.
+    QueueFull,
+    /// A kernel-internal invariant has been violated.
+    ///
+    /// This error should never happen. It can be returned if the
+    /// process is inactive (which should be caught by
+    /// [`Grant::enter`]) or `process.tasks` was taken.
+    ///
+    /// These cases cannot be reasonably handled.
+    KernelError,
```
ppannuto, line 210: Is this error worth including? I am imaging me as a handler of this code -- what do I do? The options are
 1. Just ignore the error and hope (same as if an error had not been returned)
 2. `panic!`, the world is broken

I think that making capsule authors choose between these will likely just result in a lot of arbitrary coin flips between these two options.

kernel/src/ipc.rs
```diff
@@ -225,8 +225,12 @@ impl<const NUM_PROCS: usize, const NUM_UPCALLS: usize> Driver for IPC<NUM_PROCS,
                             |target| {
                                 let ret = target.enqueue_task(process::Task::IPC((appid, cb_type)));
                                 match ret {
-                                    true => CommandReturn::success(),
-                                    false => CommandReturn::failure(ErrorCode::FAIL),
+                                    Ok(()) => CommandReturn::success(),
+                                    Err(e) => match e {
+                                        // The other side has a null upcall, choosing to ignore
+                                        ErrorCode::OFF => CommandReturn::success(),
```
ppannuto, line 231: Should the `OFF` cases be removed now that we aren't treating the null call as an error?

kernel/src/upcall.rs
```diff
@@ -78,21 +79,46 @@ impl Upcall {
...
-                argument1: r1,
-                argument2: r2,
-                argument3: self.appdata,
-                pc: fp.as_ptr() as usize,
-            }))
-        });
+    ) -> Result<(), UpcallError> {
+        let res = self.fn_ptr.map_or(
+            Ok(()), // A null-Upcall is treated as being delivered to
+            // the process and ignored
```
ppannuto, line 85:
```suggestion
            // A null-Upcall is treated as being delivered to
            // the process and ignored
            Ok(()),
```

kernel/src/grant.rs
lschuermann: I suppose the issue is that `panic!`ing in the kernel is not an option (even if these error cases were to occur, the system might not be in an entirely unrecoverable state), whereas returning `Ok(())` as if the callback was scheduled is obviously misleading.

kernel/src/upcall.rs
lschuermann: Ah! That was so well aligned before `rustfmt` came along. :)

kernel/src/grant.rs
bradjc: It's convenient to include the error, because rust forces us to handle cases that can "never" happen. I think it is better to have the error than do something like hardcode panics into the core kernel binary which will never be triggered, but still take up space.

From a capsule author's point of view, I think it's pretty clear that it's not an error they should handle, but could report upstream to us if they ever saw it.

However, we do not want capsule authors to include a debug or panic, because that just re-creates the size issue (and probably worse). But that is also true of handling this return value at all, unless a capsule can do something meaningful it should just ignore the return value.

kernel/src/process_standard.rs
```diff
@@ -213,26 +213,36 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
...
+                        // insufficient space in the ring buffer
+                        Err(ErrorCode::NOMEM)
+                    }
+                }
+            },
+            || {
+                // Make a note that we lost this upcall if the enqueue function fails.
+                self.debug.map(|debug| {
+                    debug.dropped_upcall_count += 1;
+                });
```
bradjc, line 242: This should happen in the false case on line 231, shouldn't it?

kernel/src/process_standard.rs
```diff
@@ -213,26 +213,36 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
...
-        // Make a note that we lost this upcall if the enqueue function
-        // fails.
-        if ret == false {
-            self.debug.map(|debug| {
-                debug.dropped_upcall_count += 1;
-            });
-        } else {
-            self.kernel.increment_work();
-        }
-
```
bradjc, line 234: Wouldn't keeping logic like this be easier?

phil-levis:
> For reference, reasons schedule() can fail:
...
>     5. There is no room in the tasks queue
...
> Then (5) is the only case I could see returning an error, but I don't see what a capsule could do with that information.

I agree. I am worried that making this a `Result` is going to add a lot of boilerplate. The only real answer a capsule might have is to mark the upcall as pending and try to issue it again, e.g., after a timeout or on next invocation. How many capsules handle this case? If the answer is zero, then I'd suggest changing it to no return value until we have a case to work with. It might be, for example, that now that we count the number of upcalls we create a reserved upcall queue, such that an upcall fails iff there is a copy of that upcall pending. If so, the error cases for UpcallError are going to be different.

bradjc:
> I agree. I am worried that making this a `Result` is going to add a lot of boilerplate. 

Well, the boilerplate is only adding `.ok()` to the end of the schedule calls. I pushed a commit which changes capsules to prevent the warning for unhandled result so you can see the change.

>The only real answer a capsule might have is to mark the upcall as pending and try to issue it again, e.g., after a timeout or on next invocation. How many capsules handle this case? If the answer is zero, then I'd suggest changing it to no return value until we have a case to work with. It might be, for example, that now that we count the number of upcalls we create a reserved upcall queue, such that an upcall fails iff there is a copy of that upcall pending. If so, the error cases for UpcallError are going to be different.

No capsules use the return value today. However, if we make change the return to Result now, then we can update the capsule interface now as part of 2.0, and if we expand the error types later we don't have to go back and change any capsules that do not need to handle schedule errors.

phil-levis:
> > I agree. I am worried that making this a `Result` is going to add a lot of boilerplate.
> 
> Well, the boilerplate is only adding `.ok()` to the end of the schedule calls. I pushed a commit which changes capsules to prevent the warning for unhandled result so you can see the change.

I think I am `ok` with that. 

Is there any increase in code size? I've become a bit wary of flexibility that we might need in the future that increases code size now. If the increase is minimal then I'm on board.

bradjc:
> Is there any increase in code size? I've become a bit wary of flexibility that we might need in the future that increases code size now. If the increase is minimal then I'm on board.

I think it saves space. Looking through https://github.com/tock/tock/pull/2656/checks?check_run_id=3030271308 a lot of the numbers look smaller.

bradjc:
I think this is waiting on @ppannuto to review his own PR with the latest changes.

kernel/src/process_standard.rs
```diff
@@ -213,23 +213,35 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
...
+    fn enqueue_task(&self, task: Task) -> Result<(), ErrorCode> {
         // If this app is in a `Fault` state then we shouldn't schedule
         // any work for it.
         if !self.is_active() {
-            return false;
+            return Err(ErrorCode::NODEVICE);
         }
 
-        let ret = self.tasks.map_or(false, |tasks| tasks.enqueue(task));
+        let ret = self.tasks.map_or(Err(ErrorCode::BUSY), |tasks| {
```
ppannuto, line 223: Should this be `BUSY` or `FAIL`? This is really an internal kernel error that the map cell is empty, no?

ppannuto (commented):
You go away for one long weekend and your PR flies on its own :P

Looks good to me overall; one small interface question

kernel/src/process_standard.rs
bradjc: Yes, but what ErrorCode is that?

bradjc (approved):

ppannuto (commented):
image
But I would if I coult

PR #2658, opened 2021-07-08
Minor updates to grant.rs

bradjc:
### Pull Request Overview

In grant.rs:

- Remove a kernel.process_map_or() call we didn't need.
- Update and format comments.
- Rename one local variable to match new struct naming.

### Testing Strategy

Running hail app.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -474,81 +474,75 @@ impl<'a, T: Default, const NUM_UPCALLS: usize> ProcessGrant<'a, T, NUM_UPCALLS> 
                     // Now we can calculate the entire size of the grant.
                     let alloc_size = upcalls_size + upcalls_padding + grant_t_size;
 
-                    let (ptr_upcall_count, optional_ptr_first_upcall, raw_ptr_grant_nn) = processid
-                        .kernel
-                        .process_map_or(Err(Error::NoSuchApp), processid, |process| {
-                            process
-                                .allocate_grant(grant_num, driver_num, alloc_size, alloc_align)
-                                .map_or(Err(Error::OutOfMemory), |buf| {
-                                    // Number of upcalls.
-                                    let ptr_upcall_count = NonNull::cast::<usize>(buf);
+                    let (ptr_upcall_count, optional_ptr_first_upcall, raw_ptr_grant_nn) = process
+                        .allocate_grant(grant_num, driver_num, alloc_size, alloc_align)
+                        .map_or(Err(Error::OutOfMemory), |buf| {
+                            // Number of upcalls.
+                            let ptr_upcall_count = NonNull::cast::<usize>(buf);
 
-                                    // Upcall array.
+                            // Upcall array.
 
-                                    // Only create the pointer to the first upcall
-                                    // if we actually have memory for the
-                                    // SavedUpcall to exist.
-                                    let optional_ptr_first_upcall = if num_upcalls > 0 {
-                                        // # Safety
-                                        //
-                                        // It is safe to construct a *u8 pointer to
-                                        // the start of the upcalls array because we
-                                        // ensured that the memory is both valid and
-                                        // aligned by performing the allocation.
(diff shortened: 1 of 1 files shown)


kernel/src/grant.rs
```diff
@@ -371,11 +371,11 @@ impl<'a, T: Default, const NUM_UPCALLS: usize> ProcessGrant<'a, T, NUM_UPCALLS>
...
-        // code bloat from the generic function being monomorphized, as it is common
-        // to have over 50 copies of Grant::enter() in a Tock kernel (and thus 50+
-        // copies of this function). The returned Option indicates if the returned
-        // pointer still needs to be initialized (in the case where the grant was only
-        // just allocated).
+        // code bloat from the generic function being monomorphized, as it is
+        // common to have over 50 copies of Grant::enter() in a Tock kernel (and
+        // thus 50+ copies of this function). The returned Option indicates if
+        // the returned pointer still needs to be initialized (in the case where
+        // the grant was only just allocated).
```
ppannuto, line 378: Is your editor configured to do something different that `rustfmt` somehow? There's a whole bunch of these comment line-wrapping changes [It doesn't look like any actual text changed here?] that just feel like noise in this PR.

ppannuto (changes requested):
The function call removal and renaming look fine, but I think the comment changes all look like they're just noise that we should leave to `rustfmt`

hudson-ayers (approved):

kernel/src/grant.rs
bradjc: That was the main point of this PR :)

bradjc: rustfmt doesn't touch comments, and grant.rs has been updated enough recently with things moving in and out that comments ended up wrapped to all sorts of various widths. This does some prettyifying.

Also, it would be great if rustfmt would do this for us, but I suspect it's impossible to automatically detect what is a hardwrapped paragraph and what is intentionally formatted.

hudson-ayers: By default, rustfmt does not enforce wrapping comments or maximum comment length.

There is an unstable feature (https://github.com/rust-lang/rustfmt/issues/3349) that can be enabled to make rustfmt enforce this, but I am not sure it is worth using unstable rustfmt just for comment formatting.

Given that rustfmt does not litigate this stuff, I think these changes are reasonable.

bradjc: Can we have `comment_min_width` too? :)

ppannuto (approved):
bors r+

Sigh. `rustfmt` suffers from the 'solves so many problems, I get angry when it doesn't solve all the problems, problem' :)

lschuermann:
Even if we don't enforce it with rustfmt, do we have a rough guideline on a maximum line length to follow; for code and / or for comments? I'll try to stay in the 80 characters limit my editor suggests, but some parts of Tock have rather insane indentation levels. If there is a magic number being recommended, it can serve as a basis for arguing that some parts could benefit from refactoring; however where it makes sense to exceed this limit it could still be allowed. This, as far as I know, is similar to codestyle enforcement policies in Linux. This is probably not the right place to discuss this, but roughly related.

hudson-ayers:
> Even if we don't enforce it with rustfmt, do we have a rough guideline on a maximum line length to follow; for code and / or for comments? I'll try to stay in the 80 characters limit my editor suggests, but some parts of Tock have rather insane indentation levels. If there is a magic number being recommended, it can serve as a basis for arguing that some parts could benefit from refactoring; however where it makes sense to exceed this limit it could still be allowed. This, as far as I know, is similar to codestyle enforcement policies in Linux. This is probably not the right place to discuss this, but roughly related.

I don't think we have an agreed-upon guideline anywhere. I generally try to stay below 100 characters since that is the maximum line length which rustfmt allows for code.

PR #2663, opened 2021-07-09
Auto implement Group Traits

jettr:
### Pull Request Overview

Instead of each object providing an empty implementation for a group trait, auto implement the group trait if the object has all of the necessary sub-traits.

### Testing Strategy

`cargo check` on each CL

### Documentation Updated

- No documnetation update

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/rf233.rs
+++ b/capsules/src/rf233.rs
@@ -1162,8 +1162,6 @@ impl<'a, S: spi::SpiMasterDevice> RF233<'a, S> {
     }
 }
 
-impl<S: spi::SpiMasterDevice> radio::Radio for RF233<'_, S> {}
-
 impl<S: spi::SpiMasterDevice> radio::RadioConfig for RF233<'_, S> {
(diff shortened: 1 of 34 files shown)


lschuermann:
This is great!

The only thing which this would effectively prevent is that implementers of these HILs can't opt out of certain marker traits. I don't think that's an issue, but want to mention it nonetheless.

alistair23 (approved):

bradjc (approved):
This seems like a net good thing.

hudson-ayers (approved):

jettr:
> The only thing which this would effectively prevent is that implementers of these HILs can't opt out of certain marker traits. I don't think that's an issue, but want to mention it nonetheless.

I agree. We should only do this for convenience group traits, not marker traits. That is actually why I didn't implement this for the `Timestamp` trait since it seems like it might be a marker trait on `Time` instead of a group trait

PR #2669, opened 2021-07-16
[rfc] Comptibility header (v1)

alexandruradovici:
### Pull Request Overview

This pull request adds a compatibility header to the TBF. This is the first version of such a header.

The idea is to add the ability to the kernel to ignore and not load the processes that do not meet the kernel compatibility constraints defined in this header.

This defines a fixed header that contains:
- the exact kernel major version
- the minimum kernel minor version

Processes that provide a `KernelVersion` header state that the exact required kernel version and the
minimum required kernel minor version. For instance, if a process specifies kernel version 2.12, this means it requires a kernel
version that is >= 2.12 and < 3.0.

### Testing Strategy

This was tested using a Microbit v2 and libtock-c with the new elf2tab (https://github.com/tock/elf2tab/pull/30).

EDIT: Tockloader seems to send over to the board unknown TLV headers as this works with the unmodified tockloader.

### TODO or Help Wanted

1. I'm not sure where to define the running kernel version. For now this is defined as two constants in `kernel/src/lib.rs`. Another idea was to define them within the `Config` structure. The printed kernel version is taken from the git release, which, at least for the development tree, is difficult to parse
2. Improvement to the debug messages

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/TockBinaryFormat.md
+++ b/doc/TockBinaryFormat.md
@@ -16,6 +16,7 @@
     + [`5` Fixed Addresses](#5-fixed-addresses)
     + [`6` Permissions](#6-permissions)
     + [`7` Persistent ACL](#7-persistent-acl)
+    + [`8` Compatibility](#8-compatibility)
 - [Code](#code)
 
 <!-- tocstop -->
@@ -167,6 +168,14 @@ struct TbfHeaderV2PersistentAcl {
     read_ids: [u32],
     access_length: u16,
     access_ids: [u32],
+
+// Compatibility
+struct TbfHeaderV2Compatibility {
+    base: TbfHeaderTlv,
+    length: u16,
+    abi_version: u16,
+    kernel_major: u8,
+    kernel_minor: u8
 }
 ```
 
@@ -430,3 +439,25 @@ but the specific address is determined by the platform. Code in the binary
 should be able to execute successfully at any address, e.g. using position
 independent code.
 
+#### `8` Compatibility
+
+The `compatibility` header is designed to prevent the kernel
+from running applications that are not compatible with it.
+
+It defins the following three items:
+* `ABI version` is the ABI (system call) version that the kernel provides. Tock 1.x 
+ provides ABI version 1 while Tock 2.0 provides ABI version 2. The ABI version is not
+ necessarly related to the kernel major number. The ABI version number increases 
+ each time the systm call numbers, systm call arguments or upcall parameters change.
+* `KM` is the kernel major number (for Tock 2.0, KM is 2)
+* `Km` is the kernel minor number (for Tock 2.0, Km is 0)
+
+Setting any of the `KM` or `Km` values to 0 indicates that the application
+works regadless of the actual value. The `ABI version` is required.
+
+```
+0             2             4             6             8
++-------------+-------------+---------------------------+
+| Type (7)    | Length (4)  | ABI version |  KM  |  Km  |
++-------------+-------------+---------------------------+
+```
--- a/doc/TockBinaryFormat.md
+++ b/doc/TockBinaryFormat.md
@@ -172,7 +172,6 @@ struct TbfHeaderV2PersistentAcl {
 // Kernel Version
 struct TbfHeaderV2KernelVersion {
     base: TbfHeaderTlv,
-    length: u16,
     major: u16,
     minor: u16
 }
--- a/doc/TockBinaryFormat.md
+++ b/doc/TockBinaryFormat.md
@@ -173,8 +173,8 @@ struct TbfHeaderV2PersistentAcl {
 struct TbfHeaderV2KernelVersion {
     base: TbfHeaderTlv,
     length: u16,
-    major: u8,
-    minor: u8
+    major: u16,
+    minor: u16
 }
 ```
 
--- a/doc/TockBinaryFormat.md
+++ b/doc/TockBinaryFormat.md
@@ -435,7 +435,7 @@ different regions.
 The `compatibility` header is designed to prevent the kernel
 from running applications that are not compatible with it.
 
-It defins the following two items:
+It defines the following two items:
 * `Kernel major` or `V` is the kernel major number (for Tock 2.0, it is 2)
 * `Kernel minor` or `v` is the kernel minor number (for Tock 2.0, it is 0)
 
--- a/doc/TockBinaryFormat.md
+++ b/doc/TockBinaryFormat.md
@@ -437,7 +437,7 @@ but the specific address is determined by the platform. Code in the binary
 should be able to execute successfully at any address, e.g. using position
 independent code.
 
-#### `8` Compatibility
+#### `8` Kernel Version
 
 The `compatibility` header is designed to prevent the kernel
 from running applications that are not compatible with it.
@@ -457,7 +457,7 @@ data between kernel and userspace and the the system call numbers.
 ```
 0             2             4             6             8
 +-------------+-------------+---------------------------+
-| Type (7)    | Length (4)  | Kernel major| Kernel minor|
+| Type (8)    | Length (4)  | Kernel major| Kernel minor|
 +-------------+-------------+---------------------------+
 ```
 
--- a/doc/TockBinaryFormat.md
+++ b/doc/TockBinaryFormat.md
@@ -444,7 +444,7 @@ Apps defining this header are compatible with kernel version ^V.v (>= V.v and < 
 The kernel version header refers only to the ABI and API exposed by the kernel 
 itself, it does not cover API changes within drivers. 
 
-A kernel major and minor version  guarantees the  ABI for exchanging 
+A kernel major and minor version guarantees the ABI for exchanging 
 data between kernel and userspace and the the system call numbers.
 
 ```
--- a/doc/TockBinaryFormat.md
+++ b/doc/TockBinaryFormat.md
@@ -448,6 +448,12 @@ It defins the following two items:
 
 Apps defining this header are compatible with kernel version ^V.v (>= V.v and < (V+1).0)
 
+The kernel version header refers only to the ABI and API exposed by the kernel 
+itself, it does not cover API changes within drivers. 
+
+A kernel major and minor version  guarantees the  ABI for exchanging 
+data between kernel and userspace and the the system call numbers.
+
 ```
 0             2             4             6             8
 +-------------+-------------+---------------------------+
--- a/doc/TockBinaryFormat.md
+++ b/doc/TockBinaryFormat.md
@@ -430,13 +430,6 @@ errors over the network, and once the reported errors are acked erases them
 from the log. In this case `access_ids` allow an app to erase multiple
 different regions.
 
-## Code
-
-The process code itself has no particular format. It will reside in flash,
-but the specific address is determined by the platform. Code in the binary
-should be able to execute successfully at any address, e.g. using position
-independent code.
-
 #### `8` Kernel Version
 
 The `compatibility` header is designed to prevent the kernel
--- a/doc/TockBinaryFormat.md
+++ b/doc/TockBinaryFormat.md
@@ -16,7 +16,7 @@
     + [`5` Fixed Addresses](#5-fixed-addresses)
     + [`6` Permissions](#6-permissions)
     + [`7` Persistent ACL](#7-persistent-acl)
-    + [`8` Compatibility](#8-compatibility)
+    + [`8` Kernel Version](#8-kernel-version)
 - [Code](#code)
 
 <!-- tocstop -->
@@ -169,13 +169,12 @@ struct TbfHeaderV2PersistentAcl {
     access_length: u16,
     access_ids: [u32],
 
-// Compatibility
-struct TbfHeaderV2Compatibility {
+// Kernel Version
+struct TbfHeaderV2KernelVersion {
     base: TbfHeaderTlv,
     length: u16,
-    abi_version: u16,
-    kernel_major: u8,
-    kernel_minor: u8
+    major: u8,
+    minor: u8
 }
 ```
 
@@ -444,20 +443,23 @@ independent code.
 The `compatibility` header is designed to prevent the kernel
 from running applications that are not compatible with it.
 
-It defins the following three items:
-* `ABI version` is the ABI (system call) version that the kernel provides. Tock 1.x 
- provides ABI version 1 while Tock 2.0 provides ABI version 2. The ABI version is not
- necessarly related to the kernel major number. The ABI version number increases 
- each time the systm call numbers, systm call arguments or upcall parameters change.
-* `KM` is the kernel major number (for Tock 2.0, KM is 2)
-* `Km` is the kernel minor number (for Tock 2.0, Km is 0)
+It defins the following two items:
+* `Kernel major` or `V` is the kernel major number (for Tock 2.0, it is 2)
+* `Kernel minor` or `v` is the kernel minor number (for Tock 2.0, it is 0)
 
-Setting any of the `KM` or `Km` values to 0 indicates that the application
-works regadless of the actual value. The `ABI version` is required.
+Apps defining this header are compatible with kernel version ^V.v (>= V.v and < (V+1).0)
 
 ```
 0             2             4             6             8
 +-------------+-------------+---------------------------+
-| Type (7)    | Length (4)  | ABI version |  KM  |  Km  |
+| Type (7)    | Length (4)  | Kernel major| Kernel minor|
 +-------------+-------------+---------------------------+
 ```
+
+
+## Code
+
+The process code itself has no particular format. It will reside in flash,
+but the specific address is determined by the platform. Code in the binary
+should be able to execute successfully at any address, e.g. using position
+independent code.
--- a/kernel/src/lib.rs
+++ b/kernel/src/lib.rs
@@ -89,6 +89,10 @@
 #![warn(unreachable_pub)]
 #![no_std]
 
+// Define the kernel major and minor versions
+pub const MAJOR: u16 = 2;
+pub const MINOR: u16 = 0;
+
 pub mod capabilities;
 pub mod collections;
 pub mod component;
--- a/kernel/src/process_standard.rs
+++ b/kernel/src/process_standard.rs
@@ -1411,7 +1411,7 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
             }
         } else {
             if enforce_kernel_version {
-                // If enforing the kernel version is requested, and the `KernelVersion` header is not present,
+                // If enforcing the kernel version is requested, and the `KernelVersion` header is not present,
                 // we prevent the process from loading.
                 if config::CONFIG.debug_load_processes {
                     debug!(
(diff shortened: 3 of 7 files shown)


doc/TockBinaryFormat.md
```diff
@@ -155,6 +156,14 @@ struct TbfHeaderV2Permissions {
     length: u16,
     perms: [TbfHeaderDriverPermission],
 }
+
+// Kernel Version
+struct TbfHeaderV2KernelVersion {
+    base: TbfHeaderTlv,
+    length: u16,
+    major: u8,
+    minor: u8
```
hudson-ayers, line 165: These are `u8`s here but seem to be `u16`s in the tock-tbf code?

alexandruradovici: Fixed, sorry for that

doc/TockBinaryFormat.md
```diff
@@ -155,6 +156,14 @@ struct TbfHeaderV2Permissions {
     length: u16,
     perms: [TbfHeaderDriverPermission],
 }
+
+// Kernel Version
+struct TbfHeaderV2KernelVersion {
+    base: TbfHeaderTlv,
+    length: u16,
```
bradjc, line 163: Don't need this length.

kernel/src/lib.rs
```diff
@@ -89,6 +89,10 @@
 #![warn(unreachable_pub)]
 #![no_std]
 
+// Define the kernel major and minor versions
+pub const MAJOR: u16 = 2;
+pub const MINOR: u16 = 0;
```
bradjc, line 94: We need some way to remember to update this.

doc/TockBinaryFormat.md
```diff
@@ -364,10 +372,28 @@ included, so long as no `offset` is repeated for a single driver. When
...
+#### `7` Kernel Version
+
+The `Kernel Version` header is designed to prevent the kernel
+from running applications that are not compatible with it.
+
+It defins the following two items:
+* `Kernel major` or `V` is the kernel major number (for Tock 2.0, it is 2)
+* `Kernel minor` or `v` is the kernel minor number (for Tock 2.0, it is 0)
+
+Apps defining this header are compatible with kernel version ^V.v (>= V.v and < (V+1).0)
```
phil-levis, line 384: I am wary of this kind of specification. For example, it means that we can't change a system call API for a particular device number without a major number increment. Defining "compatible" seems difficult. 

Instead, I'd suggest that we add this to the end of the main header. It should either be present in every app, or not. It should include only a major version number, which relates to the ABI with the kernel. Anything further than that (compatibility) can be determined by the app itself, e.g., by probing if devices are installed.

bradjc: The intent is for this (i.e. TLV id 7) to only specify kernel ABI. We wanted major and minor, because Tock 2.1 might include a new allow call that Tock 2.0 does not. We do need to make this clear in the docs.

We also thought that a TLV was an appropriate option for two reasons:
1. We're not entirely sure this is the long term best way to handle this sort of compatibility checking. Using a TLV makes it easy to deprecate this in the future and possibly replace it with something else.
2. Some boards/deployments might not want to include any compatibility information (and save the bytes in the header). They can easily omit the TLV.

alexandruradovici: I detailed this in the documentation.

kernel/src/process_standard.rs
```diff
@@ -1385,14 +1386,46 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
...
+            if enforce_kernel_version {
+                // If enforing the kernel version is requested, and the `KernelVersion` header is not present,
+                // we prevent the process from loading.
+                if config::CONFIG.debug_load_processes {
+                    debug!(
+                            "WARN process {:?} not loaded as it has no kernel version header, please upgrade to elf2tab >= 0.8.0",
+                            process_name.unwrap_or ("(no name")
+                        );
+                }
+                return Ok((None, remaining_memory));
```
bradjc, line 1425: I think we should return an error here. The caller asked to enforce kernel version, and the check failed.

kernel/src/process_standard.rs
```diff
@@ -1385,14 +1386,46 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
...
+                if config::CONFIG.debug_load_processes {
+                    debug!("WARN process {:?} not loaded as it requires kernel version {} (running kernel {}.{})", process_name.unwrap_or("(no name)"), major, crate::MAJOR, crate::MINOR);
+                }
+                return Ok((None, remaining_memory));
+            } else if crate::MINOR < minor {
+                // If the kernel major version is the same, we compare the kernel minor version. The current
+                // running kernel minor version has to be greater or equal to the one that the process
+                // has requested. If not, we prevent the process from loading.
+                if config::CONFIG.debug_load_processes {
+                    debug!("WARN process {:?} not loaded as it requires kernel version at least {}.{}, (running kernel {}.{})", process_name.unwrap_or("(no name)"), major, minor, crate::MAJOR, crate::MINOR);
```
bradjc, line 1411: Do we need to separate these cases?

kernel/src/process_standard.rs
```diff
@@ -1385,14 +1386,46 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
...
+                }
+                return Ok((None, remaining_memory));
```
bradjc, line 1413: I think this should error if `enforce_kernel_version` is true.

kernel/src/process_standard.rs
alexandruradovici: If we want to display different errors, but I suppose we could merge this into a single if.

kernel/src/process_standard.rs
alexandruradovici: My understanding is that if this returns an error, `load_processes_advanced` will stop loading processes, not just skip this process.

```rust
  ProcessStandard::create(
                    kernel,
                    chip,
                    entry_flash,
                    header_length as usize,
                    version,
                    remaining_memory,
                    fault_policy,
                    enforce_kernel_version,
                    i,
            )?
```

bradjc: That's correct. But without an error this will silently fail, and it's hard to differentiate an app that just doesn't do anything from one that was never actually loaded. Presumably no user who wants the kernel version compatibility checked wants to just ignore those errors.

alexandruradovici: Agree, but on the other hand this would make "some apps" (all apps after the one that cannot be loaded) silently fail.

What about always showing this warnings if enforce kernel version is true?

kernel/src/process_standard.rs
```diff
@@ -1385,14 +1386,42 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
...
+                // running kernel minor version has to be greater or equal to the one that the process
+                // has requested. If not, we prevent the process from loading.
+                if config::CONFIG.debug_load_processes {
+                    debug!("WARN process {:?} not loaded as it requires kernel version >= {}.{} and < {}.0, (running kernel {}.{})", process_name.unwrap_or("(no name)"), major, minor, (major+1), crate::MAJOR, crate::MINOR);
+                }
+                return Ok((None, remaining_memory));
+            }
+        } else {
+            if enforce_kernel_version {
+                // If enforing the kernel version is requested, and the `KernelVersion` header is not present,
```
ppannuto, line 1413:
```suggestion
                // If enforcing the kernel version is requested, and the `KernelVersion` header is not present,
```

doc/TockBinaryFormat.md
```diff
@@ -364,10 +372,34 @@ included, so long as no `offset` is repeated for a single driver. When
...
+```
+
+
 ## Code
 
 The process code itself has no particular format. It will reside in flash,
 but the specific address is determined by the platform. Code in the binary
 should be able to execute successfully at any address, e.g. using position
 independent code.
-
```
alistair23, line 373: Don't remove this

kernel/src/lib.rs
alistair23: Could we auto generate/update this from git tags?

kernel/src/process_standard.rs
bradjc: What is the use case for knowing that you have an incompatible app but continuing to run the others? I suspect that the most likely case is the user only has one app (or if there are multiple they are all incompatible) and they are going to fix it immediately.

doc/TockBinaryFormat.md
alexandruradovici: Do you mean the empty new line?

kernel/src/process_standard.rs
alexandruradovici: The user might have loaded one of the apps from a tab using tockloader. The issue is that it will load all apps up to the one that does not meet the requirements or has no header. In case we do not want to load any app, maybe it would be a good idea to panic as it is not deterministic which apps will run.

bradjc: Most boards do not panic, but do print a warning/error when load_processes fails. I've heard anecdotally that the checks for an app that does not meet ram/flash position requirements which cause load_processes to fail have worked well.

kernel/src/lib.rs
bradjc: Generate doesn't work if someone downloads the zip file and there is no git repo. But something that at least checked
somehow would be nice.

doc/TockBinaryFormat.md
hudson-ayers: yes he does

kernel/src/process_standard.rs
alexandruradovici: I got it now :)  

I just saw that most boards report a debug message upon an error.

kernel/src/process_standard.rs
```diff
@@ -1340,6 +1340,7 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
         app_version: u16,
         remaining_memory: &'a mut [u8],
         fault_policy: &'static dyn ProcessFaultPolicy,
+        enforce_kernel_version: bool,
```
bradjc, line 1343:
```suggestion
        require_kernel_version: bool,
```

kernel/src/process_standard.rs
```diff
@@ -1386,14 +1387,44 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
...
+                // has requested. If not, we prevent the process from loading.
+                if config::CONFIG.debug_load_processes {
+                    debug!("WARN process {:?} not loaded as it requires kernel version >= {}.{} and < {}.0, (running kernel {}.{})", process_name.unwrap_or("(no name)"), major, minor, (major+1), crate::MAJOR, crate::MINOR);
+                }
+                return Err(ProcessLoadError::IncompatibleKernelVersion {
+                    version: Some((major, minor)),
+                });
+            }
+        } else {
+            if enforce_kernel_version {
```
bradjc, line 1415:
```suggestion
            if require_kernel_version {
```

I think the current logic makes sense: if a process tells us it requires a newer kernel version we always error in that case. If the process doesn't tell us what it requires we assume the process is fine, unless explicitly told not to. 

We may someday want to add a "force allow incompatible apps" or something, but we can make that change if the need ever arises.

kernel/src/process_utilities.rs
```diff
@@ -45,6 +45,14 @@ pub enum ProcessLoadError {
         expected_address: u32,
     },
 
+    /// A process might provide a KernelVersion TBF header stating the
+    /// compatible kernel version (^major.minor). If `load_processes_advanced` is
+    /// called with `enforce_kernel_version` set to true, kernel will stop loading
+    /// processes when it encounters either:
+    ///    - a process that does not provide the KernelVersion TBF header
+    ///    - a process that requested an incompatible version of the kernel
```
bradjc, line 53:
```suggestion
    /// A process requires a newer version of the kernel or did not specify
    /// a required version. Processes can include the KernelVersion TBF header stating
    /// their compatible kernel version (^major.minor).
    ///
    /// Boards may not require processes to include the KernelVersion TBF header, and
    /// the kernel supports ignoring a missing KernelVersion TBF header. In that case,
    /// this error will not be returned for a process missing a KernelVersion TBF
    /// header.
    ///
    /// `version` is the `(major, minor)` kernel version the process indicates it
    /// requires. If `version` is `None` then the process did not include the
    /// KernelVersion TBF header.
```

kernel/src/process_utilities.rs
```diff
@@ -247,3 +269,30 @@ pub fn load_processes<C: Chip>(
 
     Ok(())
 }
+
+/// This is a wrapper function for `load_processes_advanced` that uses
+/// the default arguments that mainstream boards should provide.
+///
+/// Default arguments are:
+///  - `enforce_kernel_version`: prevent loading processes that do not provide a `KernelVersion`
```
bradjc, line 277:
```suggestion
///  - `require_kernel_version`: prevent loading processes that do not provide a `KernelVersion`
```

libraries/tock-tbf/src/types.rs
```diff
@@ -434,4 +462,14 @@ impl TbfHeader {
             start => Some(start),
         }
     }
+
+    pub fn get_kernel_version(&self) -> Option<(u16, u16)> {
```
bradjc, line 466:
```suggestion
    /// Get the minimum compatible kernel version this process requires.
    /// Returns `None` if the kernel compatibility header is not included.
    pub fn get_kernel_version(&self) -> Option<(u16, u16)> {
```

kernel/src/process_utilities.rs
```diff
@@ -125,13 +145,14 @@ impl fmt::Debug for ProcessLoadError {
...
 /// creation.
-pub fn load_processes<C: Chip>(
+pub fn load_processes_advanced<C: Chip>(
     kernel: &'static Kernel,
     chip: &'static C,
     app_flash: &'static [u8],
     app_memory: &mut [u8], // not static, so that process.rs cannot hold on to slice w/o unsafe
     procs: &'static mut [Option<&'static dyn Process>],
     fault_policy: &'static dyn ProcessFaultPolicy,
+    enforce_kernel_version: bool,
```
bradjc, line 155:
```suggestion
    require_kernel_version: bool,
```

kernel/src/process_utilities.rs
```diff
@@ -218,6 +239,7 @@ pub fn load_processes<C: Chip>(
                     version,
                     remaining_memory,
                     fault_policy,
+                    enforce_kernel_version,
```
bradjc, line 242:
```suggestion
                    require_kernel_version,
```

kernel/src/process_utilities.rs
```diff
@@ -98,6 +106,18 @@ impl fmt::Debug for ProcessLoadError {
                 actual_address, expected_address
             ),
 
+            ProcessLoadError::IncompatibleKernelVersion { version } => match version {
+                Some((major, minor)) => write!(
+                    f,
+                    "App is incomatible with the kernel. Running: {}.{}, Requested: {}.{}",
```
bradjc, line 112:
```suggestion
                    "Process is incompatible with the kernel. Running: {}.{}, Requested: {}.{}",
```

kernel/src/process_utilities.rs
```diff
@@ -98,6 +106,18 @@ impl fmt::Debug for ProcessLoadError {
...
+                    crate::MAJOR,
+                    crate::MINOR,
+                    major,
+                    minor
+                ),
+                None => write!(f, "App did not provide a TBF kernel version header"),
```
bradjc, line 118:
```suggestion
                None => write!(f, "Process did not provide a TBF kernel version header"),
```

kernel/src/process_standard.rs
```diff
@@ -1386,14 +1387,44 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
                         "Process not enabled flash={:#010X}-{:#010X} process={:?}",
                         app_flash.as_ptr() as usize,
                         app_flash.as_ptr() as usize + app_flash.len() - 1,
-                        process_name
+                        process_name.unwrap_or("(no name)")
```
bradjc, line 1390: I don't understand this change. Is this related?

doc/TockBinaryFormat.md
```diff
@@ -423,10 +430,34 @@ errors over the network, and once the reported errors are acked erases them
 from the log. In this case `access_ids` allow an app to erase multiple
 different regions.
 
+#### `8` Kernel Version
+
+The `compatibility` header is designed to prevent the kernel
+from running applications that are not compatible with it.
+
+It defins the following two items:
```
bradjc, line 438:
```suggestion
It defines the following two items:
```

doc/TockBinaryFormat.md
```diff
@@ -423,10 +430,34 @@ errors over the network, and once the reported errors are acked erases them
...
+* `Kernel major` or `V` is the kernel major number (for Tock 2.0, it is 2)
+* `Kernel minor` or `v` is the kernel minor number (for Tock 2.0, it is 0)
+
+Apps defining this header are compatible with kernel version ^V.v (>= V.v and < (V+1).0)
+
+The kernel version header refers only to the ABI and API exposed by the kernel 
+itself, it does not cover API changes within drivers. 
+
+A kernel major and minor version  guarantees the  ABI for exchanging
```
bradjc, line 447:
```suggestion
A kernel major and minor version guarantees the ABI for exchanging 
```

kernel/src/process_standard.rs
alexandruradovici: Its not related, but this would display the process name instead of Some(process_name). I suggest the same change at line 1452.

kernel/src/process_utilities.rs
```diff
@@ -125,13 +151,14 @@ impl fmt::Debug for ProcessLoadError {
 /// Returns `Ok(())` if process discovery went as expected. Returns a
 /// `ProcessLoadError` if something goes wrong during TBF parsing or process
 /// creation.
-pub fn load_processes<C: Chip>(
+pub fn load_processes_advanced<C: Chip>(
```
hudson-ayers, line 154:
```suggestion
#[inline(always)]
pub fn load_processes_advanced<C: Chip>(
```

Given that load_processes has the annotation we should it include it here as well.

kernel/src/process_utilities.rs
hudson-ayers: This was marked as resolved but does not look to be changed?

kernel/src/process_utilities.rs
alexandruradovici: I added the inline option to load_processes as it is a wrapper for load_processes_advanced. The original load_processes did not have this option.

bradjc (approved):
We need to merge at least the elf2tab changes and release elf2tab before we merge this.

bradjc:
Or we could set required to `false` for now, and change it back closer to the actual 2.0 release.

kernel/src/process_utilities.rs
alexandruradovici: @hudson-ayers it's changed now, sorry I forgot to rename the variable in `process_utilities.rs`.

alexandruradovici:
> Or we could set required to `false` for now, and change it back closer to the actual 2.0 release.

As long is we change it back before we release 2.0 this should be fine.

phil-levis:
Let's do this right -- @alexandruradovici can you rebase https://github.com/tock/elf2tab/pull/30 so we can review and merge? It seems like everyone is on-board and it all looks good, let's just do it.

alexandruradovici:
I rebased https://github.com/tock/elf2tab/pull/30, I would be great if @alistair23 can take a look at it the merge was with the permissions header.

lschuermann (approved):

PR #2674, opened 2021-07-19
Notes from Tock Core WG meeting July 9th

brghena:
### Pull Request Overview

Adds notes from Core Working Group meeting on July 9th. Feel behind on this one, so this was the meeting a week and a half back!

[Rendered](https://github.com/tock/tock/blob/3274b475e87eecf0b4b9982c22b03e6543fbeb18/doc/wg/core/notes/core-notes-2021-07-09.md)

### Testing Strategy

Spell check

### TODO or Help Wanted

None

### Documentation Updated

### Formatting

- [N/A] Ran `make prepush`.

### DIFF
--- a/doc/wg/core/notes/core-notes-2021-07-09.md
+++ b/doc/wg/core/notes/core-notes-2021-07-09.md
@@ -1,0 +1,157 @@
+# Tock Core Notes 2021-07-09
+
+Attending:
+- Alexandru Radovici
+- Brad Campbell
+- Branden Ghena
+- Hudson Ayers
+- Leon Schuermann
+- Johnathan Van Why
+- Arjun Deopujari
+- Vadim Sukhomlinov
+- Anthony Quiroga
+
+## Updates
+
+* Hudson: Continued to work on code size reduction including grant stuff to save space. Working on other changes that are "free" rather than removing functionality
+
+
+* Leon: Upcall swapping PR is merged. ProcessAliasing PR is on the way too, hopefully will be merged soon. Only remaining PR for Tock 2.0 after that is kernel reorganization
+* Brad: That sounds right.
+
+
+* Brad: Released a new version of tockloader on pypi. Hasn't changed in a while but there were new commits, so it seemed like time to push. It does more autodetection of boards that are not using the tock bootloader.
+* Arjun: Tockloader fixes some bugs with flashing nano33ble
+
+
+## Tock 2.0 Progress
+* Hudson: Callback swapping PR is merged, which leaves very few things on the tracking issue. Aliasing of process-allowed memory is the big one, but it's on the brink of being ready. CRC stuff was merged (and Phil has some further stuff that has CRC fully working) so #2632 is unblocked. Just some final reviews would be good.
+* Leon: I posted a comment with the remaining question. This is significant and commits to how we handle this, so we want to make sure it's right. We do want to double-check `transmute` calls with someone in the Rust community to ensure our unsafe code is actually correct. I don't have a contact there though.
+* Leon: This looks good though and is close to being mergable. Just needs a thorough review since there have been some rebases.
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):

jrvanwhy (approved):

doc/wg/core/notes/core-notes-2021-07-09.md
```diff
@@ -0,0 +1,157 @@
...
+* Hudson: I thought we had said capsules were allowed to refuse to return a buffer. For example if we were passing it into a DMA request
+* Leon: Well, we can't use userspace buffers for DMA anyways. Capsules can refuse operations, for example if there is a complex piece of hardware. So if the buffer only goes away if the process dies, that makes things easier to handle. Otherwise _every_ operation would have to be abortable.
+* Hudson: I thought there was also the idea that a capsule can store info in an allowed buffer and could rely on it still being there. So we didn't want to let processes revoke buffers at any time.
+* Leon: Yes, but it can't be trusted, since the process could change data in the buffer whenever it wants.
+* Hudson: The reason I'm asking is that maybe the text in TRD104 is incorrect if we permit capsules to refuse an allow call. Which I thought was originally the plan.
+* Leon: I'm confused.
+* Hudson: Currently the documentation of TRD104, I believe, is that if userspace requests a buffer back, it must get it back. But that doesn't sync with my memory that allows can be refused.
```
phil-levis, line 83: Yes, this is correct. A capsule is allowed to reject an allow, return a failure result, and return the passed buffer. If it accepts the allow and returns a success result, then it must return the previous buffer. At least, this is what the semantics currently are. We could revisit the ability for capsule to reject if we want.

doc/wg/core/notes/core-notes-2021-07-09.md
```diff
@@ -0,0 +1,157 @@
...
+* Leon: I think the wording is that _if_ it gets a buffer back, it must be the right one. And the "must" restriction there is what we don't enforce.
```
phil-levis, line 84: Yes. This is stating a requirement for the implementation of a capsule. That's why it's a MUST. If the document just said that the kernel always does it, then we could just make that true. Put another way, a capsule that doesn't follow this MUST is not compliant with 104.

doc/wg/core/notes/core-notes-2021-07-09.md
```diff
@@ -0,0 +1,157 @@
...
+* Hudson: Okay. So we should submit a PR to TRD104 to modify the language a little and open an issue to enforce non swapping buffers.
```
phil-levis, line 85: This is a good discussion to have. I think it reflects a changing perspective on the degree to which capsules are checked/validated and the Tock security model.

doc/wg/core/notes/core-notes-2021-07-09.md
lschuermann: Agreed. I suppose there are two perspectives, which are a great source of confusion when they're mixed:
- the capsule implementation: a capsule MUST return the correct buffer to be compliant
- the system call / kernel behavior: applications MUST NOT rely on previously allowed buffers to be returned, as least w.r.t. to the current kernel implementation.

Unfortunately, the TRD104 currently states: _Calling a Read-Write Allow system call returns a buffer (address and length).  On the first call to a Read-Write Allow system call, the kernel returns a zero-length buffer. Subsequent successful calls to Read-Write Allow return the previous buffer passed._

This, to me, seems very much written from an ABI perspective, outlining the behavior of the kernel in response to an application's actions. And here, we definitely do not want to give the guarantee that the previous buffers is going to be returned in every case, because we're currently not enforcing that.

I will open a PR to rephrase this so that it's clear that (a) a capsule must return the previously passed in buffer, and (b) that an application cannot simply rely on that fact, in the sense of it being enforced by the kernel.

lschuermann: Just opened #2692 which tried to incorporate these details into the specification. Sorry it took me a bit longer.

PR #2678, opened 2021-07-21
Add SPI to RP2040 

alexandruradovici:
### Pull Request Overview

This pull request adds SPI to the RP2040.

### Testing Strategy

This pull request was tested using a Raspberry Pi Pico.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


chips/rp2040/src/chip.rs
```diff
@@ -147,9 +149,14 @@ impl Rp2040DefaultPeripherals<'_> {
...
             adc: adc::Adc::new(),
+            spi0: spi::Spi::new_spi0(),
             sysinfo: sysinfo::SysInfo::new(),
         }
     }
+
+    pub fn set_clocks(&'a self) {
+        self.spi0.set_clocks(&self.clocks);
+    }
 }
```
bradjc, line 160: Wouldn't it be easier to just pass `clocks` in as an argument to `spi::new()`?

bradjc (approved):

chips/rp2040/src/chip.rs
alexandruradovici: I don't think so, it is a circular dependency, as peripherals has  the clock field. I can try though.

hudson-ayers: Yeah, that won't work. I used a `setup_circular_dependencies()` method to address that issue for some other chips, which at least made it only require a single call in the case where there were several circular dependencies.

hudson-ayers (approved):

chips/rp2040/src/chip.rs
alexandruradovici: I followed your code 😃

PR #2625, opened 2021-06-17
WiFiNNA (Arduino WiFi) HIL and capsule

alexandruradovici:
### Pull Request Overview

This pull request is a RFC for defining a WiFi HIL.

Boards like the Arduino Connect RP 2040 have an integrated WiFi chip. This handles all the WiFi connectivity and socket tasks. We are not sure how to design ths HILs required so we are looking for a lot of feedback.

### Testing Strategy

N/A

### TODO or Help Wanted

We are lloking for a lot of feedback on how to design the HILs for WiFi.

### Documentation Updated

- [ ] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [ ] Ran `make prepush`.

PR #2681, opened 2021-07-22
Add Arduino Nano RP2040 Connect

alexandruradovici:
### Pull Request Overview

This pull request adds the initial port for the Arduino Nano RP2040 Connect board based on the Raspberry Pi's RP2040 MCU.

Supported peripherals:
- [x] GPIO
- [x] Interrupts
- [x] ADC
- [x] Timers
- [x] SPI

The board's MCU is identical to the Raspberry Pi Pico, except for the flash memory. The board has 16 MB of flash.

The main purpose for adding this board is the WiFi support (still work-in-progress as it takes some time to decide upon a WiFi - #2625 and Socket HIL).

Until the `tock-bootloader` is ported to this board, flashing can be done using a flash storage folder mounted
in BOOTSEL mode and copying UF2 file to it.

While the board exposes SWD pads that can be used through bit-banging from a Raspberry Pi, these pads are hard to solder as they are located on the back of the board (the side that is *in contact* with the breadboard). This is why the PR opts out of using openocd.

### Testing Strategy

This pull request was tested using a Nano RP2040.

### TODO or Help Wanted

Feedback for the readme documentation, most probably some text rephrasing.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


boards/nano_rp2040/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "nano_rp2040"
```
bradjc, line 2: Seems like we should include the "connect" in case they create a non wireless version.

boards/nano_rp2040_connect/Makefile
```diff
@@ -0,0 +1,55 @@
+# Makefile for building the tock kernel for the Raspberry Pi Pico board.
```
bradjc, line 1:
```suggestion
# Makefile for building the tock kernel for the Arduino Nano RP2040 Connect board.
```

boards/nano_rp2040_connect/src/main.rs
```diff
@@ -0,0 +1,459 @@
+//! Tock kernel for the Raspberry Pi Pico.
```
bradjc, line 1:
```suggestion
//! Tock kernel for the Arduino Nano RP2040 Connect.
```

lschuermann (commented):
Apart from @bradjc's comments, this looks pretty good.

boards/nano_rp2040/Cargo.toml
alexandruradovici: I renamed the crate

bradjc (dismissed):

bradjc (approved):

boards/nano_rp2040_connect/Makefile
```diff
@@ -0,0 +1,55 @@
...
+	elf2uf2 $< $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).uf2
+	if [ -d $(BOOTSEL_FOLDER) ];
+	then
+		cp $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.uf2 $(BOOTSEL_FOLDER)"
+	else
+		@echo Please edit the BOOTSEL_FOLDER variable to point to you Nano RP2040 Flash Drive Folder
+	fi
+
+.PHONY: program
+program: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
```
lschuermann, line 43: Can you rename this Make-rule to `flash-app` (or something along those lines)? IIRC we settled on -- for the kernel -- "flash" if it's simply writing a binary using some flashing tool and "program" to use `tockloader`. Or am I missing something @bradjc?

boards/nano_rp2040_connect/Makefile
```diff
@@ -0,0 +1,55 @@
...
+
+include ../Makefile.common
+
+OPENOCD=openocd
+OPENOCD_OPTIONS=-f openocd.cfg
+
+KERNEL=$(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
+KERNEL_WITH_APP=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/release/$(PLATFORM)-app.elf
+
+BOOTSEL_FOLDER=/media/$(USER)/RP2040
```
lschuermann, line 15: I suppose we don't want to define this here? It's better to just show the user the message that `BOOTSEL_FOLDER` is not set than to assume a path and error out.

boards/nano_rp2040_connect/Makefile
```diff
@@ -0,0 +1,55 @@
+# Makefile for building the tock kernel for the Arduino Nano RP2040 Connect board.
+
+TOCK_ARCH=cortex-m0p
+TARGET=thumbv6m-none-eabi
+PLATFORM=nano_rp2040_connect
+
+include ../Makefile.common
+
+OPENOCD=openocd
```
lschuermann, line 9: Are we actually using openocd here?

lschuermann (approved):
I'd prefer for the Makefile comments to be addressed, otherwise looks good.

boards/nano_rp2040_connect/Makefile
lschuermann: I think I'm starting to understand why this happens -- I've seen some other makefiles which had been doing this. I believe `program` in this context is confused with a synonym for `application`, i.e. something running on the system instead of the verb (to) program, i.e. writing something to the board.

boards/nano_rp2040_connect/Makefile
alexandruradovici: This should be the default for Ubuntu, I think that for more inexperienced users this might be useful.

lschuermann: Okay. We should at least make this an overrideable default then (i.e. use `?=`).

alexandruradovici: Done

boards/nano_rp2040_connect/Makefile
alexandruradovici: No, I deleted it for now.

boards/nano_rp2040_connect/Makefile
alexandruradovici: Yep, renamed it.

bradjc (dismissed):

bradjc (approved):

lschuermann (approved):

PR #2685, opened 2021-07-22
hil/gpio: add Copy and Clone to enums

jettr:
Derive Copy and Clone on the remaining 2 out of 5 enums on the gpio hil trait.

We have some downstream code that puts these enums in a `Cell` so having `Copy` is very nice

### Documentation Updated

None

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/hil/gpio.rs
+++ b/kernel/src/hil/gpio.rs
@@ -12,7 +12,7 @@ pub enum FloatingState {
 }
 
 /// Enum for selecting which edge to trigger interrupts on.
-#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):
The GPIO HIL is not finalized, so I think this should be easier than the time changes.

phil-levis (approved):

ppannuto (approved):

ppannuto:
It's P-Significant as it technically changes a HIL, but I think this should be pretty non-controversial

brghena (approved):

Issue #2676, opened 2021-07-20
virtual_alarm will never fire client if dt is max ticks


jettr:
Looking at the [alarm callback within virtual_alarm][0], you can see that if `dt` is `max_ticks`, then `now` will always be between `reference` and `reference + dt`. This means we will never fire the client callback for a `max_tick` `dt`. This problem also extends to ticks near `max_ticks` if the bottom half of virtual_alarm handler doesn't run immediately and gets a `now` time that is already past the original reference.I see a couple of potential solutions:

* When someone tries to schedule a new alarm for ~`max_ticks`, we reduce that by some factor to ensure that the bottom half of virtual alarm handler has time to fire and loop through clients before the `now` goes back into `[reference, reference + dt)`
  *   This is not very desirable since the amount of ticks for each system and timer should be tuned to their parameters
* Add logic to detect if we are setting a `dt` with the highest bit available set, then toggle that bit and set the alarm for half of the `max_tick` duration. When we fire the alarm, look for all extended `dt`s that need to be set again, and the wait the remaining time left from the `dt`.
  *  This is more robust and allows for waiting max_tick without having a race window to worry about
  *  This takes more bookeeping memory to know if the future `dt` we are waiting for is an extended `dt` that need to have its top bit flipped again

We are especially vulnerability to `max_tick` being used for `set_alarm` functions since the helper `ticks_from_<time>` function will saturate at `max_ticks` if you try to convert too large of a time.

[0]:https://github.com/tock/tock/blob/0c8c7127a0e820966a5336c79d961846765bd926/capsules/src/virtual_alarm.rs#L199

jettr:
I wrote a unit test example that shows the issue at https://github.com/jettr/tock/tree/virtual-alarm

The bug happens when there is any time delay from when the hardware timer fires and the bottom half of the virtual timer handler runs (which is expected on real hardware with 1MHz timers).

phil-levis:
> Looking at the [alarm callback within virtual_alarm](https://github.com/tock/tock/blob/0c8c7127a0e820966a5336c79d961846765bd926/capsules/src/virtual_alarm.rs#L199), you can see that if `dt` is `max_ticks`, then `now` will always be between `reference` and `reference + dt`. 

I don't think this is true. E.g. suppose I have `now = 100` and a `dt = max_ticks`. Then `start=100` and `end=99`. The inequalities on `within_range` are defined such:

> Returns whether the value is in the range of [`start, `end`)

[...]

PR #2688, opened 2021-07-23
Add new time trd with ConvertTicks trait

jettr:
### Pull Request Overview

Obsolete existing trd101-time in favor of this trd which supersedes it.
This trd includes information about the `ConvertTicks` trait which is
used to converts ticks to/from time units.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/reference/trd105-time.md
+++ b/doc/reference/trd105-time.md
@@ -1,12 +1,16 @@
 Kernel Time HIL
 ========================================
 
-**TRD:** 101<br/>
+**TRD:** 105<br/>
 **Working Group:** Kernel<br/>
 **Type:** Documentary<br/>
-**Status:** Obsolete <br/>
-**Obsoleted By:** 105 <br/>
-**Author:** Guillaume Endignoux, Amit Levy and Philip Levis <br/>
+**Status:** Draft <br/>
+**Obsoletes:** 101 <br/>
+**Author:** Guillaume Endignoux, Amit Levy, Philip Levis, and Jett Rink <br/>
+**Draft-Created:** 2021/07/23 <br/>
+**Draft-Modified:** 2021/07/23 <br/>
+**Draft-Version:** 1.0 <br/>
+**Draft-Discuss:** Github PR <br/>
 
 Abstract
 -------------------------------
@@ -46,8 +50,7 @@ call capsules for alarm callbacks to work across boards and chips.
 This document describes these traits, their semantics, and the
 instances that a Tock chip is expected to implement.
 
-
-2 `Time`, `Frequency`, and `Ticks` traits
+2 `Time`, `Frequency`, `Ticks`, and `ConvertTicks` traits
 ===============================
 
 The `Time` trait represents a moment in time, which is obtained by
@@ -88,6 +91,14 @@ pub trait Ticks: Clone + Copy + From<u32> + fmt::Debug + Ord + PartialOrd + Eq {
     fn within_range(self, start: Self, end: Self);
 
     fn max_value() -> Self;
+
+    /// Converts the specified val into this type if it fits, otherwise the
+    /// `max_value()` is returned
+    fn from_or_max(val: u64) -> Self;
+
+    /// Scales the ticks by the specified numerator and denominator. If the
+    /// resulting value would be greater than u32,`u32::MAX` is returned instead
+    fn saturating_scale(self, numerator: u32, denominator: u32) -> u32;
 }
 
 pub trait Frequency {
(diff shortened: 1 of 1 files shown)


doc/reference/trd105-time.md
```diff
@@ -0,0 +1,455 @@
+Kernel Time HIL
+========================================
+
+**TRD:** 105<br/>
```
lschuermann, line 4: IIRC usually draft TRDs won't get a TRD number. It might be just fine in this case if we are committed to finalize it once the single set of changes have landed, but I suspect if we follow due process already we might as well get it entirely correct.

doc/reference/trd105-time.md
```diff
@@ -0,0 +1,455 @@
...
+**Working Group:** Kernel<br/>
+**Type:** Documentary<br/>
+**Status:** Draft <br/>
+**Obsoletes:** 101 <br/>
+**Author:** Guillaume Endignoux, Amit Levy, Philip Levis, and Jett Rink <br/>
+**Draft-Created:** 2021/07/32 <br/>
```
lschuermann, line 10: I don't think this month as 32 days. :smile:

lschuermann (commented):
Some formal nitpicking.

doc/reference/trd105-time.md
brghena: I thought we agreed we were fine with a number in this case, since we're already in agreement about it.

doc/reference/trd105-time.md
jettr: Good catch. Fixed. Thanks!

doc/reference/trd105-time.md
jettr: I understood that in this case we were okay with giving it a number since we all had verbal agreement on this, too

doc/reference/trd101-time.md
```diff
@@ -5,6 +5,7 @@ Kernel Time HIL
 **Working Group:** Kernel<br/>
 **Type:** Documentary<br/>
 **Status:** Final <br/>
+**Obsoleted By:** 105 <br/>
```
ppannuto, line 8:
```suggestion
**Status:** Obsolete <br/>
**Obsoleted By:** 105 <br/>
```

I _think_ we're supposed to do this too IIRC

ppannuto (commented):
In a way, this format is a real shame, as it's hard in the github UX to get a nice clean diff from the prior TRD :/

doc/reference/trd101-time.md
jettr: Done. I also agree it is hard to diff without pulling this change down locally and performing a diff.

doc/reference/trd105-time.md
phil-levis: Yeah, so far the process has been that we add a TRD number when it starts getting close to finalization/when it's in a complete state. I think it's OK to assign 105 here. The thing we want to avoid is holes in the TRD numbering.

doc/reference/trd105-time.md
```diff
@@ -0,0 +1,455 @@
...
+at an absolute moment while `Timer` requests a callback at a point
+relative to now.
+
+```rust
+pub trait AlarmClient {
+  fn alarm(&self);
+}
+
+pub trait Alarm: Time {
+  fn set_alarm(&self, reference: Self::Ticks, dt: Self::Ticks);
```
phil-levis, line 249: If we want to add a `max_dt` method or specify how `Alarm` and/or `Timer` respond to `dt` values that are too large, this would be the time!

doc/reference/trd105-time.md
```diff
@@ -0,0 +1,455 @@
...
+
+  if expires.wrapping_sub(now) < MIN_DELAY:
+    expires = now.wrapping_add(MIN_DELAY)
+
+  clear_alarm()
+  set_compare(expires)
+  enable_alarm()
+```
+
+10 Acknowledgements
```
phil-levis, line 429: I'd suggest that we add a short section that describes the changes from 101.

phil-levis (commented):
I think everything here is good, just the question is whether we want to solve the `dt` problem too.

phil-levis:
I think that if we're going to obsolete 101, we might as well fix the `dt` problem too. I see four possible solutions:
1. Add text in the TRD saying that you SHOULD NOT call with a `dt > Ticks::max_value() / 2` and leave the logic unchanged.
2. Modify the implementations to cap `dt` to be `Ticks::max_value() - K` for some constant `K`.
3. Handle it internally by keeping extra state and turning a long `dt` into two separate alarms/timers internally.
4. Use approach 2, adding a `max_dt` method to the `Alarm` and `Timer` traits that tells you what `K` is.

2 and 3 are suggested by @jettr in #2676. 

I am most wary of 2 and 4; there are of course cases when an Alarm doesn't fire exactly when requested, but this is usually due to Alarm interference, and not a deterministic behavior. If we do go with 2 and 4, then setting an alarm/timer should return the actual `dt` value used. 

1 basically pushes 3 into the caller; this has the advantage that an implementation doesn't required the extra state and logic for clients that don't need it, but has the disadvantage that it could lead to replicated caller code and potential bugs in the caller. 

From what I can tell, FreeRTOS doesn't handle this problem, and RIOT handles it by using relative (rather than absolute) timers that are ordered in a list (ztimer) or relying on 64 bits (xtimer), such that this practically is not an issue.

I'm personally in favor of either 1 or 3.

doc/reference/trd105-time.md
jettr: Added.

doc/reference/trd105-time.md
jettr: I think of the dt at more of an issue with the virtual alarm implementation. Without the virtual alarm layer in between, then the actually hardware driver has no problems with a max dt call. I lean toward leaving the TRD unchanged and making the virtual alarm handle the case properly.

jettr:
> I think that if we're going to obsolete 101, we might as well fix the `dt` problem too. I see four possible solutions:
> 
> 1. Add text in the TRD saying that you SHOULD NOT call with a `dt > Ticks::max_value() / 2` and leave the logic unchanged.
> 2. Modify the implementations to cap `dt` to be `Ticks::max_value() - K` for some constant `K`.
> 3. Handle it internally by keeping extra state and turning a long `dt` into two separate alarms/timers internally.
> 4. Use approach 2, adding a `max_dt` method to the `Alarm` and `Timer` traits that tells you what `K` is.
> 
> 2 and 3 are suggested by @jettr in #2676.
> 
> I am most wary of 2 and 4; there are of course cases when an Alarm doesn't fire exactly when requested, but this is usually due to Alarm interference, and not a deterministic behavior. If we do go with 2 and 4, then setting an alarm/timer should return the actual `dt` value used.
> 
> 1 basically pushes 3 into the caller; this has the advantage that an implementation doesn't required the extra state and logic for clients that don't need it, but has the disadvantage that it could lead to replicated caller code and potential bugs in the caller.
> 
> From what I can tell, FreeRTOS and TinyOS don't handle this problem, and RIOT handles it by using relative (rather than absolute) timers that are ordered in a list (ztimer), which has its own issues, or relying on 64 bits (xtimer), such that this practically is not an issue.
> 
> I'm personally in favor of either 1 or 3.

I am also very weary of option 2 and 4. I prefer option 3 the most. I see this as an issue with the virtual alarm implementation instead of an issue with the core definition of how time operations should be defined. There is nothing technically stopping us from waiting max dt. Also without the virtual alarm mux, the actual hardware driver implementation of `Alarm` has no trouble handling max dt values. Will comment further on #2676

phil-levis (approved):

hudson-ayers (approved):

PR #2692, opened 2021-07-25
trd/104-syscalls: clarify guarantees w.r.t. returned Allow buffers

lschuermann:
### Pull Request Overview

This clarifies what guarantees the kernel provides w.r.t. the returned buffer information as part of an Allow system call. See `doc/wg/core/notes/core-notes-2021-07-09.md` and tock/tock#2674 for an extended discussion of these requirements.

These changes are necessary, because the kernel does not enforce that drivers always return a zero-length buffer on the first call to Allow, and return the previously allowed buffer on subsequent calls to Allow.

The wording to use in the TRD104 Syscalls document is tricky, as there are two perspectives on this issue, which must be clearly separated:

- *the capsule implementation*: a capsule MUST return the correct buffer to be correct and compliant to the specification.
- *the system call / kernel behavior*: applications MUST NOT rely on previously allowed buffers to be returned, as least w.r.t. to the current kernel implementation.

Unfortunately, TRD104 prior to these changes expresses: "Calling a Read-Write Allow system call returns a buffer (address and length). On the first call to a Read-Write Allow system call, the kernel returns a zero-length buffer. Subsequent successful calls to Read-Write Allow return the previous buffer passed."

This sets expectations for the userspace / kernel ABI, which are not validated or enforced as part of the current kernel specification. Thus, this commit changes the wording of TRD104 to outline these rules for drivers, but also clarify that these are not enforced as of now, but may be enforced in the future. Userspace applications must expect that drivers are not behaving as specified and should employ proper error handling.

This PR further removes trailing whitespace from the file and contains minor fixes to the file's formatting. I recommend to review commit 944c0e928d634c separately

### Testing Strategy

N/A

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.


jrvanwhy (dismissed):

alistair23 (dismissed):

doc/reference/trd104-syscalls.md
```diff
@@ -509,19 +509,36 @@ kernel returns `Failure` with an error code of `NODEVICE`.
...
-write a buffer that has been allowed: access to the memory is intended
-to be exclusive either to userspace or to the kernel. To regain access
-to a passed buffer, the process calls the same Read-Write Allow system
-call again, which, if successful, returns the buffer. It can do so
-with a zero-length buffer if it wishes to pass no memory to the
-kernel. 
+length). Kernel drivers MUST be implemented such that the returned
+buffer follows a defined set of rules, as specified in the following:
+
+- On the first call to a Read-Write Allow system call, the kernel
```
hudson-ayers, line 518:
```suggestion
- On the first successful call to a Read-Write Allow system call, the kernel
```

doc/reference/trd104-syscalls.md
```diff
@@ -509,19 +509,36 @@ kernel returns `Failure` with an error code of `NODEVICE`.
...
+  returns a zero-length buffer. The returned address value is not
+  specified.
+- Subsequent successful calls to Read-Write Allow return the previous
+  buffer passed, with unmodified address and length values.
+
+However these rules are not enforced by the kernel and thus userspace
```
hudson-ayers, line 524:
```suggestion
However these rules are not enforced by the core kernel and thus userspace
```

doc/reference/trd104-syscalls.md
```diff
@@ -509,19 +509,36 @@ kernel returns `Failure` with an error code of `NODEVICE`.
...
+MUST NOT rely on the fact that drivers follow these rules as
+described. Userspace SHOULD employ additional checks to ensure that
+drivers behave as specified, and implement proper error handling in
+any other case. The kernel MAY enforce a subset of these specified
```
hudson-ayers, line 528:
```suggestion
any other case. The core kernel MAY enforce a subset of these specified
```

hudson-ayers (dismissed):

doc/reference/trd104-syscalls.md
```diff
@@ -509,19 +509,36 @@ kernel returns `Failure` with an error code of `NODEVICE`.
...
-write a buffer that has been allowed: access to the memory is intended
-to be exclusive either to userspace or to the kernel. To regain access
-to a passed buffer, the process calls the same Read-Write Allow system
-call again, which, if successful, returns the buffer. It can do so
-with a zero-length buffer if it wishes to pass no memory to the
-kernel. 
+length). Kernel drivers MUST be implemented such that the returned
+buffer follows a defined set of rules, as specified in the following:
+
+- On the first successful call to a Read-Write Allow system call, the
```
phil-levis, line 518: This seems a bit redundant with lines 554-561 below, but also not completely in sync. E.g., 561 says the address MUST be 0.

Perhaps it would be better to distinguish system call implementations from the kernel. E.g., system call implementations MUST do this but the kernel doesn't check it, so userspace libraries SHOULD? This text says the kernel MUST follow the rules but that it doesn't check them, which seems a bit weird.

phil-levis (changes requested):

phil-levis (dismissed):
I'm good with the text now, but since I rewrote it I'd appreciate other people looking at it to make sure it's exactly what we want.

doc/reference/trd104-syscalls.md
```diff
@@ -571,6 +560,28 @@ userspace needs to read or write to a buffer held by the kernel, it
...
+result, it MUST return the same address and length as those that were passed 
+in the call. When a driver implementing the Read-Write Allow system call
+returns a success result, the returned address and length MUST be those
+that were passed in the previous call, unless this is the first call.
+On the first successful invocation of a particular Read-Write Allow system 
+call, an driver implementation MUST return address 0 and size 0.
+
+Note that these requirements are typically on capsule code, and are not
+enforced by the core kernel. Because capsules may have bugs or not be fully
+trust, userspace SHOULD NOT rely on the fact that drivers follow these rules as
```
hudson-ayers, line 576:
```suggestion
trusted, userspace SHOULD NOT rely on the fact that drivers follow these rules as
```

doc/reference/trd104-syscalls.md
```diff
@@ -571,6 +560,28 @@ userspace needs to read or write to a buffer held by the kernel, it
...
+described, especially if the capsule code is not independently checked and tested.
+Instead, userspace SHOULD employ additional checks to ensure that
+drivers behave as specified, and implement proper error handling as needed.
+The core kernel MAY enforce a subset of these
+specified rules in the future.
+
+The buffer number specifies which buffer this is. A driver may
+support multiple allowed buffers.
```
hudson-ayers, line 584: I believe these lines should still be further up, closer to where the arguments to subscribe are listed.

doc/reference/trd104-syscalls.md
```diff
@@ -509,19 +509,19 @@ kernel returns `Failure` with an error code of `NODEVICE`.
...
+not read or write a buffer that has been allowed: access to the memory
+is intended to be exclusive either to userspace or to the kernel. To
+regain access to a passed buffer, the process calls the same
+Read-Write Allow system call again, which, if successful, returns the
+buffer. It can do so with a zero-length buffer if it wishes to pass no
+memory to the kernel. Once a buffer has been returned to userspace as
+part of a Read-Write Allow system call, it is guaranteed for the
+kernel to no longer have access to the described memory region, unless
+it is currently shared with the kernel as part of the passed in buffer
+or another Allow mechanism.
```
jrvanwhy, line 524: The way I read this, it is allowed for a process to replace an Allowed buffer with a different buffer which overlaps the previous buffer. Was that intentional?

If we choose to implement an overlap check, it ends up being a little awkward. There are either two ways to implement the check:

1. Ignore the existing buffer with the same driver and allow number when checking for overlaps.
2. Remove the existing buffer from the overlap check data structure, check for overlap with the new buffer, and if the new buffer overlaps (i.e. the allow fails) re-add the existing buffer to the data structure.

Not that that's unreasonable, but it would add a bit of code size.

lschuermann: Yes, that was intentional. I think that #2632 we committed to not introducing overlap checks?

doc/reference/trd104-syscalls.md
lschuermann: Yeah, this confused me.

doc/reference/trd104-syscalls.md
jrvanwhy: I thought we were allowing overlap checks to be added in the future. If we don't allow overlap checks, then this makes more sense.

doc/reference/trd104-syscalls.md
```diff
@@ -509,19 +509,19 @@ kernel returns `Failure` with an error code of `NODEVICE`.
...
-with a zero-length buffer if it wishes to pass no memory to the
-kernel. 
+shares a buffer with the kernel that the kernel can read and write.
+
+The standard access model for allowed buffers is that userspace does
+not read or write a buffer that has been allowed: access to the memory
+is intended to be exclusive either to userspace or to the kernel. To
+regain access to a passed buffer, the process calls the same
+Read-Write Allow system call again, which, if successful, returns the
+buffer. It can do so with a zero-length buffer if it wishes to pass no
```
lschuermann, line 519: I've deliberately structured the text to say that the kernel does not enforce the stated requirements way at the beginning. This was driven by the fact that I'm worried that introducing these buffer-swapping semantics here might lead people to think that this is indeed always a one-for-one swap in the Allow slot, which it isn't.

It's entirely fine to state this below, I agree it makes more sense, and it's the developer's fault for not reading all the way. However, it's tricky, given this section literally says "returns the buffer". Changing it to "returns _a_ buffer" better incorporates what ABI guarantees we give:

```suggestion
Read-Write Allow system call again, which, if successful, returns a
buffer. The process can do so with a zero-length buffer if it wishes to pass no
```

(Also resolves some possible ambiguity of who can pass a zero-length buffer)

doc/reference/trd104-syscalls.md
lschuermann: Well, now I'm not sure any longer myself :smile:. You might have a point. @phil-levis @hudson-ayers was that something we wanted to leave open for the future, or do we commit to just say "a process SHOULD NOT do this, but can technically"?

phil-levis: The conclusion was that we should allow overlapped Allows at the system call level. If userspace wants to not allow then though its libraries, that's OK, but then it's a userspace check. After discussion, the conclusion was that the cycle and code size cost of the checks were not worth it given the API protections for memory safety.

doc/reference/trd104-syscalls.md
phil-levis: I think this isn't right. This text is intended to describe how userspace regains exclusive access, which is my calling `allow` again. By saying "returns a buffer", it suggests that it might not be the same buffer that was passed: "a" means "some  buffer" while "the" means "the same buffer."

doc/reference/trd104-syscalls.md
phil-levis: Yeah I think this was inadvertently cut off. I think the whole section needs a re-org; the points are made in strange orders.

phil-levis:
I've reorganized the RW Allow section -- hopefully it is clearer.

doc/reference/trd104-syscalls.md
phil-levis: I agree with the "process" point. I did a bit of reworking, take a look.

alistair23 (approved):

hudson-ayers (approved):

hudson-ayers:
bors r+

Thanks Leon, Phil, and everyone who spent time getting this text right!

PR #2693, opened 2021-07-26
Add support for asymmetric encryption

alistair23:
### Pull Request Overview

This PR adds a HIL for asymmetric encryption. It also adds a capsule to expose the RSA engine to userspace and starts to implement RSA key support as well.

### Testing Strategy

CI and build tests

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


kernel/src/hil/asymmetric_encryption.rs
```diff
@@ -0,0 +1,124 @@
...
+/// The cyrpto buffuers
+/// encrypt -> Source, destination and key.
+/// verify -> Message, signature and key.
+/// decrypt -> Source, destination and key.
+/// Sign -> Input, destination and key.
+pub type CryptoBuffers = (
+    &'static mut [u8],
+    &'static mut [u8],
+    &'static mut dyn PubPrivKey<'static>,
+);
```
bradjc, line 50: Why not just use a normal struct with named fields?

bradjc (commented):
Don't we need some way to actually run/test all of this code?

alistair23:
> Don't we need some way to actually run/test all of this code?

Working on it! As usual I wanted to get the HIL started though as it can take a long time to get HILs accepted.

> Why not just use a normal struct with named fields?

Instead of the `CryptoBuffers` type? I can do that

kernel/src/hil/asymmetric_encryption.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// function is complete. If using an existing key this doesn't need to be
+    /// used.
+    fn set_client(&'a self, client: &'a dyn PubPrivKeyClient<'a>);
+
+    /// This generates a new private/public key pair of `length` bits.
+    /// This will call the `generation_complete()` on completion. They keys
+    /// cannot be used and will return `None` until the upcall has been called.
+    ///
+    /// The keys generated by `generate()` will depend on the implementation.
+    fn gnerate(&'a self, trng: &'a dyn entropy::Entropy32, length: usize) -> Result<(), ErrorCode>;
```
phil-levis, line 26: generate

kernel/src/hil/asymmetric_encryption.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    /// Import an existing public key. This will override any existing keys.
+    fn import_public_key(&mut self, public_key: &[u8]) -> Result<(), ErrorCode>;
```
phil-levis, line 29: Should this be typed, rather than a slice? I.e., do you expect to have variable-length keys?

kernel/src/hil/asymmetric_encryption.rs
```diff
@@ -0,0 +1,132 @@
...
+    fn import_private_key(&mut self, private_key: &[u8]) -> Result<(), ErrorCode>;
+
+    /// Return the public key if it exists.
+    fn pub_key(&'a self) -> Option<&'a [u8]>;
+
+    /// Return the private key if it exists.
+    fn priv_key(&'a self) -> Option<&'a [u8]>;
+}
+
+/// The cyrpto buffuers
```
phil-levis, line 41: crypto

phil-levis (changes requested):
I am really skittish about approving this as a general HIL for public key encryption and signatures, for a variety of reasons:
  - There are public key systems which do other things, such as Diffie-Hellman key exchange
  - There are public key systems which only do signatures (e.g., Lamport signatures)
  - The APIs are open to replay attacks, one of the easy pitfalls in public key systems
  - It leaves a lot of potential issues to run-time checks (e.g., strange key lengths)

Public key cryptography is very tricky, and for a secure OS we should be cautious in the design of its APIs. 

How does one encrypt with P256? Does this perform a key exchange? What symmetric system is used.

I might be much more comfortable with an RSA-specific HIL and P256-specific HIL. We could later try to come up with more general public key APIs.

kernel/src/hil/asymmetric_encryption.rs
alistair23: I expect variable length keys between implementations, but not in the same implementation. I should be able to change this to use const generics.

alistair23: Actually no, I do expect variable length keys. For example RSA could have 2048 or 4096 bit keys

alistair23:
I'm happy to convert this over to be RSA specified and then work on a P256 specific one later as well. It would basically just be a rename of the current version. Let me know if that's where we want to start.

alistair23:
Ok, this is now RSA specific

alistair23:
Any more comments?

kernel/src/hil/asymmetric_encryption.rs
phil-levis: I'd argue those are not variable-length keys; those are two specific lengths which can be expected. I.e., you don't have 1054-bit keys, or 4000-bit keys. Should there be stronger typing here?

kernel/src/hil/asymmetric_encryption.rs
```diff
@@ -0,0 +1,133 @@
...
+    ///
+    /// The keys generated by `generate()` will depend on the implementation.
+    fn generate(&'a self, trng: &'a dyn entropy::Entropy32, length: usize)
+        -> Result<(), ErrorCode>;
+
+    /// Import an existing public key. This will override any existing keys.
+    fn import_public_key(&mut self, public_key: &[u8]) -> Result<(), ErrorCode>;
+
+    /// Import an existing private key. This will override any existing keys.
+    fn import_private_key(&mut self, private_key: &[u8]) -> Result<(), ErrorCode>;
```
phil-levis, line 33: What are the valid ErrorCodes and what do they mean?

kernel/src/hil/asymmetric_encryption.rs
phil-levis: What are the valid ErrorCodes and what do they mean?

kernel/src/hil/asymmetric_encryption.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// function is complete. If using an existing key this doesn't need to be
+    /// used.
+    fn set_client(&'a self, client: &'a dyn PubPrivKeyClient<'a>);
+
+    /// This generates a new private/public key pair of `length` bits.
+    /// This will call the `generation_complete()` on completion. They keys
+    /// cannot be used and will return `None` until the upcall has been called.
+    ///
+    /// The keys generated by `generate()` will depend on the implementation.
+    fn generate(&'a self, trng: &'a dyn entropy::Entropy32, length: usize)
```
phil-levis, line 26: What are the valid ErrorCodes and what do they mean?

kernel/src/hil/asymmetric_encryption.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    /// Store the private key in buffer if it exists.
+    fn priv_key(&'a self, buffer: &mut [u8]) -> Option<()>;
+}
+
+pub enum Operation {
+    Encrypt,
+    Decrypt,
+    Sign,
+    Verify,
```
phil-levis, line 46: It seems weird that signatures and signature verification are in a file about `asymmetric_encryption`. Perhaps this should be called `public_key_crypto.rs`? That would then mean we could expect crypto that's just signatures in here as well. Otherwise, it might make sense separate these out into a separate `signatures.rs`.

kernel/src/hil/asymmetric_encryption.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/// Upcall from the `RsaCrypto` trait.
+pub trait Client<'a, B> {
+    /// The `encrypt()`/`decrypt()`/`sign()` command has been completed.
+    ///
+    /// On success the value of `result` should be ignored unless the `op`
+    /// was a `Operation::Verify` in which cast this indicates if the message
+    /// is valid.
+    fn operation_done(&'a self, result: Result<bool, ErrorCode>, op: Operation, buffers: B);
```
phil-levis, line 56: I'd suggest splitting encrypt/decrypt/sign/verify apart. With `dyn` references, you are going to get everything included even if you only use a small part of it. It might make sense to have encrypt/decrypt in one trait and sign/verify in another.

You want separate callbacks for the different operations. Having a shared structure for buffers, whose fields have different meanings based on the operation, is not taking advantage of type safety. This looks very C-like. 

Why not have an RsaClient? The major advantage of this approach would be that you could write generic modules using client that are parameterized by B. But since this generic isn't in the opposite direction (you have an RSA trait), I'm unsure of when this would be useful.

kernel/src/hil/asymmetric_encryption.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// Set the `Client` client to be called on completion.
+    fn set_client(&'a self, client: &'a dyn Client<'a, RsaCryptoBuffers>);
+
+    /// Encrypt the data in the `source` destination with the public `key`
+    /// specified and store the data in the `dest` buffer.
+    ///
+    /// On completion the `operation_done()` upcall will be scheduled.
+    /// The type of encryption that occurs will be based on the latest
+    /// `set_mode*()` function that was called.
+    fn encrypt(&self, buffers: RsaCryptoBuffers) -> Result<(), (ErrorCode, RsaCryptoBuffers)>;
```
phil-levis, line 99: Since you are moving `RsaCryptoBuffers`,  why not just have parameters?

phil-levis (changes requested):
Overall I think this is looking good. Making it RSA-specific definitely makes things simpler, but the traits need some changes. High-level points
  - The asymmetry between separate methods in the `RsaCrypto` trait and the `Client` trait having a single callback is problematic. The client can't be used for systems that only do encryption/decryption or only do signing/verification. 
  - Break encrypt/decrypt and sign/verify into at least 2 separate traits. Maybe even 4, with  composite traits that combine them. HIL design point #8. 
  - Document valid `ErrorCode` values and their causes

There are also some questions (e.g., about `RsaCryptoBuffers`).

alevy:
@alistair23 I know this has gone through two iterations already, but can you point to the hardware that would implement this? I.e. if it's OpenTitan than the relevant specifications for the RSA module?

alistair23:
> I'd argue those are not variable-length keys; those are two specific lengths which can be expected. I.e., you don't have 1054-bit keys, or 4000-bit keys. Should there be stronger typing here?

Yeah... I don't think it's that clear.

I agree that keys will generally be specific lengths. But for example ECC can have compressed and uncompressed versions, so I don't want to specify the exact length that the buffers must be.

kernel/src/hil/asymmetric_encryption.rs
alistair23: I have changed `Client` to be `RsaClient`.

I don't agree though that splitting the complete upcalls apart is better. If you look at the [capsule upcall code](https://github.com/tock/tock/pull/2693/files#diff-6908c29af6e08ab6f8f3882b89317a78fcff3193b01ef0a033aa98bae1074956R200) we can cleanly pass the complete command up to userspace. Splitting these in the kernel would just make that clunky.

We could debate splitting the kernel and userspace upcalls, but I'm not convinced that gives us any benefits. Consolidating all of the callbacks into a single one to userspace I think is simpler.

kernel/src/hil/asymmetric_encryption.rs
alistair23: Looking at the capsule code using this I think the `RsaCryptoBuffers` is easier to understand. We pass the buffers in and get them back for every command

alistair23:
> @alistair23 I know this has gone through two iterations already, but can you point to the hardware that would implement this? I.e. if it's OpenTitan than the relevant specifications for the RSA module?

For OT the [OTBN](https://docs.opentitan.org/hw/ip/otbn/doc/index.html) will implement RSA so the next step is to connect that up

kernel/src/hil/asymmetric_encryption.rs
phil-levis: I think it's important to distinguish a kernel HIL and a userspace API. There are going to be users of this HIL within the kernel that are not a system call capsule. We want to design the HIL for internal kernel use, where *one* of the uses is a system call capsule. A HIL is intended to be general, not for a specific use case.

phil-levis:
There was a long discussion about this PR on the core call on Friday. @jettr has some crypto experts on his team, he's going to ask them for feedback and input on what public key APIs should look like. Once the notes are up I'll link them here.

kernel/src/hil/asymmetric_encryption.rs
alistair23: I agree. I don't see why splitting it up will be easier be any better for internal kernel use either. The current way will work internally and works very well with the only current user which is the system call capsule.

alistair23:
I would like to be included in discussions about my PRs, instead of just being told about the decisions. Maybe in the future we could discuss it in the OpenTitan call instead?

phil-levis:
I just pinged the list to get the notes up soon. There are a lot of people interested in/concerned with crypto APIs in the core group, so there was a lot of discussion. It's a HIL, so under the purview of the core WG, so it makes sense for it to be discussed on the core call. Once the notes are up I can try to summarize the points that were discussed in a message here, and mention their authors, to continue the discussion here.

alevy:
Just to ease @alistair23's mind re:

> instead of just being told about the decisions.

There was no decision on the call, nor should there be, we just used the time force some of us to take a close look at the PR.

alevy:
I've looked into this a bit over the last week (sorry, my time is extremely limited atm as I'm *technically* on leave). Thoughts below:

The main question that concerns me (with *all* new HILs, not just this one) is whether it reflects both a generic interface across a set of reasonable hardware devices *and* is minimally abstract such that each implementation has minimal logic in code. The time HIL is a good example (an easy one, because alarm/timer hardware usually looks very very similar): the HIL just presents a unified interface for configuring an alarm interrupt, getting the current "tick", etc, and all the logic of actually setting future alarms based on a relative time, e.g., are in capsules _above_ the HIL.

This RSA HIL is a tricky one. OpenTitan doesn't _really_ have RSA hardware, it has a programmable, separate core, with it's own RISC-V-inspired instruction set, that includes bignum operations. Pretty cool! One can (and is presumably expected to) implement RSA in this instruction set so that this separate core is doing the high level encrypt/decrypt operations.

The issue, then, is that really it seems that we are co-designing the HIL with the RSA program on OpenTitan. The HIL can be whatever we want (because whatever complexity exists would be programmed into the bignum core, not the chip driver in Rust). However, it's not at all clear from this process that the resulting HIL would be meaningfully useful on any other hardware...

So where do we go from here? I think we should do two things: we should figure out a way to incorporate an RSA stack for OpenTitan in upstream Tock that is _allowed_ to be hardware specific, _and_ we go figure out what other hardware might be implementing (it generally looks like just bignum accelerators). Then we can either try and unify the HILs or perhaps layer interfaces (a Bignum HIL and an RSA trait in capsules) where OpenTitan implements the higher-level HIL directly while other hardware implements the HIL and a capsule does the adaptation to upper layer interfaces.

phil-levis:
Here are the notes from the call: https://github.com/tock/tock/blob/3e310821e11487ec963650219ca35769a7ea3c6b/doc/wg/core/notes/core-notes-2021-09-03.md

phil-levis:
@jettr @alevy @alistair23 @lschuermann @vsukhoml @phil-levis had a 60m phone call to discuss how to move forward on this PR. There were several conclusions:
  - This is not really a HIL, but rather a useful higher-level interface to RSA. An actual HIL is typically much closer to an RSA operation and hashing, while something like signatures combines those and adds other complexities such as padding.
  - Having a higher-level useful interface to RSA like this is valuable, as it's something closer to what capsules would actually want to use.
  - Combining the previous two points means that we likely want a "high" RSA interface like this one, as well as a "low" RSA interface which is a canonical HIL.
  - To include the high HIL in the main tree would require having an implementation of it. This could be software.
  - Some suggestions/needed changes for the traits in this PR for it to be a ready as upper HIL.

phil-levis:
Some changes I think this needs:
  - Allow for PubKey as well as PubPrivKey so that a client only having a public key doesn't require wasting storage space on the private key. Probably `PubPrivKey: PubKey`.
  - Move `generate` out of PubKey, because it requires all implementations have key generation. Having this in a separate trait would make sense.
  - Transition away from the `RsaCryptoBuffers` to standard parameters. The idea that this structure is effectively a level of indirection for parameters (which one does what depends on the call) seems prone to error. I.e., there isn't an easy way to inspect a single line of code and be sure that the right data are being passed in the right fields.
  - The `PubKey` trait needs to be able to report its length.
  - One question: would it make sense to be able to type-check the size of keys? E.g., if you want an implementation to only use keys of a certain length, can you statically enforce this? Otherwise you can get into a weird situation where the size of data you pass is related to the key, but then you could possibly pass the wrong length key. This pushes padding issues into the implementation, or necessitates more run-time errors.
  - `sign` and `verify` might need to be parameterized by their hashing; I think it would be OK for these traits to only use PKCS#1, although I not very knowledgeable about what the design space here is like.
  - the `Client` trait should be split into the different operations.

vsukhoml:
Taking into account that typical public keys (say in certificates) are big-endian, while internal representation is frequently a little-endian - need to take into account when importing keys, may be clearly say that import/export is from/to big-endian and internal representation.

RSA without padding is just modular exponent ( = input^exp mod N) - 3 parameter operation taking 3 big numbers as input and 1 output buffer). Encrypt/Decrypt/Sign/Verify just differs in what you feed in. Could it be a better name for an operation? This way you don't need notion of private/public keys - it can be on higher level. 

I'm not sure about underlying implementation, but I'd make internal storage machine word aligned [usize]. This would all more efficient implementations.

vsukhoml:
Another thought - implementation of modular exponentiation can be fast but leaky (not constant time, etc), or can be secure, use blinding, constant time but as a result slow. When you verify a signature or decrypt you would prefer a faster option, while for signing/encrypt - more secure. Secure version will also be larger. So, HIL may need to address this somehow.

alistair23:
Ok!

I'm going to close this PR and instead try to split it up into smaller PRs that we can work on merging. Step 1 is this:

Issue #2635, opened 2021-06-28
Tracking: Crypto HIL improvements


bradjc:
- [ ] The `set_mode_X()` HIL functions need to support returning an error if hardware does not support that mode or if the implementation cannot put the hardware in to that mode at that time. See for an example where panic is used instead of returning an error https://github.com/tock/tock/pull/2629/files.

phil-levis:
I think this is a larger issue, which we're facing with virtualization.

In the best case, the set mode methods are checked statically: if an implementation can provide them it does, and if it doesn't it doesn't. That way if a client needs a certain mode, you can't connect it to an implementation that doesn't implement it.

The issue with this is virtualizers. If you say that a virtualizer implements only the traits its underlying resource implements, you end up with lots of different virtualizers: this one supports AES128CBC, that one supports AES128CBC+AES128ECB, etc.

I see 4 ways forward:

1. Forget the "trait per mode" approach, and go fully dynamic, using an enum. If you request a mode not supported, it returns and error.
2. Keep the "trait per mode" approach, for static binding in the kernel, but introduce error return values because we want a single virtualizer implementation (the current approach).
3. Have a variety of virtualizer implementations, supporting different cipher suites.
4. Refactor how these traits work such that the virtualizer is parameterized by a type which implements the `set_mode` methods, such that system calls can get runtime errors but any given `set_mode` doesn't. This is pared down version of 3: we have a single virtualizer implementation, but multiple implementations of what cipher suites are supported.

There are certainly others. But what I want to avoid is pushing us towards run-time checks (motivated by userspace) that be checked statically for in-kernel use.

PR #2697, opened 2021-07-27
HIL: Digest: Add SHA/HMAC verify

alistair23:
### Pull Request Overview

This PR adds to the SHA HIL a `verify()` option. This allows users to check that a result matches what they expect.

This is beneficial as it avoids having to copy the data around when a user just wants to check if the digest matches a known value.

### Testing Strategy

Running HMAC tests on OpenTitan.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/sha.rs
+++ b/capsules/src/sha.rs
@@ -24,6 +24,7 @@
 //! ```
 
 use crate::driver;
+use kernel::errorcode::into_statuscode;
 /// Syscall driver number.
 pub const DRIVER_NUM: usize = driver::NUM::Sha as usize;
 
@@ -168,6 +169,21 @@ impl<
 
         Ok(())
     }
+
+    fn verify_digest(&self) -> Result<(), ErrorCode> {
+        self.data_copied.set(0);
+
+        if let Err(e) = self.sha.verify(self.dest_buffer.take().unwrap()) {
+            // Error, clear the appid and data
+            self.sha.clear_data();
+            self.appid.clear();
+            self.dest_buffer.replace(e.1);
+
+            return Err(e.0);
+        }
+
+        Ok(())
+    }
 }
 
 impl<
@@ -255,10 +271,31 @@ impl<
                     if app.op.get().unwrap() == UserSpaceOp::Run {
                         if let Err(e) = self.calculate_digest() {
                             upcalls
-                                .schedule_upcall(
-                                    0,
-                                    (kernel::errorcode::into_statuscode(e.into()), 0, 0),
-                                )
+                                .schedule_upcall(0, (into_statuscode(e.into()), 0, 0))
+                                .ok();
+                        }
+                    } else if app.op.get().unwrap() == UserSpaceOp::Verify {
+                        let _ = app.compare.enter(|compare| {
+                            let mut static_buffer_len = 0;
+                            self.dest_buffer.map(|buf| {
+                                // Determine the size of the static buffer we have
+                                static_buffer_len = buf.len();
+
+                                if static_buffer_len > compare.len() {
+                                    static_buffer_len = compare.len()
+                                }
+
+                                self.data_copied.set(static_buffer_len);
+
+                                // Copy the data into the static buffer
+                                compare[..static_buffer_len]
+                                    .copy_to_slice(&mut buf[..static_buffer_len]);
+                            });
+                        });
+
+                        if let Err(e) = self.verify_digest() {
+                            upcalls
+                                .schedule_upcall(1, (into_statuscode(e.into()), 0, 0))
                                 .ok();
                         }
                     } else {
@@ -305,14 +342,7 @@ impl<
                     match result {
                         Ok(_) => upcalls.schedule_upcall(0, (0, pointer as usize, 0)).ok(),
                         Err(e) => upcalls
-                            .schedule_upcall(
-                                0,
-                                (
-                                    kernel::errorcode::into_statuscode(e.into()),
-                                    pointer as usize,
-                                    0,
-                                ),
-                            )
+                            .schedule_upcall(0, (into_statuscode(e.into()), pointer as usize, 0))
                             .ok(),
                     };
 
@@ -339,12 +369,32 @@ impl<
         const L: usize,
     > digest::ClientVerify<'a, L> for ShaDriver<'a, H, L>
 {
-    fn verification_done(
-        &'a self,
-        _result: Result<bool, ErrorCode>,
-        _compare: &'static mut [u8; L],
-    ) {
-        unimplemented!();
+    fn verification_done(&'a self, result: Result<bool, ErrorCode>, compare: &'static mut [u8; L]) {
+        self.appid.map(|id| {
+            self.apps
+                .enter(*id, |_app, upcalls| {
+                    self.sha.clear_data();
+
+                    match result {
+                        Ok(equal) => upcalls.schedule_upcall(1, (0, equal as usize, 0)),
+                        Err(e) => upcalls.schedule_upcall(1, (into_statuscode(e.into()), 0, 0)),
+                    }
+                    .ok();
+
+                    // Clear the current appid as it has finished running
+                    self.appid.clear();
+                })
+                .map_err(|err| {
+                    if err == kernel::process::Error::NoSuchApp
+                        || err == kernel::process::Error::InactiveApp
+                    {
+                        self.appid.clear();
+                    }
+                })
+        });
+
+        self.check_queue();
+        self.dest_buffer.replace(compare);
     }
 }
 
@@ -392,6 +442,15 @@ impl<
                 .apps
                 .enter(appid, |app, _| {
                     mem::swap(&mut app.data, &mut slice);
+                    Ok(())
+                })
+                .unwrap_or(Err(ErrorCode::FAIL)),
+
+            // Compare buffer for verify
+            2 => self
+                .apps
+                .enter(appid, |app, _| {
+                    mem::swap(&mut app.compare, &mut slice);
                     Ok(())
                 })
                 .unwrap_or(Err(ErrorCode::FAIL)),
@@ -595,10 +654,86 @@ impl<
                         .enter(appid, |_app, upcalls| {
                             if let Err(e) = self.calculate_digest() {
                                 upcalls
-                                    .schedule_upcall(
-                                        0,
-                                        (kernel::errorcode::into_statuscode(e.into()), 0, 0),
-                                    )
+                                    .schedule_upcall(0, (into_statuscode(e.into()), 0, 0))
+                                    .ok();
+                            }
+                        })
+                        .unwrap();
+                    CommandReturn::success()
+                } else {
+                    // We don't queue this request, the user has to call
+                    // `update` first.
+                    CommandReturn::failure(ErrorCode::OFF)
+                }
+            }
+
+            // veryify
+            // Use key and data to compute hash and comapre it against
+            // the digest
+            4 => {
+                if match_or_empty_or_nonexistant {
+                    self.appid.set(appid);
+                    let _ = self.apps.enter(appid, |app, _| {
+                        app.op.set(Some(UserSpaceOp::Verify));
+                    });
+                    let ret = self.run();
+
+                    if let Err(e) = ret {
+                        self.sha.clear_data();
+                        self.appid.clear();
+                        self.check_queue();
+                        CommandReturn::failure(e)
+                    } else {
+                        CommandReturn::success()
+                    }
+                } else {
+                    // There is an active app, so queue this request (if possible).
+                    self.apps
+                        .enter(appid, |app, _| {
+                            // Some app is using the storage, we must wait.
+                            if app.pending_run_app.is_some() {
+                                // No more room in the queue, nowhere to store this
+                                // request.
+                                CommandReturn::failure(ErrorCode::NOMEM)
+                            } else {
+                                // We can store this, so lets do it.
+                                app.pending_run_app = Some(appid);
+                                app.op.set(Some(UserSpaceOp::Verify));
+                                CommandReturn::success()
+                            }
+                        })
+                        .unwrap_or_else(|err| err.into())
+                }
+            }
+
+            // verify_finish
+            // Use key and data to compute hash and comapre it against
+            // the digest, useful after a update command
+            5 => {
+                if app_match {
+                    self.apps
+                        .enter(appid, |app, upcalls| {
+                            let _ = app.compare.enter(|compare| {
+                                let mut static_buffer_len = 0;
+                                self.dest_buffer.map(|buf| {
+                                    // Determine the size of the static buffer we have
+                                    static_buffer_len = buf.len();
+
+                                    if static_buffer_len > compare.len() {
+                                        static_buffer_len = compare.len()
+                                    }
+
+                                    self.data_copied.set(static_buffer_len);
+
+                                    // Copy the data into the static buffer
+                                    compare[..static_buffer_len]
+                                        .copy_to_slice(&mut buf[..static_buffer_len]);
+                                });
+                            });
+
+                            if let Err(e) = self.verify_digest() {
+                                upcalls
+                                    .schedule_upcall(1, (into_statuscode(e.into()), 0, 0))
                                     .ok();
                             }
                         })
@@ -625,6 +760,7 @@ impl<
 enum UserSpaceOp {
     Run,
     Update,
+    Verify,
 }
 
 #[derive(Default)]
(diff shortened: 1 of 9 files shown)


bradjc (dismissed):

bradjc (approved):

phil-levis (changes requested):
Given that there are now two operations with separate callbacks, I'd suggest breaking apart hash and verify (point 8 in the HIL design TRD). So you have three base traits: data, hash, and verify, and then three combinations data+hash, data+verify, and data+hash+verify. You also have a similar structure in the Client traits, with the `set_client` operations matching the two paths.

alistair23:
The problem with breaking up the traits is that then the capsules need all three. So we still end up in a position where everything needs to implement all 3, just now the source code is more verbose.

Also I don't think it makes sense to split up data and hash. What engine would take data in but not generate a hash?

lschuermann:
Just as a side note, I don't think that the statement that `verify` doesn't require an additional buffer will be true for many cases. One still needs allocate `&'static mut [u8; L]` buffer to copy the to be verified hash into. Arguably, this means that most of the time (for instance from a Driver's perspective) one could just generate a hash and then compare to a value in a buffer provided by some client which isn't `&'static mut` (e.g. process buffer). I'm not saying this isn't useful in all cases, just that the utility might be decreased by the fact that we still need to have a `&'static mut [u8; L]` in the kernel.

phil-levis:
> The problem with breaking up the traits is that then the capsules need all three. So we still end up in a position where everything needs to implement all 3, just now the source code is more verbose.

Not everything needs to implement all 3. You can have something which only hashes, and something which only verifies. 

> Also I don't think it makes sense to split up data and hash. What engine would take data in but not generate a hash?

You split up data and hash so that you can have data+verify without hash.

So something like:

pub trait Data {...}
pub trait HashOnly {...}
pub trait VerifyOnly{...}

pub trait Hash: HashOnly + Data {}
pub trait Verify: VerifyOnly + Data {}
pub trait Digest: HashOnly + VerifyOnly + Data {}

UART is an example of this structure:

alistair23:
As you point out at https://github.com/tock/tock/issues/2635#issuecomment-890585932 with virtulisation we are unfortunately stuck with everything having to implement all implementations. Splitting it up doesn't get us anything, just longer `impl` names in the capsules

alistair23:
Ping!

phil-levis:
> As you point out at [#2635 (comment)](https://github.com/tock/tock/issues/2635#issuecomment-890585932) with virtulisation we are unfortunately stuck with everything having to implement all implementations. Splitting it up doesn't get us anything, just longer `impl` names in the capsules

Please follow the HIL TRD and split the traits.

Right now, a client that only needs to hash also is given access to verification, and a client that only needs to verify is also given access to hashing. Furthermore, an implementer of `Client` has to implement both  callbacks even if it only implements one of them. From the HIL TRD:

> Separating a HIL into fine-grained traits allows Tock to follow the security principle of least privilege. In the case of GPIO, for example, being able to read a pin does not mean a client should be able to reconfigure or write it. Similarly, for a UART, being able to transmit data does not mean that a client should always also be able to read data, or reconfigure the UART parameters.

phil-levis (changes requested):
Please split the traits.

alistair23:
Done

alistair23:
Ping!

kernel/src/hil/digest.rs
```diff
@@ -6,23 +6,50 @@ use crate::ErrorCode;
...
+    /// On success the result indicate if the hashes match or don't.
+    /// On failure the result will indicate an `ErrorCode`.
+    fn verification_done(&'a self, result: Result<bool, ErrorCode>, compare: &'static mut [u8; L]);
+}
+
+pub trait Client<'a, const L: usize>:
+    ClientData<'a, L> + ClientHash<'a, L> + ClientVerify<'a, L>
+{
+}
+
```
phil-levis, line 43: It would also make sense have standard composite traits for Data+Verify and Data+Hash? `ClientDataHash` and `ClientDataVerify`. Those seem like the two most common use cases for capsules.

kernel/src/hil/digest.rs
```diff
@@ -56,11 +93,47 @@ pub trait Digest<'a, const L: usize> {
...
+    /// Compare the specified digest in the `compare` buffer to the calculated
+    /// digest. This function is similar to `run()` and should be used instead
+    /// of `run()` if the caller doesn't need to know the output, just if it
+    /// matches a known value.
+    ///
+    /// For example:
+    /// ```rust,ignore
+    ///     // Compute a digest on data
+    ///     add_data(...);
+    ///     add_data(...);
```
phil-levis, line 111: Can you call `add_data` twice synchronously?

kernel/src/hil/digest.rs
```diff
@@ -6,23 +6,50 @@ use crate::ErrorCode;
...
+}
+
 /// Computes a digest (cryptographic hash) over data
 ///
 /// 'L' is the length of the 'u8' array to store the digest output.
-pub trait Digest<'a, const L: usize> {
+pub trait DigestData<'a, const L: usize> {
     /// Set the client instance which will receive `hash_done()` and
     /// `add_data_done()` callbacks.
     /// This callback is called when the data has been added to the digest
```
phil-levis, line 55: Should this change to receiving a `ClientData`?

kernel/src/hil/digest.rs
```diff
@@ -41,6 +68,16 @@ pub trait Digest<'a, const L: usize> {
...
+    /// This won't clear the buffers provided to this API, that is up to the
+    /// user to clear.
+    fn clear_data(&self);
+}
+
+/// Computes a digest (cryptographic hash) over data
+///
+/// 'L' is the length of the 'u8' array to store the digest output.
+pub trait DigestHash<'a, const L: usize> {
     /// Request the hardware block to generate a Digest and stores the returned
```
phil-levis, line 81: Shouldn't this have a `set_client`?

Note that for `UART`, to make things a little simpler, it disambiguates the different clients through method names. So `set_digest_client`.

kernel/src/hil/digest.rs
```diff
@@ -56,11 +93,47 @@ pub trait Digest<'a, const L: usize> {
...
 
-    /// Clear the keys and any other sensitive data.
-    /// This won't clear the buffers provided to this API, that is up to the
-    /// user to clear.
-    fn clear_data(&self);
+/// Computes a digest (cryptographic hash) over data
+///
+/// 'L' is the length of the 'u8' array to store the digest output.
+pub trait DigestVerify<'a, const L: usize> {
+    /// Compare the specified digest in the `compare` buffer to the calculated
```
phil-levis, line 102: Does this need a `set_client`?

kernel/src/hil/digest.rs
```diff
@@ -56,11 +93,47 @@ pub trait Digest<'a, const L: usize> {
...
+        compare: &'static mut [u8; L],
+    ) -> Result<(), (ErrorCode, &'static mut [u8; L])>;
+}
+
+/// Computes a digest (cryptographic hash) over data
+///
+/// 'L' is the length of the 'u8' array to store the digest output.
+pub trait Digest<'a, const L: usize>:
+    DigestData<'a, L> + DigestHash<'a, L> + DigestVerify<'a, L>
+{
```
phil-levis, line 131: It might make sense to add a `set_client` here, which sets all of the clients.

kernel/src/hil/digest.rs
alistair23: No, you have to wait for the callback. This is just pseudo code, I'll add a comment

kernel/src/hil/digest.rs
```diff
@@ -56,13 +108,70 @@ pub trait Digest<'a, const L: usize> {
...
-    /// This won't clear the buffers provided to this API, that is up to the
-    /// user to clear.
-    fn clear_data(&self);
+/// Computes a digest (cryptographic hash) over data
+///
+/// 'L' is the length of the 'u8' array to store the digest output.
+pub trait DigestVerify<'a, const L: usize> {
+    /// Set the client instance which will receive the `verification_done()`
+    /// callback.
+    /// This is not required is using the `set_client()` fuction from the
```
phil-levis, line 119: is using/if using

kernel/src/hil/digest.rs
```diff
@@ -56,13 +108,70 @@ pub trait Digest<'a, const L: usize> {
...
+    /// For example:
+    /// ```rust,ignore
+    ///     // Compute a digest on data
+    ///     add_data(...);
+    ///     add_data(...);
+    ///
+    ///     // Compare the computed digest generated to an existing digest
+    ///     verify(...);
+    /// ```
+    /// NOTE: The above is just pseudo code. The user is expected to check for
```
phil-levis, line 138: If it's just pseudocode, perhaps make it not rust formatted?

kernel/src/hil/digest.rs
```diff
@@ -56,13 +108,70 @@ pub trait Digest<'a, const L: usize> {
...
+    /// `add_data_done()` and `verification_done()` callbacks.
+    fn set_client(&'a self, client: &'a dyn Client<'a, L>);
+}
+
+/// Computes a digest (cryptographic hash) over data
+///
+/// 'L' is the length of the 'u8' array to store the digest output.
+pub trait DigestDataHash<'a, const L: usize>: DigestData<'a, L> + DigestHash<'a, L> {}
+
+impl<'a, T: DigestData<'a, L> + DigestHash<'a, L>, const L: usize> DigestDataHash<'a, L> for T {}
```
phil-levis, line 166: Can a default implementation of `set_client` call the three underlying client set methods?

kernel/src/hil/digest.rs
```diff
@@ -41,6 +76,23 @@ pub trait Digest<'a, const L: usize> {
         data: LeasableBuffer<'static, u8>,
     ) -> Result<usize, (ErrorCode, &'static mut [u8])>;
 
+    /// Clear the keys and any other sensitive data.
```
phil-levis, line 79: Where are the keys set? For a digest there shouldn't be keys?

kernel/src/hil/digest.rs
```diff
@@ -56,13 +108,70 @@ pub trait Digest<'a, const L: usize> {
     /// error with error code ENOSUPPORT.
     fn run(&'a self, digest: &'static mut [u8; L])
         -> Result<(), (ErrorCode, &'static mut [u8; L])>;
+}
 
-    /// Clear the keys and any other sensitive data.
-    /// This won't clear the buffers provided to this API, that is up to the
-    /// user to clear.
-    fn clear_data(&self);
+/// Computes a digest (cryptographic hash) over data
```
phil-levis, line 113: Over which data? Perhaps "computes a digest (cryptographic hash) over data provided through a separate trait". That way it could be DigestData or some other trait (for other use cases).

kernel/src/hil/digest.rs
```diff
@@ -56,13 +108,70 @@ pub trait Digest<'a, const L: usize> {
...
+    /// The verify function is useful to compare input with a known digest
+    /// value. The verify function also saves callers from allocating a buffer
+    /// for the digest when they just need verification.
+    fn verify(
+        &'a self,
+        compare: &'static mut [u8; L],
+    ) -> Result<(), (ErrorCode, &'static mut [u8; L])>;
 }
 
+/// Computes a digest (cryptographic hash) over data
```
phil-levis, line 150: Seems like a copied comment.

kernel/src/hil/digest.rs
```diff
@@ -56,13 +108,70 @@ pub trait Digest<'a, const L: usize> {
...
+/// 'L' is the length of the 'u8' array to store the digest output.
+pub trait Digest<'a, const L: usize>:
+    DigestData<'a, L> + DigestHash<'a, L> + DigestVerify<'a, L>
+{
+    /// Set the client instance which will receive `hash_done()`,
+    /// `add_data_done()` and `verification_done()` callbacks.
+    fn set_client(&'a self, client: &'a dyn Client<'a, L>);
+}
+
+/// Computes a digest (cryptographic hash) over data
```
phil-levis, line 161: Seems like a copied comment.

kernel/src/hil/digest.rs
```diff
@@ -56,13 +108,70 @@ pub trait Digest<'a, const L: usize> {
...
+///
+/// 'L' is the length of the 'u8' array to store the digest output.
+pub trait DigestDataHash<'a, const L: usize>: DigestData<'a, L> + DigestHash<'a, L> {}
+
+impl<'a, T: DigestData<'a, L> + DigestHash<'a, L>, const L: usize> DigestDataHash<'a, L> for T {}
+
+/// Computes a digest (cryptographic hash) over data
```
phil-levis, line 168: Copied comment.

phil-levis (commented):
Overall, looking good. Just needs some cleanup in the comments. I'll ask to put it on the core call agenda tomorrow. I think it should be smooth sailing; assuming all goes well and the comments filled out/fixed 'll try to approve tomorrow.

kernel/src/hil/digest.rs
alistair23: Yeah, I'll remove that then. I like the syntax highlighting though :)

kernel/src/hil/digest.rs
alistair23: Yes

kernel/src/hil/digest.rs
alistair23: The keys are set in the `set_mode*()` functions

kernel/src/hil/digest.rs
alistair23: It is a copied comment, is that bad though? It still describes what is happening

lschuermann:
@alistair23 what do you think of my comment regarding the verify method? https://github.com/tock/tock/pull/2697#issuecomment-889968758

We did shortly discuss this on the call today and I think my comment is still valid. One significant advantage of not having a `DigestVerify` would be that it significantly reduces complexity as we only have one "frontend" for the HIL then. We wouldn't need to worry about parallel uses of the `Digest` and `Verify` parts of the interfaces for instances which implement both. Also, it doesn't occur to me that there could be any instances of hardware which don't implement `Verify` but `Digest` or the other way around. If we were to elude having `verify`, we might also just have a single clients, given there is no reason to separate `add_data_done` and `hash_done` for a single "HIL frontend" any longer.

phil-levis:
> @alistair23 what do you think of my comment regarding the verify method? [#2697 (comment)](https://github.com/tock/tock/pull/2697#issuecomment-889968758)
> 
> We did shortly discuss this on the call today and I think my comment is still valid. One significant advantage of not having a `DigestVerify` would be that it significantly reduces complexity as we only have one "frontend" for the HIL then. We wouldn't need to worry about parallel uses of the `Digest` and `Verify` parts of the interfaces for instances which implement both. Also, it doesn't occur to me that there could be any instances of hardware which don't implement `Verify` but `Digest` or the other way around. If we were to elude having `verify`, we might also just have a single clients, given there is no reason to separate `add_data_done` and `hash_done` for a single "HIL frontend" any longer.

This goes against the recommendations in the HIL TRD. The idea is that an implementation has only one "frontend", which is `Digest`, because it provides all of the traits. However, we can constrain a client to a particular subset of the functionality if we want to. E.g., we can have a client that can verify hashes but can't compute them. The idea of one client using `DigestHash` and another `DigestVerify` is similar to one client using `UartData` and another `UartAdvanced`.

Put another way, the purpose of fine-grained traits here isn't to distinguish among hardware implementations/functionality, but instead to limit a particular client to only the functionality it needs (least privilege).

kernel/src/hil/digest.rs
phil-levis: Not really -- it combines the data and verify traits so a client can verify a hash, not compute it.

phil-levis (dismissed):

phil-levis:
My approval is on the Digest HIL, which looks good to me. I think how it's handling length (L) and the clients is clean and flexible.

phil-levis:
Overall, this looks good. I think we'll need to polish the HIL a little bit before finalizing it (particularly, how you set keys in different modes and their error cases), but it's good enough for now.

alistair23:
Marking as last call

alistair23:
Ready to merge!

capsules/src/sha.rs
```diff
@@ -573,10 +654,86 @@ impl<
...
+                        .unwrap();
+                    CommandReturn::success()
+                } else {
+                    // We don't queue this request, the user has to call
+                    // `update` first.
+                    CommandReturn::failure(ErrorCode::OFF)
+                }
+            }
+
+            // veryify
```
ppannuto, line 670:
```suggestion
            // verify
```

ppannuto (approved):
bors r+

PR #2698, opened 2021-07-27
Updated lps25hb to handle i2c errors

Cristiana959:
### Pull Request Overview

This pull request adds i2c error verification for the lps25hb drivers.

### Testing Strategy

This pull request was not tested as I do not own the sensor.

### TODO or Help Wanted

N/A

### Documentation Updated

- [x]  No updates required.

### Formatting

- [x] Ran `make prepush`.

PR #2701, opened 2021-07-27
Separate schedule_upcall number and values through Rust types

alexandruradovici:
### Pull Request Overview

This pull request groups the values returned by `schedule_upcall` into a tuple to prevent wrongly placing a value instead of the upcall number.

```rust
pub fn schedule_upcall(
        &self,
        subscribe_num: usize,
        r: (usize, usize, usize),
    ) -> Result<(), UpcallError> 
```

This tries to use Rust's type system to clearly separate the upcall number and the values. The `schedule_upcall` takes four `usize` parameters, the first one being the upcall number and the following three the values returned to the application. As all four parameters have the same type, it is easy to forget that the first parameter is the upcall number, not the first value. The compiler has no way of protecting against this error. 

This is inspired from the error within #2698.

There should be no impact in size or performance.

### Testing Strategy

N/A

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/Porting_v1_Capsules_to_v2.md
+++ b/doc/Porting_v1_Capsules_to_v2.md
@@ -311,7 +311,7 @@ replaced with code like:
 
 ```rust
 self.apps.enter(appid, |app, upcalls| {
-    upcalls.schedule_upcall(upcall_number,(r0, r1, r2));
+    upcalls.schedule_upcall(upcall_number, (r0, r1, r2));
 });
 ```
 
--- a/doc/Porting_v1_Capsules_to_v2.md
+++ b/doc/Porting_v1_Capsules_to_v2.md
@@ -311,7 +311,7 @@ replaced with code like:
 
 ```rust
 self.apps.enter(appid, |app, upcalls| {
-    upcalls.schedule_upcall(upcall_number, r0, r1, r2);
+    upcalls.schedule_upcall(upcall_number,(r0, r1, r2));
 });
 ```
 
--- a/capsules/src/adc.rs
+++ b/capsules/src/adc.rs
@@ -708,9 +708,11 @@ impl<A: hil::adc::Adc + hil::adc::AdcHighSpeed> hil::adc::Client for AdcDedicate
                         upcalls
                             .schedule_upcall(
                                 0,
-                                AdcMode::SingleSample as usize,
-                                self.channel.get(),
-                                sample as usize,
+                                (
+                                    AdcMode::SingleSample as usize,
+                                    self.channel.get(),
+                                    sample as usize,
+                                ),
                             )
                             .ok();
                     })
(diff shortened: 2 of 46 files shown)


alistair23 (dismissed):
I like this! I agree the current setup is confusing and easy to get wrong.

The other option we could do is an `enum` for the first argument, that way it isn't just a simple number.

alexandruradovici:
> I like this! I agree the current setup is confusing and easy to get wrong.
> 
> The other option we could do is an `enum` for the first argument, that way it isn't just a simple number.

How do you see this `enum`? Somthing that implements a certain trait?

lschuermann:
> How do you see this `enum`? Somthing that implements a certain trait?

I believe @alistair23 is referring to something akin to `capsules::driver::NUM` (if not that specifically):
https://github.com/tock/tock/blob/92eafdb491c3e6d14e6d798ccd8b0af959619cec/capsules/src/driver.rs#L6-L9

That being that, I don't think that would be a great idea. Driver numbers can be well-known, but it's also strictly possible and desired for developers to create their own numbers for custom drivers, or multiple instances of the same one. Thus, using a number seems much more sensible.

doc/Porting_v1_Capsules_to_v2.md
```diff
@@ -311,7 +311,7 @@ replaced with code like:
 
 ```rust
 self.apps.enter(appid, |app, upcalls| {
-    upcalls.schedule_upcall(upcall_number, r0, r1, r2);
+    upcalls.schedule_upcall(upcall_number,(r0, r1, r2));
```
bradjc, line 314:
```suggestion
    upcalls.schedule_upcall(upcall_number, (r0, r1, r2));
```

bradjc (approved):

bradjc (approved):

alistair23:
As @lschuermann said something like this:

```rust
pub enum CallbackNumber {
    ZERO = 0,
    ONE = 1,
    TWO = 2,
    THREE = 3,
    FOUR = 4,
    FIVE = 5,
    SIX = 6,
}
```

That way it's more obvious what is going on

ppannuto (approved):
I think this is good to go as-is. The `enum` is an interesting idea, but I'm not quite convinced the ergonomic overhead is going to be worth it (especially given that it also adds imports everywhere). I think I would also be worried as a user about in what ways `ZERO` is or is not just `0`, which might confuse more than help.

lschuermann (approved):
I agree with @ppannuto. The enum is indeed an interesting idea, but the fact that it requires imports, is another thing we keep around and maintain or generate through some macro, and is non-standard thus might confuse people makes me think that a simple numeric seems like the simpler option.

I've taken the liberty to transfer this out of the RFC phase, given it has been discussed on today's call.

brghena (approved):

PR #2709, opened 2021-07-30
tock: fix lowrisc/opentitan uart hw driver

ccteng-google:
The lowrisc/opentitan uart hw driver was not sending
tx_client.transmitted_buffer notification properly causing
app (user mode) console write syscall to never receive the
write complete subscription callback.

Tested with opentitan earlgrey running Tock OS in verilator
simulation and app calling console write to verify successful
return from syscall.

Change-Id: Id1e485373466d6aa414cc011b05b4b5c559e17e0

### Pull Request Overview

This pull request fixes the problem where 
the lowrisc/opentitan uart hw driver was not sending
tx_client.transmitted_buffer notification properly causing
Tock OS app (user mode) console write syscall to never
receive the write complete subscription callback.

### Testing Strategy

running opentitan earlgrey + Tock OS + app in verilator
simulation and app calling console write to verify successful
return from syscall

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/chips/lowrisc/src/uart.rs
+++ b/chips/lowrisc/src/uart.rs
@@ -192,18 +192,26 @@ impl<'a> Uart<'a> {
 
             // Read from the transmit buffer and send bytes to the UART hardware
             // until either the buffer is empty or the UART hardware is full.
-            self.tx_buffer.map(|tx_buf| {
+            if self.tx_buffer.map_or(false, |tx_buf| {
                 let tx_len = len - idx;
 
                 for i in 0..tx_len {
                     if regs.status.is_set(status::txfull) {
-                        break;
+                        return false;
                     }
                     let tx_idx = idx + i;
                     regs.wdata.write(wdata::data.val(tx_buf[tx_idx] as u32));
                     self.tx_index.set(tx_idx + 1)
                 }
-            });
+                true
+            }) {
+                // We sent everything to the UART hardware
+                self.tx_client.map(|client| {
+                    self.tx_buffer.take().map(|tx_buf| {
+                        client.transmitted_buffer(tx_buf, self.tx_len.get(), Ok(()));
+                    });
+                });
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


lschuermann:
This looks very similar to an issue I had with the LiteX codebase. Unfortunately I don't think this fix is entirely valid, given that as far as I can tell this will potentially issue a callback on the first call to `transmit` and this can cause other issues down the line. This is further illustrated in the HIL TRD.

In this case we will probably want to schedule a deferred call. I can recommend you to take a look here:

https://github.com/tock/tock/blob/ba3baef1a677704cbf8e15ef72a4161c41df1c10/chips/litex/src/uart.rs#L386-L402

chips/lowrisc/src/uart.rs
```diff
@@ -192,18 +192,26 @@ impl<'a> Uart<'a> {
...
                     regs.wdata.write(wdata::data.val(tx_buf[tx_idx] as u32));
                     self.tx_index.set(tx_idx + 1)
                 }
-            });
+                true
+            }) {
+                // We sent everything to the UART hardware
+                self.tx_client.map(|client| {
+                    self.tx_buffer.take().map(|tx_buf| {
+                        client.transmitted_buffer(tx_buf, self.tx_len.get(), Ok(()));
```
bradjc, line 211: Unfortunately we can't call a callback here. If I'm reading this driver correctly, `transmit_buffer()` is called from hil::uart::transmit::transmit_buffer(), which means a callback could occur in response to a HIL command. Per Tock's callback model, callbacks can _only_ be triggered in response to interrupts.

bradjc:
Beat me to it!

lschuermann:
As discussed on today's call, this will probably have to be fixed in some Tock 2.0 release candidate, but it's fine to tag the initial one without. I can also offer to just implement it.

ccteng-google:
@lschuermann Feel free to take it over and I'd be OK to close this PR.

lschuermann:
Let's leave this open, I think I should be able to push to your branch. I'll test with Verilator, but I'd appreciate you or @alistair23  doing HW tests.

chips/lowrisc/src/uart.rs
phil-levis: Callbacks do not have to be triggered in response to interrupts; they just can't be invoked on the same call stack from the causing downcall. This generally means that it needs to either be from an interrupt or a deferred procedure call.

But yes, invoking `transmitted_buffer` here is incorrect.

ccteng-google:
I will close this PR while I upgrade to latest opentitan for further investigation and reopen if needed.

lschuermann:
Reopening given @alistair23 has confirmed he can reproduce this bug, even on older commits of Tock and/or OpenTitan/EarlGrey.

phil-levis:
@alistair23 reports that his failure was not due to the UART, but due to memory allocation.

alistair23:
Just to clarify.

I was seeing failures with some of the print release tests yesterday. I just assumed it was this issue and I then tried to see if I could find the last working version, which I wasn't able to find.

After looking into it today I realised that the problem is that the memory requirements (the `memop` commands) for the app were failing which is why it wasn't working.

So I am unable to reproduce any failures on the FPGA.

PR #2711, opened 2021-07-30
make: remove nonexistent target

bradjc:
### Pull Request Overview

This target seems to not exist?

### Testing Strategy

### TODO or Help Wanted

I'm not sure this is the correct fix.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,6 @@ usage:
 	@echo "To run these, run 'make {target}' from the board directory for any of the"
 	@echo "following targets:"
 	@echo "        cargobloat: Runs the cargo-bloat tool for attributing binary size"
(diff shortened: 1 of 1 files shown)


hudson-ayers (changes requested):
This does exist. Lines 40/41 say: run `make {target}` from the *board directory*. Go to any boards/x directory and this target works correctly (the targets are in Makefile.common). They were referenced here for discoverability

hudson-ayers:
I think this is correct as-is and we should close this PR

lschuermann:
We might want to include a warning that these can't be issued from this Makefile instead?

bradjc:
Oh geez I don't have time to read all of that! Why did we even name them differently in the first place.

PR #2712, opened 2021-07-30
make: update stack-analysis format

bradjc:
### Pull Request Overview

Two things:

1. Makes the `make allstack` easier to read:

```
$ make stack-analysis
imix
----------------------
   main stack frame:
         2048     main

   5 largest stack frames:
         2048     main
          720     _ZN104_$LT$capsules..process_console..ProcessConsole$LT$C$GT$$u20$as$u20$kernel..hil..uart..TransmitClient$GT$18transmitted_buffer17h4c3fef6f6a95d58fE
          632     _ZN8capsules15process_console23ProcessConsole$LT$C$GT$11write_state17hee14a9c17699800aE
          608     _ZN7cortexm24kernel_hardfault_arm_v7m17h8c7b48b27a33c005E
          600     _ZN98_$LT$capsules..ieee802154..framer..Framer$LT$M$C$A$GT$$u20$as$u20$kernel..hil..radio..RxClient$GT$7receive17h8a8322dd16823f26E

   **WARNING! main is the largest stack frame!**
```

2. Enables `make stack-analysis V=1` to work fully.

### Testing Strategy

Running `make stack-analysis`.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/tools/stack_analysis.sh
+++ b/tools/stack_analysis.sh
@@ -4,13 +4,28 @@
 # size information, and is thus most easily called using the `make stack-analysis`
 # rule.
 
-# Print the stack frame size of `main`
-printf "main stack frame: \n"
-$(find $(rustc --print sysroot) -name llvm-readobj) --elf-output-style GNU --stack-sizes $1 | grep 'main'
+bold=$(tput bold)
+normal=$(tput sgr0)
 
+# Get a list of all stack frames and their sizes.
+frames=`$(find $(rustc --print sysroot) -name llvm-readobj) --elf-output-style GNU --stack-sizes $1`
+
+# Print the stack frame size of `main`
+printf "   main stack frame: \n"
+echo "$frames" | grep ' main' # Use a space before main to avoid functions with main in the name.
 printf "\n"
-printf "5 largest stack frames: \n"
 
 # Print the 5 largest stack frames
-$(find $(rustc --print sysroot) -name llvm-readobj) --elf-output-style GNU --stack-sizes $1 | sort -n -r | head -5
+printf "   5 largest stack frames: \n"
+echo "$frames" | sort -n -r | head -5
 printf "\n"
+
+# Check if main is the largest stack frame
+largest=`echo "$frames" | sort -n -r | head -1 | grep ' main'`
+largest_ret_val=$?
+
+# If it is, print a warning.
+if [ $largest_ret_val -eq 0 ]; then
+    echo "   ${bold}WARNING! main is the largest stack frame!${normal}"
+    printf "\n"
+fi
--- a/tools/stack_analysis.sh
+++ b/tools/stack_analysis.sh
@@ -27,5 +27,8 @@ largest_ret_val=$?
 # If it is, print a warning.
 if [ $largest_ret_val -eq 0 ]; then
     echo "   ${bold}WARNING! main is the largest stack frame!${normal}"
+    echo "   See https://github.com/tock/tock/issues/2425 for an explanation of"
+    echo "   why this is an issue, and https://github.com/tock/tock/pull/2715 for"
+    echo "   an example of how to fix it."
     printf "\n"
 fi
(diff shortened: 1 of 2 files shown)


tools/stack_analysis.sh
```diff
@@ -4,13 +4,28 @@
 # size information, and is thus most easily called using the `make stack-analysis`
 # rule.
 
-# Print the stack frame size of `main`
-printf "main stack frame: \n"
-$(find $(rustc --print sysroot) -name llvm-readobj) --elf-output-style GNU --stack-sizes $1 | grep 'main'
+bold=$(tput bold)
+normal=$(tput sgr0)
```
lschuermann, line 8: Just a heads-up this adds a dependency on `ncurses`. It's shipped with my distro and seems to be on many, but it's worth keeping in mind - for instance I'm not quite sure whether Cygwin would have it by default.

lschuermann (approved):
Pretty cool.

tools/stack_analysis.sh
ppannuto: We've had `tput` around in the tools and build system for a long time; not a new dependency here, so I think it's fine.

tools/stack_analysis.sh
```diff
@@ -4,13 +4,28 @@
...
+echo "$frames" | sort -n -r | head -5
 printf "\n"
+
+# Check if main is the largest stack frame
+largest=`echo "$frames" | sort -n -r | head -1 | grep ' main'`
+largest_ret_val=$?
+
+# If it is, print a warning.
+if [ $largest_ret_val -eq 0 ]; then
+    echo "   ${bold}WARNING! main is the largest stack frame!${normal}"
```
ppannuto, line 29: Why is this something we warn about?

ppannuto (approved):

tools/stack_analysis.sh
hudson-ayers: That stack frame is never dropped

hudson-ayers (approved):

phil-levis:
We need to be more careful and cautious about changing the output of tools. It can be that other tools depend on their output, such that changing it breaks those tools. This is an example of a churn/incremental improvement change which can be frustrating to people downstream.

alistair23 (approved):

tools/stack_analysis.sh
ppannuto:
```suggestion
# Check if main is the largest stack frame
largest=`echo "$frames" | sort -n -r | head -1 | grep ' main'`
largest_ret_val=$?

# If it is, print a warning.
if [ $largest_ret_val -eq 0 ]; then
    echo "   ${bold}WARNING! main is the largest stack frame!${normal}"
    echo "   See https://github.com/tock/tock/issues/2425 for an explanation of"
    echo "   why this is an issue, and https://github.com/tock/tock/pull/2715 for"
    echo "   an example of how to fix it."
```

bradjc:
This tool has only existed for 138 days. Are you aware of any downstream scripts that process the output?

hudson-ayers:
I am not aware of anyone processing the output of this script downstream, and downstream users are welcome to vendor a copy of the old version if they want the old output format. I do not think that every tool committed to the Tock repository should be locked into its original format, when I published this originally it was just a quick way to let others experiment with solutions to our stack size woes.

hudson-ayers (approved):

ppannuto (approved):

alistair23 (approved):

bradjc:
ping

ppannuto:
bors r+

Phil's point about output stability is well-taken, particularly for things that are 'promoted' (i.e. this isn't just a script in the `tools/` directory, it's advertised via the primary `make` interface). That said, I think the argument that this is a very new tool still figuring out its interface and output holds here. However, that becomes less true after this is bundled as a new `make`-accessible tool with 2.0.

hudson-ayers:
bors retry

PR #2714, opened 2021-07-30
make: add aliases for root makefile commands

bradjc:
### Pull Request Overview

In hindsight I'm not sure why we prepended "all" to our root makefile commands. I'm also not sure why we call some commands one thing in boards/ and something else in the root.

This adds aliases so the makefile just does what the user wants rather than throw errors. But if you are used to `allX` those still work.

### Testing Strategy

Running make commands.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


lschuermann (approved):
I don't think it's great to keep both `allX` and `X` around, but I admit old habits die hard.

ppannuto:
In the spirit of 2.0 being the time for breaking changes, I would be okay with just removing the `allX` variants

bradjc:
We can change the `make` printout, but swapping just annoys a different set of people.

hudson-ayers (approved):

phil-levis:
Changing the CLI is non-trivial. Can you update the self-documentation (`usage`) accordingly?

alistair23 (dismissed):

bradjc:
I didn't change anything. This just makes the makefile a little closer to "just work" to _not_ annoy users.

I got tripped up by reading:

```
There are a few helpful targets that can be run for all individual boards.
To run these, run 'make {target}' from the board directory for any of the
following targets:
        cargobloat: Runs the cargo-bloat tool for attributing binary size
        stack-analysis: Prints the 5 largest stack frames for the board
```

then:

```
$ make stack-analysis
make: *** No rule to make target `stack-analysis'.  Stop.
```

Not user friendly. But `make allstack` (the command you should run if you read the help in detail) still works.

ppannuto (dismissed):

bradjc:
ping

ppannuto:
I updated the `usage` to prefer the new aliases per Phil's suggestion. I also cleaned up a bit of the rest of the `usage` text while I was there.

ppannuto (approved):

hudson-ayers (approved):

PR #2717, opened 2021-07-30
process: make ProcessFault a generic parameter

jettr:
Instead of using dynamic dispatching for `ProcessFault`, use a generic parameter. Most times the `ProcessFault` struct is a 0-sized struct and will not take up any space. A few instances of `ProcessFault` have a usize field, which is still less than a trait object (2 words minimum).

Final images will only call load_process once, so we don't get bloat from monomorphism either.

### Documentation Updated

None

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/acd52832/src/main.rs
+++ b/boards/acd52832/src/main.rs
@@ -592,7 +592,7 @@ pub unsafe fn main() {
             &_eappmem as *const u8 as usize - &_sappmem as *const u8 as usize,
         ),
         &mut PROCESSES,
-        &FAULT_RESPONSE,
+        FAULT_RESPONSE,
         &process_management_capability,
     )
     .unwrap_or_else(|err| {
--- a/boards/arty_e21/src/main.rs
+++ b/boards/arty_e21/src/main.rs
@@ -282,7 +282,7 @@ pub unsafe fn main() {
             &_eappmem as *const u8 as usize - &_sappmem as *const u8 as usize,
         ),
         &mut PROCESSES,
-        &FAULT_RESPONSE,
+        FAULT_RESPONSE,
         &process_mgmt_cap,
     )
     .unwrap_or_else(|err| {
--- a/boards/clue_nrf52840/src/main.rs
+++ b/boards/clue_nrf52840/src/main.rs
@@ -663,7 +663,7 @@ pub unsafe fn main() {
             &_eappmem as *const u8 as usize - &_sappmem as *const u8 as usize,
         ),
         &mut PROCESSES,
-        &FAULT_RESPONSE,
+        FAULT_RESPONSE,
         &process_management_capability,
     )
     .unwrap_or_else(|err| {
--- a/boards/earlgrey-nexysvideo/src/main.rs
+++ b/boards/earlgrey-nexysvideo/src/main.rs
@@ -545,7 +545,7 @@ unsafe fn setup() -> (
             &_eappmem as *const u8 as usize - &_sappmem as *const u8 as usize,
         ),
         &mut PROCESSES,
-        &FAULT_RESPONSE,
+        FAULT_RESPONSE,
         &process_mgmt_cap,
     )
     .unwrap_or_else(|err| {
--- a/boards/esp32-c3-devkitM-1/src/main.rs
+++ b/boards/esp32-c3-devkitM-1/src/main.rs
@@ -278,7 +278,7 @@ unsafe fn setup() -> (
             &_eappmem as *const u8 as usize - &_sappmem as *const u8 as usize,
         ),
         &mut PROCESSES,
-        &FAULT_RESPONSE,
+        FAULT_RESPONSE,
         &process_mgmt_cap,
     )
     .unwrap_or_else(|err| {
--- a/boards/hail/src/main.rs
+++ b/boards/hail/src/main.rs
@@ -486,12 +486,6 @@ pub unsafe fn main() {
     // );
     // peripherals.pa[16].set_client(debug_process_restart);
 
-    // Configure application fault policy
-    let fault_policy = static_init!(
-        kernel::process::ThresholdRestartThenPanicFaultPolicy,
-        kernel::process::ThresholdRestartThenPanicFaultPolicy::new(4)
-    );
-
     let scheduler = components::sched::round_robin::RoundRobinComponent::new(&PROCESSES)
         .finalize(components::rr_component_helper!(NUM_PROCS));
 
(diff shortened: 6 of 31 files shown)


lschuermann:
I think the idea behind `ProcessFault` was that we could potentially have more complex logic incorporated in some struct, which then implements `ProcessFault` and potentially other traits, and is used in multiple places throughout the kernel. If we move it into the `load_processes`, we require it to be `Clone` and can't also use it in other places (except if we were to build a wrapper encapsulating a reference, which is not great).

I agree that we should avoid dynamic dispatch. Thus I propose that we make this a generic parameter, but still pass in a reference. This way we still allow the compiler to perform many optimizations not possible with trait objects, but are significantly more flexible in what a `ProcessFault` implementation can do. For a zero-sized `ProcessFault`, out of experience, the compiler is also smart enough to figure out it doesn't need to use the reference then. In the worst case, this requires a single word for the reference - which can be even less than `Copy`ing and moving a struct into every `Process`.

lschuermann:
After our discussion yesterday, I think this should be a post 2.0 change. It doesn't significantly change kernel interfaces (only boards) and does not introduce any desired features or bug fixes required for a 2.0 release.

phil-levis:
Having done a few rewrites of the process loading code, I'm generally in favor of removing this dynamic dispatch and reducing the amount of flexibility supported. This seems like a "would be nice to have" feature, which unfortunately means that since it isn't really used it's bloat. A simple approach that always restarts a process seems mostly sufficient; we now have `exit-terminate` if a process wants to not restart.

jettr:
I agree this is a post 2.0 features as well (I didn't know we were using the `Blocked` label for that; I would have added it initially).

I was on the fence about passing the object by value or by reference; I think it is pretty close either way. The  `Copy`  code for 0-size should be empty, and the copy code for a `usize`d struct should be a single instruction. I think passing by value here is the smaller code size and not optimizing for a case we don't have yet. That said, I am not opposed to updating passing by reference if others feel strongly about it.

Also, even though we pass `ProcessFault` into the process by value, we can still pass it as a reference from there to other section of the code. Now that I am thinking about this more, I do think each Process should have its own copy of the policy to support each process tracking its own `ProcessFault` state in the future without having a globally shared `ProcessFault` object (if we weighing discussions of where this could go in the future ;) )

jrvanwhy:
Would the following trait definition work? Note we pass `self` by value.

```rust
pub trait ProcessFaultPolicy: Copy {
    /// Decide which action the kernel should take in response to `process`
    /// faulting.
    fn action(self, process: &dyn Process) -> process::FaultAction;
}
```

This lets you implement `ProcessFaultPolicy` on a zero-sized type or a reference to a non-zero-sized type.

jettr:
> Would the following trait definition work? Note we pass `self` by value.
> 
> ```rust
> pub trait ProcessFaultPolicy: Copy {
>     /// Decide which action the kernel should take in response to `process`
>     /// faulting.
>     fn action(self, process: &dyn Process) -> process::FaultAction;
> }
> ```
> 
> This lets you implement `ProcessFaultPolicy` on a zero-sized type or a reference to a non-zero-sized type.

For non-zero-sized types, I don't understand why we would want to copy the `ProcessFaultPolicy` object when we call `action` on it. Why would we not just want to use it as a reference. What am I missing here?

We need the `Copy` trait to make the initial copies of the fault policy in kernel's `load_processes` function. But after each process has its own `ProcessFaultPolicy` copy, then each process should use that one instance (and use references to it instead of copying it).

This is making me thinking of a larger refactor: we could potentially pass in a `ProcessFaultPolicy` for every process that is passed in to `load_processes`, that way the board/main could decide that some processes can be restarted while other processes must fatally fault. Then `ProcessFaultPolicy` would not need the have `Copy` as a super trait.

jrvanwhy:
> For non-zero-sized types, I don't understand why we would want to copy the `ProcessFaultPolicy` object when we call `action` on it. Why would we not just want to use it as a reference. What am I missing here?

For non-zero-sized types, `self` would be a reference type. I.e. the data is stored in type `T` and the implementation is:

```rust
impl ProcessFaultPolicy for &T {
    fn action(self, process: &dyn Process) -> process::FaultAction {
        ...
    }
}
```

jettr:
> Perhaps `&ProcessFaultPolicy` should be part of `KernelResources` now? The intent being that this is a kernel-level config option, and not stored on a per-process basis. What might be clumsy, however, is the interface between kernel.rs and process_standard.rs to tell ProcessStandard what the resulting `FaultAction` is, but perhaps there is a nice way to organize that.

So if `&ProcessFaultPolicy` is on `KernelResources`, it still would take in a `Process` so it could potential perform different actions per-process, yet there is only one `ProcessFaultPolicy` object alive in the system. This sounds like a reasonable way to go. If that sounds good, I will upload a PR for that in the next week or so.

phil-levis:
Can you quantify what the benefits of this are? After trying to rewrite process loading/handling code several times, but conclusion was that a more significant change (just having a static policy) actually had pretty tiny effects.

That being said, now that we have exit-terminate and exit-restart, I am less convinced that we need a failure policy embedded programmatically within a process structure. Conceptually, this is about the loading policy (e.g., the equivalent of `init.d`) and so should be part of the loader. This gets us into the weird dance of first the process policy decides if it wants to restart

https://github.com/tock/tock/blob/a05b654a5b16d94c05c82e55a30a341ec0d8ddda/kernel/src/process_standard.rs#L319

then it's left to the kernel to decide whether to restart, but this is embedded within the `Process` implementation:

https://github.com/tock/tock/blob/a05b654a5b16d94c05c82e55a30a341ec0d8ddda/kernel/src/process_standard.rs#L345

I'd suggest that this makes more sense to be a programmatic component of the `Process` API: that would allow, for example, the process console (or any other kernel service with the right Capability) to dynamically change the fault behavior of processes.

jettr:
The original benefits were to reduce a small amount of flash space by using a generic trait instead of `dyn`, and to follow that paradigm in more places.

I am not that attached to this PR; it was more of a low hanging fruit cleanup. However it turns out that this fruit was attached to much large fruit up the tree.

I am okay closing this PR in favor of discussing the application restart API with respect to the `main` function discussion we were having last core call. The process restart policy does seem to fit into that larger picture.

phil-levis:
> The original benefits were to reduce a small amount of flash space by using a generic trait instead of `dyn`, and to follow that paradigm in more places.

Yeah, I am totally in agreement with this general goal. When I went to rewrite process loading, I tried to follow this and found that its effects in practice on this particular case were pretty tiny. So I was wondering if you'd done a measurement. It could be that as a smaller change it has greater impact (or perhaps not). But, for example, if this saved 1kB I'd react very differently than if it saved 20B.

bradjc:
We could go back to having a `ProcessStandard.fault_action: FaultAction` field for each process, which is one of `Enum { Stop, Restart}`, and merge `ProcessFaultPolicy` into the `trait ProcessFault` we already have.

jettr:
Cleaning/Closing this PR as we need to have a large refactor discussion.

PR #2720, opened 2021-08-02
Process console: iterate each process on command `list`

bradjc:
### Pull Request Overview

If a board has many process loaded then the process console list command looks like:

```
list
 PID    Name                Quanta  Syscalls  Dropped Callbacks  Restarts    State  Grants
  0	blink                    0        66                  0         0  Yielded    1/14
  1	c_hello                  0         8                  0         0  Yielded    1/14
  2	buttons                  0         7                  0         0
```
because the buffer fills up and printing stops.

This PR attempts to move the `list` command to an async command and have each process print on its own UART tx callback cycle. To do that I had to update the logic a little, as it seemed to assume the callbacks would only be used for printing callback state.

Now it looks like:

```
list
 PID    Name                Quanta  Syscalls  Dropped Callbacks  Restarts    State  Grants
  0	blink                    0      1002                  0         0  Yielded    1/14
  1	c_hello                  0         8                  0         0  Yielded    1/14
  2	buttons                  0         7                  0         0  Yielded    1/14
  3	blink                    0      1002                  0         0  Yielded    1/14
```

### Testing Strategy

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/capsules/src/process_console.rs
+++ b/capsules/src/process_console.rs
@@ -151,15 +151,15 @@ enum WriterState {
     KernelStack,
     KernelRoData,
     KernelText,
-    ProcessStart,
-    ProcessGrant,
-    ProcessHeap,
-    ProcessHeapUnused,
-    ProcessData,
-    ProcessStack,
-    ProcessStackUnused,
-    ProcessFlash,
-    ProcessProtected,
+    ProcessStart { process_id: ProcessId },
+    ProcessGrant { process_id: ProcessId },
+    ProcessHeap { process_id: ProcessId },
+    ProcessHeapUnused { process_id: ProcessId },
+    ProcessData { process_id: ProcessId },
+    ProcessStack { process_id: ProcessId },
+    ProcessStackUnused { process_id: ProcessId },
+    ProcessFlash { process_id: ProcessId },
+    ProcessProtected { process_id: ProcessId },
 }
 
 impl Default for WriterState {
@@ -192,7 +192,6 @@ pub struct ProcessConsole<'a, C: ProcessManagementCapability> {
     queue_buffer: TakeCell<'static, [u8]>,
     queue_size: Cell<usize>,
     writer_state: Cell<WriterState>,
-    writer_process: Cell<Option<ProcessId>>,
     rx_in_progress: Cell<bool>,
     rx_buffer: TakeCell<'static, [u8]>,
     command_buffer: TakeCell<'static, [u8]>,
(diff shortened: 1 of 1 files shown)


capsules/src/process_console.rs
```diff
@@ -1033,37 +1090,29 @@ impl<'a, C: ProcessManagementCapability> uart::TransmitClient for ProcessConsole
...
-        if self.writer_state.get() != WriterState::Empty
-            && self.writer_state.get() != WriterState::KernelStart
-            && self.writer_state.get() != WriterState::ProcessStart
-        {
-            self.write_state(WriterState::Empty, None);
-        }
+        // Check if we have anything queued up. If we do, let the queue
+        // empty.
+        let ret = self.handle_queue();
+        if ret.ok() == Some(0) || ret.is_err() {
```
phil-levis, line 1101: It seems like this logic is such that if the underlying UART is busy during this callback (https://github.com/tock/tock/blob/f33ce870aea456acaf40c2a66ec880d2142e1249/capsules/src/process_console.rs#L1046), then this will lead this to enqueue more. I'm wondering if the `is_err()` case is correct here. Can you walk me through how this is supposed to work?

bradjc: The intent is: If for whatever reason handling the queue did not cause a UART TX (and therefore we will not get another `transmitted_buffer()` callback), then we continue handling other work that might need to use the UART TX.

Since we set `tx_in_progress` in progress to false immediately before this, handle_queue can't fail due to the TX being busy in this case. Perhaps there could be other failure cases in handle_queue in the future.

capsules/src/process_console.rs
```diff
@@ -318,15 +317,37 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
...
+            WriterState::ProcessStackUnused { process_id } => {
+                WriterState::ProcessFlash { process_id }
+            }
+            WriterState::ProcessFlash { process_id } => {
+                WriterState::ProcessProtected { process_id }
+            }
+            WriterState::ProcessProtected { process_id: _ } => WriterState::Empty,
+            WriterState::List { index, total } => {
+                // Next state just increments index, unless we are at end in
+                // which next state is just the empty state.
```
hudson-ayers, line 341:
```suggestion
                // which case next state is just the empty state.
```

capsules/src/process_console.rs
```diff
@@ -1024,6 +1036,51 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
...
+
+                    // Copy elements of the queue into the TX buffer.
+                    (&mut txbuf[..txlen]).copy_from_slice(&qbuf[..txlen]);
+
+                    // TODO: If the queue needs to print over multiple TX
+                    // buffers, we need to shift the remaining contents of the
+                    // queue back to index 0.
+                    // if qlen > txlen {
+                    //     (&mut qbuf[txlen..qlen]).copy_from_slice(&qbuf[txlen..qlen]);
+                    // }
```
hudson-ayers, line 1068: Do we still need to address this?

This looks like something that should use `core::slice::copy_within()`, if the issue is just that this would not compile.

bradjc: Yes this would be an improvement, in case we cannot send the entire queued buffer.

PR #2721, opened 2021-08-02
kernel: use mutable closure on `process_each` and `process_each_capability`

bradjc:
### Pull Request Overview

I do not know the implications of this change, but it is necessary for #2720 so that the closure can modify local variables outside of the closure in each iteration.

### Testing Strategy

Working on the process console.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers (approved):
I see no reason not to allow each iteration of the closure to modify the environment it captures here.

alistair23 (approved):

PR #2724, opened 2021-08-03
debug: rename begin_debug_fmt and add non-newline version

jettr:
Remove `begin` prefix from `begin_debug_fmt` and `begin_debug_verbose_fmt` since there does not appear to be a finish or end that should accompany it.

Also add a non-newline version of this function.

This does not need to make it in by Tock 2.0 release

### Documentation Updated

- None

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/debug.rs
+++ b/kernel/src/debug.rs
@@ -527,7 +527,14 @@ impl Write for DebugWriterWrapper {
     }
 }
 
-pub fn begin_debug_fmt(args: Arguments) {
+pub fn debug_print(args: Arguments) {
+    let writer = unsafe { get_debug_writer() };
+
+    let _ = write(writer, args);
+    writer.publish_bytes();
+}
+
+pub fn debug_println(args: Arguments) {
     let writer = unsafe { get_debug_writer() };
 
     let _ = write(writer, args);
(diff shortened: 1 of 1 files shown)


alistair23 (dismissed):

hudson-ayers (dismissed):

kernel/src/debug.rs
```diff
@@ -527,23 +527,45 @@ impl Write for DebugWriterWrapper {
...
+pub fn debug_format(args: Arguments) {
     let writer = unsafe { get_debug_writer() };
 
     let _ = write(writer, args);
-    let _ = writer.write_str("\r\n");
     writer.publish_bytes();
 }
 
-pub fn begin_debug_verbose_fmt(args: Arguments, file_line: &(&'static str, u32)) {
+pub fn debug_formatln(args: Arguments) {
```
ppannuto, line 537:
```suggestion
pub fn debug_println(args: Arguments) {
```

If we're going to rename, I think this makes more sense -- I think I would expect a `formatln` to just write bytes into a buffer; this does the print.

kernel/src/debug.rs
```diff
@@ -527,23 +527,45 @@ impl Write for DebugWriterWrapper {
...
 
     let (file, line) = *file_line;
     let _ = writer.write_fmt(format_args!("TOCK_DEBUG({}): {}:{}: ", count, file, line));
     let _ = write(writer, args);
+}
+
+pub fn debug_verbose_format(args: Arguments, file_line: &(&'static str, u32)) {
+    let writer = unsafe { get_debug_writer() };
+
+    print_debug_header(writer, args, file_line);
```
ppannuto, line 561:
```suggestion
    print_debug_header(writer, file_line);
    let _ = write(writer, args);
```

Feels like this part should be here (or the "print header" function should be renamed to something like `debug_println_with_header`)

ppannuto (changes requested):
Overall looks fine -- I think the `begin` is a legacy of a long-dead implementation of debug.

Some clarification comments on the naming that I think/hope makes things a bit clearer

kernel/src/debug.rs
jettr: Ah yes, I was trying to dedup that extra line, but then the function name gets tricky. I have moved the arg print back into both functions and updated the function names. Thanks!

kernel/src/debug.rs
jettr: Good call. Done.

hudson-ayers (approved):

ppannuto (approved):
bors d+

ppannuto:
Just blocked on 2.0

lschuermann (approved):

alistair23 (approved):

PR #2725, opened 2021-08-04
First step on SPI documentation

phil-levis:
### Pull Request Overview

This pull request updates the comments on `kernel::hil::spi` to document the current interface. It also adds a first draft for a SPI TRD, which has some significant changes to the interface. The first draft of the TRD is intended for starting the discussion towards defining a solid SPI API that follows modern Tock API design principles.

### Testing Strategy

### TODO or Help Wanted

This pull request needs feedback on the SPI comments and a start of discussion around the SPI TRD.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/reference/trd-spi.md
+++ b/doc/reference/trd-spi.md
@@ -6,9 +6,9 @@ Kernel Serial Peripheral Interface (SPI) HIL
 **Type:** Documentary<br/>
 **Status:** Draft <br/>
 **Author:** Philip Levis, Alexandru Radovici<br/>
-**Draft-Created:** 2021/08/02 <br/>
-**Draft-Modified:** 2021/08/02 <br/>
-**Draft-Version:** 1 <br/>
+**Draft-Created:** 2021/08/13 <br/>
+**Draft-Modified:** 2021/08/13 <br/>
+**Draft-Version:** 2 <br/>
 **Draft-Discuss:** tock-dev@googlegroups.com<br/>
 
 Abstract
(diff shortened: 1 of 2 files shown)


doc/reference/trd-spi.md
```diff
@@ -0,0 +1,359 @@
...
+The `ControllerBus` trait combines `ControllerDevice` and `ChipSelect` to
+provide the full abstraction of a SPI bus. It is the trait that chip
+SPI implementations provide. In addition to `ControllerDevice` and
+`ChipSelect`, `ControllerBus` includes an `init` method. This `init`
+method initializes the hardware to be a SPI controller and is typically
+called at boot.
+
+```rust
+pub trait ControllerBus<'a>: ControllerDevice<'a> + ChipSelect {
+  fn init(&self) -> Result<(), ErrorCode>;
```
lschuermann, line 212: I know we had this discussion just last Friday, but it might make sense to be pedantic about it if we're formalizing it now: wouldn't we want to rather have this be `enable` and `disable`, such that one can switch between modes of underlying hardware. That's commonly supported and so I think we should expose it through this API for the advanced users.

The logic would be that a shared HW-device would need to have `spi::ControllerBus::disable` be called before something like `uart::ControllerBus::enable` can be called.

For HW which doesn't support deinitialization, we could just return an error on `disable`.

I'd like to avoid being able to "convert" a shared-HW device to a specific mode implicitly by calling `init`, without the need to uninitialize it explicitly first.

phil-levis: I'd argue that the case you're describing would be neither of these APIs: it would be an API which supports moving between controller and peripheral mode. `init` is intended to be a one-time operation that's done at boot. These traits are intended for a SPI bus being in dedicated controller mode or dedicated peripheral mode. The fact that there's a syscall driver capsule that uses both is a bit of a historical artifact (and as far as I can tell, almost never used).

phil-levis:
One issue that came up while I was writing the UART TRD #2756 : SPI and the UART have different behavior when the passed length is longer than the buffer. The UART returns `Err(SIZE)` while SPI truncates it to the length of the buffer. We should probably have a unified approach. I think the UART one is better. Changing this, though, is a pretty significant change to a HIL in that it will require lots of testing to make sure clients aren't making huge writes and expecting they'll just be truncated automatically.

phil-levis:
I've updated the TRD based on the call today, so now `read_write_bytes` returns `Err(SIZE)` if `len` is longer than one of the passed buffers. It also says that a call to `Peripheral::read_write_bytes` returns `Err(EINVAL)` if both buffers `None`.

phil-levis:
@bradjc -- Let me know if these SPI comments are sufficient. If so, then I think this is ready. The TRD will involve further work after 2.0, this just starts to sketch what the APIs look like.

bradjc (approved):

doc/reference/trd-spi.md
lschuermann: Okay, that's one way of looking at the problem. I was hoping if we were to have an `enable`/`disable` API we could have our cake and eat it too: it could allow us to both support switching device modes as in the USART example, as well as selectively power down components when they are not used. It'd both just be a `disable` away.

phil-levis: You want to separate enable/disable from power; some components can have significant power-up times (generally for things like oscillators to settle), such that you want to distinguish power state transitions from functional capabilities.

PR #2728, opened 2021-08-04
Add remove_mpu_region() functionality to mirror allocate_region()

jettr:
### Pull Request Overview

When a users maps a new mpu region to a process with `allocate_region`, they may want to later remove that region. This PR adds the appropriate removal function to the process interface and implements the new MPU helper function (`remove_memory_region`) for the MPU implementation that already implement `allocate_region`

This can wait until after Tock OS 2.0 release.

### Testing Strategy

- I have tested a variant of this code on 1.X systems. I unfortunately don't have a system to test this change on that is closer to the 2.0 kernel

### Help Wanted

 - [x] Test on RISCV with newer kernel
 - [x] Test on cortexM with newer kernel

### Documentation Updated

- None

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/mpu.rs
+++ b/arch/cortex-m/src/mpu.rs
@@ -252,6 +252,14 @@ pub struct CortexMRegion {
     attributes: FieldValue<u32, RegionAttributes::Register>,
 }
 
+impl PartialEq<mpu::Region> for CortexMRegion {
+    fn eq(&self, other: &mpu::Region) -> bool {
+        self.location.map_or(false, |(addr, size)| {
+            addr == other.start_address() && size == other.size()
+        })
+    }
+}
+
 impl CortexMRegion {
     fn new(
         logical_start: *const u8,
@@ -512,6 +520,28 @@ impl<const NUM_REGIONS: usize, const MIN_REGION_SIZE: usize> mpu::MPU
         config.is_dirty.set(true);
 
         Some(mpu::Region::new(start as *const u8, size))
+    }
+
+    fn remove_memory_region(
+        &self,
+        region: mpu::Region,
+        config: &mut Self::MpuConfig,
+    ) -> Result<(), ()> {
+        let (idx, _r) = config
+            .regions
+            .iter()
+            .enumerate()
+            .find(|(_idx, r)| **r == region)
+            .ok_or(())?;
+
+        if idx == APP_MEMORY_REGION_NUM {
+            return Err(());
+        }
+
+        config.regions[idx] = CortexMRegion::empty(idx);
+        config.is_dirty.set(true);
+
+        Ok(())
     }
 
     fn allocate_app_memory_region(
--- a/arch/rv32i/src/epmp.rs
+++ b/arch/rv32i/src/epmp.rs
@@ -180,6 +180,12 @@ pub struct PMPRegion {
     cfg: registers::FieldValue<u8, pmpcfg::Register>,
 }
 
+impl PartialEq<mpu::Region> for PMPRegion {
+    fn eq(&self, other: &mpu::Region) -> bool {
+        self.location.0 == other.start_address() && self.location.1 == other.size()
+    }
+}
+
 impl fmt::Display for PMPRegion {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         fn bit_str<'a>(reg: &PMPRegion, bit: u8, on_str: &'a str, off_str: &'a str) -> &'a str {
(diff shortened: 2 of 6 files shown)


alistair23:
Do we have an example on when this would be used?

For tests you can add a OpenTitan unit test adding and removing a region. That would allow testing with 2.0

jettr:
> Do we have an example on when this would be used?

We have an out-of-tree component that dispatches large buffers from the kernel to applications. We don't want to copy these large buffers, so we temporarily give a particular application access to the buffer that is in the kernel region. When the app is done with it, and before the kernel can reuse it, the kernel will unmap that buffer from the application.

> For tests you can add a OpenTitan unit test adding and removing a region. That would allow testing with 2.0

I found the `earlgrey-nexysvideo` board that appears to be the OpenTitian unit test vehicle you are talking about. I didn't see that there were any applications that I could attach to, since this MPU setting is really only meaningful to application. Can you point me to any test applications I can add to?

jettr:
I will hopefully be able to test this with RISCV based core with Tock 2.0 very soon. Could you help me test the CortexM version (or should I just remove that code change from this PR)?

alistair23:
Is this something we want in mainline then? Mainline is not going to pass kernel regions to applications so we don't have a use for removing mpu regions.

Whoops, I wasn't thinking. The unit tests aren't helpful for PMP tests as you are right they don't apply to M mode only. In terms of adding a test your best bet is to add support to libtock-c and then add an example app. At that point it's easy for others to run the test (I can run it on a Cortex-M board as well).

bradjc (approved):
I think this is fine.

jettr:
I was finally able to test this on a RISC-V system with a new kernel. It works :) I didn't have the patch locally and I was getting MPU failures when trying to add the same region again.

I don't have a cortex M system to test with. Should we
1. Get someone to test for me please
2. Realize that the for Cortex M and RISV is similar so it should be okay
3. Drop Cortex M from the PR

Thoughts?

hudson-ayers:
What is the easiest way for someone to test for you on cortex-m? Just modify a kernel to add/remove/add an MPU region and verify it works?

jettr:
> What is the easiest way for someone to test for you on cortex-m? Just modify a kernel to add/remove/add an MPU region and verify it works?

Yes, Trying adding the same mpu region twice. The second add should fail. Then try to remove the region between the two add calls, and the second add should succeed. Also it is worth testing that you cannot access that buffer while it is removed in the app (should get an app hardware fault)

bradjc:
Marking last call. Planning to merge soon (once rebased) unless someone has an objection.

lschuermann:
I can offer to try and test it tomorrow on an nRF52.

jettr:
> once rebased

Is that something I do as the PR owner, or the core team does before the merge? I am more thank happy to rebase; just let me know what the expectations are :)

bradjc:
You as the owner, generally :).

hudson-ayers:
> > once rebased
> 
> Is that something I do as the PR owner, or the core team does before the merge? I am more thank happy to rebase; just let me know what the expectations are :)

Typically rebasing is done by the PR owner when there is a conflict

jettr:
Good to know :) Rebase done.

bradjc (approved):

alistair23:
Can we add a unit test to make sure that:

```rust
add(1)
remove(1)
add(1)
```

works. and that 

```rust
add(1)
add(1)
```

Doesn't?

Even if we don't test accesses from usersapce it will be nice to know that we don't get errors

jettr:
> Can we add a unit test to make sure that:
> 
> ```rust
> add(1)
> remove(1)
> add(1)
> ```
> 
> works. and that
> 
> ```rust
> add(1)
> add(1)
> ```
> 
> Doesn't?
> 
> Even if we don't test accesses from usersapce it will be nice to know that we don't get errors

What test harness would I be doing that from? Wouldn't that just be testing the implementation of the host emulation chip support?

hudson-ayers:
I believe Alistair is referring to the QEMU-based unit tests in `boards/opentitan/src/tests/`. Unfortunately it seems that the directory reshuffle has broken those tests, and even if I fix the directory layout there is a fatal exception error in the hmac load binary test -- we should really add those tests to the existing QEMU CI!

alistair23:
I am referring to the unit tests in `boards/opentitan/src/tests`. You are right Hudson in that they don't currently work on QEMU (I'll send a PR now) but they do work on hardware.

By adding unit tests we would be testing that future changes to PMP/ePMP don't cause regressions inside the Tock kernel.

hudson-ayers:
I tested this on cortex-m today (Imix). add/remove/add works as expected, and add/add fails as expected. In order to run this test I had to modify the core ipc code, as I think there is nowhere else in the code where we add MPU regions dynamically today, and then run the IPC apps. 

I took a stab at adding unit tests, but got frustrated enough trying to fix the QEMU test infrastructure that I gave up (and I don't have a hardware OT board).

hudson-ayers (approved):

PR #2731, opened 2021-08-05
cdc: delay before sending data

bradjc:
### Pull Request Overview

I started testing the nano33 and found that the initial debug! messages ("Initialization...") were not showing up on my new mac with tockloader. This PR adds a 100 ms delay between when the CDC code detects a terminal is connected and when it actually starts sending data. This is inspired by https://michael.stapelberg.ch/posts/2021-04-27-linux-usb-virtual-serial-cdc-acm/.

I also added an additional check for receiving the `SET_CONTROL_LINE_STATE` message (per the blog post). So now the CDC start procedure looks like:

1. Start a 10 second timer. If this goes off we start dropping UART messages. Otherwise, we buffer and apply backpressure as necessary.
2. Wait for both `SET_CONTROL_LINE_STATE` and `SET_LINE_CODING` with baud rate 115200 to happen.
3. Then wait 100 ms.
4. Then start transmitting data.

### Testing Strategy

Verifying I can see the first debug prints on nano33ble on a new (but not M1 new) mac.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/usb/cdc.rs
+++ b/capsules/src/usb/cdc.rs
@@ -64,10 +64,15 @@ enum State {
     /// The host has enumerated this USB device. Things should be functional at
     /// this point.
     Enumerated,
-    /// We have seen the CDC messages that we expect to signal that a CDC client
-    /// has connected. We stay in the "connecting" state until the USB transfer
-    /// has completed.
-    Connecting,
+    /// We are seeing the CDC messages that we expect to signal that a CDC
+    /// client has connected. We want to see that both line coding and line
+    /// state ctrl messages have been received. We stay in the "connecting"
+    /// state until the USB transfers for both ctrl messages have completed.
+    Connecting { line_coding: bool, line_state: bool },
+    /// We have seen the necessary setup messages in the `Connecting` state, now
+    /// we delay just to ensure the host has enough time to receive _and
+    /// display_ messages.
+    ConnectingDelay,
     /// A CDC client is connected. We can safely send data.
     Connected,
 }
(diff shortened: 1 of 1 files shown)


capsules/src/usb/cdc.rs
```diff
@@ -416,29 +456,30 @@ impl<'a, U: hil::usb::UsbController<'a>, A: 'a + Alarm<'a>> hil::usb::Client<'a>
...
-                |line_coding| {
-                    // Check if we should switch our main state machine to
-                    // connecting meaning that the host is connecting to the virtual
-                    // serial port. We decide this based on if the host is
-                    // configuring the baud rate to what we expect.
-                    if self.state.get() == State::Enumerated && line_coding.baud_rate == 115200 {
-                        self.state.set(State::Connecting);
-                    }
+        match self.ctrl_state.get() {
+            CtrlState::SetLineCoding => {
```
alistair23, line 460: Why change this from an `if`?

bradjc: I guess that is left over from some things I tried that didn't end up being necessary. But if someone does want to handle data from a different ctrl message they likely want to make this change.

hudson-ayers (approved):

lschuermann (approved):

ppannuto (approved):

alevy:
Not to block on, but: I don't love about this that the 100ms delay "hack" is really for the console driver (when backed by USB) but is applied to all CDC instances (i.e. implemented at the CDC driver level). More ideal would be to apply the delay at _either_ the UART trait layer _or_ in the console driver (e.g., if there was a console driver specific for USB)

alevy (approved):
Approve with reservations in my comment above: a comment, maybe at the very top, specifying that this is a hack would be better. But also... whatever :)

PR #2732, opened 2021-08-05
capsules: hts221: fix i2c read command

bradjc:
### Pull Request Overview

I do not know why it was a read_write(1, 16).

Also format the top comments, and remove a debug that can never happen. (sorry about the diff)

### Testing Strategy

This pull request was tested by using the sensors app on nano33ble board.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/hts221.rs
+++ b/capsules/src/hts221.rs
@@ -1,5 +1,5 @@
-//! SyscallDriver for the STMicro HTS221 relative humidity and temperature sensor
-//! using the I2C bus.
+//! SyscallDriver for the STMicro HTS221 relative humidity and temperature
+//! sensor using the I2C bus.
 //!
 //! <https://www.st.com/en/mems-and-sensors/hts221.html>
 //!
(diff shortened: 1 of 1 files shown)


capsules/src/hts221.rs
```diff
@@ -130,7 +134,7 @@ impl<'a> Hts221<'a> {
                         buffer[1] = 1 << 2 | 1 << 7; // BDU + PD
                         buffer[2] = 1; // ONE SHOT
 
-                        if let Err((_error, buffer)) = self.i2c.write_read(buffer, 1, 16) {
+                        if let Err((_error, buffer)) = self.i2c.write(buffer, 3) {
```
bradjc, line 137: This is the actual fix.

capsules/src/hts221.rs
```diff
@@ -202,175 +206,166 @@ impl<'a> I2CClient for Hts221<'a> {
             self.humidity_client.map(|client| client.callback(0));
             return;
         }
-        match status {
-            Ok(()) => {
```
bradjc, line 206: I don't know why bother with this since status.is_err() is already checked above.

hudson-ayers: Really it would be most idiomatic to only match on status once, rather than doing is `status.iserr() { ... return } ... whatever else`

I am guessing the reason it is written this way is because it was originally written using `ReturnCode`.

I don't think it matters that much though

hudson-ayers (approved):
I don't know why it was a `write_read(1,16)` but this looks fine to me.

ppannuto (approved):
bors r+

PR #2733, opened 2021-08-05
kernel: sched: priority: get correct proc index

bradjc:
### Pull Request Overview

If using the priority scheduler with a padding app first in the app section followed by an actual app, the actual app would never run. This is because the priority scheduler calls `kernel.get_process_iter()` which filters for only valid processes, then uses the index in _that_ iter as the process index. The first entry in the processes array is `None` (because of the padding). So the actual app is detected at index 0, but when compared to its actual index (1) it looks like there is a higher priority app ready. But it's the same app.

Switch to `iter.find()` and then get the actual processes array index.

### Testing Strategy

Using a padding app on arty-e21.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/scheduler/priority.rs
+++ b/kernel/src/scheduler/priority.rs
@@ -62,10 +62,11 @@ impl<C: Chip> Scheduler<C> for PrioritySched {
             || self
                 .kernel
                 .get_process_iter()
-                .position(|proc| proc.ready())
-                .map_or(false, |ready_idx| {
-                    self.running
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):
Wow, great catch

alistair23 (approved):

ppannuto (approved):
bors r+

PR #2742, opened 2021-08-08
[arm] Cortex M0+ region

alexandruradovici:
### Pull Request Overview

This pull request adds the ability to set the Cortex-M MPU minimum region size.

According to the [Cortex M0+ MPU documentation](https://developer.arm.com/documentation/dui0662/b/Cortex-M0--Peripherals/Memory-Protection-Unit/MPU-Region-Attribute-and-Size-Register?lang=en) the minimum region size is 256B as opposed to Cortex-M3, M4 and M7 where the minimum size is 32B.

### Testing Strategy

Still testing

### TODO or Help Wanted

I'm not sure about the region alignment requirements, I think the minimum alignment should be 256 for Cortex M0+, as regions have to be aligned to the base address.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/mpu.rs
+++ b/arch/cortex-m/src/mpu.rs
@@ -1,4 +1,4 @@
-//! Implementation of the memory protection unit for the Cortex-M3,
+//! Implementation of the memory protection unit for the Cortex-M0+, Cortex-M3,
 //! Cortex-M4, and Cortex-M7
 
 use core::cell::Cell;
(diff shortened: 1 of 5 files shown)


arch/cortex-m/src/mpu.rs
```diff
@@ -129,7 +129,7 @@ const MPU_BASE_ADDRESS: StaticRef<MpuRegisters> =
 ///
 /// There should only be one instantiation of this object as it represents
 /// real hardware.
-pub struct MPU<const NUM_REGIONS: usize> {
+pub struct MPU<const NUM_REGIONS: usize, const MIN_SIZE: u32> {
```
bradjc, line 132: Why have `MIN_SIZE` be in number of mask bits rather than just bytes?

alexandruradovici: I thought of placing the actual value first, than I realized that this might be wrongly used. As such, I though using the bits would reduce the error space. I can change it back to use the size.

alexandruradovici: I updated the constant to represent the size of the region

bradjc (approved):

hudson-ayers (approved):

ppannuto (approved):
bors r+

PR #2748, opened 2021-08-10
fix broken error handling in allow for udp

hudson-ayers:
### Pull Request Overview

This pull request fixes incorrect error handling in the body of `allow_readwrite()` for the UDP driver. This mistake seems quite easy to make as a result of having two implementations of `From` for `Error` -> `ErrorCode` and `Error` -> `Result<(), ErrorCode>`. I plan to submit a followup PR that checks other drivers for this issue and restructures the interface somehow to make mistakes like this one more difficult, but this is a band-aid to allow us to finish release testing.

### Testing Strategy

This pull request was tested by running the UDP virtualization test apps that force allow failures and confirm they fail.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/capsules/src/net/udp/driver.rs
+++ b/capsules/src/net/udp/driver.rs
@@ -358,10 +358,10 @@ impl<'a> SyscallDriver for UDPDriver<'a> {
             _ => Err(ErrorCode::NOSUPPORT),
         };
 
-        if let Err(e) = res {
-            Err((slice, e))
-        } else {
-            Ok(slice)
+        match res {
+            Err(e) => Err((slice, e)),
+            Ok(Err(e)) => Err((slice, e)),
+            Ok(Ok(())) => Ok(slice),
         }
     }
 


capsules/src/net/udp/driver.rs
```diff
@@ -358,10 +358,10 @@ impl<'a> SyscallDriver for UDPDriver<'a> {
             _ => Err(ErrorCode::NOSUPPORT),
         };
 
-        if let Err(e) = res {
-            Err((slice, e))
-        } else {
-            Ok(slice)
+        match res {
+            Err(e) => Err((slice, e)),
```
brghena, line 362: Why don't you need to do `e.into()` here? Is it implicitly casting from `Error` to `ErrorCode`?

brghena (approved):

capsules/src/net/udp/driver.rs
hudson-ayers: In this case `e` is already an `ErrorCode`. The type of `res` is `Result<Result<(), ErrorCode>, ErrorCode>`. The `ErrorCode` in the `Err()` case can be traced as follows:

`self.apps.enter` returns a `Result<Result<(), ErrorCode>, process::Error>`

then,the call to `map_err()` converts that to `Result<Result(), ErrorCode>, ErrorCode>` by only transforming the type of the `Err` case for the outer `Result`

lschuermann (approved):

bradjc (approved):

ppannuto (approved):
bors r+

This interface definitely has some issues, but this is good for an immediate bugfix.

PR #2752, opened 2021-08-10
kernel: load_processes: do not skip indices for padding apps

bradjc:
### Pull Request Overview

Before padding "apps" would effectively use up slots in the processes array because those indices would be skipped over. So if a board had many padding apps then it could exhaust its processes array despite not actually loading that many processes. This only uses up a processes array slot if an actual process was created.

This PR is related to an optimization in tockloader. Rather than uninstall apps by effectively reading all apps off the board, re-arranging them, and then re-flashing the apps to keep, uninstalling an app is now just replacing its header with a padding app header of the same size.

### Testing Strategy

This pull request was tested by running the kernel on nrf52840dk with padding apps.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process_utilities.rs
+++ b/kernel/src/process_utilities.rs
@@ -176,7 +176,8 @@ pub fn load_processes_advanced<C: Chip>(
     let mut remaining_memory = app_memory;
 
     // Try to discover up to `procs.len()` processes in flash.
-    for i in 0..procs.len() {
+    let mut index = 0;
+    while index < procs.len() {
         // Get the first eight bytes of flash to check if there is another
         // app.
         let test_header_slice = match remaining_flash.get(0..8) {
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):

ppannuto (approved):
I wonder if some day we will want to treat disabled processes different from padding here? Reserving the slot for them in case something turns them on later. But that feels like a solution for the later day when we do something with disabled processes :D

bradjc:
> I wonder if some day we will want to treat disabled processes different from padding here? Reserving the slot for them in case something turns them on later. But that feels like a solution for the later day when we do something with disabled processes.

There is no requirement that the order in PROCESSES has to match the order in flash. They can easily be added later in the array.

lschuermann (approved):

PR #2755, opened 2021-08-11
Add a tool for finding panics in a Tock .elf file

hudson-ayers:
### Pull Request Overview

This pull request adds a tool for finding locations in Tock source code that can panic in a Tock kernel .elf file. It takes in a path to an elf file and uses the objdump and dwarfdump on your system to analyze the passed elf file.

The tool largely operates by tracing backwards from calls in the core library that call `panic_fmt`.
It outputs a list of source locations. Not all of these source locations are precise locations of calls that can lead to panics, due to limitations in the DWARF debug information available in a binary. That said it is usually possible to find the relevant panic with the information given and a little poking around in the file it outputs.

The python code is definitely still pretty hacky -- I ported it from a convoluted shell script I was using originally, and am trying to focus my efforts elsewhere for now since this is at least usable (if not particularly maintainable) in its current state. I am happy to accept refactoring though!

Notably, this tool focuses on identifying only source locations in *Tock* source code, basically by identifying functions in the core library that can call panics and flagging all calls to those functions from within the Tock source.

Example output of the tool run on Imix can be found here: https://gist.github.com/hudson-ayers/382c90973eaa786bd2b4fb2db7a1911d . 378 panic locations are identified, though many are ultimately duplicates in monomoprhized code (e.g. a single panic in `Grant::enter()` can show up N times where N is the number of unique calls to enter which pass different functions).

One important thing to note is that this tool only finds panics which actually exist in the binary -- this means that in practice many calls which look like they could panic will be excluded, if the calls are inlined and LLVM is able to infer that a panic is impossible. This means that removing a given panic may not always change the number of panics identified by the tool. For example:

```rust
fn access_buf(buf: &[u8]) -> bool {
    if buf[1] == 0 {
        return true;
    }
    if buf[0] == 1 {
        return true;
    }
    false
}
```
Removing the first buffer access in this function will not change the total number of panics, because originally LLVM will be smart enough to know the second access cannot panic, but removing the first access will mean that LLVM has to check the second access.

### Testing Strategy

This pull request was tested by running this tool on upstream Imix, Hail, and OpenTitan. It was also tested on an out-of-tree risc-v platform. I manually verified a couple dozen of the panics on the Imix binary and was able to trace most of them to the actual panics in the source code.

### TODO or Help Wanted

I formatted this using black, do we have any other requirements for python code in the Tock repo?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/tools/find_panics.py
+++ b/tools/find_panics.py
@@ -47,13 +47,19 @@
 #
 # Author: Hudson Ayers <hayers@.stanford.edu>
 
-import subprocess
-import re
-import sys
 import argparse
+import platform
+import re
+import subprocess
+import sys
 
 
-DWARFDUMP = "llvm-dwarfdump"
+if platform.system() == 'Darwin':
+    DWARFDUMP = "dwarfdump"
+elif platform.system() == 'Linux':
+    DWARFDUMP = "llvm-dwarfdump"
+else:
+    raise NotImplementedError("Unknown platform")
 # Note: In practice, GCC objdumps are better at symbol resolution than LLVM objdump
 ARM_OBJDUMP = "arm-none-eabi-objdump"
 RISCV_OBJDUMP = "riscv64-unknown-elf-objdump"
--- a/tools/find_panics.py
+++ b/tools/find_panics.py
@@ -1,0 +1,458 @@
+#!/usr/bin/env python3
+
+# Prints out the source locations of panics in a Tock kernel ELF
+#
+# This tool attempts to trace all panic locations in a Tock kernel ELF by
+# tracing calls to panic functions in the core library, using the debug information
+# embedded in the ELF file. This tool requires an ELF which includes debug information.
+# In its current state, cannot accurately provide the source locations
+# corresponding to each panic, but tries to be honest about its confidence in
+# each guess. In general, each guess is usually enough to locate the relevant panic.
+# More creative analysis might be able to increase
+# the accuracy with which this tool can identify source locations of panics. For now,
+# this tool is useful for:
+#
+# - obtaining a rough count of the number of panics in a Tock kernel binary
+#
+# - finding and removing panics in a Tock kernel binary
+#
+# - roughly determining which components of a Tock kernel binary contain the most panic
+# paths
+#
+# There are several assumptions built into this tool which may not always hold. For one,
+# the list of panic_functions are assumed to not match any strings in the actual
+# codebase, despite the fact they are incomplete function names and overlap is possible.
+# I could solve this by using full names of these functions, but I am unsure how often
+# the name mangling of these functions will change as the rust compiler changes so this
+# approach felt potentially more stable.
+#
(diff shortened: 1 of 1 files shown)


hudson-ayers:
Oh, this tool also requires building with -Z build-std=core. I realize those changes will need to be submitted and considered separately, so marking as blocked for now.

tools/find_panics.py
```diff
@@ -0,0 +1,453 @@
...
+
+            source_lines = list(map(getLine, line_matches))
+            source_line = source_lines[::-1][i]
+            return (f, source_line)
+    return ("", "")
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("elf", help="ELF file for analysis")
```
phil-levis, line 189: Usually variable arguments (rather than specific options) are all caps, so ELF. E.g.:

`Usage: tar [OPTION...] [FILE]...`

So think this should print out as

`usage: find_panics.py [-h] [--verbose] [--riscv] ELF`

phil-levis (commented):
This looks great, but I can't seem to run it:

```pal@ubuntu:pal/src/tock $ tools/find_panics.py target/thumbv7em-none-eabi/release/imix.elf 
Tock panic report for target/thumbv7em-none-eabi/release/imix.elf
Traceback (most recent call last):
  File "tools/find_panics.py", line 453, in <module>
    main()
  File "tools/find_panics.py", line 427, in main
    objdump, args.elf, args.riscv
  File "tools/find_panics.py", line 205, in find_all_panics
    result = subprocess.run((objdump, "-d", elf), capture_output=True, text=True)
  File "/usr/lib/python3.6/subprocess.py", line 423, in run
    with Popen(*popenargs, **kwargs) as process:
TypeError: __init__() got an unexpected keyword argument 'capture_output'
```

So it seems like there's a versioning issue with libraries?

tools/find_panics.py
hudson-ayers: Thanks, fixed!

hudson-ayers:
> This looks great, but I can't seem to run it:
> 
> ```
> Tock panic report for target/thumbv7em-none-eabi/release/imix.elf
> Traceback (most recent call last):
>   File "tools/find_panics.py", line 453, in <module>
>     main()
>   File "tools/find_panics.py", line 427, in main
>     objdump, args.elf, args.riscv
>   File "tools/find_panics.py", line 205, in find_all_panics
>     result = subprocess.run((objdump, "-d", elf), capture_output=True, text=True)
>   File "/usr/lib/python3.6/subprocess.py", line 423, in run
>     with Popen(*popenargs, **kwargs) as process:
> TypeError: __init__() got an unexpected keyword argument 'capture_output'
> ```
> 
> So it seems like there's a versioning issue with libraries?

I use some features which require python 3.7+, so I added a check that prints an error if a user uses an older python version than that.

I have also rebased.

bradjc (approved):

PR #2756, opened 2021-08-11
TRD for UART HIL

phil-levis:
### Pull Request Overview

This pull request contains a first draft of a UART TRD. It changes the UART HIL slightly, breaking apart `Configure` and `Configuration`.

### Testing Strategy

N/A

### TODO or Help Wanted

There are two major open questions on this TRD:
  1. For SPI, the semantics of a `read_write_bytes` operation are that the actual length of the operation is the minimum of the requested length, the length of the write buffer, and the length of the read buffer (if there is one). For UART, the semantics of `receive_buffer` and `transmit_buffer` are that if the requested length is longer than the size of the buffer, return `Err(SIZE)`. Should we make this consistent? If so, then I think the right thing to do is have the SPI return `Err(SIZE)` if an operation longer than one of the buffers is requested.
  2. `ReceiveAdvanced` has always been a bit of a wart. It's a necessary API for some use cases, but  it's only implemented on the SAM4L and seems to have no users in the tree. Should we deprecate it? We can always bring it back in a separate TRD if needed. If we choose to keep it around, I'd suggest we call it `ReceiveTimeout` or `ReceiveWithTimeout`.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


brghena:
We could split `ReceiveAdvanced` into a different HIL altogether. It's unlikely to ever be finalized.

bradjc:
`ReceiveAdvanced` is currently used by the bootloader.

What I like more about `ReceiveAdvanced` is that it serves as an example of how to handle unique hardware features with Tock HILs and capsules. I'm not really partial to any particular approach on what to do with the `receive_automatic()` function, but I do think it serves as a good example we can point developers to if their chip has a feature they want to use in a capsule, but the current HIL doesn't expose the relevant functionality.

phil-levis:
I do think it's weird for `ReceiveAdvanced` to be a HIL, but to be implemented only by SAM4L and its emulation for other platforms is not in the main repository. So it's not something that a capsule can depend on.

But very specifically, if it's a piece of chip specific functionality, then I'd argue it should be a chip API, not a generic API sitting in `kernel::hil`. If it's a generic API in `kernel::hil`, then there should be a common implementation of it. If this couldn't be emulated in software (e.g., with an alarm), this I'd argue it should definitely not be a HIL. It can, though, which makes it OK.

My discomfort is that it's in this weird middle ground: it's in `hil::uart`, but only one chip implements it, and nothing in the main repository uses it. So it's in the HIL, but you can't really use it as is (except for a SAM4L platform). For the SAM4L, at least, the two have different calling semantics despite a bunch of repeated code:
  - `receive` returns an error if `len` is too long; `receive_automatic` trims the read;
  - `receive_automatic` returns an error if a read is outstanding, `receive` starts another transfer;
  - `receive_automatic` returns `Ok(())` if the DMA isn't available, `receive` returns an error.

It seems like the major reason we have it in `kernel::hil::uart` is so that the SAM4L can provide it natively. But we could also achieve this by having a `set_timeout` method on the SAM4L USART, putting this trait into the bootloader repository, and have the SAM4L implementation of the trait call `set_timeout`. Or just have a method on the USART which does it and have the bootloader codebase put the trait on top of it.

So my preference is that we either 
  1. fully support it, and move emulation into the main repository, so it's an API that everything can use,
  2. or we move the trait into the bootloader repository and tweak the SAM4L implementation so the bootloader can easily implement it.

phil-levis:
The consensus on the call today seemed to  be (correct me if I'm wrong):
  - Providing software emulation layers for `ReceiveAdvanced` in the main tree as a standard solution isn't desirable, in part because there aren't users and also because they are heavyweight
  - This means that `ReceiveAdvanced` should transition out of the UART HIL and be something which the bootloader defines/implements

This implies that the sam4l UART implementation should change a little bit to support the bootloader being able to call its hardware timeout receive but not through a trait.

There was agreement that showing a way that hardware can have "extended" APIs is valuable, at the very least as a guide/template for implementers to follow. We should keep this in mind as we continue to polish and define the HILs and try to come up with a canonical example.

phil-levis:
I've updated the TRD following the call today. There's one detail we need to iron out: can `receive_abort` (or `transmit_abort`) cause a callback to execute synchronously?

Rule 1 of the HIL design TRD says that you shouldn't execute synchronous callbacks. But this case is a little different: the callback is in response to `receive`: the abort call simply can cause it to occur earlier. It's worth noting that the SAM4L implementation does issue them synchronously, while the nrf52 doesn't (it uses a STOPRX event to trigger an interrupt).

Currently this TRD says that abort MAY trigger a synchronous `received` interrupt. If we decide that this falls under rule 1 of the HIL design TRD then we should update the SAM4L implementation and go through our other existing implementations.

lschuermann:
> There's one detail we need to iron out: can `receive_abort` (or `transmit_abort`) cause a callback to execute synchronously?

My worry is that implementations, while they can certainly handle these cases, probably wouldn't. Thinking of a driver which performs a receive application for userspace which is then aborted, a naive driver implementation would store the process handle which initated this operation and, in the `received_*` callbacks, immediately try to enter this process' Grant region.

The fact that generally we enforce callbacks to be asynchronous -- in my opinion -- makes it even more confusing that a select number of callbacks are instead allowed to be synchronous. I don't have a particularly strong opinion about this but have a slight preference to support the naive implementation use-case and handle the additional complexity for always making an asynchronous callback in the driver, instead of putting the burden on the consumer of the interface.

phil-levis:
> The fact that generally we enforce callbacks to be asynchronous -- in my opinion -- makes it even more confusing that a select number of callbacks are instead allowed to be synchronous.

I agree. But I think what this is getting at is very precisely defining what we mean by a synchronous callback. I don't think that we want to say that you only call a callback from within a callback rooted in an interrupt handler. We have no way of effectively maintaining that invariant, given that a callback is just a calling convention. But I think we could say that a synchronous callback means "invoking a callback from a method on the trait in which you set the callback." E.g., a `transmitted_buffer` callback is synchronous if it's invoked from the trait that has the `set_client` method for configuring that callback. Otherwise it's going to be extremely difficult to have any kind of callbacks for publish/subscribe abstractions.

bradjc:
It seems like somehow the appid trd was added to this PR.

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,523 @@
...
+
+
+The UART may be unable to set the precise baud rate specified. For
+example, the UART may be driven off a fixed clock with integer
+prescalar. A call to `configure` MUST set the baud rate to the closest
+possible value to the `baud_rate` field of the `params` argument and a
+call to `set_baud_rate` MUST set the baud rate to the closest possible
+value to the `rate` argument. The `Ok` result of `set_baud_rate`
+includes the actual rate set, while an `Err(INVAL)` result means the
+requested rate is well outside the operating speed of the UART (e.g., 4MHz).
```
ppannuto, line 153:
```suggestion
prescalar. In response to a call to `configure`, the underlying UART MUST set the baud rate to the closest
possible value to that requested by the `baud_rate` field of the `params` argument, and the response to a
call to `set_baud_rate` MUST set the baud rate to the closest possible
value to the provided `rate` argument. The `Ok` result of `set_baud_rate`
includes the actual rate set, while an `Err(INVAL)` result means the
requested rate is well outside the operating speed of the UART (e.g., 4MHz).
```

I eventually got it, but I had to read this many times as my first parse kept interpreting this as the _caller_ need to set the baud rate correctly when calling; this is a bit more verbose but hopefully clearer to other thick-headed readers such as myself 😅

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,523 @@
...
+pub trait Transmit<'a> {
+    fn set_transmit_client(&self, client: &'a dyn TransmitClient);
+
+    fn transmit_buffer(
+        &self,
+        tx_buffer: &'static mut [u8],
+        tx_len: usize,
+    ) -> Result<(), (ErrorCode, &'static mut [u8])>;
+
+    fn transmit_word(&self, word: u32) -> Result<(), ErrorCode>;
```
ppannuto, line 177: Should this 'no-buffer' path be `word` or `byte`? I can't think of a ton of scenarios where a caller wants to send exactly 4 bytes over UART?

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,523 @@
...
+busy.
+
+Each byte transmitted is a data word for the UART. If the UART is
+using 8-bit data words, each data word is a byte. If the UART is using
+smaller data words, it MUST ignore the high order bits of the data
+values. For example, if the UART is using 6-bit data words and is told
+to transmit `0xff`, it will transmit `0x3f`, ignoring the first two
+bits. If a client needs to transmit data words larger than 8 bits, it
+should use `transmit_word`, as `transmit_buffer` is a buffer of 8-bit
+bytes and cannot store 9-bit values.
```
ppannuto, line 209: Is this case even possible? The `Width` enum above only goes to 8 bits. I suspect 9-bit UART to be such a rarity in modern hardware perhaps that we should omit it at this stage, and then re-evaluate later what the best approach is (i.e. one would still hope to use DMA; perhaps the engine expects 16-bit halfword streams of 9-bit data? I think the 9-bit interface is a question best answered by hardware that uses it)

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,523 @@
...
+===============================
+
+The `Transmit` and `TransmitClient` traits allow a client to transmit
+bytes over the UART.
+
+```rust
+enum AbortResult {
+    Failure,
+    Success,
+}
```
ppannuto, line 166: What is the purpose of including this enum instead of of just returning `Ok` or `Err(FAIL)`?

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,523 @@
...
+returned by `transmit_buffer` indicates whether there will be a
+callback in the future. If `transmit_buffer` returns `Ok(())`,
+implementation MUST call the `TransmitClient::transmitted_buffer`
+callback in the future when the transmission completes or fails. If
+`transmit_buffer` returns `Err` it MUST NOT issue a callback in the
+future in response to this call. If the error is `BUSY`, this is
+because there is an outstanding call to `transmit_buffer` or
+`transmit_word`: the implementation handles these calls normally and
+issues a callback for them. However, it does not issue a callback for
+the call to `transmit_buffer` that returned `Err`.
```
ppannuto, line 229:
```suggestion
because there is an outstanding call to `transmit_buffer` or
`transmit_word`: the implementation will continue to handle
the original call and issue the originally scheduled callback
(as if the call that `Err`'d with `BUSY` never happened). However,
 it does not issue a callback for the call to `transmit_buffer`
  that returned `Err`.
```

I think this is what this is saying?

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,523 @@
...
+
+The valid error codes for `transmit_buffer` are:
+  - `OFF`: the underlying hardware is not available, perhaps because it has
+    not been initialized or has been initialized into a different mode
+    (e.g., a USART has been configured to be a SPI).
+  - `BUSY`: the UART is already transmitting and has not made a transmission
+    callback yet.
```
ppannuto, line 236:
```suggestion
    complete callback yet.
```

ppannuto (commented):
Partial review, will try to read the rest later

doc/reference/trd-uart.md
phil-levis: Yeah, I get the confusion. But "in response" seems to suggest asynchrony. How about "Calling `configure` MUST ..."

phil-levis: Yeah, I get the confusion. But "in response" seems to suggest asynchrony. How about "Calling `configure` MUST ..."

doc/reference/trd-uart.md
phil-levis: It's a word in case the UART is 9-bit data.

doc/reference/trd-uart.md
phil-levis: The `Ok` vs. `Err` is about whether there will be a callback. If the return type is `Ok`, there will be a callback. If it's `Err`, there won't be (e.g., because no operation was oustanding). The `AbortResult` type is about whether the operation is actually aborted. So:
  - `Ok(AbortResult::Failure)` means there will be a callback and the operation wasn't aborted
  - `Ok(AbortResult::Success)` means there will be a callback and the operation was aborted
  - `Err(())` means there will not be a callback.

This trinary result really doesn't fit that well with `Result`, but I figured it was better to keep with linguistic support than roll a new type.

doc/reference/trd-uart.md
phil-levis: https://community.silabs.com/s/article/uart-using-9-bit-mode-of-the-uart?language=en_US

There can be all kinds of weird hardware you want to connect to: I'd hate for someone who has to talk 9N1 (on a platform that can) having to start hacking peripheral drivers to do so... The fact that it's only word-level means that there isn't great support, but at least there's *something*.

`Width` should probably include 9?

doc/reference/trd-uart.md
hudson-ayers: Given that the Error in this case is just the unit type, this seems better suited to return `Option<AbortResult>`, or to just a custom type like you said. The `?` operator makes it easy to bubble errors up the call stack, but in this case we don't want callers to bubble up the error immediately, because one level higher than the caller it will not be clear that the `Err` does not indicate an actual error, but merely that there will not be a callback.

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,527 @@
...
+and also improves transfer speeds because hardware can keep the UART
+busy.
+
+Each byte transmitted is a data word for the UART. If the UART is
+using 8-bit data words, each data word is a byte. If the UART is using
+smaller data words, it MUST ignore the high order bits of the data
+values. For example, if the UART is using 6-bit data words and is told
+to transmit `0xff`, it will transmit `0x3f`, ignoring the first two
+bits. If a client needs to transmit data words larger than 8 bits, it
+should use `transmit_word`, as `transmit_buffer` is a buffer of 8-bit
```
ppannuto, line 210: Do we want to limit `9N` people to no DMA? Maybe

> If a client needs to transmit characters larger than 8 bits, the byte array will be read as MSB halfwords. The unused upper bits MUST be set to `0`, that is for 9-bit characters: `buffer[0] ∈ [0,1]` and `buffer[1] ∈ [0-255]`.

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,527 @@
...
+time. Successfully calling either `transmit_buffer` or
+`transmit_word` causes the UART to become busy until it issues the
+callback corresponding to the outstanding operation.
+
+3.1 `transmit_buffer` and `transmitted_buffer`
+===============================
+
+`Transmit::transmit_buffer` sends a buffer of data. The result
+returned by `transmit_buffer` indicates whether there will be a
+callback in the future. If `transmit_buffer` returns `Ok(())`,
```
ppannuto, line 223:
```suggestion
callback in the future. If `transmit_buffer` returns `Ok(())`, an
```

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,527 @@
...
+
+Most devices using serial ports today use 8-bit data, but some older
+devices use more or fewer bits, and hardware implementations support
+this. If the data width of a UART is set to less than 8 bits, data is
+still partitioned into bytes, and the UART sends the least significant
+bits of each byte. Suppose a UART is configured to send 7-bit
+words. If a client sends 5 bytes, the UART will send 35 bits,
+transmitting the bottom 7 bits of each byte. The most significant bit
+of each byte is ignored.
+
```
ppannuto, line 91:
```suggestion

To avoid ambiguity with in-memory bytes, we will refer to the 5- to 9-bit entities
sent on the wire as `characters`.
```

This change will require changing "data word" out in many places in this document,
but would bring the document in line with the terminology used in the Wikipedia article
linked at the beginning of the text as well as the datasheets I skimmed.

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,527 @@
...
+was successfully transmitted.  The `tx_len` argument specifies how
+many data words (defined by `Configure`) were transmitted. If the
+`rval` of `transmitted_buffer` is `Ok(())`, `tx_len` MUST be equal to
+the size of the transmission started by `transmit_buffer`, defined
+above.  A call to `transmit_word` or `transmit_buffer` made within
+this callback MUST NOT return `Err(BUSY)` unless it is because this is
+not the first call to one of these methods in the callback.  When this
+callback is made, the UART MUST be ready to receive another call. The
+valid `ErrorCode` values for `transmitted_buffer` are all of those
+returned by `transmit_buffer` plus:
```
ppannuto, line 258: Do all of those errors make sense for the already-started callback case?
 - `OFF`: I can see this if some higher priority task pre-empted your HW access
 - `BUSY`: Same priority thought?
 - `SIZE`: I don't think this error should be possible given the narrow definition above (the slice length is known at call-time)
 - `FAIL`: Sure.

So maybe "except `SIZE`"?

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,527 @@
...
+  - `CANCEL` if the call to `transmit_buffer` was cancelled by a call
+    to `abort` and the entire buffer was not transmitted.
+  - `SIZE` if the buffer could only be partially transmitted.
```
ppannuto, line 261: Why should this be `SIZE` instead of one of the other errors (or `FAIL`)?

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,527 @@
...
+asynchronously.  The word length is determined by the UART
+configuration.  A UART implementation MAY choose to not implement
+`transmit_word` and `transmitted_word`.  There is a default
+implementation of `transmitted_word` so clients that do not use
+`receive_word` do not have to implement a callback.
+
+If `transmit_word` returns `Ok(())`, the implementation MUST call the
+`transmitted_word` callback in the future. If a call to
+`transmit_word` returns `Err`, the implementation MUST NOT issue a
+callback for this call, although if the it is `Err(BUSY)` is will
```
ppannuto, line 276:
```suggestion
callback for this call, although if the it is `Err(BUSY)` it will
```

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,527 @@
...
+issue a callback for the outstanding operation.  Valid `ErrorCode`
+results for `transmit_word` are:
+  - `OFF`: The underlying hardware is not available, perhaps because
+    it has not been initialized or in the case of a shared
+    hardware USART controller because it is set up for SPI.
+  - `BUSY`: the UART is already transmitting and has not made a
+    transmission callback yet.
+  - `NOSUPPORT`: the implementation does not support `transmit_word`
+    operations.
```
ppannuto, line 285:
```suggestion
```

I'm not sure we should allow this. If a chip really can't do this for some reason, then asking the implementing driver to statically allocate 1 word of memory and transparently convert this to a `transmit_buffer` internally is not a huge lift. I feel like runtime feature-failures like this should be avoided if we can.

doc/reference/trd-uart.md
ppannuto: Okay, I have read all of this now, and I think I have convinced myself that making abort split-phase is more confusing than helpful. The configure trait is already all synchronous, so we have implicitly discarded the possibility of an off-chip implementation requiring everything to be split. The HIL guidelines push towards split, but acknowledge pragmatism for synchronous calls.

Here, I put myself in the userspace position—cancelling operations is hard enough, but now I have to handle two separate asynchronous callbacks (the abort response and the original xmit response) in unknown order? What other case do we have multiple intertwined callbacks in flight for one driver?

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,527 @@
...
+8-bit data words, each data word is a byte. If the UART is using
+smaller data words, it MUST zero the high order bits of the data
+values. For example, if the UART is using 6-bit data words and
+receives `0x1f`, it must store `0x1f` in a byte and not set high order
+bits.  If the UART is using 9-bit words and receives `0x1ea`, it
+stores this in a 32-bit value for `receive_word` as `0x000001ea`.
+
+`Receive` supports a single outstanding receive request. A successful
+call to `receive_buffer` or `receive_word` causes the UART to be busy
+until the callback for the outstanding operation is issued.
```
ppannuto, line 344:
```suggestion
call to `receive_buffer` or `receive_word` causes the receive half of UART
to be busy until the callback for the outstanding operation is issued.
```

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,527 @@
...
+
+In addition to the 6 basic traits, the UART HIL defines several traits
+that use these basic traits as supertraits. These composite traits allow
+structures to refer to multiple pieces of UART functionality with a
+single reference and ensure that their implementations are coupled.
+
+```rust
+pub trait Uart<'a>: Configure + Configuration + Transmit<'a> + Receive<'a> {}
+pub trait UartData<'a>: Configuration + Transmit<'a> + Receive<'a> {}
+pub trait UartAdvanced<'a>: Configure + Configuration + Transmit<'a> + ReceiveAdvanced<'a> {}
```
ppannuto, line 480:
```suggestion
```

We got rid of this, right?

doc/reference/trd-uart.md
```diff
@@ -0,0 +1,527 @@
...
+request. If multiple clients make `receive_buffer` calls that overlap
+with one another, they each receive copies of the received data.
+
+Suppose, for example, that there are two clients. One of them calls
+`receive_buffer` for 8 bytes. A user starts typing "1234567890"
+at the console. After the third byte, another client calls `receive_buffer`
+for 4 bytes. After the user types "7", the second client will
+receive a `received_buffer` callback with a buffer containing "4567".
+After the user types "8", the first client will receive a callback
+with a buffer containing "12345678".
```
ppannuto, line 515:
```suggestion
with a buffer containing "12345678". If the second client then calls
`receive_buffer` for 1 byte, they will receive "9" (note that it never
sees "8", since that has been consumed by the time this second
receive call was made).
```

I think this is true?

doc/reference/trd-uart.md
phil-levis: > I think what I am missing is why we need to generate a `transmit_abort` callback in the case of `AbortResult::Failure` -- there will still be the regular `transmitted_[word,buffer]` callback. Or perhaps equivalently, when `Err(())` would ever be generated in lieu of an `Ok(AbortResult::Failure)` (nothing to cancel?).
> 
> I'm also concerned about whether we have to enforce any ordering between the `abort` callback and the `transmitted_X` callback.
> 
> I think this is explained in more depth in the text below I didn't get to on first pass; will read that then circle back here.

There is no abort callback -- the callback it's talking about is the transmitted_{word,buffer} callback.

phil-levis: > Given that the Error in this case is just the unit type, this seems better suited to return `Option<AbortResult>`, or to just a custom type like you said. The `?` operator makes it easy to bubble errors up the call stack, but in this case we don't want callers to bubble up the error immediately, because one level higher than the caller it will not be clear that the `Err` does not indicate an actual error, but merely that there will not be a callback.

The reasoning was less about bubbling up and more about trying to make it consistent with the transmit calls (`Ok` means callback, `Err` means no callback). But I agree that the benefits of this consistency are pretty tenuous. You are unlikely to have similar state transition code structures, because the call to cancel doesn't have a buffer it has to deal with in response to `Err`. 

I prefer the `Option` approach, mostly because I think the big deal is callback or no callback, not whether the operation was cancelled (you'll find that out later). So it's nice to be able to match on `Some`/`None` in the base case. Or it could be a type like

```rust 
enum AbortResult {
  Callback(bool),
  NoCallback,
}```

File: doc/reference/trd-uart.md (hunk 13 above)
Comment (phil-levis) [reply doc/reference/trd-uart.md:285]: Take a look at how many implement it. ;)

**Comment by phil-levis** [comment]
Let's put this on the next call for a final read-over and discussion before we start updating implementations.

File: doc/reference/trd-uart.md [hunk 17]
```diff
@@ -0,0 +1,540 @@
...
+===============================
+
+In addition to the 6 basic traits, the UART HIL defines several traits
+that use these basic traits as supertraits. These composite traits allow
+structures to refer to multiple pieces of UART functionality with a
+single reference and ensure that their implementations are coupled.
+
+```rust
+pub trait Uart<'a>: Configure + Configuration + Transmit<'a> + Receive<'a> {}
+pub trait UartData<'a>: Configuration + Transmit<'a> + Receive<'a> {}
```
hudson-ayers, line 490:
```suggestion
pub trait UartData<'a>:  Transmit<'a> + Receive<'a> {}
```
I thought the point of `UartData` was this it did not require `Configuration`?

phil-levis: Good catch. Fixed.

phil-levis:
OK, now that we have implementations, I think this is safe to merge. It's still a Draft.

@hudson-ayers @ppannuto @bradjc

PR #2757, opened 2021-08-11
Rv32i/pmp: Reduce size of some inefficient constructions

hudson-ayers:
### Pull Request Overview

This pull request simplifies some match statements to move register setting outside of conditional logic where possible, and replaces series of calls to pmpxxx_modify(a) with a single call to pmpxxx_set(a). The combination of these 2 changes reduces code size of the HiFive1 board by 368 bytes, with no change in functionality.

This PR also removes a `#[repr(C)]` annotation on the CSR struct, since it should be a 0-size type now that we use const generics, and there is no reason for us to be specifying its representation anyway.

One takeaway here: LLVM cannot optimize repeated calls to read/write registers because each call is a volatile write. Therefore there are significant efficiency gains to minimizing the number of unique calls to modify a given register.

This can be merged after 2.0.

### Testing Strategy

This pull request was tested on an out-of-tree board, for which the MPU appeared to still work correctly.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alistair23 (dismissed):
The ePMP has the same commands, do you mind making the same changed there?

hudson-ayers:
Marking as blocked until I make the same changes to the ePMP

hudson-ayers:
I forgot this was blocked on me, sorry for the delay. Should be ready to go now.

bradjc (approved):

alistair23 (approved):

PR #2759, opened 2021-08-12
Add a kernel config option to elide process printing code

hudson-ayers:
### Pull Request Overview

This pull request adds a config option to disable printing exta process info in ProcessStandard in the event of a panic.
It does so by early-returning from `print_full_process()` and `print_memory_map()`. This is necessary because Rust cannot use  dead code elimination to remove uncalled functions from a binary if those functions are behind a trait object reference. Turning off this config option reduces code size by 7792 bytes on Imix.

I realize this sort of use of Config can be controversial so I am open to alternatives. But 8kB of overhead is a lot to be non-optional.

This can wait until after Tock 2.0.

### Testing Strategy

This pull request was tested by compiling with the option set to both true and false.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/config.rs
+++ b/kernel/src/config.rs
@@ -40,6 +40,27 @@ pub(crate) struct Config {
     /// into which SRAM addresses. This can be useful to debug whether the kernel could
     /// successfully load processes, and whether the allocated SRAM is as expected.
     pub(crate) debug_load_processes: bool,
+
+    /// Whether the kernel should output additional debug information on panics.
+    ///
+    /// If enabled, the kernel will include implementations of `Process::print_full_process()` and
+    /// `Process::print_memory_map()` that display the process's state in a human-readable
+    /// form.
+    // This config option is intended to allow for smaller kernel builds (in
+    // terms of code size) where printing code is removed from the kernel
+    // binary. Ideally, the compiler would automatically remove
+    // printing/debugging functions if they are never called, but due to
+    // limitations in Rust (as of Sep 2021) that does not happen if the
+    // functions are part of a trait (see
+    // https://github.com/tock/tock/issues/2594).
+    //
+    // Attempts to separate the printing/debugging code from the Process trait
+    // have only been moderately successful (see
+    // https://github.com/tock/tock/pull/2826 and
+    // https://github.com/tock/tock/pull/2759). Until a more complete solution
+    // is identified, using configuration constants is the most effective
+    // option.
+    pub(crate) debug_panics: bool,
 }
 
 /// A unique instance of `Config` where compile-time configuration options are defined. These
@@ -47,4 +68,5 @@ pub(crate) struct Config {
 pub(crate) const CONFIG: Config = Config {
     trace_syscalls: false,
     debug_load_processes: false,
+    debug_panics: true,
 };
--- a/kernel/src/process_standard.rs
+++ b/kernel/src/process_standard.rs
@@ -1067,6 +1067,9 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
     }
 
     fn print_memory_map(&self, writer: &mut dyn Write) {
+        if !config::CONFIG.debug_panics {
+            return;
+        }
         // Flash
         let flash_end = self.flash.as_ptr().wrapping_add(self.flash.len()) as usize;
         let flash_start = self.flash.as_ptr() as usize;
@@ -1226,6 +1229,9 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
     }
 
     fn print_full_process(&self, writer: &mut dyn Write) {
+        if !config::CONFIG.debug_panics {
+            return;
+        }
         self.print_memory_map(writer);
 
         self.stored_state.map(|stored_state| {
@@ -1316,6 +1322,8 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
     }
 }
 
+// Only used if debug_panics == true
+#[allow(unused)]
 fn exceeded_check(size: usize, allocated: usize) -> &'static str {
     if size > allocated {
         " EXCEEDED!"


bradjc:
Process console currently duplicates a lot of the process printing in process_standard. If we had a `ProcessPrinter` trait and single implementation that both could use, would this still be straightforward?

phil-levis:
It would only be an issue if you wanted to turn off panic output but keep the process console output; however, if you're paying a one-time cost for the code (since it's in the process console), cutting it from the panic output seems not very important.

But I think that in most in cases in which you want to cut out the panic dump for code size, you aren't running ProcessConsole.

phil-levis:
I 100% think we need to be able to do this, but I don't know if `cfg` is the right way to do it. So I'm all for solving this problem but am open to better solutions.

alistair23 (dismissed):
Yes!

jettr:
There has been discussion of other potential solutions to this problem in #2772 as well. Posting here since this was the original PR.

I would like to suggest that the solution should have the following requirements

 * Does not require downstream developers to carry a custom downstream kernel patch
 * Zero cost abstraction (after compiler optimizes out code)
 * The CI can check that any/all combinations of the extra debug flags compile

Are there any other requirements we would like in a solution in this space?

bradjc:
> Are there any other requirements we would like in a solution in this space?
 

- Matches separation of concerns in the kernel organization
- Maintainable

>     * Zero cost abstraction (after compiler optimizes out code)

I'm not sure I see this as necessary. Would a small cost be unacceptable?

hudson-ayers:
> > Are there any other requirements we would like in a solution in this space?
> 
> * Matches separation of concerns in the kernel organization

I take this to mean, more or less, "It does not make sense that process debug info tracking is part of the Process trait, so any solution to this problem should also solve that inconsistency, because that inconsistency is directly contributing to the problem as it exists"?

> * Maintainable

This solution always ensures that both options compile (the code is compiled without knowledge of the value of the CONFIG variable, and dead code is only eliminated during LTO).

> > ```
> > * Zero cost abstraction (after compiler optimizes out code)
> > ```
> 
> I'm not sure I see this as necessary. Would a small cost be unacceptable?

Yes, I think a small cost could be acceptable (say, < 5% of the overhead of having no solution). But I think that if we are choosing between a zero cost abstraction (like this one), and an abstraction with a small cost (like one that adds vtable lookups to every debug operation), there should be clear benefits to the approach with the runtime cost in terms of maintainability.

hudson-ayers:
For context: at present, the code size savings from using this change downstream are 12,928 bytes.

bradjc:
> For context: at present, the code size savings from using this change downstream are 12,928 bytes.

Is this on imix? How much savings from removing the process console? The memory map debugging code is still duplicated, and solving that duplication is an easy way to save space.

bradjc:
I think there are two intertwined issues at play here.

1. Rust/LLVM does not elide functions in traits that are never called, if that trait is used somewhere. This leads to an unnecessary increase in binary size.
2. Board authors have no way of configuring how process debugging information is provided/displayed. Some boards may not want to have process debugging info. Others may want it to be encoded in some packed binary representation and unpacked and formatted by an off-board tool. Others might want even more debugging information that we provide now.

I think that rather than just looking at the specific instance of code size for panic debugging, we should be looking at these problems more holistically. Panic print is low hanging fruit now, but if we do a quick fix for that, something else becomes the low hanging fruit, and I suspect that becomes the next #define PR (slippery slope argument, but we pretty much only add features to Tock).

bradjc:
>     1. Rust/LLVM does not elide functions in traits that are never called, if that trait is used somewhere. This leads to an unnecessary increase in binary size.

Since Tock started we have been fighting rust to support our use case (remember xargo?). Things have gotten better, and @hudson-ayers mentioned that there is support somewhere in the toolchain for this feature, so it is possible that it will be default in Rust at some point. Is there any attempt to provide a version of this somehow? It might very well use features, but instead of being related to a specific task (remove process print code for panics), it would be a generic fix-a-limitation-in-rust tool that we could use anywhere in the tock kernel.

jettr:
As other said, zero cost may be too strong of a requirement. We should aim for zero to low cost abstraction.

> Board authors have no way of configuring how process debugging information is provided/displayed. Some boards may not want to have process debugging info. Others may want it to be encoded in some packed binary representation and unpacked and formatted by an off-board tool. Others might want even more debugging information that we provide now.

I agree that is one of the prime issues.

Can we use the `KernelResources` to provide `ProcessDebugTracker` trait objects (not `dyn` though), and make `()` implement it to have our zero/low cost?

For a rough example:

```rust
enum ProcessDebugCounter {
    MissedUpcall,
    Syscall,
    TimesliceExpiration,
    ...
}

trait ProcessDebugTracker {
    fn increment_counter(&mut self, counter: ProcessDebugCounter);

    fn set_stack_ptr(&mut self, stack: *const u8);
    fn set_heap_ptr(&mut self, stack: *const u8);
    ...

    fn display_memory_map(&self, f :&mut core::fmt::Formatter) ->  core::fmt::Result;
    fn display_all(&self, f :&mut core::fmt::Formatter) ->  core::fmt::Result;
}

pub trait KernelResources<C: Chip> {
    ...
    type ProcessDebugTracker: process::ProcessDebugTracker;
    fn new_process_debug_tracker(&self) -> Self::ProcessDebugTracker;
}
```

Then we  can make `ProcessStandard` generic on `ProcessDebugTracker` and get a new tracker for each process instance. We would still hold the object in a `MapCell` to access the object in a mutable way (similar to today's `ProcessStandardDebug`)

hudson-ayers:
> Is this on imix? 

This is on an out-of-tree risc-v platform. On Imix, the savings are 8,384 bytes.

> How much savings from removing the process console? The memory map debugging code is still duplicated, and solving that duplication is an easy way to save space.

Removing the process console but leaving process panic info in panics saves 11,088 bytes on Imix. Removing process info in panics *after* removing process console saves an _additional_ 10,736 bytes, so I think most of the discrepancy between Imix and our out-of-tree board is that keeping process console retains some of the strings which are exactly duplicated between the panic print and the process console print. 

While I agree for upstream Imix specifically that dedpulicating process console information would be a big savings, we use a custom pared-down process console implementation out-of-tree which does not provide memory map debugging code anyway.

In fact, we use a custom panic handler that does not call these functions on the process struct at all, which is why it is so frustrating to pay an almost 13kB tax for them in order to use the Tock kernel without carrying downstream patches.

> I think there are two intertwined issues at play here.
> 
> 1. Rust/LLVM does not elide functions in traits that are never called, if that trait is used somewhere. This leads to an unnecessary increase in binary size.
> 2. Board authors have no way of configuring how process debugging information is provided/displayed. Some boards may not want to have process debugging info. Others may want it to be encoded in some packed binary representation and unpacked and formatted by an off-board tool. Others might want even more debugging information that we provide now.
> 
> I think that rather than just looking at the specific instance of code size for panic debugging, we should be looking at these problems more holistically. Panic print is low hanging fruit now, but if we do a quick fix for that, something else becomes the low hanging fruit, and I suspect that becomes the next #define PR (slippery slope argument, but we pretty much only add features to Tock).

Yeah, I am sympathetic to this. It is clearly the case that having some low-cost mechanism that allows individual boards to fine-tune the debug info they use would be desirable. The way we handle that sort of thing in Tock is generally by having traits, and letting the board choose between different implementations of those traits, or even provide its own implementation of those traits if needed. For example, boards can opt-out of the full set of default peripherals available by implementing the InterruptService trait for a struct that contains only a subset of the available chip peripherals. 

For `Process` and debug info, things get a little trickier. Looking at functions like `print_memory_map()`, they rely extensively on directly accessing private fields of the `ProcessStandard` struct, and modifying `print_memory_map()` to be implemented on a different struct that only uses the interface exposed by the `Process` trait would probably have a significant cost to boards that *do* want that functionality, and would require us to add lots of stuff to the `Process` struct that is really just about making information accessible that is needed for debug prints.

Alternatively, we could have two separate traits like in your proposed PR, both implemented by the `ProcessStandard` struct, but switching between the different provided traits would still require use of `#[cfg]`, and does not allow board crates to re-implement the trait themselves without redefining their own Process type entirely.

Finally, a suggestion like Jett's might work, but would probably require `ProcessDebugTracker` to store or take a reference to a `ProcessStandard` (rather than `dyn Process`) to not have significant overhead. This still might be fine, and I will look into a solution along those lines. That said, I remain unconvinced that in practice the tradeoffs of that approach are substantively different than the Config based approach in this PR. In both approaches, implementations of `ProcessDebugTracker` can only exist in the kernel crate, unless we want to make far more fields of `ProcessStandard` be public. In both approaches, all implementations are compiled. In both approaches, multiple levels of debugging are possible (for the config we could replace the bool with an enum, and each debug function could dispatch to one of several possible implementations based on the value of the config enum). The advantage of not using a trait is that we do not confine each implementation to a particular interface, and then debug methods which do not depend on the internals of ProcessStandard could then be usable by other implementations of Process. For example:
```rust
impl Process for ProcessStandard {
    fn print_memory_map(&self) {
        match CONFIG::DEBUG_LEVEL {
            DEBUG_ALL => print_all_process_info(self),
            DEBUG_MPU_ONLY => print_mpu_info_only(self.chip),
            NO_DEBUG => {},
        }
    }
}
//...perhaps in a different file or module
fn print_all_process_info(proc: &ProcessStandard) {
    //.. insert current contents of print_memory_map here
}
//...
fn print_mpu_info_only<C: Chip>(chip: &C) {
    //...only print mpu info and not everything else. advantage here is that this function could be reused in an implementation of Process other than ProcessStandard
}
```

My point here is that I am not convinced that using traits is fundamentally different than using the Config struct. `#define` is very bad, but the number one reason it is very bad is because when you use it only one option is compiled. If we confine our use of `#[cfg]` exclusively to the `Config` struct to set the values of constants, this is not an issue -- all code is compiled as if all values of the constants are possible. And using `Config` gives a lot of flexibility that cannot easily be replicated using traits. For runtime testing, you run into the same conundrum either way -- at runtime only one configuration will be in use, and thus only that configuration will be tested. At least if all of the different configurations are enumerated in a single place (kernel/src/config) it is clear what matrix of configurations are available, rather than that information being hidden across a dozen different trait implementations throughout the kernel crate.

> > ```
> > 1. Rust/LLVM does not elide functions in traits that are never called, if that trait is used somewhere. This leads to an unnecessary increase in binary size.
> > ```
> 
> Since Tock started we have been fighting rust to support our use case (remember xargo?). Things have gotten better, and @hudson-ayers mentioned that there is support somewhere in the toolchain for this feature, so it is possible that it will be default in Rust at some point. Is there any attempt to provide a version of this somehow? It might very well use features, but instead of being related to a specific task (remove process print code for panics), it would be a generic fix-a-limitation-in-rust tool that we could use anywhere in the tock kernel.

There is recently-added support in LLVM for this feature, but rustc needs to be modified to make it actually usable by Rust code -- currently support only exists for C++ virtual methods. I have written and submitted a proposal for this feature to be implemented by toolchain developers, but it is not usable today with or without nightly features.

phil-levis:
I think there is an additional complication here, which is related to @bradjc 's point about board authors:

> 1. Board authors have no way of configuring how process debugging information is provided/displayed. Some boards may not want to have process debugging info. Others may want it to be encoded in some packed binary representation and unpacked and formatted by an off-board tool. Others might want even more debugging information that we provide now.

At the same time, we have so far followed a policy that there is one entry in `boards/` for each hardware board Tock supports. This touches on @hudson-ayers's point about configuration. The idea that one changes debugging output by changing executable code (e.g., which implementation of a trait you pass) in `main.rs` is problematic. The argument implies that the level of debugging for a board is a static thing, rather than something you might want to easily change quickly. That being said, given the concern here is code size, standard very dynamic approaches (settable flags in `ProcessConsole`!) won't help.

I of course agree with @bradjc's point about tackling this issue holistically. How do we want to handle debugging output? In practice, we don't generally allow standard debugging output within the kernel and capsules. I.e., we ask contributors to cut out `debug!` statements except in a few rare cases. But at the same time I think that the issues faced by capsules and chips are different than the kernel; they are expected to churn and change more, and are board-specific. So I'd like to suggest that we focus on a holistic answer to this problem *for the kernel*. So we can write down what the classes of debugging information are for the kernel, and write a configuration structure and methodology around that. This would specify *only* debugging output, and not functional behavior (e.g., process fault policies).

For something to make this list, I think it would need to satisfy at least two conditions:
  1. It's output which is commonly needed. It's not a one-off issue, but rather that you can expect to run into a few times if you develop for Tock.
  2. It imposes a significant overhead when enabled, either in terms of CPU cycles or code size (syscall tracing is the former, panic prints is the latter).

Here's what would make my list:
  1. TBF loading
  2. TBF validation/security
  3. System call tracing
  4. Process faults
  5. Kernel faults
  6. Security access violations (e.g., storage IDs)

bradjc:
> For `Process` and debug info, things get a little trickier. Looking at functions like `print_memory_map()`, they rely extensively on directly accessing private fields of the `ProcessStandard` struct, and modifying `print_memory_map()` to be implemented on a different struct that only uses the interface exposed by the `Process` trait would probably have a significant cost to boards that _do_ want that functionality, and would require us to add lots of stuff to the `Process` struct that is really just about making information accessible that is needed for debug prints.

I wasn't sure about this, so I tried implementing `print_memory_map()` without accessing private fields. My attempt is here: https://github.com/tock/tock/compare/print-memory-map-using-methods?expand=1. I got it down to only 3 uses of private fields which we have to implement getter functions for, the rest already exist.

hudson-ayers:
> > For `Process` and debug info, things get a little trickier. Looking at functions like `print_memory_map()`, they rely extensively on directly accessing private fields of the `ProcessStandard` struct, and modifying `print_memory_map()` to be implemented on a different struct that only uses the interface exposed by the `Process` trait would probably have a significant cost to boards that _do_ want that functionality, and would require us to add lots of stuff to the `Process` struct that is really just about making information accessible that is needed for debug prints.
> 
> I wasn't sure about this, so I tried implementing `print_memory_map()` without accessing private fields. My attempt is here: https://github.com/tock/tock/compare/print-memory-map-using-methods?expand=1. I got it down to only 3 uses of private fields which we have to implement getter functions for, the rest already exist.

Huh, that was easier than I expected. I think I should have used `print_full_process()` as an example instead, which also requires access to the `chip`, `kernel`, `stored_state`, `grant_pointers`, `mpu_config`, and `header` fields. Are we going to create a requirement that all implementations of the `Process` trait provide handles to all of these things? `chip` and `kernel` are probably reasonable, and maybe `stored_state` as well. But `GrantPointerEntry` seems like something that would not necessarily be stored in any implementation of `Process`, ditto for `TbfHeader` and maybe even `MpuConfig`. If we are okay with exposing all of those things via getters I do think that this may be worth considering as an approach for just panic info at least (though of course there is a cost to adding all of these getters in that it grows the Process vtable and adds the code for each of the getters).

bradjc:
How big is `print_full_process()`? The implementation relies on `userspace_kernel_boundary().print_context()` and `writer.write_fmt(format_args!("{}", mpu_config))`, so do we run into the same problems where even if `print_full_process()` is removed the MPU and UKB print implementations persist?

hudson-ayers:
> How big is `print_full_process()`? The implementation relies on `userspace_kernel_boundary().print_context()` and `writer.write_fmt(format_args!("{}", mpu_config))`, so do we run into the same problems where even if `print_full_process()` is removed the MPU and UKB print implementations persist?

It looks like removing print_memory_map removes 6720 bytes and removing print_full_process removes an addition 4096 bytes on Arm. userspace_kernel_boundary::print_context() is not included in the binary if you remove print_full_process from the binary, because the `UserspaceKernelBoundary` trait is never used as a trait object, so LLVM can correctly determine that the method is never called (currently, it is only called in `print_full_process()`). The same should apply for the implementation of fmt::Display for the respective MpuConfig type, assuming we do not print the MpuConfig anywhere else.

bradjc:
> It looks like removing print_memory_map removes 6720 bytes and removing print_full_process removes an addition 4096 bytes on Arm. userspace_kernel_boundary::print_context() is not included in the binary if you remove print_full_process from the binary, because the `UserspaceKernelBoundary` trait is never used as a trait object, so LLVM can correctly determine that the method is never called (currently, it is only called in `print_full_process()`). The same should apply for the implementation of fmt::Display for the respective MpuConfig type, assuming we do not print the MpuConfig anywhere else.

I'm unsure of how to handle `print_full_process()`. Is it even possible to make the types work to provide getters for the MPU state and UKB state?

hudson-ayers:
> I'm unsure of how to handle `print_full_process()`. Is it even possible to make the types work to provide getters for the MPU state and UKB state?

I don't think so, at least not without adding generic parameters to the Process struct, which is untenable for reasons discussed in the past. We could have methods on the `Process` trait like this:
```rust
trait Process {
    fn get_mpu_state(&self) -> &dyn MpuConfig;
    fn get_ukb(&self) -> &dyn UserspaceKernelBoundary;
}
```
but this fails to compile because those traits have associated types, and we cannot support the different associated types that will be needed depending on the Chip without adding generic parameters.

bradjc:
> I don't think so, at least not without adding generic parameters to the Process struct

So we could have

```rust
trait Process {
  ...
  fn print_mpu_state()
  fn print_grant_state()
  fn print_internal-process-X_state()
  ...
}
```

where the functions only look like
```rust
fn print_mpu_state () {
  self.mpu_config.print()
}
```

and then mpu_config could look like

```rust
struct MpuConfig {
  printer: &dyn MpuPrinter
  ...
}

impl MpuConfig {
  fn print() {
    self.printer.print(self)
  }
}
```

The general workflow would be that anything that wants to display its state for debugging has to take a reference to a printer, and boards have to provide their preferred printer to each thing that can be printed for debugging.

Hopefully, if you don't want to debug something, you provide a null printer and the resulting overhead is pretty minimal in terms of code size. But, this also provides flexibility if you want your printer to, for example, log to flash or compress the debug state and send the output over UART.

I don't think this is the solution for all in-kernel debugging. If this approach is feasible, it seems reasonable to me for process debugging in particular because process printing doesn't happen very often and is probably never on a critical path, so any overhead in jumping through functions and `dyn` calls shouldn't be a problem.

bradjc:
I started a proof of concept for a `ProcessPrinter` trait here: https://github.com/tock/tock/compare/process-printer?expand=1

The interface is basically `fn print(&self, process: &dyn Process, writer: &mut dyn Write) -> bool` where the return value is `true` if the caller should call print again, and `false` otherwise.

Adding it to the panic print in debug was pretty straightforward:

```rust
pub unsafe fn panic_process_info<PP: ProcessPrinter, W: Write>(
    procs: &'static [Option<&'static dyn Process>],
    process_printer: &'static Option<&'static PP>,
    writer: &mut W,
) {
    process_printer.map(|printer| {
        // print data about each process
        let _ = writer.write_fmt(format_args!("\r\n---| App Status |---\r\n"));
        for idx in 0..procs.len() {
            procs[idx].map(|process| {
                // print everything synchronously
                loop {
                    let keep_going = printer.print(process, writer);
                    if !keep_going {
                        break;
                    }
                }
            });
        }
    });
}
```

Hopefully it will also be straightforward to include in the process console. It will be a little more complex, as the process console has to wait for a callback before calling `print()` again.

There's also the issue of exposing all of the necessary variables outside of `ProcessStandard`, which I haven't attempted.

hudson-ayers:
> > I don't think so, at least not without adding generic parameters to the Process struct
> 
> So we could have
> 
> ```rust
> trait Process {
>   ...
>   fn print_mpu_state()
>   fn print_grant_state()
>   fn print_internal-process-X_state()
>   ...
> }
> ```
> 
> where the functions only look like
> 
> ```rust
> fn print_mpu_state () {
>   self.mpu_config.print()
> }
> ```
> 
> and then mpu_config could look like
> 
> ```rust
> struct MpuConfig {
>   printer: &dyn MpuPrinter
>   ...
> }
> 
> impl MpuConfig {
>   fn print() {
>     self.printer.print(self)
>   }
> }
> ```
> 
> The general workflow would be that anything that wants to display its state for debugging has to take a reference to a printer, and boards have to provide their preferred printer to each thing that can be printed for debugging.

I haven't looked at your branch yet, but I agree that conceptually this sounds like it should work. In fact, I think it may even be possible to use generics for things like `MpuPrinter` -- `MpuConfig` is already a generic type (cleanly hidden from the rest of the kernel behind the trait object for the `Process` struct), and I don't immediately see a reason that `MpuConfig` would be unable to have a generic parameter. This would help cut down on size and runtime overhead in the case where print() does nothing (it can be optimized out entirely) as well as the case where it does something (it can be ).

> Hopefully, if you don't want to debug something, you provide a null printer and the resulting overhead is pretty minimal in terms of code size. But, this also provides flexibility if you want your printer to, for example, log to flash or compress the debug state and send the output over UART.
> 
> I don't think this is the solution for all in-kernel debugging. If this approach is feasible, it seems reasonable to me for process debugging in particular because process printing doesn't happen very often and is probably never on a critical path, so any overhead in jumping through functions and `dyn` calls shouldn't be a problem.

I do think this approach is still pretty clearly different from one we would come up with if not for Rust being unable to optimize out uncalled methods on trait objects -- for instance, if not for that limitation I doubt that we would have separate traits for `MpuPrinter` and `UKBPrinter` etc., and would instead have a single trait for `ProcessInfoPrinter` which could be implemented to log to flash or the network or use compressed output etc.

bradjc:
> would instead have a single trait for `ProcessInfoPrinter` which could be implemented to log to flash or the network or use compressed output etc.

I started building a `ProcessPrinter` trait, and if we wanted that to print everything (not just the memory map), won't that run into the issue where we can't have `Process.get_mpu_state()`, and therefore the `ProcessPrinter` implementation can display mpu info?

hudson-ayers:
> I started building a `ProcessPrinter` trait, and if we wanted that to print everything (not just the memory map), won't that run into the issue where we can't have `Process.get_mpu_state()`, and therefore the `ProcessPrinter` implementation can display mpu info?

You're right, sorry, I've talked myself in a circle here a bit haha. This comment led me to think about this some more, and I do have one idea:

```rust
//! kernel/src/process_info.rs
trait ProcessInfoPrinter {
    fn print_full_process(&self, id: ProcessId);
}

struct ProcessStandardDefaultPrinter {
    processes: &'static [&'static ProcessStandard; NUM_PROCS],
    // ..
}

impl ProcessInfoPrinter for ProcessStandardDefaultPrinter {
    fn print_full_process(&self, id: ProcessId) {
        let proc = self.processes[id.idx()];
        // current body of print_full_process. This works because the printer knows the concrete type.
        // or any other implementation
    }
}

impl ProcessInfoPrinter for () {
    fn print_full_process(&self, id: ProcessId) {}
}

//! kernel/src/process_standard.rs
struct ProcessStandard {
    process_printer: &dyn ProcessInfoPrinter,
    // ...
}

impl Process for ProcessStandard {
    fn print_full_process(&self) {
        self.process_printer.print_full_process(self, self.mpu_config);
    }
}

//! main.rs
let pprinter = kernel::ProcessStandardDefaultPrinter;
// let pprinter = (); // comment me in to disable process printing.
kernel::load_processes(pprinter, //...);
```
This ties ProcessInfoPrinter to the underlying Process type (e.g. if you want to use this printer for an external Process type, you have to reimplement the printer as well), but it gives board the ability to specify their own ProcessPrinter, and it solves the code size issue (if you never construct a `ProcessStandardDefaultPrinter`, its code will be excluded).

Thoughts?

bradjc:
> Thoughts?

1. I'm worried this would add a de-facto restriction that only one impl of `Process` can exist in the kernel at one time. So experimenting with in-kernel processes would be impossible without hacking out the process printer. But this might be an ill founded concern.
2. One challenge is `writer` is synchronous. Having a large enough buffer to print everything in print_full_process can be problematic. Without `Printer` traits for the mpu/ukb/etc. it perhaps gets messier to support an async version.
3. If we want to support different print methods (logging to flash, binary over uart, binary over wireless, etc) we need different impls of mpu.print_context, ukb.print_context (or we just have to use the strings they provide).

bradjc:
Update on using a `trait ObjectPrinter` approach: the outlook is bleak.

The fundamental issue is that printing many objects (like the MPU configuration, the context switch state, or the grant region pointers) is tightly linked with the specific implementation. For example, printing a cortex-m MPU configuration is quite different from printing a riscv PMP configuration. That means that adding getters on the trait (i.e. the MPU trait) for all of the relevant information isn't very feasible (whereas it _is_ for the process memory map which is more constrained). That, I think, leaves two options:

1. Create a trait like `ObjectPrintable` (e.g. `MpuConfigPrintable`) and require that each implementation implement it. This trait would have a generic `print()` function, and would create an implementation-specific printable representation. Then, the `ObjectPrinter` would take a `&dyn ObjectPrintable` and call the `print()` function. However, this doesn't work because taking it as `&dyn ObjectPrintable` means that even if it is never used Rust will not remove the implementation of `ObjectPrintable`. This, then, will not address the code size issue. It could address the flexible printing backend issue.
2. Setup the `ObjectPrinter` to use an associated type, with the goal of referring to the object to be printed by its actual type. This would allow implementation-specific printers. It doesn't really address using different printing backends. But, actually getting the types to work so that it actually compiles doesn't seem to actually be possible.

Based on what we've found so far, it seems like to address the code size issue we need Rust to get smarter and elide trait functions which are never called. Until then, using early returns to get rust/llvm to remove the bulk of the function seems like our only option.

kernel/src/config.rs
```diff
@@ -40,11 +40,18 @@ pub(crate) struct Config {
     /// into which SRAM addresses. This can be useful to debug whether the kernel could
     /// successfully load processes, and whether the allocated SRAM is as expected.
     pub(crate) debug_load_processes: bool,
+
+    /// Whether the kernel should output additional debug information on panics.
+    ///
+    /// If enabled, the kernel will include Process::print_full_process() and
+    /// Process::print_memory_map()
```
bradjc, line 47:
```suggestion
    /// If enabled, the kernel will include implementations of `Process::print_full_process()` and
    /// `Process::print_memory_map()` that display the process's state in a human-readable
    /// form.
```

kernel/src/config.rs
```diff
@@ -40,11 +40,18 @@ pub(crate) struct Config {
...
+    pub(crate) debug_panics: bool,
```
bradjc, line 48:
```suggestion
    // This config option is intended to allow for smaller kernel builds (in
    // terms of code size) where printing code is removed from the kernel
    // binary. Ideally, the compiler would automatically remove
    // printing/debugging functions if they are never called, but due to
    // limitations in Rust (as of Sep 2021) that does not happen if the
    // functions are part of a trait (see
    // https://github.com/tock/tock/issues/2594).
    // 
    // Attempts to separate the printing/debugging code from the Process trait
    // have only been moderately successful (see
    // https://github.com/tock/tock/pull/2826 and
    // https://github.com/tock/tock/pull/2759). Until a more complete solution
    // is identified, using configuration constants is the most effective
    // option.
    pub(crate) debug_panics: bool,
```

bradjc (approved):
welp

alistair23 (approved):

lschuermann (approved):

PR #2760, opened 2021-08-12
Date and time support 

irina-b-dev:
### Pull Request Overview

This pull request adds date and time support for Tock. 

In order to add date and time support this pull request:

1. Describes the `Rtc` HIL (Real Time Clock)
2. Adds a date and time syscall driver
3. Implements the `Rtc` HIL for the Raspberry Pi Pico

An interesting fact is that the `Rtc ` HIL is designed to support both synchronous and asynchronous operations.
Depending on the underlying driver, setting the current date and time can be synchronous or asynchronous. On some chips setting the date and time implies simply writing on a register, while other devices might use external RTC hardware that requires an asynchronous bus transmission.

### Testing Strategy

This pull request was tested using a Raspberry Pi Pico.

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


kernel/src/lib.rs
```diff
@@ -86,6 +86,7 @@
 //!    use cases are discovered.
 
 #![feature(core_intrinsics, const_fn_trait_bound)]
+#![feature(in_band_lifetimes)]
```
alexandruradovici, line 89:
```suggestion
```

bradjc:
This looks really good to me.

For me, I'm not keen on exposing both a synchronous version and an async version. Converting a sync to async is pretty straightforward using deferred calls, and then all of the upper layers do not have to handle the complexity stemming from the underlying hardware. By exposing either, all upper layers have to include additional logic to handle both cases. Also, tock is designed to be async, and something like GPIO is an exception to the rule.

kernel/src/hil/time.rs
```diff
@@ -714,6 +714,95 @@ impl PartialEq for Ticks64 {
...
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct DateTime {
+    pub year: u16,
+    pub month: Month,
+    pub day: u8,
+    pub day_of_week: DayOfWeek,
+    pub hour: u8,
+    pub minute: u8,
+    pub seconds: u8,
+}
```
jrvanwhy, line 753: Time is complex; I have multiple questions:

1. Does all RTC hardware keep track of year/month/day/day_of_week/hour/minute/seconds separately like this? I would expect that some RTCs would track seconds, and would be used to track time since an epoch or Unix time.
2. What timezone is a `DateTime` in? If the answer is "that's application-specific", then what if multiple processes want to interact with the RTC syscall API?
3. How are leap seconds handled? Do we ignore them, and assume that the RTC time will be corrected occasionally? I don't think we want to encode a leap second table in Tock.

irina-b-dev:
Thank you for the feedback, @bradjc ! 
I modified the date_time capsule and the HIL to support only asynchronous interfaces.

kernel/src/hil/time.rs
alexandruradovici: > Time is complex; this definition poses several questions:
> 
> 
> 
> 1. Does all RTC hardware keep track of year/month/day/day_of_week/hour/minute/seconds separately like this? I would expect that some RTCs would track seconds, and would be used to track time since an epoch or Unix time.
> 

You are probably right, but I'm not sure how to proceed here without having the kernel perform a conversion. From the pure HIL point of view, I think that its purpose is to track the time, not just seconds.

A possibility would be to have another UnixRtc HIL that interacts with unix time rtc hardware.

> 2. What timezone is a `DateTime` in? If the answer is "that's application-specific", then what if multiple processes want to interact with the RTC syscall API?
> 

I think this is a userspace problem rather than a kernel one. The RTC driver reports the stored time, just like for a normal computer. It is up to the userspace library or application to convert it to local time.

> 3. How are leap seconds handled? Do we ignore them, and assume that the RTC time will be corrected occasionally? I don't think we want to encode a leap second table in Tock.

I think that the easiest way is to just ignore it for now. Systems will probably sync time with a sever from time to time and this will correct the leap second problem.

irina-b-dev:
ping

chips/rp2040/src/rtc.rs
```diff
@@ -0,0 +1,396 @@
...
+
+impl<'a> time::Rtc<'a> for Rtc<'a> {
+    fn get_date_time(&self) -> Result<(), ErrorCode> {
+        let month_num: u32 = self.registers.setup_0.read(SETUP_0::MONTH);
+        let month_name: Month = match self.month_try_from_u32(month_num) {
+            Result::Ok(t) => t,
+            Result::Err(e) => {
+                return Err(e);
+            }
+        };
```
bradjc, line 359:
```suggestion
        let month_name: Month = self.month_try_from_u32(month_num)?;
```

chips/rp2040/src/rtc.rs
```diff
@@ -0,0 +1,396 @@
...
+        let dotw_num = self.registers.setup_1.read(SETUP_1::DOTW);
+        let dotw = match self.dotw_try_from_u32(dotw_num) {
+            Result::Ok(t) => t,
+            Result::Err(e) => {
+                return Err(e);
+            }
+        };
```
bradjc, line 366:
```suggestion
        let dotw = self.dotw_try_from_u32(dotw_num)?;
```

kernel/src/hil/time.rs
```diff
@@ -714,6 +714,76 @@ impl PartialEq for Ticks64 {
...
+
+/// Interface for reading and setting the current time
+pub trait Rtc<'a> {
```
bradjc, line 756: Modifying the time HIL is complicated, and we probably want to avoid that. I think it would be better to just create a new real time clock HIL file and put this there.

kernel/src/hil/time.rs
irina-b-dev: > > Time is complex; this definition poses several questions:
> > 
> > 1. Does all RTC hardware keep track of year/month/day/day_of_week/hour/minute/seconds separately like this? I would expect that some RTCs would track seconds, and would be used to track time since an epoch or Unix time.
> 
> You are probably right, but I'm not sure how to proceed here without having the kernel perform a conversion. From the pure HIL point of view, I think that its purpose is to track the time, not just seconds.
> 
> A possibility would be to have another UnixRtc HIL that interacts with unix time rtc hardware.
>

I renamed the HIL to `DateTime` so that its name better reflects what it does: it keeps track of the date and time. Underlying chips may implement this in any way the want, such as RTC or other hardware component.

I also moved it to `hil/src/date_time.rs` to separate it from the time HILs.

irina-b-dev:
I think this is ready now.

kernel/src/hil/date_time.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    /// Sets a client that calls the callback function when date and time is requested
+    fn set_client(&self, client: &'a dyn DateTimeClient);
+}
+
+/// Callback handler for when current date is read or set.
+pub trait DateTimeClient {
+    /// Called when a date time reading has completed.
+    /// Takes `Ok(DateTime)` of current date and passes it when scheduling an upcall.
+    /// If an error is encountered it takes an `Err(ErrorCode)`
```
bradjc, line 64: What are the valid errors, and what do they mean?

kernel/src/hil/date_time.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn callback_get_date(&self, datetime: Result<Date, ErrorCode>);
+
+    /// Called when a date is set
+    /// Takes `Ok(())` if time is set correctly.
+    /// Takes  `Err(ErrorCode)` in case of an error
```
bradjc, line 69: What are the errors?

kernel/src/hil/date_time.rs
irina-b-dev: The `ErrorCode` passed as parameter is the same `ErrorCode` that the low level driver might issue if it encounters an inconsistent reading. All ErrorCodes are going to be translated into (usize) statuscodes in the `date_time` capsule. So unless the `into_statuscode` function does not support all error codes, there shouldn't be any invalid errors. 

For synchronous drivers (like the *rp2040/src/rtc.rs*) the errors can be checked before setting the deferred call and therefore in the callback function it is assumed that the result given by the driver is always correct. 
Not sure about asynchronous reading and setting. In this case I assume that the values returned by the driver should be checked right before `callback_get_date` is called. I think the errors that would make sense in this case would be : `FAIL`, `INVAL` or `UNINSTALLED`.

boards/raspberry_pi_pico/layout.ld
```diff
@@ -7,8 +7,8 @@ MEMORY
   ram (rwx) : ORIGIN = 0x20004000, LENGTH = 240K */
 
   /* boot from Flash */
-  rom (rx)  : ORIGIN = 0x10000000, LENGTH = 128K
-  prog (rx) : ORIGIN = 0x10020000, LENGTH = 256K
+  rom (rx)  : ORIGIN = 0x10000000, LENGTH = 256K
+  prog (rx) : ORIGIN = 0x10040000, LENGTH = 256K
```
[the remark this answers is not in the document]
irina-b-dev: Yes, otherwise the rom region will overflow by 8192 bytes.

kernel/src/hil/date_time.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/// Interface for reading and setting the current time
+pub trait DateTime<'a> {
+    /// Request driver to return date and time
+    ///
+    /// When successful this function call must be followed by a call
+    /// to `callback_get_date` which provides the actual date and time
+    /// or an error.
+    fn get_date_time(&self) -> Result<(), ErrorCode>;
```
phil-levis, line 48: What `ErrorCode` values can be returned, and what do they mean?

kernel/src/hil/date_time.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    /// Sets the current date and time
+    ///
+    /// When successful this function call must be followed by a call
+    /// to `callback_set_date`.
+    fn set_date_time(&self, date_time: Date) -> Result<(), ErrorCode>;
```
phil-levis, line 54: What `ErrorCode` values can be returned, and what do they mean?

kernel/src/hil/date_time.rs
phil-levis: Those could make sense: what do they mean? We typically try to document the meaning of different `ErrorCode` values similarly to how POSIX documents errno values.

capsules/src/date_time.rs
```diff
@@ -0,0 +1,350 @@
...
+    subscribed: bool,
+}
+
+pub struct DateTime<'a> {
+    date_time: &'a dyn date_time::DateTime<'a>,
+    apps: Grant<AppData, UpcallCount<1>, AllowRoCount<0>, AllowRwCount<0>>,
+    in_progress: Cell<bool>,
+}
+
+register_bitfields![u32,
```
phil-levis, line 50: It would be good to document that these aren't actually registers, it's just using register definitions for bitfields.

phil-levis (changes requested):

bradjc:

PR #2762, opened 2021-08-12
Minor updates to SD card docs and error codes

brghena:
### Pull Request Overview

Update the usage comments for the SD card to use various components. This is tested to work successfully on a Microbit v2 (wired to an SD card with a breadboard).

Also update the SD Card specific error codes so they don't overlap with tock returncodes.

None of these are bugs, just updates. So this could wait or be merged before 2.0.

### Testing Strategy

Tested on Microbit v2 with SD card hardware and the SD card app `examples/tests/sdcard/`.

### TODO or Help Wanted

How should custom capsule error codes be numbered? I picked an arbitrarily big prefix so they didn't overlap with returncodes. (SD Card error + 10000) The testing app doesn't look for anything specific and no hardware uses SD cards right now, so we can choose whatever makes sense.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/sdcard.rs
+++ b/capsules/src/sdcard.rs
@@ -8,29 +8,50 @@
 //! ```rust
 //! # use kernel::static_init;
 //! # use capsules::virtual_alarm::VirtualMuxAlarm;
+//! # use kernel::hil::spi::SpiMasterDevice;
 //!
-//! let sdcard_spi = static_init!(
-//!     capsules::virtual_spi::VirtualSpiMasterDevice<'static, usart::USART>,
-//!     capsules::virtual_spi::VirtualSpiMasterDevice::new(mux_spi,
-//!                                                        Some(&sam4l::gpio::PA[13])));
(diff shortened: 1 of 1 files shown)


bradjc (approved):

lschuermann (approved):

hudson-ayers (approved):

ppannuto (approved):
> How should custom capsule error codes be numbered?

What is the goal with the numbering scheme? Is it just to ensure that custom capsules as a class never overlap 'official' Tock return codes, or is our goal to provide a separate 'error code space' to each custom capsule? For the latter, I might imagine leveraging driver numbers somehow (maybe reserving the last N bits of the driver number; it's a huge space right now, don't need all those driver numbers...)

brghena:
The most important aspect to me is that the number doesn't overlap with standard error numbers. Previously I was getting `-2`, which has another meaning. Getting `-10002` is different enough that I'm going to go look into the capsule for it. It seems likely to me that as a developer you already know which capsule is the source of the error.

PR #2765, opened 2021-08-13
OpenTitan: Flash fixes

alistair23:
### Pull Request Overview

There was a bug in the OpenTitan flash controller where the app flash at and above 0x2004_0000 was erased (set to all 1s).

The problem is that this was really hard to fix.

It turns out that the flash on the Nexys Video FPGA board is cut in half (https://github.com/lowRISC/opentitan/issues/7730) so the address would wrap around.

This PR fixes the flash controller to not access the high addresses of flash to avoid the wrap around issue in hardware. This fixes all OT release tests.

### Testing Strategy

Run the release tests and kernel unit tests.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/tickv.rs
+++ b/capsules/src/tickv.rs
@@ -97,7 +97,7 @@ impl<'a, F: Flash> tickv::flash_controller::FlashController<64> for TickFSFlastC
 
         if self
             .flash
-            .write_page((0x20040000 + address) / 64, data_buf)
+            .write_page((0x20060000 + address) / 64, data_buf)
             .is_err()
         {
             return Err(tickv::error_codes::ErrorCode::WriteFail);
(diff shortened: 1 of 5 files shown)


capsules/src/tickv.rs
```diff
@@ -97,7 +97,7 @@ impl<'a, F: Flash> tickv::flash_controller::FlashController<64> for TickFSFlastC
 
         if self
             .flash
-            .write_page((0x20040000 + address) / 64, data_buf)
+            .write_page((0x20060000 + address) / 64, data_buf)
```
bradjc, line 100: I know this isn't relevant to this PR, but why does the tickv capsule have a hardcoded address?

bradjc (approved):

capsules/src/tickv.rs
alistair23: Yeah, it needs to be removed. I only just noticed it as well. That should be a separate PR though

hudson-ayers (approved):
Good that this fixes the tests for now, but I assume some of the changes in the lowrisc/ folder are different than what you would want if you had actual hardware instead of this FPGA board?

PR #2769, opened 2021-08-13
docs: syscalls: Update stability guarantees for 2.0

bradjc:
### Pull Request Overview

This PR revisits the stability guarantees for SyscallDriver implementations in the Tock kernel at the 2.0 release.

For 1.0, we said a handful of SyscallDrivers would provide a stable interface to userspace for the 1.x kernel. Should we do that again?

I propose we keep the same list as before. But now is the time to revisit this.

### Testing Strategy

n/a

### TODO or Help Wanted

Do we need to remove a function from alarm?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/syscalls/README.md
+++ b/doc/syscalls/README.md
@@ -36,12 +36,12 @@ Details of the [application binary interface](../Syscalls.md).
 ## Capsule Provided Drivers
 
 Each driver type that has been allocated a permanent driver number is listed in
-the tables below. The "1.0" column indicates whether the driver has been
-stabilized or not (a "✓" indicates stability) in the Tock 1.0 release.
+the tables below. The "2.0" column indicates whether the driver has been
+stabilized or not (a "✓" indicates stability) in the Tock 2.0 release.
 
 ### Base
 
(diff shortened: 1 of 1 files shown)


alistair23 (approved):

hudson-ayers (approved):

ppannuto (approved):
I think stabalizing more of these is likely very aligned with the TRD push -- stable TRD ~= stable interface

lschuermann:
I'm not against stabilization, but I'm not quite sure whether we want to do this concurrently to this release? There are still some drivers which might be worth looking into:

https://github.com/tock/tock/blob/037d174ebc97ce67977f94211ea4b53dd4d57db2/capsules/src/alarm.rs#L226-L231

I'm assuming a 2.0 release qualifies for "future versions of Tock".

bradjc:
We want to guarantee stability now so that we can compile test apps with the release that 1) will work for all users going forward and 2) we can use to test for regressions in later 2.x releases.

lschuermann (approved):

hudson-ayers:
bors d+

Issue #2594, opened 2021-05-24
`Process` trait implementation is not optimised out


alistair23:
Although I can't find documentation stating this, it seems that any implementation of `pub` traits will not be optimsied out in the final binary.

A good example of this is the `print_full_process()` function. It's an implementation of the `Process` trait. I have made the following changes to a RISC-V and an ARM board:

```diff
diff --git a/boards/redboard_artemis_nano/src/io.rs b/boards/redboard_artemis_nano/src/io.rs
index a31982232..821d2992a 100644
--- a/boards/redboard_artemis_nano/src/io.rs
+++ b/boards/redboard_artemis_nano/src/io.rs
@@ -42,12 +42,5 @@ pub unsafe extern "C" fn panic_fmt(info: &PanicInfo) -> ! {
     let led = &mut led::LedLow::new(led_pin);
     let writer = &mut WRITER;
 
-    debug::panic(
-        &mut [led],
-        writer,
-        info,
-        &cortexm4::support::nop,
-        &PROCESSES,
-        &CHIP,
-    )
+    loop {}
 }
diff --git a/boards/swervolf/src/io.rs b/boards/swervolf/src/io.rs
index 0ccf74c16..d01fdaa03 100644
--- a/boards/swervolf/src/io.rs
+++ b/boards/swervolf/src/io.rs
@@ -37,8 +37,6 @@ impl IoWrite for Writer {
 pub unsafe extern "C" fn panic_fmt(pi: &PanicInfo) -> ! {
     let writer = &mut WRITER;
 
-    debug::panic_print(writer, pi, &rv32i::support::nop, &PROCESSES, &CHIP);
-
     // By writing to address 0x80001009 we can exit the simulation.
     // So instead of blinking in a loop let's exit the simulation.
     write_volatile(0x8000_1009 as *mut u8, 20);
```

This ensures that the `print_full_process()` is never called. This can be verified by inspecting the objdump and confirming that `print_full_process()` is never called, yet the entire function is included.

Depending on the board `print_full_process()` is somewhere between the 3rd and 15th largest function. There are other `Process` debug functions like `print_full_process()` that also have the same problem and are also very large.

A quick look at the SweRV binary shows that just these two unused print functions take up about 7% of the binary size. The string formatting is also a large part.

This has come up in the past, see https://github.com/tock/tock/pull/1830 and https://github.com/tock/tock/issues/1995

[...]

PR #2772, opened 2021-08-13
kernel: process: Split panic debug functions to their own trait

bradjc:
### Pull Request Overview

Split process print functions into their own trait so that Rust will elide them if the trait is never used.

I spent 6 minutes on this as a rough proof of concept. Was going alright until:

```
$ make
   Compiling cortexm v0.1.0 (/Users/bradjc/git/tock/arch/cortex-m)
   Compiling capsules v0.1.0 (/Users/bradjc/git/tock/capsules)
   Compiling cortexm4 v0.1.0 (/Users/bradjc/git/tock/arch/cortex-m4)
   Compiling sam4l v0.1.0 (/Users/bradjc/git/tock/chips/sam4l)
   Compiling components v0.1.0 (/Users/bradjc/git/tock/boards/components)
   Compiling hail v0.1.0 (/Users/bradjc/git/tock/boards/hail)
  --> boards/hail/src/io.rs:73:9
   |
73 |         &PROCESSES,
   |         ^^^^^^^^^^ expected slice, found array of 20 elements
   |
   = note: expected reference `&'static [Option<&'static (dyn ProcessDebug + 'static)>]`
              found reference `&[Option<&'static (dyn Process + 'static)>; 20]`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `hail`
```

### Testing Strategy

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/kernel/src/debug.rs
+++ b/kernel/src/debug.rs
@@ -58,7 +58,7 @@ use crate::collections::queue::Queue;
 use crate::collections::ring_buffer::RingBuffer;
 use crate::hil;
 use crate::platform::chip::Chip;
-use crate::process::Process;
+use crate::process::ProcessDebug;
 use crate::utilities::cells::NumericCellExt;
 use crate::utilities::cells::{MapCell, TakeCell};
 use crate::ErrorCode;
@@ -103,7 +103,7 @@ pub unsafe fn panic_print<W: Write + IoWrite, C: Chip>(
     writer: &mut W,
     panic_info: &PanicInfo,
     nop: &dyn Fn(),
-    processes: &'static [Option<&'static dyn Process>],
+    processes: &'static [Option<&'static dyn ProcessDebug>],
     chip: &'static Option<&'static C>,
 ) {
     panic_begin(nop);
@@ -122,7 +122,7 @@ pub unsafe fn panic<L: hil::led::Led, W: Write + IoWrite, C: Chip>(
     writer: &mut W,
     panic_info: &PanicInfo,
     nop: &dyn Fn(),
-    processes: &'static [Option<&'static dyn Process>],
+    processes: &'static [Option<&'static dyn ProcessDebug>],
     chip: &'static Option<&'static C>,
 ) -> ! {
     // Call `panic_print` first which will print out the panic
@@ -177,7 +177,7 @@ pub unsafe fn panic_cpu_state<W: Write, C: Chip>(
 ///
 /// **NOTE:** The supplied `writer` must be synchronous.
 pub unsafe fn panic_process_info<W: Write>(
-    procs: &'static [Option<&'static dyn Process>],
+    procs: &'static [Option<&'static dyn ProcessDebug>],
     writer: &mut W,
 ) {
     // print data about each process
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -168,9 +168,11 @@ impl ProcessId {
     }
 }
 
+pub trait Process: ProcessMain + ProcessDebug {}
+
 /// This trait represents a generic process that the Tock scheduler can
 /// schedule.
-pub trait Process {
+pub trait ProcessMain {
     /// Returns the process's identifier.
     fn processid(&self) -> ProcessId;
 
@@ -536,14 +538,6 @@ pub trait Process {
     /// switched to.
     fn switch_to(&self) -> Option<syscall::ContextSwitchReason>;
 
-    /// Print out the memory map (Grant region, heap, stack, program
-    /// memory, BSS, and data sections) of this process.
-    fn print_memory_map(&self, writer: &mut dyn Write);
-
-    /// Print out the full state of the process: its memory map, its
-    /// context, and the state of the memory protection unit (MPU).
-    fn print_full_process(&self, writer: &mut dyn Write);
-
     // debug
 
     /// Returns how many syscalls this app has called.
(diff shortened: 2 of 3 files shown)


alistair23 (approved):
I like the concept! This is related to https://github.com/tock/tock/issues/2594

bradjc:
As I learn more about this, I'm starting to think this approach won't work, and `Process` will have all of the functions just like before. https://stackoverflow.com/questions/49543778/why-cant-i-cast-a-box-with-an-extended-trait-to-a-box-with-the-base-trait

hudson-ayers:
Yeah, I agree this will not work as written. In order for this approach to work, there would need to be some `ProcessStandardNoDebug` struct which would duplicate the `Process` trait implementation but have empty stubs for the `ProcessDebug` trait. But then we are back to duplicating code, and this is not really much better than without the new trait.

We could use macros to prevent code duplication, but having large code blocks in macros sucks, since rustfmt and rust-analyzer do not work as well in the body of macros, and error messages for code inside macros are much harder to parse.

I think the last option would be to have a generic type parameter on `ProcessStandard` that indicates whether to include debug info. Then you could do something like

```rust
struct ProcessStandard<const DEBUG: bool>{//...}

impl ProcessDebug for ProcessStandard<true> {
    // actually implement the trait
}

impl ProcessDebug for ProcessStandard<false> {
    // empty stubs for all methods
}
```

That said, I am not sure how easy it would actually be to make a change like this until I have a chance to try it.

bradjc:
> ```rust
> struct ProcessStandard<const DEBUG: bool>{//...}
> 
> impl ProcessDebug for ProcessStandard<true> {
>     // actually implement the trait
> }
> 
> impl ProcessDebug for ProcessStandard<false> {
>     // empty stubs for all methods
> }
> ```

Makes sense, but this doesn't really give us the API we want, and we would be back to the problem where you would need to make (at least) two changes: to the `PROCESSES` array in main.rs and to the panic handler in io.rs, and, they wouldn't be mutually enforced (you could remove the panic debug call in io.rs but not remove the process printing code).

It seems like we would actually be better off creating a new `ProcessPrinter` trait, and creating a new object that displays a process.

hudson-ayers:
> > ```rust
> > struct ProcessStandard<const DEBUG: bool>{//...}
> > 
> > impl ProcessDebug for ProcessStandard<true> {
> >     // actually implement the trait
> > }
> > 
> > impl ProcessDebug for ProcessStandard<false> {
> >     // empty stubs for all methods
> > }
> > ```
> 
> Makes sense, but this doesn't really give us the API we want, and we would be back to the problem where you would need to make (at least) two changes: to the `PROCESSES` array in main.rs and to the panic handler in io.rs, and, they wouldn't be mutually enforced (you could remove the panic debug call in io.rs but not remove the process printing code).
> 
> It seems like we would actually be better off creating a new `ProcessPrinter` trait, and creating a new object that displays a process.

Having a separate object for displaying a process requires giving that object the ability to inspect a process, which probably means keeping many of the existing debug_whatever functions on the Process trait, right? Unless we change the visibility of every field to be `pub(crate)`, which I also do not think we want to do.

bradjc:
> Having a separate object for displaying a process requires giving that object the ability to inspect a process, which probably means keeping many of the existing debug_whatever functions on the Process trait, right?

Yes. Were those ever going away?

hudson-ayers:
I was hoping to re-open a revised version of https://github.com/tock/tock/pull/1830 after this PR

bradjc:
> I was hoping to re-open a revised version of #1830 after this PR

Wait so #2759 is giving the proverbial mouse a cookie? We need to brainstorm solutions to solve whatever the entire problem is.

hudson-ayers:
> > I was hoping to re-open a revised version of #1830 after this PR
> 
> Wait so #2759 is giving the proverbial mouse a cookie? We need to brainstorm solutions to solve whatever the entire problem is.

I mentioned it on the core call, I just haven't had a chance to update #1830 to the 2.0 process code yet. It sounds like we do still need to brainstorm solutions for the larger problem though, and I didn't want to take the time to update #1830 until we reach consensus on how we want to do this.

phil-levis (changes requested):
Let's not do piecemeal tweaks on the Process trait. Let's actually redesign it. Engineering, not crafting.

alistair23:
I thought with this trait split we could then do something like:

```rust
#[cfg(debug_prints)]
impl<C: Chip> ProcessDebug for ProcessStandard<'_, C> {
    fn print_memory_map(&self, writer: &mut dyn Write) {
        // Full implementation
    }
}

#[cfg(not(debug_prints))]
impl<C: Chip> ProcessDebug for ProcessStandard<'_, C> {
    fn print_memory_map(&self, writer: &mut dyn Write) {
        // Don't do anything
    }
}
```

The idea here is then it's up to boards to use/not use the debug_prints feature. It seems like a useful board feature (based on flash contraints).

The problem with https://github.com/tock/tock/pull/2759 is it has to impact all boards or no board gets the size benefit by default.

hudson-ayers:
> I thought with this trait split we could then do something like:
> 
> ```rust
> #[cfg(debug_prints)]
> impl<C: Chip> ProcessDebug for ProcessStandard<'_, C> {
>     fn print_memory_map(&self, writer: &mut dyn Write) {
>         // Full implementation
>     }
> }
> 
> #[cfg(not(debug_prints))]
> impl<C: Chip> ProcessDebug for ProcessStandard<'_, C> {
>     fn print_memory_map(&self, writer: &mut dyn Write) {
>         // Don't do anything
>     }
> }
> ```
> 
> The idea here is then it's up to boards to use/not use the debug_prints feature. It seems like a useful board feature (based on flash contraints).
> 
> The problem with #2759 is it has to impact all boards or no board gets the size benefit by default.

If we are willing to use `cfg`, you can get the same benefit with the #2759 approach without the added complexity of splitting the traits by just using:

```rust
/// A unique instance of `Config` where compile-time configuration options are defined. These
/// options are available in the kernel crate to be used for relevant configuration.
    #[cfg(not(trace_syscalls))]
    trace_syscalls: false,
    #[cfg(trace_syscalls)]
    trace_syscalls: true,
    #[cfg(not(debug_load_processes))]
    debug_load_processes: false,
    #[cfg(debug_load_processes)]
    debug_load_processes: true,
    #[cfg(not(debug_processes))]
    debug_processes: false,
    #[cfg(debug_processes)]
    debug_processes: true,
}
```

bradjc:
Closing this as it isn't going to work.

PR #2773, opened 2021-08-14
Prevent adding duplicates to the list

alexandruradovici:
### Pull Request Overview

This pull request adds a check to the linked list that prevents adding duplicates.

I was trying to get the Adafruit Clue board to work for some days now. The board connected to the USB, sent its information, but seemed to get stuck afterwards. It turned out that the error was due to an infinite loop in the SPI Mux. The SPI device got added twice to the list. I am preparing another pull request to update that.

I think it might be a good idea to prevent linked lists from adding the same item twice as this leads to an infinite and issues that are difficult to debug.

An improvement to this would be to move the previous instance of the item to the desired position.

### Testing Strategy

This pull request was tested using the double `setup` of SPI from the Adafruit CLUE.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/collections/list.rs
+++ b/kernel/src/collections/list.rs
@@ -47,16 +47,26 @@ impl<'a, T: ?Sized + ListNode<'a, T>> List<'a, T> {
         self.head.0.get()
     }
 
+    pub fn in_list(&self, node: &'a T) -> bool {
+        self.iter()
+            .find(|list_node| *list_node as *const T == node as *const T)
+            .is_some()
+    }
+
     pub fn push_head(&self, node: &'a T) {
-        node.next().0.set(self.head.0.get());
-        self.head.0.set(Some(node));
+        if !self.in_list(node) {
+            node.next().0.set(self.head.0.get());
+            self.head.0.set(Some(node));
+        }
     }
 
     pub fn push_tail(&self, node: &'a T) {
-        node.next().0.set(None);
-        match self.iter().last() {
-            Some(last) => last.next().0.set(Some(node)),
-            None => self.push_head(node),
+        if !self.in_list(node) {
+            node.next().0.set(None);
+            match self.iter().last() {
+                Some(last) => last.next().0.set(Some(node)),
+                None => self.push_head(node),
+            }
         }
     }
 


alistair23 (approved):

lschuermann:
I'm not entirely sure whether there aren't any good reason to have a list pointing back to itself?

Furthermore, I'm not even sure if this can properly guarantee that this prevents loops in a list from happening. This fix appears to just compare the reference to the `ListNode` that is about to be inserted. However, I can have two disparate `ListNode` instances, not being the same object, but still pointing to the same next item. _If_ we want this -- and I'm not really sure -- we're going to want to compare the the `ListNode::next` versus just the inserted  reference. Even then, we'd have to handle the case of a single `ListNode` pointing back to itself.

hudson-ayers:
This makes inserting nodes O(n) instead of O(1), which is probably not what most users of a linked list data structure would expect. In most Tock use cases this is not likely to matter, but we do use this data structure in the scheduler, where performance matters more. 

That said I have run into bugs like the one you describe, and they are easy mistakes to make and quite frustrating.

alexandruradovici:
> This makes inserting nodes O(n) instead of O(1), which is probably not what most users of a linked list data structure would expect. In most Tock use cases this is not likely to matter, but we do use this data structure in the scheduler, where performance matters more.
> 

I understand your point, I totally forgot that this is used by the scheduler. This is not a solution then.

> That said I have run into bugs like the one you describe, and they are easy mistakes to make and quite frustrating.

What about adding a function `push_head_unique` which verifies if this item is already within the list and require a capability for the `push_head` function which does not perform any verification? Drivers could use `push_head_unique`.

lschuermann:
> What about adding a function `push_head_unique` which verifies if this item is already within the list and require a capability for the `push_head` function which does not perform any verification? Drivers could use `push_head_unique`.

I think a separate function would definitely be an improvement, and still allows building very flexible linked lists with loops if one is so inclined.

However, this would only cover the basic case of `ListNode::next` always returning the same items. `ListNode`s can incorporate some intelligence and querying `next` once is not a guarantee that no loops are formed in the list, neither is just comparing their references. I think this flexibility is by design and vouch to keep it this way.

Thus I'd be in favor of a function which performs a sanity check as you described. However, it must be documented that this is only a check, but in no way a guarantee that no loops can be formed in the list.

bradjc:
Why not just create a new data type in kernel/collections?

Or, is modifying the list API the correct place to handle this? Are we expecting this to be a regular problem for lists in Tock, or should we look at how we create/use virtualizers? Virtualizers are very hard to get right, and I'm not sure that patching unrelated code each time is the solution we want to pursue.

brghena:
A different way to resolve this issue is to have the List itself be in charge of creating nodes. It's not clear to me why nodes are created in drivers and then added to the list. We could instead of a ListNode struct that holds a reference to `T` which is created for the user. Then you wouldn't be able to screw up and add the same Node twice. You could add the same `T` twice, perhaps, but that wouldn't mess up the list structure.

phil-levis:
Yeah, this is a bug we've seen a few times, when a static element is initialized/added twice. This is the key bug, not the list part (if anything, the list causing an infinite loop is a nice way to find it!). 

Could one tell if the element is already inserted by looking at its next link? The change here is for a list to have a "tail" value instead of `None`.

phil-levis:
@alexandruradovici  Could we remove the O(n) cost through having a reserved tail value?

lschuermann:
> A different way to resolve this issue is to have the List itself be in charge of creating nodes.

@brghena I believe this is intended flexibility by representing the List as traits. I can imagine many examples where this could come in handy; e.g. potentially some `FlashProcess` could just implement `ListNode`. Whenever `<FlashProcess::ListNode>::next()` is called it would dynamically load the next process from flash and return a reference to that list node.

Which also helps explain why I don't think this PR in its current form is a good idea, or rather that it seems fundamentally opposed to the very dynamic concept which is implemented through the combination of `List`, `ListNode` and `ListLink` here. For instance, because we're dealing with traits, there is no specification on what the cost of calling `next` will be. Also, the behavior of `next` can be entire implementation dependent. Loops in the list can well be created dynamically at runtime.

@brghena Maybe a good idea would be to provide a concrete `LinkedList` type which implements the interface described here, but has a more rigid set of rules -- such as that `next` always returns the next list item and there may never be any loops in the list structure. This `LinkedList` type would also create the `ListNodes` themselves and thus have full control over the list's semantics. We'd have the best of both words: a _very_ flexible and generic interface for implementing list-like structures and a more simple linked list type for when that is sufficient to express the user's data structure, which also checks for and prevents loops.

bradjc (changes requested):
I'm not sure where this stands. For a general `list` type, it isn't clear to me why that would not allow duplicates.

We could add a `set` type? Or do nothing? This issue of registering handlers with a mux doesn't seem to come up too often, and components should make it unlikely to happen once the capsule is initially created.

lschuermann:
> I'm not sure where this stands. For a general `list` type, it isn't clear to me why that would not allow duplicates.
> 
> We could add a `set` type? Or do nothing? This issue of registering handlers with a mux doesn't seem to come up too often, and components should make it unlikely to happen once the capsule is initially created.

A set means the values would be unique, but unordered. Here we're specifically concerned about uniqueness of the List nodes themselves, not necessarily their values. You can have multiple different list nodes (such that they are not introducing loops in the list) which still have the same associated value. Hence my proposal to keep the generic interface (such that these advanced uses are still supported) and a concrete List type which does make sure that all nodes are indeed unique.

lschuermann:
> @brghena Maybe a good idea would be to provide a concrete `LinkedList` type which implements the interface described here, but has a more rigid set of rules -- such as that `next` always returns the next list item and there may never be any loops in the list structure. This `LinkedList` type would also create the `ListNodes` themselves and thus have full control over the list's semantics. We'd have the best of both words: a _very_ flexible and generic interface for implementing list-like structures and a more simple linked list type for when that is sufficient to express the user's data structure, which also checks for and prevents loops.

I gave this idea a shot in #2845. Let me know what you think.

Issue #2603, opened 2021-05-29
Tock syscall fuzzing - application and kernel side crash


Menzo96:
I am working on a fuzzing project on Tock OS and I have encountered some crashes and I couldn't find the cause yet.

I have created a simple fuzzer with libtock-c that tries a few Tock OS system calls. 
The issue encountered manifests in different ways for different Tock OS versions:

- For v 2.0 as a crash on the application side on a cmp instruction.
- For v 1.6 as a crash on the application side, however upon further investigation it seemed to crash in kernel space on a bx lr instruction.

I have detailed the issue on a libtock-c fork here: https://github.com/Menzo96/libtock-c/tree/dumb_fuzzer/examples/dumb_fuzzer

Do you have any idea why this happens? Do you have any advice on how to investigate this further? Is there something that I'm doing wrong (syscall interface, using tock, libtock-c)?

Thank you

alistair23:
A good place to start is to turn on the debug logging, you can find it here: https://github.com/tock/tock/blob/master/kernel/src/config.rs#L47

I don't know enough about ARM to get a good sense of the type of fault (if you can reproduce this on RISC-V I could help) but based on your `Faulting Memory Address` there is some NULL pointer access.

It seems like you have figured out the faulting line. Can you figure out where the faulting NULL pointer comes from? Is it being popped off the stack or does it come from somewhere else?

ppannuto:
<img width="645" alt="image" src="https://user-images.githubusercontent.com/339422/120139783-12ceed80-c18e-11eb-9372-47696a35ae06.png">

If I'm reading this correctly, it looks like the `bx lr` at `0x000083e6` is the last instruction that runs before the panic? To me that points to some corruption in the application stack. Right before that, we loaded up the link register with the magic value `0xffff_fff9`. When that "address" is loaded into the PC, it triggers the hardware to do a context switch. That causes several things to happen, but most importantly, a bunch of registers, including the new PC for the application, is read from the application stack pointer. This suggests that -- somehow -- the application stack is getting corrupted before the return from the kernel.

hudson-ayers:
[...]

PR #2779, opened 2021-08-16
Adjust app's watermark only for memory buffers

alexandruradovici:
### Pull Request Overview

This pull request prevents the change of the the app watermark if the read only shared buffers are in flash.

While @valexandru  was testing the iMXRT1052 EVKB board, he found an interesting behavior for the client-service examples.

Changing the watermark for flash buffers might prevent apps form further allocating memory if the flash memory is located at an address above the RAM memory.

Take as an example an app with the following code:
```c
int main()
{
  allow_readonly (driver_num, allow_num, &"hello", 5);
  printf ("hello\n");
}
```
This app will fault (due to a `printf` memory allocation error #2603) on boards like the iMXRT1052 EVKB where the flash region is at a higher address then the RAM. This is what happens:
1. allow_readonly will place the app watermark in flash
2. printf will not be able to allocate any memory as `memop (1, 0)` fails with `NOMEM` due to the app watermark that is set to a position in flash that is at a much higher address then the requested app break of `memop`. The kernel thinks the app wants to *shrink* its memory region.

Rewriting the code:

```c
int main()
{
  printf ("hello\n");
  allow_readonly (driver_num, allow_num, &"hello", 5);
  printf ("hello\n");
}
```

makes it work, as the second time `printf` is called, it does not perform any more `memop`s calls.

I think this affects all the boards where the flash region is at a higher address than the ram.

Both examples work on the Micro:bit as the flash region is at a lower address than the RAM and sharing a flash buffer does not change the app watermark.

### Testing Strategy

This pull request was tested using an iMXRT1052 EVKB board and a Microbit board.

### TODO or Help Wanted

Some feedback, I might have missed something.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process_standard.rs
+++ b/kernel/src/process_standard.rs
@@ -629,7 +629,7 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
             // TODO: Check for buffer aliasing here
 
             if self.in_app_owned_memory(buf_start_addr, size) {
-                // Valid buffer, if this is a memory buffer,
+                // Valid buffer, if this is not in a read-only (e.g. flash) memory section,
                 // we need to adjust the app's watermark
                 // note: in_app_owned_memory ensures this offset does not wrap
                 let buf_end_addr = buf_start_addr.wrapping_add(size);
--- a/kernel/src/process_standard.rs
+++ b/kernel/src/process_standard.rs
@@ -628,11 +628,14 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
         {
             // TODO: Check for buffer aliasing here
 
-            // Valid buffer, we need to adjust the app's watermark
-            // note: in_app_owned_memory ensures this offset does not wrap
-            let buf_end_addr = buf_start_addr.wrapping_add(size);
-            let new_water_mark = cmp::max(self.allow_high_water_mark.get(), buf_end_addr);
-            self.allow_high_water_mark.set(new_water_mark);
+            if self.in_app_owned_memory(buf_start_addr, size) {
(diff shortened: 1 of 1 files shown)


bradjc (approved):
Wow, great find on this regression.

hudson-ayers (approved):

kernel/src/process_standard.rs
```diff
@@ -628,11 +628,14 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
         {
             // TODO: Check for buffer aliasing here
 
-            // Valid buffer, we need to adjust the app's watermark
-            // note: in_app_owned_memory ensures this offset does not wrap
-            let buf_end_addr = buf_start_addr.wrapping_add(size);
-            let new_water_mark = cmp::max(self.allow_high_water_mark.get(), buf_end_addr);
-            self.allow_high_water_mark.set(new_water_mark);
+            if self.in_app_owned_memory(buf_start_addr, size) {
+                // Valid buffer, if this is a memory buffer,
```
lschuermann, line 632: Great find! I'm not sure I would've understood "memory buffer" at first glance though. I'd propose to clarify that we are not talking about a read-only memory section given flash is also memory.

```suggestion
                // Valid buffer, if this is not in a read-only (e.g. flash) memory section,
```

lschuermann (dismissed):

lschuermann (approved):

hudson-ayers (approved):

bradjc (approved):

alistair23 (approved):

PR #2782, opened 2021-08-16
[stm32] Simplify enabling interrupts for gpio

alexandruradovici:
enables interrupts for the first pin

### Pull Request Overview

This pull request adds a helper function to make enabling interrupts for gpio easier. The stm32 MCU does not provide a separate interrupt line for all the gpio pins. This makes enabling interrupts a little more difficult and requires some extra code in the board file. 

I am not the original author of the code for these boards, so I am not sure why this has been designed the way it is now.

This also enables interrupts for the first gpio exposed pin, allowing the `gpio` test to succeed.

### Testing Strategy

This pull request was tested using a nucleo F429ZI board

### TODO or Help Wanted

@yusefkarim please take a look if this works on your board.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/weact_f401ccu6/src/main.rs
+++ b/boards/weact_f401ccu6/src/main.rs
@@ -143,11 +143,10 @@ unsafe fn setup_dma(
 /// Helper function called during bring-up that configures multiplexed I/O.
 unsafe fn set_pin_primary_functions(
     syscfg: &stm32f401cc::syscfg::Syscfg,
-    exti: &stm32f401cc::exti::Exti,
+    _exti: &stm32f401cc::exti::Exti,
     gpio_ports: &'static stm32f401cc::gpio::GpioPorts<'static>,
 ) {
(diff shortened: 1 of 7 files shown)


boards/weact_f401ccu6/src/main.rs
```diff
@@ -143,11 +143,10 @@ unsafe fn setup_dma(
 /// Helper function called during bring-up that configures multiplexed I/O.
 unsafe fn set_pin_primary_functions(
     syscfg: &stm32f401cc::syscfg::Syscfg,
-    exti: &stm32f401cc::exti::Exti,
+    _exti: &stm32f401cc::exti::Exti,
```
hudson-ayers, line 146: I think we should just delete this argument, this function is not called anywhere else.

hudson-ayers (approved):

alexandruradovici:
Closing due to new pull request #2784

PR #2783, opened 2021-08-16
ESP32-C3: Add support for running apps

alistair23:
### Pull Request Overview

Add support for running Tock applications

### Testing Strategy

Run apps

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/esp32-c3-devkitM-1/layout.ld
+++ b/boards/esp32-c3-devkitM-1/layout.ld
@@ -1,8 +1,8 @@
 MEMORY
 {
   rom (rx)  : ORIGIN = 0x40380000, LENGTH = 0x30000
-  ram (rwx) : ORIGIN = 0x403B0000, LENGTH = 0x30000
-  prog (rx) : ORIGIN = 0x42000000, LENGTH = 0x80000
+  ram (rwx) : ORIGIN = 0x3FCA0000, LENGTH = 0x30000
+  prog (rx) : ORIGIN = 0x403B0000, LENGTH = 0x30000
 }
 
 MPU_MIN_ALIGN = 1K;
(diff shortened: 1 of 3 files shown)


hudson-ayers (approved):

lschuermann (dismissed):

boards/esp32-c3-devkitM-1/layout.ld
```diff
@@ -1,8 +1,8 @@
 MEMORY
 {
   rom (rx)  : ORIGIN = 0x40380000, LENGTH = 0x30000
-  ram (rwx) : ORIGIN = 0x403B0000, LENGTH = 0x30000
```
bradjc, line 4: How did this work?!?

bradjc (approved):

boards/esp32-c3-devkitM-1/layout.ld
lschuermann: Based on the datasheet it appears to me as both of these sections are located in SRAM. Perhaps you made the same mistake as I did and read the `B` as an `8`?

alistair23: It's inside the SRAM memory space

ppannuto (approved):
bors d+

boards/esp32-c3-devkitM-1/layout.ld
bradjc: So if you unplug the board the apps go away?

alistair23: No, it's confusing. Tock and apps are loaded into flash. The boot ROM then loads that flash into RAM and executes it.

So the read-only regions of Tock need to be in RAM as that is where the boot ROM jumps to. In theory we could store the apps in flash, but the ROM seems to do all sorts of strange things so I haven't been able to get that to works. Plus I don't think that would get us much as Tock can't use fragmented memory addresses

PR #2785, opened 2021-08-17
Add process console shell prompt

alexandruradovici:
### Pull Request Overview

This pull request adds a shell prompt to the process console. Many users have no idea that the process console exists.

This pull request modifies it in the following way:
1. the `start` function displays the *Press any key ...* message using `debug!` and should be called after the board's `debug!` message
2. it waits for a key to be pressed and displays the welcome message and the prompt. This first key is not taken into account
3. it prints the kernel version using the `MAJOR` and `MINOR` values. The value from the environment variable is now considered the `build`.

This fixes:
- a bug that allowed users to delete more characters than they have written as the \b character was printed even if the command buffer is empty.
- the kernel print, the Bss region information was not printed as it was in the same buffer with the version and was overflowing the write buffer

I renamed the variable holding the reference to the process console with an _ to be able to disable it only by commenting out the `start` function call. The compiler complains otherwise that the variable is not used.

This is how it looks like now

```
Press any key to start the process console...
Hello World!
Kernel version: 2.0 (build release-2.0-rc2-64-g3c0d323b8)
Welcome to the process console.
Valid commands are: help status list stop start fault process kernel
tock$ 
tock$ list

 PID    Name                Quanta  Syscalls  Dropped Callbacks  Restarts    State  Grants
  0	blink                   27       133                  0         0  Yielded    1/10
  1	c_hello                  2         8                  0         0  Yielded    1/10
  2	buttons                  3        11                  0         0  Yielded    1/10
tock$ list

 PID    Name                Quanta  Syscalls  Dropped Callbacks  Restarts    State  Grants
  0	blink                   39       193                  0         0  Yielded    1/10
  1	c_hello                  2         8                  0         0  Yielded    1/10
  2	buttons                  3        11                  0         0  Yielded    1/10
tock$ kernel

Kernel version: 2.0 (build release-2.0-rc2-64-g3c0d323b8)

 ╔═══════════╤══════════════════════════════╗
 ║  Address  │ Region Name    Used (bytes)  ║
 ╚0x20011FC0═╪══════════════════════════════╝
             │   Bss         10720
  0x20002028 ┼─────────────────────────────── S
             │   Relocate       40            R
  0x20002000 ┼─────────────────────────────── A
             │ ▼ Stack        8192            M
  0x20000000 ┼───────────────────────────────
             .....
  0x08025550 ┼─────────────────────────────── F
             │   RoData      41380            L
  0x0801B3AC ┼─────────────────────────────── A
             │   Code       111532            S
  0x08000000 ┼─────────────────────────────── H

tock$ list

 PID    Name                Quanta  Syscalls  Dropped Callbacks  Restarts    State  Grants
  0	blink                   95       473                  0         0  Yielded    1/10
  1	c_hello                  2         8                  0         0  Yielded    1/10
  2	buttons                  3        11                  0         0  Yielded    1/10
tock$ process buttons


 ╚0x20018000═╪══════════════════════════════════════════╝
             │ ▼ Grant        1108 |   1108          
  0x20017BAC ┼───────────────────────────────────────────
             │ Unused
  0x200169F0 ┼───────────────────────────────────────────
             │ ▲ Heap            0 |   4540               S
  0x200169F0 ┼─────────────────────────────────────────── R
             │ Data            496 |    496               A
  0x20016800 ┼─────────────────────────────────────────── M
             │ ▼ Stack        2048 |   2048          
  0x20016000 ┼───────────────────────────────────────────
             │ Unused
  0x20016000 ┴───────────────────────────────────────────
             .....
  0x08031800 ┬─────────────────────────────────────────── F
             │ App Flash      1996                        L
  0x08031034 ┼─────────────────────────────────────────── A
             │ Protected        52                        S
  0x08031000 ┴─────────────────────────────────────────── H
tock$ 
```

### Testing Strategy

This pull request was tested using a stm32 board and a microbit.

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/clue_nrf52840/src/main.rs
+++ b/boards/clue_nrf52840/src/main.rs
@@ -699,9 +699,14 @@ pub unsafe fn main() {
         nrf52840::aes::AesECB<'static>
     ));
 
-    let pconsole =
-        components::process_console::ProcessConsoleComponent::new(board_kernel, uart_mux)
-            .finalize(());
+    let pconsole = components::process_console::ProcessConsoleComponent::new(
+        board_kernel,
+        uart_mux,
+        mux_alarm,
+    )
+    .finalize(components::process_console_component_helper!(
+        nrf52840::rtc::Rtc
+    ));
     let _ = pconsole.start();
 
     //--------------------------------------------------------------------------
--- a/capsules/src/process_console.rs
+++ b/capsules/src/process_console.rs
@@ -195,8 +195,6 @@ pub struct ProcessConsole<'a, A: Alarm<'a>, C: ProcessManagementCapability> {
     command_buffer: TakeCell<'static, [u8]>,
     command_index: Cell<usize>,
 
-    first_prompt_shown: Cell<bool>,
-
     /// Flag to mark that the process console is active and has called receive
     /// from the underlying UART.
     running: Cell<bool>,
@@ -272,7 +270,7 @@ impl<'a, A: Alarm<'a>, C: ProcessManagementCapability> ProcessConsole<'a, A, C> 
             rx_buffer: TakeCell::new(rx_buffer),
             command_buffer: TakeCell::new(cmd_buffer),
             command_index: Cell::new(0),
-            first_prompt_shown: Cell::new(false),
+
             running: Cell::new(false),
             execute: Cell::new(false),
             kernel: kernel,
@@ -1025,7 +1023,6 @@ impl<'a, A: Alarm<'a>, C: ProcessManagementCapability> ProcessConsole<'a, A, C> 
     }
 
     fn prompt(&self) {
-        self.first_prompt_shown.set(true);
         let _ = self.write_bytes(b"tock$ ");
     }
 
--- a/capsules/src/process_console.rs
+++ b/capsules/src/process_console.rs
@@ -112,10 +112,12 @@ use core::fmt;
 use core::fmt::write;
 use core::str;
 use kernel::capabilities::ProcessManagementCapability;
+use kernel::hil::time::ConvertTicks;
 use kernel::utilities::cells::TakeCell;
 use kernel::ProcessId;
 
 use kernel::debug;
+use kernel::hil::time::{Alarm, AlarmClient};
 use kernel::hil::uart;
 use kernel::introspection::KernelInfo;
 use kernel::ErrorCode;
@@ -180,8 +182,9 @@ pub struct KernelAddresses {
     pub bss_end: *const u8,
 }
 
-pub struct ProcessConsole<'a, C: ProcessManagementCapability> {
+pub struct ProcessConsole<'a, A: Alarm<'a>, C: ProcessManagementCapability> {
     uart: &'a dyn uart::UartData<'a>,
+    alarm: &'a A,
     tx_in_progress: Cell<bool>,
     tx_buffer: TakeCell<'static, [u8]>,
     queue_buffer: TakeCell<'static, [u8]>,
@@ -191,6 +194,8 @@ pub struct ProcessConsole<'a, C: ProcessManagementCapability> {
     rx_buffer: TakeCell<'static, [u8]>,
     command_buffer: TakeCell<'static, [u8]>,
     command_index: Cell<usize>,
+
+    first_prompt_shown: Cell<bool>,
 
     /// Flag to mark that the process console is active and has called receive
     /// from the underlying UART.
@@ -243,9 +248,10 @@ fn exceeded_check(size: usize, allocated: usize) -> &'static str {
     }
 }
 
-impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
+impl<'a, A: Alarm<'a>, C: ProcessManagementCapability> ProcessConsole<'a, A, C> {
     pub fn new(
         uart: &'a dyn uart::UartData<'a>,
+        alarm: &'a A,
         tx_buffer: &'static mut [u8],
         rx_buffer: &'static mut [u8],
         queue_buffer: &'static mut [u8],
@@ -253,9 +259,10 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
         kernel: &'static Kernel,
         kernel_addresses: KernelAddresses,
         capability: C,
-    ) -> ProcessConsole<'a, C> {
+    ) -> ProcessConsole<'a, A, C> {
         ProcessConsole {
             uart: uart,
+            alarm: alarm,
             tx_in_progress: Cell::new(false),
             tx_buffer: TakeCell::new(tx_buffer),
             queue_buffer: TakeCell::new(queue_buffer),
@@ -265,6 +272,7 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
             rx_buffer: TakeCell::new(rx_buffer),
             command_buffer: TakeCell::new(cmd_buffer),
             command_index: Cell::new(0),
+            first_prompt_shown: Cell::new(false),
             running: Cell::new(false),
             execute: Cell::new(false),
             kernel: kernel,
@@ -276,11 +284,9 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
     /// Start the process console listening for user commands.
     pub fn start(&self) -> Result<(), ErrorCode> {
         if self.running.get() == false {
-            self.rx_buffer.take().map(|buffer| {
-                self.rx_in_progress.set(true);
-                let _ = self.uart.receive_buffer(buffer, 1);
-                self.running.set(true);
-            });
+            self.alarm
+                .set_alarm(self.alarm.now(), self.alarm.ticks_from_ms(100));
+            self.running.set(true);
         }
         Ok(())
     }
@@ -302,8 +308,10 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
         let _ = write(
             &mut console_writer,
             format_args!(
-                "Kernel version: {}\r\n",
-                option_env!("TOCK_KERNEL_VERSION").unwrap_or("unknown")
+                "Kernel version: {}.{} (build {})\r\n",
+                kernel::MAJOR,
+                kernel::MINOR,
+                option_env!("TOCK_KERNEL_VERSION").unwrap_or("unknown"),
             ),
         );
         let _ = self.write_bytes(&(console_writer.buf)[..console_writer.size]);
@@ -311,6 +319,7 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
         let _ = self.write_bytes(b"Welcome to the process console.\n");
         let _ = self
             .write_bytes(b"Valid commands are: help status list stop start fault process kernel\n");
+        self.prompt();
     }
 
     /// Simple state machine helper function that identifies the next state for
@@ -793,7 +802,7 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
                             let _ = write(
                                 &mut console_writer,
                                 format_args!(
-                                    "  {:?}\t{:<20}{:6}{:10}{:19}{:10}  {:?}{:5}/{}\n",
+                                    "  {:?}\t{:<20}{:6}{:10}{:17}{:10}  {:?}{:5}/{}\n",
                                     process_id,
                                     pname,
                                     process.debug_timeslice_expiration_count(),
@@ -809,6 +818,9 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
                             let _ = self.write_bytes(&(console_writer.buf)[..console_writer.size]);
                         }
                     });
+            }
+            WriterState::Empty => {
+                self.prompt();
             }
             _ => {}
         }
@@ -903,7 +915,7 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
                             });
                         } else if clean_str.starts_with("list") {
                             let _ = self.write_bytes(b" PID    Name                Quanta  ");
-                            let _ = self.write_bytes(b"Syscalls  Dropped Callbacks  ");
+                            let _ = self.write_bytes(b"Syscalls  Dropped Upcalls  ");
                             let _ = self.write_bytes(b"Restarts    State  Grants\n");
 
                             // Count the number of current processes.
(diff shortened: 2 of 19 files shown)


alexandruradovici:
As discussed yesterday in the core call, I updated the pull request by adding a `start_no_message` function that starts the process console without printing the "press any key" message. 

Taking into account @phil-levis's suggestion that printing the welcome message before the fist command might clutter the output, using the `start_no_message` function now prevents the display of that welcome message. It simply prints the prompt when a key is pressed.

This integrates #2793

capsules/src/process_console.rs
```diff
@@ -1147,28 +1185,39 @@ impl<'a, C: ProcessManagementCapability> uart::ReceiveClient for ProcessConsole<
...
                         }
-                    });
+                    } else {
+                        self.command_buffer.map(|command| {
+                            let index = self.command_index.get() as usize;
+                            if read_buf[0] == ('\n' as u8) || read_buf[0] == ('\r' as u8) {
+                                self.execute.set(true);
+                                let _ = self.write_bytes(&['\r' as u8, '\n' as u8]);
+                            } else if read_buf[0] == ('\x08' as u8) {
+                                if index > 0 {
```
hudson-ayers, line 1201: why move this if down to the next line if you don't do anything else outside of it?

alexandruradovici: I'm not sure I understand the question. Do you mean why index > 0 is not placed within the upper if?

hudson-ayers: Yeah that is what I mean. I only ask because I see you changed it but as far as I can tell the behavior is unchanged, so I wasn't sure if there was a mistake

alexandruradovici: Without the if, if a user presses backspace more than the number of characters he wrote, the console "deletes" whatever text was written before the user started writing the command. Most likely, he will detele the prompt text.

hudson-ayers: Oh, I understand now, sorry for the noise

bradjc:
Ok, perhaps philosophical question: does a shell start when the program starts executing, or when the user has some way of interacting with the shell?

I propose we say that the pconsole (and its prompt) should be shown when a user "connects" and is actually able to interact with the shell. For boards that use the CDC-USB stack, it would be feasibly to detect when a client connects and can interact with the user (in fact we already do this in the CDC driver). However, for normal UART connections we generally cannot do this, and perhaps having the user press a key is a good substitute.

So, I think that we should allow boards to choose if they need the "press a key..." message (which we have) and that they should also have an api for something like `user_connected_show_prompt()`. That second API could be added later.

I would like to remove the welcome message. Entering `help` is pretty easy if the user isn't sure what to do.

alexandruradovici:
> Ok, perhaps philosophical question: does a shell start when the program starts executing, or when the user has some way of interacting with the shell?
> 
> I propose we say that the pconsole (and its prompt) should be shown when a user "connects" and is actually able to interact with the shell. For boards that use the CDC-USB stack, it would be feasibly to detect when a client connects and can interact with the user (in fact we already do this in the CDC driver). However, for normal UART connections we generally cannot do this, and perhaps having the user press a key is a good substitute.
> 
Sounds like a good idea.

> So, I think that we should allow boards to choose if they need the "press a key..." message (which we have) and that they should also have an api for something like `user_connected_show_prompt()`. That second API could be added later.
> 
Do you have any idea of an API/way of doing this? Adding another function like `start_with_keypress`?

> I would like to remove the welcome message. Entering `help` is pretty easy if the user isn't sure what to do.

I deleted it.

alexandruradovici:
Now that 2.0 is out, @bradjc how do you think we should proceed with this?

bradjc:
I think that we need to decide if we want a shell prompt or not. It sounds like we do.

Then, I think this shell should work like normal shells. When it starts, you see a prompt. So, let's have `start()` start receiving characters and display the prompt.

To my last comment, boards should not call `start()` until there is something connected that can receive the prompt and use the shell. However, for UART serial channels there isn't a good way to do this, so I say we just stick with what we have now and have the bootup code call `start()`. For other channels like the USB-CDC channel, we can look into calling start when a user actually connects at some later date.

alexandruradovici:
> Then, I think this shell should work like normal shells. When it starts, you see a prompt. So, let's have `start()` start receiving characters and display the prompt.
> 
There is a small issue with this. The prompt will show before the `debug!` message. The only way that I can think of having a prompt after the `debug!` message is by using an alarm that fires after 100ms.

> To my last comment, boards should not call `start()` until there is something connected that can receive the prompt and use the shell. However, for UART serial channels there isn't a good way to do this, so I say we just stick with what we have now and have the bootup code call `start()`. For other channels like the USB-CDC channel, we can look into calling start when a user actually connects at some later date.
I think we need to connect the UART to the console, I'll think of something.

bradjc:
I'm still trying to get this PR merged. I think #2820 is a blocker for now, and depending on the outcome of that PR (adding a default deferred call for main.rs) I think it will be clear how to proceed with this PR.

phil-levis:
@bradjc and I talked about this for a while on the OT call today. The challenge is that, besides putting a very special side interface onto `Debug` for handling this special case, there doesn't seem to be a good solution. Waiting for all debug messages to flush will break if a capsule decides to emit periodic debug messages. Enforcing ordering at the virtualizer doesn't work because from its perspective the first `debug!` is the first request from `Debug`, and the rest are queued internally in `Debug`; therefore a call from the process console would appear to be next.

What we're trying to do is serialize/process in-order higher-level calls (e.g., `debug!') when these are actually one step away from where arbitration is between them and another writer (e.g, the process console). 

We briefly talked about an interface where a `Transmit` can tell the underlying implementation that it has N more bytes to send. Basically, this allows something that is actively transmitting to put itself again at the head of an empty queue, and keep on adding to its outstanding operation until another client also puts in a request, at which point its requests are rejected. E.g.:

Client A sends 10 bytes
  - send starts
Client A says it has 30 more bytes to send
  - queue marks A with 30 bytes pending
Client A says it has 20 more bytes to send
  - queue marks A with 50 bytes pending
Client B says it has 15 bytes to send
  - queue marks B with 15 bytes pending
Client A says it has 10 more bytes to send
  - queue rejects request
Client A finishes first 10 byte send, virtualizer sends another 50 bytes from A
Client A finishes 50 byte send, it tries to send 10 more, virtualizer sends 15 bytes from B
Client B finishes 15 byte send, virtualizer sends 10 bytes from A

This breaks virtualization, kinda. It seems very clunky and messy. For now, it seems like the best approach is just to wait for 100ms?

bradjc:
> Waiting for all debug messages to flush will break if a capsule decides to emit periodic debug messages.

I think we could add a version of this, with a new function `notify_when_current_queued_messages_are_flushed()` which issues a callback whenever however many bytes are in the queue at that point are flushed. This would permit implementing "wait until boot debug messages are transmitted". But that too has downsides: it requires debug to have its own dynamic deferred call in case the queue is already empty, and it imposes a perhaps strange dependency on the tock boot process for debug messages. Perhaps having this callback in the boot process is OK for process console, but it wouldn't be a good match for a generic "booted" callback, but if it is all that is available it might get used that way.

It's seeming more and more like just using a timer as a workaround is the best fix, for now at least.

phil-levis:
> It's seeming more and more like just using a timer as a workaround is the best fix, for now at least.

Yeah, it seems like these boot messages are a semantically special kind of `debug!`. If process console's requirement was a common one, then we might want to dig into a general mechanism to do this (or, if it becomes a common one, then we will want to make a general mechanism to do this).

bradjc:
Wait, what debug message is printing first?

We could add a timer, or we can just use the fact that debug is always added to the uart mux before pconsole, and let the ordering work to our advantage. If we are just adding the timer as a good enough fix, why not use the good enough fix we already have.

But looking at the code, I don't see why the prompt would print before any debug messages in main.rs.

phil-levis:
> We could add a timer, or we can just use the fact that debug is always added to the uart mux before pconsole, and let the ordering work to our advantage.

Super-brittle. We don't want to go down the path of initialization dependencies, especially manifested in assumptions about the queuing order based on initialization order: this leaks and becomes bound to two implementation details, the internal ordering of the queue and the queue servicing policy.

Phil

phil-levis:
> But looking at the code, I don't see why the prompt would print before any debug messages in main.rs.

It might suddenly do so if some implementation changes somewhere; we want to assume behavior based on interfaces not the particular implementation of those interfaces in the moment. I don't think we've ever viewed the strict priority queue of virtualizers as desirable, rather just a detail that hasn't been enough of a problem to necessitate a fix.

alexandruradovici:
The lack of ordering in the virtual UART is the reason that the code asks the user to press any key before displaying the prompt. I originally thought of using a timeout, but that would add code size to the process console as an additional virtual alarm and a client are required.

bradjc:
I'm not proposing we make this a tock precedent or a design decision we rely on, or that we make this a path we want to follow. This is simply a way to address this specific concern without overhead so we can make forward progress.

Having an interactive process console is a nice feature, but is it integral to Tock? I think this is the case where we can get away with implementation-level tricks (particularly since we don't have a principled design that addresses the issue).

bradjc:
Also, which debug message is printed first? I still don't see why this is an issue in practice.

phil-levis:
The process console shouldn't be depending on unpromised implementation details of another capsule; this creates a hidden dependency between implementations. This is a super-mega-bad anti-pattern. Someone shouldn't have to modify `main.rs` or the implementation of process console because the UART virtualizer implementation changed.

bradjc:
The UART console doesn't guarantee any ordering, and is already multiplexed in perhaps confusing ways. For example, if you run the number guessing game app (or any app which uses console receive) you will see all sorts of interleaved output between the app and the process console. I don't see this case of the shell prompt for process console as any different. Tock doesn't promise any ordering, it doesn't (currently) have any concept of "uart channels" and so things can be displayed in ways which might look strange. So I'm not proposing the process console is relying on anything, it is just using the shared UART in the standard way which (currently) doesn't provide any guarantees around ordering.

Perhaps what I am really saying is I see no issue with debug messages printing after the pconsole prompt. Since we have no good way to ensure that the pconsole prompt is always printed "last" (where it's hard to define what "last" specifically is, in the case of a capsule printing periodic debug info), then we shouldn't try, and we shouldn't promise that the pconsole will print "last". I.e., this is not an issue we should worry about or try to address.

But I am saying that I will just so happen to setup the boards I use so that the process console is added to the uart mux after debug, because, well, they have to be added in some order!

bradjc:
All that being said, I have no issue with adding a timer inside ProcessConsole, and effectively making the implementation of `start()` split-phase with a delay. I see that as a capsule-level decision, and wouldn't be unreasonable in this case.

phil-levis:
There was agreement on the core call today that the best answer is to add a timer to the process console, such that it prints its prompt a few tens or hundreds of milliseconds after it is started.

boards/clue_nrf52840/src/main.rs
```diff
@@ -690,10 +690,9 @@ pub unsafe fn main() {
         nrf52840::aes::AesECB<'static>
     ));
 
-    let pconsole =
+    let _pconsole =
```
bradjc, line 693: I don't think we need any of these board changes any more.

alexandruradovici:
I added an alarm that prints the prompt after 100ms.

boards/clue_nrf52840/src/main.rs
alexandruradovici: No we don't, I reverted the changes.

capsules/src/process_console.rs
```diff
@@ -192,6 +195,8 @@ pub struct ProcessConsole<'a, C: ProcessManagementCapability> {
     command_buffer: TakeCell<'static, [u8]>,
     command_index: Cell<usize>,
 
+    first_prompt_shown: Cell<bool>,
+
```
bradjc, line 199:
```suggestion
```

capsules/src/process_console.rs
```diff
@@ -1003,6 +1019,14 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
             command[0] = 0;
         });
         self.command_index.set(0);
+        if self.writer_state.get() == WriterState::Empty {
+            self.prompt();
+        }
+    }
+
+    fn prompt(&self) {
+        self.first_prompt_shown.set(true);
```
bradjc, line 1028:
```suggestion
```

capsules/src/process_console.rs
```diff
@@ -265,6 +272,7 @@ impl<'a, C: ProcessManagementCapability> ProcessConsole<'a, C> {
             rx_buffer: TakeCell::new(rx_buffer),
             command_buffer: TakeCell::new(cmd_buffer),
             command_index: Cell::new(0),
+            first_prompt_shown: Cell::new(false),
```
bradjc, line 275:
```suggestion
```

capsules/src/process_console.rs
alexandruradovici: Oops, forgot to remove it. :)

bradjc:
I compiled the hail kernel on this PR, and tried it out. This is what I saw:

```
[INFO   ]  ----- Waiting for serial port to reconnect...
Hi welcome to Tock. This test makes sure that a greater than 64 byte message can be printed.
Testing exit.
And a short message.
tock$ Exiting.
list
 PID    Name                Quanta  Syscalls  Dropped Upcalls  Restarts    State  Grants
  0	exit                     0        23                0         0  Terminated    0/14
  1	printf_long              0        17                0         0  Yielded    1/14
  2	blink                    0       957                0         0  Yielded    1/14
tock$ help
Welcome to the process console.
Valid commands are: help status list stop start fault process kernel
tock$ list
 PID    Name                Quanta  Syscalls  Dropped Upcalls  Restarts    State  Grants
  0	exit                     0        23                0         0  Terminated    0/14
  1	printf_long              0        17                0         0  Yielded    1/14
  2	blink                    0      1659                0         0  Yielded    1/14
tock$ fault printf_long
Process printf_long now faulted
tock$ Hi welcome to Tock. This test makes sure that a greater than 64 byte message can be printed.
And a short message.
list
 PID    Name                Quanta  Syscalls  Dropped Upcalls  Restarts    State  Grants
  0	exit                     0        23                0         0  Terminated    0/14
  3	printf_long              0        17                0         1  Yielded    1/14
  2	blink                    0      2712                0         0  Yielded    1/14
tock$
```

This may be ok, but I think we should be aware that the effect is that some process output appears to have the string "tock$" prepended to it.

alexandruradovici:
This is another reason why I initially used the "press any key approach".

bradjc:
All of these solutions are workarounds for the fact that we don't have true terminal multiplexing. I know we've been working on this for a while, but in my opinion at this point we should still with what we have. It works reasonably well, and doesn't add anything extra to the console display. The main downside is the lack of discoverability for the process console for users who don't know it is running but want to use it. I still find this a bit of a strange use case, as presumably whatever documentation alerted them to the process console being available could also instruct them on how to use it. Is there documentation we can make more clear somewhere? Also, board maintainers can add more `debug!()` messages if they so choose to add more instructions or alerting that the process console is there. I'm a huge proponent of staying friendly to new users, but we have to balance that with other users as well.

I'm not so opposed, however, that I will stand in the way. But someone else is going to have to approve+merge.

phil-levis (approved):

PR #2786, opened 2021-08-17
[imxrt1050-evkb] Fix GPIO interrupts

valexandru:
### Pull Request Overview

This pull request fixes gpio interrupts on the i.MX RT 1052 EVKB.

While I was testing the buttons example on the i.MX RT 1052 EVKB, I discovered that the interrupts are not enabled by default. However, if I set up a delay of 1000 ms so that a different interrupt is triggered before the gpio one, the gpio interrupts will work. Therefore, in order to prevent this, we have to specifically enable the interrupt in the nvic. 

This seems to be a bug in the i.MX RT 1052 EVKB board and requires to manually enable the interrupts in the nvic.

### Testing Strategy

This pull request was tested using an iMX. RT 1052 EVKB board.

### TODO or Help Wanted

This pull request needs feedback.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/imxrt10xx/src/lib.rs
+++ b/chips/imxrt10xx/src/lib.rs
@@ -232,4 +232,6 @@ pub unsafe fn init() {
     cortexm7::scb::set_vector_table_offset(
         &BASE_VECTORS as *const [unsafe extern "C" fn(); 16] as *const (),
     );
+
+    cortexm7::nvic::enable_all();
 }


hudson-ayers (dismissed):

bradjc (changes requested):
This seems bad. Why are the nvic interrupts not all enabled by default? Particularly why does the chip init look like:

https://github.com/tock/tock/blob/3c9468e522b8f67aca0ef91045e25a19fbb2fffb/chips/imxrt10xx/src/lib.rs#L228-L235

which does not enable all interrupts? Compare to the sam4l chip init:

https://github.com/tock/tock/blob/3c9468e522b8f67aca0ef91045e25a19fbb2fffb/chips/sam4l/src/lib.rs#L81-L85

Tock assumes all interrupts are enabled at the interrupt controller level, and relies on individual peripherals to disable interrupts on a per-peripheral basis. That avoids all issues like this. Did we need to special case this chip for some reason?

alexandruradovici:
I reversed the changes and enabled all interrupts. Please @valexandru  test the pio interrupt.

valexandru:
Yes, it works now, after enabling the interrupts in lib.rs. Thank you @alexandruradovici for making the changes while I was away.

hudson-ayers (approved):
bors r+

PR #2787, opened 2021-08-17
[spi] Update virtual_spi to report errors

alexandruradovici:
### Pull Request Overview

This pull request updates the virtual spi driver to report back errors. In case of errors, the virtual spi driver was setting the `op` variable to `ReadWriteDone` and was calling `do_next_op_async`. The next time `do_next_op` was run, it was not doing anything as it thought the mix was in flight.

It also:
- adds error reporting to the bus library
- fixes the screen driver to report errors
- updates the bus component to take the spi to accept the spi configuration as arguments

### Testing Strategy

This pull request was tested using a Pico Explorer Base

### TODO or Help Wanted

- [x] test using the Adafruit Clue nRF52840

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/virtual_spi.rs
+++ b/capsules/src/virtual_spi.rs
@@ -109,7 +109,6 @@ impl<'a, Spi: hil::spi::SpiMaster> MuxSpiMaster<'a, Spi> {
         } else {
             self.inflight.map(|node| {
                 let op = node.operation.get();
-                // Need to set idle here in case callback changes state
                 node.operation.set(Op::Idle);
                 match op {
                     // we have to report an error
--- a/capsules/src/virtual_spi.rs
+++ b/capsules/src/virtual_spi.rs
@@ -148,7 +148,7 @@ impl<'a, Spi: hil::spi::SpiMaster> DynamicDeferredCallClient for MuxSpiMaster<'a
     }
 }
 
-#[derive(Copy, Clone, PartialEq, Debug)]
+#[derive(Copy, Clone, PartialEq)]
 enum Op {
     Idle,
     ReadWriteBytes(usize),
--- a/capsules/src/virtual_spi.rs
+++ b/capsules/src/virtual_spi.rs
@@ -108,8 +108,7 @@ impl<'a, Spi: hil::spi::SpiMaster> MuxSpiMaster<'a, Spi> {
             });
         } else {
             self.inflight.map(|node| {
-                let op = node.operation.get();
-                match op {
+                match node.operation.get() {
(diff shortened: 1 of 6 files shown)


capsules/src/virtual_spi.rs
```diff
@@ -106,6 +106,22 @@ impl<'a, Spi: hil::spi::SpiMaster> MuxSpiMaster<'a, Spi> {
                     Op::Idle => {} // Can't get here...
                 }
             });
+        } else {
+            self.inflight.map(|node| {
+                let op = node.operation.get();
+                // Need to set idle here in case callback changes state
+                node.operation.set(Op::Idle);
```
bradjc, line 113: How can this unconditionally set the inflight operation to idle? What if two SPI clients call `read_write_bytes()` back-to-back, and `do_next_op()` is called for the second time?

alexandruradovici: If I'm not mistaking, operation cat be only two values at this point:
- `ReadWriteDone` if this was called from `command_complete`
- `ReadWriteDone` if this is called due to a an error before an action can be started
- `Idle` if this is called when another client tries to schedule an action

But I might be missing something.

capsules/src/virtual_spi.rs
```diff
@@ -106,6 +106,22 @@ impl<'a, Spi: hil::spi::SpiMaster> MuxSpiMaster<'a, Spi> {
                     Op::Idle => {} // Can't get here...
                 }
             });
+        } else {
+            self.inflight.map(|node| {
+                let op = node.operation.get();
+                // Need to set idle here in case callback changes state
```
alexandruradovici, line 112:
```suggestion
```

capsules/src/virtual_spi.rs
alexandruradovici:
```suggestion
```

capsules/src/virtual_spi.rs
```diff
@@ -106,6 +106,20 @@ impl<'a, Spi: hil::spi::SpiMaster> MuxSpiMaster<'a, Spi> {
                     Op::Idle => {} // Can't get here...
                 }
             });
+        } else {
+            self.inflight.map(|node| {
+                let op = node.operation.get();
+                match op {
```
alexandruradovici, line 112:
```suggestion
                match node.operation.get() {
```

hudson-ayers (approved):

bradjc (approved):

lschuermann (approved):

PR #2794, opened 2021-08-21
capsules: update touch driver

alexandruradovici:
### Pull Request Overview

This pull request changes the touch driver to report the first multi touch as a single touch. This is useful for applications that only know how to use single touch. 

Applications that did not require multi touch would have to use the multi touch interface if the touch panel hardware was able to provide multiple touches. The single touch interface would not fire touches.

### Testing Strategy

This pull request was tested using an stm32f412g discovery.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/touch.rs
+++ b/capsules/src/touch.rs
@@ -102,18 +102,20 @@ impl<'a> Touch<'a> {
                 break;
             }
         }
-        self.touch.map_or(
-            // if there is no single touch device
-            // try to use the multi touch device and
-            // report only a single touch
-            self.multi_touch
-                .map_or(Err(ErrorCode::NODEVICE), |multi_touch| {
-                    if enabled {
-                        multi_touch.enable()
-                    } else {
-                        multi_touch.disable()
-                    }
-                }),
+        self.touch.map_or_else(
+            || {
+                // if there is no single touch device
+                // try to use the multi touch device and
+                // report only a single touch
+                self.multi_touch
+                    .map_or(Err(ErrorCode::NODEVICE), |multi_touch| {
+                        if enabled {
+                            multi_touch.enable()
+                        } else {
+                            multi_touch.disable()
+                        }
+                    })
+            },
             |touch| {
                 if enabled {
                     touch.enable()
--- a/capsules/src/touch.rs
+++ b/capsules/src/touch.rs
@@ -17,7 +17,7 @@ use core::mem;
 use kernel::grant::Grant;
 use kernel::hil;
 use kernel::hil::screen::ScreenRotation;
-use kernel::hil::touch::{GestureEvent, TouchEvent, TouchStatus};
+use kernel::hil::touch::{GestureEvent, TouchClient, TouchEvent, TouchStatus};
 use kernel::processbuffer::{ReadWriteProcessBuffer, WriteableProcessBuffer};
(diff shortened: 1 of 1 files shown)


capsules/src/touch.rs
```diff
@@ -102,13 +102,26 @@ impl<'a> Touch<'a> {
...
+            // try to use the multi touch device and
+            // report only a single touch
+            self.multi_touch
+                .map_or(Err(ErrorCode::NODEVICE), |multi_touch| {
+                    if enabled {
+                        multi_touch.enable()
+                    } else {
+                        multi_touch.disable()
+                    }
+                }),
```
hudson-ayers, line 116: The first argument passed to `Option::map_or()` is eagerly evaluated (see https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or), so this code will run regardless of whether `touch.is_some()`. To get the behavior I think you want here, you should use `map_or_else()`.

hudson-ayers (commented):
PSA, easy to review with whitespace changes hidden

capsules/src/touch.rs
jrvanwhy: I think the documentation for `map_or` needs to be fixed. The closure is *not* eagerly evaluated (see [the implementation](https://doc.rust-lang.org/src/core/option.rs.html#514)), so this will work as intended. The "default" argument *is* eagerly evaluated in that it is computed before `map_or` is invoked.

hudson-ayers: For the outer 'map_or()' call ('self.touch.map_or()'), the "default argument" in this code is an expression with side effects. So the default argument being evaluated before the call to map_or makes those side effects happen regardless. Unless I am misunderstanding?

jrvanwhy: Oh, I didn't see the outer `map_or` call. Disregard my previous comment.

alexandruradovici: I made a beginner's mistake 😄  I absolutely overlooked the fact that the function call is a parameter.

capsules/src/touch.rs
```diff
@@ -203,71 +218,76 @@ impl<'a> hil::touch::MultiTouchClient for Touch<'a> {
...
         };
-        // debug!("{} touch(es)", len);
-        for app in self.apps.iter() {
-            app.enter(|app, upcalls| {
-                if app.ack {
-                    app.dropped_events = 0;
+        if len > 0 {
+            // report the first touch as single touch
+            // for applications that only use single touches
+            self.touch_event(touch_events[0]);
```
bradjc, line 224: Wouldn't it be odd for a multi touch app to get a single touch callback (as well as the multi touch version)?

alexandruradovici: Apps subscribe to different upcall numbers for touch and multitouch. The touch driver usually has only one of the two underlying drivers: either single touch either multitouch.

Apps that need only one touch would have to check the number of touches and subscribe to different upcalls accordingly. This allows apps that only need a single touch to subscribe to the single touch upcall regardless of what kind of hardware the touch driver uses.

bradjc: Gotcha.

bradjc (approved):

hudson-ayers (approved):

PR #2797, opened 2021-08-24
ProcessSlice augmentation: add From<> [u8] impls and express chunks() iterator as &ProcessSlice instead of &[Cell<u8>]

jettr:
### Pull Request Overview

Drivers that should operate on either ...`ProcessSlice` buffers or `u8` slices
should be able to single source their implementations. To facilitate
this, add a safe conversion from a `u8` slice into the appropriate
...`ProcessSlice` structure. This limits how you can access the `u8` slice, but
every ...`ProcessSlice` accessor is still safe to perform on a normal slice.

### Testing Strategy

Manually verified that rust lifetime rules apply correctly. For example, the follow code snippet generates a compiler error

```rust
fn test(buf: &mut [u8]) {
   let slice: WriteableProcessSlice = buf.into();
   buf[0] = 1;
   slice[0].set(2);
}
```
This above snippet complains about multiple mutable references: `error[E0503]: cannot use *buf because it was mutably borrowed`

### Documentation Updated

- None

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/processbuffer.rs
+++ b/kernel/src/processbuffer.rs
@@ -601,6 +601,28 @@ pub struct ReadableProcessSlice {
     slice: [ReadableProcessByte],
 }
 
+// Allow a u8 slice to be viewed as a ReadableProcessSlice to allow client code to be
+// authored once and accept either [u8] or ReadableProcessSlice
+impl<'a> From<&'a [u8]> for &'a ReadableProcessSlice {
+    fn from(val: &'a [u8]) -> Self {
+        // SAFETY: The layout of a [u8] and ReadableProcessSlice are guaranteed to be the same.
+        //         This also extends the lifetime of the buffer, so aliasing rules are
+        //         thus maintained properly.
+        unsafe { core::mem::transmute(val) }
+    }
+}
+
+// Allow a mutable u8 slice to be viewed as a ReadableProcessSlice to allow client code to be
+// authored once and accept either [u8] or ReadableProcessSlice
+impl<'a> From<&'a mut [u8]> for &'a ReadableProcessSlice {
+    fn from(val: &'a mut [u8]) -> Self {
+        // SAFETY: The layout of a [u8] and ReadableProcessSlice are guaranteed to be the same.
+        //         This also extends the mutable lifetime of the buffer, so aliasing rules are
+        //         thus maintained properly.
+        unsafe { core::mem::transmute(val) }
+    }
+}
+
 impl ReadableProcessSlice {
     /// Copy the contents of a [`ReadableProcessSlice`] into a mutable
     /// slice reference.
@@ -717,6 +739,17 @@ impl Index<usize> for ReadableProcessSlice {
 #[repr(transparent)]
 pub struct WriteableProcessSlice {
     slice: [Cell<u8>],
+}
+
+// Allow a mutable u8 slice to be viewed as a WritableProcessSlice to allow client code to be
+// authored once and accept either [u8] or WriteableProcessSlice
+impl<'a> From<&'a mut [u8]> for &'a WriteableProcessSlice {
+    fn from(val: &'a mut [u8]) -> Self {
+        // SAFETY: The layout of a [u8] and WriteableProcessSlice are guaranteed to be the same.
+        //         This also extends the mutable lifetime of the buffer, so aliasing rules are
+        //         thus maintained properly.
+        unsafe { core::mem::transmute(val) }
+    }
 }
 
 impl WriteableProcessSlice {
--- a/kernel/src/processbuffer.rs
+++ b/kernel/src/processbuffer.rs
@@ -601,6 +601,15 @@ pub struct ReadableProcessSlice {
     slice: [ReadableProcessByte],
 }
 
+fn cast_byte_slice_to_process_slice<'a>(
+    byte_slice: &'a [ReadableProcessByte],
+) -> &'a ReadableProcessSlice {
+    // As ReadableProcessSlice is a transparent wrapper around its inner type,
+    // [ReadableProcessByte], we can safely transmute a reference to the inner type as a reference
+    // to the outer type with the same lifetime
+    unsafe { core::mem::transmute::<&[ReadableProcessByte], &ReadableProcessSlice>(byte_slice) }
+}
+
 // Allow a u8 slice to be viewed as a ReadableProcessSlice to allow client code to be
 // authored once and accept either [u8] or ReadableProcessSlice
 impl<'a> From<&'a [u8]> for &'a ReadableProcessSlice {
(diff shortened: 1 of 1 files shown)


lschuermann:
This is pretty cool.

I'm slightly afraid this might be a little confusing and it might not have a ton of utility: we can't make asynchronous drivers take `ProcessSlice`s of non-static lifetimes, given they need to hold on to the exclusive mutable reference during the entire asynchronous operation and then pass it back to the client. This is impossible to get with a proper  process buffer.

Thus this really is only useful for drivers which are synchronous (i.e. take a non-static slice reference today). As outlined in the HIL TRD, usually HILs will be the layer which short-lived `ProcessSlice` can't cross. Thus, writing hardware drivers taking `ProcessBuffer`s won't be very useful.

jettr:
> might not have a ton of utility

I am already using in downstream code :)

We have a USB hil that let's you transfer a buffer (`&[u8]`) of data. Sometime this data comes from other kernel level capsules (in which case they have reference to  a `u8` slice); other times this USB data comes from an application that handles the logical USB data packets. We need the ability to transfer both a `&[u8]` or a `ReadableProcessSlice`. As an implementation note, the usb capsule will then copy the data from in the input parameter into the correct static circular buffer for processing.

We have changed our usb capsule to take in an `Into<ReadableProcessSlice>` parameter and use only the more limited `ReadableProcessSlice` API, and that allows us to handle either type without copying code.

Additionally, there are hils in upstream Tock OS right now, that take in a `&[u8]`, namely `digest.rs` and `symmetric_encryption.rs`. It is very conceivable that those hils would need to take in key information from an application buffer in addition to a kernel buffer.

lschuermann:
Exactly. I think this is fine to include if your not using it in downstream code, but for asynchronous HILs it wouldn't really work (as you've determined `digest` and `symmetric_encryption` could work, but those might well be changed to work asynchronously in the future).

jettr:
Another thing that might help clarity is to actually rename `WriteableProcessSlice` => `WriteableSlice` and `ReadableProcessSlice` => `ReadableSlice` (and same for `Byte` variants too).  At the point that objects are in a ..`Slice`, they are no longer tied to a process (at least in terms of their API). Where as `ReadableProcessBuffer` is very much tied to a process and should keep process in the name.

I think the rename refactor adds clarity along with this change. Since it will touch more files, I would to see what you think first. Let me know if you want me to apply the above rename.

Just so I understand, you are okay with merging this change (or a derivative of this change) to support the downstream use case, correct?

jettr:
Do y'all mind taking another pass on this PR, please? Are there any outstanding concerns?

kernel/src/processbuffer.rs
```diff
@@ -601,6 +601,37 @@ pub struct ReadableProcessSlice {
...
+    unsafe { core::mem::transmute::<&[ReadableProcessByte], &ReadableProcessSlice>(byte_slice) }
+}
+
+// Allow a u8 slice to be viewed as a ReadableProcessSlice to allow client code to be
+// authored once and accept either [u8] or ReadableProcessSlice
+impl<'a> From<&'a [u8]> for &'a ReadableProcessSlice {
+    fn from(val: &'a [u8]) -> Self {
+        // SAFETY: The layout of a [u8] and ReadableProcessSlice are guaranteed to be the same.
+        //         This also extends the lifetime of the buffer, so aliasing rules are
+        //         this maintained properly.
```
hudson-ayers, line 619:
```suggestion
        //         thus maintained properly.
```

hudson-ayers (approved):

kernel/src/processbuffer.rs
jettr: Done.

hudson-ayers (approved):

jettr:
An API can still be asynchronous when taking in a non static buffer (`&[u8]` or `ProcessSlice`); the internal API just has to copy it to its own static buffer before returning from the call. It doesn't have to transmit it (or do the thing) until later.

Having the ability to convert `&[u8]` into `ProcessSlice` won't make people start writing async drivers that take static buffers differently. It will just allow APIs that would want to take in a non-static `&[u8]` also have an easy way to single-source internal logic if it also need to take in a `ProcessSlice` version as well.

Looking at the in-tree cyrpto capsule that takes in a non-static buffer of key material, one could imagine that this  key material could come from kernel or users space. Setting the key material isn't the main async operation either. The main async operation would be the encrypt or decrypt, which could take in the static buffer.

I am not suggesting that after this change we start converting capsules to start taking `ProcessSlice`s instead of static buffers. Adding these conversion just makes it possible for people to single-source logic when needed for their use case.

jettr:
Any more outstanding concerns?

bradjc:
Am I understanding correctly then that the intent is that developers would _not_ write something like `pub fn do_action_with_buffer(&self, buf_or_process_slice: ProcessSlice)`? And that this should only be used in non-public APIs to simplify implementations or reduce code duplication?

jettr:
> Am I understanding correctly then that the intent is that developers would _not_ write something like `pub fn do_action_with_buffer(&self, buf_or_process_slice: ProcessSlice)`? And that this should only be used in non-public APIs to simplify implementations or reduce code duplication?

I think it depends on the public API boundary (I am reading that as hil/capsule API) you are going for. If you want your API to take in a buffer that could come from kernel or userspace and that API function did an internal copy, then it could be appropriate to expose `pub fn do_action_with_slice(&self, process_slice: ProcessSlice)` and `pub fn do_action_with_buffer(&self, buf: &[u8])` and have the `buffer` version internally call the `slice` version.

bradjc:
Great. Not approving since I don't think I can evaluate the safety concerns, but looks good to me.

lschuermann (approved):
I had a very careful pass over these changes. In general they look correct to me.

There is one thing which is pretty alien in a Rust setting: we can now turn a `&'a mut [u8]` reference into a `&'a WriteableProcessSlice`. Because this is an immutable reference it does not have to be an exclusive reference, thus we loose the exclusiveness of the data pointed to by the original `&'a mut [u8]`. Now, because we only access contained data through `Cell<u8>` types and the `WriteableProcessSlice::from(&'a mut [u8])` mutably borrows the original buffer for the entire lifetime of the `WriteableProcessSlice` such that the original can't be accessed any longer, this looks sound -- but at the same time pretty confusing.

Even being correct, however, these changes do bring a fair share of complexity with them. It's much easier to reason about the types and their relationships when they can only originate from `ProcessBuffer`s. Future changes to these types would need to be very careful to respect that the slice types can also at some point originate from regular borrowed Rust slices.

Thus in their current form, I'd say these changes look correct. As for their utility in the Tock ecosystem, I'm indifferent about that. Apparently there are some use cases already, but those are in stark contrast with how most defined interface layers (HILs and other interfaces) will work in Tock, given this is really only compatible with asynchronous interfaces when the lower-level implementor of the interface performs local copying. We'll want to make sure this isn't a generally endorsed paradigm.

jrvanwhy (approved):
Looks sound to me!

kupiakos:
Sorry, but the change as is is not sound 😢: https://github.com/tock/tock/issues/2882

jettr:
This PR did replicate the UB that is also present in `raw_processbuf_to_roprocessslice` in a way that we could more easily verify with a unit test. If we can eradicate the UB in `raw_processbuf_to_roprocessslice`, then we can re-use that in the conversion function that was added. We can continue to discuss the UB in the issue you opened: #2882

Issue #2593, opened 2021-05-22
tock-registers: Make Register Access Unsafe


Cryptjar:
_**TL;DR** IMO the current `get` and `set` functions are potentially unsound, thus I propose to declare them all as `unsafe` or at least allowing to do so optionally._

First of all, let me say, that I'm a user of the `tock-registers` crate as it is on crates.io, I do not have any insights into Tock itself, and I'm arguing here only in regard to `tock-registers` as a library.

I'm aware of the recent changes due to #2517 and the current release tracking issue #2579, actually, that is partially why I come up with this now. In my personal opinion, this (#2517) is going in a very good direction. But I believe, we can do even better. And since, there is already an upcoming breaking release planed, why not include this too, so only a single breaking release has to be done. I.e., I propose this to be included in #2579.

Motivation
----------

One of the most magnificent features of Rust (IMO) is the distinction between `unsafe` Rust and Safe Rust. This allows Safe Rust to invoke any safe function in any way (that the type system permits), and it will always be _sound_, i.e., it will never invoke _Undefined Behavior_.

Going directly to MMIO, there are at least two things that must be upheld when writing or reading to/from an MMIO memory address:
	1) that an address (i.e. a pointer) must be valid as required by Rust and points to the intended device
	2) the respective device and system must be in a valid state, MMIO is _no memory_ and each access individually must be valid

In my opinion, the first point has been already dealt with (i.e., requiring the constructor to ensure this contract is upheld, so the methods can rely upon that 1) is upheld). However, to the best of my knowledge, the second one is yet not handled satisfactorily (IMO), i.e. practically the constructor also has to ensure that all subsequent read & writes (in any order and with any value) will be valid, which practically is borderline impossible (depending on the device of course).

To illustrated let me give you a practical example that I recently encountered:

I'm currently working with a Raspberry Pi 4 board, and it has (like most others) an UART, in particular a so call 'Mini UART', which, for reference, is specified in Chapter 2 of the [manual](https://datasheets.raspberrypi.org/bcm2711/bcm2711-peripherals.pdf#chap-Auxiliaries). In particular, it has the `AUX_ENABLES` register besides the other UART access registers. One paragraph in its specification caught my attention:

> If the enable bits are clear\[ed\] you will have **no access** to a peripheral.
> You can not even read or write the registers!

[...]

PR #2800, opened 2021-08-27
rfc: add low level driver capability

alexandruradovici:
### Pull Request Overview

This pull request adds the `LowLevelDriverCreationCapability`. While talking a look at Tock's source together with an industry partner, we realized that there is nothing much stopping a capsule from creating a certain peripheral and using it directly. Capsules would not be able to use interrupts, as they would not be registered with the kernel's infrastructure, but could use any synchronous interfaces.

So far I have modified the Raspberry Pi Pico just as a proof of concept.

### Testing Strategy

N/A

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nano_rp2040_connect/src/io.rs
+++ b/boards/nano_rp2040_connect/src/io.rs
@@ -1,6 +1,8 @@
 use core::fmt::Write;
 use core::panic::PanicInfo;
 
+use kernel::capabilities::LowLevelDriverCreationCapability;
+use kernel::create_capability;
 use kernel::debug::{self, IoWrite};
 use kernel::hil::led::LedHigh;
 use kernel::hil::uart::{Configure, Parameters, Parity, StopBits, Width};
@@ -46,7 +48,8 @@ impl IoWrite for Writer {
         self.uart.map_or_else(
             || {
                 // If no UART is configured for panic print, use UART0
-                let uart0 = &Uart::new_uart0();
+                let capability = create_capability!(LowLevelDriverCreationCapability);
+                let uart0 = &Uart::new_uart0(&capability);
 
                 if !uart0.is_configured() {
                     let parameters = Parameters {
@@ -59,8 +62,8 @@ impl IoWrite for Writer {
                     //configure parameters of uart for sending bytes
                     let _result = uart0.configure(parameters);
                     //set RX and TX pins in UART mode
-                    let gpio_tx = RPGpioPin::new(RPGpio::GPIO0);
-                    let gpio_rx = RPGpioPin::new(RPGpio::GPIO1);
+                    let gpio_tx = RPGpioPin::new(RPGpio::GPIO0, &capability);
+                    let gpio_rx = RPGpioPin::new(RPGpio::GPIO1, &capability);
                     gpio_rx.set_function(GpioFunction::UART);
                     gpio_tx.set_function(GpioFunction::UART);
                 }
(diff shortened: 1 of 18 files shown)


hudson-ayers:
Is there a specific example of where creating two instances of a peripheral would be unsound (as opposed to just leading to unexpected behavior from the peripheral)?

In general it seems that the issue with something like a DMA peripheral would be that calling set on a register is safe even if modifying that register leads to changing arbitrary memory based on the values of other registers. In which case the fix there would be modifying the Tock-registers library to support marking certain register operations as unsafe when declaring a registers struct

lschuermann:
> In general it seems that the issue with something like a DMA peripheral would be that calling set on a register is safe even if modifying that register leads to changing arbitrary memory based on the values of other registers.

Yes. We start getting into dangerous territory in the Rust memory-safety sense as soon as we're controlling hardware which can mutate memory outside of the peripheral's memory-mapped address space. For any such peripheral I would assume that arbitrarily bad things happen once we have two drivers which think they have exclusive control over said peripheral. We can't reasonably assume any correct behavior whatsoever beyond this point.

> In which case the fix there would be modifying the Tock-registers library to support marking certain register operations as unsafe when declaring a registers struct.

I disagree. This goes in the exact same direction as #2593. At the end of that extensive discussion it seems to make the most sense to stick with the  "unsafe creation, safe use" paradigm. We explicitly put the burden on the user, when creating the `tock-registers` type (e.g. through `StaticRef::new`), to ensure that Rusts memory safety requirements will not be violated.

It also wouldn't make much sense semantically: from `tock-registers`' point of view, if we're talking about valid accesses to writable registers of DMA peripherals, the accesses themselves are entirely sound operations, no damage done here. The consequences stem from the "bigger picture", where the particular sequence of register access might lead to unsoundness. This is best gated through an `unsafe` on the high-level instance managing this sequence of accesses, i.e. the driver.

hudson-ayers:
The reason we publicly export the `StaticRef` in places where the constructor takes in a `StaticRef` is so that you can create peripherals from a board crate if you do not want to use the default set of peripherals for a given chip. You cannot mark a `const` as `unsafe` to use, unfortunately.

We could mark all of the peripheral constructors as `unsafe`, which I would consider preferable to using a capability. However in many cases I think that having multiple instances of a peripheral would not necessarily lead to memory unsafety -- it is more specific to certain peripherals which let you do DMA-like stuff.

lschuermann:
> The reason we publicly export the `StaticRef` in places where the constructor takes in a `StaticRef` is so that you can create peripherals from a board crate if you do not want to use the default set of peripherals for a given chip. You cannot mark a `const` as `unsafe` to use, unfortunately.

Can't we just simply, instead of exporting the constructed `StaticRef`, export a typed pointer to this struct? This way we still have some degree of type safety and annotation (which of course in the case of a pointer can be easily overriden through casting) but would delay the unsafety to the construction of the `StaticRef` either in the default peripherals struct (creating this would need to become unsafe then) or the custom periperhals struct in the board file.

hudson-ayers:
> Can't we just simply, instead of exporting the constructed `StaticRef`, export a typed pointer to this struct? This way we still have some degree of type safety and annotation (which of course in the case of a pointer can be easily overriden through casting) but would delay the unsafety to the construction of the `StaticRef` either in the default peripherals struct (creating this would need to become unsafe then) or the custom periperhals struct in the board file.

Yes, I think that would work fine.

bradjc (changes requested):
I don't think this is an appropriate use of capabilities. Any crate that can create chip drivers (scientifically the chip itself and boards) can also call unsafe and create its own capability at will. That eliminates the benefit of capabilities (i.e. that they are enforced by the compiler).

PR #2633, opened 2021-06-26
CI Hardware Documentation Creation

AnthonyQ619:
### Pull Request Overview

This draft pull request created the document "CI_Hardware" and added folder "images" to hold images for any markdown files that need figures of any kind.

### TODO or Help Wanted

This draft pull request still needs any review on the documentation "CI_Hardware" if wording was not clear, some definitions can be defined, or order of topics needs to be structured better.

### Documentation Updated

- [ ] Updated "/doc" with the creation of "CI_Hardware" and folder "images" to hold images

### Formatting

- [ ] Ran `make prepush`.

----

[Rendered](https://github.com/AnthonyQ619/tock/blob/aq-branch/doc/CI_Hardware.md)

PR #2801, opened 2021-08-27
Add LiteX simulator based CI

lschuermann:
### Pull Request Overview

This adds a CI workflow using the LiteX Verilated simulation to run various libtock-c apps.

In contrast to unit tests or using emulators such as QEMU, this uses a proper simulation on the gate level (compiled to C++ through Verilator) and thus is a cycle-accurate simulation of a board with a VexRiscv CPU. This comes at a significant performance cost, but might allow us to more accurately test timing-critical components such as emulated SPI controllers and peripherals.

It uses Tockloader to load apps onto a flash file together with a kernel and checks for expected output. Interaction with the simulation can be done via the console, Ethernet, or a ZeroMQ/JSON-based protocol for talking to simulation modules such as GPIO.

This workflow includes builds of the Tock kernel, libtock-c, various LiteX components as described in the tock-litex companion repository, Tockloader, etc. Thus it takes longer than other workflows and is expected to be more fragile than the regular tock-ci. Therefore, this is split into an entirely different workflow file. It does however serve as a good integration test for the larger Tock ecosystem and development workflows (e.g. using Tockloader).

A long-term goal is to eventually include all reasonable tests from the release test suite.

Given I expect this to be more fragile than the regular CI, I propose we -- at first -- don't make this a required check for Bors.

### Testing Strategy

This pull request was tested by ... testing?

### TODO or Help Wanted

- [x] Pin a `tock-litex` release.
- [x] Pin a `tockloader` release.
- [x] Pin a global `libtock-c` targeted revision for all integration tests in the Tock repository, which we can centrally update?
- [x] Wait for the `litex_sim` GPIO support to be upstream, PR enjoy-digital/litex#1035.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/tools/litex-ci-runner/src/main.rs
+++ b/tools/litex-ci-runner/src/main.rs
@@ -93,7 +93,8 @@ where
             --with-gpio \
             --rom-init {} \
             --non-interactive \
-            --with-simctrl",
+            --with-simctrl \
+            --no-compile-software",
             flash_file.to_string_lossy(),
         ),
         // The initial build might take a while, which is why we're


alistair23 (approved):
WOOOOOO!!!!!!!!!!!!!!!!!!!!!!!!!

alistair23:
I think a required check is nice. It's nice to see a failure straight away instead of waiting for a review and bors to see a failure.

7 minutes is nothing, it's a good excuse to get up and walk around :) Plus it's unlikely this will fail that often anyway as generally the format and build are the main failures

tools/litex-ci-runner/src/main.rs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+fn run() -> Result<(), Error> {
+    println!("litex-sim CI runner starting...");
+    println!("");
+
+    // Shortcut function, with the kernel binary path and the
+    // libtock-c path prefix fixed and prepended to the
+    // application. Always expects the kernel message first.
+    fn tmpl_run(
```
bradjc, line 108: "tmpl"? What is that?

tools/litex-ci-runner/src/main.rs
```diff
@@ -0,0 +1,264 @@
...
+    // This is expected to fail. The console RX is not virtualized
+    // yet, and thus one of the apps trying to receive over the
+    // console will get an error from the kernel.
+    tmpl_run(
+        &["tests/console_recv_short", "tests/console_recv_long"],
+        |p| {
+            p.exp_regex("\\[(SHORT|LONG)\\] Error doing UART receive: -2")?;
+            Ok(())
+        },
+    )?;
```
bradjc, line 252: I don't think we should include a test that would cause a PR that actually fixes this problem to fail.

.libtock_c_ci_rev
```diff
@@ -0,0 +1,8 @@
+# Pins a libtock-c revision for Tock CI tests. This makes it easy to
+# update the targeted libtock-c revision for all of Tock's CI jobs.
+#
+# This file is a valid shell file and can be sourced to obtain the
+# LIBTOCK_C_CI_REV environment variable.
+
+# libtock-c of Aug 25, 2021, 6:37 PM GMT+2
+LIBTOCK_C_CI_REV=17601d28fa3652a124abf373c86143c2532d3f3c
```
bradjc, line 8: Why do we need this hidden file?

bradjc (changes requested):
As with all new CI, let's make it non mandatory, see how it fairs for ~month and re-evaluate. The problem is the tests that can run on non-physical boards are very limited and generally easy to test. How often is a bug introduced in the console capsule, for example? So this is certainly a step in the right direction, but I think we do need to be realistic about how much of the tock kernel we can test on virtual platforms.

tools/litex-ci-runner/src/main.rs
lschuermann: I included this so it would fail and we can think of a proper test mechanism for the then fixed version. So it's essentially a reminder that we should write tests for it.

.libtock_c_ci_rev
lschuermann: See here: https://github.com/tock/tock/pull/2801#discussion_r697457888

tools/litex-ci-runner/src/main.rs
lschuermann: That's bad naming. It's a function which performs some basic checks and takes libtock-c app paths relative to the examples directory. We can rename it to "examples_test" or something along those lines.

.github/workflows/litex_sim.yml
```diff
@@ -0,0 +1,130 @@
...
+
+      # TODO: Should we pin a specific targeted libtock-c version for all CI in
+      # Tock or always refer to the master? It'd probably be good if CI is
+      # somewhat reproducible and doesn't change in behavior because of changes
+      # to another repo.
+      - name: Checkout libtock-c
+        uses: actions/checkout@v2
+        with:
+          repository: tock/libtock-c
+          ref: master
```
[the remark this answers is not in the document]
bradjc: I think it would be ok if different CI passes uses different libtock-c commits. Anyone adding a new test workflow might not want to verify that all other test workflows work on the same libtock-c, or fix (and wait for the fixes to be merged) libtock-c to make sure all can use the same libtock-c commit.

But, can we at least make the file not hidden and put it in the .github folder? Anything we add to the root pushes the readme farther down the fold on the github page.

tools/litex-ci-runner/src/main.rs
bradjc: I guess I don't understand but if you think this is important to track then 👍

lschuermann:
Sorry this has been open for so long already. I see there has been a new Tockloader release out, so I will pin that here. Furthermore, I've managed to add GPIO support in the LiteX simulation, which can controlled from the test runner. This allows to run a much more extensive test suite, hopefully almost all of the standard release tests. I have it all running locally but conditionally integrating GPIO into the litex_sim boards using `#[cfg(`s is a mess. So I'd prefer to block this on enjoy-digital/litex#1035, so that our LiteX supported boards then have proper GPIO support.

ppannuto (changes requested):
I'm wondering if this is perhaps the wrong approach for integrating LiteX CI [and possibly the QEMU CI as well]. LiteX is hardware platform, which yes we happen to be able to run in software, but that is incidental. With #2633 and the follow-on work coming there, we are building a generic mechanism to support externally-run CI of hardware platforms. In this context, @lschuermann would have a "physical copy" of the LiteX "board" that runs the LiteX CI.

This alleviates all of the complex setup identified here ('installing LiteX environment on local user machines too invasive'), and also lets updates to the LiteX platform go through without blocking on PRs to Tock proper (i.e. avoiding the series of 'update hash' commits we have been managing for the QEMU experiment). 

I think I am particularly hung up on the development of another test-runner setup. We already have one custom test runner with QEMU, now we have another one coming in here, and then there's the third one coming down for generic hardware platforms.

I would be interested to see if we couldn't treat LiteX as an opaque hardware platform that uses the same hardware CI infrastructure as any other platform.

lschuermann:
@ppannuto In general I would agree. Hardware CI should probably be done through the hardware CI interface developed as documented in #2633.

However I really wouldn't treat the LiteX simulation as a piece of hardware. Incidentally it runs the same softcore CPU and looks very similar to the LiteX Arty target, and thus is a reasonably close approximation to assert that core functionality of Tock -- specifically RISC-V processes, syscalls, the PMP and some capsules -- work as expected. Also the bitstream updates (in this case the release of the tock-litex companion repository) would still need to happen, as the "hardware" does evolve and our model in Tock needs to reflect that.

When CI is integrated into this repository directly and uses the proper pinned hashes of the tock-litex repository, we don't risk it falling out of sync with the release upstream Tock expects. As soon as we have a decoupled system for then actually running these tests, CI runs could simply fail because the simulator isn't updated at the same instant as the LiteX update to Tock is merged. Don't get me wrong, I think having the CI integration as documented in #2633 is pretty cool, however it appears that running these integration tests, which we are able to run through a regular CI workflow, through the hardware CI system introduces complexity which could be avoided. Furthermore, this currently runs hosted on GitHub Actions which is great because it only requires somewhat significant CPU power for a reasonably short time. I would presumably need to host the test harness on my own? It won't run on something like a Raspberry Pi, I'll need a more powerful PC for that. Do I understand all of this correctly?

I tried to separate this as a concept from the other Tock CI (which does, apart from the QEMU tests, in general not perform runtime tests) by placing this into a different CI workflow.

hudson-ayers:
> @ppannuto In general I would agree. Hardware CI should probably be done through the hardware CI interface developed as documented in #2633.
> 
> However I really wouldn't treat the LiteX simulation as a piece of hardware. Incidentally it runs the same softcore CPU and looks very similar to the LiteX Arty target, and thus is a reasonably close approximation to assert that core functionality of Tock -- specifically RISC-V processes, syscalls, the PMP and some capsules -- work as expected. Also the bitstream updates (in this case the release of the tock-litex companion repository) would still need to happen, as the "hardware" does evolve and our model in Tock needs to reflect that.
> 
> When CI is integrated into this repository directly and uses the proper pinned hashes of the tock-litex repository, we don't risk it falling out of sync with the release upstream Tock expects. As soon as we have a decoupled system for then actually running these tests, CI runs could simply fail because the simulator isn't updated at the same instant as the LiteX update to Tock is merged. Don't get me wrong, I think having the CI integration as documented in #2633 is pretty cool, however it appears that running these integration tests, which we are able to run through a regular CI workflow, through the hardware CI system introduces complexity which could be avoided. Furthermore, this currently runs hosted on GitHub Actions which is great because it only requires somewhat significant CPU power for a reasonably short time. I would presumably need to host the test harness on my own? It won't run on something like a Raspberry Pi, I'll need a more powerful PC for that. Do I understand all of this correctly?
> 
> I tried to separate this as a concept from the other Tock CI (which does, apart from the QEMU tests, in general not perform runtime tests) by placing this into a different CI workflow.
> 
> Edit: I don't quite understand this part: "lets updates to the LiteX platform go through without blocking on PRs to Tock proper". Are you concerned a changing upstream LiteX would break the Tock CI? That really shouldn't happen, all dependencies that this CI workflow pulls in are pinned using a Git revision hash. It _should_ never break, and an update of the tock-litex hashes would also carry the necessary updates to the Tock model.

One nice thing about software CI compared to hardware CI is that individual contributors can iterate locally on breakages rather than having to wait on the results of hardware CI runs happening elsewhere. And it looks like the LiteX simulator CI will test significant functionality that is not currently tested by our QEMU tests -- namely GPIO and IPC.

lschuermann:
Alright, all dependencies of this are in Tock mainline so this should be ready to land. Every other dependency pulled in from outside is pinned by a commit hash or version, so we shouldn't really see failures based on other repos advancing. Furthermore, this is self-testing and thus we can make sure that on a LiteX update, all other required components are updated as well.

I managed to shave off about a minute of build time by not including the ARM toolchain and selectively building the required `libtock_c` apps. When there are better ways to parallelize build steps in GitHub actions we may be able to reduce the build times for this further, but with an average of about 6-7 minutes this seems to be (much) quicker than the macOS builds almost all of the time.

I think it's still a good idea to not make this a required check for bors at first, just so we can estimate how fragile it is. I don't know any particular reason why it should break, but I also don't want to annoy people. :smile:

PR #2805, opened 2021-09-02
rp2040: Make peripherals construction functions unsafe

alexandruradovici:
### Pull Request Overview

Based on the discussion from #2800, it seems that for now this is the best solution. This pull request makes all the `new` functions from the RP2040 chip unsafe. I considered several solutions:
1. Use a capability, but the general idea is that it is not a good solution (#2800);
2. Make `new` function private to the crate, but this prevents the boards from creating single peripherals;
3. Make `new` functions unsafe, this at least prevents creating peripherals from capsules.

### Testing Strategy

N/A

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nano_rp2040_connect/src/io.rs
+++ b/boards/nano_rp2040_connect/src/io.rs
@@ -46,7 +46,7 @@ impl IoWrite for Writer {
         self.uart.map_or_else(
             || {
                 // If no UART is configured for panic print, use UART0
-                let uart0 = &Uart::new_uart0();
+                let uart0 = unsafe { &Uart::new_uart0() };
 
                 if !uart0.is_configured() {
                     let parameters = Parameters {
@@ -59,8 +59,8 @@ impl IoWrite for Writer {
                     //configure parameters of uart for sending bytes
                     let _result = uart0.configure(parameters);
                     //set RX and TX pins in UART mode
-                    let gpio_tx = RPGpioPin::new(RPGpio::GPIO0);
-                    let gpio_rx = RPGpioPin::new(RPGpio::GPIO1);
+                    let gpio_tx = unsafe { RPGpioPin::new(RPGpio::GPIO0) };
+                    let gpio_rx = unsafe { RPGpioPin::new(RPGpio::GPIO1) };
                     gpio_rx.set_function(GpioFunction::UART);
                     gpio_tx.set_function(GpioFunction::UART);
                 }
--- a/boards/pico_explorer_base/src/io.rs
+++ b/boards/pico_explorer_base/src/io.rs
@@ -45,7 +45,7 @@ impl IoWrite for Writer {
         self.uart.map_or_else(
             || {
                 // If no UART is configured for panic print, use UART0
-                let uart0 = &Uart::new_uart0();
+                let uart0 = unsafe { &Uart::new_uart0() };
 
                 if !uart0.is_configured() {
                     let parameters = Parameters {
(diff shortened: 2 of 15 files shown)


bradjc:
Best solution for what?

Is there anything memory, type, or thread unsafe about creating a peripheral?

alexandruradovici:
This would prevent code that is not specifically tagged unsafe from creating a peripheral. 

This was one observation and concern that an industry partner had related to the safety of Tock. There was absolutely nothing stopping safe code form creating a peripheral from *out of the blue*.

lschuermann:
@bradjc is correct, while having two drivers of a peripheral might lead to _unexpected_ behavior, it must not necessarily be _unsafe_ in the Rust sense. In #2800 at the end we were specifically concerned with peripherals that can write to arbitrary memory, which is where we start getting into Rust unsafety territory.

alexandruradovici:
I agree that from Rust's safety point of view creating new peripherals is not unsafe. On the other hand, I see no alternative to stop safe code (like a capsule) from creating arbitrary peripherals.

Isn't this the same case as for the implementations for `Chip`:

```rust
impl<'a, I: InterruptService<()>> Rp2040<'a, I> {
    pub unsafe fn new(interrupt_service: &'a I, sio: &'a SIO) -> Self {
        Self {
            mpu: cortexm0p::mpu::MPU::new(),
            userspace_kernel_boundary: cortexm0p::syscall::SysCall::new(),
            interrupt_service,
            sio: sio,
            processor0_interrupt_mask: interrupt_mask!(interrupts::SIO_IRQ_PROC1),
            processor1_interrupt_mask: interrupt_mask!(interrupts::SIO_IRQ_PROC0),
        }
    }
}
```

bradjc:
> I see no alternative to stop safe code (like a capsule) from creating arbitrary peripherals

Do you have a commit that illustrates the problem? I still don't see how a capsule could define its own RPi ADC driver, for example. And since that and the TBF parsing library are the only two crates that can't just call `unsafe` anyway, I think it's really the main focus.

alexandruradovici:
Out of tree users might create their own drivers crate that follows the *capsules rule*. I use this for the Tock book. This create might define any dependencies it wants, including the rp2040 crate.

bradjc:
> Out of tree users might create their own drivers crate that follows the _capsules rule_. I use this for the Tock book [template](https://github.com/WyliodrinEmbeddedIoT/tock-project). This create might define any dependencies it wants, including the rp2040 crate.

Sure, but it can also not include `#![forbid(unsafe_code)]` and then call the functions anyway. I don't see how we can prevent out-of-tree code from violating overall system stability.

hudson-ayers:
It does seem like if you are allowing the crate to pick arbitrary dependencies it could also choose to depend on any number of public crates that expose safe interfaces to unsafe operations (https://crates.io/crates/totally-safe-transmute , anyone?). I think that "crates you do not trust to manipulate hardware peripherals cannot depend on the chip peripherals crate" is not totally unreasonable. 

That said, I realize that if not for the desire to let boards choose which peripherals they create, all of these types would probably be private anyway. Is there a reason that you would want a crate which is:

- not be allowed to use unsafe 
- depends on the chip crate  
- is allowed to do anything exposed by the chip crate 
- not allowed to create peripherals?

bradjc:
Please re-open as needed.

Issue #2413, opened 2021-02-05
TicKV Hash Algorithm


alistair23:
TicKV has made some progress and is now at the point where we need a hash algorithm. The question now is which algorithm and which implementation.

The algorithm is not used for cryptography, so we don't need a crypto one.

Properties:
 * Fast
 * Easy-ish to implement on embedded systems
 * Good for kv stores and resistant to collisions
 * Produces a 64-bit hash

I have no intentions of inventing another algorithm. My plan was to use an existing one and probably also an existing implementation, unless there is a good reason to re-implement it (there rarely is).

I have found two main contenders, although I'm option to other options.

The main options I have found are:
 - [SipHasher](https://github.com/veorq/SipHash/blob/master/README.md): Provided but deprecated by [Rust core](https://doc.rust-lang.org/core/hash/struct.SipHasher.html) and now a [standalone crate](https://github.com/jedisct1/rust-siphash)
 - [FNV](https://www.isthe.com/chongo/tech/comp/fnv/index.html) has a [Rust crate](https://github.com/servo/rust-fnv)
 - ~~[SwissTable](https://abseil.io/blog/20180927-swisstables) used by Rust for it's [HashMap](https://doc.rust-lang.org/std/collections/struct.HashMap.html)~~

Other options:
 - [Spooky hash](https://burtleburtle.net/bob/hash/spooky.html)

alistair23:
@phil-levis thoughts?

jrvanwhy:
SwissTable is a hash table, not a hash function.

You should probably consider Fnv as well.

alistair23:
Thanks! I skimmed through the SwissTable section too quickly.

alistair23:
At a quick look FNV opens up the possibility of a DOS from malicious apps with large keys. For smaller operations it does seem to be faster.

l162:
Likely unrelated, but would it be possible to use set approximation (like a bloom filter) here? Have a `needs_alloc` (const generic) version of [Xor Filters](https://arxiv.org/abs/1912.08258) sitting around at home. Fuse filters may be particularly good for large number of keys (opt-in depending on how a board intends to use the kv). Could potentially decrease object overhead, but also lead to different design.

wright-joe:
[Murmur3](https://docs.rs/murmur3/0.5.1/murmur3/)

l162:
https://github.com/nickbabcock/highway-rs

l162:
[...]

PR #2804, opened 2021-09-01
OpenTitan: Bump the SHA and add CW310 support

alistair23:
### Pull Request Overview

This PR bumps the OpenTitan bitstream SHA and also adds support for the new [CW310 board supported by OpenTitan](https://docs.opentitan.org/doc/ug/fpga_boards/).

The two designs are very similar, but the CW310 has the full amount of flash and also includes some missing components (such as the OTP).

### Testing Strategy

Running the tests on the CW310 and Nexys Video

### Documentation Updated

- [X] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [X] Ran `make prepush`.

PR #2812, opened 2021-09-07
Add a software hash to Tock and connect it to TicKV

alistair23:
### Pull Request Overview

This PR adds a software SipHash to the Tock kernel, this includes a test case to ensure it's working.

Then we connect the hash to TicKV to generate keys.

SipHash was selected as it is quick while also providing resilience to DOS attacks from untrusted applications. This is compared to FNV which although quicker for short strings, is not resilient against arbitrarily long input.

This applies on top of: https://github.com/tock/tock/pull/2804

Fixes: https://github.com/tock/tock/issues/2413

### Testing Strategy

OpenTitan unit tests.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/sip_hash.rs
+++ b/capsules/src/sip_hash.rs
@@ -28,7 +28,7 @@ use core::{cmp, mem};
 use kernel::dynamic_deferred_call::{
     DeferredCallHandle, DynamicDeferredCall, DynamicDeferredCallClient,
 };
-use kernel::hil::hasher::{Client, Hasher};
+use kernel::hil::hasher::{Client, Hasher, SipHash};
 use kernel::utilities::cells::{OptionalCell, TakeCell};
 use kernel::utilities::leasable_buffer::LeasableBuffer;
 use kernel::ErrorCode;
@@ -74,6 +74,37 @@ impl<'a> SipHasher24<'a> {
         let hasher = SipHasher {
             k0: 0,
             k1: 0,
+            length: 0,
+            state: State {
+                v0: 0x736f6d6570736575,
+                v1: 0x646f72616e646f6d,
+                v2: 0x6c7967656e657261,
+                v3: 0x7465646279746573,
+            },
+            ntail: 0,
+            tail: 0,
+        };
+
+        Self {
+            client: OptionalCell::empty(),
+            hasher: Cell::new(hasher),
+            add_data_deferred_call: Cell::new(false),
+            complete_deferred_call: Cell::new(false),
+            deferred_caller,
+            deferred_handle: OptionalCell::empty(),
+            data_buffer: TakeCell::empty(),
+            out_buffer: TakeCell::empty(),
+        }
+    }
+
+    pub const fn new_with_keys(
+        deferred_caller: &'static DynamicDeferredCall,
+        k0: u64,
+        k1: u64,
+    ) -> Self {
+        let hasher = SipHasher {
+            k0,
+            k1,
             length: 0,
             state: State {
                 v0: 0x736f6d6570736575,
@@ -256,6 +287,20 @@ impl<'a> Hasher<'a, 8> for SipHasher24<'a> {
         hasher.ntail = 0;
 
         self.hasher.set(hasher);
+    }
+}
+
+impl<'a> SipHash for SipHasher24<'a> {
+    fn set_keys(&self, k0: u64, k1: u64) -> Result<(), ErrorCode> {
+        let mut hasher = self.hasher.get();
+
+        hasher.k0 = k0;
+        hasher.k1 = k1;
+
+        self.hasher.set(hasher);
+        self.clear_data();
+
+        Ok(())
     }
 }
 
--- a/capsules/src/sip_hash.rs
+++ b/capsules/src/sip_hash.rs
@@ -1,0 +1,289 @@
+//! Tock SipHash capsule.
+//!
+//! This is a async implementation of the SipHash.
+//!
+//! This capsule was originally written to be used as part of Tocks
+//! key/value store. SipHash was used as it is generally fast, while also
+//! being resilient against DOS attacks from userspace
+//! (unlike <https://github.com/servo/rust-fnv>).
+//!
+//! Read <https://github.com/veorq/SipHash/blob/master/README.md> for more
+//! details on SipHash.
+//!
+//! The implementation is based on the Rust implementation from
+//! rust-core, avaliable here: <https://github.com/jedisct1/rust-siphash>
+//!
+//! Copyright 2012-2016 The Rust Project Developers.
+//! Copyright 2016-2021 Frank Denis.
+//! Copyright 2021 Western Digital
+//!
+//! Licensed under the Apache License, Version 2.0 LICENSE-APACHE or
+//! <http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+//! LICENSE-MIT or <http://opensource.org/licenses/MIT>, at your
+//! option.
+
+use core::cell::Cell;
+use core::convert::TryInto;
+use core::{cmp, mem};
+use kernel::dynamic_deferred_call::{
+    DeferredCallHandle, DynamicDeferredCall, DynamicDeferredCallClient,
+};
+use kernel::hil::digest::{Client, Digest};
+use kernel::utilities::cells::{OptionalCell, TakeCell};
+use kernel::utilities::leasable_buffer::LeasableBuffer;
+use kernel::ErrorCode;
+
+pub struct SipHasher24<'a> {
+    client: OptionalCell<&'a dyn Client<'a, 8>>,
+
+    hasher: Cell<Hasher>,
+
+    add_data_deferred_call: Cell<bool>,
+    complete_deferred_call: Cell<bool>,
+    deferred_caller: &'static DynamicDeferredCall,
+    deferred_handle: OptionalCell<DeferredCallHandle>,
+
+    data_buffer: TakeCell<'static, [u8]>,
+    out_buffer: TakeCell<'static, [u8; 8]>,
+}
+
+#[derive(Debug, Clone, Copy)]
+struct Hasher {
+    k0: u64,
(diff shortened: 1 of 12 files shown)


bradjc (approved):

alistair23:
@phil-levis I converted this to use a new non-cryptographic hash trait

kernel/src/hil/hasher.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Call before `Hasher::run()` to perform Sha384
+    fn set_mode_sha384(&self) -> Result<(), ErrorCode>;
+}
+
+pub trait Sha512 {
+    /// Call before `Hasher::run()` to perform Sha512
+    fn set_mode_sha512(&self) -> Result<(), ErrorCode>;
+}
+
+pub trait HMACSha256 {
```
phil-levis, line 86: Why does the Hasher HIL have cryptographic hashes in it?

kernel/src/hil/hasher.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    /// This callback is called when a digest is computed.
+    /// On error or success `digest` will contain a reference to the original
+    /// data supplied to `run()`.
+    fn hash_done(&'a self, result: Result<(), ErrorCode>, digest: &'static mut [u8; L]);
+}
+
+/// Computes a non-cryptographic hash over data
+///
+/// 'L' is the length of the 'u8' array to store the digest output.
```
phil-levis, line 24: You're interchangeably using the words hash and digest here, when it seems like "digest" means cryptographic hash and "hash" means non-cryptographic hash. This looks like a copy-paste error. Please carefully read over the file and edit it.

phil-levis (changes requested):

kernel/src/hil/hasher.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// Returns the number of bytes parsed on success
+    /// There is no guarantee the data has been written until the `add_data_done()`
+    /// callback is fired.
+    /// On error the return value will contain a return code and the original data
+    fn add_data(
+        &self,
+        data: LeasableBuffer<'static, u8>,
+    ) -> Result<usize, (ErrorCode, &'static mut [u8])>;
+
+    /// Request the hardware block to generate a Hasher and stores the returned
```
phil-levis, line 44: Given this is about a software hash, should this say "Request the implementation to generate a hash"?

kernel/src/hil/hasher.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// error with error code ENOSUPPORT.
+    fn run(&'a self, hash: &'static mut [u8; L]) -> Result<(), (ErrorCode, &'static mut [u8; L])>;
+
+    /// Clear the internal state of the engine.
+    /// This won't clear the buffers provided to this API, that is up to the
+    /// user to clear.
+    fn clear_data(&self);
+}
+
+pub trait Sha224 {
```
phil-levis, line 65: Why are cryptographic hashes in a HIL for non-cryptographic hashes? It seems like there should instead be a SipHash trait. There are different parameters for SipHash: should those be settable?

kernel/src/hil/hasher.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// On error the return value will contain a return code and the original data
+    /// If there is data from the `add_data()` command asyncrously waiting to
+    /// be written it will be written before the operation starts.
+    ///
+    /// If an appropriate `set_mode*()` wasn't called before this function the
+    /// implementation should try to use a default option. In the case where
+    /// there is only one hash supported this should be used. If there is no
+    /// suitable or obvious default option, the implementation can return an
+    /// error with error code ENOSUPPORT.
+    fn run(&'a self, hash: &'static mut [u8; L]) -> Result<(), (ErrorCode, &'static mut [u8; L])>;
```
phil-levis, line 57: Should there be a method to query the correct size of the hash? E.g. for SipHash this would be 8 or 16 bytes, but for other hashes it might be different.

lschuermann: I _think_ that this is encoded in the const generic `L`, which you should be able to access given you have access to the precise `T: Hasher<L>` or a `&dyn Hasher<L>`.

capsules/src/sip_hash.rs
```diff
@@ -0,0 +1,287 @@
...
+    v0: u64,
+    v2: u64,
+    v1: u64,
+    v3: u64,
+}
+
+impl<'a> SipHasher24<'a> {
+    pub const fn new(deferred_caller: &'static DynamicDeferredCall) -> Self {
+        let hasher = SipHasher {
+            k0: 0,
```
phil-levis, line 75: There seems to be no way to set `k0` or `k1`?

kernel/src/hil/hasher.rs
lschuermann: And also: doesn't the `set_mode` style of selecting a hash function collide with how the output length is encoded as a const generic? I.e. `Hasher<64>` is a valid type specification for some SHA2-512 hash algorithm implementation, but not SHA2-256.

If we commit to a const generic style interface, we should probably have each `Hasher<L>` implement a specific algorithm (so make it configurable by the type you pass in instead of runtime-configurable through the HIL).

kernel/src/hil/hasher.rs
phil-levis: Right! Sorry, got confused.

capsules/src/sip_hash.rs
alistair23: No, there isn't. We have no need to set the keys at this point. In the future if we do we can add a `new_with_keys()` function that sets them

phil-levis: If you're going to implement SipHash, please implement it. :) If you want a partial implementation that does just what your use case needs, that's fine, but then it should be in the TickV capsule and not a general capsule that claims to implement a particular PRF.

I don't think you want a `new_with_keys()`, because this would make the key fixed for the lifetime of the PRF, which would make it not reusable. Instead, shouldn't it be done as it is with SHA HMACs in Digest, where there's a way to set it to a mode, and setting the mode passes the key? I think you want to structure this almost identically to `Digest`, it just has different computational properties, kind of like how `Entropy` and `Random` are structurally so similar.

alistair23: I can add a `new_with_keys()` then.

As for changing the keys, I don't think that's something we want to support. The original Rust implementation doesn't: https://github.com/jedisct1/rust-siphash/blob/master/src/sip.rs#L207

I also found some other implementations and they don't support changing the keys on the fly. If we want new keys we can just create a new instance

alistair23: As for structuring this like `Digest`, I don't think we want to do that. `Digest` is aimed at something that supports SHA and we can then specify different SHA-like operations.

This on the other hand is much more specilised. The idea is someone creates a `SipHash` instance and then uses that. This is more similar to Rusts Hasher trait then Digest

phil-levis: > I can add a `new_with_keys()` then.
> 
> As for changing the keys, I don't think that's something we want to support. The original Rust implementation doesn't: https://github.com/jedisct1/rust-siphash/blob/master/src/sip.rs#L207
> 
> I also found some other implementations and they don't support changing the keys on the fly. If we want new keys we can just create a new instance

But we don't have dynamic allocation. If we had dynamic allocation, then a static key at initialization is fine, because you can create new instances if new keys appear.

phil-levis: > As for structuring this like `Digest`, I don't think we want to do that. `Digest` is aimed at something that supports SHA and we can then specify different SHA-like operations.
> 
> This on the other hand is much more specilised. The idea is someone creates a `SipHash` instance and then uses that. This is more similar to Rusts Hasher trait then Digest

I'm talking about the HIL. `Digest` provides cryptographic hashes, which have particular mathematical properties. `Hasher` (which should be `Hash`) provides non-cryptographic hashes, which have other mathematical properties. If you just want to provide a specialized `SipHash`, that's fine, there just shouldn't be a HIL around it.

I agree that `Hash` wants to be a bit different from `Digest`, in particular we don't need to differentiate verification and computation. But overall its structure should be similar.

alistair23:
Ping!

alistair23:
This is ready to merge

alistair23:
Ping!

phil-levis (dismissed):
Let's talk about this in the OT call today -- it looks like there's just a bunch of error cases tossed in but it's not clear what triggers each one and how they overlap. I suspect that if we just talk for 10 minutes it should be pretty easy to sort out. The basic idea is that `ErrorCode`s need to be specified precisely, such that a user of the trait can precisely know their meaning: otherwise it's fuzzy and so implementation-specific, which is not interoperable.

kernel/src/hil/hasher.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// Request the implementation to generate a hash and stores the returned
+    /// hash in the memory location specified.
+    /// This doesn't return any data, instead the client needs to have
+    /// set a `hash_done` handler to determine when this is complete.
+    /// On error the return value will contain a return code and the original data
+    /// If there is data from the `add_data()` command asyncrously waiting to
+    /// be written it will be written before the operation starts.
+    /// The possible ErrorCodes are:
+    ///    - BUSY: The system is busy
+    ///    - ALREADY: An operation is already on going
```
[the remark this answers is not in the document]
alistair23: `BUSY` was a generic the system is busy. For example maybe there is initalisation required that hasn't completed. While `ALREADY` is explicitly that an operation has already started and we are waiting for a callback.

alistair23:
Ping!

alistair23:
Ping!

alistair23:
All the `ErrorCode` issues are addressed, this should be ready to go

bradjc:
@phil-levis

bradjc (approved):

PR #2816, opened 2021-09-10
Update Nightly - Sep 2021

bradjc:
### Pull Request Overview

Well, this was _supposed_ to be a simple periodic update to a newer nightly (we tend to do this every two months). However, things in the Rust world have changed!

Two major updates:
- Something changed with macros warnings so we have to remove `;`s.
- Descriptive label names in `asm!()` are no longer allowed. So this changes them to numbers.

Unfortunately, something else has happened where the RAM section no longer fits. It's not clear if something has changed that we need to fix, or if this is just a bad nightly to use.

### Testing Strategy

travis

### TODO or Help Wanted

Need to figure out the RAM errors.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


bradjc:
Around and around we go with the linker!

Originally we used the `.relocate : AT (_etext)` syntax in the main linker script to get the relocations in both ram and flash. That changed in 2018 (0cd5498bafe3ae1daecf6a12998d0ea1d794d298) to use the `} > ram AT > rom` syntax. Then later in 2018 we switched _back_ to `.relocate : AT (_etext)` (b5d7f8901cecd85d9eab0a50f4ba5c0eb82b4a91) because that is all that worked on llvm-ld when we switched the rust provided llvm toolchain. Now in 2021 we are switching back to `} > ram AT > rom` because now that seems to be the only thing that works in the newest llvm-ld.

arch/cortex-m/src/lib.rs
```diff
@@ -76,7 +76,7 @@ pub unsafe extern "C" fn svc_handler_arm_v7m() {
     // is something other than 0xfffffff9, then we are coming from an app which
     // has called a syscall.
     cmp lr, #0xfffffff9
-    bne to_kernel
+    bne 100f
```
ppannuto, line 79:
```suggestion
    bne 100f // to_kernel
```

🪦 useful names, but we should maybe put on both ends?

arch/rv32i/src/lib.rs
```diff
@@ -104,17 +104,17 @@ pub extern "C" fn _start() {
...
+          200: // data_init_loop
+            beq  a0, a1, 201f           // If we have reached the end of the .data
                                         // section then we are done.
             lw   a3, 0(a2)              // a3 = *a2. Load value from initial values into a3.
             sw   a3, 0(a0)              // *a0 = a3. Store initial value into
                                         // next place in .data.
             addi a0, a0, 4              // a0 = a0 + 4. Increment to next word in memory.
             addi a2, a2, 4              // a2 = a2 + 4. Increment to next word in flash.
-            j data_init_loop            // Continue the loop.
+            j 200b                      // Continue the loop.
```
ppannuto, line 115: okay, they are all "number plus surprise `f`" except this one which is "number plus surprise `b`"??

I'm going to ask a question that I probably know the dissatisfying answer to: are we sure that label-free is the future of rust, or is there a chance we can hold out and they'll revert or fix this?

ppannuto:
If we can't have labels, is it worth doing something like `/* ASM_LABEL: name */` _everywhere_ there used to be labels, and then someday we can hopefully put them all back? Basically I don't want to lose metadata that we have with labels on each end everywhere

arch/rv32i/src/lib.rs
bradjc: You can reuse labels! Because why not! https://sourceware.org/binutils/docs/as/Symbol-Names.html#Local-Labels

And, do you like to read discouraging github threads? https://github.com/rust-lang/rust/issues/81088

arch/cortex-m/src/lib.rs
bradjc: I realized that eventually, but harder when the asm is already commented properly

bradjc:
https://github.com/tock/tock/pull/2816/checks?check_run_id=3569507203#step:10:204

huh?

boards/redboard_artemis_nano/src/main.rs
```diff
@@ -148,11 +148,11 @@ pub unsafe fn main() {
     pwr_ctrl.enable_iom2();
 
     // Enable PinCfg
-    &peripherals
+    let _ = &peripherals
```
alexandruradovici, line 151: I'll probably get an obvious answer, but why is this necessary?

bradjc: Example:
```
warning: unused borrow that must be used
   --> boards/redboard_artemis_nano/src/main.rs:233:5
    |
233 |     &peripherals.iom2.enable();
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^ the borrow produces a value
    |
help: use `let _ = ...` to ignore the resulting value
    |
233 |     let _ = &peripherals.iom2.enable();
    |     +++++++
```

ppannuto:
re CI failure: looks like that usb bulk test relies on a package that hasn't been updated in 5 years, https://crates.io/crates/libusb, which is now no longer compatible with something in newest nightly :(

`tools/usb/bulk-test` only has one meaningful commit, from 2018 when it was first added. Does this test still work? Does anyone or anything run it? :/

bradjc:
> `tools/usb/bulk-test` only has one meaningful commit, from 2018 when it was first added. Does this test still work? Does anyone or anything run it? :/

It does still work, I used it when ramping up USB support on the nano33ble. It's useful to match a test in `capsules/usb`, but at this point we have enough USB support we probably don't need it anymore.

hudson-ayers:
The issue with libusb is that it depended on a very old version of bit-set, which depended on bit-vec but did not pin a version for bit-vec, which Cargo now forbids.

I forked libusb to https://github.com/tock/libusb-rs and added a commit to use a slightly newer bit-set dependency which does not have this issue, and now all the USB tests build again using a dependency on that repository. Looking at the old libusb-rs, I don't think we can expect this update to be accepted there, as similarly simple PRs have been ignored for the last 5 years.

arch/rv32i/src/syscall.rs
```diff
@@ -219,7 +219,7 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
           //  5*4(sp): x4
           //  4*4(sp): x3
           //  3*4(sp): x1
-          //  2*4(sp): _return_to_kernel (address to resume after trap)
+          //  2*4(sp): 100      (address to resume after trap)
```
bradjc, line 222:
```suggestion
          //  2*4(sp): _return_to_kernel (100) (address to resume after trap)
```

hudson-ayers:
The QEMU tests fail because something about the new nightly seems to have broken the test runner itself -- running qemu on its own works correctly, and reverting just the test-runner to the old nightly also fixes things.

hudson-ayers:
> The QEMU tests fail because something about the new nightly seems to have broken the test runner itself -- running qemu on its own works correctly, and reverting just the test-runner to the old nightly also fixes things (I edited the Makefile and test runner such that the test runner would not rebuild the board before running qemu).

I pushed a commit to have the qemu-runner check for the "QEMU: Terminated" string instead of looking for an EOF. I am not sure why updating to a new nightly would prevent it from finding the EOF, but this updated version works. @alistair23 let me know if you think that looks okay.

hudson-ayers (approved):

alistair23 (approved):

hudson-ayers:
ping?

hudson-ayers (approved):

PR #2819, opened 2021-09-13
components/{gpio,button}: support both referenced and owned GPIO pins

lschuermann:
### Pull Request Overview

Before this change the `gpio_component_helper!` and `button_component_helper!` macros could only take references to GPIO pins. This works fine for most boards, which will have a fixed number of GPIO pins which are statically allocated within the respective chip crate.

However, it is also useful for the `{gpio,button}_component_helper!` macro to support passing in owned GPIO pins, which are then internally allocated through `static_init!`. Specifically, dynamic platfoms such as LiteX do not have a fixed number of GPIOs on every board or even in different configurations. Thus it is not trivial to preallocate all GPIO pins within the board crate and take a reference to them. Instead, GPIO pin instances are created on the fly and passed to the caller as owned objects (holding a reference to the GPIO controller). This can also reduce resource usage, as unused GPIO pins will not consume a significant amount of memory.

To support easy instantiation of a GPIO or button driver using such owned GPIO pin structs, this adds support to indicate whether the passed GPIO instances are references or owned values. This is done by placing the `ref` keyword before the GPIO pin type information, in case references are being passed in. For owned values, the `{gpio,button}_component_helper!` will allocate memory to store them internally.

### Testing Strategy

This pull request was tested by compiling all boards

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/button.rs
+++ b/boards/components/src/button.rs
@@ -2,6 +2,13 @@
 //!
 //! Usage
 //! -----
+//!
+//! The `button_component_helper!` macro takes 'static references to GPIO
+//! pins. When GPIO instances are owned values, the
+//! `button_component_helper_owned!` can be used, indicating that the passed
+//! values are owned values. This macro will perform static allocation of the
+//! passed in GPIO pins internally.
+//!
 //! ```rust
 //! let button = components::button::ButtonComponent::new(
 //!     board_kernel,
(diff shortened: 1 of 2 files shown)


bradjc:
Ok, I have to be the one to ask: why not just add a new macro for this case?

lschuermann:
> Ok, I have to be the one to ask: why not just add a new macro for this case?

We can do that, of course! I thought this was nice syntactically, but I do admit it makes the macros slightly more complex.

bradjc:
I'm generally opposed to introducing more complexity for hardware boards to support soft cores. Having a second macro would mean the current boards don't have to change, yet we still get the support you are looking for. Just my 2c

lschuermann:
@bradjc that makes sense. While it may be possible that even a hardware board might make use of this functionality in the future, it's currently just the LiteX SoCs and so we can just introduce a `{gpio,button}_component_helper_owned!` macro.

hudson-ayers (approved):

alistair23 (approved):

bradjc (approved):

PR #2817, opened 2021-09-10
Add core notes of 2021-09-03

lschuermann:
### Pull Request Overview

Add core team call notes of 2021-09-03 As always, I hope to have attributed each argument correctly and not misrepresented anyone's opinion.

### Testing Strategy

Still needs spellchecking, hence draft PR. :)

### TODO or Help Wanted

Need to listen to the last part again and get it right.

### Documentation Updated

- [x] ~Updated the relevant files in `/docs`,~ or no updates are required.

### Formatting

- [ ] Ran `make prepush`.

PR #2820, opened 2021-09-13
Add "Kernel Loop Started" callback for main.rs

bradjc:
### Pull Request Overview

Inspired from the discussion on #2785 (shell prompt for process console), this is an attempt to add effectively a standard deferred call for main.rs that triggers immediately before the main kernel loop runs.

It is implemented as a new type in the `KernelResources` trait which is how boards configure the kernel. If a board wants to use this callback it provides an implementation of `trait KernelLoopStart`.

Based on this discussion: https://github.com/tock/tock/pull/2817/files#diff-2e0027bb3b0984600afd938c50eb7070f2884dfdf945be8f66a75d96804f789aR55

#### Question

The baseline implementation is actually pretty simple: call `KR.kernel_loop_start()` at the beginning of `kernel.kernel_loop()`. However, this doesn't really accomplish anything, as there hasn't been an opportunity for any kernel work to finish from the queue. So

```rust

fn main()
  ...
  capsule.start(); // operation that would be good to run after kernel loop starts
  process_load();
  kernel_loop();
}
```
and
```rust
fn kernel_loop_start() {
  capsule.start();
}

fn main()
  ...
  process_load();
  kernel_loop();
}
```
end up being virtually the same.

So, in the following case (which is the case in the pconsole shell prompt pr):

```rust
fn kernel_loop_start() {
  capsule.start(); // operation that also prints on uart
}

fn main()
  ...
  debug!("init complete");
  process_load();
  kernel_loop();
}
```

Whether `capsule.start()` or `debug!("init complete")` prints first still comes down to the implementation of the uart mux, which I think defeats the point of having the `kernel_loop_start` hook callback in the first place.

So, instead I implemented kernel loop like:

```rust
fn kernel_loop () {
  // do all pending kernel work
  loop {
    if kernel_work() {
      do_kernel_work()
    } else {
      break;
    }
  }

  // do hook callback
  kernel_loop_start()

  loop {
    kernel_loop_operation();
  }
}
```

and wrote a comment to somewhat formalize the kernel loop startup procedure.

The question is: thoughts? Does this make sense?

I'm worried that this isn't _quite_ what we want, since there could be pending operations in flight that haven't interrupted yet. Ideally, (maybe), all of those would finish first, but this is a hard problem that we don't have any way to solve. So perhaps this is good enough?

### Testing Strategy

Switching the order of the pconsole and debug console init and verifying that the `debug!()` message still prints first on hail.

### TODO or Help Wanted

Is this the design we want?

If we converge then I can update the other boards.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/boards/hail/src/main.rs
+++ b/boards/hail/src/main.rs
@@ -19,7 +19,7 @@ use kernel::hil;
 use kernel::hil::i2c::I2CMaster;
 use kernel::hil::led::LedLow;
 use kernel::hil::Controller;
-use kernel::platform::{KernelResources, SyscallDriverLookup};
+use kernel::platform::{KernelLoopStarted, KernelResources, SyscallDriverLookup};
 use kernel::scheduler::round_robin::RoundRobinSched;
 #[allow(unused_imports)]
 use kernel::{create_capability, debug, debug_gpio, static_init};
@@ -77,6 +77,10 @@ struct Hail {
     dac: &'static capsules::dac::Dac<'static>,
     scheduler: &'static RoundRobinSched<'static>,
     systick: cortexm4::systick::SysTick,
+    process_console: &'static capsules::process_console::ProcessConsole<
+        'static,
+        components::process_console::Capability,
+    >,
 }
 
 /// Mapping of integer syscalls to objects that implement syscalls.
@@ -112,7 +116,18 @@ impl SyscallDriverLookup for Hail {
     }
 }
 
+impl KernelLoopStarted for Hail {
+    fn kernel_loop_started(&self) {
+        // Setup the UART bus for nRF51 serialization..
+        self.nrf51822.initialize();
+
+        let _ = self.process_console.start();
+        self.process_console.display_welcome();
+    }
+}
+
 impl KernelResources<sam4l::chip::Sam4l<Sam4lDefaultPeripherals>> for Hail {
+    type KernelLoopStarted = Self;
     type SyscallDriverLookup = Self;
     type SyscallFilter = ();
     type ProcessFault = ();
@@ -120,6 +135,9 @@ impl KernelResources<sam4l::chip::Sam4l<Sam4lDefaultPeripherals>> for Hail {
     type SchedulerTimer = cortexm4::systick::SysTick;
     type WatchDog = ();
 
+    fn kernel_loop_started(&self) -> &Self::KernelLoopStarted {
+        &self
+    }
     fn syscall_driver_lookup(&self) -> &Self::SyscallDriverLookup {
         &self
     }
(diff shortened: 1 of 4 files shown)


phil-levis:
> end up being virtually the same.

I think I'd disagree on this. In the first case, you have to be careful about where you put `capsule.start()`, and it can interfere with things like components, which bundle instantiation and calls to things like `start()`. Having the callback allows you to decouple when the capsule is initiated from when it starts an operation, and you know that it starts its operation after everything has been created. 

I don't think it's a good idea to flush the kernel work queue before making the callback: this only helps with a narrow set of use cases and as you point out doesn't help with others. It's brittle, as it also depends on the timing of the work (e.g., if a peripheral is moved later it might not complete by the time the loop first runs, etc.).

For the USB-CDC stack, I'm wondering about this statement:

> There is a small issue with this. The prompt will show before the debug! message. The only way that I can think of having a prompt after the debug! message is by using an alarm that fires after 100ms.

I'd argue that the semantics that make the most sense to me are:
  1. UART writes execute in temporal order. So if a `debug!` is called before a different write, the `debug!` should show up first.
  2. If the UART is not read/connected, then there are two possible policies:
    a. The write returns OFF
    b. The write is queued and the callback comes when it's actually connected.

To me, a UART that does 1) and 2b) makes the most sense.

bradjc:
> I think I'd disagree on this. In the first case, you have to be careful about where you put `capsule.start()`, and it can interfere with things like components, which bundle instantiation and calls to things like `start()`. Having the callback allows you to decouple when the capsule is initiated from when it starts an operation, and you know that it starts its operation after everything has been created.

Boards can already ensure this, by how they order operations in `main()`. The only thing they _have_ to do is call `kernel_loop()` last, but in first case `kernel_loop_started()` is the first thing called by `kernel_loop()`, so a board can get exactly the same behavior (functionally at least) by putting something like `capsule.start()` immediately before calling `kernel_loop()`.

Are you looking at this more from an API design/software engineering perspective? I agree that our current paradigm makes it unclear if an operation in `main()` is intentionally placed/ordered, or just happenstance based on how the board development occurred. Then having the callback might make development/maintenance/intent more clear, but in my original PR comment I was referring to the functionality.

bradjc:
>     1. UART writes execute in temporal order. So if a `debug!` is called before a different write, the `debug!` should show up first.

When this very same issue came up last time someone told me [there was no guarantee of ordering,](https://github.com/tock/tock/issues/1569#issuecomment-580035871) so I had to figure something else out :).


>     2\. The write is queued and the callback comes when it's actually connected.

Queuing uart can be problematic. We actually do a version of this today, where the USB-CDC stack does queue the first message it receives after initialization, for 10 seconds or until a host connects, whatever comes first. At that point the USB-CDC stack issues callbacks as if it had sent that message and all subsequent messages, even though it can't actually send anything.

If it doesn't empty the queue at some point, then the entire UART state machine stops. Inside the kernel that could be OK, but for a process it would mean that `printf()` never returns, which is probably not what the app author intended.

bradjc:
Two problems:

1. Flushing the current kernel work queue before calling `kernel_loop_started()` is arguably a hack. There are certainly cases where it wouldn't provide the behavior we are after with respect to print ordering.
2. Adding ordering to the virtualizer is hard, because queuing happens above the virtualizer (e.g. with multiple `debug!()` messages in a row).

bradjc:
We don't currently have a use case for a feature like this that is compatible with an implementation we can actually guarantee.

PR #2822, opened 2021-09-14
kernel: process: add get_addrs(), get_sizes()

bradjc:
### Pull Request Overview

It seems like we might want to separate process printing from the actual process implementation. One challenge in doing that is getting the internal state of the process necessary to display information about the process. This provides a consolidated way to get insight about how memory is used by the process.

Rather than having individual getters for each item, this starts by grouping the values into "addresses" and "sizes". This approach seems more scalable than having individual functions, as any new values can be added to the relevant struct.

This is part of a larger effort to create a "process printer" trait and implementation separate from ProcessStandard. However, it is a standalone change, and can be discussed independently.

### Testing Strategy

todo

### TODO or Help Wanted

Comments?

We may want to remove the duplicated getters as part of this PR, or we may not. I wasn't sure, so I left them for now.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -536,6 +536,14 @@ pub trait Process {
     /// switched to.
     fn switch_to(&self) -> Option<syscall::ContextSwitchReason>;
 
+    /// Return process state information related to the location in memory
+    /// of various process data structures.
+    fn get_addresses(&self) -> ProcessAddresses;
+
+    /// Return process state information related to the size in memory of
+    /// various process data structures.
+    fn get_sizes(&self) -> ProcessSizes;
+
     /// Print out the memory map (Grant region, heap, stack, program
     /// memory, BSS, and data sections) of this process.
     fn print_memory_map(&self, writer: &mut dyn Write);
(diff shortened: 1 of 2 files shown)


bradjc:
Did you remove all or most of the redundant member functions?

And ProcessPrinter works for the memory map (#2826), and we decided that the trait based version was better than hardcoding it in process_standard.rs.

hudson-ayers:
> Did you remove all or most of the redundant member functions?
> 
> And ProcessPrinter works for the memory map (#2826), and we decided that the trait based version was better than hardcoding it in process_standard.rs.

Nope, I just looked at the benchmarks on this PR as-is, so you're right the overhead is probably less. Ok, that makes sense.

hudson-ayers (approved):

hudson-ayers:
Should we remove the redundant member functions before merging this, or should that be done as part of #2826 ?

bradjc:
I have a start here: https://github.com/tock/tock/compare/process-remove-dup-addr-fns?expand=1

The kernel crate compiles, but the capsules does not because of the process console usage. However, I think many of those will go away with #2826. So maybe the linked branch can serve as a down payment, and we can handle the rest after the process printer changes settle.

hudson-ayers:
SGTM

PR #2826, opened 2021-09-17
RFC: Create ProcessPrinter trait

bradjc:
### Pull Request Overview

This is a proof-of-concept of creating a kernel `ProcessPrinter` trait to display the process memory map. This new object prints the memory map for both the panic handler and the process console.

As we look at revamping how debug is done in the kernel, this is one implementation to consider.

### Testing Strategy

todo

### TODO or Help Wanted

Compare to alternatives.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.


lschuermann:
From a first look over it, I really like this design. It seems like a great solution to deduplicate code between the process console capsule and the kernel. I also quite like the `ProcessAddresses` and `ProcessSizes` types as an abstract process introspection API, though I'm having a hard time thinking about how well each field would map to a different process implementation. It does seem generic enough though that every reasonable process implementation could support it.

I presume to avoid paying a code size penalty for this printer in downstream boards, we would just add an `impl ProcessPrinter for ()` which does nothing?

bradjc:
> I presume to avoid paying a code size penalty for this printer in downstream boards, we would just add an `impl ProcessPrinter for ()` which does nothing and boards could choose to pass into the kernel instead?

ProcessPrinter isn't passed to the kernel, so a board doesn't have to do anything to avoid using it. The approach is completely opt-in.

lschuermann:
> ProcessPrinter isn't passed to the kernel, so a board doesn't have to do anything to avoid using it. The approach is completely opt-in.

You're right, it isn't. I was referring to the panic handler, where it is indeed a parameter:

https://github.com/tock/tock/blob/3125f2f6c4ed5b425a73db27ec7e5a11b9cd4459/kernel/src/debug.rs#L103-L109

The disadvantage of this is that you're going to need to have a valid instance of a `ProcessPrinter` compiled into the kernel now to even be able to call this function; no matter whether you pass in `Some(printer)` or `None`, the type needs to be resolved. If `ProcessPrinterText` is the only instance of that trait, it's always going to end up in the binary if the `panic` helpers are used. However, if we add a dummy implementation for `(): ProcessPrinter` which simply does nothing, we can get away without having the `ProcessPrinterText` included while still being able to call the panic support routines.

bradjc:
> The disadvantage of this is that you're going to need to have a valid instance of a `ProcessPrinter` compiled into the kernel now to even be able to call this function

But why would you call `panic_print()` if you aren't going to print the process?

lschuermann:
> But why would you call `panic_print()` if you aren't going to print the process?

`panic_print()` does more than just print the process state, right? It's a great helper no matter if you want process state to be added to the panic output. Just having kernel panic information is sufficient for me most times.

bradjc:
This is a pretty complete implementation at this point. It now uses a nice trick to convert a synchronous writer to an async binary printer, so we don't have to manually split up the memory map print.

bradjc:
Before I do the polishing to make this mergable and work with every board, does anyone else have any thoughts?

hudson-ayers (approved):

lschuermann (approved):

bradjc:
bors do something

bors retry

PR #2831, opened 2021-09-20
add redv board support

almindor:
### Pull Request Overview

Adds support for the SparkFun Redboard Red-V board which is a clone of the Hifive1-RevB (with minor LED changes).

### Testing Strategy

I tested the flashing on a RedV board, but I'm currently unable to test `libtock-rs` since it's not yet on v2 support.

### TODO or Help Wanted

Check functionality of LEDs with [this libtock-rs PR](https://github.com/tock/libtock-rs/pull/336) once ready for v2.

### Documentation Updated

- [x] Updated `/boards/README.md`

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/redboard_redv/Makefile
+++ b/boards/redboard_redv/Makefile
@@ -1,0 +1,31 @@
+# Makefile for building the tock kernel for the HiFive1 platform
+
+TARGET=riscv32imac-unknown-none-elf
+PLATFORM=hifive1
+QEMU ?= qemu-system-riscv32
+
+include ../Makefile.common
+
+# Default target for installing the kernel.
+.PHONY: install
+install: flash
+
+flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
+	openocd \
+		-c "source [find board/sifive-hifive1-revb.cfg]; program $<; resume 0x20000000; exit"
+
+qemu: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
+	$(QEMU) -M sifive_e,revb=true -kernel $^  -nographic
+
+qemu-app: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
+	$(QEMU) -M sifive_e,revb=true -kernel $^ -device loader,file=$(APP),addr=0x20040000 -nographic
+
+
+TOCKLOADER=tockloader
+TOCKLOADER_JTAG_FLAGS = --jlink --board hifive1b
+KERNEL_ADDRESS = 0x20010000
+
+# upload kernel over JTAG
+.PHONY: flash
+flash-jlink: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
+	$(TOCKLOADER) $(TOCKLOADER_GENERAL_FLAGS) flash --address $(KERNEL_ADDRESS) $(TOCKLOADER_JTAG_FLAGS) $<
--- a/boards/redboard_redv/Makefile
+++ b/boards/redboard_redv/Makefile
@@ -1,7 +1,7 @@
-# Makefile for building the tock kernel for the HiFive1 platform
+# Makefile for building the tock kernel for the RedV platform
 
 TARGET=riscv32imac-unknown-none-elf
-PLATFORM=hifive1
+PLATFORM=redboard_redv
 QEMU ?= qemu-system-riscv32
 
 include ../Makefile.common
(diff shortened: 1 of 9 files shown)


boards/redboard_redv/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+- https://www.sparkfun.com/products/15594
+
+Arduino-compatible dev board for RISC-V clone of the Hifive1.
+
+Programming
+-----------
+
+Running `make flash` should load the kernel onto the board. You will need a
+relatively new (i.e. from git) version of OpenOCD.
```
alistair23, line 12: Is this still true or will the latest OpenOCD release work?

boards/redboard_redv/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+The kernel also assumes there is the default HiFive1 software bootloader running
+on the chip.
+
+Running in QEMU
+---------------
+
+The HiFive1 application can be run in the QEMU emulation platform for RISC-V, allowing quick and easy testing.
+
+Unfortunately you need QEMU 5.1, which at the time of writing is unlikely to be avaliable in your distro. Luckily Tock can build QEMU for you. From the top level of the Tock source just run `make ci-setup-qemu` and follow the steps.
```
alistair23, line 22: We also probably don't need this sentence any more, or at least can re-word it as 5.1 is getting a little old

boards/redboard_redv/Makefile
```diff
@@ -0,0 +1,31 @@
...
+PLATFORM=hifive1
+QEMU ?= qemu-system-riscv32
+
+include ../Makefile.common
+
+# Default target for installing the kernel.
+.PHONY: install
+install: flash
+
+flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
```
alistair23, line 13: Doesn't this just flash the hfive1 binary as `PLATFORM` is hifive1?

boards/redboard_redv/layout.ld
```diff
@@ -0,0 +1,16 @@
+/* The HiFive1a board has 512 MB of flash. The first 0x400000 is reserved for
+ * the default bootloader provided by SiFive. We also reserve room for apps to
+ * make all of the linker files work, but don't really support them on this
+ * chip.
+ */
```
alistair23, line 5: This is out of date

boards/redboard_redv/src/main.rs
```diff
@@ -0,0 +1,301 @@
...
+        &()
+    }
+}
+
+/// Main function.
+///
+/// This function is called from the arch crate after some very basic RISC-V
+/// setup and RAM initialization.
+#[no_mangle]
+pub unsafe fn main() {
```
alistair23, line 117: I think the general hifive1 code could be split out into a helper function that they both call. This is a lot of duplicated code for a few different GPIOs

alistair23 (dismissed):

boards/redboard_redv/README.md
almindor: removed

boards/redboard_redv/Makefile
almindor: fixed

boards/redboard_redv/layout.ld
almindor: removed

boards/redboard_redv/src/main.rs
almindor: If you prefer that way, we can also make the board use features? What'd be better here?

The way we did this in the `rust-embedded` was to use features to select hifive1/hifive1-revb/redv using features.

bradjc: Given it's a clone, using a cfg/feature to _only_ change the hardware pin mapping is fine. As long as the only differences between the two versions are fixed in hardware, and the feature is only changing the kernel in a way that can be understood from reading relevant data sheets, then using features is a good solution.

In contrast, we generally want to avoid features to do any level of kernel customization or configuration, but I don't think that is the case here.

bradjc (commented):
When the only difference is a small and fixed hardware change, as in this case, I think we want to avoid the duplicated code.

One other note as an FYI: the hifive1b tends to be an awful platform for tock as it has such little RAM. As new (and better) risc-v hardware becomes available, we will eventually remove this board (and move it to tock-archive). However, I have no idea when that will be, exactly.

bradjc (approved):
Adding this new board seems reasonable since we already have the HiFive1, so this shouldn't be adding much overhead. Can we just merge this?

hudson-ayers (dismissed):
I am also fine with this just being merged. The duplicate code is annoying, but this seems similar to how we have handled things with the nordic boards in the past.

bradjc (approved):

hudson-ayers:
bors retry

PR #2834, opened 2021-09-21
console: reserve first upcall parameter as status_code

### LINKED ISSUE #2833: console: cannot distinguish between failure and success in upcall return
In the console capsule, it is impossible to distinguish from a successful write of 9 bytes and an error of `NOMEM` since the first parameter of the upcall return is the bytes written in the [success case][0] or the positive failure code in the [failure case][1].

I recommend moving the bytes written to the second parameter, and reserve the first parameter for status code. If the status code is `0`, then the second parameter will contain the number of bytes written.

We would also need to update the [documentation][2].

If people agree that this is an issue and the proposed solution fixes it, then I will upload a PR with the fix.

[0]:https://github.com/tock/tock/blob/master/capsules/src/console.rs#L307
[1]:https://github.com/tock/tock/blob/master/capsules/src/console.rs#L318
[2]:https://github.com/tock/tock/blob/master/doc/syscalls/00001_console.md#subscribe-number-2

jettr:
Always use the first parameter of the upcall as the status code, and move the bytes written parameter to the second parameter. This means that success will look like `(0, bytes_written, 0)` and `NOMEM` failure will look like `(9, 0, 0)`.

This fixes #2833

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/console.rs
+++ b/capsules/src/console.rs
@@ -304,7 +304,12 @@ impl uart::TransmitClient for Console<'_> {
                             // Go ahead and signal the application
                             let written = app.write_len;
                             app.write_len = 0;
-                            upcalls.schedule_upcall(1, (written, 0, 0)).ok();
+                            upcalls
+                                .schedule_upcall(
+                                    1,
+                                    (kernel::errorcode::into_statuscode(Ok(())), written, 0),
+                                )
+                                .ok();
                         }
                     }
                     Err(return_code) => {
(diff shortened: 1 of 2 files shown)


bradjc (changes requested):
This breaks stability guarantees we have, and will (I think) cause libtock-c apps to break for many people as basic compatibility between userspace and kernel (since most apps use printf) changes.

PR #2837, opened 2021-09-22
RFC: Allow use of cargo features to configure the `kernel::CONFIG` constant

hudson-ayers:
### Pull Request Overview

This pull request adds cargo features for the three configuration options in kernel/src/config.rs, so that they can be turned on and off without requiring users to vendor or modify the kernel crate. It also adds some comments documenting that cargo features are not an approved mechanism for adding general flexibility, and are instead limited only to setting the *values* of fields of the CONFIG struct, ensuring that throughout the actual codebase all combinations of features are guaranteed to compile (as these features rely on late-stage compiler optimizations to elide unused code, rather than cutting it out at the pree-processsing stage like if we used `#ifdef`). 

This PR sits on top of https://github.com/tock/tock/pull/2759. Notably, I consider #2759 valuable even without this PR, as carrying patches for the config file downstream is less of a maintenance burden than carrying patches in `process_standard.rs`, which has considerably more churn, and where the logic of the patches might need to change as that file changes.

### Testing Strategy

This pull request was tested by compiling with a couple of these options turned on in `imix/Cargo.toml`, e.g.:
```toml
[dependencies]
kernel = { path = "../../kernel", features = ["no_debug_panics"] }
```

### TODO or Help Wanted

N/A

### Documentation Updated

- [ ] Should we document this somewhere other than the code comments I added?

### Formatting

- [x] Ran `make prepush`.


kernel/src/config.rs
```diff
@@ -40,11 +40,44 @@ pub(crate) struct Config {
...
+    // option.
+    pub(crate) debug_panics: bool,
 }
 
 /// A unique instance of `Config` where compile-time configuration options are defined. These
 /// options are available in the kernel crate to be used for relevant configuration.
+/// Notably, this is the only location in the Tock kernel where we permit `#[cfg(x)]`
+/// to be used to configure code based on Cargo features.
 pub(crate) const CONFIG: Config = Config {
+    #[cfg(not(trace_syscalls))]
```
bradjc, line 71: Wait so you can just write this and nothing checks it? What is the purpose of defining features in the Cargo.toml file?

This isn't winning any favor with me :)

kernel/src/config.rs
```diff
@@ -40,11 +40,44 @@ pub(crate) struct Config {
...
     trace_syscalls: false,
+    #[cfg(trace_syscalls)]
+    trace_syscalls: true,
+    #[cfg(not(debug_load_processes))]
     debug_load_processes: false,
+    #[cfg(debug_load_processes)]
+    debug_load_processes: true,
+    #[cfg(not(no_debug_panics))]
```
bradjc, line 79: There isn't a way to make this default? It's unfortunate the feature name is based on the default.

kernel/Cargo.toml
```diff
@@ -8,3 +8,14 @@ edition = "2018"
 tock-registers = { path = "../libraries/tock-register-interface" }
 tock-cells = { path = "../libraries/tock-cells" }
 tock-tbf = { path = "../libraries/tock-tbf" }
+
+# In general, Tock discourages the use of cargo features. However for certain
+# kernel crate configuration, we have not found reasonable alternatives to
+# setting boolean values in kernel/src/config.rs. To make those settings configurable
+# without requiring users to modify the kernel crate itself, we allow configuration of
+# the values of the global CONFIG constant in kernel/src/config.rs via these features.
+# Other uses of cargo features are disallowed.
```
bradjc, line 17: It would be great to capture

```
[dependencies]
kernel = { path = "../../kernel", features = ["no_debug_panics"] }
```

here for those of us who do not use rust features ever.

bradjc (changes requested):
Need to update https://github.com/tock/tock/blob/master/doc/Configuration.md

lschuermann:
The `no_*` features seem to collide somewhat with how feature unification works, especially [the following text of the Cargo reference](https://doc.rust-lang.org/cargo/reference/features.html): "A consequence of this is that features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features."

It seems fine to use feature for configuration, though we should simply provide a `default` which contains features we want to typically have enabled. If users want to disable these, they'll have to use the `no-default-features` parameter on the command line or in their `Cargo.toml`s.

kernel/src/config.rs
```diff
@@ -40,11 +40,44 @@ pub(crate) struct Config {
...
+    debug_panics: true,
+    #[cfg(no_debug_panics)]
+    debug_panics: false,
```
lschuermann, line 82: Instead of conditionally generating code, it seems better to use the `cfg!` macro and just inject booleans. That has less potentially to produce invalid Rust code or have missing fields in the `Config` struct instantiation.

```suggestion
    trace_syscalls: cfg!(trace_syscalls),
    debug_load_processes: cfg!(debug_load_processes),
    debug_panics: cfg!(debug_panics),
```

lschuermann (changes requested):

hudson-ayers:
> The `no_*` features seem to collide somewhat with how feature unification works, especially [the following text of the Cargo reference](https://doc.rust-lang.org/cargo/reference/features.html): "A consequence of this is that features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features."
> 
> It seems fine to use feature for configuration, though we should simply provide a `default` which contains features we want to typically have enabled. If users want to disable these, they'll have to use the `no-default-features` parameter on the command line or in their `Cargo.toml`s.
> 
> For instance, the default feature set would include `debug_panics`. If you don't want that, just use
> 
> ```toml
> [dependencies]
> kernel = { path = "../../kernel", features = [], default-features = false }
> ```

I wrote this PR how you described originally, but if you do that you also have to modify the `Cargo.toml` of every crate that depends on the kernel crate to turn off default features (unless I am misunderstanding how unification works?). This also means that anyone adding a crate that depends on the kernel without disabling default features will render the feature useless, which is a very easy mistake to make.

lschuermann:
> I wrote this PR how you described originally, but if you do that you also have to modify the `Cargo.toml` of every crate that depends on the kernel crate to turn off default features (unless I am misunderstanding how unification works?). This also means that anyone adding a crate that depends on the kernel without disabling default features will render the feature useless, which is a very easy mistake to make.

That is true. I feel like this might be telling us that features may not be the tool we're looking for here? Is there some way to inject these parameters from the boards crate (which I think realistically is the only place where we want these values to be controlled) while at the same time cause them to trigger the same optimizations?

hudson-ayers:
> That is true. I feel like this might be telling us that features may not be the tool we're looking for here? Is there some way to inject these parameters from the boards crate (which I think realistically is the only place where we want these values to be controlled) while at the same time cause them to trigger the same optimizations? I've tried an associated constant with the `KernelResources`, but that unfortunately doesn't work given the process type doesn't have direct access to that.

I agree that features are not the ideal interface for this, I just consider them preferable to the alternative of requiring users to manually modify code in the kernel crate. Some mechanism for directly injecting these parameters from the board crate would be great, but anything I can think of would be a lot more limiting that what we have now -- we could use const generic parameters that get passed to all types where we need to configure based on these constants, but I think that messes with object safety for some types (like Process), and leads to much less clean code. 

I consider this approach better than the downstream patch approach, but am happy to consider an alternative if one exists.

kernel/src/config.rs
hudson-ayers: See my response to Leon on this -- default-features exist, but the way feature unification works makes them undesirable for this use case.

kernel/Cargo.toml
hudson-ayers: I added it to the comment

kernel/src/config.rs
hudson-ayers: Yeah, that is rough. I fixed the typo in Cargo.toml

kernel/src/config.rs
hudson-ayers: hmm..I agree that is cleaner, though it does make it a little less clear that you can just delete the cfg portion and replace it with true/false compared to the existing approach. I am happy to make this change if the consensus prefers it

hudson-ayers: I went ahead and updated to just use `cfg!()`

alistair23 (approved):

kernel/src/config.rs
```diff
@@ -65,8 +65,10 @@ pub(crate) struct Config {
 
 /// A unique instance of `Config` where compile-time configuration options are defined. These
 /// options are available in the kernel crate to be used for relevant configuration.
+/// Notably, this is the only location in the Tock kernel where we permit `#[cfg(x)]`
```
bradjc, line 68: This doesn't quite line up now.

bradjc (approved):
This is seemingly the best option we have due to Rust not eliding dead code in traits, and needing to guarantee that these flags are const so the compiler will always remove the debug/printing code when it is not wanted.

But, I think we also see why this `#define`-esque approach is so brittle and hard to scale in the development of this PR, and those learnings can get lost once the code is merged. So, to summarize from my point of view, the issues are:
- The `cfg` names are not "type" checked by the compiler or toolchain. That is, if the features listed in the Cargo.toml file are not the same as the ones used in the code, compilation does not fail and no warnings are generated.
- There isn't a way to set the default option for a feature, so the most straightforward approach is to name the features based on their current default. This makes the name of a cfg tied to the particular value at one point in time, and adds friction to changing it later. It also makes understanding the names more difficult as there isn't a one-to-one mapping of feature names to variable names.
- Configuring the features is not done in the code (i.e. in an `.rs` file), or in a dedicated kernel config file, but rather in an ad-hoc way using Cargo files. To an experienced Rust developer used to working on complex projects with features this might be second nature. But the more that the actual code that is used in a build depends on toolchain files rather than either source code or a dedicated and documented config file structure, the more difficult reading and understanding the code base becomes.

hudson-ayers:
> * The `cfg` names are not "type" checked by the compiler or toolchain. That is, if the features listed in the Cargo.toml file are not the same as the ones used in the code, compilation does not fail and no warnings are generated.
> * There isn't a way to set the default option for a feature, so the most straightforward approach is to name the features based on their current default. This makes the name of a cfg tied to the particular value at one point in time, and adds friction to changing it later. It also makes understanding the names more difficult as there isn't a one-to-one mapping of feature names to variable names.
> * Configuring the features is not done in the code (i.e. in an `.rs` file), but rather in the Cargo files. To an experienced Rust developer used to working on complex projects with features this might be second nature. But the more that the actual code that is used in a build depends on toolchain files rather than source code, the more difficult reading and understanding the code base becomes.

Agreed on all 3 points, though I consider the first 2 more annoying. For the first issue, a [PR](https://github.com/rust-lang/rust/pull/89346) was submitted yesterday which implements an [accepted RFC](https://rust-lang.github.io/rfcs/3013-conditional-compilation-checking.html) that will enable checking "type" checking of features in the way that you describe. So hopefully that pain point at least is temporary.

phil-levis:
I 100% agree with the first two points.

>     * Configuring the features is not done in the code (i.e. in an `.rs` file), but rather in the Cargo files. To an experienced Rust developer used to working on complex projects with features this might be second nature. But the more that the actual code that is used in a build depends on toolchain files rather than source code, the more difficult reading and understanding the code base becomes.

I disagree with this one: at some point, configuration is best done through a separate configuration file rather than code. For example, for all of the `#define` issues you can find in C code, imagine if instead you had to *manually change the C code for each option* or manually rework the boot sequence. Imagine if, to include a module in the Linux kernel, you had to modify the kernel code. We haven't reached a level of complexity in Tock that has yet required configuration tools for selecting capsules for a board, but we might some day.

Or do you count `build.rs`?

bradjc:
> Imagine if, to include a module in the Linux kernel, you had to modify the kernel code.

Good point. I've updated my comment to to differentiate a dedicated config system (ala `make menuconfig`) from an ad-hoc use of a general purpose mechanism. Well structured configuration (e.g. whether or not to include an entire kernel module) would, I think, be at least as intuitive as tracing a configuration through the source code.

PR #2839, opened 2021-09-24
Initial commit of RSA Key support

alistair23:
### Pull Request Overview

This is a split from https://github.com/tock/tock/pull/2693 adding RSA Key support.

#### Why add a trait and capsule?

Public/Private keys are too complex and critical to be handled in plain buffers. Using plain buffers opens us up to easy mistakes of mismatching keys type or extracting useful information.

For this reason this PR adds a generic HIL for public/private key pairs. Using this generic HIL we also add a RSA specific HIL and two implementations of RSA keys. We have two implementations to support the case where we have both public/private keys and the case where we only have a public key. This way we can avoid overhead of private key storage if we aren't going to use it.

#### Why add a new enum type?

The goal is to support 3 main use cases:
 1. Keys stored on flash. The keys are stored at some address in read only flash and we want to "import" them and use them in the kernel
 2. The app specifies a key. A userspace application obtains a key and passes it to the kernel to use for crypto operations
 3. We generate a key pair while running

In order to support all three we need to accept both mutable buffers (for userspace driver capsules and generating keys) and immutable buffers (for data stored in flash).

In order to support both I have added a new enum called `MutImutBuffer`. This allows us to pass around either a mutable static buffer or an immutable static buffer. This can be passed all the way down to the chip driver which can then unpack the buffers and write the data to hardware.

### Testing Strategy

This has been tested with the OpenTitan RSA implementation: https://github.com/tock/tock/pull/2899

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


kernel/src/hil/public_key_crypto/keys.rs
```diff
@@ -0,0 +1,90 @@
...
+/// An internal representation of a asymetric Public key.
+///
+/// This trait is useful for managing keys internally in Tock.
+pub trait PubKey<'a> {
+    /// Import an existing public key. This will override any existing keys.
+    ///
+    /// The possible ErrorCodes are:
+    ///     - `ALREADY`: A key is already imported or in the process of being
+    ///                  generated.
+    ///     - `INVAL`: An invalid key was supplied.
```
phil-levis, line 21: It would be good to distinguish INVAL from SIZE

kernel/src/hil/public_key_crypto/keys.rs
```diff
@@ -0,0 +1,90 @@
...
+    ///     - `NOMEM`: The key can not be stored.
+    fn import_public_key(&mut self, public_key: &[u8]) -> Result<(), ErrorCode>;
+
+    /// Store the public key in buffer if it exists.
+    fn pub_key(&'a self, buffer: &mut [u8]) -> Option<()>;
```
phil-levis, line 26: What does the return value mean?

kernel/src/hil/public_key_crypto/keys.rs
```diff
@@ -0,0 +1,90 @@
...
+/// An internal representation of a asymetric Public and Private key.
+///
+/// This trait is useful for managing keys internally in Tock.
+pub trait PubPrivKey<'a>: PubKey<'a> {
+    /// Import an existing private key. This will override any existing keys.
+    ///
+    /// The possible ErrorCodes are:
+    ///     - `ALREADY`: A key is already imported or in the process of being
+    ///                  generated.
+    ///     - `INVAL`: An invalid key was supplied.
```
phil-levis, line 41: Need SIZE.

kernel/src/hil/public_key_crypto/keys.rs
```diff
@@ -0,0 +1,90 @@
...
+    ///     - `NOMEM`: The key can not be stored.
+    fn import_private_key(&mut self, private_key: &[u8]) -> Result<(), ErrorCode>;
+
+    /// Store the private key in buffer if it exists.
+    fn priv_key(&'a self, buffer: &mut [u8]) -> Option<()>;
```
phil-levis, line 46: What does the Option signify?

kernel/src/hil/public_key_crypto/keys.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// The keys generated by `generate()` will depend on the implementation.
+    ///
+    /// The possible ErrorCodes are:
+    ///     - `BUSY`: The underlying `trng` is busy.
+    ///     - `ALREADY`: A key is already imported or in the process of being
+    ///                  generated.
+    ///     - `OFF`: The underlying `trng` is powered down.
+    ///     - `INVAL`: An invalid length was supplied
+    ///     - `NOMEM`: The key can not be stored.
+    fn generate(&'a self, trng: &'a dyn entropy::Entropy32, length: usize)
```
phil-levis, line 75: Where is it stored? How does one know the right/maximum length? E.g., suppose there is no key currently. Can you call `priv_key` to get the maximum length, or will it return 0 because there is no key? There should be a `max_length` method.

phil-levis: OK, looking at the RSA implementation, it seems like there is a constant key length, which defines the only size of the key allowed. So a key structure is bound to a specific key/length. So why can generate take a length? The traits need to describe this so that the meaning of the methods is clear. If the traits are intended to support both variable and static length keys, then they need some more support for the variable case.

capsules/src/public_key_crypto/rsa_keys.rs
```diff
@@ -0,0 +1,154 @@
...
+    fn set_client(&'a self, client: &'a dyn PubPrivKeyGenerateClient<'a>) {
+        self.client.set(client);
+    }
+
+    fn generate(
+        &'a self,
+        _trng: &'a dyn entropy::Entropy32,
+        _length: usize,
+    ) -> Result<(), ErrorCode> {
+        unimplemented!();
```
phil-levis, line 54: Needs to be implemented. I think if it were the inconsistency between the length parameter here and the static lengths returned by `len` would be clear.

phil-levis (changes requested):

alistair23:
Ok, I'm a little stuck.

There are three main use cases I think we need to support:
 1. Keys stored on flash. The keys are stored at some address in read only flash and we want to "import" them and use them in the kernel
 2. The app specifies a key. A userspace application obtains a key and passes it to the kernel to use for crypto operations
 3. We generate a key pair while running

Ideally we want to support all 3 without too much overhead.

My original plan was that a board would create a single `RSA2048Keys` struct. It would then copy over a key from flash, perform operations and then move on to the next key. This has the large downside that we have to store the key in memory. My hope though was that we could create a very small number of structs and re-use them. So for example:

```rust
let key = RSA2048Keys::new()
key.import(&flash_key_one);
key.do_stuff();
key.import(&flash_key_two);
key.do_stuff();
```

This means we need to keep a single key in memory, but can support any number of keys. We can also support all 3 use cases above. The problem is I think this is too much over head as we have to store large keys in memory.

The other option is to have an interface that takes a static key, like this:

```rust
    /// Import an existing public key. This will override any existing keys.
    ///
    /// The `public_key` is stored internally and can be retrieved with
    /// `pub_key()`.
    ///
    /// The possible ErrorCodes are:
    ///     - `ALREADY`: A key is already imported or in the process of being
    ///                  generated.
    ///     - `INVAL`: An invalid key was supplied.
    ///     - `SIZE`: An invalid key size was supplied.
    ///     - `NOMEM`: The key can not be stored.
    fn import_public_key(
        &'a self,
        public_key: &'static [u8],
    ) -> Result<(), (&'static [u8], ErrorCode)>;

    /// Return the public key supplied by `import_public_key()` or
    /// `generate()`.
    ///
    /// On success the return value is `Ok(())`.
    /// On failure the possible ErrorCodes are:
    ///     - `NODEVICE`: The key does not exist
    ///     - `SIZE`: An invalid buffer size was supplied.
    ///     - `NOMEM`: The key can not be stored.
    fn pub_key(&'a self) -> Result<&'static [u8], ErrorCode>;
```

The problem here is mutability. We need a mutable buffer to allow generating keys and working with apps, but that won't work with read only flash. It's a similar issue to https://github.com/tock/libtock-rs/issues/334

We could have read and read/write traits based on where the key is coming from. Am I missing something, does anyone have a better idea?

phil-levis:
I think that cases 2 and 3 are mostly equivalent, except that 2 requires copying the application buffer into a kernel buffer. But this means that in all cases the key buffer can be `'static`, which helps a bit. But the challenges here are similar to rule 5 from the HIL TRD, about always passing `mut` buffer references.

One solution is you access the internal key with a closure. So you can have a ReadOnlyKey and a ReadWriteKey and both implement a trait that has a closure that passes a read-only reference to the internal key. This needs to be carefully done, though, or perhaps we could run into the code size problems Hudson has raised?

Another is just that Key requires a RAM buffer, so you can either generate keys into it, or copy read-only data into it.

capsules/src/public_key_crypto/rsa_keys.rs
```diff
@@ -0,0 +1,197 @@
...
+        entropy: &mut dyn Iterator<Item = u8>,
+        error: Result<(), ErrorCode>,
+    ) -> Continue {
+        error.unwrap();
+
+        let pub_key = self.public_key.take().unwrap();
+
+        match pub_key {
+            MutImutBuffer::Mutable(buf) => {
+                for (i, b) in entropy.enumerate() {
```
phil-levis, line 33: Should this be generating and storing a set of primes?

alistair23: Yeah, I have a partial implementation of generating keys. The prime tests are tricky to implement with an async Rng engine.

Besides that generating keys is complex and I think should be reviewed in a separate PR, with the aim of getting lots of eyes on it.

Right now I would just like to focus on getting the HIL approved and then go from there.

alistair23:
I just added a design doc to hopefully help explain the design decisions. I have also converted this to use closures

alistair23:
Marking this as ready for review as there is now a full implementation

phil-levis (changes requested):
This should be updated to not use MutImutBuffer.

alistair23:
Updated to not use `MutImutBuffer` in any HIL

phil-levis:
Hey, getting back to this now. I think we resolved all of the issues with keys, but I suspect that `rsa_math` is going to require some more work due to the mutable/immutable problem. Can you cut that file so I can approve this?

alistair23:
Done!

phil-levis (approved):
I think the TRD needs a lot of work, but that can be improved later.

alistair23:
Can this be merged?

bradjc (approved):

PR #2841, opened 2021-09-24
lowrisc: add auto generated register constants

ccteng-google:
Add Rust constants for register specification that are auto generated
from the lowRISC/opentitan repo hardware IP JSON files.

Change-Id: Ib69aec3cc23b45c969b81cb211a24f82a69c37f3

### Pull Request Overview

This pull request adds lowrisc register specification as Rust constants.

### Testing Strategy

This pull request was tested by comparing the C header files generated in the lowRISC/opentitan repo.

### TODO or Help Wanted

This pull request still needs to update the lowrisc drivers to replace the current hard coded numbers with the constants.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.


alistair23 (changes requested):
Thanks for the PR, but we don't want this as is.

This doesn't follow the standard Tock register API and instead just adds a bunch of C style macros. We want to use the register API, like this: https://github.com/tock/tock/blob/master/chips/lowrisc/src/csrng.rs#L15

ccteng-google:
I can take a look at modifying the current OpenTitan tool to generate the register_structs! and register_bitfields!. But it will also need some clear definition on a few rules, for example,
- When to generate ReadWrite vs WriteOnly vs ReadOnly? the exact logic?
- When to use <u32> vs <u32, SOME_REGISTER::Register>? # fields == 1 AND field.size == register.size?

alistair23:
"Read Only, reading clears" would be `ReadOnly` as software can not write to the register
"Read zero, Write with 1 clears" should be `WriteOnly` as software can't read the register

As for including or not including `register_bitfields` if autogenerating the registers we should just always include the bitfields, there is no reason not to.

ccteng-google:
> "Read Only, reading clears" would be `ReadOnly` as software can not write to the register "Read zero, Write with 1 clears" should be `WriteOnly` as software can't read the register
> 
> As for including or not including `register_bitfields` if autogenerating the registers we should just always include the bitfields, there is no reason not to.

I'm OK with that.

ccteng-google:
Next topic. @alistair23 Do you have an example of how you'd like construct Multi-Register value, e.g. aes.KEY_SHARE (https://docs.opentitan.org/hw/ip/aes/doc/)?

ccteng-google:
Updated with auto gen register_structs! and register_bitfields! instead of pub const. Let me know what you think.

alistair23:
For multi contruct registers, right now we do it like this: https://github.com/tock/tock/blob/master/chips/earlgrey/src/aes.rs#L25 but it probably should be an array

alistair23:
Cool! This looks good!

The only issue now is that I don't think we want unimplemented register structs hanging around, so we should include them in the current drivers.

Maybe the best option is to check in the script for generating these and then copy the output into the current code?

ccteng-google:
> For multi contruct registers, right now we do it like this: https://github.com/tock/tock/blob/master/chips/earlgrey/src/aes.rs#L25 but it probably should be an array

I can leave it as is and we can always optimize it later as needed.

ccteng-google:
> Cool! This looks good!
> 
> The only issue now is that I don't think we want unimplemented register structs hanging around, so we should include them in the current drivers.
> 
> Maybe the best option is to check in the script for generating these and then copy the output into the current code?

The script is checked in at the lowrisc/opentitan repo since that's where the HW IP JSON files are, which has the option to generate C header file or Rust file (among other things). I'll have to create a PR for the latest patch that generate the `register_structs!` instead of `pub const`. I'll put a link here.

I'd prefer not go the manual copy/paste route. Maybe we can create a sub module with all the macro definition under the chips/lowrisc/src. And the actual driver code can bring them in with `mod` or `pub mod`. We can cherry pick the Rust files and not check in the ones that's not implemented, that's easy to do.

alistair23:
The problem is that I think using `mod` to include the structs will be very confusing. For someone reading the chip driver we want to include the register layout in that file, not in an external file.

lschuermann:
> The problem is that I think using `mod` to include the structs will be very confusing.

I don't think this is an issue. In most (non-trivial) drivers you can't have the register definitions along with the actual business logic on the screen at the same time anyways. I think it's entirely feasible to just view those in a separate buffer. Also, given they are automatically generated, it you'd only use them as a reference anyways (i.e. not actively modify them).

bradjc:
Closing in favor of #3045.

PR #2842, opened 2021-09-24
tock-registers: rename TryFromValue::try_from to try_from_value

lschuermann:
### Pull Request Overview

This pull request renames the `TryFromValue::try_from` method to be called `try_from_value`. This avoids a collision of `TryFrom::try_from`, which since Rust 2021 is included in the standard library prelude and thus collides with this method name.

Surprisingly, there seem to be no users of this interface in tree in Tock. Nonetheless it is a breaking interface change and thus requires a major version bump of the `tock-register-interface` library. Given that we're currently on v0.7 and wanted to release a v1.0.0 eventually, I propose we do this shortly after these changes. This might then be a good point at which `tock-register-interface` is split out into its own repository.

### Testing Strategy

This pull request was tested by compiling and CI.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers (approved):

ppannuto (approved):
bors r+

PR #2845, opened 2021-09-25
kernel::collections: redesign linked-list interface and types 

lschuermann:
### Pull Request Overview

In an effort to fix the issues identified in #2773, this proposes a new, redesigned linked-list interface and types for Tock.

It appears that one major issue with the old linked-list interface is that, while it defines some proper types for manipulating and holding a linked list, it relies on a significant part of the implementation to be filled by the user of the list, through an implementation of the `ListNode` trait respectively.

This has both some major advantages and drawbacks: it allows the implementors of linked-list data structures to be very flexible in their design and reuse existing structures as nodes in their linked-list data structure. Take for example the UART virtualizer: the `MuxUart` holds a linked list of `UartDevice`s. However, instead of creating some list containers which hold the `UartDevice`s, the `UartDevice`s simply implement `ListNode` and thus can be used to build the list themselves. With respect to Tock's model around memory allocation, this can be very beneficial.

Furthermore, by making `ListNode::next` a trait method, the lists structure can be entirely determined at runtime. There is no guarantee and / or requirement that a first call to `next` on some `ListNode` has to return the same reference as a second call. This interface is even flexible enough to support lazy allocation and evaluation of list nodes. A potential use case for this could be building a list of processes by scanning over flash.

However, it also makes implementing basic consistency checks as proposed in #2773 impossible. For example, because the list is entirely dynamic and implementation dependent, and because nodes are not managed by the list itself, it's impossible to tell whether a list will, during runtime, result in a loop. This makes handling lists difficult, especially for simple cases where all of this flexibility is not needed.

Thus this PR contains a redesigned list interface. It still allows for all of the flexibility of the previous interface, but instead of mandating this for every list, it defines an interface (the `SinglyLinkedList` trait) for how manipulating singly-linked lists should work.

In addition to that, it features two list implementations: a `GenericLinkedList` which supports all the flexibility of the previous interface by being generic over the underlying `ModifyableListNode`s, and a `SimpleLinkedList` which contains internal consistency checks as proposed by #2773. It also gives users a choice, whether they would like to avoid the additional overhead associated with the consistency checks by switching to the `GenericLinkedList`.

### Testing Strategy

Not yet tested!

### TODO or Help Wanted

As an example, this PR switches the `schedulers` over to the `SimpleLinkedList` and the `virtual_uart` to the `GenericLinkedList`. There are a lot of usages of this interfaces in tree, I'll happily change them once there is some basic agreement that this looks promising.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/boards/acd52832/src/main.rs
+++ b/boards/acd52832/src/main.rs
@@ -19,7 +19,6 @@ use kernel::hil::led::LedLow;
 use kernel::hil::rng::Rng;
 use kernel::hil::time::{Alarm, Counter};
 use kernel::platform::{KernelResources, SyscallDriverLookup};
-use kernel::scheduler::round_robin::RoundRobinSched;
 #[allow(unused_imports)]
 use kernel::{create_capability, debug, debug_gpio, static_init};
 use nrf52832::gpio::Pin;
@@ -86,7 +85,7 @@ pub struct Platform {
         'static,
         capsules::virtual_alarm::VirtualMuxAlarm<'static, nrf52832::rtc::Rtc<'static>>,
     >,
-    scheduler: &'static RoundRobinSched<'static>,
+    scheduler: &'static components::sched::round_robin::SchedulerType,
     systick: cortexm4::systick::SysTick,
 }
 
@@ -119,7 +118,7 @@ impl KernelResources<nrf52832::chip::NRF52<'static, Nrf52832DefaultPeripherals<'
     type SyscallDriverLookup = Self;
     type SyscallFilter = ();
     type ProcessFault = ();
-    type Scheduler = RoundRobinSched<'static>;
+    type Scheduler = components::sched::round_robin::SchedulerType;
     type SchedulerTimer = cortexm4::systick::SysTick;
     type WatchDog = ();
     type ContextSwitchCallback = ();
--- a/boards/clue_nrf52840/src/main.rs
+++ b/boards/clue_nrf52840/src/main.rs
@@ -24,7 +24,6 @@ use kernel::hil::usb::Client;
 use kernel::platform::chip::Chip;
 use kernel::platform::mpu::MPU;
 use kernel::platform::{KernelResources, SyscallDriverLookup};
-use kernel::scheduler::round_robin::RoundRobinSched;
 #[allow(unused_imports)]
 use kernel::{create_capability, debug, debug_gpio, debug_verbose, static_init};
 
@@ -167,7 +166,7 @@ pub struct Platform {
     adc: &'static capsules::adc::AdcVirtualized<'static>,
     temperature: &'static capsules::temperature::TemperatureSensor<'static>,
     humidity: &'static capsules::humidity::HumiditySensor<'static>,
-    scheduler: &'static RoundRobinSched<'static>,
+    scheduler: &'static components::sched::round_robin::SchedulerType,
     systick: cortexm4::systick::SysTick,
 }
 
@@ -203,7 +202,7 @@ impl KernelResources<nrf52::chip::NRF52<'static, Nrf52840DefaultPeripherals<'sta
     type SyscallDriverLookup = Self;
     type SyscallFilter = ();
     type ProcessFault = ();
-    type Scheduler = RoundRobinSched<'static>;
+    type Scheduler = components::sched::round_robin::SchedulerType;
     type SchedulerTimer = cortexm4::systick::SysTick;
     type WatchDog = ();
     type ContextSwitchCallback = ();
--- a/boards/components/src/sched/cooperative.rs
+++ b/boards/components/src/sched/cooperative.rs
@@ -12,22 +12,33 @@
 // Author: Hudson Ayers <hayers@stanford.edu>
 
 use core::mem::MaybeUninit;
+use kernel::collections::list::simple_linked_list::{SimpleLinkedList, SimpleLinkedListNode};
+use kernel::collections::list::SinglyLinkedList;
 use kernel::component::Component;
 use kernel::process::Process;
-use kernel::scheduler::cooperative::{CoopProcessNode, CooperativeSched};
+use kernel::scheduler::cooperative::CooperativeSched;
 use kernel::{static_init, static_init_half};
 
 #[macro_export]
 macro_rules! coop_component_helper {
     ($N:expr $(,)?) => {{
         use core::mem::MaybeUninit;
-        use kernel::scheduler::cooperative::CoopProcessNode;
+        use kernel::collections::list::simple_linked_list::SimpleLinkedListNode;
+        use kernel::process::Process;
         use kernel::static_buf;
-        const UNINIT: MaybeUninit<CoopProcessNode<'static>> = MaybeUninit::uninit();
-        static mut BUF: [MaybeUninit<CoopProcessNode<'static>>; $N] = [UNINIT; $N];
+        const UNINIT: MaybeUninit<SimpleLinkedListNode<'static, Option<&'static dyn Process>>> =
+            MaybeUninit::uninit();
+        static mut BUF: [MaybeUninit<SimpleLinkedListNode<'static, Option<&'static dyn Process>>>;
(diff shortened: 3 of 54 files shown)


hudson-ayers (commented):
I have not done an in-depth review, but let me summarize my current thoughts on this.

This PR allows us to implement the protections in #2773 for basic list types, at the cost of runtime performance on the `push_head` operation (and possibly a little code size?). It also presents a second type which allows manual implementations as before, for greater flexibility on what types can be used for the list, and to allow implementations to opt out of the performance hit of the simple linked list.

The biggest downside is probably the requirement that all accesses to node data come via a call to `node.content()`, which we cannot enforce via the compiler, and thus becomes another easy-to-make mistake that can only be caught by very careful code review.

Personally, I think whether we want this change probably depends on whether we think errors via this new footgun will be more or less common than errors of the type that would be prevented by #2773

lschuermann:
> The biggest downside is probably the requirement that all accesses to node data come via a call to `node.content()`, which we cannot enforce via the compiler, and thus becomes another easy-to-make mistake that can only be caught by very careful code review.

We can entirely remove this requirement. The `content()` method exists such that you can have list containers which are disparate from the actual contents of the list. This seems useful for a multitude of reasons, in particular consistency checks. However, nothing is stopping one from just using the same type for the list containers and content in the case of the `GenericLinkedList`.

The requirement to call `content()` regardless of whether the list container type is the same than the content type was introduced to allow lazy allocation and/or evaluation of the content. However, if we decide that the content has to be ready all the time, it's completely fine to state that as such.

hudson-ayers:
> > The biggest downside is probably the requirement that all accesses to node data come via a call to `node.content()`, which we cannot enforce via the compiler, and thus becomes another easy-to-make mistake that can only be caught by very careful code review.
> 
> We can remove this requirement entirely. The `content()` method exists such that you can have list containers which are disparate from the actual contents of the list. This seems useful for a multitude of reasons, in particular consistency checks. However, nothing is stopping one from just using the same type for the list containers and content in the case of the `GenericLinkedList`.
> 
> The requirement to call `content()` regardless of whether the list container type is the same than the content type was introduced to allow lazy allocation and/or evaluation of the content. However, if we decide that the content has to be ready all the time, it's completely fine to state that as such.
> 
> However, if the consumer of the list is not requiring a specific `ListNode` type but is instead generic over the `ListNode` (taking in a generic `L: ListNode<'a, C>` with fixed content type `C`), it will have to assume that `L != C` and thus need to call `content()` to get a `&C` reference anyways.

I think I would advocate for removing the requirement to call `content()` in that case. While I see why that requirement could be useful I have grown very wary of requirements like this where in the common case not following the requirement does not break anything, but might lead to problems eventually.

bradjc:
@alevy?

This seems good to me.

lschuermann:
I ported the rest of the Tock codebase over to this new list interface, but did not change existing usages of `List` in capsules to use the new `SimpleLinkedList`, but rather `GenericLinkList` which serves as a replacement for the previous `List` type. There still seem to be some bugs which I'll need to investigate.

That said, looking at the schedulers which have been ported to the new interface, it might make sense to also consider questions of efficiency there. While it is rather important that schedulers don't have duplicate entries of processes in their queues (in part justifying the transition to `SimpleLinkedList` and having `push_head` and `push_tail` check for duplicates during construction of the process list), when performing acutal scheduling decisions, these checks can introduce a considerable amount of overhead. Thus, perhaps we should furthermore provide deliberately unchecked variants of these functions, for code such as the following:

https://github.com/tock/tock/blob/85472f4673fb76509978bdb09267731b91792442/kernel/src/scheduler/cooperative.rs#L72-L75

With both the new and current list infrastructures, these are operations in `O(n)`. If we are to provide `push_{head,tail}_unchecked` methods, it is furthermore trivial to introduce a new but API-compatible type which maintains a head _and_ tail pointer, making scheduling decisions operations in `O(1)` (for simpler schedulers, such as round robin).

hudson-ayers:
> I ported the rest of the Tock codebase over to this new list interface, but did not change existing usages of `List` in capsules to use the new `SimpleLinkedList`, but rather `GenericLinkList` which serves as a replacement for the previous `List` type. There still seem to be some bugs which I'll need to investigate.
> 
> That said, looking at the schedulers which have been ported to the new interface, it might make sense to also consider questions of efficiency there. While it is rather important that schedulers don't have duplicate entries of processes in their queues (in part justifying the transition to `SimpleLinkedList` and having `push_head` and `push_tail` check for duplicates during construction of the process list), when performing acutal scheduling decisions, these checks can introduce a considerable amount of overhead. Thus, perhaps we should furthermore provide deliberately unchecked variants of these functions, for code such as the following:
> 
> https://github.com/tock/tock/blob/85472f4673fb76509978bdb09267731b91792442/kernel/src/scheduler/cooperative.rs#L72-L75
> 
> With both the new and current list infrastructures, these are operations in `O(n)`. If we are to provide `push_{head,tail}_unchecked` methods, it is furthermore trivial to introduce a new but API-compatible type which maintains a head _and_ tail pointer, making scheduling decisions operations in `O(1)` (for simpler schedulers, such as round robin).

I think having unchecked methods for use by the schedulers is probably important

hudson-ayers:
> If the List type is generic, then I'd suggest having two versions, one which has the unchecked methods (for when computational overhead is important) and one which doesn't (for when it isn't). Otherwise I worry that we have these dangerous methods lying around. Or perhaps mark them `unsafe`?

The "unchecked" methods don't allow for memory unsafety so I do no think `unsafe` would be the right approach here. They just protect you from ending up in an infinite loop, but that is something anyone can do using just `while(1)` anyway. I think having `unchecked` in the name is sufficient indication that it should be used carefully.

> Alternatively, how many processes do we expect to have? I've always assumed, with the current scheduler and kernel architecture, as well as resource constraints, ten or so at most. If that's the case, can these schedulers be written to use a NUM_PROCS sized array for their state, rather than a linked list?

In my initial scheduler design I did in fact use an array, but at the time const generics had not come far enough to build a useful design that could be generic on the number of processes. That is likely no longer the case and may be good to revisit now. However, for some schedulers like MLFQ the use of lists is much more straightforward -- using arrays for multiple queues you would waste space since each queue would need to be big enough to hold all processes at once.

> That being said, I think it's not necessary to iterate across the list to see if the added node can be added. This solves the problem of the node being added to the same list twice, but not if it's accidentally added to two different lists. Why not instead check if its `next` is `Some` and have a reserved value for the tail (to distinguish the last node from a node that is not in a list)? Alternatively, define an enum for the next field which has `None`, `Some()`, and `End`.

Adding a node to two different lists generally shouldn't be a problem, as doing so would not easily lead to an infinite loop? Am I missing something?

lschuermann:
> That being said, I think it's not necessary to iterate across the list to see if the added node can be added. This solves the problem of the node being added to the same list twice, but not if it's accidentally added to two different lists. Why not instead check if its `next` is `Some` and have a reserved value for the tail (to distinguish the last node from a node that is not in a list)? Alternatively, define an enum for the next field which has `None`, `Some()`, and `End`.

This is an excellent point. Adding a node to two lists can cause inconsistencies in this list, such as infinite loops, but it'd also make it impossible to retain a tail-reference for efficiently appending tail elements to a list. I like the proposed approach, and would prefer if we could get away without using another word of memory, but I will implement the simpler proposed approach for now.

> The "unchecked" methods don't allow for memory unsafety so I do no think `unsafe` would be the right approach here. They just protect you from ending up in an infinite loop, but that is something anyone can do using just `while(1)` anyway. I think having `unchecked` in the name is sufficient indication that it should be used carefully.

Agreed.

> [...] However, for some schedulers like MLFQ the use of lists is much more straightforward [...]

Right. And if we _are_ storing a tail-reference for these singly-linked lists, singly linked lists seem like a very efficient data structure to represent these operations.

bradjc:
Are we still pursuing this?

And do we need the changes to every board file?

bradjc:
Closed without prejudice, please re-open if there are future updates.

PR #2849, opened 2021-09-29
add a document on how to reduce Tock code size

hudson-ayers:
### Pull Request Overview

This pull request adds a document on how to reduce Tock code size, which contains much of the information from my presentation at the core working group meeting last week.

### Testing Strategy

This pull request was tested by spellcheck

### TODO or Help Wanted

N/A

### Documentation Updated

- [x] This is documentation!

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/CodeSize.md
+++ b/doc/CodeSize.md
@@ -1,0 +1,259 @@
+# Minimizing Tock code size
+
+Many embedded applications are ultimately limited by the flash space available on
+the board in use. This document provides tips on how to write Rust code such that
+it does not require an undue amount of flash, and highlights some options which
+can be used to reduce the size required for a particular image.
+
+## Code Style: tips for keeping Rust code small
+
+#### When to use generic types with trait bounds versus trait objects (`dyn`)
+
+Polymorphic structs and functions are one of the biggest sources of bloat in Rust
+binaries -- use of generic types can lead to bloat from monomorphization, while
+use of trait objects introduces vtables into the binary and limits opportunities
+for inlining.
+
+Use `dyn` when the function in question will be called with multiple concrete types;
+otherwise code size is increased for every concrete type used (monomorphization).
+
+```rust
+fn set_gpio_client(&dyn GpioClientTrait) -> Self {//...}
+
+// elsewhere
+let radio: Radio = Radio::new();
+set_gpio_client(&radio);
+
+let button: Button = Button::new();
+set_gpio_client(&button);
+```
+
+Use generics with trait bounds when the function is only ever called with a single public type
+per board; this reduces code size and run time cost. This increases source code
+complexity for decreased image size and decreased clock cycles used.
+
+```rust
+// On a given chip, there is only a single FlashController. We use generics so
+// that there can be a shared interface by all FlashController's on different chips,
+// but in a given binary this function will never be called with multiple types.
+impl<'a, F: FlashController> StorageDriverBackend<'a, F> {
+    pub fn new(
+        storage: &'a StorageController<'a, F>,
+    ) -> Self { ... }
+
+```
+
+Similarly, only use const generics when there will not be monomorphization, or if the body of the
+method which would be monomorphized is sufficiently small that it will be inlined anyways.
+
+#### Non-generic-inner-functions
+
+Sometimes, generic monomorphization is unavoidable (much of the code in grant.rs is an example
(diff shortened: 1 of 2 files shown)


doc/CodeSize.md
```diff
@@ -0,0 +1,260 @@
...
+let radio: Radio = Radio::new();
+set_gpio_client(&radio);
+
+let button: Button = Button::new();
+set_gpio_client(&button);
+```
+
+Use generics with trait bounds when the function is only ever called with a single public type;
+This reduces code size and run time cost. This increases source code complexity for decreased
+image size and decreased clock cycles used.
```
jrvanwhy, line 33: The wording "when the function is only ever called with a single public type" may be confusing. The "single public type" part of the phrase is per board (or per binary in the case of `libtock-rs` apps). E.g. if Imix calls it with type `A` and Hail calls it with type `B` it should still be generic argument rather than a `dyn` argument.

doc/CodeSize.md
```diff
@@ -0,0 +1,260 @@
...
+    }
+}
+```
+
+Similarly, avoid code that could divide by 0, avoid using `unwrap()` / `expect()`,
+and make sure to give the compiler enough information that it can guarantee `copy_from_slice()`
+is only being called on two slices of equal length.
+
+
+#### Non-generic-inner-functions
```
jrvanwhy, line 87: Do we want to move "Non-generic-inner-functions" to be right after the heading on picking generics versus `dyn`? They seem closely related and perhaps they belong together.

doc/CodeSize.md
```diff
@@ -0,0 +1,260 @@
...
+    if self.my_array.get(4).ok_or(ErrorCode::FAIL)? == 7 {
+        *(self.other_array.get_mut(3).ok_or(ErrorCode::FAIL)?) = false;
+        Ok(())
+    } else {
+        Err(ErrorCode::SIZE)
+    }
+}
+```
+
+Similarly, avoid code that could divide by 0, avoid using `unwrap()` / `expect()`,
```
jrvanwhy, line 82: Signed integer division has a second error case besides division by `0`: dividing the type's `MIN` value by `-1`.

Are all the divisions you encountered unsigned divisions?

doc/CodeSize.md
```diff
@@ -0,0 +1,260 @@
...
+pub const GPIO_PIN_1: Pin = Pin::new(1);
+pub const GPIO_PIN_2: Pin = Pin::new(2);
+// ...and so on.
+```
+The latter approach ensures that the compiler can remove pins which are not used from the binary.
+
+#### Combine register accesses
+Combine register accesses into as few volatile operations as possible. E.g.
+```rust
+regs.dcfg.modify(DevCibfugLLDEVSPD::FullSpeed1_1);
```
brghena, line 177:
```suggestion
regs.dcfg.modify(DevConfig::DEVSPD::FullSpeed1_1);
```

Looks like your right hand was one column too far left there... : )

bradjc:
I'm curious about the state of #1663, and if this PR means we should mark that issue as resolved.

doc/CodeSize.md
hudson-ayers: I have not removed many division-sourced panics, it is very possible that the few I have were all unsigned divisions. I added this note to the doc though, thank you.

doc/CodeSize.md
hudson-ayers: I added "per board"

doc/CodeSize.md
hudson-ayers: Thanks! fixed

hudson-ayers:
> I'm curious about the state of #1663, and if this PR means we should mark that issue as resolved.

I think there are still some ideas there that we have not applied and which are still worth considering, namely:
- building a non-relocatable kernel
- disabling frame pointers on arm (I think this should work now)
- no_debug building mode
- continuing to cut down use of &dyn in the core kernel

bradjc:
Can you add a link from `doc/README.md`?

hudson-ayers:
> Can you add a link from `doc/README.md`?

Done

jrvanwhy (approved):

bradjc (approved):

alistair23 (approved):

bradjc:
We can always continue to improve this doc with follow up PRs as needed.

bors r+

PR #2850, opened 2021-09-29
OpenTitan: Application fixups

alistair23:
### Pull Request Overview

There have been some recent regressions to running applications on OpenTitan. This PR fixes those.

### Testing Strategy

Running usersapce apps on OT.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/epmp.rs
+++ b/arch/rv32i/src/epmp.rs
@@ -665,13 +665,23 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::MPU
                         } else {
                             (16, 0x0000_FFFF)
                         };
+                        // Fully clear the PMP config
                         csr::CSR.pmpconfig_set(
                             x / 2,
-                            (disable_val | cfg_val << 8) << region_shift
+                            (disable_val << region_shift)
                                 | (csr::CSR.pmpconfig_get(x / 2) & other_region_mask),
                         );
+
+                        // Set the address *before* we enable the config
+                        // Otherwise this could take effect and block the kernel from running
                         csr::CSR.pmpaddr_set(x * 2, (start) >> 2);
                         csr::CSR.pmpaddr_set((x * 2) + 1, (start + size) >> 2);
+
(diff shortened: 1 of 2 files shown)


arch/rv32i/src/epmp.rs
```diff
@@ -667,11 +667,17 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::MPU
...
                                 | (csr::CSR.pmpconfig_get(x / 2) & other_region_mask),
                         );
+
                         csr::CSR.pmpaddr_set(x * 2, (start) >> 2);
                         csr::CSR.pmpaddr_set((x * 2) + 1, (start + size) >> 2);
+
+                        csr::CSR.pmpconfig_set(
+                            x / 2,
+                            (cfg_val << 8) << region_shift | (csr::CSR.pmpconfig_get(x / 2)),
+                        );
```
hudson-ayers, line 680: did I break something here in https://github.com/tock/tock/pull/2757 ?

alistair23: Yep, for me apps stopped working when using ePMP

alistair23: The problem is that we could have junk left over in the address register. As the high priority PMP configs are used for the apps they will take effect before the kernel configs (which are the lowest priority).

We have disabled the pmp regions in `disable_app_mpu()`. If we enable them before writing the address we can lock the kernel out of execute permissions. So we have to make sure we set the address first

alistair23: I added more comments here

bradjc (approved):

hudson-ayers (approved):

arch/rv32i/src/epmp.rs
hudson-ayers: Ah, I understand. I wanted to make sure this breakage would not also apply for the normal pmp file, sounds like it should not. Thanks for the explanation!

ppannuto (commented):
bors r+

PR #2852, opened 2021-09-30
kernel: utilities: Initial commit of mut_imut_buffer

alistair23:
### Pull Request Overview

There are a few times inside the Tock kernel where we want to support either a `'static mut` buffer or a `'static` buffer. Examples of this include the [RSA key support](https://github.com/tock/tock/pull/2839) but also the [OTBN](https://github.com/tock/tock/blob/master/chips/lowrisc/src/otbn.rs#L153).

In order to support both I have added a new enum called MutImutBuffer. This allows us to pass around either a mutable static buffer or an immutable static buffer. This can be passed all the way down to the chip driver which can then unpack the buffers and write the data to hardware.

### Testing Strategy

CI and RSA key implementation

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/kernel/src/utilities/mod.rs
+++ b/kernel/src/utilities/mod.rs
@@ -4,6 +4,7 @@ pub mod binary_write;
 pub mod helpers;
 pub mod leasable_buffer;
 pub mod math;
+pub mod mut_imut_buffer;
 pub mod peripheral_management;
 pub mod static_init;
 pub mod storage_volume;
--- a/kernel/src/utilities/mut_imut_buffer.rs
+++ b/kernel/src/utilities/mut_imut_buffer.rs
@@ -1,0 +1,54 @@
+//! An enum that can contain a reference to either a mutable or
+//! an immutable buffer.
+//!
+//! This type is intended for internal use in implementations of HILs
+//! or abstractions which need to handle both mutable and immutable
+//! buffers.
+//!
+//! One motivating use case is keys for public key
+//! cryptography. Public keys are often distributed as constant values
+//! in the flash of a kernel image (e.g., to verify signatures), which
+//! requires they be immutable.  Copying them to RAM is expensive
+//! because these keys can be very large (e.g., 512 bytes for a
+//! 4096-bit RSA key). At the same time, some clients may use
+//! dynamically generated or received keys, which are stored in
+//! mutable RAM. Requiring that keys be immutable would
+//! discard mut on this memory. An
+//! implementation can use this type to store either mutable and
+//! immutable buffers. The OTBN (OpenTitan Big Number accelerator) is
+//! one example use of MutImutBuffer.
+//!
+//! Because this type requires dynamic runtime checks that types
+//! match, it should not be used in any HILs or standard, external
+//! APIs. It is intended only for internal use in implementations.
+//!
+//! Author: Alistair Francis
+//!
+//! Usage
+//! -----
+//!
+//!  ```rust
+//! use kernel::utilities::mut_imut_buffer::MutImutBuffer;
+//!
+//! let mut mutable = ['a', 'b', 'c', 'd'];
+//! let immutable = ['e', 'f', 'g', 'h'];
+//!
+//! let shared_buf = MutImutBuffer::Mutable(&mut mutable);
+//! let shared_buf2 = MutImutBuffer::Immutable(&immutable);
+//!  ```
+
+/// An enum which can hold either a mutable or an immutable buffer
+pub enum MutImutBuffer<'a, T> {
+    Mutable(&'a mut [T]),
+    Immutable(&'a [T]),
+}
+
+impl<'a, T> MutImutBuffer<'a, T> {
+    /// Returns the length of the underlying buffer
+    pub fn len(&self) -> usize {
+        match self {
+            MutImutBuffer::Mutable(buf) => buf.len(),
+            MutImutBuffer::Immutable(buf) => buf.len(),
+        }
+    }
+}


bradjc (dismissed):
This seems like a very intuitive and straightforward approach. If it ends up being a bad idea, well at least we learned something along the way.

I'm not a huge fan of the abbreviations in the name, however.

phil-levis:
This came up on the core call today, it is very relevant:

https://github.com/tock/libtock-rs/issues/334

lschuermann (changes requested):
I think that this is indeed pretty useful, however I'm also not entirely happy with the name. I'd like to propose to make it more generic by not even taking a reference to a slice of `T`, but just a reference to `T` which then can (but not has to be) a slice.

A more appropriate name could then be `AnnotatedMutability`.

lschuermann:
I'll have the notes up next week (moving currently), but here is a fresh summary of what we've talked about on the call.

This seems useful for the use cases described here and potentially more. However, specifically talking about the use case here, it might make sense to see whether there is any hardware which actually requires persistent (asynchronous) access to crypto keys in the first place. If there is some, great, this seems like a very straightforward approach to cover these use cases. However, if access to crypto keys is only required _synchronously_ (that is, for example, writing the keys into a specific DMAable region of memory), then it might just be sufficient to pass an immutable reference of reduced lifetime or a high-level type which can provide the keys in a closure.

If this does make it in, it might have implications for other parts of the OS. For instance, should `uart::Transmit::transmit_buffer` then also take an instance of this type? It seems that this opens to a broader discussion and we'd likely want to carefully think about it usage and specify some guidelines to that. For instance, you'd try to never accept this type for an API where buffers would need to be written to -- this is important to reduce the number of failure cases which can occur at runtime.

@phil-levis @hudson-ayers and others on the call, I hope this is an objective interpretation of what we've talked about. Please correct me if I'm wrong. :)

bradjc:
> I think that this is indeed pretty useful, however I'm also not entirely happy with the name. I'd like to propose to make it more generic by not even taking a reference to a slice of `T`, but just a reference to `T` which then can (but not has to be) a slice.

Is there a use case where the type wouldn't be a buffer? I think it would be aligned with Tock ethos to start with what we have a use case for, and see if we need to make it more generic in the future.

lschuermann:
> I think it would be aligned with Tock ethos to start with what we have a use case for, and see if we need to make it more generic in the future.

I agree with that statement per se.

However I'd argue that the underlying Rust fundamentals, interactions and especially considerations of soundness and usefulness of this struct in the kernel, with respect to how Tock manages references to values (of any type) passed around the kernel, are valid regardless of whether we're dealing with buffers. I'm not sure whether it makes sense to then constrain such a generic concept to be not a single type, but an arbitrary subset of types.

phil-levis:
Just some random thoughts, I think @alistair23 has hit on a difficult wart/problem.

To touch on Leon's point about the need for asynchronous key access, for OTBN, the key would have to be copied into the OTBN data memory, so wouldn't need asynchronous access. The nRF Cryptocell doesn't allow you to pull keys from flash (they must be in RAM). Are there examples of RSA accelerators which would require asynchronous access to a flash buffer?

Virtualization does raise this issue: even if OTBN requires copying the key into data memory (so could be synchronous), it could be a request is queued on top of OTBN. This request needs to hold onto the key across a callback. 

Given the overhead of setting up OTBN, do we want to virtualize its operations? E.g., if two clients are trying to use the peripheral simultaneously for multiple operations (with different crypto), do we want them to go back and forth, changing the instruction and data memory repeatedly with different ciphers and keys? I'd guess we often don't want this behavior, but we shouldn't preclude it (e.g., with a locking mechanism).

For the virtualizer, could we pay the cost of a single read/write key at the bottom of the virtualizer, which is copied into for the operation? There could be two types of clients: RO keys and RW keys. It might be that internally we have this dynamic structure but clients can keep static typing.

lschuermann:
> Is there a use case where the type wouldn't be a buffer?

@bradjc Yes, there is: `&'static str`. It would likely be very useful to support this for passing around large constant strings from flash around the kernel.

phil-levis:
I think it would be good to see how a client uses RSA when it has RAM buffers. I'm starting to think it make make more sense of the key structure actually has a buffer, as opposed to being a handle around buffers. Two reasons:
  - Clients that use RAM buffers need to allocate their own buffers; they could just allocate key structures. This will be easier to manage than allocating a buffer and having it dynamically moved in and out of a key structure.
  -  The buffers have structure but are not typed. For example, "We store the public exponent and modulus in the public key buffer" and "`private_key` is a buffer containing the private key. The first 256 bytes are the private_exponent (also called `d`).". It's of course fine if internally this is how storage is managed, but it requires clients to know and follow this; bugs would be disastrous. Imposing structure in a type-safe way requires parameters and copying, so would require RAM buffers.

bradjc:
> @bradjc Yes, there is: `&'static str`. It would likely be very useful to support this for passing around large constant strings from flash around the kernel.

But if they are constant, why would anything need to keep a mutable reference?

alistair23:
Let me know if people want me to change the name, `AnnotatedMutability` works for me (or pretty much anything else).

I would prefer to leave this as a buffer. I think Brad is right about not having a use case.

As for @phil-levis 's point about the RAM buffers. That's a good point. The original implementation of https://github.com/tock/tock/pull/2839 was using RAM buffers.

I agree that a type safe way would be nice. Using RAM buffers doesn't really make it any more type safe though. Here is an example using RAM buffers, with the same API: https://github.com/alistair23/tock/blob/19aa9e7533bb271edb605e4c77a6d6fdc95b7677/capsules/src/public_key_crypto/rsa_keys.rs#L168
RAM buffers means internally we can split the components, which is a big plus though.

We can change the HIL to require types (although then it won't be generic for key types). We can do that with static buffers as well though.

On top of that getting the values out becomes a problem. In order to copy from an entropy source we need to use `Cell`s (https://github.com/alistair23/tock/blob/19aa9e7533bb271edb605e4c77a6d6fdc95b7677/capsules/src/public_key_crypto/rsa_keys.rs#L63), but then we can't just get the values out (https://github.com/alistair23/tock/blob/19aa9e7533bb271edb605e4c77a6d6fdc95b7677/capsules/src/public_key_crypto/rsa_keys.rs#L237 fails to compile). We instead need to copy the data to a new buffer, which is yet another buffer that we need to store keys.

alistair23:
So the problem is that we don't always have mutability.

Image a struct like this:

```rust
pub struct RSA2048Keys<'a> {
    modulus: [u8; 256],          // Also called n
    public_exponent: u32,        // Also called e
    private_exponent: [u8; 256], // Also called d
...
}
```

We can import a key with a function like this:

```rust
fn import_public_key(&mut self, public_key: &[u8]) -> Result<(), ErrorCode>
```

Where we copy a buffer into our own internal state, because we are mutable.

The problem is then generating a key.

In order to generate a key we are going to call `entropy::Entropy32::get()`, and then wait for the 

```rust
    fn entropy_available(
        &self,
        entropy: &mut dyn Iterator<Item = u32>,
        error: Result<(), ErrorCode>,
    ) -> Continue;
```

callback.

At this point the `RSA2048Keys` struct is not mutable! So we can't copy the data into our internal structures.

So the above doesn't work.

Next step is to use `Cell` in the internal state. This way the `RSA2048Keys` struct doesn't need to be mutable, we have interior mutability instead.

That works for generating the keys, but how do we then get them out?

Like you said, we can copy them out, but that's not very practical.

We could use closures, but that's not ideal either. We don't want to pass `RSA2048Keys` down to the chip driver implementation. The idea is to hide all the complex key operations and management in a capsule, so then the drivers can be simple. The hope is there is only one place we could mix up keys and that is easier to review them a number of different implementations. A closure would involve passing the `RSA2048Keys` struct to the driver, which then uses the closure. It also means we need static mutable allocations of `RSA2048Keys` to pass around everywhere.

We can use `TakeCell`, but that is basically what this implementation is anyway.

alistair23:
I have updated the PR description. There are now two justifications for this enum, I think this PR should be merged

phil-levis (changes requested):
Following up on the OT WG call today. I worked through the RSA stack and agree that it would be good to add mut_imut to the kernel. However, this type SHOULD NOT be used in HILs, because it pushes dynamic error checking to clients. Instead, the type is intended for internal use in implementations.  I'll make a PR with these changes in its documentation. Once we polish that it should be good to go.

phil-levis:
How's this?

```rust
//! An enum that can contain either a reference to either a mutable or
//! an immutable buffer.
//!
//! This type is intended for internal use in implementations of HILs
//! or abstractions which need to handle both mutable and immutable
//! buffers.  One motivating use case is keys for public key
//! cryptography. Public keys are often distributed as constant values
//! in the flash of a kernel image (e.g., to verify signatures), which
//! requires they be immutable.  Copying them to RAM is expensive
//! because these keys can be very large (e.g., 512 bytes for a
//! 4096-bit RSA key). At the same time, some clients may use
//! dynamically generated or received keys, which are stored in
//! mutable RAM. Requiring that keys be immutable would for clients to
//! discard mut on this memory and so not be able to reset keys. An
//! implementation can use this type to store either mutable and
//! immutable buffers. The OTBN (OpenTitan Big Number accelerator) is
//! one example use of MutImutBuffer.
//!
//! Because this type requires dynamic runtime checks that types
//! match, it should not be used in any HILs or standard, external
//! APIs.  It is intended only for internal use in implementations.
//!
//! Author: Alistair Francis
//!
//! Usage
//! -----
//!
//!  ```rust
//! use kernel::utilities::mut_imut_buffer::MutImutBuffer;
//!
//! let mut mutable = ['a', 'b', 'c', 'd'];
//! let immutable = ['e', 'f', 'g', 'h'];
//!
//! let shared_buf = MutImutBuffer::Mutable(&mut mutable);
//! let shared_buf2 = MutImutBuffer::Immutable(&immutable);
//!  ```
```

alistair23:
Sounds good! I have updated the PR

phil-levis (approved):

alistair23:
This should be ready to merge

hudson-ayers (approved):
bors r+

alistair23:
bors retry

PR #2853, opened 2021-09-30
make: remove relocation model flag

bradjc:
### Pull Request Overview

This flag was added in the initial switch to the LLVM linker (b5d7f8901cecd85d9eab0a50f4ba5c0eb82b4a91). I have no idea why I thought this was necessary or where I copied it from.

Removing based on this comment: https://github.com/tock/tock/issues/1663#issuecomment-613745416.

Running on CI to see if it makes a difference or breaks anything.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers:
The benchmarks seem to have some wild results here:

> earlgrey-cw310: flash use +1292 bytes (+1.02%), RAM use -10240 bytes(-50.04%)

> imix: flash use +4416 bytes (+2.52%), RAM use +0 bytes(+0.00%)

hudson-ayers:
This got me reading the documentation on relocation models: https://doc.rust-lang.org/rustc/codegen-options/index.html#relocation-model . It seems that dynamic-no-pic is almost certainly wrong, and that static sounds right. However, for most boards if you do not specify a relocation model the default is pic. So I think we should be using `-C relocation-model=static` -- which has identical code size and RAM use as what we have now, but at least seems to actually indicate the linking we are using.

ppannuto (approved):

alistair23 (approved):

PR #2854, opened 2021-09-30
Run board tests as part of the CI

alistair23:
### Pull Request Overview

This PR runs the board tests as part of the CI. This includes building all tests and running the OpenTitan QEMU tests.

This should avoid breakages in the tests being introduced.

Before we can enable the tests though we need to fix the current tests. This involves a few changes to fix the tests.

### Testing Strategy

CI

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


Makefile
```diff
@@ -302,7 +302,8 @@ ci-runner-github-tests:\
 	ci-job-capsules\
 	ci-job-chips\
 	ci-job-tools\
-	ci-job-miri
+	ci-job-miri\
+	ci-job-tests
```
ppannuto, line 306: If I understand correctly, this is just a compilation test right?

I think I should probably go under the `github-build` in that case (it's similar in spirit to the `syntax` or `compilation` run).

`-tests` is pretty generic. Can we rename to `ci-job-qemu-compile-only` or something like that?

ppannuto (changes requested):
Looks fine other than name and which group it's placed in

Makefile
alistair23: Yeah, it's just a compilation test of the `cargo test`. I'll move the section

It's not a qemu test, so I don't think that's the right name. I'll change it to `ci-job-cargo-test`

alistair23:
I'm seeing weird build failures, but only on GitHub:

```

make[1]: Entering directory '/home/runner/work/tock/tock/boards/opentitan/earlgrey-cw310'
mkdir -p /home/runner/work/tock/tock/target/riscv32imc-unknown-none-elf/release/deps/
   Compiling core v0.0.0 (/usr/share/rust/.rustup/toolchains/nightly-2021-09-08-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core)
   Compiling rustc-std-workspace-core v1.99.0 (/usr/share/rust/.rustup/toolchains/nightly-2021-09-08-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/rustc-std-workspace-core)
   Compiling earlgrey-cw310 v0.1.0 (/home/runner/work/tock/tock/boards/opentitan/earlgrey-cw310)
   Compiling compiler_builtins v0.1.49
   Compiling tock-registers v0.7.0 (/home/runner/work/tock/tock/libraries/tock-register-interface)
error: duplicate lang item in crate `core` (which `rustc_std_workspace_core` depends on): `bool`.
  |
  = note: the lang item is first defined in crate `core` (which `tock_registers` depends on)
  = note: first definition in `core` loaded from /home/runner/work/tock/tock/target/riscv32imc-unknown-none-elf/release/deps/libcore-35973b8fe6728954.rmeta
  = note: second definition in `core` loaded from /home/runner/work/tock/tock/target/riscv32imc-unknown-none-elf/release/deps/libcore-6ebb67c9463a23dd.rmeta
```

ppannuto (approved):

boards/opentitan/earlgrey-nexysvideo/test_layout.ld
```diff
@@ -0,0 +1,26 @@
+MEMORY
```
bradjc, line 1: It would be helpful to have a hint here about why these are needed.

bradjc (dismissed):

boards/opentitan/earlgrey-nexysvideo/Makefile
```diff
@@ -43,15 +43,15 @@ ifneq ($(OPENTITAN_TREE),)
...
 test-hardware:
 ifeq ($(OPENTITAN_TREE),)
 	$(error "Please ensure that OPENTITAN_TREE is set")
 endif
 	mkdir -p $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/deps/
-	$(Q)cp layout.ld $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/deps/
+	$(Q)cp test_layout.ld $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/deps/layout.ld
 	$(Q)cp ../../kernel_layout.ld $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/
-	$(Q)RUSTFLAGS="$(RUSTC_FLAGS_TOCK)" $(CARGO) test $(CARGO_FLAGS_TOCK) --bin $(PLATFORM) --release --features=hardware_tests
+	$(Q)RUSTFLAGS="$(RUSTC_FLAGS_TOCK)" $(CARGO) test $(CARGO_FLAGS_TOCK_NO_BUILD_STD) $(NO_RUN) --bin $(PLATFORM) --release --features=hardware_tests
```
hudson-ayers, line 57: why does this need to be the NO_BUILD_STD flags?

hudson-ayers (dismissed):

boards/opentitan/earlgrey-nexysvideo/Makefile
alistair23: We don't want to build the std library, it fails to build

hudson-ayers: Do you know why it fails to build? I find it a little surprising -- is it something to do with that option not working with `cargo test`?

alistair23: You can see the failures here: https://github.com/tock/tock/runs/3762108392

lschuermann: All of these errors appear to originate from `tock-registers`. This doesn't look good and we'll want to look into this anyways.

hudson-ayers: I think that some of the tests might require the standard library (not just core), so build-std=core is insufficient. In that case I think what Alistair did here makes sense.

PR #2857, opened 2021-10-04
Remove all uses of `expect("xx")`, replace with `unwrap()`

hudson-ayers:
### Pull Request Overview

This PR removes all uses of `expect()` in non-test code. Now that `\#[track_caller]` is enabled by default, `expect()` is not necessary to make it possible to backtrace panics from calls to `unwrap()`. Continuing to use `expect()` has a code size cost, as these messages are embedded in the binary. Worse, even if you remove the panic handler, the strings remain embedded in the binary, making them a truly unavoidable cost.

One place where this is tricky is `OptionalCell` -- historically, Tock did not include an implementation of `unwrap()` on `OptionalCell` to discourage developers from writing code that would panic. This PR replaces `OptionalCell::expect()` with `OptionalCell::unwrap_or_panic()` in hopes that the non-canonical name will still discourage its use.

To prevent this change from removing useful comments, I embedded the strings passed to `expect()` in comments following the new call to `unwrap()`. The mechanism for this PR was as follows:

```shell
$ vim
:args **/*.rs
:argdo %s!expect("\(.*\)")\(.*\)$!unwrap()\2 // Unwrap fail = \1!g
:wall 
:qall
```

This PR reduces Imix code size by 372 bytes.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

- [x] I will update the code size doc if this PR is accepted.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/clue_nrf52840/src/main.rs
+++ b/boards/clue_nrf52840/src/main.rs
@@ -676,9 +676,7 @@ pub unsafe fn main() {
     );
     base_peripherals.ecb.set_client(aes_mux);
     aes_mux.initialize_callback_handle(
-        dynamic_deferred_caller
-            .register(aes_mux)
-            .expect("no deferred call slot available for ccm mux"),
+        dynamic_deferred_caller.register(aes_mux).unwrap(), // Unwrap fail = no deferred call slot available for ccm mux
     );
 
     let serial_num = nrf52840::ficr::FICR_INSTANCE.address();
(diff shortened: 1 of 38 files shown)


bradjc (approved):
It feels like we kind of go back and forth on this (i.e. using expect and debuggability), but we seem to be digging into code size at the moment, and this is aligned with that, so seems like a good change. And if the default panics had been what they are now, perhaps we wouldn't have switched a few things to expect.

Anytime a PR highlights how often this happens in capsules it is concerning, but of course the issue pre-dates this PR. I did go through and remove the unwraps from two capsules, and I think LEDs can happen too in a different PR (the change is a bit more invasive).

PR #2858, opened 2021-10-04
update rust to nightly-2021-10-03

hudson-ayers:
### Pull Request Overview

This pull request updates Rust to yesterday's nightly, and adds an `allow()` for 14 dead code warnings that surface in the tbf header module.

I was motivated to submit this because upstream Rust recently switched over to LLVMs new pass manager (https://github.com/rust-lang/rust/pull/88243) which seems to be able to make better decisions about inlining functions thanks to its bottom-up pass approach ensuring that callees are always maximally optimized by the time callers are being optimized.

This change reduces code size by 1232 bytes on Imix.

I realize that two rust updates in a month is more frequently than we normally update, so if we want to wait a while I understand.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alistair23:
I think it makes sense to update again this month if there is something we want. You need to run the update script though

bradjc (changes requested):
Yeah let's update rust whenever we need new features.

But do need to run the script.

hudson-ayers:
I ran the script, and went with 2021-10-05 because that is what the script suggested

bradjc (approved):

alistair23 (approved):

ppannuto (approved):

Issue #2851, opened 2021-09-29
time hils over specify lifetime on &self parameter in functions


jettr:
Taking a look at the `Alarm` hil, it currently specifies `'a` for the `self` parameter on the `fn set_alarm_client(&'a self, client: &'a dyn AlarmClient);` function, but self should use the normal, unspecified, anonymous lifetime for self instead. The client parameter still should have `'a` since that is the reference that `Alarm<'a>` needs to hold on to.

The same thing happens on the `Timer` trait:
 * `fn set_timer_client(&'a self, client: &'a dyn TimerClient);`
 * `fn oneshot(&'a self, interval: Self::Ticks) -> Self::Ticks;`
 * `fn repeating(&'a self, interval: Self::Ticks) -> Self::Ticks;`

The `self` parameter should not require the `'a` lifetime.

Also worth noting that the `Alarm` trait already does not specify a lifetime constraint on `fn set_alarm(&self, reference: Self::Ticks, dt: Self::Ticks);`. So it isn't technically required to engage the alarm or timer.

So the next questions is why and how did we get here, and can we just remove the `'a`?

After tracing the code, it appears that we added `'a` on `self` for these methods because the implementation of `VirtualMuxAlarm` and `VirtualMuxTimer` used linked listed internally. These linked lists require a valid reference for the lifetime of the list, namely `'a`. So I believe the reason `'a` is on the hil is due to leaking implementation details of these commonly used VirtualMux implementations. 

We should stop leaking these implementation details on this hil. How do we fix it?

Once way to fix (i.e. drop the `'a` on the hil interfaces) is to require that after a  `VirtualMuxAlarm` has been created (which takes in a refernce to `MuxAlarm`) we call a new `link(virtual_alarm: &'a VirtualAlarmMux)` on `MuxAlarm` after creating the `VirtualMuxAlarm::new()`. This requires a two step initialization for `VirtualAlarmMux` and could be prone to mistakes, but it alleviates the lifetime issue. Could we perhaps use the `Component` and `finalize` paradigm to ensure that VirtualMuxAlarms are always created correctly, i.e `VirtualAlarmMux::new` then `MuxAlarm::link`?

Also there is a comment on `VirtualSchedulerTimer`:
```
        // However, because of how the MuxAlarm code is written, if the passed
        // alarm is a VirtualMuxAlarm, we must register as a client of the
        // MuxAlarm in order to guarantee that requested interrupts are not
        // dropped.
```
[...]

PR #2866, opened 2021-10-08
Remove 'a lifetime constraint on &self parameters in Time hil

jettr:
### Pull Request Overview

Add a `setup()` function on `VirtualMuxAlarm` that links that instance to the mux. We use new `setup` function instead of relying on `set_alarm_client` to link the virtual alarm to the mux. This allows us to remove the `'a` lifetime constraint on the time hil, which was an artifact of the virtual alarm implementation.

This addresses only the `Alarm` trait from #2851. I will upload another PR for the `Timer` trait fix which will be very similar.

### Testing Strategy

Rely on compiler and CI

### Documentation Updated

- [x] Updated pending time TRD to remove `'a` lifetime on `set_alarm_client` function

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/esp32-c3-devkitM-1/src/main.rs
+++ b/boards/esp32-c3-devkitM-1/src/main.rs
@@ -217,7 +217,6 @@ unsafe fn setup() -> (
         VirtualSchedulerTimer<esp32::timg::TimG<'static>>,
         VirtualSchedulerTimer::new(timer1)
     );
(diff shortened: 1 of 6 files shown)


ppannuto:
I'm inferring this was talked about on the call last week, but as the de-facto owner of the time HIL and all things time, I think it'd be nice to get a sign-off/merge from @phil-levis on this one?

phil-levis:
On my queue

PR #2876, opened 2021-10-19
kernel: hil: symmetric_encryption: Use static buffers for AES crypto

alistair23:
### Pull Request Overview

Use `'static` buffers for the AES crypto operations

### Testing Strategy

CI

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/test/aes.rs
+++ b/capsules/src/test/aes.rs
@@ -13,8 +13,8 @@ pub struct TestAes128Ctr<'a, A: 'a> {
 
     key: TakeCell<'a, [u8]>,
     iv: TakeCell<'a, [u8]>,
-    source: TakeCell<'a, [u8]>,
-    data: TakeCell<'a, [u8]>,
+    source: TakeCell<'static, [u8]>,
+    data: TakeCell<'static, [u8]>,
 
     encrypting: Cell<bool>,
     use_source: Cell<bool>,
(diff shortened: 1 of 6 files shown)


alistair23:
Everything else is moving towards `'static' buffers and removing the other lifetimes, see https://github.com/tock/tock/pull/2866 for example which is similar to this PR

jettr:
I was under the impression that we avoided specifying `'static` lifetime and instead use `'a` everywhere. This works for static and is more generic. One of the exception is when we truly need `'static` is when we passes buffer to hardware in an asynchronous manner. This seems like is going against the nature of #1074

alistair23:
> One of the exception is when we truly need `'static` is when we passes buffer to hardware in an asynchronous manner

This is a buffer being passed to hardware for an asynchronous operation.

> This seems like is going against the nature of #1074

I don't see a single other HIL that doesn't use `'static mut` buffers. Even the ones ticked in https://github.com/tock/tock/issues/1074 (like BLE) use `buf: &'static mut [u8]`

> That PR replaces a 'a lifetime with an anonymous lifetime, which is not static.

This also matches https://github.com/tock/tock/pull/2866 by removing the named `'a` lifetime of `&self` and replacing it with an anonymous lifetime.

Different to https://github.com/tock/tock/pull/2866 this PR also converts the buffers to a `'static` lieftime, as they need to be static in order to be passed down to the hardware. https://github.com/tock/tock/pull/2866 doesn't have buffers to change.

PR #2884, opened 2021-11-01
processbuffer: fix miri issue with transmuting &u8 to &Cell<u8>

### LINKED ISSUE #2882: Unsoundness transmuting &[u8] to &[Cell<u8>] when handling read-only memory
kernel/src/processbuffer.rs defines `impl<'a> From<&'a [u8]> for &'a ReadableProcesSlice` which contains only a transmute. However, this is unsound because `ReadableProcessByte`, of which `ReadableProcessSlice` is a slice of, is `Cell<u8>`, enabling interior mutability. Transmuting from a shared read-only reference to a shared read-write reference is UB regardless of whether you write or not, because you can violate the constness invariant of other shared references that may exist at the time with safe code (other than the `transmute`).

If you add the following to the bottom and run `cargo miri test`, it will fail:
```rust
#[cfg(test)]
mod test {
    use super::*;

    #[test]
    fn test_ro_slice_into_readable_process_slice() {
        let data = [10u8; 10];
        let out: &ReadableProcessSlice = (&data[..]).into();
    }
}
```

The failure:
```
test processbuffer::test::test_ro_slice_into_readable_process_slice ... error: Undefined Behavior: trying to reborrow for SharedRea
dWrite at alloc96722, but parent tag <241883> does not have an appropriate item in the borrow stack                                
   --> kernel/src/processbuffer.rs:620:18                                                                                          
    |                                                                                                                              
620 |         unsafe { core::mem::transmute(val) }                                                                                 
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^ trying to reborrow for SharedReadWrite at alloc96722, but parent tag <241883> does
 not have an appropriate item in the borrow stack                                                                                  
    |                                                                                                                              
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the rules it violated are still e
xperimental                                                                                                                        
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information     
```

Maybe instead of using interior mutability, we can somehow force the index operations to cast to a raw pointer before read, disabling optimizations? Another fix would be to change all use of `&[u8]` to `&[Cell<u8>]` directly.

I'd also like to suggest that all uses of `unsafe` require unit tests running under Miri, regardless of a reviewer scan for soundness.

jettr:
### Pull Request Overview

Miri check for UB, and it complains that we are casting an immutable `&u8`
to a `&Cell<u8>` since the `Cell` will allow for mutation (even though the
code we have in process buffer will not mutable it). It is safer to
store the read only process buffer in a normal immutable `&[u8]` slice
than a `&[Cell<u8>]` since we cannot accidentally update the data in
which we have an immutable reference.

This fixes #2882 

### Testing Strategy

- [x] `cargo miri test` does not complain anymore

### Documentation Updated

- [x] None. This has the same external API

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/processbuffer.rs
+++ b/kernel/src/processbuffer.rs
@@ -20,6 +20,7 @@
 //! these slices, however.
 
 use core::cell::Cell;
+use core::mem::MaybeUninit;
 use core::ops::{Index, Range, RangeFrom, RangeTo};
 
 use crate::capabilities;
@@ -47,13 +48,12 @@ unsafe fn raw_processbuf_to_roprocessslice<'a>(
     ptr: *const u8,
     len: usize,
 ) -> &'a ReadableProcessSlice {
-    // Transmute a reference to a slice of Cell<u8>s into a reference
+    // Transmute a reference to a slice of [u8]s into a reference
     // to a ReadableProcessSlice. This is possible as
     // ReadableProcessSlice is a #[repr(transparent)] wrapper around a
     // [ReadableProcessByte], which is a #[repr(transparent)] wrapper
-    // around a [Cell<u8>], which is a #[repr(transparent)] wrapper
-    // around an [UnsafeCell<u8>], which finally #[repr(transparent)]
-    // wraps a [u8]
+    // around a [MaybeUninit<u8>], which is a #[repr(transparent)] wrapper
+    // around [u8]
     core::mem::transmute::<&[u8], &ReadableProcessSlice>(
         // Rust has very strict requirements on pointer validity[1]
         // which also in part apply to accesses of length 0. We allow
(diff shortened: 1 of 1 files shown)


jettr:
@lschuermann FYI since you assigned the issue to yourself.

I did try to get rid of `ReadableProcesByte` entirely by replacing it with `u8` and required touching a lot more files and the external API for process buffers. I think it is cleaner to switch it is `u8` directly but I also don't want to change churn between released. What are your thoughts?

lschuermann:
@jettr Thanks! I replied on the original issue, it seems better to have all discussion regarding this particular problem in one place. :)

jettr:
> @jettr Thanks! I replied on the original issue, it seems better to have all discussion regarding this particular problem in one place. :)

Agreed, let's discuss strategy there and then we can come back here when we feel good with the direction

phil-levis:
Bump?

jettr:
Bumped conversation on #2882

jettr:
Closing based on [feedback][0] that this would be unsound.

[0]:https://github.com/rust-lang/unsafe-code-guidelines/issues/303#issuecomment-1016639879

PR #2885, opened 2021-11-01
Stored state

granaghan:
### Pull Request Overview

Adds the ability to to serialize StoredState. Along with PR #2822, this enables crash information to be exported and fetched on a later boot.

### Testing Strategy

Tested export, import, and print of RISC-V StoredState.

### TODO or Help Wanted

### Documentation Updated

N/A?

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/syscall.rs
+++ b/arch/cortex-m/src/syscall.rs
@@ -2,8 +2,10 @@
 //! system call interface.
 
 use core::fmt::Write;
-use core::mem;
+use core::mem::{self, size_of};
+use core::ops::Range;
 use core::ptr::{read_volatile, write_volatile};
+use kernel::errorcode::ErrorCode;
 
 /// This is used in the syscall handler. When set to 1 this means the
 /// svc_handler was called. Marked `pub` because it is used in the cortex-m*
@@ -45,6 +47,56 @@ pub struct CortexMStoredState {
     yield_pc: usize,
     psr: usize,
     psp: usize,
+}
+
+const MAJOR_VER: usize = 1;
+const MINOR_VER: usize = size_of::<CortexMStoredState>();
+const TAG: [u8; 4] = [b'c', b't', b'x', b'm'];
+const METADATA_LEN: usize = 3;
+
+const USIZE_SZ: usize = size_of::<usize>();
+fn usize_byte_range(index: usize) -> Range<usize> {
+    index * USIZE_SZ..(index + 1) * USIZE_SZ
+}
+
+fn usize_from_u8_slice(slice: &[u8], index: usize) -> Result<usize, ErrorCode> {
+    let range = usize_byte_range(index);
+    if range.end < slice.len() {
+        Ok(usize::from_le_bytes(
+            slice[range].try_into().or(Err(ErrorCode::FAIL))?,
+        ))
+    } else {
+        Err(ErrorCode::SIZE)
+    }
+}
+
+fn write_usize_to_u8_slice(val: usize, slice: &mut [u8], index: usize) {
+    let range = usize_byte_range(index);
+    slice[range].copy_from_slice(&val.to_le_bytes());
+}
+
+impl core::convert::TryFrom<&[u8]> for CortexMStoredState {
+    type Error = ErrorCode;
+    fn try_from(ss: &[u8]) -> Result<CortexMStoredState, Self::Error> {
+        if ss.len() >= size_of::<CortexMStoredState>() + METADATA_LEN * USIZE_SZ
+            && usize_from_u8_slice(ss, 0)? == MAJOR_VER
+            && usize_from_u8_slice(ss, 1)? == MINOR_VER
+            && usize_from_u8_slice(ss, 2)? == u32::from_le_bytes(TAG) as usize
+        {
+            let mut res = CortexMStoredState {
+                regs: [0; 8],
+                yield_pc: usize_from_u8_slice(ss, 3)?,
+                psr: usize_from_u8_slice(ss, 4)?,
+                psp: usize_from_u8_slice(ss, 5)?,
+            };
+            for (i, v) in (6..14).enumerate() {
+                res.regs[i] = usize_from_u8_slice(ss, v)?;
+            }
+            Ok(res)
+        } else {
+            Err(ErrorCode::FAIL)
+        }
+    }
 }
 
 /// Implementation of the `UserspaceKernelBoundary` for the Cortex-M non-floating point
@@ -346,5 +398,28 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
                 "!!ERROR - Cortex M Thumb only!"
             },
         ));
+    }
+
+    fn store_context(
+        &self,
+        state: &CortexMStoredState,
+        out: &mut [u8],
+    ) -> Result<usize, ErrorCode> {
+        // MINOR_VER is size_of CortexMStoredState.
+        if out.len() >= size_of::<CortexMStoredState>() + 3 * USIZE_SZ {
+            write_usize_to_u8_slice(MAJOR_VER, out, 0);
+            write_usize_to_u8_slice(MINOR_VER, out, 1);
+            write_usize_to_u8_slice(u32::from_le_bytes(TAG) as usize, out, 2);
+            write_usize_to_u8_slice(state.yield_pc, out, 3);
+            write_usize_to_u8_slice(state.psr, out, 4);
+            write_usize_to_u8_slice(state.psp, out, 5);
+            for (i, v) in state.regs.iter().enumerate() {
+                write_usize_to_u8_slice(*v, out, 6 + i);
+            }
+            // + 3 for yield_pc, psr, psp
+            Ok((state.regs.len() + 3 + METADATA_LEN) * USIZE_SZ)
+        } else {
+            Err(ErrorCode::SIZE)
+        }
     }
 }
--- a/arch/cortex-m/src/syscall.rs
+++ b/arch/cortex-m/src/syscall.rs
@@ -49,10 +49,20 @@ pub struct CortexMStoredState {
     psp: usize,
 }
 
-const MAJOR_VER: usize = 1;
-const MINOR_VER: usize = size_of::<CortexMStoredState>();
+/// Values for encoding the stored state buffer in a binary slice.
+const VERSION: usize = 1;
+const STORED_STATE_SIZE: usize = size_of::<CortexMStoredState>();
 const TAG: [u8; 4] = [b'c', b't', b'x', b'm'];
 const METADATA_LEN: usize = 3;
+
+const VERSION_IDX: usize = 0;
+const SIZE_IDX: usize = 1;
+const TAG_IDX: usize = 2;
+const YIELDPC_IDX: usize = 3;
+const PSR_IDX: usize = 4;
+const PSP_IDX: usize = 5;
+const REGS_IDX: usize = 6;
+const REGS_RANGE: Range<usize> = REGS_IDX..REGS_IDX + 8;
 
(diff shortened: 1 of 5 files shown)


hudson-ayers (changes requested):
These uses of unsafe are UB, see the following playground link: https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=e8c98a0afacd79074e07bbb8b41eecc9 (run Miri using the playground link).

In general, just checking size is insufficient for transmute to be safe to use -- you also have to verify alignment constraints are met, among other things

bradjc:
This seems like an interesting feature. It seems like it wouldn't be that hard to encode the stored state structs into a buffer and then re-create the stored state struct from the buffer without using unsafe/transmute. It would be more code, but would make the serialization much more explicit.

granaghan:
Updated to just alias a usize array and get rid of the unsafe. I only need the [u8] for flash accesses. I can handle that more locally.

lschuermann:
I think it's interesting to have the `StoredState` (de)serializable, but given this was brought on the core calls previously I'd like to voice a few concerns which we had back then:

- are we making any guarantees that the information stored in this struct won't ever change by offering (de)serialization capabilities?

  Based on the fact that this includes a MAJOR and MINOR version I suspect not. We should elaborate and at least document the version semantics then. Also, should this be bound to Tock's versioning or be independent? I guess the latter would make more sense.

- is everyone going to need the same information for debugging?

  This proposal sounds like a reasonable default serialization strategy with the advantage of being deserializable into a `StoredState` again, but that might not suit everyone's needs. Would it not make more sense to have an architecture-specific trait implemented by the `SysCall` struct, containing a privileged method (e.g. capability-based) which exposes proper Rust structs, such that downstream developers have the option to ship their own serialization strategy without modifying the UKB?

Furthermore, I find the choice of having a single fixed-size array for all platforms odd. The serialization method arguably should take a mutable slice instead of returning a fixed-size array and return an error if the slice length is insufficient. This way we can also avoid a 66-byte stack allocation in many cases.

One final thought: should we encode the target architecture in the serialized stored state as well? It's probably unlikely for users to swap serialized `StoredState`s between platforms, but it could break pretty badly if they do and we have no way of detecting that. :)

granaghan:
> * is everyone going to need the same information for debugging?
>   This proposal sounds like a reasonable default serialization strategy with the advantage of being deserializable into a `StoredState` again, but that might not suit everyone's needs. Would it not make more sense to have an architecture-specific trait implemented by the `SysCall` struct, containing a privileged method (e.g. capability-based) which exposes proper Rust structs, such that downstream developers have the option to ship their own serialization strategy without modifying the UKB?

The down side to this strategy is that because stored state is only accessible through a Process implementation, exposing an associated type from UKB would make the Process trait become a generic on the Chip type. That would cause painful churn everywhere that Process is used.

> Furthermore, I find the choice of having a single fixed-size array for all platforms odd. The serialization method arguably should take a mutable slice instead of returning a fixed-size array and return an error if the slice length is insufficient. This way we can also avoid a 66-byte stack allocation in many cases.

> One final thought: should we encode the target architecture in the serialized stored state as well? It's probably unlikely for users to swap serialized StoredStates between platforms, but it could break pretty badly if they do and we have no way of detecting that. :)

Given the above, I think doing these two things is likely the better route.

hudson-ayers:
This change increases code size by ~200 bytes on boards that do not use it -- it is the same problem as we have encountered elsewhere, where you can't add methods to the process trait for free, because Rust cannot do dead code elimination there.

I think that cost is small enough that we should not worry about it.

kernel/src/syscall.rs
```diff
@@ -648,4 +648,12 @@ pub trait UserspaceKernelBoundary {
         state: &Self::StoredState,
         writer: &mut dyn Write,
     );
+
+    /// Store architecture specific (e.g. CPU registers or status flags) data
+    /// for a process. On success returns the number of elements written to out.
+    fn store_context(
+        &self,
+        state: &Self::StoredState,
+        out: &mut [usize],
```
lschuermann, line 657: Can we change this to use a slice of `u8`s instead? It's unconventional to pass around slices of `usize` in the kernel and thus might make interop with other interfaces more tricky. Furthermore, using `usize` can be tricky w.r.t. endianness. While platforms are free to choose their encoding of the `StoredState` structs (after all, they won't ever be compatible to each other), it's at least important that the header fields can be read for other platforms as well and thus use a fixed endianness. I don't expect any (significant) overhead caused by using `usize::{from,to}_le_bytes` etc.

arch/rv32i/src/syscall.rs
```diff
@@ -40,6 +42,35 @@ const R_A2: usize = 11;
 const R_A3: usize = 12;
 const R_A4: usize = 13;
 
+const MAJOR_VER: usize = 1;
+const MINOR_VER: usize = size_of::<Riscv32iStoredState>();
+const TAG: [u8; 4] = [b'r', b'v', b'3', b'2'];
```
lschuermann, line 47: I like the idea of using meaningful semantic information in the tag, though we're going to want to avoid some obvious collisions if we can :smile:. Given we have 4 bytes here, and the state to store heavily depends on the extensions present on RISC-V (e.g. floating point registers), maybe it'd be a good idea to use something like `r5i\0`, where `r` -> RISC-V, `5` -> 2^5 bits, `i\0` for the base integer instruction set.

arch/cortex-m/src/syscall.rs
```diff
@@ -347,4 +377,28 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
             },
         ));
     }
+
+    fn store_context(
+        &self,
+        state: &CortexMStoredState,
+        out: &mut [usize],
+    ) -> Result<usize, ErrorCode> {
+        // MINOR_VER is size_of Riscv32iStoredState + size_of usize. +2 more
```
alexandruradovici, line 386:
```suggestion
        // MINOR_VER is size_of CortexMStoredState + size_of usize. +2 more
```

granaghan:
Anything else I need to do here?

alistair23 (dismissed):

hudson-ayers (approved):
I am happy with this, curious if @bradjc has any more feedback

arch/rv32i/src/syscall.rs
```diff
@@ -40,6 +44,56 @@ const R_A2: usize = 11;
...
+const U32_SZ: usize = size_of::<u32>();
+fn u32_byte_range(index: usize) -> Range<usize> {
+    index * U32_SZ..(index + 1) * U32_SZ
+}
+
+fn u32_from_u8_slice(slice: &[u8], index: usize) -> Result<u32, ErrorCode> {
+    let range = u32_byte_range(index);
+    if range.end < slice.len() {
+        Ok(u32::from_le_bytes(
+            slice[range].try_into().or(Err(ErrorCode::FAIL))?,
```
bradjc, line 61: All of the index operations make me wish Rust didn't support that syntax, because we have to rely on the compiler to figure out that there are enough checks to ensure a panic can't happen and to elide the panic handling code.

I personally would rather see `.get()` used rather than `[]` with the `if()` statement (as it is here: https://github.com/tock/tock/blob/master/libraries/tock-tbf/src/types.rs#L426), but this is a little bit of my pet peeve and not something we are rigorous about in the kernel. I think the code gets a little simpler, but I won't hold up this PR on this issue.

kernel/src/process.rs
```diff
@@ -610,6 +610,10 @@ pub trait Process {
 
     /// Return the lowest recorded address of the process stack, if known.
     fn debug_stack_end(&self) -> Option<*const u8>;
+
+    /// Export stored state as a binary blob. Returns the number of items
+    /// written on success.
+    fn get_stored_state(&self, out: &mut [u8]) -> Result<usize, ErrorCode>;
```
bradjc, line 616:
```suggestion
    /// Write stored state as a binary blob into the `out` slice. Returns the number of bytes
    /// written to `out` on success.
    ///
    /// Returns `ErrorCode::SIZE` if `out` is too short to hold the stored state binary
    /// representation. Returns `ErrorCode::FAIL` on an internal error.
    fn get_stored_state(&self, out: &mut [u8]) -> Result<usize, ErrorCode>;
```

bradjc: Also, can you move this to line 579?

kernel/src/syscall.rs
```diff
@@ -648,4 +648,8 @@ pub trait UserspaceKernelBoundary {
         state: &Self::StoredState,
         writer: &mut dyn Write,
     );
+
+    /// Store architecture specific (e.g. CPU registers or status flags) data
+    /// for a process. On success returns the number of elements written to out.
+    fn store_context(&self, state: &Self::StoredState, out: &mut [u8]) -> Result<usize, ErrorCode>;
```
bradjc, line 654:
```suggestion
    /// Store architecture specific (e.g. CPU registers or status flags) data
    /// for a process to the `out` slice. On success returns the number of bytes written to `out`.
    /// Returns `ErrorCode::SIZE` if out is too short to hold the context data.
    fn store_context(&self, state: &Self::StoredState, out: &mut [u8]) -> Result<usize, ErrorCode>;
```

arch/cortex-m/src/syscall.rs
```diff
@@ -47,6 +49,56 @@ pub struct CortexMStoredState {
     psp: usize,
 }
 
+const MAJOR_VER: usize = 1;
+const MINOR_VER: usize = size_of::<CortexMStoredState>();
```
bradjc, line 53:
```suggestion
const VERSION: usize = 1;
const STORED_STATE_SIZE: usize = size_of::<CortexMStoredState>();
```

arch/cortex-m/src/syscall.rs
```diff
@@ -47,6 +49,56 @@ pub struct CortexMStoredState {
     psp: usize,
 }
 
+const MAJOR_VER: usize = 1;
```
bradjc, line 52:
```suggestion
/// Values for encoding the stored state buffer in a binary slice.
const MAJOR_VER: usize = 1;
```

bradjc (commented):
In general looks good, just a few changes since this touches some very core interfaces.

arch/cortex-m/src/syscall.rs
```diff
@@ -47,6 +49,56 @@ pub struct CortexMStoredState {
...
+
+fn write_usize_to_u8_slice(val: usize, slice: &mut [u8], index: usize) {
+    let range = usize_byte_range(index);
+    slice[range].copy_from_slice(&val.to_le_bytes());
+}
+
+impl core::convert::TryFrom<&[u8]> for CortexMStoredState {
+    type Error = ErrorCode;
+    fn try_from(ss: &[u8]) -> Result<CortexMStoredState, Self::Error> {
+        if ss.len() >= size_of::<CortexMStoredState>() + METADATA_LEN * USIZE_SZ
```
hudson-ayers, line 81: One more question: why support slices larger than the size we are converting? Wouldn't passing a slice larger than the size of `CortexMStoredState` likely indicate some bug?

bradjc: Is it possible the slice would come from somewhere that doesn't know which architecture will be used? Or that it would want to keep access to a longer slice?

granaghan: It's read into a large buffer from flash and I was just passing the rest of the buffer from the start of the stored state. Updated to expect actual size.

hudson-ayers (approved):

bradjc (approved):

lschuermann (approved):

granaghan:
Rebased.

hudson-ayers (approved):
bors r+

PR #2886, opened 2021-11-01
OpenTitan: Add support for an RSA test using OTBN

alistair23:
### Pull Request Overview

This PR adds a OTBN test case for OpenTitan that performs an RSA encryption operation. It does this by first loading the OTBN binary from the list of Tock apps (the OTBN binary is disabled), then setting the required options and performing the operation.

### Testing Strategy

Running OTBN tests on the CW310

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


hudson-ayers (approved):

bradjc (approved):

alistair23:
Ready to go

PR #2889, opened 2021-11-04
esp32: Update timer

alexandruradovici:
### Pull Request Overview

This pull request fixes the `TimerGroup` peripheral driver for the ESP32-C3 Dev KitM1. It seems that the hardware present on this board is slightly different from the one in the *classic* esp32 (I think that this is why there is an `esp32` crate, to be used with ports of the esp32 boards).

The PR does the following changes:
 * adds a new `timg` driver in the `esp32-c3` crate
 * adds a minimal `sysreg` driver to the `esp32-c3` crate

### Testing Strategy

This pull request was tested by using an ESP32-C3 Dev KitM1

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers (dismissed):
I realize I reviewed these files in your other PR, everything here looks good now.

chips/esp32-c3/src/timg.rs
```diff
@@ -0,0 +1,245 @@
...
+        (0x064 => wdtwprotect: ReadWrite<u32>),
+
+        (0x068 => rtccalicfg: ReadWrite<u32, RTCCALICFG::Register>),
+        (0x06C => rtccalicfg1: ReadWrite<u32, RTCCALICFG1::Register>),
+
+        (0x070 => int_ena: ReadWrite<u32, INT::Register>),
+        (0x074 => int_raw: ReadWrite<u32, INT::Register>),
+        (0x078 => int_st: ReadWrite<u32, INT::Register>),
+        (0x07C => int_clr: ReadWrite<u32, INT::Register>),
+        (0x080 => @END),
```
alistair23, line 56: It seems that the address of the interrupts is the only difference. Maybe instead it's better to just make the current implementation dynamic instead of adding all this new code

alistair23 (changes requested):
Thanks for the PR.

Can you describe what doesn't work currently? Because the EPS32 runs the alarm tests as part of it's unit tests and they pass for me. It might be worth looking at adding a test case for the failure you are seeing

chips/esp32-c3/src/chip.rs
```diff
@@ -28,18 +30,22 @@ pub struct Esp32C3<'a, I: InterruptService<()> + 'a> {
...
+    pub sysreg: sysreg::SysReg,
 }
 
 impl<'a> Esp32C3DefaultPeripherals<'a> {
     pub fn new() -> Self {
         Self {
             uart0: esp32::uart::Uart::new(esp32::uart::UART0_BASE),
-            timg0: esp32::timg::TimG::new(esp32::timg::TIMG0_BASE),
+            timg0: timg::TimG::new(timg::TIMG0_BASE, timg::ClockSource::Pll),
+            timg1: timg::TimG::new(timg::TIMG0_BASE, timg::ClockSource::Pll),
```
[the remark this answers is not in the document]
alexandruradovici: No, this is a typo, thank you for pointing it out.

chips/esp32-c3/src/timg.rs
alexandruradovici: I'm not sure what you mean here.

alexandruradovici:
I started by adding the process console and noticed that the alarm was not firing. I then tried a simple app that performs a `delay_ms` and it does not work. When I started looking at the timer, it did not seem to increment any ticks.

- The frequency seed to be wrong, it is 20 MHz
- T1 timer does not seem to exist, the documentation only says T0
- The `DIVIDER_RST` was missing
- The timer had to be disabled before it can be set
- The timer would not work without setting `CONFIG::EN`
- The timer interrupt was not mapped
- The timer interrupt was not handled in `chip.rs`

chips/esp32-c3/src/timg.rs
alistair23: From what I can tell the only difference between the current implementation in `chips/esp32` and this one is that the `int_*` register addresses are different.

So instead of copying the entire file to just change some addresses, why not have a option to the `new()` function that specifies which address to use?

Something like:

```rust
fn new(... c3: bool) {
    ...
    self.c3.set(c3);
    ...
}
fn my_function() {
    if self.c3.get() {
        self.registers.int_en_c3.write(...);
    } else {
        self.registers.int_en.write(...);
    }
}
```

Does that make sense?

alistair23:
Strange. I see the alarm tests working.

Can you copy the log from the unit test results?

alexandruradovici:
> Strange. I see the alarm tests working.
> 
> Can you copy the log from the unit test results?

How do I run these tests?

chips/esp32-c3/src/timg.rs
alexandruradovici: Some registers from the ESP32 seem to overlap with the ones in C3. I have to check that again.

alistair23:
The `TestRandomAlarm` test should be incrementing the counter when `AlarmClient` is called. So the test should test the interrupts. But you are right it looks like the ESP32 currently isn't handling the timer interrupts.

If you can consolidate this and the standard EPS32 timer implementation we should be good to go

alexandruradovici:
> If you can consolidate this and the standard EPS32 timer implementation we should be good to go

I'm not sure I understand what do you mean by this? Do you mean a partial reuse of esp32 timer registers?

alistair23:
Exactly!

alexandruradovici:
I tried to consolidate the two implementations , but I am not sure how I can do this. Except for the `RTCCALICFG` register, all the the other registers are either simple `u32` types or have overlapping fields.

I did import the `RTCCALICFG` from the `esp32`, but unless I name some fields with something like `DIVIDER_OR_EDGE_INT`, I have no idea how to consolidate this. @alistair23 any ideas are welcome.

alistair23:
Yeah, it's not as nice as I was hoping. I didn't realise the register field values changed just enough to be difficult.

I was picturing something like this:

```diff
diff --git a/chips/esp32/src/timg.rs b/chips/esp32/src/timg.rs
index c4970ef7b..bb1a658c1 100644
--- a/chips/esp32/src/timg.rs
+++ b/chips/esp32/src/timg.rs
@@ -1,5 +1,6 @@
 //! TimG Group driver.
 
+use core::cell::Cell;
 use kernel::hil::time::{self, Alarm, Counter, Ticks, Ticks64, Time};
 use kernel::utilities::cells::OptionalCell;
 use kernel::utilities::registers::interfaces::{ReadWriteable, Readable, Writeable};
@@ -57,7 +58,12 @@ register_structs! {
         (0x068 => rtccalicfg: ReadWrite<u32, RTCCALICFG::Register>),
         (0x06C => rtccalicfg1: ReadWrite<u32, RTCCALICFG1::Register>),
 
-        (0x070 => _reserved2),
+        (0x070 => int_ena_c3: ReadWrite<u32, INT_C3::Register>),
+        (0x074 => int_raw_c3: ReadWrite<u32, INT_C3::Register>),
+        (0x078 => int_st_c3: ReadWrite<u32, INT_C3::Register>),
+        (0x07C => int_clr_c3: ReadWrite<u32, INT_C3::Register>),
+
+        (0x080 => _reserved2),
         (0x098 => int_ena: ReadWrite<u32, INT::Register>),
         (0x09C => int_raw: ReadWrite<u32, INT::Register>),
         (0x0A0 => int_st: ReadWrite<u32, INT::Register>),
@@ -68,20 +74,23 @@ register_structs! {
 
 register_bitfields![u32,
     CONFIG [
+        USE_XTAL OFFSET(9) NUMBITS(1) [],
         ALARM_EN OFFSET(10) NUMBITS(1) [],
         LEVEL_INT_EN OFFSET(11) NUMBITS(1) [],
-        EDGE_INT_EN OFFSET(12) NUMBITS(1) [],
+        DIVIDER_RST OFFSET(12) NUMBITS(1) [],
         DIVIDER OFFSET(13) NUMBITS(16) [],
         AUTORELOAD OFFSET(29) NUMBITS(1) [],
-        INCREATE OFFSET(30) NUMBITS(1) [],
+        INCREASE OFFSET(30) NUMBITS(1) [],
         EN OFFSET(31) NUMBITS(1) [],
     ],
     WDTCONFIG0 [
+        APP_CPU_RESET_EN OFFSET(12) NUMBITS(1) [],
+        PROC_CPU_RESET_EN OFFSET(13) NUMBITS(1) [],
         FLASHBOOT_MOD_EN OFFSET(14) NUMBITS(1) [],
         SYS_RESET_LENGTH OFFSET(15) NUMBITS(3) [],
         CPU_RESET_LENGTH OFFSET(18) NUMBITS(3) [],
-        LEVEL_INT_EN OFFSET(21) NUMBITS(1) [],
-        EDGE_INT_EN OFFSET(22) NUMBITS(1) [],
+        LEVEL_INT_EN_OR_USE_XTAL OFFSET(21) NUMBITS(1) [],
+        UPDATE_EN OFFSET(22) NUMBITS(1) [],
         STG3 OFFSET(23) NUMBITS(2) [],
         STG2 OFFSET(25) NUMBITS(2) [],
         STG1 OFFSET(27) NUMBITS(2) [],
@@ -89,6 +98,7 @@ register_bitfields![u32,
         EN OFFSET(31) NUMBITS(1) [],
     ],
     WDTCONFIG1 [
+        DIVCNT_RST OFFSET(0) NUMBITS(1) [],
         CLK_PRESCALE OFFSET(16) NUMBITS(16) [],
     ],
     RTCCALICFG [
@@ -99,6 +109,7 @@ register_bitfields![u32,
         START OFFSET(31) NUMBITS(1) [],
     ],
     RTCCALICFG1 [
+        CYCLING_DATA_VLD OFFSET(0) NUMBITS(1) [],
         VALUE OFFSET(7) NUMBITS(25) [],
     ],
     INT [
@@ -106,18 +117,32 @@ register_bitfields![u32,
         T1 OFFSET(1) NUMBITS(1) [],
         WDT OFFSET(2) NUMBITS(1) [],
     ],
+    INT_C3 [
+        T0 OFFSET(0) NUMBITS(1) [],
+        WDT OFFSET(1) NUMBITS(1) [],
+    ],
 ];
 
+#[derive(Copy, Clone)]
+pub enum ClockSource {
+    Pll = 0,
+    Xtal = 1,
+}
+
 pub struct TimG<'a> {
     registers: StaticRef<TimgRegisters>,
+    clocksource: ClockSource,
     alarm_client: OptionalCell<&'a dyn time::AlarmClient>,
+    c3: Cell<bool>,
 }
 
 impl TimG<'_> {
-    pub const fn new(base: StaticRef<TimgRegisters>) -> Self {
+    pub const fn new(base: StaticRef<TimgRegisters>, c3: bool, clocksource: ClockSource) -> Self {
         TimG {
             registers: base,
+            clocksource,
             alarm_client: OptionalCell::empty(),
+            c3: Cell::new(c3),
         }
     }
 
@@ -194,6 +219,20 @@ impl<'a> Alarm<'a> for TimG<'a> {
             expire = now;
         }
 
+        if self.c3.get() {
+            self.registers
+                .t0config
+                .modify(CONFIG::ALARM_EN::CLEAR + CONFIG::EN::CLEAR);
+
+            self.registers.t0config.modify(
+                CONFIG::USE_XTAL.val(self.clocksource as u32)
+                    + CONFIG::INCREASE::SET
+                    + CONFIG::DIVIDER.val(2 * (2 - self.clocksource as u32)),
+            );
+
+            self.registers.t0config.modify(CONFIG::DIVIDER_RST::SET);
+        }
+
         let val = expire.into_u64();
         let high = (val >> 32) as u32;
         let low = (val & 0xffffffff) as u32;
@@ -202,7 +241,11 @@ impl<'a> Alarm<'a> for TimG<'a> {
         self.registers.t0alarmhi.set(high);
         self.registers.t0alarmlo.set(low);
 
-        self.registers.int_ena.modify(INT::T0::SET);
+        if self.c3.get() {
+            self.registers.int_ena_c3.modify(INT_C3::T0::SET);
+        } else {
+            self.registers.int_ena.modify(INT::T0::SET);
+        }
 
         self.registers.t0config.modify(CONFIG::ALARM_EN::SET);
     }
```

I think the conflicting register names could be either or a `A_OR_B` setup. Up to you.

I know it's a bit clunky, but I still think it's better then duplicating all of the code.

alexandruradovici:
I don't think this is possible due to the associated `Frequency` and `Ticks` traits. The `Time` trait has to have a different implementation based on the chip type. I did that, but I get the following errors:

```
error[E0277]: the trait bound `TimG<'a, C3>: Time` is not satisfied
   --> chips/esp32/src/timg.rs:203:26
    |
203 | impl<'a, const C3: bool> Counter<'a> for TimG<'a, C3> {
    |                          ^^^^^^^^^^^ the trait `Time` is not implemented for `TimG<'a, C3>`
    |
    = help: the following implementations were found:
              <TimG<'_, false> as Time>
              <TimG<'_, true> as Time>
note: required by a bound in `Counter`

error[E0277]: the trait bound `TimG<'a, C3>: Time` is not satisfied
   --> chips/esp32/src/timg.rs:229:26
    |
229 | impl<'a, const C3: bool> Alarm<'a> for TimG<'a, C3> {
    |                          ^^^^^^^^^ the trait `Time` is not implemented for `TimG<'a, C3>`
    |
    = help: the following implementations were found:
              <TimG<'_, false> as Time>
              <TimG<'_, true> as Time>
note: required by a bound in `Alarm`

error[E0277]: the trait bound `TimG<'a, C3>: Time` is not satisfied
   --> chips/esp32/src/timg.rs:234:8
    |
234 |     fn set_alarm(&self, reference: Self::Ticks, dt: Self::Ticks) {
    |        ^^^^^^^^^ the trait `Time` is not implemented for `TimG<'a, C3>`
    |
    = help: the following implementations were found:
              <TimG<'_, false> as Time>
              <TimG<'_, true> as Time>

error[E0277]: the trait bound `TimG<'a, C3>: Time` is not satisfied
   --> chips/esp32/src/timg.rs:268:8
    |
268 |     fn get_alarm(&self) -> Self::Ticks {
    |        ^^^^^^^^^ the trait `Time` is not implemented for `TimG<'a, C3>`
    |
    = help: the following implementations were found:
              <TimG<'_, false> as Time>
              <TimG<'_, true> as Time>

error[E0277]: the trait bound `TimG<'a, C3>: Time` is not satisfied
   --> chips/esp32/src/timg.rs:268:18
    |
268 |     fn get_alarm(&self) -> Self::Ticks {
    |                  ^^^^^ the trait `Time` is not implemented for `TimG<'a, C3>`
    |
    = help: the following implementations were found:
              <TimG<'_, false> as Time>
              <TimG<'_, true> as Time>

error[E0277]: the trait bound `TimG<'a, C3>: Time` is not satisfied
   --> chips/esp32/src/timg.rs:284:8
    |
284 |     fn minimum_dt(&self) -> Self::Ticks {
    |        ^^^^^^^^^^ the trait `Time` is not implemented for `TimG<'a, C3>`
    |
    = help: the following implementations were found:
              <TimG<'_, false> as Time>
              <TimG<'_, true> as Time>

error[E0277]: the trait bound `TimG<'a, C3>: Time` is not satisfied
   --> chips/esp32/src/timg.rs:284:19
    |
284 |     fn minimum_dt(&self) -> Self::Ticks {
    |                   ^^^^^ the trait `Time` is not implemented for `TimG<'a, C3>`
    |
    = help: the following implementations were found:
              <TimG<'_, false> as Time>
              <TimG<'_, true> as Time>
```

I assume this is a compiler bug, as there are no other possible values of `C3` except `true` and `false`.

@alistair23 any suggestions would help.

alexandruradovici:
I managed somehow to make it work by adding the `Frequency` as a generic parameter.

I think this is ready to be merge, I'm not sure why the litex test fails, @lschuermann I think I need your help here.

lschuermann:
> I'm not sure why the litex test fails, @lschuermann I think I need your help here.

It appears as if compiling `elf2tab` is failing, presumably because something got updated somewhere? That's why it's a good idea to avoid impurity in the CI through automatic installs :). I'll try to fix, but in the meantime the `litex-sim-ci` is not a required check, so merging should still work just fine. Sorry for the red X next to the commit.

hudson-ayers (approved):

alistair23 (approved):

alistair23:
Looks good!

PR #2892, opened 2021-11-08
kernel: process_standard: Don't check memory regions if disabled

alistair23:
### Pull Request Overview

Check if an app is disabled before checking if the fixed memory
regions match. If an app is disabled there is no reason to report an
error for mismatched memory regions.

### Testing Strategy

Loading apps

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


hudson-ayers (approved):
Makes sense to me

bradjc (approved):

PR #2893, opened 2021-11-08
esp32: UART receive

alexandruradovici:
### Pull Request Overview

This pull request adds the ability to receive bytes from UART for the ESP32-C3 DevKitM1 board.

Additionally, it adds the process console to the board implementation. This requires a working timer (#2889).

### Testing Strategy

This pull request was tested by using an ESP32-C3 DevM1 Kit.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


chips/esp32/src/uart.rs
```diff
@@ -310,6 +312,35 @@ impl<'a> Uart<'a> {
...
+
+        if idx < len {
+            // If we are going to receive anything, we first need to enable the
+            // RX interrupt. This ensures that we will get an interrupt, where
+            // we can either call the callback from, or continue receiving
+            // bytes.
+            self.enable_rx_interrupt();
+
+            // Read from the UART hardware and write them to the receive buffer
+            // until either the buffer if full or the UART hardware is empty.
```
hudson-ayers, line 328:
```suggestion
            // until either the buffer is full or the UART hardware is empty.
```

chips/esp32/src/uart.rs
```diff
@@ -330,6 +361,27 @@ impl<'a> Uart<'a> {
                 self.tx_progress();
             }
         }
+
+        if intrs.is_set(INT::RXFIFO_FULL_INT) || intrs.is_set(INT::RXFIFO_TOUT_INT) {
+            self.disable_rx_interrupt();
```
hudson-ayers, line 366: I think this line should  be moved inside the if statement on the next line. There is no reason to disable interrupts if we have not received everything, and disabling here then re-enabling in `rx_progress()` wastes cycles and risks missing interrupts that arrive in that brief window of time.

chips/esp32-c3/src/sysreg.rs
```diff
@@ -0,0 +1,106 @@
...
+        SOC_CLK_SEL OFFSET(10) NUMBITS(2) [
+            Xtal = 0,
+            Pll = 1,
+            Fosc = 2
+        ],
+        CLK_XTAL_FREQ OFFSET(12) NUMBITS(6) [],
+    ]
+];
+
+#[repr(u8)]
```
hudson-ayers, line 50: why `#[repr(u8)]`? It looks like this is just cast to a u32 in the one place it is used anyway, and forcing a u8 representation can actually lead to extra code for casting to/from other widths, and can prevent certain niche optimizations.

chips/esp32-c3/src/sysreg.rs
```diff
@@ -0,0 +1,106 @@
...
+pub enum PllFrequency {
+    MHz320 = 0,
+    MHz480 = 1,
+}
+
+#[repr(u8)]
```
hudson-ayers, line 56: Same here -- am I missing something?

chips/esp32-c3/src/timg.rs
```diff
@@ -0,0 +1,244 @@
...
+use kernel::utilities::StaticRef;
+use kernel::ErrorCode;
+
+pub const TIMG0_BASE: StaticRef<TimgRegisters> =
+    unsafe { StaticRef::new(0x6001_F000 as *const TimgRegisters) };
+
+pub const TIMG1_BASE: StaticRef<TimgRegisters> =
+    unsafe { StaticRef::new(0x6002_0000 as *const TimgRegisters) };
+
+/// 80MHz `Frequency`
```
hudson-ayers, line 17:
```suggestion
/// 20MHz `Frequency`
```

chips/esp32-c3/src/timg.rs
```diff
@@ -0,0 +1,244 @@
...
+    RTCCALICFG1 [
+        VALUE OFFSET(7) NUMBITS(25) [],
+    ],
+    INT [
+        T0 OFFSET(0) NUMBITS(1) [],
+        WDT OFFSET(1) NUMBITS(1) [],
+    ],
+];
+
+#[repr(u8)]
```
hudson-ayers, line 102: Same here

chips/esp32-c3/src/timg.rs
```diff
@@ -0,0 +1,244 @@
...
+        }
+    }
+}
+
+impl time::Time for TimG<'_> {
+    type Frequency = Freq20MHz;
+    type Ticks = Ticks64;
+
+    fn now(&self) -> Self::Ticks {
+        self.registers.t0update.set(0xABC);
```
hudson-ayers, line 153: what is `0xABC` here? A comment or named constant would be helpful

alexandruradovici: My understanding is that you have to write anything in that register, the value does not matter. I did not write this part, I think @alistair23 can answer this better. I'll add a comment.

chips/esp32-c3/src/sysreg.rs
alexandruradovici: You are right, I modified it to u32.

chips/esp32-c3/src/timg.rs
alexandruradovici: You are right, I modified it to u32.

chips/esp32-c3/src/sysreg.rs
alexandruradovici: You are right, I modified it to u32.

chips/esp32/src/uart.rs
alexandruradovici: I just replicated what @alistair23 did for the transmission part, but you have a point. I moved both disabling of interrupts just before running the client's callback function.

chips/esp32-c3/src/timg.rs
hudson-ayers: My inclination for all of these cases would be to simply not specify the representation at all -- these types are not used when reading or writing directly to registers or across any other FFI boundary, so why do we need to specify how they are represented in memory?

alexandruradovici: I has the impression that this was needed to automatically cast to u32, but it is not. I deleted the repr.

hudson-ayers (approved):

chips/esp32/src/uart.rs
```diff
@@ -310,14 +306,36 @@ impl<'a> Uart<'a> {
...
+            });
+        }
+    }
+
     pub fn handle_interrupt(&self) {
         let regs = self.registers;
         let intrs = regs.int_st.extract();
 
         if intrs.is_set(INT::TXFIFO_EMPTY_INT) {
-            self.disable_tx_interrupt();
```
alistair23, line 318: You aren't clearing the interrupt with this change

alistair23 (changes requested):
Overall the UART part looks good, just need to make sure you clear the interrupts in the interrupt handler

chips/esp32/src/uart.rs
alexandruradovici: I did that initially, but @hudson-ayers suggested disabling them only when the transmit or receive is done (lines 338 and 354). I might be missing something though.

hudson-ayers: @alistair23 why would you disable the interrupt if you are still transmitting / still receiving?

alistair23: You don't have to disable the interrupt, but you need to clear the pending interrupt status. Currently this is done in the `disable_*()` functions

hudson-ayers: Oh, I see. I suppose this should be changed back, though renaming the function to `disable_and_clear_interrupts()` might make this clearer.

alexandruradovici: I have split the function in two.

hudson-ayers (approved):

hudson-ayers (dismissed):

alistair23 (changes requested):
This depends on https://github.com/tock/tock/pull/2889

chips/esp32-c3/src/timg.rs
```diff
@@ -0,0 +1,245 @@
+//! TimG Group driver.
+
+use kernel::hil::time::{self, Alarm, Counter, Ticks, Ticks64, Time};
+use kernel::utilities::cells::OptionalCell;
+use kernel::utilities::registers::interfaces::{ReadWriteable, Readable, Writeable};
+use kernel::utilities::registers::register_bitfields;
+use kernel::utilities::registers::{register_structs, ReadWrite};
+use kernel::utilities::StaticRef;
+use kernel::ErrorCode;
```
alistair23, line 9: This file should be removed

alistair23 (changes requested):

alistair23 (approved):

PR #2895, opened 2021-11-09
Restructure the USB stack to be more modular

alevy:
_NOTE: this PR includes relatively sweeping changes mostly in new files, though much of the code is borrowed from the modules it aims to replace. As a result, I recommend primarily reading the source tree itself rather than the diff. The relevant design is mostly encapsulated in `capsules/src/usb/{device,configuration,bulk_simple,bulk_echo}.rs` and any changes to other parts of the code are important for review, but mostly ancillary to the proposed design._

### Pull Request Overview

This pull request restructures the USB stack with a different division of responsibilities to allows for more modularity.

#### What's the problem?

Before this PR, the state of the USB stack is functional (thank you very much to those who contributed it, as getting it as functional as it is is no small feat!) but quite ad-hoc. Currently, the stack is structured something like this:

```
                  Board
                   ^
                   |
                   v
              Functional
                driver
                 |   ^
            |-----   |
            v        |
     ClientCtrl      |
         |           |
         v           |
         UsbController
```

A particular board initializes the chip's implementation of `UsbController` as well as a function-specific driver (e.g., CDC-ACM or CTAP, etc) and connects them. Because `UsbController` (correctly) is not virtualized,the functional driver "owns" the controller: it uses the global namespace for configuration, interface and endpoint numbers, it controls the device configuration itself (e.g. vendor/product id, etc).  The functionality to actually support USB enumeration is factored out into ClientCtr, but this is used as a library and because of the high-level structure, there is really no way to modularly construct a device composed of multiple functions (e.g. a device that both provides a serial interface and CTAP HID without writing a new driver that implements both).

A few ad-hoc additions to the stack over time have led to some other minor warts: HID report descriptors are handled by `ClientCtr` meaning even when there is no HID device the USB stacks suffers from the complexity and memory overhead of including that logic; via various copy-paste jobs boards generally claim to be self-powered via their configuration descriptor even though some are very much not (e.g. the nano33ble, which is powered exclusively over USB), and declare an impractically low max current draw.

#### The solution in this PR

This PR proposed a new design of the USB stack that follows a more typical multiplexing stack in Tock and solves most of the existing issues.
```
         Board
        /     \
    Console  Mouse
       ^       ^
      /         \
     v           v
Interface     Interface
       ^       ^
        \      /
         \    /
          \  /
           v
     Configuration   
           ^
           |
           v
        Device
           ^
           |
           v
      UsbController
```

This PR introduces four new components:

  - The `Device` struct, which replaces the `ClientCtl` library but is now the main `usb::Client` that interacts directly with the low-level USB controller. It also now _holds_ information such as the vendor/product ID, strings table, etc and is configured directly from the board (rather than indirectly via function-specific drivers).
  - The `Configuration` trait, which implements "contains" one or more interfaces and is responsible for encoding the configuration descriptor. It's expected that most configurations will simply use the provided `SimpleConfiguration` struct which more or less wraps a slice of `Interfaces`, but it's left as a trait for flexibility (tbd if this is actually necessary). In principle USB supports multiple configurations that the host can switch between (only one can be active at a time), and this design could support that functionality, but in practice it's not clear this is useful enough to implement (for example, I'm not sure if Linux ever chooses anything but the first configuration).
  - The `Interface` trait implements a USB interface (a collection of related endpoints) and is where developers are expected to implement function-specific drivers. The interface trait includes upcalls similar to `usb::Client` with slightly different signatures and semantics. Most notably, endpoint numbers _never_ refer to the absolute endpoint numbers but rather relative indeces in the interface (the `Device` performs this translation).
  - The `Endpoint` struct, which contains an optional buffer (which the `Device` will configure as a buffer for the endpoint during enumeration) and endpoint number (which is determined and written to by the `Device` during enumeration) as well as other meta-data (direction, type, etc).

Enumeration is handled in the Device struct, but deferred to the supplied Configuration for enumerating interfaces, endpoints and any ancillary descriptors (such as Class descriptors or HID reports descriptors). It also handles allocating configuration, interface, and endpoint numbers and routing post-enumeration requests/responses to the appropriate `Interface`.

Most of the existing stack is left in-tact in this PR. The new stack is implemented alongside it along with two actual Interface implementations: BulkSimple and BulkEcho. Both implement Linux's very simple serial interface for prototypes. `BulkSimple` also implements `UartData` and can be used as a UART device (e.g. in the console) while `BulkEcho` implements the echo server implemented for the previous stack in `usbc_client.rs`. Note that both implementations are fairly basic. For example, `BulkSimple` does not drop TX transmissions if USB hasn't connected so, in practice, the caller will block until the host starts reading transmissions over USB. In other words, these implementations are not for production use (unless that's the functionality you want for some reason).

#### Known Warts

The current implementation of this design has some outstanding warts that likely ought to be resolved either before or shortly after merging.

First, `Interface` implementations typically need to signal to the host when an endpoint is either ready to send or receive more data after it returns a `Delay`. Since this might happen out-of-bound of a USB transaction (e.g. when a process gives it a new buffer to send), it needs a reference to something that allows it to do so. For now, BulkSimple accomplishes this by simply requiring an instance of `UsbController` (the same instance as is passed to `Device`) and uses `Endpoint#endpoint_number` as a dynamic argument to `UsbController#resume_{in,out}_endpoint`. This works, but is not very satisfying and probably encourages fairly brittle code since the Interface actually _does_ have full control of the low-level device if it wants. A specialized struct which is provided by `Device` and wraps the particular Endpoint is probably close to the right solution.

Second, there is currently no way to handle CDC-ACM, which has a very unique requirement: a CDC-ACM device actually uses _two_ interfaces (this is required by the USB CDC specification) that are related to each other, where the "control" interface's class descriptors include two descriptors that need to refer to the absolute interface value of the two interfaces (the union descriptor and call management descriptors). I have not yet come up with a clean way to model this without dramatically complicated the stack solely for supporting this edge (as far as I can tell) case. 

### Testing Strategy

The second commit in this PR modifies the nano33ble board to use the new USB stack with multiple BulkSimple instances: one used for process_console and the debug writer via a UartMux and the other used for the console driver. These show up as separate `/dev/ttyUSB*` devices on Linux. You can run `tockloader listen` twice, selecting either device, and they will operate independently.

For this to work, you need to, first, enable the generic usb serial driver:

```
$ sudo modprobe usbserial
```

Then tell it that you want to interpret your device as a usbserial by supplying the vendor and product id (set in the board configuration to 0x6667 and 0xabcd):

```
$ echo 6667 abcd | sudo tee /sys/bus/usb-serial/drivers/generic/new_id
```

I have not tested but very much doubt this will work on any other host operating system. The generic serial driver is meant for prototyping only (https://www.kernel.org/doc/html/latest/usb/usb-serial.html#generic-serial-driver) and I don't think there are equivalents on other OSs.

### TODO or Help Wanted

- [ ] Document, document, document
- [ ] Remove the demonstration commit
- [ ] Solve or avoid the CDC-ACM problem
- [ ] ...

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/boards/nano33ble/src/main.rs
+++ b/boards/nano33ble/src/main.rs
@@ -302,7 +302,7 @@ pub unsafe fn main() {
     //--------------------------------------------------------------------------
 
     let dynamic_deferred_call_clients =
-        static_init!([DynamicDeferredCallClientState; 4], Default::default());
+        static_init!([DynamicDeferredCallClientState; 10], Default::default());
     let dynamic_deferred_caller = static_init!(
         DynamicDeferredCall,
         DynamicDeferredCall::new(dynamic_deferred_call_clients)
@@ -334,6 +334,8 @@ pub unsafe fn main() {
 
     let bulk_simple = static_init!(capsules::usb::bulk_simple::BulkSimple<'static, nrf52::usbd::Usbd>,
             capsules::usb::bulk_simple::BulkSimple::new(&nrf52840_peripherals.usbd));
+    let bulk_echo = static_init!(capsules::usb::bulk_simple::BulkSimple<'static, nrf52::usbd::Usbd>,
+            capsules::usb::bulk_simple::BulkSimple::new(&nrf52840_peripherals.usbd));
     let usb = {
         use capsules::usb::{device, configuration};
         // Create the strings we include in the USB descriptor. We use the hardcoded
@@ -356,32 +358,17 @@ pub unsafe fn main() {
         };
 
         let configuration = configuration::SimpleConfiguration {
-            interfaces: [bulk_simple] as [&dyn configuration::Interface; 1],
+            interfaces: [bulk_echo, bulk_simple] as [&dyn configuration::Interface; 2],
             is_self_powered: false,
             supports_remote_wakeup: false,
             max_power: 50,
             _f: core::marker::PhantomData,
         };
         static_init!(
-            device::Device<'static, nrf52::usbd::Usbd, configuration::SimpleConfiguration<'static, [&'static dyn configuration::Interface; 1]>>,
+            device::Device<'static, nrf52::usbd::Usbd, configuration::SimpleConfiguration<'static, [&'static dyn configuration::Interface; 2]>>,
             device::Device::new(&nrf52840_peripherals.usbd, device_descriptor, configuration, 0x0409))
     };
     kernel::hil::usb::UsbController::set_client(&nrf52840_peripherals.usbd, usb);
-    /*let cdc = components::cdc::CdcAcmComponent::new(
-        &nrf52840_peripherals.usbd,
-        capsules::usb::cdc::MAX_CTRL_PACKET_SIZE_NRF52840,
-        0x2341,
-        0x005a,
-        strings,
-        mux_alarm,
-        dynamic_deferred_caller,
-        Some(&baud_rate_reset_bootloader_enter),
-    )
-    .finalize(components::usb_cdc_acm_component_helper!(
-        nrf52::usbd::Usbd,
-        nrf52::rtc::Rtc
-    ));
-    CDC_REF_FOR_PANIC = Some(cdc); //for use by panic handler*/
 
     // Create a shared UART channel for the console and for kernel debug.
     let uart_mux = components::console::UartMuxComponent::new(bulk_simple, 115200, dynamic_deferred_caller)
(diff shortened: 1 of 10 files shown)


alistair23:
I think cleaning up the USB stack would be great and overall I like this layered direction.

I think we should focus on replacing the current implementation with a new one though. Which then means the CDC-ACM issue needs to be fixed. Having two live along side each other seems like a nightmare to deal with.

alevy:
I agree with @bradjc and @alistair23 that this is not a good solution without also solving CDC-ACM. Any ideas?

There are two approaches that I think could work:

1. Expose more global information to the enumeration functions, such that the CDC-ACM's interfaces can refer to each other's dynamically generated interface numbers in descriptors. I think this can work and is fairly general, but I can't figure out a way for it to not be a super hacky interface, so I don't love this.

2. CDC-ACM would be a special implementation of the Configuration trait that that handles it specially. Just like the default Configuration trait implementation, it could still allow additional interfaces, but the two CDC-ACM interfaces could have a guaranteed interface number since they would be specialized for this Configuration instance. This might entail modifying the current interface a little bit, but I don't think in a problematic way. The downside is that you would need this for every kind of composite device and it wouldn't necessarily be straight forward to combine composite devices (though, every thing can be solved with a layer of indirection). So it's not super general.

3. Surely there is a solution with a different layering or interface that I just haven't thought of.

bradjc:

PR #2896, opened 2021-11-10
util: Add CopyOrErr trait for a non-panicing copy_from_slice.

granaghan:
Change-Id: I0b5217bef21553e4db976ecc4d1f816da690f329

### Pull Request Overview

Adds copy_from_slice_or_err to process buffers and a trait to implement this on integer types.
Using this instead of copy_from_slice can reduce binary size by eliminating potential panics.

### Testing Strategy

Replacing instances in my own capsules and libraries with copy_from_slice_or_err

### TODO or Help Wanted

### Documentation Updated
N/A?

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/ble_advertising_driver.rs
+++ b/capsules/src/ble_advertising_driver.rs
@@ -114,6 +114,7 @@ use kernel::processbuffer::{ReadOnlyProcessBuffer, ReadableProcessBuffer};
 use kernel::processbuffer::{ReadWriteProcessBuffer, WriteableProcessBuffer};
 use kernel::syscall::{CommandReturn, SyscallDriver};
 use kernel::utilities::cells::OptionalCell;
+use kernel::utilities::copy_slice::CopyOrErr;
 use kernel::{ErrorCode, ProcessId};
 
 /// Syscall driver number.
@@ -272,7 +273,7 @@ impl App {
                             header[1] = (payload_len & 0x3f) as u8;
 
                             let (adva, data) = payload.split_at_mut(6);
-                            adva.copy_from_slice(&self.address);
+                            adva.copy_from_slice_or_err(&self.address)?;
                             adv_data_corrected.copy_to_slice(&mut data[..adv_data_len]);
                         }
                         let total_len = cmp::min(PACKET_LENGTH, payload_len + 2);
@@ -466,8 +467,9 @@ where
                     let success = app
                         .scan_buffer
                         .mut_enter(|userland| {
-                            userland[0..len as usize].copy_from_slice(&buf[0..len as usize]);
-                            true
+                            userland[0..len as usize]
+                                .copy_from_slice_or_err(&buf[0..len as usize])
+                                .is_ok()
                         })
                         .unwrap_or(false);
 
--- a/kernel/src/processbuffer.rs
+++ b/kernel/src/processbuffer.rs
@@ -24,6 +24,7 @@ use core::ops::{Index, Range, RangeFrom, RangeTo};
 
 use crate::capabilities;
 use crate::process::{self, ProcessId};
+use crate::ErrorCode;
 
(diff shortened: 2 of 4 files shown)


kernel/src/utilities/copy.rs
```diff
@@ -0,0 +1,88 @@
+use crate::ErrorCode;
+use core::ptr;
+
+pub trait CopyOrErr {
+    /// Copies a nonoverlapping slice from src to self. Returns Err(ErrorCode) is source and slef
```
hudson-ayers, line 5:
```suggestion
    /// Copies a nonoverlapping slice from src to self. Returns Err(ErrorCode) if source and self
```

hudson-ayers (dismissed):
I am a fan of this change. Hopefully upstream Rust catches up to the need for non-panicking APIs and gives us something like this in core one day.

bradjc:
I think if we add `CopyOrErr` we should use it somewhere. It would also be nice to change it from "copy.rs" to something more specific like "slice_copy.rs".

Also, why do the process_buffer changes help? Is it just that the compiler is more likely to detect that the panic can never happen with this change?

hudson-ayers:
> I think if we add `CopyOrErr` we should use it somewhere. It would also be nice to change it from "copy.rs" to something more specific like "slice_copy.rs".

Agreed we should use it somewhere -- https://github.com/tock/tock/blob/master/capsules/src/ble_advertising_driver.rs#L469 is one place where we could use it, and return an error to userspace in the case of failure (in the following upcall).

> Also, why do the process_buffer changes help? Is it just that the compiler is more likely to detect that the panic can never happen with this change?

As-is the process_buffer changes don't do anything, beyond add a non-panicking API for capsules that want to use it. I think the reason for changing the existing copy_from_slice etc. implementations to use the error-providing ones underneath is so that there is not a near-duplicate implementation of the panicking version and the error providing version.

granaghan:
The process_buffer changes don't really help with size since the panic function is marked cold. It's just code deduplication and adding consistent APIs since they seem to be trying to look as much like slices as possible.

I'm planning to try to add get and get_mut as well so I can get rid of subslicing that may panic.

kernel/src/utilities/copy.rs
granaghan: Done along with the rename update in ble_advertising_driver.

alistair23 (approved):

granaghan:
Looks like I need someone to approve running the remaining workflows.

brghena:
Workflows started

hudson-ayers (dismissed):

bradjc (approved):

granaghan:
Rebased.

hudson-ayers (approved):

alistair23 (approved):

bradjc (changes requested):
Wait, there still isn't anything using `kernel::utilities::copy_slice`, correct? If we want to add that utility code we should demonstrate it somewhere.

hudson-ayers:
The BLE driver now uses `kernel::utilities::copy_slice::copy_from_slice_or_err()`, are you saying Brian should find a use for every function in that file?

bradjc:
> The BLE driver now uses `kernel::utilities::copy_slice::copy_from_slice_or_err()`, are you saying Brian should find a use for every function in that file?

No, but I don't think it does. I was able to delete the entire `kernel/src/utilities/copy_slice.rs` file and `make check` still works fine. I don't think anything is using the `CopyOrErr` trait.

granaghan:
Looks like it's using copy_from_slice_or_err on a ProcessSlice. I'm guessing most things in capsules will be ProcessSlice-s.

TicKV uses copy_from_slice on [u8], so I'll update that.

bradjc (approved):

PR #2898, opened 2021-11-12
OpenTitan: OTBN: A few more OTBN fixes

alistair23:
### Pull Request Overview

This PR includes a few more OTBN cleanups

### Testing Strategy

Running on the CW310.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/src/tests/otbn.rs
+++ b/boards/opentitan/src/tests/otbn.rs
@@ -136,14 +136,17 @@ fn otbn_run_rsa_binary() {
         let slice = unsafe { core::slice::from_raw_parts_mut(dmem_start as *mut u8, dmem_length) };
 
         CALLBACK.reset();
-        assert_eq!(otbn.load_data(0, slice), Ok(()));
+        assert_eq!(otbn.load_data(0, LeasableBuffer::new(slice)), Ok(()));
         run_kernel_op(1000);
 
         // Set the RSA mode to encryption
         // The address is the offset of `mode` in the RSA elf
         unsafe {
             TEMP[0] = 1;
-            assert_eq!(otbn.load_data(0, &mut TEMP[0..4]), Ok(()));
+            assert_eq!(
+                otbn.load_data(0, LeasableBuffer::new(&mut TEMP[0..4])),
+                Ok(())
+            );
         }
         run_kernel_op(1000);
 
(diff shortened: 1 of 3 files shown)


chips/lowrisc/src/virtual_otbn.rs
```diff
@@ -57,7 +57,7 @@ impl<'a> VirtualMuxAccel<'a> {
     pub fn load_data(
         &self,
         address: usize,
-        data: &'static mut [u8],
+        data: LeasableBuffer<'static, u8>,
```
hudson-ayers, line 60: The intended use of `LeasableBuffer` is for the `LeasableBuffer` to be returned via a callback (or in the error case), and then `reset()` can be called to regain access to the entire buffer, but I don't see how `LeasableBuffer` is really providing much as used here.

alistair23: Interesting, previously I have been passing in a `LeasableBuffer` and then returning just the static buffer (https://github.com/tock/tock/blob/124e7ac9c97dbe9b4bc719a0b884ed827778f55e/kernel/src/hil/digest.rs#L76).

The idea being that the caller can specify a active slice using `LeasableBuffer` but then just wants the full buffer to be returned.

In this case it's used to provide an active slice, so the caller can specify a subset range of the buffer to be used

hudson-ayers: I see. Okay sounds fine then

boards/opentitan/src/tests/otbn.rs
```diff
@@ -136,29 +136,38 @@ fn otbn_run_rsa_binary() {
...
 
         // Set the RSA length
         // The address is the offset of `n_limbs` in the RSA elf
         unsafe {
             TEMP[0] = (MODULUS.len() / 32) as u8;
-            assert_eq!(otbn.load_data(4, &mut TEMP[0..4]), Ok(()));
+            assert_eq!(
+                otbn.load_data(4, LeasableBuffer::new(&mut TEMP[0..4])),
+                Ok(())
+            );
```
hudson-ayers, line 160: This isn't really a comment on this PR, since this code already did this before, but this is probably UB -- you're taking multiple  `&'static mut` references to the TEMP buffer and passing them to callees. Given what the underlying code does and that this is single threaded it probably will do the correct thing in practice, and this is just test code. Still, having multiple static buffers (one for each use), and initializing them using static_init!() would probably be safer.

alistair23:
Ping!

boards/opentitan/src/tests/otbn.rs
alistair23: That's probably a good idea. I'll work on this

hudson-ayers (approved):
bors r+

PR #2899, opened 2021-11-12
OpenTitan: Add support for hardware RSA operations

alistair23:
### Pull Request Overview

This PR builds on top of:
 * https://github.com/tock/tock/pull/2898
 * https://github.com/tock/tock/pull/2892
 * https://github.com/tock/tock/pull/2839

and adds support for RSA hardware operations using the OTBN.

### Testing Strategy

Running the RSA test in this PR on the CW310 FPGA.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


phil-levis:
Following up on the 11/18 OT call, I started looking into factory methods for key generation to better constrain the types. This is tricky in the context of something like OTBN. The types of keys it can support depends on what programs are loaded into it. So unless OTBN is statically configured, there's no way to be sure what keys it will support. This raises the subsequent issue of possibly passing incorrectly formatted or improper keys to the accelerator. For safety, this suggests to me that Tock should always configure imem statically, or at least directly from kernel code, such that it can change the instructions needed to support a particular key.

alistair23:
Pushed an update to more clearly specify what is/isn't allowed in the buffer length

alistair23:
Ping!

alistair23:
Ping!

phil-levis (approved):

PR #2907, opened 2021-11-24
syscall: Add missing TryInto import.

granaghan:
### Pull Request Overview

Add a missing TryInto import in cortex-m syscall. For some reason `make prepush` doesn't catch this on master but did catch in on our internal fork.

### Testing Strategy

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/syscall.rs
+++ b/arch/cortex-m/src/syscall.rs
@@ -1,6 +1,7 @@
 //! Implementation of the architecture-specific portions of the kernel-userland
 //! system call interface.
 
+use core::convert::TryInto;
 use core::fmt::Write;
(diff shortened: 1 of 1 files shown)


alistair23 (approved):

hudson-ayers (changes requested):
TryFrom / TryInto are included in the prelude in Rust 2021 edition (see https://doc.rust-lang.org/edition-guide/rust-2021/prelude.html), so we do not need this.

If you depended directly on the upstream crate, you still would not need it, as I understand -- I suspect you just need it downstream because you are still using edition 2018 even in the forked crate.

PR #2914, opened 2021-12-07
Propose a draft TRD for app completion codes

kupiakos:
Based on discussion in https://github.com/tock/libtock-rs/pull/347. This aims to define little about the `exit` syscall other than the common convention that zero indicates success, and non-zero indicates some sort of failure.

[Rendered](https://github.com/kupiakos/tock/blob/trd106/doc/reference/trd106-completion-codes.md)

### DIFF
--- a/doc/reference/trd106-completion-codes.md
+++ b/doc/reference/trd106-completion-codes.md
@@ -7,7 +7,7 @@ Application Completion Codes
 **Status:** Draft<br/>
 **Author:** Alyssa Haroldsen<br/>
 **Draft-Created:** December 6, 2021<br/>
-**Draft-Modified:** December 6, 2021<br/>
+**Draft-Modified:** December 7, 2021<br/>
 **Draft-Version:** 1<br/>
 **Draft-Discuss:** tock-dev@googlegroups.com</br>
 
@@ -34,9 +34,9 @@ understandable to newcomers by following the principle of least astonishment.
 
 3 Design
 ===============================
-An application exiting normally via the `exit` syscall MUST use a completion
-code of `0` to indicate normal termination. A non-zero completion code SHOULD be
-used to indicate abnormal termination. A non-zero completion code MAY be the 
+A completion code of `0` passed to the `exit` syscall MUST indicate normal app
+termination. A non-zero completion code SHOULD be used to indicate abnormal
+termination. A completion code between `1` and `1024` inclusive SHOULD be the
 same value as one of the error codes specified in [TRD 104][error-codes].
 
 The kernel MAY treat zero and non-zero completion codes differently.
@@ -44,7 +44,7 @@ The kernel MAY treat zero and non-zero completion codes differently.
 | **Completion Code** | **Meaning** |
 | ------------------- | ----------- |
 | 0                   | Success     |
-| 1-1024              | MAY be a [TRD 104 error code][error-codes] |
+| 1-1024              | SHOULD be a [TRD 104 error code][error-codes] |
 | 1025-`u32::MAX`     | Not defined |
 
 4 Implementation
--- a/doc/reference/trd106-completion-codes.md
+++ b/doc/reference/trd106-completion-codes.md
@@ -68,12 +68,34 @@ to values.
 | ------------------- | --------------------------------------------- |
 | 0                   | Success                                       |
 | 1-1024              | SHOULD be a [TRD 104 error code][error-codes] |
-| 1025-`u32::MAX`     | Reserved                                      |
+| 1025-`u32::MAX`     | Not defined                                   |
 
 4 Implementation
 ===============================
 As of writing, libtock [currently implements][termination] this TRD via the
 `Termination` trait.
+
+```rust
+pub trait Termination {
+    fn complete<S: Syscalls>(self) -> !;
+}
+
+impl Termination for () {
+    fn complete<S: Syscalls>(self) -> ! {
+        S::exit_terminate(0)
+    }
+}
+
+impl Termination for Result<(), ErrorCode> {
+    fn complete<S: Syscalls>(self) -> ! {
+        let exit_code = match self {
+            Ok(()) => 0,
+            Err(ec) => ec as u32,
+        };
+        S::exit_terminate(exit_code);
+    }
+}
+```
 
 5 Author's Address
 ===============================
--- a/doc/reference/trd106-completion-codes.md
+++ b/doc/reference/trd106-completion-codes.md
@@ -13,8 +13,9 @@ Application Completion Codes
 
 Abstract
 -------------------------------
-This document describes the expected behavior of application completion codes
-when terminating via the `exit` syscall, as described in [TRD 104][exit-syscall].
+This advisory document describes the expected behavior of application completion
+codes when terminating via the `exit` syscall, as described in
+[TRD 104][exit-syscall].
 
 1 Introduction
 ===============================
@@ -28,9 +29,14 @@ platforms.
 
 2 Design Considerations
 ===============================
-When possible, Tock should follow existing conventions and terminology from
-other major platforms. This assists in helping the project be more
-understandable to newcomers by following the principle of least astonishment.
+When possible, Tock applications should follow existing conventions and
+terminology from other major platforms. This assists in helping the project be
+more understandable to newcomers by following the principle of least
+astonishment.
+
+This advisory document provides guidance for the ecosystem of Tock applications
+using the `exit` syscall, and does not define the behavior of the syscall
+itself.
 
 3 Design
 ===============================
@@ -41,11 +47,11 @@ same value as one of the error codes specified in [TRD 104][error-codes].
 
 The kernel MAY treat zero and non-zero completion codes differently.
 
-| **Completion Code** | **Meaning** |
-| ------------------- | ----------- |
-| 0                   | Success     |
+| **Completion Code** | **Meaning**                                   |
+| ------------------- | --------------------------------------------- |
+| 0                   | Success                                       |
 | 1-1024              | SHOULD be a [TRD 104 error code][error-codes] |
-| 1025-`u32::MAX`     | Not defined |
+| 1025-`u32::MAX`     | Not defined                                   |
 
 4 Implementation
 ===============================
--- a/doc/reference/trd106-completion-codes.md
+++ b/doc/reference/trd106-completion-codes.md
@@ -7,7 +7,7 @@ Application Completion Codes
 **Status:** Draft<br/>
 **Author:** Alyssa Haroldsen<br/>
 **Draft-Created:** December 6, 2021<br/>
-**Draft-Modified:** December 7, 2021<br/>
+**Draft-Modified:** January 25, 2022<br/>
 **Draft-Version:** 1<br/>
 **Draft-Discuss:** tock-dev@googlegroups.com</br>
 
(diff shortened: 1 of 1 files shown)


doc/reference/trd106-completion-codes.md
```diff
@@ -0,0 +1,57 @@
...
+An application exiting normally via the `exit` syscall MUST use a completion
+code of `0` to indicate normal termination. A non-zero completion code SHOULD be
+used to indicate abnormal termination. A non-zero completion code MAY be the 
+same value as one of the error codes specified in [TRD 104][error-codes].
+
+The kernel MAY treat zero and non-zero completion codes differently.
+
+4 Implementation
+===============================
+libtock 1.0 [currently implements][termination] this TRD via the
```
jrvanwhy, line 46: Shorten this to just `libtock`.

I've been colloquially using the phrase "`libtock-rs` 1.0" to refer to the `libtock-rs` codebase that was designed for Tock 1.0, but we haven't actually named either a 1.0 or a 2.0 release of `libtock`.

doc/reference/trd106-completion-codes.md
```diff
@@ -0,0 +1,57 @@
...
+When possible, Tock should follow existing conventions and terminology from
+other major platforms. This assists in helping the project be more
+understandable to newcomers by following the principle of least astonishment.
+
+3 Design
+===============================
+An application exiting normally via the `exit` syscall MUST use a completion
+code of `0` to indicate normal termination. A non-zero completion code SHOULD be
+used to indicate abnormal termination. A non-zero completion code MAY be the 
+same value as one of the error codes specified in [TRD 104][error-codes].
```
jrvanwhy, line 40: It's probably worth making a table here. Something like:

Completion Code | Meaning
------------- | --------
0              | Success
1 - 1024   | Error code as defined in TRD 104
everything else | Not yet defined

kupiakos: Notice the MAY - I specifically don't want to reserve 1-1024. The only required convention is 0 vs. non-zero.

kupiakos: I added a table that I think indicates this

kupiakos:
(apologies on the force push, I accidentally did a Github merge commit)

doc/reference/trd106-completion-codes.md
kupiakos: Done

doc/reference/trd106-completion-codes.md
hudson-ayers: IMO, it seems a little confusing to use MAY here. Either we should reserve 1-1024, or we should just say that 1-MAX is not yet defined. Specifying that it may be defined as an error code, or may not, does not seem to have much value

kupiakos: How would you feel about SHOULD? The MAY is to indicate that using 1-1024 for TRD 104 error codes is a good baseline, but an application not following it is still well-formed. I'm thinking from the perspective of an engineer trying to figure out what completion code 12 is, reading this doc, and seeing that it indicates the error code UNINSTALLED by default, providing a valuable hint.

hudson-ayers: I am happy with SHOULD

kupiakos: Thoughts on the new language? I also changed the first sentence, switching the logic from "success implies a 0 completion code" to "0 completion code implies success".

jrvanwhy: I'm happy with the new language + table.

jrvanwhy (dismissed):

alistair23 (dismissed):

hudson-ayers (dismissed):

bradjc:
I'm wondering if this is appropriate for the kernel to have. It doesn't specify what the kernel does, but rather what processes should do. Do we want to tell processes what they should do? If we do, is a TRD the correct mechanism?

I've long been an advocate for not constraining processes beyond what is specified by syscalls. Should we consider this an update to the exit syscall spec? Are we expecting the kernel to act based on the completion code? How does that interact with exit-terminate vs exit-restart?

What about definining in libtock-rs and libtock-c a set of defined `COMPLETION_CODE_X` constants that processes can use if they wish?

bradjc:
From the call today:

- Generally this seems useful, and having it as a TRD is reasonable as there really isn't a better place at the moment.
- The text should be updated to make it clear this is advisory and separate from the exit syscall (i.e. the kernel cannot do anything with these completion codes even if applications use them).
- The main use case seems to be giving application developers or application library developers some guidance on what completion codes to use. This should be described in the document, and some likely useful completion codes should be enumerated.

kupiakos:
> * The text should be updated to make it clear this is advisory and separate from the exit syscall (i.e. the kernel cannot do anything with these completion codes even if applications use them).

@bradjc Are you suggesting that `The kernel MAY treat zero and non-zero completion codes differently.` be changed to `The kernel MUST treat zero and non-zero completion codes the same.` or removed?

> The text should be updated to make it clear this is advisory and separate from the exit syscall (i.e. the kernel cannot do anything with these completion codes even if applications use them).

I hope I've made that more clear with 1922c4c.

> some likely useful completion codes should be enumerated
The suggested convention is to use a TRD 104 error code, so `1` means FAIL, `2` means BUSY, and so on. Is the link insufficient?

phil-levis:
Following up on the call from today. I think the statement `The kernel MUST treat zero and non-zero completion codes the same` is probably too strong, or at least ambiguous, because it could be interpreted as "you can't print different strings for them." I might say something instead like `The kernel MUST NOT assume any semantic meaning to completion codes and take actions based on their values besides printing error messages. While there are common and conventional uses of certain values, applications are not required to follow these and may assign their own semantic meanings to values.`

I think the current formulation makes sense and is good. I think the text could possibly be a bit more explanatory. For example, talk explicitly about why it is useful to distinguish success and failure exits. Just being compatible isn't necessarily a good reason; it could be that the previous systems were designed to achieve very different, even contrary, goals. 

For example, explain the SHOULD on 104 error codes: say that following allows software to infer something about the cause of the error and perhaps print a useful message, but on the other hand if a process needs to do something else (e.g., for some kind of compatibility) it can.

phil-levis:
Bump?

hudson-ayers:
I went ahead and added a commit that attempts to address Phil's comment / the discussion from last week's call (I hope you don't mind my pushing to your branch, @kupiakos , just trying to keep this moving forward). The one notable change in my edits is the idea that boards could provide a process-restart policy which matches on whether the process in question exited successfully or with a failure -- I think it is important that is something which is allowed by this TRD.

kupiakos:
@hudson-ayers those edits look great! You're all good - I wouldn't have granted reviewers push access if I wasn't OK with it :)
I was having trouble wording this in a way that I thought would be fully satisfying.

@phil-levis apologies on missing this - I'm still not used to Github notifications and I missed the email

lschuermann (approved):

doc/reference/trd106-completion-codes.md
```diff
@@ -0,0 +1,86 @@
...
+codes or take actions based on their values besides printing error messages.
+While there are common and conventional uses of certain values, applications
+are not required to follow these and may assign their own semantic meanings
+to values.
+
+| **Completion Code** | **Meaning**                                   |
+| ------------------- | --------------------------------------------- |
+| 0                   | Success                                       |
+| 1-1024              | SHOULD be a [TRD 104 error code][error-codes] |
+| 1025-`u32::MAX`     | Not defined                                   |
```
hudson-ayers, line 71:
```suggestion
| 1025-`u32::MAX`     | Reserved                                     |
```

doc/reference/trd106-completion-codes.md
```diff
@@ -0,0 +1,86 @@
...
+A completion code between `1` and `1024` inclusive SHOULD be the
+same value as one of the error codes specified in [TRD 104][error-codes].
+This requirement is a SHOULD rather than a MUST because it is useful in the
+common case (it allows software to infer something about the cause of an
+error that led to an exit, and possibly print a useful message) but also
+allows a process to do something else if needed (e.g. for compatibility
+with some other standard of exit codes).
+
+Accordingly, the core kernel MUST NOT assume any semantic meaning for completion
+codes or take actions based on their values besides printing error messages.
```
lschuermann, line 62:
```suggestion
Accordingly, the core kernel MUST NOT assume any semantic meaning for completion
codes or take actions based on their values besides printing error messages, as long as
there is no explicit contract between the kernel and application regarding completion
code interpretation.
```

I like that this does imply that the kernel will not take arbitrary action based solely on the completion codes, but depending on how this is read, it does seem to severely hinder the usefulness of return codes if an app _wants_ do define an explicit contract with the kernel. It might be more clear that this is still an option if we add this phrase explicitly.

lschuermann (commented):
I like the implications of this, especially that by default it avoids accidental misinterpretation of completion codes by the kernel. I think we should allow it though, if both parties agree.

brghena (dismissed):

doc/reference/trd106-completion-codes.md
kupiakos: I like this, though wouldn't "core kernel" mean that capsules still have the freedom to assume semantic meaning for completion codes?

lschuermann: Yes, I think that interpretation is reasonable. Maybe it's worth adding an emphasis to core, as I didn't really realize capsules were deliberately excluded from this statement while reading.

I still think that, given the scheduler and process fault handler be considered part of the core kernel, _allowing_ an explicit contract between an app and the kernel may make sense.

kupiakos: @hudson-ayers, @phil-levis, what are your thoughts on this? I'm fine with the suggested wording here.

hudson-ayers: I am happy with Leon's suggested wording as well

phil-levis: Sorry, missed this chain. What's an explicit contract? How would such a contract emerge?

phil-levis: @lschuermann @hudson-ayers @kupiakos Just wanted to follow up on this. Would you like me to suggest some text to describe what we talked about today, or Leon, I think you said you'd do that?

lschuermann: @phil-levis Sure, didn't get to it until now.

Based on the feedback I suppose it's good to establish my primary motivations behind this change, if only for future record:
- the current wording might be to. restrictive. While downstream users don't technically have to abide to these TRDs, they will the direction of the upstream kernel takes. If we rule out that the kernel can take actions based on app completion codes, we might design internal interfaces that reflect that decision, making it harder for downstream users to implement something like this.
- we _MUST_ ensure that the kernel does not take action on app completion codes as long as there is no guarantee that the kernel knows how to interpret a specific app's completion codes. It is unacceptable for the core kernel to misidentify an app or otherwise misinterpret its completion codes and take action based on this information.

Thus, a more elaborate text might say

```suggestion
Accordingly, the core kernel MUST NOT assume any semantic meaning for completion
codes. Without a revision of this TRD, the upstream core kernel may not take actions
based on their values besides printing error messages. Future versions of the kernel
may implement mechanisms to take actions based on an app's completion code,
provided that
- the app provides a specification for its completion code space, specifically concerning
  completion codes encoding success and failure cases. It may further specify which types
  of actions the core kernel may enact based on its completion codes.
- the kernel adheres to the app's completion code specification and only takes actions
  allowed by the specification based on the completion code returned. This implies
  that an application with a specification that does not include any instructions about
  actions taken in response to a return code will not be different from an app without such an
  application. The kernel may not implement every action described in this specification.
- the kernel can reliably identify the app and associate it with its completion code specification.
```

This is very elaborate on what a future design could and should look like, but it does not yet fully describe how such a specification must look like and what qualifies as a completion-code specification. It does not necessarily have any benefits for us in the upstream Tock codebase right now, other than being able to design internal interfaces to incorporate such future behavior. Still, it provides downstream users with essential and sensible guidelines on how to implement such a mechanism, which is similar and in the best case compatible with future upstream requirements.

phil-levis: How about just trim it down to"

Accordingly, the core kernel MUST NOT assume any semantic meaning for completion codes or take actions based on their values besides printing error messages unless 
  - there is a specification of a particular application's completion code space written in a TRD, and 
  - the kernel can reliably identify that application and associate it with this specification.

hudson-ayers: That works for me

lschuermann: @phil-levis Just now catching up on the backlog. Sure, that sounds good and handles all of my concerns. Here's a suggestion which can be applied through GitHub:

```suggestion
Accordingly, the core kernel MUST NOT assume any semantic meaning for completion
codes or take actions based on their values besides printing error messages
unless

- there is a specification of a particular application's completion code space
  written in a TRD, and

- the kernel can reliably identify that application and associate it with this
  specification.
```

bradjc (approved):

doc/reference/trd106-completion-codes.md
hudson-ayers: I went ahead and applied the suggestion so that outstanding approvals will not get wiped when it is applied

hudson-ayers (approved):

bradjc (approved):

lschuermann (approved):

PR #2918, opened 2021-12-09
OpenTitan: Convert the OTBN to use syncronous calls

alistair23:
### Pull Request Overview

Currently the OTBN load functions are asynchronous and the callback is triggered by a a deferred call. There is no need for the callback though as the entire operation happens synchronously anyway. The async part is left over from the original HIL implementation, but that was dropped in favour of just using `pub` functions for the `chip` driver.

This patch makes the data loading a synchronous operation. This means we don't need `'static mut` buffers and avoids the overhead of the callbacks.

### Testing Strategy

Run the unit tests on the CW310.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/src/main.rs
+++ b/boards/opentitan/src/main.rs
@@ -207,7 +207,7 @@ unsafe fn setup() -> (
     let board_kernel = static_init!(kernel::Kernel, kernel::Kernel::new(&PROCESSES));
 
     let dynamic_deferred_call_clients =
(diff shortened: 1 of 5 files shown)


hudson-ayers (dismissed):

bradjc (approved):

alistair23:
Rebased on master so now the CI is passing

hudson-ayers (approved):

bradjc (approved):

PR #2921, opened 2021-12-16
boards/opentitan: tests: otbn: Improve failure output

alistair23:
### Pull Request Overview

If the OTBN tests case fails we previously just printed:

```
   [FAIL]
```

Without even specifying the test case.

This patch fixes that so we print the test case no matter what and then
explain the failure.

### Testing Strategy

`make test`

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


alistair23:
@thulithwilfred

hudson-ayers (approved):

lschuermann (approved):

PR #2922, opened 2021-12-16
tock-register-interface: replace register tests by const assertions

lschuermann:
### Pull Request Overview

This replaces the automatically generated tests for register structs defined through the `register_structs!` macro by static assertions evaluated at compile time.

This has some significant advantages over the current approach:

- errors in the struct offsets and violation of other invariants is detected immediately on compilation and not only when running tests. This increases developer ergonomics significantly and reduces the potential for errors and weird behavior at runtime.

- the crate no longer generates tests which rely on the standard library being present in the users' code. While the library itself still has tests which rely on the presence of the standard library, these tests are for internal validation only. In `no_std` environments, relying on the standard library even just for `#[test]` functions can be problematic. For developers, the choice has been to either use the standard library and have test, implement a workaround to provide the standard library when running tests but not otherwise, or not generate the tests at all. These changes enable us to make the library purely `#[no_std]` compatible and not impose any standard library requirements on user code.

It does not come without drawbacks though. Because Rust's const panic features, especially w.r.t. string formatting, are still very limited, the provided debug information is significantly reduced. For instance, it as of now seems impossible to interpolate any const variable content into the assertion message. This is likely to be resolved in the future, when Rust allows more elaborate string formatting techniques at compile time. As an additional consequence, the assertions code is less readable.

The generated error messages are quite prominent during the compilation process and easy to identify:

    error[E0080]: evaluation of constant value failed
      --> chips/lowrisc/src/uart.rs:16:1
       |
    16 | / register_structs! {
    17 | |     pub UartRegisters {
    18 | |         (0x00 => intr_state: ReadWrite<u32>),
    19 | |         (0x04 => intr_enable: ReadWrite<u32>),
    ...  |
    41 | |     }
    42 | | }
       | |_^ the evaluated program panicked at 'Invalid end offset for field val (expected 49 but actual value differs)', chips/lowrisc/src/uart.rs:16:1

### Testing Strategy

### TODO or Help Wanted

- [x] Remove the `std_unit_tests` feature and only make the standard library available when the crate itself is tested.
- [x] Update the README.md documentation.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/libraries/tock-register-interface/Cargo.toml
+++ b/libraries/tock-register-interface/Cargo.toml
@@ -15,14 +15,9 @@ edition = "2021"
 travis-ci = { repository = "tock/tock", branch = "master" }
 
 [features]
-default = [ "register_types", "std_unit_tests" ]
+default = [ "register_types" ]
 
 # Include actual register types (except LocalRegisterCopy). Disabling
 # the feature makes this an interface-only library and removes all
 # usage of unsafe code
 register_types = []
-
-# Feature flag to enable generation of unit tests for the
-# registers. Enabling this may break compilation in
-# `custom-test-frameworks` environments.
-std_unit_tests = []
--- a/libraries/tock-register-interface/README.md
+++ b/libraries/tock-register-interface/README.md
@@ -90,22 +90,16 @@ struct Registers {
 }
 ```
 
-By default, `std` unit tests for the struct are generated as well (that is,
-tests attributed with `#[test]`). The unit tests make sure that the offsets and
-padding are consistent with the actual fields in the struct, and that alignment
-is correct.
+This crate will generate additional, compile time (`const`) assertions
+to validate various invariants of the register structs, such as
 
-Since those tests would break compilation in `custom-test-frameworks`
-environments, it is possible to opt out of the test generation. To do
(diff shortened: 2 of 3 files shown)


alistair23 (approved):

bradjc:
I tried compiling the nano33ble after removing one of the "reserved" fields in acomp.rs and got:

```
error[E0080]: evaluation of constant value failed
   --> chips/nrf52/src/acomp.rs:63:1
    |
63  | / register_structs! {
64  | |     CompRegisters {
65  | |         /// TASK REGISTERS
66  | |         /// Trigger task by writing 1
...   |
125 | |     }
126 | | }
    | |_^ the evaluated program panicked at 'Invalid end offset for field tasks_sample (expected 256 but actual value differs)', chips/nrf52/src/acomp.rs:63:1
    |
    = note: this error originates in the macro `$crate::panic::panic_2021` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I wanted to include this to (hopefully) help anyone googling this error in the future. The issue is there is a gap in my register definitions.

bradjc (approved):
Excellent

hudson-ayers (approved):

PR #2926, opened 2021-12-20
boards/opentitan: tests: otbn: Remove global static buffers

thulithwilfred:

### Pull Request Overview
Removed global static buffers from the OTBN test.

### Testing Strategy
`make test`

`make ci-all`

### Documentation Updated

### Formatting
- [x] Ran `make fmt`.

### DIFF
--- a/boards/opentitan/src/tests/otbn.rs
+++ b/boards/opentitan/src/tests/otbn.rs
@@ -1,11 +1,10 @@
 use crate::tests::run_kernel_op;
 use crate::PERIPHERALS;
 use core::cell::Cell;
+use kernel::static_init;
+use kernel::utilities::cells::TakeCell;
 use kernel::{debug, ErrorCode};
 use lowrisc::otbn::Client;
-
-static mut OUTPUT: [u8; 1024] = [0; 1024];
-static mut TEMP: [u8; 4] = [0; 4];
 
 static MODULUS: [u8; 256] = [
     0xf9, 0x90, 0xc7, 0x94, 0xcf, 0x96, 0xd3, 0x12, 0x6f, 0x16, 0xa6, 0x50, 0x5d, 0xcb, 0xe9, 0x29,
@@ -47,14 +46,16 @@ static EXPECTING: [u8; 256] = [
 
 struct OtbnTestCallback {
     op_done: Cell<bool>,
+    output_buf: TakeCell<'static, [u8]>,
 }
 
 unsafe impl Sync for OtbnTestCallback {}
 
 impl<'a> OtbnTestCallback {
-    const fn new() -> Self {
+    fn new(output_buf: &'static mut [u8]) -> Self {
         OtbnTestCallback {
             op_done: Cell::new(false),
+            output_buf: TakeCell::new(output_buf),
         }
     }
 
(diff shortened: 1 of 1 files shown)


boards/opentitan/src/tests/otbn.rs
```diff
@@ -81,10 +82,26 @@ extern "C" {
     static _eapps: u8;
 }
 
+/// Static init an OtbnTestCallback, with
+/// respective buffers allocated
+/// Note: static_init!() returns an &'static mut reference.
+unsafe fn static_init_test_cb() -> &'static OtbnTestCallback {
+    let output_buf = static_init!([u8; 1024], [0; 1024]);
+    let temp_buf = static_init!([u8; 4], [0; 4]);
```
alistair23, line 90: You have identified a bug in the current code here.

The `TEMP` buffer is only ever used with `load_data()`. Previously `load_data()` took a `'static mut` as it was an async (like most calls in Tock). It was recently converted to just be a synchronous call as it's only used for the OTBN and it loads the data synchronously.

This means that `TEMP` and the new `temp_buf` don't need to be static. So you can remove this and just create a local array in `otbn_run_rsa_binary()` with

```rust
let temp_buf = [0; 4];
```

and just use that directly

boards/opentitan/src/tests/otbn.rs
```diff
@@ -81,10 +82,26 @@ extern "C" {
...
+        OtbnTestCallback::new(output_buf, temp_buf)
+    )
+}
+
 #[test_case]
 fn otbn_run_rsa_binary() {
     let perf = unsafe { PERIPHERALS.unwrap() };
     let otbn = &perf.otbn;
+    let cb = unsafe { static_init_test_cb() };
+    let output = cb.output_buf.take().unwrap();
```
alistair23, line 103: You need to save this again in `op_done()`

alistair23 (changes requested):

boards/opentitan/src/tests/otbn.rs
thulithwilfred: Thanks, I have made these changes.

boards/opentitan/src/tests/otbn.rs
```diff
@@ -66,13 +67,13 @@ impl<'a> OtbnTestCallback {
 impl<'a> Client<'a> for OtbnTestCallback {
     fn op_done(&'a self, result: Result<(), ErrorCode>, output: &'static mut [u8]) {
         self.op_done.set(true);
+        self.output_buf.replace(output);
         assert_eq!(result, Ok(()));
-        assert_eq!(output[0..256], EXPECTING);
+        // Use this copy of output for assertion
+        assert_eq!(self.output_buf.take().unwrap()[0..256], EXPECTING);
```
alistair23, line 73: Close, but this won't work.

You correctly replace the `self.output_buf` above, but then you take it here and it gets dropped.

You would instead want to do:

```rust
assert_eq!(output[0..256], EXPECTING);
self.output_buf.replace(output);
```

boards/opentitan/src/tests/otbn.rs
```diff
@@ -101,32 +114,30 @@ fn otbn_run_rsa_binary() {
...
         run_kernel_op(1000);
 
         // Set the RSA mode to encryption
         // The address is the offset of `mode` in the RSA elf
-        unsafe {
-            TEMP[0] = 1;
-            assert_eq!(otbn.load_data(0, &TEMP[0..4]), Ok(()));
-        }
+        temp[0] = 1;
+        assert_eq!(otbn.load_data(0, &temp[0..4]), Ok(()));
```
alistair23, line 132: You should be able to replace the `&temp[0..4]` with just `temp`

alistair23 (changes requested):

alistair23 (approved):

bradjc (approved):

PR #2928, opened 2021-12-21
Add userspace support for the Key/Value System

alistair23:
### Pull Request Overview

This exposes Tock's Key/value system to userspace.

### Testing Strategy

Running the libtock-c example app on OpenTitan.

### TODO or Help Wanted

- [x] Enforce permissions

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/components/src/kv_system.rs
+++ b/boards/components/src/kv_system.rs
@@ -1,0 +1,194 @@
+//! Component for KV System Driver.
+//!
+//! Usage
+//! -----
+//! ```rust
+//!    let mux_kv = components::kv_system::KVStoreMuxComponent::new(tickv).finalize(
+//!        components::kv_store_mux_component_helper!(
+//!            capsules::tickv::TicKVStore<
+//!                capsules::virtual_flash::FlashUser<lowrisc::flash_ctrl::FlashCtrl>,
+//!            >,
+//!            capsules::tickv::TicKVKeyType,
+//!        ),
+//!    );
+//!
+//!    let kv_store = components::kv_system::KVStoreComponent::new(mux_kv).finalize(
+//!        components::kv_store_component_helper!(
+//!            capsules::tickv::TicKVStore<
+//!                capsules::virtual_flash::FlashUser<lowrisc::flash_ctrl::FlashCtrl>,
+//!            >,
+//!            capsules::tickv::TicKVKeyType,
+//!        ),
+//!    );
+//!
+//!    let kv_driver_data_buf = static_init!([u8; 32], [0; 32]);
+//!    let kv_driver_dest_buf = static_init!(capsules::tickv::TicKVKeyType, [0; 8]);
+//!
+//!    let kv_driver = components::kv_system::KVDriverComponent::new(
+//!        kv_store,
+//!        board_kernel,
+//!        capsules::kv_driver::DRIVER_NUM,
+//!        kv_driver_data_buf,
+//!        kv_driver_dest_buf,
+//!    )
+//!    .finalize(components::kv_driver_component_helper!(
+//!        // capsules::kv_store::KVStore<
+//!        capsules::tickv::TicKVStore<
+//!            capsules::virtual_flash::FlashUser<lowrisc::flash_ctrl::FlashCtrl>,
+//!        >,
+//!        capsules::tickv::TicKVKeyType,
+//!    ));
+//! ```
+
+use capsules::kv_driver::KVSystemDriver;
+use capsules::kv_store::{KVStore, MuxKVStore};
+use core::mem::MaybeUninit;
+use kernel::capabilities;
+use kernel::component::Component;
+use kernel::create_capability;
+use kernel::hil::kv_system::{KVSystem, KeyType};
+use kernel::static_init_half;
+
+// Setup static space for the objects.
+#[macro_export]
+macro_rules! kv_store_mux_component_helper {
+    ($K:ty, $T:ty $(,)?) => {{
+        use capsules::kv_store::MuxKVStore;
+        use core::mem::MaybeUninit;
+        static mut BUF1: MaybeUninit<MuxKVStore<'static, $K, $T>> = MaybeUninit::uninit();
+        &mut BUF1
+    };};
+}
+
+pub struct KVStoreMuxComponent<
+    K: 'static + KVSystem<'static> + KVSystem<'static, K = T>,
+    T: 'static + KeyType,
+> {
+    kv: &'static K,
+}
+
+impl<'a, K: 'static + KVSystem<'static> + KVSystem<'static, K = T>, T: 'static + KeyType>
+    KVStoreMuxComponent<K, T>
+{
+    pub fn new(kv: &'static K) -> KVStoreMuxComponent<K, T> {
+        Self { kv }
+    }
+}
+
+impl<K: 'static + KVSystem<'static> + KVSystem<'static, K = T>, T: 'static + KeyType> Component
+    for KVStoreMuxComponent<K, T>
+{
+    type StaticInput = &'static mut MaybeUninit<MuxKVStore<'static, K, T>>;
+    type Output = &'static MuxKVStore<'static, K, T>;
+
+    unsafe fn finalize(self, s: Self::StaticInput) -> Self::Output {
+        static_init_half!(s, MuxKVStore<'static, K, T>, MuxKVStore::new(self.kv))
+    }
+}
+
+// Setup static space for the objects.
+#[macro_export]
+macro_rules! kv_store_component_helper {
+    ($K:ty, $T:ty $(,)?) => {{
+        use capsules::kv_store::KVStore;
+        use core::mem::MaybeUninit;
+        static mut BUF1: MaybeUninit<KVStore<'static, $K, $T>> = MaybeUninit::uninit();
+        &mut BUF1
+    };};
+}
+
+pub struct KVStoreComponent<K: 'static + KVSystem<'static, K = T>, T: 'static + KeyType> {
+    kv_store: &'static MuxKVStore<'static, K, T>,
+    key_buf: &'static mut T,
+    header_buf: &'static mut [u8; 9],
+}
+
+impl<K: 'static + KVSystem<'static, K = T>, T: 'static + KeyType> KVStoreComponent<K, T> {
+    pub fn new(
+        kv_store: &'static MuxKVStore<'static, K, T>,
+        key_buf: &'static mut T,
+        header_buf: &'static mut [u8; 9],
+    ) -> Self {
+        Self {
+            kv_store,
+            key_buf,
+            header_buf,
+        }
+    }
+}
+
+impl<K: 'static + KVSystem<'static, K = T>, T: 'static + KeyType> Component
+    for KVStoreComponent<K, T>
+{
+    type StaticInput = &'static mut MaybeUninit<KVStore<'static, K, T>>;
+    type Output = &'static KVStore<'static, K, T>;
+
+    unsafe fn finalize(self, static_buffer: Self::StaticInput) -> Self::Output {
+        static_init_half!(
+            static_buffer,
+            KVStore<'static, K, T>,
+            KVStore::new(self.kv_store, self.key_buf, self.header_buf)
+        )
+    }
+}
+
+// Setup static space for the objects.
+#[macro_export]
+macro_rules! kv_driver_component_helper {
+    ($K:ty, $T:ty $(,)?) => {{
+        use capsules::kv_driver::KVSystemDriver;
+        use core::mem::MaybeUninit;
+        static mut BUF1: MaybeUninit<KVSystemDriver<'static, $K, $T>> = MaybeUninit::uninit();
+        &mut BUF1
+    };};
+}
+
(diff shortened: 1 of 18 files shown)


capsules/src/kv_store.rs
```diff
@@ -0,0 +1,426 @@
...
+        &self,
+        unhashed_key: &'static mut [u8],
+        value: &'static mut [u8],
+    ) -> Result<(), (&'static mut [u8], &'static mut [u8], Result<(), ErrorCode>)> {
+        if self.mux_kv.operation.get() == Operation::None {
+            self.mux_kv.operation.set(Operation::Get);
+            if let Err((unhashed_key, hashed_key, e)) = self
+                .mux_kv
+                .kv
+                .generate_key(unhashed_key, self.hashed_key.take().unwrap())
```
bradjc, line 95: There are many `unwrap()`s in this file. Aside from this not being ideal from a "setting an example" perspective, doesn't this add many panics and increase the code size?

capsules/src/kv_driver.rs
```diff
@@ -0,0 +1,453 @@
+//! KV System
+//!
```
bradjc, line 2: Update.

Also, if you want you can put the KV code in its own folder. I don't care either way, up to you.

capsules/src/kv_driver.rs
```diff
@@ -0,0 +1,453 @@
...
+        result: Result<(), ErrorCode>,
+        key: &'static mut [u8],
+        ret_buf: &'static mut [u8],
+    ) {
+        self.data_buffer.replace(key);
+        self.dest_buffer.replace(ret_buf);
+
+        self.appid.map(move |id| {
+            self.apps.enter(*id, move |app, upcalls| {
+                if app.op.get().unwrap() == UserSpaceOp::Get {
```
bradjc, line 250: `.map()` would work here, no?

capsules/src/kv_driver.rs
```diff
@@ -0,0 +1,453 @@
...
+                        }
+                    }
+
+                    Ok(())
+                })
+                .unwrap_or_else(|err| Err(err.into()))
+        })
+    }
+
+    fn check_queue(&self) {
```
bradjc, line 213: This seems out of place here, but, I feel like you've come this far, maybe it's worth doing the userspace virtualization too?

libraries/tock-tbf/src/types.rs
```diff
@@ -702,6 +704,66 @@ impl TbfHeader {
         }
     }
 
+    /// Get the process `write_id`.
+    /// Returns `None` if a `write_id` is not included.
+    pub fn get_write_id(&self) -> Option<u32> {
```
bradjc, line 709: Are these used?

Also, perhaps we could namespace these somehow? `get_write_id` is a little vague.

alistair23:
Just pushed an update!

boards/components/src/kv_system.rs
```diff
@@ -0,0 +1,196 @@
...
+    type StaticInput = &'static mut MaybeUninit<KVStore<'static, K, T>>;
+    type Output = &'static KVStore<'static, K, T>;
+
+    unsafe fn finalize(self, static_buffer: Self::StaticInput) -> Self::Output {
+        let driver = static_init_half!(
+            static_buffer,
+            KVStore<'static, K, T>,
+            KVStore::new(self.kv_store, self.key_buf, self.header_buf)
+        );
+        // self.kv.set_client(driver);
```
hudson-ayers, line 132: Looks like this should be deleted?

boards/components/src/kv_system.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+impl<K: 'static + KVSystem<'static, K = T>, T: 'static + KeyType> Component
+    for KVStoreComponent<K, T>
+{
+    type StaticInput = &'static mut MaybeUninit<KVStore<'static, K, T>>;
+    type Output = &'static KVStore<'static, K, T>;
+
+    unsafe fn finalize(self, static_buffer: Self::StaticInput) -> Self::Output {
+        let driver = static_init_half!(
```
hudson-ayers, line 127: Is `driver` the appropriate name here? should this just be `kv_store`?

boards/components/src/kv_system.rs
```diff
@@ -0,0 +1,196 @@
...
+            static_buffer,
+            KVSystemDriver<'static, K, T>,
+            KVSystemDriver::new(
+                self.kv,
+                self.data_buffer,
+                self.dest_buffer,
+                self.board_kernel.create_grant(self.driver_num, &grant_cap)
+            )
+        );
+        // self.kv.set_client(driver);
```
hudson-ayers, line 193: why is this commented out? It seems like the best place to set the client?

boards/opentitan/src/main.rs
```diff
@@ -482,6 +493,49 @@ unsafe fn setup() -> (
...
+    let kv_store =
+        components::kv_system::KVStoreComponent::new(mux_kv, kv_store_key_buf, header_buf)
+            .finalize(components::kv_store_component_helper!(
+                capsules::tickv::TicKVStore<
+                    capsules::virtual_flash::FlashUser<lowrisc::flash_ctrl::FlashCtrl>,
+                    capsules::sip_hash::SipHasher24<'static>,
+                >,
+                capsules::tickv::TicKVKeyType,
+            ));
+    tickv.set_client(kv_store);
```
hudson-ayers, line 518: It feels like this client setting belongs in the components if possible

boards/opentitan/src/main.rs
```diff
@@ -482,6 +493,49 @@ unsafe fn setup() -> (
...
+        kv_driver_dest_buf,
+    )
+    .finalize(components::kv_driver_component_helper!(
+        capsules::tickv::TicKVStore<
+            capsules::virtual_flash::FlashUser<lowrisc::flash_ctrl::FlashCtrl>,
+            capsules::sip_hash::SipHasher24<'static>,
+        >,
+        capsules::tickv::TicKVKeyType,
+    ));
+    kv_store.set_client(kv_driver);
```
hudson-ayers, line 537: ditto for this client setting

libraries/tickv/src/async_ops.rs
```diff
@@ -555,9 +568,10 @@ mod store_flast_ctrl {
             _ => unreachable!(),
         }
 
-        let ret = tickv.append_key(get_hashed_key(b"TWO"), &VALUE);
+        #[allow(unsafe_code)]
```
hudson-ayers, line 571: Could we have a `mod test` (within this file) to encapsulate all of these test functions, and then mark only that test module with `#[allow(unsafe_code)]`? That makes it much clearer that only the test code is allowed to use unsafe.

capsules/src/kv_store.rs
```diff
@@ -0,0 +1,626 @@
...
+
+        self.mux_kv.do_next_op();
+    }
+
+    fn invalidate_key_complete(&self, result: Result<(), ErrorCode>, key: &'static mut T) {
+        self.hashed_key.replace(key);
+
+        self.mux_kv.operation.map(|op| match op {
+            Operation::Set | Operation::Get => {
+                unreachable!();
```
hudson-ayers, line 517: I advocate for just using a debug or nothing in these sort of instances, rather than embedding a panicking path in the binary which takes more size than a debug!() call and would crash the system if it was ever hit.

capsules/src/kv_store.rs
```diff
@@ -0,0 +1,626 @@
...
+        result: Result<(), ErrorCode>,
+        key: &'static mut T,
+        value: &'static mut [u8],
+    ) {
+        self.hashed_key.replace(key);
+        self.value.replace(value);
+
+        self.mux_kv.operation.map(|op| match op {
+            Operation::Get | Operation::Delete => {
+                unreachable!();
```
hudson-ayers, line 396: same here

capsules/src/kv_store.rs
```diff
@@ -0,0 +1,626 @@
...
+        &self,
+        result: Result<(), ErrorCode>,
+        key: &'static mut T,
+        ret_buf: &'static mut [u8],
+    ) {
+        self.hashed_key.replace(key);
+
+        self.mux_kv.operation.map(|op| match op {
+            Operation::Set => {
+                unreachable!();
```
hudson-ayers, line 423: same here

capsules/src/kv_driver.rs
```diff
@@ -0,0 +1,467 @@
...
+    K: kv_system::KVSystem<'a> + kv_system::KVSystem<'a, K = T>,
+    T: 'static + kv_system::KeyType,
+> {
+    kv: &'a KVStore<'a, K, T>,
+
+    active: Cell<bool>,
+
+    apps: Grant<
+        App,
+        UpcallCount<1>,
```
hudson-ayers, line 46: I am a fan of the ro_allow::COUNT approach, could we do that for the upcall as well here for consistency?

kernel/src/process.rs
```diff
@@ -400,6 +435,26 @@ pub trait Process {
...
+    fn num_read_ids(&self) -> Option<usize>;
+
+    /// Get the `access_ids`.
+    /// Returns `None` if a `access_ids` is not included.
+    fn get_access_ids(&self) -> Option<[u32; 8]>;
+
+    /// Get the number of `access_ids`.
+    /// Returns `None` if a `access_ids` is not included.
+    fn num_access_ids(&self) -> Option<usize>;
+
```
hudson-ayers, line 457: All boards pay for the code size of this support, even if they do not use the KV system. Looks like it is about 250 bytes per board, which is of course not a ton. I think long-term it is important for Tock to find ways to expose TBF header extensions to capsules in a configurable manner, but won't ask to block this PR on that support

boards/opentitan/src/main.rs
alistair23: We currently don't pass `tickv` to the component as we don't need it. I have left this here, but let me know if you feel strongly and I can move it into the component.

kernel/src/process.rs
alistair23: Hopefully long term we can eliminate virtual functions with https://github.com/rust-lang/rust/issues/68262

hudson-ayers (approved):

thulithwilfred (dismissed):

hudson-ayers:
I marked this last-call, but pinging in case anyone else wants to do a final review

alistair23:
Just pushed an update to try and address issues raised by @bradjc in the last OT call

kernel/src/process.rs
```diff
@@ -96,6 +96,19 @@ impl fmt::Debug for ProcessId {
     }
 }
 
+/// The read permissions the process has
+/// If the permissions exist this will be: `Some((num_ids, ids))`
+/// Where num_ids is the number of `read_ids` and `ids` is an array of the
+/// `read_ids`.
+pub type ReadPermissions = Option<(usize, [u32; 8])>;
```
bradjc, line 103: This still puts `8` in the public capsule-facing process API. I'm hesitant to go that way.

Would you be ok with a `persistent_storage_has_write_access(storage_id: u32) -> bool`-style API?

alistair23: That works fine for checking the permissions, but we also need to store the permissions in flash when we write data. So we need a way to get them to then write them.

Also we check permissions in `KVStore` as this allows either applications or the kernel to store data. At this point though we don't have a `Process` as the kernel might be using it. So there isn't really a way we could check against a `Process` here. Instead `KVDriver` get the app permissions and passes them down. This way the kernel can directly use `KVStore` if it wants

bradjc:
Take a look at the commits I added. The major change is adding `storage_permissions.rs` and the `StoragePermissions` object which encapsulates the storage permissions. I then used this to simplify the kv_store implementation.

kernel/src/process_standard.rs
```diff
@@ -410,24 +410,26 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
...
-    }
+    fn get_storage_permissions(&self) -> Option<storage_permissions::StoragePermissions> {
+        let (read_count, read_storage_ids) = self
+            .header
+            .get_persistent_acl_read_ids()
+            .unwrap_or((0, [0; 8]));
 
-    fn get_write_permissions(&self) -> WritePermissions {
-        let write_id = self.header.get_write_id().unwrap_or(0);
+        let (write_count, write_storage_ids) = self
```
alistair23, line 419: `access_count`, `access_storage_ids`

kernel/src/process.rs
```diff
@@ -429,15 +409,8 @@ pub trait Process {
...
-    /// This describes what the process is allowed to read
-    /// Returns `None` if `read_ids` are not included.
-    fn get_read_permissions(&self) -> ReadPermissions;
-
-    /// Get the storage write permissions for the process
-    /// This describes what the process is allowed to delete or modify
-    /// Returns `None` if `access_ids` are not included.
-    fn get_write_permissions(&self) -> WritePermissions;
+    /// Get the storage permissions for the process.
+    fn get_storage_permissions(&self) -> Option<storage_permissions::StoragePermissions>;
```
alistair23, line 413: Returns `None` if the process has no storage permissions.

alistair23:
Looks good

PR #2929, opened 2021-12-22
boards/opentitan: tests: siphash: remove global static buffers

thulithwilfred:

### Pull Request Overview
Removes global static buffers from the sip_hash test

### Testing Strategy

`make test`

`make ci-all`

### TODO or Help Wanted
N/A

### Documentation Updated

### Formatting
- [x] Ran `make fmt`.

### DIFF
--- a/boards/opentitan/src/tests/sip_hash.rs
+++ b/boards/opentitan/src/tests/sip_hash.rs
@@ -2,59 +2,58 @@ use crate::tests::run_kernel_op;
 use crate::SIPHASH;
 use core::cell::Cell;
 use kernel::hil::hasher::{self, Hasher};
+use kernel::static_init;
+use kernel::utilities::cells::TakeCell;
 use kernel::utilities::leasable_buffer::LeasableBuffer;
 use kernel::{debug, ErrorCode};
-
-static mut INPUT: [[u8; 8]; 20] = [
-    [0x31, 0x0e, 0x0e, 0xdd, 0x47, 0xdb, 0x6f, 0x72],
-    [0xfd, 0x67, 0xdc, 0x93, 0xc5, 0x39, 0xf8, 0x74],
-    [0x5a, 0x4f, 0xa9, 0xd9, 0x09, 0x80, 0x6c, 0x0d],
-    [0x2d, 0x7e, 0xfb, 0xd7, 0x96, 0x66, 0x67, 0x85],
-    [0xb7, 0x87, 0x71, 0x27, 0xe0, 0x94, 0x27, 0xcf],
-    [0x8d, 0xa6, 0x99, 0xcd, 0x64, 0x55, 0x76, 0x18],
-    [0xce, 0xe3, 0xfe, 0x58, 0x6e, 0x46, 0xc9, 0xcb],
-    [0x37, 0xd1, 0x01, 0x8b, 0xf5, 0x00, 0x02, 0xab],
-    [0x62, 0x24, 0x93, 0x9a, 0x79, 0xf5, 0xf5, 0x93],
-    [0xb0, 0xe4, 0xa9, 0x0b, 0xdf, 0x82, 0x00, 0x9e],
-    [0xf3, 0xb9, 0xdd, 0x94, 0xc5, 0xbb, 0x5d, 0x7a],
-    [0xa7, 0xad, 0x6b, 0x22, 0x46, 0x2f, 0xb3, 0xf4],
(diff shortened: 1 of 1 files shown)


thulithwilfred:
@alistair23

boards/opentitan/src/tests/sip_hash.rs
```diff
@@ -66,35 +47,65 @@ impl<'a> hasher::Client<'a, 8> for SipHashTestCallback {
         assert_eq!(ret, 0x9ed5975598371f51);
 
         self.hash_done.set(true);
+        self.output_buf.replace(digest);
```
alistair23, line 50: You should replace input in `add_data_done()` as well

alistair23 (changes requested):

boards/opentitan/src/tests/sip_hash.rs
thulithwilfred: How would you suggest this to be implemented? The cb here gets triggered as each slice is being added by the loop in the test? and since _data in the cb is a particular slice, would we want to save it into a separate buffer or track the cb count/index?

alistair23: Good question.

Could we track calls to `add_data_done()` and just replace that array?

thulithwilfred: In the new commit, tried making a new buffer with each callback to hold the data, and at the end replacing the old one with it. However, this requires 2 static buffers to be allocated per instance (might be very inefficient). 

Thoughts on this?

boards/opentitan/src/tests/sip_hash.rs
```diff
@@ -52,9 +41,28 @@ impl<'a> SipHashTestCallback {
...
+        let temp = self.input_save.take().unwrap();
+
+        // Stay within valid ranges
+        assert_eq!(self.cb_count.get() < 20, true);
+        for i in 0..data.len() {
+            assert_eq!(i < 8, true);
+            temp[self.cb_count.get() as usize][i] = data[i];
+        }
+
+        self.input_save.replace(temp);
```
alistair23, line 57: You aren't replacing `data` here.

Does something like this work?

```rust
SipHashTestCallback {
...
    input: [TakeCell<'static, [u8; 8]>; 20],
...
}

fn add_data_done(&'a self, result: Result<(), ErrorCode>, data: &'static mut [u8]) {
...
    self.input[i].replace(data);
...
}
```
temp[]

thulithwilfred: Where, `input` here acts as a separate buffer to collect the callback slices? But doesn't replace `input_buffer` (that starts off with all the input data)  once all the cbs are complete?

boards/opentitan/src/tests/sip_hash.rs
```diff
@@ -63,40 +62,152 @@ impl<'a> hasher::Client<'a, 8> for SipHashTestCallback {
...
+
+    assert_eq!(sip_hasher.run(cb.output_buf.take().unwrap()), Ok(()));
+    run_kernel_op(100);
+    assert_eq!(cb.hash_done.get(), true);
 
     run_kernel_op(100);
     debug!("    [ok]");
     run_kernel_op(100);
-}
+}
```
alistair23, line 213: You should have a new line at the end of files

boards/opentitan/src/tests/sip_hash.rs
```diff
@@ -63,40 +62,152 @@ impl<'a> hasher::Client<'a, 8> for SipHashTestCallback {
         assert_eq!(result, Ok(()));
         // Value calculated from:
         //    https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=08a45842acb9dc1abf9dbc303b5eaa50
-        assert_eq!(ret, 0x9ed5975598371f51);
+        if self.run_count.get() == 0 {
+            assert_eq!(ret, 0x9ed5975598371f51);
+        } else if self.run_count.get() == 1 {
+            assert_eq!(ret, 0xB5326A7D96F8A5B7);
+        } else {
+            panic!();
```
alistair23, line 70: Probably worth adding a message here

thulithwilfred: Do you mean like a debug message output to say that the assertion was ok for each of the runs?

alistair23: `panic!("Unsupported number of callbacks");`

alistair23 (approved):

hudson-ayers (approved):

PR #2933, opened 2022-01-05
Have Grant::each() take FnMut instead of Fn

hudson-ayers:
### Pull Request Overview

This restriction is a holdover from when all of the Grant functions took Fn closures, but there is no longer any reason to limit the functionality of the closures passed to this method.

### Testing Strategy

This pull request was tested by compiling the following closure:
```rust
        let mut next = Ticks64::max_value();
        self.apps.each(|_, app, kernel_data| {
            if let Some(deadline) = app.expiration {
                // We don't have to worry about rollover comparisons with 64-bit ms timer
                if deadline <= now {
                    // Schedule callback and remove this expiration
                    let _ = kernel_data.schedule_upcall(Sub::Fired as usize, (0, 0, 0));
                    app.expiration = None;
                } else {
                    next = next.min(deadline);
                }
            }
        });
```
which only compiles if each() accepts FnMut closures.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -1612,9 +1612,9 @@ impl<T: Default, Upcalls: UpcallSize, AllowROs: AllowRoSize, AllowRWs: AllowRwSi
     ///
     /// Calling this function when an `ProcessGrant` for a process is currently
     /// entered will result in a panic.
-    pub fn each<F>(&self, fun: F)
+    pub fn each<F>(&self, mut fun: F)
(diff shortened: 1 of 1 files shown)


alistair23 (approved):

jettr (approved):
Thank you!

lschuermann (approved):

PR #2935, opened 2022-01-07
Add embedded data analysis tool

paulmure:
### Pull Request Overview

This pull request adds an embedded data analysis tool, which allows users to inspect the distribution of embedded data in tock binaries.

Prerequisite:
* The binary must be compiled for the RISC-V architecture.
* Turn off LTO. To do so, set `lto` to `false` in the top-level `Cargo.toml` file.

To run the tool, in the Tock home directory envoke:

    $ tools/embedded_data_analyzer.py -e <elf_path> -n <name> -o <output_directory>

where 
* `elf_path` is the relative path of the compile elf file
* `name` is a name for the binary to be displayed in the HTML page
* `output_directory` is where the user wishes the output to be stored

You can also run the script with `-h` flag for help on what the arguments are.

Upon successful completion, the tool will print the path of the html page where the user can view the ouput in a browser. The page should contain a list of all the functions and information about their embedded data size. The user can click on each function to see the exact content in each embedded object for that function.

### Testing Strategy

This pull request was tested by inspecting the embedded data elements in outputted HTML against disassembly.

### TODO or Help Wanted

This pull request still needs to 
* Handle unidentified ASCII characters better
* Present embedded data content in a more visually pleasing way

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/tools/embedded_data_analyzer.py
+++ b/tools/embedded_data_analyzer.py
@@ -1,0 +1,320 @@
+#! /usr/bin/env python3
+
+from audioop import add
+from embedded_data_visualizer import create_html_file
+from print_tock_memory_usage import parse_mangled_name
+import argparse
+from collections import defaultdict, namedtuple
+import os
+import sys
+import re
+HELP_STRING = '''
+Tock embedded data analysis tool
+
+Generates a report of a Tock binary's embedded data usage in the form of a static HTML webpage.
+
+Prerequisite:
+    - The binary must be compiled for the RISC-V architecture.
+    - Turn off LTO. To do so, set lto to false in the top-level Cargo.toml file.
+
+To run the tool, in the Tock home directory invoke:
+
+    $ tools/embedded_data_analyzer.py -e <elf_path> -n <name> -o <output_directory> [-d <objdump>]
+
+arguments:
+    elf_path         -- the relative path of the compiled elf file
+    name             -- a name for the binary to be displayed in the HTML page
+    output_directory -- where the user wishes the output to be stored
+    objdump          -- (optional) path of the riscv objdump binary to use
+
+You can also run the script with -h flag for help on what the arguments are.
+'''
+
+
+def get_args():
+    # To allow for a custom help message, the default help option is turned off,
+    # this means that required arguments must be checked manually
+    ap = argparse.ArgumentParser(add_help=False)
+    ap.add_argument('-e', '--elf_path', required=False)
+    ap.add_argument('-o', '--output_directory', required=False)
+    ap.add_argument('-n', '--name', required=False)
+    ap.add_argument('-h', '--help', required=False,
+                    action='store_true', default=False)
+    ap.add_argument('-d', '--objdump', required=False)
+    return vars(ap.parse_args())
+
+
+def find_objdump():
+    objdump = os.popen('which riscv32-unknown-elf-objdump').read().strip()
+    if objdump != '':
+        return objdump
+
+    objdump = os.popen('which riscv64-unknown-elf-objdump').read().strip()
+    if objdump != '':
+        return objdump
+
+    objdump = os.popen('which riscv32-none-elf-objdump').read().strip()
+    if objdump != '':
+        return objdump
+
+    objdump = os.popen('which riscv64-none-elf-objdump').read().strip()
+    if objdump != '':
+        return objdump
+
+    critical_error(
+        'Objdump not found, please specify a riscv objdump path manually with -d or --objdump.')
+
+
+FunctionInfo = namedtuple(
+    'FunctionInfo', 'embedded_data_size_actual embedded_data_size_estimated embedded_data_count addresses')
+SymbolInfo = namedtuple('SymbolInfo', 'actual_size estimated_size name')
+
+
+def get_srodata_address(disasm):
+    pattern = '<_srodata>:'
+    for i in range(len(disasm)):
+        match = re.findall(pattern, disasm[i])
+        if match:
+            return disasm[i].split()[0], i
+    critical_error('Invalid binary, cannot find <_srodata> section')
+
+
+def map_adress_to_data(disasm):
+    _, start_index = get_srodata_address(disasm)
+    sro_address, sro_end = get_sro_range(disasm)
+    address_pattern = r'^\s*([0-9a-f]+):'
+    data_pattern = r'[ |\t][0-9a-f]{4}'
+
+    sro_data = ""
+    for i in range(start_index, len(disasm)):
+        match = re.search(address_pattern, disasm[i])
+        if match:
+            address = int(match.group(1), 16)
+            if (sro_address <= address < sro_end):
+                data = re.findall(data_pattern, disasm[i])
+                if data:
+                    # remove the \t character from first byte
+                    data[0] = re.sub('\s+', '', data[0])
+                    for j in range(len(data)):
+                        data[j] = data[j].strip()
+                        # seperate the bytes
+                        sro_data += bytes.fromhex(data[j][2:]
+                                                  ).decode("utf-8", errors='replace')
+                        sro_data += bytes.fromhex(data[j][0:2]
+                                                  ).decode("utf-8", errors='replace')
+
+    return sro_data
+
+
+def get_sro_range(disasm):
+    data_pattern = r'[ |\t][0-9a-f]{4}'
+    address_pattern = r'^[0-9a-f]{1,16}'
+    sro_string, start_index = get_srodata_address(disasm)
+    sro_address = int(sro_string, 16)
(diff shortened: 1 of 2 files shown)


alistair23:
Can you add details on how this should be run? Can you also include details on what the output should look like and what it tells you?

phil-levis:
@paulmure Can you add information into comments in the script itself and add a usage (with a --help)? Someone won't know to look in a particular PR for how to use a tool.

hudson-ayers:
Can you squash all of the commits into one? We try to avoid including merge commits other than those created by bors (our CI mergebot) on the master branch.

paulmure:
> Can you squash all of the commits into one? We try to avoid including merge commits other than those created by bors (our CI mergebot) on the master branch.

The commit history got really messy. I think it's a lot easier for me to just close this pull request and redo it with a fresh fork.

paulmure:
> @paulmure Can you add information into comments in the script itself and add a usage (with a --help)? Someone won't know to look in a particular PR for how to use a tool.

I updated the —help flag to include a custom message about the usage of the tool. The help message is also included at the top of the script.

phil-levis (commented):
This looks good to me, but I'd like to have an independent set of eyes take a look.

tools/embedded_data_analyzer.py
```diff
@@ -0,0 +1,288 @@
+#!/usr/bin/python3
+
+HELP_STRING = '''
+Tock embedded data analysis tool
+
+Generates a report of a Tock binary's embedded data usage in the form of a static HTML webpage.
+
+Prerequisite:
+    - The binary must be compiled for the RISC-V architecture.
+    - Turn off LTO. To do so, set lto to false in the top-level Cargo.toml file.
```
alistair23, line 10: The debug build does this already, can't we just use that?

tools/embedded_data_analyzer.py
```diff
@@ -0,0 +1,288 @@
...
+
+To run the tool, in the Tock home directory invoke:
+
+    $ tools/embedded_data_analyzer.py -e <elf_path> -n <name> -o <output_directory>
```
alistair23, line 14: Could we add this as a Makefile target?

tools/embedded_data_analyzer.py
```diff
@@ -0,0 +1,288 @@
...
+    sys.exit()
+
+
+def check_architecture(elf_path):
+    readelf = os.popen(f'readelf -h {elf_path} | grep Machine').read()
+    if 'RISC-V' not in readelf:
+        critical_error('Error: elf file must be compiled for RISC-V')
+
+
+def check_rto():
```
alistair23, line 248: `check_lto()`?

tools/embedded_data_analyzer.py
```diff
@@ -0,0 +1,288 @@
...
+import re
+import sys
+import os
+from collections import defaultdict, namedtuple
+import argparse
+
+from print_tock_memory_usage import parse_mangled_name
+from embedded_data_visualizer import create_html_file
+
+OBJDUMP = 'riscv64-unknown-elf-objdump'
```
alistair23, line 34: Unfortunately every distro has a different name for this. Can we add a fallback checker like we have in [libtock-c](https://github.com/tock/libtock-c/blob/master/Configuration.mk#L124)

alistair23 (changes requested):

tools/embedded_data_analyzer.py
paulmure: The user might want to use this tool on the release build to see what embedded data makes it into the final binary.

tools/embedded_data_analyzer.py
paulmure: fixed

tools/embedded_data_analyzer.py
alistair23: Agreed, but once you disable LTO it's closer to a debug binary then a release binary anyway. Using a debug binary means it could work with make, so you can just do: `make html-analysis` and get results

hudson-ayers: Yeah, the "debug binary" is equivalent to a release binary with LTO disabled

bradjc (approved):

tools/embedded_data_analyzer.py
```diff
@@ -0,0 +1,288 @@
+#!/usr/bin/python3
```
lschuermann, line 1: Unfortunately, this won't work on every distribution. It's best to use the `python3` interpreter from the PATH environment:

```suggestion
#! /usr/bin/env python3
```

tools/embedded_data_analyzer.py
lschuermann: I second this. It's unfortunate this does not seem to work with `llvm-objdump`, but many distros do indeed choose a different toolchain tuple. The list in `libtock-c` seems like a good start, or just adding an argument with a default value for this.

lschuermann (changes requested):
This looks pretty cool.

Unfortunately when trying to run it, I get the following error:

```
> ./embedded_data_analyzer.py -e ../target/riscv32imc-unknown-none-elf/release/litex_sim.elf -n litex_sim -o ../tmp/
Traceback (most recent call last):
  File "/home/leons/develop/tock/tock/tools/./embedded_data_analyzer.py", line 288, in <module>
    main()
  File "/home/leons/develop/tock/tock/tools/./embedded_data_analyzer.py", line 277, in main
    symbols_dict = build_symbols_dict(symbols_table)
  File "/home/leons/develop/tock/tock/tools/./embedded_data_analyzer.py", line 159, in build_symbols_dict
    size = int(splitted_size[0], 16)
ValueError: invalid literal for int() with base 16: 'g_loc'
```
I didn't look into this in-depth yet, maybe it's just something isolated to my system or build. Perhaps you've already stumbled over this error and can tell me what I'm doing wrong? I did disable LTO in `Cargo.toml` and rebuilt my target.

tools/embedded_data_analyzer.py
paulmure: I don't know which build the user might want to run the tool on, so I just printed a soft warning when LTO is turned on for either build. It will keep running and just give a warning to the user.

tools/embedded_data_analyzer.py
paulmure: Thanks for the note, this changed has been added.

tools/embedded_data_analyzer.py
paulmure: Yes, I added the functionality to search the system path for known distributions of riscv objdump. As well as a fallback option for manually specifying which objdump to use.

paulmure:
> This looks pretty cool.
> 
> Unfortunately when trying to run it, I get the following error:
> 
> ```
> > ./embedded_data_analyzer.py -e ../target/riscv32imc-unknown-none-elf/release/litex_sim.elf -n litex_sim -o ../tmp/
> Traceback (most recent call last):
>   File "/home/leons/develop/tock/tock/tools/./embedded_data_analyzer.py", line 288, in <module>
>     main()
>   File "/home/leons/develop/tock/tock/tools/./embedded_data_analyzer.py", line 277, in main
>     symbols_dict = build_symbols_dict(symbols_table)
>   File "/home/leons/develop/tock/tock/tools/./embedded_data_analyzer.py", line 159, in build_symbols_dict
>     size = int(splitted_size[0], 16)
> ValueError: invalid literal for int() with base 16: 'g_loc'
> ```
> 
> I didn't look into this in-depth yet, maybe it's just something isolated to my system or build. Perhaps you've already stumbled over this error and can tell me what I'm doing wrong? I did disable LTO in `Cargo.toml` and rebuilt my target. FWIW, [here's](https://gist.github.com/lschuermann/b75e0551c36cb1f6802b9104c39b85ec) a JSON-dump of the erroring method invocation's argument.

Hi Leon, this seems like a bug in my code. I added a check to skip any lines in the symbol table that contains the word ".debug", which would imply a debugging information that is not relevant to the analysis. Would you mind checking again to see if it works?

tools/embedded_data_analyzer.py
paulmure: Sounds like this should be a separate PR that can be done later.

lschuermann (approved):
Just catching up on things... Yeah, this looks good, tested to work on a bunch of boards and this contains my fix for the issue with addresses with leading zeroes.

lschuermann:
Given @bradjc already approved this previously and it has been waiting on me for a while now, I'd label this as last-call.

hudson-ayers (approved):
bors r+

alistair23 (approved):
I would like to see this launched from Make, but I don't want to hold this up

hudson-ayers (approved):
bors r+

PR #2939, opened 2022-01-12
chips/earlgrey: Don't use MIP::MTIMER bit

alistair23:
### Pull Request Overview

The RISC-V specification provides the MIP:MTIMER bit to indicate when a
timer interrupt occurs. OpenTitan also provides a RVTIMERTIMEREXPIRED0_0
interrupt for the same functionality.

We should be only using one of the timer interrupt sources, currently we
report the MIP::MTIMER bit, but never act on it, instead handling the
interrupt generated from RVTIMERTIMEREXPIRED0_0.

This patch removes all of the MIP::MTIMER functionality and disables the
interrupt in MIE to ensure that we don't get double handle a timer
interrupt. We now only use the RVTIMERTIMEREXPIRED0_0 interrupt as
passed from the PLIC.

This issue was found by @thulithwilfred when he tried to use a delay in libtock-c
on the OpenTitan QEMU model.

### Testing Strategy

Running the test on OT FPGA hardware

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/src/main.rs
+++ b/boards/opentitan/src/main.rs
@@ -319,9 +319,9 @@ unsafe fn setup() -> (
     // Need to enable all interrupts for Tock Kernel
     chip.enable_plic_interrupts();
     // enable interrupts globally
-    csr::CSR
-        .mie
-        .modify(csr::mie::mie::msoft::SET + csr::mie::mie::mtimer::SET + csr::mie::mie::mext::SET);
+    csr::CSR.mie.modify(
+        csr::mie::mie::msoft::SET + csr::mie::mie::mtimer::CLEAR + csr::mie::mie::mext::SET,
+    );
     csr::CSR.mstatus.modify(csr::mstatus::mstatus::mie::SET);
 
     // Setup the console.
--- a/chips/earlgrey/src/chip.rs
+++ b/chips/earlgrey/src/chip.rs
@@ -5,7 +5,7 @@ use kernel;
 use kernel::dynamic_deferred_call::DynamicDeferredCall;
 use kernel::platform::chip::{Chip, InterruptService};
 use kernel::utilities::registers::interfaces::{ReadWriteable, Readable, Writeable};
-use rv32i::csr::{mcause, mie::mie, mip::mip, mtvec::mtvec, CSR};
+use rv32i::csr::{mcause, mie::mie, mtvec::mtvec, CSR};
 use rv32i::epmp::PMP;
 use rv32i::syscall::SysCall;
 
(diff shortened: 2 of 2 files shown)


chips/earlgrey/src/chip.rs
```diff
@@ -212,28 +212,25 @@ impl<'a, I: InterruptService<()> + 'a> kernel::platform::chip::Chip for EarlGrey
...
-            if !mip.matches_any(mip::mtimer::SET) && self.plic.get_saved_interrupts().is_none() {
+            if self.plic.get_saved_interrupts().is_none() {
                 break;
             }
         }
 
         // Re-enable all MIE interrupts that we care about. Since we looped
         // until we handled them all, we can re-enable all of them.
-        CSR.mie.modify(mie::mext::SET + mie::mtimer::SET);
+        CSR.mie.modify(mie::mext::SET + mie::mtimer::CLEAR);
```
hudson-ayers, line 228:
```suggestion
        CSR.mie.modify(mie::mext::SET);
```

Given that mtimer should already be cleared, isn't this better?

boards/opentitan/src/main.rs
```diff
@@ -319,9 +319,9 @@ unsafe fn setup() -> (
     // Need to enable all interrupts for Tock Kernel
     chip.enable_plic_interrupts();
     // enable interrupts globally
-    csr::CSR
-        .mie
-        .modify(csr::mie::mie::msoft::SET + csr::mie::mie::mtimer::SET + csr::mie::mie::mext::SET);
+    csr::CSR.mie.modify(
+        csr::mie::mie::msoft::SET + csr::mie::mie::mtimer::CLEAR + csr::mie::mie::mext::SET,
```
hudson-ayers, line 323: Is mtimer enabled by default? If not, it seems less confusing to just leave it out entirely, or at least add a note that we do not enable that interrupt because we instead get that signal via the PLIC

alistair23: I think we should explicitly clear this. It's possible that the bootROM enabled it, plus this makes it obvious for anyone reading the code that we disable the timer interrupt

chips/earlgrey/src/chip.rs
alistair23: It should already be cleared, that's true. I like this though as it makes it explicit that we aren't using the standard RISC-V timer interrupt. This is different to every other RISC-V chip so I think it's worth making it obvious here

hudson-ayers (approved):

bradjc (approved):

lschuermann (approved):

PR #2945, opened 2022-01-14
spi: fix issue when read buffer not present

### LINKED ISSUE #2910: SPI read length incorrect
Accorrding to the comment in [spi_controller.rs][0], we should be able to perform a write operation without providing a read buffer. The way the code is written in [spi_controller][0] and [spi_peripheral][1] does not do this. It should instead be something like the following:

```rust
let wlen = app.app_write.enter(|w| w.len()).unwrap_or(0);
let rlen = app.app_read.enter(|r| r.len()).unwrap_or(0);
let len = if rlen == 0 { wlen } else { wlen.min(rlen) };
```

I want to fix this after #2906 lands as to not give myself more merge conflicts (that is also how I noticed the issue in the code).

[0]:https://github.com/tock/tock/blob/master/capsules/src/spi_controller.rs#L223
[1]:https://github.com/tock/tock/blob/master/capsules/src/spi_peripheral.rs#L213

jettr:
### Pull Request Overview

If a read buffer was not passed in, then the write operation would not occur since the read length returned would always be 0 (not none). Then the min would be 0, and the operation would fail since len was 0.

This fixes #2910

### Documentation Updated

- [x] None

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/spi_controller.rs
+++ b/capsules/src/spi_controller.rs
@@ -199,15 +199,16 @@ impl<'a, S: SpiMasterDevice> SyscallDriver for Spi<'a, S> {
                     // 1) Write and read buffers present: len is min of lengths
                     // 2) Only write buffer present: len is len of write
                     // 3) No write buffer present: no operation
-                    let mut mlen = kernel_data
+                    let wlen = kernel_data
                         .get_readonly_processbuffer(ro_allow::WRITE)
                         .map_or(0, |write| write.len());
                     let rlen = kernel_data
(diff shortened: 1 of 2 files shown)


phil-levis (approved):
Looks good to me. The key point is that 1.x had the idea of "null" buffers, when a null pointer meant no buffer. The buffer was stored as an option, so the buffer could be a None. Now the kernel just uses zero-length buffers.

alistair23 (approved):

PR #2947, opened 2022-01-21
arch/rv32i: pmp/ePMP: Fixup PMP comparision

alistair23:
### Pull Request Overview

PMP addresses are not inclusive on the high end, that is
```
pmpaddr[i-i] <= y < pmpaddr[i]
```

So let's change the `overlaps()` logic to take that into account. This 
fixes PMP errors when two regions are right next to each other.

### Testing Strategy

OT FPGA

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/epmp.rs
+++ b/arch/rv32i/src/epmp.rs
@@ -306,6 +306,10 @@ impl PMPRegion {
         self.location
     }
 
+    /// Check if the PMP regions specified by `other_start` and `other_size`
+    /// overlaps with the current region.
+    /// Matching the RISC-V spec this checks pmpaddr[i-i] <= y < pmpaddr[i] for
+    /// TOR ranges.
     fn overlaps(&self, other_start: *const u8, other_size: usize) -> bool {
         let other_start = other_start as usize;
         let other_end = other_start + other_size;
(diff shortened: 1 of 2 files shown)


hudson-ayers (approved):
Pretty surprising we had never tested two directly adjacent PMP regions until now

arch/rv32i/src/epmp.rs
```diff
@@ -318,7 +318,7 @@ impl PMPRegion {
             (region_start, region_end)
         };
 
-        if region_start < other_end && other_start < region_end {
+        if region_start < (other_end - 4) && other_start < (region_end - 4) {
```
hudson-ayers, line 321: Actually, can we add a comment here stating what is in this PR description? Otherwise at a glance this just looks like a bug

arch/rv32i/src/pmp.rs
```diff
@@ -197,7 +197,7 @@ impl PMPRegion {
             (region_start, region_end)
         };
 
-        if region_start < other_end && other_start < region_end {
+        if region_start < (other_end - 4) && other_start < (region_end - 4) {
```
hudson-ayers, line 200: same here

hudson-ayers (approved):

bradjc (approved):

lschuermann (approved):

PR #2953, opened 2022-01-31
opentitan: Makefile: fixup test flash-layout

thulithwilfred:
### Pull Request Overview

This fixes a bug, where if an application is ran (i.e qemu-app) after
`make test`, the incorrect flash layout is loaded (in layout.ld). As a result
of `make test` copying the test layout over and this not being cleared
when an app is loaded, thus resulting in a hang after OT is initialized.

Is fixed by removing `layout.ld` after a test is run with no additional cost to
build time.

edit: Applied fix for nexysvideo Makefile

### Testing Strategy
`make test` ,  loading an application through qemu

### Documentation Updated

- [x ] Removed notice to run `make clean` between `make test` and `make app` (as this fixes the bug)

### Formatting

- [ x] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/earlgrey-cw310/Makefile
+++ b/boards/opentitan/earlgrey-cw310/Makefile
@@ -66,6 +66,8 @@ endif
 	$(Q)cp test_layout.ld $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/deps/layout.ld
 	$(Q)cp ../../kernel_layout.ld $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/
 	$(Q)RUSTFLAGS="$(RUSTC_FLAGS_TOCK)" $(CARGO) test $(CARGO_FLAGS_TOCK_NO_BUILD_STD) $(NO_RUN) --bin $(PLATFORM) --release
+# Test layout should be removed so that following apps (non-test) load the correct layout.
+	$(Q)rm $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/deps/layout.ld
 
 test-hardware:
 ifeq ($(OPENTITAN_TREE),)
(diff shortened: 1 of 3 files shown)


thulithwilfred:
@alistair23

boards/opentitan/earlgrey-cw310/Makefile
```diff
@@ -66,6 +66,8 @@ endif
 	$(Q)cp test_layout.ld $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/deps/layout.ld
 	$(Q)cp ../../kernel_layout.ld $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/
 	$(Q)RUSTFLAGS="$(RUSTC_FLAGS_TOCK)" $(CARGO) test $(CARGO_FLAGS_TOCK_NO_BUILD_STD) $(NO_RUN) --bin $(PLATFORM) --release
+# Test layout should be removed so that following apps (non-test) load the correct layout.
+	$(Q)rm $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/deps/layout.ld
```
alistair23, line 70: You will need to make the same changes to the nexysvideo Makefile

alistair23 (changes requested):
> This fixes a bug, where if an application is ran (i.e qemu-app) followed by
> make test, 

This ordering looks wrong, the bug appears when running an application after `make test`

boards/opentitan/earlgrey-cw310/Makefile
thulithwilfred: oops! good catch

thulithwilfred:
> > This fixes a bug, where if an application is ran (i.e qemu-app) followed by
> > make test,
> 
> This ordering looks wrong, the bug appears when running an application after `make test`



> 

oops! good catch, thanks!

alistair23 (approved):

bradjc (approved):

PR #2957, opened 2022-02-08
doc: Update Instructions: Building build.rs requires a host toolchain for linking

dcz-self:
### Pull Request Overview

This fixes build instructions. After installing in a container, it turned out some more packages were required. Those were added. When the host toolchain is missing, the following errors appear, and don't seem to disappear after the toolchain is installed:

```
  = note: rust-lld: error: unable to find library -lgcc_s
          rust-lld: error: unable to find library -lutil
          rust-lld: error: unable to find library -lrt
          rust-lld: error: unable to find library -lpthread
          rust-lld: error: unable to find library -lm
          rust-lld: error: unable to find library -ldl
          rust-lld: error: unable to find library -lc
          rust-lld: error: cannot find linker script layout.ld
```

### Testing Strategy

This pull request was tested by typing the commands manually.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`. irrelevant?

### DIFF
--- a/doc/Getting_Started.md
+++ b/doc/Getting_Started.md
@@ -30,8 +30,9 @@ developing Tock.
 
 1. [Rust](http://www.rust-lang.org/)
 2. [rustup](https://rustup.rs/) to install Rust (version >= 1.11.0)
-3. Command line utilities: make
-4. A supported board or QEMU configuration.
+3. Host toolchain (gcc, glibc)
+4. Command line utilities: make, find
+5. A supported board or QEMU configuration.
 
    If you are just starting to work with TockOS, you should look in
    the [`boards/` subdirectory](../boards/README.md) and choose one of
@@ -61,6 +62,7 @@ $ pip3 install --upgrade tockloader
 
 Ubuntu:
 ```
+$ apt install -y build-essential python3-pip curl
 $ curl https://sh.rustup.rs -sSf | sh
 $ pip3 install --upgrade tockloader --user
 $ grep -q dialout <(groups $(whoami)) || sudo usermod -a -G dialout $(whoami) # Note, will need to reboot if prompted for password


doc/Getting_Started.md
```diff
@@ -61,6 +63,7 @@ $ pip3 install --upgrade tockloader
 
 Ubuntu:
 ```
+$ apt install -y build-essential gcc-arm-none-eabi python3-pip curl
```
alexandruradovici, line 66: While this is definitely useful for fast getting started, it is only necessary of you are building for ARM.

```suggestion
# If planning to build for ARM
$ apt install -y build-essential gcc-arm-none-eabi python3-pip curl
```

doc/Getting_Started.md
```diff
@@ -30,8 +30,10 @@ developing Tock.
 
 1. [Rust](http://www.rust-lang.org/)
 2. [rustup](https://rustup.rs/) to install Rust (version >= 1.11.0)
-3. Command line utilities: make
-4. A supported board or QEMU configuration.
+3. Host toolchain (gcc, glibc)
```
alexandruradovici, line 33: I might be wrong, but I don't think the actual host toolchain is needed. I think the only requirements are the build utilities (find, make, ...).

dcz-self: I also have a Fedora dockerfile that I used, and it kept failing the same way until I added host gcc. Feel free to try:

```
# container for building Tock kernel
# podman build -f tock.dockerfile -t tock_kernel
# podman run -ti --name=tock_kernel -v.:/mnt/build/:z tock_kernel /bin/bash
FROM fedora:35

RUN dnf install -y arm-none-eabi-gcc-cs make gcc findutils
RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
```

Sadly, the containers are big and I ran out of space, so can't double-check right now :S

alexandruradovici: Got it. Do you still have the fail log? It would be interesting to see why.

dcz-self: Those are two failure modes that I've seen, both one after another, in the Debian-bsed container.

[tock.log](https://github.com/tock/tock/files/8026285/tock.log)

doc/Getting_Started.md
dcz-self: How about this instead;

```suggestion
$ apt install -y build-essential python3-pip curl
# If planning to build for ARM
$ apt install -y gcc-arm-none-eabi
```

I don't know exactly, but I think it's unlikely that the host portions (build-essential) are dependent on the target architecture.

doc/Getting_Started.md
alexandruradovici: Thank you, it seems to be a linker problem. Maybe others can have a better response to this.

doc/Getting_Started.md
dcz-self: BTW, reactions don't show up in email, so if I didn't check the thread using web UI, I would have thought this comment was forgotten.

doc/Getting_Started.md
hudson-ayers: The discussion at https://github.com/rust-lang/rust/issues/71515 might be informative.

I *think* that using `-Zgcc-ld=lld` might fix this so that we use the Rust-shipped lld instead of gcc's, but that is an unstable option. Unless we switch to using that, it seems accurate that we require a gcc installation, which is unfortunate.

hudson-ayers (dismissed):

doc/Getting_Started.md
```diff
@@ -61,6 +63,9 @@ $ pip3 install --upgrade tockloader
 
 Ubuntu:
 ```
+$ apt install -y build-essential python3-pip curl
```
lschuermann, line 66: I'm confused. Sure, some of the scripts in `/tools` are written in Python and may require some wheels, but is pip3 really required to build the Tock kernel?

doc/Getting_Started.md
bradjc: Am I understanding this correctly that what this means is there are some gcc files that llvm-ld is using?

doc/Getting_Started.md
hudson-ayers: `tockloader` requires it, and the instructions immediately below this instruct the user to install `tockloader` using pip3

bradjc (changes requested):
Can you try again without `VERBOSE=1`? I think that is causing some weird effects. If you want more output, you can do `make V=1`.

doc/Getting_Started.md
bradjc: I tried changing the makefile to:

```
RUSTC_FLAGS ?= \
  -C link-arg=-Tlayout.ld \
  -C linker=rust-lld \
  -C linker-flavor=ld.lld \
  -C relocation-model=static \
  -C link-arg=-nmagic \
  -C link-arg=-icf=all \
  -Z gcc-ld=lld \
```

and got:

```
     Running `rustc --crate-name hail --edition=2021 boards/hail/src/main.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type bin --emit=dep-info,link -C opt-level=z -C panic=abort -C lto -C codegen-units=1 -C debuginfo=2 -C link-arg=-L/Users/bradjc/git/tock/boards/hail -C metadata=5e232a3c154cf8d9 -C extra-filename=-5e232a3c154cf8d9 --out-dir /Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps --target thumbv7em-none-eabi -L dependency=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps -L dependency=/Users/bradjc/git/tock/target/release/deps --extern capsules=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libcapsules-1202625586e27f87.rlib --extern 'noprelude:compiler_builtins=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libcompiler_builtins-46d9de604032590c.rlib' --extern components=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libcomponents-9fd8dad9e0b9bcb3.rlib --extern 'noprelude:core=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libcore-fee7ca5f2aa3d4f2.rlib' --extern cortexm4=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libcortexm4-b147e95b800d37f5.rlib --extern kernel=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libkernel-dc50ffa5fc0275b6.rlib --extern sam4l=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libsam4l-fb66e817481084b7.rlib -Z unstable-options -C link-arg=-Tlayout.ld -C linker=rust-lld -C linker-flavor=ld.lld -C relocation-model=static -C link-arg=-nmagic -C link-arg=-icf=all -Z gcc-ld=lld --remap-path-prefix=/Users/bradjc/git/tock/= --remap-path-prefix=/Users/bradjc/.rustup/toolchains/nightly-2021-12-04-x86_64-apple-darwin/lib/rustlib/src/rust/library/core=/core/`
error: option `-Z gcc-ld` is used even though linker flavor is not gcc

error: could not compile `hail` due to previous error

Caused by:
  process didn't exit successfully: `rustc --crate-name hail --edition=2021 boards/hail/src/main.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type bin --emit=dep-info,link -C opt-level=z -C panic=abort -C lto -C codegen-units=1 -C debuginfo=2 -C link-arg=-L/Users/bradjc/git/tock/boards/hail -C metadata=5e232a3c154cf8d9 -C extra-filename=-5e232a3c154cf8d9 --out-dir /Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps --target thumbv7em-none-eabi -L dependency=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps -L dependency=/Users/bradjc/git/tock/target/release/deps --extern capsules=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libcapsules-1202625586e27f87.rlib --extern 'noprelude:compiler_builtins=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libcompiler_builtins-46d9de604032590c.rlib' --extern components=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libcomponents-9fd8dad9e0b9bcb3.rlib --extern 'noprelude:core=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libcore-fee7ca5f2aa3d4f2.rlib' --extern cortexm4=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libcortexm4-b147e95b800d37f5.rlib --extern kernel=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libkernel-dc50ffa5fc0275b6.rlib --extern sam4l=/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libsam4l-fb66e817481084b7.rlib -Z unstable-options -C link-arg=-Tlayout.ld -C linker=rust-lld -C linker-flavor=ld.lld -C relocation-model=static -C link-arg=-nmagic -C link-arg=-icf=all -Z gcc-ld=lld --remap-path-prefix=/Users/bradjc/git/tock/= --remap-path-prefix=/Users/bradjc/.rustup/toolchains/nightly-2021-12-04-x86_64-apple-darwin/lib/rustlib/src/rust/library/core=/core/` (exit status: 1)
make: *** [/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/hail] Error 101
```

I'm not sure this comment is correct (https://github.com/rust-lang/rust/issues/39915#issuecomment-998789288) because we use `-C linker=rust-lld` on mac and linux.

Anyway I think the issue is overriding the `VERBOSE` variable used inside the makefile. When I do that I can't get the kernel to build and get similar errors.

dcz-self:
make does work when "VERBOSE=1" is not present, even if gcc is not installed.

I was convinced "VERBOSE" was the standard name (cmake uses it), but clearly I haven't used autotools enough, where it appears to be called "V". I'm not really sure how to counteract that assumption.

dcz-self:
Updated without the host tools. What remains are extra tools that aren't always present.

A potential solution to VERBOSE clashes: would renaming the "VERBOSE" variable as used internally to something like "VERBOSE_FLAGS" be accepted? That would eliminate one way cmake users could paint themseves into a corner. It could also emit a message in the spirit of "Try using 'V=1' instead".

bradjc:
Something doesn't quite look right, and I'm not sure why. Specifically, in your output, the final rustc command:

```
     Running `rustc --crate-name build_script_build --edition=2021 boards/nordic/sma_q3/build.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type bin --emit=dep-info,link -C embed-bitcode=no -C debuginfo=2 -C debug-assertions=off -C metadata=fb4377ec039a2c18 -C extra-filename=-fb4377ec039a2c18 --out-dir /mnt/build/tock/target/release/build/sma_q3-fb4377ec039a2c18 -L dependency=/mnt/build/tock/target/release/deps`
```

Doesn't include any of the linker arguments (`-C link-arg=-Tlayout.ld -C linker=rust-lld -C linker-flavor=ld.lld -C relocation-model=static -C link-arg=-nmagic -C link-arg=-icf=all`). That could definitely cause rustc to call the wrong linker.

Have you tried building a tock board included in the kernel?

bradjc:
Ok thanks for confirming.

After puzzling over this for a while, it seems like we do in fact require a host GCC or clang to link the compiled `build.rs` file for each board crate. I wonder if there is a way to change that, although, it might be reasonable to just ask users to install a toolchain for their own host.

doc/Getting_Started.md
bradjc, line 66: Why do we need `gcc-arm-none-eabi`?

dcz-self:
I restored a previous version with host toolchain instructions. That also removes the commits I accidentally pushed from another branch.

doc/Getting_Started.md
dcz-self: That should be resolved in the updated commit.

bradjc: But even when building for ARM based boards gcc-arm-none-eabi shouldn't be needed.

doc/Getting_Started.md
lschuermann: How did I miss that? Thanks :)

bradjc (approved):

alistair23 (approved):

lschuermann (approved):

PR #2958, opened 2022-02-09
Store Kernel Managed Grant Resource Counters in One usize Value

bradjc:
### Pull Request Overview

Because the kernel manages the upcall array, read-only allow array, and read-write allow array for grants, it must keep track of the maximum number of upcalls and each allow type that there is allocated space for. This is necessary so the kernel can check on each allow() and subscribe() syscall if there is sufficient room to store the process-provided upcall or allowed buffer. These max numbers are also contained in the capsule that is using the grant, but the kernel has no access to those values when a syscall happens.

Currently, these maximum allocation number (or in practice array lengths) are each stored as a `usize` in the grant allocation. That means we use 12 bytes of RAM per allocated grant to store the array lengths. On a complex app (like the hail app), which uses 12 grant regions, that adds up to 144 bytes of memory.

However, it's hard to imagine a grant needing to store billions of upcalls or shared buffers. So, this PR merges the array lengths into a single `usize`, granting 8 bits to each value. The structure now looks like:

```text,ignore
0x003FFC8  ┌────────────────────────────────────┐
           │   T                                |
0x003FFxx  ├  ───────────────────────── ┐ K     |
           │   Padding (ensure T aligns)| e     |
0x003FF44  ├  ───────────────────────── | r     |
           │   SavedAllowRwN            | n     |
           │   ...                      | e     | G
           │   SavedAllowRw1            | l     | r
           │   SavedAllowRw0            |       | a
0x003FF44  ├  ───────────────────────── | O     | n
           │   SavedAllowRoN            | w     | t
           │   ...                      | n     |
           │   SavedAllowRo1            | e     | M
           │   SavedAllowRo0            | d     | e
0x003FF30  ├  ───────────────────────── |       | m
           │   SavedUpcallN             | D     | o
           │   ...                      | a     | r
           │   SavedUpcall1             | t     | y
           │   SavedUpcall0             | a     |
0x003FF24  ├  ───────────────────────── |       |
           │   Counters (usize)         |       |
0x003FF20  └────────────────────────────────────┘


Counters:

0             1             2             3         bytes
|-------------|-------------|-------------|-------------|
| # Upcalls   | # RO Allows | # RW Allows | [unused]    |
|-------------|-------------|-------------|-------------|
```

This reduces the RAM overhead per grant by 2/3, since only one `usize` is used rather than three.

#### Use of `u8`

Since the number of upcalls, read-only allows, and read-write allows per grant is now capped at 256, I changed the type for capsules to express the number they need from a `usize` to a `u8`. This causes many changes that are pretty inconsequential, but lead to a larger diff.

This change does propagate up to boards, however, because of how IPC is implemented. The max number of processes on the board is used to set the allow lengths. Maybe someone can take a look and see if there is a better way to handle this?

### Testing Strategy

hail app on hail.

### TODO or Help Wanted

I'm curious if there is a better way to handle the switch to u8 for IPC.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -277,8 +277,9 @@ impl KernelManagedLayout {
     ///
     /// # Safety
     ///
-    /// The incoming base pointer must be well aligned but cannot point to
-    /// initialized data. There must not be any other `KernelManagedLayout` for
+    /// The incoming base pointer must be well aligned and reference enough
+    /// memory to hold the entire kernel managed grant structure. There must
+    /// not be any other `KernelManagedLayout` for
     /// the given `base_ptr` at the same time, otherwise multiple mutable
     /// references to the same upcall/allow slices could be created.
     unsafe fn initialize_from_counts(
@@ -323,7 +324,7 @@ impl KernelManagedLayout {
         grant_t_align: GrantDataAlign,
     ) -> usize {
         let kernel_managed_size = size_of::<usize>()
-            + (upcalls_num.0 as usize * size_of::<SavedUpcall>())
+            + upcalls_num.0 as usize * size_of::<SavedUpcall>()
             + allow_ro_num.0 as usize * size_of::<SavedAllowRo>()
             + allow_rw_num.0 as usize * size_of::<SavedAllowRw>();
         // We know that grant_t_align is a power of 2, so we can make a mask
--- a/kernel/src/grant.rs
+++ b/kernel/src/grant.rs
@@ -183,37 +183,40 @@ impl<const NUM: usize> AllowRwSize for AllowRwCount<NUM> {
 /// ```text,ignore
 /// 0x003FFC8  ┌────────────────────────────────────┐
 ///            │   T                                |
-/// 0x003FFxx  ├  ───────────────────────── ┐       |
-///            │   Padding (ensure T aligns)|       |
-/// 0x003FF44  ├  ───────────────────────── |       |
-///            │   SavedAllowRwN            | K     |
+/// 0x003FFxx  ├  ───────────────────────── ┐ K     |
+///            │   Padding (ensure T aligns)| e     |
+/// 0x003FF44  ├  ───────────────────────── | r     |
+///            │   SavedAllowRwN            | n     |
 ///            │   ...                      | e     | G
-///            │   SavedAllowRw1            | r     | r
-///            │   SavedAllowRw0            | n     | a
-/// 0x003FF44  ├  ───────────────────────── | e     | n
-///            │   NumAllowRw (usize)       | l     | t
-/// 0x003FF40  ├  ───────────────────────── |       |
-///            │   SavedAllowRoN            | O     | M
-///            │   ...                      | w     | e
-///            │   SavedAllowRo1            | n     | m
-///            │   SavedAllowRo0            | e     | o
-/// 0x003FF30  ├  ───────────────────────── | d     | r
-///            │   NumAllowRo (usize)       |       | y
-/// 0x003FF2C  ├  ───────────────────────── | D     | 1
-///            │   SavedUpcallN             | a     |
-///            │   ...                      | t     |
-///            │   SavedUpcall1             | a     |
-///            │   SavedUpcall0             |       |
+///            │   SavedAllowRw1            | l     | r
+///            │   SavedAllowRw0            |       | a
+/// 0x003FF44  ├  ───────────────────────── | O     | n
+///            │   SavedAllowRoN            | w     | t
+///            │   ...                      | n     |
+///            │   SavedAllowRo1            | e     | M
+///            │   SavedAllowRo0            | d     | e
+/// 0x003FF30  ├  ───────────────────────── |       | m
+///            │   SavedUpcallN             | D     | o
+///            │   ...                      | a     | r
+///            │   SavedUpcall1             | t     | y
+///            │   SavedUpcall0             | a     |
 /// 0x003FF24  ├  ───────────────────────── |       |
-///            │   NumUpcalls (usize)       |       |
+///            │   Counters (usize)         |       |
 /// 0x003FF20  └────────────────────────────────────┘
 /// ```
+///
+/// The counters structure is composed as:
+///
+/// ```text,ignore
+/// 0             1             2             3         bytes
+/// |-------------|-------------|-------------|-------------|
+/// | # Upcalls   | # RO Allows | # RW Allows | [unused]    |
+/// |-------------|-------------|-------------|-------------|
+/// ```
 struct KernelManagedLayout {
-    upcalls_num: *mut usize,
+    counters_ptr: *mut usize,
     upcalls_array: *mut SavedUpcall,
-    allow_ro_num: *mut usize,
     allow_ro_array: *mut SavedAllowRo,
-    allow_rw_num: *mut usize,
     allow_rw_array: *mut SavedAllowRw,
 }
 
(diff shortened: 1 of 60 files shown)


lschuermann:
I think a per-driver `u8::MAX` limit for readable and writeable allows and subscribes each sounds reasonable, if it can lead to significant savings of memory usage in the Grant regions. However, it seems important to explicitly keep this change an implementation detail of the kernel and not otherwise propagate it to userspace, i.e. for subdriver numbers larger than 255 we'd respond as if the driver simply does not have more allow/subscribe slots. I think that is the case with this PR. This would allow us to undo these changes in the future without breaking anything.

Capping the number of processes on a board is something I'm slightly more worried about. While it should be possible to re-engineer the IPC interface to still support `usize` processes, if we'd merge this I see greater potential for other parts of the kernel and userspace to be reliant on the number of processes limited to 256.

bradjc:
> While it should be possible to re-engineer the IPC interface to still support `usize` processes, if we'd merge this I see greater potential for other parts of the kernel and userspace to be reliant on the number of processes limited to 256.

I agree we don't want `NUM_PROCS` (really that should be `MAX_NUM_PROCS`) to be further tied to being a `u8`. IPC is really the only place that number is used inside the core kernel, and IPC is definitely an outlier as a syscall driver. I don't think this change actually makes it more likely we would cap the max number of processes to 256 accidentally, but it certainly adds code that suggests that we only support up to 256 processes.

kernel/src/grant.rs
```diff
@@ -343,7 +361,94 @@ impl KernelManagedLayout {
...
+
+    /// Return mutable access to the slice of stored upcalls for this grant.
+    /// This is necessary for storing a new upcall.
+    fn get_upcalls_slice(&mut self) -> &mut [SavedUpcall] {
+        // # Safety
+        //
+        // Creating a `KernelManagedLayout` object ensures that the pointer to
+        // the upcall array is valid.
+        unsafe { slice::from_raw_parts_mut(self.upcalls_array, self.get_upcalls_number()) }
+    }
```
hudson-ayers, line 404: Calling this function twice in the same scope would give the caller multiple mutable references to the same memory, which is UB, without the caller having to use `unsafe`. This function should not be marked safe, and should document that callers are responsible for ensuring that only one reference to the `SavedUpcall` memory may exist at any time. Alternatively, we could get rid of this helper function, since it is only called once.

The same comment applies to the three functions below this as well.

kernel/src/grant.rs
```diff
@@ -237,81 +240,96 @@ struct GrantDataSize(usize);
...
+    /// Creates a layout from the specified pointer and lengths of arrays and
+    /// initializes the kernel owned portion of the layout.
     ///
     /// # Safety
     ///
-    /// The incoming base pointer must be well aligned but does not need to
-    /// point to initialized data.
-    unsafe fn from_counts(
+    /// The incoming base pointer must be well aligned but cannot point to
+    /// initialized data. There must not be any other `KernelManagedLayout` for
```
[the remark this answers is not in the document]
bradjc:
```suggestion
    /// The incoming base pointer must be well aligned and reference enough
    /// memory to hold the entire kernel managed grant structure. There must
    /// not be any other `KernelManagedLayout` for
```

kernel/src/grant.rs
```diff
@@ -237,81 +240,96 @@ struct GrantDataSize(usize);
...
         allow_rw_num: AllowRwItems,
         grant_t_size: GrantDataSize,
         grant_t_align: GrantDataAlign,
     ) -> usize {
-        let kernel_managed_size = 3 * size_of::<usize>()
-            + upcalls_num.0 * size_of::<SavedUpcall>()
-            + allow_ro_num.0 * size_of::<SavedAllowRo>()
-            + allow_rw_num.0 * size_of::<SavedAllowRw>();
+        let kernel_managed_size = size_of::<usize>()
+            + (upcalls_num.0 as usize * size_of::<SavedUpcall>())
```
bradjc, line 326:
```suggestion
            + upcalls_num.0 as usize * size_of::<SavedUpcall>()
```

kernel/src/grant.rs
bradjc: I added this helper function to better encapsulate the unsafe pointer portions, so they are not as mingled with all of the rest of the somewhat complex grant code. But what you bring up is a concern with this approach.

My attempt to address this is to require `&mut self` for this function, and then note in the `KernelManagedLayout` constructors that part of the safety contract there is that only one can be created per `base_ptr`. Does that logic work?

We might still want to change this even if my thinking is ok.

kernel/src/grant.rs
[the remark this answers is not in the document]
hudson-ayers: looks good to me

kernel/src/grant.rs
hudson-ayers: I think that logic would work if we only needed to create one `KernelManagedLayout` per base pointer, but currently it looks like we create one each time `enter_grant_kernel_managed` is called, which is called in `subscribe()`, `allow_ro()` and `allow_rw()` -- it seems like for a given base_ptr each of these might be called once or multiple times.

I think what we actually want to say is "only one KernelManagedLayout per base_ptr can exist at any time", but it does seem like that requirement might force us to propagate `unsafe` to several of the places where we currently create those layouts.

The easiest solution might actually be to just not create the slice references at all and instead encapsulate those accesses in some safe function that just uses raw pointer reads, but it does seem like a shame to have to do so.

hudson-ayers:
I finally had the chance to look through this change again today. I think that as currently constructed `KernelManagedLayout` is not very workable, because it is created in several places and it is very difficult to verify in the places where one is created that it is the only place it could be created with a given `base_ptr`. I am also not in favor of any solution that requires runtime checking of such an invariant, because that would eat into the memory savings of this PR and slow things down and be inelegant.

Instead, I think that we should do one of two things: 

a) Move unsafe back to `get_rx_allow_slice()`, `get_ro_allow_slice()`, `get_upcalls_slice()` and `get_resource_slices()`, with a note for each: 
``` 
/// SAFETY: 
/// To call this function, you must make sure that the lifetime of the 
/// returned slice does not overlap with the lifetime of any other slice 
/// returned by a call to this function or `get_resource_slices()` on a 
/// KernelManagedLayout with the same base_ptr as this one.
```

then also make `allow_rw()`, `allow_ro()`, and `subscribe()` unsafe, and propogate that message. Now, `Grant::enter()` must remain safe. So there, we should note that `Grant::enter()` / `try_enter()` must be the only place marked safe to call that calls `get_resource_slices()`. We already have protection to prevent entering the same Grant twice, so no worries there.

Now, the kernel (the lone caller of `allow_rw()`, `allow_ro()`, and `subscribe()`) just has to verify that it never calls one of those functions while a grant is entered. Because those are only called in the scheduler, I think this is reasonably straightforward to show.

b) Have get_ro_allow_slice() etc. return a raw pointer instead of a slice, and never cast the underlying type to a rust reference. I don't like this as much.

I am working on a commit that I believe accomplishes option a, and removes all of the `unsafe` around creating a `KernelManagedLayout`

bradjc:
> because it is created in several places and it is very difficult to verify in the places where one is created that it is the only place it could be created with a given `base_ptr`.

Doesn't the guarantee follow directly from the invariants of entering a grant? To get the base pointer we must enter the grant, and that can only be done once at a time, as verified by ProcessStandard. I don't think this adds any complexity we haven't already introduced by restricting grants to only be opened once at a time.

bradjc:
> I think we want to tie the lifetime of `KernelManagedLayout` to the lifetime of the entered grant?

Yes absolutely. Can we do something where calling `process.enter_grant()` gives you something you have to pass back to `process.leave_grant()`? And that `KernelManagedLayout` can't live longer than that? This isn't my area of expertise.

phil-levis:
Bump! Did this make any progress?

bradjc:
Short answer: no.

bradjc:
Ok @hudson-ayers, made it a little closer. I removed `GrantLifetimeGuard` and moved its functionality into `KernelManagedLayout`. So the grant is automatically exited after (what is now) `EnteredGrantKernelManagedLayout` goes out of scope.

However, is there a way to tie `EnteredGrantKernelManagedLayout` more closely to the entered grant, or do we need to change `process.enter_grant()` to return a custom type rather than a `*mut u8` so we can enforce somehow that `EnteredGrantKernelManagedLayout` cannot live longer than the returned custom type?

hudson-ayers:
> Ok @hudson-ayers, made it a little closer. I removed `GrantLifetimeGuard` and moved its functionality into `KernelManagedLayout`. So the grant is automatically exited after (what is now) `EnteredGrantKernelManagedLayout` goes out of scope.

Thanks for revisiting this! This seems like it is getting close.

> However, is there a way to tie `EnteredGrantKernelManagedLayout` more closely to the entered grant, or do we need to change `process.enter_grant()` to return a custom type rather than a `*mut u8` so we can enforce somehow that `EnteredGrantKernelManagedLayout` cannot live longer than the returned custom type?

I think the easier solution is making it unsafe to leave a grant except for within the Drop implementation of `EnterGrantKernelManagedLayout`. I just added a commit that adds this requirement. It has been a while since I thought deeply about this so let me know if I seem to be missing something, but I think this addresses my concern from last time.

bradjc:
I don't like adding `EnterGrantKernelManagedLayout` to the process interface, as this is an implementation detail of grants, and if history is any indication, it is likely to change.

hudson-ayers:
> I don't like adding `EnterGrantKernelManagedLayout` to the process interface, as this is an implementation detail of grants, and if history is any indication, it is likely to change.

That's fine, I originally added it as an owned type as an alternative to making leave_grant unsafe, I think it is fine to change that part back. I do think it makes it a little harder to add an incorrect call to leave_grant but I suppose I agree that benefit does not outweigh the cost you described

hudson-ayers (approved):
Brad and I met today, I believe my concerns have been resolved.

alistair23 (dismissed):

kernel/src/grant.rs
```diff
@@ -343,7 +389,111 @@ impl KernelManagedLayout {
...
+        //
+        // Creating a `EnteredGrantKernelManagedLayout` object requires that the pointers
+        // are well aligned and point to valid memory.
+        let counters_val = unsafe { self.counters_ptr.read() };
+        (counters_val >> offset_bits) & 0xFF
+    }
+
+    /// Return the number of upcalls stored by the core kernel for this grant.
+    fn get_upcalls_number(&self) -> usize {
+        self.get_counter_offset(0)
```
lschuermann, line 409:
```suggestion
        u32::to_be_bytes(unsafe { self.counters_ptr.read() } as u32)[0]
```

Gets away without the helper function, same for `allow_rw/ro`.

lschuermann (dismissed):
Looks good to me, just some very minor nits.

bradjc:
Rebased and updated.

lschuermann (approved):
Looks good to me!

kernel/src/grant.rs
bradjc: I thought I commented here, but using a helper function allows only one unsafe and not having to copy the safety comment several times.

hudson-ayers (approved):
bors r+

PR #2959, opened 2022-02-09
Update Nightly to Feb 2022

bradjc:
### Pull Request Overview

The headline here being: `asm!()` is stabilized! (As of https://github.com/rust-lang/rust/pull/91728.) Most changes are to remove the asm feature and add `use core::arch:asm`.

### Testing Strategy

travis

### TODO or Help Wanted

Resolved: ~~Can someone look at https://github.com/tock/tock/compare/update-nightly-feb-2022?expand=1#diff-494cb9c3b94f58f1ab43cca69f63906088f87a97726d9260665f8529f605f2c9L259? I had to remove some clobbers to remove an error. I think that is ok (https://stackoverflow.com/questions/41380345/how-to-use-hir8-r12-register-in-cortex-m0) but I'm not sure.~~

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


arch/cortex-m0/src/lib.rs
```diff
@@ -256,8 +258,7 @@ pub unsafe extern "C" fn switch_to_user(
     ",
     inout("r0") user_stack,
     in("r1") process_regs,
-    out("r2") _, out("r3") _, out("r4") _, out("r5") _, out("r8") _,
-    out("r10") _, out("r11") _, out("r12") _);
+    out("r2") _, out("r3") _, out("r4") _, out("r5") _, out("r11") _);
```
hudson-ayers, line 261: I am confused by r11 being allowed, but not r8/r10/r12. If this was a thumb1 vs thumb2 thing then https://doc.rust-lang.org/nightly/reference/inline-assembly.html suggests that only r0-r7 should be allowed. Can you share what error you get?

alexandruradovici: I am having the same issue when trying to add cortex-m0 support libtock-rs. I tried adding (just for this example) the r11 register, it does not seem to complain.

```rust
unsafe {
    asm!("svc 0",
         inlateout("r0") r0 => _, // a1
         lateout("r1") _,         // a2
         lateout("r2") _,         // a3
         lateout("r3") _,         // a4
         // r4-r8 are callee-saved.
         // r9 is platform-specific. We don't use it in libtock_runtime,
         // so it is either unused or used as a callee-saved register.
         // r10 and r11 are callee-saved.
         lateout("r11") _, // added for testing
         lateout("r12") _, // ip
         // r13 is the stack pointer and must be restored by the callee.
         lateout("r14") _, // lr
         // r15 is the program counter.
    );
}
```
The build output

```
error: invalid register `r12`: high registers (r8+) cannot be used in Thumb-1 code
  --> runtime/src/syscalls_impl_arm.rs:18:18
   |
18 |                  lateout("r12") _, // ip
   |                  ^^^^^^^^^^^^^^^^

error: invalid register `r14`: high registers (r8+) cannot be used in Thumb-1 code
  --> runtime/src/syscalls_impl_arm.rs:20:18
   |
20 |                  lateout("r14") _, // lr
   |                  ^^^^^^^^^^^^^^^^
```

alexandruradovici: That is cont clear to me. I submitted https://github.com/tock/libtock-rs/pull/381, where I used `clobber_abi("C")`, but that might not be correct. From the disassembly it seems to have no effect, at least for `yield1`.

hudson-ayers: I think we should maybe just wait on https://github.com/rust-lang/rust/pull/90796#issuecomment-1035374829 rather than potentially breaking cortex-m0 by updating now

bradjc: I updated to the latest nightly that works (2022-02-23) and I don't get an error, so I don't think this is resolved.

bradjc: Oh wait I have that backwards, it does compile without the extra commit. I've pushed a new version.

hudson-ayers (approved):

alistair23 (approved):

PR #2962, opened 2022-02-09
make: move -L to build.rs

bradjc:
### Pull Request Overview

Per the comment for `RUSTC_FLAGS_FOR_BIN`:

> Ultimately, this should move to the Cargo.toml, for example when
> https://github.com/rust-lang/cargo/pull/7811 is merged into Cargo.

this moves setting the linker script search path to build.rs, which is enabled by the addition of the `cargo:rustc-link-arg=FLAG` instruction for build scripts.

This removes the need to special case the -L flag so that it is only used when creating a binary and not when compiling dependencies, since that is exactly what the new build script instruction handles.

### Testing Strategy

Running make on two boards and making sure all crates are not recompiled.

### TODO or Help Wanted

Does this look ok? It needs to be replicated for all build.rs files, but I wanted to make sure this is what we want first.

Also, we can move other flags from the Makefile to use this build script technique. Should we? Is there a way to have a common build.rs file that all boards can pull in so we don't have to duplicate the same lines for every board?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/Makefile.common
+++ b/boards/Makefile.common
@@ -84,27 +84,6 @@ ifeq ($(CI),true)
   RUSTC_FLAGS_TOCK += -D warnings
 endif
 
-# The following flags should only be passed to the board's binary crate, but
-# not to any of its dependencies (the kernel, capsules, chips, etc.). The
-# dependencies wouldn't use it, but because the link path is different for each
-# board, Cargo wouldn't be able to cache builds of the dependencies.
-#
-# Indeed, as far as Cargo is concerned, building the kernel with
-# `-C link-arg=-L/tock/boards/imix` is different than building the kernel with
-# `-C link-arg=-L/tock/boards/hail`, so Cargo would have to rebuild the kernel
-# for each board instead of caching it per board (even if in reality the same
-# kernel is built because the link-arg isn't used by the kernel).
-#
-# Ultimately, this should move to the Cargo.toml, for example when
-# https://github.com/rust-lang/cargo/pull/7811 is merged into Cargo.
-#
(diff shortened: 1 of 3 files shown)


jettr:
> @jettr @a-pronin not sure if either of you have thoughts on this proposal, wanted to flag it since you would have to take it into account next time you update Tock versions if it is merged

Not gonna lie, I don't immediately seen what we are going to have to change. Can you help me out a little more please? Sorry you have to spoon feed me the explanation.

hudson-ayers:
> Not gonna lie, I don't immediately seen what we are going to have to change. Can you help me out a little more please? Sorry you have to spoon feed me the explanation.

I think you will need to add the link-arg line to build.rs for each project, since previously it was being added by the Tock Makefile when building the kernel (your Makefile does not overwrite RUSTC_FLAGS_FOR_BIN to be empty, so I am assuming that linker arg was needed)

jettr:
> I think you will need to add the link-arg line to build.rs for each project, since previously it was being added by the Tock Makefile when building the kernel (your Makefile does not overwrite RUSTC_FLAGS_FOR_BIN to be empty, so I am assuming that linker arg was needed)

That makes sense. After this PR lands, we can cherry pick it now and make that change so we aren't surprise when to upgrade to 2.1. Thank you for the heads up!

bradjc:
It seems like at the end of the day cargo still doesn't really have a good solution for us. We can use `build.rs` for a few things in our build process, hidden cargo files for a few other things, and then the rest has to go in the makefile anyway. While `cargo:rustc-link-arg-bins` is clearly exactly what we want in terms of functionality, moving one thing to build.rs, and having to duplicate it for every board, is not clearly worth it.

PR #2955, opened 2022-02-04
opentitan: Verilator fixup for ot bump

thulithwilfred:
### Pull Request Overview

Updated the build-system for tock-verilator to match the changed target
(boot/otp images)
paths for the recent OT bump [1]. The path change is referenced here
[2]

Added specific target build instructions to README
as these targets are required, and since running
`ninja -C build-out all` will typically fail to complete. Also much
quicker to just build only the required targets.

[1] https://github.com/tock/tock/pull/2954
[2] https://github.com/lowRISC/opentitan/issues/9154

Signed-off-by: Wilfred Mallawa <wilfred.mallawa@wdc.com>

### Testing Strategy

Building the updated targets for both cw130 and nexysvideo

### TODO or Help Wanted
N/A

### Documentation Updated
- [x] Updated README 

### Formatting
- [x] Ran `make fmt`.

PR #2954, opened 2022-02-03
boards/opentitan: Bump the a newer build

alistair23:
### Pull Request Overview

Update the OpenTitan bitstream to a newer build.

### Testing Strategy

Running the OpenTitan tests and app on hardware.

### Documentation Updated

- [X] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [X] Ran `make prepush`.

PR #2963, opened 2022-02-10
opentitan: verilator: fixup test build

thulithwilfred:
### Pull Request Overview

Fixup an issue when `make test-verilator` is ran, the binary loaded to
Verilator is incorrect. It would load the `binary.64.vmem` file which
might have been previously created by `make
BOARD_CONFIGURATION=sim_verilator verilator` or not at all (error). Note that
a `.vmem` created by the previous command will not have been built for tests.

This PR updates the `makefile/run.sh` for cw130/nexysvideo  to ensure
that the `.vmem` created using the test binary is loaded into Verilator.

### Depends

This PR builds on: 
* https://github.com/tock/tock/pull/2955
* https://github.com/tock/tock/pull/2954
* https://github.com/tock/tock/pull/2953

All merged.

### Testing Strategy

`make test-verilator` for both boards. The test outputs observed over `screen dev/pts/x`.

### TODO or Help Wanted
To keep things tidy, I output all the Verilator test build files into `verilator_build/` within the current board directory. So you could easily just delete that after. What do you guys think about this implementation? 

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### Update
17/2 - Implemented OBJCOPY variable to be used for object copying.

### DIFF
--- a/boards/opentitan/earlgrey-cw310/run.sh
+++ b/boards/opentitan/earlgrey-cw310/run.sh
@@ -1,18 +1,31 @@
 #!/bin/bash
 
+BUILD_DIR="verilator_build/"
+
 if [[ "${VERILATOR}" == "yes" ]]; then
-	${OBJCOPY} --output-target=binary --strip-sections -S --remove-section .apps "${1}" binary
-	srec_cat binary \
+		if [ -d "$BUILD_DIR" ]; then
+			# Cleanup before we build again
+			printf "\n[CW-130: Verilator Tests]: Cleaning up verilator_build...\n\n"
+			rm -R "$BUILD_DIR"/*
+		else
+			printf "\n[CW-130: Verilator Tests]: Setting up verilator_build...\n\n"
+			mkdir "$BUILD_DIR"
+		fi
+	# Copy in and covert from cargo test output to binary
+	${OBJCOPY} ${1} "$BUILD_DIR"/earlgrey-cw310-tests.elf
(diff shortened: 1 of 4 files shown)


thulithwilfred:
@alistair23

boards/opentitan/earlgrey-cw310/run.sh
```diff
@@ -1,14 +1,27 @@
...
+		if [ -d "$BUILD_DIR" ]; then
+			# Cleanup before we build again
+			printf "\n[CW-130: Verilator Tests]: Cleaning up verilator_build...\n\n"
+			rm -R "$BUILD_DIR"/*
+		else
+			printf "\n[CW-130: Verilator Tests]: Setting up verilator_build...\n\n"
+			mkdir "$BUILD_DIR"
+		fi
+	# Copy in and covert from cargo test output to binary
+	riscv64-linux-gnu-objcopy ${1} "$BUILD_DIR"/earlgrey-cw310-tests.elf
```
alistair23, line 15: You should use `${OBJCOPY}` here instaed

alistair23 (changes requested):

boards/opentitan/earlgrey-cw310/run.sh
thulithwilfred: ${OBJCOPY} doesn't get set here, so we can't use it, without explicitly setting it i think.

thulithwilfred: @alistair23 added support for the makefile to pass it into `run.sh`. This should keep things clean.

alistair23 (approved):

PR #2966, opened 2022-02-16
opentitan: add spi_host 

thulithwilfred:
### Pull Request Overview

This pull request adds spi_host implementation to opentitan. The driver has been designed with reference to [1]. Opentitan has both spi_device and spi_host capabilities, this PR implements only the spi_host. The base register mapping can be found [2].

### Testing Strategy

Tested by using Verilator and  basic testing on FPGA with the provided unit tests. For Verilator testing, this PR depends on:
- https://github.com/tock/tock/pull/2963

Tested on the ongoing `ibex_spi` development branch on qemu.

### TODO 

Mainline qemu still requires a patch to update spi_host/spi_device base addresses. As such, testing on qemu by including the `#cfg...` will cause a store access fault when the board initializes SPI and attempts to write to 'invalid' registers in qemu. To avoid CI failure, these sections have be isolated using `#cfg...`. To be removed when qemu fixup pr is merged. 

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### References
[1] https://docs.opentitan.org/hw/ip/spi_host/doc/

[2] https://github.com/lowRISC/opentitan/blob/6c317992fbd646818b34f2a2dbf44bc850e461e4/hw/top_earlgrey/sw/autogen/top_earlgrey_memory.h#L107


thulithwilfred:
@alistair23

boards/opentitan/src/main.rs
```diff
@@ -126,6 +126,11 @@ struct EarlGreyNexysVideo {
         capsules::virtual_uart::UartDevice<'static>,
     >,
     i2c_master: &'static capsules::i2c_master::I2CMasterDriver<'static, lowrisc::i2c::I2c<'static>>,
+
```
alistair23, line 129: No newline here

chips/earlgrey/src/spi_host.rs
```diff
@@ -0,0 +1,10 @@
+use kernel::utilities::StaticRef;
+use lowrisc::spi_host::SpiHostRegisters;
+
+//These addresses have been changed in mainline of OT
+//Refer: https://github.com/lowRISC/opentitan/blob/c4f342b9349ba033a5f22fba9349999299a1b2bf/hw/top_earlgrey/sw/autogen/top_earlgrey_memory.h#L179
```
alistair23, line 5: Do you mean they have been changed since the commit we use for OT?

boards/opentitan/src/tests/spi_host.rs
```diff
@@ -0,0 +1,232 @@
...
+    }
+}
+
+impl<'a> SpiMasterClient for SpiHostCallback {
+    fn read_write_done(
+        &self,
+        tx_data: &'static mut [u8],
+        rx_done: Option<&'static mut [u8]>,
+        tx_len: usize,
+        _rc: Result<(), ErrorCode>,
```
alistair23, line 40: You should be checking this result

boards/opentitan/src/tests/spi_host.rs
```diff
@@ -0,0 +1,232 @@
...
+    ) {
+        //Transfer Complete
+        assert_eq!(tx_len, self.tx_len.get());
+
+        //Capture Buffers
+        self.tx_data.replace(tx_data);
+        if rx_done.is_none() == false {
```
alistair23, line 47: You don't need the `== false`.

You should be able to use an `if let Some()..` here which is more Rusty

boards/opentitan/src/tests/spi_host.rs
```diff
@@ -0,0 +1,232 @@
...
+    let cb = unsafe { static_init_test_partial_cb() };
+
+    debug!("[SPI] Setup spi_host0 partial_transfer... ");
+    run_kernel_op(100);
+
+    spi_host.set_client(cb);
+    cb.reset();
+
+    let tx = cb.tx_data.take().unwrap();
+    let _rx = cb.rx_data.take().unwrap();
```
alistair23, line 138: Don't use an underscore when you are using the variable

boards/opentitan/src/tests/spi_host.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    let cb = unsafe { static_init_test_cb() };
+
+    debug!("[SPI] Setup spi_host0 transfers... ");
+    run_kernel_op(100);
+    spi_host.set_client(cb);
+    cb.reset();
+
+    let tx = cb.tx_data.take().unwrap();
+    let _rx = cb.rx_data.take().unwrap();
```
alistair23, line 181: Same here

boards/opentitan/src/tests/spi_host.rs
```diff
@@ -0,0 +1,232 @@
...
+    run_kernel_op(5000);
+
+    #[cfg(feature = "hardware_tests")]
+    assert_eq!(cb.transfer_done.get(), true);
+
+    run_kernel_op(100);
+    debug!("    [ok]");
+    run_kernel_op(100);
+
+    let cb2 = unsafe { static_init_test_cb() };
```
alistair23, line 207: Can't you just re-use the existing callback?

chips/lowrisc/src/spi_host.rs
```diff
@@ -0,0 +1,672 @@
+//! Serial Peripheral Interface (SPI) Host Driver
+use core::cell::Cell;
+use core::cmp;
+use core::option::Option;
```
alistair23, line 4: I don't think you need this

chips/lowrisc/src/spi_host.rs
```diff
@@ -0,0 +1,672 @@
...
+const SPI_HOST_CMD_STANDARD_SPI: u32 = 0;
+
+impl SpiHost {
+    pub const fn new(base: StaticRef<SpiHostRegisters>, cpu_clk: u32) -> Self {
+        SpiHost {
+            registers: base,
+            client: OptionalCell::empty(),
+            busy: Cell::new(false),
+            chip_select: Cell::new(0),
+            cpu_clk: Cell::new(cpu_clk),
```
alistair23, line 159: This doesn't need to be in a `Cell`

chips/lowrisc/src/spi_host.rs
```diff
@@ -0,0 +1,672 @@
...
+    }
+
+    pub fn handle_interrupt(&self) {
+        let regs = self.registers;
+        let irq = regs.intr_state.extract();
+        self.disable_interrupts();
+
+        if irq.is_set(intr::ERROR) {
+            if self.rx_buf.is_some() {
+                self.rx_buf.take().map(|rx_buf| {
```
alistair23, line 177: You don't need to check `is_some()` then use `map()`, just use `map()` directly

alistair23: But you don't even need that, just pass `self.rx_buf.take()` directly to `read_write_done()`

chips/lowrisc/src/spi_host.rs
```diff
@@ -0,0 +1,672 @@
...
+                            self.tx_offset.get(),
+                            Err(ErrorCode::FAIL),
+                        ),
+                    });
+                });
+                //Specified to be cleared, after err_status is cleared.
+                self.clear_err_interrupt();
+                //Something went wrong, reset IP and clear buffers
+                self.reset_spi_ip();
+                self.reset_internal_state();
```
alistair23, line 192: You should probably do this before the callback. It's possible the callback will try the operation again as it failed, in which case you don't want to clear the data after a new operation has started

alistair23 (changes requested):

chips/earlgrey/src/spi_host.rs
thulithwilfred: Yep since the commit, I will remove this comment cause it's all up-to date now. I put it there as a *note to self*

boards/opentitan/src/tests/spi_host.rs
thulithwilfred: Put this in to silence build warnings (unused variable) for when not building tests with `hardware-features`. Until qemu is patched. Thoughts?

thulithwilfred: Alternatively, i can wrap them in `#cfg` too

alistair23 (approved):

PR #2968, opened 2022-02-16
capsules: Add bmp280 temperature sensor

dcz-self:
### Pull Request Overview

This pull request adds the BMP280 driver with temperature measurement functionality present (no pressure measurement).

### Testing Strategy

This pull request was tested on the SMA Q3 smart watch (sold as [Jazda](https://jazda.org/) and Bangle.js 2 versions before 2.1). Working branch: https://github.com/dcz-self/tock/tree/sma_q3

### TODO or Help Wanted

Pressure measurement is not present, but I don't think it's needed in this request.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/bmp280.rs
+++ b/boards/components/src/bmp280.rs
@@ -1,0 +1,104 @@
+//! Component for the BMP280 Temperature and Pressure Sensor.
+//!
+//! Based off the SHT3x code.
+//!
+//! I2C Interface
+//!
+//! Usage
+//! -----
+//!
+//! With the default i2c address
+//! ```rust
+//! let bmp280 = components::bmp280::Bmp280Component::new(sensors_i2c_bus, mux_alarm).finalize(
+//!         components::bmp280_component_helper!(nrf52::rtc::Rtc<'static>),
+//!     );
+//! bmp280.begin_reset();
+//! ```
+//!
+//! With a specified i2c address
+//! ```rust
+//! let bmp280 = components::bmp280::Bmp280Component::new(sensors_i2c_bus, mux_alarm).finalize(
+//!         components::bmp280_component_helper!(nrf52::rtc::Rtc<'static>, capsules::bmp280::BASE_ADDR),
+//!     );
+//! bmp280.begin_reset();
+//! ```
+
+use crate::i2c::I2CComponent;
+use crate::i2c_component_helper;
+use capsules::bmp280::Bmp280;
+use capsules::virtual_alarm::{MuxAlarm, VirtualMuxAlarm};
+use capsules::virtual_i2c::MuxI2C;
+use core::mem::MaybeUninit;
+use kernel::component::Component;
+use kernel::hil::time::Alarm;
+
+use kernel::static_init_half;
+
+/// Setup static space for the driver and its requirements.
+#[macro_export]
+macro_rules! bmp280_component_helper {
+    ($A:ty) => {{
+        use capsules::bmp280;
+        $crate::bmp280_component_helper!($A, bmp280::BASE_ADDR)
+    }};
+
+    // used for specifically stating the i2c address
+    // as some boards (like nrf52) require a shift
+    ($A:ty, $address: expr) => {{
+        use capsules::bmp280::Bmp280;
+        use core::mem::MaybeUninit;
+
+        static mut BUFFER: [u8; bmp280::BUFFER_SIZE] = [0; bmp280::BUFFER_SIZE];
+
+        static mut BMP280: MaybeUninit<Bmp280<'static, VirtualMuxAlarm<'static, $A>>> =
+            MaybeUninit::uninit();
+        static mut BMP280_ALARM: MaybeUninit<VirtualMuxAlarm<'static, $A>> = MaybeUninit::uninit();
+        (&mut BMP280_ALARM, &mut BUFFER, &mut BMP280, $address)
+    }};
+}
+
+pub struct Bmp280Component<A: 'static + Alarm<'static>> {
+    i2c_mux: &'static MuxI2C<'static>,
+    alarm_mux: &'static MuxAlarm<'static, A>,
+}
+
+impl<A: 'static + Alarm<'static>> Bmp280Component<A> {
+    pub fn new(
+        i2c_mux: &'static MuxI2C<'static>,
+        alarm_mux: &'static MuxAlarm<'static, A>,
+    ) -> Bmp280Component<A> {
+        Bmp280Component { i2c_mux, alarm_mux }
+    }
+}
+
+impl<A: 'static + Alarm<'static>> Component for Bmp280Component<A> {
+    type StaticInput = (
+        &'static mut MaybeUninit<VirtualMuxAlarm<'static, A>>,
+        &'static mut [u8],
+        &'static mut MaybeUninit<Bmp280<'static, VirtualMuxAlarm<'static, A>>>,
+        u8,
+    );
+    type Output = &'static Bmp280<'static, VirtualMuxAlarm<'static, A>>;
+
+    unsafe fn finalize(self, static_buffer: Self::StaticInput) -> Self::Output {
+        let bmp280_i2c =
+            I2CComponent::new(self.i2c_mux, static_buffer.3).finalize(i2c_component_helper!());
+
+        let bmp280_alarm = static_init_half!(
+            static_buffer.0,
+            VirtualMuxAlarm<'static, A>,
+            VirtualMuxAlarm::new(self.alarm_mux)
+        );
+        bmp280_alarm.setup();
+
+        let bmp280 = static_init_half!(
+            static_buffer.2,
+            Bmp280<'static, VirtualMuxAlarm<'static, A>>,
+            Bmp280::new(bmp280_i2c, static_buffer.1, bmp280_alarm)
+        );
+        bmp280_i2c.set_client(bmp280);
+        bmp280_alarm.set_alarm_client(bmp280);
+
+        bmp280
+    }
+}
--- a/capsules/src/bmp280.rs
+++ b/capsules/src/bmp280.rs
@@ -1,0 +1,483 @@
+//! hil driver for Bmp280 Temperature and Pressure Sensor
+//!
+//! Written by Dorota <gihu.dcz@porcupinefactory.org>
+//!
+//! Based off the SHT3x code.
+//!
+//! Not implemented: pressure
+
+use core::cell::Cell;
+use kernel::debug;
+use kernel::hil;
+use kernel::hil::i2c;
+use kernel::hil::time::{Alarm, ConvertTicks};
+use kernel::utilities::cells::{OptionalCell, TakeCell};
+use kernel::ErrorCode;
+
+pub static BASE_ADDR: u8 = 0x76;
+
+/// Currently sized enough for temperature readings only.
+pub const BUFFER_SIZE: usize = 6;
+
+#[allow(non_camel_case_types)]
+#[allow(dead_code)]
+enum Register {
+    /// First register of calibration data.
+    /// Each register is 2 bytes long.
+    DIG_T1 = 0x88,
+    DIG_T2 = 0x8a,
+    DIG_T3 = 0x8c,
+    ID = 0xd0,
+    RESET = 0xe0,
+    /// measuring: [3]
+    /// im_update: [0]
+    STATUS = 0xf3,
+    /// osrs_t: [7:5]
+    /// osrs_p: [4:2]
+    /// mode: [1:0]
+    CTRL_MEAS = 0xf4,
+    /// t_sb: [7:5]
+    /// filter: [4:2]
+    /// spi3w_en: [0]
+    CONFIG = 0xf5,
+    PRESS_MSB = 0xf7,
+    PRESS_LSB = 0xf8,
+    /// xlsb: [7:4]
+    PRESS_XLSB = 0xf9,
+    TEMP_MSB = 0xfa,
+    TEMP_LSB = 0xfb,
+    /// xlsb: [7:4]
+    TEMP_XLSB = 0xfc,
+}
+
+#[derive(Clone, Copy, PartialEq, Debug)]
+struct CalibrationData {
+    dig_t1: u16,
+    dig_t2: i16,
+    dig_t3: i16,
+    // TODO: pressure calibration
+}
+
+/// CAUTION: calibration data puts least significant byte in the lowest address,
+/// readouts do the opposite.
+fn twobyte(lsb: u8, msb: u8) -> u16 {
+    u16::from_be_bytes([msb, lsb])
+}
+
+impl CalibrationData {
+    fn new(i2c_raw: &[u8]) -> Self {
+        CalibrationData {
+            dig_t1: twobyte(i2c_raw[0], i2c_raw[1]) as u16,
+            dig_t2: twobyte(i2c_raw[2], i2c_raw[3]) as i16,
+            dig_t3: twobyte(i2c_raw[4], i2c_raw[5]) as i16,
+        }
+    }
+
+    fn temp_from_raw(&self, raw_temp: u32) -> i32 {
+        let temp = raw_temp as i32; // guaranteed to succeed because raw temp has only 20 significant bits maximum.
+        let dig_t1 = self.dig_t1 as i32; // same, 16-bits
+        let dig_t2 = self.dig_t2 as i32; // same, 16-bits
+        let dig_t3 = self.dig_t3 as i32; // same, 16-bits
+                                         // From the datasheet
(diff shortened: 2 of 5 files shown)


capsules/src/bmp280.rs
```diff
@@ -0,0 +1,378 @@
...
+                    Ok(())
+                }
+                State::MismatchedDevice => Err(ErrorCode::NODEVICE),
+                _ => Err(ErrorCode::ALREADY),
+            },
+        )
+    }
+
+    pub fn read_temperature(&self) -> Result<(), ErrorCode> {
+        use State::*;
```
bradjc, line 206: We prohibit using `use...*` in Tock, please just use `State::`.

boards/components/src/bmp280.rs
```diff
@@ -0,0 +1,104 @@
...
+        $crate::bmp280_component_helper!($A, bmp280::BASE_ADDR)
+    }};
+
+    // used for specifically stating the i2c address
+    // as some boards (like nrf52) require a shift
+    ($A:ty, $address: expr) => {{
+        use capsules::bmp280::Bmp280;
+        use core::mem::MaybeUninit;
+
+        static mut BUFFER: [u8; bmp280::CALIBRATION_BYTES + 1] = [0; bmp280::CALIBRATION_BYTES + 1];
```
bradjc, line 51: Can you define an explicit buffer size in the capsule itself? Something like https://github.com/tock/tock/blob/1fe93e8dd8ab3a89ac20ff1df332681fc4ac2e29/capsules/src/spi_controller.rs#L35. That way we keep the logic for the required minimum buffer size with the driver code.

capsules/src/bmp280.rs
```diff
@@ -0,0 +1,378 @@
...
+        &self,
+        buffer: &'static mut [u8],
+        addr: Register,
+        data: [u8; COUNT],
+    ) {
+        buffer[0] = addr as u8;
+        for (i, d) in data.iter().enumerate() {
+            buffer[i + 1] = *d;
+        }
+        // TODO: propagate errors
```
bradjc, line 139: I think it is worth handling this TODO now, because we would like to avoid the `unwrap()` calls in capsule code.

capsules/src/bmp280.rs
```diff
@@ -0,0 +1,378 @@
...
+            state: Cell::new(State::Uninitialized),
+            buffer: TakeCell::new(buffer),
+            alarm: alarm,
+        }
+    }
+
+    /// Resets the device and brings it into a known state.
+    pub fn begin_reset(&self) -> Result<(), ErrorCode> {
+        self.buffer.take().map_or_else(
+            || panic!("BMP280 No buffer available!"),
```
bradjc, line 192: I think using just `map_or` is fine here. Adding panics makes the kernel binary larger, and with this type of check this error never happens in practice because the buffer is always returned to the TakeCell (or the driver completely breaks).

capsules/src/bmp280.rs
```diff
@@ -0,0 +1,378 @@
...
+        addr: Register,
+        data: [u8; COUNT],
+    ) {
+        buffer[0] = addr as u8;
+        for (i, d) in data.iter().enumerate() {
+            buffer[i + 1] = *d;
+        }
+        // TODO: propagate errors
+        self.i2c.enable();
+        self.i2c.write(buffer, COUNT as u8 + 1).unwrap();
```
alexandruradovici, line 141: Please handle the error here instead of using unwrap. You can use this https://github.com/tock/tock/blob/1fe93e8dd8ab3a89ac20ff1df332681fc4ac2e29/capsules/src/apds9960.rs#L205 as an example.

capsules/src/bmp280.rs
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    /// Requests a read into buffer.
+    /// Parse the result using `parse_read`.
+    fn read(&self, buffer: &'static mut [u8], addr: Register, count: u8) {
```
alexandruradovici, line 146: Return the buffer in case if an error
```suggestion
    fn read(&self, buffer: &'static mut [u8], addr: Register, count: u8) -> Result<(), (ErrorCode, &'static mut [u8])> {
```

capsules/src/bmp280.rs
```diff
@@ -0,0 +1,378 @@
...
+
+    fn arm_alarm(&self) {
+        // Datasheet says temp oversampling=1 makes a reading typically take 5.5ms.
+        // (Maximally 6.4ms).
+        let delay = self.alarm.ticks_from_us(6400);
+        self.alarm.set_alarm(self.alarm.now(), delay);
+    }
+
+    fn check_ready(&self, buffer: &'static mut [u8]) {
+        self.i2c.read(buffer, Register::STATUS, 1)
```
alexandruradovici, line 254: Please handle errors

alexandruradovici (changes requested):
The driver should never panic and errors should be handled as they might *lockup* your driver and it might be difficult to debug.

capsules/src/bmp280.rs
alexandruradovici: I think you can return `ErrorCode::NOMEM` if the TakeCell returns None.

capsules/src/bmp280.rs
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    fn handle_alarm(&self) {
+        match self.state.get() {
+            State::WaitingForAlarm(calibration) => self.buffer.take().map_or((), |buffer| {
+                self.check_ready(buffer);
+                self.state.set(State::Waiting(calibration))
+            }),
+            other => {
+                debug!("BMP280 received unexpected alarm in state {:?}", other);
```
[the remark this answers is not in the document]
dcz-self: Is there a compile-time switch to mark development builds? I would prefer to keep the debug messages at least when I'm testing. Comments aren't really a great way to do it.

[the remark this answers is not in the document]
alexandruradovici: We do tend to avoid using conditional compiling in Tock, as this introduces issues related to testing. Using flags prevents all the code to be compiled in all cases and would require test builds for all the possible feature combinations.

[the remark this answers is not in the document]
dcz-self: I opted for keeping the debug statements, improving the precedent of SHT3X. https://github.com/tock/tock/blob/master/capsules/src/sht3x.rs#L153

I haven't exercised a lot of the failure paths in the driver (and some outside), so I'm not confident enough to remove the flagging of bugs. If those early warnings are dropped silently, then finding the actual problems is going to be a real pain.

capsules/src/bmp280.rs
dcz-self: Done. That caused a bit of churn, especially in the i2c handler, where the handling would have added another 5 lines to each match branch, but I managed to keep the branches lean, and instead deal with the failures in one place.

I might have gone overboard a little though.

capsules/src/bmp280.rs
dcz-self: Done.

boards/components/src/bmp280.rs
dcz-self: Done.

capsules/src/bmp280.rs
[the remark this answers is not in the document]
alexandruradovici: The SHT3X driver is rather old and should not have those panics. 

My suggestion was to keep them there while you are debugging and comment them out when the driver is ready for production.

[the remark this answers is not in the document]
dcz-self: I appreciate the suggestion, but I think there is a significant overlap between production and "not guaranteed bug free" (at least in things I build) and that I will never feel comfortable removing the debugs.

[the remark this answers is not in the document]
alistair23: It would be best to report this error to the caller instead of using a `debug!()` call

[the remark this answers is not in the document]
dcz-self: I don't see any benefit of reporting this to the caller: this message is a compromise because I was asked to remove the panic, andit means "this driver (or OS) is buggy and does not function correctly". I struggle to see what the caller could do with this information.
This would also codify the driver being buggy as part of the driver API, which I haven't generally seen done, apart from the informal removing the resource and having the resource manager report a fatal error.
But Tock OS drivers don't seem to be making an allowance for themselves being buggy, so I think returning this to the caller would have been a departure.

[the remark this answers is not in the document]
lschuermann: > But Tock OS drivers don't seem to be making an allowance for themselves being buggy, so I think returning this to the caller would have been a departure.

This is a subjective rather than authoritative response, but I do think that we do and should acknowledge that drivers can and will be buggy. Fundamentally, drivers being buggy is also a significant motivation of introducing the partly-unprivileged capsule concept in the first place. How to report such issues is a difficult question, but I would side with reporting it to userspace as well: when a capsule manages to misbehave without causing an overall denial of service, resilient userspace applications should have _the ability_ to recognize that and handle this gracefully, in which ever way they choose.  A good representation of a generic, unexpected capsule error is `ErrorCode::FAIL`.

I think it might be fine to also issue a console message on debug builds. We generally try to reduce error reporting in form of human-readable output -- at least for release builds -- because it tends to put significant strain on our flash space requirements (hence also favor `unwrap()` over `expect()`, etc.). Returning a generic internal failure to userspace also _enables_ userspace to report a human-readable error to the console, if it wishes to do so.

capsules/src/bmp280.rs
```diff
@@ -0,0 +1,378 @@
...
+struct CalibrationData {
+    dig_t1: u16,
+    dig_t2: i16,
+    dig_t3: i16,
+    // TODO: pressure calibration
+}
+
+/// CAUTION: calibration data puts least significant byte in the lowest address,
+/// readouts do the opposite.
+fn twobyte(lsb: u8, msb: u8) -> u16 {
```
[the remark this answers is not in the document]
lschuermann: I disagree. For hinting the compiler to inline a given function out of performance considerations, the weak `#[inline]` attribute should be just fine. We should refrain from using `#[inline(always)]` and `#[inline(never)]` unless we have some empirical evidence this is necessary, because presumably the compiler has some good reason to disobey our recommendation.

capsules/src/bmp280.rs
```diff
@@ -0,0 +1,469 @@
...
+    dig_t3: i16,
+    // TODO: pressure calibration
+}
+
+/// CAUTION: calibration data puts least significant byte in the lowest address,
+/// readouts do the opposite.
+#[inline]
+fn twobyte(lsb: u8, msb: u8) -> u16 {
+    ((msb as u16) << 8) + lsb as u16
+}
```
lschuermann, line 66: I think this can be replaced with a call to `u16::from_be_bytes`.

capsules/src/bmp280.rs
```diff
@@ -0,0 +1,378 @@
...
+    Error,
+}
+
+/// Complies with the reading and writing protocol used by the sensor.
+struct I2cWrapper<'a> {
+    i2c: &'a dyn i2c::I2CDevice,
+}
+
+impl<'a> I2cWrapper<'a> {
+    fn write<const COUNT: usize>(
```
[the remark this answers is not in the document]
lschuermann: That's right. We must never "loose" our `&'static mut [u8]` buffer reference, hence passing it back. This method should check that the length of the provided buffer is sufficient.

capsules/src/bmp280.rs
```diff
@@ -0,0 +1,469 @@
...
+        &self,
+        buffer: &'static mut [u8],
+        addr: Register,
+        data: [u8; COUNT],
+    ) -> Result<(), (i2c::Error, &'static mut [u8])> {
+        buffer[0] = addr as u8;
+        for (i, d) in data.iter().enumerate() {
+            buffer[i + 1] = *d;
+        }
```
lschuermann, line 138: Perhaps you can use the more elegant [`copy_from_slice`](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice) here, with appropriate subslicing and length checks.

capsules/src/bmp280.rs
```diff
@@ -0,0 +1,469 @@
...
+
+    /// Requests a read into buffer.
+    /// Parse the result using `parse_read`.
+    fn read(
+        &self,
+        buffer: &'static mut [u8],
+        addr: Register,
+        count: u8,
+    ) -> Result<(), (i2c::Error, &'static mut [u8])> {
+        buffer[0] = addr as u8;
```
lschuermann, line 151: Please validate the length of the buffer to be at least 1.

capsules/src/bmp280.rs
```diff
@@ -0,0 +1,469 @@
...
+        self.i2c.write_read(buffer, 1, count)
+    }
+
+    fn disable(&self) {
+        self.i2c.disable()
+    }
+
+    fn parse_read(buffer: &[u8], count: u8) -> &[u8] {
+        &buffer[..(count as usize)]
+    }
```
lschuermann, line 162: I'm afraid I don't fully understand the purpose of this method. Is this supposed to eventually do more than just subslicing? If not, I'd avoid this layer of indirection and just the Index operator directly.

capsules/src/bmp280.rs
```diff
@@ -0,0 +1,469 @@
...
+    Write {
+        addr: Register,
+        data: u8,
+        fail_state: State,
+    },
+    Disable,
+}
+
+impl I2cOperation {
+    fn read(addr: Register, count: u8, fail_state: State) -> Self {
```
lschuermann, line 301: Given `I2cOperation` is module-private by default, and these methods don't do anything other than construct the enum variants with the passed arguments, you can just construct the variants below (you already do that for `I2cOperation::disable` anyways.

lschuermann (changes requested):
This is pretty cool, thank you for this contribution. I might actually have a BMP280 with which I can test this lying around somewhere... It looks pretty good overall, just some minor comments.

capsules/src/bmp280.rs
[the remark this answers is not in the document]
lschuermann: Actually, nevermind, I was confused by GitHub's weird suggestions and how they integrate into the review UI and didn't realize this was already addressed but not marked as `Outdated`.

capsules/src/bmp280.rs
dcz-self: It was there to have a single place with an intepretation of how the buffer should be parsed. I had to change it a couple of times as I built up my understanding, and it was annoying to change it everywhere. It forms a complete, coherent application layer on top of i2c under the I2cWrapper struct.

I guess we don't expect this to change in this driver, so it could be thrown away, but I think it's valuable if anyone wants to learn from this layer as a basis for another driver, it would be useful to keep it.

capsules/src/bmp280.rs
dcz-self: I added those as a syntax sugar. Without those,
```
I2cOperation::read(Register::TEMP_MSB, 3, State::Idle(calibration));
```

becomes
```
I2cOperation::Read(addr: Register::TEMP_MSB, count: 3, state: State::Idle(calibration));
```

I don't have a better reason to keep those, other than that it will be annoying to write the pressure measurement code without them.

bradjc (approved):

capsules/src/bmp280.rs
dcz-self: Done.

capsules/src/bmp280.rs
dcz-self: Done.
I decided to skip length checks. At the moment I'm not ready to have indexing errors propagate all the way up and push the driver into a buggy state. This is the same situation as in the thread about buggy drivers above, and handling that correctly needed a serious update.

On the bright side, all the i2c read/write code paths are utilized before a measurement can be made, so they are flawless index-wise at the moment.

dcz-self:
Updated for the most recent round of feedback.

alistair23 (approved):

bradjc (approved):

bradjc:
@lschuermann

capsules/src/bmp280.rs
lschuermann: > I guess we don't expect this to change in this driver, so it could be thrown away, but I think it's valuable if anyone wants to learn from this layer as a basis for another driver, it would be useful to keep it.

Okay, I now understand the basic intention, I was just wondering why you'd want a `parsing` function just for subslicing, but it's probably a coincidence that parsing is effectively just a subslicing of the buffer in for this device and there is no more complex encapsulation / message layer. I'm fine with keeping it.

lschuermann (approved):
Sorry for the long wait. I'm working with some tight deadlines right now so I didn't get a chance to test it on proper hardware but it looks good to me so far.

hudson-ayers:
bors r+

We can always submit fixes once Leon gets a chance to test this, but given that it has been tested on hardware by the author I see no reason not to get this working code into the repository.

dcz-self:
Awesome, thanks!

PR #2972, opened 2022-02-18
RFC: Moving some flags from Makefile to build.rs

bradjc:
### Pull Request Overview

This pull request is a companion to #2962 and looks at what moving more flags from the Makefile to a common build.rs file could look at.

This is an initial implementation to get feedback. The [cargo build.rs instructions](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script) are somewhat limited, so there isn't too much we can move (at least right now). 

I'm targeting build.rs because it is not a hidden file, and because we can use a shared common_build.rs.

### Testing Strategy

todo

### TODO or Help Wanted

Since we can't move too much out of Makefile right now, should we bother? Or is shrinking our custom Makefile and using standard rust infrastructure still a net improvement?

If we find a version we like it needs to be propagated to all boards.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/boards/Makefile.common
+++ b/boards/Makefile.common
@@ -41,21 +41,10 @@ TARGET_DIRECTORY ?= $(TOCK_ROOT_DIRECTORY)target/
 # This will hopefully move into Cargo.toml (or Cargo.toml.local) eventually.
 #
 # - `relocation-model=static`: See https://github.com/tock/tock/pull/2853
-# - `-nmagic`: lld by default uses a default page size to align program
-#   sections. Tock expects that program sections are set back-to-back. `-nmagic`
-#   instructs the linker to not page-align sections.
-# - `-icf=all`: Identical Code Folding (ICF) set to all. This tells the linker
-#   to be more aggressive about removing duplicate code. The default is `safe`,
-#   and the downside to `all` is that different functions in the code can end up
-#   with the same address in the binary. However, it can save a fair bit of code
-#   size.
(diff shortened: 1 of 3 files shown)


ppannuto:
Is there any sense of priority in the Rust community of getting more options added to the `build.rs + cargo:` interface?

It feels like this move doesn't really buy us anything at this stage, and it's a minor regression to have half the options set in one place, and half in another. We could add 'tracking comments' to the flags in the Makefile perhaps, something like each line being `-C link-arg=-Tlayout.ld \   # println!("cargo:rustc-link-arg=-Tlayout.ld");` so that once every config has an equivalent, we can switch over easily?

Conceptually, I like the idea of eventually getting to a build that is rust-native, but driven by make as a glorified shell scripts that ease use.

bradjc:
> Is there any sense of priority in the Rust community of getting more options added to the `build.rs + cargo:` interface?

I don't know. I'm guessing that since they are OK with their build configurations sequestered in hidden files they are in no hurry to support more general build.rs files. But you never know.

Also, does it change anything if you consider #2962 too? That one does have the "this should move" flag in the makefile.

hudson-ayers:
I think there are some other small upsides to this change with regards to the amount of rust code that gets rebuilt when manually changing linker args, but I think those upsides are not worth having half the flags set in one place and half in another.

bradjc:
Same issue as #2962.

PR #2976, opened 2022-02-18
process_console: Add missing '\r' on command output strings

gustavonihei:
### Pull Request Overview

This PR intends to fix some of the command output strings of Process Console which are currently missing the carriage return character.

**Before**
```bash
tock$ status
Total processes: 0
                  Active processes: 0
                                     Timeslice expirations: 0
                                                             tock$ help
Welcome to the process console.
                               Valid commands are: help status list stop start fault process kernel
                                                                                                   tock$ 
tock$ 
```
**After**
```bash
tock$ status
Total processes: 0
Active processes: 0
Timeslice expirations: 0
tock$ help
Welcome to the process console.
Valid commands are: help status list stop start fault process kernel
tock$ 
```

### Testing Strategy

Tested with the ESP32-C3-DevKitM-1 development board and executed some of the available commands on the Process Console.
Validated with visual inspection.

### TODO or Help Wanted

Not applicable.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/process_console.rs
+++ b/capsules/src/process_console.rs
@@ -326,9 +326,10 @@ impl<'a, A: Alarm<'a>, C: ProcessManagementCapability> ProcessConsole<'a, A, C> 
         );
         let _ = self.write_bytes(&(console_writer.buf)[..console_writer.size]);
 
-        let _ = self.write_bytes(b"Welcome to the process console.\n");
-        let _ = self
-            .write_bytes(b"Valid commands are: help status list stop start fault process kernel\n");
+        let _ = self.write_bytes(b"Welcome to the process console.\r\n");
+        let _ = self.write_bytes(
+            b"Valid commands are: help status list stop start fault process kernel\r\n",
+        );
         self.prompt();
     }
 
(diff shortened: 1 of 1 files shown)


alexandruradovici:
I had this issue and solved by setting the terminal to use CRLF for \n. 

To be consistent with all the strings in Tock (and most of the OSes), I think we should leave the \n.

jrvanwhy:
As far as I am aware, `\r\n` line endings have died out in the modern world except for one use case: UNIX terminal I/O in raw mode.

So the question I think we need to answer is "Is Tock's UART interface a UNIX raw terminal interface, or is it something else?". I don't know the correct answer to that question.

When I wrote `LowLevelDebug`, I didn't know that `\r\n` line endings were still used *anywhere*, hence my use of `\n` line endings. When I implemented libtock-rs' cargo runner (https://github.com/tock/libtock-rs/pull/365), I learned what raw mode is and that raw mode still uses `\r\n` line endings. At the time, I assumed that it was correct for Tock to use `\n` line endings, and I made the runner translate line endings itself.

Note that raw mode isn't primarily about changing line modes: its primary action is to remove all the buffering and preprocessing the terminal normally does to its input. Does Tock want to do that buffering and editing itself, or should that be done by the terminal?

ppannuto:
I don't think we want to handle all of the display management; in particular, line buffering should definitely be a client-side decision.

However, as evidenced by this PR, there are still plenty of serial terminals out that that expect `\r\n`. Tock does put out a pretty serious effort to be user-friendly, and printing 'nice' for any random serial terminal falls into that bin in my mind.

Besides the obvious 1-byte overhead, is there any harm to using `\r\n`?

I would also be willing to discuss special-casing the ProcessConsole as the only place where we prioritize `\r\n`, as it's the interface that is mostly likely to be used by folks use odd or less-capable serial terminals.

thulithwilfred (approved):

alistair23 (approved):

PR #2977, opened 2022-02-21
kernel/processbuffer: migrate ProcessSlices towards raw pointers

### LINKED ISSUE #2882: Unsoundness transmuting &[u8] to &[Cell<u8>] when handling read-only memory

lschuermann:
### Pull Request Overview

This commit migrates the ProcessSlice interface in Tock to work on raw pointers instead of slices of `Cell`s. Most notably, this prevents usage of Rust's `Index` operator, which has been replaced by a `ProcessSliceIndex` trait and `get` method. Furthermore, this commit addresses

- removal of all panicing methods in relation to process buffers / process slices,

- documentation of the new safety requirements and invariants for the raw-pointer based process slice interface,

- revision of previous documentation, alignment to 80 char line width,

- adaption of capsules to this new interface, in the form of a straightforward translation. This adds a substantial amount of `unwrap()` calls in capsules, and does not apply trivial cleanups such as converting manually implemented for-loops for copying between slices with the appropriate `copy_from_slice` / `copy_to_slice` calls. This is done for easy validation of the proposed changes. Subsequent commits should clean up this code, and demonstrate an implementation with reduced panics on select capsules.

All of the above are deeply interwoven, thus they could not be separated into different commits.

##### Motivation

In short, the previous interface was unsound. Because fundamentally processes may allow overlapping memory regions to the kernel, and (mutable) aliasing of memory is unsound when used with Rust primitives (such as `&[u8]` slice references), the previous approach used a construct which allows mutable aliasing of memory: instead of holding onto a slice of bytes, it provided a slice of `Cell`s of bytes. Containing an `UnsafeCell`, these constructs allow mutable aliasing (that is, holding multiple mutable references to their contents simultaneously).

This approach has two issues: because `Cell` is a construct allowing mutation of the underlying value, it is unsound to use it over hardware-immutable memory. The exact consequences of writing to read-only memory are implementation dependant. For potentially hardware read-only memory, the previous implementation used a wrapper around `Cell`, exposing only a subset of the operations possible through `Cell`, explicitly excluding operations which mutate the underlying value. Nonetheless, wrapping such memory in `Cell` is unsound.

Furthermore, it is desirable to be able to treat immutable Rust slices (`&[u8]`) as `ReadableProcessBuffer`s to pass both through common interfaces. However, transmuting an immutable `&[u8]` Rust slice into `&[Cell<u8>]` is an unsound operation, for the same reasons as outlined above. For more information, refer to [1].

##### Approach

This commit changes the underlying structure of `ReadableProcessSlice` and `WriteableProcessSlice` to no longer be a `#[repr(transparent)]` wrapper over a slice of `Cell`s, but instead store a non-null pointer and a length to describe a given shared memory region. This is fundamentally identical to the in-memory representation of slice references. However, based on this information, raw pointer-based methods can be used to dereference the covered memory, for which Rust's aliasing rules do not apply as long as they can not access other Rust allocations.

At the same time, it introduces a significant amount of unsafe code, as many convenience methods of Rust slices can no longer be used. Thus this code requires especially careful vetting.

Fixes #2882.

[1]: https://github.com/tock/tock/issues/2882

### Testing Strategy

This pull request was tested by CI (the LiteX simulation tests a few allow operations). Needs much more testing, also in the form of unit tests.

### TODO or Help Wanted

This pull request still needs
- cleanup of some capsule code. Currently, the process slice accesses have a lot of `unwrap()` calls. While this PR can't get rid of any panics, I'm going to do at least one pass over the affected capsules and see what can be trivially optimized. For instances, many manually implemented for-loops can be replaced by their `copy_from_slice` / `copy_to_slice` equivalents.
- converting one capsule to avoid usage of `unwrap()` (probably console). This can demonstrate how the new API enables and encourages writing panic-free or panic-reduced capsule code.
- Extensive testing.
- Vetting of all new unsafe code introduced.

This code contains some TODOs regarding future code removals. I plan to follow this PR with one which cleans up some aspects of the ProcessBuffer kernel-internal interfaces, as there are some leftovers from the migration to kernel-managed allows. I wanted to avoid polluting this PR, which should be limited to only the process slice changes, to allow careful review of those.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/processbuffer.rs
+++ b/kernel/src/processbuffer.rs
@@ -1,17 +1,17 @@
 //! Data structures for passing application memory to the kernel.
 //!
-//! A Tock process can pass read-write or read-only buffers into the
-//! kernel for it to use. The kernel checks that read-write buffers
-//! exist within a process's RAM address space, and that read-only
-//! buffers exist either within its RAM or flash address space. These
-//! buffers are shared with the allow_read_write() and
-//! allow_read_only() system calls.
+//! A Tock process can pass read-write or read-only buffers into the kernel for
+//! it to use. The kernel checks that read-write buffers exist within a
+//! process's RAM address space, and that read-only buffers exist either within
+//! its RAM or flash address space. These buffers are shared with the
+//! `allow_readwrite()` and `allow_readonly()` system calls. Refer to [TRD 104
+//! -- Syscalls][1] for more information.
 //!
-//! A read-write and read-only call is mapped to the high-level Rust
-//! types [`ReadWriteProcessBuffer`] and [`ReadOnlyProcessBuffer`]
-//! respectively. The memory regions can be accessed through the
-//! [`ReadableProcessBuffer`] and [`WriteableProcessBuffer`] traits,
-//! implemented on the process buffer structs.
+//! A read-write and read-only call is mapped to the high-level Rust types
+//! [`ReadWriteProcessBuffer`] and [`ReadOnlyProcessBuffer`] respectively. The
+//! memory regions can be accessed through the [`ReadableProcessBuffer`] and
+//! [`WriteableProcessBuffer`] traits, implemented on the process buffer
+//! structs.
 //!
 //! Each access to the buffer structs requires a liveness check to ensure that
 //! the process memory is still valid. For a more traditional interface, users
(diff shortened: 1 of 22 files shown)


bradjc:
I added a commit for an effective but maybe not elegant way to remote `unwrap()` from console.

bradjc:
While working on the change to console and thinking about the grant/processbuffer API, as well as trying to reason about what would happen if a process went away in between transmitting console buffers and if using `unwrap_or(0,...)` is actually ok (I'm pretty sure it is), I think we currently check for process liveness _twice_, once when entering the grant, and once when entering the process buffer. We shouldn't have to duplicate the check (i.e. `process_map_or()` in processbuffer.enter()) since once we have entered the grant we are guaranteeing the process won't go away until after the grant is exited.

Not sure if this is actually relevant to this PR, but wanted to note that.

lschuermann:
> I think we currently check for process liveness _twice_, once when entering the grant, and once when entering the process buffer. We shouldn't have to duplicate the check (i.e. `process_map_or()` in processbuffer.enter()) since once we have entered the grant we are guaranteeing the process won't go away until after the grant is exited.

That's right, I had that exact same thought. I think we should most definitely think about this, and how to restructure some of the handling of ProcessBuffers given optimizations made possible by moving them into the Grant kernel resources area. I have some WIP code for that locally.

However, I believe that these changes, which are mostly limited to ProcessSlices, are complex enough on their own and warrant a close review without worring about throwing over other parts of the kernel, hence this isolated PR. :smiley:

bradjc:
I'm in favor of this change, and am working on converting a capsule to use slice copy functions rather than manual for loops.

lschuermann:
> I'm in favor of this change, and am working on converting a capsule to use slice copy functions rather than manual for loops.

Thanks! I'll also work on cleaning up a few capsules over the next days. I've also pushed a local commit which does a few (hopefully) correct transitions to `copy_from_slice` / `copy_to_slice` in capsule code that I had lying around, hope that doesn't collide with your change @bradjc.

lschuermann:
@bradjc I've just pushed a transition of the app flash driver towards the iterator-based copy approach we've discussed. It seems strictly better than dealing with the Result-returning indexing methods, although I realize it might be confusing at first. The "tricky" part is that, when we need to know how many bytes we've copied, we need to extract that data from the iteration, which is going to involve an `enumerate()` and `fold()` or just a `fold()`. For simply copying until the lower bound of the two buffers, when we don't need to know how much we've copied, the `for_each()` method on the zipped iterator is more elegant. Both approaches completely avoid panics and error handling.

bradjc:
I added a few more commits.

kernel/src/processbuffer.rs
```diff
@@ -519,81 +456,155 @@ impl ReadableProcessBuffer for ReadWriteProcessBuffer {
         }
     }
 
-    fn enter<F, R>(&self, fun: F) -> Result<R, process::Error>
+    fn enter<'a, F, R>(&'a self, fun: F) -> Result<R, process::Error>
     where
-        F: FnOnce(&ReadableProcessSlice) -> R,
+        F: FnOnce(ReadableProcessSlice<'a>) -> R,
```
kupiakos, line 461: This changes the lifetime behavior of `enter`. Previously the arg's lifetime was that of the function call, now it's the same as `&self`. If this is sound, and I'm not positive it is, we could instead be using a `fn get(&self) -> ReadableProcessSlice<'a>`. Considering it was switched to `enter` from a getter previously, what was the reasoning behind that?

lschuermann:
I'm going to rebase this and work through the remaining capsules over the next few days.

lschuermann:
It appears that under the new "tree borrows" model, our current implementation is legal: https://github.com/rust-lang/unsafe-code-guidelines/issues/303#issuecomment-1696984148

Super happy that we won't have to ship this proposed solution!

PR #2980, opened 2022-02-25
chips: move arch config to function level

bradjc:
### Pull Request Overview

In tock we use cfg()s based on architecture to hide assembly code when cargo builds tests. For consistency and clarity this PR moves a couple cfgs to the function level so we are not modifying the body of functions using compiler flags. Also, if there is some use case for running chip-specific code on host this will make it clear when certain functions are unimplemented.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/rp2040/src/clocks.rs
+++ b/chips/rp2040/src/clocks.rs
@@ -1038,11 +1038,11 @@ impl Clocks {
         (((source_freq as u64) << 8) / freq as u64) as u32
     }
 
+    #[cfg(all(target_arch = "arm", target_os = "none"))]
     #[inline]
     fn loop_3_cycles(&self, clock: Clock) {
         if self.get_frequency(clock) > 0 {
             let _delay_cyc: u32 = self.get_frequency(Clock::System) / self.get_frequency(clock) + 1;
-            #[cfg(target_arch = "arm")]
             unsafe {
                 asm! (
                     "1:",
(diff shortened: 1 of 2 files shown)


lschuermann (approved):
This makes sense to me. I like the function-local imports for traits which are only required when building for proper target platform.

hudson-ayers (approved):

alistair23 (approved):

thulithwilfred (approved):

chips/rp2040/src/clocks.rs
```diff
@@ -1038,11 +1038,11 @@ impl Clocks {
         (((source_freq as u64) << 8) / freq as u64) as u32
     }
 
+    #[cfg(all(target_arch = "arm", target_os = "none"))]
     #[inline]
     fn loop_3_cycles(&self, clock: Clock) {
         if self.get_frequency(clock) > 0 {
             let _delay_cyc: u32 = self.get_frequency(Clock::System) / self.get_frequency(clock) + 1;
```
lschuermann, line 1045: One thing to watch out for with these changes is that they can cause `rustc` to conclude functions are unused, depending on the target architecture. I've manually validated this doesn't seem to happen here, and the solution is pretty trivial: just add the offending function calls to the stub implementation as well.

bors r+

Issue #2982, opened 2022-03-01
Cannot use `tock-registers` on the stable release channel


dawsonc623:
I am trying my hand at some embedded Rust development on a Raspberry Pi 4B, and one of my dependencies (`cortex-a`) depends on this library. My toolchain uses the stable release channel, and when trying to compile my project I get `error[E0554]: '#![feature]' may not be used on the stable release channel`. Is this by design, or is this library intended to work for devs using the stable release channel?

alexandruradovici:
Can you point us to the exact crate that you are using and the crate (library) that it depends on from this repository?

jrvanwhy:
> Can you point us to the exact crate that you are using and the crate (library) that it depends on from this repository?

I checked `cortex-a`'s `Cargo.toml`, and it depends on `tock-registers`. It may be easy to make `tock-registers` work on stable Rust (I might send a PR today).

jrvanwhy:
I edited the issue title to mention `tock-registers`, so that Tock developers understand at a glance which library this issue refers to.

The overarching issue for making Tock build on stable Rust is https://github.com/tock/tock/issues/1654.

dawsonc623:
Sorry, I should have dug a little deeper to give a better description (I did not realize this was a mono-repo and how big the scope of the project is). Thanks for making that clearer. I am not the most advanced Rust developer out there (especially in the embedded space), but if there is anything I can help with to get unblocked I am happy to try.

jrvanwhy:
PR opened: https://github.com/tock/tock/pull/2983

The specific unstable feature in use is likely to be stabilized soon, and the PR I opened has some drawbacks, so it's possible that we will not merge the PR.

jrvanwhy:
Making `tock-registers` work on stable Rust will not solve your issue on its own, because `cortex-a` depends on unstable Rust as well: https://github.com/rust-embedded/cortex-a/blob/fd6688419ac9884abd1d79a18b3ff64b4776e1f4/src/lib.rs#L82. It'll just move the compilation error from the `tock-registers` crate to the `cortex-a` crate.

dawsonc623:
Hmmm, okay. Well, thanks for your work on this all the same. I will open an issue with `cortex-a` next and see what they say.

brghena:
@dawsonc623 One thing to check, do you actually need to be on stable, or could you just move your toolchain to a nightly version of rust?

dawsonc623:
[...]

PR #2983, opened 2022-03-01
Move `tock-registers` to stable Rust by removing some trait bounds.

jrvanwhy:
### Pull Request Overview

`tock-registers` depends on `feature(const_fn_trait_bound)`. This PR removes the dependency on `feature(const_fn_trait_bound)` by removing trait bounds on types with `const` functions.

I see two downsides to this PR:

1. Users can now use the `Field`, `FieldValue`, `InMemoryRegister`, and `LocalRegisterCopy` types with arbitrary generic arguments (in which case they will have no methods other than `new`).
2. Re-adding these trait bounds is a breaking change.

### Testing Strategy

`cargo check` and `make prepush`

### TODO or Help Wanted

**Question:** are the drawbacks of this PR worth the benefit of supporting stable Rust in `tock-registers`? Note that it looks like `const_fn_trait_bound` will likely be stabilized soon (i.e. Rust 1.61): https://github.com/rust-lang/rust/pull/93827

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/libraries/tock-register-interface/src/fields.rs
+++ b/libraries/tock-register-interface/src/fields.rs
@@ -73,13 +73,17 @@ use crate::{RegisterLongName, UIntLike};
 /// Specific section of a register.
 ///
 /// For the Field, the mask is unshifted, ie. the LSB should always be set.
-pub struct Field<T: UIntLike, R: RegisterLongName> {
+///
+/// `T` should be a `UIntLike` and `R` should be a `RegisterLongName`. Because
+/// `new` is a `const` function, we cannot add type constraints for `T` and `R`
+/// until `feature(const_fn_trait_bound)` is stabilized.
+pub struct Field<T, R> {
     pub mask: T,
     pub shift: usize,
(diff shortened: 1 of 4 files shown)


jrvanwhy:
Relevant issue: https://github.com/tock/tock/issues/2982.

alistair23:
I would say we should hold off and continue using `const_fn_trait_bound`

bradjc:
I think this PR is great to have as a merge-if-other-nightly-features-become-stable fall back plan so we can switch to stable.

But it seems like the stabilization for this feature is super close? Seems worth it to wait.

lschuermann:
While I am excited for `tock-registers` to become usable with stable Rust, as that means we should be able to split it out into a different repository in the tock GitHub organization without having to worry about compiler compatibility any longer, I agree we should try to wait a little. It really does seem like `const_fn_trait_bound` is right around the corner.

PR #2985, opened 2022-03-07
AES: Add HIL for AES GCM

alistair23:
### Pull Request Overview

AES GCM is a relatively new mode of AES, formalised by NIST in 2007. It is becoming widely adopted due to it's performance improvements.

GCM is now used in a variety of IEEE standards, including 702.11ad, as well as SSH, TLS and OpenVPN.
    
Due to it's growing popularity let's add support for it to Tock.

### Testing Strategy

This was tested on OpenTitan using a software implementation of GCM. You can see the implementation here: https://github.com/alistair23/tock/tree/alistair/aes-gcm

Note that in https://github.com/alistair23/tock/tree/alistair/aes-gcm we use an external hash crypto library for the operation, which is why it isn't being sent as a PR.

### TODO or Help Wanted

Nothing, the HIL and capsule are ready to go

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/kernel/src/hil/symmetric_encryption.rs
+++ b/kernel/src/hil/symmetric_encryption.rs
@@ -139,3 +139,40 @@ pub trait AES128CCM<'a> {
         encrypting: bool,
     ) -> Result<(), (ErrorCode, &'static mut [u8])>;
 }
+
+pub trait GCMClient {
+    /// `res` is Ok(()) if the encryption/decryption process succeeded. This
+    /// does not mean that the message has been verified in the case of
+    /// decryption.
+    /// If we are encrypting: `tag_is_valid` is `true` iff `res` is Ok(()).
+    /// If we are decrypting: `tag_is_valid` is `true` iff `res` is Ok(()) and the
+    /// message authentication tag is valid.
+    fn crypt_done(&self, buf: &'static mut [u8], res: Result<(), ErrorCode>, tag_is_valid: bool);
+}
+
+pub trait AES128GCM<'a> {
+    /// Set the client instance which will receive `crypt_done()` callbacks
+    fn set_client(&'a self, client: &'a dyn GCMClient);
+
+    /// Set the key to be used for GCM encryption
+    /// Returns `INVAL` if length is not `AES128_KEY_SIZE`
+    fn set_key(&self, key: &[u8]) -> Result<(), ErrorCode>;
+
+    /// Set the IV to be used for GCM encryption. The IV should be less
+    /// or equal to 12 bytes (96 bits) as recommened in NIST-800-38D.
+    /// Returns `INVAL` if length is greater then 12 bytes
+    fn set_iv(&self, nonce: &[u8]) -> Result<(), ErrorCode>;
+
+    /// Try to begin the encryption/decryption process
+    /// The possible ErrorCodes are:
+    ///     - `BUSY`: An operation is already in progress
+    ///     - `SIZE`: The offset and lengths don't fit inside the buffer
+    fn crypt(
+        &self,
+        buf: &'static mut [u8],
+        aad_offset: usize,
+        message_offset: usize,
+        message_len: usize,
+        encrypting: bool,
+    ) -> Result<(), (ErrorCode, &'static mut [u8])>;
+}
--- a/capsules/src/test/aes_gcm.rs
+++ b/capsules/src/test/aes_gcm.rs
@@ -1,0 +1,243 @@
+//! Test the AES GCM implementation on top of AES hardware.
+
+use core::cell::Cell;
+use kernel::debug;
+use kernel::hil::symmetric_encryption::{GCMClient, AES128GCM, AES128_KEY_SIZE};
+use kernel::utilities::cells::TakeCell;
+use kernel::ErrorCode;
+
+pub struct Test<'a, A: AES128GCM<'a>> {
+    aes_gcm: &'a A,
+
+    buf: TakeCell<'static, [u8]>,
+    current_test: Cell<usize>,
+    encrypting: Cell<bool>,
+
+    // (key, iv, pt, aad, ct, tag)
+    tests: [(
+        &'static [u8],
+        &'static [u8],
+        &'static [u8],
+        &'static [u8],
+        &'static [u8],
+        &'static [u8],
+    ); 3],
+}
+
+impl<'a, A: AES128GCM<'a>> Test<'a, A> {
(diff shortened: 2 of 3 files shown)


kernel/src/hil/symmetric_encryption.rs
```diff
@@ -139,3 +139,35 @@ pub trait AES128CCM<'a> {
...
+
+pub trait AES128GCM<'a> {
+    /// Set the client instance which will receive `crypt_done()` callbacks
+    fn set_client(&'a self, client: &'a dyn GCMClient);
+
+    /// Set the key to be used for GCM encryption
+    fn set_key(&self, key: &[u8]) -> Result<(), ErrorCode>;
+
+    /// Set the IV to be used for GCM encryption. The IV should be less
+    /// then 12 bytes (96 bites) as recommened in NIST-800-38D.
```
thulithwilfred, line 161: Should be  ‘bits’ here?

thulithwilfred (approved):
Looks good

kernel/src/hil/symmetric_encryption.rs
alistair23: It should be, fixed

thulithwilfred (dismissed):

kernel/src/hil/symmetric_encryption.rs
```diff
@@ -139,3 +139,35 @@ pub trait AES128CCM<'a> {
...
+    /// message authentication tag is valid.
+    fn crypt_done(&self, buf: &'static mut [u8], res: Result<(), ErrorCode>, tag_is_valid: bool);
+}
+
+pub trait AES128GCM<'a> {
+    /// Set the client instance which will receive `crypt_done()` callbacks
+    fn set_client(&'a self, client: &'a dyn GCMClient);
+
+    /// Set the key to be used for GCM encryption
+    fn set_key(&self, key: &[u8]) -> Result<(), ErrorCode>;
```
phil-levis, line 158: Why is the key a slice if it's AES 128?

phil-levis: What are the allowed return values/errors and what triggers them?

kernel/src/hil/symmetric_encryption.rs
```diff
@@ -139,3 +139,35 @@ pub trait AES128CCM<'a> {
...
+
+    /// Try to begin the encryption/decryption process
+    fn crypt(
+        &self,
+        buf: &'static mut [u8],
+        aad_offset: usize,
+        message_offset: usize,
+        message_len: usize,
+        encrypting: bool,
+    ) -> Result<(), (ErrorCode, &'static mut [u8])>;
```
phil-levis, line 172: What are the allowed error results and what triggers them?

kernel/src/hil/symmetric_encryption.rs
```diff
@@ -139,3 +139,35 @@ pub trait AES128CCM<'a> {
...
+pub trait AES128GCM<'a> {
+    /// Set the client instance which will receive `crypt_done()` callbacks
+    fn set_client(&'a self, client: &'a dyn GCMClient);
+
+    /// Set the key to be used for GCM encryption
+    fn set_key(&self, key: &[u8]) -> Result<(), ErrorCode>;
+
+    /// Set the IV to be used for GCM encryption. The IV should be less
+    /// then 12 bytes (96 bits) as recommened in NIST-800-38D.
+    fn set_iv(&self, nonce: &[u8]) -> Result<(), ErrorCode>;
```
phil-levis, line 162: I believe its less than or equal to 12 bytes, not less than 12 bytes?

phil-levis: What are the allowed errors and what triggers them?

phil-levis (changes requested):

phil-levis:
I thought GCM is popular because it's a parallelizable authenticated encryption that's not encumbered by patents? 

When Dan Boneh teaches his crypto class, when he gets to AES in symmetric encryption, he requires everyone to sign the AES pledge. Slide 7: https://crypto.stanford.edu/~dabo/cs255/lectures/PRP-PRF.pdf

I think it's worth considering that here.

kernel/src/hil/symmetric_encryption.rs
alistair23: It takes a slice to match all of the other `set_key()` AES functions

alistair23:
Ping!

alistair23:
Ping!

kernel/src/hil/symmetric_encryption.rs
phil-levis: Shouldn't this be SIZE, not INVAL?

kernel/src/hil/symmetric_encryption.rs
phil-levis: Shouldn't that be SIZE?

phil-levis: It looks like the comment on SIZE is "too large", but I think that was because it didn't consider operations like this. It seems like SIZE should be more general than that. Otherwise, > 12 means SIZE and < 12 means INVAL.

kernel/src/hil/symmetric_encryption.rs
phil-levis: Can you explain why you think it should be INVAL?

alistair23: Yeah, so the ErrorCode documentation says:

```rust
    /// An invalid parameter was passed
    INVAL = 6,
    /// Parameter passed was too large
    SIZE = 7,
```

As we aren't trying to store anything in a buffer I don't think it's a `SIZE` error (the buffer isn't too small), but more that the user passed an incorrect key

kernel/src/hil/symmetric_encryption.rs
alistair23: Similar to above, we aren't trying to store anything in a buffer I don't think it's a `SIZE` error (the buffer isn't too small), but more that the user passed an incorrect IV

kernel/src/hil/symmetric_encryption.rs
phil-levis: OK, that makes sense to me.

phil-levis (approved):

kernel/src/hil/symmetric_encryption.rs
phil-levis: I agree -- that makes sense.

PR #2986, opened 2022-03-07
OpenTitan: Some SPI Host fixes

alistair23:
### Pull Request Overview

This PR includes a few fixes for the SPI host when running on hardware.

- Update the QEMU commit SHA used to avoid accesses errors for the spi addresses
- Remove the feature `skip` check
- Clear interrupts before we enable them
- Revert back to GCC's Objcopy to support the features that we want

### Testing Strategy

Run on the CW310

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/earlgrey-cw310/run.sh
+++ b/boards/opentitan/earlgrey-cw310/run.sh
@@ -24,8 +24,8 @@ if [[ "${VERILATOR}" == "yes" ]]; then
 		--meminit=flash,./"$BUILD_DIR"/binary.64.vmem \
 		--meminit=otp,${OPENTITAN_TREE}/build-out/sw/device/otp_img/otp_img_sim_verilator.vmem
 elif [[ "${OPENTITAN_TREE}" != "" ]]; then
-	${OBJCOPY} --update-section .apps=${APP} ${1} bundle.elf
-	${OBJCOPY} --output-target=binary bundle.elf binary
+	riscv64-linux-gnu-objcopy --update-section .apps=${APP} ${1} bundle.elf
+	riscv64-linux-gnu-objcopy --output-target=binary bundle.elf binary
 	${OPENTITAN_TREE}/util/fpga/cw310_loader.py --firmware binary
 else
 	../../../tools/qemu/build/qemu-system-riscv32 -M opentitan -bios ../../../tools/qemu-runner/opentitan-boot-rom.elf -nographic -serial stdio -monitor none -semihosting -kernel "${1}"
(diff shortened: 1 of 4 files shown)


alistair23:
@thulithwilfred

chips/lowrisc/src/spi_host.rs
```diff
@@ -358,8 +358,11 @@ impl SpiHost {
 
     /// Enable both event/err IRQ
     fn enable_interrupts(&self) {
-        let regs = self.registers;
-        regs.intr_enable
+        self.registers
+            .intr_state
+            .write(intr::ERROR::SET + intr::SPI_EVENT::SET);
```
thulithwilfred, line 363: Why do we need to set `INTR_STATE` here? Did you mean to `clear`?

boards/opentitan/earlgrey-cw310/run.sh
```diff
@@ -24,8 +24,8 @@ if [[ "${VERILATOR}" == "yes" ]]; then
 		--meminit=flash,./"$BUILD_DIR"/binary.64.vmem \
 		--meminit=otp,${OPENTITAN_TREE}/build-out/sw/device/otp_img/otp_img_sim_verilator.vmem
 elif [[ "${OPENTITAN_TREE}" != "" ]]; then
-	${OBJCOPY} --update-section .apps=${APP} ${1} bundle.elf
-	${OBJCOPY} --output-target=binary bundle.elf binary
+	riscv64-linux-gnu-objcopy --update-section .apps=${APP} ${1} bundle.elf
```
thulithwilfred, line 27: We could probably do the same for the Verilator and get rid of OBJCOPY altogether?

thulithwilfred (changes requested):

chips/lowrisc/src/spi_host.rs
alistair23: The [`spi_host.INTR_STATE`](https://docs.opentitan.org/hw/ip/spi_host/doc/#register-table) register is `rw1c` which means that writing 1 to it will clear the bit

boards/opentitan/earlgrey-cw310/run.sh
alistair23: If the LLVM one works it's probably best to use that (I just can't get it working here)

chips/lowrisc/src/spi_host.rs
thulithwilfred: ah right!

thulithwilfred (approved):

bradjc (approved):

PR #2989, opened 2022-03-09
docs: Clarify allow usage

dcz-self:
Includes a fix to the docs of the console driver.

### Pull Request Overview

This pull request fixes pieces of documentation regarding the allow syscall.

### Testing Strategy

This pull request was tested by manual reading.

### TODO or Help Wanted

A future update should describe the `UserspaceReadableAllow` syscall. I don't know anything about it, so I skipped it.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/reference/trd104-syscalls.md
+++ b/doc/reference/trd104-syscalls.md
@@ -5,10 +5,10 @@ System Calls
 **Working Group:** Kernel<br/>
 **Type:** Documentary<br/>
 **Status:** Draft <br/>
-**Author:** Hudson Ayers, Guillaume Endignoux, Jon Flatley, Philip Levis, Amit Levy, Leon Schuermann, Johnathan Van Why <br/>
+**Author:** Hudson Ayers, Guillaume Endignoux, Jon Flatley, Philip Levis, Amit Levy, Leon Schuermann, Johnathan Van Why, dcz <br/>
 **Draft-Created:** August 31, 2020<br/>
-**Draft-Modified:** November 23, 2021<br/>
-**Draft-Version:** 6<br/>
+**Draft-Modified:** March 9, 2022<br/>
+**Draft-Version:** 7<br/>
 **Draft-Discuss:** tock-dev@googlegroups.com</br>
 
 Abstract
@@ -245,7 +245,7 @@ call drivers, which can be added and removed in different kernel
 builds. The full set of valid system calls a kernel supports therefore
 depends on what system call drivers it has installed.
 
-The 6 classes are:
+The 7 classes are:
 
 | Syscall Class    | Syscall Class Number |
 |------------------|----------------------|
@@ -518,10 +518,16 @@ on RISC-V.
 | Argument         | Register |
 |------------------|----------|
 | Driver number    | r0       |
-| Buffer number    | r1       |
+| Allow number     | r1       |
 | Address          | r2       |
 | Size             | r3       |
 
+The *allow number* argument is an ordinal number (index) of the buffer.
+When Read-Write Allow is called, the provided buffer
+SHALL get assigned to the provided *allow number*,
+replacing the previous buffer assigned to that *allow number*,
+if there was one.
+The supported *allow number*s are defined by the driver.
 
 The Tock kernel MUST check that the passed buffer is contained within
 the calling process's writeable address space. Every byte of a passed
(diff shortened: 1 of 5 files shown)


doc/reference/trd104-syscalls.md
```diff
@@ -522,6 +522,12 @@ on RISC-V.
 | Address          | r2       |
 | Size             | r3       |
 
+The *buffer number* argument is an ordinal number (index) of the buffer.
```
lschuermann, line 525: Is there some other place in which the terminology "buffer number" is used in Tock? If you're proposing to introduce it, I'm afraid it might be even more confusing than what we have currently. "buffer" is a rather overloaded term. Traditionally we've been calling these specifiers "subdriver number" or, for each system call respectively, "allow number" and "subscribe number". This makes it obvious that you're referring to the identifier of a given allow number provisioned by a driver as part of its ABI.

dcz-self: I just copied this from the argument name for consistence. It's only used in doc/reference/trd-userspace-readable-allow-syscalls.md apart from here.

Changed to "allow number".

doc/reference/trd104-syscalls.md
```diff
@@ -518,19 +518,23 @@ on RISC-V.
 | Argument         | Register |
 |------------------|----------|
 | Driver number    | r0       |
-| Buffer number    | r1       |
+| Allow number    | r1       |
```
jrvanwhy, line 521: You need an extra space to keep the `|`'s aligned.

doc/reference/trd104-syscalls.md
```diff
@@ -654,6 +658,8 @@ class. It differs in two ways:
 1. The buffer it passes to the kernel is read-only, and the process MAY
    freely read the buffer.
 2. The kernel MUST NOT write to a buffer shared with a Read-Only Allow.
+3. The *allow number*s in the Read-Only Allow
+are independent from those in the Read-Write Allow.
```
jrvanwhy, line 662: The line beginning with `are` should be indented to line up with the word `The` on the previous line (see entry `1.` in this list).

kernel/src/syscall_driver.rs
```diff
@@ -179,9 +179,11 @@ impl From<process::Error> for CommandReturn {
...
 /// driver.
 ///
-/// Note about subscribe: upcall subscriptions are handled entirely by the core
-/// kernel, and therefore there is no subscribe function for capsules to
-/// implement.
+/// Note about **subscribe** and **allow** syscalls:
+/// upcall subscriptions and buffer allows
+/// are handled entirely by the core kernel,
+/// and therefore there is no corresponding function
+/// for capsules to implement.
```
jrvanwhy, line 186: Nitpick: these comments should be wrapped at 80 columns.

This isn't quite correct, because userspace-readable Allow calls do have a corresponding method. It's only Read-Only Allow and Read-Write Allow that have no corresponding method.

dcz-self: Updated to mention userspace-readable allow.

If word wrapping in doc comments is relevant, I would suggest having it as part of the formatter run. It's a bother to reflow things manually. Thankfully, computers are good at this.

jrvanwhy: > If word wrapping in doc comments is relevant, I would suggest having it as part of the formatter run. It's a bother to reflow things manually. Thankfully, computers are good at this.

My experience begs to differ :-p

Automatic formatters can easily detect a line that is too long, but when a line is wrapped too short, they can't tell whether the line was wrapped for alignment reasons (e.g. to make a table look good) or whether it should be unwrapped. Telling an autoformatter to automatically remove line wrapping results in lists and tables getting unwrapped into paragraphs.

As a result, `rustfmt` only supports automatically wrapping comments, not un-wrapping them.

dcz-self: Rust standardizes doc comments as markdown, and markdown specifies paragraphs as their own thing, distinct from tables. It can be taken advantage of. It's a matter of taste, but I prefer manual adjustments to unusual constructs once in a while rather than having to manually re-break lines on every change in prose.

thulithwilfred (approved):

bradjc (approved):

bradjc:
I think we can move on this.

bors r+

PR #1457, opened 2019-11-08
Implement a driver, HIL, and capsule to write to flash

ia0:
### Pull Request Overview

The purpose of the capsule is to provide low-level control of the embedded flash to allow applications to implement flash-efficient data-structures using their writeable flash regions. The API is blocking since most flash either halt the CPU during write and erase operations or ask the application to wait until the operation is finished. A blocking API is also simpler to reason and less error-prone.

### Testing Strategy

In addition to `make ci-travis`, this pull request was tested on the `nrf52840dk` board by implementing in `libtock-rs`:
- a syscall interface
- a multi-set data-structure
- an example application similar to `blink.rs` but with a persistent counter

And updating the linker script.

### TODO or Help Wanted

This pull request may use feedback for other boards if needed, although this may as well be done in followup pull-requests.

### Documentation Updated

- [x] Updated the relevant files in `/docs`.

### Formatting

- [x] Ran `make formatall`.

PR #2993, opened 2022-03-13
hil: Add generic block device HIL

dcz-self:
### Pull Request Overview

This pull request adds/changes the flash HIL (https://github.com/tock/tock/pull/2248) to be more generic, and fit flash devices with or without FTL, as long as they have uniform block sizes.

Changes compared to https://github.com/tock/tock/pull/2248
- the name: block device
- most calls use block indices, not addresses. This way it's impossible to request an invalid operation.
- explicitly stated that this is for uniformly sized block devices
- erase is relaxed
- removed redundant descriptions in returns (when repeating the signature)
- documented errors on sizes
- added call to get total size

### Testing Strategy

This pull request was tested by compiling. Implementation for mx25r64 is in progress and will come later.

### TODO or Help Wanted

This pull request still needs reviews. In particular, I'm not sure if a future HIL suitable for flash should rather go in the more generic, or more specialized direction (in particular, non-uniform block/page/sector sizes).

I'm also not sure if it's a good idea to read/write/erase multiple blocks in one go. Perhaps call back multiple times, once per block? Or limit to one block per request?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/hil/block_storage.rs
+++ b/kernel/src/hil/block_storage.rs
@@ -1,0 +1,248 @@
+//! Interface for reading, writing, and erasing storage blocks
+//! on devices without a Flash Translation Layer.
+//!
+//! Operates on discardable and writeable blocks, where areas must be
+//! discarded before overwriting.
+//!
+//! Here's an example implementation for a raw flash chip:
+//!
+//! ```rust,ignore
+//! use kernel::hil;
+//! use kernel::ErrorCode;
+//!
+//! const WRITE_BLOCK_BYTES: usize = 256;
+//! const DISCARD_BLOCK_BYTES: usize = 4096;
+//!
+//! struct RawFlashChip {};
+//!
+//! type WriteBlockIndex = hil::block_storage::BlockIndex<WRITE_BLOCK_BYTES>;
+//! type DiscardBlockIndex = hil::block_storage::BlockIndex<DISCARD_BLOCK_BYTES>;
+//!
+//! impl hil::block_storage::BlockStorage<WRITE_BLOCK_BYTES, DISCARD_BLOCK_BYTES> for RawFlashChip {
+//!     //(implement associated functions here)
+//! }
+//! ```
+use crate::ErrorCode;
+use core::ops::Add;
+
+/// An index to a block within device composed of blocks containing `B` bytes.
+/// Stores the number of blocks from the start of the device.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct BlockIndex<const B: usize>(pub u32);
+
+impl<const B: usize> BlockIndex<B> {
+    /// Returns the index that contains the address.
+    pub fn new_containing(byte_address: u64) -> Self {
+        Self((byte_address / B as u64) as u32)
+    }
+
+    /// Returns the index starting at the given address, if any.
+    pub fn new_starting_at(byte_address: u64) -> Option<Self> {
+        if byte_address % B as u64 == 0 {
+            Some(Self::new_containing(byte_address))
+        } else {
+            None
+        }
+    }
+}
+
+impl<const B: usize> From<BlockIndex<B>> for u64 {
+    fn from(index: BlockIndex<B>) -> Self {
+        index.0 as u64 * B as u64
+    }
+}
+
+impl<const B: usize> Add<u32> for BlockIndex<B> {
+    type Output = Self;
+    fn add(self, other: u32) -> Self {
+        BlockIndex(self.0 + other)
+    }
+}
+
+/// Readable persistent block storage device.
+///
+/// The device is formed from equally-sized storage blocks,
+/// which are arranged one after another, without gaps or overlaps,
+/// to form a linear storage of bytes.
+///
+/// Every byte on the device belongs to exactly one block.
+///
+/// `R`: The size of a read block in bytes.
+pub trait ReadableStorage<const R: usize> {
+    /// Read data from a block, and into the buffer.
+    ///
+    /// `ErrorCode::INVAL` will be returned if
+    /// - `block` exceeds the end of the device, or
+    /// - `buf` is shorter than `block`.
+    ///
+    /// Returns `ErrorCode::BUSY` when another operation is in progress.
+    ///
+    /// On success, triggers `ReadableClient::read_complete` once.
+    fn read(
+        &self,
+        block: &BlockIndex<R>,
+        buf: &'static mut [u8],
+    ) -> Result<(), (ErrorCode, &'static mut [u8])>;
+
+    /// Returns the size of the device in bytes.
+    fn get_size(&self) -> u64;
+}
+
+/// Writeable persistent block storage device.
+///
+/// The device is formed from equally-sized storage blocks,
+/// which are arranged one after another, without gaps or overlaps,
+/// to form a linear storage of bytes.
+///
+/// The device is split into blocks in two ways, into:
+/// - discard blocks, which are the smallest unit of space
+/// that can be discarded (see `BlockStorage::discard`)
+/// - write blocks, which are the smallest unit of space that can be written
+///
+/// Every byte on the device belongs to exactly one discard block,
+/// and to exactly one write block at the same time.
+///
+/// `W`: The size of a write block in bytes.
+/// `D`: The size of a discard block in bytes.
+pub trait WriteableStorage<const W: usize, const D: usize> {
+    /// Write data from a buffer to storage.
+    ///
+    /// This function writes the contents of `buf` to memory,
+    /// at the chosen `block`.
+    ///
+    /// `ErrorCode::INVAL` will be returned if
+    /// - `block` exceeds the end of the device, or
+    /// - `buf` is shorter than `block`.
+    ///
+    /// This function SHALL NOT discard the block first.
+    /// The user of this function MUST ensure that the relevant block
+    /// has been successfully discarded first (see `discard`).
+    ///
+    /// Once a byte has been written as part of a write block,
+    /// it MUST NOT be written again until it's discarded
+    /// as part of a discard block.
+    /// Multiple consecutive writes to the same block
+    /// are forbidden by this trait, but the restriction is not enforced.
+    ///
+    /// **Note** about raw flash devices: writes can turn bits from `1` to `0`.
+    /// To change a bit from `0` to `1`, a block must be erased (discarded).
+    ///
+    /// Returns `ErrorCode::BUSY` when another operation is in progress.
+    ///
+    /// On success, triggers `WriteableClient::write_complete` once.
+    fn write(
+        &self,
+        block: &BlockIndex<W>,
+        buf: &'static mut [u8],
+    ) -> Result<(), (ErrorCode, &'static mut [u8])>;
+
+    /// Makes a block ready for writing.
+    ///
+    /// This corresponds roughly to the erase operation on raw flash devices.
+    ///
+    /// A successful `discard` leaves bytes in the selected `block` undefined.
+    /// The user of this API must not assume any property
+    /// of the discarded bytes.
+    ///
+    /// If `block` exceeds the size of the device, returns `ErrorCode::INVAL`.
+    ///
+    /// Returns `ErrorCode::BUSY` when another operation is in progress.
+    ///
+    /// On success, triggers `WriteableClient::discard_complete` once.
+    fn discard(&self, block: &BlockIndex<D>) -> Result<(), ErrorCode>;
+}
+
+pub trait Storage<const W: usize, const D: usize>:
+    ReadableStorage<W> + WriteableStorage<W, D>
+{
+}
+
+impl<const W: usize, const D: usize, T: ReadableStorage<W> + WriteableStorage<W, D>> Storage<W, D>
+    for T
+{
+}
+
+/// Specifies a storage area with byte granularity.
+pub struct AddressRange {
+    /// Byte offset from the beginning of the storage device.
+    pub start_address: u64,
+    /// Length of the range in bytes.
+    pub length_bytes: u32,
+}
+
+impl AddressRange {
+    /// Returns the address of the byte after the last byte in the range
+    pub fn get_end_address(&self) -> u64 {
+        self.start_address + self.length_bytes as u64
+    }
+}
+
+impl<const C: usize> From<BlockIndex<C>> for AddressRange {
+    fn from(block: BlockIndex<C>) -> Self {
+        AddressRange {
+            start_address: block.0 as u64 * C as u64,
+            length_bytes: C as u32,
+        }
+    }
+}
+
+/// Devices which can read arbitrary byte-indexed ranges.
+pub trait ReadRange {
+    /// Read data from storage into a buffer.
+    ///
+    /// This function will read data stored in storage at `range` into `buf`.
+    ///
+    /// `ErrorCode::INVAL` will be returned if
+    /// - `range` exceeds the end of the device, or
+    /// - `buf` is shorter than `range`.
+    ///
+    /// Returns `ErrorCode::BUSY` when another operation is in progress.
+    ///
+    /// On success, triggers `ReadableClient::read_complete` once.
+    fn read_range(
+        &self,
+        range: &AddressRange,
+        buf: &'static mut [u8],
+    ) -> Result<(), (ErrorCode, &'static mut [u8])>;
+}
+
+pub trait HasClient<'a, C> {
+    /// Set the client for this peripheral. The client will be called
+    /// when operations complete.
+    fn set_client(&'a self, client: &'a C);
+}
+
+/// Implement this to receive callbacks from `ReadableStorage` and `ReadRange`.
+pub trait ReadableClient {
+    /// This will be called when a read operation is complete.
+    ///
+    /// If the device is unable to read the block, returns `ErrorCode::FAIL`.
+    ///
+    /// On errors, the buffer contents are undefined.
+    fn read_complete(&self, read_buffer: &'static mut [u8], ret: Result<(), ErrorCode>);
+}
+
+/// Implement this to receive callbacks from `ReadableStorage`.
+pub trait WriteableClient {
+    /// This will be called when the write operation is complete.
+    ///
+    /// If the device is unable to write to the block,
+    /// returns `ErrorCode::FAIL`.
+    ///
+    /// On errors, the contents of the storage block are undefined,
+    /// and the block must be considered written.
+    fn write_complete(&self, write_buffer: &'static mut [u8], ret: Result<(), ErrorCode>);
+
+    /// This will be called when the discard operation is complete.
+    ///
+    /// If the device is unable to discard the block,
+    /// returns `ErrorCode::FAIL`.
+    ///
+    /// On errors, the contents of the storage block are undefined,
+    /// and the block's discarded status must be considered unchanged.
+    fn discard_complete(&self, ret: Result<(), ErrorCode>);
+}
+
+pub trait Client: WriteableClient + ReadableClient {}
+
+impl<T: WriteableClient + ReadableClient> Client for T {}
--- a/kernel/src/hil/mod.rs
+++ b/kernel/src/hil/mod.rs
@@ -3,6 +3,7 @@
 pub mod adc;
 pub mod analog_comparator;
 pub mod ble_advertising;
+pub mod block_storage;
 pub mod bus8080;
 pub mod crc;
 pub mod dac;


dcz-self:
I added a basic implementation and made a simple test on real HW in https://github.com/dcz-self/tock/commits/flashil .

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// This corresponds roughly to the erase operation on raw flash devices,
+    /// but may also do nothing on devices where erasure is not necessary.
+    ///
+    /// Calling `erase` may modify bytes in the selected `region` in any way
+    /// (typically, on raw flash devices, all bits will be set,
+    /// i.e. each byte turns to 0xFF).
+    ///
+    /// If `region` exceeds the size of the device, returns `ErrorCode::INVAL`.
+    /// On success, triggers `Client::erase_complete` once.
+    fn erase(&self, region: &Region<E>) -> Result<(), ErrorCode>;
```
bradjc, line 240: This description seems like a different operation than "erase". I think this should either promise to clear data or be called `prepare_write()` (or something similar).

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    /// Returns the size of the device in bytes.
+    fn get_size(&self) -> Result<u64, ErrorCode>;
```
bradjc, line 243: How could this fail?

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// when operations complete.
+    fn set_client(&'a self, client: &'a C);
+}
+
+/// Implement `Client` to receive callbacks from `BlockStorage`.
+pub trait Client<const W: u32, const E: u32> {
+    /// Block read complete.
+    ///
+    /// This will be called when the read operation is complete.
+    fn read_complete(&self, read_buffer: &'static mut [u8], ret: Result<(), ErrorCode>);
```
bradjc, line 257: What are the errors?

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,268 @@
...
+        buf: &'static mut [u8],
+    ) -> Result<(), (ErrorCode, &'static mut [u8])>;
+
+    /// Write data from a buffer to flash.
+    ///
+    /// This function writes the contents of `buf` to memory,
+    /// starting at the chosen `start_block`.
+    ///
+    /// `ErrorCode::INVAL` will be returned if
+    /// - `range` exceeds the end of the device, or
```
bradjc, line 200: No `range`

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// - `buf` is shorter than `region`.
+    ///
+    /// This function SHALL NOT prepare the block for writing first.
```
bradjc, line 203:
```suggestion
    /// This function SHALL NOT prepare the region for writing first.
```
? Or "blocks in region" maybe.

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// On failure returns a `ErrorCode` and the buffer passed in.
+    /// If `ErrorCode::NOSUPPORT` is returned then `read_block`
+    /// should be used instead.
+    fn read_range(
+        &self,
+        range: &AddressRange,
+        buf: &'static mut [u8],
+    ) -> Result<(), (ErrorCode, &'static mut [u8])>;
+
+    /// Read data from blocks starting at `block`, and into the buffer.
```
bradjc, line 181: region

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub fn new_containing(range: AddressRange) -> Self {
+        let index = BlockIndex::new_containing(range.start_address);
+        Self {
+            index,
+            length_blocks: {
+                div_ceil(
+                    range.start_address % S as u64 + range.length_bytes as u64,
+                    S as u64,
+                ) as u32
+                // CAUTION: div_ceil is nightly
```
bradjc, line 98: what does this refer to?

bradjc (commented):
I'm not the best to judge if this meets the various needs for an alternative flash hil, but I made some comments on the HIL and documentation.

kernel/src/hil/block_storage.rs
dcz-self: I think this could be reasonably expected to ask the device to identify itself, and fail if the size can't be determined.

On the other hand, perhaps it could be made infallible. The other methods are already supposed to fail if bounds are exceeded, and it doesn't make a lot of sense to not know how big a device is from the init.

I could use some feedback on this.

kernel/src/hil/block_storage.rs
dcz-self: It's the `div_ceil` function which is needed here: https://doc.rust-lang.org/std/primitive.u64.html#method.div_ceil . This is also what the FIXME is referring to; I haven't found the right incantation to do the division yet.

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,268 @@
...
+//!     { Ok(()) }
+//!     
+//!     (implement remaining associated functions here)
+//! }
+//! ```
+use crate::ErrorCode;
+
+/// An index to a block within device composed of `S`-sized blocks.
+#[derive(Clone, Copy)]
+pub struct BlockIndex<const S: u32>(pub u32);
```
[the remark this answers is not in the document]
dcz-self: I think `Index` works to disambiguate that it's an indirection, and not e.g. a copy of the data.

dcz-self:
Thanks for reviewing, especially proofreading. I changed the names a few times and that left stale traces.

After implementing a userspace driver, I'm more and more tempted to drop the ability to affect multiple blocks in a row. Supporting multiple blocks makes the internal capsule logic more complex, and it's not clear to me what the API should do about buffer sizes. Unless there's good reason to keep it in the kernel, I would rather do that in the userspace.

That being said, my file system experience is limited. Perhaps it makes sense from the performance POV. Feedback welcome.

kernel/src/hil/block_storage.rs
bradjc: Well if this leads to a device query then it would have to be asynchronous, or the device query would have to happen in advance which complicates the implementation.

But in general, are there hardware designs which wouldn't know the size of their block storage? It seems like it's a reasonable assumption for an embedded device to make that it knows the size of its storage.

kernel/src/hil/block_storage.rs
bradjc: Ok makes sense, I was just expecting to see a feature.

kernel/src/hil/block_storage.rs
dcz-self: I removed the possibility to have an error completely. The data manipulation needs to know the device size anyway to be able to error out early anyway.

kernel/src/hil/block_storage.rs
dcz-self: I left it for now. I presume that any virtualization needs to store the read offset anyway, so storing read size alongside it wouldn't be a problem. Or am I mistaken?

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,268 @@
...
+    ///
+    /// This function will read data stored in flash at `range` into `buf`.
+    ///
+    /// `ErrorCode::INVAL` will be returned if
+    /// - `range` exceeds the end of the device, or
+    /// - `buf` is shorter than `range`.
+    ///
+    /// On success, triggers `Client::read_complete` once.
+    /// On failure returns a `ErrorCode` and the buffer passed in.
+    /// If `ErrorCode::NOSUPPORT` is returned then `read_block`
```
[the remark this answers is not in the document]
dcz-self: Split into another trait.

kernel/src/hil/block_storage.rs
phil-levis: I'm wary of this SHALL NOT, because it's an implementation decision. I think you want to say that this method does not any claims of preparation, and that a client that needs a clean block MUST call `prepare_write` first. That leaves open the possibility of multiple writes, while also allowing implementations that defer preparation to this actual method call (e.g., by storing a "prepare first" bit).

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,238 @@
...
+    fn write(
+        &self,
+        region: &BlockIndex<W>,
+        buf: &'static mut [u8],
+    ) -> Result<(), (ErrorCode, &'static mut [u8])>;
+
+    /// Makes a region ready for writing.
+    ///
+    /// This corresponds roughly to the erase operation on raw flash devices,
+    /// but may also do nothing on devices where erasure is not necessary.
```
phil-levis, line 155: This suggests that this API does not support the possibility of multiple writes (e.g., on flash)?

I think that's OK, but this should be explicit.

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// This function SHALL NOT prepare the block for writing first.
+    /// The user of this function SHOULD ensure that a block is prepared
+    /// before calling `write` (see `prepare_write`).
+    ///
+    /// Once a byte has been written as part of a write block,
+    /// it SHOULD NOT be written again until it's prepared
+    /// as part of a prepare block.
+    /// Multiple writes to the same block are possible in this trait,
+    /// but the result is dependent on the nature of the underlying hardware,
+    /// and this interface does not provide a way to discover that.
```
phil-levis, line 133: I am wary of this SHOULD NOT. It means that someone can use an instance of this trait and make implementation-specific or hardware-specific assumptions, which then break on another implementation. This should either not be general (distinguish write models), or a MUST.

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,238 @@
...
+/// and to exactly one write block at the same time.
+///
+/// `W`: The size of a write block in bytes.
+/// `P`: The size of a prepare block in bytes.
+pub trait BlockStorage<const W: usize, const P: usize> {
+    /// Read data from a block, and into the buffer.
+    ///
+    /// `ErrorCode::INVAL` will be returned if
+    /// - `region` exceeds the end of the device, or
+    /// - `buf` is shorter than `region`.
```
phil-levis, line 104: Distinguish these two cases with SIZE?

kernel/src/hil/block_storage.rs
phil-levis: Given both BlockIndex and AddressRange as borrowed with an anonymous lifetime, I think this is OK. It does necessitate that the caller keeps that state. I see benefits and drawbacks in both approaches (having the caller keep the state, or having the callee pass it back).

phil-levis (changes requested):
Overall, I think this looks pretty good. By just trying to do a simpler, general block interface it avoids all of the difficult complexities of flash. However, there's one detail I'm worried about. It's often the case that the erase unit for flash is larger than the write unit. In that case, `prepare_write` can have side effects. Or is the intention that this API would just make the write unit as large as the erase unit?

Finally, I would recommend separating the read and write APIs, then joining them in a composite trait. With this API, you can't give a client just read access or just write access. Please take a look at https://github.com/tock/tock/blob/master/doc/reference/trd2-hil-design.md#rule-8-use-fine-grained-traits-that-separate-different-use-cases

kernel/src/hil/block_storage.rs
dcz-self: I think you're reading this right: regions on flash are not ready for writing unless they get erased first. Without this method, we can't support flash.

This particular passage says that if you have 3D Xpoint memory or a plain old HDD, this may be implemented as a no-op.

kernel/src/hil/block_storage.rs
dcz-self: Sure. I wanted to accommodate NOR flash devices (they allow multiple writes without erasing), but it's probably better to not touch it until there's an actual need.

dcz-self:
I originally had the "prepare_write" method called "erase", but that was not clear enough. "discard" is used in other contexts, and I think it's clearer that its block size is not the same as the write size (especially after I rename the "erase block size" to "discard block size"). Would that be clearer?

kernel/src/hil/block_storage.rs
phil-levis: This starts to get a little weird. It means that a client on top of an HDD can read discarded data, while a client on top of flash can't. This then exposes which one the client is on top of, such that a client sitting on top of a HDD will do things that then won't work if you replace the underlying block store. This means that API is no longer transparent and isn't an interchangeable API.

kernel/src/hil/block_storage.rs
phil-levis: I believe some NAND devices do as well (although fewer writes).

phil-levis:
> I originally had the "prepare_write" method called "erase", but that was not clear enough. "discard" is used in other contexts, and I think it's clearer that its block size is not the same as the write size (especially after I rename the "erase block size" to "discard block size"). Would that be clearer?

I think, for this API, separating the discard and write sizes is a big change. There are a lot of weird cases. In particular, what happens if you read after discard? With this API, a device using a block device (e.g., HDD, a RAM device, or a NVRAM device) has to manage the complexity of discard sizes potentially being larger than write sizes, even though discard is a no-op. I.e., if it wants to change one block of a page, it has to copy out the other blocks first, discard the page, then re-write the blocks. I thought you were going in the opposite direction, that for this simple API the write unit is the same as the discard unit. The challenge here, though, is that the erase unit for large NAND is often 256kB.

There are NAND flashes that allow multiple writes to a single block. This is sometimes used by flash-specific storage abstractions, e.g., to maintain free lists as bitmasks, etc. I don't think this API needs to represent this. 

There's been I think now 3 attempts to write a better API for flash, that exposes all of its pecularities. It's tough. One effort didn't even make it to be a PR, here are the other two:

https://github.com/tock/tock/pull/2249
https://github.com/tock/tock/pull/2248
https://github.com/tock/tock/pull/1457

kernel/src/hil/block_storage.rs
dcz-self: Perhaps I didn't make it clear, but my intention here is that the result of reads of discarded blocks are undefined. The note that discard may be a no-op is a hint to the driver author, rather than an API guarantee.

By assuming SSD semantics of required discard, the API remains interchangeable (if SSD/HDD interchangeability is what you refer to).

Would it be clearer with the following wording?

> This corresponds roughly to the erase operation on raw flash devices.
> An implementation may internally decide to do nothing when irrelevant,
> e.g. on devices where erasure is not necessary.

dcz-self:
> what happens if you read after discard?

As mentioned in the other comment, I think the only sane way out is to have the consumer suffer the consequences. It's infeasible to try to react to bad reads, because that would require tracking erased blocks, which I think is out of scope of a low-level layer like this.

> With this API, a device using a block device (e.g., HDD, a RAM device, or a NVRAM device) has to manage the complexity of discard sizes potentially being larger than write sizes, even though discard is a no-op.

The block sizes are set by the driver, so if the driver determines that discard is a no-op, it is free to define all block sizes as the same value.

kernel/src/hil/block_storage.rs
phil-levis: The issue with "undefined" is that a client which assumes a particular implementation is then bound to it; it's no longer an interchangeable interface. The client uses a `BlockStorage`, but it has to be a specific one or it won't work. I.e., a client built on top of an HDD or NVRAM device could read after discarding, but then if you put it on top of an SSD it won't work.

phil-levis: > This corresponds roughly to the erase operation on raw flash devices.
> An implementation may internally decide to do nothing when irrelevant,
> e.g. on devices where erasure is not necessary.

The issue is what the client will do.

dcz-self: I see "undefined" more as "forbidden, but not enforced", as opposed to "implementation-defined". If a client decides to rely on something that's outside of the API contract, then it's the client's problem. Within the constraint that a client only uses allowed operations, it won't be able to distinguish between a HDD or SDD.

I admit it's not perfect to leave an undefined operation, but I think this is a local optimum for the useablity of this API. If we tried to eliminate that undefinedness, I expect we'd have to come up with something a lot more complicated. Perhaps we want to in the future, but I think this proposal is a good stepping stone to open up the familiar block storage functionality at all.

phil-levis:
> > what happens if you read after discard?
> 
> As mentioned in the other comment, I think the only sane way out is to have the consumer suffer the consequences. It's infeasible to try to react to bad reads, because that would require tracking erased blocks, which I think is out of scope of a low-level layer like this.

I think we're talking at cross purposes here. You're discussing what an implementation could do, given an ambiguous specification that allows multiple behaviors. I'm pointing out this means that a client written for one implementation might not work with others because it assumes those behaviors. The API doesn't fully encapsulate the abstraction and create a layered system; information about the lower layer can leak into the upper layer. In that case, we should just define a block API and a flash API.

dcz-self:
> a client written for one implementation might not work with others because it assumes those behaviors

Right, I'm fine with letting the client live in a hell of its own making by choosing to adhere to the implementation rather than the contract ;)

More seriously, like I wrote before, I don't see any more practical option at the moment. This API is primarily meant to cover devices that require erasure before writing. I don't currently see a way to define an API that enforces erasure before writing without revealing underlying implementation details, so splitting this proposal into 2 APIs doesn't bring any benefits that I can see.

dcz-self:
To add on to the previous comment – if we wanted to be explicit about the underlying implementation, we'd have to define what happens to the erased blocks on NAND and NOR devices, and we'd have to define how many writes are allowed. The first one means we can only support devices which set bits to 1 on erase (reasonable enough), and those which may be rewritten at most once (nrf52 supports writing twice, so would not be supported). Because if a chip supports two rewrites, a client could try to rely on this property of the implementation, and we would have to split out an API supporting 2 rewrites as a separate driver.

I'm not sure it makes sense to try to avoid all such problems explicitly, strictly from the practical point of view. I'd rather use this HIL as a stepping stone to implement a file system, and recommend an actual file system in the long term, after marking the block driver "unsafe" in some way.

dcz-self:
Having deliberated on this some more, I know how to force an erase before a write: by returning a "ticket" from an erase operation, and consuming the ticket as a requirement for the write operation.

I wasn't able to invent any mechanism to prevent a read of erased data. But I don't think we can completely prevent that without rolling out an entire file system: it will always be possible to read erased blocks right after the driver was loaded, as erasure bitmaps are not preserved. This means that the specifics of the medium will always be accessible to the client, so it remains for us to choose whether we want to mandate particular implementations in this HIL or not.

phil-levis:
> Having deliberated on this some more, I know how to force an erase before a write: by returning a "ticket" from an erase operation, and consuming the ticket as a requirement for the write operation.

Excellent!

> I wasn't able to invent any mechanism to prevent a read of erased data. But I don't think we can completely prevent that without rolling out an entire file system: it will always be possible to read erased blocks right after the driver was loaded, as erasure bitmaps are not preserved. This means that the specifics of the medium will always be accessible to the client, so it remains for us to choose whether we want to mandate particular implementations in this HIL or not.

Given that you need to separate the read and write paths, and the fact that the actual timing of erases is implementation-dependent (they could be deferred, occur on the next write, etc), that's OK.

How does this work if the erase and write units are different? Does an erase give you a set of write tokens?

dcz-self:
> How does this work if the erase and write units are different? Does an erase give you a set of write tokens?

Correct.

The downside of this approach is that the driver now needs to be able to allocate memory. Handing out opaque tokens won't cut it. Consider:

```
let t1 = discard(0);
let t2 = discard(0);
write(t1, data1);
write(t2, data2); // 2 writes, we're doomed
```

To prevent the double-write, the driver could prevent handing out the same token twice. It must therefore keep track of all active tokens.

Option 1: the driver allocates the memory itself. The driver must either utilize allocation or have a limit on the number of tokens in flight.
Option 2: the driver uses memory provided by the client. The API would become sth like `discard(block: u32, space: &'static mut Space) -> Token` - by similarity to the current way of sending buffers to SPI devices. The buffer would be returned in `write(Token) -> &'static mut Space`, where the client could get rid of the memory. The problem with this approach is that it's now impossible to use the HIL if the number of tokens in flight is unknown, like in a userspace driver.

I think all options here suck, although the version which limits the number of active tokens seems simple enough to try out. It might even survive a meeting with reality if the file system built on top is conservative about the number of used tokens. Ultimately though, what we're trying to do here is to build an occupancy bitmap, which, as I wrote before, I consider overkill for this application as it belongs in the file system layer.

hudson-ayers:
> > > what happens if you read after discard?
> 
> > 
> 
> > As mentioned in the other comment, I think the only sane way out is to have the consumer suffer the consequences. It's infeasible to try to react to bad reads, because that would require tracking erased blocks, which I think is out of scope of a low-level layer like this.
> 
> 
> 
> I think we're talking at cross purposes here. You're discussing what an implementation could do, given an ambiguous specification that allows multiple behaviors. I'm pointing out this means that a client written for one implementation might not work with others because it assumes those behaviors. The API doesn't fully encapsulate the abstraction and create a layered system; information about the lower layer can leak into the upper layer. In that case, we should just define a block API and a flash API.

"Using the API in an unsupported way is not allowed to leak any information about the lower layer" seems like a higher isolation bar than we generally enforce for Tock HILs, and I am not sure that the benefits (preventing someone writing a driver that uses the API in a way forbidden by the docs but happens to work on only some platforms) of enforcing it here are worth the costs (significant implementation complexity, and likely size and runtime degradation).

If the concern is that someone might rely on erase to clear sensitive data, the HIL should just clearly document that you must overwrite sensitive data if you want to guarantee it is not accessible in the future.

phil-levis:
> > How does this work if the erase and write units are different? Does an erase give you a set of write tokens?
> 
> Correct.
> 
> The downside of this approach is that the driver now needs to be able to allocate memory. Handing out opaque tokens won't cut it. Consider:
> 
> ```
> let t1 = discard(0);
> let t2 = discard(0);
> write(t1, data1);
> write(t2, data2); // 2 writes, we're doomed
> ```

Yeah, without keeping per-page metadata this doesn't work; that's why FTLs are like they are.

What size flashes are you thinking of? A 16MB flash with 4kB pages, for example, can have a "erased" bitmask of 512 bytes.

It's my experience that you want to treat FTL'd and non-FTL'd chips differently. You want to use different storage abstractions. E.g., without an FTL you need to do wear-levelling or you can easily lose blocks. Unless you expect everything using this layer to use logging approaches and just be inefficient on an FTL device?

phil-levis:
> seems like a higher isolation bar than we generally enforce for Tock HILs, and I am not sure that the benefits (preventing someone writing a driver that uses the API in a way forbidden by the docs but happens to work on only some platforms) of enforcing it here are worth the costs (significant implementation complexity, and likely size and runtime degradation).

*shrug* It'ss just good software design. Otherwise you get hidden dependencies and pieces are not composable. The cost you pay here (which I mention in a separate comment above) is that you have to assume there isn't an FTL, which means all of your storage approaches need to be log-structured or you can quickly (e.g., 100 erases) lose blocks. So providing this API on erase-free devices (HDD or FTL'd flash) imposes a large complexity overhead.

If we want to say that this API is really intended for raw flash, but of course you can put it on top of an FTL device and pay a huge penalty, I'd be OK with that. I'd expect we'd have a different API for a write-in-place device, and a trivial shim (ignores erase operations). It does get a little weird in that many flash storage abstractions assume the erase behavior of flash (e.g., for bitmasks). It would mean that if you want to be sure about erase behavior your actually need to erase, write, erase.

dcz-self:
I don't like the RAM occupancy map, because it imposes an extra resource cost on all users. Twofold: if the client tries to be lean by storing occupancy on the flash device itself, it doesn't need any of this in the RAM. Secondly, when storing a bitmap of erased blocks, every block erased before initialization will be erased again before writing, ~halving the lifetime of the device in some scenarios.

I think there might be a place for such a driver. But still, implementing an erasure bitmap in RAM will not completely prevent the client from figuring out what the lower layer is: all the client needs to do is to erase a block, wait until the driver is reinitialized, and then read it.

This HIL is not suitable for FTL-naive clients. The API makes it possible to use any kind of device (FTL or not. It's impossible to prevent a write-many-times device from exposing an erase-write interface), but by requiring erasure, it necessarily requires clients to do the management themselves. Your observations on HDDs are correct. I mostly left them in because why not - HDDs are a generalization of flash, and I want my code work on them too, even when it's not the best match.

> It would mean that if you want to be sure about erase behavior your actually need to erase, write, erase.

Could you provide an example of when that would be needed?

phil-levis:
> I don't like the RAM occupancy map, because it imposes an extra resource cost on all users. Twofold: if the client tries to be lean by storing occupancy on the flash device itself, it doesn't need any of this in the RAM. Secondly, when storing a bitmap of erased blocks, every block erased before initialization will be erased again before writing, ~halving the lifetime of the device in some scenarios.

I agree, I don't either. 

> This HIL is not suitable for FTL-naive clients. The API makes it possible to use any kind of device (FTL or not. It's impossible to prevent a write-many-times device from exposing an erase-write interface), but by requiring erasure, it necessarily requires clients to do the management themselves. Your observations on HDDs are correct. I mostly left them in because why not - HDDs are a generalization of flash, and I want my code work on them too, even when it's not the best match.

The first sentence of the PR says: *and fit flash devices with or without FTL,*

I think your comments are a bit contradictory? FTL-native devices act like standard block devices, allowing in-place writes: the FTL allows a flash device to act like an HDD. So if your code works on an HDD, it'll work on an FTL flash device. 

> > It would mean that if you want to be sure about erase behavior your actually need to erase, write, erase.
> 
> Could you provide an example of when that would be needed?

A storage layer is keeping a bitmask, denoting the use of a particular resource (blocks, etc.). It calls erase and expects the bits to be cleared (let's suppose it has a clean way to know whether `1` is cleared or `0` is depending on medium). It later reads the bits to see which blocks are in use. If it's using an implementation that just discards erases it will see the old values. So the storage layer actually needs to erase-write to be sure the bits are cleared, then erase so it can start writing them again. So erase-write-erase. Alternatively, implementations of this trait on HDD/FTL/NVRAM would need to write to the block to clear it on an erase operation. Put another way, an erase is a type of write, just a very specific one.

dcz-self:
> The first sentence of the PR says: and fit flash devices with or without FTL,
>
> I think your comments are a bit contradictory? FTL-native devices

Note that I did not write "na**t**ive devices", but "naive *clients*". The client has to expect an FTL, the device is unconstrained.

Thanks for explaining the erase-write-erase. Yes, that's a problem after each reset of the state of the storage layer, so best to minimize those IMO.

Where do we stand on it? As far as I can tell:

- rename the "erase" functionality to use the word "discard",
- split out a read-only trait,

is that all to reach a level of usefulness sufficient for inclusion?

phil-levis:
> > The first sentence of the PR says: and fit flash devices with or without FTL,
> > I think your comments are a bit contradictory? FTL-native devices
> 
> Note that I did not write "na**t**ive devices", but "naive _clients_". The client has to expect an FTL, the device is unconstrained.

I'm confused. Do you mean that clients of this interface expect that it has an FTL, but implementations may be on top of raw flash and implement the FTL? If so, why is there a discard operation? Flash translation layers abstract away erases and expose a standard block-write interface, just like an HDD.

> Thanks for explaining the erase-write-erase. Yes, that's a problem after each reset of the state of the storage layer, so best to minimize those IMO.

What do you mean? It's a logical error that can lead to undefined results -- how does the system handle it? Minimization isn't sufficient; what's needed is correctness.
 
> Where do we stand on it? As far as I can tell:
> 
>     * rename the "erase" functionality to use the word "discard",
> 
>     * split out a read-only trait,
> 
> 
> is that all to reach a level of usefulness sufficient for inclusion?

It needs one of three changes:
  - Be defined only for raw flash devices (which have a clear behavior for discard),
  - Be defined for FTL and other block devices and remove the discard operation,
  - Define the behavior of discard (what happens if you read a block after discarding its page), noting that for FTL and other block devices, discard means "write all 0s" or "write all 1s".

dcz-self:
> Do you mean that clients of this interface expect that it has an FTL, but implementations may be on top of raw flash and implement the FTL?

The opposite: clients must be able to deal without an FTL. Implementations may or may not have an FTL.

> What do you mean? It's a logical error that can lead to undefined results -- how does the system handle it? Minimization isn't sufficient; what's needed is correctness.

I mean that when the layer above the device tracking erase state is uninitialized, it knows nothing about the state of the device. It can't be fixed by definition. By minimization I meant: better not keep the state in volatile storage.

> Be defined only for raw flash devices (which have a clear behavior for discard),

Ok, this seems reasonable, and if it needs extending to non-flash, that's also a no-op. I'll do it.

dcz-self:
Sure, I get your point. But

> If the implementation of the trait does have an FTL, then the client using it can treat it like a standard block device and perform arbitrary numbers in-place writes on a single block.

Even if the FTL is not present, the client can treat it like a block device. But by doing this, it's going to break the API contract, regardless of whether there's a FTL, or not. Putting a comment in the docs to the effect of "underlying device does not contain FTL" will have no bearing on how badly the API can be abused, and I meant that removing that comment is a no-op :)

dcz-self:
Thanks for all the reviews. I've done the requested changes:
- separated read and write traits
- removed comments trying to make this HIL too general
- noted the precise semantics of discard: obligatory, not enforced, leaves data undefined

phil-levis:
Two changes needed as part of the splitting into read and write traits;
  - You need to [split the `Client` traits and add `set_client` methods to the separate traits for each type of client](https://github.com/tock/tock/blob/master/doc/reference/trd2-hil-design.md#rule-12-traits-that-can-trigger-callbacks-should-have-a-set_client-method). 
  - Please need to add a [composite trait (with automatic implementation) that combines Read and Write](https://github.com/tock/tock/blob/68cf7940a745e4bba2fe4a1409683443a8348302/kernel/src/hil/uart.rs#L62).

phil-levis:
Other than that, looks good!

phil-levis:
I am not sure if we want to merge this before the 2.1 release -- I am guessing no. It would be good to write a draft TRD describing the design of the API, but merging shouldn't block on this.

dcz-self:
I've added requested changes: Client is now split, and there is a composite Client and Storage.

While adjusting the first driver, I noticed that some convenience methods were missing, so there's now a From<BlockIndex> for AddressRange.

dcz-self:
After I am finished with the driver, I'll fork the syscall interface and write a TRD as well, submitting as I go.

dcz-self:
Minor changes: typos and rewording.

dcz-self:
I've drafted a TRD https://github.com/tock/tock/pull/3157

kernel/src/hil/block_storage.rs
alistair23: We should accommodate multiple writes without an erase. It's a pretty common use case, one that [TicKV](https://github.com/tock/tock/tree/master/libraries/tickv) and [Yaffs1](https://yaffs.net/documents/how-yaffs-works) both rely on for example

dcz-self: I want to address this in a future revision. Currently we managed to work out a compromise based on the agreement that the targeted devices are basic, and the user does not need to know about how many writes they can receive.
Besides, the current version Works For Me™ ;) That is, it's already covering some useful ground.

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,247 @@
...
+    ///
+    /// This function SHALL NOT discard the block first.
+    /// The user of this function MUST ensure that the relevant block
+    /// has been successfully discarded first (see `discard`).
+    ///
+    /// Once a byte has been written as part of a write block,
+    /// it MUST NOT be written again until it's discarded
+    /// as part of a discard block.
```
alistair23, line 123: This is very restrictive and doesn't match a lot of flash use cases

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,247 @@
...
+
+impl<const W: usize, const D: usize, T: ReadableStorage<W> + WriteableStorage<W, D>> Storage<W, D>
+    for T
+{
+}
+
+/// Specifies a storage area with byte granularity.
+pub struct AddressRange {
+    /// Offset from the beginning of the storage device.
+    pub start_address: u64,
```
alistair23, line 168: `usize` for address?

alistair23 (changes requested):

kernel/src/hil/block_storage.rs
dcz-self: Covering all use cases is not the goal at this stage, see discussion leading up to https://github.com/tock/tock/pull/2993#issuecomment-1207910434

kernel/src/hil/block_storage.rs
dcz-self: The storage addresses are not in the CPU memory space, so I opted for making the size independent of CPU architecture.

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,268 @@
...
+/// A memory region composed of consecutive `S`-sized blocks.
+#[derive(Clone, Copy)]
+pub struct Region<const S: u32> {
+    pub index: BlockIndex<S>,
+    pub length_blocks: u32,
+}
+
+/// Divide, round up
+fn div_ceil(a: u64, b: u64) -> u64 {
+    a / b // FIXME: trololo
```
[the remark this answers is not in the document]
twilfredo: Is this what you are after here `(a + b-1) / b`?

kernel/src/hil/block_storage.rs
alistair23: I would say allowing multiple writes doesn't mean covering all use cases, just a large number of them

dcz-self: I agree, but I still think it's better to tackle this as a separate unit of work. Incremental progress seems to be working out better in Tock than general solutions in my experience.

phil-levis: I agree -- all of previous attempts to write a powerful but general flash API have failed because of all of the tricky edge cases: the APIs worked onlu for the particular devices/use cases the author cared about. We can go down that road again, but this shouldn't block having a simpler API for simpler use cases.

phil-levis:
Overall, I think this is in good shape. My one major comment is that since the API is now designed to encapsulate raw flash storage, rather than block devices generally, it should probably not be called "block storage". Instead, should this replace `hil::flash`? I am guessing no due to @alistair23's concerns. However, it might make sense to put it in `hil::flash` as something like "BlockFlash" and explicitly distinguish the two. Otherwise someone looking through the HILs might be confused on which one to use when and why. E.g., the current `hil::flash` only allow reading/writing of pages, while this new API
allows block-level reads and writes.

A "discard block" should be called a "page" (standard flash terminology).

I suspect that we will at some point in the future want to break out discard from write (e.g., so a client can be given the ability to write but not delete data), but I think we can leave them coupled for now.

dcz-self:
I still think that supporting discard doesn't automatically make this a flash-only driver. Discard is also seen in VM images, for example. Naming the API after flash would make less sense than naming an API without discard after HDDs: no discard makes it unsuitable for flash, presence of discard does not make it unsuitable for HDDs (even if today we ban it by decree).

On the other hand, it's a minor thing, and given the embedded focus, it's probably not going to be used on anything but flash anyway.

alistair23:
This shouldn't replace the flash HIL as the lack of multi-write support is very limiting. Adding it as a `block_device` HIL sounds fine to me

phil-levis:
> This shouldn't replace the flash HIL as the lack of multi-write support is very limiting. Adding it as a `block_device` HIL sounds fine to me

It definitely shouldn't replace. The question is whether is it is a flash HIL (as the TRD says it is) or a HIL for more general durable storage.

"Block device" is a technical term, from the original AT&T UNIX and later POSIX (character devices and and block devices). We should not reuse it unless the abstraction is extremely similar and it behaves like a block device, which this does not. If the API described block devices, it should be called `block_device`, but since it doesn't (discard, page vs. block), it shouldn't. E.g., https://docs.oracle.com/cd/E19683-01/806-5222/block-34861/index.html

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// A successful `discard` leaves bytes in the selected `region` undefined.
+    /// The user of this API must not assume any property
+    /// of the discarded bytes.
+    ///
+    /// If `region` exceeds the size of the device, returns `ErrorCode::INVAL`.
+    ///
+    /// Returns `ErrorCode::BUSY` when another operation is in progress.
+    ///
+    /// On success, triggers `WriteableClient::discard_complete` once.
+    fn discard(&self, region: &BlockIndex<D>) -> Result<(), ErrorCode>;
```
bradjc, line 152: The unit of `D` here is bytes?

Doesn't that conflict with `BlockIndex<S>` where `S` is in blocks?

What is a region?

I think we need a glossary in the top comment. We have blocks, write blocks, discard blocks, regions, `S`, `D`, `W`, `R`, at least.

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,247 @@
...
+//!     //(implement associated functions here)
+//! }
+//! ```
+use crate::ErrorCode;
+use core::ops::Add;
+
+/// An index to a block within device composed of `S`-sized blocks.
+/// Stores the number of blocks from the start of the device.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct BlockIndex<const S: usize>(pub u32);
```
bradjc, line 31: The unit of `S` is bytes?

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,247 @@
...
+    fn from(region: BlockIndex<C>) -> Self {
+        AddressRange {
+            start_address: region.0 as u64 * C as u64,
+            length_bytes: C as u32,
+        }
+    }
+}
+
+/// Devices which can read arbitrary byte-indexed ranges.
+pub trait ReadRange {
```
bradjc, line 189: This trait is for a device?

Also, it seems out of place to have byte level reads in a block storage driver.

kernel/src/hil/block_storage.rs
bradjc, line 168: This is in blocks or bytes?

bradjc (commented):
It seems like we 1) really want something like this, and 2) have a hard time agreeing on what that something is. I think the way forward is to, as discussed, not solve every problem, and not over-promise in the API. However, I think the tradeoff is we need more clarity on what this HIL actually does so that a future reader can understand if it will work for their use case or not. Right now the assumptions and intent aren't clear.

Also I get confused about whether this is for flash, or this isn't for flash. I think we either say this is designed explicitly for some subset of flash chips and make it clear that the particulars of the API are based on flash limitations, and name it as such, or we don't, and we remove the subtle references to flash.

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,247 @@
...
+    for T
+{
+}
+
+/// Specifies a storage area with byte granularity.
+pub struct AddressRange {
+    /// Offset from the beginning of the storage device.
+    pub start_address: u64,
+    /// Length of the range.
+    pub length_bytes: u32,
```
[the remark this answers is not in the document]
dcz-self: The name indicates bytes.

kernel/src/hil/block_storage.rs
dcz-self: I am not particularly attached to it, but I've noticed that flash devices include byte addressing mode for reads. Should I get rid of it?

dcz-self:
> The question is whether is it is a flash HIL (as the TRD says it is) or a HIL for more general durable storage.

I'm fine to limit the intentions for this version to raw flash only, for the sake of limiting the scope of discussion. I will definitely not hold myself back in the future when I get a chip with an FTL. I will just reuse this, and try to apply the lessons to refine the API.

After reading http://www.linux-mtd.infradead.org/faq/general.html#L_mtd_vs_hdd , I agree that this is not falling under the strict definition of "block device", although I have no proposals what to call the superset of block device and raw flash. Not that I insist on naming the HIL after that name.

I think the docs clearly indicate flash in this revision. I am happy to rename it to whatever is deemed appropriate.

kernel/src/hil/block_storage.rs
dcz-self: Corrected docstring to bytes.

dcz-self (commented):
I updated the proposal with some new feedback.

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,248 @@
...
+use core::ops::Add;
+
+/// An index to a block within device composed of blocks containing `B` bytes.
+/// Stores the number of blocks from the start of the device.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct BlockIndex<const B: usize>(pub u32);
+
+impl<const B: usize> BlockIndex<B> {
+    /// Returns the index that contains the address.
+    pub fn new_containing(byte_address: u64) -> Self {
```
bradjc, line 35: Why is this `u64`? Tock only supports 32 bit platforms.

kernel/src/hil/block_storage.rs
```diff
@@ -0,0 +1,248 @@
...
+            None
+        }
+    }
+}
+
+impl<const B: usize> From<BlockIndex<B>> for u64 {
+    fn from(index: BlockIndex<B>) -> Self {
+        index.0 as u64 * B as u64
+    }
+}
```
bradjc, line 53: Why do we need this? It's not clear to me that converting a blockindex to a u64 would give the address of the start of the block. Why not something explicit like `BlockIndex.get_start_address()`?

kernel/src/hil/block_storage.rs
bradjc: What about having `ReadableStorageBlocks` and `ReadableStorageBytes`. Then drivers can implement whatever they support.

kernel/src/hil/block_storage.rs
dcz-self: This is to support storage devices with more than 4GiB capacity.

phil-levis:
Comparing this with https://github.com/tock/tock/pull/2248, I think there are relevant differences based on different use cases. Until (unless?) we come up with the "one true flash HIL", I support having multiple HILs that can support different use cases. There is of course the challenge that you can't easily compose the use cases, but you also can't if you only support one of the HILs. :) I'd describe the differences as these:
  - Current API - Supports only page-based access, indicated by a page index. Read, write and erase are all in the same trait.
  - #2248 - Supports only page-based access, uses a `Page` object for verifying valid addresses. No implementations of the trait? Read, write and erase are all in the same trait. Can express that write and erase page sizes can be different (technically, should be block and page).
  - #2993 (this) - Separates trait into (write, erase) and read. Supports arbitrary address range reads, page-sized erases and block-sized writes. 

Here are my suggested changes (open to discussion):
  - [ ] Move these traits into `hil::flash`
  - [ ] Changes names to `Read`, `Modify`, `ReadClient`, and `ModifyClient`
  - [ ] Add `set_client()` methods to `Read` and `Modify`. 
  - [ ] Merge `ReadableStorage`/`Read` and `ReadRange` into a single trait, since they trigger the same callback. Alternatively, make `ReadRange` an extension of `Read`. Otherwise there is a problematic discontinuity, where the runtime behavior of one trait depends on the calls onto another (e.g., I assume that if a client calls `read` then `read_range` the second call will fail with BUSY?). 
  - [ ] Break `BlockIndex` into `BlockIndex` and `PageIndex` rather than just distinguishing on the size -- they really are different abstractions and following flash terminology is helpful
  - [ ] In the `Ok` value of the `read_complete` event, include a `usize` to indicate how much was read. For `read` this will always be R, but for `read_range` it could be something else. This provides a simple way for a client to know exactly how much of the data in `buf` contains read data, rather than forcing the client to track this in internal state.
  - [ ] Change the `flash` documentation to note that there are two APIs, how they differ, and why.

bradjc:
Sounds good to me, but I think we need a name for the new traits, so we don't have confusion about the difference between the current `hil::flash::Flash` and what would be `hil::flash::Read`. Also, since we aren't trying to solve all flash use cases, we should save some namespace for future interfaces.

phil-levis:
> Sounds good to me, but I think we need a name for the new traits, so we don't have confusion about the difference between the current `hil::flash::Flash` and what would be `hil::flash::Read`. Also, since we aren't trying to solve all flash use cases, we should save some namespace for future interfaces.

I don't see a good way to assign names that are simultaneously descriptive for the different APIs and are succinct. So I see two options:
  1. Use proper nouns: call them something like "flash::ainu::" and "flash::beagle::" Authors can decide on the names.
  2. Keep the names as is and have the documentation clearly articulate that certain sets of traits go together and what they are.

Or we can try to for something descriptive, like `flash::PageFlash` and `flash::PageBlockRead`, `flash::PageBlockModify`.

alevy:
This seems worth reviving to me (I have the hardware this was designed for, so I'm biased). This seems to have been very close and the dependent PRs on the surface seem fairly easy to approve other than the HIL discussion.

If @dcz-self has moved on, I'm willing to "take over" incorporating remaining feedback into the code.

dcz-self:
On Tue, 27 Feb 2024 20:19:32 -0800
Amit Levy ***@***.***> wrote:

> This seems worth reviving to me (I have the hardware this was designed for, so I'm biased). This seems to have been very close and the dependent PRs on the surface seem fairly easy to approve other than the HIL discussion.
> 
> If @dcz-self has moved on, I'm willing to "take over" incorporating remaining feedback into the code.
> 

I'm ready to be part of the discussion, but I decided that I need to remove Tock from the dependency path of my project, the reason being that it's missing required functionality that is rather complex in comparison to less powerful OS, and reviewing progresses super slow even if I write it myself.

I'll be watching this to see if things improve.

bradjc:
Looking over this again, I think I'm arriving at the same conclusion that the discussion seemed to arrive at before. Basically, adding a new HIL that is similar to the flash HIL essentially introduces a split in the codebase, which leads to confusion and an inability to layer various components. If we are going to change what we have now, we really need to look at the entire nonvolatile storage stack holistically, how the layers fit together, and specify which interfaces every existing capsule should use. And then update those capsules at the same time.

alevy:
Closing without prejudice for lack of current momentum.

alevy:
Closing without prejudice for lack of current momentum.

PR #2996, opened 2022-03-15
Console: Print the first N bytes of the write buffer, plus doc fixes.

jrvanwhy:
### Pull Request Overview

The console syscall driver's documentation did not say what happens when the number of bytes to write (which is passed to Command) is smaller than the size of the passed buffer. I think everyone reading the documentation assumed that console would write the first N bytes of the buffer, but it actually wrote the last N bytes. This PR alters console's behavior to print the first N bytes, and states that in the documentation.

While editing the documentation, I made a few other fixes:

1. It no longer has a paragraph about buffers being automatically released after use -- that behavior was specific to Tock 1.0.
2. I clarified that console will never perform partial writes. This is to clear up the ambiguity expressed at https://github.com/tock/libtock-rs/pull/395#discussion_r824693201.

### Testing Strategy

I write a `libtock-rs` app that tries to print `7` bytes of `"Hello,\nWorld!\n"`. I ran it (in QEMU on HiFive1) before this PR to confirm it printed `"World!\n"`, made this change, then ran it again with this PR and confirmed it printed `"Hello,\n"`.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/console.rs
+++ b/capsules/src/console.rs
@@ -161,7 +161,7 @@ impl<'a> Console<'a> {
                     .get_readonly_processbuffer(ro_allow::WRITE)
                     .and_then(|write| {
                         write.enter(|data| {
-                            for (i, c) in data[data.len() - app.write_remaining..data.len()]
+                            for (i, c) in data[app.write_len - app.write_remaining..app.write_len]
                                 .iter()
                                 .enumerate()
                             {
--- a/doc/syscalls/00001_console.md
+++ b/doc/syscalls/00001_console.md
@@ -11,10 +11,6 @@ write a buffer, a process must share the buffer using `allow` then initiate the
 write using a `command` call. It may also using `subscribe` to receive a
 callback when the write has completed.
 
-Once the write has completed, the buffer shared with the driver is released, so
-can be deallocated by the process. This also means that it is necessary to
-share a buffer for every write transaction, even if it's the same buffer.
-
 ## Command
 
   * ### Command number: `0`
@@ -33,7 +29,10 @@ share a buffer for every write transaction, even if it's the same buffer.
     At the end of the transaction, a callback will be delivered if the process
     has `subscribed`.
 
-    **Argument 1**: The maximum number of bytes to write.
+    **Argument 1**: The maximum number of bytes to write. If this argument is
+    greater than or equal to the buffer's size, the entire buffer will be
+    written. Otherwise, the first N bytes of the buffer will be written, where N
+    is the value of this argument.
 
     **Argument 2**: unused
 


jrvanwhy:
CI failure appears to be unrelated to this PR's changes.

dcz-self:
This makes the write bytes count in the upcall redundant. Should that stay or get removed?

jrvanwhy:
> This makes the write bytes count in the upcall redundant. Should that stay or get removed?

This PR does not make the written bytes count passed to the upcall any more redundant than it already is. I think it should stay, because removing it would be a breaking change, and there's not much of a benefit to doing so (it *might* save an instruction or two in the kernel).

hudson-ayers (approved):

dcz-self:
This is already a breaking change in that something else is printed than before (then ending, now beginning) even if the binary interface doesn't change, so I wouldn't hold off improvements for that reason.

I'd also say that the byte count was underspecified before, and now the redundancy is explicit. Given that streaming interfaces typically return byte counts when they can stop prematurely, I'm sure I would have been thrown off (again).

jrvanwhy:
> This is already a breaking change in that something else is printed than before (then ending, now beginning) even if the binary interface doesn't change, so I wouldn't hold off improvements for that reason.

My understanding of Tock's policy is that we can only make breaking changes to stabilized portions of the `console` capsule if we increment Tock's major version number, which we don't want to do anytime soon. Whether or not this PR is acceptable depends on whether the Tock core working group considers this a breaking change (a change to the intended behavior) or a bugfix (we changed its behavior to match how it was intended to behave).

alistair23 (approved):

thulithwilfred (approved):
Nice fix!

bradjc (approved):
I consider this a bugfix.

And we need to leave the number of bytes written, since a different console syscall implementation might have a use for it.

jrvanwhy:
> > > And we need to leave the number of bytes written, since a different console syscall implementation might have a use for it.
> > 
> > 
> > This suggests that the console might in fact perform partial writes, and that userspace console libraries need to include retry loops in the case that partial writes are performed. I think part of Johnathan's goal with this PR was to clarify that such retry loops in userspace are not needed (libtock-c does not use one, for example)
> 
> Isn't that a bug in libtock-c? Shouldn't `_write()` return the number of bytes written, and not the number requested? https://github.com/tock/libtock-c/blob/master/libtock/sys.c

That depends on what you think console's API was. If you believe the description in this PR (that console always writes `min(buffer length, command argument)` bytes) is correct, then there's no bug in the code you linked: the number of transmitted bytes is always `count`. If you believe that console can (or could) transmit a smaller number of bytes, then yes it would seem that `libtock` has a bug.

hudson-ayers:
The console driver in the Tock kernel loops until all bytes have been written. So why should libtock-c not treat them as the same? The only reason is that the API docs suggest that the behavior of the console might be different than it's actual implementation. I think this is a documentation bug, and not something that was intentionally decided on at some point. Perhaps I am wrong and the plan is to one day support consoles that do not internally complete writes.

jrvanwhy:
@alevy Ping as discussed during core WG call.

alevy (approved):

PR #2999, opened 2022-03-21
console: Give own buffers to each component

dcz-self:
Without this, naive attempts to create multiple Console instances using the component would result in all of them inoperable due to all using the same buffers.

### Pull Request Overview

This pull request makes it harder to end up with a broken console by making it easier to use ConsoleComponent correctly. Without this change, the user must be aware that all ConsoleComponents share the same set of read/write buffers (layering violation). With this change, each ConsoleComponent gets its own buffer, at the cost of having to call a helper macro – the user has to consciously make the mistake of reusing buffers if they insist.

ConsoleComponent has an explicit `device_num` parameter, which suggests that multiple instances were allowable, plus it's a decent interface to expose generic serial devices to the userspace.

To add weight to this problem, debugging it systematically is rather hard, because when multiple devices are created and the bug is triggered, then – drum roll – debug macros no longer work.

### Testing Strategy

This pull request was tested by running manually.

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/console.rs
+++ b/boards/components/src/console.rs
@@ -13,7 +13,8 @@
 //! let uart_mux = UartMuxComponent::new(&sam4l::usart::USART3,
 //!                                      115200,
 //!                                      deferred_caller).finalize(());
-//! let console = ConsoleComponent::new(board_kernel, uart_mux).finalize(());
+//! let console = ConsoleComponent::new(board_kernel, uart_mux)
+//!    .finalize(console_component_helper!());
 //! ```
 // Author: Philip Levis <pal@cs.stanford.edu>
 // Last modified: 1/08/2020
@@ -27,6 +28,9 @@ use kernel::dynamic_deferred_call::DynamicDeferredCall;
 use kernel::hil;
 use kernel::hil::uart;
 use kernel::static_init;
+use kernel::utilities::static_init::StaticUninitializedBuffer;
+
+use capsules::console::DEFAULT_BUF_SIZE;
 
 pub struct UartMuxComponent {
     uart: &'static dyn uart::Uart<'static>,
@@ -74,6 +78,21 @@ impl Component for UartMuxComponent {
     }
 }
 
+#[macro_export]
+macro_rules! console_component_helper {
+    () => {{
+        use capsules::console::{Console, DEFAULT_BUF_SIZE};
+        use capsules::virtual_uart::UartDevice;
+        use kernel::static_buf;
+        let read_buf = static_buf!([u8; DEFAULT_BUF_SIZE]);
+        let write_buf = static_buf!([u8; DEFAULT_BUF_SIZE]);
+        // Create virtual device for console.
+        let console_uart = static_buf!(UartDevice);
+        let console = static_buf!(Console<'static>);
+        (write_buf, read_buf, console_uart, console)
+    }};
+}
+
 pub struct ConsoleComponent {
     board_kernel: &'static kernel::Kernel,
     driver_num: usize,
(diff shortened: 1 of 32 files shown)


alistair23 (approved):

dcz-self:
It turns out I didn't need to use `static_init_half`.

I put the buffer size inside the component at first because I thought components are the place where recommended values are stored. The capsule doesn't use the buffer size after all, it just accepts whatever buffers it gets. That's moved to the capsule now.

alistair23 (approved):

hudson-ayers:
I am not sure this actually enables re-use of the console component -- calling `finalize()` twice does not lead to two different static globals being created, because a given call to `static_init!()` will only create a single static global variable. The documentation for `static_init()` states:
```
/// As this macro will write directly to a global area without acquiring a lock
/// or similar, calling this macro is inherently unsafe. The caller should take
/// care to never call the code that initializes this buffer twice, as doing so
/// will overwrite the value from first allocation without running its
/// destructor.
```

dcz-self:
Thanks. I restored the original version with `static_init_half`.

dcz-self:
As I wasn't aware how the `static` keyword worked, the original version was also useless. Instead, I went for it and implemented what I actually wanted, and tested it properly, too.

Here's the new patch in action: https://github.com/dcz-self/tock/commit/991b0e6dadcf82e3b965b1eed346b865135e7a51#diff-7aca7e314d2f0d112ca186c195040f488bf4db748f059bdc03729c3c2649e814R331

This change is based on that.

This makes the interface to the component a bit more annoying, because the UartDevice is broken out when a UartMux is in use.

dcz-self:
Given that there are no examples of using `static_buf!` in components, despite being the replacement for `static_init_half`, I decided to give it a shot too.

boards/components/src/console.rs
```diff
@@ -74,48 +97,57 @@ impl Component for UartMuxComponent {
...
+    );
     type Output = &'static console::Console<'static>;
 
-    unsafe fn finalize(self, _s: Self::StaticInput) -> Self::Output {
+    unsafe fn finalize(self, s: Self::StaticInput) -> Self::Output {
         let grant_cap = create_capability!(capabilities::MemoryAllocationCapability);
 
-        // Create virtual device for console.
-        let console_uart = static_init!(UartDevice, UartDevice::new(self.uart_mux, true));
-        console_uart.setup();
```
hudson-ayers, line 106: I am not sure I agree with moving the creation of the virtual device outside of the component. The purpose of components is to enable simple, single-line initialization of capsules for the common case, to reduce code duplication across our many upstream boards. It makes sense to not use a mux for something like GNSS, but that is not a  common use case and could just be initialized without using the component at all. We have run into many users doing things like forgetting to call `setup()` on the virtual device and that being included in the component prevents that error.

I think we should either have a `NonMuxedConsoleComponent` to support things like GNSS, or just have that use case go without a component.

hudson-ayers (changes requested):
This looks mostly good, thank you for providing a template for how to use `static_buf!()`, this component is now much closer to how we imagined they should look when we deprecated `static_init_half!()`. I have one main concern about the component no longer assuming use of a mux.

boards/components/src/console.rs
dcz-self: I brought it back to mux only.

hudson-ayers (approved):
Thanks for your patience with this, and for this very nice change!

bradjc (approved):

alistair23 (approved):

PR #3001, opened 2022-03-24
mpu: Load static apps in the middle of memory regions if needed

dcz-self:
Static processes have a hard-coded RAM placement. Before this change, they would be rejected if the start placement did not lie on a MPU access boundary.

In the case of cortex-m, the boundary is dependent on total RAM size, and that, in turn, on the stack size declared in the tbf header. As a result, changing stack size could turn a working application into a permanently rejected one.

This change adds some padding before the actual application start, to align the memory region with the application size.

### Pull Request Overview

This pull request fixes loading static applications.

Solved: ~~It still fails on making `stack_size!` value too high, so this goes only halfway to fixing the stack problem. It seems that the application neither adjusts the application break automatically, nor does the kernel preemptively set the stack higher.~~

~~As a result, an application with a stack that's above the 1st memory subregion (0x2000 works like a charm) will invariably trigger a fault the moment it tries to use the stack (or at least that's my best guess).~~

~~**I think this MR is still complete even without the above problem being solved.**~~

### Testing Strategy

This pull request was tested by installing various libtock-rust applications on a nrf52840. Tested starts like 0x20006010 and changed application size using elf2tab's `--stack 8192`, and other values.

### TODO or Help Wanted

This pull request still needs testing on risc-v (some things modified), ~~and maybe aswering the question of how to tell the kernel what the stack size is.~~ (solved)

EDIT: I'm also not sure how to test resetting.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/mpu.rs
+++ b/arch/cortex-m/src/mpu.rs
@@ -548,6 +548,7 @@ impl<const NUM_REGIONS: usize, const MIN_REGION_SIZE: usize> mpu::MPU
         &self,
         unallocated_memory_start: *const u8,
         unallocated_memory_size: usize,
+        app_memory_start: Option<*const u8>,
         min_memory_size: usize,
         initial_app_memory_size: usize,
         initial_kernel_memory_size: usize,
@@ -579,15 +580,34 @@ impl<const NUM_REGIONS: usize, const MIN_REGION_SIZE: usize> mpu::MPU
             return None;
         }
 
-        // The region should start as close as possible to the start of the unallocated memory.
-        let mut region_start = unallocated_memory_start as usize;
+        // If the start and length don't align, alter region until it fits
+        let (mut region_start, mut region_size) = match app_memory_start {
+            Some(start) => {
+                // If the app needs a fixed start, put it within the region.
+                let start = start as usize;
+                // Need to take region size into account because it has a minimal size.
+                let region_start = start - (start % region_size);
+                math::extend_to_pow2(
+                    region_start,
+                    cmp::max(start + memory_size, region_start + region_size),
+                )
+            }
+            None => {
+                // The region should start as close as possible to the start of the unallocated memory.
+                // If the start and length don't align, move region towards the end
+                // until it does
+                let start = unallocated_memory_start as usize;
+                let region_start = if start % region_size != 0 {
+                    start + region_size - (start % region_size)
+                } else {
+                    start
+                };
+                (region_start, region_size)
+            }
+        };
 
-        // If the start and length don't align, move region up until it does
-        if region_start % region_size != 0 {
-            region_start += region_size - (region_start % region_size);
-        }
-
-        // We allocate an MPU region exactly over the process memory block, and we disable
+        // We allocate an MPU region starting before or exactly
+        // over the process memory block, and we disable
         // subregions at the end of this region to disallow access to the memory past the app
         // break. As the app break later increases, we will be able to linearly grow
         // the logical region covering app-owned memory by enabling more and more subregions.
@@ -599,7 +619,10 @@ impl<const NUM_REGIONS: usize, const MIN_REGION_SIZE: usize> mpu::MPU
             if initial_kernel_memory_size == 0 {
                 8
             } else {
-                initial_app_memory_size * 8 / region_size + 1
+                let app_start = app_memory_start.map_or(region_start, |a| a as usize);
+                let app_end = app_start + initial_app_memory_size;
+                let initial_size = app_end - region_start;
+                initial_size * 8 / region_size + 1
             }
         };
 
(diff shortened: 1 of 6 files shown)


kernel/src/process_standard.rs
```diff
@@ -1408,50 +1408,12 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
         // Right now, we only support skipping some RAM and leaving a chunk
         // unused so that the memory region starts where the process needs it
         // to.
```
bradjc, line 1410: This comments needs to be updated or removed.

arch/cortex-m/src/mpu.rs
```diff
@@ -579,15 +580,32 @@ impl<const NUM_REGIONS: usize, const MIN_REGION_SIZE: usize> mpu::MPU
...
+                let region_start = start - (start % region_size);
+                math::extend_to_pow2(
+                    region_start,
+                    cmp::max(start + memory_size, region_start + region_size),
+                )
+            }
+            None => {
+                // The region should start as close as possible to the start of the unallocated memory.
+                let start = unallocated_memory_start as usize;
+                let region_start = if start % region_size != 0 {
```
bradjc, line 598:
```suggestion
                // If the start and length don't align, move region up until it does
                let region_start = if start % region_size != 0 {
```

phil-levis:
*This still has the nasty property that some stack sizes break the application if it's too close after memory which is unavailable for allocation.*

I don't quite understand your example. I fully believe there might be a problem, but I can't from your description understand it completely. Could you expand your explanation?

hudson-ayers:
I wonder whether we should instead be trying to update libtock-{rs,c} to error if an application of size X requests a memory location Y where Y is not a valid location for an MPU region of size at least X.

I understand the perks of doing this is in the kernel (users don't always have to manually update the memory location an application is being built for when they change its size), but at the end of the day it will still be possible to choose an invalid fixed address for apps:
-  libtock-rs has to track the size of the kernel which changes over time
- an increase in size could still conflict with other fixed size apps already installed

Rather than add logic to the kernel which all boards have to pay the size/cycle/complexity cost for, I would like if we just took away the "surprise" that you encountered here where increasing the stack size led to an app no longer working, by catching and explaining the error at link-time. I am generally reluctant to add overhead and complexity for a use case which we still hope to not support super long-term (fixed location apps).

If doing this in the linker is very difficult I think this idea might be fine as an alternative, but I want a chance to explore that option

dcz-self:
> update libtock-{rs,c} to error if an application of size X requests [invalid placement]

I think that direction is all right, and even does not go far enough. If the process knows which placements are wrong, it might as well be used to choose a correct placement.

The only difficulty is that some inputs (program size) are not known until linking is finished, so linking has to be done twice.

kernel/src/process_standard.rs
dcz-self: Removed. It did not apply any longer.

arch/cortex-m/src/mpu.rs
dcz-self: Altered.

dcz-self:
I have an idea about this where I interject in the linking process, link once with the default placement, and check if it looks reasonable. If not, link again with a better placement.

This means that all the positioning is offloaded to the compilation step, so the kernel doesn't have to accept apps at misaligned locations.

As a downside, the user can no longer choose the linker using .cargo/config. Would have to be done using an env var I think.

Does that sound like an idea worth exploring?

hudson-ayers:
> I have an idea about this where I interject in the linking process, link once with the default placement, and check if it looks reasonable. If not, link again with a better placement.
> 
> This means that all the positioning is offloaded to the compilation step, so the kernel doesn't have to accept apps at misaligned locations.
> 
> As a downside, the user can no longer choose the linker using .cargo/config. Would have to be done using an env var I think.
> 
> Does that sound like an idea worth exploring?

I definitely think this sounds like an idea worth exploring. Tock is generally anti `.cargo/config` as a result of a distaste for configuration in hidden files anyway. Also, I do not believe that Tock at any point has promised that any linker other than lld will work correctly for libtock-rs apps, or that any linker other than ld will work for libtock-c apps.

As an aside, I understand you have been attempting to use a different linker for libtock-rs apps -- can you say why you wanted to do so?

dcz-self:
I figured that the gnu linker is not going to hurt, assumed that linkers should not matter, so filed reports :)

Later I got the idea that I don't like non-PIC linking. I intend to distribute applications as object files and then link them at the time of flashing, at the exact right positions. Using the gnu linker is much easier because it's way more widespread than the rust linker, and installed in a standard way. So I still want to make libtock-rs compatible with it – as long as PIC is not on the horizon.

bradjc:
I think we are going to hold off on this for now. We certainly have a delicate balance between what we expect userspace and userspace toolchains to handle, and what the kernel should handle as far as managing app addresses. But, PIC has always been the expectation (and in fact when we figured out we could do PIC on arm-gcc was what really caused tock development to start in earnest). I think our current balance of supporting fixed address apps but expecting the app loader to do most of the work is a reasonable compromise until we hopefully have full PIC support someday.

Issue #2981, opened 2022-02-28
Kernel ABI for 64-bit systems


jettr:
How do we want to handle the kernel/user ABI for 64-bit systems using Tock OS? This has come up while implementing a host emulated version of Tock OS and application on a 64-bit linux machine.

Most of the kernel/user ABI is defined using `usize` or `Register` nomenclature ([example][2]), but the newer [`SyscallReturnVariant`][0] enum specifically calls out `u32` variants instead of `usize` variants. Should, for example, `FailureU32` instead be `FailureUsize`?

Otherwise, when the kernel passes back architecture-sized data (e.g. pointer), the correct `SyscallReturnVariant` variant is difficult to chose in an platform independent manner.

Another option would be to add `usize` variants to `SyscallReturnVariant`, but then that gets confusing for 32-bit system to chose the correct return variant.

[TRS 104][1] specifically calls out that it only applies to "32-bit ARM Cortex-M and RISC-V RV32I", but the definitions in that TRD seem to have directly led to the `SyscallReturnVariant` enum, which exist in a platform independent location of the kernel.

Are there any other documents or discussions on this topic that I can gain some insight from to help forge a path forward?

[0]:https://github.com/tock/tock/blob/master/kernel/src/syscall.rs#L221
[1]:https://github.com/tock/tock/blob/master/doc/reference/trd104-syscalls.md
[2]:https://github.com/tock/tock/blob/master/kernel/src/syscall.rs#L159

lschuermann:
This has come up during the design of the Tock 2.0 system call interface.

> [`SyscallReturnVariant`](https://github.com/tock/tock/blob/master/kernel/src/syscall.rs#L221) enum specifically calls out `u32` variants instead of `usize` variants

This is deliberate choice we've made for the Tock 2.0 system call interface. While an `allow`'s pointer and length, as well as a `subscribe`'s callback pointer and user data (most likely a pointer) are clearly platform dependent and thus must be represented through the platform's native integer width, `command` has a different story. While capsules may in certain cases want to communicate types of the platform's native integer width, fundamentally the `command` interface is simply a mechanism to convey data from a userspace application to a capsule, and provide a return value as feedback of that operation. It is of major importance that the ABI provided by capsules does not implicitly change between platforms. While in practice transitioning from 32-bit to 64-bit is seldom going to cause issues, a capsule using native integer widths, written for a 64-bit system, can very likely not work on a 32-bit system.

[...]

PR #3003, opened 2022-03-30
Documentation: Allow pointer success variant for Command return

jettr:
In support of #2981, update the TRD for Command to allow the kernel to pass back a pointer as a success variant.  Also we should treat a pointer are wholly different from either `u32` or `u64` values to avoid confusion on which success variant to select. Adding a pointer success variants helps common capsule code that needs to returns a pointer back to users space to be generic on architecture size.

The main driving use case for this change is having a kernel owned buffer that a capsule can lease to an application for it to read and write to. Another potential use cases for a pointer success variant is for a system information capsule where applications could request starting/ending pointers of their resources (e.g. flash and ram).

Returning a pointer from the kernel seems like a very well defined behavior especially since the kernel provides no guarantees  on the accessibility of returned pointers; only the driver implementation itself can provide any guarantees.

### DIFF
--- a/doc/reference/trd104-syscalls.md
+++ b/doc/reference/trd104-syscalls.md
@@ -5,10 +5,10 @@ System Calls
 **Working Group:** Kernel<br/>
 **Type:** Documentary<br/>
 **Status:** Draft <br/>
-**Author:** Hudson Ayers, Guillaume Endignoux, Jon Flatley, Philip Levis, Amit Levy, Leon Schuermann, Johnathan Van Why <br/>
+**Author:** Hudson Ayers, Guillaume Endignoux, Jon Flatley, Philip Levis, Amit Levy, Leon Schuermann, Johnathan Van Why, Jett Rink <br/>
 **Draft-Created:** August 31, 2020<br/>
-**Draft-Modified:** November 23, 2021<br/>
-**Draft-Version:** 6<br/>
+**Draft-Modified:** March 30, 2022<br/>
+**Draft-Version:** 7<br/>
 **Draft-Discuss:** tock-dev@googlegroups.com</br>
 
 Abstract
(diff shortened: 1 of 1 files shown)


phil-levis:
I'd like to separate the use case from the proposed change, because I think it would be a big mistake to make this change for this use case. Adding a return type to the kernel and changing the syscall ABI to support 64-bit emulation of a 32-bit platform which has a command that doesn't really follow TRD 104 seems like the tail wagging the dog. We shouldn't make architectural changes to the kernel to support esoteric, non-standard use cases that don't really follow 104.

That being said, I think there could be a use for such a return type. Particularly, it could make sense for memop. Currently, the ABI assumes 32 bits. So memop returns a `u32`. So I have a couple of questions:

  1. How does this emulation platform handle memop (or does it not)?
  2. How does the emulation platform handle `allow`? The code casts pointers to a `u32`: https://github.com/tock/tock/blob/108dbf723c25a5e645161028ad28904cc3ebb1c9/kernel/src/syscall.rs#L430
  3. How does the emulation platform handle `subscribe`?  Same problem as above.

It seems like supporting 64 bits would require many more changes than this one, although this one might be the first you are running into. I'm uncomfortable with applying a series of architectural changes to handle little edge cases.

What does the emulation platform look like? Is it possible for the test harness to figure out 64-bits in userspace and then use the low order bits passed in the register to construct the 64-bit address in userspace?

jettr:
I think it is much bigger ask/change to start defining `memop`, `subscribe`, and `allow` in terms of newly defined `sucess_pointer`, `success_pointer_pointer`, `failure_pointer_pointer` return variants than the existing `success_u32`, `success_u32_u32`, and `failure_u32_u32`. But I do think that would be a more consistent change and sets up for a 64-bit ABI quiet nicely.

Right now for host emulation, we can know that the data passed through for these explicit memory operations contain pointers, so we can transport them as `usize` to/from the kernel using the existing return `SyscallReturn` ids -- even though it is technically (wrong)/(undefined for 64-bit ABI). The kernel does stores the pointers as pointers (e.g  `*const ()`) instead of `u32` for most variants (e.g `SubscribeSuccess` and `AllowReadWriteSuccess`) until the last mile conversion, so the data isn't lost until the `SyscallReturn::encode_syscall_return` call. The host emulation version of `UserspaceKernelBoundary::set_syscall_return_value` does not need to use the default implementation of `SyscallReturn::encode_syscall_return` that truncates `*const ()` into `u32`. For commands returns however, there is no variant that holds a pointer until the last mile, that is why I would like to add one.

phil-levis:
I think we agreed to not go down this path. Closing this.

jettr:
Agreed. I just closed the issue instead both the issue and the pull. Thanks for the clean up!

PR #3008, opened 2022-04-02
debug: Add dbg! from std

dcz-self:
### Pull Request Overview

This pull request adds a dbg!() macro because typing `let foo = &a; debug!("{:?}", foo);` and breaking expressions to insert the statement for the hundredth time gets boring.

Shamelessly stolen from rust std.

### Testing Strategy

This pull request was tested by using it to test the sma_q3 integration on a non-upstreamed branch https://github.com/dcz-self/tock/tree/flashil

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alistair23 (approved):

hudson-ayers (changes requested):
I understand this name is taken directly from `std`, but I think we should use a different one here because of the confusing collision with `debug!()` in Tock, which does something very different.

dcz-self:
I considered that the names are confusingly similar, but the only reasonable idea I got was to rename "debug" to "println" instead, for the sake of consistency. Any other solutions?

hudson-ayers (approved):
I don't feel strongly either way about println!(), will let those who know the history of why Tock settled on debug!() comment on that

dcz-self:
I added an ignore to the docs, because trying to print in the doctest would fail due to `DEBUG_WRITER` being None.

I'm not sure how to fill that in without compilation conditional on the testing more, so I dropped the doctest.

brghena:
> I don't feel strongly either way about println!(), will let those who know the history of why Tock settled on debug!() comment on that

I'm fairly certain that `debug!()` came about organically and without knowledge of `std` debugging mechanisms. But since we're here, `debug_exp!()` makes sense to me.

brghena (approved):

bradjc:
Prefacing all debugging calls with `debug` makes sense to me: they are macro hacks useful for debugging. We don't have a normal stdout, and I don't think kernel could should be calling something like `println!()` in normal operation.

dcz-self:
"dbg" in stdlib is a kind of an exception: there are not many names that are contractions. While I don't know what the reasons for choosing it were, I appreciate that its invokation is only 6 characters to type. Having any name starting with a full "debug_" no longer has this advantage (but I'll gladly take it anyway).

hudson-ayers (approved):

lschuermann (approved):

lschuermann:
I realize this macro is essentially copied verbatim from the Rust standard library, nonetheless it might be interesting to see how this changes in future Rust versions / editions given [the current `match` reference lifetime behavior and the confusion stemming from that](https://fasterthanli.me/articles/a-rust-match-made-in-hell). We might need to look out to keep this macro in sync with the standard library, if something changes there. Perhaps this won't (need to) change at all, just watching out for that.

bradjc (approved):

Issue #3006, opened 2022-04-01
Screen HIL bit packing


dcz-self:
The screen HIL specifies three things that interact awkwardly;

- 1 bit per pixel mode
- selection frame that can have arbitrary bounds (including odd)
- write commands taking bytes

There are multiple ways to pack pixels into the wrrite call when the frame width is not mod 8. An example with 3 mono pixels wide. Does it get written like:

```
____
|123|
|456|
----
```

or like this, ignoring the extra bits:

```
____
|123|45678
|123|45678
----
```

Currently, the packing is undefined. What should it be?

alistair23:
@alexandruradovici

dcz-self:
Please take a look here, I addressed that and some other issues: https://github.com/tock/tock/pull/3011

dcz-self:
The reason I'm pointing out this issue is that the HIL is meant to abstract the underlying hardware somehow (if I understand what a HIL is for), and tying the interpretation of buffers means that the users are unable to submit buffers of correct length and shape unless they know which device will be used. That makes the HIL specific to one device, and supporting formats like monochrome becomes pointless if the device doesn't support them.

PR #3009, opened 2022-04-03
WIP: Add driver for LPM013M126 display

dcz-self:
### Pull Request Overview

This pull request adds a frame buffer driver for LPM013M126 graphical display.

### Testing Strategy

This pull request was tested by using in the sma_q3 branch https://github.com/dcz-self/tock/tree/flashil

### TODO or Help Wanted

This pull request still has some problems, even though it works.

The worst problem is that any graphical operation will always submit the entire buffer due to how the screen hil and the display command set interact with DMA buffer passing.

The issue is twofold. First, the display can only submit entire lines. There is no accommodation for this in the screen HIL – buffers are arbitrary pixels wide – and so, the driver needs to store each line completely to submit segments that are not being written at the moment.

Second, even though the display can receive one line at a time, the HIL for SPI does not allow a start offset but only an end limit then passing the &'static mut [u8]. When a line in the middle of the buffer needs to be updated, we could pass &buf[10..] to SPI, but when the SPI finishes, we need to receive &buf, but all we've got is &buf[10..]. This means the buffer would be single-use only.

Could we consider wrapping the &static mut [u8]s in some object that contains their length and offset? I've noticed that the length is usually passed through anyway.

* * *

A minor problem is that the screen HIL allows updating single pixels by passing byte-sized buffers. There's no opt-out, meaning that this driver cannot comply to the screen HIL: the device *does not* have a mode that aligns pixels on byte boundaries (only 1, 3, 4bit), and wasting more space/time to pass oversized buffers is not what I want to do. This relates to https://github.com/tock/tock/issues/3006

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


capsules/src/lpm013m126.rs
```diff
@@ -0,0 +1,487 @@
...
+            buffer: TakeCell::empty(),
+            client: OptionalCell::empty(),
+            state: Cell::new(State::Uninitialized),
+        }
+    }
+
+    pub fn initialize(&self) -> Result<(), ErrorCode> {
+        match self.state.get() {
+            State::Uninitialized | State::Bug => {
+                // Even if we took Pin type that implemets Output,
```
bradjc, line 208:
```suggestion
                // Even if we took Pin type that implements Output,
```

capsules/src/lpm013m126.rs
```diff
@@ -0,0 +1,487 @@
...
+            disp,
+            extcomin,
+            frame_buffer: OptionalCell::new(FrameBuffer::new(frame_buffer)),
+            buffer: TakeCell::empty(),
+            client: OptionalCell::empty(),
+            state: Cell::new(State::Uninitialized),
+        }
+    }
+
+    pub fn initialize(&self) -> Result<(), ErrorCode> {
```
bradjc, line 205: Can you add a comment explaining when initialize should be called? It appears that initialize can be safely called during the board's init process, even if the screen is never used, but it would be good to clarify.

capsules/src/lpm013m126.rs
```diff
@@ -0,0 +1,487 @@
...
+        };
+
+        if let Some(new_state) = new_state {
+            self.state.set(new_state);
+        }
+        // Thankfully, this is the only command that results in the callback,
+        // so there's no danger that this will get attributed
+        // to a command that's not finished yet.
+        self.client.map(|client| client.command_complete(Ok(())));
+        ret
```
bradjc, line 319: This doesn't quite look right. Callbacks cannot be called from HIL commands, as this breaks Tock's control flow model (i.e. the caller of `screen.set_write_frame()` can expect that its code completes before the callback function is called). If this callback is needed then it needs to be triggered originating from another callback (such as alarm.fired() for example).

capsules/src/lpm013m126.rs
```diff
@@ -0,0 +1,487 @@
...
+
+        self.buffer.replace(buffer);
+        match self.state.get() {
+            State::Writing(..) => {}
+            _ => {
+                self.client.map(|client| {
+                    self.buffer
+                        .take()
+                        .map(|buffer| client.write_complete(buffer, ret))
+                });
```
bradjc, line 363: Same here, although it's not clear to me why a write_complete callback would be triggered while the driver is still in the `Writing` state.

capsules/src/lpm013m126.rs
dcz-self: Is there some "idle" callback triggered when the kernel has nothing else to schedule? Should I try adding one?

Getting a timer involved for what should be synchronous sounds like an overkill, both in terms of code as well as resources.

alexandruradovici: You can use a deferred call, similar to the one used by this driver (https://github.com/WyliodrinEmbeddedIoT/getting-started-with-secure-embedded-systems/blob/main/chapter_10/kernel/drivers/src/led_matrix_text.rs), to schedule an upcall after your function return.

dcz-self:
Updated to remove most warnings and also use DeferredCall in order to inform about events that complete synchronously.

DeferredCall requires the client to be 'static, so I half-gave up and added `where Self: 'static` everywhere.

dcz-self:
Actually tested now under https://github.com/dcz-self/tock/tree/flashil

capsules/src/lpm013m126.rs
```diff
@@ -0,0 +1,587 @@
...
+        if frame.column != 0 {
+            // Can't be arsed to bit shift pixels…
+            panic!("Horizontal offset not supported");
+        }
+        let rows = (frame.row)..(frame.row + frame.height);
+        // There are 8 pixels in each row per byte.
+        let sources = buffer.chunks(frame.width as usize / 8);
+        for (i, source) in rows.zip(sources) {
+            let row = self.get_row_mut(i);
+            debug!("rowl {} soul {}", row.len(), source.len());
```
alistair23, line 74: Can we remove the `debug!()`?

capsules/src/lpm013m126.rs
```diff
@@ -0,0 +1,587 @@
...
+    ) -> SubmitBuffer<'static> {
+        // HILs typically can't use offsets :/
+        // Best we can do is limit length (TODO)
+        frame_buffer.data
+    }
+}
+
+/// Modes are 6-bit, network order.
+/// They use a tree-ish encoding, so only the ones in use are listed here.
+#[allow(dead_code)]
```
alistair23, line 107: Is this unused?

capsules/src/lpm013m126.rs
```diff
@@ -0,0 +1,587 @@
...
+
+    /// Needed forinit and to flip the EXTCOMIN pin at regular intervals
+    alarm: &'a A,
+}
+
+impl<'a, A: Alarm<'a>, P: Pin, S: SpiMasterDevice> Lpm013m126<'a, A, P, S>
+where
+    Self: 'static,
+{
+    /// Caution: passing `frame_buffer` that's too small will panic.
```
alistair23, line 195: Could we add a check for this instead?

capsules/src/lpm013m126.rs
dcz-self: The BlinkInversion variant is present for reference but unused here. The list is not exhaustive anyway (the encoding does not map so well into a list), should unused entries get removed?

capsules/src/lpm013m126.rs
dcz-self: Not quite sure what was meant, so I moved the panic to the component.

alistair23: I mean instead of panincing can we just return an error

dcz-self: So that's what I did: the new() returns an error, and then the component panics.

capsules/src/lpm013m126.rs
alistair23: I think it's probably best to remove the unused entries

bradjc (approved):

capsules/src/lpm013m126.rs
dcz-self: Instead of adding a comment here, I added it to setup, and un-pub'ed initialize.

dcz-self:
I updated this to use the updated Screen HIL including set_power.

Also, added some fixes to bugs revealed by using it here https://framagit.org/jazda/tock

Notably, the blit function was buggy and prone to crashing.

bradjc (approved):

Issue #1112, opened 2018-07-11
Tracking: Initialize in HIL interfaces


ppannuto:
This issue tracks implementation progress for RFC #1035, removing `initialize` from HIL interfaces.

- [ ] DAC
- [ ] LED
- [ ] Radio
- [ ] SPI
- [ ] UART: #1073 

---

- [ ] Final CHANGELOG summary of the horizontal changeset and documentation for chip authors.

PR #3010, opened 2022-04-03
Software implementation of SHA-256 in a capsule

phil-levis:
### Pull Request Overview

This pull request adds a software implementation of SHA256 as a capsule. The implementation relies in a deferred call to trigger callbacks. It includes a test, which passes. It adds a commented out line to the imix `main.rs` to include the test.

I wrote this capsule so we have a software cryptographic hash implementation that can be used on any platform. This allows testing process integrity checking and inserting credentials without requiring public keys. Public key signatures will be the next step.

This a draft because it currently only supports a single call to `add_data` per hash. You cannot call `add_data` multiple times.

### Testing Strategy

This pull request was tested by running `sha256_test::run_sha256` on the imix platform. The current test tests a multi-block hash; previous versions tested a single-block hash. When this passed I updated the test to a multi-block hash. The correct hashes were generated from https://www.movable-type.co.uk/scripts/sha256.html

### TODO or Help Wanted

We need to decide on the calling semantics of `DigestData::clear_data`.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


capsules/src/sha256.rs
```diff
@@ -0,0 +1,400 @@
+//! Software implementation of SHA-256. Implementation is based on the
```
bradjc, line 1:
```suggestion
//! Software implementation of SHA-256.
//!
//! Implementation is based on the
```

bradjc (commented):
Looks good to me, although it might be good to remove the debugs before merging.

phil-levis:
Yes, definitely -- it's still in draft. I've completed and tested the arbitrary `add_data` cases (e.g., which require internal software buffer). I do one more cleanup then transition the PR from draft.

capsules/src/test/sha256.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+impl digest::ClientVerify<'static, 32> for TestSha256 {
+    fn verification_done(
+        &'static self,
+        result: Result<bool, ErrorCode>,
+        compare: &'static mut [u8; 32],
+    ) {
+        self.hash.put(Some(compare));
+        debug!("Sha256Test: Verification result: {:?}", result);
```
alistair23, line 74: Tests should panic if they fail, that way they will cause failures in the CI

boards/imix/src/test/sha256_test.rs
```diff
@@ -0,0 +1,59 @@
+//! This tests a software SHA256 implementation. To run this test,
+//! add this line to the imix boot sequence:
+//! ```
+//!     test::sha256_test::run_sha256();
```
alistair23, line 4: Isn't this missing an argument?

boards/imix/src/test/sha256_test.rs
```diff
@@ -0,0 +1,59 @@
...
+//! hashes correctly. This string is 12 repetitions of "hello ", so is
+//! 72 bytes long. As SHA uses 64-byte/512 bit blocks, this verifies
+//! that multi-block hashes work correctly.
+
+use capsules::sha256::Sha256Software;
+use capsules::test::sha256::TestSha256;
+use kernel::dynamic_deferred_call::DynamicDeferredCall;
+use kernel::static_init;
+
+pub unsafe fn run_sha256(call: &'static DynamicDeferredCall) {
```
alistair23, line 24: Can we set this up to match the Tock test infrastructure?

As this isn't board specific (it can run on any board) it also might be worth making it more generic

capsules/src/sha256.rs
```diff
@@ -0,0 +1,402 @@
...
+        }
+    }
+
+    pub fn initialize(&self) -> Result<(), ErrorCode> {
+        if !self.busy() {
+            self.hash_values.set([
+                0x6a09e667, 0xbb67ae85, 0x3c6ef372, 0xa54ff53a, 0x510e527f, 0x9b05688c, 0x1f83d9ab,
+                0x5be0cd19,
+            ]);
+            self.round_constants.map(|k| {
```
alistair23, line 78: Why not set this in `new()`? Do we expect to call `initialize()` multiple times? That's worth documenting then

capsules/src/sha256.rs
```diff
@@ -0,0 +1,402 @@
...
+                    self.input_data.set(data);
+                    self.compute_sha256();
+                });
+                Ok(len)
+            }
+        }
+    }
+
+    fn clear_data(&self) {
+        let _ = self.initialize();
```
alistair23, line 293: This won't clear the data if we are busy. The lack of errors indicates that this should clear the state even if the implementation is busy

boards/imix/src/test/sha256_test.rs
phil-levis: How would you not pass DynamicDeferredCall? I think that's necessary. The issue is it's software but needs to issue callbacks.

phil-levis: I'm happy for this test to be generalized to other platforms, but I currently only have an imix, so would prefer to leave that to someone else.

capsules/src/sha256.rs
phil-levis: Because it needs to be invoked in `clear_data`. Initialization of the hash values is a per-hash operation.  I could factor out the round constants to `new` or make them const data. The latter probably makes more sense/?

capsules/src/sha256.rs
phil-levis: The issue I saw here is there's no return value, so its behavior needs to be deterministic. This will clear the data if it's busy? This method can't be called in the middle of an operation. It will possibly clear data that has been filled in through a SHA but whose callback is pending through the deferred call. The `DigestData` HIL doesn't say what should happen here. So I went for clearing what's in state.

phil-levis: The HIL should probably have a `Result` here and say what happens.

capsules/src/sha256.rs
phil-levis: Changed it to const data.

boards/imix/src/test/sha256_test.rs
alistair23: You would do it by creating `Sha256Software` outside the test framework, in `main.rs`

That would mean we can use the generic test framework but also expose SHA support to userspace

capsules/src/sha256.rs
alistair23: If the callback is pending then we should clear the data and set `State::Idle`. The idea with `clear_data()` is that it can't fail, we need the data cleared when we call it

phil-levis: What about the callback? I think we want to specify this in terms of the behavior as a caller sees it (design) rather than what the capsule does (implementation). So, in particular, if there are operations pending what happens to their callbacks? What error values do they return?

You can't move to the idle state because you might leak memory. E.g., 
```rust
sha.verify(verify_buf);
sha.clear_data(); 
sha.verify(verify_buf2); // verify_buf lost here?
```

alistair23: After `clear_data()` the implementation should trigger a callback with the `ErrorCode::CANCEL` to return the buffer

phil-levis: > After `clear_data()` the implementation should trigger a callback with the `ErrorCode::CANCEL` to return the buffer

This doesn't quite agree with TRD2. 

https://github.com/tock/tock/blob/master/doc/reference/trd2-hil-design.md#rule-3-split-phase-result-values-indicate-whether-a-callback-will-occur

The point I'm getting at is that `clear_data` needs to return something. It sounds like you are arguing that case 1 (not cancelled) should not occur. The example here would be that a client calls `hash`, the implemention completes the hash and pushes a deferred procedure call with the completed result, then the client calls `clear_data`. In this case, the hash would just be thrown away and `hash_done` would have an ECANCEL? In this case I think that's feasible (there are others when it's not, e.g., communication, you want to signal that the communication was in fact sent).

capsules/src/sha256.rs
```diff
@@ -0,0 +1,417 @@
...
+            });
+            self.hash_values.set([
+                0x6a09e667, 0xbb67ae85, 0x3c6ef372, 0xa54ff53a, 0x510e527f, 0x9b05688c, 0x1f83d9ab,
+                0x5be0cd19,
+            ]);
+            Ok(())
+        } else {
+            Err(ErrorCode::BUSY)
+        }
+    }
```
alistair23, line 103: We have been trying to remove [`init()` calls ](https://github.com/tock/tock/issues/1112), why not just call this in new()?

capsules/src/sha256.rs
```diff
@@ -0,0 +1,417 @@
...
+
+    // Complete the hash and produce a final hash result.
+    fn complete_sha256(&self) {
+        let mut buffered_length = self.buffered_length.get();
+        // This shouldn't be necessary, as temp buffer should never be
+        // full. But if it is fill, appending the 1 will be an
```
alistair23, line 109: s/fill/full/g

capsules/src/sha256.rs
```diff
@@ -0,0 +1,417 @@
...
+            self.data_buffer.map(|b| {
+                for i in buffered_length..SHA_BLOCK_LEN_BYTES {
+                    b[i] = 0;
+                }
+            });
+        }
+
+        self.data_buffer.map(|b| {
+            // Append the 1
+            b[buffered_length] = 0x80;
```
alistair23, line 131: Why not just use `get()` here and then you don't need the checks above? That also removes a panic :)

capsules/src/sha256.rs
```diff
@@ -0,0 +1,417 @@
...
+
+    // This method computes SHA256 on data in input_data,
+    // updating the internal hash state. `data_buffer`
+    // contains input data that did or does not fill a block:
+    // the implementation first fills temp_buffer and computes
+    // on it, then operates on input_data. If the end of
+    // input_data does not complete a block then the remainder
+    // is stored in data_buffer.
+    fn compute_sha256(&self) {
+        let mut data = self.input_data.take().unwrap();
```
alistair23, line 171: Let's use `if let` instead of `unwrap()`ing, avoiding a panic

capsules/src/sha256.rs
```diff
@@ -0,0 +1,417 @@
...
+    fn add_data(
+        &self,
+        data: LeasableBuffer<'static, u8>,
+    ) -> Result<usize, (ErrorCode, &'static mut [u8])> {
+        if self.busy() {
+            Err((ErrorCode::BUSY, data.take()))
+        } else {
+            self.state.set(State::Data);
+            if self.handle.is_none() {
+                Err((ErrorCode::FAIL, data.take()))
```
alistair23, line 295: Won't this leave the state stuck in `Data`?

capsules/src/sha256.rs
alistair23: The idea behind not returning an error is that `clear_data()` can't fail. Let's say the application crashes, it wants to call `clear_data()` to clear any secrets and know that it has been cleared. So yes, the not cancelled case should not occur.

Exactly, in that case `hash_done()` should be called with `ECANCEL`.

The current lowRISC implementation doesn't do that (it just wipes everything), so we loose the buffer (that's a bug) but it doesn't leak information by not cancelling the operation.

If we do have a good reason to allow cancels to fail, then the HIL needs to be modified to reflect that

capsules/src/sha256.rs
```diff
@@ -0,0 +1,417 @@
...
+    fn run(
+        &'a self,
+        digest: &'static mut [u8; 32],
+    ) -> Result<(), (ErrorCode, &'static mut [u8; 32])> {
+        if self.busy() {
+            Err((ErrorCode::BUSY, digest))
+        } else {
+            self.state.set(State::Hash);
+            if self.handle.is_none() {
+                Err((ErrorCode::FAIL, digest))
```
alistair23, line 323: Same here, won't the state be forever `Hash`

capsules/src/sha256.rs
```diff
@@ -0,0 +1,417 @@
...
+impl<'a> DynamicDeferredCallClient for Sha256Software<'a> {
+    fn call(&self, _handle: DeferredCallHandle) {
+        let prior = self.state.get();
+        self.state.set(State::Idle);
+        match prior {
+            State::Idle => {}
+            State::Verify => {
+                // Do the verification here so we don't have to store
+                // the result across the callback.
+                let output = self.output_data.replace(None).unwrap();
```
alistair23, line 381: Can we remove the `unwrap()`

capsules/src/test/sha256.rs
```diff
@@ -0,0 +1,120 @@
...
+        let position = self.position.get();
+        if position != data.len() {
+            // More to input
+            let next_position = cmp::min(position + CHUNK_SIZE, data.len());
+            self.position.set(next_position);
+            let mut buffer = LeasableBuffer::new(data);
+            buffer.slice(position..next_position);
+            let r = self.sha.add_data(buffer);
+            if r.is_err() {
+                debug!("Sha256Test: failed to add data: {:?}", r);
```
alistair23, line 74: Same comment about `panic!()`ing

boards/imix/src/main.rs
```diff
@@ -702,6 +702,8 @@ pub unsafe fn main() {
     );*/
     //virtual_alarm_timer.set_alarm_client(mux_timer);
 
+    test::sha256_test::run_sha256(dynamic_deferred_caller);
```
alistair23, line 705: If we use the generic test framework here then the test won't be run on every boot (where users might not want to waste time running SHA tests)

alistair23 (changes requested):

capsules/src/sha256.rs
phil-levis: I put this in a separate method because this functionality is also needed by `clear_data`. But I agree, this should not be a `pub fn`, instead `new` should invoke it.

capsules/src/sha256.rs
phil-levis: I don't understand. Can you explain more?

capsules/src/sha256.rs
phil-levis: Good catch.

capsules/src/sha256.rs
phil-levis: I'm of the opinion that there should be a fail fast here. If the unwrap isn't possible, you can't invoke the callback. If you can't invoke the callback, the state machine is wedged. This should panic, as it implies an internal error that is going to hang the system.

capsules/src/sha256.rs
phil-levis: With the clear_data semantics of never-fail (and an Err(ErrorCode::CANCEL)), this state machine is now different. The `call` deferred call was setting state to `Idle`.

capsules/src/test/sha256.rs
phil-levis: Fixed.

capsules/src/test/sha256.rs
phil-levis: Fixed.

capsules/src/sha256.rs
phil-levis: No -- `call` moves it to `Idle`.

boards/imix/src/test/sha256_test.rs
```diff
@@ -0,0 +1,63 @@
...
+    let t = static_init_test_sha256(call);
+    t.run();
+}
+
+// HSTRING is "hello world" and HHASH is the SHA-256 hash of this string.
+pub static mut HSTRING: [u8; 11] = [
+    'h' as u8, 'e' as u8, 'l' as u8, 'l' as u8, 'o' as u8, ' ' as u8, 'w' as u8, 'o' as u8,
+    'r' as u8, 'l' as u8, 'd' as u8,
+];
```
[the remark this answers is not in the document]
phil-levis: Good idea. Done.

boards/imix/src/main.rs
phil-levis: Where is the generic test framework documented, so I can use it? This is the way that all of the tests are currently done on imix. Or are you suggesting I incorporate a test on QEMU so it can be tested in the CI?

boards/imix/src/test/sha256_test.rs
phil-levis: Is there documentation on how to correctly write a test?

alistair23: Yep! There is heaps

https://doc.rust-lang.org/book/ch11-01-writing-tests.html is a good start
and https://os.phil-opp.com/testing/ has good information on an embedded setup

capsules/src/sha256.rs
alistair23: We don't do it everywhere in Tock, but for a single allocation like this it's probably a good idea.

You can see an example here: https://doc.rust-lang.org/std/primitive.slice.html#method.get_mut
and Hudson has done a great job explaining it here: https://github.com/tock/tock/blob/master/doc/CodeSize.md#panics

phil-levis: Done.

boards/imix/src/main.rs
phil-levis: The intention is this test is commented out for the final PR.

phil-levis: Removed.

boards/imix/src/test/sha256_test.rs
alistair23: Ah, ok. We don't have documentation on how to write one, just examples.

The [KV System](https://github.com/tock/tock/blob/master/capsules/src/test/kv_system.rs) is a good example of implementing a test as a capsule.

Then you can `static_init!()` it inside a Rust test case like this: https://github.com/tock/tock/blob/master/boards/opentitan/src/tests/tickv_test.rs#L32 . Note that you don't have to do it in a `#[test_case]`, you just can

and you use a global in `main.rs` to pass it to the test case, like this: https://github.com/tock/tock/blob/master/boards/opentitan/src/tests/tickv_test.rs#L32

I'm not suggesting you do anything related to QEMU. Just that you allow the use of the generic Rust test infrastructure, that way we can call the tests from a `cargo test` command instead of having to comment/uncomment the tests out. In the case of the IMIX you can still just comment it out, but it then allows other boards to add it as a test case. Which then means it can be run as part of the CI :)

phil-levis: It seems like this should be documented if it's a requested change for a testing approach?

phil-levis: > Ah, ok. We don't have documentation on how to write one, just examples.
> 
> The [KV System](https://github.com/tock/tock/blob/master/capsules/src/test/kv_system.rs) is a good example of implementing a test as a capsule.

This test seems strange to me -- the capsule test documentation says to use `tickv_test::run_tickv_tests(kvstore)`, but this is an OpenTitan-specific test. It seems like there's a completely separate testing ecosystem; it transitions tests to being `static mut` variables within main.rs, something which the rest of Tock has tried hard to remove. Put another way, you've chosen a different set of tradeoffs in testing than the rest of Tock has; I'm a little way of starting to transition the rest of Tock over to it without discussion with in the core group. So for now, I'd like to keep the testing as is, consistent with the rest of imix.

alistair23: There seems to be some confusion about QEMU or OpenTitan specifics. The `test_case` and `cargo test` infrastructure is a core part of Rust. It isn't specific to QEMU or OpenTitan. For example libtock-rs is also [using it](https://github.com/tock/libtock-rs/blob/master/doc/Testing.md), although in a slightly different way.

`tickv_test::run_tickv_tests()` will work on any board that has flash and TicKV setup. OpenTitan is just the board that it was first implemented on. OpenTitan also isn't the only board using the Rust `test_case` and `cargo test` infrastructure.

Besides which, that is unrelated. You don't need to write a `test_case` and use `cargo test`, you can just comment out the test case if you want. The main ask was just to make the test case generic and not require any arguments. Then we can run the test case on a regularly tested board to ensure there are no regressions in the SHA implementation. This is what we do with the [alarm tests for example](https://github.com/tock/tock/blob/master/boards/opentitan/src/tests/multi_alarm.rs#L28)

Then we can call the test from the OpenTitan test builds. We don't need a software implementation for OT, as it has a hardware version. By running it in the tests we can run the test case on every single PR!! That way we avoid it bit rotting, which is what happens when we comment out the test case and run it when someone thinks about it.

phil-levis: > There seems to be some confusion about QEMU or OpenTitan specifics. The `test_case` and `cargo test` infrastructure is a core part of Rust. It isn't specific to QEMU or OpenTitan. For example libtock-rs is also [using it](https://github.com/tock/libtock-rs/blob/master/doc/Testing.md), although in a slightly different way.

Yeah, I am aware that this is a standard Rust mechanism.

> `tickv_test::run_tickv_tests()` will work on any board that has flash and TicKV setup. OpenTitan is just the board that it was first implemented on. OpenTitan also isn't the only board using the Rust `test_case` and `cargo test` infrastructure.

Then why is it in `boards/opentitan`? Note line 4:

```rust
use crate::{SIPHASH, TICKV};
```

This requires there be a `static mut`, as in `boards/opentitan/src/main.rs`:

```rust
static mut SIPHASH: Option<&capsules::sip_hash::SipHasher24<'static>> = None;
```

If I follow the approach you've taken to implementing an argument-free test, it requires adding `static mut` fields to `main.rs`. Since other platforms don't do this, it means the test is only available for OT. Is this what you mean? So other boards don't use these global Options, but since OT does, then it's possible to run a software regression?

On one hand, I found that these tests are not particularly useful because they don't use hardware; they can only test pure software. On the other this SHA256 implementation is pure software, so it seems like it's actually a good fit for CI.

> Then we can call the test from the OpenTitan test builds. We don't need a software implementation for OT, as it has a hardware version. By running it in the tests we can run the test case on every single PR!! That way we avoid it bit rotting, which is what happens when we comment out the test case and run it when someone thinks about it.

Given my struggle in getting the current OT system up and running, I think "we" is a bit of a stretch here. :) But it seems like there are two separate points here, one is CI checks, the other is running tests on hardware?

alistair23: It seems like the `static mut` is causing confusion. I'll try and explain it.

So the [Rust tests](https://doc.rust-lang.org/book/ch11-01-writing-tests.html) don't take arguments. Which means we somehow need to pass the already created drivers or capsules to the test case. One option would be to create the driver or capsule in the test case, but this isn't great as it deviates from the non-testing setup and results in duplication.

So the solution we use in OpenTitan and other Tock boards is to use a `static mut` to pass capsules and drivers to test cases. This is the same thing all Tock boards do with `CHIP`, `PROCESSES` and `PROCESS_PRINTER` for the panic IO. It's not great, but it's not that awful.

If you aren't using Rust's testing infrastructure, which the IMIX isn't, then you don't really need to worry about this. You could copy the entire `test_case` into `main()` and then just call it directly. No `static mut` required!

What this means is that we now has a generic test case, that can be called via commented out code or the Rust test infrastructure. It's much more generic!

To answer your other questions:

> Then why is it in `boards/opentitan`? Note line 4:

We use it in OpenTitan, but it's not OpenTitan specific

> This requires there be a `static mut`, as in `boards/opentitan/src/main.rs`:

Yes, it does if you want to pass it to a `test_case`, if you want to call it in `main()` you don't need a `static mut`

> it requires adding `static mut` fields to `main.rs`

Only if you want to call it from outside `main()`. If you want to call the test directly from `main()` you don't need to do this

> Since other platforms don't do this, it means the test is only available for OT. Is this what you mean? So other boards don't use these global Options, but since OT does, then it's possible to run a software regression?

I'm not clear what you mean. The software regression can be called directly from `main()` on any board (obviously we need flash though). As well as that other in tree boards use the Rust test infrastructure already. You are correct that only OT does this for the KV tests though.

> On one hand, I found that these tests are not particularly useful in CI because they don't use hardware; they can only test pure software

Agreed, hopefully soon-ish we can run Tock on Verilator in the CI! Which means we are testing hardware as well!

> On the other this SHA256 implementation is pure software, so it seems like it's actually a good fit for CI.

Exactly!

> Given my struggle in getting the current OT system up and running, I think "we" is a bit of a stretch here. :) But it seems like there are two separate points here, one is CI checks, the other is running tests on hardware?

That is strange. There are a few Tock people using OT on Tock, so I'm surprised your are having so many issues. Maybe file a bug case in Tock for the specific problems and we can try and update the docs to address them?

For this PR I don't think we need to worry about hardware. The main goal should just be allowing the test to be used with the Rust test infrastructure. As an extra bonus actually running the test on OT would be awesome!

phil-levis: I understand how Rust tests work. But I think some of your claims don't make sense. The test isn't OpenTitan specific, but it's in the OpenTitan boards directory. The core group agreed to remove `static mut` globals from main.rs a long while ago, and this new testing approach re-introduces them. Unless of course you call things from main, but this is not what the tests you've written (or are suggesting I copy) do.

What are the other boards that are using these tests by declaring `static mut` globals in main.rs?

phil-levis: I incorporated a test into OpenTitan that follows the same approach used there.

phil-levis:
Yes -- Alistair cleared up the issues with Verilator in the OpenTitan WG call today and now the OT HMAC peripheral is working.

phil-levis:
@alistair23 Can you re-review and clear your changes requested?

PR #3011, opened 2022-04-04
RFC: Display HIL

dcz-self:
### Pull Request Overview

This pull request changes the Screen HIL into something hopefully more universal. Differences from Screen:

1. When implementing the Screen HIL on a monochrome microcontroller, I had lots of troubles regarding bit packing of pixels due to this being underspecified. This is addressed here: write area's rows must be aligned to bytes. I added some clarifications for pixel formats about bit ordering, and added a wildcard format for cases where submitting the format upstream might be troublesome (my display has a 3-bit format, go figure).

2. Another problem is that the default Screen is a bit bloated. There's no reason to implement rotation or inversion on a dumb device. They may not even be supported, meaning software emulation, and that shouldn't really be a requirement to add a driver. So I split out the mnimal functionality to a new FrameBuffer trait, and put the extras in FixedFunctionScreen.

3. The distinction between brightness and power was not preserved. That was fixed, so now I can unambiguously run the display without backlight, or turn it into a light source without displaying anything. Moreover, setting power is now the right way to trigger the `screen_is_ready` callback.

4. I'm not entirely sure about that, but I also removed the ability to resum4e writes. It was used by the syscall driver, but I'm not seeing much point in this extra complexity. Displays are usually fixed to a board, so the author already knows how to size the buffer appropriately. Worst case, write it row by row.

5. One thing I *did not* include is adjusting the write area. Some displays (monochrome) write 8×N pixels at a time. Some others write 1×WIDTH at a time. That calculation needs to be done by the userspace often, so some way of transporting the formula from the kernel would be needed (WASM?). I don't have a clean solution, even though it complicates my driver a lot.

6. EDIT: there's a FrameBuffer::flush operation now, which flushes writes. That should be useful for actual frame buffers which may update a pixel here and there but need to write the entire frame at once for some reason (even if the reason is APIs like in https://github.com/tock/tock/pull/3009 ).

I split this into a separate file, but I think this should rather replace the existing Screen.

### Testing Strategy

This pull request was tested by eye. Also, the FrameBuffer subset is here https://github.com/tock/tock/pull/3009

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/hil/display.rs
+++ b/kernel/src/hil/display.rs
@@ -1,0 +1,334 @@
+//! Interfaces for screens and displays.
+use crate::ErrorCode;
+use core::ops::Add;
+use core::ops::Sub;
+
+#[derive(Copy, Clone, PartialEq)]
+pub enum ScreenRotation {
+    Normal,
+    Rotated90,
+    Rotated180,
+    Rotated270,
+}
+
+impl Add for ScreenRotation {
+    type Output = Self;
+
+    fn add(self, other: Self) -> Self {
+        match (self, other) {
+            (ScreenRotation::Normal, _) => other,
+            (_, ScreenRotation::Normal) => self,
+            (ScreenRotation::Rotated90, ScreenRotation::Rotated90) => ScreenRotation::Rotated180,
+            (ScreenRotation::Rotated90, ScreenRotation::Rotated180) => ScreenRotation::Rotated270,
+            (ScreenRotation::Rotated90, ScreenRotation::Rotated270) => ScreenRotation::Normal,
+
+            (ScreenRotation::Rotated180, ScreenRotation::Rotated90) => ScreenRotation::Rotated270,
+            (ScreenRotation::Rotated180, ScreenRotation::Rotated180) => ScreenRotation::Normal,
+            (ScreenRotation::Rotated180, ScreenRotation::Rotated270) => ScreenRotation::Rotated90,
+
+            (ScreenRotation::Rotated270, ScreenRotation::Rotated90) => ScreenRotation::Normal,
+            (ScreenRotation::Rotated270, ScreenRotation::Rotated180) => ScreenRotation::Rotated90,
+            (ScreenRotation::Rotated270, ScreenRotation::Rotated270) => ScreenRotation::Rotated180,
+        }
+    }
+}
+
+impl Sub for ScreenRotation {
+    type Output = Self;
+
+    fn sub(self, other: Self) -> Self {
+        match (self, other) {
+            (_, ScreenRotation::Normal) => self,
+
+            (ScreenRotation::Normal, ScreenRotation::Rotated90) => ScreenRotation::Rotated270,
+            (ScreenRotation::Normal, ScreenRotation::Rotated180) => ScreenRotation::Rotated180,
+            (ScreenRotation::Normal, ScreenRotation::Rotated270) => ScreenRotation::Rotated90,
+
+            (ScreenRotation::Rotated90, ScreenRotation::Rotated90) => ScreenRotation::Normal,
+            (ScreenRotation::Rotated90, ScreenRotation::Rotated180) => ScreenRotation::Rotated270,
+            (ScreenRotation::Rotated90, ScreenRotation::Rotated270) => ScreenRotation::Rotated180,
+
+            (ScreenRotation::Rotated180, ScreenRotation::Rotated90) => ScreenRotation::Rotated90,
+            (ScreenRotation::Rotated180, ScreenRotation::Rotated180) => ScreenRotation::Normal,
+            (ScreenRotation::Rotated180, ScreenRotation::Rotated270) => ScreenRotation::Rotated270,
+
+            (ScreenRotation::Rotated270, ScreenRotation::Rotated90) => ScreenRotation::Rotated180,
+            (ScreenRotation::Rotated270, ScreenRotation::Rotated180) => ScreenRotation::Rotated90,
+            (ScreenRotation::Rotated270, ScreenRotation::Rotated270) => ScreenRotation::Normal,
+        }
+    }
+}
+
+/// Pixel format and color depth information.
+///
+/// In formats where pixels don't fall on byte boundaries,
+/// most significant bits encode pixels more to the left.
+#[derive(Copy, Clone, PartialEq)]
+#[repr(usize)]
+#[allow(non_camel_case_types)]
+pub enum ScreenPixelFormat {
+    /// Pixels encoded as 1-bit, used for monochromatic displays.
+    /// Leftmost pixels occupies the higher bits.
+    Mono,
+    /// Each pixel is a 4-byte half-word. Each color channel is 1 bit: RGBX,
+    /// where X is padding.
+    /// Leftmost pixel occupies the higher word.
+    RGB_4,
+    /// Pixels encoded as 2-bit red channel, 3-bit green channel, 3-bit blue channel.
+    RGB_233,
+    /// Pixels encoded as 5-bit red channel, 6-bit green channel, 5-bit blue channel.
+    RGB_565,
+    /// Pixels encoded as 8-bit red channel, 8-bit green channel, 8-bit blue channel.
+    RGB_888,
+    /// Pixels encoded as 8-bit alpha channel, 8-bit red channel, 8-bit green channel, 8-bit blue channel.
+    ARGB_8888,
+    /// For cases that were not taken into account. Specifies size in bits.
+    Other(u8),
+}
+
+impl ScreenPixelFormat {
+    const CHOICES_MASK: u32 = 0xffffff;
+
+    pub fn get_bits_per_pixel(&self) -> usize {
+        match self {
+            Self::Mono => 1,
+            Self::RGB_4 => 4,
+            Self::RGB_233 => 8,
+            Self::RGB_565 => 16,
+            Self::RGB_888 => 24,
+            Self::ARGB_8888 => 32,
+            Self::Other(size) => *size as usize,
+        }
+    }
+
+    /// Helper for encoding.
+    pub fn pack(&self) -> u32 {
+        match self {
+            ScreenPixelFormat::Mono => 0,
+            ScreenPixelFormat::RGB_233 => 1,
+            ScreenPixelFormat::RGB_565 => 2,
+            ScreenPixelFormat::RGB_888 => 3,
+            ScreenPixelFormat::ARGB_8888 => 4,
+            ScreenPixelFormat::RGB_4 => 5,
+            ScreenPixelFormat::Other(depth) => Self::CHOICES_MASK | ((*depth as u32) << 24),
+        }
+    }
+
+    /// Helper for decoding. Returns None for unsupported formats.
+    pub fn unpack(val: u32) -> Option<Self> {
+        match val & Self::CHOICES_MASK {
+            0 => Some(ScreenPixelFormat::Mono),
+            1 => Some(ScreenPixelFormat::RGB_233),
+            2 => Some(ScreenPixelFormat::RGB_565),
+            3 => Some(ScreenPixelFormat::RGB_888),
+            4 => Some(ScreenPixelFormat::ARGB_8888),
+            5 => Some(ScreenPixelFormat::RGB_4),
+            Self::CHOICES_MASK => Some(ScreenPixelFormat::Other((val >> 24) as u8)),
+            _ => None,
+        }
+    }
+}
+
+/// Describes a rectangular area in the frame buffer. Sized in pixels.
+pub struct Area {
+    /// First column
+    pub x: usize,
+    /// First row
+    pub y: usize,
+    /// Column count
+    pub width: usize,
+    /// Row count
+    pub height: usize,
+}
+
+/// A frame buffer.
+///
+/// The frame buffer may be stored on the device itself and inaccessible.
+pub trait FrameBuffer {
+    /// Returns a tuple (width, height) with the current resolution (in pixels)
+    /// This function is synchronous as the driver should know this value without
+    /// requesting it from the screen.
+    ///
+    /// note that width and height may change due to rotation
+    fn get_resolution(&self) -> (usize, usize);
+
+    /// Returns the current pixel format
+    /// This function is synchronous as the driver should know this value without
+    /// requesting it from the screen.
+    fn get_pixel_format(&self) -> ScreenPixelFormat;
+
+    /// Sets the video memory area.
+    ///
+    /// This function has to be called before the first call to the write function.
+    ///
+    /// Area's validity depends on the current pixel format:
+    /// - the first pixel of the first column must start on a byte boundary, and
+    /// - the last pixel of the last column must start on a byte boundary.
+    /// This is to reduce ambiguity regarding partial bytes.
+    ///
+    /// This will generate a `command_complete()` callback when finished.
+    ///
+    /// Return values:
+    /// - `Ok(())`: The write area is valid.
+    /// - `INVAL`: The parameters of the write area are not valid.
+    /// - `BUSY`: Unable to set the write area on the device.
+    fn set_write_area(&self, area: &Area) -> Result<(), ErrorCode>;
+
+    /// Sends a write command to write data in the selected video memory area.
+    /// When finished, the driver will call the `write_complete()` callback.
+    ///
+    /// Return values:
+    /// - `Ok(())`: Write is valid and will be sent to the screen.
+    /// - `RESERVE`: No write area reserved.
+    /// - `INVAL`: Write is invalid or length is wrong.
+    /// - `BUSY`: Another write is in progress.
+    fn write(&self, buffer: &'static mut [u8], len: usize) -> Result<(), ErrorCode>;
+
+    /// Applies all unapplied write commands.
+    ///
+    /// When finished, the driver shall call `ScreenClient::write_complete`.
+    /// Returns `BUSY` if another operation is in progress,
+    /// or `ENOSUPPORT` if the driver automatically applies all writes.
+    fn flush(&self) -> Result<(), ErrorCode> {
+        Err(ErrorCode::NOSUPPORT)
+    }
+
+    /// Set the object to receive the asynchronous command callbacks.
+    fn set_client(&self, client: Option<&'static dyn ScreenClient>);
+}
+
+/// Describes a display module with physical properties.
+///
+/// Must be implemented on top of a `FrameBuffer`.
+pub trait ScreenModule {
+    /// Sets the display brightness on a logarithmic scale
+    ///
+    /// Displays should implement this function for at least 0 and 1.
+    /// - 0 - completely no light emitted
+    /// - otherwise - on, set brightness to value
+    fn set_brightness(&self, brightness: usize) -> Result<(), ErrorCode>;
+
+    /// Controls the screen power supply.
+    ///
+    /// Use it to initialize the display device.
+    ///
+    /// Does not control backlight power (if applicable),
+    /// so call `set_brightness` to turn on/off the module completely.
+    ///
+    /// When finished, calls `ScreenClient::screen_is_ready`,
+    /// both when power was enabled and disabled.
+    fn set_power(&self, enabled: bool) -> Result<(), ErrorCode>;
+}
+
+/// A screen which has some extra processing capabilities.
+///
+/// Only available to drivers implementing FrameBuffer.
+pub trait FixedFunctionScreen {
+    /// Sets the rotation of the display.
+    ///
+    /// Pixels already in the frame buffer are not affected,
+    /// but newly submitted pixels follow the new directions and dimensions.
+    ///
+    /// The call to `set_rotation` shall invalidate the selected write area.
+    ///
+    /// The function should return Ok(()) if the request is registered
+    /// and will be sent to the screen.
+    /// Upon Ok(()), the caller has to wait for the `command_complete` callback function
+    /// that will return the actual Result<(), ErrorCode> after setting the rotation.
+    ///
+    /// Note that in the case of `Rotated90` or `Rotated270`, this will swap the width and height.
+    /// Submitting pixels still respects the same byte boundary rules.
+    ///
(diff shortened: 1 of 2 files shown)


dcz-self:
> I don't quite understand how you plan on using pack and unpack.

It's just a helper to send the enum value to the userspace in a coordinated format.

> On several boards, the touch panel is rotated differently from the screen.

Thanks, that's a very good point. While I haven't yet played with displays that are rotation-aware, I can't argue with the touch and display HILs needing to be synced in terms of the coordinate systems. But I'd choose to transform touch events to the right coordinates (easy) rather than implementing all kinds of blitting for rotated buffers in the display. It's lots of work, and would totally lock performance to the absolute bottom on monochrome, row-based ones, for example.

That being said, libraries like lvgl are already able to rotate the outputs, for example: https://docs.lvgl.io/latest/en/html/porting/display.html#rotation

> The problem here is that you get a buffer from user space

But the screen syscall driver copies the data to its internal buffer before submission (sigh). That's because the driver may take the buffer and send them to DMA with the device directly. However, userspace buffers *cannot* be sent to DMA. So the buffer will not come from the userspace, and therefore can be sized right.

I think it would be better if userspace buffers could be DMAed, and then it would make sense to consider resumptions, but now they are unneeded.

> I assume this is why you added the flush function?
> 
> I think this was covered by the write_continue function.

No, those are unrelated. Continue is removed only because I don't find it useful. Flushing is added because otherwise I need to submit 3 frames (>10000 bytes) on the buffer if I update 3 pixels in different rows (because DMA requests can't take offsets, see https://github.com/tock/tock/pull/3009). With flush, I only send one frame (<4000 bytes). It's about batching.

>> One thing I _did not_ include is adjusting the write area
> I think this is the problem of the driver rather than the HIL.

Depends - if we cache the entire frame, then it's on the driver. If we want to send writes directly to the hardware, it's on HIL. Even LVGL consumes that information: https://docs.lvgl.io/latest/en/html/porting/display.html#_CPPv4N11_disp_drv_t10rounder_cbE

alexandruradovici:
> > I don't quite understand how you plan on using pack and unpack.
> 
> It's just a helper to send the enum value to the userspace in a coordinated format.

Got it. The general idea of the HIL is to be completely decoupled from the implementation. While the HIL defines variants for enums, it should not have attach a value to it. This is rather an implementation detail. The display syscall driver, the one that uses the HIL and has to exchange information with its user space counterpart should be the one transforming the variants into numbers. This allows several drivers to implement different data exchange formats.
> 
> > On several boards, the touch panel is rotated differently from the screen.
> 
> Thanks, that's a very good point. While I haven't yet played with displays that are rotation-aware, I can't argue with the touch and display HILs needing to be synced in terms of the coordinate systems. But I'd choose to transform touch events to the right coordinates (easy) rather than implementing all kinds of blitting for rotated buffers in the display. It's lots of work, and would totally lock performance to the absolute bottom on monochrome, row-based ones, for example.
> 
The touch driver uses the rotation information provided by the display to adjust its calculation.

> That being said, libraries like lvgl are already able to rotate the outputs, for example: https://docs.lvgl.io/latest/en/html/porting/display.html#rotation
> 
Yes, but that is done in software. We want to have a correct setup for the hardware so that applications do not have to worry about the difference in rotation between the display and the touch panel.

> > The problem here is that you get a buffer from user space
> 
> But the screen syscall driver copies the data to its internal buffer before submission (sigh). That's because the driver may take the buffer and send them to DMA with the device directly. However, userspace buffers _cannot_ be sent to DMA. So the buffer will not come from the userspace, and therefore can be sized right.
> 
It has to do that as it cannot pass directly a shared slice to the driver. However it can allocate small buffers and send parts to the display driver's buffer by using `write` and `write_continue`. The screen syscall driver that uses the screen HIL should not make any assumptions about the display driver underneath it. The screen syscall driver has to be generic and be able to use any actual display driver that implements the screen HIL.

> I think it would be better if userspace buffers could be DMAed, and then it would make sense to consider resumptions, but now they are unneeded.
> 
> > I assume this is why you added the flush function?
> > I think this was covered by the write_continue function.
> 
> No, those are unrelated. Continue is removed only because I don't find it useful. Flushing is added because otherwise I need to submit 3 frames (>10000 bytes) on the buffer if I update 3 pixels in different rows (because DMA requests can't take offsets, see #3009). With flush, I only send one frame (<4000 bytes). It's about batching.
> 
I'm not sure I understand how DMA is affected by flush.

> > > One thing I _did not_ include is adjusting the write area
> > > I think this is the problem of the driver rather than the HIL.
> 
> Depends - if we cache the entire frame, then it's on the driver. If we want to send writes directly to the hardware, it's on HIL. Even LVGL consumes that information: https://docs.lvgl.io/latest/en/html/porting/display.html#_CPPv4N11_disp_drv_t10rounder_cbE

I'm not sure I am following here. The HIL should always only expose an interface rather than implementation. The display driver knows its display and what it can do, so it should be the one that provides such implementations.

dcz-self:
BTW, the screen needs improvements in several places, also including syscalls, see https://github.com/tock/tock/issues/3035

But that's not the end, so to prevent a pileup I'll try to submit only a handful of improvements at a time (even if it's tempting to unload all of them).

With several improvements applied, we get a nice userspace driver which can do that:
syscall

Sources (uncleaned): https://github.com/dcz-self/tock/commit/10d39abab588d06a29d761cd6c3d9604038f56f9 and https://github.com/dcz-self/libtock-rs/commit/2ee8ba22507789c462a973cd277b060920e48804

bradjc:
@alexandruradovici

kernel/src/hil/display.rs
```diff
@@ -0,0 +1,334 @@
...
+    /// - `INVAL`: Write is invalid or length is wrong.
+    /// - `BUSY`: Another write is in progress.
+    fn write(&self, buffer: &'static mut [u8], len: usize) -> Result<(), ErrorCode>;
+
+    /// Applies all unapplied write commands.
+    ///
+    /// When finished, the driver shall call `ScreenClient::write_complete`.
+    /// Returns `BUSY` if another operation is in progress,
+    /// or `ENOSUPPORT` if the driver automatically applies all writes.
+    fn flush(&self) -> Result<(), ErrorCode> {
```
alexandruradovici, line 192: I now remember the use case of `write_continue` and how I used it. 

The small screens that I used would reset the write window (write from its beginning) when a new SPI write command was issues. As screen data may be large (I have a 240 x 240 16 bit screen), the `ProcessBuffer` sent by an application may not fit into the driver's buffer. This is why I needed to distinguish between an actual write (that would reset the screen's write position) and continued write.

This mechanism could be implemented by using `write` and `flush`, but a driver might forget to use `flush` at the end of writes, which would prevent the screen from reseting the write position. I am leaning towards keeping al the three functions, but others might want to chip in. It is true that having only `write` and `flush` makes the API simpler, but does complicate the implementation and adds a little room for error.

alexandruradovici (changes requested):

alexandruradovici:
> It's just the choice of writing 3 times to the display versus writing 1 time. For why I'm forced to submit entire frames on every tiny update, check out the WIP driver I linked, and I think that's a better place for that discussion.
> 
I am assuming your screen has an internal memory tht you write to and that needs to be *transferred* to the display upon using `flush`?

> Both options expose an interface. I'm posing the question of: is it worth making the interface more complicated (the rounding method) but more performant on some hardware? Or is it OK to have only a simpler interface that must have an extra cache (so space and performance hit) on that hardware? In this HIL I went for the second, but also I didn't make any performance measurements (which I expect to be abysmal due to a confluence of factors).

I think that the actual driver of the screen can perform these calculations (how to write), as it knows its screen and knows how it requires its data. I think this should be rather fast. Can you provide an actual example here?

kernel/src/hil/display.rs
dcz-self: I think `flush` is not really an integral part of this API, so I'm willing to leave it out for later.

I don't think that `continue` is needed for your use case: if the device expects an *entire* frame at once, then the driver can size its buffer accordingly. A write request bigger than that will be more than one frame, and is an indication that something is wrong.

However, `continue` might be useful for situations where the device expects an entire buffer inside a single write transaction, but there's not enough RAM to hold it all.

Given that there's no way to submit such buffers today (SPI doesn't use LeasedBuffer), I'd rather drop `continue` and worry about this use case later.

alexandruradovici: > I don't think that `continue` is needed for your use case: if the device expects an _entire_ frame at once, then the driver can size its buffer accordingly. A write request bigger than that will be more than one frame, and is an indication that something is wrong.
> 

I think I have to disagree on this. The problem is the discrepancy between the application buffer size and the syscall driver's buffer size. If the application buffer size is larger that the one of the syscall driver (as the application has no way of knowing the syscall driver's buffer size), the syscall driver has to issue *several writes* to the display driver that need to appear *as one* for the actual device.

Let's say we have a buffer of 100 bytes from the application and a buffer of 30 bytes available in the syscall driver. The actual device needs to receive all the 100 bytes in one single command. This is what happens:

`write (30 bytes)` -> sends WRITE_CMD BUFFER to the communication channel
`continue_write(30bytes)` sends BUFFER to the communication channel (not WRITE followed by BUFFER)
`continue_write(30bytes)` sends BUFFER to the communication channel (not WRITE followed by BUFFER)
`continue_write(10bytes)` sends BUFFER to the communication channel (not WRITE followed by BUFFER)

Where it not for `continue_write`, the syscall driver would send WRITE BUFFER every time, which would result in a reset of the write position. The only way to avoid this would be for the sycall driver to recalculate the write frame position, which seems a lot of overhead.

> However, `continue` might be useful for situations where the device expects an entire buffer inside a single write transaction, but there's not enough RAM to hold it all.
> 

This is mostly why the syscall driver might have a smaller buffer then the application.

> Given that there's no way to submit such buffers today (SPI doesn't use LeasedBuffer), I'd rather drop `continue` and worry about this use case later.

I'm not sure what you mean by this.

dcz-self:
> I am assuming your screen has an internal memory tht you write to and that needs to be transferred to the display upon using flush?

No, I thought up flush to work around Tock's SPI capsule not accepting LeasedBuffers.

> I think that the actual driver of the screen can perform these calculations (how to write), as it knows its screen and knows how it requires its data. I think this should be rather fast. Can you provide an actual example here?

The calculation of the area size that needs to be updated can be done on the driver side. What cannot be done on the driver side is getting the data that covers this area.

Determining the minimum area to be updated is not the problem here. The problem is that the userspace needs to provide data to fill that area, and the area will depend on the hardware.

I think the easiest way to think about it is: all displays are tiled displays. They accept only M×N pixel writes. Some accept 1×1 - that's a case everyone trivially supports. This HIL assumes it. Some displays might support upddating the entire frame at once, and that's covered easily by having a frame buffer.

But some displays support e.g. 8×8 as the minimal tile size, and the tile must be at an offset of (8*k, 8*l). An application trying to update a single 1×1 pixel must get information back from the driver that it's expected to update the entire tile. That cannot be done without HIL cooperation. The only way to keep it inside the driver is to take a memory hit and add a frame buffer in the driver.

kernel/src/hil/display.rs
dcz-self: No, we're in complete agreement :) I'm saying that:

> a buffer of 30 bytes available in the syscall driver.

is a bug. The syscall driver can either:

- split the user's buffer into squares, each of which is <30 bytes long, and submit those or
- allocate a buffer big enough to keep the entire frame of the underlying device driver, and reject writes bigger than that.

That thing with LeasedBuffer is actually probably more of a consequence of another problem. I'll address this later, but for now, I think that `continue` forces additional complexity that shouldn't be present while figuring out the general shape of the HIL.

dcz-self:
> >    No, I thought up flush to work around Tock's SPI capsule not accepting LeasedBuffers.
>
> I'm not sure I fully understand this. Can you give an example?

Actually, that's LeasableBuffer.

I implemented my driver as a full frame buffer [u8; 4096] due to the problem of tiles not being supported. The display supports receiving data line by line, so I can just submit buf[A..B] to SPI. Except I can't, because SPI accepts &mut [u8], and doesn't have an "offset" parameter I could give it. LeasableBuffer does contain an offset field, but it is not supported by SPI.

That forces me to send the entire [u8; 4096] every time I write an area, because I can't submit just a part of the frame buffer.

If I could split the writing to frame buffer and the submission, I could batch 2 or more area writes.

This is all because of the tiles, so maybe it doesn't need to be tackled directly.

As for a solution to the frame buffer issue, I don't think reading frame buffer data is a good solution. In fact, I'm complaining about the need to have a frame buffer on the SoC at all. Rather, I want to communicate to the userspace: this is the minimal chunk you can submit. I think having a method/syscall to discover tile size (assuming it's rectangular and non-overlapping) might cover most displays ever made.

kernel/src/hil/display.rs
alexandruradovici: >     * split the user's buffer into squares, each of which is <30 bytes long, and submit those or
>
This is exactly what `continue`, it splits the buffer

>     * allocate a buffer big enough to keep the entire frame of the underlying device driver, and reject writes bigger than that.
> 
This is difficult to do, as applications have no idea (so far) about what that buffer is. My opinion is that these details should be completely abstracted from the application.

alexandruradovici:
> As for a solution to the frame buffer issue, I don't think reading frame buffer data is a good solution. In fact, I'm complaining about the need to have a frame buffer on the SoC at all. Rather, I want to communicate to the userspace: this is the minimal chunk you can submit. I think having a method/syscall to discover tile size (assuming it's rectangular and non-overlapping) might cover most displays ever made.
> 
> (I would also deal with this after more fundamental design is finished.)

Now I understand your problem. While I would still favor a solution that abstracts these details from the application, so that an application could send buffers of any size, I'm not sure that it is actually possible without the penalties that your are describing here.

One thing that I think we can rule out is the idea that the application would like to set only one pixel. If we take a look at lvgl, it always writes parts of the framebuffer, never single pixels.

kernel/src/hil/display.rs
dcz-self: Yes, but `continue` splits the buffer in a special, non-square way, whereas the `frame` API for submitting squares could be used instead, without adding more methods.

As for allocation, it's the system call driver that needs to allocate the big enough buffer, not the userspace. This would be hardcoded in the board file, and transparent to the userspace.

alexandruradovici: > Yes, but `continue` splits the buffer in a special, non-square way, whereas the `frame` API for submitting squares could be used instead, without adding more methods.
> 
Are you suggesting to have a function that reports to the syscall driver the size of the *tile* buffer? This needs to be somehow propagated to the user space then.

> As for allocation, it's the system call driver that needs to allocate the big enough buffer, not the userspace. This would be hardcoded in the board file, and transparent to the userspace.

I assume the syscall driver would have to allocate a buffer with a size multiple of a *tile*, not necessarily the whole framebuffer.

dcz-self:
Looking at embedded-graphics, they operate under the assumption that a pixel is the fundamental unit instead, and, having discussed with the maintainer, this is not easy to change.

I wrote a library that doesn't make that assumption (but does not explore the entire problem space either): https://framagit.org/jazda/core/-/blob/master/ray-graphics/src/lib.rs#L243

Short explanation: a TiledRenderer knows the size of the tile and can render whichever pixels it wants from the scene. It's not perfect, but it proves the concept.

kernel/src/hil/display.rs
dcz-self: The `continue` problem is mostly independent of the tile problem. I just suggest using the same function just smaller area for when a buffer needs to be split up, without proposing how exactly.

> I assume the syscall driver would have to allocate a buffer with a size multiple of a tile, not necessarily the whole framebuffer.

I think that's mostly right. In my case, the buffer would be sized to 1 line, and that's my minimal tile too.

kernel/src/hil/display.rs
```diff
@@ -0,0 +1,334 @@
...
+    ///
+    /// When finished, calls `ScreenClient::screen_is_ready`,
+    /// both when power was enabled and disabled.
+    fn set_power(&self, enabled: bool) -> Result<(), ErrorCode>;
+}
+
+/// A screen which has some extra processing capabilities.
+///
+/// Only available to drivers implementing FrameBuffer.
+pub trait FixedFunctionScreen {
```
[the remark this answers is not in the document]
dcz-self: Why this name? What I tried to convey here is that this has fancy functions that most displays are not expected to have.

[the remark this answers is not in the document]
alexandruradovici: `Display` - has only basic functions to actually display some content
`FrameBufferDisplay` - functions related to the framebuffer, like rotation, inversion. `FixedFunctionScreen`, at least for me, does not describe that. If you have any other names suggestions, I'm open to them.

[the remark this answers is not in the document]
dcz-self: In https://github.com/tock/tock/pull/3047 I called it "FancyFunction" :) The problem with "FrameBuffer" is that Display is also only framebuffer functionality, because this HIL is only for framebuffer-based displays. Having the functionality of writing data to pixels presumes that. (As opposed to oscilloscope-like displays or plotters or printers.)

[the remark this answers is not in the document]
alexandruradovici: You have a point, but we need a better name than `FixedFunction` or `Fancy`.

alexandruradovici:
> Looking at embedded-graphics, they operate under the assumption that a pixel is the fundamental unit instead, and, having discussed with the maintainer, this is not easy to change.
> 
This is a problem for displays that cannot write pixel by pixel. I see only two ways of using this:
- reading back the framebuffer from the display
- holding an in RAM (if possible) framebuffer

> I wrote a library that doesn't make that assumption (but does not explore the entire problem space either): https://framagit.org/jazda/core/-/blob/master/ray-graphics/src/lib.rs#L243
> 
> Short explanation: a TiledRenderer knows the size of the tile and can render whichever pixels it wants from the scene. It's not perfect, but it proves the concept.

lvgl uses kind of the same concept with its `render` function

I'm wondering how many displays have this *tile* problem. Is it specific to your display model or do we have many of these? Considering that `embedded-graphics` is specially created for embedded displays, is it safe to assume that most of them allow pixel grade writing?

dcz-self:
> I'm wondering how many displays have this tile problem. Is it specific to your display model or do we have many of these?

I have not seen a monochromatic display that would not have some tiling.

I don't think it's safe to assume anything about hardware based on how embedded-graphics is doing things.

alexandruradovici:
> > I'm wondering how many displays have this tile problem. Is it specific to your display model or do we have many of these?
> 
> I have not seen a monochromatic display that would not have some tiling.
> 
> I don't think it's safe to assume anything about hardware based on how embedded-graphics is doing things.

I agree on that. The problem that I see is that graphics libraries (such as lvgl) render parts of the buffer. Those parts might not be a multiple of the framebuffer tile. As long as the tile is of a size of 1, there is no problem. But if the tile size is n, libraries might loose pixels when rendering parts of the framebuffer. I wonder if this tile is specific to monochrome screens or not. On these screens, considering that we have only one bit per pixel, we could keep all the framebuffer in RAM.

dcz-self:
As for lvgl, it asks the driver to adjust the buffer size, so it avoids the problem.

If a library cannot deal with tiles, I'd be more comfortable with making it the library's problem rather than the kernel's. The library will need an adaptation/driver for Tock's syscalls anyway, and it can manage the buffer there.

alexandruradovici:
> As for lvgl, it asks the driver to adjust the buffer size, so it avoids the problem.
> 
I'm not sure it actually can do that. I might be mistaking, but lvgl's render function may send any length of pixels and a position, depending on what it needs to refresh. The documentation is this one https://docs.lvgl.io/latest/en/html/porting/display.html#display-driver.
> If a library cannot deal with tiles, I'd be more comfortable with making it the library's problem rather than the kernel's. The library will need an adaptation/driver for Tock's syscalls anyway, and it can manage the buffer there.

While I agree with you, I'm not sure how to do this. As lvgl refreshes only the part of the framebuffer that has changed, there is no way of of storing the previous values for *un rendered* pixels unless you store the whole buffer. 

Any ideas would be great.

alexandruradovici:
> > lvgl's render function may send any length of pixels and a position, depending on what it needs to refresh
> 
> Yes, but also depending on what the driver allows it to refresh: https://docs.lvgl.io/latest/en/html/porting/display.html#_CPPv4N11_disp_drv_t10rounder_cbE
> 
Now I think I understand the use of this function and yes, this should be the way to go. I think we should add an additional function that reports the tile size and a system call to report it to the application. Thank you for for being persistent with this.

kernel/src/hil/display.rs
alexandruradovici: This might add complexity to the display driver as it would have split the actual write frame into smaller parts and issue several SetFrame commands to the underlying driver. After writing, it should than reset the frame to the original frame that was requested by the application.

dcz-self: Right, there's going to be complexity in this either way. But when it's unavoidable, I'd rather have it complex in a uniform way :)

phil-levis (commented):
It seems like this PR is in good shape, there's just some disagreement on names. @dcz-self I agree with @alexandruradovici 's suggested name changes. E.g., we don't actually set the pixel format of a screen, we set it for a framebuffer; the display hardware then translates this into the actual pixel values of a screen (the hardware itself). 

As for `Fancy`, how about `RotatableInvertableDisplay`? Say what it does. :)

dcz-self:
Thanks, I'll take that part about naming into account.

This RFC has partially been already implemented (brightness/power), and there's an outstanding RFC for some other part (https://github.com/tock/tock/pull/3067), so I'm not going to update this one as a whole, but rather take smaller pieces from this to submit incremental changes separately.

This is because a more focused discussion on some of the mentioned changes allowed us to notice design problems not caught in this RFC.

phil-levis:
OK -- do you want to close this PR then?

phil-levis:
@dcz-self Should we close this PR?

dcz-self:
I have no opinion. It's still useful to me to take reference for some future work.

PR #3013, opened 2022-04-05
tock-register-interface: Provide `none` method for FieldValue type.

qwandor:
### Pull Request Overview

`Default` is useful to construct a `FieldValue` which can be passed to
`modify` as a no-op or used as an additive identity when combining other
`FieldValue`s.

`Eq` and `PartialEq` are useful for assertions in testing.

### Testing Strategy

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/libraries/tock-register-interface/src/fields.rs
+++ b/libraries/tock-register-interface/src/fields.rs
@@ -190,7 +190,7 @@ Field_impl_for!(usize);
 ///
 /// For the FieldValue, the masks and values are shifted into their actual
 /// location in the register.
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Default, Eq, PartialEq)]
 pub struct FieldValue<T: UIntLike, R: RegisterLongName> {
     mask: T,
     pub value: T,
--- a/libraries/tock-register-interface/src/fields.rs
+++ b/libraries/tock-register-interface/src/fields.rs
@@ -390,6 +390,7 @@ macro_rules! register_bitmasks {
 
             #[allow(dead_code)]
             #[allow(non_camel_case_types)]
+            #[derive(Copy, Clone, Eq, PartialEq)]
             #[repr($valtype)] // so that values larger than isize::MAX can be stored
             $(#[$outer])*
             pub enum Value {
--- a/libraries/tock-register-interface/src/fields.rs
+++ b/libraries/tock-register-interface/src/fields.rs
@@ -190,7 +190,7 @@ Field_impl_for!(usize);
 ///
 /// For the FieldValue, the masks and values are shifted into their actual
 /// location in the register.
-#[derive(Copy, Clone, Eq, PartialEq)]
+#[derive(Copy, Clone)]
 pub struct FieldValue<T: UIntLike, R: RegisterLongName> {
     mask: T,
     pub value: T,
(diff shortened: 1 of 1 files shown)


lschuermann:
I don't see the benefit of having a first-class additive identity for `FieldValue`s in mind. Do you have a particular usecase? I'm hesitant, as it might not be immediately clear that a default `FieldValue` is not just some zero-value, but in fact has a zero mask.

Also, I'm not convinced we should make `FieldValue` comparable. Given this implicitly compares both the mask and value, it combines two fundamentally distinct operations into a single, opaque and undocumented one. As a user, I could for instance think that the implementation would instead check that the values contained with two `FieldValue`s are identical, compensating for any shift. As we already have the enum abstraction, can we just use that for comparing a given field's value instead (i.e. add `#[derive(Eq)]`)? This has the field encoded in the type information instead.

qwandor:
There are several usecases I see for `Default`.

1. Setting all fields in a register to 0 in a type-safe way, like

```rust
registers.reg.write(REG::default());
```

(Yes, I realise you could just use `registers.reg.set(0)`, but I'd rather set expose the raw setters. In my actual code I have a wrapper around `write`.)

2. Accumulating a series of updates to make in one go, like

```rust
let mut field_values = REG::default();

if some_condition {
  field_values += REG::FIELD::SET;
}

if some_other_condition {
  field_values += REG::OTHER_FIELD::SET;
}

registers.reg.modify(field_values);
```

As for `Eq`, I mostly had in mind using it for testing, so I could use `assert_eq!` with values passed to a mock, but it's less important to me right now.

lschuermann:
Okay, I see the benefit of having a default implementation for `FieldValue`s now. I'm not super happy with the name still, as it's not immediately clear what "default" implies. Especially for memory-mapped registers, default is a common phrase to refer to the reset-values of a given register, which aren't necessarily zero. It seems entirely reasonable and super useful for some cases to introduce custom register types which have a proper const-generic default value attached to them, representing the reset value of this register, and perhaps also a mask of writeable vs. reserved bits. In that spirit, would it perhaps be better to call this associated function `none` or `empty`?

qwandor:
Implementing the `Default` trait has the advantage that it can then be derived on containing structs (i.e. if I make a struct that contains a `FieldValue`), and also seems more idiomatic to me, but a `none` or `empty` method would work for my immediate usecase. I think `none` is probably a better name than `empty` as it represents modifying none of the fields.

alistair23:
I agree, `Default` implies the reset value so `none` of `empty` seem like better names

hudson-ayers:
I might even recommend `zeroed()`, though I think `none` or `empty` are fine as well

qwandor:
Done. `zeroed` sounds to me like it would result in setting some fields to 0, which is not what this does, so I've gone with `none`.

alistair23 (dismissed):

lschuermann:
> As for `Eq`, I mostly had in mind using it for testing, so I could use `assert_eq!` with values passed to a mock, but it's less important to me right now.

Could you perhaps remove this change from this PR? I'm still not 100% convinced on that and it seems it wouldn't be a major inconvenience to you to just compare the mask and the read value (in conjuction with the corresponding `Field` type), if this matches your definition of equality?

qwandor:
Done.

lschuermann (approved):

bradjc (approved):

brghena (approved):

hudson-ayers (approved):

PR #3014, opened 2022-04-05
tock-registers: Implement From field enum value type for FieldValue.

### LINKED ISSUE #3012: tock-registers: Allow setting registers by enum value
If I use the `register_bitfields` macro to declare a set of possible value for a register field, then I can use `Readable::read_as_enum` to get back a value of the appropriate `REG::FIELD::Value` enum type. However, unless I'm missing something, there doesn't seem to be any way to write such an enum value back into the field (i.e. to pass it to `write` or `modify`).

This could be solved by generating a `From` implementation for `FieldValue` alongside the generated enum type.

qwandor:
### Pull Request Overview

This makes it easier to set a field based on the enum value.

Fixes #3012.

### Testing Strategy

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/libraries/tock-register-interface/src/fields.rs
+++ b/libraries/tock-register-interface/src/fields.rs
@@ -413,6 +413,12 @@ macro_rules! register_bitmasks {
                     }
                 }
             }
+
+            impl From<Value> for FieldValue<$valtype, $reg_desc> {
+                fn from(v: Value) -> Self {
+                    Self::new($crate::bitmask!($numbits), $offset, v as $valtype)
+                }
+            }
         }
     };
     {
--- a/libraries/tock-register-interface/README.md
+++ b/libraries/tock-register-interface/README.md
@@ -294,6 +294,9 @@ registers.cr.modify(Control::RANGE.val(2)); // Leaves EN, INT unchanged
 // Named constants can be used instead of the raw values:
 registers.cr.modify(Control::RANGE::VeryHigh);
 
+// Enum values can also be used:
+registers.cr.modify(Control::RANGE::Value::VeryHigh.into())
+
 // Another example of writing a field with a raw value:
 registers.cr.modify(Control::EN.val(0)); // Leaves RANGE, INT unchanged
 


libraries/tock-register-interface/src/fields.rs
```diff
@@ -413,6 +413,12 @@ macro_rules! register_bitmasks {
                     }
                 }
             }
+
+            impl From<Value> for FieldValue<$valtype, $reg_desc> {
+                fn from(v: Value) -> Self {
+                    Self::new($crate::bitmask!($numbits), $offset, v as $valtype)
```
lschuermann, line 419: This looks good to me as is, but can we perhaps instead just return the in-scope named constant here instead? That might make it more obvious that this is converting to the constant FieldValue and reduces code duplication, as these macro interactions are already sufficiently complicated. :)

lschuermann (approved):

libraries/tock-register-interface/src/fields.rs
qwandor: Do you mean adding a match expression with an arm for each enum value, returning the corresponding named constant? The seems like more code duplication, and makes for a more complicated macro than just doing the cast here.

lschuermann: Oh yeah, sorry, I've mistaken the types here. This seems to be as good a solution as we can get.

hudson-ayers (approved):

qwandor:
Now that this is approved, is there anything else I need to do to get it merged?

lschuermann:
> Now that this is approved, is there anything else I need to do to get it merged?

No, however we do like to wait a few days in case other people want to chime in. Technically [one review from someone with commit access is the minimum requirement to land a PR](https://github.com/tock/tock/blob/master/.github/CONTRIBUTING.md#step-8-landing), however, as a rule of thumb, the more complex and controversial a given PR is, and especially when it touches core aspects of a system, PRs are not merged immediately and are marked as `last-call` one or more days before merging. This change seems good and uncontroversial, though I'd like to give @ppannuto a chance to take a look still.

lschuermann:
bors try

, given the CI is red on the PR currently, but that looks to be a failure on Netlify's side?

brghena:
I don't understand what the purpose is here, although I'm suspecting that I just don't know Rust well enough.

We already had:
```
// Named constants can be used instead of the raw values:
registers.cr.modify(Control::RANGE::VeryHigh);
```

Now you have added:
```
// Enum values can also be used:
registers.cr.modify(Control::RANGE::Value::VeryHigh.into())
```

If you could already use `Control::RANGE:Value::VeryHigh` then why do you need to support `into()`? It's possible that "Enum value" has some specific definition that I'm not understanding here.

Do you have an example of where the second would be useful and the first doesn't suffice?

lschuermann:
> If you could already use `Control::RANGE:Value::VeryHigh` then why do you need to support `into()`? It's possible that "Enum value" has some specific definition that I'm not understanding here.

I think the point of this change is that, as you correctly stated, a `FieldValue` can be used in the [`modify`](https://docs.tockos.org/tock_registers/interfaces/trait.readwriteable#tymethod.modify) method of registers implementing `ReadWriteable`, but [`read_as_enum`](https://docs.tockos.org/tock_registers/interfaces/trait.readable#method.read_as_enum) on `Readable` returns an enum-type specific to the register field, where the discriminator is the value of this register.

This change is key to allow such a retrieved value, described by a typesafe enum, be converted into a `FieldValue` with the appropriate mask, which can then be used in a `modify` method to set this register again. While you can cast the register's enum-type value to it's underlying discriminator of primitive integer type `T`, and use [`Field::val`](https://docs.tockos.org/tock_registers/fields/struct.field#method.val) to cast the integer to a `FieldValue` with the proper bit-alignment of the register, which can then be used in [`modify`](https://docs.tockos.org/tock_registers/interfaces/trait.readwriteable#tymethod.modify), I agree with the sentiment of this pull request to be able to use the generated enums for every register-field to modify registers based on those extracted, type-safe values directly.

tl;dr: this is fundamentally about the round trip of (1) read as enum, (2) store as enum, (3) modify the register to write the old field value, all without converting to an intermediate primitive integer.

Does that make sense?

brghena:
This does make sense now (we talked about it a little offline). Notably I hadn't even noticed the `Control::RANGE::VeryHigh` vs `Control::RANGE::Value::VeryHigh` distinction. Thanks for clearing that up Leon.

brghena (approved):

ppannuto:
bors retry

PR #3023, opened 2022-04-17
screen HIL: Add separate set_power

dcz-self:
### Pull Request Overview

This pull request adds a separate set_power call to the Screen HIL.

The brightness value also got an explanation: it's lightness, not percent.

Since the screen HIL package seems to be stalled https://github.com/tock/tock/pull/3011, I decided to sneak in the changes piecewise instead.

### Testing Strategy

This pull request was not really tested. I don't have the hardware for the driver I modified here, and brightness wasn't used there either.

### TODO or Help Wanted

This pull request still needs testing/review.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/syscalls/90001_screen.md
+++ b/doc/syscalls/90001_screen.md
@@ -30,11 +30,26 @@ The screen driver allows the process to write data to a framebuffer of a screen.
 
     **Returns**: SUCCESS_U32 with 1 if yes, and 0 if no
 
+  * ### Command number: `2`
+
+    **Description**: Set power
+
+    **Argument 1**: 0 if off, nonzero if on.
+
+    **Argument 2**: unused
+
+    **Returns**: Ok(()) followed by the ready callback if the command was successful,
+    BUSY if another command is in progress,
+
   * ### Command number: `3`
 
     **Description**: Set brightness
 
-    **Argument 1**: Percent of brightness, 0% should turn off the screen, greater than 0% should turn it on.
+    **Argument 1**: Lightness value, relative to minimum and maximum supported.
+    0 should turn off the light if available, greater than 0 should set it to minimum.
+    65535 and above turn lightness to the maximum supported.
+    Intermediate values approximate intermediate lightness levels.
+    May take effect only after power is set (e.g. for LED displays).
 
     **Argument 2**: unused
 
--- a/kernel/src/hil/screen.rs
+++ b/kernel/src/hil/screen.rs
@@ -3,6 +3,8 @@ use crate::ErrorCode;
 use core::ops::Add;
 use core::ops::Sub;
 
+pub const MAX_BRIGHTNESS: usize = 65536;
+
 #[derive(Copy, Clone, PartialEq)]
 pub enum ScreenRotation {
     Normal,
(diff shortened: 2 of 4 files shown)


dcz-self:
Verified design (except powering off) in a branch https://github.com/dcz-self/tock/commit/5df63468ce752df843894ecb641f18fa4f54a799

alistair23 (dismissed):

alexandruradovici:
Looks good

doc/syscalls/90001_screen.md
```diff
@@ -30,11 +30,24 @@ The screen driver allows the process to write data to a framebuffer of a screen.
...
+
+    **Description**: Set power
+
+    **Argument 1**: 0 if off, nonzero if on.
+
+    **Argument 2**: unused
+
+    **Returns**: Ok(()) followed by the ready callback if the command was successful,
+    BUSY if another command is in progress,
+    INVAL if a brightness value is required but wasn't provided.
```
hudson-ayers, line 43: I might vouch for chips to select a default brightness value rather than present a requirement that set_brightness be called before set_power, especially since the language here suggests that requirement might only be present for some screens

doc/syscalls/90001_screen.md
```diff
@@ -30,11 +30,24 @@ The screen driver allows the process to write data to a framebuffer of a screen.
...
+
   * ### Command number: `3`
 
     **Description**: Set brightness
 
-    **Argument 1**: Percent of brightness, 0% should turn off the screen, greater than 0% should turn it on.
+    **Argument 1**: Lightness value, relative to minimum supported.
+    0 should turn off the light if available, greater than 0 should turn it on.
```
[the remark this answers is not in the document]
dcz-self: Yes, percentage is unclear. Brightness could mean either lightness or luminance. Luminance is unhelpful for humans because 50% is not much different from 100%. Lightness is better, but it's closer to a logarithmic relationship, and 0 logarthmiic is not 0 linear, making the linear scale unsuitable.

In addition, percent brings an unnecessary quantization to only 100 levels.

I'd rather leave figuring out the boundaries to the driver. Perhaps specify that the intended range stretches from 1 to u32::max?

doc/syscalls/90001_screen.md
dcz-self: Okay, this makes sense. I think it warrants some more changes in brightness: brightness changes are not guaranteed to take effect when the display is off.

Initially I intended the separation to make flashlight-like use cases possible, but a well-manufactured module will be a reasonable flashlight only when the pixels are set to white, and that needs the display to be on anyway.

doc/syscalls/90001_screen.md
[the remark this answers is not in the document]
hudson-ayers: Leaving the boundaries to the driver would make it pretty hard to write a screen-independent app that sets the screen to max brightness. That said, it may be unrealistic to imagine we will have screen-independent apps with this current interface anyway. Specifying the intended range to some range would be helpful though, I think.

doc/syscalls/90001_screen.md
hudson-ayers: I think that not guaranteeing brightness changes when the display is off seems reasonable

doc/syscalls/90001_screen.md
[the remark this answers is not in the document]
dcz-self: Done: I chose >65535 to be the maximum.

kernel/src/hil/screen.rs
```diff
@@ -207,12 +207,30 @@ pub trait Screen {
...
 
+    /// Controls the screen power supply.
+    ///
+    /// Use it to initialize the display device.
+    ///
+    /// Does not control backlight power (if applicable),
+    /// so call `set_brightness` to turn on/off the module completely.
+    ///
+    /// For displays where power needs nonzero brightness,
+    /// this may return `INVAL` if `set_brightness` was not called first.
```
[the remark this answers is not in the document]
dcz-self: Changed to setting a default.

doc/syscalls/90001_screen.md
hudson-ayers: You updated the doc comments in `screen.rs`, but this still suggests that a brightness value may be required, can that line be removed?

dcz-self: Thanks for catching that. Updated.

hudson-ayers (approved):

alistair23 (approved):

dcz-self:
Is there anything I can do to make reviewing this easier? The HIL and syscall driver need a whole array of changes, and I don't want to end up opening multiple parallel reviews and then ending up with conflicts against my own code.

dcz-self:
I think the closest to where I want to end up is already here: https://github.com/tock/tock/pull/3011

But I'll alter this one to include other less controversial changes if that helps.

alistair23 (approved):

hudson-ayers:
bors r+

This is already approved and ready to go so I don't see a good reason to hold these changes up.
I think other changes can be accepted in other PRs, sorry if this was holding up your other contributions.

PR #3032, opened 2022-04-27
STM32f4xx: Properly support 4MHz SPI rate

LeonMatthesKDAB:
Previously, setting 4MHz as the SPI rate was only supported in the
rate setter, but not the getter.

### Pull Request Overview

This pull request adds support for 4MHz in the spi's set_rate and get_rate functions.
Previously, it was only supported in the set_rate function, but also wasn't documented!

### Testing Strategy

**I am currently not able to test this, I just stumbled across it whilst reading the SPI code.**

### Documentation Updated

- No updates required

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/stm32f4xx/src/spi.rs
+++ b/chips/stm32f4xx/src/spi.rs
@@ -411,8 +411,8 @@ impl<'a> spi::SpiMaster for Spi<'a> {
         }
     }
 
-    /// We *only* support 1Mhz. If `rate` is set to any value other than
-    /// `1_000_000`, then this function panics
+    /// We *only* support 1Mhz and 4MHz. If `rate` is set to any value other than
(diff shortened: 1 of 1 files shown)


alistair23 (approved):

hudson-ayers (commented):
This looks fine but it would be nice if someone can test it, the comments in the original code make me slightly uneasy that this might not work for some reason.

alexandruradovici:
I'll try and test it on an STM32F412g Discovery and an SPI display.

bradjc (approved):
This looks fine to me. I don't see how this would make anything _worse_, at least.

hudson-ayers:
Fair enough.

bors r+

PR #3037, opened 2022-05-06
Clarify the definition of TBF's protected region

jrvanwhy:
### Pull Request Overview

This PR corrects and expands upon the definition of "protected region" in the TBF documentation.

[Rendered](https://github.com/jrvanwhy/tock/blob/protected-region-docs/doc/TockBinaryFormat.md)

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/doc/TockBinaryFormat.md
+++ b/doc/TockBinaryFormat.md
@@ -272,10 +272,9 @@ The `Main` element has three 32-bit fields:
   * `init_offset` the offset in bytes from the beginning of binary payload
     (i.e. the actual application binary) that contains the first instruction to
     execute (typically the `_start` symbol).
-  * `protected_size` the size of the protected region in bytes. The protected
-    region begins at the start of the app image and ends just before the
-    compiled app binary. Processes may not write to the protected region. The
-    TBF headers should be contained within the protected region.
+  * `protected_size` the size of the protected region in bytes. Processes do not
+    have write access to the protected region. TBF headers are contained in the
+    protected region.
   * `minimum_ram_size` the minimum amount of memory, in bytes, the process
     needs.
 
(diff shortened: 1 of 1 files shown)


hudson-ayers (dismissed):

doc/TockBinaryFormat.md
```diff
@@ -270,8 +272,10 @@ The `Main` element has three 32-bit fields:
   * `init_offset` the offset in bytes from the beginning of binary payload
     (i.e. the actual application binary) that contains the first instruction to
     execute (typically the `_start` symbol).
-  * `protected_size` the amount of flash, in bytes, after the header, to
-    prevent the process from writing to.
+  * `protected_size` the size of the protected region in bytes. The protected
+    region begins at the start of the app image and ends just before the
+    compiled app binary. Processes may not write to the protected region. The
+    TBF headers should be contained within the protected region.
```
lschuermann, line 278: At the risk of being too pedantic, are _may_ and _should_ really appropriate here? This might indicate that this could be allowed on some platforms and given our strict requirements around memory protection for platforms we target, I can't come up with good usecases to violate these invariants. Perhaps _must_ and _must_ are more appropriate here?

phil-levis: Since this isn't a TRD, those words don't have precise meanings, unfortunately. However, I agree there is some ambiguity. How about

"Processes do not have write access to the protected region. TBF headers are contained in the protected region."

jrvanwhy: Reading it again, my original wording sounds like it is asking processes to not write the protected region, when I intended to say they would be prevented from writing the protected region. I think Phil's wording is much clearer, so I switched to his wording.

lschuermann: Right, I think @phil-levis' version is great. Thanks for changing it!

lschuermann: Right, I think @phil-levis' version is great. Thanks for changing it!

lschuermann (approved):

lschuermann (approved):

PR #3038, opened 2022-05-07
kernel/dynamic_deferred_call: document multi-handle client behavior

### LINKED ISSUE #3024: DynamicDeferredCall multiple register()
Looking at DynamicDeferredCall, I am getting a little uncertain about getting Handles. The DynamicDeferredCallClient has `fn call(&self, handle: DeferredCallHandle);`, which suggests that it can dstinguish between multiple handles (cool), but then the handles come from `register(Client)`. I'm perhaps wrong about this, but `register(Client)` gives me vibes of idempotency: registering to a client multiple times wouldn't make a difference if it was SPI or something of the sort.

But in this case, registering multiple times actually makes sense: having two deferred call handles, each for a different purpose, is useful. A skim of the code doesn't show idempotency here either.

Perhaps this should be explicitly documented: "you may register many times, and you shall receive a new handle each time". Or, to minimize surprise, rename to `create_handle_for(client)`.

Or am I mistaken somewhere?

lschuermann:
### Pull Request Overview

This documents that it is possible for a client reference to be registered with the dynamic deferred call mechanism multiple times, retrieving a different handle each time. It furthermore derives `Eq` and `PartialEq` for the `DeferredCallHandle`, for clients to distinguish scheduled deferred calls.

Fixes #3024.

Suggested-by: @dcz <gihu.dcz@porcupinefactory.org>

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/dynamic_deferred_call.rs
+++ b/kernel/src/dynamic_deferred_call.rs
@@ -186,6 +186,14 @@ impl DynamicDeferredCall {
     ///
     /// On success, a `Some(handle)` will be returned. This handle is later
     /// required to schedule a deferred call.
+    ///
+    /// A given [`DynamicDeferredCallClient`] reference (client) can be
+    /// registered multiple times and will receive a different handle each
+    /// time. This mechanism is useful to distinguish between deferred calls
+    /// scheduled by the same client, but to be handled differently. Each issued
+    /// handle will occupy one [`DynamicDeferredCallClientState`] in the
+    /// [`DynamicDeferredCall`]. Clients can utilize the passed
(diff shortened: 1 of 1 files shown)


dcz-self:
Awesome, this makes deferred calls much more useful and less ambiguous.

dcz-self (approved):

brghena (approved):

hudson-ayers (approved):

alistair23 (approved):

Issue #2637, opened 2021-06-29
Violation of Rust's aliasing rules with DMA operations


lschuermann:
It's a common paradigm for chips in Tock to take some buffer which is written to using a DMA operation. Because this buffer must be handed back to the client when the operation completes, we need to hold on to a reference to it. This seems to be done almost exclusively through `TakeCell`s. For example, for the nRF52 family of chips:

https://github.com/tock/tock/blob/94172ecf293e6df598192111b87db9dbbcaad178/chips/nrf52/src/uart.rs#L162-L169

https://github.com/tock/tock/blob/94172ecf293e6df598192111b87db9dbbcaad178/chips/nrf52/src/uart.rs#L496-L510

However, by putting these buffers into a `TakeCell`, we are still holding on to a mutable Rust reference to the respective slice:

https://github.com/tock/tock/blob/94172ecf293e6df598192111b87db9dbbcaad178/libraries/tock-cells/src/take_cell.rs#L14-L16

If the DMA operation now mutates these buffers, we've violated Rust's assumption of there only being a single mutable reference to some object, and that object not being mutated otherwise. This, as far as I understand it, makes the kernel unsound.

---

The solution appears to be quite simple: we need to drop the slice reference prior to starting a DMA operation which might mutate the memory in question, and only store the pointer (as a `*mut u8`) and length of the respective buffer. Rust does not impose the strict aliasing requirements on pointers. Once the DMA operation is finished, it should be safe to reconstruct a proper Rust slice.

Because a `core::slice::from_raw_parts_mut` looks really scary, and furthermore is easy to mess up, I propose introducing a `DMABuffer` (naming WIP) type, which does exactly this: take a static mutable reference to a slice of bytes in the constructor, decompose it into its pointer and length, and allow to reconstruct a slice again consuming the type. This would of course need to be applied across the entire code base.

---

What do others think? Have I missed something or is this really a soundness issue?

jrvanwhy:
I agree this is a soundness issue, and that the fix is to store the slice as a raw pointer and length.

ppannuto:
+1, I agree this is an issue. I might bikeshed to `DMACell`, as the cell metaphor is well understood, and I think we can explain it as a cell whose contents is (1) empty, (2) owned by underlying hardware, (3) present and owned by software. Plus Tock has a history of custom `Cell` types.

lschuermann:
Thanks for the feedback! Will create a PR to propose a new type to handle this and subsequently apply it to all chip crates.

[...]

PR #3041, opened 2022-05-11
Rework of Digest HIL 

phil-levis:
### Pull Request Overview

This pull request reworks the Digest HIL, introducing four major changes:
  1) It allows a Digest to operate both on mutable and immutable data.This is so that one can calculate digests on read-only data (e.g., process binaries), as is needed for checking cryptographic credentials of processes without copying them into RAM first. `Digest` now has two methods to add data to a digest, `add_data` and `add_mut_data`. The callback client correspondingly has two callbacks `add_data_done` and `add_mut_data_done`.
  2) To support 1) above, it changes `LeasableBuffer` to take a `&[u8]` and adds a `LeasableMutableBuffer` which takes a `&mut [u8]`.
  3) A call to `add_data` and `add_mut_data` is expected to add all of the data that is in the active slice of the corresponding `LeasableBufffer` or `LeasableMutableBuffer`. Accepting less is an error condition. Correspondingly, a successful call to one of these methods now returns `Ok(())`: the prior version returned `Ok(usize)` to return how many bytes would be added.
  4) It changes the calls and callbacks of `Digest` to use only `LeasableBuffer` and `LeasableMutableBuffer` rather than using both these types as well as slices. This allows an error condition to indicate how much data is remaining to be added.

### Testing Strategy

This pull request has not been tested yet.

### TODO or Help Wanted

This needs to be tested on OpenTitan.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


phil-levis:
There was a lot of discussion on the core call today; there's general agreement that having two traits is undesirable, but there is not yet a better solution. So we agreed to discuss possible better approaches here on this PR before moving forward.

hudson-ayers:
I spent some time today thinking about a different design for asynchronously passing buffers to hardware that I think resolves some of the issues here. This design is partially inspired by the DMA design described here: https://docs.rust-embedded.org/embedonomicon/dma.html , but ours has the added challenge that we cannot rely on an assumption of futures / generators to simplify waiting on results, and instead have to support an event-driven state machine.

Step 1: Peripheral functions that currently take `&'static mut [u8]` despite not needing to actually mutate the underlying buffer are changed to be generic (just the function, not the entire peripheral) and instead take `B: AsRef<[u8]> + 'static`

```rust
//! kernel/src/hil/uart.rs
trait Transmit {
    fn transmit_buffer<B: AsRef<[u8]> + 'static>(
        &self,
        tx_buffer: B, // used to be &'static mut [u8]
        tx_len: usize,
    ) -> Result<BufferHandle<B>, (ErrorCode, B)>; // Notice different return type!
}
```

This is supported by a new type, `BufferHandle`, which knows the exact type of buffer that was passed to the underlying operation:

```rust
pub trait BufferHandleTrait {
    fn transfer_done(&self);
}

pub struct BufferHandle<B: AsRef<[u8]> + 'static> {
    buffer: B,
    finished: Cell<bool>,
}

impl<B: AsRef<[u8]> + 'static> BufferHandle {
    fn get_buffer(&self) -> Option<B> {
        if self.finished.get() {
            Some(B)
        } else {
            None
        }
    }
}

impl<B: AsRef<[u8]> + 'static> BufferHandleTrait for BufferHandle<B> {
    fn transfer_done(&self) {
        self.finished.set(true);
    }
}
```

The underlying code that actually passes the buffer to hardware (e.g. for a DMA operation) passes the address / length to the underlying hardware, then stashes the buffer (whatever its concrete type) in the buffer handle. The buffer handle ensures that the actual (Rust) buffer is not made available until its callback is called indicating the underlying hardware no longer will be accessing the buffer.

```rust
//! dma.rs
// sam4l transmit_buffer calls dma.prepare_transfer() to actually send the buffer, so this generic component is just forwarded
    pub fn prepare_transfer<B: AsRef<[u8]> + 'static>(
        &self,
        pid: DMAPeripheral,
        buf: B,
        mut len: usize,
    ) -> DMAHandle<B> {
        let maxlen = buf.as_ref().len()
            / match self.width.get() {
                DMAWidth::Width8Bit /*  DMA is acting on bytes     */ => 1,
                DMAWidth::Width16Bit /* DMA is acting on halfwords */ => 2,
                DMAWidth::Width32Bit /* DMA is acting on words     */ => 4,
            };
        len = cmp::min(len, maxlen);
        self.registers
            .mr
            .write(Mode::SIZE.val(self.width.get() as u32));

        self.registers.psr.set(pid);
        self.registers
            .marr
            .write(MemoryAddressReload::MARV.val(buf.as_ref()[0] as *const u8 as u32));
        self.registers
            .tcrr
            .write(TransferCounter::TCV.val(len as u32));

        self.registers.ier.write(Interrupt::TRC::SET);

        BufferHandle { buf, finished: Cell::new(false) }
    }
```

Now, we need a way to indicate to the BufferHandle that it can unlock once the transfer is done. Let's do this using the new `BufferHandleTrait` above, because we cannot allow our underlying peripheral to be generic over a parameter for which there will be multiple types used, so this access needs to occur via a vtable:

```rust
//! dma.rs
struct DMAChannel {
    // ... stuff
    // we used to store buf: &'static mut [u8] here, no more
    client: &'static dyn DMAClient, // same as before
    handle: &'static dyn BufferHandleTrait,
}

impl DMAChannel {
        self.registers
            .idr
            .write(Interrupt::TERR::SET + Interrupt::TRC::SET + Interrupt::RCZ::SET);
        let channel = self.registers.psr.get();

        self.client.map(|client| {
            client.transfer_done(channel); // normal client notification mechanism
        });
        self.handle.map(|handle| {
            handle.transfer_done(); // indicate to handle buffer can be safely retrieved
        });
    }
}
```

Finally, the capsule-level code that is calling the HIL is responsible for holding onto the handle, and retrieving its original buffer once the operation is complete:

```rust
//! console.rs (I haven't thought about virtualization yet, lets assume direct access)
struct Console<U: Uart + 'static> {
    uart: &'static U,
    rw_buffer: TakeCell<&'static mut [u8]>, // for printing from ram buf
    ro_buffer: TakeCell<&'static [u8]>, //for printing from flash buf
    rw_handle: OptionalCell<BufferHandle<&'static mut [u8]>>, 
    ro_handle: OptionalCell<BufferHandle<&'static [u8]>>,
}

fn command(cmd_num: u32) {
    match cmd_num {
        0 /* print from RAM buffer */ => {
            let rw_handle = self.uart.transmit_buffer(self.static_rw_buf.take());
            self.rw_handle.set(Some(rw_handle)); // store handle so we can retrieve buffer upon tx_complete
        }
        1 /* print from flash buffer */ => {
            let ro_handle = self.uart.transmit_buffer(self.static_ro_buf.take());
            self.ro_handle.set(Some(ro_handle)); // store handle so we can retrieve buffer upon tx_complete
        }
    }
}

impl UartClient for Console {
    transmit_complete() {
        if self.rw_handle.get().is_some() {
            self.rw_buffer.set(self.rw_handle.get().unwrap().get_buffer().unwrap());
        } else if self.ro_handle.get().is_some() {
            self.ro_buffer.set(self.ro_handle.get().unwrap().get_buffer().unwrap());
       }
    }
}
```

The beauty here is that only the top-level of code has to think about the fact that there will be multiple different types of buffers passed to the underlying peripheral, the underlying peripheral code just cares that whatever type it was passed can be used as a static read-only slice. Also, now the upper-layer code is responsible for holding onto buffers until DMA completes, but the interface of BufferHandle prevents any unsoundness resulting from the upper-layer code accessing the buffer before the dma completes, because the buffer is not actually made accessible until the DMA operation is finished.

Thoughts welcome!

@kupiakos @phil-levis

lschuermann:
@hudson-ayers I think this sounds like a great design and might have the ability to solve many of the issues we face across these interfaces. Here's a few thoughts (I have not run this through an actual compiler, so might be wrong):

1. I'm not sure, whether we can create such a `BufferHandle` down in the driver. If I understand this proposal correctly, at its core is the fact that both the capsule and driver hold onto a shared (`'static`) and, in case of the driver, `dyn` reference of this wrapper type, such that each can keep hold of and use it across call stacks. However, if this is created in the call to start an operation with the driver we cannot then move the resulting `BufferHandle` into some cell (e.g. `OptionalCell`) and have the driver still be able to hold a `'static` reference to it. More generally, once we have `'static dyn` (and deliberately not `mut`!) reference to this `BufferHandle` struct stored in the driver once, storing it in a type such as `OptionalCell` will be impossible as a reference to its contents fundamentally cannot be `'static` but is limited in scope, for instance through a closure.

   Instead, we should be able to just have a static number of `&'static BufferHandle`s in the capsules. These `BufferHandle`s can then be reused for operations with the respective peripherals.

2. Given that the capsule will ultimately hold either ownership or a reference to a `BufferHandle`, there's nothing preventing it from calling `transfer_done`. This is dangerous, as it would allow it to get access to a (potentially mutable) buffer which might still be subject to a DMA operation. Thus we should protect this method using either `unsafe` or a capability.

3. I understand this type is only to be used for DMA operations which do not mutate the memory in question, right? Otherwise, to solve the actual DMA unsoundness we have in our current approach (#2637), we would need to change the `BufferHandle` to not hold onto a slice, but the raw pointer and length instead.

Given the above thoughts (especially 1 + 2), we might actually get away without using monomorphism in the driver methods at all. A driver might really only need to have acess to a `&'static dyn BufferHandleTrait`, making it entirely analog to the way clients work currently. The capsule can then use its `&'static BufferHandle<B>` reference to exchange `B` at will as long as no operation is ongoing, whereas `&'static dyn BufferHandleTrait` can be used to get a `&'a [u8]` reference for synchronous operations, a `(*const u8, usize)` for asynchronous operations, and use it to signal the unsafe `transfer_done()`.

Generally, I am really excited to see this! I will try to build a PoC of this slightly modified version, happy to collaborate on this @hudson-ayers!

hudson-ayers:
> 1. I'm not sure whether we can create such a `BufferHandle` down in the driver. If I understand this proposal correctly, at its core is the fact that both the capsule and driver hold onto a shared (`'static`) and, in case of the driver, `dyn` reference of this wrapper type, such that each can keep hold of and use it across call stacks. However, if this is created in the call to start an operation with the driver we cannot then move the resulting `BufferHandle` into some cell (e.g. `OptionalCell`) and have the driver still be able to hold a `'static` reference to it. More generally, once we have `'static dyn` (and deliberately not `mut`!) reference to this `BufferHandle` struct stored in the driver once, storing it in a type such as `OptionalCell` will be impossible as a reference to its contents fundamentally cannot be `'static` but is limited in scope, for instance through a closure.
>    Instead, we should be able to just have a static number of `&'static BufferHandle`s in the capsules. These `BufferHandle`s can then be reused for operations with the respective peripherals.

I have not fully tested an implementation of this in Tock yet, just played around with some toy code in the playground. I think I understand that the issue here is that the BufferHandle we are creating on the stack would not have a static lifetime, so the underlying dma code cannot store a static reference to it -- that is definitely a problem. I am not sure I understand your proposed solution though? Would an empty (static) `BufferHandle` have to also be passed into downcalls so that it could be filled with the actual buffer reference in the upcall?

> 2. Given that the capsule will ultimately hold either ownership or a reference to a `BufferHandle`, there's nothing preventing it from calling `transfer_done`. This is dangerous, as it would allow it to get access to a (potentially mutable) buffer which might still be subject to a DMA operation. Thus we should protect this method using either `unsafe` or a capability.

Yeah I agree there should be a capability for calls to `transfer_done()`

> 3. I understand this type is only to be used for DMA operations which do not mutate the memory in question, right? Otherwise, to solve the actual DMA unsoundness we have in our current approach ([Violation of Rust's aliasing rules with DMA operations #2637](https://github.com/tock/tock/issues/2637)), we would need to change the `BufferHandle` to not hold onto a slice, but the raw pointer and length instead.

Yes, this was just designed to address read-only DMA operations. For DMA operations that write into the passed buffer, there is already an implicit requirement that the DMA be operating on `& 'static mut [u8]`.

> Given the above thoughts (especially 1 + 2), we might actually get away without using monomorphism in the driver methods at all. A driver might really only need to have acess to a `&'static dyn BufferHandleTrait`, making it entirely analog to the way clients work currently. The capsule can then use its `&'static BufferHandle<B>` reference to exchange `B` at will as long as no operation is ongoing, whereas `&'static dyn BufferHandleTrait` can be used to get a `&'a [u8]` reference for synchronous operations, a `(*const u8, usize)` for asynchronous operations, and use it to signal the unsafe `transfer_done()`.

I am not sure I understand the suggestion here, where in my design did I suggest monomorphization in the userspace driver code?

> Generally, I am really excited to see this! I will try to build a PoC of this slightly modified version, happy to collaborate on this @hudson-ayers!

I would love to see a PoC of this! I am definitely happy to give feedback on what you come up with

lschuermann:
@hudson-ayers and I just talked about this. Indeed it appears to be an issue if we were to allocate the `BufferHandle` on the stack, or anywhere non-`'static` really. Thus we'd need to change it to be essentially a lockable optional cell: a capsule holds on to some `&'static BufferHandle`s where it can swap out buffers freely. Once this is passed down to a peripheral driver, it will be locked, preventing the capsule from changing the buffer contents. A device driver must, after the respective operation completes, unlock this buffer before invoking the client callback, returning access to the buffer for the capsule.



> I am not sure I understand the suggestion here, where in my design did I suggest monomorphization in the userspace driver code?

I was talking about monomorphization in the kernel's peripheral drivers. While the above restriction does make it possible for these drivers to accept a `&'static dyn BufferHandle` and avoid monomorphization, as @hudson-ayers pointed out, it may in many cases be beneficial to use monomorphization in order to allow inlining and compiler optimizations across call invocations (e.g. for buffer length checks). The good thing is that is can be decided on a per-driver basis, based on empirical measurments.

bradjc:
Do I have this right:

Create a new type `HardwareBuffer` that stores something that can provide a `&[u8]`. The `HardwareBuffer` then ensures that while the hardware has access to the buffer nothing else in rust can access the buffer.

?

lschuermann:
> Create a new type `HardwareBuffer` that stores something that can provide a `&[u8]`. The `HardwareBuffer` then ensures that while the hardware has access to the buffer nothing else in rust can access the buffer.

Rather that the buffer cannot be swapped or mutated while the hardware is accessing it, and the peripheral is only handed a trait object reference to it, in order to abstract over multiple buffer types.

bradjc:
> > Create a new type `HardwareBuffer` that stores something that can provide a `&[u8]`. The `HardwareBuffer` then ensures that while the hardware has access to the buffer nothing else in rust can access the buffer.
> 
> Rather that the buffer cannot be swapped or mutated while the hardware is accessing it, and the peripheral is only handed a trait object reference to it, in order to abstract over multiple buffer types.

What prevents `BufferHandle` from mutating the buffer (while HW has access to it)? It feels like something here is still fundamentally unsafe, but Hudson's example isn't marking anything as `unsafe`.

phil-levis:
> What prevents `BufferHandle` from mutating the buffer (while HW has access to it)? It feels like something here is still fundamentally unsafe, but Hudson's example isn't marking anything as `unsafe`.

Hudson and I talked about this yesterday -- I agree, it seems problematic that both the creator and the holder of the handle can have a reference to the buffer. There's effectively a run-time check that you call `transfer_done` before you can get the buffer back, and that the creator (lower layer) doesn't hold on to the memory past them. One thought we discussed on making the borrow checker happy was using raw pointers, although this of course has its own issues. Plus I think it would preclude virtualization, which necessitates two handles.

I personally don't think Hudson's approach will work, but it's definitely closer/better than previous approaches. One other point I brought up with Hudson is that this really is a narrow case where we need 2 variants of a trait. My personal take is that generalization for a bounded 2 cases isn't necessary. If we had >2 cases, or a not precisely bounded set, then we want to generalize (e.g., as with `time::Ticks`, ).

hudson-ayers:
> What prevents `BufferHandle` from mutating the buffer (while HW has access to it)? It feels like something here is still fundamentally unsafe, but Hudson's example isn't marking anything as `unsafe`.

`unlock()` would need to be unsafe, such that it could only be called by the peripheral

phil-levis:
The place where we ran into the mut/imut problem before was RSA keys:

phil-levis:
> `unlock()` would need to be unsafe, such that it could only be called by the peripheral

We've re-invented mutexes! Transitioning from compile-time to run-time checks is definitely a drawback. It might be worth the code de-duplication, though.

That being said, I don't think we want to use `unsafe` for this purpose. Other code could unlock it as well, just not capsules. We want to be sure that only the locker can unlock it. Could we do this with a functional approach (lock and unlock change the type, stored in an enum)?

phil-levis:
After going down the two traits path for a while, I decided it just wasn't workable. With RSA it was OK because the traits were pretty narrow. For digests, however, there are many variants; we have capsules like `virtual_digest` that is simultaneously an HMAC and SHA. 

Unlike with RSA keys, it seems restrictive to require a client to choose mutable or immutable data. It's completely reasonable to compute a hash over data in flash and data in RAM.  

The current implementation therefore puts mutable and immutable in a single trait. There's an `add_data` and an `add_mut_data`, and two callbacks. To better support this with `LeasableBuffer` types, there's now a `LeasableBufferDynamic`, which behaves similarly to a `MutImutBuffer`: it isn't exposed in APIs but can be used internally for storing state. It can store either a `LeasableBuffer` or a `LeasableMutableBuffer`.

This implementation feels pretty clean to me; the only place where the logic required a refactoring was in the `Hmac` peripheral for lowrisc. At the bottom level, it borrows the type of leasable buffer as a `&dyn Index<usize, Output = u8>`, so it can be called with either type. Since the lowrisc implementation has a FIFO and not DMA, the borrow with an anonymous lifetime works.

phil-levis:
Changed to Draft since there are now some further updates to Digest that are incomplete.

chips/lowrisc/src/hmac.rs
```diff
@@ -224,35 +217,21 @@ impl<'a> hil::digest::DigestData<'a, 32> for Hmac<'a> {
...
-    ) -> Result<usize, (ErrorCode, &'static mut [u8])> {
-        let regs = self.registers;
-
-        regs.cmd.modify(CMD::START::SET);
-
-        // Clear the FIFO empty interrupt
-        regs.intr_state.modify(INTR_STATE::FIFO_EMPTY::SET);
-
-        // Enable interrupts
-        regs.intr_enable.modify(INTR_ENABLE::FIFO_EMPTY::SET);
```
hudson-ayers, line 236: These register accesses were deleted and I do not see anywhere else they have been inserted, is that intentional?

doc/reference/trd-digest.md
```diff
@@ -0,0 +1,168 @@
...
+5 Capsules
+===============================
+
+The Tock kernel provides several capsules for digests:
+
+
+6 Authors' Address
+=================================
+
+    Alistair Sinclair
```
hudson-ayers, line 160:
```suggestion
    Alistair Francis
```
Unless Alistair has a different last name I do not know about :)

capsules/src/virtual_hmac.rs
```diff
@@ -311,17 +342,27 @@ impl<
...
+                            node.add_mut_data_done(Err(err), slice);
+                        }
+                    }
                 }
-                return;
-            }
+                true
+            });
+
+            //node.data.insert(data);
```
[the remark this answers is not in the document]
phil-levis: Yes.

doc/reference/trd-digest.md
phil-levis: :)

boards/opentitan/src/tests/hmac.rs
```diff
@@ -5,12 +5,13 @@ use kernel::hil::digest::{self, Digest, DigestData, DigestVerify, HMACSha256};
...
 use kernel::utilities::cells::TakeCell;
 use kernel::utilities::leasable_buffer::LeasableBuffer;
+use kernel::utilities::leasable_buffer::LeasableMutableBuffer;
 use kernel::{debug, ErrorCode};
 
 static KEY: [u8; 32] = [0xA1; 32];
 
 struct HmacTestCallback {
-    add_data_done: Cell<bool>,
+    add_mut_data_done: Cell<bool>,
```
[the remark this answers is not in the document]
phil-levis: True.

boards/opentitan/src/tests/sip_hash.rs
```diff
@@ -43,7 +43,11 @@ impl<'a> SipHashTestCallback {
 }
 
 impl<'a> hasher::Client<'a, 8> for SipHashTestCallback {
-    fn add_data_done(&'a self, result: Result<(), ErrorCode>, data: &'static mut [u8]) {
+    fn add_data_done(&'a self, _result: Result<(), ErrorCode>, _data: &'static [u8]) {
```
kupiakos, line 46: Just a quick check: is there a reason this is `&'a self` and not `&self`?

boards/opentitan/src/tests/hmac.rs
```diff
@@ -111,20 +121,22 @@ fn hmac_check_verify() {
     let hmac = &perf.hmac;
 
     let callback = unsafe { static_init_test_cb() };
-    let buf = LeasableBuffer::new(callback.input_buffer.take().unwrap());
+
+    let _buf = LeasableMutableBuffer::new(callback.input_buffer.take().unwrap());
```
hudson-ayers, line 125:
```suggestion
    callback.input_buffer.take().unwrap();
```
it seems like the construction of the leasable buffer is unneeded?

boards/opentitan/src/tests/hmac.rs
```diff
@@ -86,19 +95,20 @@ fn hmac_check_load_binary() {
     let hmac = &perf.hmac;
 
     let callback = unsafe { static_init_test_cb() };
-    let buf = LeasableBuffer::new(callback.input_buffer.take().unwrap());
+    let _buf = LeasableMutableBuffer::new(callback.input_buffer.take().unwrap());
```
hudson-ayers, line 98:
```suggestion
    callback.input_buffer.take().unwrap();
```
it seems like the construction of the leasable buffer is unneeded

hudson-ayers (approved):

boards/opentitan/src/tests/hmac.rs
phil-levis: Needed if the test is on hardware.

doc/reference/trd-digest.md
```diff
@@ -0,0 +1,314 @@
...
+
+*Message authentication codes (MACs)* are a method for providing
+integrity when both the generator and checker share a secret. MACs are
+a distinct integrity mechanism than digests. They provide both
+integrity and authenticity that the message came from a certain sender
+(who holds the secret). Some MACs, such as
+[HMAC](https://en.wikipedia.org/wiki/HMAC), are built on top of hash
+functions.
+
+This document describes Tock's traits for com and their semantics for
```
lschuermann, line 58:
```suggestion
This document describes Tock's traits and their semantics for
```

doc/reference/trd-digest.md
```diff
@@ -0,0 +1,314 @@
...
+        data: LeasableMutableBuffer<'static, u8>,
+    );
+}
+```
+
+The `data` parameters of `add_data_done` and `add_mut_data_done` indicate what
+data was added and what remains to be added to the digest. If either callback
+has a `result` value of `Ok(())`, then the active region of `data` MUST be zero
+length and all of the data in the active region passed through the corresponding
+call MUST have been added to the digest.
```
lschuermann, line 128: This does not match my understanding of `LeasableBuffer`, but as we've talked about on the call of June 3rd, the API is currently underspecified and may require some more work. I think this is okay here, and did write you an email already @phil-levis.

doc/reference/trd-digest.md
```diff
@@ -0,0 +1,314 @@
...
+`DigestData` and `DataHash`. The `DigestDataVerify` trait is for a
+client that implements both `DigestData` and `DigestVerify`.  The
+`Digest` trait is for a client that implements `DigestData`,
+`DigestHash`, and `DigestVerify`. It also adds an additional method,
+`set_client`, which allows it to store a `Client` as a single
+reference and use it for all of the client callbacks (`add_data`,
+`add_mut_data`, `hash_done`, and `verification_done`). A digest
+implementation that implements `set_client` MAY choose to not
+implement the individual client set methods for the different traits
+(e.g., `DigestData::set_client); if it does so, each of these client
```
lschuermann, line 221:
```suggestion
(e.g., `DigestData::set_client`); if it does so, each of these client
```

lschuermann (dismissed):
Did a careful pass through the TRD and skimmed over the rest, looks good generally, thanks @phil-levis! Just some minor notes on the TRD and the comment regarding `LeasableBuffer`.

phil-levis:
> For what its worth, the mutable/immutable buffer passing approach has been a topic of the last core team call. In essence, Hudson, Alyssa and I agree that this is a good solution to solve the particular issue of passing mutable/immutable buffers down the Digest HIL, and unblocks the signature verification work, while we should continue to explore more integrated buffer passing APIs in the near future:

Sounds good to me. I do think the mutable/immutable tension has pretty limited scope, basically for digests and public keys. But the DMA issue is a deeper and separate one. Having a single simple approach that solves both is of course better than two different ones. The Digest HIL in this PR is definitely intended to be in-flux, and I'm open to improvements.

doc/reference/trd-digest.md
phil-levis: These semantics are intended only for `Digest`; it's fine for `LeasableBuffer` to behave differently in different APIs. 

The reasoning here relates to where the iteration occurs. If you have a buffer covering 1000 bytes, but can only do 64 bytes at a time, should it be that the callee iterates 16 times, or the caller? I think it should be the callee, as there's potentially many callers be callee. Replicated code is more likely to be buggy and wastes space.

phil-levis: The `LeasableBuffer` return type, though, is extremely useful for the `Err` case -- it can tell you what's remaining. If you were to receive an error code which suggests you can resume, you can pass the buffer back in.

hudson-ayers (approved):

kernel/src/utilities/leasable_buffer.rs
```diff
@@ -37,18 +37,32 @@ use core::ops::{Bound, Range, RangeBounds};
...
-    internal: &'a mut [T],
+    internal: &'a [T],
     active_range: Range<usize>,
 }
 
-impl<'a, T> LeasableBuffer<'a, T> {
+pub enum LeasableBufferDynamic<'a, T> {
+    Immutable(LeasableBuffer<'a, T>),
+    Mutable(LeasableMutableBuffer<'a, T>),
+}
```
alistair23, line 59: What about re-using `MutImutBuffer` here and storing the `active_range` separately?

That way we can keep all `Immutable`/`Mutable` buffers inside the one type. I feel that might simplify things if we ever migrate t a new way of doing it.

I has the obvious downside of converting to/from `LeasableBuffer` and `LeasableMutableBuffer` is a little clunkier, but it shouldn't be too bad

chips/lowrisc/src/hmac.rs
```diff
@@ -51,35 +54,31 @@ register_bitfields![u32,
...
+                    ],
+                    CMD [
+                        START OFFSET(0) NUMBITS(1) [],
+                        PROCESS OFFSET(1) NUMBITS(1) []
+                    ],
+                    STATUS [
+                        FIFO_EMPTY OFFSET(0) NUMBITS(1) [],
+                        FIFO_FULL OFFSET(1) NUMBITS(1) [],
+                        FIFO_DEPTH OFFSET(4) NUMBITS(5) []
+                    ]
```
alistair23, line 71: Why change these?

chips/lowrisc/src/hmac.rs
```diff
@@ -88,67 +87,64 @@ impl Hmac<'_> {
...
+        self.data.take().map_or(false, |buf| match buf {
+            LeasableBufferDynamic::Immutable(mut b) => {
+                let count = self.process(&b, b.len());
+                b.slice(count..);
+                true
+            }
+            LeasableBufferDynamic::Mutable(mut b) => {
+                let count = self.process(&b, b.len());
+                b.slice(count..);
+                true
```
alistair23, line 134: I'm a little confused. It looks like data_progress() will return `true` if we have any `self.data`. No matter if we wrote all the data or not.

Which means in `add_data()` we are going to trigger an interrupt with `INTR_TEST::FIFO_EMPTY::SET`, even if the FIFO isn't empty. Then the `self.data_progress()` in the interrupt handler will return true as well, even if we didn't write all the data.

Did you test this with a large buffer that fills the FIFO at least twice?

alistair23 (changes requested):

kernel/src/utilities/leasable_buffer.rs
phil-levis: I found it easier to keep the 3 values coupled, rather than combining them in a LeasableBuffer, breaking them apart in separate fields, then regenerating the LeasableBuffer on the callback.

chips/lowrisc/src/hmac.rs
phil-levis: Hunh -- this was not intentional. Fixed.

chips/lowrisc/src/hmac.rs
phil-levis: Good catch. I have only been running the CI tests, as I do not have an earlgrey-cw310. The QEMU tests don't do anything.

I think this code needs to test if the length is 0 (we're done). If so, then it should return false and the caller should signal the callback. Otherwise, it should feed in the next block.

phil-levis: Please take a look now.

alistair23: You can run the tests on Verilator as well, that way you can test the full system without an FPGA. It's a little slow, but pretty usable still: https://github.com/tock/tock/tree/master/boards/opentitan#verilator

chips/lowrisc/src/hmac.rs
```diff
@@ -224,37 +258,60 @@ impl<'a> hil::digest::DigestData<'a, 32> for Hmac<'a> {
...
+            // Clear the FIFO empty interrupt
+            regs.intr_state.modify(INTR_STATE::FIFO_EMPTY::SET);
+            // Enable interrupts
+            regs.intr_enable.modify(INTR_ENABLE::FIFO_EMPTY::SET);
+            let ret = self.data_progress();
 
-        // Call the process function, this will start an async fill method
-        let ret = self.data_progress();
+            if ret {
+                regs.intr_test.modify(INTR_TEST::FIFO_EMPTY::SET);
```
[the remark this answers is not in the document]
alistair23: It is correct, it's a little odd though.

We need to do this if we can an `add_data()` call that does NOT fill up the buffer. If we write all of the data in one go it's possible we won't get a FIFO empty interrupt (image writing a single byte). We could use deferred callbacks, but this feature already exists in the hardware so it seemed easier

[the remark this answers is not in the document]
phil-levis: Shouldn't there be an interrupt even if the FIFO isn't filled? I.e., the interrupt is EMPTY, not NOT_FULL. A single byte write should push an entry into the FIFO, which then falls idle. If you aren't getting an interrupt, this seems like a hardware bug/issue we should reach out to the OT team about.

The problem with this approach is that you're going to handle interrupts at the interrupt handling rate, rather than for each FIFO. E.g., it's possible you get an interrupt but there's actually no space because it was fast.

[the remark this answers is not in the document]
phil-levis: This came up on the call today. The lowrisc HMAC engine is fast enough (for SHA) that it's quite possible that it can process faster than the memory bus can feed it data. This means that `lowrisc::Hmac::process` might just loop for the entire passed slice.

chips/lowrisc/src/hmac.rs
phil-levis: This instructions don't seem to work

https://github.com/tock/tock/tree/master/boards/opentitan#build-boot-romotp-image

Says to run ninja, but there is no build.ninja file. There doesn't seem to be an `otp_image` anymore?

Are the Tock instructions out of sync with the OpenTitan instructions?

bradjc (approved):

chips/lowrisc/src/hmac.rs
alistair23: Did you use the correct commit?

https://github.com/tock/tock/pull/3056 has more up to date instructions, but there are still PMP and other issues to I wouldn't recommend using that

phil-levis: I used the most recent bitstream. :) I then used the recommended commit and that seems to be running into trouble on the OpenTitan side, I will dig some more tonight.

phil-levis: I can't find a set of instructions that works. If I go back to the recommended commit on OT  -- lowRISC/opentitan@199d45626f8a7ae2aef5d9ff73793bf9a4233711 -- there doesn't seem to be  a way to build boot_rom, which is needed in this step:

https://github.com/tock/tock/tree/master/boards/opentitan#build-and-run-tock

```shell
pal@barong:~/src/opentitan$ git status
HEAD detached at 199d45626
nothing to commit, working tree clean
pal@barong:~/src/opentitan$ grep -R boot_rom
hw/vendor/pulp_riscv_dbg/tb/boot_rom.sv:module boot_rom (
hw/vendor/pulp_riscv_dbg/tb/mm_ram.sv:    boot_rom boot_rom_i (
hw/vendor/pulp_riscv_dbg/tb/Makefile:RTLSRC_TB		:= boot_rom.sv \
hw/vendor/pulp_riscv_dbg/tb/Makefile:RTLSRC_VERI_TB          := boot_rom.sv \
Binary file .git/objects/pack/pack-0db5ff8687ec6674c9c9988e048256cba5150471.pack matches
Binary file .git/index matches
doc/ug/quickstart.md:$ ./hw/top_earlgrey/Vchip_earlgrey_verilator --rominit=./sw/device/sim/boot_rom/boot_rom_sim_verilator.scr.39.vmem --flashinit=./sw/device/sim/examples/hello_world/sw.vmem
doc/ug/quickstart.md:Rom initialized with program at ./sw/device/sim/boot_rom/rom.vmem
```

alistair23: Hmmm.... What error do you get when trying to build it with Ninja?

phil-levis: Where in the instructions is the command to build it with Ninja?

alistair23: https://docs.opentitan.org/doc/getting_started/setup_fpga/

You should just be able to run

```sh
./meson_init.sh
ninja -C build-out sw/device/lib/testing/test_rom/test_rom_export_sim_verilator
ninja -C build-out sw/device/otp_img/otp_img_sim_verilator.vmem
```

to build everything you need

phil-levis:
I've tested the OpenTitan HMAC peripheral, it now passes tests. @alistair23

phil-levis:
@alistair23 Can you re-review and clear your changes requested?

hudson-ayers (approved):

PR #3045, opened 2022-05-24
Autogenerate register definitions for lowrisc peripherals

cfrantz:
### Pull Request Overview

This pull request adds auto-generated register definitions for `chips/lowrisc` peripherals.

### Testing Strategy

This pull request was tested manually.

### TODO or Help Wanted

This PR only converts the UART to use the autogenerated definitions.  Follow-up PRs will be needed to convert the remaining peripheral drivers.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


jrvanwhy:
The CI failure appears to be unrelated to this PR.

chips/lowrisc/src/registers/adc_ctrl_regs.rs
```diff
@@ -0,0 +1,129 @@
+// Generated register constants for adc_ctrl
+
+// Built for earlgrey_silver_release_v5-5548-g933e261c9
+// https://github.com/lowRISC/opentitan/tree/933e261c91e67309b242d87a97e42b8af1bd17ef
+// Tree status: modified
```
alistair23, line 5: Can we generate these from a clean tree as well?

chips/lowrisc/src/registers/adc_ctrl_regs.rs
```diff
@@ -0,0 +1,129 @@
...
+// Build date: 2022-05-24T16:20:07
+
+// Copyright information found in source file:
+// Copyright lowRISC contributors.
+
+// Licensing information found in source file:
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
```
alistair23, line 13: This should also probably be updated to match Tock's LICENSE

alistair23 (changes requested):

chips/lowrisc/src/registers/adc_ctrl_regs.rs
cfrantz: Oops, how embarassing.  Fixed.

chips/lowrisc/src/registers/adc_ctrl_regs.rs
cfrantz: Can you clarify what you mean?  AFAICT, tock is dual-licensed under the Apache 2.0 and MIT licenses.  This comment identifies the license as an Apache 2.0 license.

chips/lowrisc/src/registers/adc_ctrl_regs.rs
```diff
@@ -0,0 +1,129 @@
+// Generated register constants for adc_ctrl
+
+// Built for earlgrey_silver_release_v5-5548-g933e261c9
+// https://github.com/lowRISC/opentitan/tree/933e261c91e67309b242d87a97e42b8af1bd17ef
```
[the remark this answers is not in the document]
cfrantz: Can you clarify who is meant by "we" in your comment?  What is the process for updating the supported commit?  As we move towards final RTL for the Earlgrey ASIC, I'm going to want to move this forward periodically to account for fixes in the hardware and in the boot software (ie: ROM).

chips/lowrisc/src/registers/adc_ctrl_regs.rs
lschuermann: *Disclaimer: not a lawyer*. Doesn't being dual-licensed under the Apache-2.0 and MIT license mean that the entire codebase can be used under the terms and conditions of *either* of those licenses? This would imply that all contributions to Tock must grant usage and distribution rights up to the extend defined by *both* the Apache-2.0 and MIT license. If I understand this correctly, if only a single file were not dual-licensed, effectively the codebase in its entire would no longer be dual-licensed. In that case, we are not able to accept any file with an explicit license header, except if it declares an Apache-2.0/MIT dual licensing scheme as defined for the rest of Tock.

chips/lowrisc/src/registers/adc_ctrl_regs.rs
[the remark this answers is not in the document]
alistair23: We as in the [Tock OpenTitan board](https://github.com/tock/tock/tree/master/boards/opentitan#programming) and the OT working group

The process just involves bumping the README and running all of the tests on the new bitstream, this is an example of a recent-ish bump: https://github.com/tock/tock/commit/9a9a114d12c7c2dae9f359164e4bc3c4957b0732. Generally some register updates are also required to handle changes to the RTL.

chips/lowrisc/src/registers/adc_ctrl_regs.rs
alistair23: I mean it should be dual licensed as described here: https://github.com/tock/tock#license

But also there is no `LICENSE` file for details in Tock, so this should point to [LICENSE-APACHE](https://github.com/tock/tock/blob/master/LICENSE-APACHE) and [LICENSE-MIT](https://github.com/tock/tock/blob/master/LICENSE-MIT).

It's also a little odd that it is a different copyright then all the other files, but I don't think that's a blocking issue.

hudson-ayers (commented):
Holding off on approving as I agree with Alistair that the version used for these registers needs to match the version we run OT QEMU testing on. But overall this looks good, and I think it is fine for followup PRs to convert subsequent peripherals after UART.

chips/lowrisc/src/registers/adc_ctrl_regs.rs
jrvanwhy: The code in this PR is dual-licensed under the Apache 2.0 and MIT licenses, as is required by the Tock project. There is no license header because the Tock project has elected not to use license headers.

I'll reiterate that OpenTitan would prefer for Tock to adopt license headers. I suspect that having license headers would have avoided this conversation.

hudson-ayers: @jrvanwhy I am a little confused, do you disagree that one of the following needs to happen:

a) remove this license header, since it implies that this file is only apache licensed and points to a LICENSE file that does not exist in this repository OR

b) update this header as per https://github.com/lowRISC/opentitan/pull/12963 such that it clearly specifies both Apache and MIT licenses

jrvanwhy: > @jrvanwhy I am a little confused, do you disagree that one of the following needs to happen:
>
> a) remove this license header, since it implies that this file is only apache licensed and points to a LICENSE file that does not exist in this repository OR

This isn't a license header. These files were generated by a tool, and that tool mentions the license of some of its input files in its output, but that license is not the license of the output files.
 
> b) update this header as per [lowRISC/opentitan#12963](https://github.com/lowRISC/opentitan/pull/12963) such that it clearly specifies both Apache and MIT licenses

That's not necessary, but very possible.

If I recall correctly, the last time we discussed license headers in Tock, there was consensus that we should either have license headers on all files or license headers on no files.

lschuermann: > This isn't a license header. These files were generated by a tool, and that tool mentions the license of some of its input files in its output, but that license is not the license of the output files.

Okay, I actually missed this part. I think it's totally fine to include the licese of source files, but (a) I don't fully understand why we'd want to do this and (b) we should definitely add some additional markup to make it immediately obvious this isn't trying to apply a license to this file as part of the Tock codebase. Perhaps add some indentation?

cfrantz: I've updated the generated files such that they (a) explicitly state they're licensed under either Apache or MIT, and (b) explicitly state the original source location and licence of the reference definitions.

hudson-ayers: I am happy with the updated text

chips/lowrisc/src/registers/adc_ctrl_regs.rs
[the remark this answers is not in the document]
hudson-ayers: @cfrantz is targeting the OT commit that Tock currently supports an issue? Do we need to update Tock to support the commit that this PR targets?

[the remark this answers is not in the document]
cfrantz: We'll have to update the OT commit for Tock.  I have no desire to backport the register generator script or build-system automation to the currently supported commit.

chips/lowrisc/src/registers/adc_ctrl_regs.rs
alistair23: I'm worried the incorrect `SPDX-License-Identifier` will break automation tools. Can we just remove the original source license information?

chips/lowrisc/src/registers/adc_ctrl_regs.rs
```diff
@@ -0,0 +1,133 @@
+// Generated register constants for adc_ctrl.
+// This file is licensed under either of:
+//   Apache License, Version 2.0 (LICENSE-APACHE <http://www.apache.org/licenses/LICENSE-2.0>)
+//   MIT License (LICENSE-MIT <http://opensource.org/licenses/MIT>)
+
+// Built for earlgrey_silver_release_v5-5654-g222658011
+// https://github.com/lowRISC/opentitan/tree/222658011c27d6c1f22f02c7f589043f207ff574
```
alistair23, line 7: I started looking at bumping the OT version that Tock supports

This SHA doesn't seem to be part of the master branch of OT, so we would have to wait for this commit to be merged first though

alistair23 (changes requested):

chips/lowrisc/src/registers/adc_ctrl_regs.rs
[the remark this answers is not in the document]
hudson-ayers: If we update to https://github.com/lowRISC/opentitan/pull/12963/commits/e2fd59b68d3830de70f3d9c40e08f37308dc28b5 , which is now merged into OT master, does that work for you?

alistair23:
I had a go at performing the [update for you](https://github.com/tock/tock/pull/3056). It looks like the flash controller and CSRNG still aren't passing the tests. There probably are still some issues with ePMP as well. It seems like there were a few changes there that weren't communicated back to us.

It would be great if you could have a look and see if you can get the remaining tests passing

chips/lowrisc/src/registers/adc_ctrl_regs.rs
phil-levis: I'm a big fan of license headers...

alistair23: Great! But this isn't the correct license header. Anyone or any tool looking at the `SPDX-License-Identifier` will see the incorrect license

cfrantz:
Converted to draft until after OpenTitan formal release process.

cfrantz:
Obsoleted by #3587

PR #3046, opened 2022-05-24
UART HIL TRD Implementation

colin-r-schultz:
### Pull Request Overview

This pull request updates the UART HIL to match the TRD proposed in #2756. It  implements the new HIL for the sam4l and nrf52 chips. For an explanation of the design decisions of the new HIL, refer to #2756.

### sam4l

**transmit/receive_buffer():**
Much of the implementation remains the same.

**transmit/receive_abort():**
A new state (`Aborted([transfer_type,] usize, Result<(), ErrorCode>, uart::Error)`) was added to indicate that an operation was aborted, and to save the size and error information. `abort()` operation results were updated to indicate the corresponding callback response: `Callback(true)` if there was an outstanding operation, which is now canceled; `Callback(false)` if there was an outstanding operation, which was not canceled, and `NoCallack` if there was no outstanding operation and therefore no future callback.

**transmit/receive_character():**
This pull request adds an implementation of character-based receives and transmits. In sam41, the client may configure the character width to be 9 bits and use transmit/receive_character to operate on this scale.

`receive_character()` uses a 2-byte DMA buffer so that the abort routine, error catching, and interrupt handling can be unified with buffer reception. The buffer must be 2 bytes to support the 9 bit character mode. Additionally, the `DMA_Transmitting` and `Aborted` states both include an additional `TransferType` field which determines whether the ongoing or aborted operation was a character or buffer reception.

### nrf52840

**transmit/receive_buffer():**
Much of the implementation remains the same. The `transmit_buffer` method now returns `ErrorCode::OFF` when the UART is not enabled before the call to `transmit_buffer()`. `receive_buffer`  only changes the function signature to match the new HIL.

**transmit/receive_abort():**
Transmit_abort:  transmit_abort works without any deferred calls or new state tracking. It simply checks if the uart tx_ready(), and if so, returns noCallBack, otherwise it aborts the transmission. There is a cautionary case for if the transmission is done, but a callback has not yet been issued. This case likely will never be hit, but it is there to conform to the HIL guidelines. Unsure if it is necessary.

Receive_abort: This works without any deferred calls or new state tracking. It simply checks if rx_buffer is some, and if so we return NoCallBack since we had nothing to abort. Otherwise, there is an outstanding reception we can abort, and the abort_in_progress field is set, and then it sets the stoprx task on the nrf52840. No changes were made in the handle_interrupt code since it appeared to follow the expected behavior already. There exists an edge case of trying to abort a reception that had already completed in the DMA hardware, but no callback had been issued yet. Handling this case did not seem necessary because by testing if the rx_buffer is some, we knew that the underlying DMA op isn’t completed or ready for a callback yet, or else we would have had the endrx interrupt come in, in which case the rx_buffer would have been handed back to the received_buffer callback and would be None to us. Therefore, this likely was not a case we would ever enter, so we would never need to return a callback(false). Essentially, if there is an outstanding reception, I think we should be able to abort it regardless.

**transmit/receive_character():**
The nrf52840 does not support 9 bit transmission. Therefore, it simply returns error code NOSUPPORT. 

**configure/configuration():**
The nrd52840 only supports 1 stop bit, 8bit width, No parity bit. While i believe it can support Hardware flow control, Tock previously ensured that we never enabled HW flow control (maybe since we use DMA?). We kept with this decision and continue to not allow HW flow control to be set.

This means that both these traits are simple, with most of the set_<feature> functions of the configure trait just returning ErrorCode::NOSUPPORT (except for set_baud_rate which is supported by the NRF52840. The configuration trait takes advantage of knowing that configure is enforcing these, and just returns hard coded uart parameter values for everything other than baud rate.

### Testing Strategy

We tested the two implementations on the imix and nrf52840dk boards. Because the UART virtualizer has not been updated yet, we created minimal versions of the kernel for each board that allowed us to interact with the hardware UART implementations directly. The code for these testing platforms can be found in `boards/imix-mini` and `boards/nordic/nrf52840dk-mini`. We have not yet updated a number of capsules that currently assume the interface of the old UART HIL. As such, we have not updated the normal imix/nrf52840dk main.rs files, as they assume the existence of these capsules, which are currently commented out in capsules/src/lib.rs. 

For the sam4l implementation, we tested basic operation of the UART via the USB interface on the imix board. This includes transmitting and receiving single characters as well as larger buffers, and aborting these operations. All testing was performed with a single UART configuration (baud rate, parity, bit width, etc.).

For the nrf52840 implementation, we tested basic operation for the UART via the USB interface on the board. This included transmitting and receiving multi-byte  buffers, but not single characters (via *_character() ). To test transmit, we set up a transmit client as the debug console, and then try to call the `debug!` macro after initialization if over. A successful print indicates that transmit_buffer is working correctly. We then tested `transmit_abort()` by running a test which makes a second debug! statement then immediately calls abort on the underlying uart hardware. We tested this way because we do not have the virtualizer implemented to handle multiple clients, or manage multiple aborts. This simple test showed that the implementation does abort an outstanding transmission (it only printed the first character of the aborted debug! statement).

For receive abort it is much  the same. We created an Echo struct to wrap the Uart, which will try to receive a set length of bytes from the unix console, and then echo them via debug!. The code tests receive abort by calling it immediately after calling receive_buffer.

### TODO or Help Wanted

We need help updating the UART implementations for other chips and the virtualizer to conform to the new HIL.

- [ ] apollo3 (@alistair23 )
- [ ] arty_e21 (@alexandruradovici )
- [ ] esp32 (@alexandruradovici )
- [ ] imxrt10xx (@valexandru ) 
- [ ] litex (@lschuermann )
- [ ] lowrisc (@alistair23 )
- [ ] msp432 (@lebakassemmerl )
- [ ] rp2040 (@alexandruradovici )
- [ ] sifive (core team?)
- [ ] stm32f303xc (@alexandruradovici )
- [ ] stm32f4xx (@alexandruradovici )
- [ ] swervolf-eh1 (@alistair23 )

We also need help implementing some capsules which depend on the UART HIL

- [ ] nrf51822_serialization 
- [x] segger_rtt
- [ ] usb
- [x] virtual_uart

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


lschuermann:
I've just ported the `sifive` implementation over to the HIL, it was pretty straightforward generally. I realize that for comments on the draft HIL I'm 3 days late to the party, but here's what I noticed while implementing it (and after re-reading the discussions of #2756):

- The text of _2 `Configuration` and `Configure`_ explains the consequences of a configuration with a character smaller than 8 bits, but does not mention that wider characters must use the `{transmit,receive}_character` calls. We might want to add that.

- `Configure::configure` seems slightly redundant, as all other parameters can be set using their respective methods, _except_ if the `Parameter` struct passed in there should be applied in an atomic fashion (either completely or not at all). If that is the case, we should document this expectation to implementations.

- When using `Configure::configure`, the configured baudrate is not returned, in contrast to calling `Configure::set_baud_rate`. If `Configure::configure` is indeed designed to be an atomic configuration mechanism for all of the different options, we can make it functionally equivalent to the individual parameter-setting methods by also returning the configured baudrate.

- The specification as of now does not consider the case when calling `transmit_buffer` with `tx_len = 0`. If a compliant implementation has to accept `tx_len = 0`, it may likely -- depending on the hardware -- require a deferred call to perform a callback just for this edge case. It seems much easier to just special case `tx_len = 0` and define `ErrorCode::SIZE` to be issued when `tx_len` is larger than the passed slice OR zero.

- Should the default implementation of `{transmitted,received}_character` do nothing, or panic if called? I'm not a fan of panicing, but when this is called without a proper callback implementation either a client forgot to provide a callback implementation (which a panic helps noticing during development) or something is seriously off with the UART implementation.

- If I understand the HIL correctly, not implementing the `receive_buffer` method is no longer possible, meaning that all transmit-only UART implementations in Tock are technically non-compliant. While we can fix them to also support receive, it appears at least somewhat common for new boards to only support UART transmit at first (for instance, `apollo3`, `sifive`, `swervolf-eh1`).

2 nits:

- We might want to rename `flow_control` to `hw_flow_control`.

- `ENOSUPPORT` should be changed to `NOSUPPORT`.

I'll create a PR to discuss these observations.

Finally, I have added a default implementation for `Configuration::get_configuration`, given we already have all of the information required to create the `Parameters` struct available through the `Configuration` trait methods.

bradjc:
We should get this across the finish line.

The new platforms need to be split into their own PRs.

lschuermann:
Closing in favor of #4181 as discussed.

Issue #3035, opened 2022-05-04
Snync Screen syscalls


dcz-self:
I'm trying to implement a syscall driver for the Screen interface, and I'm seeing that all of the synchronous HIL calls have been turned async:

> ### Command number: `23` 
>
>    **Description**: Get the screen's resolution
>
>    **Argument 1**: unused
>
>  **Argument 2**: unused
>
>    **Returns**: Ok(()) followed by a callback with the resolution, BUSY if another command is in progress.

Is this actually desired? The HIL already says:

>     /// This function is synchronous as the driver should know this value without
>    /// requesting it from the screen.
>    fn get_resolution(&self) -> (usize, usize);

so it can be called synchronously even when some operation is in progress, and even if another client is currently performing an operation.

I'm just itching to make the syscalls synchronous. Should I have a go at it? Or is there a good reason for them to stay async?

hudson-ayers:
@alexandruradovici

alexandruradovici:
While the HIL's function is synchronous, the screen syscall driver services several applications at once. If one application has issued a command to the driver and the driver is busy processing it, another application will have to wait for the first application's request to finish. This is why the syscall command is asynchronous.

While you can make this particular syscall synchronous, image the following race:
- app1 requests a resolution change
- app2 asks for the current resolution and gets the "old" one
- app1 requests is done and now we have a new resolution (different from the one app2 received) 

Any suggestions are welcome though :smile:

dcz-self:
Thanks for answering. I see this as an example of a wider problem though:

1. Client A requests current resolution and gets Res1
2. Client B requests resolution change to Res2
3. Now Client A has an outdated resolution

The ordering doesn't matter, because changes need to be broadcast. And if we go in that direction, then suddenly we have to broadcast all state changes.

While this may be useful in some use cases, I think it's way overkill, and I'd rather make use of a simple driver that doesn't try to multiplex. Most people are going to use only a single screen client in the userspace anyway, including me.

So I don't really see much value in multiplexing, given its complexity. But I'll happily improve the single client situation. Should I go ahead?

alexandruradovici:
I see your point. While you still have to multiplex some requests (write, etc), I think the resolution read can be synchronous. I would say go ahead and modify it.

dcz-self:
Yes, I don't mean to throw away all multiplexing (although I consider locking the entire screen for one process a valid solution), just keep it on a best-effort basis.

dcz-self:
[...]

PR #3047, opened 2022-05-25
Screen syscall/HIL improvements

dcz-self:
### Pull Request Overview

This pull request changes the Screen syscalls and HIL to incorporate more changes that seem rather simple.

It's another step on the way to https://github.com/tock/tock/pull/3011, and includes feedback from https://github.com/tock/tock/issues/3035 . Other notable changes:

- improved docs
- removed power on tracking in the last commit - it doesn't add much value, and it prevents set_power call from succeeding at startup

### Testing Strategy

This pull request was tested on LPM013M126 driver, out of tree: https://framagit.org/jazda/tock/-/commit/18e7b9bd0ee94d3c3b6b44c2f6771d12457e1ed2 (points to last commit), this *does not* exercise Setup interface.

### TODO or Help Wanted

This pull request still needs testing of the Setup interface, because I don't have the device that uses it. I'm also not ready yet to extend my own out of tree driver with this functionality just to test it.

There are some notes inline where I'd like to get feedback.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/screen.rs
+++ b/capsules/src/screen.rs
@@ -110,7 +110,6 @@ pub struct Screen<'a> {
     screen: &'a dyn hil::screen::Screen,
     screen_setup: Option<&'a dyn hil::screen::ScreenSetup>,
     apps: Grant<App, UpcallCount<1>, AllowRoCount<{ ro_allow::COUNT }>, AllowRwCount<0>>,
-    screen_ready: Cell<bool>,
     current_process: OptionalCell<ProcessId>,
     pixel_format: Cell<ScreenPixelFormat>,
     buffer: TakeCell<'static, [u8]>,
@@ -128,7 +127,6 @@ impl<'a> Screen<'a> {
             screen_setup: screen_setup,
             apps: grant,
             current_process: OptionalCell::empty(),
-            screen_ready: Cell::new(false),
             pixel_format: Cell::new(screen.get_pixel_format()),
             buffer: TakeCell::new(buffer),
         }
@@ -154,7 +152,7 @@ impl<'a> Screen<'a> {
         {
             Err(e) => CommandReturn::failure(e),
             Ok(r) => {
-                if self.screen_ready.get() && self.current_process.is_none() {
+                if self.current_process.is_none() {
                     self.current_process.set(process_id);
                     let r = self.call_screen(command, process_id);
                     if r != Ok(()) {
@@ -286,16 +284,12 @@ impl<'a> Screen<'a> {
     }
 
     fn schedule_callback(&self, data1: usize, data2: usize, data3: usize) {
-        if !self.screen_ready.get() {
-            self.screen_ready.set(true);
-        } else {
-            self.current_process.take().map(|process_id| {
-                let _ = self.apps.enter(process_id, |app, upcalls| {
-                    app.pending_command = false;
-                    upcalls.schedule_upcall(0, (data1, data2, data3)).ok();
-                });
+        self.current_process.take().map(|process_id| {
+            let _ = self.apps.enter(process_id, |app, upcalls| {
+                app.pending_command = false;
+                upcalls.schedule_upcall(0, (data1, data2, data3)).ok();
             });
-        }
+        });
     }
 
     fn run_next_command(&self, data1: usize, data2: usize, data3: usize) {
--- a/doc/syscalls/90001_screen.md
+++ b/doc/syscalls/90001_screen.md
@@ -50,7 +50,10 @@ may return OFF when power is not enabled (see screen HIL for details).
     **Argument 2**: unused
 
     **Returns**: Ok(()) followed by the ready callback if the command was successful,
-    BUSY if another command is in progress,
+    BUSY if another command is in progress.
+    
+    The callback will carry 1 as an argument if the display was turned on,
+    but configuration not fully applied. Otherwise, the argument is 0.
 
   * ### Command number: `3`
 
--- a/kernel/src/hil/screen.rs
+++ b/kernel/src/hil/screen.rs
@@ -1,4 +1,30 @@
-//! Interface for screens and displays.
+/*! Interface for screens and displays.
+
+The interfaces exposed here cover both configurable (`ScreenSetup`),
+and less configurable hardware (only `Screen`).
+
+It's composed of 4 main kinds of requests:
+- set power,
+- read configuration (e.g. `get_resolution`),
+- configure (e.g. `set_invert`),
+- write buffer.
+
+All requests, except for `Screen::set_power`, can return `OFF`
+under some circumstances.
+
+For buffer writes, it's when the display is powered off.
+
+While the display is not powered on, the user could try to configure it.
+In that case, the driver MUST either cache the value, or return `OFF`.
+This is to let the user power the display in the desired configuration.
+
+Configuration reads shall return the actual state of the display.
+In situations where a parameter cannot be configured
+(e.g. fixed resolution), they return value may be hardcoded.
+Otherwise, the driver should query the hardware directly,
+and return OFF if it's not powered.
+*/
+
 use crate::ErrorCode;
 use core::ops::Add;
 use core::ops::Sub;
@@ -151,12 +177,13 @@ pub trait ScreenSetup {
     fn get_supported_pixel_format(&self, index: usize) -> Option<ScreenPixelFormat>;
 }
 
+/// The basic trait for screens
 pub trait Screen {
     /// Returns a tuple (width, height) with the current resolution (in pixels)
     /// This function is synchronous as the driver should know this value without
     /// requesting it from the screen.
     ///
-    /// note that width and height may change due to rotation
+    /// Note that width and height may change due to rotation.
     fn get_resolution(&self) -> (usize, usize);
 
     /// Returns the current pixel format
@@ -237,6 +264,11 @@ pub trait Screen {
     /// The device may implement power independently from brightness,
     /// so call `set_brightness` to turn on/off the module completely.
     ///
+    /// To allow starting in the correct configuration,
+    /// the driver is allowed to cache values like brightness or invert mode
+    /// and apply them together when power is enabled.
+    /// If the display cannot use selected configuration, this call returns `INVAL`.
+    ///
     /// When finished, calls `ScreenClient::screen_is_ready`,
     /// both when power was enabled and disabled.
     fn set_power(&self, enabled: bool) -> Result<(), ErrorCode>;
@@ -268,6 +300,6 @@ pub trait ScreenClient {
     /// This is different from `command_complete` as it has to pass back the write buffer
     fn write_complete(&self, buffer: &'static mut [u8], r: Result<(), ErrorCode>);
 
-    /// Some screens need some time to start, this function is called when the screen is ready
+    /// Some screens need some time to start, this function is called when the screen is ready.
     fn screen_is_ready(&self);
 }


capsules/src/screen.rs
```diff
@@ -383,16 +284,12 @@ impl<'a> Screen<'a> {
     }
 
     fn schedule_callback(&self, data1: usize, data2: usize, data3: usize) {
-        if !self.screen_ready.get() {
-            self.screen_ready.set(true);
```
dcz-self, line 387: What was the point of this? This callback would have not been forwarded anywhere. I don't get it, and don't want to break this by accident.

dcz-self:
@alexandruradovici I'd appreciate your feedback.

dcz-self:
I tested this, and it works for me, so I removed the WIP.

capsules/src/screen.rs
alexandruradovici: Whenever the underlying driver returns a `command_complete` the screen driver might have to execute a queued command. This is done by the `run_next_command` function. This schedules a callback for the previous action and then runs the next one.

The lines that you are asking about are the way in which `schedule_callback` differentiates between the `screen_is_ready` and the `command_complete` notifications from the underlying driver.

capsules/src/screen.rs
```diff
@@ -563,23 +460,53 @@ impl<'a> SyscallDriver for Screen<'a> {
             2 => self.enqueue_command(ScreenCommand::SetPower(data1 != 0), process_id),
             // Set Brightness
             3 => self.enqueue_command(ScreenCommand::SetBrightness(data1), process_id),
-            // Invert On
-            4 => self.enqueue_command(ScreenCommand::InvertOn, process_id),
-            // Invert Off
-            5 => self.enqueue_command(ScreenCommand::InvertOff, process_id),
```
alexandruradovici, line 569: I suggest preserving these calls for backward compatibility, at least until we update the libtock-c API. You coud add a comment here saying that they will be removed as soon as the user space API is updated.

doc/syscalls/90001_screen.md
```diff
@@ -8,6 +8,14 @@ driver number: 0x90001
 
 The screen driver allows the process to write data to a framebuffer of a screen.
 
+This syscall driver is designed for single-client usage, for the sake of simplicity, and because a single client covers a significant portion of display use cases.
```
alexandruradovici, line 11:
```suggestion
This syscall driver is designed for single-client usage, for the sake of simplicity, and because a single 
client covers a significant portion of display use cases.
```

Please use shorter lines

doc/syscalls/90001_screen.md
```diff
@@ -55,26 +63,18 @@ The screen driver allows the process to write data to a framebuffer of a screen.
...
-  * ### Command number: `5`
-
-    **Description**: Turn off invert colors mode
-
-    **Argument 1**: unused
-
-    **Argument 2**: unused
-
-    **Returns**: Ok(()) if the command was successful, BUSY if another command is in progress.
-
```
alexandruradovici, line 77: Mark them as deprecated for now.

alexandruradovici (changes requested):
It looks good overall. I was wondering if we could make this a single app driver, inspired by the model of the LSM driver (https://github.com/tock/tock/blob/master/capsules/src/lsm303agr.rs).

The first application that sends a request would *own* the display. As soon as this application finishes or is restarted, another application would be able to *own* the driver and send commands.

capsules/src/screen.rs
```diff
@@ -162,7 +152,7 @@ impl<'a> Screen<'a> {
         {
             Err(e) => CommandReturn::failure(e),
             Ok(r) => {
-                if self.screen_ready.get() && self.current_process.is_none() {
+                if self.current_process.is_none() {
```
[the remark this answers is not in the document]
dcz-self: That was my goal. Preventing the call in the syscall driver would be redundant.

capsules/src/screen.rs
```diff
@@ -118,7 +110,6 @@ pub struct Screen<'a> {
     screen: &'a dyn hil::screen::Screen,
     screen_setup: Option<&'a dyn hil::screen::ScreenSetup>,
     apps: Grant<App, UpcallCount<1>, AllowRoCount<{ ro_allow::COUNT }>, AllowRwCount<0>>,
-    screen_ready: Cell<bool>,
```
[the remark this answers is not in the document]
dcz-self: The callback is still getting through according to the way I tested. It's just the flag that is not actually needed anywhere.

dcz-self:
> I was wondering if we could make this a single app driver

I'm not against it, but I'd rather take smaller steps for now.

alexandruradovici:
> > I was wondering if we could make this a single app driver
> 
> I'm not against it, but I'd rather take smaller steps for now.

I suggest (above) stating this in the driver's documentation, more precisely that the driver does accept commands from multiple without returning an actual error.

capsules/src/screen.rs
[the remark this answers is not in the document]
alexandruradovici: I might be missing something, but it seems to me from the code that you forward the application's command right away if there is no other process that runs a command.

capsules/src/screen.rs
```diff
@@ -563,23 +460,53 @@ impl<'a> SyscallDriver for Screen<'a> {
...
+                if let Some(screen) = self.screen_setup {
+                    CommandReturn::success_u32(screen.get_num_supported_resolutions() as u32)
+                } else {
+                    CommandReturn::failure(ErrorCode::NOSUPPORT)
+                }
+            }
             // Get Resolution Mode Width and Height
-            12 => self.enqueue_command(ScreenCommand::GetSupportedResolution(data1), process_id),
+            12 => {
+                if let Some(screen) = self.screen_setup {
```
alexandruradovici, line 476: I think you should verify if the screen is ready, otherwise return BUSY.

capsules/src/screen.rs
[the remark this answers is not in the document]
dcz-self: Yes, and that should return not BUSY, but rather OFF, because of the HIL (and I should add that to the HIL). That's also what the userspace should see, so there's no problem just forwarding the calls without interference.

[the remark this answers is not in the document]
alexandruradovici: Yes, OFF is better.

I would go for one of these:
1. schedule an upcall as soon as the display is ready so that the application does not need to poll that
    - this might be tricky, as an application that registers for this upcall after the screen has sent the ready notification should receive an upcall immediately upon registration
2. queue the command until the screen becomes ready (what is being done now)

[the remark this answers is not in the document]
dcz-self: I don't envision it as the application needs to busy wait. The HIL mandates that every display on/off call will result in a callback, and that's trivial to forward (and this seems to be already working in my testing).

To prevent the situation where an application registers upcall on a screen that's already turned on, the application should turn it on again. That's also in the HIL, in the same place: every display on/off call shall result in a callback.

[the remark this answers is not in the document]
alexandruradovici: > I don't envision it as the application needs to busy wait. The HIL mandates that every display on/off call will result in a callback, and that's trivial to forward (and this seems to be already working in my testing).
> 
The initialization of the screen is different from the power on. At least the screens that I used need to perform an initialization sequence before they can even be powered on.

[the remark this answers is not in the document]
dcz-self: What's the difference between power on and initialization?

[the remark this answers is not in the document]
alexandruradovici: My understanding from reading the screen's documentation is that power on and off are for powering the screen LCD/TFT, while the initialization sequence is for the screen controller.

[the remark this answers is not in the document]
dcz-self: Is there a case where they need to be done separately? If not, then no problem: they both get started on set_power.

If yes, then set_power could initialize the controller, and set_brightness could power on the panel.

dcz-self (commented):
Trying to un-Pending my comments

capsules/src/screen.rs
dcz-self: Done.

hudson-ayers:
> I've updated this proposal with a consideration regarding when OFF can be returned.
> 
> Along the way I noticed that some atomic application of configuration could be useful in the future, but I don't have hardware that needs it, so not going to try that.
> 
> I think syscall callbacks should be better documented. It seems that only one callback per client can be in flight, and the client must know which, because the callbacks don't carry any info to distinguish them.

Applications can attach data to upcalls which the kernel promises to return unmodified. This allows applications to distinguish distinct callbacks. This is documented here: https://github.com/tock/tock/blob/master/doc/reference/trd104-syscalls.md#42-subscribe-class-id-1

capsules/src/screen.rs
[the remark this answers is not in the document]
alexandruradovici: The display that I am using has a *need to reset* before it can be powered up. This reset is in the hunderds of ms
https://github.com/tock/tock/blob/35cfc73e2b70023a91bbd550b6416366c3ca4911/capsules/src/st77xx.rs#L572-L588

These seem to be necessary to reset the controller and should not for power on and off. If this time is acceptable, we could merge the reset and power on into a single command.

capsules/src/screen.rs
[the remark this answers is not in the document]
alexandruradovici: Which callback are you referring to?

dcz-self:
Is there anything left here that should get improved/redesigned to get a preliminary approval?

alistair23 (dismissed):

hudson-ayers:
The contents look good to me, but I have been deferring to @alexandruradovici as the primary reviewer who actually uses the screen HIL

capsules/src/screen.rs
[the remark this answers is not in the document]
alexandruradovici: The screens that I used do not have the ability to be fully powered off, I think only the TFT is actually powered off, while the internal memory and controller still use power.

There are two ways to go now:
1. include this sequence in the power_on and delete the `screen_is_ready` event, with the observation that the power_on function will take a about a second to complete and the screen's memory will be fully erased;
2. keep this init sequence as it is right now, but then we do need this screen_is_ready function.

@hudson-ayers your feedback would be great

kernel/src/hil/screen.rs
```diff
@@ -90,6 +154,14 @@ impl ScreenPixelFormat {
...
 }
 
+/// Returned for combined operations
+pub enum UpdateType {
+    /// All settings have been applied
+    Complete,
+    /// Not all settings have been applied
+    Partial,
+}
+
```
alexandruradovici, line 164: I am not sure that I fully understand the purpose of this. Can you provide an example?

alexandruradovici:
Everything looks good to me so far, we just need to clear out the screen_is_ready event and I think this is good to go.

kernel/src/hil/screen.rs
```diff
@@ -265,6 +349,6 @@ pub trait ScreenClient {
     /// This is different from `command_complete` as it has to pass back the write buffer
     fn write_complete(&self, buffer: &'static mut [u8], r: Result<(), ErrorCode>);
 
-    /// Some screens need some time to start, this function is called when the screen is ready
-    fn screen_is_ready(&self);
+    /// Some screens need some time to start, this function is called when the screen is ready.
+    fn screen_is_ready(&self, readiness: UpdateType);
```
alexandruradovici, line 353:
```suggestion
```

I think you are right and we can delete this. Applications should issue a power command that should initialize the screen. Any command sent before that should fail with OFF.

I do not fully understand the `UpdateType`.

alexandruradovici (changes requested):

capsules/src/screen.rs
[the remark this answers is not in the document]
alexandruradovici: I had a little bit more time to dig into this and I think @dcz-self you are right, we should drop the `screen_is_ready` event. Using `set_power` to turn off the screen (which may result in resetting the display's framebuffer memory) should be acceptable. Turning off only the display should be done by using `set_brightness` with parameter 0. 

With these changes this should be ready to go.

kernel/src/hil/screen.rs
dcz-self: UpdateType is a consequence of allowing atomic requests. Let's say that the device is able to turn on with a given resolution and color space already configured (for example because the "reset" call has space for them, and setting them after power up takes extra time). The user will want them to be configured as soon as the "ready" signal arrives.

However, we can't unconditionally require this ability from all drivers in the HIL. Another device may initialize in the native resolution no matter what. Then "UpdateType" comes in, allowing the user to know which device was just started.

The first ("atomic") device successfully started and already set the resolution. It returns UptateType::Complete.

The second ("sequential") device started, but ignored the requested resolution. It returns UpdateType::Partial.

I included this atomic mode setting, because it's the same as setting brightness before startup, and presumably costs very little extra complexity.

alexandruradovici: I think I understand somehow your use case. My take in this is that the two things should not be mixed. Powering on the device should just do that and applications should not make any assumption upon resolution and color mode. As soon as the device powers on, applications should request the color mode and resolution, and if required, set them.

I think that adding `UpdateType` adds to much complexity. My suggestion is to keep only one single callback, `command_complete` and state that all commands will fail if the display is power off.

dcz-self: This is the same as setting brightness. You do not want to turn on your device and end up blinding the user for a split second because brightness is detached from power. Similarly, you do not want to have flicker because you're changing resolution or color mode.

Sometimes you can't get away with it, but I think it's a reasonable thing to have foreseen in an API.

I think that setting the brightness before power is a very basic thing and I'm going to have to be convinced to remove it, even if it adds complexity, because of how useful it is for user experience.

alexandruradovici: I do not exactly understand how the `UpdateType` can stop the flicker.

I am not convinced that power on and brightness should be mixed. In my experience with working with low end displays, powering them off means they might loose any configuration that was made. From this point of view, an application needs to assume that when powering down a display means it might loose any configurations. If an application wants to turn off the screen without loosing configuration, it should set the brightness to 0.

Moreover, low end displays are not configurable when powered down (they might actually be powered down).

My view on this is the following: 
- applications should consider that displays need to be powered on before they can be configured or used. Any command issues before that may fail and might not be reflected after power up
- drivers should set the brightness of the screens to 0 after powering on (for the simple displays that I have used this means that they do not actually show anything)
- applications should set a brightness above 0 whenever they need to show content to the user (this might mean that applications can draw, but the user does not see anything unless the brightness is set above 0)

dcz-self: UpdateType does not stop the flicker. You could drop it and have no flicker on supporting displays.

Powered down displays are not configurable when powered down, and they are not meant to. This is not attempting to change that. Instead, it's attempting to allow changing two properties atomically: brightness and power state.

If my display has a hardware functionality: PowerOnWithBrightness(brightness_level), I don't want to go through PowerOnWithBrightness(1) and then SetBrightness(desired_level), because that causes flicker, and takes extra time. Why not use the interface that's available?

From a hardware designer standpoint, the display may have multiple properties at poweron: resolution, brightness, rotation, pixel format, filter, etc. The hardware may initialize any of them to default, or may allow the user to set any combination inside the PowerOn call. From an API designer standpoint, you don't want to have a separate function for each possible combination. One way to avoid it is to send each property to the driver before requesting power-on, and let the driver choose the right call.

To the user, it looks like "configuring" the display before it's on, but it's not.

> drivers should set the brightness of the screens to 0 after powering on

This breaks if the driver doesn't support 0 while powered on, and also doesn't solve flickering.

> applications should set a brightness above 0 whenever they need to show content to the user 

Which is as soon as the display is on.

As for UpdateType, imagine sending SetPowerWithResolution(resolution). On a display that does not support a given resolution, how do you communicate the resulting state of the display?

alexandruradovici: I understand your point of view. I still believe we need to keep system calls as simple as possible, exactly as we do not know what kind of commands a display can actually take. Functions like PowerOnWithResolution would not work exactly for this reason.

You do have a point about not having to resend all the command again. I suggest deferring these actions to the driver in the following manner.

- functions called between power on and power off are executed immediately
- for functions called after power off, the driver should store the requested command and execute it together with the power on command

In this way we do keep commands simple, while avoiding duplicates.

alexandruradovici: I am currently writing a new improved HIL and TRD (#3055) for the display, so I suggest we add a similar functionality there and exclude the `UpdateType` for this PR.

kernel/src/hil/screen.rs
dcz-self: I addressed this in another comment.

capsules/src/screen.rs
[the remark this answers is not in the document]
dcz-self: I think there's a misunderstanding somewhere, because screen_is_ready is still there. I didn't do anything that would remove it, and indeed I rely on it for initialization. As for set_power and set_brightness, this is what I intended the documentation to read. Should I reword it in some way?

capsules/src/screen.rs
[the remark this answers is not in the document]
alexandruradovici: I think we can drop it now as we have established the new way of working for power on, no?

capsules/src/screen.rs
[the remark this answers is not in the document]
alexandruradovici: I think it's ok.

capsules/src/screen.rs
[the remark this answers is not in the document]
dcz-self: Sorry, I don't know what way you mean.

kernel/src/hil/screen.rs
dcz-self: There isn't a separate upcall for partial updates. The upcall carrying UpdateType is the upcall notifying about how power on completed.

alexandruradovici: I agree, I still don't understand the benefit of knowing if the screen has all the properties applied to it as long as it cannot take any commands until it is fully ready.

dcz-self: I can drop it for this MR, except for the brightness.

alexandruradovici: I am sorry, but I still don't understand the usefulness of `UpdateType`. Let's assume we have a display that is powered off. The following steps are run:

- the application issues a set_brightness command, the driver memorizes it and will execute it after the power on command
- the application sends a power on command, the driver powers on the display, reports `UpdateType::Incomplete` and sends the set brightness command to the display
- the application receives the upcall with `UpdateType::Incomplete`. This is where I don't get how does this help the application? Any command that it tries to send to the driver will return `BUSY`
- the display finishes setting the brightness, notifies the driver that sends the `UpdateType::Complete` upcall. Now an application can send commands

dcz-self: I think we can skip UpdateType and atomic mode changes for this MR, for the sake of not holding up uncontroversial improvements. I'll update it in the coming days.

For an immediate explanation:


-    the application issues a set_brightness command, the driver memorizes it and will execute it after the power on command
-    the application sends a power on command, the driver atomically powers on the display at selected brightness
-    the application receives the upcall with UpdateType::Complete

OR 


-    the application issues a set_brightness command, the driver memorizes it and will execute it after the power on command
-    the application sends a power on command, the driver sequentially powers on the display, and then attempts to set brightness but fails
-    the application receives the upcall with UpdateType::Incomplete

alexandruradovici: This makes things more clear now. The issue that I see is that if there are more parameters than brightness, the application has no idea which one failed and there is not much that it can do (maybe power cycle the display?).

Instead of q quick fix now, I'd rather find a more general approach that allows the application to take an action.

alexandruradovici: I'm sorry, but I think your example needs to be more clear, I personally do not understand this from your code example.

Could you please use code that actually compiles, instead of functions like `ignore_off` which do no actually exist. I think it's important to have clear working examples. Of course, you can skip the registering of the callback function.

dcz-self: I've removed this entire train of thought from this proposal. We can consider it when it becomes needed.

kernel/src/hil/screen.rs
```diff
@@ -90,6 +116,14 @@ impl ScreenPixelFormat {
...
 }
 
+/// Returned for combined operations
+pub enum UpdateType {
+    /// All settings have been applied
+    Complete,
+    /// Not all settings have been applied
+    Partial,
+}
+
```
alexandruradovici, line 126: I think you can delete this as it is not used anymore.

alexandruradovici (changes requested):
Remove the `UpdateType` enum and I think we can merge it.

kernel/src/hil/screen.rs
dcz-self: Thanks for noticing that. It's gone.

alexandruradovici (approved):

hudson-ayers (approved):
bors r+

@alexandruradovici thank you for all your time reviewing this and getting it to such a good spot!

PR #3049, opened 2022-05-27
Wrap VolatileCell around Cell; save the world

alevy:
# Pull Request Overview

Uses `Cell` as a wrapper around the value stored in a `VolatileCell`
(and thus inheriting an `UnsafeCell`). This prevents triggering
undefined behavior when sharing a `VolatileCell` immutably when,
internally, it might change.

A consequence is that we have to remove the `Copy` marker, but that was
only used in one place to make initializing a large array easier, so I
changed that call site to simply initialize the array explicitly. More
source code, no more compiled code.

/cc @kupiakos

# Testing Strategy

The changes in this PR are all static refactoring, so it was tested by simply compiling all boards.

# TODO or Help Wanted

# Documentation Updated

-   [ ] <del>Updated the relevant files in `/docs`, or no updates are
    required.</del>

# Formatting

-   [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/usb/descriptors.rs
+++ b/capsules/src/usb/descriptors.rs
@@ -26,7 +26,7 @@ pub struct Buffer64 {
 impl Default for Buffer64 {
     fn default() -> Self {
         Self {
-            buf: [VolatileCell::default(); 64],
+            buf: [(); 64].map(|_| VolatileCell::default()),
         }
     }
 }
--- a/libraries/tock-cells/src/volatile_cell.rs
+++ b/libraries/tock-cells/src/volatile_cell.rs
@@ -1,29 +1,81 @@
 //! Implementation of a type for accessing MCU registers.
 
+use core::cell::UnsafeCell;
 use core::ptr;
 
 /// `VolatileCell` provides a wrapper around unsafe volatile pointer reads
 /// and writes. This is particularly useful for accessing microcontroller
-/// registers.
-// Source: https://github.com/hackndev/zinc/tree/master/volatile_cell
-#[derive(Copy, Clone, Default)]
-#[repr(C)]
-pub struct VolatileCell<T> {
-    value: T,
+/// registers by (unsafely) casting a pointer to the register into a `VolatileCell`.
+///
+/// ```
+/// use tock_cells::volatile_cell::VolatileCell;
+/// let myptr: *const usize = 0xdeadbeef as *const usize;
+/// let myregister: &VolatileCell<usize> = unsafe { core::mem::transmute(myptr) };
+/// ```
+// Originally modified from: https://github.com/hackndev/zinc/tree/master/volatile_cell
+#[derive(Default)]
+#[repr(transparent)]
+pub struct VolatileCell<T: ?Sized + Copy> {
+    value: UnsafeCell<T>,
 }
 
-impl<T> VolatileCell<T> {
+impl<T: Copy> Clone for VolatileCell<T> {
+    #[inline]
+    fn clone(&self) -> Self {
+        VolatileCell::new(self.get())
+    }
+}
+
+impl<T: Copy> VolatileCell<T> {
     pub const fn new(value: T) -> Self {
-        VolatileCell { value }
+        VolatileCell {
+            value: UnsafeCell::new(value),
+        }
     }
 
     #[inline]
-    pub fn get(&self) -> T {
-        unsafe { ptr::read_volatile(&self.value) }
-    }
-
-    #[inline]
+    /// Performs a memory write with the provided value.
+    ///
+    /// # Side-Effects
+    ///
+    /// `set` _always_ performs a memory write on the underlying location. If
+    /// this location is a memory-mapped I/O register, the side-effects of
+    /// performing the write are register-specific.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use tock_cells::volatile_cell::VolatileCell;
+    ///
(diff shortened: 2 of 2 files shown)


kupiakos:
`VolatileCell` should instead be changed to have `UnsafeCell<T>` internally.

libraries/tock-cells/src/volatile_cell.rs
```diff
@@ -1,29 +1,41 @@
...
-        VolatileCell { value }
+        VolatileCell {
+            value: Cell::new(value),
+        }
     }
 
     #[inline]
-    pub fn get(&self) -> T {
-        unsafe { ptr::read_volatile(&self.value) }
+    pub fn set(&self, value: T) {
```
kupiakos, line 31: This won't call the destructor for the previous value, but there's no way to do that without a read, defeating the point. Can we make `set` require `T: Copy` to reduce the likelihood of problems due to this, and mention this drop behavior in a doc comment?

capsules/src/usb/descriptors.rs
```diff
@@ -26,7 +26,72 @@ pub struct Buffer64 {
 impl Default for Buffer64 {
     fn default() -> Self {
         Self {
-            buf: [VolatileCell::default(); 64],
+            buf: [
```
kupiakos, line 29: This is simpler: `[(); 64].map(|_| VolatileCell::default())`

libraries/tock-cells/src/volatile_cell.rs
```diff
@@ -1,29 +1,41 @@
...
-// Source: https://github.com/hackndev/zinc/tree/master/volatile_cell
-#[derive(Copy, Clone, Default)]
-#[repr(C)]
-pub struct VolatileCell<T> {
-    value: T,
+// Originally modified from: https://github.com/hackndev/zinc/tree/master/volatile_cell
+#[derive(Default)]
+#[repr(transparent)]
+pub struct VolatileCell<T: ?Sized> {
+    value: Cell<T>,
```
kupiakos, line 13: nit: We don't use `Cell`'s ops here, is there a reason it was chosen over `UnsafeCell`?

kupiakos (dismissed):
Just some details; otherwise looks great, thank you!

hudson-ayers (dismissed):

libraries/tock-cells/src/volatile_cell.rs
alevy: We can definitely make restrict all of `volatile_cell` to `Copy` types (for MMIO this is almost certainly desirable anyway). But can you explain why this would reduce the likelihood of problems? Is it because it's uncommon for Copy types to have meaningful drop behavior?

(Also, yes, definitely will include this issue in the documentation, great catch!)

libraries/tock-cells/src/volatile_cell.rs
alevy: Because I thought it saved us some custom implementations and unsafes but it doesn't. My bad :/

libraries/tock-cells/src/volatile_cell.rs
alevy: I've fixed this, but I'm still curious to clarify why requiring `Copy` (which makes sense anyway) helps with the drop issue

kupiakos: > Is it because it's uncommon for Copy types to have meaningful drop behavior?

Yes. They're very rare, and when they do exist, their destructors aren't doing any pointer freeing, at least soundly.

jrvanwhy: > > Is it because it's uncommon for Copy types to have meaningful drop behavior?
> 
> Yes. They're very rare, and when they do exist, their destructors aren't doing any pointer freeing, at least soundly.

Not just very rare, nonexistent. No type can implement both `Copy` and `Drop`.

libraries/tock-cells/src/volatile_cell.rs
```diff
@@ -1,29 +1,81 @@
...
-    pub fn set(&self, value: T) {
-        unsafe { ptr::write_volatile(&self.value as *const T as *mut T, value) }
+    /// Performs a memory read and returns a copy of the value represented by
+    /// the cell.
+    ///
+    /// # Side-Effects
+    ///
+    /// `get` _always_ performs a memory read on the underlying location. If
+    /// this location is a memory-mapped I/O register, the side-effects of
+    /// performing the write are register-specific.
```
hudson-ayers, line 68:
```suggestion
    /// performing the read are register-specific.
```

hudson-ayers (approved):
Just one typo!

libraries/tock-cells/src/volatile_cell.rs
```diff
@@ -1,29 +1,81 @@
...
-        unsafe { ptr::read_volatile(&self.value) }
+    /// Performs a memory write with the provided value.
+    ///
+    /// # Side-Effects
+    ///
+    /// `set` _always_ performs a memory write on the underlying location. If
+    /// this location is a memory-mapped I/O register, the side-effects of
+    /// performing the write are register-specific.
+    ///
+    /// `set` does not read the value currently inside the `VolatileCell` and,
```
kupiakos, line 45: nit: With the `T: Copy` restriction, this line isn't necessary, at least in the doc string. Maybe instead have a comment inside the function that mentions that because `T: Copy`, there's no drop to consider?

kupiakos (approved):

PR #3052, opened 2022-06-05
doc/reference/trd-uart: small fixes & suggestions in response to #3046

lschuermann:
### Pull Request Overview

This pull request contains some small fixes, suggestions and additions to the UART draft TRD I noticed while implementing it for the `sifive` chips as part of #3046, namely:

- some indetation, whitespace, spelling fixes, renaming `{set,get}_flow_control` to `{set,get}_hw_flow_control` to be consistent with the `Parameters` struct field.

- [disable bulk transfers with width >= 9](https://github.com/tock/tock/commit/8327c3b91a7d118f968d5f81c8050b3509deca51) 

  The TRD mentions that `transmit_buffer` and `receive_buffer` are not supported for transfers with data-width configured to larger than 8-bit, but does not explicitly document what happens when these methods are invoked in this configuration. This changes the text to enforce these methods must return `ErrorCode::INVAL` in such a configuration.

  It furthermore extends section 3 (`Configuration` and `Configure`) to describe both the case of Width < 8-bit and Width > 8-bit.

- [apply config changes only to new operations](https://github.com/tock/tock/commit/03207fdb2eaace9f1e7b951e574c8dfc129141de) 

  It does not seem like a good idea to allow UART configuration changes mid-flight, and conceivably some hardware might not support this at all (e.g. fix parameters on the start of the DMA operation or raise an exception when changing them mid-flight).

  To unify the behavior of different implementations, it seems most sensible to define that configuration changes apply only to new operations and a device is free to refuse changing its configuration when an operation is currently ongoing.

- [ensure configuration changes are atomic](https://github.com/tock/tock/commit/516f93d79f51f32c9d56daae57506f0e87040c5e) 

  The utility of `Configure::configure` is quite limited currently, considering that other methods in the `Configure` allow setting all parameters contained in the passed `Parameters` struct already. What would make it significantly more useful is to have all changes applied through that interface be atomic: if only a single requested parameter is not supported by the underlying device, do not apply any of the passed parameters. One particularly convincing usecase for this might be a parameter change following a device-negotation: either the negotiated parameters can be used, or communication can continue in the original configuration.

- [return an error on `{tx,rx}_len == 0`](https://github.com/tock/tock/commit/7cf4ca2fc22981234d3b4d30ac3f08d00b04720b) 

  If transmissions or receptions of zero length must be supported as proper operations, they will need to result in a callback which must be generated as part of a different call stack. If a UART device does not support generating hardware interrupts for the completion of zero-length transmissions, a device driver implementation has to use a deferred call mechanism in order to be compliant with the API specified in this HIL.

  To avoid employing a deferred call for this edge case when it would otherwise not be required, avoid supporting zero-length operations.

### Testing Strategy

N/A

### TODO or Help Wanted

This pull request still needs feedback on whether these are actually changes we'd like to include.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/doc/reference/trd-uart.md
+++ b/doc/reference/trd-uart.md
@@ -87,7 +87,14 @@ still partitioned into bytes, and the UART sends the least significant
 bits of each byte. Suppose a UART is configured to send 7-bit
 words. If a client sends 5 bytes, the UART will send 35 bits,
 transmitting the bottom 7 bits of each byte. The most significant bit
-of each byte is ignored.
+of each byte is ignored. While this HIL does support UART transfers
+with a character-width of more than 8-bit, such characters cannot be
+sent or received using the provided bulk transfer mechanisms. A
+configuration with `Width` > `8` will disable the bulk buffer transfer
+mechanisms and restrict the device to single-character
+operations. Refer to [3 `Transmit` and `TransmitClient`
+](#3-transmit-and-transmitclient) and [4 `Receive` and `ReceiveClient`
+](#4-receive-and-receiveclient-traits) respectively.
 
 ```rust
 pub enum StopBits {
@@ -216,7 +223,9 @@ of the bytes passed in the buffer. For example, if the UART is using
 
 If a client needs to transmit characters larger than 8 bits, it should
 use `transmit_character`, as `transmit_buffer` is a buffer of 8-bit
-bytes and cannot store 9-bit values.
+bytes and cannot store 9-bit values. If the UART is configured to use
+characters wider than 8-bit, the `transmit_buffer` operation is
+disabled and calls to it must return `ErrorCode::INVAL`.
 
 There can be a single transmit operation ongoing at any
 time. Successfully calling either `transmit_buffer` or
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):
These all seem positive to me, thanks for making a pass.

I am not 100% sure that it is always the right decision to require implementations to detect mismatch between configuration and use of the API (this makes the implementation larger, and it does not seem unreasonable for the implementation to instead just work incorrectly if it is misused), but don't feel strongly enough to vote the other way.

lschuermann:
> I am not 100% sure that it is always the right decision to require implementations to detect mismatch between configuration and use of the API (this makes the implementation larger, and it does not seem unreasonable for the implementation to instead just work incorrectly if it is misused)

I spent some more time thinking about the consequences of either approach. It'd surely work to leave certain parts of the interface unspecified if these should not be invoked given the current configuration, but let's assume that one of these unspecified methods is used nonetheless, for the sake of an example calling `transmitt_buffer` with `Width >= 9`: I can imagine implementations refusing this method call, returning a `FAIL`. This is hard to handle in upper layers. Other implementations might actually support this in hardware, working on 16-bit words and producing garbled output data if the expectation is to only output 8-bit words using that function. Yet other implementations may output 8-bit words when the upper layer would've expected it to work on 16-bit words. And some may just break, because why handle an unspecified code path?

In the end, what convinced me was the question as to what would be considered implementation defined behavior, and what would be a bug? I can't think of a good line to draw, so just specifying that an invocation of a method with an unmatching configuration must return an error seemed like the easy way out. It increases comaptibility by narrowing down what a compliant interface looks like.

alistair23 (approved):

phil-levis (approved):

PR #3055, opened 2022-06-10
Unify text and graphics display and TRD

alexandruradovici:
### Pull Request Overview

After discussing the modifications for the display HIL and drivers (#3041 and #3047) with @dcz-self, I tried to unify the text and graphics display HILs under the Display HIL.

In a nutshell, changes are:
- single `Screen` trait that handle brightness, power and other hardware related tings
- a `FrameBuffer` trait
- a `TextBuffer` trait

I started writing a TRD using @phil-levis SPI as an inspiration.

### Testing Strategy

This pull request was not tested.

### TODO or Help Wanted

Feedback

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.


kernel/src/hil/display.rs
```diff
@@ -0,0 +1,376 @@
...
+    ASCII,
+    /// Characters are encoded using UTF8, used for monochromatic displays
+    UTF8,
+    /// Characters are encoded using 16 bits, as in the VGA text mode
+    /// 0   1   2   3   4   5   6   7   8   9   10  11  12  13  14  15  16
+    /// +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
+    /// | B |Background | Foreground    |          Code Point           |
+    /// +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
+    /// B - blink
+    VGA,
```
dcz-self, line 107: What encoding in VGA? What happens if the code point is multiple bytes? What endianness?

kernel/src/hil/display.rs
```diff
@@ -0,0 +1,376 @@
...
+    fn get_mode(&self) -> GraphicsMode;
+
+    /// Returns the minim size (in bytes) that can be written to the framebuffer.
+    ///
+    /// Due to hardware constraints, framebuffer writes have to be rounded up to a tile size.
+    /// This means that the size of the write buffer has to be a multiple of the tile. An
+    /// example use case is a framebuffer that has the minimum write unit a full line on
+    /// the display. This means that clients can only write entier lines to the framebuffer
+    /// as opossed to single pixels.
+    fn get_block_size(&self) -> usize;
```
dcz-self, line 208: There are at least 4 coordinates needed for tiling: xoffset, yoffset, xsize, ysize. The tiles also need to get internal pixel ordering: some display drivers (PCD8544) will iterate rows within a tile first, and then columns.

kernel/src/hil/display.rs
```diff
@@ -0,0 +1,376 @@
...
+    /// from the screen upfront.
+    fn get_num_supported_modes(&self) -> usize;
+
+    /// Can be called with an index from 0 .. count-1 and will
+    /// a tuple (width, height) with the current resolution (in pixels).
+    /// note that width and height may change due to rotation
+    ///
+    /// This function is synchronous as the driver should know this value without
+    /// requesting it from the screen.
+    fn get_supported_mode(&self, index: usize) -> Option<GraphicsMode>;
```
dcz-self, line 291: Why not return a slice?

kernel/src/hil/display.rs
```diff
@@ -0,0 +1,376 @@
...
+    /// - 0 - completely no light emitted
+    /// - 1..MAX_BRIGHTNESS - on, set brightness to the given level
+    ///
+    /// The display should interpret the brightness value as *lightness*
+    /// (each increment should change preceived brightness the same).
+    /// 1 shall be the minimum supported brightness,
+    /// `MAX_BRIGHTNESS` and greater represent the maximum.
+    /// Values in between should approximate the intermediate values;
+    /// minimum and maximum included (e.g. when there is only 1 level).
+    fn set_brightness(&self, brightness: usize) -> Result<(), ErrorCode>;
```
dcz-self, line 346: I don't think brightness should be in the same trait as inversion. One is very important, the other very unimportant.

dcz-self (commented):
I like the idea to generalize displays: the roations, brightness, and inversions are common to any kind of a display.

I apprecate incorporating some more high level feedback on my merge requests.

From my point of view, however, the graphical display HIL still has a couple fundamental problems, and I think we have enough on our plates. This doesn't really help me resolve any problems I've seen on my hardware. I'll only review this again after we reach the end of the improvements I proposed here https://github.com/tock/tock/pull/3047 .

kernel/src/hil/display.rs
alexandruradovici: This was modeled initially to be sent over the syscall ABI which did no support a slice. You might be right, and the driver may be able to have an array with these modes and return a slice to it, I'll think about it.

kernel/src/hil/display.rs
alexandruradovici: I agree they might not be in the best place. Where would you see the invert functions?

alexandruradovici:
> I appreciate incorporating some more high level feedback on my merge requests.
>
I'm not sure what do you mean by this.
 
> From my point of view, however, the graphical display HIL still has a couple fundamental problems, and I think we have enough on our plates. This doesn't really help me resolve any problems I've seen on my hardware. I'll only review this again after we reach the end of the improvements I proposed here #3047 .

I think it is very important to have a clear TRD before we actually start implementing things. As you have several issues, I do believe that this is the perfect timing to establish this.

dcz-self:
I'm happy that you tried to include some suggestions from me.

> I think it is very important to have a clear TRD before we actually start implementing things. As you have several issues, I do believe that this is the perfect timing to establish this.

We have one: https://github.com/tock/tock/pull/3011

I consider it an exemplification of why the approach fails:
- it's big, and it's been in review for 2 months without coming anywhere close to being accepted
- it's already incomplete, because https://github.com/tock/tock/pull/3047 is much narrower in scope and yet includes some basic improvements over it
- things get figured out while implementing, and implementations are easier to do stepwise
- small improvements are valuable to me now, because large reworks may arrive after my project has already failed, or after I've moved on

Considering all that, this draft doesn't present all that much value to my project.

alexandruradovici:
> I'm happy that you tried to include some suggestions from me.
>
I tried to include the whole discussion that we had for #3047, I do have to include the things that we have discussed today.

> We have one: #3011
> 
I think this takes in consideration most of the things in your PR, I might have missed some though.

> Considering all that, this draft doesn't present all that much value to my project.
The main purpose of TRDs is to explain how things work. We need to understand the Tock user's needs to be able to provide the closest solution that works for their projects, but that still keeps Tock general enough. This is why your feedback on this would be very valuable.

dcz-self:
I don't think I can provide good feedback without having an implementation. For feedback that doesn't need an implementation, I can't communicate clearly when it's not split into small units. Therefore I'll provide feedback in reworks that have a small scope.

alexandruradovici:
Great, I suggest we start with the tiles 😄 function.

dcz-self:
I suggest we finish https://github.com/tock/tock/pull/3047 before opening another can of worms :)

dcz-self:
I wrote up a proposal for tiles: https://github.com/tock/tock/pull/3067

bradjc:
Can one of you please open a tracking issue for the screen/display stack? It would be helpful to have a rough list of todos and a roadmap for what the screen/display stack should support.

dcz-self:
Here's the tracking issue: https://github.com/tock/tock/issues/3079

doc/reference/trd-display.md
```diff
@@ -0,0 +1,185 @@
...
+
+The Display HIL defines three main items that are present in all the displays:
+
+  1. The `FrameBuffer` that handles pixel data that is being displayed, used for graphics
+  2. The `TextBuffer` that handles text data that is being displayed, used for text
+  3. The `Screen` that handles the parameters of the actual display hardware
+
+The separation between framebuffer, textbuffer and screen has been made as they may function
+as independant systems. For instance, the framebuffer could be used either for a
+screen or fopr a virtual framebuffer, while the screen could be used either for
```
bradjc, line 35:
```suggestion
screen or for a virtual framebuffer, while the screen could be used either for
```

kernel/src/hil/display.rs
```diff
@@ -0,0 +1,376 @@
...
+
+#[derive(Copy, Clone, PartialEq)]
+pub struct TextMode {
+    columns: usize,
+    lines: usize,
+    character_format: CharacterFormat,
+}
+
+/* TextBuffer */
+pub trait TextBuffer<'a> {
```
bradjc, line 125: Some description of what this and framebuffer are would be helpful to include here. I think what is missing in the name is that buffer implies this would be a synchronous interface (just writing to the buffer). Why not `TextScreen`, which includes everything but print and set_cursor?

kernel/src/hil/display.rs
```diff
@@ -0,0 +1,376 @@
...
+    fn get_size(&self) -> (usize, usize);
+
+    /// Sends a write command to the driver, and the buffer to write from
+    /// and the len are sent as arguments. When the `write` operation is
+    /// finished, the driver will call the `write_complete()` callback.
+    ///
+    /// Return values:
+    /// - `Ok(())`: The write command is valid and will be sent to the driver.
+    /// - `BUSY`: The driver is busy with another command.
+    fn print(
```
brghena, line 142: What about text wrapping? Is that something that the device often/sometimes supports, or does it need to be implemented above the HIL?

kernel/src/hil/display.rs
dcz-self: In https://github.com/tock/tock/pull/3011/ I took them out into a separate trait with extra builtin functionality, together with setting rotation. Either way, it should not share the trait with power and brightness control, because those are expected out of almost all displays.

kernel/src/hil/display.rs
```diff
@@ -0,0 +1,376 @@
...
+    /// Some framebuffers keep in a temporary memory the changes and require a flush command
+    /// to send the changes to the hardware.
+    ///
+    /// Return values:
+    /// - `Ok(())`: Flush is in progress and the client will receive
+    ///    a call to `command_complete`.
+    /// - `ENOSUPPORT`: Flush has been done synchronous or there is no
+    ///    no need to flush the framebuffer.
+    /// - `BUSY`: Another write or flush is in progress.
+    fn flush(&self) -> Result<(), ErrorCode>;
```
dcz-self, line 251: With https://github.com/tock/tock/pull/3067 I will no longer have any need for flushing.

kernel/src/hil/display.rs
dcz-self: Thought through in https://github.com/tock/tock/pull/3067

dcz-self (changes requested):
I somewhat like the idea of having a separate trait to control power, brightness, and filters like rotation and inversion. This will reduce redundancy in HILs, even if it will not result in reuse of any actual code.

I don't need to use a character display, so I'm not going to comment on that.

Once the graphical display side is updated to match the current Screen trait, and once it includes changes from https://github.com/tock/tock/pull/3067, as well as splits off the rotation/inversion trait, then I'm for it.

bradjc:
I would like to see this updated and merged to make progress on #3079

alevy:
Screen APIs have moved on. Closing for now.

PR #3060, opened 2022-06-21
boards/opentitan: Don't run nops when targetting Verilator

alistair23:
### Pull Request Overview

The `NOP` flush in the panic handler takes a **really** long time in a simulation.

This PR changes the `NOP` flush to be optional and then doesn't use it when running on the OpenTitan Verilator build.

### Testing Strategy

Running on Verilator

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/src/io.rs
+++ b/boards/opentitan/src/io.rs
@@ -48,9 +48,20 @@ pub unsafe extern "C" fn panic_fmt(pi: &PanicInfo) -> ! {
     );
     first_led_pin.make_output();
     let first_led = &mut led::LedLow::new(first_led_pin);
-
     let writer = &mut WRITER;
 
+    #[cfg(feature = "sim_verilator")]
+    debug::panic(
+        &mut [first_led],
+        writer,
+        pi,
+        &|| {},
+        &PROCESSES,
+        &CHIP,
+        &PROCESS_PRINTER,
+    );
+
+    #[cfg(not(feature = "sim_verilator"))]
     debug::panic(
         &mut [first_led],
         writer,
@@ -59,7 +70,7 @@ pub unsafe extern "C" fn panic_fmt(pi: &PanicInfo) -> ! {
         &PROCESSES,
         &CHIP,
         &PROCESS_PRINTER,
-    )
+    );
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


thulithwilfred:
Nice, is there any potential from skipping `NOP` flush to miss out on any panic logs or things the kernel wasn't finished with  on Verilator? Noticed here it says https://github.com/tock/tock/blob/4aa174d4fa695a0f1cb1b8f207b856a71e6043c5/kernel/src/debug.rs#L149 'Let any outstanding uart DMA's finish`.

alistair23:
I don't think so. The exact comment you point out doesn't apply to OpenTitan as it doesn't have DMA. We also aren't interfacing with any external hardware that we have to wait on so I don't think this should have any downsides on simulation platforms.

Besides UART DMA it probably isn't useful for boards anyway, it just doesn't really cost us much on full speed silicon. But it's very slow in a simulator

thulithwilfred:
> 

Ah gotcha, that makes sense

thulithwilfred (dismissed):

bradjc:
What takes a long time? Running a `nop`? Or calling the `nop()` function? Or calling any function 20000 times?

In any case, I don't think `nop` should be an optional argument, instead there should be a new API in debug.rs that a simulated platform can use (if it wants).

hudson-ayers:
> What takes a long time? Running a `nop`? Or calling the `nop()` function? Or calling any function 20000 times?

The actual call to nop() should be inlined, the delay is just that actually running 20,000 instructions on a simulated CPU takes awhile, and we do that multiple times for long panic prints.

> In any case, I don't think `nop` should be an optional argument, instead there should be a new API in debug.rs that a simulated platform can use (if it wants).

I think the verilated platform should just pass a reference to an empty closure rather than modify the `nop()` function:

```rust
    debug::panic(
        &mut [first_led],
        writer,
        pi,
        &|| {},
        &PROCESSES,
        &CHIP,
        &PROCESS_PRINTER,
    )

```
The above works just fine and should achieve the same effect.

alistair23:
Ah, that's even better! Thanks

thulithwilfred (approved):

hudson-ayers (approved):

bradjc (approved):

PR #3066, opened 2022-06-27
lpm013m126: Implement turning off

dcz-self:
Not turning off was technically against the contract of the screen HIL.

### Pull Request Overview

This pull request fixes the lpm display driver by adding turn off functionality.

### Testing Strategy

This pull request was tested by having it run in a private branch for a while (https://framagit.org/jazda/tock), from where it was manually picked.

It still neglects to clear the pixel memory, but that hasn't been necessary in my testing.

### TODO or Help Wanted

This pull request still needs a sanity check.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/lpm013m126.rs
+++ b/capsules/src/lpm013m126.rs
@@ -339,6 +339,33 @@ where
         }
     }
 
+    fn uninitialize(&self) -> Result<(), ErrorCode> {
+        match self.state.get() {
+            State::Off => Err(ErrorCode::ALREADY),
+            _ => {
+                // TODO: investigate clearing pixels asynchronously,
+                // like the datasheet asks.
+                // It seems to turn off fine without clearing, but
+                // perhaps the state of the buffer affects power draw when off.
+
+                // The following stops extcomin timer.
+                self.alarm.disarm()?;
+                self.disp.clear();
+                self.state.set(State::Off);
+
+                let scheduled =
+                    schedule_deferred(self.deferred_caller, &self.ready_callback, "ready");
+
+                if let Err(()) = scheduled {
+                    self.state.set(State::Bug);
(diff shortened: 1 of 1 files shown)


capsules/src/lpm013m126.rs
```diff
@@ -339,6 +339,29 @@ where
         }
     }
 
+    fn uninitialize(&self) -> Result<(), ErrorCode> {
+        match self.state.get() {
+            State::Off => Err(ErrorCode::ALREADY),
+            _ => {
+                // TODO: clear pixels asynchronously, like the datasheet asks
+                // this stops extcomin timer
```
alexandruradovici, line 347: I assume this is still work in progress, no? If this should be merged without the solved TODO, can you describe here the implications of not implementing this yet.

dcz-self: This is not work in progress. It works as expected. I updated the comment.

alexandruradovici: My assumption is that it works, but the display's datasheet suggests clearing all pixels? What about changing the `TODO` into `Improvement`?

dcz-self: I'm not sure if it will actually be any improvement :P But I know it's something to investigate, and TODO reflects that well I think. Plus, convention says we grep code for "TODO", "FIXME", and "XXX" for rough edges.

Shall I still change it to "improvement"?

alistair23 (dismissed):

hudson-ayers (dismissed):

capsules/src/lpm013m126.rs
dcz-self: It's now "TODO: investigate". I also added an empty line to separate the two comments.

hudson-ayers (approved):

PR #3067, opened 2022-06-28
hil: Add support for tiled displays

dcz-self:
### Pull Request Overview

This pull request adds support for tiled displays to the screen HIL. "Tiled" as discussed in https://github.com/tock/tock/pull/3011 . The lpm013m126 is a tiled display, because the smallest unit that it can accept is a whole row of pixels. Currently, it's forced to hold a frame buffer, but after this change, it won't.

### Testing Strategy

This pull request was not tested or verified. lpm013m126 is in plans.

### TODO or Help Wanted

This pull request still needs verification and changes to the syscall driver.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/hil/screen.rs
+++ b/kernel/src/hil/screen.rs
@@ -34,6 +34,7 @@ use core::ops::Sub;
 
 pub const MAX_BRIGHTNESS: usize = 65536;
 
+/// Rotation in degrees counterclockwise in relation to `Normal`.
 #[derive(Copy, Clone, PartialEq)]
 pub enum ScreenRotation {
     Normal,
@@ -90,12 +91,28 @@ impl Sub for ScreenRotation {
     }
 }
 
+/// Defines the data format being fed to the display driver.
+/// Each pixel can take up more or less than a byte.
+/// Reaching the end of line in a defined region moves the cursor
+/// back to the start of the line, and downwards by 1 line
+/// (unless noted otherwise).
 #[derive(Copy, Clone, PartialEq)]
 #[repr(usize)]
 #[allow(non_camel_case_types)]
-pub enum ScreenPixelFormat {
-    /// Pixels encoded as 1-bit, used for monochromatic displays
-    Mono,
+pub enum PixelStreamFormat {
+    /// Pixels encoded as 1-bit, 8 consecutive bits (most to least significant)
+    /// form 8 horizontal pixels left-to-right.
+    Mono_1H8,
+    /// Pixels encoded as 1-bit, 8 consecutive bits (most to least significant)
+    /// form 8 vertical pixels top-to-bottom.
+    /// Reaching end of line advances the cursor to the beginning of the line
+    /// and 8 pixels downwards.
+    Mono_1V8,
+    /// A single byte carries two pixels horizontally: higher nibble on the left,
+    /// lower on the right.
+    /// Each color channel is 1 bit (starting with most significant): RGBX,
+    /// where X is a padding bit.
+    RGB_111xH2,
     /// Pixels encoded as 2-bit red channel, 3-bit green channel, 3-bit blue channel.
     RGB_233,
     /// Pixels encoded as 5-bit red channel, 6-bit green channel, 5-bit blue channel.
@@ -107,16 +124,36 @@ pub enum ScreenPixelFormat {
     // other pixel formats may be defined.
 }
 
-impl ScreenPixelFormat {
+impl PixelStreamFormat {
     pub fn get_bits_per_pixel(&self) -> usize {
         match self {
-            Self::Mono => 1,
+            Self::Mono_1H8 => 1,
+            Self::Mono_1V8 => 1,
+            Self::RGB_111xH2 => 4,
             Self::RGB_233 => 8,
             Self::RGB_565 => 16,
             Self::RGB_888 => 24,
             Self::ARGB_8888 => 32,
         }
     }
+}
+
+/// Placement of a grid of rectangles on the display.
+pub struct Grid {
+    /// Placement of a vertical edge of a rectangle
+    /// relative to the left edge of the display.
+    /// Additional rectangles are tiled to the left and right
+    /// without overlapping, until they fill the entire display area.
+    /// Greater values are further to the right.
+    pub x_offset: u16,
+    /// Placement of a horizontal edge of a rectangle,
+    /// relative to top edge of the display.
+    /// Greater values are further downwards.
+    pub y_offset: u16,
+    /// Width of a single rectangle.
+    pub width: u16,
+    /// Height of a single rectangle.
+    pub height: u16,
 }
 
 pub trait ScreenSetup {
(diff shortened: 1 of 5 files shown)


kernel/src/hil/screen.rs
```diff
@@ -186,18 +224,41 @@ pub trait Screen {
...
+    /// described by the return value is filled with a pixel buffer,
+    /// then the buffer ends on a byte boundary.
+    /// For example, if the format is `Mono_1H8` (8 pixels per byte),
+    /// then `Grid::width` cannot be 12 (12 pixels = 12 bits = 1,5 bytes).
+    ///
     /// This function is synchronous as the driver should know this value without
-    /// requesting it from the screen.
-    fn get_pixel_format(&self) -> ScreenPixelFormat;
+    /// requesting it from the hardware.
+    fn get_pixel_format(&self) -> (PixelStreamFormat, Grid);
```
dcz-self, line 245: I'm not exactly sure where the user should get that info. Grid might change with resolution, perhaps rotation, but returning a new Grid on every call is a bit silly.

kernel/src/hil/screen.rs
```diff
@@ -186,18 +224,41 @@ pub trait Screen {
...
 
     /// Returns the current rotation.
+    ///
+    /// The rotation is relative to a touch device, if present.
+    ///
```
dcz-self, line 250: I couldn't help myself mess with the rotation. Should I split it off?

alistair23:
@alexandruradovici

kernel/src/hil/screen.rs
```diff
@@ -116,6 +135,22 @@ impl ScreenPixelFormat {
...
+/// Placement of a grid of rectangles on the display.
+pub struct Grid {
+    /// Placement of a vertical edge of a rectangle
+    /// relative to the left edge of the display.
+    /// Additional rectangles are tiled to the left and right
+    /// without overlapping, until they fill the entire display area.
+    /// Greater values are further to the right.
+    pub u16: x_offset,
+    /// Greater values are further downwards.
+    pub u16: y_offset,
```
alexandruradovici, line 147: I understand `width` and `height`, but I don't quite get the meaning of the offsets.

dcz-self: I can imagine screens where the dimension is not a whole multiple of the tile size, putting the *end* at a different offset. From there, it's not a big jump to assume an odd start, perhaps due to rotation. I have no examples though.

kernel/src/hil/screen.rs
dcz-self: My goal here is to push the Grid information on the user, so that they don't have an opportunity not to notice it. After all, there's no way to operate the display without knowing all of format, resolution, and grid.

The idea to put it in get_pixel_format return value came from the fact that for a screen with only one supported format, get_pixel_format *must* be called before knowing how to submit data. (So does get_resolution, but I decided by a coin flip.)

I would like to preserve that property: shove the Grid in user's face. Does this make more sense now?

The situation gets more complicated in advanced displays where the user selects the pixel format. There, the user can change the pixel format directly, wihout a call to get_pixel_format. The path to push the Grid on to the user is broken. But I think that might have a solution:

```
fn set_pixel_format(Format) -> Result<Grid, Error>
```

(I think it's enough to put it in the return value rather than a callback, the driver will probably know how to calculate the new grid.) This way, the user will see the grid before submitting data.

Do this sound like a good idea?

kernel/src/hil/screen.rs
alexandruradovici: I would assume that the end would have some ignored pixels. Most probably the memory space would be there (as usually tiles are related to memory units), but the pixels are missing. I would suggest having only `width` and `height`.

kernel/src/hil/screen.rs
alexandruradovici: I think the `set_pixel_format` needs to be asynchronous, as many displays will have to ask this.

dcz-self: Do you mean get_pixel_format? Are there any disadvantages of making the query at initialization?

kernel/src/hil/screen.rs
dcz-self: What should be done for displays rotated 180°?

kernel/src/hil/screen.rs
```diff
@@ -212,7 +273,16 @@ pub trait Screen {
...
+    /// Writes data into the selected rectangle of pixels.
+    ///
+    /// By default, the selected rectangle covers the entire frame.
+    /// The rectangle can be set using `set_write_frame`.
+    ///
+    /// The pixel data contained in the buffer must adhere
+    /// to the current pixel format and grid shape. See `get_pixel_format`.
+    /// Consecutive pixels on the screen will be set
+    /// to consecutive values from the buffer. See `PixelStreamFormat`.
+    ///
```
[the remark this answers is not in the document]
phil-levis: Generally, you should operate on a frame buffer, then apply a rotation after. Otherwise, you have to be continually transforming. I.e., always operate in the framebuffer frame of reference, not the screen frame of reference.

kernel/src/hil/screen.rs
phil-levis: > Would there ever be a screen where the tiling didn't start at 0,0?

Yeah, I've seen this done. It happens when you want the screen to be a window into a larger framebuffer. E.g., you could do something where you load a framebuffer with a 1000x1000 image, the screen is 200x200, and you want to tile a 100x100 patch of the framebuffer 4 times into the screen.

phil-levis: > What should be done for displays rotated 180°?

I think you want to operate in framebuffer space. Suppose I have a device that can detect orientation and wants to adjust the screen rotation accordingly (e.g., a modern phone). When the device rotates 180 degrees, I continue to draw up as up (the top left of the image is at 0, 0) but then I physically rotate it 180 degrees on the screen. You would also need to do geometric inversion for other inputs (e.g., a capacitive display, accelerometers), but keying this on a global rotation bit is easier than trying to dynamically switch draw calls.

kernel/src/hil/screen.rs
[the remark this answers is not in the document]
dcz-self: Should that interpretation find itself into this change? I deliberately ignored it to limit the scope of this, and we have more semantics questions (around resolution, format, inversion) to sort out.

alexandruradovici:
I'll try to integrate this feedback into #3055.

dcz-self:
Updated: rebased, and made sure that it works (except old bugs in the lpm... driver that I'll fix in later patches).

I put this on top of the new PR: https://github.com/tock/tock/pull/3177 because the lpm... driver is broken and noncompliant to the HIL without it.

dcz-self:
Added userspace counterpart with an example: https://github.com/tock/libtock-rs/pull/429

alexandruradovici:
I suggest we fit this into #3055 which has a broader approach for displays.

dcz-self:
Fixing this doesn't stop #3055 from progressing. This is more or less done, so if we wait for #3055, we're just going to block further development until all of #3055 is solved.
I have improvements to the lpm013 driver already waiting, and I need tile support but don't need unification.

alexandruradovici:
I guess we can accept this as a temporary solution until it is merged in a similar form in #3055

alistair23 (approved):

bradjc:

I think changes like this need someone to reconcile a consistent view of the screen with PRs like #3811 and issue #3079.

PR #3069, opened 2022-07-02
boards/redboard_artemis_nano: Setup initial tests

alistair23:
### Pull Request Overview

Setup the Rust tests for the Apollo3 board

### Testing Strategy

Run the tests on the board

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,6 @@ boards/swervolf/fusesoc_libraries/
 
 # ESP C3 temp hex
 boards/esp32-c3-devkitM-1/binary.hex
+
+# Redboard Artemis Nano test binary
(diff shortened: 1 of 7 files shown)


hudson-ayers (approved):
I think its great to have this simple approach to testing more boards. We should eventually move the multi-alarm test into `capsules/` in some non-board specific way rather than duplicating the file for every board.

bradjc (approved):
I assume if we have to remove `#![feature(custom_test_frameworks)]` from one board someday it isn't much more difficult to remove it from two.

alistair23:
Ping!

PR #3070, opened 2022-07-02
kernel: hil: sensors: Add an AirQualityDriver

alistair23:
### Pull Request Overview

Add a AirQualityDriver HIL. This can be used to measure things like CO2
or TVOC levels from a sensor to reflect the current air qualtiy.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/kernel/src/hil/sensors.rs
+++ b/kernel/src/hil/sensors.rs
@@ -31,6 +31,47 @@ pub trait HumidityClient {
     fn callback(&self, value: usize);
 }
 
+/// A basic interface for a Air Quality sensor
+pub trait AirQualityDriver<'a> {
+    /// Set the client to be notified when the capsule has data ready.
+    fn set_client(&self, client: &'a dyn AirQualityClient);
+
+    /// Read the CO2 or equivalent CO2 (eCO2) from the sensor.
+    /// This will trigger the `AirQualityClient` `co2_data_available()`
+    /// callback when the data is ready.
+    ///
+    /// This function might return the following errors:
+    /// - `BUSY`: Indicates that the hardware is busy with an existing
+    ///           operation or initilisation/calibration.
+    /// - `NOSUPPORT`: Indicates that this data type isn't supported.
+    fn read_co2(&self) -> Result<(), ErrorCode>;
+
+    /// Read the Total Organic Compound (TVOC) from the sensor.
+    /// This will trigger the `AirQualityClient` `tvoc_data_available()`
+    /// callback when the data is ready.
+    ///
+    /// This function might return the following errors:
+    /// - `BUSY`: Indicates that the hardware is busy with an existing
+    ///           operation or initilisation/calibration.
+    /// - `NOSUPPORT`: Indicates that this data type isn't supported.
+    fn read_tvoc(&self) -> Result<(), ErrorCode>;
+}
+
+/// Client for receiving Air Quality readings
+pub trait AirQualityClient {
+    /// Called when a CO2 or equivalent CO2 (eCO2) reading has completed.
+    ///
+    /// - `value`: will contain the latest CO2 reading in ppm. An example value
+    ///            might be `400`.
+    fn co2_data_available(&self, value: u32);
+
+    /// Called when a Total Organic Compound (TVOC) reading has completed.
+    ///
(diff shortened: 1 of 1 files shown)


kernel/src/hil/sensors.rs
```diff
@@ -31,6 +31,47 @@ pub trait HumidityClient {
     fn callback(&self, value: usize);
 }
 
+/// A basic interface for a Air Quality sensor
+pub trait AirQualityDriver<'a> {
+    /// Set the client to be notified when the capsule has data ready.
+    fn set_client(&self, client: &'a dyn AirQualityClient);
+
+    /// Read the CO2 or equivalent CO2 (eCO2) from the sensor.
+    /// This will trigger the `AirQualityClient` `co2_data_available()
```
hudson-ayers, line 40:
```suggestion
    /// This will trigger the `AirQualityClient` `co2_data_available()`
```

kernel/src/hil/sensors.rs
```diff
@@ -31,6 +31,47 @@ pub trait HumidityClient {
...
+    /// callback when the data is ready.
+    ///
+    /// This function might return the following errors:
+    /// - `BUSY`: Indicates that the hardware is busy with an existing
+    ///           operation or initilisation/calibration.
+    /// - `NOSUPPORT`: Indicates that this data type isn't supported.
+    fn read_co2(&self) -> Result<(), ErrorCode>;
+
+    /// Read the Total Organic Compound (TVOC) from the sensor.
+    /// This will trigger the `AirQualityClient` `tvoc_data_available()
```
hudson-ayers, line 50:
```suggestion
    /// This will trigger the `AirQualityClient` `tvoc_data_available()`
```

kernel/src/hil/sensors.rs
```diff
@@ -31,6 +31,47 @@ pub trait HumidityClient {
...
+    ///
+    /// - `value`: will contain the latest CO2 reading in ppm. An example value
+    ///            might be `400`.
+    fn co2_data_available(&self, value: usize);
+
+    /// Called when a Total Organic Compound (TVOC) reading has completed.
+    ///
+    /// - `value`: will contain the latest TVOC reading in ppb. An example value
+    ///            might be `0`.
+    fn tvoc_data_available(&self, value: usize);
```
hudson-ayers, line 72: I realize we get this wrong in other places in this file, but I think it probably makes more sense to use `u32` here, the size of the value returned should probably not depend on the pointer width of the underlying platform.

alistair23: Good call. Fixed!

hudson-ayers (approved):

thulithwilfred (approved):

bradjc (approved):

kernel/src/hil/sensors.rs
```diff
@@ -31,6 +31,47 @@ pub trait HumidityClient {
...
+    fn read_tvoc(&self) -> Result<(), ErrorCode>;
+}
+
+/// Client for receiving Air Quality readings
+pub trait AirQualityClient {
+    /// Called when a CO2 or equivalent CO2 (eCO2) reading has completed.
+    ///
+    /// - `value`: will contain the latest CO2 reading in ppm. An example value
+    ///            might be `400`.
+    fn co2_data_available(&self, value: u32);
```
phil-levis, line 66: Why is CO2 in ppm but VOC in ppb? Should they both be ppb (given a 32-bit value)?

thulithwilfred: I think it makes sense to use these units, since normal outdoor eCO2 ranges are <~1000ppm. But the TVOC typically can be in the range of ~200-5500ppb. So for instance, 500ppb = 0.5ppm. Avoids having to deal with floats, if we use ppb.

thulithwilfred: I think given the ranges in which these units are measures it makes sense?
eCO2 usually <~1000ppm and tVOC is typically like 200-5500ppb.

alistair23: What @twilfredo said. It's to do with the usual ranges. It matches what I see others using and what the datasheets supply formulas to calculate

PR #3073, opened 2022-07-07
doc: porting: add note about svd2regs.py

### LINKED ISSUE #2951: can we have more document about porting new platform using code generate tools?
Hi, I'm new to TockOS and wanting to port a new chip. When I read the document (for example: https://github.com/tock/tock/blob/master/doc/Porting.md), I realized that it tell me to write chip crate by hand. I was suprised that it does not mention `svd2rust` at all. After some more search, I found that tockos has two tool called `svd2regs.py` and `nRF51_codegen.py`.

Now I'm confused:
* why there is no word about using tools to generate chip crate in the document? is it because the document is outdated, or because the tool not work well?
* what is the difference between `svd2regs.py` and `nRF51_codegen.py`?
* why not use `svd2rust`? and is it possible to use it?

I hope there is more document about those point and it can make port a new chip a lot eaiser.

bradjc:
### Pull Request Overview

Adds a note about svd2regs to the porting guide. Also updates spelling and wrapping.

Fixes #2951

### Testing Strategy

n/a

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/Porting.md
+++ b/doc/Porting.md
@@ -13,6 +13,7 @@ _It is a work in progress. Comments and pull requests are appreciated!_
 - [Crate Details](#crate-details)
   * [`arch` Crate](#arch-crate)
   * [`chip` Crate](#chip-crate)
+    + [Tips and Tools](#tips-and-tools)
   * [`board` Crate](#board-crate)
     + [Board Support](#board-support)
       - [`panic!`s (aka `io.rs`)](#panics-aka-iors)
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):

bradjc:
Fixed TOC.

hudson-ayers (approved):

PR #3080, opened 2022-07-11
arch/cortex-m: introduce CortexMVariant trait and avoid calls through `extern "C"` symbols

lschuermann:
### Pull Request Overview

Given different sub-variants of the Cortex-M architecture family, this introduces a trait to encapsulate differences in between Cortex-M variants. Most prominently, this includes interrupt handlers which are now passed as associated constants of trait implementations versus exported symbols, as well as the `switch_to_user` function, which is now exposed as a regular function by trait implementors.

This has some significant benefits over passing functions via symbols:

- By using a trait carrying proper first-level Rust functions, the type signatures of the trait and implementing functions are properly validated. Before these changes, some Cortex-M variants already used incorrect type signatures (e.g. `*mut u8` instead of `*const usize`) for the `user_stack` argument. It also ensures that all functions are provided by a respective sub-architecture at compile time, instead of throwing linker errors.

- Determining the respective functions at compile time, Rust might be able to perform more aggressive inlining, especially if more device-specific proper Rust functions (non hardware-exposed symbols, i.e. not fault or interrupt handlers) were to be added.

- Most importantly, this avoid ambiguity with respect to a compiler fence being inserted by the compiler around calls to `switch_to_user`. The `asm!` macro in that function call will cause Rust to emit a compiler fence given the `nomem` option is not passed, but the opaque `extern "C"` function call obscured that code path. While this is probably fine and Rust is obliged to generate a compiler fence when switching to C code, having a traceable code path for Rust to the `asm!` macro will remove any remaining ambiguity and allow us to argue against requiring volatile accesses to userspace memory (during context switches). See #2582 for further discussion of this issue.

Generally, this seems like the right way to go, which is further confirmed by our prior code using `#[allow(improper_ctypes)]` to annotate the `extern "C" fn switch_to_user` function reference.

The fact that some functions are proper trait functions, while others are exposed via associated constants is caused by the associated const functions being `#[naked]`. To wrap these functions in proper trait functions would require these trait functions to also be `#[naked]` to avoid generating a function prologue and epilogue, and we'd have to call the respective backing function from within an `asm!` block. The associated constants allow us to simply reference any function in scope and export it through the provided `CortexMVariant` trait infrastructure.

### Testing Strategy

This pull request was tested by compiling. ~It still needs to be tested on a (few) ARM board(s).~ It has undergone basic tests on an nRF52DK (run an application, observe console output).

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


arch/cortex-m0/src/lib.rs
```diff
@@ -191,21 +307,71 @@ pub unsafe extern "C" fn svc_handler() {
...
+
+    b {}    // Branch to the non-naked fault handler.
+    bx lr   // If continued function returns, we need to manually branch to
+            // link register.
+    ",
+    sym hard_fault_handler_continued,
+    options(noreturn));
+}
+
+pub enum CortexM0Subarch {}
```
hudson-ayers, line 352: I assume this is an enum rather than a struct so that it cannot be instantiated -- should we make a note about that in a comment?

arch/cortex-m4/src/lib.rs
```diff
@@ -3,41 +3,51 @@
 #![crate_name = "cortexm4"]
 #![crate_type = "rlib"]
 #![no_std]
+#![feature(naked_functions)]
```
hudson-ayers, line 6: Why do we need to add this feature? I don't see any naked functions added in this crate

hudson-ayers (dismissed):
Some of the code movement in `arch/cortex-m0/src/lib.rs` made this a bit harder to review than it would have been, but overall I really like this. We had effectively been using raw function pointers to abstract over interfaces when traits are purpose built to do exactly that.

I have not tested this yet but I think it looks good.

arch/cortex-m/src/lib.rs
```diff
@@ -15,6 +15,20 @@ pub mod support;
 pub mod syscall;
 pub mod systick;
 
+pub trait CortexMSubarch {
```
hudson-ayers, line 18: I think you should leave this commment from the PR description as a doc comment on this trait:

> The fact that some functions are proper trait functions, while others are exposed via associated constants is caused by the associated const functions being #[naked]. To wrap these functions in proper trait functions would require these trait functions to also be #[naked] to avoid generating a function prologue and epilogue, and we'd have to call the respective backing function from within an asm! block. The associated constants allow us to simply reference any function in scope and export it through the provided CortexMSubarch trait infrastructure.

lschuermann:
> Some of the code movement in `arch/cortex-m0/src/lib.rs` made this a bit harder to review than it would have been

Right, sorry. I think that one part of the code had to be moved given the introduction of the traits, whereas other parts have been shuffled around to collect the naked fns in a single location, above the trait implementation. Assuming I didn't mess this up, this really should only change the order of functions defined there.

I agree that we should've been using traits before, but feel like the idea to use associated constants was a blocker to implementing this (at least took me some time to think of that).

Will test these changes first thing tomorrow when I have access to an ARM board again.

arch/cortex-m/src/lib.rs
lschuermann: Absolutely.

arch/cortex-m0/src/lib.rs
lschuermann: Yes, that's the idea. I'll add a comment :+1:.

arch/cortex-m4/src/lib.rs
lschuermann: My bad, this was leftover from a previous attempt which didn't use associated constants. Wondering why I didn't get a warning about an unused feature here?

arch/cortex-m/src/lib.rs
lschuermann: I've added some documentation which hopefully explains this trait's purpose and the particular structure chosen here. I've furthermore moved the function documentation comments from the generic ARMv7-M `cortexm` functions into the trait, as this is what should be specifying the function requirements and is rendered as part of the rustdoc.

arch/cortex-m/src/lib.rs
```diff
@@ -15,6 +15,61 @@ pub mod support;
...
 
+/// Trait to encapsulate differences in between Cortex-M variants
+///
+/// This trait contains functions and other associated data (constants) which
+/// differs in between different Cortex-M architecture variants (e.g. Cortex-M0,
+/// Cortex-M4, etc.). It is not designed to be implemented by an instantiable
+/// type and passed around as a runtime-accessible object, but is to be used as
+/// a well-defined collection of functions and data to be exposed to
+/// architecture-dependent code paths. Hence, implementations can use an `enum`
+/// without variants to implement this type.
```
hudson-ayers, line 26:
```suggestion
/// without variants to implement this trait.
```

arch/cortex-m/src/lib.rs
```diff
@@ -15,6 +15,61 @@ pub mod support;
...
+///
+/// The fact that some functions are proper trait functions, while others are
+/// exposed via associated constants is caused by the associated const functions
```
hudson-ayers, line 29:
```suggestion
/// exposed via associated constants is necessitated by the associated const functions
```

arch/cortex-m/src/lib.rs
```diff
@@ -15,6 +15,61 @@ pub mod support;
...
+/// being `#\[naked\]`. To wrap these functions in proper trait functions would
+/// require these trait functions to also be `#\[naked\]` to avoid generating a
+/// function prologue and epilogue, and we'd have to call the respective backing
+/// function from within an asm! block. The associated constants allow us to
+/// simply reference any function in scope and export it through the provided
+/// CortexMSubarch trait infrastructure.
+pub trait CortexMSubarch {
+    /// All ISRs are caught by this handler. This must ensure the interrupt is
```
hudson-ayers, line 37:
```suggestion
    /// All ISRs not caught by a more specific handler are caught by this handler. This must ensure the interrupt is
```

hudson-ayers (approved):

arch/cortex-m/src/lib.rs
```diff
@@ -15,6 +15,62 @@ pub mod support;
...
+///
+/// The fact that some functions are proper trait functions, while others are
+/// exposed via associated constants is necessitated by the associated const
+/// functions being `#\[naked\]`. To wrap these functions in proper trait
+/// functions would require these trait functions to also be `#\[naked\]` to
+/// avoid generating a function prologue and epilogue, and we'd have to call the
+/// respective backing function from within an asm! block. The associated
+/// constants allow us to simply reference any function in scope and export it
+/// through the provided CortexMSubarch trait infrastructure.
+pub trait CortexMSubarch {
```
bradjc, line 36: Subarch is a little confusing, because there is no cortex-m arch. Maybe `CortexMFamily` or `CortexMVariant`.

arch/cortex-m/src/lib.rs
bradjc: Can we also capture a version of


> This has some significant benefits over passing functions via symbols:

> By using a trait carrying proper first-level Rust functions, the type signatures of the trait and implementing functions are properly validated. Before these changes, some Cortex-M variants already used incorrect type signatures (e.g. *mut u8 instead of *const usize) for the user_stack argument. It also ensures that all functions are provided by a respective sub-architecture at compile time, instead of throwing linker errors.

> Determining the respective functions at compile time, Rust might be able to perform more aggressive inlining, especially if more device-specific proper Rust functions (non hardware-exposed symbols, i.e. not fault or interrupt handlers) were to be added.

> Most importantly, this avoid ambiguity with respect to a compiler fence being inserted by the compiler around calls to switch_to_user. The asm! macro in that function call will cause Rust to emit a compiler fence given the nomem option is not passed, but the opaque extern "C" function call obscured that code path. While this is probably fine and Rust is obliged to generate a compiler fence when switching to C code, having a traceable code path for Rust to the asm! macro will remove any remaining ambiguity and allow us to argue against requiring volatile accesses to userspace memory (during context switches). See https://github.com/tock/tock/pull/2582 for further discussion of this issue.


here, perhaps as a normal `//` comment? This explanation is great to capture independent of github.

bradjc (commented):
I didn't go through and double check every chip but based on the description this seems like a good change.

bradjc (approved):
Looks good

hudson-ayers (approved):
Thanks for documenting all of this so well

PR #3083, opened 2022-07-15
Remove one use of `asm_const`

hudson-ayers:
### Pull Request Overview

This pull request removes one of our two remaining uses of `#![feature(asm_const)]`, at the low cost of a little readability. Our one remaining use is in `riscv_csr`. Removing that use is more complex, and would require going back to either the switch statement approach to CSRs (e.g. before https://github.com/tock/tock/pull/2472/files), or revisiting the macro-based approach that Alistair suggested (https://github.com/tock/tock/pull/2470/files).

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/litex_vexriscv/src/interrupt_controller.rs
+++ b/chips/litex_vexriscv/src/interrupt_controller.rs
@@ -108,7 +108,9 @@ mod vexriscv_irq_raw {
     pub unsafe fn irq_getmask() -> usize {
         let mask: usize;
         use core::arch::asm;
-        asm!("csrr {mask}, {csr}", mask = out(reg) mask, csr = const CSR_IRQ_MASK);
+        // TODO: If asm_const is stabilized, transition back to below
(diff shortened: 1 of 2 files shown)


alistair23 (dismissed):

bradjc (commented):
👍

hudson-ayers:
Rebased

ppannuto (approved):
Looks reasonable to me.

I think it would be nice at some point to try to have a 'dashboard' view, that has the features we're watching for stabilization of like this (maybe a table, with links to upstream trackers). The header in #1654 is some of this, but has understandably gotten pretty messy. I know we've false-started wiki pages a few times before, but maybe this is actually a case for one? :/

brghena (approved):

alistair23 (approved):

bradjc (approved):

PR #3084, opened 2022-07-15
Define Buzzer HIL and PWM Buzzer Driver

TeodoraMiu:
### Pull Request Overview

This pull request changes the way the buzzer driver works. Before this PR, the buzzer syscall driver assumed all buzzer systems work with PWM. This PR:
- defines a buzzer HIL
- provides a PWM buzzer driver that implements the buzzer HIL
- provides a generic buzzer syscall driver
- uses the driver for the Microbit, clue and acd boards

### Testing Strategy

### TODO or Help Wanted

Feedback for the buzzer HIL is much appreciated.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


boards/microbit_v2/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2021"
 [dependencies]
 cortexm4 = { path = "../../arch/cortex-m4" }
 capsules = { path = "../../capsules" }
-kernel = { path = "../../kernel" }
+kernel = { path = "../../kernel", features = ["trace_syscalls"] }
```
alexandruradovici, line 11:
```suggestion
kernel = { path = "../../kernel" }
```

capsules/src/buzzer.rs
```diff
@@ -0,0 +1,143 @@
...
+                } else {
+                    // If there is no active app or the same app is trying to use the buzzer,
+                    // we set/replace the frequency and duration.
+                    self.active_app.set(appid);
+                    self.buzzer.buzz(data1, data2).into()
+                }
+            }
+
+            2 =>
+            // Stop the current sound.
```
alexandruradovici, line 111:
```suggestion
            // Stop the current sound.
            2 =>
```

capsules/src/buzzer.rs
```diff
@@ -0,0 +1,143 @@
...
+    ) -> CommandReturn {
+        match command_num {
+            0 =>
+            // Check whether the driver exists.
+            {
+                CommandReturn::success()
+            }
+
+            1 =>
+            // Play a sound.
```
alexandruradovici, line 97:
```suggestion
            // Play a sound.
            1 =>
```

capsules/src/buzzer.rs
```diff
@@ -0,0 +1,143 @@
...
+    fn command(
+        &self,
+        command_num: usize,
+        data1: usize,
+        data2: usize,
+        appid: ProcessId,
+    ) -> CommandReturn {
+        match command_num {
+            0 =>
+            // Check whether the driver exists.
```
alexandruradovici, line 91:
```suggestion
            // Check whether the driver exists.
            0 =>
```

capsules/src/buzzer.rs
```diff
@@ -0,0 +1,143 @@
...
+        grant: Grant<(), UpcallCount<1>, AllowRoCount<0>, AllowRwCount<0>>,
+    ) -> Buzzer<'a, B> {
+        Buzzer {
+            buzzer: buzzer,
+            apps: grant,
+            active_app: OptionalCell::empty(),
+        }
+    }
+
+    pub fn is_valid_app(&self, appid: ProcessId) -> bool {
```
alexandruradovici, line 51: Please add a comment to explain what a valid app means for this driver.

kernel/src/hil/buzzer.rs
```diff
@@ -0,0 +1,37 @@
...
+
+use crate::ErrorCode;
+
+pub trait BuzzerClient {
+    /// Called when the current sound played by the buzzer has finished
+    /// or it was stopped.
+    fn buzzer_done(&self, status: Result<(), ErrorCode>);
+}
+
+pub trait Buzzer<'a> {
```
alexandruradovici, line 11: Please add a comment to explain that the main purpose of the Buzzer HIL is. (ex: playing a sound of a fixed frequency).

alexandruradovici (changes requested):

capsules/src/buzzer.rs
TeodoraMiu: Done

capsules/src/buzzer.rs
TeodoraMiu: Done

capsules/src/buzzer.rs
TeodoraMiu: Done

lschuermann:
> Before this PR, the buzzer syscall driver assumed all buzzer systems work with PWM.

Do you have an example for a buzzer which does not work with PWM? I I'm not mistaken, this PR only has an implementation for PWM-controlled buzzers, right?

alexandruradovici:
> Do you have an example for a buzzer which does not work with PWM? I I'm not mistaken, this PR only has an implementation for PWM-controlled buzzers, right?

You might be right, even tough I can think of having for instance an external device over some bus. 

The main idea behind the HIL came from a use case where the buzzer was needed from another capsule.

lschuermann:
I am not necessarily averse to having a buzzer HIL, but I think that there should be a compelling usecase for such peripherals which might be representable by a more general-purpose API. I figure an external chip will use some communication bus, which might be a more appropriate layer of abstraction in the kernel. Would you mind describing your usecase for controlling buzzers from within the kernel?

kernel/src/hil/buzzer.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    /// Stop the sound currenty playing.
+    /// After the buzzer is successfully stopped, the `buzzer_done()`
+    /// callback is called.
+    ///
+    /// Return values:
+    ///
+    /// - `Ok(())`: The attempt at stopping the buzzer was successful.
+    /// - `FAIL`: Cannot stop the buzzer.
+    fn stop(&self) -> Result<(), ErrorCode>;
```
bradjc, line 34: If the buzzer is not currently buzzing, what happens if `stop()` is called?

bradjc (changes requested):
Why did you remove the support for multiple apps and queuing?

TeodoraMiu:
> Why did you remove the support for multiple apps and queuing?

I figured that an app that wants to notify the user using the buzzer wants to do it immediately, not wait after another notification. Therefore the buzzer should play the sound instantly, otherwise it doesn't really serve its purpose.

kernel/src/hil/buzzer.rs
TeodoraMiu: I changed it so that the capsule returns ErrorCode::OFF if the stop command is called while the buzzer isn't buzzing.

alexandruradovici (approved):

phil-levis:
@bradjc Have your concerns been addressed?

bradjc:
> @bradjc Have your concerns been addressed?

No, not really. We currently have a virtualized buzzer that supports multiple apps, and this replaces it with one that makes the userspace app responsible for doing retries. Perhaps we need a new "buzz immediate" syscall or syscall driver that boards that only want immediate buzz (or if the platform can't provide an immediate buzz it returns an error) can use. But I don't see why we would get rid of the current support for queuing in the process.

I also don't see the need to change the name from `buzzer_driver.rs` to `buzzer.rs`. `buzzer_driver` is more clear this is a userspace driver.

TeodoraMiu:
I believe this is now good to go. I changed the driver so that there is a separate syscall for the buzzer to immediately buzz, but I also kept a syscall for queuing buzz commands. I also returned to the old buzzer_driver.rs name.

phil-levis:
> I believe this is now good to go. I changed the driver so that there is a separate syscall for the buzzer to immediately buzz, but I also kept a syscall for queuing buzz commands. I also returned to the old buzzer_driver.rs name.

@bradjc?

phil-levis:
@TeodoraMiu Looks like there are some simple conflicts to resolve.

kernel/src/hil/buzzer.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// Once the buzzer finishes buzzing, the `buzzer_done()` callback
+    /// is called.
+    /// If it is called while the buzzer is playing, the buzzer command will be
+    /// overriden with the new frequency and duration values.
+    ///
+    /// Return values:
+    ///
+    /// - `Ok(())`: The attempt at starting the buzzer was successful.
+    /// - `FAIL`: Cannot start the buzzer.
+    fn buzz(&self, frequency_hz: usize, duration_ms: usize) -> Result<(), ErrorCode>;
```
ppannuto, line 24: Should this have an `EVALUE` for impossible frequencies?

bradjc (approved):

bradjc (approved):

phil-levis:
Great!

PR #3056, opened 2022-06-14
OpenTitan: Bump to the latest RTL SHA

alistair23:
### Pull Request Overview

This bumps us to the latest SHA. This includes support for the Manifest and the ePMP changes

### Testing Strategy

Running on FPGA and Verilator

### TODO or Help Wanted

Still get all the tests passing ~~and ePMP cleaned up~~

### Documentation Updated

- [X] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [X] Ran `make prepush`.

PR #3085, opened 2022-07-15
WIP: OT bitstream update

cfrantz:
### Pull Request Overview

This pull request is a work in progress to bring tock upstream into a state that works with the current version of the OpenTitan hardware.  It includes the following:

1. Adjust the `flash_header` of the image to be consistent with the TestROM/MaskROM manifest definition.
2. Temporarily disable the EPMP initialization.  The current epmp init scheme attempts to auto-detect the state of the pmp and guess which entries are available for use.  This has the side effect of trashing the epmp and faulting the boot process.  Philosophically, I disagree with this epmp initialization approach.  The OpenTitan boot process will define which epmp entries a given implementation (chip) will use and the boot processes through the ROM, ROM_EXT and Owner Bootloader will leave the epmp in a deterministic and known state.
3. Refactor the earlgrey_cw310 crate from a `bin` crate into a `lib+bin` crate.  The OpenTitan project wants to declare tock-upstream as a dependency to its own source base and provide its own per-product board init sources.  At the same time, we want tock-upstream to have an independent and functional board definition that can initialize some of the common hardware (e.g. currently earlgrey_cw310).  We are using [cargo raze](https://github.com/google/cargo-raze) to automatically consume the Cargo.toml files and create bazel definitions within our codebase.  By making the earlgrey_cw310 crate a `lib+bin` crate, we can depend on it as a somewhat generic source of init code while defining a `main` in our own sources.
4. Make TicKV optional - it is unclear at this point what we might choose as a flash-filesystem.  Furthermore, I believe that many of the other hardware inits in the `earlgrey_cw310` crate should be feature-gated so make the initialization sequence finer-grained.
5. Automate building `rsa.tbf` which is needed as a test resource.

### Testing Strategy

This PR was tested by following the instructions in `boards/opentitan/README.md`.  Unfortunately, a number of the tests are failing for various reasons:

1. Building `rsa.tbf` does not seem to result in a valid tock binary.  In fact, elf2tab fails on that file because the start symbol appears to overlap both the `.text` and `.data` sections.  However, the OTBN accelerator in OT is a Harvard architecture and as such, the apparent overlap isn't a real overlap.
2. I'm not certain, but I don't think the current flash driver/filesystem tests _can_ pass.  I believe the flash driver has an error involving the flash block size.

### TODO or Help Wanted

This pull request still needs **review and discussion**.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


boards/opentitan/src/main.rs
```diff
@@ -7,764 +7,14 @@
 // https://github.com/rust-lang/rust/issues/62184.
 #![cfg_attr(not(doc), no_main)]
 #![feature(custom_test_frameworks)]
-#![test_runner(test_runner)]
+#![test_runner(my_test_runner)]
```
jrvanwhy, line 10: This is the only place I see `my_test_runner` -- is this change a leftover?

arch/rv32i/src/epmp.rs
```diff
@@ -108,6 +112,7 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> PMP<MAX_AVAILABLE_REGIONS_OVER
             // Reset back to how we found it
             csr::CSR.pmpconfig_set(i / 4, pmpcfg_og);
         }
+        */
```
jrvanwhy, line 115: Tock doesn't usually like leaving commented-out code in the repository. They'd prefer that PR authors remove the code, and if it is needed later someone can dig it out of git history.

alistair23:
For updating the bitstream I think it's best to continue with https://github.com/tock/tock/pull/3056

It's significantly further along. It already contains the changes to support the manifest, already contains fixes for the registers that have changed and is heading towards getting the ePMP up and running. It's mostly just stuck on the Flash and CSRNG tests failing

As for the refactoring that can just be split out into a separate PR.

For the OTBN app you will need this PR, which is still pending approval: https://github.com/tock/elf2tab/pull/37

cfrantz:
I'm going to close this PR until after the OpenTitan project institutes a formal release process.

PR #3086, opened 2022-07-15
RFC: Restructure `DeferredCall` to not require atomics

hudson-ayers:
### Pull Request Overview

This PR restructures `DeferredCall` to not require atomics, instead relying on a `DeferredCallManager` structure which is simply a wrapper around `Cell<usize>` and is held by the `Chip` struct. This allows us to remove the perma-unstable `#![feature(core_intrinsics)]` from the Tock kernel, and removes some unsafe code. This change is safe because we only ever touch the Manager from the main kernel thread (not interrupt context) so we never really needed atomics anyway. In theory, using a `static mut` global usize would have worked just as well and allowed us to more closely retain the existing API, but given the documented issues with safe/correct use of `static mut` I prefer this approach.

This change comes at the expense of about ~300 bytes of code size on platforms with multiple deferred calls -- however deferred calls are actually rarely used, with only 3 of Tock's supported chips requiring any (in contrast to dynamic deferred calls which are more widely used).

~~So far, I have only implemented this for Imix, as I would like feedback before implementing it for the remaining chips.~~
I have implemented this for all chips/boards now.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/acd52832/src/main.rs
+++ b/boards/acd52832/src/main.rs
@@ -9,6 +9,7 @@
 use capsules::virtual_alarm::VirtualMuxAlarm;
 use kernel::capabilities;
 use kernel::component::Component;
+use kernel::deferred_call::DeferredCallManager;
 use kernel::dynamic_deferred_call::{DynamicDeferredCall, DynamicDeferredCallClientState};
 use kernel::hil;
 use kernel::hil::adc::Adc;
@@ -22,6 +23,7 @@ use kernel::platform::{KernelResources, SyscallDriverLookup};
 use kernel::scheduler::round_robin::RoundRobinSched;
 #[allow(unused_imports)]
 use kernel::{create_capability, debug, debug_gpio, static_init};
+use nrf52832::deferred_call_tasks::DeferredCallTask;
 use nrf52832::gpio::Pin;
 use nrf52832::interrupt_service::Nrf52832DefaultPeripherals;
 use nrf52832::rtc::Rtc;
@@ -152,10 +154,15 @@ impl KernelResources<nrf52832::chip::NRF52<'static, Nrf52832DefaultPeripherals<'
 /// these static_inits is wasted.
 #[inline(never)]
 unsafe fn get_peripherals() -> &'static mut Nrf52832DefaultPeripherals<'static> {
+    let dc_mgr = static_init!(
+        DeferredCallManager<DeferredCallTask>,
+        DeferredCallManager::new()
+    );
+
     // Initialize chip peripheral drivers
     let nrf52832_peripherals = static_init!(
         Nrf52832DefaultPeripherals,
-        Nrf52832DefaultPeripherals::new()
+        Nrf52832DefaultPeripherals::new(dc_mgr)
     );
 
     nrf52832_peripherals
@@ -165,7 +172,6 @@ unsafe fn get_peripherals() -> &'static mut Nrf52832DefaultPeripherals<'static> 
 #[no_mangle]
 pub unsafe fn main() {
     nrf52832::init();
-
     let nrf52832_peripherals = get_peripherals();
 
     // set up circular peripheral dependencies
--- a/kernel/src/deferred_call.rs
+++ b/kernel/src/deferred_call.rs
@@ -4,81 +4,69 @@
 //! in the chip scheduler if the hardware doesn't support interrupts where
 //! they are needed.
 
-use core::cell::UnsafeCell;
+use core::cell::Cell;
 use core::convert::Into;
 use core::convert::TryFrom;
 use core::convert::TryInto;
-use core::intrinsics;
 use core::marker::Copy;
-use core::marker::Sync;
+use core::marker::PhantomData;
 
-/// AtomicUsize with no CAS operations that works on targets that have "no atomic
-/// support" according to their specification. This makes it work on thumbv6
-/// platforms.
-///
-/// Borrowed from https://github.com/japaric/heapless/blob/master/src/ring_buffer/mod.rs
-/// See: https://github.com/japaric/heapless/commit/37c8b5b63780ed8811173dc1ec8859cd99efa9ad
-struct AtomicUsize {
-    v: UnsafeCell<usize>,
+/// Any chip with peripherals which require deferred calls should
+/// instantiate exactly one of these, and a reference to that manager should be
+/// passed to all created `DeferredCall`s.
+pub struct DeferredCallManager<T: Into<usize> + TryFrom<usize> + Copy> {
+    v: Cell<usize>,
+    _p: PhantomData<T>,
 }
 
-impl AtomicUsize {
-    pub(crate) const fn new(v: usize) -> AtomicUsize {
-        AtomicUsize {
-            v: UnsafeCell::new(v),
+impl<T: Into<usize> + TryFrom<usize> + Copy> DeferredCallManager<T> {
+    pub fn new() -> Self {
+        Self {
+            v: Cell::new(0),
+            _p: PhantomData,
         }
     }
 
-    pub(crate) fn load_relaxed(&self) -> usize {
-        unsafe { intrinsics::atomic_load_relaxed(self.v.get()) }
-    }
-
-    pub(crate) fn store_relaxed(&self, val: usize) {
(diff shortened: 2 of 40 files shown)


kernel/src/platform/chip.rs
```diff
@@ -109,6 +109,20 @@ pub trait InterruptService<T> {
...
+    /// Returns true if any deferred call tasks are pending, otherwise returns false.
+    /// Default implementation just returns false, and should be overriden by chips that require
+    /// defferred calls.
+    fn has_deferred_call_tasks(&self) -> bool {
+        false
+    }
+
+    /// Returns the next pending deferred call task, if any.
+    /// Default implementation just returns None, and should be overriden by chips that require
+    /// deferred calls
```
bradjc, line 122:
```suggestion
    /// deferred calls.
```

hudson-ayers:
> Since this changes boards, I think we need to add something to clarify why we have two deferred call mechanisms, and how they should be used. Is there anything stopping a capsule from using the new `DeferredCallManager`?

A capsule cannot use the `DeferredCallManager` because capsules cannot reference code in `chips/`, and `DeferredCall` relies on a statically defined chip-specific list of deferred calls.

Peripherals *could* use `DynamicDeferredCall` instead of `DeferredCall`, but `DeferredCall` is lower cost (the dynamic version uses trait object dispatch, and the dynamic version requires boards to define an array representing the number of `DynamicDeferredCall` handles, which changes dynamically depending which capsules and peripherals you include.

I don't think it would be crazy to advocate for everything using `DynamicDeferredCall` though.

> Also, should this be `StaticDeferredCall`?

It could be, I suppose. I am indifferent to the name change.

alistair23 (approved):

bradjc:
> A capsule cannot use the `DeferredCallManager` because capsules cannot reference code in `chips/`, and `DeferredCall` relies on a statically defined chip-specific list of deferred calls.

And this is because when `DeferredCallManager` is created it is templated on a specific task list? `DeferredCallManager<sam4l::deferred_call_tasks::Task>`? So a capsule couldn't somehow use one designed for a chip, but couldn't a capsule author expect the board to pass it a `DeferredCallManager`? The list of tasks would have to be defined in the capsule crate.

If that is possible, is that what we want, or would we be ok if a capsule author did that? If this still isn't possible then this is moot.



> >    Also, should this be StaticDeferredCall?

> It could be, I suppose. I am indifferent to the name change.

This goes along with my other question, I think we need to set this up so we can easily teach people the difference and what to use when. I think having parallel naming helps with that.

lschuermann:
I think this change is good generally, but only a small selection of chips needs to have these types of static deferred calls, right? I'm not opposed to any of this, but just to throw this thought out there, is it really worth to provide some central (and actually pretty minimal) infrastructure for this, compared to implementing static deferred calls implicitly in the chip's schedulers themselves? The answer might well be yes, I'm only curious on what others think.

kernel/src/deferred_call.rs
```diff
@@ -4,81 +4,69 @@
...
+}
 
 impl<T: Into<usize> + TryFrom<usize> + Copy> DeferredCall<T> {
     /// Creates a new DeferredCall
     ///
     /// Only create one per task, preferably in the module that it will be used
-    /// in.
-    pub const unsafe fn new(task: T) -> Self {
-        DeferredCall(task)
+    /// in. Creating more than 32 tasks on a given manager will lead to
```
ppannuto, line 60:
```suggestion
    /// in. Creating more than `NUM_BITS(usize)` tasks on a given manager will lead to
```

Should probably express this like this; or use a `u32`

ppannuto (commented):
> however deferred calls are actually rarely used, with only 3 of Tock's supported chips requiring any (in contrast to dynamic deferred calls which are more widely used).

> I don't think it would be crazy to advocate for everything using DynamicDeferredCall though.

I would be interested to understand how the all-dynamic looks. Having two mechanisms for a low-level, correctness-sensitive part of the execution flow feels like it increases complexity a lot for users. If the static case is really so rare, what is the cost (size? performance?) from just removing it?

hudson-ayers:
> > however deferred calls are actually rarely used, with only 3 of Tock's supported chips requiring any (in contrast to dynamic deferred calls which are more widely used).
> 
> > I don't think it would be crazy to advocate for everything using DynamicDeferredCall though.
> 
> I would be interested to understand how the all-dynamic looks. Having two mechanisms for a low-level, correctness-sensitive part of the execution flow feels like it increases complexity a lot for users. If the static case is really so rare, what is the cost (size? performance?) from just removing it?

I took a stab at removing `DeferredCall`, implementing this only for Imix: https://github.com/tock/tock/compare/master...hudson-ayers:tock:no_def_call?expand=1 . The size cost is about 400 bytes, I have not taken any performance measurements. It does simplify the interrupt handling a good bit in sam4l/src/chips.rs, and the `InterruptService` trait. On the other hand the peripheral code that uses `DynamicDeferredCall` now becomes less straightforward, as does the peripheral initialization code.


> Perhaps we should ask the inverse as well. Are we doing truly dynamic deferred calls, where new sources of deferred calls are configured at runtime with some uncertainty? (I'm not sure the best way to phrase this, our dyn deferred calls are configured at runtime but its all predictable in main.rs.) Why don't we just use the static manager, and have main.rs assign numbers at compile time?

I will take a stab at this next.

hudson-ayers:
Update: I am still working on fully static assignment of deferred calls, from both capsules and chip peripherals, but there are some tricky subproblems that arise as a result of none of `{capsules, chips, kernel}` having visibility into all three crates, and as a result of wanting to get rid of global variables as part of this change. While I still think this is possible, I worry the final result is going to be a pretty big refactoring and will make using deferred calls a good bit less ergonomic.

bradjc:
> Update: I am still working on fully static assignment of deferred calls, from both capsules and chip peripherals, but there are some tricky subproblems that arise as a result of none of `{capsules, chips, kernel}` having visibility into all three crates, and as a result of wanting to get rid of global variables as part of this change. While I still think this is possible, I worry the final result is going to be a pretty big refactoring and will make using deferred calls a good bit less ergonomic.

Is that when using a single manager for deferred calls? If so, would it work/help to have multiple static managers? Or does having multiple managers revert the benefit in terms of scheduling clarity/simplicity?

hudson-ayers:
> Is that when using a single manager for deferred calls? If so, would it work/help to have multiple static managers? Or does having multiple managers revert the benefit in terms of scheduling clarity/simplicity?

Yes, this is using a single manager. I don't think having multiple static managers would be a huge help. Frankly I think I just got a bit overfocused on wanting the deferred call id --> callback mapping to be a closure, which leads to the following problem:

1. main.rs needs to define a mapping of `DeferredCallID` --> callback_handler, and pass that mapping to the kernel scheduler (or to the Chip struct)
2. The logical way to do this is to define a function, but callback handlers for capsules are methods on objects. So can't define a global function since we need to reference the specific object we are calling the method on.
3. Ok, we can define a closure for this! But...we cant instantiate a closure with a static lifetime, and we end up needing to store this mapping on either the Kernel or the Chip, both of which have static lifetimes, so Rust does not let us store a reference to this "shorter-lived" closure.

So, we end up with no easy way to store a mapping defined in main.rs in one of the places where having that mapping would be useful.

I think if I move away from using a closure it is possible to get something better using a custom type

kernel/src/deferred_call.rs
kupiakos: `usize::BITS` is how you spell this in Rust

boards/acd52832/src/main.rs
```diff
@@ -22,6 +23,7 @@ use kernel::platform::{KernelResources, SyscallDriverLookup};
 use kernel::scheduler::round_robin::RoundRobinSched;
 #[allow(unused_imports)]
 use kernel::{create_capability, debug, debug_gpio, static_init};
+use nrf52832::deferred_call_tasks::DeferredCallTask;
```
bradjc, line 26: I think it would be nice to remove this import, as it would make the `dc_mgr` more clear that it is chip-specific.

hudson-ayers:
Closing in favor of #3382

Issue #3019, opened 2022-04-14
Tock-registers's test_fields macro fails in undocumented and non-obvious ways


jscatena88:
The test_fields macro seems to not be able to account for the size of a struct being influenced by the alignment constraints of a repr[C] struct; specifically the Rust docs say ["The alignment of the struct is the alignment of the most-aligned field in it. ... Finally, the size of the struct is the current offset rounded up to the nearest multiple of the struct's alignment."](https://doc.rust-lang.org/reference/type-layout.html). Take for example this code:
```
extern crate tock_registers;

use tock_registers::register_structs;
use tock_registers::registers::ReadWrite;

register_structs! {
    pub RegisterBlock {
        (0x00 => foo: ReadWrite<u64>),
        (0x08 => bar: ReadWrite<u32>),
        (0x0C => @END),
    }
}
fn main() {
    println!("{}", core::mem::size_of::<RegisterBlock>());
}
```

Running `cargo test` fails with the following error:
```
running 1 test
test test_register_structs::RegisterBlock::test_offsets ... FAILED

failures:

---- test_register_structs::RegisterBlock::test_offsets stdout ----
thread 'test_register_structs::RegisterBlock::test_offsets' panicked at 'Invalid size for struct RegisterBlock (expected 0xC but was 0x10)', src/main.rs:6:1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This is because the alignment of the `u64` element dictates that the size must be a multiple of 8. However even changing the `@END` marker to `0x10` as shown here:
```
extern crate tock_registers;

use tock_registers::register_structs;
use tock_registers::registers::ReadWrite;

register_structs! {
    pub RegisterBlock {
        (0x00 => foo: ReadWrite<u64>),
        (0x08 => bar: ReadWrite<u32>),
        (0x10 => @END),
    }
}
fn main() {
    println!("{}", core::mem::size_of::<RegisterBlock>());
}
```
Also generates an error, that is actually more confusing since it seems to indicate that the size is now smaller than before: 

```
running 1 test
test test_register_structs::RegisterBlock::test_offsets ... FAILED

failures:

---- test_register_structs::RegisterBlock::test_offsets stdout ----
thread 'test_register_structs::RegisterBlock::test_offsets' panicked at 'Invalid end offset for field bar (expected 0x10 but was 0xC)', src/main.rs:6:1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

[...]

PR #3088, opened 2022-07-17
tock_registers/test_fields: respect struct size padding w/ alignment

lschuermann:
### Pull Request Overview

As illustrated in tock/tock#3019, Rust causes a struct's size to be padded to a multiple of the largest field's alignment constraints.

The tock-register-interface crate contains some sanity checks which ensure that a struct is not missing any padding fields, all fields are generally well aligned, and that the fields are placed at their respective specified memory address. As of tock/tock#2922, these checks are no longer runtime tests, but static const assertions which cause proper compiler errors, should an assertion be violated.

Because the tock-registers-interface macros check invariants by iterating over the individual fields and accumulating their size, the above property of Rust structs has not been taken into account properly and caused compiler errors for register structs which contain fields of varying alignment and size.

This commit changes the iterative processing of register fields to also track the highest alignment constraint of any field encountered and, at the end of the struct, add the remaining alignment to the expected size of the struct. Thus the following register definition works now:

```rust
register_structs! {
    pub RegisterBlock {
	(0x00 => foo: ReadWrite<u64>),
	(0x08 => bar: ReadWrite<u32>),
	(0x0C => @END),
    }
}

fn test_struct_loc() {
    // Undefined behavior here, but sufficient for checking offsets
    let regs = unsafe { &*(0x0 as *const RegisterBlock) };
    assert!(regs as *const _ as usize == 0x0);
    assert!(&regs.foo as *const _ as usize == 0x0);
    assert!(&regs.bar as *const _ as usize == 0x8);
}
```

Fixes tock/tock#3019. Perhaps you can confirm, @jscatena88?

### Testing Strategy

This pull request was tested by the example provided in the commit message.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/libraries/tock-register-interface/src/macros.rs
+++ b/libraries/tock-register-interface/src/macros.rs
@@ -179,7 +179,7 @@ macro_rules! test_fields {
     // Macro entry point.
     (@root $struct:ident $(<$life:lifetime>)? { $($input:tt)* } ) => {
         // Start recursion at offset 0.
-        $crate::test_fields!(@munch $struct $(<$life>)? ($($input)*) : 0);
+        $crate::test_fields!(@munch $struct $(<$life>)? ($($input)*) : (0, 0));
     };
 
     // Consume the ($size:expr => @END) field, which MUST be the last field in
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):

jrvanwhy:
I don't know if this affects the correctness of this PR, but just a reminder that Rust types can have an alignment larger than the maximum of the alignments of their fields:

```rust
#[repr(C,align(32))]
struct Foo {
    address: usize,
}

fn main() {
    println!("align_of::<usize>(): {}", core::mem::align_of::<usize>());
    println!("align_of::<Foo>(): {}", core::mem::align_of::<Foo>());
    println!("size_of::<Foo>(): {}", core::mem::size_of::<Foo>());
}
```

```
align_of::<usize>(): 8
align_of::<Foo>(): 32
size_of::<Foo>(): 32
```

lschuermann:
> I don't know if this affects the correctness of this PR, but just a reminder that Rust types can have an alignment larger than the maximum of the alignments of their fields

That's an excellent point. I did take this into consideration, but given we only care about the size of the composite structure, and that depends on the alignment of its _fields_, I think that the alignment of that `struct` should hopefully not matter too much. We might want to introduce checks for constructing instances of these structures though, which is where the alignment is of concern. I'd say that given @kupiakos and I wanted to look at the register structures anyways (given the compiler is technically able to insert spurious reads to these struct fields), we should go back to this issue later.

jscatena88:
Thank you for getting to fixing this!

PR #3090, opened 2022-07-18
libraries/tickv: crc32: Cleanup the CRC32 interface and expose it publically

alistair23:
### Pull Request Overview

The original interface was clunky and required creating a Crc struct and
then getting a Digest from there. This patch instead entirely hides the
Crc struct and we just provide a Crc32 struct for performing CRC
operations.

This patch also documents all of the public functions and structs.

Finally we also reset the value after calling finalise() to allow
reusing the Crc after it has completed.

We also then expose the CRC implementation publicly.

### Testing Strategy

Running the TicKV tests.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


alistair23:
@BravoHoseok

libraries/tickv/src/crc32.rs
```diff
@@ -307,15 +309,11 @@ const fn crc32(poly: u32, mut byte: u8) -> u32 {
...
 
-pub struct Crc {}
-
-pub struct Digest<'a> {
-    crc: &'a Crc,
-    value: u32,
-}
+struct Crc {}
 
 impl Crc {
```
phil-levis, line 314: Can you document the distinction between Crc32 and Crc, and why they are separated?

alistair23: Done

bradjc (approved):

alevy (approved):

lschuermann:
I think these changes are fine generally, but I would say that we shouldn't rely on TicKV in any crate just to have access to the CRC32 implementation, but rather provide an alternative implementation which doesn't pull in the entirety of TicKV (regarding making it `pub`).

PR #2809, opened 2021-09-04
Adding App IDs and userspace binary integrity to the Tock kernel

phil-levis:
### Pull Request Overview

This pull request incorporates App IDs into the Tock kernel. It is based on Johnathan's V3 proposal from this [long thread](https://groups.google.com/g/tock-dev/c/aduN7fHWXdI/m/G799efpNBwAJ). The core of the proposal is in this [message](https://groups.google.com/g/tock-dev/c/aduN7fHWXdI/m/9oOx5kfRAgAJ). The details on App IDs and how they work is described in in [doc/reference/trd-appid.md](https://github.com/tock/tock/blob/aaacca0d42128b2c2aa503e1ef6ed719bc689e88/doc/reference/trd-appid.md) in this PR.

This PR requires corresponding PRs for [elf2tab](https://github.com/tock/elf2tab/tree/credentials_support) and [tockloader](https://github.com/tock/tockloader/tree/credentials_support). All three need to be applied simultaneously. The elf2tab modifications support the new required TBF headers and footers. The tockloader modifications can parse these headers/footers so tockloader can properly scan and install binaries. The Tock modifications add support for these headers and footers and the kernel mechanisms for handling them.

Incorporating App IDs into the kernel allows a kernel to verify the integrity and authenticity of userspace binaries and make corresponding decisions on whether load them. The exact policy for userspace binary loading is defined by an implementation of the `AppCredentialsChecker` trait, which is passed to the kernel by `main.rs`. If no checker is provided, then the kernel places no restrictions on which binaries are loaded: it follows the current policy of filling process slots with binaries in the order they appear in the program memory region.

This PR includes two sample `AppCredentialsChecker` implementations. The first, [AppCheckerSimulated](https://github.com/tock/tock/blob/cef462dab5270f1b5336a634217e7a0b5e3367e5/kernel/src/process_checking.rs#L107), simulates checking credentials but always accepts every credential. This causes the kernel to load all processes, as is the standard policy now. The second, [AppCheckerSha256](https://github.com/tock/tock/blob/cef462dab5270f1b5336a634217e7a0b5e3367e5/kernel/src/process_checking.rs#L194) only accepts Sha256 credentials. If a process does not have a valid Sha256 footer, the kernel will not load the process.

[...]

PR #3092, opened 2022-07-18
capsules: Add support for AES GCM

alistair23:
### Pull Request Overview

This PR adds a new aes_gcm capsule. This capsule exposes [AES-GCM](https://en.wikipedia.org/wiki/AES-GCM-SIV) support if the hardware doesn't actually have it.

We implement AES GCM using other supported AES operations and with a software implementation of ghash.

This is similar to the existing virtual_aes_ccm capsule, except without virtulisation.

AES-GCM is commonly used in [TLS](https://blog.mozilla.org/security/2017/09/29/improving-aes-gcm-performance/). It isn't commonly used in embedded systems, but it's still a good example I think.

### Testing Strategy

Running the OpenTitan tests

### TODO or Help Wanted

This PR adds an external dependency!

This PR adds a dependency on the [ghash crate](https://github.com/RustCrypto/universal-hashes/tree/master/ghash) as part of Rust-crypto.

We are seeing more and more requirements for cryptographic operations in Tock, which means we are also seeing more and more software implementations of them.

Tock already has a SHA256 and CRC32 software implementation. It's not unreasonable that RSA becomes more common with the cryptographic app IDs as well.

I wanted to start the discussion about allowing Rust-crypto crates as capsules. Re-implementing a crypto library ourselves seem very risky, so if we do want software implementations Rust-crypto would be the way to go. I'm not convinced one way or the other, I just wanted to get some feedback on how people feel.

The Rust-crypto crates are well respected, have been audited and match the Tock software license

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/src/main.rs
+++ b/boards/opentitan/src/main.rs
@@ -16,6 +16,7 @@
 
 use crate::hil::symmetric_encryption::AES128_BLOCK_SIZE;
 use crate::otbn::OtbnComponent;
+use capsules_aes_gcm::aes_gcm;
 use capsules_core::virtualizers::virtual_aes_ccm;
 use capsules_core::virtualizers::virtual_alarm::{MuxAlarm, VirtualMuxAlarm};
 use capsules_core::virtualizers::virtual_hmac::VirtualMuxHmac;
@@ -79,9 +80,13 @@ static mut TICKV: Option<
         capsules_extra::sip_hash::SipHasher24<'static>,
     >,
 > = None;
-// Test access to AES CCM
-static mut AES: Option<&virtual_aes_ccm::VirtualAES128CCM<'static, earlgrey::aes::Aes<'static>>> =
-    None;
+// Test access to AES
+static mut AES: Option<
+    &aes_gcm::Aes128Gcm<
+        'static,
+        virtual_aes_ccm::VirtualAES128CCM<'static, earlgrey::aes::Aes<'static>>,
+    >,
+> = None;
 // Test access to SipHash
 static mut SIPHASH: Option<&capsules_extra::sip_hash::SipHasher24<'static>> = None;
 // Test access to RSA
@@ -158,7 +163,10 @@ struct EarlGrey {
     rng: &'static capsules_core::rng::RngDriver<'static>,
     aes: &'static capsules_extra::symmetric_encryption::aes::AesDriver<
         'static,
-        virtual_aes_ccm::VirtualAES128CCM<'static, earlgrey::aes::Aes<'static>>,
+        aes_gcm::Aes128Gcm<
+            'static,
+            virtual_aes_ccm::VirtualAES128CCM<'static, earlgrey::aes::Aes<'static>>,
+        >,
     >,
     kv_driver: &'static capsules_extra::kv_driver::KVSystemDriver<
         'static,
@@ -632,9 +640,6 @@ unsafe fn setup() -> (
 
     const CRYPT_SIZE: usize = 7 * AES128_BLOCK_SIZE;
 
-    let aes_source_buffer = static_init!([u8; 16], [0; 16]);
-    let aes_dest_buffer = static_init!([u8; CRYPT_SIZE], [0; CRYPT_SIZE]);
-
     let ccm_mux = static_init!(
         virtual_aes_ccm::MuxAES128CCM<'static, earlgrey::aes::Aes<'static>>,
         virtual_aes_ccm::MuxAES128CCM::new(&peripherals.aes)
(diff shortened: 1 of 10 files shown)


boards/opentitan/src/main.rs
```diff
@@ -75,9 +74,13 @@ static mut TICKV: Option<
...
     >,
 > = None;
-// Test access to AES CCM
-static mut AES: Option<&virtual_aes_ccm::VirtualAES128CCM<'static, earlgrey::aes::Aes<'static>>> =
-    None;
+// Test access to AES
+static mut AES: Option<
+    &virtual_aes_gcm::VirtualAES128GCM<
+        'static,
+        virtual_aes_ccm::VirtualAES128CCM<'static, earlgrey::aes::Aes<'static>>,
```
phil-levis, line 81: Why is GCM building on top of CCM? They are different algorithms.

phil-levis:
> Tock already has a SHA256 and CRC32 software implementation. It's not unreasonable that RSA becomes more common with the cryptographic app IDs as well.
> 
> I wanted to start the discussion about allowing Rust-crypto crates as capsules. Re-implementing a crypto library ourselves seem very risky, so if we do want software implementations Rust-crypto would be the way to go. I'm not convinced one way or the other, I just wanted to get some feedback on how people feel.

This has been discussed in the core call. The general consensus is that Tock should not have any cipher implementations used for confidentiality. E.g., there should not be Tock-specific implementations of AES or RSA. However, having simple software implementations of integrity primitives (e.g., hashes and CRCs) is OK. The expectation is that these are included to allow software testing without introducing dependencies.

The point you bring up, though, that some Tock boards are going to need software implementations of encryption/decryption and we should pull in crates to do so, is a good one. I will put it on the core call agenda.

alistair23:
> This has been discussed in the core call. The general consensus is that Tock should not have any cipher implementations used for confidentiality. E.g., there should not be Tock-specific implementations of AES or RSA. However, having simple software implementations of integrity primitives (e.g., hashes and CRCs) is OK. The expectation is that these are included to allow software testing without introducing dependencies.

This PR only adds a software hash implementation. Does that mean you agree this should be added? Or do you mean Tock is ok with writing it's own implementations?

> The point you bring up, though, that some Tock boards are going to need software implementations of encryption/decryption and we should pull in crates to do so, is a good one. I will put it on the core call agenda.

I was thinking a bit more about this.

Once we have [verified AppIDs](https://github.com/tock/tock/pull/2809) I could see a future where the default research loading mechanism is to try loading signed apps, with an allow fallback. This way we wouldn't break any users (if it's not signed we still load the app) but also provides a lot more test coverage on signed app loading. We can then update the tools to handle this as transparently as possible.

As well as that, there are lots of good Tock use cases that need crypto. LoRaWAN and other wireless communication stacks all use crypto by default.

These would only work if we had a default software crypto implementation though. So I think it's something worth considering

phil-levis:
> This PR only adds a software hash implementation. Does that mean you agree this should be added? Or do you mean Tock is ok with writing it's own implementations?

It does not mean that I agree it should be added. It means that it's something we need to discuss to explore all of the issues in play. Tock (so far) is OK writing its own software implementations of integrity primitives, such as hashes, checksums, and CRCs. However, that doesn't mean that it's the preferred thing to do. 

> I was thinking a bit more about this.
> 
> Once we have [verified AppIDs](https://github.com/tock/tock/pull/2809) I could see a future where the default research loading mechanism is to try loading signed apps, with an allow fallback. This way we wouldn't break any users (if it's not signed we still load the app) but also provides a lot more test coverage on signed app loading. We can then update the tools to handle this as transparently as possible.

This is already supported in the API. But the policies are more complex than this; there's a question of what signing keys are allowed. The current API requires the checker know something about the key, due to the exponent being implicit. We might make the exponent explicit, depending on requirements.

> As well as that, there are lots of good Tock use cases that need crypto. LoRaWAN and other wireless communication stacks all use crypto by default.

True. Most chips (LoRa being an exception) have cryptographic support, though. The issue here is introducing external dependencies into the kernel for narrow use cases. We've resisted external dependencies for a long time, for simplicity. That doesn't mean the kernel won't ever use them, but for now it doesn't.

boards/opentitan/src/main.rs
phil-levis: @alistair23 Can you explain your intended relationship between CCM and GCM?

phil-levis:
Notes from the OT call:

A `virtual_aes_gcm` capsule should virtualize an AES-GCM implementation for multiple clients. The current capsule should decompose the composition of AES-GCM from AES-CCM and virtualization into two separate capsules. The first is a translation that turns an AES implementation (and perhaps a GCM implementation) into AES-GCM. The second virtualizes AES-GCM.

phil-levis:
> Notes from the OT call:
> 
> A `virtual_aes_gcm` capsule should virtualize an AES-GCM implementation for multiple clients. The current capsule should decompose the composition of AES-GCM from AES-CCM and virtualization into two separate capsules. The first is a translation that turns an AES implementation (and perhaps a GCM implementation) into AES-GCM. The second virtualizes AES-GCM.

The fact that the current AES-CCM virtualizer *isn't* decomposed this way and combines translation and virtualization shouldn't be taken as a reference design. The current code started with an early, 5-year-old AES-CCM implementation and extended the capsule to add virtualization.

phil-levis:
@alistair23 Any progress on this?

alistair23:
I'm still waiting for an Ack or Nack about adding external crypto crates

phil-levis:
@alistair23 I will bring it up on Friday just to be sure, but I am very confident that Tock will allow external (high quality, vetted) implementations of cryptographic primitives.

alistair23:
Ok, updated to remove the virtuliser part from this.

Once this is merged we can then look at adding a single AES virtuliser (while also removing the virtulisation from `virtual_aes_ccm.rs`)

alistair23:
Any more comments? I can rebase this if it will be accepted

alistair23:
I have rebased this now that the `ExternalDependencies.md` doc has been merged

phil-levis (dismissed):
Please add a comment explaining the reason behind the CCM/GCM composition (decorator pattern for API, to enable both), as well as a README with the cargo tree from the new crate. Then it's good to go.

alistair23:
Done!

bradjc (approved):

PR #3093, opened 2022-07-18
Change Led Matrix Syscall Capsule for Microbit

TeodoraMiu:
### Pull Request Overview

This pull request changes the way the led matrix driver works for the Microbit. Before this PR, there were two separate led matrix syscall drivers which served, practically, the same purpose. This PR:

- modifies the led_matrix capsule to be used as a service capsule instead of a syscall one
- initializes only the LedD syscall driver for the Microbit and uses the Leds exported via the Led HIL by the Led matrix service driver

### Testing Strategy

### TODO or Help Wanted

Feedback for these changes is much appreciated.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


capsules/src/led_matrix.rs
```diff
@@ -81,18 +81,13 @@
```
alexandruradovici, line 81: Delete the comments about the syscall.

alexandruradovici, line 81: Add a usage example.

alexandruradovici (changes requested):

capsules/src/led_matrix.rs
TeodoraMiu: Done

TeodoraMiu: Done

alexandruradovici (approved):

boards/components/src/led_matrix.rs
```diff
@@ -103,65 +103,79 @@ macro_rules! led_line_component_helper {
...
 
-pub struct LedMatrixComponent<L: 'static + Pin, A: 'static + Alarm<'static>> {
-    col: &'static [&'static L],
-    row: &'static [&'static L],
+pub struct LedMatrixComponent<
+    L: 'static + Pin,
+    A: 'static + Alarm<'static>,
+    const ROWS: usize,
+    const COLS: usize,
+    const LED_BITS: usize,
```
alexandruradovici, line 152:
```suggestion
    const LED_BYTES: usize,
```

capsules/src/led_matrix.rs
```diff
@@ -1,125 +1,118 @@
...
         alarm: &'a A,
         col_activation: ActivationMode,
         row_activation: ActivationMode,
         refresh_rate: usize,
     ) -> Self {
         // Initialize all LEDs and turn them off
-        if (buffer.len() * 8) < cols.len() * rows.len() {
-            panic!("Matrix LED Driver: provided buffer is too small");
-        }
+        assert!(ROWS * COLS / 8 + 1 == LED_BITS);
```
alexandruradovici, line 115: Document this assertion, explain that this is used until const generics https://github.com/rust-lang/rust/issues/76560 will be stable.

capsules/src/led_matrix.rs
```diff
@@ -258,65 +251,46 @@ impl<'a, L: Pin, A: Alarm<'a>> LedMatrixDriver<'a, L, A> {
...
+impl<'a, L: Pin, A: Alarm<'a>, const ROWS: usize, const COLS: usize, const LED_BITS: usize>
+    LedMatrixLed<'a, L, A, ROWS, COLS, LED_BITS>
+{
+    pub fn new(
+        matrix: &'a LedMatrixDriver<'a, L, A, ROWS, COLS, LED_BITS>,
+        col: usize,
+        row: usize,
+    ) -> Self {
         if col >= matrix.cols_len() || row >= matrix.rows_len() {
             panic!("LED at position ({}, {}) does not exist", col, row);
```
alexandruradovici, line 285: Delete panic and return an `Option<Self>` that is None if the LED is not valid.

boards/components/src/led_matrix.rs
```diff
@@ -75,14 +75,14 @@ macro_rules! led_matrix_component_helper {
...
-        static mut BUFFER: [u8; NUM_COLS*NUM_ROWS / 8 + 1] = [0; NUM_COLS*NUM_ROWS / 8 + 1];
+        static mut BUFFER: [u8; NUM_ROWS*NUM_COLS / 8 + 1] = [0; NUM_ROWS*NUM_COLS / 8 + 1];
         components::led_matrix::LedMatrixComponent::new ($alarm_mux, $crate::led_line_component_helper!($Pin, $($C,)+), $crate::led_line_component_helper!($Pin, $($R,)+), $col_active, $row_active, $refresh_rate, &mut BUFFER)
     };};
 }
 
 #[macro_export]
 macro_rules! led_line_component_helper {
-    ($Pin:ty, $($L:expr),+ $(,)?) => {{
+    ($Pin:ty, $($L:expr),+, $(,)?) => {{
```
alexandruradovici, line 85: Merge this with the `led_matrix_component_buf`, as that one computes all the constant values.

boards/microbit_v2/src/main.rs
```diff
@@ -44,6 +44,9 @@ const UART_TX_PIN: Pin = Pin::P0_06;
 const UART_RX_PIN: Pin = Pin::P1_08;
 
 /// LED matrix
+const ROWS: usize = 5;
+const COLS: usize = 5;
+const LED_BITS: usize = ROWS * COLS / 8 + 1;
```
alexandruradovici, line 49:
```suggestion

```

The values should be computed by the macro.

alexandruradovici (changes requested):

bradjc (approved):

TeodoraMiu:
I believe this is good to go.

hudson-ayers (commented):
I am a bit confused by this PR. Are any of the helper functions in `capsules/src/led_matrix.rs` even used anymore? What is gained from the added indirection of instantiating the led_matrix capsule if userspace is just going to use the interface of the normal LED capsule?

I may be missing something obvious here

alexandruradovici:
This PR provides the ability of using the standard LED driver even if LEDs are actually connected in a matrix that can only display one column or line at a time. Before the PR, we had a syscall driver specifically for the matrix that was identical in API with the LED driver.

The timeline that got us to this PR is:

1. When micro:bit was added, we had a led matrix sycall driver that was handling both the matrix control and system calls (identical to the LED syscall driver). This resulted in code and API duplication, as we now had two LED syscall drivers.
2. While using the micro:bit, we came across the need to use single LEDs from the matrix from within another driver. This made us add code that exposes individual LEDs from the matrix through the LED HIL.
3. As the matrix driver exposes each LED through the LED HIL, we realized that we could convert the matrix syscall driver into a driver that only controls the actual matrix and use the LED syscall driver, while removing duplicate code and API.

hudson-ayers (approved):
Thank you for the explanation, that makes sense.

bors r+

Issue #1447, opened 2019-11-04
Soundness of static_init macro?


gendx:
The current static_init implementation relies on `static mut` memory and heavily unsafe transmuting of memory to initialize it.

https://github.com/tock/tock/blob/0be31eac710967b6376a637be339e7e883498b24/kernel/src/common/utils.rs#L30-L33

From a soundness point-of-view, this seems problematic for several reasons.

First, `static mut` seems inherently hard to use correctly:
- https://users.rust-lang.org/t/why-is-static-mut-bad/32888 
- https://github.com/rust-lang/rust/issues/53639
- https://github.com/rust-lang-nursery/lazy-static.rs/issues/117

Second, Tock seems to be aliasing mutable memory, which is against Rust's basic memory-safety rules. For example, unless I missed something, the debugger's `OUTPUT_BUF` and `INTERNAL_BUF` are aliased mutably multiple times.

https://github.com/tock/tock/blob/8c4baa27c29967eff577c472b2897b55cef17188/kernel/src/debug.rs#L226-L227

First, in the console component.

https://github.com/tock/tock/blob/8c4baa27c29967eff577c472b2897b55cef17188/boards/components/src/console.rs#L71-L78

Second, in the process_console component.

https://github.com/tock/tock/blob/8c4baa27c29967eff577c472b2897b55cef17188/boards/components/src/process_console.rs#L66-L73

Third, in each board's setup.

https://github.com/tock/tock/blob/8c4baa27c29967eff577c472b2897b55cef17188/boards/nordic/nrf52dk_base/src/lib.rs#L252-L259

In this case, the worst that could happen is likely debug messages being interleaved, repeated or corrupted, which is probably not terrible, but this seems to point out an inherent issue with how unsafe the `static_init` API really is, and how easy it is to get it wrong. There could be other cases where aliased mutable memory is more problematic.

Has the [lazy_static](https://crates.io/crates/lazy_static) crate been considered to initialize static memory? It seems to provide the functionality that Tock needs, but I would be much more confident with it, because it is widely used (16M+ downloads) and has undergone much more scrutiny than Tock's home-made unsafe `static_init` mechanism. `lazy_static` supports `no_std` and only depends on [spin](https://crates.io/crates/spin) to provide a locking mechanism (as far as I understand, this is used to avoid any data races when initializing the static memory).

niklasad1:
Please see https://github.com/tock/tock/blob/master/doc/Soundness.md

[...]

PR #3095, opened 2022-07-19
arm/cortex-m/syscall: replace volatile userspace RAM accesses with regular pointer reads/writes

lschuermann:
### Pull Request Overview

Given the changes introduced in the previous commits (introducing the CortexM subarch trait), which cause the Rust compiler to have direct visibility of the `#[naked]` functions and `asm!` macros used in handling interrupts and switching to userspace, it is obliged to insert a compiler fence before executing these blocks which potentially modify userspace memory.

Hence, volatile accesses to these memory regions are not actually needed, given that we only have a single thread of execution and retain total ordering of accesses in hardware.

This commit thus changes volatile memory accesses for regular accesses through pointers.

### Testing Strategy

To be tested! What's the best way to test this? Look at the LLVM-IR?

### TODO or Help Wanted

Blocked on #3080.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/syscall.rs
+++ b/arch/cortex-m/src/syscall.rs
@@ -5,8 +5,7 @@ use core::fmt::Write;
 use core::marker::PhantomData;
 use core::mem::{self, size_of};
 use core::ops::Range;
-use core::ptr::{read_volatile, write_volatile};
-
+use core::ptr::{self, read_volatile, write_volatile};
 use kernel::errorcode::ErrorCode;
 
 use crate::CortexMVariant;
@@ -228,13 +227,13 @@ impl<A: CortexMVariant> kernel::syscall::UserspaceKernelBoundary for SysCall<A> 
         //  - Instruction addresses require `|1` to indicate thumb code
         //  - Stack offset 4 is R12, which the syscall interface ignores
         let stack_bottom = state.psp as *mut usize;
-        write_volatile(stack_bottom.offset(7), state.psr); //......... -> APSR
-        write_volatile(stack_bottom.offset(6), callback.pc | 1); //... -> PC
-        write_volatile(stack_bottom.offset(5), state.yield_pc | 1); // -> LR
-        write_volatile(stack_bottom.offset(3), callback.argument3); // -> R3
-        write_volatile(stack_bottom.offset(2), callback.argument2); // -> R2
-        write_volatile(stack_bottom.offset(1), callback.argument1); // -> R1
-        write_volatile(stack_bottom.offset(0), callback.argument0); // -> R0
+        ptr::write(stack_bottom.offset(7), state.psr); //......... -> APSR
+        ptr::write(stack_bottom.offset(6), callback.pc | 1); //... -> PC
+        ptr::write(stack_bottom.offset(5), state.yield_pc | 1); // -> LR
+        ptr::write(stack_bottom.offset(3), callback.argument3); // -> R3
+        ptr::write(stack_bottom.offset(2), callback.argument2); // -> R2
+        ptr::write(stack_bottom.offset(1), callback.argument1); // -> R1
+        ptr::write(stack_bottom.offset(0), callback.argument0); // -> R0
 
         Ok(())
     }
@@ -279,20 +278,20 @@ impl<A: CortexMVariant> kernel::syscall::UserspaceKernelBoundary for SysCall<A> 
             // syscall (i.e. we return a value to the app immediately) then this
             // will have no effect. If we are doing something like `yield()`,
             // however, then we need to have this state.
-            state.yield_pc = read_volatile(new_stack_pointer.offset(6));
-            state.psr = read_volatile(new_stack_pointer.offset(7));
+            state.yield_pc = ptr::read(new_stack_pointer.offset(6));
+            state.psr = ptr::read(new_stack_pointer.offset(7));
 
             // Get the syscall arguments and return them along with the syscall.
             // It's possible the app did something invalid, in which case we put
             // the app in the fault state.
-            let r0 = read_volatile(new_stack_pointer.offset(0));
-            let r1 = read_volatile(new_stack_pointer.offset(1));
-            let r2 = read_volatile(new_stack_pointer.offset(2));
-            let r3 = read_volatile(new_stack_pointer.offset(3));
+            let r0 = ptr::read(new_stack_pointer.offset(0));
+            let r1 = ptr::read(new_stack_pointer.offset(1));
+            let r2 = ptr::read(new_stack_pointer.offset(2));
+            let r3 = ptr::read(new_stack_pointer.offset(3));
 
             // Get the actual SVC number.
-            let pcptr = read_volatile((new_stack_pointer as *const *const u16).offset(6));
-            let svc_instr = read_volatile(pcptr.offset(-1));
+            let pcptr = ptr::read((new_stack_pointer as *const *const u16).offset(6));
+            let svc_instr = ptr::read(pcptr.offset(-1));
             let svc_num = (svc_instr & 0xff) as u8;
 
             // Use the helper function to convert these raw values into a Tock
@@ -336,14 +335,14 @@ impl<A: CortexMVariant> kernel::syscall::UserspaceKernelBoundary for SysCall<A> 
                 0xBAD00BAD,
             )
         } else {
-            let r0 = read_volatile(stack_pointer.offset(0));
-            let r1 = read_volatile(stack_pointer.offset(1));
-            let r2 = read_volatile(stack_pointer.offset(2));
-            let r3 = read_volatile(stack_pointer.offset(3));
-            let r12 = read_volatile(stack_pointer.offset(4));
-            let lr = read_volatile(stack_pointer.offset(5));
-            let pc = read_volatile(stack_pointer.offset(6));
-            let xpsr = read_volatile(stack_pointer.offset(7));
+            let r0 = ptr::read(stack_pointer.offset(0));
+            let r1 = ptr::read(stack_pointer.offset(1));
+            let r2 = ptr::read(stack_pointer.offset(2));
+            let r3 = ptr::read(stack_pointer.offset(3));
+            let r12 = ptr::read(stack_pointer.offset(4));
+            let lr = ptr::read(stack_pointer.offset(5));
+            let pc = ptr::read(stack_pointer.offset(6));
+            let xpsr = ptr::read(stack_pointer.offset(7));
             (r0, r1, r2, r3, r12, lr, pc, xpsr)
         };
 


alistair23:
I'm not sure I love this.

You are saying because the compiler is inserting a fence, we don't need to access the volatile memory as volatile? What if one day the compiler doesn't insert a fence? Either because it changes or someone changes Tock code. Or more likely, what if one day Tock adds multi-thread support?

It seems like volatile accesses are the correct thing to do here, and we are just trying to avoid using them. And this will be very difficult to debug in the future.

I'm assuming that not using volatile could save us a few instructions in the syscall path, which might be a trade off worth making though. Do you have an idea of savings we get here?

lschuermann:
> It seems like volatile accesses are the correct thing to do here, and we are just trying to avoid using them.

@alistair23 I disagree with this argument. [Rust's documentation](https://doc.rust-lang.org/std/ptr/fn.read_volatile.html) has the following to say about volatile:

> Volatile operations are intended to act on I/O memory, and are guaranteed to not be elided or reordered by the compiler across other volatile operations. [...] Just like in C, whether an operation is volatile has no bearing whatsoever on questions involving concurrent access from multiple threads. Volatile accesses behave exactly like non-atomic accesses in that regard. In particular, a race between a `read_volatile` and any write operation to the same location is undefined behavior.

While volatile accesses are undoubtedly the right tool to use in many circumstances around the Tock codebase, specifically when proper memory is being modified by the underlying hardware (or rather by anything not in the current hardware-thread of execution), or for memory-mapped peripherals, for interactions between userspace and kernel space it is not. Even though we are executing two independent programs and switching between these software threads of execution, all proper instructions issued by either of those two programs still retain their order and effects relative to each other because they are issued within the same hardware thread of execution. A write-instruction issued by userspace before a context switch is always observable by kernelspace (running on this same core) regardless of using volatile accesses or not. The only and real danger we face here is that a compiler may choose to reorder reads and writes in a way which changes their order relative to context switches. Continuing in this argument, volatile accesses actually might not help us at all in this regard, given that volatile accesses _"are [only] guaranteed to not be elided or reordered by the compiler across other volatile operations."_ The context switch itself is performed via an opaque `asm!` block, and so **if** the compiler may move reads and writes beyond this block, theoretically it may do the same for volatile reads and writes.

In this argument, I conclude that the right tool to use here is a compiler fence, as it guarantees us that memory accesses are not reordered over this well-defined point (according to the semantics of the particular fence used). We do not specify the `nomem` assembly option, [which states that](https://rust-lang.github.io/rfcs/2873-inline-asm.html) _"the asm blocks does not read or write to any memory. This allows the compiler to cache the values of modified global variables in registers across the asm block since it knows that they are not read or written to by the asm."_ This should be sufficient for us to assure that no reordering of accesses to globals, variables explicitly exposed to this `asm!` block, and memory of which Rust cannot reason about (e.g. userspace RAM) should not be subject to reordering over such a block.

ppannuto:
If we can, I think I would like to separate this in to two threads of discussion, as I think there are two different cases here.

We have some uses here, e.g. `SYSCALL_FIRED`, `APP_HARD_FAULT`, that are global flag variables that we're using `volatile` to interface with -- these should probably be moved over to some proper atomic / sync primitive.

The other case is the process stack pointer. That one I need to think about a little bit more; in particular, the memory it points to _is_ being written by hardware outside the knowledge of the executing code (i.e. the exception entry/return writes/reads from the stack in HW). While my first instinct is that, yes, the `asm` block elsewhere makes this work, that assembly is far-removed from the memory operations here. Relying on that to ensure correctness feels fragile. Semantically, I believe volatile is correct here: we don't actually care about the order of the writes to the stack, but we do need to read/write it completely, as its contents has changed by hardware operations outside the visibility of the executing software.

lschuermann:
Those are some good points, but I am still not 100% convinced that volatile is the right tool here for the guarantees we need, specifically for accessing registers transferred via the stack in case of a system call and the return values respectively. Perhaps I'm missing something, going to talk about this with @ppannuto on Friday.

ppannuto:
FWIW, I think that `volatile` is necessary, but I am not arguing that `volatile` alone is sufficient – we can chat more on the latter tomorrow.

lschuermann:
@ppannuto and I talked about this yesterday. It is a complex issue, but there seems to be a substantial line of argument that volatile is not required, at least for a subset of the operations performed here. In the following, I am going to focus solely on the hardware stacked registers, so accesses such as the following:

https://github.com/tock/tock/blob/f042be27c84ba3634475c3105302d00f5f70f6ae/arch/cortex-m/src/syscall.rs#L288-L291

It is important to understand what happens here conceptually: upon a context switch (which is indicated to the kernel via an exception), the CPU implements the ARM Architecture Procedure Call Standard (AAPCS) by hardware-stacking a set of registers, namely those which are saved by the calling function as per AAPCS. This makes the calling convention of exception handlers essentially equivalent to that of any other C function following the AAPCS. Because we trigger our exception handlers not only for explicit system calls, but also to preempt userspace or as a result of faults, we're also saving the callee-saved registers (`r4` to `r11`) in the process state.

In our context switching code, we're first restoring user-space callee-saved registers (`ldmia`), executing the context switch via the `svc` instruction which will further unstack the caller-saved registers resident on the bottom of the process stack, execute userspace code, and upon an exception return to kernel-space by hardware-stacking caller-saved registers, and finally storing callee-saved registers (`stmia`). While it is true that, on entry of kernel space, the caller-saved registers are technically written by neither user-space nor kernel-space but the hardware, any writes performed by the hardware as part of this stacking/unstacking are immediately visible core-locally. From the kernel's perspective, these stacked registers could've just as well come from proper store-instructions from userspace. They don't originate from the same thread of execution as far as the kernel binary is concerned, but they are part of the ordered sequence of instructions issued to the local CPU.

It is also important to establish that these stacked registers are just conventional RAM. Writing or reading them does not have any side effects. What _is_ important is that all required store-instructions required to update the userspace register set are issued before the `svc` instruction is, and that upon returning to userspace, the kernel must assume these memory locations have changed. We do not care about the number or order of stacked register writes before the `svc` call, and we do not care about the number or order of stacked register reads after the `svc` call, but the order of instructions issued relative to the `svc` call.

As discussed above, `volatile` thus does not appear the write tool here: volatile just guarantees the number of read/write instructions generated, and the ordering of volatile reads and writes _only_ relative to _other volatile_ reads and writes. This has been discussed further [here](https://github.com/tock/tock/issues/1447#issuecomment-661959672). Instead, [`compiler_fence`s](https://doc.rust-lang.org/std/sync/atomic/fn.compiler_fence.html) can be used to "[restrict] the kinds of memory re-ordering the compiler is allowed to do. Specifically, depending on the given [Ordering](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html) semantics, the compiler may be disallowed from moving reads or writes from before or after the call to the other side of the call to compiler_fence."

I would like to argue (although I am less certain about this particular point) that, assuming we do not have any unsoundness relevant to context switching (which we do currently), a proper `asm!` block visible to the compiler and not specifying the `nomem` flag is already sufficient to effectively provide the guarantees and semantics of a compiler fence, at least for the reads and writes outside of the kernel's execution thread (i.e. stacked registers + user-space memory generally). Rust's documentation on the `asm!` macro specifies the following for the `nomem` flag:

> The asm! blocks does not read or write to any memory. This allows the compiler to cache the values of modified global variables in registers across the asm! block since it knows that they are not read or written to by the asm!.

Assuming the inverse holds when `nomem` is not passed, this would mean that the compiler must assume that global variables have been modified by the assembly and cannot rely on them not having changed. The term "global variables" is a bit of a stretch here, really, given that the user-space memory we are accessing is "alien" memory and does not fit into the context of Rust's memory model at all. The above also does not strictly specify that all prior writes to such global variables must be visible to the `asm!`, although I do presume that this is implied.

I have performed some very unscientific tests issuing non-volatile writes to Rust allocations across the assembly block, as well as to memory accessed through the userspace stack pointer. Indeed, in the generated LLVM-IR, writes to proper Rust variables can be observed to be reordered across the assembly block (or rather elided entirely), whereas writes to user-space memory are not. The assembly block in the LLVM-IR has the `~{memory}` clobber, which "indicates that the assembly writes to arbitrary undeclared memory locations – not only the memory pointed to by a declared indirect output" and hence LLVM should also not move reads/writes across this block in later optimization stages.

To summarize the above wall of text, I am quite confident that we do not need volatile accesses to these stacked registers, although I would like to see more evidence that the `asm!` provides a sufficient "effective compiler fence" concerning accesses to userspace memory. If we do not find evidence for that, we should likely insert an explicit [`compiler_fence`](https://doc.rust-lang.org/std/sync/atomic/fn.compiler_fence.html) in our codebase.

lschuermann:
Marking as blocked until we've reached a consesus on how to continue here. Also going to revert the volatile accesses to flags such as `APP_HARD_FAULT`, which I suspect are problematic on their own, but an entirely different issue.

ppannuto:
After the conversation that Leon and I had last week, I agree that volatile should not be necessary (or used) for the [un]stacking. I do still think we should split this out from the globals, as suggested.

---

Leon: Not for this PR, but I just saw this newer `asm!` feature: https://doc.rust-lang.org/nightly/reference/inline-assembly.html#abi-clobbers – looks like they're adding explicit clobber knowledge to assembly blocks; might make sense to look into this for some of the actual `asm!` blocks (looks like [libtock-rs is already using this](https://github.com/tock/libtock-rs/blob/39fec6b59a2385fafcc5ae4b52dd19be1d96c7e4/runtime/src/syscalls_impl_arm.rs#L20), though I wonder why that is `C` rather than `aapcs`..., will dig in there later)

bradjc (dismissed):
I have no thoughts on this, but convincing @ppannuto is a high bar :)

lschuermann:
As promised, I have reverted to volatile accesses to flags such as `APP_HARD_FAULT` for now, so this should be good to merge. However, I'd suggest deferring until after releasing 2.1 -- the `volatile` currently does not do any harm, and it'd be good to have these changes in-tree for some time before making a release. Keeping `blocked`.

lschuermann:
> Leon: Not for this PR, but I just saw this newer `asm!` feature: https://doc.rust-lang.org/nightly/reference/inline-assembly.html#abi-clobbers – looks like they're adding explicit clobber knowledge to assembly blocks; might make sense to look into this for some of the actual `asm!` blocks (looks like [libtock-rs is already using this](https://github.com/tock/libtock-rs/blob/39fec6b59a2385fafcc5ae4b52dd19be1d96c7e4/runtime/src/syscalls_impl_arm.rs#L20), though I wonder why that is `C` rather than `aapcs`..., will dig in there later)

@ppannuto This is a good point. I've seen the discussion on tock/libtock-rs#426 and I'm unsure about what the consequences for Tock should be, if any? When ARM exceptions clearly use the AAPCS, should our exception handler signature instead be `unsafe extern "aapcs" fn generic_isr_arm_v7m()` (which is valid Rust)? How does this mesh with system calls, which are by definition following the C calling convention, but with the CPU still performing hardware stacking as per AAPCS? The impact of all of this is low for sure, given that `C` and `aapcs` are identical now and very likely will remain so. But we might as well try and be pedantic about this.

ppannuto (approved):
bors d+

This more limited patchset looks good, and can be merged after 2.1 release

ppannuto:
re exception handler signature: we can probably pull this discussion to a separate PR, but yeah, I think the 'right' thing to do is to mark them as `aapcs` with a fairly extensive comment explaining what's going on. From the kernel's perspective, the thing triggering kernel entry will always be hardware which is following the hardware calling convention; that will always be true. For tooling convenience, libtock-rs can safely keep using the `C` nomenclature, as they're identical.

lschuermann:
Let's see whether the world breaks with this in.

Issue #3098, opened 2022-07-20
nRF52840-DK's Board Configuration Leaves 63.5Mb of storage unavailable


TheConner:
The nRF52840-DK's storage is configured [here](https://github.com/tock/tock/blob/master/boards/nordic/nrf52840dk/src/main.rs#L570-L573), partitioning out `0-0x60000` (384 Kb) for the kernel space storage region, and then `0x60000-0x80000` (128 Kb) for userspace storage region. Using the [nonvolatile storage test application](https://github.com/tock/libtock-c/tree/master/examples/tests/nonvolatile_storage) on the nRF52840-DK, I see only 128 Kb for userspace which is expected under this configuration.

This amount of storage makes sense if Tock was only using the nRF52840's flash storage (1MB total, iirc); however, the nRF52840-DK has 64Mb of [external storage](https://infocenter.nordicsemi.com/topic/ug_nrf52840_dk/UG/dk/hw_external_memory.html) on a [MX25R6435F SPI/QSPI flash chip](https://www.macronix.com/Lists/Datasheet/Attachments/8746/MX25R6435F,%20Wide%20Range,%2064Mb,%20v1.5.pdf). It seems that the `NonvolatileStorageComponent` for the nRF52840-DK [is configured to use the external flash chip](https://github.com/tock/tock/blob/master/boards/nordic/nrf52840dk/src/main.rs#L566-L569), so it should define a larger userspace flash region (or both userspace and kernel space) to enable access to the other 63.5 Mb of flash storage.

My suggestion is to increase the flash regions as follows, to fill all 64Mb of external flash (will follow up with a PR for this)

Kernel: 0-0x60000 (384 Kb, the same amount)
Userspace: 0x60000-0x4000000 (63.625 Mb, fill the rest of available flash)

hudson-ayers:
Fixed by #3099 . Thanks!

PR #3099, opened 2022-07-20
Enable all storage on nRF52840-DK #3098

TheConner:
### Pull Request Overview

This PR increases the userspace storage of the nRF52840-DK (re #3098) to use all of the available external flash on the board. New flash regions are as follows:

Kernel: 0-0x60000 (384 Kb)
Userspace: 0x60000-0x4000000 (63.625 Mb)
For a total of 64Mb.

### Testing Strategy

I built and installed the kernel on my nRF52740-DK, and used the  [nonvolatile storage test application](https://github.com/tock/libtock-c/tree/master/examples/tests/nonvolatile_storage) to see how much storage is reported. All tests suceeded, and the correct amount of storage (66715648 bytes, 63.625Mb) is reported.

### TODO or Help Wanted

Regarding testing, ideally there would be a test application to ensure all the flash is functioning correctly, instead of only 512 bytes. 

### Documentation Updated

- [x] N/A

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nordic/nrf52840dk/src/main.rs
+++ b/boards/nordic/nrf52840dk/src/main.rs
@@ -567,10 +567,10 @@ pub unsafe fn main() {
         board_kernel,
         capsules::nonvolatile_storage_driver::DRIVER_NUM,
         mx25r6435f,
-        0x60000, // Start address for userspace accessible region
-        0x20000, // Length of userspace accessible region
-        0,       // Start address of kernel region
-        0x60000, // Length of kernel region
+        0x60000,   // Start address for userspace accessible region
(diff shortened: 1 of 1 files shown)


hudson-ayers:
Thanks for this PR, this looks correct to me. Would you mind testing the nonvolatile storage app with some large offset, such that we can verify writing outside the region that we currently have made accessible works?

hudson-ayers (approved):
Thanks for testing! If you wanted to contribute that change to the `nonvolatile_storage` test to `libtock_c` I would be happy to accept that as well.

alistair23 (approved):

bradjc (approved):

PR #3106, opened 2022-07-26
Seven Segment Display Capsule

TeodoraMiu:
### Pull Request Overview

This pull request adds a capsule for controlling a 7 segment digit display. This PR:

- defines a new driver that is able to represent numbers, dots and custom patterns on a 7 segment display made of one or more digits

### Testing Strategy

Functionality demonstration: https://youtu.be/DH0zL62WDuo

### TODO or Help Wanted

Feedback is much appreciated.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/seven_segment.rs
+++ b/capsules/src/seven_segment.rs
@@ -9,7 +9,25 @@
 //! Usage
 //! -----
 //!
+//! Cannot use simultaneously with LED matrix.
+//! Example of use for a display with 4 digits and the Microbit: 
+//! Microbit Pins: https://tech.microbit.org/hardware/schematic/
+//! 4 digit 7 segment display pinout: https://www.dotnetlovers.com/images/4digit7segmentdisplay85202024001AM.jpg
+//!
 //! ```rust
+//! const NUM_DIGITS: usize = 4;
+//! const DIGITS: [Pin; 4] = [Pin::P1_02, Pin::P0_12, Pin::P0_30, Pin::P0_09]; // [D1, D2, D3, D4]
+//! const SEGMENTS: [Pin; 7] = [
+//!     Pin::P0_02, // A
+//!     Pin::P0_03, // B
+//!     Pin::P0_04, // C
+//!     Pin::P0_31, // D
+//!     Pin::P0_28, // E
+//!     Pin::P0_10, // F
+//!     Pin::P1_05, // G
+//! ];
+//! const DOT: Pin = Pin::P0_11;
+//!
 //! let segment_array = static_init!(
 //!     [&'static nrf52::gpio::GPIOPin<'static>; 8],
 //!     [
--- a/capsules/src/seven_segment.rs
+++ b/capsules/src/seven_segment.rs
@@ -2,14 +2,13 @@
 //!
 //! This capsule was developed using the following components:
 //! - Microbit_v2
-//! - Edge Connector Breakout Board for Microbit
-//! - 7 segment display with 4 digits
+//! - Edge Connector Breakout Board for Microbit (PPMB00126)
+//! - 7 segment display with 4 digits (3461BS-1)
 //! - breadboard, 220 ohms resistances and jump wires
 //!
 //! Usage
 //! -----
 //!
-//! Cannot use simultaneously with LED matrix.
 //! Example of use for a display with 4 digits and the Microbit:
 //! Microbit Pins: https://tech.microbit.org/hardware/schematic/
 //! 4 digit 7 segment display pinout: https://www.dotnetlovers.com/images/4digit7segmentdisplay85202024001AM.jpg
@@ -180,9 +179,10 @@ const DIGITS: [u8; 10] = [
 /// Holds an array of digits and an array of segments for each digit.
 
 pub struct SevenSegmentDriver<'a, P: Pin, A: Alarm<'a>, const NUM_DIGITS: usize> {
-    /// An array of 8 segments (7 for digit segments and one dot segment)
+    /// An array of 8 segment pins (7 for digit segments and one dot segment)
     segments: &'a [&'a P; 8],
-    /// An array of `NUM_DIGITS` digits
+    /// An array of `NUM_DIGITS` digit pins, each one corresponding to one digit on the display
+    /// For each digit selected, a pattern of lit and unlit segments will be represented
     digits: &'a [&'a P; NUM_DIGITS],
     /// A buffer which contains the patterns displayed for each digit
     /// Each element of the buffer array represents the pattern for one digit, and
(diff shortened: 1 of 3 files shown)


capsules/src/seven_segment.rs
```diff
@@ -0,0 +1,371 @@
...
+//!         kernel::hil::gpio::ActivationMode::ActiveLow,
+//!         kernel::hil::gpio::ActivationMode::ActiveHigh,
+//!         60
+//!     ),
+//! );
+//!
+//! virtual_alarm_digit.set_alarm_client(digit_display);
+//!
+//! digit_display.init();
+//!
```
bradjc, line 89: Missing "```"

capsules/src/seven_segment.rs
```diff
@@ -0,0 +1,371 @@
...
+    0b01111111, // 8
+    0b01101111, // 9
+];
+
+/// Holds an array of digits and an array of segments for each digit.
+
+pub struct SevenSegmentDriver<'a, P: Pin, A: Alarm<'a>, const NUM_DIGITS: usize> {
+    /// An array of 8 segments (7 for digit segments and one dot segment)
+    segments: &'a [&'a P; 8],
+    /// An array of 4 digits
```
bradjc, line 159:
```suggestion
    /// An array of `NUM_DIGITS` digits
```

capsules/src/seven_segment.rs
```diff
@@ -0,0 +1,371 @@
+//! Provides userspace access to 7 segment digit displays.
+//!
```
bradjc, line 2: Can you add what hardware this was developed with?

capsules/src/seven_segment.rs
```diff
@@ -0,0 +1,371 @@
...
+    /// An array of 8 segments (7 for digit segments and one dot segment)
+    segments: &'a [&'a P; 8],
+    /// An array of 4 digits
+    digits: &'a [&'a P; NUM_DIGITS],
+    /// A buffer which contains the patterns displayed for each digit
+    /// Each element of the buffer array represents the pattern for one digit, and
+    /// is a sequence of bits that have the value 1 for a lit segment and the value 0 for
+    /// an unlit segment.
+    buffer: TakeCell<'a, [u8; NUM_DIGITS]>,
+    alarm: &'a A,
```
bradjc, line 166: What is the alarm for? I don't think we can have a purely synchronous interface and use an alarm.

TeodoraMiu: The alarm is for the refresh rate of the display. Since the digits and the segments are multiplexed, not all digits can be represented simultaneously, so I used the alarm to switch between digits after a short amount of time, to appear as if all digits are being shown at the same instance.

capsules/src/seven_segment.rs
TeodoraMiu: Fixed

capsules/src/seven_segment.rs
TeodoraMiu: Fixed

capsules/src/seven_segment.rs
TeodoraMiu: Done

bradjc: Is there some sort of switch? How is the digit selected?

TeodoraMiu: There is one pin for each digit. The buffer I used is an array containing NUM_DIGITS elements, each element being a pattern corresponding to one digit. The driver goes through this array constantly, to represent each digit one after the other (but appearing simultaneous to the human eye), similarly to how the led matrix driver for the Microbit works. So there isn't necessarily one digit selected at a time, rather every digit is in use successively. If one of the elements in the buffer remains unchanged (0b00000000), that means that none of the segments on that specific digit are on, but the driver will go over that element anyway and represent nothing every time it loops back, until the buffer is changed.

bradjc: Is this a specific product you can link to, or tutorial? How would someone else use this driver?

TeodoraMiu: I added the pin connections in the documentation: https://github.com/WyliodrinEmbeddedIoT/tock/blob/f36f2c983e5aa9353aa562ef1e750ebe16496068/capsules/src/seven_segment.rs#L12

capsules/src/seven_segment.rs
```diff
@@ -0,0 +1,397 @@
...
+//! This capsule was developed using the following components:
+//! - Microbit_v2
+//! - Edge Connector Breakout Board for Microbit
+//! - 7 segment display with 4 digits
+//! - breadboard, 220 ohms resistances and jump wires
+//!
+//! Usage
+//! -----
+//!
+//! Cannot use simultaneously with LED matrix.
```
bradjc, line 12: Why not?

capsules/src/seven_segment.rs
```diff
@@ -0,0 +1,397 @@
...
+    0b01111111, // 8
+    0b01101111, // 9
+];
+
+/// Holds an array of digits and an array of segments for each digit.
+
+pub struct SevenSegmentDriver<'a, P: Pin, A: Alarm<'a>, const NUM_DIGITS: usize> {
+    /// An array of 8 segments (7 for digit segments and one dot segment)
+    segments: &'a [&'a P; 8],
+    /// An array of `NUM_DIGITS` digits
```
bradjc, line 185: Can this explain what these "digits" pins need to do?

capsules/src/seven_segment.rs
bradjc: Do you have a part number for the display? Capsules generally reference specific parts that they were initially tested with.

capsules/src/seven_segment.rs
TeodoraMiu: This was an error. I meant to say that it cannot be used simultaneously with LED matrix on the Microbit because of lack of pins. I will remove this statement.

capsules/src/seven_segment.rs
TeodoraMiu: I added the part numbers in the documentation.

TeodoraMiu:
I think this is good to go.

capsules/src/seven_segment.rs
```diff
@@ -0,0 +1,397 @@
...
+
+        Self {
+            segments,
+            digits,
+            buffer: TakeCell::new(buffer),
+            alarm,
+            segment_activation: segment_activation,
+            digit_activation: digit_activation,
+            current_digit: Cell::new(0),
+            timing: (1000 / (refresh_rate * digits.len())) as u8,
```
brghena, line 223: Why not make `timing` a `u32` by default? The refresh rate is _usually_ high enough that `timing` should be less than 255 ms, but there's no guarantee of that in the equation here.

brghena: You could go even further and just store `interval` instead of `timing`, since there's no option to change refresh rate anyways.

brghena (approved):
One nit here, but otherwise this looks great to me.

bradjc (approved):

PR #3110, opened 2022-07-26
Initial VirtIO support for VirtIO Net & RNG on `qemu_rv32_virt` target

lschuermann:
### Pull Request Overview

This PR adds VirtIO support for the VirtIO `NetworkCard` and `EntropySource` device types, running on the `qemu_rv32_virt` board to be introduced with #2516. Next to the LiteX boards, VirtIO is yet another way towards Ethernet support within Tock. Along with LiteX and probably one other board (perhaps the i.MX RT1060 EVK) it can serve as an example for what Ethernet controllers can look like, to define an Ethernet device HIL in the near term.

Apart from that, VirtIO seems to be generally useful as a standardized interface towards virtualized (and even some non-virtualized!) peripherals and is portable between different emulators and simulation environments. Together with the QEMU RISC-V 32bit `virt` target, it can serve as a stable base to improve on Tock's virtualized board support and is likely very useful in testing. Furthermore, QEMU also defines an ARM `virt` target, which has VirtIO devices mapped in an identical fashion.

As a follow-up to this PR, I am preparing a "TAP" driver PR for delivering Ethernet frames to userspace and implementing basic IP networking support in userspace using LwIP (web-server working locally with both QEMU and the physical LiteX FPGA board).

### Testing Strategy

This pull request was tested by running a small Ethernet "switch" (really more of a two-port hub) demo in kernel to test the `NetworkCard`, integrating this codebase with the WIP Tock Ethernet infrastructure on the `tock-ethernet` branch, and the randomness test app to test the `EntropySource`.

### TODO or Help Wanted

This pull request still needs some work in the following areas:

- [x] Generally improving code quality. I have written much of this over two years ago; some parts have not aged well.
- [x] Writing more documentation. While the VirtIO MMIO transport is reasonably well documented, some traits and especially the entire `split_queue.rs` file require more extensive documentation.
- [x] ~Safety & soundness checks. While almost all of the VirtIO support is implemented through capsules (and VirtIO device drivers in `capsules::virtio::devices` can definitely stay this way, how awesome is that?), the VirtIO transports and queues might need to move. While these parts do not use any `unsafe` code directly currently, they should not be kept in capsules given they interact with device registers through `StaticRef`s passed to them, or by having VirtIO devices write to `&mut [u8]` slices (the former might be problematic from a safety and isolation perspective, while the latter is straight up unsound and needs to use raw pointer operations).~ VirtIO support has been moved to its own chip crate, where it does use some unsafe to ensure that we're not violating Rust's memory aliasing requirements.
- [x] Removing the dummy Ethernet implementation from the `qemu_rv32_virt` board initialization and revert the Makefile hacks to instantiate two TAP network interfaces on the host for this demo.
- [x] ~Writing components for the VirtIO peripherals.~ Components postponed. Presumably we'll only have one chip supporting this, at least in the short term?
- [x] Write a primitve auto-discovery mechanism to retrieve VirtIO transports offering a specific device type to assign devices to drivers. (Users should not have to assign VirtIO devices to specific VirtIO MMIO transports exposed through QEMU)

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/qemu_rv32_virt/src/main.rs
+++ b/boards/qemu_rv32_virt/src/main.rs
@@ -40,7 +40,7 @@ const FAULT_RESPONSE: kernel::process::PanicFaultPolicy = kernel::process::Panic
 /// Dummy buffer that causes the linker to reserve enough space for the stack.
 #[no_mangle]
 #[link_section = ".stack_buffer"]
-pub static mut STACK_MEMORY: [u8; 0x2000] = [0; 0x2000];
+pub static mut STACK_MEMORY: [u8; 0x8000] = [0; 0x8000];
 
 /// A structure representing this platform that holds references to all
 /// capsules for this platform. We've included an alarm and console.
@@ -59,6 +59,7 @@ struct QemuRv32VirtPlatform {
     scheduler_timer: &'static VirtualSchedulerTimer<
         VirtualMuxAlarm<'static, qemu_rv32_virt_chip::chip::QemuRv32VirtClint<'static>>,
     >,
+    virtio_rng: Option<&'static capsules::rng::RngDriver<'static>>,
 }
 
 /// Mapping of integer syscalls to objects that implement syscalls.
@@ -71,6 +72,13 @@ impl SyscallDriverLookup for QemuRv32VirtPlatform {
             capsules::console::DRIVER_NUM => f(Some(self.console)),
             capsules::alarm::DRIVER_NUM => f(Some(self.alarm)),
             capsules::low_level_debug::DRIVER_NUM => f(Some(self.lldb)),
+            capsules::rng::DRIVER_NUM => {
+                if let Some(rng_driver) = self.virtio_rng {
+                    f(Some(rng_driver))
+                } else {
+                    f(None)
+                }
+            }
             kernel::ipc::DRIVER_NUM => f(Some(&self.ipc)),
             _ => f(None),
         }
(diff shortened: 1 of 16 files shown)


lschuermann:
I think this PR is (finally) ready now. It's still a little rough around the edges, but seems like a pretty good start for now. With the `qemu_rv32_virt` board, the RNG peripheral works out of the box now (modulo #3316, so use QEMU upstream master). The network card works as well and is going to be wired up on the `tock-ethernet` branch once this makes it in. Other peripherals should be pretty straightforward to add as well, it's really only the queue and transport parts which are tricky to get right with VirtIO.

chips/virtio/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "virtio"
+version.workspace = true
+authors.workspace = true
+edition = "2018"
```
hudson-ayers, line 5: we have been starting new crates at edition 2021 for awhile now

hudson-ayers (commented):
I don't really know enough about QEMU / VirtIO to review this PR, but after a cursory review it looks good.

One question: what is the plan for integrating this into CI? Wait until upstream QEMU tags a new release, then add QEMU tests that use this functionality?

chips/virtio/Cargo.toml
lschuermann: Great catch, thanks!

lschuermann:
Blocking this on #3334.

lschuermann:
FWIW, I've managed to integrate this with the WIP HIL and TAP driver on the `tock-ethernet` branch without any issues. These changes seem pretty solid to me. Would appreciate if we could merge this in the near future.

chips/virtio/src/queues/split_queue.rs
```diff
@@ -0,0 +1,911 @@
...
+struct SharedDescriptorBuffer<'b> {
+    ptr: NonNull<u8>,
+    len: usize,
+    _lt: PhantomData<&'b mut [u8]>,
+}
+
+impl<'b> SharedDescriptorBuffer<'b> {
+    pub fn from_slice(slice: &'b mut [u8]) -> SharedDescriptorBuffer<'b> {
+        SharedDescriptorBuffer {
+            ptr: unsafe { NonNull::new_unchecked(slice.as_mut_ptr()) },
```
hudson-ayers, line 395: I would expect the compiler to be able to optimize out any checking from using `NonNull::new().unwrap()` (https://rust.godbolt.org/z/oEdPsh1eW confirms this for at least one case). I would prefer that to adding this use of unsafe. In principle we would really prefer the only use of `unsafe` in `chips/` to be from declaring registers

chips/virtio/src/queues/split_queue.rs
```diff
@@ -0,0 +1,911 @@
...
+            len: slice.len(),
+            _lt: PhantomData,
+        }
+    }
+
+    pub fn into_slice(self) -> &'b mut [u8] {
+        unsafe { slice::from_raw_parts_mut(self.ptr.as_ptr(), self.len) }
```
hudson-ayers, line 402:
```suggestion
        // SAFETY: This is gauranteed to be safe because this struct can only be using the `from_slice()`
        // constructor, and `ptr` and `len` cannot be modified after this struct is created.
        unsafe { slice::from_raw_parts_mut(self.ptr.as_ptr(), self.len) }
```

chips/virtio/src/queues/split_queue.rs
```diff
@@ -0,0 +1,911 @@
...
+/// retain the `&'static mut [u8]` in scope though, as the VirtIO device (host)
+/// writing to it would violate Rust's memory aliasing rules. Thus, we convert a
+/// slice into this struct (converting the slice into its raw parts) for
+/// safekeeping, until we reconstruct a byte-slice from it to hand it back to
+/// the client.
+///
+/// While we technically retain an identical pointer in the
+/// [`VirtqueueDescriptors`] descriptor table, we record it here nonetheless, as
+/// a method to sanity check internal buffer management consistency.
+struct SharedDescriptorBuffer<'b> {
```
hudson-ayers, line 386: I think this is fine here for now, but it seems like this might eventually be useful as a general Tock type for buffers passed to DMA etc, and could go into `kernel/utils`

hudson-ayers (dismissed):
I reviewed all the uses of `unsafe` and they all make sense to me. Everything is confined to chips/boards so I think it is ok for this to go through without an in-depth review of all of the logic, especially if it works.

I'll mark last-call once you look at the couple of changes I suggested

chips/virtio/src/devices/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+/// virtual devices, such as network drivers, serial consoles, block devices or
+/// random number generators. A VirtIO bus endpoint announces which type of
+/// device it represents (and hence also which rules and semantics the VirtIO
+/// driver should follow).
+///
+/// This enum maps the VirtIO device IDs to human-readable variants of an enum,
+/// which can be used throughout the code base. Users should not rely on this
+/// enum not being extended. Whenever an official device ID is missing, it can
+/// be added to this enumeration.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
```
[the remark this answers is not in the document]
lschuermann: Oh, I didn't know of `#[non_exhaustive]`, this is pretty cool. Will try to add, thanks!

chips/virtio/src/queues/split_queue.rs
lschuermann: I'll validate that this is indeed optimized out, but you're right, it would make sense to be just as efficient given that Rust knows references to be non-null.

chips/virtio/src/queues/split_queue.rs
lschuermann: Yes, that's right. This is added explicitly here as #2637 hasn't progressed to an actual PR yet, but it's pretty high up my todo list and will essentially be a fancier version of this struct, maintained in the kernel.

chips/virtio/src/queues/split_queue.rs
lschuermann: I can confirm that in the compiled & optimized binary this function is literally a no-op, even without the unsafe. It seems the `SharedDescriptorBuffer` is layout-compatible with a slice reference. Even a `#[inline(never)]` doesn't cause it to be generated with a non-empty body, had to add some asm to make it visible.

chips/virtio/src/queues/split_queue.rs
hudson-ayers: Great! Try to make a note on #2367 to remove this once a fancier version is included

lschuermann:
@hudson-ayers Thanks for your review! Addressed all of your comments, seems we're down to 2 unsafe (of which one will be moved into `kernel/` soon(TM) and one should be addressed by Rust eventually). Not too shabby for VirtIO! Marking as `last-call`.

hudson-ayers (approved):

boards/qemu_rv32_virt/src/main.rs
```diff
@@ -40,7 +40,7 @@ const FAULT_RESPONSE: kernel::process::PanicFaultPolicy = kernel::process::Panic
 /// Dummy buffer that causes the linker to reserve enough space for the stack.
 #[no_mangle]
 #[link_section = ".stack_buffer"]
-pub static mut STACK_MEMORY: [u8; 0x2000] = [0; 0x2000];
+pub static mut STACK_MEMORY: [u8; 0x8000] = [0; 0x8000];
```
lschuermann, line 43: Growing the stack this large was necessary when integrating this with the TAP network driver in `tock-ethernet`, which isn't the most efficient right now (uses an in-kernel ring-buffer of packets). However, given this platform is intended as a test-vehicle for Ethernet development and purely virtual, we can afford it easily.

PR #3112, opened 2022-07-28
Fix temperature sign in bme280 and bmp280 drivers

brghena:
### Pull Request Overview

Temperature is a 20-bit, two's complement value in both the BME280 and BMP280 chips. Further,
the datasheets provide example code, which treats all units as signed,
32-bit values while doing math.

This came up when considering #2967 and looking through temperature sensor drivers and datasheets. I wanted to decouple it though, since it's chip-specific changes.

Functionally, this PR should make no difference except for temperatures
below zero Celsius. They still won't work right until #2867 lands, but that's an easy fix.

### Testing Strategy & Help Wanted

I don't have hardware for these sensors! @dcz-self and @alistair23 are you able to ensure I didn't break temperature readings?

### TODO or Help Wanted

- [ ] Test BME280
- [ ] Test BMP280

### Documentation Updated

- [N/A] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/bme280.rs
+++ b/capsules/src/bme280.rs
@@ -108,7 +108,7 @@ pub struct Bme280<'a> {
     humidity_client: OptionalCell<&'a dyn HumidityClient>,
     state: Cell<DeviceState>,
     op: Cell<Operation>,
-    t_fine: Cell<usize>,
+    t_fine: Cell<i32>,
 }
 
 impl<'a> Bme280<'a> {
@@ -292,9 +292,13 @@ impl<'a> I2CClient for Bme280<'a> {
                     Operation::None => (),
                     Operation::Temp => {
                         let calib = self.calibration.get();
-                        let adc_temperature = (buffer[0] as usize) << 12
+
+                        // note: per datasheet, measurement is 20-bit two's complement
+                        let adc_temperature: i32 = (((buffer[0] as usize) << 12
                             | (buffer[1] as usize) << 4
-                            | (((buffer[2] as usize) >> 4) & 0x0F);
+                            | (((buffer[2] as usize) >> 4) & 0x0F) << 12)
+                            as i32)
+                            >> 12; // ensure sign extension
 
                         if adc_temperature == 0 {
                             // We got a misread, try again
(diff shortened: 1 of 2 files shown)


brghena:
For double-checking me, see pages 25-26 here: https://www.mouser.com/datasheet/2/783/BST-BME280-DS002-1509607.pdf and page 22 here: https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bmp280-ds001.pdf

bradjc (approved):

capsules/src/bme280.rs
```diff
@@ -292,9 +292,13 @@ impl<'a> I2CClient for Bme280<'a> {
...
                         let calib = self.calibration.get();
-                        let adc_temperature = (buffer[0] as usize) << 12
+
+                        // note: per datasheet, measurement is 20-bit two's complement
+                        let adc_temperature: i32 = (((buffer[0] as usize) << 12
                             | (buffer[1] as usize) << 4
-                            | (((buffer[2] as usize) >> 4) & 0x0F);
+                            | (((buffer[2] as usize) >> 4) & 0x0F) << 12)
+                            as i32)
+                            >> 12; // ensure sign extension
```
alistair23, line 301: This sign extension breaks the temperature values. Everything else seems to work here

capsules/src/bme280.rs
```diff
@@ -304,28 +308,28 @@ impl<'a> I2CClient for Bme280<'a> {
...
+                            * ((adc_temperature >> 4) - (calib.temp1 as i32)))
                             >> 12)
-                            * (calib.temp3 as usize))
+                            * (calib.temp3 as i32))
                             >> 14;
 
                         self.t_fine.set(var1 + var2);
 
-                        let temperature = ((self.t_fine.get() * 5 + 128) >> 8) / 100;
+                        let temperature = (self.t_fine.get() * 5 + 128) >> 8;
```
alistair23, line 322: We still need the divide by 100

alistair23 (changes requested):

capsules/src/bme280.rs
brghena: Wouldn't dividing by 100 result in temperature in degrees Celsius? Every other temperature sensor returns hundredths of degrees Celsius instead. Am I misunderstanding something?

capsules/src/bme280.rs
brghena: Yup, sure does. I missed a parentheses in there, which meant it shifted `buffer[2]` left before or-ing it, per order of operations. Should be fixed in the new commit.

brghena:
Thanks so much for testing this @alistair23. Would you mind running the temperature app again now that I added a commit?

alistair23:
Same problem, I get values like this:

```
check run BME280 Temperature... 
Temperature: 9330
    [ok]
check run BME280 Humidity... 
Humidity: 4294967186
    [ok]
```

brghena:
Alright, I've got a BME280 breakout coming and will debug. Thanks for trying this out @alistair23

capsules/src/bmp280.rs
```diff
@@ -73,7 +73,7 @@ impl CalibrationData {
         }
     }
 
-    fn temp_from_raw(&self, raw_temp: u32) -> i32 {
+    fn temp_from_raw(&self, raw_temp: i32) -> i32 {
```
dcz-self, line 76:
```suggestion
    fn temp_from_raw(&self, temp: i32) -> i32 {
```

The following line can be removed with the above change.

dcz-self (approved):
Looks simple enough, at least the BMP280 part. I tested the calculations exhaustively (rather than in actual hardware) with https://paste.debian.net/1251850/ and I'm seeing no problems at all.

phil-levis:
@brghena ping!

hudson-ayers:
@brghena reminder on this if you get the chance to work on it! Once it is ready again feel free to re-assign Johnathan

alevy:
I rebased and (believe) correctly resolved conflicts. This seems right to me now, but can't test. @brghena ?

lschuermann (approved):
Can't speak to the correctness of these changes, but the code itself looks fine.

PR #3113, opened 2022-07-31
Correct kernel data/code disambiguation in print_tock_memory_usage

phil-levis:
### Pull Request Overview

This pull request further improves symbol counting for print_tock_memory_usage. It now correctly distinguishes kernel data and kernel code, accounting for them separately.

This has been complicated/required a series of PRs because different toolchains produce different types of symbols/symbol tables.

### Testing Strategy

This pull request was tested by running it on imix and cw310 binaries and using the results to reduce the code size of AppID.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/tools/print_tock_memory_usage.py
+++ b/tools/print_tock_memory_usage.py
@@ -47,7 +47,9 @@ sections = {}
 # computed by sorting the symbols and calculating their spacing.
 kernel_uninitialized = []
 kernel_initialized = []
+kernel_text = []
 kernel_functions = []
+kernel_data = []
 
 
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):
This looks good, but I still think eventually we should switch to using something like https://github.com/teambi0s/rust_demangler for name demangling, especially because Rust is going to switch to its new demangling scheme soon (https://github.com/rust-lang/rust/issues/60705) and that will break this tool when it happens

bradjc (approved):

Issue #2941, opened 2022-01-14
register_structs! macro: reached recursion limit during macro expansion


imDema:
I'm using `svd2regs` to parse the SVD for the STM32MP157 (which can be obtained [here](https://github.com/stm32-rs/stm32-rs/blob/master/svd/vendor/STM32MP157x_v1r6.zip)) and get an generated source which won't compile due to recursion limits in macros.

## Replicating
+ Download and extract STM32MP157x svd
+ Run `svd2regs` for the RCC component ```python3 ./tools/svd2regs.py --svd stm32mp157.svd --save generated.rs RCC```
+ (Other small issue) Fix the outdated imports by changing `kernel::common` to `kernel::utilities`
+ Try compiling

Here's an archive with pre generated code, run `cargo check` to replicate [sv2regs_issue.tar.gz](https://github.com/tock/tock/files/7870019/sv2regs_issue.tar.gz) (The tock directory is missing for archive size reasons, clone or change the path in `Cargo.toml`)

I'd be happy to help if needed

alexandruradovici:
The `register_structs` macro is indeed recursive. To prevent infinite loops, Rust has a [limit](https://doc.rust-lang.org/reference/attributes/limits.html) for the number of recursions that it will execute.

You should be able to change it as describe [here](https://doc.rust-lang.org/reference/attributes/limits.html), but you might have to edit the [lib.rs](https://github.com/tock/tock/blob/master/libraries/tock-register-interface/src/lib.rs) file from the `tock-register-interface` crate the `register_structs` macro is defined.

This should be something similar to
```rust
// place here anything higher than 128
#![recursion_limit = "512"]
```

imDema:
I figured that was the culprit, however I was wondering if a one shot (nested) macro could be beneficial instead of just increasing the limit to an arbitrary amount.
From what I can tell, the recursion is only really needed for the padding fields, which need to know the offset of the next field in order to compute the difference, which otherwise would need to be captured twice.
However by using a nested implementation, it would be possible to first duplicate the offset and then recapture it in a nested macro to achieve the same result without recursion.

The idea is something like this:
```rust
// ...
(0x0D8 => rcc_ddritfcr: ReadWrite<u32, RCC_DDRITFCR::Register>),
(0x0DC => _reserved5),
(0x100 => rcc_mp_bootcr: ReadWrite<u32, RCC_MP_BOOTCR::Register>),
(0x108 => @END),

// Duplicate offset for each field

// ...
[...]

PR #1554, opened 2020-01-20
Add support for structs with a lifetime parameter in register_structs.

gendx:
### Pull Request Overview

This pull request extends `register_structs` to allow defining structs with a lifetime parameter.

This will be useful to be able to migrate the definition of `UsbdRegisters` in `chips/nrf52/usbd` to use the `register_structs` macro.

https://github.com/tock/tock/blob/8bb3fc1fa3d1b3fe66f2ed0e85860d229df0af7c/chips/nrf52/src/usbd.rs#L46

I created this separate PR to avoid blocking https://github.com/tock/tock/pull/1543 (which is already large enough), while being able to iterate on `register_structs` in parallel.

### Testing Strategy

This pull request was tested by:
- Running `make ci-travis` with a migrated version of `UsbdRegisters` (not yet in review, to avoid blocking https://github.com/tock/tock/pull/1543 in the meantime).
- Running Travis-CI on the current code, checking that other uses of `register_structs` are unaffected.

### TODO or Help Wanted

- [x] For now, this change only allows a single lifetime parameter. I'm not sure we'll ever need more lifetimes (in the end these lifetimes are `'static` in Tock), but the syntax could be extended in the future if necessary.
- [x] Eventually, migrating `UsbdRegisters` to use `register_structs`, but this can be done after this PR, and after https://github.com/tock/tock/pull/1543 is merged.

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

The additional lifetimes should be seamless in terms of user syntax, i.e. one will be able to write:

```rust
register_structs! {
    Foo<'a> {
        (0x000 => bar: Bar<'a>),
        ...
    }
}
```

Therefore I'm not sure whether any updates to the docs are required for this specific case.

### Formatting

- [x] Ran `make formatall`.

PR #3114, opened 2022-08-01
RFC: `tock-registers` macros: avoid unsoundness, support tests

jrvanwhy:
*This PR proposes a new design for the registers structs generated by `register_structs!` that attempts to resolve an existing bug while adding support for testing driver implementations.*

## Goals of this design

Here are the goals of this design, in order from most-important to least-important:

1. The current implementation allows the Rust compiler to insert spurious reads to MMIO registers.
2. We want to be able to unit test peripheral drivers, which requires providing a fake or mock implementation of the peripheral.
3. [`register_structs!` can hit recursion limits](https://github.com/tock/tock/issues/2941).
4. Minimize breaking changes.

### Spurious MMIO reads

There isn't an open issue for this problem, so I'll describe it here.

All Rust references, including `&UnsafeCell<_>`, have the LLVM `dereferenceable` attribute. `dereferenceable` allows LLVM to insert reads to the referenced memory whenever it wants to. Some peripherals take action when an MMIO register is read (see e.g. clear-on-read registers), so spurious reads are unacceptable. It is currently unclear to me whether a register that changes between subsequent reads will cause UB, but it's possible that is an issue too. There is more background at https://github.com/rust-lang/unsafe-code-guidelines/issues/33.

## Fixing the spurious reads requires breaking changes

I don't think it is practical to fix the spurious read issue without making some backwards-incompatible changes to `tock-registers`. To demonstrate my point, consider the following register definition:

```rust
(0x100 => iser: [ReadWrite<u32, NvicSetClear::Register>; 32])
```

This emits a field that is an array. I don't think it is practical to make a type that implements the entire slice API -- or track it as it changes -- so to keep backwards compatibility, we need a real slice.

The slice cannot be zero-sized, because then `tock-registers` cannot make the following work:

```rust
iser[3].set(3);
iser[7].set(8);
```

because `iser[3]` and `iser[7]` are indistinguishable (they have the same address and type).

The slice cannot have a nonzero size and be in the MMIO region, because then you have the spurious read issue I am trying to fix.

I don't think we want to put the slice anywhere else, because it would have to be in readable memory, which requires either putting it into flash or RAM. While that would technically be possible, some of these register arrays have hundreds of elements, and the flash/RAM cost would be unreasonable.

At the moment, the only way I see to fix the issue while supporting large array-registers is to create a new type that implements a subset of Rust's slice API.

If you have another idea on how to fix the spurious reads, let me know!

## Supporting fakes/mocks

To support unit tests, we need to allow users to replace the MMIO interface with an alternative implementation, such as fakes or mocks. We can do this by using generics. Given the following register definition:

```rust
register_structs! {
    Foo {
        (0x00 => rw_single: ReadWrite<u32>),
        (0x04 => rw_array: [ReadWrite<u32>; 4]),
    }
}
```

we need the following operations (I'm ignoring error handling here):

```rust
trait AccessRegisters {
    fn read_rw_single(&self) -> u32;
    fn write_rw_single(&self, value: u32);

    fn read_rw_array(&self, idx: usize) -> u32;
    fn write_rw_array(&self, value: u32, idx: usize);
}
```

However, a `macro_rules!` macro cannot prepend the `read_` and `write_` prefixes (see the [`concat_idents` tracking issue](https://github.com/rust-lang/rust/issues/29599) for more information), so instead we need two separate traits:

```rust
trait ReadRegisters {
    fn rw_single(&self) -> u32;
    fn rw_array(&self, idx: usize) -> u32;
}

trait WriteRegisters {
    fn rw_single(&self, value: u32);
    fn rw_array(&self, value: u32, idx: usize);
}

trait AccessRegisters: ReadRegisters + WriteRegisters {}
impl<T: ReadRegisters + WriteRegisters> AccessRegisters for T {}
```

## Providing backwards-compatible-ish access

Obviously, the `AccessRegisters` trait looks very different from the existing `register_structs!` API. Each register should become a struct field with getter and setter methods (such as `Readable` and `Writeable` implementations). These fields must call into the `AccessRegisters` implementation, so we make them generic over `AccessRegisters`:

```rust
struct Registers<A: AccessRegisters> {
    access: A,

    rw_single: RwSingle<A>,
    rw_array: RwArray<A>,
}
```

oh, except a `macro_rules!` macro can't re-capitalize an identifier, so unfortunately the types will have to be snake case:

```rust
struct Registers<A: AccessRegisters> {
    access: A,

    rw_single: rw_single<A>,
    rw_array: rw_array<A>,
}
```

There is an important unanswered question in this struct definition: how do `rw_single` and `rw_array` access the `access` member? We don't want them to contain pointers, as that would require a LOT of pointers. In fact, we need them to be zero-sized, else we will end up kilobytes of wasted RAM per board. Which means that the only non-zero-sized member of `Registers` is `access` (and we aren't even sure that `access` isn't a ZST!). Thus we can put all of `Registers`' fields at the same address:

```rust
#[repr(C)]
struct Registers<A: AccessRegisters> {
    rw_single: rw_single<A>,
    rw_array: rw_array<A>,

    // access must be last
    access: A,
}
```

Now `rw_single` and `rw_array` can call into `access` by casting `self` to a `&A`!

## Generate a module, not a struct

The above design generates multiple traits and many types. Therefore, I propose that the macro generates a module named after the peripheral, with mostly generic names inside:

```rust
mod foo_peripheral {
    #[repr(C)]
    struct Registers<A: AccessRegisters> { ... }

    trait ReadRegisters ...
    trait WriteRegisters ...
    trait AccessRegisters ...

    struct rw_single ...
    struct rw_array ...
}
```

For backwards compatibility, we should give the macro a new name. I suggest `registers!`.

To avoid maintaining two sets of macro implementations, I made `register_structs!` wrap `registers!` and re-export only the `Registers` type. Unfortunately, `registers!` requires distinct module names for each register struct it generates, and `register_structs!` isn't normally passed distinct module names. So instead, I tweaked `register_structs!` so that you can pass in a register name:

```rust
register_structs! {
    Foo, mod foo_registers {
        ...
    }
}
```

## Implementation details and remaining work

There is a lot of remaining work to do in this PR:

1. The "real" and "mock" implementations of `AccessRegisters`
1. The detailed macro chaining required to implement read and write methods for only registers that are readable and writable, respectively.
1. The API for accessing array registers, which do not fit into the `Readable` and `Writeable` traits.
1. Adding support for padding fields.
1. Implement tests to make sure there are not gaps/overlaps in the register addresses.
1. Figure out what should happen if the `register_types` feature is disabled, or remove that feature. @lschuermann, any input here?
1. Properly support `Aliased` registers (I kept forgetting about them while writing macros)

### Slice API subset

As mentioned above, our API for accessing arrays of registers can mimic a subset of Rust's slice API, but it isn't practical to implement the entire thing (and we can't use real slices).

### `ReadOnly`/`ReadWrite`/`WriteOnly`/`Aliased` pickiness

Most existing users of `tock-registers` put a subset of:

```rust
use tock_registers::registers::{Aliased, ReadOnly, ReadWrite, WriteOnly};
```

in the file in which they define their registers. This implementation assumes that, and matches on those register names directly. So the following will not work:

```rust
register_structs! {
    Foo {
        (0x0 => reg: tock_registers::registers::ReadWrite<u32>),
    }
}
```

and will have to be rewritten as:

```rust
register_structs! {
    Foo {
        (0x0 => reg: ReadWrite<u32>),
    }
}
```

### No lifetime parameters

This PR removes the support for lifetime parameters added in #1554. It's not clear to me how those parameters should be integrated into the trait design, and OpenSK appears to be dead so I'm not sure I'll get an answer.

### `register_structs!` needs module names

As mentioned previously, the new `register_structs!` implementation needs module names to be provided if there is more than one register struct in a single module.

## Key Question

In creating this PR, I really bent the implementation of `registers!` to fit the existing API, so that it could be as close to a drop-in replacement as possible. However, it is not perfectly backwards-compatible, as shown by the above known limitations. I'm concerned that I'm designing `tock-registers` into a corner, where we can't improve it or add necessary features without necessitating a backwards-incompatible redesign.

As mentioned above, I *think* that fixing the MMIO reference bug requires making backwards-incompatible changes. How do we want to fix that bug? We can fix it with fewer breaking changes by leaving out support for fake peripheral implementations (and use `cargo` features to implement an expectation/mocking API instead). We can fix it with even fewer breaking changes than that if we don't try to give `register_structs!` unit testing features.

Perhaps instead of trying to do everything in one API, we should just have two APIs? `v1` which is the current `register_structs!` (including its MMIO bug), and `v2` which is a completely redesigned API that is easier to test and expand.

I look forward to your input, if you have the time to look at this!

### `macros.rs`

I recommending reading the [new `macros.rs`](https://github.com/jrvanwhy/tock/blob/reg-macro-rfc/libraries/tock-register-interface/src/macros.rs) rather than looking at the diff, as the file has been completely rewritten.

### How to examine the generated code

I recommend installing [`cargo-expand`](https://crates.io/crates/cargo-expand) (if you haven't done so already) and running

```
cargo expand --example registers -p tock-registers
```

to see an example of the generated code. The source code for the example is libraries/tock-register-interface/examples/registers.rs.

### DIFF
--- a/libraries/tock-register-interface/examples/register_structs.rs
+++ b/libraries/tock-register-interface/examples/register_structs.rs
@@ -1,0 +1,18 @@
+//! An example of using the `register_structs!` macro.
+
+use tock_registers::register_structs;
+
+register_structs! {
+    Foo {
+        (0x00 => ro_single: ReadOnly<u32, ()>),
+        (0x04 => ro_array: [ReadOnly<u32, ()>; 2]),
+        (0x0c => rw_single: ReadWrite<u32, ()>),
+        (0x10 => rw_array: [ReadWrite<u32, ()>; 3]),
+        (0x1c => wo_single: WriteOnly<u32, ()>),
+        //(0x20 => _padding),
+        (0x24 => wo_array: [WriteOnly<u32, ()>; 4]),
+        //(0x34 => @END),
+    }
+}
+
+fn main() {}
--- a/libraries/tock-register-interface/examples/registers.rs
+++ b/libraries/tock-register-interface/examples/registers.rs
@@ -1,0 +1,16 @@
+//! An example of using the `registers!` macro.
+
+use tock_registers::registers;
+
+registers! {
+    foo {
+        (0x00 => ro_single: ReadOnly<u32, ()>),
+        (0x04 => ro_array: [ReadOnly<u32, ()>; 2]),
+        (0x0c => rw_single: ReadWrite<u32, ()>),
+        (0x10 => rw_array: [ReadWrite<u32, ()>; 3]),
+        (0x1c => wo_single: WriteOnly<u32, ()>),
+        (0x20 => wo_array: [WriteOnly<u32, ()>; 4]),
+    }
+}
+
+fn main() {}
--- a/libraries/tock-register-interface/src/internal.rs
+++ b/libraries/tock-register-interface/src/internal.rs
@@ -1,0 +1,84 @@
+//! Module containing items used by `tock-registers` macros. These must be `pub`
+//! so the macros can use them, but are not intended to be part of
+//! `tock-registers`' API.
+//!
+//! In other words: don't use any of these outside this crate.
+
+use core::ptr::{read_volatile, write_volatile};
+use crate::{TooLargeIndex, WriteSuccess};
+
+// Reads a value from a MMIO non-array register.
+//
+// Safety:
+//     mmio must point to the beginning of the peripheral.
+//
+//     REL_ADDR must be the address of the register to read, relative to the
+//     beginning of the peripheral.
+//
+//     This register must be readable.
+#[inline]
+pub unsafe fn mmio_read<const REL_ADDR: usize, Mmio, T>(mmio: &Mmio) -> T
+{
+    // This is unsound under strict provenance. Under strict provenance, we
+    // should attach mmio's provenance to the resulting pointer. For now, we
+    // can't do that without enabling a nightly feature (which may be worth
+    // doing in Miri) or using the `sptr` crate.
+    let addr = (mmio as *const _ as usize + REL_ADDR) as *const T;
+    unsafe { read_volatile(addr) }
+}
+
+// Reads a value from a MMIO array register.
+//
+// Safety:
+//     mmio must point to the beginning of the peripheral.
+//
+//     REL_ADDR must be the address of the beginning of the register array,
+//     relative to the beginning of the peripheral.
+//
+//     This register must be readable.
+#[inline]
+pub unsafe fn mmio_read_array<const REL_ADDR: usize, const LEN: usize, Mmio, T>(mmio: &Mmio, idx: usize)
+    -> Result<T, TooLargeIndex>
+{
+    if idx >= LEN { return Err(TooLargeIndex); }
+    // See the comment in mmio_read about strict provenance.
+    let array_start = (mmio as *const _ as usize + REL_ADDR) as *const T;
+    Ok(unsafe { read_volatile(array_start.add(idx)) })
+}
+
+// Writes a value to a MMIO non-array register.
+//
+// Safety:
+//     mmio must point to the beginning of the peripheral.
+//
+//     REL_ADDR must be the address of the register to write, relative to the
+//     beginning of the peripheral.
+//
+//     This register must be writable.
+#[inline]
+pub unsafe fn mmio_write<const REL_ADDR: usize, Mmio, T>(mmio: &Mmio, value: T)
+{
+    // See the comment in mmio_read about strict provenance.
+    let addr = (mmio as *const _ as usize + REL_ADDR) as *mut T;
+    unsafe { write_volatile(addr, value) }
+}
+
+// Writes a value to a MMIO array register.
+//
+// Safety:
+//     mmio must point to the beginning of the peripheral.
+//
+//     REL_ADDR must be the address of the beginning of the register array,
+//     relative to the beginning of the peripheral.
+//
+//     This register must be writable.
+#[inline]
+pub unsafe fn mmio_write_array<const REL_ADDR: usize, const LEN: usize, Mmio, T>(mmio: &Mmio, value: T, idx: usize)
+    -> Result<WriteSuccess, TooLargeIndex>
+{
+    if idx >= LEN { return Err(TooLargeIndex); }
+    // See the comment in mmio_read about strict provenance.
+    let array_start = (mmio as *const _ as usize + REL_ADDR) as *mut T;
+    unsafe { write_volatile(array_start.add(idx), value) }
+    Ok(WriteSuccess)
+}
--- a/libraries/tock-register-interface/src/lib.rs
+++ b/libraries/tock-register-interface/src/lib.rs
@@ -57,9 +57,11 @@
 // If we don't build any actual register types, we don't need unsafe
 // code in this crate
 #![cfg_attr(not(feature = "register_types"), forbid(unsafe_code))]
+#![deny(unsafe_op_in_unsafe_fn)]
 
 pub mod fields;
 pub mod interfaces;
+pub mod internal;
 pub mod macros;
 
 #[cfg(feature = "register_types")]
@@ -126,3 +128,9 @@ pub trait RegisterLongName {}
 // Useful implementation for when no RegisterLongName is required
 // (e.g. no fields need to be accessed, just the raw register values)
 impl RegisterLongName for () {}
+
+/// An array index was out of bounds.
+pub struct TooLargeIndex;
+
+/// Indicates a register write succeeded.
+pub struct WriteSuccess;
--- a/libraries/tock-register-interface/src/macros.rs
+++ b/libraries/tock-register-interface/src/macros.rs
@@ -1,380 +1,273 @@
 //! Macros for cleanly defining peripheral registers.
 
 #[macro_export]
-macro_rules! register_fields {
-    // Macro entry point.
-    (@root $(#[$attr_struct:meta])* $vis_struct:vis $name:ident $(<$life:lifetime>)? { $($input:tt)* } ) => {
-        $crate::register_fields!(
-            @munch (
-                $($input)*
-            ) -> {
-                $vis_struct struct $(#[$attr_struct])* $name $(<$life>)?
+macro_rules! registers {
+    (
+        $(
+            $(#[$mod_attr:meta])*
+            $vis:vis $name:ident {
+                $(
+                    $(#[$field_attr:meta])*
+                    ($offset:literal => $reg_name:ident: $($tokens:tt)*)
+                ),*
+                $(,)?
             }
-        );
-    };
-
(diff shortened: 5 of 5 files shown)


jrvanwhy:
@cfrantz, @kupiakos, @lschuermann: I think you all would be interested in this RFC.

jrvanwhy:
Sorry about the title change: I composed the PR description in a second document, and forgot to move the title over from that document when I opened this PR.

bradjc:
I personally have no general issue with making breaking changes to tock-registers. We could make tock-registers 2.0. Having both 1.x and 2.x is also fine with me, although in that case we probably really should follow through with making it its own repo.

However, I'm not necessarily volunteering to fix all the chip crates if that becomes necessary! It would be helpful to see a chip driver updated to get a sense of what the changes would look like from an API standpoint.

lschuermann:
Thanks for getting this draft up so quickly. I am convinced that we should work on the persistent unsoundness in `tock-registers` caused by holding references to MMIO addresses. Also, I think that the ability to mock registers sounds like a sensible use case.

However, I am not yet convinced of the particular approach to supporting fakes/mocks proposed here. As far as I understand your proposal, a register struct is going to hold a collection of zero-sized types, which are being passed an instance of the `AccessRegisters` trait through a generic parameter to perform the underlying fundamental operations of reading and writing memory. While this would work for the common case, I don't particularly like it for a couple of reasons.

Most importantly, I believe that funneling all memory accesses through something passed as an associated type to _all_ register types works against our abstraction levels established within `tock-registers` currently, which are inherently defined by the traits of the `interfaces` module. The basic idea of these interfaces is that a method or struct can take any type implementing a given trait (with appropriate associate types, e.g. `Readable<T=u32, R=RXStatus>`) to define composable and reusable modules (drivers), which can be mapped to a specific chip/board/architecture by providing a concrete type meeting the trait requirement. The concrete types can be the registers provided by the `tock-registers` crate, or user-defined and thus entirely arbitrary types. This flexibility is useful for people outside of Tock, as well as internal usecases (e.g. RISC-V CSRs).

In fact, this intuitive trait-based abstraction model is alone already sufficient to provide register mocks: instead of operating on a `ReadWrite<u32, RXStatus>`, driver implementations can simply operate on some `impl ReadWriteable<T=u32, R=RXStatus`. Mock implements should provide their own `MockReadWrite` type, implementing this trait. The beauty of this approach, compared to having registers call through some `AccessRegisters` trait, is that no restrictions or assumptions are placed on the register type implementations whatsoever.

Now, some difficulty does arise when reconciling this approach to generic registers with register structs such as `register_fields!`. In theory, this is rather simple: we define a trait representing a "register collection", which represents a set of register implementations providing certain operations:

```rust
pub trait RegisterCollection {
    type ReadOnly<T: UIntLike, R: RegisterLongName>: Readable<T = T, R = R>;
    type WriteOnly<T: UIntLike, R: RegisterLongName>: Writeable<T = T, R = R>;
    type ReadWrite<T: UIntLike, R: RegisterLongName>: ReadWriteable<T = T, R = R>;
}
```

, and implement this trait on some type, which shall serve as our register collection to be plugged in to a register struct:

```rust
pub enum StandardRegisterCollection {}
impl RegisterCollection for StandardRegisterCollection {
    type ReadOnly<T: UIntLike, R: RegisterLongName> = ReadOnly<T, R>;
    type WriteOnly<T: UIntLike, R: RegisterLongName> = WriteOnly<T, R>;
    type ReadWrite<T: UIntLike, R: RegisterLongName> = ReadWrite<T, R>;
}
```

In this model (using the current unsound `tock-registers` struct layout), a register struct looks like the following:

```rust
#[repr(C)]
pub struct CsRngRegisters<RC: RegisterCollection> {
    intr_state: RC::ReadWrite<u32, INTR::Registers>,
    err_code: RC::ReadOnly<u32, ()>,
}
```

To switch a driver implementation from using MMIO registers to some mock register types is just a matter of changing `CsRngRegisters<StandardRegisterCollection>` to something like `CsRngRegisters<MockRegisterCollection>`. I've also implemented this for the current `register_fields!` macro, which works like a charm:

```rust
register_structs! {
    pub CsRngRegisters {
        (0x00 => intr_state: RC::ReadWrite<u32, ()>),
        (0x04 => intr_enable: RC::ReadWrite<u32, ()>),
        (0x08 => @END),
    }
}

type MMIOCsRngRegisters = DeviceRegisterSet<StandardRegisterCollection>;
```

The only major issue with this approach is that it requires generic associated types (GAT), which are unstable in Rust and likely will not be stabilized any time soon. There are two other approaches which would uphold the layers of abstraction as they are currently: on the one hand, one could spell out the various integer types in the register collection trait, as it is currently done in the LiteX peripherals:


```rust
pub trait RegisterCollection {
    type ReadOnly8: Readable<T = u8, R = ()>;
    type WriteOnly8: Writeable<T = u8, R = ()>;
    type ReadWrite8: ReadWriteable<T = u8, R = ()>;

    type ReadOnly16: Readable<T = u16, R = ()>;
    type WriteOnly16: Writeable<T = u16, R = ()>;
    type ReadWrite16: ReadWriteable<T = u16, R = ()>;

...
```

In fact, the `AccessRegisters` would probably have the very same issue, and would need to provide methods for the various integer sizes to support. The most major drawback of this approach is that the `RegisterLongNames` can no longer be supplied with the registers. In LiteX, I have hacked around this by providing "register wrappers", but that is not a particularly ergonomic solution:

https://github.com/tock/tock/blob/7dee0db559fa668848c1a8b1e8ccf348e875691a/chips/litex/src/litex_registers.rs#L476-L486

Finally, yet another approach could be to pass module prefixes into the `register_fields!` macro, which generates multiple instances of the register structures, each with different registers, and a common trait implemented by them. This would retain our layers of abstractions (make the register structs pluggable while not placing assumptions/restrictions on the register types), but limit the register types usable for a particular register struct at compile time:

```rust
register_structs! {
    Standard: kernel::common::registers::standard_registers,
    Mock: crate::my_mock_registers,
} {
    pub DeviceRegisterSet {
        (0x00 => intr_state: rc::ReadWrite<u32, ()>),
        (0x04 => intr_enable: rc::ReadWrite<u32, ()>),
        (0x08 => @END),
    }
}
```

I have not built a PoC for the latter approach just yet.

jrvanwhy:
I've identified a soundness issue with this design. In my proposed struct design (`pub` added where necessary):

```rust
#[repr(C)]
struct Registers<A: AccessRegisters> {
    pub rw_single: rw_single<A>,
    pub rw_array: rw_array<A>,

    // access must be last
    access: A,
}
```

User code can move the `rw_single` and `rw_array` variables out of the `Registers<_>` instance safely (imagining I implemented `Registers::new`):

```rust
fn kaboom() {
    let registers = Registers::new(FakeFoo);
    let rw_single = registers.rw_single; // Move here
    rw_single.write(0); // Zeroes out a random thing on the stack, UB
}
```

I may have a new idea that solves this problem as well as addresses Leon's concerns, currently trying to prototype enough to convince myself it is workable.

bradjc:
I believe a new version of this PR is in progress.

jrvanwhy:
Yes, I am redesigning this.

Issue #3081, opened 2022-07-12
TicKV panics at certain value sizes


eflukx:
Hi I'm using (ok.. _planning_ to use) TickV in my application, looks like a great little tool! I use it apart from Tock, using nrf-hal (and RTIC). I wrote a simple implementation of `FlashController` for the nrf-hal `nvmc` peripheral and things mostly seem to work fine!

While testing the newly found capabilities of my app I ran into the following:

- Writes of values length 1-4067 bytes work fine
- Writes of values length >=4070 bytes result in an (probably expected) error code
- **Writes of values length 4068-4069 bytes result in a panic** 

_Note that I have been testing this with a single key, on a clean (empty, initialized) TicKV store._

The panic seems to have its [origins at this source line](https://github.com/tock/tock/blob/99cf1d32ce05dfd96a98e8f6989c162e753bba57/libraries/tickv/src/tickv.rs#L496) resulting in an out-of-bounds indexing operation.
Probably the simplest (and sane) solution is to tighten the `object_length` check in `append_key()` (?)

alistair23:
Hey! I'm glad to hear you find TicKV useful!

A panic is bad, so that needs to be fixed. It would be great if you could isolate a test case. If you could share a test case then we can fix the issue and ensure it doesn't occur again.

I'm guessing the value plus the header data is larger then the memory region, which is causing the panic. We should probably change the array access to use `get()`, that way we can remove a panic

bootstrap-prime:
Do you happen to have your TickV nrf-hal impl for FlashController publically available? I'm unable to implement it because the trait requires &self instead of allowing &mut self... are you using interior mutability?

eflukx:
> Do you happen to have your TickV nrf-hal impl for FlashController publically available? I'm unable to implement it because the trait requires &self instead of allowing &mut self... are you using interior mutability?

Yes, I do use interior mutability indeed.. :) I've posted [a gist with my current interface](https://gist.github.com/eflukx/8d85fad415b8fdc4d86dc4f382d8de7b). Probably useful for someone.

(I included the check-if-pager-empty before actually erasing the flash page, as was done in the nrf-hal nvmc example code, but I don't know if that's actually useful, as TickV probably very seldomly erases already-erased pages)

eflukx:
@bootstrap-prime if you could test if you get the same the panicing-behavior i got using my implementation that would be great.

[...]

PR #3120, opened 2022-08-03
libraries: tickv: Convert panics to errors

alistair23:
### Pull Request Overview

Convert existing panics to instead return a useful error

This should fix: https://github.com/tock/tock/issues/3081

### Testing Strategy

CI

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/libraries/tickv/src/tickv.rs
+++ b/libraries/tickv/src/tickv.rs
@@ -558,18 +558,40 @@ impl<'a, C: FlashController<S>, const S: usize> TicKV<'a, C, S> {
 
                 // Copy in new header
                 // This is a little painful, but avoids any unsafe Rust
-                region_data[offset + VERSION_OFFSET] = header.version;
-                region_data[offset + LEN_OFFSET] =
+                *region_data
+                    .get_mut(offset + VERSION_OFFSET)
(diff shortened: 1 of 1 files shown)


phil-levis:
My experience going through the TBF code is doing a single check (the slice is long enough) then direct slice accesses can reduce code size. I observed that if there were 3 or more of the `get`/`ok_or` pattern, a check and direct accesses was smaller in terms of code, because those accesses would then not have panics. What is the code size effect of this change? I don't think it should block on that (this is a capsule), just wondering.

hudson-ayers:
This PR reduces the size of the OT binary by almost 500 bytes (as a reminder, you can check the benchmark CI run for these numbers on any PR that passes all checks! https://github.com/tock/tock/runs/7643510896?check_suite_focus=true)

hudson-ayers (approved):
Big fan of removing panics, and this looks pretty mechanical.

twilfredo (approved):

lschuermann (approved):
LGTM and pretty uncontroversial.

bors r+

phil-levis (approved):

PR #3123, opened 2022-08-03
RFC: Replace `DynamicDeferredCall` with modified `DeferredCall` that supports capsules and peripherals

hudson-ayers:
### Pull Request Overview

This pull request builds upon the work in https://github.com/tock/tock/pull/3086 to construct a version of `DeferredCall` that can be used by both capsules and chip peripherals, obsoleting `DynamicDeferredCall`. So far, I have ported this approach to the two sam4l peripherals with deferred calls (`CRCCU` and `FLASHCALW`) and to one capsule with a dynamic deferred call (`ieee802154/driver.rs`)

The major downside of this approach is that all capsules which use deferred calls add a generic parameter they did not previously have. It also adds some verbosity to main.rs, where boards have to define the mapping of `DeferredCallTask`s --> callback functions.

Otherwise, this approach seems pretty beneficial: it does not use any trait object dispatch, eliminating vtables and speeding the time to set a deferred call. It reduces code size compared to `DynamicDeferredCall`, at least for the radio case. It reduces the number of fields needed on capsules to store a deferred call, replacing `DynamicDeferredCaller` and `DeferredCallHandle` with a single `DeferredCall` struct, and removing the need for a `initialize_callback_handle()` function.

### Testing Strategy

### TODO or Help Wanted

This pull request still needs to port of all capsules except the ieee802154 driver.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/ieee802154/driver.rs
+++ b/capsules/src/ieee802154/driver.rs
@@ -4,6 +4,7 @@
 //! frames. Also provides a minimal list-based interface for managing keys and
 //! known link neighbors, which is needed for 802.15.4 security.
 
+use crate::driver::Task;
 use crate::ieee802154::{device, framer};
 use crate::net::ieee802154::{AddressMode, Header, KeyId, MacAddress, PanID, SecurityLevel};
 use crate::net::stream::{decode_bytes, decode_u8, encode_bytes, encode_u8, SResult};
@@ -11,8 +12,8 @@ use crate::net::stream::{decode_bytes, decode_u8, encode_bytes, encode_u8, SResu
 use core::cell::Cell;
 use core::cmp::min;
 
-use kernel::dynamic_deferred_call::{
-    DeferredCallHandle, DynamicDeferredCall, DynamicDeferredCallClient,
+use kernel::deferred_call::{
+    DeferredCall, DeferredCallManager, DeferredCallMapper, DeferredCallTask,
 };
 use kernel::grant::{AllowRoCount, AllowRwCount, Grant, UpcallCount};
 use kernel::processbuffer::{ReadableProcessBuffer, WriteableProcessBuffer};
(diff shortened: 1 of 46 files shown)


kernel/src/deferred_call.rs
```diff
@@ -1,84 +1,123 @@
...
-/// support" according to their specification. This makes it work on thumbv6
-/// platforms.
-///
-/// Borrowed from https://github.com/japaric/heapless/blob/master/src/ring_buffer/mod.rs
-/// See: https://github.com/japaric/heapless/commit/37c8b5b63780ed8811173dc1ec8859cd99efa9ad
-struct AtomicUsize {
-    v: UnsafeCell<usize>,
+
+pub trait DeferredCallMapper {
+    type PT: PeripheralTask;
```
bradjc, line 16: Probably should be `ChipTask` (for both)?

capsules/src/ieee802154/driver.rs
```diff
@@ -172,7 +173,7 @@ pub struct App {
     pending_tx: Option<(u16, Option<(SecurityLevel, KeyId)>)>,
 }
 
-pub struct RadioDriver<'a> {
+pub struct RadioDriver<'a, M: DeferredCallMapper<CT = Task> + 'static> {
```
bradjc, line 176:
```suggestion
pub struct RadioDriver<'a, DCM: DeferredCallMapper<CT = Task> + 'static> {
```
?

bradjc (commented):
I like this. I'm not sure we were ever deliberate with having two separate deferred call mechanisms, and I don't remember there being a use case for truly dynamic deferred calls.

The syntax is a little clunky, but I see that as the price of using rust.

hudson-ayers:
After the core call discussion of this a couple weeks ago I realized two problems with this approach:

1. This approach does not provide not a way for out-of-tree boards to both a) depend on capsules which use deferred calls and b) add new capsules which use deferred calls without forking our capsules crate
2. This approach probably does not work for certain in-kernel structures which need to be able to schedule deferred calls but also need to be object safe (e.g. nodes in a virtualizer)

I am working on a new approach that does not have these problems, but going to go ahead and close this PR.

PR #3124, opened 2022-08-04
AppId: move verifier to `KernelResources`

bradjc:
### Pull Request Overview

Note! This is a PR targeting the AppID PR (not master).

This pull request changes the verifier from being an argument passed in to `kernel::new()` to a type in the `KernelResources` trait. We moved at some point away from using arguments to `kernel:new()` as the way to configure the kernel because 1) using function arguments is verbose and poorly self-documenting, and 2) using a trait allows for more standardization and provides a clear tool to use for a board to configure how it wants the kernel to operate.

However, due to Rust-isms and my limited ability to figure out how to get all of the rust types to compile, I couldn't figure out a way to move how verifier is passed to the kernel while keeping the existing control flow. The issue is that `submit_process` needs a verifier, and if `kernel` no longer has a reference, everything that calls `submit_process()` must provide it one. But to get a verifier requires being passed a `KernelResources`, and that now holds an `AppVerifier`, which requires a lifetime, it seemed to get increasingly difficult to get everything (including `ProcessStandard`) the correct types to make that work.

Instead, this PR proposes removing `submit_process()` and instead replacing it with a process state that means "this process is ready to run, pending kernel uniqueness verification, please check the process and if it is ok then run it". That way, instead of having the complexity of getting the verifier types to work, all that is needed is updating the process's state, and allowing the kernel loop to do the uniqueness check, and since the kernel loop already has access to `KernelResources` this is straightforward.

I reused the `process::State::Unstarted` state to do this because it was the easiest. However, we may want a different state for unstarted and "ready to run". I think this should be easy to update, but if we can resuse unstarted then that seems simpler.

### Testing Strategy

Running blink on hail.

### TODO or Help Wanted

Is this reasonable? I think my only remaining major concern with AppId is that it splits configuring the kernel back into passing to new() and the resources trait. If this resolves that then I really think we should merge appid soon.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.


kernel/src/process_checking.rs
```diff
@@ -40,23 +40,31 @@ pub trait Client<'a> {
 
 /// Implements a Credentials Checking Policy.
 pub trait AppCredentialsChecker<'a> {
-    fn set_client(&self, client: &'a dyn Client<'a>);
-    fn require_credentials(&self) -> bool;
+    fn set_client(&self, _client: &'a dyn Client<'a>) {}
+    fn require_credentials(&self) -> bool {
+        false
```
phil-levis, line 45: There shouldn't be defaults here.

kernel/src/process_checking.rs
```diff
@@ -40,23 +40,31 @@ pub trait Client<'a> {
...
 
     fn check_credentials(
         &self,
-        credentials: TbfFooterV2Credentials,
-        binary: &'a [u8],
-    ) -> Result<(), (ErrorCode, TbfFooterV2Credentials, &'a [u8])>;
+        _credentials: TbfFooterV2Credentials,
+        _binary: &'a [u8],
+    ) -> Result<(), (ErrorCode, TbfFooterV2Credentials, &'a [u8])> {
+        Ok(())
```
phil-levis, line 53: There shouldn't be defaults here.

phil-levis:
It seems strange to couple security policy decisions (application credential checking) with context switch callback implementations. Put another way, there should be a distinction between kernel implementation decisions and security policy decisions. Also, decoupling the checker from the kernel means that it's theoretically possible to, well, decouple them. Is it possible to construct an alternate `KernelResources` and use it? 

It seems like `KernelResources` is doing three separate things: implementations of some low-level mechanisms (driver lookup, context switch callback, watchdog), some execution policies (process fault, scheduler), and some security policies (process checking, syscall filtering). Perhaps we should break it up in this way?

phil-levis:
> Instead, this PR proposes removing `submit_process()` and instead replacing it with a process state that means "this process is ready to run, pending kernel uniqueness verification, please check the process and if it is ok then run it".

Note this is different behavior. If a process passes checking, but it not unique, it is queued to run. If the process it conflicts with terminates, then this queued process will immediately run. Your implementation does day "disable this process" in a comment, this would need to be implemented. But also, it means that if there are multiple processes with the same AppID, which one will run depends on the kernel implementation for traversing them in the scheduler, rather than the order in flash. This could be OK, it's just different.

bradjc:
> It seems strange to couple security policy decisions (application credential checking) with context switch callback implementations. Put another way, there should be a distinction between kernel implementation decisions and security policy decisions.

Those could be the same, depending on how you view the kernel. But, they're coupled via a Rust mechanism, not in how the kernel is designed. The alternative is having `kernel::new(verifier, syscall_callback, watchdog, scheduler, etc)`, where they would all be grouped in the struct (or more realistically actually `kernel::new(None, None, None, None, scheduler)`).

>Also, decoupling the checker from the kernel means that it's theoretically possible to, well, decouple them. Is it possible to construct an alternate `KernelResources` and use it?

`KernelResources` is only passed into the kernel once, but we should make that explicit. The intent isn't to not have the resources in the kernel struct, I imagine we tried that initially. The issue is getting the type parameters to work without having to propagate them to every file in tock.


> It seems like `KernelResources` is doing three separate things: implementations of some low-level mechanisms (driver lookup, context switch callback, watchdog), some execution policies (process fault, scheduler), and some security policies (process checking, syscall filtering). Perhaps we should break it up in this way?

I can see breaking it up, but I think we should consider the complexity cost in the board main.rs files.


> Note this is different behavior. If a process passes checking, but it not unique, it is queued to run. If the process it conflicts with terminates, then this queued process will immediately run. Your implementation does day "disable this process" in a comment, this would need to be implemented.

Agreed. I want to make sure I understand the specific state machine we are trying to provide for processes so we have a clear idea of what is valid in each process state.

> But also, it means that if there are multiple processes with the same AppID, which one will run depends on the kernel implementation for traversing them in the scheduler, rather than the order in flash. This could be OK, it's just different.

It seems like with dynamic app loading coming we probably shouldn't try to provide any guarantees in the case of duplicate appids.

phil-levis:
> Those could be the same, depending on how you view the kernel. But, they're coupled via a Rust mechanism, not in how the kernel is designed. The alternative is having `kernel::new(verifier, syscall_callback, watchdog, scheduler, etc)`, where they would all be grouped in the struct (or more realistically actually `kernel::new(None, None, None, None, scheduler)`).

Yes, they are coupled in the API. I think having multiple parameters is better, because it forces you to design and think through what the parameters are (i.e., design the responsibilities and decomposition of the kernel). By passing a struct of arbitrary configurations, we're pushing the type checking to the construction of the type, rather than the invocation of the method. Imagine if snprintf was `snprintf(snprintf_args args)`. There's a reason why APIs are designed this way: it makes the set of things required to invoke the method more explicit.

phil-levis:
> I can see breaking it up, but I think we should consider the complexity cost in the board main.rs files.

In terms of complexity of `main.rs`, I think that the actual kernel construction should not be our big concern; it's a tiny piece of most `main.rs` files yet it's a very important thing. This could be as something as simple as breaking it up into three separate traits and passing them in `kernel_loop`. If we decoupled it, then it might be possible to factor out chip-independent parts into common implementations for policies (e.g., security policies), while keeping common chip-specific ones for stuff like `SysTick`.

bradjc:
> Yes, they are coupled in the API. I think having multiple parameters is better, because it forces you to design and think through what the parameters are (i.e., design the responsibilities and decomposition of the kernel). By passing a struct of arbitrary configurations, we're pushing the type checking to the construction of the type, rather than the invocation of the method. Imagine if snprintf was `snprintf(snprintf_args args)`. There's a reason why APIs are designed this way: it makes the set of things required to invoke the method more explicit in the method itself.

I see your point, but `snprintf` is doing a relatively constrained task. I'm imaging if `spi.transfer(spi_transfer)` in linux was `spi.transfer(tx_buf, rx_buf, len, tx_dma, rx_dma, tx_sg, rx_sg, cs_change, tx_nbits, rx_nbits, bits_per_word, delay_usecs, speed_hz, transfer_list)`. But even that is generous, it would really be `spi.transfer(tx, NULL, 100, NULL, NULL, 1, 1, 3, 8, 0, 2000000, list)`. None of the interface clarity of printf is preserved for complicated methods. Moving to configuring using a  type rather than method seems to increase clarity after a certain point of complexity.

> In terms of complexity of main.rs, I think that the actual kernel construction should not be our big concern; it's a tiny piece of most main.rs files yet it's a very important thing. This could be as something as simple as breaking it up into three separate traits and passing them in kernel_loop. If we decoupled it, then it might be possible to factor out chip-independent parts into common implementations for policies (e.g., security policies), while keeping common chip-specific ones for stuff like SysTick.

I agree the construction is relatively small, but the clarity for board authors (something many users of Tock are likely to be) may be reduced by not grouping all configuration options in a single place. Splitting up based on what is internally chip-specific leaks kernel implementation structure into how configuration is handled, which may not be intuitive. I think one big KernelResources struct may be confusing as well, but I think this interface should prioritize creating new boards and customizing kernels.

hudson-ayers:
Would the kernel no longer holding a reference to the verifier complicate the process of verifying dynamically loaded apps, down the line?

bradjc:
> Would the kernel no longer holding a reference to the verifier complicate the process of verifying dynamically loaded apps, down the line?

It's a good question. My thinking is that since verification is async, it will have to use the scheduler loop anyway, and since the loop does have access to the kernel config the kernel will have to verify processes in its loop.

ppannuto (commented):
I think I am in favor of this; I find the SPI example above compelling, and architecturally the changes sketched here seem sane.

Assuming 2.1 merges mid-next-week, I think this should be top priority for the core team call, as it's probably the biggest structural thing left in App ID?

phil-levis:
> I think I am in favor of this; I find the SPI example above compelling, and architecturally the changes sketched here seem sane.
> 
> Assuming 2.1 merges mid-next-week, I think this should be top priority for the core team call, as it's probably the biggest structural thing left in App ID?

Yes -- let's talk about it tomorrow.

I am generally in favor of shifting this into `KernelResources` but think this particular configuration approach needs more structure. It's grown pretty organically and is now quite big.

kernel/src/platform/platform.rs
```diff
@@ -31,6 +32,12 @@ pub trait KernelResources<C: Chip> {
     /// the kernel will use.
     type ContextSwitchCallback: ContextSwitchCallback;
 
+    /// Implements the kernel policy for checking cryptographic credentials in
+    /// TBF objects to determine if the contained application can be run as a
+    /// process. If `None` then no credentials are checked and the kernel runs
+    /// all successfully loaded application binaries.
+    type Verifier: process_checking::AppVerifier<'static>;
```
kupiakos, line 39: note: this adds a `'static` requirement that I don't believe was there before, which can be more problematic to test

bradjc:
What should happen if the kernel attempts to start a process that is not unique? Right now this PR puts it in the `CredentialsFailed` state, which may be confusing as this can happen even if no credentials are supplied or verified.

bradjc:
I plan on finishing this PR (specifically updating all boards), but I would like some more clarity on AppUniqueness vs. AppVerification first. Currently I don't understand the coupling, what configuration options a board author should have, and what happens if an app is not unique.

PR #3125, opened 2022-08-04
Tock Implementation for HiFive Inventor

TeodoraMiu:
### Pull Request Overview

This draft pull request is a first attempt at developing Tock compatibility with the HiFive Inventor board (https://www.hifiveinventor.com/user-guide/overview). 
I have used the hifive1 board implementation as guidance and changed the memory layout to correspond to the Hifive Inventor's FE310-G003 processor. I also made some other changes, such as the baud rate of the serial port, to match the new board. 
At the moment, I have successfully connected the board via JLink and flashed the kernel to the board. However, I am having some difficulties with the board's debugger. Debug messages are not shown, so I tried to display a panic and it seems to double panic in the panic handler. Here is the output I received (after the 'tockloader listen' command):

```
Sleep BK1 0x00000000 BK15 0xbed0bed0
After BK1 0x00000000 BK15 0xbed0bed0

PMU IE 0x00000005 
Bootloader version: 2.40 09/10/2020 12:17:19
Sleep BK1 0x00000000 BK15 0xbed0bed0
After BK1 0x00000000 BK15 0xbed0bed0

PMU IE 0x00000005 CAUSE 0x00000100 BK1 0x00000000 BK15 0xbed0bed0

panicked at 
panicked at 'fatal exception', chips/e310x/src/chip.rs:181:13
        Kernel version release-2.0-820-ga8c97d5b7

---| Global debug wl `set_debug_writer_wrapper` in board initialization.

---| RISC-V Machine State |---
Last cause (mcause): Store/AMO access fault (interrupcode=0x00000007)
Last value (mtval):  0x7FFFFFFC

System register dump:
 mepc:    0x20010514    mstatus:     0x00001880
 mcycle:  0x08251Fet:    0x048662DE
 mtvec:   0x20010100
 mstatus: 0x00001880
  uie:    false  upie:   false
  sie:    false  spie:   false
  mie:    falselse               false 
  msoft:  true                false 
  utimer: false               false 
  stimer: false               false 
       false 
  sext:   false               false 
  mext:   true                false 

---| App Status |---
```

I am unsure why this is happening. @alistair23 I think you were the one to add the hifive1 board, any help from you would be greatly appreciated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ members = [
     "boards/esp32-c3-devkitM-1",
     "boards/clue_nrf52840",
     "boards/hail",
+    "boards/hifive_inventor",
     "boards/hifive1",
     "boards/imix",
     "boards/imxrt1050-evkb",
--- a/boards/hifive_inventor/Cargo.toml
+++ b/boards/hifive_inventor/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "hifive_inventor"
+version = "0.1.0"
+authors = ["Tock Project Developers <tock-dev@googlegroups.com>"]
+build = "build.rs"
+edition = "2021"
+
+[dependencies]
+components = { path = "../components" }
+rv32i = { path = "../../arch/rv32i" }
+capsules = { path = "../../capsules" }
+kernel = { path = "../../kernel" }
+e310x = { path = "../../chips/e310x" }
+sifive = { path = "../../chips/sifive" }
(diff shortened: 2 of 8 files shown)


alistair23:
I would try moving the panic further along in `main()` it looks like you are getting the `Call `set_debug_writer_wrapper` in board initialization.` error, which means you are trying to print before `set_debug_writer_wrapper()` gets called.

If you aren't getting output, one trick is to disable all UART configuration in Tock (comment it out in `chips`). It looks like the previous stage is using the UART so you probably don't need to configure the UART to have it work. If that works you can then work backwards to re-enable the configuration and see where it stops working.

We eventually need Tock to configure the UART, but for testing it can be useful to try that

bradjc:
Moved to #3225

PR #3127, opened 2022-08-04
WIP: Remove Dynamic Deferred Call w/o changing deferred call

hudson-ayers:
### Pull Request Overview

This pull request is another alternative to https://github.com/tock/tock/pull/3123 and #3086 . Rather than modify DefferredCall, it introduces a new type in the kernel, `CapsuleDeferredCall`, for use only by capsules.
This PR keeps the use of globals and atomics, allowing a less annoying API, at the expense of keeping us on nightly Rust until we reach an agreement about how to provide atomics without using LLVM intrinsics. This PR also seems to be a little smaller (~400 bytes) than #3123, and seems to produce about 100 bytes of savings per `DynamicDeferredCall` removed.

As with the other draft PRs, this is only implemented for Imix + the radio driver capsule.

### Testing Strategy

### TODO or Help Wanted

This pull request still needs implementations for more than the radio capsule, plus a lot of cleanups of names etc.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/ieee802154.rs
+++ b/boards/components/src/ieee802154.rs
@@ -13,7 +13,6 @@
 //!     &nrf52::aes::AESECB,
 //!     PAN_ID,
 //!     SRC_MAC,
-//!     deferred_caller,
 //! )
 //! .finalize(components::ieee802154_component_helper!(
 //!     nrf52::ieee802154_radio::Radio,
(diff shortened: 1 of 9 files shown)


hudson-ayers:
Based on the core call discussion of this PR, I realized that it does not offer a good way for out-of-tree boards to use our capsules and introduce out-of-tree capsules with deferred calls. I am working on a new design to accommodate this use case, and closing this PR.

PR #3129, opened 2022-08-05
DRAFT: basic capsule unit test

kupiakos:
This doesn't solve everything in the way I'd like, since there are a few challenges that need to be worked out:

- There's no way to create a fake `ProcessId` to pass into `SyscallDriver::command`
  - This can't be done as-is without access to either `unsafe` (rejected by `deny`) or test-only capabilities
- There's no way to ensure a `CommandReturn` contains the expected value. IMO `CommandReturn::into_inner` should be `pub`, not `pub(crate)`.
- There's no easy way to safely create a `&'static` or `&'static mut` to a non-`Sync` type without leaking a `Box`. There are a couple of utility data structures that could help with this:
  - `OnceMut<T>`, a `Sync` wrapper around `T: ?Send + ?Sync` with a `fn get(&self) -> &mut T`, which allows you to retrieve a `&'a mut T`, but only _once_ for the lifetime of the object. This lets you put a `OnceMut` in a `static` directly and use it safely to create a `&static mut T`. This can be `#[no_std]` and could be a safe, but slight cost, replacement for `static_init`. It's possible the optimizer can make this zero-cost with LTO determining only one direct access.
  - `AppendVec<T>` (better name TODO), which is essentially just a `Sync` arena, conceptually a `OnceMut` but for multiple items. As a test-only library, it can just be something that leaks `Box`es 

This PR should also do more testing on the `Led` capsule itself, as it doesn't exercise every codepath.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,6 +163,7 @@ allstack stack stack-analysis:
 
 .PHONY: licensecheck
 licensecheck:
+	$(call banner,License checker)
 	@cargo run --manifest-path=tools/license-checker/Cargo.toml --release
 
 ## Commands
--- a/tools/license-checker/Cargo.toml
+++ b/tools/license-checker/Cargo.toml
@@ -11,6 +11,7 @@ edition.workspace = true
 
 [dependencies]
 clap = { features = ["derive"], version = "4.0.29" }
+colored = "2.0.1"
 ignore = "0.4"
 thiserror = "1.0.37"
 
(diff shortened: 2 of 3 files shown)


ppannuto:
> There's no way to ensure a CommandReturn contains the expected value. IMO CommandReturn::into_inner should be pub, not pub(crate).

I believe we discussed this briefly at TockWorld and agreed it would be find to switch to `pub`. It was just `pub(crate)` previously for least-privilege reasons (i.e. nothing needed it). This PR can include this change IMO.

kupiakos:
`OnceMut` appears to be the optionally thread-safe version of [`cortex_m::singleton!`](https://docs.rs/cortex-m/latest/cortex_m/macro.singleton.html)

PR #3134, opened 2022-08-09
use deferred calls to report aborted reception in sam4l uart

hudson-ayers:
### Pull Request Overview

This pull request fixes the implementation of `abort_rx()` in `sam4l/src/usart.rs` to not call an upcall within a downcall. This fixes `console_timeout` to not panic. We had not hit this bug in the past because the console capsule only recently was modified to call `abort_rx` from within a grant region, as a code size optimization. Notably, `abort_tx` still calls an upcall in a downcall, but this does not break any of our existing applications, and I would prefer to let fixing that go through as part of the switch to the new UART HIL in https://github.com/tock/tock/pull/3046 . 

This PR fixes one of the two release-blocking failures I found as part of my initial Imix release testing.

This is the first time (I think?) that we have had to implement a deferred call for a peripheral with a single implementation for multiple hardware peripherals. This forced me to use a semi-hacky "match on the base address of the peripheral" approach for mapping deferred calls. Open to other suggestions on this element of this PR, but I think my approach is probably fine.

### Testing Strategy

This pull request was tested by running the libtock-c apps `console_timeout` and `console_recv_short` on Imix.

### TODO or Help Wanted

N/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/sam4l/src/chip.rs
+++ b/chips/sam4l/src/chip.rs
@@ -229,6 +229,10 @@ impl InterruptService<Task> for Sam4lDefaultPeripherals {
         match task {
             crate::deferred_call_tasks::Task::Flashcalw => self.flash_controller.handle_interrupt(),
             crate::deferred_call_tasks::Task::CRCCU => self.crccu.handle_deferred_call(),
+            crate::deferred_call_tasks::Task::Usart0 => self.usart0.handle_deferred_call(),
+            crate::deferred_call_tasks::Task::Usart1 => self.usart1.handle_deferred_call(),
+            crate::deferred_call_tasks::Task::Usart2 => self.usart2.handle_deferred_call(),
+            crate::deferred_call_tasks::Task::Usart3 => self.usart3.handle_deferred_call(),
         }
         true
     }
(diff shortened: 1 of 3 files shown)


phil-levis:
Looks good to me. I wish there were a way to force an interrupt from the USART.

phil-levis (approved):

ppannuto (approved):
This looks good to me as a minimal fix to address this for the upcoming release.

ppannuto:
> I wish there were a way to force an interrupt from the USART.

With the caveat of never having had occasion to write code that does it, I'm pretty sure you can trigger any interrupt you like:

image
image

Some light googling confirms others have used that to emulate / generate "hardware" interrupts. As this isn't running in (higher-priority or any) exception context, the interrupt would fire immediately; not sure if that's what we want, as it's more overhead to effectively do the same thing of scheduling a task (though it is one less code path I suppose).

hudson-ayers:
Our (undocumented?) policy for chip drivers has generally been that they should only use unsafe to declare the address of their memory-mapped registers, not for anything else. I think that allowing arbitrary chip drivers to trigger arbitrary NVIC interrupts probably runs afoul of this policy. It might make sense to find some way to safely allow chip peripheral drivers to trigger only the NVIC ID associated with that particular peripheral

phil-levis:
We can use a deferred call now and then later if we change our mind about manually triggering interrupts can make that change.

PR #3135, opened 2022-08-09
LEDs array in LED Driver does not need to be mut

brghena:
### Pull Request Overview

The array of LEDs in LED Driver does not need to be a mutable reference.

The LEDs themselves were not mutable, only the array. But nothing should ever be changing to contents of that array, and the LED Driver does not do so. This seems to be legacy from when we used to have LEDs in a TakeCell.

### Testing Strategy

This pull request was tested by compiling all boards and testing the Blink app on an nRF52840DK.

### TODO or Help Wanted

Should be good to go.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


ppannuto (approved):
While pretty benign, I suppose this should probably begin our queue of things that will come in post-code-freeze.

hudson-ayers (approved):

alistair23 (approved):

twilfredo (approved):

PR #3136, opened 2022-08-10
Fix secondary alarm from firing immediately during alarm callback

jettr:
If you have two alarms and one sets the other in its callback, and the
second alarm is later in the linked list of virtual alarms, then the
second alarm callback will fire immediately.

### Pull Request Overview

We need get the latest "now" when comparing the reference time while
going through the list of virtual alarms, otherwise the assumption
that reference is always in the past breaks down and falsely
triggers alarms set during previous alarms callback.

### Testing Strategy

-  [x] Tested on real device that was exibiting this incorrect behavior
-  [x] Added unit test that verifes new behavior (failed without new change in this PR)

### Documentation Updated

- [x] None

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/virtual_alarm.rs
+++ b/capsules/src/virtual_alarm.rs
@@ -238,8 +238,6 @@ impl<'a, A: Alarm<'a>> time::AlarmClient for MuxAlarm<'a, A> {
     /// When the underlying alarm has fired, we have to multiplex this event back to the virtual
     /// alarms that should now fire.
     fn alarm(&self) {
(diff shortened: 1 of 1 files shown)


phil-levis (approved):
Looks good. Makes sense. Nice  catch!

bradjc (approved):

bradjc:
This seems like a bug fix we can merge for rc1.

ppannuto (approved):
+1 for including in 2.1 release

PR #3137, opened 2022-08-10
scheduler: round_robin: prevent infinite loop

jettr:
### Pull Request Overview

If the kernel says there is app work left, but none of the application are ready, then prevent the round robin scheduler from looping infinitely.

This is defensive coding in the sense that if the kernel work count ever is incorrect, then this loop go on indefinitely. I had local changes that made it seem like there was work left but there were no apps ready.

### Testing Strategy

- [x] Verified on actual device that was encountering infinite loop

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/scheduler/round_robin.rs
+++ b/kernel/src/scheduler/round_robin.rs
@@ -74,7 +74,18 @@ impl<'a, C: Chip> Scheduler<C> for RoundRobinSched<'a> {
 
             // Find next ready process. Place any *empty* process slots, or not-ready
             // processes, at the back of the queue.
+            let mut first_head = None;
             for node in self.processes.iter() {
+                // Ensure we do not loop forever if all processes are not not ready
+                match first_head {
+                    None => first_head = Some(node),
+                    Some(first_head) => {
+                        // We make a full iteration and nothing was ready. Try to sleep instead
+                        if core::ptr::eq(first_head, node) {
+                            return SchedulingDecision::TrySleep;
+                        }
+                    }
+                }
                 match node.proc {
                     Some(proc) => {
                         if proc.ready() {


hudson-ayers (approved):
I have been bitten by this before when modifying the kernel work counting, and it's definitely a pain to debug this sort of deadlock. On the other hand, this would allow kernel work mis-counting to go undetected, which also doesn't seem great. Interested to see what others think of the trade off here

brghena:
As a point of information, theoretically, the prior "deadlock" behavior should result in a Watchdog timeout, if the board provides a watchdog. https://github.com/tock/tock/blob/master/kernel/src/kernel.rs#L448

jettr:
If you were using another type of scheduler, you wouldn't necessarily get an infinite loop, right? I don't think a project's scheduler type should dictate whether you see this.

If we want to catch the case where the kernel work is not balanced, we should prefer an explicit check instead of relying on an implicit dead lock in my opinion.

jettr:
> As a point of information, theoretically, the prior "deadlock" behavior should result in a Watchdog timeout, if the board provides a watchdog. https://github.com/tock/tock/blob/master/kernel/src/kernel.rs#L448

That is how we caught it is practice, it was somewhat of a pain to track down what was causing the watch dog reset.

ppannuto:
I agree not deadlocking is probably the better behavior, and similarly `panic`'ing probably isn't correct either as this doesn't necessarily mean the whole system is bork'ed. But, if I'm understanding correctly, falling into this check is always indicative of a (non-fatal) bug. Seems there should at least be a `debug` message printed on this code path.

Perhaps we need some instrumentation to log/report `kernel::debug::non_fatal_error("This weird thing happened")`, but that's a bigger bit of machinery. In the short term a plain old `debug` here seems appropriate to me.

jettr:
I was working on a change related to having processes that had pending upcalls, but were not schedulable because they were waiting on something else (nothing related to this is even in a PR yet).

To me the kernel_work is an optimization, if it is 0, then there is no work, but if it is non 0, then there _might_ be work. Each process should be able to say it is not ready even if there is still some kernel work it has but it can't do it right now.

ppannuto:
The feels like [`processes_blocked`](https://github.com/tock/tock/pull/3137/files#diff-a678b193858876ccfb7bf72e14496fe0468cf3f4ca306b6a133ea80572863630R68) is either mis-named, or mis-implemented then. IIUC, you're only in that code path because at the beginning of the scheduling decision, the kernel thought there existed process that was not blocked.

jettr:
This bug wouldn't' even be possible if we didn't have to have interior mutability everywhere. We should not be able to mutate the linked list why we are iterating over it. This is just a bad idea in general.

jettr:
I just took another look at the other schedulers and they all need to be updated to handle the case where there is no process available when `!kernel.processes_blocked()`.

What is the history reasoning behind tracking `kernel.processes_blocked` with a counter? Is it for optimization to easily know if there is nothing to do that we don't need to even look at any of the processes for attempted scheduling?

What would be the down side of removing the `kernel.processes_blocked` from all of the scheduler implementations and just check if the next process (according to the scheduler type) is available or not?

hudson-ayers:
I agree this change stands on its own, and I think we have generally tried to avoid adding debug prints to the core kernel. I think a discussion of the precise meaning of processes_blocked could wait until any follow up PR. My understanding of the current implementation and work tracking (without looking back) is that it's mostly a result of the historical evolution of the scheduler that made this easiest to implement in a performant way.

kernel/src/scheduler/round_robin.rs
```diff
@@ -74,7 +74,18 @@ impl<'a, C: Chip> Scheduler<C> for RoundRobinSched<'a> {
 
             // Find next ready process. Place any *empty* process slots, or not-ready
             // processes, at the back of the queue.
+            let mut first_head = None;
             for node in self.processes.iter() {
+                // Ensure we do not loop forever if all processes are not not ready
```
kupiakos, line 79: extra `not`

kernel/src/scheduler/round_robin.rs
```diff
@@ -74,7 +74,18 @@ impl<'a, C: Chip> Scheduler<C> for RoundRobinSched<'a> {
...
+                match first_head {
+                    None => first_head = Some(node),
+                    Some(first_head) => {
+                        // We make a full iteration and nothing was ready. Try to sleep instead
```
kupiakos, line 83: `made`

kupiakos:
I think this change stands well on its own, though I do like the idea of keeping track of this situation occurring. Maybe a low level debug?

ppannuto:
Happy to defer to consensus here. @jettr, could you open an issue to discuss the more general question of the utility and intent of `processes_blocked()`? Then I think we can merge this.

bradjc (changes requested):
> To me the kernel_work is an optimization, if it is 0, then there is no work, but if it is non 0, then there _might_ be work. Each process should be able to say it is not ready even if there is still some kernel work it has but it can't do it right now.

That isn't the meaning of having work. If the process doesn't want the kernel to run something, then it shouldn't be enqueuing a task to run. 


> What is the history/reasoning behind tracking `kernel.processes_blocked` with a counter? Is it for optimization to easily know if there is nothing to do that we don't need to even look at any of the processes to attempt scheduling?
> 
> What would be the down side of removing the `kernel.processes_blocked` from all of the scheduler implementations and just check if the next process (according to the scheduler type) is available or not?

`kernel.work: usize` is an optimization that avoids needing to scan the entire array of processes before deciding to sleep or not.


> I have been bitten by this before when modifying the kernel work counting, and it's definitely a pain to debug this sort of deadlock. On the other hand, this would allow kernel work mis-counting to go undetected, which also doesn't seem great. Interested to see what others think of the trade off here

I'm not particularly in favor of the change in this PR. If `processes_blocked()` is not in sync with processes that have tasks to run, that is a bug in the kernel, and one that we want to know about. Making this change seems like a way to make it very difficult to find a hypothetical future bug that makes kernel.work out of sync.

However, I'm not _necessarily_ defending how tasks are queued, how processes_blocked() is implemented, and what having something in the task queue means. I think we could revisit that design (it is carried over from much earlier implementations of the kernel) since we haven't really discussed and iterated on it. But I think we have to clarify these issues first, before we can assess a change like what's in this PR. Right now, this change seems wrong, because `processes_blocked()` cannot return false unless some process is ready to run.

jettr:
So taking a step back, I opened #3138 for further discussion about removing the increment/decrement work and `processes_blocked `concepts altogether as I think it adds complexity and failure points, but it does not give us that much benefit. Once we come to a consensus over there, I will update this PR.

jettr:
Closing in favor of #3158

Issue #2425, opened 2021-02-10
RFC: Fixing unnecessary stack use on all platforms


hudson-ayers:
The new peripheral instantiation approach has forced us to increase the stack size on several platforms. Here is why:

`static_init!()` initializes items on the stack before copying them into the relevant global. As a result, when the peripherals are created using `static_init!(Sam4lDefaultPeripherals, Sam4lDefaultPeripherals::new(pm))`, they are allocated on the stack in `reset_handler()`. Unfortunately, `reset_handler()` never returns, so as far as I can tell all local variables in `reset_handler()` take up stack space forever. Notably, this issue actually applies to *everything* we allocate using `static_init!()` directly in `reset_handler()` -- the new peripheral approach just exposed it because the peripherals are big.

**In summary: using `static_init!()`in `reset_handler()` uses twice as much memory as the thing being statically allocated: the memory for the global is used, and the stack memory is used forever**

Another note: `lazy_static` and `Box::new()` have this same issue, though of course they do not affect us.

I have thought of 4 potential solutions to this, and would like people's thoughts on each.

## Solutions

I have been able to come up with several solutions, each with a unique set of tradeoffs.

### Solution 1
By convention, create peripherals in a separate, `#[inline(never)]` function:
```rust
#[inline(never)]
unsafe fn create_peripherals(
    pm: &'static sam4l::pm::PowerManager,
) -> &'static mut Sam4lDefaultPeripherals {
    static_init!(Sam4lDefaultPeripherals, Sam4lDefaultPeripherals::new(pm))
}

#[no_mangle]
pub unsafe fn reset_handler() {
    sam4l::init();
    let pm = static_init!(sam4l::pm::PowerManager, sam4l::pm::PowerManager::new());
    let peripherals = create_peripherals(pm);
    ...
```
Using `-Z emit-stack-sizes`, this drops the stack frame required for `reset_handler()` from 1496 bytes --> 248 bytes.

This also prevents stack overflows in practice. If I add a 6450 byte stack allocation to the end of reset_handler, the kernel still runs with this code on the board, but there is a stack overflow with the current approach.

Pros: 
- stack space used to allocate the peripherals is freed immediately

Cons:
- stack still has to be at least as large as the `DefaultPeripherals` struct
- gross boilerplate for creating peripherals with non-obvious reason for why it is needed

### Solution 2
[...]

Issue #3109, opened 2022-07-26
Curious Timing Causes Failed Syscall on Cortex-M4 MCU


bradjc:
Originally discovered in #3076.

### Observation

With a very specific kernel (on commit f6adf9d536c92b450c61430b002f8a17b9bcb286) compiled on linux, running on imix, with the `examples/tests/analog_comparator` app (commit f5e164366028b5633006299ac136ba5bf0e0db74) as the only app running, the fourth syscall (a memop sbrk) fails in some way. In code, libtock-c memop() seems to get an invalid return code from the kernel (`r0==1`, which isn't a valid memop return code).

### Comments

Observing this bug seems to take a VERY specific order of instructions and timing. The bug is very repeatable (Hudson and I can both observe it, on multiple imixes, over many trials, and it happens every time). However, modifying the kernel OR the app, OR the number of apps all cause it to go away. This suggests that the particular timing quirk of when the analog_comparator app calls sbrk matters significantly. For example, compiling the same kernel on mac and the issue goes away. Adding a second app (and lengthing the boot process), and the issue goes away. Adding even a single instruction before the `svc 5` (memop) call in the app and the issue goes away.

bradjc:
To reproduce:
- Need Hudson's imix kernel (or to compile on linux, not mac).
- Run _only_ the analog comparator test app
- The app calls exit when doing its first sbrk, and gets stuck in a while 1 loop in the exit function. The symptom appears to be the memop call returns an invalid return (fail with value, `r0 == 1`). However, the kernel does not record that the app ever called memop(1) (sbrk), so its not clear what is happening.

Also note:
- sbrk uses r0=1, and the error is happening because r0==1 "after" the memop runs. so could be that r0 is not changing (ie not being set by the kernel).


Giving up for now, but I was able to make two versions of the analog comparator test app that differ in one instruction where one causes the error and the other doesn't. Basically inserting one meaningless instruction before the `svc 5` memop call and the problem goes away (insert it after the svc 5 and still broken):

```
diff cortex-m4.userland_debug.lst.WORKING_MINOR_CHANGE cortex-m4.userland_debug.lst.BROKEN_MINOR_CHANGE
1962,1963c1962,1963
<    403c6:	4684      	mov	ip, r0
<    403c8:	df05      	svc	5
---
>    403c6:	df05      	svc	5
>    403c8:	4684      	mov	ip, r0
```

```
diff --git a/libtock/tock.c b/libtock/tock.c
[...]

PR #3139, opened 2022-08-12
boards/redboard_artemis: Workaround Hard Fault Exception

alistair23:
### Pull Request Overview

Fixup a hardfaul exception that is seen on every non-test boot of Tock.

This change ensures that the `svc` call in `apollo3::init()` is called 
inline with `main()`. I'm not clear why, but if the `svc` is called 
after a function call (a `push` op) we end up with a corrupted call 
stack.

We need to call the `svc` op to ensure the FPU is correctly disabled (as
the boot loader enables it).

While we are changing the inline let's also never inline `setup()` to 
reduce the stack size usage (see 
https://github.com/tock/tock/issues/2425).

```
panicked at 'Kernel HardFault.
        Kernel version release-2.1-rc1
        r0  0x10001ab8
        r1  0x10001da0
        r2  0x10000fe8
        r3  0x10001e08
        r12 0x10001acc
        lr  0x10001d70
        pc  0x0
        prs 0x0 [ N 0 Z 0 C 0 V 0 Q 0 GE 0000 ; ICI.IT 0 T false ; Exc 0-Thread Mode ]
        sp  0x10000f80
        top of stack     0x40008000
        bottom of stack  0x10000000
        SHCSR 0x0
        CFSR  0x20000
        HSFR  0x40000000
        Instruction Access Violation:       false
        Data Access Violation:              false
        Memory Management Unstacking Fault: false
        Memory Management Stacking Fault:   false
        Memory Management Lazy FP Fault:    false
        Instruction Bus Error:              false
        Precise Data Bus Error:             false
        Imprecise Data Bus Error:           false
        Bus Unstacking Fault:               false
        Bus Stacking Fault:                 false
        Bus Lazy FP Fault:                  false
        Undefined Instruction Usage Fault:  false
        Invalid State Usage Fault:          true
        Invalid PC Load Usage Fault:        false
        No Coprocessor Usage Fault:         false
        Unaligned Access Usage Fault:       false
        Divide By Zero:                     false
        Bus Fault on Vector Table Read:     false
        Faulting Memory Address: (valid: false) 0xE000ED34
        Bus Fault Address:       (valid: false) 0xE000ED38
', arch/cortex-m/src/lib.rs:496:5
        Kernel version release-2.1-rc1
```

Fixes: e822de1742 "boards/redboard_artemis_nano: Setup initial tests"

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/redboard_artemis_nano/src/main.rs
+++ b/boards/redboard_artemis_nano/src/main.rs
@@ -151,14 +151,13 @@ impl KernelResources<apollo3::chip::Apollo3<Apollo3DefaultPeripherals>> for Redb
     }
 }
 
+#[inline(never)]
 unsafe fn setup() -> (
     &'static kernel::Kernel,
     &'static RedboardArtemisNano,
     &'static apollo3::chip::Apollo3<Apollo3DefaultPeripherals>,
     &'static Apollo3DefaultPeripherals,
 ) {
-    apollo3::init();
-
     let peripherals = static_init!(Apollo3DefaultPeripherals, Apollo3DefaultPeripherals::new());
 
     // No need to statically allocate mcu/pwr/clk_ctrl because they are only used in main!
@@ -394,6 +393,8 @@ unsafe fn setup() -> (
 /// setup and RAM initialization.
 #[no_mangle]
 pub unsafe fn main() {
+    apollo3::init();
+
     #[cfg(test)]
     test_main();
 
--- a/boards/redboard_artemis_nano/src/main.rs
+++ b/boards/redboard_artemis_nano/src/main.rs
@@ -151,6 +151,9 @@ impl KernelResources<apollo3::chip::Apollo3<Apollo3DefaultPeripherals>> for Redb
     }
 }
 
+// WARN: This is a short-term patch applied to allow this board to boot for the
+// 2.1 release. This `inline` will be removed immediately after the 2.1 release
+// pending the fixes to cortex-m context switching which should come in 2.2.
 #[inline(never)]
 unsafe fn setup() -> (
     &'static kernel::Kernel,
@@ -158,6 +161,13 @@ unsafe fn setup() -> (
     &'static apollo3::chip::Apollo3<Apollo3DefaultPeripherals>,
     &'static Apollo3DefaultPeripherals,
 ) {
+    // WARN: This is a short-term patch applied to allow this board to
+    // boot for the 2.1 release. This will be returned here immediately
+    // after the 2.1 release pending the fixes to cortex-m context switching
+    // which should come in 2.2.
+    //
+    // apollo3::init();
+
     let peripherals = static_init!(Apollo3DefaultPeripherals, Apollo3DefaultPeripherals::new());
 
     // No need to statically allocate mcu/pwr/clk_ctrl because they are only used in main!
@@ -393,6 +403,10 @@ unsafe fn setup() -> (
 /// setup and RAM initialization.
 #[no_mangle]
 pub unsafe fn main() {
+    // WARN: This is a short-term patch applied to allow this board to
+    // boot for the 2.1 release. `init` will be removed here immediately
+    // after the 2.1 release pending the fixes to cortex-m context switching
+    // which should come in 2.2.
     apollo3::init();
 
     #[cfg(test)]
--- a/chips/apollo3/src/lib.rs
+++ b/chips/apollo3/src/lib.rs
@@ -71,6 +71,7 @@ pub static IRQS: [unsafe extern "C" fn(); 32] = [CortexM4::GENERIC_ISR; 32];
 pub static PATCH: [unsafe extern "C" fn(); 16] = [unhandled_interrupt; 16];
 
 #[cfg(all(target_arch = "arm", target_os = "none"))]
+#[inline(always)]
 pub unsafe fn init() {
     use core::arch::asm;
     let cache_ctrl = crate::cachectrl::CacheCtrl::new();
--- a/chips/apollo3/src/lib.rs
+++ b/chips/apollo3/src/lib.rs
@@ -70,6 +70,10 @@ pub static IRQS: [unsafe extern "C" fn(); 32] = [CortexM4::GENERIC_ISR; 32];
 #[cfg_attr(all(target_arch = "arm", target_os = "none"), used)]
 pub static PATCH: [unsafe extern "C" fn(); 16] = [unhandled_interrupt; 16];
 
+// WARN: This is a short-term patch applied to allow this board to
+// boot for the 2.1 release. This `inline` will be removed immediately
+// after the 2.1 release pending the fixes to cortex-m context switching
+// which should come in 2.2.
 #[cfg(all(target_arch = "arm", target_os = "none"))]
 #[inline(always)]
 pub unsafe fn init() {


phil-levis:
Why is this happening? The PC=0. which suggests that something tried to set the PC to 0, which failed because the LSB=0? 

Please document what the change is in the description.

phil-levis (approved):

bradjc (changes requested):
Isn't this the exact opposite of why we added the `setup()` functions? The idea is to ensure that all of the setup code uses its own stack frame so the stack space is returned before the kernel loop starts.

It seems like this is masking whatever the actual failure is, or I'm missing something.

alistair23:
I'm not clear on why the hard fault happens or why this fixes it. I don't have good debugging capabilities, so I can't inspect what is going on. If anyone else has a J-Link probe they might be able to figure out what is happening.

This does seem to be the opposite of what other boards where doing with the `setup()` function, which is strange. But this is the only change that I could find that actually works.

phil-levis:
If the difference is "doesn't work without this change" and "does work with this change", then I'm in favor of making it. But I 100% agree with Brad that understanding **why** is critically important. On a CortexM, this a particular fault is when the PC (either through a `b` or `ld`) is loaded with a value whose LSB is not 1 (so it's not a Thumb instruction). 

Looking at the dump, the `pc` is `0x00000000`. This looks like stack corruption; `lr` was pushed to the stack, some method corrupted the stack, then `pc` was restored with a bad value. Changing inlining fixing this also indicates this might be the issue. Finally, it also looks weird that `lr` is `0x10001d70`; that's a RAM address, and doesn't have LSB=1; as `lr` is often saved and restored from the stack, this again points at stack corruption.

alistair23:
Yeah, I also think it's stack corruption. I tried increasing the stack and that didn't make a difference though. I'm a little out of ideas of what else to try. I'm trying to see if I can get a J-link debugger, but until then I'm a little stuck on how to debug it. My cheapy debugger doesn't work well enough to catch this unfortunately

phil-levis:
I would look at the printed out registers and see what you can figure out from them. Those are all stack addresses? What do you expect to be on the stack at those addresses? You can in the hard fault handler print out some more information. Yeah, this is a methodical embedded systems debugging problem.

alistair23:
I tried that, nothing interesting stands out. They are just buffers from `setup()` on the stack:

```
        r0  0x10001ab8 .sram  00000014 _ZN21redboard_artemis_nano5setup3BUF17h6fb3132038a37d9aE
        r1  0x10001da0 .sram  00000030 _ZN21redboard_artemis_nano5setup3BUF17h80b91a79e4208070E
        r2  0x10000fe8
        r3  0x10001e08 .sram  0000003c _ZN21redboard_artemis_nano5setup3BUF17hbd865345f081752dE
        r12 0x10001acc .sram  00000018 _ZN21redboard_artemis_nano5setup3BUF17hb291be63825f8d02E
        lr  0x10001d70 .sram  00000030 _ZN21redboard_artemis_nano5setup3BUF17hf2bf023d415da105E
        pc  0x0
```

It seems to happen at the return of `setup()`, but no idea why

phil-levis:
> I tried that, nothing interesting stands out. They are just buffers from `setup()`:
> 
> ```
>         r0  0x10001ab8 .sram  00000014 _ZN21redboard_artemis_nano5setup3BUF17h6fb3132038a37d9aE
>         r1  0x10001da0 .sram  00000030 _ZN21redboard_artemis_nano5setup3BUF17h80b91a79e4208070E
>         r2  0x10000fe8
>         r3  0x10001e08 .sram  0000003c _ZN21redboard_artemis_nano5setup3BUF17hbd865345f081752dE
>         r12 0x10001acc .sram  00000018 _ZN21redboard_artemis_nano5setup3BUF17hb291be63825f8d02E
>         lr  0x10001d70 .sram  00000030 _ZN21redboard_artemis_nano5setup3BUF17hf2bf023d415da105E
>         pc  0x0
> ```
> 
> It seems to happen at the return of `setup()`, but no idea why

It happens at the return of `setup` because that's when `lr` is popped from the stack, and if the stack is corrupted, `lr` has the wrong value (0x0) and `b lr` makes `pc=0x0`. I would modify the panic handler to print out/display the stack. It's very tough to figure out what stack corruption wrote 0x0 to where `lr` was stored, but if you look at the values around it you might be able to figure out what happened.

phil-levis:
@alistair23 and I talked about this on the OT call today. It sounds like this happens with *any* number of peripherals, and so Alistair is going to strip `setup` down to just be the UART, as the assembly will be easier to read and understand. He's also going to modify the hard fault handler to print out the callee-saved registers (r4-r7) as these are adjacent to lr on the stack frame in the initial stack push. Their values may therefore give a hint on what is overwriting the stored link register. Hopefully with the simpler method, it will be easier to figure out which static structure is being placed right after these pushed registers and is causing them to be corrupted.

alistair23:
I did make some progress. I have updated the PR description with the new information. It seems to be due to the `svc` op in the `apollo3::init()`.

The tricky part is that I did try to not call it earlier, and still saw a hard fault. But after looking more closely I realised it was a different hard fault as the FPU wasn't disabled. Inlining the `svc` fixes the hard fault, although I'm not clear why

phil-levis:
Normally, in the Tock kernel, `svc` denotes either a switch from kernel space to user space, or a switch from user space to kernel space. Since `apollo3::init` is running in the kernel, then it needs to be the former. Has the PSP been set up? What do you expect to happen with the `svc`?

alistair23:
Hmmm... My understanding was that [`SVC`](https://developer.arm.com/documentation/dui0489/c/arm-and-thumb-instructions/miscellaneous-instructions/svc) causes an exception that will always jump to the supervisor mode (where Tock is running).

That should cause Tock to execute `svc_handler_arm_v7m()`. At which point the link register should be something other then `0xfffffff9`. So we should set `SYSCALL_FIRED` (which would be incorrect), but then continue running.

Life would be better without the `svc` call, but unfortunately the FPU doesn't seem to be disabled correctly without it.

I had one thought, that somehow the push/pop operation changes if the FPU is enabled/disabled. That would explain why inlining/not-inlining makes a difference. The [compiled code changes](https://stackoverflow.com/questions/27116011/arm-cortex-m4-hard-fault-floating-point), but I wasn't sure about the actual CPU operations.

I tried this diff. To ensure we disable the FPU before running any Rust

```diff
diff --git a/arch/cortex-m/src/lib.rs b/arch/cortex-m/src/lib.rs
index 4dd1034e5..dfa1093d4 100644
--- a/arch/cortex-m/src/lib.rs
+++ b/arch/cortex-m/src/lib.rs
@@ -356,6 +356,13 @@ pub unsafe extern "C" fn initialize_ram_jump_to_main() {
 
   201: // data_init_done
 
+    ldr.w r0, =0xE000ED88       // CPACR is located at address 0xE000ED88
+    ldr r1, [r0]                // Read CPACR
+    and r1, r1, #(0xFF0FFFFF)    // Clear bits 20-23 to disable CP10 and CP11 coprocessors
+    str r1, [r0]               // Write back the modified value to the CPACR
+    dsb
+    isb
+
     // Now that memory has been initialized, we can jump to main() where the
     // board initialization takes place and Rust code starts.
     bl main
diff --git a/arch/cortex-m/src/scb.rs b/arch/cortex-m/src/scb.rs
index 677547afe..dd9155d0f 100644
--- a/arch/cortex-m/src/scb.rs
+++ b/arch/cortex-m/src/scb.rs
@@ -299,10 +299,10 @@ pub unsafe fn set_vector_table_offset(offset: *const ()) {
 #[cfg(all(target_arch = "arm", target_os = "none"))]
 pub unsafe fn disable_fpca() {
     use core::arch::asm;
-    SCB.cpacr
-        .modify(CoprocessorAccessControl::CP10::CLEAR + CoprocessorAccessControl::CP11::CLEAR);
+    // SCB.cpacr
+    //     .modify(CoprocessorAccessControl::CP10::CLEAR + CoprocessorAccessControl::CP11::CLEAR);
 
-    asm!("dsb", "isb", options(nomem, nostack, preserves_flags));
+    // asm!("dsb", "isb", options(nomem, nostack, preserves_flags));
 
     if SCB.cpacr.read(CoprocessorAccessControl::CP10) != 0
         || SCB.cpacr.read(CoprocessorAccessControl::CP11) != 0
diff --git a/chips/apollo3/src/lib.rs b/chips/apollo3/src/lib.rs
index e36bd40dc..feaf08e70 100644
--- a/chips/apollo3/src/lib.rs
+++ b/chips/apollo3/src/lib.rs
@@ -89,7 +89,7 @@ pub unsafe fn init() {
     scb::disable_fpca();
 
     // This ensures the FPU is actually disabled
-    asm!("svc 0xff", out("r0") _, out("r1") _, out("r2") _, out("r3") _, out("r12") _);
+    // asm!("svc 0xff", out("r0") _, out("r1") _, out("r2") _, out("r3") _, out("r12") _);
 
     cortexm4::nvic::disable_all();
     cortexm4::nvic::clear_all_pending();
```

but still get the same fault as just disabling the `svc`, which is:

```
panicked at 'Kernel HardFault.
        Kernel version release-2.1-rc1-18-g0b88e6ada
        r0  0x10001ab8
        r1  0x10001da0
        r2  0x0
        r3  0x10001d40
        r12 0x0
        lr  0x10001d70
        pc  0x135ae
        prs 0x1000000 [ N 0 Z 0 C 0 V 0 Q 0 GE 0000 ; ICI.IT 0 T true ; Exc 0-Thread Mode ]
        sp  0x10000a18
        top of stack     0x40008000
        bottom of stack  0x10000000
        SHCSR 0x0
        CFSR  0x8200
        HSFR  0x40000000
        Instruction Access Violation:       false
        Data Access Violation:              false
        Memory Management Unstacking Fault: false
        Memory Management Stacking Fault:   false
        Memory Management Lazy FP Fault:    false
        Instruction Bus Error:              false
        Precise Data Bus Error:             true
        Imprecise Data Bus Error:           false
        Bus Unstacking Fault:               false
        Bus Stacking Fault:                 false
        Bus Lazy FP Fault:                  false
        Undefined Instruction Usage Fault:  false
        Invalid State Usage Fault:          false
        Invalid PC Load Usage Fault:        false
        No Coprocessor Usage Fault:         false
        Unaligned Access Usage Fault:       false
        Divide By Zero:                     false
        Bus Fault on Vector Table Read:     false
        Faulting Memory Address: (valid: false) 0x00000000
        Bus Fault Address:       (valid: true) 0x00000000
', arch/cortex-m/src/lib.rs:503:5
        Kernel version release-2.1-rc1-18-g0b88e6ada
```

This is the snipppet where the PC is pointing to

```
   1358c:       e9cd 46d1       strd    r4, r6, [sp, #836]      ; 0x344
   13590:       e9cd 54cf       strd    r5, r4, [sp, #828]      ; 0x33c
   13594:       90ab            str     r0, [sp, #684]  ; 0x2ac
   13596:       a8ce            add     r0, sp, #824    ; 0x338
   13598:       f005 fb74       bl      18c84 <_ZN6kernel5debug13debug_println17h7e9357a72a5b80f7E>
   1359c:       f8dd c0b0       ldr.w   ip, [sp, #176]  ; 0xb0
   135a0:       f8df e34c       ldr.w   lr, [pc, #844]  ; 138f0 <_ZN21redboard_artemis_nano5setup17hc595872dabbbd41cE+0x1ee8>
   135a4:       9a2b            ldr     r2, [sp, #172]  ; 0xac
   135a6:       f10e 0130       add.w   r1, lr, #48     ; 0x30
   135aa:       48c2            ldr     r0, [pc, #776]  ; (138b4 <_ZN21redboard_artemis_nano5setup17hc595872dabbbd41cE+0x1eac>)
   135ac:       3044            adds    r0, #68 ; 0x44
   135ae:       e9c2 c100       strd    ip, r1, [r2]
   135b2:       e9c2 0902       strd    r0, r9, [r2, #8]
   135b6:       f20d 5d44       addw    sp, sp, #1348   ; 0x544
   135ba:       e8bd 0f00       ldmia.w sp!, {r8, r9, sl, fp}
   135be:       bdf0            pop     {r4, r5, r6, r7, pc}
   135c0:       2001            movs    r0, #1
   135c2:       e7b8            b.n     13536 <_ZN21redboard_artemis_nano5setup17hc595872dabbbd41cE+0x1b2e>
   135c4:       2001            movs    r0, #1
```

ppannuto:
Exception entry/exit is definitely more complicated with the hard FPU enabled. Short version, the calling convention nominally requires the hardware to bank (push to stack) many more registers for you; that's quite slow, so there's a bunch of configuration options where you can tell the core to bank fewer of them for better performance at the cost of having to manually avoid some register clobbering. There's a pointer to the relevant bit in the ARM ARM [in my comment on the stacked exception bug](https://github.com/tock/tock/issues/3109#issuecomment-1207117500).

This complexity is also why FPU / m4f stuff keeps getting deferred: https://github.com/tock/tock/pull/1010 , https://github.com/tock/tock/pull/2348 – though clearly it's something we're going to need to figure out at some point.

hudson-ayers:
> That should cause Tock to execute svc_handler_arm_v7m(). At which point the link register should be something other then 0xfffffff9. So we should set SYSCALL_FIRED (which would be incorrect), but then continue running.

Why would the link register not be 0xfffffff9? LR should get set to 0xfffffff9 anytime svc is called and "The core was not currently executing an exception handler and the core was in “Main” (aka kernel) context" (per Pat, but that matches my reading of the ARM ARM). This would result in an attempt to restore state from the currently-uninitialized process stack, which sounds like it could lead to all sorts of weird failures depending what is currently in that memory. However, I do not know why whether the svc call was inlined would have any effect here, so maybe I am missing something.

An interesting test fix might be to try passing a different immediate value to your svc call (say, 0xcc), and modifying the svc handler to check for that `imm` value and just return immediately when it is detected. Here is my draft of what that code might look like (you would also have to change 

```
diff --git a/arch/cortex-m/src/lib.rs b/arch/cortex-m/src/lib.rs
index 4dd1034e5..bf5b0604d 100644
--- a/arch/cortex-m/src/lib.rs
+++ b/arch/cortex-m/src/lib.rs
@@ -159,6 +159,17 @@ pub unsafe extern "C" fn svc_handler_arm_v7m() {
     use core::arch::asm;
     asm!(
         "
+    // Extract the SVC number
+    tst        lr, #4
+    ite eq
+    mrseq r0, msp
+    mrsne r0, psp
+    ldr        r0, [r0, #24]
+    ldrb r0, [r0, #-2]
+    // SVC number in R0 now.
+    cmp r0, #0xcc
+    beq 300f // do_nothing
+
     // First check to see which direction we are going in. If the link register
     // is something other than 0xfffffff9, then we are coming from an app which
     // has called a syscall.
@@ -188,6 +199,7 @@ pub unsafe extern "C" fn svc_handler_arm_v7m() {
     mov r1, #1
     str r1, [r0, #0]

+  300: // do_nothing
     // Set thread mode to privileged as we switch back to the kernel.
     mov r0, #0
     msr CONTROL, r0
diff --git a/chips/apollo3/src/lib.rs b/chips/apollo3/src/lib.rs
index dc3344493..d9646c933 100644
--- a/chips/apollo3/src/lib.rs
+++ b/chips/apollo3/src/lib.rs
@@ -88,7 +88,7 @@ pub unsafe fn init() {
     scb::disable_fpca();

     // This ensures the FPU is actually disabled
-    asm!("svc 0xff", out("r0") _, out("r1") _, out("r2") _, out("r3") _, out("r12") _);
+    asm!("svc 0xcc", out("r0") _, out("r1") _, out("r2") _, out("r3") _, out("r12") _);

     cortexm4::nvic::disable_all();
     cortexm4::nvic::clear_all_pending();
```

alistair23:
> Exception entry/exit is definitely more complicated with the hard FPU enabled

Ok, so I'm going to guess that this is very related to the stack corruption then.

> Why would the link register not be 0xfffffff9?

Whoops. I just assumed we had to set it to that, I didn't realise it was some special hardware value.

> This would result in an attempt to restore state from the currently-uninitialized process stack, which sounds like it could lead to all sorts of weird failures depending what is currently in that memory. However, I do not know why whether the svc call was inlined would have any effect here, so maybe I am missing something.

I'm guessing that's what is happening then. But like you I don't understand why inlining makes any difference

> An interesting test fix might be to try passing a different immediate value to your svc call (say, 0xcc), and modifying the svc handler to check for that `imm` value and just return immediately when it is detected. Here is my draft of what that code might look like (you would also have to change

I tried your diff and I don't see a difference. I'm still seeing the same hardfault and the fix is still the same (this PR)

alistair23:
A little bit more progress.

I managed to get my debugger to connect early enough and give me a little bit of insight.

Without @hudson-ayers's change (so just the current master) the `svc_handler_arm_v7m()` takes the `to_kernel` route. So we incorrectly set `SYSCALL_FIRED` but otherwise Hudson's change from above won't make much difference to fixing the hardfault.

It looks like with `apollo3::init()` not inline we get a hardfault at the first function return.

This is the last instruction my debugger gets to, before it crashes and disconnects:

```
0x00011ad2  apollo3::iom::Iom::new0+4 bl    0x1a3c0 <compiler_builtins::arm::__aeabi_memclr4::__aeabi_memclr4>
```

So I'm guessing something about the FPU status changing is breaking the entry/exit expectations and that is causing the stack to become corrupted

hudson-ayers:
@alistair23 Do you have a pointer to any documentation on why the `svc` call is needed to complete disabling the FPU? Or is that just something we sort of stumbled across as working? I am wondering if the `svc` call only happens to be disabling the FPU as some sort of side effect, but there is a different approach we are supposed to take

phil-levis:
Here's a guess on what's happening. This is from 2-26/2-27 in [Cortex -M4 Devices Generic User Guide](https://developer.arm.com/documentation/dui0553/b).

When the boot sequence triggers `svc`, floating point is enabled. Following Figure 2-3 in the guide, this means that in addition to the standard 8 registers (xPsr, PC, LR, R12, R3, R2, R1, R0), an additional 17 floating point registers (FPSCR, S15...S0) are pushed.

On exception return, whether floating point registers are unstacked depends on the `EXC_RETURN` code passed in `lr` (Table 2-17). If you pass the normal `0xFFFFFFFF9`, the registers are not unstacked. This means that when you resume execution, the stack is larger than before and it's filled with floating point state. This would be an easy way for execution to therefore think that `lr=0`. To cause exception return to unstack the floating point registers, the handler needs to put `0xFFFFFFE9` into `pc`.

alistair23:
> @alistair23 Do you have a pointer to any documentation on why the `svc` call is needed to complete disabling the FPU? Or is that just something we sort of stumbled across as working? I am wondering if the `svc` call only happens to be disabling the FPU as some sort of side effect, but there is a different approach we are supposed to take

I don't have any documentation as to why it's required. I think I just stumbled across it. Without the `svc` we end up crashing later, I'm guessing that the next `svc` causes a fault due to the stack unwinding

hudson-ayers:
> Here's a guess on what's happening. This is from 2-26/2-27 in [Cortex -M4 Devices Generic User Guide](https://developer.arm.com/documentation/dui0553/b).
> 
> When the boot sequence triggers `svc`, floating point is enabled. Following Figure 2-3 in the guide, this means that in addition to the standard 8 registers (xPsr, PC, LR, R12, R3, R2, R1, R0), 17 floating point registers (FPSCR, S15...S0) are pushed.
> 
> On exception return, whether floating point registers are unstacked depends on the `EXC_RETURN` code passed into `pc` (Table 2-17). If you pass the normal `0xFFFFFFFF9`, the registers are not unstacked. This means that when you resume execution, the stack is larger than before and it's filled with floating point state. This would be an easy way for execution to therefore think that `lr=0`. To cause exception return to unstack the floating point registers, the handler needs to put `0xFFFFFFE9` into `pc`.

So the theory for why we only see the failure if the `svc` call is issued in a non-inlined function is: the portion of the stack that gets corrupted by this is only touched if we later pop it off the stack, and if `svc` is called in main, then we never pop off that portion of the stack again because main never returns.

What is still very strange to me is that this suggests that the apollo3 chip behavior is:
1) which registers are pushed during an exception is automatically determined in the hardware based on if the FPU is enabled. 2) For the purposes of this register pushing, whether the FPU is enabled is not determined via the standard mechanism of reading the `CPACR`, but some other mechanism that actually detects if the FPU has been fully disabled via an `svc` call.
3) disabling the FPU requires an `svc`, but this is not documented anywhere, and from what I can tell is not a requirement for most Cortex m4f chips

ppannuto:
This was discussed on the core call today. The consensus is that we understand why this patch works, namely that it works around bugs in the underlying context switch code for cortex-m in the kernel. This is not desired as a permanent fix, but should be applied for 2.1 release such that this board will be usable.

The proposal was to:
  1. Add explicit warnings to the source about pending removal.
  2. Merge this for 2.1.
  3. Immediately open a second, companion PR that reverts this change, to be merged after the 2.1 release.

I believe this PR should now be ready to merge.

bradjc (approved):
Approved for 2.1 release. I updated title to reflect temporary workaround.

brghena (approved):

PR #3140, opened 2022-08-12
boards/redboard_artemis_nano: Fixup app loading

alistair23:
### Pull Request Overview

Ensure the app doesn't override Tock

The ROM seems to store some temporary data at the load-address-blob
address. In commit f8ceb22352da we increased the addresses, but didn't
correctly increase the load-address-blob of the application enough. This
means that the temporary data can override the Tock binary. Increase the
address to a larger one that works.

Fixes: f8ceb22352da "boards/redboard_artemis_nano: Increase the blob loading address"

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/boards/redboard_artemis_nano/Makefile
+++ b/boards/redboard_artemis_nano/Makefile
@@ -23,7 +23,7 @@ flash-debug: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/debug/$(PLATFORM).bin
 
 .PHONY: flash-app
 flash-app:
-	python ambiq/ambiq_bin2board.py --bin $(APP) --load-address-blob 0x40000 -b 115200 -port $(PORT) -r 2 -v --magic-num 0xCB --version 0x0 --load-address-wired 0x40000 -i 6 --options 0x1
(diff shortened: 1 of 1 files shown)


alistair23:
The failure is a Mac CI failure, not related to this patch

brghena:
If you change that address to 0x6000, how does Tock know to look for the applications at that address? Don't you have to change the LD file too? https://github.com/tock/tock/blob/master/boards/redboard_artemis_nano/layout.ld#L4

phil-levis (dismissed):

alistair23:
In this case I'm not changing the final address the application is loaded. I'm just changing the address the ROM stores temporary data

phil-levis:
I'm trying to reconcile these two sentences:

"Ensure the app doesn't override Tock"

and 

"I'm just changing the address the ROM stores temporary data"

Which is true? Please document the change in more detail (e.g., what this flag does, how the change fixes the issue) in the main comment. Generally speaking, I would like it if you documented your PRs more, so that other people can understand what they are doing; that way, someone else can backtrack and understand what you've done.

alistair23:
Both are true. I have updated the description with this

```
The ROM seems to store some temporary data at the load-address-blob
address. In commit f8ceb22352da we increased the addresses, but didn't
correctly increase the load-address-blob of the application enough. This
means that the temporary data can override the Tock binary. Increase the
address to a larger one that works.
```

phil-levis (approved):

PR #3147, opened 2022-08-16
arch/cortex-m: Correct end stack printing

alistair23:
### Pull Request Overview

Before this change I would see these values:

    top of stack     0x40008000

After this change I now see these values:

    top of stack     0x10001000

This patch converts the end stack pointing to match the start stack
address pointing. Now the printed values match the objdump values.

### Testing Strategy

Running on Artemis Nano

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/lib.rs
+++ b/arch/cortex-m/src/lib.rs
@@ -556,7 +556,7 @@ unsafe fn kernel_hardfault_arm_v7m(faulting_stack: *mut u32) -> ! {
         exception_number,
         ipsr_isr_number_to_str(exception_number),
         faulting_stack as u32,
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):
Definitely doesn't make sense this was being printed any differently than the start of the stack

ppannuto (approved):
bors r+

This is strictly an improvement on what's there now, and is a pretty clean and clear bugfix to me. I think there are some deeper issues (still..) around our linker-variable <> Rust integration I'd like to dig into, but that can come later.

ppannuto:
bors retry

PR #3148, opened 2022-08-17
boards/esp32-c3-devkitM-1: Prepare for the 2.1 release

alistair23:
### Pull Request Overview

Remove invalid `dont-append-digest` option when loading an app and add support to the board test runner for the ESP32-C3.

### Testing Strategy

board-release-tests

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/tools/board-runner/src/esp32_c3.rs
+++ b/tools/board-runner/src/esp32_c3.rs
@@ -1,0 +1,328 @@
+use rexpect::errors::Error;
+use rexpect::spawn_stream;
+use serialport::prelude::*;
+use serialport::SerialPortSettings;
+use std::env;
+use std::fs::OpenOptions;
+use std::process::{Command, Stdio};
+use std::time::Duration;
+use std::{thread, time};
+
+fn esp32_c3_flash(
+    app_name: &str,
+) -> Result<rexpect::session::StreamSession<std::boxed::Box<dyn serialport::SerialPort>>, Error> {
+    let s = SerialPortSettings {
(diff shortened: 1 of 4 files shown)


tools/board-runner/src/esp32_c3.rs
```diff
@@ -0,0 +1,328 @@
...
+        "{}/{}",
+        env::var("LIBTOCK_C_TREE").unwrap(),
+        "examples/tests/mpu_walk_region/build/rv32imac/rv32imac.0x403B0060.0x3FCC0000.tbf"
+    );
+    let mut p = esp32_c3_flash(&app).unwrap();
+
+    p.exp_string("MPU Walk Regions")?;
+    p.exp_string("Walking flash")?;
+    // p.exp_string("Will overrun")?;
+    // p.exp_string("panicked at 'Process mpu_walk_region had a fault'")?;
```
hudson-ayers, line 276:
```suggestion
```
No need for these since the board does not have a button attached, right?

hudson-ayers (approved):

tools/board-runner/src/esp32_c3.rs
alistair23: The board has a physical button. it's not attached in Tock though

hudson-ayers: gotcha, makes sens to keep them then

bradjc (approved):

PR #3149, opened 2022-08-18
boards/litex: update pinned tock-litex release to 2022081701

lschuermann:
### Pull Request Overview

This updates the LiteX boards (sim & arty) to new hardware revisions, provided as part of the `tock-litex` companion repository. This includes pre-built bitstreams. It further updates the `litex-sim-ci` workflow to use a new `tock-litex` revision, based on the `2022081701` release, with patches for the CI workflow applied.

These changes reflect some restructuring in the way that the LiteX SoCs are built, which will contribute to making them easier to run on local systems and ultimately reproduce the `litex-sim-ci` Tock CI workflow. In particular, it reduces the number of dependencies on repository-forks carrying patches and provides more pre-built patched packages as part of the `tock-litex` releases.

Furthermore, this adds the process console component and IPC support to the LiteX Arty board, to make it a more useful general-purpose board and more applicable to the release test suite.

I have been running many of the release tests successfully with these changes and would appreciate if we can get this in prior to Tock 2.1.

### Testing Strategy

This pull request was tested by running CI & many release tests.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/.github/workflows/litex_sim.yml
+++ b/.github/workflows/litex_sim.yml
@@ -76,13 +76,10 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: lschuermann/tock-litex
-          # Currently pointing to a fixed commit, switch to a tag on the next
-          # release with the Python requirements.txt included.
-          #
-          # Also, currently using the dev/litex-sim-ci branch, which
-          # pulls in a forked LiteX with GPIO support in the
-          # simulation and the simctrl modules
-          ref: da3f33968ae018c33dd10d705fd7e3ca04e5e056
+          # The pinned revision is different from the targeted release as
+          # documented in the LiteX boards, as the CI requires special patches
+          # to LiteX for interacting with the simulation:
+          ref: 2022081701-tock-ci-0
           path: tock-litex
 
       # Install all of the required Python packages from the tock-litex'
@@ -90,10 +87,13 @@ jobs:
       - name: Install Python packages pinned by the tock-litex revision
         run: |
           pushd tock-litex
-          # Unfortunately the valentyusb package fails to install because
-          # it does not have a setup.py. Thus ignore it, we don't need it
-          # here anyways (would need it for LiteX unit tests though).
-          sed -i '/valentyusb/d' ./requirements.txt
+          # Migen is the DSL which the LiteX ecosystem uses as its
+          # hardware-description language. It effectively provides a set of
+          # Python classes and constructs which can be translated into Verilog.
+          # It is not a package of the LiteX ecosystem, and thus not in the
+          # requirements.txt, but it is required to be present on the system.
+          # It should not require any specific or patched version.
+          pip3 install migen==0.9.2
           pip3 install -r requirements.txt
           popd
 
--- a/boards/litex/arty/README.md
+++ b/boards/litex/arty/README.md
@@ -11,9 +11,9 @@ differ significantly depending on the LiteX release and configuration
 options used. This board definition currently targets and has been
 tested with
 - [the LiteX SoC generator, revision
-  c43132f81f1113](https://github.com/enjoy-digital/litex/tree/c43132f81f1113)
+  c4e635ea5c91ca](https://github.com/enjoy-digital/litex/tree/c4e635ea5c91ca)
 - using the companion [target
-  file](https://github.com/litex-hub/litex-boards/blob/f18b10d1edb4e1/litex_boards/targets/digilent_arty.py)
+  file](https://github.com/litex-hub/litex-boards/blob/bf458e388efea4/litex_boards/targets/digilent_arty.py)
   from `litex-boards`
 - built around a VexRiscv-CPU with PMP, hardware multiplication and
   compressed instruction support (named `TockSecureIMC`)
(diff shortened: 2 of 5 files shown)


.github/workflows/litex_sim.yml
```diff
@@ -76,24 +76,20 @@ jobs:
...
-          # release with the Python requirements.txt included.
-          #
-          # Also, currently using the dev/litex-sim-ci branch, which
-          # pulls in a forked LiteX with GPIO support in the
-          # simulation and the simctrl modules
-          ref: da3f33968ae018c33dd10d705fd7e3ca04e5e056
+          # The pinned revision is different from the targeted release as
+          # documented in the LiteX boards, as the CI requires special patches
+          # to LiteX for interacting with the simulation:
+          ref: tock-cipin-1
```
ppannuto, line 82: Assuming this ref is a name we control, might be nicer to name it based on a date? `tock-CI-pin-2022-08-18`? I assume these numbers will keep growing arbitrarily, so a little semantic context here might be nice.

.github/workflows/litex_sim.yml
```diff
@@ -76,24 +76,20 @@ jobs:
...
       # requirements.txt file
       - name: Install Python packages pinned by the tock-litex revision
         run: |
           pushd tock-litex
-          # Unfortunately the valentyusb package fails to install because
-          # it does not have a setup.py. Thus ignore it, we don't need it
-          # here anyways (would need it for LiteX unit tests though).
-          sed -i '/valentyusb/d' ./requirements.txt
+          # Migen is technically not a LiteX package, which is why it is not
+          # contained within the requirements.txt file:
```
ppannuto, line 91: what is it then / why do we need it (when litex doesn't?)

boards/litex/arty/README.md
```diff
@@ -35,13 +35,13 @@ hardware multiplication and compressed instruction support (such that
...
-Verilog files) can be obtained from the [Tock on LiteX companion
-repository
+Verilog files and a patched LiteX version to support them) can be
+obtained from the [Tock on LiteX companion repository
 releases](https://github.com/lschuermann/tock-litex/releases/). The
 current board definition has been verified to work with [release
-2021100501](https://github.com/lschuermann/tock-litex/releases/tag/2021100501). The
-bitstream for this board is located in `digilent_arty_a7-35t.zip` or
-`digilent_arty_a7-100t.zip` under `gateware/digilent_arty.bit`.
+2022081701](https://github.com/lschuermann/tock-litex/releases/tag/2022081701).
```
ppannuto, line 42: Ah! should the CI tag then be a derivative of this tag? `tock-CI-from-2022081701` or s/t like that?

boards/litex/arty/README.md
```diff
@@ -120,8 +120,8 @@ To boot via serial run the LiteX-included `litex_term.py` (sometimes
 available as `lxterm`):
 ```
 $ ./litex/litex/tools/litex_term.py \
-    --speed 10000000 \
-	--serial-boot \
+    --speed 1000000 \
```
ppannuto, line 123: going from 10MBaud to 1MBaud is a pretty serious slowdown – any risk of buffer overruns here? (said, having had a miserable experience tracking down a transient bug when moving a different project from 3MBaud to 1MBaud to run more reliably on USB stacks within VMs...)

ppannuto (changes requested):
looks good for pre-2.1 inclusion to me

A few small thoughts, but most doc/cosmetic stuff

hudson-ayers (approved):
This looks good to me to go in as part of 2.1, since it gets the testing working for the litex platform and the changes are confined to that platform

boards/litex/arty/README.md
lschuermann: Yes, that seems like a great idea! I'll rename the ref.

.github/workflows/litex_sim.yml
lschuermann: Migen is the DSL which the LiteX ecosystem uses as its hardware-description language. It effectively provides a set of Python classes and constructs which can be translated into Verilog. So it's not a package of the LiteX ecosystem, but it is a hard requirement. I don't include it in the auto-generated requirements.txt of `tock-litex` given it's a regular system dependency and we don't hard-require a specific or patched version of it.

boards/litex/arty/README.md
lschuermann: I should've specified that more. In this case, the documentation is wrong and has been for a long time. 10MBaud would be nice but I have deemed that to be unreliable on the Arty. The bitstreams have always been built with a 1MBaud UART (by default). This was actually reported to me some time ago by @irina-b-dev.

alistair23 (approved):

boards/litex/arty/README.md
lschuermann: Renamed the tag: `2022081701-tock-ci-0`, which is a lot more meaningfull hopefully.

.github/workflows/litex_sim.yml
ppannuto:
```suggestion
          # Migen is the DSL which the LiteX ecosystem uses as its
          # hardware-description language. It effectively provides a set of
          # Python classes and constructs which can be translated into Verilog.
          # It is not a package of the LiteX ecosystem, and thus not in the
          # requirements.txt, but it is required to be present on the system.
          # It should not requie any specific or patched version.
```

lschuermann: Fair enough :)

hudson-ayers (approved):

alistair23 (approved):

bradjc (approved):

ppannuto (approved):

PR #3150, opened 2022-08-18
cortex-m: cleanup linker constants

ppannuto:
### Pull Request Overview

This is the follow-on to #3147 that does the real fix here. I still think #3147 should go in for 2.1 as a quick-fix, and then this can come in after as the better fix.

This fixes the type imports from the linker script to match reality (they're const pointers). I think this was just some real legacy stuff that predated our understanding of Rust.

### Testing Strategy

Todo: will test on HW when I'm in the office next week.

### TODO or Help Wanted

^see above

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/lib.rs
+++ b/arch/cortex-m/src/lib.rs
@@ -15,6 +15,17 @@ pub mod support;
 pub mod syscall;
 pub mod systick;
 
+// These constants are defined in the linker script.
+extern "C" {
+    static _estack: *const u32;
+    static _sstack: *const u32;
+    static _szero: *const u32;
+    static _ezero: *const u32;
+    static _etext: *const u32;
+    static _srelocate: *const u32;
+    static _erelocate: *const u32;
+}
+
 /// Trait to encapsulate differences in between Cortex-M variants
 ///
 /// This trait contains functions and other associated data (constants) which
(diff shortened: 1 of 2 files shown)


hudson-ayers (changes requested):
I think we should also apply these changes to here: https://github.com/tock/tock/blob/915863ef09fd3b0c933484362b2fdcf44111ca22/boards/opentitan/src/main.rs#L661 and here https://github.com/tock/tock/blob/master/arch/rv32i/src/lib.rs#L23 and here https://github.com/tock/tock/blob/master/boards/components/src/process_console.rs#L73 ?

hudson-ayers:
> I think we should also apply these changes to here:
> 
> https://github.com/tock/tock/blob/915863ef09fd3b0c933484362b2fdcf44111ca22/boards/opentitan/src/main.rs#L661
> 
> and here https://github.com/tock/tock/blob/master/arch/rv32i/src/lib.rs#L23 and here https://github.com/tock/tock/blob/master/boards/components/src/process_console.rs#L73 ?

I think this comment still applies

ppannuto:
Yep, I false-started on cleaning that up and finishing this the other day; will roll this over from draft to ready when it's actually ready.

bradjc:
Can we merge this?

brghena (approved):

alistair23 (approved):

hudson-ayers (approved):
Just realized this was held up by my requested changes, sorry about that

PR #3154, opened 2022-08-23
NRF serialization: Handle restarted app in 2.0 world

bradjc:
### Pull Request Overview

I think somewhere in the 2.0 update the NRF serialization capsule no longer worked if an app faulted and restarted. This updates the capsule to support restarting an app using the nrf serialization.

The issue is with receive: if a second app (ie the new processid after the original app faulted and was restarted) calls receive, the capsule would see there is already an active app (the original app), and the kernel would return an error and everything would break at that point.

Now, we on a receive command we check: is there an active app? If so, does that app still exist? If so -> error. Otherwise, mark this new app as the active app.

### Testing Strategy

Using the process console to fault the hail app.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/nrf51822_serialization.rs
+++ b/capsules/src/nrf51822_serialization.rs
@@ -213,7 +213,7 @@ impl SyscallDriver for Nrf51822Serialization<'_> {
                     },
                     |appid| {
                         // The app is set, check if it still exists.
-                        if let Err(_err) = self.apps.enter(*appid, |_, _| {}) {
+                        if let Err(kernel::process::Error::NoSuchApp) = self.apps.enter(*appid, |_, _| {}) {
                             // The app we had as active no longer exists.
                             self.active_app.clear();
                             self.rx_buffer
--- a/capsules/src/nrf51822_serialization.rs
+++ b/capsules/src/nrf51822_serialization.rs
@@ -182,17 +182,58 @@ impl SyscallDriver for Nrf51822Serialization<'_> {
             }
             // Receive from the nRF51822
             2 => {
-                self.rx_buffer.take().map_or(CommandReturn::failure(ErrorCode::RESERVE), |buffer| {
-                    let len = arg1;
-                    if len > buffer.len() {
-                        CommandReturn::failure(ErrorCode::SIZE)
-                    } else {
-                        // Set this as the active app for the receive callback
-                        self.active_app.set(appid);
-                        let _ = self.uart.receive_automatic(buffer, len, 250);
-                        CommandReturn::success_u32(len as u32)
-                    }
-                })
+                let len = arg1;
+
+                // We only allow one app to use the NRF serialization capsule
+                // (old legacy code, and a difficult thing to virtualize).
+                // However, we would like to support restarting/updating apps.
+                // But we don't want to allow a simultaneous app to disrupt the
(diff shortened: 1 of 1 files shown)


capsules/src/nrf51822_serialization.rs
```diff
@@ -182,17 +182,58 @@ impl SyscallDriver for Nrf51822Serialization<'_> {
...
+                                }
+                            })
+                    },
+                    |appid| {
+                        // The app is set, check if it still exists.
+                        if let Err(_err) = self.apps.enter(*appid, |_, _| {}) {
+                            // The app we had as active no longer exists.
+                            self.active_app.set(*appid);
+                            self.rx_buffer
+                                .take()
```
brghena, line 220: This arm of the code sets the `active_app` before checking size of the `rx_buffer`. The above wing instead doesn't set `active_app` if the length is too long. Is this different problematic?

capsules/src/nrf51822_serialization.rs
```diff
@@ -270,14 +315,22 @@ impl uart::ReceiveClient for Nrf51822Serialization<'_> {
                 // hardware, regardless of how much space (if any) was
                 // available in the buffer provided by the app.
                 kernel_data.schedule_upcall(0, (4, rx_len, len)).ok();
-            });
+            }) {
```
brghena, line 318: Something is weird with this line. As I'm reading it, the `})` ends the closure for `self.apps.enter()`. But then it's followed by an opening `{`?? I expected there to be an `} else {` like the above code has.

Am I just misunderstanding something?

brghena (changes requested):

capsules/src/nrf51822_serialization.rs
bradjc: Yes.

capsules/src/nrf51822_serialization.rs
bradjc: The open `{` is the bracket for the `if` and its enclosed block is if the grant entry had an error.

brghena: Oh geez. Okay, that parses now.

brghena (dismissed):

capsules/src/nrf51822_serialization.rs
```diff
@@ -182,17 +182,66 @@ impl SyscallDriver for Nrf51822Serialization<'_> {
...
+                                    // receive callback.
+                                    self.active_app.set(appid);
+                                    let _ = self.uart.receive_automatic(buffer, len, 250);
+                                    CommandReturn::success_u32(len as u32)
+                                }
+                            })
+                    },
+                    |appid| {
+                        // The app is set, check if it still exists.
+                        if let Err(_err) = self.apps.enter(*appid, |_, _| {}) {
```
hudson-ayers, line 216:
```suggestion
                        if let Err(kernel::process::Error::NoSuchApp) = self.apps.enter(*appid, |_, _| {}) {
```

To protect against future changes to grant where entering the grant might fail for other reasons than `NoSuchApp` / `InactiveApp`

hudson-ayers (dismissed):

brghena (approved):

hudson-ayers (approved):
bors r+

PR #3157, opened 2022-08-24
TRD: block_storage

dcz-self:
### Pull Request Overview

This pull request adds the TRD corresponding to the [block_storage HIL](https://github.com/tock/tock/pull/2993)

### Testing Strategy

N/A

### TODO or Help Wanted

This pull request is based on another one (https://github.com/tock/tock/pull/2993) and needs that one to be merged first.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/reference/trd-block-storage.md
+++ b/doc/reference/trd-block-storage.md
@@ -1,0 +1,103 @@
+Kernel block storage HIL
+========================================
+
+**TRD:** <br/>
+**Working Group:** Kernel<br/>
+**Type:** Documentary<br/>
+**Status:** Draft <br/>
+**Author:** Dorota Czaplejewicz<br/>
+**Draft-Created:** 2021/08/24 <br/>
+**Draft-Modified:** 2021/08/24 <br/>
+**Draft-Version:** 1 <br/>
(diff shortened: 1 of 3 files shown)


phil-levis:
This is a great start. I think the document needs to be clear that this is about page/block access to raw flash and why; this means it is not "block storage", which in computing systems means in-place modification (e.g., disks, FTL). 

It should document the APIs and their semantics. E.g. take a look at

dcz-self:
Oops, the TRD was called "block_device", where the HIL "block_storage". I fixed that, and clarified which operations correspond to which calls.

TBH I don't see much point in copying the API spec verbosely like the time TRD does. It doesn't seem any different than the documentation that is already present in the HIL, so it would only result in duplication and problems synchronizing the separate documents describing the same thing. What I put in the TRD is the background and reasoning, as well as the basic information about model that is crucial but might be difficult to put together from documentation that's focused on the implementation and is spread across different parts of the HIL. Adding that back to the TRD would negate the benefit of terseness in addition to increasing duplication.

alevy:
Closing without prejudice.

PR #3158, opened 2022-08-24
RFC: remove kernel work counting optimization

### LINKED ISSUE #3138: Potential deadlocks if `kernel.process_blocked` is out of sync
Many of the schedulers use the `kernel.processes_blocked()` function to determine if the scheduler should even look at the processes to schedule them. If `processes_block` is `false`, then the schedulers assume that at least one process will have work to perform. If the assumption is wrong, then the kernel may `panic` or dead lock.

The kernel provides work increment and decrement methods that processes use when they have pending work. Processes also have a `ready()` method that lets the kernel know if the process is ready to schedule or not. Each process must carefully increment and decrement work on the shared work counter for the `kernel.processes_blocked()` to be correct. As of now, the current code does this correctly.

This optimization does not seem worth the risk of getting it wrong. We can simplify assumptions and remove the increment/decrement work concepts along with the `processes_blocked` method. This will only affect the case when there are no processes to run, and the schedulers will loop through all processes once to see if there is any work. This is when the chip is about to go to sleep anyway, so this isn't really time the core needs to optimize for.

This is related to #3137 but taking a step back.

Thoughts?

hudson-ayers:
### Pull Request Overview

This pull request removes the kernel work counter which tracked how many outstanding tasks were present in the kernel. Now, schedulers must determine whether any work is outstanding by iterating through all processes and calling `process.ready()`. For all of our currently implemented schedulers, we would ultimately iterate through processes until finding a ready one anyway, so this is not a significant performance hit. The only scenario in which this increases the amount of work done is when no processes are ready and the board is about to go to sleep -- in this scenario, it used to be the case that the scheduler could sleep immediately, but now the scheduler has to iterate through all processes before sleeping.

This PR reduces the size of the kernel by 176 bytes and simplifies the scheduling logic by not having kernel work tracked in two places (the work counter + the process state/process task lists). It also reduces the amount of work done for things like scheduling callbacks / removing callbacks since the work counter does not need to be incremented and decremented.

As a basic attempt to measure the performance impact of this change, I used the cortex-m DWT to measure the number of cycles between the first time the LED turns on and then back off when running the blink app, and found that with this change there were actually 1200 fewer total cycles (though this does seem to be a noisy measurement).

This PR fixes #3138 and obsoletes #3137 (cc @jettr )

### Testing Strategy

This pull request was tested by running a few combos of apps, such as blink and blink + whileone, and taking cycle measurements using the DWT.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/scheduler/cooperative.rs
+++ b/kernel/src/scheduler/cooperative.rs
@@ -12,7 +12,7 @@
 //! that was executing.
 
 use crate::collections::list::{List, ListLink, ListNode};
-use crate::kernel::{Kernel, StoppedExecutingReason};
+use crate::kernel::StoppedExecutingReason;
 use crate::platform::chip::Chip;
 use crate::process::Process;
 use crate::scheduler::{Scheduler, SchedulingDecision};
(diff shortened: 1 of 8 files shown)


jettr (approved):
Woot! This is a nice simplification and makes it is harder to have incorrect states!

bradjc:
> As a basic attempt to measure the performance impact of this change, I used the cortex-m DWT to measure the number of cycles between the first time the LED turns on and then back off when running the blink app, and found that with this change there were actually 1200 fewer total cycles (though this does seem to be a noisy measurement).

You should try with like 10 blink apps (or null apps) installed.

bradjc (dismissed):
I agree this is probably the right tradeoff to make. Until the language itself can provide support for this type of optimization, it's hard to always get it right.

brghena (approved):
Looks good to me

ppannuto (approved):
Hah, should've read this before commenting on #3158.. the 1200 cycle reduction is a bit surprising, but maybe we had more overhead in state tracking that I realized.

The other thing I like which I'd meant to comment on there is that I appreciate the scheduler no longer taking a reference to the kernel. I think that makes for a cleaner pluggable scheduler interface.

kernel/src/scheduler/cooperative.rs
```diff
@@ -52,33 +52,39 @@ impl<'a> CooperativeSched<'a> {
...
-    fn next(&self, kernel: &Kernel) -> SchedulingDecision {
-        if kernel.processes_blocked() {
-            // No processes ready
-            SchedulingDecision::TrySleep
-        } else {
-            let mut next = None; // This will be replaced, bc a process is guaranteed
-                                 // to be ready if processes_blocked() is false
+    fn next(&self) -> SchedulingDecision {
+        let mut first_head = None;
+        let mut next = None; // This will be replaced, bc a process is guaranteed
```
ppannuto, line 57: This comment needs updating

kernel/src/scheduler/round_robin.rs
```diff
@@ -64,41 +64,47 @@ impl<'a> RoundRobinSched<'a> {
...
-    fn next(&self, kernel: &Kernel) -> SchedulingDecision {
-        if kernel.processes_blocked() {
-            // No processes ready
-            SchedulingDecision::TrySleep
-        } else {
-            let mut next = None; // This will be replaced, bc a process is guaranteed
-                                 // to be ready if processes_blocked() is false
+    fn next(&self) -> SchedulingDecision {
+        let mut first_head = None;
+        let mut next = None; // This will be replaced, bc a process is guaranteed
```
ppannuto, line 69: here too

bradjc:
I would like to block this until after appid is merged.

hudson-ayers:
Rebased

hudson-ayers:
ping!

brghena (approved):

bradjc (approved):

PR #3166, opened 2022-08-26
stm: improve app programming

alexandruradovici:
### Pull Request Overview

This pull request changes the makefile and readme files for STM32 boards to improve the app programming.

It uses the same methods as used for the Raspberry Pi Pico board. It defines a command line variable `APP` that contains the path to a TBF.

### Testing Strategy

This pull request was tested using STM32F412g Discovery and Nucleo429zi boards.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nucleo_f429zi/Makefile
+++ b/boards/nucleo_f429zi/Makefile
@@ -8,6 +8,9 @@ include ../Makefile.common
 OPENOCD=openocd
 OPENOCD_OPTIONS=-f openocd.cfg
 
+KERNEL=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/release/$(PLATFORM).elf
+KERNEL_WITH_APP=$(TOCK_ROOT_DIRECTORY)/target/$(TARGET)/release/$(PLATFORM)-app.elf
+
 # Default target for installing the kernel.
 .PHONY: install
 install: flash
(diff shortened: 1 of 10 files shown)


bradjc:
Why don't we use tockloader?

alexandruradovici:
STM32F412g has a strange Flash layout, the kernel breaks when apps are loaded. Let me try with the rest of them.

hudson-ayers (approved):
These changes seem fine if we want to get something in before the release, but `tockloader` support should probably be a priority for these boards.

Could we just lie to tockloader about the page size being 128 kB? That appears to be the largest page size for that chip, depending on where you are in flash. It would make flashing slower than the optimal case but thats not a huge deal

bradjc:
> Could we just lie to tockloader about the page size being 128 kB? That appears to be the largest page size for that chip, depending on where you are in flash. It would make flashing slower than the optimal case but thats not a huge deal

Is the start of apps on some sort of page boundary? If so `--bundle-apps` was intended to address issues like this, as that flag tells tockloader to write all apps as one large write every time. But if the page overlaps with the kernel then I don't know what to do.

hudson-ayers:
> > Could we just lie to tockloader about the page size being 128 kB? That appears to be the largest page size for that chip, depending on where you are in flash. It would make flashing slower than the optimal case but thats not a huge deal
> 
> Is the start of apps on some sort of page boundary? If so `--bundle-apps` was intended to address issues like this, as that flag tells tockloader to write all apps as one large write every time. But if the page overlaps with the kernel then I don't know what to do.

It looks like currently the kernel/app boundary is halfway through a page -- we should just move it later. I'll submit a PR

alevy:
@alexandruradovici keep this around or close it? Reopen if you want

alexandruradovici:
Thanks

PR #3167, opened 2022-08-26
Refresh the process console list command output

bradjc:
### Pull Request Overview

Before:

```
tock$ list
 PID    Name                Quanta  Syscalls  Dropped Upcalls  Restarts    State  Grants
  0	blink                    0        57                0         0  Yielded    1/14
```

After:
```
tock$ list
 PID    Name                Quanta  Syscalls  Restarts  Grants  State
 1      blink                    0        84         1   1/14   Yielded
```

- Implement alignment on `ProcessId` so we don't have to use the tab hack.
- Move state to the end so that its length can run off the end of the line.
- Format grants properly.
- Remove `Dropped Upcalls` because it is always zero and now that we have the process command it can be viewed there.

### Testing Strategy

Using the process console.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/process_console.rs
+++ b/capsules/src/process_console.rs
@@ -24,8 +24,6 @@
 //! - `Quanta`: How many times this process has exceeded its allotted time
 //!   quanta.
 //! - `Syscalls`: The number of system calls the process has made to the kernel.
-//! - `Dropped Upcalls`: How many upcalls were dropped for this process
-//!   because the queue was full.
 //! - `Restarts`: How many times this process has crashed and been restarted by
 //!   the kernel.
 //! - `Grants`: The number of grants that have been initialized for the process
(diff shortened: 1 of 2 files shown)


ppannuto (approved):
This looks reasonable – is this something that will (that should?) conflict with AppId? Seems like that would be useful information here as well. We should consider PR ordering and prioritizing AppId I think?

hudson-ayers (approved):
These are nice changes.

I will note that a lot of the fancy formatting stuff would become impossible if we ever wanted to switch to something substantially smaller for in-kernel debug output like `ufmt`

bradjc:
> is this something that will (that should?) conflict with AppId? Seems like that would be useful information here as well. We should consider PR ordering and prioritizing AppId I think?

I pulled most of the pconsole changes from app id into their own PR. Only a slight change to how processes are restarted, shouldn't be an issue to merge this separately from appid.

alistair23 (approved):

PR #3169, opened 2022-08-26
Update stored kernel version to 2.1

bradjc:
### Pull Request Overview

For the release.

Also make it easier to find by including the word "version" in the variable name.

### Testing Strategy

n/a

### TODO or Help Wanted

I think these are the two places. Anyone remember any others?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/Makefile.common
+++ b/boards/Makefile.common
@@ -149,7 +149,7 @@ endif
 # Ask git what version of the Tock kernel we are compiling, so we can include
 # this within the binary. If Tock is not within a git repo then we fallback to
 # a set string which should be updated with every release.
-export TOCK_KERNEL_VERSION := $(shell git describe --tags --always 2> /dev/null || echo "1.4+")
+export TOCK_KERNEL_VERSION := $(shell git describe --tags --always 2> /dev/null || echo "2.1+")
 
 # Allow users to opt out of using rustup.
 ifeq ($(NO_RUSTUP),)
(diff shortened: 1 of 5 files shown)


boards/Makefile.common
```diff
@@ -149,7 +149,7 @@ endif
 # Ask git what version of the Tock kernel we are compiling, so we can include
 # this within the binary. If Tock is not within a git repo then we fallback to
 # a set string which should be updated with every release.
-export TOCK_KERNEL_VERSION := $(shell git describe --tags --always 2> /dev/null || echo "1.4+")
+export TOCK_KERNEL_VERSION := $(shell git describe --tags --always 2> /dev/null || echo "2.1+")
```
ppannuto, line 152:
```suggestion
export TOCK_KERNEL_VERSION := $(shell git describe --tags --always 2> /dev/null || echo "nogit-version-2.1+")
```

Maybe we make this a little more verbose so we have a better shot of noticing it in the future? It's a bit behind :/. Nothing wrong with more letters here, the normal result is something like `release-2.1-rc1-19-g50c1c14e3`.

ppannuto (approved):

hudson-ayers (approved):

alistair23 (dismissed):

boards/Makefile.common
bradjc: Maybe it should just be `2.1`? That is the version on the release tag, after all. I don't see the need to tell people what tools they were not using.

hudson-ayers (approved):

boards/Makefile.common
ppannuto: That's true.. I think I'd originally named it `1.4+` as a notion of 'this started at 1.4, but `+` who knows what folks have done with it since'. But it's not like there's any way to tell whether they did anything or not. I'd support backing down to just `2.1`.

alistair23 (approved):

PR #3171, opened 2022-08-26
Process Console New(ish) Commands

bradjc:
### Pull Request Overview

This adds the `terminate` and `boot` commands from the app id PR to process console.

This also restores my panic command which was removed for some reason at some point.

### Testing Strategy

Using the process console on hail.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/process_console.rs
+++ b/capsules/src/process_console.rs
@@ -140,7 +140,8 @@ pub static mut COMMAND_BUF: [u8; 32] = [0; 32];
 
 /// List of valid commands for printing help. Consolidated as these are
 /// displayed in a few different cases.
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):
Looks good, though until AppId is in it seems like `boot` basically does nothing, since a process can never be unstarted while the process console is running?

bradjc:
> Looks good, though until AppId is in it seems like `boot` basically does nothing, since a process can never be unstarted while the process console is running?

Aha, except for using the new `terminate` command with the process console!

bradjc:
ping!

I really want the `panic` command back :)

PR #3174, opened 2022-08-27
stm: add adc, gpio pins and temperature

alexandruradovici:
### Pull Request Overview

This pull request adds ADC, GPIO pins and Temperarture drivers to the Nucleo446re board.

### Testing Strategy

This pull request was tested using a Nucleo446re board.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alistair23 (dismissed):

bradjc (approved):

bradjc (changes requested):
Actually, seems like maybe RNG doesn't work.

alexandruradovici:
RNG seems to works, the IPC system seems to crash the kernel.

chips/stm32f429zi/src/trng.rs
```diff
@@ -0,0 +1,160 @@
+//! True random number generator
```
hudson-ayers, line 1: I do not believe it is necessary to copy this file here. It should be able to continue to live in `stm32f4xx/src/`, since you only actually instantiate the peripheral for the correct chips in the corresponding `interrupt_service.rs` files

hudson-ayers (changes requested):

chips/stm32f429zi/src/trng.rs
bradjc: I agree, but why would the interrupt servicer instantiate peripherals?

hudson-ayers: in hindsight that file should probably just be named `chip.rs`. But that file just defines a DefaultPeripherals struct which then implements the interrupt service trait. All upstream boards use this default struct, but out-of-tree boards can define a smaller set of peripherals to use and then re-implement the interrupt service trait if they want to.

alexandruradovici: I think @hudson-ayers made these changes when he introduced `DefaultPeripherals`

hudson-ayers: That is correct

bradjc: I prefer to just put it in a rng.rs file, ala: https://github.com/tock/tock/blob/master/chips/arty_e21_chip/src/uart.rs

alexandruradovici: @bradjc I'm not sure what you mean.

bradjc:
Can you cherry pick d9c34006d24204dfe08423386b17ab1d76722375?

Then I think this is ready to merge.

chips/stm32f429zi/src/trng.rs
hudson-ayers: I think the commit brad suggested is fine, but it is only really necessary to have any file in the specific chip crate if the peripheral is at different addresses depending on the chip, whereas it seems at least for the stm boards we support if the trng is present it is always at the same address, so the `StaticRef` can be safely declared in the stm32f4xx crate

bradjc: I'm not sure I agree, doing the static ref in the shared crate, and then letting any hw-specific chip crate have access to it is wrong (the other crates have safe access to an invalid pointer). If we say that the author of the specific crate needed to validate the `unsafe` before choosing a dependent crate, then that would address the problem but then some specific chip crates wouldn't be able to use the shared crate.

hudson-ayers: Oh, that's right, I didn't consider that once you have the `StaticRef` you can actually create a peripheral without any unsafe. I agree with you

alexandruradovici: I think I made the right changes.

hudson-ayers (approved):

bradjc (approved):

PR #3175, opened 2022-08-29
syscall: Fix SuccessU32U64 format

dcz-self:
TRD104 specifies that the 32-bit value goes before, not after the 64-bit one.

### Pull Request Overview

This pull request fixes syscall format to match TRD104 3.2 Return Values

### Testing Strategy

This pull request was tested by matching with libtock-rs in a test project.

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/syscall.rs
+++ b/kernel/src/syscall.rs
@@ -228,7 +228,7 @@ pub enum SyscallReturnVariant {
     SuccessU32U32 = 130,
     SuccessU64 = 131,
     SuccessU32U32U32 = 132,
-    SuccessU64U32 = 133,
+    SuccessU32U64 = 133,
 }
 
 /// Enumeration of the possible system call return variants specified
@@ -269,7 +269,7 @@ pub enum SyscallReturn {
     SuccessU64(u64),
     /// Generic success case, with an additional 32-bit and 64-bit
     /// data field
-    SuccessU64U32(u64, u32),
+    SuccessU32U64(u32, u64),
 
     // These following types are used by the scheduler so that it can
     // return values to userspace in an architecture (pointer-width)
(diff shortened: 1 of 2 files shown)


bradjc (approved):

hudson-ayers (approved):
I submitted https://github.com/tock/libtock-c/pull/294 to match this change

ppannuto (approved):
Nice catch. Mercifully nothing (in-tree at least) is using this yet, so hopefully no impact from the syscall behavior technically changing. I'm comfortable calling this a bugfix to match documented intent.

lschuermann:
Do we want to merge this before 2.1? I know we're technically changing ABI semantics w.r.t. allow and subscribe already, but this seems like it would constitute a major change in the system call ABI, would it not? I know we're probably not strictly following semantic versioning, but as far as I know changing the system call encoding was one of the arguments for performing a major version bump with 2.0. Alternatively, we might just be able to update the TRD instead, to match the current implementation?

lschuermann:
Okay, didn't see @ppannuto's comment. If nothing is using this yet, I'm fine with this change in 2.1.

ppannuto:
Yeah, IIRC the TRD was designed with some understanding of architectural calling conventions here: namely, when passing a 64-bit value, ARM will pass it in `{r0,r1}` or `{r2,r3}`, so aligning it this way allows the whole return to live in registers; while the other way would require some manipulation in the low-level bits of libtock-c after the syscall once things are types again.

brghena:
@ppannuto That was exactly the information I was looking for: whether the TRD or implementation were correct given real-world constraints. It looks like the TRD was correct, which makes this indeed a bugfix.

bradjc:
I don't see how we can merge this without approval from either of the two lead architects of the 2.0 syscalls.

ppannuto:
Sorry, I got trigger happy with our goals of trying to get the release out the door.

lschuermann:
> Yeah, IIRC the TRD was designed with some understanding of architectural calling conventions here: namely, when passing a 64-bit value, ARM will pass it in `{r0,r1}` or `{r2,r3}`, so aligning it this way allows the whole return to live in registers; while the other way would require some manipulation in the low-level bits of libtock-c after the syscall once things are types again.

@ppannuto This makes sense, and I recall these discussions around these issues, which makes me wonder why we haven't implemented the proper encoding. One very likely explanation is indeed that we haven't used this particular encoding yet. I was on the fence about this change when I saw it, which also motivated my above comment, since -- even though this wasn't used **by us (i.e. capsules)** -- it's still a change to the *implemented* ABI contract. I'd say that with this change merged, we should at the very least take a look at the ASM on the kernel and userspace side of code using this particular return value encoding, and perform some basic sanity tests for good measure before 2.1, just like we did when developing 2.0.

PR #3177, opened 2022-08-29
Screen HIL: Fix set_brightness

dcz-self:
The set_brightness method was not causing a callback, confusing the syscall driver and revealing a design/documentation mistake.

### Pull Request Overview

This pull request fixes an oversight in Display HIL.

I put some non-functional improvements to LPM013M126 driver on top.

### Testing Strategy

This pull request was tested by applying to a privte branch and running on LPM013M126 display.

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/lpm013m126.rs
+++ b/capsules/src/lpm013m126.rs
@@ -459,6 +459,7 @@ where
                 self.frame_buffer
                     .take()
                     .map_or(Err(ErrorCode::NOMEM), |mut frame_buffer| {
+                        // TODO: reject if buffer is shorter than frame
                         frame_buffer.blit(&buffer[..cmp::min(buffer.len(), len)], &frame);
                         let send_buf = FrameBuffer::with_raw_rows(
                             frame_buffer,
@@ -543,7 +544,7 @@ where
 
     fn set_brightness(&self, _brightness: usize) -> Result<(), ErrorCode> {
         // TODO: add LED PWM
-        Ok(())
+        Err(ErrorCode::NOSUPPORT)
     }
 
     fn set_invert(&self, _inverted: bool) -> Result<(), ErrorCode> {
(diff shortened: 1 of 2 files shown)


alistair23 (approved):

bradjc:
Isn't the fix to return an error instead of `Ok(())`? It looks like this isn't actually setting brightness.

dcz-self:
Good point. The problem I aimed to fix here was "hangs instead of progressing". I didn't consider what the return value for "I don't do brightness" should be. NOSUPPORT perhaps?

bradjc:
> Good point. The problem I aimed to fix here was "hangs instead of progressing". I didn't consider what the return value for "I don't do brightness" should be. NOSUPPORT perhaps?

agreed

dcz-self:
I have no idea why it started failing. I even rebased. This doesn't seem related at first glance.

brghena:
It's not your fault that it's failing. There's a CI bug right now causing all PRs to fail and we're working on it. You can ignore the `litex-sim-ci` for now.

capsules/src/lpm013m126.rs
```diff
@@ -543,7 +543,17 @@ where
 
     fn set_brightness(&self, _brightness: usize) -> Result<(), ErrorCode> {
         // TODO: add LED PWM
-        Ok(())
+        let scheduled = schedule_deferred(
+            self.deferred_caller,
+            &self.command_complete_callback,
+            "set_brightness complete",
+        );
```
bradjc, line 550: Wait why does this need to set a deferred call? The client should not expect a callback if the `set_brightness()` command fails.

dcz-self: You're right, I did that change too mechanically.

bradjc (approved):

alistair23 (approved):

PR #3181, opened 2022-08-29
stm32f412g: make user/kernel code boundary fall on flash sector boundary

hudson-ayers:
### Pull Request Overview

This pull request makes the user/kernel code boundary fall on flash sector boundary, which should hopefully enable tockloader to work with the `--bundle-apps flag.`

### Testing Strategy

@alexandruradovici can you test this with tockloader?

### TODO or Help Wanted

This pull request still needs testing

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/stm32f412gdiscovery/chip_layout.ld
+++ b/boards/stm32f412gdiscovery/chip_layout.ld
@@ -1,13 +1,13 @@
 /* Memory layout for the STM32F412G
  * rom = 1MB (LENGTH = 0x01000000)
- * kernel = 192KB
- * user = 832KB
+ * kernel = 256KB
+ * user = 744KB
  * ram = 256KB */
 
 MEMORY
 {
-  rom (rx)  : ORIGIN = 0x08000000, LENGTH = 0x00030000
-  prog (rx) : ORIGIN = 0x08030000, LENGTH = 0x000D0000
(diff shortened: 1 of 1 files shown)


bradjc:
We also need to tell tockloader to use 0x40000 not 0x30000.

bradjc:
I would like to merge this but I'm worried it will break app flashing in a way that is difficult to tell.

hudson-ayers:
Looking at tockloader, there is only a single stm32f4discovery profile (https://github.com/tock/tockloader/blob/master/tockloader/tockloader.py#L112), and it seems to have the settings from the stm32f446rediscovery, rather than the stm32f412gdiscovery (which has a different layout). The profile of the f446re is actually very close to what this PR changes the f412g layout to be, with the exception of it having *less* RAM than the f412g, which should not be a problem.

So I think merging this would strictly improve things relative to Tockloader's expectations, and I don't think this PR is blocked on any changes to tockloader. @alexandruradovici if you have a chance to test this PR with tockloader master, that would be great!

bradjc (approved):
As long as this shouldn't break existing workflows looks good to me.

alexandruradovici:
I will try to test this asap.

bradjc:
Good enough.

bors r+

PR #3185, opened 2022-08-30
readme: Update for 2.1

bradjc:
### Pull Request Overview

For 2.1 we should update the main readme, as 2.0 is no longer brand new.

I'm not sure what we should put. I just removed most of what we had there, hoping that by now we don't need the disclaimers. Please add commits as you see fit.

### Testing Strategy

n/a

### TODO or Help Wanted

What text should we have about 2.x?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,9 @@ from Tock 1.x, including:
 - Updated kernel types.
 - Many new and improved HILs.
 
+For a summary of the latest new features and improvements, check out the
+[changelog](CHANGELOG.md).
+
 
 Learn More
 ----------
--- a/README.md
+++ b/README.md
@@ -17,27 +17,16 @@ kernel.
 
 [tock-ci]: https://github.com/tock/tock/actions?query=branch%3Amaster+workflow%3Atock-ci
 
-Tock 2.0!
+Tock 2.x!
 ---------
 
-Tock is now on its second major release! Here are some 2.0 highlights, and see
-the [release notes](https://github.com/tock/tock/releases/tag/release-2.0) for
-more detail, or the [changelog](CHANGELOG.md#new-in-20) for the complete set of
-changes.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,28 +17,17 @@ kernel.
...
 - Many new and improved HILs.
 
-As 2.0 includes many breaking changes, to use the new kernel you will need to
-ensure you have updated versions of userspace apps and the various Tock tools.
-We recommend if you are using git that you do a `git pull`, and if you are using
-tagged releases be sure to update to the 2.0 release.
-
-Initially, only [libtock-c](https://github.com/tock/libtock-c) is compatible
-with Tock 2.0. Work on supporting Tock 2.0 for libtock-rs is [under
-development](https://github.com/tock/libtock-rs/issues/322).
```
brghena, line 40: Seeing as https://github.com/tock/libtock-rs/issues/322 is still open, we should probably keep this note.

jrvanwhy: `libtock-rs` *works* on Tock 2.1, it's just *incomplete*. In particular, it is missing test infrastructure and documentation. That sentence was true for 2.0 because the Tock 2.0 `libtock-rs` wasn't usable at the time of the Tock 2.0 release.

brghena: Never mind then! As long as the basics work at least, we definitely shouldn't steer people away from it.

ppannuto:
I think the way this is written, the bullet points of what makes Tock2 different from Tock1 still hold. Maybe could put back something like, `For a summary of the latest new features and improvements [in recent point releases?], check out the [Changelog].`

bradjc:
Please add commits!

hudson-ayers (approved):
bors r+

PR #3190, opened 2022-09-01
Add "Squared" Capsule

bradjc:
### Pull Request Overview

This pull request adds a capsule for squaring a number. It's actual purpose is to support testing the U32U64 command return type.

### Testing Strategy

Writing a userspace app and verifying the squared value comes back successfully.

### TODO or Help Wanted

This isn't the most...useful capsule. However to test the userspace side as well we need a capsule that implements the less common return types. 

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/capsules/src/squared.rs
+++ b/capsules/src/squared.rs
@@ -1,0 +1,41 @@
+//! Provide userspace service to calculate the square of a number.
+//!
+//! This uses the U32U64 command return type.
+
+use kernel::syscall::{CommandReturn, SyscallDriver};
+use kernel::ErrorCode;
+use kernel::ProcessId;
+
+/// Syscall driver number.
+use crate::driver;
+pub const DRIVER_NUM: usize = driver::NUM::Squared as usize;
+
+pub struct Squared {}
+
+impl SyscallDriver for Squared {
+    fn command(
+        &self,
+        command_type: usize,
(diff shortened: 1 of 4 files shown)


hudson-ayers (approved):

twilfredo (approved):

ppannuto:
Should we just add an 'ABI test' capsule instead? One that implements every possible syscall type?

bradjc:
> Should we just add an 'ABI test' capsule instead? One that implements every possible syscall type?

Yeah, probably.

bradjc:
I'm planning on updating this capsule to include all of the return values.

capsules/src/squared.rs
```diff
@@ -0,0 +1,41 @@
...
+        _appid: ProcessId,
+    ) -> CommandReturn {
+        match command_type {
+            0 => CommandReturn::success(),
+
+            // Square a number.
+            1 => {
+                let multiplicand: u64 = arg1 as u64;
+                let multiplier: u64 = arg1 as u64;
+                let product = multiplicand * multiplier;
```
twilfredo, line 30:
```suggestion
                let product = multiplicand.saturating_mul(multiplier);
```
Would something this be better to avoid overflow/wrapping?

PR #3198, opened 2022-09-07
boards: linker: remove MPU alignment

bradjc:
As far as I know, the `MPU_MIN_ALIGN` alignment directive is very old in Tock, probably from a much simpler MPU implementation. I'm not sure why we still have it, as the MPU implementation should handle alignment, not the linker.

### Testing Strategy

I tested this on hail, and the MPU app memory code sets up the same regions before and after this change since the mpu implementation first sets the region to a power of two and then aligns to that region size.

### TODO or Help Wanted

Does this cause any issue for RISC-V platforms?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


brghena (approved):
This makes sense to me. The start of app memory can be at any possible address, and then the kernel can decide from there where it actually wants to grab chunks and allocate them based on MPU requirements.

hudson-ayers:
Removing this would mean that `_sappmem` could be placed not aligned with an MPU region, such that the some portion of "app_memory" would not actually be usable for applications, which is semantically a bit weird, but I guess not a big deal.

I do think this would break anyone that statically places applications (cc @jettr / @a-pronin ) and uses our `kernel_layout.ld` as a starting point for that, but I suspect most users who do so would be using a modified copy our of our layout rather than including our layout and building on it.

bradjc:
> Removing this would mean that `_sappmem` could be placed not aligned with an MPU region, such that the some portion of "app_memory" would not actually be usable for applications, which is semantically a bit weird, but I guess not a big deal.

We could set the alignment to 256 bytes on cortex-m platforms, so that it is technically possible it could be used. However, that seems like something we would do since we already have the variable, and not something we would add if this variable had never existed.

hudson-ayers:
If `MPU_MIN_ALIGN` never existed, would `_sappmem` instead be named `_ekernelmem`? (since it doesn't actually show the location of the first app, just the last address used by the kernel). With this change, it seems like `_sappmem` and `_ezero` should always be the same address anyway.

bradjc:
> If `MPU_MIN_ALIGN` never existed, would `_sappmem` instead be named `_ekernelmem`? (since it doesn't actually show the location of the first app, just the last address used by the kernel). With this change, it seems like `_sappmem` and `_ezero` should always be the same address anyway.

We don't necessarily want to preclude users from putting in other sections by embedding assumptions about where app mem starts.

hudson-ayers (approved):
LGTM

ppannuto:
bors r+

I dug back in history a bit: it looks like this dates back to when userland was in this repo and built together, and it was used to place the beginning of applications. Definitely not needed now :)

PR #3200, opened 2022-09-07
cortex-m: mpu: do not give all memory to process in certain case

bradjc:
The MPU should not make decisions about how much memory to give to a process, the kernel should decide.

It is not clear to me why this check if the kernel does not need any process memory then give the process all of the memory was included.

### Testing Strategy

Quickly using hail.

### TODO or Help Wanted

Does anyone know if this covers a specific use case? The caller should just request the full amount of RAM rather than relying on the MPU to make this decision.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/mpu.rs
+++ b/arch/cortex-m/src/mpu.rs
@@ -595,13 +595,8 @@ impl<const NUM_REGIONS: usize, const MIN_REGION_SIZE: usize> mpu::MPU
         // multiple of an eighth of the MPU region length.
 
         // Determine the number of subregions to enable.
-        let mut num_subregions_used = {
-            if initial_kernel_memory_size == 0 {
-                8
-            } else {
-                initial_app_memory_size * 8 / region_size + 1
-            }
-        };
+        // Want `round_up(app_memory_size / subregion_size)`.
+        let mut num_subregions_used = initial_app_memory_size * 8 / region_size + 1;
 
         let subregion_size = region_size / 8;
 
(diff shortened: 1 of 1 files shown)


bradjc:
Ping? I don't think we want the MPU to decide how much memory apps should get. This also conflicts with the brk memop call.

hudson-ayers:
Yeah, this seems weird. It took me awhile to review this PR because it was not easy for me to understand the logic before/after the change. If I understand correctly, the current logic is:

1. process::create() decides the amount of memory a process requires for userspace, and for its grant region
2. MPU region size is determined as the closest power-of-2 size larger than the total requested memory for the process; or 256 if less than 256 bytes requested. Subregion size is just this / 8.
3. Determine how many subregions to enable. If the grant region size is > 0, determine the number of enabled subregions for this process' MPU region via `roundup(initial_app_memory_size / (region_size / 8))`. Effectively, enable all the subregions containing any of the memory requested by the process. (The grant memory is at the end of the region, and thus that MPU subregion is never enabled by this process, and if it would be we just double the total memory requirement for that app to fix it.)
4. Anytime an app requests more memory (via `brk`) we give it more memory by just enabling some more subregions, up until the point that doing so would touch the subregion that covers the grant memory, at which point we return an Error.
5. If in step 3), the grant_region_size is == 0, we just enable all the subregions and let the app touch any of its memory it wants regardless of if it has called break, because we don't have to protect the grant region anyway since this app does not have one. This is a little weird, because it means that for apps with a grant region, you can only write a little bit past your app break before getting an MPU fault (specifically, you can write past your app break up until the next subregion boundary, wherever that is), whereas for apps with no grant region, you can write past your app break all the way up to the end of the entire MPU region without getting an MPU fault.

This PR just removes the edge case in step 5, such that the MPU is always configured to only allow apps to write to subregions they have specifically requested access to based on initial memory requirements in the tbf header + calls to `brk()`.

That makes sense to me, I see no real upside from letting apps with no grant region write further past their break before generating a fault.

hudson-ayers (approved):

ppannuto:
bors r+

Hudson, thank you for the extremely helpful summary. I wonder if when we _first_ started with the static array of process structs, if a board image didn't include any capsules that used grants (so no empty grant pointers) if it was possible to have an empty grant space? Not relevant any more however.

PR #3202, opened 2022-09-07
RFC: Cortex-M: Use two MPU regions for app memory rather than one

bradjc:
### Pull Request Overview

Currently the Cortex-M MPU uses a single MPU region for process memory. That region must be a power of two and aligned to its size. In practice, this can lead to a huge amount of wasted memory between the end of the kernel's memory and the start of process memory. For example, on hail with the c_hello app installed, the memory map looks like:

```
tock$ process c_hello
𝐀𝐩𝐩: c_hello   -   [Yielded]
 Events Queued: 0   Syscall Count: 8   Dropped Upcall Count: 0
 Restart Count: 0
 Last Syscall: Yield { which: 1, address: 0x0 }
 Completion Code: None


 ╚0x20008000═╪══════════════════════════════════════════╝
             │ Grant Ptrs      112
             │ Upcalls         360
             │ Process         880
  0x20007AB8 ┼───────────────────────────────────────────
             │ ▼ Grant          76
  0x20007A6C ┼───────────────────────────────────────────
             │ Unused
  0x20006A04 ┼───────────────────────────────────────────
             │ ▲ Heap            0 |   4200               S
  0x20006A04 ┼─────────────────────────────────────────── R
             │ Data            516 |    516               A
  0x20006800 ┼─────────────────────────────────────────── M
             │ ▼ Stack         128 |   2048
  0x20006780 ┼───────────────────────────────────────────
             │ Unused
  0x20006000 ┴───────────────────────────────────────────
             .....
  0x00040800 ┬─────────────────────────────────────────── F
             │ App Flash      1972                        L
  0x0004004C ┼─────────────────────────────────────────── A
             │ Protected        76                        S
tock$ kernel
Kernel version: 2.1 (build release-2.1-1-gefb03eadb)

 ╔═══════════╤══════════════════════════════╗
 ║  Address  │ Region Name    Used (bytes)  ║
 ╚0x20004470═╪══════════════════════════════╝
             │   BSS         13424
  0x20001000 ┼─────────────────────────────── S
             │   Relocate        0            R
  0x20001000 ┼─────────────────────────────── A
             │ ▼ Stack        4096            M
  0x20000000 ┼───────────────────────────────
             .....
  0x0002B88C ┼─────────────────────────────── F
             │   RoData      26858            L
  0x00024FA2 ┼─────────────────────────────── A
             │   Code        85922            S
  0x00010000 ┼─────────────────────────────── H
```
So between the end of the kernel RAM at 0x20004470 and the start of the process RAM at 0x20006000, there are 7056 bytes of unused memory.

## Question

What, if anything, should we do about the wasted RAM on Cortex-M platforms?

## This PR

This PR switches the MPU to use two regions for app memory, rather than one. The amount of ram given to processes stays the same, it is just split evenly between two regions, so each region is half of the size and therefore aligned to half the size.

```
tock$ process c_hello
𝐀𝐩𝐩: c_hello   -   [Yielded]
 Events Queued: 0   Syscall Count: 8   Dropped Upcall Count: 0
 Restart Count: 0
 Last Syscall: Yield { which: 1, address: 0x0 }
 Completion Code: None


 ╚0x20007000═╪══════════════════════════════════════════╝
             │ Grant Ptrs      112
             │ Upcalls         360
             │ Process         880
  0x20006AB8 ┼───────────────────────────────────────────
             │ ▼ Grant          76
  0x20006A6C ┼───────────────────────────────────────────
             │ Unused
  0x20005A04 ┼───────────────────────────────────────────
             │ ▲ Heap            0 |   4200               S
  0x20005A04 ┼─────────────────────────────────────────── R
             │ Data            516 |    516               A
  0x20005800 ┼─────────────────────────────────────────── M
             │ ▼ Stack         128 |   2048
  0x20005780 ┼───────────────────────────────────────────
             │ Unused
  0x20005000 ┴───────────────────────────────────────────
             .....
  0x00040800 ┬─────────────────────────────────────────── F
             │ App Flash      1972                        L
  0x0004004C ┼─────────────────────────────────────────── A
             │ Protected        76                        S
tock$ kernel
Kernel version: 2.1 (build release-2.1-2-g204b40400)

 ╔═══════════╤══════════════════════════════╗
 ║  Address  │ Region Name    Used (bytes)  ║
 ╚0x20004470═╪══════════════════════════════╝
             │   BSS         13424
  0x20001000 ┼─────────────────────────────── S
             │   Relocate        0            R
  0x20001000 ┼─────────────────────────────── A
             │ ▼ Stack        4096            M
  0x20000000 ┼───────────────────────────────
             .....
  0x0002B93C ┼─────────────────────────────── F
             │   RoData      26836            L
  0x00025068 ┼─────────────────────────────── A
             │   Code        86120            S
  0x00010000 ┼─────────────────────────────── H
```

This also has the effect of making the subregions half the size, so we get finer-grained protection around the app_brk.

## Downsides

- Adds a little bit more complexity
- Uses up another of the scarce MPU regions

### Testing Strategy

Running apps on hail

### TODO or Help Wanted

Is this worth it? Should we ignore this problem? Anyone have other ideas?

Another question: why stop at two? Fairly easy to implement with two, but we could use more.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/mpu.rs
+++ b/arch/cortex-m/src/mpu.rs
@@ -164,7 +164,7 @@ pub struct CortexMConfig<const NUM_REGIONS: usize> {
 /// Regions 0-APP_MEMORY_REGION_MAX_NUM are used for application RAM. Regions
 /// with indices above APP_MEMORY_REGION_MAX_NUM can be used for other MPU
 /// needs.
-const APP_MEMORY_REGION_MAX_NUM: usize = 0;
+const APP_MEMORY_REGION_MAX_NUM: usize = 1;
 
 impl<const NUM_REGIONS: usize> Default for CortexMConfig<NUM_REGIONS> {
     fn default() -> Self {
@@ -548,6 +548,9 @@ impl<const NUM_REGIONS: usize, const MIN_REGION_SIZE: usize> mpu::MPU
         Ok(())
     }
 
+    // When allocating memory for apps, we use two regions, each a power of two
+    // in size. By using two regions we halve their size, and also halve their
+    // alignment restrictions.
     fn allocate_app_memory_region(
         &self,
         unallocated_memory_start: *const u8,
@@ -558,7 +561,8 @@ impl<const NUM_REGIONS: usize, const MIN_REGION_SIZE: usize> mpu::MPU
         permissions: mpu::Permissions,
         config: &mut Self::MpuConfig,
     ) -> Option<(*const u8, usize)> {
-        // Check that no previously allocated regions overlap the unallocated memory.
+        // Check that no previously allocated regions overlap the unallocated
+        // memory.
         for region in config.regions.iter() {
             if region.overlaps(unallocated_memory_start, unallocated_memory_size) {
                 return None;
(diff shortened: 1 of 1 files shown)


alistair23:
ls there a reason to save the MPU regions? On RISC-V we use them to protect the kernel, but I don't think ARM can do that right?

bradjc:
> ls there a reason to save the MPU regions? On RISC-V we use them to protect the kernel, but I don't think ARM can do that right?

The reason is because the current IPC implementation needs MPU regions for each shared buffer.

alistair23:
Ah, forgot about that.

Are we hitting the limit on ARM MPU regions? Otherwise the memory savings seem valuable

hudson-ayers (approved):
I think that the memory savings are worth losing one IPC region

hudson-ayers (approved):

lschuermann (approved):

ppannuto (dismissed):
bors r+

I read over this twice, and it all looks correct, but man if there is a place to prioritize effective unit tests, MPU configuration feels like it

ppannuto (approved):
bors r+

My kingdom for a clearer 'changes since you last viewed' :(

PR #3205, opened 2022-09-09
OpenTitan: Prepare to bump to the latest RTL SHA

alistair23:
### Pull Request Overview

This is just the first few commits of https://github.com/tock/tock/pull/3056, focusing on the ePMP improvements to handle already configured ePMP regions.

See each commit for details

### Testing Strategy

Tested as part of https://github.com/tock/tock/pull/3056

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/pmp.rs
+++ b/arch/rv32i/src/pmp.rs
@@ -629,8 +629,9 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::KernelM
                             csr::CSR.pmpaddr_set(x * 2, start >> 2);
 
                             // Set access to end address
-                            csr::CSR
-                                .pmpconfig_set(x / 2, cfg_val << 8 | csr::CSR.pmpconfig_get(x / 2));
+                            let new_cfg =
+                                cfg_val << 8 | (csr::CSR.pmpconfig_get(x / 2) & 0xFFFF_00FF);
+                            csr::CSR.pmpconfig_set(x / 2, new_cfg);
                             // Lock the CSR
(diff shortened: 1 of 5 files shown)


twilfredo (approved):

arch/rv32i/src/pmp.rs
```diff
@@ -629,8 +633,9 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::KernelM
                             csr::CSR.pmpaddr_set(x * 2, start >> 2);
 
                             // Set access to end address
-                            csr::CSR
-                                .pmpconfig_set(x / 2, cfg_val << 8 | csr::CSR.pmpconfig_get(x / 2));
+                            let new_cfg =
+                                cfg_val << 8 | (csr::CSR.pmpconfig_get(x / 2) & 0xFFFF_00FF);
+                            csr::CSR.pmpconfig_set(x / 2, new_cfg);
```
bradjc, line 638: Isn't this what `pmpconfig_modify()` is for?

alistair23: The problem is that `cfg_val` is just a number. This is because it can either hold `csr::pmpconfig::pmpcfg::x1` or `csr::pmpconfig::pmpcfg::x3` values. So to be more flexible we accept any value for `cfg_val` then shift it as required

bradjc (approved):

arch/rv32i/src/pmp.rs
bradjc: I tried this, and yeah because there are actually 5 fields in those 8 bits, it is very cumbersome to use modify.

hudson-ayers (approved):

PR #3208, opened 2022-09-12
opentitan: add watchdog/aon_timer

twilfredo:
### Pull Request Overview

Uses the `Always on timer` (AON_TIMER) for Opentitan [hw_ip](https://docs.opentitan.org/hw/ip/aon_timer/doc/), to implement watchdog for OpenTitan. 

This AON_TIMER has both a `watchdog` timer and a `wakeup` timer. The driver implemented has support for both hw functionality but only runs the watch dog timer for now  (unused code tagged with `#[cfg(aon_wkup_timer)]`, that should be (?) compiled out and is left for reference/use later.

The following patch for [qemu](https://www.mail-archive.com/qemu-devel@nongnu.org/msg907656.html) adds support to the watchdog, so we can use this in the future with qemu also. 

### Testing Strategy

Driver has been written to support to latest version of OpenTitan supported by TockOS, testing was done on Verilator on this https://github.com/tock/tock/pull/3056. 

### TODO

- Tested on Verilator, would be good to test on CW130  (since the AON_TIMER runs on different clock rates on both devices)
- Implement a WakeUp timer interface (?) and connect it to this driver.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/lowrisc/src/aon_timer.rs
+++ b/chips/lowrisc/src/aon_timer.rs
@@ -1,0 +1,232 @@
+//! AON/Watchdog Timer Driver
+
+use kernel::platform;
+use kernel::utilities::registers::interfaces::{Readable, Writeable};
+use kernel::utilities::registers::{register_bitfields, register_structs, ReadWrite, WriteOnly};
+use kernel::utilities::StaticRef;
+
+// Based on the latest commit of OpenTitan supported by tock:
+// Refer: https://github.com/lowRISC/opentitan/blob/217a0168ba118503c166a9587819e3811eeb0c0c/hw/ip/aon_timer/rtl/aon_timer_reg_pkg.sv#L136
+register_structs! {
+    pub AonTimerRegisters {
+        //AON_TIMER: Alert Test Register
+        (0x000 => alert_test: WriteOnly<u32, ALERT_TEST::Register>),
+        //AON_TIMER: Wakeup Timer Control Register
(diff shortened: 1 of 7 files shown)


bradjc (approved):

hudson-ayers (approved):

twilfredo:
Ping! @alistair23

chips/lowrisc/src/aon_timer.rs
```diff
@@ -0,0 +1,232 @@
...
+        let prescaler = 0x1000; // 2^12
+        self.registers
+            .wkup_ctrl
+            .write(WKUP_CTRL::PRESCALER.val(prescaler));
+    }
+
+    /// Set threshold ticks for the wake-up interrupt,
+    /// Note: the timing here depends on the prescaler set in
+    /// `set_wkup_prescaler()`
+    #[cfg(aon_wkup_timer)]
```
alistair23, line 163: What is this `cfg` for?

twilfredo: Since we aren't using the `aon wakeup timer` it's just there to compile out this code. I left the code there for reference, perhaps if we end up implementing this timer later, you could remove the tags and call it.

alistair23: Ha, I should have read the description

alistair23 (dismissed):

twilfredo:
Rebased, interrupt line names had changed in the recent OT bump.

bradjc (approved):

PR #3214, opened 2022-09-13
chips/earlgrey/aes: Fixup AES for Verilator

twilfredo:
### Pull Request Overview

On the latest commit of OpenTitan supported by Tock (https://github.com/tock/tock/pull/3056), on `Verilator` AES does not work as per the tests. The driver returns `ErrorCode::BUSY` since the hw_ip has not moved into the `idle` state when we expect. This issue only occurs on Verilator and works as expected on HW.

This patch adds a Verilator specific change to make sure we add a small wait where appropriate. Now passing all AES tests on Verilator.

### Testing Strategy

Running the `aes_test` on Verilator. 

````
OpenTitan initialisation complete. Entering main loop
check run AES128 CCM...
AES CCM* encryption/decryption tests
aes_ccm_test passed: (current_test=0, encrypting=true, tag_is_valid=true)
aes_ccm_test passed: (current_test=0, encrypting=false, tag_is_valid=true)
aes_ccm_test passed: (current_test=1, encrypting=true, tag_is_valid=true)
aes_ccm_test passed: (current_test=1, encrypting=false, tag_is_valid=true)
aes_ccm_test passed: (current_test=2, encrypting=true, tag_is_valid=true)
aes_ccm_test passed: (current_test=2, encrypting=false, tag_is_valid=true)
    [ok]
check run AES128 ECB...
aes_test passed (ECB Enc Src/Dst)
aes_test passed (ECB Dec Src/Dst)
aes_test passed (ECB Enc In-place)
aes_test passed (ECB Dec In-place)
    [ok]
check run AES128 CBC...
aes_test passed (CBC Enc Src/Dst)
aes_test passed (CBC Dec Src/Dst)
aes_test passed (CBC Enc In-place)
aes_test passed (CBC Dec In-place)
    [ok]
check run AES128 CTR...
aes_test CTR passed: (CTR Enc Ctr Src/Dst)
aes_test CTR passed: (CTR Dec Ctr Src/Dst)
    [ok]

````

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/earlgrey/src/aes.rs
+++ b/chips/earlgrey/src/aes.rs
@@ -149,6 +149,25 @@ impl<'a> Aes<'a> {
         self.registers.status.is_set(STATUS::IDLE)
     }
 
+    /// Must wait for IDLE, for trigger to set.
+    /// On reset, AES unit will first reseed the internal PRNGs
+    /// for register clearing and masking via EDN, and then
+    /// clear all key, IV and data registers with pseudo-random data.
+    /// Only after this sequence has finished, the unit becomes idle
+    ///
+    /// NOTE: This is needed for Verilator, and is suggested by documentation
+    ///       in general.
+    /// Refer: https://docs.opentitan.org/hw/ip/aes/doc/#programmers-guide
+    fn wait_on_idle_ready(&self) -> Result<(), ErrorCode> {
+        for _i in 0..10000 {
+            if self.idle() {
+                return Ok(());
+            }
+        }
+        // AES Busy
+        Err(ErrorCode::BUSY)
+    }
+
     fn input_ready(&self) -> bool {
         self.registers.status.is_set(STATUS::INPUT_READY)
     }
@@ -295,9 +314,7 @@ impl<'a> hil::symmetric_encryption::AES128<'a> for Aes<'a> {
     }
 
     fn set_iv(&self, iv: &[u8]) -> Result<(), ErrorCode> {
-        if !self.idle() {
-            return Err(ErrorCode::BUSY);
-        }
+        self.wait_on_idle_ready()?;
 
         if iv.len() != AES128_BLOCK_SIZE {
             return Err(ErrorCode::INVAL);
@@ -323,9 +340,7 @@ impl<'a> hil::symmetric_encryption::AES128<'a> for Aes<'a> {
     }
 
     fn set_key(&self, key: &[u8]) -> Result<(), ErrorCode> {
-        if !self.idle() {
-            return Err(ErrorCode::BUSY);
-        }
+        self.wait_on_idle_ready()?;
 
         if key.len() != AES128_KEY_SIZE {
             return Err(ErrorCode::INVAL);
@@ -433,10 +448,7 @@ impl<'a> hil::symmetric_encryption::AES128<'a> for Aes<'a> {
 
 impl kernel::hil::symmetric_encryption::AES128Ctr for Aes<'_> {
     fn set_mode_aes128ctr(&self, encrypting: bool) -> Result<(), ErrorCode> {
-        if !self.idle() {
-            return Err(ErrorCode::BUSY);
-        }
-
+        self.wait_on_idle_ready()?;
         self.mode.set(Mode::AES128CTR);
 
         let mut ctrl = if encrypting {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by bradjc — CHANGES_REQUESTED** [review]
We should not use cfg features in chip crates. OT is an odd case where verilator and FPGA can differ, and we have `chip_config.rs` to contain those differences. We want to avoid the case where the actual peripheral code is heavily dependent on which features were set to understand what is happening.

**Comment by twilfredo** [comment]
> We should not use cfg features in chip crates. OT is an odd case where verilator and FPGA can differ, and we have `chip_config.rs` to contain those differences. We want to avoid the case where the actual peripheral code is heavily dependent on which features were set to understand what is happening.

Ah okay fair enough, the simplest solution I can think of would be to just call `wait_on_idle()` (on both platforms). Otherwise, aes won't work on Verilator.  What do you guys think @alistair23 @bradjc ?

**Comment by bradjc** [comment]
I'm ok with some sort of `aes_bootup_delay_loops` variable in chip_config, that is set to 0 for fpga and something else for verilator.

File: chips/earlgrey/src/aes.rs [hunk 1]
```diff
@@ -275,6 +294,10 @@ impl<'a> Aes<'a> {
 
 impl<'a> hil::symmetric_encryption::AES128<'a> for Aes<'a> {
     fn enable(&self) {
+        #[cfg(feature = "config_sim_verilator")]
+        {
+            self.wait_on_idle().unwrap();
```
bradjc, line 299: No `unwrap()`, at the very least an expect.

twilfredo: I removed, this entirely (turns out we don't need the wait for enable)

twilfredo:
> I'm ok with some sort of `aes_bootup_delay_loops` variable in chip_config, that is set to 0 for fpga and something else for verilator.

Thanks, Implemented the suggested changes.

chips/earlgrey/src/aes.rs
```diff
@@ -149,6 +151,31 @@ impl<'a> Aes<'a> {
         self.registers.status.is_set(STATUS::IDLE)
     }
 
+    /// Must wait for IDLE, for trigger to set.
+    /// On reset, AES unit will first reseed the internal PRNGs
+    /// for register clearing and masking via EDN, and then
+    /// clear all key, IV and data registers with pseudo-random data.
+    /// Only after this sequence has finished, the unit becomes idle
+    ///
+    /// NOTE: This is necessary for Verilator
```
alistair23, line 160: Is there a harm in just doing this for both hardware and ventilator? It's a bounded loop waiting for startup, which seems pretty reasonable

twilfredo: Probably not, might be a little faster on HW if the AES IP is not responsive (I guess?). Should we just add it to both then?

alistair23: Even when there is an OT ASIC, I still expect people to use Verilaotr. It's open RTL, I image people would be intested in using that . The implementation would be clearer without the extra config switches

bradjc: Anytime there is a wait loop it can be confusing: why was this added? is it needed all the time, or just on some chips? is this documented in the datasheet? if I remove this, what will break? Having the config switch provides that clarity, as it won't be included when it isn't needed, and makes it clear when it is.

alistair23: Wouldn't a comment just fix that?

bradjc (dismissed):

twilfredo:
Reading through the programming guide for the hwip, it does mention that ["software must ensure that the AES unit is idle by checking STATUS.IDLE"](https://docs.opentitan.org/hw/ip/aes/doc/#programmers-guide). Even though hardware is fast enough (at-least during tests) to return to idle, it might still be worth adding the bounded delay just to allow it a bit more time to process (when required).

Should we make this change platform agnostic  @alistair23 @bradjc ?

alistair23:
I think we should

bradjc:
With that comment and link, I think it's fine to have the loop in all cases.

twilfredo:
Reverted the Verilator specific delays, so now the changes apply in all cases. Tested by running the aes tests on Verilator.

phil-levis (approved):

alistair23 (approved):

bradjc (approved):

PR #3215, opened 2022-09-13
Added missing base frequency config for e310x

mateibarbu19:
### Pull Request Overview

Currently the SiFive FE310X chips have do not configure any bits in the Power Reset Clock Interrupt controller driver.

There was a [commit](https://github.com/tock/tock/commit/54d8c9d5c5f2a39321614f3809369abd7fdd2389) trying to decouple the setup done in `main.rs` for a board implementation from what a bootloader configures. But this also assumes that the bootloader leaves the chip running at the external crystal oscilator frequency.

I have found that in the case of the [BBC HiFive Inventor](https://www.hifiveinventor.com/) the bootloader configures the chip to run off the PLL which is configured to give a 100MHz frequency.

So I made the following changes inspired by this [Zephyr template](https://github.com/sifive/zephyr-sifive-freedom-template/blob/master/clock.c)

- bypass and power off the PLL (bypass = 1)
- PLL input driven from external HFXOSC oscillator, then pllout is driven 
    directly by it (refsel = 1)
- set the frequency divider to 1
- drive the final hfclk with the PLL output, bypassed or otherwise (sel = 1)
- disable the ring oscillator

While I got working result only by setting bypass to 1, I thought that disabling the unused ring oscilator and choosing, instead of it, the PLL input the external oscillator to drive the PLL when it will be activated. See the Clock Generation chapther in any of the FE310 chips manual for more details.

### Help Wanted

The changes I made work for the SiFive FE310-G003 chip, but I cannot test them for the FE310-G002 which is featured on the [HiFive1 Rev B](https://www.sifive.com/boards/hifive1-rev-b) board.

@alistair23 could you please test this if it works?

### DIFF
--- a/chips/sifive/src/prci.rs
+++ b/chips/sifive/src/prci.rs
@@ -1,5 +1,6 @@
 //! Power Reset Clock Interrupt controller driver.
 
+use kernel::utilities::registers::interfaces::ReadWriteable;
 use kernel::utilities::registers::{register_bitfields, ReadWrite};
 use kernel::utilities::StaticRef;
 
@@ -59,10 +60,17 @@ impl Prci {
     }
 
     pub fn set_clock_frequency(&self, frequency: ClockFrequency) {
-        let _regs = self.registers;
+        let regs = self.registers;
 
         match frequency {
-            ClockFrequency::Freq16Mhz => {}
+            ClockFrequency::Freq16Mhz => {
+                regs.pllcfg
+                    .modify(pllcfg::bypass::SET + pllcfg::refsel::SET);
+                regs.plloutdiv
+                    .modify(plloutdiv::divby1.val(1) + plloutdiv::div.val(0));
+                regs.pllcfg.modify(pllcfg::sel::SET);
+                regs.hfrosccfg.modify(hfrosccfg::enable::CLEAR);
+            }
         };
     }
 }


bradjc:
Thanks for submitting this! Let's see if we can get this tested on other sifive boards.

alistair23:
Looks good to me. I don't know if I even have a working HiFive board anymore so I don't think I'll be able to test this

ppannuto:
@gemarcano is actively developing on the hifive right now -- can you test this patch?

gemarcano:
Yes, I can test it today. What am I looking for, exactly? That it boots? Or are there any other aspects I can poke at to confirm that it's working properly? Any way to check the actual frequency the board is running at?

mateibarbu19:
> Yes, I can test it today. What am I looking for, exactly? That it boots? Or are there any other aspects I can poke at to confirm that it's working properly? Any way to check the actual frequency the board is running at?

I didn't find any way to discover the frequency either by reading a register or by hooking an oscilloscope to a pin. The way I deduced that the HiFive Inventor was running ar 100Mhz was by interpreting the register described in the PRCI chapter. And proved correct when configuring the UART baudrate with this base frequency.

Then main aspect to test if the UART interface still works and the timings. Thank you.

gemarcano:
Alright, so I think it is working. I built this branch, flashed it, flashed a couple of test applications, and I'm still able to see console logs through tockloader (from reading the code for the HiFive revB, it looks like the UART port might also be connected to an NXP MCU that's managing USB communication, so it does look like UART is negotiating timing correctly. Additionally I attached an external UART to USB converter, and I was also able to see the console).

On a bare-metal application one way to get a rough estimate is to have a tight, controlled loop of instructions (say, maybe a simple NOP and jump?), and a timer, and use the performance counters to count how many instructions executing in the time it took for the timer to fire, or something like that. Still not precise, but might give an estimate. Not sure how to do that with Tock, though.

In any case, at a minimum, UART works, and I can load applications on the HiFive1 RevB I have.

gemarcano:
Alright, I had checked out the wrong branch. I re-ran the tests with the right branch (e310x_frequency) and everything still works fine (UART and console output is fine).

bradjc (approved):

ppannuto (approved):
bors r+

PR #1618, opened 2020-02-21
[RFC] Simpler and more robust component interface

gendx:
### Context

While writing my first components in https://github.com/tock/tock/pull/1556, I hit some difficulties in how to design things.

This pull request is an RFC to refactor the components interface, with the goals of (1) being less verbose and (2) more difficult to use incorrectly. It contains:
- New ways of initializing static objects, in `kernel/src/common/utils.rs`.
- An example of how the refactoring would look like on the component side (`boards/components/src/alarm.rs`) as well as the board side (`boards/imix/src/main.rs`). I picked the alarm component, which uses the `static_init_half` pattern, and therefore things should generalize well to the other components as far as I understand it. But please let me know if there are more complex cases that components can handle.

#### Current rationale for `static_init_half`

In principle, static objects in kernel boards are initialized with the `static_init` macro.

However, this relies on defining a `static` buffer inline in the current function, which doesn't work in Rust when the `static` type depends on a generic type of the function (because the compiler would somehow have to instantiate a distinct `static` object for each type of the function - which may not be well-defined in general).

This led to splitting `static_init` into `*_helper` macros defined in components to create `static` buffers with concrete types, and a `static_init_half` macro that initializes the buffer with some value. Besides creating some confusion about what this and that half mean, I don't think exposing the internal details of `static_init` *in this way* is very robust, as it relies on too many assumptions.

- `static_init_half` expects a `MaybeUninit<T>` buffer that contains `uninit`. Giving it an already initialized buffer would be incorrect as it would erase the previous value without notice. However, the current API of `static_init_half` just takes a `MaybeUninit<T>`.
- The `*_helper` macros return a `&mut MaybeUninit<T>`, which doesn't prevent the buffer from being used directly, or passed to various `static_init_half`. The API should only allow to pass such a buffer to `static_init_half` and only once.
- `static_init_half` doesn't have to be a macro, it can be implemented as a function. The only part that needs to be in a macro is the `static mut BUF: MaybeUninit<$T> = MaybeUninit::uninit();`, because that cannot be put in a function generic on `$T` (due to Rust rules). But this part is already in the `*_helper` macros.

#### Proposed new API for static initialization

Instead, I propose the following API (see the changes in `kernel/src/common/utils.rs`).

- An `uninit_static_buf` macro, which takes a type and instantiates inline a `static mut` buffer, containing `MaybeUninit::uninit()`. The role of this macro is similar to the `*_helper` macros in components.
[...]

PR #3216, opened 2022-09-13
Fix UartMux to be able to use multiple muxes

alexandruradovici:
### Pull Request Overview

This pull request fixes the way in which `UartMux` is being instantiated. Before this PR:
- the `new` function from the `UartMuxComponent` would instantiate the `UartMux` using a `static_init!` macro. Due to this, instantiating two `UartMux` structures would result in the second instance overwriting the first one.
- the `RX_BUF` was defined within the `virtual_uart` capsule, meaning that two instances of `UartMux` would use the same buffer.

This PR:
- moves the instantiating of the `UartMux` in the `uart_mux_component_helper` macro.
- moves the instantiating of the `RX_BUF` in the `uart_mux_component_helper` macro.

#### Example

```rust
let uart_mux1 = components::console::UartMuxComponent::new(
    &peripherals.usart1,
    115200,
    dynamic_deferred_caller,
)
.finalize(components::uart_component_helper!(64));

let uart_mux2 = components::console::UartMuxComponent::new(
    &peripherals.usart2,
    115200,
    dynamic_deferred_caller,
)
.finalize(components::uart_component_helper!(64));
```

Without this PR `uart_mux2` overwrites `uart_mux1`, as the structure is allocated in the same memory space.

This PR enables the `usart2` for the STM32F3Discovery board.

This is similar to what the I2C and SPI components do.

### Testing Strategy

This pull request was tested using an STM32F3 Discovery board with two serial ports. One serial port was used for the kernel messages and process console, while the second one was used for applications.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/console.rs
+++ b/boards/components/src/console.rs
@@ -35,9 +35,14 @@ use capsules::console::DEFAULT_BUF_SIZE;
 #[macro_export]
 macro_rules! uart_mux_component_helper {
     () => {{
-        $crate::uart_mux_component_helper!(capsules::virtual_uart::RX_BUF_LEN)
+        use capsules::virtual_uart::MuxUart;
+        use core::mem::MaybeUninit;
+        use kernel::static_buf;
+        let UART_MUX = static_buf!(MuxUart<'static>);
+        let RX_BUF = static_buf!([u8; capsules::virtual_uart::RX_BUF_LEN]);
+        (UART_MUX, RX_BUF)
     }};
-    ($rx_buffer_len: expr) => {{
+    ($rx_buffer_len: literal) => {{
         use capsules::virtual_uart::MuxUart;
         use core::mem::MaybeUninit;
         use kernel::static_buf;
--- a/boards/components/src/console.rs
+++ b/boards/components/src/console.rs
@@ -12,12 +12,14 @@
 //! ```rust
 //! let uart_mux = UartMuxComponent::new(&sam4l::usart::USART3,
 //!                                      115200,
-//!                                      deferred_caller).finalize(());
+//!                                      deferred_caller).finalize(components::uart_mux_component_helper!(64));
 //! let console = ConsoleComponent::new(board_kernel, uart_mux)
 //!    .finalize(console_component_helper!());
 //! ```
 // Author: Philip Levis <pal@cs.stanford.edu>
 // Last modified: 1/08/2020
+
+use core::mem::MaybeUninit;
 
 use capsules::console;
 use capsules::virtual_uart::{MuxUart, UartDevice};
@@ -32,18 +34,29 @@ use kernel::static_init;
 
 use capsules::console::DEFAULT_BUF_SIZE;
 
-pub struct UartMuxComponent {
+#[macro_export]
+macro_rules! uart_mux_component_helper {
+    ($rx_buffer_len: literal) => {{
+        use capsules::virtual_uart::MuxUart;
+        use core::mem::MaybeUninit;
+        static mut UART_MUX: MaybeUninit<MuxUart<'static>> = MaybeUninit::uninit();
+        static mut RX_BUF: [u8; $rx_buffer_len] = [0; $rx_buffer_len];
+        (&mut UART_MUX, &mut RX_BUF)
+    }};
+}
+
+pub struct UartMuxComponent<const RX_BUF_LEN: usize> {
     uart: &'static dyn uart::Uart<'static>,
     baud_rate: u32,
     deferred_caller: &'static DynamicDeferredCall,
 }
 
-impl UartMuxComponent {
+impl<const RX_BUF_LEN: usize> UartMuxComponent<RX_BUF_LEN> {
     pub fn new(
         uart: &'static dyn uart::Uart<'static>,
         baud_rate: u32,
         deferred_caller: &'static DynamicDeferredCall,
-    ) -> UartMuxComponent {
+    ) -> UartMuxComponent<RX_BUF_LEN> {
         UartMuxComponent {
             uart,
             baud_rate,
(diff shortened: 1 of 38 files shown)


bradjc:
Does this conflict or overlap #3046?

alexandruradovici:
> Does this conflict or overlap #3046?

I don't think it does, #3046 does not seem to change anything for the mux initialization or the uart receive buffer. The `boards/components/src/console.rs` file is not changed.

hudson-ayers (approved):
Looks good to me -- it was always a problem that the receive buffer was defined in the capsule.

I am adding a new label for this PR to track changes that modify the board interface in a way that an out-of-tree board maintainer would need to replicate, just so we can track these types of changes between releases.

bradjc (changes requested):
I see that this is inline with the I2C component, but I'm not sure about this change. `static_ini_half!` was created for a very specific purpose: to support sharing components when the actual object is templated on a chip-specific type. Here there is no chip-specific type in the virtual uart object, and as such the console component didn't use static_init_half.

I don't think we should continue to use static_init_half for a different purpose (to allow the same component to be used multiple times). Also, there is currently no indication of which components can be used multiple times. If components are going to be maintainable and usable, I think we need something more clear about how they can be used.

hudson-ayers:
Ideally the policy should be that every component is reusable; but that would require a high-effort overhaul of all components to get there, and we have seen with https://github.com/tock/tock/pull/1618 that there isn't anyone motivated for a whole-scale components overhaul. IMO it seems like a shame to not at least make usability improvements on an as-needed basis, or to block those improvements on a very high-effort complete overhaul.

alexandruradovici:
@bradjc are you suggesting moving the arguments of `new` to the macro?

The `UartMux` needs to be initialized in a macro, otherwise it will be overwritten every time we call `new`.

bradjc:
I'm not suggesting we do a full change to components, or that we cannot upgrade the console component independently of all other components. What I'm saying is that `static_init_half!()` is a hack I added to allow us to share components across microcontrollers. According to our own documentation, we aren't supposed to be using static_init_half anymore:

https://github.com/tock/tock/blob/d21666de503955ab9bae7d1ed95c58b144b007e9/kernel/src/utilities/static_init.rs#L136-L142

But what I want to avoid is adding a new hack using an old hack to get re-usable components. I assume there must be some mechanism we can add or use that makes it clear that the console component is now reusable.

alexandruradovici:
So using `static_buf` instead would work?

alexandruradovici:
I updated it to use the `static_buf` macro.

capsules/src/virtual_uart.rs
```diff
@@ -52,9 +52,6 @@ use kernel::hil::uart;
 use kernel::utilities::cells::{OptionalCell, TakeCell};
 use kernel::ErrorCode;
 
-const RX_BUF_LEN: usize = 64;
```
bradjc, line 55: This should be left here, as the recommended rx buf len from the author of this capsule.

alexandruradovici: Yes, that makes sense.

boards/stm32f3discovery/src/main.rs
```diff
@@ -251,15 +251,15 @@ unsafe fn set_pin_primary_functions(
...
     });
 
-    // channel 3
-    gpio_ports.get_pin(PinId::PA02).map(|pin| {
-        pin.set_mode(stm32f303xc::gpio::Mode::AnalogMode);
-    });
+    // // channel 3
+    // gpio_ports.get_pin(PinId::PA02).map(|pin| {
+    //     pin.set_mode(stm32f303xc::gpio::Mode::AnalogMode);
+    // });
```
alistair23, line 257: This seems like an unrelated change

boards/litex/arty/src/main.rs
```diff
@@ -7,6 +7,7 @@
 #![cfg_attr(not(doc), no_main)]
 
 use capsules::virtual_alarm::{MuxAlarm, VirtualMuxAlarm};
+
```
alistair23, line 10: Extra newline

boards/stm32f3discovery/src/main.rs
alexandruradovici: Those are the pins used to test the second uart, reverted.

hudson-ayers (dismissed):

alistair23 (dismissed):

alexandruradovici:
I rebased it to use the new `static_buf` with `MaybeUninit`. I think this is good to go.

bradjc (dismissed):

alistair23 (dismissed):

alexandruradovici:
Rebased

boards/components/src/console.rs
```diff
@@ -28,22 +28,41 @@ use kernel::create_capability;
...
-use kernel::static_init;
 
 use capsules::console::DEFAULT_BUF_SIZE;
 
-pub struct UartMuxComponent {
+#[macro_export]
+macro_rules! uart_mux_component_static {
+    () => {{
+        use capsules::virtual_uart::MuxUart;
+        use core::mem::MaybeUninit;
```
bradjc, line 38:
```suggestion
```

boards/components/src/console.rs
```diff
@@ -28,22 +28,41 @@ use kernel::create_capability;
...
+        use kernel::static_buf;
+        let UART_MUX = static_buf!(MuxUart<'static>);
+        let RX_BUF = static_buf!([u8; capsules::virtual_uart::RX_BUF_LEN]);
+        (UART_MUX, RX_BUF)
+    }};
+    ($rx_buffer_len: literal) => {{
+        use capsules::virtual_uart::MuxUart;
+        use core::mem::MaybeUninit;
```
bradjc, line 46:
```suggestion
```

bradjc (changes requested):

bradjc (approved):

PR #3218, opened 2022-09-14
ESP32-c3 Build Fixes, add docs for flashing apps

TheConner:
### Pull Request Overview

This PR fixes the following warning from esptool when running `make flash-app`:

```
Warning: Image file at 0x0 is protected with a hash checksum, so not changing the flash mode setting. Use the --flash_mode=keep option instead of --flash_mode=dio in order to remove this warning, or use the --dont-append-digest option for the elf2image command in order to generate an image file without a hash checksum
```

It seems that the flashed apps seem to overwrite critical regions of flash, causing the board to enter a infinite loop of trying to boot tock and failing. In order to recover from this I had to re-flash the tock kernel and apps. 

Solution: add `--dont-append-digest` as the warning suggests, resolves issues with the boot loop. 

Additionally, I add some extra documentation detailing how to flash apps to the board. It should save others some time in the future.

To reproduce this issue:
- Build and flash tock kernel `make && make flash`
- Build an example app for RISCV target from libtock-c
- Flash example app to board `make flash-app APP=../../../libtock-c/examples/<some-app>/build/rv32imac/rv32imac.0x403B0060.0x3FCC0000.tbf`
- 
This results in an infinite loop of trying to boot tock and failing:
```
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0x10 (RTCWDT_RTC_RST),boot:0xc (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:QIO, clock div:2
load:0x40380000,len:0x125c4
ets_loader.c 78 
... etc
```

### Testing Strategy

Repeat the reproduction steps, it should not result in the bootloader failing to boot Tock. 

### TODO or Help Wanted

Someone else to validate this would be nice

### Documentation Updated

- [x] Updated the readme for the ESP32-c3 board. 

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/esp32-c3-devkitM-1/Makefile
+++ b/boards/esp32-c3-devkitM-1/Makefile
@@ -19,7 +19,7 @@ ifeq ($(APP),)
 	$(error "Please specify an APP to be flashed")
 endif
 	$(RISC_PREFIX)-objcopy --update-section .apps=$(APP) $^ $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.elf
-	esptool.py --port /dev/ttyUSB0 --chip esp32c3 elf2image --use_segments --output binary.hex $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.elf
+	esptool.py --port /dev/ttyUSB0 --chip esp32c3 elf2image --use_segments --output binary.hex $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM)-app.elf --dont-append-digest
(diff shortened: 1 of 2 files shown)


twilfredo:
Same problem as https://github.com/tock/tock/pull/3059 I believe. It looks like this was removed in https://github.com/tock/tock/pull/3148 . 

I think issue is that older versions of the `esptool.py` don't support `dont-append-digest` but the current release version installed with pip is `esptool.py v4.3` and support this. 

I tested the above with `esptool.py v4.3` and can reproduce the bug and is fixed with this patch.

alistair23 (approved):

hudson-ayers (approved):

Issue #2564, opened 2021-04-27
Tracking: Update `static_init!()` to `singleton!()` semantics


bradjc:
This is the tracking issue for the discussion on #1447. At a high level, if static_init is called twice it can lead to mutable memory aliasing.

- [ ] Update `static_init!()` to have `singleton!()` semantics.
  - Singleton: https://docs.rs/cortex-m/0.7.2/cortex_m/macro.singleton.html
  - Will not be able to disable interrupts.

Optional, or more discussion needed:
- [ ] Provide a safe interface for `static_init!()` if there is some way to guarantee it's safe whenever used.
  - https://github.com/tock/tock/issues/1447#issuecomment-827336421
  - Intent is it is only used during initialization before the kernel loop starts.

kupiakos:
Note: if a type is [`FromBytes`](https://docs.rs/zerocopy/latest/zerocopy/trait.FromBytes.html), meaning any bit pattern is valid, it could be safely mutated on a single-core system even during an interrupt, as the most any race could do is create an arbitrary byte pattern in the value, which is safe to do according to `FromBytes`'s safety guarantees. This doesn't solve the multiple `&mut` issue.

hudson-ayers:
https://github.com/tock/tock/pull/3239 took us almost all the way there, all that is left is making the macro safe to call from single-core systems

bradjc:
I propose we close this.

hudson-ayers:
I still think a safe `static_init!()` would be a very significant upgrade to Tock, so I would prefer we keep this open, even if no one is working on it now.

kupiakos:
An idea I'd brought up in a previous call: rather than protecting each singleton with a flag, we could protect a group of singletons with a single flag, or even better, have a primitive for ensuring a block of code can only run once.

PR #3219, opened 2022-09-14
kernel: convert static_buf!() to be like singleton

bradjc:
### Pull Request Overview

This PR changes `static_buf!()` to now panic if the same static buffer is used multiple times. This happens if the same `static_buf!()` expansion is run multiple times, such is if `static_buf!()` is in a function which is called multiple times. Since doing this causes memory to be aliased which is bad, we panic to let the board author know this needs to be fixed.

Implements #2564.

### Testing Strategy

I added

```rust
    let crc2 = components::crc::CrcComponent::new(
        board_kernel,
        capsules::crc::DRIVER_NUM,
        &peripherals.crccu,
    )
    .finalize(components::crc_component_helper!(sam4l::crccu::Crccu));
```
to the end of hail's main.rs, and when running I get:

```
[INFO   ]  ----- Waiting for serial port to reconnect...

panicked at 'ERROR! The same static_buf!() cannot be used multiple times.', boards/components/src/crc.rs:63:23
	Kernel version release-2.1-76-g4f51e0ed5

---| Cortex-M Fault Status |---
No Cortex-M faults detected.

---| App Status |---
```

### TODO or Help Wanted

Thoughts? Comments?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/Soundness.md
+++ b/doc/Soundness.md
@@ -63,10 +63,10 @@ However, in a few cases, it is useful to start with a mutable reference in order
 to enforce _who_ can make certain calls. For example, setting up buffers in the
 SPI driver is, for practical reasons, deferred until after construction but we
 would like to enforce that it can only be called by the platform initialization
-function (before `main` starts). This is enforced because all references after
-the platform is setup are immutable, and the `config_buffers` method takes an
-`&mut self` (_Note: it looks like this is not strictly necessary, so maybe not a
-big deal if we can't do this_).
+function (before the kernel loop starts). This is enforced because all
+references after the platform is setup are immutable, and the `config_buffers`
+method takes an `&mut self` (_Note: it looks like this is not strictly
+necessary, so maybe not a big deal if we can't do this_).
 
 ### Soundness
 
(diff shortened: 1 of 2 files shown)


hudson-ayers:
I pushed a commit that brings the size impact of this change on Imix from 4688 bytes to 1920 bytes. This is still a pretty big size hit, even though it is likely < 20 bytes per call to `static_init!()`.

alexandruradovici:
What about having the check only for debug builds? I think this is somehow consistent with Rust's approach to number overflowing.

hudson-ayers:
> What about having the check only for debug builds? I think this is somehow consistent with Rust's approach to number overflowing.

With this check, `static_init!()` will not have to be `unsafe`, which is very valuable in reducing the total amount of `unsafe` in the kernel. However if the checks only apply for debug builds we cannot remove `unsafe` since misuse can lead to UB. Also, Tock builds are `release` by default, so no one ever uses debug builds in practice.

bradjc:
I added a commit updating the `Soundness.md` documentation. Please take a look that what I added is correct.

doc/Soundness.md
```diff
@@ -63,10 +63,10 @@ However, in a few cases, it is useful to start with a mutable reference in order
...
 SPI driver is, for practical reasons, deferred until after construction but we
 would like to enforce that it can only be called by the platform initialization
-function (before `main` starts). This is enforced because all references after
-the platform is setup are immutable, and the `config_buffers` method takes an
-`&mut self` (_Note: it looks like this is not strictly necessary, so maybe not a
-big deal if we can't do this_).
+function (before the kernel loop starts). This is enforced because all
+references after the platform is setup are immutable, and the `config_buffers`
+method takes an `&mut self` (_Note: it looks like this is not strictly
+necessary, so maybe not a big deal if we can't do this_).
```
hudson-ayers, line 69: IMO, we could just remove this paragraph. The SPI buffers are held in `TakeCell`s now anyway, and using `&mut self` as a way to enforce that buffers are only configured at boot is not a pattern we have really carried over to other capsules.

hudson-ayers (approved):

bradjc:
Closed in favor of https://github.com/tock/tock/pull/3239

PR #3221, opened 2022-09-15
AppID: Convert ShortID to an enum

bradjc:
_This PR references AppID, not master._

### Pull Request Overview

Right now, `ShortID` is essentially an `Option<u32>`, where `Some(value)` represents a fixed, known 32 bit nonzero short ID, and `None` means the ShortID is not specifically known, but is guaranteed to be unique among all runnable processes. This means `ShortID` can be used to check for runnable process uniqueness, as well as support cases where the board does not need a policy for creating ShortIDs, as it can just use the `None` type to mean some arbitrary unique ShortID.

The downside to this approach in code is that this is somewhat counterintuitive from how `Option<ShortID>` looks at first glance. The normal interpretation of a `None` would be the process does not have a `ShortID`, but that is not quite correct.

This PR proposes some sort of enum for ShortID to remove the need for an `Option`, as all processes have a short ID, and makes the "we don't specifically know what the id is but it is unique" case more explicit.

Feedback would be helpful, as I know the `Option<NonZeroU32>` format currently is to ensure ShortID fits in a single word. Can we replicate that somehow, but while using our own names?

### Testing Strategy

draft

### TODO or Help Wanted

RFC

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.


hudson-ayers:
> Feedback would be helpful, as I know the Option<NonZeroU32> format currently is to ensure ShortID fits in a single word. Can we replicate that somehow, but while using our own names?

Looks like you already have! https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f33b0871f20e54e2e9f7951bb6eab35f

jrvanwhy:
> Feedback would be helpful, as I know the `Option<NonZeroU32>` format currently is to ensure ShortID fits in a single word. Can we replicate that somehow, but while using our own names?

Yes. As I understand it (this isn't documented well), niche optimization is guaranteed to occur when both of the following are true:

1. Exactly one enum variant has nonzero size
2. The number of niches in that enum variant is at least as large as the number of zero-sized enum variants.

ShortID meets both conditions, so it receives niche optimization and is 4 bytes in size.

jrvanwhy:
The previous design (`Option<ShortID>`) makes it really easy to accidentally think two different short IDs are the same, because `None == None`. With this PR, we can make them compare unequal, which is an improvement.

kernel/src/process.rs
```diff
@@ -196,10 +196,28 @@ impl ProcessId {
...
+    fn eq(&self, other: &Self) -> bool {
+        match self {
+            ShortID::LocalUnique => false,
+            ShortID::Fixed(a) => match other {
+                ShortID::LocalUnique => false,
+                ShortID::Fixed(b) => a == b,
+            },
+        }
+    }
 }
```
jrvanwhy, line 221: I find nested `match` statements hard to read. Maybe this is better?

```rust
impl PartialEq for ShortID {
    fn eq(&self, other: &Self) -> bool {
        match (self, other) {
            (Self::Fixed(a), Self::Fixed(b)) => a == b,
            _ => false,
        }
    }
}
```

Issue #3222, opened 2022-09-16
Tracking: Permit components to be used multiple times by the same board


bradjc:
Some components will not work if used multiple times in the same board. This is counterintuitive, and there are many good reasons to be able to use the same component twice.

The reason certain components cannot be used multiple times is that they use `static_init!()` or `static_buf!()` internally in such a way that causes static memory to be aliased. The fix is to only use `static_buf!()` in a macro which is called from the main.rs file directly, instead of in a function in the component.

Components which need to be updated:

- [x] air_quality
- [x] bme280
- [x] ccs811
- [x] console
- [x] crc
- [x] debug_queue
- [x] debug_writer
- [x] fxos8700
- [x] gpio
- [x] hts221
- [x] humidity
- [x] isl29035
- [x] lldb
- [x] mlx90614
- [x] nrf51822
- [x] process_console
- [x] process_printer
- [x] rng
- [x] screen
- [x] segger_rtt
- [x] si7021
- [x] sound_pressure
- [x] temperature
- [x] text_screen
- [x] touch
- [x] udp_driver

My rough search for components which likely have this issue:

```
❯ git grep -B 10 static_init\! | grep -B 5 -A 5 finalize
```

bradjc:
I believe all components that use `static_init!()` or `static mut` in `finalize()` have been updated.

PR #3223, opened 2022-09-16
Convert Temperature Component to support multiple instances

bradjc:
### Pull Request Overview

Part of #3222 

I thought the temperature component is somewhat simple, so this PR converts the component to (hopefully) support being used multiple times by the same board. I'm hoping we can discuss this PR to get a template for converting all the rest.

The main change here is adding:

```rust
#[macro_export]
macro_rules! temperature_component_static_init {
    () => {{
        kernel::static_buf!(capsules::temperature::TemperatureSensor<'static>)
    };};
}
```
to the component.

I chose the name `temperature_component_static_init` as the purpose of adding this macro is to encapsulate the static_init!() portion which is what needs to be instantiated in the main.rs file itself, not in the finalize() function. The name is not `_helper`, as I created the "_helper" name because those macros helped me create some way to enable component sharing across microcontrollers. I didn't have a better name at the time.

I'm thinking we should revisit the naming convention for the macro. Perhaps we just use `X_static!()` for these macros, to signify this is where the static buf is defined?

### Testing Strategy

travis

### TODO or Help Wanted

Can we do better than this?

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/boards/clue_nrf52840/src/main.rs
+++ b/boards/clue_nrf52840/src/main.rs
@@ -615,7 +615,7 @@ pub unsafe fn main() {
         capsules::temperature::DRIVER_NUM,
         sht3x,
     )
-    .finalize(());
+    .finalize(components::temperature_component_static!());
 
     let humidity = components::humidity::HumidityComponent::new(
         board_kernel,
--- a/boards/components/src/temperature.rs
+++ b/boards/components/src/temperature.rs
@@ -3,15 +3,23 @@
 //! Usage
 //! -----
 //! ```rust
-//! let temp = TemperatureComponent::new(board_kernel, nrf52::temperature::TEMP).finalize(());
+//! let temp = TemperatureComponent::new(board_kernel, nrf52::temperature::TEMP)
+//!     .finalize(components::temperature_component_static!());
 //! ```
 
 use capsules::temperature::TemperatureSensor;
+use core::mem::MaybeUninit;
 use kernel::capabilities;
 use kernel::component::Component;
 use kernel::create_capability;
 use kernel::hil;
-use kernel::static_init;
+
+#[macro_export]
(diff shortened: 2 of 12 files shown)


hudson-ayers (dismissed):
I think this PR looks good -- it lets the component be used for multiple instances in the "tock-idiomatic" way. 

I do think that given our experience with components so far we might want to consider making components just be macros in the future. `static_init!()` requires us to use macros for most components in order for them to be reusable anyway, and I have doubts that this half-macro / half-trait approach is any more reader-friendly than a pure-macro approach (which would also be much more brief, both in implementation and instantiation).

alistair23 (dismissed):

bradjc:
I changed the name to `temperature_component_static!()`, and I think we should only use that naming convention. The purpose of those macros is to handle the static defines---nothing else.

I also added some documentation to the porting guide.

hudson-ayers:
> I changed the name to `temperature_component_static!()`, and I think we should only use that naming convention. The purpose of those macros is to handle the static defines---nothing else.

Agreed this should be the purpose of these macros, though we do not follow this everywhere today (e.g. https://github.com/tock/tock/blob/master/boards/components/src/temperature_stm.rs#L11)

bradjc:
> > I changed the name to `temperature_component_static!()`, and I think we should only use that naming convention. The purpose of those macros is to handle the static defines---nothing else.
> 
> Agreed this should be the purpose of these macros, though we do not follow this everywhere today (e.g. https://github.com/tock/tock/blob/master/boards/components/src/temperature_stm.rs#L11)

Yes, working on it #3231

hudson-ayers (approved):

alistair23 (approved):

bradjc:
I guess I meant to

PR #3232, opened 2022-09-20
Component: Update bmX280 components

bradjc:
### Pull Request Overview

Update these components to use _static macro, static_buf, and use new() for all configuration.

### Testing Strategy

travis

### TODO or Help Wanted

Any thoughts on components using other components? bmp280 was using the I2C component, but it's hard to update all components at once, so I just replicated the i2c init because it is pretty simple.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/bme280.rs
+++ b/boards/components/src/bme280.rs
@@ -4,19 +4,19 @@
 //! -----
 //! ```rust
 //!     let bme280 =
-//!         Bme280Component::new(mux_i2c, 0x77).finalize(components::bme280_component_helper!());
+//!         Bme280Component::new(mux_i2c, 0x77).finalize(components::bme280_component_static!());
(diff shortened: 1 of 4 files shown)


hudson-ayers (dismissed):
Re: components using other components, I think we want to avoid that, unless the call to the `helper!()`/`static!()` macro for the other component is contained within the macro for the outer component. Otherwise, the component is still not reusable

alistair23 (dismissed):

phil-levis (approved):

alistair23 (approved):

hudson-ayers (dismissed):

ppannuto (dismissed):

ppannuto:
bors r+

looks good; and just waiting on mac build which should finish by the time bors gets to it

ppannuto (approved):

PR #3234, opened 2022-09-20
component: update ambient light

bradjc:
### Pull Request Overview

This pull request updates the ambient light component to use new static format.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/isl29035.rs
+++ b/boards/components/src/isl29035.rs
@@ -12,38 +12,46 @@
 //! Usage
 //! -----
 //! ```rust
-//! let isl = Isl29035Component::new(mux_i2c, mux_alarm)
-//!     .finalize(isl29035_component_helper!(sam4l::ast::Ast));
-//! let ambient_light = AmbientLightComponent::new(board_kernel, sensors_i2c, mux_alarm)
-//!     .finalize(isl29035_component_helper!(sam4l::ast::Ast));
+//! let isl29035 = Isl29035Component::new(mux_i2c, mux_alarm)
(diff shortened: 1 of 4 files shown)


hudson-ayers (dismissed):
It seems that before this PR, there were 2 components: `Isl29035Component` and `AmbientLightComponent`, and that `AmbientLightComponent` would setup the isl29035 and then use it to expose the `AmbientLight` driver, while one would use `Isl29035Component` only if one did not want to use it with the `AmbientLight` driver. Now, `AmbientLightComponent` does not include the setup of `Isl29035Component`, that has to be done seperately. 

I think this change makes sense, however, because it makes it much harder to accidentally call one of the static macros twice, so this looks good to me.

alistair23 (approved):

bradjc:
> It seems that before this PR, there were 2 components: `Isl29035Component` and `AmbientLightComponent`, and that `AmbientLightComponent` would setup the isl29035 and then use it to expose the `AmbientLight` driver, while one would use `Isl29035Component` only if one did not want to use it with the `AmbientLight` driver. Now, `AmbientLightComponent` does not include the setup of `Isl29035Component`, that has to be done seperately.
> 
> I think this change makes sense, however, because it makes it much harder to accidentally call one of the static macros twice, so this looks good to me.

Yes. I think this component was originally created when components were new and not as settled. This update makes it consistent with the others.

phil-levis (approved):

hudson-ayers (approved):

alistair23 (approved):

PR #3239, opened 2022-09-21
`static_buf!()`: Remove `StaticUninitializedBuffer` and `UninitializedBuffer` in favor of `MaybeUninit`, update component conventions to match

hudson-ayers:
### Pull Request Overview

This pull request builds on top of #3219 , with some additional optimizations for size to reduce the overhead introduced in #3219. This PR removes `StaticUninitializedBuffer` and `UninitializedBuffer`, since I believe they now offer very little over `MaybeUninit` today (other than a slightly nicer name for `initialize()` vs `write()`.

This combination of changes makes it no longer `unsafe` to call `static_init!()` or `static_buf!()`. A logical followup change is to remove `unsafe` from `static_buf!()` and make `Component::finalize()` safe. This change should allow us to remove >90% of the unsafe code in `main.rs`, a big jump forward for validating memory safety in Tock.

On Imix, this brings the overhead of #3219 down to 1104 bytes. Notably, once Brad finishes porting uses of `static_init_half!()` to `static_buf!()`, this overhead will increase.

Apologies for the weird commit ordering, where #3219 comes after my changes -- I implemented my changes first and then pulled in the #3219 changes in a squashed commit. Ultimately I think that #3219 should just be closed in favor of this PR.

This PR also defines the expected process for a component: _all_ statically declared memory is created in a macro and passed in to `finalize()`. That macro has a naming convention of `[name]_component_static!()`. No memory is statically allocated in `finalize()` (i.e. no static_init! or static_buf! in the finalize function). That macro _only_ statically declares memory.

### Testing Strategy

### TODO or Help Wanted

~~After this PR is merged, Brad's various component PRs will need to be rebased to use `MaybeUninit` as the `StaticInput` to components.~~

### Documentation Updated

- [x]  Updated.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/src/tests/hmac.rs
+++ b/boards/opentitan/src/tests/hmac.rs
@@ -72,21 +72,23 @@ impl<'a> digest::ClientVerify<32> for HmacTestCallback {
 
 /// Static init an HmacTestCallback, with
 /// respective buffers allocated for data fields.
-unsafe fn static_init_test_cb() -> &'static HmacTestCallback {
-    let input_data = static_init!([u8; 32], [32; 32]);
-    let digest_data = static_init!(
-        [u8; 32],
-        [
-            0xdc, 0x55, 0x51, 0x5e, 0x30, 0xac, 0x50, 0xc7, 0x65, 0xbd, 0xe, 0x2, 0x82, 0xf7, 0x8b,
-            0xe1, 0xef, 0xd1, 0xb, 0xdc, 0xa8, 0xba, 0xe1, 0xfa, 0x11, 0x3f, 0xf6, 0xeb, 0xaf,
-            0x58, 0x57, 0x40,
-        ]
-    );
+macro_rules! static_init_test_cb {
+    () => {{
+        let input_data = static_init!([u8; 32], [32; 32]);
+        let digest_data = static_init!(
+            [u8; 32],
+            [
+                0xdc, 0x55, 0x51, 0x5e, 0x30, 0xac, 0x50, 0xc7, 0x65, 0xbd, 0xe, 0x2, 0x82, 0xf7,
+                0x8b, 0xe1, 0xef, 0xd1, 0xb, 0xdc, 0xa8, 0xba, 0xe1, 0xfa, 0x11, 0x3f, 0xf6, 0xeb,
+                0xaf, 0x58, 0x57, 0x40,
+            ]
+        );
 
-    static_init!(
-        HmacTestCallback,
-        HmacTestCallback::new(input_data, digest_data)
-    )
+        static_init!(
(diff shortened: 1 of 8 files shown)


bradjc (approved):
I think this is a major step forward for components. The singleton-without-enforcement property of components has long been a pain point for Tock, and this is a step towards eliminating that rought edge.

bradjc (dismissed):

boards/opentitan/src/tests/hmac.rs
```diff
@@ -72,29 +72,31 @@ impl<'a> digest::ClientVerify<32> for HmacTestCallback {
...
+    };};
 }
 
 #[test_case]
 fn hmac_check_load_binary() {
     let perf = unsafe { PERIPHERALS.unwrap() };
     let hmac = &perf.hmac;
 
-    let callback = unsafe { static_init_test_cb() };
+    let callback = unsafe { static_init_test_cb!() };
```
hudson-ayers, line 99: I am surprised you don't get warnings that this unsafe is not needed

bradjc (dismissed):

chips/sifive/src/prci.rs
```diff
@@ -59,10 +60,17 @@ impl Prci {
...
         match frequency {
-            ClockFrequency::Freq16Mhz => {}
+            ClockFrequency::Freq16Mhz => {
+                regs.pllcfg
+                    .modify(pllcfg::bypass::SET + pllcfg::refsel::SET);
+                regs.plloutdiv
+                    .modify(plloutdiv::divby1.val(1) + plloutdiv::div.val(0));
+                regs.pllcfg.modify(pllcfg::sel::SET);
+                regs.hfrosccfg.modify(hfrosccfg::enable::CLEAR);
+            }
```
alistair23, line 73: Why is this change in here?

bradjc: Rebased and fixed.

bradjc (dismissed):

alevy (approved):

alevy:
The most recent version of this PR (which @bradjc just updated) requires `unsafe` to use `static_`.

alevy:
There was some discussion outside GitHub about the desire for a zero-cost variant that doesn't dynamically check if a value has already been initialized, though the uses of this would likely be only out-of-tree (at least for now).

I think this is pretty simple to implement, but would vote for leaving this either out-of-tree for now or doing it in a separate pull request.

bradjc (approved):

ppannuto (approved):
Got a chance to actually read this over now, sorry I wasn't so engaged earlier today. I agree this is better than what we have, and we should make the stepwise improvement here.

PR #3246, opened 2022-09-27
cortex-m: Add initial dwt and dcb support

codingHahn:
### Pull Request Overview

The DWT and DCB peripherals are used to access various debug functionality in the chip, such as various counters.
This commit adds register descriptions for most of the DWT and some of the DCB. It also adds a very simple capsule that allows reading the cycle counter from userspace.

### Testing Strategy

This pull request was tested by getting the cycle count and it checks out.

### TODO or Help Wanted

This pull request still needs the rest of the registers of some peripherals. 

The capsule needs to be added to every supported board and could be expanded to grant access to the other capabilities of the peripherals. We also need to consider security implications for allowing debug peripherals in userspace.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/dcb.rs
+++ b/arch/cortex-m/src/dcb.rs
@@ -52,7 +52,7 @@ pub fn enable_debug_and_trace() {
 }
 
 /// Disable the Debug and Trace unit `DWT`
-pub unsafe fn disable_debug_and_trace() {
+pub fn disable_debug_and_trace() {
     DCB.demcr
         .modify(DebugExceptionAndMonitorControl::TRCENA::CLEAR);
 }
--- a/arch/cortex-m/src/dcb.rs
+++ b/arch/cortex-m/src/dcb.rs
@@ -1,6 +1,6 @@
-//! ARM Data Watchpoit and Trace Unit
+//! ARM Debug Control Block
 //!
-//! <https://developer.arm.com/documentation/100166/0001/Debug/Debug-configuration/Debug-register-summary?lang=en>
+//! <https://developer.arm.com/documentation/ddi0403/latest>
 
 use kernel::utilities::registers::interfaces::ReadWriteable;
 use kernel::utilities::registers::{register_bitfields, register_structs, ReadWrite};
--- a/arch/cortex-m/src/dcb.rs
+++ b/arch/cortex-m/src/dcb.rs
@@ -1,0 +1,58 @@
+//! ARM Data Watchpoit and Trace Unit
+//!
+//! <https://developer.arm.com/documentation/100166/0001/Debug/Debug-configuration/Debug-register-summary?lang=en>
+
+use kernel::utilities::registers::interfaces::ReadWriteable;
+use kernel::utilities::registers::{register_bitfields, register_structs, ReadWrite};
+use kernel::utilities::StaticRef;
+
+//TODO: Rest of the Registers
+register_structs! {
+    DcbRegisters{
+        (0x00 => demcr: ReadWrite<u32, DebugExceptionAndMonitorControl::Register>),
+
+        (0x04 => @END),
+    }
+}
+
+register_bitfields![u32,
+    DebugExceptionAndMonitorControl [
+        /// Write 1 to globally enable all DWT and ITM features.
+        /// RW.
+        TRCENA          OFFSET(24)  NUMBITS(1),
+
+        /// Debug monitor semaphore bit.
+        /// Monitor software defined.
+        /// RW.
+        MON_REQ         OFFSET(19)  NUMBITS(1),
+
+        /// Write 1 to make step request pending.
+        /// RW.
+        MON_STEP        OFFSET(18)  NUMBITS(1),
+/// Write 0 to clear the pending state of the DebugMonitor exception.
+        /// Writing 1 pends the exception.
+        /// RW.
+        MON_PEND        OFFSET(17)  NUMBITS(1),
+
+        /// Write 1 to enable DebugMonitor exception.
+        /// RW.
+        MON_EN        OFFSET(16)  NUMBITS(1),
+
+        //TODO: Rest
+    ],
+];
+
+const DCB: StaticRef<DcbRegisters> = unsafe { StaticRef::new(0xE000EDFC as *const DcbRegisters) };
+
+/// Enable the Debug and Trace unit `DWT`
+/// This has to be enabled before using any feature of the `DWT`
+pub fn enable_debug_and_trace() {
+    DCB.demcr
+        .modify(DebugExceptionAndMonitorControl::TRCENA::SET);
+}
+
+/// Disable the Debug and Trace unit `DWT`
+pub unsafe fn disable_debug_and_trace() {
+    DCB.demcr
+        .modify(DebugExceptionAndMonitorControl::TRCENA::CLEAR);
+}
--- a/arch/cortex-m/src/dcb.rs
+++ b/arch/cortex-m/src/dcb.rs
@@ -1,6 +1,7 @@
 //! ARM Debug Control Block
 //!
 //! <https://developer.arm.com/documentation/ddi0403/latest>
+//! Implementation matches `ARM DDI 0403E.e`
 
 use kernel::utilities::registers::interfaces::ReadWriteable;
 use kernel::utilities::registers::{register_bitfields, register_structs, ReadWrite};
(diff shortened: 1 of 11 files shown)


arch/cortex-m/src/dcb.rs
```diff
@@ -0,0 +1,58 @@
+//! ARM Data Watchpoit and Trace Unit
```
bradjc, line 1: update

arch/cortex-m/src/dwt.rs
```diff
@@ -0,0 +1,502 @@
...
+        /// RW.
+        FUNCTION    OFFSET(0)    NUMBITS(4),
+    ],
+
+];
+
+const DWT: StaticRef<DwtRegisters> = unsafe { StaticRef::new(0xE0001000 as *const DwtRegisters) };
+
+/// Enable the cycle counter
+/// [`DCB::enable_debug_and_trace()`] needs to be enabled for this to work
```
bradjc, line 479: Why not just do that then?

capsules/Cargo.toml
```diff
@@ -8,3 +8,4 @@ edition = "2021"
 kernel = { path = "../kernel" }
 enum_primitive = { path = "../libraries/enum_primitive" }
 tickv = { path = "../libraries/tickv" }
+cortexm = { path = "../arch/cortex-m" }
```
bradjc, line 11: We cannot make capsules dependent on something chip specific.

Instead, we have to use a HIL to abstract HW from capsules.

bradjc (changes requested):

capsules/Cargo.toml
codingHahn: Ok, I'll take a look and amend the pr

arch/cortex-m/src/dcb.rs
```diff
@@ -0,0 +1,58 @@
+//! ARM Debug Control Block
+//!
+//! <https://developer.arm.com/documentation/ddi0403/latest>
+
+use kernel::utilities::registers::interfaces::ReadWriteable;
+use kernel::utilities::registers::{register_bitfields, register_structs, ReadWrite};
+use kernel::utilities::StaticRef;
+
+//TODO: Rest of the Registers
```
ppannuto, line 9: image

There's only three more here; I think we'd ask for the register definition to be fully implemented before pulling this in (i.e., flesh out the rest of `register_structs` and `register bitfields`; don't need all the getters/setters/syscalls at first).

arch/cortex-m/src/dcb.rs
```diff
@@ -0,0 +1,58 @@
+//! ARM Debug Control Block
+//!
+//! <https://developer.arm.com/documentation/ddi0403/latest>
```
ppannuto, line 3:
```suggestion
//! <https://developer.arm.com/documentation/ddi0403/latest>
//! Implementation matches `ARM DDI 0403E.e`
```

Not that this doc actually changes in substantial ways that often, but it's useful to have a pointer to which version is actually implemented here.

arch/cortex-m/src/dcb.rs
```diff
@@ -0,0 +1,58 @@
...
+        (0x00 => demcr: ReadWrite<u32, DebugExceptionAndMonitorControl::Register>),
+
+        (0x04 => @END),
+    }
+}
+
+register_bitfields![u32,
+    DebugExceptionAndMonitorControl [
+        /// Write 1 to globally enable all DWT and ITM features.
+        /// RW.
```
ppannuto, line 21: This whole register is declared `ReadWrite` already; are the `RO/WO` fields? If not, these `RW`s should probably be removed.

arch/cortex-m/src/dcb.rs
```diff
@@ -0,0 +1,58 @@
...
+/// This has to be enabled before using any feature of the `DWT`
+pub fn enable_debug_and_trace() {
+    DCB.demcr
+        .modify(DebugExceptionAndMonitorControl::TRCENA::SET);
+}
+
+/// Disable the Debug and Trace unit `DWT`
+pub unsafe fn disable_debug_and_trace() {
+    DCB.demcr
+        .modify(DebugExceptionAndMonitorControl::TRCENA::CLEAR);
```
ppannuto, line 57: Consider wrapping DWT in a `PeripheralManager` instance (e.g. https://github.com/tock/tock/blob/7fcb766909b8093f3169cbb67e3b531b00a9b35a/chips/sam4l/src/i2c.rs#L569) to automate these calls and ensure enable-before-access.

ppannuto (changes requested):

kernel/src/hil/debug.rs
```diff
@@ -0,0 +1,15 @@
+/// Interface for interacting with debug hardware integrated in various SoCs.
+/// Currently allows reading the cycle counter and can be expanded to allow
+/// access to other features in the future.
```
bradjc, line 3:
```suggestion
//! Interface for interacting with debug hardware integrated in various SoCs.
//!
//! Currently allows reading the cycle counter.
```

kernel/src/hil/debug.rs
```diff
@@ -0,0 +1,15 @@
...
+use crate::ErrorCode;
+
+pub trait PerformanceCounters {
+    /// Enable the cycle counter. Returns an error, if the cycle counter is not present.
```
bradjc, line 7: Which error?

kernel/src/hil/debug.rs
```diff
@@ -0,0 +1,15 @@
+/// Interface for interacting with debug hardware integrated in various SoCs.
+/// Currently allows reading the cycle counter and can be expanded to allow
+/// access to other features in the future.
+use crate::ErrorCode;
+
+pub trait PerformanceCounters {
```
bradjc, line 6:
```suggestion
pub trait CycleCounter {
```
?

kernel/src/hil/debug.rs
```diff
@@ -0,0 +1,15 @@
+/// Interface for interacting with debug hardware integrated in various SoCs.
```
bradjc, line 1: "debug.rs" is somewhat generic, and there is the potential for confusion with `kernel/src/debug.rs`. Is there a name that would be somewhat more informative? Maybe `hw_debug`?

capsules/src/perf.rs
```diff
@@ -0,0 +1,51 @@
+//! Provides a performance counter interface for userspace.
```
bradjc, line 1: `perf.rs` is too short. If this is for cycle counts, let's name it that.

bradjc (changes requested):

arch/cortex-m/src/dwt.rs
```diff
@@ -0,0 +1,525 @@
...
+}
+
+impl Dwt {
+    pub const fn new() -> Self {
+        Self { registers: DWT }
+    }
+}
+
+impl hil::debug::PerformanceCounters for Dwt {
+    fn enable_cycle_counter() -> Result<(), ErrorCode> {
```
hudson-ayers, line 499: Personally I would call this `start_cycle_counter()` since it actually starts it counting up, rather than just enabling the functionality. Same for `disable_cycle_counter()` -- I would just call that `stop_cycle_counter()`, and not have it disable the DCB.

arch/cortex-m/src/dwt.rs
```diff
@@ -0,0 +1,525 @@
...
+    }
+}
+
+impl Dwt {
+    pub const fn new() -> Self {
+        Self { registers: DWT }
+    }
+}
+
+impl hil::debug::PerformanceCounters for Dwt {
```
hudson-ayers, line 498: I think it would also be helpful to add a `reset_cycle_counter()` method to this HIL to reset the counter to 0. This saves from having to get the current count and then perform subtraction, and it just an easy interface (imo).  Amit wrote a rough DWT implementation a few years ago that I often use when benchmarking, and it just does the following:

```rust
pub fn reset_timer() {
    DEMCR.set(DEMCR.get() | 0x01000000);
    DWT.cycnt.set(0); // reset the counter
    DWT.ctrl.set(0); // disable counter;
}
```

Keeping in mind that on something like the nrf52 this counter will overflow in 67 seconds, it can actually be pretty helpful to be able to reset to 0 rather than have to consider the (somewhat high) possibility of overflow.

arch/cortex-m/src/dwt.rs
```diff
@@ -0,0 +1,525 @@
...
+        if is_cycle_counter_present() {
+            // The cycle counter has to be enabled in the DCB block
+            dcb::enable_debug_and_trace();
+            DWT.ctrl.modify(Control::CYCNTENA::SET);
+            Ok(())
+        } else {
+            Err(ErrorCode::NOSUPPORT)
+        }
+    }
+
```
hudson-ayers, line 509: One other useful function from Amit's old DWT implementation was `bench()` which could be used to easily benchmark in-kernel code:

```rust
pub unsafe fn bench<F: FnOnce()>(f: F) -> u32 {
    reset_timer();
    start_timer();
    f();
    stop_timer();
    get_time()
}
```

I think it would be nice to see something like this included here.

hudson-ayers (dismissed):
I would love to see DWT support merged, I end up using some modified version of what Amit wrote for the PowerClocks paper all the time when taking performance measurements, and its always a pain to have to throw it somewhere in the kernel.

While I don't think this PR has to be updated to include this, I do think it would be really nice to see this functionality added to `kernel/src/debug.rs` (similar to `debug!()` and `debug_gpio!()`) so that it is easy to take performance measurements from anywhere. This would require some sort of global that boards would initialize with the appropriate implementation of the HIL, or perhaps we could configure it using `KernelResources`

hudson-ayers (commented):
@codingHahn will you have a chance to revisit this? If not, I will try to find time to update this with the various suggestions

codingHahn:
Yeah, sorry. I finished my bachelors thesis and had to settle in to working a full time job. I'd love to apply all your suggestions. Problem is that I don't have the board anymore because it belonged to my employer.

hudson-ayers:
Replaced by https://github.com/tock/tock/pull/3934

Issue #3229, opened 2022-09-19
Flash Controller <-> Capsule: Potential vulnerability. 


twilfredo:
I've run into an interesting scenario while writing the initial version of [these tests](https://github.com/tock/tock/blob/master/boards/opentitan/src/tests/flash_ctrl.rs). Where in the initial commit it would erase and write over pages containing code. So during extended testing you may run into an `illegal_instruction` caused by the code that was overwritten. 

I'm wondering since the flash driver technically allows to R/W/Er all pages (at least for OpenTitan), and capsules that use the flash controller are 'untrusted' in Tock. They could potentially request a flash op on a page that should not be operated on. 
Does this fundamentally create point of vulnerability? 

OpenTitan `flash_ctrl` does allow setting memory protection regions for flash (similar to `pmp`). This could be implemented to mitigate the above, but then what about devices that don't have this functionality?

bradjc:
Hopefully I can capture how we think about this in Tock:

Capsules are untrusted, and therefore they should not be able to arbitrarily write flash, which could have security implications. However, capsules can (and should and do) have the ability to write flash, but that ability must be provided by trusted code. So, the trust lies in the board author verifying the capsule is not poorly implemented before passing a flash controller to the capsule. That trust is explicit in main.rs as the board must explicitly pass a reference to something that can write flash to the capsule.

What we have to ensure is that a capsule cannot write flash without that ability being explicit in main.rs. We ensure this because there is no way for a capsule to create a flash controller object without 1) being able to use `unsafe` and/or 2) having a particular chip crate as a cargo dependency.

alistair23:
> So, the trust lies in the board author verifying the capsule is not poorly implemented before passing a flash controller to the capsule

So in this case any capsule that has a `hil::flash::Flash` interface can read/write/erase any flash address. I thought that was against the [threat model](https://github.com/tock/tock/blob/master/doc/threat_model/Capsule_Isolation.md), but apparently it's not. It's just up to the board writer to verify the capsule will act correctly.

Still, considering that the OT hardware has access controls, it might be worth adding a new HIL that the board maintainer can use to enforce access in hardware (not sure how though, I haven't looked that closely at the hardware).

twilfredo:
> > So, the trust lies in the board author verifying the capsule is not poorly implemented before passing a flash controller to the capsule
> 
[...]

PR #3250, opened 2022-09-29
kernel/hil/flash: add memory protection, impl/connect/test for OpenTitan

twilfredo:
### Pull Request Overview

This pull request adds a new `HIL` to interface flash memory protection (fmp). 

The hardware support for fmp may vary between vendors. For example, OpenTitan has ['advanced' control options](https://docs.opentitan.org/hw/ip/flash_ctrl/doc/#Reg_mp_region_cfg_0) but `nRF52840` only allows r/w control (see [section 6.3](https://infocenter.nordicsemi.com/pdf/nRF52840_PS_v1.0.pdf#%5B%7B%22num%22%3A153%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C85.039%2C123.823%2Cnull%5D)).

Since we don't want to limit the hardware functionality, this `HIL` adds a basic implementation which should work with hardware with basic functionality (i.e `nRF52840`) and an advanced `HIL` that extends the basic one (could be used by `OpenTitan` to allow full control of fmp.  An implementation example can be seen with the new additions to the OpenTitan flash controller in this PR. 

OpenTitan Specific Changes:

- Unit tests to test FMP functionality
- Driver implementation
- Add FMP rules to ROM/Kernel regions

Also, the addition of fmp to supported devices can address some of the concerns discussed in https://github.com/tock/tock/issues/3229

### Testing Strategy

Running the tests (new/old) on Verilator
```
�I00000 test_rom.c:81] Version:    earlgrey_silver_release_v5-5665-g217a0168b
Build Date: 2022-09-29, 14:35:47

I00001 test_rom.c:115] Test ROM complete, jumping to flash!
Unable to find otbn-rsa, disabling RSA support
OpenTitan initialisation complete. Entering main loop
[FLASH_CTRL] Test page read/write....
    [ok]
[FLASH_CTRL] Test page erase....
    [ok]
[FLASH_CTRL] Test memory protection basic....
    [ok]
[FLASH_CTRL] Test memory protection functionality....
    [ok]
trivial assertion... 
    [ok]
```

### TODO or Help Wanted

- RFC on HIL changes

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/src/main.rs
+++ b/boards/opentitan/src/main.rs
@@ -36,6 +36,7 @@ use kernel::platform::{KernelResources, SyscallDriverLookup, TbfHeaderFilterDefa
 use kernel::scheduler::priority::PrioritySched;
 use kernel::utilities::registers::interfaces::ReadWriteable;
 use kernel::{create_capability, debug, static_init};
+use lowrisc::flash_ctrl::FlashMPConfig;
 use rv32i::csr;
 
 pub mod io;
@@ -455,6 +456,32 @@ unsafe fn setup() -> (
         /// Beginning on the ROM region containing app images.
         static _sstorage: u8;
         static _estorage: u8;
+    }
+
+    // Flash setup memory protection for the ROM/Kernel
+    // Only allow reads for this region, any other ops will cause an MP fault
+    let mp_cfg = FlashMPConfig {
+        read_en: true,
+        write_en: false,
+        erase_en: false,
+        scramble_en: false,
+        ecc_en: false,
+        he_en: false,
+    };
+
+    // Allocate a flash protection region (associated cfg number: 0), for the code section.
+    if let Err(e) = peripherals.flash_ctrl.mp_set_region_perms(
+        &_manifest as *const u8 as usize,
+        &_etext as *const u8 as usize,
+        0,
+        &mp_cfg,
+    ) {
+        debug!("Failed to set flash memory protection: {:?}", e);
+    } else {
+        // Lock region 0, until next system reset.
+        if let Err(e) = peripherals.flash_ctrl.mp_lock_region_cfg(0) {
+            debug!("Failed to lock memory protection config: {:?}", e);
+        }
     }
 
     // Flash
--- a/chips/lowrisc/src/flash_ctrl.rs
+++ b/chips/lowrisc/src/flash_ctrl.rs
@@ -2,14 +2,13 @@
 
 use core::cell::Cell;
 use core::ops::{Index, IndexMut};
+use kernel::hil;
 use kernel::utilities::cells::OptionalCell;
 use kernel::utilities::cells::TakeCell;
 use kernel::utilities::registers::interfaces::{ReadWriteable, Readable, Writeable};
 use kernel::utilities::registers::{
     register_bitfields, register_structs, ReadOnly, ReadWrite, WriteOnly,
 };
-
-use kernel::hil;
 use kernel::utilities::StaticRef;
 use kernel::ErrorCode;
 
@@ -51,7 +50,7 @@ register_structs! {
         (0x16C => op_status: ReadWrite<u32, OP_STATUS::Register>),
         (0x170 => status: ReadOnly<u32, STATUS::Register>),
         (0x174 => debug_state: ReadOnly<u32>),
-        (0x178 => err_code: ReadOnly<u32>),
+        (0x178 => err_code: ReadWrite<u32, ERR_CODE::Register>),
         (0x17C => std_fault_status: ReadOnly<u32>),
         (0x180 => fault_status: ReadOnly<u32>),
         (0x184 => err_addr: ReadOnly<u32>),
@@ -110,6 +109,15 @@ register_bitfields![u32,
         INFO_SEL OFFSET(9) NUMBITS(2) [],
         NUM OFFSET(16) NUMBITS(12) []
     ],
+    ERR_CODE [
+        OP_ERR OFFSET(0) NUMBITS(1) [],
+        MP_ERR OFFSET(1) NUMBITS(1) [],
+        RD_ERR OFFSET(2) NUMBITS(1) [],
+        PROG_ERR OFFSET(3) NUMBITS(1) [],
+        PROG_WIN_ERR OFFSET(4) NUMBITS(1) [],
+        PROG_TYPE_ERR OFFSET(5) NUMBITS(1) [],
+        UPDATE_ERR OFFSET(6) NUMBITS(1) [],
+    ],
     PROG_TYPE_EN [
         NORMAL OFFSET(0) NUMBITS(1) [],
         REPAIR OFFSET(1) NUMBITS(1) [],
@@ -121,7 +129,12 @@ register_bitfields![u32,
         START OFFSET(0) NUMBITS(32) []
     ],
     REGION_CFG_REGWEN [
-        REGION OFFSET(0) NUMBITS(1) []
+        REGION OFFSET(0) NUMBITS(1) [
+            // Once locked, region cannot be modified till next reset
+            Locked = 0,
+            // Region can be configured.
+            Enabled = 1,
+        ]
     ],
     MP_REGION_CFG [
         // These config register fields require a special value of
@@ -157,7 +170,7 @@ register_bitfields![u32,
     ],
     MP_REGION [
         BASE OFFSET(0) NUMBITS(9) [],
-        SIZE OFFSET(9) NUMBITS(9) []
+        SIZE OFFSET(10) NUMBITS(10) []
     ],
     BANK_INFO_REGWEN [
         REGION OFFSET(0) NUMBITS(1) [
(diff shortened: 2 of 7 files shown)


bradjc:
I have a couple questions:

1. Do we need a HIL? Do we expect that other platforms will be able to implement this HIL?

2. How do we reason about this interface in the context of an MPU/PMP? There is overlapping functionality, do we want to expose the additional complexity in Tock?

twilfredo:
> I have a couple questions:
> 
>     1. Do we need a HIL? Do we expect that other platforms will be able to implement this HIL?
> 

The primary motive behind this was to add support for OpenTitan flash mp features, so technically we *could* just expose an api for the driver (exclusive to OT) to implement these features, without the need for this HIL (similar to the lowrisc OTBN). However, looking at other hardware (like Nordic socs, nRF52840 for example) they have similar functionality present. I was thinking that It would be good for Tock to implement these features for devices where possible.  Since Tock does not have a way to currently do this, adding a HIL to govern flash mp would be a starting point.

>     2. How do we reason about this interface in the context of an MPU/PMP? There is overlapping functionality, do we want to expose the additional complexity in Tock?

I'm not too familiar with the MPU/PMP usage in Tock,

twilfredo:
> I have a couple questions:
> 
>     1. Do we need a HIL? Do we expect that other platforms will be able to implement this HIL?
> 
The primary motive behind this was to add support for OpenTitan flash mp features, so technically we *could* just expose an api for the driver (exclusive to OT) to implement these features, without the need for this HIL (similar to the lowrisc OTBN). However, looking at other hardware (like Nordic socs, nRF52840 for example) they have similar functionality present. I was thinking that It would be good for Tock to implement these features for devices where possible.  Since Tock does not have a way to currently do this, adding a HIL to govern flash mp would be a starting point.

>     2. How do we reason about this interface in the context of an MPU/PMP? There is overlapping functionality, do we want to expose the additional complexity in Tock?

I'm not super familiar with the MPU/PMP usage in Tock. Do we currently utilize these for flash memory protection, perhaps not all MPUs might support this(?).  @alistair23 maybe you can chime in on this abit?

kernel/src/hil/flash.rs
```diff
@@ -142,3 +145,154 @@ pub trait Client<F: Flash> {
...
+    /// Region can be read.
+    pub read_en: bool,
+    /// Region can be programmed.
+    pub write_en: bool,
+}
+
+/// Memory access control protection for flash. For hardware that supports
+/// flash memory protection, this can be used to implement the relevant functionality.
+///
+/// Requires the Flash interface to have been implemented as well.
```
hudson-ayers, line 163: In the context of the possibility for additional future traits as alternatives to the current `Flash` trait, do we need to require this?

hudson-ayers (commented):
My feeling is that a HIL only makes sense when we know of multiple chips with support, so the basic HIL makes sense but the advanced does not.

kernel/src/hil/flash.rs
twilfredo: Wouldn't we want to make sure flash is implemented before flash memory protection is though?

twilfredo:
> My feeling is that a HIL only makes sense when we know of multiple chips with support, so the basic HIL makes sense but the advanced does not.

Yea fair enough, would it make more sense to expose the `advanced` features as a part of the OpenTitan flash driver (new api ?)

kernel/src/hil/flash.rs
hudson-ayers: I don't think it could really hurt for something to implement this trait without implementing flash, it just wouldn't be useful. And if https://github.com/tock/tock/pull/2993 is merged, it is very possible there could be something that implements `BlockStorage` but not `Flash` which still might benefit from flash memory protection

bradjc:
We discussed this in the OT WG, and the conclusion was this should be an API in the lowrisc chip for now. There are two questions that are hard to answer right now as a HIL:

1. Will this HIL API work for a reasonable subset of chips that have this functionality?
2. How does this API intersect with the MPU trait?

By making it an API we can learn more about how to answer those questions without having to address them now.

twilfredo:
> 
that sounds good, I'll work on converting this into a chip api.

twilfredo:
Updates:
- Rebased
- Reverted HIL changes, and replaced with a chip specific public api for interfacing to `flash_mp`... similar to the `otbn` the implementation.
- Kept the `FlashMPError` code in the flash HIL, as this generic and could be used by other drivers.

twilfredo:
Ping!

boards/opentitan/src/main.rs
```diff
@@ -220,6 +221,11 @@ impl KernelResources<earlgrey::chip::EarlGrey<'static, EarlGreyDefaultPeripheral
     }
 }
 
+#[inline(always)]
+fn div_ceil(a: u32, b: u32) -> u32 {
```
alistair23, line 225: It's worth adding a comment that we can remove this and use the Rust standard one at some point.

See: https://doc.rust-lang.org/std/primitive.i32.html#method.div_ceil and https://github.com/rust-lang/rust/issues/88581

boards/opentitan/src/main.rs
```diff
@@ -457,6 +463,36 @@ unsafe fn setup() -> (
...
+        scramble_en: false,
+        ecc_en: false,
+        he_en: false,
+    };
+
+    // Starting from page 0 (start of flash) to num_pages_occupied on `region 0`.
+    if let Err(e) =
+        peripherals
+            .flash_ctrl
+            .mp_set_region_perms(0, num_pages_occupied as usize, 0, &mp_cfg)
```
alistair23, line 486: I feel like this API should match the MPU/PMP API. So in `main()` we can just pass in the same addresses and handle the conversion to pages in the flash controller. Especially as this isn't a generic HIL but an OT specific API

chips/lowrisc/src/flash_ctrl.rs
```diff
@@ -564,6 +610,187 @@ impl<'a> FlashCtrl<'a> {
...
+
+    /// Setup the specified flash memory protection configuration
+    ///
+    /// # Arguments
+    ///
+    /// * `page_number` - Defines the starting page number to apply configuration to
+    /// * `num_pages` - Number of pages the configs are applied to (region size)
+    /// * `region_num`  - The configuration region number associated with this region.
+    ///                   This associates the specified permissions to a configuration region.
+    /// * `mp_perms` - Specifies the permissions to set
```
alistair23, line 624: We should document the return types here as well

alistair23 (changes requested):

boards/opentitan/src/main.rs
twilfredo: sounds good, I'm thinking this:
```
peripherals
            .flash_ctrl
            .mp_set_region_perms(start_addr, end_addr, region_num, &mp_cfg)
```
and in `pub fn mp_set_region_perms()`, we check if the addresses are valid then convert to pages with a helper function to calculate the num pages between `start_addr & end_addr`.

Does that sound okay?

alistair23: That sounds good!

The only question left is do we need region number?

twilfredo: cool! hmm... in this instance, we don't I suppose, since we are just setting up the single region. But later down the line could be useful if we wanted to have multiple isolated regions with varying levels of access control.

alistair23: Ok, might be best to provide a bit more info in the doc comments about what a region number is

twilfredo:
Updates:

- The api now uses the following convention as discussed above
```
    pub fn mp_set_region_perms(
        &self,
        start_addr: usize,
        end_addr: usize,
        region_num: usize,
        mp_perms: &FlashMPConfig,
    )
```

- Added more comments for how regions work
- Added comments for return conditions
- Updated tests to match the new api
- Updated `main.rs` to match new api, now just passing in the addresses similar to MPU/PMP

alistair23 (changes requested):

boards/opentitan/src/main.rs
```diff
@@ -457,6 +458,35 @@ unsafe fn setup() -> (
...
+        scramble_en: false,
+        ecc_en: false,
+        he_en: false,
+    };
+
+    // Starting from page 0 (start of flash) to num_pages_occupied on `region 0`.
+    if let Err(e) =
+        peripherals
+            .flash_ctrl
+            .mp_set_region_perms(0x0, text_end_addr as usize, 0, &mp_cfg)
```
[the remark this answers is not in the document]
twilfredo: as for `_manifest` which in our case would be at 20000000h (start of bank 0 page 0), the flash registers relatively refer to this as address 0. Similarly for `_etext` which would be `20000000h + x`. 
image


We could pass in the actual addresses as `_manifest` and `_etext` and just subtract the offset...? But then we would have to keep an eye of this, in-case the base address changes.

boards/opentitan/src/main.rs
```diff
@@ -457,6 +458,35 @@ unsafe fn setup() -> (
         static _estorage: u8;
     }
 
+    // Flash setup memory protection for the ROM/Kernel
+    let etext_addr = &_etext as *const u8 as u32;
+    let smanifest_addr = &_manifest as *const u8 as u32;
+    // Protect from ROM start to end of text
+    let text_end_addr = etext_addr.checked_sub(smanifest_addr).unwrap(); // Unwrap fail = addresses have updated in an unexpected way, this should be revised.
```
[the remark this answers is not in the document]
twilfredo: technically don't need this, I had it in there in-case the linker file in a weird way and this goes uncaught...but perhaps such a case wouldn't even be possible? where etext < smanifest ?

boards/opentitan/src/main.rs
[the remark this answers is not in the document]
twilfredo: I do think passing `_manifest and _etext` is alot more cohesive though, should we do this and offset it in the driver?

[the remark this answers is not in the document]
alistair23: Yeah, I think we should pass the address and offset them in the driver

boards/opentitan/src/main.rs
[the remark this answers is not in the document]
alistair23: You could add a linker assert to the board file, otherwise I think remove this. It adds a runtime overhead that isn't required

twilfredo:
Updates:
- Added a linker assert and removed the `unwrap()` code in `main.rs` for checking the address layout. 
- `mp_set_region_perms()` now takes the absolute addresses and offsets them in the driver.

alistair23 (dismissed):

kernel/src/hil/flash.rs
```diff
@@ -100,6 +100,9 @@ pub enum Error {
 
     /// An error occurred during the flash operation.
     FlashError,
+
+    /// A flash memory protection violation was detected
+    FlashMPError,
```
bradjc, line 105:
```suggestion
    /// A flash memory protection violation was detected.
    FlashMemoryProtectionError,
```

twilfredo: done!

twilfredo:
Ping!

bradjc (approved):

alistair23 (approved):

PR #3252, opened 2022-09-30
syscall: add notification mechanism

alexandruradovici:
### Pull Request Overview

This pull request adds a notification mechanisms that informs syscall drivers about the succesfull `subscribe` and `allow_readonly` and `allow_readwrite` system calls sent by applications.

As of Tock 2.1, drivers are not in charge of handling these system calls, but the kernel performs them in the driver's name. This makes drivers unaware of what happens with an application's upcalls and buffers. While most drivers do not need this information, some drivers might want to know when a an application swaps a buffer.

Applications are not allowed to access buffers while they are shared with a driver. In most cases, when an application swaps out a buffer (either unallows it, either provides a replacement for it) it means that the application is consuming the provided data. An example of such behavior is the `touch` driver. The driver receives touch events and places them into a readwrite allowed buffer and notifies the application using an upcall. When the application needs to consume the buffer, it has to swap it out with an empty one. This notification function provides the means to inform the driver that the application has swapped the buffer and that the driver has a new buffer in which it can place a new set of events.

Without this type of notification, the application had to issue a supplementary `command` system call to inform the driver about the consumption of the buffer.

I think this approach will be useful for communication drivers that needs fill in data as fast as possible.

### Testing Strategy

This pull request was tested on an STM32F412g Discovery board using the touch driver.

### TODO or Help Wanted

Feedback is welcome.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


kernel/src/syscall_driver.rs
```diff
@@ -312,4 +312,38 @@ pub trait SyscallDriver {
...
+    // Without this type of notification, the application had to issue a
+    // supplementary `command` system call to inform the driver about the consumption
+    // of the buffer.
+    fn syscall_notification(&self, process_id: ProcessId, which: SycallNotification) {}
+}
+
+/// The system call request type used to inform a driver upon
+/// a system call that is performed by the kernel in the
+/// driver's name
+pub enum SycallNotification {
```
teonaseverin, line 345:
```suggestion
pub enum SyscallNotification {
```

lschuermann:
I'm sympathetic towards this addition, as it addresses a disparity we introduced in the semantics around buffer sharing in Tock 2.1: capsules can no longer control whether allow operations are granted, and hence have to accept any changes in shared buffers by an application; at the same time they cannot efficiently check whether an allowed buffer was changed.

It is noteworthy that in current implementations, userspace can still technically modify the contents of allowed buffers, but this is disallowed for `allow_readonly` and `allow_readwrite` by TRD 104. Hence I'd argue that this change is in line with the expectations set by TRD 104, although we'd need to clearly communicate that this mechanism is not sufficient to be informed of all permutations of the underlying buffer.

I think it's conceivable to use other mechanisms to tell whether a buffer was changed, such as an incrementing counter for allow operations per buffer, in case the proposed method of a method invocation on the driver is deemed to expensive.

bradjc:
I'm convinced we need something like this. However, we've only identified a very specific use case (i.e. reasonably fast streaming data) where this is needed. Yet, adding a new callback to the `SyscallDriver` trait is a very blunt tool, in that it could be (even if it should not be) used to implement a wide range of functionality. This then burdens code review to ensure the callback is used only for known purposes.

My question is, can we come up with some API that makes it very clear what the callback is for? Can we make it so that the callback can only modify the grant? Can we pass a reference somehow and have the kernel internally mark something in the grant (and therefore not add the generic callback)?

hudson-ayers:
I will point out this adds 2kB to Imix and 1.5 kB to OT, neither of which have any drivers that use the mechanism

alexandruradovici:
> My question is, can we come up with some API that makes it very clear what the callback is for? Can we make it so that the callback can only modify the grant? Can we pass a reference somehow and have the kernel internally mark something in the grant (and therefore not add the generic callback)?

#3258 does that

phil-levis:
Why is there a `subscribe` notification?

I am conceptually 100% in favor of this, but the code size increase concerns me. Where is that code coming from, and is there a way to remove it?

alexandruradovici:
> Why is there a `subscribe` notification?
> 
I added this as is became technically possible. Considering @bradjc and @phil-levis concerns about making code review more difficult and adding the ability for capsules to start actions within this notification, I am not against removing it.

alexandruradovici:
> I am conceptually 100% in favor of this, but the code size increase concerns me. Where is that code coming from, and is there a way to remove it?
These lines seems to add about 700B to the code.

https://github.com/OxidosAutomotive/tock/blob/339ebed78f736f318bc286c6ab46664f4100527b/kernel/src/kernel.rs#L1298-L1310

alexandruradovici:
Using the code changes described in #3276, the code size is now 800 bytes less (imix) then the current master. 

The notification seems to increase the code size (if #3276 is merged) by only 160 bytes (imix).

hudson-ayers:
When we pick this up we should be sure to compare against https://github.com/tock/tock/pull/3258

lschuermann:
If we add a notification about process state changes (start, stop, restart), this can also re-enable optimizations in drivers such as the `num_armed` optimization in alarm: https://github.com/tock/tock/pull/3975#issuecomment-2136125626

Currently drivers have no reliable way of determining whether a process is still alive, and can't perform actions in response to these state changes. These issues also affect other drivers, such as ones that enable a peripheral in response to a userspace request, but won't disable it when the application stops / faults.

lschuermann:
Closing in favor of #4023.

PR #3253, opened 2022-09-30
Sifive uart abort

mateibarbu19:
### Pull Request Overview

This pull request adds UART abort functionality to all boards which feature a SiFive chip and finishing touches to the last UART changes brought to these chips.

### Testing Strategy

This pull request was tested by running the [console_timeout](https://github.com/tock/libtock-c/tree/master/examples/tests/console_timeout) app on a HiFive Inventor.

I would also like to note that Tock is running fine with these changes on the `arty_e21` board, however I didn't test with apps.

### Help Wanted

If anyone is willing to see if they can test this new functionality on a `hifive1` or a `redboard_redv`. I tried to do it myself using QEMU, but ran in a not enough memory error. Thanks.

### Documentation Updated

- [x] None.

### Formatting

- [x] Ran `make prepush`.


chips/sifive/src/uart.rs
```diff
@@ -91,15 +93,21 @@ pub struct Uart<'a> {
...
     pub baud_rate: u32,
 }
 
-impl<'a> Uart<'a> {
-    pub fn new(base: StaticRef<UartRegisters>, clock_frequency: u32) -> Uart<'a> {
+impl<'a, T> Uart<'a, T> {
+    pub fn new(
+        base: StaticRef<UartRegisters>,
+        clock_frequency: u32,
+        deferred_call: &'a DeferredCall<T>,
```
alistair23, line 109: You should be able to use a `DynamicDeferredCall` here instead. Have a look at other examples in the `chips` directory.

boards/redboard_redv/src/main.rs
```diff
@@ -247,10 +268,7 @@ pub unsafe fn main() {
     )
     .finalize(());
 
-    // Need two debug!() calls to actually test with QEMU. QEMU seems to have
-    // a much larger UART TX buffer (or it transmits faster).
-    debug!("Red-V initialization complete.");
-    debug!("Entering main loop.");
```
alistair23, line 253: Why not leave this as is?

alistair23 (changes requested):

boards/redboard_redv/src/main.rs
mateibarbu19: Because I couldn't reproduce the bug and have tried many times to do so.

chips/sifive/src/uart.rs
hudson-ayers: `DynamicDeferredCall` has notable size overhead relative to this implementation. I suspect switching to it would cost ~250 bytes on boards with two UARTs

chips/sifive/src/uart.rs
```diff
@@ -264,6 +274,34 @@ impl<'a> Uart<'a> {
...
+            self.tx_client.map(|client| {
+                self.tx_buffer.take().map(|buf| {
+                    client.transmitted_buffer(buf, self.tx_position.get(), Err(ErrorCode::CANCEL));
+                });
+            });
+            self.tx_status.set(UARTStateTX::Idle);
+        }
+
+        if self.rx_status.get() == UARTStateRX::AbortRequested {
+            // self.transmit_sync(&[65, 66, 67, 68]);
```
hudson-ayers, line 289: delete?

hudson-ayers (dismissed):

boards/redboard_redv/src/main.rs
alistair23: Is there a reason not to just leave it though?

chips/sifive/src/uart.rs
alistair23: Ah ok

boards/redboard_redv/src/main.rs
hudson-ayers: IIRC, the bug was not "without two lines nothing prints" but rather: "with a single debug!() statement, things will seem to work in QEMU even if there is a breakage". Basically, because QEMU has a large buffer / transmits quickly, the entire message is printed to console even if the kernel loop does not run. So without a two line message here, our QEMU CI will fail to catch basic breakages. Now, we don't even run the redboard_redv in QEMU as part of CI, so it may not matter much here, but I seem to recall that being the justification for this when it was first added for the hifive1.

mateibarbu19: I understand your point of view. I will leave it as it is and add it back to the HiFive1 with a more in depth explanation.

chips/sifive/src/uart.rs
mateibarbu19: Sure, I forgot it indeed.

boards/arty_e21/src/main.rs
```diff
@@ -1,4 +1,4 @@
-//! Board file for the SiFive E21 Bitstream running on the Arty FPGA
+//! Board file for the iFive E21 Bitstream running on the Arty FPGA
```
bradjc, line 1: ?

bradjc (approved):

hudson-ayers (approved):

bradjc (approved):
I don't like how the e310x chip code is designed, as updating uart to support abort shouldn't also change how boards are setup (i.e. switching `e310x` to `e310_g002`). The e310* crates should be updated so that there is only one way to setup a board (i.e. boards should have to depend on the specific chip variant crate, and not the generic shared crate).

mateibarbu19:
> I don't like how the e310x chip code is designed, as updating uart to support abort shouldn't also change how boards are setup (i.e. switching `e310x` to `e310_g002`). The e310* crates should be updated so that there is only one way to setup a board (i.e. boards should have to depend on the specific chip variant crate, and not the generic shared crate).

I'm not sure what you mean by this, as far as I can see boards depend on the specific crate, not on the generic shared one. Am I missing something? Can you please be a little bit more specific?

bradjc:
> I'm not sure what you mean by this, as far as I can see boards depend on the specific crate, not on the generic shared one. Am I missing something? Can you please be a little bit more specific?


This PR is about adding UART abort support, yet it also changes the dependency of the boards, which is really a separate change. But it highlights an issue: it shouldn't be possible to create a sifive-based board in two different ways. Ideally PRs like this shouldn't have to change the board dependencies, and we want to guide future board authors into the canonical method without having to wait for a PR review.

mateibarbu19:
> > I'm not sure what you mean by this, as far as I can see boards depend on the specific crate, not on the generic shared one. Am I missing something? Can you please be a little bit more specific?
> 
> This PR is about adding UART abort support, yet it also changes the dependency of the boards, which is really a separate change. But it highlights an issue: it shouldn't be possible to create a sifive-based board in two different ways. Ideally PRs like this shouldn't have to change the board dependencies, and we want to guide future board authors into the canonical method without having to wait for a PR review.

I have a better understanding of what you mean, but the reason some dependencies changed in this PR is because of a mistake of mine. The boards `redboard_redv` and `arty_e21` should have been updated along with the `sifive` and `e310x` changes I brought before this PR.

One reason specific board instantiation had to change was because the number of UART channels should not be limited by hard-coding the number of deferred calls in the `sifive` crate. Thus each chip instantiates how many deferred calls it needs, each one linked to a UART channel. 

Is there anything I need to change to improve the situation?

boards/hifive1/src/main.rs
```diff
@@ -187,10 +187,6 @@ pub unsafe fn main() {
         &peripherals.e310x.gpio_port[17],
         &peripherals.e310x.gpio_port[16],
     );
-    peripherals.e310x.uart1.initialize_gpio_pins(
-        &peripherals.e310x.gpio_port[18],
-        &peripherals.e310x.gpio_port[23],
-    );
```
alistair23, line 193: Why remove this?

mateibarbu19: Because when I placed it there, the clearing of pending interrupts didn't work. After the [patch](https://github.com/tock/tock/commit/2ce997d5680b9cdcd47c2ec70d25901132df2a40) there was no need for it.

alistair23 (approved):

bradjc:
> Is there anything I need to change to improve the situation?

Ideally the e310x crate wouldn't expose default peripherals so boards would be forced to use the specific variant crate.

PR #3255, opened 2022-09-30
Update components to use _static and static_buf!(): straightforward components

bradjc:
### Pull Request Overview

This pull request updates components to the new format, and includes the changes that seemed fairly straightforward (i.e. mostly just changing static_init_half -> .write() and making sure all memory is statically defined in macros.

I can treat this like a "consent agenda" and move commits to their own PR as necessary (each component change is its own commit).

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/acd52832/src/main.rs
+++ b/boards/acd52832/src/main.rs
@@ -232,7 +232,7 @@ pub unsafe fn main() {
     //
     // LEDs
     //
-    let led = components::led::LedsComponent::new().finalize(components::led_component_helper!(
+    let led = components::led::LedsComponent::new().finalize(components::led_component_static!(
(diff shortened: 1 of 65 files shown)


hudson-ayers (approved):
I think this is a major step in the right direction. I looked through all of the changes in `components/` as well as to main.rs for Imix and did not catch any errors.

ppannuto (approved):
I am grateful to hudson for looking at all of these – I looked at an arbitrary sample via random scroll, and they all looked good as well. I think a monolithic PR is fine here.

PR #3258, opened 2022-10-02
Add attributes to shared buffers

alexandruradovici:
### Pull Request Overview

This pull request adds attributes to shared process buffers. It uses part of the `len` bits to encode them. This is a different approach to the issue described in https://github.com/tock/tock/pull/3252. It solves the problem of informing capsules about buffer swaps without giving them the possibility to start actions when this happens.

The pull request defines one single attribute, the `accessed` attribute. This single bit attribute states whether the capsule has tried to access the buffer since it has been shared. When an application shares a buffer, the `accessed` attribute is cleared. Whenever the capsule retrieves the buffer using `get_readonly_processbuffer` or `get_readwrite_processbuffer`, this attribute is set. All the buffers returned by subsequent calls to these functions will have this attribute set.

#### Restrictions
The buffer attributes are saved within the `len` field in the `SavedAllowRo` and `SavedAllowRw` structures. This implies that the maximum length of a shared buffer is reduced to `2^(size_of::<usize>()-1)`. Considering that Tock runs on MCUs, this should not be an issue.

#### The issue

This is the issue that led to this pull reqeuest.

As of Tock 2.1, drivers are not in charge of handling these system calls, but the kernel performs them in the driver's name. This makes drivers unaware of what happens with an application's upcalls and buffers. While most drivers do not need this information, some drivers might want to know when a an application swaps a buffer.

Applications are not allowed to access buffers while they are shared with a driver. In most cases, when an application swaps out a buffer (either unallows it, either provides a replacement for it) it means that the application is consuming the provided data. An example of such behavior is the `touch` driver. The driver receives touch events and places them into a readwrite allowed buffer and notifies the application using an upcall. When the application needs to consume the buffer, it has to swap it out with an empty one. The `accessed` attribute provides the means to inform the driver that the application has swapped the buffer and that the driver has a new buffer in which it can place a new set of events.

Without this attribute, the application had to issue a supplementary `command` system call to inform the driver about the consumption of the buffer. This leads to a data race, as there capsule might execute in between the application's `allow` and `command`.

### Testing Strategy

This pull request was not tested yet.

### TODO or Help Wanted

As this touches core grant architecture, a lot of feedback is welcome.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/touch.rs
+++ b/capsules/src/touch.rs
@@ -35,9 +35,9 @@ mod rw_allow {
     // | id (u8) | type (u8) | x (u16)      | y (u16)      | size (u8) | pressure (u8) |         ...
     // +---------+-----------+--------------+--------------+-----------+---------------+-------- ...
     // | Touch 0                                                                       | Touch 1 ...
-    pub const EVENTS: usize = 2;
+    pub const EVENTS: usize = 0;
     /// The number of allow buffers the kernel stores for this grant
-    pub const COUNT: u8 = 3;
+    pub const COUNT: u8 = 1;
 }
 
 fn touch_status_to_number(status: &TouchStatus) -> usize {
@@ -50,7 +50,6 @@ fn touch_status_to_number(status: &TouchStatus) -> usize {
 }
 
 pub struct App {
-    ack: bool,
     dropped_events: usize,
     x: u16,
     y: u16,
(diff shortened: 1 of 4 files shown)


alexandruradovici:
During the Tock call we discussed that one downside, besides the code size, of #3252 is that capsules could use the callback to start actions, which would contradict the fact that actions should be started only as a consequence of `command` calls.

alexandruradovici:
For imix the code size increase is 432 bytes.

lschuermann:
Closing in favor of #4023.

PR #3260, opened 2022-10-02
component: update udp to new static

bradjc:
### Pull Request Overview

This pull request updates the UDP component to the new standard format.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/udp_driver.rs
+++ b/boards/components/src/udp_driver.rs
@@ -1,7 +1,7 @@
 //! Component to initialize the userland UDP driver.
 //!
-//! This provides one Component, UDPDriverComponent. This component initializes a userspace
-//! UDP driver that allows apps to use the UDP stack.
+//! This provides one Component, UDPDriverComponent. This component initializes
+//! a userspace UDP driver that allows apps to use the UDP stack.
 //!
 //! Usage
 //! -----
@@ -14,7 +14,7 @@
 //!        local_ip_ifaces,
 //!        PAYLOAD_LEN,
 //!     )
-//!     .finalize();
+//!     .finalize(components::udp_driver_component_static!());
 //! ```
 
 use capsules;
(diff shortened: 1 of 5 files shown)


hudson-ayers (approved):
This is a good example where the "static!() macros return a tuple" convention starts to get a little dicey, but I still think this is a step in the right direction. For macros that need to initialize 15 static objects (which is partially just because the network stack has some problematic APIs, like taking a reference to otherwise zero-size capabilities), we might want to consider creating named structs just for all of the types returned by the macro. This would hep prevent bugs arising when elements are added/removed from the tuple of static objects

bradjc:
> This would hep prevent bugs arising when elements are added/removed from the tuple of static objects

Having looked at every component, this isn't really a problem. Because the compiler type checks the "return" of the macro against the type of `StaticInput`, any mismatches are detected. Now, the error itself is somewhat misleading, as the most common way this manifests is that the compiler complains that return value of `finalize()` does not match the stated _output_ type because it implicitly assumes the input static type to the `MaybeUninit.write()` call was correct.

Now, what issues could arise? Well, it is easy to mixup two buffers (e.g. txbuf, rxbuf) that have the same length (i.e. "returning" `(txbuf, rxbuf)` from the macro and then declaring StaticInput in the other order. This is more a problem for code readability than it is for the system, as the name of the buffers doesn't _really_ matter.

By _far_ the most frustrating issue I had was I would make a mistake like this in the macro:

```rust
#[macro_export]
macro_rules! ccs811_component_static {
    () => {{
        let i2c_device = kernel::static_buf!(capsules::virtual_i2c::I2CDevice);
        let buffer: kernel::static_buf!([u8; 6]);
        let ccs811 = kernel::static_buf!(capsules::ccs811::Ccs811<'static>);

        (i2c_device, buffer, ccs811)
    };};
```

and the compiler would give absolutely useless errors. After spending 20 minutes trying to find what was wrong, I'm pretty convinced the last thing we should do is convert components entirely to macros. We actually get quite a bit of type checking right now.

hudson-ayers:
The bug in that snippet is the extra semicolon, I assume? But yeah that is really frustrating. And we would get basically the same amount of type checking even if components were all-macro, but I agree that because the errors are hidden within the macro they become much harder to debug (unless https://github.com/rust-lang/rust/issues/82068 ever becomes stable + default).

bradjc:
> The bug in that snippet is the extra semicolon, I assume?

No. Well that makes me feel better at least, it is difficult to see because in some contexts the error it is completely valid syntax.

hudson-ayers:
Oh, wow yeah I totally missed the colon instead of the `=` even after looking at it a few times

ppannuto (approved):
that giant list is indeed hard to read, but not too many things have this problem I don't think, so probably okay until proven a bigger problem

PR #3231, opened 2022-09-20
Component: Move ADC components to new _static macro

bradjc:
### Pull Request Overview

This pull request updates the ADC components to use the _static component macro convention and static_buf!().

### Testing Strategy

travis

### TODO or Help Wanted

Need temperature PR to go through first.

### Documentation Updated

- [x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x] Ran `make prepush`.

PR #3267, opened 2022-10-05
remove ADC `HighSpeedClient` trait

hudson-ayers:
### Pull Request Overview

This pull request removes the ADC `HighSpeedClient` trait and moves the callback it contained into the basic adc `Client` trait. The reasoning behind this is that having two separate Client traits is not compatible with the use of generic components. 

Today, all chips which support high speed sampling do so by creating an "EverythingClient" trait which is a supertrait of `Client`+`HighSpeedClient`. Each chip ADC then creates a normal, non-trait method `set_client(client: &'static dyn EverythingClient){...}`. However, there is no way for a component to know that such a method will exist if it is not part of any ADC trait. Currently, there is only a component for `AdcVirtualized`, which does not support high speed sampling. Creating a component for `AdcDedicated`, which does support high speed sampling, is impossible without this PR. Attempting to create one without this PR (as @bradjc did in https://github.com/tock/tock/pull/3231/commits/744ed25fdce58a07863c5385fbba5d8cf3c14125) would lead you into a very very subtle bug, where previously working code would not work once copied into the component, because within the component `adc_peripheral.set_client(client)` resolves to the `set_client()` method on the Adc trait, instead of the `set_client()` that accepts an `EverythingClient`. 

With this change, implementations that do not support high speed sampling should simply leave this callback unimplemented.

### Testing Strategy

This pull request has not been tested.

### TODO or Help Wanted

N/A

### Documentation Updated

- [x] Updated TRD102

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/imix/src/imix_components/adc.rs
+++ b/boards/imix/src/imix_components/adc.rs
@@ -19,6 +19,7 @@ use capsules::adc;
 use kernel::capabilities;
 use kernel::component::Component;
 use kernel::create_capability;
+use kernel::hil::adc::Adc;
 use kernel::static_init;
 use sam4l::adc::Channel;
 
--- a/capsules/src/adc.rs
+++ b/capsules/src/adc.rs
@@ -767,10 +767,8 @@ impl<A: hil::adc::Adc + hil::adc::AdcHighSpeed> hil::adc::Client for AdcDedicate
             let _ = self.adc.stop_sampling();
         }
     }
-}
 
-/// Callbacks from the High Speed ADC driver
-impl<A: hil::adc::Adc + hil::adc::AdcHighSpeed> hil::adc::HighSpeedClient for AdcDedicated<'_, A> {
+    /// Callback from the High Speed ADC driver
     /// Internal buffer has filled from a buffered sampling operation.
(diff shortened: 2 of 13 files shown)


bradjc:
How does this compare to adding a `set_client()` fn to `HighSpeed`, and then having multiple components? The critical error in the ADC is that `set_client()` is not in the trait.

phil-levis:
High speed sampling is pretty rare; making all clients handle it seems like a mistake. Instead, having separate clients, as @bradjc suggests, seems like a better approach. The cost is one extra reference in the ADC. There is a larger design issue with the conflicts between the two ADC trait types (they can trigger BUSY on each other), but fixing that requires a significant revision of the APIs.

phil-levis (changes requested):
High speed sampling is pretty rare; making all clients handle it seems like a mistake. Instead, having separate clients, as @bradjc suggests, seems like a better approach. The cost is one extra reference in the ADC. There is a larger design issue with the conflicts between the two ADC trait types (they can trigger BUSY on each other), but fixing that requires a significant revision of the APIs.

bradjc:
Merged #3290 instead.

PR #3249, opened 2022-09-29
boards/opentitan: bump to latest master RTL SHA

twilfredo:
### Pull Request Overview

This patch series adds support for OpenTitan [latest master](https://github.com/lowRISC/opentitan/commit/7550a3512290c3c4df978fcc70112b787c963076) (**as of this PR**).  Most of the heavy lifting for this was done in https://github.com/tock/tock/pull/3056.

This PR updates the since changes, registers/irqs etc... to bring Tock up to speed.

Going forward: Latest master of OT will most likely work, I'll test any major  changes (maybe weekly) and update here as necessary. 

### Testing Strategy

Running OpenTitan tests on Verilator (with the latest master build of OT).

```
I00000 test_rom.c:98] Version: earlgrey_silver_release_v5-7719-g7550a3512, Build Date: 2022-09-29 11:24:47
I00001 test_rom.c:180] Test ROM complete, jumping to flash!
Unable to find otbn-rsa, disabling RSA support
OpenTitan initialisation complete. Entering main loop
check run AES128 CCM... 
AES CCM* encryption/decryption tests
aes_ccm_test passed: (current_test=0, encrypting=true, tag_is_valid=true)
aes_ccm_test passed: (current_test=0, encrypting=false, tag_is_valid=true)
aes_ccm_test passed: (current_test=1, encrypting=true, tag_is_valid=true)
aes_ccm_test passed: (current_test=1, encrypting=false, tag_is_valid=true)
aes_ccm_test passed: (current_test=2, encrypting=true, tag_is_valid=true)
aes_ccm_test passed: (current_test=2, encrypting=false, tag_is_valid=true)
    [ok]
check run AES128 ECB... 
aes_test passed (ECB Enc Src/Dst)
aes_test passed (ECB Dec Src/Dst)
aes_test passed (ECB Enc In-place)
aes_test passed (ECB Dec In-place)
    [ok]
check run AES128 CBC... 
aes_test passed (CBC Enc Src/Dst)
aes_test passed (CBC Dec Src/Dst)
aes_test passed (CBC Enc In-place)
aes_test passed (CBC Dec In-place)
    [ok]
check run AES128 CTR... 
aes_test CTR passed: (CTR Enc Ctr Src/Dst)
aes_test CTR passed: (CTR Dec Ctr Src/Dst)
    [ok]
check run CSRNG Entropy 32... 
Entropy32 test: first get Ok(())
Entropy test: obtained all 8 values. They are:
[00]: 735b27a0
[01]: 497b246f
[...]

PR #3271, opened 2022-10-07
boards/opentitan/Makefile: add OT version check

twilfredo:
### Pull Request Overview

For running tests on HW/Verilator make sure we are testing only on a supported version of OpenTitan by default. This might be a little annoying to keep changing if we are jumping back and forth between versions (normally shouldn't need to do that though). But can save some time when developing on different versions of OT by ensuring the current tests are ran on the respective version (specially Verilator, where tests take hours only to fail on the last one because of a load fault 😪).  

### Testing Strategy

Trying to run tests when on an unsupported commit of OpenTitan. 

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/README.md
+++ b/boards/opentitan/README.md
@@ -18,11 +18,13 @@ for more details.
 
 Programming
 -----------
+The latest supported version (commit SHA) of OpenTitan is specified in the [OPENTITAN_SUPPORTED_SHA](https://github.com/tock/tock/blob/master/boards/opentitan/earlgrey-cw310/Makefile) make variable found:
+>  boards/opentitan/earlgrey-cw310/Makefile
 
-Tock on OpenTitan requires
-lowRISC/opentitan@d072ac505f82152678d6e04be95c72b728a347b8. In
-general it is recommended that users start with the specified commit as newer
-versions have not been tested.
+In *general* it is recommended that users start with the commit specified by `OPENTITAN_SUPPORTED_SHA` as newer
+versions **have not been** tested.
+
+> Note: when building, you can pass in `SKIP_OT_VERSION_CHECK=yes` to skip the trivial OpenTitan version check, this maybe useful when developing or testing across multiple versions of OpenTitan.
 
 Unfortunately the OpenTitan documentation is out of sync with the Tock setup.
 For instructions that match the OpenTitan version Tock supports you will need
--- a/boards/opentitan/earlgrey-cw310/Makefile
+++ b/boards/opentitan/earlgrey-cw310/Makefile
@@ -10,6 +10,9 @@ QEMU ?= ../../../tools/qemu/build/qemu-system-riscv32
 # This offset is calculated (from the linker file) as:
 # ORIGIN(rom) + size_manifest = 0x20000000 + 0x400
 QEMU_ENTRY_POINT=0x20000400
+# Latest supported commmit of OpenTitan
+OPENTITAN_SUPPORTED_SHA := d072ac505f82152678d6e04be95c72b728a347b8
+
 
 include ../../Makefile.common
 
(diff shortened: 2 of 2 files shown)


boards/opentitan/earlgrey-cw310/Makefile
```diff
@@ -6,6 +6,8 @@ PLATFORM=earlgrey-cw310
 FLASHID=--dev-id="0403:6010"
 RISC_PREFIX ?= riscv64-elf
 QEMU ?= ../../../tools/qemu/build/qemu-system-riscv32
+# Latest supported commmit of OpenTitan
+OPENTITAN_SUPPORTED_SHA := 217a0168ba118503c166a9587819e3811eeb0c0c
```
twilfredo, line 10: As per: https://github.com/tock/tock/blob/13bbb77161e365a5bf8c6c41ecc34d4489854f2a/boards/opentitan/README.md?plain=1#L23

alistair23: It's a pain having two copies of this SHA to update. Maybe we could change the README to point to this?

twilfredo: Good idea! Updated the `README` to point here.

boards/opentitan/README.md
```diff
@@ -18,11 +18,11 @@ for more details.
...
 -----------
+The latest supported version (commit SHA) of OpenTitan is specified in the [OPENTITAN_SUPPORTED_SHA](https://github.com/tock/tock/blob/master/boards/opentitan/earlgrey-cw310/Makefile) make variable found:
+>  boards/opentitan/earlgrey-cw310/Makefile
 
-Tock on OpenTitan requires
-lowRISC/opentitan@217a0168ba118503c166a9587819e3811eeb0c0c. In
-general it is recommended that users start with the specified commit as newer
-versions have not been tested.
+In *general* it is recommended that users start with the specified commit as newer
+versions **have not been** tested.
```
alistair23, line 25: This should say that `OPENTITAN_SUPPORTED_SHA` specifies this commit

twilfredo: Fixed!

alistair23 (approved):

twilfredo:
> Is this check in the right spot? It looks like someone could 1) build OT on the incorrect commit, 2) switch the OT repo to the correct commit, 3) run the test, and this check would succeed but the hardware would be out-of-sync.

Yea that's definitely still doable with this, I was trying to avoid using `bazel` here to get a build version (not sure if this is even possible - seems like it ?). Is there a simpler solution to this, without over complicating the Makefile?

twilfredo:
Changes:
- Rebased / Fixup conflict

alistair23 (dismissed):

jrvanwhy:
I don't know if there's a way to do better.

Can you add a way to skip the version check? Maybe you can skip it if an environment variable `SKIP_OT_VERSION_CHECK` is set? Otherwise, we'll have to comment this out every time we are working on an OT version update.

twilfredo:
Updates:
- Added `SKIP_OT_VERSION_CHECK` make variable to skip this trivial check @jrvanwhy 
- Update `README` to add usage for this skip.
- TLDR: pass in `SKIP_OT_VERSION_CHECK=yes` when building to skip the checks.

hudson-ayers (approved):

lschuermann (dismissed):

twilfredo:
Updates:

- Resolved conflicts in README created by https://github.com/tock/tock/pull/3249

alistair23 (approved):

hudson-ayers (approved):

bradjc (approved):

PR #3276, opened 2022-10-07
Decrease kernel size and return NODEVICE error

alexandruradovici:
### Pull Request Overview

This pull request changes the kernel code to decrease its size by around 900 bytes on arm and 400 bytes on risc-v. 

The is achieved by moving all the system calls that request a driver number in one single `with_driver` closure. In this way, the kernel searches only in one place for a driver.

| Board | Master | PR |
|-------|--------|---|
| imix  | 169676 | 168700 |
| stm32f412 | 109216 | 108400 |
| hivive1 | 79408 | 79024 |

This PR also fixes the error returned by by the `subscribe` and `allow` system calls if the `driver_number` is not found. before this change, the kernel was returning `INVAL` instead of `NODEVICE`. The kernel did not check the existence of the driver, it was just failing to allocate the driver's grant.

### Testing Strategy

This pull request was not tested yet.

### TODO or Help Wanted

Feedback

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/kernel.rs
+++ b/kernel/src/kernel.rs
@@ -28,6 +28,7 @@ use crate::process::{self, Process, ProcessId, ShortID, Task};
 use crate::process_checker::{self, CredentialsCheckingPolicy};
 use crate::process_loading::ProcessLoadError;
 use crate::scheduler::{Scheduler, SchedulingDecision};
+use crate::syscall::SyscallDriver;
 use crate::syscall::{ContextSwitchReason, SyscallReturn};
 use crate::syscall::{Syscall, YieldCall};
 use crate::syscall_driver::CommandReturn;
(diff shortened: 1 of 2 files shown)


bradjc:
No comment on this PR, just marking as blocked until appid is in.

kernel/src/kernel.rs
```diff
@@ -873,188 +865,206 @@ impl Kernel {
...
+                completion_code,
+            } => match which {
+                // The process called the `exit-terminate` system call.
+                0 => process.terminate(Some(completion_code as u32)),
+                // The process called the `exit-restart` system call.
+                1 => process.try_restart(Some(completion_code as u32)),
+                // The process called an invalid variant of the Exit
+                // system call class.
+                _ => process.set_syscall_return_value(SyscallReturn::Failure(ErrorCode::NOSUPPORT)),
+            },
```
bradjc, line 879: Why did this move here?

alexandruradovici: I was hoping that Rust would be smart enough to determine that https://github.com/tock/tock/blob/3ecb382cec9684052fc373087c2bf2f070cfd843/kernel/src/kernel.rs#L1365 is not possible and not necessary. I will move it back.

hudson-ayers (dismissed):
(Note to other reviewers: I recommend toggling the "hide whitespace" option here)

This is a great example of a change that saves us flash size at no-cost other than reduced readability. I am an in favor of it.

bradjc (approved):

hudson-ayers:
@alexandruradovici can this be updated to include the final version of what is in #3300 (i.e. also include the additional error code fixes that made it into that PR)?

alexandruradovici:
Yes, I'll make the changes

hudson-ayers:
We could obviously just cherry-pick the commits onto the master branch, but we have branch protection rules setup to prevent anyone directly pushing the master branch, and I think it is more transparent for all changes to master to still go through PRs. So I would recommend resetting this branch back to master, then cherry-picking the commits from #3300 onto this branch, and then we can just merge this PR

alexandruradovici:
I think these are the changes pushed to 2.1.1

hudson-ayers (approved):

kernel/src/kernel.rs
```diff
@@ -1231,58 +1254,102 @@ impl Kernel {
...
-                    );
-                }
 
-                process.set_syscall_return_value(res);
+                        process.set_syscall_return_value(res);
+                    }
+                    _ => {
+                        // This should never be reachable due to the
+                        // previous match statement.
+                    },
```
lschuermann, line 1351:
```suggestion
                    Syscall::Yield { .. }
                    | Syscall::Exit { .. }
                    | Syscall::Memop { .. } => {
                        // These variants must not be reachable due to
                        // the outer match statement:
                        debug_assert!(false, "Kernel system call handling invariant violated!");
                    },
```

Let's carry [the same patch](https://github.com/tock/tock/pull/3300#discussion_r1002533300) here.

lschuermann (changes requested):

lschuermann (approved):

hudson-ayers (approved):

PR #3277, opened 2022-10-07
virtual_alarm: add unit test for quick alarm case and fix

jettr:
### Pull Request Overview

Ensure we test the case when there are multiple alarms that do not get fired on the first pass of checking, but they would have fired by the time we do the calculation to set the next arm time. In that scenario the next arm time is way in the future (rollover) and the alarms do not get called when they are supposed to.

@phil-levis provided final fix for issue.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/virtual_alarm.rs
+++ b/capsules/src/virtual_alarm.rs
@@ -278,7 +278,19 @@ impl<'a, A: Alarm<'a>> time::AlarmClient for MuxAlarm<'a, A> {
             .virtual_alarms
             .iter()
             .filter(|cur| cur.armed.get())
-            .min_by_key(|cur| cur.dt_reference.get().reference_plus_dt().wrapping_sub(now));
+            .min_by_key(|cur| {
+                let when = cur.dt_reference.get();
+                // If the alarm has already expired, then it should be
+                // considered as the earliest possible (0 ticks), so it
+                // will trigger as soon as possible. This can happen
+                // if the alarm expired *after* it was examined in the
+                // above loop.
+                if !now.within_range(when.reference, when.reference_plus_dt()) {
+                    A::Ticks::from(0u32)
+                } else {
+                    when.reference_plus_dt().wrapping_sub(now)
+                }
+            });
 
         // Set the alarm.
         if let Some(valrm) = next {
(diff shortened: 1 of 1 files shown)


jettr:
FYI I updated the unit test to better reflect the scenario.

We need another alarm in the future to make the test more accurate since that it way really triggers/exposes the issue.

phil-levis:
Can you explain the exact issue? The way that the virtualizer is currently implemented, this case should be handled by setting a low-level alarm in the very near future. I.e., suppose I have these three Alarms:

A: reference: 100, dt: 50
B: reference: 100, dt: 54
C: reference: 100, dt: 500

At time 150, the low-level alarm should fire. This should trigger alarm A. Suppose that, when the loop across the alarms examines B, the current time t=153, so B does not fire. However, by the time the loop completes it is now t=155. This means that B should have fired but didn't.

My understanding is that https://github.com/tock/tock/blob/b4ac85abed0045a5729c2b6c702a9bc8e58c97e2/capsules/src/virtual_alarm.rs#L277 should handle this. After firing the alarms, the virtualizer scans across the alarms and finds the earliest one. It should find that B is the earliest and set an alarm for (100, 54). The underlying Alarm implementation should know that 100 is in the past, so 154 is in the past, so this means fire as soon as possible, e.g.:

https://github.com/tock/tock/blob/b4ac85abed0045a5729c2b6c702a9bc8e58c97e2/chips/sam4l/src/ast.rs#L347
https://github.com/tock/tock/blob/b4ac85abed0045a5729c2b6c702a9bc8e58c97e2/arch/rv32i/src/machine_timer.rs#L48

But it sounds like this is not happening?

I see this approach (delays sometimes in alarms that are very close) as preferable to the potential of introducing an infinite loop. Delaying the alarms (to a subsequent interrupt) doesn't break concurrency in the system: the kernel will continue to handle interrupts, etc. In contrast, with this approach, if I repeatedly set an alarm with a tiny duration, I might enter an infinite loop. 

But you're seeing this problem happen in practice, so I must be missing something? Perhaps that line of code at the end lf the `alarm` handler isn't quite right?

phil-levis:
> But you're seeing this problem happen in practice, so I must be missing something? Perhaps that line of code at the end lf the `alarm` handler isn't quite right?

I think this is the problem.

```rust
        let next = self
            .virtual_alarms
            .iter()
            .filter(|cur| cur.armed.get())
            .min_by_key(|cur| cur.dt_reference.get().reference_plus_dt().wrapping_sub(now));
```

is incorrect. It should not be a simple `wrapping_sub`, as this fails in the above case. Instead, this needs to properly consider an expired alarm as earlier than any unexpired alarm.

phil-levis:
@jettr Can you check whether this fixes the problem?

jettr:
Yes, the [code](https://github.com/tock/tock/compare/master...jettr_alarm_fix) above does fix the issue and pass the unit test. I forced pushed this PR with your fix and the unit test (which I also updated to be a little more clear).

PR #3279, opened 2022-10-07
ADC: Add set_client() to `AdcHighSpeed` trait

bradjc:
### Pull Request Overview

This pull request is an alternative to #3267.

The base changes are:
- Adds a `set_client()` function to the `adc::AdcHighSpeed` trait. This is standard practice that we expect HILs to follow.
- Makes the `adc::HighSpeedClient` also include the base `adc::Client` callback.
- Updates the chip ADC implementations to implement the `adc::HighSpeedClient::set_client()` function. 

The downside to just making those changes is that any chip that implements `adc::AdcHighSpeed` now _has_ to use the `adc::HighSpeedClient::set_client()` function. So it is feasible that a board/capsule could use the `adc::Adc` trait for a chip, and then the chip driver could be updated to support `adc::AdcHighSpeed`, and then that board/capsule would break because the client would no longer be used. To address this, this PR:

- Adds a `EitherClient` enum to the adc drivers.

This enables the sam4l and msp432 adc drivers to use either client depending on which `set_client()` was called.

The downside is this adds some complexity on the implementation of the ADC trait. It's also possible that the wrong `set_client()` is used, but I don't see this as a larger problem than forgetting to call `set_client()` at all.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/msp_exp432p401r/src/main.rs
+++ b/boards/msp_exp432p401r/src/main.rs
@@ -13,6 +13,7 @@ use kernel::capabilities;
 use kernel::component::Component;
 use kernel::dynamic_deferred_call::DynamicDeferredCall;
 use kernel::dynamic_deferred_call::DynamicDeferredCallClientState;
+use kernel::hil;
 use kernel::hil::gpio::Configure;
 use kernel::platform::{KernelResources, SyscallDriverLookup};
 use kernel::scheduler::round_robin::RoundRobinSched;
@@ -409,8 +410,7 @@ pub unsafe fn main() {
             &mut capsules::adc::ADC_BUFFER3
         )
     );
-
-    peripherals.adc.set_client(adc);
+    hil::adc::AdcHighSpeed::set_client(&peripherals.adc, adc);
 
     // Set the reference voltage for the ADC to 2.5V
     peripherals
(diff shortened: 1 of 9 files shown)


boards/msp_exp432p401r/src/main.rs
```diff
@@ -409,8 +410,7 @@ pub unsafe fn main() {
             &mut capsules::adc::ADC_BUFFER3
         )
     );
-
-    peripherals.adc.set_client(adc);
+    hil::adc::AdcHighSpeed::set_client(&peripherals.adc, adc);
```
hudson-ayers, line 413: if someone were to call `peripherals.adc.set_client(adc)`, does rustc throw a warning that this is ambiguous and could call one of two methods, or does it just pick one based off of what is in scope? I suspect the latter. If so, this still seems to make it easy to set a non-high-speed client on accident, since this is one of very few places in main.rs where you have to fully specify which trait you are using in order to get correct behavior.

hudson-ayers:
This costs 192 bytes of flash on Imix, I think just storing two separate clients is probably preferable between the size and the "calling the wrong trait" footgun.

I personally still prefer https://github.com/tock/tock/pull/3267 but am fine to be outvoted

PR #3283, opened 2022-10-07
sifive and hifive1: support 344MHz system clock

gemarcano:
### Pull Request Overview

The SiFive FE310-G002 and FE310-G003 appear to have the same clock control hardware, and expose identical registers to manage the clocking of the system. This adds a ~~320MHz~~ (**edit: 344MHz**) clock frequency option, achieved by adjusting the PLL. Some additional code makes sure changing frequency is done safely.

~~UART needs to update its divider when the system clock changes to maintain a constant baud rate. A new function
`Uart::update_clock_frequency` handles this, and should be called after a system clock change. Additionally, now `Uart::set_baud_rate` is called on creation to make sure the internal baud rate and hardware state is consistent.~~

**Edit:** UART now takes in the desired system clock on creation, so that the `set_baud` calculations take that into account.

There is no need currently to do adjustments to the SPI peripheral serial clock divider. Default settings on the SPI peripheral divide the clock by 8, limiting the maximum frequency of the serial clock to 48MHz (maximum possible output from PLL peripheral is 384MHz). All QSPI chips listed as in use for HiFive boards support frequencies below 50MHz.

I am also looking for feedback on how to make it more flexible to choose semi-arbitrary clocks (there are over 100 valid unique clock outputs from the PLL block), or if this is something that is unnecessary, and whether we should just support a small set of frequencies.

A lot of the changes are comments just to make it easier to follow what is going on and why. I may have gone overboard, as I tend to err on the side of too much.

### Testing Strategy

I tested this on a HiFive1 revB. It actually helped me catch a bug in the SiFIve UART driver, which should be fixed by #3272.

I tested this by making sure UART communication happened, indicating that the baud rate was still properly clocked at 115200 even after adjusting the divider in the UART peripheral to account for the increased system clock. To confirm that the divider mattered, I removed the update logic to compensate for a change in the system clock, and I observed that UART communication was no longer possible (as expected).

### TODO or Help Wanted

I checked the documentation for the G003 chip, and I believe the registers are all the same, and possibly even the PLL peripheral, but it would be really good to have someone with a FE310-G003 to check (I think that's a HiFive Inventor???).

Additionally I have a few design questions, which I'll highlight in the comments.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/sifive/src/prci.rs
+++ b/chips/sifive/src/prci.rs
@@ -1,8 +1,11 @@
 //! Power Reset Clock Interrupt controller driver.
 
+use core::cell::Cell;
 use kernel::utilities::registers::interfaces::ReadWriteable;
+use kernel::utilities::registers::interfaces::Readable;
 use kernel::utilities::registers::{register_bitfields, ReadWrite};
 use kernel::utilities::StaticRef;
+use rv32i::csr;
 
 #[repr(C)]
 pub struct PrciRegisters {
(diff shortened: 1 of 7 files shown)


chips/sifive/src/prci.rs
```diff
@@ -48,29 +51,111 @@ register_bitfields![u32,
...
+
+        // Make sure external clock subsystem is enabled before adjusting the PLL
+        self.enable_external_clock();
 
         match frequency {
             ClockFrequency::Freq16Mhz => {
+                // Bypass enabled, feeds clock directly from external clock. For HiFive1 revB, this
+                // is a 16 MHz clock
                 regs.pllcfg
                     .modify(pllcfg::bypass::SET + pllcfg::refsel::SET);
```
gemarcano, line 123: One thing that bothers me here is the implicit assumption that all SiFive chips will be running off a 16MHz external clock. Should we, instead, provide some kind of input parameter when creating the peripheral indicating the value of the external clock as installed on the board? This way different SiFive chips on multiple boards can specify what their actual external clocks are.

chips/sifive/src/uart.rs
```diff
@@ -117,7 +119,11 @@ impl<'a> Uart<'a> {
             rx_len: Cell::new(0),
             rx_position: Cell::new(0),
             rx_status: Cell::new(UARTStateRX::Idle),
-        }
+        };
+
+        // Make sure current_baud matches what the hardware is actually configured for
+        result.set_baud_rate(115200);
```
gemarcano, line 125: Is it OK to execute a function while creating a Uart object? I am doing this just to synchronize the hardware state with what the software is assuming (specifically, the `current_baud` new variable).

chips/sifive/src/prci.rs
```diff
@@ -48,29 +51,111 @@ register_bitfields![u32,
 
 pub enum ClockFrequency {
     Freq16Mhz,
+    Freq320Mhz,
 }
```
gemarcano, line 55: What would be the best way to support more frequencies? Is it possible/do we want to support arbitrary frequencies (well, semi-arbitrary, there are rules about what frequencies are valid).

chips/sifive/src/prci.rs
alistair23: You can have an input parameter, we do that in [OpenTitan](https://github.com/tock/tock/blob/master/chips/earlgrey/src/chip_config.rs) for the UART speeds for example.

If you want to feel free to open a PR. The problem with the SiFive SoCs is that they aren't [great Tock candidates](https://github.com/tock/libtock-rs/issues/156) so there hasn't been a lot of interest.

chips/sifive/src/prci.rs
alistair23: I think an enum of supported and tested frequencies is the best option

chips/sifive/src/uart.rs
alistair23: I don't think this is right, the caller should specify the baud rate

chips/sifive/src/prci.rs
bradjc: Clocks are complicated, but the reality is most people just use the same crystals and oscillators that others use, and we may never see additional boards using the sifive g0002 chip. I wouldn't worry about trying to support what is possible, only what we actually need to use.

bradjc:
Following up on my comment in a thread on this PR, why do we need clock flexibility for this platform?

gemarcano:
From a solely "is someone using this" perspective, probably I'm the only one. The hardware has the capability to support over 100 unique viable clock frequencies, heck, even down into the kHz range.

My main motivation for this was mostly to see if I could do it. Additionally, I prefer running hardware faster rather than slower, especially if it is mostly just going to be plugged into some USB power source.

bradjc:
> running hardware faster rather than slower

That sounds like a good reason to me. Why not just change to the higher frequency? I agree that it doesn't seem likely anyone is going to try to run the hifive1 off of batteries.

gemarcano:
Sure, I can do that. From a quick test, it looks like 344MHz is probably a good value to set the frequency to. I can run Tock on my board up to around 344MHz, but above that the board is between unstable to outright locking up.

I'll leave the code for the 16MHz clock there, as I think the HiFive Inventor should probably be tested at 344MHz to make sure it can support that frequency stably.

Related, in Tock, there's no API to modify clocks, right? In other words, are clocks are set up upon board initialization and is it expected for them to remain constant?

alistair23:
Tock doesn't have support for variably scaling the clock frequency based on usage.

It does have a `sleep(&self)` function in the chip though. So you can enter a low power state there

hudson-ayers:
Tock supports a large variety of oscillators and clock speeds for the sam4l, including support for dynamically changing the clock at runtime (see sam4l/src/pm.rs). However this support is not exposed outside of the chip crate in mainline Tock, so there is no way for applications or even capsules to take advantage of this functionality. Holly's powerclocks work exposed all of this via additional HILs such that clocks would be automatically set based on the ongoing requirements and workload of apps, allowing for near optimal energy consumption. However this work was not merged into mainline Tock. I'm about to board a flight but I believe the code is linked in the paper: https://sing.stanford.edu/site/publications/93

gemarcano:
Thanks! I'll take a look at it. As the HiFive1 revB is so underpowered (the 16K of SRAM really hurts), it might not make sense to spend too much energy in replicating the SAM4L work for it, but it certainly has quite a lot of flexibility over its clocking. I think for this PR I'll just focus on exposing 344MHz as an alternative to 16MHz.

I might need to take a look at `sleep` as I think the PLL needs some extra work to wake up from a proper sleep.

gemarcano:
OK, for our current implementation of sleep for rv32i, it just uses `wfi`, for which there's nothing special that needs to be done with PLL. The HiFive1 does support a deeper sleep, where only a particular subset of peripherals (including the RTC) remain powered on, but the main CPU complex is turned off. When waking up, the CPU effectively restarts from scratch, and goes through the normal reset vector (there is a register somewhere that indicates if the reset is due to a wakeup or something else).

In other words, I don't think this matters for this PR.

chips/sifive/src/uart.rs
gemarcano: OK. In the latest code, I changed it so that the system clock is provided when the UART object is created, alleviating the issue this is trying to fix.

chips/sifive/src/prci.rs
gemarcano: For now, per another comment by @bradjc , I think we can assume that 16MHz external clock. If in the future we support another SiFive chip based board with a different external clock, we can add that later, I think.

gemarcano:
I've updated the code. The main change is switching to 344MHz for the HiFive1, and making the UART code take in the system clock rate when it is created, which simplified a lot of the code.

I updated code for the `hifive_invetor` and `redboard_redv` as both used the `E310x` based peripherals which I modified to pass through the clock information for the UART objects. I've only checked that these build, but don't have them to check that they run properly (do we know anyone that has these?). These other boards are still configured to run at 16MHz, which is what they have been using up to now.

alistair23 (dismissed):

bradjc (dismissed):

hudson-ayers:
@gemarcano this needs a rebase

gemarcano:
Sorry, I am planning on doing it, I just don't have access to my HiFives for the time being due to an ongoing labor strike at UCSD where the boards currently are. As soon as I can get access to the board I'll rebase and make sure the board is still stable.

hudson-ayers:
No rush! Just wanted to make sure this didn't get lost

lschuermann:
@gemarcano Did you get a chance to rebase and test this in the meantime?

gemarcano:
Coincidentally, I just finally managed to get back into the lab. I just rebased, and did some light testing with a second board, and it appears stable. I can run a custom application I made, and I can run the hello world examples from libtock-c. I can also connect over UART (`/dev/ttyACM0` on my computer) just fine.

hudson-ayers (approved):

lschuermann (approved):

PR #3290, opened 2022-10-11
ADC: Add set_client() to AdcHighSpeed trait (take #2)

bradjc:
### Pull Request Overview

This pull request is an alternative to #3279 and #3267.

This adds a set client function to the `AdcHighSpeed` trait, and then treats the high speed and regular traits separately, requiring implementations that want to use both to have two clients. The users of the drivers then need to set both clients.

This seems like the most direct approach. It simply implements what we do with all HILs, and is the same if a chip driver wanted to support, for example, both I2C and SPI: it implements both traits and has both clients.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/hail/src/main.rs
+++ b/boards/hail/src/main.rs
@@ -451,7 +451,7 @@ pub unsafe fn main() {
         )
     );
     hil::adc::Adc::set_client(&peripherals.adc, adc);
-    hil::adc::AdcHighSpeed::set_client(&peripherals.adc, adc);
+    hil::adc::AdcHighSpeed::set_highspeed_client(&peripherals.adc, adc);
 
     // Setup RNG
     let rng = components::rng::RngComponent::new(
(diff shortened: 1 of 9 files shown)


hudson-ayers:
I wonder if we should just call the method `set_highspeed_client()` to save us from having to worry about name collisions

bradjc:
While we have many opportunities for `set_client()` name collisions, it's more likely that the same object will implement Adc and AdcHighSpeed than some other arbitrary set of two different HILs. I added a commit so we do not have two closely related set_client functions in the ADC HILs.

hudson-ayers (approved):

hudson-ayers:
Yeah, I consider it analogous to like `set_alarm_client()` and `set_overflow_client()` in the time HIL.

hudson-ayers:
This approach is exactly what Phil suggested in https://github.com/tock/tock/pull/3267#pullrequestreview-1133853988 ; I think this can also be merged with one more approval.

(Notably, https://github.com/tock/tock/pull/3231 is blocked on this, and other improvements to components are blocked on #3231, so trying to keep things moving)

brghena (approved):
Looks fine to me

PR #3291, opened 2022-10-11
Rename appid and app_id to processid

alexandruradovici:
### Pull Request Overview

This pull request renames the usage of `appid` and `app_id` of type `ProcessId` to `processid`. 

This is now useful as the `AppID` pull request was merged and the AppID now has a different meaning now from the ProcessId. Most usages of `appid` and `app_id` actually are referring `ProcessId`.

### Testing Strategy

N/A

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/mpu.rs
+++ b/arch/cortex-m/src/mpu.rs
@@ -754,16 +754,16 @@ impl<const NUM_REGIONS: usize, const MIN_REGION_SIZE: usize> mpu::MPU
         Ok(())
     }
 
-    fn configure_mpu(&self, config: &Self::MpuConfig, app_id: &ProcessId) {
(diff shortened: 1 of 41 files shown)


bradjc (approved):
I agree, this seems like the time to finally make this change. When we changed the type to what is now `ProcessId` we intentionally did not change the variable names to reduce the complexity of that change and hopefully reduce the potential for mistakes. But with proper app id now, seems like now is the time to update the variable names.

alistair23 (approved):

ppannuto (approved):
+1 and same as Hudson re review 😅

PR #3297, opened 2022-10-17
Make `Component::finalize()` a safe trait method

hudson-ayers:
### Pull Request Overview

This pull request makes `Component::finalize()` safe to call. Now that `finalize()` cannot be used to initialize static objects, there is no reason for it to be an unsafe trait method. To support this change, I also had to make some modifications to one of the nrf52 startup components, and add a few local unsafe blocks within `finalize()`.

This further switches OpenTitan to use the new generic USB component and hence removes its USB component.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


bradjc (approved):
I looked through the files, all look good to me.

lschuermann (approved):

alistair23 (approved):

Issue #3278, opened 2022-10-07
Subscribe error code for a nonexistent driver_num


jrvanwhy:
[TRD 104](https://github.com/tock/tock/blob/master/doc/reference/trd104-syscalls.md#421-the-null-upcall) says:

If a userspace process invokes subscribe on a driver ID that is not installed in the kernel, the kernel MUST return a failure with an error code of `NODEVICE` and an upcall of the Null Upcall.

However, while implementing `libtock-rs`, I [observed](https://github.com/tock/libtock-rs/blob/1d785a043a95d83b410f6a099a6121fc101ca3b7/unittest/src/fake/syscalls/subscribe_impl.rs#L76) that as of Tock 2.0, the kernel returns `NOMEM` instead of `NODEVICE`.

I do not know what Tock 2.1 does -- Subscribe's implementation was overhauled between 2.0 and 2.1 -- but at least in 2.0 we had an implementation/documentation mismatch.

alexandruradovici:
#3276 fixes this issue.

alexandruradovici:
As far I noticed, 2.1 does not return `NODEVICE` for any of the `subscribe` and `allow` system calls.

jrvanwhy:
Best as I can tell, here are the documented and actual return codes for stable Tock releases:

| Syscall                  | TRD 104    | Tock 2.0     | Tock 2.1   |
| ------------------------ | ---------- | ------------ | ---------- |
| Command                  | `NODEVICE` | `NODEVICE`   | `NODEVICE` |
| Read-Only Allow          | `NODEVICE` | `NODEVICE`   | `NOMEM`    |
| Read-Write Allow         | `NODEVICE` | `NODEVICE`   | `NOMEM`    |
| Subscribe                | `NODEVICE` | `NOMEM`      | `NOMEM`    |
| Userspace-Readable Allow | `NODEVICE` | Didn't exist | `NODEVICE` |

alexandruradovici:
IMHO I think we would treat this as a bug. Based on @jrvanwhy table, I think it is clearly visible that this has been an honest implementation mistake.

The arguments being:
1. this was not an intentional change, we simply do not have any unit tests (I guess we should) and nobody actually checked that the new `subscribe` in 2.0 and new `allow` in 2.1 where actually following TRD 104
2. semantically they should return `NODEVICE` as this is the actual error. A user will not be able do distinguish between a driver that does not exist and a driver that does not properly allocate a grant.
[...]

PR #3300, opened 2022-10-18
2.1.1: Fix syscall return codes for wrong driver_number

alexandruradovici:
### Pull Request Overview

This PR is not for the master branch, but for the `release-2.1`. As there is no branch for it, I am can to reopen this towards another branch.

This pull request fixes the value returned by the `subscribe` and `allow` system calls. Before this PR: 
- providing a wrong driver_number would return `NOMEM`. The issue is presented in #3278;
- providing a wrong subscribe or allow number would return `INVAL`, now it returns `NOSUPPORT`.

This PR will be superseded by  #3276, it is only for releasing Tock 2.1.1. This PR will have a slightly different behavior than its successor:

| |Tock 2.1.1 (this PR) | master (#3276) |
| --- |-----------|---------|
| `subscribe (invalid_number, invalid_callback, ...)` | `INVAL` | `NODEVICE` |
| `subscribe (invalid_number, valid_callback, ...)` | `NODEVICE` | `NODEVICE` |
| `allow... (invalid_number, invalid_reference, ...)` | `INVAL` | `NODEVICE` |
| `allow... (invalid_number, valid_reference, ...)` | `NODEVICE` | `NODEVICE` |

In #3276 the check for a valid driver is performed before the check for parameters, while in this PR the check is performed after. Checking for a valid driver before the validity of arguments will add at least 600 KB of code to the kernel, while this adds only 32 B.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/kernel.rs
+++ b/kernel/src/kernel.rs
@@ -1350,9 +1350,12 @@ impl Kernel {
 
                         process.set_syscall_return_value(res);
                     }
-                    _ => {
-                        // This should never be reachable due to the
-                        // previous match statement.
+                    Syscall::Yield { .. }
+                    | Syscall::Exit { .. }
+                    | Syscall::Memop { .. } => {
+                        // These variants must not be reachable due to
+                        // the outer match statement:
+                        debug_assert!(false, "Kernel system call handling invariant violated!");
                     },
                 })
             }
--- a/kernel/src/kernel.rs
+++ b/kernel/src/kernel.rs
@@ -950,8 +950,8 @@ impl Kernel {
                                             }
                                         }
                                     }
-                                    alloc_failure @ AllocResult::NoAllocation
-                                    | alloc_failure @ AllocResult::SameAllocation => {
+                                    alloc_failure @ (AllocResult::NoAllocation
+                                    | AllocResult::SameAllocation) => {
                                         // We didn't actually create a new
                                         // alloc, so just error.
                                         match (config::CONFIG.trace_syscalls, alloc_failure) {
--- a/kernel/src/kernel.rs
+++ b/kernel/src/kernel.rs
@@ -1256,8 +1256,8 @@ impl Kernel {
                                             }
                                         }
                                     }
-                                    alloc_failure @ AllocResult::NoAllocation
-                                    | alloc_failure @ AllocResult::SameAllocation => {
+                                    alloc_failure @ (AllocResult::NoAllocation
+                                    | AllocResult::SameAllocation) => {
                                         // We didn't actually create a new
                                         // alloc, so just error.
                                         match (config::CONFIG.trace_syscalls, alloc_failure) {
--- a/kernel/src/kernel.rs
+++ b/kernel/src/kernel.rs
@@ -865,200 +865,194 @@ impl Kernel {
                     process.set_yielded_state();
                 }
             }
-            Syscall::Subscribe {
-                driver_number,
-                subdriver_number,
-                upcall_ptr,
-                appdata,
-            } => {
-                let res = resources
-        .syscall_driver_lookup()
-        .with_driver(driver_number, |driver| match driver {
-            Some(driver) => {
-                // A upcall is identified as a tuple of the driver number and
-                // the subdriver number.
-                let upcall_id = UpcallId {
-                    driver_num: driver_number,
-                    subscribe_num: subdriver_number,
-                };
+            Syscall::Subscribe { driver_number, .. }
+            | Syscall::Command { driver_number, .. }
+            | Syscall::ReadWriteAllow { driver_number, .. }
+            | Syscall::UserspaceReadableAllow { driver_number, .. }
+            | Syscall::ReadOnlyAllow { driver_number, .. } => {
+                resources
+                .syscall_driver_lookup()
+                .with_driver(driver_number, |driver| match syscall {
+                    Syscall::Subscribe {
+                        driver_number,
+                        subdriver_number,
+                        upcall_ptr,
+                        appdata,
+                    } => {
+                        // A upcall is identified as a tuple of the driver number and
+                        // the subdriver number.
+                        let upcall_id = UpcallId {
+                            driver_num: driver_number,
+                            subscribe_num: subdriver_number,
+                        };
 
-                // First check if `upcall_ptr` is null. A null `upcall_ptr` will
-                // result in `None` here and represents the special
-                // "unsubscribe" operation.
-                let ptr = NonNull::new(upcall_ptr);
+                        // First check if `upcall_ptr` is null. A null `upcall_ptr` will
+                        // result in `None` here and represents the special
+                        // "unsubscribe" operation.
+                        let ptr = NonNull::new(upcall_ptr);
 
-                // For convenience create an `Upcall` type now. This is just a
-                // data structure and doesn't do any checking or conversion.
-                let upcall = Upcall::new(process.processid(), upcall_id, appdata, ptr);
+                        // For convenience create an `Upcall` type now. This is just a
+                        // data structure and doesn't do any checking or conversion.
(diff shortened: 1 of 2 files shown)


bradjc (approved):
Looks good to me.

Also I pushed a branch called release-2.1.

kernel/src/kernel.rs
```diff
@@ -949,7 +950,8 @@ impl Kernel {
                                             }
                                         }
                                     }
-                                    alloc_failure => {
+                                    alloc_failure @ AllocResult::NoAllocation
+                                    | alloc_failure @ AllocResult::SameAllocation => {
```
hudson-ayers, line 954:
```suggestion
                                    alloc_failure @ (AllocResult::NoAllocation
                                    | AllocResult::SameAllocation) => {
```
nit, this is slightly more concise / better motivates use of an @ binding

kernel/src/kernel.rs
```diff
@@ -1071,7 +1076,8 @@ impl Kernel {
...
```
hudson-ayers, line 1080:
```suggestion
                                    alloc_failure @ (AllocResult::NoAllocation
                                    | AllocResult::SameAllocation) => {
```

kernel/src/kernel.rs
```diff
@@ -1242,7 +1256,8 @@ impl Kernel {
...
```
hudson-ayers, line 1260:
```suggestion
                                    alloc_failure @ (AllocResult::NoAllocation
                                    | AllocResult::SameAllocation) => {
```

hudson-ayers:
> will add at least 600 KB of code to the kernel

I hope not :) 

It feels very strange to me to have the 2.1.1 bugfix behave differently than what the syscall TRD specifies and differently than what we are planning for 2.2. It seems unlikely many people will use the 2.1.1 tagged release anyway, so might as well just take the 600 Byte hit on it and let it go away in master once #3276 is merged

twilfredo (approved):

alexandruradovici:
> 

I think it does behave like the TRD, just that error checking is in a different order.

hudson-ayers:
> I think it does behave like the TRD, just that error checking will be in a different order.
> 
> @hudson-ayers do you want me to change it back to the less efficient way?

I suppose you are right that it follows the TRD because the TRD is ambiguous about which error should take precedence. It still seems weird to have an API "bugfix" for the 2.1 release not present the same API as the bugfix we will apply to master/2.2.

@tock/core-wg thoughts?

jrvanwhy:
I'm surprised that changing the order of checks results in 600 bytes of bloat. Is something funny happening with inlining?

lschuermann:
> I suppose you are right that it follows the TRD because the TRD is ambiguous about which error should take precedence. It still seems weird to have an API "bugfix" for the 2.1 release not present the same API as the bugfix we will apply to master/2.2.

@hudson-ayers I agree. This seems like an important part of the ABI contract which is underspecified in TRD104 (by means of not unambiguously specifying precedence). I don't think we should now make use of these accidental degrees of freedom to implement different behaviors for minor releases of Tock, just because it is beneficial to us for some orthogonal (e.g., code size) reasons.

In search for a canonical interpretation of this aspect of TRD104, I would like to argue for an interpretation which says that the driver existence check and return of `NODEVICE` shall take precedence over other return codes, given that the top-level heading "4 System Call API" specifies, before mentioning any other explicit return variants: "If userspace invokes a system call on a peripheral driver that is not installed in the kernel, the kernel MUST return a Failure result with an error of NODEVICE." However, I realize this argument is arbitrary and I don't feel particularly strongly about the interpretation we would ultimately decide upon.

One more thing to consider is: what does this mean for other return codes? Immediately following the above specification, TRD104 reads: "If userspace invokes an unrecognized system call on a peripheral driver, the peripheral driver MUST return a Failure result with an error of NOSUPPORT." Does this mean that `NOSUPPORT` must also take precedence over `NOMEM`? (judging from the code this currently does not seem to be the case currently)

hudson-ayers:
I think a discussion of which error codes should take precedence in general can be deferred past this PR being merged, but agreed that is something we should resolve and clarify in TRD104.

For this particular case, I believe that NODEVICE should always take precedence and we should accept the code size hit in this PR since it will go away in master once #3276 is merged.

alexandruradovici:
What about merging #3276 into 2.1.1? The change is not major and, without it, the only way to respect the TRD is to search twice for the driver.

I can submit come changes to TRD 104 to reflect @lschuermann view.

alexandruradovici:
I made the modifications, but the performance hit is 1500 B for imix and 768 B on hifive1. This seems to be due to the closure code that is repeated in every match arm when searching for the driver. I would still recommend merging #3276 directly into 2.1.1 as the modifications there are what I would have done for this PR to reduce the size hit.

jrvanwhy:
I don't agree that we should specify an error code precedence order. I don't see a significant benefit to doing so, and there is a code size benefit to *not* specifying the error code precedence. I think this should be discussed elsewhere, but I'm saying this here so the discussion starts with "should we specify a precedence order" rather than "what should the precedence order be".

hudson-ayers:
> I made the modifications, but the performance hit is 1500 B for imix and 768 B on hifive1. This seems to be due to the closure code that is repeated in every match arm when searching for the driver. I would still recommend merging https://github.com/tock/tock/pull/3276 directly into 2.1.1 as the modifications there are what I would have done for this PR to reduce the size hit.

I would support this. I think when we decided on the call to try to separate out the change in API behavior from the code size optimizations it was not clear just how intertwined those changes were -- it seems obviously silly to have to lookup the driver number twice.

alexandruradovici:
I pushed all the modifications.

The questions that I still have is what happens if the user application wants to us an invalid subscribe number of allow number? The kernel now returns `INVAL`. Should it return `NOSUPPORT`?

phil-levis:
> I think a discussion of which error codes should take precedence in general can be deferred past this PR being merged, but agreed that is something we should resolve and clarify in TRD104.
> 
> For this particular case, I believe that NODEVICE should always take precedence and we should accept the code size hit in this PR since it will go away in master once #3276 is merged.

I agree that it's something we should resolve and clarify in TRD104 (even if it is to say that no ordering is promised and it's an implementation decision). I am not convinced, but am open to the idea, that NODEVICE should always take precedence over INVAL for kernel-checked values.

hudson-ayers (approved):

kernel/src/kernel.rs
```diff
@@ -1236,58 +1259,102 @@ impl Kernel {
...
-                    );
-                }
-
-                process.set_syscall_return_value(res);
+                        process.set_syscall_return_value(res);
+                    }
+                    _ => {
+                        // This should never be reachable due to the
+                        // previous match statement.
+                    },
```
lschuermann, line 1356:
```suggestion
                    Syscall::Yield { .. }
                    | Syscall::Exit { .. }
                    | Syscall::Memop { .. } => {
                        // These variants must not be reachable due to
                        // the outer match statement:
                        debug_assert!(false, "Kernel system call handling invariant violated!");
                    },
```

We should try to reduce the number of default match arms in `handle_syscall`, given its size and complexity.

I'm not sure about the benefit of the `debug_assert!` statement, but it can't hurt to have it around for tests.

lschuermann (approved):
This looks good to me generally, except for my nit in `handle_syscall`.

The discussion on yesterday's call has convinced me that we should not specify precedence of kernel-generated errors in response to system calls, in accordance with @phil-levis' comment. However, I think it would be beneficial for us to have tests which cover the _expected_ precedence of errors, such that we are made aware of changes influencing this implementation-codified and implicit precedence of errors returned to applications.

alevy (approved):

PR #3306, opened 2022-10-26
arch: cortex-m: format and comment assembly

bradjc:
### Pull Request Overview

After reading through the cortex-m assembly 100 times, I think we should capture what is actually going on so future readers can get up to speed quicker. This also tries to make the format consistent since rustfmt doesn't help here.

I also changed assembly like:

```asm
movw LR, #0xFFF9
movt LR, #0xFFFF
```

to:

```asm
ldr lr, =0xFFFFFFF9
```

which asks the assembler to figure out the best way to load that 32 bit immediate into the register. Since we are mostly doing this with numbers which are mostly 0xf-s, the assembler uses the `mvn` instruction:

```asm
mvn lr, #6
```

which loads the bitwise not of the immediate into the register.

Note! The psuedoinstruction _has_ to use the `=` symbol, as the `#` does something very different.

```asm
ldr lr, =0xFFFFFFF9  // works as expected
ldr lr, #0xFFFFFFF9  // does something very different and wrong
```

### Testing Strategy

Running on hail.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/lib.rs
+++ b/arch/cortex-m/src/lib.rs
@@ -126,19 +126,28 @@ pub unsafe extern "C" fn systick_handler_arm_v7m() {
     use core::arch::asm;
     asm!(
         "
-    // Set thread mode to privileged to switch back to kernel mode.
-    mov r0, #0
-    msr CONTROL, r0
-    /* CONTROL writes must be followed by ISB */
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):

bradjc:
Anyone else ok with this?

bradjc:
Can we merge this?

PR #3307, opened 2022-10-27
Add correct handling of Userspace Binary version numbers to boot sequence

phil-levis:
### Pull Request Overview

This pull request updates credentials checking so that it properly handles userspace binary version numbers. The AppID TRD says that, at boot, if the kernel encounters two processes with the same Application ID or Short ID, it'll boot the one with the higher version number, of if they have the same version number, the one that the schedule runs first. This patch implements this functionality and cleans up some text in the TRD.

It also adds a new checker for testing, in `kernel::process_checker::basic`, called `AppCheckerRsaSimulated`. This checker acts like a Checking Policy that accepts RSA signatures, assigns Application IDs as the public key of the signature, and short IDs derived from high order bits of the public key (if 0 is the MSB, the short ID is bits 1-31). However, it does not actually check the signature itself, instead blindly accepting an RSA signature. This checker allows you to easily check ID collisions and therefore test version numbers.

### Testing Strategy

This pull request was tested by installing two applications with RSA credentials using the same public key and different version numbers. The kernel correctly only booted the one with the higher version number. If this process is terminated at the process console, the one with the lower version number could successfully be booted from the console.

It was also tested by loading two applications with RSA credentials using the same public key and the same version number. It checks that the kernel runs one of them, whichever the scheduler runs first.

This adds 98 bytes to the kernel:
  - 30 of the bytes are adding an accessor method for application version number to ProcessStandard
  - 52 bytes added to kernel_loop (inlined from loading processes) to handle version numbers
  - 12 bytes because terminate has two extra cases, allowing you to terminate Faulted and Approved apps. You can terminate Faulted processes so that you can restart them; it is not valid to push a stack frame (init task) onto a Faulted process, as it might be storing fault state. So you must first transition it to Terminated. You can terminate CredentialsApproved apps because they are scanned in the main kernel loop to see if they can start. If the kernel wants to check credentials but then not start the process it needs to call terminate() after approving credentials.
  - The remaining bytes are due to alignment noise (+2 here or there).

### TODO or Help Wanted

Feedback on process `State` names.

### Documentation Updated

This also adds a lengthy comment to `process.rs` that explains the process state machine.

### Formatting

- [X] Ran `make prepush`.


phil-levis:
This adds 98 bytes to the kernel
  - 30 of the bytes are adding an accessor method for application version number to `ProcessStandard`
  - 52 bytes added to `kernel_loop` (inlined from loading processes) to handle version numbers
  - 12 bytes because `terminate` has two extra cases, allowing you to terminate Faulted and Approved apps. You can terminate Faulted processes so that you can restart them; it is not valid to push a stack frame (init task) onto a Faulted process, as it might be storing fault state. So you must first transition it to Terminated. You can terminate CredentialsApproved apps because they are scanned in the main kernel loop to see if they can start. If the kernel wants to check credentials but then not start the process it needs to call terminate() after approving credentials.
  - The remaining bytes are due to alignment noise (+2 here or there).

hudson-ayers:
I consider that size increase acceptable

kernel/src/process.rs
```diff
@@ -771,8 +776,44 @@ impl From<Error> for ErrorCode {
...
+/// and either leaves it in the `Faulted` state, restarts it, or takes
+/// some other action defined by the kernel fault policy. If the
+/// process issues an `exit-terminate` system call, it enters the
+/// `Terminated` state. If it issues an `exit-restart` system call, it
+/// terminates then tries to transition to `CredentialsApproved`.
+///
+/// When a process faults, it enters the `Faulted` state. To be
+/// restarted, it must first transition to the `Terminated` state,
+/// which means that all of its state has been cleaned up.
+
```
bradjc, line 816:
```suggestion
```

kernel/src/process.rs
```diff
@@ -771,8 +776,44 @@ impl From<Error> for ErrorCode {
 
 /// States a process can be in.
 ///
-/// This is made public in case external implementations of `Process` want
-/// to re-use these process states in the external implementation.
+/// This is public so external implementations of `Process` want to
```
bradjc, line 779:
```suggestion
/// This is public so external implementations of `Process` can
```

kernel/src/process_checker.rs
```diff
@@ -58,40 +60,81 @@ impl<'a> AppCredentialsChecker<'a> for () {
...
-pub fn has_unique_identifiers<AU: AppUniqueness>(
+/// Return whether `process` can run given the identifiers, version
+/// numbers, and execution state of other processes. A process is
+/// runnable if:
+///
+///   - Its Application Identifier and Short ID are different from
+///   all other processes, or
+///   - Other processes that share an Application Identifier or Short
+///   ID with it have a lower version number and are not running, or
+///   - Other processes that shared
```
bradjc, line 71: ?

phil-levis: Looks like it was truncated. Filled it in.

phil-levis:
@bradjc -- I think I addressed your comments.

doc/reference/trd-appid.md
```diff
@@ -379,50 +379,76 @@ deterministic behavior in the presence of colliding Application
...
+examine the process.
+
+The `CredentialsApproved` state indicates that the process is runnable
+in terms of its credentials. However, at any given time it might not
+be allowed to run because its Application Identifier or Short ID
+conflicts with another processs. When the main kernel loop examines a
+process in the `CredentialsApproved` state, it determines whether to
+run the process based on its Application Identifier, Short ID, and the
+Application Binary version number (stored in the Program Header,
+described in Section 5.1). At boot, the kernel starts a process if:
```
lschuermann, line 409:
```suggestion
described in Section 5.1). At boot, the kernel starts a process if
either of:
```

A nit, just to make it very explicit that this is ORed.

alevy (approved):

doc/reference/trd-appid.md
```diff
@@ -379,50 +379,76 @@ deterministic behavior in the presence of colliding Application
...
 
 The Unloaded, CredentialsUnchecked, CredentialsFailed,
-CredentialsPassed and Running states describe the conceptual state
+CredentialsApproved and Running states describe the conceptual state
 machine of a process at loading; the values or names of state
 variables in the kernel implemementation might be different. The exact
 Tock process state machine and names of its states is outside the
 scope of this document.
 
+One a Tock system is running, management interfaces may change the set
```
lschuermann, line 427:
```suggestion
Once a Tock system is running, management interfaces may change the set
```

doc/reference/trd-appid.md
```diff
@@ -379,50 +379,76 @@ deterministic behavior in the presence of colliding Application
...
+that a process at boot is blocked from running even if the process
+that blocks it does not run. For example, suppose there are three
+processes: A, B and C.
+
+  - A has a version number of 1, an Application ID of X and a Short ID of 6
+  - B has a version number of 2, an Application ID of Y and a Short ID of 6
+  - C has a version number of 3, an APplication ID of Y and a Short ID of 7
+
+Following the above logic, A cannot run because it shares a Short ID
+with A and has a lower version number. At the same time, B cannot run
```
hudson-ayers, line 445:
```suggestion
Following the above logic, A cannot run because it shares a Short ID
with B and has a lower version number. At the same time, B cannot run
```

doc/reference/trd-appid.md
alexandruradovici, line 427:
```suggestion
Once a Tock system is running, management interfaces may change the set
```

kernel/src/process_standard.rs
```diff
@@ -448,7 +453,10 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
     }
 
     fn terminate(&self, completion_code: Option<u32>) {
-        if !self.is_running() {
+        if !self.is_running()
+            && self.get_state() != State::Faulted
+            && self.get_state() != State::CredentialsApproved
+        {
```
brghena, line 459: Can you add a comment here explaining why these three states are all part of the early return? I think before it made sense to not terminate a process that isn't running anyways. Now I don't find the logic obvious though.

doc/reference/trd-appid.md
```diff
@@ -379,50 +379,76 @@ deterministic behavior in the presence of colliding Application
...
+Short ID among running processes.
+
+Note that the algorithm above can lead to transitive blocking, such
+that a process at boot is blocked from running even if the process
+that blocks it does not run. For example, suppose there are three
+processes: A, B and C.
+
+  - A has a version number of 1, an Application ID of X and a Short ID of 6
+  - B has a version number of 2, an Application ID of Y and a Short ID of 6
+  - C has a version number of 3, an APplication ID of Y and a Short ID of 7
```
hudson-ayers, line 443: This suggests that both of the following are true:

1. A given AppID can map to multipl shortIDs
2. A given shortID can map to multip AppIDs

That is a little surprising to me

kernel/src/process_standard.rs
```diff
@@ -448,7 +453,10 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
     }
 
     fn terminate(&self, completion_code: Option<u32>) {
-        if !self.is_running() {
+        if !self.is_running()
+            && self.get_state() != State::Faulted
+            && self.get_state() != State::CredentialsApproved
```
hudson-ayers, line 458: Why can we not terminate a function that has had its credentials approved, but has not yet run?

doc/reference/trd-appid.md
phil-levis: Done.

kernel/src/process_standard.rs
phil-levis: Great catch. I've added a comment.

kernel/src/process_standard.rs
phil-levis: Oh, we *can* terminate it. Note the !=. The logic here is confusing, I agree. @brghena pointed this out. I have added a comment to explain.

phil-levis:
Alright, I think I've addressed the comments; I explained the logic in `terminate` and removed the incorrect text in the TRD about transitive blocking. Thank you, @hudson-ayers and @brghena for the detailed read-through.

hudson-ayers (approved):

brghena (approved):

PR #3312, opened 2022-11-03
doc: Add ExternalDependencies.md

alistair23:
### Pull Request Overview

This adds a document that describes some of the requirements of adding external crates.

Once this is accepted the idea is then to add an external crate with https://github.com/tock/tock/pull/3092

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


doc/ExternalDependencies.md
```diff
@@ -0,0 +1,60 @@
...
+     * Wireless implementations are difficult to get the correct timing
+     * Wireless protocols are also very expensive to certify
+ * Rust embedded HALs
+     * For some platforms there are mature Rust Embedded HALs. This would take
+       a long time for Tock developers to re-implement themselves.
+
+### Project maturity
+
+The external crate being added must be a mature project, with a high quality
+of code. The project must be well regarded in the Rust community.
```
twilfredo, line 41: What do we typically look at to see if a crate is `high quality`?

doc/ExternalDependencies.md
```diff
@@ -0,0 +1,60 @@
...
+
+The Tock dependency tree can be generated by running the below command in the
+top level of the Tock repo.
+
+```shell
+cargo tree | grep -v /tock
+```
+
+The below tree should be updated whenever a dependency change is made.
+
```
twilfredo, line 58: was there meant to be a tree here?

doc/ExternalDependencies.md
```diff
@@ -0,0 +1,60 @@
...
+## Adding a dependency
+
+External dependencies can be added to Tock on a case by case basis. Each
+dependency will be reviewed for inclusion. Below is a general list of
+requirements for adding a new external crate.
+
+### Provide Important Functionality
+
+The external crate must provide important functionality that couldn't
+easily or realisticly be provided by the Tock developers.
```
twilfredo, line 25: *realistically

doc/ExternalDependencies.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Project maturity
+
+The external crate being added must be a mature project, with a high quality
+of code. The project must be well regarded in the Rust community.
+
+### Limited dependencies
+
+The external crate should include only a few dependencies. The fewer
+dependencies the crate introduces the more likeley it is to be accepted.
```
twilfredo, line 46: *likely

doc/ExternalDependencies.md
alistair23: It's kind of vague I know, but I don't really have a good way to describe whats "high quality"

alistair23: I think it'll just be up to reviewers

twilfredo (approved):

hudson-ayers:
I added this as an agenda item for tomorrow's call

hudson-ayers:
We ended up not having much time to discuss this on the call.

I made one comment on the call which I wanted to try to summarize here:

Having dependencies which use `unsafe` internally directly in capsules violates Tock's written threat model for capsules. I am of the opinion that even relatively high quality dependencies are more likely to have soundness bugs than Rust's core library -- especially once we are using a stable compiler -- and these are not equivalent cases. That said, I realize it is important that some capsules be able to call code (e.g. crypto code) that will be defined in external dependencies.

I proposed that all dependencies should be hidden in a top-level directory named `external`. This top-level directory would not itself be a crate (like `boards/`, for instance), but would hold a collection of crates, with each crate having a single external dependency. For the `ghash` example, there would be a `ghash-tock` crate in the `external` folder. This crate would implement a trait which should be defined in `capsules`. For the ghash example, this trait might look like:
```rust
trait Hash {
    fn update_padded(&self, &'static mut [u8]);
    fn finalize(&self); 
    // etc. 
}
```
and the ghash-tock crate would define a struct which would implement this trait using the external dependency. The capsule (e.g. `aes_gcm`) would be generic over this trait, and accept an instance of it when created. Accordingly, the board using the capsule would be responsible for passing a GhashTock struct into the aes_gcm capsule in main.rs. Thus it is the board author explicitly choosing to use the ghash-tock crate, a crate clearly marked as containing an external dependency, and doing this mapping itself. 

The upsides of this approach:
1. `capsules` does not itself gain any external dependencies, and all dependencies on potentially unsafe crates are expressed directly from the board repo, as is currently the case.
2. This works well to express situations where a given interface might be implemented by a trusted, high quality software library *or* by an accelerator, since the trait could be implemented for either and used equivalently (with some potential problems dealing with async vs. sync interfaces...)
3. The way that code can interact with the dependency is very clearly defined in this external interface crate, as opposed to the alternative approach where auditing the use of the external dependency requires tracking down every place that `ghash::x` is called throughout all of capsules. 
4. Boards not using the individual capsules which require external dependencies do not have to compile those dependencies. This is nice because of a problem with external dependencies directly in `capsules`: even boards that do not need the dependency at all will be forced to compile them. Compiling external crates with large dependency trees can be quite slow, and compiling external crates could lead to the execution of build.rs files or procedural macros that themselves represent potential security vulnerabilities. 

---------------------------
I also realize that this approach has a size and ergonomics overhead relative to direct use of the dependency. It also may be difficult to incorporate certain dependencies behind traits (procedural macro support traits being one such example). If this turns out to be impractical, I am sympathetic to that. Regardless, I think we should try to find a solution that addresses my final point above. So if this trait-based separation approach is impractical, I think that we should still create a seperate crate or crates for capsules with dependencies, we could just call it `external-capsules` or something, and that way boards that do not want to compile dependencies are not forced to.

bradjc:
I somehow missed Hudson's comment before doing this, but I did a pass updating the doc. My main changes:

1. I specified we only do external deps for crypto from https://github.com/RustCrypto.
2. I specified one approach Hudson mentioned, specifically that all dependencies are included from a specific crate.
3. I wrote down what we talked about before with board crates having more latitude to use external dependencies.

I'm not overly attached to item 2, but I do like the idea that it would be easily grep-able.

bradjc:
> Having dependencies which use `unsafe` internally directly in capsules violates Tock's written threat model for capsules.

Right, do we treat external dependencies like unaudited capsule code, or like highly managed code (like the kernel crate)? I think I end up on the side of like highly managed code. This is more consistent with the ethos that we should be very selective of our external dependencies, and really we are just adding a convenience layer versus creating a vendor-specific copy in the tock repo itself.

> Thus it is the board author explicitly choosing to use the ghash-tock crate, a crate clearly marked as containing an external dependency, and doing this mapping itself.

In my view, this is how board-specific dependencies should (and would be forced to) work. For core dependencies like cryptographic libraries, I find it dissonant that we are including external crates to promote secure code, but then falling back to trusting board authors to do the right thing. I view the external dependency inclusion process as the audit that the dependency is ok, and we do not want board authors overriding that choice.

Now, I do understand the SE challenge of allowing SW or HW implementations. But I think that comes at a different layer, using HILs, and not with a new collection of capsule interfaces.

doc/ExternalDependencies.md
```diff
@@ -5,56 +5,147 @@ External Dependencies
...
+External dependencies can be added to Tock on a case-by-case basis. Each
+dependency will be reviewed for inclusion, according to the criteria in this
+section. The requirements are intentionally strict.
+
+There are two general methods to for including an external dependency in the
+Tock kernel: core external dependencies, and board-specific external
+dependencies. Core external dependencies may be used in "core" Tock crates, such
+as the `kernel`, `chips`, and `capsules` crates. Board-specific external
+dependencies may _only_ be used by crates in the `board/` folder. The processes
+for inclusion between these two methods are different.
```
hudson-ayers, line 59: For crypto crates that will ultimately need to be used by the core kernel, this makes sense to me -- it is inevitable that those crates will need to be compiled by all boards unless we want to rely on some hacky conditional includes that will be a mess to maintain.

For external dependencies which are *only* needed by individual capsules, I think things look a little different -- we really don't want to force all boards to compile those external crates if they are not using those capsules, so I think any such capsules should exist in stand-alone crates rather than inside the `capsules` crate.

`ghash` is an interesting example because currently it is only needed in capsules, but my understanding is that the plan is to eventually use it as part of the kernel app verification workflow (is this correct, @phil-levis ?), which would require the kernel to depend on it anyway. So I don't know that we need to move Alistair's capsule into a standalone crate.

hudson-ayers:
I think that I am happy with the current text of the document, modulo my one comment

doc/ExternalDependencies.md
bradjc: What's the right place to put the separation? I worry that we could be including fine-grained slicing before it's clear we need to. Does anyone care if the first build takes a couple seconds longer? I agree that once it's minutes then that is a lot of overhead, but perhaps we cross that bridge when we get there.

We also could look at separating capsules in to multiple crates, which could address this issue as well. Or put all capsules-with-dependencies in their own crate.

alistair23: `ghash` isn't specifically for app signing. The idea with the `ghash` crate is that we can expose AES-GCM on all platforms that have any AES hardware. The goal here would be to support all AES block ciphers (GCM, CCM, CTR, CBC and ECB) on any platform that supports AES. That way we can rely on all implementations existing and expose them all to userspace and the kernel.

Basically CTR, CBC and ECB are commonly supported in hardware, and the other two can be built with software on top of the hardware.

In the longer run though, I think it would be worth supporting crypto operations (AES, RSA, ECC) on all platforms with software fallbacks. The goal there is to allow loading signed/encrypted applications on all Tock platforms. As a secure operating system, it seems very useful for Tock to be able to do that.

I do agree with Brad. It does seem like we are worrying about something that isn't a problem yet. In terms of crypto crates we only ever expect to include a handful. We can also do the `external-capsules` direction, but I feel like the crypto ones might end up being included all the time anyway

doc/Design.md
```diff
@@ -289,27 +289,18 @@ that the operation is synchronous. These cases are rare, though: the operation
...
+Generally it's best to avoid adding any external crates to the Tock project.
+
+In certain situations Tock does allow external depenencies. This is decided on
+a case by case basis. For more details on this see
+[External Dependencies](ExternalDependencies.md).
+
+The other option to add an external library is to put specific portions of
+libraries into the `libraries` folder. This puts the library's source in the
+same repository, while keeping the library as a clearly separate crate. We do
+try to limit how often this happens.
```
lschuermann, line 303: Possible interop with the discussions around #3318 here? We want to make sure that licenses are compatible and it may be worth linking to that.

hudson-ayers:
We talked about this on the core call last week, some people read the updated version of this for the first time. The conclusion was that if we are gonna do this, we should go ahead and restructure the capsules directory to not be a crate, and instead have a few crates contained within it. I opened https://github.com/tock/tock/issues/3346 to propose a division, and will implement whatever consensus is reached there

doc/ExternalDependencies.md
phil-levis: Yeah, `ghash` isn't needed for application credentials. However, public key signatures (RSA, ECDSA, DH, etc.) will need to use tested-and-true software implementations (existing crates).

hudson-ayers:
Update: we reached a consensus in #3346, but I have not had a chance to implement the division (it is mostly just a large mechanical refactor). If someone else has a chance to implement it before me please feel free.

alistair23:
Ok! Now that https://github.com/tock/tock/issues/3346 is fixed and https://github.com/tock/tock/pull/3396 is merged we can get back to this.

Is the plan to have a `extern` crate that contains all of the capsules that have external dependencies? Or do we want a individual crate per external capsule?

doc/Design.md
brghena: I don't think we'll need to document that here. Licenses will indeed matter, but that seems like something we can handle on a per-PR basis.

bradjc:
I updated the document with a first pass based on the call. @lschuermann please revise as needed.

lschuermann (commented):
Thanks @bradjc! I restructured those sections a little, to reflect the proposed policy around the various capsule-crates.

While our previous discussions really only focused on adding dependencies to the capsule-crates, they did not explicitly consider board-specific external dependencies. However, those seem even less controversial, given a board can simply choose to depend on a capsule-crate which has external dependencies.

Finally, I'd appreciate some input on my comment regarding core external dependencies. I recall that there was a lot of pushback against that, which is why we converged on the capsule-crate based approach in the first place. The document may look inconsistent as of now, given that it establishes a policy around core external dependencies (which judging from the discussions we might not even want right now, at least not without concrete motivating examples), but not a process for including these.

It should be noted that the motivating example for this PR (#3092) could be resolved by, for instance, a `crypto` capsule crate.

doc/ExternalDependencies.md
```diff
@@ -0,0 +1,255 @@
...
+dependency will be reviewed for inclusion, according to the criteria in this
+section. The requirements are intentionally strict.
+
+There are two general methods to for including an external dependency in the
+Tock kernel: core external dependencies, and capsule- or board-specific
+external dependencies.
+
+- Core external dependencies may be used in "core" Tock crates, such as the
+  _kernel_, _arch_, and _chip crates_. Notably, individual boards generally do
+  not have a reasonable choice whether to depend on those crates.
```
bradjc, line 136: Why don't we just remove this?

I think the two categories are "boards" and "not boards".

bradjc: The "core" notation is I think a leftover from the earlier version of the draft, and I'm not sure there is a good reason to keep it.

lschuermann: > The "core" notation is I think a leftover from the earlier version of the draft, and I'm not sure there is a good reason to keep it.

I agree, and I think this would make the document much less confusing overall. Along with that I would remove the policies around _core external dependencies_ overall. We can punt on that until we need it, and resurrect the text. I just didn't want to delete it without discussing first.

> I think the two categories are "boards" and "not boards".

I think we do now have "boards", "capsules" and the rest, but the document should reflect that properly now.

lschuermann: Okay, I have removed the notion of "core external dependencies" and changed the strict requirements on external dependencies (maturity, small dependency tree, etc.) to be guidelines for dependency selection. Let me know what you think!

doc/ExternalDependencies.md
```diff
@@ -0,0 +1,233 @@
...
+
+Tock limits its use of external libraries for all crates in the kernel. This is
+done to promote safety, as auditing the Tock code only requires inspecting the
+code in the Tock repository. Tock tries to be very specific with its use of
+`unsafe`, and tries to ensure that when it is used it is clear as to why. With
+external dependencies, verifying uses of `unsafe` are valid is more challenging
+to, particularly as external libraries evolve.
+
+External dependencies also typically themselves rely on dependencies, so
+including one external crate likely pulls in several external crate. As of Nov
```
brghena, line 50:
```suggestion
including one external crate likely pulls in several external crates. As of Nov
```

doc/ExternalDependencies.md
```diff
@@ -0,0 +1,233 @@
...
+* Wireless protocols.
+  * Wireless implementations are difficult to get the correct timing.
+  * Wireless protocols are also very expensive to certify.
+
+Note, however, that _only_ the board crate itself may include such an external
+dependency in its `Cargo.toml` file.
+
+A possible way to have other crates indirectly use such a dependency is through
+a wrapper-trait. Such traits abstract the external dependency in a way that
+allows other crates to still be built without the dependency included.
```
brghena, line 177: But we said up above that we want to avoid wrapper-traits. This feels like a contradiction.

brghena: From our discussion, this point is good, and it's up top that should be made a little more flexible. Use wrappers where they make sense, but not required.

hudson-ayers:
Items from the call:
- We should clarify that this document only applies to things that will end up in the kernel binary
- Would be good to lighten the language about cargo in the intro
- Move the "specific approach motivation" down to the end of the doc, and possibly rename that section

bradjc:
I updated based on Amit's comment of the discussion.

Is this ready to merge?

hudson-ayers (approved):
I think this is ready!

lschuermann (approved):

PR #3315, opened 2022-11-04
boards: update `let _` syntax

twilfredo:
### Pull Request Overview

Since rust 1.59, we can simplify the use of an unused return value from `let _ = bar()` to just `_ = bar()`.

### Testing Strategy

Compiling each target

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/esp32-c3-devkitM-1/src/main.rs
+++ b/boards/esp32-c3-devkitM-1/src/main.rs
@@ -290,7 +290,7 @@ unsafe fn setup() -> (
     .finalize(components::process_console_component_static!(
         esp32_c3::timg::TimG
     ));
-    let _ = process_console.start();
+    _ = process_console.start();
 
     let esp32_c3_board = static_init!(
         Esp32C3Board,
--- a/boards/opentitan/src/io.rs
+++ b/boards/opentitan/src/io.rs
@@ -91,7 +91,7 @@ pub unsafe extern "C" fn panic_fmt(pi: &PanicInfo) -> ! {
         &PROCESS_PRINTER,
     );
 
-    let _ = writeln!(writer, "{}", pi);
+    _ = writeln!(writer, "{}", pi);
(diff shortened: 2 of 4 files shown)


alistair23 (approved):

hudson-ayers (approved):
This looks fine to me, I honestly don't really have a preference between the two different approaches.

lschuermann (commented):
I don't feel strongly about this. On the one hand I think forcefully changing this doesn't give us any advantage whatsoever and actually prevents us from using code outside of the context of the project on older Rust compiler versions. On the other hand, we are happily taking advantage of newly introduced `rustc` features elsewhere without worrying about this. So I'm indifferent.

bradjc:
Does this feature have a name?

twilfredo:
> Does this feature have a name?

Think it's just called `Ignored pattern binding`, the operator that is. Not sure if there a 'feature' name for it.

ppannuto:
I don't have a strong feeling one way or the other on this, but is there a clippy lint that will ensure we are consistent with this?

bradjc:
I don't understand why rust needs two nearly identical ways to do the same thing. If we can add a clippy lint to effectively enforce this (as we do with something like not explicitly calling `return` at the end of a function), which suggests this is or would become Rust Standard Style TM, then let's make the change. Otherwise, I'm opposed to this PR.

bradjc (changes requested):
Requesting clippy lint so we have a goal we are working towards rather than arbitrary code syntax changes.

twilfredo:
I couldn't find any clippy options for this, does anyone know if there are any? if not, it's probably best to just close this one off.

hudson-ayers:
I don't think such a lint exists, so I am going to close this for now. If a lint is created that would let us favor one approach or the other and enforce that throughout the codebase, we can revisit

PR #3317, opened 2022-11-04
Add default license notice to all Rust files

alevy:
### Pull Request Overview

This pull request adds a default copyright and license notice to all Rust files in the repository

### Testing Strategy

None

### Documentation Updated

- [] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [ ] Ran `make prepush`.

PR #3318, opened 2022-11-05
trd: license and copyright policy

ppannuto:
### Pull Request Overview

First draft of TRD for license and copyright rules for the Tock project.

### Testing Strategy

N/A

### TODO or Help Wanted

Review of language and policy decisions.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/doc/reference/trd-legal.md
+++ b/doc/reference/trd-legal.md
@@ -24,8 +24,8 @@ Explicitly not discussed in this TRD are issues of trademark, the Tock brand,
 logos, or other non-code assets and artifacts.
 
 
-1. Introduction
-===============
+1 Introduction
+==============
 
 Tock’s goal is to provide a safe, secure, and efficient operating system for
 microcontroller-class devices. The Tock project further believes it is
@@ -39,8 +39,8 @@ The intent of these policies is to best satisfy the needs of all stakeholders
 in the Tock ecosystem.
 
 
-2. Licensing
-============
+2 Licensing
+===========
 
 All software artifacts under the umbrella of the Tock project are
 dual-licensed as Apache License, Version 2.0 (LICENSE-APACHE or
@@ -50,8 +50,8 @@ http://opensource.org/licenses/MIT).
 All contributions to Tock projects MUST be licensed under these terms.
 
 
-3. Copyright
-============
+3 Copyright
+===========
 
 Entities contributing resources to open-source projects often require
 attribution to recognize their efforts. Copyright notices are a common means to
@@ -69,8 +69,8 @@ the original copyright holder.
 For full authorship information, see the version control history.
 
 
-4. Implementation
-=================
+4 Implementation
+================
 
 All code artifacts in Tock projects MUST include a license notice.
 
@@ -173,7 +173,7 @@ this policy. Such situations MUST include public explanation and public
 record of non-anonymized vote results. This is not expected to ever occur.
 
 
-6 Author’s Address
+5 Author’s Address
 ==================
 
     Pat Pannuto
--- a/doc/reference/trd-legal.md
+++ b/doc/reference/trd-legal.md
@@ -3,7 +3,7 @@ Licensing and Copyrights
 
 **TRD:** <br/>
 **Working Group:** Core<br/>
-**Type:** Documentary<br/>
+**Type:** Best Current Practice<br/>
 **Status:** Draft <br/>
 **Author:** Pat Pannuto<br/>
 **Draft-Created:** 2022/11/05 <br/>
@@ -42,11 +42,6 @@ in the Tock ecosystem.
 2. Licensing
 ============
 
-Tock’s primary licensing goal is to ensure ease of use for Tock project elements.
-For this reason, Tock offers its resources under the minimally encumbered MIT
-license and under the Apache2 license for those entities concerned about possible
-litigation issues.
-
 All software artifacts under the umbrella of the Tock project are
 dual-licensed as Apache License, Version 2.0 (LICENSE-APACHE or
 http://www.apache.org/licenses/LICENSE-2.0) or MIT license (LICENSE-MIT or
@@ -58,17 +53,18 @@ All contributions to Tock projects MUST be licensed under these terms.
 3. Copyright
 ============
 
-Entities contributing resources to open-source projects often require attribution
-to recognize their efforts. Copyright is a common means to provide this. For
-downstream users, the license terms of Tock ensures unencumbered use.
+Entities contributing resources to open-source projects often require
+attribution to recognize their efforts. Copyright notices are a common means to
+provide this. For downstream users, the license terms of Tock ensures
+unencumbered use.
 
 Copyrights in Tock projects are retained by their contributors. No
 copyright assignment is required to contribute to Tock projects.
 
 Artifacts in the Tock project MAY include explicit copyright notices.
 Substantial updates to an artifact MAY add additional copyright notices
-to an artifact. Copyright notices MUST NOT be removed by any party except
-the original copyright holder, and removal is done solely at their discretion.
+to an artifact. Copyright notices SHOULD NOT be removed by any party except
+the original copyright holder.
 
 For full authorship information, see the version control history.
 
@@ -78,8 +74,8 @@ For full authorship information, see the version control history.
 
 All code artifacts in Tock projects MUST include a license notice.
 
-Code artifacts MAY include copyright notice(s). Copyright notices MUST
-include a year. Newer copyright notices MUST be placed after existing
+Code artifacts MAY include copyright notice(s). Copyright notices SHOULD
+include a year. Newer copyright notices SHOULD be placed after existing
 copyright notices. If non-trivial updates are performed by an original
 copyright author, they MAY amend the year(s) indicated on their existing
 copyright statement or MAY add an additional copyright line, at their
@@ -89,49 +85,46 @@ discretion.
 4.1 Format
 ----------
 
-License and copyright information MUST appear near the top of a file.
-Generally, license and copyright information SHOULD be the first lines
-in a file, however exceptions may be made where technically necessary
-(e.g. to accommodate the shebang in a shell script).
+License and copyright information SHOULD appear near the top of a file.
+Generally, license and copyright information SHOULD be the first lines in a file
+unless necessary for technical (e.g. to accommodate the shebang in a shell
+script) or other reasons (e.g. importing a file with an existing copyright
+notice in a different location).
 
-License and copyright information MUST have exactly one (1) blank line
+License and copyright information SHOULD have exactly one (1) blank line
 separating it from any other content in the file.
 
-License information MUST come before copyright information.
-There MUST NOT be any whitespace between lines in the license/copyright block.
-
-Text described in this section MAY be pre-fixed or post-fixed with
+Text described in this section SHOULD be pre-fixed or post-fixed with
 technically necessary characters (i.e. to mark as a comment in source
 code) as appropriate.
 
-The first line of license text MUST appear exactly as-follows:
+The first line of license text SHOULD appear as-follows:
 
 > Licensed under the Apache License, Version 2.0 or the MIT License.
 
-The second line of license MUST adhere to the [SPDX](https//spdx.dev)
+The second line of license SHOULD adhere to the [SPDX](https//spdx.dev)
 specification for license description. As of this writing, it SHOULD
 appear as-follows:
 
 > SPDX-License-Identifier: Apache-2.0 OR MIT
 
-The [current (v2.3) normative rules](https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/)
-permit case-insensitive matches of the license identifier but require
-case-sensitive matching of the disjunction operator. To simplify
-enforcement of licensing and documentation rules, Tock dictates that
-license information MUST preserve case as-shown in the SPDX license
-list (i.e. as-presented above).
+The [current (v2.3) normative
+rules](https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/) permit
+case-insensitive matches of the license identifier but require case-sensitive
+matching of the disjunction operator. To simplify enforcement of licensing and
+documentation rules, license information SHOULD preserve case as-shown in the
+SPDX license list (i.e. as-presented above).
 
-Copyright lines MUST follow this pattern:
+Copyright lines SHOULD follow this pattern:
 
 > Copyright {entity} {year}([,year],[-year]).
 
-The `{entity}` field is REQUIRED and should reflect the entity wishing
-to claim copyright. The `{year}` field is REQUIRED and should reflect
-when the copyright is first established. Substantial updates in the
-future MAY indicate renewed copyright, via additional comma-separated
-years or via range syntax, at the copyright holder’s discretion. The
-initial year SHALL NOT be removed unless it is the express intent of
-the copyright holder to relinquish the initial copyright.
+The `{entity}` field should reflect the entity wishing to claim copyright. The
+`{year}` field SHOULD reflect when the copyright is first established.
+Substantial updates in the future MAY indicate renewed copyright, via additional
+comma-separated years or via range syntax, at the copyright holder’s discretion.
+The initial year SHALL NOT be removed unless it is the express intent of the
+copyright holder to relinquish the initial copyright.
 
 
 ### 4.1.1 Examples
@@ -144,6 +137,8 @@ The common-case format is the top of a file with only license information:
 
 //! Module-level documentation...
 ```
+
+It need not include a copyright line.
 
 A file with a long history and multiple copyrights may look as follows:
 
@@ -168,7 +163,7 @@ Many additional examples are available throughout the Tock repositories.
 ---------------
 
 To ensure coverage and compliance with these policies, the Core Team
-SHALL author and maintain tooling which enforces the presence and correct
+SHALL author and maintain tooling which checks the presence and expected
 format of license and copyright information. This SHOULD be automated and
 integrated with continuous integration systems. Contributions which do
 not satisfy these license and copyright rules MUST NOT be accepted.
--- a/doc/reference/trd-legal.md
+++ b/doc/reference/trd-legal.md
@@ -1,0 +1,188 @@
+Licensing and Copyrights
+========================================
+
+**TRD:** <br/>
+**Working Group:** Core<br/>
+**Type:** Documentary<br/>
+**Status:** Draft <br/>
+**Author:** Pat Pannuto<br/>
+**Draft-Created:** 2022/11/05 <br/>
+**Draft-Modified:** 2022/11/05 <br/>
+**Draft-Version:** 1 <br/>
+**Draft-Discuss:** tock-dev@googlegroups.com<br/>
+
+Abstract
+--------
+
+This document describes Tock’s policy on licensing and copyright. It explains
+the rationale behind the license selection (dual-license, MIT or Apache2) and
+copyright policy (optional, and authors may retain copyright if desired). It
+further outlines how licensing and copyright shall be handled throughout
+projects under the Tock umbrella.
+
+Explicitly not discussed in this TRD are issues of trademark, the Tock brand,
+logos, or other non-code assets and artifacts.
+
+
+1. Introduction
+===============
+
+Tock’s goal is to provide a safe, secure, and efficient operating system for
+microcontroller-class devices. The Tock project further believes it is
+important to enable and support widespread adoption of safe, secure, and
+efficient software. Tock also seeks to be an open and inclusive project, and
+Tock welcomes contributions from any individuals or entities wishing to
+improve the safety, security, reliability, efficiency, or usability of Tock
+and the Tock ecosystem.
+
+The intent of these policies is to best satisfy the needs of all stakeholders
+in the Tock ecosystem.
+
+
+2. Licensing
+============
+
+Tock’s primary licensing goal is to ensure ease of use for Tock project elements.
+For this reason, Tock offers its resources under the minimally encumbered MIT
+license and under the Apache2 license for those entities concerned about possible
+litigation issues.
+
+All software artifacts under the umbrella of the Tock project are
+dual-licensed as Apache License, Version 2.0 (LICENSE-APACHE or
+http://www.apache.org/licenses/LICENSE-2.0) or MIT license (LICENSE-MIT or
+http://opensource.org/licenses/MIT).
+
+All contributions to Tock projects MUST be licensed under these terms.
+
+
+3. Copyright
+============
+
+Entities contributing resources to open-source projects often require attribution
+to recognize their efforts. Copyright is a common means to provide this. For
+downstream users, the license terms of Tock ensures unencumbered use.
+
+Copyrights in Tock projects are retained by their contributors. No
+copyright assignment is required to contribute to Tock projects.
+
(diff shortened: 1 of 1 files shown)


doc/reference/trd-legal.md
```diff
@@ -0,0 +1,188 @@
...
+=================
+
+All code artifacts in Tock projects MUST include a license notice.
+
+Code artifacts MAY include copyright notice(s). Copyright notices MUST
+include a year. Newer copyright notices MUST be placed after existing
+copyright notices. If non-trivial updates are performed by an original
+copyright author, they MAY amend the year(s) indicated on their existing
+copyright statement or MAY add an additional copyright line, at their
+discretion.
```
hudson-ayers, line 86: So this is slightly different than Amit's approach in https://github.com/tock/tock/pull/3317, in that it does not require a copyright notice in all files. I personally think excluding the copyright notice except when individuals want to include it sounds nice.

hudson-ayers (commented):
Updating https://github.com/tock/tock/pull/3317 to match the process in this document, as best as I can tell, would require the following changes:

1. Updating non-rust files that are still code files (e.g. the python/bash code in `tools/` and our Makefiles) to include the license notice
2. Remove the "Copyright Tock Contributors" line from each file, only have a Copyright line if someone chooses to add it.
3. Add the line "Licensed under the Apache License, Version 2.0 or the MIT License." as the first line of each file, before the SPDX license identifier.

Overall this approach looks good to me. Curious to hear what @jrvanwhy in particular thinks about not requiring the "Copyright Tock Contributors" line on all files without a different copyright assignment

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,188 @@
...
+4.1 Format
+----------
+
+License and copyright information MUST appear near the top of a file.
+Generally, license and copyright information SHOULD be the first lines
+in a file, however exceptions may be made where technically necessary
+(e.g. to accommodate the shebang in a shell script).
+
+License and copyright information MUST have exactly one (1) blank line
+separating it from any other content in the file.
```
brghena, line 98: "exactly one" or "at least one"? I don't know if it's necessary to be this picky here.

brghena (dismissed):
This all looks like a reasonable approach to me.

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,188 @@
...
+in the Tock ecosystem.
+
+
+2. Licensing
+============
+
+Tock’s primary licensing goal is to ensure ease of use for Tock project elements.
+For this reason, Tock offers its resources under the minimally encumbered MIT
+license and under the Apache2 license for those entities concerned about possible
+litigation issues.
```
jrvanwhy, line 48: I don't think we should explain *why* a user should prefer one license over the other. IMO that gets too close to sounding like legal advice.

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,188 @@
...
+http://opensource.org/licenses/MIT).
+
+All contributions to Tock projects MUST be licensed under these terms.
+
+
+3. Copyright
+============
+
+Entities contributing resources to open-source projects often require attribution
+to recognize their efforts. Copyright is a common means to provide this. For
```
jrvanwhy, line 62: Was this supposed to say "Copyright notices are a common means to provide this."? "Copyright" is a much broader term than "Copyright notices".

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,188 @@
...
+
+
+4. Implementation
+=================
+
+All code artifacts in Tock projects MUST include a license notice.
+
+Code artifacts MAY include copyright notice(s). Copyright notices MUST
+include a year. Newer copyright notices MUST be placed after existing
+copyright notices. If non-trivial updates are performed by an original
```
jrvanwhy, line 83: How do we deal with files like [`chips/nrf52/src/crt1.rs`](https://github.com/tock/tock/blob/0f7fe7d9355002e4e7065cc3fb940a0d121e8d21/chips/nrf52/src/crt1.rs) that have a copyright notice in the middle of the file?

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,188 @@
...
+copyright author, they MAY amend the year(s) indicated on their existing
+copyright statement or MAY add an additional copyright line, at their
+discretion.
+
+
+4.1 Format
```
jrvanwhy, line 89: The entire "Format" section makes it difficult to re-use code from other projects, as their license and copyright information may not match our requirements, and I do not think we can arbitrarily reformat those notices.

An example of a file that does not match this format is [`capsules/src/sip_hash.rs`](https://github.com/tock/tock/blob/0f7fe7d9355002e4e7065cc3fb940a0d121e8d21/capsules/src/sip_hash.rs).

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,188 @@
...
+
+> Copyright {entity} {year}([,year],[-year]).
+
+The `{entity}` field is REQUIRED and should reflect the entity wishing
+to claim copyright. The `{year}` field is REQUIRED and should reflect
+when the copyright is first established. Substantial updates in the
+future MAY indicate renewed copyright, via additional comma-separated
+years or via range syntax, at the copyright holder’s discretion. The
+initial year SHALL NOT be removed unless it is the express intent of
+the copyright holder to relinquish the initial copyright.
```
jrvanwhy, line 134: I'm not sure that "relinquish the initial copyright" is something you can *do* in many jurisdictions. Probably better to leave this half of the sentence out.

doc/reference/trd-legal.md
alevy: Agree. Just say less

doc/reference/trd-legal.md
alevy: We can move that copyright notice

alevy:
We're going to take another pass on this to condense it and remove some unnecessary requirements.

Unlike most other PRs, let's hold off on merging after approvals in order to get feedback from external contributors first.

But please approve once you're comfortable so we _can_ be confident that what we get feedback on is similar to what we'll merge.

(draft coming later today)

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,188 @@
...
+License and copyright information MUST appear near the top of a file.
+Generally, license and copyright information SHOULD be the first lines
+in a file, however exceptions may be made where technically necessary
+(e.g. to accommodate the shebang in a shell script).
+
+License and copyright information MUST have exactly one (1) blank line
+separating it from any other content in the file.
+
+License information MUST come before copyright information.
+There MUST NOT be any whitespace between lines in the license/copyright block.
```
phil-levis, line 101: I think we want to avoid specifying this at such a low level. E.g., the standard format of a BSD license in a comment block has lines of whitesapce between the enumerated elements. That is, remove this MUST NOT.

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,188 @@
+Licensing and Copyrights
+========================================
+
+**TRD:** <br/>
+**Working Group:** Core<br/>
+**Type:** Documentary<br/>
```
phil-levis, line 6: Best Common Practice?

ppannuto:
Quick comment re the formatting bits: the pedantic, low-level bits were written with an eye towards making Johnathan's life easier re: automatic enforcement. I do think top of file is important and clustering licenses and copyright; the rest I have less attachment to.

jrvanwhy:
> Quick comment re the formatting bits: the pedantic, low-level bits were written with an eye towards making Johnathan's life easier re: automatic enforcement. I do think top of file is important and clustering licenses and copyright; the rest I have less attachment to.

In this respect, I think there are two distinct groups of license notices:

**Tock project notices:** Notices added by Tock contributors when they contribute to the Tock project.
**3p notices:** Notices in code that the Tock project copied from another project, such as the `sip_hash.rs` notice.

When I commented on the Format section, I assumed this policy would apply to both sets of license notices. I think that being picky about 3p notices makes it difficult to import code from other projects.

We could limit our formatting requirements and automated enforcement to Tock project notices. Then we could import code from other projects as long as that code is also (Apache 2.0 OR MIT)-licensed, while remaining picky about *most* of the license headers in the repository.

We'll probably end up with an exceptions list in the tool. For example, if someone copies a file from another project without making any changes to it, then it should be committed without a Tock project notice. In that case, we'll need to list the file as an exception in the tool. When we eventually make our own contributions to that file, we will then add a Tock project notice and remove the license exception.

jrvanwhy:
[Rendered view](https://github.com/tock/tock/blob/doc-copy-license/doc/reference/trd-legal.md)

jrvanwhy:
> We'll probably end up with an exceptions list in the tool. For example, if someone copies a file from another project without making any changes to it, then it should be committed without a Tock project notice. In that case, we'll need to list the file as an exception in the tool. When we eventually make our own contributions to that file, we will then add a Tock project notice and remove the license exception.

I just finished typing up the [notes](https://github.com/tock/tock/pull/3326) from today's core WG call. Having listened to the discussion again, I've changed my mind slightly on this.

If we copy a file from another (Apache 2.0 OR MIT) project into Tock, I think it is totally fine to add something like:

```
// Licensed under the Apache License, Version 2.0 or the MIT License.
// SPDX-License-Identifier: Apache-2.0 OR MIT
```

at the top of the file. It's the copyright line that we shouldn't add if we didn't contribute to the file.

However, I do think we should allow some variation in the format if we use a file from another project. For example, if we copy in a file from another project with the following header:

```
// Copyright ExampleCo
// Licensed under the Apache License, Version 2.0 or the MIT license, at your option.
// SPDX-License-Identifier: Apache-2.0 OR MIT
```

I think we should leave it intact, rather than requiring us to change it to:

```
// Licensed under the Apache License, Version 2.0 or the MIT License.
// SPDX-License-Identifier: Apache-2.0 OR MIT
// Copyright ExampleCo
// Licensed under the Apache License, Version 2.0 or the MIT license, at your option.
// SPDX-License-Identifier: Apache-2.0 OR MIT
```

and the enforcement tool should be flexible enough to make that easy.

brghena (approved):

alevy (dismissed):

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,183 @@
...
+to an artifact. Copyright notices SHOULD NOT be removed by any party except
+the original copyright holder.
+
+For full authorship information, see the version control history.
+
+
+4. Implementation
+=================
+
+All code artifacts in Tock projects MUST include a license notice.
```
jrvanwhy, line 75: What about files that do not support comments? Markdown comes to mind -- I suspect we want to omit license notices in our Markdown documents.

jrvanwhy (commented):
Based on my reading of the document, I think the enforcement tool should:

1. Verify the license information is present near the top of each file and formatted according to this policy.
1. NOT look for license information elsewhere in the file, and ignore any that is present.

With this implementation, we could copy code from other projects with license/copyright information in the middle of the file without adding an exception in the tool. The drawback is that it would not catch errant license headers in the middle of new contributions.

What do you think?

doc/reference/trd-legal.md
jrvanwhy: I suspect that is fine.

hudson-ayers: It sounds like you two have different answers to the question "is markdown code?"

jrvanwhy: I'm not sure that either "Markdown vs not-Markdown" or "code vs not-code" is the right place to make the distinction. How about requiring headers in all file formats that support comments? I don't want someone to think that images have to have license info embedded in their EXIF tags, so maybe all *textual* file formats that support comments.

I don't particularly care where we draw the line, I just think we need to be unambiguous.

alevy: Sure, that seems fine to me.

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,183 @@
...
+
+
+4.1 Format
+----------
+
+License and copyright information SHOULD appear near the top of a file.
+Generally, license and copyright information SHOULD be the first lines in a file
+unless necessary for technical (e.g. to accommodate the shebang in a shell
+script) or other reasons (e.g. importing a file with an existing copyright
+notice in a different location).
```
lschuermann, line 92: We may want to use this "loophole" to avoid having the license specification be the first lines of a file which also includes a module-level documentation comment. In my (rather subjective) opinion, something like the following actually looks rather odd to people who expect module documentation to always be at the top of a file (which is motivated by Rust not allowing any actual code before such a parent-module doc comment). This would be a generous interpretation of _technical or other reasons_:

```
// Licensed under the Apache License, Version 2.0 or the MIT License.
// SPDX-License-Identifier: Apache-2.0 OR MIT
// Copyright Jane Doe <jdoe@example.org> 2019,2022

//! Provides userspace with access to a serial interface.
//!
//! Setup
//! -----
//!
//! You need a device that provides the `hil::uart::UART` trait.
```

I am aware this conflicts with the example provided below. Not trying to start a flamewar about code style here, just my 2ct.

lschuermann (approved):
This looks pretty good to me.

doc/reference/trd-legal.md
jrvanwhy: I think I interpret "necessary" different than you do. If we want the headers to go below module documentation, then we should state that explicitly.

lschuermann: Oh, yeah, I agree with you. My comment is a convoluted way to ask the question "to me this looks weird, should we do this instead?". If we come to the conclusion that license specs should indeed follow doc-comments or even be part of them (at the end, which I've seen throughout the ecosystem), we might want to change the TRD.

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,183 @@
...
+
+2. Licensing
+============
+
+All software artifacts under the umbrella of the Tock project are
+dual-licensed as Apache License, Version 2.0 (LICENSE-APACHE or
+http://www.apache.org/licenses/LICENSE-2.0) or MIT license (LICENSE-MIT or
+http://opensource.org/licenses/MIT).
+
+All contributions to Tock projects MUST be licensed under these terms.
```
alistair23, line 50: This doesn't seem right.

What about tools? What if there are tools that run on the host. For example scripts that we might use to run tests, checks or flash binaries.

Also, what about newlib which is used by libtock-c? The binaries are [committed to a Tock project](https://github.com/tock/libtock-c/tree/master/newlib/cortex-m/v7-m)

doc/reference/trd-legal.md
hudson-ayers: I realize this was already updated from MUST to SHOULD, but I think we should still say "at least one". For example, [PEP 8 requires top-level functions and classes be surrounded by two blank lines](https://peps.python.org/pep-0008/#blank-lines), so most Python files without other comments at the top of the file would want two spaces after this information.

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,183 @@
...
+provide this. For downstream users, the license terms of Tock ensures
+unencumbered use.
+
+Copyrights in Tock projects are retained by their contributors. No
+copyright assignment is required to contribute to Tock projects.
+
+Artifacts in the Tock project MAY include explicit copyright notices.
+Substantial updates to an artifact MAY add additional copyright notices
+to an artifact. Copyright notices SHOULD NOT be removed by any party except
+the original copyright holder.
```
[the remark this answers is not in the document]
hudson-ayers: Looking...*very* far into the future, one reason for a non-author to remove a copyright would be if the copyright has expired. 

Regardless, I don't think we should interpret the use of `SHOULD NOT` as implying `MAY`, it is just a hedge that we may not have considered every possible case. Obviously any removal of a copyright notice is something that should get attention in a code review.

doc/reference/trd-legal.md
hudson-ayers: I do think it is a good question whether this should actually apply to all Tock project code -- for example, I don't think we want to go and add copyright headers to every file in the tock-teensy repo. Similarly, a tool like tockloader is a published artifact, and even most corporate users will not be depending on any of its source files directly -- they just might invoke the tool at some point during the process of building/flashing.

@jrvanwhy for Google's purposes, do you know what subset of Tock project repositories need to have license headers? For example, would just the kernel be sufficient? Or the kernel + libtock-c? I think we should limit the scope of this as much as possible.

doc/reference/trd-legal.md
alevy: Resolution from call: "SHOULD have *at least one*"

doc/reference/trd-legal.md
alevy: Conclusion: Textual files that allow comments

(e.g. Markdown files don't support comments, so we don't expect it)

doc/reference/trd-legal.md
alevy: Conclusion: redundant since we're going to be more liberal with code imported from third-parties. Normal practice would be license header above module docs.

doc/reference/trd-legal.md
alevy: Conclusion 1: It's a lot easier if everything in a repository is under the same/compatible licenses. Let's keep it to that for Tock.

Conclusion 2: For other repositories, let's punt with this document. I.e. change the language to refer _specifically_ to _this_ Tock repository, not the Tock "project" as a whole.

alevy:
Additional conclusions:
- The expectations for formatting are about the Tock notices, other notices don't have to follow these requirements.
- If notices different enough to modify then instead create a new one above it, but should be flexible to accept headers that are close enough.

bradjc:
I fixed the header formatting.

jrvanwhy:
> Overall this approach looks good to me. Curious to hear what @jrvanwhy in particular thinks about not requiring the "Copyright Tock Contributors" line on all files without a different copyright assignment

I just noticed I never answered this question. Personally, I'm not a fan of making the copyright notice optional. Making the copyright notice optional will tend to overrepresent the contributions of some contributors relative to others.

For example, suppose contributors A, B, C, D, and E work together on a file (all contributing significantly), but the only contributor who wants to add a copyright notice is A. In that case, the license header will become:

```
# Licensed under the Apache License, Version 2.0 or the MIT License.
# SPDX-License-Identifier: Apache-2.0 OR MIT
# Copyright A 2022.
```

Someone who is *not* from the Tock project who stumbles across this file will see it as primarily A's work, rather than a community project.

Requiring the "Copyright Tock Contributors" line would change that to:

```
# Licensed under the Apache License, Version 2.0 or the MIT License.
# SPDX-License-Identifier: Apache-2.0 OR MIT
# Copyright Tock Contributors 2022.
# Copyright A 2022.
```

which is a more accurate representation of the code's authorship.

lschuermann:
I share the spirit of @jrvanwhy's comment. Yet, I'm unsure whether requiring such a line might not introduce some issues: obviously we'd not add such a line on a compatibly-licensed, vendored file. But even when a file is added e.g. by its original author to Tock, I think it might be weird to force this developer to attribute the wider "Tock contributors" despite the file not having received any contributions other than the original author's (yet). I think having a "Copyright Tock Contributors" line in each file as a convention is good, and I would adapt that, but I'm wary about enforcing it. I presume every contributor would be able to decide on their own whether to attribute just themselves, the "Tock contributors", or both.

jrvanwhy:
We could require the "Copyright Tock Contributors" line only in cases where non-trivial contributions are made by contributors who do not have their own copyright lines. In practice, I think this would result in most files carrying only the "Copyright Tock Contributors" line (because most of y'all seem to prefer that over per-person copyrights), and a handful of files having 1 or 2 specific contributors named.

The downside is that "non-trivial contributions" is a bit vague and therefore this cannot be automatically enforced, but I don't expect that will be an issue in practice.

hudson-ayers:
> We could require the "Copyright Tock Contributors" line only in cases where non-trivial contributions are made by contributors who do not have their own copyright lines. In practice, I think this would result in most files carrying only the "Copyright Tock Contributors" line (because most of y'all seem to prefer that over per-person copyrights), and a handful of files having 1 or 2 specific contributors named.
> 
> The downside is that "non-trivial contributions" is a bit vague and therefore this cannot be automatically enforced, but I don't expect that will be an issue in practice.

I disagree that this will not be an issue in practice. What will happen without automatic enforcement is there will be tons of files with "Copyright FIRST_AUTHOR" at the top that have in fact been modified by dozens of other people. For example, Phil used to often add an `Author: Philip Levis` line to the top of files he created. If you grep for that in the Tock repo, you will find dozens of files with just "Author: Philip Levis" or maybe "Author: Philip Levis, Leon Schuermann" where looking at the git history will reveal many other authors with significant contributions.

I think it is much easier in practice to just require "Copyright Tock Contributors". If an individual author wishes to add an additional copyright for just their own name they are free to do so.

alevy:
I'm going to convert this from a Draft PR to a real PR. I _believe_ the remaining issue is just whether to require a copyright line or make it optional. Please comment if there is more missing still.

alevy (dismissed):

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,178 @@
...
+
+The `{entity}` field should reflect the entity wishing to claim copyright. The
+`{year}` field SHOULD reflect when the copyright is first established.
+Substantial updates in the future MAY indicate renewed copyright, via additional
+comma-separated years or via range syntax, at the copyright holder’s discretion.
+The initial year SHALL NOT be removed unless it is the express intent of the
+copyright holder to relinquish the initial copyright.
+
+
+### 4.1.1 Examples
```
jrvanwhy, line 125: The Examples section needs to be updated to indicate that a copyright line is necessary.

We seem to have come to the conclusion (see [Hudson's comment](https://github.com/tock/tock/pull/3318#issuecomment-1324225961)) that we should require "Copyright Tock Contributors" to be one of the copyright lines, or at least make it standard, so it should probably be part of the examples as well.

lschuermann: At the risk of repeating what I said above, I’d vouch for making it standard, as I foresee issues with requiring this line for files added without modification by a “Tock Contributor”. Otherwise I agree that this section should be updated.

doc/reference/trd-legal.md
```diff
@@ -0,0 +1,178 @@
...
+copyright notices.
+
+For full authorship information, see the version control history.
+
+
+4 Implementation
+================
+
+All textual files that allow comments MUST include a license notice and
+copyright notice(s).
```
jrvanwhy, line 77: From the 2023-01-13 core wg call: This line needs an exception for files that Tock contributors have not made contributions for (e.g. vendored files).

jrvanwhy: Merged #3377 which adds an exception for vendored and other imported files.

doc/reference/trd-legal.md
jrvanwhy: Merged #3377 which updated the examples. `Copyright Tock Contributors` is part of the examples but is not required in the text.

doc/reference/trd-legal.md
jrvanwhy: This PR has been updated to only apply to the `tock` repository. We expect to discuss applying this policy to other repositories after this PR has been merged.

doc/reference/trd-legal.md
jrvanwhy: I changed it to "at least one".

hudson-ayers (approved):

lschuermann (approved):

brghena (approved):

alevy (approved):

PR #3322, opened 2022-11-11
arch/rv32i/e|pmp: return on addr unalignment 

twilfredo:
### Pull Request Overview

Instead of rounding up an un-aligned e/pmp start address, lets return an
error specifying we can't carry out the specified work at this address,
so return `None`.

It is then upto the board to specify the correct aligned address to
configure a pmp region.

### Testing Strategy

Compile and load kernel on Verilator.

### TODO or Help Wanted

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/epmp.rs
+++ b/arch/rv32i/src/epmp.rs
@@ -575,22 +575,13 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::MPU
         let region_num = config.unused_region_number(self.locked_region_mask.get())?;
 
         // Logical region
-        let mut start = unallocated_memory_start as usize;
-        let mut size = min_region_size;
+        let start = unallocated_memory_start as usize;
+        let size = min_region_size;
 
-        // Region start always has to align to 4 bytes
-        if start % 4 != 0 {
-            start += 4 - (start % 4);
-        }
-
-        // Region size always has to align to 4 bytes
-        if size % 4 != 0 {
-            size += 4 - (size % 4);
-        }
-
-        // Regions must be at least 8 bytes
-        if size < 8 {
-            size = 8;
+        // Region start and size always has to align to 4 bytes
+        // and size must be at-least 8 bytes
+        if start % 4 != 0 || size % 4 != 0 || size < 8 {
+            return None;
         }
 
         let region = PMPRegion::new_app(start as *const u8, size, permissions);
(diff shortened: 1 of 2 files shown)


hudson-ayers (commented):
What is the motivation for this change?

twilfredo:
> What is the motivation for this change?

I was thinking we include the initial address requested in a case where it is not 4-byte aligned and gets rounded up. So with this, we round down with a downside being exposing slightly more memory to the region but allowing the initially requested address to be in this region also.

hudson-ayers:
Note: if we do change this for `allocate_kernel_region()` we need to also change `allocate_region()` to mirror it.

On one hand, I agree conceptually that a client of this API requesting memory starting at a specific address probably wants that address to be a part of the memory block that it gets handed.

On the other hand, this behavior could be a little surprising -- in general, I would not expect to get handed a region starting before the address I requested, and it seems...vaguely possible that this surprising behavior could somehow make bugs with overlapping regions more likely (though I will be honest that I am having trouble thinking of a specific way that could happen, since the end of the region also has to be aligned in any case).

Interested to hear thoughts from someone who has written code that interacts with this API

alistair23:
> On one hand, I agree conceptually that a client of this API requesting memory starting at a specific address probably wants that address to be a part of the memory block that it gets handed.

Exactly! I think if we can't meet this requirement we should return an error. We can't do what has been asked of us.

What we currently do seems like a bug, where we just ignore the requested address and round up (if it's not aligned). From memory the PMP spec allows larger alignment requirements (depending on the hardware). So on some boards this might be a substantial difference between what was requested and what we provide.

> On the other hand, this behavior could be a little surprising -- in general, I would not expect to get handed a region starting before the address I requested, and it seems...vaguely possible that this surprising behavior could somehow make bugs with overlapping regions more likely (though I will be honest that I am having trouble thinking of a specific way that could happen, since the end of the region also has to be aligned in any case).

I agree, I think we should remove the alignment fixes entirely and just return an error (or `None`). It can then be up to the board to perform any alignment adjustments it wants

twilfredo:
I've implemented changes to return `None` when not aligned, feel like we should do the same here? https://github.com/tock/tock/blob/5da9c41bc8ef048e0c18eacc1a60f6c5dcda07d4/arch/cortex-m/src/mpu.rs#L424

alistair23 (approved):

hudson-ayers:
@bradjc thoughts?

I think I agree with Alistair's reasoning, but it does make sense that we want the MPU and PMP behavior to match with regards to whether we are going to accept rounding.

The doc comments for the MPU trait do not mention any shifting for alignment as being allowed:
```rust
    /// Allocates a new MPU region.
    ///
    /// An implementation must allocate an MPU region at least `min_region_size`
    /// bytes in size within the specified stretch of unallocated memory, and
    /// with the specified user mode permissions, and store it in `config`. The
    /// allocated region may not overlap any of the regions already stored in
    /// `config`.
    ///
    /// # Arguments
    ///
    /// - `unallocated_memory_start`: start of unallocated memory
    /// - `unallocated_memory_size`:  size of unallocated memory
    /// - `min_region_size`:          minimum size of the region
    /// - `permissions`:              permissions for the region
    /// - `config`:                   MPU region configuration
    ///
    /// # Return Value
    ///
    /// Returns the start and size of the allocated MPU region. If it is
    /// infeasible to allocate the MPU region, returns None.
```

Returning `None` when we cannot begin a region at exactly the requested start address seems in-line with that doc comment. Do you know if our current implementation automatically aligning is currently necessary for some particular reason?

bradjc (changes requested):
Where does the caller request a specific start address? And where does that comment specify any guarantee on start address?

The API is designed _specifically_ to support padding for alignment or any other reason within the available memory. The caller has to provide the range of memory that is available, and the implementation finds some region which does not overlap, is enforceable, and is at least `min_region_size` in bytes.

twilfredo:
> Where does the caller request a specific start address? And where does that comment specify any guarantee on start address?
> 
> The API is designed _specifically_ to support padding for alignment or any other reason within the available memory. The caller has to provide the range of memory that is available, and the implementation finds some region which does not overlap, is enforceable, and is at least `min_region_size` in bytes.

Ah, I didn't realize that this designed specifically with padding, in this case, I don't this this PR is relevant anymore (?) if so, feel free to close.

PR #3325, opened 2022-11-11
Fix MapCell safety

LawrenceEsswood:
MapCell is unsafe as it does not distinguish between an empty cell and a full cell that is borrowed.
It also fails to drop its contents if put is called. I have fixed these issues and added in a more convenient try_borrow_mut to bring the interface closer in line with RefCell.

Example breaking the old implementation:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=414f6a3ec3b3578db1308d0e55169e8f

### Pull Request Overview

MapCell implementation

### Testing Strategy

Building and running a hello world example on one of the boards

### TODO or Help Wanted

Review.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.


LawrenceEsswood:
I would like to bring a particular note to the reviewer on my decision "replace" and "put" trapping, but make "take" return None.

My rationale is that users of the library are already handling failure to take a value in a sensible way, and treating "empty" and "full but borrowed" the same way makes sense.

There is no way for put (currently) to indicate failure, and nothing sensible to do for replace. 

Possibly you will think that take should also panic if the cell is borrowed, I have no strong feelings.

hudson-ayers:
To summarize the changes to the interface:

- `MapCell::take()` will now return `None` if called on a `MapCell` which is currently borrowed
- `MapCell::put()` will panic if called on a `MapCell` which is currently borrowed
- `MapCell::replace()` will panic if called on a `MapCell` which is currently borrowed
- adds `MapCell::try_borrow_mut()` which is a helper method that uses a new type with a `Drop` implementation to allow code to borrow from inside a mapcell without having to use closures
- adds roughly 300/500 bytes to imix/OT respectively, probably mostly as a result of the extra panics added

Did I miss anything?

LawrenceEsswood:
Thats everything.

bradjc:
First of all, thank you for finding/noticing this.

It seems like from a MapCell usage standpoint, there are two issues:

1. Once inside `map()`, the caller has a mutable reference, so calling `put()` or `replace()` on the same MapCell is odd/unnecessary.
2. Once inside `map()`,  the nested call to `map()` shouldn't be allowed, as the caller already has the mutable reference.

We've run into this general issue before with grants, where entering the same grant twice simultaneously is unsafe. Our fix there was to panic if capsule attempts to enter the same grant again.

My thought is to follow that same logic and handle the error when .map() is called twice. I.e., attempting to get the mutable reference twice is the fundamental issue. I believe in that case since `map()` already returns an `Option` we can just not call the closure and return `None`, as would happen if `.map()` was called when there was nothing in the MapCell.

There is still the question of what to do with `.replace()`. I propose we just remove it. I _think_ we only ever use it in Tock when `.put()` would work just as well, and I believe it can still be implemented manually with `.take()` and `.put()`.

LawrenceEsswood:
I believe, as it stands, the refactor will do what you suggest with respect to ```map```. Possibly the sneaky inline ```?``` makes that not so obvious.

Having just a ```put``` still has the unfortunate side effect of either panicking if the cell is full, or (I think worse) silently dropping the value. I don't think ```replace``` is better or worse than ```put```. They are pretty similar in the end, and implementing one from the other is not hard.

If anyone wants to avoid the panic maybe offering a ```try_put``` is the answer.

hudson-ayers:
> Having just a `put` still has the unfortunate side effect of either panicking if the cell is full, or (I think worse) silently dropping the value. I don't think `replace` is better or worse than `put`. They are pretty similar in the end, and implementing one from the other is not hard.

Just including one in the library does have the advantage of only one panic!() (and associated size from message / location data) in the final binary.

> If anyone wants to avoid the panic maybe offering a `try_put` is the answer.

I would be in favor of this.

libraries/tock-cells/src/map_cell.rs
```diff
@@ -80,27 +131,67 @@ impl<T> MapCell<T> {
...
                 let result: MaybeUninit<T> = ptr::replace(self.val.get(), MaybeUninit::new(val));
                 // `result` is _initialized_ and now `self.val` is now a new uninitialized value
                 Some(result.assume_init())
+            },
+            InitBorrowed => {
+                // I know the point of MapCell was possibly also to reduce the source of panics,
+                // but this always unsafe. The description says it returns the value in the MapCell,
+                // so defaulting to just returning the argument might lead to unintended behavior.
+                // I say that any (old) code that tries to replace a borrowed value just fault,
+                // it is likely that it was broken anyway.
```
hudson-ayers, line 162: Before we merge, we should update this comment to something like

```
// Trying to replace a borrowed value is invalid use of the API
```

And we should also update the doc comment on this method to indicate as such.

libraries/tock-cells/src/map_cell.rs
```diff
@@ -80,27 +131,67 @@ impl<T> MapCell<T> {
...
+                panic!();
```
hudson-ayers, line 163:
```suggestion
                panic!("MapCell: Can't replace borrowed value");
```

bradjc:
> Having just a `put` still has the unfortunate side effect of either panicking if the cell is full, or (I think worse) silently dropping the value. 
> 
> If anyone wants to avoid the panic maybe offering a `try_put` is the answer.

I'm thinking from the other perspective, if I want to try to ensure that code (like capsules) cannot use a library with only safe APIs that can subtly panic the system, then I only want to provide `try_put()`, and having `put()` silently fail _is_ better because it only breaks the buggy code rather than crash the system.

> I don't think `replace` is better or worse than `put`. They are pretty similar in the end, and implementing one from the other is not hard.

What does `replace()` return if it is going to fail? What makes `put()` easier is it doesn't have to return anything. If `replace()` returned on `Option` then it could return `None`, but then `.replace()` wouldn't really be better than `.take()` + `.put()`.

LawrenceEsswood:
> If replace() returned on Option then it could return None

Replace already does return an option. Documentation says it is meant to be anything that was previously there, but we can always change the documentation to say that it can also return what was provided if the ```replace``` fails. If we would like to avoid panic (and  are happy for silent failure), we could always return the input, wrapping with a ```Some()```. I think I prefer this to returning ```None``` as at least then a caller _could_ do something with the return value. ```Put``` is then just ```replace```, but silently dropping whatever happens to result. This would leave ```Put``` and ```Replace``` slightly misnamed, but would cause least disruption for existing code, would not panic, and any new code that wants to initialize a ```MapCell``` can do replace and compare the result with ```None``` to check they have correctly done so.

phil-levis:
Let's discuss this on the core call this week? It would be great to resolve it or at least have a clear next step.

libraries/tock-cells/src/map_cell.rs
```diff
@@ -17,23 +65,23 @@ pub struct MapCell<T> {
     // `.occupied` before calling `.val.get()` or `.val.assume_init()`. See
     // [mem::MaybeUninit](https://doc.rust-lang.org/core/mem/union.MaybeUninit.html).
     val: UnsafeCell<MaybeUninit<T>>,
```
kupiakos, line 67: This does not drop a contained `T` and instead drops it early during `replace`

libraries/tock-cells/src/map_cell.rs
```diff
@@ -80,27 +131,67 @@ impl<T> MapCell<T> {
...
     /// empty, the previous value is returned, otherwise `None` is returned.
     pub fn replace(&self, val: T) -> Option<T> {
-        if self.is_none() {
-            self.put(val);
-            None
-        } else {
-            unsafe {
+        match self.occupied.get() {
+            Uninit => {
+                unsafe {
```
kupiakos, line 143: `self.occupied` is not updated

kupiakos (commented):
This library is in _desperate_ need of unit tests - there are additional bugs such as `MapCell<T>` not dropping a held `T`. Can you please add unit tests, including the UB playground you have provided?

hudson-ayers:
@LawrenceEsswood do you plan to adjust the changes suggested by @kupiakos ? If not, someone from the Tock core team can try to take this PR over and get it across the finish line

libraries/tock-cells/src/map_cell.rs
```diff
@@ -80,27 +131,67 @@ impl<T> MapCell<T> {
...
+    /// use tock_cells::map_cell::MapCell;
+    /// let cell = MapCell::new(1234);
+    /// let x = &cell;
+    /// let y = &cell;
+    ///
+    /// *x.try_borrow_mut().unwrap() += 1;
+    ///
+    /// assert_eq!(y.take(), Some(1235));
+    /// ```
+    pub fn try_borrow_mut(&self) -> Option<MapCellRef<T>> {
```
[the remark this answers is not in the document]
LawrenceEsswood: There are two possible errors that could actually occur here (being un-init and already borrowed). I've gone for a shared error type with try_replace even though try_replace can only have one of the errors occur.

libraries/tock-cells/src/map_cell.rs
LawrenceEsswood: Fixed.

libraries/tock-cells/src/map_cell.rs
LawrenceEsswood: Fixed

LawrenceEsswood:
Changes:

replaced some of the Options<> with explicit errors
added try_replace as non-panic option (try_borrow already was)
fixed drop
fixed the error with forgetting to set state to init in replace
added some tests

libraries/tock-cells/src/map_cell.rs
```diff
@@ -80,24 +156,66 @@ impl<T> MapCell<T> {
...
     /// Replaces the contents of the `MapCell` with `val`. If the cell was not
     /// empty, the previous value is returned, otherwise `None` is returned.
-    pub fn replace(&self, val: T) -> Option<T> {
-        if self.is_none() {
-            self.put(val);
-            None
-        } else {
-            unsafe {
+    /// In the event the cell is currently borrowed, returns Err(AlreadyBorrowed)
+    pub fn try_replace(&self, val: T) -> Result<Option<T>, MapCellErr> {
```
kupiakos, line 166: Can you add tests for `try_replace` as well?  Exercising all code paths

libraries/tock-cells/src/map_cell.rs
```diff
@@ -17,23 +76,37 @@ pub struct MapCell<T> {
...
+
+impl<T> Drop for MapCell<T> {
+    fn drop(&mut self) {
+        let state = self.occupied.get();
+        debug_assert!(state != InitBorrowed);
+        if state == Init {
+            unsafe {
+                // Safety: state being Init means that the MaybeUninit data was initted.
+                // the pointer to the data can never be used again as this
+                drop_in_place(self.val.get_mut().as_mut_ptr())
```
kupiakos, line 90: This code isn't exercised in any tests; can you add a `MapCell<String>` test?

libraries/tock-cells/src/map_cell.rs
```diff
@@ -2,7 +2,66 @@
 
 use core::cell::{Cell, UnsafeCell};
 use core::mem::MaybeUninit;
+use core::ops::{Deref, DerefMut};
 use core::ptr;
+use core::ptr::drop_in_place;
+use map_cell::MapCellErr::AlreadyBorrowed;
```
kupiakos, line 8: nit: it'd be more readable to use `self` here and on the line below

LawrenceEsswood: Sure!

libraries/tock-cells/src/map_cell.rs
LawrenceEsswood: It is exercised in test_drop. Just checked that removing this line does indeed cause that test to fail

libraries/tock-cells/src/map_cell.rs
LawrenceEsswood: I added a further test.

hudson-ayers:
I think we should get this PR in, the soundness hole has been open too long -- one downside is that it increases flash use on imix and OT by about 1.5 kB, which is a lot. 

I just added a commit which moves the replace() panic into a non-monomorphized function, which gets us back 312 bytes, so now the increase is about 1.2 kB (still kind of a lot, but a bit better)

hudson-ayers (approved):

alistair23 (approved):

libraries/tock-cells/src/map_cell.rs
```diff
@@ -80,24 +162,69 @@ impl<T> MapCell<T> {
...
+    /// let cell = MapCell::new(1234);
+    /// let x = &cell;
+    /// let y = &cell;
+    ///
+    /// *x.try_borrow_mut().unwrap() += 1;
+    ///
+    /// assert_eq!(y.take(), Some(1235));
+    /// ```
+    #[inline]
+    pub fn try_borrow_mut(&self) -> Result<MapCellRef<T>, MapCellErr> {
```
kupiakos, line 220: ...do we need this? I'm glancing at the CL and don't see a previous equivalent `borrow_mut`. It should simplify the necessary code in `try_replace`

bradjc (commented):
I'm not blocking this, I guess, but I don't see anything in this PR conversation that makes a strong case for adding a panic. Our discussion from before also seemed to steer away from panics, but I was certainly participating in that conversation... (https://github.com/tock/tock/blob/master/doc/wg/core/notes/core-notes-2022-12-16.md#fixing-mapcell-safety-3325).

I still believe it is better for capsules to silently fail than to provide a trivial way for capsules to call panic! whenever they want without directly calling `panic!()`.

lschuermann:
I might not get the full picture here, but the `MapCell` documentation states that "[you] can think of a `MapCell` as an `Option` wrapped in a `RefCell`". Looking at all the unsafe code in this file, it seems that we're building something awfully complex for such a helper.

Why aren't we just building a small wrapper around `RefCell` in the first place? Superficially it appears that we're implementing more or less the exact same logic here, modulo some slightly different API which happens to integrate an inner `Option` as well. Here's something I cobbeled together in a Rust playground:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=237197b4a2fcedb1d91b4878fa1b94de

This should implement the exact same API, possible entirely without unsafe through use of the very convenient [`RefMut::map` API](https://doc.rust-lang.org/stable/core/cell/struct.RefMut.html#method.map).

lschuermann:
> @lschuermann `RefCell` has separate shared and mutable counters that `MapCell` doesn't need and would be wasteful to use.

I don't see any separate counters, just one `isize` that will be positive or negative depending on shared vs. mutable borrow. But alas, `RefCell<Option<usize>>` does take up 3 words, whereas this PR's `MapCell<usize>` uses 2. On a non-nullable type, `RefCell<Option<usize>>` also uses just 2 words; same as for `MapCell`. So technically, our custom implementation is more efficient memory-wise, at least in a subset of cases.

I'm not against the separate `MapCell` implementation per se, and definitely not against this PR. But seeing these soundness issues makes me uneasy, and I'd generally voice a strong opinion against re-inventing the wheel and using tried-and-tested code from the standard library, if possible.

libraries/tock-cells/src/map_cell.rs
hudson-ayers: I agree, I do not think that this needs to be a `pub` interface, and we do not call this externally. If a more efficient function could be made used by not making this `pub` that works for me.

alevy:
The purpose of `MapCell` (and `TakeCell` before it) rather than just using `RefCell` was to avoid the possibility of panics in favor of (as @bradjc suggested) potentially buggy capsules. If we're going to introduce `panic`, we may as well use `RefCell`.

These changes still avoid a `panic` on a `take` or `map`. Is it possible that there is use for a type that _only_ has `map` and no `take`, `put`, or `replace`? (or perhaps a `take` and only a `try_put`, `try_replace`)

kupiakos:
> Is it possible that there is use for a type that _only_ has `map` and no `take`, `put`, or `replace`?

After the discussion this morning, I think the perfect middle ground is to _not_ have `MapCell` follow `RefCell`, and instead be more like `Cell<Option<T>>` with an extra state and check to allow `map` to be sound against re-entrancy. We can have:
- `map` marks the cell as borrowed while the closure is running, in order to detect re-entrancy. Simplest is to just have `Cell<MapCellState>` as it is now.
- `map` does not call its closure if currently inside `map`, like it does for the empty state.
- `take` returns `None` and does not mutate if currently inside `map`, unlike the empty state.
- `put` drops its input and does not mutate if currently inside `map`, unlike the empty state.
- `replace` drops its input, does not mutate, and returns `None` if currently inside map, unlike the empty state.

We can also provide `try_` methods of the above that return `Err` if re-entrant, which I don't think most users should use but could be useful in isolated instances - what do you folks think?

This will still impact code size but *should* be a much smaller impact while allowing the same API to compile. In every case though, calling a mutator while inside `map` is a bug by the programmer. To aid in debugging and for unit tests to catch, when debug assertions are enabled, `map`, `take`, `put`, and `replace` all panic if run inside of `map`. It's a bit of a silent failure otherwise, unfortunately.

> I don't see any separate counters, just one `isize` that will be positive or negative depending on shared vs. mutable borrow. But alas, `RefCell<Option<usize>>` does take up 3 words, whereas this PR's `MapCell<usize>` uses 2.

You're correct @lschuermann, I was misremembering with `Rc` when it comes to size impact. But yes, we can also take advantage of combining the set and in-use states into the same value which `RefCell` cannot. That doesn't just save RAM size, the code to check should be simpler too.

kupiakos:
I've got a [draft implementation](https://github.com/kupiakos/tock/blob/1654dd975ea4797e354e138e774b976c4ae0900a/libraries/tock-cells/src/map_cell.rs) that implements the above with two major changes:

- There are no `try_` methods, but it does panic in debug mode, similar to how integer overflow is only ever a panic in debug mode.
- I added `get` for `T: Copy`, because why _shouldn't_ it be there? Granted, most users there should probably be using `Cell<Option<T>>`.

I'm not sure what I did or if this is correct, but it seems running `./tools/github_actions_size_changes.sh` tells me I _saved_ size? I've not investigated further, but it could be an unrelated change to the soundness check while I refactored elsewhere, or my setup is broken.

```
weact-f401ccu6: flash use -2468 bytes (-2.87%), RAM use -1640 bytes(-9.31%)
esp32-c3-board: flash use -2780 bytes (-3.75%), RAM use -1628 bytes(-21.21%)
litex_arty: flash use -2624 bytes (-3.22%), RAM use -1612 bytes(-10.83%)
swervolf: flash use -764 bytes (-1.38%), RAM use -1240 bytes(-22.47%)
earlgrey-cw310: flash use -3048 bytes (-2.17%), RAM use -1442 bytes(-7.60%)
litex_sim: flash use -192 bytes (-0.26%), RAM use -1224 bytes(-8.28%)
teensy40: flash use -864 bytes (-1.34%), RAM use -1240 bytes(-8.55%)
imix: flash use -1200 bytes (-0.68%), RAM use -1515 bytes(-5.12%)
hail: flash use -1412 bytes (-1.27%), RAM use -1596 bytes(-8.18%)
redboard_artemis_nano: flash use +64 bytes (+0.08%), RAM use -1189 bytes(-10.28%)
stm32f412gdiscovery: flash use -2040 bytes (-1.91%), RAM use -1628 bytes(-2.14%)
acd52832: flash use +632 bytes (+1.05%), RAM use -1217 bytes(-10.97%)
nrf52dk: flash use -2336 bytes (-2.62%), RAM use -1616 bytes(-14.46%)
nano33ble: flash use -636 bytes (-0.40%), RAM use -1560 bytes(-8.97%)
nrf52840dk: flash use -1028 bytes (-0.65%), RAM use -1560 bytes(-5.63%)
particle_boron: flash use -1812 bytes (-1.41%), RAM use -1564 bytes(-10.77%)
imxrt1050-evkb: flash use -2312 bytes (-2.78%), RAM use -1620 bytes(-9.97%)
microbit_v2: flash use -2548 bytes (-2.39%), RAM use -1904 bytes(-10.79%)
stm32f3discovery: flash use -1512 bytes (-1.50%), RAM use -1600 bytes(-8.80%)
stm32f429idiscovery: flash use -4272 bytes (-4.62%), RAM use -1728 bytes(-9.26%)
lora_things_plus: flash use -672 bytes (-0.79%), RAM use -3487 bytes(-25.09%)
msp-exp432p401r: flash use -292 bytes (-0.38%), RAM use -1252 bytes(-10.99%)
sma_q3: flash use -1556 bytes (-1.26%), RAM use -1516 bytes(-10.86%)
clue_nrf52840: flash use -860 bytes (-0.61%), RAM use -1540 bytes(-2.10%)
nrf52840_dongle: flash use -1988 bytes (-1.67%), RAM use -1588 bytes(-12.05%)
nucleo_f429zi: flash use -9972 bytes (-9.93%), RAM use -1790 bytes(-9.50%)
nucleo_f446re: flash use -2804 bytes (-3.20%), RAM use -1640 bytes(-9.52%)
raspberry_pi_pico: flash use -16016 bytes (-16.21%), RAM use -4229 bytes(-31.60%)
nano_rp2040_connect: flash use -15328 bytes (-15.01%), RAM use -4216 bytes(-31.44%)
pico_explorer_base: flash use -15664 bytes (-14.56%), RAM use -4140 bytes(-5.78%)
redboard_redv: flash use -2972 bytes (-3.82%), RAM use -1616 bytes(-18.97%)
qemu_rv32_virt: flash use -29056 bytes (-31.54%), RAM use -29608 bytes(-73.97%)
arty_e21: flash use -624 bytes (-0.99%), RAM use -1224 bytes(-15.13%)
hifive_inventor: flash use -3084 bytes (-3.98%), RAM use -1616 bytes(-13.97%)
hifive1: flash use -3740 bytes (-4.72%), RAM use -592 bytes(-7.87%)
```

kupiakos:
As discussed on the call, I've created a new PR with my changes: #3537

bradjc:
Fixed in https://github.com/tock/tock/pull/3537

PR #3327, opened 2022-11-12
Temporally order userspace print statements with kernel debug statements

phil-levis:
### Pull Request Overview

Currently it possible for userspace `printf`(or `putstr`) calls to be re-ordered with respect to kernel `debug!` calls. This means that a kernel `debug!` call made *before* a userspace print operation can appear later in the console output. This can make debugging difficult, as it is challenging to correctly correlate kernel and userspace debug output.

This problem is caused by how serial port output is queued. `debug` operations synchronously write to a circular queue, which is then asynchronously written to the serial port. The userspace `putstr` API is technically asynchronous, but `putstr` and `printf` yield until completion, so they appear synchronous to userspace.

There are two problems. First, when there are multiple writers, their ordering is independent of their arrival. The first writer will start writing, and subsequent writers are queued. The order of this queue depends on the order in which they were initialized (it's a fixed order linked list determined at initialization), not the order of operations. So if writers A, B and C write in that order, it is possible that A, C, B will be the output.

Second, the fact that debug uses a circular queue that's written out asynchronously breaks any ordering of its operations. For example, suppose the kernel calls `debug` for an upcall to a userspace process, but this can't be written because it's busy writing something else (e.g., another userspace printf). In the upcall, the userspace process calls `printf`. The code for `printf` issues another `debug` call. The text of these two debug calls (one from the upcall, one from the printf) will be written together when it's `debug''s turn to write, and the actual `printf` may occur either before or after them.

The basic issue is that kernel logging and userspace print statements use different UART clients, which write independently of one another.

This PR provides a ConsoleOrdered capsule (name should be different) that provides the same writing interface as `Console` but uses `debug` to write to the same circular buffer as kernel circular buffer. This means that kernel prints and userspace print operations have the correct ordering. The one edge case is that userspace print operations are queued and have an inserted delay until completion. Since `debug` is synchronous, there is no way to exert backpressure besides polling. The current approach simply writes as much as it can to the debug log, and then tries to write more after a timeout (which is configurable when instantiating the component). So this means that in real time it is possible for userspace to issue a printf system call, have that write delayed because the debug buffer was full, then see kernel `debug` operations that occured in wall time later appear earlier in the log. However, since the userspace process is blocked, none of these `debug` operations could be in response to its operations. In theory, the userspace process could have been rescheduled before issuing the `printf`. So ConsoleOrdered does not provide true wall clock time ordering of print operations, but assuming that userspace prints are blocking, it maintains causal ordering.

The system call driver maintains ordering between userspace printf operations by tracking a sequence number. Each call to the write system call increments a sequence number, which is then stored in that client. While a client (process) finishes sending, the driver looks for the client with a write pending that has the lowest sequence number. This handles wraparound for 32-bit values, under the assumption that there are at most 2^31 writes at any time.

The imix board has been configured to use this new capsule. The ConsoleOrdered capsule provides the same system call interface and device ID as the Console capsule.

### Testing Strategy

This pull request was tested on imix by running two userspace processes which both write to the serial port with `printf` in a loop. The printfs are large, such that `newlibc` breaks them into multiple writes. The capsule prints each write in entirety before interleaving another write. Adding `debug!` statements into the kernel has those statements appear in temporal order with respect to the userspace `printf` writes.

### TODO or Help Wanted

This needs to incorporate reading.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.


capsules/src/print_log.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// - `0`: Driver check.
+    /// - `1`: Transmits a buffer passed via `allow`, up to the length
+    ///        passed in `arg1`
+    fn command(&self, cmd_num: usize, arg1: usize, _: usize, appid: ProcessId) -> CommandReturn {
+        //debug!("PrintLog receiving command: {} {} from {:?}", cmd_num, arg1, appid);
+        let res = self
+            .apps
+            .enter(appid, |app, kernel_data| {
+                match cmd_num {
+                    0 => Ok(()),
```
bradjc, line 255: I don't think this really matters for this capsule, but I don't think we want anyone to copy this example, as command(0) should still return Ok() even if something went wrong with creating the grant.

kernel/src/debug.rs
```diff
@@ -555,6 +557,16 @@ pub fn debug_println(args: Arguments) {
     writer.publish_bytes();
 }
 
+pub fn debug_slice(slice: &ReadableProcessSlice) {
+    let writer = unsafe { get_debug_writer() };
+
+    for b in slice.iter() {
+        let buf: [u8; 1] = [b.get(); 1];
+        let _ = writer.write(&buf);
+    }
```
hudson-ayers, line 566: We could make this a bit more idiomatic by changing the definition of write() to take a `ReadableProcessSlice` instead of `&[u8]`, since any `&[u8]` can be cast for free to a `ReadableProcessSlice`

capsules/src/driver.rs
```diff
@@ -18,6 +18,7 @@ pub enum NUM {
     AnalogComparator      = 0x00007,
     LowLevelDebug         = 0x00008,
     ReadOnlyState         = 0x00009,
+    PrintLog              = 0x0000a,
```
hudson-ayers, line 21: This suggests that we plan to maintain both the `Console` and `PrintLog` implementations going forward -- is there any reason for `PrintLog` not to just replace `Console`? Alternatively, we could keep both capsules, but have them use the same driver number. Then board authors can choose whether they need in-order prints, but regardless can use the same userspace libraries

boards/components/src/debug_writer.rs
```diff
@@ -31,7 +31,7 @@ use kernel::hil::uart;
 // least a 1 KiB boundary). This is not _semantically_ critical, but helps keep buffers on 1 KiB
 // boundaries in some cases. Of course, these definitions are only advisory, and individual boards
 // can choose to pass in their own buffers with different lengths.
-pub const DEBUG_BUFFER_KBYTE: usize = 1;
+pub const DEBUG_BUFFER_KBYTE: usize = 8;
```
hudson-ayers, line 34: Hopefully this was just for debugging? This is a big increase

boards/components/src/print_log.rs
```diff
@@ -0,0 +1,84 @@
...
+//! let ast = &sam4l::ast::AST;
+//! let mux_alarm = components::alarm::AlarmMuxComponent::new(ast)
+//!     .finalize(components::alarm_mux_component_static!(sam4l::ast::Ast));
+//! ast.configure(mux_alarm);
+//! let print_log = components::print_log::PrintLogComponent::new(board_kernel, driver_num, mux_alarm)
+//!     .finalize(components::printlog_component_static!(sam4l::ast::Ast));
+
+//! ```
+// Author: Philip Levis <pal@cs.stanford.edu>
+// Last modified: 1/08/2020
```
hudson-ayers, line 24:
```suggestion
// Last modified: 11/29/2022
```
I realize these always end up out of date, but at least the first one should be right!

boards/imix/src/main.rs
```diff
@@ -143,6 +145,10 @@ struct Imix {
     >,
     nrf51822: &'static capsules::nrf51822_serialization::Nrf51822Serialization<'static>,
     nonvolatile_storage: &'static capsules::nonvolatile_storage_driver::NonvolatileStorage<'static>,
+    printlog: &'static capsules::print_log::PrintLog<
+        'static,
+        VirtualMuxAlarm<'static, sam4l::ast::Ast<'static>>,
+    >,
```
hudson-ayers, line 151: So now Imix exports both console and printlog .. is there a reason that a userspace app would want to use both at the same time?

capsules/src/print_log.rs
```diff
@@ -0,0 +1,276 @@
...
+            .get_readonly_processbuffer(ro_allow::WRITE)
+            .map_or(0, |write| write.len())
+            .min(len);
+        // We have nothing to write
+        if app.write_len == 0 {
+            return Err(ErrorCode::NOMEM);
+        }
+        // Order the prints through a global counter.
+        app.tx_counter = self.tx_counter.get();
+        self.tx_counter.set(app.tx_counter.wrapping_add(1));
```
hudson-ayers, line 130: When this wraps, an application with u32::MAX as its sequence number could get starved indefinitely. Obviously that is a lot of prints so maybe not worth worrying about?

capsules/src/driver.rs
phil-levis: Great question -- I think this is worth a discussion in the core WG. The current console driver allows userspace to submit arbitrary length strings and always prints them completely. This driver might truncate them (because it's ultimately relying on copying into a finite in-kernel buffer). 

So I think there are two possible ways forward:
  - Have them as alternative implementations of the same device, with different runtime behavior. In this case, we will want to look at the API to make sure that the API is such that both behaviors are acceptable, and userspace receives enough information (in return values or callbacks) to respond appropriately.
  - Have them as separate devices, with potentially different APIs and behaviors, and have them accessible through different userspace library calls (e.g., printf for console, klog for this one). 

I personally kind of lean towards the second option, but I'm interested to hear other people's thoughts.

boards/components/src/debug_writer.rs
phil-levis: I think it's a good idea to have such an increase with this particular driver. It allows userspace to be able to dump a lot of information. 1K is often too small, for example, for what one of our kernel dumps looks like. I suppose the issue is that this is shared across platforms, and a RAM-starved one might not want this?

Also worth discussing. I rarely find RAM tight on systems, but I do often find debug overflows when debugging. :)

boards/components/src/print_log.rs
phil-levis: :)

boards/imix/src/main.rs
phil-levis: They have different semantics. Console always fully prints the userspace message (barring failures), while printlog will truncate if it's too long.

printlog also doesn't handle reading from the console.

I'm not arguing that imix should definitely always have both going forward, but they are different.

capsules/src/print_log.rs
```diff
@@ -0,0 +1,276 @@
+//! Provides userspace with access to a serial interface whose output
```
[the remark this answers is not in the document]
phil-levis: The key property is that the write is synchronous: this is what allows it to be in-order. But it bounds the length of what you can write. I avoided calling it a Console because you can't read from it. But I agree that a different name is needed. I thought log was a good name because of the lack of reading?

capsules/src/print_log.rs
phil-levis: Ah, good point; once you wrap around, all of the outstanding writes from before the wraparound could be reordered, or starved indefinitely (if the console is always fully busy). I hadn't considered this case. As soon as it falls, idle, starvation isn't a problem, but re-ordering will be in all cases. So I think this is a major concern. I'll work on a fix.

capsules/src/print_log.rs
phil-levis: This was copied from console; I was trying to keep the same behavior. I can change both in a separate PR?

kernel/src/debug.rs
```diff
@@ -428,11 +429,11 @@ impl DebugWriter {
     }
 
     /// Write as many of the bytes from the internal_buffer to the output
-    /// mechanism as possible.
-    fn publish_bytes(&self) {
+    /// mechanism as possible, returning the number written.
+    fn publish_bytes(&self) -> usize {
```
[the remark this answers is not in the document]
phil-levis: I was thinking that if we do want Console/PrintLog to be the same device, then PrintLog will need to return how many bytes were written, so userspace can resend if it's less than the full amount.

kernel/src/debug.rs
phil-levis: Maybe? This seems a bit backwards. It would mean that regular `debug!` calls of kernel information are cast into a `ReadableProcessSlice`? I agree that it makes this particular bit of code more idiomatic, but it would be a strange hoop to jump through and be confusing to most readers of `Debug`.

capsules/src/print_log.rs
phil-levis: I don't think that's right; I handled wraparound. Sorry, this code was a month old so I forgot.

https://github.com/tock/tock/blob/5d939e1363d41e7fda2167581b7cc69f706a6b92/capsules/src/print_log.rs#L207

For example, suppose for simplicity that it's using an 8-bit rather than 32-bit counter. There are four writes enqueued: 0xfd , 0xfe, 0xff and 0x00. 0xfd is currently in flight.

When the alarm fires, 

https://github.com/tock/tock/blob/5d939e1363d41e7fda2167581b7cc69f706a6b92/capsules/src/print_log.rs#L179

after checking if it should issue a callback, the handler goes to figure out which write to write next. `self.tx_counter = 1` since there is write with 0x00 enqueued.

Suppose that the order of the writes in `apps.iter()` is 0, 0xfe, 0xff, 0xfd. The loop

https://github.com/tock/tock/blob/5d939e1363d41e7fda2167581b7cc69f706a6b92/capsules/src/print_log.rs#L201

Here are the iterations:
  1. check if `1.wrapping_sub(0) < (0xff / 2)` so `1 < 0x7f`. It is,  so `seqno = 0`.
  2. check if `0.wrapping_sub(0xfe) < (0xff / 2)`, so `2 < 0x7f`. It is, so `seqno = 0xfe`.
  3. check if `0xfe.wrapping_sub(0xff) < (0xff / 2)`, so `0xff < 0x7f`. It isn't, so `seqno = 0xfe`.
  4. check if `0xfe.wrapping_sub(0xfd) < (0xff / 2), so `0x01 < 0x7f`. It is, so `seqno = 0xfd`.

Since the gap between the lowest and the highest sequence numbers is at most the number of processes, this works as long as the number of processes is less than `usize::MAX / 2`. The sequence number increments only if there's a valid grant, and the send can only fail if there isn't a valid grant. So, there isn't a danger that an application calls this in a loop, there's a problem with its grant and it causes the sequence numbers to advance and spread more than the number of processes.

capsules/src/driver.rs
hudson-ayers: +1 for discussing this on the core call

boards/components/src/debug_writer.rs
hudson-ayers: I think this change would break the HiFive board today. We should probably just adjust the component to add an optional paramater to the macro, so the HiFive can maintain a smaller debug buffer than other boards.

boards/imix/src/main.rs
hudson-ayers: Marking as resolved, this seems like the same discussion as in my other comment

capsules/src/print_log.rs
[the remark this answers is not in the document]
hudson-ayers: I personally don't associate all consoles as being read+write; for example the console object in javascript is write only AFAIK (https://developer.mozilla.org/en-US/docs/Web/API/console). That said, I think `InOrderLog` is fine; I just don't like the collision with the log capsule meant for logging to flash.

capsules/src/print_log.rs
hudson-ayers: Ok, I agree that wraparound is handled -- sorry for making you revisit this, I should have checked the actual code. I was thrown off by this line in your PR description: "While a client (process) finishes sending, the driver looks for the client with a write pending that has the lowest sequence number.", and did not verify that the actual implementation wasn't in fact smarter.

kernel/src/debug.rs
hudson-ayers: Yeah, I don't feel too strongly about it. Worth nothing that even in the final binary, this does not get optimized to something more efficient -- there is actually a unique call to `write()` for every byte in the slice:

```
   1b336:       f011 fdcb       bl      2ced0 <_ZN6kernel5debug16get_debug_writer17h9eb5ff7542767abcE>
   1b33a:       f1a7 0919       sub.w   r9, r7, #25
   1b33e:       4605            mov     r5, r0
   1b340:       b156            cbz     r6, 1b358 <_ZN8capsules9print_log17PrintLog$LT$A$GT$4send17h5a74adc2852bad0fE+0x92>
   1b342:       f814 0b01       ldrb.w  r0, [r4], #1
   1b346:       4649            mov     r1, r9
   1b348:       f807 0c19       strb.w  r0, [r7, #-25]
   1b34c:       4628            mov     r0, r5
   1b34e:       2201            movs    r2, #1
   1b350:       f011 fb3e       bl      2c9d0 <_ZN76_$LT$kernel..debug..DebugWriterWrapper$u20$as$u20$kernel..debug..IoWrite$GT$5write17hbb6dbc4f34f5e269E>
   1b354:       3e01            subs    r6, #1
   1b356:       e7f3            b.n     1b340 <_ZN8capsules9print_log17PrintLog$LT$A$GT$4send17h5a74adc2852bad0fE+0x7a>
   1b358:       4628            mov     r0, r5
   1b35a:       f011 fddf       bl      2cf1c <_ZN6kernel5debug18DebugWriterWrapper13publish_bytes17hddce97158654c963E>
```

capsules/src/print_log.rs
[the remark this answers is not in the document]
phil-levis: We could change Console to ReadWriteConsole and this to InOrderConsoleLog.... :)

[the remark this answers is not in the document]
hudson-ayers: I suppose any naming should wait for the conclusions of discussing this on the core call -- if we wanted these to implement the same device we would need to port over the read functionality from Console anyway

capsules/src/driver.rs
phil-levis: The resolution on the core call was to go with option #1 (two alternative implementations of the same device/system call API).

hudson-ayers: Does that mean we will also port over the read portion of the console driver to this capsule?

phil-levis: I think so, yes.

phil-levis:
This was discussed in the core call on December 16th:

https://github.com/tock/tock/blob/912f18de861164afd51817b2d57164f1d7b32346/doc/wg/core/notes/core-notes-2022-12-16.md

The conclusion is that synchronous writes to the console (the code in this PR) should be considered an alternative implementation of the console system call API. Tock will therefore have two different implementations of the console system calls. One (the existing one today) performs asynchronous writes of arbitrary length, while another (the one in this PR) performs synchronous writes of finite length. This will require some changes to the API, as now the number of bytes written must be passed back to userspace so it knows whether the whole buffer was written. I am going to start working on this.

phil-levis:
It looks like there is a complication: the newlib implementation of printf assumes that `putnstr` writes the entire buffer; `printf` returns the size of the complete buffer regardless of what `putnstr` returns. I see two options going forward:

1. Calls to `printf` always return the buffer size passed, even if it is not fully output.
2. Calls to `printf` always output the entire buffer, but the write is no longer atomic: other debug statements may be interleaved inside the `printf`.

There is also kind of a third option, which is to re-implement newlib's printf.

I'll put this on the agenda for the call tomorrow.

boards/components/src/debug_writer.rs
hudson-ayers: Submitted https://github.com/tock/tock/pull/3369 to address this

phil-levis:
@kupiakos @bradjc 

I am finally getting back to this. The call on [12/16/22](https://github.com/tock/tock/blob/master/doc/wg/core/notes/core-notes-2022-12-16.md) concluded that this new console and the old one should have the same system call device number, and be alternative implementations. This leads to the complication that the two behave differently. One (asynchronously) prints the entire offered string, while the other prints potentially a truncated subset. This means that we have two options: 
  1) have a write state machine in userspace that can means large prints might be broken into multiple, smaller ones which the kernel can handle. This means a print from userspace will not be atomic and can have other writes interleaved within it.
  2) truncate long prints from userspace: they atomically print as much as they can in the kernel buffer and return.

Looking through the call notes, I don't see a clear resolution. I suggested that we should do #2, but I don't see substantial discussion or agreement, so want to make sure that's the right conclusion before wrapping this up.

hudson-ayers:
+1 for approach #2

phil-levis:
@kupiakos @vsukhoml  -- option 1 or option 2?

vsukhoml:
Option 3: have a state machine in kernel console implementation which will print full lines from applications, possibly use line buffer in apps grant. Benefit vs option 1 - state machine is implemented once and shared across apps. Benefit vs. option 2 - print full lines.

There are certain corner cases I can think of:
1. Adding timestamps - probably time stamps should be added at the time of first symbol of string received by driver no matter when last symbol / line feed is received.
2. Probably need a `flush` call to force print to support prints with control codes which overwrite previously printed characters. E.g. some dynamic/animated prompt with `/ - \ |`  transition.

kupiakos:
Option 1 resembles what our Ti50 console does right now - we line buffer in userspace and issue `blocking_command` syscalls for every line (grouping multiple lines into a single syscall when available). We've managed to strip down most of the size cost from doing this, but it's still there and duplicated per-app. An issue I have with approach 2 is how are userspace applications supposed to know the max amount they can send to the console at a time without data being dropped? For cases where we want to print a lot of data and guarantee it's all sent, we'd need to have some sort of breakup system in userspace anyways.

I'm curious what y'all think of Option 3 offered by Vadim.

phil-levis:
@vsukhoml @kupiakos 

Just getting back to this. I think I would need to understand the proposed semantics better. @vsukhoml You have described an implementation, not an interface. What is the desired behavior in response to a large printf that is larger than the available space in the kernel buffer? The current options are:
  1. Prints the entire printf but other prints (concurrent, ordered) may be interspersed with it. I.e., if a print P2 appears between the start of print P1 and the end of print P1, this means it was invoked after P1 was invoked and before P1 returned.
  2. Prints as much of the printf as it can atomically, truncating it if there is not enough space.

From your description, I can't tell if you want:
  a. Prints the entire print, other prints may be interspersed with it, but interspersals will always occur at line boundaries. Note that kernel prints may be truncated, as they are currently.
  b. Prints the entire printf without an interspersed prints.

b) is not possible without console locking or blocking within the kernel. Console locking would mean that other prints would be lost while it is locked (there is no way to buffer them).

phil-levis:
@vsukhoml and I discussed the desired semantics. They are temporal ordering and atomicity of a line (up to a newline) up to a finite length L. If userspace tries to print a line longer than L, the kernel will print up to L and return to userspace how much was printed.  If userspace prints a buffer that is < L bytes and has more than one newline in it, it's up to the kernel on how many lines it prints (it could print 1, 2, or N).

phil-levis:
One update. libtock-c's `newlib` semantics for `printf` are:
  - An output string that contains multiple newlines is printed one line at a time (each line is a separate invocation of `putnstr`)
  - The maximum length of a single print is 1024 characters; if more than 1023 characters are before the newline it prints them in 1024 chunks and the last chunk ends in a newline

I talked with @vsukhoml and we concluded this means the kernel console driver shouldn't worry about newlines, merely ordered and atomic writes. As console writes no longer guarantee writing every byte, libtock-c will require some updates to `sys.c`.

The code now assures atomicity of writes up to 200 bytes (by waiting for 200 bytes to be available in the debug buffer) and ensures that userspace writes do not interleave. This means that kernel debug writes can interleave at 200 byte granularity if the debug buffer is full. I've tested the current code with two userspace writers and interleaved kernel writes (logging the system calls of the writes themselves). I'll test it some more in the next week.

phil-levis:
On another note, I think I've also found some bugs in the newlib printf implementation. Writes that are between 1400-1800 bytes or so (without a newline) behave erratically, sometimes truncating the first 1024 bytes. It's deterministic based on the length, but different lengths behave differently. E.g., writes longer than 1480 bytes do not print the first 1024 bytes, while writes 1600-2000 or so do not print anything at all. I will look at the source to try to get to the bottom of this.

phil-levis:
> On another note, I think I've also found some bugs in the newlib printf implementation. Writes that are between 1400-1800 bytes or so (without a newline) behave erratically, sometimes truncating the first 1024 bytes. It's deterministic based on the length, but different lengths behave differently. E.g., writes longer than 1480 bytes do not print the first 1024 bytes, while writes 1600-2000 or so do not print anything at all. I will look at the source to try to get to the bottom of this.

We discussed this on the core call today, and @bradjc suggested it was likely a malloc failure issue within newlib. He's correct; if you increase the size of the process heap, then prints within this range complete correctly.

vsukhoml:
Might be a reason to switch from newlib to https://github.com/vsukhoml/noc :-)  - I tried to add all functionality required by libtock-c.

phil-levis:
bors try

hudson-ayers:
Is the status of this that it is ready for final review now?

phil-levis:
> Is the status of this that it is ready for final review now?

Not quite yet. Right now, this requires updates to libtock-c. I think I can change the logic to always complete writes before issuing a callback, which would not require an update to libtock-c.

phil-levis:
@hudson-ayers @kupiakos

This is now ready for review.

bradjc (approved):
Looks good!

PR #3333, opened 2022-11-25
chips/sifive/clint: make generic over timebase `Frequency`

lschuermann:
### Pull Request Overview

It turns out that at least the QEMU emulated boards (`hifive1` and `qemu_rv32_virt`) provide a Clint timebase of 10MHz, whereas the previous implementation (and presumably physical chips) assumed a timebase of 32KHz. This forwards the frequency as a generic parameter over the `Clint` struct, such that the driver implementation can be shared between chips but the frequency adjusted to meet the actual underlying hardware. It furthermore corrects the `qemu_rv32_virt` board to use the proper 10MHz frequency.

While we're at it and introducing a new 10KHz `Frequency` into the time HIL, this further converts the existing `Frequency` structs into variant-less enums, which is best practice for marker-types which are not supposed to be used at runtime, and formats the frequency number literals for readability. 

### Testing Strategy

This pull request was tested by compiling and running in QEMU.

### TODO or Help Wanted

This doesn't solve the problem around the QEMU `hifive1` board, where there is a mismatch between the expected timebase frequency still. Presumably the 32KHz are correct (this [thread](https://forums.sifive.com/t/clint-and-sleep/2355) seems to indicate that), so maybe this should be changed in QEMU?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/arty_e21/src/main.rs
+++ b/boards/arty_e21/src/main.rs
@@ -49,7 +49,7 @@ struct ArtyE21 {
     gpio: &'static capsules::gpio::GPIO<'static, arty_e21_chip::gpio::GpioPin<'static>>,
     alarm: &'static capsules::alarm::AlarmDriver<
         'static,
-        VirtualMuxAlarm<'static, sifive::clint::Clint<'static>>,
+        VirtualMuxAlarm<'static, arty_e21_chip::chip::ArtyExxClint<'static>>,
     >,
     led: &'static capsules::led::LedDriver<
         'static,
@@ -176,7 +176,7 @@ pub unsafe fn main() {
     // Create a shared virtualization mux layer on top of a single hardware
     // alarm.
     let mux_alarm = static_init!(
-        MuxAlarm<'static, sifive::clint::Clint>,
+        MuxAlarm<'static, arty_e21_chip::chip::ArtyExxClint>,
         MuxAlarm::new(&peripherals.machinetimer)
     );
     hil::time::Alarm::set_alarm_client(&peripherals.machinetimer, mux_alarm);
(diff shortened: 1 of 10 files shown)


hudson-ayers (approved):
These changes look good to me, especially the switch to variant-less enums.

I do wonder whether we should just replace the Frequency trait with the use of a `F` const generic parameter everywhere that currently accepts a generic Frequency type. We could still define constants like `Freq32KHz`, and we could replace everywhere that uses `Frequency` as an associated type with an associated constant. Seems like this is what we might have done originally if const generics support existed at the time.

lschuermann:
> I do wonder whether we should just replace the Frequency trait with the use of a `F` const generic parameter everywhere that currently accepts a generic Frequency type. We could still define constants like `Freq32KHz`, and we could replace everywhere that uses `Frequency` as an associated type with an associated constant. Seems like this is what we might have done originally if const generics support existed at the time.

Agreed, that sounds like an excellent idea. However, this would touch much of the existing infrastructure relying on the `time` HIL, so we'd better split that out into its own separate PR.

lschuermann:
Okay, I've tried to switch the `time` HIL over to use an associated constant for the frequency and hit a roadblock:

https://github.com/tock/tock/blob/fbf96bd2667f56ceb367506e44172bd3b676b5c3/chips/imxrt10xx/src/gpt.rs#L339-L364

This code makes me really uneasy. On the one hand I admire its cleverness, but on the other hand this is a perfect demonstration of why we would want to use associated constants over associated types for specifying the frequency in time-HIL interfaces: the entire interface and (presumably) all implementations of higher-level abstractions (such as virtual alarms) are not at all built to handle switching the underlying timebase frequency at runtime. While perhaps legitimate in this case, it nonetheless demonstrates that changing the returned value of `Frequency::frequency` at runtime is not only possible, but done in practice. This introduces concerns we'd normally not have: simply instantiating a struct/component (e.g., `AlarmDriver`) at the wrong time can cause it to read (and cache) an invalid frequency value, assuming it wouldn't change over time. This seems like a nightmare to debug.

While I am open to supporting ideas like frequency scaling as part of the time HIL, I think this would deserve its own, separate and explicit mechanism. For the current time HIL as specified in TRD 105, I think its safe to establish
- that the HIL is really only applicable to hardware or software time-sources which do not change their reference frequency,
- and the HIL is underspecified in that it does not explicitly require the `Frequency::frequency` method to return a constant value, likely because this possibility was not thought of in the design phase.

Circling back to the issue at hand: I feel like a switch to associated constants in the long term would be worth it, alongside a corresponding update to TRD 105 (still a draft). For the current board we may want to collaborate with the author of this respective code to find a workaround. What do you think, @hudson-ayers?

hudson-ayers:
I completely agree that the time HIL was not designed to support dynamic frequency changes at runtime, and that fixing it to return a constant value would be a strict improvement in that it would prevent people from attempting to use it in this way. 

Holly's power clocks work supported dynamic frequency changes at runtime, and required a much more complex API to work correctly alongside peripherals that require locked frequencies whie operating, or require scalar adjustments if frequencies change, or that only support certain clock sources: https://github.com/hchiang/tock/blob/e95d9a99ef5cda834f35e7932b704d4b7fafc7cb/kernel/src/hil/clock_pm.rs#L63-L83

hudson-ayers:
Marked last call!

lschuermann:
Hm, the current draft TRD does define the standard implementations of `Frequency` and that they are structs. I believe we can update this (draft) TRD, but perhaps that's worth considering?

hudson-ayers:
I am in favor of updating the TRD once we have a working implementation using const generics. Agreed that level of detail is probably not helpful.

We might want to move the discussion about switching `Frequency` to just be a const generic parameter over to an issue, unless you think you'll have a separate PR for that soon

alistair23 (approved):

PR #3335, opened 2022-11-28
chips/lowrisc: small refactor / reduce some bloat

twilfredo:
### Pull Request Overview

Trim a little bit of the bloat...
```
For hmac.rs:
    0.0%   0.2%     266B         lowrisc <lowrisc::hmac::Hmac as kernel::hil::digest::HmacSha256>::set_mode_hmacsha256
->  0.0%   0.2%     236B         lowrisc <lowrisc::hmac::Hmac as kernel::hil::digest::HmacSha256>::set_mode_hmacsha256

For uart.rs:
   0.0%   0.1%     102B         lowrisc lowrisc::uart::Uart::tx_progress
-> 0.0%   0.1%      84B         lowrisc lowrisc::uart::Uart::tx_progress

For i2c.rs
   0.0%   0.2%     276B         lowrisc lowrisc::i2c::I2c::write_data
-> 0.0%   0.2%     266B         lowrisc lowrisc::i2c::I2c::write_data

   0.0%   0.2%     244B         lowrisc lowrisc::i2c::I2c::write_read_data
-> 0.0%   0.2%     232B         lowrisc lowrisc::i2c::I2c::read_data

   0.0%   0.2%     232B         lowrisc lowrisc::i2c::I2c::read_data
-> 0.0%   0.2%     224B         lowrisc lowrisc::i2c::I2c::  write_read_data

For flash.rs
   0.0%   0.0%      60B         lowrisc lowrisc::flash_ctrl::FlashCtrl::configure_data_partition
-> 0.0%   0.0%      56B         lowrisc lowrisc::flash_ctrl::FlashCtrl::configure_data_partition

   0.0%   0.0%      46B         lowrisc lowrisc::flash_ctrl::FlashCtrl::configure_info_partition
-> 0.0%   0.0%      42B         lowrisc lowrisc::flash_ctrl::FlashCtrl::configure_info_partition

Total: ~86B
```

### Testing Strategy

Running the `hmac`, `flash` tests on Verilator, `i2c` was **NOT** tested.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/lowrisc/src/i2c.rs
+++ b/chips/lowrisc/src/i2c.rs
@@ -270,7 +270,8 @@ impl<'a> I2c<'_> {
                     break;
                 }
                 // Send the data
-                regs.fdata.write(FDATA::FBYTE.val(buf[i as usize] as u32));
+                regs.fdata
+                    .write(FDATA::FBYTE.val(*buf.get(i).unwrap_or(&0) as u32));
                 data_pushed = i;
             }
 
@@ -278,7 +279,7 @@ impl<'a> I2c<'_> {
             if regs.status.read(STATUS::FMTFULL) == 0 && data_pushed == (len - 1) {
                 // Send the last byte with the stop signal
                 regs.fdata
-                    .write(FDATA::FBYTE.val(buf[len as usize] as u32) + FDATA::STOP::SET);
+                    .write(FDATA::FBYTE.val(*buf.get(len).unwrap_or(&0) as u32) + FDATA::STOP::SET);
 
                 data_pushed = len;
             }
@@ -317,7 +318,8 @@ impl<'a> I2c<'_> {
                     break;
                 }
                 // Send the data
-                regs.fdata.write(FDATA::FBYTE.val(buf[i as usize] as u32));
+                regs.fdata
+                    .write(FDATA::FBYTE.val(*buf.get(i).unwrap_or(&0) as u32));
                 data_pushed = i;
             }
 
@@ -325,7 +327,7 @@ impl<'a> I2c<'_> {
             if regs.status.read(STATUS::FMTFULL) == 0 && data_pushed == (len - 1) {
                 // Send the last byte with the stop signal
                 regs.fdata
-                    .write(FDATA::FBYTE.val(buf[len as usize] as u32) + FDATA::STOP::SET);
+                    .write(FDATA::FBYTE.val(*buf.get(len).unwrap_or(&0) as u32) + FDATA::STOP::SET);
 
                 data_pushed = len;
             }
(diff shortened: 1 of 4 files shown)


chips/lowrisc/src/i2c.rs
```diff
@@ -317,15 +318,16 @@ impl<'a> I2c<'_> {
...
+                regs.fdata
+                    .write(FDATA::FBYTE.val(*buf.get(i).unwrap_or(&0) as u32));
                 data_pushed = i;
             }
 
             // Check if we can send the last byte
             if regs.status.read(STATUS::FMTFULL) == 0 && data_pushed == (len - 1) {
                 // Send the last byte with the stop signal
                 regs.fdata
-                    .write(FDATA::FBYTE.val(buf[len as usize] as u32) + FDATA::STOP::SET);
```
twilfredo, line 328: Not sure if it's better to panic in a case like this or to 'silently fail' and write 0?

chips/lowrisc/src/i2c.rs
```diff
@@ -270,15 +270,16 @@ impl<'a> I2c<'_> {
...
+                    .write(FDATA::FBYTE.val(*buf.get(len).unwrap_or(&0) as u32) + FDATA::STOP::SET);
```
twilfredo, line 282: Not sure if it's better to panic in a case like this or to 'silently fail' and write 0?

bradjc (dismissed):

chips/lowrisc/src/i2c.rs
hudson-ayers: If you just care about binary size, usually adding a
```rust
if len >= buf.len() { return; } 
```
at the top and then keeping the normal array access should be sufficient for LLVM to optimize out the later panicking bounds check. 

When I made a similar series of passes to a driver for code size, I preferred to add Result<(), ErrorCode> return values to these functions and then just panic in the event of any error at the highest level possible, which meant that attempts at OOB accesses would still generate panics, but would not force a size cost of 12 panics in the binary and instead just have the size cost of 1.

twilfredo: > If you just care about binary size, usually adding a
> 
> ```rust
> if len >= buf.len() { return; } 
> ```
> 
> at the top and then keeping the normal array access should be sufficient for LLVM to optimize out the later panicking bounds check.
> 
Ah cool!
> When I made a similar series of passes to a driver for code size, I preferred to add Result<(), ErrorCode> return values to these functions and then just panic in the event of any error at the highest level possible, which meant that attempts at OOB accesses would still generate panics, but would not force a size cost of 12 panics in the binary and instead just have the size cost of 1.

Sounds like a better idea, I'll try to implement these where possible.

alistair23 (approved):

hudson-ayers (dismissed):

twilfredo:
Rebased to resolve conflicts with this guy https://github.com/tock/tock/pull/3250

hudson-ayers (approved):

Issue #3311, opened 2022-11-02
tock-registers: CLEAR FieldValue does not perform match check with inverted mask


Fighter19:
Consider following test case:
```rust
        #[test]
        fn test_clear_match() {
            register_bitfields! {
                u32,
            
                TEST [
                    FLAG OFFSET(18) NUMBITS(1) [],
                ]
            }

            let value = 266241;
            match TEST::FLAG::CLEAR.matches_any(value) {
                true => panic!("Field Value should not have matched {}", TEST::FLAG::CLEAR.mask()),
                false => (),
            }
        }
```

Even though the test value has bit 18 set, the result of the function returns true.

brghena:
266241 -> 0x41001 which sets bits 0, 12, and 18. So I agree that this code _should_ definitely work.

Do you know if it's a problem with all functions, or just `matches_any()` maybe?

Fighter19:
I only noticed it with this particular function.

hudson-ayers:
I see two issues here:

**Poor Name**

`matches_any()` is a poor name for the functionality it implements, it would be better named `is_any_set()`, since it just checks if any bits in the passed parts of the field are set -- hence why the provided example using `TEST::FLAG::CLEAR` does not work. It would also work incorrectly for a multi-bit field:
```rust
//! dma.rs

register_bitfields![u32,
    Mode [
        /// Size of Transfer
        SIZE OFFSET(0) NUMBITS(2) [
            Byte = 0,
            Halfword = 1,
            Word = 2
        ]
    ],
];

// The below variable will be `true` if the dma register is set to either `Halfword` 
// or `Word`, which is clearly incorrect
let is_halfword_transfer = dma_regs.matches_any(Mode::SIZE::HALFWORD);
```

I don't think it is possible to provide a "correct" implementation for the current interface, since the current interface relies on multiple `FieldValue`s being added into a single `FieldValue`, but a correct implementation which could actually check on a field-by-field basis for any matches would need to retain the information about which individual fields were passed.

**I think we should rename the current method to `any_bits_set()`. If we want to be able to provide the functionality suggested by the current name, we would need a new interface:**

```rust
trait Readable {
    #[inline]
    /// Check if any specified parts of a field match
    fn matches_any(&self, fields: &[FieldValue<Self::T, Self::R>]) -> bool {
        for field in fields {
            if self.get() & field.mask() == field.value {
                return true;
            }
        }
        return false;
    }
}
```

[...]

PR #3336, opened 2022-11-28
Rename `matches_any()` to `any_matching_bits_set()`, implement new `matches_any()`

hudson-ayers:
### Pull Request Overview

The current implementation of `matches_any()` does not implement the functionality the name implies. This PR renames the existing implementation to a name which better describes its functionality, and introduces a new `matches_any()` function (with a different interface) that actually correctly implements the functionality suggested by the name. 

This PR also adds several tests to the tock-registers test suite to verify the new version works as expected, and removes a feature gate on a feature that no longer exists for the crate which was preventing some of the tock-registers test suite from being run as part of `cargo test`.

https://github.com/tock/tock/issues/3311#issuecomment-1328113050 contains a complete description of the issues with the current method.

Fixes https://github.com/tock/tock/issues/3311

### Testing Strategy

### TODO or Help Wanted

This pull request still needs feedback on method names

### Documentation Updated

- [x] `README.md` updated.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/libraries/tock-register-interface/src/interfaces.rs
+++ b/libraries/tock-register-interface/src/interfaces.rs
@@ -230,16 +230,29 @@ pub trait Readable {
         field.is_set(self.get())
     }
 
+    /// Check if any bits corresponding to the mask in the passed `FieldValue` are set.
+    /// This function is identical to `is_set()` but operates on a `FieldValue` rather
+    /// than a `Field`, allowing for checking if any bits are set across multiple,
+    /// non-contiguous portions of a bitfield.
     #[inline]
-    /// Check if any specified parts of a field match
-    fn matches_any(&self, field: FieldValue<Self::T, Self::R>) -> bool {
-        field.matches_any(self.get())
+    fn any_matching_bits_set(&self, field: FieldValue<Self::T, Self::R>) -> bool {
+        field.any_matching_bits_set(self.get())
     }
 
     #[inline]
     /// Check if all specified parts of a field match
     fn matches_all(&self, field: FieldValue<Self::T, Self::R>) -> bool {
         field.matches_all(self.get())
+    }
+
+    /// Check if any of the passed parts of a field exactly match the contained
+    /// value. This allows for matching on unset bits, or matching on specific values
+    /// in multi-bit fields.
+    #[inline]
+    fn matches_any(&self, fields: &[FieldValue<Self::T, Self::R>]) -> bool {
+        fields
+            .iter()
+            .any(|field| self.get() & field.mask() == field.value)
     }
 }
 
--- a/chips/apollo3/src/stimer.rs
+++ b/chips/apollo3/src/stimer.rs
@@ -152,7 +152,7 @@ impl<'a> Counter<'a> for STimer<'a> {
 
     fn is_running(&self) -> bool {
         let regs = self.registers;
-        regs.stcfg.matches_any(STCFG::CLKSEL::XTAL_DIV2)
+        regs.stcfg.matches_any(&[STCFG::CLKSEL::XTAL_DIV2])
     }
 }
 
--- a/chips/e310x/src/chip.rs
+++ b/chips/e310x/src/chip.rs
@@ -147,7 +147,9 @@ impl<'a, I: InterruptService<DeferredCallTask> + 'a> kernel::platform::chip::Chi
                 }
             }
 
-            if !mip.matches_any(mip::mtimer::SET) && self.plic.get_saved_interrupts().is_none() {
+            if !mip.any_matching_bits_set(mip::mtimer::SET)
+                && self.plic.get_saved_interrupts().is_none()
+            {
                 break;
             }
         }
--- a/chips/litex/src/litex_registers.rs
+++ b/chips/litex/src/litex_registers.rs
@@ -186,7 +186,7 @@ pub trait Read<T: UIntLike> {
     fn is_set(&self, field: Field<T, Self::Reg>) -> bool;
 
     /// Check if any specified parts of a field match
-    fn matches_any(&self, field: FieldValue<T, Self::Reg>) -> bool;
+    fn any_matching_bits_set(&self, field: FieldValue<T, Self::Reg>) -> bool;
 
     /// Check if all specified parts of a field match
     fn matches_all(&self, field: FieldValue<T, Self::Reg>) -> bool;
@@ -273,8 +273,8 @@ where
     }
 
     #[inline]
-    fn matches_any(&self, field: FieldValue<T, Self::Reg>) -> bool {
-        field.matches_any(self.get())
+    fn any_matching_bits_set(&self, field: FieldValue<T, Self::Reg>) -> bool {
+        field.any_matching_bits_set(self.get())
     }
 
     #[inline]
--- a/chips/nrf52/src/acomp.rs
+++ b/chips/nrf52/src/acomp.rs
@@ -231,7 +231,11 @@ impl<'a> Comparator<'a> {
     fn enable(&self) {
         // Checks if it's already enabled
         // Assumes no one else is writing to comp registers directly
-        if self.registers.enable.matches_any(Enable::ENABLE::Enabled) {
+        if self
+            .registers
+            .enable
+            .any_matching_bits_set(Enable::ENABLE::Enabled)
+        {
             return;
         }
 
--- a/chips/nrf5x/src/gpio.rs
+++ b/chips/nrf5x/src/gpio.rs
@@ -483,7 +483,7 @@ impl<'a> hil::gpio::Interrupt<'a> for GPIOPin<'a> {
     fn is_pending(&self) -> bool {
         if let Ok(channel) = self.find_channel(self.pin) {
             let ev = &self.gpiote_registers.event_in[channel];
-            ev.matches_any(EventsIn::EVENT::Ready)
+            ev.any_matching_bits_set(EventsIn::EVENT::Ready)
         } else {
             false
         }
@@ -576,7 +576,7 @@ impl<'a, const N: usize> Port<'a, N> {
         let pin_registers = self.pins[0].gpiote_registers;
 
         for (i, ev) in pin_registers.event_in.iter().enumerate() {
-            if ev.matches_any(EventsIn::EVENT::Ready) {
+            if ev.any_matching_bits_set(EventsIn::EVENT::Ready) {
                 ev.write(EventsIn::EVENT::NotReady);
                 // Get pin number for the event and `trigger` an interrupt manually on that pin
                 let pin = pin_registers.config[i].read(Config::PSEL) as usize;
--- a/chips/qemu_rv32_virt_chip/src/chip.rs
+++ b/chips/qemu_rv32_virt_chip/src/chip.rs
@@ -131,7 +131,9 @@ impl<'a, I: InterruptService<()> + 'a> Chip for QemuRv32VirtChip<'a, I> {
                 }
             }
 
-            if !mip.matches_any(mip::mtimer::SET) && self.plic.get_saved_interrupts().is_none() {
+            if !mip.any_matching_bits_set(mip::mtimer::SET)
+                && self.plic.get_saved_interrupts().is_none()
+            {
                 break;
             }
         }
--- a/chips/sam4l/src/i2c.rs
+++ b/chips/sam4l/src/i2c.rs
@@ -1027,7 +1027,7 @@ impl I2CHw {
             let interrupts = status.bitand(imr.get());
 
             // Check for errors.
-            if interrupts.matches_any(
+            if interrupts.any_matching_bits_set(
                 StatusSlave::BUSERR::SET
                     + StatusSlave::SMBPECERR::SET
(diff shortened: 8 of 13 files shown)


bradjc:
Need to update readme.

lschuermann (dismissed):
Looks good except for the pending README update. Thanks @hudson-ayers!

libraries/tock-register-interface/src/interfaces.rs
```diff
@@ -230,17 +230,30 @@ pub trait Readable {
...
-    fn matches_any(&self, field: FieldValue<Self::T, Self::R>) -> bool {
-        field.matches_any(self.get())
+    fn any_matching_bits_set(&self, field: FieldValue<Self::T, Self::R>) -> bool {
+        field.any_matching_bits_set(self.get())
     }
 
     #[inline]
     /// Check if all specified parts of a field match
     fn matches_all(&self, field: FieldValue<Self::T, Self::R>) -> bool {
         field.matches_all(self.get())
```
bradjc, line 245: It seems like `matches_all()` needs to change as well. This is now `all_matching_bits_set()`, right?

hudson-ayers:
README updated

bradjc (commented):
Fixing this is a little confusing.

If we look at what is probably a very old use of this API (in SAM4L USART), it makes it more clear what `matches_any()` was trying to do do:

```rust
        let ints_active = self.registers.imr.matches_any(
            Interrupt::RXBUFF::SET
                + Interrupt::TXEMPTY::SET
                + Interrupt::TIMEOUT::SET
                + Interrupt::PARE::SET
                + Interrupt::FRAME::SET
                + Interrupt::OVRE::SET
                + Interrupt::TXRDY::SET
                + Interrupt::RXRDY::SET,
        );

...

        if !(rx_active || tx_active || ints_active || is_panic) {
            pm::disable_clock(self.clock);
        }
```

Beside my point about `matches_all()` no longer being consistent, do we actually want two functions? It seems like what we want from an API point of view is the new `matches_any()`, and to update all current uses to the new version.

For example, this code in apollo3 stimer looks suspicious:

```rust
    fn is_running(&self) -> bool {
        let regs = self.registers;
        regs.stcfg.any_matching_bits_set(STCFG::CLKSEL::XTAL_DIV2)
    }
```

Particularly when given what start() does:

```rust
    fn start(&self) -> Result<(), ErrorCode> {
        // Set the clock source
        self.registers.stcfg.write(STCFG::CLKSEL::XTAL_DIV2);
        Ok(())
    }
```

Shouldn't the is_running() check be `regs.stcfg.matches_all([STCFG::CLKSEL::XTAL_DIV2])`?

I can see an efficiency argument for including `any_matching_bits_set()`, but I'm worried that if we can't restrict it to only single bit fields then it can be used in ways it shouldn't be and lead to very strange behavior.

libraries/tock-register-interface/src/interfaces.rs
hudson-ayers: `matches_all()` is actually already correct -- since it was already checking for an exact match across the entire passed `FieldValue`, it works correctly for multi-bit fields and for matching on cleared bits as well as set bits. 

For convenience here are the two implementations:
```rust
    /// Check if any of the bits covered by the mask for this
    /// `FieldValue` are set in the passed value
    #[inline]
    pub fn any_matching_bits_set(&self, val: T) -> bool {
        val & self.mask != T::zero()
    }

    /// Check if all specified parts of a field match
    #[inline]
    pub fn matches_all(&self, val: T) -> bool {
        val & self.mask == self.value
    }
```

`matches_all()` actually looks at `self.value()` and makes sure that the passed value, when masked, matches it. So the problem in #3311 does not apply.

bradjc: Ok good point. Seems like if there is only one FieldValue, then it should be `.matches()`.

hudson-ayers:
> Beside my point about `matches_all()` no longer being consistent, do we actually want two functions? It seems like what we want from an API point of view is the new `matches_any()`, and to update all current uses to the new version.
> 
> For example, this code in apollo3 stimer looks suspicious:
> 
> ```rust
>     fn is_running(&self) -> bool {
>         let regs = self.registers;
>         regs.stcfg.any_matching_bits_set(STCFG::CLKSEL::XTAL_DIV2)
>     }
> ```
> 

 Yeah, good call, this is incorrect, because CLKSEL is a multi-bit field.

> 
> Shouldn't the is_running() check be `regs.stcfg.matches_all([STCFG::CLKSEL::XTAL_DIV2])`?

For the interfaces currently defined in this PR, the correct change would be either `regs.stcfg.matches_all(STCFG::CLKSEL::XTAL_DIV2)` or `regs.stcfg.matches_any([STCFG::CLKSEL::XTAL_DIV2])`

> I can see an efficiency argument for including `any_matching_bits_set()`, but I'm worried that if we can't restrict it to only single bit fields then it can be used in ways it shouldn't be and lead to very strange behavior.

Yeah, the reason I included it is for efficiency -- when just matching on a bunch of single bit fields, it is much more efficient than the new `matches_any()`, which has to loop through each passed field. I don't think there is an easy way to enforce that it can only be called with a `FieldValue` composed of multiple single-bit `FieldValue`s added together though.

hudson-ayers:
Pushed an update to fix that stimer example (which was broken before this PR, to be clear)

hudson-ayers:
I would be onboard with removing `any_matching_bits_set()` and just forcing users to use the less-efficient, more correct `matches_any()` if people prefer that. If someone really needs the efficient implementation it is easy enough to handwrite 
```rust
    // current approach
    // let any_bits_set_efficient = self.regs.srp.any_matching_bits_set(TX::INT::SET + RX::INT::SET);
  
    // new approach if we remove `any_matching_bits_set()`
    let any_bits_set_efficient = self.regs.srp.extract() & (TX::INT + RX::INT) != 0;

```

hudson-ayers:
Any thoughts on these two approaches? I personally am in favor of what is currently in the PR, but don't feel too strongly

bradjc:
What I still feel like is swapping one quirk for another is that `reg.any_matching_bit_set(INTERRUPT::BIT1::CLR + INTERRUPT::BIT2::CLR)` works the same as `reg.any_matching_bit_set(INTERRUPT::BIT1::SET + INTERRUPT::BIT2::SET)`. This API really shouldn't use `FieldValue`s it should use `Field`s, but we can't do concatenation on `Field`s (`INTERRUPT::BIT1 + INTERRUPT::BIT2` doesn't compile) so if we want this efficient functionality we need to use FieldValues.

Can we `&` in the FieldValue value? That way it 1) forces the more intuitive use case (`::SET` rather than `::CLR`) and 2) makes the adjective "set" apply to both the FieldValue and the underlying register value, rather than just one (leaving the caller to understand which applies).

```rust
#[inline]
    pub fn any_matching_bits_set(&self, val: T) -> bool {
        val & self.mask & self.value != T::zero()
    }
```

We could also call it `any_set()`

```rust
           if interrupts.any_set(
                Status::BUSERR::SET
                    + Status::SMBPECERR::SET
                    + Status::SMBTOUT::SET
                    + Status::ORUN::SET
                    + Status::URUN::SET,
            ) {...}
```

hudson-ayers:
> What I still feel like is swapping one quirk for another is that `reg.any_matching_bit_set(INTERRUPT::BIT1::CLR + INTERRUPT::BIT2::CLR)` works the same as `reg.any_matching_bit_set(INTERRUPT::BIT1::SET + INTERRUPT::BIT2::SET)`. 
> 
> Can we `&` in the FieldValue value? That way it 1) forces the more intuitive use case (`::SET` rather than `::CLR`) and 2) makes the adjective "set" apply to both the FieldValue and the underlying register value, rather than just one (leaving the caller to understand which applies).
> 
> ```rust
> #[inline]
>     pub fn any_matching_bits_set(&self, val: T) -> bool {
>         val & self.mask & self.value != T::zero()
>     }
> ```
> 
> We could also call it `any_set()`
> 
> ```rust
>            if interrupts.any_set(
>                 Status::BUSERR::SET
>                     + Status::SMBPECERR::SET
>                     + Status::SMBTOUT::SET
>                     + Status::ORUN::SET
>                     + Status::URUN::SET,
>             ) {...}
> ```

Yeah, I think that seems like a reasonable approach. We will have to re-write the unit tests for the function, as this change breaks the current unit tests, but I find the current unit tests somewhat incomprehensible anyway

lschuermann:
> This API really shouldn't use `FieldValue`s it should use `Field`s, but we can't do concatenation on `Field`s (`INTERRUPT::BIT1 + INTERRUPT::BIT2` doesn't compile) so if we want this efficient functionality we need to use FieldValues.

@bradjc I mean, presumably we could override the bitwise-OR operator for `Field`s such that we could do something like the following?

```
interrupts.any_set(
    Status::BUSERR
    | Status::SMBPECERR
    | Status::SMBTOUT
    | Status::ORUN
    | Status::URUN
)
```

Although this may also make it more confusing, as we don't know specifically which values we're using (presumably this would be `T::max() & self.mask`).

> We could also call it `any_set()`

Not a fan of that as long as we're accepting any `T`, given that nothing's stopping a user from composing arbitrary things as the function argument, or just using raw integer types. Then "any" looses all meaning.

libraries/tock-register-interface/src/interfaces.rs
hudson-ayers: Yeah, that makes sense to me, but I don't think the current name is really *wrong* ... not sure how to balance that with a desire not to modify the API of tock-registers more than needed, especially since it is the one library with non-Tock downstream users

hudson-ayers:
> > This API really shouldn't use `FieldValue`s it should use `Field`s, but we can't do concatenation on `Field`s (`INTERRUPT::BIT1 + INTERRUPT::BIT2` doesn't compile) so if we want this efficient functionality we need to use FieldValues.
> 
> @bradjc I mean, presumably we could override the bitwise-OR operator for `Field`s such that we could do something like the following?
> 
> ```
> interrupts.any_set(
>     Status::BUSERR
>     | Status::SMBPECERR
>     | Status::SMBTOUT
>     | Status::ORUN
>     | Status::URUN
> )
> ```
> 
> Although this may also make it more confusing, as we don't know specifically which values we're using (presumably this would be `T::max() & self.mask`).

Yeah, I think that overriding the bitwise OR operator is probably going to make things more confusing, I would vote against that.

> 
> > We could also call it `any_set()`
> 
> Not a fan of that as long as we're accepting any `T`, given that nothing's stopping a user from composing arbitrary things as the function argument, or just using raw integer types. Then "any" looses all meaning.

In practice, I think it was a mistake to make most of the functions in `fields.rs` `pub`, instead of `pub(crate)`. It really only makes sense for these functions to be called by the outer functions on the `Readable/Writable` traits. Those functions do not accept arbitrary `T` anyway, which would resolve that concern. Would you like `any_set()` if the implementation in `fields.rs` was `pub(crate)`?

lschuermann:
> In practice, I think it was a mistake to make most of the functions in `fields.rs` `pub`, instead of `pub(crate)`. It really only makes sense for these functions to be called by the outer functions on the `Readable/Writable` traits. Those functions do not accept arbitrary `T` anyway, which would resolve that concern.

Yes, I agree with that. This function seem destined for misunderstanding their implementation and do not constrain the input types well to limit it to a reasonable subset of possible behaviors.

> Would you like `any_set()` if the implementation in `fields.rs` was `pub(crate)`?

I like `&`ing `self.value()` in its implementation as proposed by @bradjc, but I'm not sold on the `any_set` name. I think something more explicit like `any_matching_bits_set` is much more expressive.

As a side note, something that made me skittish when seeing the example above was the use of the `+` operator. Even though I'd consider myself very familiar with the library, my intuition was that presumably something like `Status::BUSERR::SET` must return a `T` (instead of our `FieldValue`), and then `+` would only be equal to a bitwise-or if these were distinct, non-overlapping fields. Considering the usual add-operator semantics, it seems that overloading `Add` instead of `BitOr` on `FieldValue` was a mistake.

hudson-ayers:
I rebased this PR, and made the following updates:
2. `FieldValue::any_matching_bits_set()` now `&`'s in `self.value`, forcing the more intuitive use case (i.e. `my_interrupts.any_matching_bits_set(MY_FIELD::CLEAR)` will never return true).
3. Updated the tests to reflect this functionality change

I did not make the following changes:
1. I did not change the name to `any_set()` based on Leon's pushback.
2. I did not make `FieldValue::any_matching_bits_set()` `pub(crate)`, since it turns out that `litex_registers.rs` has to be able to call it. This bums me out a bit, but I didn't see an easy fix for this that did not involve moving much of the logic from `litex_registers` into `tock_registers`. Still, it doesn't really feel right that we are basing the interface to `tock_registers` off the peculiarities of `litex_registers`, especially in this case where the function being public makes it much more likely to be used incorrectly in place of the correct method on the `Field` struct

hudson-ayers:
ping? @lschuermann @bradjc

alistair23 (approved):

bradjc (approved):
I've stood in the way long enough!

PR #1714, opened 2020-03-25
Use a Cargo workspace to make builds reproducible and speed up CI.

gendx:
### Pull Request Overview

This pull request adds a [Cargo workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) at the top-level directory to make build reproducible (see the analysis in https://github.com/tock/tock/issues/1666#issuecomment-603863887).

With a workspace, Cargo now only writes to a single `target/` directory in the root folder, instead of having a separate `target/` in each board. This means that some changes have to be made in the Makefiles and tools.

Another benefit of a single workspace is that shared dependencies can be built once (per CPU architecture) and cached across boards. For example, the kernel or the capsules only need to be compiled once per CPU architecture, instead of once per board, which can significantly speed up commands like `make allboards`.

### Testing Strategy

This pull request was tested by:
- Duplicating the Tock directory (two identical copies in `/path/to/tock` and `/path/to/tock2`), and running `make allboards`, checking that the SHA-256 checksum match.
- Running `make flash` in the nRF52840-DK board's folder, and checking that flashing the kernel worked properly.
- Travis-CI.

### TODO or Help Wanted

This pull request still needs:
- [x] Fixing netlify. Having a single `target/` folder allows to greatly simplify the `tools/build-all-docs.sh` script. I checked the output on a few files and it looks consistent, but some pages might break. On the other hand, some crates such as the [arty_e21](https://docs.tockos.org/arty_e21/index.html) board are currently not indexed on the navigation bar, and the `arty_e21` chip was not even documented so far on https://docs.tockos.org/.
- [x] A bit of cleanup throughout Makefiles. ~~For now I made just enough changes in `boards/Makefile.common` for `make ci-travis` to work, but other make rules should be checked.~~
- [ ] Checking that the `tools/` still work properly with a single target folder now at the top-level.
  - [x] `build-all-docs.sh` Simplified, as mentioned above, now that all docs are built in the same folder.
  - [x] `post_size_changes_to_github.sh` Updated, but it's currently broken for other reasons (see #1722).
- [x] Now fixed with a minimum search depth of 2. ~~I had to remove the `tools/check_wildcard_imports.sh` in `make ci-travis` because otherwise it reported the following error. The tool would need some refactoring.~~
  ```
  Wildcard import(s) found in ..
  Tock style rules prohibit this use of wildcard imports.
  
  The following wildcard imports were found:
  libraries/tock-register-interface/src/macros.rs:                use super::super::*;
[...]

PR #3341, opened 2022-12-01
Make `tools/` a cargo workspace.

jrvanwhy:
### Pull Request Overview

The root cargo workspace was created in PR #1714. #1714's description asked the question "Should `tools/` also be part of the workspace? Have their own workspace?", which doesn't appear to have been answered.

This PR makes `tools/` its own cargo workspace.

### Testing Strategy

Ran `cargo check --workspace --exclude litex-ci-runner` in `tools/` (I excluded `litex-ci-runner` as it requires a dependency I don't have installed).
Ran `cargo check` in the root of the repository.
Ran `make prepush`.

### Open Questions

Do we want to make `tools/` part of the root workspace rather than making it its own workspace? I made it a separate workspace because the compilation profiles in the root workspace are intended for embedded code, whereas `tools/` is host-side code, but I'm open to making it all one workspace. Building host-side tools with embedded profiles will make things a bit slower but shouldn't break anything (unless something relies on panic unwinding).

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,18 +81,7 @@ members = [
     "libraries/tock-register-interface",
     "libraries/tickv",
 ]
-exclude = [
-    "tools/alert_codes",
-    "tools/board-runner",
-    "tools/qemu",
-    "tools/litex-ci-runner",
-    "tools/qemu-runner",
-    "tools/sha256sum",
-    "tools/usb/bulk-echo",
-    "tools/usb/bulk-echo-fast",
-    "tools/usb/bulk-test",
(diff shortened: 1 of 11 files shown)


hudson-ayers (approved):
+1 for tools being their own workspace

bradjc (approved):

alistair23 (approved):

hudson-ayers (approved):

twilfredo (approved):

lschuermann (approved):

PR #3343, opened 2022-12-05
timer: Add support for 64-bit timers

alistair23:
### Pull Request Overview

The current timer implementation doesn't work with 64-bit timers. This PR adds support for the HIL and capsules for 64-bit timers.

When userspace performs a relative expiration syscall we calculate the current time as 64-bits. If userspace passes a reference time it's only a 32-bit value, but we update the top 32-bits to match the current time.

Some libtock-c changes are also required, I will send a PR

### Testing Strategy

Modelling a 64-bit timer

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/alarm.rs
+++ b/capsules/src/alarm.rs
@@ -58,7 +58,7 @@ impl<'a, A: Alarm<'a>> AlarmDriver<'a, A> {
     // underlying alarm is wider than 32 bits.
     fn reset_active_alarm(&self) {
         let mut earliest_alarm = Expiration::Disabled;
-        let mut earliest_end: A::Ticks = A::Ticks::from(0);
+        let mut earliest_end: A::Ticks = A::Ticks::from(0u32);
         // Scale now down to a u32 since that is the width of the alarm;
         // otherwise for larger widths (e.g., u64) now can be outside of
         // the range of what an alarm can be set to.
@@ -140,7 +140,7 @@ impl<'a, A: Alarm<'a>> AlarmDriver<'a, A> {
                 // This uses the invariant that reference <= now.
                 if now_lower_bits.into_u32() < reference {
                     // Build 1<<32 in a way that just overflows to 0 if we are 32 bits
-                    let bit33 = A::Ticks::from(0xffffffff).wrapping_add(A::Ticks::from(0x1));
+                    let bit33 = A::Ticks::from(0xffffffffu32).wrapping_add(A::Ticks::from(0x1u32));
                     high_bits = high_bits.wrapping_sub(bit33);
                 }
                 let real_reference = high_bits.wrapping_add(A::Ticks::from(reference));
--- a/capsules/src/alarm.rs
+++ b/capsules/src/alarm.rs
@@ -4,7 +4,7 @@
 use core::cell::Cell;
 
 use kernel::grant::{AllowRoCount, AllowRwCount, Grant, UpcallCount};
-use kernel::hil::time::{self, Alarm, Frequency, Ticks, Ticks32};
+use kernel::hil::time::{self, Alarm, Frequency, Ticks};
 use kernel::syscall::{CommandReturn, SyscallDriver};
 use kernel::{ErrorCode, ProcessId};
 
@@ -248,15 +248,30 @@ impl<'a, A: Alarm<'a>> SyscallDriver for AlarmDriver<'a, A> {
 
 impl<'a, A: Alarm<'a>> time::AlarmClient for AlarmDriver<'a, A> {
     fn alarm(&self) {
-        let now: Ticks32 = Ticks32::from(self.alarm.now().into_u32());
+        let now = self.alarm.now();
         self.app_alarms.each(|_processid, alarm, upcalls| {
             if let Expiration::Enabled { reference, dt } = alarm.expiration {
-                // Now is not within reference, reference + ticks; this timer
-                // as passed (since reference must be in the past)
-                if !now.within_range(
-                    Ticks32::from(reference),
-                    Ticks32::from(reference.wrapping_add(dt)),
-                ) {
+                // `now` is a 64-bit value, but we are only allowed a 32-bit
+                // `reference` or `dt`.
+                let within_range = if now < u32::MAX.into() {
+                    // The current time fits in a 32-bit value. We can compare
+                    // `now` with `reference + dt` where all values are 64-bit.
+                    now.within_range(
+                        A::Ticks::from(reference),
+                        A::Ticks::from((reference as u64).wrapping_add(dt as u64)),
+                    )
+                } else {
+                    // `now` has 'overflowed" the 32-bit value, so let's mask
+                    // off the top bits to represent a 32-bit overflow.
+                    let now = A::Ticks::from(now.into_u32());
+
+                    now.within_range(
+                        A::Ticks::from(reference),
+                        A::Ticks::from(reference).wrapping_add(dt.into()),
+                    )
+                };
+
+                if !within_range {
                     alarm.expiration = Expiration::Disabled;
                     self.num_armed.set(self.num_armed.get() - 1);
                     upcalls
--- a/capsules/src/buzzer_pwm.rs
+++ b/capsules/src/buzzer_pwm.rs
@@ -102,7 +102,7 @@ impl<'a, A: hil::time::Alarm<'a>, P: hil::pwm::PwmPin> hil::buzzer::Buzzer<'a>
         // Disarm the current alarm and instantly fire another.
         self.alarm.disarm()?;
         // This method was used to reduce the size of the code.
-        self.alarm.set_alarm(self.alarm.now(), A::Ticks::from(0));
+        self.alarm.set_alarm(self.alarm.now(), A::Ticks::from(0u32));
         Ok(())
     }
 }
--- a/capsules/src/read_only_state.rs
+++ b/capsules/src/read_only_state.rs
@@ -73,7 +73,7 @@ impl<'a, T: Time> ContextSwitchCallback for ReadOnlyStateDriver<'a, T> {
                         buf[4..8].copy_from_slice(&(pending_tasks as u32).to_le_bytes());
                     }
                     if buf.len() >= 16 {
-                        let now = self.timer.now().into_usize() as u64;
+                        let now = self.timer.now().into_u64();
                         buf[8..16].copy_from_slice(&now.to_le_bytes());
                     }
                 });
--- a/capsules/src/test/alarm_edge_cases.rs
+++ b/capsules/src/test/alarm_edge_cases.rs
@@ -34,7 +34,7 @@ impl<'a, A: Alarm<'a>> TestAlarmEdgeCases<'a, A> {
         let counter = self.counter.get();
         let delay = self.alarm.ticks_from_ms(self.alarms[counter % 20]);
         let now = self.alarm.now();
-        let start = now.wrapping_sub(A::Ticks::from(10));
+        let start = now.wrapping_sub(A::Ticks::from(10u32));
 
         debug!(
             "{}: Setting alarm to {} + {} = {}",
--- a/capsules/src/virtual_alarm.rs
+++ b/capsules/src/virtual_alarm.rs
@@ -53,7 +53,7 @@ impl<'a, A: Alarm<'a>> ListNode<'a, VirtualMuxAlarm<'a, A>> for VirtualMuxAlarm<
 impl<'a, A: Alarm<'a>> VirtualMuxAlarm<'a, A> {
     /// After calling new, always call setup()
     pub fn new(mux_alarm: &'a MuxAlarm<'a, A>) -> VirtualMuxAlarm<'a, A> {
-        let zero = A::Ticks::from(0);
+        let zero = A::Ticks::from(0u32);
         VirtualMuxAlarm {
             mux: mux_alarm,
             dt_reference: Cell::new(TickDtReference {
@@ -182,7 +182,7 @@ impl<'a, A: Alarm<'a>> Alarm<'a> for VirtualMuxAlarm<'a, A> {
         let extension = if dt_reference.extended {
             Self::Ticks::half_max_value()
         } else {
-            Self::Ticks::from(0)
+            Self::Ticks::from(0u32)
         };
         dt_reference.reference_plus_dt().wrapping_add(extension)
     }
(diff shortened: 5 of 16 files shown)


capsules/src/alarm.rs
```diff
@@ -217,15 +205,32 @@ impl<'a, A: Alarm<'a>> SyscallDriver for AlarmDriver<'a, A> {
...
+                        // This logic handles when the underlying Alarm is wider than
+                        // 32 bits; it sets the reference to include the high bits of now
+                        let mut high_bits = now.wrapping_sub(now_lower_bits);
+                        // If lower bits have wrapped around from reference, this means the
+                        // reference's high bits are actually one less; if we don't subtract
+                        // one then the alarm will incorrectly be set 1<<32 higher than it should.
+                        // This uses the invariant that reference <= now.
+                        if now_lower_bits.into_u32() < reference as u32 {
+                            // Build 1<<32 in a way that just overflows to 0 if we are 32 bits
+                            let bit33 = A::Ticks::from(0xffffffffu32).wrapping_add(A::Ticks::from(0x1u32));
```
twilfredo, line 228:
```suggestion
                            let bit33 = A::Ticks::from(u32::MAX).wrapping_add(A::Ticks::from(0x1u32));
```

alistair23:
> With this approach, systems with 32-bit counters return only 32 bits (in a 64-bit value), while systems with 64-bit counters fill the entire 64 bits

Correct! This way all platforms can request the current time as a 64-bit value. It's then up to the hardware of what that value could be.

> So this is a minimalist tweak that lets RISC-V systems have 64 bits but makes applications not portable across RISC-V and CortexM.

Why is it not portable?

On a RISC-V system, the value of the `uint64_t`/`u64` could be between 0 and 0xFFFF_FFFF_FFFF_FFFF on an ARM system it could be between 0 and 0xFFFF_FFFF. For an application that is just requesting the time and using it for logging, or busy waiting for a certain amount of time (to avoid syscall overhead) I don't see why it matters.

The last change here obviously does require libtock-c changes. I haven't checked if converting from a `SuccessU32` to a `SuccessU64` would break existing users. If we want we can add new 64-bit syscalls and deprecate the existing 32-bit ones. Then we will ensure we don't break any users and but still swap over to the new syscalls.

phil-levis:
> On a RISC-V system, the value of the `uint64_t`/`u64` could be between 0 and 0xFFFF_FFFF_FFFF_FFFF on an ARM system it could be between 0 and 0xFFFF_FFFF. For an application that is just requesting the time and using it for logging, or busy waiting for a certain amount of time (to avoid syscall overhead) I don't see why it matters.

```c
uint64_t deadline = alarm_read() + timeout;
int gpio_val = 1;
/* Spin waiting for a pin to go low or a timeout */
while (alarm_read() < deadline && gpio_val == 1) { 
  gpio_read(STATUS_PIN, &gpio_val);
}
```

If alarm_read() only returns the bottom 32 bits of a 64 bit value, then it's possible additions to this value will be larger than its range, and so the above inequality might never be true.

phil-levis:
I brought this up on the core call; there was significant interest, especially from RISC-V folks, in making 64-bit time standard in userspace, with the caveat that this might lead to a code size increase for 32-bit platforms. One question that will come up is passing system call arguments for 64-bit time on 32-bit platforms (e.g., you can't fit two 64-bit values into command arguments for a 32-bit architecture). I'm going to look into it.

One note, TRD105 does say the kernel needs to provide 64-bit Time but 32-bit Alarm.

alistair23:
Ok, good point. So your example doesn't work where `deadline` overflows, so the real life code would be more complex to handle that case. The entire userspace loop would probably have to look something like this to be generic on all platforms (untested), which is a little clunky. Maybe we could provide a wrapper to help with this.

```c
uint64_t start = alarm_read()
uint64_t deadline = start + timeout;
int gpio_val = 1;

if (deadline < start) {
  // We overflowed the timer
  // Wait until the current time overflows and then try again

  while (alarm_read() > deadline) && gpio_val == 1) { 
    gpio_read(STATUS_PIN, &gpio_val);
  }
}

/* Spin waiting for a pin to go low or a timeout */
while (alarm_read() < deadline && gpio_val == 1) { 
  gpio_read(STATUS_PIN, &gpio_val);

  if (alarm_read() < start) {
    // The alarm has wrapped around, let's update deadline with a wrapped around value

    deadline = (uint32_t)start + (uint32_t)timeout;
  }
}
```

Anyway, it does break backwards compatibility, so we can add new syscalls and deprecate the current ones. I think a helper for "run this while I loop for some amount of time" would be useful.

It's worth noting that the read only system time is already reporting to userspace a 64-bit value.

alistair23:
> One question that will come up is passing system call arguments for 64-bit time on 32-bit platforms (e.g., you can't fit two 64-bit values into command arguments for a 32-bit architecture). I'm going to look into it.

So this PR should handle that. I went with the assumption that we only ever will pass a 64-bit time and a 32-bit dt. So you can only set an alarm 32-bit time in the future, but the current time can be read as a 64-bit time.

alistair23:
I have dropped the last patch, which means there are no userspace visible changes now. Let's merge this and then I'll update the patch to add new syscalls instead of updating the existing ones

capsules/src/alarm.rs
```diff
@@ -248,14 +248,14 @@ impl<'a, A: Alarm<'a>> SyscallDriver for AlarmDriver<'a, A> {
 
 impl<'a, A: Alarm<'a>> time::AlarmClient for AlarmDriver<'a, A> {
     fn alarm(&self) {
-        let now: Ticks32 = Ticks32::from(self.alarm.now().into_u32());
+        let now = self.alarm.now();
```
phil-levis, line 251: I am a little worried about this change. `Expiration` stores values in 32 bits.  But `now` can return a 64-bit value. So what happens on line 256 below if now > 2^32? I think the calculation will be incorrect. For example, suppose

```
reference       = 0xffffff00
dt              = 0x00000110
reference + dt  = 0x00000010
now            = 0x100000020
```

Technically, `now` is after `reference + dt`, so it's not within their range. However, since reference +  dt is considered in the 64-bit space, it will be. This change of moving to the `Ticks` type only works if `Expiration` is also in terms of the `Ticks` type, which is why I chose to stick with 32 bits.

phil-levis (changes requested):

capsules/src/alarm.rs
alistair23: Good catch!

My original PR did update `Expiration` so I missed this. It still would have broken bisectability though, so it needs to be fixed.

Wouldn't something like this fix the issue? That way `now` and `reference + dt` will be the same:

```diff
diff --git a/capsules/src/alarm.rs b/capsules/src/alarm.rs
index 6900c1da1..82a808840 100644
--- a/capsules/src/alarm.rs
+++ b/capsules/src/alarm.rs
@@ -255,7 +255,7 @@ impl<'a, A: Alarm<'a>> time::AlarmClient for AlarmDriver<'a, A> {
                 // as passed (since reference must be in the past)
                 if !now.within_range(
                     A::Ticks::from(reference),
-                    A::Ticks::from(reference.wrapping_add(dt)),
+                    A::Ticks::from((reference as u64).wrapping_add(dt as u64)),
                 ) {
                     alarm.expiration = Expiration::Disabled;
                     self.num_armed.set(self.num_armed.get() - 1);
```

It also aligns with the future `u64` value for `Expiration`

phil-levis: This clearly doesn't work. What happens when `now` is significantly greater than 2^32, i.e. well after the first 32-bit increment? Since `reference` <= 2^32, I this means `now` will not fall within `reference + dt`.

There shouldn't be a plan to change `Expiration` to `u64`. The [TRD105](https://github.com/tock/tock/blob/master/doc/reference/trd105-time.md) is clear that `Alarm` is provided for 32 bits but `Time` is provided for 64 bits. If we want a 64-bit alarm interface that should become a new device/system call driver.

phil-levis: I think at this point it might be good for you to set up a testing configuration. When I redesigned the time system, I'd test by running timer tests for a few days and checking the output. I'd also test 32-bit wraparound more heavily by setting the counter register value to close to wraparound.

alistair23: Hmm... If we aren't allowed a `u64` in `Expiration` it starts to become really hard. As our `reference` will always possibly be less then the current time.

I have pushed an update that handles a `now` greater then 2^32. I'll run some tests on it as well

phil-levis: Tock the system call driver provides 32-bit alarms, but it can support 64-bit time. Why change it to propagate the underlying alarm width into its logic?

alistair23: The current time can be a 64-bit value, so we want to use that for setting alarms, otherwise we might alarm at the wrong time if we ignore the high bits of the current time

phil-levis: You don't ignore the high bits. The TRD is pretty clear that it's 32-bit alarms and 64-bit time.

alistair23: The current implementation is a 32-bit time. This line:

```rust
let now: Ticks32 = Ticks32::from(self.alarm.now().into_u32());
```

gets the current time as a 32-bit value. So we need a change like this to instead use a 64-bit time. The same applies for `reference` which is `now` when we schedule the alarm (so a 64-bit time)

alistair23:
Any more thoughts? Is it worth rebasing this?

alistair23:
Ping!

capsules/src/alarm.rs
```diff
@@ -248,15 +248,30 @@ impl<'a, A: Alarm<'a>> SyscallDriver for AlarmDriver<'a, A> {
...
+                } else {
+                    // `now` has 'overflowed" the 32-bit value, so let's mask
+                    // off the top bits to represent a 32-bit overflow.
+                    let now = A::Ticks::from(now.into_u32());
+
+                    now.within_range(
+                        A::Ticks::from(reference),
+                        A::Ticks::from(reference).wrapping_add(dt.into()),
+                    )
+                };
```
phil-levis, line 272: I'm confused why we need these two cases. If `now < u32::MAX.into()`, can't one still just execute line 266? In use, the variable `now` is always less than `u32::MAX` because its truncated in line 266. But then why does the top case use `u64` and the bottom cases use just `A::Ticks`?

phil-levis (changes requested):

capsules/src/alarm.rs
alistair23: I haven't looked at this for awhile. I'm guessing that it's separated out to make the control flow clearer to understand what is going on and what we are comparing against.

alistair23:
Ping!

phil-levis:
> Ping!

I'm not sure how to proceed here -- I asked why a piece of code was doing what it is doing, your response was "I don't know but can guess." I'm a little wary of merging code like that.

lschuermann:
I think in the discussion around #3975 (and related PRs) we have, at least for the immediate challenges that we had there, decided that we'd use 32 bit as the canonical width for timers exposed to userspace, irrespective of the underlying timer's width. If the timer's width is smaller, we scale its frequency accordingly. If it is larger, we can truncate it.

Having timers that predictably wrap at 2^32 allows userspace to, with relative ease, extend them to arbitrary widths, by scheduling alarms for 2^31 ticks into the future and implementing a manual overflow counter. Would this cover your use-cases @alistair23, or are there other good reasons to have 64 bit timers?

Also, you may want to look at #3975 as it is a major revision to this subsystem, and should make changes like these much more straightforward. It removes many of the conversions between Ticks of different widths entirely, and only converts once when going from/to userspace respectively.

alistair23:
> I'm not sure how to proceed here -- I asked why a piece of code was doing what it is doing, your response was "I don't know but can guess." I'm a little wary of merging code like that.

The problem is that this PR is a year and a half old

I can spend time to read over it and jog my memory, but I'm hesitant to invest more time in this, just to have it stall for another year.

> Having timers that predictably wrap at 2^32 allows userspace to, with relative ease, extend them to arbitrary widths, by scheduling alarms for 2^31 ticks into the future and implementing a manual overflow counter. Would this cover your use-cases @alistair23, or are there other good reasons to have 64 bit timers?

That results in a bunch of extra wake ups though.

We needed a 64-bit timer (hence this PR). I also remember a few lowRISC people were keen to see this merged, so they might want a 64-bit timer as well.

Realistically all RISC-V users will probably want a 64-bit timer as you have the hardware so not utilising it seems a bit silly

lschuermann:
@alistair23 Gotcha. We actually discussed this on the call last Friday, I'll try to have notes up soon. tl;dr: the best way forward seems to be a dedicated, separate alarm capsule for 64-bit timers.

I think we're all convinced that 64-bit timers in userspace seem generally advantageous to some people, in particular when they know that they're on hardware that has 64-bit timers underneath. However, this is only a subset of the users of Tock.

Our current strategy to support hardware timers of differing widths is to left-justify them to a 32-bit value, such that userspace can reliably predict when they are going to wrap. This works well for 32-bit, 24-bit or 16-bit timers. It even works well for 64-bit hardware timers, at the expense of requiring userspace to implement wrapping logic itself. Left-justifying something like an, e.g., 24-bit hardware timer to a 64-bit value, though, is rather wasteful as it becomes a very low-precision timer that also incurs using 64-bit arithmetic everywhere.

It seems like there is not a single, one-size fits all solution here. Tock's current 32-bit timer interface arguably works well for the majority of platforms. For those use-cases where a 64-bit timer would be advantageous, it seems that the best way forward is a second alarm capsule, that accepts only types of `A: <Alarm<'_> as Time>::Ticks = Ticks64` and exposes a proper 64-bit timer to userspace. That new timer will also not have to implement any wrapping logic and will thus be much simpler than the existing alarm capsule, at the expense of only working in the presence of 64-bit timers in the kernel.

lschuermann:
As a side note, it definitely also seems possible to have a userspace driver that probes the existence of the 64-bit alarm capsule, and falls back onto the 32-bit one when it's not present. Or even have this be a compile-time flag for userspace builds. Really the only thing we'd like to avoid is to have every Tock board need to expose a 64-bit timer to userspace, hence the need for a different capsule.

lschuermann:
@alistair23 I'm closing this as per the above, please feel free to re-open if there's something we haven't addressed. I think we'd be happy about a dedicated 64-bit timer capsule that's distinct from the alarm capsule.

Here's the notes of the call where we talked about this change.

PR #3344, opened 2022-12-05
Update `VolatileCell` using the `vcell` crate's `VolatileCell`.

jrvanwhy:
### Pull Request Overview

Tock's `VolatileCell` originally came from the [zinc.rs](https://github.com/hackndev/zinc) project. zinc.rs is no longer maintained. This updates `VolatileCell` by switching to the version from the [vcell](https://github.com/japaric/vcell) crate, which appears to still be maintained.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/libraries/tock-cells/src/volatile_cell.rs
+++ b/libraries/tock-cells/src/volatile_cell.rs
@@ -9,7 +9,6 @@
 
 #![deny(missing_docs)]
 #![deny(warnings)]
-#![no_std]
 
 use core::cell::UnsafeCell;
 use core::ptr;
@@ -18,6 +17,7 @@ use core::ptr;
 ///
 /// [`Cell`]: https://doc.rust-lang.org/std/cell/struct.Cell.html
 /// [volatile]: https://doc.rust-lang.org/std/ptr/fn.read_volatile.html
+#[derive(Default)]
 #[repr(transparent)]
 pub struct VolatileCell<T> {
     value: UnsafeCell<T>,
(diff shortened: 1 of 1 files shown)


libraries/tock-cells/src/volatile_cell.rs
```diff
@@ -1,81 +1,93 @@
-//! Implementation of a type for accessing MCU registers.
+//! A type for accessing MMIO registers. `VolatileCell` is just like [`Cell`]
+//! but with [volatile] read / write operations
+//!
+//! [`Cell`]: https://doc.rust-lang.org/std/cell/struct.Cell.html
+//! [volatile]: https://doc.rust-lang.org/std/ptr/fn.read_volatile.html
```
lschuermann, line 5: This should probably say where the code was vendored from and under which license it was placed originally.

lschuermann (changes requested):
Looks good to me generally.

libraries/tock-cells/src/volatile_cell.rs
jrvanwhy: I added a comment indicating where the code was vendored from. A license header will need to wait for #3318.

lschuermann (approved):
Thanks!

twilfredo (dismissed):

hudson-ayers (dismissed):

alistair23 (approved):

jrvanwhy:
@alevy have you seen this?

bradjc:
Does this change anything other than the order of the methods?

Why do we want to add `as_ptr()`?

alevy (approved):

jrvanwhy:
> Does this change anything other than the order of the methods?

It also changes the license on the code from Apache 2.0 to Apache 2.0 or MIT.

> Why do we want to add `as_ptr()`?

We don't need `as_ptr()`, and I can remove it if you'd prefer.

bradjc:
I don't think we should add it without explaining why. I guess I would side on not adding it.

jrvanwhy:
> I don't think we should add it without explaining why. I guess I would side on not adding it.

So would you like me to push a commit removing `as_ptr()`?

jrvanwhy:
I removed `as_ptr`.

hudson-ayers (approved):

lschuermann (approved):

PR #3345, opened 2022-12-05
Add a license checking tool.

jrvanwhy:
### Pull Request Overview

This pull request adds a tool that checks for license headers as specified in #3318.

### Testing Strategy

I adapted `ci-job-tools` to run tests in the `tools/` workspace, and ran the tool by hand to verify it reports errors + success correctly.

### TODO or Help Wanted

#3318 is undergoing revision.

I currently have the license checker disabled on all files *except new files added in this PR*, because otherwise it does its job and fails CI.

Also, after license headers are added, I need to evaluate the speed of the tool. It currently runs in 100 milliseconds, which seems fast enough, but it may slow down when license headers are added (it may also speed up). If it slows down too much (e.g. takes over a second), I'll revert to a simpler -- but less accurate -- parsing implementation.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/.lcignore
+++ b/.lcignore
@@ -8,7 +8,7 @@
 
 # TODO: This disables the license checker over most of the repository, enabling
 # it on a handful of files for demonstration purposes. The next few lines should
-# be removed when license headers are removed.
+# be removed when license headers are added.
 *
 !/.lcignore
 !/tools/
--- a/.lcignore
+++ b/.lcignore
@@ -1,0 +1,7 @@
+# Licensed under the Apache License, Version 2.0 or the MIT License.
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Copyright Tock Contributors 2022
+# Copyright Google LLC 2022
+
+/.git/
+*.md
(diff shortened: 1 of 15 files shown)


alevy (approved):

alevy:
@jrvanwhy I'm not sure it's worth worrying about performance if we're talking about +/- a few seconds in the CI

hudson-ayers:
> @jrvanwhy I'm not sure it's worth worrying about performance if we're talking about +/- a few seconds in the CI

This PR also adds it to the `prepush` Make rule, so its a few seconds before every `git push` -- I still think a few seconds doesn't matter there, but ~30 seconds would imo

hudson-ayers (approved):
As discussed on the call today, this is fine to go in before https://github.com/tock/tock/pull/3318/ because it is not currently set to enforce anything for most files, and can always be adapted to reflect any additional changes to that PR before it is merged.

lschuermann (approved):

brghena (approved):

PR #3358, opened 2022-12-19
Enable virtual function elimination by default for OpenTitan

jwnrt:
### Pull Request Overview

This PR:
* Updates OpenTitan's tests to use `-Zbuild-std`, fixing `VFUNC_ELIM=1 make test`.
* Enables the `VFUNC_ELIM=1` option by default for OpenTitan's earlgrey board.
* Moves the default board configuration into the Cargo manifest, fixing `make test-verilator`.

Virtual function elimination (`VFUNC_ELIM`) currently miscompiles if crates built without VFE are mixed into the LTO unit. This includes the pre-built `core`, which creates and uses trait objects (virtual functions). Building `core` from source enables VFE.

Moving the default board configuration (`fpga_cw310`) into the Cargo manifest makes it easier to disable when running Verilator tests. Configurations are mutually exclusive and conflict. Previously, `BOARD_CONFIGURATION=sim_verilator` had to be set to enable `make test-verilator`.

### Testing Strategy

Running tests for `opentitan/earlgrey-cw310`:
1. `make test`.
2. `VFUNC_ELIM=1 make test` (was previously failing to link).
3. `make test-verilator` (previously required `BOARD_CONFIGURATION=sim_verilator`).

### TODO or Help Wanted

Should other boards be updated to use `-Zbuild-std` for tests too?

- [ ] `apollo3/redboard_artemis`.
- [ ] `apollo3/lora_things_plus`.
- [ ] `esp32-c3-devkitM-1`.

### Documentation Updated

The `-Zpanic-abort-tests` flag's purpose is documented in the `Makefile`.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/earlgrey-cw310/Makefile
+++ b/boards/opentitan/earlgrey-cw310/Makefile
@@ -1,6 +1,5 @@
 # Makefile for building the tock kernel for the OpenTitan platform
 
-DEFAULT_BOARD_CONFIGURATION=fpga_cw310
 TARGET=riscv32imc-unknown-none-elf
 PLATFORM=earlgrey-cw310
 FLASHID=--dev-id="0403:6010"
@@ -20,8 +19,6 @@ include ../../Makefile.common
 # Cargo `--features`. Please see `Cargo.toml` for available options.
 ifneq ($(BOARD_CONFIGURATION),)
 	CARGO_FLAGS += --features=$(BOARD_CONFIGURATION)
-else
-	CARGO_FLAGS += --features=$(DEFAULT_BOARD_CONFIGURATION)
 endif
 
 # Build test crates with the `panic = "abort"` strategy. Tests use `unwind` by
@@ -112,4 +109,4 @@ test-verilator: ot-check
 	mkdir -p $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/deps/
 	$(Q)cp test_layout.ld $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/deps/layout.ld
 	$(Q)cp ../../kernel_layout.ld $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/
-	$(Q)RUSTFLAGS="$(RUSTC_FLAGS_TOCK)" VERILATOR="yes" OBJCOPY=${OBJCOPY} TOCK_ROOT_DIRECTORY=${TOCK_ROOT_DIRECTORY} TARGET=${TARGET} $(CARGO) test $(CARGO_FLAGS_TOCK) $(NO_RUN) --bin $(PLATFORM) --release --features=hardware_tests --features sim_verilator
+	$(Q)RUSTFLAGS="$(RUSTC_FLAGS_TOCK)" VERILATOR="yes" OBJCOPY=${OBJCOPY} TOCK_ROOT_DIRECTORY=${TOCK_ROOT_DIRECTORY} TARGET=${TARGET} $(CARGO) test $(CARGO_FLAGS_TOCK) $(NO_RUN) --bin $(PLATFORM) --release --no-default-features --features=hardware_tests,sim_verilator
--- a/boards/opentitan/earlgrey-cw310/Makefile
+++ b/boards/opentitan/earlgrey-cw310/Makefile
@@ -11,6 +11,8 @@ QEMU ?= ../../../tools/qemu/build/qemu-system-riscv32
 QEMU_ENTRY_POINT=0x20000400
 # Latest supported commmit of OpenTitan
 OPENTITAN_SUPPORTED_SHA := d072ac505f82152678d6e04be95c72b728a347b8
+# Enable virtual function elimination
+VFUNC_ELIM=1
 
 
 include ../../Makefile.common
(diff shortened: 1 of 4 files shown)


boards/opentitan/earlgrey-cw310/Makefile
```diff
@@ -20,10 +19,12 @@ include ../../Makefile.common
 # Cargo `--features`. Please see `Cargo.toml` for available options.
 ifneq ($(BOARD_CONFIGURATION),)
 	CARGO_FLAGS += --features=$(BOARD_CONFIGURATION)
-else
-	CARGO_FLAGS += --features=$(DEFAULT_BOARD_CONFIGURATION)
 endif
 
+# Build test crates with the `panic = "abort"` strategy. Tests use `unwind` by
+# default which causes incompatability with `core` built via `-Zbuild-std`.
+CARGO_FLAGS += -Zpanic-abort-tests
```
hudson-ayers, line 26: An alternative to this would be for tests to be built using `-Z build-std=core,std,alloc,test`, which would enable virtual function elimination without breaking unwinding. We would probably have to modify the main Makefile a bit to make it ergonomic to only use this configuration for tests. I don't think any of the tests in the kernel repo currently rely on unwinding, but I know we do rely on unwinding for libtock-rs unit tests already (e.g. https://github.com/tock/libtock-rs/blob/d07eaa2df10b07668b2ed51cf993b9fd50951545/unittest/src/fake/syscalls/command_impl_tests.rs#L66)

hudson-ayers (commented):
This generally looks good. I am curious about the motivation for this change -- the primary reason to use `VFUNC_ELIM` is for code size savings. Are the test binaries becoming big enough that size is an important constraint? Or is the idea that opentitan will likely want to enable this by default, so running tests under this configuration helps make sure that there are not breakages the tests fail to catch?

If the latter, perhaps we should also change the Makefile to build OT using virtual function elimination by default?

jwnrt:
> Or is the idea that opentitan will likely want to enable this by default, so running tests under this configuration helps make sure that there are not breakages the tests fail to catch?

Yes, this is generally the motivation. We'd like the code savings from VFE, but in its current state we get some miscompilations. Getting the tests to run with VFE makes it a little easier to catch these in cases where the link still succeeds.

> perhaps we should also change the Makefile to build OT using virtual function elimination by default?

Ultimately it feels right to have it enabled by default, but whether it's ready right now I'm not sure. I don't think Tock is getting any miscompilations as long as it's using `build-std`.

boards/opentitan/earlgrey-cw310/Makefile
jwnrt: I looked into this briefly but had some trouble disabling atomics support while building `alloc` for `riscv32imc`. Looks like [Rust has a `no_sync`](https://github.com/rust-lang/rust/blob/3f2b2eee8f46f2252d2919d7a57bf3068d7df285/library/alloc/src/lib.rs#L241) config but I wasn't sure how to set this through `build-std-[features]`.

I'm not sure that the problem is missing `libtest` since we don't seem to link to it anyway due to using a custom `#![test_runner]` in the binary crate. Is the idea that we get Tock itself to support unwinding when building with tests, and that requires `std`?

hudson-ayers: Hmm, I think if this looks challenging the current approach is fine. I hadn't really thought all the way through the idea of this -- my thought process was basically "hmm, don't we have tests that rely on unwinding -- wait no, that was in libtock-rs -- still, that suggests unwinding could be useful for tests -- can't we just build the whole std library with build-std?". 

Thinking a bit more about it now, libtock-rs actually only uses unwinding for on-host tests, and it seems unlikely we would have any on-target tests that would use unwinding, especially in the kernel. So I will go ahead and mark this resolved

hudson-ayers:
> > perhaps we should also change the Makefile to build OT using virtual function elimination by default?
> 
> Ultimately it feels right to have it enabled by default, but whether it's ready right now I'm not sure. I don't think Tock is getting any miscompilations as long as it's using `build-std`, so it might not be a problem.


I think it makes sense for us to test the same configuration that we provide to users by default -- it seems possible that virtual function elimination could also hide some other soundness bug, and it would suck to miss that because we test a different configuration than we deploy. Also, if upstream OT is going to use VFE, it probably is good for Tock's release testing etc. to test this configuration. Would you mind making that change as part of this PR, and re-titling the PR to reflect that it includes that change?

hudson-ayers (approved):

boards/opentitan/earlgrey-cw310/Makefile
jrvanwhy: > I looked into this briefly but had some trouble disabling atomics support while building `alloc` for `riscv32imc`. Looks like [Rust has a `no_sync`](https://github.com/rust-lang/rust/blob/3f2b2eee8f46f2252d2919d7a57bf3068d7df285/library/alloc/src/lib.rs#L241) config but I wasn't sure how to set this through `build-std-[features]`.

I don't expect any embedded code in the `tock/tock` repository to use `alloc`, so we should be able to dodge that problem here. That is a problem for `libtock-rs`, as it should eventually have an optional dependency on `alloc`, but we can cross that bridge when we get to it.

> Thinking a bit more about it now, libtock-rs actually only uses unwinding for on-host tests, and it seems unlikely we would have any on-target tests that would use unwinding

Correct, I don't depend on unwinding in embedded code. I'm not sure we want unwinding support -- I think it has an adverse impact on code size.

alistair23 (approved):

PR #3330, opened 2022-11-18
boards/apollo3: lora_things_plus: Expose SPI to userspace for SX1262

alistair23:
### Pull Request Overview

This builds on https://github.com/tock/tock/pull/3324 and adds support to communicate with the SX1262 LoRa chip via SPI.

### Testing Strategy

Communicating with SX1262

### Documentation Updated

- [X] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [X] Ran `make prepush`.

PR #3360, opened 2022-12-20
lora-phy: Add support for a LoRa capsule

alistair23:
### Pull Request Overview

This builds on top of https://github.com/tock/tock/pull/3330 and adds a LoRa capsule.

The capsule is pretty simple and mostly just exposes the SPI device and some GPIO pins. This could pretty much all be done with the current SPI and GPIO capsules, but this has a few advantages.

 * It allows us to expose other SPI busses to userspace
 * It hides the GPIO numbering so that it doesn't change per board. This allows the userspace code to be board independent

With this PR, support in libtock-c and LoRaMac-node we can run LoRa code on Tock

### Testing Strategy

I have tested communicating and interfacing with the LoRa device, using [libtock-c](https://github.com/tock/libtock-c/pull/317)

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/chips/apollo3/src/gpio.rs
+++ b/chips/apollo3/src/gpio.rs
@@ -155,6 +155,96 @@ impl Port<'_> {
         }
     }
 
+    /// This function configures some GPIO pins on the Apollo3 to allow
+    /// communication with a SX1262 LoRa module.
+    ///
+    /// The pin mapping is setup to match what is used by the NM180100 SoC (shown below)
+    ///
+    /// IOM3: Semtech SX1262
+    ///     Apollo 3 Pin Number | Apollo 3 Name | SX1262 Pin Number | SX1262 Name | SX1262 Description
+    ///                      H6 |       GPIO 36 |                19 |  NSS        | SPI slave select
+    ///                      J6 |       GPIO 38 |                17 |  MOSI       | SPI slave input
+    ///                      J5 |       GPIO 43 |                16 |  MISO       | SPI slave output
+    ///                      H5 |       GPIO 42 |                18 |  SCK        | SPI clock input
+    ///                      J8 |       GPIO 39 |                14 |  BUSY       | Radio busy indicator
+    ///                      J9 |       GPIO 40 |                13 |  DIO1       | Multipurpose digital I/O
+    ///                      H9 |       GPIO 47 |                6  |  DIO3       | Multipurpose digital I/O
+    ///                      J7 |       GPIO 44 |                15 |  NRESET     | Radio reset signal, active low
+    ///
+    /// This should be used by the lora_things_plus board or any other board using
+    /// the Apollo3 based NM180100 SoC. This function would also work for any
+    /// Apollo3 board using the same pins and IOM as specified above.
+    pub fn enable_sx1262_radio_pins(&self) {
+        let regs = GPIO_BASE;
+
+        regs.padkey.set(115);
+
+        // Pin 36 NSS
+        regs.padreg[9].modify(
+            PADREG::PAD0PULL::CLEAR + PADREG::PAD0INPEN::CLEAR + PADREG::PAD0FNCSEL.val(0x1),
+        );
+        regs.cfg[4].modify(CFG::GPIO4INCFG.val(0x00) + CFG::GPIO4OUTCFG.val(0x00));
+        regs.altpadcfgj
+            .modify(ALTPADCFG::PAD0_DS1::CLEAR + ALTPADCFG::PAD0_SR::CLEAR);
+
+        // Pin 39 Busy
+        regs.padreg[9].modify(
+            PADREG::PAD3INPEN::SET
+                + PADREG::PAD3STRNG::CLEAR
+                + PADREG::PAD3FNCSEL.val(0x3)
+                + PADREG::PAD3RSEL.val(0x0),
+        );
+        regs.cfg[4].modify(
+            CFG::GPIO7INCFG.val(0x00) + CFG::GPIO7OUTCFG.val(0x00) + CFG::GPIO7INTD.val(0x00),
+        );
+        regs.altpadcfgj
+            .modify(ALTPADCFG::PAD3_DS1::CLEAR + ALTPADCFG::PAD3_SR::CLEAR);
+
+        // Pin 40 DIO1
+        regs.padreg[10].modify(
+            PADREG::PAD0PULL::CLEAR
+                + PADREG::PAD0INPEN::SET
+                + PADREG::PAD0STRING::CLEAR
+                + PADREG::PAD0FNCSEL.val(0x3)
+                + PADREG::PAD0RSEL.val(0x0),
+        );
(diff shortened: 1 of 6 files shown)


twilfredo (dismissed):

lschuermann:
I think this PR hits on the issue of how generic or specific we'd like the userspace/kernel boundary and corresponding driver interfaces to be. In my humble opinion, this new `lora_node` driver seems to cross the line here: it doesn't add any new API which could not be represented through existing drivers, and it seems also rather specific for a given (class of) LoRa module(s), specifically the SX1262. Adding this as a separate capsule appears to have some benefits (as listed above), but more importantly adds a significant amount of code and the potential for this implementation to diverge from the other SPI / GPIO capsules.

This is not a strong opinion I'm voicing, especially given we're talking about an addition to capsules here. But it seems that a much better solution would be to -- in the long term -- have a standardized mechanism of exposing multiple instances of a given peripheral to userspace, perhaps using a driver discovery mechanism. We'd still be able to have a separate SPI and GPIO driver just for this peripheral, but would not need to vendor an entirely new capsule. I proposed something like this (#1735) a long time ago, perhaps we could revisit that.

phil-levis:
In the past, when we've implemented radio drivers in userspace (e.g., the original RF233 driver was a port from Atmel's Contiki code), we've used the standard SPI and GPIO system call APIs. Why not do that here? Is this because you want to keep one SPI for other use, and expose a second SPI for the radio?

Way back when (5 years ago?), we agreed that the way to do this is to use the high order bits of system call numbers. So that way you can demux on which bus to use for a command by looking at this most significant bits. Shane Leonard (who wrote the first components interface, the registers crate and the Teensy 3.6 port) proposed this but graduated before he implemented it.

alistair23:
> it doesn't add any new API which could not be represented through existing drivers

That's true. We are just adding APIs that do already exist. The problem with the current APIs though is that they aren't consistent across boards.

For example we need to toggle some specific GPIOs. Using the current GPIO driver these could be a different GPIO pin on each board. So we are unable to write code that can run on multiple boards. Instead we need to have a board specific port for each board.

As well as that, there is the limitation in Tock userspace that we can only have a single driver type. So this will use the entire SPI driver for the bus the LoRa is connecte to, and we are then unable to use other SPI buses.

> and it seems also rather specific for a given (class of) LoRa module(s), specifically the SX1262

This would work with all SemTech LoRa modules (which is all LoRa modules)

> have a standardized mechanism of exposing multiple instances of a given peripheral to userspace, perhaps using a driver discovery mechanism

I think in the long term that would also address the issue. But we are currently a very long way away from that. If we get to a point where this is possible we can deprecate this crate and switch over to using a discovery mechanism instead.

alistair23:
> we've used the standard SPI and GPIO system call APIs. Why not do that here? Is this because you want to keep one SPI for other use, and expose a second SPI for the radio?

As the PR description mentions, there are two main reasons.

One is that in this boards case the LoRa module is on a dedicated SPI bus. So we want to expose a different SPI bus to userspace. This will allow userspace to communicate with other devices connected by the SPI bus broken out on the board, while still using the LoRa specific SPI.

The second is that the GPIO numbering will be different across different boards. If we use the standard GPIO capsule we can't write portable code as each board will have different GPIO mappings.

> Way back when (5 years ago?), we agreed that the way to do this is to use the high order bits of system call numbers. So that way you can demux on which bus to use for a command by looking at this most significant bits.

That would require a pretty substantial change to Tock capsules to handle multiple underlying drivers. But it then has the same problem of board compatibility. How do you know which SPI to use on different boards?

alistair23:
I wanted to come back to this. I think it does make sense to add a new API for a complex enough syscall because of the benefits I mentioned above. With a syscall like this we can then run the exact same application on any board and it will just work (besides architectures obviously).

If in the future Tock adds a discovery mechanism we can always just drop this capsule and use the new discovery mechanism. The libtock-c interface wouldn't change so we wouldn't break existing app (unless they use raw syscalls). I do feel that a mechanism like that is a long way off if it ever even happens.

hudson-ayers:
I'll try to bring this PR up on Fridays call, I agree that a driver discovery mechanism is a ways off, and I do think that even basic LoRa support could get some additional people to try Tock

alistair23:
@bradjc That works for me as well. I liked combining them as then there is a single LoRa capsule, but splitting it in two still addresses the main issues.

I don't think we mention LoRaWAN anywhere in this PR

brghena:
The name "loramac-node" is from the LoRaWAN implementation by Semtech: https://github.com/Lora-net/LoRaMac-node So I think it would imply to outsiders that you support LoRaWAN rather than just the LoRa PHY.

alistair23:
Ah ok. I can change the name to lora-phy

alistair23:
Updated to address the comments

chips/apollo3/src/gpio.rs
```diff
@@ -116,7 +116,7 @@ impl Port<'_> {
         let mut count = 0;
         while irqs != 0 && count < self.pins.len() {
             if (irqs & 0b1) != 0 {
-                self.pins[count].handle_interrupt();
+                self.pins[count + 32].handle_interrupt();
```
brghena, line 119: Can you add a comment explaining where the "32" comes from here?

brghena (approved):
Just one nit, but I think this looks great

hudson-ayers (approved):
I think this looks like a nice approach until we have a driver registry, then we can revisit.

chips/apollo3/src/gpio.rs
alistair23: Done!

hudson-ayers (dismissed):

chips/apollo3/src/gpio.rs
```diff
@@ -155,6 +157,77 @@ impl Port<'_> {
         }
     }
 
+    pub fn enable_sx1262_radio_pins(&self) {
```
bradjc, line 160: This is not ideal, but it seems like the GPIO controller on this chip is so weird that it kind of forces code like this.

Can you at least comment this function explaining that it is board-specific and would ideally be made generic in the future?

bradjc (commented):
I really like where this PR is now. This is a really nice solution to the current challenge of supporting multiple SPI busses and portable GPIO applications, and our design of allowing boards to select the driver nums is key here.

phil-levis (approved):
Looks great!

chips/apollo3/src/gpio.rs
alistair23: Yeah, the GPIO controller is very flexible, which also means we need to configure it for every use case.

Eventually we can collapse this all into a generic function for this chip (which is what the HAL does) but for the time being I find these explicit functions simpler to follow

alistair23: I added a comment though :)

alistair23:
Added a comment to the function as requested by Brad. This should be good to go now :)

PR #3367, opened 2023-01-05
PWM Syscall Driver

TeodoraMiu:
### Pull Request Overview

This pull request adds a syscall driver for PWM, allowing driver users to control a PWM pin's frequency and duty cycle. 

### Testing Strategy

This pull request was tested with a Microbit.

### TODO or Help Wanted

Feedback is much appreciated.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/pwm.rs
+++ b/capsules/src/pwm.rs
@@ -18,7 +18,7 @@ pub struct Pwm<'a, const NUM_PINS: usize> {
     /// Per-app state.
     apps: Grant<App, UpcallCount<1>, AllowRoCount<0>, AllowRwCount<0>>,
     /// An array of apps associated to their reserved pins.
-    active_app: [OptionalCell<ProcessId>; NUM_PINS],
+    active_process: [OptionalCell<ProcessId>; NUM_PINS],
 }
 
 impl<'a, const NUM_PINS: usize> Pwm<'a, NUM_PINS> {
@@ -31,15 +31,15 @@ impl<'a, const NUM_PINS: usize> Pwm<'a, NUM_PINS> {
         Pwm {
             pwm_pins: pwm_pins,
             apps: grant,
-            active_app: [EMPTY; NUM_PINS],
+            active_process: [EMPTY; NUM_PINS],
         }
     }
 
-    pub fn is_valid_app(&self, processid: ProcessId, pin: usize) -> bool {
+    pub fn claim_pin(&self, processid: ProcessId, pin: usize) -> bool {
         // Attempt to get the app that is using the pin.
-        self.active_app[pin].map_or(true, |id| {
-            // If the app is empty, that means that there is no app currently using the pins,
-            // therefore the pin is usable by the new app
+        self.active_process[pin].map_or(true, |id| {
+            // If the app is empty, that means that there is no app currently using this pin,
+            // therefore the pin could be usable by the new app
             if id == &processid {
                 // The same app is trying to access the pin it has access to, valid
                 true
@@ -59,7 +59,7 @@ impl<'a, const NUM_PINS: usize> SyscallDriver for Pwm<'a, NUM_PINS> {
     ///
     /// - `0`: Return number of PWM pins if this driver is included on the platform.
     /// - `1`: Start the PWM pin output. First 16 bits of `data1` are used for the duty cycle, as a
-    ///     percentage with 2 decimals, and the last 16 bits of `data2` are used for the PWM channel
+    ///     percentage with 2 decimals, and the last 16 bits of `data1` are used for the PWM channel
     ///     to be controlled. `data2` is used for the frequency in hertz. For the duty cycle, 100% is
     ///     the max duty cycle for this pin.
     /// - `2`: Stop the PWM output.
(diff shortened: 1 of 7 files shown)


capsules/src/pwm.rs
```diff
@@ -0,0 +1,137 @@
...
+        data1: usize,
+        data2: usize,
+        processid: ProcessId,
+    ) -> CommandReturn {
+        match command_num {
+            // Check whether the driver exists, return number of usable PWM pins.
+            0 => CommandReturn::success_u32(NUM_PINS as u32),
+
+            // Start the pwm output.
+            1 => {
```
alexandruradovici, line 79: Please explain similar to the [touch driver](https://github.com/tock/tock/blob/602ede370120961b3ad772fcc5b8be1809cff698/capsules/src/touch.rs#L31-L36) how data is sent to the driver.

capsules/src/pwm.rs
```diff
@@ -0,0 +1,137 @@
...
+#[derive(Default)]
+pub struct App;
+
+pub struct Pwm<'a, const NUM_PINS: usize> {
+    /// The usable pwm pins.
+    pwm_pins: &'a [&'a dyn hil::pwm::PwmPin; NUM_PINS],
+    /// Per-app state.
+    apps: Grant<App, UpcallCount<1>, AllowRoCount<0>, AllowRwCount<0>>,
+    /// An array of apps associated to their reserved pins.
+    active_app: [OptionalCell<ProcessId>; NUM_PINS],
```
alexandruradovici, line 21:
```suggestion
    active_process: [OptionalCell<ProcessId>; NUM_PINS],
```

alexandruradovici (changes requested):
Add the syscall driver documentation to https://github.com/tock/tock/tree/master/doc/syscalls.

capsules/src/pwm.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn command(
+        &self,
+        command_num: usize,
+        data1: usize,
+        data2: usize,
+        processid: ProcessId,
+    ) -> CommandReturn {
+        match command_num {
+            // Check whether the driver exists, return number of usable PWM pins.
+            0 => CommandReturn::success_u32(NUM_PINS as u32),
```
phil-levis, line 76: Please separate out these two pieces of functionality. 0 should be just an "exists" call, there should be a separate command for how many pins.

capsules/src/pwm.rs
```diff
@@ -0,0 +1,137 @@
...
+        assert!(NUM_PINS <= (u16::MAX as usize));
+        const EMPTY: OptionalCell<ProcessId> = OptionalCell::empty();
+        Pwm {
+            pwm_pins: pwm_pins,
+            apps: grant,
+            active_app: [EMPTY; NUM_PINS],
+        }
+    }
+
+    pub fn is_valid_app(&self, processid: ProcessId, pin: usize) -> bool {
```
phil-levis, line 38: Ah -- I think this method is a bit confusing. I assumed it was checking the validity of the application (e.g., is running), rather than whether the app can acquire the pin. I'd suggest changing this to `claim_pin`, and add a `release_pin`.

capsules/src/pwm.rs
```diff
@@ -0,0 +1,137 @@
...
+            1 => {
+                let pin = data1 & ((1 << 16) - 1);
+                let duty_cycle = data1 & (((1 << 16) - 1) << 16);
+                let frequency_hz = data2;
+
+                if pin >= NUM_PINS {
+                    // App asked to use a pin that doesn't exist.
+                    CommandReturn::failure(ErrorCode::INVAL)
+                } else {
+                    if !self.is_valid_app(processid, pin) {
```
phil-levis, line 88: It was not clear to me at first reading that this is a lock/guard. Please see comment above.

phil-levis (changes requested):
I think this wants to be more explicit that there's a per-pin lock; it would help to note that this is FCFS and not virtualized (of course).

capsules/src/pwm.rs
alexandruradovici: I kind of disagree with this, as:
- ADC https://github.com/tock/tock/blob/500caa5e882b7efe40158849fb489205f07c6a2e/capsules/src/adc.rs#L1149
- GPIO https://github.com/tock/tock/blob/500caa5e882b7efe40158849fb489205f07c6a2e/capsules/src/gpio.rs#L198
- LED https://github.com/tock/tock/blob/500caa5e882b7efe40158849fb489205f07c6a2e/capsules/src/led.rs#L95

are implemented in the same way.

It is true that the TRD104 specifies that [command 0](https://github.com/tock/tock/blob/master/doc/reference/trd104-syscalls.md#431-command-identifier-0) should always return `Success`, but this might not be clear if it is the success class of return values or the actual success without payload.

While I do not favor one or the other, we should keep implementations consistent for all drivers.

phil-levis: Ah. Hunh. If this is consistent with some major system call capsules, then I am OK with it. My concern was about consistency, not that one way or the other is better.

phil-levis (approved):

capsules/src/pwm.rs
alexandruradovici: @TeodoraMiu please revert this so that command 0 returns the number of PWM pins

capsules/src/pwm.rs
```diff
@@ -0,0 +1,150 @@
...
+        match command_num {
+            // Check whether the driver exists.
+            0 => CommandReturn::success(),
+
+            // Return number of usable PWM pins.
+            1 => CommandReturn::success_u32(NUM_PINS as u32),
+
+            // Start the pwm output.
+
+            // data1 format
```
alexandruradovici, line 88:
```suggestion
            // data1 stores the duty cycle and the pin number in the format
            // data2 stores the frequency
```

Please add the explanation about why you chose this format

capsules/src/pwm.rs
```diff
@@ -0,0 +1,150 @@
...
+            // +------------------+---------------+
```
alexandruradovici, line 89:
```suggestion
            // +------------------+------------------+
```

doc/syscalls/00010_pwm.md
```diff
@@ -0,0 +1,71 @@
...
+
+    **Argument 2**: unused
+
+    **Returns**: The number of PWM pins on this board.
+
+  * ### Command number: `2`
+
+    **Description**: Start the PWM output.
+
+    **Argument 1**: First 16 bits represent the duty cycle, as a percentage with 2 decimals (100% being the maximum possible duty cycle), and the last 16 bits represent the PWM pin to be controlled.
```
alexandruradovici, line 39: Please add the explanation about why you chose this format

alexandruradovici (changes requested):

capsules/src/pwm.rs
hudson-ayers: It seems like all of these capsules are potentially non-compliant with trd104: https://github.com/tock/tock/blob/master/doc/reference/trd104-syscalls.md#431-command-identifier-0 . I vaguely recall some discussion of this back around Tock 2.0, and I think there was a desire for userspace to be able to assume the same system call return variant for all driver-exists checks

hudson-ayers: cc @lschuermann I think you may have been a part of those discussions?

alexandruradovici:
As there are several drivers using it this way, I suggest merging this as is and, if needed, opening another pull request that fixes all of them. We need the PWM for the Rust Nation presentation.

hudson-ayers (approved):

hudson-ayers:
> As there are several drivers using it this way, I suggest merging this as is and, if needed, opening another pull request that fixes all of them. We need the PWM for the Rust Nation presentation.

SGTM

alexandruradovici (approved):

PR #3368, opened 2023-01-05
ADC pending command functionality

TeodoraMiu:
### Pull Request Overview

This pull request changes the way the ADC capsule receives commands from apps. Before this PR, the ADC capsule could only handle one command at a time. This PR would allow the ADC capsule to set an incoming command as pending if the requested ADC channel is already in use, and run the pending command after the currently active command finishes.

### Testing Strategy

This pull request was tested on a Microbit, using a fake ADC driver to simulate the case in which two commands can overlap.

### TODO or Help Wanted

Feedback is greatly appreciated.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


capsules/src/adc.rs
```diff
@@ -653,16 +653,18 @@ impl<'a> AdcVirtualized<'a> {
...
-                .enter(processid, |app, _| {
-                    if self.current_app.is_none() {
-                        self.current_app.set(processid);
-                        let value = self.call_driver(command, channel);
-                        if value != Ok(()) {
-                            self.current_app.clear();
-                        }
-                        value
-                    } else {
+            if self.current_app.is_none() {
```
alexandruradovici, line 656: We renamed `appid` to `processid`, you might want to update your branch or just change the name of the variable here.

boards/components/src/adc.rs
```diff
@@ -27,7 +27,6 @@ macro_rules! adc_component_static {
 macro_rules! adc_syscall_component_helper {
     ($($P:expr),+ $(,)?) => {{
         use kernel::count_expressions;
-        use kernel::static_init;
```
alexandruradovici, line 30: Please keep this line, as the macro might be used in places where `static_init` is not defined.

capsules/src/adc.rs
```diff
@@ -653,16 +653,18 @@ impl<'a> AdcVirtualized<'a> {
...
+                self.current_app.set(appid);
```
alexandruradovici, line 657:
```suggestion
                self.current_process.set(appid);
```
Please rename `current_app` to `current_process` in all places.

alexandruradovici (changes requested):
Please add a comment explaining how the syscall driver works and how applications can you it, describing the exact rules that apply.

phil-levis (approved):
Ah, nice. It looks like the machinery was there (`pending_command`) but it wasn't being used.

Can you explain your refactoring of `enqueue_command`?

alistair23 (approved):

alexandruradovici (approved):

TeodoraMiu:
> Can you explain your refactoring of `enqueue_command`?

The enqueue_command function has been slightly changed: if the current process is none, there is no need to enter the grant. Therefore, I only enter the grant when there is an active process and the new command must be set as pending.

PR #3369, opened 2023-01-06
debug_writer_component: make debug buffer size configurable; double default debug buffer size

hudson-ayers:
### Pull Request Overview

This pull request makes the debug buffer size configurable via an optional macro argument, and doubles the default size of the buffer to 2kB. I modified the Hifive1 to use the configurable parameter so that it still uses a 1kB buffer, since it is highly RAM constrained.

This PR was inspired by https://github.com/tock/tock/pull/3327, which will ultimately require a much larger buffer (~8kB) for boards that use it.

### Testing Strategy

### TODO or Help Wanted

Are any other boards sufficiently RAM constrained as to require a smaller buffer than 2kB?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/debug_writer.rs
+++ b/boards/components/src/debug_writer.rs
@@ -31,66 +31,79 @@ use kernel::hil::uart;
 // least a 1 KiB boundary). This is not _semantically_ critical, but helps keep buffers on 1 KiB
 // boundaries in some cases. Of course, these definitions are only advisory, and individual boards
 // can choose to pass in their own buffers with different lengths.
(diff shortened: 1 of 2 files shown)


hudson-ayers:
An aside: This PR is a good example where the component trait is an obstacle. The requirement of using an associated type for the `StaticInput` that is passed to `finalize()` forced me to make the entire component generic over the buffer size, when otherwise only the finalize method would have needed to be generic (a much cleaner design with less boilerplate and no need for `PhantomData`).

phil-levis (approved):

PR #3372, opened 2023-01-09
Implementation of PWM functionality for RP2040.

Ioan-Cristian:
### Pull Request Overview

This pull request adds PWM functionality to RP2040.

### Testing Strategy

This pull request was tested using both unit test and integration tests on Raspberry Pi Pico W. The unit tests can be found inside chips/rp2040/src/pwm.rs and the integration tests in chips/rp2040/src/test/pwm.rs

### TODO or Help Wanted

Running tests on other RP2040-based MCUs.

### Documentation Updated

- [x] Added relevant documentation files.

### Formatting

- [x] Ran `make prepush`.


chips/rp2040/src/pwm.rs
```diff
@@ -0,0 +1,1254 @@
...
+            CH6 = 64,
+            CH7 = 128
+        ]
+    ]
+];
+
+const NUMBER_CHANNELS: usize = 8;
+
+#[repr(C)]
+struct Ch {
```
hudson-ayers, line 108: Why not `Channel`? (https://github.com/tock/tock/blob/master/doc/Style.md#using-descriptive-names)

chips/rp2040/src/pwm.rs
```diff
@@ -0,0 +1,1254 @@
...
+///
+/// **Note**:
+///
+/// + The same PWM output can be selected on two GPIO pins. The same signal will appear on each
+/// GPIO.
+/// + If a PWM B pin is used as an input, and is selected on multiple GPIO pins, then the PWM
+/// channel will see the logical OR of those two GPIO inputs
+impl From<RPGpio> for ChannelNumber {
+    fn from(gpio: RPGpio) -> Self {
+        match gpio as u8 >> 1 & 7 {
```
hudson-ayers, line 231:
```suggestion
        match gpio as u8 >> 1 & 0b111 {
```

nit, I just think this is a bit clearer

chips/rp2040/src/pwm.rs
```diff
@@ -0,0 +1,1254 @@
...
+    // once the increment is complete.
+    fn advance_count(&self, channel_number: ChannelNumber) {
+        self.registers.ch[channel_number as usize]
+            .csr
+            .modify(CSR::PH_ADV::SET);
+        while self.registers.ch[channel_number as usize]
+            .csr
+            .read(CSR::PH_ADV)
+            == 1
+        {}
```
hudson-ayers, line 479: Can you comment this with the maximum number of cycles you would expect this to block?

chips/rp2040/src/pwm.rs
```diff
@@ -0,0 +1,1254 @@
...
+    // is complete.
+    fn retard_count(&self, channel_number: ChannelNumber) {
+        self.registers.ch[channel_number as usize]
+            .csr
+            .modify(CSR::PH_RET::SET);
+        while self.registers.ch[channel_number as usize]
+            .csr
+            .read(CSR::PH_RET)
+            == 1
+        {}
```
hudson-ayers, line 494: Same here

chips/rp2040/src/pwm.rs
```diff
@@ -0,0 +1,1254 @@
...
+        let divider = threshold_freq_hz as f32 / selected_freq_hz as f32;
+        // If the desired frequency is too low, then it can't be achieved using the divider.
+        // In this case, notify the caller with an error.
+        if divider >= 256.0f32 {
+            return Err(());
+        }
+        // At this point, the divider is a valid value. Its integral and fractional part
+        // can be computed.
+        let int = divider as u8;
+        let frac = ((divider - int as f32) * 16.0) as u8;
```
hudson-ayers, line 642: These floating point calculations will be really slow/generate a lot of instructions since Tock does not currently support hardware floating point for arm cores. 

Fortunately, since you return an error if the divider is > 256, and only need 4-bit precision for the fractional component, I think the following should give the same results without requiring floating point:
```suggestion
        let divider = threshold_freq_hz / selected_freq_hz;
        // If the desired frequency is too low, then it can't be achieved using the divider.
        // In this case, notify the caller with an error.
        if divider >= 256.0f32 {
            return Err(());
        }
        // At this point, the divider is a valid value. Its integral and fractional part
        // can be computed.
        let int = divider as u8;
        // threshold_freq_hz *must* be less than u16::MAX since `get_maximum_duty_cycle()` 
        // returns u16::MAX + 1. Therefore neither of these multiplications by 16 can overflow.
        let frac = ((threshold_freq_hz * 16) / (selected_freq_hz) - divider * 16) as u8;
```

You may want to double check my math on this.

hudson-ayers (commented):
Overall looks really good, love all the unit tests. A few comments

chips/rp2040/src/pwm.rs
alexandruradovici: Additionally I suggest defining a `wait_for` function as in https://github.com/OxidosAutomotive/tock/blob/026361c7e830586c3eb25f2345b0449618c16078/chips/stm32f4xx/src/can.rs#L502-L517 and suing it like https://github.com/OxidosAutomotive/tock/blob/026361c7e830586c3eb25f2345b0449618c16078/chips/stm32f4xx/src/can.rs#L532-L534 and report an error.

chips/rp2040/src/pwm.rs
```diff
@@ -0,0 +1,1254 @@
...
+    // Direct access to the PWM counter register
+    ctr: ReadWrite<u32, CTR::Register>,
+    // Counter compare values register
+    cc: ReadWrite<u32, CC::Register>,
+    // Counter wrap value register
+    top: ReadWrite<u32, TOP::Register>,
+}
+
+#[repr(C)]
+struct PwmRegisters {
```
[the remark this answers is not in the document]
alexandruradovici: You can start form this exmaple https://github.com/tock/tock/blob/fd669565a407dd03018a8acdbfe6de1a885ee912/chips/rp2040/src/adc.rs#L8

chips/rp2040/src/pwm.rs
Ioan-Cristian: The datasheet doesn't say anything about the maximum number of cycles the pooling should be done. Should I assume that the pooling should be over after 1 PWM period?

alexandruradovici: Just wait a few cycles (maybe max 100) before reporting an error.

chips/rp2040/src/pwm.rs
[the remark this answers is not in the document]
Ioan-Cristian: Done.

chips/rp2040/src/pwm.rs
Ioan-Cristian: Changed the name.

chips/rp2040/src/pwm.rs
Ioan-Cristian: Done.

chips/rp2040/src/pwm.rs
Ioan-Cristian: Implemented `wait_for()` method.

chips/rp2040/src/pwm.rs
Ioan-Cristian: Done.

chips/rp2040/src/pwm.rs
Ioan-Cristian: Changed the code. Is it OK now?

chips/rp2040/src/pwm.rs
```diff
@@ -0,0 +1,1254 @@
...
+    // selected_freq_hz ==> user's desired frequency
+    //
+    // Return value: Ok(top, int, frac) in case of no error, otherwise Err(())
+    fn compute_top_int_frac(&self, selected_freq_hz: usize) -> Result<(u16, u8, u8), ()> {
+        // If the selected frequency is high enough, then there is no need for a divider
+        // Note that unwrap can never fail.
+        let max_freq_hz = hil::pwm::Pwm::get_maximum_frequency_hz(self);
+        let threshold_freq_hz = max_freq_hz / hil::pwm::Pwm::get_maximum_duty_cycle(self);
+        if selected_freq_hz > threshold_freq_hz {
+            return Ok(((max_freq_hz / selected_freq_hz - 1) as u16, 1, 0));
```
hudson-ayers, line 629:
```suggestion
            return Ok(((max_freq_hz / (selected_freq_hz - 1)) as u16, 1, 0));
```
Is this what you meant here? if not, this subtraction can underflow for a number of different input frequencies depending on the value of `max_freq_hz`. I am not sure why you are subtracting 1 since I am not sure what the `top` value means. If you meant to divide, then subtract, change this to use `saturating_sub()` instead

chips/rp2040/src/pwm.rs
```diff
@@ -0,0 +1,1254 @@
...
+        }
```
hudson-ayers, line 630:
```suggestion
        } else if selected_freq_hz == 0 {
            return Err(());
        }
```

Otherwise this function panics if the caller passes 0 as the input frequency

chips/rp2040/src/pwm.rs
hudson-ayers: It is ok now once you make my other two suggested changes. I verified the new version is functionally equivalent to the old version using this playground https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3eaa2b45c63ca9a153ef00e2d49c8c6c (I had to run it locally to get past the playground timeout, though)

chips/rp2040/src/pwm.rs
Ioan-Cristian: It's a division followed by a substraction. **top** is the maximum value of the counter. Once the counter reaches it, it eithers wraps to 0 or starts counting down (if running in double-edge mode).

Normally, no underflow should occur since one should choose the desired frequency based on the maximum value, i.e. `let freq = max_freq / 3`. I didn't add a constraint on this since I thought it is in the responsability of the caller to ensure the value is less than or equal to maximum value. But now I added a check to see if either the frequency is greater than **max_freq_hz** or if its value is 0.

chips/rp2040/src/pwm.rs
Ioan-Cristian: See my comment above.

chips/rp2040/src/pwm.rs
hudson-ayers: Makes sense, looks good to me

hudson-ayers (approved):

chips/rp2040/src/pwm.rs
```diff
@@ -0,0 +1,1282 @@
...
+        pwm.init();
+        pwm
+    }
+
+    // Enable or disable the given PWM channel
+    //
+    // enable == false ==> disable channel
+    // enable == true ==> enable channel
+    fn set_enabled(&self, channel_number: ChannelNumber, enable: bool) {
+        self.registers.ch[channel_number as usize]
```
twilfredo, line 343:
```suggestion
        if channel_number >= NUMBER_CHANNELS {
                // Do something... perhaps return a `Result`
                return;
        }
        self.registers.ch[channel_number as usize]
```
something like this could help to avoid out of bound errors in places where do we ` self.registers.ch[channel_number as usize]` and might reduce the panic bloat generated.

twilfredo (approved):

chips/rp2040/src/pwm.rs
Ioan-Cristian: I don't think it is necessary to add this check, since `ChannelNumber` enum has exactly `NUMBER_CHANNELS` variants.

alexandruradovici: I agree with @Ioan-Cristian

alexandruradovici (approved):

chips/rp2040/src/pwm.rs
hudson-ayers: Agreed, marking resolved

twilfredo: Ah whoops I missed that!

alexandruradovici:
@Ioan-Cristian can you please fix the conflicts.

Ioan-Cristian:
> @Ioan-Cristian can you please fix the conflicts.

I think I fixed them. Sorry for the delay!

alexandruradovici:
This seems to be merged, but not closed. @hudson-ayers this might be a bors issue.

hudson-ayers:
Yeah, confirmed this did in fact get merged into master, but failed to close as merged..just going to close this then

PR #3381, opened 2023-01-23
Process console command history

CosminGGeorgescu:
### Pull Request Overview

This pull request adds an easy-to-use Bash-like command history for the process console.

### Added features

   - User can enable/disable the command history from the board's `main.rs`.
   - If enabled, user can scroll `up` or `down` by pressing `up` and `down` arrow keys respectively.
   - Pressing on accident `up` or `down` arrow saves the unfinished command and can be fetched back by going to the bottom of command history. 
   - Modifying a command from history is also saved (if accidentally pressed `up` or `down` arrow).
   - Empty commands (whitespace-filled) don't get registered in the command history.
   - Invalid commands also are saved into the history structure.

### Code size for IMIX binary build with enabled/disabled command history:

| Build | .text | .data | .bss | .dec |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| Tock master  | 174468  |  0 |  29592  | 204060  |
| Enabled with default size  | 175492  |  0  |  28944  |  204436  |
| Disabled  |  174292  | 0  |  28576  |  202868  |

If disabled, the code size decreases.

### Testing Strategy

The implementation of the command history was tested on the `nrf52840dk` and `esp32-c3-devkitM-1` boards.

### Files changed
`capsules/process_console.rs`
 ~ implementation for command history

`boards/components/process_console.rs`
 ~ added a new branch to process_console_component_static macro for easily enable/disable the command history

### Documentation Updated

Updated the documentation in `doc/Process_Console.md` accordingly : 
- Code support on how to enable/disable the command history
- How to use command history
- Support for selecting a custom size for the command history

### TODO or Help Wanted

Feedback is much appreciated.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/process_console.rs
+++ b/capsules/src/process_console.rs
@@ -162,9 +162,9 @@ impl Command {
         (&mut self.buf).copy_from_slice(buf);
     }
 
-    fn insert_byte(&mut self, byte: &u8) {
+    fn insert_byte(&mut self, byte: u8) {
         if self.len < COMMAND_BUF_LEN {
-            self.buf[self.len] = *byte;
+            self.buf[self.len] = byte;
             self.len = self.len + 1;
         }
     }
@@ -1050,10 +1050,13 @@ impl<'a, const COMMAND_HISTORY_LEN: usize, A: Alarm<'a>, C: ProcessManagementCap
                                 self.command_history.map(|cmd_arr| {
                                     if self.scrolled_in_history.get() {
                                         cmd_arr[0].clear();
+                                        for i in 0..(self.command_index.get() - 1) {
+                                            cmd_arr[0].insert_byte(command[i]);
+                                        }
                                         self.scrolled_in_history.set(false);
                                     }
 
-                                    cmd_arr[0].insert_byte(&read_buf[0]);
+                                    cmd_arr[0].insert_byte(read_buf[0]);
                                 });
                             }
                         }
--- a/capsules/src/process_console.rs
+++ b/capsules/src/process_console.rs
@@ -34,12 +34,14 @@ pub const READ_BUF_LEN: usize = 4;
 /// characters, limiting arguments to 25 bytes or so seems fine for now.
 pub const COMMAND_BUF_LEN: usize = 32;
 
-pub const COMMAND_HISTORY_LEN: usize = 10;
+pub const DEFAULT_COMMAND_HISTORY_LEN: usize = 10;
 
 /// List of valid commands for printing help. Consolidated as these are
 /// displayed in a few different cases.
 const VALID_COMMANDS_STR: &[u8] =
     b"help status list stop start fault boot terminate process kernel panic\r\n";
+
+const ESC: u8 = '\x1B' as u8;
 
 /// States used for state machine to allow printing large strings asynchronously
 /// across multiple calls. This reduces the size of the buffer needed to print
@@ -70,34 +72,35 @@ impl Default for WriterState {
 }
 
 #[derive(Copy, Clone)]
-pub struct Command<const LEN: usize> {
-    buf: [u8; LEN],
+pub struct Command {
+    buf: [u8; COMMAND_BUF_LEN],
     len: usize,
 }
-impl<const LEN: usize> Command<LEN> {
-    pub fn new() -> Command<LEN> {
+impl Command {
+    pub fn new() -> Command {
         Command {
-            buf: [0; LEN],
+            buf: [0; COMMAND_BUF_LEN],
             len: 0,
         }
     }
 
+    /// Fill the buffer with the provided data.
+    /// If the provided data length is smaller than the buffer length,
+    /// the left over bytes are not modified due to '\0' termination.
     pub fn fill(&mut self, buf: &[u8], terminator_idx: usize) -> Result<(), ErrorCode> {
-        if terminator_idx >= LEN {
-            self.len = LEN;
+        if terminator_idx >= COMMAND_BUF_LEN {
+            self.len = COMMAND_BUF_LEN;
         } else {
             self.len = terminator_idx;
         }
 
-        for i in 0..self.len {
-            self.buf[i] = buf[i];
-        }
+        self.buf[..self.len].copy_from_slice(&buf[..self.len]);
 
         Ok(())
     }
 
-    pub fn is_buffer_empty(&mut self) -> Option<bool> {
-        Some(self.len == 0 && self.buf[0] == 0)
+    pub fn is_buffer_empty(&mut self) -> bool {
+        self.len == 0 && self.buf[0] == 0
     }
 
     pub fn is_byte_empty(&mut self, byte_idx: usize) -> Option<bool> {
@@ -116,18 +119,12 @@ impl<const LEN: usize> Command<LEN> {
         Some(self.buf[byte_idx])
     }
 
-    pub fn same_bytes(&mut self, buf: &[u8], terminator_idx: usize) -> Option<bool> {
+    pub fn same_bytes(&mut self, buf: &[u8], terminator_idx: usize) -> bool {
         if terminator_idx != self.len {
-            return Some(false);
+            return false;
         }
 
-        for i in 0..self.len {
-            if self.buf[i] != buf[i] {
-                return Some(false);
-            }
-        }
-
-        Some(true)
+        self.buf.iter().zip(buf.iter()).all(|(a, b)| a == b)
     }
 }
 
@@ -163,7 +160,7 @@ pub struct ProcessConsole<'a, A: Alarm<'a>, C: ProcessManagementCapability> {
     command_index: Cell<usize>,
 
     /// Keep a history of inserted commands
-    command_history: TakeCell<'static, [Command<{ COMMAND_BUF_LEN }>; COMMAND_HISTORY_LEN]>,
+    command_history: TakeCell<'static, [Command; DEFAULT_COMMAND_HISTORY_LEN]>,
 
     control_seq_in_progress: Cell<bool>,
 
@@ -237,7 +234,7 @@ impl<'a, A: Alarm<'a>, C: ProcessManagementCapability> ProcessConsole<'a, A, C> 
         rx_buffer: &'static mut [u8],
         queue_buffer: &'static mut [u8],
         cmd_buffer: &'static mut [u8],
-        cmd_history_buffer: &'static mut [Command<{ COMMAND_BUF_LEN }>; COMMAND_HISTORY_LEN],
+        cmd_history_buffer: &'static mut [Command; DEFAULT_COMMAND_HISTORY_LEN],
         kernel: &'static Kernel,
         kernel_addresses: KernelAddresses,
         capability: C,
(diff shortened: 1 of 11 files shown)


alexandruradovici (dismissed):

alexandruradovici:
This feature is very helpful in education, most of our students tried immediately to scroll up and down. I think this can be added, as it does not increase code size. 

I curious though why the code size actually decreases with the feature turned off.

doc/Process_Console.md
```diff
@@ -600,3 +603,49 @@ tock$
...
+          process_printer,
+      )
+      .finalize(components::process_console_component_static!(
+          nrf52833::rtc::Rtc,
+          COMMAND_HISTORY_LEN // or nothing for the default behaviour
+      ));
+
+  /// ...
+ ```
+> Note: In order to disable any functionality for the command history set the `COMMAND_HISTORY_LEN` as `0`, also the history will be disabled for a size of `1`, because the first position from the command history is reserved for accidents by pressing `up` or `down` arrow key.
```
alexandruradovici, line 651:
```suggestion
> Note: In order to disable any functionality for the command history set the `COMMAND_HISTORY_LEN` to `0` or `1` (the history will be disabled for a size of `1`, because the first position from the command history is reserved for accidents by pressing `up` or `down` arrow key.
```

alexandruradovici (dismissed):

alexandruradovici:
Kind reminder to review this, we need this the for Rust Nation presentation. :)

lschuermann (approved):
I'm by no means an expert on the process console code, but from a coarse review this looks good to me. I'm not too worried about the code size or memory usage impact of this (as long as it doesn't grow astronomically), given it's intended solely for debugging and experimentation.

hudson-ayers (approved):
The decreased code size when this is disabled is just an indication that LLVM was making some suboptimal inlining decisions regarding the process console before this was added, but that may not apply universally across boards and could be fixed with explicit inline directives anyway.

Since this can be disabled from main I am indifferent to the addition -- I personally would probably advocate for it being off-by-default (i.e. unless you specify a length when creating the component no commands will be saved) but will not block this on that.

CosminGGeorgescu:
+1 to @hudson-ayers's opinion for it being off by default

alexandruradovici:
> +1 to @hudson-ayers's opinion for it being off by default

I think that for educational purposes and for getting started with Tock, the history should be on by default. Most developers do expect this feature from the console, all of our students being frustrated that it was not available.

capsules/src/process_console.rs
```diff
@@ -120,6 +144,64 @@ pub struct ProcessConsole<'a, A: Alarm<'a>, C: ProcessManagementCapability> {
...
+            .iter()
+            .position(|a| *a == EOL)
+            .unwrap_or(COMMAND_BUF_LEN);
+
+        (&mut self.buf).copy_from_slice(buf);
+    }
+
+    fn insert_byte(&mut self, byte: u8) {
+        if self.len < COMMAND_BUF_LEN {
+            self.buf[self.len] = byte;
```
alistair23, line 168: Not blocking, but in these type of cases I feel it's better to use `get()/get_mut()`.

So for example (untested):

```rust
if Ok(b) = self.get_mut(self.len) {
    *b = byte;
}
```

That avoids the panic. I would assume that rustc is smart enough here to understand that this can't panic and optimise it out, but AFAIK there is no guarantee that it will

capsules/src/process_console.rs
```diff
@@ -120,6 +144,64 @@ pub struct ProcessConsole<'a, A: Alarm<'a>, C: ProcessManagementCapability> {
...
+    }
+}
+
+impl PartialEq<[u8; COMMAND_BUF_LEN]> for Command {
+    fn eq(&self, other_buf: &[u8; COMMAND_BUF_LEN]) -> bool {
+        self.buf
+            .iter()
+            .zip(other_buf.iter())
+            .take_while(|(a, b)| **a != EOL || **b != EOL)
+            .all(|(a, b)| *a == *b)
```
alistair23, line 201: Won't this not work if one is longer then the other?

For example if the buffers are: "abc" and "abcde" won't zip just iterate "abc" and they all match?

alistair23 (changes requested):

capsules/src/process_console.rs
Matrix22: Cheked once again on the board and seems to work as expected, also have tested it on your example and seems to work.

The `take_while` will stop until both bytes from buffer are set to `EOL`.
In other words will stop when will encounter a `(EOL, EOL)` pair.
So for your example we have the following buffers:
```text
b1 = ['a', 'b', 'c', EOL, EOL, EOL, ....];
b2 = ['a', 'b', 'c', 'd', 'e', EOL, EOL, EOL, ...];
```

when comparing we will have the following pairs:
```text
zip = [
('a', 'a'),
('b', 'b'),
('c', 'c'),
(EOL, 'd'),
(EOL, 'e')
]
```
Which will evaluate as false.

Instead if we put on the take_while predicate something like this `|(a, b)| **a != EOL && **b != EOL` it would generate the behaviour that you were talking about.

capsules/src/process_console.rs
Matrix22: I do not know if rustc will optimize this, becasue the `insert` and `delete` methods are called when user is typing a character or is pressing `backspace` in order to delete a character, so if the command buffer is empty (delete case) and user is pressing a backspace this would cause a panic and the same way for the insert case.

However I changed the code for the `insert_byte` and `delete_last_byte` to match the suggestion.

capsules/src/process_console.rs
alistair23: Ah! Your right. While I was looking at this I forgot they are fixed length buffers :facepalm:

alistair23 (approved):

bradjc (approved):

phil-levis (approved):
As the original process console author, I agree with this improvement conceptually. I haven't done a code review.

capsules/src/process_console.rs
```diff
@@ -860,8 +978,131 @@ impl<'a, A: Alarm<'a>, C: ProcessManagementCapability> uart::ReceiveClient
...
+                                    });
+                                }
+                            }
+                        } else if (COMMAND_HISTORY_LEN > 1)
+                            && (read_buf[0] == ESC || self.control_seq_in_progress.get())
+                        {
+                            // Catch the Up and Down arrow keys
+                            if read_buf[0] == ESC {
+                                // Signal that a control sequence has started and capture it
+                                self.control_seq_in_progress.set(true);
```
jettr, line 998: I meant to collaborate with Alex on this since we have a downstream version of process console that handles command history.

I want to share the small state machines that we use down stream that you can either choose to use or not (your call). I just wanted it to be an option. Consider this last minute collaboration.


```rust

/// Start of terminal escape sequence
const CHAR_ESC_START: u8 = b'\x1B'; 

/// The FSM for processing terminal escape sequences
#[derive(Copy, Clone)]
enum EscState {
    /// Most recent character is not part of an escapee sequence, so process it elsewhere
    Bypass,
    /// The escape sequence completed with a recognized escape key
    Complete(EscKey),
    Started,
    Bracket,
    Bracket1,
    Bracket3,
    Bracket4,
    Unrecognized,
    UnrecognizedDone,
}

impl EscState {
    /// Process the data for this state and returns the next state in the FSM
    fn next_state(self, data: u8) -> Self {
        use self::{EscKey::*, EscState::*};
        match (self, data) {
            (Bypass, CHAR_ESC_START)
            | (UnrecognizedDone, CHAR_ESC_START)
            | (Complete(_), CHAR_ESC_START) => Started,
            (Bypass, _) | (UnrecognizedDone, _) | (Complete(_), _) => Bypass,
            (Started, b'[') => Bracket,
            (Bracket, b'A') => Complete(Up),
            (Bracket, b'B') => Complete(Down),
            (Bracket, b'C') => Complete(Right),
            (Bracket, b'D') => Complete(Left),
            (Bracket, b'1') => Bracket1,
            (Bracket1, b'~') => Complete(Home),
            (Bracket, b'3') => Bracket3,
            (Bracket3, b'~') => Complete(Delete),
            (Bracket, b'4') => Bracket4,
            (Bracket4, b'~') => Complete(End),
            _ => {
                if EscState::terminator_esc_char(data) {
                    UnrecognizedDone
                } else {
                    Unrecognized
                }
            }
        }
    }

    /// Returns true if the processing next state logically consumed the data and it should not
    /// be processed elsewhere
    pub fn data_consumed(&self) -> bool {
        !matches!(self, EscState::Bypass)
    }

    fn terminator_esc_char(data: u8) -> bool {
        data.is_ascii_alphabetic() || data == b'~'
    }
}

```

And when it is used, we do something like the follow. Note that `cursor` is the position of the cursor and `len` is the total pending command length.

```rust
        let esc_state = self.esc_state.get().next_state(ch);
        self.esc_state.set(esc_state);

        if let EscState::Complete(key) = esc_state {
            match key {
                EscKey::Right if cursor < len => {
                    Direction::Right.move_cursor(1);
                    cursor += 1;
                }
                EscKey::Left if cursor > 0 => {
                    Direction::Left.move_cursor(1);
                    cursor -= 1;
                }
                EscKey::End if cursor < len => {
                    Direction::Right.move_cursor(len - cursor);
                    cursor = len;
                }
                EscKey::Home if cursor > 0 => {
                    Direction::Left.move_cursor(cursor);
                    cursor = 0;
                }
                EscKey::Delete if cursor < len => {
                    // Remove the character under the cursor
                    try_copy_within(&mut cmd_buf, cursor + 1..len, cursor)?;
                    len -= 1;
                    // Print the updated buffer with a clearing space at end
                    for c in cmd_buf.iter().skip(cursor).take(len - cursor) {
                        console_output_no_eol!("{}", *c as char);
                    }
                    console_output_no_eol!(" ");
                    Direction::Left.move_cursor(len - cursor + 1);
                }
                EscKey::Up | EscKey::Down => {
                    let res: Result<(), ErrorCode> = self.mru.map_or(Ok(()), |mru| {
                        // Clear the current line on console
                        Direction::Left.move_cursor(cursor);
                        for _ in 0..len {
                            console_output_no_eol!(" ");
                        }
                        Direction::Left.move_cursor(len);

                        // Copy in the next data from mru circular buffer slices
                        let (b1, b2) = if matches!(key, EscKey::Up) {
                            mru.get_next_older()
                        } else {
                            mru.get_next_newer()
                        };
                        let total = b1.len() + b2.len();
                        cmd_buf
                            .get_mut(..b1.len())
                            .ok_or(ErrorCode::INVAL)?
                            .copy_from_slice(b1);
                        cmd_buf
                            .get_mut(b1.len()..total)
                            .ok_or(ErrorCode::INVAL)?
                            .copy_from_slice(b2);
                        cursor = total;
                        len = total;

                        // Print all new characters
                        for c in cmd_buf.iter().take(len) {
                            console_output_no_eol!("{}", *c as char);
                        }
                        Ok(())
                    });
                    res?
                }
                _ => {}
            };
        }
        // If the escape sequence consumed the character, we are done processing this character
        if esc_state.data_consumed() {
            return Ok(());
        }
```

Also here is the Direction helper
```rust
/// Direction to move a cursor
enum Direction {
    Left,
    Right,
}

impl Direction {
    fn move_cursor(&self, count: usize) {
        if count != 0 {
            console_output_no_eol!(
                "\x1b[{}{}",
                count,
                match self {
                    Direction::Left => "D",
                    Direction::Right => "C",
                }
            );
        }
    }
}
```

Lastly, we have a Most Recently Used (`mru` in code above) circular buffer cache for console history that compactly stores previous commands with an upper bound limit on the totally number of bytes and stored commands, which ever comes first. I can also share this too with its unit tests.

Also don't feel like you have to make these changes before landing this PR. If you want to incorporate this feedback in a follow up PR, I would be happy to help out. I just wanted to start talking about this.

alexandruradovici: I agree, @Matrix22 @CosminGGeorgescu you can prepare a new pull request as soon as this gets merged.

hudson-ayers:
bors r+

(On the Friday call, there was agreement that we could merge this as-is then seperately incorporate Jett's suggestion in a new PR, so as to ensure this makes it in before RustNation)

PR #3384, opened 2023-01-24
fixed i2c buffer len

teonaseverin:
### Pull Request Overview

This pull request changes the length of the buffer that can be read or written on the I2C bus to a `usize` variable. Currently the length is represented by a `u8`, but there are use cases in which an application needs to send more than 256 characters on the I2C bus at once. Is there any reason behind this limitation that I missed? 

### Testing Strategy

This pull request was tested by connecting an I2C peripheral to the `nucleo_429zi` board.

### TODO or Help Wanted

Any review or comment is welcome. 

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/src/bmp280.rs
+++ b/capsules/src/bmp280.rs
@@ -150,7 +150,7 @@ impl<'a> I2cWrapper<'a> {
         buffer[0] = addr as u8;
         buffer[1..][..COUNT].copy_from_slice(&data);
         self.i2c.enable();
-        self.i2c.write(buffer, COUNT as u8 + 1)
+        self.i2c.write(buffer, COUNT + 1)
     }
 
     /// Requests a read into buffer.
@@ -159,7 +159,7 @@ impl<'a> I2cWrapper<'a> {
         &self,
         buffer: &'static mut [u8],
         addr: Register,
-        count: u8,
+        count: usize,
     ) -> Result<(), (i2c::Error, &'static mut [u8])> {
         buffer[0] = addr as u8;
         self.i2c.enable();
@@ -304,7 +304,7 @@ impl<'a, A: Alarm<'a>> Bmp280<'a, A> {
 enum I2cOperation {
     Read {
         addr: Register,
-        count: u8,
+        count: usize,
         fail_state: State,
     },
     Write {
--- a/capsules/src/bus.rs
+++ b/capsules/src/bus.rs
@@ -315,7 +315,7 @@ impl<'a, I: I2CDevice> Bus<'a> for I2CMasterBus<'a, I> {
             debug!("write len {}", len);
             self.len.set(len);
             self.status.set(BusStatus::Write);
-            match self.i2c.write(buffer, (len * bytes) as u8) {
+            match self.i2c.write(buffer, len * bytes) {
                 Ok(()) => Ok(()),
                 Err((error, buffer)) => Err((error.into(), buffer)),
             }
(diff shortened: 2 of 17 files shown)


brghena:
Why `usize` and not `u16` then? It seems better to use fixed, clear sizes wherever possible.

Also, do you really have an i2c peripheral capable of sending more than 255 bytes in one message? I don't know that I've ever seen a peripheral capable of that, which is probably why the limit is set to a `u8` now. If you do, then I'm for changing it.

teonaseverin:
I chose `usize` because it is the data type used also by the SPI HIL. Although, if there is any problem with `usize`, `u16` would also help. Also, as far as I know, there is no limit on the I2C bus.

The peripheral that I used is a SSD1306 screen that communicates via I2C and can receive indeed more than 255 bytes  at once.

hudson-ayers:
I think the "I2C peripheral" that Branden is referring to is the on-chip I2C device on the stm32f429zi chip on your nucleo board (peripheral is a confusing word here given that "I2C peripheral" has replaced "I2C slave" as the terminology for the non-controlling device in an I2C interaction).

That said, I looked at the datasheet and the I2C peripheral code for your chip and I see now that the Tock driver just sends one byte at a time over I2C, receiving an interrupt after every byte sent to continue through its state machine. This design obviously does not impose any length limit on the buffer passed down to the peripheral (but it much less efficient than if the driver was written to use DMA)

brghena:
Yes, sorry about the terminology!! I should have been more clear that I did indeed mean the i2c hardware on the stm32.

The exact thing Hudson looked at does resolve my second question. You can send more than 255 bytes at once because you're NOT using DMA. It's still pretty rare for an external I2C device to want more than 255 bytes, but a screen is a perfect example of such a device.

So the only question left is what to use. If SPI uses `usize` I guess it does make more sense to copy that. I generally just dislike a non-fixed-size types like `usize` that could change between architectures when demonstrating fixed size information like a number of bytes to transfer.

alexandruradovici:
> The exact thing Hudson looked at does resolve my second question. You can send more than 255 bytes at once because you're NOT using DMA. It's still pretty rare for an external I2C device to want more than 255 bytes, but a screen is a perfect example of such a device.
> 

Now it makes sense why this was fixed to 255 bytes. Is there any restriction for the DMA size in general, or was this an implementation detail for the nRF?

> So the only question left is what to use. If SPI uses `usize` I guess it does make more sense to copy that. I generally just dislike a non-fixed-size types like `usize` that could change between architectures when demonstrating fixed size information like a number of bytes to transfer.

The idea behind using `usize` is that the HIL is able to accommodate any I2C master, regardless of the hardware that it has beneath. I would still go for `usize` to match the other buses (SPI, UART, etc).

brghena:
> Is there any restriction for the DMA size in general, or was this an implementation detail for the nRF?

It's not a global thing. It's not just an nRF thing though either. I've seen similar limits on many microcontrollers. There has to be a "length" field in a register somewhere, so you have to pick some limit. Sometimes this is in the DMA. Sometimes it's in the I2C peripheral itself.

Actually, I just checked the nRF52840 documentation, and the length field is 16-bit there. So, that's even more evidence for why this change is valuable. (MAXCNT register)

Meanwhile the SAM4L _is_ limited to an 8-bit length before a stop bit is generated. (NBYTES field of the CMDR register). Although there might be some way to get around that, like disabling the stop bit or something.

I didn't see any particular limit in a brief scan of the STM32. Probably whatever DMA's limit is then.


> The idea behind using usize is that the HIL is able to accommodate any I2C master, regardless of the hardware that it has beneath.

I'm not sure I follow. Wouldn't a `u32` or a `u16` still be a better choice? Then you're sure of exactly how many bits of length you have. 


> I would still go for usize to match the other buses (SPI, UART, etc).

I agree here though. It seems better to match than to be different.

hudson-ayers:
We have talked for sometime about changing every use of `usize` which isn't being used for pointers to `u32`, but I don't think that this PR needs to be the start of that change (which will only be useful in practice if it happens to every single driver in Tock), I think it looks fine as-is.

hudson-ayers (approved):

brghena (approved):

brghena:
From discussion on the working group call, `usize` is important for indexing into arrays in Rust. So there was a reason the APIs were designed to use `usize` isntead of `u32` in the first place. So we should definitely keep `usize` in this case as well.

twilfredo (approved):

Issue #3346, opened 2022-12-07
RFC: Reorganizing capsules into `core/`, `contrib/`, `external/`


hudson-ayers:
On last week's core call we discussed making capsules not a single crate, but a few crates. There are several benefits to this, the motivating one from the call was so that we could have a dedicated crate for capsules with external dependencies. I believe the conclusion we came to on the call was to minimize the amount that we split up `capsules` to start by just making two crates -- `core/`, for core capsules we expect almost all boards to use, and `contrib/`, for everything else. Finally, there would be an `external/` directory which would hold individual crates for each capsule with external dependencies. 

We did not come to a conclusion on exactly how we would split up `core/` and `contrib/`. I would like some level of agreement before I implement this, since it is going to require renaming a ton of paths, so lots of changing back and forth will be a pain. I propose the following breakdown:

## Core:
adc.rs
alarm.rs
button.rs
console.rs
gpio.rs
led.rs
i2c_master.rs
i2c_master_slave_driver.rs
low_level_debug/
process_console.rs
rng.rs
spi_controller.rs
spi_peripheral.rs
virtual_adc.rs
virtual_aes_ccm.rs
virtual_alarm.rs
virtual_digest.rs
virtual_flash.rs
virtual_hmac.rs
virtual_i2c.rs
virtual_pwm.rs
virtual_rng.rs
virtual_sha.rs
virtual_spi.rs
virtual_timer.rs
virtual_uart.rs

## Contrib:
adc_microphone.rs
air_quality.rs
ambient_light.rs
analog_comparator.rs
analog_sensor.rs
apds9960.rs
app_flash_driver.rs
ble_advertising_driver.rs
bme280.rs
bmp280.rs
bus.rs
buzzer_driver.rs
buzzer_pwm.rs
ccs811.rs
crc.rs
ctap.rs
dac.rs
debug_process_restart.rs
fm25cl.rs
ft6x06.rs
fxos8700cq.rs
gpio_async.rs
hd44780.rs
hmac.rs
hts221.rs
humidity.rs
ieee802154/
isl29035.rs
kv_driver.rs
kv_store.rs
l3gd20.rs
led_matrix.rs
log.rs
lpm013m126.rs
lps25hb.rs
lsm303agr.rs
lsm303dlhc.rs
lsm303xx.rs
lsm6dsoxtr.rs
ltc294x.rs
max17205.rs
mcp230xx.rs
mlx90614.rs
mx25r6435f.rs
net/
ninedof.rs
[...]

PR #3396, opened 2023-02-10
capsules: split into `core/` and `extra/` (implementing #3346)

lschuermann:
### Pull Request Overview

This pull request splits capsules into `core/` and `extra/` as proposed and discussed in #3346 for #3312.

### Testing Strategy

This pull request was tested by `cargo check`.

### TODO or Help Wanted

This PR is still in development. Specifically, I'd like to check whether it actually moves capsules into the correct crates, and update documentation pointing to capsules, as well as updating the capsules (now `core/`) README.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


hudson-ayers:
Took a quick look through this and I think it is pretty much exactly what I had in mind! 

> updating the capsules (now core/) README.

I think we probably want a capsules/README as well as a capsules/core/README and a capsules/extra/README. The top level one should just explain the basic division of capsules and where new capsules should be placed.

lschuermann:
I have split up the README between `core/` and `extra/` accordingly, and also introduced a new top-level capsules README file.

One additional thing to note: as part of this refactoring, I needed to promote the `stream` module, implementing a macro-infrastructure for encoding & decoding byte-streams, to a module in `core`. This is because `virtual_aes_ccm` uses this module, and all virtualizers are to be kept in `core`. I don't think that is going to cause issues, but am open to other approaches.

Also, this PR currently does not move all virtualizers into their own module. I'm personally in favor of doing that, as it helps to separate the different reasoning behind a given module being in core: either it's a well-vetted module crucial to a given boards' operation, or it is a virtualizer and thus only bridging kernel-internal interfaces and not interacting with external hardware or userspace directly. I'd appreciate feedback on this.

alevy (dismissed):

bradjc (dismissed):
I like the change, but I think having the default in boards as "capsules_core" and "capsules_extra" makes sense (particularly if we have more crates in the future).

lschuermann:
I will rename `{core,extra}_capsules` to `capsules_{core,extra}`, and split the virtualizers into their own module, as discussed on today's call. Thanks for the feedback!

lschuermann:
Okay, I think the discussed changes should be addressed now.

hudson-ayers (approved):

brghena (dismissed):

brghena:
Okay, this is a dumb question, but what makes it so that the module is named "capsules-core", but when you import it inside rust code it's "capsules_core"? Is that some Rust rule that I just didn't know?

lschuermann:
> but what makes it so that the module is named "capsules-core", but when you import it inside rust code it's "capsules_core"?

According to the [Rust reference on external crates](https://doc.rust-lang.org/reference/items/extern-crates.html), apparently this is a hard-coded rule: "When naming Rust crates, hyphens are disallowed. However, Cargo packages may make use of them. In such case, when Cargo.toml doesn't specify a crate name, Cargo will transparently replace - with _".

brghena:
Thanks for the link. Makes me feel a little better that I didn't understand what was happening.

There's no reason to "fix" anything. I was just trying to figure out how it worked and couldn't find it in the code anywhere.

hudson-ayers (approved):

bradjc (approved):

Issue #3375, opened 2023-01-11
All drivers should return `Success` rather than `SuccessU32` for command 0


hudson-ayers:
https://github.com/tock/tock/blob/master/doc/reference/trd104-syscalls.md#431-command-identifier-0 specifies that command 0 should always return `Success`. However, at least the following 4 drivers do not follow this direction:

- ADC https://github.com/tock/tock/blob/500caa5e882b7efe40158849fb489205f07c6a2e/capsules/src/adc.rs#L1149
- GPIO https://github.com/tock/tock/blob/500caa5e882b7efe40158849fb489205f07c6a2e/capsules/src/gpio.rs#L198
- LED https://github.com/tock/tock/blob/500caa5e882b7efe40158849fb489205f07c6a2e/capsules/src/led.rs#L95
- PWM https://github.com/tock/tock/pull/3367

I recall there being a desire for userspace to be able to build helper functions around command 0 with the assumption of a fixed Success variant for that command, so we should update these 4 drivers by adding a new command number to check the number of resources for that particular driver. This will require matching updates to `libtock-rs` and `libtock-c` for each

jrvanwhy:
ADC and LED are stable capsules, and changing this call is a breaking change, so wouldn't this require a version bump to 3.0?

Or is this okay because "TRD 104 says it should be this way"?

ppannuto:
I would like to think the "stable" does indeed mean "stable" for Tock 2.x, [we currently say](https://github.com/tock/tock/tree/master/doc/syscalls):
> Each driver type that has been allocated a permanent driver number is listed in the tables below. The "2.0" column indicates whether the driver has been stabilized or not (a "✓" indicates stability) in the Tock 2.0 release.

I think a path forward might be to mark driver `0x00002` as deprecated, and pending removal in the Tock 3.x release, and have the revised driver from #3397 be a new `0x0000a` driver, available from Tock 2.2+, and as the new stable LED interface for Tock 3.x?

Downstream platforms concerned about code size could, of course, omit the `0x00002` driver in their internal implementations to avoid duplicate copies of the LED driver.

---

LED happens to be stateless, so in this case we could likely have correct operation simply keeping two copies of the LED capsules around (caveat: two Rust references to the same MMIO region). The better thing would probably be to write a "legacy LED" capsule or some such that lives underneath `0x00002` and forwards / translates commands to the "real LED" capsule, wired directly to `0x0000a`?

hudson-ayers:
[...]

PR #3397, opened 2023-02-13
led: change command 0 to always return success

thedcap:
command 0 now always returns Success and the led counting functionality has been moved into the new command 4.

### Pull Request Overview

This pull request changes command 0 in the LED capsule to return success (based on issue #3375).
The previous command 0 was checking for led existence as well as returning the number of LEDs
available and has been moved into the new command 4.

### Testing Strategy

This pull request was tested by compiling tock along with modified libtock-c and libtock-rs that have
corresponding changes and then running the example blink (libtock-c) and led (libtock-rs) programs
and witnessing their continued functionality.

### TODO or Help Wanted

This pull request still needs...Nothing, I hope. Please advise me.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/src/led.rs
+++ b/capsules/src/led.rs
@@ -37,7 +37,7 @@
 //!
 //! #### `command_num`
 //!
-//! - `0`: Return the number of LEDs on this platform.
+//! - `0`: Check if there are LEDs on the board (always returns success).
 //!   - `data`: Unused.
 //!   - Return: Number of LEDs.
 //! - `1`: Turn the LED on.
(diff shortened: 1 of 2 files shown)


alexandruradovici:
I suggest we block this until we discuss how to handle this issue, as this is valid for GPIO, LEDs and many other capsules.

hudson-ayers:
I agree with the approach taken in this PR, and think that we should try to apply it to other capsules which are currently non-compliant with TRD104.

Happy to take a poll on this in Friday's meeting before we merge this though

hudson-ayers:
Based on the discussion on the core call, we actually cannot accept this PR because the LED driver has been marked as stable and we are not ready for a major version bump for some time.

Very sorry that you invested time into this!

However, there are several capsules that are not stable and can be converted now if you are interested in taking on one of those: https://github.com/tock/tock/issues/3375#issuecomment-1440647330

PR #3398, opened 2023-02-14
Add reset to process console

alexandruradovici:
### Pull Request Overview

This pull request adds a *reset* command to the process console. This is very useful for boards like the Raspberry Pi Pico that have no reset button and require a USB cable (power) disconnect for reset. For boards where there is not reboot implementation, it displays a message saying that the function is not implemented.

Additionally this pull requests adds:
- watchdog reboot for the RP2040
- stack size for the RP2040
- `scb::reset` for all cortex-m0+ and cortex-m4 boards

This pull request will be very useful for the Rust Nation presentation, so that attendees do not have to plug and unplug their boards.

### Testing Strategy

This pull request was tested using a Pico Explorer Base.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


lschuermann:
I do think that we should have some discussion about whether we can generalize "reboot" / "reset" and its various cross-chip and cross-board effects towards a more reusable core-interface. I am not necessarily opposed to "hacking" it into the process console temporarily and I can definitely see its utility, I constantly find myself typing `reset` into the process console.

Can we call it `reset` though? It's not actually "shutting" down anything and instead literally performing a device reset, and "reset" seems to be a more generally established term in the embedded space anyways.

boards/clue_nrf52840/src/main.rs
```diff
@@ -135,6 +135,13 @@ fn baud_rate_reset_bootloader_enter() {
     }
 }
 
+// Function for the process console to use to reboot the board
+fn reboot() {
+    unsafe {
+        cortexm4::scb::reset();
+    }
+}
+
```
hudson-ayers, line 144: Why not use this implementation on all of the cortex-m4 boards?

lschuermann (dismissed):

lschuermann:
Also, I presume that the boards where you've added the `reboot` command are just the ones where you've been able to test this on?

boards/components/src/process_console.rs
```diff
@@ -76,12 +77,14 @@ impl<const COMMAND_HISTORY_LEN: usize, A: 'static + Alarm<'static>>
         uart_mux: &'static MuxUart,
         alarm_mux: &'static MuxAlarm<'static, A>,
         process_printer: &'static dyn ProcessPrinter,
+        reboot_function: Option<&'static (dyn Fn() + 'static)>,
```
hudson-ayers, line 80: This is fine if we need to get it in in a hurry, but is there a reason not to use generics here to pass the function pointer?

boards/clue_nrf52840/src/main.rs
lschuermann, line 144: I'm unsure whether this should really be an unsafe function. It's difficult to reason about this from a Rust safety-perspective, given we're literally resetting all state. A valid argument for keeping it safe may be that `std::process::exit` is also a safe function.

In either case, it should return the never type though. We can guarantee to Rust that this function indeed never returns by simply adding a loop which will (assuming that the reset is indeed performed) just never be entered.

```suggestion
fn reboot() -> ! {
    unsafe {
        cortexm4::scb::reset();
    }
    loop {}
}
```

boards/components/src/process_console.rs
alexandruradovici: I just copied what the capsule driver does, I am happy to improve it with generics.

alexandruradovici:
> Can we call it `reset` though? It's not actually "shutting" down anything and instead literally performing a device reset, and "reset" seems to be a more generally established term in the embedded space anyways.
> 
Renamed to `reset`.

boards/components/src/process_console.rs
alexandruradovici: I am not sure how to specify the actual type of the closure in the component.

lschuermann: Shouldn't an `Option<fn() -> !>` work? Doesn't even use generics.

boards/clue_nrf52840/src/main.rs
alexandruradovici: This seems to be an unstable feature https://github.com/rust-lang/rust/issues/35121.

alexandruradovici: > I'm unsure whether this should really be an unsafe function. It's difficult to reason about this from a Rust safety-perspective, given we're literally resetting all state. A valid argument for keeping it safe may be that `std::process::exit` is also a safe function.
> 
The `cortexm4::scb::reset` function is unsafe.

alexandruradovici:
> Also, I presume that the boards where you've added the `reboot` command are just the ones where you've been able to test this on?

Yes, cortex-m4 boards and rp2040.

boards/components/src/process_console.rs
alexandruradovici: It works, and we do not actually use a closure.

boards/clue_nrf52840/src/main.rs
alexandruradovici: Done.

alexandruradovici: Done, thank you @lschuermann.

alexandruradovici:
It seems that using scb::reset works for cortexm0+.

lschuermann (approved):

hudson-ayers (approved):

twilfredo (approved):

brghena (commented):
This is fine, and I don't need to hold it up on this, but couldn't we have put the `reset()` function in the `arch` crate so each board file didn't have to separately implement it?

brghena (approved):

hudson-ayers:
> 

I think the problem is that `reset()` needs to be a safe function so it can be called from capsules, but we don't really want to expose a `safe` function in the arch crate that allows for resetting because we still are sort of relying on `unsafe` for access control to certain functions, even though resetting the chip is not memory unsafe. Though now that I think about it, only `chip` crates and `boards` crates depend on `arch` crates, so it may not really matter.

brghena:
Yeah, I see what you mean, but I think we could get away with it in `arch/`, just like we're getting away with it here in `boards/`.

alexandruradovici:
The `arch` create does have the reset function, just that it is unsafe and it does return a value (`scb::reset`). We could modify that. However, it would be great to have this merged for tomorrow morning's (this evening in the US) presentation, and I can submit a new PR with the modifications.

hudson-ayers:
bors r+

Merging so this is in for the presentation, a PR soon with a single (safe) reset function would be nice

PR #3404, opened 2023-02-21
Add reset function to cortexm::support

alexandruradovici:
### Pull Request Overview

This pull request adds a reset function to the `cortexm` crate in `cortexm::support::reset`. The function can be directly used by the process console as a *reset* function. Before this, boards had to define a function `fn reset() -> !` that used the `cortexm::scb::reset` function.

As discussed in https://github.com/tock/tock/pull/3398#pullrequestreview-1302001313, the `reset` function is marked as being safe as capsules do jot depend on the `cortexm` crate. I am not sure about this, maybe we should guard this using a permission so that even if a capsules does depend ion the arch crate, it cannot use this function.

### Testing Strategy

N/A

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


arch/cortex-m/src/support.rs
```diff
@@ -54,3 +55,14 @@ where
...
+pub fn reset() -> ! {
+    unsafe {
+        scb::reset();
+    }
+    loop {
+        // This is required to avoid the empty loop clippy
+        // warning #[warn(clippy::empty_loop)]
+        nop();
+    }
+}
```
hudson-ayers, line 68: I think that scb::reset should also have the signature `fn () -> !` and the nop loop should be moved into there.

bradjc:
I think in theory, boards would wrap the unsafe cortexm::reset function in a safe version, and pass that to pconsole. That would effectively be using the board's authority to call unsafe to determine that it is ok for _that_ capsule to have what is an unsafe function.

But, I think that adds unnecessary overhead. What is here looks fine.

bradjc (dismissed):

brghena (dismissed):

lschuermann:
I agree that `reset` is an incredibly useful command to have in the console and this is a net improvement. I rebased this as a562de8f8de5294b3d7a9b23e10e0022a3fd7828. Can you force push your branch to that commit, @alexandruradovici?

alexandruradovici:
@lschuermann I am not sure what you want me to do 😄 , my git skills are not that good.

lschuermann (approved):

alistair23 (approved):

arch/cortex-m/src/support.rs
bradjc: Sorry hudson

Issue #3244, opened 2022-09-26
RFC: Unsafe `WriteOnly`, `ReadWrite`, and `ReadOnly` (?) variants for tock-register-interface?


andre-richter:
In application grade processors, there potentially are registers that can compromise Rust's memory safety one way or the other. For example, registers that control virtual memory could be used to install page tables that create aliases which would break Rust's memory safety guarantees.

Manipulating such registers should be guarded with `unsafe` to stay true to Rust. If such registers should be implemented using tock-register-interface, unsafe variants of the `WriteOnly`, `ReadWrite`, and maybe `ReadOnly` interfaces (couldn't think of a real-world example for a register yet that compromises memory safety by just reading it) would be needed.

I am not sure if `tock` itself currently supports processor architectures that could make use of this. [`cortex-a`](https://github.com/rust-embedded/cortex-a), though, which is one of the more prominent users of tock-register-interface, would definitely embrace it.

So I wanted to first check with you if there is a general willingness to add unsafe interfaces variants?

jrvanwhy:
> So I wanted to first check with you if there is a general willingness to add unsafe interfaces variants?

I am supportive of adding unsafe interface variants.

I have been working on a rewrite of the `register_structs!` macro and register types that would collide with work on adding unsafe interface variants. My new design makes it easier to add and compose these variants (e.g. you could specify that one register is safe to read and unsafe to write, while another is unsafe to read or write), so it's probably worth waiting for the redesign to add this functionality.

andre-richter:
That sounds great! We can wait, no urgency here. It's just something that crossed my mind recently and that I'd like `cortex-a` to do better in the future than it does currently.

ppannuto:
+1 here that we would be happy to add support for something like this.

The only potential caveat I can think of is that we are always a bit wary of code size. However, my expectation is that an implementation of this would likely end up looking something like a sum of traits, and if Tock platforms don't use any of the unsafe traits, we won't see any size impact.

lschuermann:
We've had discussions like this before and previously sided with a different line of argument: https://github.com/tock/tock/issues/2593#issuecomment-847139818

[...]

PR #3405, opened 2023-02-21
[RFC] `tock-registers` macro rewrite try 2

jrvanwhy:
This is a request for comments on a possible redesign of the `register_structs!` macro in `tock-registers`. Note: the code is not ready for review -- it is incomplete, and I've implemented the most difficult parts without writing any comments.

Instead, read the [Design document](https://github.com/jrvanwhy/tock/blob/registers-proc-macro/doc/RegistersPeripheralDesign.md) and see if you can make sense of it.

1. Allows unit testing -- a driver test can implement a fake version of the hardware and use that to test the driver's functionality.
2. Resolves the unsoundness with pointers pointing into MMIO memory. The constructed registers type is zero-sized, so references to it do not reference any bytes, so the compiler cannot insert arbitrary reads.
3. Operations can be defined outside `tock_registers`. This design allows the `riscv-csr` crate to define its operations and retain the full functionality of `tock_registers` (including unit test functionality).
4. Hideously complex and hard to explain.

Note: I realize the design document may be incomprehensible -- please don't waste too much time trying to comprehend it. Instead, ask questions and I will respond (and improve the document if possible). I tried to explain the *why* behind the design but it got way too verbose to be useful, so instead I'm dumping the finished design in the doc without explaining the reasoning behind it.

### DIFF
--- a/doc/RegistersPeripheralDesign.md
+++ b/doc/RegistersPeripheralDesign.md
@@ -1,0 +1,68 @@
+`tock_registers::peripheral!` Macro and Trait Interface Design
+==============================================================
+
+## An oversimplified explanation
+
+`peripheral!` turns a register definition like:
+
+```rust
+peripheral! {
+    pub foo {
+        0x00 => ctrl: u32 { read + write }
+        0x04 => received: u8 { read }
+    }
+}
+```
+
+into a module, which contains an `Accessor` trait and a `Registers` struct:
+
+```rust
+pub mod foo {
+    trait Accessor: /* To be explained */ {}
+
+    struct Registers<A: Accessor> {
+        pub ctrl: tock_registers::Register<0x00, Self, A>,
+        pub received: tock_registers::Register<0x04, Self, A>,
+    }
+
+    /* Plus trait impls that I will explain later */
+}
+```
+
+## What is this `Accessor` trait?
+
+`Accessor` exists to facilitate unit testing. An `Accessor` provides access to a
+peripheral: either the real peripheral (e.g. via MMIO access) or a mock/fake
+version. Operations performed on the members of `foo::Registers` forward to the
+accessor; if you call `foo.ctrl.get()`, it will invoke a method on an instance
+of `A`.
+
+## Where is this `A` instance?
+
+Each `tock_registers::Register` instance needs access to an instance of `A`, so
+they must contain it:
+
+```rust
+// In tock_registers
+pub struct Registers<const REL_ADDR: usize, Peripheral, Accessor> {
+    pub accessor: A,
+    /* ... */
+}
+```
+
+Now the implementation of `foo.ctrl.get()` can call the accessor:
+
+```rust
+impl<Peripheral, A: ...> /*Mystery trait*/ for Register<..., Peripheral, A> {
+    fn get(&self) -> ... {
+        self.accessor.get::<...>()
+    }
+}
+```
+
+## What is this mystery trait?
+
+`Register` is defined inside `tock_registers`, but some of the operations are
+not. For example, RISC-V CSR traits are defined in the `riscv-csr` crate, which
+depends on `tock_registers`. Therefore, the methods on `Register` cannot be
+inherent impls. Instead, the 
--- a/doc/RegistersPeripheralDesign.md
+++ b/doc/RegistersPeripheralDesign.md
@@ -1,89 +1,9 @@
 `tock_registers::peripheral!` Macro and Trait Interface Design
 ==============================================================
 
-## Separating operations from the `tock_registers` crate
+## Basic terminology
 
-`register_structs!` requires callers to specify a type for each register that
-indicates which operations the registers can support. Examples include
-`ReadOnly`, `WriteOnly`, `ReadWrite`, and `ReadWriteRiscvCsr`. Most of these
-happen to be implemented in `tock_registers`, but that's not true of all. In
-general, the crate that specifies operations depends on `tock_registers`, and
-a generic view of the dependency tree looks like:
-
-```
-    +--------+
-    | Driver |
-    +--------+
-         |
-         V
-+------------------+
-| Operations crate |
-| (e.g. riscv-csr) |
-+------------------+
-         |
-         V
- +----------------+
- | tock_registers |
- +----------------+
-```
-
-For the purposes of this document, we'll assume that the read and write
-operations are defined outside the `tock_registers` crate, in their own crate.
-In practice, they'll be defined inside `tock_registers`, but other operations
-(such as `ReadWriteRiscvCsr`) will remain in external crates.
-
-Note that I referred to `read` and `write` as separate operations: the fact they
-are grouped together in a single type is an artifact of `register_structs!`'
-implementation. `peripheral!` treats each register as having a list of supported
-operations. This results in a dependency tree with multiple operations crates:
-
-```
-   +--------+
-   | Driver |
-   +--------+
-     |    |
-     V    V
-+------+ +-------+
-| read | | write |
-+------+ +-------+
-    |        |
-    V        V
-+----------------+
-| tock_registers |
-+----------------+
-```
-
-## Adding unit test support to the diagram
-
-One of the features we want to support in `peripheral!` is unit testing: it
-should be possible to use the generated struct with a fake or mock peripheral
-rather than the real hardware peripheral. This means that `peripheral!` cares
-about an even larger view of the dependency graph:
-
-```
-+-------------+ +-----------+
-| Real Kernel | | Unit test |
-+-------------+ +-----------+
-            |    |
-            V    V
-          +--------+
-          | Driver |
-          +--------+
-            |    |
-            V    V
-       +------+ +-------+
-       | read | | write |
-       +------+ +-------+
-           |        |
-           V        V
-       +----------------+
-       | tock_registers |
-       +----------------+
-```
-
-## High level: what should `peripheral!` generate?
-
-Given the following peripheral definition:
+A `peripheral!` invocation looks like the following:
 
 ```rust
 peripheral! {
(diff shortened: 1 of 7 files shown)


jrvanwhy:
**Discussion question:** This design can be simplified considerably if we make `peripheral!` only support MMIO registers, as it would no longer be necessary to allow external crates to add register types. Should we go forward with a complex design like this one, or a simpler design that doesn't support `riscv-csr`?

brghena:
What impact would this have on the [`register_bitfields!`](https://github.com/tock/tock/tree/master/libraries/tock-register-interface#defining-bitfields) stuff? Would that still be possible?

brghena:
I don't understand how this would actually change things from a "someone who uses registers" perspective. Could you show an example chips file (real or fake) and how it would use this.

jrvanwhy:
> What impact would this have on the [register_bitfields!](https://github.com/tock/tock/tree/master/libraries/tock-register-interface#defining-bitfields) stuff? Would that still be possible?

The `register_bitfields!` stuff would still be possible, yes. I did oversimplify the explanation: [`read::Register` in my design](https://github.com/jrvanwhy/tock/blob/registers-proc-macro/doc/RegistersPeripheralDesign.md#operation-implementation) would serve the same purpose as the [`Readable` trait in the current design](https://github.com/tock/tock/blob/master/libraries/tock-register-interface/src/interfaces.rs), and I would envision it having all the same operations (or in fact just using `Readable`). `Register` is a part of the "operation interface" that the `peripheral!` macro doesn't actually rely on, so individual operations are free to do other things.

Also, another thing I left out for simplification purposes: the operations can have generic parameters of their own which are appended to the trait instantiations. So an example of a register with bitflags would be:

```rust
0 => ctrl: u32 { read<CTRL::Register> + write<CTRL::Register> }
```

"Aliased" registers (using terminology from the existing design) can be represented by passing *different* generic arguments to `read` and `write`. We can also allow the following syntax:

```rust
0 => ctrl: u32 { [read,write]<CTRL::Register> }
```
as a shortcut so we aren't duplicating the generic arguments everywhere.

> I don't understand how this would actually change things from a "someone who uses registers" perspective. Could you show an example chips file (real or fake) and how it would use this?
>
>A good example could be the rp2040 [watchdog.rs](https://github.com/tock/tock/blob/master/chips/rp2040/src/watchdog.rs). It was written recently and is very few lines of real code.

I added [an example implementation of the rp2040 driver](https://github.com/jrvanwhy/tock/blob/registers-proc-macro/doc/RegistersPeripheralDesign.md#example-driver-implementation-rp2040-watchdog).

One other thing that would change is arrays of registers. Instead of:

```rust
(0 => data: [ReadOnly<u8>; 32]),
```

we would have separate array operations:

```rust
0 => data: [u8; 32] { array_read },
```

`array_read` would have an operation like `get(&self, index: usize) -> Option<Value>`. Arguably, a register could have both `read` and `array_read`; in that case, `read` would read the entire array while `array_read` would read the value at a particular index.

brghena:
Thank you! That RP2040 implementation is really helpful to me.

To my eyes, it seems like the definition of `peripheral!` is just arbitrarily different from `register_structs!`. Can you talk about why the DSL that people type needs to be different? It would theoretically be nice if we could change Tock Registers without changing contents of chip drivers.

As for your question about keeping external crates as register types, I personally think that _is_ valuable. If we want Tock registers to be useful to anyone outside of Tock, they should be able to extend it. So we can tank the complexity internally, as long as the external interface remains clean. However, that opinion is uninformed about the costs. I have no concept for what would have changed in your design if we didn't need external crates.

Related to that, generally, I don't have a concept for why any of the parts of the implementation exist. Particularly the `Access`, `Has`, and `Register` traits. What is each one used for? Your example showed me that `Accessor` is used for stubbing out implementations. This makes it hard to see which parts of the design are there to support which features.

brghena:
Also, this is a dumb comment and I hope someone can correct me, but is Unit Testing actually useful for Chips drivers in reality?

Conceptually, I can imagine creating some crazy unit test harness that takes in register values, has a state machine, and emulates hardware. Practically though, I can't imagine ever doing that. The amount of work it would take would be _significant_ without clear advantages (to me) over using other emulators like QEMU which have already done the work of emulating hardware.

So I'd expect that chip Unit Tests would be written for trivial peripherals (like a GPIO or something), but wouldn't be very useful because they're trivial. Or would be incredibly useful for complex peripherals (think USART on SAM4L) but would be so complex to write that they'll never actually exist.

I suspect I might be missing some important use case and am open to being corrected.

jrvanwhy:
> Also, this is a dumb comment and I hope someone can correct me, but is Unit Testing actually useful for Chips drivers in reality?
> 
> Conceptually, I can imagine creating some crazy unit test harness that takes in register values, has a state machine, and emulates hardware. Practically though, I can't imagine ever doing that. The amount of work it would take would be _significant_ without clear advantages (to me) over using other emulators like QEMU which have already done the work of emulating hardware or just testing on actual hardware.
> 
> So I'd expect that chip Unit Tests would be written for trivial peripherals (like a GPIO or something), but wouldn't be very useful because they're trivial. Or would be incredibly useful for complex peripherals (think USART on SAM4L) but would be so complex to write that they'll never actually exist.

I have done so as part of porting Tock to H1. There are a number of benefits of unit testing drivers over using an emulator like QEMU, including:

1. It allows you to unit test hardware that is not emulated. In particular: OpenTitan is not interested in maintaining emulators for its hardware. This was a carefully-reasoned decision based on past experience with the cost/benefit tradeoff of maintaining hardware-specific emulators.
2. It allows you to test code that handles hardware failures. QEMU emulation (and other testing approaches) typically don't let you simulate hardware that *breaks* and throws error flags back at the driver.
3. It allows you to test your code under Miri, and run expensive tests like fuzz tests that are far too slow to run in QEMU.

brghena:
Thanks for the thoughts! Hardware failures in particular resonates with me and wasn't something I was considering.

jrvanwhy:
> To my eyes, it seems like the definition of `peripheral!` is just arbitrarily different from `register_structs!`. Can you talk about why the DSL that people type needs to be different? It would theoretically be nice if we could change Tock Registers without changing contents of chip drivers.

The DSL doesn't actually *need* to be different, but there are several good reasons to make it different:

## `mod` not `struct`

It generates a `mod` instead of a `struct`, so the capitalization of the name changes (`WatchdogRegisters` becomes `watchdog`). Technically it could procedurally generate trait names and output something like:

```rust
struct WatchdogRegisters { ... }

// Generated by appending "Access" to whatever the caller passed in
trait WatchdogRegistersAccess: ... {}
```

but that isn't hermetic (these could collide) and I wasn't happy with it.

## Fixed register type -> variable length operations list

Conceptually, the new design treats each register as having a variable-length list of operations, whereas the old design only supports a fixed set of operation lists. This makes it more extensible. For example, some systems have registers that are unsafe to write (see https://github.com/tock/tock/issues/3244). We can introduce an `unsafe_write` trait that has the same interface as `write` except using it is unsafe, then a register can mix and match safe and unsafe reads and writes:

```rust
peripheral! {
    // A hardware RNG that writes out lots of bytes using direct memory access.
    dma_rng {
        // When written to, this will write 32 bytes of random data starting
        // at the provided address. When read from, returns the last address
        // that was written.
        0x0 => dma_target: usize { read + unsafe_write },
    }
}
```

As a corollary, I didn't want you to have to write:

```rust
0x0 => ctrl { read<u32> + write<u32> }
```

so I moved the data type of the register out of the operation list's generics and into its own section.

## The exposed API is changing anyway, so why not?

The API exposed by the generated types is changing, so the drivers need to change:

1. Using register operations requires a `use <op>::Register` statement, because they're all in macros. Technically this doesn't need to apply to operations defined in `tock_registers` (because they can be inherent impls on `tock_registers::Register`), but it's conceptually part of the design.
2. Array-typed registers will have a different API, because this design does not allow us to implement `Index` on them. Instead, we'll have to implement our own indexing APIs.

Ultimately, most of the work in updating drivers to the new design will be in changing the driver code to use the new generated API, not in tweaking the DSL passed to `register_structs!`/`peripheral!`

jrvanwhy:
> As for your question about keeping external crates as register types, I personally think that _is_ valuable. If we want Tock registers to be useful to anyone outside of Tock, they should be able to extend it. So we can tank the complexity internally, as long as the external interface remains clean. However, that opinion is uninformed about the costs. I have no concept for what would have changed in your design if we didn't need external crates.

I've been thinking about how I could simplify this if we didn't support external crates as register types, and I'm not having much success. So I think it makes sense to continue aiming to support operations defined in external traits.

jrvanwhy:
> Related to that, generally, I don't have a concept for why any of the parts of the implementation exist. Particularly the `Access`, `Has`, and `Register` traits. What is each one used for? Your example showed me that `Accessor` is used for stubbing out implementations. This makes it hard to see which parts of the design are there to support which features.

`Register` is just the trait implemented on `tock_register::Register` to provide a given operation. It allows a crate outside `tock_registers` to add methods to `tock_register::Register` so that driver crates can use those methods.

`Has` is just a metaprogramming signal: it doesn't do anything on its own, it just tells other `impls` what operations should be implemented on a trait. Its only use is in the trait bounds on the `Register` impls.

`Access` is the underlying implementation of the operation, which may be real (e.g. via MMIO or special assembly instructions) or fake (e.g. a fake peripheral for testing).

Does that make sense? I find this part of the design really hard to explain.

phil-levis:
I read through the design document. I'd suggest formatting/structure it this way:
  1. Start with the first 3 properties of the design, except frame them as requirements. We need `registers` to do these 3 things, and the current approach does not.
  2. Show the example peripheral and how you use it. Trim down `Watchdog`, e.g., just show `ctrl`, `load`, and `tick`. Remove the resets part (for simplicity). Explain what an `Accessor` is and why it's needed. 
  3. Show how a different `Accessor` (`Fake`?) can be used.
  4. Show how the registers are expanded, describing what the different traits do (`read::Has`, `read::Access`, `read::Register`) do and why.
  5. Describe how `read` and `write` are both Operations, and actually show the definitions/implementations of the Operations.
  6. Show a full/complete example.

PR #3407, opened 2023-03-01
arch/rv32i: clobber kernel registers instead of manual (un)stacking

lschuermann:
### Pull Request Overview

It turns out that Rust allows us to mark the majority of kernel registers as clobbered in an `asm!()` block on `rv32i`. We can use this in the context switch code to drastically reduce the number of registers we need to manually stack and unstack, and instead let Rust do this work for us.

In the worst case, this simply means that we have less lines of asm, and thus less things for us to break / a slightly more readable context switch handler. It is not any more or less efficient for Rust to stack and unstack registers before and after this `asm!()` block, then it is for us to do it ourselves.

However, it turns out that this change allows Rust to actually only stack the registers which need to be restored in practice, according to the surrounding function's ABI. Thinking in terms of caller/callee-saved registers is a bit tricky here, given that Rust doesn't make any promises concerning the ABI of these functions (not tagged as e.g. `extern "C"`), in practice we can expect quite a few instructions to be saved here.

From some initial, very unscientific tests (aka. looking at the generated assembly) it seems that we could save a good ~55 instructions per context switch in the best case here. Also, this does seem to come with some code size savings (about 100 bytes on the LiteX sim board).

Further investigation is necessary as to the true runtime impact & correctness of this, but it seems to work in initial tests.

### Testing Strategy

### TODO or Help Wanted

it needs more testing / squinting at the generated assembly / scrutinizing the diff.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/syscall.rs
+++ b/arch/rv32i/src/syscall.rs
@@ -242,85 +242,39 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
         // will issue arbitrary memory accesses (acting as a compiler
         // fence).
         asm!("
-          // Before switching to the app we need to save the kernel registers to
-          // the kernel stack. We then save the stack pointer in the mscratch
-          // CSR (0x340) so we can retrieve it after returning to the kernel
-          // from the app.
+          // Before switching to the app we need to save some kernel registers
+          // to the kernel stack, specifically ones which we can't mark as
+          // clobbered in the asm!() block. We then save the stack pointer in
+          // the mscratch CSR (0x340) so we can retrieve it after returning to
+          // the kernel from the app.
           //
           // A few values get saved to the kernel stack, including an app
           // register temporarily after entering the trap handler. Here is a
           // memory map to make it easier to keep track:
           //
           // ```
-          // 34*4(sp):          <- original stack pointer
-          // 33*4(sp):
-          // 32*4(sp): x31
-          // 31*4(sp): x30
-          // 30*4(sp): x29
-          // 29*4(sp): x28
-          // 28*4(sp): x27
-          // 27*4(sp): x26
-          // 26*4(sp): x25
-          // 25*4(sp): x24
-          // 24*4(sp): x23
-          // 23*4(sp): x22
-          // 22*4(sp): x21
-          // 21*4(sp): x20
-          // 20*4(sp): x19
-          // 19*4(sp): x18
-          // 18*4(sp): x17
-          // 17*4(sp): x16
-          // 16*4(sp): x15
-          // 15*4(sp): x14
-          // 14*4(sp): x13
-          // 13*4(sp): x12
-          // 12*4(sp): x11
-          // 11*4(sp): x10
-          // 10*4(sp): x9
-          //  9*4(sp): x8
-          //  8*4(sp): x7
-          //  7*4(sp): x6
-          //  6*4(sp): x5
-          //  5*4(sp): x4
-          //  4*4(sp): x3
-          //  3*4(sp): x1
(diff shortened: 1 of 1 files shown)


alistair23 (approved):

arch/rv32i/src/syscall.rs
```diff
@@ -422,45 +376,39 @@ impl kernel::syscall::UserspaceKernelBoundary for SysCall {
...
           // particularly relevant, however we must avoid using t0
           // as that is overwritten prior to being accessed
           // (although stored and later restored) in the assembly
           in("a0") state as *mut Riscv32iStoredState,
+
+          // Clobber all registers which can be marked as clobbered, except
+          // for `a0` / `x10`. By making it retain the value of `&mut state`,
+          // which we need to stack manually anyway, we can avoid Rust/LLVM
+          // stacking it redundantly for us.
+          lateout("x1") _, lateout("x5") _, lateout("x6") _,
```
brghena, line 403: Could you explain the purpose of using `lateout` here instead of `out`?

It seems like the only difference is that `lateout` registers could be reused input registers, after they've already be read and utilized in the code and their value no longer matters. That doesn't seem to apply here though, since all registers are explicitly named, and there's only one input register anyways, which doesn't overlap with any of the `lateout` registers.

brghena: Okay, that makes sense to me then. I thought there was some deeper reason why they _needed_ to be `lateout`.

alistair23: The Rust docs say that lateout "can be used on any output that is written only after all inputs have been consumed.", which applies to us here. So it does seem like the one to use

brghena (approved):

arch/rv32i/src/syscall.rs
lschuermann: > "can be used on any output that is written only after all inputs have been consumed.", which applies to us here

While technically true, I don't think this is what the Rust developers intended it to mean. I read it as simply a notice that you'd need to be careful with `lateout` to not clobber a register after you've read its contents. There's really no reason to use `lateout` if a register is solely defined as `in` and it won't be clobbered by the assembly. For cases where a register is solely defined as `out`, Rust will not rely on its contents not changing, and allow subsequent code to access this register's value. However, it won't work when a register is also defined as `in`; the inability to use `in` + `out` on a single register is the sole purpose of `lateout` (which we don't do here).

hudson-ayers: It seems a little unclear to me which is preferable (I believe either would be "correct"). On one hand, using `out` gives the benefit leon mentioned -- you get a warning if you later change the input registers such that there is an alias, and that is nice because you probably do not want to unintentionally alias your in and output registers. But using `lateout` presumably decreases register pressure a little, since the compiler can feel free to keep whatever was already in the `lateout` register up until the final input is read.

That said, I am not sure the compiler would even make use of that second optimization -- it seems possible that `lateout` is only beneficial if you are not selecting a specific register, and instead using the generic `reg`, such that the compiler can just reuse any of the input registers.

brghena: > it seems possible that lateout is only beneficial if you are not selecting a specific register, and instead using the generic reg, such that the compiler can just reuse any of the input registers.

After doing some reading, this was my takeaway. I think it's irrelevant when picking specific registers.

alistair23 (approved):

bradjc (approved):
This is great.

I wrote that original stacking code to reduce uncertainty, hoping eventually we would scrutinize it, so I'm glad to see that happening :).

PR #3411, opened 2023-03-08
Update OpenTitan Makefile and build examples 

jwnrt:
## Pull Request Overview

This pull updates parts of OpenTitan's Makefile and README that I ran into when trying to build and run Tock:

## Disable default board configuration

Disabled the default `fpga_cw310` Cargo feature when a `BOARD_CONFIGURATION` is specified.

Previously, both `fpga_cw310` and `sim_verilator` would be set, leading to conflicting `const CONFIG` definitions in `chips/earlgrey`.

### Objcopy binary

The Makefile is using `riscv64-linux-gnu-objcopy` as documented in https://github.com/tock/tock/pull/2104. The reasoning was that `llvm-objcopy` wasn't providing correct output.

From https://github.com/tock/tock/pull/2104 it looks like `riscv64-linux-gnu` was chosen because users were likely to have it installed. `riscv32-unknown-elf-objcopy` is provided by the lowRISC toolchain used for building OpenTitan software, so I thought developers would be more likely to have this already installed.

Happy not to include this change, though. I also haven't checked whether the `llvm-objcopy` bug is now resolved and that could be used instead.

### README examples

The QEMU examples in the README were outdated, possibly since `libtock-rs` has updated. Updated these examples, and tidied up some Markdown along the way.

## Testing Strategy

This pull request was tested by running the updated README examples and make rules.

## TODO or Help Wanted

Any thoughts on using `riscv32-unknown-elf-objcopy` by default over `riscv64-linux-gnu`?

## Documentation Updated

- [x] Updated example commands in OpenTitan's README

## Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/README.md
+++ b/boards/opentitan/README.md
@@ -287,11 +287,10 @@ The TBF must be compiled for the OpenTitan board. For example, you can build
 the Hello World example app from the libtock-rs repository by running:
 
 ```shell
-cd [LIBTOCK-RS-DIR]
-make flash-opentitan
-tar xf target/riscv32imac-unknown-none-elf/tab/opentitan/hello_world.tab
-cd [TOCK_ROOT]/boards/opentitan
-make APP=[LIBTOCK-RS-DIR]/rv32imac.tbf qemu-app
+cd "$libtock_rs_dir"
+make opentitan EXAMPLE=console
+cd "${tock_dir}/boards/opentitan/earlgrey-cw310"
+make APP=$"{libtock_rs_dir}/target/tbf/opentitan/console.tbf" qemu-app
 ```
 
 QEMU GDB Debugging [**earlgrey-cw310**]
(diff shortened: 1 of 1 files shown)


bradjc (dismissed):

hudson-ayers:
> The Makefile is using riscv64-linux-gnu-objcopy as documented in https://github.com/tock/tock/pull/2104. The reasoning was that llvm-objcopy wasn't providing correct output.
> 
> From https://github.com/tock/tock/pull/2104 it looks like riscv64-linux-gnu was chosen because users were likely to have it installed. riscv32-unknown-elf-objcopy is provided by the lowRISC toolchain used for building OpenTitan software, so I thought developers would be more likely to have this already installed.

`libtock-c` provides instructions for users to download a riscv toolchain, and those instructions install the riscv64-* tools (https://github.com/tock/libtock-c#prerequisites), so I think that many tock developers are more likely to have those installed. Because just building the binary requires objdump, plenty of tock developers who are just testing that their changes build (e.g. `make boards` or `make ci-all`) will hit this even if they never have actually looked into building opentitan at all. I would probably prefer keeping the default how it is for now unless we also update the toolchain instructions so there is a clearer path for users to get the default toolchain installed

jwnrt:
> I would probably prefer keeping the default how it is for now unless we also update the toolchain instructions so there is a clearer path for users to get the default toolchain installed

Thanks you, this makes sense

I haven't looked at `libtock-c`, so it would be nice to get toolchain instructions into `libtock-rs` too. For now, I've just dropped that commit.

hudson-ayers (approved):

twilfredo (approved):

alistair23 (dismissed):

alistair23 (approved):

hudson-ayers (approved):
bors r+

PR #3422, opened 2023-04-14
nrf52: support new access port protection mechanism on new nRF52 chips

bradjc:
### Pull Request Overview

It seems like recently nordic got bad press for having a vulnerability in how they disable jtag access, so newer versions of the nrf52840 change how they restrict jtag access, which essentially restricts jtag by default. This PR adds support to disable the protection so we can use and reprogram newer nRFs without re-flashing every time.

This requires setting both a "HW" register (aka in flash), and a "SW" register (aka in RAM) to 0x5a. The HW register can just be done once. The SW register needs to be done on every boot. Note, this is only for the new versions. So this code checks the variant and only includes it on the new variant.

### Testing Strategy

Running tock on a new nrf52840dk board.

### TODO or Help Wanted

~~Still need to set the UICR, or for now running `nrfjprog --recover` works.~~

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/nrf52/src/approtect.rs
+++ b/chips/nrf52/src/approtect.rs
@@ -4,7 +4,7 @@
 
 //! Access port protection
 //!
-//! https://infocenter.nordicsemi.com/index.jsp?topic=%2Fps_nrf52840%2Fdif.html&cp=5_0_0_3_7_1&anchor=register.DISABLE
+//! <https://infocenter.nordicsemi.com/index.jsp?topic=%2Fps_nrf52840%2Fdif.html&cp=5_0_0_3_7_1&anchor=register.DISABLE>
 //!
 //! The logic around APPROTECT was changed in newer revisions of the nRF52
 //! series chips (Oct 2021) and later which requires more careful disabling of
@@ -67,8 +67,8 @@ impl Approtect {
     /// disabled both in the UICR register (hardware) and in this register
     /// (software). For older variants this is just a no-op.
     ///
-    /// - https://devzone.nordicsemi.com/f/nordic-q-a/96590/how-to-disable-approtect-permanently-dfu-is-needed
-    /// - https://devzone.nordicsemi.com/nordic/nordic-blog/b/blog/posts/working-with-the-nrf52-series-improved-approtect
+    /// - <https://devzone.nordicsemi.com/f/nordic-q-a/96590/how-to-disable-approtect-permanently-dfu-is-needed>
+    /// - <https://devzone.nordicsemi.com/nordic/nordic-blog/b/blog/posts/working-with-the-nrf52-series-improved-approtect>
     pub fn sw_disable_approtect(&self) {
         let factory_config = ficr::Ficr::new();
         match factory_config.variant() {
--- a/chips/nrf52/src/approtect.rs
+++ b/chips/nrf52/src/approtect.rs
@@ -79,7 +79,27 @@ impl Approtect {
                 // updated to recognize it.
                 self.registers.disable.write(Disable::DISABLE::SWDISABLE);
             }
-            _ => {
+
+            // Exhaustively list variants here to produce compiler error on
+            // adding a new variant, which would otherwise not match the above
+            // condition.
+            ficr::Variant::AAA0
+            | ficr::Variant::AAAA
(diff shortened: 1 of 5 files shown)


twilfredo (approved):

chips/nrf52/src/uicr.rs
```diff
@@ -188,4 +191,28 @@ impl Uicr {
     pub fn set_ap_protect(&self) {
         self.registers.approtect.write(ApProtect::PALL::ENABLED);
     }
+
+    /// Disable the access port protection in the UICR register. This is stored
+    /// in flash and is persistent. This behavior can also be accomplished
+    /// outside of tock by running `nrfjprog --recover`.
+    ///
+    /// The `hardware_only` flag indicates whether
+    pub fn disable_ap_protect(&self) {
```
brghena, line 200: It doesn't look like you ever call this. Did you mean to?

bradjc: Added.

bradjc:
I added functionality to now set the approtect field in the UICR in the startup component.

I also refactored how we check for which variant in uicr.rs when disabling the approtect function.

I tested this PR on a new board and an old board (nrf52840dk) and both work.

brghena (approved):

twilfredo (approved):

bradjc:
What's wrong with netlify? Something in this PR?

twilfredo:
> What's wrong with netlify? Something in this PR?

Yea the error is `this URL is not a hyperlink`, I think it wants the urls to be wrapped in `<>`

chips/nrf52/src/approtect.rs
```diff
@@ -0,0 +1,87 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright Tock Contributors 2023.
+
+//! Access port protection
+//!
+//! https://infocenter.nordicsemi.com/index.jsp?topic=%2Fps_nrf52840%2Fdif.html&cp=5_0_0_3_7_1&anchor=register.DISABLE
```
twilfredo, line 7: here `<url>`

chips/nrf52/src/approtect.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /// Software disable the Access Port Protection mechanism.
+    ///
+    /// On newer variants of the nRF52, to enable JTAG, APPROTECT must be
+    /// disabled both in the UICR register (hardware) and in this register
+    /// (software). For older variants this is just a no-op.
+    ///
+    /// - https://devzone.nordicsemi.com/f/nordic-q-a/96590/how-to-disable-approtect-permanently-dfu-is-needed
+    /// - https://devzone.nordicsemi.com/nordic/nordic-blog/b/blog/posts/working-with-the-nrf52-series-improved-approtect
```
twilfredo, line 71: these two here `<url>`

bradjc:
What? This tool can tell these are URLs so it can give us errors informing us it needs us to tell it these are URLs? What problem is this solving, rustdoc?

twilfredo:
> What? This tool can tell these are URLs so it can give us errors informing us it needs us to tell it these are URLs? What problem is this solving, rustdoc?

I think this is just a lint being treated as an error, see [bare_urls](https://doc.rust-lang.org/rustdoc/lints.html)

twilfredo (approved):

brghena (approved):

ppannuto (dismissed):

lschuermann (approved):
Thanks, this is great! I'm yet to test it as I didn't have access to my (old) nRF52840DK.

I added a minor change which exhaustively lists all the chip variants for which the `SwDisable` switch is not required. This is to prevent the scenario where a new revision would be supported at first (because it matches on the `Unspecified` variant), and someone adds a definition for it to the `ficr` module, which would then implicitly break it in a non-obvious way due to the default match. Having the exhaustive list ensures that the compiler yells at us when this `match` diverges from the `ficr::Variant` enum definition.

bradjc:
Ready?

brghena (approved):

PR #3426, opened 2023-04-21
chips: apollo3: Set a second timer for alarms

alistair23:
### Pull Request Overview

From the Apollo3 errata list there is an entry:

"ERR026: STIMER: Glitch on XTAL clock input may cause missed compare"

that describes some alarms not triggering interrupts due to jitter in the clock.

The workaround is to set a second alarm 1 tick later, to ensure that at least one of them will trigger.

This patch adds that second alarm as I was seeing timer interrupts be missed on some boards. This fixes the issue

### Testing Strategy

Running timers on an Apollo3 board

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/chips/apollo3/src/stimer.rs
+++ b/chips/apollo3/src/stimer.rs
@@ -4,15 +4,14 @@
 
 //! STimer driver for the Apollo3
 
-use kernel::utilities::cells::OptionalCell;
-use kernel::ErrorCode;
-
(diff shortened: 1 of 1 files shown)


twilfredo (approved):

alistair23:
This is ready to go :)

hudson-ayers (approved):

PR #3440, opened 2023-05-19
capsules: change ctap to usb hid

bradjc:
### Pull Request Overview

I'm looking for comments on this PR.

I have a USB HID keyboard device working, and if we want to expose that functionality to userspace then we need a driver capsule. Since I based my implementation on the ctap code, that seemed like a logical place to start. The current ctap capsule is a pretty straightforward pass through for buffers between userspace and the USB stack.

This PR renames that ctap capsule to be generic USB HID. It also adds command 0 as a check command per convention.

@alistair23 thoughts?

I imagine the idea was since CTAP can support multiple transport layers the ctap capsule could expand to support multiple layers, isolating applications from that complexity. However, much like with the lora PR, perhaps it's better to have one capsule per transport, and have boards use the same driver num and then choose the correct (generic) transport layer capsule.

### Testing Strategy

Building the USB HID keyboard.

### TODO or Help Wanted

Need to decide if we want this PR, or should just create a separate (essentially duplicate) USB HID or USB HID keyboard capsule).

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alistair23:
I think this is a good idea.

As we currently only support CTAP over USB HID, this seems like the way to go.

The only downside is that if we ever add BLE or NFC support we then need userspace to figure out how to send/receive messages. For some use cases that might be the best bet anyway.

bradjc:
This is ready to go.

alistair23 (approved):

ppannuto (approved):
I think this is a good simplification for the current state of things. If/when we actually support multiple transport layers, that will be a better time to figure out the 'right' way to multiplex.

alevy (approved):

PR #3443, opened 2023-05-22
make: llvm-tools no longer preview

bradjc:
### Pull Request Overview

Without this change I was getting:

```
$ make
info: component 'llvm-tools' for target 'x86_64-apple-darwin' is up to date
```
on every build. It seems llvm-tools is now a real thing, and no longer a preview.

### Testing Strategy

running make

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/Makefile.common
+++ b/boards/Makefile.common
@@ -23,12 +23,12 @@ RUSTC_SYSROOT := "$(shell rustc --print sysroot)"
 
 # Common defaults that specific boards can override, but likely do not need to.
 #
-# The TOOLCHAIN parameter is set to the magic value "llvm-tools-preview", which
-# will cause the Makefile to resolve the llvm toolchain installed as part of the
-# rustup component "llvm-tools-preview". In case a system toolchain shall be
-# used, this can be overridden to specify the toolchain prefix, e.g. "llvm" for
+# The TOOLCHAIN parameter is set to the magic value "llvm-tools", which will
+# cause the Makefile to resolve the llvm toolchain installed as part of the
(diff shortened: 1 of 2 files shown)


alistair23 (approved):

hudson-ayers (approved):

bradjc:
This one is easy!

shell.nix
```diff
@@ -73,7 +73,7 @@ let
     ];
     extensions = [
       "rust-src" # required to compile the core library
-      "llvm-tools-preview" # currently required to support recently added flags
+      "llvm-tools" # currently required to support recently added flags
```
lschuermann, line 76:
```suggestion
      "llvm-tools-preview" # currently required to support recently added flags
```

This breaks on Nix unfortunately. I tried to look into the root cause of this, and it seems that the Rust overlay we use extracts the extensions from the actual upstream Rust package, so it seems there's at least one point in the chain where this rename hasn't propagated.

I tested these changes with `llvm-tools-preview` and managed to compile a board / run a few makefile commands. Happy to see this merged without this particular change, and I can investigate it when I'm not on airport wifi.

lschuermann (changes requested):

shell.nix
bradjc: I reverted the change to shell.nix.

lschuermann (approved):

bradjc:
This one is a real quality of life improvement.

alistair23 (approved):

ppannuto (approved):
bors r+

PR #3445, opened 2023-05-23
Add an RSA2048 credential checker

bradjc:
### Pull Request Overview

This pull request is a rough proof of concept for doing a software-based RSA2048, SHA256, PKCS1 v1.5 signature validation for apps.

Issues:

1. It uses the RustCrypto RSA library which needs an allocator. I'm using the https://docs.rs/emballoc/latest/emballoc/ because it seems to work.
2. It adds a RSA signature HIL. I'm not sure what that should look like.
3. The kernel binary is a lot bigger and I had to move the app address on hail to 0x40000.
4. I got a ton of errors like:

    ```
       Compiling base64ct v1.6.0
       Compiling getrandom v0.2.9
    error[E0152]: duplicate lang item in crate `core` (which `alloc` depends on): `sized`.
      |
      = note: the lang item is first defined in crate `core` (which `base64ct` depends on)
      = note: first definition in `core` loaded from /Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/deps/libcore-80d7c25ab46a392b.rmeta
      = note: second definition in `core` loaded from /Users/bradjc/.rustup/toolchains/nightly-2023-04-13-x86_64-apple-darwin/lib/rustlib/thumbv7em-none-eabi/lib/libcore-3cc9bf0d133d2d4c.rlib
    ```
    so I removed our custom build of `core`.

I chose RSA2048 and SHA256 only because the ARM® TrustZone® Cryptocell 310 security subsystem on the nRF52840 supports those values.

This also builds on my attempt to differentiate processes based on credential checks with the sha256 credential.

### Testing Strategy

Tockloader can add a credential to an app:

```
tockloader tbf-add-credential rsa2048 --public-key tockkey2048.public.der --private-key tockkey2048.private.der
```

Then on hail this PR checks that the signature is correct.

I used Python to to get n and e for the verifier:

```python
from Crypto.PublicKey import RSA
f=open('tockkey2048.public.der', 'rb')
key = RSA.importKey(f.read())
key.n
key.e
```

### TODO or Help Wanted

A lot.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ members = [
     "boards/weact_f401ccu6/",
     "capsules/core",
     "capsules/extra",
+    "capsules/rsa_sw",
     "chips/apollo3",
     "chips/arty_e21_chip",
     "chips/e310_g002",
--- a/boards/Makefile.common
+++ b/boards/Makefile.common
@@ -231,7 +231,6 @@ CARGO_FLAGS ?=
 #   https://github.com/tock/tock/pull/2847 for more details.
 CARGO_FLAGS_TOCK ?= \
   $(VERBOSE_FLAGS) \
-  -Z build-std=core,compiler_builtins \
   --target=$(TARGET) \
   --package $(PLATFORM) \
   --target-dir=$(TARGET_DIRECTORY) $(CARGO_FLAGS)
--- a/boards/hail/Cargo.toml
+++ b/boards/hail/Cargo.toml
@@ -17,3 +17,5 @@ sam4l = { path = "../../chips/sam4l" }
 
 capsules-core = { path = "../../capsules/core" }
 capsules-extra = { path = "../../capsules/extra" }
+rsa-sw = { path = "../../capsules/rsa_sw" }
+emballoc = "0.1.2"
--- a/boards/hail/chip_layout.ld
+++ b/boards/hail/chip_layout.ld
@@ -2,7 +2,7 @@
 /* Bootloader is at address 0x00000000 */
 MEMORY
 {
-  rom (rx)  : ORIGIN = 0x00010000, LENGTH = 0x00020000
-  prog (rx) : ORIGIN = 0x00030000, LENGTH = 0x00040000
+  rom (rx)  : ORIGIN = 0x00010000, LENGTH = 0x00030000
+  prog (rx) : ORIGIN = 0x00040000, LENGTH = 0x00030000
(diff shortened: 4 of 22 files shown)


alistair23:
Cool!

I think it's worth splitting the HIL changes out and getting them merged first. I also think we should add a RSA HIL as part of this. The idea being that there can be a software implementation of the RSA HIL, but then we can also extend it to add a hardware implementation of RSA in the future.

That way we can easily swap the RSA backend depending on the board. With this design we will need to swap the actual verifier

bradjc:
Yeah I need help with integrating this. We have RSA math hil, do we have a HIL for RSA operations (in this case `verify()`)?

alistair23:
We currently just have basic RSA math operations `(message ^ exponent) % modulus`. So that that should be supported using the modpow() function (https://github.com/RustCrypto/RSA/blob/b65f352299c8535df7271600dbd04ea6060ba1e3/src/algorithms/rsa.rs#L16).

Then we will need a verify function on top of that

bradjc:
It seems like unfortunately this is blocked on https://github.com/RustCrypto/RSA/issues/51 (and https://github.com/RustCrypto/RSA/issues/342). I don't actually think that converting the portion of the RSA crate we need to use crypto-bigint (and stack-allocated keys) would be too terribly difficult, but, then what? We probably don't want our own vendored RSA crate in libraries, and RustCrypto/RSA (I assume) won't upstream a partial switch.

Maybe we should skip RSA. Anytime I mention it the first thing that comes up is wait what about ECDSA.

bradjc:
Work on moving the rsa library to crypto-bigint is underway https://github.com/RustCrypto/RSA/pull/394

bradjc:
https://github.com/RustCrypto/RSA/pull/394 is merged! I don't think that PR is everything we need, but I believe it enables a clear path to using non-dynamic types in the RSA library.

alistair23:
> Maybe we should skip RSA. Anytime I mention it the first thing that comes up is wait what about ECDSA.

Tock has hardware accelerated ECDSA checking now, so maybe a software RSA verifier isn't worth implementing

https://github.com/tock/tock/blob/master/boards/apollo3/lora_things_plus/src/main.rs#L870

bradjc:
Closing as replaced by #4372. It seems like non-alloc RSA is still a long ways off.

PR #3446, opened 2023-05-23
nrf52: remove unsafe and static mut from adc

bradjc:
### Pull Request Overview

Since the nRF52 adc peripheral uses DMA, we were using a static mut buffer to store the ADC result. This requires `unsafe`. This PR is one attempt to remove that unsafe.

Of course, now we are doing perhaps undesirable things with buffers stored in a TakeCell.

### Testing Strategy

travis

### TODO or Help Wanted

What is better? This, or having `unsafe` in the driver?

If OK, needs to be added to each nrf board.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alistair23 (approved):

bradjc:
Turns out this is used throughout the nrf52 peripheral drivers. I don't think it makes sense to complicate this for the adc without a more comprehensive look.

PR #3448, opened 2023-05-24
nRF52 AES Improvements

bradjc:
### Pull Request Overview

This pull request:

- Updates the nRF52 AES driver to support ECB and CBC, as well as in-place operations for ECB, CBC, and CTR.
- Adds an AES component.
- ~~Adds the userspace AES driver to the nRF52840dk board.~~
- Adds the AES tests to the nRF52840dk board.
- Updates the AES tests to run all configurations (encryption first, both diff buffers and in place, then decryption).

Note, the nRF52 ECB module does not support decryption, so the ECB and CBC decryption tests fail.

This pull request is me getting familiar with the AES support on the nRF52 in hopes of using it for the Tock Tutorial.

### Testing Strategy

Running the kernel AES tests.

### TODO or Help Wanted

When I run the libtock-c/aes test app, I get:

```
[TEST] AES CTR Example Test
Loading in the key...
   done
Loading in the IV buf...
   done
Loading in the first half of the source buf...
   done
Setting up the first half of the dest buf...
   done
Setting up the callback...
   done
Starting the initial run...
   done
Loading in second half of the source buf...
   done
Running AES twice...
   done
Finish AES operation...
   done
Encrypted text: '���␒�␇	ѕ�6␐��ΓSY␕R'
Loading in the decryption source buf...
   done
Setting up the decryption dest buf...
   done
Running decryption...
   done
Finish AES operation...
   done
Original text: 'A language empowering everyone to build reliable and efficient software.'
[TEST] AES CCM Example Test
Loading in the key...
   done
Loading in the nonce buf...
   done
Loading in the source buf...
tock$    done
Setting up the dest buf...
   done
Setting up the callback...
   done
Starting the initial run...
   done
Finish AES operation...
   done
Comparison failure at 0
Expected value: 0x4a
 But got value: 0x2
```

~~So the AES CTR works, but the CCM does not. I did not change any of the CCM code. Do we know if that test is supposed to work? I believe the userland test app uses a different size IV than the in-kernel CCM test.~~

Nevermind, since we don't have GCM we cannot use the userspace driver.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/aes.rs
+++ b/boards/components/src/aes.rs
@@ -1,0 +1,160 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright Tock Contributors 2022.
+
+//! Components for various AES utilities.
+//!
+//! Usage
+//! -----
+//! ```rust
+//! let aes_driver_device = components::aes::AesVirtualComponent::new(aes_mux).finalize(
+//!     components::aes_virtual_component_static!(nrf52840::aes::AesECB<'static>),
+//! );
+//!
+//! let aes = components::aes::AesDriverComponent::new(
+//!     board_kernel,
+//!     capsules_extra::symmetric_encryption::aes::DRIVER_NUM,
+//!     aes_driver_device,
+//! )
+//! .finalize(components::aes_driver_component_static!(
+//!     capsules_core::virtualizers::virtual_aes_ccm::VirtualAES128CCM<
+//!         'static,
+//!         nrf52840::aes::AesECB<'static>,
+//!     >
+//! ));
+//! ```
+
+use core::mem::MaybeUninit;
+use kernel::capabilities;
+use kernel::component::Component;
+use kernel::create_capability;
+use kernel::hil;
+use kernel::hil::symmetric_encryption::{
+    AES128Ctr, AES128, AES128CBC, AES128CCM, AES128ECB, AES128GCM,
+};
+
+const CRYPT_SIZE: usize = 7 * hil::symmetric_encryption::AES128_BLOCK_SIZE;
+
+#[macro_export]
+macro_rules! aes_virtual_component_static {
+    ($A:ty $(,)?) => {{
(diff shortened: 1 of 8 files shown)


boards/nordic/nrf52840dk/src/main.rs
```diff
@@ -712,11 +742,17 @@ pub unsafe fn main() {
...
+        aes,
         scheduler,
         systick: cortexm4::systick::SysTick::new_with_calibration(64000000),
     };
 
     let _ = platform.pconsole.start();
+
+    // test::aes_test::run_aes128_ctr(&base_peripherals.ecb);
+    // test::aes_test::run_aes128_cbc(&base_peripherals.ecb);
+    // test::aes_test::run_aes128_ecb(&base_peripherals.ecb);
```
alistair23, line 754: Why not use the unit test framework to run these instead of commenting them out?

boards/components/src/aes.rs
```diff
@@ -0,0 +1,152 @@
...
+use kernel::capabilities;
+use kernel::component::Component;
+use kernel::create_capability;
+use kernel::hil;
+use kernel::hil::symmetric_encryption::{AES128Ctr, AES128, AES128CBC, AES128CCM, AES128ECB};
+
+const CRYPT_SIZE: usize = 7 * hil::symmetric_encryption::AES128_BLOCK_SIZE;
+
+#[macro_export]
+macro_rules! aes_virtual_component_static {
```
alistair23, line 37: It'd probably be worth updating the OT board to use this as well

boards/nordic/nrf52840dk/src/main.rs
bradjc: That's the right thing to do but not something I'm going to take on right now.

bradjc:
Rebased. Ready to go again.

ppannuto (approved):
It's unfortunate that the use of a static mut buffer for DMA also means those array accesses also end up in the unsafe block, but I don't think things are any meaningfully worse than before. Added some explainer comments for the `unsafe`'s

PR #3452, opened 2023-05-24
kernel: process: freshen comments

bradjc:
### Pull Request Overview

Make formatting consistent, add missing comment docs, minor clarifications.

### Testing Strategy

n/a

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


kernel/src/process.rs
```diff
@@ -952,10 +983,10 @@ pub struct ProcessAddresses {
...
     /// the kernel has reserved for its own use.
     pub flash_non_protected_start: usize,
-    /// The address immediately after the end of part of the process
-    /// binary that is covered by integrity; the integrity region is
-    /// [flash_start - flash_integrity_end). Footers are stored in
-    /// the flash after flash_integrity_end.
+    /// The address immediately after the end of part of the process binary that
+    /// is covered by integrity; the integrity region is [flash_start -
+    /// flash_integrity_end). Footers are stored in the flash after
+    /// flash_integrity_end.
```
ppannuto, line 989: this looks strictly less readable

ppannuto (changes requested):
I think we either need to have a tool automatically wrapping comments to some maintained, specified width or you need to turn off automatic, whole file rewrapping in your editor.

It's really hard to figure out what actually changed with a lot of the doc updates, and churn for different, arbitrary line wrap setting doesn't feel worth it — here, it made what looks like some intentional formatting worse in a least one case.

bradjc:
> I think we either need to have a tool automatically wrapping comments to some maintained, specified width

I wouldn't argue with that!

> or you need to turn off automatic, whole file rewrapping in your editor.

I'm doing this manually. To me, the kernel crate is the most core crate to understand Tock, and it's worth having consistency in the presentation, even if that means periodic updates to the file because it is difficult to handle on each PR. I personally am willing to block PRs to the core kernel files until the comments are consistent, but I don't think that overhead is less than just doing it periodically.

> arbitrary line wrap setting

I wouldn't say it's completely arbitrary, at least not within this PR, it's 80 characters like the other comments in the file.

bradjc:
Anyone else have any thoughts? I think formatting comments is valuable. I also think that if the specific wrapping of a comment matters it should be in code block, since the formatting is not going to be preserved on the generated HTML docs anyway.

lschuermann:
> Anyone else have any thoughts? I think formatting comments is valuable.

My 2ct are: well-formatted docs are nice, but if something doesn't look too terrible we shouldn't necessarily feel the need to reformat it. So I don't think it's a big deal either way.

To give some examples: if somethings formatted to wrap at 70 chars, it seems fine to keep it that way. If inserting a missing word pushes a comment above the 80 chars limit, I don't know whether we need to re-wrap 20 other lines because of that.

I think the way we handled this previously was just fine, where people generally wouldn't re-wrap comments for minor changes, and then every once in a while or on major refactors, we'd adjust formatting.

lschuermann (approved):

hudson-ayers (approved):
I generally agree with Leon. For future PRs, I would prefer a two-commit approach where one commit includes the new text, and a separate commit just adjusts comment wrapping, to simplify reviewing

bradjc:
Easy enough to do. First commit is all of the formatting changes, second commit is text changes.

lschuermann (approved):

PR #3453, opened 2023-05-25
nrf52: adc: Add highspeed implementation.

bradjc:
### Pull Request Overview

This pull request adds an implementation of `hil::adc::AdcHighSpeed` for the nRF52.

The main issue is that the internal clock divider for the SAADC only allows dividing a 16 MHz clock by a value in the range 80 to 2047 (so 8 kHz to 200 kHz sampling). So there is a limited range of possible frequencies. This is also why it's hard to implement continuous sampling.

This also adds VDD detection to the calibrate step, since the nRF52840dk uses a 3V reference and I couldn't figure out why my ADC readings were wrong.

### Testing Strategy

Using the libtock-c adc app.

### TODO or Help Wanted

How good of a driver do we need?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nordic/nrf52840dk/src/main.rs
+++ b/boards/nordic/nrf52840dk/src/main.rs
@@ -844,6 +844,7 @@ pub unsafe fn main() {
     };
 
     let _ = platform.pconsole.start();
+    base_peripherals.adc.calibrate();
 
     // test::aes_test::run_aes128_ctr(&base_peripherals.ecb);
     // test::aes_test::run_aes128_cbc(&base_peripherals.ecb);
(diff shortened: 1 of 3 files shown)


lschuermann (dismissed):

lschuermann (dismissed):

ppannuto (dismissed):
I moved the default VREF setting up a level so that some other caller of `adc::new()` could set something else if needed. @bradjc I'll leave this for you to click merge in case there was some reason you'd not done that I didn't see.

bradjc:
Rebased. Looks good to me.

ppannuto (approved):

PR #3466, opened 2023-06-02
chips: nrf52: spi: auto disable when unused

bradjc:
### Pull Request Overview

This pull request fixes the issues I was having with using nrfjprog to read the external flash on the nRF52840dk boards. The root issue is the main nrf52840 mcu was driving the MOSI pin to the flash chip low. Since the MOSI pin is the LSB pin when the flash is in Quad SPI mode, the LSB was always reading 0.

I'm guessing that when the SPIM peripheral is enabled it is actively driving the MOSI pin. This PR then only enables the SPIM peripheral when it is busy, aka when a read/write is actually happening.

The driver wasn't really using the `initialized` flag, and since I don't think we should have init() functions in our HILs (#1112) I removed that in favor of just setting the interrupt and enabling when the peripheral is used.

### Testing Strategy

I verified the nonvolatile storage test app works, as well as the `nrfjprog --memrd 0x12000000 --n 32` command.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/nrf52/src/spi.rs
+++ b/chips/nrf52/src/spi.rs
@@ -241,7 +241,6 @@ pub struct SPIM<'a> {
     registers: StaticRef<SpimRegisters>,
     client: OptionalCell<&'a dyn hil::spi::SpiMasterClient>,
     chip_select: OptionalCell<&'a dyn hil::gpio::Pin>,
-    initialized: Cell<bool>,
     busy: Cell<bool>,
     tx_buf: TakeCell<'static, [u8]>,
     rx_buf: TakeCell<'static, [u8]>,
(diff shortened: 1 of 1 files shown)


lschuermann (approved):

brghena:
Woah, that's really tricky. How did you debug that?

It fully explains the behavior we saw too. Your board and Hudson's were probably running Tock, while mine had some class Arduino stuff loaded on it.

bradjc:
> Woah, that's really tricky. How did you debug that?
> 
> It fully explains the behavior we saw too. Your board and Hudson's were probably running Tock, while mine had some class Arduino stuff loaded on it.

Yeah when I had a board that I could read successfully from with the same UICR settings as one that didn't work, and flashing tock seemed to break it, it occurred to me the issue is likely with something that Tock is running, and not any of the jlink/nrfjprog tools/infrastructure.

alistair23 (dismissed):

twilfredo (approved):

lschuermann (approved):

lschuermann (approved):

PR #3468, opened 2023-06-06
capsules: spi_controller: Don't provide read buffer for write ops

alistair23:
### Pull Request Overview

The SPI HIL uses a single `read_write_bytes()` function which is used for both reading/writing. The `read_buffer` is optional if the caller just wants to write data and doesn't care about reading any data back they can provide `None`

The current userspace capsule always supplies a read buffer, even if userspace doesn't want to perform a read. This results in the data being read from the bus, passed back to the capsule and then not being used.

To avoid reading and copying data that was never requested, let's avoid passing the read buffer to the HIL implemenation if userspace doesn't want to read the data. This way implementations can avoid copying the data from the bus when it isn't going to be used.

### Testing Strategy

Running SPI operations

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/capsules/core/src/spi_controller.rs
+++ b/capsules/core/src/spi_controller.rs
@@ -120,12 +120,22 @@ impl<'a, S: SpiMasterDevice> Spi<'a, S> {
             app.index = start + tmp_len;
             tmp_len
         });
+
+        let rlen = kernel_data
+            .get_readwrite_processbuffer(rw_allow::READ)
+            .map_or(0, |read| read.len());
+
         // TODO verify SPI return value
-        let _ = self.spi_master.read_write_bytes(
-            self.kernel_write.take().unwrap(),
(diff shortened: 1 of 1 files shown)


capsules/core/src/spi_controller.rs
```diff
@@ -322,7 +332,9 @@ impl<S: SpiMasterDevice> SpiMasterClient for Spi<'_, S> {
                     src
                 });
 
-                self.kernel_read.put(rbuf);
+                if !rbuf.is_none() {
```
bradjc, line 335:
```suggestion
                if rbuf.is_some() {
```

bradjc (dismissed):

bradjc (approved):

PR #3469, opened 2023-06-06
Convert all CRLF files to Unix line endings.

jrvanwhy:
This conversion was performed using the `dos2unix` program.

The other `.jlink` files in the repository use Unix line endings, so CRLF line endings are not a requirement of the file type.


bradjc (approved):

alistair23 (approved):

twilfredo (approved):

ppannuto (commented):
bors r+

This touches a small corner, plus "The other .jlink files in the repository use Unix line endings, so CRLF line endings are not a requirement of the file type." puts this over the line move through as maintenance to IMO

Issue #3457, opened 2023-05-26
TicKV: Support for kernelspace and userspace


bradjc:
I'm experimenting with using TicKV in the kernel and as a way to share data between a host computer and a tock board (i.e. the host computer would use jtag to read and write the tickv kv store).

Currently, it looks like the capsule support for TicKV in Tock is aimed towards supporting userspace. What would it look like to support both userspace and in-kernel users of TicKV?

Some potential issues I see:
1. The kv_store capsule is both a wrapper on top of TicKV for tock userspace (defines a header format and uses StoragePermissions), as well as virtualizer. It seems like we would need to separate the virtualizer functionality.
2. Key collisions between userspace and the kernel.

alistair23:
You should be able to use `capsules/extra/src/kv_store.rs` in the kernel. Only the `KVSystemDriver` capsule (in `kv_driver.rs`) should be userspace specific.

You just need to have a kernel specific `StoragePermissions` for `kv_store.rs`. That could be as simple as reserving `0xFFFF_FFFF` for the kernel and not allowing applications to use/access those.

That doesn't fix key collisions, which are always a possible problem. One answer is to have separate instances (with different flash backing) for applications and kernel

bradjc:
How should this look? I ran into this because the constructor for `StoragePermissions` is `pub(crate)` so I couldn't create one in my main.rs file. Should the kernel just define a kernel storage permissions? Or should there be a capability-enforced constructor?

alistair23:
I think the kernel should define it's own `StoragePermissions`. The kernel is then in a position to enforce that no one else can use those permissions. Either in `get_storage_permissions()` or in the application loading.

I don't think it should be created in `main.rs`. The kernel permissions should be the same across all boards, not up to each board. Does that sound reasonable?

bradjc:
In that case I'm imagining something like:

in kernel crate:

```rust
fn get_kernel_storage_permissions(cap: &Capability) -> StoragePermissions {
   // return standard kernel permissions
}
```

alistair23:
Yeah, I think that seems like the way to go

hudson-ayers:
Closed by #3470

PR #3470, opened 2023-06-07
kernel: StoragePermissions: add "kerneluser" storage access

bradjc:
### Pull Request Overview

This pull request adds a mechanism for the kernel to have special access to its items on a board, _and_ to use a storage platform shared with processes while keeping keys reserved for itself. This should allow us to use TicKV in the kernel.

Since the TBF definition already effectively reserved write_id 0 to mean "this process cannot write items" this PR uses write_id 0 for the kernel. This also formalizes that 0 means no writing by saving `write_id` as an `Option<NonZeroU32>`.

To give the kernel access I added a `kerneluser` flag to the `StoragePermission`. It allows the kernel to access items with write_id 0.

To create a kernel level `StoragePermission` I added a capability (`KerneluserStorageCapability`) and then made the constructor protected on that capability.

### Testing Strategy

This pull request was tested by compiling. The changes are largely mechanical.

### TODO or Help Wanted

Thoughts on reserving 0 for the kernel?

### Documentation Updated

The TBF doc already reserves the write id 0, so I didn't need to update anything.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/storage_permissions.rs
+++ b/kernel/src/storage_permissions.rs
@@ -5,6 +5,9 @@
 //! Mechanism for managing storage read & write permissions.
 
 use core::cmp;
+use core::num::NonZeroU32;
+
+use crate::capabilities;
 
 /// List of storage permissions for a storage user.
 ///
@@ -20,21 +23,26 @@ use core::cmp;
 /// we may want to revamp this interface.
 #[derive(Clone, Copy)]
 pub struct StoragePermissions {
-    // How many entries in the `read_permissions` slice are valid, starting at
-    // index 0.
+    /// How many entries in the `read_permissions` slice are valid, starting at
+    /// index 0.
     read_count: usize,
-    // Up to eight 32 bit identifiers of storage items the process has read
-    // access to.
+    /// Up to eight 32 bit identifiers of storage items the process has read
+    /// access to.
     read_permissions: [u32; 8],
-    // How many entries in the `write_permissions` slice are valid, starting at
-    // index 0.
+    /// How many entries in the `write_permissions` slice are valid, starting at
+    /// index 0.
     write_count: usize,
-    // Up to eight 32 bit identifiers of storage items the process has write
-    // (update) access to.
+    /// Up to eight 32 bit identifiers of storage items the process has write
+    /// (update) access to.
     write_permissions: [u32; 8],
-    // The identifier for this storage user when creating new objects. If `None`
-    // there is no `write_id` for these permissions.
-    write_id: Option<u32>,
+    /// The identifier for this storage user when creating new objects. If
+    /// `None` there is no `write_id` for these permissions.
+    write_id: Option<NonZeroU32>,
+    /// If `superuser` is true, this permission grants full access to all stored
+    /// items. New items created with `superuser == true` will use the specified
+    /// ID if  `write_id.is_some()`, otherwise new items will be created with
+    /// the reserved ID (i.e., 0).
+    superuser: bool,
 }
 
 impl StoragePermissions {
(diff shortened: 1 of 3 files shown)


bradjc:
I should have referenced #3457.

kernel/src/storage_permissions.rs
```diff
@@ -53,29 +61,76 @@ impl StoragePermissions {
...
     /// Check if this permission object grants read access to the specified
     /// `storage_id`. Returns `true` if access is permitted, `false` otherwise.
     pub fn check_read_permission(&self, storage_id: u32) -> bool {
-        self.read_permissions
-            .get(0..self.read_count)
-            .unwrap_or(&[])
-            .contains(&storage_id)
+        if self.superuser {
+            // Superuser grants all permissions.
+            true
```
alistair23, line 90: I have a few thoughts about this.

Obviously the kernel can do anything, it has direct access to the flash. But I think it would make more sense to allow apps to specify if the kernel can access their files. Accesses happen from capsules, which are only semi-trusted in Tock. That allows applications to apply some restrictions on what the kernel can/can't do.

I also worry about having a single branch instruction to allow/disallow access like this. Doesn't that expose us to glitch attacks? In saying that though, I'm not sure if anything else here is hardened against attacks like that.

I think it would be better to make the kernel ID `0x00` a user, the same as all other users. That way apps can choose to share data with the kernel or not.

alistair23 (changes requested):

kernel/src/storage_permissions.rs
bradjc: That seems reasonable to me. I think my current use case doesn't need superuser permissions, just having reserved kernel storage is sufficient. Updated.

kernel/src/storage_permissions.rs
```diff
@@ -20,21 +23,26 @@ use core::cmp;
...
-    // there is no `write_id` for these permissions.
-    write_id: Option<u32>,
+    /// The identifier for this storage user when creating new objects. If
+    /// `None` there is no `write_id` for these permissions.
+    write_id: Option<NonZeroU32>,
+    /// If `kerneluser` is true, this permission grants access to all objects
+    /// stored stored with `write_id` 0. New items created with `kerneluser ==
+    /// true` will use the specified ID if  `write_id.is_some()`, otherwise new
+    /// items will be created with the reserved ID (i.e., 0).
+    kerneluser: bool,
```
alistair23, line 45: Do we need this? Isn't the `write_id` Option enough

alistair23 (changes requested):

kernel/src/storage_permissions.rs
bradjc: `write_id.is_none()` means these permissions do not allow writes at all, so we wouldn't be able to differentiate "no writes" from "kernel writing".

libraries/tock-tbf/src/types.rs
```diff
@@ -227,7 +227,7 @@ pub struct TbfHeaderV2Permissions<const L: usize> {
 /// A list of persistent access permissions
 #[derive(Clone, Copy, Debug)]
 pub struct TbfHeaderV2PersistentAcl<const L: usize> {
-    write_id: u32,
+    write_id: Option<core::num::NonZeroU32>,
```
alistair23, line 230: Why convert this to an `Option` though?

bradjc: Because it expresses the property that `write_id==None` means the app does not have write access. It also forces the tbf library to explicitly handle that case.

alistair23 (approved):

lschuermann (approved):

PR #3471, opened 2023-06-07
Add license headers to the `.jlink` files.

jrvanwhy:
I tweaked the license checker's fallback parser so it accepts comments that start with `#` rather than `# `, which allows for a `#\n` line at the end of the header (as we allow for other file types).


bradjc (approved):

ppannuto (approved):
bors d+

Erring on the side of giving folks a few more days as this touches a bunch of license stuff, but seems like an easy merge

PR #3472, opened 2023-06-07
Add license headers to `.ld` files.

jrvanwhy:
I don't have time to dig through the history of all the linker scripts, so I followed the practice from #3317 and used `Copyright 2023`.

### DIFF
--- a/.lcignore
+++ b/.lcignore
@@ -34,7 +34,5 @@
 
 # File types that the license checker does not support but should support (see
 # issue #3417)
-*.ld
-!/tools/license-checker/testdata/block_comments.ld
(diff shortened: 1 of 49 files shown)


bradjc (dismissed):

ppannuto (approved):
bors d+

Same here — seems an easy merge, but give a bit more time as it touches licenses in a bunch of files.

w.r.t. linker provenance, I wrote most of the original linker scripts; the layout linker scripts are basically all copy/paste config files with different constants, so don't think there's much concern there either

bradjc (approved):

PR #3479, opened 2023-06-12
kernel: HIL: digest: add set_client() for `ClientDataHash` and `ClientDataVerify`

bradjc:
### Pull Request Overview

This pull request updates the digest HIL to add `set_client()` methods for two client types which were defined but could not be set in a capsule using the HIL. This makes it possible to implement a digest module which does 2 of the 3 operations (data, hash, verify).

But, like `Digest`, we can no longer apply these traits automatically. So, we have to explicitly add them (as in the sha256 capsule).

### Testing Strategy

Creating an HMAC-SHA256 capsule.

### TODO or Help Wanted

Is this the right way to go? I might be missing something. 

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/core/src/virtualizers/virtual_digest.rs
+++ b/capsules/core/src/virtualizers/virtual_digest.rs
@@ -148,6 +148,8 @@ impl<'a, A: digest::Digest<'a, L>, const L: usize> digest::DigestData<'a, L>
             self.mux.digest.clear_data()
         }
     }
+
+    fn set_data_client(&'a self, _client: &'a (dyn digest::ClientData<L> + 'a)) {}
 }
 
 impl<'a, A: digest::Digest<'a, L>, const L: usize> digest::DigestHash<'a, L>
@@ -175,6 +177,8 @@ impl<'a, A: digest::Digest<'a, L>, const L: usize> digest::DigestHash<'a, L>
             }
         }
     }
+
+    fn set_hash_client(&'a self, _client: &'a (dyn digest::ClientHash<L> + 'a)) {}
 }
 
 impl<'a, A: digest::Digest<'a, L>, const L: usize> digest::DigestVerify<'a, L>
@@ -200,6 +204,8 @@ impl<'a, A: digest::Digest<'a, L>, const L: usize> digest::DigestVerify<'a, L>
             }
         }
     }
+
+    fn set_verify_client(&'a self, _client: &'a (dyn digest::ClientVerify<L> + 'a)) {}
 }
 
 impl<'a, A: digest::Digest<'a, L>, const L: usize> digest::Digest<'a, L>
--- a/capsules/core/src/virtualizers/virtual_digest.rs
+++ b/capsules/core/src/virtualizers/virtual_digest.rs
@@ -149,7 +149,9 @@ impl<'a, A: digest::Digest<'a, L>, const L: usize> digest::DigestData<'a, L>
         }
     }
 
-    fn set_data_client(&'a self, _client: &'a (dyn digest::ClientData<L> + 'a)) {}
+    fn set_data_client(&'a self, _client: &'a (dyn digest::ClientData<L> + 'a)) {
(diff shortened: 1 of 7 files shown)


alistair23 (approved):

lschuermann (approved):

phil-levis (approved):

bradjc:
I ran into an issue while trying to support different clients for the different operations:

```
error[E0658]: cannot cast `dyn kernel::hil::digest::Client<32>` to `dyn ClientData<32>`, trait upcasting coercion is experimental
   --> chips/lowrisc/src/hmac.rs:381:30
    |
381 |         self.client_data.set(client);
    |                              ^^^^^^
    |
    = note: see issue #65991 <https://github.com/rust-lang/rust/issues/65991> for more information
    = help: add `#![feature(trait_upcasting)]` to the crate attributes to enable
    = note: required when coercing `&'a (dyn kernel::hil::digest::Client<32> + 'a)` into `&'a (dyn ClientData<32> + 'a)`

error[E0658]: cannot cast `dyn kernel::hil::digest::Client<32>` to `dyn ClientHash<32>`, trait upcasting coercion is experimental
   --> chips/lowrisc/src/hmac.rs:382:30
    |
382 |         self.client_hash.set(client);
    |                              ^^^^^^
    |
    = note: see issue #65991 <https://github.com/rust-lang/rust/issues/65991> for more information
    = help: add `#![feature(trait_upcasting)]` to the crate attributes to enable
    = note: required when coercing `&'a (dyn kernel::hil::digest::Client<32> + 'a)` into `&'a (dyn ClientHash<32> + 'a)`

error[E0658]: cannot cast `dyn kernel::hil::digest::Client<32>` to `dyn ClientVerify<32>`, trait upcasting coercion is experimental
   --> chips/lowrisc/src/hmac.rs:383:32
    |
383 |         self.client_verify.set(client);
    |                                ^^^^^^
    |
    = note: see issue #65991 <https://github.com/rust-lang/rust/issues/65991> for more information
    = help: add `#![feature(trait_upcasting)]` to the crate attributes to enable
    = note: required when coercing `&'a (dyn kernel::hil::digest::Client<32> + 'a)` into `&'a (dyn ClientVerify<32> + 'a)`
```

It seems like we actually cannot store the generic ("super") client three times. Thoughts?

bradjc:
@phil-levis I pushed my WIP changes but it doesn't seem like it will work.

We could change the `Digest::set_client()` to take three separate clients (and the others to take two) so that we aren't changing the types.

capsules/core/src/virtualizers/virtual_digest.rs
```diff
@@ -95,6 +95,10 @@ impl<'a, A: digest::Digest<'a, L>, const L: usize> VirtualMuxDigest<'a, A, L> {
 impl<'a, A: digest::Digest<'a, L>, const L: usize> digest::DigestData<'a, L>
     for VirtualMuxDigest<'a, A, L>
 {
+    fn set_data_client(&'a self, _client: &'a dyn ClientData<L>) {
+        unimplemented!()
```
alevy, line 99: Why is it ok for this to be `unimplemented`? Why would, for example, this being a `noop` not be better? (ditto for the other `unimplemented`)

bradjc: Well, we merged the `unimplemented!()` on line 211/223 of the same file, so, for consistency?

The challenge here is we are using the same trait twice, so our trick of putting the set_client in the HIL doesn't work (or at least I don't think it does, since I don't think we can implement `Digest::set_client()` twice, can we?).

phil-levis:
> It seems like we actually cannot store the generic ("super") client three times. Thoughts?

Ugh -- I'd forgotten this about Rust.

phil-levis (approved):

brghena (approved):

PR #3482, opened 2023-06-14
Add license headers to `*.xml` files.

### LINKED ISSUE #3417: Expand file type support in the license checker.
Linker scripts do not support line comments (`//`-style comments). The license checker should support linker scripts, and should therefore be adjusted to support block comments as well.

Before implementing, I'll check whether there are other file types in the repository that only support block comments and make sure the implementation works for those too.

jrvanwhy:
AFAIK, these are the last files in the repository that should have license headers but don't, so this PR should resolve #3417.

### DIFF
--- a/.lcignore
+++ b/.lcignore
@@ -31,8 +31,3 @@
 /tools/license-checker/testdata/many_errors.rs
 /tools/license-checker/testdata/no_copyright.rs
 /tools/license-checker/testdata/no_spdx.rs
-
-# File types that the license checker does not support but should support (see
-# issue #3417)
-*.xml
-!/tools/license-checker/testdata/block_comments.xml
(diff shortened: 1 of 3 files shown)


lschuermann (approved):

lschuermann:
Actually, looking at the XML I'm not even sure whether this isn't just copied from the MCUXpresso IDE, and thus it's unclear to me whether we can just assert those files to be licensed under our MIT/Apache-2.0 license. Perhaps @valexandru can help clear this up?

bradjc (approved):

jrvanwhy:
> Actually, looking at the XML I'm not even sure whether this isn't just copied from the MCUXpresso IDE, and thus it's unclear to me whether we can just assert those files to be licensed under our MIT/Apache-2.0 license.

This was discussed at today's core WG call, and the consensus was that the license headers should be merged as they are in this PR.

PR #3485, opened 2023-06-15
Add docs-ci workflow to build docs through GitHub actions

lschuermann:
### Pull Request Overview

GitHub's merge queues work by pushing to branches named `gh-readonly-queue/{target_branch}/{some_string}`. Then, pull requests will be merged once all required checks (as defined in the branch protection settings) have passed. However, our Netlify CI is only configured to build deploy previews for PRs (via a webhook), and build the production branch (master). This means that we can no longer mark the deploy previews as required checks, since they will not delivered for the aforementioned branches (only PRs). We could enable Netlify to build all branches, however this would significantly increase the builds carried out through Netlify, at least given how much development is done through branches on the main Tock repository.

A reasonable workaround is to simply build the documentation through GitHub actions itself. While we're not running in the exact same envrionment as Netlify's bots, our pinned rustup toolchain means at least rustdoc should behave identical. Note that this only applies for PRs enqueued in the merge queue; once merged into master the docs will still be built with Netlify. The deploy preview on PRs is also unaffected.

### Testing Strategy

This pull request was tested by running this on the `push` event on a fork.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.


.github/workflows/mergequeue_docs.yml
```diff
@@ -0,0 +1,34 @@
+# Licensed under the Apache License, Version 2.0 or the MIT License.
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Copyright Tock Contributors 2023.
+
+# Netlify's own CI builds both deploy previews for PRs, as well as the
+# production deploy for the master branch. We use this workflow purely as we
+# can't have Netlify build the wildcard gh-readonly-queue/* branches. This
+# workflow thus ensures that docs need to build (albeit not in the exact same
+# envrionment as Netlify's).
```
jrvanwhy, line 9: This is a sentence fragment. I think you meant to say "need to build do build" where you wrote "need to build".

jrvanwhy (approved):

.github/workflows/mergequeue_docs.yml
lschuermann: Ah, yes, I should've phrased this more clearly. What I meant to say was "[...] thus ensures that the docs build successfully".

jrvanwhy (approved):

bradjc (approved):
This looks great, and having docs which are few commits out of date is not too big of a deal.

PR #3489, opened 2023-06-15
capsules: tickv: call error cb on append error

bradjc:
### Pull Request Overview

In the tickv capsule we need to signal a callback even if the append fails.

### Testing Strategy

Writing to an existing key.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/extra/src/tickv.rs
+++ b/capsules/extra/src/tickv.rs
@@ -258,10 +258,6 @@ impl<'a, F: Flash, H: Hasher<'a, 8>, const PAGE_SIZE: usize> flash::Client<F>
             .replace(pagebuffer);
         let (ret, buf_buffer) = self.tickv.continue_operation();
 
-        buf_buffer.map(|buf| {
-            self.ret_buffer.replace(buf);
-        });
-
         match self.operation.get() {
             Operation::Init => match ret {
                 Ok(tickv::success_codes::SuccessCode::Complete)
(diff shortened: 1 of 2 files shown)


alevy (dismissed):

capsules/extra/src/tickv.rs
```diff
@@ -270,37 +266,71 @@ impl<'a, F: Flash, H: Hasher<'a, 8>, const PAGE_SIZE: usize> flash::Client<F>
...
-                    });
+            Operation::GetKey => {
+                buf_buffer.map(|buf| {
+                    self.ret_buffer.replace(buf);
+                });
+
+                match ret {
+                    Ok(tickv::success_codes::SuccessCode::Complete)
+                    | Ok(tickv::success_codes::SuccessCode::Written)
+                    | Err(tickv::error_codes::ErrorCode::BufferTooSmall(_)) => {
```
alistair23, line 277: Why do we signal this as a success to the caller?

bradjc: Because on delete our kv store layer only reads the header, not the entire value, so it only provides a smaller buffer.

bradjc: @alistair23

capsules/extra/src/tickv.rs
```diff
@@ -45,7 +45,7 @@ use kernel::utilities::leasable_buffer::LeasableMutableBuffer;
 use kernel::ErrorCode;
 use tickv::{self, AsyncTicKV};
 
-#[derive(Clone, Copy, PartialEq)]
+#[derive(Clone, Copy, PartialEq, Debug)]
```
alistair23, line 48: Why do we need this to be `Debug`?

capsules/extra/src/tickv.rs
alistair23: I don't follow. What delete?

alistair23 (changes requested):

capsules/extra/src/tickv.rs
bradjc: `kv_store::delete()`.

To remove an object we call get with a buffer large enough to read just the tock tickv header.

capsules/extra/src/tickv.rs
bradjc: I've done _a lot_ of debugging.

alistair23: Ha!

But it does add code size and it isn't being used by default

capsules/extra/src/tickv.rs
alistair23: But this is the `Operation::GetKey` operation, does it call delete?

bradjc: https://github.com/tock/tock/blob/c91004925ef9699dc962940f4e67701db7de18c6/capsules/extra/src/kv_store.rs#L354

Can you look at the kv store implementation?

alistair23: Shouldn't the kv_store implementation handle this error then? It knows if a delete operation is happening

bradjc: I mean, the code does what the code does. This PR is to change it. Currently this error gets returned as FAIL. What is the caller supposed to do?

Why are we going back and forth? You can just try this and see the issue, no?

alistair23: I still don't follow.

So in `kv_store` the `generate_key_complete()` function calls `get_value()` with a hashed key and a buffer to store the value.

That ends up triggering the `read_complete()` callback for a `Operation::GetKey` in `tickv.rs`. 

The `ret` is an `Err(tickv::error_codes::ErrorCode::BufferTooSmall(_))` which triggers the `get_value_complete()` callback.

But then `get_value_complete()` doesn't check `result` because it expects a failure. So what doesn't work? The PR and commit messages don't really explain what this is fixing

bradjc: How can kv_store not check `result`? What if some other error occurs?

alistair23: Yeah, I agree it should check result.

A small change like this:

```diff
diff --git a/capsules/extra/src/tickv.rs b/capsules/extra/src/tickv.rs
index d70e4e13a..fa91f5319 100644
--- a/capsules/extra/src/tickv.rs
+++ b/capsules/extra/src/tickv.rs
@@ -282,6 +282,16 @@ impl<'a, F: Flash, H: Hasher<'a, 8>, const PAGE_SIZE: usize> flash::Client<F>
                         );
                     });
                 }
+                Err(tickv::error_codes::ErrorCode::BufferTooSmall(_)) => {
+                    self.operation.set(Operation::None);
+                    self.client.map(|cb| {
+                        cb.get_value_complete(
+                            Err(ErrorCode::SIZE),
+                            self.key_buffer.take().unwrap(),
+                            self.ret_buffer.take().unwrap(),
+                        );
+                    });
+                }
                 Err(tickv::error_codes::ErrorCode::EraseNotReady(_)) | Ok(_) => {}
                 _ => {
                     self.operation.set(Operation::None);
```

and then `get_value_complete()` can check the result and ignore the `ErrorCode::SIZE` as expected

bradjc: Can you create a branch or suggested change?

alistair23: Something like this? https://github.com/alistair23/tock/commit/76714971943bbd4d46e3a980748e34b3ddca94fa

bradjc:
Added `SIZE` error when the retrieved value does not fit within the provided buffer.

capsules/extra/src/tickv.rs
```diff
@@ -270,37 +266,85 @@ impl<'a, F: Flash, H: Hasher<'a, 8>, const PAGE_SIZE: usize> flash::Client<F>
...
+            }
+            Operation::AppendKey => {
+                buf_buffer.map(|buf| {
+                    self.value_buffer.replace(Some(buf));
+                });
+
+                match ret {
+                    Ok(tickv::success_codes::SuccessCode::Complete)
+                    | Ok(tickv::success_codes::SuccessCode::Written) => {
+                        self.operation.set(Operation::None);
```
brghena, line 326: Appending doesn't need a callback?

capsules/extra/src/tickv.rs
```diff
@@ -270,37 +266,85 @@ impl<'a, F: Flash, H: Hasher<'a, 8>, const PAGE_SIZE: usize> flash::Client<F>
...
-                        cb.get_value_complete(
-                            Err(ErrorCode::FAIL),
-                            self.key_buffer.take().unwrap(),
-                            self.ret_buffer.take().unwrap(),
-                        );
-                    });
+            Operation::GetKey => {
+                buf_buffer.map(|buf| {
+                    self.ret_buffer.replace(buf);
+                });
```
brghena, line 272: Generally, the system of `continue_operation()` taking a buffer and then returning it, then this code needing to put that buffer right back, but you've got to make sure you put it in the right place, seems like a bug waiting to happen.

I'm satisfied that you did it correctly for GetKey and AppendKey (I think). And I'm not saying you should rewrite it now, but to me it feels _bad_.

brghena (approved):
I don't fully understand the TicKV design, but these changes all look valid to me. I've left a pair of comments to double-check.

alevy (approved):

capsules/extra/src/tickv.rs
hudson-ayers: I don't think it should add size in the case that it isn't being called

alistair23 (approved):

PR #3490, opened 2023-06-16
capsules: kv_store: implement virtualization

bradjc:
### Pull Request Overview

The existing kv_store capsule has a mux but doesn't do any virtualization. This re-writes the kv_store capsule to support virtualization and multiple users.

### Testing Strategy

I've tested with a single user, but not with multiple. Need to do that still.

### TODO or Help Wanted

I have a bunch of debugs to remove, but I wanted to get this out there.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/extra/src/kv_store.rs
+++ b/capsules/extra/src/kv_store.rs
@@ -366,9 +366,10 @@ impl<'a, K: KVSystem<'a, K = T>, T: kv_system::KeyType> kv_system::Client<T>
                                     }
                                     Err((key, value, e)) => {
                                         self.hashed_key.replace(key);
+                                        self.header_value.replace(value);
                                         node.operation.clear();
                                         node.client.map(move |cb| {
-                                            cb.get_complete(e, unhashed_key, value);
+                                            cb.delete_complete(e, unhashed_key);
                                         });
                                     }
                                 }
@@ -420,8 +421,6 @@ impl<'a, K: KVSystem<'a, K = T>, T: kv_system::KeyType> kv_system::Client<T>
                 match op {
                     Operation::Set => {}
                     Operation::Delete => {
-                        node.operation.clear();
-
                         let mut access_allowed = false;
 
                         if result.is_ok() {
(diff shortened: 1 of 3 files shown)


bradjc:
Now handles SIZE error when the retrieved value does not fit in the returned buffer. In the case of a delete this is intentional: we only provide a buffer to store the tock kv header. In the case of a normal get we now notify the upper layer that the value didn't fit.

capsules/extra/src/kv_store.rs
```diff
@@ -201,172 +169,229 @@ impl<'a, K: KVSystem<'a, K = T>, T: kv_system::KeyType> KVStore<'a, K, T> {
...
+        });
+
+        // If we have nothing scheduled, and we have recently done a delete, run
+        // a garbage collect.
+        if ret == Err(ErrorCode::NODEVICE) && self.cleanup.contains(&StateCleanup::CleanupRequested)
+        {
+            self.cleanup.set(StateCleanup::CleanupInProgress);
+            // We have no way to report this error, and even if we could, what
+            // would a user do?
+            let _ = self.kv.garbage_collect();
```
brghena, line 288: Would it make sense to instead have another member that tracks whether garbage collection has failed, and if so errors all further requested operations? That would make it visible to the user that something has broken rather than hiding it.

That will make the driver totally broken if garbage collection fails, but presumably it should never fail, and if it does the driver would likely be totally broken afterwards anyways, right?

brghena (dismissed):
This facially looks valid, and as long as we intend to "eat our own dogfood" and work with it, I think it's fine to move more quickly than usual and possibly break things.

capsules/extra/src/kv_store.rs
hudson-ayers: We sort of have this tradeoff in many Tock capsules, where there are rare underlying errors that could occur, and if they do occur, there is pretty much nothing a user can do to rectify it, and the capsule should probably no longer be used. Some capsules have a 'Error::BROKEN' in their state machine to account for this, but adding such a state comes at a size/complexity cost since you have to check against this possible error state in response to every system call. I am generally in favor of just dropping these errors for now, except in cases where we see people are actually hitting them in practice

hudson-ayers (dismissed):

hudson-ayers (approved):

capsules/extra/src/kv_store.rs
phil-levis: Is the idea that since this is an extra capsule, it shouldn't have a separate virtualizer?

bradjc: This capsule does two things: 1) add the tock header to the k-v objects, and 2) virtualize the kv_store interface.

We could split those two, but...maybe not right now.

PR #3493, opened 2023-06-16
capsules: kv driver: need to clear buffers

bradjc:
### Pull Request Overview

Because the entire buffer is used we need to make sure we clear it since it could have values in it from previous operations.

### Testing Strategy

Running an interactive key get/set/delete userspace app.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/extra/src/kv_driver.rs
+++ b/capsules/extra/src/kv_driver.rs
@@ -102,7 +102,9 @@ impl<'a, K: kv_system::KVSystem<'a, K = T>, T: kv_system::KeyType> KVSystemDrive
                                                 let static_buffer_len =
                                                     buf.len().min(unhashed_key.len());
 
-                                                // Copy the data into the static buffer
+                                                // Copy the data into the
+                                                // cleared static buffer.
+                                                buf.fill(0);
                                                 unhashed_key[..static_buffer_len]
                                                     .copy_to_slice(&mut buf[..static_buffer_len]);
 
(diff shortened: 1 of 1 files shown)


capsules/extra/src/kv_driver.rs
```diff
@@ -102,7 +102,9 @@ impl<'a, K: kv_system::KVSystem<'a, K = T>, T: kv_system::KeyType> KVSystemDrive
                                                 let static_buffer_len =
                                                     buf.len().min(unhashed_key.len());
 
-                                                // Copy the data into the static buffer
+                                                // Copy the data into the
+                                                // cleared static buffer.
+                                                buf.fill(0);
```
alevy, line 107: I assume we expect these to be pretty small such that filling in slots that we're about to write over in the next line is the same or faster (rather than _only_ clearing the left over after the copy) because of something-something vectorization.

alevy, line 107:
```suggestion
                                                buf[static_buffer_len..].fill(0);
```

bradjc: Yeah I have them set at 64 bytes.

Really I think the fix is to add a length parameter so we aren't using the entire buffer unconditionally. https://github.com/tock/tock/issues/3495

alevy (approved):

capsules/extra/src/kv_driver.rs
alevy: fine with me

PR #3494, opened 2023-06-16
kv: use only 64 byte buffers

bradjc:
### Pull Request Overview

The buffer size affects the key hash, so we need to standardize on something. I've been using 64 bytes.

### Testing Strategy

My interactive KV app.

### TODO or Help Wanted

It's probably better to _not_ make the hash dependent on the buffer size. I'm not sure why it's implemented this way.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alevy (approved):

alistair23 (approved):

bradjc:
I'm adding support for keeping track of sizes so we do not need this.

PR #3502, opened 2023-06-22
Add hmac-sha256 software capsule

bradjc:
### Pull Request Overview

This pull request adds an HMAC-sha256 capsule and a simple test.

### Testing Strategy

Running the test on the nrf52840dk.

### TODO or Help Wanted

This requires the update to the digest trait.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


ppannuto (approved):
I believe this all looks correct.

Long term, the near-duplication of implementation (e.g. `add_data` vs `add_mut_data`) up and down the security/crypto ecosystem based on whether it's an in-place or two buffer feels fragile, but I don't have a better thought at the moment.

PR #3503, opened 2023-06-22
capsules: remove virtual_[digest|flash|sha].rs which do not virtualize

bradjc:
### Pull Request Overview

These capsules do not actually provide virtualized digest resources, and I think it's better to not give the impression that we support virtualized digest resources when we do not currently.

### Testing Strategy

n/a

### TODO or Help Wanted

Thoughts? Obviously another option is to re-write the virtualizers, but someone needs to be willing to do that.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.


lschuermann (approved):
We can always resurrect these from git.

hudson-ayers:
Wait, so like...what do these do now?

bradjc:
> Wait, so like...what do these do now?

Nothing, really. They insert a little code when a HIL function is called, but for the most part the code isn't used. This is because the callbacks are not handled by these capsules, but instead passed directly to the upper layer.

alevy (approved):

lschuermann (approved):

hudson-ayers (approved):

hudson-ayers:
This still breaks CI

bradjc:
Yeah if there is consensus to doing this then I can remove their uses.

hudson-ayers:
I am on board

phil-levis (approved):

PR #3505, opened 2023-06-24
kernel: leasablebuffer: add APIs

bradjc:
### Pull Request Overview

This pull request updates `LeasableBuffer` to add:

- `as_slice()`: gets a slice
- `is_sliced()`: whether .slice has been called
- ~~`unslice_left()`: open up room to the left~~

### Testing Strategy

Updating k-v stack.

### TODO or Help Wanted

This doesn't have to be the last PR that adds APIs, but I would like to not stall on this because these are the changes needed to update the K-V stack.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/utilities/leasable_buffer.rs
+++ b/kernel/src/utilities/leasable_buffer.rs
@@ -179,25 +179,6 @@ impl<'a, T> LeasableMutableBuffer<'a, T> {
             end: new_end,
         };
     }
-
-    /// Increase the range of the LeasableBuffer that is accessible to the left
-    /// of the starting point of the currently available range.
-    ///
-    /// This is particularly useful to insert a header before the current
-    /// contents of the buffer.
-    ///
-    /// This returns `Ok(())` if there is room to increase the range the
-    /// requested number of elements to the left. This returns `Err(())` if
-    /// there is not enough room, and the accessible range remains unchanged.
-    pub fn unslice_left(&mut self, left: usize) -> Result<(), ()> {
-        let new_start = self.active_range.start.checked_sub(left).ok_or(())?;
-
-        self.active_range = Range {
-            start: new_start,
-            end: self.active_range.end,
-        };
-        Ok(())
-    }
 }
 
 impl<'a, T, I> Index<I> for LeasableMutableBuffer<'a, T>
@@ -298,25 +279,6 @@ impl<'a, T> LeasableBuffer<'a, T> {
             start: new_start,
             end: new_end,
         };
-    }
-
-    /// Increase the range of the LeasableBuffer that is accessible to the left
-    /// of the starting point of the currently available range.
-    ///
-    /// This is particularly useful to insert a header before the current
-    /// contents of the buffer.
-    ///
-    /// This returns `Ok(())` if there is room to increase the range the
-    /// requested number of elements to the left. This returns `Err(())` if
-    /// there is not enough room, and the accessible range remains unchanged.
-    pub fn unslice_left(&mut self, left: usize) -> Result<(), ()> {
-        let new_start = self.active_range.start.checked_sub(left).ok_or(())?;
-
-        self.active_range = Range {
-            start: new_start,
-            end: self.active_range.end,
-        };
-        Ok(())
     }
 }
 
--- a/kernel/src/utilities/leasable_buffer.rs
+++ b/kernel/src/utilities/leasable_buffer.rs
@@ -137,6 +137,22 @@ impl<'a, T> LeasableMutableBuffer<'a, T> {
         self.active_slice().as_ptr()
     }
 
+    /// Returns a slice of the currently accessible portion of the
+    /// LeasableBuffer.
+    pub fn as_slice(&mut self) -> &mut [T] {
+        &mut self.internal[self.active_range.clone()]
+    }
+
+    /// Returns `true` if the LeasableBuffer is sliced internally.
+    ///
+    /// This is a useful check when switching between code that uses
+    /// LeasableBuffers and code that uses traditional slice-and-length. Since
+    /// slice-and-length _only_ supports using the entire buffer it is not valid
+    /// to try to use a sliced LeasableBuffer.
+    pub fn is_sliced(&self) -> bool {
+        self.internal.len() != self.len()
+    }
+
     /// Reduces the range of the LeasableBuffer that is accessible. This should be called
     /// whenever an upper layer wishes to pass only a portion of a larger buffer down to
     /// a lower layer. For example: if the application layer has a 1500 byte packet
@@ -162,6 +178,25 @@ impl<'a, T> LeasableMutableBuffer<'a, T> {
             start: new_start,
             end: new_end,
         };
+    }
+
+    /// Increase the range of the LeasableBuffer that is accessible to the left
+    /// of the starting point of the currently available range.
+    ///
+    /// This is particularly useful to insert a header before the current
+    /// contents of the buffer.
+    ///
+    /// This returns `Ok(())` if there is room to increase the range the
+    /// requested number of elements to the left. This returns `Err(())` if
+    /// there is not enough room, and the accessible range remains unchanged.
+    pub fn unslice_left(&mut self, left: usize) -> Result<(), ()> {
+        let new_start = self.active_range.start.checked_sub(left).ok_or(())?;
+
+        self.active_range = Range {
+            start: new_start,
+            end: self.active_range.end,
+        };
+        Ok(())
     }
 }
 
@@ -222,6 +257,22 @@ impl<'a, T> LeasableBuffer<'a, T> {
         self.active_slice().as_ptr()
     }
 
+    /// Returns a slice of the currently accessible portion of the
+    /// LeasableBuffer.
+    pub fn as_slice(&self) -> &[T] {
+        &self.internal[self.active_range.clone()]
+    }
+
+    /// Returns `true` if the LeasableBuffer is sliced internally.
+    ///
+    /// This is a useful check when switching between code that uses
+    /// LeasableBuffers and code that uses traditional slice-and-length. Since
+    /// slice-and-length _only_ supports using the entire buffer it is not valid
+    /// to try to use a sliced LeasableBuffer.
+    pub fn is_sliced(&self) -> bool {
+        self.internal.len() != self.len()
+    }
+
     /// Reduces the range of the LeasableBuffer that is accessible. This should be called
     /// whenever an upper layer wishes to pass only a portion of a larger buffer down to
     /// a lower layer. For example: if the application layer has a 1500 byte packet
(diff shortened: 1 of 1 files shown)


kernel/src/utilities/leasable_buffer.rs
```diff
@@ -163,6 +179,25 @@ impl<'a, T> LeasableMutableBuffer<'a, T> {
...
+    /// contents of the buffer.
+    ///
+    /// This returns `Ok(())` if there is room to increase the range the
+    /// requested number of elements to the left. This returns `Err(())` if
+    /// there is not enough room, and the accessible range remains unchanged.
+    pub fn unslice_left(&mut self, left: usize) -> Result<(), ()> {
+        let new_start = self.active_range.start.checked_sub(left).ok_or(())?;
+
+        self.active_range = Range {
+            start: new_start,
```
alistair23, line 196: This seems dangerous doesn't it?

Isn't the point of `LeasableBuffer` that we can pass a portion of the buffer to lower levels and trust that they aren't changing any data outside of the active range. This seems to encourage lower levels to change data outside of the range.

What if a lower level calls this but the upper level doesn't expect it to? Won't it then override data?

alistair23 (changes requested):

kernel/src/utilities/leasable_buffer.rs
bradjc: What if it calls take() (like siphash did)? Or reset()?

You can see my use case here: https://github.com/tock/tock/pull/3508/files#diff-d4c838ce4fe52e989c391df3c23c81855d0eb3396656f4ca281a609a89c02391R215

This saves us from having to move bytes in the buffer to make room for a header.

alistair23: I agree that currently nothing enforces it. I thought the idea with `take()` is that the lower layers still used the length (but that would break with a non zero start offset). `reset()` I thought was never to be called by lower layers.

I understand that it avoids copying data within the buffer, but it seems like a risky function to implement as then all lower layers can use it and there isn't a way to determine if we should or shouldn't safely call it

bradjc: Why is it more risky than expecting that the provided buffer has room to the right?

Either we unslice to the left, or move bytes to the right then fill in the vacated space. Either way we make assumptions about the buffer passed in. If there isn't room we return an error in either case.

But, I'm arguing that this is the API we _want_. Rather than implicitly expecting buffers to be large, we explicitly direct callers to leave space so we don't have to move bytes around.

alistair23: But we know the length of the buffer. When a lower layer gets passed a `LeasableBuffer` it has a `start` and an `end` offset. So we know for sure that when we move data it is inside that buffer, there aren't any assumptions.

How do we direct callers to leave space at the start? Just a comment in the function that we hope they read? That seems not a very Rust-y way to do this

bradjc: We _don't_ know there is room. We can check at runtime, but we don't know at compile time.

We can call unslice left, at runtime, and either it succeeds, or it doesn't.

Either way, we check at runtime. What makes hoping that the caller happened to allocate a large enough buffer more rusty?

alistair23: How about an example. There is capsule A which creates a `LeasableBuffer` and passes it to a HAL implementation B.

Capsule A creates a buffer of 64 bytes. It stores some data in byte 0 to 8 buffer[0..7], which it expects to not be changed. It creates a `LeasableBuffer::new(&buffer[8..63])` and passes that to the HAL implementation B.

The implementation B can then copy its data into the `LeasableBuffer`, without modifying the data that was already stored there. You are right that this is a run time check, but we know at run time that we have a slice `&buffer[8..63]`.

The problem with `unslice_left()` is that capsule A might expect the data not to be modified, while the HAL implementation B might modify it. I thought the entire point of `LeasableBuffer` was to pass slices to lower layers. If we start modifying the slice in the lower layers then we loose the contract.

Obviously lower layers can call `reset()` or `take()`, but they shouldn't. `unslice_left()` on the other hand encourages lower layers to modify the slice. It's something that can't be enforced (at either runtime or compile time) which makes it not Rust-y.

bradjc: Unslice left lets us append headers without copying. If you want to write a malicious capsule, call reset and you're done.

If A is adamant that B must absolutely be unable to modify the first 8 bytes under any circumstances, then it needs to pass a different buffer to B.

If we don't want to support prepending headers, that's a design choice, but adding unslice left isn't changing the guaranteed safety of using leasable buffers.

alistair23: We aren't defending against something malicious though. `LeasableBuffer` will be created by a a capsule (something we reasonably trust) and passed to a HAL implementation. The HAL implementation is completely trusted. It has access to the hardware and unsafe code, if it's malicious it can do anything anyway.

We are trying to protect against programming errors. Maybe a SPI capsule wants to call `spi_read_write()` twice, each time saving data to a different place in a single buffer. That type of situation is a good use case for `LeasableBuffer` and we don't want the lower layer editing our slices.

I agree, as I said earlier that `unslice_left()` does not change the guaranteed safety. What it does provide is an API to modify the slice and the caller has no way to know if the data outside the active range has been modified. That is the risk. Once we start modifying `LeasableBuffer` active ranges in the lower layers what is the use for it?

kernel/src/utilities/leasable_buffer.rs
```diff
@@ -163,6 +179,25 @@ impl<'a, T> LeasableMutableBuffer<'a, T> {
...
+    /// Increase the range of the LeasableBuffer that is accessible to the left
+    /// of the starting point of the currently available range.
+    ///
+    /// This is particularly useful to insert a header before the current
+    /// contents of the buffer.
+    ///
+    /// This returns `Ok(())` if there is room to increase the range the
+    /// requested number of elements to the left. This returns `Err(())` if
+    /// there is not enough room, and the accessible range remains unchanged.
+    pub fn unslice_left(&mut self, left: usize) -> Result<(), ()> {
```
alevy, line 192: What about a more general `extend(&mut self, left: usize, right: usize) -> Result<(), ()>`.

kernel/src/utilities/leasable_buffer.rs
alevy: @alistair23 I don't think I agree that the purpose of LeasableBuffer is to protect against errors (though, that's of course always nice).

The main purpose is to systemetize the pattern that capsules often end up needing to pass around complete `&'static mut [u8]` along with an offset and length into that slice _because_ Tock is generally asynchronous and the caller can only get back ownership of the slice via a callback later (as opposed to leveraging borrows in synchronous code in "normal Rust").

So, the point is to pass around a slice-like thing that internalizes an offset and length and, by default, only exposes the elements in that range, but is, inherently, able to restore the entire buffer or any other subportion of the complete buffer.

alistair23: Yeah, I agree that `LeasableBuffer` is used to pass a slice-like thing.

I always pictured something like this

Say we had a function that mutates a slice, in non-Tock code we would have a function like this

```rust
/// Non-Tock code
/// Does something magic to a slice passed in
fn my_function(&self, buffer: &mut [u8]) {
...
}

my_function(&mut my_buffer[2..16]);
```

That falls apart with Tock's static life times, so the Tock function instead would be:


```rust
/// Tock code
/// Does something magic to a slice passed in
fn my_function(&self, buffer: &'static mut [u8], offset: usize, len: usize) {
...
}

my_function(my_buffer, 2, 16);
```

In both these cases though I expect `my_function()` to only change the data inside the slice. In the first case Rust enforces that `my_function()` can only access `2..16` while in the second case there is no enforcement, but we expect `my_function()` to use the `offset` and `len` arguments.

I always saw `LeasableBuffer` as just a wrapper for the above, which is why I think we shouldn't encourage lower layers to edit parts of the slice

hudson-ayers: I agree with Alistair on this. If there was an easy way to statically enforce that `LeasableBuffer` only provides access to the subsliced range once it is created, I would have implemented that. Unfortunately there is not an efficient way to statically enforce this, but I still generally consider it a code smell when any component calls take() or reset() that is not the component that originally created the `LeasableBuffer`. I do not think we should make this pattern easier. If a higher layer expects a lower layer to prepend headers, then it should provide more room in the passed buffer, or the lower layer should use a copy, imo

bradjc: Seems like the question is: is LB a _protection_ mechanism, designed to enforce `my_slice[2..16]` like semantics, or a _convenience_ mechanism, designed to help tock programmers encapsulate buffer length, slice length, and offset?

Personally, I don't think we need a protection mechanism. I don't think we have any use cases for wanting to hide data in buffers passed to lower layers that we then expect to get back. I'm not even sure we make any guarantees you would get the _same_ buffer back (for example if something is double buffered, you might get a different static buffer back). I also don't think we care if a lower layer decides to use the remainder of a buffer for its own purposes before passing the buffer back. Put another way, if we could somehow statically enforce access only to the subslice, I don't see how an upper layer would be written in any different way to take advantage of that guarantee.

bradjc: But, if we do want to preserve the ability to make LB a protection mechanism, what would it look like to expand the type somehow to support both uses? Could we add an API to allow the creator to specify that extending/unslicing is ok? Or is that compromising the usability of the type?

hudson-ayers: IMO LB is a convenience mechanism to emulate slices, that we are forced to use because no_heap Rust requires us to reuse static buffers, and static lifetimes are not compatible with passing slices. If we could pass slices of a static buffer and then reassemble the static buffer after a callback LB would not exist. So I think LB should try to emulate the slice API as closely as possible, for both the convenience and protection that provides, except where we are forced to make concessions so that it is usable in practice.

I think if we design LB such that people are encouraged to use it differently than they would a slice that will lead to programming mistakes due to mismatched expectations. Maybe I am wrong though, as you and Amit clearly do not see LB as a slice replacement as much as I do

phil-levis: I don't think SubSlice can be an enforced protection mechanism; once you move to an asynchronous model, you can't be assured of a linear call/return chain that allows you to telescope out of reducing access scopes. Put another way, it can only work as a protection mechanism if you assume that SubSlices always return to their callers (which they might not), and that would require storing some state for each caller. Put another way, the bottom of the call chain does in the end need access to the complete buffer because they might need to pass it off to someone else. 

That being said, following Hudson's point, it is intended that they are *often* used kind of like slices. They're a way to easily pass the state of a sub-slice down. So making them look like slices as much as possible is good, as long as we don't break assumptions/lead to confused use.

bradjc: While I don't agree that SubSlice _should_ emulate what LB would look like if Rust gave us a native way to do what we want (as Hudson explained well), I do understand that mimicking what a rust developer would reasonably _expect_ a "sub" slice type to behave like has quite a bit of value.

If I'm extending this discussion just slightly, what I'm hearing is that if it is going to be called `SubSlice` (or some other riff on "slice") then it should act like a normal slice to the extent possible (and we just need to tell people not to call take). It's not that we expect to enforce this with the compiler, just that we want to match developer expectations.

But, I'm also hearing that it's not that we _shouldn't_ provide a type that exposes unslice or extend, it's just that would be a different type, with a name that reflects its purpose, if it were to exist.

bradjc:
I removed `unslice_left()` to punt that to potentially a different discussion.

alistair23 (approved):

brghena (approved):

PR #3507, opened 2023-06-24
TicKV: Keep track of the value length, reverse async op return

bradjc:
### Pull Request Overview

This pull request makes two changes to TicKV. First, it updates TicKV to change the `get_key()` API to:

```
pub fn get_key(&self, hash: u64, buf: &mut [u8]) -> Result<(SuccessCode, usize), ErrorCode>;
```

so that the caller knows the size of the value that was retrieved.

Second, it changes the async_ops interface to return `Ok()` when things are ok, and `Err()` when something is actually wrong. TicKV uses errors to allow a synchronous interface to also support synchronous operation. But once we are in the async interface, it doesn't make sense to continue to report correct values as errors.

### Testing Strategy

Running tickv in the kernel.

### TODO or Help Wanted

This obviously breaks the tickv capsule, so there are more PRs to come. But I'm trying to isolate changes for review.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/libraries/tickv/src/async_ops.rs
+++ b/libraries/tickv/src/async_ops.rs
@@ -105,7 +105,7 @@
 //!
 //! // Add a key
 //! static mut VALUE: [u8; 32] = [0x23; 32];
-//! let ret = unsafe { tickv.append_key(get_hashed_key(b"ONE"), &mut VALUE) };
+//! let ret = unsafe { tickv.append_key(get_hashed_key(b"ONE"), &mut VALUE, 32) };
 //!
 //! match ret {
 //!     Err((_buf, ErrorCode::ReadNotReady(reg))) => {
@@ -139,6 +139,8 @@ type ContinueReturn = (
     Result<SuccessCode, ErrorCode>,
     // Buf Buffer
     Option<&'static mut [u8]>,
+    // Length of valid data inside of the buffer.
+    usize,
 );
 
 /// The struct storing all of the TicKV information for the async implementation.
@@ -147,7 +149,7 @@ pub struct AsyncTicKV<'a, C: FlashController<S>, const S: usize> {
     pub tickv: TicKV<'a, C, S>,
     key: Cell<Option<u64>>,
     value: Cell<Option<&'static mut [u8]>>,
-    buf: Cell<Option<&'static mut [u8]>>,
+    value_length: Cell<usize>,
 }
 
 impl<'a, C: FlashController<S>, const S: usize> AsyncTicKV<'a, C, S> {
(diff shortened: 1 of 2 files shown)


libraries/tickv/src/async_ops.rs
```diff
@@ -147,7 +149,7 @@ pub struct AsyncTicKV<'a, C: FlashController<S>, const S: usize> {
     pub tickv: TicKV<'a, C, S>,
     key: Cell<Option<u64>>,
     value: Cell<Option<&'static mut [u8]>>,
-    buf: Cell<Option<&'static mut [u8]>>,
```
alistair23, line 150: This fails to build, there are still users of this

alistair23 (changes requested):

libraries/tickv/src/async_ops.rs
bradjc: Yeah I'm trying to balance reviewable PRs with reasonable programming overhead.

#3508 if you want to see the full changes that pass CI.

libraries/tickv/src/async_ops.rs
```diff
@@ -193,18 +195,30 @@ impl<'a, C: FlashController<S>, const S: usize> AsyncTicKV<'a, C, S> {
...
-            Ok(code) => Ok(code),
+        length: usize,
+    ) -> Result<SuccessCode, (&'static mut [u8], ErrorCode)> {
+        match self.tickv.append_key(hash, &value[0..length]) {
+            Ok(_code) => {
+                // Ok is a problem, since that means no asynchronous operations
+                // were called, which means our client will never get a
+                // callback. We need to error.
+                Err((value, ErrorCode::WriteFail))
+            }
```
hudson-ayers, line 206: This seems better than how things worked before, though I do think we might eventually want to support an API that can sit on top of maybe-async operations, which would rely on this library actually scheduling a callback in the case that the underlying `append_key()` happened synchronously. I don't want to block this on that though, as I think getting these changes in so things are in a workable state is important.

hudson-ayers (approved):

libraries/tickv/src/async_ops.rs
bradjc: Hmm, that makes sense. That does get fairly complicated, however. Since the caller has to implement both the lower layer (flash) and upper layer (in our case tickv capsule), it would be strange for the caller to both implement a synchronous flash but then try to use the async API.

hudson-ayers: Makes sense. Lets stick with this for now and can change it if/when a use case comes along

bradjc:
This PR will never pass CI, so again, it will have to be merged in the full PR (#3508), but you can see this change more clearly here.

bradjc:
Merged in #3508.

PR #3509, opened 2023-06-24
kernel: process loading: update comments & remove capabilities

bradjc:
### Pull Request Overview

In process loading we don't need capabilities for non-pub functions.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process_loading.rs
+++ b/kernel/src/process_loading.rs
@@ -176,6 +176,11 @@ impl fmt::Debug for ProcessLoadError {
 /// This function is made `pub` so that board files can use it, but loading
 /// processes from slices of flash an memory is fundamentally unsafe. Therefore,
 /// we require the `ProcessManagementCapability` to call this function.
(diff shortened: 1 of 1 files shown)


lschuermann (approved):
I guess I agree with these changes, but I feel like it's going to be an easy mistake to change these functions to be `pub`, but not take a capability. I wouldn't have an issue passing the capability onward to these internal functions, given that those are the actual process loading implementations for which the capability is actually relevant.

bradjc:
> I guess I agree with these changes, but I feel like it's going to be an easy mistake to change these functions to be `pub`, but not take a capability. I wouldn't have an issue passing the capability onward to these internal functions, given that those are the actual process loading implementations for which the capability is actually relevant.

I think this is true of numerous functions in the kernel crate. I think we should keep the use case of capabilities narrow and not speculative.

hudson-ayers (approved):

PR #3511, opened 2023-06-26
kernel: LeasableBuffer: fix double slice end value

bradjc:
### Pull Request Overview

This pull request fixes the `slice()` operation in LeasableBuffer when .slice() is called twice. In the second .slice(), if the end was `..` (ie `mybuf.slice(4..)`), the buffer would go back to its full length rather than just the part originally visible.

The fix is just to use the saved end as the default (which is set in `new()` so it works for the first slice as well).

### Testing Strategy

KV stack work.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


alistair23 (approved):

brghena (approved):

brghena:
Yes, this is definitely the fix

PR #3499, opened 2023-06-21
kernel: process: do not inline load_*processes() functions

bradjc:
### Pull Request Overview

This causes the stack size of the main function to be quite large, and since main does not return its stack frame is not recovered.

Related to #2425.

With this patch:

```
❯ make allstack
hifive_inventor
----------------------
   main stack frame:
         1408     main

   5 largest stack frames:
         1648     _ZN6kernel15process_loading14load_processes17h4086cff94706f9b2E
         1568     _ZN95_$LT$kernel..process_standard..ProcessStandard$LT$C$GT$$u20$as$u20$kernel..process..Process$GT$18print_full_process17he131d7be106924b4E
         1408     main
          976     _ZN13capsules_core15process_console31ProcessConsole$LT$_$C$A$C$C$GT$11write_state17hbf87fa67f6d6c55cE
          960     _ZN6kernel5debug11panic_print17h54911765252f592eE

nrf52dk
----------------------
   main stack frame:
          128     main

   5 largest stack frames:
         2408     _ZN7cortexm24kernel_hardfault_arm_v7m17h88793f49ade3c395E
         1728     _ZN6kernel15process_loading14load_processes17h6134321b823806bbE
          960     _ZN13capsules_core15process_console31ProcessConsole$LT$_$C$A$C$C$GT$11write_state17h12a38ccfec0ca229E
          832     _ZN95_$LT$kernel..process_standard..ProcessStandard$LT$C$GT$$u20$as$u20$kernel..process..Process$GT$18print_full_process17had5cd228b8bd15a2E
          672     _ZN117_$LT$capsules_core..process_console..ProcessConsole$LT$_$C$A$C$C$GT$$u20$as$u20$kernel..hil..uart..TransmitClient$GT$18transmitted_buffer17h5706731d0f1a1db8E

nrf52840_dongle
----------------------
   main stack frame:
          160     main

   5 largest stack frames:
         2408     _ZN7cortexm24kernel_hardfault_arm_v7m17h88793f49ade3c395E
         1728     _ZN6kernel15process_loading14load_processes17hb5eabb632f44b429E
          960     _ZN13capsules_core15process_console31ProcessConsole$LT$_$C$A$C$C$GT$11write_state17h00e75e951ec52fb3E
          832     _ZN95_$LT$kernel..process_standard..ProcessStandard$LT$C$GT$$u20$as$u20$kernel..process..Process$GT$18print_full_process17h0eb0f6766ab104a8E
[...]

PR #3514, opened 2023-06-26
boards: microbitv2: use helper function in main

bradjc:
### Pull Request Overview

By outsourcing all of `main()` to a separate function (in this case `start()` other than running the kernel, we can minimize the stack frame size of main. This is important because the main stack frame is never popped since main does not return.

This is an alternative to doing this piecemeal for individual functions.

This PR is an alternative to #3499.

### Testing Strategy

travis

### TODO or Help Wanted

I only did this for one board. Thoughts? Naming? Is using `Chip.unwrap()` ok / is there a better option?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/microbit_v2/src/main.rs
+++ b/boards/microbit_v2/src/main.rs
@@ -220,22 +220,13 @@ impl KernelResources<nrf52833::chip::NRF52<'static, Nrf52833DefaultPeripherals<'
 /// removed when this function returns. Otherwise, the stack space used for
 /// these static_inits is wasted.
 #[inline(never)]
-unsafe fn create_peripherals() -> &'static mut Nrf52833DefaultPeripherals<'static> {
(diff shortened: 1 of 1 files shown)


hudson-ayers:
I think I would just return `chip` from `start()` so it can be used in `main()`, since I am not sure that Rust will optimize out that panicking branch with inlining prevented.

hudson-ayers:
This PR actually seems to reduce code size for the microbit, and gives us a much smaller main() stack frame (and worst-case stack depth, I am sure). I much prefer this to #3499, and advocate using it for all boards

bradjc:
Maybe `setup()` is a better name?

hudson-ayers (approved):

ppannuto (approved):
~bors r+~ jk

PR #3515, opened 2023-06-26
Redefine TBF `protected_size` field to a `protected_trailer_size`

lschuermann:
### Pull Request Overview

In practice, all TBFs created with `elf2tab` have set their TBF header `protected_size` field to indicate the size of the protected region, *excluding* the actual TBF headers. This inconsistency leads to confusion when inspecting generated TBF headers against their specification.

The `tock-tbf` library in the kernel contains a workaround for this issue by adding the TBF headers size to the `protected_size` header field and returning this to higher-level layers (such as the process loading infrastructure in `process_standard.rs`) as the "protected size".

Including the TBF headers in the protected region makes sense semantically. However, at this point it seems infeasible to change the kernel & elf2tab behavior. As there is nothing wrong with the behavior itself, apart from not matching its specification, this change proposes to update the specification instead: we redefine the `protected_size` header field to specify a `protected_trailer_size`, which specifies the number of bytes between the end of the TBF headers and the end of the protected region size.

This change updates the `tock-tbf` library accordingly. By keeping the (specification-conforming) `get_protected_size` method this change should not have any user-visible impact. Ultimately, `elf2tab` should be updated to refer to the new field name as well. However, given that `elf2tab` already corrects for this error internally, no user-visible change should be required on that front either.

### Testing Strategy

N/A

### TODO or Help Wanted

The name could be bikeshedded. It might be confusing to call it a "trailer", as it's really only a trailer for the protected region itself, but not the overall TBF file. Then again, calling it "padding" is confusing, because the binary itself (following this protected region) could require padding, so we'd have two very different types of padding sections adjacent to each other.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/TockBinaryFormat.md
+++ b/doc/TockBinaryFormat.md
@@ -41,7 +41,8 @@ Tock App Binary:
 Start of app -> +-------------------+---
               ^ | TBF Header        | ^
               | +-------------------+ | Protected region
-              | | Optional padding  | V
+              | | (Optional)        | |
+              | | protected trailer | V
  Covered by   | +-------------------+---
  integrity    | | Userspace Binary  |
               | |                   |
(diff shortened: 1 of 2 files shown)


bradjc (approved):
I agree that it is time to cut our losses and change the definition of `protected_size` field in the headers to how it is used in practice.

bradjc (approved):

alistair23 (approved):

brghena (approved):

Issue #3518, opened 2023-06-27
Tracking: Use `start()` function in board main.rs


bradjc:
Since the main() function never returns, its stack frame is never reclaimed. If that stack frame is large, that results in significant stack memory waste.

We tried addressing this in #2425. However, in retrospect that was a piecemeal approach, where at the time peripheral creation was the only thing causing main's stack frame to be large. Since then, we've realized other operations cause a large stack frame (for example inlining the load_processes function #3499).

We should just avoid this once-and-for-all by moving everything that is currently in `main()` into a new function called `start()`, except for the call to `kernel::kernel_loop()`.

Boards to be updated:
- [x] acd52832 #3766
- [x] apollo3/lora_things_plus
- [x] apollo3/redboard_artemis_nano
- [x] arty_e21 #3765
- [x] clue_nrf52840 #3767
- [x] esp32-c3-devkitM-1
- [x] hail #3678 
- [x] hifive1 #3768
- [x] hifive_inventor #3768
- [x] imix #4085
- [x] imxrt1050-evkb #4085
- [x] litex/arty #4084
- [x] litex/sim #4084
- [x] microbit_v2 #3514
- [x] msp_exp432p401r #4085
- [x] nano33ble #3602 
- [x] nano_rp2040_connect #3914
- [x] nordic/nrf52840_dongle #3836
- [x] nordic/nrf52840dk #3836
- [x] nordic/nrf52dk #3836
- [x] nucleo_f429zi #3860
- [x] nucleo_f446re #3860
- [x] opentitan/earlgrey-cw310
- [x] particle_boron #3733
- [x] pico_explorer_base #3914
- [x] qemu_rv32_virt #4085
- [x] raspberry_pi_pico #3914
- [x] redboard_redv #4085
- [x] sma_q3 #4085
- [x] stm32f3discovery #4083
- [x] stm32f412gdiscovery #4083
- [x] stm32f429idiscovery #4083
- [x] swervolf #4085
- [x] teensy40  #4060
- [x] weact_f401ccu6 #4085

twilfredo:
The Particle Boron now has support for start() fixed in https://github.com/tock/tock/pull/3733

bradjc:
Done

PR #3520, opened 2023-06-28
boards: apollo3: Use helper function in main

alistair23:
### Pull Request Overview

Update to use a helper function in main to reduce the stack usage. This is to address the stack usage issue as described in https://github.com/tock/tock/issues/3518.

### Documentation Updated

### Formatting

- [X] Ran `make prepush`.

### DIFF
--- a/chips/apollo3/src/lib.rs
+++ b/chips/apollo3/src/lib.rs
@@ -74,7 +74,10 @@ pub static IRQS: [unsafe extern "C" fn(); 32] = [CortexM4::GENERIC_ISR; 32];
 #[cfg_attr(all(target_arch = "arm", target_os = "none"), used)]
 pub static PATCH: [unsafe extern "C" fn(); 16] = [unhandled_interrupt; 16];
 
+// The SVC call in this function means that we need to ensure it's inlined in
+// `main()` otherwise we end up with a clobbered stack.
 #[cfg(all(target_arch = "arm", target_os = "none"))]
+#[inline(always)]
 pub unsafe fn init() {
     use core::arch::asm;
     let cache_ctrl = crate::cachectrl::CacheCtrl::new();
--- a/boards/apollo3/lora_things_plus/src/main.rs
+++ b/boards/apollo3/lora_things_plus/src/main.rs
@@ -203,14 +203,15 @@ impl KernelResources<apollo3::chip::Apollo3<Apollo3DefaultPeripherals>> for LoRa
     }
 }
 
+// Ensure that `setup()` is never inlined
+// This helps reduce the stack frame, see https://github.com/tock/tock/issues/3518
+#[inline(never)]
 unsafe fn setup() -> (
     &'static kernel::Kernel,
     &'static LoRaThingsPlus,
     &'static apollo3::chip::Apollo3<Apollo3DefaultPeripherals>,
     &'static Apollo3DefaultPeripherals,
 ) {
-    apollo3::init();
-
     let peripherals = static_init!(Apollo3DefaultPeripherals, Apollo3DefaultPeripherals::new());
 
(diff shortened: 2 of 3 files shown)


bradjc:
I'm not sure what this change does. It looks like the `main` stack frame is the same size:

```
From github.com:tock/tock
 * [new ref]             refs/pull/3520/head -> pr/3520
Switched to branch 'pr/3520'
❯ cd ..
❯ cd ../apollo3/lora_things_plus
❯ make stack-analysis
info: component 'llvm-tools' for target 'x86_64-apple-darwin' is up to date
lora_things_plus
----------------------
   main stack frame:
           24     main

   5 largest stack frames:
         2408     _ZN7cortexm24kernel_hardfault_arm_v7m17h88793f49ade3c395E
         1856     _ZN16lora_things_plus5setup17h49e91a52febf179eE
          832     _ZN95_$LT$kernel..process_standard..ProcessStandard$LT$C$GT$$u20$as$u20$kernel..process..Process$GT$18print_full_process17hd2851a4b78b89a34E
          592     _ZN6kernel5debug11panic_print17hf573d43131026b23E
          520     _ZN95_$LT$kernel..process_standard..ProcessStandard$LT$C$GT$$u20$as$u20$kernel..process..Process$GT$11try_restart17h699970668029a181E

❯ git co master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
❯ make stack-analysis
lora_things_plus
----------------------
   main stack frame:
           24     main

   5 largest stack frames:
         2408     cortexm::kernel_hardfault_arm_v7m
         1856     lora_things_plus::setup
          832     <kernel::process_standard::ProcessStandard<apollo3::chip::Apollo3<apollo3::chip::Apollo3DefaultPeripherals>> as kernel::process::Process>::print_full_process
          600     kernel::debug::panic_print::<lora_things_plus::io::Writer, apollo3::chip::Apollo3<apollo3::chip::Apollo3DefaultPeripherals>, kernel::process_printer::ProcessPrinterText>
          520     <kernel::process_standard::ProcessStandard<apollo3::chip::Apollo3<apollo3::chip::Apollo3DefaultPeripherals>> as kernel::process::Process>::try_restart
```

alistair23:
The main difference is that we are now have `#[inline(never)]` on `setup()`. `setup()` might be inlined (as I guess it is in that comparison) but we now ensure it's never inlined.

chips/apollo3/src/lib.rs
```diff
@@ -75,6 +75,7 @@ pub static IRQS: [unsafe extern "C" fn(); 32] = [CortexM4::GENERIC_ISR; 32];
 pub static PATCH: [unsafe extern "C" fn(); 16] = [unhandled_interrupt; 16];
 
 #[cfg(all(target_arch = "arm", target_os = "none"))]
+#[inline(always)]
```
bradjc, line 78: Why are we now inlining this?

Can you please comment to explain? We're learning that we can't keep track of why we are including inline directives on various functions, and we need to start documenting them.

alistair23: Comments added!

hudson-ayers (approved):

hudson-ayers:
I think this PR is correct.

The two `#[inline(never)]`'s that were added are both on `setup()` calls to reduce stack frame size.

The added `#[inline(always)]` is definitely needed. If that call is not inlined, the board *will not run* because the `svc` call within relies on it being called from the main function. It not being there before was a bug, but Alistair probably did not hit the bug because it happened to be getting inlined in practice. This change might have caused LLVM to not inline the function, which would break things, so he probably noticed it as a result.

bradjc:
Ok I somehow missed that we've been adding inline never with create_peripherals.

bradjc (approved):

PR #3521, opened 2023-06-28
Add jlink to shell.nix

Samir-Rashid:
### Pull Request Overview

The nix shell config does not support jlink. It errors as so:
`FileNotFoundError: [Errno 2] No such file or directory: 'JLinkExe'`

I added a line users can edit to add it. 

### Testing Strategy

`make install` works. Added EULA accept as required by jlink.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -14,7 +14,10 @@
 #  $ nix-shell
 #
 
-{ pkgs ? import <nixpkgs> {} }:
+{ pkgs ? import <nixpkgs> {
+  config.allowUnfree = true;
+  config.segger-jlink.acceptLicense = true;
+} }:
 
 with builtins;
 let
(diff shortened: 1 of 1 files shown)


bradjc:
Since not all boards/users use jlink we might not want to include this by default.

But I don't use nix so I don't have a horse in this race, @lschuermann @alevy thoughts?

lschuermann:
Yes, I don't think that we should include this unconditionally. We most certainly should not ship any unfree code and accept EULA / ToS on a user's behalf.

Apart from those concerns, I'm not sure whether this is useful in the `shell.nix` file: the JLink installation also requires udev rules to be included system-wide. Given that, I don't see why we shouldn't just use a system-wide JLink or have users enter into a nix-shell with JLink installed themselves.

Samir-Rashid:
Closing. Reasonable for user to install board specific dependencies themselves.

alevy:
Maybe we can add board-specific nix shell scripts in their respective board directories that pull in additional tools?

Issue #3308, opened 2022-10-31
Tracking: Initial IEEE 802.3 Wired Ethernet and IP support in Tock


lschuermann:
This issue shall serve as an overview and progress tracker for introducing IEEE 802.3 Ethernet support in Tock. If you feel there is something missing / incorrect here, feel free to leave a comment and/or edit this issue. This based on discussions at TockWorld 5 and #3103.

## Motivation

Tock already has (limited) upstream support for some communication protocols, such as Bluetooth LE, IEEE 802.15.4 with 6LoWPAN, as well as serial transports over UART & USB. Nonetheless, having at least limited support for wired Ethernet and IP in Tock seems attractive for a variety of reasons:
- Ethernet is a comparatively simple communication interface. While its high-bandwidth properties and relatively large packets may not map well to some of the smaller chips we support, it is easy to prototype networked applications on larger and Ethernet-capable boards.
- Ethernet is an accessible technology. No special hardware is required to get started (except for an Ethernet-capable supported board). This is unlike many wireless interfaces, which require secondary boards or hardware to integrate with other systems.
- Ethernet is easy to debug. Whereas wireless protocols may require specialized sniffers and debugging hardware, most Ethernet issues can be diagnosed using a regular PC and a packet sniffer application.
- Ethernet & IP can allow Tock devices to talk over the Internet and operate independently of other gateway devices or host computers. This may be especially interesting for prototyping and showcasing OTA updates.

## Approach

Developing an efficient, maintainable and virtualized (routing UDP & TCP flows to different clients) in-kernel TCP/IP stack is not trivial. Relying on pre-existing TCP/IP implementations can bridge the gap between having IP-capable Tock applications and a well-designed and functional IP stack implemented in the Tock kernel. An example for such a preexisting implementation is the [LwIP library](https://www.nongnu.org/lwip/2_1_x/index.html), released under a 3-clause BSD license. Hence I propose the following approach and scope for initial support of IEEE 802.3 Wired Ethernet and IPv6:

1. Introduce support for a set of Ethernet-capable chips and/or external Ethernet MACs (for example the ENC28J60 connected via SPI)
2. Design an Ethernet adapter HIL based on these implementations. Developing the HIL to fit multiple different chips and external MACs should make it more flexible and support a wide range of different devices.
3. Implement a system call driver, exposing a TAP-like interface to userspace (i.e., passing Ethernet frames from/to a userspace application).
4. Provide LwIP as a `libtock-c` library and supply a dummy application.

[...]

PR #3523, opened 2023-06-28
Ethernet support for STM32F429

Ioan-Cristian:
### Pull Request Overview

This pull request adds basic Ethernet support for STM32F429. At the moment, the
peripheral supports only raw Ethernet frames and there is no support for
address filtering, hardware checksum control or chaining descriptors for Ethernet DMA.

### Testing Strategy

The Ethernet implementation was tested through basic unit tests and manually running a web server. (lwIP)

### TODO or Help Wanted

Issues:

+ In rare situations, the peripheral (or the capsule?) hangs and no frame is
processed anymore.
+ When compiling with level 2 or 3 optimizations, the web server doesn't work.

TODO:

+ Try to successfully compile and run the web server with optimizations (high priority)
+ Full UDP stack in the kernel
+ Offload workload from software to hardware
+ Chaining descriptors for Ethernet DMA.
+ Address filtering
+ Packets statistics: drop frames, bad frames, received/transmitted frames etc
for profiling

### Documentation Updated

- [+] Updated the relevant files in `/docs`

### Formatting

- [+] Ran `make prepush`.

### Updates

+ 29/06/2023: This pull request was converted to a draft, since the clock implementation was split up in #3528


boards/weact_f401ccu6/layout.ld
```diff
@@ -1,13 +1,13 @@
...
+ * ram = 64KiB */
 
 MEMORY
 {
-  rom (rx)  : ORIGIN = 0x08000000, LENGTH = 0x00020000
-  prog (rx) : ORIGIN = 0x08020000, LENGTH = 0x00020000
+  rom (rx)  : ORIGIN = 0x08000000, LENGTH = 150K
+  prog (rx) : ORIGIN = 0x08025800, LENGTH = 102K
   ram (rwx) : ORIGIN = 0x20000000, LENGTH = 64K
 }
```
Ioan-Cristian, line 13: The total flash size for the kernel was increased because otherwise the kernel couldn't be compiled in debug mode. This problem seems to also occur on the `master` branch.

capsules/extra/src/ethernet_app_tap.rs
```diff
@@ -213,7 +213,7 @@ pub const DRIVER_NUM: usize = capsules_core::driver::NUM::EthernetTAP as usize;
 
 /// Maximum size of a packet which can be transmitted over the underlying
 /// [`EthernetAdapter`] device. Currently hard-coded to `1522` bytes.
-pub const MAX_MTU: usize = 1522;
+pub const MAX_MTU: usize = 1524;
```
Ioan-Cristian, line 216: Receive descriptors allow buffer lengts multiple of 4, 8 or 16, depending how the DMA is configured.

chips/stm32f4xx/Cargo.toml
```diff
@@ -12,3 +12,9 @@ edition.workspace = true
 cortexm4 = { path = "../../arch/cortex-m4" }
 enum_primitive = { path = "../../libraries/enum_primitive" }
 kernel = { path = "../../kernel" }
+
+[features]
+stm32f401 = []
+stm32f412 = []
+stm32f429 = []
+stm32f446 = []
```
Ioan-Cristian, line 20: To avoid a lot of code duplication, conditional compilation was preferred.

chips/stm32f4xx/src/flash.rs
```diff
@@ -0,0 +1,711 @@
...
+//! debug!("Current flash latency is {}", flash_latency);
+//! ```
+
+use kernel::debug;
+use kernel::utilities::registers::interfaces::{ReadWriteable, Readable};
+use kernel::utilities::registers::{register_bitfields, register_structs, ReadWrite, WriteOnly};
+use kernel::utilities::StaticRef;
+use kernel::ErrorCode;
+
+register_structs! {
```
Ioan-Cristian, line 43: This struct represents the common superset of all chips Flash. The field that is not available on some chips shouldn't be accessed. This was done to avoid code duplication.

chips/stm32f4xx/src/flash.rs
```diff
@@ -0,0 +1,711 @@
...
+        /// Flash option control register
+        (0x014 => optcr: ReadWrite<u32, OPTCR::Register>),
+        /// Flash option control register 1
+        // NOTE: This register is present only on some chip models
+        (0x018 => optcr1: ReadWrite<u32>),
+        (0x01C => @END),
+    }
+}
+
+register_bitfields![u32,
```
Ioan-Cristian, line 65: This is the common superset of all possible bitfieds for all chips. This was done to avoid code duplication.

bradjc (changes requested):
Blocking because of the added nightly feature which since we want to build Tock on stable rust we would be hesitant to add.

Generally this looks good, and it's exciting to see this implementation!

I have to say, the cargo features make me uneasy. I'm not entirely against adding them because _when_ to use them is not ambiguous: either the kernel is being compiled for a specific chip or its not. They aren't features in the sense that different developers might turn them on or off to get different functionality.

However, they do present an opportunity for confusion: is the MCU variant code in the variant crate, or in the general chip crate guarded by a feature? In general, to avoid this confusion we put shared code in the general chip crate, and then "customize it" (often for example by setting the MMIO address of a peripheral) in the variant-specific crate. I think that would be our preference.

However, there may still be reasons to use features. Another way to address the confusion about where all the variant-specific customization is is to group the settings in one (and only one) place, like we did here: https://github.com/tock/tock/blob/master/chips/earlgrey/src/chip_config.rs. That way, the feature is only "used" once, it is clear where to document all of the differences and why the differences need to be included in Tock, and the remaining code reads as traditional normal rust code.

What I would like to avoid is having features used in many places haphazardly.

Also, I'm not worried about features in the tests.

boards/nucleo_f429zi/src/main.rs
```diff
@@ -44,7 +50,7 @@ const FAULT_RESPONSE: kernel::process::PanicFaultPolicy = kernel::process::Panic
 /// Dummy buffer that causes the linker to reserve enough space for the stack.
 #[no_mangle]
 #[link_section = ".stack_buffer"]
-pub static mut STACK_MEMORY: [u8; 0x2000] = [0; 0x2000];
+pub static mut STACK_MEMORY: [u8; 0x4000] = [0; 0x4000];
```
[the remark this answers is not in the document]
lschuermann: This is expected because of my terribly inefficient `tap_driver` implementation on the `tock-ethernet` branch. This PR is targeting this branch, and so we should focus on the diff concerning the added driver here. When it comes time to merge `tock-ethernet` into master, I think it's a good opportunity to consider the remaining infrastructure on that branch. Things are still WIP there.

capsules/extra/src/ethernet_app_tap.rs
lschuermann: This is a good point. We should reconsider this in the design when we revisit the packet buffer passing story around Tock.

chips/stm32f429zi/src/ethernet/mod.rs
```diff
@@ -0,0 +1,2328 @@
...
+
+mod transmit_descriptor;
+use crate::ethernet::transmit_descriptor::TransmitDescriptor;
+
+mod receive_descriptor;
+use crate::ethernet::receive_descriptor::ReceiveDescriptor;
+
+register_structs! {
+    /// Ethernet: media access control
+/// (MAC)
```
lschuermann, line 45: We should fix the formatting in those register definitions.

chips/stm32f429zi/src/ethernet/mod.rs
```diff
@@ -0,0 +1,2328 @@
...
+/// register
+        (0x05C => maca3lr: ReadWrite<u32>),
+        (0x060 => @END),
+    }
+}
+
+register_bitfields![u32,
+MACCR [
+    /// RE
+    RE OFFSET(2) NUMBITS(1) [],
```
lschuermann, line 113: Perhaps fix the indentation here as well. It's unfortunate that `rustfmt` doesn't really format our custom macros well.

chips/stm32f429zi/src/ethernet/mod.rs
```diff
@@ -0,0 +1,2328 @@
...
+        assert_eq!(false, ethernet.did_abnormal_interrupt_occur());
+    }
+
+    /// Test Ethernet initialization
+    pub fn test_ethernet_init(ethernet: &Ethernet) {
+        debug!("~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~");
+        debug!("Testing Ethernet initialization...");
+
+        // This is broken for some reasons
+        assert_eq!(Ok(()), ethernet.init());
```
lschuermann, line 2048: If this is broken, we probably shouldn't ship this test? Do you know what the underlying root cause is?

lschuermann (changes requested):
I did a preliminary review of these changes, the Ethernet driver code looks pretty good! Certainly an impressive feat to get this complex of a MAC under control. :+1: This is still lacking a pass over the code to address @bradjc's comments, and the TODOs and formatting through the changeset.

To give some more context for these changes: this is part of #3308. The PR is made against the `tock-ethernet` branch, which already contains some initial in-kernel infrastructure and a minimal "Ethernet HIL" to support userspace network stacks (e.g., currently an LwIP-based webserver in `libtock-c`, `smoltcp` integration in `libtock-rs` is WIP).

I think it would help for the STM clocking support to be split out of this changeset (perhaps using git filter) and PRed against the upstream master branch directly. I can then rebase `tock-ethernet` to incorporate those changes and reduce the diff in this PR.

Ioan-Cristian:
I am going to close this pull request and create two new ones: one for the clocks and one for the Ethernet as suggested. Both of them should be ready in the coming hours.

Ioan-Cristian:
The clock implementation was split up in a new pull request. For all clock related issues, see #3528 .

chips/stm32f429zi/src/ethernet/mod.rs
Ioan-Cristian: Fixed formatting.

chips/stm32f4xx/src/lib.rs
```diff
@@ -9,6 +9,7 @@
 #![crate_name = "stm32f4xx"]
 #![crate_type = "rlib"]
 #![no_std]
+#![feature(core_panic)]
```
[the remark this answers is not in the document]
Ioan-Cristian: No longer related to this pull request.

boards/weact_f401ccu6/layout.ld
```diff
@@ -1,13 +1,13 @@
...
+ * kernel = 150KiB
+ * user = 102KiB
+ * ram = 64KiB */
 
 MEMORY
 {
-  rom (rx)  : ORIGIN = 0x08000000, LENGTH = 0x00020000
-  prog (rx) : ORIGIN = 0x08020000, LENGTH = 0x00020000
+  rom (rx)  : ORIGIN = 0x08000000, LENGTH = 150K
+  prog (rx) : ORIGIN = 0x08025800, LENGTH = 102K
```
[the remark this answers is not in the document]
Ioan-Cristian: No longer related to this pull request.

boards/nucleo_f429zi/src/main.rs
```diff
@@ -16,13 +16,19 @@ use capsules_core::virtualizers::virtual_alarm::VirtualMuxAlarm;
...
 use kernel::hil::led::LedHigh;
 use kernel::platform::{KernelResources, SyscallDriverLookup};
 use kernel::scheduler::round_robin::RoundRobinSched;
 use kernel::{create_capability, debug, static_init};
 
 use stm32f429zi::gpio::{AlternateFunction, Mode, PinId, PortId};
 use stm32f429zi::interrupt_service::Stm32f429ziDefaultPeripherals;
+#[allow(unused_imports)]
+use stm32f429zi::rcc::{APBPrescaler, MCO1Source, SysClockSource};
+use stm32f429zi::syscfg::EthernetInterface;
```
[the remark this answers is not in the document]
Ioan-Cristian: Removed `#[allow(unused_imports)]`.

boards/nucleo_f429zi/src/main.rs
```diff
@@ -288,6 +299,63 @@ unsafe fn setup_peripherals(
     can1.enable_clock();
 }
 
+// Helper function to setup Ethernet pins
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
Ioan-Cristian: Removed `#[allow(dead_code)]`.

alevy:
@Ioan-Cristian I think I shot you in the foot a bit by pushing directly to the `tock-ethernet` branch a merge with master, so _this_ PR now has >400 files changed compared to current `tock-ethernet` and >700 commits. This was my mistake, but I think we can fix this with another merge or cherry picking your commits here onto `tock-ethernet`

lschuermann:
Sorry for the confusion here! We merged `master` into `tock-ethernet` to keep it up to date, which is reasonable for a shared development branch. However, this has seemed to result in this PR including all of these commits in its diff as well.

I have verified locally that git is clever enough to produce a merge commit which only includes the files actually changed in this PR. Instead it appears that this PR's diff can be explained by [GitHub's three-dot diff](https://stackoverflow.com/a/74871448) algorithm: whereas a `git diff $TARGET_BRANCH..$PR_BRANCH` would show you the full delta between the merge target and the PR (and thus also include changes made as the target moves forward), GitHub picks a common ancestor on both the target branch and PR branch. It then effectively displays a `git diff $COMMON_ANCESTOR..$PR_BRANCH`. However, this is not what a merge-commit would contain: merges can detect duplicate commits and identical changes, and so the actual merge-diff for this PR indeed solely contains the actual changes.

We should be able to fix this PR's diff by either rebasing this branch on `tock-ethernet`, or merging `tock-ethernet` into this branch. The latter seems like a more reasonable option, given this branch also contains merge-commits.

chips/stm32f429zi/src/ethernet/ethernet.rs
```diff
@@ -0,0 +1,2288 @@
...
+use crate::rcc;
+use crate::rcc::PeripheralClock;
+use crate::rcc::PeripheralClockType;
+
+use crate::ethernet::receive_descriptor::ReceiveDescriptor;
+use crate::ethernet::transmit_descriptor::TransmitDescriptor;
+
+register_structs! {
+    /// Ethernet: media access control (MAC)
+    Ethernet_MacRegisters {
```
lschuermann, line 42: Nit: usually, names in Rust follow either "snake_case", "UPPER_CASE_SNAKE_CASE" or "PascalCase", but not a mixture of those. Typically we try to stick to those, but here it might not matter: the tock-registers macros generate the appropriate `#[allow(_)]` markers for names to violate these rules (as we use the names for both modules and structs).

chips/stm32f429zi/src/ethernet/ethernet.rs
```diff
@@ -0,0 +1,2288 @@
...
+        self.forward_undersized_good_frames();
+        self.enable_transmit_store_and_forward()?;
+        self.enable_receive_store_and_forward()?;
+
+        self.set_transmit_descriptor_list_address(
+            &self.transmit_descriptor as *const TransmitDescriptor as u32,
+        )?;
+        self.set_receive_descriptor_list_address(
+            &self.receive_descriptor as *const ReceiveDescriptor as u32,
+        )?;
```
lschuermann, line 808: Note to self: This is safe, given that both `TransmitDescriptor` and `ReceiveDescriptor` are `#[repr(C)]` structs which only contain `InMemoryRegister`s, which in turn wrap `VolatileCell` -- so we don't need to main Rust's aliasing requirements here, and spurious reads are also not an issue.

chips/stm32f429zi/src/ethernet/ethernet.rs
```diff
@@ -0,0 +1,2288 @@
...
+    }
+
+    fn did_normal_interrupt_occur(&self) -> bool {
+        self.dma_registers.dmasr.is_set(DMASR::NIS)
+    }
+
+    // TODO: Am I allowed to clear this bit?
+    //fn clear_dma_normal_interrupt(&self) {
+    //self.dma_registers.dmasr.modify(DMASR::NIS::SET);
+    //}
```
lschuermann, line 1098: We should probably resolve these comments and add these functions or remove them appropriately. More generally, I think it's perfectly fine for a driver to not expose an exhaustive set of hardware capabilities, especially as accessor methods. Such methods typically suggest that some functionality is either required, or at least supported in a driver. Not implementing them can be interpreted as "we don't know whether touching this bit is safe, so we don't provide a method for it".

chips/stm32f429zi/src/ethernet/ethernet.rs
```diff
@@ -0,0 +1,2288 @@
...
+
+    fn did_abnormal_interrupt_occur(&self) -> bool {
+        self.dma_registers.dmasr.is_set(DMASR::AIS)
+    }
+
+    // TODO: Am I allowed to clear this bit?
+    //fn clear_dma_abnormal_interrupt(&self) {
+    //self.dma_registers.dmasr.modify(DMASR::AIS::SET);
+    //}
```
lschuermann, line 1107: Same here, resolve TODO.

chips/stm32f429zi/src/ethernet/ethernet.rs
```diff
@@ -0,0 +1,2288 @@
...
+            0b011 => DmaReceiveProcessState::WaitingForReceivePacket,
+            0b100 => DmaReceiveProcessState::Suspended,
+            0b101 => DmaReceiveProcessState::ClosingReceiveDescriptor,
+            _ => DmaReceiveProcessState::TransferringReceivePacketDataToHostMemory,
+        }
+    }
+
+    fn did_normal_interrupt_occur(&self) -> bool {
+        self.dma_registers.dmasr.is_set(DMASR::NIS)
+    }
```
lschuermann, line 1093: In general, Tock drivers typically use the tock-registers interface directly to query such bits -- accessor methods like these are mostly used in case of some more complex bit-queries / bit-manipulations. However, in the form of private methods they're also totally fine to have around; feel free to leave them in if that's preferred.

chips/stm32f429zi/src/ethernet/ethernet.rs
```diff
@@ -0,0 +1,2288 @@
...
+        self.dma_registers.dmaomr.modify(DMAOMR::FTF::SET);
+
+        // TODO: Adjust this value
+        for _ in 0..100 {
+            if self.dma_registers.dmaomr.read(DMAOMR::FTF) == 0 {
+                return Ok(());
+            }
+        }
+
+        Err(ErrorCode::BUSY)
```
lschuermann, line 1287: For such methods featuring a timeout, it'd be great to document the exact reasons why this could fail, what the magic timeout parameter means (i.e. why you came up with this value, even if it just happens to work in practice that's worth documenting), and perhaps make it configurable via a parameter.

chips/stm32f429zi/src/ethernet/ethernet.rs
```diff
@@ -0,0 +1,2288 @@
...
+    }
+
+    fn is_transmit_buffer_unavailable_interrupt_enabled(&self) -> bool {
+        self.dma_registers.dmaier.is_set(DMAIER::TBUIE)
+    }
+
+    #[allow(dead_code)]
+    fn get_current_host_transmit_descriptor_address(&self) -> u32 {
+        self.dma_registers.dmachtdr.get()
+    }
```
lschuermann, line 1580: I think it may be better to not keep such unused functions around, assuming you don't plan on integrating them in the near term. If required, we can always re-add them later.

chips/stm32f429zi/src/ethernet/ethernet.rs
```diff
@@ -0,0 +1,2288 @@
...
+
+    fn is_mac_enabled(&self) -> bool {
+        self.is_mac_receiver_enabled() || self.is_mac_transmiter_enabled()
+    }
+
+    fn enable_transmitter(&self) -> Result<(), ErrorCode> {
+        self.enable_dma_transmission()?;
+        self.enable_mac_transmitter();
+
+        for _ in 0..10 {
```
lschuermann, line 1597: Same here, explain timeout.

chips/stm32f429zi/src/ethernet/ethernet.rs
```diff
@@ -0,0 +1,2288 @@
...
+            assert_eq!(Ok(()), self.receive_packet());
+        } else if self.did_early_receive_interrupt_occur() {
+            self.clear_early_receive_interrupt();
+        }
+    }
+
+    fn handle_abnormal_interrupt(&self) {
+        if self.did_fatal_bus_error_interrupt_occur() {
+            self.clear_fatal_bus_error_interrupt();
+            panic!("Fatal bus error");
```
lschuermann, line 1703: Perhaps we should avoid panicing here in the long term, except if these are errors which really put the chip in an unrecoverable state. However, something like a "transmit buffer underflow" seems like the kind of interrupt which could be occurring at runtime without major impact (except for a broken Ethernet frame), depending on the AHB bus.

chips/stm32f429zi/src/ethernet/ethernet.rs
```diff
@@ -0,0 +1,2288 @@
...
+
+    // TODO: Remove this once Transmit trait is implemented
+    fn start_transmit(&self) -> Result<(), ErrorCode> {
+        self.enable_transmitter()
+    }
+
+    // TODO: Remove this once Receive trait is implemented
+    fn start_receive(&self) -> Result<(), ErrorCode> {
+        self.enable_receiver()
+    }
```
lschuermann, line 1845: I believe these functions are no longer relevant? In general, a receive API will likely be asynchronous, with the MAC maintaining a (shared) buffer poll which it can pass to a registered client on demand.

lschuermann (changes requested):
My apologies for the (significantly) delayed review! I did my first full thorough pass through the `stm32f429zi/src/ethernet/ethernet.rs`, and overall it looks great! I especially appreciate the clean separation of low-level methods, which essentially just map hardware-operations to Rust methods, and high-level methods that configure the hardware & implement TX / RX.

I left a few comments, many of which are nits. The driver contains a lot of simple accessor methods, many of which are only used once or twice; those could be inlined, but that's a question of style.

More generally, I don't think the additions to the preliminary WIP Ethernet HIL in this PR are objectionable, but they certainly warrant further discussion. I presume we'd workshop them as part of the Network WG, and under consideration of the other Ethernet MACs, but that is outside of the scope of this PR.

Assuming #3528 gets in (which I hope to be able to take a look at next), I think this is pretty solid in terms of its driver implementation. I don't see many blockers for the other (HIL) changes to land in `tock-ethernet` as well.

chips/stm32f429zi/src/ethernet/ethernet.rs
alevy: Alternatively, if this is really guaranteed to succeed _eventually_ it's OK to loop until the condition is true. E.g., some of the GPIO hardware (such as the SAM4L) have a busy bit which will eventually clear, so the drivers for that hardware just loops on the busy bit before reading or setting the GPIO state and there is no error condition.

chips/stm32f429zi/src/ethernet/ethernet.rs
Ioan-Cristian: > we don't know whether touching this bit is safe, so we don't provide a method for it

That's exactly the issue. The general description of DMA interrupts (page 1189):

> When all the enabled interrupts within a group are cleared, the corresponding summary bit is cleared.

The page describing registers (1226):

> The normal interrupt summary bit value is the logical OR of the following when the corresponding interrupt bits are enabled in the ETH_DMAIER register

So I far I understand that the summary bit is automatically cleared. However, the same page describing summary bits:

> This is a sticky bit and it must be cleared (by writing a 1 to this bit) each time a corresponding bit that causes NIS to be set is cleared
> This is a sticky bit and it must be cleared each time a corresponding bit that causes AIS to be
set is cleared

Do I miss something?

chips/stm32f429zi/src/ethernet/ethernet.rs
Ioan-Cristian: I find it's easier to understand what code does when using methods with explicit names compared to registers named `dmasr`, especially that the Tock guideline suggests keeping the name of the registers the same as in the reference manual. However, I'm not opposed to remove them if you wish so.

chips/stm32f429zi/src/ethernet/ethernet.rs
Ioan-Cristian: Some of these for loops can fail if the peripheral mode configuration register doesn't match the actual hardware. Not sure if I should replace them with while loops since by the time the MAC and DMA cores are initialized, `syscfg` is already configured in the right mode. The sole benefit of keeping a for loop would be spotting easier the case where the syscfg is not configured correctly, since this DMA/MAC core seems to be present on multiple boards Tock supports (MSP432 for instance).

chips/stm32f429zi/src/ethernet/ethernet.rs
lschuermann: It seems fine to keep them, just a style choice.

chips/stm32f429zi/src/ethernet/ethernet.rs
Ioan-Cristian: See above.

chips/stm32f429zi/src/ethernet/ethernet.rs
Ioan-Cristian: Fixed names.

chips/stm32f429zi/src/ethernet/ethernet.rs
Ioan-Cristian: The story is that when I started working on the Ethernet, I was not aware of the `tock-ethernet` branch, so I started to create my own Ethernet HIL. Since Ethernet is another mean for transmitting and receiving data, I mimicked the UART HIL. Now, all traits are removed from the HIL.

chips/stm32f429zi/src/ethernet/ethernet.rs
Ioan-Cristian: Removed most of the panics and added additionally comments to explain the intent of the panic handler actions. Might require future revision.

Ioan-Cristian:
Addressed most of the comments. However, I don't have any board near me to test the changes, so I suggest you to wait until tomorrow before reviewing them. I'll inform you when testing is done.

Ioan-Cristian:
Completely removed any addition from the Ethernet HIL.
lwIP and smoltcp apps still working.
The only `#[allow(dead_code)]` functions that I kept are those for disabling the Ethernet peripheral, but you think there will never be a need to disable the Ethernet peripheral once started, I can delete them.
I hope I didn't forget to address any comment. If I did, please tell me.
Waiting for your other comments/issues. Thank you!

chips/stm32f429zi/src/ethernet/ethernet.rs
Ioan-Cristian: Added comment with your explanation.

capsules/extra/src/ethernet_app_tap.rs
brghena: The comment here needs to be updated. Probably worth noting the length limitations of 4/8/16

Ioan-Cristian: Updated comment. Probably this constant will be removed from the capsule and instead queried from the peripheral itself once the Networking Working Group establish a HIL for that.

Ioan-Cristian:
Hello! Sorry for my long inactivity! Is there a possibility to merge the upstream master branch?

Ioan-Cristian:
I managed to cherry pick all the relevant commits to reduce the number of changed files. Would it be possible to merge the `master` branch into `tock-ethernet` branch? The Ethernet implementation relies on clocks.

bradjc:
@lschuermann

PR #3528, opened 2023-06-29
Basic system clock management for the STM32F4 family

Ioan-Cristian:
### Pull Request Overview

This pull request adds support for basic clock management for the STM32F4 family:

+ high-speed internal (HSI) clock configuration
+ main phase-locked loop (PLL) clock configuration
+ Dynamic system clock source and frequency management with system integrity
checks

### Testing Strategy

This pull request was thoroughly tested through unit tests, integration tests
and logic analyzer. The tests were run on STM32F429ZI Nucleo-144 and STM32F446
Nucleo-64 boards.

### TODO or Help Wanted

TODO:

+ Add support for high-speed external (HSE) clock
+ Add support for secondary PLL clock

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


chips/stm32f4xx/src/clocks/mod.rs
```diff
@@ -0,0 +1,991 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
```
bradjc, line 1: Please rename this file to match its contents.

chips/stm32f429zi/Cargo.toml
```diff
@@ -11,5 +11,5 @@ edition.workspace = true
 [dependencies]
 cortexm4 = { path = "../../arch/cortex-m4" }
 kernel = { path = "../../kernel" }
-stm32f4xx = { path = "../stm32f4xx" }
+stm32f4xx = { path = "../stm32f4xx", features = ["stm32f429"] }
```
Ioan-Cristian, line 14: Addressed the issue described in #3523 : change to inline features.

chips/stm32f4xx/src/flash.rs
```diff
@@ -0,0 +1,711 @@
...
+            5 => FlashLatency::Latency5,
+            6 => FlashLatency::Latency6,
+            // The hardware allows 3-bit latency values
+            _ => FlashLatency::Latency7,
+        }
+    }
+
+    // TODO: Take into the account the power supply
+    //
+    // NOTE: This method is pub(crate) to prevent a capsule from modifying the flash latency. Flash
```
bradjc, line 350: A capsule wouldn't be able to call this function even if it was just `pub` because the capsule crates do not have the stmX crates as dependencies.

chips/stm32f4xx/src/lib.rs
```diff
@@ -9,6 +9,7 @@
 #![crate_name = "stm32f4xx"]
 #![crate_type = "rlib"]
 #![no_std]
+#![feature(core_panic)]
```
bradjc, line 12: Nightly feature.

bradjc (changes requested):
Same comments as before about the nightly feature and use of features in the chip crate.

chips/stm32f4xx/src/lib.rs
Ioan-Cristian: This was added because of the `check_and_panic!` macro:

```rust
macro_rules! check_and_panic {
	($left:expr, $right:expr, $clocks: ident) => {
		match (&$left, &$right) {
			(left_val, right_val) => {
				if !(*left_val == *right_val) {
					set_default_configuration($clocks);
					let kind = core::panicking::AssertKind::Eq;
					// The reborrows below are intentional. Without them, the stack slot for the
					// borrow is initialized even before the values are compared, leading to a
					// noticeable slow down.
					core::panicking::assert_failed(
						kind,
						&*left_val,
						&*right_val,
						core::option::Option::None,
					);
				}
			}
		};
	};
}
```

This macro ensures that if a test fails, the system clock is reverted back to
its default configuration. This prevents messing up with the UART baud rate.

A solution to this problem could be:

```rust
macro_rules! check_and_panic {
	($left:expr, $right:expr, $clocks: ident) => {
		match (&$left, &$right) {
			(left_val, right_val) => {
				if *left_val != *right_val {
					set_default_configuration($clocks);
				}
				assert_eq!($left, $right);
			}
		};
	};
}
```

The downside of this approach is that `left` and `right` are evaluated twice.

Another solution is to use `panic!` with a custom message for each test. The
downside of this approach is evident: one has to write a custom panic message
for each occurrence of `check_and_panic`.

Do you have a better solution?

Ioan-Cristian:
@bradjc If I understood you correctly, you would like to get rid of the features. I have tried all day to come up with a solution that doesn't rely on features and prevents code duplication. Without success so far.

Creating a file containing all constants, types and functions is trivial to implement, even though I am not a big fan of it, since I prefer to keep flash-related stuff in `flash.rs`, pll-related stuff in `clocks/pll.rs` and so on. However, this solution doesn't eliminate the need for features.

I will keep trying to find a better solution this evening and tomorrow and if I cannot make any progress, I will revert back to your solution. Are you OK with this?

bradjc:
Re: features: I don't think we must not use them in this case (as they are not optional and there is no uncertainty about when these type of features must be used). But I would like to resolve two questions:

1. How should future developers know which code to put in a variant-specific crate, and which code should be behind a feature? When someone is reading the code, how should they determine whether to look for a cfg or for it in a variant-specific crate?
2. Can the feature be used exactly once (or in one very specific place)?

chips/stm32f4xx/src/clocks/mod.rs
Ioan-Cristian: This file defines a structure called `Clocks` which encapsulates all clocks present on the chip. Do you want me to rename the file to `clocks.rs`? I find this ugly: `crate::clocks::clocks::Clocks`.

Ioan-Cristian: Yes. My fault. Sorry!

chips/stm32f4xx/src/flash.rs
Ioan-Cristian: The comment text was wrong. It's `pub(crate)` to prevent modifying the flash latency from the board main file. Fixed it.

chips/stm32f4xx/src/lib.rs
Ioan-Cristian: Implemented the first solution.

boards/Makefile.common
```diff
@@ -66,7 +66,7 @@ RUSTC_FLAGS ?= \
   -C linker-flavor=ld.lld \
   -C relocation-model=static \
   -C link-arg=-nmagic \
-  -C link-arg=-icf=all \
+  -C link-arg=-icf=all
```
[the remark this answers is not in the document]
Ioan-Cristian: Solved this.

chips/stm32f4xx/src/clocks/mod.rs
Ioan-Cristian: Renamed the file.

Ioan-Cristian:
@bradjc Moved all chip-specific code to `chip_specific.rs` file. What do you think about it?

bradjc:
I think this is close, but I'm wary of using features to define a trait (with different mcus defining different version of the trait). That seems complex and unexpected to me.

chips/stm32f4xx/src/rcc.rs
```diff
@@ -727,14 +707,243 @@ register_bitfields![u32,
...
+    pub(crate) fn set_mco1_clock_source(&self, source: MCO1Source) {
+        self.registers.cfgr.modify(CFGR::MCO1.val(source as u32));
+    }
+
+    pub(crate) fn get_mco1_clock_source(&self) -> MCO1Source {
+        match self.registers.cfgr.read(CFGR::MCO1) {
+            0b00 => MCO1Source::HSI,
+            //0b01 => MCO1Source::LSE,
+            //0b10 => MCO1Source::HSE,
+            _ => MCO1Source::PLL,
```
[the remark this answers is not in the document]
ppannuto: Oh... I get it, https://github.com/tock/tock/pull/3528/files#diff-31fea138f9c2d02594479e3892b86f77b6a254a8647e6f1e5352448a2e45ae20R1269 doesn't define `HSE` yet so folks can't try to pass something unsupported.

Maybe better to have an `MC01Source::UNSUPPORTED` or `::INVALID` or similar in that enum for the default match here (and match `PLL` explicitly)

chips/stm32f4xx/src/flash.rs
```diff
@@ -0,0 +1,549 @@
...
+        (0x008 => optkeyr: WriteOnly<u32>),
+        /// Status register
+        (0x00C => sr: ReadWrite<u32, SR::Register>),
+        /// Control register
+        (0x010 => cr: ReadWrite<u32, CR::Register>),
+        /// Flash option control register
+        (0x014 => optcr: ReadWrite<u32, OPTCR::Register>),
+        /// Flash option control register 1
+        // NOTE: This register is present only on some chip models
+        (0x018 => optcr1: ReadWrite<u32>),
```
ppannuto, line 65: This is one of the cases where we explicitly allow use of `#cfg[` based on the specific chip in use. As example, `chips/nrf5x/*` uses the pattern somewhat heavily to differentiate the hardware physically available on the `nrf51` vs `nrf52` family. It should be compile-time impossible to try to access a register on a chip that doesn't have it.

ppannuto: Though I fear I'm slightly contradicting or giving confusing feedback with @bradjc's comment here: https://github.com/tock/tock/pull/3523#pullrequestreview-1503774449

I think the balance we've come to to-date is that generic `abcXX` crate should use `cfg` to define the presence / absence of hardware correctly, but any implementation or function/etc that would try to _use_ that register should not be `cfg`'d around in the `abcXX` crate, but rather should be in the specific `abc22` chip crate that actually has the hardware.

chips/stm32f4xx/src/chip_specific.rs
```diff
@@ -0,0 +1,269 @@
...
+        /// 5 wait cycles
+        Latency5,
+        /// 6 wait cycles
+        Latency6,
+        /// 7 wait cycles
+        Latency7,
+    }
+
+    // Chip-specific trait that allows to read the current configured flash latency
+    pub(crate) trait SpecificFlashTrait {
```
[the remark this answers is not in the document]
ppannuto: Hmm.. if I understand what's happening correctly, the concept here is that the _only_ file with `#cfg[`'s unrelated to unit testing is this `chip_specific.rs` file.

While I'm always supportive of efforts to contain and ease reasoning about `#define`-hell, in this case I think it makes things a bit more confusing as part of the flash implementation is no longer in the `flash.rs` file.

This is an interesting ergonomics question though. Maybe something we should bring up briefly on a core call.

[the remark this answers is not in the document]
bradjc: > Hmm.. if I understand what's happening correctly, the concept here is that the _only_ file with `#cfg[`'s unrelated to unit testing is this `chip_specific.rs` file.

Yes, as we did with the earlgrey chip.

> While I'm always supportive of efforts to contain and ease reasoning about `#define`-hell, in this case I think it makes things a bit more confusing as part of the flash implementation is no longer in the `flash.rs` file.

I think we can just leave the config/options/settings here and all the flash stuff can move to flash.rs. The earlgray chip really is a good example: https://github.com/tock/tock/blob/master/chips/earlgrey/src/chip_config.rs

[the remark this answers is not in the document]
ppannuto: So basically leave this as-is, only as a function called from the `flash.rs` rather than a trait on the Flash object?

The big difference I see here is that earlgrey is all const's. Here, there is runtime code difference—critically a different number of frequency ranges to check per chip. This will always* have to be a function somewhere whose body is `cfg`'d.

*I guess you could do something like a array of (mhz_threshold, Latency)'s where the array is the `cfg`'d const and then iterate the array at runtime, but that seems a lot worse and more confusing than what is effectively picking the right jump table currently. Or probably have some macro that creates the right match block, but that again feels really opaque.

chips/stm32f4xx/src/chip_specific.rs
```diff
@@ -0,0 +1,269 @@
...
+            {
+                if frequency_mhz <= 30 {
+                    FlashLatency::Latency0
+                } else if frequency_mhz <= 64 {
+                    FlashLatency::Latency1
+                } else if frequency_mhz <= 90 {
+                    FlashLatency::Latency2
+                } else {
+                    FlashLatency::Latency3
+                }
```
ppannuto, line 205:
```suggestion
                match frequency_mhz {
                    0..=30 => FlashLatency::Latency0,
                    31..=64 => FlashLatency::Latency1,
                    65..=90 => FlashLatency::Latency2,
                    _ => FlashLatency::Latency3,
                }
```

I can't believe I'm asking this, but, would this be more a idiomatic way to express this in Rust?

chips/stm32f4xx/src/chip_specific.rs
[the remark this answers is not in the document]
bradjc: > So basically leave this as-is, only as a function called from the `flash.rs` rather than a trait on the Flash object?

Yeah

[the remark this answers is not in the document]
ppannuto: Okay, so I looked at this again, decided it wouldn't work because the flash_specific needed to call back into the flash object to read the right register, but then decided that could work, and ultimately just did it to check myself

https://github.com/tock/tock/commit/a9b8c6d2359af8a9d3e96d2c7dca5b92029f7236 — I think this is what @bradjc is getting at. I think this is a bit cleaner, and I like that everything that is specific to a given flash instance is cleanly scoped behind `flash_specific::`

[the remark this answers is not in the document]
Ioan-Cristian: I don't like the fact that `chip_specific.rs` depends on `flash.rs`, but if you prefer this solution, I'm happy with it.

chips/stm32f4xx/src/rcc.rs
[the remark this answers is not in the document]
Ioan-Cristian: In theory (and in practice too I think), `::UNSUPPORTED` or `::INVALID` should never be returned. Should impossible invalid state be introduced in code? Also, I added a few comments to better explain the intent of the commented lines.

Even if you still prefer to introduce invalid state, wouldn't it be better to return  `Result`?

chips/stm32f4xx/src/chip_specific.rs
Ioan-Cristian: Yes, it is. I'm sorry for this!

chips/stm32f4xx/Cargo.toml
```diff
@@ -12,3 +12,9 @@ edition.workspace = true
 cortexm4 = { path = "../../arch/cortex-m4" }
 enum_primitive = { path = "../../libraries/enum_primitive" }
 kernel = { path = "../../kernel" }
+
+[features]
+stm32f401 = []
+stm32f412 = []
+stm32f429 = []
+stm32f446 = []
```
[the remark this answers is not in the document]
Ioan-Cristian: Added comments.

chips/stm32f4xx/src/flash.rs
Ioan-Cristian: I wasn't able to achieve this using `register_structs!` macro. Any help?

chips/stm32f4xx/src/chip_specific.rs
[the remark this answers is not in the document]
bradjc: This is what I am comfortable with: https://github.com/tock/tock/compare/pr/3528?expand=1

With those commits we only reference chips we claim to support and not variants outside of that list. We also only affirmatively use features so it is very clear exactly when certain code is being selected. With any `not()` when that code will get compiled is ambiguous. Also chips which we do not claim to support will not compile until they are explicitly added.

This also removes the conditionally included flash trait and replaces it with chip-specific functions. The flash driver then uses these functions to implement the functionality that was in the trait. This simplifies the code constructs in the chip_specific file to reduce complexity.

Note, I'm not doubting that the extended list of stm variants were correctly mapped to their various settings. But, in general I want to avoid giving the impression of more support than Tock actually offers, and I don't want there to be an example of including code/config that isn't ever used/tested so that others don't copy. If those settings for variants are hard to find then I think it would be fine to include them as comments for future developers who wish to add support for different variants and can then easily include them.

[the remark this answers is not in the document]
Ioan-Cristian: The problem with `#[cfg]` for `const` is that clippy tests don't pass anymore. I tried finding a solution for this without success. Any help?

I prefer the idea of commenting out unused features, but I'm also fine with removing them completely. They are not hard to find in documentation.

Personally, I don't like decoupling flash-related functionality from the `Flash` struct, but since you think it's clearer I'll implement it.

chips/stm32f4xx/src/rcc.rs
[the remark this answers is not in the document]
ppannuto: No.. I've come around to how you did this, the extra case just adds complexity for a situation that won't really happen.

chips/stm32f4xx/src/chip_specific.rs
ppannuto: Hah, no reason to apologize — I was laughing at myself because I only ever _get_ comments like that re: idiomatic Rust. I think this is the first time I've ever been able to give one correctly. Only took me eight or so years, but I guess I finally grok the Rust ethos 😅

chips/stm32f4xx/src/chip_specific.rs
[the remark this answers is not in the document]
bradjc: I added a commit which makes `cargo test` pass for me.

This isn't germane to this issue I guess but the fact that we run into these kind of issues continues to suggest to me that features continue to be a hindrance and should be avoided. The fact that the tests passed before not because of any sort of principled selection of hardware, with intentional testing, but instead because of which branch of the `if` statements happened to be selected is worrying to me.

chips/stm32f4xx/src/clocks/clocks.rs
```diff
@@ -0,0 +1,926 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright OxidOS Automotive SRL.
+//
+// Author: Ioan-Cristian CÎRSTEA <ioan.cirstea@oxidos.io>
+
+#![deny(dead_code)]
+#![deny(missing_docs)]
+#![deny(unused_imports)]
```
bradjc, line 9: Do we need the dead_code and unused_imports? Aren't those already on?

chips/stm32f4xx/src/flash.rs
ppannuto: Ohh... interesting... nrf5x gpio, temperature, etc don't use the `register_structs!` macro — they have hand-written register structs (likely because they predate that macro...). Frankly, it is _hard_ to read the `cfg`'d struct with the mixed chip definitions as they get bigger... it looks like your latest code here just went with the 'hand-written register struct' approach. I think that is a fine stopgap, especially for a small device like this, but long-term this is probably something that `register_structs!` should figure out how to support (at risk of making that macro even more complex... :/ )

bradjc:
As a status update, I'm waiting on either commits from https://github.com/tock/tock/compare/pr/3528?expand=1 or something similar to be added. Please let me know if I shouldn't be.

chips/stm32f4xx/src/clocks/clocks.rs
Ioan-Cristian: For me, they are not on by default. Any missing documentation and unused imported are checked during CI. However, I prefer that compilation fails before I can even make `git push`. Any new future developer can simply switch the `#![deny]` to `#![warning]` if they found it annoying or trying to develop something quickly.

Ioan-Cristian:
> As a status update, I'm waiting on either commits from https://github.com/tock/tock/compare/pr/3528?expand=1 or something similar to be added. Please let me know if I shouldn't be.

Sorry for being inactive for so long!

I pushed some of your suggestions, but not the one regarding `cfg!()` macro. Using `#[cfg]` attributes makes `cargo clippy` fail. One additional issue with the attribute version is that a documentation comment must be added for each variant. With macro version, only one comment is necessary. Do you still want to use attributes?

Ioan-Cristian:
I wonder if we should exclude `stm32f4xx` crate from `tools/list_chips.sh`? In this way, no default feature will be required. And after all, it is not a chip by itself, just a common dependency for STM32F4 family. Compiling `stm32f429zi` implies compiling `stm32f4xx` for instance.

I find default feature really ugly and unnatural. What do you think?

bradjc:
> I wonder if we should exclude `stm32f4xx` crate from `tools/list_chips.sh`? In this way, no default feature will be required. And after all, it is not a chip by itself, just a common dependency for STM32F4 family. Compiling `stm32f429zi` implies compiling `stm32f4xx` for instance.
> 
> I find default feature unnatural. What do you think?

That makes sense to me. I personally think cargo/rust features are a hack, so doing other hacks to make them work for real-world use cases doesn't seem any more strange to me. Cargo/rust have been very clear they do not care to support embedded use cases properly with features so we will always have to work around them.

I don't have any issue with setting a default. That is the same as the if..else... pattern does: whatever falls through gets set. At least with a default feature it is explicit and consistent, and not whatever was in the else block.

Copying documentation seems like a pretty minor issue to me compared to the complexity of reasoning about if statements wrapped with #ifdefs. And there should not be else cases because that violates the idea that we are being explicit and proactive about exactly which code is included.

Ioan-Cristian:
Removed `stm32f4xx` crate from `list_chips.sh` output and replaced `cfg!()` macros with `#[cfg]` attributes. Also, added internal variables to avoid the need to duplicate documentation comments. Is that fine for you?

Ioan-Cristian:
In order to pass CI tests, I found two solutions:

1. Using `#[cfg]` attributes and dirty hacks: https://github.com/OxidosAutomotive/tock/compare/sysclk-stm32f4xx...Ioan-Cristian:tock:sysclk-stm32f4xx-syntax?expand=1
2. Using `cfg!()` macros and panics at compile time if no feature is selected: https://github.com/OxidosAutomotive/tock/compare/sysclk-stm32f4xx...Ioan-Cristian:tock:sysclk-stm32f4xx-cfg-macros?expand=1

@bradjc Even though you seem to dislike `cfg!()` macros, I found them much clearer and there's no need to use dirty hacks. What do you think? Am I missing something?

Note: default feature doesn't help in the first case.

chips/stm32f4xx/src/chip_specific.rs
```diff
@@ -0,0 +1,166 @@
...
+        {
+            match frequency_mhz {
+                0..=30 => FlashLatency::Latency0,
+                31..=60 => FlashLatency::Latency1,
+                61..=90 => FlashLatency::Latency2,
+                91..=120 => FlashLatency::Latency3,
+                121..=150 => FlashLatency::Latency4,
+                _ => FlashLatency::Latency5,
+            }
+        }
```
bradjc, line 127: For example, this is the default setting for `get_number_wait_cycles_based_on_frequency`, correct? So if a user sets no features, they get these values, or if a user tries a new board with a different stm32f4 variant and sets a different feature they get these values? Are these the values for every stm32f4 variant (now and into the future) not called "stm32f412"? If the datasheet says so then that is good enough for me. If STM changes the datasheet in the future we can't realistically plan for that.

The purpose of our crate-based approach is we don't have these ambiguities. Everything in the stm32f4xx crate should apply to all stm32f4 chips. Everything in the stm32f446re crate only needs to apply to that specific variant. If there is anything in the stm32f4xx crate that could be used on the wrong variant that is a bug in my opinion.

Ioan-Cristian: I agree with you. Check the second link I sent you. I think it solves this issue. The only place where features are not explicitly stated is `PLL_MIN_FREQ_MHZ`, since only STM32F401 supports 13MHz as the minimum PLL frequency. All the other chips in the family support 24MHz.

brghena (approved):

brghena:
This seems to me to be on the right path, and is clearly a high-quality effort.

It would be good to get a summary of the remaining outstanding issues. There's something about the cfg and CI interactions? But I'm not exactly following. Is that the only thing?

chips/stm32f4xx/src/chip_specific.rs
ppannuto: I think the approach in the second link looks pretty reasonable. I'm not sure I have a strong opinion between `#[cfg` and `cfg!(`. The biggest negative of the latter is that it can 'hide' in the middle of lines (`if (...blah && cfg!(..`), but when contained to the `chip_specific` file already, where it's more sane to be on the lookout for `cfg` stuff, I think it is a reasonable solution.

Ioan-Cristian: I understand your concerns about `cfg!()` macros. However, used in `chip_specific.rs` and in simple if statements without any logical operators, they are easy to understand and easy to write, at least for me. Also, they don't require any dirty hacks to pass CI tests (or probably I missunderstand how to use `#[cfg]` attributes).

Ioan-Cristian:
> It would be good to get a summary of the remaining outstanding issues. There's something about the cfg and CI interactions? But I'm not exactly following. Is that the only thing?

Yes, that's the only thing. We're trying to figure out how to use either `#[cfg]` attributes or `cfg!()` macros to:

1. Avoid code duplication since clocks are shared between the entire F4 family.
2. Avoid obfuscating code.

The main problems are:

1. Impossibility to use `register_structs!()` macro. Personally, I couldn't make it work with `#[cfg]` attribute.
2. Some CI tools like `cargo clippy` activate all features at once or none of them, so the code fails to pass some tests, even though it is correct and compiles. Currently, this is the biggest hurdle.
3. Whether `#[cfg]` attributes or `cfg!()` macros are easier to understand.
4. How to prevent code from compiling without any feature enabled. Two solutions were found so far: default feature and use of  `cfg!()` macros in if-else statements with panics.

I hope I didn't forget any other problem.

ppannuto:
I think the `cfg!(`-based approach, _so long as `cfg!` appears only in the `chip_specific.rs` file_, is the best solution we have here and think we should move forward with that. @bradjc you're the other voice in the room here, does that sound good to you?

Ioan-Cristian:
@bradjc Sorry if I am insistent, but have you decided which approach is better?

chips/stm32f4xx/src/flash.rs
```diff
@@ -0,0 +1,458 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright OxidOS Automotive SRL.
+
+#![deny(dead_code)]
+#![deny(missing_docs)]
+#![deny(unused_imports)]
```
bradjc, line 7: Remove.

chips/stm32f4xx/src/chip_specific.rs
```diff
@@ -0,0 +1,166 @@
...
+    // The number of wait cycles depends on two factors: system clock frequency and the supply
+    // voltage. Currently, this method assumes 2.7-3.6V voltage supply (default value).
+    // TODO: Take into the account the power supply
+    //
+    // The number of wait states varies from chip to chip.
+    pub(crate) fn get_number_wait_cycles_based_on_frequency(frequency_mhz: usize) -> FlashLatency {
+        // feature = "stm32f401"
+        // feature = "stm32f429"
+        // feature = "stm32f446"
+        #[cfg(not(feature = "stm32f412"))]
```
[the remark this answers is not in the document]
ppannuto: I think that's the issue with the `#[cfg`-based approach as-opposed to the `cfg!(`-based approach.

From https://github.com/tock/tock/pull/3528#issuecomment-1691034340, do you prefer option (1) [which I believe is currently what this code attached to this PR does] or option (2) [which is linked from [this comment](https://github.com/tock/tock/pull/3528#issuecomment-1684269209)]. I think (2) is better and would alleviate this problem.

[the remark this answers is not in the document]
bradjc: For (2), do we need the ifs? Can we just have a default thing that happens which is the panic? Sort of like how the arch crate has a bunch of empty functions which only show up if the correct function can't be included because we're compiling for a host machine.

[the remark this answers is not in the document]
Ioan-Cristian: Are you talking about functions like this one:
```rust
#[cfg(not(any(target_arch = "arm", target_os = "none")))]
pub unsafe extern "C" fn initialize_ram_jump_to_main() {
    unimplemented!()
}
```

The problem with this approach is that code compiles. The code will panic only if the `unimplemented!()` macro is called at runtime. Not sure if this is the desirable behaviour.

Ioan-Cristian:
Hello!

As discussed last Friday, I reimplemented this pull request using type parameters and trait bounds. What do you think about it? @bradjc @ppannuto 

Note that documentation is still missing. I'll fix that once the changes are approved.

Also, tests still rely on features. Probably I'll reimplement them using the same strategy as for the implementation itself.

chips/stm32f4xx/src/chip_specific/chip_specs/mod.rs
```diff
@@ -0,0 +1,15 @@
...
+
+//! This module contains specification for different chips in the STM32F4 family
+
+pub mod chip_specs;
+pub mod stm32f401;
+pub mod stm32f412;
+pub mod stm32f429;
+pub mod stm32f446;
+
+pub use chip_specs::*;
```
bradjc, line 15: Please expand to remove the `*`.

chips/stm32f4xx/src/chip.rs
```diff
@@ -12,28 +12,35 @@ use kernel::platform::chip::InterruptService;
...
     pub i2c1: crate::i2c::I2C<'a>,
+    pub clocks: crate::clocks::Clocks<'a, ChipSpecs>,
     pub spi3: crate::spi::Spi<'a>,
     pub tim2: crate::tim2::Tim2<'a>,
     pub usart1: crate::usart::Usart<'a, dma::Dma2<'a>>,
     pub usart2: crate::usart::Usart<'a, dma::Dma1<'a>>,
     pub usart3: crate::usart::Usart<'a, dma::Dma1<'a>>,
-    pub gpio_ports: crate::gpio::GpioPorts<'a>,
-    pub fsmc: crate::fsmc::Fsmc<'a>,
+    _marker: PhantomData<ChipSpecs>,
```
bradjc, line 40:
```suggestion
```

We don't need the `PhantomData`s.

chips/stm32f4xx/src/clocks/clocks.rs
```diff
@@ -0,0 +1,927 @@
...
+
+/// Main struct for configuring on-board clocks.
+pub struct Clocks<'a, ChipSpecs> {
+    rcc: &'a Rcc,
+    flash: OptionalCell<&'a Flash<ChipSpecs>>,
+    /// High speed internal clock
+    pub hsi: Hsi<'a>,
+    /// Main phase loop-lock clock
+    pub pll: Pll<'a, ChipSpecs>,
+    _marker: PhantomData<ChipSpecs>,
```
bradjc, line 183:
```suggestion
```

Same here.

chips/stm32f4xx/src/flash.rs
```diff
@@ -0,0 +1,508 @@
...
+    optkeyr: WriteOnly<u32>,
+    /// Status register
+    sr: ReadWrite<u32, SR::Register>,
+    /// Control register
+    cr: ReadWrite<u32, CR::Register>,
+    /// Flash option control register
+    optcr: ReadWrite<u32, OPTCR::Register>,
+    /// Flash option control register 1
+    #[cfg(feature = "stm32f429")]
+    optcr1: ReadWrite<u32>,
```
bradjc, line 64:
```suggestion
    optcr1: ReadWrite<u32, OPTCR1::Register>,
```

tools/list_chips.sh
```diff
@@ -8,5 +8,5 @@
 for b in $(find chips -maxdepth 4 -name 'Cargo.toml'); do
     b1=${b#chips/}
     b2=${b1%/*}
-    echo $b2
+    echo ${b2}
```
bradjc, line 11: Do we need this change?

bradjc (commented):
Wow this is great.

One question, should the chip-specific implementations go in the specific chip crates? Is there a benefit to centralizing them?

Ioan-Cristian:
> Wow this is great.
> 
> One question, should the chip-specific implementations go in the specific chip crates? Is there a benefit to centralizing them?

Hmm, good question. I assumed that you wanted them to be centralized just like as with features. There might be a small benefit from implementing them in the `stm32f4xx` crate: the respective modules can be made `pub(crate)`, exposing to the other crates just `ChipSpecs` trait. This can help in hiding implementation details giving more freedom to change them in the future. However, both solutions seem fine to me. As a side note, I first implemented them in the chip-specific crates.

bradjc:
I think from the perspective of "if I'm updating/creating a specific variant, I shouldn't have to change the shared crate" it would make sense to me to put the chip-specific content in the specific crates.

tools/list_chips.sh
Ioan-Cristian: Removed curly braces.

chips/stm32f4xx/src/flash.rs
Ioan-Cristian: Added type parameter.

chips/stm32f4xx/src/clocks/clocks.rs
Ioan-Cristian: Removed type marker.

chips/stm32f4xx/src/chip.rs
Ioan-Cristian: Removed type marker.

chips/stm32f4xx/src/chip_specific/chip_specs/mod.rs
Ioan-Cristian: Removed wild card import.

bradjc (approved):

PR #3532, opened 2023-07-02
capsules: remove doc leftovers on `SyscallDriver` impls

elenaf9:
### Pull Request Overview

In #2639 `subscribe` was remove from the `SyscallDriver` trait, and in #2906 also `allow_readwrite` and `allow_readonly` were removed.
Both PRs however left some of the code docs related to the removed functions. I am not sure if that was intentional, but it confused me quite a bit when I was looking into the code.

This PR proposes to remove these old docs. But I am quite new to tock, so maybe I am missing something and there was a reason for keeping them?

### Testing Strategy

_Not relevant.__

### TODO or Help Wanted

Open Questions:
- Is any information lost by removing the docs? 
- Are there any other related docs that should be updated?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.
<details><summary>Fails due to an unrelated error in `kernel/src/process_standard.rs:535:13` </summary>

```sh

   --> kernel/src/process_standard.rs:535:13
    |
530 |         Some(storage_permissions::StoragePermissions::new(
    |              -------------------------------------------- arguments to this function are incorrect
...
535 |             write_id,
    |             ^^^^^^^^ expected `Option<NonZeroU32>`, found `Option<u32>`
    |
    = note: expected enum `Option<NonZeroU32>`
               found enum `Option<u32>`
note: associated function defined here
   --> kernel/src/storage_permissions.rs:49:19
    |
49  |     pub(crate) fn new(
    |                   ^^^
...
54  |         write_id: Option<NonZeroU32>,
    |         ----------------------------

```
 </details>

### DIFF
--- a/capsules/core/src/console.rs
+++ b/capsules/core/src/console.rs
@@ -235,25 +235,6 @@ impl<'a> Console<'a> {
 }
 
 impl SyscallDriver for Console<'_> {
-    /// Setup shared buffers.
-    ///
-    /// ### `allow_num`
-    ///
-    /// - `1`: Writeable buffer for read buffer
-
-    /// Setup shared buffers.
-    ///
-    /// ### `allow_num`
-    ///
-    /// - `1`: Readonly buffer for write buffer
-
-    // Setup callbacks.
-    //
-    // ### `subscribe_num`
-    //
-    // - `1`: Write buffer completed callback
-    // - `2`: Read buffer completed callback
-
     /// Initiate serial transfers
     ///
     /// ### `command_num`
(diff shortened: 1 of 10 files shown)


bradjc:
We did intentionally keep them. Even though the capsules are no longer implementing the logic for storing the process buffers, the contract between userspace and the capsule for how the buffers are used is still implemented in the capsule.

Now, the rewrite in those PRs was fairly substantial, and we didn't put really any effort in to what to do with those comments, other than not delete them. If there is some way we would keep the documentation but reduce the confusion that would be ideal.

elenaf9:
> If there is some way we would keep the documentation but reduce the confusion that would be ideal.

:+1: I'll try to come up something.

alevy:
Thanks @elenaf9! The current documentation is indeed confusing now. Perhaps there is a way to refer to some central documentation of the semantics of each high-level system call (`allow_read`, `subscribe`, etc) but keeping the specific use of each subdriver number in the drivers themselves, by convention.

bradjc:
I think we should just move these to the top of the files.

bradjc:
Ok this should be ready to go.

PR #3533, opened 2023-07-04
Small licence checker enhancements

Ioan-Cristian:
### Pull Request Overview

This pull request adds a red error message when errors are encountered. This
way, it's more obvious whether the licence checker failed or not. Additionally,
a banner is printed every time the test is launched.

### Testing Strategy

This pull request was manually tested.

### TODO or Help Wanted

No help wanted.

### Documentation Updated

- No updates are required.

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/tools/license-checker/src/main.rs
+++ b/tools/license-checker/src/main.rs
@@ -87,11 +87,11 @@
 
 #![allow(rustdoc::invalid_rust_codeblocks)]
 
+use colored::ColoredString;
+use colored::Colorize;
 use ignore::WalkBuilder;
 use std::path::{Path, PathBuf};
 use std::process::exit;
-
-use colored::Colorize;
 
 mod parser;
 use parser::{Cache, LineContents, ParseError, Parser};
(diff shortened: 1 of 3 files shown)


alevy (dismissed):
lgtm. @jrvanwhy ? (I think he might be on vacation, in which we can probably move ahead without waiting for him, and revert if there are problems later).

tools/license-checker/src/main.rs
```diff
@@ -91,6 +91,8 @@ use ignore::WalkBuilder;
 use std::path::{Path, PathBuf};
 use std::process::exit;
 
+use colored::Colorize;
```
jrvanwhy, line 94: I think you're trying to separate `use std` from non-`std` `use` statements, but they're all already clumped together anyway (see `use ignore::WalkBuilder`). Lets combine `use colored::...` with the above block of `use` statements.

tools/license-checker/src/main.rs
```diff
@@ -115,24 +117,32 @@ struct Args {
     verbose: bool,
 }
 
+const ERROR_MESSAGE: &str = "error:";
+
+macro_rules! bright_red_and_bold_error_message {
+    () => {
+        ERROR_MESSAGE.bright_red().bold()
+    };
+}
```
jrvanwhy, line 126: I'm having trouble seeing why this is a macro. Can it just be a function? Something like:

```rust
fn error_prefix() -> ColoredString {
    "error:".bright_red().bold()
}
```

Note that I intentionally shortened the name as well -- the function name was as long as its body!

tools/license-checker/src/main.rs
Ioan-Cristian: Moved use statements.

tools/license-checker/src/main.rs
Ioan-Cristian: Replaced macro with function.

jrvanwhy (approved):

PR #3534, opened 2023-07-04
Fix negative padding reports in print_tock_memory_usage.py

nbdd0121:
### Pull Request Overview

`print_tock_memory_usage.py` reports negative paddings due to overlapping symbol spans.

Symbols can overlap, e.g. when two identical functions are merged (which can happen to Rust functions because their addresses are not significant) or two address-not-significant constants are merged due to one being a subslice to another.

This PR changes it to only consider the symbol with largest size when multiple overlapping symbols are present.

### Testing Strategy

```make -C boards/opentitan/earlgrey-cw310/ memory```

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/tools/print_tock_memory_usage.py
+++ b/tools/print_tock_memory_usage.py
@@ -345,6 +345,13 @@ def group_symbols(groups, symbols, waste, section):
     waste_sum = 0
     prev_symbol = ""
     for (symbol, addr, size, real_size, _) in symbols:
+        if addr < expected_addr:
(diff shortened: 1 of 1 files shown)


lschuermann (approved):

nbdd0121:
Looks like spurious failure in the CI.

lschuermann:
> Looks like spurious failure in the CI.

Indeed, restarted.

alevy (approved):

PR #3535, opened 2023-07-07
tock-cells: change `OptionalCell::map` to pass a copy into the closure

lschuermann:
### Pull Request Overview

This pull request changes the semantics of `OptionalCell::map` to move a copy of the internal `Option`'s value into the provided closure. Proving a mutable reference to a copy that is then immediately discarded is very confusing.

An alternative design would remove the `map` APIs and add a `get()` method, which returns an `Option<T>`. Then we can use the `map*` methods on that returned type.

### Testing Strategy

### TODO or Help Wanted

This pull request still needs porting of the rest of the kernel to this new interface. This should be mostly mechanical changes.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/capsules/core/src/adc.rs
+++ b/capsules/core/src/adc.rs
@@ -336,7 +336,7 @@ impl<'a, A: hil::adc::Adc<'a> + hil::adc::AdcHighSpeed<'a>> AdcDedicated<'a, A> 
         let mut app_buf_length = 0;
         let exists = self.processid.map_or(false, |id| {
             self.apps
-                .enter(*id, |_, kernel_data| {
+                .enter(id, |_, kernel_data| {
                     app_buf_length = kernel_data
                         .get_readwrite_processbuffer(0)
                         .map(|b| b.len())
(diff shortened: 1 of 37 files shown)


kupiakos:
On the call, we also discussed renaming the `extract` field to `get`. Should that be in a different PR or this one? I brought up my reasoning for calling it `get` [on the original PR that named it `extract`](https://github.com/tock/tock/pull/2531#issuecomment-1625703637).

lschuermann:
@kupiakos It's simple enough to be tacked onto here, but given there's already a history behind that (#2531, #2531), having the discussion in a separate PR seems reasonable. Happy to be the one to take the blame here and creating the PR, given I initially introduced `get()` and then also renamed it `extract()`. :laughing: 

Apart from that, I do think that these changes are mostly orthogonal, right?

lschuermann:
I implemented this change throughout the Tock codebase. Most of these changes have been done mechanically through `sed(1)` magic, and then verified in the git diff.

Unsurprisingly, the diff confirms that we've been dereferencing the closure parameter almost everywhere, implicitly copying it. This confirms that this change more closely reflects the intended behavior.

The actual change to `OptionalCell` is quite small; here's a link to avoid having to scroll through the diff: https://github.com/tock/tock/pull/3535/files#diff-8e0f1d8c375dc7a41a4ffd6511e8029fc478ba6425c80cd455710f54e2b75188L188-R210

lschuermann:
For context, this has been brought up by @kupiakos on the call today, and was also the source of confusion for some OpenTitan developers.

hudson-ayers (approved):
Yeah, this is definitely better

bradjc (approved):
Yeah I never understood why we needed the `*` there and just did what the compiler told me to.

PR #3536, opened 2023-07-07
tock-cells/OptionalCell: rename extract-method to get

lschuermann:
### Pull Request Overview

In response to https://github.com/tock/tock/pull/2531#issuecomment-1625703637, this renames `OptionalCell::extract` to `OptionalCell::get`. Hence this reverts commit 3136f00a77173708625256011be54ee24778364f.

I agree that `extract` is extremely confusing in this context -- we're operating on a cell, and `extract` could reasonably suggest that it removes the value from the `OptionalCell`. This is amplified by the fact that `take` (the natural method name for a "destructive get" in `TakeCell`) implies move-semantics, which we don't have in `OptionalCell<T>`, as it requires that `T: Copy`. Even though I originally made the `get -> extract` change, I still get this wrong.

Now, do I want to open this can of worms again? I don't know. I still believe we need a method on `OptionalCell` whose signature is `fn somename(&self) -> Option<T>`. Is `get` the right name? At least some evidence (in @kupiakos) seems to suggest that it's not entirely atypical for a `get` method to return an `Option<T>`, and it is the only interpretation of `get` that makes sense given `OptionalCell`'s semantics. On the other hand, I think the concerns over this name as raised on the initial PR (#2531) are fair.

### Testing Strategy

### TODO or Help Wanted

This pull request still needs bike-shedding of the method name.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/libraries/tock-cells/src/optional_cell.rs
+++ b/libraries/tock-cells/src/optional_cell.rs
@@ -154,8 +154,8 @@ impl<T> OptionalCell<T> {
 }
 
 impl<T: Copy> OptionalCell<T> {
-    /// Returns a copy of the contained [`Option`].
-    pub fn extract(&self) -> Option<T> {
+    /// Returns a copy the contained [`Option`].
+    pub fn get(&self) -> Option<T> {
         self.value.get()
     }
 
--- a/libraries/tock-cells/src/optional_cell.rs
+++ b/libraries/tock-cells/src/optional_cell.rs
@@ -154,7 +154,35 @@ impl<T> OptionalCell<T> {
 }
 
 impl<T: Copy> OptionalCell<T> {
-    /// Returns a copy the contained [`Option`].
+    /// Returns a copy of the contained [`Option`].
+    //
+    // This was originally introduced in PR #2531 [1], then renamed to `extract`
+    // in PR #2533 [2], and finally renamed back in PR #3536 [3].
+    //
+    // The rationale for including a `get` method is to allow developers to
+    // treat an `OptionalCell<T>` as what it is underneath: a `Cell<Option<T>>`.
+    // This is useful to be interoperable with APIs that take an `Option<T>`, or
+    // to use an *if-let* or *match* expression to perform case-analysis on the
+    // `OptionalCell`'s state: this avoids using a closure and can thus allow
+    // Rust to deduce that only a single branch will ever be entered (either the
+    // `Some(_)` or `None`) branch, avoiding lifetime & move restrictions.
+    //
+    // However, there was pushback for that name, as an `OptionalCell`'s `get`
+    // method might indicate that it should directly return a `T` -- given that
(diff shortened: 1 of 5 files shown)


kupiakos:
> the concerns over this name as raised on the initial PR (https://github.com/tock/tock/pull/2531) are fair

For reference, I've [replied](https://github.com/tock/tock/pull/2531#issuecomment-1625703637) to those concerns on that PR.

> `take` (the natural method name for a "destructive get" in `TakeCell`) implies move-semantics, which we don't have in `OptionalCell<T>`, as it requires that `T: Copy`

[`OptionalCell<T>::take`](https://github.com/tock/tock/blob/ee3973f/libraries/tock-cells/src/optional_cell.rs#L140-L142) is already present and does not require `T: Copy`. It doesn't need to - `take` semantics don't requiring cloning, just like how [`core::cell::Cell<T>::take`](https://doc.rust-lang.org/std/cell/struct.Cell.html#method.take) doesn't require `T: Copy` as its [`impl` block only requires `T: Default`](https://doc.rust-lang.org/std/cell/struct.Cell.html#impl-Cell%3CT%3E-3)

hudson-ayers (approved):
I am on board with this based on the discussion on the call

bradjc (approved):
At least personally what I always do is 1. call .get() 2. compiler error 3. google optionalcell 4. look for the function that has the correct signature 5. oh right call extract(). So, I guess I'm in favor of reverting.

libraries/tock-cells/src/optional_cell.rs
```diff
@@ -154,8 +154,8 @@ impl<T> OptionalCell<T> {
 }
 
 impl<T: Copy> OptionalCell<T> {
-    /// Returns a copy of the contained [`Option`].
-    pub fn extract(&self) -> Option<T> {
+    /// Returns a copy the contained [`Option`].
+    pub fn get(&self) -> Option<T> {
```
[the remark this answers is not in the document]
lschuermann: Sounds good, will add!

libraries/tock-cells/src/optional_cell.rs
```diff
@@ -155,6 +155,34 @@ impl<T> OptionalCell<T> {
 
 impl<T: Copy> OptionalCell<T> {
     /// Returns a copy the contained [`Option`].
```
brghena, line 157:
```suggestion
    /// Returns a copy of the contained [`Option`].
```

brghena (approved):

libraries/tock-cells/src/optional_cell.rs
lschuermann: Thanks!

brghena (approved):

bradjc (approved):

ppannuto (approved):

PR #3537, opened 2023-07-08
Fix MapCell safety 2.0

kupiakos:
### Pull Request Overview

MapCell is unsound if mutated inside of `map`, since it invalidates the `&mut T` in the closure.
It also fails to drop its contents if put is called.

This fixes the soundness hole by making these re-entrant mutations silently do nothing, while panicking with debug assertions to inform unit tests of a bug.

More rustdoc has also been added. This replaces pull #3325 - see that for further context.

`MapCell` is like a `Cell<Option<T>>` with an extra state and check to allow `map` to be sound against re-entrancy.

- `map` marks the cell as borrowed while the closure is running, in order to detect re-entrancy.
- `map` does not call its closure if currently inside `map`, like it does for the empty state.
- `take` returns `None` and does not mutate if currently inside `map`, unlike the empty state.
- `put` drops its input and does not mutate if currently inside `map`, unlike the empty state.
- `replace` drops its input, does not mutate, and returns `None` if currently inside map, unlike the empty state.
- A `get` function has been added that returns `Option<T>` and returns `None` if currently inside `map`, like it does for the empty state.

### Testing Strategy

I have not tested this on hardware. This PR will be primarily tested by CI unless someone chips in.

### TODO or Help Wanted

The /docs for `MapCell` could use some updating - [what it says about comparing to `TakeCell` is confusing](https://github.com/tock/tock/blob/master/doc/Mutable_References.md#mapcell).

A way to test the release behavior of mutating inside of a `map` while inside of a unit test would be helpful - right now it's solely `should_panic` tests.

### Documentation Updated

### Formatting

- [x] ~Ran `make prepush`.~ Ran `cargo fmt` and `cargo clippy`

### DIFF
--- a/libraries/tock-cells/src/map_cell.rs
+++ b/libraries/tock-cells/src/map_cell.rs
@@ -2,61 +2,179 @@
 
 use core::cell::{Cell, UnsafeCell};
 use core::mem::MaybeUninit;
-use core::ptr;
+use core::ptr::drop_in_place;
 
-/// A mutable memory location that enforces borrow rules at runtime without
-/// possible panics.
+#[derive(Clone, Copy, PartialEq)]
+enum MapCellState {
+    Uninit,
+    Init,
+    Borrowed,
+}
+
+#[inline(never)]
+#[cold]
+fn access_panic() {
+    panic!("`MapCell` accessed while borrowed");
+}
+
+macro_rules! debug_assert_not_borrowed {
+    ($slf:ident) => {
+        if cfg!(debug_assertions) && $slf.occupied.get() == MapCellState::Borrowed {
+            access_panic();
+        }
+    };
+}
+
+/// A mutable, possibly unset, memory location that provides checked `&mut` access
+/// to its contents via a closure.
 ///
-/// A `MapCell` is a potential reference to mutable memory. Borrow rules are
-/// enforced by forcing clients to either move the memory out of the cell or
-/// operate on a borrow within a closure. You can think of a `MapCell` as an
-/// `Option` wrapped in a `RefCell` --- attempts to take the value from inside a
-/// `MapCell` may fail by returning `None`.
+/// A `MapCell` provides checked shared access to its mutable memory. Borrow
+/// rules are enforced by forcing clients to either move the memory out of the
+/// cell or operate on a `&mut` within a closure. You can think of a `MapCell`
+/// as a `Cell<Option<T>>` with an extra "in-use" state to prevent `map` from invoking
+/// undefined behavior when called re-entrantly.
+///
+/// # Examples
+/// ```
+/// # use tock_cells::map_cell::MapCell;
+/// let cell: MapCell<i64> = MapCell::empty();
+///
+/// assert!(cell.is_none());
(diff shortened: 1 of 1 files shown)


libraries/tock-cells/src/map_cell.rs
```diff
@@ -124,22 +263,33 @@ impl<T> MapCell<T> {
...
+            // `occupied` is reset to initialized at the end of scope,
+            // even if a panic occurs in `closure`.
+            struct ResetToInit<'a>(&'a Cell<MapCellState>);
+            impl Drop for ResetToInit<'_> {
+                #[inline(always)]
+                fn drop(&mut self) {
+                    self.0.set(MapCellState::Init);
+                }
+            }
+            let _reset_to_init = ResetToInit(&self.occupied);
```
hudson-ayers, line 286: While this unwind-safety is nice (especially for unit tests), the Tock kernel has generally been designed with the assumption that the kernel will be built with panic=abort, and I don't think we aim to guarantee sound operation if unwinding panics are used / caught. And this presumably does come at some size cost, so may be worth considering whether it makes sense to include. 

That said, I don't want this PR to get held up on a general discussion of whether we should provide unwind-safety in Tock.

hudson-ayers (approved):
This looks inline with the direction we agreed on in the call, and the size results are good!

libraries/tock-cells/src/map_cell.rs
kupiakos: > And this presumably does come at some size cost

The size cost should only be there if it needs to catch an unwind to reset the state. This is only true if you have `-Cpanic=unwind` _and_ the compiler cannot prove the closure is panic-less. Otherwise it's the same as writing the normal `self.occupied.set(MapCellState::Init)` - note the `#[inline(always)]` to ensure this is the case. In other words, it won't affect Tock users that write `-Cpanic=abort` but will help users in unit tests and systems with `-Cpanic=unwind`.

When it comes to fundamental data structures like this, I think we should lean towards caution here for unwind-safety. Otherwise, it would cause missed drop glue and weird behavior as it would then forever think it is still borrowed - fortunately not unsoundness though.

bradjc (approved):
I think it might be helpful to comment on what helps with size reductions, but if you don't think we are likely to accidentally undo it then not a big deal.

kupiakos:
> comment on what helps with size reductions

Compared to the original, or with the refactors in #3325? There's one fewer branch in `replace` compared to the original, and it tries to change as few instructions as possible to achieve safety. If I get the chance I'd also like to experiment with using `-1` and `1` as discriminants, since checking `<0`, `>0`, and `==0` require fewer instructions than `==1` or `==2` (as there's no immediate to load). I specifically kept `Uninit` as `0` so `OptionalCell::empty()` can be stored in `.bss`.

When compared to the refactors in #3325 this saves space by doing less. There's no smart pointer involved, and there's one fewer state for users to branch on. I also tried to consolidate branches for uninit and init mutation where possible.

ppannuto (approved):

lschuermann (approved):

PR #3539, opened 2023-07-11
Switch to v0 mangling scheme and port print_tock_memory_usage.py

### LINKED ISSUE #3529: Migrate from legacy mangling scheme to v0 scheme
Tock currently uses the default symbol mangling scheme, which is the legacy scheme at the moment, but will eventually be moved to the v0 mangling scheme.

The v0 scheme contains more information (most notably, generic arguments), e.g. this is the demangling in v0:
```rust
<<capsules_core::led::LedDriver<kernel::hil::led::LedHigh<lowrisc::gpio::GpioPin>, 8> as kernel::syscall_driver::SyscallDriver>::allocate_grant>
```
where this is for legacy:
```rust
<<capsules_core::led::LedDriver<L,_> as kernel::syscall_driver::SyscallDriver>::allocate_grant>
```

There are a few scripts that currently rely on the legacy scheme, e.g. `print_tock_memory_usage.py` and `find_panics.py` which would need to be migrated to support the v0 scheme. Some scripts, e.g. `stack_analysis.sh` on the other rely on v0 scheme.

We should migrate entirely away from legacy scheme to the v0 mangling scheme.

Rust tracking issue for the v0 mangling scheme: https://github.com/rust-lang/rust/issues/60705

nbdd0121:
### Pull Request Overview

This PR switches the symbol mangling scheme to v0. I also ported the `print_tock_memory_usage.py` script to parse the demangled symbol from v0 scheme.

Fix #3529

### Testing Strategy

`make -C boards/opentitan/earlgrey-cw310/ memory`

### TODO or Help Wanted

`tools/find_panics.py` currently does not work

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/Makefile.common
+++ b/boards/Makefile.common
@@ -60,6 +60,9 @@ TARGET_DIRECTORY ?= $(TOCK_ROOT_DIRECTORY)target/
 #   and the downside to `all` is that different functions in the code can end up
 #   with the same address in the binary. However, it can save a fair bit of code
 #   size.
+# - `-C symbol-manging-version=v0`: Opt-in to Rust v0 symbol mangling scheme.
+#   See https://github.com/rust-lang/rust/issues/60705 and
+#   https://github.com/tock/tock/issues/3529.
 RUSTC_FLAGS ?= \
   -C link-arg=-Tlayout.ld \
   -C linker=rust-lld \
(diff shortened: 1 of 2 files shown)


boards/Makefile.common
```diff
@@ -61,6 +61,7 @@ TARGET_DIRECTORY ?= $(TOCK_ROOT_DIRECTORY)target/
 #   with the same address in the binary. However, it can save a fair bit of code
 #   size.
 RUSTC_FLAGS ?= \
+  -C symbol-mangling-version=v0 \
```
bradjc, line 64: Can you add this at the end and update the comment?

PR #3540, opened 2023-07-11
kernel/platform/mpu: remove dependence on ProcessID and assumptions

lschuermann:
### Pull Request Overview

The current MPU configuration trait assumes that implementations will perform an optimization to avoid unnecessary re-configuring the MPU for a process for which it is already configured. To enable this optimization, it passes the `ProcessId` into `MPU::configure_mpu`.

This approach has multiple potential issues. First and foremost, it assumes that a Tock process will only ever use a single `MPUConfig` type. This assumption can be trivially violated, as `MPUConfig: Default`. Furthermore, it seems like a practical optimization to store different, pre-computed MPU configurations for a single process, for instance when performing IPC operations with different processes. The MPU would then silently not be configured if a process were to not mark the respective region as dirty (actually, it can't, given that this is an implementation-specific private field).

Additionally, the requirement to pass a `ProcessId` makes it impossible to use the MPU trait to configure memory protection for services which do not fit the standard Tock process model. Naturally, this is a more exotic limitation.

To fix the aforementioned issue, this commit removes the `Default` trait requirement from the `MpuConfig` type, and instead adds a method to the `MPU` trait to allocate a new `MpuConfig`. This method can then assign unique IDs to the returned `MpuConfig` structs, which enable internally optimizing MPU configuration when applying an identical and unmodified `MpuConfig` multiple times in a row. It thus removes the implicit dependency on `ProcessId` and assumptions on the behavior of Tock's `Process`es. `MpuConfig` types now only need to be internally consistent.

### Testing Strategy

This pull request was tested by compiling / CI.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


bradjc (dismissed):
This seems a lot better. Given your explanation not sure why we didn't think of this originally.

Why `NonZeroUsize` instead of `usize`?

lschuermann:
> Why `NonZeroUsize` instead of `usize`?

This shaves some words off of RAM usage, as we commonly wrap these types in an `Option` / `OptionalCell`, which is then free in terms of memory (provided the structs start with this as their first field).

ppannuto (dismissed):

lschuermann:
Rebased to resolve conflicts.

I've also changed `new_config` to return an `Option<Self::MpuConfig>` instead of a `Self::MpuConfig` directly. This is because the `MpuConfig` types will likely be generated from some exhaustible counter, and thus at some point fail new allocation requests. 

While I don't expect Tock to actually reach this counter in practice, and we have other exhaustible resources with similar semantics (e.g., `ProcessId`), it is good practice to communicate this clearly in the API. This moves the `.unwrap()`-panic up from the MPU implementations into the `ProcessStandard`, which could gracefully handle this case eventually.

Removing *last-call* for this change.

bradjc (changes requested):
I definitely understand your reasoning, but I would rather not see `unwrap()` in the process implementation. My quick search looks like the file is currently unwrap free.

I think improperly handling the case where we create 4 billion mpu configurations is better than the code precedent of using unwrap.

alistair23 (dismissed):

bradjc:
The other option is the mpuconfig id is not _always_ unique (ie wrap back to zero). I'm partial to that option because it's simple. But, if there is a way for an app to force new mpuconfigs to be created continuously, then perhaps we should avoid it.

Really I think I'm good with either solution you proposed.

lschuermann:
Already started implementing my proposed solution. :)

I think this implementation is decently elegant, and captures the semantics of our MPU implementations well:
- the `ProcessStandard` implementation stays free of (explicit) panics
- the `reset_config` method is effectively a more elegant version of `clear`. I don't know why `clear` exists at all, and it violates our otherwise consistent paradigm that the only way to change the currently active MPU configuration is through `configure_mpu`, `enable_app_mpu` and `disable_app_mpu`. It has particularly unclear semantics when the MPU is further used for kernel memory protection. I vouch to eventually remove it, as `reset_config` followed by `configure_mpu` should cover all use-cases.
- when using an MPU as a means to enforce memory protection in both the kernel and userspace, the MPU configuration may fail in a variety of ways, which we cannot possibly exhaustively cover through an error enum. The newly introduced `ProcessLoadError::MpuConfigurationError` can cover these rare resource exhaustion events, and other MPU-specific internal errors.

bradjc (approved):

ppannuto (approved):

alistair23 (approved):

bradjc:
Marking this last call again.

PR #3547, opened 2023-07-17
Capsule: AT24C32/64 support

j-kk:
### Pull Request Overview

This change add capsule for handling AT24C32/64 EEPROM read/write operations via NonvolatileStorageComponent with NonvolatileToPages

### Testing Strategy

This capsule was tested on raspberry pi pico with AT24C32 connected. Data consistency was verified with an existing library in Arduino SDK.

### TODO or Help Wanted

None

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


capsules/extra/src/eeprom.rs
```diff
@@ -0,0 +1,259 @@
+//! Driver for the AT24C32/64 EEPROM memory. Built on top of the I2C interface.
```
bradjc, line 1: Since we don't know how general this is, can we call this file `at24c32.rs`?

capsules/extra/src/eeprom.rs
```diff
@@ -0,0 +1,259 @@
...
+                self.i2c.disable();
+                self.flash_client.map(|client| {
+                    if let Some(client_page) = self.client_page.take() {
+                        if status.is_err() {
+                            client.write_complete(client_page, hil::flash::Error::FlashError);
+                        } else {
+                            client.write_complete(client_page, hil::flash::Error::CommandComplete);
+                        }
+                    } else {
+                        panic!("EEPROM Writing op completed without client_page");
```
bradjc, line 210: I don't think we need this panic. It's pretty clear that client_page will be there.

capsules/extra/src/eeprom.rs
```diff
@@ -0,0 +1,259 @@
...
+                self.buffer.replace(local_buffer);
+                self.i2c.disable();
+                Err((error.into(), buf))
+            } else {
+                self.client_page.replace(buf);
+                Ok(())
+            }
+        } else {
+            Err((ErrorCode::RESERVE, buf))
+        }
```
brghena, line 113: Shouldn't this branch also disable the `i2c`? Or is the absence of a buffer to `take()` a sign that a transaction is in progress?

A separate solution, which might be better, is to only enable after taking the buffer.

This applies to `read_sector()`, `write_sector()`, and `erase_sector()`.

capsules/extra/src/eeprom.rs
```diff
@@ -0,0 +1,259 @@
...
+//!
+//! Usage
+//! -----
+//!
+//! ```rust
+//! let i2cmux = I2CMuxComponent::new(i2c0, None).finalize(components::i2c_mux_component_static!());
+//!
+//! let eeprom_buffer = static_init!([u8; 34], [0; 34]);
+//!
+//! let eeprom_i2c_device = static_init!(I2CDevice, I2CDevice::new(i2cmux, 0x50));
```
brghena, line 23: `0x50` should be a named constant in this file. Like: https://github.com/tock/tock/blob/master/capsules/extra/src/bmp280.rs#L21

We have not been good about doing this generally though, and many drivers do exactly what you did here.

brghena (commented):
Overall, this is great. A few small comments, but this is almost good to go.

capsules/extra/src/eeprom.rs
j-kk: right, however I'd stay without 32 at the end
when it comes to the capsule (struct) name I'm still thinking about `at24c_eeprom` or sth like that

brghena: Either `at24c_eeprom` or something like `at24cxx_eeprom` seems totally reasonable. For the filename too.

capsules/extra/src/eeprom.rs
j-kk: `0x50` just specifies the i2c address so it's more depends on that how user connects it's eeprom, that is why I refrained from that

j-kk: What is more concerning to me is buffer initialisation - it should always have size `PAGE_SIZE + 2` so we can define it inside the `new()`. Is there a way to initialise a buffer without referencing to kernel?

capsules/extra/src/eeprom.rs
j-kk: tbh this should be inside the `if`, after taking the buffer, moved

capsules/extra/src/eeprom.rs
j-kk: same behaviour is in lines 194-196, but without panic, removed that

bradjc (approved):

capsules/extra/src/at24c_eeprom.rs
```diff
@@ -0,0 +1,259 @@
...
+pub struct AT24C<'a> {
+    i2c: &'a dyn I2CDevice,
+    buffer: TakeCell<'static, [u8]>,
+    client_page: TakeCell<'a, EEPROMPage>,
+    flash_client: OptionalCell<&'a dyn hil::flash::Client<AT24C<'a>>>,
+    state: Cell<State>,
+}
+
+impl<'a> AT24C<'a> {
+    pub fn new(i2c: &'a dyn I2CDevice, buffer: &'static mut [u8]) -> Self {
```
brghena, line 85: Does `buffer` always need to be `PAGE_SIZE+2` here? If so, does `buffer: &'static mut [u8; PAGE_SIZE+2]` compile? Similarly the `TakeCell` can hold a `[u8; PAGE_SIZE+2]`.

I know that works with hard-coded numbers, as other things in Tock do it: https://github.com/tock/tock/blob/master/kernel/src/process_checker/basic.rs#L119

j-kk: Yes, buffer always need to hold `PAGE_SIZE+2` bytes, `PAGE_SIZE` for the data, 2 bytes for addressing

bradjc (approved):

brghena (approved):

PR #3550, opened 2023-07-18
kernel/platform/mpu: don't provide default impls for methods

lschuermann:
### Pull Request Overview

Providing default implementations for the various `kernel::mpu::MPU` methods makes it easy to forget a method in an implementation, and would not cause compiler errors when adding a new method that should be implemented. Default implementations should really only be used when it is legal for an implemenation to not provide a specialized method and fall back onto the default. This is not the case for any of the `MPU` trait methods.

This instead moves those method implementations to the unit type, which implements an `MPU` for boards that don't have one present.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/platform/mpu.rs
+++ b/kernel/src/platform/mpu.rs
@@ -100,13 +100,13 @@ pub trait MPU {
     /// MPU where possible.
     /// On some hardware it is impossible to reset the MPU after it has
     /// been locked, in this case this function wont change those regions.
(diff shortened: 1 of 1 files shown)


bradjc (approved):
Agreed. I think this was just a way to get the `()` version to compile.

ppannuto (approved):

PR #3551, opened 2023-07-20
KV: Separate virtualization from KV implementation, New KV trait, Set/Add/Update

bradjc:
### Pull Request Overview

This pull request updates a fair amount of the KV stack.

- Create new HILs: `kv:KV` and `kv:KVPermissions`
- Split the kv_store layer into three capsules:
   1. The kv_store capsule which provides `kv:KV` on top of TicKV.
   2. The kv_store_permissions capsule which provides `kv:KVPermissions` on top of kv_store.
   3. The virtual_kv capsule wich implements virtualization.
   
    This brings the kv stack more in line with how tock capsules are implemented.
- Remove the `kv_system` HIL.

Critically, this change fixes an issue where the kv_store layer could call a callback from a downcall if the first kv operation (in this case hashing the key) failed. Now all immediate errors are reported synchronously and only errors after an interrupt are returned via callback.

#### HIL Updates

This PR gives us a standard KV interface:
- `get(key) -> value`: Lookup a key
- `set(key, value)`: Set the key to val, overwriting if needed.
- `add(key, value)`: Set the key to val if key does not exist.
- `update(key, value)`: Replace the existing value.
- `delete(key)`: Remove key-value object.

To separate our permissions/access control and to give KV stack users flexibility, we have both regular and permissions versions.

We also no longer assume any hashed keys in the HIL. This avoids the issue where hashes could collide.

By separating get/add/update we can now have implementations which will not lose data on a set(). Note, however, that with the current tickv implementation we cannot guarantee that after a set either the new data or the old data will be present.

### Testing Strategy

Tested with kv-interactive app.

### TODO or Help Wanted

- [x] Formalize the error returns from the KV interface.
  - The HIL now specifies the return ErrorCodes and the code matches them.
- [x] Add SIZE error on callbacks if key/value too big to store.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


capsules/extra/src/kv_store.rs
```diff
@@ -79,88 +78,178 @@ impl KeyHeader {
...
+    /// ### Return
+    ///
+    /// - On success returns `Ok(())`. A callback will be issued.
+    /// - On error, returns the buffers and:
+    ///   - `ENOSUPPORT`: The key could not be found.
+    ///   - `SIZE`: The value is longer than the provided buffer. The amount of
+    ///     the value that fits in the buffer will be provided.
+    fn get(
+        &self,
+        key: LeasableMutableBuffer<'static, u8>,
```
ppannuto, line 154: Is there a reason to call this `key` instead of `unhashed_key`? That expectation used to be embedded in the name, and now it's just in the comment above. What was the rationale for just `key` (and changing the `unhashed_key`s earlier)? Seems like it's adding extra cognitive load for the user?

capsules/extra/src/virtual_kv.rs
```diff
@@ -0,0 +1,349 @@
...
+
+        // // Create the Tock header.
+        // let header = KeyHeader {
+        //     version: HEADER_VERSION,
+        //     length: (value.len() - HEADER_LENGTH) as u32,
+        //     write_id,
+        // };
+
+        // // Copy in the header to the buffer.
+        // header.copy_to_buf(value.as_slice());
```
ppannuto, line 158: Is this intentionally left as comment?

capsules/extra/src/virtual_kv.rs
```diff
@@ -0,0 +1,349 @@
...
+
+        self.operation.set(Operation::Set);
+        self.valid_ids.set(permissions);
+        self.unhashed_key.replace(key);
+        self.value.replace(value);
+        // self.mux_kv.do_next_op();
+        // Ok(())
```
ppannuto, line 165:
```suggestion
```

?

capsules/extra/src/virtual_kv.rs
```diff
@@ -0,0 +1,349 @@
...
+    ) -> Result<(), (LeasableMutableBuffer<'static, u8>, Result<(), ErrorCode>)> {
+        if self.operation.is_some() {
+            return Err((key, Err(ErrorCode::BUSY)));
+        }
+
+        self.operation.set(Operation::Delete);
+        self.valid_ids.set(permissions);
+        self.unhashed_key.replace(key);
+        // self.mux_kv.do_next_op();
+        // Ok(())
```
ppannuto, line 189:
```suggestion
```

ppannuto (changes requested):
I think this is a good change overall. It was easier to think through some of the new state progression than the old, largely because it matches what my mind is already trained on for our virtualizer pattern. Just a few small things I caught.

capsules/extra/src/kv_store.rs
bradjc: I would put it the other way: is there a reason to call it `unhashed_key`? Describing something in terms of what it isn't adds its own burden: why would I have thought it was that in the first place? Perhaps `unhashed_uncompressed_key`?

capsules/extra/src/virtual_kv.rs
bradjc: No that it is a mistake. I intend to fix.

ppannuto (approved):

bradjc:
I'm working on this PR, new version soon.

kernel/src/hil/kv.rs
```diff
@@ -0,0 +1,468 @@
...
+    /// operation.
+    fn header_size(&self) -> usize;
+}
+
+/// Key-Value interface.
+///
+/// This interface provides access to key-value storage where access control.
+/// Each object is marked with a `write_id` (based on the `StoragePermissions`
+/// used to create it), and all further accesses and modifications to that
+/// object require suitable permissions.
```
alistair23, line 328: This comment seems out of date.

capsules/extra/src/tickv.rs
```diff
@@ -39,12 +33,165 @@
...
 use tickv::{self, AsyncTicKV};
 
+/// The type of keys, this should define the output size of the digest
+/// operations.
+pub trait KeyType: Eq + Copy + Clone + Sized + AsRef<[u8]> + AsMut<[u8]> {}
+
+impl KeyType for [u8; 8] {}
+
+/// Implement this trait and use `set_client()` in order to receive callbacks.
+pub trait KVSystemClient<K: KeyType> {
```
alistair23, line 48: I don't really follow why this is moved from the HIL. Won't other backend implementations want something similar?

bradjc:
I just realized I need to add a SIZE error on the set callbacks in case the value or key was too large to store.

bradjc:
> I just realized I need to add a SIZE error on the set callbacks in case the value or key was too large to store.

Done.

alevy:
> The kv_store capsule which provides kv:KV on top of TicKV.

Seems like this should just be called "TickV" or something, since presumably the idea is that another capsule would implement the same trait for a different underlying implementation.

alevy:
> The virtual_kv capsule wich implements virtualization.

What _kind_ of virtualization?

* Presumably to other capsules, not processes.
* Is it actually virtualized (i.e. the key space is somehow namespaced or something), or just shared (i.e. many capsules can have outstanding operations to the same KV store)?

alevy:
> Create new HILs: kv:KV and kv:KVPermissions

Probably shouldn't _actually_ be a HIL, but that's getting into potentially moving a bunch of traits out of HILs and into... I'm not entirely sure where

capsules/extra/src/tickv.rs
```diff
@@ -39,12 +33,165 @@
...
+    fn set_client(&self, client: &'a dyn KVSystemClient<Self::K>);
+
+    /// Generate key.
+    ///
+    /// - `unhashed_key`: A unhashed key that should be hashed.
+    /// - `key_buf`: A buffer to store the hashed key output.
+    ///
+    /// On success returns nothing.
+    /// On error the unhashed_key, key_buf and `Result<(), ErrorCode>` will be returned.
+    fn generate_key(
```
alevy, line 111: I think this is a broad general design question about a KV store interface and implementation, so maybe it shouldn't be part of this PR, _but_...

I don't understand as either an implementer or user of this interface what `generate_key` is supposed to do. Is the resulting key guaranteed to be unique for a given input key? Do I always need to use it? Is it allowed to be the identity function?

Also, why not return the resulting key instead of passing in a mutable "buffer," c-style? In particular, if it's being hashed I expect `Self::K` to _always_ be fixed size (either a usize, or a fixed-length array or something)

alevy (approved):
I broadly support these changes as improvements from the status quo. I think this interface could benefit from a more wholistic design though

bradjc:
> > The kv_store capsule which provides kv:KV on top of TicKV.
> 
> Seems like this should just be called "TickV" or something, since presumably the idea is that another capsule would implement the same trait for a different underlying implementation.

I agree. I can add that to the todo list, I just didn't do that because changing file names can make PRs even harder to review.

> > The virtual_kv capsule wich implements virtualization.
> 
> What _kind_ of virtualization?
> 
>     * Presumably to other capsules, not processes.

Yes, same as all other `virtual_X.rs` crates.

>     * Is it actually virtualized (i.e. the key space is somehow namespaced or something), or just shared (i.e. many capsules can have outstanding operations to the same KV store)?

Access to the resource is virtualized, the namespace is not.



> I broadly support these changes as improvements from the status quo. I think this interface could benefit from a more wholistic design though

Do you mean the `KV` HIL or something else?

capsules/extra/src/tickv.rs
bradjc: > I think this is a broad general design question about a KV store interface and implementation, so maybe it shouldn't be part of this PR, _but_...

I think KV interface questions are within scope. The PRs that have already been merged got things to a working state.

> I don't understand as either an implementer or user of this interface what `generate_key` is supposed to do. Is the resulting key guaranteed to be unique for a given input key? Do I always need to use it? Is it allowed to be the identity function?

Currently no guarantees of uniqueness. You don't have to use it in the sense that if you already have a `K` type object than you can use that, but somehow you need to get `K`. It can be any function.

> Also, why not return the resulting key instead of passing in a mutable "buffer," c-style? In particular, if it's being hashed I expect `Self::K` to _always_ be fixed size (either a usize, or a fixed-length array or something).

I think the expectation is that it could be passed to a hardware peripheral.

libraries/tickv/src/async_ops.rs
```diff
@@ -324,7 +324,7 @@ impl<'a, C: FlashController<S>, const S: usize> AsyncTicKV<'a, C, S> {
                     .tickv
                     .append_key(self.key.get().unwrap(), &value[0..value_length]);
                 self.value.replace(Some(value));
-                (ret, 0)
+                (ret, value_length)
```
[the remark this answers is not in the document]
bradjc: I did update the comment.

alevy:
> Do you mean the KV HIL or something else?

I think the method signatures in the HIL are probably close to right, but the actual semantics aren't very spelled out (and when we spell them out, we might actually want slightly different method signatures).

I think there is also a question of where and how to virtualize (not saying that the current PR is wrong, just that I actually don't *exactly* know what it does or whether it's right).

Finally, specifically where things live is probably part of a bigger revisit of the hil module in the kernel.

alevy:
> Access to the resource is virtualized, the namespace is not.

I think I would technically call this a multiplexer (or a mux), not a virtualizer, but meh... It's what I assumed and I think it's fine as is.

capsules/extra/src/tickv.rs
alevy: @alistair23 I think this PR is basically redefining which interface it makes sense for alternative backends to target.

A client of KVSystem (in this PR `tickv_kv_store` spends much of it's effort basically translating keys between high-level keys (byte strings) and TickV-keys using the asynchronous `generate_key` API which, I think, is very specific to TicKV. I suspect alternative back-ends are unlikely to use a similar mechanism for indexing the datastore, so this interface just wouldn't make sense for them.

Of course, if that proves wrong, and `tickv_kv_store.rs` would otherwise be just copy-pasted for a different KV store we can (a) revisit this or (b) move the trait somewhere more shared.

alevy (dismissed):

bradjc:
> > > The kv_store capsule which provides kv:KV on top of TicKV.
> > 
> > 
> > Seems like this should just be called "TickV" or something, since presumably the idea is that another capsule would implement the same trait for a different underlying implementation.
> 
> I agree. I can add that to the todo list, I just didn't do that because changing file names can make PRs even harder to review.

I changed this, so tickv-specific code is in capsules with filenames `tickv_X.rs`.

bradjc:
@alistair23 This function doesn't work: https://github.com/tock/tock/blob/master/libraries/tickv/src/tickv.rs#L219. It doesn't really make sense as it casts a negative number to a usize and then checks if it is greater than some number.

The return value isn't used in `get_key()` any way, causing an infinite loop. This is only a problem when the starting index is 0 because otherwise `increment_region_offset()` returns `None`.

There are several issues here and I'm not sure how to fix it. How did you intend for this to work?

bradjc:
> There are several issues here and I'm not sure how to fix it. How did you intend for this to work?

I spent a couple minutes trying to formalize the search but because tickv is sync it is difficult to keep track of the states and what needs to have access to the current search progress.

@alistair23 Is this something you can fix? If not, I'm not sure what to do. I think maybe we just delete the promise that tickv will search for new pages and if a key doesn't fit on its assigned page then we return an error. That's what is happening now, except for region 0 which if there is 1 key stored then tickv infinite loops.

bradjc:
I added several bug fixes after testing with userspace. Related to https://github.com/tock/libtock-c/pull/337

alevy (approved):

PR #3553, opened 2023-07-21
usb_hid: disambiguate RESERVE/BUSY

ppannuto:
### Pull Request Overview

The send path has two possible failures
 1. Userspace forgot to allow a buffer to send from
 2. The driver already sending something

Previously, we returned ERESERVE in both cases, but (2) should be EBUSY instead, which this corrects.

I believe this will explain https://github.com/tock/libtock-c/pull/333

### Testing Strategy

Reading and reasoning through behavior.

### TODO or Help Wanted

I'm pretty certain this is correct, but after this goes in, @brghena can you re-test https://github.com/tock/libtock-c/pull/333 and see if you're (my guess) in the EBUSY case?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/extra/src/usb_hid_driver.rs
+++ b/capsules/extra/src/usb_hid_driver.rs
@@ -212,7 +212,7 @@ impl<'a, U: usb_hid::UsbHid<'a, [u8; 64]>> SyscallDriver for UsbHidDriver<'a, U>
                             .and_then(|send| {
                                 send.enter(|data| {
                                     self.send_buffer.take().map_or(
-                                        CommandReturn::failure(ErrorCode::RESERVE),
+                                        CommandReturn::failure(ErrorCode::BUSY),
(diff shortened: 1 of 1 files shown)


bradjc (approved):

alistair23 (approved):

brghena (approved):
Yes, this changes the error to EBUSY, as you guessed.

It still doesn't explain _why_ it's busy, or why it still works even though it's busy. That debugging should almost certainly be focused on the `libtock-c/libtock/usb_keyboard_hid.c` driver though.

bradjc:
One thing that could be happening is the _receive_ callback could be firing perhaps, which then triggers an upcall causing the yield to (mistakenly) finish when the libtock-c code should be waiting on the send done callback.

lschuermann (approved):

PR #3554, opened 2023-07-21
storage permissions: clarify naming

bradjc:
### Pull Request Overview

Tock has a notion of storage permissions that traces from the TBF header to a type in the kernel crate. The naming changes from module-to-module, and this PR attempts to make it consistent.

- Remove the "persistent acl" terminology. Replace with "storage permissions". This is more clear that it relates to storage and not something else that is persistent.
- Remove "access id" and replace with "modify id". I think this is more clear, as access and read are similar, but really having an access id gives the app modify permissions.
- Make the StoragePermissions object in the kernel consistent. Before it had two write ids, the app's write id and a list of write ids the app could modify. That list is now modify ids.

### Testing Strategy

This is only a cosmetic change and so verified by the compiler.

### TODO or Help Wanted

n/a

If this is good I will update elf2tab and tockloader to match.

### Documentation Updated

- [x] Updated doc/tbf.md

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/TockBinaryFormat.md
+++ b/doc/TockBinaryFormat.md
@@ -15,7 +15,7 @@
     + [`3` Package Name](#3-package-name)
     + [`5` Fixed Addresses](#5-fixed-addresses)
     + [`6` Permissions](#6-permissions)
-    + [`7` Persistent ACL](#7-persistent-acl)
+    + [`7` Storage Permissions](#7-storage-permissions)
     + [`8` Kernel Version](#8-kernel-version)
     + [`9` Program](#9-program)
     + [`128` Credentials Footer](#128-credentials-footer)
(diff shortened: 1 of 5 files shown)


brghena (approved):

lschuermann (approved):

ppannuto (approved):
This falls a bit into the name churn bin, which I'd normally nudge to deferring to a major version bump with other messiness, but I think this is worth an exception as it's a key user-facing aspect of the upcoming tutorial.

alistair23 (approved):

bradjc:
> This falls a bit into the name churn bin, which I'd normally nudge to deferring to a major version bump with other messiness, but I think this is worth an exception as it's a key user-facing aspect of the upcoming tutorial.

I disagree, it's quite confusing as the name changes with every layer. We also _specifically_ marked this portion of code as experimental with the exact intention we would change it without a major version because we weren't sure how it would work.

Issue #1739, opened 2020-04-06
allocate_app_memory_region needs to enforce alignment


ppannuto:
This came up in #1648: https://github.com/tock/tock/pull/1648#discussion_r402822448

Basically, this is working out fine right now because all the MPU region allocators do allocate regions larger than the pointer size (which is certainly unlikely to change). Nonetheless, we have an unexpressed requirement here that we should capture.

ia0:
This is only working out fine right now as long as a custom MPU implementation is used. The default implementation does not enforce any alignment.

Here's the bug (undefined behavior when the default implementation of `allocate_app_memory_region` is used):
https://github.com/tock/tock/blob/6e0aeb0328a114775ce301bac48682173a54a4bd/kernel/src/process_standard.rs#L1666-L1679

Here's the default implementation breaking the invariant the unsafe code above relies on:
https://github.com/tock/tock/blob/6e0aeb0328a114775ce301bac48682173a54a4bd/kernel/src/platform/mpu.rs#L221-L240

Also, it looks like the documentation of `allocate_app_memory_region` does not guarantee alignment, so the code in `ProcessStandard::create()` is actually assuming things it should not.

hudson-ayers:
Did you hit this bug in practice? If so I am curious to see how. I guess just using the default MPU config on a board with no MPU?

I agree that we should update the documentation for allocate_app_memory_region() and the default implementation to guarantee these constraints.

ia0:
> Did you hit this bug in practice? If so I am curious to see how. I guess just using the default MPU config on a board with no MPU?

Yes, I've hit this bug in practice (the kernel panics with an Unaligned Access Usage Fault). Indeed, I'm using the default MPU config (the board could have an MPU, but the layout constraints imposed by Tock and the MPU are too restrictive). I don't have a minimal reproduction example, but to get you an idea here's how one could be forged:

```rust
static BUSY: AtomicBool = AtomicBool::new(false);
static COUNTER: AtomicU32 = AtomicU32::new(0);
pub fn start_work(...) {
    assert!(!BUSY.swap(true, SeqCst)); // must not already be running
    COUNTER.fetch_add(1, SeqCst);
    ...
}
pub fn stop_work(...) {
    assert!(BUSY.load(SeqCst)); // must be running
    ...
    BUSY.store(false, SeqCst);
}
```

[...]

Issue #3544, opened 2023-07-17
RISC-V PMP region allocation overflows provided memory section


lschuermann:
The RISC-V PMP implementation of the `MPU::allocate_region` method is implemented as follows:
https://github.com/tock/tock/blob/d0e7132d7b641e0b8b4cd7207e02db474bb28bd3/arch/rv32i/src/pmp.rs#L360-L376

It starts with the kernel-provided parameters for allocating the memory region, and then proceeds to adjust them to meet PMP TOR start and end address alignment constraints. In practice, TOR regions require a start and end address aligned to a 4-byte boundary, so an unaligned start address can shift the region's start by at most 3 bytes, and an unaligned size can shift the region's end address by another 3 bytes.

Unfortunately, after making these adjustments, the implementation does not check whether the resulting region is still fully contained in the provided _unallocated memory_ section (`start + size < unallocated_memory_start + unallocated_memory_size`).

When loading a process, its flash-region is allocated with the MPU as follows:
https://github.com/tock/tock/blob/d0e7132d7b641e0b8b4cd7207e02db474bb28bd3/kernel/src/process_standard.rs#L1473-L1487

Because the kernel does not have any freedom in prepending or appending padding to application flash, it provides an "unallocated memory" region of the exact size of the application binary in flash. In this `app_flash.len()`-size region, it requests an allocation of `app_flash.len()` bytes. The only way for this allocation to work is if the region's start and end address are compatible with the aforementioned PMP alignment constraints.

Unfortunately, many apps today don't have 4-byte alignment of their end-address. Furthermore, applications could potentially request a _protected region_ size which causes their start address to not be aligned on a 4-byte boundary (although linker script section alignment constraints may mask this). Combined with the current implementation, depending on app binary alignment, this means that we are providing an app access to at most 6 bytes of flash beyond its accessible memory section, and risk not granting an application access to its first 3 bytes.

We can't fix this without breaking existing applications, as the PMP alignment constraints are hardware requirements. On a breaking release, we should likely fix this issue by requiring applications to adhere to the PMP constraints, and properly implementing the PMP region allocation bounds checks _after_ adjusting the region for hardware constraints.

In the short term, we should ensure that applications build with current toolchains do meet the requirement alignment constraints, to minimize breakage when fixing this bug in the kernel.

PR #3556, opened 2023-07-21
kernel/process_standard: ensure kernel_memory_break align & cleanup

lschuermann:
### Pull Request Overview

This commit addresses issue tock/tock#1739 by ensuring that the kernel memory break is always well-aligned to a word-boundary, regardless of the MPU's internal alignment constraints.

It further cleans up the memory allocation code. Tracing the previous slices and pointers through the previous code was rather difficult. In an effort to simply this logic, the returned memory is now continuously sub-sliced until all individual memory sub-sections have been isolated into their own slices. This avoids indexing into larger slices, encompassing multiple logically-distinct memory allocations.

It further renames the allocation returned by `allocate_app_memory_region` to no longer have "app" in its name, as this can be mistaken for app-accessible memory. Instead, it simply calls it `allocated_memory`, which is then sub-sliced into `app_accessible_memory` and `allocated_kernel_memory` slices.

Visually, the memory is now sliced as follows:

    +-----------------------------------------------------------------
    | remaining_memory
    +----------------------------------------------------+------------
    v                                                    v
    +----------------------------------------------------+
    | allocated_padded_memory                            |
    +--+-------------------------------------------------+
       v                                                 v
       +-------------------------------------------------+
       | allocated_memory                                |
       +-------------------------------------------------+
       v                                                 v
       +-----------------------+-------------------------+
       | app_accessible_memory | allocated_kernel_memory |
       +-----------------------+-------------------+-----+
                                                   v
                                 kernel memory break
                                                    \---+/
                                                        v
                                          optional padding

### Testing Strategy

Needs testing. How to best test this? Deliberately have the dummy MPU (`()`) return a non word-aligned kernel break?

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/kernel/src/process_standard.rs
+++ b/kernel/src/process_standard.rs
@@ -1505,8 +1505,17 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
         // Initial size of the kernel-owned part of process memory can be
         // calculated directly based on the initial size of all kernel-owned
         // data structures.
-        let initial_kernel_memory_size =
-            grant_ptrs_offset + Self::CALLBACKS_OFFSET + Self::PROCESS_STRUCT_OFFSET;
+        //
+        // We require our kernel memory break (located at the end of the
+        // MPU-returned allocated memory region) to be word-aligned. However, we
+        // don't have any explicit alignment constraints from the MPU. To ensure
+        // that the below kernel-owned data structures still fit into the
+        // kernel-owned memory even with padding for alignment, add an extra
+        // `sizeof(usize)` bytes.
+        let initial_kernel_memory_size = grant_ptrs_offset
+            + Self::CALLBACKS_OFFSET
+            + Self::PROCESS_STRUCT_OFFSET
+            + core::mem::size_of::<usize>();
 
         // By default we start with the initial size of process-accessible
         // memory set to 0. This maximizes the flexibility that processes have
@@ -1586,9 +1595,17 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
             remaining_memory
         };
 
-        // Determine where process memory will go and allocate MPU region for
-        // app-owned memory.
-        let (app_memory_start, app_memory_size) = match chip.mpu().allocate_app_memory_region(
+        // Determine where process memory will go and allocate an MPU region.
+        //
+        // `[allocation_start, allocation_size)` will cover both
+        //
+        // - the app-owned `min_process_memory_size`-long part of memory (at
+        //   some offset within `remaining_memory`), as well as
+        //
+        // - the kernel-owned allocation growing downward starting at the end
+        //   of this allocation, `initial_kernel_memory_size` bytes long.
+        //
+        let (allocation_start, allocation_size) = match chip.mpu().allocate_app_memory_region(
             remaining_memory.as_ptr() as *const u8,
             remaining_memory.len(),
             min_total_memory_size,
@@ -1613,11 +1630,14 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
             }
         };
 
-        // Get a slice for the memory dedicated to the process. This can fail if
-        // the MPU returns a region of memory that is not inside of the
-        // `remaining_memory` slice passed to `create()` to allocate the
-        // process's memory out of.
-        let memory_start_offset = app_memory_start as usize - remaining_memory.as_ptr() as usize;
+        // Determine the offset of the app-owned part of the above memory
+        // allocation. An MPU may not place it at the very start of
+        // `remaining_memory` for internal alignment constraints. This can only
+        // overflow if the MPU implementation is incorrect; a compliant
+        // implementation must return a memory allocation within the
+        // `remaining_memory` slice.
+        let app_memory_start_offset =
+            allocation_start as usize - remaining_memory.as_ptr() as usize;
 
         // Check if the memory region is valid for the process. If a process
         // included a fixed address for the start of RAM in its TBF header (this
@@ -1625,7 +1645,7 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
         // need a fixed address) then we check that we used the same address
         // when we allocated it in RAM.
         if let Some(fixed_memory_start) = tbf_header.get_fixed_address_ram() {
-            let actual_address = remaining_memory.as_ptr() as u32 + memory_start_offset as u32;
+            let actual_address = remaining_memory.as_ptr() as u32 + app_memory_start_offset as u32;
             let expected_address = fixed_memory_start;
             if actual_address != expected_address {
                 return Err((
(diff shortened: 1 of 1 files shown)


kernel/src/process_standard.rs
```diff
@@ -1505,8 +1505,17 @@ impl<C: 'static + Chip> ProcessStandard<'_, C> {
...
+        // We require our kernel memory break (located at the end of the
+        // MPU-returned allocated memory region) to be word-aligned. However, we
+        // don't have any explicit alignment constraints from the MPU. To ensure
+        // that the below kernel-owned data structures still fit into the
+        // kernel-owned memory even with padding for alignment, add an extra
+        // `sizeof(usize)` bytes.
+        let initial_kernel_memory_size = grant_ptrs_offset
+            + Self::CALLBACKS_OFFSET
+            + Self::PROCESS_STRUCT_OFFSET
+            + core::mem::size_of::<usize>();
```
bradjc, line 1518: Why not just calculate the exact needed padding?

bradjc (commented):
It seems like you should copy your diagram in to the code.

I assume you can run the process command on the pconsole with a null app before and after to ensure all this doesn't change anything?

kernel/src/process_standard.rs
lschuermann: We don't know this ahead of time, it's up to the MPU where the end of the allocation is placed. We have three options:
- Add an additional "allocation end" alignment constraint to `allocate_app_memory_region`, where we can then specify that we want the end of the region word-size aligned. This seems sub-optimal, as this is not even a kernel-invariant, but rather one of `ProcessStandard`.
- Add `core::mem::size_of::<usize>() - 1` to the initial kernel memory size, which is the maximum amount of padding we'll ever need. However, with our current MPU implementations, this will mean that we'll *always* end on an unaligned address, and always waste those 3 bytes.
- Add `core::mem::size_of::<usize>()`. This is one byte more than required, but it would mean that an MPU implementation which naturally aligns regions to a word boundary places this padding also on a word boundary. We would have a slightly larger `initial_kernel_memory_size` than required, but wouldn't waste any of this memory, for instance when we do grant allocations.

bradjc: We have precedent for the first option with `grant`: we request both the size and the alignment when creating a grant. I think it's reasonable to say that alignment is a fundamental operation for allocation.

But I'm not sure I feel strongly about changing the interface for this purpose.

lschuermann: I'm all for adding alignment to the interface, but what feels weird is to add an alignment requirement towards the _end_ of a region. Usually the MPU performs alignment inserting padding in the front. In this case, we'd need the MPU to place the kernel region towards the end of the allocated section, then move it further towards the front to achieve alignment, while not colliding with the app's memory, and then passing back the allocated _start_ address (which isn't even subject to that alignment) to the caller.

Apparently, even our current interface has sufficient complexity implement it incorrectly (#3544), and adding all of the above constraints adds so many more edge-cases.

I don't like either, but for now I'd be leaning towards allocating another word of memory, which can be used by grants at runtime anyways.

ppannuto (dismissed):
I think this is _better_ than what we have currently, but doing a more thorough read through of the whole process creation I think TheRightWay(TM) would move more intelligence and work into the `userspace_kernel_boundary`.

Ultimately, alignment requirements derive from architecture specifications. To that end, all the little final padding, shifting, casting bits really should be in arch-specific implementations. As one example, `initial_app_brk` ultimately gets passed to `userspace_kernel_boundary.app_brk` where it's used as the initial process stack pointer. On ARM, this requires **8-byte** alignment. Nothing enforces this*. On the flip side, technically ARM Cortex-M3—M7 actually allow for unaligned access (with degraded runtime performance as the HW silently does the alignment), but if some more practical architecture had looser requirements, we would bake padding we don't need into processes.

In the short term, I think we should merge this, including the current "extra word of padding for alignment if-needed" solution. In the longer run, I'd like to play with a slightly more compartmentalized and federated architecture for process creation—which, at minimum, might break the current 500-line monolith that is `create()` into more manageable, conceptual pieces.

I'm also skeptical that a "default implementation" for the MPU should exist, especially if the MPU is to be responsible for things like defining region sizes. Indeed, the _real_ root cause IMO to the error reported on the original issue here is that the MPU region allocation did not follow rules for the currently executing architecture (which a default impl can't possibly do!). Maybe I'll see if it makes sense for that to go away as well.

*though it does happen to work out in our common case paths right now. It derives from `min_process_memory_size`, which comes `UKB.initial_process_app_brk_size` which is `SVC_FRAME_SIZE` which is 32 bytes. This is then added `app_mpu_mem_start`, which ultimately chains through a few MPU things to something aligned. Frankly, it's unlikely this wouldn't work out on real ARM implementations, but that isn't a case for hidden, cross-layer dependencies.

ppannuto:
Also, possibly of interest, removing the clippy allows doesn't generate clippy warnings anymore (on `master`, or this branch).

kernel/src/process_standard.rs
bradjc: Ok yes I agree. Given that we are allocating a larger region, it makes sense that the process standard impl just ensures it can both give the process enough accessible memory and make its grant bookkeeping aligned.

bradjc:
> would move more intelligence and work into the `userspace_kernel_boundary`.

I'm skeptical, but, building on the idea that we need `UKB.initial_process_app_brk_size` might make this better.

ppannuto:
Yeah, I'm not certain that UKB is the right way to do it either :/. I think we have a bunch of places where we assume alignment that in practice work out on all 'normal' systems, but host-based unit testing and folks exploring 64-bit arch's, etc are starting to expose / violate these assumptions.

I opened #3583 to flag/track the more substantial alignment issue.

I think we should merge this as-is as it fixes a real problem in the now and has some nice buffer management cleanups. I don't think this will be the final word here, but think the 'most right' fix is a much bigger undertaking.

bradjc (approved):
on master:

```
tock$ process bare_bones
𝐀𝐩𝐩: bare_bones   -   [Yielded]
 Events Queued: 0   Syscall Count: 1   Dropped Upcall Count: 0
 Restart Count: 0
 Last Syscall: Yield { which: 1, address: 0x20007000 }
 Completion Code: None
 Credential: None


 ╚0x20007800═╪══════════════════════════════════════════╝
             │ Grant Ptrs      120
             │ Upcalls         320
             │ Process         920
  0x200072B0 ┼───────────────────────────────────────────
             │ ▼ Grant           0
  0x200072B0 ┼───────────────────────────────────────────
             │ Unused
  0x20007020 ┼───────────────────────────────────────────
             │ ▲ Heap            ? |      ?               S
  ?????????? ┼─────────────────────────────────────────── R
             │ Data              ? |      ?               A
  ?????????? ┼─────────────────────────────────────────── M
             │ ▼ Stack           ? |      ?
  0x20007000 ┼───────────────────────────────────────────
             │ Unused
  0x20007000 ┴───────────────────────────────────────────
             .....
  0x00041000 ┬─────────────────────────────────────────── F
             │ App Flash      4016                        L
  0x00040050 ┼─────────────────────────────────────────── A
             │ Protected        80                        S
```

with this pr:

```
tock$ process bare_bones
𝐀𝐩𝐩: bare_bones   -   [Yielded]
 Events Queued: 0   Syscall Count: 1   Dropped Upcall Count: 0
 Restart Count: 0
 Last Syscall: Yield { which: 1, address: 0x20007000 }
 Completion Code: None
 Credential: None


 ╚0x20007800═╪══════════════════════════════════════════╝
             │ Grant Ptrs      120
             │ Upcalls         320
             │ Process         920
  0x200072B0 ┼───────────────────────────────────────────
             │ ▼ Grant           0
  0x200072B0 ┼───────────────────────────────────────────
             │ Unused
  0x20007020 ┼───────────────────────────────────────────
             │ ▲ Heap            ? |      ?               S
  ?????????? ┼─────────────────────────────────────────── R
             │ Data              ? |      ?               A
  ?????????? ┼─────────────────────────────────────────── M
             │ ▼ Stack           ? |      ?
  0x20007000 ┼───────────────────────────────────────────
             │ Unused
  0x20007000 ┴───────────────────────────────────────────
             .....
  0x00041000 ┬─────────────────────────────────────────── F
             │ App Flash      4016                        L
  0x00040050 ┼─────────────────────────────────────────── A
             │ Protected        80                        S
tock$
```

Looks good to me!

PR #3562, opened 2023-07-25
Add `capsules/tutorials` crate and add encryption oracle capsule

lschuermann:
### Pull Request Overview

This adds `capsules/tutorials/`, a new crate for capsules used in Tock tutorials. Capsules contained in this crate may not be particularly useful, or even complete (e.g., they can be "checkpoint" versions for following along in a tutorial), but they must compile. This ensures that breaking changes in upstream Tock will also propagate to kernel-code used in tutorials.

It further adds various checkpoints of the "encryption oracle" capsule, to be built by participants in the upcoming Tock World 6 tutorial.

### Testing Strategy

Running the last checkpoint of the `encryption_oracle` capsule (`chkpt5`) alongside the Tock book instructions together with the HOTP userspace app.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


brghena (dismissed):

alexandruradovici (approved):

boards/nordic/nrf52840_chip_layout.ld
```diff
@@ -5,8 +5,8 @@
 /* Memory Space Definitions, 1M flash, 256K ram */
 MEMORY
 {
-  rom (rx)  : ORIGIN = 0x00000000, LENGTH = 192K
-  prog (rx) : ORIGIN = 0x00030000, LENGTH = 832K
+  rom (rx)  : ORIGIN = 0x00000000, LENGTH = 256K
+  prog (rx) : ORIGIN = 0x00030000, LENGTH = 768K
```
brghena, line 9: I'm 99% certain that Amit reverted this. It made apps not work. And it's not in: https://github.com/tock/tock/compare/master...dev/software-hmac-integration

brghena (approved):

boards/nordic/nrf52840_chip_layout.ld
brghena: Oh, you addressed this on Slack. As long as it's intentional and apps still work, I'm fine with it.

bradjc:
I'm not sure I understand why create a new crate for tutorial capsules rather than a module in extra. Not only is it more overhead (we need another readme, and we really should update the docs to describe when we create new capsule crates), but it (currently) makes it impossible to share types between capsules extra and this crate (for example if you wanted to use the `KV` trait.

We have a precedent for non formal code in the extra crate (tests).

alexandruradovici:
> I'm not sure I understand why create a new crate for tutorial capsules rather than a module in extra. Not only is it more overhead (we need another readme, and we really should update the docs to describe when we create new capsule crates), but it (currently) makes it impossible to share types between capsules extra and this crate (for example if you wanted to use the `KV` trait.
> 
> We have a precedent for non formal code in the extra crate (tests).

I agree with this if the tutorial is for future Tock developers, as this will show them how to add a capsule.

brghena:
The split to a `tutorials/` directory is mostly my fault. I wanted some way to segment these partially working, redundant, tutorial capsules from "normal" capsules. So someone doesn't just look at "encryption_oracle.rs" as a totally valid thing.

The namespacing was what seemed important to me. Maybe a `tutorials/` folder inside `capsules/extra/`? Or we could add `tutorial` to the name of the capsules?

lschuermann:
I do like @brghena's idea of the extra crate, as it very clearly says "this is not production code, don't expect this to be working necessarily". A module in `extra` would work, but to me this semantic separation was a compelling reason to split it into a `tutorials` crate (in a similar spirit to how we separate `core` from `extra`). It also limits the TCB for board-builds, as regular boards won't depend on this code.

This is not a particularly strong opinion though, if this is an easier path forward to get this merged, I'll move it to `extra/src/tutorials` and update the text.

boards/nordic/nrf52840dk_demo/src/main.rs
```diff
@@ -0,0 +1,995 @@
...
+        spi_controller,
+        hmac,
+        keyboard_hid_driver,
+        app_flash,
+	oracle: oracle,
+        scheduler,
+        systick: cortexm4::systick::SysTick::new_with_calibration(64000000),
+    };
+
+    // let _ = platform.pconsole.start();
```
bradjc, line 950:
```suggestion
    let _ = platform.pconsole.start_hibernated();
```

bradjc (approved):

PR #3577, opened 2023-07-28
[RFC] Yield-WaitFor syscall

ppannuto:
### Pull Request Overview

Following the discussion at TockWorld6, this describes the proposed Yield-WaitFor and provides a (**untested**) rough implementation of how the kernel could easily implement it.

For ease of viewing, this draft PR edits TRD 104 directly so it can be seen as a diff. A final PR would follow the proper, full TRD process.

The primary motivation to move this functionality from a userspace `yield_for` into a specialized system call is to simplify correctness for userspace applications. Userspace upcall handlers do not have to worry about reentrancy if the kernel guarantees that exactly one and only one specific one of userspace's choosing will be called. It becomes an opt-in synchronous API for userspace without reducing the fundamental asynchronous design of Tock.

### Testing Strategy

Compiling (and no more!).

### TODO or Help Wanted

This is currently designed and architected as a minimal-impact change. In particular, if you want the actual status or return value from the upcall, you still need to have supplied a callback function. If (e.g., often in the case of prints) you don't care about the return success/failure, then you can just leave the default Null Upcall in place and this will do what you want.

The implementation is _not_ intended as final. Rather, it's just trying to demonstrate how this can be a very lightweight change in the kernel (indeed, one should not write careful code while also trying to participate in a meeting 😅 ).

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

---

[Rendered](https://github.com/tock/tock/blob/yeild-for/doc/reference/trd104-syscalls.md#413-yield-waitfor-nocallback)


bradjc:
I think we want something like this. This yield-for makes mixing async and sync code in userspace easier. Currently, it is fairly easy to write _all_ async or _all_ sync code in userspace, but when mixing them async callbacks can lead to unexpected call chains when using `yield_for()` in userspace. With this yield for syscall, sync code knows no other callbacks will happen.

The decision to make is whether to go with this lightweight version, or something more integrated (or something else).

Instead (or in addition to), we could add yield-for-blocking, which would remove the upcall entirely, and instead pass the upcall arguments to userspace via the return arguments of yield-for-blocking.

**Advantages of yield-for-blocking:**

1. Userspace sync code is easy to write:

    ```c
    allow();
    command();
    ret = yield_for_blocking(UPCALL_ID, &arg1, &arg2, &arg3);
    unallow();
    ```

    instead of:

    ```c
    void callback(e, arg1, arg2, arg3) {
        arg1=arg1;
        arg2=arg2;
        arg3=arg3;
    }

    allow();
    subscribe(callback);
    command();
    yield_for(UPCALLID);
    unallow();
    ```

2. Eliminates the context switches for subscribe and the upcall.

**Disadvantages:**

1. Semantic confusion: `schedule_upcall()` in a capsule may not, actually, schedule an upcall any more. Instead the "upcall" happens as the return to a yield.
2. If userspace _does_ call subscribe, but then only uses yield-for-blocking, that upcall would never get called, which is odd. But with a mix of yield-for-blocking and yield, the upcall would get called sometimes. This inconsistency complicates the interface.

Having articulated this, I see the appeal of yield-for. I think Pat pointed out that the complexity in userspace code is fairly easy to hide in a library layer, so that driver code would look like the yield-for-blocking case. Perhaps the prudent move is to address the complexity of writing sync and async code in userspace with yield-for, and not try to introduce blocking semantics before we know the aysnc version is insufficient.

ppannuto:
A quick note, `UPCALL_ID` is made up of `driver_number` and `subscribe_number` in practice, so it has to be two arguments. The upcall signature is currently three integer arguments and a user data pointer; while the blocking call can eliminate the last argument, as the pointer can just be local to the userland callsite, one of the arguments would have to overwrite one passed in. That makes this pseudocode more awkward:

> ```rust
> allow();
> command();
> // Can't do this:   ret = yield_for_blocking(UPCALL_ID, &arg1, &arg2, &arg3);
> int subscribe_num_and_arg1 = SUBSCRIBE_NUMBER;
> ret = yield_for_blocking(DRIVER_NUMBER, &subscribe_num_and_arg1, &arg2, &arg3);
> unallow();
> ```

There are probably other games that could be played to ease this, but it's all more complex than the lightweight change I think we should start with as-proposed here.

bradjc:
> ```rust
> allow();
> command();
> // Can't do this:   ret = yield_for_blocking(UPCALL_ID, &arg1, &arg2, &arg3);
> int subscribe_num_and_arg1 = SUBSCRIBE_NUMBER;
> ret = yield_for_blocking(DRIVER_NUMBER, &subscribe_num_and_arg1, &arg2, &arg3);
> unallow();
> ```

But the arg1 arg2 arg3 are going back up (kernel to userspace) and driver num and subscribe num are going down (userspace to kernel).

ppannuto:
Oh, you were thinking some more library magic here? I guess that works so long as the lower layer does the remapping and the pointer writing. I don't think there is any way to pass all the pointers for arg1-arg3 in the syscall and have the kernel do the write, but that's not necessary.... I think the below would work?

```c
// libtock.c
int yield_for_blocking(struct upcall_id, &arg1, &arg2, &arg3) {
    register uint32_t r0 __asm__ ("r0") = YIELD_FOR_BLOCKING_ID;
    register uint32_t r1 __asm__ ("r1") = upcall_id.drv_num;
    register uint32_t r2 __asm__ ("r2") = upcall_id.sub_num;
    register int retval __asm__ ("r0");
    register int rv1 __asm__ ("r1");
    register int rv2 __asm__ ("r2");
    register int rv3 __asm__ ("r3");
    __asm__ volatile (
    "svc 0"
    : "=r" (retval), "=r" (rv1), "=r" (rv2), "=r" (rv3)
    : "r" (r0), "r" (r1), "r" (r2)
    : "memory");

    if (retval & YIELD_FOR_ACTUALLY_YIELDED_FLAG_MASK) {
        *arg1 = rv1;
        *arg2 = rv2;
        *arg3 = rv3;
    }
    return retval;
}
```

brghena:
The C wrapper _could_ just take five arguments though instead of a struct:

```
int yield_for_blocking(driver_num, subscribe_num, &arg1, &arg2, &arg3) {
```

ppannuto:
True, though, since we're size-optimizing, if that function didn't get inlined everywhere, adding the fifth argument would require spilling to the stack [assuming `stuct upcall_id` were passed by ref, unlike the implicit copy I wrote above..] — for something low-level and apparently on the hot path like this, I'd push to limit to four args generally.

hudson-ayers:
> True, though, since we're size-optimizing, if that function didn't get inlined everywhere, adding the fifth argument would require spilling to the stack [assuming `stuct upcall_id` were passed by ref, unlike the implicit copy I wrote above..] — for something low-level and apparently on the hot path like this, I'd push to limit to four args generally.

This function seems like something that should get inlined everywhere, provided that it is being built with reasonable optimization settings, which should be the case for any project particularly concerned about size. So I would advocate for the five argument function definition, though I don't feel super strongly about it. If we used your approach, wouldn't we need to pass a pointer to upcall_id in order to not spill to the stack?

That said, I think that the disadvantages Brad described are valid:

> Disadvantages:
>
> Semantic confusion: schedule_upcall() in a capsule may not, actually, schedule an upcall any more. Instead the "upcall" happens as the return to a yield.
> If userspace does call subscribe, but then only uses yield-for-blocking, that upcall would never get called, which is odd. But with a mix of yield-for-blocking and yield, the upcall would get called sometimes. This inconsistency complicates the interface.

These are both disadvantages that do not exist with `blocking_command()` as Alyssa described it at TockWorld.

`yield_for()`, in cases where the return value matters, does not remove any system calls compared to the current status quo,  or the need for a callback function. It just allows userspace to ensure that only one callback will ever be called. Note that libtock-rs today only even allows one outstanding callback, so this seems to be of very limited usefulness in libtock-rs.

`yield_for_blocking()`, as Brad described, removes the context switch for subscribe (and unsubscribe in libtock-rs), as well as the need to implement a callback. But it comes with the two disadvantages that Brad mentioned, and still requires one more context switch than `blocking_command()`. The potential for mixing of functions that call `yield_for_blocking()` and `subscribe()` takes away a lot of the purported advantage here of "making it easier to simplify correctness for userspace apps", since there would be a new class of bug if apps mix async and sync driver methods that use the same callbacks.

ppannuto:
> wouldn't we need to pass a pointer to upcall_id in order to not spill to the stack?

Yes, that's what I meant by "[assuming `struct upcall_id` were passed by ref, unlike the implicit copy I wrote above..]"

I'm generally wary about relying on compiler optimizations for performance correctness, but I think we can shelve this side-hypothetical for now.

---

> Note that libtock-rs today only even allows one outstanding callback

Why is that limitation in place? That's a pretty crippling limitation for a lot of applications — can't have something like a periodic timer and a event-based sensor subscription live at the same time? That's not really an acceptable or realistic limitation for the long term.

If the issue is concurrent execution concerns of callbacks, `yield_for` as-proposed can be used to ensure that only one callback stack/chain is active at any given time.

---

I'm not opposed to something of the spirit of `yield_for_blocking` in addition to `yield_for`, but I think that is a more complex and nuanced interface for the reasons we've articulated, and that the simpler `yield_for` is useful on its own merits.

ppannuto:
One thing I had been ruminating on was whether there would be value in something closer to a `select/poll` style interface, or in this context a `yeild_for_any_of([array of upcall ids])` (with return value of index of which array entry fired).

In practice, I think the ergonomics around first `allow`ing the array and then sending the next syscall with the pointer is awkward, especially for the likely common-case of just wanting to yield on one specific upcall. I'd be inclined to steer clear of this more complex interface until it's proven necessary.

bradjc:
> > Note that libtock-rs today only even allows one outstanding callback
> 
> Why is that limitation in place? That's a pretty crippling limitation for a lot of applications — can't have something like a periodic timer and a event-based sensor subscription live at the same time? That's not really an acceptable or realistic limitation for the long term.

I'm not sure it is? In my little dabble yesterday it seems like subscribe in libtock-rs is essentially the `with` operator in python: the subscribe is removed when the context goes away, but you nest "with"s. Maybe that doesn't actually work in practice, but that's the impression the libtock-rs API gives.



> But it comes with the two disadvantages that Brad mentioned, and still requires one more context switch than `blocking_command()`.

It's really hard to discuss blocking command without something like this PR. What exactly _is_ blocking command? Critically, how does a capsule give the blocking command its return data? Also, can someone write its `#### Disadvantages` section? I kinda made it easy by writing my own for the idea I was supporting.

It _seems_ like the major disadvantage is going to be: when writing a capsule driver, after my interrupt fires, do I call `grant.sched_upcall()` or `kernel.command_return()`? And when using a capsule, can I call subscribe or do I have to call blocking command?

------

Imagine we implement `yield-for-blocking` and it turns out that we like it, and who knows maybe we rename `grant.sched_upcall()` to `grant.deliver()` or something and we end up with a nice way to merge an async kernel with a convenient sequential userspace.

It _seems_ like it might not be too big of a step to implement command-yield-for-blocking, which is a command immediately followed by a yield-for-blocking in a single system call (how exactly to do that tbd).

I say this because that to me seems more plausible than re-writing all of our capsules to have both an async and a sync version (if in fact that is what is required for blocking_command).

jrvanwhy:
> > Note that libtock-rs today only even allows one outstanding callback
> 
> Why is that limitation in place? That's a pretty crippling limitation for a lot of applications — can't have something like a periodic timer and a event-based sensor subscription live at the same time? That's not really an acceptable or realistic limitation for the long term.

`libtock-rs` allows multiple outstanding callbacks (e.g. you can wait for multiple sensors concurrently, with a timeout alarm), but it does not allow for independent tasks. That is, you cannot have a callback running in the background that blinks an LED continuously while your code does something completely unrelated to that.

There is more context on the decision to not support independent tasks at https://github.com/tock/libtock-rs/issues/334. The short version is that supporting that requires making a lot of important design decisions that I did not have the data to make at the time.

I also want to re-evaluate whether futures are acceptable in `libtock-rs`. Over the past few years, I've received a lot of suggestions on how to do more lightweight `Future`-based APIs than I achieved when I benchmarked the code size impacts of `Future`. I haven't had the time to investigate those suggestions, but I want to do so, and if any of them pan out then that would we wonderful. Using futures would allow us to support independent tasks without having to make those painful design tradeoffs.

ppannuto:
That state of libtock-rs is much more sane.

w.r.t. to an eventual more async world, there's some more complex stuff in libtock-c apps, but really nothing that would demand more capability than what is sounds like libtock-rs already supports. I do think one of the good outcomes from TockWorld6 is a push for the core team to get more parity in libtock-c / libtock-rs examples, so if I'm wrong, I suspect we'll see soon :).

doc/reference/trd104-syscalls.md
```diff
@@ -293,8 +293,8 @@ result with an error of `NOSUPPORT`.
 The Yield system call class is how a userspace process handles
 upcalls, relinquishes the processor to other processes, or waits for
 one of its long-running calls to complete.  The Yield system call
-class implements the only blocking system call in Tock that returns,
-`yield-wait`.
+class implements the only blocking system calls in Tock that return,
```
alevy, line 296:
```suggestion
class implements the only blocking system calls in Tock that return:
```

doc/reference/trd104-syscalls.md
```diff
@@ -303,61 +303,92 @@ call to invoke. The kernel invokes upcalls only in response to Yield
...
+
+### 4.1.2 Yield-Wait
+
+Yield number 1, Yield-Wait, blocks until an upcall executes. It is
+commonly used to provide a blocking I/O interface to userspace or to
+indicate to the kernel that the process has no work to do so the system
+can be put into a sleep state. A userspace library starts a long-running
+operation that has an upcall, then calls Yield-Wait to wait for an
+upcall.  When the Yield-Wait returns, the process checks if the resuming
+upcall was the one it was expecting, and if not calls Yield-Wait again.
```
alevy, line 370: This whole description misleading describes the primary use of `Yield-Wait` as simulating blocking calls. This is, in fact, a secondary use of `Yield-Wait` in the scenario that `Yield-WaitFor` exists.

Instead, it is the _only_ way to respond to asynchronous upcalls from the kernel. It is equivalent to:

```
while !yield-no-wait() { }
```
except that it doesn't busy loop.

For example, the _main_ use of `Yield-Wait` is in the main loop of an asynchronous program (as in `libtock-c`'s `_start` procedure)

doc/reference/trd104-syscalls.md
```diff
@@ -303,61 +303,92 @@ call to invoke. The kernel invokes upcalls only in response to Yield
...
+
+`yield-param-1`, `yield-param-2`, and `yield-param-3` are unused and reserved.
+
+
+### 4.1.3 Yield-WaitFor
+
+The third call, Yield-WaitFor, blocks until a specific upcall executes.
+In contrast to the Yield-Wait call, only the specified upcall is
+eligible to execute. Any other upcall that arrives will be queued.
```
alevy, line 379: This could be explained a bit better, but I think this is saying the right thing: that no upcalls except for the one specified in `Yield-WaitFor` are executed, and this syscall doesn't return until the specified upcall has been queued/executed (subject to the note below for what it means for an upcall to be queued/executed).

doc/reference/trd104-syscalls.md
```diff
@@ -303,61 +303,92 @@ call to invoke. The kernel invokes upcalls only in response to Yield
...
+
+`yield-param-1` is the Driver number and `yield-param-2` is the
+Subscribe number to wait for.
+
+`yield-param-3` is unused and reserved.
+
+_Note:_ In the case that the specified upcall is set to the Null Upcall,
+no upcall will execute, but Yield-WaitFor will still return. In this
+case, the contents of r0-r3 upon return are UNDEFINED and should not be
+relied on in any way.
```
alevy, line 389: OK, this is great except why not have `r0`-`r3` be the values that would have been passed a non-null upcall? This would allow many cases to avoid callbacks entirely and just process the results directly, and doesn't cost anything (I don't think).

bradjc: That is `yield-for-blocking`.



>  Disadvantages:
> 
>  Semantic confusion: schedule_upcall() in a capsule may not, actually, schedule an upcall any more. Instead the "upcall" happens as the return to a yield.
>  If userspace does call subscribe, but then only uses yield-for-blocking, that upcall would never get called, which is odd. But with a mix of yield-for-blocking and yield, the upcall would get called sometimes. This inconsistency complicates the interface.
>

alevy:
I'm strongly supportive of something very close to this.

# High bit

`Yield-WaitFor` solves several related problems concisely and elegantly:

1. It allows for true blocking semantics, most notably for debug statements in callbacks, where "simulated" blocking using the user-space `yield_for` relies heavily on user-written callbacks not, themselves calling `yield_for` recursively, e.g.. This is a big semantic improvement to userspace.

2. It can dramatically reduce the code size required of user-space applications that using a synchronous I/O API---they could avoid callbacks all-together, avoid corresponding `subscribe` calls, and reduces the number of system calls invoked per operation for some common operations. This seems _particularly_ useful in Rust applications where having callbacks always subscribed in the background (without unsubscribing them) may be unsound.

3. It doesn't require any additional functionality in capsules to support this. Everything is entirely differentiated in the process scheduler. This is important because it means that any code size implications in the kernel are fixed, and it means that userspace doesn't need to wait for a particular capsule to adapt to the new API in order to use this new yield variant.

# Some more details separated so they can be quoted separately

More on 1: it's worth looking through our various libtock-c examples and drivers and seeing where we would actually replace calls to the current library `yield_for` with a new `Yield-WaitFor`. In practice, these have slightly different semantics, so it's not obvious it would be everywhere, but it might be everywhere or almost everywhere.

More on 2: with some specialization in userspace as well as some sort of combined system-call system call ("please do a `command` followed by a `yield-waitfor`") I think we can get nearly optimal code size in applications (compared to a functionality-specific specialized blocking call). So we should view this as a step towards even more code size optimization.

# (There are no!) Disadvantages

@bradjc points out some disadvantages. These are very well worth considering. I think an evolution of this proposal avoids these pitfalls.

> 1. Semantic confusion: schedule_upcall() in a capsule may not, actually, schedule an upcall any more. Instead the "upcall" happens as the return to a yield.

Yes. As suggested later in the thread, either renaming or aliasing `schedule_upcall` to `deliver` may help avoid this confusion. I think this is _just_ a naming issue, which we should resolve, but doesn't break this proposal.

> 2. If userspace does call subscribe, but then only uses yield-for-blocking, that upcall would never get called, which is odd. But with a mix of yield-for-blocking and yield, the upcall would get called sometimes. This inconsistency complicates the interface.

Agree completely. `Yield-WaitFor` should use the existence of a subscribed callback to determine whether to execute a callback or return directly to the system call sight with the values passed to `schedule_upcall`. If a process has a callback subscribed, that callback should always execute. If it doesn't, `Yield-WaitFor` should always return the values that would have been passed to a callback.

There is a remaining question of what to do with `r0-r3` if a callback _was_ executed. I think, in this case, it might be reasonable to allow the callback (currently a `void`) to determine what those values should be.

alevy:
# Naming versions of the Yield variant

We've now described several variations of the proposed yield variant and I believe I've added to the confusion by conflating names for them above. Let's refer to them as follows:

* `Yield-WaitFor` is the original proposal as described in the [current revision of the TRD text](https://github.com/tock/tock/blob/60dd230642cce90ea34090b6432e6af948bbb25d/doc/reference/trd104-syscalls.md)
* `Yield-WaitFor-NoCallback` (proposed by @bradjc) never executes a callback, regardless of whether one is registered, and instead returns the values that would be passed to the callback as return values to the system call site.
* `Yield-WaitFor-CallbackIfPresent` (proposed implicitly by me) always executes a callback if one is present, otherwise it returns directly to the callsight returning the values that would have been passed to the callback in registers `r0-r3`.

bradjc:
What is the usecase for calling subscribe and using `Yield-WaitFor-CallbackIfPresent`? Maybe writing async code with ordering semantics where the code can know which order the callbacks will happen?

alevy:
> What is the usecase for calling subscribe and using Yield-WaitFor-CallbackIfPresent? Maybe writing async code with ordering semantics where the code can know which order the callbacks will happen?

One simple example may be logging statistics on how often something happens.

If the callback is able to return different values to the syscall site in r0-r3, it might be useful for filtering or customizing the result...

Just general modularity.

Indeed, I suspect this will rarely be used.

doc/reference/trd104-syscalls.md
ppannuto: I don't disagree—but that is also the original text from the TRD for the existing Yield-Wait call (https://github.com/tock/tock/pull/3577/files#diff-b402e09d7e051fa824752128efdf9f032782a83987280bb3905e05f2edbb059bL311-L316). For first pass, I tried to minimize changes to the existing TRD, but I'm happy to rephrase this to try to be clearer.

doc/reference/trd104-syscalls.md
ppannuto: Okay, so, the reason I didn't do this is because I couldn't decide the right behavior for `r3`.

TRD104 starts by describing upcalls abstractly as something that have (up to) 4 return arguments passed in r0-r3. We have one type of upcall the kernel actually emits right now, namely the upcall produced in response to subscribe [n.b., nothing in this TRD forbids other upcall signatures in the future*]. For clarity, I will use `SubscribableUpcall` to refer to an upcall invoked on a pointer passed via Subscribe.

Much to my surprise, as far as I can tell, the signature of `SubscribableUpcall` is not expressly defined anywhere except in [§5.2 of this TRD, where the C prototype for a `subscribe_upcall` is given as an example](https://github.com/tock/tock/blob/9d63213714f41ba1a8c57e6981688c867c08922f/doc/reference/trd104-syscalls.md#52-subscribe).

> Indeed, [we define the `FunctionCall` object passed to UKB](https://docs.tockos.org/kernel/process/struct.functioncall) with this note:
> 
> Struct that defines a upcall that can be passed to a process. The upcall takes four arguments that are Driver and upcall specific, so they are represented generically here.

In practice, the type signature for an upcall is established by the kernel in [kernel::upcall::schedule()](https://github.com/tock/tock/blob/9d63213714f41ba1a8c57e6981688c867c08922f/kernel/src/upcall.rs#L133-L139) and an implicit assumption in the [definition/creation of an `Upcall` object](https://github.com/tock/tock/blob/9d63213714f41ba1a8c57e6981688c867c08922f/kernel/src/upcall.rs#L78-L79) that upcalls come only from subscribes. But that's just an implementation artifact at the moment, not part of our ABI*.

Now, if `YieldFor` were to say something like 'sets r0-r3 to the Upcall Arguments if the current Upcall is the NullUpcall', then we would expect all Upcall Arguments to be set, right? Now, for the case of a `SubscribableUpcall` r0-r2 are clear, but what do we do with r3, which would be the `userdata` pointer? Does `YieldFor` need to do different things based on which type of Upcall it's passing?

We could further amend TRD104 to formally specify the function signature for _all_ upcalls, and then define `YieldFor` to skip r3. Or define `SubscribableUpcall` and specify `YieldFor` behavior only for that upcall type.

I was shooting for minimum delta with first RFC, and I don't think we can do this "skip the callback" without also making more changes around Upcall definition. I lean to formalizing `SubscribableUpcall` as a no-overhead option for today (since there is only one Upcall type still) while leaving the door open to other Upcall signatures in the future.

I do think the "skip the callback" would be nice. I can update this RFC to capture a version of that.

---

*Subscribe does start like this:

> The Subscribe system call class is how a userspace process registers upcalls with the kernel. 

Which if we really read into the "a" there maybe suggests the 1:1 mapping, but I wouldn't blink an eye at a future section that said something like "The `GetASignal` system call is another way userspace processes register upcalls with the kernel". If there really is _OneTrueUpcallSignature_, this TRD needs to say that much more directly.

ppannuto:
I've updated this to match the `Yield-WaitFor-CallbackIfPresent` proposed by @alevy.

There are some design choices not fully elucidated in the RFC yet, as I'd like us to decide what we want to do about https://github.com/tock/tock/pull/3577#discussion_r1284942276 before fleshing that text out completely.

PoC code is updated (again, compile-tested only) to realize the proposed interface.

doc/reference/trd104-syscalls.md
```diff
@@ -794,7 +814,52 @@ If an exit syscall is successful, it does not return. Therefore, the return
...
+linker places an upcall function, the first instruction of the function
+should be a no-op and the address of the second instruction passed
+instead.
+
+If a userspace process invokes subscribe on a driver ID that is not
+installed in the kernel, the kernel MUST return a failure with an
+error code of `NODEVICE` and an upcall of the Null Upcall.
+
+
+5.2 SubscribableUpcall
```
bradjc, line 856: Yes I think we need this. I was looking for this and I'm not sure we define anywhere in this document how an upcall is encoded.

doc/reference/trd104-syscalls.md
```diff
@@ -303,61 +303,108 @@ call to invoke. The kernel invokes upcalls only in response to Yield
...
+### 4.1.1 Yield-NoWait
+
+Yield number 0, Yield-NoWait, executes a single upcall if any is
+pending.  If no upcalls are pending it returns immediately.
+
+`yield-param-1` contains the memory address of an 8-bit byte that
+Yield-NoWait writes to indicate whether an upcall was invoked. If
+invoking Yield-NoWait resulted in an upcall executing, Yield-NoWait
+writes 1 to the field address. If invoking Yield-NoWait resulted in no
+upcall executing, Yield-NoWait writes 0 to the field address. This field
```
bradjc, line 355: I don't understand why we use pointers for yield-nowait and return values for yield-waitfor. Is there an optimization for the yield-nowait case that made the pointer the correct choice?

doc/reference/trd104-syscalls.md
```diff
@@ -303,61 +303,108 @@ call to invoke. The kernel invokes upcalls only in response to Yield
...
 |------------------------|----------|
 | Yield number           | r0       |
-| No wait field          | r1       |
-| unused                 | r2       |
-| unused                 | r3       |
+| yield-param-1          | r1       |
+| yield-param-2          | r2       |
+| yield-param-3          | r3       |
 
+The Yield system call class has no return value. This is because
```
bradjc, line 321: This would no longer be true.

bradjc:
With `Yield-WaitFor-CallbackIfPresent`, is it true that userspace can't determine at runtime if r0-r3 are valid, or if an upcall executed?

So libtock tries to write:

```c
int my_command_sync() {
  driver_command();
  asm("yield wait-for-callbackifpresent");
  // did an upcall execute?
  // or is r0-r3 valid?
  return ??
}
```

But the then the user at some point called `subscribe(my_driver)` and then forgot about it. So when the user calls `my_command_sync()` it seems to work ok, but the values are wrong.

Is this possible or am I missing something?

doc/reference/trd104-syscalls.md
bradjc: > *Subscribe does start like this:
> 
> > The Subscribe system call class is how a userspace process registers upcalls with the kernel.
> 
> Which if we really read into the "a" there maybe suggests the 1:1 mapping, but I wouldn't blink an eye at a future section that said something like "The `GetASignal` system call is another way userspace processes register upcalls with the kernel". If there really is _OneTrueUpcallSignature_, this TRD needs to say that much more directly.

But wouldn't we need to add that everywhere? A command syscall is THE ONLY way to issue a command. Yield is THE ONLY way to yield, etc. It seems clear to me that subscribe is the only way to subscribe, since that is all it does.

But yes we need to document the upcall signature.

ppannuto: I agree it's clear that "subscribe is the only way to subscribe".

What I do not think is clear is that "subscribe is the only way to get upcalls". Upcalls are defined completely independently of subscribe in the previous section. As written, subscribe is just one thing (and currently happens to be the only thing) than can generate an upcall.

bradjc: I don't think we need to say anything about subscribe being the only way to set upcalls, any more than we need to say command is the only way to instruct the kernel to start an operation. Commands are documented elsewhere. But, we do need to define what an upcall is, in my opinion.

doc/reference/trd104-syscalls.md
ppannuto: When the kernel returns to userspace, it only gets to set registers r0-r3 for one stack frame. In practice, we have two cases:
 - "DirectResume" {command, subscribe, allow, memop; yield-no-wait_DidntWaitCase}: We are resuming execution directly after the `svc` invocation, so the assembly that follows the `svc` command can use the values in r0-r3 as-set by the kernel
 - "PushedCallback" {yield-wait, yield-no-wait_DidWaitCase}: We resume execution at the start of the callback function. The values in r0-r3 are _consumed_ by the callback. When the callback finishes, it will `pop {lr}`, which has been set by the kernel to the instruction after the `svc`, and the assembly that invoked the syscall now gets to run. At this point `r0-r3` are _unknown_ as those are caller-save registers, so the callback can clobber them freely. The assembly that invoked the `svc` cannot make any assumptions about the values in `r0-r3`, nor can the kernel use them to pass things "to" the calling assembly. Thus, the "PushedCallback" case has to use the pointer approach for the kernel to communicate with the assembly that invokes the `svc`.

For Yield-WaitFor as-written, we again have two cases.
 - There _was not_ a subscribed callback, so we "DirectResume" and the kernel 'passes arguments to the calling assembly'
 - There _was_ a subscribed callback, so we are in the "PushedCallback" case, and the assembly that invoked the YieldFor can't get any information from the kernel.

bradjc: Thanks that is very clear and it would be great to capture it somewhere.

ppannuto:
> With Yield-WaitFor-CallbackIfPresent, is it true that userspace can't determine at runtime if r0-r3 are valid, or if an upcall executed?

As currently written, this is correct, yes. It assumes that userspace tracks whether it has subscribed a callback or not correctly, and if userspace cares that a callback ran, the callback can set a flag.

We could use `yield-param-3` to do the same pointer-indicator thing as Yield-NoWait if desired.

bradjc:
> > With Yield-WaitFor-CallbackIfPresent, is it true that userspace can't determine at runtime if r0-r3 are valid, or if an upcall executed?
> 
> As currently written, this is correct, yes. It assumes that userspace tracks whether it has subscribed a callback or not correctly, and if userspace cares that a callback ran, the callback can set a flag.
> 
> We could use `yield-param-3` to do the same pointer-indicator thing as Yield-NoWait if desired.

I propose we pass all yield return arguments via a pointer specified by `yield-param-3` unconditionally.

Otherwise, I don't know how we would use `Yield-WaitFor-CallbackIfPresent` in userspace libraries other than calling `subscribe(driver, upcall, 0)` before `Yield-WaitFor-CallbackIfPresent` every time.

bradjc (changes requested):
It would be great if I was missing something, but it seems with `Yield-WaitFor-CallbackIfPresent` a userspace implementation of yield_for is unable to tell if the registers after the yield_for syscall returns are valid or not, without some other tracking state or program knowledge. Marking as blocked to resolve this, because that seems like a problematic interface we do not want.

kupiakos:
OK! I'm finally able to get to this, my apologies on the delay.

## `blocking_command`

First off, a quick summary of how `blocking_command` works as currently implemented in Ti50:
- The syscall ABI is the same as command, other than its own syscall ID
- `blocking_command` is its own method in `SyscallDriver` that returns `Option<CommandReturn>`. `None` indicates there is further work to be done, and all upcalls on that app are queued.
- `ProcessStandard` tracks its own `wait_for_blocking_command: Cell<bool>`, which is exposed to capsules via `GrantKernelData::complete_blocking_command`, which calls `<ProcessStandard as Process>::set_blocking_command_return`.
- The `Process` trait also exposes `set_yielded_state_for_async_cmd`, called when blocking command returns `None`, which sets the process in a yielding state.

This simplified this:

```
allow
subscribe to an upcall that writes to an `Option<OutputOfUpcall>`
command
loop yield until that upcall has written to that var
unsubscribe
unallow
```

into:

```
allow
blocking command
unallow
```

## Thoughts on the current design

`yield-for` seems like a good stepping stone to simple sync code for those of us that want it. In particular, I like how it's capable of both acting as a normal-ish yield while being able to remove the upcall entirely (!) if you never subscribe.

I'm concerned that the validity of extracting r0-r3 from a yield-for is dependent on global knowledge that the current upcall is the null upcall. Is that OK? How would that work in libtock-rs soundly?

I have a possible design for resolving this:
- yield-for takes an upcall pointer in `r2` and application data in `r3`, and first performs a subscribe (possibly dropping  the previous upcall without returning).
- The proposal changes to instead depend on `r2` for determining whether to make an upcall or just write the args to r0-r3.
- This is similar to yield_blocking_for, but is capable of performing an upcall.
- Users that want to do an upcall with yield-for no longer need to subscribe, and users that don't can have confidence that they're providing the null upcall and can read from r0-r3.

bradjc:
I added

> I have a possible design for resolving this, mostly just throwing ideas out there:
> 
>     * yield-for takes an upcall pointer in `r2` and application data in `r3`, and first performs a subscribe (possibly dropping  the previous upcall without returning).
> 
>     * If `r2` is the null upcall, `r3` is ignored.
> 
>     * The proposal changes to instead depend on `r2` for determining whether to make an upcall or just write the args to r0-r3.
> 
>     * This is similar to yield_blocking_for, but is capable of performing an upcall.
> 
>     * Users that want to do an upcall with yield-for no longer need to subscribe if they didn't care about the previous value, and users that don't can have confidence that they're providing the null upcall and can read from r0-r3.

I added this to the summary comment: https://github.com/tock/tock/pull/3577#issuecomment-1666228991 as `Yield-WaitFor-OptionalSubscribe`.

ppannuto:
Okay, I've tried to update this to capture the current discussions here and to a current-best proposal.

Start with the summary of design discussion, here: https://github.com/tock/tock/blob/yeild-for/doc/reference/trd104-syscalls.md#design-rationale-and-alternatives
   - @kupiakos There are a few open questions with your proposal documented in this section — biggest problem is that upcall id takes two registers, so there aren't two registers available for subscribe pointer and app data

The biggest noteworthy changes:

 - I followed @bradjc's thinking and updated the whole `Yield Syscall Class` to always/explicitly use r1 as a "yield result" mechanism — this addresses the "did a callback run?" for all of the Yeild-Wait-* variations
 - I updated the main proposal to follow @alevy's implicit `CallbackIfPresent` proposal

I think this is getting somewhere pretty reasonable. Not written up in the RFC yet, because it occurred to me as I was writing this comment, but I think that @kupiakos's 'choose which callback when calling yield' can be implemented in userspace with the `CallbackIfPresent` as-written. The low-level libtock can take an optional callback argument which it can then call _assuming there is no callback subscribed that was already called_ (of course, the low-level impl could defensively subscribe the null callback before yield-for'ing, but that's not really the target design here..). As I write that sentence, then, I realize this is may be a case for the `NeverCallback`, as it lets the low-level userspace always make the callback decision based on what was passed to `libtock::yield-for`.

ppannuto:
`yield-result` is a pointer passed in r1 to the syscall, the kernel uses that pointer to write a byte in process memory. The "return value" or result from yield is that byte in memory. (This is what Yield-NoWait already does [and Yield-Wait by implementation bug currently]). The pointer is lost by the time userspace executes again.

bradjc:
Do we need a new error code?

```c
int my_sync_syscall(int* len) {
  uint8_t yield_result
  command()
  yield_for(upcall_id, &yield_result)
  if (yield_result == 0) {
    *len = r2
    return SUCCESS
  } else {
    return ???
  }
}
```

What happens if sync api is expecting yield to return r0-r3 but it doesn't?

ppannuto:
Oh shoot, I meant to call attention to this too. At the [end of §4.1.0](https://github.com/tock/tock/blob/yeild-for/doc/reference/trd104-syscalls.md#41-yield-class-id-0) I added another open question:

> **OPEN QUESTION:** ^^ This documents what we currently do, but it is the case right now that there is no way for userspace to identify an errorneous call to yield (e.g. invalid value in r0)---should instead yeild-result be written no matter what and an error case defined?

I went ahead with the first half of that (yield-result written no matter what), but the only two options currently are:
   * `*result == 0` — No upcall executed
   * `*result == 1` — An upcall executed

I _think_ a more general approach would have the kernel setting `*result == {§3.2-style Return Code}`, ~~but that would require two words at minimum to fully support just the `Success` and `Failure` cases. It's already a bit weird that a syscall is writing an non-Allowed byte; writing two words (or four words if we wanted to expose the whole Return Code space) gets a bit messier perhaps.~~ Which we could support by using `*result == -1` (or whatever) means that r0-r3 are set as-per §3.2-style Return Code.

doc/reference/trd104-syscalls.md
```diff
@@ -303,61 +336,226 @@ call to invoke. The kernel invokes upcalls only in response to Yield
...
-indicate whether an upcall was invoked. If invoking `yield-no-wait`
-resulted in an upcall executing, `yield-no-wait` writes 1 to the
-field address. If invoking `yield-no-wait` resulted in no upcall
-executing, `yield-no-wait` writes 0 to the field address. This field
-allows userspace loops that want to flush the upcall queue to
-execute `yield-no-wait` until the queue is empty.
+**OPEN QUESTION:** ^^ This documents what we currently do, but it is the
+case right now that there is no way for userspace to identify an
+errorneous call to yield (e.g. invalid value in r0)---should instead
+`yeild-result` be written no matter what and an error case defined?
```
bradjc, line 391: I'm not sure what this means. What is an erroneous call to yield? Also, `yield-result` is written no matter what, assuming it is a valid process pointer. What would the error case capture? When are you going to enable spell check in your text editor :)?

ppannuto: The only one I can think of is an illegal yield type, i.e. the invalid value in r0. FWIW, this is what the current TRD says:
image
which basically kind of just doesn't say anything about what else does or doesn't happen if you make an impossible syscall.

Frankly, I don't really think it matters too much what we do, as so much else has gone wrong likely if you're asking for an invalid Yield, but I do feel like we should define what does or doesn't happen—especially if we have a "r1 is the yield-result" paradigm.

---

> When are you going to enable spell check in your text editor :)?

😬 , never... I learned a long time ago that spellcheck is too distracting for my brain to write/synthesize anything non-trivial. I do try to make a point of spellchecking non-draft PRs before pushing though :)

bradjc: > Frankly, I don't really think it matters too much what we do,

I think we are already clear on what happens:

> If userspace tries to invoke a system call that the kernel does not support, the system call will return a Failure result with an error code of NODEVICE or NOSUPPORT (Section 4).

bradjc: I think that a system call with class number 0 and r0==<anything not defined> is an unknown syscall and not an unknown yield.

bradjc:
I remain in favor of `Yield-WaitFor-NoCallback`. Are there any use want-to-implement-today use cases that `Yield-WaitFor-NoCallback` doesn't support? I'm counting efficiency in terms of number of syscalls as something else, and not a use case.

Are there any use cases that _want_ to use yield_waitfor_X with upcalls? Amit has mentioned hypothetical ones, but are there any we would implement immediately if we could? Is anyone asking for this?

The discord of doing a _blocking_ syscall with an _asynchronous_ callback makes me adverse to any of the upcall yield-waitfor options. Trying to explain that to someone would be difficult.

I propose we implement the thing which 1. supports our use case and only use cases we know of and 2. is low impact to implement in the kernel, which is `Yield-WaitFor-NoCallback`. This provides userspace with a clearly desired API without adding features and complexity no one is asking for.

hudson-ayers:
@bradjc to clarify your most recent comment, are you now in favor of *just* `Yield-WaitFor-NoCallback`? Your previous comment indicated you wanted both it and `Yield-WaitFor`.

If the former, I think I agree. `Yield-WaitFor-NoCallback` seems clearly useful, efficient, and relatively easy to understand -- it is the most similar to `blocking_command()`. I could not immediately think of a use case where I would really need in-kernel `Yield-WaitFor` (over userspace `yield-for()`) and not prefer `Yield-WaitFor-NoCallback`. I understand the idea that this enables prioritizing some particular ordering of callback handling, but the same thing can be handled with `Yield-WaitFor-NoCallback` followed by a manual call to the relevant "callback" function. 

The main point of confusion with `Yield-WaitFor-NoCallback` is still the scenario in which a capsule calls `subscribe()` with an upcall and then calls `yield_for_blocking()` -- the upcall is defined, but never called, which seems confusing. And if another function is using that upcall to receive periodic events it is possible that its not totally clear that calling `Yield-WaitFor-NoCallback` would clear that upcall from future events. 
One way to prevent this confusion would be for `Yield-WaitFor-NoCallback` to return an error if a non-null callback is already set (perhaps we would add another sentinel value in the kernel to indicate a "fake upcall". This would mean adding another error code to userspace, but IMO would prevent some possibly confusing bugs.

bradjc:
> @bradjc to clarify your most recent comment, are you now in favor of _just_ `Yield-WaitFor-NoCallback`? Your previous comment indicated you wanted both it and `Yield-WaitFor`.

My current opinion is to add only Yield-WaitFor-NoCallback`.

bradjc:
> The main point of confusion with `Yield-WaitFor-NoCallback` is still the scenario in which a capsule calls `subscribe()` with an upcall and then calls `yield_for_blocking()` -- the upcall is defined, but never called, which seems confusing. And if another function is using that upcall to receive periodic events it is possible that its not totally clear that calling `Yield-WaitFor-NoCallback` would clear that upcall from future events.
> One way to prevent this confusion would be for `Yield-WaitFor-NoCallback` to return an error if a non-null callback is already set (perhaps we would add another sentinel value in the kernel to indicate a "fake upcall". This would mean adding another error code to userspace, but IMO would prevent some possibly confusing bugs.

I think it's fine if there is an upcall handler subscribed that doesn't get used. Having `Yield-WaitFor-NoCallback` be able to error if there is a subscription means every call now has to look like:

```c
r0, r1, r2, r3 = yield_wait_for_no_callback();
if (r0 == SUBSCRIPTION_ERROR) {
  subscribe(NULL);
  r0, r1, r2, r3 = yield_wait_for_no_callback();
}
```

And if you want to be even more correct:

```c
old_subscribe = NULL;
r0, r1, r2, r3 = yield_wait_for_no_callback();
if (r0 == SUBSCRIPTION_ERROR) {
  old_subscribe = subscribe(NULL);
  r0, r1, r2, r3 = yield_wait_for_no_callback();
  subscribe(old_subscribe);
}
```

Because having a library secretly unregister your subscription would also be a subtle bug.

brghena:
I also support `Yield-WaitFor-NoCallback` of these choices. From there, the library _could_ still make a callback with very few additional cycles. I also think that even if someone has subscribed previously, it's pretty clear that calling this version of yield won't lead to a callback.

I think it's worth starting an implementation for `Yield-WaitFor-NoCallback` and seeing what it feels like in terms of effort, code size, possible bugs, etc.

phil-levis:
`Yield-WaitFor-NoCallback` also makes sense to me. It captures an extremely common case of completion callbacks, which is just waiting on the stack-based boolean. With respect to @hudson-ayers concern about subcribed callbacks not being invoked, I think this is consistent with subscribe semantics. In particular, calling this variant of yield is equivalent to subscribe being called to install a new upcall, yielding for this upcall, then re-installing the old one before returning.

To follow up on the discussion in the call today, I'd suggesting looking at how 4 different system call drivers (userspace and kernel) would change if re-implemented with `Yield-WaitFor-NoCallback`: Console, Alarm, ADC, and SPI Controller. They cover a bunch of different cases. The initial stated reasoning for this system call was:

> The primary motivation to move this functionality from a userspace yield_for into a specialized system call is to simplify correctness for userspace applications. Userspace upcall handlers do not have to worry about reentrancy if the kernel guarantees that exactly one and only one specific one of userspace's choosing will be called. It becomes an opt-in synchronous API for userspace without reducing the fundamental asynchronous design of Tock.

Can we quantify "simplify" as "shorter code text", "fewer branches", or "smaller binary"? Are there other considerations we care about?

doc/reference/trd104-syscalls.md
ppannuto: > I think that a system call with class number 0 and r0== is an unknown syscall and not an unknown yield.

The problem here is "how does the kernel tell userspace it was an invalid syscall?"

I think we the challenge with the current text / design is that these two parts are at odds:

> **the system call will return** a Failure result with an error code...

> The Yield system call class has no return value.

How can a system call class that has no return value return a Failure result?

We're in a pretty deep never-should-happen corner case here, but what I'm struggling with is 'what is the right thing for userspace to do'? Once userspace invokes a syscall with `svc 0`, i.e. something in the Yield class, it _cannot_ first 'read the return value to see if the syscall failed' as the Yield class defines different return semantics than every other syscall.

ppannuto:
Main text is updated for Yield-WaitFor-NoCallback semantics. 

> Can we quantify "simplify" as...

The original / primary motivation discussed at TockWorld was not any quantitative metric. It was correctness-oriented—i.e., no reentrancy concerns for userspace developers. I think the current text captures that as-intended.

The subsequent paragraphs already discussed reduction in syscall overhead as a secondary motivation. I also added a paragraph on code size. I think those are all the driving motivations here.

---

Will work on the proposed impl's next...

ppannuto:
The userspace updates are actually pretty substantial if the intent is to switch the library fully to a synchronous interface, so I just did console to start: https://github.com/tock/libtock-c/pull/345

With the enormous caveat that all of these are currently **compile-tested only**, for the kernel:

size before:
```
   text	   data	    bss	    dec	    hex	filename
 110636	      0	  19928	 130564	  1fe04	/Volumes/code/helena-project/tock/target/thumbv7em-none-eabi/release/hail
```

size after:
```
   text	   data	    bss	    dec	    hex	filename
 112228	      0	  19928	 132156	  2043c	/Volumes/code/helena-project/tock/target/thumbv7em-none-eabi/release/hail
```

So for rough order-of-magnitude sense of impact, this adds ~1500 bytes to the kernel while removing ~500 bytes from the userspace console implementation — i.e., a code size improvement for systems >3 apps.

For simplest case (`getnstr`), results in >=1 syscall reduction (no `subscribe` == 1, >1 if other upcalls would have run).

For the primary goal of code complexity reduction, the `putnstr` implementation is qualitatively night and day different. To try to quantify, `console.c` went from 119 to 36 SLOC (source lines of code, from `cloc`).

phil-levis:
> The original / primary motivation discussed at TockWorld was not any quantitative metric. It was correctness-oriented—i.e., no reentrancy concerns for userspace developers. I think the current text captures that as-intended.

That makes sense, but how do you evaluate that? The flipside is that you re-order callbacks, which has its own issues. E.g., if a do a yield-wait-for on packet reception, I can delay a timer upcall indefinitely. This seems a little tricky -- how do you write a userspace library when any other library (particularly one written by a developer who doesn't want to handle re-entrancy) can arbitrarily delay an upcall? Part of the challenge is that this is about the *implementation* of a library, i.e., which yield call it makes.

This makes me skittish because it's an extremely large architectural change (i.e., how upcalls are handled) without a substantive evaluation or tradeoff analysis plan. I buy the arguments -- it *seems like* it could be a good idea and so is worth exploring. But you have to build it. Is there a way to evaluate the cost of reentrancy concerns?  I'm not arguing that this isn't the best solution, just that it's going to create *different* correctness problems. It seems very likely that those correctness problems are much less bad than the current ones. But this isn't a silver bullet. 

Perhaps this would call for labeling particularly library functions "non-reentrant", and if you call a non-reentrant function, you are non-reentrant? That way, userspace functions that need re-entrance can know not to call these functions. I.e., if you need to handle a timer, you don't call "nr_receive" or something like that.

It's important that the final TRD not include long discussions of design rationales; this isn't an academic paper. Following the command 0 example we could include a link?

kernel/src/process.rs
```diff
@@ -927,6 +943,10 @@ pub enum FaultAction {
 /// This is public for external implementations of `Process`.
 #[derive(Copy, Clone)]
 pub enum Task {
+    /// A task that should not actually be executed. This is used to resume a
+    /// suspended process without invoking any callbacks in userspace (e.g.,
+    /// when YieldFor resolves to a Null Upcall).
+    NullSubscribableUpcall(NullSubscribableUpcall),
```
bradjc, line 949: This addition seems not strictly necessary. If we make `fn_ptr` in `FunctionCall` optional, we can re-use that enum option.

Perhaps it could be `FunctionalCallOrReturn`?

alevy:
RE: @ppannuto's initial evaluation:

> So for rough order-of-magnitude sense of impact, this adds ~1500 bytes to the kernel while removing ~500 bytes from the userspace console implementation — i.e., a code size improvement for systems >3 apps.

and

> For the primary goal of code complexity reduction, the putnstr implementation is qualitatively night and day different. To try to quantify, console.c went from 119 to 36 SLOC (source lines of code, from cloc).

I buy this generally, but _note_ I don't think this is an apples to apples comparison for putnstr since the current main-brach implementation queues prints to avoid blocking while the modified version doesn't. Most of the complexity in the user-space console driver is from this queuing and can be removed. As an exercise, I did this a while ago and got substantially more concise code and substantial savings, without changing the kernel (I don't think I got 500 bytes, and it was also untested, so I'm not saying there are no savings here, just that this particular comparison probably shouldn't be taken completely at face value).

Also, I didn't even share my findings in a place I can now find them, so @ppannuto's exercise is about 1000x more valuable and useful than mine and my comment here should not be interpreted as a complaint.

alevy:
## Brief summary of our state

- It seems that at least the plurality are in favor of primarily opinion is to explore `Yield-WaitFor-NoCallback`.
- I seems to me that `Yield-WaitFor-OptionalSubscribe` is different in interface (and I am partial to the options it might provide, particular for Rust userland) but not so different in how it might affect most application correctness and code size etc. So I think it might be a good idea to go ahead exploring the plurality opinion here, and gather evidence one way or the other for whether/how to include subscribe options at the yield callsite.
- We don't have a concrete plan for how to "measure" whether this change will be an net benefit in terms of correctness. I think some of this comes down to articulating what choices belong in the userland library vs the app itself, but also an agreement on what _precisely_ we're looking for.
- We have a concrete idea of what we want from a performance perspective: smaller application code size with at limited code size increase in the kernel (such that a small number of savings in apps "wins"), and _secondarily_ performance gains. It would be best to have some specific threshold of apps complexity/size (i.e. some examples) we might care about, but we can probably defer on that.

I suggest we forge a plan for exploring this completely, which would include a prototype implementation in the kernel (which I believe we already have) and porting several drivers & apps in both C and Rust to use this new interface.

alevy:
The relevant text describing `Yield-WaitFor-NoCallback` (YWFNC) is this:

> | Argument               | Register |
> |------------------------|----------|
> | Yield number           | r0       |
> | yield-result           | r1       |
> | yield-param-A          | r2       |
> | yield-param-B          | r3       |

> Yield-WaitFor-NoCallback, suspends the process until one
> specific upcall---the `WaitedForUpcallId`---is ready to execute. If
> other events arrive that would invoke an upcall on this process, they
> are queued by the kernel, and will be delivered after the event for the
> `WaitedForUpcallId`.  Event order in this queue is maintained.
> 
> This process will resume execution when an event in the kernel generates
> an upcall that matches the `WaitedForUpcallId`. No userspace callback
> function will invoked by the kernel. Instead, the contents of r0-r3 will
> be set to Upcall Arguments as-feasible, specifically:
>  - A `SubscribeUpcall` will set r0-r2 and leave r3 untouched.
>
> `yield-param-A` is the Driver number and `yield-param-B` is the
> Subscribe number to wait for. Together, these make up the
> `WaitedForUpcallId`.

Important properties:

- Between YWFNC being called and returning, _no callbacks will run_.
- There may be a callback subscribed to that subscribe number, and it will be ignored
- Library/application may also call normal yield elsewhere, which would induce callbacks, but because YWFNC blocks the process completely, these are never in competition.

brghena:
From discussion: the arguments originally listed above are incorrect. There should only be three arguments.

bradjc:
This would also make it much easier to mix sync and async code in userspace, wouldn't it? I'm currently having my app crash from calling `printf` in a callback. I would like to see us do this just to simplify writing libtock-c apps.

doc/reference/trd104-syscalls.md
```diff
@@ -794,24 +971,93 @@ If an exit syscall is successful, it does not return. Therefore, the return
...
+application code and where the linker places an upcall function, the
+first instruction of the function should be a no-op and the address of
+the second instruction passed instead.
+
+If a userspace process invokes subscribe on a driver ID that is not
+installed in the kernel, the kernel MUST return a failure with an
+error code of `NODEVICE` and an upcall of the Null Upcall.
+
+
+5.2 SubscribeUpcall
```
bradjc, line 1013: This is confusing, because the NullUpcall section describes the arguments to `subscribe()` and this describes the arguments to `schedule()`. Do we need to add this section right now?

kernel/src/kernel.rs
```diff
@@ -752,41 +803,57 @@ impl Kernel {
                 }
                 process.set_syscall_return_value(rval);
             }
-            Syscall::Yield { which, address } => {
+            Syscall::Yield {
+                which,
+                param1,
+                param2,
```
ppannuto, line 809:
```suggestion
                paramA,
                paramB,
```

Not that it really matters, but might be a bit nicer for readability to match the names used in the TRD

kernel/src/process.rs
```diff
@@ -957,6 +977,23 @@ pub struct FunctionCall {
     pub pc: usize,
 }
 
+/// This is similar to `FunctionCall` but for the special case of the Null
+/// Upcall for a subscribe. Because there is no function pointer in a Null
+/// Upcall we can only return these values to userspace. This is used to pass
+/// around upcall parameters when there is no associated upcall to actually call
+/// or userdata for arg3.
```
ppannuto, line 984:
```suggestion
/// or userdata for argument2.
```
?, matching the actual struct below

kernel/src/syscall.rs
```diff
@@ -82,10 +96,12 @@ impl TryFrom<u8> for SyscallClass {
...
     /// Structure representing an invocation of the Yield system call class.
-    ///
-    /// - `which`: the Yield identifier value
-    /// - `address`: the no wait field
-    Yield { which: usize, address: *mut u8 },
+    /// `which` is the Yield identifier value and `address` is the no wait field.
+    Yield {
+        which: usize,
+        param1: usize,
+        param2: usize,
```
ppannuto, line 103:
```suggestion
        paramA: usize,
        paramB: usize,
```

kernel/src/syscall.rs
```diff
@@ -190,7 +206,8 @@ impl Syscall {
         match SyscallClass::try_from(syscall_number) {
             Ok(SyscallClass::Yield) => Some(Syscall::Yield {
                 which: r0,
-                address: r1 as *mut u8,
+                param1: r1,
+                param2: r2,
```
ppannuto, line 210:
```suggestion
                paramA: r1,
                paramB: r2,
```

ppannuto (commented):
This looks good to me!

bradjc:
I managed to test this with https://github.com/tock/libtock-c/pull/345 and verify that printf with yield-waitfor worked.

alevy (approved):

alevy:
@tock/core-wg This is definitely P-Significant, and I marked it so, so we need another approval. Having said that, I think this is pretty good to go.

kernel/src/kernel.rs
```diff
@@ -638,6 +644,46 @@ impl Kernel {
                         },
                     }
                 }
+                process::State::YieldedFor(upcall_id) => {
+                    // If this process is waiting for a specific upcall, see if
+                    // it is ready. If so, dequeue and execute it, otherwise move on.
```
alexandruradovici, line 649: I would use another word instead of *execute*, as this does not actually execute the callback function.

kernel/src/kernel.rs
```diff
@@ -638,6 +644,46 @@ impl Kernel {
...
+                    match process.remove_upcall(upcall_id) {
+                        None => break,
+                        Some(task) => {
+                            let (a0, a1, a2) = match task {
+                                Task::ReturnValue(rv) => {
```
alexandruradovici, line 654:
```suggestion
                                // There is not callback function registered, we just return the
                                // values provided by the driver
                                Task::ReturnValue(rv) => {
```

kernel/src/kernel.rs
```diff
@@ -638,6 +644,46 @@ impl Kernel {
...
+                                            "[{:?}] Yield-WaitFor: [NU] ({:#x}, {:#x}, {:#x})",
+                                            process.processid(),
+                                            rv.argument0,
+                                            rv.argument1,
+                                            rv.argument2,
+                                        );
+                                    }
+                                    (rv.argument0, rv.argument1, rv.argument2)
+                                }
+                                Task::FunctionCall(ccb) => {
```
alexandruradovici, line 666:
```suggestion
                                // There is an callback function registered, but as the application
                                // used `Yield-WaitFor`, we do not execute it, we just return
                                // its arguments values to the application
                                Task::FunctionCall(ccb) => {
```

alexandruradovici (changes requested):
I suggested adding a few comments, as it was not clear for me when reading the code that the registered callback will not be called. I had to read the TRD for this. I think this will make code understanding easier.

kernel/src/kernel.rs
alevy: "schedule?"

alexandruradovici: Not sure about this, we do not actually schedule the callback, just return its arguments. What about *return its arguments to the application*.

alevy, line 649:
```suggestion
                    // If this process is waiting for a specific
                    // upcall, see if it is ready. If so, dequeue it
                    // and return its values to the process without
                    // scheduling the callback.

```

alevy (approved):

alevy:
@alexandruradovici Comments adapted. I think all good.

In general, I don't _love_ the structure of the `Task` enum, which is now kind pulling double-duty, and thus requires a bit more documentation (as you've added) to make clear how it's actually used. Having said that, I think it's worth deferring refactoring the code until IPC is potentially reworked, which might unify `Task::IPC` and `Task::FunctionCall`, and overall result in cleaner code.

alexandruradovici (approved):

bradjc:
How did we merge a significant PR in 3 days with only 2 approvals?

bradjc:
Did anyone else test this? I ran exactly one test on one board with one process. And nothing on riscv.

bradjc:
> We need 2 approvals for a p-significants

Can we write this down somewhere? That is not my understanding nor what our documentation says

> **Significant pull requests** require review by the entire core team. Each
> core team member is expected to respond within one week.

alevy:
Ok, sorry.

My sense was that it had sat since your rebase and testing, I confirmed with similar testing, and there has been for a while either consensus or apathy about the design (I think I'm the most dissenting voice about the design specifically). And we'll do more testing for release and now that it's merged.


We can/should clarify that language. Maybe your interpretation is right and I was wrong, maybe my interpretation was right, but either way it's not clear

doc/reference/trd104-syscalls.md
lschuermann, line 1: We should also update the draft version of this TRD in a follow-up PR.

lschuermann: Created a PR: #4032

lschuermann:
> We can/should clarify that language. Maybe your interpretation is right and I was wrong, maybe my interpretation was right, but either way it's not clear

I don't know whether requiring approvals from all core team members is feasible in practice, but we should definitely allow for more time from the point where reviews from the WG are requested, to when someone is able to hit merge.

alistair23:
> How did we merge a significant PR in 3 days with only 2 approvals?

It's also still an `RFC`

PR #3580, opened 2023-07-31
boards: remove nrf52840dk demo

bradjc:
### Pull Request Overview

This board was added for the July 2023 tutorial which is complete.

The book has been updated to provide guides to setup the kernel to support the HOTP application: https://book.tockos.org/key-overview.html#stages

### Testing Strategy

I built this board myself during the tutorial, and the changes to the book are based on my changes to nordic/nrf52840dk/src/main.rs.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


lschuermann (approved):
Agreed that we shouldn't keep this board around, but worried that the setup documentation will go out of date. I guess we're going to have to think about this more for the book in general, as alluded to at TockWorld.

On the upside, this means that we don't have to update the board path in the encryption oracle capsule any more.

bradjc:
> but worried that the setup documentation will go out of date

I think the beauty of components is that I'm not too worried about this. It seems like PRs which modify every/multiple board are somewhat rare these days.

bradjc:
It may also be time to add "meta components", where a component encapsulates all of the setup for a tock use case. In this case, there would be the HOTP component which would create all of the HOTP stuff.

I would be supportive of creating the nrf52840dk-hotp board, which was just HOTP + console + alarm. While the book guides I think are helpful for learning tock and how to configure a kernel, they are not convenient for us when we just want to test that hotp still works.

brghena (approved):

alistair23 (approved):

ppannuto (approved):

PR #3585, opened 2023-08-01
Console: notify app if transmit fails

### LINKED ISSUE #3433: Console empty string bug
Printing to console a formatted output, with an empty string given as a parameter causes the app to yield infinitely.

I tried this code:

```
writeln!(
                Console::writer(),
                "Temperature: {}{}.{}*C\n",
                if temp_val > 0 { "" } else { "-" },
                i32::abs(temp_val) / 100,
                i32::abs(temp_val) % 100
            )
```

I expected to see this happen: a positive temperature value, with no explicit before the absolute value, printed to the console.

Instead, this happened: `Temperature: ` was printed to the console and nothing else happened.

It looks like the `WRITE command` in the [console capsule](https://github.com/tock/tock/blob/44f39d7c8cf5db0038606f08640bfae670127eef/capsules/src/console.rs) does not return an error in this case, but still no upcall is scheduled.

## Meta

Hardware platform: micro:bit v2

bradjc:
### Pull Request Overview

This pull request fixes #3433. It corrects the virtual uart to return an error on 0 length tx, and handles that error in console.

The tricky part is ensuring that the console continues after an error. There are two cases:

1. A command happened and console called `send()`. It is ok if this does not trigger a interrupt (buf_txd) because if the command was able to immediately send, then nothing was pending and there is no other work to do.
2. A buf_txd interrupt happened, found work to do, called send, which errored. This _does_ need to check if there was other work to do. The easy fix is to check the tx_in_progress flag rather than assuming send actually sent a buffer which will cause a callback.

### Testing Strategy

Running temperature and console libtock-rs apps.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/core/src/console.rs
+++ b/capsules/core/src/console.rs
@@ -148,7 +148,11 @@ impl<'a> Console<'a> {
     ) -> bool {
         if app.write_remaining > 0 {
             self.send(processid, app, kernel_data);
-            true
+
+            // The send may have errored, meaning nothing is being transmitted.
+            // In that case there is nothing pending and we return false. In the
+            // common case, this will return true.
+            self.tx_in_progress.is_some()
         } else {
             false
         }
@@ -193,7 +197,20 @@ impl<'a> Console<'a> {
                     })
                     .unwrap_or(0);
                 app.write_remaining -= transaction_len;
-                let _ = self.uart.transmit_buffer(buffer, transaction_len);
+                match self.uart.transmit_buffer(buffer, transaction_len) {
+                    Err((_e, tx_buffer)) => {
+                        // The UART didn't start, so we will not get a transmit
+                        // done callback. Need to signal the app now.
+                        self.tx_buffer.replace(tx_buffer);
+                        self.tx_in_progress.clear();
+
+                        // Go ahead and signal the application
+                        let written = app.write_len;
+                        app.write_len = 0;
+                        kernel_data.schedule_upcall(1, (written, 0, 0)).ok();
+                    }
+                    Ok(()) => {}
+                }
             });
         } else {
             app.pending_write = true;
--- a/capsules/core/src/virtualizers/virtual_uart.rs
+++ b/capsules/core/src/virtualizers/virtual_uart.rs
@@ -413,7 +413,9 @@ impl<'a> uart::Transmit<'a> for UartDevice<'a> {
         tx_data: &'static mut [u8],
         tx_len: usize,
     ) -> Result<(), (ErrorCode, &'static mut [u8])> {
-        if self.transmitting.get() {
+        if tx_len == 0 {
+            Err((ErrorCode::SIZE, tx_data))
+        } else if self.transmitting.get() {
             Err((ErrorCode::BUSY, tx_data))
         } else {
             self.tx_buffer.replace(tx_data);


ppannuto (commented):
Ugh... I don't like this :/, I feel like this _should_ be something more like

```diff
diff --git a/capsules/core/src/console.rs b/capsules/core/src/console.rs
index 4854c886e..cb0b5bce4 100644
--- a/capsules/core/src/console.rs
+++ b/capsules/core/src/console.rs
@@ -198,7 +198,7 @@ impl<'a> Console<'a> {
                     .unwrap_or(0);
                 app.write_remaining -= transaction_len;
                 match self.uart.transmit_buffer(buffer, transaction_len) {
-                    Err((_e, tx_buffer)) => {
+                    Err((err, tx_buffer)) => {
                         // The UART didn't start, so we will not get a transmit
                         // done callback. Need to signal the app now.
                         self.tx_buffer.replace(tx_buffer);
@@ -207,7 +207,13 @@ impl<'a> Console<'a> {
                         // Go ahead and signal the application
                         let written = app.write_len;
                         app.write_len = 0;
-                        kernel_data.schedule_upcall(1, (written, 0, 0)).ok();
+                        kernel_data.schedule_upcall(1,
+                            (
+                                kernel::errorcode::into_statuscode(Err(err)),
+                                written,
+                                0,
+                            )
+                        ).ok();
                     }
                     Ok(()) => {}
                 }
```

As-is, this drops errors silently :(. But, we're hamstrung by the interface of the stabilized console driver here, which didn't have any way to report asynchronous errors on the write path.

I feel like we should deprecate upcall 1 and add an upcall 3 or something.

This is simple and clean enough that this might be the right problem to explore how we want to manage sustaining deprecated-and-future APIs in-tree. We punted on that problem once already with the LEDs, but this is going to keep coming up, and we need to develop some principled approaches.

bradjc:
> As-is, this drops errors silently :(. But, we're hamstrung by the interface of the stabilized console driver here, which didn't have any way to report asynchronous errors on the write path.

I don't disagree, but the fact that nearly every app ever uses printf and it took this many years to have a "failure" case likely indicates that addressing this isn't too pressing.

alistair23 (approved):

bradjc:
I know this isn't as pressing because the previous PR happens to fix the issue in #3433, but I assume we still want this?

lschuermann (approved):

lschuermann:
Created a release-blocker issue based on @ppannuto's comment. Given this is the non-breaking version of this change, it seems like we should be able to merge this. Marking as `last-call`.

alevy (approved):

ppannuto:
This isn't the most important thing right now, and I won't stand in the way, but I do think it would be _better_ if we figured out a way to carry real fixes pending major version releases 'in-tree'. The 'leave it as a blocking release issue' approach just means there's a pile of code changes to make when we do want to do a release, and we have to relearn the context for each of those changes at release time rather than just implementing the fix now while the context is fresh. We could leave "unmerged branches" with blocker-issues, but the likelihood of those applying cleanly by release time is also low — we need a means to keep pending fixes up to date with master, which in practice I think necessarily means having them merged in-tree.

This specific example is so trivial, it doesn't really matter, but that's also what makes it a great candidate to test out infrastructure ideas for carrying 'next-major-rev' deprecations.

ppannuto (dismissed):
My desire to stand in the way has grown with the appearance #3613, however.

We need a way to carry code that will change with a future major release.

It would be nicer to prototype with the smaller surface area here, compared to the swath of changes across #3613.

alevy:
@ppannuto, can you articulate what problems would arise from merging this now with the plan to switch to your implementation/semantics for 3.0?

Adding a new upcall number doesn't seem to help the obvious downsides of ^, since it really seems silly to just never ever have an upcall 1 post 3.0 for the console driver, unless we're committed to previous major verison applications running on new major version kernels breaking only in very specific ways (e.g. an upcall will never be invoked, rather than it might be invoked with unexpected argument order)

ppannuto:
I'm worried about death by 1000 cuts when we do want to do a 3.0.

Whenever something comes along that 'demands' or otherwise motivates the major version bump, we have a growing list of 'comparatively insignificant' stuff that _someone_ will have to do first which is wholly unrelated to whatever is motivating the major version release. Each of these 'little' things means paging in a lot of old problems. I would rather we get in the habit of fixing things when they come up. Here it's a question of a little extra code (on an execution path which is fresh in the PR author's head); for #3613, it's a case of asking to keep a (functionally correct!) patch alive (from a non-core contributor!) for an unbounded, arbitrary amount of time—it's not really a reasonable ask.

For this case, it seems like we should add a new upcall that does propagate the error correctly, and update libtock-{c,rs} to use the new upcall [this probably adds a new command as well in practice]. We would ideally mark the legacy code path in some way that makes it a straightforward deletion — the code size impact here with enough of a built-up set of deprecations could even be used as a trigger for a major release perhaps.

I'm realizing that I should just write up an RFC on this... maybe I can pull that off before tomorrow's call

> since it really seems silly to just never ever have an upcall 1 post 3.0 for the console driver

Why? They're just arbitrary numbers, and we certainly have enough to spare.

alevy:
#3433 is closed, fixed in a different PR. Is this PR now unnecessary?

bradjc:
This is still needed. The side effect of the other pr is that the app works without this, but this fix is still correct.

alevy:
Sorry @ppannuto, I think merging this is sufficiently orthogonal to #3622 (which is written up, so doesn't need this PR to be immortalized).

PR #3587, opened 2023-08-01
Use autogenerated register constants for OpenTitan

cfrantz:
### Pull Request Overview

This pull request adds auto-generated register definitions for lowrisc/opentitan peripherals and updates the lowrisc/uart driver to use the autogenerated definitions.

### Testing Strategy

This pull request was tested by executing a tock image on the CW310 FPGA board and observing kernel initialization.

### TODO or Help Wanted

- Follow-on PRs to update other lowrisc/opentitan peripherals to use the autogenerated register definitions.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/chips/earlgrey/src/lib.rs
+++ b/chips/earlgrey/src/lib.rs
@@ -8,6 +8,9 @@
 #![no_std]
 #![crate_name = "earlgrey"]
 #![crate_type = "rlib"]
+// `registers/rv_plic_regs` has many register definitions in `register_structs()!`
+// and requires a deeper recursion limit than the default to fully expand.
+#![recursion_limit = "256"]
 
 pub mod chip_config;
 mod interrupts;
@@ -23,6 +26,7 @@ pub mod i2c;
 pub mod otbn;
 pub mod plic;
 pub mod pwrmgr;
+pub mod registers;
 pub mod spi_host;
 pub mod timer;
 pub mod uart;
--- a/chips/earlgrey/src/registers/README.md
+++ b/chips/earlgrey/src/registers/README.md
@@ -1,0 +1,16 @@
+# Earlgrey Register Definitions
+
+The files in this directory are auto-generated register definitions for
+Earlgrey peripherals.  These files were auto-generated from the OpenTitan
+codebase as follows:
+
+```bash
+$ cd $OPENTITAN_TREE
+$ git checkout earlgrey_es
+$ bazel build //sw/device/tock:tock_earlgrey_registers
+$ tar -C $TOCK_TREE -xvf bazel-bin/sw/device/tock/tock_earlgrey_registers.tar
+```
+
+Note: the existence of a file in this directory does not necessarily mean
+that a tock peripheral implementation exists.  These files are _only_
+the register definitions.
--- a/chips/earlgrey/src/registers/alert_handler_regs.rs
+++ b/chips/earlgrey/src/registers/alert_handler_regs.rs
@@ -1,0 +1,500 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright lowRISC contributors 2023.
+
+// Generated register constants for alert_handler.
+// Built for Earlgrey-M2.5.1-RC1-438-gacc67de99
+// https://github.com/lowRISC/opentitan/tree/acc67de992ee8de5f2481b1b9580679850d8b5f5
+// Tree status: clean
+// Build date: 2023-08-08T00:15:38
+
+// Original reference file: hw/top_earlgrey/ip_autogen/alert_handler/data/alert_handler.hjson
+use kernel::utilities::registers::ReadWrite;
+use kernel::utilities::registers::{register_bitfields, register_structs};
+/// Number of alert channels.
+pub const ALERT_HANDLER_PARAM_N_ALERTS: u32 = 65;
+/// Number of LPGs.
+pub const ALERT_HANDLER_PARAM_N_LPG: u32 = 24;
+/// Width of LPG ID.
+pub const ALERT_HANDLER_PARAM_N_LPG_WIDTH: u32 = 5;
+/// Width of the escalation timer.
+pub const ALERT_HANDLER_PARAM_ESC_CNT_DW: u32 = 32;
+/// Width of the accumulation counter.
+pub const ALERT_HANDLER_PARAM_ACCU_CNT_DW: u32 = 16;
+/// Number of classes
+pub const ALERT_HANDLER_PARAM_N_CLASSES: u32 = 4;
+/// Number of escalation severities
+pub const ALERT_HANDLER_PARAM_N_ESC_SEV: u32 = 4;
+/// Number of escalation phases
+pub const ALERT_HANDLER_PARAM_N_PHASES: u32 = 4;
+/// Number of local alerts
+pub const ALERT_HANDLER_PARAM_N_LOC_ALERT: u32 = 7;
+/// Width of ping counter
+pub const ALERT_HANDLER_PARAM_PING_CNT_DW: u32 = 16;
+/// Width of phase ID
+pub const ALERT_HANDLER_PARAM_PHASE_DW: u32 = 2;
+/// Width of class ID
+pub const ALERT_HANDLER_PARAM_CLASS_DW: u32 = 2;
+/// Local alert ID for alert ping failure.
+pub const ALERT_HANDLER_PARAM_LOCAL_ALERT_ID_ALERT_PINGFAIL: u32 = 0;
+/// Local alert ID for escalation ping failure.
+pub const ALERT_HANDLER_PARAM_LOCAL_ALERT_ID_ESC_PINGFAIL: u32 = 1;
+/// Local alert ID for alert integrity failure.
+pub const ALERT_HANDLER_PARAM_LOCAL_ALERT_ID_ALERT_INTEGFAIL: u32 = 2;
+/// Local alert ID for escalation integrity failure.
+pub const ALERT_HANDLER_PARAM_LOCAL_ALERT_ID_ESC_INTEGFAIL: u32 = 3;
+/// Local alert ID for bus integrity failure.
+pub const ALERT_HANDLER_PARAM_LOCAL_ALERT_ID_BUS_INTEGFAIL: u32 = 4;
+/// Local alert ID for shadow register update error.
+pub const ALERT_HANDLER_PARAM_LOCAL_ALERT_ID_SHADOW_REG_UPDATE_ERROR: u32 = 5;
+/// Local alert ID for shadow register storage error.
+pub const ALERT_HANDLER_PARAM_LOCAL_ALERT_ID_SHADOW_REG_STORAGE_ERROR: u32 = 6;
+/// Last local alert ID.
+pub const ALERT_HANDLER_PARAM_LOCAL_ALERT_ID_LAST: u32 = 6;
+/// Register width
+pub const ALERT_HANDLER_PARAM_REG_WIDTH: u32 = 32;
+
+register_structs! {
+    pub AlertHandlerRegisters {
+        /// Interrupt State Register
+        (0x0000 => pub(crate) intr_state: ReadWrite<u32, INTR::Register>),
+        /// Interrupt Enable Register
+        (0x0004 => pub(crate) intr_enable: ReadWrite<u32, INTR::Register>),
+        /// Interrupt Test Register
+        (0x0008 => pub(crate) intr_test: ReadWrite<u32, INTR::Register>),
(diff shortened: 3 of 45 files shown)


boards/opentitan/earlgrey-cw310/Makefile
```diff
@@ -72,6 +72,10 @@ qemu-app-gdb : $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
 flash: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
 	$(OPENTITAN_TREE)/bazel-bin/sw/host/opentitantool/opentitantool.runfiles/lowrisc_opentitan/sw/host/opentitantool/opentitantool --interface=cw310 bootstrap $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).bin
 
+update_registers:
+	$(OPENTITAN_TREE)/bazelisk.sh build //sw/device/tock:tock_lowrisc_registers
+	tar -C $(TOCK_ROOT_DIRECTORY) -xf $(OPENTITAN_TREE)/bazel-bin/sw/device/tock/tock_lowrisc_registers.tar
+
```
jrvanwhy, line 78: `update-registers` should probably be in its own section of the Makefile, maybe at the bottom. It doesn't really belong with all the "run in X environment" targets, like "flash" and "flash-app".

chips/earlgrey/src/registers/alert_handler_regs.rs
```diff
@@ -0,0 +1,504 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright lowRISC contributors.
```
jrvanwhy, line 3: The Copyright line of the license header should contain a year (as specified in [trd-legal section 4.1](https://github.com/tock/tock/blob/master/doc/reference/trd-legal.md#41-format).

I recommend using one of the following license headers headers for these generated files:

```
// Licensed under the Apache License, Version 2.0 or the MIT License.
// SPDX-License-Identifier: Apache-2.0 OR MIT
// Copyright Tock Contributors 2023.
// Copyright Google LLC 2023.
```

```
// Licensed under the Apache License, Version 2.0 or the MIT License.
// SPDX-License-Identifier: Apache-2.0 OR MIT
// Copyright Tock Contributors 2023.
// Copyright lowRISC Contributors 2023.
```

boards/opentitan/earlgrey-cw310/Makefile
cfrantz: I didn't even realize that snuck into the Makefile.  I'll eliminate it.

chips/lowrisc/src/uart.rs
```diff
@@ -12,97 +12,10 @@ use kernel::hil::uart;
...
-    val [
-        rx OFFSET(0) NUMBITS(16) []
-    ],
-    timeout_ctrl [
-        val OFFSET(0) NUMBITS(23) [],
-        en OFFSET(31) NUMBITS(1) []
-    ]
-];
+#[allow(clippy::wildcard_imports)]
+use crate::registers::uart_regs::*;
```
bradjc, line 18:
```suggestion
use crate::registers::uart_regs::UartRegisters;
use crate::registers::uart_regs::{CTRL, FIFO_CTRL, INTR, STATUS, WDATA};
```

bradjc (commented):
Looks good.

It would be nice to have a readme in the registers directories explaining where these generated files came from, and that the presence of register definitions does not imply a peripheral implementation.

chips/earlgrey/src/registers/alert_handler_regs.rs
cfrantz: I'm a bit confused by this: the register definitions are an export out of the opentitan codebase.  I don't understand how they acquire "Copyright Tock Contributors".

If it were the case that we're assigning copyright to Tock by contributing the definitions, I can understand then adding that copyright notice.  However, section 3 of the trd-legal document states that "No copyright assignment is required...".

I'm happy to add the notice; I just want to understand why.

chips/lowrisc/src/uart.rs
cfrantz: Done.

cfrantz:
> It would be nice to have a readme in the registers directories explaining where these generated files came from, and that the presence of register definitions does not imply a peripheral implementation.

Done.

chips/earlgrey/src/registers/alert_handler_regs.rs
lschuermann: I think @jrvanwhy's comment was more about the format, i.e. including a year in the copyright statement. I believe the above are just examples of how that could look like. This is to support our CI which automatically checks & enforces that copyright notices are present in files in the Tock codebase.

Indeed, most of the rules of `trd-legal` don't apply to code vendored from other projects (which, I presume, includes these auto-generated code snippets). However, given that the files in this PR do include a copyright notice, and that notice is *very* close to our intended format, it would be nice if they could abide to the following rule from `trd-legal`:

> Copyright notices SHOULD include a year.

chips/lowrisc/src/registers/README.md
```diff
@@ -0,0 +1,16 @@
...
+```bash
+$ cd $OPENTITAN_TREE
+$ git checkout Earlgrey-M2.5.2-RC0
+$ bazel build //sw/device/tock:tock_lowrisc_registers
+$ tar -C $TOCK_TREE -xvf bazel-bin/sw/device/tock/tock_lowrisc_registers.tar
+```
+
+Note: the existence of a file in this directory does not necessarily mean
+that a tock peripheral implementation exists.  These files are _only_
+the register definitions.
```
ppannuto, line 16:
```suggestion
Note: the existence of a file in this directory does not necessarily mean
that a tock peripheral implementation exists.  These files are _only_
the register definitions. As an estimate of implementation status, entities
exported in `registers/mod.rs` are those with non-zero implementation.
```

I suspect this will generally hold true?

ppannuto (dismissed):

bradjc (dismissed):

chips/earlgrey/src/registers/README.md
```diff
@@ -0,0 +1,16 @@
+# Earlgrey Register Definitions
+
+The files in this directory are auto-generated register definitions for
+Earlgrey peripherals.  These files were auto-generated from the OpenTitan
+codebase as follows:
+
+```bash
+$ cd $OPENTITAN_TREE
+$ git checkout Earlgrey-M2.5.2-RC0
```
alistair23, line 9: We want to make sure this matches `OPENTITAN_SUPPORTED_SHA` in the Makefile.

Let's get the non README changes (SHA bump and linker script) from https://github.com/tock/tock/pull/3586 merged, then this PR can go next

chips/lowrisc/src/registers/flash_ctrl_regs.rs
```diff
@@ -0,0 +1,660 @@
...
+        (0x01a8 => pub(crate) fifo_rst: ReadWrite<u32, FIFO_RST::Register>),
+        /// Current program and read fifo depth
+        (0x01ac => pub(crate) curr_fifo_lvl: ReadWrite<u32, CURR_FIFO_LVL::Register>),
+        /// Memory area: Flash program FIFO.
+        (0x01b0 => pub(crate) prog_fifo: [WriteOnly<u32>; 1]),
+        /// Memory area: Flash read FIFO.
+        (0x01b4 => pub(crate) rd_fifo: [ReadOnly<u32>; 1]),
+
+
+        (0x0000 => pub(crate) csr0_regwen: ReadWrite<u32, CSR0_REGWEN::Register>),
```
alistair23, line 157: It looks like something broke here

alistair23: @lschuermann  any idea why the tests wouldn't catch this? It's a duplicate value, that is something we should be looking for right?

chips/lowrisc/src/registers/otp_ctrl_regs.rs
```diff
@@ -0,0 +1,712 @@
...
+        (0x0080 => pub(crate) secret1_digest: [ReadWrite<u32, SECRET1_DIGEST::Register>; 2]),
+        /// Integrity digest for the SECRET2 partition.
+        (0x0088 => pub(crate) secret2_digest: [ReadWrite<u32, SECRET2_DIGEST::Register>; 2]),
+        (0x0090 => _reserved1),
+        /// Memory area: Any read to this window directly maps to the corresponding offset in the creator
+        /// and owner software
+        (0x1000 => pub(crate) sw_cfg_window: [ReadOnly<u32>; 512]),
+
+
+        (0x0000 => pub(crate) csr0: ReadWrite<u32, CSR0::Register>),
```
alistair23, line 429: Same here

chips/lowrisc/src/registers/rom_ctrl_regs.rs
```diff
@@ -0,0 +1,56 @@
...
+        /// Alert Test Register
+        (0x0000 => pub(crate) alert_test: ReadWrite<u32, ALERT_TEST::Register>),
+        /// The cause of a fatal alert.
+        (0x0004 => pub(crate) fatal_alert_cause: ReadWrite<u32, FATAL_ALERT_CAUSE::Register>),
+        /// The digest computed from the contents of ROM
+        (0x0008 => pub(crate) digest: [ReadWrite<u32, DIGEST::Register>; 8]),
+        /// The expected digest, stored in the top words of ROM
+        (0x0028 => pub(crate) exp_digest: [ReadWrite<u32, EXP_DIGEST::Register>; 8]),
+        /// Memory area: ROM data
+        (0x0000 => pub(crate) rom: [ReadOnly<u32>; 8192]),
```
alistair23, line 35: and again! Something seems off

alistair23 (changes requested):

chips/lowrisc/src/registers/flash_ctrl_regs.rs
lschuermann: Yes, this is not added to `mod.rs`. We should probably add all of these register files there unconditionally, to make sure they're checked for syntax & type-system errors, and have their const-assertions evaluated.

alistair23: Ah, I didn't notice that. Yeah, these should all be added to the project, we don't want dead code sitting around bit rotting

chips/lowrisc/src/registers/README.md
lschuermann: This allows such issues as identified by @alistair23 in https://github.com/tock/tock/pull/3587#discussion_r1282022178 to slip through. We should add all of them to the `mod.rs` unconditionally, which unfortunately would be in conflict with that statement.

ppannuto: Yep. This wasn't critical, just a nicety.

chips/earlgrey/src/registers/README.md
cfrantz: That is the plan - this was always meant to go after #3586.

chips/lowrisc/src/registers/flash_ctrl_regs.rs
cfrantz: There was a bug in the register generator for peripherals that have more than one register block.  The bug doesn't matter much for C headers because you'd we expected to use the correct base address constant for the particular block.  It is, as you've pointed out, a bug for the rust register structs.

I've corrected the bug in the register generator and added all of the generated files to their respective `mod.rs` (the original intention was to add them as we convert the drivers over to using the autogenerated definitions).

chips/earlgrey/src/lib.rs
```diff
@@ -8,6 +8,7 @@
 #![no_std]
 #![crate_name = "earlgrey"]
 #![crate_type = "rlib"]
+#![recursion_limit = "256"]
```
bradjc, line 11:
```suggestion
// `registers/rv_plic_regs` has many register definitions in `register_structs()!`
// and requires a deeper recursion limit than the default to fully expand.
#![recursion_limit = "256"]
```

cfrantz: Done.  I don't particularly like what the register generator did for the PLIC registers.  I'm leaning towards declaring the PLIC to be an exception to auto-generated registers.

The problem is that the source-of-truth JSON description of the registers individually enumerates all of the `PRIOn` registers when clearly it should be though of as a sequence of length `n` of identical registers.

If we make an exception for the PLIC regs, we can evict `rv_plic_regs.rs` from the codebase and remove the change to the recursion limit.  We can make that decision when we work on converting the individual peripheral/driver implementations over to the auto-generated definitions.

bradjc (approved):

jrvanwhy (approved):

lschuermann (approved):
Skimmed all the auto-generated files and looked more closely at the UART changes. LGTM!

alistair23 (approved):

Issue #695, opened 2017-11-27
[RFC] Tock Attributes


bradjc:
Vision
====

High level goal: all binaries in a Tock system (bootloader, kernel, apps) have metadata stored at or near the beginning of their flash region. The metadata should be coupled with the binary it is relevant to.

Contents at each layer
---------

- **bootloader**: The bootloader metadata should contain two types of data:

    1. State that is immutable about the board. This includes its name, serial number, MCU, etc.
    2. State that the bootloader needs to know. This would include the location in flash of the kernel.

- **kernel**: The kernel metadata holds certain parameters of the kernel that could change over the lifetime of the device and that the kernel needs to know after a power cycle. In particular, this would include the address where apps are located in flash. We use `_sapps` now, but a future system may want to place new apps later in flash and change which apps are active while not losing an older copy of the apps. Another attribute may be the git commit hash or the version of the compiled kernel.

- **apps**: This is the content in the TBF header. It specifies the size of the app and where the first instruction is, among other things.

Advantages to this approach
============

- There is clear parallelism where metadata is stored with each binary.
- Because metadata is stored with the binary, it gets updated when the binary itself gets updated (e.g. we generate a new TBF header when we compile an app, and it gets loaded with the app).
- Any future metadata has a clear place to go.
- It is obvious how Tockloader gets access to any state that it needs.

Disadvantages to this approach
============

- Compiling metadata in with the binary is tricky (ugly in C, really ugly in rust/Cargo).
- We have an entirely different tool (elf2tbf) that generates the metadata for apps.
- Might be clumsy for binaries to have to pull the metadata out of a header in its flash rather than having it in the code somehow.
- Binaries may wish to store their metadata elsewhere rather than having to read/write the MCU flash.

Implementations
===============

We currently have implementations for the bootloader and apps.

Bootloader
----------

All metadata is stored in two flash pages after the vector table for the bootloader. Attributes are stored as key-value pairs, where the key can be up to 8 bytes and the value can be up to 55 bytes (one byte is for the length of the value). There are up to 16 supported attributes.

The attributes are created by a special C struct [hand-crafted](https://github.com/helena-project/tock-bootloader/blob/master/src/attributes.c) in the bootloader source.

Apps
----

All metadata for apps is stored at the beginning of their flash blob. The header is composed of TLV elements, and more documentation is [here](https://github.com/helena-project/tock/blob/master/doc/Compilation.md#tock-binary-format).

[...]

PR #367, opened 2017-05-02
Kernel attributes

ppannuto:
This is a first pass at implementing attribute support for kernel images. It puts kernel attributes immediately after the vector table (which is padded out to the fixed size of 0x400, the max vector table for any arm core; i.e. attributes will always be at "kernel_start + 0x400") but before any code.

The attribute section format follows the same guidelines as the bootloader, the attribute section has:

  - "TockKernel"
  - 1 byte version (currently '0')
  - Attributes:
    - 8 byte name
    - 1 byte length
    - up to 55 byte value

PR #3588, opened 2023-08-01
New approach for kernel attributes

bradjc:
### Pull Request Overview

This is the spiritual successor to #367 and part of #695.

This PR modifies the linker script to insert kernel attributes at the end of the kernel's ROM region. These can be used to store metadata about the kernel that is loaded on the board.

The location at the end of ROM makes them easy to find since they are right before the app flash start address which tockloader already knows about.

#### First use case

The motivation for these is to specify the app RAM region in a way that tockloader can find. Tockloader can then use knowledge about the ram available to processes to choose which version of an app to load onto the board.

#### Downsides

- Every kernel image is now always the full size of the available ROM. This makes loading kernels take longer.

#### Comparison to #367

- This PR puts the attributes at the end of the flash region rather than the beginning. That avoids issues with different boards needing different amounts of data at the start of flash, and makes it easier for tockloader to find (since it already needs to know where to put apps).
- Back when we created the previous PR, kernel attributes were more of a "nice to have" rather than a "must have". That meant we never really wanted to deal with the overhead of creating and maintaining attributes. Now, however, since PIC is _still_ not a thing in LLVM and riscv-gcc, and libtock-rs has no PIC whatsoever, writing interesting apps and using multiple apps with libtock-rs is very difficult. Even when building TBFs for different plausible addresses, loading the correct ones is challenging since we need fixed flash _and_ RAM addresses. I think this means today's need is much more pressing than before.
- We've done a lot with TLVs, and they work for us. We use them in TBF headers and footers. The risk of committing to something now and keeping it stable is low since we've been able to do it with TBFs.
- The changes are entirely contained in the linker script.

### Testing Strategy

Using this with tockloader to load console libtock-rs app.

### TODO or Help Wanted

RFC

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/kernel_layout.ld
+++ b/boards/kernel_layout.ld
@@ -351,11 +351,8 @@ SECTIONS
      *
      * `Length` specifies the size of the value (i.e. not include the header).
      */
-    .attributes :
+    .attributes : AT (ORIGIN(rom) + LENGTH(rom) - SIZEOF(.attributes))
     {
-        . = ABSOLUTE(ORIGIN(rom) + LENGTH(rom) - 16);
-        _sattributes = .;
-
         /* TLV: App Memory
          * This indicates the start address and size of RAM available for
          * userspace apps.
@@ -369,7 +366,6 @@ SECTIONS
         BYTE(79) /* O */
         BYTE(67) /* C */
         BYTE(75) /* K */
-        _eattributes = .;
     }  > rom
 
 
@@ -381,6 +377,30 @@ SECTIONS
     }
 }
 
+/* LOADADDR: Return the absolute load address of the named section. This is
+ * normally the same as ADDR, but it may be different if the AT keyword is used
+ * in the section definition. */
+_sattributes = LOADADDR(.attributes);
+_eattributes = LOADADDR(.attributes) + SIZEOF(.attributes);
+_erom = ORIGIN(rom) + LENGTH(rom);
+
+/* This.. this is a dirty, not-fully-understood, hack. In some way, linking is
+ * a multi-pass process. i.e., if you were to ASSERT(0, "how many links")
+ * you'll get three assertions printed for one invocation of rust-lld.
+ *
+ * Emprically, in the first pass, _eattributes ends up at
+ *     _erom + SIZEOF(.attributes)
+ * So, we need to allow that case.
+ *
+ * That's really not the dream, since we can't identify passes. We really only
+ * want to let the wrong location through on the first pass. Thus, the _best_
+ * solution would be to extract the _erom and _eattributes symbols with readelf
+ * or similar and verify they're equal, but that would be yet-one-more-tool.
+ * Maybe something we can do in CI at some point, and remove this assert.
+ */
+ASSERT((_eattributes == _erom) || (_eattributes == _erom + SIZEOF(.attributes)), "Kernel attributes are not at the end of ROM.")
+
+/* This assert works out becuase even though some of the relative positions are
+ * off, the sizes are sane in each pass. */
 ASSERT((_etext - _stext) + (_erelocate - _srelocate) + (_eattributes - _sattributes) < LENGTH(rom),
 "Text plus relocations plus attributes exceeds the available ROM space.");
-ASSERT(_eattributes == ORIGIN(rom) + LENGTH(rom), "Kernel attributes are not at the end of ROM.")
--- a/boards/kernel_layout.ld
+++ b/boards/kernel_layout.ld
@@ -354,6 +354,7 @@ SECTIONS
     .attributes :
     {
         . = ABSOLUTE(ORIGIN(rom) + LENGTH(rom) - 16);
+        _sattributes = .;
 
         /* TLV: App Memory
          * This indicates the start address and size of RAM available for
@@ -368,6 +369,7 @@ SECTIONS
         BYTE(79) /* O */
         BYTE(67) /* C */
         BYTE(75) /* K */
+        _eattributes = .;
     }  > rom
 
 
@@ -379,5 +381,6 @@ SECTIONS
     }
 }
 
-ASSERT((_etext - _stext) + (_erelocate - _srelocate) < LENGTH(rom), "
-Text plus relocations exceeds the available ROM space.");
+ASSERT((_etext - _stext) + (_erelocate - _srelocate) + (_eattributes - _sattributes) < LENGTH(rom),
+"Text plus relocations plus attributes exceeds the available ROM space.");
+ASSERT(_eattributes == ORIGIN(rom) + LENGTH(rom), "Kernel attributes are not at the end of ROM.")
--- a/boards/kernel_layout.ld
+++ b/boards/kernel_layout.ld
@@ -343,13 +343,14 @@ SECTIONS
      *
      * Attributes are stored as TLVs, but going backwards in flash.
      *
-     *                           2 byte   2 byte   4 byte
-     * -----------+------------+--------+--------+------------+
-     *            | Value...   | Type   | Length | TOCK       | Start of `prog`
-     * -----------+------------+--------+--------+------------+
-     * <-TLVs...-> <-----TLV--------------------> <-Sentinel->
+     *                           2 byte   2 byte   3 byte     1 byte    4 byte
+     * -----------+------------+--------+--------+----------+---------+---------+ Start
+     *            | Value...   | Type   | Length | Reserved | Version | TOCK    | of
+     * -----------+------------+--------+--------+----------+---------+---------+ `prog`
+     * <-TLVs...-> <-----TLV--------------------> <----Header--------> <-Sentl->
      *
-     * `Length` specifies the size of the value (i.e. not include the header).
+     * The TLV `Length` specifies the size of the TLV value (i.e. not including
+     * the TLV header).
      */
     .attributes : AT (ORIGIN(rom) + LENGTH(rom) - SIZEOF(.attributes))
     {
@@ -362,6 +363,15 @@ SECTIONS
         SHORT(0x0101) /* Type = App RAM Region = 0x0101 */
         SHORT(8)      /* Length = 8 bytes */
 
+        /* Version and Reserved. Current version is 0x01.
+         */
+        BYTE(0) /* Reserved */
+        BYTE(0) /* Reserved */
+        BYTE(0) /* Reserved */
+        BYTE(1) /* Version = 0x01 */
+
+        /* Sentinel.
+         */
         BYTE(84) /* T */
         BYTE(79) /* O */
         BYTE(67) /* C */
--- a/boards/kernel_layout.ld
+++ b/boards/kernel_layout.ld
@@ -354,6 +354,15 @@ SECTIONS
      */
     .attributes : AT (ORIGIN(rom) + LENGTH(rom) - SIZEOF(.attributes))
     {
+        /* TLV: Kernel Flash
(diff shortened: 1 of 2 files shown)


alistair23:
> * Every kernel image is now always the full size of the available ROM. This makes loading kernels take longer.

That is a shame, but I do see a lot of value in this

boards/kernel_layout.ld
```diff
@@ -337,6 +337,40 @@ SECTIONS
...
+     * -----------+------------+--------+--------+------------+
+     *            | Value...   | Type   | Length | TOCK       | Start of `prog`
+     * -----------+------------+--------+--------+------------+
+     * <-TLVs...-> <-----TLV--------------------> <-Sentinel->
+     *
+     * `Length` specifies the size of the value (i.e. not include the header).
+     */
+    .attributes :
+    {
+        . = ABSOLUTE(ORIGIN(rom) + LENGTH(rom) - 16);
```
brghena, line 356: (line of code irrelevant) How does Tockloader know where the end of the kernel flash region / start of the apps flash region is?

boards/kernel_layout.ld
```diff
@@ -337,6 +337,40 @@ SECTIONS
...
+        LONG(_sappmem) /* Address of start of app memory. */
+        LONG(_eappmem - _sappmem) /* Length of app memory. */
+        SHORT(0x0101) /* Type = App RAM Region = 0x0101 */
+        SHORT(8)      /* Length = 8 bytes */
+
+        BYTE(84) /* T */
+        BYTE(79) /* O */
+        BYTE(67) /* C */
+        BYTE(75) /* K */
+    }  > rom
```
brghena, line 371: (line of code irrelevant) Do you have a sense for how bad the cost is in terms of time? Maybe just for whatever board you were testing with at least? Presumably it scales with the size of Flash we have reserved for the kernel.

brghena (commented):
I think this is a good idea and a good implementation.

boards/kernel_layout.ld
```diff
@@ -337,6 +337,40 @@ SECTIONS
...
+    {
+        . = ABSOLUTE(ORIGIN(rom) + LENGTH(rom) - 16);
+
+        /* TLV: App Memory
+         * This indicates the start address and size of RAM available for
+         * userspace apps.
+         */
+        LONG(_sappmem) /* Address of start of app memory. */
+        LONG(_eappmem - _sappmem) /* Length of app memory. */
+        SHORT(0x0101) /* Type = App RAM Region = 0x0101 */
```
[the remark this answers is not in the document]
bradjc: Yeah these are not in apps so they are different, but given that the numbers don't actually matter, it seems prudent to just choose different numbers to avoid potential confusion.

boards/kernel_layout.ld
bradjc: 1. Command line flag.
2. Bootloader attributes.
3. Hardcoded into tockloader.
4. `0x30000`

brghena: Fair enough

boards/kernel_layout.ld
bradjc: Well for jlink it doesn't really change anything.

For serial, on hail-master:

```
make program
    Finished release [optimized + debuginfo] target(s) in 0.09s
   text	   data	    bss	    dec	    hex	filename
 112324	      0	  19932	 132256	  204a0	/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/hail
30d94f325136ae6a83cfe968f9a30f2ad4026502991ce8fbe0513b9f4b352299	/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/hail.bin
tockloader  flash --address 0x10000 /Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/hail.bin
[INFO   ] No device name specified. Using default name "tock".
[INFO   ] Using "/dev/cu.usbserial-c098e513000a - Hail IoT Module - TockOS - Hail IoT Module - TockOS".
[STATUS ] Flashing binary to board...
[INFO   ] Padding binary with 320 0xFFs.
100%|█████████████████████████████████████████████████| 220/220 [00:14<00:00, 15.63it/s]
[INFO   ] CRC check passed. Binaries successfully loaded.
[INFO   ] Finished in 16.800 seconds
```

And with this PR:

```
make program
    Finished release [optimized + debuginfo] target(s) in 0.10s
   text	   data	    bss	    dec	    hex	filename
 112036	  19040	  19932	 151008	  24de0	/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/hail
afea7a7e09427cb4799a9f7a480d2c6ea8161dc62fc67d724de768c2cb3d3f45	/Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/hail.bin
tockloader  flash --address 0x10000 /Users/bradjc/git/tock/target/thumbv7em-none-eabi/release/hail.bin
[INFO   ] No device name specified. Using default name "tock".
[INFO   ] Using "/dev/cu.usbserial-c098e513000a - Hail IoT Module - TockOS - Hail IoT Module - TockOS".
[STATUS ] Flashing binary to board...
100%|█████████████████████████████████████████████████| 256/256 [00:16<00:00, 15.63it/s]
[INFO   ] CRC check passed. Binaries successfully loaded.
[INFO   ] Finished in 19.128 seconds
```

So, a little longer.

boards/kernel_layout.ld
```diff
@@ -337,6 +337,40 @@ SECTIONS
...
+        SHORT(8)      /* Length = 8 bytes */
+
+        BYTE(84) /* T */
+        BYTE(79) /* O */
+        BYTE(67) /* C */
+        BYTE(75) /* K */
+    }  > rom
+
```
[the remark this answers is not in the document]
bradjc: I have:
```
ASSERT((_etext-_stext) + (_erelocate-_srelocate) + (_eattributes-_sattributes) < LENGTH(rom),
"Text plus relocations plus attributes exceeds the available ROM space.");
ASSERT(_sattributes > _erelocate, "Kernel attributes overlaps relocations.");
```

I'm not sure it's possible to add more, since if they are the same then whatever is first will trigger and the rest are superfluous. And the linker won't let you overflow rom.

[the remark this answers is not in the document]
brghena: "And the linker won't let you overflow rom." That's great. I didn't think about that. So if we change things and forget to update the 16 it'll already fail at build time.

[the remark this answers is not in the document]
brghena: Ooh, you're right. Although it could have been `_etext + (_erelocate-_srelocate)` then.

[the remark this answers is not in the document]
bradjc: The linker will already check that it doesn't go beyond ROM or overlap other sections.

I added a check that .attributes is actually at the end of ROM.

brghena (approved):

boards/kernel_layout.ld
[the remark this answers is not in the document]
brghena: Yeah, these asserts are good. Thanks for thinking through those.

boards/kernel_layout.ld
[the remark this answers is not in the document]
bradjc: I'm happy to document this should we choose to go forward with it.

ppannuto (dismissed):
Well, I went into this with the naiveté that I understood linkers.

2 hours later, I have something that works, and _no idea_ what TheRightWay(TM) is to handle the multi-pass parsing w.r.t. asserts.

That said, I think sizing `.attributes` automatically is useful, especially if at some point in the future boards start adding their own attributes (which would be trivial; add a `*(.attributes.*)` above the TLVs defined here, and boards can just define `.attributes.hail` or whatever in the linker files already present in their board directories).

bradjc:
I added a version number, another TLV for the kernel binary size, and documented these attributes.

doc/Kernel_Attributes.md
```diff
@@ -0,0 +1,117 @@
...
++----------+----------+----------+----------+
+```
+
+- Type: Indicates which TLV this is. Little endian.
+- Length: The length of the value. Little endian.
+- Value: Length bytes corresponding to the TLV.
+
+## TLVs
+
+### App Memory (0x0101)
```
brghena, line 74:
```suggestion
## TLVs

The TLV types used for kernel attributes are unrelated to the TLV types used for the [Tock Binary Format](./doc/TockBinaryFormat.md#tlv-types). However, to minimize possible confusion type values for each should not use the same numbers.

### App Memory (0x0101)
```

boards/kernel_layout.ld
```diff
@@ -345,5 +396,30 @@ SECTIONS
...
+ * want to let the wrong location through on the first pass. Thus, the _best_
+ * solution would be to extract the _erom and _eattributes symbols with readelf
+ * or similar and verify they're equal, but that would be yet-one-more-tool.
+ * Maybe something we can do in CI at some point, and remove this assert.
+ */
+ASSERT((_eattributes == _erom) || (_eattributes == _erom + SIZEOF(.attributes)), "Kernel attributes are not at the end of ROM.")
+
+/* This assert works out because even though some of the relative positions are
+ * off, the sizes are sane in each pass. */
+ASSERT((_etext-_stext) + (_erelocate-_srelocate) + (_eattributes-_sattributes) < LENGTH(rom),
```
brghena, line 424: Need to add spaces here, per #3601 

```suggestion
ASSERT((_etext - _stext) + (_erelocate - _srelocate) + (_eattributes - _sattributes) < LENGTH(rom),
```

boards/kernel_layout.ld
```diff
@@ -337,6 +337,57 @@ SECTIONS
...
+     * the TLV header).
+     */
+    .attributes : AT (ORIGIN(rom) + LENGTH(rom) - SIZEOF(.attributes))
+    {
+        /* TLV: Kernel Flash
+         * This indicates the start address of the kernel flash and the size of
+         * the kernel binary.
+         */
+        LONG(ORIGIN(rom)) /* Address of start of kernel binary. */
+        LONG((LOADADDR(.relocate) + (_erelocate-_srelocate)) - ORIGIN(rom)) /* Length of kernel binary. */
```
brghena, line 362: Need to add spaces here, per https://github.com/tock/tock/pull/3601

```suggestion
        LONG((LOADADDR(.relocate) + (_erelocate - _srelocate)) - ORIGIN(rom)) /* Length of kernel binary. */
```

brghena (commented):
A couple of small edits here, but I approve of the idea and implementation of this PR

bradjc:
Updated.

brghena (approved):

ppannuto (approved):

doc/Kernel_Attributes.md
```diff
@@ -0,0 +1,122 @@
...
+  * [Kernel Binary (0x0102)](#kernel-binary-0x0102)
+- [Kernel Attributes Location](#kernel-attributes-location)
+
+<!-- tocstop -->
+
+## Format
+
+Kernel attributes are stored in a descending TLV (type-length-value) structure.
+That means they start at the highest address in flash, and are appended in
+descending flash addresses.
```
lschuermann, line 27: Last minute question: what about alignment? I don't think TLVs have any size-alignment constraints, and byte-aligned reads might make reads from within Tock more expensive?

bradjc: Why would tock read this? It already has the linker values in the code. This is for tockloader.

PR #3589, opened 2023-08-01
RFC: Add a <component>_type!() macro in components

bradjc:
### Pull Request Overview

This pull request adds a couple ideas for how to make the `Platform` struct perhaps easier to write. Or maybe copy board-to-board.

Components help with making static_init easy to port from board to board. But editing the Platform struct is always manual.

This PR tries out adding a new macro to components: `component_type!()`, like:

```rust
#[macro_export]
macro_rules! alarm_component_type {
    ($A:ty $(,)?) => {
        capsules_core::alarm::AlarmDriver<
                'static,
                capsules_core::virtualizers::virtual_alarm::VirtualMuxAlarm<
                    'static,
                    $A,
                >,
            >
        };
}
```

That can then be used to create the platform struct:

```rust
pub struct Platform {
    ...
    alarm: &'static components::alarm_component_type!(nrf52840::rtc::Rtc<'static>),
    ...
}
```

The savings aren't _huge_ in general, but with complex stacks (e.g. tickv), the savings would be pretty substantial.

I'm not normally a fan of macros, but in for a penny in for a pound?

### Testing Strategy

travis

### TODO or Help Wanted

RFC

Obviously I didn't do all components, just some as an example.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/components/src/alarm.rs
+++ b/boards/components/src/alarm.rs
@@ -31,6 +31,19 @@ use kernel::component::Component;
 use kernel::create_capability;
 use kernel::hil::time::{self, Alarm};
 
+#[macro_export]
+macro_rules! alarm_component_type {
+    ($A:ty $(,)?) => {
+        capsules_core::alarm::AlarmDriver<
+                'static,
+                capsules_core::virtualizers::virtual_alarm::VirtualMuxAlarm<
+                    'static,
+                    $A,
+                >,
+            >
+        };
+}
+
 // Setup static space for the objects.
 #[macro_export]
 macro_rules! alarm_mux_component_static {
--- a/boards/components/src/alarm.rs
+++ b/boards/components/src/alarm.rs
@@ -70,6 +70,35 @@ macro_rules! alarm_component_static {
     };};
 }
 
+// Setup static space for the objects.
+#[macro_export]
+macro_rules! alarm_component_static_and_type {
+    ($A:ty $(,)?) => {
+        /// Static definition for the Alarm Component.
+        pub mod alarm {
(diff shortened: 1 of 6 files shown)


lschuermann:
Why not something like?

```rust
pub type AlarmComponentType<A> = capsules_core::alarm::AlarmDriver<
    'static,
    capsules_core::virtualizers::virtual_alarm::VirtualMuxAlarm<
        'static,
        A,
    >,
>;
```

bradjc:
That could be better. I generally don't like that rust has `type` so I forgot about it.

I have some affinity for the parallel nature of the _static and _type macros, but might not be better than the language-provided feature.

lschuermann:
I share your distaste of type aliases, but macros have a strict superset of their anti-features. In the interest of readability, debug-ability and the fact that type aliases are at least somewhat type-checked by the compiler at their definition, I would express a strong aversion towards using macros for this purpose.

bradjc:
What about something like https://github.com/tock/tock/pull/3589/commits/a903841270809d71865912b93b8c13b3d9f19ceb

bradjc:
Note: I've been working in the nrf52840dk board because that's what I've been using lately, and honestly it doesn't showcase why this would be helpful very well for two reasons:

1. Many of our capsules use `&'a dyn X`, avoiding types.
2. Many of our components encapsulate a lot of work (such as creating virtual devices) AND use dyn types.

Without dyn types and with more re-usable components, this type of functionality becomes a lot more useful.

bradjc:
Closing in favor of #3657.

PR #3590, opened 2023-08-04
opentitan: tests: flash_ctrl: fixup test order

twilfredo:
### Pull Request Overview

These tests need to run in a particular order, because we are locking certain flash regions (sticky) for negative testing. The test-runner seems to invoke them alphabetically by fn name (new since last time I tested).

### Testing Strategy

Running the tests through Verilator.

### TODO or Help Wanted

There's gotta be a nicer way to do this right? Any ideas? We need to re-order the actual tests order as well. Before it used to run by the order listed in `tests/mod.rs`. Now it seems alphabetical (I think).  Let me know if there's any ways to fix the order. It's needed for tests like the flash_ctrl tests that leave hw in a odd state till next reset.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/opentitan/src/tests/flash_ctrl.rs
+++ b/boards/opentitan/src/tests/flash_ctrl.rs
@@ -13,15 +13,15 @@ use kernel::errorcode::ErrorCode;
 use kernel::hil;
 #[allow(unused_imports)]
 use kernel::hil::flash::Flash;
-#[allow(unused_imports)]
-use lowrisc::flash_ctrl::FlashMPConfig;
 use kernel::hil::flash::HasClient;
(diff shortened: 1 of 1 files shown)


twilfredo:
@alistair23

alistair23:
We should avoid having dependencies on test orders. The orders can change, but also it makes debugging tests difficult for developers, as they might disable a test to speed up run time but cause a surprise breakage.

We should just consolidate the tests into a single test so that we know it will be run in order

twilfredo:
this has been added to https://github.com/tock/tock/pull/3586.

PR #3591, opened 2023-08-04
Update EOL GitHub Actions dependencies

jwnrt:
### Pull Request Overview

This pull request updates GitHub Actions dependencies which use Node v12 and will stop working soon to Node v16.

GitHub [will remove Node v12](https://github.blog/changelog/2023-07-17-github-actions-removal-of-node12-from-the-actions-runner/) from the action runners on the 14th of August.

### Testing Strategy

All the updated workflows run `on: pull_request`, so if this PR passes then the CI still works.

I checked the major-version release notes of the updated actions and didn't see any breaking changes apart from updating Node v12 to Node v16.

I see that [one workflow is using actions/setup-node](https://github.com/tock/tock/blob/master/.github/workflows/ci.yml#L37) so there could be breakage here.

### TODO or Help Wanted

There is one action left that hasn't been updated to Node v16: [actions-rs/toolchain](https://github.com/actions-rs/toolchain/pull/220) which doesn't seem to be maintained any more. It's used in most workflows.

There's a replacement action called [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain/blob/master/action.yml) but I'm not sure it supports using the toolchain specified in `rust-toolchain`.

I think GitHub's container comes with `rustup`, so I think you could drop this action if the repo had a `rust-toolchain.toml` which [specified the required components](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file), but [the update script](https://github.com/tock/tock/blob/08ff147363d1351ed278d3eff5fcbc6c3b402dfc/tools/update_rust_version.sh#L34) currently depends on this being a plaintext file.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -24,12 +24,12 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: actions-rs/toolchain@v1 # pulls version from rust-toolchain file
         with:
           components: rustfmt, clippy
       - name: Set up Python
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v4
         with:
           python-version: '3.x'
       - name: Install dependencies
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,11 +32,11 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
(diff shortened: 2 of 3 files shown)


bradjc (approved):

ppannuto (approved):
This is an easy yes, thanks!

W.r.t. the toolchain version and its action, I'll make a separate PR on that in a little bit — it looks like the new `rust-toolchain.toml` also supports explicit lists of required components and versions; we currently have a bunch of make machinery that that could displace as well, which is exciting.

Issue #3497, opened 2023-06-19
EarlGrey CW310 QEMU `spi_host` test bug


lschuermann:
Over the past few weeks, I've noticed the `spi_host` QEMU CI test fail a couple of times, on PRs which are entirely unrelated to that board / subsystem. We should investigate whether there's a race either in the Tock drivers for that board, or in QEMU. Example of failing workflow: https://github.com/tock/tock/actions/runs/5312037775/jobs/9616026681

alistair23:
@twilfredo

alistair23:
In case the logs get deleted:

```
[SPI] Setup spi_host0 partial_transfer... 

panicked at 'assertion failed: `(left == right)`
  left: `false`,
 right: `true`', boards/opentitan/earlgrey-cw310/src/tests/spi_host.rs:163:5
	Kernel version aab0546

---| No debug queue found. You can set it with the DebugQueue component.

---| OpenTitan Earlgrey configuration for fpga_cw310 |---
---| RISC-V Machine State |---
Last cause (mcause): Machine external interrupt (interrupt=1, exception code=0x0000000B)
Last value (mtval):  0x00000000

System register dump:
 mepc:    0x20021650    mstatus:     0x00000088
 mcycle:  0x2194CE6E    minstret:    0x219505E2
 mtvec:   0x20000501
 mstatus: 0x00000088
  uie:    false  upie:   false
  sie:    false  spie:   false
  mie:    true   mpie:   true
  spp:    false
 mie:   0x00000808   mip:   0x00000000
  usoft:  false               false 
  ssoft:  false               false 
  msoft:  true                false 
  utimer: false               false 
  stimer: false               false 
  mtimer: false               false 
  uext:   false               false 
  sext:   false               false 
  mext:   true                false 
 ePMP regions:
  [0]: addr=0x00000000, end=0x00000000, cfg=0x0 (OFF) (----)
  [1]: addr=0x00000000, end=0x00000000, cfg=0x0 (OFF) (----)
  [2]: addr=0x00000000, end=0x00000000, cfg=0x0 (OFF) (----)
  [3]: addr=0x00000000, end=0x00000000, cfg=0x0 (OFF) (----)
  [4]: addr=0x00000000, end=0x00000000, cfg=0x0 (OFF) (----)
  [5]: addr=0x00000000, end=0x00000000, cfg=0x0 (OFF) (----)
  [6]: addr=0x00000000, end=0x40000000, cfg=0x0 (OFF) (----)
[...]

PR #3597, opened 2023-08-07
arch/rv32i: re-design PMP architecture & implement OpenTitan EarlGrey-specific ePMP

lschuermann:
### Pull Request Overview

This pull-request redesigns the Tock implementation of the RISC-V Physical Memory Protection Unit. It improves on the previous design in the areas of simplicity, modularity, runtime efficiency and testability.

It is split into separate changes, each of which has a separate section illustrating its motivation and design below:
- a general redesign of Tock's RISC-V PMP infrastructure,
- an implementation of the OpenTitan EarlGrey ePMP, conforming to the new APIs introduced in the first commit.

#### RISC-V PMP Redesign

The RISC-V PMP is complex and can be used to enforce memory protection in various modes (machine, supervisor and user mode). Depending on the exact extension set present (e.g., ePMP) and the machine's security configuration bits, it may expose a vastly different set of constraints and application semantics. Chips such as the OpenTitan EarlGrey SoC are using their ePMP hardware to protect against various machine-mode threats, for instance by limiting machine-mode execute permissions to the kernel's text section, and utilizing the "machine-mode whitelist policy" (MMWP) to enforce a deny-by-default memory access policy.

However, this also means that it is virtually impossible to write an efficient, secure, auditable and maintainable PMP implementation which works across this broad spectrum of possible hardware-features, pre-configuration from prior boot stages, and software requirements.

Compared to the previous, monolithic implementation of the RISC-V PMP driver, responsible for hardware initialization and re-configuration, as well as PMP region allocation as part of Tock's memory protection unit (MPU) infrastructure, this commit introduces a layered model, illustrated below:

    +-----------------------------------------------------------------+
    | Kernel                                                          |
    |         							  |
    | Uses an implementation of the `MPU` trait to allocate "MPU      |
    | regions" for userspace processes.                               |
    +-----------------------------------------------------------------+
       |
       | kernel::platform::mpu::MPU
       v
    +-----------------------------------------------------------------+
    | rv32i::pmp::PMPUserMPU                                          |
    |                                                                 |
    | Responsible for implementing the kernel's MPU region allocation |
    | API, bridging this to the constraints of the RISC-V PMP's TOR   |
    | regions. It keeps track of regions, prevents regions from       |
    | overlapping, etc.                                               |
    |                                                                 |
    | It interacts with a `TORUserPMP<NUM_REGIONS>` implementation,   |
    | which exposes an interface to configure a given number of       |
    | user-mode "top of range" (TOR) regions, that can be enabled     |
    | and disabled around a context switch.                           |
    +-----------------------------------------------------------------+
       |     |
       |     | arch::rv32i::pmp::TORUserPMP
       |     v
       |  +-----------------------------------------------------------+
       |  | rv32i::pmp::simple::SimplePMP                             |
       |  |                                                           |
       |  | A "simple" PMP implementation  for chips with no kernel   |
       |  | memory protection or locked regions.                      |
       |  +-----------------------------------------------------------+
       v
    +--------------------------------------------------------------+
    | chips::some_chip::SomePMP                                    |
    |                                                              |
    | An arbitrarily simple or complex custom PMP implementation,  |
    | which supports this chip's set of constraints. It implements |
    | `arch::rv32i::pmp::TORUserPMP`, which requires it to simply  |
    | provide a number of TOR regions within its PMP configuration |
    | space that must be accessible when in user-mode.             |
    +--------------------------------------------------------------+

Advantages of this approach include:
- **Increased flexibility:** Choosing PMP TOR regions as the layer of abstraction allows the `PMPUserMPU` implementation to implement the more complex user-mode memory protection region allocation and management logic, while supporting arbitrary PMP-compliant hardware memory protection units. When implementing machine-mode (kernel) memory protection, these PMP units can simply "slice out" a number of their PMP regions and make those accessible to user-mode. For example, here are two possible PMP configuration layouts under different ePMP protection bits:

      MML = 1, MMWP = 1               | MML = 0, MMWP = 1
      +-----------------------------+ | +-------------------------------+
      | High-priority kernel        | | | ⎡                           ⎤ |
      | mode "lockdown" regions     | | | ⎢ TORUserPMP regions        ⎥ |
      | (e.g., .text with R/X)      | | | ⎣                           ⎦ |
      +-----------------------------+ | +-------------------------------+
      | ⎡                         ⎤ | | | User-mode "deny all"          |
      | ⎢ TORUserPMP regions      ⎥ | | | fallback entry                |
      | ⎣                         ⎦ | | +-------------------------------+
      +-----------------------------+ | | Machine-mode accessible       |
      | Low-priority kernel         | | | locked memory regions         |
      | mode "accessability"        | | +-------------------------------+
      | regions (e.g., ram,         | |
      | flash, MMIO, ...)           | |
      +-----------------------------+ |
                                      |
      An example ePMP configuration   | An ePMP configuration with MMWP
      using the machine-mode lockdown | enabled, but without MML.
      (MML) mode, alongside the       | Machine-mode requires explicit
      machine-mode whitelist policy   | locked regions for all accesses,
      (MMWP). "High-priority" kernel  | but those also apply to user-
      regions are always inaccessible | mode. Hence the TORUserPMP slice
      to user-mode, whereas low-      | is followed by a "deny all" PMP
      priority kernel regions alias   | entry, which limits user-mode
      userspace memory. Machine-mode  | memory accesses to the allocated
      is denied access to user-mode   | regions.
      regions, so the kernel disables |
      the TORUserPMP regions prior to |
      accessing them.                 |
- **Testability**: By implementing the `PMPUserMPU` as its own isolated component, we can write unit-tests for its behavior. For instance, the previous PMP implementation contains a bug where an allocation can overflow the provided memory section. In this new architecture, we can trivially test such behaviors by operating on a mock PMP type implementing the `TORUserPMP` trait. To demonstrate the testability of this architecture, this commit introduces an extensive test around PMP region overlaps.

  To demonstrate the importance of testing this code, we show that through a combination of the `MPU::allocate_region` and `MPU::allocate_app_memory_region` methods, a caller can cause two MPU region allocations to overlap -- this issue is further present in the previous PMP implementation, and the Cortex-M MPU implementation:
  https://github.com/tock/tock/blob/7e9b2703226868e90efc763b7570d6b864fe991b/arch/rv32i/src/pmp.rs#L915-L953
- **Efficiency**: The `TORUserPMP` interface is designed with re-configuration efficiency in mind. Its `configure_pmp` trait accepts PMP TOR regions as a three-tuple consisting of a `pmpcfg` octet, a start and an end address. This `pmpcfg` octet is wrapped in a `TORUserPMPCFG` type, which guarantees -- by construction -- that this octet configures a non-locked TOR-addressed region. By using these type-level guarantees, we can avoid expensive conversions from the `mpu::Permissions` enum or runtime-checks as part of the re-configuration routine.

In terms of the actual PMP implementation, this commit replaces the previous `PMP` struct with a `simple::SimplePMP`. As the name suggests, this new struct is greatly simplified by extracting most of its business-logic into the centralized `PMPUserMPU` implementation. It further implements a now trivial optimization: the `SimplePMP<AVAILABLE_ENTRIES_OVER_TWO>` implements the `TORUserPMP<NUM_REGIONS>` for all `NUM_REGIONS <= AVAILABLE_ENTRIES_OVER_TWO`. By choosing a small `NUM_REGIONS` const-generic, a chip can greatly accelerate the PMP re-configuration routine at the cost of having less MPU regions available.

As a tradeoff, the `SimplePMP` no longer supports kernel-mode memory protection, locked PMP regions, or advanced ePMP security bits (MML, MMWP, RLB). This is a deliberate choice, as supporting these features correctly introduces significant runtime and code complexity, which decreases performance, readability and auditability. To illustrate this, the previous `PMP` implementation contains an implicit assumption that the kernel memory protection regions are configured prior to any userspace regions. Otherwise, a user-space region colliding with a PMP entry for a kernel-mode region can silently not be configured.

Apart from these architectural changes, this commit addresses several other issues, for example:
- It introduces an explicit, documented compatability hack around the previous implementation's memory overshoot bug, as documented in issue https://github.com/tock/tock/issues/3544 [1] (hack in `arch/rv32i/src/pmp.rs:L508-537`). This hack should be removed on the next major kernel release. Removing it breaks existing userspace applications.
- It prevents overshooting the provided memory section for all writeable memory regions, limiting the security impact of this issue.
- It ensures that unused PMP regions are invalidated when unused on re-configuration (as patched in PR https://github.com/tock/tock/pull/3541 [2]), and introduces the aforementioned `NUM_REGIONS <= AVAILABLE_ENTRIES_OVER_TWO` optimization to reduce the runtime overhead of this change.
- It prevents the `SimplePMP` from being instantiated on a system which has any ePMP security bits set. This is a safeguard, as instantiating a PMP implementation not explicitly targeting these configurations will almost certainly render the system unusable.

[1]: https://github.com/tock/tock/issues/3544
[2]: https://github.com/tock/tock/pull/3541

#### OpenTitan ePMP Re-Implementation

This commit reimplements the ePMP support for the OpenTitan EarlGrey SoC, following the previous re-design of the RV32I PMP interfaces. It replaces the dynamic ePMP configuration with a static initialization routine, and integrates with the Tock kernel's MPU infrastructure through the generic `rv32i::pmp::PMPUserMPU` implementation, by means of implementing the `rv32i::pmp::TORUserPMP` trait.

##### The RISC-V ePMP (`-Smepmp`)

The OpenTitan EarlGrey SoC makes extensive use of the RISC-V extended physical memory protection (ePMP, `-Smepmp`) extension. This extension allows using the ePMP to implement extensive memory protection policies in both user-mode and machine-mode. It introduces a new CSR, `mseccfg`, which introduces three new security-relevant configuration bits:

- Machine-Mode Lockdown (MML): enabling this bit changes the rule enforcement semantics of the PMP. For instance, depending on an entry's permission, setting its lock-bit makes the memory region described by said entry inaccessible to user-mode, and exclusively accessible to machine-mode. Furthermore, MML-mode does not support any regions with R/W/X permissions. This is a sticky bit, it cannot be disabled once set.

- Machine-Mode Whitelist Policy (MMWP): this bit control the PMP behavior in case no matching PMP entry was found for a memory access. When this bit is enabled, all memory accesses from within machine-mode must have a corresponding locked PMP entry defined. This is a sticky bit, it cannot be disabled once set.

- Rule-Lock Bypass (RLB): this bit allows overriding the "lock"-bit of the pmpcfg octets. It is useful for handing over a partially-configured ePMP from one execution stage to the next, with MMWP enabled. For instance, the EarlGrey ROM will set up the PMP with MMWP enabled. To hand over execution to Tock, it sets the RLB bit and adds a permissive rule (R/W/X on the entire address space) in the last `pmpaddrX` CSR. Tock can then replace this rule with a more restrictive set of rules and disable RLB. This is an inverse-sticky bit: once a locked rule has been added with RLB disabled, it can never be set again.

For more information on the `Smepmp` extension, refer to the following document:
https://github.com/riscv/riscv-tee/blob/1e9c1ca07ee2eaa5c0379f77a912cb3b2410496f/Smepmp/Smepmp.pdf

##### OpenTitan EarlGrey + Tock ePMP Configuration

To properly harden the Tock kernel on the OpenTitan EarlGrey, the ePMP shall enforce the following machine-mode memory protection rules:
- use the machine-mode lockdown mode, to prevent security issues such as user-mode write + machine-mode excecute,
- read and execute permissions on the kernel's `.text` section,
- read permissions on the entire flash,
- read and write permissions on the entire RAM,
- read and write permissions on the MMIO memory,
- deny other memory accesses.

Userspace applications must have read-write access to their RAM and read-execute access to their flash allocations, and be denied other memory accesses.

Naively implemented, the above rules require 5 of the available 8 TOR regions for machine-mode, leaving 3 TOR regions available to user-mode. This severely restricts more advanced memory-sharing scenarios, such as required with zero-copy IPC.

This commit instead uses a more compact layout with NAPOT regions, configured through a static initialization routine. By making the configuration static and predictable, we can carefully modify the PMP configuration to set up the required rules while occupying only 5 PMP entries, or 3 PMP TOR regions. This leaves 5 regions for userspace. Furthermore, we can increase the configuration resiliency against run-time mis-configurations or attacks by defining certain memory regions as "high-priority lockdown" rules, which are never accessible to user-mode (if we had more ePMP regions available, we would do so for more than just the kernel's `.text`). The `EarlGreyEPMP` struct's documentation contains a visual representation of the final ePMP configuration:
https://github.com/tock/tock/blob/a6466582dcac84ec70758365df61a9a553ee1685/chips/earlgrey/src/epmp.rs#L332-L360

##### Enabling JTAG Debug Access on OpenTitan EarlGrey

The OpenTitan EarlGrey's Ibex CPU further contains a RISC-V Debug Manager (RVDM) component for implementing JTAG debug access. This RVDM peripheral is mapped into the CPU's address space at a NAPOT-region compatible address space. When debug access is required, this memory must be machine-mode accessible with read, write and execute privileges. Unfortunately, the machine-mode lockdown (MML) mode does not support defining such regions. To enable debug access, the ePMP can be configured to not enable MML mode. However, without MML enabled, all locked (machine-mode) regions are also accessible to user-mode. Still, the MMWP bit requires us to define explicit regions for all kernel-mode memory accesses. Hence, in this confiugration, we change the ePMP configuration layout: we first define a set of non-locked user-mode regions. Then we insert a user-mode "deny all" region, which limits user-mode accesses to the preceding non-locked regions. Finally, we can grant access permissions to the kernel through locked regions. The `EarlGreyEPMP` struct's documentation also contains a visual representation of this debug-enabled ePMP configuration:
https://github.com/tock/tock/blob/a6466582dcac84ec70758365df61a9a553ee1685/chips/earlgrey/src/epmp.rs#L427-L456

Switching between debug- and non-debug modes of the ePMP at runtime is generally not feasible: once the MML sticky bit is set, it effectively prevents us from defining the appropriate RVDM memory region until a hart reset. Thus, we introduce a generic parameter to annotate whether the ePMP is configured to operate in debug-mode or not.

Because the non-MML mode requires us to insert the additional "deny-all" user-mode fallback region, it reduces the effective number of available user-mode TOR regions from 5 to 4. This is reflected in the implementations of the `TORUserPMP` traits for the `EPMPDebugEnable` and `EPMPDebugDisable` configurations respectively.

##### Other Changes

Apart from these architectural changes, this implementation reduces the runtime-complexity of the ePMP configuration process. Accessing userspace-memory from the kernel under the MML-mode requires disabling the user-mode memory protection regions first. By using shadow-registers for the `pmpcfgX` octets of those regions, and statically encoding the number of regions and their offsets as constants, enabling and disabling the ePMP on a context switch is reduced to copying these values into their respective `pmpcfgX` CSRs.

##### Tradeoff: Static vs. Dynamic Configuration

While the OpenTitan EarlGrey chip benefits from the increased readability, auditability and ease of hardening of this simple and static ePMP implementation, there is value in a more dynamic approach as used in the previous implementation. This commit removes the previous, generic implementation (`arch/rv32i/src/epmp.rs`), as it was exclusively used by the OpenTitan EarlGrey chip. It might be a good idea to resurrect this implementation if ePMP support becomes more widespread in the RISC-V ecosystem, and vendors can agree on a standard "handover" configuration.

### Testing Strategy

Running it in on the OpenTitan EarlGrey `Earlgrey-M2.5.2-RC0` FPGA target, alongside the changes of #3586.
It needs more testing on other RISC-V boards, and more unit tests for the `PMPUserMPU` implementation.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


arch/rv32i/src/pmp.rs
```diff
@@ -39,306 +35,427 @@ register_bitfields![u8,
...
+impl<const MAX_REGIONS: usize> fmt::Display for PMPUserMPUConfig<MAX_REGIONS> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, " PMP regions:\r\n")?;
-        for (n, region) in self.regions.iter().enumerate() {
-            match region {
-                None => write!(f, "  <unset>\r\n")?,
-                Some(region) => write!(f, "  [{}]: {}\r\n", n, region)?,
-            }
-        }
+        write!(f, " PMP regions: TODO\r\n")?;
```
lschuermann, line 383: TODO: Implement `PMPConfig` formatting.

arch/rv32i/src/pmp.rs
```diff
@@ -490,210 +683,472 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::MPU
...
 
-        // Logical region
-        let mut start = memory_start as usize;
-        let mut size = memory_size;
+    #[test]
+    pub fn test_app_memory_alloc() {
+        // TODO! implement more test cases, such as:
+        //
+        // - Try to update the app memory break with an invalid pointer below
+        //   its allocation's start address.
```
lschuermann, line 964: TODO: More test cases here! Especially w.r.t. to alignment & avoiding memory region overrun.

arch/rv32i/src/pmp.rs
```diff
@@ -490,210 +683,472 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::MPU
...
-                                cfg_val << 24 | (csr::CSR.pmpconfig_get(x / 2) & 0x00FF_FFFF);
-                            csr::CSR.pmpconfig_set(x / 2, new_cfg);
-                            // Lock the CSR
-                            csr::CSR.pmpconfig_modify(x / 2, csr::pmpconfig::pmpcfg::l3::SET);
-                        }
-                        // unreachable, but don't insert a panic
-                        _ => (),
+                    i += 2;
+                } else {
+                    // TODO: check overhead of code
```
lschuermann, line 1107: TODO: Analyze overhead of the implemented PMP configuration routine. Initial results seem to suggest that this is much more efficient than before, but calling out to the `pmpconfig_*` methods is still very costly. We might be able to cut down on runtime cost by using a hand-written ASM jump table for these functions.

arch/rv32i/src/pmp.rs
```diff
@@ -39,306 +35,427 @@ register_bitfields![u8,
...
+    ///   to the next 4-byte boundary.
+    ///
+    /// - third value (`*const u8`): the region's end addres. As a PMP TOR
+    ///   region has a 4-byte address granularity, this address is rounded down
+    ///   to the next 4-byte boundary.
+    ///
+    /// To disable a region, set its configuration to [`TORUserPMPCFG::OFF`]. In
+    /// this case, the start and end addresses are ignored and can be set to
+    /// arbitrary values.
+    fn configure_pmp(
```
cfrantz, line 334: Would is make sens to have these functions return specific error codes for the error conditions they might encounter?

bradjc:
What happens to `kernel/src/platform/mpu.rs::KernelMPU`?

alistair23:
> As a tradeoff, the `SimplePMP` no longer supports kernel-mode memory protection, locked PMP regions, or advanced ePMP security bits (MML, MMWP, RLB).

This seems like a big tradeoff though. I suspect that there will be lots of RICS-V boards where the bootloader or ROM locks itself with PMP. That's what OpenSBI on Linux platforms is already doing (and I suspect what OT will do)

It also means we no longer support ePMP. It looks like RISC-V MCUs will move from PMP to ePMP, I think we need to keep this supported.

I completely agree that the current implementation is less then ideal. It has grown organically and has ended up being clunnky. I do think we need to keep the core features though, things like ePMP and detecting already locked regions are going to be really important in the future

lschuermann:
> > As a tradeoff, the SimplePMP no longer supports kernel-mode memory protection, locked PMP regions, or advanced ePMP security bits (MML, MMWP, RLB).
> 
> This seems like a big tradeoff though. I suspect that there will be lots of RICS-V boards where the bootloader or ROM locks itself with PMP. That's what OpenSBI on Linux platforms is already doing (and I suspect what OT will do)

This is certainly a good point. I absolutely agree that we should support boards with more advanced security configurations. Unfortunately, it seems that the RISC-V ecosystem is still rather fragmented when it comes to which of these security bits are implemented, and in which configuration (locks + security bits) the hardware will be handed over to Tock.

In fact, this PR already introduces support for a platform with a complex and diverse set of security requirements: the OpenTitan EarlGrey chip (which has its own dedicated ePMP implementation in `chips/earlgrey/src/pmp.rs`, interacting with the general PMP infrastructure through the the `TORUserPMP` trait). The EarlGrey ROM hands over the PMP with both MMWP and RLB enabled, and memory access is provided through a R/W/X rule in `pmpaddr15`. Adding to the re-configuration complexity comes the fact that OpenTitan shall use the MMU mode, which requires us to perform an intricate dance to reach our desired, complex PMP layout (as I've tried to illustrate [in the `EarlGreyEPMP` implementation's docs](https://deploy-preview-3597--docs-tockosorg.netlify.app/earlgrey/epmp/struct.earlgreyepmp)).

The reason why the new `SimplePMP` implementation does not (yet) support any of these "advanced" features is that they introduce unpredictable complexity, which makes it hard to write correct and efficient implementations that cover a wide-range of different combinations of locked regions & security bits, *especially so when we don't have a chip for them yet*. It seems feasible to me to extend the `SimplePMP` into an implementation which, for instance, supports a number of locked regions at the end, and plays nicely with MMWP enabled. However, I would like to extend the implementation as we go, and cross this bridge when we get there. And the beauty of the `TORUserPMP` abstraction layer is that, once a PMP setup diverges too much from what we already support, we can introduce a new, dedicated implementation for it.

lschuermann:
> What happens to `kernel/src/platform/mpu.rs::KernelMPU`?

Good question. In my honest opinion, I don't think that this trait captures all necessary information to set up some arbitrary kernel memory protection hardware correctly. While it is rather generic, it does not account for the complex interplay between regions affecting userspace, the kernel, or (depending on the hardware) both. Notably, rule precedence is implicit and implementation dependent in this API: for example, to set up a read-only flash region with a read-execute kernel `.text` section, on RISC-V you would first need to configure the larger flash region, and then the `.text` section. This assumes that the implementation will set up the PMP entries in reverse-order, such that entries added later precede earlier ones. And even then it still does not account for potential locked regions, which may alias some of this memory; entry placement relative to those locked regions is important. Looking at the `allocate_kernel_region` method documentation, statements such as "note that kernel level permissions also apply to apps" are, for instance, simply not true with the ePMP MML mode.

I'd vouch to remove this interface for the time being. If and when we support a generic (but limited to certain "handover configurations") PMP/ePMP implementation with kernel-mode protection capabilities, we should revisit this interface.

chips/earlgrey/src/epmp.rs
```diff
@@ -0,0 +1,1154 @@
...
+
+        // ---------- Check current HW config
+
+        // Ensure that the `mseccfg` CSR has the expected value, namely that
+        // we're in "machine-mode whitelist policy" and have "rule-lock bypass"
+        // enabled. If this register has an unexpected value, we risk
+        // accidentally revoking important permissions for the Tock kernel
+        // itself.
+        if csr::CSR.mseccfg.get() != 0x00000006 {
+            return Err(());
```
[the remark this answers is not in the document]
lschuermann: Yes, I think this would be a strict improvement. During development I used to add explicit panics in these branches around violated assumptions.

alistair23:
> This is certainly a good point. I absolutely agree that we should support boards with more advanced security configurations. Unfortunately, it seems that the RISC-V ecosystem is still rather fragmented when it comes to which of these security bits are implemented, and in which configuration (locks + security bits) the hardware will be handed over to Tock.

The RISC-V spec only defines the PMP/ePMP hardware implementations. So there is always going to be fragmentation around what the PMP state is when Tock is started. I even except it to change on some boards over their lifetime. For example if the ROM on a board supports secure boot, it might be default not configure any PMP. Then once a certificate is flashed/fused the ROM will use the keys and also set PMP bits to protect itself.

> In fact, this PR already introduces support for a platform with a complex and diverse set of security requirements: the OpenTitan EarlGrey chip (which has its own dedicated ePMP implementation in `chips/earlgrey/src/epmp.rs`, interacting with the general PMP infrastructure through the the `TORUserPMP` trait). OpenTitan does not use the `SimplePMP`, but makes use of the `PMPUserMPU`. The EarlGrey ROM hands over the PMP with both MMWP and RLB enabled, and memory access is provided through a R/W/X rule in `pmpaddr15`. Adding to the re-configuration complexity comes the fact that OpenTitan shall use the MMU mode, which requires us to perform an intricate dance to reach our desired, complex PMP layout (as I've tried to illustrate [in the `EarlGreyEPMP` implementation's docs](https://deploy-preview-3597--docs-tockosorg.netlify.app/earlgrey/epmp/struct.earlgreyepmp)).

I do see the justification that OpenTitan is a special DICE like security chip and therefore deserves special treatment. It's unlikely most other RISC-V MCUs will hand over to Tock with such a locked down state. So I think having some special treatment in chips makes sense.

I don't think it's so unbelievably different though that ePMP should only be OT specific. ePMP was recently ratified so I expect future RISC-V board to start supporting it.

> The reason why the new `SimplePMP` implementation does not (yet) support any of these "advanced" features is that they introduce unpredictable complexity, which makes it hard to write correct and efficient implementations that cover a wide-range of different combinations of locked regions & security bits, _especially so when we don't have a chip for them yet_. It seems feasible to me to extend the `SimplePMP` into an implementation which, for instance, supports a number of locked regions at the end, and plays nicely with MMWP enabled. However, I would like to extend the implementation as we go, and cross this bridge when we get there. And the beauty of the `TORUserPMP` abstraction layer is that, once a PMP setup diverges too much from what we already support, we can introduce a new, dedicated implementation for it.

My worry with extending it as we go is that is how we got here :)

I think it would be better to aim to support the features we will need in the future (and now with OT) during the initial design.

Also, I wanted to throw a shout out to still supporting kernel protections with regular PMP. It can be useful for catching overflows and other issues.

lschuermann:
@alistair23 I opened #3599 to address these concerns in the long term. It also contains a practical example of how the current `KernelMPU` trait can actually substantially weaken the system's security when used on a non-MML PMP. I agree with your opinion that we can't just simply hack this support into the PMP infrastructure as we go, it needs to be a properly engineered design which takes all constraints into an account.

This PR demonstrates that the new `PMPUserMPU` implementation and the `TORUserPMP` trait support such advanced configurations (two, actually, one with MMWP && MML and one with MMWP && !MML). Even if we don't stick with the `SimplePMP` in the long-term, I think that this PR may be a substantial improvement over the current design. I'd rather take the time to carefully craft some well-designed interfaces in #3599 and not rush some implementation within this PR.

lschuermann:
The current QEMU CI failures are due to the fact that the EarlGrey ePMP implementation checks the handover configuration prior to attempting to reconfigure any ePMP regions. This is to avoid accidentally running an unsupported (and thus non-validated) configuration, and to prevent certain types of malfunction / attacks (such as trying to reconfigure the ePMP without a hart reset). We can work around that by introducing a flag which disables these checks, and sets up an ePMP configuration identical to the handover configuration the EarlGrey SoC:

```rust
if !HANDOVER_CONFIG_CHECK {
    // We aren't supposed to run a handover configuration check. This is useful for
    // environments which don't replicate the OpenTitan EarlGrey chip behavior
    // entirely accurately, such as QEMU. However, in those environments, we cannot
    // guarantee that this configuration is actually going to work, and not break
    // the system in the meantime.
    //
    // We perform a best-effort configuration, starting by setting rule-lock
    // bypass...
    csr::CSR.mseccfg.set(0x00000004);
    // ...adding our required kernel-mode mode memory access rule...
    csr::CSR.pmpaddr15.set(0x7FFFFFFF);
    csr::CSR.pmpcfg3.set(0x9F000000);
    // ...and enabling the machine-mode whitelist policy:
    csr::CSR.mseccfg.set(0x00000006);
}
```

However, even then QEMU does not work. It seems that there is a bug in its rule-lock bypass (RLB) implementation: while this bit allows changing the `pmpcfgX` octet of a locked rule, it does not allow changing its address:

```rust
        // With the new configuration in place, we can adjust the last region's
        // address to be limited to the MMIO region, ...
        //
        // Current configuration:
        // - mseccfg: 0x00000006 (RLB && MMWP && !MML)
        // - pmpaddr15: 0x7FFFFFFF (all of memory), pmpcfg3[31:24]: 0x9F (L && NAPOT && R && W && X)
        // - mmio.0.napot_addr(): 0x11FFFFFF
        csr::CSR.pmpaddr15.set(mmio.0.napot_addr());
        // - pmpaddr15: 0x7FFFFFFF (unchanged)
        csr::CSR.pmpcfg3.set(0x9B9B9980);
        // - pmpcfg3: 0x9B9B9980 (where pmpcfg3[31:24] has been changed despite being locked)
```

@alistair23 What should I do about this? RLB is a critical feature to setting up the ePMP on the EarlGrey SoC, and I'd rather not have the ePMP configuration diverge in Tock.

arch/rv32i/src/pmp.rs
```diff
@@ -490,210 +683,478 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::MPU
...
-/// This is PMP support for kernel regions
-/// PMP does not allow a deny by default option, so all regions not marked
-/// with the below commands will have full access.
-/// This is still a useful implementation as it can be used to limit the
-/// kernels access, for example removing execute permission from regions
-/// we don't need to execute from and removing write permissions from
-/// executable regions.
-impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::KernelMPU
-    for PMP<MAX_AVAILABLE_REGIONS_OVER_TWO>
-{
```
alistair23, line 583: This should continue to be supported

As https://github.com/tock/tock/issues/3599 points out, this is difficult to add support later. We really should support this in the original design and hopefully avoid the situation we are in now

alistair23 (changes requested):

alistair23:
> I added this patch in a commit to this PR, just to make sure that CI passes otherwise. We'll want to remove this commit prior to merge.

Looks like a bug fix to me. Do you want to send the patch upstream

arch/rv32i/src/pmp.rs
lschuermann: I really think that the trait-based structure proposed here would allow us to add an implementation for this later, but I acknowledge that me just making this claim is less convincing than proof in the form of an actual implementation. Let me whip up an implementation which still supports this (albeit less flexible than previously).

lschuermann:
Apologies for the delay with this. @alistair23 I added an implementation of the PMP kernel memory protection with and without relying on the ePMP MML bit respectively. I made sure to use each of those implementations with at least one board in tree. Also, I bumped the pinned QEMU version to pull in the fix around ePMP RLB bit enforcement.

Unfortunately, during development I noticed that this QEMU release is more likely to cause the SPI test bug (https://github.com/tock/tock/issues/3497). I don't believe that this behavior is correlated with these changes. Given that we're regularly observing this issue on the current pinned QEMU commit as well, I would suggest temporarily disabling this test / ignoring it for this changeset.

A more critical issue is that I believe the VexRiscv PMP implementation to be broken: its enforcement of locked regions in machine-mode does not work. I will debug this issue in the coming days. This issue has been spotted in CI (which is great!), and fixing this bug should allow us to have a CI-tested board exercising the non-ePMP kernel memory protection enforcement.

Apart from these two issues, and addressing the minor review comments above, these changes should be ready. I recommend reviewing them commit-by-commit.

lschuermann:
> A more critical issue is that I believe the VexRiscv PMP implementation to be broken

Update: after a bunch of testing & investigating, the CPU's PMP implementation is indeed broken (integer overflow bug). Will propose a fix to upstream, then update the LiteX packages, generate a new target, build images, and then have a CI target which can actually test the non-ePMP kernel memory protection feature. :)

lschuermann:
It took a while, but I finally got all dependencies updated. With this new `tock-litex` revision and the patched CPU, we now get:

```
Starting test. This will compile a Verilated LiteX simulation and thus might take a bit...
We're up! Got the kernel greeting. Connecting to ZeroMQ simulation control socket...
Walking flash worked.
Set button input to true, currently driven by input.
Expecting to overrun flash region.
Walking RAM worked, walking flash now, will overrun!
Process faulted.
Process had a load access fault, as expected.
Test succeeded!
```

(This is with the non-ePMP `KernelProtectionPMP` module.)

I'll split the LiteX update itself into a different PR. This should be good to go modulo some cosmetic fixes, and LiteX being updated.

arch/rv32i/src/pmp.rs
lschuermann: While I agree that we should ultimately have an error enumeration here, currently we can't even meaningfully handle these errors (the kernel's MPU trait and infrastructure assumes these operations to be infallible). I think that we should introduce more meaningful error variants once we change the general MPU interfaces to support erroring on an MPU configuration failure and acting accordingly (e.g., by suspending / failing the process instead of panicing).

arch/rv32i/src/pmp.rs
lschuermann: I added the `KernelProtectionPMP`, which demonstrates that the abstractions proposed here are flexible enough to continue supporting this type of PMP implementation.

arch/rv32i/src/pmp.rs
```diff
@@ -486,210 +838,1388 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::MPU
...
-                    .value;
-
-                let (region_shift, other_region_mask) = if x % 2 == 0 {
-                    (0, 0xFFFF_0000)
-                } else {
-                    (16, 0x0000_FFFF)
-                };
+        if !self.last_configured_for.contains(&config.id) || config.is_dirty.get() {
+            self.pmp.configure_pmp(&config.regions).unwrap()
+        }
```
lschuermann, line 870: This code currently always reconfigures the PMP, it's missing the code to remember whether a given `MpuConfig` has already been configured:

```suggestion
        if !self.last_configured_for.contains(&config.id) || config.is_dirty.get() {
            self.pmp.configure_pmp(&config.regions).unwrap()
            config.is_dirty.set(false);
            self.last_configured_for.set(config.id);
        }
```

bradjc:
Litex updates being merged

lschuermann:
> Litex updates being merged

Awesome! Rebased this PR.

bradjc (approved):

lschuermann:
I realize this is quite a large changeset, but how can we push this over the finish line? Should this get more reviews from @tock/core-wg or the OpenTitan WG? This does change / remove some core kernel APIs (`MPU` and `KernelMPU`) and security-relevant architecture code.

boards/opentitan/earlgrey-cw310/Makefile
```diff
@@ -60,16 +60,16 @@ endif
 install: flash
 
 qemu: $(TOCK_ROOT_DIRECTORY)target/$(TARGET)/release/$(PLATFORM).elf
-	$(QEMU) -M opentitan -kernel $^ -nographic -serial mon:stdio -global driver=riscv.lowrisc.ibex.soc,property=resetvec,value=${QEMU_ENTRY_POINT}
+	$(QEMU) -M opentitan -kernel $^ -nographic -serial mon:stdio -global driver=riscv-cpu,property=priv_spec,value=v1.12.0 -global driver=riscv.lowrisc.ibex.soc,property=resetvec,value=${QEMU_ENTRY_POINT}
```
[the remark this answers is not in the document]
lschuermann: We don't, apparently! This works without the spec version argument on QEMU 8.1.3 and 8.2.0.

boards/qemu_rv32_virt/Makefile
```diff
@@ -67,6 +67,7 @@ QEMU_BASE_CMDLINE := \
   $(QEMU_CMD) \
     -machine virt \
     -semihosting \
+    -global driver=riscv-cpu,property=x-epmp,value=true \
```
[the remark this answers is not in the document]
lschuermann: I'm not sure how to specify that (maybe 8.2.0 is too old still)? `-global driver=riscv-cpu,property=sepmp,value=true` doesn't work. When I tried to find this argument I ended up spending hours reading through the QEMU object parameters and guessing the CLI arguments, can you enlighten me on what to pass exactly? Thanks!

bradjc (approved):

alistair23 (approved):

PR #3598, opened 2023-08-09
Update OpenTitan to use auto-generated base address constants

cfrantz:
### Pull Request Overview

This pull request changes the OpenTitan/Earlgrey peripheral definitions to use auto-generated constants from the OpenTitan codebase.

### Testing Strategy

This pull request was tested by observing the Tock kernel boot.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


jrvanwhy (dismissed):

alistair23 (dismissed):

chips/earlgrey/src/top_earlgrey.rs
```diff
@@ -0,0 +1,3188 @@
...
+
+// Built for Earlgrey-M2.5.1-RC1-389-g21ce4e9761
+// https://github.com/lowRISC/opentitan/tree/21ce4e9761abdf5c919b46e5ae64a5a8e24992f7
+// Tree status: clean
+// Build date: 2023-07-20T09:36:12.725946
+
+// This file was generated automatically.
+// Please do not modify content of this file directly.
+// File generated by using template: "toplevel.rs.tpl"
+// To regenerate this file follow OpenTitan topgen documentations.
```
lschuermann, line 13: I'm wondering whether it might be better to move this file into `registers`, to keep all auto-generated code contained in a single module (as it essentially only relates to registers' base offsets). However, I don't want to voice a strong opinion either way, to me the current module path is also fine.

lschuermann (approved):

chips/earlgrey/src/top_earlgrey.rs
```diff
@@ -0,0 +1,3188 @@
...
+pub const TOP_EARLGREY_UART0_BASE_ADDR: usize = 0x40000000;
+
+/// Peripheral size for uart0 in top earlgrey.
+///
+/// This is the size (in bytes) of the peripheral's reserved memory area. All
+/// memory-mapped registers associated with this peripheral should have an
+/// address between #TOP_EARLGREY_UART0_BASE_ADDR and
+/// `TOP_EARLGREY_UART0_BASE_ADDR + TOP_EARLGREY_UART0_SIZE_BYTES`.
+pub const TOP_EARLGREY_UART0_SIZE_BYTES: usize = 0x40;
+/// Peripheral base address for uart1 in top earlgrey.
```
bradjc, line 42: Spacing could be better. There are not newlines between peripherals.

chips/earlgrey/src/top_earlgrey.rs
bradjc, line 1: What is "top"? Is there a "bottom"? This distinguisher seems out of place in this codebase.

Why not call this file "constants.rs"?

chips/earlgrey/src/top_earlgrey.rs
bradjc: Moving it to registers/ seems appropriate.

cfrantz: Moved.

chips/earlgrey/src/top_earlgrey.rs
cfrantz: A "top" (or top-level) is a term of art in hardware design.  It refers to the top-level module that pulls in all other modules and constructs the final design.  In OpenTitan, there is some automation in binding all of the modules together into the top-level and the `top_<design>.h` (and `.rs`) files are outputs of that tool.

Because this file is auto-generated, I'd prefer to not rename it so that if it ever needs to be updated in the future (e.g. a correction to a constant, or addition of more consts) its obvious which file to replace.

chips/earlgrey/src/top_earlgrey.rs
cfrantz: This looks like a typo in the generating template - an extra newline was omitted in the register block.

chips/earlgrey/src/top_earlgrey.rs
bradjc: I don't see why we want that hardware-design detail to leak into the OS. From my point of view earlgrey is a chip like any other, and if that is common in the industry then I imagine all other chips were "top"s too. But if that is how this chip is identified (i.e. matches the datasheet), then I would support renaming the chip crate to "top_earlgrey".

We could also have our own tock-specific autogen script that converts the name.

Finally, why would this need to be re-executed? I suppose a revision of the chip might change it, but that is presumably rare. Given these are HW constants this seems like exactly the file that is generated once and never touched.

lschuermann (approved):

chips/earlgrey/src/top_earlgrey.rs
ppannuto: Long-term, I agree with @bradjc here. We shouldn't leak HDL conventions to the SW interface.

In the now, it looks like OT doesn't have a datasheet yet, or really anything even close to one, at least not publicly. The "datasheet" link I found on a quick google tells you what it really is in its URL: https://opentitan.org/book/hw/top_earlgrey/doc/specification.html :) [which, also leaks "top"]

Short term, I think it's probably okay to pull this as-is to help keep OT's tree more in sync with upstream, so long as it's understood by all that eventually it'll need to be updated to the actual naming convention that release silicon chooses to use.

cfrantz: Would a bit of module renaming-foo be an acceptable workaound?

chips/earlgrey/src/registers/mod.rs:
```
pub mod top_earlgrey.rs;
pub use top_earlgrey as constants;
```

As to why the file might need to be regenerated: All human beings, without exception, make mistakes.

I don't think it is particularly likely that any of the base addresses are wrong, but there is the possibility that there are some other consts or enumerated definitions in the top-level that we aren't exposing that we may later determine that we need.  In such a case, we'd regenerate this file and re-export it to tock.

alistair23 (approved):

alevy (approved):

chips/earlgrey/src/top_earlgrey.rs
cfrantz: BTW, I haven't applied the module-rename workaround - I was asking if that sort of rename would be acceptable to hide the `top_earlgrey` module name.

twilfredo (approved):

chips/earlgrey/src/top_earlgrey.rs
ppannuto: I realize I'm answering this post-merge, but long-term I think the answer is that "top_" shouldn't appear anywhere in Tock, or really any part of the software-facing ecosystem around OT.

The notion of a top-level module stops at the VLSI/HDL engineer—above that, it's just "a chip".

AFAICT, this chip from the OpenTitan chip manufacturer is called EarlGrey, just as Nordic makes an nrf52840 or TI makes a series of chips in the MSP430 family, such as the MSP430FR2353. "top_" doesn't appear in any of their public-facing datasheets or software libraries, nor I expect will "top_" appear in anything OT produces targeting the software ecosystem by this time next year or whatever.

Eventually, part of the tooling for 'automated export from hardware ecosystem to software ecosystem' will include logic to adapt conventions and naming from what bottom-up folks work with to what's expected from a top-down engineer. IME that's just a part of 'growing up' for EE/chip startups that comes later down the pipe as it's not critical for early-stage functional correctness, but is important for widespread adoption. If the first thing I do when evaluating a chip to choose is have to ask "top what?" to even figure out what chip I'm working with, and there isn't really a meaningful answer, that's a pretty major red flag for me as a designer that this is (pre-?)alpha silicon and I shouldn't use it for anything "real" (which, of course, is right where OT is right now! :) )

Issue #3572, opened 2023-07-27
Review Clippy Flags


bradjc:
We run clippy as part of CI, with the following flags:

https://github.com/tock/tock/blob/910821abb95417997798ce615de1e9b2d6727670/tools/run_clippy.sh#L26-L43

These flags were chosen when we added clippy to minimize the amount of re-writing of code. However, we should revisit these flags to increase the number of checks clippy is helping us with by improving the source tree incrementally.

bradjc:
This is where we are at now:

https://github.com/tock/tock/blob/3df5b00e00121f645740ec39441f93cb101feec1/tools/run_clippy.sh#L20-L208

I think we should continue to remove clippy allows as appropriate. But I think the spirit of this issue from TW is finished: we are now very explicit about which lints we specifically allow.

PR #3607, opened 2023-08-11
tools: clippy: list all complexity and style lints

bradjc:
### Pull Request Overview

This lists out which complexity and styles lints we currently need to allow. I grouped them based on lints that seem like reasonable issues we should fix, lints we may want to fix depending on the diff, and lints we most likely will not use.

Part of #3572.

This is potentially an important change because we are adding many clippy lints that we weren't previously.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/tools/run_clippy.sh
+++ b/tools/run_clippy.sh
@@ -17,9 +17,7 @@
 #
 # - `clippy::if_same_then_else`: There are often good reasons to enumerate
 #   different states that have the same effect.
-
 CLIPPY_ARGS="
--A clippy::complexity
 -A clippy::pedantic
 -A clippy::nursery
 -A clippy::style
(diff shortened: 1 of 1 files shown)


brghena (approved):

tools/run_clippy.sh
```diff
@@ -37,4 +34,123 @@ CLIPPY_ARGS="
 -D clippy::wildcard_imports
 "
 
-cargo clippy -- $CLIPPY_ARGS
+# Disallow all complexity lints, then re-allow each one Tock does not comply
+# with.
+#
+# There are three sections:
+# 1. The first section are lints we almost certainly don't want.
+# 2. The section section are lints we may not want, we probably have to see the
```
hudson-ayers, line 42:
```suggestion
# 2. The second section are lints we may not want, we probably have to see the
```

tools/run_clippy.sh
```diff
@@ -37,4 +34,123 @@ CLIPPY_ARGS="
...
+#    resulting diff.
+# 3. The third section are lints that we do want we just need to fixup the code
+#    to pass the lint checks.
+CLIPPY_ARGS_COMPLEXITY="
+-D clippy::complexity
```
hudson-ayers, line 47: I might put the complexity link on a separate line, at a glance it looks like you accidentally denied one of the lints from the first section that you meant to allow

hudson-ayers (approved):
I don't have a ton of thoughts on this particular grouping, but this approach seems good to me. I assume we already pass a lot of these

bradjc:
> I assume we already pass a lot of these

We pass none of these. That is how I found them.

ppannuto (approved):

bradjc:
Do we want to move forward with this? This is a step towards using clippy more generally. I have a few branches with fixes/changes to meet various clippy lints. Luckily `clippy --fix` is a thing.

lschuermann (approved):

PR #3611, opened 2023-08-11
doc: Update compilation document with information about fixed-address apps

bradjc:
### Pull Request Overview

Opening this now in case the writeup helps with understanding how we can support fixed addresses. But this documentation isn't quite true today given the outstanding PRs.

### Testing Strategy

n/a

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/doc/Compilation.md
+++ b/doc/Compilation.md
@@ -164,7 +164,7 @@ compiled binary.
 Tock supports resource constrained microcontrollers which do not support virtual
 memory. This means Tock process cannot assume a known address space. Tock
 supports two methods for enabling processes despite the lack of virtual memory:
-embedded PIC (ePIC) and fixed address loading.
+embedded PIC (FDPIC) and fixed address loading.
 
 #### Position Independent Code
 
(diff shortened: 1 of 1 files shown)


alistair23 (dismissed):

doc/Compilation.md
```diff
@@ -129,27 +147,26 @@ in the stm32f412gdiscovery.elf kernel ELF, and creates a new .elf called
...
+compiled binary.
 
-The first requirement is explained directly below while the other two are
-detailed in [Tock Binary Format](#tock-binary-format).
+### Executing without Virtual Memory
 
+Tock supports resource constrained microcontrollers which do not support virtual
+memory. This means Tock process cannot assume a known address space. Tock
+supports two methods for enabling processes despite the lack of virtual memory:
+embedded PIC (ePIC) and fixed address loading.
```
ppannuto, line 167:
```suggestion
Position Independent Code (PIC) and fixed address loading.
```

I don't think "ePIC" is a thing (yet?). I tried looking it up, and the only thing I could find was a [whitepaper](https://riscv-europe.org/media/proceedings/posters/2023-06-08-Hugo-MCNALLY-abstract.pdf) and [fairly inactive repo](https://github.com/luismarques/epic-spec/) talking about an ABI proposal for RISC-V. We use bog-standard PIC; all the "ePIC" changes should be reverted.

ppannuto (changes requested):
Mostly great, just the ePIC thing

doc/Compilation.md
bradjc: But we don't use standard PIC because of the magic `-mno-pic-data-is-text-relative` flag.

https://github.com/riscv-collab/riscv-gcc/issues/158

ppannuto: Huh, looks like "FDPIC" might be the right word based on the comment linked there, and a quick google that turns up a ton of info on FDPIC

bradjc: Ok that's right. Updated.

ppannuto (approved):

alistair23 (approved):

doc/Compilation.md
jrvanwhy: I don't think we need to update the document, but this isn't quite right. FDPIC is distinct from the ePIC that lowRISC is working on, which is distinct from the type of PIC we use on ARM in `libtock-c`. I don't think there's any standard name for the type of PIC that `libtock-c` uses on ARM.

For reference, FDPIC supports shared libraries, while ePIC does not.

PR #3613, opened 2023-08-13
capsules: ensure driver existence check

### LINKED ISSUE #3375: All drivers should return `Success` rather than `SuccessU32` for command 0

Samir-Rashid:
### Pull Request Overview

The driver liveliness check is defined in TRD104 4.3.1. I have updated all drivers in advance of Tock v3 (#3375). Make code comments and documentation consistent with regards to "driver existence check" terminology -- change as per #2908. "Driver check" falsely implies some sort of functionality checking.

### Testing Strategy

Unsure how to test. These are breaking changes.

### TODO or Help Wanted

There seems to be an established convention that command code 0 should return the number of things the driver is controlling. I moved those commands to the next available command code. Should these syscalls have some standardized code number?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/core/src/adc.rs
+++ b/capsules/core/src/adc.rs
@@ -1186,10 +1186,10 @@ impl<'a, A: hil::adc::Adc<'a> + hil::adc::AdcHighSpeed<'a>> SyscallDriver for Ad
             return CommandReturn::failure(ErrorCode::NOMEM);
         }
         match command_num {
-            // check if present
+            // Driver existence check
             // TODO(Tock 3.0): TRD104 specifies that Command 0 should return Success, not SuccessU32,
             // but this driver is unchanged since it has been stabilized. It will be brought into
-            // compliance as part of the next major release of Tock.
+            // compliance as part of the next major release of Tock. See #3375.
             0 => CommandReturn::success_u32(self.channels.len() as u32),
 
             // Single sample on channel
--- a/capsules/core/src/gpio.rs
+++ b/capsules/core/src/gpio.rs
@@ -178,7 +178,7 @@ impl<'a, IP: gpio::InterruptPin<'a>> SyscallDriver for GPIO<'a, IP> {
     ///
     /// ### `command_num`
     ///
-    /// - `0`: Number of pins.
+    /// - `0`: Driver existence check.
     /// - `1`: Enable output on `pin`.
     /// - `2`: Set `pin`.
     /// - `3`: Clear `pin`.
@@ -188,6 +188,7 @@ impl<'a, IP: gpio::InterruptPin<'a>> SyscallDriver for GPIO<'a, IP> {
     /// - `7`: Configure interrupt on `pin` with `irq_config` in 0x00XX00000
     /// - `8`: Disable interrupt on `pin`.
     /// - `9`: Disable `pin`.
+    /// - `10`: Get number of GPIO ports supported.
     fn command(
         &self,
         command_num: usize,
@@ -198,8 +199,8 @@ impl<'a, IP: gpio::InterruptPin<'a>> SyscallDriver for GPIO<'a, IP> {
         let pins = self.pins;
         let pin_index = data1;
         match command_num {
-            // number of pins
-            0 => CommandReturn::success_u32(pins.len() as u32),
+            // Check existence.
+            0 => CommandReturn::success(),
 
             // enable output
             1 => {
@@ -330,6 +331,9 @@ impl<'a, IP: gpio::InterruptPin<'a>> SyscallDriver for GPIO<'a, IP> {
                     }
                 }
             }
+
+            // number of pins
+            10 => CommandReturn::success_u32(pins.len() as u32),
 
             // default
             _ => CommandReturn::failure(ErrorCode::NOSUPPORT),
--- a/capsules/extra/src/analog_comparator.rs
+++ b/capsules/extra/src/analog_comparator.rs
@@ -118,7 +118,7 @@ impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> SyscallDriver
     ///
     /// ### `command_num`
     ///
-    /// - `0`: Driver check.
+    /// - `0`: Driver existence check.
     /// - `1`: Perform a simple comparison.
     ///        Input x chooses the desired comparator ACx (e.g. 0 or 1 for
     ///        hail, 0-3 for imix)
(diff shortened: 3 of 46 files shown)


bradjc:
It's great that this is a comprehensive change because we do want cmd 0 to be a way to check if the driver exists. We don't want people starting from capsules that do not follow this convention so it's great to remove this broadly.

We can't change the cmd==0 for the drivers with a check in the "2.0" column here: https://github.com/tock/tock/blob/master/doc/syscalls/README.md As the comments indicate we can't change those until Tock 3.0.

For the others we can change, and I think we should.

> Should these syscalls have some standardized code number?

I think generally using "1" would make sense, but I don't think it's worth changing existing drivers to match. Just adding them as the next available number makes sense to me for existing drivers.

bradjc (changes requested):
We can't break the stability guarantees.

Samir-Rashid:
As I reverted stable drivers, I noticed there is no PWM driver [listed](https://github.com/tock/tock/blob/master/doc/syscalls/README.md#hardware-access). It should be there as it has an allocated driver number 10 (in hex?).

hudson-ayers:
> As I reverted stable drivers, I noticed there is no PWM driver [listed](https://github.com/tock/tock/blob/master/doc/syscalls/README.md#hardware-access). It should be there as it has an allocated driver number 10 (in hex?).

Good catch, feel free to add that in this PR or another standalone one!

capsules/core/src/adc.rs
```diff
@@ -1191,10 +1191,10 @@ impl<'a, A: hil::adc::Adc<'a> + hil::adc::AdcHighSpeed<'a>> SyscallDriver for Ad
             return CommandReturn::failure(ErrorCode::NOMEM);
         }
         match command_num {
-            // check if present
+            // Driver existence check
             // TODO(Tock 3.0): TRD104 specifies that Command 0 should return Success, not SuccessU32,
             // but this driver is unchanged since it has been stabilized. It will be brought into
-            // compliance as part of the next major release of Tock.
+            // compliance as part of the next major release of Tock. See #3613.
```
bradjc, line 1197: Can we point to #3375 instead? That is the tracking issue that will remain open for a possible 3.0 release.

capsules/core/src/gpio.rs
```diff
@@ -198,8 +199,8 @@ impl<'a, IP: gpio::InterruptPin<'a>> SyscallDriver for GPIO<'a, IP> {
         let pins = self.pins.as_ref();
         let pin_index = data1;
         match command_num {
-            // number of pins
-            0 => CommandReturn::success_u32(pins.len() as u32),
+            // Check existence.
+            0 => CommandReturn::success(),
```
bradjc, line 203: I think we need PRs open for libtock-c and libtock-rs to make this change.

https://github.com/tock/libtock-c/blob/master/libtock/gpio.c
https://github.com/tock/libtock-rs/blob/master/apis/gpio/src/lib.rs#L64

capsules/extra/src/analog_comparator.rs
```diff
@@ -136,8 +137,8 @@ impl<'a, A: hil::analog_comparator::AnalogComparator<'a>> SyscallDriver
         processid: ProcessId,
     ) -> CommandReturn {
         if command_num == 0 {
-            // Handle this first as it should be returned unconditionally.
-            return CommandReturn::success_u32(self.channels.len() as u32);
+            // Handle unconditional driver existence check.
+            return CommandReturn::success();
```
bradjc, line 141: Need pr: https://github.com/tock/libtock-c/blob/master/libtock/analog_comparator.c#L8

capsules/extra/src/gpio_async.rs
```diff
@@ -169,9 +170,14 @@ impl<Port: hil::gpio_async::Port> SyscallDriver for GPIOAsync<'_, Port> {
         let other = (data >> 16) & 0xFFFF;
         let ports = self.ports.as_ref();
 
-        // Special case command 0; everything else results in a process-owned,
-        // split-phase call.
+        // Driver existence check.
         if command_number == 0 {
```
bradjc, line 174: We don't use this in libtock-c so we are good here.

capsules/extra/src/read_only_state.rs
```diff
@@ -127,8 +128,11 @@ impl<'a, T: Time> SyscallDriver for ReadOnlyStateDriver<'a, T> {
         _processid: ProcessId,
     ) -> CommandReturn {
         match command_number {
-            // get version
-            0 => CommandReturn::success_u32(VERSION),
+            // Check existence
+            0 => CommandReturn::success(),
+
+            // Get version
+            1 => CommandReturn::success_u32(VERSION),
```
bradjc, line 135: https://github.com/tock/libtock-c/blob/master/libtock/read_only_state.c#L3

Samir-Rashid:
Apology for delay. [Here](https://github.com/Samir-Rashid/tock/commit/ce1d50804fcf85a2779ec62503b50073838b24f8) is a diff of what I changed in case GitHub gets confused when I squash again. I made accompanying PRs https://github.com/tock/libtock-c/pull/349 and https://github.com/tock/libtock-rs/pull/517

doc/syscalls/00007_analog_comparator.md
```diff
@@ -55,7 +55,7 @@ the index of the last AC module.
 
     **Returns**: `Ok(())` if starting interrupts was succesful.
 
-* ### Command number: `4`
+* ### Command number: `3`
```
ppannuto, line 58: Just noting for myself: this is not a change per se, but a fix of a documentation bug

ppannuto (approved):
I think this is looking pretty good. It's a lot of changes, but it's a nice consistency pass over a pretty old part of the kernel interface. It doesn't look like any stabilized interfaces are changed (though some of the documentation gets small updates in the name of consistency—which is good IMO).

bradjc (approved):

PR #3615, opened 2023-08-14
Rewritten nRF52840 802.15.4 driver adding software ACK on receive

tyler-potyondy:
### Pull Request Overview

This pull request rewrites the existing nRF52840 driver. The current driver does not support sending an ACK upon receiving a packet. Additionally, the current driver power cycles the radio on/off after receiving or transmitting a packet. 

The rewrite of the driver removes the power cycling and allows the radio to switch between receiving and sending packets without powering off. Additionally, the nRF52840 radio chip supports hardware shortcuts that allow for interrupt events to be handled in hardware and automatically trigger tasks. The current receiving and sending now utilizes these shortcuts to improve the speed at which the driver operates. Finally, the driver adds support for sending an ACK when a received packet requests to be acknowledged.

This driver still lacks support to "listen" for an ACK on a packet the radio has sent. Likewise, each sent packet defaults to a false value for the acknowledged field.

### Testing Strategy

The libtock-c ieee802154/radio_tx and ieee802154/radio_rx examples were primarily used for testing while also observing behavior on wireshark. This branch was also merged into the ongoing Thread network development to test a more real world application that switches between sending and receiving. With this driver, Tock was able to join an openthread network consistently and correctly ACK openthread packets before the ACK timeout.

### TODO or Help Wanted

As mentioned in the overview, this driver still lacks support for listening for ACK on sending. Furthermore, I am open to comments on some design decisions primarily surrounding buffers/hardware shortcuts. 

The previous driver utilized the unwrap() function when taking the rx/tx buffers. I preserved this in the current driver, but wanted to draw attention to this as it may not be a good practice. Overall, this seems logical, as this panic will only occur if a faulty implementation of the driver or client receive function forgets to replace the buffer. In the current implementation, it is not possible for this panic to occur, and likewise only serves as a check against future faulty implementations.

Additionally, to enforce that the assumed shortcuts are enabled, I added condition checks in the interrupt handler to panic if the assumed shortcuts are not set as expected. The logic for this is similar to the previous comment as Tock discourages panic statements. Without these shortcuts set, the driver will exhibit undefined behavior.

I am curious what others think about these design decisions and if there is a better practice to achieve the same end.

### Documentation Updated

- [ x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [x ] Ran `make prepush`.


capsules/extra/src/rf233.rs
```diff
@@ -1346,6 +1346,8 @@ impl<'a, S: spi::SpiMasterDevice<'a>> radio::RadioData<'a> for RF233<'a, S> {
         self.rx_buf.replace(buffer);
     }
 
+    fn set_ack_buffer(&self, _buffer: &'static mut [u8]) {}
+
```
ppannuto, line 1350:
```suggestion
```

Ideally, fixing nrf52840-specific stuff wouldn't need to touch other radio implementation

ppannuto (changes requested):
Doing one series on ACK buffer to keep thoughts clustered. I think ideally the software ack machinery stays contained to the nrf family. It's a bit awkward with all the component machinery, I know, but I think the easiest might be to pull the `ieee802154_radio` out of the generic nrf `base_peripherals` and create a new entry in the `nrf52_components/` to create the chip-specific radio object with an ack buffer for the nrf radio driver

chips/nrf52/src/ieee802154_radio.rs
```diff
@@ -767,128 +797,237 @@ impl<'a> Radio<'a> {
...
+                    //in the IEEE 802.15.4 standard (see Figure 69 in
+                    //section 7.5.1.4 The CSMA-CA algorithm of the
+                    //standard).
+                    if self.cca_count.get() < IEEE802154_MAX_POLLING_ATTEMPTS {
+                        self.cca_count.set(self.cca_count.get() + 1);
+                        self.cca_be.set(self.cca_be.get() + 1);
+                        let backoff_periods = self.random_nonce() & ((1 << self.cca_be.get()) - 1);
+                        let current_time = self.timer0.unwrap_or_panic().get_current_time();
+                        self.timer0
+                            .unwrap_or_panic() // Unwrap fail = Missing timer reference for CSMA
```
ppannuto, line 931: This was already here, so not critical to fix at this stage, but this really shouldn't be a runtime check / possible failure. This should be passed in with component creation, as it's not possible to implement this radio (correctly) without a timer. If you're already mucking with components, would be nice to eliminate this...

chips/nrf52/src/ieee802154_radio.rs
```diff
@@ -767,128 +797,237 @@ impl<'a> Radio<'a> {
...
+                                ),
+                            );
+                    } else {
+                        // We have exceeded the IEEE802154_MAX_POLLING_ATTEMPTS and should
+                        // fail the transmission/return buffer to sending client
+
+                        let result = Err(ErrorCode::BUSY);
+                        //TODO: Acked is hardcoded to always return false; add support to receive tx ACK
+                        self.tx_client.map(|client| {
+                            // Unwrap fail = TX Buffer produced error when sending it back to the requestor after the channel was busy
```
ppannuto, line 945: Can this old comment be updated / clarified as well?

chips/nrf52/src/ieee802154_radio.rs
```diff
@@ -1035,11 +1170,18 @@ impl<'a> kernel::hil::radio::RadioConfig<'a> for Radio<'a> {
...
     fn is_on(&self) -> bool {
-        true
+        if self.registers.power.is_set(Task::ENABLE) {
+            true
+        } else {
+            false
+        }
     }
+
+    // Previous driver implementation //
```
ppannuto, line 1184: can this fn just be deleted?

chips/nrf52/src/ieee802154_radio.rs
```diff
@@ -718,20 +753,15 @@ impl<'a> Radio<'a> {
...
+        // Unwrap fail = Radio RX Buffer produced an invalid result when setting the DMA pointer.
+        let rbuf = self.rx_buf.take().unwrap();
+        self.rx_buf.replace(self.set_dma_ptr(rbuf));
 
-            self.tx_buf.replace(self.set_dma_ptr(tbuf));
-        } else {
-            let rbuf = self.rx_buf.take().unwrap(); // Unwrap fail = Radio RX Buffer produced an invalid result when setting the DMA pointer.
-            self.rx_buf.replace(self.set_dma_ptr(rbuf));
-        }
+        self.registers.shorts.write(Shortcut::READY_START::SET);
```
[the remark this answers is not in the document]
tyler-potyondy: You are correct in your assumption; the shortcuts are stateful. The reason for turning on/off the shortcuts when switching from receiving/transmitting/acknowledging is that they each require a different set of shortcuts that are not overlapping. I agree that a one time init would simplify the implementation. However, this would mean that we remove shortcuts and handle each progression of the state machine in the interrupt handler. This is entirely feasible, but will likely lead to a higher latency for each of our events. Does it make sense to have a higher latency for radio operations to simplify the state machine/readability of what the driver is doing?

chips/nrf52/src/ieee802154_radio.rs
tyler-potyondy: No, this fn is part of the RadioConfig trait (see below)

`pub trait RadioConfig<'a> {
    /// buf must be at least MAX_BUF_SIZE in length, and
    /// reg_read and reg_write must be 2 bytes.
    fn initialize(
        &self,
        spi_buf: &'static mut [u8],
        reg_write: &'static mut [u8],
        reg_read: &'static mut [u8],
    ) -> Result<(), ErrorCode>;
    fn reset(&self) -> Result<(), ErrorCode>;
    fn start(&self) -> Result<(), ErrorCode>;
    fn stop(&self) -> Result<(), ErrorCode>;
    fn is_on(&self) -> bool;
    fn busy(&self) -> bool;

    fn set_power_client(&self, client: &'a dyn PowerClient);

    /// Commit the config calls to hardware, changing the address,
    /// PAN ID, TX power, and channel to the specified values, issues
    /// a callback to the config client when done.
    fn config_commit(&self);
    fn set_config_client(&self, client: &'a dyn ConfigClient);

    fn get_address(&self) -> u16; //....... The local 16-bit address
    fn get_address_long(&self) -> [u8; 8]; // 64-bit address
    fn get_pan(&self) -> u16; //........... The 16-bit PAN ID
    fn get_tx_power(&self) -> i8; //....... The transmit power, in dBm
    fn get_channel(&self) -> u8; // ....... The 802.15.4 channel

    fn set_address(&self, addr: u16);
    fn set_address_long(&self, addr: [u8; 8]);
    fn set_pan(&self, id: u16);
    fn set_tx_power(&self, power: i8) -> Result<(), ErrorCode>;
    fn set_channel(&self, chan: u8) -> Result<(), ErrorCode>;
}`

chips/nrf52/src/ieee802154_radio.rs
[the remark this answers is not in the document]
ppannuto: > that they each require a different set of shortcuts that are not overlapping.

🤯 😭 — okay, I didn't quite appreciate that the active shortcuts were changing on different paths. If that's the case, then as-is is likely best.

chips/nrf52/src/ieee802154_radio.rs
ppannuto: Yeah, it looks like that's what the rf233 does.

The radio TRD says:
> The `busy` function returns whether the radio is currently busy.
It MUST return false if the radio is currently idle and can accept
reconfiguration or packet transmission requests. If it is busy and
cannot accept reconfiguration or packet transmission requests, it
MUST return true.

Would be good to make this an honest implementation eventually, but that can wait until a future PR.

kernel/src/hil/radio.rs
```diff
@@ -111,6 +111,7 @@ pub trait RadioData<'a> {
     fn set_transmit_client(&self, client: &'a dyn TxClient);
     fn set_receive_client(&self, client: &'a dyn RxClient, receive_buffer: &'static mut [u8]);
     fn set_receive_buffer(&self, receive_buffer: &'static mut [u8]);
+    fn set_ack_buffer(&self, ack_buffer: &'static mut [u8]);
```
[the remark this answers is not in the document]
tyler-potyondy: The most recent commit makes an attempt to resolve these changes. Since the nrf52_components is altered, it will require updating additional function signatures across other board families using the nrf52_components/peripherals (this is the reason for the PR checks failing). Once I get feedback on these changes, I will update the other board families.

chips/nrf5x/src/timer.rs
```diff
@@ -293,6 +293,10 @@ impl<'a> TimerAlarm<'a> {
         });
     }
 
+    pub fn get_current_time(&self) -> hil::time::Ticks32 {
```
alevy, line 296: Why is this useful if it's just a wrapper for `self.now()`? You can always get access to the `now` method with just `use kernel::hil::time::Time`

chips/nrf52/src/ieee802154_radio.rs
```diff
@@ -4,10 +4,11 @@
 
 //! IEEE 802.15.4 radio driver for nRF52
 
+use crate::timer::TimerAlarm;
 use core::cell::Cell;
 use core::convert::TryFrom;
 use kernel;
-use kernel::hil::radio::{self, PowerClient};
+use kernel::hil::radio::{self, PowerClient, RadioData};
 use kernel::hil::time::{Alarm, AlarmClient};
```
alevy, line 12: E.g., import `Time` here to be able to use `.now()` instead of the new `get_current_time()`

```suggestion
use kernel::hil::time::{Alarm, AlarmClient, Time};
```

alevy (changes requested):

alevy:
@tyler-potyondy this is an exceptionally well written PR overview. Thank you!

alevy:
> The previous driver utilized the unwrap() function when taking the rx/tx buffers. I preserved this in the current driver, but wanted to draw attention to this as it may not be a good practice. Overall, this seems logical, as this panic will only occur if a faulty implementation of the driver or client receive function forgets to replace the buffer. In the current implementation, it is not possible for this panic to occur, and likewise only serves as a check against future faulty implementations.

I think this is not ideal but fine. The obvious alternative is to `map(...)` the Option, instead, which is also unsatisfying. In general, this is a pattern that is worth (continuously) revisiting and seeing if we can come up with better practices or solutions. But for the purpose of this re-implementation, I think you made the right choice to preserve the existing "behavior"

chips/nrf52/src/ieee802154_radio.rs
```diff
@@ -767,128 +797,250 @@ impl<'a> Radio<'a> {
...
-
-        //   IF we receive the go ahead (channel is clear)
-        // THEN start the transmit part of the radio
-        if self.registers.event_ccaidle.is_set(Event::READY) {
-            self.registers.event_ccaidle.write(Event::READY::CLEAR);
-            self.registers.task_txen.write(Task::ENABLE::SET)
-        }
+        match self.state.get() {
+            // It should not be possible to receive an interrupt while the tracked radio state is OFF
+            RadioState::OFF => panic!("Received interrupt while off"),
```
alevy, line 820: This is a place where I think just ignoring the interrupt, rather than panicking, makes more sense, since panics _do_ incur some overhead (code size, particularly for the format string).

chips/nrf52/src/ieee802154_radio.rs
```diff
@@ -767,128 +797,250 @@ impl<'a> Radio<'a> {
...
+            RadioState::RX => {
+                // Enforce READY_START shortcut enablement if radio is receiving
+                if !self.registers.shorts.is_set(Shortcut::READY_START) {
+                    panic!("READY_START shortcut not enabled while sending.")
```
alevy, line 824: This seems like a waste. It's an invariant that's enforced everywhere, and if there is a bug in the logic, we probably don't want it bringing down the whole system. Seems better to just _test_ for it, rather than continuously test at runtime.

chips/nrf52/src/ieee802154_radio.rs
```diff
@@ -767,128 +797,250 @@ impl<'a> Radio<'a> {
...
+                }
             }
+            RadioState::TX => {
+                // Confirm proper hardware shortcuts are enabled
+                if !self.registers.shorts.matches_all(
+                    Shortcut::DISABLED_RXEN::SET
+                        + Shortcut::CCAIDLE_TXEN::SET
+                        + RXREADY_CCASTART::SET,
+                ) {
+                    panic!("Incorrect shortcut configuration for TX")
```
alevy, line 912: ditto

ppannuto:
I took a look at the ack buffer / component situation. Really, this should just be something passed in at creation time, so I switched the buffer to go that way. In so doing, I realized that we (eventually) need a nrf528XX crate as well, since the nrf58232 does **NOT** have 15.4 support, but the 58233 and 58240 both do.

In the short term, I pulled the 15.4 driver into the nrf52840 crate, which caused a bunch of boilerplate changes in all the boards that are built on the 52840. The makes 15.4 inaccessible to the microbit briefly, but it wasn't using it anyway. Adding a proper 528XX crate can come after this PR.

alevy:
@ppannuto wouldn't it be reasonable to just put stuff in `nrf5x` as long as the register files/functionality is always consistent and individual chips can simply cherry pick what they use? Like... it's ok for it to be in `nrf5x` even if not all drivers there are using by all `nrf5*` chips. Or does that not work?

ppannuto:
Perhaps. It was in the nrf52, so could have moved to nrf5x instead of nrf52. The reason I went the way I did is that there is no nrf5xDefaultPeripherals object, but there is a nrf52840DefaultPeripherals, which I moved the 15.4 radio instantiation to. I don't think there are any chips outside of nrf528XX that support 15.4, so scoping it down to make such accidental inclusion impossible for this or other more chip-specific peripherals held appeal.

ppannuto:
Someday, rust will sort out delegates and composition of chip capabilities will be a lot smoother I think, but a quick peruse of RFCs make it look that while highly desired, it's also pretty far off.

bradjc:
> @ppannuto wouldn't it be reasonable to just put stuff in `nrf5x` as long as the register files/functionality is always consistent and individual chips can simply cherry pick what they use? Like... it's ok for it to be in `nrf5x` even if not all drivers there are using by all `nrf5*` chips. Or does that not work?

I do think we need to write down the invariant we are trying to preserve. For me it is something like: 

> Chip crates will never expose an instantiated resource or setting that is incompatible with a MCU matching the chip name.

So including the source for a driver is fine, as long as that driver isn't created. Including functions that return MCU constants or settings are OK, as long as they cannot be wrong. For example, if I had a function in a hypothetical `nrf52xxx` crate:

```rust
fn get_number_spi_busses -> usize {}
```

that's fine, as long as it ALWAYS returns the correct value for every nrf52xxx chip that exists (and will exist). That is, it cannot fallback to some default which could be wrong in the future should nordic release the nRF52843 chip (or whatever). It's ok if it fails to compile.

tyler-potyondy:
> > The rewrite of the driver removes the power cycling and allows the radio to switch between receiving and sending packets without powering off
> 
> I don't think I understand whether this means the radio is never turned off (which presumably is a problem for power consumption) or if it's only not turned off when we know we immediately want to switch to receiving/sending?

chips/nrf52/src/ieee802154_radio.rs
tyler-potyondy: I agree that this makes more sense. I will go ahead and remove this. Do we have an established methodology or way to test this in future PRs or changes to this driver?

chips/nrf52/src/ieee802154_radio.rs
tyler-potyondy: If this condition ever occurs, we have diverged from the state machine and some bug exists in the driver. From a "handling the interrupt perspective" I agree that it makes sense to just ignore here. 

Is there some other way to still prevent a "silent fail" if our state diverges from the state machine?

alevy:
@tyler-potyondy looking at the code in the PR again, I understand now. Can you confirm?

The behavior that this PR changes is _just_ that transitioning between RX & TX doesn't _first_ turn the radio off and then re-initialize it, which is _never_ necessary.

However, it will _still_ be the case that `stop()` powers off the radio and so, for example, the XMAC capsule still will successfully have the radio completely off most of the time.

With regard to when and how to power the radio off for Thread, it ultimately depends on what the interface to applications is, and what the Thread MAC is (which I don't recall). In general, it should probably not be the case that each application has control over the radio power and, instead, power consumption is simply optimized to comply with Thread's expectation for the radio being on (presumably a  sleep edge device has some schedule for when it listens for incoming frames or something). This should _probably_ be managed in either a Thread capsule _or_ in a specially privileged Thread management process (which should be the only one with access to, e.g., the power state of the radio).

chips/nrf52/src/ieee802154_radio.rs
```diff
@@ -767,128 +797,237 @@ impl<'a> Radio<'a> {
...
+                // Since READY_START shortcut enabled, always clear READY event
+                self.registers.event_ready.write(Event::READY::CLEAR);
+
+                // Completed receiving a packet, now determine if we need to send ACK
+                if self.registers.event_end.is_set(Event::READY) {
+                    self.registers.event_end.write(Event::READY::CLEAR);
+                    let result = self.crc_check();
+
+                    // Unwrap fail = RX Buffer produced error when sending received packet to requestor
+                    let rbuf = self.rx_buf.take().unwrap();
```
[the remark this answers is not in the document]
tyler-potyondy: Looking further into this, I do not think this has to do with the DMA event. The only instance this would occur is if the client receive(...) method did not call `self.mac.set_receive_buffer(buf)` to return the buffer to the radio driver. I may be overlooking something related to the DMA event, but this is my understanding.

The set_dma_ptr method always returns the buffer, so the only case the buffer "wouldn't be returned from DMA" is if someone forgets to replace the buffer after the set_dma_ptr method is called. Please let me know if the updated comments are more clear.

chips/nrf52/src/ieee802154_radio.rs
alevy: Not that I can think of. The trouble is that `panic` is both catastrophic and has high size overhead. If ignoring completely is unpalatable (particularly for a still early implementation), perhaps replacing this with a `debug!` might be better. This won't fail "silently" but is also might be easier (though not currently supported) to "turn off" in production.

alevy (dismissed):
I think this is pretty close and is worth merging as soon as another 802.15.4 expert (@ppannuto) is satisfied.

chips/nrf52/src/ieee802154_radio.rs
bradjc: For what it is worth, the only place I know of where we handle something like this is in Process within the kernel, where we have a error type which is basically "kernel_bug" that is returned in all of these cases that should never happen but the compiler cannot verify that.

But, these type of state machine errors don't seem to happen in practice. Either there are bugs and even simple testing identifies them, or they are correct and panic is never called.

tyler-potyondy: Thank you for the input. I'll go ahead and add a debug! statement instead of the panic.

ppannuto (dismissed):
🎉

brghena:
I'll take a look at this PR right now.

ppannuto:
@tyler-potyondy I fixed the nano33 merge conflict; it was benign.

ppannuto (approved):

chips/nrf52840/src/ieee802154_radio.rs
```diff
@@ -1146,23 +1347,47 @@ impl<'a> kernel::hil::radio::RadioData<'a> for Radio<'a> {
...
+            // check insures we do not interrupt an ungoing transmission
             return Err((ErrorCode::BUSY, buf));
         } else if radio::PSDU_OFFSET + frame_len >= buf.len() {
             // Not enough room for CRC
             return Err((ErrorCode::SIZE, buf));
         }
 
+        if let RadioState::OFF = self.state.get() {
+            self.radio_initialize();
+        }
```
brghena, line 1361: Do we also need to `radio_initialize()` in the `start()` or `reset()` functions? If not, why do we need to do so here?

chips/nrf52840/src/ieee802154_radio.rs
```diff
@@ -767,128 +835,254 @@ impl<'a> Radio<'a> {
...
+                    // or some instance in  driver taking buffer without properly replacing).
+                    let rbuf = self.rx_buf.take().unwrap();
+
+                    let frame_len = rbuf[MIMIC_PSDU_OFFSET as usize] as usize - radio::MFR_SIZE;
+                    // Length is: S0 (0 Byte) + Length (1 Byte) + S1 (0 Bytes) + Payload
+                    // And because the length field is directly read from the packet
+                    // We need to add 2 to length to get the total length
+
+                    // 6th bit in 2nd byte of received 15.4 packet determines if sender requested ACK
+                    if rbuf[2] & ACK_FLAG != 0 {
```
brghena, line 884: Shouldn't we only be sending an ACK if the packet requests one AND the CRC is valid?

chips/nrf52840/src/ieee802154_radio.rs
```diff
@@ -767,128 +835,254 @@ impl<'a> Radio<'a> {
...
+                // Completed receiving a packet, now determine if we need to send ACK
+                if self.registers.event_end.is_set(Event::READY) {
+                    self.registers.event_end.write(Event::READY::CLEAR);
+                    let result = self.crc_check();
+
+                    // Unwrap fail = Radio RX Buffer is missing (may be due to receive client not replacing in receive(...) method,
+                    // or some instance in  driver taking buffer without properly replacing).
+                    let rbuf = self.rx_buf.take().unwrap();
+
+                    let frame_len = rbuf[MIMIC_PSDU_OFFSET as usize] as usize - radio::MFR_SIZE;
```
brghena, line 878: Should there be any sanity checking on this `frame_len`, or is that up to the user? For example, in the case that the CRC is bad, this could be incorrect or invalid (>127).

chips/nrf52840/src/ieee802154_radio.rs
```diff
@@ -669,41 +728,52 @@ pub struct Radio<'a> {
...
-    timer0: OptionalCell<&'a crate::timer::TimerAlarm<'a>>,
+    timer0: OptionalCell<&'a TimerAlarm<'a>>,
+    state: Cell<RadioState>,
 }
 
 impl<'a> AlarmClient for Radio<'a> {
     fn alarm(&self) {
-        self.rx();
+        // This alarm function is the callback for when the CCA backoff alarm completes
+        // Attempt a new CCA period by issuesing CCASTART task
```
brghena, line 746:
```suggestion
        // Attempt a new CCA period by issuing CCASTART task
```

chips/nrf52840/src/ieee802154_radio.rs
```diff
@@ -767,128 +835,254 @@ impl<'a> Radio<'a> {
...
+                        client.send_done(tbuf, false, result);
+                    });
+                    rx_init = true;
+                }
             }
+            RadioState::ACK => {
+                // Enforce READY_START shortcut enablement if radio is sending ACK
+                if !self.registers.shorts.is_set(Shortcut::READY_START) {
+                    panic!("READY_START shortcut not enabled while undergoing ACK.")
+                }
```
brghena, line 1024: Why enforce the shortcut here in the ACK state, but assume the shortcuts in the TX or RX states?

brghena (commented):
This is excellent. I really appreciate the clear comments that explain why/how for sections of code.

A few questions for you, which might lead to changes or maybe you've already thought through them (in which case a comment in the code could be helpful). Also a few small spelling fixes.

chips/nrf52840/src/ieee802154_radio.rs
tyler-potyondy: All three originally were enforced. A comment from Amit prompted the switch to assuming rather than enforcing. I forgot to remove this one. Thanks for catching this.

chips/nrf52840/src/ieee802154_radio.rs
tyler-potyondy: I added `radio_initialize()` to the `start()` and `reset()` methods

chips/nrf52840/src/ieee802154_radio.rs
tyler-potyondy: I am unsure what is best here. On one hand, it makes sense to have the sanity check since the size is enforced by the 15.4 spec. The status of the CRC check is passed to the user/client so theoretically higher levels can handle this. I am curious as to your/others thoughts on if this should be checked here or handled by the client

alevy: Recall that this is a hardware specific implementation of a general interface, not an abstraction layer, so it is not in the business of helping clients avoid shooting themselves in the foot (that may or may not be a job for middle layers).

This should reliably imementation the HIL, without unnecessarily panicking or otherwise breaking the rest of the system.

I'm making some assumptions here without fully being switched into to the specifics of the interface, but I think that means that if a client passed in an invalid frame (e.g. it's too long), that's the client's business.

chips/nrf52840/src/ieee802154_radio.rs
```diff
@@ -767,128 +835,255 @@ impl<'a> Radio<'a> {
...
-        self.registers.intenset.write(
-            Interrupt::READY::SET
-                + Interrupt::CCAIDLE::SET
-                + Interrupt::CCABUSY::SET
-                + Interrupt::END::SET
-                + Interrupt::FRAMESTART::SET,
-        );
+        self.registers
+            .intenset
+            .write(Interrupt::READY::SET + Interrupt::CCABUSY::SET + Interrupt::END::SET);
```
hudson-ayers, line 1086: What is the reasoning behind this change? Seems odd that this just enables a different set of interrupts now

chips/nrf52840/src/ieee802154_radio.rs
hudson-ayers: Is it possible for transmit() to work correctly if start() or reset() has not been called? If not, why do we still need to initialize here?

chips/nrf52840/src/ieee802154_radio.rs
brghena: That all makes sense to me. The client is aware of whether the CRC is valid or not, and can make decisions based on that information (like not trusting the length field).

chips/nrf52840/src/ieee802154_radio.rs
brghena: I assumed that you could call `stop()` to turn off the radio, and then `start()` to start receiving again. Or `stop()` and then later `transmit()` to go straight to transmitting without hanging out in receive mode for a bit.

(I'm actually not sure why `reset()` exists at all, but it's part of the HIL, so maybe it makes more sense on other chips)

tyler-potyondy: Do others have thoughts on if transmit should fail if the radio is turned off? A case can be made for both implementations, and I am unsure which makes more sense.

chips/nrf52840/src/ieee802154_radio.rs
tyler-potyondy: The previous implementation enabled: 
`Interrupt::READY::SET`
`Interrupt::CCAIDLE::SET`
`Interrupt::CCABUSY::SET`
`Interrupt::END::SET`
`Interrupt::FRAMESTART::SET`

The framestart interrupt notifies the kernel of the progressing state of the rx/tx, but does not require any response or change of state to the radio so there does not exist a need to trigger an interrupt in the kernel for this radio event.

To reduce tx latency, I enabled the nrf52840 radio's hardware shortcuts feature so that the tx progression can automatically launch the next task on triggered radio interrupts. These shortcuts made it redundant to enable these interrupts as the internal radio events automatically trigger the next task and we no longer need to handle those events in the kernel.

chips/nrf52840/src/ieee802154_radio.rs
brghena: I think `transmit()` should turn on the radio AND successfully transmit. It shows intent from the client to use the radio, so just do the right thing. 

What Hudson was asking above was _whether_ this is the case or not. I.e., does `transmit()` work in the current radio state is OFF?

brghena (approved):

chips/nrf52840/src/ieee802154_radio.rs
alevy: > I think transmit() should turn on the radio AND successfully transmit.

@brghena I think this is misreading at least the current HIL (which unfortunately is underdocumented so we have to look at how it's used instead).

There are separate interfaces for managing the power state of the radio and for transmitting frames and `start` (power on) is asynchronous (the client waits for a state change notification in the `PowerClient` trait)---most likely because at least one implementation (the rf233) is over i2c, so any operation needs to be async. The _current_ MAC implementation (xmac) explicitly uses this---when it wants to transmit, if first calls `start` and then actually transmits from the power-on callback.

I think it's _fine_ for this implementation to do one better and turn the radio on for transmit, but probably _better_ to avoid it and push that functionality to the client---this would avoid buggy clients that only work on some hardware that happens to turn the radio on in `transmit`.

brghena: Thanks for the higher-level thinking Amit! Looking closer, you're definitely right here about the intent of the HIL as designed.

That being said, the Radio HIL is only loosely implemented by this chip driver as it stands (in part because of how undocumented it is, I would guess). The `PowerClient` for example, isn't implemented and `start()` doesn't have a way to provide a callback. There is a question as to whether calling start and then transmit quickly would mess anything up, but I suspect not for this hardware implementation.

So what would be the best course of action for this PR? I don't think meeting every aspect of the HIL should be the immediate requirement since this PR is really just implementing ACKs.

alevy: @brghena got it yes. `start` is still doing a fair amount of work and it's still that case that, but your point is well taken.

I'd suggest, in this PR, _not_ adding the above code that checks if the radio state is off and, if so, initializes the radio. Instead, subsequent changes can maybe either bring this chip implementation closer to something reasonable with the HIL (probably simply through a deferred call) or else change the HIL.

tyler-potyondy: @brghena @alevy Is there consensus on removing this?

brghena: Consensus: leave this alone for now entirely. It's fine as-is.

Long-term: the 15.4 driver here should better implement the HIL which expects callbacks for power state changes. But that doesn't have to happen here. (And simultaneously we would be reviewing whether that HIL is appropriate or not)

PR #3617, opened 2023-08-16
tickv: remove key search function

bradjc:
### Pull Request Overview

This isn't ideal, but currently tickv does not correctly implement search and this fixes a bug when keys hash to region 0 in the store.

Currently, if there is a key stored in region 0 and a get() for a key hashes to region 0 tickv will enter an infinite loop checking the same flash region for the key.

I suppose it would be better to correctly implement the search, but it is difficult to do correctly.

### Testing Strategy

This pull request was tested calling get on the keys `cdf` and `cddf`.

### TODO or Help Wanted

@alistair23 Can we do better?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/libraries/tickv/src/tickv.rs
+++ b/libraries/tickv/src/tickv.rs
@@ -611,113 +611,111 @@ impl<'a, C: FlashController<S>, const S: usize> TicKV<'a, C, S> {
     pub fn get_key(&self, hash: u64, buf: &mut [u8]) -> Result<(SuccessCode, usize), ErrorCode> {
         let region = self.get_region(hash);
 
-        loop {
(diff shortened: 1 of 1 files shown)


alistair23:
Hmmm.... I'm not sure what's going on there.

The entire library seems pretty broken now though. The tests no longer compile

This seems to be the commit that breaks compilation

```
bb3063da1 tickv: use value length (by Brad Campbell 8 weeks ago)
```

But a few other commits add a bunch of warnings.

Reverting `bb3063da1` doesn't fix things up either.

I don't think this PR should be merged. Can you fix the tests then I can look at fixing this issue?

PR #3621, opened 2023-08-17
workflows/litex_sim: add mirror for the pre-compiled RISC-V toolchain

lschuermann:
### Pull Request Overview

This pull request fixes the LiteX sim CI workflow by adding another mirror to fetch the pre-compiled toolchain from. UVA seems to have some network issues right now, and luckily I still had a version of this toolchain in my local archives. I won't make any promises on the availability of my mirror, but we're still preferring UVA and performing checksumming on the downloaded file, so I think this should be relatively low risk.

### Testing Strategy

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/.github/workflows/litex_sim.yml
+++ b/.github/workflows/litex_sim.yml
@@ -52,10 +52,23 @@ jobs:
       # follows the libtock-c GitHub actions definition.
       - name: Setup RISC-V GCC toolchain
         run: |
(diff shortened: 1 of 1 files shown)


ppannuto (approved):

bradjc:
Apparently it has to be www.cs.virginia.edu. Not sure if this is intentional or not...

PR #3622, opened 2023-08-18
RFC: Managing versions and long-pending changes

ppannuto:
### Pull Request Overview

This tries to write up a more complete summary of my thinking from #3585. As I was working on this, it bled over to a couple distinct concerns that arguably could be 4-5 separate RFCs, but these are all interrelated to a degree. I tried to keep the document reasonably organized which hopefully can facilitate logical threads of conversation.

### Testing Strategy

N/A

### TODO or Help Wanted

Your thoughts! And talk me off the ledge of `cfg` maybe being a good idea for something. I think I talked myself out of it already, but there was a real moment there where I was almost going to advocate for it.

---

[Rendered](https://github.com/tock/tock/blob/managing-versions/doc/rfcs/2023-08-17--ManagingVersions.md)


doc/Maintenance.md
```diff
@@ -36,6 +37,63 @@ academic and professional conferences.
 The project also maintains a [book](https://book.tockos.org) which includes
 self-guided tutorials for various Tock features.
 
+## Between Releases
```
bradjc, line 40: This section header doesn't seem to match the contents.

doc/Maintenance.md
```diff
@@ -36,6 +37,63 @@ academic and professional conferences.
...
+
+Generally speaking, the primary development branch of Tock holds "next
+minor version" semantics. That is, a default compile of a fresh checkout
+of the Tock repository should maintain backwards-compatibility with the
+currently released major version, but may also include bugfixes and new,
+additional features.
+
+### Patch Releases
```
bradjc, line 48: Can this just move to the "Preparing a Release" section?

doc/rfcs/2023-08-17--ManagingVersions.md
```diff
@@ -0,0 +1,199 @@
...
+I think it worked pretty well last time, but open to more thoughts here.
+
+
+Understanding the delta between master and last release
+------------------------------------
+
+While we have a CHANGELOG file, to-date we are not great about keeping
+it up to date during active development. Rather, as part of release
+preparation the core team retroactively looks through all of the PRs and
+commit history since the last release and synthesizes the key changes.
```
bradjc, line 56: CHANGELOG takes work, and it's work no one wants to do (as evidenced by it not getting updated). Perhaps eventually we hire someone who can help. But I don't think any amount of documenting is going to get the current set of core members interested in maintaining a changelog.

This is fairly on-the-nose, forgive me, but rather than creating this PR, why not direct your "service" tock time towards maintaining the changelog? You created it (https://github.com/tock/tock/commit/a32a34de6ec9eebf21f47737f397e213413dbf3a) but I don't think have touched it since 2018, and I think it's reasonable to say you are the main person who remembers it is there.

doc/rfcs/2023-08-17--ManagingVersions.md
```diff
@@ -0,0 +1,199 @@
...
+        someone's head a few months ago.
+
+
+### What can we do instead?
+
+1. Keep PRs with fixes tagged and open until a major release process starts.
+   - Pro: Have the "real code" and "real fix" already there.
+   - Con: Likely that the PR will not merge cleanly by the time the next
+       major release occurs.
+   - Con: Lots of noise in the PR queue
```
bradjc, line 124: This is a substantial con.

doc/rfcs/2023-08-17--ManagingVersions.md
```diff
@@ -0,0 +1,199 @@
...
+           if we released a new version we'd shave off XX% of kernel
+           size for large enough X.
+   - Nuance: It may be worth distinguishing between "next minor" and
+       "next major". In particular, "next minor" `cfg`'d stuff can be
+       default-on, and largely no change from current practice, except
+       for this sentinel sprinkled in the codebase identifying what's
+       not in the released version [though, this could arguably also be
+       accomplished with some type of comment keyword].
+
+4. ...?
```
bradjc, line 160:
```suggestion
4. Open a tracking issue for the change needed for the next release. Include the diff required to implement the change in the issue text.
    - Pro: Work to implement the change already done.
    - Pro: Because by definition this is a change to some stabilized portion of the kernel, the patch is likely to cleanly apply or be very close when the next release is prepared.
    - Con: Someone has to remember to address the tracking issue.
```

doc/rfcs/2023-08-17--ManagingVersions.md
```diff
@@ -0,0 +1,199 @@
...
+**Proposal:** We need something between "stable" and "unstable". Maybe
+not the [7 tiers of Python stability](https://pypi.org/classifiers/),
+but something that helps folks get a sense of 'this has been around for
+a while' versus 'this is brand new and we are still figuring it out'.
+
+
+**Proposal:** We should not stabilize whole driver interfaces, but
+rather individual syscalls within them. Low-level-debug Command 2,
+"print a number", seems pretty safe to call immutable; I don't think
+we want to lock the LLD interface yet, however.
```
bradjc, line 185: This seems like the opposite of this section, as it would make it _harder_ to know if your app relies on stable functionality or not.

jrvanwhy:
Just came up with some more ideas.

## Idea 4: Procedural macro

Add a new procedural macro crate that exports attribute macros for each version (`tock2` and `tock3)`. The `tock2` macro would only build an item if `TOCK3` is unspecified and the `tock3` macro would only build an item if `TOCK3` is set. Example (not-production-quality) implementation:

```rust
extern crate proc_macro;
use proc_macro::TokenStream;

// Builds the specified item if and only if TOCK3 is empty.
#[proc_macro_attribute]
pub fn tock2(_attr: TokenStream, item: TokenStream) -> TokenStream {
    match std::env::var_os("TOCK3") {
        None => item,
        Some(_) => Default::default(),
    }
}

// Builds the specified item if and only if TOCK3 is specified.
#[proc_macro_attribute]
pub fn tock3(_attr: TokenStream, item: TokenStream) -> TokenStream {
    match std::env::var_os("TOCK3") {
        None => Default::default(),
        Some(_) => item,
    }
}
```

In each crate that has Tock 3.0 changes, we would add a dependency on this procedural macro and use it as follows:

```rust
#[tock2]
fn foo_v2() {}

#[tock3]
fn foo_v3() {}
```

This has the advantage of working for users who do not use Cargo. Unlike the `cfg`-based options, the above code will stop compiling when the `tock2` macro is removed, giving us a way to ensure that we don't accidentally use Tock 2.0 code after Tock 3.0 is released. One downside is we have to depend on this new crate from any Tock crate that has Tock 3.0 code, but I think that's a reasonable compromise.

Big downside: The build system will be oblivious to the fact our crates depend on the value of `TOCK3`, as procedural macros [do not yet have a way to specify that they depend on an environment variable](https://github.com/rust-lang/rust/issues/99515).

## Idea 5: idea 4 + idea 3

We could have `build.rs` in the procedural macro crate read `TOCK3` and set a `cfg` variable that is used by the macro implementations to control their behavior. That way, Cargo will know to rebuild the procedural macro crate if the value of `TOCK3` changes...

Procedural macro `build.rs`:
```rust
fn main() {
    println!("cargo:rerun-if-env-changed=TOCK3");
    if std::env::var_os("TOCK3").is_some() {
        println!("cargo:rustc-cfg=tock3");
    }
}
```

Procedural macro `lib.rs`:
```rust
extern crate proc_macro;
use proc_macro::TokenStream;

#[proc_macro_attribute]
pub fn tock2(_attr: TokenStream, _item: TokenStream) -> TokenStream {
    #[cfg(not(tock3))]
    return _item;
    #[cfg(tock3)]
    return Default::default();
}

#[proc_macro_attribute]
pub fn tock3(_attr: TokenStream, _item: TokenStream) -> TokenStream {
    #[cfg(not(tock3))]
    return Default::default();
    #[cfg(tock3)]
    return _item;
}
```

Usage:
```rust
#[tock2]
fn foo_v2() {}

#[tock3]
fn foo_v3() {}
```

This has the advantage of just Doing The Right Thing (building Tock 2.0 code) if someone builds the crate without invoking `build.rs`. We also only need to remember to clean up old `#[cfg()]` attributes in one place (the proc macro crate); if we leave any `#[tock2]` annotations around we'll get a compile error.

jrvanwhy:
I realize I should summarize my previous comments, especially because this may be discussed at the next core WG call, which I do not expect to attend.

I thought through how we could implement "The `cfg` option" in Tock. There are some issues with using feature flags or custom `rustc` `--cfg=` arguments, but there is a viable option for implementing conditional compilation. The viable option is to introduce a (relatively simple) procedural macro crate that looks for the presence of an environment variable to determine whether to compile the current-release code or the next-release code. This option plays reasonably well with non-Cargo build systems and gives us a reliable way to clean up the old code once we've created a new major Tock release.

Truthfully, I think the conditional-compilation option is lower maintenance than the "keep PRs open until the next major release" and "always keep a next-major-release branch open" options. If we don't come up with any better ideas, then I would advocate for it.

bradjc:
## Idea 6: Diff in tracking issue

Open a tracking issue for the change needed for the next release. Include the diff required to implement the change in the issue text.

Pro:
- Work to implement the change already done.
- Because by definition this is a change to some stabilized portion of the kernel, the patch is likely to cleanly apply or be very close when the next release is prepared.

Con:
- Someone has to remember to address the tracking issue.

bradjc:
My pushback to all of the automatic/cfg/etc. options is that we would effectively be maintaining two versions of the kernel (and giving users a [relatively] easy option to use either). I argue we have trouble keeping up with _one_ version of the kernel, and don't want the inevitable issue: "I enabled the tock-3.0 flag and userspace doesn't work how do I fix it".

doc/Maintenance.md
```diff
@@ -36,6 +37,63 @@ academic and professional conferences.
...
+its first release candidate, a final minor release for the current major
+version will be prepared.
+
+After this release, all development will move to the major release
+branch. Critical fixes that merit a new point release for the old major
+version will be accepted, but all other changes should target the new
+major release. The core team will then initiate the release process for
+the new major release. When the final release is tagged, the major
+version branch will be merged into the default branch for the Tock
+repository, and development will continue as-normal.
```
lschuermann, line 94: This seems potentially problematic, as the merge could potentially change the code-base due to minor changes made to the main branch, from after the point where the next major version branch has been branched off.

By merging only _after_ the release, we'd immediately have a potentially quite different kernel right after tagging a release, which seems counter-intuitive. Additionally, minor fixes going into the main branch, while ideally ported to the release branch, seem important to include in any future release. By merging only after tagging a release, I fear it would be easy for these fixes to be missed and not included in a release.

An easier, more predictable approach, would be to perform the "swap" (merge of `tock-v{N+1}` into `master`, and branch off a "maintenance branch" for the previous major version from before that merge commit. Then, when things are settled, testing and the release process for `tock-v{N+1}` would commence on the main branch.

This has the disadvantage that there is a short period of time where the main branch is ahead of and breaking w.r.t. the current stable major release. But it's the process we've used for Tock 2.0 and worked quite well for that.

@ppannuto thoughts?

lschuermann (commented):
@ppannuto I want to revive this PR, and with that extending the `Maintenance.md` doc with more concrete guidance on what steps to do for a release (e.g., which version numbers to bump, how to test, ...).

The changes to both `trd104-syscalls.md` and `Maintenance.md` seem great, modulo the above comment. I'm less sure about the value in the RFC, which seemed to have kicked off a significant and unresolved discussion on this PR. My goal is to separate the documentation about the current processes from any discussion on how to optimize this in the future.

doc/Maintenance.md
ppannuto: I agree, swap semantics make sense. I _think_ I had been trying to write down what we ended up doing for Tock 2.0, since it did work reasonably well.

ppannuto:
@lschuermann I think the RFC doc does have some real questions that would be nice to resolve (in particular, how to ergonomically handle "long-deferred" code that awaits a major release). Most of the discussion in this PR thread is tied to the RFC questions. I think the best path forward would be to extract the TRD and Maintenance updates to a separate PR, and this one can focus on the fuzzier RFC questions.

PR #3624, opened 2023-08-18
rv32i: use la instruction, pass in linker vars

bradjc:
### Pull Request Overview

This pull request uses some shortcuts in the riscv assembly.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/lib.rs
+++ b/arch/rv32i/src/lib.rs
@@ -26,8 +26,9 @@ pub use riscv::csr;
 
 extern "C" {
     // Where the end of the stack region is (and hence where the stack should
-    // start).
+    // start), and the start of the stack region.
     static _estack: usize;
+    static _sstack: usize;
 
     // Boundaries of the .bss section.
     static mut _szero: usize;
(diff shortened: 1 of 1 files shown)


lschuermann (commented):
Passing in these addresses using `sym` seems like a strict improvement, but I'm not sure how I feel about `la`.

While I haven't looked at the assembly, and wasn't aware of this pseudo-instruction before, it seems to expand to an `auipc` and `addi` instruction. Effectively the only runtime difference would thus be that we switch from absolute to PC-relative addressing.

If that's indeed the case, I'd be indifferent on this -- however it is nice that absolute addressing encodes these addresses directly in the assembly, and a PC-relative jump often requires some more manual decoding work to "decipher" addresses in the assembly.

alistair23 (approved):

bradjc:
We use `la` throughout the file. I think the major benefit is it halves the instructions we have to maintain and reason about.

But is there a similar psuedo instruction for absolute addressing?

lschuermann:
> But is there a similar psuedo instruction for absolute addressing?

There is `li` (load immediate), which expands into the current `lui` and `addi`. However, the compiler / assembler disallows us using that here. I believe this is because an `li` can also potentially generate just an `addi` if the immediate is small enough, but that would then depend on function placement, which is only determined at link time.

ppannuto (commented):
I don't know enough risc-v detail to assess too critically, but generally this looks reasonable to me 👍🏼

bradjc:
Yay or nay? I think we should be consistent with the rest of the file and choose the option which is fewer lines of code for us to maintain.

lschuermann (approved):

bradjc:
Merge this?

PR #3626, opened 2023-08-18
Syscall Spec: Allow Command 0 to return Success with value or lock it down

ppannuto:
# [Rendered](https://github.com/tock/tock/blob/cmd0-counts/doc/rfcs/2023-08-18--CommandZeroSemantics.md)

---

### Pull Request Overview

This pull request capture the proposal(s) brought up by @alevy / @bradjc / others on the call today and in https://github.com/tock/tock/issues/3375#issuecomment-1681535604.

### Testing Strategy

N/A

### TODO or Help Wanted

Do we want to go this way?

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

---

# Motivation

The utility of an "exists" syscall is laid out well in the current TRD104 text:

> If userspace tries to invoke a system call that the kernel does not
support, the system call will return a Failure result with an error
code of `NODEVICE` or `NOSUPPORT` (Section 4). As the Allow and
Subscribe system call classes have defined Failure types, the kernel can
produce the expected type with known failure variants. Command, however,
can return any variant. This means that Commands can appear to have
two failure variants: the one expected (e.g., Failure with u32) as well
as Failure. To avoid this ambiguity for `NODEVICE`, userspace can use the
reserved "exists" command (Command Identifier 0), described in Section
4.3.1. If this command returns Success, the driver is installed and will
not return a Failure with `NODEVICE` for Commands. The driver may still
return `NOSUPPORT`, however. Because this implies a misunderstanding of
the system call API by userspace (it is invoking system calls that do not
exist), userspace is responsible for handling this case.

However, there is nothing in this rationale that would require _only_ allowing `Success` as-opposed to any of the success variants.

During the call, a few design constraints were discussed:

 - **Existence checks should be idempotent and side-effect free.**
    - This is not actually enforced by the current TRD text. However, the motivation is somewhat clear — a generic component polling for hardware existence / platform capabilities should not trigger hardware events.
    - The secondary implication here is that this implicitly forbids Command 0 from triggering asynchronous events. Currently, a driver _could_ generate an Upcall in response to a Command 0 event, which, given that 'anything' might check existence, probably isn't desirable.
    - We could codify these limitations without the more permissive change here though.

 - **Why not pass this information in a second Command 1 or equivalent for drivers?**
    - Something that cares about hardware count could just call Command 1 and skip the existence check as [the kernel will automatically return `NODEVICE` if it's missing already](https://github.com/tock/tock/blob/aa33bf1bad61ff8f3ba99a36a0760368bc4e6c3f/kernel/src/kernel.rs#L1038).
    - However, now drivers have to implement two syscalls (0: exists, 1: info-about-existence) where one could easily be collapsed into the other, reducing code overhead.

 - **Having every driver implement an empty Command 0 is silly, we might as well let them do something with it.**
    - Yes.., but really that's an artifact of the legacy drivers that do something with command 0. Once all of those are gone, the kernel can take over management of Command 0.

---

I captured the two main branches here as
 - Proposal A: Modify Command 0 to be more permissive
 - Proposal B: Lock down Command 0 and give the kernel clear ownership

I included some rough-sketch code of what the two variations might look like to help discussion. It obviously will not compile as the surface area for the complete change is huge.

### DIFF
--- a/doc/reference/trd104-syscalls.md
+++ b/doc/reference/trd104-syscalls.md
@@ -7,7 +7,7 @@ System Calls
 **Status:** Draft <br/>
 **Author:** Hudson Ayers, Guillaume Endignoux, Jon Flatley, Philip Levis, Amit Levy, Pat Pannuto, Leon Schuermann, Johnathan Van Why, dcz <br/>
 **Draft-Created:** August 31, 2020<br/>
-**Draft-Modified:** August 21, 2023<br/>
+**Draft-Modified:** August 23, 2023<br/>
 **Draft-Version:** 8<br/>
 **Draft-Discuss:** tock-dev@googlegroups.com</br>
 
@@ -497,15 +497,20 @@ handle userspace/kernel mismatches should be able to handle `Failure` in
 addition to the expected failure variant (if different than `Failure`).
 
 
-4.3.1 Command Idenfitier 0
+4.3.1 Command Identifier 0
 --------------------------
 
- - [Added August, 2023](../rfcs/2023-08-18--CommandZeroSemantics.md).
+ - [Design RFC](../rfcs/2023-08-18--CommandZeroSemantics.md).
 
-Command Identifier 0 is implemented by the core kernel and provides an
-existence check for drivers. If a driver is installed, the kernel will
-return `Success` for Command 0. If a driver is not installed, the kernel
-will return `Failure` with and error code of `NODEVICE`.
+Command Identifier 0 provides an existence check for drivers. If a
+driver is accessible to calling application, the kernel will return
+`Success` for Command 0. If a driver is not accessible, the kernel will
+return `Failure` with an error code of `NODEVICE`.
+
+_Note:_ The kernel may still elect to return the error condition even
+though driver and its associated hardware may be physically present on
+the current platform (for example, if the application does not have the
+required permissions to access this driver).
 
 Device drivers CANNOT modify the behavior of Command Identifier 0.
 
--- a/doc/reference/trd104-syscalls.md
+++ b/doc/reference/trd104-syscalls.md
@@ -972,3 +972,8 @@ Leon Schuermann <leon@is.currently.online>
 
 Johnathan Van Why <jrvanwhy@google.com>
 ```
+
+7 References and Additional Information
+=======================================
+
+- [Design RFC for Command 0 Semantics](../rfcs/2023-08-18--CommandZeroSemantics.md).
--- a/doc/reference/trd104-syscalls.md
+++ b/doc/reference/trd104-syscalls.md
@@ -807,7 +807,7 @@ Because C allows a single return value but Tock system calls can return multiple
 they do not easily map to idiomatic C. These low-level APIs are translated into standard C
 code by the userspace library. The general calling convention is that the complex return types
 are returned as structs. Since these structs are composite types larger than a single word, the
-ARM and RISCV calling conventions pass them on the stack.
+ARM and RISC-V calling conventions pass them on the stack.
 
 The system calls are implemented as inline assembly. This assembly moves arguments into the correct
 registers and invokes the system call, and on return copies the returned data into the return type
--- a/doc/reference/trd104-syscalls.md
+++ b/doc/reference/trd104-syscalls.md
@@ -7,7 +7,7 @@ System Calls
 **Status:** Draft <br/>
 **Author:** Hudson Ayers, Guillaume Endignoux, Jon Flatley, Philip Levis, Amit Levy, Pat Pannuto, Leon Schuermann, Johnathan Van Why, dcz <br/>
 **Draft-Created:** August 31, 2020<br/>
-**Draft-Modified:** August 21, 2022<br/>
+**Draft-Modified:** August 21, 2023<br/>
 **Draft-Version:** 8<br/>
 **Draft-Discuss:** tock-dev@googlegroups.com</br>
 
--- a/doc/reference/trd104-syscalls.md
+++ b/doc/reference/trd104-syscalls.md
@@ -500,20 +500,13 @@ addition to the expected failure variant (if different than `Failure`).
 4.3.1 Command Identifier 0
 --------------------------
 
- - [Design RFC](../rfcs/2023-08-18--CommandZeroSemantics.md).
-
-Command Identifier 0 provides an existence check for drivers. If a
-driver is accessible to calling application, the kernel will return
-`Success` for Command 0. If a driver is not accessible, the kernel will
-return `Failure` with an error code of `NODEVICE`.
-
-_Note:_ The kernel may still elect to return the error condition even
-though driver and its associated hardware may be physically present on
-the current platform (for example, if the application does not have the
-required permissions to access this driver).
-
-Device drivers CANNOT modify the behavior of Command Identifier 0.
-
+Command Identifier 0 provides an existence check for drivers. Command
+Identifier 0 MUST return either `Success` or `Failure` with `ENODEVICE`.
+`Success` indicates that the driver is present and the userspace process can
+issue system calls to it. If the driver is not accessible, Command Identifier 0
+returns `Failure` with an error code of `ENODEVICE`. A driver may be not
+accessible because the kernel does not have it, the process does not have the
+required permissions to use it, or other reasons.
 
 4.4 Read-Write Allow (Class ID: 3)
 ---------------------------------
--- a/doc/reference/trd104-syscalls.md
+++ b/doc/reference/trd104-syscalls.md
@@ -7,7 +7,7 @@ System Calls
 **Status:** Draft <br/>
 **Author:** Hudson Ayers, Guillaume Endignoux, Jon Flatley, Philip Levis, Amit Levy, Pat Pannuto, Leon Schuermann, Johnathan Van Why, dcz <br/>
 **Draft-Created:** August 31, 2020<br/>
-**Draft-Modified:** August 23, 2023<br/>
+**Draft-Modified:** September 8, 2023<br/>
 **Draft-Version:** 8<br/>
 **Draft-Discuss:** tock-dev@googlegroups.com</br>
 
--- a/doc/reference/trd104-syscalls.md
+++ b/doc/reference/trd104-syscalls.md
@@ -496,14 +496,32 @@ failure variant of `Failure`, with an associated error code of
 handle userspace/kernel mismatches should be able to handle `Failure` in
 addition to the expected failure variant (if different than `Failure`).
 
-4.3.1 Command Identifier 0
+4.3.1 Command Identifier 0 [PROPOSAL A]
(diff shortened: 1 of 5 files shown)


bradjc:
I support proposal B. It is easier to teach, can be centralized, is easier to implement in userspace, and promotes better design patterns† for drivers. It does not prevent drivers from using other command numbers to expose counts or capabilities. It does not require us to modify or revisit this provision of TRD104.

†I argue that dynamic information that drivers currently share with cmd0 has minimal practical benefit to userspace, and in many cases if there is a value to return a value in cmd0 then the driver should be redesigned. The LED driver is maybe the best counterexample to this point; I can see where userspace can adapt to the number of LEDs on the board. But for others, returning a value for cmd0 often means the driver exposes some raw hardware resource: ADC channels, GPIO pins, buttons. But when there is more than one resource (e.g. multiple ADC channels) the driver provides no context for what the resource actually is (e.g. what the ADC channel is attached to). When drivers try to represent multiple copies of some resource, and hence would want to expose a count via cmd0, I think that is a bad design we want to discourage. This all to say, not only should we _not_ change TRD104, we should actively _desire_ to have drivers follow TRD104, because it means a more robust and more usable driver.

brghena:
I was going to say I was on the side of proposal A, but Brad's points are persuasive. Making command 0 about existence only seems like the right call, and the special case for the few drivers which really want to present additional information can use a different command number.

alevy:
OK, I'm basically in agreement with @bradjc now. I think there _might_ be a compelling reason _in the future_ to modify TRD 104 to allow some returned information from `command 0`. In particular, it could save the amount of code necessary for compliant drivers that _also_ need a way to return, e.g., the number of instances _if_ there are ever such drivers that are not toys.

However, at the moment, this is _only_ used in drivers like GPIO and LED. In those cases it's a nice showcase that you can have a blink example that works seemless on boards with a varying number of buttons/leds, but it's unlikely these kinds of drivers ever really make it only "real" products, so their code size footprint isn't so important, and that functionality can still be supported in a different command number.

Conversely, there is both a cleanliness and potential code size saving from providing command 0 in the kernel itself, rather than each driver implementing it on its own. Suggesting that this may simply be the better design, rather than sticking with a "mistake" in TRD 104.

_If_ we run into compelling use cases in the future where changing command 0, we can revisit TRD 104, most likely in a highly backwards compatible way.

ppannuto:
I'm also pro-B. I tried to update the TRD draft appropriately and capture this discussion in the repo in a reasonable / permanent way.

I'm unfortunately away with family this Friday, but hopefully this can be discussed on the core call and finalized? I think it's a relatively small change.

alexandruradovici:
I was also going to side with A, but I think @bradjc has a very good point. 

The only observation that I have is that the code in the kernel should be placed after the boards' decision of wheter to deliver or not the system call. I would like to keep the boards' ability to deny it.

alevy:
> The only observation that I have is that the code in the kernel should be placed after the boards' decision of wheter to deliver or not the system call. I would like to keep the boards' ability to deny it.

@alexandruradovici. This is important for access control, for example.

doc/reference/trd104-syscalls.md
```diff
@@ -5,10 +5,10 @@ System Calls
 **Working Group:** Kernel<br/>
 **Type:** Documentary<br/>
 **Status:** Draft <br/>
-**Author:** Hudson Ayers, Guillaume Endignoux, Jon Flatley, Philip Levis, Amit Levy, Leon Schuermann, Johnathan Van Why, dcz <br/>
+**Author:** Hudson Ayers, Guillaume Endignoux, Jon Flatley, Philip Levis, Amit Levy, Pat Pannuto, Leon Schuermann, Johnathan Van Why, dcz <br/>
 **Draft-Created:** August 31, 2020<br/>
-**Draft-Modified:** March 9, 2022<br/>
-**Draft-Version:** 7<br/>
+**Draft-Modified:** August 21, 2022<br/>
```
bradjc, line 10:
```suggestion
**Draft-Modified:** August 21, 2023<br/>
```

doc/reference/trd104-syscalls.md
```diff
@@ -496,14 +496,19 @@ failure variant of `Failure`, with an associated error code of
...
 
-Every device driver MUST implement command number 0 as the
-"exists" command.  This command always returns `Success`. This command
-allows userspace to determine if a particular system call driver is
-installed; if it is, the command returns `Success`. If it is not, the
-kernel returns `Failure` with an error code of `NODEVICE`.
+4.3.1 Command Idenfitier 0
+--------------------------
+
+ - [Added August, 2023](../rfcs/2023-08-18--CommandZeroSemantics.md).
```
bradjc, line 503: What does this mean to be added August 23? I'm not sure how future readers should interpret this link.

bradjc:
What happened to the kernel impl for returning success?

Can we still merge that if it was something like 

```rust
with_driver(|d| {
   if command == 0 {
     ret = d.command()
     if ret.is_success() {
      return ret;
    } else {
      return ReturnCode::SUCCESS
    }
  }
...
}
```

alevy:
@bradjc I think it's fine for that to be a separate PR, especially since it has implications on userspace. Perhaps a PR that "fixes" this behavior (by implementing it in the kernel) in GPIO/LED/ADC/etc while adding a new additional command to those drivers.

doc/reference/trd104-syscalls.md
```diff
@@ -496,14 +496,19 @@ failure variant of `Failure`, with an associated error code of
...
+
+Command Identifier 0 is implemented by the core kernel and provides an
```
alevy, line 505: whether or not it is provided by the core kernel seems out of scope of this TRD.

```suggestion
Command Identifier 0 provides an
```

alevy (changes requested):

bradjc:
> @bradjc I think it's fine for that to be a separate PR, especially since it has implications on userspace. Perhaps a PR that "fixes" this behavior (by implementing it in the kernel) in GPIO/LED/ADC/etc while adding a new additional command to those drivers.

Agreed, I was just wondering if that code was in this PR or I imagined it.

doc/reference/trd104-syscalls.md
ppannuto: I wanted to in some way link the context of this discussion without dumping a whole lot of text into the TRD. I updated that to `[Design RFC](../rfcs/2023-08-18--CommandZeroSemantics.md)`

doc/reference/trd104-syscalls.md
ppannuto: Updated

ppannuto:
Updated with latest comments.

Brad: You aren't crazy. I pulled the actual, in-kernel changes out of this PR when I changed it from 'draft' to 'ready' and captured them at https://github.com/tock/tock/blob/cmd0-counts/doc/rfcs/2023-08-18--CommandZeroSemantics.md#proposals-and-prototypes

alevy (approved):

bradjc (commented):
I haven't approved this because I'm not sure about the `- <a>Design RFC</a>` syntax. This would introduce a new convention for TRDs which doesn't relate to cmd 0. It gives the impression that the other designs haven't been discussed.

bradjc:
What is this blocked on?

ppannuto:
Not blocked anymore — I put that tag on when I had direct code changes on this nominally doc/spec only PR; fixed.

phil-levis:
One question: should we finalize 104 for 2.x and then make subsequent things follow-up TRDs?

alexandruradovici:
I would prefer B.

doc/reference/trd104-syscalls.md
```diff
@@ -496,14 +496,24 @@ failure variant of `Failure`, with an associated error code of
...
 
+
 4.3.1 Command Identifier 0
---------------------------------
+--------------------------
+
+ - [Design RFC](../rfcs/2023-08-18--CommandZeroSemantics.md).
+
+Command Identifier 0 provides an existence check for drivers. If a
+driver is accessible to calling application, the kernel will return
```
phil-levis, line 506: Why is the text rewritten to say the exact same thing? It doesn't constrain the Success type, which was the start of this whole discussion? It is not using keywords.  This is also prescribing *who* returns the `Success`, which this as an API specification shouldn't do (I agree the current text is not great on this either). I would suggest:

Command Identifier 0 provides an existence check for drivers. A return value of `Success` indicates that the driver is present and the userspace process can issue system calls to it. If the driver is not accessible, Command Identifier 0 returns `Failure` with an error code of `ENODEVICE`. A driver may be not accessible because the kernel does not have it, the process does not have the required permissions to use it, or other reasons.

phil-levis (changes requested):

doc/reference/trd104-syscalls.md
ppannuto: Updated with this text √

ppannuto:
Updated with Phil's suggested text. I added a section at the end for "References and Additional Information" per our discussion on the call of 'adding the pointer to the RFC as a reference' — I avoided the heading "Reference" as that has a different meaning per TRD1, not quite sure if that's what folks had in mind?

alevy (dismissed):

bradjc (approved):

lschuermann (dismissed):

doc/rfcs/2023-08-18--CommandZeroSemantics.md
```diff
@@ -0,0 +1,239 @@
+Syscalls: Commando 0 Semantics
```
jrvanwhy, line 1: This was supposed to be `Command` not `Commando`, right?

doc/rfcs/2023-08-18--CommandZeroSemantics.md
```diff
@@ -0,0 +1,239 @@
...
++Device drivers MUST return `Success` OR a success variant for command
++number 0. Drivers MAY use success variants to convey additional
++information, e.g. an LED driver might return the number of LEDs
++physically present on the board via `Success with u32`. Command 0
++implementation MUST NOT have ANY runtime effects. Any code present in
++the `0` match arm MUST be suitable for [constant evaluation by the
++Rust compiler](https://doc.rust-lang.org/reference/const_eval.html).
++
++
++4.3.1 Command Idenfitier 0 [PROPOSAL B]
```
jrvanwhy, line 116: Identifier

bradjc (approved):

phil-levis (approved):
I'm fine with moving to Approve, but looking back on the original text, I think this was very much a non-problem. It was clear only `Success` was allowed, and ultimately a specification is its rules, not its reasoning; we don't have supplementary links for the detailed discussions behind subscribe and allow semantics (which were big and detailed problems that required a lot of discussion and revisions), so having a solitary supplementary material link at the bottom for this little detail seems weird. That being said, it sounds like I'm the outlier here, and it's not a big deal, so happy to go along.

PR #3628, opened 2023-08-19
Update deprecated VSCode settings

CosminGGeorgescu:
### Pull Request Overview

This pull request removes the `rust-client.channel` field, part of the [Rust VSCode extension](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust), which is deprecated as of 15.11.2022, and replaces it with the equivalent setting for the new [rust-analyzer extension](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust-analyzer). However, there could be a case made for developers that install the extension by source code (somehow ? I'm not familiar with this way), as the [Github repo](https://github.com/rust-lang/vscode-rust) is still public, just archived.

The second field `"rust-analyzer.check.allTargets": false,` was added to fix the `found duplicate lang item ``panic_impl`` ` error and associated squiggles sometimes raised by rust-analyzer, as it runs `cargo check` with the `--all-targets` flag by default and raises an error for each `pub unsafe extern "C" fn panic_fmt(pi: &PanicInfo)` that is not preceded by `#[cfg(not(test))]`.

### Testing Strategy

This pull request was tested by checking with the official rust-analyzer documentation and restarting the rust-analyzer server to see that the added changes had the desired effects.

### TODO or Help Wanted

If there are any suggestion as what more to be enforced upon the VSCode developer, let me know and I'll add them here.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


ppannuto (dismissed):
I don't believe anyone on the core team (except maybe @hudson-ayers) use VSCode, so this has always been something we defer to community folks to maintain. Changes look reasonable to me — thanks!

.vscode/launch.json
```diff
@@ -30,4 +30,4 @@
             ]
         }
     ]
-}
+}
```
bradjc, line 33: ?

.vscode/settings.json
```diff
@@ -1,4 +1,7 @@
 {
     "editor.formatOnSave": true,
-    "rust-client.channel": "nightly-2023-07-30",
-}
+    "rust-analyzer.server.extraEnv": {
+        "RUSTUP_TOOLCHAIN": "nightly-2023-07-30"
+    },
+    "rust-analyzer.check.allTargets": false,
+}
```
bradjc, line 7: ?

bradjc: (why remove newline?)

CosminGGeorgescu: Was a case of `"files.insertFinalNewline": false` being the default in VSCode and `formatOnSave` just doing it's job. Should I add them back ? Additionally, I could add `"files.insertFinalNewline": true` to `settings.json` and override the user's default, to keep a rule.

.vscode/launch.json
CosminGGeorgescu: Same as below.

hudson-ayers:
> I don't believe anyone on the core team (except maybe @hudson-ayers) use VSCode, so this has always been something we defer to community folks to maintain. Changes look reasonable to me — thanks!

Nope, I use neovim! So no stake in this, but the changes look fine to me.

alexandruradovici:
I am using VS Code, I'll take a look at the changes, but so far they look good to me.

bradjc (approved):

PR #3633, opened 2023-08-21
clippy: -D unnecessary_cast

bradjc:
### Pull Request Overview

This pull request updates the tock source so that we pass the `unnecessary_cast` lint.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/cortex-m/src/mpu.rs
+++ b/arch/cortex-m/src/mpu.rs
@@ -472,7 +472,7 @@ impl<const NUM_REGIONS: usize, const MIN_REGION_SIZE: usize> mpu::MPU
                 let tz = start.trailing_zeros();
                 if tz < 32 {
                     // Find the largest power of two that divides `start`
-                    (1 as usize) << tz
+                    1_usize << tz
                 } else {
                     // This case means `start` is 0.
                     let mut ceil = math::closest_power_of_two(size as u32) as usize;
--- a/boards/qemu_rv32_virt/src/io.rs
+++ b/boards/qemu_rv32_virt/src/io.rs
@@ -8,7 +8,6 @@ use core::str;
 
 use kernel::debug;
 use kernel::debug::IoWrite;
-use rv32i;
 
 use crate::CHIP;
 use crate::PROCESSES;
--- a/capsules/extra/src/touch.rs
+++ b/capsules/extra/src/touch.rs
@@ -177,7 +177,7 @@ impl<'a> Touch<'a> {
                 }
                 ScreenRotation::Rotated270 => {
                     mem::swap(&mut width, &mut height);
-                    (width as u16 - touch_event.y as u16, touch_event.x)
+                    (width as u16 - touch_event.y, touch_event.x)
                 }
                 _ => (touch_event.x, touch_event.y),
             };
--- a/arch/cortex-m/src/systick.rs
+++ b/arch/cortex-m/src/systick.rs
@@ -62,8 +62,7 @@ pub struct SysTick {
 }
 
 const BASE_ADDR: *const SystickRegisters = 0xE000E010 as *const SystickRegisters;
-const SYSTICK_BASE: StaticRef<SystickRegisters> =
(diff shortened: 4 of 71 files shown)


arch/cortex-m/src/mpu.rs
```diff
@@ -472,7 +472,7 @@ impl<const NUM_REGIONS: usize, const MIN_REGION_SIZE: usize> mpu::MPU
                 let tz = start.trailing_zeros();
                 if tz < 32 {
                     // Find the largest power of two that divides `start`
-                    (1 as usize) << tz
+                    1_usize << tz
```
ppannuto, line 475: I acknowledge that this is presumably the more idiomatic Rust, and fine, whatever, we can write this way, but wow this is much less clear to my eyes.

boards/qemu_rv32_virt/src/io.rs
```diff
@@ -8,7 +8,6 @@ use core::str;
 
 use kernel::debug;
 use kernel::debug::IoWrite;
-use rv32i;
```
ppannuto, line 11: Weird this was caught in this particular lint, but w/e I guess

capsules/extra/src/touch.rs
```diff
@@ -177,7 +177,7 @@ impl<'a> Touch<'a> {
                 }
                 ScreenRotation::Rotated270 => {
                     mem::swap(&mut width, &mut height);
-                    (width as u16 - touch_event.y as u16, touch_event.x)
+                    (width as u16 - touch_event.y, touch_event.x)
```
ppannuto, line 180: This one is a little weirder as I think `touch_event.y` is still being coerced to a `u16` but _only_ becuase the `width` in the LHS has been coerced. This feels fragile. Maybe better to split into separate lines / temporary variables?

ppannuto (changes requested):
I swear once upon a compiler a lot of these were required.

I kinda liked the intentionality of type conversion Rust asked for once upon a time, but the ergonomics of implicit conversion of unambiguous casts is admittedly nice.

capsules/extra/src/touch.rs
lschuermann: Seems that `touch_event.y` is actually a `u16`: https://github.com/tock/tock/blob/0a2b74eba355dba42907fb6f5d693166ad4539be/kernel/src/hil/touch.rs#L35

AFAIK, Rust (still) does not coerce types.

lschuermann:
> I kinda liked the intentionality of type conversion Rust asked for once upon a time, but the ergonomics of implicit conversion of unambiguous casts is admittedly nice.

I could be wrong, I don't think that's what's happening here. Rust only implicitly assumes an integer time for literals which aren't otherwise "driven to a type" (e.g. through the `1_usize`). I believe many of these cases are just ones where we've changed the actual type underneath, such that our casts have literally become unnecessary.

lschuermann (approved):

boards/qemu_rv32_virt/src/io.rs
bradjc: Yeah it wasn't, clippy fix sometimes deletes a bunch of imports from some board's io.rs file. When I fixed this one I only put back what was necessary.

I don't really understand why io.rs and rust behave so strangely.

hudson-ayers (approved):

arch/cortex-m/src/mpu.rs
brghena: I think it's the underscore that makes this seem weird to me. Would it be more clear without the underscore @ppannuto?

```rust
    1usize << tz
```

versus

```rust
    1_usize << tz
```

brghena: But we should really be consistent. So if `0x8000000_u32` and `0b01_u32` seem like the clear way to write them (and I think the underscore _is_ better there), then `1_usize` is for the best.

brghena (approved):

ppannuto (approved):

PR #3635, opened 2023-08-21
clippy: -D needless_borrow

bradjc:
### Pull Request Overview

This pull request removes cases where we borrow something we don't have to (which I'm assuming Rust just undoes anyway).

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


lschuermann (dismissed):
Slightly skittish about this one: when removing the `ref` in some of the match statements, I'm worried that sometimes we might now be doing a copy where there might've been a reason to use a borrow (for performance reasons, drop semantics, etc.).

In general this change seems good though -- double borrows can also do a lot of harm. We should be open to allowing this lint in individual cases where it makes sense.

hudson-ayers (approved):
All of the binaries are identical before/after this change so I am not too concerned about any performance implications of this

alevy (approved):

PR #3643, opened 2023-08-23
Clippy: Enable many style lints

bradjc:
### Pull Request Overview

This pull request enables many of the clippy style lints that to me seem noncontroversial.

You can see the change made by each one in each commit. If one seems like it needs more discussion I'm happy to move it to a new PR.

I also fixed a mistake from an earlier clippy PR where we both denied and allowed a flag (before I decided to enumerate all allows).

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/extra/src/net/network_capabilities.rs
+++ b/capsules/extra/src/net/network_capabilities.rs
@@ -49,7 +49,7 @@ impl AddrRange {
                 let full_bytes: usize = prefix_len / 8;
                 let remainder_bits: usize = prefix_len % 8;
                 // initial bytes -- TODO: edge case
-                if &allowed_addr.0[0..full_bytes] != &addr.0[0..full_bytes] {
+                if allowed_addr.0[0..full_bytes] != addr.0[0..full_bytes] {
                     false
                 } else if remainder_bits == 0 {
                     true //this case is necessary bc right shifting a u8 by 8 bits is UB
--- a/chips/lowrisc/src/spi_host.rs
+++ b/chips/lowrisc/src/spi_host.rs
@@ -360,16 +360,15 @@ impl<'a> SpiHost<'a> {
     /// Currently only Bi-Directional transactions are supported
     fn start_transceive(&self) {
         let regs = self.registers;
-        //8-bits that describe command transfer len (cannot exceed 255)
-        let num_transfer_bytes: u32;
         //TXQD holds number of 32bit words
         let txfifo_num_bytes = regs.status.read(status::TXQD) * 4;
 
-        if txfifo_num_bytes > u8::MAX as u32 {
-            num_transfer_bytes = u8::MAX as u32;
+        //8-bits that describe command transfer len (cannot exceed 255)
+        let num_transfer_bytes: u32 = if txfifo_num_bytes > u8::MAX as u32 {
+            u8::MAX as u32
         } else {
-            num_transfer_bytes = txfifo_num_bytes;
-        }
+            txfifo_num_bytes
+        };
 
         //Flush all data in TXFIFO and assert CSAAT for all
         // but the last transfer segment.
--- a/chips/lowrisc/src/spi_host.rs
+++ b/chips/lowrisc/src/spi_host.rs
@@ -278,7 +278,7 @@ impl<'a> SpiHost<'a> {
                             break;
                         }
                         self.rx_offset.set(self.rx_offset.get() + 1);
-                        shift_mask = shift_mask << 8;
+                        shift_mask <<= 8;
                     }
                 }
                 //Save buffer!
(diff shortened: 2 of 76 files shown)


lschuermann (approved):
These all broadly make sense to me.

capsules/extra/src/hmac.rs
```diff
@@ -147,9 +147,7 @@ impl<
                             })
                         })
                         .unwrap_or(Err(ErrorCode::RESERVE));
-                    if ret.is_err() {
-                        return ret;
-                    }
+                    ret?;
```
ppannuto, line 150: I know this is super-Rusty, but I need my IDE to quintuple the font size of the `?` or something — hiding an early-exit from a function in a single character is not my favorite design decision.

capsules/extra/src/net/network_capabilities.rs
```diff
@@ -49,7 +49,7 @@ impl AddrRange {
                 let full_bytes: usize = prefix_len / 8;
                 let remainder_bits: usize = prefix_len % 8;
                 // initial bytes -- TODO: edge case
-                if &allowed_addr.0[0..full_bytes] != &addr.0[0..full_bytes] {
+                if allowed_addr.0[0..full_bytes] != addr.0[0..full_bytes] {
```
ppannuto, line 52: The fact that these two expressions are equivalent... this is why borrows are so hard to teach :/

chips/lowrisc/src/spi_host.rs
```diff
@@ -361,15 +361,14 @@ impl<'a> SpiHost<'a> {
...
 
-        if txfifo_num_bytes > u8::MAX as u32 {
-            num_transfer_bytes = u8::MAX as u32;
+        let num_transfer_bytes: u32 = if txfifo_num_bytes > u8::MAX as u32 {
+            u8::MAX as u32
         } else {
-            num_transfer_bytes = txfifo_num_bytes;
-        }
+            txfifo_num_bytes
+        };
```
ppannuto, line 371: The comment for this variable needs to move too

kernel/src/process_standard.rs
```diff
@@ -578,16 +578,12 @@ impl<C: Chip> Process for ProcessStandard<'_, C> {
...
-                return None;
-            }
+            )?;
 
             for region in self.mpu_regions.iter() {
                 if region.get().is_none() {
-                    region.set(new_region);
-                    return new_region;
+                    region.set(Some(new_region));
+                    return Some(new_region);
```
ppannuto, line 586: This is weird — why did this compile before?

kernel/src/utilities/math.rs
```diff
@@ -115,7 +115,7 @@ fn ln_1to2_series_approximation(x: f32) -> f32 {
     //https://en.wikipedia.org/wiki/Remez_algorithm
     let ln_1to2_polynomial: f32 = -1.741_793_9_f32
         + (2.821_202_6_f32
-            + (-1.469_956_8_f32 + (0.447_179_55_f32 - 0.056_570_851_f32 * x_working) * x_working)
+            + (-1.469_956_8_f32 + (0.447_179_55_f32 - 0.056_570_85_f32 * x_working) * x_working)
```
ppannuto, line 118: Did this just trim a bit off the constant here?

ppannuto (changes requested):
Two screaming into the void comments, three real ones

kernel/src/utilities/math.rs
brghena: That final `1` is beyond the precision of a 32-bit float. Your options are 0.05657085**0**3127 or 0.05657085**4**038. It looks like there are a couple clippy lints that look at floating point literals to ensure that they make sense.

capsules/extra/src/hmac.rs
bradjc: Yeah, that's the pain point with not having these lints on from the start. The `?` should be on the original function call. But then that changes the types and this is no longer a mechanical change, it's more invasive.

If the developer had gotten this clippy warning, they would have realized they could have used the `?` earlier and we wouldn't have this odd line. This is fixable, but not in the spirit of these clippy prs.

kernel/src/process_standard.rs
bradjc: Yeah ok I guess I did the more invasive change here. I'll move this change to its own PR.

kernel/src/utilities/math.rs
ppannuto: On the one hand, that makes sense — on the other hand, curious what it was doing before... always wary about changing magic constants, even in the insignificant bits.

@alexandruradovici , looks like you added this in 8daa35 — any issues here?

bradjc: As I understand it this is the same as before, Rust was truncating the one.

https://rust-lang.github.io/rust-clippy/master/index.html#/excessive_precision

chips/lowrisc/src/spi_host.rs
bradjc: Moved.

ppannuto (dismissed):

lschuermann (dismissed):

hudson-ayers (dismissed):

alevy (dismissed):

ppannuto (approved):

PR #3645, opened 2023-08-23
Clippy: -D question_mark

bradjc:
### Pull Request Overview

This pull request fixes the question mark clippy lint that says to use `?` instead of checking the error and then returning it.

This makes some changes to make the ? use more natural.

Skips the epmp because it is changing anyway.

### Testing Strategy

travis

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/arch/rv32i/src/epmp.rs
+++ b/arch/rv32i/src/epmp.rs
@@ -612,9 +612,7 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::MPU
 
         let region = PMPRegion::new_app(start as *const u8, size, permissions);
 
-        if region.is_none() {
-            return None;
-        }
+        region?;
 
         config.regions[region_num] = region;
         config.is_dirty.set(true);
@@ -706,9 +704,7 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::MPU
             permissions,
         );
 
-        if region.is_none() {
-            return None;
-        }
+        region?;
 
         config.regions[region_num] = region;
 
@@ -962,9 +958,7 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::KernelM
 
         let region = PMPRegion::new_kernel(start as *const u8, size, permissions);
 
-        if region.is_none() {
-            return None;
-        }
+        region?;
 
         config.regions[region_num] = region;
 
--- a/capsules/extra/src/hmac.rs
+++ b/capsules/extra/src/hmac.rs
@@ -147,9 +147,7 @@ impl<
                             })
                         })
                         .unwrap_or(Err(ErrorCode::RESERVE));
-                    if ret.is_err() {
-                        return ret;
-                    }
+                    ret?;
 
                     kernel_data
                         .get_readonly_processbuffer(ro_allow::DATA)
--- a/capsules/extra/src/hmac.rs
+++ b/capsules/extra/src/hmac.rs
@@ -120,7 +120,7 @@ impl<
         self.processid.map_or(Err(ErrorCode::RESERVE), |processid| {
             self.apps
                 .enter(processid, |app, kernel_data| {
-                    let ret = kernel_data
+                    kernel_data
                         .get_readonly_processbuffer(ro_allow::KEY)
                         .and_then(|key| {
                             key.enter(|k| {
@@ -146,8 +146,7 @@ impl<
                                 }
                             })
                         })
-                        .unwrap_or(Err(ErrorCode::RESERVE));
-                    ret?;
+                        .unwrap_or(Err(ErrorCode::RESERVE))?;
 
                     kernel_data
                         .get_readonly_processbuffer(ro_allow::DATA)
--- a/capsules/extra/src/sha.rs
+++ b/capsules/extra/src/sha.rs
@@ -115,16 +115,12 @@ impl<
         self.processid.map_or(Err(ErrorCode::RESERVE), |processid| {
             self.apps
                 .enter(processid, |app, kernel_data| {
-                    let ret = if let Some(op) = &app.sha_operation {
-                        match op {
-                            ShaOperation::Sha256 => self.sha.set_mode_sha256(),
-                            ShaOperation::Sha384 => self.sha.set_mode_sha384(),
-                            ShaOperation::Sha512 => self.sha.set_mode_sha512(),
-                        }
-                    } else {
-                        Err(ErrorCode::INVAL)
-                    };
-                    ret?;
+                    match app.sha_operation {
+                        Some(ShaOperation::Sha256) => self.sha.set_mode_sha256()?,
+                        Some(ShaOperation::Sha384) => self.sha.set_mode_sha384()?,
+                        Some(ShaOperation::Sha512) => self.sha.set_mode_sha512()?,
+                        _ => return Err(ErrorCode::INVAL),
+                    }
 
                     kernel_data
                         .get_readonly_processbuffer(ro_allow::DATA)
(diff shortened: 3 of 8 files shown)


arch/rv32i/src/epmp.rs
```diff
@@ -612,9 +612,7 @@ impl<const MAX_AVAILABLE_REGIONS_OVER_TWO: usize> kernel::platform::mpu::MPU
 
         let region = PMPRegion::new_app(start as *const u8, size, permissions);
 
-        if region.is_none() {
-            return None;
-        }
+        region?;
```
lschuermann, line 615:
```suggestion
        let _ = region?;
```
Maybe that could be a more natural way to write this? I agree with @ppannuto that an expression consisting of a variable name followed by a `?`, while silently discarding the unpacked value, is not a particularly great way to implement an early-abort. However, it is quite natural to use a `?` for this purpose to extract the contained value at the end of more generic, complex expressions. Maybe `let _ =` can be used to make it clear that we're not actually interested in the contained value here.

bradjc: Aren't you rewriting this anyway?

ppannuto: I'm actually more okay with this 'line of code on its own with a question mark'. It's clear that the `?` has to be doing _something_ otherwise why would this line even be here?

It's the stuff like this:

    let region = PMPRegion::new_app(start as *const u8, size, permissions)?;

where that little, easy-to-miss `?` hanging out at the end magically injects very easily overlooked control flow that I think is a an absolute crime and a terrible anti-feature.

capsules/extra/src/hmac.rs
```diff
@@ -146,10 +146,7 @@ impl<
                                 }
                             })
                         })
-                        .unwrap_or(Err(ErrorCode::RESERVE));
-                    if ret.is_err() {
-                        return ret;
-                    }
+                        .unwrap_or(Err(ErrorCode::RESERVE))?;
```
ppannuto, line 149: This is a great example. Let's tack a single `?` onto the end of a 25-line block that can invisibly cause an early-return from the function.

Now, realistically, I think I should get over myself and we should do this the 'rust-y' way and make this change. But I will 'old man yell at cloud' and assert that this is an absolutely insane language feature that dramatically harms the readability of Rust code for anyone except full-time Rust developers.

arch/rv32i/src/epmp.rs
lschuermann: > Aren't you rewriting this anyway?

Well, yes. I was just trying to see whether @ppannuto's concerns could've been addressed by a `let _ =` binding.

capsules/extra/src/hmac.rs
alevy: FWIW, I think `?` really becomes most clearly useful when it's used really ubiquitously, along-side `Into` implementations between error types. It encourages using meaningful error types (potentially including some state), and can make really long control-flowy code much terser and easier to read, where the control flow is all uniform and easy to reason about---this function is a combinator of a bunch of functions that may or may not succeed, and each of them may return early. Think parser-combinators.

It's a really common pattern in Haskell and is really great when it's used to it's full extent.

All to say, if we start using this _more_ we might get over a hump where there is better local maximum

arch/rv32i/src/epmp.rs
alevy: I guess, yeah, the "Rust-y" version would be

```
let region = PMPRegion::new_app(start as *const u8, size, permissions)?;
```

capsules/extra/src/hmac.rs
alevy, line 149: Two better options

```suggestion
                        .or(Err(ErrorCode::RESERVE))?;
```

```suggestion
                        .map_err(|e| ErrorCode::RESERVE)?;
```

capsules/extra/src/sha.rs
```diff
@@ -115,17 +115,11 @@ impl<
...
-                        Err(ErrorCode::INVAL)
-                    };
-                    if ret.is_err() {
-                        return ret;
+                    match app.sha_operation {
+                        Some(ShaOperation::Sha256) => self.sha.set_mode_sha256()?,
+                        Some(ShaOperation::Sha384) => self.sha.set_mode_sha384()?,
+                        Some(ShaOperation::Sha512) => self.sha.set_mode_sha512()?,
+                        _ => return Err(ErrorCode::INVAL),
                     }
```
alevy, line 123: I think either of these is more clear what's going on:

```
app.sha_operation.ok_or(ErrorCode::INVAL).map(|op| {
                        ShaOperation::Sha256 => self.sha.set_mode_sha256(),
                        ShaOperation::Sha384 => self.sha.set_mode_sha384(),
                        ShaOperation::Sha512 => self.sha.set_mode_sha512(),
                    })?;
```

```
let op = app.sha_operation.ok_or(ErrorCode::INVAL)?;
match op {
    ShaOperation::Sha256 => self.sha.set_mode_sha256()?,
    ShaOperation::Sha384 => self.sha.set_mode_sha384()?,
    ShaOperation::Sha512 => self.sha.set_mode_sha512()?,
}
```

First, we transform the `Option` into a `Result` that basically says: if you passed in `None` that was invalid. Then, we get to assume it's not erroneous, and we match on the operation, erroring if setting the mode is unsupported (or whatever the error is).

capsules/extra/src/symmetric_encryption/aes.rs
```diff
@@ -91,25 +91,19 @@ impl<
...
-                            }
-                            AesOperation::AES128CBC(encrypt) => {
-                                self.aes.set_mode_aes128cbc(*encrypt)
-                            }
-                            AesOperation::AES128ECB(encrypt) => {
-                                self.aes.set_mode_aes128ecb(*encrypt)
-                            }
-                            AesOperation::AES128CCM(_encrypt) => Ok(()),
-                            AesOperation::AES128GCM(_encrypt) => Ok(()),
+                    match app.aes_operation {
```
alevy, line 94: Ditto here to the SHA version. Though, of course, many correct patterns can co-exist. It's just my taste.

capsules/extra/src/symmetric_encryption/aes.rs
```diff
@@ -134,21 +128,15 @@ impl<
                                             AesOperation::AES128Ctr(_)
                                             | AesOperation::AES128CBC(_)
                                             | AesOperation::AES128ECB(_) => {
-                                                if let Err(e) = AES128::set_key(self.aes, buf) {
-                                                    return Err(e);
-                                                }
+                                                AES128::set_key(self.aes, buf)?;
```
alevy, line 131: This whole block can be simplified dramatically. I can take a pass, or we can avoid additional changes in this PR, and just focus on making clippy happy, cleaning up code style even more separately.

alevy (dismissed):
Some comments, but personally I'm for merging as is rather than blocking the clippy work on style bikeshedding

bradjc:
I would prefer to keep the clippy PRs just fixes for clippy so we don't have to review these PRs for other changes as well.

capsules/extra/src/hmac.rs
lschuermann: @alevy The first one changes the semantics (and I'm not sure would pass the type-checker). For a `Result<T, E>`, `unwrap_or()` takes a `T` and returns either the `Ok(T)` contained in the result, or the default value. `or()` will return a different `Result<T, F>` passed in, in case the original result is an `Err(_)`.

Second case is good, but we should rename `e` or `_` to avoid an unused closure argument warning.

ppannuto (dismissed):

hudson-ayers (approved):

capsules/extra/src/hmac.rs
alevy: You're right, the first one should not have an enclosing `Err`

alevy (approved):

alevy:
The dismissed reviews are from doing a trivial merge with `master`, thus I'm making an executive decision to merge this as long as I didn't break the build, assuming that @hudson-ayers and @ppannuto would trivially re-appprove.

PR #3652, opened 2023-08-30
15.4 link layer security - encryption/decryption

tyler-potyondy:
### Pull Request Overview

This pull request fixes the 15.4 framer's encryption/decryption of 15.4 packets secured with the 15.4 link layer security. The existing 15.4 driver exposes functionality to add network keys, specifies the level of encryption required when sending a 15.4 packet, and decrypts received packets using the added keys. However, in testing this functionality, Tock either crashed or sent/received errant values (tested on the nrf52840dk).

A majority of these errors were initially due to parsing/offset issues that have now been resolved. After resolving these more minor errors, a larger issue appeared that led the entire kernel to crash. For context, the nrf52840dk 15.4 radio driver performs an error check that panics if the radio driver's receive buffer is empty when beginning a receive operation. The driver is written such that this buffer will only be empty if the `client.receive(...)` function fails to set the radio's receive buffer prior to returning. The current state machine for the 15.4 framer fails to set the radio's buffer prior to returning from the receive function.

The original framer would not set the buffer until after the crypto operation had completed. However, it would return from `receive(...)` prior to the crypto operation completing. This creates a challenge as the crypto operation must hold the buffer until completing the operation. This leads to either solving the issue by: (1) delaying the return from the framer until after the crypto operation completes, or (2) creating a separate buffer to copy the packet into. Solution (1) creates an obvious issue as the radio will not receive any new packets while waiting for the crypto operation to complete. This led to the adoption of solution (2). This involved adding a static buffer to the framer so that when receiving an encrypted packet, the packet could be copied from the radio's buffer to the framer's buffer. This allows for the radio's recv buffer to be set prior to returning. Although solution (2) possess the obvious setback of greater memory usage, this seems superior to simply dropping packets while waiting for a crypto operation to complete. 

The original 15.4 capsule does not send the extended address by default. Although this does not necessarily violate the 15.4 spec or impair the capsule's operation, this creates a challenge in monitoring encrypted packets in wireshark as wireshark cannot decrypt the packet without the extended source address. Moreover, the framer requires the extended source address when creating the nonce for transmitting encrypted packets. The 15.4 spec states that the extended MacAddress should be an EUI-64. This address is now passed as an argument when creating the 15.4 radio driver component in `main.rs`. This creates a challenge as not all board families have an existing function to access the device's EUI-64. For these boards, I added a default extended source address.

As a minor note, the nrf52840dk device ID should be used instead of the serial number for generating the MacAddress. The device ID is a 64 bit unique device identifier, while the serial number is only 48 bit. I updated the address function to read this register instead.

### Testing Strategy

This pull request was tested by running:
-the libtock-c 15.4 rx/tx apps to test standard 15.4 unsecured operation
-new libtock-c 15.4 rx/tx encrypted apps to test encrypted operation (will submit PR to libtock-c for these new apps)
-UDP send/rx to test higher layers use of the 15.4 functionality 

### TODO or Help Wanted

With the change of requiring the extended MacAddress when creating a 15.4 component, all boards operating with 15.4 must implement a method for generating a EUI64. The nrf52840dk has been updated accordingly, but support for this must still be added to imix, nrf52840 dongle, particle boron, sma q3. For now, these boards have a default extended mac address that is hard coded.

To determine the length of the m data, the previous implementation computed a bitwise or as seen in line 217 of the unmodified framer.rs. This led to the wrong value being computed. I revised this value to be computed as the difference. I do not see a reason for a bitwise or to compute this length, but I wanted to draw attention to this in case I am overlooking something here.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/nordic/nrf52840dk/src/main.rs
+++ b/boards/nordic/nrf52840dk/src/main.rs
@@ -562,17 +562,18 @@ pub unsafe fn main() {
     // IEEE 802.15.4 and UDP
     //--------------------------------------------------------------------------
 
-    let serial_num = nrf52840::ficr::FICR_INSTANCE.address();
-    let serial_num_bottom_16 = serial_num[0] as u16 + ((serial_num[1] as u16) << 8);
-    let src_mac_from_serial_num: MacAddress = MacAddress::Short(serial_num_bottom_16);
+    let device_id = nrf52840::ficr::FICR_INSTANCE.address();
+
+    let device_id_bottom_16: u16 = device_id[0] as u16 + ((device_id[1] as u16) << 8);
+    let short_mac_addr: MacAddress = MacAddress::Short(device_id_bottom_16);
     let (ieee802154_radio, mux_mac) = components::ieee802154::Ieee802154Component::new(
         board_kernel,
         capsules_extra::ieee802154::DRIVER_NUM,
         &nrf52840_peripherals.ieee802154_radio,
         aes_mux,
         PAN_ID,
-        serial_num_bottom_16,
-        serial_num,
+        device_id_bottom_16,
+        device_id,
     )
     .finalize(components::ieee802154_component_static!(
         nrf52840::ieee802154_radio::Radio,
@@ -591,7 +592,7 @@ pub unsafe fn main() {
                 0x1e, 0x1f,
             ]),
             IPAddr::generate_from_mac(capsules_extra::net::ieee802154::MacAddress::Short(
-                serial_num_bottom_16
+                device_id_bottom_16
             )),
         ]
     );
@@ -601,7 +602,7 @@ pub unsafe fn main() {
         DEFAULT_CTX_PREFIX_LEN,
         DEFAULT_CTX_PREFIX,
         DST_MAC_ADDR,
-        src_mac_from_serial_num,
+        short_mac_addr,
         local_ip_ifaces,
         mux_alarm,
     )
--- a/boards/nordic/nrf52840dk/src/main.rs
+++ b/boards/nordic/nrf52840dk/src/main.rs
@@ -572,6 +572,7 @@ pub unsafe fn main() {
         aes_mux,
         PAN_ID,
         serial_num_bottom_16,
+        serial_num,
     )
     .finalize(components::ieee802154_component_static!(
         nrf52840::ieee802154_radio::Radio,
--- a/capsules/extra/src/ieee802154/framer.rs
+++ b/capsules/extra/src/ieee802154/framer.rs
@@ -496,6 +496,8 @@ impl<'a, M: Mac<'a>, A: AES128CCM<'a>> Framer<'a, M, A> {
 
         match result {
             None => {
+                // The packet was not encrypted, we completed the 15.4 framer procedure, and passed the packet to the
+                // client. We can now return the recv buffer to the radio driver and enter framer's idle state.
                 self.mac.set_receive_buffer(buf);
                 RxState::Idle
             }
@@ -634,7 +636,7 @@ impl<'a, M: Mac<'a>, A: AES128CCM<'a>> Framer<'a, M, A> {
                                 //   the radio drivers recv buffer and return the framer recv state machine to idle
                                 // - (4) The crypto buffer is empty (in use elsewhere) and we are unable to copy the received
                                 //   packet. This packet is dropped and we must return the buffer to the radio driver. This
-                                //   scenario is handled in None case
+                                //   scenario is handled in the None case
                                 match res {
                                     // Scenario 1
                                     Some(Ok(())) => {
@@ -757,7 +759,10 @@ impl<'a, M: Mac<'a>, A: AES128CCM<'a>> MacDevice<'a> for Framer<'a, M, A> {
         // TODO: For Thread, in the case of `KeyIdMode::Source4Index`, the source
         // address should instead be some constant defined in their
         // specification.
-        let mut src_addr = src_addr; // NOT SURE IF THIS IS OPTIMAL
+
+        // copy function parameter src_addr as a mut variable so that we can update the src address to extended address if we are encrypting
+        let mut src_addr = src_addr;
+
         let src_addr_long = self.get_address_long();
         let security_desc = security_needed.and_then(|(level, key_id)| {
             self.lookup_key(level, key_id).map(|key| {
@@ -799,7 +804,7 @@ impl<'a, M: Mac<'a>, A: AES128CCM<'a>> MacDevice<'a> for Framer<'a, M, A> {
             dst_pan: Some(dst_pan),
             dst_addr: Some(dst_addr),
             src_pan: Some(src_pan),
-            src_addr: Some(src_addr), //Not entirely sure what to do here?
+            src_addr: Some(src_addr),
             security: security,
             header_ies: Default::default(),
             header_ies_len: 0,
--- a/capsules/extra/src/ieee802154/framer.rs
+++ b/capsules/extra/src/ieee802154/framer.rs
@@ -89,6 +89,7 @@ use kernel::hil::radio;
 use kernel::hil::symmetric_encryption::{CCMClient, AES128CCM};
 use kernel::processbuffer::ReadableProcessSlice;
 use kernel::utilities::cells::{MapCell, OptionalCell};
+use kernel::utilities::leasable_buffer::SubSliceMut;
 use kernel::ErrorCode;
 
 /// A `Frame` wraps a static mutable byte slice and keeps just enough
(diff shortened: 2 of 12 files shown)


boards/nordic/nrf52840dk/src/main.rs
```diff
@@ -562,16 +562,18 @@ pub unsafe fn main() {
     // IEEE 802.15.4 and UDP
     //--------------------------------------------------------------------------
 
-    let serial_num = nrf52840::ficr::FICR_INSTANCE.address();
-    let serial_num_bottom_16 = serial_num[0] as u16 + ((serial_num[1] as u16) << 8);
-    let src_mac_from_serial_num: MacAddress = MacAddress::Short(serial_num_bottom_16);
+    let device_id = nrf52840::ficr::FICR_INSTANCE.address();
+
+    let device_id_bottom_16: u16 = device_id[0] as u16 + ((device_id[1] as u16) << 8);
```
ppannuto, line 567:
```suggestion
    let device_id_bottom_16 = u16::from_le_bytes([device_id[0], device_id[1]]);
```
match the other boards / simpler?

capsules/extra/src/ieee802154/framer.rs
```diff
@@ -567,39 +580,84 @@ impl<'a, M: Mac<'a>, A: AES128CCM<'a>> Framer<'a, M, A> {
...
+                                // Log debug statement here so that this error does not occur silently
+                                kernel::debug!(
+                                    "[15.4 RECV FAIL] - Failed setting crypto key/nonce."
                                 );
+                                RxState::Idle
+                            } else {
+                                // The crypto operation requires multiple steps through the receiving pipeline and
+                                // an unknown quanitity of time to perform decryption. Holding the 15.4 radio's
+                                // receive buffer for this period of time is suboptimal as packets will be dropped.
+                                // The radio driver assumes the mac.set_receive_buffer(...) funcion is called prior
```
ppannuto, line 610:
```suggestion
                                // The radio driver assumes the mac.set_receive_buffer(...) function is called prior
```

capsules/extra/src/ieee802154/framer.rs
```diff
@@ -567,39 +580,84 @@ impl<'a, M: Mac<'a>, A: AES128CCM<'a>> Framer<'a, M, A> {
...
-                            } else {
-                                let res = self.aes_ccm.crypt(
-                                    buf,
-                                    a_off,
-                                    m_off,
-                                    m_len,
-                                    info.mic_len,
-                                    level.encryption_needed(),
-                                    true,
+                                // No error is returned for the receive function because recv occurs implicitly
```
ppannuto, line 600: Does this case need to return the buffer to the radio driver? i.e., `self.mac.set_receive_buffer(???);`

ppannuto (commented):
Quick review, will try to look more in-depth later. Overall this looks great, and rationale is very clear, thank you for the extended description here.

One question: I can't remember the spec well enough, but the short ID is the bottom 16 of the long address by definition, right? Would it make more sense to only take the long address as an argument for component / framer instance creation, and then have the framer slice out the bottom two bytes whenever it needs the short address? [Or are there cases where the short ID can be something else?]

chips/nrf52/src/ficr.rs
```diff
@@ -411,16 +411,10 @@ impl Ficr {
...
-        let lo = self
-            .registers
-            .deviceaddr0
-            .read(DeviceAddress0::DEVICEADDRESS);
-        let hi = self
-            .registers
-            .deviceaddr1
-            .read(DeviceAddress1::DEVICEADDRESS) as u16;
-        let mut addr = [0; 6];
+    pub fn address(&self) -> [u8; 8] {
```
bradjc, line 414: Why not just add a new function `id()`?

capsules/extra/src/net/ipv6/ipv6_recv.rs
```diff
@@ -71,7 +71,8 @@ impl<'a> SixlowpanRxClient for IP6RecvStruct<'a> {
         }
         match IP6Header::decode(buf).done() {
             Some((offset, ip6_header)) => {
-                let checksum_result = ip6_header.check_transport_checksum(&buf[offset..len]);
+                let checksum_result =
+                    ip6_header.check_transport_checksum(&buf[offset..offset + len]);
```
bradjc, line 75: I'm a little confused by this, why is it valid to index into buf beyond the received `len`?

capsules/extra/src/ieee802154/framer.rs
```diff
@@ -706,12 +759,19 @@ impl<'a, M: Mac<'a>, A: AES128CCM<'a>> MacDevice<'a> for Framer<'a, M, A> {
         // TODO: For Thread, in the case of `KeyIdMode::Source4Index`, the source
         // address should instead be some constant defined in their
         // specification.
+
+        // copy function parameter src_addr as a mut variable so that we can update the src address to extended address if we are encrypting
+        let mut src_addr = src_addr;
```
bradjc, line 764: Does it work to mark `src_addr` `mut` in the function signature?

capsules/extra/src/net/ipv6/ipv6_recv.rs
tyler-potyondy: The `len` field is only the length of the ip6_header. `offset` is where the header begins so `offset+len` constitutes the entire packet. 

For non-encrypted 15.4 packets, the index for len extends beyond the passed buf. This has to do with the way the offsets are calculated in the framer. This comment from a previous implementation addressed the issue of the index being too long, but reasoned that it was okay. I originally looked into why this was occurring, but could not find a satisfactory answer. I copied the comment below. I'll look some more into this. 

Previous Comment from `framer.rs`:
    ```
                // Note: there is a complication here regarding the offsets.
                // When the received frame has security enabled, the payload
                // (including the payload IEs) is encrypted, and hence the data
                // payload field includes the encrypted payload IEs too.
                // However, when the frame is not encrypted, the data payload
                // field does not include the payload IEs.
                //
                // This is fine because we re-parse the unsecured frame before
                // exposing it to the user. At that time, the data payload field
                // will not include the payload IEs.
```

capsules/extra/src/ieee802154/framer.rs
tyler-potyondy: Yes it should. Good catch!

tyler-potyondy:
> One question: I can't remember the spec well enough, but the short ID is the bottom 16 of the long address by definition, right? Would it make more sense to only take the long address as an argument for component / framer instance creation, and then have the framer slice out the bottom two bytes whenever it needs the short address? [Or are there cases where the short ID can be something else?]

@ppannuto Looking further into the spec, I am not actually seeing any definition of the short address as the bottom 16 of the long address. It _appears_, from briefly looking around, that the short address is mainly used/assigned by the PAN (except during broadcasts (0xFFFF). If anyone knows definitively on this matter, please let me know.

The Tock 15.4 implementation seems to be a partial combination of a "PAN" network with the add/remove neighbor and generic sending (the short address is likely an artifact of this). Long story short, it seems that maybe we should not even set the short address in the first place. We likely need to come up with a more complete plan/vision for 15.4 (discussion for the networking WG). The changes in this PR are more of a band-aide for the time being.

To answer your question, yes I think it makes sense to simplify this and have only the long address as an argument to the component to initialize the short address. Longer term, the decisions on how 15.4 should be structured with the short address will impact how framer implements getting/setting the short address.

capsules/extra/src/net/ipv6/ipv6_recv.rs
bradjc: I am certainly no expert on this stack. I tried to find where `SixlowpanRxClient` is used, and I found it in sixlowpan_state.rs. 

The receive callback provides dgram_size:

https://github.com/tock/tock/blob/d068b0549e847430e79637c4d94ca125cf06c343/capsules/extra/src/net/sixlowpan/sixlowpan_state.rs#L775-L779

Which seems to primarily be set by start_receive:

https://github.com/tock/tock/blob/d068b0549e847430e79637c4d94ca125cf06c343/capsules/extra/src/net/sixlowpan/sixlowpan_state.rs#L924-L930

It seems like if you are allowed to read beyond that len then the api is bad and we should fix that, or change the calling side.

tyler-potyondy:
@bradjc  Sorry for the delay looking into this. I've been pulled away working on configuring/setting up some testing infrastructure for a Thread network. 

Looking further into this and reading through the 6LoWPAN/IPV6 layers, I think we have totally missed the point with the discussion we are having regarding the 6LoWPAN/IPV6 recv function changes. I think we should remove those changes and leave the 6LoWPAN as is for this PR. I originally made the changes to prevent a kernel panic that was occurring for encrypted packets. However, this packet should never arrive at the IPV6 function. The current networking stack doc states:

> Any received packet is passed to ALL MacUsers, which are expected to filter packets themselves accordingly.

These packets should be filtered out at the 6LoWPAN layer and never be passed to IPV6 where they are panicking. I propose we remove these changes and perhaps make a separate PR fixing the 6LoWPAN issue of not filtering packets correctly. This PR will not be functional without making the changes to the 6LoWPAN filtering so I think we should create/merge a 6LoWPAN fix and then finalize this PR. Please let me know your thoughts.

bradjc (commented):
I'm not familiar enough with this to really _approve_, but I don't see anything which stands out as an issue.

ppannuto (approved):

PR #3654, opened 2023-09-04
capsules: BUGFIX: backspace/delete would send an extra EOL char

george-cosma:
### Pull Request Overview

Whilst working on [tockloader-rs/listen-prototype](https://github.com/tock/tockloader-rs/pull/2) I've noticed a few peculiarities with the handling of backspace & delete. To cut a very long story short - every time backspace/delete/ANSI Delete is pressed in the process-console, an additional "null"/"end of line" byte is send. Python tockloader does not have an issue with this because its terminal emulator automatically converts it into a unicode character. The issue comes when developing for tockloader-rs where there isn't (or I haven't found) a fully-functional terminal emulator. As such, all bytes are treated as-is.

This also means null is a by out-of-place. Tock sends <null><backspace><space><backspace>, which in effect should just delete the last byte. For the Unicode Null this works. For run-of-the-mill null, this doesn't work, and in fact it deletes the byte behind the <null>. I don't know why this happens, but the same effect can be had by just sending <space><backspace>.

Moreover, delete (ascii) was handled the same as backspace, which is confusing. ANSI Escape Sequence Delete is being handled properly (minus the null shenanigans). I've just made a logical short-circuit to make the code believe it is dealing with the escape sequence rather than a del byte.

### Testing Strategy

I tested this using [this branch](https://github.com/george-cosma/tockloader-rs/tree/no-magic) of tockloader-rs, without the fixes implemented in the aforementioned pull request.

`cargo run -- listen`

Board: microbit v2

### TODO or Help Wanted

The short-circuit condition to make a "DEL" char be treated the same as the escape sequence is a bit wonky. It works, but it I'd like a second opinion if we can leave it as-is or maybe extract the handling of "delete" (either) into a function or something else.

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/core/src/process_console.rs
+++ b/capsules/core/src/process_console.rs
@@ -1211,7 +1211,7 @@ impl<'a, const COMMAND_HISTORY_LEN: usize, A: Alarm<'a>, C: ProcessManagementCap
                 0 => debug!("ProcessConsole had read of 0 bytes"),
                 1 => {
                     self.command_buffer.map(|command| {
-                        let esc_state = self.esc_state.get().next_state(read_buf[0]);
+                        let mut esc_state = self.esc_state.get().next_state(read_buf[0]);
                         self.esc_state.set(esc_state);
 
                         let previous_byte = self.previous_byte.get();
@@ -1219,6 +1219,14 @@ impl<'a, const COMMAND_HISTORY_LEN: usize, A: Alarm<'a>, C: ProcessManagementCap
                         let index = self.command_index.get();
 
                         let cursor = self.cursor.get();
+
+                        // If we get the ANSI "DELETE" byte, we want to treat it the same as
+                        // the "delete" escape sequence. This is a dirty short-circuit.
+                        if read_buf[0] == DEL {
+                            if let EscState::Bypass = esc_state {
+                                esc_state = EscState::Complete(EscKey::Delete);
+                            }
+                        }
 
                         if let EscState::Complete(key) = esc_state {
                             match key {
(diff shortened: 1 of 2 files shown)


capsules/core/src/process_console.rs
```diff
@@ -1220,6 +1220,14 @@ impl<'a, const COMMAND_HISTORY_LEN: usize, A: Alarm<'a>, C: ProcessManagementCap
...
                         let cursor = self.cursor.get();
 
+                        // If we get the ANSI "DELETE" byte, we want to treat it the same as
+                        // the "delete" escape sequence. This is a dirty short-circuit.
+                        if read_buf[0] == DEL {
+                            if let EscState::Bypass = esc_state {
+                                esc_state = EscState::Complete(EscKey::Delete);
+                            }
+                        }
+
```
bradjc, line 1230: It seems like this logic should just go inside the `next_state()` function for EscState.

bradjc (commented):
It would also be great if you could add this case to https://github.com/tock/tock/blob/master/tools/check_process_console.py

george-cosma:
I've moved the "DEL" logic to next_state() and I've modified the "deleting" test to run same checks for both types of delete sequences.

I also had to modify the evaluation function of the tests because they were interpreting "<space><backspace>" quite literally. Of course, in actuality this has no effect over the actual output as all it does is cover up a remaining character. As an alternative to it, I've tried to send an escape sequence to clear the line, but no luck. Neither versions of tockloader would actually interpret the escape sequence.

bradjc (approved):

PR #3657, opened 2023-09-07
Add `pub type` to components

bradjc:
### Pull Request Overview

This is an alternative to #3589 

The motivation remains the same: types for structs in main.rs get unwieldy and repetitive. Our current approach of using `dyn` traits is probably not what we want. If components can help with types, and make it easier to add resources to a board that is an improvement.

The major change for this version is that components provide templated types, and boards create the specific version for their use at the top of the main.rs. Then boards can use those types in both the `Platform` struct and when creating other components.

The downside is that the approach is somewhat brute force. We have to use the `type` keyword a fair bit. There is still redundancy, although much less. Some of the "magic" with macros is lost. In particular, there is embedded redundancy when defining both a new `type` and the macro to finalize.

The advantage is this doesn't add more macros and there's less "convention" that component authors need to follow to create easy to understand and reuse components.

### Testing Strategy

A few changes to nrf52840dk board as a test.

### TODO or Help Wanted

Is this a step forward? In the worst case (kv stack) it saves hundreds of lines in redundant type signatures. It won't have nearly as much of an impact with other components. That is sometimes because some components encapsulate several steps (ie creating a virtual device) and also because we use `dyn` quite a bit.

### Documentation Updated

### Formatting

- [ ] Ran `make prepush`.

### DIFF
--- a/boards/components/src/alarm.rs
+++ b/boards/components/src/alarm.rs
@@ -57,6 +57,11 @@ macro_rules! alarm_component_static {
     };};
 }
 
+pub type AlarmDriverComponentType<A> = capsules_core::alarm::AlarmDriver<
+    'static,
+    capsules_core::virtualizers::virtual_alarm::VirtualMuxAlarm<'static, A>,
+>;
+
 pub struct AlarmMuxComponent<A: 'static + time::Alarm<'static>> {
     alarm: &'static A,
 }
--- a/boards/components/src/kv.rs
+++ b/boards/components/src/kv.rs
@@ -30,6 +30,8 @@ macro_rules! kv_driver_component_static {
     };};
 }
 
+pub type KVDriverComponentType<V> = capsules_extra::kv_driver::KVStoreDriver<'static, V>;
+
 pub struct KVDriverComponent<V: hil::kv::KVPermissions<'static> + 'static> {
     kv: &'static V,
     board_kernel: &'static kernel::Kernel,
@@ -84,6 +86,9 @@ macro_rules! kv_permissions_mux_component_static {
     };};
 }
 
+pub type KVPermissionsMuxComponentType<V> =
+    capsules_extra::kv_store_permissions::KVStorePermissions<'static, V>;
+
 pub struct KVPermissionsMuxComponent<V: hil::kv::KVPermissions<'static> + 'static> {
     kv: &'static V,
 }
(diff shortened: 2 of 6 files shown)


lschuermann:
I think this is a strict improvement over #3589. Analog to how components encapsulate an instantiation of a given peripheral, these type aliases provide a template for composing the nested type signatures describing the types created by the respective components.

As we away from trait objects towards generics (which enable significant gateway optimizations) as in #3681, these types will get increasingly unwieldy.

hudson-ayers (approved):
I am a big fan of this change. Anything that makes it easier to remove `dyn` when possible is something I am in favor of

phil-levis (commented):
I think this is a big improvement. Complexity or verbosity in declaration is better than verbosity in use.

bradjc:
I tried what I was talking about on the call. It doesn't help much.

```diff
diff --git a/boards/components/src/kv.rs b/boards/components/src/kv.rs
index 0597eb807..3939d7153 100644
--- a/boards/components/src/kv.rs
+++ b/boards/components/src/kv.rs
@@ -15,6 +15,16 @@ use kernel::component::Component;
 use kernel::create_capability;
 use kernel::hil;

+///////////////////////
+// KV Stack
+///////////////////////
+
+pub type KVDriverStackType<K, T> = KVDriverComponentType<
+    VirtualKVPermissionsComponentType<
+        KVStorePermissionsComponentType<TicKVKVStoreComponentType<K, T>>,
+    >,
+>;
+
 ///////////////////////
 // KV Userspace Driver
 ///////////////////////
diff --git a/boards/nordic/nrf52840dk/src/main.rs b/boards/nordic/nrf52840dk/src/main.rs
index 84bb83bfd..719fac12f 100644
--- a/boards/nordic/nrf52840dk/src/main.rs
+++ b/boards/nordic/nrf52840dk/src/main.rs
@@ -179,7 +179,8 @@ type TicKVKVStore = components::kv::TicKVKVStoreComponentType<
 >;
 type KVStorePermissions = components::kv::KVStorePermissionsComponentType<TicKVKVStore>;
 type VirtualKVPermissions = components::kv::VirtualKVPermissionsComponentType<KVStorePermissions>;
-type KVDriver = components::kv::KVDriverComponentType<VirtualKVPermissions>;
+type KVDriver =
+    components::kv::KVDriverStackType<TicKVDedicatedFlash, capsules_extra::tickv::TicKVKeyType>;

 /// Supported drivers by the platform
 pub struct Platform {
```

Arguably it's worse.

alevy:
Why is this still an RFC? I think all signs point to this being an improvement, and #3681 seems to corroborate. Perhaps there was discussion on the call that's missing here that would jog our memory?

alevy (approved):

lschuermann (approved):

PR #3660, opened 2023-09-07
6LoWPAN Packet Filtering

tyler-potyondy:
### Pull Request Overview

A received 15.4 packet is passed to both the radio driver and 6LoWPAN RxClient. This is in accordance with the networking stack documentation stating: 

> Any received packet is passed to ALL MacUsers, which are expected to filter packets themselves accordingly.

The current 6LoWPAN module checks if the packet is in fact a 6LoWPAN packet, but incorrectly copies the packet/passes the packet along to the higher network layers when the packet is not 6LoWPAN (i.e. simply a generic 15.4 packet containing data). This is incorrect behavior. This PR resolves this issue with a minor change.

### Testing Strategy

This pull request was tested by running the 15.4 rx/tx (both encrypted and non encrypted) libtock-c example tests. 

### TODO or Help Wanted

N/A

### Documentation Updated

- [ x] Updated the relevant files in `/docs`, or no updates are required.

### Formatting

- [ x] Ran `make prepush`.

### DIFF
--- a/capsules/extra/src/net/sixlowpan/sixlowpan_state.rs
+++ b/capsules/extra/src/net/sixlowpan/sixlowpan_state.rs
@@ -955,7 +955,7 @@ impl<'a, A: time::Alarm<'a>, C: ContextStore> Sixlowpan<'a, A, C> {
                     }
                 }
             } else {
-                packet[0..payload_len].copy_from_slice(&payload[0..payload_len]);
+                return (None, Ok(()));
             }
             state.packet.replace(packet);
             (Some(state), Ok(()))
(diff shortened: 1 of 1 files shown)


capsules/extra/src/net/sixlowpan/sixlowpan_state.rs
```diff
@@ -955,7 +955,7 @@ impl<'a, A: time::Alarm<'a>, C: ContextStore> Sixlowpan<'a, A, C> {
                     }
                 }
             } else {
-                packet[0..payload_len].copy_from_slice(&payload[0..payload_len]);
+                return (None, Ok(()));
```
ppannuto, line 958: Having peed at Google yesterday, I am reminded by their bathroom coding style flyer of a comment Johnathan has made on a few PRs. This would be slightly cleaner switched to a "guard" idiom, i.e. move

```rust
if !is_lowpan(payload) {
    return (None, OK());
}

// ... other code continues here not in an else block
```

ppannuto (changes requested):

ppannuto (approved):

PR #3663, opened 2023-09-08
Labeler updates

brghena:
### Pull Request Overview

This pull request changes auto-labeling GitHub action to apply WG-Network to relevant files. Since the Network WG isn't entirely contained in certain directories, I've added individual files in many places.

Also adds the WG-OpenTitan label to their documentation directory.

### TODO or Help Wanted

Can @lschuermann and @alexandruradovici take a brief skim to make sure I haven't missed obvious files?

Can @bradjc or @jrvanwhy confirm the WG-OpenTitan change? I'll remove it if you don't want it.

### Documentation Updated

### Formatting

- [N/A] Ran `make prepush`.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -32,6 +32,7 @@ WG-OpenTitan:
   - boards/opentitan/**/*
   - chips/earlgrey/**/*
   - chips/lowrisc/**/*
+  - doc/wg/opentitan/**/*
 
 # add kernel label unless already covered by hil label
 kernel:
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -34,6 +34,25 @@ WG-OpenTitan:
   - chips/lowrisc/**/*
   - doc/wg/opentitan/**/*
 
+WG-Network
+  - capsules/extra/src/ble_advertising_driver.rs
+  - capsules/extra/src/can.rs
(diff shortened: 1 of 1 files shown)


brghena:
image

Why bot? Please don't. It's exactly as Brad suspected.

bradjc (approved):
Looks good to me.

brghena:
Looking through the [labeler repo](https://github.com/actions/labeler), we could disable ever removing labels with the `sync-labels` option. Or we can leave it as-is and have it auto-remove labels. There doesn't seem to be an in-between option that could choose on a per-label basis.

Keeping manually set labels would also suffice, but the option doesn't exist currently: https://github.com/actions/labeler/issues/624

hudson-ayers:
> Looking through the [labeler repo](https://github.com/actions/labeler), we could disable ever removing labels with the `sync-labels` option. Or we can leave it as-is and have it auto-remove labels. There doesn't seem to be an in-between option that could choose on a per-label basis.
> 
> Keeping manually set labels would also suffice, but the option doesn't exist currently: [actions/labeler#624](https://github.com/actions/labeler/issues/624)

Even with `sync-labels=false`, the labeler still removes manually added labels (https://github.com/actions/labeler/issues/624#issuecomment-1682626478), unfortunately.

hudson-ayers (approved):

PR #3664, opened 2023-09-08
clippy: fix some pedantic lints

bradjc:
### Pull Request Overview

```diff 
- -A clippy::bool_to_int_with_if
- -A clippy::inconsistent_struct_constructor
```

### Testing Strategy

travis

### TODO or Help Wanted

Ok with enforcing these?

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/boards/apollo3/lora_things_plus/src/main.rs
+++ b/boards/apollo3/lora_things_plus/src/main.rs
@@ -472,7 +472,7 @@ unsafe fn setup() -> (
             humidity,
             air_quality,
             scheduler,
-            systick
+            systick,
         }
     );
 
--- a/boards/apollo3/lora_things_plus/src/main.rs
+++ b/boards/apollo3/lora_things_plus/src/main.rs
@@ -460,9 +460,9 @@ unsafe fn setup() -> (
         LoRaThingsPlus,
         LoRaThingsPlus {
             alarm,
-            console,
-            gpio,
             led,
+            gpio,
+            console,
             i2c_master,
             external_spi_controller,
             sx1262_spi_controller,
@@ -472,7 +472,7 @@ unsafe fn setup() -> (
             humidity,
             air_quality,
             scheduler,
-            systick,
+            systick
         }
     );
 
--- a/boards/components/src/proximity.rs
+++ b/boards/components/src/proximity.rs
@@ -38,9 +38,9 @@ impl<P: hil::sensors::ProximityDriver<'static>> ProximityComponent<P> {
         driver_num: usize,
     ) -> ProximityComponent<P> {
         ProximityComponent {
+            sensor,
             board_kernel,
             driver_num,
-            sensor,
         }
     }
 }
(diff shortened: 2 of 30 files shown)


boards/apollo3/lora_things_plus/src/main.rs
```diff
@@ -472,7 +472,7 @@ unsafe fn setup() -> (
             humidity,
             air_quality,
             scheduler,
-            systick,
```
ppannuto, line 475: ugh—can we set this one to the opposite? Missing trailing commas on arbitrarily-last entries just lead to noise in the diff and history; I think these should be required.

boards/components/src/proximity.rs
```diff
@@ -38,9 +38,9 @@ impl<P: hil::sensors::ProximityDriver<'static>> ProximityComponent<P> {
         driver_num: usize,
     ) -> ProximityComponent<P> {
         ProximityComponent {
+            sensor,
             board_kernel,
             driver_num,
```
ppannuto, line 43: Now I'm just confused — it's not even a consistent change? Here it leaves the `,`?

capsules/extra/src/pwm.rs
```diff
@@ -30,7 +30,7 @@ impl<'a, const NUM_PINS: usize> Pwm<'a, NUM_PINS> {
         pwm_pins: &'a [&'a dyn hil::pwm::PwmPin; NUM_PINS],
         grant: Grant<App, UpcallCount<1>, AllowRoCount<0>, AllowRwCount<0>>,
     ) -> Pwm<'a, NUM_PINS> {
-        assert!(NUM_PINS <= (u16::MAX as usize));
+        assert!(u16::try_from(NUM_PINS).is_ok());
```
ppannuto, line 33: I find this an impressive conversion for the tool to make — I like the change, but had to read it twice to convince myself the effect is the same.

ppannuto (changes requested):
On first read, I was confused by the `u8::from(boolean variable) << shift` pattern, but thinking a bit more, I think the explicit conversion to the `u8` `u32` etc is superior to our prior pattern of if/else's returning constants which the compiler interpreted as-needed.

I ~~assume~~ the compiler is smart enough that adding 'functions' to convert doesn't actually add overhead. Looking at size report, this PR adds 8 bytes total, which all seem to be padding / layout related.

Think this is all good but for the comma thing

boards/apollo3/lora_things_plus/src/main.rs
bradjc: I'm actually a little confused as well. I thought that rustfmt enforced having the trailing `,`. I wonder if this has something to do with the macro?

The clippy lint is just checking that the fields match the definition in order.

bradjc: This might be relevant: https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#trailing_comma

bradjc:
Good with this?

PR #3667, opened 2023-09-09
tickv: handle additional read

bradjc:
### Pull Request Overview

If the first read does not find a key we may need to do an additional read.

This wasn't handled before because the search feature wasn't working.

### Testing Strategy

Porting HTOP to kv.

### TODO or Help Wanted

n/a

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.

### DIFF
--- a/capsules/extra/src/tickv.rs
+++ b/capsules/extra/src/tickv.rs
@@ -474,6 +474,9 @@ impl<'a, F: Flash, H: Hasher<'a, 8>, const PAGE_SIZE: usize> flash::Client<F>
                     }
                     Err(tickv::error_codes::ErrorCode::ReadNotReady(_)) => {
                         // Need to do another flash read.
+                        //
+                        // `self.operation` will still be `GetKey`, so this will automatically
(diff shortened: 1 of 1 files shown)


capsules/extra/src/tickv.rs
```diff
@@ -472,6 +472,9 @@ impl<'a, F: Flash, H: Hasher<'a, 8>, const PAGE_SIZE: usize> flash::Client<F>
                             );
                         });
                     }
+                    Err(tickv::error_codes::ErrorCode::ReadNotReady(_)) => {
+                        // Need to do another flash read.
```
ppannuto, line 476:
```suggestion
                        // Need to do another flash read.
                        //
                        // `self.operation` will still be `GetKey`, so this will automatically
                        // be retried by the primary state machine.
```
I think this is why this works? I thought this was a "todo: need to do another read" looking at this at first — the implicit retry isn't terribly obvious.

ppannuto (approved):

PR #3680, opened 2023-09-18
boards: fix clippy ==true

bradjc:
### Pull Request Overview

A couple more of these for the lint.

### Testing Strategy

Uh, good question. Why doesn't our CI catch these?!?

### TODO or Help Wanted

nothing, hopefully

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


ppannuto (approved):
I think CI let it slide by because clippy is currently marked as a non-required status to merge.

Once all the dust settles on clippy settings, we should probably make it required.

I suspect it would be very annoying in practice as a required check to have the formatting requirements moving underneath other PRs, however.

PR #3682, opened 2023-09-20
Bug Fixes for Hs3003 sensor

TheButterMineCutter:
### Pull Request Overview

This pull request moves the functionality of the last `State::Sleep` section into the `State::Read` section. Having the functionality in the `State::Sleep` section did not make the capsule work properly. Also divided the final temperature calculation by 10. This doesn't follow the datasheet but will otherwise return 10 times the actual temperature.

### Testing Strategy

Tested on the onboard hs3003 sensor of the `Nano 33 BLE Sense Rev2`.

### TODO or Help Wanted

N/A

### Documentation Updated

### Formatting

- [x] Ran `make prepush`.


capsules/extra/src/hs3003.rs
```diff
@@ -186,7 +186,10 @@ impl<'a, I: I2CDevice> I2CClient for Hs3003<'a, I> {
                     self.pending_humidity.set(false);
                     self.humidity_client.map(|client| client.callback(humidity));
                 }
+
+                self.state.set(State::Sleep(temperature, humidity));
```
bradjc, line 190: Do we need the two values with the `Sleep` state or can those be removed?

TheButterMineCutter: Your right, there is no need for the `Sleep` state to have values at all since I removed the last `Sleep` state.

bradjc (approved):

